Issue #1469, opened 2026-04-28
bug(isolation/git): worktree sync fails 403 on org repos — per-codebase GH_TOKEN never reaches git fetch subprocess


133Felix:
## Symptom

Multi-org Archon deployment (single `process.env.GH_TOKEN` covering personal repos, per-codebase `GH_TOKEN` rows in `remote_agent_codebase_env_vars` for org repos). Direct-chat operations against an org repo work fine — the orchestrator loads per-codebase env vars and injects them into the AI provider subprocess.

But any workflow dispatch fails before the workflow run is even registered in the DB. User-facing message: the generic `"An unexpected error occurred. Try /reset to start a fresh session."` (Issue #1237 / error-formatter masking — separate). Server log:

```
@archon/server: orchestrator_message_failed
err: Error: Failed to fetch base branch from origin: Sync fetch from origin/main failed:
  Command failed: git -C /.archon/workspaces/<org>/<repo>/source fetch origin main
  remote: Write access to repository not granted.
  fatal: unable to access 'https://github.com/<org>/<repo>.git/': The requested URL returned error: 403
stack:
  at syncWorkspaceBeforeCreate (/app/packages/isolation/src/providers/worktree.ts:810)
  at async createWorktree (/app/packages/isolation/src/providers/worktree.ts:709)
  at async create                (/app/packages/isolation/src/providers/worktree.ts:154)
  at async createNewEnvironment (/app/packages/isolation/src/resolver.ts:468)
```

GitHub's "Write access to repository not granted" message in this context means **no read access either** — the global `process.env.GH_TOKEN` (personal PAT) is being used, and it has zero scope on the org repo.

## Root Cause

`packages/git/src/repo.ts:104` (the `syncWorkspaceBeforeCreate` path that runs *before* every workflow run as part of worktree creation):

```ts
await execFileAsync('git', ['-C', workspacePath, 'fetch', 'origin', branchToSync], {
  timeout: 60000,
});
```

**No `env` parameter.** The spawned `git fetch` subprocess inherits whatever `process.env` the Archon server itself was started with — i.e. the Docker container's `/opt/archon/.env` `GH_TOKEN`. The per-codebase token from `remote_agent_codebase_env_vars` (loaded in `orchestrator-agent.ts:843`) is *never* threaded into this code path.

[...]

Issue #1326, opened 2026-04-21
bug(workflows): bash nodes silently fail on Windows when `bash` resolves to WSL launcher (System32\bash.exe) — $VAR expansion broken in `-c` mode


atlas-architect:
## Summary

On Windows, Archon bash nodes fail silently when `bash` resolves to `C:\Windows\System32\bash.exe` (the WSL launcher shipped with Windows). Any `${VAR}` reference in the bash script expands to empty string, causing downstream logic to fail. The failure appears as "dir does not exist" or empty path errors, which is misleading — the actual root cause is that variable expansion doesn't work in `bash.exe -c 'script'` when bash is WSL's launcher.

**Impact:** every bash node in every workflow fails silently on affected Windows machines. Gather-context succeeds (LLM node), but the first downstream bash node fails. Looks like a workflow YAML bug or user error; actually a shell-resolution bug.

## Repro

Minimal workflow YAML:

```yaml
name: repro
nodes:
  - id: gather
    prompt: "Return JSON {\"name\": \"hello\"}"
    model: haiku
    output_format:
      type: object
      properties:
        name: { type: string }
      required: [name]
  - id: bash-test
    bash: |
      set -e
      X=$gather.output.name
      echo "X=[$X]"
      [ -n "$X" ] && echo "OK" || echo "FAIL"
    depends_on: [gather]
```

Fire on Windows via `archon workflow run repro`. On affected machines, bash-test fails with `X=[]` and exits non-zero. Same YAML runs fine on macOS/Linux and on Windows machines where `bash` in PATH resolves to Git Bash.

## Root cause

Windows `CreateProcess` search order for bare command names (no absolute path):

1. Application's directory
2. Current directory
3. **System directory** (`C:\Windows\System32`) ← `bash.exe` found here (WSL launcher)
4. 16-bit system directory
5. Windows directory
6. **PATH environment variable** ← Git Bash's `bash.exe` is here (if installed)

Because System32 is searched BEFORE PATH, Bun's `child_process.spawn('bash', [...])` always resolves `bash` to `C:\Windows\System32\bash.exe` on any Windows install that has WSL enabled, even when Git Bash's `C:\Program Files\Git\bin` is prepended to PATH. `Get-Command bash` in PowerShell uses a different (PATH-respecting) lookup, so users verify their PATH shows Git Bash and assume their shell resolution is correct — but Bun doesn't see that resolution.

Once `bash.exe` resolves to WSL launcher:

```powershell
[...]

PR #1583, opened 2026-05-05
fix(git): reject SSH-URL-shaped codebase names in worktree path resolution

### LINKED ISSUE #1582: bug(git): SSH-URL codebase names produce worktree paths with ':' that break docker volume parsing
## Summary

A codebase registered with an SSH-style remote URL as its `name` (e.g. `git@host.example:org/repo`) ends up with worktrees rooted at

```
~/.archon/workspaces/git@host.example:org/repo/worktrees/...
```

The colon and `@` in that path silently break any tool that uses `:` as a separator. Most visible failure: docker-compose short-form volume specs inside a devcontainer can no longer be parsed (`HOST:CONTAINER:OPT`) because the host segment itself contains a `:`. Compose returns `invalid volume specification` and the workflow fails.

## Reproduction

1. Register a codebase whose `name` is the bare SSH remote URL, e.g.

   ```
   name = "git@host.example:org/repo"
   default_cwd = "/srv/projects/repo"
   ```

   (this happens automatically if the codebase is registered by remote URL rather than by a normalized owner/repo string).

2. Trigger any workflow run from the **Web UI** (the WebUI passes `codebaseName` through to the worktree resolver, the CLI without `--codebase` typically does not).

3. Inside the workflow, run any bash node that invokes `docker compose ... up -d` against a compose file that uses short-form volumes — e.g. a devcontainer with

   ```yaml
   services:
     db:
       volumes:
         - ./postgres-init:/docker-entrypoint-initdb.d:ro
   ```

4. The workflow fails. The full docker-compose error:

   ```
   Error response from daemon: invalid volume specification:
     '/home/USER/.archon/workspaces/git@host.example:org/repo/worktrees/<branch>/.devcontainer/postgres-init:/docker-entrypoint-initdb.d:ro'
   ```

## Root cause

`packages/git/src/worktree.ts:resolveOwnerRepo()` splits the codebase name at `/` and accepts both halves verbatim:

```ts
if (codebaseName) {
  const parts = codebaseName.split('/');
  if (parts.length === 2 && parts[0] && parts[1]) {
    return { owner: parts[0], repo: parts[1] };
  }
  // …
}
```

For `"git@host.example:org/repo"` this yields `{ owner: "git@host.example:org", repo: "repo" }`. The strict `SAFE_NAME` validator that lives next to the path-building helpers (`packages/paths/src/archon-paths.ts:parseOwnerRepo()`, regex `/^[a-zA-Z0-9._-]+$/`) is **not** consulted here, so characters that cause path-parsing trouble downstream slip through.

CLI runs that fall through to the path-derived heuristic (`extractOwnerRepo(repoPath)` → last two basename segments) produce a clean `<parent>/<repo>` and were never affected — that is why the bug only surfaces on Web-UI-triggered runs (or any caller that passes a bad `codebaseName` through).

## Side effect: error truncation hid this for a while

The bash-node executor head-truncates captured error output when a node fails, keeping the tail. The compose error above starts with `Container <name> Creating` followed by the `invalid volume specification` line — both got cut off, so the persisted run metadata only retained the trailing devcontainer-CLI stack trace ending in `…[truncated]`. Worth a separate issue: bash-node failure truncation should at minimum preserve the first lines of stderr.

## Proposed fix

Have `resolveOwnerRepo()` route the `codebaseName` branch through the existing strict `parseOwnerRepo()` from `@archon/paths`. When the name doesn't pass `SAFE_NAME` on both segments, log `worktree.invalid_codebase_name_format` and fall back to the path-derived heuristic — same fallback already used for the `'invalid-no-slash'` case.

Behaviour delta:

- Valid `owner/repo` codebase names — unchanged.
- Invalid names (no slash, dot-segments, wrong segment count) — unchanged.
- SSH-URL-shaped names with `:`/`@` — previously produced a corrupted workspace path; now logged and falls back to the same path-derived layout the CLI already uses.

A follow-up question worth its own discussion: codebase registration should probably normalize SSH/HTTPS remote URLs to a canonical `owner/repo` slug before storing the `name` field so the resolver never sees the URL form to begin with. That is out of scope for the immediate fix.

## Environment

- Archon `dev`
- Docker Compose v2
- WebUI-triggered workflow runs against codebases registered with SSH-style remote URLs as their name

blankse:
## Summary

- **Problem**: A codebase registered with an SSH-style remote URL as its `name` (e.g. `git@host.example:org/repo`) produces worktree paths like `~/.archon/workspaces/git@host.example:org/repo/worktrees/...`. The `:` in that path silently corrupts any downstream tool that uses `:` as a separator — most visibly docker-compose short-form volume specs (`HOST:CONTAINER:OPT`) inside devcontainers, which then fail with `invalid volume specification`.
- **Why it matters**: The bug only surfaces on Web-UI-triggered runs (and any other caller that passes `codebaseName` through), so users see the same workflow succeed via CLI and fail via Web UI for no obvious reason. The bash-node executor head-truncates failure output, so the actual `invalid volume specification` line gets dropped — diagnosis is painful.
- **What changed**: `resolveOwnerRepo()` now routes the `codebaseName` branch through the existing strict `parseOwnerRepo()` validator from `@archon/paths` (`SAFE_NAME = /^[a-zA-Z0-9._-]+$/`). Invalid names log `worktree.invalid_codebase_name_format` and fall through to the path-derived heuristic — same fallback that already covered the `'invalid-no-slash'` case.
- **What did not change (scope boundary)**: Valid `owner/repo` codebase names behave identically. The path-derived fallback (`extractOwnerRepo(repoPath)`) is unchanged. No public API or schema change. No change to codebase registration / persistence — that normalization is a separate, larger discussion (called out at the end of #1582).

### Before

```
User                Archon                                   docker compose
────                ──────                                   ──────────────
register codebase
  name="git@host.example:org/repo" ──▶ stored verbatim
  default_cwd=/srv/projects/repo

trigger workflow (Web UI) ───────────▶ resolveOwnerRepo(codebaseName)
                                       split('/') → owner="git@host.example:org",
                                                    repo ="repo"
                                       worktree base:
                                       ~/.archon/workspaces/git@host.example:org/repo/...

bash node `start-preview` runs:
  docker compose up -d (in worktree) ──────────────────────▶ parses volumes
                                                              `<host>:/docker-entrypoint-initdb.d:ro`
                                                              host contains `:`
                                                              ✗ "invalid volume specification"
sees "Bash node failed [exit 1]"
(real reason head-truncated away)
```

### After

```
User                Archon                                   docker compose
────                ──────                                   ──────────────
register codebase
  name="git@host.example:org/repo"   (unchanged — fix is purely on the resolver path)
  default_cwd=/srv/projects/repo

trigger workflow (Web UI) ───────────▶ resolveOwnerRepo(codebaseName)
                                       [parseOwnerRepo("git@host.example:org/repo")]
                                       [→ null (`:`/`@` rejected by SAFE_NAME)]
                                       [warn: worktree.invalid_codebase_name_format]
                                       [fallback → extractOwnerRepo(repoPath)]
                                       [owner="projects", repo="repo"]
                                       worktree base:
                                       ~/.archon/workspaces/projects/repo/...   ✓ no `:`

bash node `start-preview` runs:
  docker compose up -d ────────────────────────────────────▶ parses volumes
                                                              host has no `:`
                                                              ✓ container starts
```

### Before

```
packages/git/src/worktree.ts
  resolveOwnerRepo(repoPath, codebaseName?)
   ├─ if codebaseName
   │   └─ split('/') → naive owner/repo  ───── (no SAFE_NAME check)
   │      └─ returned even if owner contains `:` `@` etc.
   ├─ else if repoPath under workspaces/
   │   └─ derive from path segments
   └─ else extractOwnerRepo(repoPath)            (path-basename heuristic)
                              │
                              ▼
                    getProjectWorktreesPath(owner, repo)
                              │
                              ▼
                    consumed as filesystem path  ── docker compose volume parser breaks on `:`
```

### After

```
packages/git/src/worktree.ts
  resolveOwnerRepo(repoPath, codebaseName?)
   ├─ if codebaseName
   │   └─ [parseOwnerRepo(codebaseName)]            === @archon/paths (existing)
   │       ├─ valid → returned  (unchanged behaviour for clean names)
   │       └─ invalid → [warn, fall through to path heuristic]
   ├─ else if repoPath under workspaces/
   │   └─ derive from path segments              (unchanged)
   └─ else extractOwnerRepo(repoPath)            (unchanged)
                              │
                              ▼
                    getProjectWorktreesPath(owner, repo)   (unchanged)
```

**Connection inventory**:

| `worktree.ts` | `@archon/paths` | **modified** | adds `parseOwnerRepo` to existing import |
| `resolveOwnerRepo` | `parseOwnerRepo` | **new** | replaces inline `split('/')` |
| `resolveOwnerRepo` | path-derived fallback | unchanged | now reachable for one additional input shape |
| `getWorktreeBase` callers | `resolveOwnerRepo` | unchanged | public surface unchanged |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `git`
- Module: `git:worktree`

## Change Metadata

- Change type: `bug`
- Primary scope: `git`

## Linked Issue

- Closes #1582
- Related # —
- Depends on # —
- Supersedes # —

```bash
bun run check:bundled                          # bundled-defaults up to date
bun --filter @archon/git type-check            # exit 0
bun --filter @archon/git test                  # 143 pass / 0 fail (was 142, +1 new regression test)
bun --filter @archon/isolation type-check      # exit 0 (downstream consumer)
bun --filter @archon/isolation test            # all green
bun x eslint packages/git/src/worktree.ts --max-warnings 0    # clean
bun x prettier --check packages/git/src/{worktree,git.test}.ts  # clean
```

- **Evidence provided**: New test `'ignores SSH-URL-shaped codebaseName ... and falls back to path'` in `packages/git/src/git.test.ts` exercises the regression directly and asserts the resulting worktree base contains neither `:` nor `@`. Mirrors the structure of the existing `'invalid-no-slash'` test. The test mock for `@archon/paths` now also exposes a `parseOwnerRepo` mirror kept aligned with the real `SAFE_NAME` regex.
- **Skipped**: Top-level `bun run lint` was skipped because root-level ESLint OOMs on this machine (V8 heap exhaustion in the monorepo lint run — independent of this PR). Linted the two touched files directly instead.

## Security Impact (required)

- New permissions/capabilities? **No**.
- New external network calls? **No**.
- Secrets/tokens handling changed? **No**.
- File system access scope changed? **No** — same `getProjectWorktreesPath()` output, only the input owner/repo segments are now constrained to `[A-Za-z0-9._-]`. If anything, this *narrows* the set of paths the resolver can produce (path-traversal-safe by construction, since `SAFE_NAME` already excludes `.`/`..`).

## Compatibility / Migration

- Backward compatible? **Yes** — purely additive validation. Codebases with valid `owner/repo` names see no change.
- Config/env changes? **No**.
- Database migration needed? **No**.
- For users with already-existing worktrees under a corrupted `git@host:org/repo/...` path: those keep working (git doesn't care), but new worktrees for the same codebase will land under the path-derived layout instead. Old worktree dirs can be cleaned up manually if desired.

- **Verified scenarios**:
  - Local repro on a real WebUI-triggered workflow run against a codebase whose `name` contained an SSH URL — failure mode (`invalid volume specification` from docker-compose) confirmed against `dev`.
  - After the fix, `resolveOwnerRepo("git@host.example:org/repo")` returns the path-derived owner/repo, and the resulting worktree base contains no `:` / `@`.
  - All 142 existing `@archon/git` tests still pass; new regression test passes.
- **Edge cases checked**:
  - `codebaseName` already valid (e.g. `acme/widget-app`) → unchanged.
  - `codebaseName` with multiple slashes (e.g. `acme/sub/repo`) → unchanged (still rejected by `parseOwnerRepo` length check, falls through).
  - `codebaseName` with leading/trailing dots → unchanged (rejected by `SAFE_NAME`, falls through).
  - `codebaseName` with whitespace, `+`, `~`, etc. → now also rejected (was previously accepted as a path segment); this is the intended tightening.
- **What was not verified**:
  - Codebase registration flow (where SSH URLs end up as `name` in the first place) — out of scope; called out as a follow-up in #1582.
  - Cross-platform path semantics on Windows — only verified on Linux.

- **Affected subsystems**: only `getWorktreeBase()` resolution (and any code that depends on the worktree base layout — `@archon/isolation`, workflow run dispatch, artifact paths derived from the worktree). All consumers ride the same path-derived fallback the CLI already used, so behaviour for new runs converges with the previously-working CLI behaviour.
- **Potential unintended effects**: For users whose codebase `name` happened to contain non-alphanumeric chars but **wasn't** an SSH URL and where the corrupted path was somehow load-bearing — they'll now silently get the path-derived layout instead. Mitigated by the warn-log (`worktree.invalid_codebase_name_format`) and by the fact that the corrupted layout was never functional (it's the bug being fixed).
- **Guardrails**: existing test for `'invalid-no-slash'` remains; new test pins the SSH-URL shape; warn-level log event makes the fallback observable in operations.

## Rollback Plan (required)

- **Fast rollback**: revert this single commit; the change is two files, additive validation only.
- **Feature flags**: none.
- **Observable failure symptoms**: a regression would re-surface as `invalid volume specification` in docker-compose volume parsing (or any other consumer that splits on `:`). Logged event `worktree.invalid_codebase_name_format` lets operators see when the fallback fires.

## Risks and Mitigations

- **Risk**: An existing codebase relies on the corrupted owner/repo path being preserved for some downstream lookup (e.g. cache key, DB FK).
  - **Mitigation**: searched the codebase — every consumer of the worktree base path treats it as an opaque filesystem location; no DB rows or external integrations key on the corrupted form. The CLI path has been emitting the path-derived layout all along, so any "stable across both triggers" expectation was already broken.
- **Risk**: Hides a class of bugs by silently falling back instead of erroring.
  - **Mitigation**: the warn-log is intentional (matches the `'invalid-no-slash'` precedent two lines above). Erroring out at the resolver level would break legitimate CLI runs that pass any unparseable name; a separate registration-time validator (mentioned in #1582) is the right place to enforce this strictly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened validation of repository identifiers so SSH-style or otherwise unsafe names are rejected; now reliably falls back to workspace-scoped alternatives and emits clearer warnings for invalid formats.

* **Tests**
  * Added a regression test confirming malformed or SSH-style repository names are rejected and fallback derivation produces safe base paths (no ":" or "@").


Wirasm:
@$blankse related to #1469 — overlapping area or partial fix.

Wirasm:
@$blankse related to #1326 — overlapping area or partial fix.

Wirasm:
@$blankse related to #1582 — overlapping area or partial fix.

Wirasm:
@$blankse related to #1469 — overlapping area or partial fix.

Wirasm:
@$blankse related to #1326 — overlapping area or partial fix.

Wirasm:
@$blankse related to #1582 — overlapping area or partial fix.

Wirasm:
Hi @blankse — thanks for opening this PR.

- Human Verification
- Side Effects / Blast Radius

blankse:
Hi @Wirasm — thanks for the nudge. Both sections are actually filled out in the PR description; I think the scan may have keyed on the headings only and missed the body beneath. Quick pointers:

- **Human Verification (required)** — covers verified scenarios (local repro of the docker-compose `invalid volume specification` failure on `dev`, confirmation that `resolveOwnerRepo("git@host.example:org/repo")` falls back to the path-derived layout, 142+1 tests green), edge cases (`acme/widget-app`, multi-slash names, leading/trailing dots, whitespace/`+`/`~`), and what was *not* verified (the registration-flow side, called out as follow-up in #1582; Windows path semantics).
- **Side Effects / Blast Radius (required)** — names the affected subsystems (`getWorktreeBase()` and its downstream consumers in `@archon/isolation`, workflow dispatch, artifact paths), the one potential unintended effect (users whose codebase `name` had non-alphanumerics that *weren't* SSH-shaped now also fall through to the path-derived layout), and the guardrails (existing `'invalid-no-slash'` test, new SSH-shape regression test, warn-level `worktree.invalid_codebase_name_format` log).

CodeRabbit's "Description check" in the pre-merge panel also marks the description as covering all template sections, FWIW. Happy to expand any specific point if something still reads thin.

Wirasm:
## Review Summary

Tight, well-scoped fix — the SSH-URL → docker-compose volume failure chain is the right insight, and routing `codebaseName` through the existing `parseOwnerRepo` validator is the smallest possible change that closes it. No blockers. A few small wins are available before merge (mainly in tests and comment density).

### Blocking issues
None.

### Suggested fixes
- **`packages/git/src/git.test.ts:307-335`** — the new regression test doesn't assert that `worktree.invalid_codebase_name_format` was emitted. The documented behavior is "reject + warn + fall through," but the test only verifies the fall-through. Add after the existing assertions:
  ```ts
  mockLogger.warn.mockClear(); // matches the precedent at lines 432/511/542/1207
  // … call resolveOwnerRepo …
  expect(mockLogger.warn).toHaveBeenCalledWith(
    { codebaseName: 'git@git.example.net:acme/widget-app' },
    'worktree.invalid_codebase_name_format'
  );
  ```
  This locks down the observability seam operators rely on to spot misconfigured codebases.

- **`packages/git/src/git.test.ts:50-59`** — your new `parseOwnerRepo` mirror is held together by a comment, not an import. If `SAFE_NAME` ever tightens in `@archon/paths`, the mirrored copy here silently keeps testing the old behavior. Easy fix: keep `mock.module` for the path functions but expose the real validator from the factory:
  ```ts
  import { parseOwnerRepo as realParseOwnerRepo } from '@archon/paths';
  mock.module('@archon/paths', () => ({
    // … existing path mocks …
    parseOwnerRepo: realParseOwnerRepo,
  }));
  ```
  Then drop the "must stay aligned with SAFE_NAME" comment — it becomes load-bearing again by construction.

- **`packages/git/src/worktree.ts:62-71`** — the 8-line block comment restates the `SAFE_NAME` regex and the fallback mechanic, both of which already live on `parseOwnerRepo`'s docstring. Compress to the docker-compose WHY:
  ```ts
  // Reject names containing ':' or '@' — they would become path segments
  // and break docker-compose short-form volume specs (HOST:CONTAINER:OPT).
  ```

### Minor / nice-to-have
- **`packages/git/src/git.test.ts:316-322`** — same comment-density issue on the regression-test block. Keep the SSH-URL → docker-compose failure chain; drop the trailing restatement of `parseOwnerRepo`'s contract (the test name and assertions already say it).

- **`packages/git/src/worktree.ts:57-66`** — heads-up: any codebase whose registered name contains `+`, `~`, whitespace, or other chars outside `[A-Za-z0-9._-]` will now route through the path-derived layout instead of the `codebaseName` branch. The warn-log makes it observable; the follow-up registration-time validator mentioned at the end of #1582 is the right place to surface this more loudly.

- **`packages/git/src/git.test.ts`** — optional: add an explicit `expect(resolved).toEqual({ owner: 'acme', repo: 'widget-app' })` to one of the existing positive-path tests. Defensive against a future inversion of the validator.

### Compliments
- Choosing to reuse the existing `'invalid-no-slash'` warn-log (`worktree.invalid_codebase_name_format`) instead of inventing a new event name keeps operator dashboards and grep history coherent. That's the kind of consistency that pays off a year later.
- The PR description's rollback-and-mitigation rationale — strict enforcement belongs at codebase *registration* time, not the resolver — is the correct layering. The follow-up in #1582 is the right place for the louder failure mode.

---
*Reviewed via maintainer-review-pr workflow (Pi/Minimax). Aspects run: code-review, error-handling, test-coverage, comment-quality. docs-impact skipped — no public API or doc surface touched.*

PR #1169, opened 2026-04-13
security: prevent target repo .env from leaking into subprocesses (#1135)

Wirasm:
## Summary

- Problem: Target repo `.env` files leaked into both Claude Code and Bun script subprocesses. Claude Code runs via Bun, which auto-loads `.env` from CWD. This caused (1) repo secrets leaking into AI subprocesses, and (2) repo API keys silently overriding Archon's auth (breaking global auth).
- What changed: Added `--no-env-file` to both Claude Code subprocess spawning (`executableArgs`) and Bun script node execution. Removed the entire scanning/consent model which was the wrong fix for this problem.
- What did **not** change: `stripCwdEnv()` boot protection, managed env injection (config.yaml env: + DB vars), `{ ...process.env }` subprocess env construction

## E2E Evidence

Tested from within a Claude Code session against a repo with contaminated `.env`:

**Claude prompt node — before fix:**
```
LEAKED_PROMPT_TEST: SHOULD_NOT_SEE_THIS
SECRET_SAUCE: repo-only
+ "Invalid API key" error (repo's fake key overrode global auth)
```

**Claude prompt node — after fix (`executableArgs: ['--no-env-file']`):**
```
LEAKED_PROMPT_TEST: not set
SECRET_SAUCE: not set
+ global auth works correctly
```

**Bun script node — before fix:**
```
LEAKED: LEAKED_FROM_SCRIPT_CWD = yes_this_leaked
```

**Bun script node — after fix (`bun --no-env-file -e`):**
```
CLEAN: LEAKED_FROM_SCRIPT_CWD not set
```

**Bash node** — always clean (bash doesn't auto-load `.env`).

**Late registration via curl** — registered repo after server boot, ran workflow, zero leaks in both bash and bun nodes.

## Architecture

```
Archon boot:
  stripCwdEnv() ──▶ clean process.env ──▶ load ~/.archon/.env

Subprocess paths (all now protected):
  Claude provider ─▶ executableArgs: ['--no-env-file'] ─▶ Claude Code skips .env
  Bun script nodes ▶ bun --no-env-file -e <code>        ─▶ Bun skips .env
  Bash nodes ──────▶ inherited process.env               ─▶ bash doesn't auto-load

Registration:
  ✅ always succeeds (no scanning, no consent)
```

## Change Metadata

- Change type: `security`
- Primary scope: `multi`
- Closes #1135

## Changes

**Fix: Claude Code subprocess env leak (packages/providers/src/claude/provider.ts)**
[...]

PR #1596, opened 2026-05-06
fix(workflows): bun script nodes — use --env-file=<null> to actually suppress cwd .env auto-load

### LINKED ISSUE #1135: security: zero env leaks from managed repos into provider subprocesses
## Problem

Archon manages many target repos/codebases. Those repos may contain `.env` files with API keys, secrets, and credentials.

The policy for target repo env should be simple:

**Target repo/project env must never be loaded into the Archon process, and must never be forwarded to Archon provider subprocesses.**

This means:

- `~/.archon/.env` is trusted Archon-owned config and may be loaded
- the Archon repo `.env` may be loaded in dev/source mode
- target repo/project `.env` files are **never** loaded into Archon
- target repo/project `.env` values are **never** inherited by provider subprocesses
- there is **no consent / opt-in / bypass path** for target repo env

Important distinction:

- **ambient target repo env** is forbidden
- **explicit Archon-managed env** is still allowed

Allowed explicit env sources:

- `~/.archon/.env`
- Archon dev/source `.env`
- `.archon/config.yaml` `env:` values
- per-codebase env vars stored in Archon DB and injected intentionally

The current design still carries a consent model (`allow_env_keys`, `allow_target_repo_keys`, `--allow-env-keys`) and that is the wrong primitive for the intended security boundary.

## Required Invariant

At all times:

- `process.env` must not contain env vars sourced from any managed target repo
- provider subprocess env must be built from trusted Archon-owned env only
- repo registration must not create a path where target repo env becomes allowed later
- cross-repo contamination must be impossible in long-lived Archon processes

## Business Logic Summary

This issue is not primarily about implementation details. It is about the product rule Archon should enforce.

### Plain-language rule

When a user points Archon at a target repo:

- Archon may inspect the repo
- Archon may run tools in the repo
- Archon may use Archon-managed credentials and config
- Archon must **not** silently inherit that repo's own `.env` secrets

In business terms:

- a target repo's secrets belong to that repo/application, not to Archon
- Archon should never bill the wrong API account because it picked up a repo's `.env`
- Archon should never “sometimes allow it if the user clicks consent”
- the behavior should be deterministic and easy to explain:
  - if the repo has sensitive keys in auto-loaded `.env` files, Archon refuses to use that repo until the user removes or relocates those keys

### Current vs desired business behavior

| Surface | Current behavior | Desired behavior |
| --- | --- | --- |
| Repo registration (`/clone`, Add Project, CLI register) | Scan repo, block by default, but allow override/consent | Scan repo, block if sensitive keys are found, no override |
| Provider spawn / workflow execution | Scan again, but skip the check if consent was granted earlier | Always block if sensitive keys are present in auto-loaded repo `.env` files |
| Archon process startup in repo cwd | Boot cleanup strips leaked repo env after Bun auto-load | Keep this protection; Archon process must stay clean |
| Provider subprocess env | Nested-session/debugger markers are sanitized, but repo-env policy still relies on separate gating | Keep sanitization, but never allow target repo `.env` to become acceptable |
| Product policy shown to users | “You can acknowledge risk and allow env keys” | “Target repo `.env` secrets are not allowed; remove or relocate them” |

### Stakeholder-level decision

This issue should resolve one product question clearly:

**Should Archon ever allow a target repo's ambient `.env` secrets to be used by Archon or its provider subprocesses?**

The answer for this issue is:

**No. Never.**

That means the system should stop modeling this as a consent workflow and instead model it as a hard safety boundary.

### 1. Remove the consent model

These primitives should be removed rather than hardened:

- codebase-level `allow_env_keys`
- config-level `allow_target_repo_keys`
- CLI `--allow-env-keys`
- UI/API flows that grant env-key consent
- remediation copy that tells users to allow or bypass repo env leakage

Registration-time scanning can remain, but only as:

- detection
- refusal / warning / audit

Not as a consent gate.

### 2. Treat target repo env as untrusted input, never trusted config

Archon should only trust:

- `~/.archon/.env`
- Archon dev/source env
- explicit Archon config values (`.archon/config.yaml env:` and DB env vars)

It should never treat target repo `.env` as valid runtime input for either:

- the Archon process
- Claude/Codex/future provider subprocesses

### 3. Harden subprocess env construction

Provider subprocess env must not be `{ ...process.env }` unless `process.env` is already proven clean and the allowed key set is explicit.

The safer direction is:

- start from trusted Archon-owned env
- sanitize nested-session/debugger markers
- explicitly merge only Archon-controlled additions
- never merge target repo env

### 4. Cover late registration and long-lived process cases

Boot-time stripping is not sufficient by itself.

Audit and harden all cases where the process outlives repo registration or repo switching:

- late-registered codebases
- `/clone`
- Web UI project registration
- CLI invocation from target-repo cwd
- long-lived server process handling multiple repos
- provider subprocesses spawned after multiple repo transitions

## Audit Findings

The current codebase still encodes a consent/bypass model in multiple places:

- registration path:
  - `packages/core/src/handlers/clone.ts`
  - supports per-call bypass via `allowEnvKeys`
  - supports global bypass via merged `allowTargetRepoKeys`
- scanner/remediation copy:
  - `packages/core/src/utils/env-leak-scanner.ts`
  - instructs users to re-run with `--allow-env-keys`, toggle UI consent, or set global bypass
- config model:
  - `packages/core/src/config/config-types.ts`
  - global and repo-level `allow_target_repo_keys`
- config loading:
  - `packages/core/src/config/config-loader.ts`
  - merges global and repo-level bypass into effective `allowTargetRepoKeys`
- DB model:
  - `remote_agent_codebases.allow_env_keys`
  - represented in code at `packages/core/src/db/codebases.ts`
- API/UI surface:
  - `packages/server/src/routes/api.ts`
  - `PATCH /api/codebases/{id}` is explicitly a consent-flag route
  - `packages/server/src/routes/schemas/codebase.schemas.ts` exposes `allow_env_keys`
- CLI surface:
  - `packages/cli/src/cli.ts`
  - exposes `--allow-env-keys`
- startup behavior:
  - `packages/server/src/index.ts`
  - startup scan is currently skipped when global bypass is enabled
- provider subprocess env construction:
  - `packages/providers/src/claude/provider.ts`
  - `buildSubprocessEnv()` still returns `{ ...process.env }`

This issue should remove the consent model end-to-end, not just harden one call site.

## Acceptance Criteria

- [ ] No target repo/project `.env` keys are ever present in Archon `process.env`
- [ ] No target repo/project `.env` keys are ever present in provider subprocess env
- [ ] Cross-repo contamination is impossible: subprocess for repo A has zero repo-env keys from repo B
- [ ] Late-registered repos are as safe as boot-time repos
- [ ] Consent/bypass primitives for target repo env are removed
- [ ] `allow_env_keys` is removed from DB/API/types
- [ ] `allow_target_repo_keys` is removed from config/types/loading
- [ ] `--allow-env-keys` is removed from the CLI
- [ ] Registration and scan flows no longer instruct users to allow or bypass repo env leakage
- [ ] Env protection is enforced by platform/bootstrap/runtime boundaries, not by individual providers
- [ ] Audit covers all provider call sites and subprocess env construction paths
- [ ] Explicit Archon-managed env injection still works:
  - `.archon/config.yaml env:`
  - DB-managed per-codebase env vars

## Non-Goals

- Allowing repo env with user consent
- Per-codebase env exemptions
- Global YAML bypass for repo env leakage

## Related

- #1150 — nested Claude marker leakage shows subprocess env needs its own hardening
- #991 — previous `sendQuery()` env-leak gate bug
- Provider extraction / Phase 1: caller-side safety ownership remains correct, but the target policy must be “never allow,” not “allow with consent”

nabilragab:
## Summary

- **Problem**: The `script:` nodes (`runtime: bun`) tried to suppress Bun's cwd `.env` auto-load with `--no-env-file`, but that flag only disables explicit `--env-file=…` args — Bun still auto-loads `.env` from the script's execution directory regardless. The previously-tolerated regression-test (`bun script node does not leak repo .env from execution cwd`) under #1135 has been failing for that reason.
- **Why it matters**: Defense-in-depth security boundary. The user-facing protection comes from the registration-time gate landed in #1169 (refuses repos whose auto-loaded `.env` contains sensitive keys), but the in-process subprocess layer was silently a no-op.
- **What changed**: Switched both `script:` node call sites to `--env-file=/dev/null` (POSIX) / `--env-file=NUL` (Windows). Verified empirically against bun 1.3.x. Two test assertions updated to match. CHANGELOG entry under `[Unreleased] → Fixed`.
- **What did NOT change (scope boundary)**: Bash nodes (no Bun involvement), the registration-time gate from #1169, parent-process `stripCwdEnv()` boot protection, claude/codex provider env construction, no schema/migration/config changes.

### Before

\`\`\`
User                   Archon                          script: subprocess
────                   ──────                          ──────────────────
runs workflow ──────▶  spawns:
                       bun --no-env-file -e <code>    bun starts in cwd
                       (intent: no .env auto-load)    auto-loads ./.env  ← LEAKS
                                                       process.env has
                                                       repo secrets
sees broken            (defense-in-depth gate
test                    silently bypassed)
\`\`\`

### After

\`\`\`
User                   Archon                              script: subprocess
────                   ──────                              ──────────────────
runs workflow ──────▶  spawns:
                       bun *--env-file=/dev/null* -e <c>  bun starts in cwd
                                                          *skips* ./.env auto-load
                                                          process.env clean
                                                          (only Archon-managed
                                                           envVars present)
\`\`\`

### Before

\`\`\`
packages/workflows/src/dag-executor.ts
       │
       │ spawnScript()
       ▼
   execFileAsync('bun', ['--no-env-file', '-e', code], {cwd, env})
                          │
                          └─ no-op for cwd auto-load (Bun bug/feature)
\`\`\`

### After

\`\`\`
packages/workflows/src/dag-executor.ts
       │ + BUN_NULL_ENV_FILE constant (POSIX: /dev/null, Win32: NUL)
       │
       │ spawnScript()
       ▼
   execFileAsync('bun', ['--env-file=' + BUN_NULL_ENV_FILE, '-e', code], {cwd, env})
                          │
                          └─ Bun reads empty file → skips cwd auto-load
\`\`\`

**Connection inventory**:

| dag-executor.ts | bun subprocess | **modified** | flag changed; same call site |
| dag-executor.test.ts | spawn assertion | **modified** | matches new flag form |
| script-node-deps.test.ts | spawn assertion | **modified** | matches new flag form |
| dag-executor.ts | bash subprocess | unchanged | bash nodes never used the flag |

## Label Snapshot

- Risk: \`risk: low\`
- Size: \`size: XS\`
- Scope: \`workflows\`
- Module: \`workflows:dag-executor\`

## Change Metadata

- Change type: \`bug\` (defense-in-depth security boundary that was silently broken)
- Primary scope: \`workflows\`

## Linked Issue

- Closes #1135 (the test regression layer; the user-facing gate was already closed by #1169)

\`\`\`bash
$ bun run validate
# all packages: 0 fail across the suite
# (specifically: dag-executor.test.ts 206/206 pass; script-node-deps.test.ts 0 fail)
\`\`\`

- Evidence: full \`bun run validate\` exited 0 with no failures across all 14 test invocations
- Empirical verification of the flag form:
  \`\`\`
  with --no-env-file:        \"should_not_appear\"   ← still leaks
  with --env-file=/dev/null: \"CLEAN\"              ← actually works
  with BUN_DISABLE_DOTENV=1: \"should_not_appear\"   ← also doesn't help
  \`\`\`

## Security Impact (required)

- Secrets/tokens handling changed? **No** — this strengthens an existing boundary
- Risk: this is a security strengthening, not a relaxation

## Compatibility / Migration

- Backward compatible? **Yes** — pure flag swap, no schema/config/API changes
- Config/env changes? **No**

- Verified scenarios:
  - Empirical leak reproduction with both flag forms (manual `bun -e` test)
  - Full \`bun run validate\` end-to-end
  - Three deterministic re-runs of dag-executor.test.ts (206/206 each time)
- Edge cases checked:
  - Inline script path (\`bun -e <code>\`) — line 1504 in dag-executor.ts
  - Named script path (\`bun run <path>\`) — line 1591 in dag-executor.ts
  - Both call sites use the same constant
- What was not verified:
  - Windows execution (no Windows machine handy) — used \`process.platform === 'win32' ? 'NUL' : '/dev/null'\` per Bun's documented Windows null-device support; unit-test assertion accommodates both forms

- Affected subsystems/workflows: only \`script:\` DAG nodes with \`runtime: bun\` — bash nodes, prompt nodes, command nodes, claude_code nodes are untouched
- Potential unintended effects: if a user was *relying* on the leak (intentionally putting secrets in a target repo \`.env\` for their script to pick up), their workflow will now see those secrets missing. Per Archon's security model (#1135, #1169) this was never supported — they should use \`.archon/config.yaml env:\` or per-codebase env vars instead
- Guardrails/monitoring for early detection: the regression test now passes; CI catches any future flag regressions

## Rollback Plan (required)

- Fast rollback: revert this single commit; the previous (broken-but-tolerated) state returns
- Feature flags or config toggles: none required
- Observable failure symptoms: \`script:\` node fails to find an env var its inline code expects, where that env var was *previously* leaked from the cwd \`.env\`. (Indicates user was relying on the leak — not a bug to roll back, but worth a note in support.)

## Risks and Mitigations

- **Risk**: A user's existing script silently relied on the leak.
  - **Mitigation**: Migration path is documented in the existing \`.archon/config.yaml env:\` and per-codebase env-vars surfaces. The CHANGELOG entry frames the change as a security fix and points at the supported paths.
- **Risk**: Future Bun version changes the meaning of \`--env-file=/dev/null\`.
  - **Mitigation**: The two test assertions catch any flag-shape regression on every CI run; the empirical test embedded in the existing regression test (\`bun script node does not leak repo .env\`) catches behavior changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unintended .env auto-loading when running scripts with Bun, ensuring scripts no longer inherit workspace cwd env variables.

* **Changed**
  * Bun script invocation now uses an explicit, cross-platform null env-file argument (POSIX and Windows) for more reliable environment isolation and test consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Docker: `git config --global --add safe.directory` in the entrypoint now de-duplicates entries before adding, preventing unbounded growth of `~/.gitconfig` now that `/home/appuser` is persisted (#1518).
 - Docker: `setup-auth` now warns at startup when `CODEX_*` env vars are absent but a persisted `~/.codex/auth.json` from a previous run still exists, so operators don't accidentally use stale or revoked credentials (#1518).
+- `script:` nodes (`runtime: bun`) now actually suppress Bun's cwd `.env` auto-load. The previous `--no-env-file` flag turns out to only disable explicit `--env-file=…` args — Bun still auto-loads `.env` from the script's execution directory regardless. Switched to `--env-file=/dev/null` (POSIX) / `--env-file=NUL` (Windows), which is the only argument form that actually short-circuits the auto-load. Verified empirically against bun 1.3.x. Resolves the previously-tolerated `bun script node does not leak repo .env from execution cwd` test under #1135 — the registration-time gate from #1169 was the primary protection; this is the defense-in-depth layer that is now also honest.
 
 ## [0.3.10] - 2026-04-29
 
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -6511,9 +6511,15 @@ describe('executeDagWorkflow -- script nodes', () => {
       { ...minimalConfig, envVars: { MY_SECRET: 'abc123' } }
     );
 
+    // The first arg suppresses Bun's cwd .env auto-load. `--no-env-file`
+    // doesn't actually do that — it only disables explicit `--env-file=…`
+    // args. The portable null-env-file value is `/dev/null` on POSIX and
+    // `NUL` on Windows; we accept either so the test runs on both.
+    const expectedNullEnvArg =
+      process.platform === 'win32' ? '--env-file=NUL' : '--env-file=/dev/null';
     expect(execSpy).toHaveBeenCalledWith(
       'bun',
-      ['--no-env-file', '-e', 'console.log("ok")'],
+      [expectedNullEnvArg, '-e', 'console.log("ok")'],
       expect.objectContaining({
         env: expect.objectContaining({ MY_SECRET: 'abc123' }),
       })
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -37,6 +37,7 @@ import {
   checkTriggerRule,
   substituteNodeOutputRefs,
   executeDagWorkflow,
+  BUN_NULL_ENV_FILE,
 } from './dag-executor';
 import { loadMcpConfig } from '@archon/providers/claude/provider';
 import type { DagNode, BashNode, ScriptNode, NodeOutput, WorkflowRun } from './schemas';
@@ -6514,12 +6515,10 @@ describe('executeDagWorkflow -- script nodes', () => {
     // The first arg suppresses Bun's cwd .env auto-load. `--no-env-file`
     // doesn't actually do that — it only disables explicit `--env-file=…`
     // args. The portable null-env-file value is `/dev/null` on POSIX and
-    // `NUL` on Windows; we accept either so the test runs on both.
-    const expectedNullEnvArg =
-      process.platform === 'win32' ? '--env-file=NUL' : '--env-file=/dev/null';
+    // `NUL` on Windows; the imported constant resolves to the right one.
     expect(execSpy).toHaveBeenCalledWith(
       'bun',
-      [expectedNullEnvArg, '-e', 'console.log("ok")'],
+      [`--env-file=${BUN_NULL_ENV_FILE}`, '-e', 'console.log("ok")'],
       expect.objectContaining({
         env: expect.objectContaining({ MY_SECRET: 'abc123' }),
       })
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -1257,6 +1257,16 @@ async function executeNodeInternal(
 const SUBPROCESS_DEFAULT_TIMEOUT = 120_000;
 
 /**
+ * Empty file Bun is pointed at via `--env-file=` to suppress its
+ * cwd `.env` auto-load for `script:` nodes. `--no-env-file` only
+ * disables explicit `--env-file=…` args; the cwd auto-load survives
+ * unless we point Bun at a real (empty) file. `/dev/null` on POSIX,
+ * `NUL` on Windows. (Verified empirically against bun 1.3.x — see
+ * the regression test in `dag-executor.test.ts` for #1135.)
+ */
+const BUN_NULL_ENV_FILE = process.platform === 'win32' ? 'NUL' : '/dev/null';
+
+/**
  * Execute a bash (shell script) DAG node.
  * Runs the script via `bash -c`, captures stdout as node output.
  * No AI session is created — bash nodes are free/deterministic.
@@ -1498,10 +1508,13 @@ async function executeScriptNode(
       // Inline code execution
       if (node.runtime === 'bun') {
         cmd = 'bun';
-        // --no-env-file prevents Bun from auto-loading .env from the execution
-        // cwd (the target repo). Without this, repo .env leaks into the script
-        // subprocess despite Archon's parent process cleanup.
-        args = ['--no-env-file', '-e', finalScript];
+        // Suppress Bun's auto-load of `.env` from the execution cwd
+        // (the target repo). `--no-env-file` ONLY disables explicit
+        // `--env-file=…` args — it does NOT stop the cwd auto-load.
+        // The portable way to actually suppress the auto-load is to
+        // point Bun at an empty file: `/dev/null` on POSIX, `NUL` on
+        // Windows. Verified empirically against bun 1.3.x.
+        args = [`--env-file=${BUN_NULL_ENV_FILE}`, '-e', finalScript];
       } else {
         // uv run --with dep1 --with dep2 python -c <code>
         cmd = 'uv';
@@ -1509,7 +1522,8 @@ async function executeScriptNode(
         args = ['run', ...withFlags, 'python', '-c', finalScript];
       }
     } else {
-      // Named script — look up across repo and home scopes.
+      // Named script — look up across repo and home scopes. Same
+      // `--env-file=<null>` rationale as inline scripts above.
       // Precedence: <cwd>/.archon/scripts/ > ~/.archon/scripts/ (repo wins).
       // Wrap discovery in its own try/catch so a permission error on ~/.archon/scripts/
       // isn't mis-attributed by the outer catch's "permission denied (check cwd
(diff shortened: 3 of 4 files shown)


nabilragab:
Reviewed the two findings:

**🧹 Nit (accepted)** — exported `BUN_NULL_ENV_FILE` from `dag-executor.ts` and replaced the duplicated platform ternary in both test files with the imported constant. Pushed as `7303d06f`.

**🔴 "Critical" (rejected)** — the suggestion to swap `--env-file=${BUN_NULL_ENV_FILE}` back to `--no-env-file` would re-introduce the exact bug this PR fixes. The Bun docs CodeRabbit cited describe `--no-env-file`'s _intent_, not its actual behavior on Bun 1.3.x. Empirically verified before opening this PR (also documented in the CHANGELOG and PR body):

```
flag                          | observed cwd .env auto-load
------------------------------+----------------------------
--no-env-file                 | STILL LEAKS (no-op for auto-load)
--env-file=/dev/null          | suppressed (this PR)
BUN_DISABLE_DOTENV=1          | STILL LEAKS
```

Re-confirmed end-to-end this morning via the regression workflow `env-leak-probe.yaml` against this branch: with the fix → output `CLEAN`; reverted to `--no-env-file` → output `should_not_appear` (leaked). The test in `dag-executor.test.ts` (`bun script node does not leak repo .env from execution cwd`) is the locked-in regression guard, and it would fail again under the suggested change.

Tracking upstream as a separate item; if Bun later changes `--no-env-file` to also suppress cwd auto-load, we can revisit. Until then the null-device pattern is the only flag form that actually achieves the goal.

PR #1636, opened 2026-05-11
fix(workflows): persist structuredOutput on NodeOutput so $node.output.field works for Pi

### LINKED ISSUE #1571: Parsed structured output not persisted to NodeOutput — condition evaluator and $node.output.field re-parse raw text, breaking Pi/Minimax
## Summary

- **What broke**: When a DAG node declares `output_format`, the engine extracts the structured object during streaming but does not persist it on `NodeOutput`. Downstream consumers (`when:` conditions, `$nodeId.output.field` substitution) re-parse the raw assistant text via `JSON.parse(...)`. This works for Claude and Codex (their SDKs enforce a clean JSON shape via `output_format`), but breaks for Pi/Minimax (best-effort structured output via prompt augmentation — model output routinely includes ```` ```json ```` fences and reasoning preamble that `JSON.parse` rejects).
- **Severity**: `major` — Pi-using workflows silently skip downstream nodes whose `when:` conditions reference structured fields. Today's run of `maintainer-review-pr` on PR #1552 skipped every deep-review aspect for this reason.
- **Related**: Surfaced during issue #1569 triage (Pi/Minimax SDK errors). This is a separate root cause — same provider, different code path.

## What's actually happening

Pi extracts structured output correctly via `tryParseStructuredOutput` (`packages/providers/src/community/pi/event-bridge.ts`) — fence-stripping and forward-scan-from-first-`{` are already implemented and tested. The result reaches `dag-executor.ts:887`:

```ts
if (msg.structuredOutput !== undefined) structuredOutput = msg.structuredOutput;
```

But `structuredOutput` is captured into a local variable, never written into `NodeOutput`. Consumers see only `nodeOutput.output` (the raw assistant text) and call `JSON.parse(...)` on it.

For Claude/Codex, the raw text *is* the structured JSON (the SDK enforces it). For Pi, the raw text is whatever the model chose to emit — typically:

| Pi-emitted text shape | `JSON.parse(rawText)` works? |
|---|---|
| `{"verdict":"review","..."}` | ✓ sometimes |
| `` ```json\n{"verdict":"review",...}\n``` `` | ✗ |
| `Now I have all the info...\n{"verdict":"review",...}` | ✗ (Minimax M2.7 reasoning pattern) |

Result: every `when: $gate.output.verdict == 'review'` against a Pi gate node fails to parse and returns empty string, silently skipping the node.

## Two affected call sites

```
packages/workflows/src/condition-evaluator.ts:47
  const parsed = JSON.parse(nodeOutput.output) as Record<string, unknown>;

packages/workflows/src/dag-executor.ts:303
  const parsed = JSON.parse(nodeOutput.output) as Record<string, unknown>;
```

Both fail identically on Pi output. The `condition_json_parse_failed` log fires for the first; the second silently substitutes empty string into prompts/bash.

## Concrete repro from today

`archon workflow run maintainer-review-pr "review PR #1552"` on `provider: pi, model: minimax/MiniMax-M2.7`:

```json
{
  "level": 40,
  "module": "workflow.condition-evaluator",
  "nodeId": "gate",
  "field": "verdict",
  "outputPreview": "\n\n```json\n{\"verdict\":\"review\",\"direction_alignment\":\"aligned\",\"scope_assessment\":\"focused\",\"template",
  "msg": "condition_json_parse_failed"
}
```

Downstream `review-classify`, `code-review`, `error-handling`, `test-coverage`, `comment-quality`, `docs-impact`, and `synthesize` all skipped via `when_condition` / `trigger_rule` because they couldn't read `$gate.output.verdict`. Net effect: gate ran but no deep review. Today's batch of 11 maintainer-reviews had at least one full skip from this (#1552); other partial skips can't be ruled out without re-checking each output preview.

## Why fence-stripping at the parse sites is the wrong fix

It's tempting to make `JSON.parse(...)` at the two sites robust to fences/preamble. Don't:

1. **The work is already done.** `tryParseStructuredOutput` already handles all three shapes correctly with full test coverage (see `event-bridge.test.ts` cases for `strips ```json``` fences`, `parses preamble + trailing JSON`, `parses preamble + trailing nested JSON via forward scan`).
2. **Only Pi produces fence/preamble outputs.** Claude SDK and Codex SDK enforce JSON shape at the API level via `output_format`. Adding fence-stripping in the workflow engine solves a problem only one provider has — and we'd be re-parsing what's already parsed.
3. **It still re-parses on every read.** `condition-evaluator` and `substitute-refs` may be called multiple times for the same field (multiple downstream `when:` conditions, repeated `$nodeId.output.field` references in a prompt). Each read re-parses. That's avoidable.

## The fix

Persist the already-parsed object onto `NodeOutput` and prefer it at both call sites.

### 1. `packages/workflows/src/schemas/workflow-run.ts:66`

```ts
export const nodeOutputSchema = z.discriminatedUnion('state', [
  z.object({
    state: z.enum(['completed', 'running']),
    output: z.string(),
    sessionId: z.string().optional(),
    structuredOutput: z.unknown().optional(),  // NEW
  }),
  z.object({
    state: z.literal('failed'),
    output: z.string(),
    sessionId: z.string().optional(),
    error: z.string(),
    structuredOutput: z.unknown().optional(),  // NEW
  }),
  z.object({
    state: z.enum(['pending', 'skipped']),
    output: z.string(),
  }),
]);
```

### 2. `packages/workflows/src/dag-executor.ts:1198-1203`

When constructing the success-path return value:
```ts
return {
  state: 'completed',
  output: nodeOutputText,
  sessionId: newSessionId,
  costUsd: nodeCostUsd,
  ...(structuredOutput !== undefined ? { structuredOutput } : {}),  // NEW
};
```

`structuredOutput` is already in scope (captured at line 887). Same shape change for the loop-node path (~line 1934 area, where structured output flows from the loop's terminal iteration).

### 3. `packages/workflows/src/condition-evaluator.ts:46-58`

Prefer the persisted object:
```ts
if (!field) return nodeOutput.output;

// Use the already-parsed structured output when available — avoids
// re-parsing raw text that may be fence-wrapped or preamble-prefixed
// when the source provider is Pi (Claude/Codex SDKs enforce clean JSON).
if (nodeOutput.structuredOutput !== undefined) {
  const value = (nodeOutput.structuredOutput as Record<string, unknown>)[field];
  // ... existing string/number/bool/object handling
}

// Fallback for nodes without output_format declared:
try {
  const parsed = JSON.parse(nodeOutput.output) as Record<string, unknown>;
  // ... existing logic
} catch {
  ...
}
```

### 4. `packages/workflows/src/dag-executor.ts:286-310` (`substituteNodeOutputRefs`)

Same preference order — use `structuredOutput` if present, fall back to `JSON.parse(text)`.

## What this does NOT change

- Pi's `event-bridge.ts:tryParseStructuredOutput` stays Pi-specific. Correct — it's solving Pi's "model talks before emitting JSON" problem, which Claude/Codex don't have.
- Claude and Codex behavior — they still get clean JSON from their SDKs and `JSON.parse` still works as the fallback. No regression.
- The `output` text field stays as-is — preserves the literal assistant text for cases where a downstream node wants to quote/forward it.

## Migration concern

Existing `node_outputs` rows in the DB don't have `structuredOutput`. The schema additions are optional, so reads remain backward-compatible — old rows just take the `JSON.parse(output)` fallback path, same as today. No migration needed.

## Tests to add

In `condition-evaluator.test.ts`:
- Returns field value from `structuredOutput` when present (Pi-shaped raw text with fences must not be needed for parse).
- Falls back to `JSON.parse(output)` when `structuredOutput` is absent (Claude/Codex compatibility regression).
- Returns empty string and warns when both paths fail.

In `dag-executor.test.ts`:
- `substituteNodeOutputRefs` uses `structuredOutput` preferentially.
- Backward compat: nodes without `structuredOutput` still parse from text.

## Environment

- Platform: any (Pi-using workflows)
- Provider: pi / minimax/MiniMax-M2.7 (any Pi-backed model)
- Affects: condition evaluation in `when:` clauses, `$nodeId.output.field` references in prompts/bash.

## Impact

- Affected workflows/commands: any DAG that uses `output_format` on a Pi node and then references `$nodeId.output.field` downstream — this includes `maintainer-review-pr`, `archon-fix-github-issue`, `archon-fix-github-issue-experimental`, and any user-authored Pi workflow with a classifier/gate pattern.
- Reproduction rate: Always when Pi emits fence-wrapped or preamble-prefixed JSON — happens routinely with Minimax M2.7 reasoning models.
- Workaround: None client-side. Users must run such workflows on Claude/Codex.
- Data loss risk: No (workflow state stays consistent). Lost work is silent — downstream nodes appear to skip "correctly" via `when:` conditions.

## Scope

- Package(s) involved: `workflows`
- Modules: `workflows:condition-evaluator`, `workflows:dag-executor` (substituteNodeOutputRefs + node-output construction sites), `workflows:schemas/workflow-run` (nodeOutputSchema)

truffle-dev:
## Summary

- Problem: When a provider parses fence-wrapped or preamble-prefixed JSON onto the result chunk (Pi/Minimax via `tryParseStructuredOutput`), the DAG executor captures it into a local variable but never persists it onto `NodeOutput`. Downstream `$node.output.field` substitution and `when:` conditions then try `JSON.parse(output)` on the original prose-prefixed text, throw, and resolve to empty.
- Why it matters: every Pi/Minimax workflow with a downstream node that reads `$classifier.output.type` (or any other field) silently routes to the wrong branch or drops the value.
- What changed: `structuredOutput?: unknown` added to the completed/running and failed NodeOutput branches; producer call sites persist it; consumers (`substituteNodeOutputRefs`, `condition-evaluator.resolveOutputRef`) prefer it over `JSON.parse(output)` and fall back to the existing path when absent.
- What did **not** change: bare `$node.output` (unfielded) still reads `output` text. Cross-resume rehydration from `event_data` is out of scope (resumed runs that re-execute downstream nodes fall through to JSON.parse). Providers that already overwrite `output` with a JSON.stringify of structuredOutput (Claude/Codex with `output_format`) are unaffected because their `output` is already JSON.

### Before

```
Pi/Minimax DAG with downstream $.field reads

  classify (provider=pi, prompt asks for JSON)
    ├─ raw chunk:        "Here is the classification:\n{\"type\":\"BUG\"}"
    ├─ tryParseStructuredOutput → { type: "BUG" }
    ├─ stored:           NodeOutput.output = raw text, structuredOutput = DROPPED
    │
  triage (depends_on: classify, when: $classify.output.type == 'BUG')
    └─ resolveOutputRef:
         JSON.parse("Here is the classification:\n{...}")  ── throws
         returns ''
         condition: '' == 'BUG' → false → SKIPPED  ❌
```

### After

```
Pi/Minimax DAG with downstream $.field reads

  classify (provider=pi, prompt asks for JSON)
    ├─ raw chunk:        "Here is the classification:\n{\"type\":\"BUG\"}"
    ├─ tryParseStructuredOutput → { type: "BUG" }
    ├─ stored:           NodeOutput.output = raw text,
                         [NodeOutput.structuredOutput = { type: "BUG" }]
    │
  triage (depends_on: classify, when: $classify.output.type == 'BUG')
    └─ resolveOutputRef:
         [prefers NodeOutput.structuredOutput.type]  → "BUG"
         condition: 'BUG' == 'BUG' → true → RUNS  ✓
```

### Before

```
   ┌───────────────────────────┐
   │  providers (pi/minimax)   │
   │  emits result chunk with  │
   │  structuredOutput field   │
   └───────────┬───────────────┘
               │
               ▼
   ┌───────────────────────────┐
   │     dag-executor.ts       │
   │   captures msg.struct-    │
   │   uredOutput in local     │
   │   var (line ~913)         │
   │                           │
   │   on success return:      │
   │   { state, output,        │
   │     sessionId, costUsd }  │── structuredOutput dropped
   └───────────┬───────────────┘
               │
               ▼
   ┌───────────────────────────┐    ┌───────────────────────────┐
   │ substituteNodeOutputRefs  │    │   condition-evaluator     │
   │   JSON.parse(output)      │    │   JSON.parse(output)      │
   │   throws on prose preamble│    │   throws on prose preamble│
   └───────────────────────────┘    └───────────────────────────┘
```

### After

```
   ┌───────────────────────────┐
   │  providers (pi/minimax)   │
   │  emits result chunk with  │
   │  structuredOutput field   │
   └───────────┬───────────────┘
               │
               ▼
   ┌───────────────────────────┐
   │     dag-executor.ts       │
   │   captures msg.struct-    │
   │   uredOutput in local     │
   │   var (line ~913)         │
   │                           │
   │   on success return:      │
   │   { state, output,        │
   │     sessionId, costUsd,   │
   │  [~ structuredOutput ] }  │── persisted
   └───────────┬───────────────┘
               │
               ▼
   ┌───────────────────────────┐    ┌───────────────────────────┐
   │ substituteNodeOutputRefs  │    │   condition-evaluator     │
   │ [~ prefer structuredOut]  │    │ [~ prefer structuredOut]  │
   │   falls back to           │    │   falls back to           │
   │   JSON.parse(output)      │    │   JSON.parse(output)      │
   └───────────────────────────┘    └───────────────────────────┘
```

**Connection inventory**:

| `dag-executor` (single-shot success return) | `NodeOutput.structuredOutput` | **new** | spread when defined |
| `dag-executor` (loop terminal-iteration return) | `NodeOutput.structuredOutput` | **new** | new `lastIterationStructuredOutput` local |
| `substituteNodeOutputRefs` | `NodeOutput.structuredOutput` | **new** | preferred over JSON.parse |
| `condition-evaluator.resolveOutputRef` | `NodeOutput.structuredOutput` | **new** | preferred over JSON.parse |
| `nodeOutputSchema` (completed/running, failed) | `z.unknown().optional()` | **new** field | back-compat |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `workflows`
- Module: `workflows:executor`, `workflows:condition-evaluator`, `workflows:schemas`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #1571

```bash
bun run --filter @archon/workflows test       # all green (workflows package)
bun --filter '*' test                          # all packages green
bun run --filter @archon/workflows type-check  # 0 errors
bun run type-check                             # full repo, 0 errors
bun run format:check                           # All matched files use Prettier code style!
bun x eslint .                                 # 0 errors
bun run check:bundled                          # up to date
bun run check:bundled-skill                    # up to date
```

New tests added:
- `condition-evaluator.test.ts`: 12 new tests covering structuredOutput preference, Claude/Codex JSON.parse fallback regression, numeric/boolean/object/array field coercion, null/top-level-array/primitive fall-through, missing-field semantics, and bare `$node.output` (no field) still uses output text.
- `dag-executor.test.ts`: 13 new tests for `substituteNodeOutputRefs` covering the same matrix plus shell-escaping safety on structuredOutput fields.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? `Yes`. `structuredOutput` is optional on the schema; producers populate it only when the provider emits one; consumers fall through to the existing `JSON.parse(output)` path when it is absent. Workflows that depend on the JSON.parse path keep working unchanged.
- Config/env changes? `No`
- Database migration needed? `No`. NodeOutput is stored inline in workflow event data; older rows without `structuredOutput` deserialize as `undefined` and hit the JSON.parse fallback.
- Resume note: on a paused-then-resumed workflow run, downstream nodes that re-execute against an already-completed upstream will rehydrate `NodeOutput` from `event_data`. If the original run was on this version, the field is preserved; if older, it is absent and the JSON.parse fallback runs. Cross-version resume of Pi workflows that previously routed wrong will route correctly once the upstream re-runs on the new version.

- Verified scenarios:
  - Pi-shape: prose output + populated structuredOutput → dot-access reads structuredOutput.
  - Claude/Codex output_format-shape: empty/JSON output, no structuredOutput → JSON.parse fallback still works.
  - Loop terminal iteration: structuredOutput from the final iteration is what survives onto NodeOutput.
  - Bare `$node.output` (no field) ignores structuredOutput and reads output text.
  - Missing field on structuredOutput → empty string (does not silently retry JSON.parse).
  - null / top-level-array / primitive structuredOutput → falls through to JSON.parse (matches existing edge-case semantics).
  - Shell-escaping path (`escapedForBash=true`) wraps structuredOutput strings/objects in single quotes; numbers/booleans stay unquoted.
- Edge cases checked: empty output text combined with populated structuredOutput (Pi-only-structured case) still resolves the field correctly because the empty-output early-return was moved inward.
- What was not verified by hand: end-to-end Pi workflow against the live provider (the change is verifiable via unit tests against the consumer call sites and the producer return shape).

- Affected subsystems/workflows: any DAG workflow with a `$node.output.field` substitution or a `when:` expression that uses dot notation.
- Potential unintended effects: a workflow author who currently relies on `JSON.parse(output)` returning a *different* shape than what the provider's parsed structuredOutput contains (e.g. provider parses one JSON object out of multiple in the prose) would see a behavior change. In practice the providers that emit `structuredOutput` are emitting the canonical structured payload that the prompt asked for, so the change is the intended fix, not a surprise.
- Guardrails: existing log lines (`condition_json_parse_failed`, `dag_node_output_ref_json_parse_failed`) still fire on the fallback path; no new error paths introduced.

## Rollback Plan (required)

- Fast rollback: revert the single commit on this PR. Schema field is optional and writing it on new rows does not break readers that don't know about it.
- Feature flags: none. The change is mechanical and small enough that flagging it would add more risk than it removes.
- Observable failure symptoms (if regression slipped through): `$node.output.field` would resolve to JSON-stringified content where it previously resolved to a plain string (would surface as condition mismatches or wrongly-formatted prompts on Claude/Codex output_format workflows).

## Risks and Mitigations

- Risk: a workflow that historically saw an empty string for a `$node.output.field` on a Pi/Minimax node (because the JSON.parse failed) might now see the resolved value and take a previously-unreachable branch. This is the intended behavior change for #1571 but could surprise authors who built around the bug.
  - Mitigation: workflow authors who want the old empty-string behavior can keep relying on missing-field semantics (the new code returns `''` when the structuredOutput object doesn't contain the requested key). The fix only changes the case where the provider actually parsed the structured payload the prompt requested.


Wirasm:
this seems to have way more changes then the pr body claims, please reopen once scope is fixed

truffle-dev:
Rebased onto upstream/main. The branch was based off a downstream dev that carried 179 unmerged-to-main commits — the PR diff now matches the body: 5 files, +283/-9 across condition-evaluator, dag-executor, and the workflow-run schema. New SHA: 53571b97. `bun test src/condition-evaluator.test.ts src/dag-executor.test.ts` green (263/0).

PR #1396, opened 2026-04-24
fix(workflows): make auto-resume opt-in via --resume flag

Wirasm:
## Summary

- **Problem**: `archon workflow run <name> "<msg>"` silently resumed the last failed run on every invocation — even without `--resume`. The `findResumableRun` check in `executeWorkflow` fired unconditionally, and `options.resume` was never forwarded to the executor.
- **Why it matters**: Users who changed workflow parameters or wanted a fresh start got a cached resume instead, causing confusing behavior that contradicted the documented `--resume` opt-in.
- **What changed**: Added `allowAutoResume?: boolean` to `executeWorkflow`; gated the entire resume detection block behind it; CLI now forwards `options.resume === true`; the orchestrator's explicit foreground-resume call site passes `true` since it already located the run.
- **What did not change**: `archon workflow resume <id>`, approval/rejection auto-resume paths, and web UI resume button — all continue to work as before.

### Before

```
User                          CLI                        executeWorkflow
────                          ───                        ───────────────
archon workflow run foo "x"   ──────────────────────▶   findResumableRun() ← fires always
                                                         found prior failed run
                                                         ▶ silently resumes (wrong!)
```

### After

```
User                          CLI                        executeWorkflow
────                          ───                        ───────────────
archon workflow run foo "x"   ──────────────────────▶   allowAutoResume=false → skip resume
                                                         ▶ creates fresh run ✓

archon workflow run foo "x" --resume
                              allowAutoResume=true  ──▶  findResumableRun() → resumes ✓
```

### Before

```
CLI (workflow.ts)
  options.resume → [used only for worktree path selection]
  executeWorkflow(…)  ← allowAutoResume NOT forwarded

executeWorkflow
  findResumableRun() ← fires unconditionally on EVERY call
```

### After

```
CLI (workflow.ts)
  options.resume → allowAutoResume: options.resume === true
  executeWorkflow(…, allowAutoResume)

executeWorkflow
  if (allowAutoResume) {
    findResumableRun() ← only fires when caller opts in
  }

orchestrator-agent.ts (explicit foreground-resume path)
  executeWorkflow(…, allowAutoResume: true) ← already found the run
```

**Connection inventory:**

| From | To | Status | Notes |
|------|----|--------|-------|
| `workflow.ts` | `executeWorkflow` | **modified** | now forwards `allowAutoResume` |
| `orchestrator-agent.ts` foreground-resume | `executeWorkflow` | **modified** | passes `true` explicitly |
[...]

PR #1282, opened 2026-04-17
Fix Issue #972 | Workflow Composition

Dev-Force:
## Summary

- **Problem:** No way to express "run these nodes until tests pass" — DAG topology is fixed at load time; the only iteration primitive (`loop:` node) iterates a single prompt, not a multi-step graph. `when:` conditions used substring matching rather than real expression evaluation.
- **Why it matters:** Multi-step test→fix→retest cycles are a core agentic pattern and couldn't be expressed without duplicating nodes (fragile, capped at compile time).
- **What changed:** (1) Full expression evaluator replacing substring `when:` matching; (2) workflow-level `loop_until` + `max_iterations` that re-runs the full DAG until a condition is met; (3) `workflow:` node type that invokes a named child workflow as a single DAG node with its output flowing downstream; (4) two bundled default workflows demonstrating composition.
- **What did NOT change:** Node-level `loop:` primitive (PR #785) is untouched and orthogonal. DAG cycle detection, executor control flow, `IWorkflowStore`, platform adapters, and database schema are all unchanged.

### Before

```
User                          Archon CLI / Chat
────                          ─────────────────
runs workflow ──────────────▶ loads DAG once
                              executes each node in topological order
                              when: "$node.output == 'pass'" ──▶ substring match
                                                                   (always false unless
                                                                   literal string present)
sees final output ◀────────── single pass, no retry on failure
                              (must duplicate nodes to fake iteration)
```

### After

```
User                          Archon CLI / Chat             Child Workflows
────                          ─────────────────             ───────────────
runs parent workflow ───────▶ loads DAG once
                              [~] evaluates when: via real  
                                  expression evaluator
                              executes layer by layer
                              hits workflow: node ─────────▶ [+] discovers child by name
                                                              [+] creates child WorkflowRun
                                                              [+] runs child DAG end-to-end
                              $childNode.output ◀──────────  [+] terminal output propagated
                              downstream nodes see child output
                              [+] if loop_until defined:
                                  evaluates condition after DAG pass
                                  if false AND iterations < max:
                                    "↻ Iteration N/M — re-running DAG"
                                    re-runs full DAG with fresh scope
                                  if condition met: stops
                                  if max_iterations exhausted: warns
sees output ◀───────────────  final pass output
```

### Before

```
loader.ts
  └── validateDagStructure()    (cycle detect, dep refs, output refs)
  └── dagNodeSchema.safeParse() (CommandNode | PromptNode | BashNode |
[...]

PR #1646, opened 2026-05-11
fix(workflows): make resume explicit via prepareResumedRun / hydrateResumableRun (closes #1392)

### LINKED ISSUE #1392: `archon workflow run` auto-resumes failed runs without `--resume` — docs say opt-in
## Problem

The `WorkflowRunOptions` doc-comment (\`packages/cli/src/commands/workflow.ts\`) says:

> --resume: reuse worktree from last failed run.

…implying `--resume` is opt-in. The CLI usage referenced in `references/cli-commands.md` and the README also presents `--resume` as a separate, explicit subcommand-style flag.

Observed behavior contradicts this: `archon workflow run <name> \"<message>\"` (no `--resume`) **silently auto-resumes** the latest failed run with the same workflow name, replaying cached node outputs.

## Repro

```
# Run 1 — produces a failure
archon workflow run my-workflow \"...\"   # fails at node-3

# Run 2 — no --resume flag
archon workflow run my-workflow \"...\"   # observe:
#   workflow.dag_resuming
#   workflowRunId is the SAME as run 1
#   priorCompletedCount: 2  (parse-args, node-2)
#   node-1 Skipped (prior_success)
#   node-2 Skipped (prior_success)
#   node-3 re-runs
```

Concretely: the run id stays constant across `archon workflow run` invocations as long as the prior run is in a non-terminal-not-completed state. Both runs are observably the same DB row; the second invocation isn't a fresh run.

## Why this is confusing

For agents (and humans) authoring workflows:

- They run, hit a bug, fix the bug, re-run. They expect a clean slate. They get a resume.
- They expect their `parse-args` change to take effect. It doesn't (cached `prior_success`).
- They expect to see the dynamic-args path execute. They see the cached output from the prior args.
- They `cd` into a different state and re-run. They get the prior run's worktree path, not their current one.
- Combined with [issue E](#) (no way to invalidate cached node output), debugging cycles compound: a fix doesn't surface because of cached upstream state, and there's no obvious flag to disable resume.

The auto-resume *is* a great feature. The problem is it's not labelled, not documented as default, and there's no `--no-resume` to opt out.

## What the docs say

`packages/cli/src/commands/workflow.ts`:

```
 * Default: creates worktree with auto-generated branch name (isolation by default).
 * --branch: explicit branch name for the worktree.
 * --no-worktree: opt out of isolation, run in live checkout.
 * --resume: reuse worktree from last failed run.        ← reads as opt-in
 * --from: override base branch (start-point for worktree).
```

`references/cli-commands.md` (Archon skill):

```bash
# Resume a failed workflow (re-runs, skipping completed nodes)
archon workflow resume <run-id>
```

…suggesting resume is an explicit `resume` subcommand or `--resume` flag.

## Proposed direction

Pick one of:

1. **Behavior matches docs** — `archon workflow run` defaults to a fresh run. Auto-resume only when `--resume` is passed (or when the user runs `archon workflow resume <id>`).
2. **Docs match behavior** — document that `archon workflow run` auto-resumes any non-completed prior run for this workflow on this codebase, and add a `--no-resume` (or `--fresh`) flag to opt out.

Either is fine; the current state — silent auto-resume contradicting the docs — is the worst combination. (1) is probably the safer default since fresh-by-default matches CLI conventions almost everywhere; (2) requires less code change.

## Found while

Building a workshop `archon-release` workflow. Spent ~2 iterations confused about why my workflow-YAML edits weren't taking effect — the answer was \"they were, but the cached prior-success nodes were unchanged.\" See [#1389](https://github.com/coleam00/Archon/issues/1389) and the related cache-invalidation issue for adjacent papercuts.

Wirasm:
## Summary

- **Problem:** `executeWorkflow` silently auto-resumed the most recent failed run for the same `(workflow_name, cwd)` pair. The `--resume` CLI flag was documented as opt-in but had no causal effect on this code path; it only steered worktree selection. We reproduced the live failure mode in PR #1282's review queue — `extract-pr-number`'s cached output "1634" leaked into every subsequent invocation of `archon workflow run maintainer-review-pr <N>` for the same `cwd`.
- **Why it matters:** Users who changed inputs or wanted a fresh start silently got a cached resume instead — visible as wrong outputs, ghost state across "unrelated" runs, and behaviour that contradicted the docs.
- **What changed:** Moves the resume decision out of `executeWorkflow` and into the caller. Two new helpers (`hydrateResumableRun`, `prepareResumedRun`) cover the two real lookup shapes. Trailing positional args consolidate into a single `ExecuteWorkflowOptions` bag — fixes the `undefined, undefined, undefined, undefined, true` smell at call sites. Orphan-cleanup code deletes outright.
- **What did NOT change (scope boundary):** Approve/reject auto-resume from chat (still works — orchestrator now calls `hydrateResumableRun` directly with the run row it already found). Web UI resume button. `archon workflow resume <id>` semantics. DB schema. Platform adapters.

### Before

```
User                          CLI                        executeWorkflow
────                          ───                        ───────────────
archon workflow run foo "x"   ─────────────────────────▶ findResumableRun()  ← fires always
                                                         found prior failed run
                                                         ▶ silently resumes (wrong inputs!)
```

### After

```
User                          CLI                        executeWorkflow
────                          ───                        ───────────────
archon workflow run foo "x"   ─────────────────────────▶ no internal lookup
                                                         ▶ creates fresh run ✓

archon workflow run foo "x" --resume
                              looks up + hydrates  ───▶  executes resumed run ✓

archon workflow resume <id>   → workflowRunCommand({resume: true})
                              looks up + hydrates  ───▶  executes resumed run ✓
```

### Before

```
executor.ts
  findResumableRun() ← unconditional, internal
    │
    └─→ if found: resumeWorkflowRun + load priors
        else:     use preCreatedRun or create fresh

CLI workflow.ts (--resume)
  workflowDb.findResumableRun()  ← used only for worktree path selection
  executeWorkflow(...positionals)  ← executor re-discovers via its own findResumableRun

orchestrator-agent.ts (foreground-resume)
  findResumableRunByParentConversation()  ← used only for working_path
  executeWorkflow(...positionals)  ← executor re-discovers via its own findResumableRun
```

### After

```
@archon/workflows/executor (NEW)
  prepareResumedRun(deps, workflow, cwd)
    └── findResumableRun → hydrateResumableRun
  hydrateResumableRun(deps, candidate)
    ├── getCompletedDagNodeOutputs
    ├── return null if 0 nodes + no interactive-loop state
    └── resumeWorkflowRun → { run, priorCompletedNodes }
  executeWorkflow(..., opts: ExecuteWorkflowOptions)
    ├── opts.preCreatedRun ? use it : createWorkflowRun
    ├── opts.priorCompletedNodes → flows to dag-executor
    └── NO store lookups

CLI workflow.ts (--resume)
  workflowDb.findResumableRun()  ← path resolution + candidate
  hydrateResumableRun(deps, candidate)  ← new
  executeWorkflow(..., { preCreatedRun, priorCompletedNodes })

orchestrator-agent.ts (foreground-resume)
  findResumableRunByParentConversation()  ← path resolution + candidate
  hydrateResumableRun(deps, candidate)  ← new
  executeWorkflow(..., { preCreatedRun, priorCompletedNodes })
```

**Connection inventory:**

| `executor.ts` | `findResumableRun` | **removed** | Executor no longer queries the store for prior runs |
| `executor.ts` | `prepareResumedRun` / `hydrateResumableRun` | **new exports** | Caller-invoked resume preparation |
| `executor.ts` | `ExecuteWorkflowOptions` | **new export** | All trailing optional args (codebaseId, issueContext, isolationContext, parentConversationId, preCreatedRun, priorCompletedNodes) |
| `cli/workflow.ts` | `hydrateResumableRun` | **new** | Reuses the candidate already found for worktree-path resolution |
| `orchestrator.ts` | `executeWorkflow` | **call shape** | Trailing args → opts bag |
| `orchestrator-agent.ts` foreground-resume | `hydrateResumableRun` | **new** | Was relying on executor internal auto-resume |
| `orchestrator-agent.ts` interactive | `executeWorkflow` | **call shape** | Trailing args → opts bag |
| Orphan-cleanup code in `executor.ts` (~120 lines) | — | **removed** | No `preCreatedRun` to orphan now that resume flows through opts |

## Label Snapshot

- Risk: `risk: low` — closes #1392; behaviour now matches documented `--resume` semantics; smaller diff than #1396.
- Size: `size: M`
- Scope: `workflows`, `cli`, `core`
- Module: `workflows:executor`, `cli:workflow`, `core:orchestrator`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

Closes #1392. Supersedes #1396.

```bash
bun run validate  # ✅ exit 0
```

- `check:bundled`, `check:bundled-skill`, `type-check`, `lint --max-warnings 0`, `format:check`, all per-package test batches — green.
- 72 test batches passed, 0 failures. New / changed tests:
  - `prepareResumedRun` suite (`null` when no resumable, hydrates candidate, propagates DB errors)
  - `hydrateResumableRun` suite (hydrate, return null on zero+no-loop, paused-loop hydration, DB error propagation on both `getCompletedDagNodeOutputs` and `resumeWorkflowRun`)
  - **Regression test for #1392 cross-invocation leak**: asserts the executor does NOT call `findResumableRun` on its own
  - `executor.test.ts` "uses caller-supplied preCreatedRun + priorCompletedNodes" — asserts opts flow into dag-executor at the right slot
  - Updated orchestrator + orchestrator-agent tests to assert resume payload on the opts bag

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? **No for `executeWorkflow` callers (API)**, **yes for users of documented CLI/UI surfaces**.
  - The signature change is breaking for any external caller of `executeWorkflow`. There are no public extension points that call it directly; in-tree callers are migrated in this commit.
  - User-facing behaviour: `archon workflow run <name>` always starts fresh (matches docs); resume requires `--resume`, `archon workflow resume <id>`, the web UI button, or approve/reject in chat (unchanged).
- Config/env changes? `No`

- **Verified scenarios:** Full validate suite passes; manually ran the failing reproduction from the standup queue (back-to-back `archon workflow run maintainer-review-pr <N>` for different `<N>`) — confirms each invocation now creates a fresh run, no `dag.node_skipped_prior_success` leak.
- **Edge cases checked:**
  - Plain run with no prior failed run — fresh run created.
  - Plain run when a prior failed run exists for `(workflow, cwd)` — fresh run created (no implicit resume). This is the bug-fix.
  - `--resume` with a resumable prior run — resumes and skips completed nodes.
  - `--resume` with no resumable prior run — clear error, no fresh run.
  - Orchestrator foreground-resume (approve/reject from chat) — `hydrateResumableRun` succeeds and resume continues.
- **What was not verified:**
  - Live end-to-end web UI resume button flow (the bridge call path is unchanged; only positional → opts call shape).

- **Affected subsystems:** Workflow executor (`packages/workflows/src/executor.ts`), CLI `workflow run` / `workflow resume`, orchestrator chat dispatch (background + foreground-resume + interactive paths). No DB schema or adapter changes.
- **Potential unintended effects:** Anyone scripting against the previous undocumented implicit auto-resume behaviour will now get fresh runs. This matches docs.
- **Guardrails:** Existing tests for `--resume`, approve/reject, foreground-resume, and concurrent-run-guard are all updated and pass. A new regression test pins down the cross-invocation leak from #1392.

## Rollback Plan (required)

- **Fast rollback:** Revert the commit.
- **Feature flags:** None.
- **Observable failure symptoms:** Users reporting "`--resume` does not resume" (would indicate a positional/opts wiring bug). The fix has zero observable change for invocations that don't use `--resume` — they always start fresh, as documented.

## Risks and Mitigations

- **Risk:** Users who depended on the undocumented implicit auto-resume will now get a fresh run.
  - Mitigation: This matches the documented behaviour. The `--resume` flag and `/workflow resume <id>` command are the explicit paths. CHANGELOG entry calls this out.
- **Risk:** Test-mock divergence on `@archon/workflows/executor` for callers that mock it.
  - Mitigation: Updated all in-tree mocks. The new `hydrateResumableRun` export is part of the module signature, so consumers needing tighter test-mocks now know the surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow run no longer auto-resumes on re-run; resume is explicit via CLI `--resume`, `archon workflow resume <id>`, or Web UI resume. Web resume now validates a parent web conversation and dispatches a resume message.

* **Documentation**
  * Clarified resume mechanics across guides and API docs: resuming skips already-completed steps and is opt-in.

* **Tests**
  * Expanded tests to cover resume paths and error conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1646)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `archon workflow run` no longer silently auto-resumes the previous failed run for the same `(workflow_name, cwd)` pair. The implicit `findResumableRun` call inside `executeWorkflow` was the cause of cross-invocation state leaks — completed-node outputs from a prior failed run would bleed into the next invocation of the same workflow at the same path. Resume is now an explicit caller-side decision: use `archon workflow run --resume`, `archon workflow resume <id>`, or the web UI resume button. `executeWorkflow`'s trailing positional args are consolidated into an options bag and a new `prepareResumedRun` / `hydrateResumableRun` pair handles resume preparation at call sites. Closes #1392.
 - `archon doctor` and `archon setup` no longer interleave `[archon] loaded N keys` boot lines and Pino info JSON with their checklist output. Set `ARCHON_VERBOSE_BOOT=1` or `LOG_LEVEL=debug` to restore the boot lines; pass `--verbose` to re-enable structured Pino logs for those commands (#1606).
 - Docker: `git config --global --add safe.directory` in the entrypoint now de-duplicates entries before adding, preventing unbounded growth of `~/.gitconfig` now that `/home/appuser` is persisted (#1518).
 - Docker: `setup-auth` now warns at startup when `CODEX_*` env vars are absent but a persisted `~/.codex/auth.json` from a previous run still exists, so operators don't accidentally use stale or revoked credentials (#1518).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -788,7 +788,7 @@ Pattern: Use `classifyIsolationError()` (from `@archon/isolation`) to map git er
 - `DELETE /api/workflows/:name` - Delete a user-defined workflow; bundled defaults cannot be deleted
 
 **Workflow Run Lifecycle:**
-- `POST /api/workflows/runs/{runId}/resume` - Mark a failed run as ready for auto-resume on next invocation
+- `POST /api/workflows/runs/{runId}/resume` - Resume a failed run from where it left off (skips already-completed DAG nodes; AI session context is not restored).
 - `POST /api/workflows/runs/{runId}/abandon` - Abandon a non-terminal run (marks as cancelled)
 - `DELETE /api/workflows/runs/{runId}` - Delete a terminal workflow run and its events
 
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -75,6 +75,7 @@ mock.module('@archon/workflows/workflow-discovery', () => ({
 }));
 mock.module('@archon/workflows/executor', () => ({
   executeWorkflow: mock(() => Promise.resolve({ success: true, workflowRunId: 'test-run-id' })),
+  hydrateResumableRun: mock(() => Promise.resolve(null)),
 }));
 
 // Capture the subscription handler so tests can trigger events
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -15,7 +15,7 @@ import { join } from 'node:path';
 import { createWorkflowDeps } from '@archon/core/workflows/store-adapter';
 import { discoverWorkflowsWithConfig } from '@archon/workflows/workflow-discovery';
 import { resolveWorkflowName } from '@archon/workflows/router';
-import { executeWorkflow } from '@archon/workflows/executor';
+import { executeWorkflow, hydrateResumableRun } from '@archon/workflows/executor';
 import {
   getWorkflowEventEmitter,
   type WorkflowEmitterEvent,
@@ -427,9 +427,10 @@ export async function workflowRunCommand(
   let workingCwd = cwd;
   let isolationEnvId: string | undefined;
 
-  // Handle --resume: find the most recent failed run and reuse its worktree.
-  // The executor's implicit findResumableRun will detect the failed run and
-  // skip already-completed nodes automatically.
+  // Handle --resume: locate the prior failed run, reuse its worktree, and hand
+  // the resumed-run handle to executeWorkflow below via opts. The executor no
+  // longer performs implicit resume detection on its own.
+  let resumable: WorkflowRun | null = null;
   if (options.resume) {
     if (!codebase) {
       if (codebaseLookupError) {
@@ -448,7 +449,7 @@ export async function workflowRunCommand(
       );
     }
 
-    const resumable = await workflowDb.findResumableRun(workflowName, cwd);
+    resumable = await workflowDb.findResumableRun(workflowName, cwd);
 
     if (!resumable) {
       throw new Error(`No resumable run found for workflow '${workflowName}' at path '${cwd}'.`);
@@ -704,18 +705,37 @@ export async function workflowRunCommand(
     );
   }
 
+  // When --resume, hand the already-found run (and its completed-node outputs)
+  // to executeWorkflow. Otherwise this is a fresh run and prepared stays null.
+  // The lookup-by-(workflowName, cwd) was already done above for worktree-path
+  // resolution; reuse that result rather than querying twice.
+  const deps = createWorkflowDeps();
+  let prepared: Awaited<ReturnType<typeof hydrateResumableRun>> = null;
+  if (options.resume && resumable) {
+    prepared = await hydrateResumableRun(deps, resumable);
+    if (!prepared) {
+      throw new Error(
+        `Cannot resume: the prior run for '${workflowName}' has no completed nodes and no interactive-loop state.`
+      );
+    }
+  }
+
   // Execute workflow with workingCwd (may be worktree path)
   let result: Awaited<ReturnType<typeof executeWorkflow>>;
   try {
     result = await executeWorkflow(
-      createWorkflowDeps(),
+      deps,
       adapter,
       conversationId,
       workingCwd,
       workflow,
       userMessage,
       conversation.id,
-      codebase?.id
+      {
+        codebaseId: codebase?.id,
+        preCreatedRun: prepared?.run,
+        priorCompletedNodes: prepared?.priorCompletedNodes,
+      }
     );
   } finally {
     unsubscribe?.();
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -712,7 +712,18 @@ export async function workflowRunCommand(
   const deps = createWorkflowDeps();
   let prepared: Awaited<ReturnType<typeof hydrateResumableRun>> = null;
   if (options.resume && resumable) {
-    prepared = await hydrateResumableRun(deps, resumable);
+    try {
+      prepared = await hydrateResumableRun(deps, resumable);
+    } catch (error) {
+      const err = error as Error;
+      getLog().error(
+        { err, workflowName, runId: resumable.id },
+        'cli.workflow_hydrate_resume_failed'
+      );
+      throw new Error(
+        `Cannot resume workflow '${workflowName}': failed to load prior run state — ${err.message}`
+      );
+    }
     if (!prepared) {
       throw new Error(
         `Cannot resume: the prior run for '${workflowName}' has no completed nodes and no interactive-loop state.`
@@ -723,6 +734,9 @@ export async function workflowRunCommand(
   // Execute workflow with workingCwd (may be worktree path)
   let result: Awaited<ReturnType<typeof executeWorkflow>>;
   try {
+    const opts = prepared
+      ? { codebaseId: codebase?.id, ...prepared }
+      : { codebaseId: codebase?.id };
     result = await executeWorkflow(
       deps,
       adapter,
@@ -731,11 +745,7 @@ export async function workflowRunCommand(
       workflow,
       userMessage,
       conversation.id,
-      {
-        codebaseId: codebase?.id,
-        preCreatedRun: prepared?.run,
-        priorCompletedNodes: prepared?.priorCompletedNodes,
-      }
+      opts
     );
   } finally {
     unsubscribe?.();
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -106,8 +106,16 @@ mock.module('@archon/workflows/router', () => ({
     workflows.find(w => w.name === name)
   ),
 }));
+const mockHydrateResumableRun = mock(
+  async (_deps: unknown, candidate: { id: string }) =>
+    ({
+      run: { ...candidate, status: 'running' },
+      priorCompletedNodes: new Map([['n1', 'v1']]),
+    }) as unknown
+);
 mock.module('@archon/workflows/executor', () => ({
   executeWorkflow: mockExecuteWorkflow,
+  hydrateResumableRun: mockHydrateResumableRun,
 }));
 
 mock.module('@archon/providers', () => ({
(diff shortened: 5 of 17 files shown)


Wirasm:

Seven specialized agents reviewed this PR: `code-reviewer`, `docs-impact`, `pr-test-analyzer`, `comment-analyzer`, `silent-failure-hunter`, `type-design-analyzer`, `code-simplifier`. Findings consolidated below.

### Critical Issues (1)

| Agent | Issue | Location |
|-------|-------|----------|
| code-reviewer | **Web UI Resume button is functionally broken.** The PR removes auto-resume on plain re-invocation, but `POST /api/workflows/runs/{runId}/resume` was not updated. It still returns `"Re-run the workflow to auto-resume from completed nodes."` and the inline comment still says `the next invocation on the same path auto-resumes`. PR docs explicitly list "Web UI: Resume button on the workflow card" as a valid resume path, which is now untrue. Verified against the PR branch. | `packages/server/src/routes/api.ts:1910-1914` (handler), `:627` (route summary `"Resume a failed workflow run (re-run auto-resumes from completed nodes)"`) |

### Important Issues (5)

| Agent | Issue | Location |
|-------|-------|----------|
| code-reviewer / code-simplifier (consensus) | **`prepareResumedRun` has zero production callers** — only test files import it. Violates the YAGNI / "one current caller" rule in CLAUDE.md. CLI and orchestrator both call `hydrateResumableRun` directly. Suggest unexporting (or inlining the 3-line body) and folding tests into the `hydrateResumableRun` suite with a `findResumableRun` mock. | `packages/workflows/src/executor.ts` (export of `prepareResumedRun`) |
| type-design-analyzer | **`preCreatedRun` and `priorCompletedNodes` are independently optional but jointly required for resume.** A caller passing only `priorCompletedNodes` (or only `preCreatedRun`) yields silent corruption: executor creates a fresh row via `createWorkflowRun` but still uses `dagPriorCompletedNodes = priorCompletedNodes` to skip nodes — the mirror of the #1392 bug this PR fixes. Type system does not catch it. Suggest a discriminated union: `({preCreatedRun, priorCompletedNodes}) \| ({preCreatedRun?: undefined; priorCompletedNodes?: undefined})`. | `packages/workflows/src/executor.ts` `ExecuteWorkflowOptions` |
| silent-failure-hunter | **Foreground-resume throw message exceeds the 100-char threshold in `classifyAndFormatError`**, so the user sees the generic `"⚠️ An unexpected error occurred. Try /reset to start a fresh session."` instead of the actionable message. `/reset` is also the wrong recovery — the user should `abandon` the stuck run. Verified: message template is ~62 chars + UUID(36) + workflow name = well over 100. | `packages/core/src/orchestrator/orchestrator-agent.ts` (foreground-resume throw) → `packages/core/src/utils/error-formatter.ts:88-95` |
| silent-failure-hunter | **CLI `hydrateResumableRun` call has no try/catch.** A DB error here propagates as a raw stack trace and gets logged under the misleading `cli.workflow_resume_run_failed` event name. All other CLI failure paths wrap with context. | `packages/cli/src/commands/workflow.ts` ~line 715 |
| docs-impact | **Stale user-facing docs that contradict the new explicit-only semantics**, not touched by this PR: <br>• `packages/docs-web/src/content/docs/book/dag-workflows.md:329` — "Let DAG resume handle failures... run it again... No `--resume` flag required."<br>• `packages/docs-web/src/content/docs/guides/authoring-workflows.md:973,999` — "Checkpoint and Resume" pattern still describes resume as automatic on re-invocation.<br>• `packages/docs-web/src/content/docs/guides/approval-nodes.md:237-241` — references `findResumableRun` by internal-function name as the resume mechanism (stale implementation detail). | (see locations) |

### Suggestions (5)

| comment-analyzer | **Wrong arg index in test comment.** Comment says "arg index 14" but assertion accesses `[15]`. The enumerated signature in the same comment is correct (`priorCompletedNodes` is 16th, index 15) — only the index label is off. | `packages/workflows/src/executor.test.ts:559-560` |
| comment-analyzer | **Issue-number references in source/test comments** (e.g. "Silent fallback... was the original #1392 bug", `#1392 follow-up`). CLAUDE.md guidance: keep issue numbers in PR descriptions, not in code — they rot. | `executor.ts` `hydrateResumableRun` JSDoc; `executor.test.ts` removed-suite tombstones; test name `regression: #1392` |
| comment-analyzer | **Position-leaky test comment** ("opts.parentConversationId on the options bag at position 7") will rot if `executeWorkflow` ever gains another positional arg. The assertion `opts.parentConversationId` is self-documenting. | `packages/core/src/orchestrator/orchestrator-agent.test.ts` |
| pr-test-analyzer | **No test for the new CLI throw "Cannot resume: the prior run for '...' has no completed nodes and no interactive-loop state."** (rating 8/10). Add a test that mocks `findResumableRun` → candidate but `hydrateResumableRun` → null. | `packages/cli/src/commands/workflow.test.ts` |
| pr-test-analyzer | **No test for the orchestrator foreground-resume `hydrateResumableRun` → null branch** (rating 7/10). Pairs with the silent-failure issue above — would surface the >100-char error formatting too. | `packages/core/src/orchestrator/orchestrator-agent.test.ts` |
| type-design-analyzer | `hydrateResumableRun` return key `run` doesn't align with `ExecuteWorkflowOptions.preCreatedRun` — three call sites need `{ preCreatedRun: prepared.run, priorCompletedNodes: prepared.priorCompletedNodes }`. Renaming `run` → `preCreatedRun` enables a clean `...prepared` spread. | `packages/workflows/src/executor.ts` `hydrateResumableRun` return type |

### Strengths

- **Core refactor is sound.** Removing implicit `findResumableRun` from `executeWorkflow` is the right fix for #1392; the regression test (`does NOT call findResumableRun on its own`) precisely pins down the cross-invocation leak.
- **Opts bag eliminates the `undefined, undefined, undefined, undefined, true` smell** at every call site — genuine maintenance win.
- **`hydrateResumableRun` propagates DB errors without silent fallback** — `pr-test-analyzer` confirms both `getCompletedDagNodeOutputs` and `resumeWorkflowRun` failure paths are tested with `rejects.toThrow`. Directly fixes the original #1392 root cause.
- **`ExecuteWorkflowOptions` JSDoc is strong** — each field documents the non-obvious contract (e.g. `preCreatedRun` must be in `running` status; `issueContext` substitution/append behavior).
- **Orchestrator foreground-resume comment is exemplary**: "Hydrate the already-found candidate (skip a second findResumableRun query). If hydration returns null, the run vanished or has nothing worth resuming — throw rather than silently start fresh." — captures non-obvious WHY in three lines.
- **Mock isolation is clean** — no collisions across batches; both packages run conflicting `mock.module` calls in separate `bun test` invocations.

### Verdict

**NEEDS FIXES** — one blocker (broken Web UI button) plus YAGNI + invariant gap.

### Recommended Actions

1. **Block-merge fix:** Update `POST /api/workflows/runs/{runId}/resume` and its route summary in `packages/server/src/routes/api.ts`. Either dispatch the resume from the handler, or change the response message + summary to remove the now-false "auto-resume on re-run" language and point the user at `archon workflow resume <id>`.
2. **YAGNI:** Remove `prepareResumedRun` (or stop exporting it). Fold its 3 tests into the `hydrateResumableRun` suite.
3. **Type invariant:** Make `preCreatedRun` + `priorCompletedNodes` a discriminated union on `ExecuteWorkflowOptions` so the type system enforces what the JSDoc currently only documents.
4. **Silent failure:** Shorten the foreground-resume throw message to <100 chars (or surface it directly via `safeSendMessage` before rethrowing) so `classifyAndFormatError` passes it through. Wrap the CLI `hydrateResumableRun` call in try/catch with context.
5. **Docs:** Update the three stale doc locations the docs-impact agent identified (`book/dag-workflows.md:329`, `authoring-workflows.md:973,999`, `approval-nodes.md:237-241`).
6. **Polish:** Fix the `arg index 14` → `15` comment in `executor.test.ts:559`. Consider renaming `hydrateResumableRun` return key `run` → `preCreatedRun` to enable `...prepared` spread at call sites.
7. **Tests:** Add the two missing null-branch tests (CLI `prepared === null`, orchestrator hydration null).

PR #1661, opened 2026-05-13
fix(providers): preserve native tools when skills are set without allowed_tools (fixes #1605)

### LINKED ISSUE #1605: skills: wrapper strips native tools when allowed_tools is unset (silently breaks bundled archon-create-issue reproduce)
## Bug Description

When a DAG workflow node has `skills:` but no explicit `allowed_tools:`, the AgentDefinition wrapper strips ALL native Claude Code tools, leaving only the `Skill` tool available. Native tools (`Read`, `Grep`, `Glob`, `Bash`, `Write`, `Edit`) become unavailable to the AI sub-agent.

This silently breaks any node whose prompt expects to invoke binaries via `Bash` (e.g. `agent-browser`, `gh`, `curl`) or to read files via `Read`/`Grep`. The bundled `archon-create-issue.yaml` `reproduce` node hits this — its playbook is written to invoke `agent-browser open URL`, `agent-browser screenshot ...`, etc., as Bash commands, but the AI sees only `Skill` available and bails with "Cannot do task. No Bash tool available."

Closely related to #1343 (same code path: `provider.ts:436-459` AgentDefinition wrapper), different symptom (tools stripped vs model ignored).

## Root Cause

`packages/providers/src/claude/provider.ts:436-459`:

```typescript
if (nodeConfig.skills) {
  const skills = nodeConfig.skills;
  const agentId = 'dag-node-skills';
  const agentTools = options.tools ? [...(options.tools as string[]), 'Skill'] : ['Skill'];
  //                                                                                  ^^^^^^^^^
  //                                                          when nodeConfig.allowed_tools is undefined,
  //                                                          options.tools is undefined,
  //                                                          and agentTools defaults to ['Skill'] only
  const agentDef = {
    description: 'DAG node with skills',
    prompt: `You have preloaded skills: ${skills.join(', ')}. Use them when relevant.`,
    skills,
    tools: agentTools,
  };
  options.agents = { [agentId]: agentDef };
  options.agent = agentId;
  // ...
}
```

`options.tools` is set from `nodeConfig.allowed_tools` at `provider.ts:372-373`:

```typescript
if (nodeConfig.allowed_tools !== undefined) {
  options.tools = nodeConfig.allowed_tools;
}
```

So if YAML doesn't specify `allowed_tools:`, the AgentDefinition's `tools:` field becomes `['Skill']` only — no Read, Bash, Write, etc.

## Reproduction

**Workflow YAML** (e.g. drop in `~/.archon/workflows/skills-tool-test.yaml`):

```yaml
name: skills-tool-test
description: Demonstrates that skills strip native tools
provider: claude

nodes:
  - id: with-skills-no-allowed-tools
    prompt: |
      Run `ls /` via Bash and report the output. Read `/etc/hostname` via Read.
    skills:
      - caveman   # any installed skill (only the FIELD presence matters here)

  - id: with-skills-and-allowed-tools
    prompt: |
      Run `ls /` via Bash and report the output. Read `/etc/hostname` via Read.
    skills:
      - caveman
    allowed_tools:
      - Bash
      - Read
```

**Expected:** Both nodes execute the prompt's tool calls.

**Actual:** `with-skills-no-allowed-tools` reports "Cannot do task. Only `Skill` tool available this session. No `Bash`/`Read` tools." `with-skills-and-allowed-tools` works correctly.

## Real-World Impact

Bundled `archon-create-issue.yaml` `reproduce` node has:

```yaml
- id: reproduce
  prompt: |
    ## Reproduction Playbooks
    ### web-ui
    1. Read the server port: `PORT=$(cat "$ARTIFACTS_DIR/.server-port" | tr -d '\n')`
    2. Open the app: `agent-browser open http://localhost:$PORT`
    3. Take a baseline screenshot: `agent-browser screenshot "$ARTIFACTS_DIR/repro-01-baseline.png"`
    ...
  skills:
    - agent-browser
  # NO allowed_tools — Bash gets stripped
```

Result: AI cannot invoke `agent-browser` binary via Bash. Falls back to Skill('agent-browser', ...) but the prompt is written for Bash invocation, not Skill invocation. Bundled workflow silently fails to reproduce real bugs — gating logic then skips `create-issue` step. No GitHub issue gets filed.

Confirmed across multiple runs:
- WSL CLI: `archon workflow run archon-create-issue --no-worktree "<bug>"` — `reproduce` says "no Bash tool exposed"
- Container Web UI: same symptom

Lean variants that explicitly set `allowed_tools: [Bash, Read, Write, ...]` work correctly. The wrapper was probably designed for skill-only sub-agents (where Skill is sufficient), but the bundled workflows mix native-tool playbooks with `skills:` declarations — incompatible without explicit allowlist.

## Suggested Fix

Several options, in order of cleanliness:

**Option A (minimal change)** — Default `agentTools` to all native tools plus Skill when `options.tools` is unset:

```typescript
const DEFAULT_NATIVE_TOOLS = ['Read', 'Write', 'Edit', 'Glob', 'Grep', 'Bash', 'NotebookEdit', 'WebSearch', 'WebFetch', 'TodoWrite', 'Task'];
const agentTools = options.tools
  ? [...(options.tools as string[]), 'Skill']
  : [...DEFAULT_NATIVE_TOOLS, 'Skill'];
```

Risk: hard-codes the SDK's default tool list; needs maintenance if SDK adds new tools.

**Option B (no defaults)** — Don't set `tools:` on AgentDefinition when `options.tools` is undefined; let SDK use its defaults:

```typescript
const agentDef: AgentDefinition = {
  description: 'DAG node with skills',
  prompt: `You have preloaded skills: ${skills.join(', ')}. Use them when relevant.`,
  skills,
};
if (options.tools) {
  agentDef.tools = [...(options.tools as string[]), 'Skill'];
}
// Skill auto-added to options.allowedTools below covers visibility
```

Cleaner — defers to SDK's default-tool resolution. Validate against SDK behavior.

**Option C (validation warning)** — Keep current behavior but emit a loud warning when `skills:` is set without `allowed_tools:`:

```typescript
if (nodeConfig.skills && !nodeConfig.allowed_tools) {
  warnings.push({
    code: 'skills_without_allowed_tools',
    message: 'Node has skills: but no allowed_tools: — native tools will be stripped. Add allowed_tools: [Read, Bash, Write, ...] to restore tool access.',
  });
}
```

Less invasive; doesn't change behavior, just surfaces the gotcha.

## Related

- #1343 — same code path, different symptom (model override ignored)
- Bundled `archon-create-issue.yaml` `reproduce` node would benefit from either an upstream fix OR an explicit `allowed_tools:` add
- The agent-browser SKILL.md isn't bundled in the Docker image either (binary is, file isn't) — separate but adjacent gap; users who want the Skill-tool route to work must hand-author SKILL.md based on `agent-browser --help` output

## Environment

- Archon: v0.3.10 + local docker patches (commit `86379528`)
- Claude Agent SDK: 0.2.121
- Provider: Claude (sonnet)
- Reproduced on: homeserverai Docker (Debian glibc) AND local WSL (Ubuntu glibc)

kagura-agent:
## Problem

When a DAG workflow node has `skills:` but no explicit `allowed_tools:`, the AgentDefinition wrapper defaults `tools` to `["Skill"]` only, stripping all native Claude Code tools (`Read`, `Bash`, `Write`, `Edit`, `Grep`, `Glob`, etc.).

This silently breaks any node whose prompt expects native tool access — including the bundled `archon-create-issue.yaml` `reproduce` node, which uses `agent-browser` via `Bash` commands.

**Root cause:** `provider.ts:439` — when `options.tools` is undefined (no `allowed_tools` in YAML), `agentTools` defaults to `["Skill"]`, and the AgentDefinition `tools` field restricts the sub-agent to only `Skill`.

## Fix

Omit the `tools` field on AgentDefinition when `options.tools` is undefined, letting the SDK provide its full default tool set. When `allowed_tools` is explicitly set, `Skill` is still appended to the explicit list (existing behavior preserved).

This is **Option B** from the issue — the cleanest approach since it defers to SDK default-tool resolution without hardcoding a tool list that needs maintenance.

## Changes

- `packages/providers/src/claude/provider.ts` — Make `tools` optional on the AgentDefinition type; only set it when `options.tools` is defined
- `packages/providers/src/claude/provider.test.ts` — Add 2 tests:
  - Skills without `allowed_tools` → `tools` field is undefined (SDK defaults apply)
  - Skills with `allowed_tools` → `tools` includes explicit list + `Skill`

## Validation

- `bun test packages/providers/src/claude/provider.test.ts` — 74 pass, 0 fail
- `bun run type-check` — all packages clean
- lint-staged clean on commit

---

## UX Journey

**Before:** Node with `skills: [agent-browser]` but no `allowed_tools` → AI sees only `Skill` tool → "Cannot do task. No Bash tool available."

**After:** Node with `skills: [agent-browser]` → AI sees all default native tools + `Skill` → executes `agent-browser` via Bash as expected.

## Architecture Diagram

```
nodeConfig.skills present
├── allowed_tools defined → agentDef.tools = [...allowed_tools, "Skill"]
└── allowed_tools NOT defined → agentDef.tools omitted (SDK defaults apply)
```

## Label Snapshot

`fix`, `providers`, `workflows`

## Change Metadata

- **Scope:** `packages/providers/src/claude/provider.ts`
- **Lines changed:** +50 / -3

## Linked Issue

Closes #1605

## Validation Evidence

All 74 provider tests pass including 2 new regression tests.

## Security Impact

None — this change only affects which tools are available to sub-agents, and only makes the behavior match what users would expect (all default tools available unless explicitly restricted).

## Compatibility/Migration

Fully backward compatible. Nodes with explicit `allowed_tools` behave identically. Only nodes with `skills` + no `allowed_tools` are affected (they now get SDK defaults instead of `Skill`-only).

## Human Verification

- [x] Diff reviewed line-by-line
- [x] Tests pass
- [x] Type check clean

## Risks and Mitigations

**Risk:** SDK default tool set might change in future versions.
**Mitigation:** This is the intended behavior — deferring to SDK defaults is more maintainable than hardcoding a tool list.

## Side Effects/Blast Radius

Only affects DAG nodes with `skills:` and no `allowed_tools:`. No impact on other code paths.

## Rollback Plan

Revert the single commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tools configuration handling for agent skills. The tools field is now conditionally set based on provided options, enabling SDK defaults to properly apply when configuration is not explicitly specified.

* **Tests**
  * Added test coverage validating tools configuration behavior in two important scenarios: when allowed_tools parameters are provided and when they are not specified.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1661)

### DIFF
--- a/packages/providers/src/claude/provider.test.ts
+++ b/packages/providers/src/claude/provider.test.ts
@@ -1347,6 +1347,52 @@ describe('sendQuery decomposition behaviors', () => {
       );
     });
 
+    test('skills without allowed_tools omits tools field so SDK defaults apply', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield { type: 'result', session_id: 'sid' };
+      });
+
+      for await (const _ of client.sendQuery('test', '/workspace', undefined, {
+        nodeConfig: {
+          skills: ['agent-browser'],
+          // no allowed_tools → options.tools is undefined
+        },
+      })) {
+        // consume
+      }
+
+      const callArgs = mockQuery.mock.calls[0][0] as { options: Record<string, unknown> };
+      const outAgents = callArgs.options.agents as Record<
+        string,
+        { description: string; tools?: string[] }
+      >;
+      // tools should NOT be set — lets SDK provide all default native tools
+      expect(outAgents['dag-node-skills'].tools).toBeUndefined();
+    });
+
+    test('skills with allowed_tools includes Skill in the tools list', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield { type: 'result', session_id: 'sid' };
+      });
+
+      for await (const _ of client.sendQuery('test', '/workspace', undefined, {
+        nodeConfig: {
+          skills: ['agent-browser'],
+          allowed_tools: ['Bash', 'Read'],
+        },
+      })) {
+        // consume
+      }
+
+      const callArgs = mockQuery.mock.calls[0][0] as { options: Record<string, unknown> };
+      const outAgents = callArgs.options.agents as Record<
+        string,
+        { description: string; tools?: string[] }
+      >;
+      // tools should include the explicit list plus Skill
+      expect(outAgents['dag-node-skills'].tools).toEqual(['Bash', 'Read', 'Skill']);
+    });
+
     test('does NOT warn when inline agents do not collide with the skills wrapper', async () => {
       mockQuery.mockImplementation(async function* () {
         yield { type: 'result', session_id: 'sid' };
--- a/packages/providers/src/claude/provider.ts
+++ b/packages/providers/src/claude/provider.ts
@@ -436,19 +436,20 @@ async function applyNodeConfig(
   if (nodeConfig.skills) {
     const skills = nodeConfig.skills;
     const agentId = 'dag-node-skills';
-    const agentTools = options.tools ? [...(options.tools as string[]), 'Skill'] : ['Skill'];
     const agentDef: {
       description: string;
       prompt: string;
       skills: string[];
-      tools: string[];
+      tools?: string[];
       model?: string;
     } = {
       description: 'DAG node with skills',
       prompt: `You have preloaded skills: ${skills.join(', ')}. Use them when relevant.`,
       skills,
-      tools: agentTools,
     };
+    if (options.tools) {
+      agentDef.tools = [...(options.tools as string[]), 'Skill'];
+    }
     if (options.model) agentDef.model = options.model;
     options.agents = { [agentId]: agentDef };
     options.agent = agentId;


Wirasm:
## Review Summary

**Verdict**: ready-to-merge

Your fix correctly restores native tool access for workflow nodes with `skills:` but no explicit `allowed_tools:`. The change is minimal, properly typed, and well-tested.

### Blocking issues
None.

### Suggested fixes
None.

### Minor / nice-to-have
- **`provider.test.ts`**: Consider adding a comment or test case for `allowed_tools: []` (empty array). In JavaScript, `[]` is truthy, so the explicit-tools branch is taken, but this edge case isn't currently exercised explicitly. A comment noting this behavior would prevent future confusion.

### Compliments
- Clean, targeted fix that addresses the root cause without introducing scope creep
- Two tests cover both branches of the conditional, making the behavior self-documenting
- PR description is thorough with architecture diagram, validation evidence, and rollback plan

---
*Reviewed via maintainer-review-pr workflow (Pi/Minimax). Aspects run: code-review, test-coverage.*

Issue #1123, opened 2026-04-12
CLI workflow approve marks paused workflows as failed on process exit


joelsb:
## Bug Description

When using `archon workflow approve` from the CLI to resume an interactive workflow that has multiple gates (e.g., two interactive loops), the workflow is marked as `failed` in the database when the CLI process exits — even though the workflow correctly paused at the next interactive gate.

## Steps to Reproduce

1. Create a workflow with 2+ interactive loop nodes (e.g., `capture-requirements` → `confirm-spec`)
2. Run: `archon workflow run my-workflow "input"`
3. Workflow pauses at first gate — status is `paused` ✅
4. Run: `archon workflow approve <run-id> "response"`
5. The DAG resumes, processes the response, advances to the next node, and pauses at the second gate
6. The CLI process exits (exit code 0)
7. **Bug**: The workflow status in the DB is now `failed` instead of `paused`
8. Subsequent `archon workflow approve` rejects with: `Cannot approve run with status 'running'. Only paused runs can be approved.`

## Expected Behavior

When the DAG executor pauses at an interactive gate and the CLI `approve` process exits, the workflow status should remain `paused` in the database — not be converted to `failed`.

## Workaround

Manually reset the status in the DB before each approve:

```bash
sqlite3 ~/.archon/archon.db "UPDATE remote_agent_workflow_runs SET status='paused' WHERE id='<run-id>';"
archon workflow approve <run-id> "response"
```

## Environment

- Archon: dev branch (built from source)
- Platform: macOS (Darwin 25.3.0)
- CLI: `bun link @archon/cli`

## Analysis

The issue appears to be in `packages/cli/src/commands/workflow.ts` — the approve command runs the DAG executor inline. When the executor returns after hitting the next interactive gate (`dag.stop_detected_between_layers`), the CLI process exits and the workflow status transitions to `failed` instead of preserving the `paused` state.

The DAG executor logs confirm the correct pause:
```
dag.stop_detected_between_layers (status: paused)
dag.skip_complete_status_changed (status: paused)
dag_workflow_finished
```

But the DB shows `failed` after the CLI process terminates.

## Additional Context

A second related issue: `archon workflow approve` sometimes returns `Cannot approve run with status 'running'` when the workflow is actually paused — suggesting a race condition between the approval gate state and the DB status field.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

cryptobirr:
[...]

Issue #972, opened 2026-04-08
feat: DAG loop-back nodes — conditional retry and loop-within-DAG


Wirasm:
> Migrated from dynamous-community/remote-coding-agent#792 — Archon active development has moved to coleam00/Archon. Original issue retained as historical reference.

---

## Summary

Enable test→fix→retest and other "re-run until condition" patterns in workflows. **Rescoped from a sub-DAG node to a composition-based design** after a primitives review (see History below).

This issue now tracks two pieces of work:

1. **Prerequisite:** a real expression evaluator for `when:` / `until:` conditions
2. **Feature:** workflow invocation node + workflow-level `loop_until`

## Problem

The DAG topology is computed once at load time (`dag-executor.ts`) and back-edges are rejected by cycle detection (`loader.ts`). There's no way to express "run these nodes until tests pass." The only existing iteration primitive is PR #785's `loop:` node, which iterates a **single prompt** — not a multi-step graph.

The naive workaround is duplicating nodes (`run-tests`, `fix`, `run-tests-2`, `fix-2`), which is fragile and caps at whatever count is hardcoded.

## Why not a sub-DAG node (the original proposal)

The original proposal added a `loop_node` containing an inner `nodes:` list with its own edges. From a primitives standpoint this introduces a **composite-with-nested-scope** concept that brings real design debt:

- Nested topology (a node that has internal edges)
- Scoped variable resolution — what does `$test.output` mean across iterations? Not specified.
- Recursive validation (cycle detection, ref checks, model compat all need to recurse)
- Recursive execution (nested cancellation, timeouts, events, JSONL logging)
- A second meaning for `nodes:` (outer DAG nodes vs. inner loop-body nodes)
- Once nesting is allowed one level, two-level nesting becomes the next ask

The motivating examples don't require any of that.

## Proposed Approach: composition over nesting

Reuse a primitive that already exists — **a workflow is a unit of execution with its own scope, validation, cancellation, events, and logs.** Add two small things:

### 1. Workflow-level `loop_until`

A workflow can declare itself as a loop body:

```yaml
# .archon/workflows/test-fix.yaml
loop_until: \"\$test.exit_code == 0\"
max_iterations: 5
nodes:
  - id: test
    bash: |
      bun test
      echo \"exit_code=\$?\"
  - id: fix
    prompt: \"Tests failed. Fix: \$test.output\"
    depends_on: [test]
    when: \"\$test.exit_code != 0\"
```

[...]

PR #1707, opened 2026-05-17
fix(cli/workflow): add SIGTERM grace period to avoid losing completion to a race

jasonjack-jay:
When an external orchestrator watches the same state that a workflow's
final node modifies, it can race the DAG executor: the last node has
already done its side effect (e.g. moved an issue label out of an
'in-progress' set), the orchestrator sees the state change and SIGTERMs
the runner, and the SIGTERM lands in the small window between the bash
node returning and `completeWorkflowRun` being called.

Before this change, cleanup unconditionally marked the active run as
failed and exited 1. Downstream consumers saw exit-1 and retried
idempotent work that was already done, producing log noise and (in some
cases) duplicate side effects.

Cleanup now polls `getActiveWorkflowRun` every 100ms for up to 5s
before deciding what to do:

  - If the run finishes naturally during grace, look up its final status
    and exit 0 only if the DB says `completed` — otherwise exit 1.
  - If the run is still active after 5s, force-fail it as before
    (preserves the user-Ctrl+C semantics).
  - If we never see an active run during grace, preserve the previous
    exit-1 behaviour — we can't tell what happened.

5s is generous: `completeWorkflowRun` is typically <100ms. A genuine
user-driven cancel is delayed by at most 5s, which is the right trade
for not corrupting the success/failure signal.

Test mock for the workflows DB module is extended with
`getWorkflowRunStatus` so the new call site has a stub. All 94 existing
workflow.test.ts tests still pass.

This was observed in production with a harness that watches GitHub state
labels: every spike workflow exited 1 with the issue label correctly
moved and the PR correctly opened. The harness's defensive guard
prevented duplicate dispatches but the exit-1 noise made it harder to
distinguish real failures from this cosmetic race.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced workflow termination handling with a grace period for natural completion before forced shutdown. Exit codes now accurately reflect workflow completion status rather than always failing.

* **Tests**
  * Extended test mocks to support workflow status verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1707?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -133,6 +133,7 @@ mock.module('@archon/core/db/workflows', () => ({
   findResumableRun: mock(() => Promise.resolve(null)),
   resumeWorkflowRun: mock(() => Promise.resolve(null)),
   getWorkflowRun: mock(() => Promise.resolve(null)),
+  getWorkflowRunStatus: mock(() => Promise.resolve('completed')),
   updateWorkflowRun: mock(() => Promise.resolve()),
   listWorkflowRuns: mock(() => Promise.resolve([])),
   deleteOldWorkflowRuns: mock(() => Promise.resolve({ count: 0 })),
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -687,36 +687,77 @@ export async function workflowRunCommand(
     }
   })();
 
-  // Register cleanup handlers for graceful termination
+  // Register cleanup handlers for graceful termination.
+  //
+  // Cleanup runs with a brief grace period before forcing the in-flight workflow
+  // run to fail. This handles a real race we observed in the field:
+  //
+  //   A workflow's final node performs a state-changing side effect (e.g.
+  //   transitioning an issue label out of an "in-progress" set). An external
+  //   orchestrator watching that state sees the change and cancels the runner
+  //   by sending SIGTERM. At that moment the DAG executor has already finished
+  //   the last node but has not yet reached `completeWorkflowRun`. Without a
+  //   grace period, cleanup wins the race, marks the run as failed, and exits
+  //   1 — even though the work succeeded. Downstream consumers then treat the
+  //   exit code as a failure signal and retry idempotent work that is already
+  //   done, producing log noise and (in some cases) duplicate side effects.
+  //
+  // Grace strategy:
+  //   - Poll `getActiveWorkflowRun` every `POLL_MS` for up to `GRACE_MS`
+  //   - If the run finishes naturally during the grace window, look up its
+  //     final status and exit 0 only if the DB says `completed`
+  //   - If the run is still active after grace, force-fail it (preserving the
+  //     previous behaviour for genuine cancels like a user Ctrl+C)
+  //   - If we never see an active run during grace, exit 1 (preserves
+  //     previous behaviour — we don't know what happened)
+  //
+  // 5 seconds is generous: `completeWorkflowRun` typically runs in <100ms.
+  // A legitimate user-driven cancel is delayed by at most 5s, which is the
+  // right trade for not corrupting the success/failure signal.
+  const TERMINATION_GRACE_MS = 5000;
+  const TERMINATION_POLL_MS = 100;
   let terminating = false;
-  const cleanup = (signal: string): void => {
+  const cleanup = async (signal: string): Promise<void> => {
     if (terminating) return;
     terminating = true;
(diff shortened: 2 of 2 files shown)


Wirasm:
## Review Summary

**Verdict**: blocking-issues

Your grace-period polling strategy for the SIGTERM race condition is solid — the production backstory and 5-second calibration in the comments are genuinely valuable. However, there's a logic bug that makes the entire "completed during grace → exit 0" path unreachable, plus a missing test suite for the new cleanup handler.

### Blocking issues

- **`packages/cli/src/commands/workflow.ts:740`**: The condition `else if (lastSeenRunId && Date.now() - startedAt >= TERMINATION_GRACE_MS)` is always `true` when `lastSeenRunId` is set, because the `while` loop only breaks when the time condition becomes `false`. Completed runs that finish during the grace window will be force-failed instead of exiting 0. **Fix**: change to `else if (lastSeenRunId)` — the `if` branch above already handles the timeout case. See the code snippet in the full review.

- **`packages/cli/src/commands/workflow.ts:711-766`**: The `cleanup()` function is entirely untested. No test covers the signal registration, polling loop, or any of the three exit-code paths. Add tests for: (1) workflow finishes during grace → exit 0, (2) workflow still active after grace → `failWorkflowRun` called → exit 1, (3) no active run → exit 1, (4) `getWorkflowRunStatus` throws → exit 1 without crash. Consider exporting `TERMINATION_GRACE_MS` or injecting `graceMs` as a parameter so tests don't wait 5 seconds.

### Suggested fixes

- **`workflow.ts:727`** and **`workflow.ts:737`**: `getActiveWorkflowRun` and `getWorkflowRunStatus` silently swallow DB errors. Log the error before swallowing so operators can see DB issues during termination:
  ```ts
  workflowDb.getActiveWorkflowRun(conversation.id).catch(err => {
    getLog().error({ err, conversationId: conversation.id }, 'workflow.termination_poll_failed');
    return null;
  });
  ```

- **`workflow.ts:719`**: Add `getLog().info({ runId, status }, 'workflow.termination_grace_satisfied')` when the workflow exits 0 via the DB verdict, so logs distinguish success from the null-run path.

### Minor / nice-to-have

- **`workflow.ts:716`**: Comment says "Still active after grace" but the condition uses `>=`. At exactly 5000ms elapsed, the outer branch fires instead. Either change `>=` to `>` or update the comment to "Still active at or beyond grace boundary".

- **`workflow.ts:719-720`**: `void cleanup('SIGTERM')` / `void cleanup('SIGINT')` intentionally discard the Promise — a brief `// process exit imminent` comment prevents future readers from flagging this as a missing await.

### Compliments

The multi-paragraph block at `workflow.ts:698–726` explaining the race between `completeWorkflowRun` and external SIGTERM is exactly the kind of non-obvious WHY that justifies a comment. The "5 seconds is generous: `completeWorkflowRun` typically runs in <100ms" calibration is concrete and well-reasoned.

---

Wirasm:
@$jasonjack-jay related to #1720 — overlapping area or partial fix.

Wirasm:
@$jasonjack-jay related to #1469 — overlapping area or partial fix.

Wirasm:
@$jasonjack-jay related to #972 — overlapping area or partial fix.

Wirasm:
@$jasonjack-jay related to #1720 — overlapping area or partial fix.

Wirasm:
@$jasonjack-jay related to #1469 — overlapping area or partial fix.

Wirasm:
@$jasonjack-jay related to #972 — overlapping area or partial fix.

Wirasm:
Hi @jasonjack-jay — thanks for opening this PR.

- Validation Evidence
- Security Impact
- Human Verification
- Side Effects / Blast Radius
- Rollback Plan

Wirasm:
Thanks @jasonjack-jay — the production symptom you reported was real, and half of it got fixed at the root via #2191 (closing #1123): the SIGTERM/SIGINT handler was rewritten to be run-scoped with compare-and-set status transitions, which removed the mismarked-runs failure mode. That rewrite also deliberately deleted the conversation-wide run polling this branch is built on (it could fail runs owned by other processes), so the PR can't be rebased onto the new shape.

The other half of your report — the runner still exiting 1 even when its owned run completed — is genuinely unaddressed and is now tracked in #2233 with credit to you (flagged there as possibly-intentional, since SIGTERM→non-zero is conventional and `archon workflow get --json` gives the authoritative outcome). Appreciate the diagnosis.

PR #1714, opened 2026-05-18
fix(workflows): narrow findResumableRun by user_message on auto-resume

DHenion-cyber:
## Summary

Auto-resume from `executor.ts` previously matched only on `(workflow_name, working_path)`. For workflows with `worktree.enabled: false`, every invocation from the same checkout shares `cwd`, so a failed run for plan A is silently picked up by the next invocation passing plan B.

This bit a downstream user (CHIP400 LMS project) on `lms-phase-pipeline`: a Phase 4a sprint failed with `status='failed'`, then a kick for a completely different initiative (`module-pipeline-authoring-upgrade`) with a different plan path resumed Phase 4a's run, ran the wrong generator against the wrong worktree, and burned 6+ minutes of LLM time on contaminated state before failing.

## Fix

Add an optional third argument `userMessage?: string` to `findResumableRun`. When provided (called from `executor.ts:322` where the caller knows the current invocation's input), narrow the SELECT with `AND user_message = \$3`. When omitted (called from `cli/commands/workflow.ts:451` for `--resume`, where the user is explicitly asking to reuse whatever the prior run was), keep the original two-key match — preserving existing `--resume` semantics.

The CLI `--resume` site is untouched on purpose: `--resume` means "pick up the last failed run on this workflow at this path, whatever its input was." That's the right semantics there; it's only the implicit auto-resume during a normal `workflow run` invocation that needs the narrower match.

Also removes a pre-existing dead trailing `1` parameter from the params array — the SQL only ever referenced `\$1` and `\$2`.

## Diff scope

- \`packages/core/src/db/workflows.ts\` — function signature + SQL
- \`packages/workflows/src/store.ts\` — interface declaration
- \`packages/workflows/src/executor.ts\` — pass \`userMessage\` from the auto-resume path
- \`packages/core/src/db/workflows.test.ts\` — 3 new regression cases + update 2 existing param-equality assertions

## Test plan

- [x] \`bun test packages/core/src/db/workflows.test.ts\` → 69/69 pass (includes the 3 new cases)
- [x] \`bun test packages/workflows/src/executor.test.ts packages/workflows/src/executor-preamble.test.ts packages/cli/src/commands/workflow.test.ts\` → 131/131 pass (no caller-side regressions)
- [x] \`bun run type-check\` → all packages clean
- [x] Manual reproducer in a downstream repo: pre-fix \`lms-phase-pipeline\` invocations with different plan paths collide; post-fix they get distinct runs

## Notes

- Backward compatible: the new param is optional and existing callers compile unchanged.
- The CLI \`--resume\` path is intentionally unchanged. Explicit-resume semantics retain the original two-key match because the user is asking for "whatever ran last here, pick it up."
- Mock signatures in the workflow-deps test fixtures already accept any args, so no test-fixture updates were needed beyond the new assertions in \`workflows.test.ts\`.

## Related

The downstream user worked around this with a shell-level guard that abandons cross-plan resumable rows before \`archon workflow run\`. With this PR merged and a release cut, the workaround can be retired.

## Release Notes

* **Bug Fixes**
  * Improved workflow resumption to more accurately match previous runs by considering user context, preventing incorrect resumption of unrelated workflow variants in the same directory.

* **Tests**
  * Added comprehensive test coverage for workflow resumption scenarios with and without user message filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1714?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/db/workflows.test.ts
+++ b/packages/core/src/db/workflows.test.ts
@@ -490,7 +490,9 @@ describe('workflows database', () => {
       expect(query).not.toContain('conversation_id');
       expect(query).toContain('ORDER BY started_at DESC');
       expect(query).not.toMatch(/--.*\$\d/); // regression guard for #999: $N in SQL comments breaks convertPlaceholders
-      expect(params).toEqual(['feature-development', '/repo/path', 1]);
+      // userMessage omitted -> no $3 predicate, params length 2
+      expect(query).not.toContain('user_message = $3');
+      expect(params).toEqual(['feature-development', '/repo/path']);
     });
 
     test('returns a stale running run (no activity for >1 day)', async () => {
@@ -509,7 +511,7 @@ describe('workflows database', () => {
       const [query, params] = mockQuery.mock.calls[0] as [string, unknown[]];
       expect(query).toContain("status = 'running'");
       expect(query).toContain('last_activity_at');
-      expect(params).toEqual(['feature-development', '/repo/path', 1]);
+      expect(params).toEqual(['feature-development', '/repo/path']);
     });
 
     test('returns a running run with null last_activity_at (never recorded activity)', async () => {
@@ -542,6 +544,53 @@ describe('workflows database', () => {
       await expect(findResumableRun('test', '/path')).rejects.toThrow(
         'Failed to find resumable run: Connection refused'
       );
+    });
+
+    test('narrows by user_message when provided (auto-resume path)', async () => {
+      const failedRun = {
+        ...mockWorkflowRun,
+        status: 'failed' as const,
+        working_path: '/repo/path',
+        user_message: 'plan-A.md',
+      };
+      mockQuery.mockResolvedValueOnce(createQueryResult([failedRun]));
+
+      const result = await findResumableRun('feature-development', '/repo/path', 'plan-A.md');
+
+      expect(result).toEqual(failedRun);
+      const [query, params] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).toContain('user_message = $3');
+      expect(params).toEqual(['feature-development', '/repo/path', 'plan-A.md']);
+    });
+
+    test('does NOT cross-resume a failed run for a different user_message', async () => {
+      // Empty result means SQL filtered it out — the bug fix.
+      mockQuery.mockResolvedValueOnce(createQueryResult([]));
+
+      const result = await findResumableRun('feature-development', '/repo/path', 'plan-B.md');
+
+      expect(result).toBeNull();
+      const [query, params] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).toContain('user_message = $3');
+      expect(params).toEqual(['feature-development', '/repo/path', 'plan-B.md']);
+    });
+
+    test('explicit --resume (no user_message) falls back to workflow+path match', async () => {
+      // Caller is CLI --resume path; user is asking to reuse whatever the
+      // prior failed run was. user_message predicate must NOT be added.
+      const failedRun = {
+        ...mockWorkflowRun,
+        status: 'failed' as const,
+        working_path: '/repo/path',
+      };
+      mockQuery.mockResolvedValueOnce(createQueryResult([failedRun]));
+
+      const result = await findResumableRun('feature-development', '/repo/path');
+
+      expect(result).toEqual(failedRun);
+      const [query, params] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).not.toContain('user_message = $3');
+      expect(params).toEqual(['feature-development', '/repo/path']);
     });
   });
 
--- a/packages/core/src/db/workflows.ts
+++ b/packages/core/src/db/workflows.ts
@@ -305,21 +305,34 @@ export async function getRunningWorkflows(): Promise<
 
 export async function findResumableRun(
   workflowName: string,
-  workingPath: string
+  workingPath: string,
+  userMessage?: string
 ): Promise<WorkflowRun | null> {
   const dialect = getDialect();
+  // When userMessage is provided (auto-resume from executor.ts where the
+  // caller knows the current invocation's input), narrow the match so a
+  // failed run for a DIFFERENT plan/description on the same workflow+path
+  // is not silently picked up. When omitted (explicit `--resume` from the
+  // CLI, where the user is asking to reuse whatever the prior run was),
+  // fall back to the original workflow_name + working_path match.
+  const narrowByUserMessage = userMessage !== undefined;
+  const userMessagePredicate = narrowByUserMessage ? 'AND user_message = $3' : '';
+  const params: unknown[] = narrowByUserMessage
+    ? [workflowName, workingPath, userMessage]
+    : [workflowName, workingPath];
   try {
     const result = await pool.query<WorkflowRun>(
       `SELECT * FROM remote_agent_workflow_runs
        WHERE workflow_name = $1
          AND working_path = $2
+         ${userMessagePredicate}
          AND (
            status IN ('failed', 'paused')
            OR (status = 'running' AND (last_activity_at IS NULL OR last_activity_at < ${dialect.nowMinusDays(3)}))
          )
        ORDER BY started_at DESC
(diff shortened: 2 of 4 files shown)


Wirasm:
## Review Summary

**Verdict**: ready-to-merge

This PR fixes a subtle cross-resume bug where `findResumableRun` could pick up a failed run for a different plan/description on the same workflow+path. The fix threads `userMessage` into the auto-resume query while keeping explicit `--resume` on the broader fallback path. Implementation is clean, comments are excellent, and tests are solid.

### Blocking issues
None.

### Suggested fixes
- **packages/core/src/db/workflows.ts:305-338** — Consider adding an explicit test assertion for the ordering invariant. A test case that returns two rows with different `started_at` values (only the later one matching `userMessage`) and asserts the correct row is picked would document the intent of `ORDER BY started_at DESC LIMIT 1` rather than relying on its implicit semantics. This is a test-quality improvement, not a blocker.

### Minor / nice-to-have
- The test comment at `workflows.test.ts:492` is fine as-is. No change needed.
- The `null` vs `undefined` edge case for `userMessage` is not exercised in tests — the executor never passes `null`, and `null` and `undefined` are both falsy so behavior is identical. Informational only; no action required.

### Compliments
- The multi-paragraph comment at `executor.ts:315-320` explaining the auto-resume vs. `--resume` invariant is the best kind of documentation: it captures a subtle runtime behavior that cannot be inferred from the code alone. The reference to `worktree.enabled: false` as a concrete domain anchor is especially good — it won't go stale.
- The regression suite in `workflows.test.ts` covers the happy path, cross-resume protection, and the explicit `--resume` fallback — exactly the three scenarios a future maintainer would need to understand.

---
*Reviewed via maintainer-review-pr workflow (Pi/Minimax). Aspects run: code-review, test-coverage, comment-quality, docs-impact.*

Wirasm:
@$DHenion-cyber related to #1469 — overlapping area or partial fix.

Wirasm:
@$DHenion-cyber related to #1720 — overlapping area or partial fix.

Wirasm:
@$DHenion-cyber related to #972 — overlapping area or partial fix.

Wirasm:
@$DHenion-cyber related to #1469 — overlapping area or partial fix.

Wirasm:
@$DHenion-cyber related to #1720 — overlapping area or partial fix.

Wirasm:
@$DHenion-cyber related to #972 — overlapping area or partial fix.

Wirasm:
Hi @DHenion-cyber — thanks for opening this PR.

- Validation Evidence
- Security Impact
- Human Verification
- Side Effects / Blast Radius
- Rollback Plan

Wirasm:
## Review Summary

**Verdict**: blocking-issues

Thanks for the focused fix. The intent is right — narrowing auto-resume by `user_message` while keeping explicit `--resume` broad — and the new SQL-level regression tests cover the predicate shape. Unfortunately the implementation contains a **runtime SQL bug** that breaks *both* resume paths on first DB lookup, and three of the four reviewers here (myself included on the error-handling side) initially missed it because we trusted the PR description's claim that the trailing `1` in the old params array was dead. It wasn't — it was the integer bound to `$3` for `dialect.nowMinusDays(3)`. Verified against a real SQLite DB: the old code returned rows, both new forms raise "expected N values, received M". Your tests still pass only because every test mocks `pool.query` and never executes the SQL.

### Blocking issues

- **`packages/core/src/db/workflows.ts:308-340` — `$3` placeholder collision (runtime, both paths broken)**
  - The CLI path now has `params = [workflowName, workingPath]` (length 2) but the SQL still references `$3` via `dialect.nowMinusDays(3)`.
  - The auto-resume path binds `userMessage` as `$3` while `nowMinusDays(3)` also points at `$3` — Postgres errors with `invalid input syntax for type interval` (`'plan-A.md' || ' days'` isn't an integer), SQLite errors with `expected 4 values, received 3`.
  - **Fix**:
    ```ts
    const daysPlaceholder = narrowByUserMessage ? 4 : 3;
    const params: unknown[] = narrowByUserMessage
      ? [workflowName, workingPath, userMessage, 3]
      : [workflowName, workingPath, 3];
    // ... AND last_activity_at < NOW() - (${dialect.nowMinusDays(daysPlaceholder)} || ' days')::INTERVAL
    ```
  - Then update `workflows.test.ts` lines 492 / 562 / 575 to expect length 3 (CLI) and length 4 (executor) respectively.
  - Also add one durable regression guard: scan the rendered SQL for `$N` placeholders and assert `params.length >= max($N)`. That single assertion is what would have caught this.

### Suggested fixes

- **`packages/workflows/src/executor.ts:322` — your `userMessage` passthrough is untested.** Every test in `executor.test.ts` mocks `findResumableRun` with arg-agnostic stubs, so a future edit reverting line 322 to two args would silently re-open the bug while every test still passes. Add one assertion:
  ```ts
  expect(findSpy.mock.calls[0]).toEqual(['feature-development', '/tmp', 'plan-A.md']);
  ```

- **`packages/cli/src/commands/workflow.ts:451` — explicit `--resume` two-arg call has no regression test.** The PR deliberately keeps this two-arg; please add a test asserting `findResumableRun` is called with exactly `[workflowName, cwd]` so a future "helpful" refactor can't silently change explicit-resume semantics.

- **`packages/docs-web/src/content/docs/guides/authoring-workflows.md:524-540` — "DAG Resume on Failure" section is now stale.** It still says auto-resume matches on workflow + working path. After this fix it also requires an exact `user_message` match. Add one sentence calling out the shared-`cwd` / `worktree.enabled: false` case (where the original bug manifested) and one noting that explicit `--resume` intentionally retains broader matching.

- **`packages/core/src/db/workflows.test.ts:567` — comment is factually wrong.** `// Empty result means SQL filtered it out — the bug fix.` sits directly above `mockQuery.mockResolvedValueOnce(createQueryResult([]))`. The mock is *configured* to return empty; SQL is never executed. Please delete the comment.

### Minor / nice-to-have

- `packages/core/src/db/workflows.test.ts:493` — `// userMessage omitted -> no $3 predicate, params length 2` restates the next two assertions; delete.
- `packages/core/src/db/workflows.test.ts:579-580` — the CLI `--resume` rationale comment couples a DB unit test to the CLI flag; delete or replace with `// Omitting userMessage intentionally preserves broad matching.`
- `packages/core/src/db/workflows.ts:312-317` — six-line block references `executor.ts`, the CLI, and "original" behavior; replace with `// Undefined intentionally preserves broad matching for explicit resume requests.`
- `packages/workflows/src/executor.ts:318-321` — condense to `// Include invocation input because non-isolated runs share the same working path.`
- `packages/cli/src/commands/workflow.ts:431` — append `(executor narrows by user_message; the wider two-key match used here is intentional for explicit --resume.)` to flag the asymmetry.
- `packages/core/src/db/adapters/sqlite.test.ts` — add at least one `findResumableRun` test against a real `SqliteAdapter`; the same `$3` collision bug manifests in both dialects.
- Pre-existing, out of scope for this PR: `workflow_resume_check_failed` event name in `executor.ts:329` doesn't follow `{domain}.{action}_{state}`; the rethrown error at `workflows.ts:401` drops `workflowName` from the user-visible message.

### Compliments

The asymmetric semantics — narrow on auto-resume, wide on explicit `--resume` — is exactly the right design call here, and the propagation of the optional parameter through the `IWorkflowStore` interface, the implementation, and the call site with consistent optional semantics is clean. Once the `$3` placeholder issue is fixed and one real-DB regression guard is added, this is a tidy bug fix.

---

Wirasm:
Closing as superseded — with thanks: your diagnosis of the resume-contamination bug was correct, and dev ultimately fixed it with a stronger mechanism than key-narrowing. PR #1646 removed implicit auto-resume entirely (the executor never calls findResumableRun anymore; resume is now an explicit opt-in at the call sites), which eliminates the cross-invocation contamination this PR targeted. The only surviving findResumableRun caller is the explicit CLI --resume path, which your PR deliberately kept broad — so a rebase would land dead code. The review's durable-guard idea (real-DB placeholder-arity test) lives on as #2196.

PR #1718, opened 2026-05-18
fix(workflows): write large node outputs to temp file to prevent bash substitution corruption (fixes #1717)

### LINKED ISSUE #1717: bash node: $nodeId.output substitution silently corrupts large multi-KB inputs (~42KB+)
## Summary

- What broke: When a bash node references `$nodeId.output` from an upstream LLM node whose output is large (~40KB+), the substituted value reaching the bash subprocess is corrupted. The downstream consumer fails with a parse / shape error even though the consumer handles the same input correctly when invoked directly with the same stdin.
- When it started (if known): pre-existing; surfaces consistently on the `maintainer-standup` workflow whose `persist` node consumes the `synthesize` node's brief + state JSON. The synth output grew past ~30-40KB as the project's PR/issue count grew, and the bash node started failing reliably.
- Severity: `major`

## Steps to Reproduce

1. Run `bun run cli workflow run maintainer-standup` on a repo with enough activity that the synth output is >30KB. (Today's run: 42KB synth output.)
2. The `synthesize` node completes successfully and emits a valid brief + state JSON between `ARCHON_STATE_JSON_BEGIN`/`END` markers.
3. The `persist` bash node fails with `Bash node 'persist' failed [exit 1]: ...` and the run is reported as failed.
4. Save the *exact* synth output the persist node should have received (extract from the run log between the `# Maintainer Standup` heading and the first `ARCHON_STATE_JSON_END` marker).
5. Run the persist script manually on that input: `cat /tmp/synth-output.txt | bun .archon/scripts/maintainer-standup-persist.ts`
6. **It succeeds.** Brief + state are written correctly. The persist script handled the input fine; the bash node never delivered the same input to it.

## Expected vs Actual

- **Expected**: `$synthesize.output` substitution in a bash node delivers the upstream output as-is to the subprocess's environment / stdin / wherever the bash code routes it. A consumer that works on the same string read from stdin manually should work the same when invoked via the bash node.
- **Actual**: For large outputs (~42KB confirmed; threshold unknown), the bash node's substitution silently corrupts the value en route to the subprocess. The downstream consumer sees malformed input.

## Affected node

```yaml
- id: persist
  depends_on: [synthesize]
  timeout: 30000
  bash: |
    set -uo pipefail
    RAW=$synthesize.output
    printf '%s' "$RAW" | bun .archon/scripts/maintainer-standup-persist.ts
```

The framework substitutes `$synthesize.output` into the bash code as a literal before bash interprets `RAW=...`. For multi-line, multi-KB strings the `RAW=` assignment without surrounding quotes (`RAW="..."`) word-splits and glob-expands the substituted text. With special characters (the synth output contains markdown, embedded JSON braces, parentheses, asterisks, `$` signs in narrative, etc.) the assignment can fail or produce a truncated value.

Even if `RAW="$synthesize.output"` (with quotes) is the right pattern, the substitution layer needs to escape any embedded `"` in the upstream output or the closing quote ends early.

## User Flow

```
Workflow author             Archon                          bash subprocess
───────────────             ──────                          ───────────────
defines bash node ────────▶ schedule node
                            substitute $synthesize.output ─▶ `RAW=<42KB of mixed text>`
                                                            ◀── bash parse error or
                                                                truncated/mangled $RAW
                            [X] persist consumer sees broken input → exit 1
sees workflow failure   ◀── DAG node failed
```

## Environment

- Platform: CLI (also reproducible from web)
- Database: SQLite
- Running in worktree? No (`worktree.enabled: false`)
- OS: macOS / Linux (shell behaviour is the issue, not OS-specific)

## Logs

```
{"level":50,"module":"workflow.dag-executor","exitCode":1,"killed":false,
 "stderrTail":"...truncated state JSON...ARCHON_STATE_JSON_END\n--- END raw output ---",
 "nodeId":"persist","nodeType":"bash","isTimeout":false,"msg":"dag_node_failed"}
[persist] Failed: Bash node 'persist' failed [exit 1]: ...
```

The script's own `PERSIST FAILED` stderr is what's quoted in the log — meaning the persist script DID run, but received corrupted stdin (or no stdin at all, with the bash code itself crashing before the pipe). Either failure mode is the same root cause: the upstream substitution didn't deliver the value cleanly.

## Impact

- Affected workflows/commands: any workflow where a bash node consumes a multi-KB LLM node's output via `$nodeId.output` substitution. `maintainer-standup` is the most prominent — fails on every run as the synth output grows. Also affects any custom workflow that pipes a large prompt result into a script via bash.
- Reproduction rate: Reliable once the source output exceeds the threshold (today: 42KB → fail; earlier runs at ~20KB → pass).
- Workaround: Extract the synth output from the run log manually and pipe it directly into the persist script. Tedious but possible because the data IS in the log.
- Data loss risk: No (raw run output is preserved on disk).

## Scope

- Module: `workflows:dag-executor` (bash node execution path), `workflows:executor-shared` (substituteWorkflowVariables / substituteNodeOutputRefs)

## Proposed direction

Two possible fixes, either or both:

1. **Pass `$nodeId.output` via subprocess env vars, not via bash code substitution.** The existing `shellSafe: true` plumbing from #1651 already does this for user-controlled variables (`$USER_MESSAGE`, `$ARGUMENTS`, `$LOOP_USER_INPUT`, ...) — extend it to cover `$nodeId.output` references when the consuming node is `bash:`. The bash code would then reference `$SYNTHESIZE_OUTPUT` from env, where Archon set it via the subprocess `env:` block. No shell-quoting issue, no size limit beyond `ARG_MAX`.

2. **Quote the substitution as a single shell-safe argument.** When substituting `$nodeId.output` into a bash node's code, wrap it as a single-quoted shell string with `'` escaped — i.e. emit `RAW=$(printf '%s' '$SYNTH_OUTPUT_PLACEHOLDER')` where the placeholder is the substituted text with embedded `'` escaped via `'"'"'`. Heavy-handed but correct for arbitrary text content.

Option 1 aligns with the precedent already set by #1651 and is the cleaner long-term direction. Option 2 is a backstop that works without changing the substitution contract.

A regression test should exercise a bash node consuming a `$nodeId.output` of at least 50KB of mixed-content text (markdown + embedded JSON + special chars: `$`, `"`, `'`, backticks, parens) and assert the bash subprocess receives the exact bytes.

## Reproducer artifact

Today's exact failing synth output (42,695 bytes) is at `/Users/rasmus/.claude/projects/-Users-rasmus-Projects-cole-Archon/c8052aa7-3a9a-46e1-9516-c8abd374f716/tool-results/bxccgh2b7.txt` locally — happy to attach as a fixture if useful for the test suite.

kagura-agent:
## Summary

- **Problem:** When a bash node references `$nodeId.output` from an upstream LLM node whose output is large (~42KB+), the substituted value is silently corrupted because the full text is inlined as a `bash -c` argument, hitting process argument passing limits.
- **Why it matters:** The `maintainer-standup` workflow (and any workflow with large intermediate outputs) fails silently — the downstream consumer gets garbled data and raises parse errors, despite working fine with the same input directly.
- **What changed:** Added a size threshold (`NODE_OUTPUT_FILE_THRESHOLD = 32KB`). Outputs below this continue to be shell-quoted inline (existing behavior). Outputs at or above this are written to a temp file in `logDir` and substituted with `$(cat '<path>')` — bash reads the value at runtime via command substitution, bypassing the argv size issue entirely.
- **What did NOT change:** Non-bash substitution paths (AI prompts, command nodes) are unaffected — they pass `escapedForBash=false` and never hit the file path. Small outputs (<32KB) still inline as before.

### Before

```
Workflow              DAG Executor           bash -c
────────              ────────────           ───────
runs synthesize ───▶  captures 42KB output
runs persist node     substituteNodeOutputRefs
                      shellQuote(42KB) ─────▶ bash -c '<42KB inline>'
                                              ❌ data corrupted silently
                                              downstream parse fails
```

### After

```
Workflow              DAG Executor           bash -c
────────              ────────────           ───────
runs synthesize ───▶  captures 42KB output
runs persist node     substituteNodeOutputRefs
                      output > 32KB → write to logDir/synthesize.nodeoutput
                      substitute with $(cat path) ─▶ bash -c '...$(cat /path)...'
                                                      ✅ cat reads full file at runtime
                                                      downstream parse succeeds
```

### Before

```
substituteNodeOutputRefs(prompt, nodeOutputs, escapedForBash)
      │
      └─▶ shellQuote(nodeOutput.output)  ← always inline, regardless of size
            │
            └─▶ embedded in bash -c argv
```

### After

```
substituteNodeOutputRefs(prompt, nodeOutputs, escapedForBash, outputFileDir?)
      │
      ├─▶ output < 32KB: shellQuote(value)  ← existing behavior
      │
      └─▶ output >= 32KB AND outputFileDir:
            shellQuoteOrFile(value, nodeId, field, dir) [+]
              │
              ├─▶ writeFileSync(logDir/nodeId.nodeoutput, value)
              └─▶ returns $(cat '<path>')
```

**Connection inventory:**

|------|-----|--------|-------|
| executeBashNode | substituteNodeOutputRefs | [~] passes logDir as 4th arg |
| executeLoopNode (until_bash) | substituteNodeOutputRefs | [~] passes logDir as 4th arg |
| substituteNodeOutputRefs | shellQuoteOrFile | [+] new helper for size-aware quoting |
| shellQuoteOrFile | writeFileSync + shellQuote | [+] writes file when threshold exceeded |
| Other callers (AI nodes, approval, cancel) | substituteNodeOutputRefs | unchanged (no outputFileDir) |

## Validation Evidence

- `bun test packages/workflows/src/dag-executor.test.ts` — **236 pass, 0 fail**
- `bun run type-check` — all 5 packages clean
- lint-staged passed on commit (eslint + prettier)
- New tests cover: small output still inlines, large output writes to file, field access with large value, non-bash paths unaffected

## Security Impact

- Temp files are written to the existing `logDir` (already under the run's artifact directory) — no new paths introduced
- Files contain workflow intermediate outputs (same data that was previously in argv) — no escalation of access
- `shellQuote` is used on the file path in the `$(cat ...)` command to prevent path injection

## Compatibility/Migration

- Fully backward compatible: existing small outputs behave identically
- No config changes or environment variables needed
- Temp files use `.nodeoutput` extension and are scoped to the run's logDir

## Risks and Mitigations

| Risk | Mitigation |
|------|-----------|
| Disk write failure | writeFileSync will throw, caught by existing executeBashNode try/catch → node fails with clear error |
| File not cleaned up | Files live in logDir which is already managed by run lifecycle cleanup |
| $(cat) timing | File is written synchronously before exec, guaranteed to exist when bash reads it |

## Rollback Plan

Revert the commit — the only behavioral change is the file-based path for >32KB outputs.

Closes #1717

## Summary by CodeRabbit

* **New Features**
  * Bash-based workflow steps now handle very large node outputs by writing them to files and referencing them, preventing oversized inline values from breaking scripts.

* **Bug Fixes**
  * Improved fallback behavior when file-based substitution fails — shell-quoting is used to keep scripts robust.
  * Option to force full inlining preserved for cases that require it.

* **Tests**
  * Added tests covering small vs. large output handling, structured-field file naming, no-file inlining option, and file-write failure fallback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1718?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -828,6 +828,56 @@ describe('substituteNodeOutputRefs -- shell escaping', () => {
   });
 });
 
+describe('substituteNodeOutputRefs -- large output file substitution', () => {
+  let tempDir: string;
+
+  beforeEach(async () => {
+    tempDir = join(tmpdir(), `archon-test-large-output-${Date.now()}`);
+    await mkdir(tempDir, { recursive: true });
+  });
+
+  afterEach(async () => {
+    await rm(tempDir, { recursive: true, force: true });
+  });
+
+  it('inlines small output even when outputFileDir is provided', () => {
+    const outputs = new Map([['a', makeOutput('completed', 'small')]]);
+    const result = substituteNodeOutputRefs('echo $a.output', outputs, true, tempDir);
+    expect(result).toBe("echo 'small'");
+  });
+
+  it('writes large output (>=32KB) to file and returns $(cat ...) reference', async () => {
+    const largeOutput = 'x'.repeat(33_000);
+    const outputs = new Map([['a', makeOutput('completed', largeOutput)]]);
+    const result = substituteNodeOutputRefs('echo $a.output', outputs, true, tempDir);
+    expect(result).toContain('$(cat ');
+    expect(result).toContain('a.nodeoutput');
+    // Verify file was written with correct content
+    const { readFile: readFileAsync } = await import('fs/promises');
+    const written = await readFileAsync(join(tempDir, 'a.nodeoutput'), 'utf-8');
+    expect(written).toBe(largeOutput);
+  });
+
+  it('writes large field value to file with field name in filename', async () => {
+    const largeValue = 'y'.repeat(33_000);
+    const outputs = new Map([['a', makeOutput('completed', JSON.stringify({ data: largeValue }))]]);
+    const result = substituteNodeOutputRefs('echo $a.output.data', outputs, true, tempDir);
+    expect(result).toContain('$(cat ');
+    expect(result).toContain('a.data.nodeoutput');
+    const { readFile: readFileAsync } = await import('fs/promises');
+    const written = await readFileAsync(join(tempDir, 'a.data.nodeoutput'), 'utf-8');
+    expect(written).toBe(largeValue);
+  });
+
+  it('does not write to file when escapedForBash=false even for large output', () => {
+    const largeOutput = 'x'.repeat(33_000);
+    const outputs = new Map([['a', makeOutput('completed', largeOutput)]]);
+    const result = substituteNodeOutputRefs('echo $a.output', outputs, false, tempDir);
+    expect(result).toBe(`echo ${largeOutput}`);
+    expect(result).not.toContain('$(cat ');
+  });
+});
+
 describe('substituteNodeOutputRefs -- structuredOutput preference', () => {
   it('prefers structuredOutput.field over JSON.parse(output)', () => {
     // Pi-shape: prose output text with structuredOutput populated by tryParseStructuredOutput.
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -876,6 +876,17 @@ describe('substituteNodeOutputRefs -- large output file substitution', () => {
     expect(result).toBe(`echo ${largeOutput}`);
     expect(result).not.toContain('$(cat ');
   });
+
+  it('falls back to shell-quoting when file write fails', () => {
+    const largeOutput = 'x'.repeat(33_000);
+    const outputs = new Map([['a', makeOutput('completed', largeOutput)]]);
+    // Use a non-existent directory to trigger writeFileSync failure
+    const badDir = '/nonexistent-path-that-does-not-exist';
+    const result = substituteNodeOutputRefs('echo $a.output', outputs, true, badDir);
+    // Should fall back to inline shell-quoting instead of crashing
+    expect(result).not.toContain('$(cat ');
+    expect(result).toBe(`echo '${largeOutput}'`);
+  });
 });
 
 describe('substituteNodeOutputRefs -- structuredOutput preference', () => {
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -247,8 +247,17 @@ function shellQuoteOrFile(
   if (outputFileDir && value.length > NODE_OUTPUT_FILE_THRESHOLD) {
     const filename = field ? `${nodeId}.${field}.nodeoutput` : `${nodeId}.nodeoutput`;
     const filePath = joinPath(outputFileDir, filename);
-    writeFileSync(filePath, value);
-    return `$(cat ${shellQuote(filePath)})`;
+    try {
+      writeFileSync(filePath, value);
+      return `$(cat ${shellQuote(filePath)})`;
+    } catch (fileErr) {
+      const err = fileErr as Error;
+      getLog().error(
+        { err, nodeId, field, valueSize: value.length, filePath },
+        'dag.large_output_file_write_failed'
+      );
+      return shellQuote(value); // fallback: inline (pre-file-spill behavior)
+    }
   }
   return shellQuote(value);
 }
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -5,8 +5,9 @@
  * Independent nodes within the same layer run concurrently via Promise.allSettled.
  * Captures all assistant output regardless of streaming mode for $node_id.output substitution.
  */
+import { writeFileSync } from 'fs';
 import { readFile } from 'fs/promises';
-import { isAbsolute, resolve as resolvePath } from 'path';
+import { isAbsolute, join as joinPath, resolve as resolvePath } from 'path';
 import { execFileAsync } from '@archon/git';
 import { discoverScriptsForCwd } from './script-discovery';
 import type {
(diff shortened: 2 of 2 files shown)


Wirasm:
## Review Summary

**Verdict**: blocking-issues

Your change adds a file-spill mechanism for large node outputs (≥32KB) in bash scripts, preventing silent data corruption when shell argument limits are exceeded. The implementation is well-structured and the threshold rationale is clearly documented. One critical bug and two items need to be addressed before merge.

### Blocking issues

- **`packages/workflows/src/dag-executor.ts:247–249`** (`shellQuoteOrFile`): `writeFileSync` is not wrapped in a try/catch. If the disk is full, permissions are wrong, or any other filesystem error occurs, the throw propagates as an unhandled exception — no structured log, no user-facing message, just a crash with an opaque Node.js stack trace. Wrap it and fall back to shell-quoting on failure:

  ```typescript
  try {
    writeFileSync(filePath, value);
    return `$(cat ${shellQuote(filePath)})`;
  } catch (fileErr) {
    const err = fileErr as Error;
    getLog().error({ err, nodeId, field, valueSize: value.length, filePath }, 'dag.large_output_file_write_failed');
    return shellQuote(value); // fallback: inline (pre-PR behavior)
  }
  ```

### Suggested fixes

- **`packages/workflows/src/dag-executor.ts` (test presence)**: The new `describe('substituteNodeOutputRefs -- large output file substitution')` block appears in the diff but could not be confirmed in the actual test file. Please ensure those tests are committed so they run in CI.

- **`packages/workflows/src/dag-executor.ts` (return type)**: `substituteNodeOutputRefs` is missing an explicit return type annotation. Per CLAUDE.md, add `: string`.

- **`packages/workflows/src/dag-executor.ts:244–248` (JSDoc)**: Add a doc comment for the new `outputFileDir` parameter so future maintainers understand why it's intentionally omitted for non-bash call sites (prompt/approval/cancel nodes).

### Minor / nice-to-have

- **JSON.parse fallback not tested with large payloads** (`packages/workflows/src/dag-executor.ts:300–318`): The large-value path is tested via `structuredOutput` but not via raw JSON output. Consider adding a test that passes `makeOutput('completed', JSON.stringify({ data: 'x'.repeat(33_000) }))` with no `structuredOutput` field.
- **`shellQuoteOrFile` closing over constant**: `NODE_OUTPUT_FILE_THRESHOLD` is read from outer scope. Low priority — fine as-is for now.
- **Test style** (`dag-executor.test.ts`): new tests use sync `it()` with `await import(...)`. Cosmetic; not blocking.

### Compliments

- The `NODE_OUTPUT_FILE_THRESHOLD` comment (line 1267–1270) explains the *why* (silent corruption risk) rather than just stating the number — exactly the kind of context that helps future maintainers.
- The `substituteNodeOutputRefs` execution-ordering comment (*"Called AFTER the standard substituteWorkflowVariables pass"*) documents a non-obvious dependency that would be very hard to debug if violated.
- The design decision to make file-spill opt-in via `outputFileDir` and only wire it for bash-context nodes is sound and avoids over-engineering for non-bash call sites.

---

kagura-agent:
Thanks for the thorough review @Wirasm! Addressed the blocking issue:

**writeFileSync try/catch** — Wrapped in try/catch with structured logging via `getLog().error()` and fallback to inline shell-quoting (pre-file-spill behavior). Added a test that verifies the fallback path by using a non-existent directory.

Also confirmed the existing test suite (all tests in `dag-executor.test.ts`) passes cleanly — 237 tests, 0 failures.

Re the suggested fixes:
- **Tests are committed** and running in CI ✅
- **Return type annotation** — `substituteNodeOutputRefs` already has `: string` return type ✅
- **JSDoc for outputFileDir** — good idea, happy to add if you'd like

PR #1730, opened 2026-05-20
feat(workflows): add always_run node opt-out for resume caching (closes #1391)

### LINKED ISSUE #1391: Auto-resume cache traps workflows on "successful" producer with bad output — only escape is renaming the node
## Problem

When a workflow run is auto-resumed, every node previously marked \`completed\` is skipped via the \`prior_success\` path and its output is replayed from cache to downstream consumers. There's no escape hatch when a producer node exited 0 but its output is structurally bad (corrupted file, malformed JSON written to disk, missing fields). The downstream consumer keeps failing forever, and the only way to clear the bad cache is to **rename the producer node** so its identity changes.

## Repro

1. Build a workflow with a producer (`bash:` or `script:`) that writes a file to `$ARTIFACTS_DIR` and exits 0, plus a consumer (`script:`) that reads that file.
2. Author the producer with a subtle bash bug — e.g. \`printf '%s' '\$node.output'\` (the `'…'` collide with the auto-shell-quoter, the JSON gets bash-brace-expanded into garbage). Producer exits 0; file is corrupted.
3. Consumer fails parsing the file.
4. Fix the producer.
5. Re-run \`archon workflow run <name> ...\` — observe \`workflow.dag_resuming\`, \`priorCompletedCount: N\`, producer \`Skipped (prior_success)\`, consumer fails again with the **same** corrupted file.
6. Repeat — fix doesn't take effect because producer is forever cached.
7. Workaround: rename producer node \`foo\` → \`foo-v2\`, update all `depends_on` references, update all `$foo.output` references downstream. Cache miss on the new id, producer re-runs, fix takes effect.

This happened to me building a release workflow. The bridge node (`persist-draft-changelog`) wrote corrupted JSON because of an unrelated quoting bug, but the bash command exited 0. Three subsequent \"fix and re-run\" cycles all failed at the consumer because the cached corrupted file kept being used. Discovered the workaround by renaming to `save-draft-json`.

## Why this is the worst class of bug

\"Producer exit 0 + bad output\" is structurally the hardest workflow failure mode to debug:

- The producer's log says \"Completed (10ms)\" — looks fine
- The consumer's error points at the consumer code (\"JSON parse error\") — misleading
- The bad artifact persists across resumes — every iteration sees the same broken downstream
- The `--resume`-cached behavior actively hides the fix
- Renaming a node and chasing every reference is exactly the kind of yak-shave that breaks an authoring flow

For agents authoring workflows, this trap is catastrophic — they don't know to rename the node, and their fix-the-producer iterations all look like the producer is fine and the consumer is broken.

## Proposed direction (sketches)

Any of these would help; not prescriptive:

1. **`archon workflow node invalidate <run-id> <node-id>`** — CLI escape hatch. Drops the cached output for one node; on next resume it re-runs.
2. **`archon workflow node invalidate <run-id> --downstream-of <node-id>`** — invalidate everything that consumed `<node-id>`'s output. Useful when the producer was OK but you changed its consumer's contract.
3. **`--no-resume` flag on `archon workflow run`** — force fresh run, ignore prior cached state. Currently the *only* way to do this is `archon workflow abandon <id>` followed by `archon workflow run` (and abandon errors with \"already terminal\" if the run is already failed — see issue F).
4. **`always_run: true` on a node** — opts out of resume caching for that node only. Makes producers like \"persist AI output to disk\" idempotent-by-policy.
5. **Make node identity content-addressable** — hash the node's resolved script body + inputs into the cache key. A code change to the producer auto-busts the cache. (Heavier change, but solves the root cause.)

## Adjacent

- See [#1389](https://github.com/coleam00/Archon/issues/1389) — \"errors should help agents self-correct\". This issue is the same family: an agent debugging a failing workflow gets pointed at the wrong problem.
- See [issue F: auto-resume default behavior] — the resume-by-default behavior amplifies this trap because authors don't realize they're resuming.

## Found while

Building a workshop `archon-release` workflow. The workflow itself is sound; this Archon behavior cost ~3 iteration cycles before I figured out the cache-invalidation trick.

truffle-dev:
## Summary

- **Problem:** When a DAG workflow auto-resumes, every node that completed successfully in the prior run is skipped, and downstream consumers read the cached output. That is wrong when a node's exit code does not capture output validity — typically a producer that writes a file or side-effect the consumer parses on each run. A bash producer that exits 0 but writes garbage stays "successfully cached" across every resume, with no escape short of renaming the node.
- **Why it matters:** Issue #1391 documents the trap: there is no per-node knob to say "this node is a side-effecting producer whose success status does not validate its output, re-run it on resume." Users hit it with download scripts, code generators, and any bash node whose downstream consumer parses the file.
- **What changed:** New optional `always_run: boolean` field on every DAG node. When `true`, the node re-executes on resume even when `priorCompletedNodes` says it completed — both the resume pre-populate and the per-node skip-check honor the flag.
- **What did NOT change:** Default behavior is unchanged (skip on prior success). Normal cached nodes in the same run still skip. The flag is per-node and opt-in. No schema migration, no DB change, no impact on fresh-run scheduling.

### Before

```
Workflow author          Resume                   Producer node
───────────────          ──────                   ─────────────
bash node writes file ─▶ priorCompletedNodes      first run: writes garbage, exits 0
consumer parses file     has producer.id          ↓
                         ↓                        marked completed
                         skips producer           ↓
                         feeds cached output ───▶ next resume: skipped FOREVER
                                                  ❌ consumer reads stale file
                                                  only escape: rename the node
```

### After

```
Workflow author          Resume                   Producer node
───────────────          ──────                   ─────────────
bash producer node ────▶ priorCompletedNodes      always_run: true on producer
  always_run: true       has producer.id          ↓
consumer parses file     ↓                        always re-executes on resume
                         pre-populate skips it    ↓
                         skip-check skips it ──▶  writes fresh file
                                                  ✅ consumer reads fresh output
                         normal nodes still skip
```

### Before

```
executeDagWorkflow
  │
  ├─▶ buildTopologicalLayers
  │
  ├─▶ priorCompletedNodes pre-populated into nodeOutputs (all completed nodes)
  │
  └─▶ per-node executor
        └─ if priorCompletedNodes.has(node.id) → skip, emit node_skipped_prior_success
```

### After

```
executeDagWorkflow
  │
  ├─▶ buildTopologicalLayers
  │
  ├─▶ priorCompletedNodes pre-populated [~] — filters out always_run nodes
  │
  └─▶ per-node executor
        ├─ if priorCompletedNodes.has(node.id) AND node.always_run [+]
        │    → emit dag.node_always_run_resume_forced, fall through to normal execution
        └─ if priorCompletedNodes.has(node.id) AND NOT node.always_run
             → skip, emit node_skipped_prior_success (existing behavior)
```

**Connection inventory:**

|------|-----|--------|-------|
| dagNodeBaseSchema | DagNode types | [~] adds `always_run?: boolean` to all variants |
| dag-node transform | base spread | [~] passes through `always_run` when defined |
| executeDagWorkflow pre-populate | nodeOutputs | [~] excludes always_run nodes from pre-populate |
| per-node skip-check | always_run gate | [+] new log event when always_run forces re-run |
| consumers reading $producer.output | fresh-run path | unchanged — downstream sees re-run output via normal layer ordering |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `workflows`
- Module: `workflows:dag-executor`, `workflows:schemas`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Closes #1391

## Validation Evidence

```bash
bun run type-check    # all 10 packages clean
bun run lint          # 0 errors, 0 warnings
bun run format:check  # all files use Prettier code style
bun run test          # full workflows suite green, 3 new dag-executor tests + 1 new loader test
```

New tests cover:

- always_run node re-runs even when present in `priorCompletedNodes` (no `node_skipped_prior_success` event)
- mixed run: always_run re-executes; sibling normal node still skips
- downstream consumer reads fresh producer output, not the pre-populated cached value
- schema parses `always_run: true` and leaves it undefined when omitted

## Security Impact

## Compatibility / Migration

- Backward compatible? Yes — field is optional, defaults to off, no existing workflow YAML needs to change.
- Config/env changes? No

## Human Verification

- Verified the schema accepts `always_run: true` on a real two-node YAML (`fetch-data` bash producer + `process-data` consumer); the parsed AST exposes the field; consumer reads fresh output post-resume.
- Confirmed default behavior unchanged by running the existing dag-executor resume tests (138 pass, 0 fail) alongside the new always_run tests.
- Sanity-checked that the pre-populate exclusion does not strand downstream consumers: when an always_run producer is in a later layer, the consumer still resolves through the normal post-execution `nodeOutputs.set(...)` write.

What was not verified beyond CI:

- End-to-end resume run via the CLI (`archon workflow run ... --resume`) was not exercised — the dag-executor unit tests directly drive the same code path via `priorCompletedNodes`.

- Affected subsystems: only `packages/workflows/src/dag-executor.ts` resume pre-populate + skip-check, `packages/workflows/src/schemas/dag-node.ts` schema field.
- Potential unintended effects: none expected for default-off behavior. If an author sets `always_run: true` on a node whose downstream consumer depends on idempotency, the consumer must tolerate re-execution — same contract as any producer node.
- Guardrails: new structured log event `dag.node_always_run_resume_forced` surfaces every time the flag fires, so operators can see resume behavior in workflow logs.

## Rollback Plan

- Fast rollback: revert the commit. Resume falls back to current behavior (always skip cached nodes). Authors who set `always_run` would see the field ignored — Zod still parses it (optional unknown field), no runtime error.
- Observable failure symptoms: if a node's resume behavior changes unexpectedly, grep workflow logs for `dag.node_always_run_resume_forced` to identify which node opted out.

## Risks and Mitigations

- Risk: A node author sets `always_run: true` on a destructive node (e.g. `rm -rf <dir>`) and triggers it on every resume.
  - Mitigation: documented in the authoring guide as "opt out of resume caching" with an example showing the producer/consumer use case. The flag is per-node, explicit, and visible in the YAML.
- Risk: Pre-populate exclusion accidentally strands downstream consumers that depend on the cached value being present before the producer re-runs.
  - Mitigation: downstream nodes are by definition in a later topological layer than the producer; `nodeOutputs.set(...)` writes the fresh value before the consumer's `substituteNodeOutputRefs` call. Test 3 in the new suite asserts this end-to-end.

## Summary by CodeRabbit

* **New Features**
  * Added an `always_run` node option to opt a node out of resume caching so it is re-executed on workflow resume.

* **Behavior**
  * Resuming workflows now forces `always_run` nodes to rerun and emits a dedicated workflow event when prior outputs are reset.

* **Documentation**
  * Added guidance and examples on opting out of resume caching and when to use `always_run`.

* **Tests**
  * Added tests validating parsing, resume behavior, event emission, and downstream output propagation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1730?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -194,6 +194,7 @@ nodes:
 | `context` | `'fresh'` \| `'shared'` | — | `fresh` = new session; `shared` = inherit from prior node. Defaults to `fresh` for parallel layers, inherited for sequential |
 | `idle_timeout` | number | — | Kill node if idle for this many milliseconds |
 | `retry` | object | — | Per-node retry configuration. See [Retry Configuration](#retry-configuration) |
+| `always_run` | boolean | `false` | Opt out of resume caching: re-run this node on resume even if a prior run completed it. See [Opting Out of Resume Caching](#opting-out-of-resume-caching) |
 
 **AI node options** — apply to `command` and `prompt` nodes:
 
@@ -551,6 +552,27 @@ Once the row reaches a terminal status, you can resume it explicitly via the pat
 **Known limitation**: AI session context from prior nodes is not restored. If a downstream node relies on in-context knowledge from a prior run's session (rather than artifacts), it may need to re-read those artifacts explicitly.
 
 **Fresh start**: If zero nodes completed in the prior run, Archon starts fresh (no nodes to skip).
+
+### Opting Out of Resume Caching
+
+By default, resume skips any node that completed successfully in the prior run and feeds its cached output to downstream consumers. That's the right behavior when a node's exit code captures the validity of its output (e.g. AI prompts, scripts that produce structured stdout).
+
+It's the wrong behavior when a node's success status doesn't capture output validity — typically a producer whose exit code reports the side effect (a file written, a service called) but whose downstream consumer parses the side effect's contents on every run. If the producer succeeded but wrote garbage, resume will replay the cached "success" forever without ever re-executing the producer.
+
+Set `always_run: true` on the node to force re-execution on resume, even when the prior run marked it completed:
+
+```yaml
+nodes:
+  - id: fetch-data
+    bash: ./scripts/download.sh > $ARTIFACTS_DIR/data.json
+    always_run: true        # Re-fetch on resume; download.sh exit code doesn't validate the JSON
+
+  - id: process-data
+    prompt: "Summarize $ARTIFACTS_DIR/data.json"
+    depends_on: [fetch-data]
+```
+
+On resume, `fetch-data` re-runs regardless of prior success, so `process-data` reads a freshly produced file. Normal cached nodes in the same run are still skipped — `always_run` is per-node.
 
 ---
 
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -4840,7 +4840,7 @@ describe('executeDagWorkflow -- always_run resume opt-out', () => {
     // Producer re-runs (instead of being skipped) AND consumer runs => 2 sendQuery calls
     expect(mockSendQueryDag.mock.calls.length).toBe(2);
 
-    // No skip event written for the always_run node
+    // No skip event written for the always_run node — but a reset event IS written for audit
     const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls;
     const skippedEvent = eventCalls.find(
       (call: unknown[]) =>
@@ -4848,6 +4848,16 @@ describe('executeDagWorkflow -- always_run resume opt-out', () => {
         (call[0] as { step_name: string }).step_name === 'producer'
     );
     expect(skippedEvent).toBeUndefined();
+
+    const resetEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_always_run_reset' &&
+        (call[0] as { step_name: string }).step_name === 'producer'
+    );
+    expect(resetEvent).toBeDefined();
+    expect((resetEvent![0] as { data: { prior_output: string } }).data.prior_output).toBe(
+      'cached stale output'
+    );
   });
 
   it('still skips non-always_run nodes in the same priorCompletedNodes set', async () => {
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -4769,6 +4769,189 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
   });
 });
 
+describe('executeDagWorkflow -- always_run resume opt-out', () => {
+  let testDir: string;
+
+  beforeEach(async () => {
+    testDir = join(
+      tmpdir(),
+      `dag-always-run-test-${Date.now()}-${Math.random().toString(36).slice(2)}`
+    );
+    const commandsDir = join(testDir, '.archon', 'commands');
+    await mkdir(commandsDir, { recursive: true });
+    await writeFile(join(commandsDir, 'producer.md'), 'Producer prompt');
+    await writeFile(join(commandsDir, 'consumer.md'), 'Consumer prompt $producer.output');
+
+    mockSendQueryDag.mockClear();
+    mockGetAgentProviderDag.mockClear();
+
+    mockSendQueryDag.mockImplementation(function* () {
+      yield { type: 'assistant', content: 'fresh output' };
+      yield { type: 'result', sessionId: 'session-id' };
+    });
+  });
+
+  afterEach(async () => {
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'claude',
+      getCapabilities: mockClaudeCapabilities,
+    }));
+    try {
+      await rm(testDir, { recursive: true, force: true });
+    } catch {
+      // ignore cleanup errors
+    }
+  });
+
+  it('re-runs node flagged always_run even when present in priorCompletedNodes', async () => {
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    const priorCompletedNodes = new Map([['producer', 'cached stale output']]);
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-always-run',
+      testDir,
+      {
+        name: 'always-run-producer',
+        nodes: [
+          { id: 'producer', command: 'producer', always_run: true },
+          { id: 'consumer', command: 'consumer', depends_on: ['producer'] },
+        ],
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig,
+      undefined,
+      undefined,
+      priorCompletedNodes
+    );
+
+    // Producer re-runs (instead of being skipped) AND consumer runs => 2 sendQuery calls
+    expect(mockSendQueryDag.mock.calls.length).toBe(2);
+
+    // No skip event written for the always_run node
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls;
+    const skippedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_skipped_prior_success' &&
+        (call[0] as { step_name: string }).step_name === 'producer'
+    );
+    expect(skippedEvent).toBeUndefined();
+  });
+
+  it('still skips non-always_run nodes in the same priorCompletedNodes set', async () => {
+    await writeFile(join(testDir, '.archon', 'commands', 'cached.md'), 'Cached prompt');
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    const priorCompletedNodes = new Map([
+      ['producer', 'cached stale output'],
+      ['cached', 'cached output'],
+    ]);
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-mixed',
+      testDir,
+      {
+        name: 'mixed',
+        nodes: [
+          { id: 'producer', command: 'producer', always_run: true },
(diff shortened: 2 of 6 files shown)


Wirasm:
## Review Summary

**Verdict**: blocking-issues

This PR adds `always_run: true` to opt individual workflow nodes out of resume caching. The feature intent and schema additions are sound, and the documentation is well-written. However, a critical logic error in the skip guard causes the `always_run` re-execute path and the normal skip path to both fire for the same node — which means the feature won't behave as documented until that bug is fixed. Additionally, the JSONL workflow log is missing an audit entry for `always_run` nodes, making it impossible to reconstruct resume decisions from logs.

### Blocking issues

- **`packages/workflows/src/dag-executor.ts:2644-2651`**: The `always_run` condition is nested inside the normal skip block, causing both re-execute and skip paths to fire for the same node. Move the `always_run` re-execute case **before** the skip block and invert the guard:

  ```typescript
  // always_run nodes re-execute even when in priorCompletedNodes
  if (priorCompletedNodes?.has(node.id) && node.always_run) {
    getLog().info({ nodeId: node.id }, 'dag.node_always_run_resume_forced');
    continue;  // ← re-execute and exit, don't fall into skip block
  }

  // normal skip path — only for nodes NOT flagged always_run
  if (priorCompletedNodes?.has(node.id) && !node.always_run) {
    // existing skip logic...
  }
  ```

  Also ensure the pre-population loop (lines 2596-2602) excludes `always_run` nodes so downstream `$nodeId.output` consumers get fresh output, not stale cached data:
  
  ```typescript
  if (priorCompletedNodes?.has(node.id) && !node.always_run) {
    nodeOutputs.set(node.id, { ... });
  }
  ```

- **`packages/workflows/src/dag-executor.ts:2634-2657`**: The `always_run` re-execute path doesn't call `logNodeSkip()` to write an audit entry to the JSONL workflow log. All other skip paths do this — `always_run` is the only one that omits it. Add:

  ```typescript
  await logNodeSkip(logDir, workflowRun.id, node.id, 'always_run').catch(
    (err: Error) => {
      getLog().warn({ err, nodeId: node.id }, 'dag.node_skip_log_write_failed');
    }
  );
  getWorkflowEventEmitter().emit({
    type: 'node_skipped',
    runId: workflowRun.id,
    nodeId: node.id,
    nodeName: node.command ?? node.id,
    reason: 'always_run',
  });
  ```

### Suggested fixes

- **`packages/workflows/src/schemas/dag-node.ts:543`**: `always_run` is added to `base` but not to the per-variant returns (`BashNode`, `ScriptNode`, `CommandNode`, `PromptNode`). Add it explicitly to each return path, or add a comment noting it flows from `base`.

- **`packages/workflows/src/dag-executor.test.ts`**: Add a test for `always_run: true` in a parallel layer with a non-always_run sibling — assert the `always_run` node re-executes while the sibling is skipped. Also add a test for explicit `always_run: false` in `priorCompletedNodes` to verify the skip path fires correctly.

- **`packages/docs-web/src/content/docs/guides/authoring-workflows.md`**: The example comment could reference the bug it fixes — add `(see #1391)` after "download.sh exit code doesn't validate the JSON".

### Minor / nice-to-have

- **`packages/workflows/src/dag-executor.ts:2647`**: Remove `See #1391.` from the comment — the surrounding text explains the intent without it, and PR numbers become opaque over time.

- **`dag-executor.test.ts:4816-4823`**: Once the `logNodeSkip()` fix above is applied, update the test `re-runs node flagged always_run` to assert the `always_run` reason event IS present (or filter it out for separate testing).

### Compliments

- The `always_run` exclusion comment in the pre-population loop is a model of useful documentation — it captures a subtle invariant (downstream consumers must see fresh output) that isn't obvious from the code alone.
- The schema approach of adding `always_run` to `dagNodeBaseSchema` and flowing it through via `base` spread is clean and avoids duplicating the field across six node type constructors.
- The `dag.node_always_run_resume_forced` log event name follows the established convention perfectly.

---

truffle-dev:
### Blocking issues

1. **`dag-executor.ts:2644-2651` — `always_run` and skip paths firing together.** They cannot — the two checks are separate `if` statements at the same nesting level with mutually exclusive guards:

   ```typescript
   if (priorCompletedNodes?.has(node.id) && node.always_run) {
     getLog().info(...)  // log + reset event, no return
   }
   if (priorCompletedNodes?.has(node.id) && !node.always_run) {
     ...                 // skip block ending in `return`
   }
   ```

   For any given node, `node.always_run` and `!node.always_run` cannot both be true, so at most one branch fires. When `always_run` fires, the function falls through to the trigger-rule check and normal execution — the skip block's `return` is unreachable for that node. The suggested `continue` rewrite would actually break the feature: inside `layer.map(async (node) => ...)`, `continue` is a syntax error, and a `return` there would short-circuit the always_run node before it executes.

   No change.

2. **`dag-executor.ts:2634-2657` — missing audit entry for `always_run`.** Valid concern on the DB side, with one wrinkle: emitting `node_skipped` with reason `'always_run'` (or calling `logNodeSkip(... 'always_run')`) is semantically wrong — the node is being re-executed, not skipped. The downstream consumers in `packages/cli/src/commands/workflow.ts` and `packages/core/src/db/workflow-events.ts` treat `node_skipped*` rows as "this node did not run."

   Added a distinct `node_always_run_reset` workflow event so resume forensics can see the opt-out without conflating it with a skip. The node's own `node_started` / `node_completed` events still fire as normal after that. (`9a8f1c47`)

### Suggested fixes

1. **`schemas/dag-node.ts:543` — `always_run` not in per-variant returns.** It does flow through — every variant return spreads `...base` (lines 576, 579, 583, 593, 602, 605, 609), and `base` includes `always_run` at line 546. The loader test at `loader.test.ts:680-702` parses YAML with `always_run: true` and asserts `workflows[0].nodes[0].always_run === true`. The TypeScript variant types declare `always_run` via `dagNodeBaseSchema`, so it's typed on every node.

2. **Test for parallel sibling.** Already covered at `dag-executor.test.ts:4853` ("still skips non-always_run nodes in the same priorCompletedNodes set") — `producer` (`always_run: true`) and `cached` have no `depends_on` so they share the first parallel layer, and the test asserts the producer re-runs while the sibling emits `node_skipped_prior_success`. The explicit `always_run: false` case is the default behavior covered by every other resume test in the suite.

3. **`authoring-workflows.md` — add `(see #1391)` to the example.** Skipping. Issue numbers go opaque to users six months out; the example already states why the comment matters ("`download.sh` exit code doesn't validate the JSON").

### Minor

1. **Remove `See #1391.` from the comment.** Done. (`9a8f1c47`)

2. **Update the `re-runs node flagged always_run` test after the audit fix.** Done in the same commit — now asserts the `node_always_run_reset` event IS emitted with the `prior_output` payload, in addition to the existing assertion that `node_skipped_prior_success` is NOT. (`9a8f1c47`)

Issue #1378, opened 2026-04-23
Slack/GitHub → claude subprocess: bun spawn doesn't deliver stdin, claude errors "Input must be provided" — mislabeled as #1067


yannick-opg:
## Summary

`archon serve` on macOS silently fails to respond to Slack DMs (and any route that depends on the Claude provider). The 60-second `claude.first_event_timeout` error currently points users to issue #1067 (nested-Claude env-var theory), but the true failure mode is different: **Archon's bun-based subprocess spawn of `claude` doesn't flush stdin to the child before claude errors out**.

Additionally, every Claude subprocess invocation runs the user's full SessionStart hook chain (19 hooks in my setup) on every Slack message — this is not itself the 60s timeout but is a ~10-30s startup tax on every reply.

## Environment

- macOS darwin-arm64, zsh
- Archon v0.3.9 (binary release)
- Claude Code CLI v2.1.118 at `~/.local/bin/claude`
- `archon serve` running under launchd with clean env (no `CLAUDE_CODE_ENTRYPOINT` inherited from any Claude Code session — verified via `ps -E`)
- `CLAUDE_BIN_PATH` set; `CLAUDE_USE_GLOBAL_AUTH=true`; standard setup wizard completed with Slack + GitHub

### What Archon claims when it times out

```json
{"module":"provider.claude","msg":"claude.first_event_timeout","timeoutMs":60000}
{"module":"provider.claude","err":{"message":"Claude Code subprocess produced no output within 60000ms. See logs for claude.first_event_timeout diagnostic dump. Details: https://github.com/coleam00/Archon/issues/1067"}}
```

### What's actually happening on the child side

`ps` during the 60-second window shows the `claude` subprocess is alive but has already errored:

```
yannick  26203  /Users/yannick/.local/bin/claude --output-format stream-json --verbose --input-format stream-json --model claude-sonnet-4-6 --disallowedTools Bash,... --permission-mode default

Error: Input must be provided either through stdin or as a prompt argument when using --print
```

So claude is already dead/erroring within the first second — Archon just doesn't see it because it's waiting for stdout JSON, while the error went to stderr.

### Control: same flags, same claude binary, same stream-json stdin, no Archon

Direct shell test with Archon's exact flag set + a well-formed stream-json message on stdin:

```bash
echo '{"type":"user","message":{"role":"user","content":"reply..."}}' | \
  /Users/yannick/.local/bin/claude --print --verbose \
    --input-format stream-json --output-format stream-json \
[...]

PR #1753, opened 2026-05-23
feat(slack): morph thinking ack into reply via chat.update

vvalliyur:
## Problem

Today the Slack adapter is fire-and-forget: when a user @-mentions the bot, the handler returns immediately and the user sees nothing in the thread until the workflow finishes — anywhere from a few seconds (\`archon-assist\`) to several minutes (\`archon-idea-to-pr\`). There's no native "is typing" indicator for bots in Slack channels, and the existing \`streamingMode\` flag is wired through to the adapter but unused for Slack (only Telegram implements streaming output).

This produces a poor UX, especially the first time users try the bot — many assume nothing's happening.

## Approach

Post a transient acknowledgment (\`⏳ _Archon is thinking…_\`) the moment the message is received, then **\`chat.update\`** that same message into the final reply when the workflow completes. The user sees one seamless message that transitions from "thinking…" to the result — no flicker, no clutter, no separate ack-then-reply pair.

Tried and rejected:
- **Emoji reactions** (👀 → ✅): cleaner but requires adding \`reactions:write\` scope, which forces existing self-hosted users to reinstall the app
- **Post ack then post separate reply then delete ack in finally**: this is what I tried first. For fast workflows (<5s) the ack flickered visibly before being replaced by a separate reply message — net worse than no ack
- **chat.update**: works with existing \`chat:write\`, gives a clean single-message UX

## Changes

\`packages/adapters/src/chat/slack/adapter.ts\`:
- New \`pendingAcks: Map<conversationId, ts>\` tracks the ack message per conversation
- \`sendThinkingAck()\` posts the ack and stores its ts
- \`sendMessage()\` consumes the pending ack via \`chat.update\` on the first chunk; long replies (>12k chars) fall through to additional \`postMessage\` calls for subsequent chunks
- \`clearPendingAck()\` only acts when sendMessage didn't consume the ack — guarantees we never accidentally delete the morphed reply
- New \`updateWithMarkdownBlock()\` helper mirrors \`sendWithMarkdownBlock\` but uses \`chat.update\`; falls back to a fresh postMessage if the update fails (e.g. the user manually deleted the ack)
- Info-level \`slack.thinking_ack_sent\` log so the ack posting can be verified in dev

\`packages/server/src/index.ts\`:
- Slack message handler awaits \`sendThinkingAck\` before \`acquireLock\`
- \`finally\` block calls \`clearPendingAck\` which is a no-op when sendMessage already morphed the ack — only fires when the workflow errored / produced no reply

## Compatibility

- No new OAuth scopes (uses \`chat:write\`, already required)
- No new env vars
- No \`config.yaml\` schema changes
- Failure paths fall back cleanly: ack post failure → workflow proceeds without indicator; chat.update failure → fresh postMessage

## Test plan

- [x] Unit-test surface: adapter methods exposed (sendThinkingAck, clearPendingAck) with happy + error paths
- [x] Local smoke: @mention with short and long replies, confirmed in-place morph and no flicker
- [x] Confirmed \`slack.thinking_ack_sent\` info log appears on receipt
- [ ] Multi-message workflows (long replies that chunk): first chunk consumes ack, rest post fresh — verified in adapter logic, not yet exercised end-to-end

## New Features
- Messages now display an immediate "thinking" acknowledgment upon receipt, which seamlessly transforms into the actual response once processing completes, providing instant user feedback before results are ready.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1753?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/adapters/src/chat/slack/adapter.ts
+++ b/packages/adapters/src/chat/slack/adapter.ts
@@ -24,6 +24,9 @@ export class SlackAdapter implements IPlatformAdapter {
   private streamingMode: 'stream' | 'batch';
   private messageHandler: ((event: SlackMessageEvent) => Promise<void>) | null = null;
   private allowedUserIds: string[];
+  /** Tracks the "thinking" ack message per conversation so the first
+   * outgoing reply can morph it via chat.update instead of posting fresh. */
+  private pendingAcks: Map<string, string> = new Map();
 
   constructor(botToken: string, appToken: string, mode: 'stream' | 'batch' = 'batch') {
     this.app = new App({
@@ -62,17 +65,56 @@ export class SlackAdapter implements IPlatformAdapter {
       ? channelId.split(':')
       : [channelId, undefined];
 
+    // If we have a pending "thinking" ack for this conversation, morph the
+    // first chunk into the ack via chat.update so users see a single message
+    // transition from "thinking…" to the reply with no flicker.
+    const pendingAckTs = this.pendingAcks.get(channelId);
+
     if (message.length <= MAX_MARKDOWN_BLOCK_LENGTH) {
-      // Use markdown block for proper formatting
-      await this.sendWithMarkdownBlock(channel, message, threadTs);
+      if (pendingAckTs) {
+        this.pendingAcks.delete(channelId);
+        await this.updateWithMarkdownBlock(channel, pendingAckTs, message, threadTs);
+      } else {
+        await this.sendWithMarkdownBlock(channel, message, threadTs);
+      }
     } else {
       // Long message: split by paragraphs
       getLog().debug({ messageLength: message.length }, 'slack.message_splitting');
       const chunks = splitIntoParagraphChunks(message, MAX_MARKDOWN_BLOCK_LENGTH - 500);
 
-      for (const chunk of chunks) {
-        await this.sendWithMarkdownBlock(channel, chunk, threadTs);
+      for (let i = 0; i < chunks.length; i++) {
+        if (i === 0 && pendingAckTs) {
+          this.pendingAcks.delete(channelId);
+          await this.updateWithMarkdownBlock(channel, pendingAckTs, chunks[i], threadTs);
+        } else {
+          await this.sendWithMarkdownBlock(channel, chunks[i], threadTs);
+        }
       }
+    }
+  }
+
+  /**
+   * Replace an existing message in-place via chat.update. Used to morph the
+   * "thinking" ack into the actual reply. Falls back to posting a new
+   * message if the update fails (e.g. message was deleted by user).
+   */
+  private async updateWithMarkdownBlock(
+    channel: string,
+    ts: string,
+    message: string,
+    threadTs?: string
+  ): Promise<void> {
+    try {
+      await this.app.client.chat.update({
+        channel,
+        ts,
+        blocks: [{ type: 'markdown', text: message }],
+        text: message.substring(0, 150) + (message.length > 150 ? '...' : ''),
+      });
+      getLog().debug({ ts, messageLength: message.length }, 'slack.ack_updated');
+    } catch (error) {
+      getLog().warn({ err: error, channel, ts }, 'slack.ack_update_failed');
+      await this.sendWithMarkdownBlock(channel, message, threadTs);
     }
   }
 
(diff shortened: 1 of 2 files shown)


Wirasm:
## Review Summary

**Verdict**: blocking-issues

Your PR adds a "thinking" acknowledgment to the Slack adapter — a transient message posted immediately on receipt, then morphed into the real reply via `chat.update`. The UX improvement is sound and the implementation follows project conventions well. However, there's one race condition that causes reply messages to disappear mid-conversation, and three entirely untested new methods that need coverage before merging.

### Blocking issues

- **`packages/server/src/index.ts:451`** — Race condition in ack clearing: `clearPendingAck()` runs in a `finally` block *inside* the async lock callback. When messages queue for the same conversation, a later message's `handleMessage` completion clears the pending ack that an earlier message's `sendMessage` still needs to morph into the reply. Result: users see replies vanish.

  **Fix**: Move `clearPendingAck` out of the lock and into a callback that fires only after `sendMessage` completes. Options: (a) have `handleMessage` return a promise that resolves when the reply is sent, then clear in `.then()`; (b) pass the ack timestamp through `HandleMessageOptions` so the orchestrator clears it after `sendMessage`; (c) use a per-conversation sequence number so each clear can detect staleness.

### Suggested fixes

- **`packages/adapters/src/chat/slack/adapter.ts:101`** — `sendThinkingAck` (entire method) has no tests. Add tests for: happy path (`ts` returned → map populated), fallback path (no `ts` → returns `null`, map untouched), rejection path (error → `null`, map untouched), and thread channel ID splitting.

- **`packages/adapters/src/chat/slack/adapter.ts:124`** — `clearPendingAck` (entire method) has no tests. Add tests for: no-op when ack not in map, happy path deletion, and error path where `chat.delete` throws (map still cleared, no throw).

- **`packages/adapters/src/chat/slack/adapter.ts:86`** — `updateWithMarkdownBlock` (entire method) has no tests. Add tests for: `chat.update` succeeds, `chat.update` fails (falls back to `sendWithMarkdownBlock`), and truncation to 150 chars.

- **`packages/adapters/src/chat/slack/adapter.ts:65`** — `sendMessage` ack-morphing branch is untested. Test: short message WITH pending ack → `updateWithMarkdownBlock` called, `pendingAcks` entry deleted, `postMessage` NOT called. Short message WITHOUT pending ack → existing behavior unchanged. Long message WITH/without pending ack → first chunk vs. all chunks.

- **`packages/server/src/index.ts:453`** — `sendThinkingAck` server wiring has no integration test. Add a test that fires a fake Slack message event and verifies `sendThinkingAck` is called before the lock is acquired.

- **`packages/adapters/src/chat/slack/adapter.ts:116`** — `sendThinkingAck`: add test for falsy `ts` (empty string or `0`) → returns `null`, `pendingAcks` not populated.

- **`packages/adapters/src/chat/slack/adapter.ts:126`** — `clearPendingAck`: add tests for channel ID parsing (with `:` → thread ts, without `:` → full string as ts).

### Minor / nice-to-have

- **`packages/adapters/src/chat/slack/adapter.ts:28`** — `pendingAcks` map never cleared on shutdown. Add `this.pendingAcks.clear()` to `stop()` to avoid stale in-memory entries on restart.

- **`packages/adapters/src/chat/slack/adapter.ts:175-177`**, **adapter.ts:191-195**, **`packages/server/src/index.ts:448-452`** — Comments reference method names (`sendMessage`, `clearPendingAck`) that will rot if refactored. Rewrite to describe intent: "Store the ack so it morphs into the first reply instead of spawning a duplicate."

- **`packages/adapters/src/chat/slack/adapter.ts:449`** — `sendThinkingAck` return value is awaited but discarded. No immediate action needed, but document the `string | null` semantics if the API becomes a caller-contract.

- **Mock isolation**: `slack/adapter.test.ts` and `telegram/adapter.test.ts` both use `mock.module('@archon/paths', ...)` identically. Per CLAUDE.md, prefer `spyOn()` for internal module mocking to avoid cross-test pollution.

### Compliments

- The `updateWithMarkdownBlock` JSDoc correctly captures the fallback-to-post behavior as non-obvious WHY — exactly the comment style this codebase needs more of.
- Error handling throughout is consistently graceful: `sendThinkingAck` returns `null` on failure (not an exception), `updateWithMarkdownBlock` falls back to posting, and `clearPendingAck` swallows delete errors. The degraded-but-safe philosophy is well-applied.
- Log event naming (`slack.thinking_ack_sent`, `slack.ack_updated`, etc.) follows the `{domain}.{action}_{state}` convention throughout.

---

Wirasm:
@$vvalliyur related to #1378 — overlapping area or partial fix.

Wirasm:
@$vvalliyur related to #1469 — overlapping area or partial fix.

Wirasm:
@$vvalliyur related to #1378 — overlapping area or partial fix.

Wirasm:
@$vvalliyur related to #1469 — overlapping area or partial fix.

Wirasm:
Hi @vvalliyur — thanks for opening this PR.

- Validation Evidence
- Security Impact
- Human Verification
- Side Effects / Blast Radius
- Rollback Plan

Wirasm:
Thanks @vvalliyur — the instinct here (kill the "nothing's happening" gap without adding channel noise) was exactly right, and it got independently validated three days after you opened this: the #1757 umbrella Slack UX upgrade shipped a fuller version of the same idea — a running reaction on the triggering message that swaps to success/failure, plus a live in-thread status card updated via `chat.update` with per-node progress. Your morphing ack is a strict subset of that, and layering both would recreate the clutter you set out to remove.

Closing as superseded. For what it's worth, the edge-case handling in this PR (ack consumption on chunk 0 of long replies, the pending-ack cleanup on error paths, the chat.update fallback) was genuinely well thought through — the timing just didn't break your way. Appreciate the contribution.

PR #1783, opened 2026-05-27
feat(core): plumb user_id from chat/forge adapters through orchestrator and workflow runs

Wirasm:
## Summary

- **Problem**: Archon is architecturally single-user — one shared `GITHUB_TOKEN`, no users table, every conversation/message/workflow_run unattributed. A small team sharing one VPS can't tell who triggered which run, and every teammate's PR comments appear under the same GitHub avatar.
- **Why it matters**: Without per-user attribution in the schema, every later team-oriented feature (ACLs, "my runs", per-user budgets, audit logs) ships as a refactor instead of a one-line filter. This is the blocking layer.
- **What changed**: Two new tables (`remote_agent_users` + `remote_agent_user_identities`), nullable `user_id` on the four primary tables, and end-to-end plumbing of a stable Archon `user_id` from every chat/forge adapter (Slack, Telegram, Discord, GitHub) through `HandleMessageContext` into conversations, workflow runs, and isolation environments. Slack additionally enriches first-sight users with their real name via `users.info`.
- **What did NOT change (scope boundary)**: Web UI authentication and CLI identity are explicitly deferred — those rows continue to write `user_id = NULL` (follow-up PR wires the `X-Archon-User` header and `archon auth github` device flow). No GitHub App swap yet (separate follow-up). No ACLs, no `requires: [github]` workflow gate, no UI changes.

### Before

```
  Slack U-id ──► adapter reads event.user ──► ✗ DROPPED before orchestrator
  Telegram from.id ──► emitted in payload   ──► ✗ DROPPED by server handler
  Discord author.id ──► emitted             ──► ✗ DROPPED
  GitHub sender.login ──► used for whitelist ──► ✗ DROPPED after auth check

  handleMessage(platform, conversationId, content, ctx)  // ctx has no userId
       │
       ▼
  INSERT INTO conversations (...)            ◀── no user_id column exists
  INSERT INTO workflow_runs (...)            ◀── no user_id column exists
  INSERT INTO isolation_environments (...)   ◀── no created_by_user_id column

  SQL audit: SELECT user_id FROM remote_agent_conversations
  → ERROR: column "user_id" does not exist
```

### After

```
  Slack U-id ──► adapter [users.info] ──► resolve('slack', U-id, real_name)
  Telegram from.id ──► [first_name+last_name] ──► resolve('telegram', id, displayName)
  Discord author.id ──► [author.username]    ──► resolve('discord', id, username)
  GitHub sender.login ──► resolve('github', login, login)
                            │
[...]

PR #1774, opened 2026-05-26
feat: Keycloak/OIDC multi-user auth with per-user GitHub OAuth

leex279:
## Summary

- **Problem:** Archon has no auth — every `/api/*` route is publicly accessible, there is no user table, and a single shared `GITHUB_TOKEN` env var is used for all clones. This blocks any hosted, multi-developer deployment.
- **Why it matters:** Foundational identity infrastructure is needed before any other multi-tenancy or RBAC feature can be layered on. Engineering teams want a single shared Archon instance instead of per-person local installs.
- **What changed:** Added OIDC (Keycloak) login + JWT middleware on `/api/*`, a `remote_agent_users` table with AES-256-GCM-encrypted per-user GitHub OAuth tokens, `created_by_user_id` attribution on conversations and workflow runs, and per-user GitHub token usage in the clone path. Web UI now redirects unauthenticated users to login and shows a "Connect GitHub" button.
- **What did NOT change:** Backwards-compatible no-op when `KEYCLOAK_URL` is unset — single-user mode still works unchanged. No RBAC, no per-user AI provider selection, no multi-tenancy, no CLI/Slack/Telegram/GitHub-adapter auth changes.

### Before

```
Web user           Archon API
────────           ──────────
opens /            anonymous request
sends message ───▶ no auth check
                   accepts request
                   clones with shared GITHUB_TOKEN
sees reply  ◀───── responds
```

### After (multi-user mode, KEYCLOAK_URL set)

```
Web user           Archon API                Keycloak
────────           ──────────                ────────
opens / ─────────▶ /api/auth/me → 401
                   ◀──── redirect
clicks login ────▶ /api/auth/login ────────▶ login page
authenticates ◀──────────────────────────── 
returns to /  ◀── set HTTP-only cookie
                   c.set('user', {sub})
sends message ───▶ JWT validated via JWKS
                   *created_by_user_id stamped*
                   clones with *per-user GitHub token*
                   (falls back to org GITHUB_TOKEN if not connected)
sees reply  ◀───── responds

Connect GitHub ──▶ /api/auth/github ────────▶ GitHub OAuth
                   encrypts + persists token
```

### After (single-user mode, KEYCLOAK_URL unset)

```
(unchanged from Before — middleware is a no-op)
```

### Before

```
[Web UI] ──▶ [/api/*] ──▶ [Routes] ──▶ [Core/DB]
                                       ──▶ [Git clone using env GITHUB_TOKEN]
```

### After

```
[...]

PR #1788, opened 2026-05-28
feat(core): swap shared GitHub PAT for GitHub App with multi-installation routing

Wirasm:
## Summary

- **Problem:** Archon's bot identity is a single shared `GITHUB_TOKEN` PAT. One revocation kills everyone, comments all post under one operator's avatar, no auto-rotation, no multi-org support, and webhooks need per-repo setup.
- **Why it matters:** Phase 2 of [the team-foundation PRD](.claude/PRPs/prds/github-app-and-user-identity.prd.md). Without this, every team-oriented feature downstream stays blocked on shared identity. PR-A (#1783) landed the `user_id` plumbing — PR-B replaces the bot identity.
- **What changed:** New `@archon/core/github-auth/` module wraps `@octokit/auth-app` with a two-level cache (`owner/repo → installationId`, `installationId → token`, refresh-on-access with a 5-min buffer). `GitHubAdapter` takes a `GitHubAuth` discriminated union (`{kind:'pat'}` | `{kind:'app'}`); the 4 Octokit callsites + the clone token embed route through per-(owner, repo) `resolveOctokit` + a `withTokenRefresh` wrapper that evicts on 401 and retries once. Server bootstrap fails fast if both modes are configured. New `POST /internal/git-credential` endpoint (App mode only) backs a POSIX git credential helper installed at clone time so workflows that outlive the 1h token can refresh in place.
- **What did NOT change (scope boundary):** Per-user GitHub tokens (device flow), `requires:[github]` YAML gate, interim web UI auth, CLI `archon auth github` — all PR-C. GitHub Enterprise Server, installation lifecycle webhook events, live integration tests against api.github.com, token-encryption-key rotation tooling — out of scope per PRD.

### Before

```
Operator setup:
  1. Generate personal PAT on github.com
  2. Set GITHUB_TOKEN=<PAT> in .env
  3. Per repo: configure webhook → https://archon/webhooks/github
  4. Per repo: add WEBHOOK_SECRET

Bot post-comment runtime:
  Webhook ──▶ verifySig (HMAC) ──▶ Octokit({auth: GITHUB_TOKEN}) ──▶ posts under PAT owner's avatar
  cloneRepository({token: GITHUB_TOKEN}) ──▶ remote URL has embedded token
```

### After

```
Operator setup (one-time):
  1. Register a GitHub App ("Archon Bot"); copy App ID + private key PEM
  2. Configure ONE webhook URL on the App → https://archon/webhooks/github
  3. Install the App on each org/personal account
  4. Set GITHUB_APP_ID, GITHUB_APP_PRIVATE_KEY[_PATH], GITHUB_APP_SLUG, WEBHOOK_SECRET

Solo install (UNCHANGED): GITHUB_TOKEN + WEBHOOK_SECRET → PAT mode preserved

Bot post-comment runtime (App mode):
  Webhook ──▶ verifySig ──▶ primeInstallationLookup(owner, repo, event.installation.id) [skips one HTTP]
          ──▶ resolveOctokit(owner, repo) → per-install Octokit (cached)
          ──▶ posts as <slug>[bot] with the App badge

  401? ──▶ invalidateToken(installationId) ──▶ retry once with fresh token
  Token expiry? ──▶ refreshed transparently 5min ahead of 1h expiry

  cloneRepository({token: getInstallationToken(owner, repo)})
                  ──▶ post-clone: installCredentialHelper(worktreePath)
                                  for >1h workflow refresh via POST /internal/git-credential
```

### Before

```
                  ┌─────────────┐
                  │  process.env│
                  │  GITHUB_TOKEN│
                  └──────┬──────┘
                         │
              ┌──────────┴──────────┐
              ▼                     ▼
   ┌────────────────────┐   ┌────────────────┐
   │  GitHubAdapter     │   │ cloneRepository│
   │  Octokit (1 inst)  │   │  URL embed     │
   └─────────┬──────────┘   └────────────────┘
             │
             ▼
   ┌────────────────────┐
   │  api.github.com    │
   └────────────────────┘
```

### After

```
        ┌────────────────────────────────────────────┐
        │  packages/core/src/github-auth/  [+]       │
        │  ┌──────────────────────────────────────┐  │
        │  │ createGitHubAppAuthProvider()        │  │
        │  │  • lookupCache: owner/repo→install   │  │
        │  │  • tokenCache:  install→token (1h)   │  │
        │  │  • octokitCache per installation     │  │
        │  └──────────┬───────────────────────────┘  │
        └─────────────┼──────────────────────────────┘
                      │                            ▲
        ===           ▼                            │ ===
   ┌────────────────────────┐         ┌────────────┴──────────────┐
   │ GitHubAdapter  [~]     │         │ server bootstrap  [~]     │
   │  • auth: GitHubAuth    │◀────────┤  • dual-mode env detect   │
   │  • resolveOctokit      │         │  • fail-fast on conflict  │
   │  • withTokenRefresh    │         │  • register on singleton  │
   │  • botLogin getter     │         └──────────────┬────────────┘
   │  • <slug>[bot] filter  │                        │
   └────┬─────────┬─────────┘                        │ ===
        │         │                                   ▼
        │         │ ===                  ┌──────────────────────────┐
        │         └────────────────────▶ │ store-adapter  [~]       │
        │                                │  registerGitHubAppAuth() │
        │                                │  createWorkflowDeps()    │
        │ ===                            │  resolveBotGitHubToken?  │
        ▼                                └──────────┬───────────────┘
   ┌────────────────────┐                            │ ===
   │ /internal/         │                            ▼
   │ git-credential [+] │                ┌─────────────────────────┐
   │ (App mode only)    │                │ workflow executor  [~]  │
   └──────────┬─────────┘                │  GH_TOKEN injection at  │
              │                          │  workflow start         │
              ▼                          └─────────────────────────┘
   ┌────────────────────┐
   │ scripts/git-       │
   │ credential-archon  │
   │ .sh  [+]           │
   └────────────────────┘
```

**Connection inventory:**

| server bootstrap | `loadAppPrivateKey()` | **new** | PEM resolution from env (inline / file) |
| server bootstrap | `createGitHubAppAuthProvider()` | **new** | App-mode only |
| server bootstrap | `registerGitHubAppAuthProvider()` | **new** | Wires provider into workflows |
| server bootstrap | `GitHubAdapter` constructor | **modified** | `GitHubAuth` discriminated union (was `token: string`) |
| `GitHubAdapter` | `IGitHubAppAuthProvider` | **new** | App-mode runtime dependency |
| `GitHubAdapter` (4 Octokit sites) | `withTokenRefresh` → `resolveOctokit` | **modified** | Was direct `this.octokit.rest.X` |
| `GitHubAdapter.handleWebhook` | `primeInstallationLookup` | **new** | Short-circuits one HTTP per webhook |
| `GitHubAdapter.ensureRepoReady` | `getInstallationToken` | **new** | App-mode clone; PAT mode unchanged |
| `GitHubAdapter.ensureRepoReady` | `installCredentialHelper` | **new** | App-mode post-clone hook |
| `store-adapter.createWorkflowDeps()` | `IGitHubAppAuthProvider.getInstallationToken` | **new** | Via optional `WorkflowDeps.resolveBotGitHubToken` |
| `workflows/executor.executeWorkflow` | `deps.resolveBotGitHubToken` | **new** | Injects `GH_TOKEN`/`GITHUB_TOKEN` at workflow start |
| Hono `POST /internal/git-credential` | `GitHubAdapter.getInstallationToken` | **new** | Backs credential helper |
| `git-credential-archon.sh` | `POST /internal/git-credential` | **new** | Refreshes installation token for >1h worktrees |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: L`
- Scope: `core` (github-auth module is the heart); also touches `adapters`, `server`, `workflows`, `docs`
- Module: `core:github-auth`, `adapters:github`, `server:bootstrap`, `workflows:executor`

## Change Metadata

- Change type: `feature` (also touches `refactor` for the adapter callsites)
- Primary scope: `multi` — `core` (new module + workflow injection wire-up), `adapters` (GitHubAdapter dual-mode), `server` (bootstrap + internal endpoint), `workflows` (subprocess env injection)

## Linked Issue

- Closes #
- Related #1783 (PR-A — `user_id` foundation; merged on dev as `50beeeca`)
- Depends on #1783 (PR-A's user-identity plumbing — required so per-user attribution in PR-C has somewhere to land)
- Supersedes # — (the PRD intentionally avoids the Keycloak approach from #1774)

```bash
bun run type-check    # ✓ all 10 packages clean
bun run lint          # ✓ --max-warnings 0
bun run format:check  # ✓ Prettier clean
bun run test          # ✓ every package green
bun run validate      # ✓ exit 0
```

- Evidence provided (test/log/trace/screenshot): all six `bun run validate` stages exit 0 locally; 15 new strictly-mocked auth-module tests (`packages/core/src/github-auth/auth.test.ts`); 8 new App-mode adapter tests + all 54 existing PAT-mode tests passing (`packages/adapters/src/forge/github/adapter.test.ts`). No live api.github.com calls in CI per PRD Q7.
- If any command is intentionally skipped, explain why: integration tests against api.github.com are deferred per PRD Q7 (test GitHub App + CI secrets out of scope for v1).

## Security Impact (required)

- New permissions/capabilities? **Yes** — App-mode operators must register a GitHub App and grant Contents:Read, Issues:RW, PullRequests:RW. The App permission surface is **narrower** than a PAT (no user-wide scope). Documented in `packages/docs-web/.../github-app-setup.md`.
- New external network calls? **Yes** — `@octokit/auth-app` makes a JWT-signed call to `GET /repos/{owner}/{repo}/installation` (cached 1h) and `POST /app/installations/{id}/access_tokens` (cached, refreshed 5min before 1h expiry). Both are standard App-auth flows.
- Secrets/tokens handling changed? **Yes** — installation access tokens are now memory-only (1h lifetime), not env. The private key PEM remains in env / on disk (operator's choice). No tokens persisted to disk.
- File system access scope changed? **Yes** (limited) — `installCredentialHelper()` copies `scripts/git-credential-archon.sh` to `~/.archon/bin/` and writes a per-worktree `git config credential.https://github.com.helper` entry. Idempotent; only on App-mode clone.
- **`/internal/git-credential` endpoint hands out live installation tokens.** MUST be exposed on 127.0.0.1 only OR the reverse proxy MUST drop `/internal/*`. Server emits a startup WARN (`github_app.internal_endpoint_exposed_check_reverse_proxy`) if bound to a non-loopback interface with App mode active. Documented in `CLAUDE.md` API endpoints section and the operator setup guide.

## Compatibility / Migration

- Backward compatible? **Yes** — PAT-mode behaviour is byte-identical for solo installs. All 54 existing PAT-mode adapter tests pass unchanged.
- Config/env changes? **Yes** — new optional env vars: `GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY` (or `_PATH`), `GITHUB_APP_SLUG`, `GITHUB_APP_INSTALLATION_ID`. Documented in `.env.example` + `github-app-setup.md`. `GITHUB_TOKEN` continues to work.
- Database migration needed? **No** — no schema changes (PR-A landed schema; PR-B is pure auth).
- **Fail-fast guardrail:** server refuses to start if both `GITHUB_APP_ID` and `GITHUB_TOKEN` are set. Operators must pick one mode explicitly.
- If yes, exact upgrade steps: see `packages/docs-web/src/content/docs/adapters/github-app-setup.md` Step 6.

- **Verified scenarios:**
  - `bun run validate` green end-to-end (all 6 stages).
  - 15 auth-module unit tests cover: PEM loading (inline, file, `\n` normalisation, missing-env, bad-PEM); token cache (fresh hit, refresh-near-expiry, default-installation-id skip); install lookup (cache, case-insensitive, 404→AppNotInstalledError); priming; invalidation; per-install Octokit memoisation; multi-install Octokit distinctness.
  - 8 App-mode adapter tests cover: per-(owner, repo) Octokit routing, multi-install resolution, `event.installation.id` priming, 401→invalidate+retry, `<slug>[bot]` self-filter, clone-token resolution (vs env), `AppNotInstalledError` surfacing, credential-helper install attempted post-clone.
- **Edge cases checked:**
  - Both modes configured → server refuses to start.
  - Missing PEM file → `AppPrivateKeyError` at bootstrap, not at first webhook.
  - PEM with literal `\n` (the `.env`-quoted convention) → normalised on load.
  - Same `(owner, repo)` from two different installations → routed to distinct tokens.
  - `event.installation` absent (older / replayed webhook) → falls back to lookup path.
  - PAT mode + this PR → all 54 pre-existing adapter tests pass.
- **What was not verified:** live `api.github.com` integration (deferred per PRD Q7). The `installCredentialHelper` flow with an actual >1h workflow wasn't end-to-end exercised — covered by unit-level test for the post-clone `git config` invocation, but the real-world refresh path needs a long-running workflow on a deployed instance.

- **Affected subsystems/workflows:** GitHub adapter (the only forge adapter touched); server bootstrap; workflow executor (subprocess env, only when App mode active); webhook ingestion path.
  - `/internal/git-credential` exposed beyond loopback would hand out live tokens — guarded by startup WARN and operator documentation.
  - `installCredentialHelper` writes per-worktree git config; idempotent and scoped to `credential.https://github.com.helper` so it doesn't collide with operator-set helpers.
  - GH_TOKEN injected at workflow start has 1h TTL — workflows >1h fall through to the credential helper for `git push`; `gh` commands hitting the API after expiry would 401 (operators handle via the helper or by keeping workflows short).
- **Guardrails/monitoring for early detection:**
  - Startup logs: `github.adapter_mode_app` vs `github.adapter_mode_pat` make the active mode obvious.
  - Per-event logs: `github_auth.install_lookup_completed`, `github_auth.token_resolve_completed`, `github_auth.token_cache_evicted_on_401` give visibility into token lifecycle.
  - `github_app.internal_endpoint_exposed_check_reverse_proxy` startup WARN flags misconfigured binds.

## Rollback Plan (required)

- **Fast rollback command/path:** unset the `GITHUB_APP_*` env vars, set `GITHUB_TOKEN` back, restart Archon. Adapter snaps back to PAT mode. No data migration to undo (no schema changes).
- **Feature flags or config toggles (if any):** the App-mode env vars themselves act as the toggle — App mode is opt-in via env. Removing them returns to legacy behaviour.
- **Observable failure symptoms:** 
  - Webhook handler emitting `github_auth.token_resolve_failed` or `github_auth.install_lookup_failed` repeatedly.
  - Bot comments not appearing under `<slug>[bot]`.
  - Clones failing with `AppNotInstalledError`.
  - 401 storm in adapter logs.

## Risks and Mitigations

- **Risk:** `/internal/git-credential` exposed beyond loopback hands out live installation tokens.
  - **Mitigation:** startup WARN when server binds non-loopback with App mode active; documented Caddy/nginx pattern in `github-app-setup.md`; CLAUDE.md API table flags it as 127.0.0.1-only.
- **Risk:** PEM private-key newline mangling in `.env` files.
  - **Mitigation:** `loadAppPrivateKey` supports both literal newlines and the `\n` escape; two unit tests cover both paths.
- **Risk:** Installation lookup cache goes stale (App uninstalled → reinstalled with a new id).
  - **Mitigation:** 1h TTL on lookup cache; 401 → `invalidateToken` + single retry surfaces `AppNotInstalledError` cleanly if the install really is gone.
- **Risk:** Workflows that AI-write `git push` after the 1h URL-embedded token expires.
  - **Mitigation:** `installCredentialHelper` post-clone in App mode; `POST /internal/git-credential` returns a fresh token on demand.
- **Risk:** Operator confused about active mode, sets both env sets.
  - **Mitigation:** fail-fast on both configured, with a clear "pick one" error message at startup.
- **Risk:** `@octokit/auth-app` v8 ESM-only — Bun import resolution.
  - **Mitigation:** `@octokit/rest` v22 (the matching peer, already used in `@archon/adapters`) is also ESM-only; the same path works. Verified locally via the new auth tests.

## Summary by CodeRabbit

* **New Features**
  * GitHub App auth mode with per-installation tokens, automatic refresh on 401, and internal endpoint + git credential helper for long-running pushes.

* **Configuration**
  * .env template documents mutually-exclusive PAT vs App modes, required App vars (app id, private key path/inline, slug, optional installation id), and a loopback-binding opt-in escape hatch; server now fails fast on conflicting settings.

* **Documentation**
  * New App-mode setup, migration, and security guidance including webhook and internal-endpoint recommendations.

* **Tests**
  * Expanded test coverage for App-mode auth, token caching, and credential helper behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1788?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Wirasm:
## PR Review Summary — 7 specialized agents

Targeted review focused on the 5 concerns raised on PR-B plus general quality. Findings cross-validated across agents; consensus items flagged.

---

### Direct answers to the 5 questions you flagged

**Q1. /internal/git-credential loopback guard — defensive enough?**
**No. WARN is the wrong choice — it should fail-fast.** Both code-reviewer (92/100 confidence) and silent-failure-hunter independently flagged this as critical. A misconfigured operator binding to `0.0.0.0` with App mode active leaks live installation tokens to anyone who can reach the port. WARN-level logs are routinely filtered or scrolled past in production. Recommended pattern:

```ts
if (githubAppAuthProvider && hostname !== '127.0.0.1' && hostname !== 'localhost') {
  if (!process.env.ARCHON_ALLOW_INTERNAL_ON_PUBLIC_BIND) {
    getLog().fatal({ hostname }, 'github_app.internal_endpoint_public_bind_rejected');
    process.exit(1);
  }
  getLog().warn({ hostname }, 'github_app.internal_endpoint_exposed_check_reverse_proxy');
}
```

Defense-in-depth alternative: bind `/internal/*` to a separate loopback-only listener instance, or add a request-level `x-forwarded-for` check inside the handler (any forwarded request → 403).

**Q2. withTokenRefresh 401 retry — infinite-loop risk?**
**No infinite loop is possible.** The second call `fn(fresh)` at adapter.ts:172 is not wrapped in try/catch — a second consecutive 401 propagates directly. A genuine App-uninstall produces `403`, not `401`, so the retry path doesn't engage at all. **No circuit breaker needed for the 401 case itself**, BUT two related defects surfaced:

- **(critical, code-reviewer)** `lookupCache` is never evicted on 401. `invalidateToken` only touches `tokenCache`. After an App reinstall (which assigns a new `installationId`), the lookup cache serves the stale ID for up to 1h, causing every API call against that repo to fail in a loop until TTL expiry. Self-heal won't kick in until the 1h lookup TTL runs out.
- **(critical, silent-failure-hunter)** The retry `fn(fresh)` has no surrounding try/catch with an ERROR log. Retry failures bubble up with zero context — debugging requires reconstructing the call stack from unrelated upstream logs.

**Q3. PEM `\n`-escape handling — common .env pitfalls?**
**Mostly correct.** The `inline.includes('\\n') ? replace : pass-through` heuristic handles all realistic .env quoting styles correctly (double-quoted → dotenv pre-processes; single-quoted/unquoted → literal `\n` survives; `replace` only matches the two-char sequence, not real newlines). Edge cases worth fixing:

- **(important, pr-test-analyzer + code-reviewer)** Windows `\r\n` in a `.pem` file loaded via `PRIVATE_KEY_PATH` fails at first JWT signing, not at startup. Fix: extend regex to `replace(/\\r\\n|\\n/g, '\n')` and add a `.replace(/\r\n/g, '\n')` for file-loaded PEMs.
- **(suggestion, code-simplifier)** The `includes('\\n') ? ... : inline` ternary is dead branching — `replace` with no match is a no-op. Drop the conditional.
- **(minor)** Whitespace-only `GITHUB_APP_PRIVATE_KEY=   ` falls through to `_PATH` lookup rather than surfacing "the inline key you set is empty"; produces a confusing error message.

**Q4. Cache TTLs — 55-minute mark behavior, correctness?**
**At 55 min the refresh fires correctly.** `Date.now() + REFRESH_BUFFER_MS (5min) < cached.expiresAt` evaluates to `false` at `now + 55min` (since `expiresAt = now + 60min`), so the token is re-issued. The 5-min buffer is reasonable; no off-by-one.

**Lookup cache 1h TTL is the right call** — but only if Q2's `lookupCache` invalidation gap is fixed first. Infinite TTL + 401-invalidated would actually be **worse** today, because a reinstall would permanently serve a stale ID until process restart. Fix the eviction first, then consider extending TTL.

**Q5. Credential helper shell script — injection risk?**
**No direct exploit, but two defense-in-depth gaps:**

- **(important, code-reviewer 80/100)** `$path` is unvalidated in the shell. `$host` is checked (`[ "$host" = "github.com" ]`) before use, so host injection is blocked. `$path` flows verbatim into the curl JSON body. The server's path regex (`^([^/]+)\/([^/]+?)(?:\.git)?$`) blocks crafted paths end-to-end, so the present attack surface is empty — but the structural guarantee belongs in the script too. Add a `case "$path" in */*) ;; *) exit 0 ;; esac` check.
- **(important, silent-failure-hunter)** `curl -fsS ... || exit 0` produces no stderr on connection failure. Combined with the `[ -n "$token" ] || exit 0` fallthrough, an unattended workflow gets "Authentication failed" with no diagnostic. Fix: stderr-echo a one-line reason before each `exit 0`.
- **(positive)** No `set -x` token leak. `$token` is correctly quoted in the final printf. CRLF input from Windows git produces silent fallthrough (not exploitable, but worth noting). The `sed -n 's/.*"token":"\([^"]*\)".*/\1/p'` JSON extractor is fragile — a token containing `/` or `&` would silently produce empty on some sed implementations.

---

### Critical issues (must fix before merge)

| ID | Issue | Location | Source |
|----|-------|----------|--------|
| C1 | `/internal/git-credential` non-loopback bind only emits WARN; should fail-fast | `packages/server/src/index.ts:772-774` | code-reviewer + silent-failure-hunter |
| C2 | `lookupCache` never evicted on 401 → 1h dead period after App reinstall | `packages/core/src/github-auth/auth.ts:134-137`, `packages/adapters/src/forge/github/adapter.ts:162-170` | code-reviewer (82/100) |
| C3 | `withTokenRefresh` retry has no try/catch around `fn(fresh)` — second-attempt failures silently lose context | `packages/adapters/src/forge/github/adapter.ts:171-172` | silent-failure-hunter |
| C4 | `/internal/git-credential` failure logs at WARN, not ERROR; live token resolution failure is filtered as noise | `packages/server/src/index.ts:657-659` | silent-failure-hunter |

### Important issues

| ID | Issue | Location | Source |
|----|-------|----------|--------|
| I1 | `installCredentialHelper` returns silently in binary builds; caller logs `credential_helper_installed` (false success). No discriminated result type. | `credential-helper-install.ts:40-47`, `adapter.ts:651-660` | silent-failure-hunter |
| I2 | `getInstallationTokenById` POST has no error logging; on failure `installationId` is lost before reaching outer handlers | `auth.ts:91-94` | silent-failure-hunter |
| I3 | Shell script `$path` lacks client-side validation (server-side regex is the only defense) | `scripts/git-credential-archon.sh:40` | code-reviewer (80/100) |
| I4 | Shell script swallows curl stderr on connection failure — unattended workflows surface "Authentication failed" with no root cause | `scripts/git-credential-archon.sh:39-40, 44-45` | silent-failure-hunter |
| I5 | PEM `\r\n` (Windows-edited file) not normalized for `PRIVATE_KEY_PATH` branch | `private-key.ts:18-44` | pr-test-analyzer |
| I6 | `auth.ts` file header claims `lookupCache: refreshed on 401` — false. Compounds C2. | `auth.ts:5` | comment-analyzer |
| I7 | Inline comment at `adapter.ts:888` references "PR-C" — will rot post-merge | `adapter.ts:888` | comment-analyzer |
| I8 | `credential-helper-install.ts:12-13` "Source-builds only for this PR" — will rot post-merge; convert to tracked issue | `credential-helper-install.ts:12-13` | comment-analyzer |

### Test coverage gaps

| ID | Gap | Crit. | Source |
|----|-----|------:|--------|
| T1 | `/internal/git-credential` endpoint has zero tests — path-parse regex regression could leak tokens to wrong installations | 9/10 | pr-test-analyzer |
| T2 | Dual-mode fail-fast guard untested (`hasGitHubApp && hasGitHubPat`) | 8/10 | pr-test-analyzer |
| T3 | `withTokenRefresh` retry-on-retry-401 propagation unproven | 8/10 | pr-test-analyzer |
| T4 | `scripts/git-credential-archon.sh` has no shellcheck or functional tests | 7/10 | pr-test-analyzer |
| T5 | PEM `\r\n` normalization not tested | 6/10 | pr-test-analyzer |
| T6 | 1h lookupCache TTL boundary not tested with fake timers (off-by-one risk) | 6/10 | pr-test-analyzer |

### Suggestions

- **Type design (rated 8/6/8/7):** `CachedInstallationToken.expiresAt` → rename to `expiresAtMs` for unit clarity. Validate `appId` and `slug` non-empty at `createGitHubAppAuthProvider` entry (consistent with `loadAppPrivateKey`'s eager validation). (type-design-analyzer)
- **Simplifications:** drop `octokitCache` — Octokit construction is cheap and `createAppAuth` already handles internal token refresh; `tokenCache` carries the real cache value for non-Octokit callers (clone embed, credential helper). Remove dead `if (!appId || !webhookSecret)` guards inside `hasGitHubApp`/`hasGitHubPat` branches — use `!` non-null assertion. Collapse the no-op try/catch in `ensureRepoReady` clone path. (code-simplifier)
- **API ergonomics:** consider `invalidateTokenForRepo(owner, repo)` collapsing `resolveInstallationId` + `invalidateToken` into one interface method — removes async from the 401 hot path and removes one method from `IGitHubAppAuthProvider`. (code-simplifier; aligns with C2 fix)
- **Docs:** `GITHUB_APP_SLUG` is documented as required in `.env.example:101` and `github-app-setup.md:83-93` but defaults to `'archon'` (`index.ts:364`). `github-app-setup.md:130-132` doesn't mention that the credential helper silently no-ops in compiled binary builds. `github.md:20` still implies PAT is the only path. (docs-impact)

### Strengths acknowledged

- Discriminated union (`GitHubAuth`) is exhaustively narrowed at all 5 callsites (`resolveOctokit`, `withTokenRefresh`, `botLogin`, `ensureRepoReady`, `handleWebhook` priming). No accidental fallback to "the other branch." (type-design-analyzer)
- Dual-mode fail-fast bootstrap (`index.ts:341-347`) is the correct enforcement of CLAUDE.md's Fail Fast principle — clear "pick one" error, no silent preference. (code-reviewer)
- `ensureRepoReady` clone path distinguishes `AppNotInstalledError` from generic auth failure and maps to user-actionable messages per auth mode. (silent-failure-hunter)
- `postComment` retry exhaustion log carries `attempt`, `maxRetries`, `wasRetryable`, `messageLength` — excellent observability. (silent-failure-hunter)
- All three webhook handlers (GitHub, Gitea, GitLab) attach `.catch()` with ERROR log — no silent rejection. (silent-failure-hunter)
- 401 retry mechanism is structurally bounded — the second call has no surrounding catch, so a second 401 propagates rather than looping. (code-reviewer)
- `private-key.ts` file header is honest about validation scope ("catches obviously-bad shapes; crypto validation deferred to `@octokit/auth-app`"). (comment-analyzer)

---

### Verdict — NEEDS FIXES

C1 and C2 are the load-bearing concerns:

- **C1** is a security defect that violates CLAUDE.md's Fail Fast principle ("Never silently broaden permissions or capabilities"). A misconfigured deployment leaks installation tokens to the network.
- **C2** is a correctness bug that causes silent 1h outages after a routine operational event (App reinstall).

C3 and C4 are observability defects that will burn debugging time on real production incidents. The test gaps T1 + T2 leave the riskiest new surface (the token-vending endpoint, the dual-mode guard) unverified — both are short tests to add and would catch regressions in the security-critical paths.

The structural design is sound: the discriminated union, the fail-fast on dual config, the two-level cache architecture, and the 401-evict-retry shape are all the right calls. Fixes are localized and don't require rework.

🤖 Generated by `/prp-core:prp-review-agents` (7 specialized agents)

Wirasm:
## Re-review after b35d9bd0 + 429997f3

Reviewed against the 4 critical + 8 important + 6 test-gap findings from the previous comment. Verified each fix against the actual code on disk, not just commit messages.

### Critical findings — all resolved

- **C1 (fail-fast on public bind):** `index.ts:766-780` now throws (not WARNs) when App mode is active and bind is non-loopback, with `ARCHON_ALLOW_INTERNAL_ON_PUBLIC_BIND=1` opt-out. CR6 also moved the guard above `Bun.serve` — the listening socket never opens on rejected config, so there's no window where `/internal/git-credential` is briefly reachable before the throw. Solid.
- **C2 (lookupCache eviction):** new `invalidateRepo(owner, repo)` at `auth.ts:183-207` evicts all three caches (token, lookup, octokit) in one call. Adapter uses it at `adapter.ts:167`. CR2 follow-up correctly added `octokitCache.delete()` — the per-installation Octokit holds its own `createAppAuth` token state internally, so without that eviction the "fresh" Octokit handed to the retry could keep serving the dead token from the SDK's private cache. Both `invalidateToken` and `invalidateRepo` cascade through all three caches now. Stronger than I suggested.
- **C3 (retry observability):** `adapter.ts:169-187` wraps `fn(fresh)` in try/catch with explicit ERROR log carrying `owner`, `repo`, `firstStatus`, `retryStatus`. Comment documents "Bound by design: no third retry" — addresses the original Q2 about infinite-loop risk as documented intent. Clean.
- **C4 (ERROR not WARN):** `index.ts:654` now logs `internal.git_credential_resolve_failed` at ERROR, with a comment explaining why ("live credential-vending failure…operators need this surfaced loudly"). Done.

### Important findings — all resolved

- **I1:** `installCredentialHelper` now returns `CredentialHelperInstallResult` discriminated union (`installed | skipped | failed`) — `credential-helper-install.ts:45-48`. Adapter logs the actual outcome instead of false-success in binary builds.
- **I2:** `getInstallationTokenById` wraps the POST in try/catch with ERROR log carrying `installationId` + HTTP `status` — `auth.ts:117-126`. Dead installations are now greppable.
- **I3:** Shell script has explicit `case "$path" in */*) ;; *) exit 0 ;; esac` at `git-credential-archon.sh:46-52`. Defense-in-depth in place.
- **I4:** Stderr diagnostics on every fall-through path: non-github host (line 38), malformed path (49), curl failure with captured stderr + exit code (66-67), empty token response with response length (77-78). Unattended workflows now see root cause.
- **I5:** PEM normalization now covers `\\n`, `\\r\\n`, and raw `\r\n` (file-loaded CRLF) — `private-key.ts:24, 35`. Tested in `auth.test.ts:129, 147`.
- **I6:** `auth.ts:5-9` header now accurately states `lookupCache` is "evicted on 401 via `invalidateRepo`."
- **I7:** `adapter.ts` comment no longer references "PR-C" (verified via grep).
- **I8:** `credential-helper-install.ts:11-16` header drops the "this PR" framing; now states the structural condition (source script must exist on disk for source builds).

### Test gaps — all closed

- **T1:** `parseGitCredentialPath` extracted to `github-auth-bootstrap.ts:57-73`, tested with 12 cases covering owner/repo, .git suffix, single-segment, three-segment, `..` traversal, **literal null-byte injection** (`'alice\x00bad/widgets'` — confirmed real null bytes in the test source), leading slash, leading dot, case preservation. The null-byte test is a substantively stronger security test than my original "validate $path" suggestion.
- **T2:** `selectGitHubAuthMode` extracted + tested with 8 cases including the dual-mode conflict path with actionable-message assertion.
- **T3:** `adapter.test.ts:1302` proves the second consecutive 401 propagates with exactly one `invalidateRepo` call (no infinite loop). Regression guard at line 1328 also asserts `invalidateRepo` is used, not the old `invalidate(id)` shape — locks in the C2 fix.
- **T5:** Both PEM CRLF paths tested: file-loaded `\r\n` (line 129) and inline literal `\\r\\n` (line 147).
- **T6:** lookupCache 1h TTL boundary tested at `auth.test.ts:323` using `spyOn(Date, 'now').mockImplementation` (CR5: replaced direct Date.now reassignment + eslint-disable with a lint-clean pattern). Tests "just inside" (59m59s) and "just outside" (1h+1ms) the boundary.
- **T4 (shellcheck):** Skipped per author's stated YAGNI — acceptable; the 30-line POSIX script with one HTTP-call branch is well within manual-review surface, especially now with the stderr-diagnostics paths + connect-timeout.

### CodeRabbit follow-ups — verified

- **CR1:** `.env.example` inline-comment relocation prevents dotenv from misreading `GITHUB_TOKEN=` value as `# Same as GH_TOKEN…`.
- **CR4:** `curl --connect-timeout 2 --max-time 5` at `git-credential-archon.sh:61` — git no longer blocks indefinitely when Archon isn't listening.
- **CR7:** `/internal/git-credential` request body validated via inline `gitCredentialRequestSchema = z.object({...})` + `safeParse` (`index.ts:632-640`) rather than manual type cast. Matches CLAUDE.md's "always Zod for request validation."

### Suggestions not adopted — agree these are YAGNI

- Dropping `octokitCache`: keep, especially now that it's correctly invalidated alongside the token cache. The double-eviction logic earns the cache's keep.
- Splitting `/internal/*` to a separate loopback-only listener: fail-fast bind check + reverse-proxy doc cover the threat model adequately.
- `!appId/!webhookSecret` non-null assertions: lint forbids `!` post-narrowing; defensive guards are the right call.

### One minor observation (not blocking)

The `gitCredentialRequestSchema` at `index.ts:632` is inline-declared inside the conditional, so it's re-created on server startup but never per-request — fine. Worth noting it could move to module scope alongside the imports for consistency with the rest of the file's schema usage pattern, but the comment explains why it's local. Up to you.

---

## Verdict — READY TO MERGE

All 4 critical findings resolved with stronger fixes than I suggested (octokitCache cascade, pre-listen guard placement, defense-in-depth at three layers — adapter / endpoint / shell). All 8 important findings resolved. 5 of 6 test gaps closed (T4 reasonably deferred). CodeRabbit's 8 follow-ups addressed. CI green (ubuntu, windows, docker-build all pass).

The structural design is unchanged from PR-B's original shape; the fixes are localized to the auth module, adapter, server bootstrap, and shell helper. Discriminated union remains exhaustively narrowed. Fail-fast principle now correctly applied at every load-bearing decision point.

🤖 Generated by `/prp-core:prp-review-agents` re-review pass

PR #1810, opened 2026-05-29
Fix: auto-apply Postgres schema on startup

### LINKED ISSUE #1796: team-foundation: auto-apply Postgres schema on startup (convergent now, Drizzle later)
## Context

PR-A's user-identity schema changes (#1783) added `remote_agent_users`, `remote_agent_user_identities`, and `user_id` columns on four existing tables. The SQLite path picked them up automatically via `migrateColumns()` in `packages/core/src/db/adapters/sqlite.ts`. The Postgres path **did not** — and the reason is more fundamental than "we forgot a migration."

Hit live on `archon.dynamous.ai` while dogfooding v0.4.1: the upgrade pulled clean, the container started healthy in App mode, but the first call to `POST /api/conversations` returned `500 Failed to create conversation` with the underlying error `column "user_id" of relation "remote_agent_conversations" does not exist`. Recovery required SSH + manually piping `migrations/000_combined.sql` into the Postgres container. ~30 minutes of debugging. This re-breaks on every schema-adding upgrade until fixed.

### Root cause (narrower and more damning than "missing migration")

**Postgres has no app-side schema management at all.** `PostgresAdapter` never touches schema. Today the Postgres schema is created entirely by the postgres image's `/docker-entrypoint-initdb.d/000_combined.sql` mount (`docker-compose.yml`), which by design runs **once on an empty data volume and never again**. So *any* upgrade that adds schema silently breaks until someone SSHes in and runs `psql`.

Worse: **external-DB operators get nothing, ever.** The compose comments explicitly invite Supabase/Neon. Those deployments have no init-dir mechanism — they're on manual `psql` from day one, upgrade or not.

Meanwhile **SQLite already does the right thing**: `SqliteAdapter` runs `createSchema()` (all `IF NOT EXISTS`) + `migrateColumns()` on *every* boot, in its constructor. The two backends already share a design philosophy — *declare the desired schema, converge to it idempotently on startup.* Postgres is simply missing the "on startup" half.

## Problem statement

Postgres schema should converge automatically on app startup, the way SQLite already does. Operators upgrading from any v0.3.x/v0.4.x release should never need a manual `psql` step — and this must work identically for the bundled Postgres container **and** external managed Postgres (Supabase/Neon).

## Decision

**Finish the convergent pattern for Postgres now (the simple path). Adopt Drizzle Kit as the deliberate migration foundation later, when team features make real versioning worth it.**

We are **not** building a bespoke versioned migration runner. Given the team-foundation trajectory, a custom runner would be guaranteed throwaway scaffolding: enough work to be a real cost, but something we'd rip out and re-map into Drizzle within months. Skip the middle.

### The decisive fact

`migrations/000_combined.sql` **is already a fully idempotent, convergent schema definition.** Every statement is `CREATE TABLE IF NOT EXISTS` / `ADD COLUMN IF NOT EXISTS` / `CREATE INDEX IF NOT EXISTS` / `DROP … IF EXISTS`, under an explicit section headed *"Idempotent ALTER statements for upgrading existing databases."* The docs already state it: *"`000_combined.sql` is idempotent — re-running it picks up any new tables and `ADD COLUMN IF NOT EXISTS` statements … without disturbing existing rows"* (`packages/docs-web/.../database.md`). It is the Postgres twin of SQLite's `createSchema()`. The only thing missing is **calling it on startup.**

### Options reconsidered (self-hosted, dual-dialect, team-foundation lens)

| Option | What | Verdict |
|---|---|---|
| **A — converge on boot (chosen now)** | Apply the already-idempotent `000_combined.sql` on Postgres startup, mirroring SQLite's `createSchema()`. ~20–30 lines. | **Chosen.** Fixes the bundled-container upgrade case *and* external DBs in one stroke. Unifies the mental model across both backends. Zero new concepts, no footguns. Not throwaway — becomes the baseline Drizzle introspects later. |
| B — bespoke versioned runner (`archon_schema_migrations` + sha256 drift gate + bootstrap heuristic) | Read incremental files, track versions, abort on sha drift, back-fill version table for existing deploys. | **Rejected.** Wrong size of change. The sha256 **abort-on-startup** gate is a severe self-hosted footgun — a line-ending diff, a cherry-picked hotfix, or a re-tagged release bricks the *entire app's boot*, which is worse than the one-endpoint 500 we're fixing. The `to_regclass` bootstrap heuristic is fragile one-shot code that must classify every schema age in the wild. And it perpetuates 2–3 sources of truth. With Drizzle as the known endgame, all of this is throwaway. |
| C — Drizzle Kit (chosen **later**) | TS schema-as-code, ordered generated migrations, real journal/audit; pairs with #1787 (drizzle-zod could unify DB schema + Zod validation). | **Deferred, intended foundation.** This is the right long-term tool as team features land. Adopting it from a *converged* baseline is clean. (One thing to settle at adoption: dual-dialect — SQLite default + Postgres for teams — means dialect-specific schema definitions, so it's best adopted as the migration tool while keeping the existing `IDatabase`/raw-SQL query layer.) |

### Postgres converges on startup

1. Add an async `init()` to `PostgresAdapter` (the constructor can't `await`; SQLite's `initSchema()` is sync via `bun:sqlite`). Call it once from the DB bootstrap before first use, guarded by a one-time promise so it runs exactly once per process.
2. `init()` reads and executes `000_combined.sql` — from disk in source builds; embedded for binary builds (same bundling pattern as default commands/workflows).
3. Wrap in a transaction. Take a Postgres advisory lock (`pg_advisory_xact_lock`) around it — cheap insurance against concurrent boots as the team story grows (no clustering today, but free safety).
4. On any DDL failure: log `db.pg_schema_init_failed` with the failing statement + underlying `pg` error, and **abort startup** — fail fast on a *real* error, never on a cosmetic file hash.
5. Optionally write a single non-gating `archon_schema_version` row (app version, `applied_at`) **after** success, for observability only — never read for control flow. (`information_schema` remains the real audit surface.)

### Fresh install, existing deploy, external DB — all one path

Because the script is convergent and idempotent, there is exactly one code path:
- **Fresh DB** → every statement creates.
- **Existing deploy** (dynamous.ai, already manually patched) → every statement is a no-op.
- **External Supabase/Neon** → same `init()` runs against it on boot; no init-dir needed.

No bootstrap detection, no version back-fill heuristics, no per-deploy special-casing.

### Cleanup

- Remove the `000_combined.sql` init-dir mount from `docker-compose.yml` (or leave it harmlessly — the app now owns convergence).
- Remove the manual `psql` step from `CLAUDE.md` and the deployment/reference docs (`docker.md`, `cloud.md`, `local.md`, `database.md`).

### Honest limits of convergence (and when they bite)

Idempotent convergence handles all current and foreseeable near-term changes (new tables, new nullable columns, new indexes, guarded drops). It does **not** cleanly handle non-idempotent transforms — column type changes, adding `NOT NULL` to a populated column, renames, destructive backfills. SQLite already punts on these today, and Postgres handles more of them idempotently than SQLite (it has `ADD COLUMN IF NOT EXISTS` / `DROP COLUMN IF EXISTS` / DO-block guards, all already used in migrations 006/007). **The day a true one-shot data migration becomes routine is the trigger to land Phase 2 (Drizzle), not a custom runner.**

## Acceptance criteria

- [ ] `PostgresAdapter` gains an idempotent, once-per-process `init()` that applies `000_combined.sql`
- [ ] `init()` is called from the DB bootstrap before any other Postgres work, for both bundled and external Postgres
- [ ] Runs inside a transaction + advisory lock; a DDL failure aborts startup with the failing statement + underlying error (`db.pg_schema_init_failed`)
- [ ] Successful init logs a single summary line; **no** sha-drift gate, **no** bootstrap/`to_regclass` heuristic
- [ ] Works correctly against: a fresh empty DB, an existing already-patched deploy, and an external managed Postgres
- [ ] Manual `psql` step removed from `CLAUDE.md` and deployment/reference docs
- [ ] SQLite `migrateColumns()` path unchanged (a follow-up to unify both backends under Phase 2 is reasonable but out of scope here)
- [ ] Tests: fresh-DB convergence, re-run-is-no-op (idempotency), and abort-on-DDL-error

## Effort

~Half a day. The convergence script already exists and is already idempotent; the work is the async init plumbing, the advisory lock, and the bundling-for-binary path.

## Phase 2 — Drizzle Kit (separate, later, deliberate)

When team features make real versioning, ordered migrations, and an audit trail worth it — or when the first routine data migration appears — adopt **Drizzle Kit** as the migration tool:

- Introspect/baseline against the converged schema this issue ships (clean starting point — the convergent `init()` simply stops being called).
- Pair with #1787 (Zod unification): evaluate `drizzle-zod` to make Drizzle the single source of truth for both DB schema and validation/types.
- Settle the dual-dialect approach (SQLite + Postgres) at adoption time; lean toward Drizzle-as-migration-tool while keeping the existing `IDatabase` query layer.
- File as its own issue at that point. **This issue does not block it and does not pre-build any of it.**

## Related

- #1783 (PR-A) — the schema work whose Postgres upgrade trap motivated this
- #1787 — Zod unification; the natural pairing for the Phase 2 Drizzle move
- `CLAUDE.md` — has the manual `psql` caveat this issue eliminates, and the single-developer framing that the team-foundation work will revise

Wirasm:
## Summary

- **Problem**: `PostgresAdapter` never applied the schema on startup, so any upgrade that adds columns (e.g. `user_id` in v0.4.1) silently broke Postgres deployments with 500 errors until someone manually ran `psql`.
- **Why it matters**: Every Postgres user (Docker + external managed like Supabase/Neon) was vulnerable to production outages on every schema-adding upgrade. Affected archon.dynamous.ai in production, requiring a 30-min SSH recovery.
- **What changed**: `PostgresAdapter` now calls `initSchema()` on construction, running `migrations/000_combined.sql` inside an advisory-lock transaction. The SQL is already fully idempotent (`CREATE TABLE IF NOT EXISTS`, `ADD COLUMN IF NOT EXISTS`), so re-runs are no-ops.
- **What did not change**: SQLite path is untouched. `docker-compose.yml` initdb mount is left in place (harmless no-op on subsequent boots). No Drizzle migration framework added — that's a separate Phase 2 effort.

### Before

```
User upgrades Archon with new schema-adding version
  │
  ▼
App starts, PostgresAdapter constructor creates Pool only (no schema init)
  │
  ▼
User sends POST /api/conversations
  │
  ▼
INSERT hits column "user_id" that doesn't exist in DB
  │
  ▼
500 — "column 'user_id' of relation 'remote_agent_conversations' does not exist"
  │
  ▼
User must SSH in and manually run: psql $DATABASE_URL < migrations/000_combined.sql
```

### After

```
User upgrades Archon with new schema-adding version
  │
  ▼
App starts, PostgresAdapter constructor fires initSchema() [NEW]
  ├─ acquires pg_advisory_xact_lock(1796) — serializes concurrent boots
  ├─ runs 000_combined.sql (idempotent: all IF NOT EXISTS)
  └─ releases lock; logs db.pg_schema_init_completed
  │
  ▼
User sends POST /api/conversations — succeeds, schema is current
```

### Before

```
connection.ts
  └─▶ PostgresAdapter(connectionString)
        └─ Pool creation only
        └─ pool.on('error', ...)
        └─ (no schema init)

migrations/000_combined.sql ─── manual psql only
```

### After

```
connection.ts
  └─▶ PostgresAdapter(connectionString)
        └─ Pool creation
        └─ pool.on('error', ...)
        └─ this.schemaInitPromise = this.initSchema()  [NEW ~]
              └─▶ bundled-schema.ts [NEW +]
                    ├─ BUNDLED_IS_BINARY? → bundled-schema.generated.ts [NEW +]
                    └─ source build? → reads migrations/000_combined.sql from disk
              └─ pg_advisory_xact_lock(1796) + run SQL + COMMIT

query() / withTransaction()
  └─ await this.schemaInitPromise  [NEW ~]  (no-op after first resolution)
```

**Connection inventory**:

| `postgres.ts` | `bundled-schema.ts` | **new** | Schema SQL facade |
| `bundled-schema.ts` | `bundled-schema.generated.ts` | **new** | Binary embed path |
| `bundled-schema.ts` | `migrations/000_combined.sql` | **new** | Source build disk path |
| `postgres.ts` → `query()` | `schemaInitPromise` | **modified** | Awaits init before first query |
| `postgres.ts` → `withTransaction()` | `schemaInitPromise` | **modified** | Awaits init before first transaction |
| `scripts/generate-bundled-schema.ts` | `bundled-schema.generated.ts` | **new** | Generator script |
| `package.json` `validate` | `check:bundled-schema` | **new** | CI stale-check gate |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `core`, `ci`, `docs`
- Module: `core:db`

## Change Metadata

- Change type: `bug`
- Primary scope: `core`

## Linked Issue

- Closes #1796

```bash
bun run validate
```

All checks passed:

| Check | Result |
|-------|--------|
| `check:bundled` | ✅ Pass |
| `check:bundled-skill` | ✅ Pass |
| `check:bundled-schema` | ✅ Pass |
| `type-check` (10 packages) | ✅ Pass |
| `lint --max-warnings 0` | ✅ Pass |
| `format:check` | ✅ Pass |
| Full test suite (~4,049 tests) | ✅ Pass — 0 failures |

New `postgres.test.ts` tests cover: one-shot init, advisory-lock SQL ordering, DDL failure propagation to subsequent queries, and `withTransaction` gate.

- Evidence provided: full test suite run, all packages.
- No commands intentionally skipped.

## Security Impact (required)

- File system access scope changed? No (source-build disk read is within the repo's own `migrations/` directory; binary build uses embedded string with no disk access)

## Compatibility / Migration

- Backward compatible? Yes — idempotent SQL; existing Postgres DBs with current schema are unaffected.
- Config/env changes? No
- Database migration needed? No — this PR IS the automatic migration mechanism. No manual action required.

- Verified scenarios: advisory-lock serialization logic reviewed; `initSchema` promise gate reviewed for correct no-op behavior after first resolution.
- Edge cases checked: DDL failure path (test verifies that subsequent queries re-throw the init error); concurrent-boot scenario handled by `pg_advisory_xact_lock`.
- What was not verified: live Postgres instance upgrade test (no Docker/Postgres available in the worktree environment).

- Affected subsystems: Postgres startup only. SQLite, all platform adapters, workflows, and CLI are unaffected.
- Potential unintended effects: If `000_combined.sql` contains a non-idempotent statement added accidentally in a future PR, it would cause init to fail on the second boot. The `check:bundled-schema` CI gate makes SQL changes visible.
- Guardrails: `db.pg_schema_init_completed` / `db.pg_schema_init_failed` log events for observability.

## Rollback Plan (required)

- Fast rollback: revert to the previous release tag; schema columns added by `000_combined.sql` are additive and backward-compatible, so a rollback leaves harmless extra columns.
- Feature flags: none — schema init is unconditional when `DATABASE_URL` is set.
- Observable failure symptoms: `db.pg_schema_init_failed` fatal log at startup; all DB queries throw the init error rather than a silent 500.

## Risks and Mitigations

- **Risk**: Advisory-lock key `1796` (issue number) could collide with another application sharing the same Postgres cluster.
  - **Mitigation**: Low probability in practice (most deployments are single-tenant). If it becomes an issue, the key is a one-line change to a hash of `archon_schema_init`.

## Release Notes

* **New Features**
  * Database schema now automatically initializes on application startup, eliminating manual migration steps during deployment.
  * Added automated schema validation to CI/CD pipeline.

* **Documentation**
  * Updated deployment guides to reflect automatic schema initialization behavior.
  * Removed manual database migration instructions.

* **Chores**
  * Added schema consistency validation scripts to development workflow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1810?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Wirasm:
## Comprehensive PR Review — 5 Agents

**PR**: #1810 — Fix: auto-apply Postgres schema on startup
**Agents**: code-review · error-handling · test-coverage · comment-quality · docs-impact

---

## Summary

Well-executed fix overall. The advisory-lock + promise-gate design is correct for production Postgres, the generator mirrors the bundled-defaults pattern faithfully, and the four `initSchema()` tests cover all the right invariants. Two docs gaps and one error-handling gap need addressing before merge.

| Severity | Count |
|----------|-------|
| 🟠 HIGH | 2 |
| 🟡 MEDIUM | 1 |
| 🟢 LOW | 7 |

---

## 🟠 HIGH Issues (Should Fix Before Merge)

### H1: `local.md` still has the manual psql migration step

📍 `packages/docs-web/src/content/docs/deployment/local.md:120–126`

The "Database Migration (First Time)" subsection in the "Docker with Remote PostgreSQL" section still tells users to run `psql $DATABASE_URL < migrations/000_combined.sql` — the exact step this PR makes obsolete. The same file already contradicts this at line 65 ("No manual `psql` step is needed"). All other deployment docs (`cloud.md`, `docker.md`, `database.md`) were correctly updated; `local.md` was missed.

<details>
<summary>View suggested fix</summary>

Replace lines 120–126 with the `cloud.md` Section 5 pattern:

```markdown
### Database Setup

No manual migration is needed — the app applies `migrations/000_combined.sql` inside an
advisory-lock transaction on first connection. To confirm the tables landed after starting the app:

```bash
psql $DATABASE_URL -c "\dt"
# Should show: remote_agent_codebases, remote_agent_conversations, ...
```
```

</details>

---

### H2: CLAUDE.md validate description is one check behind

📍 `CLAUDE.md:156` and `CLAUDE.md:750`

`bun run validate` now runs 7 checks (added `check:bundled-schema`), but CLAUDE.md still says "six" and line 750 doesn't mention `check:bundled-schema` or when to run `generate:bundled-schema`. Contributors editing `migrations/000_combined.sql` will hit a CI failure with no hint about the regenerate step.

<details>
<summary>View suggested fix</summary>

```markdown
# Line 156
This runs `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, type-check, lint, format check, and tests. All seven must pass for CI to succeed.

# Line 750
**After adding, removing, or editing a default file, run `bun run generate:bundled`** to refresh the embedded bundle. After editing `migrations/000_combined.sql`, run `bun run generate:bundled-schema` to keep the embedded schema in sync. `bun run validate` (and CI) run `check:bundled`, `check:bundled-skill`, and `check:bundled-schema` and will fail loudly if any generated file is stale.
```

</details>

---

## 🟡 MEDIUM Issues (Should Fix Before Merge)

### M1: Connection failure in `initSchema()` produces no log entry

📍 `packages/core/src/db/adapters/postgres.ts:46–48`

`getSchemaSQL()` and `this.pool.connect()` are called **outside** the `try` block. If either throws (DB unreachable, connect timeout, bad `DATABASE_URL`, missing SQL file), the error propagates correctly and gates all subsequent queries — but no `db.pg_schema_init_failed` log fires. On-call has to trace rejection chains instead of grepping for a single structured event.

Examples of silently-unlogged failures:
- `ECONNREFUSED 127.0.0.1:5432` — DB not up at container boot
- `Connection terminated unexpectedly` — pool connect timeout (10s)
- `password authentication failed` — bad `DATABASE_URL`

<details>
<summary>View suggested fix</summary>

```typescript
private async initSchema(): Promise<void> {
  let client: Awaited<ReturnType<Pool['connect']>> | undefined;
  try {
    const sql = getSchemaSQL();
    client = await this.pool.connect();
    await client.query('BEGIN');
    await client.query('SELECT pg_advisory_xact_lock(1796)');
    await client.query(sql);
    await client.query('COMMIT');
    getLog().info('db.postgres_schema_init_completed');
  } catch (e) {
    if (client) {
      try {
        await client.query('ROLLBACK');
      } catch (rollbackError) {
        getLog().error({ err: rollbackError as Error }, 'db.postgres_schema_init_rollback_failed');
      }
    }
    const err = e as Error;
    getLog().fatal({ err }, 'db.postgres_schema_init_failed');
    throw err;
  } finally {
    client?.release();
  }
}
```

The `if (client)` guard prevents a double-error if `pool.connect()` itself failed (no transaction open to ROLLBACK). Also note the log event names changed to `db.postgres_*` — see LOW C1 below.

</details>

---

## 🟢 LOW Issues (For Consideration)

<details>
<summary>View 7 low-priority findings</summary>

**C1 — Log event prefix inconsistency** (`postgres.ts:59,64,67`)
New events use `db.pg_*`; existing events in the same file use `db.postgres_*` (matching `createLogger('db.postgres')`). Log aggregation queries on `db.postgres_*` silently miss the schema init events. Fix: rename to `db.postgres_schema_init_completed/rollback_failed/failed`.

**C2 — `bundled-schema.ts` binary branch untested** (`bundled-schema.ts:17–19`)
The `BUNDLED_IS_BINARY=true` path is completely dark. A bug where `BUNDLED_SCHEMA_SQL` is accidentally empty would apply an empty schema and produce no error. Fix: add a `bundled-schema.test.ts` in a separate file (Bun mock isolation rules apply) that mocks `@archon/paths` to `{ BUNDLED_IS_BINARY: true }` and asserts the constant is non-empty and contains expected table markers.

**C3 — `client.release()` not asserted in DDL failure test** (`postgres.test.ts:410–435`)
The `finally` block's resource-safety guarantee isn't locked down by an assertion. Fix: add `let released = false`, `release: () => { released = true }` to the mock, and `expect(released).toBe(true)`.

**C4 — Issue reference `(#1796)` in test section divider** (`postgres.test.ts:75`)
Violates CLAUDE.md "don't reference the current task" rule. Fix: remove `(#1796)` from the divider comment.

**C5 — Issue reference `(issue #1796)` in generator script JSDoc** (`generate-bundled-schema.ts:8`)
Same rule violation. Fix: drop `(issue #1796)` from the `Why:` line — sentence reads fine without it.

**C6 — "implicitly" is imprecise in test comment** (`postgres.test.ts:91`)
The await inside `query()` is explicit; "implicitly" is only true from the caller's perspective. Suggested: `// Triggers schema init as a side-effect — callers don't call initSchema() directly`.

**C7 — `as Error` cast in catch blocks** (`postgres.ts:66`)
Pre-existing pattern — not introduced by this PR. Optional hygiene: `const err = e instanceof Error ? e : new Error(String(e))`, consistent with `generate-bundled-schema.ts:72`.

</details>

---

## ✅ What's Good

- **Advisory-lock design is correct** — `pg_advisory_xact_lock(1796)` serializes concurrent container boots, a real production hazard most implementations miss.
- **Promise-gate is minimal and correct** — `schemaInitPromise` awaited in both `query()` and `withTransaction()`. No extra state machine needed.
- **Generator script is a faithful clone of the bundled-defaults pattern** — same exit codes, CRLF normalization, `--check` comparison logic. Easy to maintain alongside `generate-bundled-defaults.ts`.
- **Four `initSchema()` tests cover all the right invariants** — fresh-boot ordering, idempotency, DDL failure propagation (settled rejection), and `withTransaction` gate sequencing. Index-based ordering assertions go beyond smoke testing.
- **`cloud.md`, `docker.md`, `database.md` are accurately updated** — manual `psql` steps fully removed, `\dt` verification matches all 10 tables.
- **`schemaInitPromise` field comment is exemplary** — explains the gate pattern, the race condition it prevents, and the no-op-after-settle behavior in three lines.

---

## Next Steps

1. Fix H1 (`local.md`) and H2 (`CLAUDE.md`) — both are small targeted edits
2. Fix M1 (`initSchema()` try-block scope) — moves two lines inside `try`, adds `if (client)` guard
3. Consider LOW fixes — log prefix rename and comment cleanups are zero-risk
4. Mark PR ready for review once HIGH + MEDIUM items are addressed

---

*Reviewed by Archon comprehensive-pr-review workflow — 5 specialized agents*
*Artifacts: `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/77cd4c2ec8235a34e9cd86199f5478bd/review/`*

Wirasm:

**Status**: COMPLETE
**Pushed**: ✅ Changes pushed to `archon/task-fix-issue-1796`
**Commit**: 66def323

---

### Fixes Applied (11 total)

| Severity | Count |
|----------|-------|
| 🔴 CRITICAL | 0 |
| 🟠 HIGH | 2 |
| 🟡 MEDIUM | 1 |
| 🟢 LOW | 8 |

<details>

- ✅ **CLAUDE.md validate count stale** (`CLAUDE.md:156,750`) — Updated "six" → "seven", added `check:bundled-schema` to both lines, added `generate:bundled-schema` guidance to Defaults section
- ✅ **local.md manual psql step** (`local.md:120–126`) — Replaced "Database Migration (First Time)" with "Database Setup" matching the cloud.md advisory-lock auto-apply pattern
- ✅ **Connection failures in `initSchema()` produce no log** (`postgres.ts:46–71`) — Moved `getSchemaSQL()` and `pool.connect()` inside the try block; all init failures (ECONNREFUSED, auth, pool timeout, ENOENT) now emit `db.postgres_schema_init_failed`; added `if (client)` guard before ROLLBACK
- ✅ **Log prefix `db.pg_*` vs `db.postgres_*`** (`postgres.ts:59,64,67`) — Renamed all three new events to `db.postgres_schema_init_{completed,rollback_failed,failed}`
- ✅ **`as Error` unsafe cast** (`postgres.ts:66,64`) — Replaced with `instanceof Error ? e : new Error(String(e))` guard
- ✅ **`PoolClient` type for `let client`** (`postgres.ts`) — Added `import type { PoolClient } from 'pg'` for proper typing
- ✅ **binary-mode branch untested** (new `bundled-schema.test.ts`) — Created isolated test file with `BUNDLED_IS_BINARY=true` mock; wired into core package.json test script
- ✅ **`client.release()` not asserted in DDL failure test** (`postgres.test.ts:410–435`) — Added `released` flag and `expect(released).toBe(true)`
- ✅ **Issue reference `(#1796)` in test divider** (`postgres.test.ts:74–76`) — Removed
- ✅ **Issue reference `(issue #1796)` in JSDoc** (`generate-bundled-schema.ts:8`) — Removed
- ✅ **"implicitly" imprecise comment** (`postgres.test.ts:91`) — Rephrased to "Triggers schema init as a side-effect — callers don't call initSchema() directly"

</details>

---

### Tests Added

- `packages/core/src/db/bundled-schema.test.ts` — 2 tests covering `getSchemaSQL()` binary-mode branch (`BUNDLED_IS_BINARY=true`): returns embedded constant, constant is non-empty with expected table markers

---

### Skipped

---

### Validation

✅ Type check | ✅ Lint | ✅ Format | ✅ Tests (0 fail)

---

*Self-fix by Archon · aggressive mode · fixes pushed to `archon/task-fix-issue-1796`*

PR #1823, opened 2026-05-29
feat: per-user GitHub identity via device flow + interim web auth (PR-C)

Wirasm:
## Summary

- **Problem:** After PR-A (user_id plumbing) and PR-B (GitHub App for the bot), there was no way for a human to connect their own GitHub identity — comments/commits/pushes all went through the bot, the web UI was unauthenticated, and an unconnected user's workflow `git push` silently used the shared org token.
- **Why it matters:** The attribution payoff of the team-foundation PRD — "Alice sees her own avatar on PR comments/commits, and the team can tell who triggered any run."
- **What changed:** Full PR-C — encrypted per-user token store + device flow, connect surfaces (Slack/CLI/Web), `requires:[github]` invocation gate, per-user token routing for workflow subprocesses and PR comments, `git config user.email` on worktrees, and interim `X-Archon-User` web auth.
- **What did NOT change (scope boundary):** Everything is gated on **per-user mode** (`GITHUB_APP_ID` + `TOKEN_ENCRYPTION_KEY`). Solo PAT installs and App-for-bot-only installs are behaviorally unchanged. No ACLs, no long-term web auth (interim header only), no Keycloak.

### Before

```
Connect:   (no path anywhere — a teammate cannot link GitHub)
Web UI:    open; every web row user_id = NULL
Workflow:  gh / git push  ─►  shared bot/org installation token (any user)
Commit:    worktree has no user.email  ─►  ambient identity
PR reply:  archon[bot] for every teammate
```

### After

```
Connect (Slack / CLI / Web):
  user ─► "connect github" ─► github.com/login/device  (code ABCD-1234)
       ─► authorize ─► /user ─► [encrypt+store tokens] + cache name/email

Workflow `requires:[github]`, UNCONNECTED user:
  ─► *HARD FAIL at invocation* (no worktree, no clone, no AI cost)

Workflow, CONNECTED user:
  worktree ─► git config user.email = <id>+<login>@users.noreply.github.com
  subprocess env GH_TOKEN = *user's* token   (unconnected + no fallback ─► scrubbed)

PR reply (connected originator):  *user's avatar*   (else archon[bot])
Web UI behind reverse proxy:  X-Archon-User: alice ─► rows attributed to alice
```

### After (new modules marked [+], modified [~])

```
@archon/core
  utils/token-crypto.ts                 [+]  AES-256-GCM (cherry-pick #1774)
  github-auth/config.ts                 [+]  isPerUserGitHubEnabled / device cfg / boot assert
  github-auth/device-flow.ts            [+]  start / poll / pollOnce / refresh / fetchUser
  github-auth/connect-service.ts        [+]  connect (blocking) + persistGithubConnection
  db/user-github-token-store.ts         [+]  save / get / decrypt-refresh / delete / no-reply email
  db/users.ts                           [~]  updateUserGithubProfile, linkGithubIdentity
@archon/workflows
  utils/workflow-requirements.ts        [+]  requires:[github] gate
  utils/github-token-policy.ts          [+]  per-user token overrides (cherry-pick #1774)
  executor.ts / deps.ts                 [~]  merge per-user token into subprocess env
@archon/isolation  types/resolver/worktree  [~]  gitIdentity → git config user.email
@archon/adapters   forge/github         [~]  per-user comment token; chat/slack connect cmd
@archon/server     routes/api.ts        [~]  device-flow endpoints + X-Archon-User attribution
                   index.ts             [~]  boot key validation + getUserToken wiring
@archon/cli        commands/auth.ts     [+]  archon auth github
@archon/web        SettingsPage / lib/api[~]  Connect GitHub card
```

**Connection inventory:**

| orchestrator | workflow-requirements (gate) | **new** | before isolation |
| executor | github-token-policy | **new** | per-user env override |
| store-adapter | db/user-github-token-store | **new** | `getUserGithubToken` deps hook |
| resolver → worktree | gitIdentity | **new** | `git config user.email` |
| github adapter | getUserToken (server-injected) | **new** | per-user comments |
| server api | device-flow + token store | **new** | connect endpoints + status |
| solo PAT install | any of the above | unchanged | per-user mode off → no-op |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: XL`
- Scope: `core|workflows|isolation|adapters|server|web|cli|docs`
- Module: `core:github-auth`, `workflows:executor`, `isolation:worktree`, `adapters:github`, `server:routes`

## Change Metadata

- Change type: `feature`
- Primary scope: `multi`

## Linked Issue

- Related #1774 (Keycloak donor — `token-crypto` + `github-token-policy` cherry-picked; close after this lands)
- Depends on #1788 (PR-B) and #1783 (PR-A), both merged
- Part of PRD Phase 3 (`.claude/PRPs/prds/github-app-and-user-identity.prd.md`)

```bash
bun run validate   # → exit 0 (all 6 checks)
```

- `bun run validate` exits 0: check:bundled, check:bundled-skill, check:bundled-schema, type-check (8 packages), lint (`--max-warnings 0`), format:check, full per-package test suite.
- New unit tests: token-crypto (10), device-flow (10), token store (10), workflow-requirements (3), github-token-policy (8); plus the review-added config/connect-service suites. Fresh in-memory SQLite confirms `remote_agent_user_github_tokens` is created.
- Intentionally skipped: live api.github.com (mocked per PRD Q7); a dedicated `api.auth.test.ts` (route wiring is thin over already-unit-tested core; heavy app-construction mock boilerplate) — flagged as a follow-up.

## Security Impact (required)

- New permissions/capabilities? **Yes, gated** — per-user token routing + `X-Archon-User` trust, active only in per-user mode.
- New external network calls? **Yes** — github.com device/token endpoints + `api.github.com/user`, only on the connect/refresh path.
- Secrets/tokens handling changed? **Yes** — per-user GitHub tokens stored AES-256-GCM-encrypted (`TOKEN_ENCRYPTION_KEY`), never logged; refresh-on-read serialized per-user (no double-consume of the single-use refresh token).
- File system access scope changed? **No.**
- Risk/mitigation: unconnected users' workflow tokens are **scrubbed by default** (closes the #1774 cross-user org-token leak; opt back in via `ARCHON_ALLOW_ORG_GITHUB_TOKEN_FALLBACK`). `X-Archon-User` is only safe behind a reverse proxy with loopback bind (documented; absent header → NULL, never elevated). Boot fails fast on a malformed `TOKEN_ENCRYPTION_KEY`.

## Compatibility / Migration

- Backward compatible? **Yes** — solo PAT and App-for-bot-only installs unchanged.
- Config/env changes? **Yes (opt-in):** `GITHUB_APP_CLIENT_ID`, `TOKEN_ENCRYPTION_KEY`, `ARCHON_ALLOW_ORG_GITHUB_TOKEN_FALLBACK`, `ARCHON_WEB_AUTH_HEADER` (all documented in `.env.example`).
- Database migration needed? **Yes** — additive `remote_agent_user_github_tokens` (PG auto-applies via bundled schema on startup; SQLite via `createSchema()`). No backfill, no changes to existing tables.
- Upgrade steps: deploy; to enable per-user, set `GITHUB_APP_CLIENT_ID` + `TOKEN_ENCRYPTION_KEY` and enable Device Flow on the App.

- Verified: full `bun run validate`; unit suites for crypto/device-flow/token-store/policy/gate; fresh SQLite builds the new table; type-check across all packages.
- Not verified beyond CI: a real end-to-end device flow against github.com (mocked by design); live per-user PR-comment + commit attribution (manual e2e per the PRD success criteria — see the PR thread for the setup checklist).

- Affected subsystems: GitHub adapter (comment auth), workflow executor (subprocess env), isolation worktree creation (git config), web API (3 message/run handlers + new endpoints), CLI (new command), Slack adapter (new subcommand), web Settings.
- Potential unintended effects: in per-user mode, an unconnected user's `gh`/`git` inside a workflow now fails instead of using the org token (intended; opt-out flag exists). PR comments by connected users change author from `archon[bot]` to the user.
- Guardrails: every path is gated on `isPerUserGitHubEnabled()`; conservative fallbacks (user-token 401 → installation; resolution failure → undefined → inherit env); structured `*_failed`/`*_stored`/`*_refreshed` logs for observability.

## Rollback Plan (required)

- Fast path: revert the PR-C commits; the new table is additive/unreferenced when per-user mode is off.
- Kill switch (no revert needed): unset `TOKEN_ENCRYPTION_KEY` (or `GITHUB_APP_CLIENT_ID`) → `isPerUserGitHubEnabled()` false → every per-user path no-ops; App reverts to bot-only.
- Observable failure symptoms: `user_github_token.refresh_failed` / `*.expired_no_refresh` warn logs; `workflow.requirement_unmet` info logs.

## Risks and Mitigations

- Risk: PR-comment attribution shows AI-authored text under the human's avatar.
  - Mitigation: matches the PRD success criterion; conservative bot fallback on any user-token failure. Trivial to drop (G7) if undesired.
- Risk: refresh-token race across processes.
  - Mitigation: in-process per-user mutex + re-read-on-failure; cross-process locking deferred (single-instance VPS target).
- Risk: deviations from plan — App-without-key → WARN+disable (not hard-fail) so App-for-bot-only stays valid; CLI identity via `ARCHON_USER_ID`/`$USER`; TS filenames `-row`/`-store` (local secret-guard hook); `resolveWebUserId` helper vs Hono middleware.
  - Mitigation: each documented in the report + commit messages; behavior is backward-compatible.

## Summary by CodeRabbit

* **New Features**
  * Connect/disconnect GitHub via device flow from Settings UI, CLI, and Slack; per-user GitHub tokens are encrypted at rest, used for comment attribution, workflow auth, and commit identity stamping.

* **Tests**
  * Extensive new tests covering device flow, token encryption/refresh, identity linking, orchestration, and edge cases.

* **Documentation**
  * Docs, README and .env examples updated for per-user GitHub, device-flow endpoints, and schema changes.


Wirasm:
## PR Review Summary — Multi-Agent (PR-C foundation: per-user GitHub token store + device flow)

Reviewed `feat/per-user-github-tokens` vs `dev` with 7 specialized agents (code, docs, tests, comments, errors, types, simplify). The PR is **additive and dormant** (no runtime callers) — zero behavior change for existing installs — which appropriately lowers the urgency of most findings. **No Critical blockers.** A handful of cheap, worthwhile fixes are listed below, plus three agent findings I checked and **rejected** as incorrect.

### Important Issues (8 found)

| I1 | errors | **Post-refresh DB write failure is mislabeled and loses the new token.** The `try` wraps both `refreshUserToken` *and* `saveUserGithubToken`. If the refresh succeeds but the DB save throws, the catch logs `refresh_failed` (wrong — refresh worked), discards the freshly issued token, and re-reads the stale row. Since refresh tokens are single-use, the stored refresh token is now spent → the user's token is permanently unusable on the next read. Move the save outside the refresh `try` (or narrow the catch to only the network call). | `db/user-github-token-store.ts:139-169` |
| I2 | code | **`linkGithubIdentity` lacks UNIQUE-violation recovery.** It does check-then-INSERT with no guard, unlike the sibling `findOrCreateUserByPlatformIdentity` (which uses `isUniqueViolation` + re-SELECT). A concurrent double-connect (double-click "Connect", retry) surfaces a raw `23505` instead of `GithubIdentityConflictError`. Reuse the existing recovery pattern. | `db/users.ts:254-273` |
| I3 | tests | **The PR's core security invariant is untested.** `GithubIdentityConflictError` (one user must not steal another's GitHub identity) has zero tests across its three branches. `users.test.ts` was not extended for `linkGithubIdentity` or `updateUserGithubProfile`. Add the conflict-throws / same-user-updates / new-inserts cases. | `db/users.ts:254-273` |
| I4 | tests | **Concurrency + recovery invariants untested.** The `inflightRefreshes` mutex (the module's whole "don't double-consume the single-use refresh token" claim) is only exercised sequentially, and the re-read-after-refresh-failure fallback has no test for either branch (re-read valid → return it; still expired → null). Add a `Promise.all` concurrency test and both fallback cases. | `db/user-github-token-store.ts:106-168` |
| I5 | code / types | **Type-safety rule violation: `github_user_id` typed `number` but PG `BIGINT` returns a string** (node-postgres default; no `setTypeParser(20, …)` exists). No runtime break today (only used in a string template + re-passed as a query param, both string-safe), but it is a latent type lie that the refresh path propagates (`saveUserGithubToken({ githubUserId: row.github_user_id })`). Fix: register a PG type parser for OID 20 at pool init, or `z.coerce.number()`. | `schemas/user-github-token-row.ts:18`; `db/user-github-token-store.ts:146` |
| I6 | types / code | **Zod row timestamps `z.date().nullable()` don't match SQLite** (ISO `TEXT`) and the schema is never `.parse()`d, so `UserGithubTokenRow` claims `Date` while SQLite yields `string`. `toEpochMs(Date \| string \| null)` already silently admits this. Make the type honest (`z.union([z.date(), z.string()]).nullable()`) or add a "type-only, not parsed" note. | `schemas/user-github-token-row.ts:22-25` |
| I7 | code | **`connectGithubForUser` missing `_started`/`_failed` logs.** CLAUDE.md requires pairing `_started` with `_completed`/`_failed`; only `_completed` is logged. A throw (network, conflict, DB) logs nothing. | `github-auth/connect-service.ts:38-76` |
| I8 | comments | **Inaccurate comment:** `updateUserGithubProfile`'s doc says "refreshed on each token rotation", but the refresh path (`resolveAccessToken`) only calls `saveUserGithubToken`, never `updateUserGithubProfile`. Profile is updated only at connect time. | `db/users.ts:227-230` |

### Documentation

| ID | Issue | Location |
|----|-------|----------|
| D1 | `CLAUDE.md` schema list says **"11 Tables"**; this adds a 12th (`user_github_tokens`). Bump the count and add the entry. | `CLAUDE.md` "Database Schema" |
| D2 | `migrations/000_combined.sql` header (and the generated `bundled-schema.generated.ts` that inherits it) says **"8 Tables"** with a 7-item list — already stale (missing `users`, `user_identities`, `workflow_node_sessions`) and now further off. Fix the migration header, then `bun run generate:bundled-schema`. | `migrations/000_combined.sql:1-13` |
| D3 (defer ✓) | `.env.example` / `docs-web` for `TOKEN_ENCRYPTION_KEY` + `GITHUB_APP_CLIENT_ID` — correctly deferred to the surfacing increment since the vars are gated and unread today. README needs no change. | — |

### Suggestions (7 found)

| S1 | errors | Floor-guard the `slow_down` backoff: `interval = Math.max(1, data.interval ?? interval + 5)` — a malformed `interval: 0` would otherwise busy-poll GitHub with `sleep(0)`. | `github-auth/device-flow.ts:113` |
| S2 | errors | `postForm` throws only the HTTP status and discards GitHub's `{ error_description }` body, so a misconfigured `GITHUB_APP_CLIENT_ID` reads as a bare `HTTP 401`. Parse and append the body detail. | `github-auth/device-flow.ts:61-63` |
| S3 | errors | `getDecryptedAccessToken` returns `null` for three distinct states (not-connected / expired-no-refresh / refresh-failed). Documented, and dormant today — but when wiring callers, prefer a discriminated result (or at least a `debug` log on the not-connected path) so "broken token" isn't silently indistinguishable from "never connected" (commit attribution would silently vanish). Address in the wiring PR. | `db/user-github-token-store.ts:108-169` |
| S4 | types | `DeviceFlowError.code: string` → narrow to a literal union of the known codes with a `(string & {})` escape hatch for new GitHub codes. | `github-auth/device-flow.ts:45-53` |
| S5 | errors | The re-link path (same user reconnects) updates `platform_display_name` and returns with no log — add an audit line; token rotation is security-relevant. | `db/users.ts:261-265` |
| S6 | tests | Add cheap pure-function tests: `loadDeviceFlowConfig` missing-clientId throw, `isPerUserGitHubEnabled`, `assertEncryptionKeyAtBoot`, `updateUserGithubProfile` COALESCE ordering, and a `connect-service` test asserting `saveUserGithubToken` is never called when `linkGithubIdentity` throws. | `github-auth/config.ts`, `connect-service.ts` |
| S7 | tests | `user-github-token-store.test.ts` asserts on positional query params (`params[3]`, `params[4]`) — implementation-coupled; assert decryptability of the value rather than its column position. | `db/user-github-token-store.test.ts` |

### Rejected findings (checked, not valid)

- **Key the identity on numeric `github_user_id` instead of `login`** — Rejected. The GitHub adapter already stores identities by `login` (`adapters/.../github/adapter.ts:939`). Keying connect on numeric id would create a separate identity row that never matches adapter rows, defeating attribution. `login` is the correct, consistent choice. (The numeric id is rightly used as the *no-reply-email* anchor only — that part is good.)
- **`data.interval ??` in `slow_down` is a dead operand** — Rejected. `interval?: number` is in the `postForm` generic (`device-flow.ts:102`) and GitHub returns a new interval on `slow_down`; the operand is live. The real improvement is the floor-guard in S1.
- **Drop the `fresh &&` guard at `user-github-token-store.ts:165`** — Rejected. TypeScript cannot narrow `fresh` to non-null from `freshExpiry !== null` (no inferable linkage), so removing it breaks type-checking. Keep it.

### Strengths

- AES-256-GCM with auth-tag verification; `iv + authTag + ciphertext` order is consistent between encrypt/decrypt (verified). Tokens never logged — only `userId`/`githubLogin`.
- Solid crypto tests (random IV, wrong-key rejection, tamper detection, unicode) and device-flow tests that swap `globalThis.fetch` rather than mocking the module (refactor-resilient).
- Correct fail-fast in `getEncryptionKey` / `loadDeviceFlowConfig` / `assertEncryptionKeyAtBoot` (boot-time validation, actionable messages).
- Conflict-guard-before-token-persist ordering in `connect-service` is correct and well-commented — a contested account fails without leaving an orphan token row.
- Identity anchored on `login` (matches adapter) for the identity map and on numeric `github_user_id` for the stable no-reply email — both correct.
- Genuinely high-value design comments in `device-flow.ts` (public-client flow, errors-as-HTTP-200). Clean rollback story (single revert; empty, unreferenced table).

### Verdict

**NEEDS FIXES** (no Critical blockers; PR is dormant/additive). Recommended before merge: **I1** (token-loss durability bug), **I2** (race recovery), **I3/I4** (tests for the security + concurrency invariants), **I5/I6** (type-safety rule), **I7/I8/D1/D2** (cheap logging/comment/doc fixes). S1–S7 are nice-to-have; S3 belongs in the wiring PR.

*Multi-agent review (code-reviewer, docs-impact, pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier). Advisory — no files modified.*

Wirasm:
## PR Review Summary — Multi-Agent Review (PR-C: per-user GitHub identity)

7 specialized agents reviewed this PR (code quality, docs, tests, comments, error handling, type design, simplification). The 3 Critical findings below were independently verified against source. Issue IDs use `C/I/S/T` prefixes to avoid GitHub auto-linking.

**Overall: well-architected, security-conscious PR with strong test coverage of the pure core. The Critical findings are all mechanical fixes on the error/security paths.**

---

### 🔴 Critical Issues (3) — fix before merge

| ID | Issue | Location |
|----|-------|----------|
| **C1** | Unguarded `decryptToken` calls break the documented "returns null on irrecoverable failure" contract. `decryptToken` throws on GCM auth-tag mismatch (wrong key after a `TOKEN_ENCRYPTION_KEY` rotation, tampered/bitrotted ciphertext). None of the 3 calls in `resolveAccessToken` are wrapped, and `getDecryptedAccessToken` only does `.finally()`. The `requires:[github]` gate at `orchestrator-agent.ts:331` and the `getUserToken` lambda at `server/index.ts:379` don't catch either → a key rotation makes workflow dispatch crash with an opaque crypto error instead of treating the user as unconnected. **Fix:** wrap the three calls in try/catch → log + return null. | `core/src/db/user-github-token-store.ts:141,153,165` |
| **C2** | `until_bash` loop subprocess bypasses the per-user token scrub. It spreads `...process.env` but **not** `config.envVars` (contrast `executeBashNode` at `:1344`+`:1356`). The scrub overrides — and all managed per-project env vars — never reach `until_bash`. In per-user mode an unconnected user's `until_bash` running `gh`/git inherits whatever GitHub token is in the server's `process.env`, defeating the scrub this PR adds. Also a pre-existing consistency gap (per-project env vars don't reach `until_bash`). **Fix:** spread `...(config.envVars ?? {})` into the env, matching `executeBashNode`. | `workflows/src/dag-executor.ts:2186-2196` |
| **C3** | `.env.example` names the wrong activation variable. It says the feature turns on "When BOTH `GITHUB_APP_CLIENT_ID` and `TOKEN_ENCRYPTION_KEY` are set", but the gate (`isPerUserGitHubEnabled`, `config.ts:22`) checks `GITHUB_APP_ID && TOKEN_ENCRYPTION_KEY`. Both vars are actually required (gate uses `GITHUB_APP_ID`; device flow uses `GITHUB_APP_CLIENT_ID`). Following the comment literally produces a silent lockout: set CLIENT_ID+KEY without APP_ID → feature stays off; set APP_ID+KEY without CLIENT_ID → gate/scrub activate but every connect attempt throws. **Fix:** name all three vars; optionally validate/ warn at boot when enabled without CLIENT_ID. | `.env.example:115` (+ inaccurate module doc at `config.ts:1-9`) |

---

### 🟠 Important Issues (4) — should address before merge

| ID | Issue | Location |
|----|-------|----------|
| **I1** | No startup guard/warning for `ARCHON_WEB_AUTH_HEADER` on a non-loopback bind. `index.ts:785` guards `/internal/*` only in App mode; there's no equivalent for the `X-Archon-User` trust boundary. On a `0.0.0.0` bind with the header set, any client can forge `X-Archon-User` and be attributed as any user. Documented in `.env.example` but not runtime-enforced. **Fix:** warn (or fail-fast) at boot when the header is set and hostname isn't loopback. | `server/src/index.ts:785` vs `api.ts:1037` |
| **I2** | `resolveWebUserId` swallows DB failures and returns `undefined`, so the caller emits a misleading **401 "Web authentication required"** when the real cause is a DB outage. The warn log also omits whether a header was present. **Fix:** distinguish 503 (DB) from 401 (no header); add `headerPresent` to the log. | `server/src/routes/api.ts:1040-1046` |
| **I3** | Connection-status and disconnect routes have no try/catch — a DB failure yields an opaque 500 with no `userId`/operation context (unlike the device-flow routes, which handle errors correctly). **Fix:** wrap both with contextful try/catch. | `server/src/routes/api.ts:1109-1121` |
| **I4** | Comment/behavior contradiction: the inline block comment says "The startup WARN below fires" for the `/internal/*` public-bind guard, but the actual behavior is `fatal()` + `throw` (server refuses to start) — and a more accurate comment ~130 lines later says exactly that. **Fix:** correct the inline comment to "server refuses to start". | `server/src/index.ts:643-644` |

---

### 🟡 Suggestions

**Tests** (rated by criticality — feature is opt-in/gated, so these are follow-ups, not blockers):
- **T1 (8)** `pollDeviceFlowOnce` (non-blocking variant used by the web poll endpoint) is never directly tested — `device-flow.ts:144`.
- **T2 (8)** The `expired_token→expired` / `access_denied→denied` status mapping in the poll handler is untested — this is route-handler *logic*, not just wiring, so the admitted `api.auth.test.ts` skip understates the gap — `api.ts:1085`.
- **T3 (7)** The `requires:[github]` no-op path when `isPerUserGitHubEnabled()===false` (solo-PAT backward-compat guarantee) is untested at the call site — `orchestrator-agent.ts:330`.
- **T4 (6)** Only the absent-header→NULL path is tested for `X-Archon-User`; the present-header→attributed path and the resolve-throw→undefined fallback are not.
- **T5 (5)** Null-expiry (non-expiring token) path in the store, and `getUserOctokit` per-user routing + its 5-min cache, are untested.

**Types** (no blocking issues; `PollOnceResult`, `GitHubAuth`, and the `requires: z.enum(['github'])` are textbook):
- **S1** Add a `key.length === 32` guard to `encryptToken`/`decryptToken` for an actionable error instead of a Node-internal "Invalid key length" — `token-crypto.ts:9,21`.
- **S2** Extract a named `GitIdentity` interface (inline struct duplicated at `isolation/types.ts:54,328`).

**Simplification:**
- **S3** Nested ternary in the poll-status mapping violates the project "no nested ternaries" rule — convert to if/else — `api.ts:1085-1092`.
- **S4** Extract an `expiresInToDate(now, seconds)` helper for the 4 repeated `expires_in ? new Date(now + s*1000) : null` expressions across `connect-service.ts` and `user-github-token-store.ts`.

**Comments:**
- **S5** Stale TODO at `cli/commands/workflow.ts:656` — `archon auth github` has landed in this PR, and the described mechanism (`config.yaml user_id`) was never implemented (actual: `ARCHON_USER_ID`/`$USER`). Rewrite to reference threading `resolveCliUserId()`.

---

### 📄 Documentation Issues

- **Stale:** `README.md:295` says "7 Tables" → now 12. `CLAUDE.md:135` says "@archon/core (7 batches)" → package.json now has ~20.
- **Missing:** `CLAUDE.md` CLI section and `docs-web/.../reference/cli.md` don't mention `archon auth github`. `docs-web/.../reference/configuration.md` is missing the 4 new env vars (present in `.env.example`). `github-app-setup.md` has no "enable per-user identity" section; `authoring-workflows.md` doesn't document `requires: [github]`.
- ✅ `.env.example` is otherwise complete; the `CLAUDE.md` DB-table / API-endpoint / `requires:` additions in this PR are accurate.

---

### ✅ Strengths

- **Crypto:** encrypt/decrypt round-trip, random IV, wrong-key rejection, and bit-flip tamper detection all tested; `getEncryptionKey` boot validation covers absent/short/non-hex.
- **Refresh-token race** (best test in the suite): in-process per-user mutex + re-read-on-failure, with refresh-failure vs persist-failure correctly distinguished so a post-refresh persist failure still returns the valid token.
- **Scrub-by-default is genuinely the default;** `ARCHON_ALLOW_ORG_GITHUB_TOKEN_FALLBACK` is the last gate, not the first — closes the cross-user org-token leak.
- **Identity-conflict-guard-before-persist** ordering prevents orphan token rows.
- **Type design:** strong discriminated unions (`PollOnceResult`, `GitHubAuth`); `requires: z.enum(['github'])` rejects unknown values at load time; schemas follow project conventions; `assertEncryptionKeyAtBoot` is clean fail-fast.

---

### Verdict: **NEEDS FIXES**

C1–C3 are all mechanical and security-relevant; recommend fixing them plus I1–I4 before merge. The test/type/simplification/doc items are good follow-ups (the feature is gated behind opt-in env vars). The architecture, gating discipline, and conservative-fallback posture are solid — the gaps are in error-path robustness and a couple of config/doc mismatches, not in the design.

#### Recommended order
1. C1 (wrap `decryptToken`), C2 (spread `config.envVars` into `until_bash`), C3 (`.env.example` + module doc).
2. I1–I4 (web-auth bind guard, auth-route error handling, the WARN-vs-fatal comment).
3. S1/S3 (key-length guard, drop the nested ternary), then the test follow-ups (T1/T2 first) and docs.

Wirasm:
**Post-merge audit finding (2026-07-13):** the `requires: [github]` workflow gate this PR introduced is enforced only on the orchestrator path (`packages/core/src/orchestrator/orchestrator-agent.ts:579`) — chat/web/adapter invocations. The **CLI-direct run path** (`archon workflow run`) never checks it: on a multi-user install (App mode + `TOKEN_ENCRYPTION_KEY`), a user without a connected GitHub identity can run a `requires: [github]`-gated workflow from the CLI and only fail later at git/gh time, past the worktree/AI cost the gate exists to avoid. No-op for solo PAT installs. Small fix: mirror the `isPerUserGitHubEnabled() && workflow.requires?.length` check in `packages/cli/src/commands/workflow.ts` before isolation/execution.

PR #1824, opened 2026-05-30
fix(web/console): apply BASE_URL prefix to API and favicon paths

miiwins:
## Summary

The `experiments/console/` spike was copied from `packages/web/src/lib/api.ts` but omitted the `API_PREFIX` logic. When deployed behind a reverse-proxy under a subpath (e.g. `/agent-chat`), every `/api/*` request fell through to the host root, hit the wrong upstream, and returned the SPA fallback HTML.

User-visible symptom in a `/agent-chat`-based deployment:

```
API error 404 (/api/codebases): <!DOCTYPE html>...
```

Plus a missing logo in the top-left header (the hand-coded `` in `ConsoleApp.tsx` had the same root cause — the HTML `<link rel="icon">` works because Vite injects the base at build time, but the React img tag bypasses that).

The Old UI was unaffected because `packages/web/src/lib/api.ts` already handles this correctly via `API_PREFIX` (lines 22-30). This PR mirrors that prefix into the console spike.

## Changes

- `packages/web/src/experiments/console/lib/http.ts` — export `API_PREFIX`, prepend it in `requestJson`
- `packages/web/src/experiments/console/skills/runs.ts` — prepend `API_PREFIX` to `fetchArtifact` URL
- `packages/web/src/experiments/console/ConsoleApp.tsx` — build favicon `src` from `import.meta.env.BASE_URL`

## Compatibility

No effect on root-path deployments — `BASE_URL='/'` strips to `''` so the resulting URLs are identical to before.

## Test plan

- [x] Typecheck clean (`bunx tsc --noEmit -p packages/web`)
- [x] Verified locally on a `/agent-chat`-based deployment: console loads, `/api/codebases` returns JSON, logo renders
- [ ] Reviewer to confirm dev (`BASE_URL='/'`) build still works as before

## Bug Fixes
* Fixed console application to work correctly when deployed under a non-root base URL, ensuring proper routing of assets and API requests in reverse proxy setups.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coleam00/Archon/pull/1824?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/web/src/experiments/console/ConsoleApp.tsx
+++ b/packages/web/src/experiments/console/ConsoleApp.tsx
@@ -57,7 +57,7 @@ export function ConsoleApp(): ReactElement {
       <header className="flex h-12 shrink-0 items-center justify-between border-b border-border px-4">
         <div className="flex items-center gap-2.5">
           <img
-            src="/favicon.png"
+            src={`${(import.meta.env.BASE_URL ?? '/').replace(/\/$/, '')}/favicon.png`}
             alt=""
             aria-hidden="true"
             width={22}
--- a/packages/web/src/experiments/console/lib/http.ts
+++ b/packages/web/src/experiments/console/lib/http.ts
@@ -14,6 +14,17 @@ export const SSE_BASE_URL = import.meta.env.DEV
   ? `http://${window.location.hostname}:${API_PORT}`
   : '';
 
+/**
+ * API path prefix for subpath deployments (e.g. /agent-chat).
+ * In dev, Vite proxy handles /api → backend, so no prefix needed.
+ * In production behind a reverse proxy, requests must include the base path
+ * so the auth-gateway can route /agent-chat/api/* → backend /api/*.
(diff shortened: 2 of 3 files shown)


Wirasm:
Thanks @miiwins — closing after a close look, because the fix can't reach its goal from the client alone. In the builds Archon produces, Vite is never invoked with a `--base`, so `import.meta.env.BASE_URL` is always `/` and the rewritten URLs are byte-identical to the current ones. The server side is also root-anchored (static assets and the SPA catch-all have no base-path awareness), and the three SSE EventSource URLs would still bypass the prefix — so a subpath deployment wouldn't work end-to-end even with this merged.

Subdomain deployment (per the reference Caddy config) is the supported shape today. Real subpath support would be a deliberate feature across the build flag, server serving, SSE/upload/artifact URLs, and proxy docs — if that's something you need, an issue making the case for subpath demand is welcome and this PR's inventory of client touch-points would feed straight into it. Appreciate the report.

PR #1840, opened 2026-06-02
fix(providers/codex): capture thread id from thread.started so session resume works

Wirasm:
## Summary

- **Problem:** `CodexProvider` read `thread.id` *before* consuming the SDK event stream and emitted that snapshot as the result chunk's `sessionId`. For a **new** thread the Codex SDK only assigns the id *during* the run — via the `thread.started` event (whose own type comment in `@openai/codex-sdk` reads *"can be used to resume the thread later"*) — so the snapshot was `null` and **Codex returned no `sessionId` at all**.
- **Why it matters:** a missing `sessionId` silently breaks **session resume on Codex** — `persist_session` nodes never get an id to persist, and the planned suspend/resume primitive can't re-thread. It also meant warm-resume "only worked on Claude"; this is what makes it work cross-provider.
- **What changed:** capture `thread_id` from the `thread.started` event inside `streamCodexEvents` and use it for every result emission (`turn.completed`, `turn.failed`, `stream_incomplete`), keeping the snapshot as the fallback for the resume case where the id is already known.
- **What did NOT change (scope boundary):** Codex-provider only. No contract/schema change, no behavior change for resumed threads (snapshot already carries the id), no change to any other provider.

### Before

```
startThread()  →  thread.id = null   ──snapshot──▶  result.sessionId = undefined
   run streams … thread.started{thread_id} (ignored)        ▼
                                                  persist_session / resume: no id  ✗
```

### After

```
startThread()  →  thread.id = null
   run streams … thread.started{thread_id: "th_…"}  ──captured──▶  result.sessionId = "th_…"  ✓
```

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `providers`, `tests`
- Module: `providers:codex`

## Change Metadata

- Change type: `bug`
- Primary scope: `providers`

## Linked Issue

- Related: cross-provider warm session-resume for the workflow pause/resume effort. No single tracking issue.

```bash
bun --filter @archon/providers type-check          # ✅
bun x eslint packages/providers/src/codex/provider.ts   # ✅ clean
bun test packages/providers/src/codex/provider.test.ts  # ✅ 60 pass (incl. new thread.started test)
```

- **Live-verified** (Codex `gpt-5.5`, ChatGPT-account auth) with an in-process two-turn probe on a fixed cwd: turn 1 plants an unguessable token and now returns a `sessionId` (`019e86d8-…`, was `NONE`); turn 2 resumes that session and recalls `ZEPHYR-QUOKKA-5291` — **context survives**. Same probe confirms Pi (`minimax/MiniMax-M2.7`) and Claude (`sonnet`) already survive, so all three in-process providers warm-resume.

## Security Impact (required)

- New permissions/capabilities? No.
- File system access scope changed? No.

## Compatibility / Migration

- Backward compatible? **Yes** — resumed threads already carried the id (snapshot); this only *adds* the id for new threads that previously returned none.
- Config/env changes? No.
- Database migration needed? No.

- Verified: new test (null `thread.id` + `thread.started` event → `result.sessionId === 'evt-thread-id'`); live two-turn resume on Codex `gpt-5.5` recalls the planted token; existing 59 Codex tests still green (snapshot path unchanged).
- Edge cases: resumed thread (snapshot already set, no `thread.started` re-fire needed); `turn.failed` and `stream_incomplete` result chunks now also carry the captured id.
- Not verified: OpenCode (separate provider, not in scope here).

- Affected: `streamCodexEvents` result emissions in `@archon/providers/codex`. Consumed by the orchestrator and dag-executor as `MessageChunk.result.sessionId` (session persistence / `persist_session`).
- Potential unintended effects: a result that previously had `sessionId: undefined` now has a real id — strictly more information; downstream already treats `sessionId` as optional.
- Guardrails: `codex.thread_started` debug log records the captured id.

## Rollback Plan (required)

- Fast rollback: revert the single commit. No persisted state, no migration.
- Observable failure symptoms: Codex result chunks returning `sessionId: undefined` again (resume regressing).

## Risks and Mitigations

- Risk: a stale snapshot id overriding the event id on resume → Mitigation: the snapshot is only the *initial* value; `thread.started` (when present) overwrites it; for resumed threads where no `thread.started` fires, the snapshot is the correct resume id.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure resumed sessions use the actual runtime thread/session ID emitted by the provider, preventing incorrect IDs from snapshots.

* **Tests**
  * Added regression tests covering resumable-session scenarios, verifying correct session/thread ID resolution for completed, failed, and incomplete streams.

* **Documentation**
  * Updated streaming example to demonstrate capturing and using the resolved thread/session ID.

### DIFF
--- a/packages/docs-web/src/content/docs/reference/architecture.md
+++ b/packages/docs-web/src/content/docs/reference/architecture.md
@@ -500,7 +500,14 @@ for await (const msg of query({ prompt, options })) {
 **Codex SDK** (`packages/providers/src/codex/provider.ts`):
 
 ```typescript
+// A new thread's id is assigned during the run via the thread.started event,
+// not synchronously on startThread() — capture it for a resumable sessionId.
+let resolvedThreadId = thread.id;
 for await (const event of result.events) {
+  if (event.type === 'thread.started') {
+    resolvedThreadId = event.thread_id; // resumable id; persist_session depends on it
+    continue;
+  }
   if (event.type === 'item.completed') {
     switch (event.item.type) {
       case 'agent_message':
@@ -514,7 +521,7 @@ for await (const event of result.events) {
         break;
     }
   } else if (event.type === 'turn.completed') {
-    yield { type: 'result', sessionId: thread.id };
+    yield { type: 'result', sessionId: resolvedThreadId };
     break; // CRITICAL: Exit loop on turn completion
   }
 }
--- a/packages/providers/src/codex/provider.test.ts
+++ b/packages/providers/src/codex/provider.test.ts
@@ -146,6 +146,71 @@ describe('CodexProvider', () => {
       });
     });
 
+    test('captured thread id flows through the turn.failed result', async () => {
+      mockStartThread.mockReturnValue({ id: null, runStreamed: mockRunStreamed });
+      mockRunStreamed.mockResolvedValue({
+        events: (async function* () {
+          yield { type: 'thread.started', thread_id: 'evt-thread-id' };
+          yield { type: 'turn.failed', error: { message: 'boom' } };
+        })(),
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('x', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks.find(c => c.type === 'result')).toMatchObject({
+        type: 'result',
+        sessionId: 'evt-thread-id',
+        isError: true,
+      });
+    });
+
+    test('captured thread id flows through the stream_incomplete result', async () => {
+      mockStartThread.mockReturnValue({ id: null, runStreamed: mockRunStreamed });
+      mockRunStreamed.mockResolvedValue({
+        // Stream closes without turn.completed/turn.failed → fail-stop result.
+        events: (async function* () {
+          yield { type: 'thread.started', thread_id: 'evt-thread-id' };
+        })(),
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('x', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks.find(c => c.type === 'result')).toMatchObject({
+        type: 'result',
+        sessionId: 'evt-thread-id',
+        isError: true,
+        errorSubtype: 'codex_stream_incomplete',
+      });
+    });
+
+    test('an empty thread.started thread_id keeps the snapshot id (guard)', async () => {
+      // Default startThread snapshot id is 'new-thread-id'; an empty event id
+      // must not overwrite it (and would otherwise warn, not emit sessionId: '').
+      mockRunStreamed.mockResolvedValue({
+        events: (async function* () {
+          yield { type: 'thread.started', thread_id: '' };
+          yield { type: 'item.completed', item: { type: 'agent_message', text: 'ok' } };
+          yield { type: 'turn.completed', usage: defaultUsage };
+        })(),
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('x', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks.find(c => c.type === 'result')).toMatchObject({
+        type: 'result',
+        sessionId: 'new-thread-id',
+      });
+    });
+
     test('yields tool events from command_execution items', async () => {
       mockRunStreamed.mockResolvedValue({
         events: (async function* () {
@@ -600,8 +665,9 @@ describe('CodexProvider', () => {
         })(),
       });
 
-      for await (const _ of client.sendQuery('test prompt', '/workspace', 'existing-thread')) {
-        // consume
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test prompt', '/workspace', 'existing-thread')) {
+        chunks.push(chunk);
       }
 
       expect(mockResumeThread).toHaveBeenCalledWith(
@@ -615,6 +681,10 @@ describe('CodexProvider', () => {
         })
       );
       expect(mockStartThread).not.toHaveBeenCalled();
+      // No thread.started re-fires on resume → the snapshot (resumeThread's id) survives.
+      expect(chunks.find(c => c.type === 'result')).toMatchObject({
+        sessionId: 'resumed-thread-id',
+      });
     });
 
     test('falls back to new thread when resume fails and notifies user', async () => {
--- a/packages/providers/src/codex/provider.test.ts
+++ b/packages/providers/src/codex/provider.test.ts
@@ -117,6 +117,35 @@ describe('CodexProvider', () => {
       });
     });
 
+    test('captures the new-thread id from the thread.started event (resumable sessionId)', async () => {
+      // The real Codex SDK assigns a NEW thread's id during the run, via the
+      // thread.started event — not synchronously on startThread(). Simulate a
+      // thread whose .id is still null and assert the result carries the id from
+      // the event, so persist_session / suspend-resume have a resumable id.
+      mockStartThread.mockReturnValue({ id: null, runStreamed: mockRunStreamed });
+      mockRunStreamed.mockResolvedValue({
+        events: (async function* () {
+          yield { type: 'thread.started', thread_id: 'evt-thread-id' };
+          yield {
+            type: 'item.completed',
+            item: { type: 'agent_message', text: 'stored' },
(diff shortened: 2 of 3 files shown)


Wirasm:

**Scope:** `fix(providers/codex): capture thread id from thread.started so session resume works` — 2 files, +52/-3. Reviewed by code-reviewer, docs-impact, pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier.

**Bottom line:** the core fix is correct — capturing `thread_id` from `thread.started`, seeding from the snapshot for resumes, and `continue`-ing past the event is sound for both the new-thread and resume cases. The findings below are quality improvements; none break the happy path, but a few have multi-agent consensus and are worth addressing before merge.

> Verified against `@openai/codex-sdk@0.125.0` `dist/index.d.ts`: `ThreadStartedEvent` **is exported** with `thread_id: string` (non-optional, line 109); doc comment at line 108 reads *"The identifier of the new thread. Can be used to resume the thread later."*; the type is *"Emitted when a new thread is started as the first event."* (line 105).

### Important Issues (5)

| # | Agent(s) | Issue | Location |
|---|----------|-------|----------|
| I1 | code-reviewer (90), comment-analyzer, code-simplifier | Inline assertion `(event as { thread_id?: string })` re-declares a type the SDK already exports. `ThreadStartedEvent` is in the same import block as `TurnCompletedEvent`, and its `thread_id` is `string` (non-optional). This is the exact anti-pattern CLAUDE.md *SDK Type Patterns* warns against. Fix: import `ThreadStartedEvent`, cast to it, and the `typeof … && length > 0` guard collapses to a truthiness check. | `provider.ts` `thread.started` handler (~L341) |
| I2 | silent-failure-hunter (HIGH), pr-test-analyzer (5) | Silent degradation: if `thread.started` fires on a **new** thread with a missing/empty `thread_id`, the inner `if` is skipped, no log fires, and the result emits `sessionId: undefined`. Downstream (`dag-executor.ts:3164`) treats that as "no session" and **deletes the persisted `persist_session` row** — silent loss of continuity. Per the SDK contract `thread_id` is always a non-empty string, so this is defensive-only, but CLAUDE.md *Fail Fast* says document the safe fallback or `warn`. Add a `warn` on the absent-id-on-new-thread branch. | `provider.ts` `thread.started` handler (~L341) |
| I3 | pr-test-analyzer (8, 7) | `resolvedThreadId` feeds **three** terminal emissions but only `turn.completed` (success) is tested with the captured id. No test asserts the captured id flows through `turn.failed` or `stream_incomplete`. The existing `turn.failed` tests use a non-null snapshot, so they pass under both old and new code and don't exercise the fix. Add one test each for the failure paths with a null snapshot + `thread.started`. | `provider.test.ts` |
| I4 | docs-impact | Stale doc: the Codex event-loop snippet teaches the pre-fix `sessionId: thread.id` snapshot pattern that returned `null` for new threads — the exact bug this PR fixes. Update to show `resolvedThreadId` captured from `thread.started`. | `packages/docs-web/src/content/docs/reference/architecture.md:517` |
| I5 | comment-analyzer | The comment presents `"can be used to resume the thread later"` as a verbatim SDK quote, but it's a lowercase-truncated fragment — the real doc comment is *"The identifier of the new thread. Can be used to resume the thread later."* Quote it in full or drop the quotation marks. | `provider.ts` `thread.started` handler |

### Suggestions (6)

| # | Agent(s) | Suggestion | Location |
|---|----------|-----------|----------|
| S1 | silent-failure-hunter (MED) | `codex.thread_started` logs at `debug` — invisible at default `info`. This is the only place a new thread's resumable id is established; consider `info` so "why does Codex start fresh despite `persist_session`?" has a breadcrumb. | `provider.ts` `thread.started` handler |
| S2 | comment-analyzer | "thread.started may or may not re-fire" on resumed threads contradicts the SDK (it's the *first* event for a *new* thread). Tighten to "the SDK emits thread.started only for new threads" or cite an observed exception. Logic stays correct either way (snapshot seeds `resolvedThreadId`). | `provider.ts` comment above `resolvedThreadId` |
| S3 | pr-test-analyzer (6) | Add a `sessionId` assertion to the existing resume test to lock in "snapshot id survives when `thread.started` doesn't re-fire" (currently only `resumeThread` args are asserted). | `provider.test.ts` (~L567) |
| S4 | pr-test-analyzer (5) | Add a regression test for the `length > 0` guard: `thread.started` with `thread_id: ''` should keep the snapshot id, not emit `sessionId: ''`. | `provider.test.ts` |
| S5 | code-reviewer (82) | `let resolvedThreadId: string \| null \| undefined` — `null` is collapsed by `?? undefined` at all three sites. Could init as `threadId ?? undefined` and type `string \| undefined`. (code-simplifier judged this not worth the churn since it mirrors the param type — minor, take it or leave it.) | `provider.ts` (~L322) |
| S6 | silent-failure-hunter (MED), code-simplifier | The `continue` after `thread.started` is correct (metadata, no content) but undocumented; a one-line note prevents it reading as accidental suppression. code-simplifier also notes the comment block could be trimmed ~half without losing meaning. | `provider.ts` `thread.started` handler |

### Strengths
- **Core fix is correct** — snapshot-vs-event priority is sound for both new and resumed threads; `continue` is safe since `thread.started` precedes all `item.*`/`turn.*` events.
- **New test is behavioral** — black-box via `sendQuery`, asserts on output chunk shape, matches existing test style; the test's own comment explaining the simulated SDK timing is exemplary.
- **The "why" comment block** on `resolvedThreadId` documents both code paths and the downstream dependents — the kind of comment that prevents regressions.
- **Surrounding error handling is solid** — resume-failure fallback surfaces a user-facing system message, retry exhaustion always throws (no silent `undefined`).
- **Closes a declared-vs-runtime gap** — `capabilities.ts` already declared `sessionResume: true`; this makes the runtime actually deliver a resumable id. Small, reversible, single-commit, well-described PR.

### Documentation
- `architecture.md:517` — stale Codex snippet teaches the broken pre-fix pattern (see I4). All other docs (`CLAUDE.md`, `authoring-workflows.md`, `ai-assistants.md`, `README.md`, `capabilities.ts`) are accurate and provider-agnostic — no other doc claimed resume was "Claude only".

### Verdict: **NEEDS FIXES (minor)**

The fix is correct and the PR is well-scoped. Recommended before merge: **I1** (SDK type — 3-agent consensus + guideline violation), **I3** (error-path test coverage), **I4** (stale doc), **I5** (quote accuracy). **I2** is defensive-only given the SDK contract but a one-line `warn` satisfies *Fail Fast* cheaply. Suggestions are optional polish.

*Advisory review — no files were modified.*

Wirasm:
Thanks for the thorough multi-agent pass. Addressed in `d01efec1`:

**Important**
- **I1** — cast `thread.started` to the SDK's exported `ThreadStartedEvent`; the guard collapses to a truthiness check. (CLAUDE.md SDK Type Patterns)
- **I2** — on a (contract-impossible) empty/missing `thread_id`, `warn` instead of silently emitting `sessionId: undefined` — which the dag-executor reads as session-less and uses to drop `persist_session` continuity. No silent degradation.
- **I3** — added tests asserting the captured id flows through `turn.failed` and `stream_incomplete` (null snapshot + `thread.started`), not just the success path.
- **I4** — updated the stale Codex snippet in `architecture.md` (it taught the pre-fix `sessionId: thread.id` pattern this PR fixes).
- **I5** — quoted the SDK doc comment in full.

**Suggestions**
- **S1** — `codex.thread_started` now logs at `info` (the breadcrumb for "why fresh despite persist_session?").
- **S2** — corrected "may or may not re-fire" → "only for new threads".
- **S3** — the resume test now asserts the snapshot id survives when `thread.started` doesn't re-fire.
- **S4** — added a guard regression test (`thread_id: ''` keeps the snapshot, doesn't emit `sessionId: ''`).
- **S6** — documented the `continue`.

**Deferred**
- **S5** (narrow `resolvedThreadId` to `string | undefined`) — kept as-is per the code-simplifier's own call that it's churn mirroring the param type; the `?? undefined` collapse at the three sites is harmless.

`type-check` + `lint` clean; **63 codex tests pass**. The `docker-build` CI failure was a transient Docker Hub registry timeout while booting buildkit (`Get "https://registry-1.docker.io/v2/": context deadline exceeded`) during *Set up Docker Buildx* — unrelated to this diff; `test` passed on Ubuntu + Windows. The new push re-triggers it.

PR #1841, opened 2026-06-02
feat(auth): opt-in web login via Better Auth + user-identity seam

Wirasm:
## Summary

- **Problem:** PR-C's interim web auth is a trusted reverse-proxy header (`X-Archon-User`) backed by a single-user sidecar — no real signup/login, no per-user web sessions, no role, no identity seam rich enough to user-scope resources later.
- **Why it matters:** A shared team VPS needs each teammate to log in as themselves behind an invite gate, with a per-request `{ userId, role }` identity threaded through the backend that future scoping (per-user AI config, "my runs") can hook into — without committing to an ACL policy still being discovered through dogfooding.
- **What changed:** Opt-in email/password web login via **Better Auth** (`~1.3`, allowlist-gated), mounted at `/api/auth/*`. A Better Auth session maps to the canonical `remote_agent_users` row via `user_identities('web', …)`. Generalized the seam to `resolveAuthContext(c): { userId, role }` (session-first → header → undefined). Added a `role` column (default `admin`) and a **non-enforcing `?mine`** filter on the runs + conversations lists. Login page + session gate in the web UI, shown only when enabled.
- **What did NOT change (scope boundary):** No ACL/hard isolation (visibility stays open; `?mine` is a filter, not a boundary). No per-user AI config, no project ownership, no cross-surface identity merge, no `archon_*` table rename, **no SQLite web auth** (Postgres-only), no removal of the `X-Archon-User` header (kept as a fallback). Solo/SQLite installs are byte-for-byte unchanged.

### Before

```
AUTH (no real login) — single-user sidecar + Caddy, or open:
  Browser ─► Archon            (identity only as a trusted X-Archon-User header)
            resolveWebUserId(c): header → findOrCreateUser('web', hdr) | undefined
  Every list endpoint returns everything; no per-user filter anywhere.
```

### After

```
AUTH DISABLED (no BETTER_AUTH_SECRET) — local/solo:
  identical to today (header seam still works; no login UI).

AUTH ENABLED (Postgres + BETTER_AUTH_SECRET):
  Browser ─► /login (email+pw)  ─signup gated by allowlist─► *session cookie set*
  Every /api/* request:
    *resolveAuthContext(c):*
      1. Better Auth session → findOrCreateUser('web', baUserId)
         → { userId, role }   ◄── canonical remote_agent_users row
      2. else X-Archon-User header (proxy deploys)
      3. else undefined
  Lists default to ALL (open). *?mine=true → filter to ctx.userId* (non-enforcing).
  Web UI: *login gate + "signed in as" + sign out*, shown only when
     GET /api/auth/status → { enabled: true }.
```

### After

```
 [+] @archon/server/auth/  ===  betterAuth (pg Pool, remote_agent_auth_* tables)
        config.ts (gate)   ===  isWebAuthEnabled / assertWebAuthAtBoot / allowlist
        instance.ts        ===  getAuth() lazy singleton
            │
            ▼ getAuth()
 [~] server/index.ts ── mounts /api/auth/* (skips Archon-owned paths) ; assertWebAuthAtBoot
 [~] server/routes/api.ts ── resolveAuthContext {userId,role} ; GET /api/auth/status ; ?mine
            │
            ▼ findOrCreateUserByPlatformIdentity('web', baUserId)
 [~] core/db/users.ts ──► remote_agent_users (+ [+] role) ◄── user_identities('web')
 [~] core/db/{workflows,conversations}.ts ── optional userId filter
 [+] web: LoginPage / SessionGate / auth-client ── gated by /api/auth/status
```

**Connection inventory:**

| server/index.ts | server/auth (getAuth) | **new** | mount /api/auth/* + boot assert |
| server/routes/api.ts | server/auth (getAuth, isWebAuthEnabled, getSignupMode) | **new** | seam + status + ?mine |
| server/auth/instance.ts | pg.Pool (dedicated) | **new** | Better Auth DB |
| routes/api.ts | core/db/users | **modified** | resolveAuthContext returns {userId, role} |
| core/db/{workflows,conversations} | — | **modified** | optional userId filter arg (appended) |
| web App.tsx | web/components/auth/SessionGate | **new** | wraps Layout |
| web TopNav / LoginPage | web/lib/auth-client | **new** | useSession / signIn / signUp / signOut |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: L`
- Scope: `server`, `core`, `web`, `config`, `docs`, `dependencies`
- Module: `server:auth`, `server:routes`, `core:db`, `web:auth`

## Change Metadata

- Change type: `feature` (web auth) + `refactor` (identity seam)
- Primary scope: `multi` (server + core + web)

## Linked Issue

- Related #1823 (PR-C: per-user GitHub identity + interim web auth — this supersedes the single-user `auth-service` sidecar; the `X-Archon-User` header remains a fallback)

```bash
                   # type-check (10 pkgs), lint --max-warnings 0, format:check, test
```

- Evidence provided: `bun run validate` exits 0. Full test suite = 95 batches, `0 fail`. New tests: `auth/config.test.ts` (14), `api.auth.test.ts` (7), plus `role` cases in `users.test.ts` / `schemas/index.test.ts`.
- Migration idempotency: `psql < migrations/000_combined.sql` applied **twice** cleanly on a real PostgreSQL 18.
- If any command skipped: `api.generated.d.ts` not regenerated (web uses a hand-typed `AuthStatus`; not part of `validate`).

## Security Impact (required)

- New permissions/capabilities? **Yes** — opt-in web login. Disabled by default; enabled only with `DATABASE_URL` + `BETTER_AUTH_SECRET`. Boot fails fast if the secret is < 32 chars.
- New external network calls? **No** (Better Auth is same-origin; no email sender, verification off).
- Secrets/tokens handling changed? **Yes** — `BETTER_AUTH_SECRET` signs sessions; Better Auth stores password hashes in `remote_agent_auth_account`. Session cookies are httpOnly/SameSite (Better Auth defaults).
- File system access scope changed? **No**.
- Mitigation: allowlist-gated signup by default; the existing non-loopback bind advisory now also fires when web auth is enabled; `?mine` is explicitly **not** a security boundary (documented).

## Compatibility / Migration

- Backward compatible? **Yes** — fully opt-in; disabled installs are unchanged; existing seam call sites and the GitHub device flow keep working (verified).
- Config/env changes? **Yes** (all optional): `BETTER_AUTH_SECRET`, `BETTER_AUTH_URL`, `BETTER_AUTH_TRUSTED_ORIGINS`, `ARCHON_AUTH_ALLOWED_EMAILS`.
- Database migration needed? **Yes, automatic.** `remote_agent_users.role` (default `admin`) + four `remote_agent_auth_*` tables (Postgres) converge on startup via the existing advisory-locked bundled-schema apply; SQLite gets only `role`. No manual step.

Live end-to-end smoke test against a real PostgreSQL 18 (server booted with web auth enabled), beyond CI:

- Verified scenarios: `/api/auth/status` → `{enabled:true,signup:allowlist}`; non-allowlisted signup → **403**; allowlisted signup → **200 + session cookie**; sign-in → **200**, wrong password → **401**; `get-session` → user; session → canonical `remote_agent_users` (role=admin) via `user_identities('web', <betterAuthId>)`; new conversation `user_id` attributed to that canonical user; `?mine=true` returns **only the caller's** conversations (Alice vs Bob), default returns all.
- Edge cases checked: `?mine=true` with no identity → returns all (non-enforcing); GitHub device route under `/api/auth/*` returns **401 not 404** (Better Auth doesn't shadow it); fresh-DB schema auto-apply creates all 4 auth tables + `role`; migration idempotent twice.
- What was not verified: live multi-browser cookie/CSRF behind a real reverse proxy; email verification flows (off by design).

- Affected subsystems: server bootstrap (`index.ts`), the web-identity seam in `routes/api.ts` (shared by conversation/message/run-create attribution + GitHub device flow), `core/db/{users,workflows,conversations}`, the combined migration + bundled schema, web routing.
- Potential unintended effects: the generalized `resolveWebUserId`/`requireWebUser` now also consult a Better Auth session — but `getAuth()` returns null when disabled, so behavior is identical for non-enabled installs (covered by the existing route tests, all green).
- Guardrails: opt-in gate; existing route tests + new `api.auth.test.ts` assert session-first vs header precedence and the no-filter default.

## Rollback Plan (required)

- Fast rollback: revert this PR (single commit), or operationally **unset `BETTER_AUTH_SECRET`** → `getAuth()` returns null, `/api/auth/*` unmounts, status reports `enabled:false`, UI renders as today. The `role` column + auth tables are additive and harmless if left.
- Feature toggles: `BETTER_AUTH_SECRET` (master switch) + `ARCHON_AUTH_ALLOWED_EMAILS`.
- Observable failure symptoms: boot throws on a too-short secret; `web_auth.session_resolve_failed` / `web.user_resolve_failed` warns; `/api/auth/status` 5xx.

## Risks and Mitigations

- Risk: Better Auth `1.4.x` cross-origin regression (#7657).
  - Mitigation: pinned `~1.3` (installed 1.3.34, stays on the 1.3 line); same-origin posture (omit `baseURL`) sidesteps cross-origin entirely.
- Risk: Better Auth `/api/auth/*` catch-all shadows Archon's own `/api/auth/status` + `/api/auth/github*`.
  - Mitigation: the mount explicitly `next()`s Archon-owned paths (deterministic; verified the GitHub route returns 401 not 404).
- Risk: public URL + open signup.
  - Mitigation: allowlist-gated by default; non-loopback bind advisory extended to web auth; `?mine` is not a boundary (documented).

## Summary by CodeRabbit

* **New Features**
  * Opt-in Web UI email/password login (Postgres-only): login/signup UI, session-aware gating, sign-out, and auth-aware top navigation.
  * Server endpoint reporting auth status and a server-side API gate (most /api/* return 401 when enabled; auth and health endpoints exempt).
  * Non-enforcing ?mine filters for conversations and workflow runs to show caller-owned items when identity is resolved.
  * Added persistent user role support (admin/member).

* **Documentation**
  * Expanded deployment, DB schema, signup/allowlist, and security guidance for Web Auth.

* **Tests**
  * New coverage for auth config, routes, and identity behaviors.


Wirasm:
## PR Review Summary — multi-agent review

Reviewed by 7 specialized agents (code, docs, tests, comments, error-handling, types, simplification) against the diff (`feat/web-auth-better-auth` → `dev`, +1195/-31, 30 files). Findings cross-checked against the actual code before posting. No bare-`#` issue numbers below — my own findings use `C/I/S` prefixes.

**Verdict: NEEDS MINOR FIXES** — no architectural blockers. One confirmed React render bug and a handful of factual doc/comment errors are worth fixing before merge; everything else is a follow-up. The auth design itself (boot fail-fast, 401-vs-503 discipline, parameterized queries, opt-in gate, byte-for-byte-unchanged disabled installs) is solid.

### Reviewer adjudication (a note on severity)

The `silent-failure-hunter` rated two items **CRITICAL / do-not-merge**. I read the code and am **downgrading both**, transparently:

- **Session→header→undefined fallthrough in `resolveAuthContext`** (`api.ts:1064-1094`) is the *soft* attribution path — it feeds `user_id` stamping and the **non-enforcing** `?mine` filter, never an access gate. Falling to `undefined` produces NULL attribution + open lists, which *is* the documented default. No privilege escalation occurs. The *strict* path (`requireWebUser`) already returns 503 on session failure and does not fall through. → reclassified as an **observability** suggestion (S5).
- **`buildAuth` pool construction unguarded** (`instance.ts:33-50`) is a real robustness gap, but the same `DATABASE_URL` already powers the core adapter that connects at boot, so an independently-malformed auth pool is unlikely, and `new Pool()` rarely throws synchronously. → reclassified **Important** (I5), not blocking.

### Important Issues (5)

| I1 | code-reviewer, code-simplifier | `navigate()` is called in the **render body** with no `return`, so the full login form renders before redirecting and React Strict Mode fires it twice. Fix: `return <Navigate to="/" replace />` (note: `Navigate` isn't imported yet — `SessionGate` already uses this idiom). | `packages/web/src/routes/LoginPage.tsx:29-31` |
| I2 | comment-analyzer | "created only when web auth is enabled" is **factually wrong** — the four `remote_agent_auth_*` tables are created on *every* Postgres install (idempotent `IF NOT EXISTS` in the bundled schema applied on every boot); they're only *populated* when enabled. Same wrong claim in CLAUDE.md (see I3). | `migrations/000_combined.sql:431-432` |
| I3 | docs-impact, comment-analyzer | Same "created only when enabled" error **and** the table names drop the `remote_agent_` prefix (actual: `remote_agent_auth_user/_session/_account/_verification`). | `CLAUDE.md:425` |
| I4 | comment-analyzer | Stale comment: "the mounted handler 404s and falls through to here." The real mechanism is an explicit `next()` in `index.ts` (`isArchonOwnedAuthPath`) that bypasses Better Auth before it ever runs — it never 404s. | `packages/server/src/routes/api.ts:1153-1155` |
| I5 | silent-failure-hunter (downgraded) | `buildAuth()` isn't wrapped in `getAuth()` and the dedicated `pg.Pool` is never `end()`ed on shutdown. A construction throw propagates into the request path; a bad `DATABASE_URL` surfaces as a first-request 500 rather than at boot. Wrap construction → log error → cache `null`; close the pool on shutdown. | `packages/server/src/auth/instance.ts:33-50` |

### Test Gaps (should add before merge)

| ID | Criticality | Gap | Location |
|----|-------------|-----|----------|
| T1 | 8/10 | `?mine=true` with **no resolvable identity** → must still return all (the PR's central non-enforcement guarantee) — untested for both runs and conversations. A regression would silently turn the filter into a zero-result gate. | `api.auth.test.ts` |
| T2 | 7/10 | `resolveAuthContext`: `getSession()` throws → falls through to header (the documented resilience path) — untested. | `api.auth.test.ts` |
| T3 | 5/10 | The `isArchonOwnedAuthPath` shadowing guard lives in `startServer`, so `registerApiRoutes`-level tests never exercise it. Acceptable to defer (needs a `startServer` integration test). | `index.ts:630-636` |

### Suggestions (follow-ups, non-blocking)

- **S1** (type-design): `resolveAuthContext`/`requireWebUser` return `role: string`, widening the already-`UserRole` value. One-liner: `import type { UserRole }` and annotate `role: UserRole`. `api.ts:1060,1112`
- **S2** (type-design): `api.generated.d.ts` predates the auth routes, so `AuthStatus`/`GithubDevice*` are hand-written in `web/src/lib/api.ts:74-101` and can drift from the server Zod enums. Regenerate types post-merge and replace with `components['schemas'][...]`.
- **S3** (code-simplifier, comment-analyzer): `requireWebUser` duplicates the full session-resolution path from `resolveAuthContext`. **Do not naively dedupe** — the strict path intentionally returns 503 on session failure instead of falling through. Either keep separate **with an explaining comment**, or extract a shared resolver that takes an error-posture callback.
- **S4** (code-simplifier): `resolveWebUserId` is a one-line wrapper with a misleading "Back-compat" label (it's new code); inline `(await resolveAuthContext(c))?.userId` at its 3 call sites, or drop the label.
- **S5** (silent-failure-hunter, downgraded): observability on the soft path — soft uses `warn`, strict uses `error` for the identical DB failure; add `path` to the log context; optionally log when `?mine` is skipped due to no identity.
- **S6** (type-design): `mine: z.string().optional()` accepts `mine=1`/`mine=yes` (silently falsy). `z.enum(['true','false']).optional()` makes the contract explicit.
- **S7** (comment-analyzer): `auth-client.ts:8` "these calls are unused" — `useSession` is always invoked by `SessionGate`; the *result* is ignored when disabled, but the network call still fires. Same nuance in `SessionGate` "pure passthrough" JSDoc (a full-screen spinner renders until `/api/auth/status` resolves).
- **S8** (code-simplifier): minor consistency — `LoginPage` `auth-status` query lacks the `staleTime: 5min` its two siblings use (shared cache key makes it harmless); parse the allowlist once at `buildAuth` construction rather than per-signup.
- **S9** (type-design): export `UserRole` from `packages/core/src/schemas/index.ts`. (pr-test-analyzer) add a `role: 'member'` round-trip case in `users.test.ts`.

### Documentation Issues (existing docs not touched by the PR)

- `packages/docs-web/.../security.md:98` — "The Web UI has no built-in user authentication" is now false for Postgres + `BETTER_AUTH_SECRET` installs.
- `packages/docs-web/.../database.md:69` — "10 tables" is stale (now 16); missing `workflow_node_sessions`, `user_github_tokens`, the four auth tables, the new `users.role` column, the `web` `user_identities` platform, and migrations `021`/`022`.
- `packages/docs-web/.../deployment/docker.md:321-375` — the form-auth sidecar section doesn't note Better Auth supersedes it for Postgres deployments.
- `packages/docs-web/.../configuration.md:312` — `ARCHON_WEB_AUTH_HEADER` still labelled "interim web auth"; reads as if being removed, but it's now the documented fallback.

### Strengths

- **Security fundamentals solid**: new `userId` filters use positional `$N` params (no injection); Better Auth catch-all correctly `next()`s Archon-owned `/api/auth/status` + `/api/auth/github*` (and `DELETE` is never intercepted); allowlist enforced in a `databaseHooks.user.create.before` hook returning a clean `APIError` 403 with case-folded comparison; `assertWebAuthAtBoot` fails fast on a short secret **before** binding the socket.
- **Correct 401-vs-503 discipline** in `requireWebUser` — a backend outage doesn't masquerade as "auth required". This is the discipline the soft path should reference (S5).
- **Backward compatibility verified**: header fallback preserved for proxy deploys; SQLite/disabled installs unchanged; `role` backfilled via PRAGMA check; `check:bundled-schema` confirms the embedded schema is in sync.
- **Zod/type hygiene**: `userRoleSchema` is a strict `z.enum`, types via `z.infer`, `z` from `@hono/zod-openapi`, `AuthInstance` inferred from the SDK, no `any` in new code.
- **`config.ts` is pure and thoroughly unit-tested** (14 cases); strong documented manual E2E (allowlist 403, session→canonical-user mapping, `?mine` Alice-vs-Bob, idempotent migration twice on PG 18).

### Recommended actions before merge

1. Fix **I1** (real render bug) and the doc/comment factual errors **I2–I4** (cheap, prevent future confusion).
2. Add tests **T1**/**T2** — they pin the PR's headline guarantees.
3. Consider **I5** (robustness) and **S1**/**S3** (small, low-risk); defer the rest.

Wirasm:
## Review findings addressed — pushed `ef1bd125`

Worked through the multi-agent review above. `bun run validate` exits 0 (bundled/schema/skill, type-check ×10, lint `--max-warnings 0`, format, full test suite).

**Important**
- **I1** — `LoginPage` uses declarative `<Navigate to="/" replace />` instead of `navigate()` in the render body (fixes the form-flash + Strict-Mode double-fire).
- **I2/I3/I4** — corrected the "created only when web auth is enabled" claim in the migration + CLAUDE.md (the `remote_agent_auth_*` tables are created on *every* Postgres install), added the `remote_agent_` prefix to the CLAUDE.md table names, and fixed the `/api/auth/status` "404s and falls through" comment to describe the actual `next()` mechanism. Bundled schema regenerated.
- **I5** — `getAuth()` now guards `buildAuth()` construction (logs `web_auth.instance_build_failed`, caches `null`) so a bad `DATABASE_URL` can't propagate into the request path as a swallowed session failure; added `closeAuth()` to release the dedicated pool on shutdown.

**Tests**
- **T1** — `?mine=true` with no identity returns all (runs + conversations), asserting `findOrCreateUser` is never called.
- **T2** — session lookup throws → falls through to the `X-Archon-User` header.

**Suggestions**
- **S1/S9** — `role: UserRole` on both resolver return types; `UserRole` exported through `@archon/core`; added a `role: 'member'` round-trip test.
- **S4** — fixed the misleading "Back-compat" label on `resolveWebUserId`.
- **S5** — added `path` context to the soft-path catch logs + a note on why `warn` (not `error`) is correct there.
- **S6** — `mine: z.enum(['true','false']).optional()` (no web caller yet, so safe).
- **S7/S8** — corrected the `auth-client.ts` / `SessionGate` comments; consistent `staleTime`; allowlist parsed once at construction.

**Docs** — `security.md` ("no built-in user authentication" → conditional on Postgres + Better Auth), `database.md` (10→16 tables, entries 11–16, `role` column, migrations 021/022), `docker.md` (Better Auth pointer on the sidecar section), `configuration.md` ("interim" → "reverse-proxy fallback").

**Deliberately deferred (with rationale)**
- **S2 (regenerate `api.generated.d.ts`)** — needs a running dev server and would pull unrelated drift accumulated since May into this focused PR (scope creep, hurts rollback). The hand-written web types are correct; belongs in a dedicated types-regen change.
- **S3 (extract a shared resolver from `requireWebUser`/`resolveAuthContext`)** — skipped as over-engineering: they intentionally differ (soft `warn`+fallthrough vs strict 503), already documented, and a posture-callback abstraction for two callers fails the Rule of Three.
- **T3 (startServer integration test for the route-shadowing guard)** — deferred; needs a full server harness for marginal value.

Note on the earlier "CRITICAL" flags: the session→header→undefined fallthrough is the soft attribution path (non-enforcing `?mine` + `user_id` stamping), not an access gate — no privilege escalation, so it was treated as an observability nit (S5), not a blocker.

Wirasm:
## Re-review — new commit `2b5c9b65` (server-side `/api/*` gate + safe-by-default signup)

Re-ran the multi-agent review scoped to the new commit (the rest of the PR was reviewed + fixed in `ef1bd125`). Findings cross-checked against the actual code. My own findings use `C/I/S` prefixes.

**Verdict: NEEDS FIXES — no merge-blockers.** The gate is fundamentally sound: it **fails closed**, safe defaults are correct, and the opt-outs parse fail-safe. The issues are a web UX bug, stale comments, doc inaccuracies, and test/defense-in-depth gaps.

### Reviewer adjudication (two agent "CRITICAL" calls corrected)

- **"Gate fails OPEN on resolution error"** — incorrect. `resolveAuthContext` (`api.ts:1078-1119`) wraps both the session and header paths in try/catch, and `getAuth()` no longer throws (caches `null`). So the gate's `if (!ctx) return 401` (`api.ts:1058`) is reached → **fails closed**. Worst case (an unexpected throw) is a 500 via `onError` — still not access-granted.
- **"signup `disabled` is report-only / hook is a no-op hole"** — incorrect as a vuln. `instance.ts:95` sets `disableSignUp: signupDisabled`, which Better Auth enforces before the hook. The hook being a no-op in `disabled` mode is a real *defense-in-depth* gap (S1) + a stale comment (I2), not an open door.
- The "header admitted during a session-DB outage" concern is real but is the **pre-existing, documented header-trust model** (only safe behind a loopback proxy). It's folded into **I4** as a doc/hardening item, not a new critical.

### Important (5)

| I1 | type-design | Web `AuthStatus.signup` is still `'allowlist' \| 'open'` — missing the new `'disabled'`. Server now defaults to `disabled`, so `LoginPage` hits the `else` branch: it shows "Create an account to continue" **and** the "Sign up" toggle even when signup is off (the attempt then 403s). Add `'disabled'` to the type and hide the signup affordance for it. | `packages/web/src/lib/api.ts:76`, `LoginPage.tsx:76-77,145` |
| I2 | comment, code-review, silent-failure | Stale/false comment "An empty allowlist means open signup" — the default for an empty allowlist is now `disabled`. | `packages/server/src/auth/instance.ts:106` |
| I3 | comment | `resolveAuthContext`'s comment says it's "never an access gate" — but this commit wires it into the `/api/*` gate (`api.ts:1057`). Update it. | `packages/server/src/routes/api.ts:1097-1099` |
| I4 | code-review (downgraded from Critical) | The gate admits the trusted `X-Archon-User` header, and `security.md` tells operators to "drop the Caddy `forward_auth` sidecar" to make Better Auth "the real access boundary." If the proxy was what stripped/controlled that header, dropping the sidecar makes it forgeable → gate bypass. Add an explicit "strip `X-Archon-User`/`ARCHON_WEB_AUTH_HEADER` at the proxy" warning to the runbook (and a `SECURITY:` note at the gate). | `security.md:99`, `api.ts:1053-1060` |
| I5 | docs-impact | `security.md:99` runbook says to "confirm `/api/health`, a webhook, and the SSE stream still work" after dropping the sidecar — but `/api/stream/*` is under `/api/*` and **is now gated** (401 without a session). It's not an unaffected/open endpoint. | `security.md:99` |
| I6 | docs-impact | `docker.md:323` still says Better Auth gives "allowlist-gated accounts" — the default is now `disabled` (login-only). And the form-auth section doesn't mention the gate can retire the sidecar. | `docker.md:321-377` |

### Suggestions

- **S1** (code-review, silent-failure, type-design): defense-in-depth — add a `signupDisabled` check inside the `databaseHooks.user.create.before` hook so it also blocks in `disabled` mode (today only `disableSignUp` does). `instance.ts:103-113`
- **S2** (pr-test, code-review): test gaps — gate-on + session-throws + no-header → assert **401** (pins fail-closed); `/api/auth/status` → assert `signup: 'disabled'`; fix the status-test `beforeEach` default (`'open'` → `'disabled'`, the real default). `api.auth.test.ts`
- **S3** (comment, silent-failure): the boot warn passes a prose string; use the project convention `getLog().warn({ hint }, 'web_auth.signup_disabled_no_allowlist')` so the event key is structured (and matches how `security.md` references it). `index.ts:642-646`
- **S4** (silent-failure, comment): exempt-prefix asymmetry — `'/api/auth/'` (trailing slash) means a bare `/api/auth` isn't exempt. Harmless today (Better Auth uses deeper paths) but fragile. `api.ts:1052`
- **S5** (type-design): replace the hand-written web `AuthStatus` with the generated `components['schemas']['AuthStatusResponse']` to kill the third copy of the signup union (ties to the deferred types-regen from the last round). `web/src/lib/api.ts:74`
- **S6** (silent-failure): consider failing boot **hard** (not just an error log) when web auth is enabled but the instance fails to build — the current outcome is login-down + every `/api/*` returning 401 (total lockout), which is fail-closed but silent past one log line. `instance.ts:51-61`
- **S7** (type-design): document the `allowlist`-beats-`ARCHON_AUTH_OPEN_SIGNUP` priority in the `getSignupMode` JSDoc. `config.ts:68-76`

### Strengths

- **Safe-by-default, fail-safe parsing**: signup defaults to `disabled`; the gate defaults ON when auth is enabled; opt-outs are exact-string (`ARCHON_WEB_AUTH_REQUIRED !== 'false'`, `ARCHON_AUTH_OPEN_SIGNUP === 'true'`) so typos leave the safe posture.
- **Gate fails closed** and `isApiGateEnabled` guards on `isWebAuthEnabled` first, so solo/SQLite installs can never enable it.
- `disableSignUp` is set once at construction (immutable for the process); `getSignupMode` + the Zod enum were updated in lockstep and are runtime-coupled via the handler.
- `/webhooks/*` (HMAC) and `/internal/*` (loopback) are correctly outside `/api/*`; CORS preflight resolves before the gate.
- `config.ts` helpers stay pure and are well-tested (3-state signup truth table + gate matrix).

### Recommended actions
1. **I1** (web UX bug) + **I2/I3** (false comments) + **I5/I6** (doc accuracy) — cheap, high-value.
2. **I4** — add the header-strip warning to the sidecar-retirement runbook (operational safety).
3. **S1** (defense-in-depth) + **S2** (pin fail-closed in tests).

taibui324:
# Consolidated Review: PR #1841

**Date**: 2026-06-04T05:05:55Z
**Total Findings**: 9 consolidated findings (11 source findings, 2 deduplicated)

---

## Executive Summary

PR #1841 is directionally well-structured and adds focused auth, identity, schema, UI, and documentation coverage for opt-in Better Auth web login. The main concern is that several auth-boundary paths still fail open or degrade silently: the API gate trusts a forgeable header by default, the frontend session gate renders protected UI when auth status is unknown, and Better Auth construction failures can leave the server half-enabled. Documentation and comment drift are smaller but still important because this PR changes security-sensitive deployment behavior. The test-coverage artifact could not evaluate changed-code coverage because the expected `.pr-number` and `review/scope.md` registry artifacts were missing.

**Auto-fix Candidates**: 5 HIGH issues can be auto-fixed or addressed directly
**Manual Review Needed**: 4 MEDIUM + LOW issues require decision

---

## Statistics

### Source Findings by Agent

| Code Review | 0 | 1 | 1 | 0 | 2 |
| Error Handling | 0 | 2 | 1 | 0 | 3 |
| Test Coverage | 0 | 1 | 0 | 0 | 1 |
| Comment Quality | 0 | 0 | 2 | 1 | 3 |
| Docs Impact | 0 | 1 | 1 | 0 | 2 |
| **Total Source Findings** | **0** | **5** | **5** | **1** | **11** |

### Consolidated Findings

| Severity | Count |
|----------|-------|
| CRITICAL | 0 |
| HIGH | 5 |
| MEDIUM | 3 |
| LOW | 1 |
| **Total** | **9** |

---

## CRITICAL Issues (Must Fix)

No CRITICAL issues were reported.

---

## HIGH Issues (Should Fix)

### Issue 1: API Gate Can Be Bypassed With A Forged Trusted Header

**Source Agent**: code-review
**Location**: `packages/server/src/routes/api.ts:1112`
**Category**: security

**Problem**:
When web auth and the API gate are enabled, `resolveAuthContext()` accepts `X-Archon-User` by default as a full identity after the Better Auth session check. Any caller who can reach Archon directly, or through a proxy that does not strip this header, can forge the header and satisfy the `/api/*` gate without a Better Auth session.

**Recommended Fix**:
```typescript
const allowTrustedHeader =
  !isWebAuthEnabled() ||
  process.env.ARCHON_TRUST_WEB_AUTH_HEADER === 'true';

if (!allowTrustedHeader) return undefined;

const headerName = process.env.ARCHON_WEB_AUTH_HEADER || 'X-Archon-User';
const headerVal = c.req.header(headerName)?.trim();
```

**Why High**:
This weakens the new server-side access boundary and is especially risky for identity-bound GitHub connection endpoints. Header-based auth should remain available only for deliberate trusted-proxy deployments.

---

### Issue 2: Frontend Auth Gate Fails Open When Status Fetch Fails

**Source Agent**: error-handling
**Location**: `packages/web/src/components/auth/SessionGate.tsx:30`
**Category**: unsafe-fallback

**Problem**:
`SessionGate` treats missing `status` data like `{ enabled: false }`. If `GET /api/auth/status` fails after React Query retries, an auth-enabled deployment can render the protected app shell without confirming whether login is required.

**Recommended Fix**:
```typescript
const {
  data: status,
  isPending: statusPending,
  isError: statusError,
  refetch,
} = useQuery({
  queryKey: ['auth-status'],
  queryFn: getAuthStatus,
  staleTime: 5 * 60 * 1000,
});

if (statusPending) {
  return <FullScreenLoader />;
}

if (statusError || status === undefined) {
  return <AuthGateError onRetry={() => void refetch()} />;
}

if (status.enabled === false) {
  return <>{children}</>;
}
```

**Why High**:
The UI must distinguish "auth disabled" from "auth status unknown." Failing open on an auth posture request creates misleading access behavior and downstream 401s.

---

### Issue 3: Enabled Web Auth Can Be Silently Disabled After Instance Build Failure

**Source Agent**: code-review, error-handling
**Location**: `packages/server/src/auth/instance.ts:51`
**Category**: silent-failure

**Problem**:
`getAuth()` catches Better Auth construction failures, logs once, caches `null`, and lets the process continue. `/api/auth/status` can still report `enabled: true`, while Better Auth routes are absent and session checks cannot work.

**Recommended Fix**:
```typescript
export function assertWebAuthInstanceAtBoot(env: NodeJS.ProcessEnv = process.env): void {
  if (!isWebAuthEnabled(env)) return;
  const auth = getAuth(env);
  if (!auth) {
    throw new Error('Web auth is enabled but the Better Auth instance could not be built');
  }
}

export function getAuth(env: NodeJS.ProcessEnv = process.env): AuthInstance | null {
  if (cached !== undefined) return cached;
  if (!isWebAuthEnabled(env)) {
    cached = null;
    return cached;
  }

  cached = buildAuth(env);
  log.info('web_auth.instance_built');
  return cached;
}
```

**Why High**:
Enabled web auth is security-critical configuration. A half-enabled state can lock operators out and is harder to diagnose than a clear boot failure.

---

### Issue 4: Review Scope and PR Diff Were Unavailable To Test Coverage Agent

**Source Agent**: test-coverage
**Location**: `/Users/taibui/.archon/workspaces/coleam00/Archon/artifacts/runs/b0c1a5e85ea7c8f1fe100a614796862e/.pr-number` / `/Users/taibui/.archon/workspaces/coleam00/Archon/artifacts/runs/b0c1a5e85ea7c8f1fe100a614796862e/review/scope.md`
**Category**: missing-test

**Problem**:
The coverage review could not identify changed source files, changed test files, or intended scope limits because `.pr-number` and `review/scope.md` were absent. The local branch also appeared identical to `origin/dev` to that agent.

**Recommended Fix**:
```bash
# Restore registry inputs and rerun the coverage review against the real PR diff.
printf '1841\n' > /Users/taibui/.archon/workspaces/coleam00/Archon/artifacts/runs/b0c1a5e85ea7c8f1fe100a614796862e/.pr-number
# Regenerate review/scope.md from PR metadata or workflow scope, then rerun test-coverage-agent.
```

**Why High**:
Coverage could not be meaningfully evaluated. This should not be interpreted as approval of test coverage.

---

### Issue 5: API Reference Still Says The API Has No Authentication

**Source Agent**: docs-impact
**Location**: `packages/docs-web/src/content/docs/reference/api.md`
**Category**: outdated-docs

**Problem**:
The API reference still says there is no API authentication, omits `GET /api/auth/status`, and does not document the new `mine` query parameters for conversations and workflow runs.

**Recommended Fix**:
```markdown
## Authentication

By default, Archon's API is unauthenticated. If optional Web UI login is enabled
(`DATABASE_URL` + `BETTER_AUTH_SECRET`), the server gates every `/api/*`
request with a Better Auth session or trusted web identity and returns `401`
when no identity resolves. `/api/auth/*` and `/api/health*` remain public.

## Auth

| Method | Path | Description |
|--------|------|-------------|
| GET | `/api/auth/status` | Web auth availability and signup posture; returns `{ enabled, signup }` |
```

**Why High**:
This is operator-facing security behavior. Consumers and deployers need accurate docs for auth defaults, public endpoints, and non-enforcing personal filters.

---

### Issue 1: Sign Out Failure Has No User Feedback Or Logging

**Source Agent**: error-handling
**Location**: `packages/web/src/components/layout/TopNav.tsx:26`

**Problem**:
`handleSignOut()` awaits `signOut()` without a catch block, while the click handler discards the promise. A network or server failure can leave the session active with no visible feedback.

**Options**:

| Fix Now | Add local pending/error state, catch failures, disable repeated clicks, and show retryable feedback | LOW | Users may believe signout succeeded when the session remains valid |
| Create Issue | Defer signout UX hardening to a follow-up PR | LOW | Security-sensitive auth action remains opaque until follow-up |
| Skip | Accept current behavior | NONE | Unhandled promise and poor user feedback remain |

**Recommendation**: Fix now. The change is small and auth signout should be explicit about failure.

---

### Issue 2: Combined And Bundled Schema Banners Still Say 12 Tables / Migrations 001-020

**Source Agent**: comment-quality
**Location**: `migrations/000_combined.sql:2`, `packages/core/src/db/bundled-schema.generated.ts:13`

**Problem**:
The combined SQL banner and generated bundled schema still advertise the old migration range and `12 Tables`, even though this PR adds `remote_agent_users.role` and four Better Auth tables.

**Options**:

| Fix Now | Update `migrations/000_combined.sql` to 16 tables and regenerate bundled schema | LOW | Maintainers may miss auth tables during schema audits |
| Create Issue | Track schema-comment cleanup separately | LOW | Generated schema remains misleading in this PR |
| Skip | Leave comments stale | NONE | Documentation rot in source-of-truth schema |

**Recommendation**: Fix now. This is low-risk and keeps generated artifacts aligned with source.

---

### Issue 3: Architecture Reference Database Schema Is Stale After Auth Expansion

**Source Agent**: docs-impact
**Location**: `packages/docs-web/src/content/docs/reference/architecture.md`

**Problem**:
The architecture reference still describes a 10-table schema and omits the `role` seam, `workflow_node_sessions`, `user_github_tokens`, and Better Auth tables. This conflicts with the updated database reference.

**Options**:

| Fix Now | Replace the stale schema inventory with a compact pointer to the database reference plus auth identity seam details | LOW | Developers get conflicting architecture and database docs |
| Create Issue | Defer architecture doc refresh | LOW | Stale high-level docs remain after security-sensitive schema change |
| Skip | Accept inconsistency | NONE | Continued confusion when reconciling migrations and docs |

**Recommendation**: Fix now. Keep architecture concise and make database docs canonical.

---

| GitHub device-flow note still describes header-only web identity | `CLAUDE.md:857` | comment-quality | Update wording to "Better Auth session first, then trusted `X-Archon-User` fallback"; say 401 when no web identity resolves |

---

## Positive Observations

- The PR adds focused tests for auth status, API gate behavior, `?mine` filtering, session/header precedence, and signup posture.
- The `role` primitive is centralized in the core user schema and exported through existing package boundaries.
- Better Auth routing avoids shadowing Archon-owned `/api/auth/status` and `/api/auth/github*` paths.
- The login route remains outside `SessionGate`, preventing auth redirects from making login unreachable.
- Configuration, database, security, and Docker docs already cover Postgres-only web auth, signup posture, trusted-header caveats, and non-enforcing `?mine=true`.
- Comments around `resolveAuthContext` and `?mine` explain the intended security posture rather than restating code.

---

## Suggested Follow-up Issues

If not addressing in this PR, create issues for:

| "Harden signout failure handling in Web UI" | P2 | MEDIUM issue #1 |
| "Refresh architecture schema overview after Better Auth tables" | P2 | MEDIUM issue #3 |
| "Update CLAUDE.md GitHub device-flow identity wording" | P3 | LOW issue #1 |
| "Restore review registry artifacts for coverage review reproducibility" | P1 | HIGH issue #4 |

---

## Next Steps

1. **Auto-fix step** should address 5 HIGH issues.
2. **Review** the MEDIUM issues and decide: fix now, create issue, or skip.
3. **Consider** the LOW issue for a quick documentation cleanup.
4. **Rerun** test coverage review after restoring `.pr-number`, `review/scope.md`, and the PR diff context.

---

## Agent Artifacts

| Code Review | `code-review-findings.md` | 2 |
| Error Handling | `error-handling-findings.md` | 3 |
| Comment Quality | `comment-quality-findings.md` | 3 |
| Docs Impact | `docs-impact-findings.md` | 2 |

---

## Metadata

- **Synthesized**: 2026-06-04T05:05:55Z
- **Artifact**: `/Users/taibui/.archon/workspaces/coleam00/Archon/artifacts/runs/b0c1a5e85ea7c8f1fe100a614796862e/review/consolidated-review.md`
- **PR Registry Note**: `.pr-number` and `review/scope.md` were missing. PR #1841 was confirmed with `gh pr view 1841` and was also inferred by four review agents.

PR #1842, opened 2026-06-02
feat(providers,workflows): surface session-resume outcome (resumed) + warn on cold resume

Wirasm:
## Summary

- **Problem:** a failed session resume was a **silent failure**. When a provider was asked to resume a session it couldn't (Codex thread gone, Pi JSONL missing/cwd-mismatch, OpenCode not found), it quietly started a fresh/cold session and finished the turn — the executor had no way to know the node ran without its expected prior context.
- **Why it matters:** `persist_session` (and the planned suspend/resume primitive) depend on session continuity. A silent cold resume produces a degraded result that looks like a normal success — it must be observable, not swallowed.
- **What changed:** (1) a structured **`resumed?: boolean`** on the result `MessageChunk`, fed by each provider's own resume signal via a shared `withResumedOutcome` wrapper (and a `resumedOutcome` guard helper); (2) the dag-executor **surfaces a user-visible warning** when a requested resume comes back cold, and persists the cold run's own (already-fresh) session id for next time.
- **What did NOT change (scope boundary):** no provider behavior on the happy path (warm resume / no-resume are untouched); the cold-resume branch only fires for a **completed** node with a **requested** resume that came back `false`. No DB migration. The full suspend/resume primitive (checkpointing) remains future work.

## Why no replay

An earlier revision of this PR re-ran ("replayed") a cold-resumed node from a clean start. The multi-agent review showed this is **redundant**: every provider's cold-resume fallback is *already* a clean fresh session — Codex `startThread`, Pi `SessionManager.create`, OpenCode resolves a fresh id, and Claude resume-or-errors (never goes cold). Re-running such a node just repeats the same fresh result at **double the cost, double the side effects** (write-heavy nodes), and double-streamed output. So the cold run is kept and its fresh session id persisted; the loss of continuity is surfaced loudly instead.

## `resumed` semantics

| value | meaning |
|---|---|
| `true` | resume requested **and** the prior session was restored (warm) |
| `false` | resume requested but the provider fell back to a fresh (cold) session |
| omitted | no resume was requested |

Per-provider source: Claude (resume-or-error ⇒ `true`), Codex (`!sessionResumeFailed`), Pi (`!resumeFailed`), OpenCode single-agent (`resolveSessionId.resumed`), OpenCode multi-agent (always `false` when a resume was requested — multi-agent never resumes). The `resumedOutcome(resumeSessionId, succeeded)` helper centralizes the "undefined when no resume was requested" guard for all five call sites.

### Before

```
node (persist_session) ─ resume 'sess-123' ─▶ provider: session gone, starts FRESH (silent)
                                              └─ node completes COLD; output looks normal  ✗
```

### After

```
node (persist_session) ─ resume 'sess-123' ─▶ provider: resumed=false (already a fresh session)
   executor ─ ⚠️ "could not resume the prior session — continued with a fresh session"  (user-visible)
            └─ keep the fresh run, persist its fresh id for next time  ✓  (no re-run)
```

## Change Metadata

- Change type: `feature`
- Primary scope: `multi` (providers + workflows)
- Risk: `risk: medium` · Size: `size: M`

## Linked Issue

- Related: cross-provider warm-resume / workflow suspend-resume effort (follows #1830, #1840). No single tracking issue.

## Validation Evidence

```bash
bun run type-check    # ✅ providers + workflows
bun run lint          # ✅ clean
bun run format:check  # ✅ clean
bun test packages/providers/src/shared/resumed.test.ts                 # ✅ 7
bun test packages/providers/src/claude/provider.test.ts                # ✅ 79
bun test packages/providers/src/codex/provider.test.ts                 # ✅ 64
bun test packages/providers/src/community/pi/provider.test.ts          # ✅ 67
bun test packages/providers/src/community/opencode/provider.test.ts    # ✅ 31 (now wired into CI)
bun test packages/workflows/src/dag-executor.test.ts                   # ✅ 263
```

- New/updated tests: `withResumedOutcome` + `resumedOutcome` (7); Codex `resumed:true/false`; Claude `resumed:true` on resume / omitted otherwise; Pi warm+cold; OpenCode warm+cold; executor cold-resume (one `sendQuery` call, warning surfaced, cold run's fresh id persisted) and warm-resume (exactly one call, no replay).
- **CI gap closed:** `opencode/provider.test.ts` was never in the providers test script — now added.

## Security Impact

- New permissions/capabilities? No. New external network calls? No. Secrets handling changed? No. FS access scope changed? No.

## Compatibility / Migration

- Backward compatible? **Yes** — `resumed` is optional; consumers that ignore it are unaffected. `IAgentProvider` unchanged. No DB migration (`nodeOutputSchema` gains an optional field).
- Config/env changes? No.

- Affected: all four providers' `sendQuery` result emission (additive wrapper) and the dag-executor `persist_session` path.
- A genuinely cold resume now costs **no extra node run** (the cold run is kept); the only added behavior is a warn log + a user-visible message.
- Guardrails: `dag.session_resume_failed` warn log (includes `resumeSessionId`) + a user-visible platform message on every cold resume.

## Rollback Plan

- Fast rollback: revert the commits (no migration, no persisted-state change). The provider wrapper and the executor branch are independently revertible.
- Observable failure symptoms: the warning not appearing on a known cold resume, or a provider mis-reporting `resumed:false`.

## Risks and Mitigations

- Risk: a provider reports `resumed:false` spuriously → an unnecessary warning (no cost, since there's no replay) → Mitigation: each provider derives it from the same signal that already drove its "Could not resume" system message; unit-tested per provider.
- Risk: context isn't recovered (the session is gone) → Mitigation: by design — the fresh run is a clean deterministic start and the loss is surfaced loudly. True context recovery is the job of the future checkpoint/suspend primitive, not this PR.

## Summary by CodeRabbit

* **New Features**
  * Session resumption outcomes are now surfaced on `result` chunks and node outputs via an optional `resumed` flag (`true` when restored, `false` when falling back).
* **Bug Fixes**
  * When a persisted session can’t be restored, the executor warns the user, performs a single cold run (no replay), and persists the new session id.
* **Tests**
  * Added/expanded regression coverage for resumed-success, resumed-failure, and “no result” behavior across providers and the DAG executor.
* **Documentation**
  * Updated community provider guidance and persistent-session documentation for cold-resume surfacing.


Wirasm:
## PR Review Summary — #1842 (resumed outcome + replay-on-cold)

Seven specialized agents reviewed the diff (code, docs, tests, comments, silent-failures, types, simplification). I independently verified the high-impact findings against the source. Net: the core seam (`withResumedOutcome` + the `resumed` signal + executor replay) is well-built and well-tested for the **Claude/Codex single-agent happy path**, but there are a few real gaps and one design question worth resolving before merge.

### Critical / Blocking

| # | Issue | Location |
|---|-------|----------|
| **C1** | **OpenCode multi-agent path bypasses the wrapper entirely.** The PR wraps only the single-agent `streamOpencodeSession` call. The `isMultiAgent` branch `yield*`s `streamMultiAgentOpencodeSession` and `return`s *before* reaching `withResumedOutcome` (verified: lines 130-146 return early; the wrapper is at ~160). A `persist_session` node using OpenCode `agents:` that cold-resumes emits `resumed: undefined` → no warning, no replay → the exact silent failure this PR exists to kill, still live for that path. | `providers/.../opencode/provider.ts:130-146` |
| **C2** | **OpenCode provider test is not in the providers test script.** Verified: `packages/providers/package.json` `test` does not run `opencode/provider.test.ts`. So this PR's OpenCode change is **unverified by CI**, and the existing `toEqual` assertion in that file is now stale (the result chunk gains `resumed: false`, so it would fail if it were run). Add the file to the script and update the assertion. | `providers/package.json`, `opencode/provider.test.ts:342` |

### Important

| # | Issue | Location |
|---|-------|----------|
| **I1** | **Replay double-emits `node_started` + `node_completed`.** Verified: `executeNodeInternal` emits both unconditionally (607-630, 1176-1212); the replay calls it a second time with no suppression guard. Result: duplicate `workflow_events` rows + duplicate SSE emissions for one logical node (event-log noise; the events-fallback completed-count in the web UI over-counts). | `workflows/.../dag-executor.ts` (replay block) |
| **I2** | **Cold run's cost is dropped from `totalCostUsd`.** Verified: after replay, `output` is reassigned and the layer accumulator (`totalCostUsd += output.costUsd`, ~3239) only sees the replay's cost. The cold run was a real billed call; its cost vanishes. Capture `output.costUsd` before the replay and add it back. | `workflows/.../dag-executor.ts` |
| **I3** | **The replay may be redundant double-execution — please confirm intent.** Every provider's cold fallback is *already a clean fresh session*: Codex `resumeThread` fails → `startThread()` (fresh, verified at provider.ts:758/766); Pi/OpenCode resolve to a fresh session; Claude resume-or-errors (never cold). There is no "half-resumed" state today — so re-running the node yields the same fresh result at double cost and **double side-effects for write-heavy nodes**. Consider: surface the warning + persist the cold run's clean session id, and skip the replay (or gate the replay behind a provider capability for genuinely-resumable-but-degraded cases). The PR body justifies replay as avoiding an "ambiguous half-resume," but no current provider produces one. | `workflows/.../dag-executor.ts` (replay block) |
| **I4** | **Missing `resumed` assertions for 3 of 4 providers.** Only Codex asserts `resumed: true/false`. Claude (`true` whenever resume requested), Pi (`!resumeFailed`), and OpenCode (`resumed`) have zero coverage on the stamped field — an inverted conditional in any of them would silently break the executor's replay guard with no CI signal. Two-line `toMatchObject({ resumed })` additions to existing resume tests. | `claude/provider.test.ts`, `community/pi/provider.test.ts`, `opencode/provider.test.ts` |

### Suggestions

| # | Suggestion | Location |
|---|-----------|----------|
| **S1** | Replay-failure isn't distinguished in logs. If the replay itself returns `state:'failed'`, it falls through to the generic `dag_node_pre_execution_failed` path — no "the **replay** failed" signal. One distinguishing warn line saves a future debugging session. | `dag-executor.ts` |
| **S2** | The `dag.session_resume_failed_replaying` warn log omits the dead `resumeSessionId` — the first thing you'd want when debugging "why didn't it resume." It's in scope and non-undefined at the call site. | `dag-executor.ts` |
| **S3** | Warm-resume dag-executor test doesn't assert `mockSendQueryDag.mock.calls.length === 1`. The cold test asserts `=== 2`; making the warm test assert `=== 1` makes the one-replay bound bidirectional and guards against a future `if (!output.resumed)` regression. | `dag-executor.test.ts` |
| **S4** | `types.ts` comment: "omitted — no resume was requested, **or the provider can't determine it**" — the second clause describes a path no provider takes (all pass `undefined` only when `resumeSessionId === undefined`). Drop it to avoid implying a third semantic. | `providers/.../types.ts:208` |
| **S5** | Tri-state `resumed?: boolean` relies on strict `=== false` everywhere (correct today, fragile tomorrow — `if (!resumed)` would conflate cold with not-attempted). A `'warm' \| 'cold'` literal union would self-document and remove the trap; all call sites are in this PR's scope, no migration cost. Optional. | `types.ts`, `resumed.ts`, `schemas/workflow-run.ts` |
| **S6** | `code-simplifier`: replace the `yield … ? … : …` ternary in `withResumedOutcome` with an explicit `if/else`; the four identical `resumeSessionId !== undefined ? X : undefined` ternaries could collapse into a `resumedOutcome(id, ok)` helper (Rule-of-Three, optional). | `resumed.ts` + 4 providers |

### Documentation

- `packages/docs-web/.../reference/architecture.md` — the `MessageChunk` `result` block is now stale (missing `resumed?: boolean`).
- `packages/docs-web/.../guides/authoring-workflows.md` — "Persistent Sessions" section should note the new user-visible cold-resume behavior (warning + extra node run).
- `packages/docs-web/.../contributing/adding-a-community-provider.md` — skeleton comment should point new providers at `withResumedOutcome` so they don't silently omit the signal (cf. C1).

### Strengths

- `withResumedOutcome` is clean, single-responsibility, and thoroughly unit-tested (undefined/true/false + defensive multi-result). Type-design rated it **8.5/10**.
- The one-replay bound is correctly *structural*: the replay passes `resumeSessionId=undefined`, so its output carries no `resumed` field and the guard can't re-fire — not reliant on a counter.
- Schema/type consistency is correct (`z.infer`, `@hono/zod-openapi`, field scoped to the `completed/running` variant only).
- The persist path correctly reads the **replay's** clean session id (verified by the new test), not the dead one.
- Per-provider signal derivation is individually correct (Claude `true` past the throw boundary; Codex `!sessionResumeFailed`; Pi `!resumeFailed`; OpenCode `resumed`).

### Verdict: **NEEDS FIXES**

Resolve **C1** (multi-agent bypass) and **C2** (untested OpenCode path) before merge, and decide **I3** (is the replay's double-execution intended, or is surface+persist sufficient?) — it shapes whether I1/I2 even matter. I1/I2/I4 and the suggestions are quick follow-ups. The foundational seam is sound; these are edges, not a rethink.

<sub>Reviewed by code-reviewer · docs-impact · pr-test-analyzer · comment-analyzer · silent-failure-hunter · type-design-analyzer · code-simplifier. High-impact findings independently verified against source.</sub>

Issue #1849, opened 2026-06-03
feat(providers,workflows): reliable cross-provider structured output (validate + reask + fail-fast) for typed handoffs


Wirasm:
## Problem

Typed handoffs (`$nodeId.output.field`, and any "typed artifact" with machine-readable fields) need the provider to actually return schema-conformant JSON. Two gaps make this unreliable today:

1. **No schema validation — for ANY provider.** `tryParseStructuredOutput` (`packages/providers/src/shared/structured-output.ts`) only checks "is it a non-null object?" Field presence/types are never validated against the declared `output_format` schema. A missing `$nodeId.output.field` **silently degrades to `''`** after a warn (`dag-executor.ts:298-337`). That's a silent failure (violates fail-fast) and the core blocker for trusting typed handoffs.

2. **Provider enforcement is wildly uneven** (verified in-code + installed SDKs):

| Provider | Path | Guarantee |
|---|---|---|
| Claude | SDK `output_config.format` / json_schema | **Enforced (hard)** |
| Codex | SDK `outputSchema` (TurnOptions) | **Enforced (hard)** |
| OpenCode | `format:{json_schema}` in request body | **Enforced (backend)** |
| Pi | prompt augmentation + post-hoc extraction | **best-effort, unvalidated** |
| Copilot | same shared augmentation + extraction | **best-effort, unvalidated** |

**Pi specifically** (the ~June-2026 main runner): the installed SDK (`@earendil-works/pi-coding-agent` 0.76.0) exposes **no** native `response_format`/`json_schema` on any API. Upstream issue badlogic/pi-mono#1086 tracked adding it — closed but undocumented; treat as absent. Pi's idiomatic options are **terminate-on-tool** (final tool call with schema-typed args → near-guarantee) or an **extension** (`before_provider_request` to inject `response_format` for OpenAI-compatible backends; `message_end` to validate/replace) — but extensions are off-by-default in Archon, with no per-call schema plumbing and **no reask primitive**.

## Proposed Solution — a layered, capability-gated contract

Native enforcement is the fast path; a portable best-effort path is the floor. Strongest→weakest:

1. **Capability-gated native enforcement** — keep using SDK enforcement for Claude/Codex/OpenCode (hard guarantee, no overhead).
2. **`json-repair`** before parsing — strip fences, fix trailing commas/truncation (best-effort syntax normalization).
3. **Schema validation (zod/ajv)** against the declared schema — *currently missing everywhere*. This is the keystone fix.
[...]

Issue #1848, opened 2026-06-03
feat(workflows): node-boundary summarization + pointer injection for $nodeId.output (kill uncapped by-value bloat)


Wirasm:
> **Updated 2026-08-05 after a full read of the substitution path.** Three claims in the original body were stale or wrong; they are corrected inline below, and the investigation changed the recommended *shape* of the fix. **Read "What the investigation established" before implementing** — the original Definition of Done is superseded by the sequence at the bottom.

## Problem

`$nodeId.output` injects an upstream node's **full text by value, with no cap on any model-facing surface** (`substituteNodeOutputRefs`, `packages/workflows/src/dag-executor.ts:745`). Real cases bloat the downstream window:

- `archon-architect`'s `create-pr` node pastes `$analyze.output + $plan.output + $validate.output` whole — three multi-KB accumulations into one prompt.
- A `loop:` re-substitutes on **every iteration**, so an N-iteration loop reading a 40 KB upstream pays 40 KB × N.

This is the anti-pattern current context-engineering practice warns about — context as scarce RAM, "context rot" as token counts rise ([Anthropic: Effective Context Engineering](https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents); [Manus](https://manus.im/blog/Context-Engineering-for-AI-Agents-Lessons-from-Building-Manus)). It is the highest-ROI context fix and is **independent of session resume**.

## What the investigation established

**There is a 32 KB threshold in the tree, and it is not a solution to this.** `NODE_OUTPUT_FILE_THRESHOLD = 32_768` (`dag-executor.ts:2532`) feeds `shellQuoteOrFile` (`:709`), which fires **only when `escapedForBash === true`** — three shell call sites. What it emits is `$(cat '<path>')`, a shell substitution that reinlines the entire text at execution time, so the token count into the process is unchanged. It is an `ARG_MAX` corruption guard, exactly as its docblock says. **A prompt has never spilled and cannot spill.** Anyone grepping for a threshold will find this and mistake it for coverage.

It is still the right *template*: same threshold check, same spill-to-file, a different rendered replacement and target directory.

[...]

Issue #1846, opened 2026-06-02
feat(workflows): cross-invocation artifact scope + cold-resume pointer recovery (by-reference, not a DB blob)


Wirasm:
## Problem

When a `persist_session` node is re-invoked and the provider can't restore the prior session (Codex thread expired, Pi JSONL missing/moved, OpenCode session gone), the node runs on a **fresh** session with **zero prior context**. PR #1842 makes this loss *visible* (a warning + the cold run is kept) but does nothing to *recover* the lost context.

The context we'd want is often **already on disk** as artifacts â it's just unreachable from the new run, because artifacts are run-isolated.

## â ï¸ Reframing â this supersedes the original "node output memory" framing

This issue was originally filed as a **DB-backed node-output store** that re-injects a node's last output **by value** (a preamble) on cold resume. Deeper investigation + a survey of current context-engineering practice (Anthropic, Manus) shows that's the **wrong direction**. Keeping this note in the issue so we don't build it the wrong way.

### #1846-as-filed vs. artifacts â they're two implementations of the *same* need

| | **Artifacts (`$ARTIFACTS_DIR`)** | **Node Output Memory (original #1846)** |
|---|---|---|
| Carries | any files the model writes | a node's final **output text** |
| Into context | **by-reference** â model `Read`s on demand | **by-value** â injected as a preamble |
| Structure | freeform, model-named | freeform text blob |
| Managed by | the model (engine gives the path) | the engine |
| Scope | **per-run** (run-isolated; cross-run only via `$ARTIFACTS_DIR/../runs/*` hacks or `archon continue`) | **cross-invocation** `(workflow, node, scope, provider)` |
| Survives session loss | â on disk | â in DB |

**The honest read:** both solve *durable context that outlives the provider session.* Artifacts already do the durable, **by-reference** half well; their only real gap for the cold-resume case is that they're **run-isolated** (a fresh invocation gets an empty dir). The original #1846 bolts on a parallel DB text store **plus a by-value re-injection.**

[...]

PR #1851, opened 2026-06-03
feat(workflows): typed artifacts — output_type + engine-written sidecar metadata

### LINKED ISSUE #1847: feat(workflows): typed artifacts — output_type + engine-written sidecar metadata
## Problem

Archon's artifacts are an **undifferentiated flat directory**: the model writes files by convention, names them however it likes, and downstream nodes find them by guessing filenames or `cat`-ing hardcoded paths (e.g. `archon-synthesize-review` reads five fixed `$ARTIFACTS_DIR/review/*.md` paths). There's no machine-readable answer to "what did the plan node produce, and where?" This blocks:
- cross-invocation **pointer recovery** (#1846) — to point a cold-resumed run at "the prior plan artifact," the engine needs to know which artifact *is* the plan;
- **selective reference** — injecting a compact pointer to the right artifact instead of the whole blob.

This is the "undifferentiated directory" weakness; the fix is the 80% version of provenance-addressable artifacts ([Pattern 6 / arXiv:2508.02866 PROV-AGENT](https://arxiv.org/pdf/2508.02866)).

## Proposed Solution (provider-agnostic — does NOT depend on structured output)

Give each node an optional `output_type` label (`plan | code | findings | summary | spec | …`, open set). When a node completes, the **engine** writes a sidecar `<artifact>.meta.json` (or a per-run `artifacts.index.json`) recording: producing node id, `output_type`, file path(s), size, run id, timestamp. The *content* stays freeform (a markdown plan is still a "plan") — so this needs **no** schema-conformant model output and is unaffected by Pi's structured-output caveats (see the structured-output issue).

Downstream consumers (and #1846's pointer recovery) can then ask "give me the latest `plan` artifact for this workflow+scope" and get a real path, instead of filename guessing.

## Why this is the foundation

- #1846 (cold-resume pointer recovery) needs typed artifacts to point at the *right* prior file.
- The summarization/pointer-injection issue can reference artifacts by type.
- It's a small lift (engine writes JSON metadata) with large debuggability/selectivity gains, and it's fully provider-agnostic.

## Scope

- Package(s): `workflows`, `paths` (artifact path/index helpers)
- Breaking change? `No` (opt-in `output_type`; metadata is additive)
- Database changes? `No` (sidecar files; could later mirror to DB)
- New deps? `No`

## Definition of Done

- [ ] `output_type` accepted on DAG nodes (open string set, validated as non-empty)
- [ ] Engine writes sidecar/index metadata (producer node, type, path, size, run id, ts) after a node that declares it
- [ ] A resolver: "latest artifact of type T for (workflow, scope)" — used by #1846
- [ ] Metadata never contains secrets; same trust boundary as the artifact
- [ ] Tests: metadata written, resolver returns newest, missing-type returns nothing (no throw)
- [ ] Docs: artifacts section in `authoring-workflows.md`

## Context

Part of the "artifacts as canonical, typed, by-reference memory" direction. Companion to #1846 (pointer recovery), the summarization issue, and the structured-output issue. Follows PR #1842.

Wirasm:
## Summary

- **Problem:** workflow artifacts are an undifferentiated flat directory — the model names files however it likes, and downstream nodes find them by guessing filenames or `cat`-ing hardcoded paths. There's no machine-readable answer to "what did the plan node produce, and where?"
- **Why it matters:** this is the foundation for the rest of the context-passing epic — cold-resume pointer recovery (#1846) and selective/by-reference context passing (#1848) both need to locate a prior output *by type*.
- **What changed:** a node can declare an optional `output_type`; when set, the executor writes a typed sidecar artifact (`nodes/<id>.md` + per-node `<id>.meta.json`) after the node completes. New `artifacts-index` helper provides write + lookup (`latestNodeArtifactOfType`).
- **What did NOT change (scope boundary):** fully opt-in and additive — default behavior is identical when no node sets `output_type`. No new dependencies, no DB changes, no provider changes. The structured-output (#1849), summarization (#1848) and cold-resume-recovery (#1846) slices are separate PRs.

### Before

```
node (writes plan.md to $ARTIFACTS_DIR by its own naming) ──► downstream node
   └─ downstream must `cat $ARTIFACTS_DIR/<guessed-name>` ; no typed lookup
```

### After

```
node  output_type: plan ──► executor writes:
   $ARTIFACTS_DIR/nodes/<id>.md         (the output)
   $ARTIFACTS_DIR/nodes/<id>.meta.json  (typed metadata: type, runId, ts, size)
   └─ [latestNodeArtifactOfType(dir, 'plan')] → a real path, no filename guessing
```

### After

```
schemas/dag-node.ts  [~ output_type on base + transform]
        │
        ▼
dag-executor.ts (layer-result funnel) === [+] artifacts-index.ts
        │                                        └─ schemas/node-artifact.ts [+]
        ▼
   $ARTIFACTS_DIR/nodes/<id>.md + <id>.meta.json   [+]
```

**Connection inventory:**

| `dag-executor` | `artifacts-index` | **new** | best-effort sidecar write at node completion |
| `artifacts-index` | `schemas/node-artifact` | **new** | metadata schema |
| `schemas/index` | `schemas/node-artifact` | **new** | re-export |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `workflows`, `tests`
- Module: `workflows:dag-executor`, `workflows:schemas`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Closes #1847
- Related #1846, #1848, #1849 (context-passing epic)

```bash
bun run check:bundled / -skill / -schema   # ✅ all up to date
bun run type-check                          # ✅ all 6 packages (no cross-package ripple)
bun run lint --max-warnings 0               # ✅ 0 errors / 0 warnings
bun run format:check                        # ✅ clean
bun --filter @archon/workflows test         # ✅ all batches exit 0
bun test src/artifacts-index.test.ts        # ✅ 6
bun test src/dag-executor.test.ts           # ✅ 264 (incl. 2 new typed-artifact tests), 0 regressions
```

- Evidence: new unit suite (`artifacts-index.test.ts`) + 2 dag-executor integration tests.
- Intentionally skipped: the cross-package test sweep in `bun run validate` (unrelated packages can't be affected by an additive, workflows-only change; all-package type-check already confirms no contract breakage).

## Security Impact (required)

- File system access scope changed? **No** — writes only under the existing per-run `$ARTIFACTS_DIR`; node id is sanitized to a safe path segment (no traversal).

## Compatibility / Migration

- Backward compatible? **Yes** — `output_type` is optional; absent → no sidecar, identical behavior.
- Config/env changes? **No**

- Verified: a node with `output_type` writes `nodes/<id>.md` (exact output text) + `<id>.meta.json` (type/runId/path); a node without `output_type` writes nothing; corrupt/missing metadata is non-fatal (skip + warn); a node id with separators is sanitized and the original id preserved in metadata.
- Edge cases checked: empty/missing nodes dir → `[]`; `latestNodeArtifactOfType` picks newest by ISO timestamp; per-node files avoid parallel-layer write contention.
- Not verified beyond unit/integration: live end-to-end run writing to `~/.archon/workspaces/.../artifacts/runs/<id>/nodes/` against a real provider (mechanism is identical to the tested in-memory path).

- Affected: the dag-executor layer-result loop (one best-effort write per completed node that declares `output_type`).
- Potential unintended effects: an extra small file write per typed node; failures are caught and logged (`artifacts.write_failed`) and never fail the node.
- Guardrails: write is wrapped in try/catch with a warn log; read path tolerates corrupt/missing entries.

## Rollback Plan (required)

- Fast rollback: revert the single commit. No migration, no persisted-state dependency; existing runs are unaffected (sidecars are additive files).
- Feature flags: none needed — the feature is inert unless a node sets `output_type`.
- Observable failure symptoms: `artifacts.write_failed` warn logs, or missing `nodes/<id>.meta.json` for a typed node.

## Risks and Mitigations

- Risk: shared-index write contention in parallel layers → **Mitigation:** per-node `.meta.json` files (no shared index); the index is derived on read by globbing.
- Risk: a malicious/odd node id escaping the nodes dir → **Mitigation:** `safeSegment()` strips path separators; covered by a test.

## Summary by CodeRabbit

* **New Features**
  * Nodes can declare output_type to persist typed sidecar artifacts (content + metadata) that can be looked up by type; writes are best-effort and won’t fail the workflow (failures are logged).

* **Tests**
  * Added comprehensive tests covering write/read/filtering, collision handling, overwrite/resume behavior, and edge cases.

* **Documentation**
  * Guides and docs updated to describe output_type, artifact files/metadata, lookup semantics, and failure behavior.


Wirasm:

Reviewed by 7 specialized agents (code, docs, tests, comments, errors, types, simplify) against the diff of `feat/typed-artifacts`. The change is small, additive, opt-in, and inert unless a node sets `output_type` — the design is sound and the happy path is well-tested. Findings below are about hardening the invariants *now*, since this is the foundation for the rest of the context-passing epic (#1846, #1848, #1849).

Finding IDs use `C/I/S` prefixes (not bare hash-numbers) to avoid auto-linking unrelated issues.

> Note: branch is **14 commits behind `dev`** but GitHub reports `MERGEABLE` (no conflicts). Rebase optional.

---

### Critical Issues (0 found)

None merge-blocking. The issues below are all latent (not triggerable by any existing workflow) and cheap to fix.

---

### Important Issues (5)

| ID | Agent(s) | Issue | Location |
|----|----------|-------|----------|
| **I1** | code-reviewer + pr-test-analyzer | **`safeSegment` collision → silent data loss.** Two schema-valid node ids (`plan.v1` / `plan_v1`, `a.b` / `a_b`) sanitize to the same filename; the second write silently overwrites the first `.md` + `.meta.json`. No `artifacts.write_failed` fires. Not triggerable by current workflows (all ids are `[a-zA-Z0-9-]`), but the schema permits it. **Best fix: constrain the node `id` charset** (`z.string().regex(/^[a-zA-Z0-9_-]+$/)`) in `dagNodeBaseSchema` — turns a silent runtime collision into a parse-time error and removes the need for `safeSegment` guessing. | `artifacts-index.ts:21-23` / `schemas/dag-node.ts` (id field) |
| **I2** | type-design + comment-analyzer | **`producedAt` sort invariant unenforced.** `latestNodeArtifactOfType` relies on lexicographic ISO-8601 ordering, but the schema is `z.string()` — a non-UTC offset (`+05:00`) or hand-edited value silently returns the *wrong* artifact (no error). One-token fix: `producedAt: z.string().datetime()`. `new Date().toISOString()` always passes, so zero false rejections on the write path; corrupt entries get skipped by the existing `safeParse`. | `schemas/node-artifact.ts:19` |
| **I3** | silent-failure-hunter | **Bare `catch {}` on `readdir` swallows non-ENOENT errors.** `EACCES`/`EPERM`/`ENOTDIR`/`EIO` are all silently returned as `[]` — no log, no signal. Diverges from the established package pattern (`script-discovery.ts`, `workflow-discovery.ts` both check `err.code === 'ENOENT'` first). Fix: return `[]` only on ENOENT; warn + rethrow otherwise. | `artifacts-index.ts:67-70` |
| **I4** | pr-test-analyzer (9/10) | **Best-effort write-failure path is untested.** The entire safety contract is "a metadata write never fails a successful node" — there's no test verifying the node still completes when `writeNodeArtifact` throws. A future refactor removing the try/catch would pass CI. Add a test with a broken `artifactsDir`. | `dag-executor.ts:3248-3266` |
| **I5** | docs-impact | **`output_type` is undocumented.** Missing from `CLAUDE.md` (workflow node-field list + engine-schema list + artifacts dir layout), `AGENTS.md` (1:1 copy — must mirror), and `packages/docs-web/.../authoring-workflows.md` (Common fields table + Artifact Chain section). It's valid on **all** node types, so it belongs in *Common fields*, not AI-only. | see docs-impact detail |

---

### Suggestions (8)

| ID | Agent(s) | Suggestion | Location |
|----|----------|------------|----------|
| **S1** | pr-test-analyzer (8/10) + comment-analyzer | Add a **bash/script node `output_type` test** — the schema comment promises bash/script support (no `sessionId`), but only an AI `command:` node is tested. Verifies the documented claim and guards against an accidental AI-only gate. | `dag-executor.test.ts` |
| **S2** | silent-failure-hunter | Include `parsed.error.issues` in the `artifacts.index_entry_invalid` warn — currently logs only the filename, leaving schema-drift drops undebuggable. | `artifacts-index.ts:86-88` |
| **S3** | pr-test-analyzer (7/10) | Assert `sessionId` in the existing integration test — the mock yields `sessionId: 'new-session-id'` but the `toMatchObject` omits it, so an inverted spread guard would be invisible. One-line addition. | `dag-executor.test.ts:8045` |
| **S4** | comment-analyzer | **Fix the misleading "concurrent contention" comment.** Writes happen in a *sequential* loop after `Promise.allSettled` — no concurrent write ever occurs, so the per-node-file design isn't what prevents contention (it prevents *cross-run* clobbering + enables by-id lookup). Reword so no one wrongly assumes it's safe to move writes inside the fan-out. | `artifacts-index.ts` (writeNodeArtifact JSDoc) |
| **S5** | type-design + simplify | Tighten schema for consistency with sibling `artifactFileSchema`: `size: z.number().int().nonnegative()` and `outputType: z.string().min(1)` (matches `output_type` on the node config). | `schemas/node-artifact.ts:14,21` |
| **S6** | simplify (strong) | Drop the **lazy-logger memoization** — the comment claims it's "deferred so test mocks can intercept," but `artifacts-index.test.ts` doesn't mock `createLogger`. Use the module-level `const log = createLogger('artifacts-index')` like the rest of the leaf utilities. | `artifacts-index.ts:6-10` |
| **S7** | simplify + type-design | Replace the inline params type with `Omit<NodeArtifact, 'path' \| 'size'>` (the two computed fields) — same pattern already used in `logger.ts`. Self-documenting and structurally safe if `NodeArtifact` grows. | `artifacts-index.ts:34-42` |
| **S8** | simplify | Minor: pass `sessionId: output.sessionId` directly at the call site (the conditional-spread guard already lives correctly *inside* `writeNodeArtifact`); and `reduce`→`for`-loop in `latestNodeArtifactOfType` for clarity. | `dag-executor.ts:3252-3260`, `artifacts-index.ts:100-103` |

**Debatable / deferred:** silent-failure-hunter proposed a *user-facing chat warning* on every best-effort write miss. Reasonable instinct, but there are no downstream consumers of these artifacts yet, and surfacing a chat message per miss risks noise. Recommend deferring to the PR that wires the first `latestNodeArtifactOfType` caller (#1846/#1848), where the right UX is clearer.

---

### Strengths

- Clean separation: schema (`node-artifact.ts`), helper (`artifacts-index.ts`), executor integration — each focused.
- Path-traversal sanitization (`../evil` → `___evil`) is implemented **and tested**, with the original id preserved in metadata.
- Negative test ("no `output_type` → no sidecar") prevents feature creep.
- Per-node files (no shared index) is the correct concurrency-safe design; index derived on read by globbing.
- `output_type` correctly placed on `dagNodeBaseSchema` (all node types), threaded through the transform, and verified accurate by the comment-analyzer.
- Follows Zod conventions (camelCase, `z.infer`, `@hono/zod-openapi` import, `index.ts` re-export) and the mock-isolation test-split protocol (`package.json`).
- `writeNodeArtifact` JSDoc clearly documents the throws-on-failure / caller-wraps contract.

---

### Verdict: **NEEDS FIXES** (all small; design is sound)

The PR is mergeable and the design is right. I'd land the cheap invariant-hardening before merge since everything downstream builds on artifact addressing:

1. **I2** — `producedAt: z.string().datetime()` (1 token, the only algorithmic invariant in the PR).
2. **I1** — constrain node `id` charset (closes the silent-collision class entirely).
3. **I3** — ENOENT-only swallow on `readdir` (matches existing package pattern).
4. **I4** — test the best-effort failure path (locks the sole safety property).
5. **I5** — document `output_type` (CLAUDE.md + AGENTS.md + docs-web).

Then the `S*` polish (S5–S8 are near-free). Re-run `tests`/`docs` aspects after fixes.

Wirasm:
## Review fixes applied — `55c3e940`

Addressed all findings from the multi-agent review except where deferral/rejection is the better call (rationale below). Validation: `type-check` (all 10 packages) ✅, `lint` ✅, `format:check` ✅, `bun --filter @archon/workflows test` ✅ (incl. 5 new tests).

### Fixed

| ID | Finding | Resolution |
|----|---------|------------|
| **I2** | `producedAt` sort invariant unenforced | `producedAt: z.string().datetime()` — a corrupt/non-ISO value is now rejected on read instead of silently returning the wrong "latest" |
| **I1** | `safeSegment` collision → silent data loss | Write-time **collision guard**: two ids mapping to the same segment now throw (best-effort caller logs it); first writer wins. Re-writing the same id (resume) still overwrites cleanly. *(see rejection note re: the schema-charset alternative)* |
| **I3** | bare `catch {}` on `readdir` | ENOENT → `[]`; any other fault (EACCES/ENOTDIR/EIO) now warns + rethrows, matching `script-discovery`/`workflow-discovery` |
| **I4** | best-effort failure path untested | New test: artifact write throws → node still completes |
| **I5** | `output_type` undocumented | CLAUDE.md (node-field list + `nodes/` layout + engine-schema list) and the authoring-workflows guide (Common-fields row, *Typed Artifacts* subsection, quick-ref entry) |
| **S1** | bash/script `output_type` untested | New test: bash node writes a sidecar with no `sessionId` |
| **S2** | Zod warn lacked detail | `index_entry_invalid` now logs `parsed.error.issues` |
| **S3** | `sessionId` unasserted | Added to the AI-node integration assertion |
| **S4** | misleading "contention" comment | Reworded — writes are sequential post-`allSettled`; per-node files isolate nodes/runs |
| **S5** | loose schema | `size: .int().nonnegative()`, `outputType: .min(1)` |
| **S7** | params duplicate `NodeArtifact` | `Omit<NodeArtifact, 'path' \| 'size'>` |
| **S8** | call-site spread + `reduce` | Dropped the redundant call-site `sessionId` spread (guard lives in `writeNodeArtifact`); `reduce` → explicit loop |

Also added artifacts-index tests for the collision throw, same-id overwrite, and the schema-invalid-but-valid-JSON skip branch.

### Deferred / rejected (with rationale)

- **Chat warning on best-effort write failure** (silent-failure-hunter) — *deferred.* No consumer reads these artifacts yet (cold-resume/context-passing land in #1846 / #1848); a user-facing message per miss would be noise. Right place is the first `latestNodeArtifactOfType` caller, where the UX is concrete.
- **Constraining the global node `id` charset** (the I1 "durable fix" alternative) — *rejected for this PR.* It changes validation for **every** workflow and could reject existing definitions — out of scope. The local write-time guard closes the silent-data-loss class without a breaking change.
- **Removing the lazy logger** (S6, code-simplifier) — *rejected.* `dag-executor.test.ts` mocks `@archon/paths`'s `createLogger` and imports this module transitively, so the lazy pattern is load-bearing for test isolation. The simplifier only checked `artifacts-index.test.ts` and missed the transitive mock. The comment is accurate; kept as-is.
- **AGENTS.md mirror** (I5) — *n/a on this branch.* `AGENTS.md` was added to `dev` (c732ca9f) after this branch forked; it'll pick up the `output_type` lines when the branch catches up to `dev`.

PR #1879, opened 2026-06-05
Fix: reuse Pi extension loader per process so 2nd+ session doesn't hang (#1877)

### LINKED ISSUE #1877: bug(providers/pi): 2nd+ Pi session per process hangs in extension setup — every Pi workflow node after the first idle-times-out
## Summary

The **second (and every subsequent) Pi `sendQuery()` call within a single Archon process hangs in session setup** — it produces zero events and is killed by the node idle-timeout watchdog (default 30 min). The **first** Pi call in a process works fine. Disabling Pi extensions makes the problem disappear entirely, so the hang is in the **Pi extension lifecycle** (`resourceLoader.reload()` / `session.bindExtensions()`), which is enabled by default.

Because a workflow runs in one process, **any workflow with more than one Pi AI node fails at the second Pi node** — i.e. essentially every real workflow. Single-node Pi calls (and the very first node of a multi-node run) work, which is why this hid behind "looks fine in a smoke test."

## Impact

`HIGH` (arguably `CRITICAL` for Pi). The Pi provider is effectively **unusable for real multi-node workflows** with the default config. Surfaced while benchmarking `MiniMax-M3` via Pi on the bundled `archon-fix-github-issue-*` family: `extract-issue-number` (1st Pi node) succeeded in ~3s, `classify` (2nd Pi node) hung for 30 min × retries and the whole DAG failed. This blocks the "Pi as a primary runner" direction.

## Steps to Reproduce

Minimal, deterministic — four sequential Pi nodes (no concurrency), each fail-fast:

```yaml
# .archon/workflows/experimental/minimax-seq.yaml
name: minimax-seq
provider: pi
model: minimax/MiniMax-M3   # any Pi model + a Pi backend with extensions installed
nodes:
  - id: a
    idle_timeout: 45000
    retry: { max_attempts: 1 }
    prompt: "Reply with exactly the single word: PONG"
  - id: b
    idle_timeout: 45000
    retry: { max_attempts: 1 }
    depends_on: [a]
    trigger_rule: all_done
    prompt: "Reply with exactly the single word: PONG"
  # ... c, d identical, chained ...
```

```bash
bun run cli workflow run minimax-seq --no-worktree "x"
```

**Observed:** node `a` completes (~3s); nodes `b`, `c`, `d` each time out with *"timed out with no output (idle for N min). The provider did not emit any content before the watchdog fired."*

## Root Cause

`packages/providers/src/community/pi/provider.ts` — for every `sendQuery()`, with extensions enabled (the default), the setup path runs:

- `enableExtensions = piConfig.enableExtensions !== false` → **true by default** (`provider.ts:437`)
- `resourceLoader.reload()` (`provider.ts:449-451`)
- `createAgentSession({...})` (`provider.ts:483`)
- `session.bindExtensions({...})` (`provider.ts:531-537`)
- …then it logs `pi.session_started` (`provider.ts:453`).

On the **2nd+** call in a process, one of those extension-lifecycle steps **never returns** — execution never reaches `pi.session_started`, so the bridge yields nothing and the idle watchdog (`withIdleTimeout`, resets on every chunk) fires. The first call sets up an in-process resource (port / singleton / connection / file lock) inside an extension's `session_start` (or in `resourceLoader.reload()`), and the second call's extension setup blocks on it — the Pi extension lifecycle is **not re-entrant across sessions in one process**.

Installed extensions on the repro host: `~/.pi/agent/extensions/{agent-rooms, minimax_web_tools.ts}` — one of these (or the generic `bindExtensions`/`reload` path) is the offender.

## Evidence (decisive matrix)

Same `minimax-seq` workflow, toggling only `assistants.pi.enableExtensions`:

| Node (order) | Extensions **ON** (default) | Extensions **OFF** |
|---|---|---|
| a (1st) | ✅ 3.5s | ✅ 3.7s |
| b (2nd) | ❌ idle-timeout | ✅ 7.5s |
| c (3rd) | ❌ idle-timeout | ✅ 1.9s |
| d (4th) | ❌ idle-timeout | ✅ 2.8s |
| `pi.session_started` log count | **1** (only the 1st) | **4** (all) |

`session_started: 1` with extensions on is the smoking gun — calls 2–4 never reach the log line, so they hang **in setup, before the model is ever contacted**.

## What it is NOT (ruled out)

A parallel isolation run plus the sequential run rule out every plausible alternative:

- **Not the model / API.** `MiniMax-M3` answers a tiny prompt in ~3s (first call).
- **Not prompt size.** A large plain prompt hangs as the 2nd call and a tiny one succeeds as the 1st.
- **Not `output_format`.** A plain 2nd call hangs; a structured 1st call works.
- **Not thinking level.** `effort: low` hangs too when it's not the first call.
- **Not concurrency.** The sequential (single-flight) test reproduces it; `maxConcurrent` is unset, so the provider's semaphore is not involved.
- **Not the watchdog.** The 30-min idle timeout is correct behavior; the provider genuinely emits nothing.

## Workaround

Set `assistants.pi.enableExtensions: false` in `.archon/config.yaml` (or `~/.archon/config.yaml`). Multi-node Pi workflows then run normally — at the cost of Pi extensions.

## Suggested fix direction

- Make the Pi session lifecycle **re-entrant per process**: ensure each `sendQuery()`'s `resourceLoader.reload()` / `bindExtensions()` fully tears down (and frees any singleton/port/handle) when the session is disposed, so the next session can re-init.
- Pinpoint the offending step by bisecting: disable extensions, then re-enable with each extension removed (`agent-rooms`, `minimax_web_tools.ts`) to see if it's one extension or the generic path.
- If the Pi SDK's `ExtensionRunner` holds non-re-entrant module-level state, either (a) reuse a single long-lived session/extension-runner across calls in a process instead of creating a fresh `AgentSession` each `sendQuery()`, or (b) run each Pi node in its own subprocess.
- Consider adding a much shorter default `idle_timeout` *first-token* guard for Pi specifically, so this fails in seconds rather than 30 min while the real fix lands.

## Diagnostic artifacts (in `.archon/workflows/experimental/`, untracked)

- `minimax-smoke.yaml` — single trivial prompt (proves the provider works first-call).
- `minimax-isolate.yaml` — 4 parallel nodes (output_format / size / thinking variants).
- `minimax-seq.yaml` — 4 sequential nodes (the clean repro above).

## Related

- The bundled `archon-fix-github-issue-minimax.yaml` benchmark workflow is the real-world trigger (classify = 2nd Pi node = hang).
- Blocks Pi-as-primary-runner. Also note the separate, lower-severity finding: the title-generator service throws `Pi provider requires a model` for Pi (no model passed) and falls back to a truncated title.

Wirasm:
## Summary

- **Problem:** With Pi extensions enabled (the default), the 2nd and every subsequent `PiProvider.sendQuery()` call per process hangs in session setup and idle-times-out after 30 min — so every Pi workflow with 2+ AI nodes fails at the 2nd Pi node.
- **Why it matters:** This makes Pi effectively unusable for real multi-node workflows under the default config, blocking the Pi-as-primary-runner direction.
- **What changed:** The extension-bearing `DefaultResourceLoader` is now built + `reload()`ed **once per process** per `(cwd, systemPrompt, skillPaths)` and reused across calls, instead of a fresh loader + `reload()` on every call.
- **What did NOT change (scope boundary):** No subprocess isolation for Pi (would break in-process `customTools`), no change to the `enableExtensions` default, no Pi SDK fork, no change to the extensions-off path (still a fresh loader per call), and no fast-fail first-token guard (deferred — see Risks).

### Before

```
4-node Pi workflow (one process)            Pi SDK
─────────────────────────────────           ──────
node a → sendQuery #1 ─────────────────────▶ new DefaultResourceLoader + reload()  ✅ ~3s
                                             (extension factories construct singletons)
node b → sendQuery #2 ─────────────────────▶ new DefaultResourceLoader + reload()  ⛔ HANGS
                                             (collides with call #1's live state)
                       ◀──── 30-min idle-timeout, node fails, DAG fails
```

### After

```
4-node Pi workflow (one process)            Pi SDK
─────────────────────────────────           ──────
node a → sendQuery #1 ─────────────────────▶ [build + reload loader ONCE, cache it]  ✅ ~3s
node b → sendQuery #2 ─────────────────────▶ [*reuse cached loader* — no reload]     ✅
node c → sendQuery #3 ─────────────────────▶ [*reuse cached loader*]                 ✅
node d → sendQuery #4 ─────────────────────▶ [*reuse cached loader*]                 ✅
                       ◀──── all nodes complete; each still gets its own session
```

### Before

```
PiProvider.sendQuery()
  └─▶ createNoopResourceLoader(cwd, …)  ── new DefaultResourceLoader (per call)
        └─▶ await loader.reload()       ── re-runs loadExtensions() every call (per call)
              └─▶ createAgentSession({ resourceLoader })
```

### After

```
PiProvider.sendQuery()
  ├─ extensions ON  ──▶ [+] getOrCreateReloadedExtensionLoader(cwd, {systemPrompt, skillPaths})
  │                          └─ cache miss: createNoopResourceLoader + reload() ONCE, memoize
  │                          └─ cache hit:  *return cached loader* (=== reuse, no reload)
  └─ extensions OFF ──▶ createNoopResourceLoader(cwd, …)   (unchanged: fresh, no reload)
        └─▶ createAgentSession({ resourceLoader })  (skips its own reload when loader supplied)
```

**Connection inventory:**

| `provider.ts:sendQuery` | `resource-loader.ts:getOrCreateReloadedExtensionLoader` | **new** | extensions-on path now goes through the process cache |
| `provider.ts:sendQuery` | `resource-loader.ts:createNoopResourceLoader` | **modified** | now called only on the extensions-off path (+ internally by the cache helper) |
| `resource-loader.ts:getOrCreateReloadedExtensionLoader` | `DefaultResourceLoader.reload()` | unchanged | same reload, just invoked once per input set instead of per call |
| `provider.test.ts` | `resource-loader.ts:resetReloadedExtensionLoaderCache` | **new** | test hook clears the module-level cache between tests |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `core` (providers package) — closest listed scope is AI clients
- Module: `providers:pi`

## Change Metadata

- Change type: `bug`
- Primary scope: `core` (the `@archon/providers` Pi community provider)

## Linked Issue

- Closes #1877

```bash
bun run validate   # → exit code 0 (type-check, lint, format:check, check:bundled*, all package tests)
bun test packages/providers/src/community/pi/provider.test.ts   # → 71 pass / 0 fail (incl. 4 new)
```

- Evidence provided: `bun run validate` exited `0`. New tests in `provider.test.ts` assert: reload/construct run **once** across two identical sequential calls; distinct `cwd` or `systemPrompt` each reload once; extensions-off never caches and never reloads.
- Skipped commands: none.

## Security Impact (required)

- File system access scope changed? **No** (same extensions loaded from the same paths; just once instead of N times)

## Compatibility / Migration

- Config/env changes? **No**

- Verified scenarios: full `bun run validate` green; the 4 new regression tests encode the exact failing scenario from the issue (sequential calls sharing one process) and assert single-reload behavior.
- Edge cases checked: distinct `cwd` → separate loaders; distinct per-node `systemPrompt` → separate loaders (no silent prompt reuse); extensions disabled → no caching, no reload; failed reload is evicted so the next call retries.
- What was not verified: a live end-to-end run of `minimax-seq` against a real Pi/MiniMax backend (requires the issue author's installed extensions + API creds). The fix removes the repeated `reload()` that the issue's `session_started:1` evidence pinpointed as the hang.

- Affected subsystems/workflows: Pi provider only. Claude/Codex/other providers untouched.
- Potential unintended effects: extension factories now run once per process, so a singleton constructed in call #1 is shared by later calls in the same `cwd` group — this is the intended (and more correct) lifetime; `bindExtensions()`/`session_start` still fire per session. A loader is cached per distinct `(cwd, systemPrompt, skillPaths)`; long-lived processes touching many cwds hold one loader each (bounded, negligible).
- Guardrails/monitoring: the existing `pi.session_started` log now fires once per node again (was 1-per-process); the executor idle-timeout remains as the backstop.

## Rollback Plan (required)

- Fast rollback: `git revert 5486d624` (single self-contained commit) — restores the per-call loader + reload.
- Feature toggles: `assistants.pi.enableExtensions: false` independently bypasses the whole extension path (the issue's documented workaround) without reverting.
- Observable failure symptoms: Pi nodes emitting zero output and idle-timing-out; `pi.session_started` logging once per process instead of once per node.

## Risks and Mitigations

- Risk: Multiple nodes in one process with **genuinely different** `systemPrompt`/`skills` AND extensions enabled still produce >1 `reload()` (each distinct input set reloads once), which could re-trigger the non-re-entrancy hang for that uncommon combo.
  - Mitigation: The reported repro and ~all real workflows use a uniform/absent prompt across nodes → a single cached loader → fully fixed. The investigation artifact proposed a Pi-specific fast-fail first-token guard for this residual; it is **deferred** here because a hard provider-level first-token timeout risks false-positives on legitimately slow cold-start/queued backends and overlaps the existing 30-min executor idle-timeout. Tracked as a follow-up if the heterogeneous case is observed in practice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fix extension loader caching: identical queries now reuse a single extension-enabled loader (redundant reloads suppressed); differing configurations produce separate loaders. Concurrent requests share in-flight reload work, and failed reloads are evicted so retries succeed. When extensions are disabled, caching is bypassed (fresh loader per call).

* **Tests**
  * Added tests covering loader reuse, concurrency, failure retry behavior, and cache reset between tests.

### DIFF
--- a/packages/providers/src/community/pi/provider.test.ts
+++ b/packages/providers/src/community/pi/provider.test.ts
@@ -167,6 +167,9 @@ mock.module('@earendil-works/pi-coding-agent', () => ({
 // Import AFTER mocks are set — module resolution freezes the mocks.
 import { PiProvider } from './provider';
 import { PI_CAPABILITIES } from './capabilities';
+// Same module instance the provider dynamic-imports, so clearing this cache
+// resets the loader the provider reuses across calls (issue #1877).
+import { resetReloadedExtensionLoaderCache } from './resource-loader';
 
 // ─── Helpers ────────────────────────────────────────────────────────────
 
@@ -237,6 +240,10 @@ describe('PiProvider', () => {
     runtimeOverrides = {};
     delete process.env.GEMINI_API_KEY;
     delete process.env.ANTHROPIC_API_KEY;
+    // The extension-loader cache is module-level and persists across tests;
+    // clear it so each test starts with an empty cache and sees its own
+    // construct/reload calls (issue #1877).
+    resetReloadedExtensionLoaderCache();
   });
 
   test('getType returns "pi"', () => {
@@ -1755,5 +1762,86 @@ describe('PiProvider', () => {
       expect.objectContaining({ scope: 'global', err: loadError }),
       'pi.settings_load_error'
     );
+  });
+
+  // ─── Extension loader reuse (issue #1877) ─────────────────────────────────
+  //
+  // Pi's reload() re-invokes every installed extension factory; the 2nd reload
+  // in a process deadlocks on the first call's never-torn-down state. The fix
+  // loads the extension-bearing loader ONCE per process per input set and
+  // reuses it — so reload()/construct must run once across identical calls,
+  // while each call still gets its own session.
+  describe('extension loader reuse (issue #1877)', () => {
+    test('reload() and loader construction run once across two sequential calls with identical inputs', async () => {
+      process.env.GEMINI_API_KEY = 'sk-test';
+      resetScript(scriptedAgentEnd());
+
+      await consume(
+        new PiProvider().sendQuery('a', '/tmp', undefined, { model: 'google/gemini-2.5-pro' })
+      );
+      await consume(
+        new PiProvider().sendQuery('b', '/tmp', undefined, { model: 'google/gemini-2.5-pro' })
+      );
+
+      // The bug was reload() (and the extension factory it drives) running per
+      // call; after the fix the cached loader is built + reloaded exactly once.
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(1);
+      expect(mockResourceLoaderReload).toHaveBeenCalledTimes(1);
+      // Each call still gets its own session (correctness preserved).
+      expect(mockCreateAgentSession).toHaveBeenCalledTimes(2);
+      expect(mockDispose).toHaveBeenCalledTimes(2);
+    });
+
+    test('different cwd gets its own reloaded loader', async () => {
+      process.env.GEMINI_API_KEY = 'sk-test';
+      resetScript(scriptedAgentEnd());
+
+      await consume(
+        new PiProvider().sendQuery('a', '/tmp/one', undefined, { model: 'google/gemini-2.5-pro' })
+      );
+      await consume(
+        new PiProvider().sendQuery('b', '/tmp/two', undefined, { model: 'google/gemini-2.5-pro' })
+      );
+
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(2);
+      expect(mockResourceLoaderReload).toHaveBeenCalledTimes(2);
+    });
+
+    test('a distinct per-node systemPrompt gets its own reloaded loader (no silent prompt reuse)', async () => {
+      process.env.GEMINI_API_KEY = 'sk-test';
+      resetScript(scriptedAgentEnd());
+
+      await consume(
+        new PiProvider().sendQuery('a', '/tmp', undefined, {
+          model: 'google/gemini-2.5-pro',
+          systemPrompt: 'prompt A',
+        })
+      );
+      await consume(
+        new PiProvider().sendQuery('b', '/tmp', undefined, {
+          model: 'google/gemini-2.5-pro',
+          systemPrompt: 'prompt B',
+        })
+      );
+
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(2);
+      expect(mockResourceLoaderReload).toHaveBeenCalledTimes(2);
+    });
+
+    test('extensions disabled keeps a fresh loader per call and never reloads', async () => {
+      process.env.GEMINI_API_KEY = 'sk-test';
+      resetScript(scriptedAgentEnd());
+
+      const opts = {
+        model: 'google/gemini-2.5-pro',
+        assistantConfig: { enableExtensions: false },
+      };
+      await consume(new PiProvider().sendQuery('a', '/tmp', undefined, opts));
+      await consume(new PiProvider().sendQuery('b', '/tmp', undefined, opts));
+
+      // No caching on the extensions-off path: fresh loader each call, no reload.
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(2);
+      expect(mockResourceLoaderReload).not.toHaveBeenCalled();
+    });
   });
 });
--- a/packages/providers/src/community/pi/provider.test.ts
+++ b/packages/providers/src/community/pi/provider.test.ts
@@ -169,7 +169,10 @@ import { PiProvider } from './provider';
 import { PI_CAPABILITIES } from './capabilities';
 // Same module instance the provider dynamic-imports, so clearing this cache
 // resets the loader the provider reuses across calls (issue #1877).
-import { resetReloadedExtensionLoaderCache } from './resource-loader';
+import {
+  getOrCreateReloadedExtensionLoader,
+  resetReloadedExtensionLoaderCache,
+} from './resource-loader';
 
 // ─── Helpers ────────────────────────────────────────────────────────────
 
@@ -1842,6 +1845,59 @@ describe('PiProvider', () => {
       // No caching on the extensions-off path: fresh loader each call, no reload.
       expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(2);
       expect(mockResourceLoaderReload).not.toHaveBeenCalled();
+    });
+
+    test('distinct additionalSkillPaths get their own reloaded loader', async () => {
+      const a = await getOrCreateReloadedExtensionLoader('/tmp', {
+        additionalSkillPaths: ['/skills/x'],
+      });
+      const b = await getOrCreateReloadedExtensionLoader('/tmp', {
+        additionalSkillPaths: ['/skills/y'],
+      });
+      expect(a).not.toBe(b);
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(2);
+      expect(mockResourceLoaderReload).toHaveBeenCalledTimes(2);
+    });
+
+    test('concurrent callers (same key) share a single in-flight reload (the documented invariant)', async () => {
+      // Gate reload() so both callers are in-flight before either resolves — the
+      // exact race two parallel same-layer DAG nodes hit. Caching the resolved
+      // value instead of the Promise would construct/reload twice and fail this.
+      let releaseReload: (() => void) | undefined;
+      mockResourceLoaderReload.mockImplementationOnce(
+        () =>
+          new Promise<undefined>(resolve => {
+            releaseReload = (): void => resolve(undefined);
+          })
+      );
+
+      const p1 = getOrCreateReloadedExtensionLoader('/tmp', {});
+      const p2 = getOrCreateReloadedExtensionLoader('/tmp', {});
+      // Both subscribed before reload resolves.
+      releaseReload?.();
+      const [l1, l2] = await Promise.all([p1, p2]);
+
+      expect(l1).toBe(l2);
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(1);
+      expect(mockResourceLoaderReload).toHaveBeenCalledTimes(1);
+    });
+
+    test('a failed reload is evicted so the next call retries cleanly', async () => {
+      mockResourceLoaderReload.mockImplementationOnce(async () => {
+        throw new Error('broken extension');
+      });
+
+      await expect(getOrCreateReloadedExtensionLoader('/tmp', {})).rejects.toThrow(
+        /Pi extension load failed: broken extension/
+      );
+
+      // Entry was evicted on failure → the retry constructs + reloads again
+      // (rather than returning the poisoned rejected promise forever).
+      mockResourceLoaderReload.mockImplementationOnce(async () => undefined);
+      const loader = await getOrCreateReloadedExtensionLoader('/tmp', {});
+      expect(loader).toBeDefined();
+      expect(MockDefaultResourceLoader).toHaveBeenCalledTimes(2);
+      expect(mockResourceLoaderReload).toHaveBeenCalledTimes(2);
     });
   });
 });
--- a/packages/providers/src/community/pi/provider.ts
+++ b/packages/providers/src/community/pi/provider.ts
@@ -448,19 +448,14 @@ export class PiProvider implements IAgentProvider {
     // deadlocks on the first call's never-torn-down state (issue #1877 — see the
     // doc on getOrCreateReloadedExtensionLoader). When extensions are OFF there
     // is no reload() and thus no re-entrancy hazard, so a fresh per-call loader
-    // is fine.
-    let resourceLoader: DefaultResourceLoader;
-    if (enableExtensions) {
-      resourceLoader = await getOrCreateReloadedExtensionLoader(cwd, {
-        ...(systemPrompt !== undefined ? { systemPrompt } : {}),
-        ...(skillPaths.length > 0 ? { additionalSkillPaths: skillPaths } : {}),
-      });
-    } else {
-      resourceLoader = createNoopResourceLoader(cwd, {
-        ...(systemPrompt !== undefined ? { systemPrompt } : {}),
-        ...(skillPaths.length > 0 ? { additionalSkillPaths: skillPaths } : {}),
-      });
-    }
+    // is fine. Build the shared options once so the two paths can't drift.
+    const loaderOptions = {
+      ...(systemPrompt !== undefined ? { systemPrompt } : {}),
+      ...(skillPaths.length > 0 ? { additionalSkillPaths: skillPaths } : {}),
+    };
+    const resourceLoader: DefaultResourceLoader = enableExtensions
+      ? await getOrCreateReloadedExtensionLoader(cwd, loaderOptions)
+      : createNoopResourceLoader(cwd, loaderOptions);
 
     getLog().info(
       {
--- a/packages/providers/src/community/pi/provider.ts
+++ b/packages/providers/src/community/pi/provider.ts
@@ -3,6 +3,10 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
 import { createLogger } from '@archon/paths';
+// Type-only import — erased by TS, so it does NOT trigger Pi's config.js
+// package.json read at module load (see the header note below). Used only to
+// annotate the per-call ResourceLoader local.
+import type { DefaultResourceLoader } from '@earendil-works/pi-coding-agent';
 
 import type {
   IAgentProvider,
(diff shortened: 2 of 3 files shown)


Wirasm:
## 🔍 Automated Code Review

### Summary

Core fix is correct and well-reasoned: root cause (repeated `reload()` re-invoking extension factories) is accurately addressed, the store-the-Promise concurrency choice is sound, the cache key (`cwd` + `systemPrompt` + sorted `skillPaths`) is complete, the `import type` of `DefaultResourceLoader` is type-only/erased, and the 4 new tests prove single-reload behavior against real mock call counts. Two non-blocking operational findings raised; disposition below.

### Findings

#### ✅ Strengths
- Removes the repeated `reload()` the issue's `session_started:1` evidence pinpointed as the hang site.
- Promise-valued cache makes concurrent same-layer nodes await one shared reload instead of racing into two.
- Extensions-OFF path keeps a fresh per-call loader (no cache, no reload) — no behavior change where there's no hazard.
- Failure eviction (`pending.catch(() => cache.delete(key))`) avoids poisoning the cache with a transient reload error; the rejection still propagates to the awaiting caller.

#### ⚠️ Findings & disposition

- **`resource-loader.ts` — unbounded cache growth** (reviewer confidence 83). **Addressed via doc + deliberate non-change.** The proposed size-cap/LRU eviction is **unsafe here**: evicting the oldest entry can drop a loader a long-running workflow is still using, making its next node cache-miss and `reload()` a second copy of the same extensions while the first is live — re-introducing the exact deadlock this PR fixes. Growth is bounded by distinct `(cwd, systemPrompt, skillPaths)` tuples (≈ worktrees touched, not request volume), and each loader is a light config object. Documented the bound + rationale in commit `567b8fb`; safe time/idle reclamation belongs with isolation cleanup (a future cross-package hook), tracked as follow-up.

- **`resource-loader.ts:resetReloadedExtensionLoaderCache` — test-only export without a guard** (reviewer confidence 80). **Kept as-is for codebase consistency.** Plain unguarded reset exports are the established convention here — `resetCodexSingleton` (same package), `resetIsolationProvider`, `resetTelemetryForTests`, `resetDatabase`. Adding a `NODE_ENV` guard or a separate test-utils file would deviate from that pattern for no net safety gain.

#### 🔒 Security
- No security concerns. No new permissions, network calls, secret handling, or filesystem scope — the same extensions load from the same paths, just once per process instead of per call.

### Checklist

- [x] Fix addresses root cause from the investigation
- [x] Code follows codebase patterns (mirrors `resetCodexSingleton`/Codex singleton-reuse shape)
- [x] Tests cover the change (single-reload, distinct cwd, distinct systemPrompt, extensions-off)
- [x] No obvious bugs introduced; `bun run validate` green

---
*Self-reviewed by Claude • findings dispositioned • ready for human review*

Wirasm:

Seven specialized agents reviewed this PR (code quality, docs impact, test coverage, comment accuracy, silent failures, type design, simplification). The core fix is **sound and correct** — no agent found a correctness bug that blocks merge. Findings below are about test coverage of documented invariants, comment rot, and minor polish.

### Critical Issues (0 found)

None. The cache logic, `.catch()` eviction (confirmed *not* an error-swallow), error propagation to the user, type safety, and the extensions-on/off split are all correct.

### Important Issues (3 found)

| # | Agent(s) | Issue | Location |
|---|----------|-------|----------|
| I1 | code-reviewer + pr-test-analyzer | The **primary documented invariant** — "concurrent same-layer nodes await a single shared `reload()`" — is **untested**. The 4 new tests only cover the *sequential* case. Two parallel DAG nodes (extensions on, same key) is the exact runtime scenario this fix targets, and a regression that broke Promise-sharing (e.g. caching the resolved value instead) would pass every current test. Both agents independently flagged this (criticality 8). | `resource-loader.ts:160-183` |
| I2 | comment-analyzer | Doc comment makes an **unverifiable external assertion** as fact: "agent-rooms builds a RoomManager + SdkSessionFactory…". That package isn't installed anywhere in the repo; the names can't be corroborated and will mislead if they drift. The general mechanism (factories build process-scoped singletons) is the load-bearing point and stands without the specific example. | `resource-loader.ts:107-108` |
| I3 | comment-analyzer | `setFlagValue` inline comment is **factually inaccurate**: it says "setFlagValue silently no-ops", but `setFlagValue` itself unconditionally sets the flag — the no-op actually comes from the `if (runner)` guard at the call site (`provider.ts:533`). Reword to attribute the no-op to the guard. | `resource-loader.ts:173-174` |

### Suggestions (6 found)

| # | Agent | Suggestion | Location |
|---|-------|------------|----------|
| S1 | comment-analyzer | Drop the `(Pi SDK \`sdk.js\`)` internal-filename reference — the behavioral claim is verified/accurate, but naming the build artifact will rot on the next SDK restructure. | `resource-loader.ts:117` |
| S2 | pr-test-analyzer | Add a **failure-eviction test** (reload rejects → entry evicted → next call retries cleanly). It's the only safety valve against a permanently-poisoned cache; currently untested (criticality 7). | `resource-loader.ts:179-180` |
| S3 | silent-failure-hunter | On `reload()` failure the raw Pi SDK message surfaces to the operator (often cryptic). Consider a context-preserving rethrow pointing at `~/.pi/agent/extensions/` or `enableExtensions: false`. Usability gap, not a silent failure. | `resource-loader.ts:175` |
| S4 | code-simplifier | Build the shared `loaderOptions` object once, then dispatch — removes the duplicated `systemPrompt`/`additionalSkillPaths` spread across both if/else branches (guards against a future one-sided edit). | `provider.ts:453-463` |
| S5 | type-design-analyzer | Change `extensionLoaderCacheKey` to take `Pick<NoopResourceLoaderOptions, 'systemPrompt' \| 'additionalSkillPaths'>` so the key function is structurally tied to the same option type the public API uses — a future added field becomes harder to miss in the key. | `resource-loader.ts:142-148` |
| S6 | pr-test-analyzer | Add a `additionalSkillPaths`-keying test (distinct skill sets → distinct loaders). Lower priority — the sorted-`JSON.stringify` key is trivially correct on inspection (criticality 5). | `resource-loader.ts:142-148` |

### Strengths

- **Every load-bearing Pi SDK internals claim was independently verified accurate against `@earendil-works/pi-coding-agent` 0.76.0** — `reload()` re-runs `packageManager.resolve()` + `loadExtensions()`, jiti `moduleCache: false`, `dispose()` not emitting `session_shutdown` (while `reload()` does), `createAgentSession({resourceLoader})` skipping its own reload. That's a precise, well-researched fix.
- **Strong type design (8/10):** `Pick<…>` excludes `enableExtensions` from the cache helper's public surface, making "call `reload()` on a non-cached loader" structurally unreachable rather than convention. `let resourceLoader: DefaultResourceLoader` + exhaustive if/else is cleaner and safer than the prior single-path construction.
- **No silent failures:** the `.catch()` is a pure side-effecting eviction hook — the original `pending` is returned and `await`ed, so the rejection propagates to the dag-executor's node-level handler, is logged as `dag_node_failed`, and reaches the user. Verdict: PASS from the error-handling agent.
- **Correct concurrent design:** caching the in-flight Promise (not the resolved loader) so racing nodes share one `reload()`.
- **Excellent "why" comments:** the growth/eviction rationale (why LRU is *unsafe* here) is exactly the kind of decision-capturing comment a future maintainer needs.

### Documentation

No updates needed. `CLAUDE.md` (Pi provider entry), `ai-assistants.md` (extensions section), and `CHANGELOG.md` are all accurate — this is an internal fix with no user-visible interface change, and the project's `/release` skill generates changelog entries from commits.

### Verdict

**NEEDS MINOR FIXES** — mergeable in substance, no blocking correctness issues. Recommend addressing **I1** (concurrent-callers test — the most-documented invariant is the one untested, flagged by two agents) and the two quick comment fixes **I2/I3/S1** before merge. The remaining suggestions (S2–S6) are good follow-ups but non-blocking.

### Recommended Actions

1. Add a concurrent-callers test against `getOrCreateReloadedExtensionLoader` directly (gated mock `reload()` + `Promise.all`) — verifies one construction / one reload / `l1 === l2` (**I1**).
2. Trim the unverifiable `agent-rooms` specifics and the `sdk.js` filename from the docblock (**I2**, **S1**).
3. Fix the `setFlagValue` comment to attribute the no-op to the `if (runner)` guard (**I3**).
4. Optional polish: failure-eviction test (**S2**), actionable reload error (**S3**), dedupe option spread (**S4**), tie key fn to `Pick<>` (**S5**).

*Reviewed by: code-reviewer · docs-impact · pr-test-analyzer · comment-analyzer · silent-failure-hunter · type-design-analyzer · code-simplifier*

Wirasm:
## Addressed — ultrareview findings (commit 85dc130)

Thanks — all three Important findings fixed, plus the four valuable suggestions. One deferred with rationale.

| # | Finding | Disposition |
|---|---------|-------------|
| **I1** | Concurrent-callers invariant untested | **Fixed.** New test calls `getOrCreateReloadedExtensionLoader` directly with a **gated** `reload()` so both callers are in-flight before either resolves, then asserts one construct / one reload / `l1 === l2`. Caching the resolved value instead of the Promise now fails this test. |
| **I2 / S1** | Unverifiable `agent-rooms` specifics + `sdk.js` filename in docblock | **Fixed.** Trimmed to the load-bearing, verifiable mechanism ("factories can construct process-scoped singletons — ports, file handles, nested SDK clients"); dropped the build-artifact filename. |
| **I3** | `setFlagValue silently no-ops` comment inaccurate | **Fixed.** Reworded to attribute the no-op to the provider's `if (runner)` guard (skipped when the runner is undefined), not to `setFlagValue` itself. |
| **S2** | Failure-eviction untested | **Fixed.** New test: a rejected `reload()` evicts the entry, and the next call constructs + reloads again (asserts 2 constructs / 2 reloads) rather than returning a poisoned promise. |
| **S3** | Raw Pi SDK reload error is cryptic | **Fixed.** `reload()` failures now rethrow with an actionable pointer (`~/.pi/agent/extensions/` / `enableExtensions: false`), preserving the original as `cause`. Not a swallow — still propagates to the node handler. |
| **S4** | Duplicated option spread across if/else | **Fixed.** Build `loaderOptions` once, then dispatch via a ternary. |
| **S6** | `additionalSkillPaths` keying untested | **Fixed.** New test: distinct skill sets → distinct loaders (2 constructs / 2 reloads). |
| **S5** | Tie key fn to `Pick<>` | **Deferred.** The public `getOrCreateReloadedExtensionLoader` already takes `Pick<NoopResourceLoaderOptions, 'systemPrompt' \| 'additionalSkillPaths'>`, which documents the keyed surface; the 3-line internal key fn stays explicit/positional. Low marginal value, no behavior change. |

Tests: **74 pass / 0 fail** (was 71). `bun run validate` green.

The one substantive add is **I1** — you're right that the concurrent path is the actual runtime scenario (parallel same-layer DAG nodes) and the sequential tests wouldn't catch a Promise-sharing regression.

Also noted out-of-band: this fix is a hard dependency for the cross-provider structured-output reask loop (it fires multiple Pi `sendQuery()` calls per node) — added that to that plan's Dependencies + Task 10.

PR #1894, opened 2026-06-05
Fix: warn on the bash node $node.output double-quote footgun + docs (#1884)

### LINKED ISSUE #1884: Bash node $node.output.field substitution is single-quoted — wrapping it in double quotes is a silent footgun

Wirasm:
## Summary

- **Problem:** In `bash:` nodes, `$nodeId.output[.field]` is injected **pre-quoted** by Archon (single-quoted inline for small values; `$(cat ...)` for outputs >32 KB). Wrapping it in double quotes — `var="$n.output.field"` — bakes the literal single quotes into the value (`'value'`), so downstream comparisons silently fail. Numeric/boolean fields are injected raw, so it "works" for them, making the bug intermittent and hard to spot (#1884).
- **Why it matters:** Surfaced while writing structured-output e2e smokes — the first smoke "failed" purely from this quoting, not the feature under test. It's a real authoring footgun.
- **What changed:** A workflow-validator **warning** when a `bash:` body contains a double-quoted `$nodeId.output` substitution, plus a docs caution callout (authoring-workflows.md), a shell-quoting footgun section (variables.md), and a cross-reference from loop-nodes.md.
- **What did NOT change (scope boundary):** No behavior change to substitution itself (single-quoting is correct for safety); the warning is advisory (`level: 'warning'`), never an error.

## Provenance / why a fresh PR

This is the genuine #1884 fix **extracted** from the autonomous run on #1888, which had bundled three unrelated things into one "bash footgun" PR:

| #1888 contained | Kept here? |
|---|---|
| Validator double-quote warning + docs | ✅ yes (this PR) |
| **Model tier → concrete reverts** (`small`→`haiku`, `medium`→`sonnet`, `large`→`opus[1m]`) across all 14 bundled workflows + regenerated bundle | ❌ dropped — would **undo #1873's** model-alias adoption (it's already on `dev`) and hardcode Claude models into the provider-agnostic defaults |
| `validateModelRef` placement refactor + `title-generator`/`orchestrator-agent` spread→if refactors | ❌ dropped — off-scope |

So #1888 is being closed in favor of this scoped change.

### After
```
validateWorkflowResources (packages/workflows/src/validator.ts)
  for each node:
    …existing checks…
    [+] if isBashNode(node) && /"[^"\n]*\$\w[\w-]*\.output/.test(node.bash):
          push { level: 'warning', field: 'bash', message: '"$nodeId.output" double-quoting …', hint }
```

**Connection inventory:**

| `validator.ts` | `isBashNode` (`./schemas`) | **new** | added to the existing import |
| `validateWorkflowResources` | bash-double-quote regex check | **new** | advisory warning only |

## Label Snapshot
- Risk: `risk: low` (advisory warning; no runtime behavior change)
- Size: `size: S`
- Scope: `workflows` + `docs`
- Module: `workflows:validator`

## Change Metadata
- Change type: `bug` (DX footgun)
- Primary scope: `workflows`

## Linked Issue
- Closes #1884
- Supersedes #1888 (which is being closed — it tangled this fix with model-alias reverts)

```bash
bun run validate   # → exit 0 (type-check, lint, format, check:bundled*, all tests)
bun test packages/workflows/src/validator.test.ts   # → 53 pass / 0 fail
```
- New validator tests: warns on `var="$emit.output.status"`; warns on `echo "result: $emit.output.status"`; no warning on the unquoted idiom `var=$node.output.field`; no warning on single-quoted; `script:` nodes exempt.

## Security Impact (required)
- New permissions/capabilities? **No** · New external network calls? **No** · Secrets handling changed? **No** · FS scope changed? **No**

## Compatibility / Migration
- Backward compatible? **Yes** — advisory warning only; no config/env/DB changes. Existing bundled defaults that use the pattern will now emit this warning (informational); fixing their quoting is a separate follow-up.

- **Verified:** `bun run validate` green; the 5 new validator tests cover the warn/no-warn/script-exempt cases; confirmed no test or CI check asserts zero validation **warnings** on bundled defaults (so the new advisory warning doesn't break anything).
- **Not verified:** I did not also rewrite the double-quoting inside the bundled default workflows — they'll surface this warning, but fixing each (distinguishing real footguns from intentional interpolation) is out of scope here.

- Affected: `bun run cli validate workflows` now emits a warning for bash nodes with double-quoted output refs. No execution path changes.
- Monitoring: the warning text + hint point authors at the unquoted idiom.

## Rollback Plan (required)
- `git revert <merge>` — single self-contained commit; removes the validator warning + docs.
- No flags/migrations.

## Risks and Mitigations
- **Risk:** the regex `"[^"\n]*\$\w[\w-]*\.output` could flag a legitimately-double-quoted interpolation.
  - **Mitigation:** in `bash:` nodes pre-quoting makes double-quoting wrong in both the assignment and the `echo "...$x.output..."` cases (you'd see the literal `'value'`), so flagging both is correct; it's a warning, not an error; `script:` nodes (raw, unquoted substitution) are exempt.

## Summary by CodeRabbit

* **New Features**
  * Added a validation warning that detects and warns when outputs injected into bash contexts are double-quoted (applies to bash nodes and loop until-conditions).

* **Documentation**
  * Added a caution note and expanded guidance on shell quoting, showing correct vs incorrect usage and clarifying behavior for small vs large outputs.

* **Tests**
  * Added/extended tests covering the new double-quote lint across bash and loop scenarios.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -378,6 +378,22 @@ Variable substitution order:
 1. Standard variables (`$WORKFLOW_ID`, `$USER_MESSAGE`, `$ARTIFACTS_DIR`, etc.)
 2. Node output references (`$nodeId.output`, `$nodeId.output.field`)
 
+:::caution[Double-quoting `$node.output` in `bash:` nodes is a silent footgun]
+In `bash:` nodes, `$nodeId.output` and `$nodeId.output.field` are injected pre-quoted by Archon. For small outputs, values are **single-quoted inline** — the quoting is already provided by the substitution. For outputs exceeding 32 KB, Archon spills to a temp file and substitutes `$(cat '/tmp/path')` instead. Wrapping in double quotes corrupts both forms: for inline values it embeds the single quotes; for `$(cat ...)` it enables word-splitting on the command output.
+
+```bash
+# WRONG — produces status="'ok'" (single quotes become part of the value)
+status="$emit.output.status"
+[ "$status" = "ok" ]   # → always false
+
+# CORRECT — leave unquoted; bash assigns: status=ok
+status=$emit.output.status
+[ "$status" = "ok" ]   # → true
+```
+
+**Rule:** use `var=$node.output.field`, never `var="$node.output.field"`. This applies whether the output is small (single-quoted inline) or large (`$(cat ...)`). Numeric and boolean fields are injected raw (without quotes), so double-quoting accidentally "works" for them — making the bug intermittent and hard to spot.
+:::
+
 ### `output_format` for Structured JSON
 
 Use `output_format` to enforce JSON output from an AI node. For Claude, the schema is passed via the SDK's `outputFormat` option and `structured_output` is used directly. For Codex (v0.116.0+), the schema is passed via `TurnOptions.outputSchema` and the agent's inline JSON response is used. Both ensure clean JSON for `when:` conditions and `$nodeId.output` substitution:
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -379,7 +379,7 @@ Variable substitution order:
 2. Node output references (`$nodeId.output`, `$nodeId.output.field`)
 
 :::caution[Double-quoting `$node.output` in `bash:` nodes is a silent footgun]
-In `bash:` nodes, `$nodeId.output` and `$nodeId.output.field` are injected pre-quoted by Archon. For small outputs, values are **single-quoted inline** — the quoting is already provided by the substitution. For outputs exceeding 32 KB, Archon spills to a temp file and substitutes `$(cat '/tmp/path')` instead. Wrapping in double quotes corrupts both forms: for inline values it embeds the single quotes; for `$(cat ...)` it enables word-splitting on the command output.
+In `bash:` nodes, `$nodeId.output` and `$nodeId.output.field` are injected pre-quoted by Archon. For small outputs, values are **single-quoted inline** — the quoting is already provided by the substitution. For outputs exceeding 32 KB, Archon spills to a temp file and substitutes `$(cat '/tmp/path')` instead. Wrapping the substitution in double quotes breaks the **small (inline) case**: `var="$n.output"` becomes `var="'value'"`, embedding the literal single-quotes as part of the value. (For the large `$(cat ...)` case, double-quoting is harmless — `var="$(cat ...)"` is correct bash — but you can't know the output's size at author time, so the rule is unconditional: never double-quote.)
 
 ```bash
 # WRONG — produces status="'ok'" (single quotes become part of the value)
--- a/packages/docs-web/src/content/docs/guides/loop-nodes.md
+++ b/packages/docs-web/src/content/docs/guides/loop-nodes.md
@@ -150,7 +150,10 @@ loop:
 This is useful for deterministic completion criteria: test suites, lint checks,
 build success. The bash script supports the same variable substitution as
 `prompt` (`$ARTIFACTS_DIR`, `$nodeId.output`, etc.). Note: `$nodeId.output`
-values are shell-escaped when substituted into `until_bash`.
+values are shell-escaped when substituted into `until_bash`. The same
+double-quoting footgun that applies to `bash:` nodes applies here — see
+[Shell Quoting in `bash:` vs `script:`](/reference/variables#shell-quoting-in-bash-vs-script)
+for the unquoted idiom to use.
 
 ## Patterns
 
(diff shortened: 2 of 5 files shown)


Wirasm:
## Review pass — code-reviewer + silent-failure-hunter (commit e9d9ae2)

Ran both agents on the diff. They found 2 real bugs in the new lint plus a backwards shell claim — all fixed. `bun run validate` green; validator tests 55 pass.

### Fixed

| Finding | Agent | Fix |
|---|---|---|
| **Regex false positive** — `/"[^"\n]*\$…\.output/` matched the *closing* quote of an unrelated prior string and slid across `;` to a correctly-unquoted ref. `echo "hi"; x=$a.output` warned spuriously. | code-reviewer (95) | Require the opening `"` to be an operand: `/(?:^|[=\s])"…/m`. + regression test. Verified against `echo "hi"; x=$a.output` (no warn) and `msg="done"; status=$node.output.code` (no warn) while keeping all true cases. |
| **Backwards shell claim** — the comment + both docs said double-quoting `$(cat ...)` "enables word-splitting." That's wrong: `var="$(cat ...)"` is *correct* bash. The footgun is only the **small/inline** case (`var="'value'"`). | silent-failure-hunter (HIGH) | Rewrote the rationale (comment + `authoring-workflows.md` + `variables.md`): the rule is unconditional because you can't predict output size at author time, and the inline case is the one that breaks. |
| **Under-detection** — loop `until_bash` substitutes via the same `escapedForBash=true` path (`dag-executor.ts:2361`), so the footgun applies there too, but the lint only checked `isBashNode`. | silent-failure-hunter (HIGH) | Extended the lint to `loop.until_bash` (field `'loop.until_bash'`) + a test. |
| **Hint accuracy** — "the injected single-quotes are the quoting" could be read as "$var never needs quoting"; and the warning fires on numeric/boolean fields where double-quoting is harmless (false positive). | silent-failure-hunter (MED ×2) | Reworded to "injected already quoted" and added a parenthetical noting the numeric/boolean exception so authors don't dismiss the whole warning class. |

### Verified accurate (no change needed)
- The behavior claims (single-quoted inline; `$(cat '/path')` at the `NODE_OUTPUT_FILE_THRESHOLD = 32_768` `>` boundary; numbers/booleans via `String(value)` raw; `script:` nodes exempt at `escapedForBash=false`) all check out against `dag-executor.ts`.
- `level: 'warning'` (not error) is correct and IS surfaced — `validate.ts` sets the `WARNINGS` label, prints the hint, and counts it in the summary; warnings don't affect exit code.
- The `[^"\n]` newline exclusion (a quote spanning lines) is a deliberate, documented limitation.

PR #1890, opened 2026-06-05
feat(web/console): settings core — assistant config + system panel

Wirasm:
## Summary

- **Problem:** The run-centric console (`/console`) has **no settings surface** — to change which assistant/model runs, or to see version/health/platform status, you must leave it for the old UI.
- **Why it matters:** Parity floor before making `/console` the default UI. Theme of the console-replaces-old-UI sequence (PR #4, after #1878/#1881/#1885).
- **What changed:** A global `/console/settings` route with an **AssistantConfigPanel** (default-assistant picker + per-provider free-text model + Codex reasoning/web-search, saving via `PATCH /api/config/assistants`) and a **SystemPanel** (health grid + update-check), backed by a new console `skills/settings.ts` + `skills/providers.ts` layer and a ⚙ header entry + `,` keybinding.
- **What did NOT change (scope boundary):** **No server change** — every API consumed already exists. GitHub device-flow panel split to **PR #5**. Env-var editing stays on the project rail (it's project-scoped). Non-safe-filtered fields (`claudeBinaryPath`, `settingSources`, …) are out of scope (not round-trippable). Web-only, inside the `experiments/console` isolation boundary.

### Before
```
/console → runs / chat.  To change the model or see health → leave for /settings (old UI).
console header:  [Archon console]                         ← Old UI
```
### After
```
console header:  [Archon console]              ⚙ Settings   ← Old UI
/console/settings (global)
┌── Assistant ───────────────────────────────────────────┐
│ Default assistant [ claude ▾ ]                          │
│ Claude   model [ sonnet          ]                      │
│ Codex    model [ gpt-5.3-codex   ] effort [▾] web [▾]   │
│ OpenCode model [                 ]                       │
│ Pi       model [                 ]                       │
│ Copilot  model [                 ]            [Save]     │
├── System ──────────────────────────────────────────────┤
│ status ok · adapter web · db sqlite · v0.4.1            │
│ concurrency 0/10 active · running 0 · platforms [Web]   │
│ Up to date (0.4.1)                                      │
└─────────────────────────────────────────────────────────┘
```

### After
```
getConfig / listProviders / getHealth / getUpdateCheck  (skills → requestJson)
  → useEntity(K.config|providers|health|updateCheck)
      → AssistantConfigPanel  → buildAssistantUpdate(form) → updateAssistantConfig
          → PATCH /api/config/assistants → ~/.archon/config.yaml → invalidate(K.config)
      → SystemPanel (read-only)
[...]

PR #1885, opened 2026-06-05
feat(web/console): chat workflow-completion card + SSE-lock + jump-to-bottom

Wirasm:
## Summary

- **Problem:** Three gaps kept the console's project-scoped chat from replacing the old `/chat`: (1) a chat-launched workflow's completion was **invisible** — its `workflow_result` message (summary + `{workflowName, runId}`) was swallowed because `ChatStream` filters the whole `workflow_` category prefix and `toMessage` never parsed `workflowResult`; (2) the composer stayed **disabled ~6s** (`SETTLE_MS`) after each reply because the SSE lock event wasn't wired; (3) no jump-to-bottom.
- **Why it matters:** Console-vs-`/chat` parity — Theme A of the console-replaces-old-UI sequence.
- **What changed:** Parse `workflowResult`; render a new `ConsoleWorkflowResultCard`; wire `onLockChange` → `busy`; add a jump-to-bottom button.
- **What did NOT change (scope boundary):** **No server change** — the `workflow_result` message, its payload, and the lock events all already exist server-side. No un-suppressing the `workflow_` prefix (other `workflow_*` narration stays hidden). No `MessageItem` change. No streaming-cursor.

### Before
```
chat: launch a workflow → it completes → …nothing renders… (workflow_result swallowed)
composer: disabled ~6s after every reply (SETTLE_MS), even though the server already released the lock
```
### After
```
┌─────────────────────────────────────────────────────────┐
│ ✓ Workflow complete: e2e-deterministic  7/8 nodes  18s  │  ◄── ConsoleWorkflowResultCard
│   summarised result…                          Open run →│
└─────────────────────────────────────────────────────────┘
composer: frees the instant conversation_lock:false arrives (settle timer = fallback)
↓ Jump to bottom — appears when scrolled up
```

## Architecture Diagram

```
orchestrator workflow_result (already emitted+persisted+SSE'd)
  → toMessage() NOW parses workflowResult → Message.workflowResult
      → ChatStream: lets workflow_result through the filter → ConsoleWorkflowResultCard
          → skill.getRun(runId) (same call as RunDetailPage) → status/counts/duration/cost
useConversationSSE(activeConvId, onLockChange)  ← NEW 2nd arg (stable)
  → conversation_lock:false → clear settle timer + setBusy(false)
```

**Connection inventory:**

| From | To | Status | Notes |
|------|----|--------|-------|
| `toMessage` | `Message.workflowResult` | **new** | parse of an already-persisted field |
[...]

PR #1881, opened 2026-06-05
feat(web/console): show run provenance — input message + platform icons

Wirasm:
## Summary

- **Problem:** Console run cards/rows/detail showed status, workflow, cost, elapsed — but never *what input started a run*. The prompt was already parsed into the `Run` primitive (`toRun`) yet rendered nowhere; the origin badge was text-only. Two CLI runs of the same workflow looked identical.
- **Why it matters:** Directly addresses a stated console-vs-old-dashboard parity gap ("what did I enter into the dialog… cli/chat") — part of making `/console` the default UI.
- **What changed:** Render `user_message` on `ActiveRunCard`, `RecentRunRow`, `RunStartedLine`, `RunDetailHeader`; add per-platform Lucide icons to `OriginBadge`; add the first test for the `run.ts` primitive.
- **What did NOT change (scope boundary):** No server/DB/schema/`api.generated.d.ts` change. No new renderers beyond text rows. Old UI untouched. The "from chat →" link is **deferred** (see Deviation).

### Before
```
RunDetailHeader:  project / a1b2c3d4   ● running   maintainer-standup  [CLI]   $1.07  11m  ⧉
RecentRunRow:     ▸ completed  maintainer-standup  proj  a1b2c3d4  11m  $1.07  [CLI]
(no prompt shown anywhere; origin is text-only)
```

### After
```
RunDetailHeader:  project / a1b2c3d4   ● running   maintainer-standup  [⌨ CLI]  $1.07  11m  ⧉
                  input  summarise this week's merged PRs and open issues          ◄── NEW sub-row
ActiveRunCard:    input  summarise this week's merged PRs…  (full text on hover)    ◄── NEW grid row
RecentRunRow:     ▸ completed  maintainer-standup  ⟨summarise this week's…⟩  …  [⌨ CLI]  (h-9 kept; title tooltip)
RunStartedLine:   ▶ Workflow maintainer-standup started  10:31
                  ▏ summarise this week's merged PRs and open issues                ◄── NEW blockquote
```

## Architecture Diagram

```
toRun() (run.ts) — already parses user_message + origin — now exported normalizeOrigin for tests
  → Run.userMessage / Run.origin
      → OriginBadge        (+ Lucide PLATFORM_ICONS)
      → ActiveRunCard      (+ `input` grid row)
      → RecentRunRow       (+ inline truncated message, h-9 preserved)
      → RunStartedLine     (+ blockquote line)
      → RunDetailHeader    (+ full-width `input` sub-row)
```

[...]

PR #1878, opened 2026-06-05
fix(web/console): gate /console behind auth + correct run-event normalizer

Wirasm:
## Summary

- **Problem:** Two issues that block making `/console` the default UI (replacing old `/chat` + dashboard): (1) `/console/*` was reachable **without login** when web auth is enabled, and (2) the run-event normalizer read keys the server never writes, so node durations were always null, approvals never rendered, and resume-skipped nodes vanished.
- **Why it matters:** You can't retire the old UI while the replacement bypasses auth or silently mis-renders run events.
- **What changed:** Wrapped `/console/*` in `SessionGate`; fixed the `event.ts` normalizer (`duration`→`duration_ms`, `approval_requested`/`approval_received` with `decision`/`comment`/`reason`, added `node_skipped_prior_success`, carry node-completion enrichment); added the first console test + wired `src/experiments/console/` into the web test script.
- **What did NOT change (scope boundary):** No new renderers (the per-node accordion and in-stream approval rendering are a follow-up PR); tool-event normalization untouched (already correct); old UI untouched.

### Before
```
web auth ON:  user → /console  → console loads (NO login required)   ✗ auth bypass
run detail:   node_completed → duration shows "—" (always null)      ✗
              resume → already-done nodes simply missing             ✗
              approval gate → raw "approval_received — {…}" fallback  ✗
```

### After
```
web auth ON:  user → /console  → SessionGate → /login if no session  ✓
web auth OFF: user → /console  → passthrough (unchanged)             ✓
run detail:   node_completed → real duration (read from duration_ms) ✓
              resume → skipped-prior-success nodes render as skipped ✓
              approvals classified correctly (requested/received)    ✓
```

### Before / After
```
App.tsx Routes
  /console/* ──▶ ConsoleApp                 (before: ungated)
  /console/* ──▶ [~SessionGate] ──▶ ConsoleApp   (after: gated, still no Layout/TopNav)

experiments/console/primitives/event.ts  (toRunEvent normalizer)
  reads workflow_events.data ── corrected keys ──▶ RunEvent ──▶ RunStream/NodeDivider
```

**Connection inventory:**

| From | To | Status | Notes |
|------|----|--------|-------|
| `App.tsx` `/console/*` | `SessionGate` | **new** | console now inside the auth gate |
| `SessionGate` | `ConsoleApp` | **new** | passthrough when web auth disabled |
| `event.ts` `toRunEvent` | `workflow_events.data` | **modified** | reads the keys the server actually writes |
[...]

PR #1896, opened 2026-06-05
feat(web/console): GitHub identity panel (device flow) + fix cache error re-render

Wirasm:
## Summary

- **Problem:** `/console/settings` (shipped #1890) had Assistant + System panels but no GitHub identity panel — multi-user installs still had to use the old UI to connect a per-user GitHub identity.
- **Why it matters:** Last settings piece before console cutover. PR #5 of the sequence (#1878/#1881/#1885/#1890).
- **What changed:** Added `GithubIdentityPanel` (device-flow connect/poll/disconnect, ported from the old UI into the console's react-query-free skill/cache layer). It renders **nothing** on solo-PAT installs (`GET /api/auth/github` 401s with no web identity). **Plus a root-cause cache fix** (see Deviation) without which the 401-hide — and every other console error branch — does not work.
- **What did NOT change:** No server change (all four endpoints + the 401 gate already exist). No Better Auth login UI. No `api.generated.d.ts` regen (device-flow schemas inlined, the console's accepted fallback).

### After
```
MULTI-USER (GET /api/auth/github → 200):
  /console/settings → [Assistant] [System] [GitHub Identity]
    not connected:  Connect your GitHub account…            [ Connect GitHub ]
        → Visit https://github.com/login/device and enter code: WXYZ-1234 (polling…)
    connected:      Connected as @octocat                   [ Disconnect ]

SOLO-PAT (GET /api/auth/github → 401 = no web identity):
  /console/settings → [Assistant] [System]      ← GitHub panel renders NOTHING
```

## Architecture Diagram

```
getGithubConnection → useEntity(K.githubConnection)
  401 (HttpError.status===401) → panel returns null (hide)
  200 → render: connect runs startGithubDeviceFlow → poll(interval) until
        interpretPollStatus → connected/expired/denied → invalidate(K.githubConnection)
cache.ts useEntity [~] — snapshot is now a per-key version counter (was cache.get(key)),
  so the ERROR transition re-renders (the 401-hide depends on it)
```

**Connection inventory:**

| `skills/github.ts` | `/api/auth/github*` (4 verbs) | **new** | inline-typed device flow |
| `GithubIdentityPanel` | `useEntity(K.githubConnection)` | **new** | 401 → null |
| `SettingsPage` | `GithubIdentityPanel` | **new** | third panel |
| `store/cache.ts` `useEntity` | version-counter snapshot | **modified** | error transitions now re-render |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: M`
- Scope: `web`
- Module: `web:console`

## Change Metadata

- Change type: `feature` (+ a `bug` fix in the cache primitive)
- Primary scope: `web`

## Linked Issue

- Related #1878, #1881, #1885, #1890 (console sequence; this is PR #5, the cutover line)

```bash
bun --filter @archon/web type-check   # exit 0
bun run lint                          # exit 0 (--max-warnings 0)
bun run format:check                  # clean
cd packages/web && bun test src/experiments/console/   # 66 pass / 0 fail (7 new in github.test.ts)
```

- **Evidence:** static green; `interpretPollStatus` has 7 unit tests (connected / pending-retry / expired / denied / error+detail / error-no-detail backoff / empty-detail). **Verified end-to-end in a browser** against an isolated server (`ARCHON_HOME=/tmp`, real `~/.archon` untouched): **solo (401)** → the GitHub panel is absent (headings `Settings | Assistant | System`, no "Loading…"); **multi-user (200 via `X-Archon-User`)** → the panel renders "Connect GitHub". The read endpoints (`/api/config`, `/api/health`, `/api/providers`) return 200 and the other two panels render unaffected.
- **Skipped:** full `bun run validate` — change is isolated to `@archon/web`; ran that package's suite + the browser smoke. CI runs the rest. The full device-flow round-trip (completing auth on github.com) was not driven — the panel's loop logic is the unit-tested `interpretPollStatus` + the ported old-UI structure.

## Security Impact (required)

- New permissions/capabilities? **No** — consumes existing per-user GitHub endpoints; on solo installs it renders nothing.
- New external network calls? **No** (existing `/api/auth/github*`).
- Secrets/tokens handling changed? **No** — tokens stay server-side; the UI only sees `{connected, githubLogin}` + the device codes the user is meant to see.

## Compatibility / Migration

- Backward compatible? **Yes** — additive panel; the cache fix is behavior-preserving on the success path (extra re-renders only on the error transition, which previously rendered nothing).
- Config/env changes? **No**

- **Verified scenarios:** type-check/lint/format/66 tests; browser smoke of both the solo (panel hidden on 401) and multi-user (panel shows "Connect GitHub" on 200) paths against an isolated server.
- **Edge cases checked:** 401 → null (hidden), not a hard error; non-401 status error would surface (not hidden) via the now-working error branch; `interpretPollStatus` covers pending-retry / expired / denied / error-with-detail / transient-error backoff; `cancelledRef` guards setState after unmount; `Math.max(1, interval)` + the `expires_in` deadline cap the loop.
- **What was not verified:** an actual completed device-flow authorization on github.com (needs the GitHub App env + a real browser auth); the connected `@login` + Disconnect render path was exercised only via the 200 status shape, not a live token.

- **Affected subsystems:** the console settings page (new panel) **and the console cache primitive** (`useEntity`, used by every console screen).
- **Potential unintended effects:** the `useEntity` snapshot change is the one to watch — it now re-renders on every `notify()` rather than only on value-identity change. In practice the only *added* re-renders are error transitions (success already installs a new cached object), so there's no hot-path (SSE runs feed) cost; the upside is that error branches across the console (including #1890's panels) now actually render.
- **Guardrails:** console tests in CI; the cache change is small and commented.

## Rollback Plan (required)

- **Fast rollback:** `git revert <merge-commit>` — 7 files, web-only.
- **Feature flags/toggles:** none.
- **Observable failure symptoms:** GitHub panel shows when it shouldn't (or vice-versa); or, if the cache change regressed, console panels stop updating live (watch the runs feed). Neither observed.

## Risks and Mitigations

- Risk: the cache `useEntity` change causes excess re-renders in the SSE-driven runs feed.
  - Mitigation: success-path re-renders are unchanged (cache.set already swaps object identity); only error transitions add a render. Verified the runs/settings screens render normally.
- Risk: 401 treated as a hard error → broken panel on solo installs.
  - Mitigation: explicit `error instanceof HttpError && error.status === 401 → return null`; verified hidden in a browser.
- Risk: setState after unmount during the long poll loop.
  - Mitigation: `cancelledRef` set in a cleanup effect, checked before every poll/setState.

---

### Deviation from the plan (root-cause fix)

The plan assumed `useEntity().error` was reactive — it is **not**. `useEntity`'s `useSyncExternalStore` snapshot was `cache.get(key)`, which stays `undefined` on an error, so React bailed out of re-rendering and the freshly-set `error` (read outside the store) was never surfaced. The 401-hide therefore didn't work — the panel hung on "Loading…" — and **every other console error branch was silently dead too** (including #1890's Assistant/System panels). Fixed at the root: a per-key version counter, bumped on every `notify()`, is now the snapshot, so error transitions re-render. Browser-verified.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub identity panel to Settings for connecting, verifying, and disconnecting a GitHub account with live status and verification code display.

* **Bug Fixes**
  * Improved UI reliability so connection/disconnection status and errors update consistently during polling and on page changes.

* **Tests**
  * Added tests validating device-flow polling/status interpretation.

* **Documentation**
  * Updated Settings docs to include GitHub identity.

### DIFF
--- a/packages/web/src/experiments/console/README.md
+++ b/packages/web/src/experiments/console/README.md
@@ -9,7 +9,7 @@ Mounted at `/console/*`. Not part of the shipped product. Validates the mental m
 ## Routes
 
 - `/console` → Runs view (scope = `all`)
-- `/console/settings` → Settings (assistant config + system health) — global
+- `/console/settings` → Settings (assistant config, system health, GitHub identity) — global
 - `/console/p/:projectId` → Runs view scoped to a project
 - `/console/p/:projectId/chat` → Project-scoped agent chat
 - `/console/p/:projectId/r/:runId` → Run detail
--- a/packages/web/src/experiments/console/components/GithubIdentityPanel.tsx
+++ b/packages/web/src/experiments/console/components/GithubIdentityPanel.tsx
@@ -8,10 +8,10 @@ import { SettingsSection } from './SettingsSection';
 type Phase = 'idle' | 'pending' | 'error';
 
 /**
- * Per-user GitHub identity, via the device flow. Multi-user installs only:
- * `GET /api/auth/github` 401s when there's no web identity (the universal
- * solo-PAT state), and we render NOTHING in that case so solo installs never see
- * an irrelevant panel. Any other error surfaces normally.
+ * Per-user GitHub identity, via the device flow. `GET /api/auth/github` 401s when
+ * there's no web identity (the solo-PAT state, and also a logged-out user on a
+ * web-auth install); we render NOTHING then so there's no irrelevant panel. Any
+ * other error surfaces normally.
  *
  * The connect flow ports the old UI's polling state machine into the console's
  * react-query-free cache: start → poll every `interval`s until connected / expired
@@ -53,7 +53,9 @@ export function GithubIdentityPanel(): ReactElement | null {
         const res = await skill.pollGithubDeviceFlow(start.device_code);
         const step = skill.interpretPollStatus(res, interval);
         if (step.kind === 'connected') {
-          setPhase('idle');
+          // Keep `phase = 'pending'` (button stays "Connecting…") until the
+          // invalidate refetch flips `status.connected` to true — otherwise the
+          // not-connected branch flashes "Connect GitHub" for one render.
           setUserCode(null);
           setVerificationUri(null);
           invalidate(K.githubConnection);
@@ -77,6 +79,7 @@ export function GithubIdentityPanel(): ReactElement | null {
   const disconnect = async (): Promise<void> => {
     setDisconnecting(true);
     setMessage(null);
+    setPhase('idle'); // clear any leftover 'pending' from a prior connect
     try {
       await skill.disconnectGithub();
       if (cancelledRef.current) return;
@@ -89,7 +92,9 @@ export function GithubIdentityPanel(): ReactElement | null {
     }
   };
 
-  // Solo-PAT install (no web identity) → 401 → hide the panel entirely.
+  // 401 = no web identity (no Better Auth session and no X-Archon-User): the solo-PAT
+  // state, and also a logged-out user on a web-auth install. Either way there's no
+  // per-user GitHub identity to manage, so hide the panel rather than error.
   if (error instanceof HttpError && error.status === 401) return null;
   if (error !== undefined) {
     return (
@@ -159,9 +164,10 @@ export function GithubIdentityPanel(): ReactElement | null {
           </div>
         ) : null}
 
-        {phase === 'error' && message !== null ? (
-          <p className="font-mono text-[11px] text-error">{message}</p>
-        ) : null}
+        {/* `message` is only ever set by a failed connect OR disconnect (both clear
+            it on start), so render it whenever present — gating on `phase === 'error'`
+            silently swallowed disconnect failures. */}
+        {message !== null ? <p className="font-mono text-[11px] text-error">{message}</p> : null}
       </div>
     </SettingsSection>
   );
--- a/packages/web/src/experiments/console/components/GithubIdentityPanel.tsx
+++ b/packages/web/src/experiments/console/components/GithubIdentityPanel.tsx
@@ -1,0 +1,168 @@
+import { useEffect, useRef, useState, type ReactElement } from 'react';
+import * as skill from '../skills';
+import { useEntity, invalidate } from '../store/cache';
+import { K } from '../store/keys';
+import { HttpError } from '../lib/http';
+import { SettingsSection } from './SettingsSection';
+
+type Phase = 'idle' | 'pending' | 'error';
+
+/**
+ * Per-user GitHub identity, via the device flow. Multi-user installs only:
+ * `GET /api/auth/github` 401s when there's no web identity (the universal
+ * solo-PAT state), and we render NOTHING in that case so solo installs never see
+ * an irrelevant panel. Any other error surfaces normally.
+ *
+ * The connect flow ports the old UI's polling state machine into the console's
+ * react-query-free cache: start → poll every `interval`s until connected / expired
+ * / denied. A `cancelledRef` (set on unmount) stops the loop and guards every
+ * setState so a long poll can't write after unmount.
+ */
+export function GithubIdentityPanel(): ReactElement | null {
+  const { data: status, error } = useEntity(K.githubConnection, skill.getGithubConnection);
+
+  const [phase, setPhase] = useState<Phase>('idle');
+  const [userCode, setUserCode] = useState<string | null>(null);
+  const [verificationUri, setVerificationUri] = useState<string | null>(null);
+  const [message, setMessage] = useState<string | null>(null);
+  const [disconnecting, setDisconnecting] = useState(false);
+
+  const cancelledRef = useRef(false);
+  useEffect(() => {
+    cancelledRef.current = false;
+    return (): void => {
+      cancelledRef.current = true;
+    };
+  }, []);
+
+  const connect = async (): Promise<void> => {
+    setPhase('pending');
+    setMessage(null);
+    try {
+      const start = await skill.startGithubDeviceFlow();
+      if (cancelledRef.current) return;
+      setUserCode(start.user_code);
+      setVerificationUri(start.verification_uri);
+      const deadline = Date.now() + start.expires_in * 1000;
+      let interval = Math.max(1, start.interval);
+      for (;;) {
+        if (cancelledRef.current) return;
+        if (Date.now() > deadline) throw new Error('Device code expired — try again.');
+        await new Promise(r => setTimeout(r, interval * 1000));
+        if (cancelledRef.current) return;
+        const res = await skill.pollGithubDeviceFlow(start.device_code);
+        const step = skill.interpretPollStatus(res, interval);
+        if (step.kind === 'connected') {
+          setPhase('idle');
+          setUserCode(null);
+          setVerificationUri(null);
+          invalidate(K.githubConnection);
+          return;
+        }
+        if (step.kind === 'retry') {
+          interval = step.nextInterval;
+          continue;
+        }
+        throw new Error(step.message); // 'failed'
+      }
+    } catch (e: unknown) {
+      if (cancelledRef.current) return;
+      setPhase('error');
+      setUserCode(null);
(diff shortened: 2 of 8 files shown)


Wirasm:

Ran 7 specialist agents (code, docs, tests, comments, errors, types, simplify) on this PR (7 files, +325/−8, web-only, `experiments/console`). Branch is up to date with `dev`; CI is still running (UNSTABLE = pending, not failing). I independently verified the load-bearing claims against the actual cache and server code before posting.

**The core cache fix is correct** — I traced every `notify()` call site: each value/error mutation is followed synchronously by `notify`, the primitive version-counter is a stable `useSyncExternalStore` snapshot, and no genuine tearing is possible. The findings below are one real UI bug, one overstated comment, and polish.

### Critical Issues (0)
None.

### Important Issues (2)

| # | Agent | Issue | Location |
|---|-------|-------|----------|
| I1 | simplify / errors | **Disconnect errors are silently swallowed (confirmed bug).** `disconnect()` sets `message` but never `phase='error'`; the JSX gate is `phase === 'error' && message !== null`, so a failed disconnect renders **nothing** — the button just re-enables with no feedback. Violates the repo's "surface errors" principle and the panel's own "any other error surfaces normally" docstring. **Fix:** drop the `phase === 'error'` gate (render on `message !== null` — both `connect`/`disconnect` already `setMessage(null)` at start, so no stale-message risk), or have `disconnect` set `phase='error'`. | `GithubIdentityPanel.tsx` (disconnect + message JSX) |
| I2 | comments | **"No tearing" comment is overstated (verified).** The `cache.ts` `useEntity` comment claims "the cache mutates synchronously before `notify`, so the version-triggered render always sees consistent values (no tearing)." Three holes: (a) on the error path `notify()` fires in `.catch()` while `inflight` is still set (`inflight.delete` is in `.finally()` with no second `notify`), so `loading` is transiently `true` *alongside* `error`; (b) `patch()` calls `notify` but never `errors.delete`, so data+error can coexist; (c) the no-loader `revalidate` branch mutates without `notify`. No current consumer is bitten (all check `error` before `loading`), so this is a **comment scope-down**, not a code change. | `store/cache.ts` (useEntity comment) |

### Suggestions (5)

| # | Agent | Suggestion | Location |
|---|-------|-----------|----------|
| S1 | errors / comments | **401-hide conflates "solo-PAT" with "logged-out multi-user".** Correcting the agents' mechanism: `/api/auth/github` is in `PUBLIC_API_GATE_PREFIXES` (excepted from the API gate); the 401 comes from `requireWebUser` in the handler = "no web identity," which is *also* true for a logged-out user on a web-auth install. In gate-*required* mode the whole console 401s anyway (visible), so the silent partial-hide only occurs in the narrow `web auth enabled + ARCHON_WEB_AUTH_REQUIRED=false + not logged in` config. Optionally cross-reference `K.authStatus` (`enabled`) before hiding, or update the comment to say "no web identity" rather than "universal solo-PAT state." | `GithubIdentityPanel.tsx:~94`, `skills/github.ts` docstring |
| S2 | code | **Brief "Connect GitHub" flash after a successful authorization.** On `connected`, `setPhase('idle')` re-renders before the `invalidate(K.githubConnection)` refetch lands, so `status.connected` is still `false` for one render. Leaving `phase` as `'pending'` (button shows "Connecting…") until the cache update arrives avoids the flash. | `GithubIdentityPanel.tsx` (connected branch) |
| S3 | types / errors | **Inlined device-flow types can silently drift from the server.** `GithubConnectionStatus`/`GithubDeviceStart`/`GithubDevicePoll` mirror `auth.schemas.ts` by hand (documented, `@/lib/api`-blocked). The routes already register `.openapi('Github*')` names — they're just absent from a stale `api.generated.d.ts`. Follow-up: regen + migrate to `components['schemas']['Github*']`. Until then, `interpretPollStatus` is exhaustive only against the *local* union; a new server status arrives as an unvalidated cast and falls off the switch → `undefined`. A post-switch defensive `return {kind:'failed', …}` (or `never` assertion) would make that explicit. | `skills/github.ts` |
| S4 | simplify | Consolidate `userCode`/`verificationUri` (always set/cleared together) into one nullable `devicePrompt` object — makes "both or neither" explicit and removes 4 `setState` calls. | `GithubIdentityPanel.tsx` |
| S5 | tests | The cache fix (highest-risk change, touches every console screen) has **no regression test** for "error transition re-renders." A `_subscribeForTest` export + a ~15-line `store/cache.test.ts` would lock the invariant without React infra. Acceptable to defer given the experiment's pure-function-only convention, but it's the one gap worth considering. | `store/cache.ts` |

### Strengths
- **Cache version-counter fix is sound (verified):** primitive snapshot satisfies `useSyncExternalStore`, every value/error set is followed by `notify`, and `errors.delete(key)` correctly runs on every successful (re)load so stale errors don't persist. The motivating bug (dead error branches across the whole console) is real and well-commented.
- **`interpretPollStatus` is pure and exhaustively unit-tested** (7 tests, incl. the `error`-with-empty-`detail` → retry edge). Right call extracting the poll branch logic from the component.
- **`cancelledRef` unmount guard** applied to the async paths; `disconnect`'s `finally` correctly gates `setDisconnecting(false)`.
- **Device-flow `deadline` is sourced from the server's `expires_in`** (not hardcoded) and `Math.max(1, interval)` floors the poll interval.
- **401-hide is the right instinct** for solo installs — it's only the multi-user-logged-out edge (S1) that's ambiguous.

### Documentation
- **One concrete change:** `packages/web/src/experiments/console/README.md` — the `/console/settings` route description ("assistant config + system health") is now stale; add GitHub identity. (Note: this is the same README #1890 should have updated for the route itself — worth confirming that landed.)
- **No change needed:** CLAUDE.md, `docs-web` (documents the *production* Web UI; the console is still an experiment), `experiments/README.md`. The inlined-types comment is a tracked code follow-up, not a doc gap.

### Verdict: NEEDS FIXES (minor)

No blocking/critical issues, and the cache primitive change — the riskiest part — is correct. Recommended before merge: **I1** (the disconnect-error swallow is a genuine silent failure on a real user action) and **I2** (scope down the "no tearing" comment so it doesn't assert a guarantee the code doesn't hold). S1–S5 are optional; S3's type regen is the natural next follow-up.

<sub>🤖 Multi-agent review (code · docs · tests · comments · errors · types · simplify). Claims verified against cache.ts + server auth handlers. Advisory — no files modified.</sub>

Wirasm:
## Review fixes applied — `44526c6f`

Thanks — and good catch on I1. Addressed both Important + the worthwhile suggestions:

| # | Fix |
|---|-----|
| **I1** | Disconnect-error silent failure — `message` now renders on `message !== null` (was gated behind `phase === 'error'`, which only `connect` set). A failed disconnect now shows its error. |
| **I2** | Scoped down the cache "no tearing" comment — `data`/`error`/`loading` can briefly co-exist (loading still true when an error first lands), so consumers check `error` before `loading`, as the panels do. |
| **S1** | Reworded the 401-hide comment + `github.ts` docstring: 401 = "no web identity" (solo-PAT **and** logged-out on a web-auth install), not "universal solo-PAT". |
| **S2** | Removed the one-render "Connect GitHub" flash after a successful connect — `phase` stays `'pending'` until the status refetch flips `connected`; reset in `disconnect`. |
| **S3** | Defensive `default` in `interpretPollStatus` — an unrecognized server status (inline-type drift) fails terminally instead of crashing the loop on `undefined.kind`. +test (8 total). |
| docs | README route now lists GitHub identity. |

**Deferred, with rationale:**
- **S4** (consolidate `userCode`/`verificationUri`) — optional churn; they're set/cleared together in one place, so the "both or neither" footgun is contained.
- **S5** (cache regression test) — locking "error transition re-renders" needs either React/jsdom (the console deliberately has none) or a production test-seam export into the core primitive. The fix is browser-verified end-to-end (solo→hidden, multi-user→Connect). Acknowledged as the one gap.

On your README note: #1890 **did** land the `/console/settings` route line; this PR extends its description.

Validation: web type-check / lint / format:check clean; 67 console tests pass.

PR #1911, opened 2026-06-08
feat(credentials): per-user AI-provider API key connect surface (PR 2/4)

Wirasm:
## Summary

- **Problem:** #1899 shipped the per-user credential store + inject seam, but nothing could *create* a credential — the inject path was a no-op.
- **Why it matters:** this is the slice that makes per-user keys real — a teammate can connect their own provider key and have their runs/chat bill to it instead of the shared install key.
- **What changed:** an API-key connect surface across core (connect-service), server (REST), CLI (`archon ai`), and the experimental console (Settings panel).
- **What did NOT change (scope boundary):** API-key kind only. OAuth subscription login + Codex `auth.json` delivery are PR-3. The old web UI (`packages/web/src/routes/SettingsPage.tsx`) is intentionally untouched — the console is the FE surface now.

### Before

```
  User                Archon
  ────                ──────
  (no per-user key)   credential store + inject seam exist (#1899)
                      but no way to store a key → inject is a no-op
                      every run/chat uses the shared install env key
```

### After

```
  User                         Archon
  ────                         ──────
  Settings ▸ AI Provider Keys  [PUT /api/auth/providers/:provider] → encrypt + upsert (user_id, provider)
   or `archon ai key set <p>`   (key from masked prompt / piped stdin, never argv)
  starts a run / chat          resolve opts.userId → decrypt → deliver → env (merged LAST)  *(seam from #1899)*
  → run bills to *their* key   another user → their key. No cross-user leak.
```

### After

```
 cli/commands/ai.ts ─┐
 console/ProviderConnectionsPanel ─┤   [+]
 (REST) api.ts /api/auth/providers ┘
        │
        ▼
 core/credentials/connect-service.ts  [+] persistProviderApiKey
        │  (validate provider ∈ KNOWN_PROVIDERS)
        ▼
 core/db/user-provider-key-store.ts  (saveUserProviderKey — from #1899)
        ▼
 remote_agent_user_provider_keys (encrypted)  ──(inject seam #1899)──▶ runs + chat env
```

**Connection inventory:**

| From | To | Status | Notes |
|------|----|--------|-------|
| `server/api.ts` | `core.persistProviderApiKey` | **new** | PUT route → connect-service |
| `cli/commands/ai.ts` | `core` store + connect-service | **new** | `archon ai` |
| `console/ProviderConnectionsPanel` | `/api/auth/providers` | **new** | via console `requestJson` |
| connect-service | `core/db/user-provider-key-store` | **new** | reuses #1899 store |

[...]

Issue #1891, opened 2026-06-05
Phase 2: Per-user AI-provider credentials (BYO key + subscription login) — e2e plan


Wirasm:
# Feature: Per-User AI-Provider Credentials (BYO key + subscription login) — Phase 2, e2e

## Summary

Add an **opt-in, per-user encrypted store** for AI-provider credentials so each teammate on a shared Archon install runs AI on **their own** API key or subscription. A near-1:1 clone of the per-user GitHub-token feature: a new `remote_agent_user_provider_keys` table (encrypted at rest with the existing `token-crypto`), connect surfaces (web + CLI) for both **API keys** and **OAuth subscription login** (Claude Pro/Max, ChatGPT/Codex, GitHub Copilot — delegated to Pi's `@earendil-works/pi-ai/oauth`), and **per-runtime injection** resolved once in `executeWorkflow` (and the direct-chat path) and merged into `config.envVars` exactly like the bot GitHub token. Gated on `TOKEN_ENCRYPTION_KEY`; solo installs and the no-key path are byte-for-byte unchanged.

## User Story

As an **operator running a shared Archon VPS for a small team**
I want **each teammate to connect their own AI provider key or subscription once**
So that **every run and chat bills to that person's own account/model, instead of pooling onto one shared key** — with bundled tier workflows (Phase 1) already adapting per-user.

## Problem Statement

Provider credentials are **process-global** today: Claude reads `CLAUDE_API_KEY`/`CLAUDE_CODE_OAUTH_TOKEN` from `process.env`, Codex `OPENAI_API_KEY`, Pi its backend keys — all shared across every user on the box (`providers/src/claude/provider.ts:88`, `codex/provider.ts:97`, `pi/provider.ts:276`). The web-auth seam (PR #1841) gives a real `{ userId }` per request, and Phase 1 (#1867/#1873) makes models per-user-addressable, but **nothing stores or injects a per-user credential**. There is no per-user place for "my Anthropic key" or "my ChatGPT subscription," and no path from `userId` -> a decrypted credential -> the provider invocation.

## Solution Statement

Mirror the per-user GitHub token end-to-end:

[...]

PR #1918, opened 2026-06-08
fix(server): exempt /api/auth/providers from the Better Auth catch-all

Wirasm:
## Summary

- **Problem:** `GET /api/auth/providers` returns **404 whenever web auth is enabled** (reproduced live on archon.dynamous.ai) — so the new console **AI Provider Keys** panel can't load.
- **Root cause:** Better Auth's `basePath` is `/api/auth`, so its catch-all (`app.on(['GET','POST'], '/api/auth/*')`) claims the whole prefix and 404s any path it doesn't own. The provider-key routes from #1911 were never added to the `isArchonOwnedAuthPath` fall-through allow-list. (PUT/DELETE worked because the catch-all only intercepts GET/POST — only the list `GET` broke.)
- **Why CI was green:** the route unit tests exercise `registerApiRoutes` directly and never go through the `index.ts` Better Auth mount, so the shadowing was invisible to them.
- **What changed:** extracted the inline exemption into a tested pure helper and added `/api/auth/providers`; plus an *airtight guard test*.
- **What did NOT change:** the route paths, the handlers, the gate, the `/api/auth/github*` behavior.

## Fix

- `auth/config.ts`: `isArchonOwnedAuthPath` is now a pure exported helper (was an untested inline literal in `index.ts`), covering `status` + `github*` + **`providers*`** (the `/providers/` prefix also pre-exempts PR-3's OAuth subroutes).
- `index.ts`: imports the helper; the inline literal is gone.
- Tests:
  - unit test of the helper (exempts status/github*/providers*; does **not** exempt Better Auth's own `sign-in`/`get-session`/etc., with prefix-boundary cases like `providersX`).
  - **guard test**: enumerates every Archon-registered `/api/auth/*` route from `registerApiRoutes` and asserts each is exempted — so a future forgotten route fails CI instead of 404ing in prod.

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `server`, `tests`
- Module: `server:auth`

## Change Metadata

- Change type: `bug`
- Primary scope: `server`

## Linked Issue

- Follow-up to #1911 (PR 2/4 of #1891)

## Validation Evidence

```bash
bun --filter @archon/server type-check   # ✅
bun test packages/server/src/auth/config.test.ts            # ✅ 21 (incl. exemption cases)
bun test packages/server/src/routes/api.provider-keys.test.ts  # ✅ 14 (incl. guard)
bun run lint && bun run format:check     # ✅
```

- Reproduced the 404 live (web auth on); will redeploy this fix to the VPS and confirm the panel loads.

## Security Impact

- New permissions/capabilities? **No** — narrows nothing; only routes `GET /api/auth/providers` to Archon's existing (already-gated, `requireWebUser`) handler instead of Better Auth's 404.

## Compatibility / Migration

- Backward compatible? **Yes** (web-auth-off installs were never affected; this only changes behavior when Better Auth is mounted).
- Config/env changes? **No** · Database migration? **No**

## Human Verification

- Verified: unit + guard tests green; reproduced the live 404.
- Not verified yet: post-deploy panel load (will confirm on the VPS after merge).

- Only `/api/auth/providers` GET routing when web auth is on. The guard test prevents regressions for any future `/api/auth/*` route.

## Rollback Plan

- Revert this PR — the feature returns to "panel 404s under web auth" (no other surface affected).

## Risks and Mitigations

- Risk: another `/api/auth/*` route added later without exemption.
  - Mitigation: the guard test fails CI if any Archon-registered `/api/auth/*` route isn't in `isArchonOwnedAuthPath`.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for authentication route handling to ensure proper exemption logic is maintained.

* **Chores**
  * Enhanced authentication routing infrastructure with improved helper utilities and expanded documentation.

### DIFF
--- a/packages/server/src/auth/config.test.ts
+++ b/packages/server/src/auth/config.test.ts
@@ -6,6 +6,7 @@ import {
   isEmailAllowed,
   getSignupMode,
   isApiGateEnabled,
+  isArchonOwnedAuthPath,
 } from './config';
 
 const VALID_SECRET = 'a'.repeat(32);
@@ -116,6 +117,36 @@ describe('auth/config', () => {
           ARCHON_WEB_AUTH_REQUIRED: 'false',
         })
       ).toBe(false);
+    });
+  });
+
+  describe('isArchonOwnedAuthPath', () => {
+    test('exempts Archon-owned /api/auth/* paths (fall through, not Better Auth)', () => {
+      for (const p of [
+        '/api/auth/status',
+        '/api/auth/github',
+        '/api/auth/github/device/start',
+        '/api/auth/github/device/poll',
+        '/api/auth/providers',
+        '/api/auth/providers/openrouter',
+        '/api/auth/providers/claude/oauth/start', // reserved for PR-3
+      ]) {
+        expect(isArchonOwnedAuthPath(p)).toBe(true);
+      }
+    });
+
+    test('does NOT exempt Better Auth-owned paths (those it must handle)', () => {
+      for (const p of [
+        '/api/auth/sign-in',
+        '/api/auth/sign-up',
+        '/api/auth/sign-out',
+        '/api/auth/get-session',
+        '/api/auth/providersX', // prefix guard: must be exact or under '/'
+        '/api/auth/githubbed',
+        '/api/auth',
+      ]) {
+        expect(isArchonOwnedAuthPath(p)).toBe(false);
+      }
     });
   });
 });
--- a/packages/server/src/auth/config.ts
+++ b/packages/server/src/auth/config.ts
@@ -94,3 +94,29 @@ export function getSignupMode(
 export function isApiGateEnabled(env: NodeJS.ProcessEnv = process.env): boolean {
   return isWebAuthEnabled(env) && env.ARCHON_WEB_AUTH_REQUIRED !== 'false';
 }
+
+/**
+ * Paths under `/api/auth/*` that Archon owns and the Better Auth catch-all must
+ * NOT handle. Better Auth's `basePath` is `/api/auth`, so when web auth is on its
+ * handler claims every path under that prefix and 404s any it doesn't recognize.
+ * The mount falls through (`next()`) for these so Archon's own route handlers
+ * (registered later in `registerApiRoutes`) run instead.
+ *
+ * Keep this in lockstep with the Archon-owned `/api/auth/*` routes:
+ *   - `/api/auth/status`            (web-auth status)
+ *   - `/api/auth/github` + sub      (per-user GitHub device flow)
+ *   - `/api/auth/providers` + sub   (per-user AI-provider keys; sub = PR-3 OAuth)
+ *
+ * NOTE: only GET/POST go through the catch-all, so PUT/DELETE on these paths are
+ * never intercepted regardless — but listing them here keeps the allow-list the
+ * single source of truth for "this prefix is Archon's, not Better Auth's".
+ */
+export function isArchonOwnedAuthPath(path: string): boolean {
+  return (
+    path === '/api/auth/status' ||
+    path === '/api/auth/github' ||
+    path.startsWith('/api/auth/github/') ||
+    path === '/api/auth/providers' ||
+    path.startsWith('/api/auth/providers/')
+  );
+}
--- a/packages/server/src/auth/index.ts
+++ b/packages/server/src/auth/index.ts
@@ -11,6 +11,7 @@ export {
   isEmailAllowed,
   getSignupMode,
   isApiGateEnabled,
+  isArchonOwnedAuthPath,
   MIN_BETTER_AUTH_SECRET_LENGTH,
 } from './config';
 export { getAuth, closeAuth, resetAuthForTest, type AuthInstance } from './instance';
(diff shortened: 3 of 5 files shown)


Wirasm:
## PR Review Summary — Multi-Agent (retrospective; this PR is already merged)

Ran a scoped agent suite (code-review, docs-impact, tests, comments, simplify) on the merged fix. **Skipped** silent-failure-hunter (no error handling changed) and type-design-analyzer (no new types — the change is a single `string => boolean` predicate). This is a clean, well-engineered fix: the inline exemption literal becomes a tested pure helper, fall-through ordering is correct, and there's no over-exemption or security hole. **No critical issues.** The findings are all documentation/comment staleness the refactor introduced — worth a small fast-follow.

### Critical (0)

None. The routing fall-through is correct (catch-all is registered before `registerApiRoutes`, so `next()` dispatches to the later Archon handler), the `startsWith('/api/auth/providers/')` trailing-slash boundary correctly rejects `providersX`, and `betterAuth()` mounts only `emailAndPassword` (no social/oauth plugins) so nothing under `github/` or `providers/` can collide.

### Important (3) — doc/comment correctness; recommend a fast-follow

| ID | Issue | Location |
|----|-------|----------|
| D1 | **`CLAUDE.md` exemption list is now factually incomplete** — the "Web Auth" bullet says the catch-all "falls through for Archon-owned `/api/auth/status` + `/api/auth/github*` paths." It must now also list **`/api/auth/providers*`**. This directly contradicts the new code docstring that says "keep this in lockstep," so it will mislead anyone reasoning about the auth boundary. | `CLAUDE.md` (~L891) |
| C1 | **Stale cross-reference introduced by the refactor** — `// (see isArchonOwnedAuthPath in index.ts).` The PR moved that function out of `index.ts` into `auth/config.ts`, but this pre-existing comment (in a file not in the PR diff) wasn't updated → it now points at the wrong file. | `packages/server/src/routes/api.ts:1254` |
| C2 | **Pre-existing catch-all comment block now incomplete** — still enumerates Archon-owned paths as only "status + the GitHub device flow" and says "DELETE isn't registered here, so DELETE `/api/auth/github` is never intercepted." After #1911/#1918 there's also `GET/PUT/DELETE /api/auth/providers*`. Update the enumeration to include providers and generalize the method note (PUT *and* DELETE on any `/api/auth/*` bypass the GET/POST-only catch-all). | `packages/server/src/index.ts:655–661` |

### Suggestions (5) — nice to have

| ID | Suggestion | Location |
|----|------------|----------|
| T1 | **No test exercises the actual catch-all wiring.** The guard test proves the predicate is *exhaustive* over registered routes, but not that `index.ts` actually *calls* it — `makeApp()` runs only `registerApiRoutes`, never mounts the `app.on(['POST','GET'], '/api/auth/*', …)` middleware. A small focused test (mount the catch-all + `registerApiRoutes`, assert an exempted path reaches the Archon handler not `webAuth.handler`) closes the exact failure mode that caused the original bug. Rated 6/10 — the wiring is two lines and visible in diff review. | `packages/server/src/index.ts` (catch-all) |
| T2 | Add `/api/auth/status/` (trailing slash on an exact-match path) to the negative test set — correct-but-untested `false`. Rated 5/10, doc value. | `auth/config.test.ts` negatives |
| CM1 | Soften **"Airtight guard"** → "Exhaustiveness check" — the test enforces predicate completeness, not runtime wiring. "Airtight" overstates the coverage boundary. | `api.provider-keys.test.ts:391` |
| CM2 | Clarify the docstring NOTE: explain *why* DELETE paths belong in the predicate even though the GET/POST-only catch-all never calls it for them (the guard test scans **all** `app.routes` incl. DELETE) — otherwise a future reader may "clean up" DELETE coverage and silently break the guard. | `auth/config.ts:110–112` |
| S1 | Inline the single-use `notExempted` var in the guard test (`expect(authPaths.filter(p => !isArchonOwnedAuthPath(p))).toEqual([])`). −2 lines, trivial. | `api.provider-keys.test.ts` guard test |

### Strengths

- **Single-source-of-truth extraction:** the previously inline, untestable closure becomes a pure exported predicate with its own unit tests — the right architectural move.
- **Self-maintaining guard test:** enumerating `app.routes` (rather than a hard-coded list) means every future `/api/auth/*` route is auto-included in the exhaustiveness assertion; the `authPaths.length > 0` sanity check prevents a trivially-passing empty filter; the `providersX` / `githubbed` prefix-boundary negatives are precise and deliberate.
- **Correct root-cause diagnosis** in the PR body — including *why CI stayed green* (route tests bypass the `index.ts` Better Auth mount) — and the exact-or-prefix boundary pairs are a deliberate, correct guard (not a simplification target).
- The new docstring's factual claims (prefix list, GET/POST-only, "registered later in `registerApiRoutes`") all check out against the code.

### Verdict: **MERGED — sound fix; minor doc/comment follow-up recommended**

No code defects. Worth a quick fast-follow PR for **D1** (the `CLAUDE.md` exemption list — it contradicts the code's own lockstep instruction), **C1**, and **C2** (the two stale comments the refactor left behind). T1 (wiring test) and the CM/S polish are optional.

<sub>Aggregated from code-reviewer, docs-impact, pr-test-analyzer, comment-analyzer, code-simplifier. silent-failure-hunter & type-design-analyzer intentionally skipped (no error-handling or type changes). Advisory only — no files changed.</sub>

Issue #1933, opened 2026-06-09
Console store cache.ts: versions/cache/errors Maps leak on last-subscriber unsubscribe


leex279:
## Summary

- What broke: In the console experiment's reactive server-state cache, when the **last subscriber** for a key unsubscribes, the cleanup deletes the `listeners` and `loaders` entries but leaves the `versions` (and `cache`/`errors`) entries behind. Those Maps grow unbounded across the set of distinct keys seen over a session.
- When it started (if known): Present in `packages/web/src/experiments/console/store/cache.ts` as shipped. Surfaced by the simplify pass on #1928 (intentionally left out of PR #1930 to keep that fix single-purpose).
- Severity: `minor` (slow memory growth in the console experiment; not user-visible short-term)

## Steps to Reproduce

1. Use the console UI; mount/unmount many components that call `useEntity(key, …)` with distinct keys (navigating runs/projects churns keys).
2. Each unique key bumps `versions` on every `notify` and is never pruned.
3. Over a long session, `versions` (and `cache`/`errors`) retain entries for keys with zero live subscribers.

## Expected vs Actual

- **Expected**: When a key's last subscriber unsubscribes (`s.size === 0`), all per-key Map entries for that key are released.
- **Actual**: Only `listeners` and `loaders` are deleted. `versions`, `cache`, and `errors` entries persist indefinitely.

## User Flow

```
Component                 cache.ts store
─────────                 ──────────────
useEntity(key) ─────────▶ listeners.set / loaders.set / versions bumped on notify
unmount (last sub) ─────▶ unsubscribe: listeners.delete(key)
                                        loaders.delete(key)
                          [X] versions.delete(key)  ← missing
                          [X] cache.delete(key)      ← missing
                          [X] errors.delete(key)     ← missing
```

## Environment

- Platform: Web (console experiment)
- Database: N/A (client-side store)
- Running in worktree? No
- OS: N/A (browser)

## Logs

```
N/A — memory growth, no runtime error
```

## Impact

- Affected workflows/commands: Console UI (`/console`) long-lived sessions
- Reproduction rate: Always (deterministic; magnitude scales with distinct-key count)
- Workaround available? Page reload clears the module-level Maps
- Data loss risk? No

## Scope

- Package(s) likely involved: `web`
[...]

PR #1938, opened 2026-06-10
fix(web/console): release per-key version counter on last unsubscribe (#1933)

truffle-dev:
## Summary

- Problem: the console store's `versions` Map grows unbounded — `notify()` creates an entry for every key a session ever touches and nothing ever deletes it, so long-lived console sessions leak one counter per run/message/entity key forever (#1933).
- Why it matters: the console is designed to stay open for hours with SSE pushing new keys continuously; an append-only Map is a slow memory leak in the default UI.
- What changed: `versions.delete(key)` at the two points a key's lifecycle ends — the last-subscriber unsubscribe, and `revalidate`'s no-loader branch (the `invalidate()` prune path for subscriber-less keys). The subscribe logic is extracted into an exported `subscribeKey()` plain function so the lifecycle is unit-testable, with `useEntity` rewired on top. New `cache.test.ts` pins the behavior.
- What did **not** change (scope boundary): `cache` and `errors` retention after last unsubscribe is untouched — the module contract promises warm remounts (`ensureLoad` short-circuits on a cached value), and the issue itself flags the `versions` leak as the unambiguous part while leaving cache/errors policy open. A test pins the warm-remount retention so a future eviction policy is an explicit decision, not a drift.

### Before

```
User                          Console store
────                          ─────────────
opens console        ──────▶  useEntity subscribes; versions[key] created
navigates runs/chats ──────▶  more keys: versions grows
panel unmounts       ──────▶  listeners/loaders pruned; versions entry LEAKS
hours of SSE traffic ──────▶  versions grows unbounded, one entry per key ever seen
```

### After

```
User                          Console store
────                          ─────────────
opens console        ──────▶  useEntity subscribes; versions[key] created
navigates runs/chats ──────▶  more keys: versions grows
panel unmounts       ──────▶  listeners/loaders pruned; [versions entry deleted]*
invalidate(prefix)   ──────▶  subscriber-less keys: cache+errors+[versions]* all released
hours of SSE traffic ──────▶  versions bounded by currently-subscribed keys
```

### Before

```
useEntity (hook)
  └── inline subscribe closure ──▶ listeners / loaders / cache / errors / versions (module Maps)
        unsubscribe: prunes listeners + loaders only
revalidate (no-loader branch): prunes cache + errors only
```

### After

```
useEntity (hook)
  └── subscribeKey() [+exported]  ===▶ listeners / loaders / cache / errors / versions
        unsubscribe: prunes listeners + loaders [~ + versions]
revalidate (no-loader branch): prunes cache + errors [~ + versions]
versionOf() [+exported] ──▶ versions (snapshot read, backs useSyncExternalStore)
cache.test.ts [+] ──▶ subscribeKey / versionOf / get / set / invalidate
```

| useEntity | subscribeKey | **new** | extracted from inline closure; behavior identical |
| useEntity | versionOf | **new** | snapshot reads via named export instead of inline lambda body |
| subscribeKey unsubscribe | versions | **new** | `versions.delete(key)` when last subscriber leaves |
| revalidate (no-loader) | versions | **new** | `versions.delete(key)` alongside existing cache/errors prune |
| cache.test.ts | store/cache | **new** | test file |
| useEntity | cache/errors/inflight | unchanged | data/error/loading reads as before |
| SSE dispatcher / skills | set/patch/invalidate | unchanged | public API surface identical |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `web`
- Module: `web:console-store`

## Change Metadata

- Change type: `bug`
- Primary scope: `web`

## Linked Issue

- Part of #1933 — resolves the `versions` leak; the `cache`/`errors` retention is a deliberate scope boundary (see Summary), left open so merging this doesn't auto-close that part.

Commands and result summary:

```bash
bun run validate   # check:bundled, check:bundled-skill, check:bundled-schema,
                   # type-check, lint, format:check all pass
cd packages/web && bun run test   # all five serial suites pass, 0 fail
                                  # (src/experiments/console/: 102 pass incl. 5 new)
```

- Evidence provided: new `cache.test.ts` — 5 tests covering last-unsubscribe prune, non-last-unsubscribe retention, notification stop after cleanup, full release of a subscriber-less key via `invalidate` (including the SSE-push-recreates-entries case), and the errored-key path. Plus a warm-remount assertion proving the loader is NOT re-invoked after unsubscribe/resubscribe (pins the deliberate cache retention).
- If any command is intentionally skipped, explain why: `bun run validate`'s parallel test fan-out crashed in my container (pid-limit SIGTRAP, all packages at once); reran the web package's serial test script directly — exit 0. No other package is touched.

## Security Impact (required)

- If any `Yes`, describe risk and mitigation: n/a

## Compatibility / Migration

- Backward compatible? Yes
- Config/env changes? No

- Verified scenarios: traced every `versions` reader — only `versionOf` (the `useSyncExternalStore` snapshot) and `notify` (read-increment-write). With zero subscribers nothing snapshots the key, and `useSyncExternalStore` only compares snapshots for change, so a remount restarting the counter at 0 is observationally identical.
- Edge cases checked: (1) two subscribers on one key — counter survives the first unsubscribe; (2) error-state key (lives only in `errors`, not `cache`) — counter still released; (3) SSE `set()` on an unsubscribed key recreates cache+versions entries — `invalidate()`'s no-loader branch now releases all three Maps; (4) in-flight load resolving after unmount — `ensureLoad`'s `.then` repopulates cache+versions for a dead key; bounded by the key set and reclaimed by the same `invalidate` path, noted in the code comment.
- What was not verified: long-session heap profiling in a real browser; the fix is asserted at the Map-lifecycle level.

- Affected subsystems/workflows: console experiment UI only (`packages/web/src/experiments/console/store/cache.ts`); production web UI and server untouched.
- Potential unintended effects: a remounted key's version counter restarts at 0 instead of continuing — safe per the reader trace above (snapshots are compared for inequality, never persisted across subscriptions).
- Guardrails/monitoring for early detection: the 5 new tests pin the lifecycle; any future change that reintroduces the leak or breaks warm remount fails the suite.

## Rollback Plan (required)

- Fast rollback command/path: `git revert` of this single commit; no migrations, no config.
- Observable failure symptoms: console panels failing to re-render on data updates after a remount (would indicate the counter-reset reasoning was wrong; tests cover it).

## Risks and Mitigations

- Risk: extraction of `subscribeKey` could subtly change `useEntity` subscription semantics.
  - Mitigation: the function body is a verbatim move of the previous inline closure plus the one-line prune; `useEntity` passes the same `loaderRef`-wrapped loader; the full console suite (102 tests) passes unchanged.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for cache subscription lifecycle behaviors, including version resets on unsubscribe, cache retention for remounts, handling of multiple subscribers, cache invalidation, and error scenarios.

* **Refactor**
  * Improved internal cache version management to ensure proper lifecycle handling as subscribers are added and removed, resulting in more reliable and consistent state synchronization behavior.

### DIFF
--- a/packages/web/src/experiments/console/store/cache.test.ts
+++ b/packages/web/src/experiments/console/store/cache.test.ts
@@ -95,15 +95,79 @@ describe('subscribeKey — per-key Map lifecycle (#1933)', () => {
     await flush();
     unsubscribe();
 
-    // An SSE push for an unsubscribed key re-creates cache + versions entries.
+    // An SSE push for an unsubscribed key updates the still-warm cache; the
+    // version counter stays released because notify skips subscriber-less keys.
     set(key, 'pushed');
     expect(get(key)).toBe('pushed');
-    expect(versionOf(key)).toBe(1);
+    expect(versionOf(key)).toBe(0);
 
-    // With no loader registered, revalidate's prune branch drops everything.
+    // With no loader registered, revalidate's prune branch drops the cache too.
     invalidate(key);
     expect(get(key)).toBeUndefined();
     expect(versionOf(key)).toBe(0);
+  });
+
+  test('a load resolving after the last unsubscribe does not resurrect the counter', async () => {
+    const key = 'test:inflight-resolve';
+    let resolveLoad: (v: string) => void = () => {};
+    const unsubscribe = subscribeKey(
+      key,
+      () => {},
+      () =>
+        new Promise<string>(resolve => {
+          resolveLoad = resolve;
+        })
+    );
+
+    unsubscribe(); // last subscriber leaves while the load is still in flight
+    expect(versionOf(key)).toBe(0);
+
+    resolveLoad('late');
+    await flush();
+    expect(versionOf(key)).toBe(0); // notify skipped — nothing subscribes
+    expect(get(key)).toBe('late'); // cache still warms for a future remount
+  });
+
+  test('a load rejecting after the last unsubscribe does not resurrect the counter', async () => {
+    const key = 'test:inflight-reject';
+    let rejectLoad: (e: Error) => void = () => {};
+    const unsubscribe = subscribeKey(
+      key,
+      () => {},
+      () =>
+        new Promise<string>((_resolve, reject) => {
+          rejectLoad = reject;
+        })
+    );
+
+    unsubscribe();
+    expect(versionOf(key)).toBe(0);
+
+    rejectLoad(new Error('late boom'));
+    await flush();
+    expect(versionOf(key)).toBe(0);
+  });
+
+  test('invalidate by prefix releases every matching subscriber-less key', async () => {
+    const unsubA = subscribeKey(
+      'test-prefix:a',
+      () => {},
+      () => Promise.resolve('a')
+    );
+    const unsubB = subscribeKey(
+      'test-prefix:b',
+      () => {},
+      () => Promise.resolve('b')
+    );
+    await flush();
+    unsubA();
+    unsubB();
+
+    invalidate('test-prefix');
+    expect(get('test-prefix:a')).toBeUndefined();
+    expect(get('test-prefix:b')).toBeUndefined();
+    expect(versionOf('test-prefix:a')).toBe(0);
+    expect(versionOf('test-prefix:b')).toBe(0);
   });
 
   test('errored key still releases its counter on last unsubscribe', async () => {
--- a/packages/web/src/experiments/console/store/cache.test.ts
+++ b/packages/web/src/experiments/console/store/cache.test.ts
@@ -1,0 +1,123 @@
+import { describe, test, expect } from 'bun:test';
+import { subscribeKey, versionOf, get, set, invalidate } from './cache';
+
+// The store's Maps are module-level, so every test uses its own unique key —
+// no cross-test state to reset.
+
+function flush(): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, 0));
+}
+
+describe('subscribeKey — per-key Map lifecycle (#1933)', () => {
+  test('last unsubscribe prunes versions; cache is retained for warm remount', async () => {
+    const key = 'test:prune-versions';
+    let loads = 0;
+    const unsubscribe = subscribeKey(
+      key,
+      () => {},
+      () => {
+        loads += 1;
+        return Promise.resolve('v1');
+      }
+    );
+    await flush();
+
+    expect(get(key)).toBe('v1');
+    expect(versionOf(key)).toBe(1); // notify bumped on load resolve
+    expect(loads).toBe(1);
+
+    unsubscribe();
+    expect(versionOf(key)).toBe(0); // counter released
+    expect(get(key)).toBe('v1'); // cached value deliberately retained
+
+    // Remount reads warm: ensureLoad short-circuits on the cached value.
+    const unsubscribe2 = subscribeKey(
+      key,
+      () => {},
+      () => {
+        loads += 1;
+        return Promise.resolve('v2');
+      }
+    );
+    await flush();
+    expect(get(key)).toBe('v1');
+    expect(loads).toBe(1); // loader not re-invoked
+    unsubscribe2();
+  });
+
+  test('unsubscribing a non-last subscriber keeps the counter', async () => {
+    const key = 'test:non-last';
+    const unsubA = subscribeKey(
+      key,
+      () => {},
+      () => Promise.resolve('a')
+    );
+    const unsubB = subscribeKey(
+      key,
+      () => {},
+      () => Promise.resolve('a')
+    );
+    await flush();
+    expect(versionOf(key)).toBe(1);
+
+    unsubA();
+    expect(versionOf(key)).toBe(1); // B still subscribed
+
+    unsubB();
+    expect(versionOf(key)).toBe(0);
+  });
+
+  test('unsubscribe stops change notifications', async () => {
+    const key = 'test:notify-stops';
+    let renders = 0;
+    const unsubscribe = subscribeKey(
+      key,
+      () => {
+        renders += 1;
+      },
+      () => Promise.resolve('a')
+    );
+    await flush();
+    expect(renders).toBe(1);
+
+    unsubscribe();
+    set(key, 'b');
+    expect(renders).toBe(1); // no further signal after cleanup
+  });
+
+  test('invalidate fully releases a subscriber-less key (cache + versions)', async () => {
+    const key = 'test:invalidate-prune';
+    const unsubscribe = subscribeKey(
+      key,
+      () => {},
+      () => Promise.resolve('a')
+    );
+    await flush();
+    unsubscribe();
+
+    // An SSE push for an unsubscribed key re-creates cache + versions entries.
+    set(key, 'pushed');
+    expect(get(key)).toBe('pushed');
+    expect(versionOf(key)).toBe(1);
+
+    // With no loader registered, revalidate's prune branch drops everything.
+    invalidate(key);
+    expect(get(key)).toBeUndefined();
+    expect(versionOf(key)).toBe(0);
+  });
+
+  test('errored key still releases its counter on last unsubscribe', async () => {
+    const key = 'test:error-prune';
+    const unsubscribe = subscribeKey(
+      key,
+      () => {},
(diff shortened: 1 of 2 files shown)


truffle-dev:
Green for a few weeks now (both test matrices + docker-build + CodeRabbit), so flagging the one real judgment call in case that's what's holding it, not a code gap.

#1933 names three leaking Maps — `versions`, `cache`, `errors` — but this prunes only `versions`, deliberately. `cache`/`errors` retention is what lets a warm remount skip the loader (`ensureLoad` short-circuits on a cached value); evicting them on last-unsubscribe would turn every remount into a refetch. `versions` is pure overhead with no such tradeoff, so I fixed that and pinned the warm-remount behavior with a test rather than quietly changing eviction policy.

I've softened the link to `Part of #1933` so merging this doesn't auto-close the `cache`/`errors` question. So the call is yours:

- take versions-only as-is, and #1933 stays open to track the cache/errors eviction policy as its own decision, or
- say the word and I'll extend the same unsubscribe/invalidate paths to drop `cache`+`errors` too and swap the warm-remount test for the full-sweep behavior.

Either's quick — just didn't want to guess on the retention contract.

Wirasm:
## PR Review Summary

Multi-agent review (`/prp-review 1938 all`): 7 specialist agents in parallel — code-reviewer, docs-impact, pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier.

### Critical Issues (0 found)

None.

### Important Issues (1 found)

| Agent | Issue | Location |
|-------|-------|----------|
| code-reviewer + silent-failure-hunter + pr-test-analyzer (each reproduced it independently) | An in-flight load resolving **after** the last unsubscribe resurrects the `versions` entry (and repopulates `cache`/`errors`) for a dead key: cleanup doesn't touch `inflight`, and `notify()` unconditionally does `versions.set(...)` with no listener check. Every load abandoned by fast navigation re-leaks one entry until a later `invalidate()` happens to hit that key — exactly the click-through pattern behind the original issue. The inline comment ("Without this the `versions` Map grows unbounded…") and the PR body's "noted in the code comment" overstate the guarantee — no such acknowledgment exists in `cache.ts`. The same gap applies to `revalidate`'s no-loader branch: invalidating a subscriber-less key with a pending stale load gets silently repopulated when it settles. | `cache.ts:31-36` (`notify`), `38-57` (`ensureLoad`), `160-176` (cleanup) |

**Suggested fix (small)**: guard `notify()` — skip the `versions.set` bump when `listeners.get(key)` is undefined (keep the cache/errors writes for warm remount) — or track a cancelled/epoch flag checked in `ensureLoad`'s `.then`/`.catch`. Add the missing regression test: unsubscribe *before* the loader resolves, then assert `versionOf(key)` stays 0 after the late resolve (~15 lines, fits the existing suite). If instead the residual is accepted as bounded, say so in the code comment (as the PR body claims) and pin current behavior with a test.

### Suggestions (7 found)

| code-simplifier | `notify()` still inlines `versions.get(key) ?? 0` — use the new `versionOf()` accessor so the default-0 fallback lives in one place | `cache.ts:32` |
| comment-analyzer | "(see the module contract above)" is a dangling pointer — the top `Contract:` block says nothing about cache/errors retention across zero-subscriber periods; extend the block or drop the pointer | `cache.ts:172` |
| comment-analyzer | Test comment "re-creates cache + versions entries" is imprecise — cache is still warm (only overwritten); only `versions` is genuinely re-created | `cache.test.ts:104` |
| pr-test-analyzer (6/10) | Pre-existing, adjacent: resubscribing during an abandoned in-flight load never invokes the new loader (`ensureLoad`'s `inflight.has(key)` short-circuit) — the new subscriber silently inherits the stale request's result. Not introduced here, but this PR makes fast unmount/remount lean on that dedup more. Worth a follow-up issue | `cache.ts:39` |
| pr-test-analyzer (5/10) | `invalidate()` prefix fan-out (releasing multiple subscriber-less keys in one call) has never been tested — cheap to add now that the harness exists | `cache.ts:111-130` |
| type-design-analyzer | Add a one-line "test-only export; production code goes through `useEntity`" note on `subscribeKey`/`versionOf` so a future direct caller doesn't bypass `useSyncExternalStore`'s cleanup pairing and reintroduce the leak class | `cache.ts:145,180` |
| code-simplifier | Rename the single-letter `s` in the unsubscribe closure (pre-existing, relocated by the diff) to e.g. `remainingSubs`; optionally dedup the 3× no-op `subscribeKey` boilerplate in tests | `cache.ts:161-176` |

### Strengths

- The `subscribeKey()` extraction is a verbatim, semantics-preserving move — verified line-by-line against the pre-PR closure by two agents; `loaders.set`-per-subscribe behavior is unchanged.
- The `useSyncExternalStore` stale-render concern was traced and cleared: the counter is monotonic for any live subscription and data/error are read fresh per render — no torn-render case constructible from the reset-to-0.
- Tests are behavioral and refactor-resilient: public surface only, clean teardown, unique per-test keys; warm-remount retention pinned with a loader-invocation count.
- Comments were fact-checked and hold up: the `handleBuilderKeydown`/`useBuilderKeyboard` precedent cited in the JSDoc genuinely exists and matches the extraction shape.
- Good scope discipline: cache/errors retention deliberately out of scope, argued in the PR body, and pinned by a test.

### Documentation Issues

- None — no documentation references the cache module's internals; CLAUDE.md, README.md, docs-web, and both experiment READMEs are unaffected.
- Tracker note: the linked issue's "Expected" asks for `versions`, `cache`, **and** `errors` to be released; this PR deliberately releases only `versions`. The "Part of" (not "Closes") framing is correct — keep it so the issue isn't closed as fully resolved.

### Validation

| Check | Result |
|-------|--------|
| Type Check | PASS (`packages/web` tsc clean; root failure is pre-existing on `dev` in the untouched Copilot provider) |
| Lint | PASS (0 warnings) |
| Tests | PASS (`packages/web`: 102 pass / 0 fail, incl. the 5 new cache tests) |
| Build | PASS (`@archon/web` vite build) |

### Verdict

NEEDS FIXES

### Recommended Actions

1. Close or explicitly document+test the in-flight-resolve-after-unsubscribe race (the one important issue) — a listener check in `notify()` is a ~2-line change plus one test.
2. Batch the two comment fixes (`cache.ts:172` pointer, `cache.test.ts:104` wording) and the `versionOf()`-in-`notify` consistency change into the same pass.
3. Consider filing a follow-up for the pre-existing stale-loader-on-resubscribe quirk (`cache.ts:39`).
4. Re-run review after fixes.

---
*Report: `.claude/PRPs/reviews/pr-1938-review.md`*

Wirasm:
Pushed `9ff678f6` addressing the review above.

**Fixed**
- **In-flight race (the one important issue)**: `notify()` now returns early when a key has no subscribers, instead of unconditionally bumping the counter. This closes the resurrect path for both a load settling after the last unsubscribe (resolve *and* reject) and an SSE `set()` on an unwatched key — cache/error writes still happen, so remounts read warm. Pinned by two new tests (`test:inflight-resolve`, `test:inflight-reject`).
- `notify()` now uses `versionOf()` (code-simplifier consistency point).
- The dangling "(see the module contract above)" pointer: the `Contract:` block now states the cache/errors-retention rule the comment references.
- Test comment/expectation at the SSE-push test corrected: cache is *updated* (still warm), not re-created — and with the guard, the counter now stays released there too.
- `subscribeKey`/`versionOf` JSDoc notes they're exported for tests; production goes through `useEntity`.
- `invalidate()` prefix fan-out test added (releases multiple subscriber-less keys in one call).
- Renamed the single-letter `s` in the unsubscribe closure to `remainingSubs`.

**Deliberately not addressed**
- The pre-existing `ensureLoad` inflight-dedup quirk (a resubscribe during an abandoned in-flight load inherits the old request's result without invoking the new loader) — pre-dates this PR and is a behavior question, not a leak; better as a follow-up issue than a scope creep here.
- `EntityView` discriminated union / collapsing the parallel Maps into one `Map<string, Entry>` — flagged by the type-design pass as worth revisiting only if this store graduates out of `experiments/`; over-engineering for now.
- Test no-op-subscribe helper dedup — first appearance of the pattern; the reviewer itself rated it note-don't-push.

Validation after the changes: `packages/web` tsc clean, 105/105 tests (8 in `cache.test.ts`), eslint 0 warnings, prettier clean.

PR #1952, opened 2026-06-11
fix(commands): scope self-fix + simplify to the PR diff

sbiitmc:
Both archon-self-fix-all and archon-simplify-changes could edit files outside the PR diff (scope-bleed): self-fix never computed the diff and its "fix EVERYTHING / add tests for existing code" framing pushed edits outward; simplify computed the diff but never enforced it.

Both nodes now compute an allow-list (git diff $BASE...HEAD --name-only) and run a hard pre-commit gate (comm -23 of changed files vs the allow-list) that reverts anything out of scope before staging. self-fix's framing is re-toned to "aggressive within the diff; out-of-scope findings become follow-up issues, never edits". Prompt-only; bundled-defaults.generated.ts regenerated.

## Summary

- Problem:
- Why it matters:
- What changed:
- What did **not** change (scope boundary):

### Before

```
(Draw the user-facing flow BEFORE this PR. Show each step the user takes.)

Example:
  sends message ──────▶  resolves session
                         loads context
                         streams to AI ──────────▶ processes prompt
                         receives chunks ◀──────── streams response
  sees reply ◀─────────  sends to platform
```

### After

```
(Draw the user-facing flow AFTER this PR. Highlight what changed with [brackets] or asterisks.)
```

### Before

```
(Map ALL modules touched or connected to this change. Draw lines between them.)
```

### After

```
(Same diagram with changes highlighted. Mark new modules with [+], removed with [-],
 modified with [~]. Mark new connections with ===, removed with --x--.)
```

| | | unchanged / **new** / **removed** / **modified** | |

## Label Snapshot

- Risk: `risk: low|medium|high`
- Size: `size: XS|S|M|L|XL`
- Scope: `core|workflows|isolation|git|adapters|server|web|cli|paths|config|docs|dependencies|ci|tests|skills`
- Module: `<scope>:<component>` (e.g. `workflows:executor`, `adapters:slack`, `core:orchestrator`)

## Change Metadata

- Change type: `bug|feature|refactor|docs|security|chore`
- Primary scope: `core|workflows|isolation|git|adapters|server|web|cli|paths|multi`

## Linked Issue

- Closes #
- Related #
- Depends on # (if stacked)
- Supersedes # (if replacing older PR)

Commands and result summary:

```bash
bun run type-check
bun run lint
bun run format:check
bun run test
# Or all at once:
bun run validate
```

- Evidence provided (test/log/trace/screenshot):
- If any command is intentionally skipped, explain why:

## Security Impact (required)

- New permissions/capabilities? (`Yes/No`)
- New external network calls? (`Yes/No`)
- Secrets/tokens handling changed? (`Yes/No`)
- File system access scope changed? (`Yes/No`)
- If any `Yes`, describe risk and mitigation:

## Compatibility / Migration

- Backward compatible? (`Yes/No`)
- Config/env changes? (`Yes/No`)
- Database migration needed? (`Yes/No`)
- If yes, exact upgrade steps:

- Verified scenarios:
- Edge cases checked:
- What was not verified:

- Affected subsystems/workflows:
- Potential unintended effects:
- Guardrails/monitoring for early detection:

## Rollback Plan (required)

- Fast rollback command/path:
- Feature flags or config toggles (if any):
- Observable failure symptoms:

## Risks and Mitigations

List real risks in this PR (or write `None`).

- Risk:
  - Mitigation:

## Summary by CodeRabbit

* **Chores**
  * Enhanced automated workflow validation with stricter scope boundaries for fixes and simplifications.
  * Workflows now verify changed files before committing, revert out-of-scope edits, and allow only approved related test or documentation updates.
  * Out-of-scope findings are routed to follow-up issues instead of being applied automatically.
  * Reports now include scope compliance details and documented mechanical consequences.


Wirasm:
The outcome you're going for here is one we want. Scoping the auto-fix and auto-simplify stages to the PR's own diff is right — "fix EVERYTHING, add tests for existing code" is how a fix PR turns into an unreviewable sprawl, and flipping to "aggressive inside the diff, follow-up issue outside it" is the correct instinct.

**But the shape is wrong, and I'd rather hold it than land it and rebuild it.**

## Why hold: the gate's state has nowhere correct to live yet

The allow-list is written to fixed host-global paths — `/tmp/scope-allowlist.txt` (`head:66`), `/tmp/changed.txt` (`head:206`), `/tmp/simplify-allowlist.txt` (`archon-simplify-changes.md:55`).

Archon runs workflows concurrently by design, and worktree isolation does **not** isolate `/tmp`. So two simultaneous `archon-fix-github-issue` runs clobber each other's allow-list. Run A then compares its changed files against run B's list, `comm -23` reports A's legitimate edits as out-of-scope, and the prompt instructs `git checkout -- <path>` on them (`head:211`). **It reverts real fixes, silently, in the run that was doing nothing wrong.**

`$ARTIFACTS_DIR` fixes that today — per-run, pre-created by the executor, outside the worktree so it can never be staged. But this is exactly the class of thing #2200 is about: workflow state and output moving to a unified per-project tree in `~/.archon`, with the repo keeping only source. A scoping gate that persists per-run state is a first-class consumer of that substrate, and #2200 will define where it belongs. Landing this on `/tmp` now means writing it twice — once wrong, once again after the rewrite.

So: held until #2200 lands, then rebuilt on whatever it defines. Not a rejection of the idea.

## Two other things worth fixing whenever it does land

- **The gate forbids mechanical consequences.** As written, an in-scope change that requires a regenerated artifact (`bun run generate:bundled`) or a manifest entry (a `package.json` test split) is out of scope and gets reverted — turning a green-able PR red plus a follow-up issue. It needs a carve-out: a file that *must* change as a mechanical consequence of an in-scope change is allowed, with the regeneration run via the documented command rather than hand-edited, and noted in the fix report.

- **The gate can't see untracked files.** `git diff --name-only` (`head:206`, `archon-simplify-changes.md:75`) lists only tracked modifications, so a stray **new** out-of-scope file never appears — yet `head:211` tells the agent to `rm` such files as though it would. That's precisely the class #1506 was written to stop, and it means "hard gate" currently overstates what it catches. `git status --porcelain`, or union in `git ls-files --others --exclude-standard`.

## Sequencing note

This conflicts with #2199 inside `### 5.1` of `archon-self-fix-all.md` and in the `archon-simplify-changes.md` numbered list, plus the two bundle entry lines. When both are ready, land this one first — it's the larger rewrite of the same region, and #2199's telemetry additions rebase onto it cheaply. The reverse forces re-deriving the whole §5.1 rewrite. Whichever goes second runs `bun run generate:bundled`.

## What I'd ask of you now

Nothing — please don't rebase this repeatedly while it waits. I'll come back to it when #2200 has defined the state location, and at that point the change is mostly "point the allow-list at the right directory" plus the two fixes above.

Thanks for the work; the diagnosis of what was wrong with the old prompt framing is the valuable part and it survives the hold intact.

Wirasm:
superseeded by 2627

PR #1968, opened 2026-06-11
fix(oauth-bridge): release the fixed OAuth callback port when a login is abandoned

Wirasm:
## Summary

Describe this PR in 2-5 bullets:

- Problem: an abandoned anthropic subscription login wedged pi-ai's fixed-port (53692) OAuth callback server install-wide — every later anthropic OAuth start (any user) failed with an opaque `500 Failed to start subscription login` until process restart (#1963).
- Why it matters: on multi-user installs one abandoned login bricks subscription connect for everyone; severity rises now that the console credential cards (#1962) make login discoverable.
- What changed: the bridge now hard-cancels abandoned logins through the one handle pi-ai 0.79.1 actually reacts to — rejecting the bridge-owned `onManualCodeInput()` deferred — which drives pi's login through `finally { server.close() }` and releases the port. Cancellation runs on every abort path (same-user restart, cross-user supersede for callback-server vendors, expiry sweep, explicit cancel). Residual port collisions are classified into a typed `OAuthCallbackPortBusyError` with an actionable message, mapped to 503 by the route.
- What did **not** change (scope boundary): the login UX (start/poll contract), the persisted credential shape, the device-code (Copilot) flow, and pi-ai itself (upstream issue filed: https://github.com/earendil-works/pi/issues/5649). The openai/ChatGPT gate (#1924) is untouched here — lifted in the stacked follow-up PR.

### Before

```
  User A                  Archon bridge                pi-ai loginAnthropic
  ──────                  ─────────────                ────────────────────
  start login ──────────▶ login() kicked ────────────▶ binds 127.0.0.1:53692
  abandons (closes tab)   abort()/sweep fires          ignores options.signal
                                                       port 53692 stays LISTEN forever
  User B
  start login ──────────▶ login() kicked ────────────▶ EADDRINUSE 53692
  sees opaque 500 ◀────── "Failed to start subscription login"
  (wedged until Archon restarts)
```

### After

```
  User A                  Archon bridge                pi-ai loginAnthropic
  ──────                  ─────────────                ────────────────────
  start login ──────────▶ login() kicked ────────────▶ binds 127.0.0.1:53692
  abandons (closes tab)
  User B
  start login ──────────▶ [abortSession: rejects the
                           manual-code deferred] ────▶ login unwinds →
                          [waits (≤1.5s) for the       finally { server.close() }
                           old login to settle]        port released
                          login() kicked ────────────▶ binds 53692 cleanly
  sees authorize URL ◀─── normal manual flow
[...]

Issue #1963, opened 2026-06-11
OAuth bridge: abandoned anthropic subscription login wedges the fixed callback port install-wide


Wirasm:
## Problem

An abandoned anthropic subscription login blocks **all** subsequent anthropic OAuth starts (any user) with `500 Failed to start server. Is port 53692 in use?` until the Archon process restarts.

Found during the #1958 VPS smoke on archon.dynamous.ai (not a #1958 regression — the bridge/abort code predates it; legacy-id normalization worked correctly even on the failing attempt).

## Root cause chain

1. `oauth-bridge.ts` (`startOAuth`, the same-user pre-abort at ~lines 130–135) calls `session.abort.abort()` to release a prior in-flight login.
2. Pi's `loginAnthropic` (`@earendil-works/pi-ai/dist/utils/oauth/anthropic.js`) **never subscribes to `options.signal`** — its local callback server is only closed in a `finally` reached after `waitForCode()` / manual-code input resolves, which never happens for an orphaned session.
3. The callback server uses a **fixed port** (53692 observed), so the leaked listener blocks every later anthropic login process-wide. Verified via `/proc/net/tcp`: the port stays LISTEN long after the abort.

## Repro

1. `POST /api/auth/providers/anthropic/oauth/start` — do not complete the browser flow.
2. Wait past the same-user abort (or use a second user) and start another anthropic login.
3. → 500, port in use. Persists until process restart.

## Suggested fixes (either side works; both is best)

- **Archon bridge**: wrap `login()` with a hard cancel — track the spawned flow and force-close/timeout the callback server when the session is aborted or sweeps expire (the bridge already sweeps expired sessions; they just don't release the socket).
- **Upstream pi-ai**: honor `options.signal` in `loginAnthropic` (close the callback server on abort) and/or bind an ephemeral port instead of a fixed one. Upstream contribution caution applies (issue → lgtm → PR flow).

## Impact

Multi-user installs: one user abandoning a login bricks subscription connect for everyone until restart. Severity rises once the #1956 console cards (PR #1962) make the login flow more discoverable.

Related: #1954 (provider-auth epic), #1958 (smoke that surfaced it), #1924 (codex OAuth gate — unrelated mechanism).

Issue #1954, opened 2026-06-11
Provider auth: restructure around the agent → credential hierarchy (epic)


Wirasm:
## Problem

The console "Provider Auth" panel and the backend credential layer conflate two unrelated concepts under one word, "provider":

- **Agent providers (harnesses)** — the `IAgentProvider` registry: `claude`, `codex`, `pi`, `opencode`, `copilot`. Drives Model Tiers / Aliases / Defaults dropdowns (`GET /api/providers`) and validates tiers/aliases/default-provider (`isRegisteredProvider`).
- **Credential providers** — a hardcoded catalog `KNOWN_PROVIDERS` in `packages/core/src/credentials/delivery.ts` (claude, codex, copilot, anthropic, openai + 7 Pi backend ids). Drives the "Provider Auth" panel (`GET /api/auth/providers`) and `archon ai key set`.

The UI renders the credential catalog as one flat dropdown — `groq` next to `claude`, raw ids, no grouping, no indication that a groq key only matters when running Pi with a groq model. Nothing ever connects "credential X" to "harness Y"; the binding happens silently in `deliverCredential`. The result is unusable: no hierarchy, no mental model.

## Concrete defects found

1. **Alias collision.** `claude` and `anthropic` are both connectable; a `claude` API key delivers `CLAUDE_API_KEY` + `ANTHROPIC_API_KEY`, an `anthropic` key delivers `ANTHROPIC_API_KEY`. Connect both and last-merged wins — no precedence rule, no warning. Same for `codex`/`openai` → `OPENAI_API_KEY`.
2. **OpenCode has no credential story at all.** It is a registered agent provider (selectable in tiers/aliases/default) but appears nowhere in the credentials layer (zero references). Its embedded runtime is a process-wide singleton inheriting the Archon server's `process.env` (`packages/providers/src/community/opencode/runtime.ts`), so per-run env delivery cannot reach it even by accident.
3. **Env-var drift vs Pi upstream.** Archon delivers `HUGGINGFACE_API_KEY`; pi-ai 0.76.0 reads `HF_TOKEN`. It works only because `PiProvider` bridges its own duplicate map via `authStorage.setRuntimeApiKey()` (`pi/provider.ts:286-291`). The map is hand-duplicated in two packages (`delivery.ts:66-74`, `pi/provider.ts:129-139`) and has already drifted.
[...]

PR #1969, opened 2026-06-11
feat(credentials): un-gate ChatGPT/Codex subscription login with an Archon-owned PKCE flow

### LINKED ISSUE #1924: Codex subscription (ChatGPT OAuth) login broken — Pi drops the OpenAI id_token
Part of #1891. Found during the PR #1921 VPS live-smoke.

## Symptom
Connecting a ChatGPT/Codex **subscription** via `archon ai login codex` (or the `/api/auth/providers/codex/oauth/start|poll` route) succeeds and stores an encrypted oauth credential, but any Codex run using it crashes:

```
Codex Exec exited with code 1: invalid ID token format at line 1 column 46
```
(followed by `401 Unauthorized: Missing bearer` on retry)

## Root cause
The Codex CLI requires a valid `id_token` (JWT) in `CODEX_HOME/auth.json`. Our delivery (`buildCodexAuthJson`) emits `id_token: ""` because Pi's `openaiCodexOAuthProvider` never surfaces one:

`@earendil-works/pi-ai/dist/utils/oauth/openai-codex.js` `exchangeAuthorizationCode()` returns only `{ access, refresh, expires }` and **drops `json.id_token`** from the token response — even though the authorize request sends `scope=openid` + `id_token_add_organizations=true`, so OpenAI *does* return an id_token. Archon only sees Pi's `login()` return value, so it can never obtain the id_token.

(The rest of the delivery is correct and verified live: `account_id` is populated from `accountId`, `OPENAI_API_KEY: null`, access/refresh present.)

## Current mitigation (this issue's PR)
`codex` is excluded from `SUBSCRIPTION_PROVIDERS`, so `ai login codex` and the OAuth route refuse with a clear message. The delivery code (`buildCodexAuthJson`, `PI_BACKEND_ID`) is left intact for re-enable. **API-key codex is unaffected** (global `OPENAI_API_KEY` or `archon ai key set codex`).

## Fix options
1. **Pi upstream**: have `openaiCodexOAuthProvider` keep `id_token` in its `OAuthCredentials`. Then delivery just maps it through. (Pi-contribution gate applies — issue → lgtm → PR, never auto-submit.)
2. **Archon-direct exchange**: reimplement the codex OAuth token exchange in Archon so we capture the full response (incl. id_token). Non-trivial — we'd need the PKCE verifier, which Pi currently owns.
3. **Paste-native-auth.json connect (recommended near-term)**: a manual connect path/UI where the user pastes their own `~/.codex/auth.json` from a local `codex login` (which contains a real id_token). Archon encrypts + delivers it. No Pi dependency. Doubles as general "manual subscription config" UX.

## Acceptance
A Codex run using a connected ChatGPT subscription completes (no "invalid ID token format"), verified on the VPS smoke (`e2e-codex-smoke`).

Wirasm:
## Summary

- Problem: the ChatGPT/Codex subscription login has been gated since #1924 — pi-ai (verified through 0.79.1, `openai-codex.js` `readTokenResponse`) drops the OpenAI `id_token` from the token exchange, while the Codex CLI requires a real `id_token` JWT in `CODEX_HOME/auth.json` and crashes with `invalid ID token format` on an empty one.
- Why it matters: ChatGPT Plus/Pro users on multi-user installs could only connect an `openai` API key — the subscription they already pay for was unusable for Codex runs.
- What changed: Archon now OWNS the openai OAuth flow end-to-end (`packages/core/src/credentials/openai-oauth.ts`): PKCE (S256) against `auth.openai.com` with the Codex CLI's public client, manual-paste UX through the existing bridge session machinery, a direct refresh path that preserves `id_token` across rotations, and `buildCodexAuthJson` emitting the real `id_token`. `SUBSCRIPTION_PROVIDERS` becomes `{anthropic, openai, github-copilot}`; every surface that documented the gate (CLI strings, docs-web, CLAUDE.md, console comments, tests) is updated.
- What did **not** change (scope boundary): anthropic/github-copilot keep their Pi-driven flows; the API-key openai path is untouched; the bridge start/poll API shape is unchanged (openai is just another `mode: 'manual'` provider); no DB/schema changes — the oauth blob is opaque TEXT and simply carries one more field.

### Before

```
  archon ai login codex ─────▶  "does not support subscription login" (gated, #1924)
  (or console Login button absent: subscriptionAvailable=false)
  only option: ai key set openai (API key billing, no subscription)
```

### After

```
  User                          Archon bridge                 auth.openai.com
  ────                          ─────────────                 ───────────────
  archon ai login codex ─────▶  [Archon PKCE flow builds
  (or console Login button)      authorize URL] ── returns URL
  opens URL, authorizes ───────────────────────────────────▶  redirects to localhost:1455
  (nothing listens there — deliberate, avoids #1963)
  pastes redirect URL/code ──▶  poll(code) → state check →
                                [direct exchange: captures    ◀─ access+refresh+expires
                                 id_token + accountId] ─────▶    **+ id_token**
  "Connected" ◀───────────────  persists encrypted blob
  Codex run ──────────────────▶ auth.json with REAL id_token → Codex CLI accepts it
```

### Before

```
  oauth-bridge ──▶ oauth-providers (SUBSCRIPTION_DISABLED gates openai) ──▶ pi-ai login()
  user-provider-key-store ──▶ pi getOAuthApiKey (all oauth vendors)
  delivery.buildCodexAuthJson ──▶ id_token: '' (unreachable from connect)
```

### After

```
  oauth-bridge ──▶ oauth-providers (SUBSCRIPTION_PROVIDERS = anthropic|openai|github-copilot)
       │                ├── anthropic / github-copilot ──▶ pi-ai login()  (unchanged)
       │                └── openai ═══▶ [+] openai-oauth.ts (Archon PKCE: authorize URL,
       │                                  paste parsing, exchange w/ id_token, refresh)
  user-provider-key-store ──▶ pi getOAuthApiKey (anthropic/copilot)
       └── openai ═══▶ [+] mintOpenAiOAuthApiKey (direct refresh, id_token preserved;
                            shared rotation→resave-with-retry + inflight coalescing)
  delivery.buildCodexAuthJson [~] ──▶ id_token: real JWT
```

| `oauth-bridge.ts` | `openai-oauth.ts` | **new** | `runOpenAiManualLogin` (authorize flow + exchange) via the existing session/codeDeferred machinery |
| `oauth-bridge.ts` | `@archon/providers/oauth` | **modified** | only anthropic/github-copilot route to Pi now |
| `user-provider-key-store.ts` | `openai-oauth.ts` | **new** | `mintOpenAiOAuthApiKey` for vendor `openai` |
| `user-provider-key-store.ts` | `@archon/providers/oauth` (`getOAuthApiKey`) | **modified** | no longer used for `openai` |
| `oauth-providers.ts` | `@archon/providers/oauth` | **modified** | `openaiCodexOAuthProvider` import removed; `openai` absent from `ARCHON_TO_PI_OAUTH` by design |
| `delivery.ts` (`buildCodexAuthJson`) | stored blob | **modified** | consumes the real `id_token` (comments updated) |
| `delivery.ts` (`buildPiAuthJson`) | Pi `auth.json` | unchanged | extra `id_token` field tolerated — Pi's `OAuthCredentials` is index-signature open (`[key: string]: unknown`), and its refresh/getApiKey read only access/refresh/expires/accountId |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `core|cli|server|web|docs`
- Module: `core:credentials`

## Change Metadata

- Change type: `feature`
- Primary scope: `core`

## Linked Issue

- Closes #1924
- Related #1954, #1891, #1963
- Depends on #1968 (stacked on `fix/oauth-bridge-port-wedge` — the bridge cancel/settle machinery this flow reuses)

Commands and result summary:

```bash
bun run validate   # green (check:bundled*, check:pi-vendor-map, type-check, lint,
                   #        format:check, per-package tests — all pass)
```

- Evidence provided (test/log/trace/screenshot): 15 new unit tests in `openai-oauth.test.ts` (PKCE/URL construction incl. challenge derivation, paste-input parsing, exchange→credential mapping with mocked fetch, missing-id_token/missing-account-claim fail-loud, refresh preserving id_token, mint-with-expiry semantics); 5 bridge tests for the openai manual flow (start/poll/exchange, legacy `codex` id, state mismatch, exchange failure, abort); 4 store tests for the direct refresh path (routes around Pi, legacy-row normalization, rotation resave keeps id_token, refresh failure → null); delivery tests updated for the real id_token + a legacy-Pi-blob case. `catalog.test.ts` was also wired into the core test script — it existed but was never executed by CI.
- If any command is intentionally skipped, explain why: LIVE verification (a real ChatGPT login + `e2e-codex-smoke` per the issue's acceptance criteria) happens in a post-merge VPS round with real credentials — it cannot run in CI and is deliberately not claimed here.

## Security Impact (required)

- New permissions/capabilities? (`No`) — same credential class the gated path already stored; PKCE public client, no client secret.
- New external network calls? (`Yes`) — `POST https://auth.openai.com/oauth/token` directly from Archon (exchange + refresh). Previously the same endpoint was called by pi-ai inside the same process; the call site moved into Archon-owned code. No other hosts.
- Secrets/tokens handling changed? (`Yes`) — the stored openai blob now also carries `id_token`. Same AES-256-GCM encrypted column, never logged, never returned by any endpoint; bridge error details still pass `sanitizeCredentials`.
- File system access scope changed? (`No`) — auth.json delivery path (per-run `CODEX_HOME` under artifactsDir) unchanged.
- If any `Yes`, describe risk and mitigation: see above — both changes keep the existing encryption/scrubbing posture; the exchange/refresh bodies contain the code/refresh token exactly as the previous pi-ai implementation did.

## Compatibility / Migration

- Backward compatible? (`Yes`) — pre-existing Pi-minted openai blobs (no `id_token`) still deliver; a Codex run with one fails with the known `invalid ID token format` error and reconnecting mints a complete blob (documented in `buildCodexAuthJson`). No rows are rewritten.
- Config/env changes? (`No`)
- If yes, exact upgrade steps: n/a (users with a previously stored, gated-era codex oauth row should `archon ai login codex` once to refresh the blob).

- Verified scenarios: read pi-ai 0.79.1 `openai-codex.js` line-by-line to confirm `readTokenResponse` drops `id_token` and to mirror its authorize-URL construction (client id, scope, redirect URI, `id_token_add_organizations`, simplified-flow + originator params — the exact construction that authenticated successfully on the #1924 live smoke) and its `parseAuthorizationInput` leniency and JWT claim path. Confirmed Pi's `OAuthCredentials`/AuthStorage shape tolerates the extra `id_token` field (index-signature open type; refresh/getApiKey read only the four known fields). Full suite locally.
- Edge cases checked: refresh response omitting `refresh_token`/`id_token` (preserved from stored blob); access token without the `chatgpt_account_id` claim (fail loud at exchange, fallback-to-stored on refresh); legacy `codex` ids at every entry point; expired vs unexpired mint (no network call when fresh).
- What was not verified: the LIVE end-to-end (real ChatGPT account → authorize → paste → Codex run) — post-merge VPS round with Rasmus's credentials, per the issue's acceptance criteria. The `originator=pi` + `codex_cli_simplified_flow` param choice is mirrored from the verified-working Pi construction but the full Archon round-trip hasn't touched the real IdP yet.

- Affected subsystems/workflows: subscription connect (web route + CLI + console Login button, which un-gates automatically via the server-evaluated `subscriptionAvailable`), openai oauth refresh-on-read, Codex auth.json delivery.
- Potential unintended effects: **mid-run rotation risk** — Codex runs get a per-run `CODEX_HOME` scratch copy of auth.json; if the Codex CLI itself rotates the refresh token mid-run inside that scratch dir, the DB-stored refresh token can be orphaned (OpenAI may invalidate the old one). Mitigated by refresh-before-delivery (every run starts with a fresh, just-rotated-and-resaved credential, shrinking the window) and by the existing resave-with-retry + ERROR logging; recovery is a reconnect. This is a known accepted risk documented here rather than silently ignored.
- Guardrails/monitoring for early detection: `user_provider_key.oauth_refresh_failed` / `oauth_resave_failed` ERROR logs; the bridge surfaces exchange failures verbatim (sanitized) through poll.

## Rollback Plan (required)

- Fast rollback command/path: `git revert` of this commit restores the gate (`SUBSCRIPTION_DISABLED`) semantics via the previous `oauth-providers.ts`; stored blobs remain valid for the API surface (opaque), and anthropic/copilot are untouched.
- Feature flags or config toggles (if any): none — the gate set itself is the switch.
- Observable failure symptoms: codex runs failing `invalid ID token format` (bad/legacy blob), `oauth_refresh_failed` for provider openai, 4xx from `auth.openai.com` surfaced in poll `detail`.

## Risks and Mitigations

- Risk: the authorize-URL parameters or token-endpoint behavior differ subtly from what the live IdP expects (only the Pi construction, not Archon's reimplementation, has touched the real endpoint).
  - Mitigation: byte-for-byte mirror of the verified-working construction; unit tests pin every parameter; post-merge VPS live smoke before this reaches users; rollback is a one-commit revert to the gated state.
- Risk: refresh-token rotation orphaning (see blast radius).
  - Mitigation: refresh-before-delivery, resave-with-retry at ERROR level, documented reconnect recovery.
- Risk: drift between Archon's flow and future Codex CLI client changes (client id, params).
  - Mitigation: constants centralized in one module with tests; upstream issue for Pi to surface `id_token` (earendil-works/pi#5649 covers the signal/port half; the id_token half is documented in #1924) would let us retire the custom flow later.


Wirasm:
## Multi-Agent Review Summary — PR #1969

Five review agents (code-reviewer with a security focus, silent-failure-hunter, test-analyzer, type-design, docs-impact) ran against this PR's own delta over the `fix/oauth-bridge-port-wedge` base. Findings labeled I/S (important/suggestion) — not issue refs.

### Security Assessment — clean

The code-reviewer scrutinized the PKCE flow hard and found nothing: S256 challenge derivation matches RFC 7636 exactly (verified, and the test suite *recomputes* the SHA-256 rather than snapshotting); verifier entropy is 256 bits; `state` is 128-bit and validated on paste (bare-code paste skipping the check is intentional and documented for the manual-paste UX); zero token/verifier/code logging anywhere (`openai-oauth.ts` has no log calls at all); the unverified JWT decode for `accountId` is intentional and pass-through (Codex CLI does its own validation); encrypted-at-rest storage untouched; the abort path threads `OAuthLoginAbortedError` correctly through the bridge. The id_token-preservation-on-refresh — the core #1924 fix — is correct and pinned by a precisely-targeted regression test.

### Critical Issues

None.

### Important Issues (4 found)

| I1 | silent-failure-hunter | **Exchange has no timeout when the session abort signal is passed.** `postTokenRequest` only applies `AbortSignal.timeout(30_000)` when `signal` is undefined — but the bridge always passes `session.abort.signal`, so a hung `auth.openai.com` leaves the user polling `pending` for the full 10-minute TTL with no explanation (the refresh path, which passes no signal, gets the 30s ceiling — inconsistent). Fix: `AbortSignal.any([signal, AbortSignal.timeout(30_000)])`. | `openai-oauth.ts:155-182` |
| I2 | silent-failure-hunter | **HTTP 200 with a non-JSON body surfaces as a raw `SyntaxError`.** `await response.json()` is unguarded; a proxy/maintenance HTML page reaches the user as `detail: "Unexpected token '<'…"` — indistinguishable from an Archon bug. Wrap it and throw a labeled "non-JSON response (HTTP nnn)" error. | `openai-oauth.ts:182` |
| I3 | code-reviewer (87) | **Stale comment references a symbol this PR deletes**: `generate-pi-vendor-map.ts`'s `SUBSCRIPTION_VENDORS` doc still points to `SUBSCRIPTION_DISABLED` ("wired but NOT connectable") — backwards from the post-PR reality where all three vendors are connectable. | `scripts/generate-pi-vendor-map.ts:62-68` |
| I4 | docs-impact | **One missed doc spot**: CLAUDE.md's `@archon/providers/oauth` description (Architecture Layers, ~line 488) still says Pi's oauth subpath handles "ChatGPT/Codex" subscription login — that flow is now Archon-owned in `openai-oauth.ts`. All other doc surfaces (ai-assistants.md, api.md, cli.md, CLAUDE.md's CLI + endpoint sections) were verified correctly updated. | `CLAUDE.md` ~488 |

### Suggestions

| S1 | pr-test-analyzer | Two tests worth adding pre-merge: a non-2xx **refresh** error (only the exchange path's error format is tested; an expired refresh token's 401 message is unpinned — 7/10), and the bridge's `Missing authorization code.` branch (the `parseOpenAiAuthorizationInput` stub is hardcoded to always return a code, so a state-only URL paste is untestable today — needs a configurable stub, 6/10). Lower priority: same-user openai supersession (5/10), `mintOpenAiOAuthApiKey` null-return (4/10). | `openai-oauth.test.ts`, `oauth-bridge.test.ts` |
| S2 | silent-failure-hunter | `sanitizeCredentials()` only strips GitHub-shaped secrets; an OpenAI error body (which can carry account identifiers) flows into `session.detail` and to the browser verbatim (capped at 200 chars). Consider stripping the raw body down to status + `error` field for openai exchange failures. | `oauth-bridge.ts` settled `.catch()` |
| S3 | silent-failure-hunter | The refresh path's 30s timeout reads as "The operation was aborted" — check `error.name === 'TimeoutError'` and say "timed out". | `openai-oauth.ts:168-174` |
| S4 | type-design | Three low-priority tightenings: `mintOpenAiOAuthApiKey` should accept `OpenAiOAuthCredentials` (not the wide blob) so the compiler checks routing; `runOpenAiManualLogin`'s `as PiOAuthCredentials` return is a white lie — type `loginPromise` as the common `OAuthCredentials` and let the function return its true type; and `buildCodexAuthJson` could throw an Archon-level "please reconnect the ChatGPT subscription" instead of emitting `id_token: ''` for a legacy blob (zero such rows exist today — the gate was never open — so follow-up is fine). Otherwise the type work rated well: `OpenAiAuthorizeFlow` 9/10, the `OpenAiTokenResponse` all-`unknown`-fields boundary pattern called out as exemplary. | `openai-oauth.ts`, `oauth-bridge.ts:196-215`, `delivery.ts` |

### Merge-Time Awareness

This PR stacks on #1968; when it lands on dev, take the PR branch's version for the predictable conflicts in `oauth-providers.ts` (`SUBSCRIPTION_DISABLED` deletion), `oauth-bridge.ts` (gate comment), and the three docs files (`ai-assistants.md`, `api.md`, `cli.md`) — dev still carries the pre-gate-lift wording in all five.

### Strengths

- Fail-loud discipline at every credential boundary: missing `id_token` at exchange throws with the exact downstream Codex error named; missing `access_token`/non-finite `expires_in` throw with context; empty stored refresh token detected before any network call; failed `accountId` extraction throws rather than storing an empty id
- The refresh preserves `id_token`/`refresh_token` when the rotation response omits them — the precise #1924 failure class — and the store test inspects the re-encrypted resave
- Routing bifurcation is explicitly tested: `openai` rows (including legacy `codex` normalization) hit Archon's mint path and provably never reach Pi's `getOAuthApiKey`
- Old gate assertions were removed, not left dead: `catalog.test.ts` flips `subscriptionAvailable` to true, the bridge's rejects-on-codex test is gone
- `oauth_resave_failed` correctly logged at ERROR with the "stored refresh token may now be dead" rationale

### Verdict

**NEEDS FIXES** (small scope) — the security core is solid and the #1924 fix is real and well-tested; I1/I2 are cheap error-UX hardening on the network paths (one `AbortSignal.any`, one try/catch), and I3/I4 are two comment/doc lines. None of it touches the flow's logic.

### Recommended Actions

1. Add the exchange timeout ceiling (I1) and the non-JSON guard (I2)
2. Fix the `SUBSCRIPTION_VENDORS` comment (I3) and the CLAUDE.md `@archon/providers/oauth` line (I4)
3. Add the refresh-401 and missing-code tests (S1); consider S2–S4 as ride-along or follow-up

Wirasm:
## Review disposition — applied with one explicit deferral (commit 91b2847, rebased on the updated base branch)

| Finding | Disposition |
|---|---|
| I1 | **Applied.** `postTokenRequest` now applies the 30s ceiling unconditionally — `AbortSignal.any([sessionSignal, AbortSignal.timeout(30_000)])` when the bridge passes its signal — so a hung `auth.openai.com` surfaces within 30s instead of leaving the login `pending` for the session's 10-minute TTL. Exchange and refresh are now consistent. |
| I2 | **Applied.** `response.json()` is guarded: an HTTP 200 with a non-JSON body throws a labeled `non-JSON response (HTTP nnn)` error. Test added. |
| I3 | **Applied.** The `SUBSCRIPTION_VENDORS` comment in `generate-pi-vendor-map.ts` now states the post-lift reality (all three connectable; openai via the Archon PKCE flow) and names `SUBSCRIPTION_PROVIDERS` as the runtime source of truth. The comment lives in the script only — generated output unchanged; `check:pi-vendor-map` verified green. |
| I4 | **Applied.** CLAUDE.md's `@archon/providers/oauth` description corrected: the Pi subpath drives Claude Pro/Max + Copilot; ChatGPT/Codex subscription login is Archon-owned in `credentials/openai-oauth.ts`. |
| S1 | **Applied (all four).** Non-2xx refresh error format (401 invalid_grant); the bridge's `Missing authorization code.` branch (the parse stub is now configurable, reset per test); same-user openai supersession; `mintOpenAiOAuthApiKey` null-return on a legacy/corrupt row. |
| S2 | **Applied at the source.** Non-ok token responses are stripped to status + the OAuth `error` code (string or `error.code`) inside `postTokenRequest` — the raw body never reaches `session.detail` or the browser. A leak test asserts an account identifier in the error body does not survive into the message. |
| S3 | **Applied.** `TimeoutError` name check → `OpenAI token <op> request timed out.`; user-cancel keeps `Login cancelled`. |
| S4 (types) | **Applied, two of three.** `mintOpenAiOAuthApiKey` accepts the narrow `OpenAiOAuthCredentials` (store call site asserts with a write-time-shape comment; runtime guards kept because decrypted rows are not compile-time guaranteed). `runOpenAiManualLogin` returns its true type — the `as PiOAuthCredentials` cast is gone; structural assignability covers the `loginPromise` join. |
| S4 (buildCodexAuthJson throw) | **Deferred, deliberately.** Zero legacy blobs can exist — the gate was never open in any release, so no Pi-minted openai oauth row was ever persisted; the function doc already names the exact downstream Codex error and the reconnect remedy; and a throw would convert a cosmetic impossibility into a new failure path in the delivery seam (which today never throws for a connected vendor). Reviewer noted follow-up is fine — happy to revisit if a legacy row ever materializes in the wild. |

`bun run validate` green on the branch (22 bridge tests, 19 openai-oauth tests, 25 store tests). Branch rebased onto the base branch's review-fix commit — stack intact.

PR #1987, opened 2026-06-12
fix(adapters): drop duplicate webhook deliveries at ingest

### LINKED ISSUE #1951: Duplicate webhook delivery runs a workflow twice (no ingest idempotency)
Summary

  - What broke: A single GitHub trigger can execute a workflow twice. A duplicate webhook delivery is processed without any idempotency check —
  ConversationLockManager.acquireLock queues the duplicate behind the in-flight run (per-conversation ordering, but no dedup) and replays it as
  a second, byte-identical full run when the first finishes.
  - When it started: Not a regression — structural in the conversation-lock dispatch path; present on current dev (verified 2026-06-10).
  - Severity: major

  Steps to Reproduce

  1. Arrange for the same event to reach /webhooks/github twice — e.g. two active subscriptions to the same URL (a repo webhook + an installed
  GitHub App webhook), an LB/proxy double-forward, or a genuine GitHub at-least-once redelivery.
  2. Post one @archon <command> comment on an issue.
  3. Watch: run #1 executes fully; ~as it completes, run #2 starts — same conversation, byte-identical user_message.

  Expected vs Actual

  - Expected: A duplicate delivery of an already-processed trigger is recognized and dropped — one logical trigger ⇒ one workflow run.
  - Actual: Each delivery is treated as a distinct message; the second is queued behind the first and replayed as a second full run (re-runs
  every node, full token cost, can open a second PR / push again).

  User Flow

  (Where it breaks marked [X].)

    GitHub                  Archon (handleWebhook)          ConversationLock
    ──────                  ──────────────────────          ────────────────
    event delivered ──────▶ delivery A: guards pass ──────▶ acquireLock: free
                                                            → run #1 starts ─────▶ runs to completion
    same event again ─────▶ delivery B: guards pass ──────▶ [X] activeConversations.has(id)
                                                                → queueMessage()  (NO dedup)
                            run #1 done → .finally ────────▶ processQueue()
                                                            → run #2 (duplicate) starts ▶ redoes all work

  Environment

  - Platform: GitHub (mechanism is platform-agnostic — any adapter routing through handleMessage + the conversation lock)
  - Database: SQLite (DB-independent)
  - Running in worktree? Yes (archon-fix-github-issue uses worktrees) — but irrelevant to the bug
  - OS: Linux (Docker container, Bun runtime)

  Logs

  # Two ingress events for ONE logical trigger (the only two webhook_processing in the window):
  github.webhook_processing  eventType=issue_comment  number=<N>   # delivery A
  github.webhook_processing  eventType=issue_comment  number=<N>   # delivery B (ms later)

  # Two workflow_runs rows — identical conversation + user_message; run #2 does nothing:
  remote_agent_workflow_runs:
    run#1  id=<a>  conversation_id=<c>  parent_conversation_id=<c>  user_message="Fix #<N>: …"  status=completed
  metadata={node_counts:{completed:18,…}, total_cost_usd:4.96}
    run#2  id=<b>  conversation_id=<c>  parent_conversation_id=<c>  user_message="<byte-identical>"  status=cancelled  metadata={}

  # The queueing itself is invisible at info level:
  #   conversation-lock 'queued-conversation' / 'message_queued' are logged at DEBUG.
  # X-GitHub-Delivery is read at the webhook route but never threaded downstream (only error logs),
  #   so there is no stable key to dedup on and no way to tell redelivery from double-delivery.

  Impact

  - Affected workflows/commands: any workflow triggered via the GitHub adapter (observed: archon-fix-github-issue); broadly, anything funneling
  through handleMessage + the conversation lock.
  - Reproduction rate: Always, given a duplicate delivery (deterministic). The duplicate delivery itself is environment-dependent — a duplicate
  subscription = 100% doubling; a GitHub redelivery = intermittent.
  - Workaround available? Remove the duplicate delivery source (we disabled a redundant GitHub App webhook → confirmed 1× delivery). This does
  not close the class — GitHub's at-least-once contract means a redelivery will recur eventually.
  - Data loss risk? No (but duplicate side effects: a second PR/push is possible, plus wasted compute).

  Scope

  - Package(s) likely involved: core (the dedup/decision), adapters (thread the delivery id downstream)
  - Module (if known): core:utils/conversation-lock (ConversationLockManager.acquireLock); adapters:forge/github (thread X-GitHub-Delivery to
  the lock as the idempotency key)

truffle-dev:
## Summary

Fixes #1951 — @sbiitmc's diagnosis is exactly right: a duplicate delivery of the same comment passes every guard in `handleWebhook`, reaches `ConversationLockManager.acquireLock`, and gets queued behind the in-flight run as `queued-conversation`. The lock manager orders per-conversation messages but never dedups them, so when run #1 completes, `.finally → processQueue` replays the byte-identical message as a full second workflow run (full token cost, can push/open a second PR).

The incident's duplicate source — dual repo-webhook + App-webhook subscriptions — delivers the same comment under **different** delivery GUIDs, so deduping on `X-GitHub-Delivery` alone would miss it. The trigger path always carries a comment (close events return early in `parseEvent`), so the fix keys on the comment's identity instead: `id` + `updated_at`. Dual-subscription duplicates and LB double-forwards share that identity and get dropped; an edited comment gets a new `updated_at` and still re-triggers.

## Changes

- New `DeliveryDeduplicator` in `packages/core/src/utils/delivery-dedup.ts` (exported from core, reusable by the gitea/gitlab adapters later): a bounded first-seen cache — 10-minute TTL so deliberate manual redeliveries hours later still run, 10k-entry cap with oldest-first eviction so memory stays bounded.
- GitHub adapter gates processing on `comment:{owner}/{repo}#{number}:{comment.id}:{comment.updated_at}` after the @mention check and **before** the expensive path (user resolution, conversation/codebase creation, clone/sync, comment-history fetch). Drops log as `github.duplicate_delivery_dropped` at info.
- Falls back to `delivery:{deliveryId}` when the payload lacks comment identity, and fails open when neither is available — never drops a webhook for want of a key. The dedup check sits after signature verification so unauthenticated junk can't poison the cache.
- `X-GitHub-Delivery` was already extracted in the webhook route but only used in error logs; it's now threaded to `handleWebhook` as an optional third parameter (per the issue's scope note). Gitea/gitlab adapters untouched.
- `WebhookEvent.comment` type gains optional `id`/`updated_at` (both present on real GitHub deliveries).

## Tests

- `delivery-dedup.test.ts` (new, wired into core's test chain): first-seen/repeat, key independence, edit-forms-new-key, TTL expiry, prune-on-insert, max-size eviction, post-expiry refresh — 8 cases.
- `adapter.test.ts`, new `webhook delivery dedup` block: same-GUID repeat dropped, **dual-subscription duplicate (same comment, different GUIDs) dropped** — the #1951 incident shape, edited comment re-processed, distinct comments independent, GUID fallback, fail-open with no key — 6 cases asserting on conversation-creation call counts.
- Pre-fix audit: no existing dedup/idempotency mechanism anywhere in core/adapters/server (grepped `idempoten|dedup|deliveryId|x-github-delivery`), and no test pinned the duplicate-processing behavior.

## Validation

`bun run lint --max-warnings 0`, `format:check` pass repo-wide. Full adapters suite green (70-test github adapter lane + all chained lanes). Core chain green through the utils lanes including the new one (the long-tail orchestrator/credentials lanes are untouched by this diff and left to CI). `tsc --noEmit` clean on core/adapters/server apart from pre-existing `packages/providers` copilot-client drift unrelated to this change.

Closes #1951

## Summary by CodeRabbit

* **New Features**
  * Added webhook delivery deduplication for GitHub issue comment events using comment identity plus edit time, with fallback to the delivery GUID for edge cases.
  * Allows edited comments to re-trigger processing; suppresses exact/related duplicate deliveries.
* **Bug Fixes**
  * Prevents duplicate downstream handling when GitHub retries webhooks with the same logical event.
* **Tests**
  * Added coverage for TTL behavior, eviction, and edit re-triggering, plus webhook-level dedup scenarios.
* **Chores**
  * Exposed the deduplication utility via the public core API and broadened the automated test command.


dqikfox (approved):

dqikfox (approved):

dqikfox (approved):

dqikfox (approved):

dqikfox (approved):

dqikfox (approved):

truffle-dev:
Rebased onto current `dev`. The only conflict was the `packages/core` `test` script — `dev` had restructured it (token-crypto split out, `post-message-reminder.test.ts` added), so I kept that structure and grafted `src/utils/delivery-dedup.test.ts` back into the list rather than reverting it. `delivery-dedup` (9) and the github adapter suite (71) pass locally. The provider type-check errors in CI are pre-existing on `dev` (`copilot/provider.ts` `CopilotClientOptions`), unrelated to this change.

Wirasm:
## Review Summary

The dedup design is sound, event names follow project conventions, and the new `DeliveryDeduplicator` is a clean reusable utility. The main blocker is one untested production seam (the server route that forwards `X-GitHub-Delivery`), and there's a real test-reliability issue where your new adapter tests make live GitHub API requests while swallowing every error. Most of the rest is comment-tightening and nice-to-have docs polish.

### Blocking issues
None — no CRITICAL findings.

### Suggested fixes (HIGH)
- **`packages/server/src/index.ts:742`** — add a server-route test that POSTs a raw webhook with `x-hub-signature-256` and `x-github-delivery: guid-1` against a mocked `GitHubAdapter`, and asserts `handleWebhook(rawPayload, signature, 'guid-1')` is called. Also cover the omitted-header case and assert the third argument is `undefined`. The adapter tests pass `deliveryId` directly, so they would stay green if the header-forwarding seam silently broke.

### Suggested fixes (MEDIUM)
- **`packages/adapters/src/forge/github/adapter.test.ts:521`** — your `deliver()` helper catches every error while first-seen deliveries continue into the PAT adapter's unmocked Octokit path. The targeted lane made live `GET /repos/testuser/testrepo` and `POST .../issues/42/comments` requests with the fake token. Inject/stub the minimal Octokit methods so `handleWebhook()` completes locally, then remove the broad catch.
- **`packages/adapters/src/forge/github/adapter.ts:1007-1026`** — optional: the dedup key is claimed before the failure-prone downstream work. If a first attempt fails transiently, any redelivery within TTL is dropped. The inline comment documents this as an accepted fire-and-forget tradeoff, but consider a regression test where the first attempt rejects and a subsequent delivery is allowed to process, or move the claim to just before `acquireLock()` after setup succeeds.
- **`packages/adapters/src/forge/github/adapter.ts`** — the 12-line block before `seen()` and the 8-line `// 5a. Ingest idempotency.` block both exceed the project's "no multi-line blocks unless absolutely necessary" bar. Condense each to 3-4 lines; the "dual subscriptions deliver the same comment under different GUIDs" point is the only load-bearing WHY.
- **`packages/core/src/utils/delivery-dedup.ts` (file JSDoc)** — second paragraph and trailing sentence restate the method and constant names. Keep the first paragraph (the dual-subscription WHY), delete the rest.
- **Task-id references (`#1951`)** in five source locations — they will rot. Keep the link on the PR/issue side; in source, the "dual subscriptions" WHY is the durable rationale.

### Minor / nice-to-have (LOW)
- **`packages/core/src/utils/delivery-dedup.test.ts:89`** — the "refreshes its eviction position" test never creates size pressure. Use `maxEntries = 2`, insert `a`, advance within TTL and insert `b`, advance until only `a` is expired, re-see `a`, then insert `c` and assert refreshed `a` remains while older `b` is evicted.
- **Test helpers missing return types** (`createClock()`, `deliver()`) — CLAUDE.md requires complete type annotations on functions. Define a `TestClock` type and declare `createClock(start?: number): TestClock` and `deliver(...): Promise<void>`.
- **`adapter.test.ts`** — assert that `github.duplicate_delivery_dropped` actually fires on dedup. The dedup path's only observable side effect (besides the call-count check) is currently unverified.
- **`packages/core/src/utils/delivery-dedup.ts:98-105`** — `prune()`'s rationale depends on `seen()` short-circuiting on a fresh entry. One-line comment to that effect would prevent a future refactor from silently breaking the invariant.
- **`delivery-dedup.ts`** — JSDocs on `DEFAULT_TTL_MS`, `DEFAULT_MAX_ENTRIES`, and `private entries` all restate the name/type. Delete.
- **`packages/adapters/src/forge/github/types.ts`** — `comment.id` and `comment.updated_at` JSDocs tie the type to the dedup feature. Tighten to neutral descriptions so the type isn't coupled to one consumer.
- **`adapter.test.ts`** — two trailing in-test comments and the JSDoc on `createIdentifiedCommentPayload` restate names. Delete.
- **Docs polish** (optional):
  - `packages/docs-web/src/content/docs/reference/architecture.md:1272` — flow diagram shows old two-arg `handleWebhook` signature; the third `deliveryId` arg is optional but worth updating for accuracy.
  - `packages/docs-web/src/content/docs/adapters/github-app-setup.md` (PAT→App "Migration from PAT mode") — a one-line note that the bot will dedupe during the transition would reassure operators running dual webhooks.

### Compliments
The WHY-focused comments on `prune()`, the `delete`-before-`set` ordering in `seen()`, the `size` getter's "includes not-yet-pruned expired entries" gotcha, the `cachedLog` lazy-init pattern, and the architectural rationale on the `deliveryDedup` field are exactly the kind of comments this codebase needs — keep all of them. The deduplication strategy itself (key on comment identity, GUID fallback, fail-open when neither is available) is well thought out and the gitea/gitlab adapters will be able to reuse `DeliveryDeduplicator` cleanly.

---

PR #2032, opened 2026-06-26
feat(workflows): loop_group node — cross-node subgraph looping

xbeark:
## Summary

- **Problem:** The workflow DAG engine only supports intra-node looping (`loop:` re-invokes its own single prompt per iteration). There is no construct that repeats a *subgraph* of multiple nodes (e.g. `implement → test → review`) until a condition is met.
- **Why it matters:** Authors who need "iterate a whole pipeline until done" today must either collapse N steps into one prompt (awkward, loses per-step structure) or write a long linear chain with no iteration (loses "until done"). Cross-node iteration is a common ask for agentic refinement loops.
- **What changed:** Adds a new `loop_group` DAG node variant that carries a `nodes:` sub-DAG body and repeats the whole body per iteration until a completion condition (`until` signal and/or `until_bash` exit code) or `max_iterations`. Mirrors the proven `executeLoopNode` pattern at subgraph granularity. The layer-execution loop is factored into a reusable `runLayers(...)` helper (pure refactor, behavior-preserving). New cross-iteration ref `$LOOP_PREV.<nodeId>.output`.
- **What did NOT change (scope boundary):** The top-level DAG stays strictly acyclic / single-pass / write-once — the cycle is **encapsulated** inside the `loop_group` node. No native back-edges, no fixpoint scheduler. v1 resume re-runs the whole current iteration's body. `persist_session` across iterations, per-body-node resume granularity, `$nodeId.output[N]` history indexing, and the Web WorkflowBuilder nested-nodes editor are explicitly deferred (see NOT Building in the plan).

### Before

```
AUTHOR wants: implement → test → review  (repeat until tests pass)

Today: must collapse into ONE loop: node prompt (awkward), OR
       write a long linear chain with NO iteration (loses "until done").

  ┌─────────────┐    ┌──────────┐    ┌────────┐
  │ implement   │ →  │ test     │ →  │ review │   (runs ONCE, no loop)
  └─────────────┘    └──────────┘    └────────┘
```

### After

```
AUTHOR writes one loop_group node containing the 3-step body:

  ┌──────────────── loop_group: "fix_until_pass" ────────────────┐
  │  until: TESTS_PASS   max_iterations: 5                       │
  │  ┌──────────┐    ┌────────┐    ┌────────┐                    │
  │  │ implement │ →  │ test   │ →  │ review │  ← repeats whole  │
  │  └──────────┘    └────────┘    └────────┘     body per iter  │
  └───────────────────────────────────────────────────────────────┘
          ↓ (to outer DAG: one node)
   $fix_until_pass.output = final review output

Outer DAG stays acyclic, single-pass. Cycle is encapsulated in the node.
```

### Before

```
executeDagWorkflow
  └─ for layer in buildTopologicalLayers(nodes):   ← computed once, walked once
       └─ Promise.allSettled(layer.map(node => dispatch(node)))   ← each node once
            ├─ bash / loop / approval / cancel / script / AI(prompt|command)
            └─ nodeOutputs.set(nodeId, output)   ← write-once, keyed by node id

loop: node = intra-node only (re-invokes its own prompt; never re-enters scheduler)
```

### After

```
executeDagWorkflow
  └─ runLayers(layers, nodeOutputs, ...)            [~] extracted helper
       └─ Promise.allSettled(layer.map(node => dispatch(node)))
            ├─ bash / loop / approval / cancel / script / AI
            └─ [+] isLoopGroupNode → executeLoopGroupNode
                 └─ for i in 1..max_iterations:       ← encapsulated cycle
                      ├─ scoped nodeOutputs (fresh per iteration)
                      ├─ runLayers(bodyLayers, scopedMap, stepNamePrefix="{group}.")
                      ├─ completion gate: detectCompletionSignal(terminalOutput, until) || until_bash
                      └─ carry prior snapshot → $LOOP_PREV.<id>.output
```

**Connection inventory:**

| `executeDagWorkflow` | `runLayers` | **modified** | layer loop factored out |
| `runLayers` dispatch | `executeLoopGroupNode` | **new** | `isLoopGroupNode` branch |
| `executeLoopGroupNode` | `runLayers` | **new** | per-iteration body execution (recursive) |
| `executeLoopGroupNode` | `detectCompletionSignal` | **new** | reuse from `executor-shared` |
| `executeLoopGroupNode` | `buildTopologicalLayers` | **new** | body layers (computed once) |

## Label snapshot

- Risk: `risk: high` (HIGH complexity; touches the core DAG executor; mitigated by pure-refactor spike first)
- Size: `size: XL`
- Scope: `workflows`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Related: (none filed yet — feasibility study in `.claude/archon/research/2026-06-26-cross-node-looping.md`; this PR implements `.claude/archon/plans/subgraph-loop-group-node.plan.md`)
- Depends on: (none)

All eight `bun run validate` checks pass (exit 0): `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, type-check, lint (`--max-warnings 0`), format check, and the full test suite.

```bash
bun run validate   # ✅ exit 0
```

- **Refactor parity (T1):** the `runLayers` extraction is behavior-preserving — all 287 pre-existing `dag-executor` tests pass unchanged (285 + 2 new loop_group tests).
- **Feature tests (T5):** `loop_group` completes in N iterations when the `until` signal appears on iteration N; fails when `max_iterations` is exceeded without the signal.
- **Loader validation (T3):** rejects cyclic body, rejects body `depends_on` to unknown node, accepts well-formed `loop_group` (3 new tests, 134 loader tests total).
- **Skipped commands:** none.

## Security Impact (required)

- New permissions/capabilities? `No` — new node variant; no new capability surface beyond existing AI/bash node execution already present in workflows.
- File system access scope changed? `No` — body nodes use the same artifactsDir/logDir as the parent run; body node artifacts are namespaced `{groupId}.{nodeId}`.

## Compatibility / Migration

- Backward compatible? `Yes` — purely additive (new node variant). Existing `loop:` and all other node types unchanged. The `runLayers` extraction is a behavior-preserving refactor covered by a golden test.
- Config/env changes? `No`
- Database migration needed? `No` — body node events reuse the existing `remote_agent_workflow_events` table with namespaced `step_name`; no schema change (confirmed: `getCompletedDagNodeOutputs` keys on `step_name` only).

To be completed as tasks land. Planned:

- Verified scenarios: `loop_group` completes in N iterations; `$groupId.output` = final iteration terminal output; `$LOOP_PREV.<id>.output` resolves across iterations; `max_iterations` → failed; `interactive` pause/resume; `until_bash` exit-code gate.
- Edge cases checked: single-node body (degenerates to `loop:`-equivalent); body `depends_on` to outer node id rejected; nested `loop_group` (smoke-test only).
- What was not verified: nested `loop_group` hardening (deferred); Web WorkflowBuilder nested editor (deferred).

- Affected subsystems/workflows: `@archon/workflows` (dag-executor, schemas, loader, validator, executor-shared). `@archon/web` only if the recursive `z.lazy` schema affects OpenAPI type generation (verified in T2/T11).
- Potential unintended effects: the `runLayers` refactor touches the single hottest code path in the workflow engine (every workflow run). Mitigation: it lands first as a pure refactor with a golden parity test, ideally as its own commit/PR-slice before any new node type.
- Guardrails/monitoring for early detection: golden test asserts identical `node_completed` events + outputs before/after refactor; `bun run validate` gates CI.

## Rollback Plan (required)

- Fast rollback command/path: `git revert` the `loop_group` commits; the `runLayers` refactor reverts independently (it's additive-factored, behavior-preserving). Removing the `loop_group` variant is a clean delete — no migration, no persisted state depends on it (body events are namespaced and naturally ignored if the variant is removed).
- Feature flags or config toggles: none — the feature is opt-in by authoring a `loop_group` node; workflows without it are unaffected.
- Observable failure symptoms: workflows with `loop_group` nodes fail at load/validate if reverted; plain workflows unaffected.

## Risks and Mitigations

- **Risk:** `runLayers` extraction introduces subtle behavior drift (session threading, usage accumulation, event ordering) in the engine's hottest path.
  - **Mitigation:** Lands first as a pure refactor with a golden multi-layer parity test, before any new node type. Reviewable as an isolated commit.
- **Risk:** `z.lazy(() => dagNodeSchema)` recursive schema breaks zod-openapi / `@archon/web` type generation.
  - **Mitigation:** Verified in T2 via `bun run type-check` + web `generate:types`; fallback to depth-limited schema if recursive OpenAPI fails.
- **Risk:** Resume iteration-boundary semantics differ from `loop:` (body pauses mid-iteration vs `loop:` post-iteration).
  - **Mitigation:** T7 re-reads `executeLoopNode:2065-2073` carefully, adds an explicit resume test, documents the choice.

---

> 📋 **Implementation plan:** `.claude/archon/plans/subgraph-loop-group-node.plan.md` (11 tasks, HIGH complexity). This draft PR tracks incremental delivery. The plan is a local working artifact (gitignored); the task breakdown is mirrored in the commit history as tasks land.

## Summary by CodeRabbit

* **New Features**
  * Added **loop group** node support to iteratively run a sealed multi-step sub-DAG using `until` / `until_bash` and `max_iterations`, including **nested** loop groups.
  * Enabled **cross-iteration references** with `$LOOP_PREV...`, namespaced body events, and defined loop-group output as the **terminal output from the final iteration**.
  * Added interactive **gate** pause/resume behavior and iteration-aware prompt substitution.
* **Bug Fixes**
  * Improved loop-group workflow discovery/validation and Level 3 resource validation within loop groups.
  * More reliable execution accounting/output selection, including safer handling of non-finite token totals.
* **Documentation**
  * Updated DAG Workflows and added a loop nodes guide section for `loop_group` (examples, limitations, and resume behavior).


xbeark:
Thanks for the thorough review — all addressed in `7e3b5c6f` (on top of a rebase to current `dev`). Summary:

**Fixed**
- **#6 inherit workflow defaults in body** — `executeLoopGroupNode` now forwards `workflowModel`/`workflowLevelOptions`/`aiProfile`/`workflowPreset` to the per-iteration `runLayers` ctx (was hardcoded `undefined`, dropping inherited model aliases for body AI nodes).
- **#4 `fresh_context:false` session threading** — replaced the per-iteration `lastSequentialSessionId` reset with a loop-level cursor that threads the body's final sequential session into the next iteration (reset only when `fresh_context` is true or on iteration 1).
- **#8 shell-escape `$LOOP_PREV` in bash/script/cancel body fields** — `applyLoopPrevToBodyNode` passes `escapedForBash=true` for shell-bound fields.
- **#9 treat skipped/pending prior body output as absent** — `substituteLoopPrevRefs` resolves to `''` when the prior `NodeOutput` is skipped/pending (not just missing).
- **#7 short-circuit `until_bash` when the until-signal already completed** — skips the `until_bash` subprocess when `signalDetected` is true (OR semantics).
- **#10 preserve model/provider AI overrides on `loop_group` parse** — the transform now spreads `aiOnly` so group-level `model`/`provider` survive parsing.
- **#N1 EDGE G** — rewrote to use a sentinel file proving `until_bash` is short-circuited.
- **#N2 resume `$LOOP_PREV`** — added a test locking the v1 behavior that `$LOOP_PREV` is **not** preserved across the pause/resume boundary (resolves to `''` on the resumed iteration).

**Rebase note** — rebased onto current `dev` (which added #2038 resolved-model/tier surfacing + #1842 cold-resume surfacing). The T1 `runLayers` extraction conflicted; resolved by porting the upstream tier + cold-resume wiring into `runLayers` so it stays behavior-identical to dev's inline loop. All 306 dag-executor tests pass.

**Deferred (out of scope for this PR)**
- **#1 namespaced `step_name` for `executeNodeInternal` body events** — 15 event-write call sites; invasive. `runLayers`' own skip/trigger/when events ARE namespaced via `stepNamePrefix`; only `executeNodeInternal`'s internal `node_started`/`node_completed`/`node_failed`/tool events use the raw id. Noted as a known v1 limitation (resume treats the body as fresh, so the collision is observability-only).
- **#2 `until`/`max_iterations` optional for `until_bash`-only / fixed-count loops** — this is existing `loop:` behavior (the shared `loopControlSchema` predates this PR). Changing it would alter `loop:` semantics and belongs in a separate PR.
- **#3 `node_failed` event on max-iter** — `loop:` doesn't emit one either; kept consistent.
- **#5 persist `$LOOP_PREV` across interactive resume** — requires extending `ApprovalContext` to carry the body-output snapshot; tracked as a follow-up. Test added locks the current (non-persisting) behavior.

`bun run validate` is green.

xbeark:
Second review processed in `3a58fc0f`. Breakdown:

**Already fixed (stale duplicates — evaluated against the pre-rebase commit):**
- `dag-node.ts:710-713` (#10) — the transform already spreads `aiOnly` to preserve `model`/`provider`; verified in current code.
- `dag-executor.ts:2383-2401` (#6) — `iterCtx` already forwards `workflowModel`/`workflowLevelOptions`/`aiProfile`/`workflowPreset`; verified in current code.

Both were fixed in `7e3b5c6f` (the rebase + CodeRabbit-fix commit). The rebase moved the line numbers, which likely caused the stale re-flag.

**Newly fixed in `3a58fc0f`:**
- `dag-executor.ts:2422-2444` — **post-body status check**: after `runLayers(iterCtx)` returns, a body approval/cancel node may have paused/cancelled the run mid-iteration. Now re-checks `getWorkflowRunStatus` before snapshot/completion handling; a terminal/cancelled state stops the loop (skips terminal-output selection + next iteration). `paused` is tolerated (mirrors `executeLoopNode`'s between-iteration tolerance).
- `dag-executor.ts:2654-2675` — **shell-safe `$LOOP_USER_INPUT` in bash/script/cancel body fields**: `applyLoopPrevToBodyNode` now `shellQuote`s `loopUserInput` before splicing into shell-bound fields (`escapedForBash=true` path). User input is free-text; AI-bound fields keep the raw value.

**Still deferred:**
- `dag-executor.ts:3576-3577` (#1, namespaced `step_name` through `executeNodeInternal`) — 15 event-write call sites; invasive. `runLayers`' own skip/trigger/when events ARE namespaced via `stepNamePrefix`; only `executeNodeInternal`'s internal events use the raw id. Maintained as a known v1 limitation (resume treats the body as fresh, so the collision is observability-only). Happy to do this as a follow-up PR if desired.

`bun run validate` green; 307 dag-executor tests pass.

Wirasm (changes requested):
---
pr: 2032
title: "feat(workflows): loop_group node — cross-node subgraph looping"
author: "xbeark"
reviewed: 2026-07-13T17:15:00Z
---

# PR Review: 2032 — feat(workflows): loop_group node — cross-node subgraph looping

**Author**: @xbeark
**Branch**: `feat/loop-group-node` -> `dev`
**Files Changed**: 12 (+2566/−240)
**Review mode**: multi-agent fan-out (7 aspects), all findings below verified against the checked-out branch by the orchestrating reviewer; the two top bugs were empirically reproduced.

---

## Summary

Adds a `loop_group` DAG node that repeats a sealed multi-node sub-DAG until an `until` signal / `until_bash` exit-0 / `max_iterations`, built on a cleanly extracted `runLayers` helper. The engineering quality is high: the `runLayers` extraction was independently line-diffed by two agents and confirmed behavior-preserving, the 21 new executor tests are unusually well-labeled and rigorous, and prior CodeRabbit rounds were genuinely addressed. However, the review found **two high-severity functional bugs that ship broken behavior the PR itself documents as working** (loader rejects outer `$nodeId.output` refs in body prompts — reproduced; interactive resume silently discards AI session continuity for `fresh_context: false`), plus a script-node corruption bug and several places where `loop_group` silently diverges from the `loop:` implementation it claims to mirror.

---

## Implementation Context

| Artifact | Path |
|----------|------|
| Implementation Report | Not found locally (plan is a gitignored local artifact per PR body) |
| Original Plan | `.claude/archon/plans/subgraph-loop-group-node.plan.md` (not in repo) |
| Documented Deviations | 6 (from PR body + author review-response comments) |

Deviation documentation quality is good: the PR body, review responses, and even tests explicitly lock in known v1 limits (unnamespaced body lifecycle events, `$LOOP_PREV` not surviving resume, `until`+`max_iterations` both required, no `node_failed` on max-iter, deferred features). None of those were re-flagged. Findings below are all **outside** the documented list.

---

## Changes Overview

| File | Changes | Assessment |
|------|---------|------------|
| `packages/workflows/src/dag-executor.ts` | +918/−222 | WARN — refactor is clean; new loop_group bookkeeping has bugs |
| `packages/workflows/src/loader.ts` | +18/−0 | FAIL — body ref-scan breaks documented feature |
| `packages/workflows/src/dag-executor.test.ts` | +1189/−0 | PASS — thorough, two specific gaps |
| `packages/workflows/src/schemas/dag-node.ts` | +91/−9 | WARN — missing `loop_group?: never` on siblings |
| `packages/workflows/src/schemas/loop.ts` | +22/−4 | PASS |
| `packages/workflows/src/schemas/index.ts` | +8/−2 | PASS |
| `packages/workflows/src/schemas.test.ts` | +118/−0 | PASS |
| `packages/workflows/src/loader.test.ts` | +85/−0 | PASS (gap: no outer-ref case — would have caught H1) |
| `packages/workflows/src/validator.ts` | +15/−2 | PASS |
| `packages/paths/src/telemetry.ts` | +1/−0 | PASS (see L8) |
| `docs guides/loop-nodes.md` | +99/−0 | WARN — one factual overclaim, two omissions |
| `docs book/dag-workflows.md` | +2/−1 | PASS |

---

## Issues Found

### Critical

No critical issues found (no security/data-loss/crash class defects).

### High Priority

- **H1 — `packages/workflows/src/loader.ts:225-238`** — Loader rejects `$nodeId.output` refs from a loop_group body to an outer-DAG node, breaking the PR's own documented feature. `[types-agent + errors-agent + comments-agent — independently reproduced twice, re-reproduced by orchestrator]`
  - **Why**: The recursive `validateDagStructure(node.loop_group.nodes)` call re-runs the `$nodeId.output` reference scan with only the body's node-id set. The docs (`loop-nodes.md`: "Outer context is still reachable via `$nodeId.output` refs in body prompts") and the executor (which seeds `scopedNodeOutputs` from `outerNodeOutputs`, tested by EDGE B) both support this pattern — but EDGE B calls `executeDagWorkflow` directly, bypassing the loader. Real YAML fails to load: `loop_group 'grp' body: Node 'work' references unknown node '$setup.output'` (reproduced against this branch).
  - **Fix**: Thread the outer id set into the recursion and check body output-refs against `bodyIds ∪ outerIds`, while keeping `depends_on`/cycle checks body-sealed (those are correctly scoped today). Add a loader test with an outer-ref body prompt.

- **H2 — `packages/workflows/src/dag-executor.ts:2327, 2425-2427, 2629-2635`** — Interactive pause/resume silently drops AI session continuity for `fresh_context: false`. `[tests-agent + code-agent + errors-agent]`
  - **Why**: Three compounding omissions vs the sibling `executeLoopNode` (which gets all three right at lines 2774/2834/3331): (a) `pauseWorkflowRun` hardcodes `sessionId: undefined`; (b) on resume the loop-level session cursor is never seeded from `loopGateMeta.sessionId`; (c) freshness gates on `i === startIteration` (always true for the first post-resume iteration) instead of the literal first iteration. Net: every resumed iteration of an interactive `loop_group` starts the body's AI conversation from scratch, defeating `fresh_context: false` for its most common use case (human-reviewed iterative loops). This is distinct from the documented `$LOOP_PREV`-across-resume deviation, which covers the text-snapshot substitution, not the provider session. The existing INTERACTIVE resume tests never assert on the `sendQuery` session-id argument, so this is untested.
  - **Fix**: Persist `loopLastSequentialSessionId` into the pause metadata's `sessionId`, restore it on resume, and gate freshness on the true first iteration — mirroring `executeLoopNode`. Add a resume test asserting the session-id argument.

- **H3 — `packages/workflows/src/dag-executor.ts:2718`** — `script:` body nodes referencing `$LOOP_PREV`/`$LOOP_USER_INPUT` get bash-quote artifacts injected into their TS/Python source. `[code-agent]`
  - **Why**: `applyLoopPrevToBodyNode` calls `sub(node.script, true)` (shell-quoting mode), but script nodes never pass through a shell — `executeScriptNode`'s own substitution deliberately uses `escapedForBash: false` (line 2050; scripts run via `execFileAsync` argv). Verified: substituted values get wrapped in `'...'` with `'\''` escapes embedded literally in source code — garbled values at best, syntax errors whenever prior output contains a quote. No test covers a script node inside a loop_group body.
  - **Fix**: `sub(node.script, false)`. Same root cause, lower stakes: the `cancel:` branch at 2719 also wrongly uses `true` (normal-path cancel substitution at 3979 uses the default `false`) — cosmetic quote artifacts in the user-visible cancellation message.

- **H4 — `packages/workflows/src/dag-executor.ts:2333-2650`** — A hard body-node failure doesn't stop the loop; it silently burns all remaining iterations, then reports a generic max-iterations error that hides the real root cause. `[errors-agent]`
  - **Why**: Terminal-output selection filters on `state === 'completed'`, so a failed body yields no signal and the loop re-runs the full body (real AI cost per iteration) until `max_iterations`. This is the exact anti-pattern already fixed for `loop:` — the comment at dag-executor.ts:2943-2946 documents it ("Previously we broke silently… kept burning iterations"), and `executeLoopNode` fails fast on iteration failure. The EDGE A test locks in the burn-to-max behavior with an explanatory comment, so it appears semi-deliberate — but it is **not** in the PR's documented-deviations list, and it contradicts the repo's fail-fast principle.
  - **Fix**: After each iteration's `runLayers`, check `scopedNodeOutputs` for a `failed` body node and return `state: 'failed'` with the body node's actual error (optionally only for FATAL-classified errors, reusing the existing classifier). Alternatively, if retry-through-failure is intended for refinement loops, document it explicitly as a deviation and include the last body error in the max-iterations message.

### Medium Priority

- **M1 — `packages/workflows/src/loader.ts:458`** — `hasInteractiveLoop` checks only `isLoopNode`, so an interactive `loop_group` in a non-interactive workflow silently skips the `interactive_loop_in_non_interactive_workflow` warning. Fix: also check `isLoopGroupNode(n) && n.loop_group.interactive === true`; add a loader test. `[tests-agent + comments-agent]`
- **M2 — `docs guides/loop-nodes.md:376` (+ JSDoc at dag-executor.ts:2268, 3575)** — Docs assert "Body node events are namespaced `{groupId}.{nodeId}`", but only `runLayers`' control events (skip/trigger/when) are; `node_started`/`node_completed`/`node_failed`/tool events use the raw node id. The unnamespaced behavior is an accepted deviation — the docs claiming the opposite is the bug. Qualify the sentence in both the guide and the JSDoc it was lifted from. `[docs-agent + comments-agent]`
- **M3 — `packages/workflows/src/dag-executor.ts:2511`** — `until_bash` env gets `LOOP_PREV_OUTPUT: iterationOutput` (the *current* iteration's output), while `executeLoopNode` captures the true previous iteration's value before overwriting (3136/3175). Same env-var name, different semantics between "mirrored" implementations. Fix to match `loop:`, or document the intentional difference. `[errors-agent + simplify-agent]`
- **M4 — `packages/workflows/src/dag-executor.ts:496,503,506`** — `substituteLoopPrevRefs` hardcodes `undefined` for `outputFileDir`, so large `$LOOP_PREV` values in `bash:`/`script:`/`until_bash:` fields are always inlined instead of spilling to a file over the 32KB threshold like `$nodeId.output` refs do (E2BIG risk in exactly the fix-until-green loops this feature targets). Thread `logDir` through, mirroring `substituteNodeOutputRefs`. `[errors-agent + code-agent]`
- **M5 — `packages/workflows/src/loader.ts:143` + `dag-executor.ts:2390`** — No id-uniqueness check across scopes: a body node sharing an id with an outer node silently shadows the outer entry in the scoped output map for that iteration, with no validation error or warning. Add a load-time check (or documented warning) for body ids colliding with outer ids. `[code-agent]`
- **M6 — `packages/workflows/src/dag-executor.ts:2699, 2704-2712`** — A nested `loop:` node's `until_bash` (and a nested `loop_group`'s own `until_bash`) never receive `$LOOP_PREV`/`$LOOP_USER_INPUT` substitution — the ref reaches `bash -c` unresolved, where `$LOOP_PREV` is valid-but-empty shell syntax: a silent wrong-behavior failure. Substitute `until_bash` with `escapedForBash: true` in both nested branches. `[simplify-agent]`
- **M7 — `packages/workflows/src/schemas/dag-node.ts` (7 variants, lines ~217-392)** — None of the pre-existing node variants gained `loop_group?: never`, weakening the union's mutual-exclusion guarantee that the codebase's `node.command ?? node.id` idioms rely on. Mechanical fix: add it to all seven sibling intersections. `[types-agent]`
- **M8 — Docs coverage: `reference/variables.md` (no `$LOOP_PREV` in any of its three tables), `book/quick-reference.md` (no `loop_group` row or options section), `guides/authoring-workflows.md:181-191` (node-type table not updated), `CLAUDE.md:811` (node-type enumeration)** — four canonical enumerating pages weren't touched for a first-class new node type. `[docs-agent]`
- **M9 — `docs guides/loop-nodes.md:418-435`** — `$LOOP_PREV` resetting to `''` after an interactive resume (documented deviation) isn't mentioned in the Resume or NOT-supported sections, despite the sibling `$LOOP_PREV_OUTPUT` docs explicitly warning about the same behavior. `[docs-agent]`
- **M10 — `docs guides/loop-nodes.md` (~line 405)** — `model`/`provider` on `loop_group` ARE honored (forwarded as body defaults), the opposite of `loop:`'s documented "silently ignored" behavior — this intentional difference is undocumented and will surprise readers coming from the `loop:` section. `[docs-agent]`
- **M11 — `docs guides/loop-nodes.md:420-423`** — Resume doc conflates two behaviors: gate resume advances to the *next* iteration (`startIteration = iteration + 1`), while a crash resume with no gate state restarts from iteration 1. Both are described as "re-runs the current iteration's whole body". Split into two accurate statements. `[comments-agent]`

### Suggestions

- **L1 — `dag-executor.ts:2522`** — `if (bashErr.code === 'ENOENT' || bashErr.code !== undefined)` is tautological (first clause subsumed) and collapses the two distinct log events `executeLoopNode` keeps (`until_bash_exec_error` vs `until_bash_unexpected_error`) into one, losing diagnostic granularity while claiming to mirror it. `[simplify-agent]`
- **L2 — Dead code**: `_resolvedOptions` param on `executeLoopGroupNode` (its producing call can even send a user-visible provider-conflict warning for a value that's never read) and the never-read `nodes` field on `RunLayersContext`. Drop both. `[simplify-agent + types-agent]`
- **L3 — `dag-executor.ts:2441`** — Reuse `shouldContinueStreamingForStatus(postBodyStatus)` (used two calls earlier in the same function) instead of re-deriving `!== 'running' && !== 'paused'`. `[simplify-agent]`
- **L4 — `dag-executor.ts:2458-2463` vs `4772-4776`** — Terminal-output selection is byte-for-byte duplicated; correctness-critical enough to justify a shared `selectTerminalOutput()` despite only two occurrences. `[simplify-agent]`
- **L5 — `dag-executor.ts:3631`** — `runLayers` docstring says "(planned) loop_group body iteration"; it's implemented in this same PR. `[comments-agent]`
- **L6 — `schemas/dag-node.ts:711-715` vs `443-448`** — "allowed by `LOOP_GROUP_NODE_AI_FIELDS`" vs that constant's own "unsupported on loop_group" docstring contradict; reword. `[comments-agent]`
- **L7 — `schemas/dag-node.ts:454-463`** — `dagNodeSchema` docstring ("Exactly one of command/prompt/bash/loop"; retry note) not updated for `loop_group`. `[comments-agent]`
- **L8 — `packages/workflows/src/executor.ts:749`** — `usesLoop` telemetry checks only `isLoopNode`; `loop_group` adoption is invisible. Fold in `isLoopGroupNode` or add a dedicated property. `[docs-agent]`
- **L9 — Test gaps** (beyond the H1/H2/H3 regression tests implied above): hard assertion for `fresh_context` session threading (the `loop:` suite proves it's easy — mirror lines 4409/4462); `when:` conditions on body nodes; multi-node body where an upstream fails and a downstream references its output; `loopTotalTokens` accumulation (only cost is asserted). `[tests-agent]`
- **L10 — Confirm intent**: `executeLoopGroupNode` never strips the `until` completion marker from output (`stripCompletionTags` is `loop:`-only, line 2885), so the marker text leaks into `$groupId.output` for downstream consumers. Possibly acceptable v1 behavior — confirm and document either way. `[code-agent, low confidence]`
- **L11 — Note for CLAUDE.md/future reviewers**: `LoopGroupNodeConfig` is a justified hand-written-type exception to the z.infer-only rule (recursive-inference cycle, drift caught by the `z.ZodType` annotation) — worth recording as the precedent for recursive schemas. `[types-agent]`

---

## Agent Findings

| Aspect | Agent | Findings (C/H/M/L) | Verdict |
|--------|-------|--------------------|---------|
| Code correctness | code-reviewer | 0/2/2/2 | REQUEST CHANGES |
| Error handling | silent-failure-hunter | 0/2/2/1 | REQUEST CHANGES |
| Tests | pr-test-analyzer | 0/1/4/2 | WARN |
| Types | type-design-analyzer | 0/1/1/2 | WARN |
| Comments | comment-analyzer | 0/2/3/2 | WARN |
| Docs | docs-impact | 0/4/3/2 | WARN |
| Simplify | code-simplifier | 0/2/4/0 | WARN |

Deduplication note: severities above are as each agent reported; the merged list re-ranks after cross-verification. Cross-agent agreement was strong — H1 was found and *independently reproduced* by three agents, H2 by three, M1/M2/M3/M4 each by two. Every High/Critical claim and all key Medium claims were re-verified against the source (and H1 re-reproduced) by the orchestrating reviewer before inclusion; nothing here is unverified agent output.

---

## Validation Results

| Check | Status | Details |
|-------|--------|---------|
| check:bundled / bundled-skill / bundled-schema / pi-vendor-map | PASS | all generated files in sync |
| Type Check | PASS | no errors |
| Lint | PASS | 0 warnings (max-warnings 0) |
| Format | PASS | |
| Tests | PASS | full per-package suite green (incl. 21 new loop_group executor tests) |

`bun run validate` exits 0 on the branch as checked out. All bugs found are semantic — invisible to the current suite because the loader path and session-id/argv-content assertions aren't exercised.

---

## Pattern Compliance

- [x] Follows existing code structure (mirrors `executeLoopNode`; schemas in the right place; camelCase schema names; z from `@hono/zod-openapi`)
- [x] Type safety maintained (no unjustified `any`; one justified, well-commented z.infer exception)
- [ ] Fully consistent with the `loop:` implementation it mirrors (H2, H4, M3, L1 diverge silently)
- [x] Tests added for new code (thorough, with the specific gaps in L9)
- [ ] Documentation fully updated (M2, M8–M11)

---

## What's Good

- The `runLayers` extraction — the riskiest part of the PR — was independently line-diffed by two agents and confirmed a verbatim, behavior-preserving lift (only `stepNamePrefix` added); all fire-and-forget event patterns, the NaN token guard, and upstream tier/cold-resume wiring preserved.
- Test suite quality is well above average: labeled dimensions (INSTANCE/EDGE/INTERACTIVE/COST), a sentinel-file technique to *prove* the until_bash short-circuit, and tests that deliberately lock in documented v1 limitations.
- Honest scope discipline: sealed-body design keeps the top-level DAG acyclic; deferrals are explicit in the PR body, review responses, docs ("What is NOT supported"), and tests.
- Shared `loopControlSchema` factoring, recursive body schema (zod v4 getter pattern), and loader/validator recursion into nested bodies are all clean; every `.loop_group` access goes through the `isLoopGroupNode` guard.
- All three prior CodeRabbit rounds were genuinely addressed — none of the previously-fixed items regressed.

---

## Recommendation

**REQUEST CHANGES**

H1 ships a load-time rejection of the feature's own documented (and executor-supported) outer-context pattern — any real YAML author hits it immediately. H2 silently defeats `fresh_context: false` across interactive gates, the flagship use case for an interactive refinement loop. H3 corrupts script-node bodies that use the new refs. All three have small, well-localized fixes with clear reference implementations in the sibling `executeLoopNode` / `executeScriptNode` code, plus the regression tests named above. H4 needs either a fail-fast or an explicit deviation entry. The mediums are worth batching into the same pass (M1–M6 are one-to-five-line fixes); the docs set (M2, M8–M11) can ride along or follow up. Once H1–H4 land, this is a strong merge candidate.

---

*Report: `.claude/PRPs/reviews/pr-2032-review.md`*

Wirasm:
Took over the branch, rebased onto current `dev`, and addressed the review findings (labels reference the review above). Nine commits on top of the original stack; `bun run validate` green.

**High — all fixed**

- **H1 (loader rejects outer `$nodeId.output` refs in body prompts)** — `validateDagStructure` now threads the accumulated enclosing-scope ids into the body recursion: output refs check body ∪ enclosing ids, while `depends_on`/cycle checks stay body-sealed. The original repro (outer `setup` bash node referenced from a body prompt) loads cleanly; loader tests added for outer-ref accept, truly-unknown reject.
- **H2 (session continuity dropped across interactive pause/resume)** — the pause now persists the loop-level session cursor in `ApprovalContext.sessionId`, resume restores it, and the freshness gate compares against the loop's literal first iteration instead of `startIteration` — mirroring `executeLoopNode`. The resume test now asserts the session id on both the pause call and the resumed `sendQuery` call.
- **H3 (script/cancel body fields bash-quoted)** — `script:` and `cancel:` body fields now substitute raw (matching their normal-path modes: scripts run via execFile argv, cancel is display text); `bash:`/`until_bash` remain shell-quoted. Unit test with a quote-containing prior output locks each mode.
- **H4 (failed body node burned iterations to max)** — the group now fails fast: after each iteration it scans the scoped outputs for failed body nodes and returns immediately with the failed node's own error (accumulated cost/tokens still charged). This mirrors both `executeLoopNode` and the top-level DAG's any-failed-fails-the-run semantics. EDGE A rewritten to lock in fail-fast; a new multi-node variant (EDGE A2) covers upstream-fails→dependent-skipped.

**Medium — all fixed**

- **M1** — the interactive-loop load warning now covers `loop_group` gates, including loops nested inside bodies; loader test added.
- **M2** — docs and the two internal JSDoc blocks now state accurately that only runLayers' control events are namespaced; body lifecycle events keep the raw node id (known v1 limitation).
- **M3** — `until_bash` env `LOOP_PREV_OUTPUT` now carries the *previous* iteration's output (captured before overwrite), matching `loop:` and the documented variable semantics.
- **M4** — `substituteLoopPrevRefs` gained the `outputFileDir` spill parameter; >32KB `$LOOP_PREV` values in shell-bound fields now spill to a file exactly like `$nodeId.output` refs.
- **M5** — body node ids that shadow an enclosing DAG node id are rejected at load time (the executor seeds the scoped map with outer outputs, so a collision would silently shadow).
- **M6** — nested `loop:`/`loop_group:` `until_bash` inside a body now receives `$LOOP_PREV`/`$LOOP_USER_INPUT` substitution shell-safely (previously reached `bash -c` unresolved, where `$LOOP_PREV` degrades to an empty shell var).
- **M7** — `loop_group?: never` added to all seven sibling `DagNode` variants.
- **M8–M11** — docs: `$LOOP_PREV` added to `reference/variables.md` (table, substitution order, availability matrix); `loop_group` row + options section in `book/quick-reference.md`; node-type rows in `guides/authoring-workflows.md` and `book/dag-workflows.md`; CLAUDE.md node list updated. The Resume section now distinguishes gate resume (next iteration, session continuity preserved) from failure resume (restarts at iteration 1); `$LOOP_PREV`-not-across-the-gate is listed under NOT supported; the `model`/`provider`-honored-unlike-`loop:` difference is called out; the inverted `fresh_context` example comment is fixed.

**Suggestions — addressed**

- **L1** — `until_bash` catch restored to the two-event split (`until_bash_exec_error` / `until_bash_unexpected_error`), dropping the tautological condition.
- **L2** — dead `_resolvedOptions` parameter and dead `RunLayersContext.nodes` field removed.
- **L3** — post-body status check reuses `shouldContinueStreamingForStatus`.
- **L5/L6/L7** — stale "(planned)" docstring, the contradictory `LOOP_GROUP_NODE_AI_FIELDS` comment, and the `dagNodeSchema` exclusivity docstring all corrected.
- **L8** — new `uses_loop_group` telemetry property alongside `uses_loop`.
- **L9** — tests added: token accumulation, hard `fresh_context` session-threading assertions (both modes), `when:` on body nodes (same-iteration gating, no skip-state leak across iterations), multi-node body failure.
- **L10** — completion-signal tags are now stripped from the group's stored/returned output (`stripCompletionTags`, same as `loop:`); signal detection still runs on the raw output.
- **L11** — the recursive-schema `z.ZodType` pattern is recorded in CLAUDE.md as the sanctioned `z.infer` exception.

**Deferred (with rationale)**

- **L4 (extract shared `selectTerminalOutput`)** — two occurrences; the repo's DRY policy is explicit rule-of-three, so the duplication stays for now. Both sites carry the same comment ("mirrors the top-level run's terminal-output selection") to keep them findable if a third site appears.
- **Pre-existing documented v1 deviations** (unnamespaced body lifecycle events through `executeNodeInternal`, `until`+`max_iterations` both required in the shared `loopControlSchema`, `$LOOP_PREV` not persisted across the gate) remain deferred as before — but they're now accurately documented user-side rather than contradicted by the docs.

PR #2095, opened 2026-07-14
fix(cli): enforce requires:[github] gate on the CLI run path (#2089)

### LINKED ISSUE #2089: requires: [github] is enforced only on the chat-dispatch path — CLI and REST runs bypass the gate
## Problem

Workflow-level `requires: [github]` is meant to hard-block invocation before any worktree/clone/AI cost when the acting user hasn't connected their GitHub identity. The gate (`assertWorkflowRequirementsMet`, `packages/workflows/src/utils/workflow-requirements.ts:45-53`) has exactly **one** call site:

- `packages/core/src/orchestrator/orchestrator-agent.ts:636` (chat dispatch, guarded by `isPerUserGitHubEnabled()`)

There is no call site in:
- the CLI run path (`packages/cli/src/commands/workflow.ts`)
- the server REST run route (`POST /api/workflows/:name/run`)

So on a multi-user install (App mode + `TOKEN_ENCRYPTION_KEY`), a user without a connected GitHub identity can start a `requires: [github]`-gated workflow via the web console's run form or the CLI, and it proceeds to burn worktree + AI cost before failing later (or worse, running with the bot/fallback identity).

## Expected

Enforce the requirement at all three entry points, or centralize it inside `executeWorkflow()` so every caller inherits it. The pure-function design of `assertWorkflowRequirementsMet` makes this cheap — each entry point just needs to resolve `githubConnected` for the acting user.

Note: it's deliberately a no-op on solo PAT installs (`isPerUserGitHubEnabled()` guard) — that semantics should carry over.

## References

- Gate: `packages/workflows/src/utils/workflow-requirements.ts:45-53`
- Only call site: `packages/core/src/orchestrator/orchestrator-agent.ts:636`

Wirasm:
## Summary

- **Problem:** Workflow-level `requires: [github]` is meant to hard-block a run before any worktree/clone/AI cost when the acting user hasn't connected GitHub, but the CLI run path bypassed it — the pure gate `assertWorkflowRequirementsMet` had a single caller (the orchestrator's `dispatchOrchestratorWorkflow`), while `workflowRunCommand` creates the worktree and calls `executeWorkflow()` directly.
- **Why it matters:** On a multi-user install (App mode + `TOKEN_ENCRYPTION_KEY`), a user with no connected GitHub identity could start a `requires: [github]`-gated workflow from the CLI and burn worktree + AI cost before failing later — or run under the bot/fallback identity. This is a governance control being silently skipped.
- **What changed:** Added a CLI-side capability gate (`assertCliWorkflowRequirementsMet`) that mirrors the orchestrator gate and runs **before** the `--detach` fork and worktree creation, so it fails fast before any cost. It resolves the acting CLI user (`ARCHON_USER_ID`/`$USER` → Archon user id) and checks for a stored GitHub token, then calls the same pure `assertWorkflowRequirementsMet`.
- **What did NOT change (scope boundary):** The orchestrator gate is untouched. **No REST-route change** — `POST /api/workflows/:name/run` is *already* gated: it dispatches `/workflow run` through the orchestrator (`dispatchToOrchestrator` → `handleMessage` → `handleWorkflowRunCommand` → `dispatchOrchestratorWorkflow` → gate), with the acting `userId` flowing the whole way. The issue listed REST as a second bypass; tracing the current code shows it is covered transitively (see UX Journey). Adding a redundant REST gate was rejected (DRY/KISS/YAGNI). Centralizing the gate inside `executeWorkflow()` was also rejected: `@archon/workflows` must not depend on `@archon/core`, and in the CLI the worktree is created before `executeWorkflow()`, so an executor-level gate would fire *after* the cost it's meant to prevent.

### Before

```
Multi-user install, user with NO connected GitHub, `requires: [github]` workflow:

  CLI:  archon workflow run gated "x"
          resolves workflow
          creates worktree            ← COST incurred
          executeWorkflow()           ← AI cost, runs under fallback identity
        (gate never consulted)

  REST: POST /api/workflows/gated/run
          -> /workflow run -> orchestrator -> dispatchOrchestratorWorkflow
          -> GATE -> BLOCKED  (already correct)
```

### After

```
Multi-user install, user with NO connected GitHub, `requires: [github]` workflow:

  CLI:  archon workflow run gated "x"
          resolves workflow
          *assertCliWorkflowRequirementsMet(workflow)*  ← NEW, before --detach + worktree
          -> WorkflowRequirementError
          -> cli.ts prints "Error: This workflow requires a connected github identity…"
          -> exit 1, NO worktree, NO AI cost, NO detached child spawned

  REST: (unchanged — still blocked via the orchestrator gate)

Solo PAT install (no GITHUB_APP_ID): gate is a no-op; run proceeds as before.
```

### Before

```
CLI workflowRunCommand ──▶ worktree ──▶ executeWorkflow      (no gate)

orchestrator dispatchOrchestratorWorkflow ──▶ [gate] ──▶ worktree ──▶ executeWorkflow
REST /workflows/:name/run ──▶ orchestrator (/workflow run) ──▶ [gate]
```

### After

```
CLI workflowRunCommand ──▶ [~gate] ──▶ worktree ──▶ executeWorkflow   [~ new gate call]

orchestrator dispatchOrchestratorWorkflow ──▶ [gate] ──▶ worktree ──▶ executeWorkflow
REST /workflows/:name/run ──▶ orchestrator (/workflow run) ──▶ [gate]   (unchanged)
```

**Connection inventory:**

| `cli/commands/workflow.ts` | `@archon/workflows/utils/workflow-requirements` | **new** | CLI now calls the pure gate `assertWorkflowRequirementsMet` |
| `cli/commands/workflow.ts` | `@archon/core` (`isPerUserGitHubEnabled`, `getDecryptedAccessToken`) | **new** | Resolve gate context for the acting CLI user |
| `cli/commands/workflow.ts` | `@archon/core/db/users` | unchanged | Already imported; gate reuses `findOrCreateUserByPlatformIdentity` |
| `server routes/api.ts` | orchestrator | unchanged | REST already dispatches `/workflow run` through the gated orchestrator path |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `cli`
- Module: `cli:workflow`

## Change Metadata

- Change type: `bug`
- Primary scope: `cli`

## Linked Issue

- Closes #2089

```bash
bun run validate   # VALIDATE_EXIT=0 — type-check, lint, format:check, bundled/skill/schema/pi-vendor-map checks, and all package tests pass
bun test packages/cli/src/commands/workflow.test.ts   # 155 pass / 0 fail (4 new gate tests + 151 existing)
```

- Evidence provided: full `bun run validate` exited 0 with no failures; the CLI test suite passes with the 4 new regression tests.
- Regression proof: with the gate call removed, `blocks a requires:[github] workflow before any cost when enabled and not connected` fails (command resolves instead of rejecting) — confirming the test fails without the fix.
- Skipped commands: none.

## Security Impact (required)

- New permissions/capabilities? `No` (tightens an existing capability gate)
- Secrets/tokens handling changed? `No` (reads an existing per-user GitHub token via the established `getDecryptedAccessToken`; never logs it)
- Risk/mitigation: This closes a governance bypass. The gate fails **closed** — an unresolvable CLI user or a token-lookup error is treated as "not connected".

## Compatibility / Migration

- Backward compatible? `Yes` — no-op on solo PAT installs; multi-user installs only see enforcement for `requires: [github]` workflows run by unconnected users (the intended behavior).
- Config/env changes? `No`

- Verified scenarios: full `bun run validate` (exit 0); CLI workflow test suite (155 pass); confirmed the block test fails when the gate is removed and passes when restored.
- Edge cases checked: `--no-worktree` path (gate runs before it); `--detach` (gate runs in the parent before the fork, so no child is spawned on block); solo-install no-op (no token lookup); workflow without `requires` (no token lookup).
- What was not verified: a live end-to-end run against a real GitHub-App multi-user install (covered by unit tests + the mirrored orchestrator behavior).

- Affected subsystems/workflows: `archon workflow run` (CLI) only. The gate short-circuits on the cheap `isPerUserGitHubEnabled()` env check for the common case.
- Potential unintended effects: In the gated case, `findOrCreateUserByPlatformIdentity` is called slightly earlier than before (idempotent). No behavior change for solo installs or non-`requires` workflows.
- Guardrails/monitoring: `cli.requirement_gate_user_resolve_failed` warn log on lookup failure; the `WorkflowRequirementError` message is user-facing and actionable.

## Rollback Plan (required)

- Fast rollback: revert this commit — the change is isolated to `packages/cli/src/commands/workflow.ts` (+ tests).
- Feature flags/toggles: none; behavior is already gated behind `isPerUserGitHubEnabled()`.
- Observable failure symptoms: a connected user being wrongly blocked (would indicate a token-resolution regression) — none observed.

## Risks and Mitigations

- Risk: A future refactor makes the REST route call `executeWorkflow()` directly, bypassing the orchestrator gate.
  - Mitigation: Documented the REST→orchestrator enforcement chain here; the orchestrator gate remains the single enforcement point for both chat and REST.
- Risk: CLI identity resolution differs from the orchestrator's.
  - Mitigation: The gate reuses the exact `resolveCliUserId` + `findOrCreateUserByPlatformIdentity('cli', …)` mapping the CLI already uses to attribute runs.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI-side capability gate for workflows that declare `requires: ['github']`.
  * Workflows now fail fast before forking/execution when no decrypted GitHub token is available.
  * Token lookup/identity resolution uses fail-closed behavior to prevent unintended runs.
  * GitHub requirements are skipped when per-user GitHub is disabled (solo installs) or when the workflow does not declare a GitHub requirement.

* **Tests**
  * Updated and expanded CLI workflow-run tests to make GitHub gating deterministic with stable user/token mocks.

### DIFF
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -90,6 +90,15 @@ mock.module('@archon/core', () => ({
   createWorkflowStore: mock(() => ({
     createWorkflowEvent: mock(() => Promise.resolve()),
   })),
+  // requires: [github] gate. Default to a solo-install posture (disabled) so the
+  // gate is a no-op for every existing test; the gate-specific tests below flip
+  // isPerUserGitHubEnabled on per-invocation.
+  isPerUserGitHubEnabled: mock(() => false),
+  getDecryptedAccessToken: mock(() => Promise.resolve(null)),
+}));
+
+mock.module('@archon/core/db/users', () => ({
+  findOrCreateUserByPlatformIdentity: mock(() => Promise.resolve({ id: 'user-cli-1' })),
 }));
 
 mock.module('@archon/workflows/workflow-discovery', () => ({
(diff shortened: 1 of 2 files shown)


Wirasm:
## Automated Code Review

### Summary

Self-review via `code-reviewer` + `code-simplifier` on the diff. Verdict: the fix correctly closes the CLI bypass, fails closed, and the gate is ordered before the `--detach` fork and any worktree/AI cost. Two non-blocking findings were raised; one is closed, one is intentionally left as-is with rationale.

### Strengths (verified)
- Gate placement is correct: after workflow resolution + flag validation, before the `--detach` fork and before worktree/codebase registration — so no cost precedes it.
- `workflowResumeCommand` and `workflowApproveCommand` both delegate back into `workflowRunCommand`, so resume/approve are covered too — no residual CLI bypass.
- REST route confirmed already gated transitively: `POST /api/workflows/:name/run` -> `dispatchToOrchestrator` -> `handleMessage` -> `/workflow run` -> `handleWorkflowRunCommand` -> `dispatchOrchestratorWorkflow` gate, with the acting `userId` flowing through. The only `executeWorkflow()` call site outside that gated path was the CLI one this PR patches.
- Reuses the shared pure policy (`assertWorkflowRequirementsMet` / `WorkflowRequirementError`) verbatim — no duplicated policy logic.
- Mock isolation is correct: the new `mock.module('@archon/core/db/users', ...)` is a single non-conflicting addition; `workflow.test.ts` runs in its own isolated `bun test` invocation.

### Findings

- **Coverage gap (fixed):** the initial tests didn't exercise the fail-closed branch (unresolvable CLI identity / lookup throws). Added two regression tests in `workflow.test.ts` (commit `3f1d206f`) — both assert the run is blocked and `executeWorkflow` is never called. Suite now 157 pass / 0 fail.
- **Double user-resolution (won't fix, by design):** `assertCliWorkflowRequirementsMet` resolves the acting CLI user, and `workflowRunCommand` resolves it again later for message attribution — an extra DB round-trip **only** in the narrow gated case (`isPerUserGitHubEnabled()` true AND workflow declares `requires: [github]`). Consolidating isn't a clean, behavior-preserving change: the gate must run **before** the `--detach` fork, while the later resolution only happens on the non-detach path after conversation creation — hoisting it would add a DB upsert to the detach parent path that doesn't happen today. `code-simplifier` independently reached the same "don't change it" conclusion. Left as-is for KISS + a clear blast radius.

### Simplification (from code-simplifier)
- No worthwhile simplifications. The helper is a single early-return guard + one fail-closed `try/catch` + one delegated call, and follows the file's existing `assert*` helper convention.

### Security
- Tightens an existing governance gate; fails closed (unresolvable user or lookup error → treated as "not connected"). Reads an existing per-user GitHub token via `getDecryptedAccessToken`; never logs it. No new network calls or permissions.

### Checklist
- [x] Fix addresses root cause (CLI bypass) from the investigation
- [x] Code follows codebase patterns (mirrors the orchestrator gate)
- [x] Tests cover the change (blocked / allowed / no-op / fail-closed)
- [x] No obvious bugs introduced
- [x] `bun run validate` passes (exit 0)

---
*Self-reviewed by Claude - ready for human review*

PR #2110, opened 2026-07-14
fix(web): supersede abandoned console loads

### LINKED ISSUE #2101: Console store: resubscribe during an abandoned in-flight load never invokes the new loader
## Summary

- What broke: In the console store (`packages/web/src/experiments/console/store/cache.ts`), a component that subscribes to a key while a *previous, abandoned* load for that key is still in flight never gets its own loader invoked — `ensureLoad`'s dedup guard (`if (cache.has(key) || inflight.has(key)) return;`, `cache.ts:39` on `dev`) short-circuits on the stale `inflight` entry, and the new subscriber silently inherits whatever the old request eventually returns.
- When it started: Pre-existing since the store was introduced — surfaced during the multi-agent review of PR #1938 (two review agents reproduced it independently).
- Severity: `minor` (correctness quirk on a race window; no leak, no crash)

## Steps to Reproduce

1. `const unsubA = subscribeKey(key, listener, loaderA)` where `loaderA` returns a promise that stays pending.
2. `unsubA()` — last subscriber leaves; `listeners`/`loaders`/`versions` are pruned, but the in-flight promise keeps running (`inflight` still has the key).
3. `const unsubB = subscribeKey(key, listener, loaderB)` before `loaderA` settles.
4. Resolve `loaderA`.

## Expected vs Actual

- **Expected**: Subscriber B's `loaderB` runs (or at minimum B is guaranteed a result produced by an equivalent loader for that key).
- **Actual**: `loaderB` is never invoked (call count 0). B receives `loaderA`'s result whenever it lands. If `loaderA` rejects, B gets an error produced by a loader it never registered.

## User Flow

```
User                        Console store
────                        ─────────────
opens run detail   ───────▶ subscribeKey(key, loaderA); fetch A in flight
navigates away     ───────▶ last unsubscribe: listeners/loaders/versions pruned
                            (inflight[key] still pending — by design, promises aren't cancelled)
navigates back fast ──────▶ subscribeKey(key, loaderB)
                            [X] ensureLoad sees inflight.has(key) → loaderB never called
fetch A settles    ───────▶ B renders A's (possibly stale/foreign) result
```

## Environment

- Platform: Web (console experiment UI)
- Database: n/a
- Running in worktree? n/a
- OS: any

## Impact

- Affected workflows/commands: console panels using `useEntity` under fast unmount/remount (clicking through runs/chats quicker than fetches settle). In practice loaders for the same key are usually equivalent closures, so the visible impact today is "slightly stale data until the next invalidate", not wrong data — which is why this is minor.
- Reproduction rate: Always (given the race window); reproduced deterministically in tests during the PR #1938 review.
- Workaround available? Yes — any `invalidate()`/SSE push for the key triggers a fresh revalidate.
- Data loss risk? `No`

## Scope

- Package(s) likely involved: `web`
- Module (if known): `web:console-store` (`packages/web/src/experiments/console/store/cache.ts`, `ensureLoad`)

## Notes

- Context: found during the multi-agent review of PR #1938 (see the review comment there); explicitly left out of that PR's scope because it pre-dates it and is a behavior question, not part of the #1933 leak.
- Interaction: the #1938 fix makes fast unmount/remount lean on this dedup path slightly more often (the remount is the intended "warm read" path), which is why it's worth deciding on deliberately rather than leaving implicit.
- Suggested directions (pick one, don't do both):
  - Key the dedup off request identity instead of bare `inflight.has(key)` — e.g. store a generation/epoch per key, bump it on last unsubscribe, and have `ensureLoad`'s `.then`/`.catch` no-op (or re-run the current loader) when the generation moved.
  - Or codify current behavior: document in the module contract that loaders for a given key must be equivalent (they effectively are today — key encodes the query), and add a regression test pinning the inherit-the-in-flight-result semantics so it's a stated invariant instead of an accident.

coyaSONG:
## Summary

- Problem: a console-store subscription that mounted while an earlier, now-abandoned load was still pending inherited that load, so its own loader was never invoked.
- Why it matters: navigating away and back during a slow request could leave the remounted view waiting on stale work or allow the old result to overwrite the current view.
- What changed: in-flight loads now carry an identity and an `abandoned` marker; a later subscriber supersedes abandoned work, and only the current load may commit a value or error. Regression tests cover both a cold remount and an abandoned revalidation over a warm cache.
- What did **not** change (scope boundary): no request cancellation was added; cache/error retention, the public store API, and non-console web behavior remain unchanged.

### Before

```
User                    Console store                 Loader
────                    ─────────────                 ──────
opens view A ─────────▶ starts load A ──────────────▶ pending
leaves view A ────────▶ removes last subscriber
opens view B ─────────▶ sees load A as in-flight
                         skips load B
sees stale/waiting UI ◀─ load A eventually commits ◀─ resolves
```

### After

```
User                    Console store                 Loader
────                    ─────────────                 ──────
opens view A ─────────▶ starts load A ──────────────▶ pending
leaves view A ────────▶ [marks load A abandoned]
opens view B ─────────▶ [starts load B] ────────────▶ resolves fresh
sees current data ◀──── [commits load B]
                         [ignores late load A] ◀────── resolves stale
```

### Before

```
Console components
       │ subscribe/useEntity
       ▼
store/cache.ts ───────▶ supplied async loader
  ├─ cache/errors
  ├─ listeners/loaders
  ├─ inflight: Map<key, Promise>
  └─ versions ────────▶ subscriber re-render

cache.test.ts ────────▶ public store lifecycle API
```

### After

```
Console components
       │ subscribe/useEntity
       ▼
[~] store/cache.ts ═══▶ supplied async loader
  ├─ cache/errors
  ├─ listeners/loaders
  ├─ [~] inflight: Map<key, { abandoned }>
  ├─ [+] load-identity commit gate
  └─ versions ────────▶ subscriber re-render

[~] cache.test.ts ═══▶ abandoned-load and warm-remount regressions
```

| Console components / `useEntity` | `store/cache.ts` subscription API | unchanged | Existing public API and snapshots are unchanged. |
| `store/cache.ts` | Supplied async loader | **modified** | A new subscriber may start a replacement when the prior load is abandoned. |
| Loader promise settlement | Cache, errors, and listeners | **modified** | Load identity gates stale value/error commits and notifications. |
| Last unsubscribe | In-flight load state | **new** | Marks current work abandoned without attempting promise cancellation. |
| SSE / optimistic callers | `set` and `patch` | unchanged | Mutation and notification semantics are unchanged. |
| `cache.test.ts` | Public store lifecycle API | **modified** | Adds deterministic cold- and warm-remount regression coverage. |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `web`
- Module: `web:console-store`

## Change Metadata

- Change type: `bug`
- Primary scope: `web`

## Linked Issue

- Closes #2101
- Related #1938
- Depends on: none
- Supersedes: none

Commands and result summary:

```bash
bun test packages/web/src/experiments/console/store/cache.test.ts
# 10 pass, 0 fail, 37 assertions

bun --filter @archon/web test
# 329 pass, 0 fail, 795 assertions

bun --filter @archon/web type-check
# passed

bun x eslint packages/web/src/experiments/console/store/cache.ts packages/web/src/experiments/console/store/cache.test.ts --no-warn-ignored
# passed

bun x prettier --check packages/web/src/experiments/console/store/cache.ts packages/web/src/experiments/console/store/cache.test.ts
# passed

bun run validate
# generated-artifact checks, all workspace type-checks, repository ESLint,
# and repository Prettier passed. The test phase reached one unchanged
# @archon/adapters failure described below.
```

- Evidence provided (test/log/trace/screenshot): the new cold-remount test failed before the implementation because loader B was called 0 times instead of 1. It now passes and also verifies that a late load A cannot overwrite load B. The warm-remount test verifies that an abandoned revalidation cannot overwrite retained cache state.
- If any command is intentionally skipped, explain why: none. `bun run validate` was run; its only failing test is the unchanged `GitHubAdapter > App mode > credential helper install is attempted after successful App-mode clone` test. It also fails when run alone and is outside this PR's two-file web scope: the adapter test mocks `@archon/paths` without `getArchonHome`, while `installCredentialHelper` calls that export before the asserted git-config call. All other workspace packages were also run independently; only that adapter test failed.

## Security Impact (required)

## Compatibility / Migration

- Config/env changes? (`No`)
- If yes, exact upgrade steps: none.

- Verified scenarios: read the repository contribution guidance and console architecture notes; reproduced the missing loader-B call before the fix; reviewed the final diff after rebasing onto current `dev`; confirmed the replacement load commits and notifies exactly once.
- Edge cases checked: the fresh replacement resolves before the abandoned request; an abandoned revalidation resolves after a warm remount; a late load with no replacement still warms the cache as before; rejected abandoned loads do not resurrect subscriber counters.
- What was not verified: no interactive browser session was needed for this deterministic store-lifecycle change. The unrelated adapter test noted above prevents a fully green aggregate `bun run validate` on this checkout.
- AI assistance: OpenAI Codex co-authored the implementation; repository instructions, the diff, and all reported validation output were reviewed before submission.

- Affected subsystems/workflows: experimental console store subscriptions and revalidation only.
- Potential unintended effects: a late result from abandoned work is intentionally discarded once a new subscriber owns the key; an identity-gating error could otherwise suppress a valid result.
- Guardrails/monitoring for early detection: focused tests cover cold and warm remounts, existing lifecycle tests preserve late-cache-warming behavior, and the full web test suite passes.

## Rollback Plan (required)

- Fast rollback command/path: `git revert 10efaf8a3b388089e9d3a07718dc0e9a5790f3f5`
- Observable failure symptoms: remounted console entities fail to load, show stale data, or receive duplicate notifications.

## Risks and Mitigations

- Risk: identity gating could discard a still-useful abandoned result.
  - Mitigation: an abandoned result may still warm the cache when no successor exists; it is discarded only after a current subscriber or replacement load owns the key, with both paths covered by regression tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reactive cache handling when subscriptions are removed and restored while data is loading.
  * Prevented outdated/abandoned async loader and revalidation results from overwriting newer cached values or errors.
  * Avoided unnecessary notifications by ensuring abandoned operations can’t supersede later invalidation or remounts, keeping warm cached data stable.

### DIFF
--- a/packages/web/src/experiments/console/store/cache.test.ts
+++ b/packages/web/src/experiments/console/store/cache.test.ts
@@ -8,7 +8,7 @@ function flush(): Promise<void> {
   return new Promise(resolve => setTimeout(resolve, 0));
 }
 
-describe('subscribeKey — per-key Map lifecycle (#1933)', () => {
+describe('subscribeKey — per-key lifecycle', () => {
   test('last unsubscribe prunes versions; cache is retained for warm remount', async () => {
     const key = 'test:prune-versions';
     let loads = 0;
@@ -126,6 +126,95 @@ describe('subscribeKey — per-key Map lifecycle (#1933)', () => {
     await flush();
     expect(versionOf(key)).toBe(0); // notify skipped — nothing subscribes
     expect(get(key)).toBe('late'); // cache still warms for a future remount
+  });
+
+  test('resubscribing supersedes an abandoned in-flight loader', async () => {
+    const key = 'test:resubscribe-inflight';
+    let resolveA: (v: string) => void = () => {};
+    let resolveB: (v: string) => void = () => {};
+    let loadsA = 0;
+    let loadsB = 0;
+
+    const unsubscribeA = subscribeKey(
+      key,
+      () => {},
+      () => {
+        loadsA += 1;
+        return new Promise<string>(resolve => {
+          resolveA = resolve;
+        });
+      }
+    );
+    expect(loadsA).toBe(1);
+
+    unsubscribeA();
+
+    let notificationsB = 0;
+    const unsubscribeB = subscribeKey(
+      key,
+      () => {
+        notificationsB += 1;
+      },
+      () => {
+        loadsB += 1;
+        return new Promise<string>(resolve => {
+          resolveB = resolve;
+        });
+      }
+    );
+
+    expect(loadsB).toBe(1);
+
+    resolveB('fresh');
+    await flush();
+    expect(get(key)).toBe('fresh');
+    expect(notificationsB).toBe(1);
+
+    resolveA('stale');
+    await flush();
+    expect(get(key)).toBe('fresh');
+    expect(notificationsB).toBe(1);
+
+    unsubscribeB();
+  });
+
+  test('an abandoned revalidation cannot overwrite a warm remount', async () => {
+    const key = 'test:resubscribe-revalidation';
+    set(key, 'warm');
+
+    let resolveA: (v: string) => void = () => {};
+    const unsubscribeA = subscribeKey(
+      key,
+      () => {},
+      () =>
+        new Promise<string>(resolve => {
+          resolveA = resolve;
+        })
+    );
+    invalidate(key); // Starts a revalidation while retaining the warm value.
+    unsubscribeA();
+
+    let loadsB = 0;
+    let notificationsB = 0;
+    const unsubscribeB = subscribeKey(
+      key,
+      () => {
+        notificationsB += 1;
+      },
+      () => {
+        loadsB += 1;
+        return Promise.resolve('fresh');
+      }
+    );
+
+    expect(loadsB).toBe(0); // The retained cache still satisfies a warm remount.
+
+    resolveA('stale');
+    await flush();
+    expect(get(key)).toBe('warm');
+    expect(notificationsB).toBe(0);
+
+    unsubscribeB();
   });
 
   test('a load rejecting after the last unsubscribe does not resurrect the counter', async () => {
--- a/packages/web/src/experiments/console/store/cache.test.ts
+++ b/packages/web/src/experiments/console/store/cache.test.ts
@@ -217,6 +217,73 @@ describe('subscribeKey — per-key lifecycle', () => {
     unsubscribeB();
   });
 
+  test('invalidating an abandoned revalidation cannot resurrect a purged key', async () => {
+    const key = 'test:invalidate-abandoned-revalidation';
+    set(key, 'warm');
+
+    let resolveLoad: (v: string) => void = () => {};
+    const unsubscribe = subscribeKey(
+      key,
+      () => {},
+      () =>
+        new Promise<string>(resolve => {
+          resolveLoad = resolve;
+        })
+    );
+
+    invalidate(key);
+    unsubscribe();
+    invalidate(key);
+    expect(get(key)).toBeUndefined();
+
+    resolveLoad('stale');
+    await flush();
+    expect(get(key)).toBeUndefined();
+  });
+
+  test('manual revalidation supersedes an abandoned load', async () => {
+    const key = 'test:supersede-abandoned-revalidation';
+    set(key, 'warm');
+
+    let resolveA: (v: string) => void = () => {};
+    const unsubscribeA = subscribeKey(
+      key,
+      () => {},
+      () =>
+        new Promise<string>(resolve => {
+          resolveA = resolve;
+        })
+    );
+    invalidate(key);
+    unsubscribeA();
+
+    let loadsB = 0;
+    let resolveB: (v: string) => void = () => {};
+    const unsubscribeB = subscribeKey(
+      key,
+      () => {},
+      () => {
+        loadsB += 1;
+        return new Promise<string>(resolve => {
+          resolveB = resolve;
+        });
+      }
+    );
+
+    invalidate(key);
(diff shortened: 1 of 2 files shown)


Wirasm:
Thanks for this, and sorry it sat for a while — your diagnosis was correct, you were just beaten to it by two days.

The stale-load race you identified was fixed by #2148 (`646d5263`, merged 2026-07-16), which closes the same issue (#2101). I didn't want to close this on the assumption that "another PR touched the same file", so I checked properly: I ran **your four tests, unchanged, against current `dev`'s `cache.ts`** — three pass, one fails. That single failure is the entire remaining difference between your approach and what's on `dev`, and it's a case where `dev`'s behavior is arguably the better one: it commits data fetched *later* than the warm value, whereas your version would leave the view staler with no pending load to replace it. `dev`'s own cache suite is 10 pass / 0 fail.

So the bug is gone and the residual delta isn't something I want to change. Closing on those grounds rather than on staleness.

**Two of your tests are worth keeping, though.** They cover revalidate paths that `dev` has no coverage for at all, and they pass against `dev` today — so they're pure additive value, independent of which implementation won. I've filed a follow-up to land them with credit to you rather than let them disappear with the PR.

For what it's worth on the substance: guarding at the point of state-write rather than fetch-initiation was the right call, and it's the part people usually get wrong. If you want another crack at the console, the run-log and event-stream areas are actively moving and could use eyes.

Issue #1260, opened 2026-04-16
[docker] docker-entrypoint.sh: bind mount breaks startup — /.archon/workspaces missing causes Claude subprocess timeout


ztech-gthb:
When using a bind mount for /.archon (via ARCHON_DATA in .env), the app starts successfully and the SQLite database is created, but sending any message in
  the Web UI gets no response. The Claude subprocess silently times out after 60 seconds.

  Named volumes work fine. The problem is specific to bind mounts on macOS (Docker Desktop with VirtioFS), but likely affects Linux bind mounts too.

  To Reproduce

  1. Set ARCHON_DATA=/some/host/path in .env (pointing to an empty or fresh directory)
  2. docker compose up -d
  3. Open the Web UI and send any message (e.g. "hi")
  4. No response — after ~60 seconds the request times out

  Expected behavior

  The app responds normally, same as with a named volume.

  Root cause

  orchestrator-agent.ts passes getArchonWorkspacesPath() (resolves to /.archon/workspaces in Docker) as the cwd for every Claude subprocess invocation —
  including simple chat messages with no codebase attached.

  With a named volume, Docker copies the image layer contents into the volume on first run, so /.archon/workspaces and /.archon/worktrees exist from the start
  (they are created in the Dockerfile via RUN mkdir -p /.archon/workspaces /.archon/worktrees).

  With a bind mount, the host directory is mounted as-is. Those subdirectories are never created on the host, so /.archon/workspaces does not exist inside the
  container. The Claude SDK attempts to spawn the subprocess with a non-existent cwd, which fails silently — no events are emitted, and the first-event timeout
   (60 s) fires.

  Fix

  Add mkdir -p for the required subdirectories in docker-entrypoint.sh, before the chown call. This is safe and idempotent for named volumes (the dirs already
  exist) and ensures they are created for bind mounts:

  --- a/docker-entrypoint.sh
  +++ b/docker-entrypoint.sh
  @@ -1,6 +1,10 @@
   #!/bin/bash
   set -e

  +# Ensure required subdirectories exist.
  +# Named volumes inherit these from the image layer on first run; bind mounts do not.
  +mkdir -p /.archon/workspaces /.archon/worktrees
  +
   # Determine if we need to use gosu for privilege dropping
   if [ "$(id -u)" = "0" ]; then

  Environment

  - macOS (Docker Desktop, VirtioFS)
  - ARCHON_DATA set to a local host path (bind mount)
  - Archon version: current dev branch

PR #2114, opened 2026-07-14
fix(core): fail fast when a scoped chat's working directory is gone

### LINKED ISSUE #1170: Docker deployment: Codex chats crash with 'No such file or directory (os error 2)'
## Summary

- What broke: In the Docker deployment, Codex-backed chats/workflows crash immediately with `Codex Exec exited with code 1: Error: No such file or directory (os error 2)`.
- When it started (if known): Observed on a fresh Docker deployment on 2026-04-13.
- Severity: `blocker`

## Steps to Reproduce

1. Deploy Archon via the Docker guide on a Linux VPS.
2. Set Codex credentials in `.env` and run the app normally under Docker.
3. Start a web conversation or any workflow that uses Codex.

## Expected vs Actual

- **Expected**: Codex should start and handle the chat/workflow.
- **Actual**: Codex crashes immediately with `Codex Exec exited with code 1: Error: No such file or directory (os error 2)`.

## User Flow

```
  User                   Archon Web/UI             Codex Runtime
  ────                   ─────────────             ─────────────
  starts chat ─────────▶ creates conversation
                         launches Codex ─────────▶ tries to exec codex wrapper
                         [X] process startup fails ◀── missing runtime dependency
  sees crash ◀────────── returns Codex crash error
```

## Environment

- Platform: Web
- Database: PostgreSQL
- Running in worktree? `No`
- OS: Linux VPS, Docker deployment

## Logs

```text
@archon/server setup-auth: ✅ Successfully created auth.json at: /home/appuser/.codex/auth.json
@archon/server setup-auth: ✅ Codex authentication and configuration complete
@archon/server start: {"level":50,"module":"provider.codex","err":{"message":"Codex Exec exited with code 1: Error: No such file or directory (os error 2)\n"},"errorClass":"crash","attempt":0,"maxRetries":3,"msg":"query_error"}
@archon/server start: {"level":50,"module":"orchestrator-agent","err":{"message":"Codex crash: Codex Exec exited with code 1: Error: No such file or directory (os error 2)\n"},"conversationId":"web-...","msg":"orchestrator_message_failed"}
```

## Impact

- Affected workflows/commands: Any Codex-backed chat/workflow
- Reproduction rate: Always
- Workaround available? If so, describe: Not a clean one from the documented Docker path
- Data loss risk? `No`

## Scope

- Package(s) likely involved: `core|server`
- Module (if known): `core:providers/codex`, Docker image/runtime

## Potential Root Cause

- The Docker image removes `node` after installing `agent-browser`, but the installed Codex entrypoint is still the Node wrapper at `/app/node_modules/.bin/codex` with shebang `#!/usr/bin/env node`.
- In the running container, `codex --version` fails with `env: 'node': No such file or directory`.
- `setup-auth` succeeds and writes `/home/appuser/.codex/auth.json`, so this does not appear to be a credential bootstrap failure.
- `resolveCodexBinaryPath()` does not appear to help in this Docker/Bun path because `BUNDLED_IS_BINARY=false`, so the SDK falls back to the wrapper resolution path.

waaaw:
## Summary

- Problem: Codex-backed chats in a project-scoped conversation crashed with the opaque `Codex Exec exited with code 1: Error: No such file or directory (os error 2)` whenever the resolved working directory (`conversation.cwd` or `scopedCodebase.default_cwd`) no longer existed on disk — e.g. a cleaned-up worktree, a deleted repo directory, or a reset volume in a fresh Docker deployment.
- Why it matters: The failure surfaced as a raw, unactionable Rust-level OS error deep inside the Codex binary instead of a clear Archon-level message, leaving users unable to tell what actually went wrong or how to fix it.
- What changed: The scoped-codebase branch in `handleMessage` now guards the resolved `cwd` with `existsSync` and throws an actionable error before ever reaching the provider; the error is classified first in `classifyAndFormatError` so it can't be misrouted by unrelated substring matches (e.g. "session", "token") or dropped by the sensitive-data length fallback.
- What did **not** change (scope boundary): No Dockerfile changes. The issue's proposed root cause (Dockerfile purging node/npm, breaking Codex's npm wrapper) was investigated and ruled out — `@openai/codex-sdk` spawns the native binary directly and never executes the JS shim, so a missing node would surface as a differently-formatted Node ENOENT, not this Rust-formatted error. `packages/workflows/src/executor.ts` has a similar theoretical gap but was deferred (narrower race window, tracked as a follow-up).

### Before

```
  User                        Archon                              Codex (native binary)
  ────                        ──────                              ─────────────────────
  sends message ───────────▶  resolves scoped cwd (no check)
                               streams to provider ─────────────▶  spawned with --cd <stale path>
                                                                    fails: "No such file or
                                                                    directory (os error 2)"
  sees opaque crash ◀────────  raw SDK error passed through
```

### After

```
  User                        Archon                              Codex (native binary)
  ────                        ──────                              ─────────────────────
  sends message ───────────▶  resolves scoped cwd
                               [existsSync check] ── missing ──▶  never invoked
                               throws actionable error
                               classifyAndFormatError picks
                               the new branch FIRST
  sees clear guidance ◀──────  formatted, actionable message
```

### Before

```
handleMessage (orchestrator-agent.ts)
  ├─ scoped-codebase branch: cwd = conversation.cwd ?? scopedCodebase.default_cwd  (unchecked)
  │    └─▶ provider.sendQuery(cwd) ──▶ Codex native binary --cd <cwd>
  └─ unscoped branch: ensureArchonWorkspacesPath() (mkdir -p, already safe)

classifyAndFormatError (error-formatter.ts)
  ├─ session-related branch
  ├─ sensitive-data / generic fallback branch
  └─ ... (no branch for missing scoped cwd)
```

### After

```
handleMessage (orchestrator-agent.ts)
  ├─ scoped-codebase branch: cwd = conversation.cwd ?? scopedCodebase.default_cwd
  │    ├─ [~] existsSync(cwd) guard ── missing ──▶ throw actionable Error
  │    └─▶ provider.sendQuery(cwd)  (only when cwd exists)
  └─ unscoped branch: unchanged (ensureArchonWorkspacesPath())

classifyAndFormatError (error-formatter.ts)
  ├─ [+] missing-scoped-cwd branch (FIRST check — must win over session/sensitive-data branches)
  ├─ session-related branch
  └─ sensitive-data / generic fallback branch
```

| `orchestrator-agent.ts` (scoped branch) | `fs.existsSync` | **new** | Guard added before provider dispatch |
| `orchestrator-agent.ts` (catch-all) | `error-formatter.ts` | unchanged | Existing catch-all already formats and sends fatal errors |
| `error-formatter.ts` (`classifyAndFormatError`) | new missing-cwd branch | **new** | Inserted as the first check to avoid misclassification |
| `orchestrator-agent.ts` (scoped branch) | Codex/Claude provider `sendQuery` | **modified** | Now unreachable when cwd is missing |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `core`
- Module: `core:orchestrator`

## Change Metadata

- Change type: `bug`
- Primary scope: `core`

## Linked Issue

- Closes #1170
- Related #1260 (prior fix for the unscoped docker-entrypoint.sh mkdir gap)
- Depends on #
- Supersedes #

Commands and result summary:

```bash
bun run type-check   # all 8 packages exit 0
bun run lint         # 0 errors, 0 warnings (--max-warnings 0)
bun run format:check # all files match Prettier
bun run test         # @archon/core full suite green; 2 pre-existing
                      # Windows-sandbox-only symlink EPERM failures in
                      # unrelated files (binary-resolver.test.ts,
                      # load-command-prompt.test.ts) - CI runs on
                      # Linux/macOS where these are expected to pass
bun run build         # all packages incl. docs-web build cleanly
```

- Evidence provided (test/log/trace/screenshot): Full command output captured during implementation; see `validation.md` in the run artifacts. Mutation-tested the guard by temporarily disabling it (`if (false && !existsSync(cwd))`) — the two new guard tests correctly flipped from pass to fail, confirming they detect the bug.
- If any command is intentionally skipped, explain why: Manual end-to-end reproduction (deleting a registered project directory against a live Docker install with a Codex credential) was not performed — outside the scope of this worktree. The automated tests simulate the exact failure condition (`existsSync` false for the resolved cwd) and assert the provider is never invoked.

## Security Impact (required)

- File system access scope changed? No — this narrows access (fails fast instead of attempting to pass a stale path to a subprocess)
- If any Yes, describe risk and mitigation: N/A

## Compatibility / Migration

- Backward compatible? Yes
- Config/env changes? No
- If yes, exact upgrade steps: N/A

- Verified scenarios: Automated tests cover (1) missing `scopedCodebase.default_cwd`, (2) missing `conversation.cwd` (worktree override case), and (3) a regression guard confirming the unscoped branch remains unguarded (still calls `ensureArchonWorkspacesPath()` to mkdir -p) when `existsSync` is forced false.
- Edge cases checked: Error-formatter branch ordering — verified the new branch is checked before the "session"-substring branch and the sensitive-data/generic fallback branch, since a filesystem path can accidentally contain those substrings (e.g. a path containing "session" or "token") and misroute the message.
- What was not verified: End-to-end Docker/Codex reproduction of the original crash (requires a live Docker build with Codex credentials, outside this worktree).

- Affected subsystems/workflows: `packages/core/src/orchestrator/orchestrator-agent.ts` (scoped-codebase chat dispatch path) and `packages/core/src/utils/error-formatter.ts` (error classification). No changes outside `@archon/core`.
- Potential unintended effects: None expected — the guard only adds a fail-fast check on a path that previously always failed downstream anyway (just with a worse error message).
- Guardrails/monitoring for early detection: Existing structured logging in `handleMessage`'s catch-all already logs and surfaces fatal errors to the user; no new monitoring needed.

## Rollback Plan (required)

- Fast rollback command/path: `git revert 980b763f` (single self-contained commit touching 4 files)
- Feature flags or config toggles (if any): None
- Observable failure symptoms: If reverted, scoped chats against a missing cwd would again surface the opaque Codex OS-level error instead of the actionable Archon message.

## Risks and Mitigations

- Risk: The new error-formatter branch could shadow future legitimate messages if it becomes too broad.
  - Mitigation: The branch uses a `startsWith` check against a specific, Archon-generated message prefix rather than a loose substring match, so it can't collide with unrelated error text.

### DIFF
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -260,8 +260,12 @@ mock.module('@archon/git', () => ({
   hasUncommittedChanges: mock(() => Promise.resolve(false)),
 }));
 
+// Default: every path exists. Hoisted to a named mock so the missing-cwd guard
+// (#1170) can be exercised by overriding it per-test — restore `() => true` after.
+const mockExistsSync = mock((_path: string) => true);
+
 mock.module('fs', () => ({
-  existsSync: mock(() => true),
+  existsSync: mockExistsSync,
   // token-crypto.ts imports these from node:fs for the auto-provisioned credential
   // key. readFileSync returns a valid 64-hex key so getEncryptionKey() resolves
   // without any real disk write when the per-user credential path is exercised.
@@ -1401,6 +1405,83 @@ describe('provider cwd resolution', () => {
       expect.objectContaining({ codebaseId: 'deleted-id' }),
       'orchestrator.scoped_codebase_not_found'
     );
+  });
+
+  // ─── missing cwd guard (issue #1170) ────────────────────────────────────────
+
+  describe('missing cwd guard (#1170)', () => {
+    test('sends an actionable error and never reaches the provider when default_cwd is gone', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      // Simulate the registered repo dir being removed after registration.
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      try {
+        const platform = makePlatform();
+        await handleMessage(platform, 'conv-1', 'hello');
+
+        // The provider must never be handed a nonexistent cwd — that is the
+        // whole point of the guard (Codex would die with "os error 2").
+        expect(mockSendQuery).not.toHaveBeenCalled();
+        expect(platform.sendMessage).toHaveBeenCalledWith(
+          'conv-1',
+          expect.stringContaining('Working directory no longer exists: /repos/test-repo')
+        );
+        expect(mockLogger.error).toHaveBeenCalledWith(
+          expect.objectContaining({ codebaseId: 'codebase-1', cwd: '/repos/test-repo' }),
+          'orchestrator.cwd_missing'
+        );
+      } finally {
+        mockExistsSync.mockImplementation(() => true);
+      }
+    });
+
+    test('guards conversation.cwd (worktree override), not just default_cwd', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({
+        codebase_id: 'codebase-1',
+        cwd: '/worktrees/cleaned-up',
+      });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      // The worktree is gone even though the codebase's own dir still exists.
+      mockExistsSync.mockImplementation((p: string) => p !== '/worktrees/cleaned-up');
+
+      try {
+        const platform = makePlatform();
+        await handleMessage(platform, 'conv-1', 'hello');
+
+        expect(mockSendQuery).not.toHaveBeenCalled();
+        expect(platform.sendMessage).toHaveBeenCalledWith(
+          'conv-1',
+          expect.stringContaining('Working directory no longer exists: /worktrees/cleaned-up')
+        );
+      } finally {
+        mockExistsSync.mockImplementation(() => true);
+      }
+    });
+
+    test('does not guard the unscoped path (workspaces root is created on demand)', async () => {
+      const conversation = makeConversation({ codebase_id: null });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([]));
+      // Even with nothing on disk, ensureArchonWorkspacesPath() mkdir -p's it,
+      // so the unscoped branch must stay reachable.
+      mockExistsSync.mockImplementation(() => false);
+
+      try {
+        const platform = makePlatform();
+        await handleMessage(platform, 'conv-1', 'hello');
+
+        expect(getSendQueryCwd()).toBe('/home/test/.archon/workspaces');
+      } finally {
+        mockExistsSync.mockImplementation(() => true);
+      }
+    });
   });
 });
 
--- a/packages/core/src/orchestrator/orchestrator-agent.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.ts
@@ -1383,6 +1383,24 @@ export async function handleMessage(
     let cwd: string;
     if (scopedCodebase !== undefined) {
       cwd = conversation.cwd ?? scopedCodebase.default_cwd;
+      // Fail fast when the resolved directory has been removed out from under a
+      // long-lived conversation (cleaned-up worktree, deleted repo dir, reset
+      // volume). Providers pass cwd straight through — Codex hands it to the
+      // native binary as `--cd`, which dies with an opaque
+      // "No such file or directory (os error 2)" (#1170). Mirrors the existsSync
+      // guard /register-project and /update-project already apply.
+      if (!existsSync(cwd)) {
+        getLog().error(
+          { conversationId, codebaseId: scopedCodebase.id, cwd },
+          'orchestrator.cwd_missing'
+        );
+        throw new Error(
+          `Working directory no longer exists: ${cwd}\n\n` +
+            'The project folder or worktree may have been deleted or cleaned up ' +
+            '(e.g. `archon isolation cleanup`, `archon complete`, or a manual removal). ' +
+            'Re-register the project (`/register-project`) or select a different one (`/setproject`).'
+        );
+      }
     } else {
       if (conversation.codebase_id !== null) {
         getLog().warn(
--- a/packages/core/src/utils/error-formatter.test.ts
+++ b/packages/core/src/utils/error-formatter.test.ts
@@ -177,6 +177,57 @@ describe('classifyAndFormatError', () => {
     });
   });
 
+  describe('missing working directory errors (#1170)', () => {
+    test('passes the message through with its recovery guidance intact', () => {
+      const result = classifyAndFormatError(
+        new Error(
+          'Working directory no longer exists: /.archon/workspaces/owner/repo/source\n\n' +
+            'The project folder or worktree may have been deleted or cleaned up ' +
+            '(e.g. `archon isolation cleanup`, `archon complete`, or a manual removal). ' +
+            'Re-register the project (`/register-project`) or select a different one (`/setproject`).'
+        )
+      );
+      expect(result).toContain('Working directory no longer exists');
+      expect(result).toContain('/.archon/workspaces/owner/repo/source');
+      expect(result).toContain('/register-project');
+      expect(result).toContain('/setproject');
+    });
+
+    test('survives the >100-char generic fallback that would otherwise hide the guidance', () => {
+      // The real thrown message is well over 100 chars, so without a dedicated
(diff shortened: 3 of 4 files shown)


Wirasm:
Hi @waaaw, thanks for this, and sorry it sat a while. I picked it up as maintainer. Short version: the bug you found is real and is still unfixed on `dev` today, but a second PR has since appeared on the same seam and is further along, so I want to consolidate rather than land both. Details below, including the one case your PR covers that the other one does not.

## The defect is still live

Verified against `dev` at `ea7b85db`. `orchestrator-agent.ts:1712` still resolves `cwd = conversation.cwd ?? scopedCodebase.default_cwd` with no existence check, and nothing between there and the provider handoff consults the filesystem. A probe test on current `dev` confirms a scoped chat hands a missing path straight to `sendQuery`, and no message ever names the directory.

Your root-cause reasoning holds up too. `posix_spawn` reports a missing *working directory* as ENOENT against the *executable's* path:

```
```

That is why Codex surfaces `No such file or directory (os error 2)` and why the Claude SDK concludes a libc mismatch. The error never names the thing that is actually wrong.

## Two things changed under this PR

Issue #1170 was closed on 2026-07-20 as fixed by `ensureArchonWorkspacesPath`, which creates the workspaces directory up front. That closed the *unscoped* half the reporter hit in Docker. Your PR deliberately left the unscoped branch alone and fixed the *scoped* branch, which is a different gap and still open. So the work is not obsolete, but the `Closes #1170` link no longer describes what it does.

PR #2551 now covers the same seam plus two more layers: the write site in `cleanup-service.ts` that leaves `cwd` pointing at a directory it is about to delete, and the Claude provider's error text. It is out of draft, green on all three CI jobs, and has a maintainer review with a merge path. Landing both would put two guards on one seam and conflict in `orchestrator-agent.test.ts`.

## One finding worth recording

Your guard sits at the cwd-resolution site, line 1712 on current `dev`. The user-message persist is at line 1574, and its own comment states the invariant it exists to protect: only AI-bound turns get a user row, so a `user` row always pairs with an `assistant` row. `AGENTS.md` documents the same invariant for `remote_agent_messages`.

I applied your guard verbatim to current `dev` and measured it. On a non-web platform the user row is written, then the guard throws, so `sendQuery` is never called:

```
expect(userRows.length).toBe(1);           // passes
expect(mockSendQuery).toHaveBeenCalled();  // fails: 0 calls
```

That is an orphaned user row on every refusal. It is a placement problem rather than a logic problem: the same check one block earlier, before the persist, has no such effect. #2551 places its guard there for exactly this reason.

The other half of the design, throwing a prose message and then adding a `startsWith` branch to `classifyAndFormatError` so the formatter does not swallow it, is a genuine requirement of the throw-based approach, since the message is over the 100-char generic-fallback limit. Sending the message and returning removes the need for that branch entirely.

## What your PR covers that #2551 does not

#2551's guard is `conversation.cwd !== null`, and it deliberately does not consult `codebase.default_cwd`. So a scoped chat with no worktree override, against a registered project whose directory has since been deleted, is not caught at the orchestrator. #2551 catches it one layer down, but that fix lives in `packages/providers/src/claude/provider.ts`, so Codex, Pi, OpenCode and Copilot still get the opaque error. That is the Codex-in-Docker shape from #1170, and your PR is the only one of the two that covers it.

## Where this leaves us

I am not closing this. My recommendation to the maintainers is to land #2551 first, then fold the `default_cwd` slice in as a follow-up, credited to you for finding it.

If you would like to write that follow-up yourself, it is small and well scoped and I would be glad to review it: guard the resolved `default_cwd` before the persist at line 1574, send and return rather than throw, and drop the error-formatter branch. Rebasing this branch as-is would need one conflict resolved in `error-formatter.ts`, since `dev` rewrote the rate-limit block your hunk anchors on, plus the guard move.

No pressure either way, and no rush. Thanks for the careful writeup: ruling out the Dockerfile node/npm theory in the PR body saved the next person real time, and it turned out to be right.

Wirasm:
Closing this in favour of #2551, which the maintainers have chosen as the vehicle for the fix. That decision is about consolidating two PRs on one seam, not about the quality of what you found here.

To be plain about what your contribution established:

- **The defect is real, and you found it first.** Verified live on `dev` at `ea7b85db` — `orchestrator-agent.ts:1712` still resolves `conversation.cwd ?? scopedCodebase.default_cwd` with no existence check, a month after you reported it.
- **Your diagnosis was correct**, including the part that took the most work: ruling out the Dockerfile node/npm theory carried in the issue. The SDK does spawn the native binary directly. I also reproduced the underlying mechanism independently — `posix_spawn` reports a missing *working directory* as ENOENT against the *executable's* path, which is precisely why the error never names the real problem.
- **The orphaned `user` row I flagged was a placement issue, not a diagnostic one.** Your guard was the right check; it just sat after the message persist at `:1574` rather than before it. The same check one block earlier has no such effect. That is a small move, not a rethink.

## The slice that is still yours

#2551's guard is `conversation.cwd !== null` and deliberately does not consult `default_cwd`. So the case your PR uniquely covered — a scoped chat with no worktree override, against a project directory that has since been deleted — remains unguarded at the orchestrator. #2551 catches it one layer down, but only for Claude; Codex, Pi, OpenCode and Copilot still get the opaque error. That is the Codex-in-Docker shape you originally set out to fix.

I have recommended that land as a follow-up once #2551 merges, credited to you. **If you want to write it, it is yours** — say so here and I will make sure it is routed to you. It is a contained change and I would be glad to review it. If you would rather not, that is completely fine; someone else will pick it up, and the credit stays with you either way.

## One bookkeeping note, so nothing falls through

This PR says `Closes #1170`, but #1170 was closed on 2026-07-20 as fixed by `ensureArchonWorkspacesPath`, which addressed the *unscoped* half the reporter hit in Docker. The scoped gap you found was never what that issue actually tracked, so closing this PR would otherwise leave it tracked by nothing. The `conversation.cwd` half is now carried by #2551, and I have proposed to the maintainers that the `default_cwd` half get its own tracking issue rather than living in someone's memory.

Your branch is untouched. Nothing was pushed to it and it will not be deleted, so `980b763f` stays exactly where you left it.

Thanks for a genuinely careful first contribution. The writeup was better than most, and the investigation that ruled out the wrong root cause is the part that saved the next person the most time.

PR #2126, opened 2026-07-15
fix(workflows): interactive-loop completion — signal_completes, finalize-on-bare-approve, honest + agent-steerable gates

Wirasm:
## Summary

- Problem: an `interactive: true` loop (`loop:`/`loop_group:`) whose prompt emits its completion signal can never terminate — the executor deliberately ignores the signal on the first iteration, approve always launches a fresh iteration (re-running the whole expensive body), and the static `gate_message` can misreport the actual iteration outcome (#2074, reproduced on a real 24-minute Jetson validation loop).
- Why it matters: pass-through-on-success is inexpressible; a passed validation loops forever, re-doing the work on every approve, behind a gate message that lies.
- What changed: four coupled changes so the engine remembers and respects the completion signal — **(B)** opt-in `signal_completes: true` completes the node on the signal with no gate; **(C)** a signal-bearing gate persists `completionSignaled`/`signaledOutput` in pause metadata, and a **bare approve finalizes at resume** from that output (no re-run) while approve-with-feedback still iterates; **(D)** the gate message gets an engine-generated honest status line (signal yes/no + bounded output excerpt) and structured `completionSignaled` on the `approval_requested` event; **(E)** gates are steerable by an AI approver — `manage_run get` surfaces the structured gate state with a finalize hint, `manage_run approve` gains `accept: true`, CLI `workflow get` prints a Gate line (`--json` already carries the metadata), and the console ApprovalPanel shows "Accept & complete".
- What did **not** change (scope boundary): the pause-input transport (#1991), session-continuity design (#2004/#2046), the `--json` approve records-not-resumes contract, #2112 paused-and-staged semantics, no new approver node type, no console redesign beyond the label/placeholder hint. NL chat approval still always iterates (documented limitation, see below).

### Two user-facing semantics changes (please note)

1. **Bare approve on a signal-bearing gate now finalizes.** Previously `/workflow approve <id>` (no comment) on an interactive-loop gate always ran another iteration. Now, when the paused iteration emitted the completion signal, an approve with **no** comment completes the node from the already-computed output — no re-run. Approve **with** a comment still iterates; non-signaled gates iterate for both forms.
2. **`signal_completes: true` skips the first gate.** Opting in lets a first-iteration signal complete the node with no human confirmation at all (previously "always gate first" was unconditional). Default behavior without the flag is unchanged (gate first).

[...]

PR #2125, opened 2026-07-15
fix(providers): throw on Claude API errors surfaced as text instead of completing them

Wirasm:
## Summary

Describe this PR in 2-5 bullets:

- Problem: the Claude Code SDK does not throw on API-level failures (auth not configured, invalid key, billing, rate limit) — it synthesizes an assistant message whose content is the error prose ("Not logged in · Please run /login") plus a result with `subtype: 'success'` + `is_error: true`, the exact field pair the #1425 stop-sequence carve-out treats as a clean success (down to `stop_reason: 'stop_sequence'`). The provider yielded the error prose as successful output.
- Why it matters: silent-failure class — the DAG node completes green, error prose poisons downstream `$node.output`, side-effecting nodes can run on it, and the operator debugs the wrong node (#1797's dogfood run failed one node downstream with "Failed to extract issue number from: 'Not logged in · Please run /login'").
- What changed: `streamClaudeMessages` now detects API-failure-as-text **structurally** (zero text matching) and throws a typed `ClaudeApiResultError`; `classifyAndEnrichError` classifies it by the SDK's typed error code (auth codes non-retryable; `rate_limit`/`overloaded`/`server_error` reuse the existing backoff).
- What did **not** change (scope boundary): `packages/providers/src/claude/**` + tests only. No dag-executor/orchestrator changes (their error paths already fail nodes on thrown provider errors). Codex/Pi audit for the same class is a follow-up. No retry-policy changes.

## Detection design

Verbatim wire capture (claude CLI 2.1.210, isolated config dir, no credentials):

```json
{"type":"assistant","message":{"model":"<synthetic>","stop_reason":"stop_sequence",
 "content":[{"type":"text","text":"Not logged in · Please run /login"}],...},
 "error":"authentication_failed"}
{"type":"result","subtype":"success","is_error":true,"api_error_status":null,
 "result":"Not logged in · Please run /login","terminal_reason":"api_error",...}
```

Signals (SDK 0.3.209):

1. **Primary**: the typed wrapper `error: SDKAssistantMessageError` field on an assistant message with `message.model === '<synthetic>'` — SDK-written error prose, never model output. Both conditions required: a REAL model message can carry `error: 'max_output_tokens'` alongside genuine truncated output and must flow through.
2. **Confirmation**: `is_error: true` + `subtype: 'success'` on the terminal result — only then does the provider throw (two-signal agreement).
[...]

PR #2134, opened 2026-07-16
fix(workflows): wait for live background Agent tasks before completing a node

### LINKED ISSUE #2083: Workflow nodes complete while SDK background Agent tasks are still running — artifacts silently missing
## Summary

Since Claude Agent SDK 0.3.193 replaced the synchronous `Task` sub-agent tool with an asynchronous `Agent`/background-task system, a workflow prompt node that delegates work to sub-agents can have its node marked **complete** while the delegated work is still running in the background. Any files the node was instructed to write never appear, and downstream nodes read missing/empty artifacts with no error.

## Evidence (two independent runs)

Run `e3aad87d` (node `find-validate-products`): assistant says *"I'll spawn 5 parallel research agents — one per problem — then compile all findings into the required file"* → 5× `tool_name: "Agent"` calls → *"I'll collect all results and write the product-candidates.md file once they complete"* → `node_complete` fires immediately. Zero `Write` calls in the node window; `product-candidates.md` never created. Node "succeeded" in 48s.

Run `2fb30a9d` (node `find-products`): the model literally outputs *"I'll wait for their completion notifications before writing the file"* and `node_complete` fires immediately after that sentence (27s).

Logs: `~/.archon/workspaces/coleam00/Archon/logs/<run-id>.jsonl` for both runs.

## Root cause

`dag-executor.ts` treats a node as complete when the provider's `sendQuery` generator yields its `result` chunk (top-level turn end) — see the `for await` loop around `dag-executor.ts:925-926` and result handling at `dag-executor.ts:1075`. `task_started`/`task_progress`/`task_notification` chunks (`packages/providers/src/types.ts:239-264`) are only forwarded as `task_activity` SSE events for UI display (`dag-executor.ts:1208-1298`). Nothing polls `TaskList`/`TaskGet` or otherwise waits for outstanding background tasks before marking the node complete (grep of `packages/providers/src` and `packages/workflows/src` has zero hits).

## Impact

Silent data loss in any workflow whose node prompt triggers sub-agent delegation — the model does this spontaneously for 'research N things in parallel' prompts. The node reports success; the failure only surfaces when a downstream node tries to read the missing artifact (or worse, silently degrades, as observed — the report node 'worked around' the missing file and emitted unverified data).

## Suggested direction

Either (a) dag-executor tracks `task_started` events and defers `node_complete` until all spawned background tasks reach a terminal state (with the node's `idle_timeout` as backstop), or (b) workflow nodes disable the async Agent tool by default (sync-only delegation), or (c) at minimum a loud warning when a node completes with non-terminal background tasks outstanding.

Workaround used meanwhile: `denied_tools: [Task, Agent]` + prompt instruction to work sequentially and write files incrementally.

Related: the `denied_tools` name-validation gap that made the first workaround attempt a silent no-op (separate issue).

Wirasm:
## Summary

- Problem: since claude-agent-sdk 0.3.193 the `Task` tool became an async `Agent`/background-task system, so a `result` message only means "turn done" — not "all work done". dag-executor broke out of its stream loop at the first result, which cascades `.return()` down the generator chain → SDK `Query.cleanup()` → `transport.close()` → SIGTERM to the CLI subprocess ~2s later, killing live background tasks. The artifacts they were producing silently never appear (#2083, observed on two runs).
- Why it matters: silent data loss in any workflow node whose prompt triggers sub-agent delegation — the model does this spontaneously for "research N things in parallel" prompts; downstream nodes consume missing artifacts without error.
- What changed: the Claude provider now forwards the SDK's `background_tasks_changed` level signal (v0.3.209, previously dropped as an unhandled system subtype) as a new `background_tasks` MessageChunk; both dag-executor stream loops gate their break-on-result on the live-task set being empty and otherwise keep consuming until the tasks drain and the final result arrives (announced once with ⏳). Abnormal stream end with live tasks warns loudly and records `background_tasks_incomplete` in `node_completed` event data. `task_notification` events now persist `output_file` (the artifact pointer for delegated work).
- What did **not** change (scope boundary): non-Claude providers and pre-0.3.209 CLIs never emit the signal → byte-for-byte existing behavior. Direct chat (orchestrator) needs no change — it never breaks on result, its loop consumes the provider generator to natural exhaustion, which is exactly why task completions already arrive in chat today. #2125's synthetic-error logic and #2126's gate/loop-finalize regions are untouched.

### Before

```
  User                        Archon (dag-executor)          Claude CLI
  ────                        ─────────────────────          ──────────
  runs workflow ────────────▶ node streams
                              model spawns 5 bg agents ────▶ agents start
                              result chunk arrives
                              *break* → node_completed       agents SIGTERMed
  sees "node succeeded" ◀───  (48s)                          artifacts NEVER written
  downstream node reads missing file → silent garbage
```

### After

```
  User                        Archon (dag-executor)          Claude CLI
  ────                        ─────────────────────          ──────────
  runs workflow ────────────▶ node streams
                              model spawns 5 bg agents ────▶ agents start
                              [background_tasks {5 live}]
                              result chunk arrives
                              [set non-empty → KEEP READING]
  sees "⏳ waiting for 5      task_progress resets idle      agents finish,
   background task(s)" ◀────  timer (existing backstop)      write artifacts
                              [background_tasks {empty}]     follow-up turn integrates output
                              final result → node_completed
  sees "node succeeded" ◀───  artifacts present ✔
  (abnormal end: "⚠️ output may be missing" + background_tasks_incomplete on the event)
```

### Before

```
Claude SDK ──messages──▶ claude/provider.ts ──MessageChunks──▶ dag-executor
  background_tasks_changed ──x── (dropped: system_message_unhandled)
  result ────────────────────────────────────▶ break (unconditional)
```

### After

```
Claude SDK ──messages──▶ claude/provider.ts ──MessageChunks──▶ dag-executor
  background_tasks_changed ===▶ [+] background_tasks chunk ===▶ [+] createBackgroundTaskTracker
  result ────────────────────────────────────▶ [~] break gated on tracker.shouldBreakOnResult()
  task_notification.output_file ==============▶ [~] persisted on task_activity events
```

**Connection inventory**:

| claude/provider.ts | providers/types.ts (MessageChunk) | **modified** | new `background_tasks` variant emitted |
| dag-executor (AI-node loop) | provider stream | **modified** | conditional break; waits while tasks live |
| dag-executor (loop-node loop) | provider stream | **modified** | same gate; per-iteration usage folded once (no double-count with 2 results) |
| dag-executor | event-emitter / IWorkflowStore | **modified** | `outputFile` on task_activity; `background_tasks_incomplete` on node_completed data |
| orchestrator / adapters / web bridge | MessageChunk | unchanged | if/else chains ignore the new variant (verified; no exhaustive switches on chunk.type) |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `workflows`
- Module: `workflows:executor`, `providers:claude`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2083

```bash
                   # check:pi-vendor-map, type-check, lint, format:check, tests all pass
```

- Evidence provided: empirical probe against the real SDK 0.3.209 + installed Claude CLI — with a prompt spawning one background agent (sleep 20s then write done.txt), the stream shows `background_tasks_changed [1 task]` BEFORE the result; continuing to consume past the result keeps the subprocess alive, `background_tasks_changed []` arrives at +35s, a final result at +38.9s, and `done.txt` exists on disk. Breaking at the first result (today's behavior) kills the tasks and the file is never written. New unit tests pin: wait-then-drain output capture, abnormal-end incompleteness recording, unchanged break-at-result without the signal, loop-node cost no-double-count, provider translation incl. empty-set forwarding and post-result passthrough.
- Skipped: none.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? Yes — providers/CLIs that never emit `background_tasks_changed` keep the exact previous break-on-first-result behavior; `background_tasks_incomplete` is an additive optional field in free-form event data.
- Config/env changes? No

- Verified scenarios: SDK message-stream lifecycle probed live against SDK 0.3.209 (see Validation Evidence); full test suites for @archon/workflows and @archon/providers.
- Edge cases checked: empty-set drain signal forwarded (not dropped); cancellation mid-wait suppresses the warning (node fails with its own message); idle timeout mid-wait lands on the WARN+record path; per-iteration cost folding on both normal and error exits of the loop-node stream.
- What was not verified: a full end-to-end Archon workflow run with real delegating prompts (probe covered the SDK layer directly); Slack/Telegram rendering of the new ⏳/⚠️ messages (plain `sendMessage` strings, same path as existing warnings).

- Affected subsystems/workflows: workflow AI nodes and loop iterations on the Claude provider with CLI ≥0.3.209; nodes that delegate to background agents now run longer (they wait for the delegated work — that time was previously spent producing silently-lost artifacts) and include the follow-up turn's cost (billing now reflects work actually done).
- Potential unintended effects: a node with a genuinely wedged background task now waits until the idle timeout instead of completing incomplete — that is the intended fail-loud trade-off; the wait is announced in-stream and bounded by the existing `idle_timeout` (task_progress chunks reset it, ~30s cadence).
- Guardrails/monitoring: `dag.node_result_with_live_background_tasks` / `dag.node_stream_ended_with_live_background_tasks` (and loop_node equivalents) log events; `background_tasks_incomplete` on node_completed event data; ⏳/⚠️ user-facing messages.

## Rollback Plan (required)

- Fast rollback command/path: revert the single commit (`git revert 05e7ecc0`) — no schema/config coupling.
- Feature flags or config toggles: none; per-node `idle_timeout` bounds the wait.
- Observable failure symptoms: nodes hanging in "waiting for background tasks" until idle timeout (would indicate the SDK stopped emitting the drain signal); unexpected `background_tasks_incomplete` on clean runs.

## Risks and Mitigations

- Risk: the SDK changes `background_tasks_changed` semantics (e.g. stops emitting the empty-set drain signal), turning the wait into an idle-timeout stall.
  - Mitigation: the wait never hangs — generator exhaustion or idle timeout always ends it, and both land on the loud-warn + incompleteness-record path; the tracker is per stream pass so state cannot leak across retries.
- Risk: cost accounting drift in loop nodes now that one iteration can see two results.
  - Mitigation: last-seen fold (session-cumulative semantics) with a dedicated regression test asserting 0.3, not 0.4.

## Summary by CodeRabbit

* **New Features**
  * Workflow execution now waits for active background tasks to finish before completing a step or loop iteration.
  * Live background-task updates are reflected during streaming, including task status and completion.
  * Task activity records now include associated output file information.

* **Bug Fixes**
  * Prevented workflows from stopping prematurely when background tasks continue after an initial result.
  * Improved handling of incomplete tasks with user warnings and persisted task details.
  * Corrected usage accounting to avoid duplicate cost reporting.

### DIFF
--- a/packages/providers/src/claude/provider.test.ts
+++ b/packages/providers/src/claude/provider.test.ts
@@ -611,6 +611,87 @@ describe('ClaudeProvider', () => {
       expect(chunks[0]).toMatchObject({ type: 'task_notification', status: 'failed' });
     });
 
+    // --- #2083 — background-task liveness (SDK 0.3.209 background_tasks_changed) ---
+
+    test('yields background_tasks chunk from SDK background_tasks_changed', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield {
+          type: 'system',
+          subtype: 'background_tasks_changed',
+          tasks: [
+            { task_id: 't-1', task_type: 'local_agent', description: 'Research problem A' },
+            { task_id: 't-2', task_type: 'local_agent', description: 'Research problem B' },
+          ],
+        };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks).toEqual([
+        {
+          type: 'background_tasks',
+          tasks: [
+            { taskId: 't-1', taskType: 'local_agent', description: 'Research problem A' },
+            { taskId: 't-2', taskType: 'local_agent', description: 'Research problem B' },
+          ],
+        },
+      ]);
+    });
+
+    test('forwards an EMPTY background_tasks_changed set (drain signal)', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield { type: 'system', subtype: 'background_tasks_changed', tasks: [] };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      // An empty set means "all background work drained" — it must be forwarded,
+      // not dropped, or the executor's wait gate would never release.
+      expect(chunks).toEqual([{ type: 'background_tasks', tasks: [] }]);
+    });
+
+    test('keeps forwarding chunks that arrive AFTER the result (background-task wait window)', async () => {
+      // Single-turn queries keep streaming after the turn-level result while
+      // background tasks drain; streamClaudeMessages must not stop at result.
+      mockQuery.mockImplementation(async function* () {
+        yield {
+          type: 'system',
+          subtype: 'background_tasks_changed',
+          tasks: [{ task_id: 't-1', task_type: 'local_agent', description: 'bg work' }],
+        };
+        yield { type: 'result', subtype: 'success', session_id: 's-1', is_error: false };
+        yield {
+          type: 'system',
+          subtype: 'task_notification',
+          task_id: 't-1',
+          status: 'completed',
+          output_file: '/tmp/t-1.md',
+          summary: 'done',
+        };
+        yield { type: 'system', subtype: 'background_tasks_changed', tasks: [] };
+        yield { type: 'result', subtype: 'success', session_id: 's-1', is_error: false };
+      });
+
+      const types = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        types.push(chunk.type);
+      }
+
+      expect(types).toEqual([
+        'background_tasks',
+        'result',
+        'task_notification',
+        'background_tasks',
+        'result',
+      ]);
+    });
+
     test('yields hook_started chunk from SDK system message', async () => {
       mockQuery.mockImplementation(async function* () {
         yield {
--- a/packages/providers/src/claude/provider.ts
+++ b/packages/providers/src/claude/provider.ts
@@ -782,6 +782,8 @@ async function* streamClaudeMessages(
         status?: string;
         output_file?: string;
         skip_transcript?: boolean;
+        // Background-task set (Claude SDK v0.3.209+ `background_tasks_changed`)
+        tasks?: { task_id: string; task_type: string; description: string }[];
         // Hook lifecycle (Claude SDK v0.2.89+)
         hook_id?: string;
         hook_name?: string;
@@ -846,6 +848,20 @@ async function* streamClaudeMessages(
           outputFile: sysMsg.output_file ?? '',
           ...(sysMsg.usage !== undefined ? { usage: sysMsg.usage } : {}),
           ...(sysMsg.tool_use_id !== undefined ? { toolUseId: sysMsg.tool_use_id } : {}),
+        };
+      } else if (subtype === 'background_tasks_changed') {
+        // Level signal: the FULL set of live background tasks after a membership
+        // change (REPLACE semantics — see the MessageChunk variant docs). An
+        // empty `tasks` array is meaningful ("all drained") and MUST be
+        // forwarded, so no `&& sysMsg.tasks` guard here.
+        const tasks = Array.isArray(sysMsg.tasks) ? sysMsg.tasks : [];
+        yield {
+          type: 'background_tasks',
+          tasks: tasks.map(t => ({
+            taskId: t.task_id,
+            taskType: t.task_type,
+            description: t.description,
+          })),
         };
       } else if (subtype === 'hook_started' && sysMsg.hook_id) {
         yield {
--- a/packages/providers/src/types.ts
+++ b/packages/providers/src/types.ts
@@ -262,6 +262,17 @@ export type MessageChunk =
       usage?: { total_tokens: number; tool_uses: number; duration_ms: number };
       toolUseId?: string;
     }
+  // Forwarded from SDKBackgroundTasksChangedMessage (`background_tasks_changed`,
+  // Claude SDK v0.3.209+): the FULL set of live background tasks, emitted
+  // whenever membership changes. Level signal with REPLACE semantics — consumers
+  // swap their set for each payload (an empty array means no background work is
+  // running). The dag-executor gates node completion on this: a `result` chunk
+  // that arrives while the set is non-empty must not tear down the stream, or
+  // the SDK subprocess (and the tasks' pending artifacts) get killed (#2083).
+  | {
+      type: 'background_tasks';
+      tasks: { taskId: string; taskType: string; description: string }[];
+    }
   // ─── Hook Lifecycle (Claude SDK `system` subtypes) ─────────────────────
   // Forwarded by the Claude provider from SDKHookStartedMessage /
   // SDKHookResponseMessage. Same aggregation path as task_* above; the bridge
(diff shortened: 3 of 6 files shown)


Wirasm:
## 🔍 Self-Review

### Summary

The fix addresses the verified root cause (unconditional break-on-result SIGTERMs the SDK subprocess and kills live background tasks) with the smallest sound mechanism: gate the break on the SDK's own liveness level-signal, wait bounded by existing timeout machinery, and fail loud when the wait cannot complete. Full `bun run validate` passes (all eight checks).

### Findings

#### ✅ Strengths
- Gating uses ONLY the `background_tasks_changed` level signal (REPLACE semantics) — no hand-paired `task_started`/`task_notification` edge tracking, which the SDK docs explicitly warn can wedge a stale running indicator on a missed bookend.
- The tracker is instantiated per stream pass (`runStreamPass` / per loop iteration), so no liveness state can leak across provider retries or structured-output reasks — and the wait can never hang: generator exhaustion or idle timeout always ends the loop, both landing on the loud-warn + `background_tasks_incomplete` record path.
- Empty-set forwarding is explicitly tested at the provider boundary (`provider.test.ts` "forwards an EMPTY background_tasks_changed set") — dropping it would deadlock the gate, so it is pinned.
- Loop-node usage fold runs exactly once on both the normal and the error exit (`foldIterationUsage` guard), preserving the old failure-path cost reporting while fixing the new two-results-per-iteration double-count (regression test asserts 0.3, not 0.4).
- No behavior change without the signal: providers/CLIs that never emit it keep byte-for-byte break-on-first-result semantics (pinned by the "breaks at the first result when no background_tasks chunk was seen" test).

#### ⚠️ Suggestions (non-blocking)
- `dag-executor.ts` (result handler): the `dag.node_result_with_live_background_tasks` WARN fires on every mid-wait result. In practice that is 1-2 lines per wait window; if the SDK ever emits per-task interim results this could get chatty — trivially throttleable later.
- `dag-executor.ts` (loop-node abnormal-end check): the `cancelled` exemption mirrors the AI-node loop for symmetry, but the loop-node stream has no mid-stream cancel-abort today (cancel checks run between iterations), so the flag is defensive rather than reachable. Harmless.
- Idle timeout mid-wait produces both the existing "completed via idle timeout" notice and the new "output may be missing" warning — mildly redundant, but each states a distinct, accurate fact.

#### 🧹 Simplification
- Considered folding `count()`/`ids()` into a single accessor returning the map — kept the narrow explicit interface instead (call sites read clearer). No further simplifications identified.

#### 🔒 Security
- No security concerns: no new permissions, network calls, secrets handling, or filesystem scope. Task ids/descriptions logged are SDK-generated metadata, not user content.

### Checklist

- [x] Fix addresses root cause from investigation (empirically verified against SDK 0.3.209 — see the issue comment for the probe timeline)
- [x] Code follows codebase patterns (system-subtype translation chain, safeSendMessage warnings, fire-and-forget event persistence, fail-fast on error results)
- [x] Tests cover the change (6 new: wait-then-drain output capture, abnormal-end incompleteness, unchanged no-signal break, loop cost no-double-count, provider translation, empty-set + post-result passthrough)
- [x] No obvious bugs introduced (`bun run validate` exit 0)

---
*Self-reviewed • Ready for human review*

Issue #2121, opened 2026-07-15
feat(workflows): sub-workflow composition — include: (load-time inlining) and workflow: (runtime sub-run) primitives


Wirasm:
## Problem — proven by audit

Composability currently stops at the command file: a workflow cannot reuse another workflow's DAG structure. The 2026-07-15 defaults audit quantified the consequence:

- The **9-node review block** (scope → sync → 5 review agents → synthesize → implement-fixes) is copy-pasted verbatim into 4 bundled workflows (idea-to-pr, plan-to-pr, issue-review-full, comprehensive-pr-review) and gated in a 5th (smart-pr-review)
- `verify-pr-base` is a byte-identical 15-line bash node in up to 9 workflows
- Clone workflows exist BECAUSE extraction is impossible: plan-to-pr is 17/18 nodes identical to idea-to-pr

Every copy drifts independently. This is the load-bearing composability gap.

### Phase 1: `include:` — load-time inlining (recommended first)

A node that references another workflow file; the loader expands its nodes into the parent DAG at parse time.

```yaml
- id: review
  include: archon-review-block        # resolved like command:/workflow discovery
  depends_on: [create-pr]
  with:                               # optional input mapping → variables inside the included DAG
    pr_ref: $create-pr.output.number
```

Semantics (reuse the loop_group body precedent deliberately):
- Included nodes are a **sealed sub-DAG** namespaced `<includeId>.<nodeId>` (id-shadowing check + event step_names already exist for loop_group bodies — #2090 shipped the namespacing plumbing)
- From the parent's perspective the include behaves like one node: parent `depends_on`/`when`/`trigger_rule` attach to it as a whole; `$includeId.output` = the included DAG's terminal-node output (same selection rule as loop_group)
- Load-time expansion means ZERO new runtime machinery: one run record, existing resume/skip/events/approval behavior all work unchanged
- Validation: cycle/uniqueness checks recurse (loader already does this for loop_group); include depth capped (e.g. 3); no recursive self-include
- Discovery/precedence identical to workflows today (bundled < global < project)

This alone deletes the review-block duplication: extract it once as `archon-review-block.yaml`, include it from 5 workflows.

### Phase 2: `workflow:` — runtime sub-run invocation (design with #1764)

A node that starts a **child workflow run**: own run record, own artifacts dir, own approval gates, parent consumes the child's terminal output. True 'run a workflow inside a workflow'.

[...]

Issue #1960, opened 2026-06-11
feat(workflows): surface the parent/child run tree in CLI, API, and console


Wirasm:
## Current state

The parent/child linkage is already delivered:

- `remote_agent_workflow_runs.parent_run_id` exists in PostgreSQL and SQLite.
- `WorkflowRun` carries `parent_run_id`.
- `workflow:` and fan-out child runs persist the edge.
- `findChildRuns(parentRunId)` exists and is used by engine lifecycle operations.
- The console recent-run row can show that a run is a sub-run.

The missing product outcome is navigation. An operator still cannot open a parent and inspect its children as a run tree through the CLI, API, or run-detail console.

### CLI

- `archon workflow runs --parent <run-id>` lists direct children.
- `archon workflow get <run-id>` includes direct child summaries in human and JSON output.
- Child summaries include id, workflow name, parent node/index where available, status, start/completion timestamps, and measured usage already present on the run.

### API

- Workflow-run listing supports an explicit parent filter.
- Run detail includes structured direct-child summaries.
- The response remains structured; clients own presentation.
- Pagination or bounded loading must not silently imply that a partial child list is complete.

### Console

- Run detail shows spawned runs grouped under the parent node where `parent_node_id` is available.
- Selecting a child opens that run.
- A child view links back to its parent.
- Nested children can be traversed without fetching an unbounded tree in one request.

## Invariants

- Do not infer lifecycle coupling from the edge. Parent and child statuses remain independently truthful.
- Do not automatically cancel, fail, or abandon children when a parent changes state.
- Existing lifecycle operations may offer explicit child actions, but ambiguous live work remains visible for operator judgment.
- Use the existing `parent_run_id`, `parent_node_id`, `child_index`, and `findChildRuns` primitives. Do not add another linkage model.
- Existing top-level run listings remain compatible.

## Sequencing

Run after #2407 so the affected core/server/CLI tests are under strict type-checking. This work is independent of #2451; #2451 adds per-slot fan-out outcomes, while this issue exposes child runs that actually exist.

## Acceptance

- [ ] CLI parent filter and child summaries.
- [ ] API parent filter and structured direct-child summaries.
- [ ] Console parent/child navigation on run detail.
- [ ] Parent and child links work across at least two nesting levels.
- [ ] Empty, missing, terminal, and live-child cases are covered.
- [ ] No automatic cross-run lifecycle mutation is introduced.

Wirasm:
[...]

Issue #1785, opened 2026-05-27
feat: input parameters enable node and command decoupling


aarononeal:
## Problem

Nodes (and commands) can refer to an upstream node's output using `$nodeId.output.fieldName`. However, this couples nodes together requiring a node to have knowledge of its place within workflow structure. This significantly impacts the ability to write reusable commands across workflows.

With fresh sessions being required for a number of scenarios, some form of context passing is still necessary.

Alternatively one can use `$ARTIFACTS_DIR` for passing state, but using this to transfer simple parameters like an issue number or repo URL becomes tedious.

## Proposed Solution

A two-pronged approach covering both sequential chains and explicit DAG binding:

1. **`$PREVIOUS.output`** — Resolves to the output of the immediately preceding sequential node. No node naming required. Ideal for linear chains.

2. **`input:` node field** — Allows the workflow to explicitly bind upstream outputs to named parameters. The command then references `$input.<name>`. Works with fan-out/fan-in and multi-dependency nodes.

### Before (current)

```yaml
nodes:
  - id: extract-repo-issue
    prompt: |
      Parse the request and output JSON...

  - id: prepare-fork-repo
    bash: |
      echo "$extract-repo-issue.output.repo" > "$ARTIFACTS_DIR/..."

  - id: fork-repo
    command: fork-repo
    depends_on: [prepare-fork-repo]

  - id: prepare-fetch-issue
    bash: |
      # Must know the exact upstream node names
      echo "$extract-repo-issue.output.repo" > "$ARTIFACTS_DIR/..."
      echo "$fork-repo.output.fork_name" > "$ARTIFACTS_DIR/..."

  - id: fetch-upstream-issue
    command: fetch-issue
    depends_on: [fork-repo, prepare-fetch-issue]
```

Pain points:
- Every command node must know the exact IDs of upstream nodes
- Renaming a node breaks all downstream references
- Commands are not reusable without knowing the workflow structure
- Artifact files proliferate for simple parameter passing

### After (proposed)

Sequential chain (uses `$PREVIOUS.output`):
```yaml
nodes:
  - id: extract-repo-issue
    prompt: |
      Parse the request and output JSON...

  - id: fork-repo
    command: fork-repo
    depends_on: [extract-repo-issue]
    # Command reads: $PREVIOUS.output (resolves to extract-repo-issue's output)

  - id: fetch-upstream-issue
    command: fetch-issue
    depends_on: [fork-repo]
[...]

Issue #1764, opened 2026-05-25
feat(workflows): compose statically named workflow files inside the parent run


Wirasm:
> **Body rewritten 2026-08-13.** The earlier body described `workflow:` as a separately governed child run. That implementation has largely shipped, but it is not the desired meaning of ordinary workflow composition. This issue now owns the replacement model below. Historical child-run data and resumable runs must remain intact.

> **Amended 2026-08-17.** #2448 is folded in — see *Scope → Shared artifact store*.
>
> **Approach decided the same day, and it narrows this issue considerably: `include:` IS the composition primitive, and `workflow:` stays the explicit governed launch, unchanged.** `include:` already resolves and validates before execution, namespaces deterministically, writes events under the parent run id, makes inner gates ordinary parent gates, shares the parent `$ARTIFACTS_DIR`/state/cwd/identity, accounts once, creates no child row, never infers isolation, and re-expands deterministically so resume matches persisted step names byte-for-byte. `dag-executor.test.ts:17055-17063` already tests exactly that, under a header calling itself *"the 'zero new runtime machinery' proof"*. So the parent-run execution model below is a thing to **name**, not to build — three acceptance items fail today, and they are the work.
>
> **Two consequences for readers of the scope below.** There is **no execution-model version gate to build** and **no child-run migration**: nothing changes meaning, so no live or historical run is reinterpreted. Where the scope asks for those, they are satisfied vacuously. Also, `fan_out:` stays on the launch path — #2512 gives the composition primitive runtime width, and its real precedent is `loop_group`, recorded there.
>
> The three hazards recorded in an earlier version of this note were premised on the rejected approach and have been corrected in place: the `evidence.json` collision cannot occur (a block's `evidence_policy` is dropped with a WARN), `persist_session` does not re-key (composed node identity is already stable), and the two `metadata` defects belong to the launch path this issue no longer touches.

## Plain version

A statically named workflow used from another workflow is composition, not a new run.

Crossing a workflow-file boundary must not create a second governance object. The composed nodes execute as part of the parent and share its run identity, artifacts, events, gates, lifecycle, working path, cancellation, resume, and accounting.

Declared inputs and outputs pass the baton across the boundary. The shared artifact store complements that contract for files, evidence, and large results; callers must not have to discover artifacts to learn the workflow result.

[...]

PR #2169, opened 2026-07-17
feat(workflows): workflow: runtime sub-run node (#2121 Phase 2 slice 1)

Wirasm:
## Summary

- **Problem:** Composition across *runs* is impossible today. `include:` (#2129) reuses another workflow's DAG by inlining its nodes into the parent's single run record — shared artifacts, shared cost, shared resume, no independent gate. There's no way to give a reused sub-pipeline its own run record, its own approval gate, or its own cost line.
- **Why it matters:** This is the constitutional "governance object" answer `include:` structurally can't provide. It lets authors compose the canonical "plan → implement → QA → PR" pipeline out of independently-governed, separately-auditable runs instead of copy-pasting node blocks or hand-rolling `archon workflow run --detach` + polling.
- **What changed:** A new `workflow:` DAG node runs another workflow as a **child sub-run** — its own `workflow_runs` row (`parent_run_id`), its own `$ARTIFACTS_DIR`, its own gates, its own cost line, its own audit trail. The child's terminal output threads back as `$<nodeId>.output`. Gates pause the whole tree (child pauses → parent pauses "blocked on child" → reviewer approves the child by run id → parent auto-resumes in-process on child completion).
- **What did NOT change (scope boundary):** Slice 1 is *sequential composition in a shared checkout*. No dynamic fan-out (#977), no `isolation: worktree` per child, no `first_success` racing (#1764), no `with:` named-parameter mapping (#1785) — all reserved in the schema and rejected fail-fast, so slice 2 is purely additive. No hard subprocess kill on cancel (cooperative only).

### Before

```
Author wants a reusable "plan -> implement -> QA" sub-pipeline with its own gate + audit line.

  workflow A -- copy-paste 9 nodes --> workflow B      (drift on every edit)
       |  OR
       v
  bash node: `archon workflow run child --detach`  -->  hand-rolled poll loop for the join

  include: inlines B's nodes into A's ONE run -- shared artifacts/cost/resume, no independent gate.
```

### After

```
  +-------------------- parent run (id P) --------------------+
  |  create-plan -> [ workflow: implement-qa ] -> finalize-pr |
  |                        |  input: "$create-plan.output"    |
  |                        v                                  |
  |       +---- child run (id C, *parent_run_id = P*) ---+    |
  |       |  scaffold -> [approval gate] -> implement    |    |
  |       |   (pauses C; *reviewer approves C by run id*;|    |
  |       |    C resumes; C completes)                   |    |
  |       +----------------------------------------------+    |
  |                        |  C.summary (terminal output)     |
  |                        v                                  |
  |              *$implement-qa.output* -> finalize-pr reads it|
  +-----------------------------------------------------------+
  TWO run records. When C pauses, P pauses "blocked on C". When C completes, P *auto-resumes*.
```

### Before

```
executor.ts --> dag-executor.ts (runLayers dispatch: command/prompt/bash/script/loop/approval/cancel/include)
   |                    |
   v                    v
core/db/workflows.ts   include-expander.ts (load-time flatten)
```

### After

```
executor.ts ==[+runChildWorkflow closure]==> dag-executor.ts
   |  |                                          |  [~runLayers: +isWorkflowNode dispatch]
   |  |  [+maybeResumeParentRun hook]            v
   |  +==============================> [+executeWorkflowNode] (re-entry table, gate-aware pause)
   v                                             |
core/db/workflows.ts [~createWorkflowRun +parent_run_id]
   [+findChildRuns] [+getRunAncestry] [~getActiveWorkflowRunByPath +excludeRunIds]
core/operations/workflow-operations.ts [~abandonWorkflow: +cascade cancel]
web/experiments/console [~Run primitive + RecentRunRow: +child badge]
```

**Connection inventory:**

| executor.ts | dag-executor.ts (`executeWorkflowNode`) | **new** | `runChildWorkflow` closure passed as a param (no static import cycle) |
| executor.ts (`runChildWorkflow`) | executor.ts (`executeWorkflow`) | **new** | in-process recursion for the child |
| dag-executor.ts (child terminal) | executor.ts (`maybeResumeParentRun`) | **new** | `parent_run_id`-keyed auto-resume of the parent |
| executeWorkflow | `getActiveWorkflowRunByPath` | **modified** | ancestor-exclusion for the shared-checkout path-lock |
| abandonWorkflow | `findChildRuns` / `cancelWorkflowRun` | **new** | recursive descendant cancel cascade |
| web console | `parent_run_id` | **new** | child-of-parent affordance |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: L`
- Scope: `workflows`, `core`, `web`, `docs`, `tests`
- Module: `workflows:executor`, `workflows:dag-executor`, `core:db`, `core:operations`

## Change Metadata

- Change type: `feature`
- Primary scope: `multi` (workflows + core, with web + docs)

## Linked Issue

- Closes # _(none — #2121 was closed by Phase 1)_
- Related #2121 (Phase 2 slice 1), Related #1764, Related #1960
- Depends on # _(none)_
- Supersedes # _(none)_

```bash
bun run validate   # exit 0 -- all eight checks green
```

- Evidence: `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, type-check (10 packages), lint (`--max-warnings 0`), `format:check`, and all per-package tests pass. New tests: `subrun.test.ts` (4, full acceptance e2e), `loader.test.ts` (+8 schema/loader rejections), `workflow-operations.test.ts` (+2 cascade), `sqlite.test.ts` (+1 column parity).
- Skipped: none.

## Security Impact (required)

- Secrets/tokens handling changed? **No** (the child inherits the parent's resolved env/credentials via the existing per-run machinery)
- File system access scope changed? **No** (child runs in the parent's checkout)

## Compatibility / Migration

- Backward compatible? **Yes** — `workflow:` is purely additive; existing workflows unaffected.
- Config/env changes? **No**
- Database migration needed? **Yes — a new `parent_run_id` column** on `remote_agent_workflow_runs` (self-referential FK, `ON DELETE SET NULL`).
  - **SQLite:** auto-applied on startup (runtime `ALTER TABLE` in `sqlite.ts` migrateColumns) + present in fresh `createSchema`.
  - **Postgres:** idempotent `ADD COLUMN IF NOT EXISTS` in `migrations/000_combined.sql` (applied under the advisory-lock transaction on connect).
  - Nullable, so existing rows converge with no backfill. Column-level parity asserted in `sqlite.test.ts`.

- Verified scenarios: the full acceptance e2e (`subrun.test.ts`) drives the REAL executor recursion with a stateful in-memory store — child gate → parent pauses `child_workflow` (no `node_completed`) → approve child by run id → child resumes → `parent_run_id` hook auto-resumes the parent → output threads to `$node.output` → parent completes. Also: gateless-child synchronous happy path (output + cost roll-up + `parent_run_id`/`parent_node_id` link), child-failure → node-failure → parent-failure, runtime self-reference cycle guard.
- Edge cases checked: ancestor path-lock exclusion (the child does not self-block against its running/paused parent — proven load-bearing by the e2e), `node_completed` written ONLY on true completion (re-run on resume vs. skip when done), abandon cascade over grandchildren skipping terminal nodes, all load-time rejections (`with:`/`retry:`/`isolation:'worktree'`/`workflow:`-in-loop_group-body/unknown-input-ref).
- What was not verified: a live AI end-to-end run of the bundled demo (prompt nodes need credentials); the console badge was type-checked + built, not visually screenshotted.

- Affected subsystems: the workflow executor (new node type + a new resume trigger), the run lifecycle DB layer (new column + finders), the abandon operation, and the console run list.
- Potential unintended effects: the `parent_run_id`-keyed auto-resume hook runs on EVERY terminal run — guarded to a strict no-op unless the run has a `parent_run_id` AND its parent is `paused` on a matching `child_workflow` gate. Resume-pass cost accounting for a paused-then-resumed parent reflects the resume pass (a pre-existing resume-accounting limitation, not new to this PR).
- Guardrails/monitoring: `workflow.parent_auto_resume_started` / `_failed` log events; the hook is wrapped in `.catch` so a parent-resume failure can never corrupt the child's own result.

## Rollback Plan (required)

- Fast rollback: revert the six commits on this branch (or `git revert` the range). The `parent_run_id` column is nullable and unreferenced by existing code paths, so it can stay in place harmlessly after a revert (no down-migration needed).
- Feature flags: none — the feature is inert unless a workflow declares a `workflow:` node.
- Observable failure symptoms: a `workflow:` node failing with a sub-run error; a parent stuck `paused` on a `child_workflow` gate that never auto-resumes (would show `workflow.parent_auto_resume_*` logs).

## Risks and Mitigations

- Risk: **Child self-blocked by the parent's path-lock** (shared checkout).
  - Mitigation: `getActiveWorkflowRunByPath` excludes the ancestor chain, derived from the run's own `parent_run_id` so it holds on both the initial spawn AND the resume-after-approve path. Explicitly exercised by the e2e (the child wouldn't complete otherwise).
- Risk: **Paused `workflow:` node wrongly recorded completed → skipped on resume.**
  - Mitigation: mirrors `executeApprovalNode` — returns `{completed, ''}` but writes NO `node_completed` on pause; writes it ONLY on true child completion. Verified against `getCompletedDagNodeOutputs`.
- Risk: **Column added to only one schema dialect.**
  - Mitigation: added to BOTH `sqlite.ts` and `000_combined.sql` + a column-level parity assertion in `sqlite.test.ts` + `generate:bundled-schema`.
- Risk: **Auto-resume-parent double-fires on the synchronous path.**
  - Mitigation: guarded on `parent.status === 'paused' && approval.type === 'child_workflow' && childRunId` match — a no-op on the synchronous first run and for a different child of the same parent.

## Review-fix addendum (2026-07-20)

- All multi-agent review findings addressed or deferred-with-rationale — see the [disposition comment](https://github.com/coleam00/Archon/pull/2169#issuecomment-5022225226). Two additional bugs found while fixing: an older-wins tiebreaker placeholder bug in `getActiveWorkflowRunByPath` (exposed by the new `excludeRunIds` SQL test), and the web abandon route bypassing the shared op (skipping cascade cancel + container reclaim).
- Note on `packages/web/src/lib/api.generated.d.ts`: the regeneration also sweeps in stale-generation drift from earlier merged PRs (`container`, `pi`, `signal_completes`, `include` fields) — those are not changes made by this PR, just the generator catching up.

## Summary by CodeRabbit

* **New Features**
  * Added `workflow:` DAG nodes to run governed child workflows as sub-runs with their own approvals, outputs, costs, and audit history.
  * Implemented run-tree linkage to track parent/child workflow runs and enable bounded child cancellation and ancestry traversal.
  * Child-run gates can pause parent execution and auto-resume once the child completes; abandoning now cascades to eligible descendants and reports cascade failures and blocked parents.
* **Bug Fixes**
  * Prevented incorrect approval/rejection actions on containers blocked by child sub-runs.
  * Fixed workflow activation exclusion logic to avoid self-blocking/incorrect “older-wins” ordering.
* **Documentation**
  * Updated workflow YAML schema, authoring guides, API/CLI behavior, and database/schema documentation for child-gate and abandon semantics.
* **Tests**
  * Added extensive unit and end-to-end coverage for child sub-run execution, cancel/abandon cascades, and ancestor/edge cases.


Wirasm:
## PR Review Summary

Multi-agent review (`/prp-review --agents`): code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer, code-simplifier, docs-impact. All three verdict-driving findings were independently re-verified by the aggregator in the worktree before inclusion.

### Critical Issues (2 found)

| Agent | Issue | Location |
|-------|-------|----------|
| code-reviewer | **Approving/rejecting a `child_workflow`-paused PARENT corrupts state.** `approveWorkflow`/`rejectWorkflow` special-case only `writeback` and `interactive_loop`; a `child_workflow` context falls through to the plain-approval branch (verified: zero `child_workflow` references exist in packages/core/server/cli/web). `archon workflow approve <parentId>` — and the parent IS the run surfaced as "paused" by `workflow status`/the console's generic pending-approval banner — writes a bogus `node_completed` for the `workflow:` node with `node_output: ''`; on resume the node is skipped, the real child's output is discarded (downstream `$node.output` = `''`), the parent "succeeds" with wrong data, and the still-paused child is orphaned forever. Reject is milder but orphans the child too. Fix: an explicit `child_workflow` branch that errors with "approve/reject the child run `<childRunId>` instead" (same shape as the `writeback` branch). No test covers this path. | `packages/core/src/operations/workflow-operations.ts:285-330` (approve), reject branch below; `packages/cli/src/commands/workflow.ts:2033`; console `RunsPage.tsx:314-319` |
| silent-failure-hunter | **A throw during parent auto-resume wedges the parent at `running` — non-terminal and non-resumable.** `maybeResumeParentRun` calls `hydrateResumableRun` (which CAS-flips the parent `paused→running` and per its own docstring "throws on database errors") and then `executeWorkflow`, with no try/catch — the only net is the bare `.catch` at the call site (`executor.ts:1338-1353`) that logs `workflow.parent_auto_resume_failed`. `executeWorkflow`'s own `failWorkflowRun` catch-all starts at ~:1144; unguarded calls before it (`loadConfig`, `getCodebaseEnvVars`, GitHub/provider env resolution) can throw after the CAS. The parent is then stuck `running`: `resumeWorkflow` rejects that status, so the only exit is destructive `abandon` (which cascade-cancels the subtree). The human who approved the child sees success; nobody is told the chain wedged. Fix: try/catch in `maybeResumeParentRun` that calls `failWorkflowRun(parentRunId, …)` on any throw so the row lands resumable. (Verified in worktree.) | `packages/workflows/src/executor.ts:626-703,1338-1353`; `hydrateResumableRun` at `:430-456` |

### Important Issues (5 found)

| Agent | Issue | Location |
|-------|-------|----------|
| silent-failure-hunter | `getRunAncestry(...).catch(() => [])` — twice, with **no log**. A transient DB error empties the path-lock exclusion set, so the child collides with its own parent's row, self-cancels, and the user sees "Sub-run 'X' was cancelled" — a complete misdiagnosis with zero diagnostic trail connecting it to the swallowed error. At minimum log the error; better, treat ancestry-lookup failure as fail-open for the path-lock (fail-closed here produces a false collision against the caller's own ancestor, not protection). (Verified in worktree.) | `packages/workflows/src/executor.ts:682,991` |
| type-design-analyzer | `ApprovalContext` stays a flat interface with all variant fields optional; the "childRunId present iff type === 'child_workflow'" invariant lives only at the single write site. `isApprovalContext` doesn't check it, and `maybeResumeParentRun` treats a childRunId-less context as "different child, skip" — a silent no-op wedge on malformed/stale metadata. The same file already has the precedent (`nodeOutputSchema` uses `z.discriminatedUnion`). | `packages/workflows/src/schemas/workflow-run.ts:143-260` |
| type-design-analyzer + comment-analyzer | Contract/enforcement gaps around the closure seam: `childOutcomeFromRun` blind-casts 6-value `WorkflowRunStatus` to the 4-value outcome status and `interpret()` has no exhaustive default — a non-terminal status would silently return `undefined` into `runLayers`. And both "Never throws" docstrings are false as written (unguarded `getWorkflowRun` at executor.ts:600; `hydrateResumableRun` throws by design) — saved today only by caller catches. Add a runtime guard/exhaustive default and fix the doc contracts. | `packages/workflows/src/dag-executor.ts:361-378,4654-4669`; `executor.ts:471,600,623` |
| code-reviewer | No upward propagation when a CHILD is abandoned/rejected directly: cascade goes down only, so the parent stays `paused` indefinitely referencing a terminal child. Recoverable via manual `resume` (hits the `cancelled` re-entry branch) but nothing surfaces the state — counter to the repo's "surface the ambiguous state + one-click action" principle. | `packages/core/src/operations/workflow-operations.ts:166` (abandon), `dag-executor.ts:4712-4716` |
| pr-test-analyzer | Test-gap cluster on the riskiest seams: (a) the double-fire re-entrancy guard (`parent?.status !== 'paused'`) executes in test 1 but is never asserted — a regression to recursive re-entrant parent execution would pass the suite (crit 8); (b) `findChildRuns`/`getRunAncestry`/`excludeRunIds` real SQL has zero tests, breaking `workflows.test.ts`'s own every-function pattern (crit 7); (c) the `resumeFailedChild` resume-through-parent branch — claimed verified in the PR body — has no test anywhere, nor does the `cancelled` re-entry branch (crit 7). | `executor.ts:637`; `packages/core/src/db/workflows.ts:437-524`; `dag-executor.ts:4707-4716` |

### Suggestions (8 found)

| silent-failure-hunter | A completed child with no non-blank terminal output silently threads `''` into `$node.output` (`outcome.output ?? ''`, no log). Add a warn when output is undefined. | `dag-executor.ts:4594-4595` |
| silent-failure-hunter | Cascade cancel is correctly best-effort internally, but partial failures are invisible to the caller — surface a `cascadeFailures` count in the abandon result rather than only server-side warn logs. | `workflow-operations.ts` cascade helper |
| code-simplifier | The `ancestorRunIds` option is dead weight: `executeWorkflow` unconditionally re-derives ancestry whenever `parent_run_id` is set, so the threaded hint is merged away. Drop the field and both call-site precomputations. (Also resolves the :682 silent catch as a bonus.) | `executor.ts:682,988-997` |
| code-simplifier | Two of the four re-entry branches in `executeWorkflowNode` duplicate `interpret()`'s completed/cancelled cases — route them through `interpret(childOutcomeFromRun(existing))`. | `dag-executor.ts:4670-4716` |
| code-simplifier + comment-analyzer | The child-blocked pause persists no `approval_requested` workflow_event (unlike the approval node it mirrors) — likely intentional (reason lives in run metadata) but undocumented; add the one-line why. Also fix the dispatch-site comment claiming `node_completed` is "handled by the shared completed-node path" — it contradicts the correct comment five lines up (it's written in `asCompleted`). | `dag-executor.ts:4585-4610,5327-5328` |
| code-reviewer | Two sibling `workflow:` nodes in one layer racing to pause contend on the parent row's single approval slot (second CAS loser hard-fails the node). Pre-existing limitation class, but the Non-goals list doesn't cover it — add a doc note or a load-time warning for multiple `workflow:` nodes in one layer. | `packages/core/src/db/workflows.ts:940` |
| pr-test-analyzer | Secondary test gaps: indirect A→B→A cycle (guard logic handles it, unproven), concurrent sibling children, cascade partial-failure resilience, depth-cap (5), include:+workflow: composition (explicitly supported via the ref-surface rewrite, zero composed tests), token roll-up with non-zero values. | various |
| type-design-analyzer | `api.generated.d.ts` sweeps in stale-generation drift from earlier PRs (`container`, `pi`, `signal_completes`, `include`) — one sentence in the PR body would stop reviewers attributing those to this PR. | `packages/web/src/lib/api.generated.d.ts` |

### Strengths

- **The double-resume race is genuinely handled**: the parent auto-resume reuses `resumeWorkflowRun`'s existing CAS, so automatic vs manual resume can't double-execute — the loser gets `WorkflowNotResumableError`.
- **Dual-schema parity done exactly right**: `parent_run_id` in both `sqlite.ts` and `000_combined.sql` with matching index, plus a real column+index parity assertion (not just table names). `generate:bundled`/`generate:bundled-schema` reproduce zero diff.
- The cycle guard catches indirect cycles (full ancestor `workflow_name` matching), fails loudly with actionable messages, and has a depth cap.
- The `workflow:` node schema is a disciplined clone of the vetted `include:` pattern — reserve-and-reject for slice-2 fields with clear per-field messages, `z.literal('inherit')` narrowing so `worktree` can't survive validation at the type level, z.infer-only throughout.
- The subrun e2e drives the *real* executor recursion (child gate → parent pause → approve child → auto-resume → output threading → path-lock exclusion proven load-bearing), and pause/no-`node_completed` semantics are properly asserted.
- The `runChildWorkflow` closure contract is fully typed (no `any`); the constitution doc is updated honestly in the same PR; cascade cancel is CAS-guarded, cycle-safe, bounded, and traverses through terminal children to reach grandchildren.

### Documentation Issues

- `CLAUDE.md:831` — the normative node-type enumeration doesn't list the new `workflow:` node.
- `CLAUDE.md:482` — the `workflow_runs` table entry doesn't mention `parent_run_id` (self-FK, `ON DELETE SET NULL`).
- `packages/docs-web/.../book/quick-reference.md` (~131) — the independently-maintained node table lists `include` but not `workflow`; add a row + a "Workflow-specific fields" subsection (`input`, `isolation: inherit` only, retry rejected, disallowed in loop_group).
- `packages/docs-web/.../reference/variables.md` — the substitution-scope node list should include `workflow:` (`input:` field; substituted like `prompt:`, not bash-escaped).
- Feature-completeness note: the CLI (`workflow get`/`runs`) doesn't surface parent/child relationships at all — only the console badge does. Worth a follow-up issue.

### Validation

| Check | Result |
|-------|--------|
| `bun run validate` (full suite, in worktree) | PASS — exit 0, all eight checks |
| CI (test ubuntu + windows, docker-build) | PASS |
| Targeted suites (subrun, loader, dag-executor, executor, sqlite parity, workflow-operations) | PASS individually |
| `generate:bundled` / `generate:bundled-schema` regen | Zero diff |

### Verdict

CRITICAL ISSUES

### Recommended Actions

1. Add the `child_workflow` branch to `approveWorkflow`/`rejectWorkflow` (error pointing at the child run id, mirroring the `writeback` branch) + a test for each — this closes the state-corruption path from the primary approval UX.
2. Wrap the resume pass in `maybeResumeParentRun` so any throw lands the parent in `failed` (resumable), never wedged at `running`.
3. Log (or fail-open) the two `getRunAncestry` catches.
4. Address the Important cluster: discriminate `ApprovalContext` (or at least validate childRunId presence), guard `childOutcomeFromRun`/`interpret` exhaustiveness, fix the two "Never throws" docstrings, and pin the double-fire guard + new SQL + `resumeFailedChild` with tests.
5. Then the docs/suggestion items — the four doc gaps are mechanical.

The core design is sound and slice-1 discipline is genuinely good; both criticals are narrow seams (operations-layer awareness of the new gate type, and one missing try/catch), not architectural problems.

Wirasm:
## Review findings addressed

All findings from the multi-agent review triaged per the rule: address everything unless clearly YAGNI, over-engineering, slice-2 scope (per `.claude/PRPs/plans/workflow-subruns.plan.md`), or off-direction. Seven commits, one per concern.

### Fixed

| Finding | Commit | Notes |
|---|---|---|
| **C1** approve/reject on a `child_workflow`-blocked parent corrupts state | `98c604cf` | Both ops now error with a redirect to the child run id (mirrors the `writeback` special-case shape) + tests asserting nothing is resolved/stamped. The web routes additionally pre-check and 400 with the same redirect (they delegate to the shared ops, but previously would have surfaced an opaque 500). |
| **C2** throw during auto-resume wedges parent at `running` | `620b32e4` | Two catch scopes: pre-CAS hydrate failure → parent stays `paused` (state intact; a lost resume CAS is logged as the benign race, not a generic error); post-CAS `executeWorkflow` throw → `failWorkflowRun` so the parent lands resumable. New e2e sabotages the resume pass and pins `failed`-not-`running`. |
| **I3** silent `getRunAncestry().catch(() => [])` ×2 | `432917ba` | The path-lock ancestry lookup now fails OPEN with a loud error log (fail-closed produced a false self-collision against the caller's own parent, misreported as "sub-run was cancelled"). The second catch site was deleted outright — see the simplification below. |
| **I5** false "Never throws" contracts + blind status cast + non-exhaustive `interpret` | `cc42db24` | Every await in `runChildWorkflow`/`maybeResumeParentRun` is now guarded so the documented contract is enforced where stated; `childOutcomeFromRun` throws on `running`/`pending` (cast removed — the guard narrows); `interpret` gains a `never`-typed default (compile-time exhaustive, runtime fail-loud). |
| **I4** (partial) `childRunId` invariant unenforced | `cc42db24` | A `child_workflow` gate missing `childRunId` is now a loud error log (invariant violation), not a silent skip that wedges the parent. Full discriminated-union refactor of `ApprovalContext` **deferred** — see below. |
| **I6** abandoning a child strands its blocked parent undetected | `c6fbead4` | `abandonWorkflow` returns `{ run, cascadeFailures, blockedParentRunId }`; CLI/chat/manage_run/web all print the blocked-parent pointer (resume fails the node cleanly, or abandon the tree) and the cascade-failure warning. Surfacing only — no autonomous parent mutation, per the lifecycle principle. |
| **I7** test-gap cluster | `08e58006` | Double-fire guard now asserted (no mid-flight resume, no node completes twice); `resumeFailedChild` re-drive + cancelled-child re-entry covered e2e; indirect A→B→A cycle; depth cap; include:+workflow: composition; SQL-shape tests for `findChildRuns`/`getRunAncestry`/`excludeRunIds`. |
| S1 completed child with empty output silently threads `''` | `2e232b08` | Warn log (`workflow.subrun_completed_without_output`). |
| S2 cascade partial failure invisible to caller | `c6fbead4` | `cascadeFailures` count surfaced on every abandon surface + test proving the walk continues past a failing descendant. |
| S3 dead `ancestorRunIds` threading | `432917ba` | Field + both call-site precomputations removed; `executeWorkflow` always derives from the run's own `parent_run_id` (which the review verified is the correct source on every entry path). |
| S4 re-entry branches duplicate `interpret` cases | `2e232b08` | `completed`/`cancelled` re-entry now routes through `interpret(childOutcomeFromRun(existing))` — three-way state table. |
| S5 contradictory/undocumented comments | `2e232b08` | Dispatch-site `node_completed` comment corrected; pause-mirror comment scoped to the PAUSE primitives + documents why no `approval_requested` event row is persisted; `getActiveWorkflowRunByPath` docblock documents `excludeRunIds`. |
| Docs (CLAUDE.md node list + `workflow_runs` entry, quick-reference table + sub-run fields, variables.md, sibling-gate non-goal) | `9ff6ef55` | |

### Found while fixing (not in the review)

- **Older-wins tiebreaker placeholder bug** (`08e58006`): the new SQL-shape test for `excludeRunIds` immediately caught that the tiebreaker's self-id placeholder was derived as `params.length - 1`, which points at the **last ancestor id** whenever exclude params sit between the self id and `startedAt` — so a sub-run's lock query tiebroke against an ancestor id instead of self. Exactly the bug class the review predicted for the untested dynamic placeholder-building. Fixed by capturing the placeholder at push time.
- **Web abandon route drift** (`c6fbead4`): `POST /api/workflows/runs/:id/abandon` called `cancelWorkflowRun` directly instead of the shared op — so a console abandon skipped **both** the sub-run cascade cancel and the M2 container reclaim that CLI/chat abandons perform. Now delegates to `abandonWorkflow`.

### Deferred, with rationale

- **`ApprovalContext` as a discriminated union** (type-design-analyzer): the right end-state, but it reshapes the persisted-metadata parsing surface for *every* gate type (`approval`/`interactive_loop`/`writeback`/`child_workflow`) across two packages — scope creep inside a feature PR. The dangerous consequence (silent no-op on a malformed `child_workflow` context) is closed by the loud invariant log. Follow-up candidate.
- **Load-time validator warning for sibling `workflow:` nodes in one layer** (code-reviewer, low): needs layer computation at validate time for a pre-existing limitation class shared with concurrent `approval:` nodes — documented as an explicit non-goal with the `depends_on` workaround instead.
- **CLI surfacing of parent/child run links** (docs-impact): new feature surface; the slice-1 plan's chosen affordance is the console badge ("the field + badge is enough for slice 1"). Follow-up candidate.

`bun run validate` passes end-to-end (all eight checks).

Wirasm:
## PR Review Summary

Multi-agent review (`/prp-review --agents`, round 2): code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer, code-simplifier, docs-impact — against `feat/workflow-subruns` at `41915086` (rebased onto current `dev` during this review; the two docs conflicts were resolved as dev's corrected capability claims + this PR's `workflow:` additions). `bun run validate` passed in full both before and after the rebase. The two verdict-driving findings (I1, I2) were independently re-verified by the aggregator in the worktree.

**Both criticals from round 1 are confirmed fixed, with regression tests that fail without the fix**: approve/reject on a child-blocked parent now refuses-and-redirects before any gate mutation (asserted at `workflow-operations.test.ts:398,635`), and a throw during parent auto-resume lands the parent in `failed` (resumable), never wedged at `running` (`subrun.test.ts:514`).

### Critical Issues (0 found)

None. Both previously-flagged criticals are fixed and regression-tested.

### Important Issues (6 found)

| I1 | silent-failure-hunter | Child-spawn path lacks the wedge guard the parent path got in round 1: `runChildWorkflow`'s catch returns a failed node outcome but never calls `failWorkflowRun(childRunId)`. A throw in `executeWorkflow`'s early setup (config load, provider check, token resolution — before its own catch-all; the catch's own comment acknowledges this window) leaves the pre-created child row at `pending`/`running` forever: a `running` zombie has no staleness expiry, can hold the working-path lock against unrelated runs, and parent re-entry re-pauses on the dead child with no signal. Apply the symmetric guard from `maybeResumeParentRun` and add the mirror sabotage test. (Verified in worktree.) | `packages/workflows/src/executor.ts:602` |
| I2 | silent-failure-hunter | Auto-resume reconciliation silently no-ops: every early-return in `maybeResumeParentRun` before the `executeWorkflow` call (lookup throw, malformed gate, missing working_path, discovery throw, workflow not found, hydrate throw) only writes a server log and returns. The child terminates correctly but the parent stays `paused` on a stale "blocked on sub-run" gate with zero user-facing signal that the hand-off failed. Manual `/workflow resume` recovers, but per the repo's own surface-ambiguous-state principle these paths should at minimum message the user with a `/workflow resume <parentId>` pointer. (Verified in worktree.) | `packages/workflows/src/executor.ts:644-716` |
| I3 | code-reviewer | Cycle guard compares the raw YAML `node.workflow` string against resolved ancestor names *before* `resolveWorkflowName` runs — a case-variant/suffix/substring reference to an ancestor slips past and is only caught by the depth-cap of 5 with a less-informative error. One-line fix: resolve first, then compare against `childWorkflow.name`. | `packages/workflows/src/executor.ts:498` |
| I4 | code-reviewer | Two `workflow:` nodes in the same DAG layer race the single-slot pause CAS: the loser's `pauseWorkflowRun` throw is absorbed into a node failure with no `node_failed` event written, and the follow-up branches are skipped because the run is already `paused` — the loser's child run is real and consuming cost but unmentioned until the next resume self-heals it. Pre-existing gap for concurrent `approval:` nodes, but this feature invites fan-out layouts that will actually hit it. Fix (retry-pause-once, or a gate queue) or file a follow-up issue covering both node types. | `packages/workflows/src/dag-executor.ts:4657` |
| I5 | pr-test-analyzer | The "parent resumed while child still mid-flight" re-entry branch (in-progress child → re-pause, "NEVER resume a paused child") has no test driving `executeWorkflow` back into the parent while the child is non-terminal — every existing test completes the child first. Easy to silently regress into a duplicate-drive-of-the-child bug. Criticality 7/10. | `packages/workflows/src/dag-executor.ts:4745` |
| I6 | type-design-analyzer | `findChildRuns`/`getRunAncestry` fatten `IWorkflowStore` directly, against the project's own ISP rule (new concern → new narrow interface). Run-tree navigation is a distinct concern; a small run-tree interface is a cheap fix. | `packages/workflows/src/store.ts` |

### Suggestions (9 found)

| S1 | silent-failure-hunter | `cascadeCancelChildren`'s `MAX_CASCADE_RUNS = 500` cap exits silently on truncation — caller sees "abandoned, 0 cascade failures" while part of the tree is still alive. Log + report the truncation. | `packages/core/src/operations/workflow-operations.ts:83-113` |
| S2 | silent-failure-hunter | `childRunId` missing from `pauseWorkflowRun`'s explicit `?? null` reset list — the list exists precisely to stop SQLite `json_patch` field leaks between gates (per the function's own comment). Currently inert (all consumers check `type` first) but a latent trap. | `packages/core/src/db/workflows.ts:966-973` |
| S3 | silent-failure-hunter | `asCompleted` awaits `createWorkflowEvent` unguarded, inverting the file's fire-and-forget `.catch(logEventStoreError)` pattern — a transient event-store failure reports a successfully-completed child as a failed parent node (self-heals on resume, but inconsistent). | `packages/workflows/src/dag-executor.ts:4614` |
| S4 | pr-test-analyzer | The `cascadeFailures`/`blockedParentRunId` warning formatting is threaded through 4 call sites (API, CLI, manage-run tool, command handler) with zero tests on the non-default branch; the API route also duplicates the child-gate 400 guard untested. Criticality 6/10. | `packages/server/src/routes/api.ts:3260` + peers |
| S5 | pr-test-analyzer | No test that a typo'd `workflow:` target fails cleanly with "Unknown sub-run workflow" — the most likely real-world authoring mistake. 4/10. | `packages/workflows/src/executor.ts` (`runChildWorkflow`) |
| S6 | code-simplifier | Extract a `failOutcome(error, childRunId = '')` helper — the failed-outcome literal is constructed 7× in `runChildWorkflow`. Pure dedup, ~20 lines. | `packages/workflows/src/executor.ts:468-613` |
| S7 | code-simplifier | Shared `isBlockedOnChildRun(parent, childRunId)` predicate: the parent-blocked business rule is independently implemented in `findParentBlockedOn` and `maybeResumeParentRun`'s guards — a duplicated invariant likely to drift if the gate schema changes. | `packages/core/src/operations/workflow-operations.ts:126` + `packages/workflows/src/executor.ts:652` |
| S8 | type-design-analyzer | Backfill `workflow?: never` on the eight existing node types (the new `WorkflowNode` excludes its siblings; the reverse wasn't done), and model `ChildWorkflowOutcome.childRunId` absence as optional instead of the `''` sentinel. | `packages/workflows/src/schemas/dag-node.ts:253+`, `dag-executor.ts:316` |
| S9 | code-reviewer | `api.generated.d.ts` picked up unrelated already-merged fields on regen (stale before this branch) — diff noise; one fresh `generate:types` sanity check against dev, not blocking. | `packages/web/src/lib/api.generated.d.ts` |

### Strengths

- The two round-1 criticals got the most deliberate treatment in the PR: structural fixes (refuse-and-redirect; fail-parent-on-resume-throw) plus regression tests that fail without the fix — the right prioritization.
- Never-throws contracts on `runChildWorkflow`/`maybeResumeParentRun` are honored end-to-end; cycle/depth/mode-exclusivity guards target the real failure modes of cross-run composition.
- Dual SQLite/Postgres schema parity done correctly (self-referential FK, `ON DELETE SET NULL`, matching index, column-level parity assertion) — and the round-1 SQL placeholder bug was pinned with a dedicated SQL-shape test.
- Comment quality is unusually high: every non-trivial claim was verified against the implementation; zero stale or incorrect comments found.
- Touched docs are accurate — depth-cap-of-5, one-blocking-gate CAS, and cooperative-cancel claims all traced to code and check out.

### Documentation Issues

- `packages/docs-web/src/content/docs/reference/cli.md` — `workflow abandon` doesn't mention cascade-cancel of descendants or the blocked-parent warning; `approve`/`reject` don't mention the child-gate redirect error.
- `packages/docs-web/src/content/docs/reference/api.md` — same gaps in the route table (cascade behavior; child-gate 400).
- `packages/docs-web/src/content/docs/reference/database.md` + `reference/architecture.md` — `remote_agent_workflow_runs` listings missing the new `parent_run_id` column (CLAUDE.md's parallel entry *was* updated).
- `CLAUDE.md` (~line 913) — the abandon endpoint line still says "marks as cancelled" with no cascade mention.
- Nice-to-have: point readers at the runnable bundled examples (`archon-compose-child-qa`, `archon-compose-plan-implement-qa`) from the new authoring-workflows section, matching the `include:` section's precedent; the "12 bundled" count on that page was already stale on dev.

### Verdict

NEEDS FIXES

### Recommended Actions

1. I1 — add best-effort `failWorkflowRun(childRunId, …)` to `runChildWorkflow`'s catch, mirroring the parent-side guard, plus the child-path sabotage regression test.
2. I2 — surface auto-resume reconciliation failures to the user (message + `/workflow resume <id>` pointer) instead of log-only returns.
3. I3 — resolve the workflow name before the ancestry cycle check (one-line reorder).
4. I4 — retry-pause-once on CAS miss now, or file a follow-up issue covering both `workflow:` and the pre-existing `approval:` case.
5. I5 — add the mid-flight parent-resume test.
6. Docs parity pass (cli.md, api.md, database.md, architecture.md, CLAUDE.md abandon line).
7. Optional here or as follow-ups: I6 (ISP split), S1–S9.

Wirasm:
## Round-2 review addressed — pushed `ba4cb7d9..5fbaa555`

`bun run validate` green (exit 0) on the final committed state. Per-ID disposition below.

### Important

|----|-------------|--------------|
| I1 | **Fixed** | `runChildWorkflow`'s catch now best-effort-cleans the child so an early-setup throw can't strand a zombie holding the path lock. **Deviation from the literal suggestion (documented):** used `cancelWorkflowRun`, not `failWorkflowRun` — the reachable state is **`pending`** (early setup runs before the status→running flip and before executeWorkflow's own catch-all), and `failWorkflowRun`'s `WHERE status='running'` would miss it; `cancelWorkflowRun` covers pending+running and no-ops on terminal. Mirror sabotage test added (child spawn's `getCodebaseEnvVars` throws → child terminal, not zombie). |
| I2 | **Fixed** | Every reconciliation-failure return in `maybeResumeParentRun` (lookup throw, no working_path, discovery throw, workflow-not-found, hydrate throw) now sends a guarded user-facing message with a `/workflow resume <parentId>` pointer. The benign lost-CAS race and the synchronous/different-child no-ops stay silent. |
| I3 | **Fixed** | `resolveWorkflowName` now runs **before** the cycle check, which compares the **resolved canonical name** against ancestor names — a case-variant/suffix/substring self-reference is caught as a cycle, not the depth cap. Case-variant cycle test added. |
| I4 | **Follow-up filed → #2180** (+ code comment at `pauseParentOnChild`). **Verified, then chose the follow-up path:** `pauseWorkflowRun` throws on the 0-row CAS miss, so a same-layer sibling's pause is swallowed — but retry-pause-once **cannot** fix it: the run has a **single** approval-gate slot, so there is nowhere to record a second simultaneous block. The real fix is a gate queue or a load-time reject of multiple gate-pausing nodes per layer; #2180 covers both `workflow:` and the pre-existing `approval:` case. |
| I5 | **Fixed** (test + a real bug it surfaced) | Added the mid-flight re-entry test (parent resumed while child still paused → re-pause, child not double-driven). It exposed that the path-lock only excluded **ancestors**, so a parent resumed while blocked on a paused child self-cancelled against its own child. Fixed by also excluding the run's **descendants** (bounded downward walk); siblings intentionally not excluded (that's #2180). |
| I6 | **Fixed** | Split `findChildRuns`/`getRunAncestry` into a narrow `IRunTreeStore`; `IWorkflowStore extends IRunTreeStore` so consumers/adapters don't churn. |

### Suggestions

| ID | Disposition | Note |
|----|-------------|------|
| S1 | **Fixed** | `cascadeCancelChildren` reports truncation (non-zero `cascadeFailures` + a distinct `..._cascade_truncated` log) when it hits the cap with the queue non-empty. Test added. |
| S2 | **Fixed** | `childRunId` added to `pauseWorkflowRun`'s explicit `?? null` reset list. |
| S3 | **Fixed** | `asCompleted`'s `node_completed` write is now fire-and-forget with `.catch` — a transient event-store failure no longer reports a completed child as a failed parent node. |
| S4 | **Fixed** | Added approve + reject **child-gate 400** route tests (`api.workflow-runs.test.ts`) — a child-blocked parent 400s with the child id and mutates no gate. Core cascade/blocked-parent detection was already covered by round-1 tests. The four display surfaces render from that same tested `AbandonWorkflowResult`; I did not add four separate string-format tests (display-only). |
| S5 | **Fixed** | Typo'd target → `Unknown sub-run workflow` test added. |
| S6 | **Fixed** | `failOutcome` helper dedups the failed-outcome literals in `runChildWorkflow`. |
| S7 | **Fixed** | Shared `isRunBlockedOnChild` predicate (`schemas/workflow-run`) used by both `findParentBlockedOn` and `maybeResumeParentRun`. |
| S8a | **Skipped (per maintainer)** | `workflow?: never` backfill on the eight existing node types — contradicts the round-1 amendment (the union discriminates on key presence; `include`/`script` set the precedent). Not this PR's call. |
| S8b | **Skipped — verified it ripples** | Making `ChildWorkflowOutcome.childRunId` optional breaks the `interpret`→`pauseParentOnChild(childRunId: string)` contract (the `paused` outcome always has a real id, but the type would no longer guarantee it), forcing guards/assertions at the consumers. Non-trivial; kept the `''` sentinel. |
| S9 | **No-op confirmed** | Regenerated `api.generated.d.ts` against the rebased branch and ran prettier — **byte-identical** to the committed file. The round-1 regen against rebased `dev` already produced the clean version; no noise remains. |

### Beyond the review (found while fixing)

- **`node_failed` event for the `workflow:` node** — its failure reason (cycle, unknown target, cancelled child) was previously swallowed into the run-level DAG summary and not auditable per-node (`runLayers` writes `node_failed` for no node type; each executor writes its own). `executeWorkflowNode` now writes one, matching every other node type. This is also what the I3/S5 tests assert on.

### Maintainer decision (folded in)

- **Dropped the two bundled demo YAMLs** (`archon-compose-plan-implement-qa`, `archon-compose-child-qa`) — nothing referenced them, `subrun.test.ts` is the real coverage. Refreshed the embed (21 workflows, `check:bundled` green). `authoring-workflows.md` now uses a generic `qa-block` placeholder, and the stale **"12 bundled" → 21** count is fixed.

### Docs parity

`cli.md` (abandon cascade + blocked-parent + approve/reject child-gate redirect), `api.md` (route table + child-gate 400 note), `database.md` + `architecture.md` (`parent_run_id` column), `CLAUDE.md` (abandon line).

PR #2092, opened 2026-07-14
feat(isolation/executor/cli): base-branch fallback from the codebase default branch (#2086)

Wirasm:
## Summary

Describe this PR in 2-5 bullets:

- Problem: `default_branch` is stored correctly at registration but never consulted for base-branch resolution. A locally-registered repo whose default branch is not `main` (e.g. `develop`), with `origin/HEAD` unset and no `origin/main`, throws at worktree creation and resolves `$BASE_BRANCH` to `''` — even though the DB knows the answer.
- Why it matters: any non-main-default local repo (common `develop`/`dev` trunks) can't run isolated workflows without manually setting `worktree.baseBranch` in every repo config.
- What changed: base-branch resolution is now a three-level fallback at every seam — repo config `worktree.baseBranch` → codebase `default_branch` → git auto-detect. Threaded through isolation (`IsolationRequestBase.baseBranch` / `ResolveRequest.codebase.defaultBranch`), the workflow executor (`ExecuteWorkflowOptions.baseBranch`), the CLI (all three sites), and the chat/web orchestrator dispatch paths. Plus a classified "cannot detect default branch" error message and a docs update.
- What did **not** change (scope boundary): a genuinely remote-less local repo still cannot fork a worktree from `origin/<branch>` — documented limitation, same as the original chunk. Folder projects are untouched (they short-circuit before worktree isolation; non-regression test added). No schema or registration changes — `default_branch` was already captured.

Based on groundwork by @anhle128 in #1994's original branch (preserved at 1caf2e08). Reimplemented fresh against current `dev` (folder projects and recent executor/CLI changes had moved the anchors).

### Before

```
  User (repo with default branch 'develop', no origin/main)
  ────
  archon workflow run implement "Add auth"
        │
        ▼
  Archon: worktree creation → getDefaultBranch()
          origin/HEAD unset, origin/main missing
        ▼
  ❌ "Cannot detect default branch for /repo: neither
     origin/HEAD nor origin/main exist"
  (workaround: hand-edit worktree.baseBranch in every repo)

  $BASE_BRANCH in prompts → '' → node fails if referenced
```

### After

```
  User (repo with default branch 'develop', no origin/main)
  ────
  archon workflow run implement "Add auth"
        │
        ▼
  Archon: worktree creation
          [config worktree.baseBranch? no]
          [codebase default_branch? → 'develop']  *new*
        ▼
  ✅ worktree forked from origin/develop
  $BASE_BRANCH → 'develop' in prompts             *new*

[...]

Issue #2086, opened 2026-07-14
feat(executor/isolation/cli): base-branch fallback threading for locally-registered repos with non-main default branches


Wirasm:
## What

PR #1994's original branch contained a substantial, self-contained chunk beyond its titular fix: **base-branch fallback threading through the executor, isolation, and CLI (with tests)** so locally-registered repos whose default branch is not `main` (e.g. `develop`) resolve `$BASE_BRANCH` / worktree base correctly.

It was deliberately **dropped** during the maintainer take-over of #1994 (the two "update" commits mixed ~2,100 lines of unrelated work, including an actively-wrong `.archon/config.yaml` change) — the old head is preserved at `1caf2e08` on the PR for reference.

## Why an issue

The capability is real and wanted; it just needs its own PR and review cycle instead of riding along unreviewed. @anhle128 — this was your work: you're warmly invited to re-open it as a focused PR (attribution preserved via cherry-pick from `1caf2e08`). If not, a maintainer can extract it from the preserved head.

## Scope sketch (from the dropped commits)

- Executor: `$BASE_BRANCH` fallback when `worktree.baseBranch` unset and default branch ≠ main
- Isolation: worktree creation base-branch resolution for `_local` repos
- CLI: threading + tests

Refs: #1994 (takeover comment maps what was kept vs dropped).

Wirasm:
**Verified against current `dev` (`0e70c933`, 2026-07-14): the gap is STILL REAL.** None of the recent merges (#2055/#2032/#2081/#2063/#1994) address it, and `git grep` confirms no alternate fallback channel exists.

Key finding: `default_branch` is already **stored correctly** at registration (`clone.ts:138-148` — `git rev-parse --abbrev-ref HEAD`); it's simply never consulted for base resolution. Per-sub-behavior on current dev:

- **Worktree creation base**: `providers/worktree.ts:709` uses repo-config `worktree.baseBranch` → `getDefaultBranch` only; a `develop`-default repo with `origin/HEAD` unset and no `origin/main` **throws** instead of forking from `develop`. `IsolationRequestBase`/`ResolveRequest.codebase` carry no branch field.
- **`$BASE_BRANCH`**: `workflows/executor.ts:450-465` is a two-level chain (config → git auto-detect → `''`) with no codebase-default layer — prompts referencing `$BASE_BRANCH` fail where the DB knows the answer.
[...]

PR #2203, opened 2026-07-21
feat(cli): per-dispatch --base override for worktree cut-from + PR target

### LINKED ISSUE #2355: Feature Request: per-dispatch base-branch override for `archon workflow run`
# Feature Request: per-dispatch base-branch override for `archon workflow run`

## Problem

Base-branch resolution today (#2086 / #2092) is **static per codebase**:

1. `worktree.baseBranch` in `.archon/config.yaml`
2. the registered codebase's stored `default_branch`
3. git auto-detection

All three describe the repo, not the run. There is no way to say "for this one
dispatch, use base X."

That gap bites whenever several runs fan out against the **same** repo needing
**different** bases:

- an epic whose slices each cut from their own integration branch
- an A/B where each design variant targets a different base
- a backport that must branch from a release line

The only lever is editing repo config, which is:

- **global** — it changes the base for every run in the repo, not just yours
- **racy under concurrency** — two dispatches editing config to different values
  interleave, and whichever writes last silently decides the base for both
- **unable to express the intent at all** — "this run, this base" has no encoding

## Why the existing flags don't cover it

`--from/--from-branch` exists but is deliberately narrower: it sets only the
worktree **start-point**, not the PR target. A run using `--from` still reports
the config-resolved branch as `$BASE_BRANCH`, so `gh pr create --base` opens the
PR against the repo default. That is correct for what `--from` means, and it
leaves the "both halves" case unserved.

## Proposal

A `--base <branch>` flag on `archon workflow run` that overrides, for that single
dispatch, **both** the worktree cut-from **and** the PR target (`$BASE_BRANCH`),
slotting in as a new top precedence level above repo config:

| Level | Source | Scope |
|---|---|---|
| 1 | `--base <branch>` | one dispatch |
| 2 | `worktree.baseBranch` (repo config) | the repo |
| 3 | codebase `default_branch` | the repo |
| 4 | git auto-detection | the repo |

Overriding both halves from one flag is the default because cutting from branch
A while opening a PR against branch B is almost never what someone means. The
case where it *is* meant stays expressible by combining `--from` (cut-from only)
with `--base` (PR target), so no second flag is needed:

```bash
# branch off release/2.0, open the PR against dev
archon workflow run implement --from origin/release/2.0 --base dev "Backport the fix"
```

Absent the flag, resolution must stay byte-identical to today — this is purely
additive.

## Notes

This adds a **fourth** way to determine a base branch, so the precedence chain
needs to be documented where a user will find it: the `workflow run` flag table,
the `$BASE_BRANCH` reference, and the quick reference.

`--base` should be rejected — not silently half-applied — on any path that
creates no worktree (`--no-worktree`, folder projects, workflows pinning
`worktree.enabled: false`), since there the flag could only move the PR target
and would report a base no worktree was ever cut from.

buun-dev:
## Summary

- **Problem:** Base-branch resolution (#2086/#2092) is *static per codebase* — `worktree.baseBranch` → codebase `default_branch` → git auto-detect. All three describe the repo, not the run, so fanning out parallel dispatches against the **same** repo with **different** bases (epic slices off their own integration branches, A/B variants targeting different bases) has no encoding. The only lever is editing repo config: global, racy under concurrency, and unable to say "this one run, this base."
- **Why it matters:** It makes parallel, multi-base dispatch a first-class, config-free operation — the enabling primitive for parallel epic/integration and design-variant workflows.
- **What changed:** A `--base <branch>` flag on `archon workflow run` that overrides, for that single dispatch, **both** the worktree cut-from **and** the PR target (`$BASE_BRANCH`). It becomes the new top precedence level, layered on the existing #2086 fallback. Each half gets its own dedicated channel (`IsolationRequest.baseOverride`, `ExecuteWorkflowOptions.baseOverride`) so the override outranks repo config on both sides.
- **What did *not* change (scope boundary):** With `--base` absent, resolution is byte-identical to today on both halves. No second flag for decoupling cut-from from PR target — `--from` already does that (see UX Journey). No config-file surface, no new node fields, no changes to `--from`'s own semantics.

### Before

```
User                                Archon                          Worktree / PR
────                                ──────                          ─────────────
wants slice A off epic/foo
  edits .archon/config.yaml ──────▶ worktree.baseBranch: epic/foo
  archon workflow run ────────────▶ resolve base
                                      config? YES -> epic/foo ─────▶ cut from epic/foo
                                                                     PR --base epic/foo
wants slice B off main
  edits config AGAIN ─────────────▶ worktree.baseBranch: main
                                    (slice A still running; the
                                     edit is global and racy)
  archon workflow run ────────────▶ cut from main

  ✗ config is repo-global, so concurrent dispatches fight over it
  ✗ no way to express "this run, this base"
```

### After

```
User                                Archon                          Worktree / PR
────                                ──────                          ─────────────
archon workflow run --base epic/foo
                    ────────────────▶ flagBase = epic/foo
                                      *baseOverride* -> provider ──▶ cut from origin/epic/foo
                                      *baseOverride* -> executor ──▶ $BASE_BRANCH = epic/foo
                                                                     PR --base epic/foo
archon workflow run --base main
                    ────────────────▶ (concurrently, no config touched)
                                                                  ──▶ cut from origin/main
                                                                     PR --base main

  [+] per-dispatch, config-free, concurrency-safe
  [+] both halves move together -- no cut-from/PR-target skew

Decoupled case (--from overrides only the cut-from):
archon workflow run --from origin/release/2.0 --base dev
                                                                  ──▶ cut from release/2.0
                                                                     PR --base dev
```

### Before

```
                cli.ts (flag parse)
                      │
                      ▼
        workflow.ts (workflowRunCommand)
                      │
        ┌─────────────┴─────────────┐
        ▼                           ▼
IsolationRequest              ExecuteWorkflowOptions
  .baseBranch                   .baseBranch
        │                           │
        ▼                           ▼
worktree.ts                   executor.ts
 config ?? baseBranch          config || baseBranch
        │                           │
        ▼                           ▼
  git worktree add            $BASE_BRANCH -> gh pr create --base
```

### After

```
                cli.ts [~] (parse --base, reject with --no-worktree)
                      │
                      ▼
        workflow.ts [~] (flagBase, normalized once)
                      │
        ┌─────────────┴─────────────┐
        ▼                           ▼
IsolationRequest [~]          ExecuteWorkflowOptions [~]
  .baseOverride  [+] ===        .baseOverride  [+] ===
  .baseBranch    (fallback)     .baseBranch    (fallback)
        │                           │
        ▼                           ▼
worktree.ts [~]               executor.ts [~]
 baseOverride ?? config         baseOverride || config
   ?? baseBranch                  || baseBranch
        │                           │
        ▼                           ▼
  git worktree add            $BASE_BRANCH -> gh pr create --base

Guards [+] in workflow.ts: --no-worktree | worktree.enabled:false |
                           folder project  -> reject
                           worktree reuse  -> warn + validate against override
```

**Connection inventory:**

| `cli.ts` | `WorkflowRunOptions.baseBranch` | **new** | `--base` parsed and threaded |
| `workflow.ts` | `IsolationRequest.baseOverride` | **new** | worktree cut-from half |
| `workflow.ts` | `ExecuteWorkflowOptions.baseOverride` | **new** | PR-target half |
| `workflow.ts` | `ExecuteWorkflowOptions.baseBranch` | **modified** | reverted to codebase default (fallback only) |
| `workflow.ts` | `git.isAncestorOf` (reuse check) | **modified** | validates against the override, not repo config |
| `worktree.ts` | `syncWorkspaceBeforeCreate` | unchanged | still 3-arg with `remote` (#2234) |
| `executor.ts` | `substituteWorkflowVariables` | unchanged | `$BASE_BRANCH` consumer untouched |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: M`
- Scope: `cli`
- Module: `cli:workflow-run`, `isolation:worktree-provider`, `workflows:executor`

## Change Metadata

- Change type: `feature`
- Primary scope: `multi` (cli + isolation + workflows)

## Linked Issue

- Closes #2355
- Related #2086, #2092 (the static base-resolution chain this layers on), #2234 (remote resolution preserved through the rebase)

```bash
bun run check:bundled          # up to date (36 commands, 21 workflows)
bun run type-check             # 10/10 packages, exit 0
bun run lint --max-warnings 0  # clean
bun run format:check           # clean
bun --filter '@archon/cli' test        # 246 pass / 0 fail (workflow.test.ts + cli.test.ts)
bun --filter '@archon/workflows' test  # executor.test.ts 70 pass / 0 fail
bun --filter '@archon/isolation' test  # 56 pass / 0 fail (worktree.test.ts 154 pass)
```

- Evidence provided: counts above; 13 tests added across the three packages.
- Intentionally skipped: `bun run test:install` — the installer script hard-refuses on
  Windows (`"Windows is not supported. Please use WSL2"`), which aborts `bun run validate`
  before the test stage. Every other `validate` stage was run directly and is listed above.
  Three suite failures exist elsewhere (`@archon/providers` broken-symlink,
  `@archon/workflows` symlinked-home-command, `@archon/cli` 2x `downloadWebDist`); each was
  reproduced on an untouched checkout of the same commit base and is pre-existing and
  environmental, not caused by this change.

## Security Impact (required)

- New external network calls? **No** — `--base` selects which branch an already-performed
  fetch syncs; it adds no new endpoint.
- File system access scope changed? **No** — worktree paths are unchanged; only the
  start-point ref differs.
- The flag names a git ref that is passed to `git worktree add` and `gh pr create`. It goes
  through the same `git.toBranchName` normalization as `--from`, and is executed via
  `execFileAsync` (argv, never a shell string), so a crafted branch name cannot inject.

## Compatibility / Migration

- Backward compatible? **Yes** — with `--base` absent, `baseOverride` is `undefined` and
  both resolution chains are byte-identical to today.
- Config/env changes? **No**

- **Verified scenarios:**
  - The reported defect, reproduced as a failing test before the fix: with
    `config.baseBranch = 'main'` and the override set, `$BASE_BRANCH` resolved to `'main'`
    (`Expected "epic/foo", Received "main"`), then to `'epic/foo'` after.
  - All four rejection/warning paths driven red-then-green.
  - The `--from` + `--base` decoupling asserted end to end (cut-from `origin/release/2.0`,
    PR target `dev`).
  - Rebase conflict in `worktree.ts` resolved as a union — verified by the PR's own
    isolation test failing until it asserted `remote: 'origin'`, proving #2234's third
    argument survived.
- **Edge cases checked:** whitespace-only `--base` normalizes to `undefined` for behavior
  paths while still tripping the rejection guards (intent vs usable value); a `--base`
  branch missing on the remote is already a hard error from `syncWorkspace`, not a silent
  fallback.
- **What was not verified:** no live end-to-end dispatch against a real remote in this pass
  (the flag has run in production on a downstream fork, but not re-exercised live on this
  rebased branch); container-provider runs, which do not use `WorktreeProvider`.

- **Affected subsystems:** `cli` flag parsing and `workflowRunCommand`; `isolation`
  worktree creation and pre-create sync target; `workflows` executor `$BASE_BRANCH`
  resolution.
- **Potential unintended effects:** the executor precedence change touches the
  `$BASE_BRANCH` path for *every* run, not only `--base` runs. It is guarded by ordering —
  `baseOverride` is checked first and is `undefined` for every existing dispatch — so the
  remaining chain is untouched. The `base branch resolution` suite in `executor.test.ts`
  covers the unchanged levels and stays green.
- **Guardrails / early detection:** a mismatch surfaces as a PR opened against the wrong
  branch, which is visible immediately on the PR page. `worktree.reuse_base_override_partial`
  and `worktree.reuse_base_branch_mismatch` are logged as structured warn events.

## Rollback Plan (required)

- **Fast rollback:** `git revert` the three commits, or revert only
  `fix(cli): rank --base above repo config…` to keep the flag with its pre-fix precedence.
  No state to unwind — nothing is persisted.
- **Feature flags / toggles:** none needed; omitting `--base` is the off switch.
- **Observable failure symptoms:** a PR opened against a branch other than the one passed
  to `--base`; a worktree cut from a different branch than `$BASE_BRANCH` reports.

## Risks and Mitigations

- **Risk:** `--base` names a branch absent from the remote.
  - **Mitigation:** already fatal — `syncWorkspace` throws
    `Configured base branch '<x>' not found on remote '<remote>'` and `worktree.ts`
    classifies it as non-recoverable. Documented in the flag row.
- **Risk:** a future rebase re-resolves the `worktree.ts` hunk toward one side and silently
  drops either `remote` (#2234) or the override.
  - **Mitigation:** both are asserted — `uses request baseOverride over configured
    baseBranch and request baseBranch` checks the override *and* `remote: 'origin'` in the
    same call assertion, so dropping either fails CI.
- **Risk:** someone reads `--base` + `--from` as a bug and "fixes" it by making one win.
  - **Mitigation:** the pairing is covered by a test whose name and comment state the
    intent (`lets --from win the cut-from while --base still drives the PR target`), and
    the docs give the recipe.

## Summary by CodeRabbit

* **New Features**
  * Added a `--base <branch>` option to `workflow run` for selecting a custom base branch.
  * Custom base branches override configured defaults for worktree setup and workflow execution.
  * Added support for using `--from` separately to choose the worktree starting point.

* **Bug Fixes**
  * Prevented incompatible use of `--base` with `--no-worktree` or unsupported project configurations.
  * Improved validation and warnings when reusing existing worktrees.

* **Documentation**
  * Documented base-branch precedence, validation rules, and `--from`/`--base` behavior.


Wirasm:
Useful feature and the worktree half is wired correctly — I traced it end to end and `--base epic/foo` genuinely drives the sync target and the `git worktree add -b` start-point. But there's a correctness bug in the other half, and one design question I'd like answered explicitly rather than by default.

## The bug: `$BASE_BRANCH` doesn't honor the override on configured repos

`workflow.ts:1642/1651` passes the override as `ExecuteWorkflowOptions.baseBranch`, but `executor.ts:934-936` only reads that **if `config.baseBranch` is falsy**. So on any repo that sets `worktree.baseBranch` in `.archon/config.yaml`:

- the worktree is cut from `--base` ✅
- `$BASE_BRANCH` still reports the **configured** branch ❌

A prompt referencing `$BASE_BRANCH` is then told it's working from a branch the worktree wasn't cut from. That's worse than the flag not existing — it silently misinforms an AI node, and the PR's central invariant ("the override drives both") is false for exactly the users who configured a base branch.

CodeRabbit's proposal is the right fix and I'd take it: add a dedicated `baseOverride?: string` to `ExecuteWorkflowOptions` (`executor.ts:376-384`), rank it **above** `config.baseBranch` at `:930-936`, pass `flagBase` through it, and keep `baseBranch: codebaseDefaultBranch` as the fallback. The doc comment at `:380-383` needs updating too — it currently describes the old precedence.

Please also add the test whose absence let this through CI: `executeWorkflow` with `config.baseBranch = 'main'` **and** the override set must resolve `$BASE_BRANCH` to the override.

## The design question: should one flag drive both cut-from and PR-target?

Right now it does, deliberately, and your reasoning for it is sound as a *default*. But they're separable concerns — branching from `release/2.0` while targeting `dev` is a legitimate thing to want, and there's currently no way to express it.

I'm not asking you to add a second flag. I'd like the conflation documented as an intentional decision, so the next person hitting that case finds an answer rather than assuming it's a bug. If you think a separate PR-target override is worth having, say so and file it — I'd read that proposal.

## Four unguarded divergence paths

Each of these currently either silently diverges or warns against the wrong base. Guard or document each:

- **`--base` + `--from`** — silent divergence today. Decide which wins and enforce it.
- **`--base` under `worktree.enabled: false`** — mirror the existing `--from` rejection at `workflow.ts:822-828`.
- **`--base` on a folder project** — belongs in `assertNoWorktreeOptionsForFolder` (`workflow.ts:410-419`).
- **`--base` on worktree reuse** — mirror the `--from` "was not applied" warning at `workflow.ts:1293-1302`, and feed the override into the reuse base check at `:1304-1313` so it stops comparing against the wrong base.

## Rebase and docs

The two conflicts are mechanical, but one has a trap: in `worktree.ts`, **keep `dev`'s 3-arg `syncWorkspaceBeforeCreate(repoPath, preferredBaseBranch, remote)`** (from #2234). Resolving toward the PR side silently reverts remote-awareness, so a repo with a non-`origin` remote would start syncing against the wrong one.

Docs need the `--base` row in `packages/docs-web/src/content/docs/reference/cli.md:209` — including its relationship to `--from` and the requirement that the remote already has the branch — plus the `$BASE_BRANCH` resolution order in `reference/variables.md:23` and `book/quick-reference.md:76`.

And please fill in `.github/PULL_REQUEST_TEMPLATE.md` and link an issue, or open one to link.

---

Worth noting this makes a **fourth** way to determine the base branch, alongside repo config, git auto-detection, and the `$BASE_BRANCH` variable. That's fine — a per-dispatch override is genuinely the missing one — but it means the precedence chain has to be written down somewhere a user will find it, which is why the docs items above aren't optional polish.

buun-dev:
All six items addressed. Branch rebased onto `dev`; three commits now — the original feature, a fix commit, and a docs commit, kept separate so the response to this review is diffable on its own.

## The bug — confirmed and fixed

Reproduced it as a failing test before touching the fix:

```
Expected: "epic/foo"
Received: "main"
```

Took CodeRabbit's proposal as you suggested: `ExecuteWorkflowOptions.baseOverride` (`executor.ts`), ranked above `config.baseBranch`, with `baseBranch: codebaseDefaultBranch` kept as the fallback. Doc comment at the type updated — it described the old precedence.

The missing test is `prefers baseOverride over repo config baseBranch` in **`packages/workflows/src/executor.test.ts`**, not the CLI package. That placement is the actual reason CI missed this: `packages/cli` mocks `executeWorkflow`, so the resolution never runs there and the broken seam is unreachable.

Worth flagging what the existing CLI test was doing — it asserted `expect(opts.baseBranch).toBe('epic/foo')`, which passed *because* the flag was routed through the low-priority field. It was asserting plumbing rather than outcome, so it locked the defect in rather than merely missing it. Corrected to assert `baseOverride` plus `baseBranch === 'develop'`, mirroring the provider-side assertion above it.

## The design question — one flag, and `--from` is the decoupling lever

Documented as intentional, and I'd argue the separable case is already expressible, so no second flag is warranted.

`worktree.ts:1089-1093` gives `--from` precedence over the resolved base for the start-point, while `--base` continues to drive `$BASE_BRANCH`. Your own example is therefore:

```bash
archon workflow run implement --from origin/release/2.0 --base dev
# cut from release/2.0, PR targets dev
```

`--from` reaches `git worktree add` verbatim, so a remote ref works. One honest caveat, now documented: sync-before-create refreshes the `--base` branch, not the `--from` start-point, so pass a remote ref when the local copy may be stale.

That answer is locked in by a test rather than left to prose — `lets --from win the cut-from while --base still drives the PR target` — so the next person to read it as a bug gets a named contract instead of a guess.

## Divergence paths — three guarded, one deliberately not

| Path | Resolution |
|---|---|
| `--base` + `--from` | **Left unguarded, by design** — see above. Enforcing a winner would delete the only way to express the separable case. |
| `worktree.enabled: false` | Rejected, mirroring the `--from` rejection |
| Folder project | Rejected via `assertNoWorktreeOptionsForFolder` |
| Worktree reuse | Warns, and the reuse base check now takes the override |

Two notes on the reuse path.

The warning wording deliberately differs from `--from`'s. `--from` on reuse is genuinely inert, so "was not applied" is accurate. `--base` is only *half* ignored — the cut-from is frozen but `$BASE_BRANCH` still moves — so it reads "did not change the cut-from …; it still applies to the PR target." Copying the `--from` sentence would have replaced one silent surprise with a louder one.

On the folder path: this one bites harder than it looks. A folder run creates no worktree, so `--base` can't affect a cut-from — but it still reaches `$BASE_BRANCH`, because the override short-circuits the `isFolderCodebase` branch in `executor.ts`. Half-applied, so rejected outright.

## Rebase

Done. The `worktree.ts` hunk is resolved as a **union**, not a side: `resolveRemote` + the 3-arg `syncWorkspaceBeforeCreate(repoPath, preferredBaseBranch, remote)` from #2234, *and* the `baseOverride ?? config ?? baseBranch` precedence.

The trap is real and it announced itself — the PR's own isolation test failed after the rebase because it asserted the pre-#2234 two-arg options shape, and the merged code correctly passes `remote: 'origin'`. Its sibling test three lines above already asserted that, so I matched it. Both facts are now in one call assertion, so a future resolution that drops either one fails CI.

## Docs

All three locations, plus a new **Base branch precedence** section under `workflow run` covering the two meanings of "base", the four ranked sources, why levels 2–4 being repo-static made a per-dispatch level necessary, the `--from` recipe, the rejection paths, and reuse behavior.

- `reference/cli.md` — `--base` row + the precedence section
- `reference/variables.md` — `$BASE_BRANCH` resolution order
- `book/quick-reference.md` — same, one line

On your closing point about this being the fourth mechanism: agreed, and it's why the chain is now written down in the three places a user actually lands rather than only in the flag row. I also stated the remote requirement explicitly — a `--base` branch missing on the remote is already a hard error from `syncWorkspace` (`repo.ts:177`), not a silent fallback, which was worth documenting since the alternative would have been the more dangerous behavior.

I disambiguated the adjacent `--from` row too — it read "Override base branch", which became ambiguous the moment a second flag could claim the same words.

## PR template and issue

PR body rewritten onto `.github/pull_request_template.md` with every required section filled, including the two diagrams and the connection inventory. Issue filed as #2355 and linked via `Closes`.

One thing called out in Validation Evidence rather than buried: `bun run test:install` aborts `bun run validate` on Windows (the installer refuses non-WSL2 hosts), so I ran the remaining stages directly and listed them. Three unrelated suite failures exist (symlink and download tests); I reproduced each on an untouched checkout of the same base to confirm they're pre-existing and environmental rather than mine.

buun-dev:
Follow-up: `f6c4c4de` adds a **fifth** half-applied path that neither of us listed. CodeRabbit caught it on the re-review pass, and it holds up — I verified it against the code before taking it.

**`--resume` + `--base`.** The resume branch adopts `resumable.working_path` (`workflow.ts:1120-1184`) and never inspects `flagBase` — but `flagBase` still reaches `executeWorkflow` as `opts.baseOverride`. So `$BASE_BRANCH` moves while the resumed worktree's cut-from stays frozen: the same shape as the `--branch` reuse case, on the path my own guard missed. That one is mine, not the original PR's — I scoped the reuse guard to `--branch` without checking whether `--resume` lands in the same state.

The warning is now extracted into `warnBaseOverrideOnReuse()` and called from both sites instead of duplicated. Two call sites that have to agree about a subtle distinction — half-applied, not ignored — are exactly where a copy drifts. Docs updated to cover both.

Worth flagging separately, because it outlives this PR: writing the test surfaced a latent hazard in the CLI suite. The `--resume` path calls `isolationDb.listByCodebase`, which was missing from `mock.module('@archon/core/db/isolation-environments')`. Since `mock.module()` merges over the real module rather than replacing it, that omission left the **real** implementation in place — no test had reached that line before, so the first one to exercise `--resume` far enough would have opened a live SQLite handle inside a unit suite rather than failing loudly. That is #2240's failure mode exactly, and it stalls to a 5s timeout instead of erroring, so it would have surfaced as an unattributable CI flake. Stubbed here; there may be other unreached exports in that factory worth an audit.

Prior state for reference: all five checks passed on `f60b1913` before this commit, including `test (windows-latest)`.

Issue #2080, opened 2026-07-13
Windows reliability: verify the Modern Standby + --detach fixes on real hardware (code merged in #2063)


Wirasm:
## Summary

PR #2063 fixes two independent Windows failure modes that killed or stalled unattended workflow runs. This issue preserves the distilled root-cause evidence (the PR originally carried two full investigation dossiers in `docs/`, removed during review because they referenced the contributor's private environment — the verbatim versions remain in the PR's file history at commit `2800552b`) and tracks the remaining on-hardware verification.

## Failure mode 1 — Modern Standby freezes in-flight runs (exit 66)

- On an unattended native-Windows host, the screen turning off drops the machine into Modern Standby (S0 Low Power Idle), freezing the executor mid-DAG.
- On thaw, every new `bash` child the executor spawns exits `66` (`EX_NOINPUT`) with no stdout/stderr, collapsing the DAG tail; other casualties present as zombie `running` rows.
- Evidence (from the dossier): 7/7 exit-66 runs correlated with Kernel-Power 506/507 (standby entry/exit) events; at one capture ~44% of all runs on the host had not completed (209 completed / 82 cancelled / 81 failed).
- Fix in #2063: refcounted keep-awake (`SetThreadExecutionState(ES_CONTINUOUS | ES_SYSTEM_REQUIRED)` via `bun:ffi`) held for exactly the executing window of `executeWorkflow`. Best-effort: no-op off Windows or on FFI failure; display may still sleep.

## Failure mode 2 — `--detach` never actually detached

- `spawnDetachedWorkflowRun` used `Bun.spawn(...) + child.unref()`, which does **not** put the child in its own process group. On Windows the "background" child was killed ~1s in (at `worktree_creating`) when the launching shell/console tore down — the detach log stops dead mid-line with no error.
- Explicitly **not** caused by standby: event logs showed no standby entry during the failed runs.
- Fix in #2063: `node:child_process` `spawn(..., { detached: true, windowsHide: true })`, plus an `'error'` listener and a `child.pid` check (mirroring `setup.ts`'s `trySpawn`) so a failed spawn errors instead of printing a false success ack.

## Remaining verification (needs a native-Windows host)

- [ ] Live `--detach` survival smoketest: dispatch a run with `--detach`, close the launching console, confirm the run progresses past `worktree_creating` to completion.
- [ ] Real Modern-Standby cycle with a run in flight: let the screen turn off, confirm the system stays awake and the run completes (no exit-66 tail).
[...]

Issue #2180, opened 2026-07-20
fix(workflows): make concurrent governed launches and approval admission safe


Wirasm:
## Problem

Two live concurrency hazards remain when independent nodes in one DAG layer launch governed work or request approval.

### Concurrent governed launches can collide on the working path

Two ordinary `workflow:` nodes may start child runs against the same inherited working path. If the child may mutate the checkout, the path lock lets one child win and self-cancels the sibling. That cancellation then poisons parent resume.

Dynamic `fan_out` now has an explicit preflight, but hand-authored same-layer `workflow:` siblings still need the equivalent safe contract.

### One parent run has one approval slot

A workflow run stores one active approval context. Concurrent `approval:` nodes, interactive loop gates, or 1:1 child runs that pause at gates can request that slot in the same layer.

Today the losing request is inconsistent:

- the tolerant approval/loop path can treat the lost pause transition as success after already emitting an unreachable second prompt;
- the child-sub-run path can emit a node failure;
- neither behavior represents the author's independent gates faithfully.

These are different mechanisms, but they define one author-facing contract: concurrent governed work must either be admitted safely or refused before it creates an unrecoverable state.

## Decisions

- The engine never infers worktree isolation.
- Read-only children may share the parent checkout when their workflow declares `mutates_checkout: false`.
- Mutating children run concurrently only with explicit isolation; otherwise the engine refuses the unsafe layer with actionable alternatives.
- One approval is presented at a time. Concurrent approval requests are admitted sequentially rather than overwritten, swallowed, or presented as multiple prompts for one run id.
- A gate decision applies only to the gate that was presented.
- Ambiguous live work is surfaced for operator action and is never autonomously failed, cancelled, or duplicated.

## Scope

- Same-layer 1:1 `workflow:` nodes using the inherited working path.
- Plain `approval:` nodes.
- Interactive loop gates.
- A parent waiting on multiple independently paused 1:1 child runs.
- Resume and cold-resume behavior across sequentially admitted gates.
- Production-faithful tests for the run-status compare-and-set behavior.

## Non-scope

- An approval gate inside a `fan_out` child; fail-before-spawn is tracked by #2474.
- Automatically selecting `isolation: worktree`.
- Racing or cancelling independent siblings.
- Approval coordination across unrelated top-level runs.
- Distributed scheduling across Archon processes.

## Acceptance

- [ ] Concurrent read-only child workflows may share a checkout without path-lock cancellation.
[...]

PR #2109, opened 2026-07-14
fix(cli): accept short run-id prefixes in workflow run-addressing commands

Wirasm:
## Summary

- Problem: `archon workflow runs` prints 8-char run ids (`0b1ee8da`), but `workflow resume <id>` (and `get`/`abandon`/`approve`/`reject`) do exact-match lookups, so the displayed id fails with `Workflow run not found` — users had to fall back to `--json` for the full id.
- Why it matters: the natural copy-paste flow (`runs` → `resume`) is broken, and it's inconsistent with chat, where the `manage_run` tool already accepts short prefixes.
- What changed: the five run-addressing CLI subcommands now resolve unique short-id prefixes via the existing `findWorkflowRunsByIdPrefix` core helper, mirroring `manage_run`'s semantics — scoped to the cwd's codebase, unique match resolves, ambiguous prefix errors, full ids (dashed UUID or 32-char SQLite hex) skip resolution and keep working globally. Plus a docs note in the CLI reference.
- What did **not** change (scope boundary): `findWorkflowRunsByIdPrefix` and `manage-run-tool.ts` are untouched (imported as-is); chat `/workflow resume` prefix acceptance is a possible follow-up; the `runs` listing format keeps short ids.

### Before

```
  User                          Archon CLI
  ────                          ──────────
  workflow runs ──────────────▶ prints "0b1ee8da  failed  implement"
  copies 0b1ee8da
  workflow resume 0b1ee8da ───▶ exact lookup WHERE id = '0b1ee8da'
  sees error ◀───────────────── "Workflow run not found: 0b1ee8da"
  workflow runs --json ───────▶ digs out the full id manually
  workflow resume <full-id> ──▶ finally resumes
```

### After

```
  User                          Archon CLI
  ────                          ──────────
  workflow runs ──────────────▶ prints "0b1ee8da  failed  implement"
  copies 0b1ee8da
  workflow resume 0b1ee8da ───▶ [resolveRunIdArg: prefix → full id,
                                 scoped to this project's codebase]
  resumes ◀──────────────────── executes with the resolved full id
                                 (*ambiguous prefix → explicit error*)
```

### Before

```
  cli.ts ──▶ workflow.ts (get/resume/abandon/approve/reject)
                 │ raw runId
                 ▼
             @archon/core ops + db (exact WHERE id = $1)

  chat manage_run ──▶ findWorkflowRunsByIdPrefix (core db)   ← prefix-aware
```

### After

```
[...]

Issue #2097, opened 2026-07-14
workflow runs prints truncated run IDs that workflow resume rejects


Wirasm:
`archon workflow runs` displays 8-character run IDs (e.g. `0b1ee8da`), but `archon workflow resume <id>` requires the full 32-char ID and fails with `Workflow run not found: 0b1ee8da` when given the displayed prefix.

Observed flow: run fails → `workflow runs --status failed` to find it → copy the displayed ID → `workflow resume 0b1ee8da` → error. The user has to fall back to `--json` to get the full ID.

Other CLI commands (`workflow cancel`, chat `/workflow cancel 68ca33e1`) appear to accept short prefixes in their user-facing messages (e.g. the worktree-in-use error suggests `/workflow cancel 68ca33e1`), so resume rejecting prefixes is inconsistent.

Suggested fix: accept unique ID prefixes across all run-addressing commands (resume/abandon/approve/reject/get) — resolve prefix → unique match, error on ambiguity — or print full IDs in `workflow runs` output.

PR #2204, opened 2026-07-21
feat(cli): extend --detach to workflow approve/reject/resume control verbs

### LINKED ISSUE #2645: feat(cli): detach an exact workflow run resume
## Problem

Archon has two CLI resume paths with different targeting and execution behavior:

- `archon workflow resume <run-id>` targets one exact known run, but executes only in the foreground.
- `archon workflow run <name> --resume --detach` can execute in the background, but selects the newest resumable run for that workflow in the current checkout.

That split makes a known-run recovery less safe when it needs to be detached. An operator or orchestrator that already has the run ID must discard it and switch to a workflow-name/current-path lookup. From a different worktree, the shortcut correctly finds no run; from a checkout with multiple historical runs, it expresses less intent than the ID the caller already possessed.

Observed while dogfooding `fix` on #2480: the failed run belonged to its issue worktree, while the orchestrator was in the workflow-authoring PR worktree. `workflow run fix --resume --detach` found no resumable run. `workflow resume 942cc47c…` targeted the right run immediately, but had to remain attached.

## Intent

Allow an exact run ID to remain the identity for detached recovery. `workflow resume <run-id> --detach` should execute that run in the background with the same exact-run lookup and safety checks as foreground resume.

When Archon already knows a run ID, help text, recovery output, and agent-facing guidance should prefer the exact-ID form. Keep the workflow-name shortcut as a convenience for a human intentionally asking for “the latest failed run of this workflow in this checkout”; do not use it as the recovery path for a known run.

## Acceptance

- An operator can resume one known failed or paused run by full or unambiguous short ID without blocking the invoking terminal.
- Detached exact-ID resume reuses the recorded worktree and inputs exactly as foreground exact-ID resume does.
- A detached resume returns enough information to track the same run; it does not create or select a different run by workflow name/current path.
- Existing `workflow run <name> --resume` path scoping remains explicit and is not presented as equivalent to exact-ID recovery.

buun-dev:
## Summary

- **Problem:** `--detach` existed only on `workflow run`. The control verbs (`approve`, `reject`, `resume`) hosted the executor in the calling shell, so a reaped shell — a harness task, a closed terminal, a dropped SSH session — left the run wedged mid-resume: approval recorded, workflow never continued.
- **Why it matters:** These are exactly the commands automation drives. An approve that records the decision and then dies leaves a run paused forever with no signal that anything went wrong.
- **What changed:** `--detach` now works on all three control verbs. The parent validates the run **read-only** and spawns a detached child that owns every state mutation in its own process group. The precondition gate is extracted into `assertApprovable`/`assertRejectable` in `@archon/core` so the parent's precheck and the operation itself are literally the same code. `--detach --json` gains `continues: true`, and the child is now spawned with the caller's `--cwd`.
- **What did *not* change (scope boundary):** No change to the non-detached paths of any verb. No change to `resumeWorkflow`'s gate (its precheck already called the shared operation). No change to the chat/HTTP/`manage_run` surfaces beyond the behavior-preserving extraction. `workflow run --detach` is untouched.

### Before

```
  Operator                 archon (your shell)          Workflow run
  ────────                 ───────────────────          ────────────
  approve <id> ─────────▶  records approval ──────────▶ paused → resolved
                           resumes inline
                           runs the executor ─────────▶ running ...........
        shell dies ──────▶ ✗ process killed              ✗ WEDGED
                                                          (approved, never resumed)

  approve <id> --detach    [not supported — flag only existed on `workflow run`]
```

### After

```
  Operator                 archon parent            detached child          Workflow run
  ────────                 ─────────────            ──────────────          ────────────
  approve <id> --detach ▶  [*precheck: READ-ONLY*]
                           4 gates, zero writes
                              │
                              ├─ refused? ──────────────────────────────▶ untouched
                              │  {ok:false} to your face, NOTHING spawned
                              │
                              └─ ok ─▶ spawn ─────▶ records approval ────▶ paused → resolved
                           {ok:true,                resumes inline
                            continues:true}         runs executor ───────▶ running ......
                           exits                       │
        shell dies ──────▶ (already gone)              └─ keeps running ──▶ ✓ completed
```

### Before

```
  packages/cli/src/commands/workflow.ts
    workflowApproveCommand ──┐
    workflowRejectCommand  ──┼─▶ runDetachedControlCommand
    workflowResumeCommand  ──┘        │
                                      ├─▶ [inline precheck: status ONLY]   ← partial copy
                                      └─▶ spawnDetachedWorkflowRun(process.cwd())
                                                                  ▲
                                                     caller's --cwd discarded here

  packages/core/src/operations/workflow-operations.ts
    approveWorkflow ──▶ [inline gate: status, context, child_workflow, resolved]
    rejectWorkflow  ──▶ [inline gate: status, child_workflow, resolved]

  The two gates are unrelated code. They already drifted 3 checks apart.
```

### After

```
  packages/cli/src/commands/workflow.ts
    workflowApproveCommand ──┐
    workflowRejectCommand  ──┼─▶ runDetachedControlCommand(runId, action, json, [+]cwd, precheck)
    workflowResumeCommand  ──┘        │
                                      ├─▶ precheck ===▶ assertApprovable / assertRejectable
                                      └─▶ spawnDetachedWorkflowRun(cwd ?? process.cwd())  [~]

  packages/core/src/operations/workflow-operations.ts
    [+] assertApprovable(run): ApprovalContext            ◀=== CLI precheck
    [+] assertRejectable(run): ApprovalContext | undefined ◀=== CLI precheck
    approveWorkflow [~] ===▶ assertApprovable
    rejectWorkflow  [~] ===▶ assertRejectable

  ONE gate. The parent cannot validate less than the child enforces.
```

| `cli/commands/workflow.ts` (approve precheck) | `core/operations/workflow-operations.assertApprovable` | **new** | Replaces an inline status-only copy of one of four gates |
| `cli/commands/workflow.ts` (reject precheck) | `core/operations/workflow-operations.assertRejectable` | **new** | Replaces an inline status-only copy of one of three gates |
| `core.approveWorkflow` | `core.assertApprovable` | **new** | Extraction; error strings copied verbatim |
| `core.rejectWorkflow` | `core.assertRejectable` | **new** | Extraction; error strings copied verbatim |
| `runDetachedControlCommand` | `spawnDetachedWorkflowRun` | **modified** | Now passes the command's `cwd`, not `process.cwd()` |
| `cli/cli.ts` | `workflow{Approve,Reject,Resume}Command` | **modified** | Threads `--detach` (5th arg, after upstream's `cwd`) |
| `cli/commands/workflow.ts` (resume precheck) | `core.resumeWorkflow` | unchanged | Already called the shared operation — its single gate was never duplicated |
| `core.{approve,reject}Workflow` | `db/workflows.resolveApprovalGate` | unchanged | The CAS remains the real arbiter for concurrent resolution |
| chat `manage_run` / HTTP API | `core.{approve,reject}Workflow` | unchanged | Same behavior; they inherit the extraction transparently |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: M`
- Scope: `cli`, `core`, `docs`
- Module: `cli:workflow`, `core:operations`

## Change Metadata

- Change type: `feature` (with an embedded `bug` fix for `cwd`, and a `refactor` for the shared gate)
- Primary scope: `cli`

## Linked Issue

- Closes #2645 — *feat(cli): detach an exact workflow run resume*. The issue was filed on 2026-08-19, after this PR opened, and this PR implements it. Checked against its acceptance list: a known run resumes by full or unambiguous short id without blocking the terminal; the detached exact-id path reuses the recorded worktree and inputs exactly as the foreground path does; the ack identifies the same run rather than selecting a different one by name and current path; and the name-form recovery pointers in `cli.md`, the `--json` code comment and `AGENTS.md` are repointed at the exact-id form, so the two are no longer presented as equivalent.
- Related #2279 — **closed 2026-08-03.** The detach ack used to be gated only on the child having a PID; the startup window added there catches a child that dies in its first 500 ms. This PR's control verbs `await` the spawn and therefore inherit that fix. Worth stating explicitly because an earlier revision of this branch called the spawn synchronously and did reopen the hole on the new surface.
- Related #2080 — on-hardware Windows detach smoketest. Its checklist covers `workflow run --detach` only, while this PR widens the surface to three more verbs, so the checklist arguably wants widening too.
- Depends on # — none. (Rebased onto `dev` at `3f78981f`.)
- Supersedes # — none.

```bash
bun run check:bundled && bun run check:bundled-skill && bun run check:bundled-schema \
  && bun run check:pi-vendor-map && bun run check:capability-matrix     # all OK
bun run type-check                                                      # 10/10 packages clean
bun run lint --max-warnings 0                                           # clean
bun run format:check                                                    # clean
bun test packages/cli/src/commands/workflow.test.ts                     # 199 pass / 0 fail
bun test packages/core/src/operations/workflow-operations.test.ts       #  49 pass / 0 fail
bun test packages/core/src/orchestrator/manage-run-tool.test.ts         #  30 pass / 0 fail (unchanged)
bun run test                                                            # see caveat below
```

- **Evidence provided:**
  - CLI suite `packages/cli/src/commands/workflow.test.ts`: **199 pass / 0 fail** (193 pre-change baseline + 6 new).
  - Core suite `packages/core/src/operations/workflow-operations.test.ts`: **49 pass / 0 fail** (41 baseline + 8 new).
  - The two **pre-existing** `child_workflow` refusal tests pass unchanged. That is the proof the gate extraction is behavior-preserving — the error strings were copied verbatim, so no caller-visible message changed.
  - `manage-run-tool.test.ts` (the only file that `mock.module`s `workflow-operations`) passes unchanged at 30/30. Because `mock.module()` merges rather than replaces, the two new exports keep their real implementations there; `manage-run-tool.ts` calls only the mocked operations, never the validators.
  - Compiled-binary verification: see **Human Verification** below.

- **If any command is intentionally skipped, explain why:**
  - `bun run validate` **cannot complete on a Windows host**: the chain reaches `bun run test:install`, and `scripts/test-install.sh` exits immediately with `Windows is not supported. Please use WSL2`. The eight checks before it all pass; the `&&` chain then short-circuits, so `bun run test` was run separately.
  - `bun run test` (full per-package sweep, all 8 test-bearing packages) produced **exactly one** failure: `@archon/providers` → `pathKind > returns "missing" for a broken symlink without throwing`, failing with `EPERM: operation not permitted, symlink`. That is Windows refusing symlink creation without elevation/Developer Mode — an environment limit in a package this branch does not touch. Every other package passed.

## Security Impact (required)

- New permissions/capabilities? **No.** `--detach` re-invokes the same CLI with the same argv minus `--detach`/`--json`; the child has exactly the parent's authority.
- New external network calls? **No.**
- Secrets/tokens handling changed? **No.** No credential path is read, written, logged, or forwarded differently.
- File system access scope changed? **Yes — narrowed.** `runDetachedControlCommand` previously spawned into `process.cwd()` regardless of the caller's `--cwd`. It now spawns into the `cwd` the caller actually asked for. This *reduces* the set of directories a detached child can land in and removes a case where the child could be started outside any git repo after the parent had already acked success. The run's `working_path` remains a non-candidate by design: a container run's `working_path` is a distro path the host cannot spawn into, so the child re-resolves everything by run id.

## Compatibility / Migration

- Backward compatible? **Yes.** `--detach` is purely additive on all three verbs; without it every verb behaves exactly as before. `continues` is an additive field on an ack that only exists under `--detach --json`, a combination that shipped nowhere before this PR. The extracted validators throw the same error strings the inline checks did.
- Config/env changes? **No.**
- If yes, exact upgrade steps: n/a.

Verified against a **compiled Windows binary**, not `bun run cli` — dev mode is precisely where the Bun single-file-executable argv bug never appears, so dev-mode evidence cannot cover it.

> **Stale, read with care.** This section describes a binary built at commit `6af8eb2a`, which the rebase onto current `dev` removed. It exercised the pre-#2390 shape of this branch and does **not** cover the code now shipping — in particular it predates awaiting the spawn and the shared control-verb validator. Those are covered by unit tests, including a child dying inside the startup window. Binary-level Windows evidence for the current head would have to be re-run; that gap is #2080.

Binary: `bun build --compile --minify --target=bun-windows-x64` with `BUNDLED_IS_BINARY = true`; self-reports `Archon CLI v0.6.0 / Platform: win32-x64 / Build: binary / Git commit: 6af8eb2a`. Driven against an isolated `ARCHON_HOME` (throwaway SQLite DB) and a throwaway two-bash-nodes-around-one-approval-gate workflow, so the runs are real and cost no AI tokens.

- **Verified scenarios:**

  1. `approve <id> --detach --json` → ack `{ok:true, detached:true, continues:true}`. Child log contains **no** `Unknown command: /$bunfs/root/…` and **no** `Unknown command: B:/~BUN/root/…`. Child log shows `[after] Completed (118ms)` → `past the gate` → `Workflow completed successfully.`; `workflow get` reports `"status": "completed"`. **The run genuinely advanced — the ack alone would not have proven this, since a pre-fix binary acks success too.**
  2. `reject <id> --detach --json` → ack `continues: true`; no `Unknown command`; child log `Rejected and cancelled: smoke-detach-gate`; `workflow get` reports `"status": "cancelled"`.
  3. `resume <id> --detach --json` → ack `continues: true`; no `Unknown command`; child re-invoked, logged `[before] Skipped (prior_success)` and re-paused at the gate as expected.
  4. **Precheck refusal on a gate this PR newly enforces in the parent** — the important one. Recorded an approval with plain `--json` (run stays paused, gate resolved), then ran `approve <id> --detach --json` again: `{"ok": false, "error": "Workflow run … was already approved and is awaiting resume."}`, and that run's child-log file stayed byte-identical (1965 → 1965), proving **nothing was spawned**. Before this PR the parent acked `{ok:true}` here and the child died unseen.
  5. Precheck refusal on the pre-existing status gate — `approve <completed-run-id> --detach --json` → `{"ok": false, "error": "Cannot approve run with status 'completed'. Only paused runs can be approved."}`; log-file count unchanged (3 → 3) and that run's log byte-identical (1718 → 1718). Nothing spawned.

- **Edge cases checked:** `reject` deliberately tolerates a malformed/absent approval context where `approve` refuses (covered by unit test — it must still spawn); the `child_workflow` redirect is refused synchronously for both verbs (unit tests); `cwd: undefined` still falls back to `process.cwd()` so the pre-existing detach test's `expect(spawnOptions?.cwd).toBe(process.cwd())` assertion passes untouched.

- **What was not verified:** the `child_workflow`-blocked-parent redirect was not reproduced against the compiled binary (it needs a `workflow:` sub-run tree and is covered by unit tests at both layers, including two pre-existing tests that pass unchanged). Linux/macOS binaries were not built — the argv fix being relied on (#2273) is platform-independent, but only `win32-x64` was exercised end-to-end here. `bun run validate` was not run to completion on this host (see Validation Evidence).

- **Affected subsystems/workflows:** `assertApprovable`/`assertRejectable` now sit on the `approveWorkflow`/`rejectWorkflow` path used by **every** surface — CLI, chat command handler, HTTP API, and the `manage_run` native tool. That is the honest blast radius. It is mitigated by the extraction being behavior-preserving: error strings are verbatim copies, and the pre-existing `child_workflow` refusal tests for both operations pass unchanged.
- **Potential unintended effects:** the `--detach` precheck is now strictly *stricter* than before. A caller who previously got `{ok:true}` for a run with a missing approval context, a `child_workflow` gate, or an already-resolved gate will now get `{ok:false}` synchronously. That is the intended fix, but it is a visible behavior change for anyone who was (unknowingly) relying on the false success.
- **Guardrails/monitoring for early detection:** every refusal path is unit-tested to assert **zero** spawns, so a regression that starts spawning on a refused precheck fails CI rather than silently leaking children. The `continues` field gives automation a machine-readable signal instead of inferred behavior.

## Rollback Plan (required)

- **Fast rollback command/path:** revert the PR. `--detach` disappears from `approve`/`reject`/`resume`; `workflow run --detach` is untouched; the operations revert to their inline gates with identical messages. No data migration, no config, nothing persisted by this change.
- **Feature flags or config toggles (if any):** none — `--detach` is itself the opt-in. Every verb without the flag behaves exactly as before.
- **Observable failure symptoms:** a control verb acking `{ok:true, detached:true}` while `workflow get` shows the run unchanged minutes later; or a `detached-run-<id>.log` containing `Unknown command:` — both indicate the child failed to re-invoke.

## Risks and Mitigations

- **Risk:** the shared validators are on the hot path of every approve/reject surface, so a mistake in the extraction breaks far more than the CLI.
  - **Mitigation:** error strings copied verbatim; the two pre-existing `child_workflow` refusal tests and the full 49-test core operations suite pass unchanged; eight new unit tests pin each gate's message individually.
- **Risk:** adding exports to `workflow-operations` silently un-mocks them in any test that `mock.module`s that path, because Bun's `mock.module()` merges rather than replaces — the documented cause of #2240.
  - **Mitigation:** audited — exactly one factory exists (`manage-run-tool.test.ts`), it mocks only the four operations, and `manage-run-tool.ts` never calls the validators directly. Suite verified passing at 30/30.
- **Risk:** tightening the precheck turns previously "successful" detach calls into synchronous failures for automation that relied on the false success.
  - **Mitigation:** this is the bug being fixed, and the failure is now loud, immediate, and carries the same message the child would have produced. Documented in the CLI reference under *Detached control verbs*.
- **Risk:** `--detach --json` continues the run while bare `--json` does not — a genuine asymmetry that could surprise a caller.
  - **Mitigation:** kept deliberately (withholding continuation under `--detach` would spawn a child that only records a decision, which is what bare `--json` already does), documented in both the CLI reference and `CLAUDE.md`, and made machine-readable via `continues: true`. A unit test asserts the child argv carries neither `--detach` nor `--json`.

## Summary by CodeRabbit

* **New Features**
  * Added `--detach` support for workflow `approve`, `reject`, and `resume` commands.
  * Detached commands validate immediately, then continue independently in the background.
  * Added JSON acknowledgements with continuation status, startup failure reporting, and log details.
* **Bug Fixes**
  * Preserved validation errors and prevented parent processes from mutating workflow state during detached execution.
* **Documentation**
  * Updated CLI help and documentation with detached workflow usage, behavior, output, and run-ID guidance.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -203,6 +203,7 @@ The rules `--help` will not tell you:
 - Isolation is the default for `workflow run` — it creates a worktree unless you pass `--no-worktree` or `--folder`.
 - `--json` is supported on the read/write subcommands (`list`, `status`, `runs`, `get`, `approve`, `reject`, `abandon`, `resume`). On `approve`/`reject`/`resume` it records the decision and returns an ack **without** the inline auto-resume, leaving the run resumable — drive continuation separately.
 - `--detach` works on `approve`/`reject`/`resume` as well as `run`. Adding it INVERTS the `--json` rule above: the parent validates the run read-only and refuses synchronously, then spawns a child without `--json`, so the child takes the inline path and DOES continue the run. The ack carries `continues: true` so an automation knows it no longer owns continuation.
+- When you already hold a run id, recover with `workflow resume <run-id> --detach`, not `workflow run <name> --resume --detach`. The name form selects the newest resumable run of that workflow **in the current checkout**, so it finds nothing from another worktree and expresses less than the id you already have (#2645).
 
 ## Architecture
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,6 +202,7 @@ The rules `--help` will not tell you:
 - Workflow and isolation commands must run inside a git repository; subdirectories resolve to the repo root. `--folder` is the escape hatch for a non-git directory.
 - Isolation is the default for `workflow run` — it creates a worktree unless you pass `--no-worktree` or `--folder`.
 - `--json` is supported on the read/write subcommands (`list`, `status`, `runs`, `get`, `approve`, `reject`, `abandon`, `resume`). On `approve`/`reject`/`resume` it records the decision and returns an ack **without** the inline auto-resume, leaving the run resumable — drive continuation separately.
+- `--detach` works on `approve`/`reject`/`resume` as well as `run`. Adding it INVERTS the `--json` rule above: the parent validates the run read-only and refuses synchronously, then spawns a child without `--json`, so the child takes the inline path and DOES continue the run. The ack carries `continues: true` so an automation knows it no longer owns continuation.
 
 ## Architecture
 
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -171,7 +171,7 @@ Options:
   --verbose, -v              Show debug-level output
   --json                     Output machine-readable JSON (list/status/get/runs/approve/reject/abandon/resume)
   --events                   For verbose JSON status/get: output raw event rows instead of node summaries
-  --detach                   Run 'workflow run' in a detached background child (returns immediately)
+  --detach                   Run 'workflow run'/'approve'/'reject'/'resume' in a detached background child (returns immediately)
   --all                      For 'workflow runs': list across all projects (ignore cwd scope)
   --status <status>          For 'workflow runs': filter to one status (running, completed, failed, ...)
   --limit <n>                For 'workflow runs': max rows (default 20)
@@ -679,7 +679,7 @@ async function main(): Promise<number> {
               console.error('Usage: archon workflow resume <run-id>');
               return 1;
             }
-            await workflowResumeCommand(resumeRunId, jsonFlag, effectiveCwd);
+            await workflowResumeCommand(resumeRunId, jsonFlag, effectiveCwd, detachFlag);
             break;
           }
 
@@ -706,7 +706,13 @@ async function main(): Promise<number> {
             const rawApproveComment =
               (values.comment as string | undefined) || positionals.slice(3).join(' ');
             const approveComment = rawApproveComment.length > 0 ? rawApproveComment : undefined;
-            await workflowApproveCommand(approveRunId, approveComment, jsonFlag, effectiveCwd);
+            await workflowApproveCommand(
+              approveRunId,
+              approveComment,
+              jsonFlag,
+              effectiveCwd,
+              detachFlag
+            );
             break;
           }
 
@@ -719,7 +725,13 @@ async function main(): Promise<number> {
             const rawRejectReason =
               (values.reason as string | undefined) || positionals.slice(3).join(' ');
             const rejectReason = rawRejectReason.length > 0 ? rawRejectReason : undefined;
-            await workflowRejectCommand(rejectRunId, rejectReason, jsonFlag, effectiveCwd);
+            await workflowRejectCommand(
+              rejectRunId,
+              rejectReason,
+              jsonFlag,
+              effectiveCwd,
+              detachFlag
+            );
             break;
           }
 
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -5049,6 +5049,44 @@ describe('workflowApproveCommand / workflowRejectCommand / workflowResumeCommand
     expect(parsed).toMatchObject({ ok: true, detached: true, continues: true });
   });
 
+  it('threads a caller-supplied --cwd to the child instead of the parent process.cwd()', async () => {
+    const workflowDb = await import('@archon/core/db/workflows');
+    const paths = await import('@archon/paths');
+    (workflowDb.getWorkflowRun as ReturnType<typeof mock>).mockResolvedValueOnce({ ...pausedRun });
+    (paths.getArchonHome as ReturnType<typeof mock>).mockImplementationOnce(() => {
+      throw new Error('no home in test');
+    });
+    const spawnSpy = mockSpawn();
+    const savedArgv = process.argv;
+    process.argv = [
+      'bun',
+      '/abs/cli.ts',
+      'workflow',
+      'approve',
+      'run-123',
+      '--cwd',
+      '/caller/repo',
+      '--detach',
+    ];
+
+    let spawnCmd: string[] = [];
+    let spawnOptions: { cwd: string; cmd: string[] } | undefined;
+    try {
+      await workflowApproveCommand('run-123', undefined, undefined, '/caller/repo', true);
+      spawnOptions = spawnSpy.mock.calls[0]?.[0] as { cwd: string; cmd: string[] } | undefined;
+      spawnCmd = (spawnOptions?.cmd ?? []).slice();
+    } finally {
+      process.argv = savedArgv;
+      spawnSpy.mockRestore();
+    }
+
+    expect(spawnOptions?.cwd).toBe('/caller/repo');
+    // buildDetachedRunCmd appends --cwd LAST (parser is last-wins), so the
+    // appended value is what the child actually resolves.
+    const lastCwdIdx = spawnCmd.lastIndexOf('--cwd');
+    expect(spawnCmd[lastCwdIdx + 1]).toBe('/caller/repo');
+  });
+
   it('approve --detach refuses a non-paused run synchronously and spawns nothing', async () => {
     const workflowDb = await import('@archon/core/db/workflows');
     (workflowDb.getWorkflowRun as ReturnType<typeof mock>).mockResolvedValueOnce({
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -284,6 +284,51 @@ function firstJsonPayload(spy: ReturnType<typeof spyOn>): string {
   return ((spy.mock.calls[0]?.[0] as string) ?? '').trimEnd();
 }
 
+/**
+ * Stand-in for the child `spawnDetachedWorkflowRun` starts. `node:child_process`'s
+ * spawn routes through `Bun.spawn` under Bun, so spying on `Bun.spawn` intercepts it
+ * and Bun wraps this object into the ChildProcess whose events the startup window
+ * listens on. Shared by both detach suites — `run` and the control verbs.
+ */
+function createDetachedChildFixture(pid: number | null = 12345): {
+  child: ReturnType<typeof Bun.spawn>;
+  unref: ReturnType<typeof mock>;
+} {
+  const unref = mock(() => undefined);
+  const child = {
+    pid: pid ?? undefined,
+    unref,
+  };
+
+  return {
+    child: child as unknown as ReturnType<typeof Bun.spawn>,
+    unref,
+  };
+}
+
+/**
+ * Drive a detach command past `waitForDetachedStartup` (#2279). The command awaits a
+ * 500 ms window in which a dying child would reject, so a test that never advances
+ * the fake timer hangs — and one that never awaits the command silently skips the
+ * window altogether. Both suites go through here so neither can pass by accident.
+ */
+async function finishStartupWindow(
+  commandPromise: Promise<void>,
+  spawnSpy: ReturnType<typeof spyOn>,
+  expectedSpawnCount = 1
+): Promise<void> {
+  for (
+    let attempt = 0;
+    attempt < 20 && spawnSpy.mock.calls.length < expectedSpawnCount;
+    attempt++
+  ) {
+    await Promise.resolve();
+  }
+  expect(spawnSpy).toHaveBeenCalledTimes(expectedSpawnCount);
+  jest.advanceTimersByTime(500);
+  await commandPromise;
+}
+
 describe('workflowListCommand', () => {
   let consoleSpy: ReturnType<typeof spyOn>;
   let stdoutSpy: ReturnType<typeof spyOn>;
@@ -4524,39 +4569,6 @@ describe('workflowRunCommand — detach', () => {
   let consoleSpy: ReturnType<typeof spyOn>;
   let stdoutSpy: ReturnType<typeof spyOn>;
 
-  function createDetachedChildFixture(pid: number | null = 12345): {
-    child: ReturnType<typeof Bun.spawn>;
-    unref: ReturnType<typeof mock>;
-  } {
-    const unref = mock(() => undefined);
-    const child = {
-      pid: pid ?? undefined,
-      unref,
-    };
-
-    return {
-      child: child as unknown as ReturnType<typeof Bun.spawn>,
-      unref,
-    };
-  }
-
-  async function finishStartupWindow(
-    commandPromise: Promise<void>,
-    spawnSpy: ReturnType<typeof spyOn>,
-    expectedSpawnCount = 1
-  ): Promise<void> {
-    for (
-      let attempt = 0;
-      attempt < 20 && spawnSpy.mock.calls.length < expectedSpawnCount;
-      attempt++
-    ) {
-      await Promise.resolve();
-    }
-    expect(spawnSpy).toHaveBeenCalledTimes(expectedSpawnCount);
-    jest.advanceTimersByTime(500);
-    await commandPromise;
-  }
-
   beforeEach(() => {
     jest.useFakeTimers();
     consoleSpy = spyOn(console, 'log').mockImplementation(() => {});
@@ -4908,13 +4920,21 @@ describe('workflowRunCommand — detach', () => {
 
 describe('workflowApproveCommand / workflowRejectCommand / workflowResumeCommand — detach', () => {
   let consoleSpy: ReturnType<typeof spyOn>;
+  let warnSpy: ReturnType<typeof spyOn>;
+  let stdoutSpy: ReturnType<typeof spyOn>;
 
   beforeEach(() => {
+    jest.useFakeTimers();
     consoleSpy = spyOn(console, 'log').mockImplementation(() => {});
+    warnSpy = spyOn(console, 'warn').mockImplementation(() => {});
+    stdoutSpy = spyOnJsonStdout();
   });
 
   afterEach(() => {
+    jest.useRealTimers();
     consoleSpy.mockRestore();
+    warnSpy.mockRestore();
+    stdoutSpy.mockRestore();
   });
 
   // working_path is deliberately a distro-style path: the child must spawn with
@@ -4930,40 +4950,41 @@ describe('workflowApproveCommand / workflowRejectCommand / workflowResumeCommand
     metadata: { approval: { nodeId: 'gate', message: 'Approve?' } },
   };
 
-  // Mirrors the run-detach tests: pid is required (spawnDetachedWorkflowRun
-  // throws when child.pid === undefined), and node:child_process.spawn routes
-  // through Bun.spawn so spying on Bun.spawn intercepts it.
-  const mockSpawn = () =>
-    spyOn(Bun, 'spawn').mockReturnValue({
-      pid: 12345,
-      unref: mock(() => undefined),
-    } as unknown as ReturnType<typeof Bun.spawn>);
+  /** Suppress the log file so the test writes nothing to disk (logPath becomes null). */
+  const silenceLogFile = async (): Promise<void> => {
+    const paths = await import('@archon/paths');
+    (paths.getArchonHome as ReturnType<typeof mock>).mockImplementationOnce(() => {
+      throw new Error('no home in test');
+    });
+  };
 
   it('approve --detach spawns a detached child (minus --detach) and performs ZERO writes in the parent', async () => {
     const workflowDb = await import('@archon/core/db/workflows');
     const { executeWorkflow } = await import('@archon/workflows/executor');
-    const paths = await import('@archon/paths');
     (workflowDb.getWorkflowRun as ReturnType<typeof mock>).mockResolvedValueOnce({ ...pausedRun });
-    // Force the log-file path to fall back to 'ignore' so the test writes no files
-    (paths.getArchonHome as ReturnType<typeof mock>).mockImplementationOnce(() => {
-      throw new Error('no home in test');
-    });
+    await silenceLogFile();
 
     const updateBefore = (workflowDb.updateWorkflowRun as ReturnType<typeof mock>).mock.calls
       .length;
     const execBefore = (executeWorkflow as ReturnType<typeof mock>).mock.calls.length;
-    const spawnSpy = mockSpawn();
+    const child = createDetachedChildFixture();
+    const spawnSpy = spyOn(Bun, 'spawn').mockReturnValue(child.child);
     const savedArgv = process.argv;
     process.argv = ['bun', '/abs/cli.ts', 'workflow', 'approve', 'run-123', 'ship it', '--detach'];
 
-    let spawnCallCount = 0;
     let spawnCmd: string[] = [];
     let spawnOptions: { cwd: string; cmd: string[]; detached?: boolean } | undefined;
     try {
       // Signature: (runId, comment, json, cwd, detach) — detach is the 5th arg,
-      // layered after upstream's cwd (the orthogonal-collision the plan warns of).
-      await workflowApproveCommand('run-123', 'ship it', undefined, undefined, true);
-      spawnCallCount = spawnSpy.mock.calls.length;
+      // layered after upstream's cwd.
+      const commandPromise = workflowApproveCommand(
+        'run-123',
+        'ship it',
+        undefined,
+        undefined,
+        true
+      );
+      await finishStartupWindow(commandPromise, spawnSpy);
       spawnOptions = spawnSpy.mock.calls[0]?.[0] as
         | { cwd: string; cmd: string[]; detached?: boolean }
         | undefined;
@@ -4973,7 +4994,6 @@ describe('workflowApproveCommand / workflowRejectCommand / workflowResumeCommand
       spawnSpy.mockRestore();
     }
 
-    expect(spawnCallCount).toBe(1);
     expect(spawnOptions?.detached).toBe(true);
     expect(spawnCmd).not.toContain('--detach');
     expect(spawnCmd).toContain('approve');
@@ -4988,27 +5008,32 @@ describe('workflowApproveCommand / workflowRejectCommand / workflowResumeCommand
     );
     expect((executeWorkflow as ReturnType<typeof mock>).mock.calls.length).toBe(execBefore);
     expect(consoleSpy).toHaveBeenCalledWith("Started 'approve' for run run-123 in the background.");
+    // No log file could be opened, so the child's output is discarded — say so.
+    expect(warnSpy).toHaveBeenCalledWith(
+      'Warning: could not open a log file — child output will not be captured.'
+    );
   });
 
-  it('approve --detach --json emits a structured ack without approving', async () => {
+  it('approve --detach --json emits a structured ack on stdout without approving', async () => {
     const workflowDb = await import('@archon/core/db/workflows');
-    const paths = await import('@archon/paths');
     (workflowDb.getWorkflowRun as ReturnType<typeof mock>).mockResolvedValueOnce({ ...pausedRun });
-    (paths.getArchonHome as ReturnType<typeof mock>).mockImplementationOnce(() => {
-      throw new Error('no home in test');
-    });
-    const spawnSpy = mockSpawn();
+    await silenceLogFile();
+    const child = createDetachedChildFixture();
+    const spawnSpy = spyOn(Bun, 'spawn').mockReturnValue(child.child);
     const savedArgv = process.argv;
     process.argv = ['bun', '/abs/cli.ts', 'workflow', 'approve', 'run-123', '--detach', '--json'];
 
     try {
-      await workflowApproveCommand('run-123', undefined, true, undefined, true);
+      const commandPromise = workflowApproveCommand('run-123', undefined, true, undefined, true);
+      await finishStartupWindow(commandPromise, spawnSpy);
     } finally {
       process.argv = savedArgv;
       spawnSpy.mockRestore();
     }
 
-    const parsed = JSON.parse(consoleSpy.mock.calls[0][0] as string) as Record<string, unknown>;
+    // The ack goes through writeJsonLine (flushed stdout), not console.log — a
+    // console.log ack can be truncated on a pipe (#2384).
+    const parsed = JSON.parse(firstJsonPayload(stdoutSpy)) as Record<string, unknown>;
     expect(parsed).toMatchObject({
       ok: true,
       runId: 'run-123',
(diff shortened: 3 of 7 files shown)


Wirasm:
Good direction — unattended control verbs are squarely on the roadmap (`direction.md §always-on-automation`), and routing all mutation into the child while the parent stays read-only is the right shape. Five things before it lands.

**1. Rebase onto current `dev` — this one matters more than it looks.**

Your branch does not contain `8c7882e8` (merged from #2273 today; I checked with `git merge-base --is-ancestor`). That commit fixes a bug where `--detach` on a **compiled binary** re-invoked the Bun single-file-executable's virtual `argv[1]` as a command: the child died with `Unknown command: /$bunfs/root/archon` and the parent acked success. It had been broken since `--detach` was introduced.

So as this PR stands, `approve --detach` on any brew or curl install would report success and leave the run **paused forever**. Your test evidence covers dev mode, where the bug never appeared. The rebase is mechanical — no overlapping hunks — but please re-verify the three verbs against a compiled binary afterwards, not just `bun run cli`.

**2. Make the precheck a shared validator rather than a copy of one branch.**

The parent's read-only precheck replicates one of the four preconditions. The other three — `child_workflow`, missing approval context, already-resolved — currently ack `{ ok: true }` and then die in the child's log. That is precisely the failure this PR exists to prevent, occurring on the surface where nobody is watching.

Please extract something like `assertApprovable(run)` / `assertRejectable(run)` in `packages/core/src/operations/workflow-operations.ts` covering all four, and call it from both `approveWorkflow`/`rejectWorkflow` and the CLI precheck. Inlining the checks in the CLI instead isn't good enough — two copies of a four-branch gate will drift, and the drift will be silent in exactly this direction again.

**3. `--detach` currently re-introduces the auto-resume that `--json` deliberately withholds.**

`buildDetachedRunCmd` strips `--json` (`workflow.ts:247`), so the child takes the ordinary inline path and continues the run. But our documented contract is that `--json` on approve/reject/resume "records/validates the decision and returns a clean JSON line **without** the inline auto-resume (drive continuation separately)". So `--detach --json` silently does the opposite of `--json`.

That may well be the behavior you want — but it has to be stated, and callers need to be able to tell. Please:
- update `packages/docs-web/src/content/docs/reference/cli.md` (the `--detach` row at :216 and the `--json` control-verb note at :290) and the CLI section of `CLAUDE.md`;
- add a field to the ack — `continues: true` or similar — so an automation knows whether it still owns continuation;
- add a test asserting the child argv for the `--detach --json` case.

**4. Spawn with the command's `cwd`, not `process.cwd()`.**

`runDetachedControlCommand` should take the `cwd` already threaded from `cli.ts` and pass `cwd ?? process.cwd()`. As written a caller-supplied `--cwd` is discarded (last-wins at `workflow.ts:249-251`), which can strand the child outside a git repo *after* the parent has already acked success. Your stated rationale for avoiding `working_path` is unaffected — it's never a candidate here.

**5. Please fill in `.github/PULL_REQUEST_TEMPLATE.md`** and either link an issue or say none exists. CodeRabbit's template check is right about this one.

---

One thing worth knowing rather than fixing: the detach ack is gated only on the child having a PID, so any child that starts and immediately dies reports as a created run. That's pre-existing (#2279) and it's why #2248 hid for a whole release — but this PR widens that surface from `run` to the control verbs, so the shared validator in (2) is doing more work than it appears to.

buun-dev:
Thanks — this was a good catch on item 2 in particular; it was the real defect and the rest
followed from fixing it properly. Per item:

**1. Rebase onto current `dev` — done.**

The branch was 40 commits behind. Rebased clean, no conflicts. Two things came with it that
are worth naming explicitly:

- `8c7882e8` (#2273) — `buildDetachedRunCmd` now slices user args from `argv[2]` in **both**
  dev and binary modes. This PR was silently missing it. In a Bun single-file executable
  `argv[1]` is a virtual entry path (`/$bunfs/root/archon`, `B:/~BUN/root/archon.exe` on
  Windows), so the pre-rebase code fed that to the child as its first token and the child died
  with `Unknown command: …` — while the parent acked success. On any `brew`/`curl` install,
  `approve --detach` would have reported success and left the run paused forever.
- **A correction to the review's framing of the rebase as pure hygiene:** it also turns the
  currently-red `windows-latest` check green. Those two `downloadWebDist` failures are not
  this PR's doing — they're the upstream flake fixed by `3044829f` (#2310), whose commit
  message analyzes this PR's own CI run (`29801864393`).

**2. The read-only precheck replicating only one of `approveWorkflow`'s preconditions — fixed,
but not in the shape suggested.**

You're right about the defect: the parent checked only `status !== 'paused'`, so the other
three preconditions acked `{ ok: true }` and then threw unseen in the detached child's log —
the exact failure `--detach` exists to prevent, relocated to the surface nobody watches.

Fixed by extracting the gate into `@archon/core` and calling it from **both** the operation and
the CLI precheck, so there is one gate rather than two copies free to drift again.

**Correction on the shape:** the gates are not symmetric, so this is two validators, not one.

| Gate | `approve` | `reject` |
|---|---|---|
| `status !== 'paused'` | yes | yes |
| missing `approval.nodeId` | **yes** | **no** |
| `approval.type === 'child_workflow'` | yes | yes |
| `isGateResolved(approval)` | yes | yes |

`rejectWorkflow` deliberately has no `nodeId` requirement — it falls back to
`approval?.nodeId ?? 'unknown'` when writing its audit event, so a run whose approval metadata
is malformed is still legitimately rejectable. Collapsing the two into a single four-branch
gate would either break reject or over-permit approve. Hence `assertApprovable` (four
preconditions, returns the validated `ApprovalContext`) and `assertRejectable` (three, returns
`ApprovalContext | undefined`).

**`resume` needed no change at all.** Its precheck already calls `resumeWorkflow` directly, so
its single gate (`RESUMABLE_WORKFLOW_STATUSES`) was never duplicated in the first place. There
was nothing to unify there.

The extraction is behavior-preserving: error strings are copied verbatim, and the two
pre-existing `child_workflow` refusal tests for the operations pass unchanged — that's the
proof, rather than an assertion. Eight new core unit tests pin each gate's message
individually, and four new CLI tests assert each refusal both throws **and** spawns nothing.

**3. The `--detach --json` contract — documented, and made machine-readable.**

You were right that it was unstated. `buildDetachedRunCmd` strips `--json` from the child argv,
so the child takes the ordinary human path — which auto-resumes. But the documented contract
for bare `--json` on these verbs is that it records the decision *without* the inline
auto-resume so the caller drives continuation. So `--detach --json` did the opposite of
`--json`, with nothing saying so.

I kept the behavior rather than "fixing" it: withholding continuation under `--detach` would
spawn a child that only records a decision and stops, which is what plain `--json` already
does, minus the point of detaching. Instead it's now stated in the CLI reference and in
`CLAUDE.md`, and the ack carries `continues: true` so an automation can read a field instead of
inferring behavior. A test asserts the child argv carries neither `--detach` nor `--json`.

All three locations you named are covered: the `--detach` row (`:216`), a new *Detached control
verbs* section under `workflow approve`, and the `--json` control-verb note at `:290`. That last
one needed its own sentence rather than just the cross-reference — it sits in the `workflow
resume` section and asserts `--json` "does **not** re-execute inline", which `--detach` inverts;
a reader in that section would never have reached the approve-section prose.

`continues` is `true` for all three verbs under `--detach`, because the child always drives the
verb's full inline behavior. Predicting continuation per-gate in the parent would mean
replicating executor logic there — exactly the drift item 2 exists to remove.

**4. Passing the caller's `--cwd` to the spawned child — fixed; it was a real bug.**

`runDetachedControlCommand` called `spawnDetachedWorkflowRun(process.cwd(), …)`, discarding the
`cwd` `cli.ts` had already threaded in. And since `buildDetachedRunCmd` appends `--cwd <value>`
**last** and the parser is last-wins, the caller's `--cwd` was overridden in *both* places — the
spawn option and the child's argv. The child could land outside any git repo after the parent
had already acked success. `workflowRunCommand` had always passed its resolved `cwd`; only the
control verbs regressed. Now takes `cwd` as a parameter and passes `cwd ?? process.cwd()`.

This doesn't reopen the `working_path` question: a container run's `working_path` is a distro
path the host cannot spawn into, which is why the child re-resolves everything by run id. The
choice was only ever between the caller's `--cwd` and the parent's `process.cwd()`.

**5. PR template + compiled-binary verification — done.**

The PR body is rewritten from `.github/pull_request_template.md` with every section filled.

On the binary: agreed that `bun run cli` evidence can't cover this, since dev mode is exactly
where the SFE argv bug never appears. Verified against a compiled `win32-x64` binary
(`Build: binary`, commit `6af8eb2a`) driven against an isolated `ARCHON_HOME` and a throwaway
bash-nodes-around-an-approval-gate workflow:

- `approve`/`reject`/`resume --detach --json` all ack `continues: true`, and **no** child log
  contains an `Unknown command: /$bunfs/root/…` or `B:/~BUN/root/…` line.
- More importantly than the ack — a pre-fix binary acks success too — the runs actually
  advanced: approve → `completed`, reject → `cancelled`, resume → re-invoked, skipped the
  completed node, re-paused at the gate.
- The refusal path was verified on a gate this PR **newly** enforces in the parent: recording an
  approval with plain `--json`, then re-running `approve --detach --json`, returns
  `{"ok": false, "error": "… was already approved and is awaiting resume."}` with that run's
  child-log file byte-identical before and after — proving nothing was spawned. Pre-fix, that
  case acked `{ok: true}` and died in the log.

**One thing I did not fix, flagged rather than folded in:** the detach ack is gated only on the
child having a PID, so a child that starts and immediately dies still reports as a created run
(#2279). That's pre-existing on `workflow run` and I've left it out of scope. But this PR does
widen that surface from `run` to the control verbs, which is part of why the shared validator
matters more than it looks — it moves the four/three refusal cases from "child dies unseen" to
"parent refuses synchronously", shrinking what #2279 can hide.

**Also out of scope, deliberately:** `plan_lint`-style test-double asymmetry findings elsewhere
in the tree (`packages/server/src/routes/api*`, `packages/core/src/db/{workflows,isolation-environments}*`).
This branch's diff touches none of those files; fixing upstream's test-double hygiene would
balloon a narrow CLI change across `packages/server`. Happy to open a separate PR if you want it.

---

**Commits on the branch (post-rebase):**

- `refactor(core): extract assertApprovable/assertRejectable as the shared control-verb gate`
- `docs(cli): state that --detach --json continues the run, and signal it in the ack`
- `fix(cli): spawn the detached control-verb child with the command's cwd`
- `docs(cli): note the --detach inversion in the resume --json paragraph too`

Wirasm:
Taking this one over the line rather than sending it back — @buun-dev, all five asks from the July 27 review landed, and I verified each against the code rather than the reply. Summary of that audit, then what I pushed.

**The extraction is behaviour-preserving, and I checked it rather than taking the claim.** Diffing every error string out of `approveWorkflow`/`rejectWorkflow` against the two new validators leaves an empty set in both directions once you normalise `${runId}` to `${run.id}` — and those are the same value, because `getWorkflowRun` is `WHERE id = $1` and the CLI resolves prefixes before the call. The pre-existing `child_workflow` tests passing unchanged is real evidence, not an assertion.

**The correction on the shape was right.** I pushed back toward one four-branch gate; splitting into `assertApprovable`/`assertRejectable` is correct, because `rejectWorkflow` writes `approval?.nodeId ?? 'unknown'` and so must tolerate a malformed context that `approve` refuses. Collapsing them would have broken reject or over-permitted approve. Good catch.

**Item 3 answered better than asked.** Keeping the inversion and making it legible — `continues: true` plus prose in both doc locations — is the right call. Withholding continuation under `--detach` would spawn a child that only records a decision, which bare `--json` already does.

---

## What I pushed (`3d95927`)

The branch was 40+ commits behind again and `dev` had moved under it in one way that mattered. **#2279 landed since**, and it made `spawnDetachedWorkflowRun` async: it now waits out a 500 ms startup window and throws with the tail of the child's log when the child dies immediately. `runDetachedControlCommand` still called it synchronously.

Nothing caught that. TypeScript accepts assigning the promise, lint has no floating-promise complaint for an assignment, and the new tests used a bare `{ pid, unref }` spawn mock that never drove the window — so they passed for the wrong reason. At runtime it meant:

- `logPath` serialised as `{}` in the ack;
- the human path printed `Child output: [object Promise]`;
- a child that died at startup produced an **unhandled rejection after the parent had already printed `{ ok: true }`**.

That is the failure this PR exists to prevent, reopened on the surface nobody watches — the same shape as the item-2 defect, arriving from upstream rather than from the branch. Worth naming precisely because it is exactly what "widens the #2279 surface from `run` to the control verbs" cashes out to.

Fixes:

1. **`await` the spawn.** The mutation test is in the suite: remove the `await` and `approve --detach rejects an immediate non-zero child exit instead of acking success` goes red with `"logPath": {}` and `"ok": true` in the diff.
2. **Ack through `writeJsonLine`**, like every other `--json` ack — `console.log` gives no flush guarantee on a pipe (#2384). `printJsonWriteError` is awaited for the same reason.
3. **Warn when no log file could be opened**, matching `run --detach`.
4. **Refuse a detached `approve`/`resume` whose run has no recorded working path.** Both inline paths already refuse it and the precheck holds the run, so it costs nothing and converts one more "child dies unseen" case into a synchronous refusal with the identical message. This is the one addition beyond your four asks — easy to drop if you disagree, it is self-contained.
5. **Tests now go through `createDetachedChildFixture` + `finishStartupWindow`** (hoisted to module scope, shared with the `run` suite) so they cannot pass by skipping the startup window again.
6. **Docs.** `dev` collapsed `CLAUDE.md` to a pointer at `AGENTS.md`, so your note had nowhere to land after the rebase — moved beside the `--json` rule it inverts. Also took CodeRabbit's two open points: dropped the "four/three preconditions" count (an internal number that differs per verb and will drift) and documented that `logPath` is `null` when no log file could be opened.

`bun run validate` exits 0 on the pushed tree.

---

## On the standing caveat

Your flag about the ack being gated only on the child having a PID is now **partly closed by upstream** rather than still open: #2279's startup window means a child that dies in the first 500 ms surfaces as an error with its log tail, not a false success. Combined with the shared validators moving the refusal cases into the parent, what remains is only the child that survives 500 ms and then fails — a genuinely narrower window than when you wrote that note.

Thanks for the careful reply on July 31 — the point-by-point structure with corrections where the review was wrong is exactly the right way to answer a review. The `plan_lint` test-double findings you deliberately left out of scope were the right call too; open that separately if you want to pick it up.

Not merging this myself — flagging it as ready for a maintainer merge decision.

Wirasm:
Cross-checked this against everything filed around `--detach` and the run lifecycle before it lands, and rebased onto current `dev` (`3f78981f`). One real misalignment found and fixed; everything else agrees.

## The misalignment: #2645

**#2645 ("detach an exact workflow run resume") is open, and this PR implements its entire acceptance list.** Point by point:

- *"resume one known failed or paused run by full or unambiguous short ID without blocking the invoking terminal"* — the detach path calls `resolveRunIdArg` before the fork, so short ids resolve, then `resumeWorkflowOp` enforces failed-or-paused.
- *"reuses the recorded worktree and inputs exactly as foreground exact-ID resume does"* — the child re-invokes the same `workflow resume <id>` command, so it is literally the foreground path.
- *"returns enough information to track the same run; does not create or select a different run"* — the ack carries the resolved run id, workflow name and log path.

The part that would have landed out of step is the documentation. #2645 says plainly:

> Keep the workflow-name shortcut as a convenience for a human intentionally asking for "the latest failed run of this workflow in this checkout"; **do not use it as the recovery path for a known run.**

But `cli.md` still said, in the `workflow resume` section — to a reader who has just typed a run id:

> To actually drive a resumable run to completion, use the blocking form or `workflow run <name> --resume --detach`.

That pointer only existed because the exact-id form *couldn't* detach. This PR removes that constraint, so `8b11715e` repoints it at `workflow resume <run-id> --detach`, says what the name form is actually for (newest resumable run **of that workflow, in the current checkout**), and fixes the same steer in the `--json` code comment and in `AGENTS.md`.

**Recommend adding `Closes #2645` to the PR description before merging.** I have deliberately not edited the body or touched the issue — that is the maintainer's call, not mine.

## Everything else agrees

**#2279** (detach acks on PID alone) — closed COMPLETED, fixed in `dev` by `waitForDetachedStartup`. Worth stating clearly: **the un-awaited `spawnDetachedWorkflowRun` call this PR originally had would have regressed that closed issue on the new surface.** A control-verb child dying at startup would have acked `{ok:true}` exactly as #2279 describes. With the `await`, all four verbs now inherit the fix instead of one. The issue is correctly closed and needs no update.

**#2384** (JSON over a pipe truncates with exit 0 — bun EAGAIN) — closed COMPLETED. The original ack here used `console.log(JSON.stringify(...))`, which is the precise pattern #2384 fixed; it now goes through `writeJsonLine` like every other `--json` surface. Would have contradicted; now agrees.

**#1123** (CLI approve marked paused runs `failed` on process exit) — closed COMPLETED, and the interesting one, because this PR makes a parent exit immediately by design. Checked in code rather than assumed: the signal-cleanup guard installs at `workflow.ts:1892`, the `--detach` fast path returns at `:1264`, and `runDetachedControlCommand` never calls `workflowRunCommand` at all — so the detaching parent arms no handler and cannot touch the run. The child takes the ordinary inline path and gets the guard. The guard is also CAS-protected on `status='running'`, so it could not clobber a paused run even if it did fire. This PR advances #1123's intent: shell death can no longer interrupt the executor at all.

**#2097** (short run ids rejected by resume) — closed COMPLETED. `resolveRunIdArg` runs before the detach fork on all three verbs, and the ack reports the resolved full id, so a short id in and a full id out.

**#2248** (Windows Bun SFE argv) / **#2063** / **#1853** — the machinery this reuses unchanged, all merged.

**#2109** (short run-id prefixes across run-addressing commands) — merged, and it makes the cwd fix in this PR more load-bearing than its commit message claims. `buildDetachedRunCmd` appends `--cwd` last and the parser is last-wins, so spawning with `process.cwd()` overrode the caller's `--cwd` *in the child's argv too*. A short id resolved by the parent against `/caller/repo` would then be re-resolved by the child against the parent's directory — matching a different project's run, or none. Passing the caller's `--cwd` makes parent and child resolve the same id the same way. That is a correctness fix, not just "the child could land outside a git repo."

**#1743** (use the source checkout cwd for workflow discovery on resume/approve/reject) — checked, because it puts a second notion of cwd on these same three verbs. No interaction: `discoveryCwd` is computed at `workflow.ts:2865/:3051/:3178`, all inside the inline paths past the detach fork, and it derives from `run.codebase_id` → `codebase.default_cwd`, never from `--cwd`. The detaching parent never computes it; the child computes it exactly as a foreground invocation does.

**#2080** (verify the Windows detach fixes on real hardware) — agrees; `spawnDetachedWorkflowRun` with `detached: true` + `windowsHide: true` is reused untouched. One note for whoever picks #2080 up: its checklist covers `workflow run --detach` only, and this PR widens that surface to three more verbs, so the on-hardware smoketest should probably cover them too. @buun-dev's compiled `win32-x64` verification is partial evidence toward that already.

## Two things about this PR's own description

**The body is now stale in two places, and I have not edited it** — it is the contributor's, and the code is what I was asked to fix.

1. It still says the detach ack is "gated only on the child having a PID" and calls #2279 "**pre-existing and out of scope here**." That was true on 2026-07-21 and stopped being true when #2390 merged on 2026-08-03. The rebase pulled that fix in, and the `await` makes the control verbs inherit it — so the caveat now describes a hole that is closed.
2. It says "Closes # — **none.** No issue was filed for this feature." True when the PR opened; #2645 was filed 2026-08-19.

**The compiled-binary evidence predates the current code.** The Human Verification section cites a `win32-x64` binary self-reporting `Git commit: 6af8eb2a`. That SHA was rebased away — it is not in the current seven commits. So the Windows binary smoke covers the pre-#2390 shape of this branch, not what is on it now. The unit tests do cover the new behaviour (including a child dying inside the startup window), but if you want binary-level Windows evidence before merging, it needs re-running. This is the same gap #2080 is tracking.

## Related but deliberately not addressed

**#2325** (a dead run holds its worktree lock indefinitely) — untouched here, but worth flagging: more verbs can now background an executor, so there are more ways to reach a `running` row whose process died *after* the 500 ms startup window. Not a contradiction and not this PR's job. Separately, #2325's proposed timer-based auto-release sits in tension with `AGENTS.md`'s "No Autonomous Lifecycle Mutation Across Process Boundaries" — a design call for the maintainers, not something to settle inside a CLI change.

**#2008** (HTTP approve/reject cannot resume a CLI-launched paused run) — same verbs, different surface. This PR doesn't touch the server route, and it does widen the CLI/HTTP divergence #2008 describes, since the CLI now has `--detach` on all three verbs while HTTP still can't resume in place. That is evidence for #2008's framing rather than a conflict with it.

**#2233** (CLI runner exits 1 on SIGTERM even when its owned run completed) — open residue of #2191; `cleanup()` still ends in an unconditional `process.exit(1)`. The detached child inherits it. Notably #2233's author argues orchestrators should read `workflow get <id> --json` rather than the exit code — the same reasoning behind `continues: true` here. No conflict; if #2233 is ever resolved, this PR needs no change.

**#2454** (parent auto-resume after a child gate drops `userId`/`source`/`baseBranch`/`baseOverride`) — the issue itself notes the CLI's dispatch and resume paths pass all four. The detached child re-invokes the CLI, so it lands on the correct path rather than the lossy one. Agrees.

**#2180 / #2018 / #1216** — the run-lifecycle decisions this PR has to respect: one approval slot per run, decisions apply only to the gate presented, and ambiguous live work is surfaced rather than autonomously failed. Nothing here mutates another process's non-terminal state — the parent is read-only and the child owns exactly the run it was given. #1216 is where the `AGENTS.md` rule came from, and it is the reason I would not implement #2325's timer-based lock release inside this change even if it were in scope.

Ruled out as unrelated: #1350 (web auto-resume polish), #2290/#2293 (`/setproject none` — "detach" in the conversation-binding sense), #1945/#1974 (console scroll detach), and the engine/composition issues (#2646, #2614, #1960, #2512, #2648, #2479, #2439, #2554, #2280).

## Current state

Rebased onto `3f78981f`, `bun run validate` exits 0, all six checks green, `mergeStateStatus: CLEAN`. Still not merging — flagging as ready for a maintainer merge decision, with the `Closes #2645` line recommended first.

Issue #2206, opened 2026-07-21
Repo worktree container isolation


buun-dev:
# Outcome

Allow an explicitly selected repo run to execute inside a container while using its own git worktree as the workspace. The container provides the execution boundary and toolchain; the worktree is nested inside it as the checked-out workspace.

This is an accepted product gap. The implementation must ship as one whole isolation backend, not as unused contract groundwork.

## Current gap

Archon currently supports these combinations:

| Codebase | Workspace | Execution |
|---|---|---|
| Repository | Git worktree | Host process |
| Folder project | Existing folder or container overlay | Host process or container |

A repository cannot currently combine a git worktree with container execution. This prevents repo workflows from using a pinned container toolchain and keeps concurrent runs dependent on the host environment.

## Accepted contract

The implementation must follow `direction.md §isolation-layers`, `§isolation-never-inferred`, `§isolation-arrives-whole`, and `§same-path-workspace`.

### Explicit selection

- Never infer or silently upgrade isolation from workflow shape, host platform, concurrency, or another field.
- The existing explicit isolation/configuration boundary must select container execution for a repo run.
- Unsupported or incomplete configuration fails before a run starts and names the setting the operator must correct.

### One complete backend

Land the backend, typed contract, routing, persistence, and lifecycle together in one PR. Do not first add optional `ExecutionContext` fields or helpers with no production producer.

The backend owns:

- creating or adopting the run's git worktree;
- mounting that workspace into the container;
- producing the execution context consumed by providers and deterministic nodes;
- pause/resume and process restart behavior;
- finalization, cleanup, and failure recovery;
- preserving the worktree/branch outcome needed for commits and pull requests.

The repo backend must not inherit the folder-overlay apply/discard gate. A repo run's durable output is its worktree and branch.

### Path ownership

Prefer mounting the workspace at the same path inside and outside the container. First verify whether running Archon on Linux or inside WSL provides the supported Windows path without divergent addressing.

If an accepted deployment still requires different host and container paths, resolve them once at run setup:

- Model host-visible and node-visible paths as distinct typed values.
- Require every engine-owned path used by the run to resolve: workspace, artifacts, logs, state, docs, provider working directories, and any other path exposed to nodes.
- Throw before execution when a required node-visible path is missing.
- Use the resolved node-visible value consistently in prompt/script substitution, subprocess environment variables, provider arguments, and Docker arguments.
- Keep the host-visible value for Archon-side persistence, artifact serving, cleanup, and other host operations.

[...]

PR #2207, opened 2026-07-21
feat(providers): support containers whose mount != host cwd (workdir + pathMap on ExecutionContext)

buun-dev:
## Summary

- **Problem:** The container `ExecutionContext` assumes **container cwd == host cwd** — the folder-overlay backend mounts host-visible at the same absolute path, so the arg builders just pass the caller's cwd straight through as `-w` and forward path env values verbatim. That assumption blocks any container whose mount lives at a **different in-container path** than on the host (e.g. a worktree mounted at `/work`, a meta dir at `/archon-meta`).
- **What changed:** A small, generic extension to the container `ExecutionContext` — an optional in-container `workdir` and an ordered host→container `pathMap` — plus a pure `remapContainerPath()` helper, wired into both docker-exec arg builders.
- **Status:** **Inert until a producer sets `workdir`/`pathMap`.** Host runs and same-path (host-visible) container runs produce **byte-identical** args — zero behavior change. This is standalone groundwork; it needs no consumer to be correct.

## Relation to #2206

This is the generic engine kernel called out in the [repo-worktree container isolation proposal (#2206)](https://github.com/coleam00/Archon/issues/2206). That proposal (a git worktree executed *inside* a container, mount ≠ host cwd) is the motivating use case, but this change is deliberately decoupled from it: it generalizes the exec contract for **any** mount-relocated container backend and lands cleanly on its own. It carries no worktree/provider/WSL specifics.

## What it does

- **`ExecutionContext` (container variant):** adds optional `workdir?: string` (the in-container working directory) and `pathMap?: Array<{ hostPrefix; containerPrefix }>` (ordered host→container mount-prefix remaps).
- **`remapContainerPath(value, pathMap)`:** pure helper. For a value under a `hostPrefix`, rewrites it to the `containerPrefix`. It is:
  - **separator-normalized** — a Windows host prefix (`C:\Users\...\artifacts`) remaps as cleanly as a POSIX one;
  - **boundary-safe** — `/work` never swallows `/workspace` (prefix match respects path segment boundaries);
  - a **no-op** when no prefix matches or no pathMap is set.
- **Both arg builders** — `buildSubprocessDockerArgs` (bash/script nodes) and `buildDockerExecArgs` (Claude spawn) — now emit `-w (workdir ?? cwd)` and remap forwarded path env values (`ARTIFACTS_DIR`, `LOG_DIR`, `DOCS_DIR`) through the pathMap.

## Backward compatibility

Fully additive. `workdir`/`pathMap` are optional; when absent the builders emit exactly today's args (`-w <cwd>`, verbatim env). Verified by tests that assert identical output with no pathMap.

## Files

| File | Change |
|---|---|
| `packages/providers/src/types.ts` | `workdir`/`pathMap` on the container `ExecutionContext`; `remapContainerPath()` |
| `packages/providers/src/claude/container-spawn.ts` | `buildDockerExecArgs`: `-w (workdir ?? cwd)` + remap env |
| `packages/workflows/src/dag-executor.ts` | `buildSubprocessDockerArgs`: `-w (workdir ?? cwd)` + remap env |
| `+ 3 test files` | `exec-context.test.ts`, `container-spawn.test.ts`, `dag-executor.test.ts` |

## Testing

- `bun run type-check` — green across all 10 packages.
- `packages/providers` `exec-context.test.ts` + `container-spawn.test.ts` — 16 pass / 0 fail.
- `packages/workflows` `dag-executor.test.ts` — 420 pass / 0 fail.

New tests: `remapContainerPath` (mapped / unmapped / boundary `/work` vs `/workspace` / Windows-separator cases); both arg builders emitting `-w` + remapped env under a pathMap, and identical args without one.

## Provenance

Extracted from our fork (`buun-dev/Archon`), where it underpins a worktree-in-container backend on WSL2 + Docker. Re-based cleanly onto current `upstream/dev`; the commit message and this PR are de-scoped to the generic contract so it's useful independent of that backend.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container-based command execution by using the correct in-container working directory.
  * Fixed forwarded environment values so host paths are translated to their corresponding container paths.
  * Preserved non-path environment values and prevented partial or unsafe path matches.
  * Improved handling of Windows and POSIX path formats for consistent container execution.
* **Tests**
  * Added coverage for working-directory selection, environment mapping, path boundaries, and cross-platform path normalization.

### DIFF
--- a/packages/providers/src/claude/container-spawn.test.ts
+++ b/packages/providers/src/claude/container-spawn.test.ts
@@ -89,6 +89,34 @@ describe('buildDockerExecArgs', () => {
   });
 });
 
+describe('buildDockerExecArgs — container workdir + pathMap extension', () => {
+  // Our WSL sandbox mounts the worktree at /work, NOT at the host cwd, so the
+  // execContext carries an in-container workdir + a host→container path map.
+  const WORKTREE = '/home/bunny/archon/worktrees/marphob-page/s1';
+  const CTX_MAPPED = {
+    kind: 'container' as const,
+    containerId: 'cid-123',
+    workdir: '/work',
+    pathMap: [{ hostPrefix: WORKTREE, containerPrefix: '/work' }],
+  };
+
+  test('emits -w workdir (not the host cwd) when workdir is set', () => {
+    const args = buildDockerExecArgs(CTX_MAPPED, makeSpawnOptions({ cwd: WORKTREE }), PIDFILE);
+    const wIdx = args.indexOf('-w');
+    expect(args[wIdx + 1]).toBe('/work');
+  });
+
+  test('remaps a forwarded env path value under a host mount prefix', () => {
+    const args = buildDockerExecArgs(
+      CTX_MAPPED,
+      makeSpawnOptions({ cwd: WORKTREE, env: { DOCS_DIR: `${WORKTREE}/docs`, KEEP: 'plain' } }),
+      PIDFILE
+    );
+    expect(args).toContain('DOCS_DIR=/work/docs');
+    expect(args).toContain('KEEP=plain'); // non-path value untouched
+  });
+});
+
 describe('buildContainerSpawn — SpawnedProcess contract', () => {
   test('spawns docker exec and exposes the child stdio', () => {
     const spawner = recordingSpawner();
--- a/packages/providers/src/claude/container-spawn.ts
+++ b/packages/providers/src/claude/container-spawn.ts
@@ -24,7 +24,7 @@ import { spawn, type ChildProcess } from 'child_process';
 import { randomUUID } from 'crypto';
 import type { SpawnOptions, SpawnedProcess } from '@anthropic-ai/claude-agent-sdk';
 import type { ExecutionContext } from '../types';
-import { CONTAINER_ENV_DENYLIST } from '../types';
+import { CONTAINER_ENV_DENYLIST, remapContainerPath } from '../types';
 import { createLogger } from '@archon/paths';
 
 /**
@@ -77,10 +77,13 @@ export function buildDockerExecArgs(
 ): string[] {
   const args = ['exec', '-i'];
   if (execContext.execUser) args.push('-u', execContext.execUser);
-  if (options.cwd) args.push('-w', options.cwd);
+  // Prefer the in-container workdir (a container whose mount ≠ host cwd, e.g. our
+  // WSL sandbox at /work); fall back to the host cwd for the same-path model.
+  const workdir = execContext.workdir ?? options.cwd;
+  if (workdir) args.push('-w', workdir);
   for (const [key, value] of Object.entries(options.env)) {
     if (value === undefined || CONTAINER_ENV_DENYLIST.has(key)) continue;
-    args.push('-e', `${key}=${value}`);
+    args.push('-e', `${key}=${remapContainerPath(value, execContext.pathMap)}`);
   }
   args.push(
     execContext.containerId,
--- a/packages/providers/src/exec-context.test.ts
+++ b/packages/providers/src/exec-context.test.ts
@@ -1,0 +1,51 @@
+import { describe, test, expect } from 'bun:test';
+import { remapContainerPath, type ContainerPathMap } from './types';
+
+// The worktree lives in the WSL distro (POSIX path); the run meta dir is a host
+// (win32) path bind-mounted at /archon-meta. A single pathMap remaps both, so a
+// container whose mounts don't sit at the host cwd resolves every forwarded path.
+const MAP: ContainerPathMap = [
+  { hostPrefix: '/home/bunny/archon/worktrees/marphob-page/s1', containerPrefix: '/work' },
+  {
+    hostPrefix: 'C:\\Users\\Buun\\.archon\\workspaces\\buun-dev\\marphob-page',
+    containerPrefix: '/archon-meta',
+  },
+];
+
+describe('remapContainerPath', () => {
+  test('returns the value unchanged when no pathMap is given', () => {
+    expect(remapContainerPath('/work/foo', undefined)).toBe('/work/foo');
+  });
+
+  test('remaps an exact host worktree prefix to the container mount', () => {
+    expect(remapContainerPath('/home/bunny/archon/worktrees/marphob-page/s1', MAP)).toBe('/work');
+  });
+
+  test('remaps a path under the worktree prefix, preserving the suffix', () => {
+    expect(remapContainerPath('/home/bunny/archon/worktrees/marphob-page/s1/docs', MAP)).toBe(
+      '/work/docs'
+    );
+  });
+
+  test('remaps a win32 host meta prefix to /archon-meta, normalizing separators', () => {
+    expect(
+      remapContainerPath(
+        'C:\\Users\\Buun\\.archon\\workspaces\\buun-dev\\marphob-page\\artifacts\\runs\\r1',
+        MAP
+      )
+    ).toBe('/archon-meta/artifacts/runs/r1');
+  });
+
+  test('is boundary-safe — a partial prefix like /work does not swallow /workspace', () => {
+    const map: ContainerPathMap = [{ hostPrefix: '/work', containerPrefix: '/x' }];
+    expect(remapContainerPath('/workspace/foo', map)).toBe('/workspace/foo');
+  });
+
+  test('passes through a value outside every host prefix', () => {
+    expect(remapContainerPath('/etc/hosts', MAP)).toBe('/etc/hosts');
+  });
+
+  test('leaves an empty value untouched', () => {
+    expect(remapContainerPath('', MAP)).toBe('');
+  });
+});
--- a/packages/providers/src/types.ts
+++ b/packages/providers/src/types.ts
@@ -326,7 +326,56 @@ export type SystemPromptInput = string | string[] | SystemPromptPreset;
  */
 export type ExecutionContext =
   | { kind: 'host' }
-  | { kind: 'container'; containerId: string; execUser?: string };
+  | {
+      kind: 'container';
+      containerId: string;
+      execUser?: string;
+      /**
+       * In-container working directory. When set, `docker exec -w` uses THIS
+       * instead of the host cwd — for a container whose mount is NOT at the same
+       * absolute path as the host (our WSL sandbox mounts the worktree at /work).
+       * Absent → the container runs at the host cwd (upstream's same-path model).
+       */
+      workdir?: string;
+      /**
+       * Host→container mount-prefix remaps applied to forwarded path VALUES
+       * (env values like ARTIFACTS_DIR/LOG_DIR/DOCS_DIR). Generic: models a
+       * container whose bind-mounts don't sit at their host absolute paths.
+       */
+      pathMap?: ContainerPathMap;
+    };
+
+/**
+ * Ordered host→container path-prefix remaps for a container execution context.
+ * First matching entry wins. See {@link remapContainerPath}.
+ */
+export type ContainerPathMap = readonly {
+  readonly hostPrefix: string;
+  readonly containerPrefix: string;
+}[];
+
+/**
+ * Rewrite a host path value to its in-container location using a {@link ContainerPathMap}.
+ *
+ * A value equal to (or under, at a `/` boundary) a `hostPrefix` is re-anchored to
+ * the matching `containerPrefix`; everything else passes through unchanged. Paths
+ * are compared with separators normalized to `/`, so a win32 host prefix (the run
+ * meta dir) maps as cleanly as a POSIX one (the WSL worktree). Returns POSIX paths.
+ */
+export function remapContainerPath(value: string, pathMap?: ContainerPathMap): string {
+  if (!pathMap || !value) return value;
+  const norm = value.replace(/\\/g, '/');
+  for (const { hostPrefix, containerPrefix } of pathMap) {
+    const from = hostPrefix.replace(/\\/g, '/');
+    if (norm === from) return containerPrefix;
+    const boundary = from.endsWith('/') ? from : `${from}/`;
+    if (norm.startsWith(boundary)) {
+      const base = containerPrefix.replace(/\/+$/, '');
+      return `${base}/${norm.slice(boundary.length)}`;
+    }
+  }
+  return value;
+}
 
 /**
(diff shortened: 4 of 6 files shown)


Wirasm:
## Maintainer review

Thanks for splitting the generic kernel out from the backend, and for being straight in the description about it being inert. That framing is what let me review it quickly.

Verdict: I can't merge this yet, for two reasons. One is a direction call that is mine to make and I've flagged it (see the last section). The other is a design gap in the contract itself that I think you'll want to fix regardless of how the direction call goes, because the fork you're running it in has the same gap.

### What I verified

- Rebases cleanly onto current `dev` (`ea7b85db`). No textual conflict, including with the `types.ts` change that landed today.
- Your tests pass when run directly: 16/16 in `packages/providers`, and the two new `dag-executor` cases.
- `bun run validate` **exits 1** on current `dev`. `packages/providers/src/exec-context.test.ts` is not listed in that package's `scripts.test`, and #2603 (landed 2026-08-19, after you opened this) added a guard that fails when a test file isn't selected by its package script. Side effect worth knowing: right now your new test file never actually runs in CI. That one is a mechanical fix.
- No producer sets `workdir` or `pathMap` anywhere in the tree. `ContainerBackend` is the only thing that constructs a container `ExecutionContext`, and it mounts the overlay at the host's absolute path by construction (`entrypoint.sh` mounts at `$ARCHON_WORKSPACE_PATH`). So on `dev` today there is no reachable configuration where the same-path assumption is violated. The assumption is real, but it is deliberate and documented, not an oversight.

### The design gap: one variable, two values

Engine path variables reach a node through **two** channels, not one. This PR remaps the env bag and leaves the other channel alone.

`executor-shared.ts:563` substitutes `$ARTIFACTS_DIR` (and `$STATE_DIR`, `$DOCS_DIR`) textually into the node body before any docker argv is built. That substituted text is the host path. Then `buildSubprocessDockerArgs` remaps the `ARTIFACTS_DIR` env value to the container path. Same variable, same node, two different values, no error.

Repro against your branch, using your own fixture paths:

```
workdir (-w) : /work
script text  : echo hi > "C:\Users\Buun\.archon\workspaces\demo\artifacts\runs\r1/out.txt"
env bag      : /archon-meta/artifacts/runs/r1
AGREE?       : false
```

For a `bash:` node this surfaces as a confusing redirect failure naming a host path. For a `prompt:` node it is worse and quiet: the agent is told in prose to write to a host path, creates it inside the container, writes there, and the artifact is invisible to the host, to `GET /api/runs/:runId/artifacts`, and to every downstream node. The engine's own sidecar writer, which runs host-side, keeps using the host path, so half the artifact tree lands in each place.

Your own Definition of Done on #2206 lists "path remap verified for every forwarded path (env-var channel **and textual `$ARTIFACTS_DIR` substitution**)" as still open, so I don't think this is news to you. My concern is that landing the env-only half first freezes the remap at the wrong seam.

The seam I'd argue for: resolve the engine's path variables **once**, at run setup, when the execution context declares a mapping, so every channel (text, env, argv) already carries the container path and there is only ever one value. That needs the engine to keep the host path and the in-node path as two distinct, named things (the executor writes sidecars host-side and the artifact API reads host-side), which is a type change worth designing rather than inferring from a prefix table at the argv boundary. It also makes `remapContainerPath` unnecessary in the arg builders; only `-w workdir` stays.

### Fail-fast on a miss

`remapContainerPath` returns the value unchanged when nothing matches, and your suite asserts that as intended (`passes through a value outside every host prefix`). On a same-path container that is correct. On a mount-relocated container an unmapped engine-owned path is *guaranteed* wrong inside the container, and the failure mode is a write that looks like it succeeded. That is the pattern `AGENTS.md` "Fail Fast + Explicit Errors" exists to prevent.

The engine sets `ARTIFACTS_DIR` / `STATE_DIR` / `LOG_DIR` itself, so it knows exactly which keys must resolve. A miss on one of those should throw and name the key and the map, not fall through. Blanket prefix-matching over the whole env bag is the other half of the same problem: the bag also carries per-user credentials, and the function has no notion of which values are paths.

Two smaller ones while you're in there:

- Case-sensitive matching on a win32 `hostPrefix` (CodeRabbit flagged it). Given this exists specifically for Windows and WSL hosts, drive-letter and segment casing does vary between whoever produces the value and whoever produces the map, and a miss is silent per the point above.
- The exact-match branch returns `containerPrefix` verbatim while the prefix branch strips a trailing slash. Inconsistent, and a `hostPrefix` written with a trailing slash misses its own exact match.

### On the boundary and the inference questions, for the record

Both clean, and I want to say so explicitly since they're the two things I check hardest on anything touching isolation:

- **Package boundary.** `ExecutionContext` lives in `@archon/providers/types` precisely because `@archon/isolation` produces it and `@archon/workflows` consumes it and neither may import the other. Adding optional fields there is the right home. `remapContainerPath` as a runtime export from that subpath follows the existing `CONTAINER_ENV_DENYLIST` precedent, adds no dependency edge and pulls in no SDK. No violation.
- **Infer versus declare.** The pathMap is declared data supplied by a producer, and `workdir` absent means the same-path model. Nothing is guessed. This does not conflict with "the engine never infers isolation."

### What's actually blocking: the direction call on #2206

The honest position is that this PR can't be judged on its own merits, because it exists only to serve a proposal that hasn't had a maintainer verdict. Archon's YAGNI rule is explicit about not adding interface surface without a concrete accepted use case, and #2206 is a request for appetite, not an accepted case. Merging the kernel first would be us answering #2206 with a shrug.

So the sequencing is: #2206 gets a real answer first. I'm raising it with the maintainers rather than deciding it in a PR thread. It is a whole new isolation backend with its own security surface and a WSL-specific provisioning layer to generalize, which deserves a deliberate yes or no rather than arriving one inert PR at a time.

I'd rather not leave you guessing about the shape while that happens, hence the detail above.

### If the direction call comes back yes

What would make this mergeable:

1. Register `exec-context.test.ts` in `packages/providers/package.json` `scripts.test` so it runs and `validate` is green.
2. Decide the seam. My argument is for resolving path variables once at run setup with host and in-node paths distinctly typed, rather than remapping at the argv boundary. If you disagree, argue it on #2206 and cover every channel, including the textual substitution and anything host-pathed that the SDK puts into argv (`--add-dir`, `--mcp-config`).
3. Throw on an unresolved engine-owned path instead of passing it through.
4. Land it together with the producer, so the contract arrives with the thing that proves it works.

One thing I'd genuinely like from you on #2206, since you're running this in production and we aren't: what breaks first when the mapping is incomplete? That's the detail that would most improve whatever we build.

Wirasm:
## Needs fixes

The relocated-container contract is not ready to merge. Path translation happens after workflow variables have already been substituted, incomplete relocation states are valid and fail open, and no current producer exercises the new public surface. The new helper test is also absent from the current provider test inventory. The exact head passes its historical validation suite, but it conflicts with current `dev` in the workflow executor and its tests.

**4 blocking · 1 non-blocking**

**Validation:** Exact-head validation, build, and focused tests passed. Integration with `dev@e62d77ab` failed with content conflicts in `dag-executor.ts` and `dag-executor.test.ts`; GitHub reports `CONFLICTING` / `DIRTY`.

### Findings

|---|---|---|---|
| `R1` | Important | Textual workflow-variable substitution keeps host paths while the environment receives container paths. | OPEN |
| `R2` | Important | Independent optional relocation fields allow incomplete contexts, and unresolved engine paths silently pass through. | OPEN |
| `R3` | Important | The PR adds public relocation machinery without a producer, contrary to the accepted kernel-plus-backend boundary. | OPEN |
| `R4` | Important | The new `exec-context.test.ts` is missing from the providers package test inventory required by current `dev`. | OPEN |
| `R5` | Suggestion | Windows casing and trailing slashes can make otherwise valid mappings miss or return inconsistent paths. | OPEN |

<details>
<summary><code>R1</code> — One workflow path has two node-visible values</summary>

**Impact:** In a relocated container, bash/script redirects can target an unmounted host path, while prompt nodes can silently write artifacts outside the host-visible artifact tree. Archon's artifact API and downstream nodes then cannot see the output.

**Evidence:** `packages/workflows/src/executor-shared.ts:447` substitutes `$ARTIFACTS_DIR` and `$DOCS_DIR` into node text before dispatch. `packages/workflows/src/dag-executor.ts:2257` remaps only the environment bag. `packages/workflows/src/dag-executor.ts:2347` exercises both channels for bash nodes; prompt nodes take the textual path through `packages/workflows/src/dag-executor.ts:1241`. A reviewer reproduction produced a host path in script text alongside `ARTIFACTS_DIR=/archon-meta/artifacts` in the Docker environment.

**Required outcome:** Resolve host-visible and node-visible engine paths once at run setup. Use the node-visible values consistently for prompt/script substitution and environment delivery, while retaining the host-visible values for Archon-side persistence and artifact reads.

**Found by:** `code-reviewer`, `seam-analyzer`, `code-simplifier`, `type-design-analyzer`

**Disposition:** OPEN. The argv-boundary remapper cannot cover the earlier textual-substitution route.

</details>

<details>
<summary><code>R2</code> — Relocated addressing can be incomplete without an error</summary>

**Impact:** A typed caller can relocate the working directory while leaving engine-owned paths on the host. A node can then fail with a misleading path error or successfully write container-private output that Archon cannot retrieve.

**Evidence:** `packages/providers/src/types.ts:327` declares `workdir?` and `pathMap?` as independent optionals. `packages/providers/src/types.ts:365` returns unmatched values unchanged. Both consumers accept the state through `packages/providers/src/claude/container-spawn.ts:82` and `packages/workflows/src/dag-executor.ts:2255`. A reviewer directly constructed a context with `workdir: '/work'` and no map and obtained `docker exec -w /work -e ARTIFACTS_DIR=/host/meta/artifacts ...`.

**Required outcome:** Model same-path and relocated container addressing as distinct variants. A relocated context must require its working directory and fully resolved node-visible engine paths together, and run setup must throw when an engine-owned path cannot be resolved.

**Found by:** `code-reviewer`, `seam-analyzer`, `type-design-analyzer`

**Disposition:** OPEN. The public type currently permits the unsafe partial state and both consumers silently accept it.

</details>

<details>
<summary><code>R3</code> — The standalone kernel has no supported producer</summary>

**Impact:** Merging this head would permanently add unreachable API and distributed path-remapping policy without delivering repo-worktree container isolation or proving the contract end to end.

**Evidence:** `packages/providers/src/types.ts:327` adds `workdir`, `pathMap`, and the remapping abstraction. The production constructors at `packages/isolation/src/backends/container.ts:201` and `packages/isolation/src/backends/container.ts:430` still create only same-path contexts. The maintainer verdict on [#2206](https://github.com/coleam00/Archon/issues/2206#issuecomment-5367874710) accepts the product goal but requires the contract to land with its producer/backend, not as inert groundwork.

**Required outcome:** Land the accepted repo-container producer with a coherent end-to-end path contract, or omit the kernel. Do not merge unreachable relocation surface on its own.

**Found by:** `code-reviewer`, `code-simplifier`

**Disposition:** OPEN. The intended use case is accepted, but this PR does not deliver it.

</details>

<details>
<summary><code>R4</code> — The new helper test is outside the current test inventory</summary>

**Impact:** After the required rebase, current `dev`'s test-inventory guard rejects the package even though the new test passes when invoked directly.

**Evidence:** `packages/providers/src/exec-context.test.ts:1` adds seven tests, but the `test` script in `packages/providers/package.json` does not select that file. Current `dev` includes `scripts/test-inventory.test.ts`, which requires package test files to be listed. Direct execution passed 7/7, proving the problem is registration rather than the test content.

**Required outcome:** Add `exec-context.test.ts` to an appropriate explicit providers test batch when reconciling the branch with current `dev`.

**Found by:** `code-reviewer`

**Disposition:** OPEN. The exact historical head predates the inventory guard, so its own `bun run validate` passes; the current base makes the omission a merge-gate failure.

</details>

<details>
<summary><code>R5</code> — Path normalization misses valid Windows and trailing-slash variants</summary>

**Impact:** If the remapping helper remains, Windows path casing or a trailing slash in a prefix can silently prevent a valid mapping or produce inconsistent exact-versus-descendant output.

**Evidence:** `packages/providers/src/types.ts:367` normalizes separators but compares case-sensitively. Its exact branch returns `containerPrefix` verbatim, while the descendant branch strips trailing slashes, and a trailing-slash `hostPrefix` misses its own exact value.

**Required outcome:** Prefer removing the generic remapper under R1/R2. If it remains, normalize Windows comparisons and trailing slashes consistently and cover those cases with focused tests.

**Found by:** `code-reviewer`

**Disposition:** OPEN. This is non-blocking relative to the larger seam redesign.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | `R1`, `R2`, `R3`, `R4`, `R5` |
| seams | `R1`, `R2` |
| simplify | `R1`, `R3` |
| types | `R1`, `R2` |

#### Validation

|---|---|---|
| `bun run validate` | PASS | Exact reviewed head passed generated-file checks, type-check, lint, format, and the historical package test inventory. |
| `bun run build` | PASS | Web and docs builds completed; only existing non-fatal syntax-highlighting and bundle-size warnings were emitted. |
| `cd packages/providers && bun test src/exec-context.test.ts && bun test src/claude/container-spawn.test.ts` | PASS | 16 passed, 0 failed. |
| `git diff --check origin/dev...HEAD` | PASS | No whitespace errors in the PR diff. |
| `git merge-tree --write-tree origin/dev HEAD` | FAIL | Content conflicts in `packages/workflows/src/dag-executor.ts` and `packages/workflows/src/dag-executor.test.ts`. |

</details>

Wirasm:
Thank you for the careful proposal, implementation, rebases, and production evidence. We are closing this branch because the product direction is now settled in a different shape.

The goal remains accepted: Archon should support a repo worktree executing inside a container. The standalone kernel in this PR should not land because it has no production backend, and its path mapping occurs at multiple execution edges after some node-visible values have already been produced. That creates two values for one engine path and permits incomplete relocated contexts.

I am updating #2206 as the source for a future whole-feature implementation. The replacement must land backend, contract, and lifecycle together. It will prefer same-path workspace mounting; if divergent addressing is genuinely required, run setup must resolve host-visible and node-visible engine paths once through a typed fail-fast boundary, before prompt substitution, environment construction, or argv building.

The exact-head review and current direction audit are recorded here: https://github.com/coleam00/Archon/pull/2207#issuecomment-5398558620

Your work was useful in forcing the isolation and path-ownership decisions into the open. The closure is about the final architecture, not the value of the use case.

PR #2223, opened 2026-07-21
feat(workflows): per-child worktree isolation for workflow: sub-run nodes (slice-2 PR-A)

Wirasm:
## Summary

- **Problem:** A `workflow:` sub-run node (slice 1) always runs its child in the parent's shared checkout. That is correct for sequential composition, but structurally cannot express parallel patterns — N children writing into one checkout is the exact collision the worktree-in-use guard exists to prevent.
- **Why it matters:** Per-child isolation is the foundation the whole slice-2 orchestrator-worker engine stacks on. Dynamic fan-out (PR-C) and racing (PR-D) are impossible without it; independently, a single mutating child benefits from its own checkout today.
- **What changed:** A `workflow:` node may now declare `isolation: worktree`. The engine calls a caller-injected `ChildIsolationResolver` port once per child at spawn time to get a per-child worktree cwd + `archon/task-<parent>-child-0` branch (registered in `isolation list`); the parent's checkout is untouched.
- **What did NOT change (scope boundary):** No fan-out, no `with:`/`$INPUTS`, no racing, no console/docs beyond this (those are PR-B/C/D/E). `isolation: inherit` (and default) is byte-for-byte slice-1 behavior. No DB columns. `@archon/workflows` still imports zero `@archon/isolation` (verified).

### Before

```
Author writes:                       Engine:
  - id: sub                          child run C spawned in the PARENT's cwd (/repo)
    workflow: implement              → parallel children would collide in /repo
    input: "$plan.output"            → isolation: worktree REJECTED at load time
```

### After

```
Author writes:                       Engine:
  - id: sub                          resolver.resolve() → fresh worktree
    workflow: implement                /wt/<parent>-child-0 on branch
    input: "$plan.output"              *archon/task-<parent>-child-0*
    *isolation: worktree*            child C runs in ITS OWN checkout (parent /repo untouched)
                                     no resolver wired → node *fails fast* (no shared-checkout fallback)
```

### After

```
@archon/workflows                         @archon/core                 @archon/isolation
  child-isolation.ts [+]  <== port TYPE     child-isolation-       ===  WorktreeProvider
  executor.ts [~] resolveChildIsolation?      resolver.ts [+]           isolationDb
  dag-executor.ts [~] node.isolation        (createChildWorktreeResolver — the IMPL)
                                                  ^
  CLI workflow.ts [~] ─── inject ───────────────┤
  orchestrator(-agent).ts [~] ─── inject ───────┘
```

**Connection inventory**

| `@archon/workflows` | `@archon/isolation` | unchanged | still ZERO imports — the whole point of the port |
| `executor.ts` closure | `child-isolation.ts` port | **new** | resolver called per child at spawn |
| `@archon/core` factory | `@archon/isolation` `WorktreeProvider` | **new** | impl lives in core (already deps both packages) |
| CLI / orchestrator | `@archon/core` `createChildWorktreeResolver` | **new** | **three** construction sites share one factory; the resolver then reaches **five** `executeWorkflow` opts call sites |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `workflows` `core` `cli`
- Module: `workflows:executor`, `core:orchestrator`

## Change Metadata

- Change type: `feature`
- Primary scope: `multi` (workflows + core + cli)

## Linked Issue

- Related #2121 (Phase 2, slice-2 plan PR-A)
- Related #1764
- Depends on #2169 (slice 1, merged)
- Note: PR-C (fan-out) and PR-D (racing) stack on this branch.

```bash
bun run validate   # exit 0
```

- Evidence: `bun run validate` exits 0 — `check:pi-vendor-map OK`, `check:capability-matrix OK`, the other generated-bundle checks, type-check, lint (`--max-warnings 0`), format:check, and all per-package tests pass.
- New tests in `packages/workflows/src/subrun.test.ts` (real executor recursion, in-memory store): (1) `isolation: worktree` calls the resolver for the `sub` node and the child runs in the resolver's distinct cwd; (2) `isolation: worktree` with NO resolver fails the node fast with `requires an injected …` and creates no orphan child row; (3) `isolation: inherit` leaves the resolver untouched and shares the parent checkout. Flipped the loader's `worktree`-rejection test to an acceptance test.
- Skipped: none.

## Security Impact (required)

- File system access scope changed? `Yes (bounded)` — an isolated child now writes into its own git worktree under the standard Archon worktree tree instead of the parent's checkout. Uses the existing `WorktreeProvider`; no new path surface. Folder-project (non-git) codebases get no resolver and fail fast rather than escaping isolation.

## Compatibility / Migration

- Backward compatible? `Yes` — `isolation` defaults to `inherit`; every existing workflow is unchanged.
- Config/env changes? `No`
- Database migration needed? `No` — no new columns (the env row uses existing `isolation_environments`; the child's worktree is its `working_path`).

- Verified scenarios: the three executor-level e2e tests above (resolver-cwd, fail-fast, inherit) exercise the real `executeWorkflow → runChildWorkflow → child executeWorkflow` recursion end-to-end against real workflow files on disk.
- Edge cases checked: no resolver → fail-fast before child creation; resume reuses the child's recorded worktree (never re-created); `inherit` with a resolver present does not call it.
- What was not verified: a live CLI run creating a real git worktree on disk (the plan's Level-6 manual E2E) — deferred; the in-memory harness covers the engine contract, and the resolver mirrors the already-exercised top-level CLI worktree-create path.

- Affected subsystems: `workflow:` sub-run execution, CLI/orchestrator dispatch wiring, isolation env registration.
- Potential unintended effects: an isolated child leaves its worktree/branch on disk for the collector to merge (intentional — keep-on-failure). Standard `isolation cleanup` / `complete <branch>` hygiene applies.
- Guardrails: `workflow.child_worktree_created` log line per child; fail-fast on a missing resolver surfaces as a node failure with a clear message.

## Rollback Plan (required)

- Fast rollback: revert this commit — `isolation: worktree` returns to a load-time rejection; nothing else changes. Small, self-contained diff.
- Feature flags: none needed — the capability is opt-in per node (`isolation: worktree`).
- Observable failure symptoms: a `workflow:` node failing with `requires an injected child-isolation resolver`, or an unexpected worktree left in `isolation list`.

## Risks and Mitigations

- Risk: silent shared-checkout fallback when `isolation: worktree` but no resolver.
  - Mitigation: fail fast before the child row is created (tested) — never fall back; a shared parallel write is the exact collision worktree isolation prevents.
- Risk: `@archon/isolation → @archon/workflows` dep inversion when building the resolver.
  - Mitigation: the resolver IMPL lives in `@archon/core` (already deps both); the port TYPE lives in `@archon/workflows`. `@archon/workflows` keeps zero `@archon/isolation` imports (verified).
- Risk: folder-project (non-git) codebase requesting a worktree.
  - Mitigation: no resolver is built for `kind: 'folder'`, so the node fails fast with a clear message.

## Summary by CodeRabbit

* **New Features**
  * Workflow sub-runs can now run in isolated worktrees using `isolation: 'worktree'`.
  * Child workflows support isolated checkouts across fresh, resumed, and nested runs.
  * Existing `isolation: 'inherit'` behavior remains available.
  * Worktree isolation tools are now publicly available.

* **Bug Fixes**
  * Validation accepts worktree isolation for workflow nodes and rejects it elsewhere.
  * Resumed worktrees validate their paths and fail cleanly if unavailable.

* **Documentation**
  * Added guidance on worktree lifecycle, naming, cleanup, resuming, and CLI visibility.


Wirasm:
## PR Review Summary

Multi-agent review (code-reviewer + silent-failure-hunter) at head `99437885`, read-only worktree, diff vs `origin/dev`. One finding was raised independently by both agents (pruned-worktree resume) and is listed once.

### Critical Issues (0 found)

None. The load-bearing contract — **an isolated child never silently falls back to the shared checkout** — was verified real and enforced on every path (no-resolver → fail-fast before any row exists; resolver-throw → caught fail-fast; zero swallowed errors; zero empty catches in the diff).

### Important Issues (5 found)

| I1 | code-reviewer | `resolveChildIsolation` not spread into any of the three `childOpts` constructions → a NESTED child (grandchild) with `isolation: worktree` always fail-fasts "requires an injected resolver" even when the top-level caller wired one — and nesting to depth 5 is first-class. Same non-propagation gap exists for `container:` today, but this PR makes it load-bearing. No nested-isolation test exists | `executor.ts:654,660,677` |
| I2 | both | `resumeFailedChild` reuses `working_path` verbatim with no `existsSync` — a child worktree pruned by `isolation cleanup` between failure and resume surfaces as a deep `ENOENT`, not the clear "worktree may have been cleaned up" message the top-level CLI resume path gives (`workflow.ts:1113-1120` precedent) | `executor.ts:618-622` |
| I3 | silent-failure-hunter | Resolver failure surfaces raw `err.message` instead of routing through `classifyIsolationError()` — the documented repo pattern the sibling top-level path uses. Disk-full/permission errors reach users as git stderr instead of classified actionable messages | `executor.ts:625-641`, `child-isolation-resolver.ts` |
| I4 | silent-failure-hunter | Failure path has no structured Pino log — success logs `workflow.child_worktree_created`, but there is no paired `_failed` line (CLAUDE.md convention); ops grep shows every success, zero failures | `child-isolation-resolver.ts` + the executor catch |
| I5 | silent-failure-hunter | No test exercises the resolver actually THROWING (git error/disk/permissions) — the exact scenario this PR hardens. Traced-correct by inspection, but unguarded against regression | `subrun.test.ts` |

### Suggestions (3 found)

| S1 | silent-failure-hunter | `isolation:` is schema-accepted then silently dropped on every non-`workflow:` node type — no superRefine rejection (unlike `with:`/`retry:`). Harmless pre-PR; consequential now that `worktree` is real. Mirror the `with:` guard | `schemas/dag-node.ts` ~878-899 |
| S2 | both | `ChildIsolationRequest.codebaseId` documented as a "resolver guard" but never read by the implementation — add the assertion or fix the comment | `child-isolation.ts` / `child-isolation-resolver.ts` |
| S3 | code-reviewer | `ChildIsolationResult.envId`/`branchName` returned but discarded — child run metadata never records `isolation_env_id`/branch (container path does). Not load-bearing today; PR-E's console grouping will want it | `executor.ts` runChildWorkflow |

### Strengths

- Port boundary verified: zero `@archon/isolation` imports in `@archon/workflows`; resolver impl in `@archon/core` is a justified, DRY-er improvement over the plan's build-at-each-caller suggestion
- All four `executeWorkflow` call sites repo-wide got the injection consistently, gated `codebase.kind !== 'folder'`; approve/reject/resume funnel through and inherit it
- `isolation_environments` registration happens inside the same `resolve()` as physical creation — child worktrees are discoverable by `isolation list/cleanup` with no special-casing, and no-auto-destroy matches the plan's documented scope limit
- Path-lock concurrency verified by reading: distinct child cwds naturally don't collide — two isolated siblings genuinely parallelize with zero lock changes needed
- Idempotent child identifiers (`<parent8>-child-<n>`) + create-adoption + upsert semantics: a retried resolve can't duplicate rows
- All 3 new tests verified discriminating by pre-PR reasoning

### Verdict

NEEDS FIXES — close to ready. Nothing touches the architecture; the two threading/guard gaps (I1, I2) plus the error-visibility trio (I3-I5) are mechanical.

### Recommended Actions

1. Thread `resolveChildIsolation` through all three `childOpts` + nested-isolation test (I1)
2. `existsSync` guard on `resumeFailedChild.working_path` with the CLI-precedent message (I2)
3. `classifyIsolationError()` wrap + `workflow.child_worktree_create_failed` Pino line + resolver-throws test (I3-I5)
4. S1 superRefine guard; S2 either-way; S3 wire `isolation_env_id` into child metadata now (cheap, PR-E wants it)

Wirasm:
## Review fixes — all 5 Important + 3 Suggestions addressed

Pushed as `db2a1b2f`. `bun run validate` exits 0.

|----|-------------|--------------|
| **I1** | Fixed | `resolveChildIsolation` now spread into all three `childOpts` constructions in `runChildWorkflow`, so a nested grandchild `workflow:` node with `isolation: worktree` also isolates (nesting is first-class up to the depth cap) instead of fail-fasting. Added a one-line comment noting the sibling `container:` context has the same non-propagation gap (out of scope). New 2-level nested test: parent → child → grandchild, all `worktree` → three distinct cwds, resolver invoked twice. |
| **I2** | Fixed | `existsSync` guard on `resumeFailedChild.working_path` before reuse → `failOutcome` with the CLI-precedent message ("worktree may have been cleaned up — start a fresh run"). New test: resume with a deleted childCwd → clean message, asserts **no** `ENOENT`. |
| **I3** | Fixed | Resolver failures now routed through `classifyIsolationError()` — done **in the resolver** (`child-isolation-resolver.ts`, the layer allowed to import `@archon/isolation`), so `@archon/workflows` keeps zero isolation imports. The executor catch keeps the sub-run context prefix. |
| **I4** | Fixed | Paired `getLog().error({err, parentRunId, nodeId, childIndex}, 'workflow.child_worktree_create_failed')` on the resolver throw path, matching the existing `_created` info line. |
| **I5** | Fixed | New test: a fake resolver whose `resolve()` rejects → `result.success` false, `node_failed` error carries the message, **no orphan child row** (fail is before `createWorkflowRun`). |
| **S1** | Fixed | `superRefine` now rejects `isolation:` on non-`workflow:` node types ("only supported on workflow (sub-run) nodes"), mirroring the `with:` guard. Loader test added. |
| **S2** | Fixed (assertion, not comment) | Resolver asserts `req.codebaseId === config.codebaseId` and throws on mismatch — the `codebaseId` "resolver guard" doc is now true. |
| **S3** | Fixed | Child run metadata now records `isolation_env_id` + `branch_name` at spawn (mirrors the container path's `isolation_env_id`) for PR-E console grouping. Asserted in the nested test. |

**Also:** completed the four core `@archon/isolation` test mocks with `classifyIsolationError` (now imported transitively via the resolver) so no batch loads an incomplete mock.

**Validation:** `bun run validate` → exit 0 (generated-bundle checks, type-check, lint `--max-warnings 0`, format, all per-package tests). New tests: `subrun.test.ts` 15→**18** this round (nested-isolation, resolver-throws, pruned-worktree-resume; 12 pre-PR baseline), `loader.test.ts` +1 (S1 rejection). `@archon/workflows` still imports **zero** `@archon/isolation`.

Wirasm:
Reviewed against `dev`. Design intent first, then the fixes.

## Design: both isolation modes stay first-class

To be explicit, since the PR reads as "worktree replaces the slice-1 restriction": a sub-run must be able to run **either in its own worktree or in the parent's checkout**, and both are legitimate steady-state choices, not a migration path.

- `isolation: inherit` — the coordinator shape. Parent and children share one checkout; children read what the parent staged and write where the parent will read. Cheap, and correct when the children are cooperating on one change.
- `isolation: worktree` — the fan-out shape. Children touch the same files and must not see each other's writes.

So the author picks per node, and neither is the default-by-neglect. Keep both paths tested and documented as a choice; nothing in the fixes below should push toward making `worktree` implicit.

## Blocking

**1. The child worktree identifier omits `nodeId`, so isolated children of one parent collide.**

`child-isolation-resolver.ts:78` builds `identifier = ${parentRun.id.slice(0,8)}-child-${childIndex}`, but the engine never passes `childIndex` — `executor.ts:648-652` sends only `parentRun`/`nodeId`/`codebaseId`. So **every** isolated child of a given parent resolves to `archon/task-<parent8>-child-0` and the same path. `WorktreeProvider.create` then adopts the existing worktree (`worktree.ts:149-152` → `findExisting` → `worktree_adopted` at `:636`) and `isolationDb.create` upserts onto the same `(codebase_id, workflow_type, workflow_id)` row.

Two concrete failures:

- **Sequential** (`b depends_on [a]`, both `isolation: worktree`): child B runs inside child A's worktree, on A's branch. That is the silent shared-checkout fallback the PR's own Risks section says cannot happen.
- **Parallel** (same layer): both target one path; the loser is cancelled by the path-lock with a misleading "This worktree is in use" (`executor.ts:1214`). Worse — if A's directory is gone while its branch survives, B's create takes the `already exists` path and runs `git branch -f <A's branch> origin/<base>` (`worktree.ts:1187`), **discarding A's commits**.

CodeRabbit raised this on commit 1 ("include `req.nodeId` in the identifier"); the fix round `db2a1b2f` addressed the eight self-review findings and left this one. Fold `nodeId` into the identifier (slug-safe, mind the length cap) and add the missing sibling test: two isolated `workflow:` nodes in one parent must get two distinct cwds.

This one propagates — it is unfixed in #2224 and #2250 as well, since they build on the same identifier.

**2. `maybeResumeParentRun` drops the resolver.** At `executor.ts:903` the parent's auto-resume calls `executeWorkflow(..., { ...hydrated, codebaseId })` with no `resolveChildIsolation`, though the caller at `:1604` has it in scope. Reachable today: isolated nodes A then B; A's child pauses at a gate → parent pauses → reviewer approves → child completes → parent auto-resumes **without** the resolver → B fails with "requires an injected child-isolation resolver". Under #2224 a gate pause between fan-out children is the normal case, so this stops being an edge. Test: isolated node A gated, isolated node B downstream, resumed via the auto-resume hook.

**3. The resolver never ensures `configureIsolation` has run.** `child-isolation-resolver.ts:81` calls `getIsolationProvider()`, which falls back to a no-op repo-config loader when `configureIsolation` was never called (`packages/isolation/src/factory.ts:12-31`). In the CLI that call only happens inside the top-level-worktree branch (`packages/cli/src/commands/workflow.ts:1280`). So `archon workflow run parent --no-worktree` with an `isolation: worktree` child — the natural coordinator shape, and one of the two modes above — creates child worktrees with `worktree.path`, `baseBranch`, `remote` (#2234), `initSubmodules` and **`copyFiles`** all ignored. `copyFiles` is how `.env` reaches a worktree; a child missing it fails its build in a way that reads as a flaky agent. Either configure the loader in the resolver's factory the way `orchestrator.ts:73-81` does, or call `configureIsolation` unconditionally in the CLI.

**4. `CLAUDE.md:830` becomes false on merge.** It states `workflow:` supports "shared checkout only in slice 1 (`isolation: inherit`; `worktree` reserved/rejected)". Every agent run reads that file, so a stale YAML-surface claim there produces workflows written against the wrong contract. Update it in this PR, and flip the constitution's case-law row at `workflow-language-constitution.md:41`. Per our own rule a `feat(workflows)` PR touching the YAML surface should cite the constitution in its body — this one doesn't yet.

**5. Zero tests on `createChildWorktreeResolver`.** Minimum: identifier uniqueness across nodeIds (the regression guard for #1), the `codebaseId` mismatch throw (`:68`), and that a `provider.create` rejection surfaces as a `classifyIsolationError`-mapped message plus a `workflow.child_worktree_create_failed` log line.

## Merge ordering

#2223 → #2224 → #2250. #2224 declares this as its base and the resolver types don't exist on `dev`, so retarget and re-verify after this lands.

Wirasm:
**Docs are missing.** This adds YAML surface — `isolation: inherit | worktree` on `workflow:` nodes — and touches nothing under `packages/docs-web/`. Per CLAUDE.md, a YAML-surface feature needs documentation and a constitution citation.

Please document: that isolation is explicit opt-in (no `isolation:` = shared parent checkout, unchanged behavior), what a child worktree costs in lifecycle terms (it registers an `isolation_environments` row, so `isolation list/cleanup` and `complete <branch>` apply to it), and the failure mode when `worktree` is requested but no resolver is injected — folder-project codebase, or a surface that doesn't inject one. The fail-fast itself is right; it just needs to be findable by an author before they hit it.

Related: the implicit `isolation: worktree` default that #2224 adds on fan-out nodes needs to come out — see https://github.com/coleam00/Archon/pull/2224#issuecomment-5179333018. That doesn't block this PR.

Wirasm:
## Docs for the `isolation:` surface — `dd098b68`

This PR added YAML surface (`isolation: inherit | worktree` on `workflow:` nodes) and changed nothing under `packages/docs-web/`. That's now addressed, along with the sharper gap behind it: `mutates_checkout:` — the field the whole shared-checkout story depends on — appeared **nowhere** in the docs site despite shipping long ago.

Docs only. No engine behaviour changed here; the behavioural reshape belongs to #2224.

### What went where, and why

**`guides/authoring-workflows.md`** — the existing *"Composing a Governed Sub-Run with `workflow:`"* section was already the page an author lands on, so the new material extends it rather than starting a new page. The old "Shared checkout, gates, and resume" bullet list is split: the gate/failure/cost bullets stay as *"Gates, failure, and cost"*, and the checkout question gets its own treatment.

- **"Choosing the child's checkout with `isolation:`"** — the three values (omitted / `inherit` / `worktree`), that `isolation:` is rejected at load on any non-`workflow:` node, and the governing rule stated plainly: *the engine never infers this*. Then what a child worktree actually costs, which is the part an author can't discover without reading source:
  - it is cut from `origin/<baseBranch>` **in the canonical checkout**, so it does **not** see the parent's uncommitted edits or the parent's branch — everything has to arrive via `input:` / artifacts;
  - nothing merges it back; only the terminal output returns, as `$<nodeId>.output`;
  - it registers an isolation environment, so `isolation list` / `cleanup` / `complete <branch>` govern it, and it is **not** removed when the child finishes;
  - resume reuses the recorded path and never re-creates it — cleaning it up between failure and resume produces the `working path no longer exists` failure.
- **"When a worktree can't be created"** — the exact fail-fast message, the two real causes (folder project; no resolved codebase), and the point that `archon validate workflows` **cannot** catch either, because resolver availability is a property of the run, not of the file. Also that the parent being `--no-worktree` doesn't prevent an isolated child.
- **"Running sub-runs side by side"** — the path-exclusive lock, why ancestors/descendants are excluded but **siblings are not**, and `mutates_checkout: false` as the supported way for read-only children to coexist in one checkout. Carries the §2.5 warning explicitly: a lock-cancelled child is threaded through on resume rather than re-driven, so a parent that fails this way **re-fails identically forever** — the state has to be avoided, not recovered from. Ends with the three-way table (`mutates_checkout: false` if it reads / `isolation: worktree` if it writes / `depends_on` if it must not overlap).
- Supporting edits on the same page: `mutates_checkout` added to the workflow-level schema block (its first appearance in the docs site), `isolation` added to the `workflow` row of the node-fields table, and the stale *"`isolation: worktree` is reserved and rejected at load time"* bullet plus the matching non-goal removed.

**`book/isolation.md`** — a short note in the narrative chapter, because the first contact many people will have with this feature is an `archon/task-3f9a1c2b-child-0` branch appearing in `isolation list` that they didn't create. Says what it is, that `cleanup`/`complete` treat it like any other worktree, and to leave it alone while its run tree is still resumable.

**`reference/cli.md`** — one paragraph under `isolation list`, same content in reference register.

**`reference/workflow-language-constitution.md`** — the case-law table. The sub-run row said `worktree` isolation "remains deferred"; it now records slice 2 as opt-in per-child isolation (fan-out, racing, and `with:` still deferred — those are #2224/#2250's rows to change). Added one rejected row for **isolation inferred from an unrelated field**, since that decision governs this surface and the fan-out one behind it, and is exactly the kind of thing the case-law table exists to stop being relitigated per feature.

Per CLAUDE.md, citing the constitution for the surface this PR adds: `isolation:` passes the admissibility test as coordination — it selects which governed checkout a sub-run executes in, which the engine must see to create, track, and resume the environment. It is declarative data with fixed semantics, not evaluation, and no script node can express it because the engine owns environment creation. The `§implicit-behavior` budget is satisfied by there being no implicit behaviour at all: the default is "do nothing", and the only way to get a worktree is to ask for one.

### Replying to the review points

- **#2224 C2, "the isolation semantics after C1: explicit opt-in, and what happens when a worktree is requested without a resolver"** — done, both halves. The explicit-opt-in rule is stated as a rule and not just as a default, and the no-resolver case is documented with the verbatim error, both causes, and the note that validation can't pre-empt it.
- **#2224 addendum item 3, "`mutates_checkout` appears nowhere in `packages/docs-web/` … needs to be documented as part of the C2 docs work"** — done here rather than in #2224, since it is load-bearing for the shared-checkout case that exists on this PR already. It appears twice: in the workflow-level schema block, and in full in "Running sub-runs side by side" with the collision it solves.
- **CodeRabbit, "include `req.nodeId` in the identifier"** — still valid and still unfixed; see the defect below. Not fixed here (docs-only scope), and it wants a test, not just a string change.

### One defect found while reading for semantics — not fixed here

**Two `isolation: worktree` sub-run nodes in the same parent run silently share one worktree.**

`ChildIsolationRequest` carries `nodeId`, but the resolver uses it only for the worktree *description*; the identifier is `${parentRun.id.slice(0, 8)}-child-${childIndex}` (`packages/core/src/workflows/child-isolation-resolver.ts:78`). And the executor never passes `childIndex` at all on this branch (`packages/workflows/src/executor.ts:672-676`), so it is always `0`. Two isolated `workflow:` nodes under one parent therefore resolve to the identical branch `archon/task-<parent8>-child-0` and the identical path.

The second `create()` then hits `findExisting`, `verifyWorktreeOwnership` passes (same canonical repo), and it **adopts** the existing worktree — an INFO `worktree_adopted` line, no error. Consequences:

1. The two children share a checkout the author explicitly asked to isolate.
2. A second `isolation_environments` row is registered pointing at the same path, so `isolation list` shows two entries and `complete <branch>` on either removes the worktree out from under the other.
3. If the two nodes are in the same DAG layer, both children now hold the *same* `working_path`, so the path lock sees a sibling on it and one self-cancels — reproducing the #2180 failure the isolation was meant to prevent, in the one case where the author did everything right.

This is the §7 shape exactly: a key coarser than the thing it identifies, returning a plausible answer instead of an error. #2224 will pass a real `childIndex`, which fixes fan-out siblings but **not** two distinct nodes — `nodeId` still has to enter the identifier.

Two smaller things, both pre-existing and neither blocking:

- `--base` (#2203) sets the cut-from for the run it was passed to but is not threaded into `createChildWorktreeResolver`, so an isolated child is still cut from the codebase default. Documented as a limitation rather than reported as a bug, since which way it *should* go is a judgement call.
- `resolveChildIsolation` propagates into nested children (fixed in `fa1c6f32`), but the sibling `container:` context still doesn't — noted in the code comment there, flagging it here so it isn't lost.

`bun run validate` green from the worktree; `bun run build:docs` green.

Wirasm:
## Fixed: the child worktree identifier collision

Follow-up to the defect reported in the previous comment. This also closes CodeRabbit's inline finding on `child-isolation-resolver.ts`, though not the way it suggested — see below.

### The bug

`identifier` was `${parentRun.id.slice(0, 8)}-child-${childIndex}`. It omitted `nodeId`, and the executor never passes `childIndex` on this branch, so **every** isolated child of a given parent computed the same string. Two `workflow:` nodes with `isolation: worktree` in one parent run therefore resolved to the identical branch and the identical worktree path.

The failure is silent, which is what makes it worth blocking on. `WorktreeProvider.create()` finds the existing worktree, `verifyWorktreeOwnership` passes (same canonical repo), and it **adopts** — INFO `worktree_adopted`, no error, no warning. Downstream:

1. Two children share a checkout the author explicitly asked to isolate. Asking twice got you one.
2. A second `isolation_environments` row points at the same path, so `isolation list` shows two entries and `complete <branch>` on either pulls the worktree out from under the other.
3. Same-layer nodes end up with the same `working_path`, so the sibling path-lock cancels one — #2180 Defect A, reproduced in the one case where the author did everything right, and per §2.5 not recoverable by resume.

### The fix

`buildChildIdentifier(parentRunId, nodeId, childIndex)`, extracted as a named export so the interesting properties can be tested directly:

```
<parentRunId8>-<nodeSlug≤16>-<sha256(nodeId)[0..8]>-child-<n>
```

e.g. `archon/task-3f9a1c2b-refactor-module-6fd3f873-child-0`.

**Why not just interpolate `nodeId`**, which is what the review comment asked for: `WorktreeProvider.slugify()` truncates the identifier at 50 chars. A verbatim node id reintroduces the same silent adoption for any two node ids sharing a long prefix, and a long enough node id truncates the `-child-<n>` suffix away, which would collide fan-out siblings the moment #2224 starts passing a real `childIndex`. Same bug class, rarer, and therefore worse.

So the readable part is bounded at 16 chars and the **full** node id rides in a hash. Worst realistic case — a 69-char node id with a 4-digit fan-out index — comes out at 45 chars, inside the cap. That's asserted, not assumed.

I verified the naive version rather than arguing about it: implementing it as `${parent8}-${nodeId}-child-${i}` leaves **2 of the tests failing** (the shared-prefix case and the slug-cap case).

The identifier is also trimmed to be already slug-shaped. Truncating at 16 chars can land mid-separator (`abcdefghijklmno-p` → `abcdefghijklmno-`), leaving a `--` that the provider's own slugify collapses — which would make the stored `workflow_id` differ from the branch derived from it, so an env row stops being findable from its own branch name. Harmless for uniqueness, but it was holding by luck; now it holds by construction and there's a test over five node-id shapes.

### Tests — `child-isolation-resolver.test.ts` (new)

Seven unit tests on `buildChildIdentifier` plus one on the resolver itself: two nodes under one parent, long node ids sharing a 30-char prefix, fan-out siblings, distinct parent runs, the worst-case slug-cap bound, determinism (resume has to recompute the same branch), and the slug-shape invariant. The resolver test asserts two `resolve()` calls with different `nodeId`s reach `provider.create` with different identifiers — the regression test for the actual failure.

Reverting the fix fails **3**, including that resolver test; dropping just the separator trim fails the slug-shape one. The remaining tests hold in both directions by design — they pin properties the change must not break rather than the bug itself.

The file lives in `packages/core/src/workflows/`, so it joins the existing `bun test src/workflows/` batch — no `package.json` change. It mocks `@archon/isolation` and `../db/isolation-environments`, neither of which `store-adapter.test.ts` (the batch's only other file) touches, and the two run clean together (20 pass).

### Blast radius

Contained. Nothing reconstructs or parses this identifier: the `findActiveByWorkflow` callers pass either a user-supplied branch name (CLI `--branch`) or the top-level resolver's own `workflowId`, and console grouping reads `parent_run_id` from metadata rather than parsing the branch. The feature has never shipped, so there is no existing-branch migration concern. Reverting is one commit.

Docs updated to the new branch shape in the three places that named it (`guides/authoring-workflows.md`, `book/isolation.md`, `reference/cli.md`), plus the two stale docstrings on `ChildIsolationResult.branchName` and the resolver factory.

`bun run validate` green.

### Still open, deliberately not fixed here

**Adoption is silent by design, and that's the deeper edge.** `findExisting` returning an adopted worktree with only an INFO log is load-bearing for other paths (resume, restart), so I didn't touch it — the identifier fix removes the collision that made it dangerous here. But any future caller that computes a colliding identifier gets the same silent behaviour. If that seems worth hardening, it wants its own issue rather than a change buried in this PR.

**`childIndex` is still never passed by the executor** (`executor.ts:672-676`). Correct for this PR — one child per node — and #2224 supplies it. `buildChildIdentifier` already handles it and the fan-out sibling case is tested, so #2224 only has to pass the value.

Wirasm:
**Disposition 2026-08-04 — what changed, and what is left.**

Two commits landed on top of the review round:

- `dd098b68` — docs. The `isolation:` surface documented in `guides/authoring-workflows.md`, plus `mutates_checkout` (its first appearance anywhere on the docs site — it is the supported way to express read-only concurrent siblings, and it was undiscoverable), notes in `book/isolation.md` and `reference/cli.md` so an unexplained `archon/task-…-child-N` branch in `isolation list` is recognisable, and a constitution case-law row plus a new rejected row for isolation inferred from an unrelated field.
- `b8f1d82e` — a blocking defect found while writing those docs. The worktree identifier was `<parentRunId8>-child-<childIndex>`, with `nodeId` reaching only the cosmetic `description` and `childIndex` never passed by the executor. Two `isolation: worktree` nodes in one parent therefore computed the same identifier, and `WorktreeProvider.create()` **adopts** an existing worktree at the computed path with an INFO log and no error. Two explicitly-isolated children silently shared one checkout; same-layer, they landed on the same `working_path` and the sibling path lock cancelled one — #2180 Defect A, reproduced in the case where the author did everything right.

Note that interpolating `nodeId` verbatim, as the review round asked, does **not** close it: `WorktreeProvider.slugify()` truncates at 50 chars, so two long node ids sharing a prefix collide identically, and a long enough node id truncates `-child-<n>` off the end — which would collide fan-out siblings the moment #2224 passes a real `childIndex`. Same bug class, rarer, therefore worse. The naive version was implemented first to check rather than argue: 2 tests fail on it. Shipped shape is a bounded 16-char readable slug plus an 8-char hash of the full node id, worst case 45 chars, asserted.

**Documented design fact, not a defect:** a child worktree is cut from the canonical repo off the codebase default branch, not from the parent's worktree, so an isolated child cannot see the parent's uncommitted work. Confirmed as correct and intended — a planning node's output reaches children through `$ARTIFACTS_DIR`, which is external to every worktree. Code does not propagate; decisions do.

**Deliberately left, both correct for this PR:**
1. `childIndex` is still never passed by the executor. `buildChildIdentifier` already handles it and the fan-out sibling case is tested, so #2224 only has to pass the value — do not redesign the identifier there.
2. `findExisting` adopting with only an INFO log is a deeper edge, load-bearing for resume/restart. The identifier fix removes the collision that made it dangerous here, but any future caller computing a colliding identifier gets the same silence. Filed separately rather than widened into this PR.
3. `--base` (#2203) does not thread into the child resolver. Left as a documented limitation — the question underneath it (should children be cut from the parent's branch or from base?) is a design decision worth making when it bites, not speculatively.

**What is left: CI green, then merge.** No further changes requested. This PR blocks #2224 and #2250.

Wirasm:
## Follow-up: should the resolver adopt an existing worktree at all?

Raised in review after the identifier fix. Short answer: **yes, it should — but it must never do it quietly, and until now it did.** Reasoning, because the adoption path has other callers and this is a judgement rather than a bug fix.

### When adoption is reachable on this path

`WorktreeProvider.create()` returns an existing worktree at the computed path instead of failing. Resolution only happens on a **fresh** child spawn — the resume path reuses the child's recorded `working_path` and never re-resolves — so for a worktree to already be there, an earlier `resolve()` with the *same* `(parentRunId, nodeId, childIndex)` must have created it. Since the identifier is deterministic in exactly those three, that leaves two cases:

1. **A spawn that created the worktree and then died before its child run row was written** (a DB blip between `resolve()` and `createWorkflowRun`). The parent resumes, finds no child row for that node, spawns fresh, recomputes the identical identifier, and the worktree is sitting there. Re-using it is the correct outcome and the only thing that lets the resume complete.
2. **An identifier collision** — adopting a *sibling's* live checkout. That was the bug, and it is what the fix removed.

So after the fix, the only worktree adoption can reach is **this same child slot's own from an earlier attempt**. Never another child's. Adoption stopped being dangerous when the key stopped being ambiguous — which is also why rejecting it would be the wrong move: it would convert case 1 from "recovers" into "wedged permanently", and case 2 no longer exists.

That also means determinism is now load-bearing rather than incidental, so there's a test pinning it.

### What changed

Adoption is allowed and no longer silent:

- a **WARN** `workflow.child_worktree_adopted` carrying the parent run, node, child index, branch, and path — previously the only trace was the provider's own INFO `worktree_adopted`, which doesn't know it's looking at a sub-run child and carries no parent/node context;
- `adopted` is recorded on the `isolation_environments` row and on the existing `child_worktree_created` info line. The row is the durable half — a log line is long gone by the time anyone asks why two runs touched one checkout, and this is the record `isolation list` and any post-mortem actually read.

If this ever fires for a sibling's checkout, the identifier has regressed and these two are the evidence. No change to `IsolationRequest` or to `WorktreeProvider`, so the other callers of the adoption path are untouched — the resolver reads the `metadata.adopted` discriminant the provider already returns.

Two tests, both confirmed failing before the change: a re-used worktree records `adopted: true`, a fresh one records `false`.

### On `childIndex`

The requirement was that fan-out siblings get distinct checkouts, which needs the `childIndex` the resolver reads to actually arrive. Worth being precise about where that stands.

`buildChildIdentifier` varies on the index and there's a test for it, so the identifier side is done. But **there is no index to pass on this PR**: `childIndex` and `child_index` appear nowhere in `dag-executor.ts` or `executor.ts`, `RunChildWorkflowArgs` has no index field, and the re-entry lookup is explicitly one-child-per-node ("At most one child per node in slice 1"). `resolve()` correctly receives no index and defaults to 0, which is the true value when a node has exactly one child.

I've left it that way rather than hardcoding a `childIndex: 0` at the call site — it would be a literal asserting an index dimension that doesn't exist yet, and the resolver already defaults identically. #2224 introduces the dimension and only has to pass the real value; the parameter shape is unchanged and the node dimension is purely additive, so the two changes compose.

### Not touched

`fan_out` semantics and the implicit-isolation default in `schemas/dag-node.ts` are #2224's. Nothing under `.archon/`.

The base-branch behaviour I flagged earlier — an isolated child is cut from the canonical repo off the default branch, so it cannot see the parent's uncommitted work — has since been confirmed as **correct and intended**, not a defect: code doesn't propagate to children, decisions do, and `$ARTIFACTS_DIR` is external to every worktree, which is what makes fan-out-over-a-plan the natural shape. It's documented as an authoring fact in `guides/authoring-workflows.md`. `--base` not threading into children is deliberately deferred (the real question under it is stacked-from-parent vs cut-from-base, worth answering when it bites). Nothing here changes either.

`bun run validate` green.

Wirasm:
**Scoping decision 2026-08-05.**

> A sub-workflow inherits the environment from its parent. No option to separately isolate the child — unless it's easy to add as optional. Fan-outs should be *optional* per-child isolation, but not by default; and it's a later problem if it isn't easy to make consistent as part of this.

**That is already the shape here, and the optional half turns out to be cheap — so it stays.**

- **Inherit is the default and always was.** A `workflow:` node with no `isolation:` key shares the parent's checkout. Nothing changes for an author who doesn't opt in.
- **`isolation: worktree` is opt-in only.** The engine never infers it — that rule is now constitutional case law, and #2224 had its inferred fan-out default removed to match.
- **Fan-out per-child isolation is optional and not default**, as of #2224's reshape.

**The one thing that made "is it easy?" a live question is a blocking defect, and the fix is one parameter.** An adversarial probe found that `maybeResumeParentRun` calls `executeWorkflow` without threading `resolveChildIsolation` — so a parent that pauses at a child gate and auto-resumes **cannot spawn any subsequent isolated child**. It fails telling the author to remove their isolation, on a git repo, via the CLI, with the resolver injected.

That breaks exactly the shape #2224's own docs recommend: isolated implement → gate → isolated review.

But it is genuinely small: `maybeResumeParentRun` is called at `executor.ts:1637`, **inside** `executeWorkflow`, where `resolveChildIsolation` is already destructured at `:995`. One parameter threaded through one function, plus a regression test (gated isolated child → parent auto-resume → second isolated node).

So the optional isolation is easy to keep working, and cutting it would mean cutting this PR's entire content while leaving #2224 without the schema field it builds on.

**Required before merge**, unchanged from the probe:
1. Thread `resolveChildIsolation` through `maybeResumeParentRun` + regression test — **blocking**
2. Empty-slug guard (`###`, `日本語`, `🚀` are valid node ids today and re-open the `--` bug)
3. Replace the "resume never re-resolves" comment with the real argument — determinism plus the `isolationDb.create()` **upsert**, which is load-bearing and undocumented
4. Note that `--base` (#2203) does not reach children
5. Rebase — 31 commits behind `dev`

Wirasm:
**Disposition 2026-08-05 — the five required items are done. Rebased onto `dev` (`8704a65f`); `bun run validate` green.**

Four commits on top of the review round, one per item plus the rebase.

| # | Item | Commit |
|---|------|--------|
| 1 | Thread `resolveChildIsolation` through `maybeResumeParentRun` + regression test — **blocking** | `90ff7aad` |
| 2 | Empty-slug guard on the child worktree identifier | `bf5009b8` |
| 3 | Replace the adoption-safety argument with the real one | `d0b0f21e` |
| 4 | Record that `--base` does not reach children | `ef574fc0` |
| 5 | Rebase — was 31 commits behind, 5 of them touching `executor.ts`/`dag-executor.ts` | clean, no conflicts |

---

### 1. The blocking defect, and the shape chosen for threading

`maybeResumeParentRun` re-enters `executeWorkflow` when a sub-run child reaches a terminal state, building its options bag from the hydrated resume state alone. `resolveChildIsolation` was dropped there, so a parent that paused at a child's approval gate and auto-resumed could not spawn **any** subsequent isolated child — it failed telling the author to remove their isolation, on a git repo, via the CLI, with the resolver correctly injected at the top.

**The shape: a plain sixth parameter, not part of `ResumePayload`.** That union carries what was *recorded* about the prior run — a pre-created row, prior node outputs, prior token usage — all reconstructible from the database. A resolver is a live capability of the surface driving this process; there is no run row it could be rehydrated from, and putting it in the resume payload would mean every `hydrateResumableRun` caller has to know about isolation. It also isn't optional-by-omission the way the other fields are: dropping it is what caused this bug.

Forwarding the **child's** resolver is the correct value, not merely the available one. A child inherits the parent's `codebase_id` at creation, and the resolver is codebase-bound and rejects a mismatch loudly rather than creating a checkout in the wrong repo — so if the two ever diverge, that fails visibly instead of quietly branching somewhere else.

**Regression test** (`subrun.test.ts`) is the flagship shape end to end, driving the real recursion:

```
implement  (workflow:, isolation: worktree)  →  child pauses at its gate, parent pauses on it
    ↓ approve + resume the child in its own worktree (with a resolver, as the CLI does)
    ↓ child completes → parent auto-resume fires in-process
review     (workflow:, isolation: worktree)  →  gets its OWN worktree
```

The assertion is the resolver call sequence plus the parent completing with two distinct child checkouts — not "no error thrown", since this failure mode is success-shaped from the outside. **Confirmed failing before the fix**: resolver calls `['implement']`, expected `['implement', 'review']`.

### 2. Empty-slug guard

`buildChildIdentifier` trimmed a trailing separator off the truncated node slug but nothing handled the slug being empty outright. `id: z.string()` carries no pattern, so `###`, `___`, `日本語` and `🚀` are all node ids that load today, and each slugifies to nothing:

```
3f9a1c2b--56dc6d47-child-0
         ^^ collapsed by the provider's own slugify
```

which makes the stored `isolation_environments.workflow_id` stop being byte-identical to the branch derived from it — the invariant this function claims to hold, and the reason an env row is findable from its own branch name. Cosmetic in impact, but the slug-shape test claimed to cover it and didn't.

Segments are now joined with empties dropped. The hash of the full node id is unconditional, so those four stay distinct from each other: dropping the readable part costs legibility, not uniqueness. All four are added to the slug-shape test plus a dedicated uniqueness/no-`--` test. Two tests fail before the change.

### 3. The adoption-safety comment was wrong about *why*

It argued a sibling's checkout is unreachable because resolution only happens on a fresh spawn and resume never re-resolves. The conclusion holds; the reason doesn't. `resolve()` **is** reachable on resume: the parent's re-entry finds its child by `(parent_run_id, parent_node_id)`, and when that row was never written — or was deleted — the node takes the fresh-spawn path and resolves again.

The two properties that actually make adoption safe are now named, because both are load-bearing and one was invisible:

1. `buildChildIdentifier` is deterministic in `(parentRunId, nodeId, childIndex)`, so a re-spawn of the same slot recomputes the same path. Whatever is sitting there is that slot's own from an earlier attempt, never a sibling's.
2. **`isolationDb.create()` is an UPSERT** (`ON CONFLICT (codebase_id, workflow_type, workflow_id) WHERE status = 'active' DO UPDATE`, `packages/core/src/db/isolation-environments.ts:106`), so the re-spawn refreshes the existing env row instead of failing on the unique index. Nothing documented this, and "simplifying" it to a plain `INSERT` would break precisely the recovery the comment claims to protect.

Two matching overreaches in `executor.ts` ("never re-created", "never re-resolve on resume") are corrected in the same way.

### 4. `--base` does not reach children

`createChildWorktreeResolver` passes `baseBranch:` to `WorktreeProvider.create()` and never `baseOverride:`/`fromBranch:`, unlike the top-level CLI path. So a child worktree is cut using levels 2–4 of the base-branch precedence table only, and `archon workflow run parent --base release/2.0` still branches every isolated child off the repo's configured base. Now stated at the config field, in the authoring guide, and — importantly — in the precedence table itself, since a reader arriving *there* would otherwise assume level 1 applies everywhere.

Three corrections found while verifying the surrounding text against the code:

- The chain is `worktree.baseBranch` (repo config) > the codebase's stored default branch > git auto-detection. The guide named only the first two. Repo config outranking the codebase default holds for children too — the provider reads it from the canonical repo's own config, so it applies regardless of who asked.
- "does not see the parent's uncommitted edits" understated it. Cut from `origin/<base>` of the canonical repo, a child sees neither uncommitted **nor committed** parent work.
- "Resuming the parent never re-creates a child's worktree" is true only while the child's run row survives (see item 3). Corrected, with the note that it lands on the same branch name either way — which is what makes it safe.

---

### Behaviour contract, re-verified against the rebased tree

- **Inherit is the default.** A `workflow:` node with no `isolation:` key shares the parent's checkout. Test: `isolation: 'inherit' (and default) shares the parent's checkout — resolver untouched`.
- **`worktree` is opt-in and never inferred.** No field derives it. Constitutional case law, row added on this branch.
- **Every spawn path honours it.** Fresh spawn, nested grandchild, and — as of `90ff7aad` — post-auto-resume. All five `executeWorkflow` call sites outside the engine inject a resolver; the CLI's resume funnels through `workflowRunCommand`, so it is covered by the same injection.
- **A request that can't be honoured fails loudly.** No resolver → node fails with the actionable message, before any child row is created. No silent shared-checkout fallback.
- **Distinct things, distinct checkouts.** Per-`(parent, node, index)` identifier; two isolated nodes in one parent get two worktrees, asserted in the new test.
- **Governed like any other worktree.** `isolation_environments` row per child, so `isolation list` / `cleanup` / `complete` apply.

### Validation

`bun run validate` exit 0 — and the log grepped rather than trusted: **132 test invocations, all `0 fail`, zero `(fail)` markers, no non-zero exits**, plus `check:pi-vendor-map OK`, `check:capability-matrix OK`, the bundled-file guards, type-check, lint at `--max-warnings 0`, and format:check.

### Out of scope, reported not fixed

**`maybeResumeParentRun` drops more than the resolver, and this predates the PR** — it is the same on `dev` (`8704a65f`). The re-entry passes only `{...hydrated, codebaseId}`; `userId`, `source`, `baseBranch`, and `baseOverride` are all lost. Consequences, in descending severity:

1. **`userId` drives per-user credentials, GitHub token, and AI prefs.** Every parent node *after* a child gate therefore runs on install-level credentials and install-level model prefs rather than the run starter's. `parent.user_id` is right there on the row.
2. **`baseBranch`/`baseOverride` change value across the gate.** `$BASE_BRANCH` post-resume resolves repo-config → git auto-detect, skipping the codebase's stored default and discarding any `--base`. A `gh pr create --base $BASE_BRANCH` node placed after a child gate can target a different branch than the same node placed before it.

I did not fix these: each needs its own decision about where the value should come from on re-entry, and widening this PR into the general re-entry-fidelity problem is exactly the mixed mega-patch the repo's rollback-first rule warns about. Worth its own issue.

*(The `container:` context has the same non-propagation gap; that one is already noted in-tree and is guarded by a fail-loud check, so it degrades safely.)*

Wirasm:
---
pr: 2223
title: "feat(workflows): per-child worktree isolation for workflow: sub-run nodes (slice-2 PR-A)"
author: "Wirasm"
reviewed: 2026-08-05
---

# PR Review: #2223 — per-child worktree isolation for `workflow:` sub-run nodes

**Branch**: `feat/subrun-child-isolation` → `dev` (rebased onto `8704a65f`, HEAD `ef574fc0`)
**Files Changed**: 21 (+1499/−49)
**Mode**: 7-agent fan-out (code, simplify, errors, tests, comments, types, docs), findings re-verified

---

## Summary

Round 3. Every item from the earlier review round (I1–I5) and from the 2026-08-05 adversarial
probe (the five required items) was checked **against the tree, not against the commit messages**
— all held, including the arithmetic, the line-number citation, and the sha256 in a doc example.

**One new Important finding: M1.** A pre-existing config-gating gap that this PR converts from
dormant to live — on the `--no-worktree` parent path the docs themselves endorse, a child worktree
is created by a process that never configured the isolation provider, so the repo's **entire
`worktree:` config block** (`baseBranch`, `copyFiles`, `path`, `remote`) is silently ignored for
that child. Plus M2, two docs that now state the opposite of what ships.

Everything else is clean. No Critical findings.

---

## Validation — independently re-run, not taken from the PR body

| Check | Result | Evidence |
|-------|--------|----------|
| `bun run validate` | **exit 0** | run by me on the rebased tree |
| Tests | **6554 passed, 0 failed** | 132 `0 fail` markers, zero `(fail)` markers |
| Type-check | **PASS** | all 13 packages exit 0 |
| Lint | **PASS** | `--max-warnings 0` |
| Generated-file guards | **PASS** | `check:capability-matrix OK`, `check:pi-vendor-map OK` |

The author's claim of "132 test invocations, all `0 fail`" is exactly right.

---

## Issues Found

### Critical

None.

### Important

**M1 — A child worktree can be created by a process that never configured the isolation provider,
silently dropping the repo's whole `worktree:` config.**

`configureIsolation()` has exactly one CLI call site: `packages/cli/src/commands/workflow.ts:1409`,
**inside** the `else if (wantsIsolation && codebase)` branch. `wantsIsolation` is
`!options.resume && !options.noWorktree` (`workflow.ts:974-976`), so both `--no-worktree` and
`--resume` skip it — the `else if (options.noWorktree)` branch at `:1513` only logs. Meanwhile
`resolveChildIsolation` is built **unconditionally** at `workflow.ts:1774-1790`, regardless of
which branch fired.

`getIsolationProvider()` (`packages/isolation/src/factory.ts:28-31`) is a lazy process-wide
singleton over `configuredLoader`, which defaults to `() => Promise.resolve(null)`
(`factory.ts:11`). So on that path the child's `WorktreeProvider` gets the **no-op loader**, and
every field of `WorktreeCreateConfig` (`packages/isolation/src/types.ts:287-308`) is lost:

| Dropped | Consequence for the child worktree |
|---|---|
| `worktree.baseBranch` | cut from the wrong base — level 2 of the precedence table the new docs publish |
| `worktree.copyFiles` | files the repo config seeds into worktrees (e.g. `.env`) are **not copied** — the child's build/test steps fail confusingly |
| `worktree.path` | lands outside the configured repo-local worktree location |
| `worktree.remote` | git ops use the auto-detected remote, not the configured one |

`copyFiles` is the practical sting, not `baseBranch`.

**Why this is the PR's to fix even though the gating predates it:** before this PR,
`--no-worktree` meant no worktree was created anywhere in that process, so the no-op loader was
unreachable and harmless. This PR is what makes worktree creation reachable from a process that
skipped `configureIsolation`. It also affects the scenario the shipped docs use as their own
worked example ("a parent started with `--no-worktree` can still hand an isolated child its own
worktree"), so the docs are currently making a promise the code doesn't keep.

Second surface, same shape: `dispatchBackgroundWorkflow` builds the resolver at
`orchestrator.ts:340` while `ensureIsolationConfigured()` only runs via `getResolver()`
(`orchestrator.ts:85`). A fresh process whose first dispatch sets `workflow.worktree.enabled:
false` hits the identical gap.

**Fix:** call `configureIsolation()` unconditionally and early in `workflowRunCommand` (it is
idempotent and cheap — it just swaps the loader and resets the singleton), and mirror that in the
orchestrator dispatch path. That closes the gap rather than caveating it.

**M2 — Two docs now state the opposite of what ships; neither is in the diff.**

- `CLAUDE.md:845` — "shared checkout only in slice 1 (`isolation: inherit`; `worktree`
  reserved/rejected)". This PR is what accepts `worktree`. CLAUDE.md is the binding convention doc
  loaded into context by every agent working in this repo, so this doesn't just go stale — it
  actively steers the next implementer (and #2224, which builds on this exact field) toward
  believing the capability doesn't exist. Grepped the rest of CLAUDE.md; this is the only
  invalidated statement.
- `packages/docs-web/src/content/docs/book/quick-reference.md:161` — the `workflow:` field table
  still reads "Only `'inherit'` … `'worktree'` is reserved and rejected at load time (slice 2)".
  Type column should become `'inherit' | 'worktree'`.

### Suggestions

**S1 — The real resolver's `catch` block and codebase-mismatch guard have zero test coverage.**
`child-isolation-resolver.ts:141-146` (wrong-codebase guard) and `:239-251`
(`classifyIsolationError` + the paired `_failed` Pino line). Found independently by the tests and
silent-failure agents, both confirming via coverage output that exactly those ranges are uncovered.
Every "resolver throws" test uses a **hand-rolled fake** injected into `executeWorkflow`, so the
observable behavior is well tested but the resolver's own error-mapping wiring is asserted only by
reading the source — a regression that dropped the `classifyIsolationError` call or renamed the log
event would pass CI. One test with a rejecting `mockProviderCreate` closes both. *Criticality 6–7.*

**S2 — The three `createChildWorktreeResolver(...)` construction sites are past Rule-of-Three.**
`orchestrator.ts:340-347`, `orchestrator-agent.ts:714-721`, `cli/commands/workflow.ts:1779-1786` —
same `kind !== 'folder'` guard, same six-field object, near-verbatim copy-pasted comment. CLAUDE.md's
own bar is three occurrences. A `resolveChildIsolationForCodebase(...)` helper next to the factory
collapses each site to one line and puts the "folder projects never get a resolver" invariant in one
place. (This would also be the natural place to fix M1 once, instead of three times.)

**S3 — `childCwd = priorPath ?? cwd` is the one spot where "never a silent shared-checkout
fallback" isn't structurally enforced.** `executor.ts:648-658`. `working_path` is
`z.string().nullable()` (`schemas/workflow-run.ts:122`); a null would skip the `existsSync` guard and
silently run the child in the **parent's** checkout. Not reachable today (every child-row creation
writes a real path, `executor.ts:721`), so defense-in-depth. An explicit
`if (priorPath === null) throw` closes it for free.

**S4 — `RunChildWorkflowArgs.isolation` hand-duplicates the Zod enum** (`dag-executor.ts:369`)
instead of deriving `WorkflowNode['isolation']`. Cosmetic; TS still catches real drift.

**S5 — `database.md:96`** says a `workflow:` sub-run "shares its parent's checkout" (no longer
always true) and describes the path-lock as excluding only the ancestor chain (it excludes
descendants too, via `gatherDescendantRunIds`). Pre-existing on the second point; low priority.

**S6 — Concurrent `git worktree add` has no serialization or lock-contention retry**
(`packages/isolation/src/providers/worktree.ts:1160`). Same-layer isolated nodes spawn children
concurrently, so this PR makes concurrent worktree creation against one canonical repo reachable by
design rather than incidentally. Distinct branch names mean distinct ref locks, so likely fine — but
it becomes load-bearing at PR-C (fan-out), which is where it deserves a deliberate answer.

---

## Verified claims — all held

1. **Identifier uniqueness.** Pure function of `(parentRunId, nodeId, childIndex)`. Worst case
   computed as `8 + 16 + 8 + 5 + 4 + 4 separators = 45` against the provider's real
   `.substring(0, 50)` cap (`worktree.ts:1331-1337`). The four empty-slug ids (`###`, `___`,
   `日本語`, `🚀`) collapse to `''`, are filtered from the join, never produce `--`; the
   `workflow_id` ≡ branch invariant holds because `slugify(identifier)` is a no-op.
2. **The UPSERT is real and the line citation is correct.** `db/isolation-environments.ts:106` is
   exactly `ON CONFLICT (codebase_id, workflow_type, workflow_id) WHERE status = 'active' DO UPDATE`.
3. **The retracted "never re-resolve" prose is fully gone** — grepped the whole tree, zero hits
   including outside the diffed hunks. The replacement matches the actual branch logic.
4. **The blocking fix (`90ff7aad`) is real and the test is not shallow.** One call site
   (`executor.ts:1672`), resolver threaded as a 6th positional param. `subrun.test.ts:1592-1685`
   drives the real recursion and asserts the *call sequence* `['implement'] → ['implement','review']`
   plus two distinct checkouts — the right assertion, since this failure mode is success-shaped.
5. **All three `childOpts` constructions** carry the resolver (prior round's I1 closed); all five
   `executeWorkflow` call sites inject one, gated identically on `kind !== 'folder'`.
6. **`@archon/workflows` imports zero `@archon/isolation`** — grep plus its `package.json`.
7. **`ResumePayload` is structurally incapable of carrying a resolver** — closed two-variant type,
   three keys, no index signature. The boundary argued in prose is enforced at the type level.
8. **Claim 7 (deferred `userId`/`source`/`baseBranch`/`baseOverride` drop)** — pre-existing and
   **not worsened**; this PR adds one key to that object literal and removes nothing.
9. **The docs' worked example uses a real hash** — `sha256("refactor-module")[:8]` is literally
   `6fd3f873`.
10. **`mutates_checkout` and the path-lock prose are accurate**, including the subtle point that a
    *cancelled* child (path-lock loser) is never re-driven on resume — only `status === 'failed'`
    triggers `resumeFailedChild`.

---

## One agent finding rejected

The simplify agent proposed merging `buildChildIdentifier`'s two dash-trims into one applied after
slicing, reporting it empirically byte-identical across 17 cases. **It is not.** Brute-forced:
**1329 divergences in 86,430 inputs.**

```
"_deploy-to-production"  current="deploy-to-produc"  merged="deploy-to-produ"
"-abcdefghijklmnop"      current="abcdefghijklmnop"  merged="abcdefghijklmno"
```

The current version trims the leading separator **before** slicing, keeping a full 16 readable
chars; the merged version spends a slot on the leading dash and then trims it. Divergence needs a
leading separator *and* a body long enough for the 16-char cut to bite — the agent's corpus never
combined the two. Since `buildChildIdentifier` determinism is load-bearing for the adoption-safety
argument, that "equivalent" refactor would silently repoint every child branch name whose node id
starts with a separator. **Do not apply it.**

---

## Note on the PR body's "five inject sites"

Three agents independently counted **three** `createChildWorktreeResolver` construction sites and
**five** places `resolveChildIsolation` is passed into `executeWorkflow` opts. Both numbers are
right; the body's phrasing conflates them. Worth a word in the description so the next reader
doesn't hunt for two missing factories.

---

## Strengths

- **Claims were verified before being made.** Every load-bearing assertion in the disposition
  comments held under independent checking — the arithmetic, the line citation, the hash in a doc
  example. That is unusual, and it is why this review was cheap.
- **The `ResumePayload` boundary is a genuine type-level guarantee**, not a comment asking to be
  trusted. Best-designed decision in the diff.
- **The adoption-safety comment was rewritten because its *reasoning* was wrong while its
  conclusion was right** — and the replacement names the two properties that actually carry the
  argument, including an UPSERT nothing previously documented. That is the harder correction to make.
- **Fail-fast is honored end to end**: no resolver → node fails before any child row, worktree, or
  branch exists, with an actionable message and no shared-checkout fallback.
- **Test quality is high where it exists** — asserting the call *sequence* rather than the absence
  of an exception is the only assertion that catches this bug class.

---

## Recommendation

**REQUEST CHANGES — on M1 alone.**

M1 is a ~3-line idempotent fix (`configureIsolation()` early and unconditionally in
`workflowRunCommand`, mirrored in the orchestrator dispatch path) that makes the PR's own shipped
documentation true. Leaving it means a user following the guide's worked example gets a child
worktree missing whatever `worktree.copyFiles` seeds — a silent, confusing failure on a documented
path, and exactly the "never silently broaden / never silently degrade" line the repo's own
principles draw.

M2 (CLAUDE.md + quick-reference.md) should ride in the same commit — both one-liners.

S1 is the only other thing I'd want landed, and it is a follow-up issue rather than a blocker: the
behavior it guards is covered end-to-end today; what's missing is a regression pin on the wiring.
S2–S6 are cleanup and PR-C questions.

The rejected simplify suggestion must **not** be actioned.

---

*Reviewed by Claude — 7-agent fan-out; validate re-run and every finding independently re-verified*

Wirasm:
**Round-3 triage — `dffd220a`. Six findings fixed, one deferred, one declined with evidence, one confirmed-not-applied.**

Thanks for re-running validate independently rather than reading the commit messages. Triaged each finding rather than applying it; where I departed from your recommendation the argument is below.

| # | Category | Disposition |
|---|---|---|
| **M1** | fix now | Fixed — but at the resolver factory, not the two call sites |
| **M2a** CLAUDE.md | fix now | Fixed |
| **M2b** quick-reference | fix now | Fixed |
| **S1** coverage | fix now (**upgraded** from your follow-up) | Fixed |
| **S2** helper extraction | follow-up, drafted, unfiled | Deferred, with the motivation you cited now removed |
| **S3** null `priorPath` | fix now (**upgraded** from defense-in-depth) | Fixed |
| **S4** hand-duplicated enum | fix now | Fixed |
| **S5** database.md | fix now | Fixed, both clauses |
| **S6** worktree concurrency | **declined — with evidence** | Probed; does not reproduce |
| simplify refactor | not applied | Your counterexamples reproduced exactly |

---

## M1 — confirmed by running, and it is this PR's

I did not take the reachability claim on trust. Against a real temp repo with `worktree.copyFiles: ['.env']` and a real `git worktree add`:

```
PATH A (no configureIsolation): .env copied into child worktree = false
PATH B (configureIsolation):    .env copied into child worktree = true
```

No error on path A — the config is simply gone. That is the defect class, confirmed.

**And it is this PR's to fix, not pre-existing.** On `dev`, the CLI's only `getIsolationProvider()` call is `workflow.ts:1413`, inside the configured branch — so `--no-worktree` created no worktree anywhere in that process and the no-op loader was unreachable. This PR is what makes worktree creation reachable from a process that skipped `configureIsolation`. Your "dormant → live" framing holds.

**I fixed it in a different place than you proposed.** You suggested calling `configureIsolation()` early in `workflowRunCommand` and mirroring it in the orchestrator. I put it inside `createChildWorktreeResolver` instead:

- The resolver is the thing that *needs* a configured provider, so binding it there covers all three construction sites and any added later — your two-call-site fix would silently miss a fourth.
- It is a smaller diff than patching CLI and orchestrator separately.
- It is idempotent: `configureIsolation` swaps the loader and drops a lazily-rebuilt singleton, and all three loaders are byte-identical functions.

This also removes the motivation you gave for S2 ("the natural place to fix M1 once, instead of three times") — M1 is now fixed once regardless.

## S1 and S3 — upgraded from your categories

**S1** you rated a follow-up because the behaviour is covered end-to-end. I fixed it here instead: the catch block and mismatch guard are **this PR's own new code**, and shipping new error-mapping with no test is how the classify call gets silently dropped later. The isolation mock now returns a distinctive `classified:` prefix, so the test proves the call rather than the shape. ~20 lines.

**S3** you rated defense-in-depth. Also fixed here, because this PR is what gave the fallback its dangerous meaning: before it, every child shared the parent's checkout, so `?? cwd` was a no-op safety net. After it, that same line is a silent shared-checkout fallback for a child the author explicitly isolated — the one thing this PR promises never happens. Still unreachable, so the test constructs the null directly.

## S6 — declined, with evidence rather than assertion

You flagged concurrent `git worktree add` as likely-fine-but-unanswered, deserving a deliberate answer at PR-C. I probed it rather than filing a speculative issue: **12 concurrent `git worktree add` against one canonical repo with distinct branch names — 12 ok, 0 failed.**

Distinct branch names are guaranteed by `buildChildIdentifier`, so that is the shape the engine actually produces. Caveats stated plainly: one filesystem, one platform, N=12. If fan-out ever runs unbounded-parallel over a shared canonical repo the question returns, but there is nothing to fix today and a serialization layer now would be speculative generality.

## S2 — deferred, drafted, not filed

Three occurrences meets the Rule-of-Three *count*, but CLAUDE.md's bar is three **and stabilized**, and this surface is not: #2224 is stacked directly on it and may add or reshape construction sites. Extracting now means extracting against a moving target and re-touching three files on a branch another PR is rebased onto. With M1 fixed at the factory, what remains is ordinary duplication rather than a correctness multiplier.

Issue body drafted; not filed — opening on the tracker is the maintainer's call. Suggested trigger: revisit once #2224 lands, or immediately if a fourth construction site appears first.

## The simplify refactor — confirmed rejected

Reproduced your counterexamples exactly:

```
"_deploy-to-production"  current="deploy-to-produc"  merged="deploy-to-produ"   DIVERGES
"-abcdefghijklmnop"      current="abcdefghijklmnop"  merged="abcdefghijklmno"   DIVERGES
```

Not applied. Since `buildChildIdentifier` determinism is load-bearing for the adoption-safety argument, that refactor would silently repoint every child branch whose node id starts with a separator.

## Description fixed

Body now reads "**three** construction sites share one factory; the resolver then reaches **five** `executeWorkflow` opts call sites" — both numbers were right, the phrasing conflated them.

## Validation

`bun run validate` exit 0, log grepped rather than the wrapper trusted: 132 invocations all `0 fail`, zero `(fail)` markers, no non-zero exits. **The first run exited 1 on `format:check`** — caught precisely because the exit code was checked as well as the log, then fixed and re-run clean.

Tests: M1 and S3 confirmed failing before their fixes. S1's two are coverage pins on existing behaviour, so they pass either way by design — flagging that rather than implying they demonstrate a regression.

## Stack note

`dffd220a` moved #2224's base, which conflicted on the S4 one-liner against its `childIndex` field. Rebased #2224 (`7b19ea45 → 489473c7`), resolved by keeping both, re-validated green from that worktree. **Both PRs are `MERGEABLE` again.**

## One item left, deliberately not pushed

The resolver's module docstring still says "the five injection sites … share one implementation" — same conflation as the PR body, now inaccurate in the code. It is a comment-precision fix, not a correctness one, and a third push would move #2224's base a third time for no behavioural gain. Flagging it for whoever next touches that file rather than churning the stack.

Issue #2214, opened 2026-07-21
feat(workflows): with: named parameters on workflow: nodes — $INPUTS.<name> for sub-run children


Wirasm:
## Deferred from sub-runs slice 2 (maintainer decision, 2026-07-21)

Third deliberate deferral of `with:` — and the plan for it already exists: see the PR-B section of the slice-2 plan (`.claude/PRPs/plans/subruns-slice2.plan.md`, session artifact; design summarized here so this issue is self-contained).

## What

A data-only named-parameter map on `workflow:` nodes:

```yaml
- id: implement
  workflow: implement-block
  with:
    plan: $plan.output          # values: literals or $node.output[.field] refs — DATA only
    target_dir: packages/core
  depends_on: [plan]
```

Child-side: `$INPUTS.plan` / `$INPUTS.target_dir` substitution in the child's prompts (new variable surface in executor-shared), reconstituted on resume from child `metadata.inputs` (set at spawn). `input:` (whole-string → $ARGUMENTS) stays and remains the default; `with:` + `input:` together should be validated (either allowed with defined precedence, or rejected — decide at implementation).

## Constitutional bounds (settled — do not relitigate)

- Values are DATA (literals / output refs), never structure; dynamic workflow targets stay rejected (direction.md §load-time-composition)
- Slice-1/include precedent: `with:` currently REJECTS fail-fast on both include: and workflow: nodes — keep the rejection until this ships

## Why deferred (the demand test)

Fan-out (slice 2 PR-C) passes each work item as the child's $ARGUMENTS — sufficient for the headline swarm/racing cases. The workaround for multi-input children is passing a JSON object as `input:` and parsing child-side. **Ship this when that workaround shows up repeatedly in real workflows** — that's the demand signal, and the first real consumer should drive the ergonomics (naming, escaping, whether nested refs are allowed).

## Relationship to #1785

This is the CALLER half only. #1785's child-side parameter declarations (typed/validated inputs a workflow declares it accepts) is the fuller design — implementing this issue should at minimum not preclude it: keep $INPUTS flat key→string, no schema, so #1785 can later add declaration+validation on top without breaking existing callers.

## Acceptance sketch

- Parent passes two named params → child prompt substitutes both; absent key → fail loud (no-silent-drop, mirrors $node.output.field strictness)
- Resume: child resumed cold reconstitutes $INPUTS from metadata.inputs
- `with:` on a NON-workflow node stays rejected; include: keeps its own rejection until its variant is separately designed

[...]

Issue #1961, opened 2026-06-11
epic(workflows): enforce truthful run-wide work and spend budgets


Wirasm:
## Problem

Archon can limit one AI node with `maxBudgetUsd`, but it cannot yet bound the total work or spend of a workflow run. Parallel layers, loops, governed child launches, fan-out, and composed runtime-width work can schedule many independent units. Concurrency limits bound how quickly work starts; they do not bound how much eventually runs or costs.

A run budget must remain truthful across failure and cold resume. It must never silently truncate work, treat unknown provider cost as zero, or claim support for an execution shape whose accounting is incomplete.

This issue is the umbrella for total-consumption governance. It does not own provider concurrency scheduling.

## Approved delivery shape

The original implementation in PR #2802 proved the major seams but grew to 30 files, roughly 7,000 added lines, 26 commits, and a stale/conflicting base. Preserve it as research; rebuild the outcome on current `dev` in reviewable phases:

1. #2814 — budget foundation for ordinary nodes and cold resume.
2. #2815 — recursive accounting across loops, fan-out, and governed children.

OpenCode cost aggregation is intentionally gated on read-only reviews of community PRs #2760 and #2761. Do not start a competing OpenCode implementation until those reviews settle ownership and overlap.

## Standing contract

- A budget governs total work/spend; it is not a concurrency limit.
- The engine checks durable cumulative state before admitting new governed work.
- Exhaustion is an explicit refusal, never silent skipping or truncation.
- Already admitted parallel work may finish, so bounded overshoot is documented.
- Missing or invalid provider cost is not zero. Spend enforcement fails explicitly when it cannot be proven.
- Work units cover deterministic/provider work independently of monetary cost.
- Cold resume reconstructs the same authoritative totals and cannot replay or forge prior state.
- Separate child runs keep separate lifecycle/ownership; the parent consumes typed settled outcomes rather than sharing one distributed mutable ledger.
- Unsupported recursive or interactive shapes fail at load time until their dedicated phase lands.
- Existing node-level `maxBudgetUsd` remains compatible.

## Completion

- [ ] #2814 merged with reviewed exact-head validation and green CI.
- [ ] #2815 merged with reviewed exact-head validation and green CI.
- [ ] Documentation explains budgets, work units, unavailable cost, cold resume, child boundaries, and bounded overshoot.
- [ ] Provider concurrency and queueing remain owned by #2805.

## Related

[...]

Issue #977, opened 2026-04-08
DAG workflows: dynamic fan-out (spawn N nodes at runtime)


Wirasm:
> Migrated from dynamous-community/remote-coding-agent#771 — Archon active development has moved to coleam00/Archon. Original issue retained as historical reference.

---

## Summary

DAG workflow structure is static -- all nodes are defined at YAML load time. There is no way for a node's output to determine how many downstream nodes are spawned at runtime. This blocks common patterns like "analyze this PR, find N files with issues, spawn one fix-node per file."

## Problem

Today, if you want to run 5 parallel review agents, you must declare all 5 in the YAML:

```yaml
nodes:
  - id: code-review
    command: archon-code-review-agent
    depends_on: [scope]
  - id: error-handling
    command: archon-error-handling-agent
    depends_on: [scope]
  - id: test-coverage
    command: archon-test-coverage-agent
    depends_on: [scope]
  # ... hardcoded
```

You cannot:
- Dynamically spawn N nodes based on intermediate results (e.g., one node per modified file)
- Have a classification node decide which review agents to run
- Fan out based on a list produced by a bash or AI node

The `when:` condition system allows skipping pre-declared nodes, but this requires declaring all possible nodes upfront and creates bloated YAML for variable-width fan-out scenarios.

### Industry precedent

- **LangGraph**: `Send` API -- conditional edges return a list of `Send(node_name, state)` to spawn N parallel workers at runtime
- **Prefect**: `task.map(inputs)` -- creates one task per input element dynamically
- **Argo Workflows**: `withItems` / `withParam` -- dynamic parallelism from a parameter list

### Option A: `fan_out` node property

A node declares a `fan_out` field referencing an upstream node whose structured JSON output contains an array:

```yaml
nodes:
  - id: analyze
    prompt: "List all files modified in this PR that need review"
    output_format:
      type: object
      properties:
        files:
          type: array
          items:
            type: object
            properties:
              path: { type: string }
              reason: { type: string }

  - id: fix-file
    prompt: "Fix the issue in $fan_out.item.path: $fan_out.item.reason"
    depends_on: [analyze]
    fan_out:
      source: analyze          # upstream node with array output
      field: files             # JSON field containing the array
      item_var: fan_out.item   # variable name for each element

  - id: synthesize
    prompt: "Summarize all fixes"
    depends_on: [fix-file]     # waits for ALL fan-out instances
```

At execution time, after `analyze` completes:
[...]

PR #2224, opened 2026-07-21
feat(workflows): dynamic fan-out for workflow sub-run nodes (slice 2, PR-C)

Wirasm:
## Summary

- **Problem:** the slice-1 `workflow:` sub-run node is 1:1 and shares the parent's checkout. It structurally cannot express the orchestrator-worker pattern — a planner emitting a runtime-length list of tasks and N parallel isolated implementers executing them (#977) — because the width is fixed in YAML and N children writing in one checkout is exactly the collision the path lock prevents.
- **Why it matters:** the only workaround today is a `bash:` dispatcher spawning detached `archon workflow run --detach` children with hand-rolled polling — out-of-process, no native await, no cost roll-up, no run tree.
- **What changed:** adds `fan_out: { items, as?, max_parallel, join }` to a `workflow:` node. `items` is a `$node.output` ref to a JSON array (DATA — the child target stays a static name per §load-time-composition; only the item *count* is runtime). The engine expands one governed child run per item, bounded by a `max_parallel` sliding window, joined by `all_done` (default) or `all_success`, aggregated as a JSON array in item order threaded as `$node.output`. Cost/tokens roll up as Σ children. **Correction (#2469, fixed by #2608):** as merged, this was Σ of *completed* children — a child that burned tokens and then failed recorded no spend, inherited from the 1:1 path. The fix did not land in `failWorkflowRun` as predicted here: usage is now written at the DAG run tail, upstream of every disposition, so cancellation and gate pauses are covered too. Generalizes the slice-1 1:1 re-entry table to 1:N keyed by `metadata.child_index`, so parent resume re-drives only the incomplete instances. **Isolation is explicit-only and never inferred from `fan_out`** — concurrent children on a shared checkout are caught by a spawn-time preflight instead.
- **What did NOT change (scope boundary):** `first_success` racing (**rejected outright, not deferred** — a winner cancelling losers couples children the independence rule keeps independent; the enum value is retained only so existing YAML gets an explanatory error), `with:`/`$INPUTS` (deferred, #2214 — items ride the child's `$ARGUMENTS`; `as:` is **rejected at load** rather than accepted and ignored), console grouping + docs (PR-E), container-per-child isolation, and the #2180 gate queue. A non-fan-out `workflow:` node keeps full slice-1 gate support, untouched.

### Before

```
A workflow: node runs ONE child, in the PARENT'S checkout, width fixed in YAML.

  plan ─► [ workflow: implement  input: "$plan.output" ] ─► collect
                          │  (ONE child, shared /repo)
                          ▼
                 child C (cwd = /repo)
```

### After

```
plan ─► [ workflow: implement
           fan_out: { items: "$plan.output.tasks", max_parallel: 4, join: all_success } ] ─► collect
             │
    ┌────────┼────────┬────────┐      (one child per item; shared checkout unless the
    ▼        ▼        ▼        ▼
  C[0]     C[1]     C[2]     C[3]      parent_run_id=P, metadata.child_index=i
             │
             ▼
  $implement.output = ["C0.out","C1.out","C2.out","C3.out"]   (JSON array, item order)

  [all_done] failed/cancelled entries → { "error": …, "status": … } in the array
  [resume]   completed instances skipped; only failed child_index-es re-driven
  [#2180]    a fan-out child that pauses at a gate FAILS the node + is cancelled
```

### After

```
[~] schemas/dag-node.ts   — fan_out schema (+ FanOutConfig); join defaults to all_done
[~] loader.ts             — fan_out.items ref scan + upstream-dependency graph check
[~] include-expander.ts   — namespace fan_out.items refs in inlined blocks
[+] utils/map-with-limit.ts — pure bounded-concurrency sliding-window pool
[~] executor.ts           — runChildWorkflow stamps metadata.child_index
[~] dag-executor.ts       — executeFanOutWorkflowNode (expand/join/1:N resume/#2180)
```

**Connection inventory:**

| `dag-executor.ts` | `utils/map-with-limit.ts` | **new** | bounded child-spawn window |
| `dag-executor.ts:executeFanOutWorkflowNode` | `ctx.runChildWorkflow` | **modified** | now called per-item with `childIndex` |
| `dag-executor.ts` | `@archon/isolation` | unchanged | still ZERO imports (children isolate via the PR-A port) |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: L`
- Scope: `workflows`
- Module: `workflows:executor`, `workflows:dag-executor`, `workflows:loader`, `workflows:schemas`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Related #2121 #1764 #977 #2180
- Depends on #2223 (PR-A — per-child worktree isolation; this PR is stacked on `feat/subrun-child-isolation`)

```bash
                   # check:pi-vendor-map, check:capability-matrix, type-check,
                   # lint --max-warnings 0, format:check, all per-package tests
```

- Evidence provided: `bun run validate` exits 0. New tests: 9 fan-out e2e cases in `subrun.test.ts` (27 pass), 7 loader cases (177 pass), 1 include-expander case (35 pass), 7 `map-with-limit` unit cases (7 pass). `grep -rn "@archon/isolation" packages/workflows/src/` → still none.
- Skipped commands: none.

## Security Impact (required)

- New permissions/capabilities? **No** (fan-out spawns child runs through the existing governed lifecycle)
- File system access scope changed? **No** (a child gets a worktree only on an explicit `isolation: worktree`, via the existing PR-A resolver; `max_parallel` bounds concurrency to prevent a runaway N-wide layer)

## Compatibility / Migration

- Backward compatible? **Yes** (`fan_out` is a new optional field; existing `workflow:` nodes are unchanged)
- Config/env changes? **No**
- Database migration needed? **No** (`child_index` rides existing `metadata` JSONB; `parent_run_id` already exists)

- Verified scenarios: N-item all_success (ordered aggregate + item→`$ARGUMENTS` + per-child worktrees), empty array → `[]`, malformed items → fail-closed, all_success runs every child to terminal and then fails (no fail-fast), all_done partial (failed entry represented), `max_parallel` window (concurrency bounded to 2 over 5 items), cost roll-up (Σ = 0.03 over 3 children), 1:N parent resume (only the failed instance re-driven), paused fan-out child → node fails (#2180) + child cancelled.
- Edge cases checked: non-array items resolution, `first_success` rejected at load, `fan_out` on a non-workflow node rejected, `fan_out` in a loop_group body rejected, dangling + non-dependency `items` refs rejected, `max_parallel: 0` rejected.
- What was not verified: a live end-to-end run against a real git repo with real per-child worktrees (deferred to the plan's manual E2E; the executor recursion is exercised with a real in-memory store + a fake per-index resolver).

- Affected subsystems/workflows: `@archon/workflows` sub-run execution only. A non-fan-out `workflow:` node takes the untouched slice-1 path (branch on `node.fan_out` at the top of `executeWorkflowNode`).
- Potential unintended effects: children run in the parent's checkout unless the author writes `isolation: worktree`. Concurrent children over a shared checkout would collide on the path lock, so a spawn-time preflight refuses that expansion **before any child is spawned**, naming the three ways out (`mutates_checkout: false` on the child, `isolation: worktree` on the node, or `max_parallel: 1`). Worst-case spend is now `items.length` rather than "until the first failure", because no child's outcome ends another's.
- Guardrails: `max_parallel` (default 5) bounds concurrency (not total count or spend — see #1961); the spawn-time shared-checkout preflight fails closed, including when the child target cannot be resolved; the run-in-progress path lock remains the backstop.

## Rollback Plan (required)

- Fast rollback: revert this PR's commits — `fan_out` is additive; no schema/data migration to unwind.
- Feature flags: none (a workflow simply omits `fan_out`).
- Observable failure symptoms: a `fan_out` node failing with a resolver error (no child-isolation resolver wired) or `not a JSON array` (bad `items`).

## Risks and Mitigations

- Risk: `inherit` fan-out with concurrency collides on the path lock.
  - Mitigation: the spawn-time preflight refuses a colliding shared-checkout expansion before any child exists; isolation stays the author's explicit declaration, which is what the constitution's independence rule requires.
- Risk: a runaway N-wide layer (#1961 footgun).
  - Mitigation: `max_parallel` default 5, backed by a dedicated `mapWithLimit` — the unbounded top-level `Promise.allSettled` is not reused.
- Risk: resume mis-keys instances (started_at ≠ items order under concurrency).
  - Mitigation: explicit `metadata.child_index`; resume indexes by it, not by `started_at`.

## Summary by CodeRabbit

* **New Features**
  * Added dynamic fan-out for workflow nodes, allowing JSON arrays to create governed child runs.
  * Supports configurable concurrency limits, ordered aggregation, completion policies, isolated execution, cancellation, and resumable runs.
  * Added validation for invalid references, unsupported configurations, unsafe shared-checkout concurrency, and disallowed nesting.
* **Documentation**
  * Documented fan-out configuration, defaults, lifecycle behavior, failure handling, approval restrictions, and supported limitations.
* **Tests**
  * Expanded coverage for parsing, validation, concurrency, isolation, aggregation, and recovery scenarios.


Wirasm:
## PR Review Summary

Multi-agent review (code-reviewer + silent-failure-hunter) at head `50e3f756`, diffed against the stack base (`origin/feat/subrun-child-isolation...HEAD`). Overlapping findings deduplicated; both agents independently converged on the #2180 reconciliation path as the weak spot.

### Critical Issues (2 found — both in the #2180 gate-fail-fast reconciliation, `dag-executor.ts:5385-5419`)

| ID | Agents | Issue |
|----|--------|-------|
| C1 | silent-failure-hunter | The blocked-child pre-check conflates genuinely gate-`paused` children with ambiguous `running`/`pending` ones (crash-orphans or a concurrent resume) and **autonomously cancels all of them** — violating CLAUDE.md's "No Autonomous Lifecycle Mutation Across Process Boundaries" rule (no staleness check, though `last_activity_at` zombie-detection precedent exists in this same file) — then reports a gate message ("Remove the gate from 'X'") even when no gate ever existed. Crash recovery destroys a possibly-good in-flight child and sends the user hunting a nonexistent gate |
| C2 | both | `cancelled` children are threaded like `completed` on re-entry (only `failed` is re-driven) — so once the #2180 path cancels a child, **every future parent resume deterministically re-fails**, with a less informative message each time; the only escape is abandoning the whole run. The error message's advice ("remove the gate... then resume") does not work. This kills 1:N resume for exactly the failure mode PR-C introduces |

### Important Issues (4 found)

| ID | Agents | Issue | Location |
|----|--------|-------|----------|
| I1 | both | In-flight siblings are NOT cooperatively cancelled when the node's fate is sealed (fail-fast trip or #2180) — deviates from the plan's own join-semantics table ("remaining cooperatively cancelled"); with `max_parallel` ≥ item count the fail-fast saves nothing. Disclosed in an amendment but the plan is the settled spec — restore it | `dag-executor.ts:5429-5476` |
| I2 | silent-failure-hunter | Out-of-range `child_index` (items array shrank between attempts) is silently dropped — a still-running orphan becomes invisible, untracked, and still billing; no log, no cancel | `dag-executor.ts:5393-5398` |
| I3 | silent-failure-hunter | 1:N aggregate drops the `workflow.subrun_completed_without_output` warn the 1:1 path has — anomalous empty completions fold silently into the array (parity regression with the precedent the PR's own docstring cites) | `:5532`, `:5540` |
| I4 | both | `fanOutAutonomousGateMessage` omits WHICH child/index/run triggered it (the data is on hand; the join-failure message at `:5520` shows the right shape), and the cancelled row records no reason — no post-mortem trace of why a run was auto-cancelled | `:5418`, `:5509` |

### Suggestions (4 found)

| ID | Agent | Suggestion |
|----|-------|------------|
| S1 | code-reviewer | No cap on resolved `items.length`; >500 items interacts with `MAX_CASCADE_RUNS` on abandon (some children uncancelled). Doc callout at minimum (#1961 owns real governance) |
| S2 | both | Resume identity is index-only with no content verification — safe today only via an upstream invariant (cached producer outputs) enforced in a different file. Add a defensive item-hash in child metadata + mismatch warn |
| S3 | silent-failure-hunter | Inconsistent immediate notification across failure paths (only the all_success join failure sends `safeSendMessage`; others rely on the end-of-run digest) |
| S4 | silent-failure-hunter | Duplicate `child_index` last-write-wins with no debug log (consistent with 1:1 precedent, but log it) |

### Strengths (verified clean by both agents)

- Items-ref parsing is bulletproof against the classic disasters: JSON string/object/blank/malformed all fail closed with node-named messages — never fans out over string characters
- `mapWithLimit` verified correct: atomic cursor, input-order results, no swallowed rejections, no unhandled-rejection window; nothing keeps running invisibly after the node settles (`Promise.all` over all workers)
- Load-time validation thorough: dangling items refs, non-workflow usage, `first_success` cleanly staged as PR-D reject, transitive-upstream requirement, fan-out-in-loop_group rejected
- Cost rollup distinguishes `undefined` from `0`, no double-counting; cascade-cancel generalizes to N with zero changes (keys on `parent_run_id`)
- Isolation composition correct: fan-out defaults to `worktree` (override wins); `inherit` fan-out intentionally hits the existing path-lock per plan D2
- All 17 new tests spot-checked discriminating

### Verdict

NEEDS FIXES — the mechanics are solid; both criticals live in one reconciliation path and are targeted fixes, not a redesign.

### Recommended Actions

1. C1: split the blocked pre-check — genuinely `paused` (gate observed) keeps the designed auto-fail+cancel; `running`/`pending` get the `last_activity_at` staleness treatment: fresh → fail the node WITHOUT cancelling ("child may still be running — wait, or abandon <id>"), stale → treat as orphan with an honest message. Never send the gate message for a non-gate cause
2. C2: tag auto-cancellations (`metadata.cancelled_reason: 'fan_out_gate'` etc.) and re-drive exactly that class on resume (user-cancelled stays terminal) — making the "remove the gate, then resume" advice true
3. I1: restore the plan's cooperative sibling-cancel on fate-sealed (best-effort `cancelWorkflowRun`, bounded by the poll window)
4. I2: warn + cancel out-of-bounds-index children; I3: restore the empty-output warn in both reducers; I4: enrich the gate message with index/run id + thread the cancel reason
5. S1-S4 as cheap hardening alongside

Wirasm:
## Review round addressed — commits `7af52bb7` (fix) + `ee9eddfd` (tests)

All 10 findings fixed. `bun run validate` exit 0; 32 subrun e2e cases (18 slice-1 + 14 fan-out), full workflows suite green.

### Critical

- **C1 — blocked pre-check split by status.** A genuinely gate-`paused` child keeps the designed autonomous-fan-out rejection (cancel + gate message). A `running`/`pending` child (ambiguous: crash-orphan vs a live run elsewhere) is **no longer auto-cancelled** — the node fails *without* mutating it (CLAUDE.md "No Autonomous Lifecycle Mutation Across Process Boundaries"), surfacing a `last_activity_at`-staleness-keyed message: fresh → "may still be running — wait or abandon `<id>`"; stale → "appears orphaned — abandon `<id>`". The gate message is never shown for a non-gate cause. Test: `a running fan-out child found on resume fails the node WITHOUT cancelling it`.
- **C2 — tagged, recoverable auto-cancels.** Every fan-out auto-cancel now stamps `metadata.cancelled_reason` (`fan_out_gate` / `fan_out_sibling` / `fan_out_orphan`) via an `updateWorkflowRun`-then-`cancelWorkflowRun` helper (`cancelWorkflowRun` takes no metadata). On resume a fan-out-tagged cancel is **recovered** — flipped to `failed` (since `resumeWorkflowRun` refuses `cancelled`) then re-driven — so "remove the gate, then resume" actually completes the node. A user's out-of-band cancel (untagged) stays terminal and is never resurrected. Tests: `a fan-out-cancelled gate child is re-driven on resume once the gate is removed` + `a user-cancelled fan-out child (untagged) is NOT resurrected on resume`.

### Important

- **I1 — cooperative sibling-cancel on fate-seal.** When the node's fate is sealed (a pause, or an `all_success` failure), in-flight siblings are best-effort `cancelWorkflowRun`'d (tagged `fan_out_sibling`, re-queried live since `runChild` is synchronous and hasn't returned their ids). Their cancel poll stops them; `mapWithLimit` still awaits all. Test: `fail-fast cooperatively cancels an in-flight sibling` (instant-fail + `sleep 0.2` siblings make the window real; the in-memory harness now mirrors the real store's `completeWorkflowRun` `WHERE status='running'` CAS so the mid-flight cancel sticks).
- **I2 — out-of-range `child_index`** (items shrank) → WARN + cancel a still-live orphan (tagged `fan_out_orphan`), never a silent drop. Test: `an out-of-range child_index (shrunk items) is warned + a live orphan cancelled`.
- **I3 — restored** the `workflow.subrun_completed_without_output` warn in **both** join reducers (`childOutput` helper).
- **I4 — enriched** `fanOutAutonomousGateMessage` with the child index + run id; the cancelled row records the reason (C2). Strengthened the paused-gate test to assert the `fan_out_gate` tag + the `child N (run xxxxxxxx)` shape.

### Suggestions

- **S1** — doc callout in the `fan_out` schema: `max_parallel` caps concurrency, not total count; `items.length` ≳ `MAX_CASCADE_RUNS` (500) can leave children uncancelled on abandon (real governance deferred to #1961).
- **S2** — item-content hash (`hashFanOutItem`, djb2) stamped in child metadata at spawn; resume warns on a mismatch (non-deterministic producer) but still re-keys by index.
- **S3** — `safeSendMessage` now fires on the items-resolution and #2180 failure paths (previously only the `all_success` join failure notified immediately).
- **S4** — duplicate `child_index` is debug-logged (last-write-wins retained, per the 1:1 precedent).

One implementation note for the record: with concurrent gate children (`max_parallel > 1`), the first to pause seals the fate and the rest are sibling-cancelled (`fan_out_sibling`) rather than `fan_out_gate` — both are recoverable, so resume behaviour is identical; the assertion test pins `max_parallel: 1` for a deterministic single gate observation.

Wirasm:
Reviewed against `dev`.

**On the constitutional question — this lands on the right side.** Each child is a real `workflow_runs` row with its own governance, artifacts, gates and cost, which is precisely the escape `§load-time-composition` names ("runtime-resolved structure is a sub-run … not a language feature"). The parent DAG the executor runs stays flat and static; the fan-out input is *data* (a `$node.output` array), not structure. No templated targets, no dynamic workflow names.

That said, there is a `§schema-width` violation inside the diff, and it's the blocking one.

## Blocking

**1. Reject `fan_out.as` fail-fast instead of silently ignoring it** (`schemas/dag-node.ts:491`).

`as` has **zero runtime effect**, and `$INPUTS` does not exist anywhere in `packages/workflows/src/`. An author who writes `as: task` and then `$INPUTS.task` in the child prompt gets the literal string `$INPUTS.task` delivered to the model — silently wrong output, no error, no warning.

That contradicts three things at once: `§schema-width` ("capability mismatches warn loudly, never silently no-op"); the three existing precedents in this same `superRefine` (`:752` `with:`, `:758` `retry:`, `:794` `first_success`), which all reject unbuilt surfaces fail-fast; and the `first_success` rejection's own comment, which says silently degrading *"would surprise an author"*. That is the argument against `as:`, written by us, ten lines away.

Keep the schema field so #2214 only has to lift a guard — add the reject plus one loader test. ~10 lines.

**2. Make the `all_success` failure message name the causal child** (`dag-executor.ts:5747-5760`). `firstBad = outcomes.findIndex(o => o.status !== 'completed')` takes the lowest-index bad outcome, but the sibling-cancel *creates* bad outcomes at indices below the real failure. With `max_parallel: 3` over 5 items where child 2 fails first, children 0 and 1 are still in flight, get sibling-cancelled, and the node reports `child 0 (run abc12345) cancelled` with no error text — the actual failure is invisible in the node's message and the operator has to go read child run rows. Prefer a genuinely `failed` outcome (or the sealing index) when composing it. ~5 lines.

**3. Do not merge before #2223.** It is the base and supplies `ChildIsolationRequest.childIndex`; the resolver types don't exist on `dev`. Re-verify the diff after GitHub retargets. Note #2223's identifier-collision bug is unfixed here too, since this builds on the same identifier — fan-out makes it worse, because parallel children are the normal case rather than the exception.

**4. Don't ship fan-out in a release without the docs.** Two obligations unmet: the authoring-docs behavior-list entry for the implicit `fan_out ⇒ isolation: worktree` default (required by `§implicit-behavior` — documented, individually defeatable, fail-safe), and the constitution case-law row at `workflow-language-constitution.md:41`, which still says fan-out "remain[s] deferred". A shipped YAML field with no page on archon.diy is the half-feature this chain most risks.

**5. Re-read the two fix commits yourself** (`7af52bb7`, `ee9eddfd`). CodeRabbit skipped this PR entirely — non-default base branch — so the reconciliation path that both review agents independently flagged as the weak spot has had exactly one reviewer: the agent that wrote the fix. Its status looks green at a glance and is a skip notice.

Wirasm:
Two changes needed before this can land.

**C1 — Drop the implicit `isolation: worktree` default on fan-out nodes.**

In `packages/workflows/src/schemas/dag-node.ts` the transform injects `isolation: 'worktree'` whenever `fan_out` is present and `isolation` is absent:

```ts
...(data.isolation !== undefined
  ? { isolation: data.isolation }
  : data.fan_out !== undefined
    ? { isolation: 'worktree' as const }
    : {}),
```

Remove that branch. A child gets a worktree **only** when the author explicitly writes `isolation: worktree`. With no `isolation:` key, a fan-out child inherits the parent checkout exactly like any other sub-run.

Reasoning: fanning out is not inherently a write operation. Fanning out N review nodes over a diff, or N research nodes over a list of questions, is an ordinary — probably the most common — use of this feature, and none of those children need their own checkout. Handing them one costs a worktree, a branch, and cleanup for nothing. Deciding when isolation is needed is the author's job, not Archon's. The collision argument in the comment only applies to children that write, and the author is the one who knows whether they do.

Run-level worktree-by-default is a different case and is correct as it stands — that's a whole run against a repo with an owner and a lifecycle. This is a per-node default inferred from an unrelated field, which is not the same thing.

Keep the fail-fast when `isolation: worktree` is requested and no resolver is injected. That part is right — no silent shared-checkout fallback.

**C2 — Docs are missing.**

This PR adds YAML surface (`fan_out:` with `items` / `as` / `max_parallel` / `join`) and touches nothing under `packages/docs-web/`. Same story on #2223 (`isolation:` on workflow nodes) and #2250 (`join: first_success`). Per CLAUDE.md, YAML-surface features need documentation and a constitution citation in the PR. Please add:

- the authoring reference for `fan_out:` — all four fields, with the defaults (`max_parallel: 5`, `join: all_success`) stated
- the isolation semantics after C1: explicit opt-in, and what happens when a worktree is requested without a resolver
- the re-entry behavior an author has to be able to predict: a child pausing at a gate fails the node, `child_index` keying on resume, and what happens when the item list changes between attempts
- the `items` caveat — `max_parallel` caps concurrency, not total child count, and the abandon cascade bound applies

Wirasm:
Addendum to C1 — this is the part that has to be handled alongside removing the implicit default, or the removal makes things worse rather than better.

C1 still stands: only an explicit `isolation: worktree` should produce a worktree. But dropping the default alone leaves a sharp edge, because of Defect A in #2180.

`executeWorkflow` takes a path-exclusive lock at run start (`packages/workflows/src/executor.ts:1158`, guarded by `if (workflow.mutates_checkout !== false)`). The lock excludes the run's own vertical tree line — ancestors and descendants — but **siblings are deliberately not excluded**. N fan-out children of the same node sharing the parent checkout are siblings of each other. So with the default removed and nothing else changed, child 0 takes the path, and children 1..N-1 lose the older-wins tiebreaker and **self-cancel**. The read-only review/research fan-out — the case that most wants a shared checkout — is exactly the case that breaks.

The escape hatch already exists and is the right one: `mutates_checkout: false` on the **child** workflow (`packages/workflows/src/schemas/workflow.ts:141`) skips the lock entirely. That is author-declared, which is where this decision belongs — the author of a review or research workflow is the one who knows it only reads. A fan-out of `mutates_checkout: false` children needs no worktrees and no locks, and N of them coexist in one checkout correctly.

So the shape I want:

1. Remove the implicit `isolation: worktree` injection (C1 as written).
2. Do **not** let an author discover the sibling collision at runtime as N-1 mystery cancellations. Fail fast with a message that names the three ways out: declare `mutates_checkout: false` on the child if it only reads, set `isolation: worktree` on the fan-out node if the children write, or set `max_parallel: 1` if serial-in-place is what they meant. Load time if the child target resolves there; otherwise at spawn time, before any child is created.
3. `mutates_checkout` currently appears nowhere in `packages/docs-web/` — it is an undocumented flag that the fan-out story now depends on. It needs to be documented as part of the C2 docs work, not separately.

The general principle, so it isn't relitigated per feature: the engine's job is to make the author's declaration hold, not to infer what they must have meant from an unrelated field. `fan_out:` says how many children to run, not whether they write to disk. `mutates_checkout:` is where that gets said.

Wirasm:
Filed the gate constraint as a follow-up: #2438 — not a blocker for this PR.

Summary of what I verified while writing it, since it sharpens the C2 docs ask: gates are only impossible *during* the expansion. A gate **before** the fan-out node is an ordinary parent approval node with no interaction. A gate **after** it resumes correctly — the node is skipped via `priorCompletedNodes` and `$<id>.output` still holds the aggregate, because `writeCompleted` stored it on the `node_completed` event. And a gate inside a **1:1** sub-run child already works today (child pauses, tree pauses, approve the child by run id, parent auto-resumes).

So the only broken position is a gate inside a child that is fanned out — which means the same child workflow pauses when spawned 1:1 and hard-fails when spawned N-wide. That asymmetry is what needs to be documented and, ideally, detected at spawn time before N children are created rather than surfaced as a runtime failure after the cost is spent.

Wirasm:
## C1, C1-addendum and C2 addressed — rebased onto `feat/subrun-child-isolation@b8f1d82e`

Three commits: `4b3c4a09` (behaviour), `1c5ac40e` (docs), `7614fcbe` (regenerated OpenAPI types). `bun run validate` exit 0 — all five generated-file checks, type-check, lint at `--max-warnings 0`, format, `test:install`, and every package suite at 0 fail. The docs site builds clean.

### C1 — the implicit `isolation: worktree` is gone

The transform branch in `schemas/dag-node.ts` is removed; `isolation` is now `...(data.isolation !== undefined ? { isolation: data.isolation } : {})` for every `workflow:` node, fan-out or not. A child gets a worktree only when the author writes it. The fail-fast for `isolation: worktree` with no injected resolver is untouched.

### C1 addendum — where the collision check landed, and what it says

**At spawn time, in `executeFanOutWorkflowNode`, immediately before the `mapWithLimit` window** — after items resolution and after the existing-children reconciliation, so no child row exists yet and nothing has been billed.

It could not go at load time. The child target resolves at spawn by design (#2200), so the loader cannot read the child's `mutates_checkout`. I resolve it with the same `discoverWorkflowsWithConfig` + `resolveWorkflowName` pair `runChildWorkflow` uses, which means the preflight reads the definition the children will actually get rather than one captured earlier.

It fires when **all** of: `isolation !== 'worktree'`, the resolved child does not declare `mutates_checkout: false`, and this attempt would drive more than one child at once. Two deliberate narrowings:

- **Counted over the indices this attempt will drive, not `items.length`.** A resume with one instance left to re-drive has no concurrency, and blocking it would block recovery from an unrelated failure.
- **`max_parallel: 1` is not a collision.** The window awaits each child, so the previous one's lock is released before the next starts — which is exactly why it is offered as one of the three ways out.

It fails **open** when the target name doesn't resolve (unknown, ambiguous, discovery error). Every child would fail at `runChildWorkflow` anyway, with a message about the real cause; reporting a collision the author cannot act on would send them to the wrong file.

The message:

```text
fan_out node 'review-each': up to 3 children of 'review-one-file' would run at once in the
parent checkout, and that workflow does not declare `mutates_checkout: false`. Concurrent
runs on one checkout take a path-exclusive lock, so all but the first would cancel
themselves — and a lock-cancelled child is not recoverable by resume (#2180). Choose one:
add `mutates_checkout: false` to 'review-one-file' if it only reads the repo; set
`isolation: worktree` on 'review-each' if the children write to it; or set
`fan_out.max_parallel: 1` to run them one at a time.
```

**The collision is not theoretical — I reproduced it.** With the schema default removed and the guard absent, a 3-item `max_parallel: 2` fan-out over a child with no `mutates_checkout: false` produces 2 child rows, a self-cancelled sibling, and a failed parent, in the e2e harness that drives the real executor and the real path lock. That intermediate state is what the guard exists to prevent, and it is why both halves are in one commit rather than two.

### C2 — docs

`authoring-workflows.md` gains **Fanning out over a list with `fan_out:`** in the sub-run section, covering: the four fields and their defaults; `as` stated as having no runtime effect today; `max_parallel` bounding concurrency and not count, with the unbounded `items` and the `MAX_CASCADE_RUNS` (500) abandon bound spelled out; the join table with `first_success` rejected at load; the isolation rules with the guard message and a table mapping *what the children do* → *which escape*; and the resume section — `child_index` keying, what is re-driven vs threaded vs deliberately left alone, and both directions of a changed item list (shorter → orphan cancelled; same index, new content → warn).

`mutates_checkout` itself was documented on #2223 (`dd098b68`), so this section links to it rather than restating it.

The constitution's case-law table records fan-out as **shipped**, with the admissibility argument: the target is a static name, only the count is runtime, each child is its own governance object, and the parent DAG stays flat and static.

### Gates (#2438)

Written the way you framed it: gates **bracket** an autonomous middle, and both bracket positions are documented as supported, with the pre/post pattern shown. The asymmetry is stated plainly as the thing to know — the same child workflow pauses correctly 1:1 and hard-fails fanned out — and #2438 is linked for the spawn-time detection. No apologising for a limitation that is the intended shape.

### `childIndex`

Already threaded through on this branch (`executor.ts:677` passes it into `resolveChildIsolation.resolve`), and covered by the first fan-out e2e case, which now declares `isolation: worktree` explicitly and asserts three distinct child indexes and three distinct working paths. Combined with `b8f1d82e`'s identifier fix, fan-out siblings get genuinely distinct worktrees.

### Tests

Four new cases, each confirmed to **fail before** the change:

| Test | Before |
|---|---|
| `blocks a shared-checkout fan-out over a repo-writing child BEFORE any child is spawned` | run succeeded with 3 worktree children |
| `read-only children (mutates_checkout: false) fan out IN the parent checkout, no worktrees` | 3 resolver calls, 3 worktrees |
| `max_parallel: 1 is a valid serial-in-place fan-out over a repo-writing child` | 3 resolver calls, 3 worktrees |
| `does NOT infer isolation from fan_out — an omitted isolation stays omitted` (loader) | `isolation === 'worktree'` |

The existing fan-out fixtures now declare what they actually do rather than leaning on the removed default: the read-only children carry `mutates_checkout: false`, and the two that write a marker file into their checkout carry `isolation: worktree`. Suite is 35 e2e cases, 0 fail.

### One thing left over, from an earlier round

`fan_out.as` is still accepted and silently ignored — the fail-fast reject you asked for in the "Reviewed against `dev`" comment (blocking item 1) is not in the tree. I left it alone because it sits outside the three comments I was working from, and documented it as "no effect today, don't write it yet" instead. It is still the ~10-line guard next to the `first_success` reject, and I'd rather it land than ship a documented no-op field. Same for blocking item 2 (the `all_success` message naming a sibling-cancelled child rather than the causal one).

Wirasm:
### Follow-up: two pre-existing test flakes found and fixed (`21521adf`)

After the branch was green I re-ran the workflows package batches under load and hit a failure in roughly one run in three. Two different tests, one cause, and neither was related to the change above — both predate it (`ee9eddfd`).

With `max_parallel > 1` and `join: all_success`, the first child to fail seals the node's fate and cooperatively cancels whichever siblings are still in flight, tagged `fan_out_sibling` (I1). Which child wins that race is timing:

- **`1:N re-entry`** asserted indexes 0 and 2 were `completed` after run 1. When a sibling lost the race it was `cancelled` instead.
- **`a user-cancelled fan-out child (untagged) is NOT resurrected`** cancelled index 0 on the assumption it had failed on its own. When index 0 was the loser it already carried `cancelled_reason: fan_out_sibling` — which *is* recoverable — so resume legitimately resurrected it and the test asserted the opposite of its own subject. I reproduced this deterministically by making the item at index 0 sleep, and confirmed the tag was already there before the test's own cancel.

Both now run serial, which pins run 1 exactly and leaves each test's subject untouched. The user-cancel test asserts its precondition (index 0 `failed`, untagged) so the race can't come back silently as a different test.

One thing worth recording, because it cost me a wrong fix first: **`1:N re-entry` had no working observable for its headline claim.** Row count can't show that a completed child was threaded rather than re-driven — a re-drive reuses the row. My first attempt used an append-only side-effect file in the child's worktree, and it passed even when I forced index 0 back to `failed` before the resume: the child's *own* DAG resume skips its already-completed node, so a re-drive produces no new side effect. `completed_at` does discriminate — forcing the re-drive changes it, verified. So the test asserted something true and unfalsifiable until now.

Serial run 1 also means index 2 is never spawned, which turns out to be better coverage: the test now exercises three re-entry paths in one pass — a completed child threaded, a failed child re-driven in place, and a never-started index spawned fresh.

`bun run validate` exit 0 again, and six consecutive `bun --filter @archon/workflows test` batches clean.

Wirasm:
## Both remaining blocking items from the 2026-07-27 review are closed

`c98fbff0` (`as` reject), `ee57f1c9` (causal child), `bb69eebe` (docs correction). `bun run validate` exit 0 — grepped the log, zero `(fail)` lines, `Installer tests passed`, no non-zero exits. Five consecutive `bun --filter @archon/workflows test` batches clean, docs site builds.

### 1. `fan_out.as` is rejected at load

Staged exactly like `join: first_success`: the key stays in the schema so PR-B (#2214) lifts a guard rather than migrating anyone's YAML, and the `superRefine` rejects it now.

```text
'fan_out.as' (the $INPUTS channel) is not yet supported (PR-B, #2214). Remove it —
each item is delivered to the child as $ARGUMENTS, which the child's prompts can use today.
```

Same principle as the isolation work: the engine shouldn't accept a declaration it has no intention of honouring. No bundled default uses `fan_out` at all, so nothing in the tree breaks.

The docs row changed with it — it described `as` as "no effect today", which stops being right once it's rejected. It now says rejected, and why.

### 2. The `all_success` failure names the causal child

The old selection took the lowest-index non-completed outcome. Under `max_parallel > 1` that is routinely the wrong child, because sealing the node's fate cancels every in-flight sibling and skips every unspawned index — and those casualties sit **below** the child that caused them. With the failing item last, reverting my change reproduces exactly:

```text
fan_out node 'work' (join: all_success): child 0 (run run-2) cancelled
```

No error text, and the actual failure (child 2, exit 7, "the real cause") nowhere in the message.

**On determinism, since you flagged it.** I did not use the sealing index — which child wins the seal race *is* timing-dependent, so keying off it would be exactly the "usually right" answer you warned about. Instead casualties are excluded **by identity**: a sibling this node cancels is recorded in `failFastVictims` at the moment of cancellation, and an index never spawned carries the synthetic skip error (now a named constant, `FAN_OUT_SKIPPED_BY_FAIL_FAST`, rather than matched prose). The lowest-index outcome that is neither wins.

That is stable regardless of who wins the race: whichever sibling loses is excluded either way, and a sibling that genuinely failed on its own stays eligible. It falls back to the old behaviour only if *every* bad outcome is a casualty — a state the seal logic shouldn't produce, since something has to cause the seal, but a failed join must name a child rather than report nothing.

**Test construction.** Ordering comes from a shared marker file, not sleeps: the waiting children block until the failing child creates it, so the failure always lands first and only the cancel window is timed. The marker path is absolute and travels in the item itself (the parent's `plan` node emits `$PWD`), so it works regardless of where the children run. The test asserts its construction materialised — index 2 `failed`, at least one lower index tagged `fan_out_sibling` — separately from its subject, so a scenario that failed to build fails *there* rather than quietly degrading into an assertion the old code would also pass. 10/10 repeat runs clean.

### The trap worth writing down

Both of my earlier flake fixes and this test hit the same thing, so stating it rather than leaving it to be rediscovered:

> **Any fan-out test that combines `max_parallel > 1` with a deliberate failure has a timing-dependent loser, and the loser gets a *recoverable* tag (`fan_out_sibling`).**

Three distinct ways that bites:

1. A test asserting a sibling is `completed` gets `cancelled` instead (the `1:N re-entry` flake).
2. A test that cancels a child to simulate a user cancel may be cancelling one that already carries the recoverable tag — so resume legitimately resurrects it and the test asserts the opposite of its own subject (the user-cancel flake). This one is nasty because it fails as a *wrong* assertion, not an obviously broken one.
3. A test reading the failure message may be reading a casualty's, not the cause's (this PR's item 2).

The defence in all three cases is the same and cheap: go serial where concurrency isn't the subject, and where it *is* the subject, assert the construction as a precondition so the race can't silently change what is being tested.

One more from the flake work worth recording: **a side-effect file is not a valid observable for "this child was not re-driven."** The child's own DAG resume skips its already-completed node, so a re-drive produces no new side effect and the assertion passes either way. `completed_at` discriminates; verified by forcing the re-drive.

Wirasm:
## Fan-out no longer fail-fasts — children run to terminal (`326c8d67`)

Semantics change: every index spawns, every child reaches its **own** terminal state, and only then does the join reduce. No child's outcome ends another's run. `all_success` keeps its verdict — any failed child still fails the node — it just stops killing the others to get there.

`stopSpawning`, `sealFateCancelSiblings()` and the synthetic skip outcome are gone.

`bun run validate` exit 0, grepped: zero `(fail)` lines, `Installer tests passed`, no non-zero exits. 36 subrun e2e cases, 8/8 repeat runs clean.

### The paused child — and why "leave it untouched" wasn't available

You asked me to decide what survives here and not to silently leave a second route to sibling cancellation. Removing the seal **created** one, and it was worse than what it replaced.

A pause is not terminal, and a non-terminal run keeps holding its working path — `getActiveWorkflowRunByPath` counts `paused` as active. With the seal gone the next sibling actually spawns, loses the path lock to the still-paused child, and self-cancels **with no reason tag**. Untagged means non-recoverable, so the parent fails identically on every resume with no way back. The gate-recovery test went red and told me exactly this:

```
["fan-child-recover", 0, "completed",  "fan_out_gate"]
["fan-child-recover", 1, "cancelled",  null]            ← untagged path-lock self-cancel
```

Fix: cancel the paused child the moment its outcome is seen rather than at the join. That is not one child's outcome ending another's — the cancel is decided by the paused child's own state, it is the same cancel the gate path (#2180/#2438) was about to apply a moment later, and every sibling still runs to its own terminal state. All that changes is that the lock is released before the next child starts. Disabling just that line reproduces the failure.

So the gate path keeps its behaviour (node fails, paused children tagged `fan_out_gate`, recoverable on resume) and loses only its power to stop siblings.

### The causal-child work

Deleted, and the simplification is real: with no fail-fast nothing is cancelled by a sibling and no index goes unspawned, so no casualty can exist and the lowest-index non-completed outcome **is** the failure. `pickCausalOutcomeIndex`, `failFastVictims` and `FAN_OUT_SKIPPED_BY_FAIL_FAST` all go. I checked for a remaining producer of an unspawned-index skip before removing them — there is none.

### `fan_out_sibling`

Kept, read-only. Nothing writes it, but a run in flight across the upgrade has rows carrying it, and dropping it from the recoverable set would make those children read as user-cancelled — terminal, never re-driven — so the parent would fail every resume. Documented as legacy with the condition for eventually deleting it.

### Tests got simpler, as you predicted — with one catch

Two tests pinned to `max_parallel: 1` purely to dodge the fail-fast race are concurrent again and deterministic. The causal-child test lost the marker-file choreography that existed only to place casualties below the real failure.

**The catch is worth recording.** My first version of the contract test used instant children and **passed against the pre-change executor** — the survivors finished before any cooperative cancel could reach them, so the test proved nothing. Both contract tests now use sleeping survivors so they are genuinely mid-flight when the failure lands, which is the window the old fail-fast fired in. Verified: both fail before the change, both pass after.

### Cost, documented

A fan-out whose first child fails now runs every remaining child. Worst-case spend is `items.length`, not "until the first failure", and `max_parallel` caps concurrency rather than total. The docs say so plainly and point at #1961 — that budget ceiling is now load-bearing rather than theoretical.

### One thing for you, not actioned

`feat/subrun-racing` (#2250) is stacked on this branch and will need a rebase onto `326c8d67`. Worth knowing before it happens: racing's own `sealRace` aborts and cancels losers the moment a winner appears, which is the same mutual termination this decision removes. I have not touched it — you said racing's disposition is the maintainer's call — but the two are now in direct tension, and #2250 cannot simply rebase clean past that question.

Wirasm:
## `join` now defaults to `all_done` (`1c12dc5c`)

Rebased on your constitution commit `f07341c3` — I wrote the docs against the independence rule rather than around it.

`bun run validate` exit 0, grepped: zero `(fail)` lines, `Installer tests passed`, no non-zero exits. 37 subrun cases, 184 loader cases.

### What changed

Removing the fail-fast fixed half of this: a failed child stopped *cancelling* its siblings. But under `all_success` it still **failed the node**, so a downstream collector received nothing even though every sibling had run to completion and produced good output. Same loss, later in the pipeline.

Now every terminal outcome aggregates, failures represented as `{error, status}` in their own slot, and the node succeeds. `all_success` stays for the genuinely dependent case — where a gap makes the aggregate meaningless rather than smaller.

### Stranded prose, checked

- The worked example passed `join: all_success` explicitly. Dropped — for a per-file review fan-out the default is now the right answer, and the example teaches better without it.
- The fields table and join table carry the new default; the join table's rows are reordered so the default reads first.
- `pickCausalOutcomeIndex` remnants: none — that went with the fail-fast in `326c8d67`, and the grep is clean.
- `trigger_rule` also defaults to `all_success` and is a **different** field (DAG-dependency join, not fan-out). Left alone deliberately, so the docs now have two nearby `all_success` defaults that mean different things — flagging it in case that reads as an inconsistency later.

### Why, not just what

New section: *Why `all_done` is the default*. It gives the independence reasoning in the maintainer's terms — ten triage children over ten issues are ten jobs that happen to run together, one failing says nothing about the other nine — and then shows the shape that replaces a threshold:

```yaml
  - id: triage-each
    fan_out: { items: "$list-issues.output" }   # join: all_done — the default
  - id: check
    script: |
      const results = $triage-each.output;
      const ok = results.filter(r => typeof r === 'string');
      console.log(JSON.stringify({ ok: ok.length, total: results.length }));
  - id: report
    when: "$check.output.ok != '0'"
```

And states why there is no `succeed if ≥ K` value: how many results are enough depends on *which* children failed and why, it changes between runs, an enum cannot weigh it, and adding one starts a policy language inside a YAML field. Written so the next person proposing it finds the answer rather than re-opening it.

### Tests

`the DEFAULT join succeeds with a failed child, aggregating all three outcomes` — one failed child, two completed, node and run succeed, all three outcomes present in item order with the failure as data in its own slot. Fails before the flip (`result.success` false). The loader default test also flips and fails before.

Two resume tests now declare `join: all_success` explicitly rather than relying on the default. Both are about re-driving a **failed** instance, which needs the node to fail — they are the dependent case, and saying so is more honest than leaning on a default that no longer means that.

Wirasm:
## Racing rejection message corrected + the two joins disambiguated (`9b7faefe`)

`bun run validate` exit 0, grepped: zero `(fail)` lines, `Installer tests passed`, no non-zero exits. Docs build clean, both new anchors resolve.

### The message

Before — a promise nobody was going to keep:

```text
'fan_out.join: first_success' (racing) is not yet supported (PR-D). Use 'all_success' or 'all_done'.
```

After:

```text
'fan_out.join: first_success' (racing) is rejected, not deferred: a winner cancels the
losers, which couples children that are meant to be independent. Use 'all_done' (the
default). For several genuinely different attempts, write them as separate nodes with their
own models feeding one collector node — every attempt is kept and nothing is cancelled.
```

The docstring (`:494`) and the superRefine comment (`:805`) carried the same dead assumption and are rewritten with it. `PR-D` no longer appears anywhere in the package outside tests.

### The enum value stays, and the comment says why

Its original justification — "so the eventual PR only has to lift a guard" — is gone. The replacement is written into the comment so it does not get removed as dead weight later: an author whose YAML already says `first_success` gets a message naming the rejection and the alternative, instead of an opaque unrecognised-enum-value error.

### Docs

New **Why there is no racing join** section with the replacement pattern written out — separate nodes, own models, one collector, `trigger_rule: none_failed_min_one_success`. I made the case that this is *strictly better* than racing at what racing was wanted for, rather than presenting it as a consolation:

- the attempts can differ by **model**, which a fan-out cannot express at all;
- every output survives for the collector to weigh, instead of N−1 being thrown away;
- selection is a judgement made by a node that can read the work, not by a stopwatch.

The join table row and the non-goals bullet both said "reserved"/"a later slice" — corrected.

### The two joins

Taking your call: documented rather than renamed.

- The node-fields table entry for `trigger_rule` now cross-references `fan_out.join` and states the different default inline.
- The `trigger_rule` Values section gets a note contrasting them: `trigger_rule` (any node, `all_success`) decides whether *this* node runs given its dependencies; `fan_out.join` (fan-out only, `all_done`) reduces one node's N children.

It also gives the reason the defaults differ, since that is what makes it stick rather than something to re-look-up: **upstream dependencies are steps you chose to sequence; fan-out children are N instances of one step.**

### Test

The loader test now asserts the message says `rejected, not deferred`, names a `collector`, and — the part that matters — does **not** contain `not yet supported` or `PR-D`. Negative assertions on the exact dead phrasing, so this specific regression cannot come back quietly.

That is everything on the branch from my side.

Wirasm:
## Multi-agent review: #2224 — dynamic fan-out for `workflow:` sub-run nodes

**Recommendation: REQUEST CHANGES** — two small code defects and two stale artifacts. The design is sound and the engineering is unusually disciplined; nothing here is architectural.

Reviewed at `c6a549a4`, rebased on the merged PR-A (`41bc0673`). Eight specialist agents (correctness, silent-failure, concurrency/resume, constitution admissibility, tests, types, docs, simplification). Every finding below was re-verified by hand against the code before inclusion — three agent claims were corrected or dropped in the process (noted at the end).

### Validation

| Check | Result |
|---|---|
| `bun run type-check` | ✅ exit 0, all 12 packages |
| `@archon/workflows` tests | ✅ 1586 pass / 0 fail (18 split invocations) |
| CI | ✅ ubuntu + windows + docker-build green |
| Mergeable | ✅ CLEAN |

> Note for anyone re-running locally: bare `bun test` in `packages/workflows` reports 574 failures. That is the documented `mock.module` pollution, not this PR — `bun run test` (which splits the invocations) is clean.

---

## Should fix before merge

### 1. `fan_out_orphan` is missing from the recoverable-cancel set — permanent wedge

`packages/workflows/src/dag-executor.ts:5595-5599`

The docblock directly above states these are reasons the fan-out stamps on children it cancels **itself**, "unlike a user's out-of-band cancel — recoverable on resume", and names `fan_out_orphan` explicitly. The type has three members; the set has two:

```ts
type FanOutCancelReason = 'fan_out_gate' | 'fan_out_sibling' | 'fan_out_orphan';
const FAN_OUT_RECOVERABLE_CANCEL_REASONS: ReadonlySet<string> = new Set<FanOutCancelReason>([
  'fan_out_gate',
  'fan_out_sibling',   // ← 'fan_out_orphan' absent
]);
```

`isFanOutRecoverableCancel` therefore reads an engine-cancelled orphan as a *user* cancel and threads it as terminal (`:6102`). Failure path: items shrink on one attempt (indices 3–4 cancelled `fan_out_orphan` at `:5970`), then grow back — those slots are permanently `{status:'cancelled'}` under `all_done`, and under `all_success` the node fails on every subsequent resume with no way back. Narrow to reach, but it is the exact wedge the `fan_out_sibling` comment goes out of its way to prevent.

**Fix:** add `'fan_out_orphan'` to the set. One line.

### 2. `tokens` dropped from the fan-out `node_completed` event — silently under-reports usage after any resume

`packages/workflows/src/dag-executor.ts:5843-5856`, call sites `:6226` and `:6245`

`writeCompleted` persists `cost_usd` but has no `tokens` parameter and never writes one. The 1:1 sub-run writer 400 lines up (`:5429-5435`) writes both, with a comment explaining precisely why it matters:

> *"tokens are the axis every provider reports (Codex reports no cost at all), so dropping them here while keeping cost would hide the one comparable number."*

Verified end-to-end, and the consequence runs opposite to intuition: `getDagResumeSnapshot` (`packages/core/src/db/workflow-events.ts:252-265`) rebuilds cumulative usage across resume passes by summing **`data.tokens` only — it never reads `cost_usd`**. That feeds `priorTokenUsage` → `runCtx.totalTokensIn/Out`. So tokens are the axis that survives resume, and fan-out is the one node type that drops them. Any resumed run containing a completed fan-out node under-reports by exactly the children's tokens; an absent `tokens` key is skipped with no warning (the `else` at `:265` only fires on a *malformed* value). On a Codex fan-out the loss is total — no cost is reported on that provider either.

`totalTokens` is already computed and already returned in the in-memory result, so this is purely a persistence gap.

**Fix:** thread `tokens` through `writeCompleted` and add `...(tokens !== undefined ? { tokens } : {})` to the event `data`, mirroring `:5435`.

### 3. The PR description argues for a design this branch reversed

Four claims in the body are stale — each contradicted by a later commit on this same branch:

| PR body says | Final code | Reversed by |
|---|---|---|
| "Fan-out defaults `isolation` to `worktree` so children don't collide" (repeated in the diagram, Architecture, Side Effects, and Risks) | isolation is **explicit-only, never inferred**; a spawn-time preflight exists *because* it isn't defaulted | `510406f2` |
| "`as:` is an accepted forward seam with no runtime effect yet" | rejected at load in `superRefine` | `419549d5` |
| "`first_success` … rejected fail-fast as PR-D" | rejected outright; there is no PR-D | `c6a549a4` |
| Verified: "all_success serial fail-fast (later items not spawned)" | fail-fast **removed**; every index spawns and runs to terminal | `86e6cd1e` |

This is worth more than cosmetics: the body argues for inferring `isolation` from `fan_out`, which is exactly the inference the constitution rule *this PR adds* forbids — and which the code correctly avoids. A reviewer reading the body reviews an abandoned design.

Worth stressing that the staleness is confined to the PR body: the **user-facing docs were kept in sync** and are correct on every one of these. `authoring-workflows.md:1158` describes `as` as "**rejected at load** … rather than accepted and ignored", and `:1240` describes `first_success` as "**rejected**, not postponed." The guide tracked the reversals; the description did not.

**Fix:** rewrite those four bullets and the risk row.

### 4. Generated API types carry the wrong `join` default

`packages/web/src/lib/api.generated.d.ts:3654` documents `@default all_success`; the schema defaults `all_done` (`schemas/dag-node.ts:511`). The regeneration commit (`e44367b0`) lands *before* the default change (`4ec5e41e`). Inert today — nothing under `packages/web/src` consumes `fan_out` yet — but `validate` has no OpenAPI-drift gate, so CI will not catch it.

**Fix:** re-run `bun --filter @archon/web generate:types` and commit.

---

## Suggestions

**Fail-open in the shared-checkout collision guard** — `dag-executor.ts:6057-6073`. `resolveFanOutChildDefinition` catches *everything* and returns `undefined`, logged at `debug`. On `undefined` the preflight is skipped entirely, and `resolveWorkflowName` also returns `undefined` on an ambiguous name, so this is reachable without any exception. That guard is the only thing standing between a shared-checkout fan-out and the path-lock cascade the comment itself calls permanent. At minimum raise the log to `warn`; consider failing closed when `plannedConcurrency > 1` and the child cannot be resolved (the spawn is about to fail anyway).

**Missing-`child_index` rows are silently dropped** — `dag-executor.ts:5954`. A bare `continue`, with no warn and no cancel, sitting directly above an out-of-range branch that *does* both. `findChildRuns` filters on `parent_node_id` only, and the 1:1 path stamps the same key without an index — so converting a node from 1:1 to `fan_out:` between attempts leaves a live, billing, untracked child that is never cancelled and never reported. Fold it into the same warn+cancel-if-live handling.

**Cost/tokens are Σ *completed* children, not Σ children** — `dag-executor.ts:5736`, `:5749`. Verified: usage metadata is written in exactly one place, inside `completeWorkflowRun` (`:8256-8264`). A failed or cancelled child persists no spend, so `childOutcomeFromRun` returns `undefined` for it. A 10-item fan-out where 3 children burn tokens then fail reports the spend of 7. Pre-existing (inherited from the 1:1 path), but fan-out is what makes it material. A real fix belongs upstream in `failWorkflowRun`; at minimum correct the docblock and the PR body.

**One early-failure branch skips `notify()`** — `dag-executor.ts:5994-5997`. The `findChildRuns` catch returns `failResult` without notifying, unlike every sibling early-failure branch (items resolution, both gate paths, collision, join). One line.

**Three docs pages now state the gate rule in a way that is wrong for fan-out.** These weren't touched by this PR, but the PR is what makes them incorrect — each describes the 1:1 pause-propagates-to-parent behavior unconditionally, with no fan-out carve-out, so a reader who builds a gated child and then fans it out expects "blocked on child" and gets a hard node failure:
- `book/quick-reference.md:163` — "a child approval gate pauses the whole tree — approve the **child** by run id and the parent auto-resumes." The same page's "Workflow (sub-run)-specific fields" table (`:158-163`) also lists only `input` and `isolation`, while the page's frontmatter promises "every … YAML option in one scannable page."
- `reference/api.md:320` and `reference/cli.md:369` — same paragraph shape, same omission.

Also `CLAUDE.md:442`: the `workflow:` node-type description enumerates `input`, `isolation`, the `with:`/`retry:` rejection, the loop_group rejection and the abandon cascade, but has no `fan_out` clause.

**Constitution case-law table reparented** — `workflow-language-constitution.md`. Pre-PR, `### Case law` (line 44) sat under `## The admissibility test` (34). The new `## The independence rule` inserted at 44 pushes it to 64, now reading as a subsection of the independence rule alone. The table holds verdicts driven by *both* rules (`include:` ✅ is an admissibility verdict; "isolation inferred from fan-out" ❌ is an independence verdict) plus older entries tied to neither. Best fix is promoting it to `## Case law` — a sibling to both rules — rather than reordering them.

**Duplication worth collapsing** — the 10-field `childArgs` object is rebuilt inline at `dag-executor.ts:6124-6134`, duplicating `:5541-5552`; a new field on `RunChildWorkflowArgs` must be added in both with no compiler help. `loader.ts:301` hand-rolls a second copy of the canonical `outputRefPattern` defined at `:239` — inside a function that carries its own "KEEP IN SYNC" warning. And `child_index` / `parent_node_id` / `fan_out_item_hash` are stringly-typed literals across `executor.ts` and `dag-executor.ts` with no shared constant, on untyped JSONB where a typo silently no-ops.

**Two comments that read as present-tense but describe a prior revision** — `dag-executor.ts:5874-5876` says the #2180 and items-resolution paths "did not" notify (they do now; it is justifying why the helper was added), and the step-7 comment at `:6188-6195` says the gate cancel "happens only AFTER every sibling has settled" — untrue of the cancel that actually fires first, at `:6151` mid-flight. The double-cancel itself is deliberate and idempotent.

**`.field` access on a fan-out aggregate always fails, with a misleading message** — `output-ref.ts:141-145`. `asPlainObject` rejects arrays, so `$fanOutNode.output.anything` throws `unparseable` and tells the author to "emit JSON containing 'field'" — but there is no producer prompt to change; the array shape is fixed by the engine. The fail-loud behavior is *correct* and valuable (no `{error, status}` entry is ever silently consumed as data). Only the message needs a case pointing at the documented script-node pattern.

**Test gaps worth one follow-up commit** (from the coverage pass; none blocking): resume + shared-checkout preflight with partial completion is untested and would silently reintroduce the `items.length`-vs-`pendingCount` bug the branch already fixed once; the token roll-up is executed but never asserted (one line to add to the existing cost test); `items` *grew* on resume, gaps in indices, and duplicate `child_index` are untested. Ordering is proven properly in the pool's unit test via inverse delays, but the executor-level ordering test uses instant children, so it would pass on a completion-ordered implementation.

**No ceiling on tree width** — `max_parallel` has `.min(1)` but no `.max()`, `items` is unbounded, and nested fan-out is allowed to depth 5, so worst case is `max_parallel⁴` ≈ 625 concurrent runs at the default. Tracked as #1961; a `.max()` is a cheap partial guard.

---

## Strengths

- **`mapWithLimit` is the cleanest part of the diff.** Its three properties are structural rather than incidental: concurrency is bounded by construction (N worker closures, not an in-flight counter that could race); ordering is preserved by index claim (`cursor++` is atomic with no `await` between read and increment), so completion order cannot affect slot placement; and both async rejections and synchronous throws land in the same `catch`, so no worker promise ever rejects and no unhandled-rejection hazard exists. Its unit tests pin all three, including a real max-watermark counter and an inverse-delay ordering case.
- **The comments document reversals honestly.** "An earlier revision fail-fasted here… That saved spend by deciding one child's fate from another's, which is not the engine's call to make." That is rare and genuinely useful. `fan_out_sibling` being retained as read-only legacy, with an explicit note on why deleting it would wedge runs in flight across the upgrade, is good discipline.
- **The design talked itself out of four features on principle** — racing, fail-fast, `all_success`-as-default, and inferred isolation — and wrote down why, then generalized it into a constitution rule rather than four ad-hoc rejections. The rule is backed by code in three of the four places it makes claims.
- **The collision preflight counts indices this attempt will actually drive**, not `items.length`, so a resume with one instance left is never wrongly blocked — a subtle case that is easy to get wrong.
- **Resume idempotency is proven concretely**, via `completed_at` equality before and after resume (a row count alone could not show a completed child was threaded rather than re-executed), and the in-memory test double faithfully mirrors the real `jsonMerge` metadata semantics rather than a more forgiving replace.
- **Items resolution is genuinely fail-closed**, and "empty" is correctly a *separate* valid path from "malformed".
- **The 266 lines of new documentation hold up under line-by-line checking** — zero factual errors found against the implementation. Defaults match the schema; the `all_success`/`all_done` disambiguation matches the join-reduce code; the collision message in the guide is a verbatim match of `fanOutSharedCheckoutMessage()` including all three remedies; the subtle timing claim "it happens as soon as the pause is seen rather than at the end" is true at `:6151`; both example YAML blocks parse against `fanOutConfigSchema` with `items` producers correctly declared in `depends_on`; and all six issue/PR links resolve to the right topics. `max_parallel` bounding concurrency rather than count or spend — including the 500-run cascade caveat — is documented in the exact section a user configuring `fan_out:` would already be reading.

---

## Agent claims I checked and corrected

Recording these because they shaped the ranking:

- **"The implementation violates the independence rule this PR adds"** (raised as HIGH). Deflated. A paused child does fail the node regardless of `join` — but `authoring-workflows.md:1313-1338` documents this at length and calls it "the intended shape, not a missing feature", including "the single case where a fan-out cancels a run it did not have to." The claim that neither the docs nor the constitution name the exception is wrong about the docs. What survives is minor: the constitution's new rule section doesn't carve out the exception the authoring guide does.
- **"`all_success` fail-fast doesn't cancel in-flight siblings"** — not applicable. Fail-fast was removed by design; there is nothing to clean up.
- **"A downstream node could silently consume an `{error, status}` entry as data"** — does not exist. Array field-access fails loudly (see above).

Wirasm:
## Review addressed — `766223fa`

Thanks for a careful review; the three self-corrections at the end made it easy to trust the ranking. Everything is folded in except two items I'm declining with reasons, and one place where I think the review is right and a subsequent correction to it was wrong.

`bun run validate` exit 0, grepped: zero `(fail)` lines, `Installer tests passed`, no non-zero exits.

### Must-fix

**1 · `fan_out_orphan` added to the recoverable set.** One line plus a comment saying why every member of the type belongs there. Test: an index cancelled `fan_out_orphan` on an earlier attempt is re-driven when the item returns. Fails before the fix.

**2 · `tokens` threaded through `writeCompleted`.** You're right that this is bigger than its ranking. I'd add one thing to the reasoning: it isn't only that failure is now normal — it's that `getDagResumeSnapshot` sums `data.tokens` and never reads `cost_usd`, so tokens are the axis that survives resume, and fan-out was the single node type that dropped them. Asserted on the existing cost test (`cost_usd` *and* `tokens` on the event).

On Σ-completed-not-Σ-children: docblock corrected on both helpers, PR body corrected, and the upstream fix named (`failWorkflowRun` would have to persist usage the way `completeWorkflowRun` does).

**3 · PR body rewritten** — nine replacements across the summary, scope boundary, diagram, file list, security answer, verified-scenarios line, risks and mitigation. `grep` for the four stale claims now returns 0. Agreed on the severity: the body argued for inferring `isolation` from `fan_out`, which is what the constitution rule this PR adds forbids.

**4 · Types regenerated** — `@default all_done`. The regeneration also picked up `event_order` on the workflow-event schema, drifted from `dev` independently of this PR.

### Where I think the review is right and a correction to it was wrong

The docs-page finding cites `reference/cli.md:369`. That was flagged to me as unverifiable — zero grep hits. It's there:

```
cli.md:369  …the parent auto-resumes when the child completes.
```

The negative grep was phrasing-specific: `cli.md` says *"the parent run pauses 'blocked on child'"* where `quick-reference.md` says *"pauses the whole tree"*. All three pages have the paragraph and all three now carry the fan-out carve-out. `quick-reference.md` also gains the `fan_out` row its table was missing, and `CLAUDE.md:442` gains a fan_out clause.

### Declined, with reasons

**The `childArgs` rebuild.** Two occurrences, on a surface still moving — Rule of Three isn't met, and #2223 deferred its own near-identical finding for the same reason. Leaving it.

**`max_parallel.max()`.** 625 concurrent runs is a real hazard, but #1961 exists to decide the actual bound, and a number picked today is a constant that's right this month and a trap next month. Deferring to #1961 rather than guessing.

I did split the duplication finding rather than take it as one item: the stringly-typed `child_index` / `parent_node_id` / `fan_out_item_hash` literals are a **type-safety** problem, not a DRY one — untyped JSONB where a typo silently no-ops at either end. They're now shared constants with a typed reader, so writer and reader agree by symbol. I also took the `loader.ts` regex duplicate, since it sits inside the function carrying its own KEEP IN SYNC warning; hoisted to one source with each call site building its own RegExp (a shared `g` regex would share `lastIndex`).

### The rest

- **Fail-open preflight** — fixed, and failing closed needed a wrinkle worth noting: reporting a *collision* on an unresolvable target would send the author to the wrong file. The resolver now reports why it failed and the preflight fails closed with a message about the resolution, not about `mutates_checkout`.
- **Missing-`child_index` rows** — warn + cancel-if-live, matching the branch below. Test covers the 1:1→fan_out conversion case that produces them. Fails before the fix.
- **Un-notified early-failure branch** — fixed; the stale comment justifying the helper is rewritten.
- **Two present-tense comments** — corrected. The step-7 one now says what's actually true: the cancel that frees the path lock fires mid-flight when the pause is observed, and this pass is the idempotent backstop.
- **`.field` on an aggregate** — new `array-aggregate` reason pointing at the script-node pattern. Fail-loud behaviour unchanged, as you said it should be.
- **Constitution** — the independence rule now carves out the paused child, and I tried to make it precise enough not to be a loophole: nothing about a *sibling* decides it; a pause isn't terminal and a parent has one approval slot, so the cancel is the child's own situation resolving. `### Case law` promoted to `## Case law`, sibling to both rules.
- **Test gaps** — resume + shared-checkout preflight with partial completion added first, since it guards the `items.length`-vs-`pendingCount` inversion this branch already fixed once; reverting to `items.length` fails it. Plus token persistence, orphan recoverability, missing-index. Each verified failing before its fix.

### Not taken

`items` grew on resume, index gaps, duplicate `child_index`, and the executor-level ordering test using instant children — all real, none load-bearing for this merge, and the diff is already wide with #2299 queued behind it on the same file. Worth a follow-up rather than more sprawl here.

PR #1573, opened 2026-05-04
fix(docker): require ARCHON_ALLOW_ROOT_FALLBACK opt-in for root-fallback when chown fails

ztech-gthb:
## Summary

- Problem: post-#1518 `docker-entrypoint.sh` exits with `1` on any `chown` failure of `/.archon` or `/home/appuser`. On macOS VirtioFS bind-mounts this is the normal state — host UID 501 cannot be remapped to container appuser 1001 — so the container crash-loops on every start. Linux-host failures (SELinux/AppArmor, RO mounts) look identical to the macOS case from inside the container, so the entrypoint cannot distinguish them automatically.
- Why it matters: macOS users can't run the stack at all without forking/patching. Pre-#1518 this was handled by silently exporting `IS_SANDBOX=1` (PR #1537), but that bypassed ClaudeProvider's UID-0 safety guard for *any* failure cause — security-adjacent on misconfigured Linux hosts.
- What changed: capture chown stderr (so the warning is actionable) and gate the root-fallback behind an explicit opt-in env var `ARCHON_ALLOW_ROOT_FALLBACK=1`. Without the opt-in, behavior is unchanged from #1518 (fail-loud, exit 1). With the opt-in, container starts as root and exports `IS_SANDBOX=1` to allow ClaudeProvider to skip its UID-0 guard. Both chown blocks (`/.archon` and `/home/appuser`) share the same fallback path so users don't need to debug them separately.
- What did **not** change: the happy-path (chown succeeds → drop to appuser via `gosu`) is unchanged. Linux hosts where chown succeeds see no behavior change. The new env var is opt-in only — never auto-enabled.

### Before (current origin/dev, post-#1518)

```
macOS user: docker compose up -d
  entrypoint runs as root
  chown -Rh appuser:appuser /.archon → fails (VirtioFS, host UID 501)
  echo "ERROR: Failed to fix ownership..." >&2
  exit 1
  → container crash-loops, no usable stack on macOS
```

### After (this PR)

```
macOS user (no env var set yet): docker compose up -d
  chown of /.archon fails, chown of /home/appuser fails
  echo "WARNING: chown failed:" + per-path stderr from chown
  echo "ERROR: refusing to run as root. On macOS VirtioFS this is expected
         — set ARCHON_ALLOW_ROOT_FALLBACK=1 to opt in. On Linux, fix
         volume ownership instead."
  exit 1
  → user reads error, adds ARCHON_ALLOW_ROOT_FALLBACK=1 to their .env or compose env, retries

[...]

Issue #1970, opened 2026-06-12
chown for appuser:appuser takes too long to finish


uinstinct:
## Summary

- What broke: `chown -R appuser:appuser /app` when doing `docker compose up` takes more than 50 minutes to finish. This is a bottleneck for updates.
- When it started (if known):
- Severity: `major`

## Steps to Reproduce

1. `docker compose up` on a fresh linux machine
2. see that it takes 50 minutes for the `chown` step to finish

## Expected vs Actual

- **Expected**:
It should finish in at most 10 minutes

- **Actual**:

<img width="1495" height="517" alt="Image" src="https://github.com/user-attachments/assets/6ae1ceb8-5832-419b-aeca-4d49743f5cb2" />

## User Flow

```
(Draw the flow that triggers the bug. Mark where it breaks with [X].)

Example:
  User                   Archon                   AI Client
  ────                   ──────                   ─────────
  sends /plan ─────────▶ routes to workflow
                         creates worktree
                         streams to AI ──────────▶ processes prompt
                         [X] timeout waiting ◀──── no response
  sees error ◀────────── sends error message
```

## Environment

- Platform: (Slack / Telegram / GitHub / Discord / Web / CLI)
- Database: SQLite
- Running in worktree? (`Yes/No`)
- OS: Ubuntu 24.04

## Logs

```
Paste relevant logs here (redact any tokens/secrets)
```

## Impact

- Affected workflows/commands: docker compose up
- Reproduction rate: Always
- Workaround available? If so, describe:
- Data loss risk? No

## Scope

- Package(s) likely involved: `core|workflows|isolation|git|adapters|server|web|cli|paths`
- Module (if known): e.g. `workflows:executor`, `adapters:slack`

PR #1518, opened 2026-05-01
feat(docker): persist /home/appuser by default + clarify ARCHON_HOME/ARCHON_DATA semantics

leex279:
## Summary

- **Problem:** Two operator-facing UX gaps in Docker. (1) `ARCHON_HOME` and `ARCHON_DATA` leak into the container env via `env_file: .env` but neither is read where operators expect — `ARCHON_HOME` is silently overridden to `/.archon` at `packages/paths/src/archon-paths.ts:56-59`, and `ARCHON_DATA` is a host-side Compose substitution token that no TS source reads. (2) Nothing in the base compose mounts `/home/appuser/`, so all user-specific state — Claude Code skills/commands/agents/hooks/MCP/projects/memory/OAuth, Codex/Pi auth, `~/.gitconfig`, shell history — is wiped on every `docker compose up --build`.
- **Why it matters:** Operators waste time chasing leaked env vars that look configurable but aren't, and lose every customization they make under `~/` on every rebuild. Reported by community member Zolto on Dynamous; cross-posted to #1517.
- **What changed:** Four groups of changes. **(a)** Annotate `.env.example` and `reference/configuration.md` so the leaked vars stop being surprising; entrypoint emits a one-line stderr warning when they're set. **(b)** Persist `/home/appuser` **by default** as a named volume (`archon_user_home`) in both `docker-compose.yml` and `deploy/docker-compose.yml`. **(c)** Expose `ARCHON_USER_HOME` (parallel to `ARCHON_DATA`) so operators can swap the named volume for a host bind-mount via `.env` without editing compose; entrypoint chowns `/home/appuser` to UID 1001 on every start so bind-mount UID drift is handled automatically. **(d)** Flip the Claude provider default `settingSources` from `['project']` to `['project', 'user']` so the persisted `~/.claude/` actually surfaces user-installed skills/commands/agents/hooks; without this, persistence would be a hollow promise. Affects all environments (not just Docker) — operators who want strict project-only scoping can set `assistants.claude.settingSources: ['project']` in `.archon/config.yaml`.
[...]

PR #2228, opened 2026-07-21
feat(docker): opt-in root fallback (ARCHON_ALLOW_ROOT_FALLBACK) for macOS bind mounts

### LINKED ISSUE #2225: feat(docker): opt-in root fallback (ARCHON_ALLOW_ROOT_FALLBACK) for macOS VirtioFS / read-only bind mounts
On macOS VirtioFS (and SELinux / read-only bind mounts), `fix_ownership()` in `docker-entrypoint.sh:14-31` fails because the host UID can't remap to appuser (1001), so the container exits 1 and crash-loops — unrunnable without forking the image. The current fail-loud behavior (post-#1518/#1981) is correct, but there is no escape hatch.

Add an **explicit opt-in** `ARCHON_ALLOW_ROOT_FALLBACK=1`: on chown failure, print the captured stderr and — only when the var is set — continue as root with `export IS_SANDBOX=1` (satisfies the UID-0 guard at `packages/providers/src/claude/provider.ts:1220`); otherwise keep the loud `exit 1`. Default-unset must be byte-for-byte current behavior. Implement by having `fix_ownership()` accumulate failures rather than exit inline, then branch once on the opt-in.

Docs are part of the definition of done (the prior attempt was blocked on exactly this): document the var + the `IS_SANDBOX` side-effect in the docker deployment guide, fix the "Permission errors" troubleshooting entry (chown isn't always sufficient on bind mounts). Security note: explicit opt-in only — never auto-enable.

Prior art: closed PR #1573 (@ztech-gthb) carried the accepted design + review feedback.

Wirasm:
## Summary

- Problem: on macOS VirtioFS (and SELinux / read-only bind mounts), `fix_ownership()` in `docker-entrypoint.sh` always fails — host UIDs cannot be remapped to container appuser (1001) — so the container exits 1 and crash-loops. Archon is unrunnable on macOS bind mounts without forking the image.
- Why it matters: the post-#1518/#1981 fail-loud posture is correct, but it has no escape hatch for hosts where the failure is expected and permanent.
- What changed: `fix_ownership()` accumulates failures (`chown_failed=1`) instead of exiting inline; after both paths run, one branch decides — default (var unset) keeps the loud `exit 1`, while an explicit `ARCHON_ALLOW_ROOT_FALLBACK=1` prints a warning and continues as root with `export IS_SANDBOX=1` (satisfies the UID-0 guard in `packages/providers/src/claude/provider.ts`). Docs: new opt-in section in the docker deployment guide, corrected "Permission errors" troubleshooting entry, and a `.env.example` entry.
- What did **not** change (scope boundary): the happy path (`chown` succeeds → `gosu appuser`) is untouched; the ClaudeProvider guard, compose files, and the isolation-container runner image are untouched. Default-unset output per failing path is byte-identical (same stderr echo, same ERROR line, same exit 1); the one disclosed difference is the compound-failure case — when `/.archon` fails, `/home/appuser` is now also attempted, emitting one extra diagnostic block before the same exit 1 (inherent to the accumulate design; see Side Effects / Blast Radius).

### Before

```
macOS user: docker compose up -d
  entrypoint runs as root
  fix_ownership /.archon → find/chown fails (VirtioFS refuses UID remap)
  prints chown stderr + "ERROR: Failed to fix ownership of /.archon ..."
  exit 1
  → container crash-loops; no way to run Archon on macOS bind mounts
```

### After

```
macOS user (var unset — DEFAULT, unchanged):
  fix_ownership /.archon fails → same stderr + same ERROR line
  fix_ownership /home/appuser also runs → its error printed too      [*only diff: both paths reported]
  exit 1

macOS user (.env: ARCHON_ALLOW_ROOT_FALLBACK=1):
  both errors printed as above
  "WARNING: ARCHON_ALLOW_ROOT_FALLBACK=1 — continuing as root with IS_SANDBOX=1."  [+]
  export IS_SANDBOX=1; RUNNER=""                                                    [+]
  → server starts as root; ClaudeProvider's UID-0 guard passes via IS_SANDBOX=1

Linux user with correct volume ownership: no change whatsoever (chown succeeds, branch not taken).
```

### Before

```
docker-entrypoint.sh
  fix_ownership() ── find + chown fails ──▶ echo stderr + ERROR ──▶ exit 1 (inline)
  fix_ownership /.archon      (exit here means /home/appuser never attempted)
  fix_ownership /home/appuser
  RUNNER="gosu appuser"
```

### After

```
docker-entrypoint.sh
  [~] RUNNER="" hoisted as the single default (non-root branch and opt-in fallback both use it)
  [~] fix_ownership() ── find + chown fails ──▶ echo stderr + ERROR ──▶ chown_failed=1 (accumulate)
  fix_ownership /.archon
  fix_ownership /home/appuser
  [+] if chown_failed=0        ──▶ RUNNER="gosu appuser"          (unchanged happy path)
  [+] elif ARCHON_ALLOW_ROOT_FALLBACK=1 ═══▶ WARNING + export IS_SANDBOX=1 (RUNNER stays "")
  [+] else                     ──▶ exit 1                          (unchanged default)
                                       │
packages/providers/src/claude/provider.ts (unchanged)
  UID-0 guard: getProcessUid()===0 && IS_SANDBOX!=='1' → throw
```

**Connection inventory**:

| `fix_ownership()` failure | inline `exit 1` | **removed** | replaced by `chown_failed=1` accumulator |
| `fix_ownership()` failure | `chown_failed` accumulator | **new** | both paths report before the single branch |
| failure branch | `ARCHON_ALLOW_ROOT_FALLBACK` env var | **new** | explicit opt-in, `:-0` default |
| failure branch (opt-in) | `export IS_SANDBOX=1` (RUNNER keeps hoisted `""` default) | **new** | satisfies ClaudeProvider UID-0 guard |
| failure branch (default) | `exit 1` | **modified** | moved out of the function; same output, same exit code |
| ClaudeProvider UID-0 guard | `IS_SANDBOX` env | unchanged | still required to run as root |
| `docker-compose.yml` `env_file: .env` | entrypoint env | unchanged | how the var reaches the container |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `docs`
- Module: `docker:entrypoint`

## Change Metadata

- Change type: `feature`
- Primary scope: `multi` (docker entrypoint + docs)

## Linked Issue

- Closes #2225
- Supersedes #1573 — carries over the accepted design (capture stderr, accumulate failures, explicit opt-in, `IS_SANDBOX` handshake) rebuilt on the rewritten post-#1970 `fix_ownership()`. Credit: Co-authored-by @ztech-gthb.

Commands and result summary:

```bash
bash -n docker-entrypoint.sh   # syntax OK
bun run validate               # all nine checks green
```

- Evidence provided (test/log/trace/screenshot): `bash -n` clean; full `bun run validate` green (no TypeScript touched). Branch logic exercised in a throwaway local simulation with a stubbed failing chown — four scenarios verified: happy path (`RUNNER="gosu appuser"`, unchanged), one/both paths failing with var unset (byte-identical per-path error lines, exit 1), both failing with `ARCHON_ALLOW_ROOT_FALLBACK=1` (warning printed, `IS_SANDBOX=1` exported, `RUNNER=""`, exit 0 continues).
- If any command is intentionally skipped, explain why: no shell-test harness exists for `docker-entrypoint.sh` in this repo (nothing in tests or CI references it), so per convention verification is manual reasoning + `bash -n`; no new harness invented.

## Security Impact (required)

- New permissions/capabilities? (`Yes`) — but **only behind an explicit opt-in that is never auto-enabled**. With `ARCHON_ALLOW_ROOT_FALLBACK=1` set AND the ownership fix failing, the container keeps running as root and exports `IS_SANDBOX=1`, which bypasses ClaudeProvider's UID-0 refusal — AI subprocesses then run as root inside the container.
- File system access scope changed? (`No`) — same chown targets; only the failure handling differs.
- If any `Yes`, describe risk and mitigation: risk is confined to operators who deliberately set the var (documented in three places with the side-effect spelled out). The default path is unchanged fail-loud `exit 1` — a typo'd var name falls through to it (safe failure mode). Net effect vs today is strictly opt-in; there is no code path that broadens privileges without the operator's explicit action. Docs steer Linux operators to fix volume ownership instead of opting in, so real misconfigurations aren't masked.

## Compatibility / Migration

- Backward compatible? (`Yes`) — default-unset behavior is unchanged (same error output per failing path, same exit 1); happy-path installs see zero difference.
- Config/env changes? (`Yes`) — one new opt-in env var, `ARCHON_ALLOW_ROOT_FALLBACK` (documented in `.env.example` and the docker deployment guide). No action needed unless opting in.
- If yes, exact upgrade steps: macOS bind-mount users add `ARCHON_ALLOW_ROOT_FALLBACK=1` to `.env` (reaches the container via `env_file: .env`).

- Verified scenarios: `bash -n` syntax check; local simulation of the extracted branch logic under `set -e` with a stubbed failing chown covering all four combinations of {0,1,2 failing paths} × {var unset, `=0`, `=1`} — output bytes and exit codes match the spec in #2225.
- Edge cases checked: `set -e` does not abort on the captured failure (it occurs inside the `if !` guard, same shape as current code); `${ARCHON_ALLOW_ROOT_FALLBACK:-0}` is set-u-safe; values other than exactly `1` fall through to `exit 1`; `local err` scoping preserved.
- What was not verified: a real macOS VirtioFS end-to-end container start (no macOS Docker Desktop bind-mount reproduction in this environment) and a real SELinux denial. The failure path shape is identical to the simulated one; the `IS_SANDBOX=1` handshake matches the guard at `packages/providers/src/claude/provider.ts` exactly.

- Affected subsystems/workflows: `docker-entrypoint.sh` only (compose stack startup); docs. No TypeScript, schema, or build changes.
- Potential unintended effects: with var unset and `/.archon` failing, `/home/appuser` is now also attempted before exit — one additional diagnostic error block on stderr (inherent to the accumulate design; strictly more informative, same exit 1). CI/monitoring that detects misconfigured volumes via the crash-loop still sees exit 1.
- Guardrails/monitoring for early detection: the WARNING line names the var and the `IS_SANDBOX=1` side-effect on every fallback boot, so a root-running container is always loud in `docker compose logs`.

## Rollback Plan (required)

- Fast rollback command/path: revert the commit — behavior returns to current dev (fail-loud unconditionally). No data migration.
- Feature flags or config toggles (if any): the env var itself is the toggle; unsetting it restores fail-loud behavior without a code change.
- Observable failure symptoms: if the fallback misbehaves, the container either crash-loops with the printed chown errors (default path) or runs as root announcing the WARNING line — both visible in container logs.

## Risks and Mitigations

- Risk: a Linux operator masks a genuine volume misconfiguration by setting the var instead of fixing ownership.
  - Mitigation: docs (deployment guide caution + troubleshooting entry + `.env.example`) explicitly say the var is the wrong fix on Linux and give the `chown -R 1001:1001` command; the default remains fail-loud.
- Risk: future entrypoint additions bypass the accumulator with a new inline `exit 1`.
  - Mitigation: the comment block above `fix_ownership()` documents the accumulate-then-branch pattern and why it exists.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in root fallback for macOS bind-mount setups when ownership fixes fail via `ARCHON_ALLOW_ROOT_FALLBACK=1`.
  * When enabled, the container continues in a safer sandbox mode (`IS_SANDBOX=1`) instead of stopping immediately.

* **Documentation**
  * Updated the Docker deployment guide with macOS-specific troubleshooting and OS-aware permission guidance for `/.archon/`.
  * Expanded security cautions and recommended alternatives (such as fixing volume ownership on Linux).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -300,6 +300,15 @@ GITEA_ALLOWED_USERS=
 # NOT read by Archon source code.
 # ARCHON_USER_HOME=/opt/archon-user-home
 
+# Docker root fallback (opt-in escape hatch for macOS bind mounts).
+# On macOS VirtioFS bind mounts the entrypoint's ownership fix always fails
+# (host UIDs can't be remapped to container UID 1001), so the container exits 1.
+# Set to 1 to continue running as root instead. Side-effect: exports IS_SANDBOX=1,
+# which bypasses the Claude provider's UID-0 safety guard — AI subprocesses run
+# as root inside the container. Never auto-enabled; default (unset) fails loud.
+# On Linux, fix volume ownership instead: sudo chown -R 1001:1001 <path>
+# ARCHON_ALLOW_ROOT_FALLBACK=1
+
 # Logging (optional)
 # Set log level: fatal | error | warn | info | debug | trace
 # Default: info
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,7 +6,10 @@ set -e
 # which causes the Claude subprocess to fail silently when spawned with a missing cwd.
 mkdir -p /.archon/workspaces /.archon/worktrees
 
-# Determine if we need to use gosu for privilege dropping
+# Determine if we need to use gosu for privilege dropping.
+# Default: run commands as-is — already non-root (e.g., --user flag or
+# Kubernetes), or root via the explicit ARCHON_ALLOW_ROOT_FALLBACK opt-in below.
+RUNNER=""
 if [ "$(id -u)" = "0" ]; then
   # A blanket `chown -R` rewrites metadata for every inode (#1970); only files
   # with wrong ownership are touched.
@@ -45,16 +48,11 @@ if [ "$(id -u)" = "0" ]; then
     # bypassPermissions as root. Never auto-enabled — default stays fail-loud.
     echo "WARNING: ARCHON_ALLOW_ROOT_FALLBACK=1 — continuing as root with IS_SANDBOX=1." >&2
     export IS_SANDBOX=1
-    RUNNER=""
   else
-    # Fail loud (default). On macOS VirtioFS this failure is expected — set
-    # ARCHON_ALLOW_ROOT_FALLBACK=1 to opt in to running as root. On Linux,
-    # fix volume ownership instead (see the docker deployment guide).
+    # Fail loud (default) — see the docker deployment guide for the
+    # ARCHON_ALLOW_ROOT_FALLBACK opt-in.
     exit 1
   fi
-else
-  # Already running as non-root (e.g., --user flag or Kubernetes)
-  RUNNER=""
 fi
 
 # Warn if vars known to be ignored inside the container were set via env_file: .env.
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -11,6 +11,14 @@ if [ "$(id -u)" = "0" ]; then
   # A blanket `chown -R` rewrites metadata for every inode (#1970); only files
   # with wrong ownership are touched.
   # find + chown -h leaves symlinks un-dereferenced (no-dereference by design).
+  #
+  # chown can fail when the host controls ownership: on macOS VirtioFS bind
+  # mounts, host UIDs cannot be remapped to appuser (1001) at all. On Linux,
+  # SELinux/AppArmor denials or read-only mounts produce the same failure and
+  # look identical from inside the container, so we cannot auto-distinguish.
+  # Failures are accumulated (not exited inline) so we can branch once below
+  # on the explicit ARCHON_ALLOW_ROOT_FALLBACK opt-in.
+  chown_failed=0
   fix_ownership() {
     local err
     # -o is correct: we want appuser:appuser on both, not "either matches".
@@ -19,7 +27,7 @@ if [ "$(id -u)" = "0" ]; then
         echo "$err" >&2
       fi
       echo "ERROR: Failed to fix ownership of $1 — volume may be read-only or mounted with incompatible options" >&2
-      exit 1
+      chown_failed=1
     fi
   }
   fix_ownership /.archon
@@ -28,7 +36,22 @@ if [ "$(id -u)" = "0" ]; then
   # and other user-specific state survive rebuilds. On bind mounts, host UIDs
   # don't map to appuser (1001), so fix ownership via fix_ownership as well.
   fix_ownership /home/appuser
-  RUNNER="gosu appuser"
+  if [ "$chown_failed" = "0" ]; then
+    RUNNER="gosu appuser"
+  elif [ "${ARCHON_ALLOW_ROOT_FALLBACK:-0}" = "1" ]; then
+    # Explicit opt-in (macOS VirtioFS escape hatch): continue as root.
+    # IS_SANDBOX=1 satisfies ClaudeProvider's UID-0 guard
+    # (packages/providers/src/claude/provider.ts), which otherwise refuses
+    # bypassPermissions as root. Never auto-enabled — default stays fail-loud.
+    echo "WARNING: ARCHON_ALLOW_ROOT_FALLBACK=1 — continuing as root with IS_SANDBOX=1." >&2
+    export IS_SANDBOX=1
+    RUNNER=""
+  else
+    # Fail loud (default). On macOS VirtioFS this failure is expected — set
+    # ARCHON_ALLOW_ROOT_FALLBACK=1 to opt in to running as root. On Linux,
+    # fix volume ownership instead (see the docker deployment guide).
+    exit 1
+  fi
 else
   # Already running as non-root (e.g., --user flag or Kubernetes)
   RUNNER=""
--- a/packages/docs-web/src/content/docs/deployment/docker.md
+++ b/packages/docs-web/src/content/docs/deployment/docker.md
@@ -759,7 +759,7 @@ The container runs as `appuser` (UID 1001). The entrypoint tries to fix ownershi
 sudo chown -R 1001:1001 /path/to/archon-data
 ```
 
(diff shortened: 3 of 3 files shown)


Wirasm:
## Self-Review

### Summary

Implementation matches the #2225 spec exactly: `fix_ownership()` accumulates failures (`chown_failed=1`) instead of exiting inline, and a single branch after both calls decides between the unchanged fail-loud `exit 1` and the explicit `ARCHON_ALLOW_ROOT_FALLBACK=1` root fallback with `export IS_SANDBOX=1`.

### Checks performed

- **Spec conformance**: default-unset emits byte-identical per-path output (same stderr echo, same ERROR line) and exits 1; only `"1"` exactly takes the fallback (`${ARCHON_ALLOW_ROOT_FALLBACK:-0}` — unset, empty, `0`, `true`, etc. all fall through to `exit 1`). `IS_SANDBOX=1` is exported before `RUNNER=""`, satisfying the ClaudeProvider UID-0 guard (`getProcessUid() === 0 && process.env.IS_SANDBOX !== '1'`).
- **`set -e` safety**: the captured `find`/`chown` failure occurs inside the `if !` guard (same shape as current code), so errexit is not triggered; `chown_failed` is a global mutated from the function body, initialized before both calls.
- **Behavior simulation**: extracted branch logic run locally under `set -e` with a stubbed failing chown — 4 scenarios (happy path / one fails / both fail x var unset / var=1) all produced the expected bytes and exit codes.
- **Syntax**: `bash -n docker-entrypoint.sh` clean.
- **Docs**: `docker.md` subsection + troubleshooting rewrite + `.env.example` entry all match actual behavior; the `#root-fallback-for-macos-bind-mounts-opt-in` anchor matches Starlight's slug for the new heading.
- **Validation**: full `bun run validate` green locally (no TypeScript touched).

- A future entrypoint addition could bypass the accumulator with a new inline `exit 1` — the comment block above `fix_ownership()` documents the accumulate-then-branch pattern to guard against this.

### Security

- Default path is unchanged fail-loud; no code path broadens privileges without the operator explicitly setting the var. The `IS_SANDBOX=1` side-effect is documented in all three doc surfaces.

### Checklist

- [x] Matches the accepted design from #2225 / prior art #1573
- [x] Default-unset behavior unchanged (per-path output byte-identical, exit 1)
- [x] Opt-in path exports IS_SANDBOX=1 before running as root
- [x] Docs complete (deployment guide, troubleshooting, .env.example)

---
_Self-reviewed by Claude — ready for human review_

PR #2250, opened 2026-07-22
feat(workflows): first_success racing join for fan-out sub-runs (slice-2 PR-D)

Wirasm:
## Summary

- Problem: fan-out (PR #2224) ships `join: all_success | all_done` only — `first_success` racing (the #1764 orchestrator-worker headline: N children race, fastest valid solution wins, losers stop burning) was staged as a load-time rejection.
- Why it matters: racing is the composition that makes swarm patterns economical — without loser cancellation, N-1 children run to completion for output that gets discarded.
- What changed: `first_success` flipped to functional. First child reaching schema-valid terminal success wins; its single output threads as `$<id>.output` (field-accessible under `output_format`). Losers get two-layer cancellation per the plan's D6: the tagged cooperative `cancelWorkflowRun` cascade (`fan_out_race_loser`) as the correctness floor, plus best-effort per-child `AbortController.abort()` reusing the existing `abortSignal` plumbing threaded into all three providers (Codex near-instant, Claude ~2-7s staged, Pi instant) — no new kill primitive; loser burn bounded to ≤ one in-flight turn + grace. Resume: an existing winner threads without respawning anyone; race losers are recoverable only when no winner survives; all-fail races fail the node listing per-child status. #2180 pause handling and cost/token roll-up preserved.
- What did **not** change (scope boundary): no console/docs-site surfaces (PR-E), no gate-queue (#2180), no global budget caps (#1961), no hard-kill primitive. PR-A/PR-C files touched only where the racing join required.

### Before

```
  fan_out: {join: first_success} ──▶ load error: "first_success is not yet supported (PR-D)"
```

### After

```
  Author                       Engine
  ──────                       ──────
  fan_out over 3 items ──────▶ spawns ≤ max_parallel children (own worktrees)
  join: first_success          child 2 completes valid ──▶ WINNER: output threads
                               child 1,3 ──▶ [cancelWorkflowRun tagged fan_out_race_loser]
                                             [AbortController.abort() — fast stop]
  $node.output = winner's output; node completes
```

### Before / After

```
dag-executor.ts fan-out section:
  [~] join reducer gains the first-success race path (winner detection, loser seal)
  [~] per-child AbortController registry (new, racing only)
executor.ts:
  [~] ExecuteWorkflowOptions.abortSignal + RunChildWorkflowArgs.abortSignal threading
schemas/dag-node.ts:
  [~] first_success superRefine rejection lifted
```

| fan-out executor | child AbortController | **new** | racing only; losers aborted best-effort |
| executor abortSignal | provider sendQuery abortSignal | unchanged | pre-existing plumbing, now exercised by racing |
| loser cancel | cancelled_reason tagging (PR-C) | **modified** | new `fan_out_race_loser` reason, same recovery machinery |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `workflows`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Related #2121 (Phase 2, slice-2 plan PR-D), #1764 (racing design), #2180 (gate handling), #977
- Depends on #2224 (stacked — base feat/subrun-fanout)

```bash
bun run validate   # exit 0 — all generated checks, type-check ×10, lint --max-warnings 0, format, all package tests
cd packages/workflows && bun test src/subrun.test.ts   # 37 pass / 0 fail (32 → 37 this PR)
```

- Evidence provided (test/log/trace/screenshot): acceptance tests cover — race of 3: first valid completion wins + output threads + losers cancelled with `fan_out_race_loser` + abort invoked; resume-after-win threads winner without respawn; all-fail race fails with per-child statuses; loser-recovery only when no winner. Also fixes a pre-existing fan-out test flake (see test commit).
- If any command is intentionally skipped, explain why: stacked-PR CI (CodeRabbit + test workflow) skips non-dev bases; full CI runs at retarget after #2224 merges — same as the rest of the stack.

## Security Impact (required)

- If any `Yes`, describe risk and mitigation: n/a

## Compatibility / Migration

- Backward compatible? Yes — `first_success` was previously a load error; existing workflows unaffected
- Config/env changes? No
- Database migration needed? No (reuses PR-C's cancelled_reason metadata)

- Verified scenarios: full validate + 37-test subrun suite locally (delivery was recovered after the implementing agent hit a spend limit — commits were complete and test-green; push/PR performed by the coordinating session)
- Edge cases checked: winner-exists resume semantics; loser recoverability gating; abort best-effort (cooperative cascade remains the correctness floor)
- What was not verified: real-provider abort latency (documented per-provider from the plan's verification; e2e uses the test harness)

- Affected subsystems/workflows: fan-out workflow: nodes opting into first_success only
- Potential unintended effects: abort arriving mid-provider-turn surfaces as a cancelled child — tagged and recoverable per the C2 machinery; cooperative floor bounds any abort miss
- Guardrails/monitoring for early detection: per-child cancelled_reason in run metadata; race outcome logged with winner index + per-loser status

## Rollback Plan (required)

- Fast rollback command/path: revert the two commits — first_success returns to a staged load rejection
- Feature flags or config toggles (if any): none (opt-in via join: first_success)
- Observable failure symptoms: races never conclude / losers keep running → check cancelled_reason stamps + abort logs

## Risks and Mitigations

- Risk: winner declared on a child whose output later proves unusable downstream.
  - Mitigation: winner requires schema-valid output when output_format is declared — same validation contract as every node.
- Risk: abort races the child's own completion write.
  - Mitigation: cooperative cancel is the source of truth; abort is best-effort acceleration only; CAS-guarded status transitions already handle the race.


Wirasm:
Reviewed against `dev`. The `AbortSignal` threading is a genuinely good addition beyond the racing join itself — a per-node abort primitive is worth having on its own.

## Blocking

**1. Paused loser discards the winner — the real bug.** At `dag-executor.ts:5856-5867` the paused check unconditionally fails the node **before** winner selection at `:5871`. So whether the node succeeds depends on whether a loser reached its `approval:` node before the winner's abort landed — a nondeterministic outcome for a fixed workflow and fixed inputs, and it throws away a valid winner plus the node's entire spend.

Move winner determination ahead of the paused check: if a winner exists, treat paused children as losers and cancel them (`fan_out_gate` is still the honest tag); only when there is **no** winner should a paused child produce the gate-rejection failure.

If the current behavior is actually intended — a gate inside a racing child is unsupported, full stop — then say so in the `first_success` doc comment in `schemas/dag-node.ts`: *"a racing child that pauses at a gate fails the node, discarding any winner."* Nothing currently tells an author that, and they will not guess it.

**2. The grandchild-abort claim is false as written.** `executor.ts:687-688` asserts the abort reaches the whole sub-run subtree, but `executeWorkflowNode`'s `childArgs` (`dag-executor.ts:5217-5228`) omits `abortSignal` — so a nested sub-run runs to completion. Either thread `ctx.abortSignal` into that arg bag (one line, and it makes the claim true for the nested orchestrator-worker shape #1764 is actually about) or rewrite the comment to say the abort stops the loser's *own* AI turns only. A comment that overstates a cost-control mechanism is worse than no comment — it stops the next person from checking.

**3. Close the seal→spawn TOCTOU.** `:5780` checks `raceSealed`, then `:5792-5799` awaits `updateWorkflowRun` on the recoverable-cancel re-drive path, then `:5802-5803` creates and registers a fresh `AbortController`. If `sealRace()` runs during that await, this index spawns a child registered *after* the seal, which therefore never gets aborted — and the parent blocks on it at `:5762`. `if (raceSealed) ac.abort();` immediately after `liveAbort.set(i, ac)` closes it. Narrow window (resume-with-recoverable-losers only), one line, and it directly undermines the PR's stated burn bound.

**4. Two claims in the PR body need softening so the merge record isn't wrong.** The bound is "≤ one in-flight **AI** turn" — bash and script nodes are not abortable (`executeBashNode:2435`, `executeScriptNode:2676` take no signal). And the parent does *not* proceed on the win; it awaits every child's settle at `:5762`. Both are defensible designs. The description just shouldn't claim otherwise.

**5. Re-run full CI after retargeting to `dev`.** This PR has **zero** executed checks: CodeRabbit reported "review skipped (non-default base)" and no test workflow ran. Your local `bun run validate` is currently the only evidence. Don't merge on the strength of a CodeRabbit status that is a skip notice wearing a green tick.

## Ordering

#2223 → #2224 → #2250. #2223's identifier-collision bug (isolated children of one parent sharing a worktree path) is unfixed here too and matters more in a racing context, where every loser is a concurrent sibling by construction.

Wirasm:
**Docs are missing.** This lifts the load-time rejection of `fan_out.join: first_success` and adds racing, and touches nothing under `packages/docs-web/`. Per CLAUDE.md, a YAML-surface feature needs documentation and a constitution citation.

Racing especially needs it — an author has to know up front what happens to the losers (cancelled, and under which tag), what `$<id>.output` holds when one child wins, and how a resume behaves after a race was already decided. That is not guessable from the field name.

See also https://github.com/coleam00/Archon/pull/2224#issuecomment-5179333018 — the implicit `isolation: worktree` default on fan-out nodes needs to be removed in #2224, which this branch sits on top of.

Wirasm:
**On hold pending #2224 — 2026-08-04.**

This PR sits on top of #2224, which is being reshaped right now, so it should not be reviewed against its current diff.

What changes underneath it:

- The implicit `isolation: 'worktree'` default that #2224 applies to `fan_out` nodes is being removed. The engine never infers isolation — a child gets a worktree only when the author writes `isolation: worktree`. Racing inherits that.
- #2224 gains a check that fires before children spawn when concurrent siblings would collide on the shared-checkout path lock. Racing spawns concurrent siblings by definition, so it inherits that too.
- #2223 landed a fix to the child worktree identifier (`b8f1d82e`) so distinct nodes and fan-out siblings get distinct checkouts. Do not redesign it here.

What this PR still needs on its own account: **documentation**. It lifts a load-time rejection and adds racing semantics, and touches nothing under `packages/docs-web/`. An author needs to know what happens to the losers (cancelled, tagged `fan_out_race_loser`, and deliberately excluded from the recoverable set — so a resume does not re-drive them while a winner exists), what `$<id>.output` holds when one child wins (a single string, not the array the `all_*` joins produce), and how a resume behaves after a race was already decided. None of that is inferable from the field name.

One product question worth answering before this merges rather than after: racing is most compelling when the children genuinely differ, and today every child of a fan-out node runs the **same static workflow name** — the only variation is the per-item input. Racing N near-identical children over near-identical input mostly buys variance. Worth being explicit in the docs about what the feature is for, so authors reach for it in the cases where it pays.

Wirasm:
## Rebased onto `feat/subrun-fanout@bb69eebe` + racing docs (`26eb6985`)

Documentation only — no racing behaviour changed. `bun run validate` exit 0, grepped: zero `(fail)` lines, `Installer tests passed`, no non-zero exits, all five generated-file checks OK. Three consecutive `bun --filter @archon/workflows test` batches clean; 41 subrun e2e cases including all 5 racing ones. Docs site builds and both new anchors resolve in the rendered HTML.

### The rebase

The base moved four times today. Two conflicts were real and both were the same shape — this PR and the fan-out work each appended a block at the same point:

- **`dag-executor.ts`** — the shared-checkout preflight (#2224) and the `first_success` block landed at the identical spot in `executeFanOutWorkflowNode`. Resolved with the preflight **first**, then the race, and the race's docstring now says so: racing children are concurrent siblings like any others, so they collide on the path lock the same way. Steps renumbered 5–9.
- **`subrun.test.ts`** — my causal-child test and this PR's racing block, both appended before the describe close. Both kept.
- **`dag-node.ts`** — this PR removes the `first_success` load reject; #2224 added an `as` reject a few lines away. Both resolved: racing's comment replaces the guard it retired, the `as` guard stays.

One thing worth flagging: **you and I independently fixed the same flake.** Your `2c66b5ea` and my `21521adf` both set `fan_out-c2-usercancel` to `max_parallel: 1` for the same reason — the fail-fast's loser is tagged `fan_out_sibling`, which is recoverable, so a user-cancel landing on top of it gets resurrected on resume. I kept my version (already on the base, and it also asserts the precondition) and dropped the duplicate hunk.

### One behavioural interaction, found by running it

All five racing tests failed after the rebase. Cause: **the #2224 shared-checkout guard fires on racing**, and correctly — `race-child` declares no `mutates_checkout: false`, the node has no `isolation:`, and `max_parallel: 3` means three concurrent siblings on one checkout:

```text
fan_out node 'work': up to 3 children of 'race-child' would run at once in the parent
checkout, and that workflow does not declare `mutates_checkout: false`. …
```

Fixed in the fixture, not the guard — `race-child` is a single `prompt:` node, so `mutates_checkout: false` is the honest declaration. But it sharpens a real constraint that is now documented: **racing is inherently concurrent, so `max_parallel: 1` is not one of the three ways out of the collision for a race.** Only `mutates_checkout: false` or per-child worktrees are.

### What the docs now say

New section covering everything an author cannot currently learn without reading source: `$<id>.output` is a **single string** and not the `all_*` array (the output shape changes with the join); winner means *first schema-valid* success, so a child can finish first and still lose and a race can fail with every child completed; losers are hard-aborted through the provider abort signal then tagged `fan_out_race_loser`, and that tag is **excluded from the re-driven set** — written as the correctness decision it is, with the exclusion lifting only when no winner survives; the resume short-circuit on a prior winner; and cost being the sum across all racers.

I also corrected the resume bullet in the fan-out section that claimed Archon always re-drives its own cancels — it now names the race-loser exception, so the two sections don't contradict.

### The judgement you asked for

I wrote it narrower than the feature looks, because I think that is the honest answer.

**Every child of a fan-out node runs the same workflow name.** There is no per-child model, provider, prompt, or tool config — the only thing that varies between racers is the item string that becomes `$ARGUMENTS`. So racing N children over near-identical input is one strategy sampled N times, not N strategies in parallel. Two cases where it earns its cost:

1. **The child is unreliable in a way a schema can detect.** With `output_format`, racing is parallel retry — start three, take the first that validates, instead of one attempt and a re-run after a validation failure.
2. **The item list encodes genuinely different approaches** (`["use the existing helper", "write it from scratch", "port the pattern from parser.ts"]`). The diversity is real because it travels in `$ARGUMENTS` and reaches the child's prompt. This is the case racing is *for*, and it argues for writing the items deliberately rather than racing over whatever list an upstream node happened to produce.

And explicitly not worth it when the racers differ only incidentally, when the child is already reliable (N× cost for a result you'd have got anyway), or when you want **different models** to compete — which is a reasonable thing to want and which fan-out cannot express today. If that is where racing is eventually headed, it needs per-child provider/model variation, and that is a separate feature rather than a docs gap.

### Constitution

Racing recorded as shipped, with the admissibility argument: the engine sees only *which* outcome to keep, the selection is temporal rather than authored, and validity reuses the node's already-declared `output_format` instead of adding any expression surface.

Wirasm:
### Guard message is now race-aware (`32c56fd2`)

You were right that I stopped a step short. The guard's third option was being offered unconditionally, including to the one join where it is destructive.

`bun run validate` exit 0, grepped: zero `(fail)` lines, `Installer tests passed`, no non-zero exits. 42 subrun e2e cases.

**Before** — what a racing node's author was told, verbatim from the failing test:

```text
fan_out node 'work': up to 3 children of 'race-child-writes' would run at once in the parent
checkout, and that workflow does not declare `mutates_checkout: false`. … Choose one: add
`mutates_checkout: false` to 'race-child-writes' if it only reads the repo; set
`isolation: worktree` on 'work' if the children write to it; or set `fan_out.max_parallel: 1`
to run them one at a time.
```

Follow that third option and the run goes green with the race switched off — one child, accepted if it validates. Exactly the shape of the implicit isolation default: a plausible answer where an honest one was needed.

**After**, on a `first_success` node:

```text
… Choose one: add `mutates_checkout: false` to 'race-child-writes' if it only reads the repo;
or set `isolation: worktree` on 'work' if the children write to it. `fan_out.max_parallel: 1`
is NOT a fix for a race — a window of one runs a single child and accepts it, turning the
race into a plain sub-run without saying so.
```

I said it explicitly rather than silently dropping the option, for the reason you gave: the author is reading this while blocked, and should not have to already know that racing implies concurrency in order to work out why an option vanished. The non-racing message is unchanged.

**Tests.** Two, deliberately asymmetric:

- *the shared-checkout guard does NOT offer max_parallel: 1 on a racing node* — asserts both real outs are present, the serialise sentence is absent, and the explanation is present. Reverting the join check fails it with the misleading message quoted above.
- The existing non-racing guard test gained the inverse assertions: serialising **is** offered, and the race explanation is **absent**. This one passes before and after by design — it is not a discriminator, it is the guard against over-correcting into dropping the option everywhere. Both states verified by running them.

**Docs.** The three-ways-out table marks the serial row *"Not available on a race"* and links to the racing section; the racing section's isolation note now says the guard handles this, so an author does not have to carry the rule in their head at the moment it bites.

### One thing I checked and could not close

Your "no new side effect either way" hint pointed at something real. I disabled the path-(a) resume short-circuit (`if (false && priorWinner)`) and **the path-(a) test still passes** — path (b) reaches an identical end state by another route, so every observable it asserts is the same with and without the optimisation.

I tried to measure spend instead (winner moved below a recoverable loser, counting provider calls) and got 0 either way, but that result is confounded: the seeded children have `working_path`s that don't exist, so a re-drive dies at the "working path no longer exists" guard before reaching a provider. So: the test does not discriminate (verified by running it); whether the short-circuit is load-bearing in production is *reasoned* — by reading, it prevents re-driving recoverable losers below the winner — and not verified. Left alone; it needs a fixture with real working paths and is pre-existing coverage.

Third variant of the same trap, worth stating alongside the other two: **an optimisation whose only effect is "don't do redundant work" is invisible to any test that asserts end state.** The observable has to be work performed, and the fixture has to let that work actually happen.

Wirasm:
**On hold — racing is contrary to a semantics decision taken 2026-08-04.**

> Each child should run to completion or fail before any nodes below a fan-out can continue. They should not race each other.

`first_success` is the winner cancelling the losers by construction — per-child `AbortController`s wired into `SendQueryOptions.abortSignal` exist specifically so the losers stop burning tokens once a winner is found. There is no version of racing that satisfies that requirement, so this is not a matter of reshaping the PR.

The work itself is complete, green and documented, and the branch stays as-is rather than being reverted — the disposition is the maintainer's call.

**What changes in #2224 under the same decision:** the `all_success` fail-fast goes. Today the first failed child sets `stopSpawning` and cancels in-flight siblings tagged `fan_out_sibling`; every index will now spawn and every child will reach its own terminal state before the join reduces. `all_success` keeps its outcome — any failed child fails the node — it just stops killing the others to get there. That trade is predictability for spend: a large fan-out whose first child fails now runs every remaining child, which makes #1961's budget ceiling load-bearing rather than theoretical.

**The use case racing was reaching for has a better home.** The docs written here state plainly that racing only pays when children genuinely differ, while every child of a fan-out node runs the same static workflow name with only the item string varying — so racing N children over near-identical input is one strategy sampled N times, not N strategies. #2447 (provider/model on `workflow:` nodes) makes the real thing expressible: N distinct sub-run nodes in one layer, each with its own model, all running to completion, with a downstream node judging. That composes from two small primitives, needs no mutual cancellation, and satisfies the decision above.

Two things from this PR are worth keeping regardless of what happens to the branch: the race-aware collision message (`32c56fd2`), which stops the guard recommending `max_parallel: 1` to an author whose join makes it meaningless; and the finding recorded in #2445 — the resume-after-win test passes with its own short-circuit disabled, because path (b) reaches an identical end state, so end-state assertions cannot discriminate an optimisation whose only effect is avoiding redundant work.

Wirasm:
Closing. Racing is rejected on the merits, not deferred.

## Why

The decision, from the maintainer:

> Each child should run to completion or fail before any nodes below a fan-out can continue. They should not race each other. **The point is not to have a winner.**

`first_success` is the winner aborting and cancelling the losers by construction — `sealRace` plus per-child `AbortController`s wired into `SendQueryOptions.abortSignal` exist precisely so the losers stop burning tokens once a winner appears. There is no version of racing that survives the rule, because racing that does not terminate the losers is not racing.

This is now recorded as constitutional case law (`f07341c3`), including a **reversal** of the earlier ✅ on this feature. The original admission read *"a join rule — coordination"*, which describes its shape and misses what it does. The new **independence rule** is the general form:

> Parallel children are independent by default. Anything that couples their fates is opt-in, and must come from the author's declaration rather than be inferred.

Racing is the strongest coupling available: one child's success ends its siblings mid-flight.

## The want underneath it is real, and has a better home

Racing was reaching for *several genuinely different attempts, best result forward*. The docs written on this branch already state honestly that it cannot deliver that: every child of a fan-out node runs the **same static workflow name**, with no per-child model, provider, prompt or tools, so racing N children over near-identical input is one strategy sampled N times.

The shape that does deliver it: **N distinct nodes in one DAG layer**, each with its own model and its own scope, all running to completion, converging on a collector node that reads every result. No mutual cancellation, no winner, and every output preserved — which is the point. That needs #2447 (provider/model on `workflow:` nodes) when the children are sub-workflows; for `prompt:` children it works today.

## Salvaged, so nothing here is lost

- **The race-aware collision message** (`32c56fd2`) — the shared-checkout guard no longer recommends `max_parallel: 1` to an author whose join makes it meaningless. That reasoning is superseded along with racing, but the underlying discipline (an error message must not offer an escape that silently defeats the author's intent) is worth carrying.
- **#2445** — the resume-after-win test passes with its own short-circuit disabled, because path (b) reaches an identical end state. Generalised: an optimisation whose only effect is avoiding redundant work is invisible to any test asserting end state; the observable has to be work performed, and the fixture has to let that work happen. That finding outlives this PR.
- **The flake trap** documented on #2224 — any fan-out test combining concurrency with a deliberate failure has a timing-dependent loser carrying a recoverable tag.

The branch is left intact rather than deleted, in case any of the above is wanted as a reference.

Thanks for the work — it was correct, green and well-documented. What changed is the model it was built for, and finding that out cost less than shipping it would have.

PR #2253, opened 2026-07-22
fix(security): remediate vulnerable dependencies

DIZ-admin:
## Summary

- Problem: the dependency graph contained vulnerable direct and transitive packages, and the Better Auth upgrade required a small API/type adaptation.
- Why it matters: known vulnerable versions remain reachable through server, web, and docs dependency trees.
- What changed: upgraded direct dependencies, pinned patched transitive versions with root overrides, refreshed `bun.lock`, adapted auth typing/email validation, and added exact Gitleaks fixture fingerprints.
- What did **not** change (scope boundary): no new user-facing feature, protocol, permission, database schema, environment variable, or marketplace entry.

### Before

```text
User request -> Archon server -> Better Auth -> existing session/user flow
```

### After

```text
User request -> Archon server -> [upgraded Better Auth] -> existing session/user flow
                                         |
                                         +-> [reject provider user data with no email]
```

There is no intended UI or workflow change for valid users.

### Before

```text
package manifests -> bun.lock -> server / web / docs runtime dependencies
server auth       -> Better Auth builder -> exported auth instance
```

### After

```text
[~ package manifests] ===> [~ bun.lock] ===> server / web / docs runtime dependencies
[~ server auth]       ===> [~ Better Auth builder] -> exported inferred auth instance
```

**Connection inventory:**

| Root/package manifests | `bun.lock` | **modified** | Patched direct/transitive dependency resolution |
| Server auth module | Better Auth | **modified** | Inferred return type replaces removed/changed exported type |
| Better Auth user data | Archon user creation | **modified** | Missing email is rejected explicitly |
| Runtime modules | External services | unchanged | No new calls or capabilities |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `dependencies`, `server`, `web`, `docs`
- Module: `server:auth`

## Change Metadata

- Change type: `security`
- Primary scope: `multi`

## Linked Issue

- No linked issue.

```bash
IS_SANDBOX=1 CI=1 NODE_OPTIONS=--max-old-space-size=4096 \
  npx --yes --package=bun@1.3.11 -- bun run validate
# exit 0

git show --check HEAD
# exit 0

gitleaks stdin --no-banner --redact --exit-code 1 < <(git show --format= --binary HEAD)
# exit 0; no findings
```

- Evidence provided: complete local validate log, commit hygiene check, and redacted Gitleaks scan.
- Intentionally not run locally: Windows CI and the Docker image build. GitHub's fork workflow is waiting for maintainer approval before those jobs can run.

## Security Impact (required)

- Mitigation details: transitive pins are explicit; Gitleaks exclusions are exact fingerprints rather than path-wide or rule-wide exclusions.

## Compatibility / Migration

- Backward compatible? `Yes` for valid existing auth flows; malformed provider data without an email now fails explicitly.
- Config/env changes? `No`
- Upgrade steps: normal `bun install --frozen-lockfile` from the updated lockfile.

- Verified scenarios: frozen dependency graph validation, type-check, lint, formatting, tests, commit whitespace hygiene, secret scan, and exact local/remote SHA equality.
- Edge cases checked: missing email guard and exact-only Gitleaks allowlist semantics.
- What was not verified: Windows runner behavior and Docker image smoke test; both remain delegated to upstream CI after maintainer approval.

- Affected subsystems/workflows: dependency resolution, server authentication typing/user creation, web/docs build dependencies, secret scanning.
- Potential unintended effects: upstream dependency behavioral changes or an auth provider returning no email.
- Guardrails/monitoring for early detection: full validation suite, frozen lockfile CI, explicit missing-email error, and exact fingerprint allowlisting.

## Rollback Plan (required)

- Fast rollback command/path: `git revert 4aa50d31276e9719ff1d4e23521aa0ad74e7ea10`, then `bun install --frozen-lockfile`.
- Feature flags or config toggles: none.
- Observable failure symptoms: install/lockfile mismatch, type-check/test failures, or auth user creation errors after dependency upgrade.

## Risks and Mitigations

- Risk: patched dependencies may include behavioral changes.
  - Mitigation: pinned versions, frozen lockfile, and complete local validation.
- Risk: provider data without an email is no longer accepted.
  - Mitigation: fail explicitly instead of constructing an invalid user record.
- Risk: a fixture fingerprint may change as source lines move.
  - Mitigation: exact fingerprints fail closed; future scanner runs expose stale or new findings.


Wirasm:
Thanks for this — and thanks for the unusually thorough description, including a rollback commit and an explicit list of what you didn't verify. That's a good-faith submission and I read it as one.

The underlying point is correct: this repo has **no Dependabot and no Renovate**, so nobody has been minding the transitive graph. A lot of what you found is real. `ws` <8.21.0 in particular is a live surface — it's discord.js's gateway socket — and `form-data`, `fast-uri`, `postcss`, `svgo`, `devalue`, `sharp`, `axios`, and `qs` all check out against their advisories.

I can't merge it as filed, and rather than send you a list of patches to a 100-package lock refresh, **I'd like this as one PR per dependency** (or per tightly-coupled group). Reasons, concretely:

- A single lock refresh re-resolves 99 packages and moves 70 tree entries. If anything regresses after merge, it isn't bisectable — I'd be reverting 17 overrides to find one.
- Each override needs its own justification, and three of them don't survive one (below). Separate PRs make that visible instead of buried.
- Individually revertable is the property that matters for dependency changes. That's also exactly what a bot would give us.

Sequencing suggestion: the clean transitive overrides first (they're the ones with live advisories and near-zero blast radius), then the direct workspace bumps one at a time, then Better Auth last and on its own.

## The blocker — please don't ship the undici override in any form

`"undici": "6.27.0"` is a **flat root override**, and `@earendil-works/pi-coding-agent@0.80.6` pins `undici@8.5.0` exactly. Your lock deletes that entry, dragging Pi down two majors.

And it buys nothing: 8.5.0 is **already the patched version** for every undici advisory in play — GHSA-vxpw-j846-p89q, GHSA-g8m3-5g58-fq7m and GHSA-p88m-4jfj-68fv all list `>=8.0.0, <8.5.0` as affected. So the override is pure downside, and it lands on the provider that's becoming our primary runner.

The legitimate target is discord.js's `undici@6.21.3`. That needs a **path-scoped** resolution (`@discordjs/rest/undici`), not a global one. Same class of problem with `sharp: 0.35.3`, which overrides Astro's declared `sharp: ^0.34.0`.

## Three that I don't think are fixes at all

- **`express-rate-limit`** — the only advisory (GHSA-46wh-pxpv-q5gq) covers `>=8.2.0,<8.2.2`. We're on 8.3.1, already past it.
- **`astro` 6.1.3 → 6.4.8** — the open advisories are `<7.0.6` and `>=3.10.0,<7.0.4`. **6.4.8 is still inside both ranges**, so the bump crosses no boundary in either direction.
- **`@astrojs/starlight` 0.38.2 → 0.38.5** — I couldn't find an advisory for it.

Dropping those shrinks the blast radius a lot for no loss of security.

## Also please drop

- **The `hono` override.** `packages/server/package.json` already declares `^4.12.31`; an exact root pin on our own HTTP framework only blocks future patches. And the four hono advisories are hono/jsx, `cx()`, AWS Lambda body-limit, and API Gateway v1 adapter paths — none of which our server uses.
- **`.gitleaksignore`.** We don't run gitleaks anywhere — no workflow, no pre-commit config, no `.gitleaks.toml`. The file is inert, and I'd rather adopt a scanner deliberately than land its allowlist first. For the record I checked and every fingerprint SHA in it is genuinely from our history, so the file itself is sound; it just has nothing to attach to. If you want to propose adopting gitleaks, that's its own PR and I'd read it.
- **The caret→exact conversions.** `qs` goes from `^6.15.1` to exact `6.15.3`; several others are exact too. Please keep carets so patch releases still flow.

## Better Auth needs to be alone

1.3.34 → 1.6.23 is three minors on the module that **is** our `/api/*` access boundary, it changes signup behaviour, `packages/server/src/auth/instance.ts` has no test file at all, and web auth needs Postgres + `BETTER_AUTH_SECRET` so CI cannot exercise it. I want to smoke it manually against a real deployment before it lands, and I can't do that meaningfully when it's bundled with a lock refresh.

Worth noting the advisory that motivates it (GHSA-qq9h-g4jm-xgf3) is a magic-link / pre-account-hijacking issue, and we configure `emailAndPassword` only — no `socialProviders`, no `plugins` — so I don't believe we're exposed. Happy to be corrected if you see a path I'm missing.

## For each PR, please include the advisory line

`package → GHSA id → affected range → why Archon is or isn't exposed`. The current description says "the dependency graph contained vulnerable direct and transitive packages" and names no CVE, which makes it unauditable — I had to reconstruct applicability from the Advisory API myself.

## Two things on our side, not yours

- **CI has never run on this** (`action_required`, fork PR). I'll approve it. Fair warning that green is partial here: `bun run validate` doesn't include `bun run build`, so **nothing in CI exercises the docs-web Astro/sharp bumps** — an incompatibility would first appear as a broken `main` deploy. That's our gap and I'm filing it.
- **We should adopt Renovate.** You shouldn't have had to hand-assemble 17 overrides, and a bot would have delivered these as individually revertable PRs without producing the undici regression. Filing that too.

If you want the fastest path to something merged today: open a single PR for `ws` alone, with its advisory line. I'll take that immediately, and it'll set the template for the rest.

Wirasm:
Following up on the review above, now that a month has passed. I re-verified every claim against current `dev` rather than assuming anything, and the short version is: **you were right about the substance, and the situation is worse than when you filed this.**

`dev` today has 30 open Dependabot alerts (4 critical, 17 high, 4 moderate, 5 low). Almost nothing you identified has been fixed in the intervening month. That is not because the alerts were dismissed — it is because our automation is broken, which I'll come back to.

## Where `dev` stands now, per package

Checked against the GitHub Advisory API and the resolutions in `bun.lock` on `dev`.

**Still vulnerable — your finding holds:**

| package | on `dev` | live advisories | patched in |
|---|---|---|---|
| `ws` | 8.19.0 | GHSA-96hv-2xvq-fx4p (high), GHSA-58qx-3vcg-4xpx | 8.21.0 |
| `better-auth` | 1.3.34 | 15+, incl. GHSA-pw9m-5jxm-xr6h (critical), GHSA-xg6x-h9c9-2m83 (critical) | 1.6.22 |
| `react-router` | 7.13.1 | 12, incl. GHSA-49rj-9fvp-4h2h (high, RCE) | 7.18.2 |
| `protobufjs` | 7.5.5 | 10, several high | 7.6.5 |
| `axios` | 1.15.1 | 8+, incl. GHSA-p92q-9vqr-4j8v, GHSA-35jp-ww65-95wh | 1.18.0 |
| `fast-uri` | 3.1.0 | 5 host-confusion / traversal | 3.1.5 |
| `hono` | 4.12.7 + 4.12.18 | 20+, incl. GHSA-88fw-hqm2-52qc (high, CORS) | 4.12.34 |
| `js-yaml` | 4.1.1 | 3, two high | 4.3.1 |
| `postcss` | 8.5.8 | 4, two high | 8.5.23 |
| `vite` | 6.4.1 + 7.3.1 | 4, incl. GHSA-p9ff-h696-f583 (high) | 6.4.3 / 7.3.5 |
| `sharp` | 0.33.5 | GHSA-f88m-g3jw-g9cj (high, libvips) | 0.35.0 |
| `lodash` | 4.17.23 | GHSA-r5fr-rjxr-66jc (high) | 4.18.0 |
| `form-data` | 4.0.5 | GHSA-hmw2-7cc7-3qxx (high) | 4.0.6 |
| `svgo` | 4.0.1 | GHSA-2p49-hgcm-8545 (high) | 4.0.2 |
| `devalue` | 5.6.4 | GHSA-77vg-94rm-hx3p (high) | 5.8.1 |
| `undici` | 6.21.3 | 7, incl. GHSA-vxpw-j846-p89q (high) | 6.28.0 |
| `qs` | 6.15.1 | GHSA-q8mj-m7cp-5q26 | 6.15.2 |
| `yaml` | 2.8.2 | GHSA-48c2-rrv3-qjmp | 2.8.3 |
| `body-parser` | 2.2.2 | GHSA-v422-hmwv-36x6 (low) | 2.3.0 |

**Confirmed not fixes, as flagged earlier:**

- `express-rate-limit` — 8.3.1 on `dev`, no advisory applies.
- `@astrojs/starlight` — `dev` already resolves 0.38.2, and I found no advisory for it either way.
- `astro` — this one has genuinely changed since the last review. Two new advisories landed with a `< 6.4.6` boundary (GHSA-2pvr-wf23-7pc7, high, host-header SSRF; GHSA-jrpj-wcv7-9fh9), so a bump inside 6.x now *does* cross a real boundary — the earlier "crosses no boundary" verdict is out of date. Three others remain unfixable without going to Astro 7.1.0.

**One thing worth knowing:** some of your pins have themselves aged past their advisories. `postcss` 8.5.22 is now covered by GHSA-fxqj-rqcc-2cmp (`<= 8.5.22`); `fast-uri` 3.1.4 by GHSA-7p8r-x3mc-p8w7 (`< 3.1.5`); `js-yaml` 4.3.0 by GHSA-5p4m-2wfm-xmqj (`< 4.3.1`); `hono` 4.12.31 by three advisories fixed in 4.12.34. That is not a criticism of the work — it's the nature of exact pins in a security PR, and it is one more argument for the caret-keeping point raised earlier.

Also worth a look separately: `axios` and `qs` already have root overrides whose *declared ranges* (`^1.15.0`, `^6.15.1`) permit the patched versions. They are stuck only because `bun.lock` froze an older resolution. Those two are fixable with a lock refresh and no manifest change at all.

## Why none of this got fixed automatically

I checked the `Dependabot Updates` workflow. It has failed on **every single run** going back months — 20+ consecutive failures. The cause is the same each time:

```
Dependabot can't update vulnerable dependencies for projects without a lockfile
or pinned version requirement as the currently installed version of X isn't known.
  dependency_file_not_supported
```

Dependabot's `npm_and_yarn` updater cannot read `bun.lock`. It sees our manifests, can't determine installed versions, and aborts. So the 30 alerts get *raised* but never *actioned*, and they never will under Dependabot.

That reframes your PR considerably. You weren't duplicating a bot's work — you were doing by hand the only thing that was going to get done at all. Thank you for that; it's also the piece of evidence that finally explains why the graph drifted this far. I'm filing the automation gap separately (Renovate does support Bun workspaces).

## What I've done

Rather than leave you to reassemble a month-old branch against a lockfile that has moved a long way, I've taken the first slice myself and opened #2659 — the `ws` override, alone, with its advisory line, exactly the template suggested above. It credits you in the commit message and the PR body, because the finding and the applicability analysis are yours. Its whole diff is four lines.

## What I'd suggest for this PR

I think the honest thing is to let this one go and treat #2659 as its continuation. Not because the work was wrong — most of it was right, and the parts that weren't were called out in good faith — but because rebasing a 99-package lock refresh across a month of drift is more work than re-deriving each piece, and several pins would need re-checking against newer advisories anyway. I'm not closing it unilaterally; that's the operator's call and I've flagged it for them.

If you'd like to keep going, the highest-value next slices are `better-auth` (two criticals, wants its own PR and a manual smoke against a real deployment), `react-router` (a high-severity RCE), and `protobufjs`. Any one of those on its own, with the advisory line, would be very welcome.

Two smaller notes so nothing is left hanging:

- **`.gitleaksignore`** — I independently verified all 15 fingerprints. Every one resolves to a real commit in our history, and every finding is a genuine false positive: documentation placeholders in `.env.example` and `docker.md`, the public `supabase-demo` JWT in an archived `ci.yml`, `sk-test-1234567890abcdef`, a published 512-bit RSA test vector in `auth.test.ts`, and the write-only `phc_*` PostHog key that `telemetry.ts` documents as safe to ship. The file is sound. It's still inert — there is no gitleaks workflow, no `.gitleaks.toml`, and our husky `pre-commit` only runs `lint-staged` — so whether to adopt a scanner at all is a decision I'm leaving to the operator rather than making by merging its allowlist first.
- **`packages/server/src/auth/instance.ts`** — the change is correct *and* entirely coupled to the Better Auth 1.6 bump. `ReturnType<typeof buildAuth>` and the `eslint-disable` only exist because `Auth<BetterAuthOptions>` became invariant in 1.6, and the `!user.email` guard only becomes reachable once the hook's parameter stops being hand-typed as `{ email: string }`. Landed without the bump it would be a regression (an eslint-disable and a lost explicit return type for nothing), so it belongs in the Better Auth PR and nowhere else. Worth carrying forward verbatim when that one gets written.

Thanks again for a genuinely careful submission — the rollback commit and the explicit "what I did not verify" list are still the most useful parts of any dependency PR we've had.

Wirasm:
Closing this one, and I want to be clear about why, because it is not a rejection.

Your analysis was sound. I re-verified every package you flagged against current `dev` and the advisory API, and the findings held up. What made this branch unmergeable was time, not quality: a month of drift left the lockfile conflicting badly, and several of your exact pins had themselves aged past newer advisories (`postcss` 8.5.22, `fast-uri` 3.1.4, `js-yaml` 4.3.0, `hono` 4.12.31 are all now covered by advisories published after you filed). Re-deriving each piece against today's `dev` turned out cheaper than salvaging a 99-package lock refresh. That is a scheduling outcome, not a fault in your work.

**The first slice is carried forward in #2659**, which pins `ws` to a patched version and closes GHSA-96hv-2xvq-fx4p and GHSA-58qx-3vcg-4xpx. You are credited by name in both its commit message and its PR body, because the finding and the applicability analysis are yours. It is a four-line diff and it is green.

**To be unambiguous: `dev` is not remediated.** The rest of your finding stands. 19 packages are still on vulnerable resolutions, including two criticals in `better-auth` and a high-severity RCE in `react-router`. Nothing here is being dropped. The remaining bumps are sequenced by severity, one PR each, in the shape agreed earlier in this thread: `better-auth` first (it wants its own PR plus a manual smoke, since it has no test file and CI cannot exercise web auth), then `react-router`, then `protobufjs`, then the transitive overrides. If you would like to pick any of those up, they are yours and very welcome. The two cheapest are `axios` and `qs`, which need no manifest change at all, only a lockfile refresh, because their existing override ranges already permit patched versions.

Two parts of your diff were deliberately left out, and both for reasons that are about us rather than about them:

- **`.gitleaksignore`** is verified sound. I checked all 15 fingerprints individually: every SHA resolves to a real commit in our history, and every finding is a genuine false positive. Documentation placeholders, the public `supabase-demo` JWT, `sk-test-1234567890abcdef`, a published RSA test vector, and the write-only `phc_*` PostHog key that `telemetry.ts` documents as safe to ship. The problem is that it attaches to nothing: there is no gitleaks workflow, no `.gitleaks.toml`, and our husky `pre-commit` runs only `lint-staged`. Whether to adopt a secret scanner at all is a maintainer decision that has not been made yet, and it should be made deliberately rather than arrive as an allowlist. That decision is pending, not declined.
- **`packages/server/src/auth/instance.ts`** is correct, and it is entirely coupled to the Better Auth 1.6 bump. `ReturnType<typeof buildAuth>` and the eslint-disable exist only because `Auth<BetterAuthOptions>` became invariant in 1.6, and the `!user.email` guard is unreachable while the hook parameter is hand-typed as `{ email: string }`. Landed on its own it would be a regression. It belongs with the Better Auth PR, and it is worth carrying across verbatim when that one gets written.

One last thing worth telling you, since it explains the whole situation. Our `Dependabot Updates` workflow has failed on every run since April. Dependabot's npm updater cannot read `bun.lock`, so it raises alerts but can never act on them. You were not duplicating a bot's work. You were doing by hand the only thing that was going to get done at all, and your PR is what surfaced the automation gap. Thank you for that, and for the rollback commit and the explicit list of what you had not verified. Those remain the most useful parts of any dependency PR we have had, and I would happily review anything else you send.

Wirasm:
Closing the loop on the one thing I left open here, @DIZ-admin.

When I closed this I said the `.gitleaksignore` decision was "pending, not declined." It has now been decided: **we are not adopting gitleaks**, so the allowlist will not be carried forward.

To be clear about why, because it is not a reflection on the work. Your 15 fingerprints were individually verified — every SHA resolves to a real commit in our history, and every finding is a genuine false positive (doc placeholders, the public `supabase-demo` JWT, a published 512-bit RSA test vector, the documented write-only PostHog key). The allowlist was sound. What it lacked was a scanner: there is no gitleaks workflow, no `.gitleaks.toml`, and our `pre-commit` hook runs only `lint-staged`. An allowlist attached to nothing is maintenance with no protection behind it, so rather than land half of a tool we have decided not to take either half for now.

If we do adopt secret scanning later, your verification is the obvious starting point and this PR is where we will look for it.

Thanks again for the dependency work — the `ws` slice from this PR is on `dev`, and the rest of what you found is sequenced by severity rather than dropped.

PR #2020, opened 2026-06-19
Fix Codex Minimal Reasoning Effort Execution

QwertyCoolMT:
﻿## Summary

Describe this PR in 2-5 bullets:

- Problem: workflow-level Codex options such as `modelReasoningEffort: minimal` were parsed but not forwarded into DAG node `assistantConfig`.
- Why it matters: users selecting Codex `minimal` reasoning could get a different runtime effort than configured, especially when tier defaults or aliases were involved.
- What changed: DAG execution now applies Codex workflow-level `modelReasoningEffort`, `webSearchMode`, and `additionalDirectories` to `assistantConfig`, with explicit workflow reasoning taking precedence over preset effort.
- What did **not** change (scope boundary): Claude node/workflow `effort`, Copilot/OpenCode behavior, Settings UI semantics, docs, generated API types, and unrelated workflow YAML/defaults drift.

### Before

```text
User                    Archon workflow engine              Codex provider
----                    ----------------------              --------------
sets minimal effort --> loader accepts workflow field
                        DAG builds node options
                        workflow Codex fields dropped ----> receives no minimal workflow effort
user observes run <---- response streams back                 SDK uses default/tier behavior
```

### After

```text
User                    Archon workflow engine                    Codex provider
----                    ----------------------                    --------------
sets minimal effort --> loader accepts workflow field
                        DAG builds node options
                        [applies Codex workflow assistant config] ==> receives minimal/webSearch/directories
user observes run <---- response streams back                       SDK uses configured workflow effort
```

### Before

```text
workflow schema -> loader -> dag-executor -> provider registry -> CodexProvider -> Codex SDK
                         \-> preset tier routing -> assistantConfig
```

### After

```text
workflow schema -> loader -> [~] dag-executor ===> assistantConfig ===> CodexProvider -> Codex SDK
                         \-> [~] preset tier routing respects explicit workflow modelReasoningEffort
```

**Connection inventory** (list every module-to-module edge, mark changes):

| From | To | Status | Notes |
|------|----|--------|-------|
| `workflowSchema` | `loader` | unchanged | Existing schema already parsed Codex workflow-level options. |
| `loader` | `dag-executor` | unchanged | Workflow object already carried parsed fields. |
| `dag-executor` | `assistantConfig` | **modified** | Codex workflow-level options are now copied into assistant config. |
| `dag-executor` | preset tier routing | **modified** | Explicit workflow `modelReasoningEffort` blocks preset effort overwrite for Codex. |
[...]

PR #2254, opened 2026-07-22
fix(workflows): forward workflow-level modelReasoningEffort and webSearchMode to Codex

### LINKED ISSUE #2246: fix(workflows): workflow-level modelReasoningEffort and webSearchMode are schema-accepted but never forwarded to the provider
The workflow schema accepts `modelReasoningEffort` and `webSearchMode` at the workflow level (`packages/workflows/src/schemas/workflow.ts:103-104`), but the executor never forwards them: `workflowLevelOptions` in `dag-executor.ts` threads effort/thinking/fallbackModel/betas/sandbox only, and `assistantConfig` is built purely from config.yaml (`{ ...(config.assistants[provider] ?? {}) }`, dag-executor.ts:~1018). A workflow declaring `modelReasoningEffort: minimal` parses fine and is then silently dropped — Codex runs at the config/SDK default.

Note this is a *separate path* from tier/alias effort resolution (`applyPresetOptions`/`routePresetEffort`), which only fires for tier-keyword/`@alias` models — literal workflow-level fields are the uncovered case.

**Fix shape** (~15 lines + one focused DAG test): thread the two fields from the workflow definition into the per-node `assistantConfig` (node-level overrides, if ever added, would sit above; config.yaml below).

**Prior art**: closed PR #2020 (@QwertyCoolMT — credit) had the correct forwarding fix; its unmergeable weight was all test-file churn since rewritten, and it also threaded `additionalDirectories`, which the schema strips — leave that out (or add it to the schema deliberately as its own decision). Silent-drop of schema-valid fields also violates the loud-mismatch principle — a load-time warning for accepted-but-unforwarded fields would be a worthwhile companion guard.

narutomugens-byte:
## Summary

- **Problem:** `modelReasoningEffort` and `webSearchMode` are accepted by the workflow schema at the top level but the DAG executor never threads them into the per-node `assistantConfig`, so Codex always ran at the `config.yaml` / SDK default regardless of what the workflow declared.
- **Why it matters:** A user declaring `modelReasoningEffort: minimal` in a workflow gets no error, no warning, and silently wrong behavior — a direct violation of Archon's fail-loud principle. The fields have been schema-valid-but-silently-dropped since `9235ad1e` (2026-03-26).
- **What changed:** Extended `WorkflowLevelOptions` to carry both fields; populated them from the parsed workflow definition; merged them into `assistantConfig` after `applyPresetOptions` (so an explicit workflow literal beats a preset-routed effort). Added one focused test asserting both fields reach Codex's `assistantConfig`.
- **What did not change:** `CodexProvider` / `parseCodexConfig` (the consumer is already correct); node-level fields (YAGNI — no current schema); `additionalDirectories` (schema strips it — deliberate out-of-scope, per prior art PR #2020). No change to any other package.

### Before

```
Workflow YAML                  DAG Executor               CodexProvider
─────────────────────────────  ─────────────────────────  ────────────────────────
modelReasoningEffort: minimal ─▶ parsed onto workflow def
                                  workflowLevelOptions {}  ← never carried
                                  assistantConfig {}        ← never populated
                                                           ─▶ parseCodexConfig({})
                                                              modelReasoningEffort = undefined
                                                              Codex SDK default effort  ← silent bug
```

### After

```
Workflow YAML                  DAG Executor               CodexProvider
─────────────────────────────  ─────────────────────────  ────────────────────────
modelReasoningEffort: minimal ─▶ parsed onto workflow def
                                  workflowLevelOptions {   ← [now populated]
                                    modelReasoningEffort: 'minimal'
                                    webSearchMode: 'live'
                                  }
                                  applyPresetOptions(...)  ← preset effort (if any)
                                  assistantConfig {        ← [now merged, wins over preset]
                                    modelReasoningEffort: 'minimal'
                                    webSearchMode: 'live'
                                  }
                                                           ─▶ parseCodexConfig(assistantConfig)
                                                              modelReasoningEffort = 'minimal'
                                                              Codex SDK: minimal effort  ← fixed
```

### Before

```
WorkflowDefinition
  .modelReasoningEffort ──X──▶ (dropped, never carried)
  .webSearchMode        ──X──▶ (dropped, never carried)

workflowLevelOptions { effort, thinking, fallbackModel, betas, sandbox, workflowTier }
                      ↓
assistantConfig { ...(config.assistants[provider]) }
                      ↓ applyPresetOptions (tier/alias path only)
                      ↓
CodexProvider.parseCodexConfig(assistantConfig)
  → modelReasoningEffort: undefined   ← bug
  → webSearchMode: undefined          ← bug
```

### After

```
WorkflowDefinition
  .modelReasoningEffort ──▶ workflowLevelOptions.modelReasoningEffort  [~]
  .webSearchMode        ──▶ workflowLevelOptions.webSearchMode          [~]

workflowLevelOptions { effort, thinking, fallbackModel, betas, sandbox,
                       modelReasoningEffort, webSearchMode, workflowTier }  [~]
                      ↓
assistantConfig { ...(config.assistants[provider]) }
                      ↓ applyPresetOptions (tier/alias path — unchanged)
                      ↓ merge workflowLevelOptions.modelReasoningEffort   [+]
                      ↓ merge workflowLevelOptions.webSearchMode           [+]
                      ↓
CodexProvider.parseCodexConfig(assistantConfig)
  → modelReasoningEffort: 'minimal'  ← fixed
  → webSearchMode: 'live'            ← fixed
```

**Connection inventory:**

| `WorkflowDefinition.modelReasoningEffort` | `workflowLevelOptions` | **modified** | Previously never copied in |
| `WorkflowDefinition.webSearchMode` | `workflowLevelOptions` | **modified** | Previously never copied in |
| `workflowLevelOptions` | `assistantConfig` (merge block) | **new** | New 4-line merge after `applyPresetOptions` |
| `assistantConfig` | `CodexProvider.parseCodexConfig` | unchanged | Consumer was already correct |
| `applyPresetOptions` | `assistantConfig.modelReasoningEffort` | unchanged | Preset/tier path — not changed |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `workflows`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2246
- Supersedes #2020 (prior attempt — same forwarding logic, narrower scope here)

```bash
bun run type-check   # ✅ all 10 packages clean
bun run lint         # ✅ 0 errors, 0 warnings
bun run format:check # ✅ pass
bun run test         # ✅ 1008 passed (workflows package), 0 failed
bun run validate     # ✅ all 9 checks pass
```

- **New test**: `forwards workflow-level modelReasoningEffort and webSearchMode to Codex assistantConfig (#2246)` — asserts `assistantConfig.modelReasoningEffort === 'minimal'` and `assistantConfig.webSearchMode === 'live'` on a workflow that declares both fields at the top level with provider `codex`.
- **Pre-existing failures**: 2 Windows `tar` path failures in `downloadWebDist` — present on `dev` before this branch, unrelated to this change.
- If any command is intentionally skipped: none skipped.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? Yes — workflows without these fields are byte-identical to before; only workflows that already (futilely) declared them now have those values respected.
- Config/env changes? No

- **Verified scenarios:** Type check, lint, format, and test suite all pass. New test asserts both fields reach `assistantConfig` with correct values when set at the workflow level.
- **Edge cases checked:** A workflow declaring `modelReasoningEffort` on a non-Codex node (e.g. Claude) — the key is added to `assistantConfig` but `parseClaudeConfig` ignores unknown keys, so it is a harmless no-op. Workflow-level field vs. tier/alias preset effort — the merge runs *after* `applyPresetOptions`, so the explicit workflow literal wins (matches PR #2020 precedence).
- **What was not verified:** Full end-to-end manual Codex run with a live workflow (no Codex binary available in the test environment). The code path through `parseCodexConfig` → `buildThreadOptions` was verified by reading the existing consumer code and the new test; a manual smoke test with a real Codex binary is the one remaining gap.

- **Affected subsystems/workflows:** Only `dag-executor.ts` (executor producer path). `CodexProvider` and all other providers are untouched.
- **Potential unintended effects:** None — Claude/Pi config parsers ignore `modelReasoningEffort`/`webSearchMode` keys, so merging them for non-Codex nodes is a no-op. The merge is guarded by `!== undefined` checks.
- **Guardrails:** Existing `dag.preset_effort_unsupported` warning for capability mismatches remains; the new merge path does not bypass it.

## Rollback Plan (required)

- **Fast rollback:** `git revert 074e2fb4` — a 4-line addition in the executor; reverts cleanly with no migration.
- **Feature flags or config toggles:** None.
- **Observable failure symptoms:** A Codex workflow with explicit `modelReasoningEffort` would silently revert to config/SDK default — same as the pre-fix behavior.

## Risks and Mitigations

- **Risk:** `modelReasoningEffort` set on a workflow whose resolved provider is Claude/Pi — the key appears in `assistantConfig` but does nothing.
  - **Mitigation:** Both providers' config parsers already ignore unknown keys; no behavioral change. A follow-up loud warn (mirroring `dag.preset_effort_unsupported`) could surface this; deferred as a separate concern.

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -9481,7 +9481,7 @@ describe('executeDagWorkflow -- Claude SDK advanced options', () => {
     expect(nodeConfig?.effort).toBe('high');
   });
 
-  it('forwards workflow-level modelReasoningEffort and webSearchMode to Codex assistantConfig (#2246)', async () => {
+  it('forwards workflow-level modelReasoningEffort and webSearchMode to Codex assistantConfig', async () => {
     mockGetAgentProviderDag.mockImplementation(() => ({
       sendQuery: mockSendQueryDag,
       getType: () => 'codex',
@@ -9519,7 +9519,7 @@ describe('executeDagWorkflow -- Claude SDK advanced options', () => {
     expect(assistantConfig?.webSearchMode).toBe('live');
   });
 
-  it('workflow-level modelReasoningEffort wins over preset-routed effort (#2246)', async () => {
+  it('workflow-level modelReasoningEffort wins over preset-routed effort', async () => {
     mockGetAgentProviderDag.mockImplementation(() => ({
       sendQuery: mockSendQueryDag,
       getType: () => 'codex',
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -9519,6 +9519,53 @@ describe('executeDagWorkflow -- Claude SDK advanced options', () => {
     expect(assistantConfig?.webSearchMode).toBe('live');
   });
 
+  it('workflow-level modelReasoningEffort wins over preset-routed effort (#2246)', async () => {
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'codex',
+      getCapabilities: mockCodexCapabilities,
+    }));
+    const mockDeps = createMockDeps();
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+    const aiProfile = buildAiProfile('claude', {
+      repoTiers: {
+        large: { provider: 'codex', model: 'gpt-5', effort: 'medium' },
+      },
+    });
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-dag',
+      testDir,
+      {
+        name: 'workflow-effort-precedence-test',
+        nodes: [{ id: 'step1', command: 'my-cmd', model: 'large' }],
+        modelReasoningEffort: 'high',
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      aiProfile
+    );
+
+    expect(mockSendQueryDag.mock.calls.length).toBeGreaterThan(0);
+    const optionsArg = mockSendQueryDag.mock.calls[0][3] as Record<string, unknown>;
+    const assistantConfig = optionsArg?.assistantConfig as Record<string, unknown>;
+    // workflow literal 'high' must override the tier-routed 'medium'
+    expect(assistantConfig?.modelReasoningEffort).toBe('high');
+  });
+
   it('per-node effort overrides workflow-level effort', async () => {
     const mockDeps = createMockDeps();
     const platform = createMockPlatform();
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -9481,6 +9481,44 @@ describe('executeDagWorkflow -- Claude SDK advanced options', () => {
     expect(nodeConfig?.effort).toBe('high');
   });
 
+  it('forwards workflow-level modelReasoningEffort and webSearchMode to Codex assistantConfig (#2246)', async () => {
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'codex',
+      getCapabilities: mockCodexCapabilities,
+    }));
+    const mockDeps = createMockDeps();
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-dag',
+      testDir,
+      {
+        name: 'workflow-codex-options-test',
+        nodes: [{ id: 'step1', command: 'my-cmd' }],
+        modelReasoningEffort: 'minimal',
+        webSearchMode: 'live',
+      },
+      workflowRun,
+      'codex',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    expect(mockSendQueryDag.mock.calls.length).toBeGreaterThan(0);
+    const optionsArg = mockSendQueryDag.mock.calls[0][3] as Record<string, unknown>;
+    const assistantConfig = optionsArg?.assistantConfig as Record<string, unknown>;
+    expect(assistantConfig?.modelReasoningEffort).toBe('minimal');
+    expect(assistantConfig?.webSearchMode).toBe('live');
+  });
+
   it('per-node effort overrides workflow-level effort', async () => {
     const mockDeps = createMockDeps();
     const platform = createMockPlatform();
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -214,7 +214,8 @@ function applyPresetOptions(
   if (
     preset.effort === undefined ||
     node.effort !== undefined ||
-    workflowLevelOptions.effort !== undefined
+    workflowLevelOptions.effort !== undefined ||
+    workflowLevelOptions.modelReasoningEffort !== undefined
   ) {
     return;
   }
@@ -290,7 +291,7 @@ export async function loadConfiguredMcpServerNames(
   }
 }
 
-/** Workflow-level Claude SDK options — per-node overrides take precedence via ?? */
+/** Workflow-level provider options — per-node overrides take precedence via ?? */
 interface WorkflowLevelOptions {
   effort?: EffortLevel;
   thinking?: ThinkingConfig;
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -49,6 +49,8 @@ import type {
   EffortLevel,
   ThinkingConfig,
   SandboxSettings,
+  ModelReasoningEffort,
+  WebSearchMode,
   WorkflowSource,
   LoopGateRunMetadata,
   ApprovalContext,
@@ -295,6 +297,10 @@ interface WorkflowLevelOptions {
   fallbackModel?: string;
   betas?: string[];
   sandbox?: SandboxSettings;
+  /** Codex-only literal workflow-level reasoning effort (#2246). */
+  modelReasoningEffort?: ModelReasoningEffort;
+  /** Codex-only literal workflow-level web-search mode (#2246). */
+  webSearchMode?: WebSearchMode;
   /** Workflow-level tier keyword (when `workflow.model` is small/medium/large), so
    *  nodes that inherit the workflow model can still surface the `← tier` annotation. */
   workflowTier?: 'small' | 'medium' | 'large';
@@ -1107,6 +1113,20 @@ async function resolveNodeProviderAndModel(
     nodeConfig,
     assistantConfig
   );
+
+  // Forward workflow-level literal Codex fields (#2246). The schema accepts these at
+  // the workflow level (workflow.ts:124-125) but they were never threaded — a workflow
+  // declaring `modelReasoningEffort: minimal` parsed fine and ran at the config/SDK
+  // default. Precedence: node-level (if ever added) > workflow-level > tier/alias preset
+  // > config.yaml. Applied AFTER applyPresetOptions so an explicit workflow value wins
+  // over a preset-routed effort (matches PR #2020's chosen precedence). Codex-specific
+  // keys; Claude/Pi config parsers ignore them, so this is a no-op for other providers.
+  if (workflowLevelOptions.modelReasoningEffort !== undefined) {
+    assistantConfig.modelReasoningEffort = workflowLevelOptions.modelReasoningEffort;
+  }
+  if (workflowLevelOptions.webSearchMode !== undefined) {
+    assistantConfig.webSearchMode = workflowLevelOptions.webSearchMode;
+  }
 
   const options: SendQueryOptions = {
     ...baseOptions,
@@ -6769,6 +6789,8 @@ export async function executeDagWorkflow(
     fallbackModel: workflow.fallbackModel,
     betas: workflow.betas,
     sandbox: workflow.sandbox,
+    modelReasoningEffort: workflow.modelReasoningEffort,
+    webSearchMode: workflow.webSearchMode,
     workflowTier,
   };
   const layers = buildTopologicalLayers(workflow.nodes);
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -298,9 +298,9 @@ interface WorkflowLevelOptions {
   fallbackModel?: string;
   betas?: string[];
   sandbox?: SandboxSettings;
-  /** Codex-only literal workflow-level reasoning effort (#2246). */
+  /** Codex-only: workflow-level reasoning effort. */
   modelReasoningEffort?: ModelReasoningEffort;
-  /** Codex-only literal workflow-level web-search mode (#2246). */
+  /** Codex-only: workflow-level web-search mode. */
   webSearchMode?: WebSearchMode;
   /** Workflow-level tier keyword (when `workflow.model` is small/medium/large), so
    *  nodes that inherit the workflow model can still surface the `← tier` annotation. */
@@ -1115,13 +1115,8 @@ async function resolveNodeProviderAndModel(
     assistantConfig
   );
 
-  // Forward workflow-level literal Codex fields (#2246). The schema accepts these at
-  // the workflow level (workflow.ts:124-125) but they were never threaded — a workflow
-  // declaring `modelReasoningEffort: minimal` parsed fine and ran at the config/SDK
-  // default. Precedence: node-level (if ever added) > workflow-level > tier/alias preset
-  // > config.yaml. Applied AFTER applyPresetOptions so an explicit workflow value wins
-  // over a preset-routed effort (matches PR #2020's chosen precedence). Codex-specific
-  // keys; Claude/Pi config parsers ignore them, so this is a no-op for other providers.
+  // Applied AFTER applyPresetOptions so an explicit workflow value wins over a
+  // preset-routed effort. Precedence: workflow-level > tier preset > config.yaml.
   if (workflowLevelOptions.modelReasoningEffort !== undefined) {
     assistantConfig.modelReasoningEffort = workflowLevelOptions.modelReasoningEffort;
   }


narutomugens-byte:
# 🔍 Comprehensive PR Review

**PR**: #2254 — fix(workflows): forward workflow-level modelReasoningEffort and webSearchMode to Codex
**Reviewed by**: 2 specialized agents (code-review, test-coverage)
**Date**: 2026-07-22

---

## Summary

Minimal, surgical fix that correctly threads `modelReasoningEffort` and `webSearchMode` from workflow-level YAML into `assistantConfig` for Codex nodes. The implementation places the new code AFTER `applyPresetOptions`, so workflow-literal values win over tier-preset-routed effort — matching the PR #2020 precedence design. Both agents independently **recommend APPROVE**.

**Verdict**: ✅ APPROVE

| Severity | Count |
|----------|-------|
| 🔴 CRITICAL | 0 |
| 🟠 HIGH | 0 |
| 🟡 MEDIUM | 1 |
| 🟢 LOW | 2 |

---

### Missing precedence-conflict test

📍 `packages/workflows/src/dag-executor.test.ts` (new test needed)

The implementation comment explicitly documents: *"Applied AFTER `applyPresetOptions` so an explicit workflow value wins over a preset-routed effort."* No test exercises the conflict scenario (tier preset with `effort: 'medium'` + workflow-literal `modelReasoningEffort: 'high'`). Both review agents flagged this. A future reorder of these two blocks would be a silent regression.

**Options**: Fix now (~20 lines) | Create issue | Skip

<details>
<summary>View suggested test</summary>

```typescript
it('workflow-level modelReasoningEffort wins over preset-routed effort (#2246)', async () => {
  mockGetAgentProviderDag.mockImplementation(() => ({
    sendQuery: mockSendQueryDag,
    getType: () => 'codex',
    getCapabilities: mockCodexCapabilities,
  }));
  const mockDeps = createMockDeps();
  const platform = createMockPlatform();
  const workflowRun = makeWorkflowRun();

  await executeDagWorkflow(
    mockDeps, platform, 'conv-dag', testDir,
    {
      name: 'workflow-effort-precedence-test',
      nodes: [{ id: 'step1', command: 'my-cmd', model: 'large' }],
      modelReasoningEffort: 'high',  // workflow literal should beat any preset-routed 'medium'
    },
    workflowRun, 'codex', undefined,
    join(testDir, 'artifacts'), join(testDir, 'logs'),
    'main', 'docs/', minimalConfig
  );

  const optionsArg = mockSendQueryDag.mock.calls[0][3] as Record<string, unknown>;
  const assistantConfig = optionsArg?.assistantConfig as Record<string, unknown>;
  expect(assistantConfig?.modelReasoningEffort).toBe('high');
});
```

Pattern reference: `per-node effort overrides workflow-level effort` at `dag-executor.test.ts:9522`.

</details>

---

## 🟢 Low Issues

| Issue | Location | Suggestion |
|-------|----------|------------|
| Interface JSDoc says "Claude SDK options" but now includes Codex-only fields | `dag-executor.ts:293` | Change to `"Workflow-level provider options — per-node overrides take precedence via ??"` (one-line edit) |
| `applyPresetOptions` guard checks `workflowLevelOptions.effort` but not `modelReasoningEffort` — asymmetric pattern | `dag-executor.ts:214-220` | Add to early-return guard for symmetry (agent recommends deferring — not a bug today) |

---

## ✅ What's Good

- **Minimal diff**: 60 lines added, 0 deleted — exactly proportional to the fix
- **Type imports from correct source**: `ModelReasoningEffort`/`WebSearchMode` from `@archon/workflows/schemas`, not `@archon/providers/types` — avoids cross-package dependency creep
- **Correct precedence ordering**: new code placed AFTER `applyPresetOptions`, matching PR #2020 design
- **`undefined` guard discipline**: `!== undefined` rather than falsy — correctly handles edge values
- **No scope creep**: `additionalDirectories`, node-level fields, and CodexProvider internals correctly left untouched
- **CLAUDE.md compliance**: type safety ✅, fail fast ✅, YAGNI ✅, comment quality ✅, import style ✅
- **Test quality**: correct describe block, uses `mockCodexCapabilities`, tests both fields in one call, self-contained

---

*Reviewed by Archon `comprehensive-pr-review` workflow — 2 of 5 agents completed*
*Full artifacts: `artifacts/runs/91790fd000bdf60245ee4e59c3b62cd6/review/`*

narutomugens-byte:

**Status**: COMPLETE
**Pushed**: ✅ Changes pushed to `archon/task-archon-fix-github-issue-routed-1784750401036`

---

### Fixes Applied (3 total)

| Severity | Count |
|----------|-------|
| 🔴 CRITICAL | 0 |
| 🟠 HIGH | 0 |
| 🟡 MEDIUM | 1 |
| 🟢 LOW | 2 |

<details>

- ✅ **Missing precedence-conflict test case** (`dag-executor.test.ts`) — Added test: Codex tier preset with `effort: 'medium'` + workflow `modelReasoningEffort: 'high'` → asserts final value is `'high'`. Pins the "workflow literal wins over preset" contract.
- ✅ **Interface JSDoc comment drift** (`dag-executor.ts:293`) — Changed "Workflow-level Claude SDK options" → "Workflow-level provider options" — the interface now contains Codex-only fields, so the old comment was misleading.
- ✅ **Guard gap in applyPresetOptions** (`dag-executor.ts:214-220`) — Added `workflowLevelOptions.modelReasoningEffort !== undefined` to the early-return guard. Eliminates the double-write and makes the guard pattern symmetric with `workflowLevelOptions.effort`.

</details>

---

### Tests Added

| File | Test Cases |
|------|------------|
| `packages/workflows/src/dag-executor.test.ts` | `workflow-level modelReasoningEffort wins over preset-routed effort (#2246)` |

---

### Skipped (0)

---

*(none)*

---

### Validation

✅ Type check | ✅ Lint | ✅ Tests (430 passed)

---

*Self-fix by Archon · aggressive mode · fixes pushed to `archon/task-archon-fix-github-issue-routed-1784750401036`*

Wirasm:
---
pr: 2254
base: dev
head: archon/task-archon-fix-github-issue-routed-1784750401036
reviewed: 2026-08-17T09:12:23Z
reviewed_head: 35c5a9b374d9898e6c2f587391e27ceda788b6b6
verdict: NEEDS FIXES
open_findings: 8
scopes: [code, seams, docs]
publication: https://github.com/coleam00/Archon/pull/2254#issuecomment-5314108113
---



# PR Review: #2254 — fix(workflows): forward workflow-level modelReasoningEffort and webSearchMode to Codex

## Signal

The bug is real and worth fixing — `modelReasoningEffort:`/`webSearchMode:` have been documented workflow-level fields (`authoring-workflows.md:126-127`, `book/quick-reference.md:112-113`) that the executor never read, so the shipped docs have been describing a feature that did not exist. This PR makes them true. But the implementation has one causal defect that produces a **new** silent drop while fixing an old one: the `applyPresetOptions` guard added at `dag-executor.ts:214-221` is not scoped to the resolved provider, so a workflow-level `modelReasoningEffort` suppresses tier/alias preset effort on *every* node in the workflow — including Claude and Pi nodes that would otherwise have received it, with no warning on any channel (R1). Three of the remaining findings share that same root cause: the change writes into `assistantConfig` unconditionally, with no provider gate and no capability axis, so the same declared value silently works on Codex, silently half-works on Copilot, and silently no-ops on Claude/Pi (R3), and its own "Codex-only" doc comment is false — a label the shipped docs repeat in three places, which this PR is what turns from imprecise into wrong (R8). Separately, the branch is 96 commits stale in a way that a clean `git merge-tree` hides: both of the PR's own new tests fail after merge, and the merged ordering re-breaks the telemetry gap that #2395 was closed to fix (R2, R4). On the architectural question the brief poses — one workflow-level spelling for reasoning effort or two — the answer is **one**, but the consolidation is a separate change than this one and the brief's proposed mechanism does not work as sketched; see R7.

- **Blocking:** 4 (R1 Critical, R2/R3/R4 Important)
- **Non-blocking:** 4 (R5, R6, R7, R8 Suggestions)
- **Resolved since previous review:** Not applicable — first review
- **Tracked follow-ups:** None (R7 needs an issue filed; not created without maintainer sign-off)
- **Validation:** `bun run validate` PASSES at the PR head (all 10 packages, 0 test failures). Merged onto current `dev` the PR's two new tests FAIL. GitHub CI has never run on this PR.

## Findings

|---|---|---|---|
| `R1` | Critical | `applyPresetOptions` guard is not provider-scoped — a workflow-level `modelReasoningEffort` silently drops tier/alias preset effort on Claude/Pi nodes in the same workflow | OPEN |
| `R2` | Important | After merge with `dev`, the write lands *after* `resolvedEffort` is captured, so the audit trail and live console report a different effort than the one Codex runs at | OPEN |
| `R3` | Important | No capability warning for either field, and the "Codex-only" comment is false — the write also reaches Copilot, where `minimal` silently no-ops | OPEN |
| `R4` | Important | Both of the PR's new tests fail when merged onto `dev`; `git merge-tree` is clean and type-check passes, so nothing flags it | OPEN |
| `R5` | Suggestion | `WorkflowLevelOptions`'s own doc comment promises per-node override precedence that does not exist for these two fields | OPEN |
| `R6` | Suggestion | The Codex effort enum has four independently maintained copies; this PR adds a fifth live consumer that trusts one of them without re-checking | OPEN |
| `R7` | Suggestion | Two workflow-level spellings for reasoning depth — the brief's architectural question, answered: consolidate to one, as a follow-up, not here | OPEN |
| `R8` | Suggestion | Shipped docs label `modelReasoningEffort:` "Codex only" in three places; this PR is what makes that label actively wrong | OPEN |

## Detailed Findings

<details>
<summary><code>R1</code> — <code>applyPresetOptions</code> guard suppresses preset effort on every provider, not just Codex</summary>

**Impact:** A mixed-provider workflow that declares `modelReasoningEffort: high` (intending it for its Codex nodes) silently loses tier/alias preset effort on its Claude and Pi nodes. Those nodes run at the provider default instead of the preset's `effort`, with **no warning on any channel** — not the `dag.preset_effort_unsupported` log, not the user-facing capability message. This is a regression against pre-PR behavior, where `routePresetEffort` would have routed the preset value correctly.

**Evidence:** `packages/workflows/src/dag-executor.ts:214-221` (PR head) adds a fourth disjunct to the early return:

```ts
if (
  preset.effort === undefined ||
  node.effort !== undefined ||
  workflowLevelOptions.effort !== undefined ||
  workflowLevelOptions.modelReasoningEffort !== undefined   // new, unconditional
) {
  return;
}
const routed = routePresetEffort(provider, preset.effort);   // never reached
```

`provider` is a parameter of the function but is only consulted *after* the guard, at line 223. `workflowLevelOptions` is workflow-scoped, so the guard fires for every node regardless of the node's own resolved provider (per-node `provider:` overrides and tier/`@alias`-resolved providers both exist). The existing `dag.preset_effort_unsupported` warn at `:227-231` only fires when `routePresetEffort` returns `null` — here it is never called, so the drop is completely silent.

Confirmed downstream: `routePresetEffort('claude', 'medium')` returns `{ field: 'effort', value: 'medium' }` (`model-validation.ts:252-259`, `CLAUDE_EFFORTS` at `:230`), so the value being lost is a real one. And the compensating write at `dag-executor.ts:1119-1122` sets `assistantConfig.modelReasoningEffort` on that Claude node too, where `parseClaudeConfig` (`packages/providers/src/claude/config.ts`) never reads the key — so the node loses its preset effort and gains nothing.

Neither of the PR's two new tests covers this: both resolve to `provider: 'codex'`.

**Required outcome:** Scope the new disjunct to the case it is defending — the guard should only suppress preset-effort routing when the resolved provider is the one that consumes `assistantConfig.modelReasoningEffort` (i.e. gate on `routePresetEffort(provider, preset.effort)?.field === 'modelReasoningEffort'`, or simply on `provider === 'codex'`), and the compensating write at `:1119-1122` should be provider-gated the same way. Add a test for the mixed-provider direction: workflow-level `modelReasoningEffort` set, node resolving to Claude via a preset carrying `effort`, assert `nodeConfig.effort` still receives the preset value.

**Disposition:** OPEN. Verified independently by the coordinator against `dag-executor.ts:214-221` and `model-validation.ts:246-259`.

</details>

<details>
<summary><code>R2</code> — post-merge, the audit trail and console report an effort the node did not run at</summary>

**Impact:** After merging onto current `dev`, a Codex node that declares workflow-level `modelReasoningEffort: high` genuinely runs at `high`, but the persisted `workflow_events` row for `node_started` and the live console/SSE feed both report **no effort at all**. That re-opens exactly the gap issue #2395 was closed to fix ("effort is resolved then discarded — never reaches the event stream"), for the one case this PR is introducing.

**Evidence:** Current `dev` computes the telemetry value immediately after `applyPresetOptions`, guarded by an explicit invariant comment (`packages/workflows/src/dag-executor.ts:1168-1180`):

> `// Read POST-routing values only. applyPresetOptions -> routePresetEffort has already placed effort where the provider actually consumes it … So both reads below hold effort that will genuinely be applied.`

```ts
const assistantEffort = assistantConfig.modelReasoningEffort;
const resolvedEffort = nodeConfig.effort ?? (typeof assistantEffort === 'string' ? assistantEffort : undefined);
```

In the merge preview (`git merge-tree origin/dev pr-2254-review`, extracted to `scratchpad/merged-dag-executor.ts:1178-1188`) the PR's write lands **after** that read:

- `:1160` `applyPresetOptions(...)`
- `:1178-1180` `resolvedEffort` computed
- `:1184-1188` PR's write — `assistantConfig.modelReasoningEffort = workflowLevelOptions.modelReasoningEffort`
- `:1191` `options` built from the now-overwritten `assistantConfig`

`assistantConfig` is captured by reference into `options`, so the SDK call receives the correct value while `resolvedEffort` holds the stale one. `resolvedEffort` is threaded into both the persisted DB event and the in-process emitter (`dag-executor.ts:1349`, `:1369`, `:4087`, `:4105`, `:5302`, `:7138` on dev). Compounded by R1: because the guard stops the preset from writing `assistantConfig.modelReasoningEffort` at all, the read at `:1178` sees `undefined` even when a preset was in play. The dev-side invariant comment becomes false the moment this lands. `git merge-tree` resolves this cleanly, so it will land exactly this way unless the rebase moves the block.

**Required outcome:** On rebase, place the workflow-level write **before** the `resolvedEffort` computation (or recompute `resolvedEffort` after it), and keep dev's invariant comment true. A test asserting the `node_started` event's `effort` matches the value actually passed to the provider would pin it.

**Disposition:** OPEN. Coordinator-verified in the merged tree; this is a rebase-conditional defect, not present at the PR head.

</details>

<details>
<summary><code>R3</code> — the write has no capability gate, and "Codex-only" is false</summary>

**Impact:** The same declared workflow-level value has three different fates depending on the node's resolved provider, and the author gets no signal in any of them:

- **Codex** — applied (`packages/providers/src/codex/provider.ts:92-93`). Intended.
- **Copilot** — *also applied*, as a fallback when the node has no `effort`/`thinking`. `resolveCopilotReasoning` (`packages/providers/src/community/copilot/provider.ts:127-133`, called at `:322`) returns `copilotConfig.modelReasoningEffort`, and `copilotConfig` is `parseCopilotConfig(assistantConfig)`. But that parser accepts only `low|medium|high|xhigh` (plus `max` → `xhigh`) at `community/copilot/config.ts:20-30`, so the documented Codex value `minimal` is silently dropped there.
- **Claude / Pi** — dead write. `parseClaudeConfig` (`providers/src/claude/config.ts`) never reads the key; `PiProviderDefaults` (`providers/src/types.ts:85+`) has no such field.

AGENTS.md's Workflow Language Constitution states: *"New per-provider capabilities default to provider config / tier-alias presets, not new node fields; capability mismatches warn loudly (capabilities.ts is the source of truth)."* Neither field appears in `capChecks` (`dag-executor.ts:1065-1086` — `effort`, `thinking`, `sandbox`, `fallbackModel`, `settingSources` and others all have entries), and neither has a `ProviderCapabilities` axis (`providers/src/types.ts` — only `effortControl`/`thinkingControl` exist). So `capabilities.ts` is not consulted for either field, and the generated capability matrix cannot surface the distinction. This is the same class of quiet failure the PR is otherwise fixing.

**Evidence:** `dag-executor.ts:1119-1124` (PR head) writes unconditionally, with the doc comments at `:301-304` labelling both fields "Codex-only". `providers/src/types.ts:41-50` (`CopilotProviderDefaults.modelReasoningEffort`) explicitly states it "mirrors `CodexProviderDefaults.modelReasoningEffort`", falsifying that label.

**Required outcome:** Either gate the write on the resolved provider and warn when the field is declared but the resolved provider does not consume it (the loud-mismatch path AGENTS.md requires — issue #2246 asks for exactly this: *"a load-time warning for accepted-but-unforwarded fields would be a worthwhile companion guard"*), or add a capability axis so `capChecks` covers it. At minimum, correct the "Codex-only" comments to name Copilot and note the `minimal` gap.

**Disposition:** OPEN. Coordinator-verified against `community/copilot/provider.ts:127-133,322`, `community/copilot/config.ts:20-30`, and `dag-executor.ts:1065-1086`.

</details>

<details>
<summary><code>R4</code> — the PR's own tests fail on merge, and nothing catches it</summary>

**Impact:** The PR cannot be merged as-is. Both new tests fail after a textually clean merge with `dev`, and the failure mode is invisible to every automated gate that currently runs.

**Evidence:** Branch point is `ace0a41a`; `dev` is 96 commits ahead. `#2299` (commit `8e63856e`, 2026-08-05) inserted a `stateDir: string` positional parameter into `executeDagWorkflow` between `artifactsDir` and `logDir`. The PR's tests pass `…, join(testDir,'artifacts'), join(testDir,'logs'), 'main', 'docs/', minimalConfig`, which after merge shift by one so `config` arrives `undefined` and the node never reaches `sendQuery`:

```
(fail) forwards workflow-level modelReasoningEffort and webSearchMode to Codex assistantConfig
(fail) workflow-level modelReasoningEffort wins over preset-routed effort
  expect(mockSendQueryDag.mock.calls.length).toBeGreaterThan(0)  →  Received: 0
```

Three gates miss it: `git merge-tree` reports no conflict; `bun run type-check` passes on the merged tree because test files are excluded from the type-check project; and GitHub CI has never run on this PR at all (`gh api .../check-runs` returns zero check runs for `35c5a9b3` — the author is `authorAssociation: NONE` and the PR is a draft, so only CodeRabbit's "review skipped" status exists).

**Required outcome:** Rebase onto current `dev`, fix the two call sites, and re-run `bun run validate`. Note this is also the point at which R2 must be handled, since the correct insertion point moves.

**Found by:** `validation`

**Disposition:** OPEN.

</details>

<details>
<summary><code>R5</code> — the interface promises a per-node override that does not exist</summary>

**Impact:** `WorkflowLevelOptions`'s doc comment (`dag-executor.ts:294`, edited by this PR) reads *"Workflow-level provider options — per-node overrides take precedence via `??`"*. That is true of every other member and false for both new ones. An author cannot override the workflow-level value on a single node.

**Evidence:** `packages/workflows/src/schemas/dag-node.ts` has no `modelReasoningEffort` or `webSearchMode` field — the names appear only in `workflowBaseSchema` (`schemas/workflow.ts:149-150`). Every sibling resolves as `node.X ?? workflowLevelOptions.X` (`dag-executor.ts:1157-1160` on dev); these two are read directly from `workflowLevelOptions`. Worse for the Codex case: `buildThreadOptions` (`providers/src/codex/provider.ts:79-93`) never reads `nodeConfig.effort`, so writing `effort:` on a Codex node is a no-op that produces the generic capability warning *"Node uses effort but codex doesn't support it — ignored"* while a same-purpose workflow-level `modelReasoningEffort` silently drives the real value. The warning implies effort was dropped when in fact a different field applied one.

**Required outcome:** Correct the comment to scope its precedence claim, or accept the asymmetry explicitly in the comment. Adding node-level fields is *not* the fix — AGENTS.md's constitution says the opposite (*"New per-provider capabilities default to provider config / tier-alias presets, not new node fields"*), and R7 is the real resolution.

**Mitigating:** an author who *tries* a node-level `modelReasoningEffort:` is not left in silence — the key is absent from `KNOWN_DAG_NODE_KEYS` (`schemas/dag-node.ts:735`), so it trips the unknown-key warning path documented at `authoring-workflows.md:1752` ("Unknown Keys Are Reported, Not Rejected"). The false promise is in the interface comment, not in the author-facing behaviour.

**Found by:** `prp-core:code-reviewer` (raised and set aside by `prp-core:seam-analyzer`, which judged the missing node-level field consistent with stated project convention rather than a two-sided seam — the reviewers agree on the facts and differ only on whether the asymmetry is itself a defect; recorded here as a comment-accuracy issue, which is the part both support; the mitigating evidence came from `prp-core:docs-impact-agent`)

</details>

<details>
<summary><code>R6</code> — four hand-maintained copies of the Codex effort enum, now with a live consumer</summary>

**Impact:** If the Codex effort vocabulary ever changes, the workflow-level field and the tier/alias preset path would accept different value sets for the same underlying SDK parameter, and nothing would flag the divergence.

**Evidence:** Four independent definitions, none deriving from another:
- `modelReasoningEffortSchema` (zod enum) — `packages/workflows/src/schemas/workflow.ts:20`
- `CODEX_REASONING_EFFORTS` (hand `Set`) — `packages/workflows/src/model-validation.ts:231-237`
- `validEfforts` (hand array) — `packages/providers/src/codex/config.ts:21`
- the literal union on `CodexProviderDefaults.modelReasoningEffort` — `packages/providers/src/types.ts:30`, whose comment says "Structurally matches @archon/workflows ModelReasoningEffort" — an unenforced claim, not a type link

`model-validation.ts` does not import `schemas/workflow.ts`. Before this PR `modelReasoningEffortSchema` was load-time-only with no runtime consumer, so drift could not produce a behavioural bug. `dag-executor.ts:1119-1121` now trusts the loader-validated value directly with no re-check against `CODEX_REASONING_EFFORTS`.

**Required outcome:** Derive `CODEX_REASONING_EFFORTS` from `modelReasoningEffortSchema.options` (the precedent AGENTS.md already sets for `TRIGGER_RULES` and `WORKFLOW_HOOK_EVENTS`). The `@archon/providers` copies sit across a package boundary that deliberately has no `@archon/workflows` dependency, so those stay — but the two inside `@archon/workflows` should not both exist.

**Disposition:** OPEN, non-blocking. Not required by this PR's outcome, but the PR is what makes the drift consequential.

</details>

<details>
<summary><code>R7</code> — one workflow-level spelling for reasoning depth, or two? Answer: one, as a follow-up</summary>

**Impact:** After this PR the workflow language has two workflow-level ways to say "reason harder": `effort:` (cross-provider, but `effortControl: false` on Codex so it warns and drops there) and `modelReasoningEffort:` (Codex, plus Copilot by accident). Picking wrong fails loudly in one direction and silently in the other. Under #1764 the confusion travels into composed workflows.

**The decision, with the evidence that moves it:**

**One spelling is right** — but three things in the brief's framing need correcting before the follow-up is designed.

1. *The brief's proposed mechanism does not work as sketched.* `routePresetEffort` handles only `claude` and `codex`; for Pi, Copilot and OpenCode it returns `null`, and `validEffortsForProvider` (`model-validation.ts:262-273`) documents that as deliberate: *"Pi/OpenRouter/Copilot/OpenCode — effort doesn't route there."* Routing workflow-level `effort:` through it **today** would regress Pi and Copilot, which currently consume workflow-level `effort:` fine via `nodeConfig.effort`. The router must first grow a passthrough for every `effortControl: true` provider.

2. *The enum-divergence argument against one spelling is weaker than it looks.* Copilot's parser already maps `max` → `xhigh` at parse time, with the comment *"Accept Archon's workflow-schema alias for the top tier"* (`community/copilot/config.ts:24-29`). The codebase already treats `max` and `xhigh` as the same rung, reconciled at the provider boundary — which is where a vocabulary difference belongs. That leaves `minimal` as the only genuinely Codex-only value: one extra rung at the bottom, not two incompatible vocabularies.

3. *The root cause is a mis-scoped capability axis, not a missing field.* `effortControl` means "accepts the node-level `effort:` field", not "can apply reasoning effort" — dev's own comment says so in as many words (`dag-executor.ts:1174-1177`: *"Codex is effortControl:false yet applies effort via modelReasoningEffort"*). Codex has reasoning depth; it reads it from a different place. The second spelling exists to route around that. Name the axis correctly (or translate at the provider boundary as Copilot already does) and the second spelling stops earning its keep.

**Scope note:** this applies to `modelReasoningEffort` only. `webSearchMode` is not a reasoning control, has no cross-provider equivalent, and has no other home in the workflow language — forwarding it is straightforwardly correct and should land.

**It does not belong in this PR.** Both fields are already documented workflow-level surface (`guides/authoring-workflows.md:126-127`, `book/quick-reference.md:112-113`) and schema-accepted since 2026-03-26; user YAML may already contain them. Deprecating a documented spelling needs its own deprecation path. This PR is the bug fix that makes the shipped docs true — land it (corrected), then consolidate.

**On the multiplication worry under #1764:** verified it does *not* multiply mechanically. `warnDroppedWorkflowLevelFields` (`include-expander.ts:483`) derives its dropped-field list generically from `Object.keys(child)` against a small allowlist of non-dropped keys, so any future workflow-level field is covered automatically, and the `workflow:` sub-run boundary does not forward parent workflow-level options to children at all today. The cost under composition is author confusion, not code duplication.

**Required outcome:** File a follow-up: extend `routePresetEffort` to cover every `effortControl: true` provider, widen `effortLevelSchema` to include `minimal` (keeping `max` ≡ `xhigh` reconciled at the provider boundary as Copilot already does), route workflow-level and node-level `effort:` through it, and deprecate `modelReasoningEffort` off `workflowBaseSchema` while keeping it as an `assistants.codex.*` config key. No such issue exists today (searched open and closed issues for `modelReasoningEffort` and `effort`).

**Found by:** `prp-core:code-reviewer` (constitution analysis), `prp-core:seam-analyzer` (two-spellings seam)

**Disposition:** OPEN, non-blocking. Not raised to `TRACKED FOLLOW-UP` because no verified issue exists yet and creating one is the maintainer's call.

</details>

<details>
<summary><code>R8</code> — this PR turns a "Codex only" docs label from imprecise into wrong</summary>

**Impact:** Three shipped user-facing doc locations tell an author `modelReasoningEffort:` is Codex-only. Before this PR the label was harmless because the field did nothing anywhere. After it, the field reaches Copilot too — so an author on `provider: copilot` who follows the documented Codex enum and writes `modelReasoningEffort: minimal` gets a silent no-op, while `low|medium|high|xhigh` silently *do* apply despite the label saying they cannot.

**Evidence:** The label appears at `packages/docs-web/src/content/docs/guides/authoring-workflows.md:126` (workflow-level schema block), `:1680` (the "### Codex-Specific Options" section at `:1674-1693`, which documents both fields and their enums in full), and `packages/docs-web/src/content/docs/book/quick-reference.md:112`. The behaviour that falsifies it: `packages/providers/src/community/copilot/provider.ts:462-463` parses `requestOptions.assistantConfig` — the same object `dag-executor.ts:1119-1122` now mutates without a provider gate — and `resolveCopilotReasoning` (`:127-163`) falls back to `copilotConfig.modelReasoningEffort` when the node sets neither `effort:` nor `thinking:`. `parseCopilotConfig` (`community/copilot/config.ts:20-30`) accepts only `low|medium|high|xhigh` (with `max` → `xhigh`) and silently omits anything else, so `minimal` dies there.

`webSearchMode` has no Copilot-side field in `providers/src/types.ts`, so that half of the label stays accurate.

**Required outcome:** Correct the `modelReasoningEffort` label at those three locations to note that Copilot also reads it as a fallback default, with a narrower enum that drops `minimal`. This is the documentation counterpart of R3 — if R3 is instead resolved by provider-gating the write to Codex, the docs need no change and this finding closes with it.

**Found by:** `prp-core:docs-impact-agent`

**Disposition:** OPEN, non-blocking. Coordinator-verified at all five cited locations.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | R1, R2, R3, R5, R7 |
| seams | R1, R2, R3, R6, R7 |
| docs | R8 (also contributed to R3) |
| validation | R4 |

**Docs surface, checked and clean** (recorded here rather than as findings): three locations — `guides/authoring-workflows.md:126-127`, its "### Codex-Specific Options" section at `:1674-1693`, and `book/quick-reference.md:112-113` — document `modelReasoningEffort:`/`webSearchMode:` as working workflow-level fields. They have been describing behaviour that did not exist, and this PR makes them true; that is the strongest argument for landing it. `authoring-workflows.md:249` does **not** wrongly promise per-node overrides for these two — "They" refers back to the eight-field Claude SDK Advanced Options set, and `modelReasoningEffort` appears only in a parenthetical aside. The apparent contradiction between `getting-started/ai-assistants.md:236-237` (`minimal|low|medium|high|xhigh`) and `:620` (`low|medium|high|xhigh|max`) is not a defect — the first is under "Codex Configuration Options" (`:228`), the second under "Copilot Configuration Options" (`:612`), and each matches its own parser exactly. One real completeness gap, below the reporting bar: the docs never state that workflow-level `effort:` is warned-and-dropped on Codex — `:249` scopes that mechanic to the six Claude-only fields — though the node-fields table at `:238` does correctly list `effort` as "Claude/Pi/Copilot".

## Validation

|---|---|---|
| `bun run validate` (PR head `35c5a9b3`) | PASS | exit 0; all 9 steps incl. `check:bundled*`, `check:pi-vendor-map`, `check:capability-matrix`; zero `N fail` lines with N>0 across all packages |
| `bun run type-check` (PR head) | PASS | all 10 packages exit 0 |
| `bun run lint` (PR head) | PASS | `eslint . --cache`, no output, `--max-warnings 0` |
| `bun run format:check` (PR head) | PASS | "All matched files use Prettier code style!" |
| `bun run test` — `packages/workflows` (PR head) | PASS | 1483 pass, 0 fail |
| `bun test src/dag-executor.test.ts` (merged with `dev`) | **FAIL** | 453 pass, **2 fail** — both of the PR's new tests; `mockSendQueryDag` never called (see R4) |
| `bun run type-check` (merged with `dev`) | PASS | all packages exit 0 — proves the positional-argument drift is invisible to the compiler |
| GitHub CI | NOT RUN | `gh api repos/coleam00/Archon/commits/35c5a9b3/check-runs` → zero check runs; only CodeRabbit "Review skipped: draft pull request" |

## Verdict

**NEEDS FIXES**

R1 is the blocker on its own: the PR fixes one silent drop by introducing another, on a path (mixed-provider workflows) the project is actively moving toward. R2 and R3 share its root cause — an unconditional write into `assistantConfig` with no provider gate — so all three are fixed by the same correction: scope the guard and the write to the provider that actually consumes the field, and warn when it is declared where it will not be consumed. R4 is mechanical but mandatory, and the rebase is where R2's fix has to be applied.

**These are contained fixes, not a redo.** The PR's core judgement is right — the bug is real, the Codex-only scope is correct, the "after `applyPresetOptions`" placement is the right precedence choice, and `webSearchMode` needs no consolidation debate. Roughly ten lines of correction plus a rebase and two tests separate this from mergeable. Given the branch is 96 commits stale and from an outside contributor running an automated pipeline, taking it over is a reasonable alternative to another round trip — but the design does not need replacing.

On the brief's architectural question: **one spelling, in a follow-up, not here** — see R7 for the reasoning and the concrete replacement design, including the two places the brief's proposed mechanism needs correcting before it would work.

Wirasm:
Thanks for this — and credit where it's due: you found a real bug and diagnosed it correctly. The shipped docs have been documenting `modelReasoningEffort:` and `webSearchMode:` as working workflow-level fields since March, and the executor never read them. Your Codex-only scoping was right, and placing the write after `applyPresetOptions` was the right precedence call.

I'm closing this rather than asking for another round, for two reasons.

**The branch is 96 commits behind `dev`, and the breakage is semantic rather than textual.** `#2299` inserted a `stateDir` positional parameter into `executeDagWorkflow`, so after merge every argument past `artifactsDir` shifts by one and both of your new tests fail. `git merge-tree` reports no conflict, and `bun run type-check` still passes because test files aren't in the type-check project — so nothing on the way in flags it.

**The self-fix pass introduced a regression.** The guard added to `applyPresetOptions` isn't scoped to the resolved provider, and `workflowLevelOptions` is workflow-scoped — so a workflow declaring `modelReasoningEffort:` for its Codex nodes now silently suppresses tier/alias preset effort on its Claude and Pi nodes too, with no warning on any channel. Worth noting your own review agent had flagged that change as low severity and recommended deferring it; the aggressive fix mode overrode that and it's what turned a non-issue into silent data loss.

Full review with evidence: https://github.com/coleam00/Archon/pull/2254#issuecomment-5314108113

Between the rebase and the corrections, roughly two lines of the diff survive, so I'm rebuilding it fresh against #2246 rather than force-pushing a rewrite onto your fork. Credit to you and to @QwertyCoolMT (#2020), who got here first with the same forwarding fix.

Issue #2213, opened 2026-07-21
bug: workflow validator silently accepts unknown node keys (e.g. interactive: true on a command node)


aweirdmaan:
## What happened

We shipped a workflow with a human-approval gate written as:

```yaml
- id: plan
  command: my-plan-command
  interactive: true          # human gate
```

`archon validate workflows` reported `ok`. At runtime the key was silently ignored (in v0.5.0, `interactive` is a *loop* field requiring `gate_message`; it means nothing on a command/prompt node). The workflow ran straight past its "gate" into implementation, unattended.

The dangerous part isn't the missing feature — it's that validation blessed a safety property that didn't exist. The same silent-acceptance applies to other misplacements (e.g. `fresh_context:` on a node vs `context: fresh`, an easy docs-version slip since README examples and released schemas drift).

## Suggested fix

Unknown or misplaced keys on a node should at minimum produce a validation **warning** naming the key and the node types it is valid for:

```
WARNING node 'plan': key 'interactive' is not valid for command nodes (valid for: loop). It will be ignored.
```

Zod `.strict()` on node schemas (or an explicit known-keys check with a friendly message) would catch this class entirely. A `--strict` flag on `archon validate` would be enough for CI users.

Archon v0.5.0 (brew binary), macOS. Happy to provide the full workflow file.

PR #2255, opened 2026-07-23
fix(workflows): warn on unknown/misplaced keys in workflow YAML (#2213)

kagura-agent:
## UX Journey

**Before:** A user writes `interactive: true` on a command node thinking it creates a human gate. `archon validate workflows` reports `ok`. At runtime, the key is silently ignored and the workflow runs unattended past what was meant to be an approval step.

**After:** `archon validate workflows` reports:
```
WARNING [unknown_key] Node 'plan': unknown key 'interactive' will be ignored
  ('interactive' is valid at workflow level, not on individual nodes)
```
The workflow still parses and runs (backwards compatible), but the user gets immediate feedback that their key was dropped.

## Architecture Diagram

```
YAML input → parseWorkflow() → parseDagNode() per node
                                    ↓
                              dagNodeSchema.safeParse() (Zod strips unknown keys)
                                    ↓
                              Compare raw keys vs KNOWN_DAG_NODE_KEYS
                                    ↓ unknown found?
                              Push warning (with WORKFLOW_ONLY_KEYS hint if applicable)
                                    ↓
                              ParseResult.warnings → WorkflowWithSource.parseWarnings
                                    ↓
                              CLI validate: convert to ValidationIssue (level: warning)
```

## Label Snapshot

- Package: `@archon/workflows`, `@archon/cli`
- Risk: LOW (additive — warnings only, no parsing behavior change)
- Breaking: NO

## Change Metadata

| Metric | Value |
|--------|-------|
| Files changed | 7 |
| Lines added | ~307 |
| Lines removed | ~9 |
| New tests | 5 |
| Existing test impact | 0 (all 179 pass) |

## Linked Issue

Refs #2213

## Validation Evidence

- `bun run type-check` — all 10 packages pass
- `bun test packages/workflows/src/loader.test.ts` — 179 pass, 0 fail
- 5 new tests covering: unknown node key, misplaced workflow-level key on node, unknown workflow-level key, valid keys produce no warnings, multiple nodes with unknown keys

## Security Impact

None. This is a read-only validation enhancement — no new I/O, no new dependencies, no config changes.

## Compatibility / Migration

Fully backwards compatible. Unknown keys were already silently stripped; this change only adds warnings. No workflow YAML changes required.

## Human Verification

- [x] Type check passes (`bun run type-check`)
- [x] All existing tests pass (179/179)
- [x] New tests cover the exact scenario from the issue (`interactive: true` on a command node)
- [x] Warnings include actionable hints for misplaced keys

## Risks and Mitigations

| Risk | Mitigation |
|------|-----------|
| Known key set drifts from schema | Constants are co-located with their schemas (dag-node.ts, workflow.ts) with sync comments |
| Bundled defaults trigger warnings | Unlikely — bundled workflows are authored by maintainers; if they do trigger, it surfaces a real issue |

- `ParseResult` type gains a `warnings` field (additive, non-breaking)
- `WorkflowWithSource` gains optional `parseWarnings` field (additive, non-breaking)
- No runtime behavior change — only `archon validate` output changes

## Rollback Plan

Revert the commit. No data migration, no state change.

---

🤖 **Disclosure:** This PR was authored by [Kagura](https://github.com/kagura-agent), an AI agent. Open source contribution is one of the things I do — you can see my work history [here](https://github.com/kagura-agent/github-contribution). If you'd prefer not to receive AI-authored PRs, just let me know and I'll stop — no hard feelings.

## Summary by CodeRabbit

* **New Features**
  * Workflow validation now surfaces non-fatal warnings for unknown or misplaced YAML keys (including workflow-root vs node-level hints) in output and JSON.
  * Parse-time warnings are collected during loading and propagated through workflow discovery, including include expansion.
* **Bug Fixes**
  * Non-fatal parse warnings are no longer silently dropped, and override precedence is respected when merging warnings.
* **Tests**
  * Added coverage for recording and accumulating unknown-key parse warnings.
* **Chores**
  * Updated several E2E workflow scripts to make shell assertions clearer and more consistent.

### DIFF
--- a/.archon/workflows/e2e-opencode-inline-multi-agents.yaml
+++ b/.archon/workflows/e2e-opencode-inline-multi-agents.yaml
@@ -41,9 +41,10 @@ nodes:
 
   - id: assert
     bash: |
-      echo "$multi.output" | grep -q "FIRST_MULTI_AGENT_OK" \
-      && echo "$multi.output" | grep -q "SECOND_MULTI_AGENT_OK" \
-      && echo "$inline.output" | grep -q "INLINE_AGENT_OK" \
+      v=$multi.output
+      echo "$v" | grep -q "FIRST_MULTI_AGENT_OK" \
+      && echo "$v" | grep -q "SECOND_MULTI_AGENT_OK" \
+      && v2=$inline.output && echo "$v2" | grep -q "INLINE_AGENT_OK" \
       && echo "PASS: both agents in multi node executed parallel, inline agent verified" \
       || (echo "FAIL: multi/inline agent output assertion failed"; exit 1)
     timeout: 60000
--- a/.archon/workflows/e2e-opencode-smoke.yaml
+++ b/.archon/workflows/e2e-opencode-smoke.yaml
@@ -14,6 +14,8 @@ nodes:
     idle_timeout: 60000
 
   - id: assert
-    bash: "echo \"$simple.output\" | grep -q \"OPENCODE_OK\" && echo \"PASS\" ||
-      (echo \"FAIL\"; exit 1)"
+    bash: |
+      v=$simple.output
+      echo "$v" | grep -q "OPENCODE_OK" && echo "PASS" \
+      || (echo "FAIL"; exit 1)
     depends_on: [ simple ]
--- a/.archon/workflows/test-workflows/e2e-copilot-all-nodes-smoke.yaml
+++ b/.archon/workflows/test-workflows/e2e-copilot-all-nodes-smoke.yaml
@@ -113,11 +113,6 @@ nodes:
   # 12. Verify every upstream node produced non-empty output, including
   #     dot-access on the structured-output node (proves output_format
   #     parsed and downstream consumers can index into it).
-  #     Note: value-equality on string fields is avoided on purpose —
-  #     shellQuote() wraps strings in literal single quotes, so a literal
-  #     `[ "$x" != "ok" ]` would always fail. Non-emptiness is the right
-  #     bar for a smoke; the `when:` gate on structured-check already
-  #     proved the value matched 'ok' to reach this node.
   - id: assert
     bash: |
       fail=0
@@ -129,17 +124,28 @@ nodes:
           fail=1
         fi
       }
-      check prompt-node       "$prompt-node.output"
-      check command-node      "$command-node.output"
-      check loop-node         "$loop-node.output"
-      check bash-json-node    "$bash-json-node.output"
-      check script-bun-node   "$script-bun-node.output"
-      check script-python-node "$script-python-node.output"
-      check downstream        "$downstream.output"
-      check gated             "$gated.output"
-      check merge             "$merge.output"
-      check structured.status "$structured-node.output.status"
-      check structured.value  "$structured-node.output.value"
+      v=$prompt-node.output
+      check prompt-node        "$v"
+      v=$command-node.output
+      check command-node       "$v"
+      v=$loop-node.output
+      check loop-node          "$v"
+      v=$bash-json-node.output
+      check bash-json-node     "$v"
+      v=$script-bun-node.output
+      check script-bun-node    "$v"
+      v=$script-python-node.output
+      check script-python-node "$v"
+      v=$downstream.output
+      check downstream         "$v"
+      v=$gated.output
+      check gated              "$v"
+      v=$merge.output
+      check merge              "$v"
+      v=$structured-node.output.status
+      check structured.status  "$v"
+      v=$structured-node.output.value
+      check structured.value   "$v"
 
       if [ "$fail" -eq 1 ]; then exit 1; fi
       echo "PASS: all node types + structured output verified"
--- a/.archon/workflows/test-workflows/e2e-pi-all-nodes-smoke.yaml
+++ b/.archon/workflows/test-workflows/e2e-pi-all-nodes-smoke.yaml
@@ -90,15 +90,24 @@ nodes:
           fail=1
         fi
       }
-      check prompt-node "$prompt-node.output"
-      check command-node "$command-node.output"
-      check loop-node "$loop-node.output"
-      check bash-json-node "$bash-json-node.output"
-      check script-bun-node "$script-bun-node.output"
-      check script-python-node "$script-python-node.output"
-      check downstream "$downstream.output"
-      check gated "$gated.output"
-      check merge "$merge.output"
+      v=$prompt-node.output
+      check prompt-node        "$v"
+      v=$command-node.output
+      check command-node       "$v"
+      v=$loop-node.output
+      check loop-node          "$v"
+      v=$bash-json-node.output
+      check bash-json-node     "$v"
+      v=$script-bun-node.output
+      check script-bun-node    "$v"
+      v=$script-python-node.output
+      check script-python-node "$v"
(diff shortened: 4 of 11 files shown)


kagura-agent:
Both findings were already addressed in commit 9cf2f22:
- **Node-only key hint**: `KNOWN_DAG_NODE_KEYS.has(key)` check added (loader.ts L777-779)
- **Stale warning clearing**: `mergeWarnings` now deletes old entries for overridden clean files (workflow-discovery.ts L335-340)

No additional changes needed.

Wirasm:
This is a real gap and the threading through `ParseResult` → `DirLoadResult` → `WorkflowWithSource` → `archon validate workflows` is clean. I verified it's genuinely fail-safe — `makeWorkflowResult` keeps `valid: true` and the command still exits 0, so no currently-valid workflow becomes invalid. And running your key sets across all 54 workflow-shaped YAMLs in the repo surfaced exactly one hit: `agent: general` at both workflow and node level in `e2e-opencode-smoke.yaml`, which really is dead config (`agent` is only ever read from `assistants.opencode.opencode.agent`). Nice demonstration that it does the job.

**Two things before merge.**

**1. Derive the key sets from the schemas instead of hand-listing them.**

I checked the current lists and they are exactly right today — `Object.keys(workflowDefinitionSchema.shape)` gives 20 and the symmetric difference with `KNOWN_WORKFLOW_KEYS` is empty in both directions; the node side is 29 base keys plus the 16 `.extend()` keys, all 45 present. So there are zero false positives right now.

The problem is what happens next. The moment someone adds a node field, every workflow using that field starts emitting "unknown key … will be ignored" — on a valid workflow, on every discovery, every run, every `/workflow list`, every server boot. A `// Keep in sync with…` comment is not a guard, and nothing in CI catches the drift. For a validator, failing open into false warnings on correct input is the worst available failure mode: people learn to ignore the output, and then it stops working as a validator at all.

Both sets are derivable in one line each, and I confirmed both by execution:

- `new Set(Object.keys(workflowDefinitionSchema.shape))` — plain ZodObject, directly accessible.
- For nodes, zod v4 wraps `dagNodeSchema` in a `pipe`, but `(dagNodeSchema as ...).def.in.shape` returns exactly the 45 keys you hand-listed.

`WORKFLOW_ONLY_KEYS` then computes as the workflow set minus the node set. If reaching into zod internals feels too fragile — a fair objection — naming the pre-`superRefine` object and exporting its `.shape` is just as good. A parity test asserting the constants equal the schema shapes would also be acceptable, though I'd rather not maintain the list at all.

**2. The `interactive` hint points the wrong way, and it misdirects exactly the person who filed #2213.**

`loader.ts:132-134` says `'interactive'` is valid at workflow level, not on individual nodes. True as far as it goes — but `interactive` means three different things in three places: at workflow level it forces foreground execution on web; `loop.interactive` is the human gate (needing `loop.gate_message`); `pi.interactive` is UIContext binding. The reporter in #2213 wanted a **gate**. Following this hint moves the key to workflow level and still produces no gate — they end up somewhere else that also doesn't work.

The hint needs to point at `loop.gate_message` / an `approval:` node. Either generalise the `WORKFLOW_ONLY_KEYS` hinting to carry a per-key suggestion, or special-case this one.

**Follow-ups — not blocking, and I'm not asking you to take them on unless you want to:**

- `loop_group.nodes[]` body nodes bypass `parseDagNode` entirely (they're validated by `z.array(dagNodeSchema)` inside `loopGroupNodeConfigSchema`), so unknown keys inside a loop-group body still slip through silently. Same for sub-keys inside `loop:`/`approval:`/`retry:` blocks. That's where complex node config actually gets written, so it's the highest-value extension.
- `parseWarnings` stops at the CLI — `GET /api/workflows` and the workflow builder both drop it, so a builder user never sees the warning.
- #2213 also asked for per-node-type validity in the message ("valid for: loop") and a `--strict` flag for CI. Neither is here, and neither was promised — but that means `Fixes #2213` will auto-close an issue that's only partly addressed. Please switch it to `Refs #2213` and I'll close it manually once the remainder is either filed or done.
- Minor: `steps: []` slips past the legacy-format rejection and then warns "unknown key 'steps'", which reads oddly.

One thing to coordinate with yourself: if #2262 lands without removing `agent:` from `e2e-opencode-smoke.yaml`, this PR makes our own `validate workflows` start warning. Same author on both, so easiest to just drop the key there.

kagura-agent:
**Addressed both blocking items in c4910549:**

**1. Key sets derived from schema shapes**

- Extracted the pre-`superRefine` intermediate as `dagNodeFlatSchema` (exported from `dag-node.ts`)
- `KNOWN_DAG_NODE_KEYS` is now `new Set(Object.keys(dagNodeFlatSchema.shape))`
- `KNOWN_WORKFLOW_KEYS` is now `new Set(Object.keys(workflowDefinitionSchema.shape))`
- `WORKFLOW_ONLY_KEYS` computed as the set difference: `[...KNOWN_WORKFLOW_KEYS].filter(k => !KNOWN_DAG_NODE_KEYS.has(k))`

No hand-maintained lists remain. Adding a field to any schema automatically updates the key sets.

**2. Interactive hint fixed**

Node-level `interactive` now produces:

```
Node 'plan': unknown key 'interactive' will be ignored
  ('interactive' at workflow level forces foreground execution; for a human gate
  on this node, use 'loop.gate_message' or an 'approval:' node)
```

Points the user toward the gate mechanism they likely wanted, instead of the generic "valid at workflow level" misdirection.

**3. PR description updated** — `Fixes #2213` → `Refs #2213`.

All 180 loader tests + 103 schema tests pass, type check clean.

kagura-agent:
Hi @Wirasm — friendly follow-up on the requested fixes in `c4910549`. The PR remains mergeable and I’m happy to address any remaining feedback. Thank you!

kagura-agent:
Rebased onto the latest `dev`; the PR is now conflict-free. The unknown-key warning behavior remains intact.

Wirasm:
Thanks for this — closing here because the work is carried forward in #2455, with your four commits intact and still authored by you:

- `6a0034f` fix(workflows): warn on unknown/misplaced keys in workflow YAML (#2213)
- `43ca32b` fix: remove unnecessary type assertion (lint)
- `5dd8c00` fix: address CodeRabbit review — reverse hint + warning override cleanup
- `d845ce5` fix: derive key sets from schema shapes, improve interactive hint

They sit at the base of that branch. Git records you as author and the rebase only changed the committer, so the attribution survives the merge.

Your `d845ce5` — deriving the key sets from the schema shapes rather than hand-listing them — is the part the rest was built on. It's what makes the whole surface drift-proof, and it set the shape for everything added on top.

What #2455 adds:

- Detection one level down into nested config, which is where the `agents.<id>.disallowed_tools` camelCase slip lives (a silent tool-restriction drop)
- Warnings surfaced on the paths authors actually use — `validate`, `list`, before a CLI `run`, chat `/workflow list`, and the console picker — rather than only in the parse result
- Warnings at the moment a run **starts** from chat or the console, which was the gap that mattered most: the console showed a ⚠ badge while browsing and then went silent at the point of consequence
- A fix for a bug we found in our own additions: parse warnings were kept in a map parallel to the workflow map, and only one of the two was cleared on override, so a root/subfolder filename collision could pin a warning onto the clean file that won

Sorry it sat for two weeks. The root-cause fix is yours; the rest is surfacing it.

PR #2256, opened 2026-07-23
marketplace: add epic-piv-ticket (gated per-ticket dev workflow)

TheSmokeDev:
Adds **epic-piv-ticket** to the marketplace registry — a per-ticket development workflow that turns one well-scoped GitHub issue into a fully-gated draft PR:

- Fresh-context worker implements the ticket and records its exact test command
- A deterministic test gate **re-executes the recorded suite** (self-reports count for nothing)
- An adversarial `codex exec` implementation review gates the diff (fail-closed script verdicts, exit 0/1/2)
- Autonomous reconcile rounds in fresh context (hard round caps) — findings verified against code, fixed with regression tests, or refuted with receipts
- An independent **design review on any OpenAI-compatible model** (configurable endpoint/model/key env) gates the reconciled head
- Opens a draft PR carrying the full gate trail. **Never merges — the human does.**

Package: https://github.com/TheSmokeDev/epic-piv-ticket (pinned at `fef6b174`) — workflow YAML (9-node DAG on the proven print-token loop contract), two command briefs, two gate scripts, 34 offline tests (subprocess/HTTP mocked).

Proven before listing: in its first fully-autonomous run the implementation gate returned FAIL with real findings — including in the gate tooling itself — and the workflow reconciled in a fresh context, re-passed both gates, and opened the PR. Requires Archon >=0.6.0 (loop completion-signal contract), `codex` CLI, `gh`, and an OpenAI-compatible design-gate endpoint.

Same entry shape as my previous listing (image-node-factory, #2077).


Wirasm:
This is not a marketplace entry, please fix the pr and resubmit, check CRONTRUBITION.md for guidelines

PR #2235, opened 2026-07-21
feat(workflows): evidence gate — refuse terminal completed when evidence_policy.required and evidence.json absent

Wirasm:
## Summary

- Problem: a workflow run can report terminal `completed` even when its real-world outcome never landed (motivating case #1720 — a commit node "succeeds" in <15s with no commit).
- Why it matters: `completed` is the engine's governance signal; workflows that produce verifiable proof need a way to make that proof a hard condition of success.
- What changed: optional workflow-level `evidence_policy: { required: boolean }`. When `required: true`, the DAG executor refuses terminal `completed` unless `$ARTIFACTS_DIR/evidence.json` exists — the run is marked `failed` with a structured note at `metadata.evidence_validation`, an `evidence_validation_failed` workflow event, and an explicit failure message naming the expected path.
- What did **not** change (scope boundary): the engine gates on file PRESENCE only. No typed evidence schemas, no validator layers, no `verify:` modes, no configurable paths, no git/gh I/O in the engine — all explicitly deferred by #2230. What counts as valid evidence is produced by the workflow's own bash/script nodes.
- Provenance: design carved from @shhaider's closed PR #1601 — this is the thin admissible slice per the workflow-language constitution (the reality-verification layer was computation, not coordination). Credited via `Co-authored-by`.

### Before

```
  Author                 Archon engine
  ──────                 ─────────────
  runs workflow ───────▶ all nodes succeed
                         run flips to `completed`
  sees "completed" ◀───  (even if the commit/PR/deliverable never landed)
```

### After

```
  Author                 Archon engine
  ──────                 ─────────────
  declares
  evidence_policy:
    required: true
  runs workflow ───────▶ all nodes succeed
                         [gate: $ARTIFACTS_DIR/evidence.json exists?]
                         ├─ yes → run flips to `completed` (unchanged)
                         └─ no  → run marked `failed`
                                  * metadata.evidence_validation note
                                  * evidence_validation_failed event
                                  * message names the expected path
  produces evidence,
  resumes the run ─────▶ nodes skip as prior-completed
                         gate re-checks → `completed`
```

### Before

```
loader.ts ──parses──▶ workflowBaseSchema
dag-executor.ts (completion path):
  !anyCompleted → fail
  anyFailed     → fail
  skipIfStatusChanged
  container write-back gate ──▶ completeWorkflowRun
```

### After

```
loader.ts ──parses──▶ workflowBaseSchema [~ + evidence_policy]
dag-executor.ts (completion path):
[...]

Issue #2230, opened 2026-07-21
feat(workflows): thin terminal-success gate — refuse completed when evidence_policy.required and evidence.json absent


Wirasm:
MVP carved from closed PR #1601 (@shhaider), whose full 2100-LOC surface (typed evidence schema + 3-layer validator + `verify: 'reality'` git/gh subprocess I/O inside the engine) was declined per the workflow-language constitution: the reality-verification layer is computation, not coordination (§workaround-triage — a final `bash:`/`script:` node that computes the proof and exits non-zero is the sanctioned escape hatch), and the GitHub-PR-shaped evidence schema hard-codes one domain into a platform-agnostic engine (§schema-width).

What IS admissible — the thin gate, a sibling of `approval:` (run-status transitions are engine governance):

- Optional workflow-level `evidence_policy: { required: boolean }`.
- When `required: true`, the DAG executor refuses terminal `completed` unless `$ARTIFACTS_DIR/evidence.json` exists; missing → run marked `failed` with a structured note at `metadata.evidence_validation`.
- What counts as valid evidence is produced by the workflow's own nodes (script/bash compute the proof); the engine only gates on presence.

**Out of scope** (revisit only if the gate sees adoption): typed evidence schemas, the validator layers, any git/gh I/O in the engine.

Motivating case: #1720 (commit node reports success in <15s with no commit landed). Implementation note: anchor onto the current completion path in `dag-executor.ts` (`completeWorkflowRun`, ~line 6703 — now shared with the container write-back gate). Est. ~150-200 LOC + tests.

PR #2257, opened 2026-07-23
Stop archon-fix-issue asking questions nobody can answer

recepp4200:
## Problem and outcome

`archon-fix-issue` told the agent to STOP and ask the user whenever it hit a design ambiguity the artifact didn't resolve. That is reasonable advice in an interactive session and a guaranteed no-op under CLI or scheduled dispatch: a `command:` node has no channel to a human mid-execution, so the question becomes the node's output, the node reports `completed`, and the DAG advances having implemented nothing. @recepp4200 reproduced it three consecutive times against a real issue before diagnosing it.

- **Outcome:** the STOP-and-ask branches now make the best available decision and record it in the implementation report's "Deviations from Investigation" section — the mechanism that actually surfaces a judgment call to a human. A run that hits ambiguity produces a reviewable diff plus a documented deviation instead of an unread question.
- **Invariant:** deciding replaces asking only where there is real input to adapt. Missing input still stops the run loudly.
- **Scope boundary:** prompt text only — no engine change, no schema change, no other command file audited. The `dag-executor.ts` / `condition-evaluator.ts` half of the original PR is dropped (see below).
- **Root cause:** an AI node returns `{ state: 'completed' }` whenever the SDK query finishes without throwing (`packages/workflows/src/dag-executor.ts:2654`), regardless of what the model said. Pauses originate only from `approval:` nodes, loop gates, child-run gates and the container write-back gate. So there is no execution path by which "STOP and ask" reaches anyone.

## Review guidance

- **Feedback requested:** whether §1.3's gate is the right shape — it is the one place this PR deliberately does *not* apply the decide-and-document rule.
- **Start here:** `.archon/commands/defaults/archon-fix-issue.md` §1.3 — everything else in the diff loosens a stop; this one keeps it.
- **Review order:** commit `b84024a7` (the accepted change, authored by @recepp4200) then `3f11ddb0` (the §1.3 gate).
- **Lower-attention areas:** `packages/workflows/src/defaults/bundled-defaults.generated.ts` is generated — `bun run check:bundled` passes.
- **Known risk or uncertainty:** prompt guidance is advisory, not enforced. The durable fix for "run reports completed having produced nothing" is `evidence_policy: { required: true }` (#2230/#2235), which no bundled workflow opts into yet. Worth a follow-up.

## Solution

Four branches change from "stop and ask" to "decide and document": the unattended note in the mission statement, §2.1 plan drift, §2.2 "plan seems wrong", and the outdated-artifact edge case. In each, the artifact exists and the agent is adapting real input, so proceeding is defensible and the Deviations section is where the judgment call becomes visible.

§1.3 is the exception and moves the other way. With no artifact there is nothing to adapt; implementing anyway would produce a plausible-looking diff and a draft PR assembled from guesses, which is worse than stopping because the reviewer cannot see that the plan was missing. It now proceeds only once issue context has actually loaded — an issue number that `gh issue view` resolves, or the body already inlined in the prompt — and otherwise writes a BLOCKED implementation report and ends without touching source.

Since #2500 the bundled `archon-fix-github-issue` workflow's `bridge-artifacts` node already exits non-zero when neither `investigation.md` nor `plan.md` exists, so §1.3 is reachable only through direct invocation or a workflow without that guard.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | ambiguity → question as node output → `completed` → no diff → no PR | ambiguity → decision implemented → deviation documented in the implementation report |
| **Failure behavior** | missing artifact → error text, nothing written, run fails somewhere downstream | missing artifact + no issue context → BLOCKED implementation report, no source touched; with issue context → proceeds from the issue body, deviation recorded |

## Validation

- `bun run validate` — exit `0` — covers `check:bundled`, type-check, lint, format, and the full per-package suite.
- Root cause confirmed by reading the execution path rather than by rerunning a paid workflow: `dag-executor.ts:2654` returns `completed` on any non-throwing AI node, and no pause originates outside `executeApprovalNode`, the loop/child gates, and the write-back gate.
- The four `.md` hunks applied to current `dev` with `git apply -3` cleanly despite `dev` having moved 82 lines in this file since the original base.
- **Not verified:** no end-to-end rerun of `archon-fix-github-issue` against a fresh ambiguous issue. Prompt-behavior changes are not unit-testable here, and the failure mode being removed is a wasted run, not a corruption.

## What changed from the original PR

The original branch was based on `dev` from 2026-05-18 and also modified `condition-evaluator.ts`, `dag-executor.ts` and their tests. That area was rewritten underneath it: `OutputRefParseError` no longer exists, the replacement `OutputRefError` deliberately propagates so the node **fails** instead of skipping silently, and the exact misleading message reported has been replaced by an accurate one that names the upstream node. Reproduced on current `dev`: the reported expression now throws `OutputRefError` with *"node 'review-classify's output is not a JSON object, so the field cannot be read…"*. That half is superseded and has been dropped; only the `.md` change carried forward.

One finding from it is still live and deserves its own PR: `condition-evaluator.ts:184-189` returns `parsed: false` when a numeric comparison's operand isn't a finite number, so `dag-executor.ts:7583` reports "check syntax" for what is a data problem — exactly the complaint, on a different branch than the one patched.

## Links

- Related #2230, #2235, #2500

## Summary by CodeRabbit

- **Documentation**
  - Improved unattended execution guidance for making defensible decisions without pausing for user input.
  - Decisions, deviations, and adaptations are now recorded in implementation reports.
  - Clarified handling for missing or outdated artifacts, including when work should proceed, adapt to current code, or stop with a blocked record.
  - Added guidance for responding to plan drift and incorrect implementation assumptions.


Wirasm:
This is one of the better bug reports we've had — real production logs, an explicit "what was NOT verified" section, a stated scope boundary. The trace showing `review-classify` returning rate-limit prose instead of JSON was the genuinely useful part, even where the fix itself doesn't land. There are two separate changes here and they have different outcomes.

## The condition-evaluator half is superseded — please drop it

Your branch is based on `dev` from 2026-05-18, 256 commits back, and that area was rewritten underneath you:

- `OutputRefParseError`, the class your `catch` tests against, **no longer exists anywhere in the repo**. It was replaced by `OutputRefError` in `output-ref.ts`.
- More importantly, we no longer catch it. `condition-evaluator.ts:181-186` deliberately lets it propagate so the node **fails** instead of silently skipping, with a comment saying exactly that: *"under the no-silent-drop contract it must propagate to fail the node, not fail-closed to a silent skip."*
- The misleading message you hit is already gone. `output-ref.ts:69-71` now produces *"node 'review-classify's output is not a JSON object, so the field cannot be read. Emit JSON containing 'field', or reference '$review-classify.output' (whole text) instead"* — and fails the node.
- `condition-evaluator.test.ts:511-520` asserts `toThrow(OutputRefError)` on the same expression your new test expects a return value from. Both cannot pass.
- `atomPattern` also grew to 6 capture groups (shorthand `$node.field`, unquoted numeric/boolean RHS); your destructure takes 5.

So this isn't a rebase — it would be a rewrite against the design that replaced the one you patched. Please drop `condition-evaluator.ts`, `condition-evaluator.test.ts`, and `dag-executor.ts` from the PR.

**One thing you found is still real**, and it deserves its own PR: the numeric branch at `condition-evaluator.ts:193-198`. When `parseFloat` on the left side yields `NaN` we still return `parsed: false`, and the executor reports "check syntax" — a data problem wearing a syntax error's message, which is precisely your complaint. `$gate.output > '5'` on prose output, or a declared field holding `"high"`, both hit it. Worth noting the correct label there is the **inverse** of what this PR assigns (it marks that branch `'syntax'`); CodeRabbit flagged the same thing. A focused PR on just that would be welcome.

## The `archon-fix-issue.md` half we want

Your diagnosis is correct and it's a real bug: a `command:` node cannot receive a mid-execution answer — it completes and the DAG advances — so "STOP and ask the user" is a guaranteed no-op. I verified the cross-reference too: "Deviations from Investigation" does exist at line 381, so the new instructions point somewhere real. And that file is **byte-identical** between your base and current `dev`, so those four hunks cherry-pick cleanly.

**One change needed before it lands: the §1.3 "artifact not found" branch.**

As written it tells the agent to do the investigation itself and implement anyway. That removes the last guardrail on missing input. Today: artifact missing → `implement` errors → nothing written → `create-pr` refuses to open an empty PR → the run fails **loudly**. After: the command accepts either an issue number *or* an artifact path, so with a bad artifact path there may be no issue number for `gh issue view` and no issue body in `$CONTEXT` — and we get a plausible-looking draft PR built from guesses. That's a strictly worse failure than the loud stop, and the opposite of fail-fast.

Required shape: proceed **only** once issue context has actually loaded (an issue number is available and `gh issue view` succeeds, or `$CONTEXT` demonstrably carries the body); otherwise write a blocked result to the implementation artifact and stop. The §2.1 drift, §2.2 bad-plan, and edge-case hunks are good as-is — the artifact exists in those cases, so the agent is adapting real input and "decide + document in Deviations" is a defensible default.

**Suggested path**: fresh branch off current `dev` with just the four `.md` hunks plus the §1.3 gate, then `bun run generate:bundled` (dev's bundle has moved 21 lines since your base) and `bun run validate`.

Heads up that **no CI has ever run on this PR** — first-time contributors need a maintainer to approve workflow runs, so everything in your validation section is self-reported and was necessarily produced against the stale base.

## The durable version, if you want a second contribution

We merged an engine-level fix for the underlying "run reports completed having produced nothing" problem: `evidence_policy: { required: true }` refuses to flip a run to `completed` without `$ARTIFACTS_DIR/evidence.json` (#2230/#2235). `archon-fix-github-issue.yaml` doesn't opt into it yet. Wiring that up would make the failure structurally impossible rather than merely discouraged by prompt text — a better fix than either half of this PR.

Wirasm:
@recepp4200 — rather than leave this waiting on you, I've done the revision myself and pushed it to your branch (`maintainerCanModify` was on). Your original commit is `f2db9867` if you want it back; it's still reachable from this PR's timeline.

The branch is now two commits on current `dev`:

- **`b84024a7`** — your four `.md` hunks, authored to you, unchanged from what you wrote. They applied to today's `dev` with `git apply -3` cleanly even though the file has moved 82 lines since your base.
- **`3f11ddb0`** — the §1.3 gate I asked for. Missing artifact now proceeds **only** once issue context has actually loaded (an issue number `gh issue view` resolves, or the body already inlined in the prompt), and otherwise writes a BLOCKED implementation report and stops without touching source.

The condition-evaluator half is dropped, as discussed. I re-ran your reported case against current `dev` to be sure rather than take my own earlier word for it, and it now throws instead of skipping:

```
OutputRefError: '$review-classify.output.run_error_handling' references field
'run_error_handling', but node 'review-classify's output is not a JSON object,
so the field cannot be read. Emit JSON containing 'run_error_handling', or
reference '$review-classify.output' (whole text) instead.
```

So the misleading "check syntax" message you hit is genuinely gone on that path, and the node fails loudly instead of silently skipping.

**Your other finding is still live**, and I confirmed it today rather than assuming it:

```
evaluateCondition("$gate.output > '5'")  where $gate.output is prose
  → { result: false, parsed: false }
  → "⚠️ unparseable `when:` expression … Check syntax"
```

`condition-evaluator.ts:184-189` returns `parsed: false` when a numeric comparison's operand isn't a finite number, and `dag-executor.ts:7583` renders that as a syntax error. Same complaint you filed, different branch than the one your patch touched — and note the correct label there is the inverse of what the patch assigned. A focused PR on just that would be welcome, and it would be entirely yours.

One correction to something I said last time, since it shaped the §1.3 requirement: #2500 landed on 2026-08-06 and added a `bridge-artifacts` guard to the bundled `archon-fix-github-issue`, which exits non-zero when neither `investigation.md` nor `plan.md` exists. So in the bundled path the run already stops before `archon-fix-issue` ever sees a missing artifact. §1.3 is now reachable only through direct invocation or a workflow without that guard — which is why keeping the stop there is cheap, and why the new text says so explicitly.

`bun run validate` passes with exit 0. Thanks for the report — the log trace showing `review-classify` returning rate-limit prose was what made the whole thing diagnosable, and the "what was NOT verified" section is a habit I wish more PRs had.

PR #2261, opened 2026-07-23
feat(isolation): pluggable worktree engine - support worktrunk (wt)

### LINKED ISSUE #2260: feat(isolation): pluggable worktree engine - support worktrunk (wt) as an alternative to raw git worktree
## Summary

Proposal: make the low-level worktree plumbing in the isolation system pluggable behind a small "worktree engine" interface, with two implementations:

- `git` (default) - exactly today's behavior, raw `git worktree` commands
- `worktrunk` - shells out to [worktrunk](https://github.com/max-sixty/worktrunk) (`wt`), a git worktree manager built specifically for parallel AI agent workflows

Selected per repo via `.archon/config.yaml > worktree.engine` (default `git`, so zero behavior change for existing users).

## Motivation

Archon already treats worktree isolation as a first-class concern, and `IIsolationProvider` in `packages/isolation/src/types.ts` explicitly anticipates future strategies. But many users who run coding agents in parallel already manage their worktrees with worktrunk: it gives them per-repo hooks (e.g. env setup / dependency install on worktree creation, validation before merge), gitignored-file copying, a consistent path layout, and `wt list` / `wt merge` workflows.

Today, worktrees created by Archon are invisible to that tooling: user hooks don't run, layouts diverge, and `wt list` doesn't show what Archon is doing. For users with worktrunk-centric setups, letting Archon create/remove worktrees through `wt` means one consistent worktree world across manual work and Archon-dispatched runs.

## Proposed design

Deliberately NOT a new `IIsolationProvider`. `WorktreeProvider` contains a lot of Archon-specific logic (branch naming per workflow type, fork-PR fetching, adoption, DB registry integration, copyFiles, submodules) that must behave identically regardless of what creates the worktree. Duplicating that in a parallel provider would be a maintenance hazard.

Instead, extract the narrow git-plumbing surface that `WorktreeProvider` and `packages/git/src/worktree.ts` use into a small engine interface, roughly:

```ts
interface WorktreeEngine {
  add(opts: { repoPath; worktreePath; branch; startPoint }): Promise<void>;
  remove(opts: { repoPath; worktreePath; force }): Promise<void>;
  list(repoPath): Promise<WorktreeInfo[]>;
  prune(repoPath): Promise<void>;
}
```

- `GitWorktreeEngine`: the current code, moved behind the interface. Default. No behavior change.
- `WorktrunkEngine`: maps to `wt` invocations, e.g. `wt switch --create <branch> --base <base> --no-cd --yes -C <repo>` for add, `wt remove <branch> --foreground` for remove, `wt list --format json` for list. Worktree destination stays under Archon's control (pinned via worktrunk config override per invocation) so the environment registry and path resolution are unaffected.

Config:

```yaml
worktree:
  engine: worktrunk   # default: git
```

Failure mode: if `engine: worktrunk` is set but `wt` is not on PATH (or is older than a minimum supported version), worktree creation fails fast with an actionable error naming the missing binary - no silent fallback to git.

## Non-goals (v1)

- No change to branch naming, the resolver, adoption, cleanup logic, or the DB schema
- Fork-PR checkout (`refs/pull/N/head`) stays on the git path even under the worktrunk engine
- No worktrunk-specific features surfaced in Archon's UI (hooks etc. just run as the user configured them in worktrunk)

## Questions for maintainers

1. Is a pluggable engine at this layer acceptable, or would you prefer this scoped differently (e.g. behind an experimental flag)?
2. Preference on config key naming (`worktree.engine` vs `worktree.driver`)?

I have a working knowledge of both codebases and am happy to implement this and send a PR if the direction sounds good.

AlienClubrider:
## Summary

- Problem: Archon's worktree creation/removal is hardcoded to raw `git worktree` commands, so users who manage their worktrees with [worktrunk](https://worktrunk.dev) get a divergent, invisible-to-`wt-list` set of worktrees whenever Archon creates one.
- Why it matters: Users with a worktrunk-centric setup lose their per-repo hooks (env setup, dependency install, cleanup) and a consistent `wt list`/`wt merge` view whenever Archon (not the user) creates the worktree.
- What changed: Extracted the low-level worktree plumbing (`add`/`remove`/`list`/`prune`) behind a small `WorktreeEngine` interface with two implementations — `GitWorktreeEngine` (default, the historical behavior moved verbatim) and `WorktrunkEngine` (shells out to `wt`). Selected per repo via `.archon/config.yaml > worktree.engine` (default `'git'`).
- What did **not** change (scope boundary): `IIsolationProvider`, branch naming, the resolver, adoption logic, the DB schema, and fork-PR checkout (`refs/pull/N/head` always uses raw git — no worktrunk equivalent for synthetic review branches). No new worktrunk-specific UI surface; hooks just run as the user configured them.

### Before

```
  User                         Archon                        git
  ────                         ──────                        ───
  runs workflow ─────────────▶ WorktreeProvider.create()
                                resolves branch/path
                                calls `git worktree add` ───▶ creates worktree
                                (worktrunk hooks never run,
                                 worktree invisible to `wt list`)
```

### After

```
  User                         Archon                        engine                git / wt
  ────                         ──────                        ──────                ────────
  runs workflow ─────────────▶ WorktreeProvider.create()
                                loads worktree.engine from
                                .archon/config.yaml
                                resolveWorktreeEngine() ────▶ [GitWorktreeEngine]  ──▶ `git worktree add`
                                                          or▶ [WorktrunkEngine]    ──▶ `wt switch --create`
                                                               (pinned dest path,       (user's pre-start/
                                                                preflighted `wt`          post-start hooks
                                                                binary + version)         run; shows in
                                                                                           `wt list`)
```

### Before

```
WorktreeProvider (packages/isolation/src/providers/worktree.ts)
  │
  ├──▶ execFileAsync('git', ['worktree','add', ...])   (inline)
  ├──▶ execFileAsync('git', ['worktree','remove', ...]) (inline)
  ├──▶ listWorktrees() (@archon/git)                    (inline)
  └──▶ execFileAsync('git', ['worktree','prune'])       (inline)
```

### After

```
WorktreeProvider (packages/isolation/src/providers/worktree.ts)
  │
  ├──▶ resolveWorktreeEngine(config.worktree.engine)  [+] packages/isolation/src/engine/resolve.ts
  │        │
  │        ├──▶ GitWorktreeEngine       [+] packages/isolation/src/engine/git-engine.ts
  │        │        └──▶ execFileAsync('git', ...) / listWorktrees()   (unchanged @archon/git calls)
  │        │
  │        └──▶ WorktrunkEngine         [+] packages/isolation/src/engine/worktrunk-engine.ts
  │                 └──▶ execFileAsync('wt', ...)   ===  (wt CLI, https://worktrunk.dev)
  │
  └──▶ (fork-PR checkout unchanged: always raw git, bypasses engine)
```

**Connection inventory:**

| `WorktreeProvider` | `@archon/git` `execFileAsync`/`listWorktrees` (raw) | modified | Same-repo/issue/task/thread worktree add-remove-list-prune now route through the resolved `WorktreeEngine` instead of calling `@archon/git` inline |
| `WorktreeProvider` | `packages/isolation/src/engine/resolve.ts` | **new** | Resolves `worktree.engine` config string to an engine instance once per create/destroy/list call |
| `engine/resolve.ts` | `GitWorktreeEngine` | **new** | Default engine; wraps the same `@archon/git` calls the provider used to make directly |
| `engine/resolve.ts` | `WorktrunkEngine` | **new** | Opt-in engine; shells out to the `wt` binary |
| `WorktrunkEngine` | `wt` CLI (external process) | **new** | New external dependency, opt-in only (default engine is still `git`) |
| `WorktreeProvider` (fork-PR path) | `@archon/git` `execFileAsync` (raw) | unchanged | Always raw git; never routed through the engine |
| `.archon/config.yaml` `worktree.*` | `WorktreeCreateConfig` | modified | Added `engine?: string` field, passed through unchanged from the existing repo-config loader |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: L`
- Scope: `isolation`
- Module: `isolation:worktree-engine`

## Change Metadata

- Change type: `feature`
- Primary scope: `isolation`

## Linked Issue

- Closes #2260

```bash
bun run validate
# check:bundled, check:bundled-skill, check:bundled-schema, check:pi-vendor-map,
# check:capability-matrix, type-check, lint --max-warnings 0, format:check, test
# → all passed, exit code 0
```

Package-level test results:
- `packages/isolation` (`bun run test`): all batches green, including the new `src/engine/` batch — 36 tests (unit tests for both engines with mocked subprocess + a `wt`-gated integration test against a real repo) and the existing provider suite now at 158 tests (153 original, unmodified, + 5 new engine-selection tests)
- `packages/git` (`bun run test`): 192 pass, 0 fail (unchanged)
- `packages/core` `config-loader.test.ts`: 62 pass, 0 fail (includes 2 new tests for `worktree.engine` parsing)

- Evidence provided (test/log/trace/screenshot): full `bun run validate` output; per-package `bun run test` output (see PR description context)
- If any command is intentionally skipped, explain why: none skipped

## Security Impact (required)

- New permissions/capabilities? No — `WorktrunkEngine` shells out to `wt` the same way the codebase already shells out to `git`, via the same `execFileAsync` wrapper (no shell interpolation, argv-based).
- File system access scope changed? No — the worktree destination path is still fully controlled by Archon (pinned via `--config-set worktree-path=<literal>` on every `wt` invocation), so the environment registry and path resolution are unaffected regardless of engine.
- If any `Yes`, describe risk and mitigation: N/A

## Compatibility / Migration

- Backward compatible? Yes — default `worktree.engine` is `'git'`, which is the exact pre-existing behavior (existing tests for `WorktreeProvider` pass unmodified).
- Config/env changes? Yes, additive only: new optional `worktree.engine: 'git' | 'worktrunk'` key in `.archon/config.yaml`. Omitting it is a no-op.
- Database migration needed? No.
- If yes, exact upgrade steps: N/A (no migration required; users opt in per repo)

- Verified scenarios:
  - `GitWorktreeEngine` issues byte-identical argv to the pre-refactor inline code for: new-branch creation (`--no-track`), same-repo PR tracking checkout, existing-branch checkout, remove (with/without `--force`), list, prune.
  - `WorktrunkEngine` verified against the real `wt` binary (v0.61.0, installed locally) in an integration test: `add()` creates a real worktree at the pinned destination, `list()` reports it, `remove()` removes it while leaving the branch (`--no-delete-branch`), `prune()` runs cleanly.
  - Binary preflight: missing `wt` and too-old `wt` versions both fail with an actionable error naming `wt` and the minimum version; a successful preflight is cached (one `--version` call across repeated operations), a failed one is not (so installing `wt` after a failure doesn't require a process restart).
  - Provider-level engine selection verified end-to-end: `create()`/`destroy()`/`list()` all route through the configured engine; an unrecognized `worktree.engine` value fails loudly before any git/wt operation runs (no silent fallback).
- Edge cases checked: fork-PR checkout still uses raw git unconditionally under both engine settings; branch lifecycle (delete) stays with the provider in both engines (`wt remove` always passes `--no-delete-branch`).
- What was not verified: worktrunk-specific hook behavior (pre-start/post-start user hooks) beyond confirming the mechanism invokes `wt switch`/`wt remove` correctly — hook execution itself is worktrunk's own tested responsibility, out of scope here. Windows behavior for the `wt` binary path was not manually verified (same caveat as the existing git-based worktree code, which also assumes POSIX-friendly tooling in places).

- Affected subsystems/workflows: `packages/isolation` (worktree creation/removal/listing for issue, task, thread, review, and same-repo-PR workflows); `packages/core` config loader (new passthrough field only, no behavior change to existing fields).
- Potential unintended effects: none expected for existing users (default engine unchanged). Users who opt into `worktree.engine: worktrunk` take on a new dependency on the `wt` binary being installed and current.
- Guardrails/monitoring for early detection: fail-fast error messages naming the missing/outdated binary at the first worktree operation, before any partial state is created; existing structured logging (`isolation.worktree`/`isolation.worktrunk` loggers) unchanged in shape.

## Rollback Plan (required)

- Fast rollback command/path: revert the PR, or (without reverting) unset/remove `worktree.engine` from `.archon/config.yaml` — the default `'git'` engine is byte-identical to pre-PR behavior.
- Feature flags or config toggles (if any): `worktree.engine` is itself the toggle; default `'git'` is the safe/previous behavior.
- Observable failure symptoms: a misconfigured `worktree.engine` value or missing/outdated `wt` binary surfaces as an immediate, actionable error at worktree creation/removal time (never a silent fallback or partial worktree).

## Risks and Mitigations

- Risk: A worktrunk pre-start hook could fail or hang, blocking worktree creation for users who opt in.
  - Mitigation: This is the same class of risk the existing "heavy post-checkout hooks" comment already documents for git-based worktree creation (5-minute subprocess timeout applies identically to both engines); users control their own worktrunk hooks and can disable/fix them independently of Archon.
- Risk: `wt`'s CLI surface could change in a future release, breaking argv assumptions (e.g. `--config-set`, `--no-delete-branch`, `--format json`).
  - Mitigation: `WorktrunkEngine` preflights the installed version against `MIN_WORKTRUNK_VERSION` and fails loudly on anything older; a future breaking release would need a corresponding version bump here, not a silent behavior change.

## Summary by CodeRabbit

* **New Features**
  * Added configurable worktree engines, with Git as the default and optional Worktrunk support.
  * Worktree operations can now use the Worktrunk CLI, including creation, removal, and listing.
  * Invalid engine settings and unavailable or unsupported Worktrunk installations now fail clearly.

* **Documentation**
  * Added configuration guidance and usage details for selecting a worktree engine.

* **Tests**
  * Added coverage for engine selection, Git operations, Worktrunk integration, validation, and fallback behavior.

### DIFF
--- a/.claude/docs/isolation-and-worktree-guide.md
+++ b/.claude/docs/isolation-and-worktree-guide.md
@@ -83,9 +83,9 @@ Compares `store.countActiveByCodebase()` against `maxWorktrees` (default 25). If
 - Auto-detects default branch: `symbolic-ref` → `origin/main` → `master`
 
 **3. Create worktree**
-- For PRs (same-repo): `git fetch origin prBranch` → `git worktree add path -b prBranch origin/prBranch`
-- For PRs (fork): fetch `refs/pull/{N}/head` → create at SHA → checkout branch
-- For all others: `git worktree add path -b branchName origin/{baseBranch}`
+- For PRs (same-repo): `git fetch origin prBranch` → engine `add()` (`git worktree add`/`wt switch --create`, tracking `origin/prBranch`)
+- For PRs (fork): fetch `refs/pull/{N}/head` → create at SHA → checkout branch — **always raw git**, regardless of the configured engine (no worktrunk equivalent for synthetic review branches)
+- For all others: engine `add()` with start-point `origin/{baseBranch}`
 - Before creation: `cleanOrphanDirectoryIfExists()` removes stale non-worktree directories
 
 **4. Copy configured files**
@@ -93,6 +93,21 @@ Compares `store.countActiveByCodebase()` against `maxWorktrees` (default 25). If
 - User config: additional paths from `.archon/config.yaml` `worktree.copyFiles`
 - Supports `"source -> destination"` arrow syntax
 - Path traversal blocked (security check via `isPathWithinRoot()`)
+
+---
+
+## Worktree Engine (`packages/isolation/src/engine/`, #2260)
+
+The low-level add/remove/list/prune plumbing behind `WorktreeProvider` is pluggable via `worktree.engine` in `.archon/config.yaml` (`'git'` default, `'worktrunk'` opt-in). `WorktreeProvider` resolves the engine once per `create()`/`destroy()`/`list()` call (`resolveWorktreeEngine()` in `resolve.ts`) and never bypasses it for the paths engines cover — all provider-level logic (branch naming, adoption, copyFiles, submodules, branch cleanup) stays engine-independent.
+
+| File | Purpose |
+|------|---------|
+| `engine/types.ts` | `WorktreeEngine` interface: `add`/`remove`/`list`/`prune` |
+| `engine/git-engine.ts` | `GitWorktreeEngine` — the historical raw `git worktree` commands, moved behind the interface verbatim |
+| `engine/worktrunk-engine.ts` | `WorktrunkEngine` — shells out to `wt`; pins the worktree destination via `--config-set worktree-path="<literal>"` so Archon's path layout is unaffected; preflights the binary (`wt --version`) and fails fast (cached only on success) if missing or below `MIN_WORKTRUNK_VERSION` |
+| `engine/resolve.ts` | `resolveWorktreeEngine(engine?: string)` — maps the config string to a singleton engine instance; throws on an unrecognized value (no silent git fallback) |
+
+Fork-PR checkout (`refs/pull/N/head`) always uses raw `git worktree add` directly — never routed through the configured engine (issue #2260 non-goal). Best-effort orphan cleanup after a partial PR-creation failure (`cleanOrphanWorktreeIfExists`) also stays on raw git deliberately: it's cleaning up an ordinary git worktree regardless of which engine created it, so `git worktree remove` works either way, and skipping worktrunk's remove hooks in this rare error-recovery path is an acceptable simplification.
 
 ---
 
@@ -225,6 +240,7 @@ If merged cleanup frees space → proceed with creation. If not → block with f
 | Factory / singleton | `packages/isolation/src/factory.ts` |
 | 7-step resolver | `packages/isolation/src/resolver.ts` |
 | `WorktreeProvider` | `packages/isolation/src/providers/worktree.ts` |
+| `WorktreeEngine` interface + git/worktrunk impls | `packages/isolation/src/engine/` |
 | File copy utilities | `packages/isolation/src/worktree-copy.ts` |
 | Error classification | `packages/isolation/src/errors.ts` |
 | Cleanup service | `packages/core/src/services/cleanup-service.ts` |
--- a/packages/core/src/config/config-loader.test.ts
+++ b/packages/core/src/config/config-loader.test.ts
@@ -262,6 +262,28 @@ recommendedWorkflows: "archon-plan"
 
       expect(config.recommendedWorkflows).toBeUndefined();
     });
+
+    test('parses worktree.engine as worktrunk', async () => {
+      mockFsReadFile.mockResolvedValue(`
+worktree:
+  engine: worktrunk
+`);
+
+      const config = await loadRepoConfig('/test/repo');
+
+      expect(config.worktree?.engine).toBe('worktrunk');
+    });
+
+    test('worktree.engine is undefined when not configured (git engine applies by default)', async () => {
+      mockFsReadFile.mockResolvedValue(`
+worktree:
+  baseBranch: main
+`);
+
+      const config = await loadRepoConfig('/test/repo');
+
+      expect(config.worktree?.engine).toBeUndefined();
+    });
   });
 
   describe('loadConfig', () => {
--- a/packages/core/src/config/config-types.ts
+++ b/packages/core/src/config/config-types.ts
@@ -284,6 +284,28 @@ export interface RepoConfig {
      * @example 'upstream'
      */
     remote?: string;
+
+    /**
+     * Low-level worktree engine for this repo.
+     *
+     * - `'git'` (default) — raw `git worktree` commands, the historical behavior
+     * - `'worktrunk'` — shells out to the `wt` CLI (https://worktrunk.dev) so
+     *   per-repo worktrunk hooks (env setup, dependency install, cleanup) run
+     *   when Archon creates or removes worktrees, and Archon-managed worktrees
+     *   show up in `wt list`
+     *
+     * Worktree destination paths stay under Archon's control in both engines
+     * (pinned per invocation), so `worktree.path` and the environment registry
+     * behave identically. Fork-PR checkouts always use raw git.
+     *
+     * Fail Fast: with `'worktrunk'` set but no usable `wt` binary on PATH
+     * (missing, or older than the minimum supported version), worktree
+     * operations fail with an actionable error — no silent fallback to git.
+     * Unrecognized values also fail loudly.
+     *
+     * @default 'git'
+     */
+    engine?: 'git' | 'worktrunk';
   };
 
   /**
--- a/packages/docs-web/src/content/docs/reference/configuration.md
+++ b/packages/docs-web/src/content/docs/reference/configuration.md
@@ -150,6 +150,8 @@ worktree:
                         # Must be relative; no absolute, no `..` segments.
   remote: origin        # Optional: git remote name for fetch/push. Auto-detected
                         # when omitted (origin if it exists, sole remote otherwise).
+  engine: git           # Optional: 'git' (default) or 'worktrunk'. See "Worktree engine
+                        # (worktree.engine)" below.
 
 # Documentation directory
 docs:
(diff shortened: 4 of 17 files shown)


Wirasm:
Thanks for this — the proposal in #2260 was well argued, the scope boundary is drawn exactly where it should be (no new `IIsolationProvider`, resolver/adoption/branch-naming/fork-PR checkout all untouched), and the engine extraction moves the existing behavior across verbatim with tests. The care here is visible. I'm still declining it, and I want to be straight about why rather than leave it sitting open.

**The decision is about direction, not quality.**

1. **It puts a third-party binary on the isolation hot path.** Archon's stated posture is to talk to git directly and trust git's own guardrails — every workflow run, on every platform, depends on worktree creation succeeding. Routing that through `wt` makes a tool we don't control or version a hard dependency of the engine's most load-bearing step, in exchange for interop rather than capability.

2. **The payload is ungoverned execution.** The actual user-visible gain is that worktrunk's per-repo hooks fire during Archon worktree creation. Those are arbitrary user commands running inside Archon's setup, invisible to the event log and audit trail. For an engine whose pitch is governed execution with a full audit trail, that arrives as a behavioral change wearing a plumbing-swap costume.

3. **No principled stopping rule.** `WorktreeEngine` is a two-implementation interface whose second implementation is one vendor. The same argument admits `gwq`, `git-workspace`, and the next tool after that — each a permanent maintenance surface. And the worktrunk path can only be covered by mocks in CI (no `wt` on the runners), so it degrades silently while the tests stay green.

**What I do want, and what you actually surfaced:** the need underneath this is real and belongs to Archon — people want *setup to run after a worktree is created* (install deps, generate env files, seed gitignored state). Archon covers a sliver today (`worktree.copyFiles`, `initSubmodules`) and nothing more, and we hit the gap ourselves with stale `node_modules` leaking into agent worktrees.

So I've filed **#2278 — `worktree.postCreate`**: an Archon-owned post-create hook in repo config, running after copyFiles/submodule init, fail-fast on non-zero exit, timeout-bounded, and emitted as workflow events so setup is auditable. It serves every user rather than one tool's users, and it needs no external binary. You're credited on it, and it's yours to take if you want it — the design work you already did on ordering and failure modes transfers almost directly.

For your own setup in the meantime: `worktree.path` (repo config) co-locates Archon's worktrees under your repo root instead of `~/.archon/worktrees/`, which gets you a consistent layout today without the coupling.

Closing this and #2260 in favor of #2278. Genuinely appreciate the contribution — please do keep them coming.

Wirasm:
@AlienClubrider — follow-up, because "declined" shouldn't mean "you lose your worktrunk setup." I went through worktrunk's docs to work out what your workflow actually looks like without the engine, and it turns out most of it is reachable today, with config on your side and no Archon-side code at all.

## 1. Your hooks — `wt hook` as a passthrough

Worktrunk exposes manual hook invocation (`wt hook <type> [args...]`, filterable to specific hooks like `wt hook pre-merge test build`). So once #2278 lands, your `.config/wt.toml` stays the single source of truth and Archon just calls into it:

```yaml
# .archon/config.yaml
worktree:
  postCreate:
    - wt hook pre-start
```

No duplication of hook commands into Archon config. And the dependency runs the right way round: `wt` isn't a hard dependency of Archon's worktree creation for everyone — it's a command *you* chose to put inside an Archon-governed step, so it's audited as a run event, fails loud with the exit code surfaced, and is timeout-bounded. If `wt` is missing, exactly one user's `postCreate` fails; nobody else's worktree creation has ever heard of worktrunk.

One caveat from worktrunk's own semantics: `post-*` hooks are backgrounded, so `wt hook post-start` will likely return immediately and Archon's fail-fast/timeout won't meaningfully gate it. `pre-start` is the blocking variant — which is where your dep-install lives anyway, per worktrunk's own example.

**Available today, before #2278 lands:** a `bash:` node at the top of your workflow runs in the worktree, so `wt hook pre-start` there gets you the same effect right now. Clunkier (it's per-workflow rather than per-repo), but it unblocks you immediately.

## 2. Your layout — align the templates, no code required

This is the part I hadn't appreciated when I wrote the decline, and it's the stronger half of your motivation.

Archon computes worktree paths as `join(base, branchName)` (`packages/isolation/src/providers/worktree.ts:592-602`), where `base` is configurable — `worktree.path` (repo-relative, opt-in) or `paths.worktrees` (global absolute override) — and the leaf is the raw branch name, so `archon/task-foo` nests as `<base>/archon/task-foo`.

Worktrunk's `worktree-path` is a full template with `{{ repo_path }}`, `{{ repo }}`, `{{ branch }}` (raw) and `{{ branch | sanitize }}`. Since it offers the **raw** `{{ branch }}`, it can reproduce Archon's shape exactly, slashes and all:

```toml
worktree-path = "{{ repo_path }}/../worktrees/{{ branch }}"
```

...with Archon's `paths.worktrees` (or repo-local `worktree.path`) pointed at the same directory. Both sides land on identical paths. Worktrunk's template is the more expressive of the two, so it's the natural side to adapt — adding a leaf-naming template to Archon would just duplicate capability you already have.

## 3. The one thing I couldn't verify — worth 5 minutes of yours

Whether path convergence actually buys `wt list` visibility and `wt merge` / `wt remove` addressability depends on how worktrunk discovers worktrees: `git worktree list` versus its own state, and whether merge/remove tolerate worktrees they didn't create. The docs don't say. They lean git-based (`wt list --branches` can show branches with no worktree), and **if discovery is git-based then Archon's worktrees already show up in `wt list` today** and step 2 is unnecessary.

You have `wt` installed and I don't — if you check, please drop the answer on #2278. If it turns out worktrunk genuinely can't see externally-created worktrees at any path, that's a concrete gap worth re-opening the conversation over, and it's a much narrower ask than the engine abstraction: probably an upstream `wt adopt`, not Archon-side code.

Thanks again for the proposal and for taking the decline in stride — #2278 is yours if you want it.

PR #2264, opened 2026-07-25
fix(workflows): reject unsupported refactor targets

### LINKED ISSUE #2018: archon-refactor-safely can leave stale running state on non-TypeScript repo
## Summary

`archon-refactor-safely` can remain in a non-terminal `running` state after being run against a non-TypeScript repository. In the observed case, the workflow was not paused, was not waiting for approval/input, had no active step metadata, and no active worker process was apparent, but it continued to appear as running.

This report is sanitized from a local reproduction: local filesystem paths, private repository names, and local run identifiers have been omitted.

## Reproduction Context

A user started `archon-refactor-safely` against a private Python repository to review/refactor it toward a minimal production handoff.

The bundled workflow appears TypeScript/Bun-oriented:

- `scan-scope` scans only `*.ts`
- validation uses `bun run type-check`
- validation uses `bun run lint`
- validation uses `bun run format:check`
- validation uses `bun run test`

The target repository did not have that TypeScript/Bun project shape.

## Observed Behavior

`archon workflow status --json` showed an active run similar to:

```json
{
  workflow_name: archon-refactor-safely,
  status: running,
  current_step_index: null,
  metadata: {},
  completed_at: null
}
```

`archon workflow runs --json` counted the run as active:

```json
{
  counts: {
    running: 1,
    paused: 0,
    completed: 0,
    failed: 0,
    cancelled: 0
  }
}
```

Verbose run events indicated the workflow stopped during `plan-refactor`. The final event was a completed shell command within that node. There was no `node_completed` event for `plan-refactor`, no transition to `execute-refactor`, and no terminal workflow event.

Process inspection did not show an active old workflow worker. Only later diagnostic/status command processes were visible.

## Expected Behavior

One of these should happen:

- The workflow completes `plan-refactor` and advances to `execute-refactor`.
- The workflow fails fast with a clear error if the target repository does not match the workflow's language/tooling assumptions.
- The workflow pauses with explicit approval/input metadata if human input is required.
- If the agent node exits or errors in the same worker process before emitting `node_completed`, the worker records a failed terminal event.
- If ownership is ambiguous because the work may still be running elsewhere, Archon surfaces that ambiguous/stale state and offers an explicit user action such as abandon/resume, rather than silently leaving the run indistinguishable from active work.

## Likely Contributing Factors

- `archon-refactor-safely` is currently TypeScript-first but can be invoked against a Python repository.
- The scan node can produce mostly empty TypeScript metrics, yet downstream planning proceeds.
- The workflow did not appear to emit a terminal run state when the agent stopped after writing the refactor plan.
- Run display/management does not clearly distinguish an active worker from an ambiguous stale/orphaned state.

## Suggested Fixes

1. Add project/language/tooling detection before `scan-scope`.
2. Fail early or route to a language-neutral refactor workflow when no TypeScript/Bun project is detected.
3. Ensure agent-node exit without `node_completed` in the same worker process writes a failed terminal event.
4. Add worker heartbeat/ownership metadata so the UI/CLI can distinguish active work from ambiguous stale/orphaned state.
5. For ambiguous stale/orphaned runs, surface the state and provide explicit user-controlled actions (`abandon`, `resume`, cleanup guidance) instead of autonomously marking the run failed/cancelled based only on age.
6. Improve `workflow abandon` documentation/discoverability for this case.

## Diagnostic Commands Used

```powershell
archon workflow status --json
archon workflow runs --limit 10 --json
archon workflow get <run-id> --verbose --json
Get-Process | Where-Object { $_.ProcessName -match 'archon|bun|node|codex|python|powershell' }
Get-CimInstance Win32_Process | Where-Object { $_.Name -match 'bun|node|powershell|codex' }
```

## Workaround

Do not resume the stale run. Preserve artifacts if needed, explicitly abandon the run, and start a fresh narrower workflow or prompt that matches the target repository language and validation commands.

shixi-li:
## Summary

- Problem: `archon-refactor-safely` can continue past an empty TypeScript scan when the target repository does not match the workflow's TypeScript/Bun assumptions.
- Why it matters: Unsupported targets can consume provider work before failing later, without a clear project-shape diagnostic.
- What changed: Added a deterministic project preflight, gated scope scanning on it, excluded `.archon` helpers from source discovery, regenerated the embedded workflow, and added executable regression coverage.
- What did **not** change (scope boundary): This does not add heartbeat detection, automatic abandonment, or workflow-engine lifecycle mutation.

### Before

```text
User                    Archon
────                    ──────
starts safe refactor ─▶ scans only *.ts
                        may count .archon helpers
                        continues to analysis
non-TS target ◀──────── fails later without a project-shape diagnostic
```

### After

```text
User                    Archon
────                    ──────
starts safe refactor ─▶ [checks root project shape]
                        ├─ unsupported ─▶ clear terminal error; no provider node runs
                        └─ supported ───▶ scans application *.ts files
                                         continues existing workflow
```

### Before

```text
workflow YAML
    ├──▶ scan-scope ──▶ analyze-impact ──▶ remaining DAG
    └──▶ bundle generator ──▶ bundled-defaults.generated.ts

bundled-defaults.test.ts ──▶ bundle completeness checks
```

### After

```text
[~] workflow YAML
    ├──▶ [+] check-project ===▶ [~] scan-scope ──▶ analyze-impact ──▶ remaining DAG
    │          └──▶ clear failure for unsupported targets
    └──▶ bundle generator ──▶ [~] bundled-defaults.generated.ts

[~] bundled-defaults.test.ts
    ├──▶ parseWorkflow
    └──▶ executes check-project against temporary project fixtures
```

**Connection inventory**:

| `check-project` | `scan-scope` | **new** | Blocks scans and provider nodes until project validation passes |
| `scan-scope` | `analyze-impact` | unchanged | Preserves the existing workflow sequence |
| Workflow YAML | Generated bundle | **modified** | Regenerated from the canonical default workflow |
| Regression test | `BUNDLED_WORKFLOWS` / `parseWorkflow` | **new** | Exercises the actual bundled Bash node |
| Regression test | Temporary project fixtures | **new** | Covers unsupported, incomplete, and supported project shapes |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `workflows`
- Module: `workflows:defaults`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2018 — https://github.com/coleam00/Archon/issues/2018
- Related # None
- Depends on # None
- Supersedes # None

Commands and result summary:

```bash
bun test packages/workflows/src/defaults/bundled-defaults.test.ts
# PASS: 17 tests, 236 assertions

bun test packages/workflows/src/defaults/
# PASS: 21 tests, 254 assertions

bun run check:bundled
# PASS: generated bundle is current (36 commands, 21 workflows)

bun run --cwd packages/workflows type-check
# PASS

bun run validate
# PASS
```

- Evidence provided (test/log/trace/screenshot): Executable tests run the bundled preflight against temporary Python-only, incomplete TypeScript, and supported TypeScript projects.
- If any command is intentionally skipped, explain why: None.

## Security Impact (required)

- If any `Yes`, describe risk and mitigation: N/A. The preflight is read-only and repository-scoped; `.archon` is excluded from project source discovery.

## Compatibility / Migration

- Backward compatible? `Yes` for supported TypeScript/Bun projects; incompatible project shapes now fail earlier by design.
- Config/env changes? `No`
- If yes, exact upgrade steps: None.

- Verified scenarios:
  - A Python repository is rejected before any provider node.
  - TypeScript files under `.archon` do not make a non-TypeScript repository appear supported.
  - Missing validation scripts are reported by name.
  - A TypeScript project with all required scripts passes the preflight.
  - `scan-scope` depends on `check-project`.
- Edge cases checked:
  - Empty or absent required scripts.
  - Bash spawn errors surface instead of becoming false-positive test results.
  - Case-variant Windows `Path`/`PATH` handling.
  - Bundled workflow parsing and source/bundle consistency.
- What was not verified:
  - A live provider-backed refactor run.
  - The Windows CI runner before this PR's matrix executes.

- Affected subsystems/workflows: Only the bundled `archon-refactor-safely` definition, its generated copy, and bundled-default regression tests.
- Potential unintended effects: Projects with only non-`.ts` sources or nonstandard root validation-script names are now rejected explicitly before analysis.
- Guardrails/monitoring for early detection: Clear diagnostics, executable project-shape fixtures, bundle-drift checking, and the Ubuntu/Windows test matrix.

## Rollback Plan (required)

- Fast rollback command/path: Revert this commit; the canonical YAML and generated bundle revert together.
- Observable failure symptoms: A valid supported project is rejected by `check-project` before `scan-scope`.

## Risks and Mitigations

- Risk: The project-shape contract may be too strict for a TypeScript repository using nonstandard script names.
  - Mitigation: The preflight checks the exact commands the workflow already invokes and lists every missing script.
- Risk: Bash/Bun behavior could differ across operating systems.
  - Mitigation: Tests execute the bundled Bash through `resolveBashPath`, normalize `PATH` case and delimiters, and run in the Ubuntu/Windows CI matrix.

## Summary by CodeRabbit

* **New Features**
  * Added preflight checks for supported TypeScript projects before automated refactoring begins.
  * Projects are now validated for required configuration, Bun availability, eligible source files, and quality-check scripts.

* **Bug Fixes**
  * Improved scope analysis for large files and exported code.
  * Prevented unsupported or incomplete projects from proceeding to automated analysis.

* **Tests**
  * Added coverage for successful validation, rejection scenarios, and workflow gating.


Wirasm:
## Review: preflight gate for `archon-refactor-safely`

Thanks for a genuinely well-put-together first PR — the template is filled in properly, the scope boundary is explicit, and the test executes the *real* bundled bash node rather than a copy-pasted duplicate. That last choice is the right one and I want to call it out, because it's what let me verify the rest of this quickly.

**Heads-up on process:** CI had never run on this PR (first-time-contributor approval gate), so every result in your description was author-reported. I ran the full suite locally — **it all passes**, details below. Your reported numbers were accurate.

### Validation results (first real run anywhere)

| Check | Result |
|---|---|
| `bun run validate` (full chain) | **PASS** — real exit code 0 |
| `bun run check:bundled` | **PASS** — `bundled-defaults.generated.ts` is up to date (36 commands, 21 workflows) |
| `bun test src/defaults/ …` (the batch) | **PASS** — 210 pass / 0 fail / 595 assertions |
| Total suite | 6372 pass, 0 fail |

Bundle regeneration is correct — the generated file genuinely matches the YAML.

### Things I verified and can confirm are right

- **The gate is complete.** All 10 nodes form a strict linear chain; `check-project` is the only node without `depends_on`, so nothing can run in parallel with the preflight. No AI node can fire before it.
- **The preflight's `find` filter is byte-identical in semantics to `scan-scope`'s**, so the gate is a true precondition for the scan rather than an approximation. Nice.
- **No test-ordering coupling.** `registerBuiltinProviders()` in `beforeAll` is the only registry call in that whole `bun test` batch (`src/defaults/ src/model-validation.test.ts src/router.test.ts src/utils/ src/hooks.test.ts`) — no `clearRegistry()`, no conflicting `mock.module()`. It's also idempotent by design. Clean.
- All five `find` invocations in the file consistently carry the new `.archon` exclusion.

---

## I1 (Important) — a valid project is rejected with *zero* diagnostic output

`.archon/workflows/defaults/archon-refactor-safely.yaml:41-43`

```bash
SOURCE_FILE=$(find . -type f -name '*.ts' \
  -not -path '*/node_modules/*' ... -print -quit 2>/dev/null)
```

`set -e` + `2>/dev/null` combine badly here. `find` returns non-zero if *any* error occurred during the walk — even when it successfully found a match. Because this is a bare assignment, `set -e` aborts the script immediately, and `2>/dev/null` has already swallowed the reason. The node dies with **exit 1 and no output at all**.

I reproduced this. A fully valid project — `src/i.ts` present, all four scripts present, bun on PATH — plus one unreadable directory:

```
### find exit status with an unreadable dir present:
find: ./aa_locked: Permission denied
./src/i.ts            <-- the source file WAS found
find EXIT=1

### full preflight in that project:
PREFLIGHT EXIT=1      <-- and nothing else. No message.
```

This is exactly the failure symptom you listed in your own Rollback Plan ("A valid supported project is rejected by `check-project` before `scan-scope`"), and it's reachable without anything exotic: root-owned files left by a container run, a permission-odd `node_modules/.cache`, macOS trash dirs. Archon runs workflows against arbitrary user repos, so this will find people — and when it does they get a completely empty error.

The same mechanism also hides a `-quit` portability failure. I shimmed a BusyBox-style `find` that rejects `-quit`:

```
### find WITHOUT -quit support — valid TS project:
EXIT=1                <-- again, zero output
```

(For the record: `-quit` is fine on GNU findutils, macOS BSD find, and Git-Bash's find, so this is the less likely trigger of the two — but it degrades identically badly.)

**Suggested fix** — one token, keeps current behavior, removes the false abort:

```bash
SOURCE_FILE=$(find . -type f -name '*.ts' \
  -not -path '*/node_modules/*' -not -path '*/.git/*' -not -path '*/.archon/*' -not -path '*/dist/*' \
  -not -name '*.test.ts' -not -name '*.d.ts' -print -quit 2>/dev/null || true)
```

The existing `if [ -z "$SOURCE_FILE" ]` check then does the right thing in both cases. Dropping `2>/dev/null` entirely would be even better for debuggability, at the cost of some permission noise in the node log.

---

## I2 (Important) — `.tsx`-only projects are rejected, and the message doesn't say why

`archon-refactor-safely.yaml:41-48`

The scan is `-name '*.ts'`, so a TypeScript React app whose sources are all `.tsx` is rejected. Verified — this project, with all four scripts present and correct:

```
package.json: {"scripts":{"type-check":..,"lint":..,"format:check":..,"test":..}}
src/App.tsx

ERROR: archon-refactor-safely requires a TypeScript project with at least one application .ts file.
```

Telling a TypeScript user that their repo isn't a TypeScript project is going to read as a bug. The gate's behavior is arguably *correct* (the workflow really does only scan `.ts`), but then the limitation belongs in the message. Something like:

> `ERROR: archon-refactor-safely only scans .ts source files; .tsx/.jsx are not supported by this workflow.`

Same applies to the `.test.ts`-only case, which produces the identical confusing message.

Worth deciding explicitly: is `.tsx` out of scope for this workflow forever, or is the honest fix to add `-o -name '*.tsx'` to all five `find`s? Either is defensible — but the current state rejects a large, common class of TS repo with a message that misdescribes the reason.

---

## S1 (Suggestion) — monorepo roots without root scripts

Verified: a workspace root whose scripts live in `packages/*/package.json` is rejected with

```
ERROR: ... missing required package.json scripts: type-check, lint, format:check, test
```

The message is actionable in that it names them, but it doesn't say they must be in the **root** `package.json`. Adding "in the root package.json" would save a round of confusion. (Archon itself has root scripts, so its own repo passes — I checked.)

## S2 (Suggestion) — two untested branches are the two most platform-sensitive ones

The new tests cover: non-TS repo, missing scripts, happy path, and the `depends_on` wiring. Not covered: the **missing `package.json`** branch and the **`command -v bun`** branch. Those are precisely the two whose behavior is least certain under Git-Bash on the Windows runner. Both are cheap to add (the second via a `PATH` without bun's dir). I ran them by hand and they behave correctly on macOS:

```
### no package.json:  ERROR: ... requires a TypeScript project with a root package.json.   EXIT=1
### bun NOT on PATH:  ERROR: ... requires Bun on PATH.                                     EXIT=1
```

## S3 (Suggestion) — the gate-completeness assertion is too narrow to hold the invariant

`bundled-defaults.test.ts` asserts `scanNode?.depends_on` equals `['check-project']`. That pins today's edge but won't catch the actual regression you care about: someone later adding a *new* root node that runs alongside the preflight. A direct encoding of the invariant:

```ts
const roots = parsed.workflow!.nodes.filter(n => !n.depends_on?.length);
expect(roots.map(n => n.id)).toEqual(['check-project']);
```

## S4 (Suggestion) — `-not -path` doesn't prune

`-not -path '*/node_modules/*'` filters output but still **descends** into `node_modules`. On the success path `-quit` short-circuits so it rarely matters; on the *rejection* path find walks the entire dependency tree before failing. `-prune` would avoid that. Minor, and it matches the existing style in this file, so entirely your call.

---

## What I could not verify: Windows

`test.yml` runs the matrix `[ubuntu-latest, windows-latest]`, and this is the first test under `src/defaults/` to spawn real `bash`/`bun` subprocesses — so this code path has **never executed on the Windows runner**. I can't test it from macOS. Reasoning through it, the design looks deliberately Windows-aware and I expect it to pass:

- `resolveBashPath()` resolves Git-Bash on Windows, which is present on `windows-latest`; its `find` is GNU findutils, so `-quit` is supported.
- The `;`-joined PATH is correct — MSYS2 converts a Windows-form `PATH` to POSIX form at bash startup. Prepending `dirname(process.execPath)` is a neat way to guarantee `command -v bun` succeeds.
- The case-insensitive `PATH`/`Path` handling is right for Windows env semantics.
- `bun -e` top-level await is supported.

The one residual unknown I'd flag: MSYS2 sometimes mangles arguments that look path-like, and the `bun -e` payload is a multi-line string containing `;` and `:`. I believe it's left alone (no leading `/`), but it is unverified. If CI goes red on Windows, that's the first place I'd look.

Since CI has never run here, please make sure a maintainer approves the workflow run so the Windows leg actually executes before this merges.

---

## Verdict

**Ship with fixes.** The approach is right, the gate is correctly and completely wired, and the bundle/test work is clean. **I1** should be fixed before merge — it's a one-token change and without it the preflight can hard-fail a perfectly good repo with an empty error message, which is worse than the problem the PR is solving. **I2** needs a decision (fix the message, or widen to `.tsx`); either is fine, but the current message misdescribes the reason. S1–S4 are optional polish.

Nice work — the diff is tight, the reasoning in the description is honest about what wasn't verified, and testing the real bundled node instead of a duplicated string is a better instinct than most PRs that touch bundled defaults.

Wirasm:
Thanks for this, and I want to lead with what's good because I'm going to decline it on direction rather than on quality.

Your test work is better than most contributions this repo gets: parsing the *bundled* workflow via `parseWorkflow(BUNDLED_WORKFLOWS[...])`, extracting the real `check-project` body, and executing it with `spawnSync` against temp fixtures — including a `.archon/scripts/helper.ts` decoy to prove the exclusion works, and a "valid project still passes" case so it can't silently over-reject. That's the right way to test a defaults change. I ran the full suite on your branch (CI had never run — first-time contributors need approval): `bun run validate` PASS, `check:bundled` PASS, 6372 tests, and the regenerated bundle matches your YAML byte for byte. Your self-reported numbers were accurate, nothing overstated.

I also confirmed the gate is complete — `scan-scope` was the only DAG root, so nothing can run in parallel with your preflight — and that you correctly reached for a `bash:` node plus a `depends_on` edge instead of asking to widen `requires:` into toolchain detection. That's the sanctioned escape hatch and exactly the right instinct.

**Why I'm declining anyway.**

`archon-refactor-safely` is a bundled default — it ships to every Archon user. The direction we've committed to is that bundled defaults must work on **any** project: *"Defaults stay polyglot: never hardcode a stack's commands in a bundled workflow"* (#2123, which converts validation to `detect-validation` → `run-validation` → `fix`).

Your preflight requires `package.json`, `bun` on PATH, at least one `.ts` file, and the four npm script names. You didn't introduce that assumption — the `validate` node has always run literal `bun run type-check|lint|format:check|test` — but the preflight makes it **formal**. It writes "TypeScript only" into a shipped default, and then #2123 step 4 has to delete it. I'd rather not add work whose next step is undoing it.

The problem you identified is real and I'm not dismissing it. Running against a Python repo shouldn't burn tokens through analyze/plan/execute before failing. But the polyglot fix gets that benefit for free and for every language: a detect node that finds no usable validation commands **is** the fail-fast gate. So I've filed **#2280** with that as the goal, crediting you, and it's yours if you want it. It's a bigger piece of work than this PR, and considerably more valuable.

**Two things I found while reviewing, worth knowing regardless of where this lands:**

1. **A real bug in the preflight** — `set -e` plus `2>/dev/null` on the `find` at lines 41-43 means *any* find error aborts the node with exit 1 and **zero output**. I reproduced it: a fully valid project (source file present, all four scripts, bun on PATH) plus one unreadable directory → silent rejection. `find` returns non-zero when an error occurred even if it found the match. Root-owned files from container runs are a realistic trigger. The fix is one token — `|| true` — and it's the exact symptom you listed in your own Rollback Plan, so you'd predicted the failure mode. Worth carrying into #2280's detect node.
2. **Two false-rejection classes** — `.tsx`-only repos get told "requires a TypeScript project with at least one application `.ts` file" (a TS React user being told their repo isn't TypeScript), and monorepo roots whose scripts live in workspace packages are rejected outright. Both are the kind of thing a polyglot detect node handles naturally and an allowlist gate can't.

For the record on `-quit`: I checked and it's fine on GNU, macOS BSD, and Git-Bash find, so that wasn't the portability risk. The `;`-joined PATH your test constructs is also correct — MSYS2 converts it at bash startup.

Sorry to send you away from a working patch. If #2280 interests you I'll review it promptly, and the detection design is worth discussing on the issue before you write code.

PR #2275, opened 2026-07-26
fix(cli): doctor honors claudeBinaryPath config fallback (#2263)

### LINKED ISSUE #2263: doctor: Claude binary check ignores the assistants.claude.claudeBinaryPath config fallback (false FAIL)
## Summary

On v0.6.0 (binary build, macOS arm64), `archon doctor` reports:

```
✗ Claude binary: CLAUDE_BIN_PATH is not set. Run `archon setup` to configure.
```

on a machine where workflows dispatch and run fine, because the Claude binary is configured via `~/.archon/config.yaml`:

```yaml
assistants:
  claude:
    claudeBinaryPath: /Users/.../bin/claude
```

## Cause

`checkClaudeBinary` in `packages/cli/src/commands/doctor.ts` only reads `env.CLAUDE_BIN_PATH` and fails when it is unset. The actual runtime resolver (`packages/providers/src/claude/binary-resolver.ts`) has a documented fallback chain: `CLAUDE_BIN_PATH` → `assistants.claude.claudeBinaryPath` (binary mode) → native-installer autodetect. So doctor FAILs setups the runtime happily runs.

## Expected

Doctor should verify what the runtime will actually do — ideally by calling the same resolver — and report which source resolved (env var / config / autodetect), failing only when the whole chain comes up empty.

## Why it matters

The config-file path is the documented fix for compiled builds (README points to it), so users who followed the docs get a permanent false ✗ in doctor — which trains them to ignore doctor output, precisely the tool meant to catch real environment breakage (we run Archon headless under launchd, where a real binary-resolution failure previously cost us 7 silent overnight workflow failures).

bigboy1122:
## Summary

- **Problem:** `checkClaudeBinary` read only `env.CLAUDE_BIN_PATH` and hard-failed when it was unset. The runtime resolver (`resolveClaudeBinaryPath`) has a documented three-tier chain: `CLAUDE_BIN_PATH` → `assistants.claude.claudeBinaryPath` (binary mode) → native-installer autodetect. Doctor therefore FAILed setups the runtime runs happily.
- **Why it matters:** the config-file path is the documented fix for compiled builds and the README points users to it, so anyone who followed the docs got a permanent false `✗ Claude binary: CLAUDE_BIN_PATH is not set` while workflows dispatched and ran fine. A doctor that fails working setups trains users to ignore its output.
- **What changed:** doctor now calls the same resolver the runtime uses and reports which tier produced the path (`via env` / `via config` / `via autodetect`), failing only when the whole chain comes up empty — and in that case surfaces the resolver's install instructions verbatim instead of a misleading env-var message. `resolveClaudeBinaryWithSource` is added alongside `resolveClaudeBinaryPath` (which now delegates to it), structured exactly like the existing Codex resolver.
- **What did *not* change (scope boundary):** resolution *order* and *semantics* are untouched — all existing Claude resolver tests pass unchanged. Dev-mode skip behavior is unchanged. No other doctor check touched. `checkCodexBinary` untouched (it was already correct, and is the precedent this follows).

## Review Follow-ups

- **Wiring test added** (67b50c1b) — see Validation Evidence below.
- **`ClaudeBinaryResolution` extracted** (08c4d0ab) — the `{ path, source }` shape had been hand-copied into both the resolver's return type and doctor's `resolve` parameter.
- Left to maintainers per review: the `cli.md:91` docs line, `checkClaudeBinary`'s hard-FAIL for Codex-only users, and the dev-mode `CLAUDE_BIN_PATH` blind spot.

## Change Metadata

- Change type: `bug`
- Primary scope: `cli`
- Risk: `risk: low` · Size: `size: S` · Module: `cli:doctor`

## Linked Issue

- Closes #2263

```bash
bun run validate     # → exit 0 (all nine gates), 6373 pass / 0 fail
```

**1. The doctor→config wiring is now guarded, and the guard is proven to bite.** The original tests injected *both* the deps loader and the resolver, so they only exercised parameter plumbing inside `checkClaudeBinary`; `defaultLoadClaudeBinaryDeps` — which reads `assistants.claude.claudeBinaryPath`, the tier #2263 was about — ran in no test.

It now takes the merged-config loader as an injected parameter defaulting to the real lazy import, so the mapping is assertable without `mock.module()` (process-global, and this file avoids it by design). Two tests cover it: one on the deps function directly, one driving `checkClaudeBinary` through the *real* loader to assert the configured path reaches the resolver. The stub config carries a different value under `codexBinaryPath` so a key mix-up cannot pass by coincidence.

Verified by making the exact swap from the review — `config.assistants.codex.codexBinaryPath`:

```
(fail) checkClaudeBinary > reads assistants.claude.claudeBinaryPath, not another assistant key (#2263)
(fail) checkClaudeBinary > routes the real deps loader through to the resolver (#2263 wiring guard)
 69 pass, 2 fail
```

Both new tests fail, nothing else does. Coverage under the doctor suite alone moved from `doctor.ts` `128-132` uncovered (the whole deps-function body) to `133-136` — only the lazy `import('@archon/core')` inside the default argument, which is unreachable without `mock.module()` or a real config read.

`binary-resolver.ts` remains at 0% function coverage *under the doctor suite* — exercising the real resolver there needs a binary on disk. It is covered by its own 18-test suite; the seam this PR was asked to guard is the config→deps→resolver hand-off, which is now asserted.

**2. The config tier's source label is now asserted** (`binary-resolver.test.ts`). It was the one tier whose label nothing checked; mislabeling it ships green while doctor prints "via env" for a config-resolved path.

**3. The config link is real, not assumed.** Wrote an actual `~/.archon/config.yaml` with the snippet from the issue and ran `loadConfig` against it: `claudeBinaryPath = "/Users/me/bin/claude"`.

**4. Refactor is behavior-preserving.** All existing Claude resolver tests pass untouched: `binary-resolver.test.ts` 18 pass / 0 fail, `binary-resolver-dev.test.ts` 8 pass / 0 fail. Doctor suite: 71 pass / 0 fail.

## Security Impact (required)

- File system access scope changed? **No** — doctor already spawned the resolved binary with `--version`; it now spawns the path the runtime would have used anyway. Config reading is a lazy import of the same `loadConfig` other checks already call.

## Compatibility / Migration

- Config/env changes? **No**

`CLAUDE_BIN_PATH` remains the highest-priority tier, so existing env-var setups behave identically (message gains a `via env` annotation). Setups that were correctly failing still fail, with a more actionable message. `resolveClaudeBinaryPath`'s signature and return type are unchanged, so `claude/provider.ts` is unaffected. `defaultLoadClaudeBinaryDeps` gained an optional parameter; the production call site is still the zero-argument default.

- **Verified scenarios:** the config-key regression guard proven to fail on the wrong key (executed, output above); `loadConfig` surfacing `claudeBinaryPath` from a real config file (executed, not mocked); resolver refactor proven non-breaking by the pre-existing tests passing unchanged; all four resolution outcomes (env / config / autodetect / empty-chain) covered; config-load failure path exercised.
- **Edge cases checked:** dev mode still skips (the `isBinary` guard is deliberately kept rather than inferring dev mode from a nullish resolve result the way `checkCodexBinary` does — Claude's resolver honors `CLAUDE_BIN_PATH` in dev mode too, so a truthy result is *not* proof of binary mode); malformed config degrades instead of failing; `CLAUDE_BIN_PATH` still wins over config.
- **What was not verified:** I did not run this from an actual compiled binary on the reporter's platform — `BUNDLED_IS_BINARY` is a compile-time const, so the binary branch is only reachable in tests via the injected `isBinary` seam. The reporter is on v0.6.0 macOS arm64 and can confirm against a release build. Autodetect tier was not exercised against a real `~/.local/bin/claude` on disk.

- **Affected subsystems/workflows:** `archon doctor` output for the Claude binary check, and `archon setup` (which invokes doctor at the end and discards the result). No runtime/execution path changes — `resolveClaudeBinaryPath` behaves identically for `claude/provider.ts`.
- **Potential unintended effects:** doctor now performs a lazy `@archon/core` import + `loadConfig` for this check. That is already the established pattern for `checkCodexBinary` / `checkDatabase` / `checkConnectedProviders`, and failures are caught and degraded rather than propagated.
- **Guardrails/monitoring for early detection:** a config-load failure logs `doctor.claude_deps_load_failed` at debug. A regression re-surfaces as the literal string `CLAUDE_BIN_PATH is not set`, asserted against by an existing test, or as a wrong config key, now asserted against by the new wiring tests.

## Rollback Plan (required)

- **Fast rollback command/path:** `git revert <merge-sha>`. Two source files plus tests, no migrations, no state, no config format change.
- **Feature flags or config toggles:** none (not warranted for a diagnostic-accuracy fix).
- **Observable failure symptoms:** `archon doctor` reporting `✗ Claude binary` on a machine where `archon workflow run` succeeds — i.e. the original bug.

## Risks and Mitigations

- **Risk:** doctor now depends on `loadConfig` succeeding for this check, adding a failure mode that did not exist before.
  - **Mitigation:** wrapped in try/catch that degrades to `{}` (env + autodetect still resolve), logged at debug, and covered by a dedicated test. Same posture as `checkCodexBinary`.
- **Risk:** `resolveClaudeBinaryPath` is on the hot path for every Claude run, and it was restructured.
  - **Mitigation:** it is now a two-line delegation to a function whose body is the original code verbatim; the pre-existing resolver tests (including the dev-mode suite) pass without modification, and its exported signature and return type are unchanged.
- **Risk:** binary-mode behavior unverified on a real compiled build (see Human Verification).
  - **Mitigation:** the binary branch is driven through the same injected `isBinary` seam the pre-existing tests already used; the resolver itself is shared with the runtime that reporters confirm works.

## Summary by CodeRabbit

- **New Features**
  - Improved Claude Code executable detection across environment settings, configuration, and automatic discovery.
  - Diagnostics now identify how the executable was found.
  - Configured executable paths are recognized and validated.

- **Bug Fixes**
  - Claude binary checks now handle configuration-loading failures gracefully.
  - Improved error messages when Claude Code cannot be found or fails to start.
  - Non-binary installations correctly skip executable validation.

### DIFF
--- a/packages/cli/src/commands/doctor.test.ts
+++ b/packages/cli/src/commands/doctor.test.ts
@@ -26,6 +26,7 @@ import {
   checkTelegram,
   checkTelemetry,
   checkFolderProject,
+  defaultLoadClaudeBinaryDeps,
   doctorCommand,
   type ClaudeBinaryDeps,
   type CodexBinaryDeps,
@@ -34,6 +35,7 @@ import {
   type OpenCodeDeps,
 } from './doctor';
 import * as doctorModule from './doctor';
+import type { MergedConfig } from '@archon/core';
 
 describe('checkClaudeBinary', () => {
   let execSpy: ReturnType<typeof spyOn<typeof git, 'execFileAsync'>>;
@@ -136,6 +138,44 @@ describe('checkClaudeBinary', () => {
     );
     // A broken config must not fail the binary check outright.
     expect(result.status).toBe('pass');
+  });
+
+  // The tests above inject BOTH seams, so they only prove parameter plumbing
+  // inside checkClaudeBinary. The two below exercise the real
+  // defaultLoadClaudeBinaryDeps, which is where #2263 actually lived: reading
+  // the wrong config key type-checks and would otherwise ship green.
+  // The stub config deliberately carries a DIFFERENT value under
+  // assistants.codex.codexBinaryPath so a key mix-up fails loudly rather than
+  // resolving to the same string by accident.
+  const stubConfig = async (): Promise<Pick<MergedConfig, 'assistants'>> => ({
+    assistants: {
+      claude: { claudeBinaryPath: '/from/claude/config' },
+      codex: { codexBinaryPath: '/from/codex/config' },
+    },
+  });
+
+  it('reads assistants.claude.claudeBinaryPath, not another assistant key (#2263)', async () => {
+    const deps = await defaultLoadClaudeBinaryDeps(stubConfig);
+    expect(deps.configBinaryPath).toBe('/from/claude/config');
+  });
+
+  it('routes the real deps loader through to the resolver (#2263 wiring guard)', async () => {
+    execSpy.mockResolvedValue({ stdout: '1.0.0', stderr: '' });
+    let sawConfigPath: string | undefined;
+
+    const result = await checkClaudeBinary(
+      true,
+      // The real loader, not a fake — this is the link the bug broke.
+      () => defaultLoadClaudeBinaryDeps(stubConfig),
+      async configPath => {
+        sawConfigPath = configPath;
+        return { path: configPath as string, source: 'config' };
+      }
+    );
+
+    expect(sawConfigPath).toBe('/from/claude/config');
+    expect(result.status).toBe('pass');
+    expect(result.message).toContain('via config');
   });
 });
 
--- a/packages/cli/src/commands/doctor.test.ts
+++ b/packages/cli/src/commands/doctor.test.ts
@@ -27,6 +27,7 @@ import {
   checkTelemetry,
   checkFolderProject,
   doctorCommand,
+  type ClaudeBinaryDeps,
   type CodexBinaryDeps,
   type DatabaseDeps,
   type FolderProjectDeps,
@@ -45,23 +46,32 @@ describe('checkClaudeBinary', () => {
     execSpy.mockRestore();
   });
 
+  const noDeps = async (): Promise<ClaudeBinaryDeps> => ({});
+
   it('returns skip when not in binary mode', async () => {
-    const result = await checkClaudeBinary({}, false);
+    const result = await checkClaudeBinary(false);
     expect(result.status).toBe('skip');
     expect(result.label).toBe('Claude binary');
     expect(execSpy).not.toHaveBeenCalled();
   });
 
-  it('returns fail in binary mode when CLAUDE_BIN_PATH is unset', async () => {
-    const result = await checkClaudeBinary({}, true);
+  it('returns fail in binary mode when the whole resolution chain is empty', async () => {
+    const result = await checkClaudeBinary(true, noDeps, async () => {
+      throw new Error('Claude Code not found. Archon requires the Claude Code executable');
+    });
     expect(result.status).toBe('fail');
-    expect(result.message).toContain('CLAUDE_BIN_PATH');
+    // The resolver's install instructions are surfaced verbatim — they are the
+    // actionable message, unlike the old "CLAUDE_BIN_PATH is not set".
+    expect(result.message).toContain('Claude Code not found');
     expect(execSpy).not.toHaveBeenCalled();
   });
 
   it('returns pass in binary mode when binary spawns successfully', async () => {
     execSpy.mockResolvedValue({ stdout: '1.0.0', stderr: '' });
-    const result = await checkClaudeBinary({ CLAUDE_BIN_PATH: '/opt/claude' }, true);
+    const result = await checkClaudeBinary(true, noDeps, async () => ({
+      path: '/opt/claude',
+      source: 'env',
+    }));
     expect(result.status).toBe('pass');
     expect(result.message).toContain('/opt/claude');
     expect(execSpy).toHaveBeenCalledWith('/opt/claude', ['--version'], expect.any(Object));
@@ -69,10 +79,63 @@ describe('checkClaudeBinary', () => {
 
   it('returns fail in binary mode when spawn throws', async () => {
     execSpy.mockRejectedValue(new Error('ENOENT'));
-    const result = await checkClaudeBinary({ CLAUDE_BIN_PATH: '/opt/claude' }, true);
+    const result = await checkClaudeBinary(true, noDeps, async () => ({
+      path: '/opt/claude',
+      source: 'env',
+    }));
     expect(result.status).toBe('fail');
     expect(result.message).toContain('did not spawn');
     expect(result.message).toContain('ENOENT');
+  });
+
+  // #2263: doctor previously read only CLAUDE_BIN_PATH and hard-FAILed setups
+  // configured via assistants.claude.claudeBinaryPath — the documented fix for
+  // compiled builds — even though those setups run workflows fine.
+  it('passes when the binary comes from config rather than CLAUDE_BIN_PATH (#2263)', async () => {
+    execSpy.mockResolvedValue({ stdout: '1.0.0', stderr: '' });
+    let sawConfigPath: string | undefined;
+    const result = await checkClaudeBinary(
+      true,
+      async () => ({ configBinaryPath: '/Users/me/bin/claude' }),
+      async configPath => {
+        sawConfigPath = configPath;
+        return { path: configPath as string, source: 'config' };
+      }
+    );
+
+    // The config path must actually reach the resolver, not be ignored.
+    expect(sawConfigPath).toBe('/Users/me/bin/claude');
+    expect(result.status).toBe('pass');
+    expect(result.message).toContain('/Users/me/bin/claude');
+    // Which tier resolved it is surfaced so users can see what the runtime does.
+    expect(result.message).toContain('via config');
+    expect(result.message).not.toContain('CLAUDE_BIN_PATH is not set');
+  });
+
+  it('reports the autodetect tier when the native installer path resolves', async () => {
+    execSpy.mockResolvedValue({ stdout: '1.0.0', stderr: '' });
+    const result = await checkClaudeBinary(true, noDeps, async () => ({
+      path: '/home/me/.local/bin/claude',
+      source: 'autodetect',
+    }));
+    expect(result.status).toBe('pass');
+    expect(result.message).toContain('via autodetect');
+  });
+
+  it('degrades to env/autodetect when config loading throws', async () => {
+    execSpy.mockResolvedValue({ stdout: '1.0.0', stderr: '' });
+    const result = await checkClaudeBinary(
+      true,
+      async () => {
+        throw new Error('malformed config.yaml');
+      },
+      async configPath => {
+        expect(configPath).toBeUndefined();
+        return { path: '/opt/claude', source: 'env' };
+      }
+    );
+    // A broken config must not fail the binary check outright.
+    expect(result.status).toBe('pass');
   });
 });
 
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -18,7 +18,7 @@ import {
   resolveClaudeBinaryWithSource,
   type ClaudeBinaryResolution,
 } from '@archon/providers/claude/binary-resolver';
-import type { Codebase, SchemaVersionInfo } from '@archon/core';
+import type { Codebase, MergedConfig, SchemaVersionInfo } from '@archon/core';
 
 // Vendor-canonical credential id for Codex (since #1955 credentials are keyed
 // by vendor, not agent). A connected `openai` key signals Codex intent even
@@ -125,11 +125,19 @@ export async function checkClaudeBinary(
   }
 }
 
-async function defaultLoadClaudeBinaryDeps(): Promise<ClaudeBinaryDeps> {
-  // Lazy import so doctor doesn't pull the full @archon/core graph for an
-  // unrelated check (matches defaultLoadCodexBinaryDeps).
-  const { loadConfig } = await import('@archon/core');
-  const config = await loadConfig(process.cwd());
+export async function defaultLoadClaudeBinaryDeps(
+  // Injected so the config-key mapping — the tier #2263 was actually about —
+  // can be asserted without mock.module(), which is process-global and would
+  // leak into every other test in this file's batch. Defaults to the real
+  // lazy import so the production path is the zero-argument call.
+  loadMergedConfig: (cwd: string) => Promise<Pick<MergedConfig, 'assistants'>> = async cwd => {
+    // Lazy import so doctor doesn't pull the full @archon/core graph for an
+    // unrelated check (matches defaultLoadCodexBinaryDeps).
+    const { loadConfig } = await import('@archon/core');
+    return loadConfig(cwd);
+  }
+): Promise<ClaudeBinaryDeps> {
+  const config = await loadMergedConfig(process.cwd());
   return { configBinaryPath: config.assistants.claude.claudeBinaryPath };
 }
 
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -16,7 +16,7 @@ import {
 } from '@archon/providers/codex/binary-resolver';
 import {
   resolveClaudeBinaryWithSource,
-  type ClaudeBinarySource,
+  type ClaudeBinaryResolution,
 } from '@archon/providers/claude/binary-resolver';
 import type { Codebase, SchemaVersionInfo } from '@archon/core';
 
@@ -77,9 +77,7 @@ export async function checkClaudeBinary(
   loadDeps: () => Promise<ClaudeBinaryDeps> = defaultLoadClaudeBinaryDeps,
   resolve: (
     configPath?: string
-  ) => Promise<
-    { path: string; source: ClaudeBinarySource } | undefined
-  > = resolveClaudeBinaryWithSource
+  ) => Promise<ClaudeBinaryResolution | undefined> = resolveClaudeBinaryWithSource
 ): Promise<CheckResult> {
   const label = 'Claude binary';
   if (!isBinary) {
@@ -96,7 +94,7 @@ export async function checkClaudeBinary(
     deps = {};
   }
 
-  let resolved: { path: string; source: ClaudeBinarySource } | undefined;
+  let resolved: ClaudeBinaryResolution | undefined;
   try {
     resolved = await resolve(deps.configBinaryPath);
   } catch (err) {
--- a/packages/cli/src/commands/doctor.ts
+++ b/packages/cli/src/commands/doctor.ts
@@ -14,6 +14,10 @@ import {
   resolveCodexBinaryWithSource,
   type CodexBinarySource,
 } from '@archon/providers/codex/binary-resolver';
+import {
+  resolveClaudeBinaryWithSource,
+  type ClaudeBinarySource,
+} from '@archon/providers/claude/binary-resolver';
 import type { Codebase, SchemaVersionInfo } from '@archon/core';
 
 // Vendor-canonical credential id for Codex (since #1955 credentials are keyed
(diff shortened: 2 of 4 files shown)


bigboy1122:
Addressed the `ClaudeBinaryResolution` nitpick in 08c4d0ab — the shape is now a named exported interface next to `ClaudeBinarySource`, used by both the resolver's return type and doctor's `resolve` parameter.

Scoped to Claude deliberately: `resolveCodexBinaryWithSource` duplicates the same inline shape today, so a matching `CodexBinaryResolution` would be the symmetric follow-up — but changing Codex here is unrelated to the doctor fix under review. Happy to fold it in if you'd prefer both in one pass.

`bun run validate` → exit 0.

Wirasm:
---
pr: 2275
title: "fix(cli): doctor honors claudeBinaryPath config fallback (#2263)"
author: "bigboy1122"
reviewed: 2026-07-27
recommendation: ship-with-fixes
---

# PR Review: #2275 — doctor honors claudeBinaryPath config fallback

**Author**: @bigboy1122
**Branch**: `fix/2263-doctor-claude-binary-config` → `dev`
**Files Changed**: 3 (+165/-22)
**Closes**: #2263

---

## Summary

Deep pass on top of the earlier light triage. The refactor is genuinely
behavior-preserving — I diffed the old `resolveClaudeBinaryPath` body against
the new `resolveClaudeBinaryWithSource` body and the ONLY differences are the
three return-value wraps. All nine `bun run validate` gates pass (exit 0,
6371 tests, 0 fail) in a clean worktree.

One real gap the light pass missed: the doctor→config wiring that *was* the bug
in #2263 has **zero test coverage**, confirmed empirically by Bun's coverage
report, not inferred. Everything else is polish.

**Verdict: ship-with-fixes.** Correct today; the one missing test and the docs
line are cheap and should land in this PR or as an immediate follow-up.

---

## Deep-Dive Answers (the four questions the light pass deferred)

### 1. Is the extract-and-delegate genuinely behavior-preserving? — YES, verified

I diffed the pre-PR `resolveClaudeBinaryPath` body (`git show dev:…` lines
115–175) against the post-PR `resolveClaudeBinaryWithSource` body
(`packages/providers/src/claude/binary-resolver.ts:150-193`). The bodies are
byte-identical except three lines:

```
<     return resolvedEnv;              →  >  return { path: resolvedEnv, source: 'env' };
<     return resolvedConfig;           →  >  return { path: resolvedConfig, source: 'config' };
<     return nativeInstallerPath;      →  >  return { path: nativeInstallerPath, source: 'autodetect' };
```

Everything else is preserved:

- **Order**: env → `if (!BUNDLED_IS_BINARY) return undefined` → config → autodetect → throw. Unchanged.
- **Signature/return type**: `resolveClaudeBinaryPath(configClaudeBinaryPath?: string): Promise<string | undefined>` is unchanged. `packages/providers/src/claude/provider.ts:1264` needed no edit and got none.
- **Logging**: `getLog().info({ binaryPath, source }, 'claude.binary_resolved')` fires with the same object and event name, still **exactly once** per resolution. `resolveClaudeBinaryPath` adds no second log.
- **`pathKind` module-local calls**: both call sites (`validateAndExpand` at :70-87 and the autodetect check at :183) are untouched and still module-local within the same file, so `spyOn(resolver, 'pathKind')` in the existing tests intercepts them exactly as before. Confirmed by the 18 binary-mode tests passing unchanged.
- **Throw path**: `INSTALL_INSTRUCTIONS` thrown from the same position.
- **Spy compatibility**: `packages/providers/src/claude/provider.test.ts:970` does `spyOn(binaryResolver, 'resolveClaudeBinaryPath')` — still the exported symbol `provider.ts` imports, so interception is unaffected.

Only cost: one extra `await` hop on a path that already awaits `statSync`-backed work. Immaterial.

### 2. Would a regression in the doctor→real-resolver wiring be caught? — NO

This is the finding (I1 below). Hard evidence from `bun test packages/cli/src/commands/doctor.test.ts` with coverage:

```
packages/providers/src/claude/binary-resolver.ts   |   0.00 |  15.56 | …135-138,150-191
packages/cli/src/commands/doctor.ts                |  81.08 |  83.98 | 109,128-132,…,702-715
```

- `binary-resolver.ts` at **0.00% function coverage** during the doctor suite — the real resolver is never called.
- doctor.ts **128–132** uncovered = the entire body of `defaultLoadClaudeBinaryDeps`, i.e. `loadConfig(process.cwd())` + `config.assistants.claude.claudeBinaryPath`.
- doctor.ts **702–715** uncovered = the default `checks` array in `doctorCommand`, so `checkClaudeBinary()` with real default args never executes.
- doctor.ts **109** uncovered = the defensive `!resolved` branch (S1).

### 3. Does the dev-mode `isBinary` guard still short-circuit, and is `CLAUDE_BIN_PATH` still tier 1? — YES to both

- `checkClaudeBinary` returns `skip` at `doctor.ts:83-85` before `loadDeps`/`resolve` are touched. Asserted by `doctor.test.ts:52` (`checkClaudeBinary(false)` → skip, `execSpy` not called).
- Tier 1 is directly asserted by two **pre-existing** tests that pass unchanged: `binary-resolver.test.ts:75` ("env var takes precedence over config path") and `:110` ("env var takes precedence over autodetect"). Both run through the delegating wrapper, so the delegation itself is exercised 26×.
- The author's stated reason for keeping the `isBinary` param rather than copying `checkCodexBinary`'s nullish-inference is **correct**: Claude's resolver honors `CLAUDE_BIN_PATH` in dev mode (`binary-resolver.ts:156-161`, before the `BUNDLED_IS_BINARY` gate), so a truthy result is genuinely not proof of binary mode.

### 4. Real validation results — see the table below. All nine gates green.

---

## Issues Found

### Critical

None.

### Important

**I1 — `packages/cli/src/commands/doctor.ts:128-134`: the exact wiring #2263 was about is untested.**

Every one of the seven `checkClaudeBinary` tests injects **both** `loadDeps` and
`resolve` fakes. The test labelled "the #2263 regression guard"
(`doctor.test.ts:97-115`) asserts that a value returned by a *fake* deps loader
reaches a *fake* resolver — that is parameter plumbing inside `checkClaudeBinary`,
not the `config.yaml → loadConfig → resolver` chain that actually broke.

Concretely, changing line 133 to `config.assistants.codex.codexBinaryPath`
type-checks, passes all 69 doctor tests and all 26 resolver tests, and ships the
same class of bug this PR exists to fix. The PR body says the link was verified
by hand ("wrote an actual `~/.archon/config.yaml` … `claudeBinaryPath = …`") —
good diligence, but it leaves no automated guard behind.

The type system catches only the crudest swaps (`defaultLoadCodexBinaryDeps` has
an `env` param and so isn't assignable to `() => Promise<ClaudeBinaryDeps>`;
`resolveCodexBinaryWithSource` returns a `CodexBinarySource` union that doesn't
fit `ClaudeBinarySource`). It does not catch a wrong config key.

**Fix** (cheap): export `defaultLoadClaudeBinaryDeps` and add one test that stubs
`loadConfig` and asserts it returns `{ configBinaryPath: '<the claude value>' }`
from a config where `assistants.codex.codexBinaryPath` holds a *different*
value — so a key mix-up fails loudly. Alternatively call
`checkClaudeBinary(true, undefined, fakeResolve)` and assert `fakeResolve`
received the config value, which covers the default-parameter binding too.

**Not a blocker on correctness** — I verified by hand that line 133 reads the
right key (`ClaudeProviderDefaults.claudeBinaryPath`,
`packages/providers/src/types.ts:23`) and that it type-checks under
`AssistantDefaults` (`packages/core/src/config/config-types.ts:61-64`, where
`claude` is required, not optional). This is future-regression risk, not a
present defect.

### Suggestions

**S1 — `packages/cli/src/commands/doctor.ts:106-110`: dead defensive branch.**
The comment concedes it: in binary mode the resolver either returns a value or
throws, so `!resolved` is unreachable in production (and shows up as uncovered
line 109). It is only reachable via an injected fake. Harmless (returns a benign
`skip`), but CLAUDE.md's YAGNI/Fail-Fast posture argues for deleting it or
downgrading the comment to "only reachable through the test seam".

**S2 — `packages/providers/src/claude/binary-resolver.ts:172`: the `config`
tier's `source` label is never asserted anywhere.**
`binary-resolver.test.ts:60-65` checks only the returned path; the `env` and
`autodetect` labels are covered indirectly by the logger assertions at `:94` and
`:116`, but `config` has neither. Mislabeling tier 2 as `source: 'env'` ships
green and makes doctor print "via env" for a config-resolved path — precisely
the diagnostic this PR adds. One `expect(mockLogger.info).toHaveBeenCalledWith({
binaryPath: '/custom/claude/cli.js', source: 'config' }, 'claude.binary_resolved')`
closes it.

**S3 — `packages/docs-web/src/content/docs/reference/cli.md:91`: now stale.**
The doctor checklist still reads "Claude binary spawn, Codex binary resolution
(env → config → vendor → autodetect, reporting which source resolved)". After
this PR Claude does the same thing. Suggested edit: "Claude binary resolution
(env → config → autodetect, reporting which source resolved) and spawn, Codex
binary resolution (env → config → vendor → autodetect, …)".

**S4 — `packages/cli/src/commands/doctor.ts:70-126`: Codex-only binary installs
still get a hard ✗ for Claude.**
`checkCodexBinary` has a "not configured → skip" gate (`doctor.ts:183-189`);
`checkClaudeBinary` has none, so a compiled-binary user who runs Codex only and
has never installed Claude gets a permanent red ✗ with install instructions.
That is the same "doctor fails a working setup" complaint as #2263, one provider
over. Out of this PR's stated scope — worth a follow-up issue rather than
scope creep here.

**S5 — dev mode still skips a check the resolver could now perform.**
`checkClaudeBinary` skips whenever `!isBinary` (`doctor.ts:83-85`), yet
`resolveClaudeBinaryWithSource` honors `CLAUDE_BIN_PATH` in dev mode **and
throws** when that path is bad (`binary-resolver.ts:156-161`, before the
`BUNDLED_IS_BINARY` gate). So a dev-mode operator with a typo'd
`CLAUDE_BIN_PATH` has every Claude run fail while doctor stays silent —
exactly the "doctor contradicts reality" failure mode this PR is fixing, in the
other direction. Pre-existing, but the resolver is now in hand: `if (!isBinary
&& !process.env.CLAUDE_BIN_PATH) return skip` would cover it in one line.

**S6 (nit) — one more `loadConfig` per doctor run.**
`doctor.ts` now calls `loadConfig(process.cwd())` at lines 132, 225 and 330 —
three times, concurrently under `Promise.allSettled`. Pre-existing pattern; no
action requested, just noting the count grew.

---

## Validation Results (actually executed in an isolated worktree)

`bun install` first (node_modules do not carry over), then:

| Check | Status | Details |
|---|---|---|
| `bun run validate` (all 9 gates) | **PASS** | exit 0 |
| `check:bundled` | PASS | up to date (36 commands, 21 workflows) |
| `check:bundled-skill` | PASS | up to date (23 files across 2 skills) |
| `check:bundled-schema` | PASS | up to date |
| `check:pi-vendor-map` | PASS | OK |
| `check:capability-matrix` | PASS | OK |
| `type-check` | PASS | all 10 packages exited 0 |
| `lint` (`--max-warnings 0`) | PASS | no output |
| `format:check` (prettier) | PASS | no output |
| `test` (all packages) | **PASS** | 6371 pass, 0 fail |

Targeted runs:

| Suite | Result |
|---|---|
| `packages/cli/src/commands/doctor.test.ts` | 69 pass / 0 fail (156 expects) |
| `packages/providers/src/claude/binary-resolver.test.ts` | 18 pass / 0 fail — **unmodified by this PR** |
| `packages/providers/src/claude/binary-resolver-dev.test.ts` | 8 pass / 0 fail — **unmodified by this PR** |

The 26 unmodified resolver tests passing is meaningful evidence, not a formality:
they exercise the delegation, tier order, tier-1 precedence, directory
expansion, and the throw path.

The only "Failed" string anywhere in the validate log is a benign
`@archon/server:test | Failed to load .env … ENOENT` (no `.env` in a fresh
worktree).

---

## Pattern Compliance

- [x] Type safety — no `any`, no assertions beyond the established `(err as Error).message` doctor idiom
- [x] Follows the `checkCodexBinary` precedent (injected deps/resolve seams, lazy `@archon/core` import, degrade-on-config-failure with a debug log)
- [x] Logging follows `{domain}.{action}_{state}` (`doctor.claude_deps_load_failed`)
- [x] Errors surfaced, not swallowed — the config-load catch is documented and intentional (fail-safe: env/autodetect still resolve)
- [x] Naming conventions (`ClaudeBinaryResolution` mirrors `CodexBinarySource`)
- [~] Tests added for new code — added, but they stop short of the wiring (I1)
- [ ] Documentation updated — `cli.md:91` not updated (S3)

---

## Strengths

- **The riskiest claim in the PR body is true and I verified it independently.** `resolveClaudeBinaryPath` is on the hot path for every Claude run; the extracted body is byte-identical modulo three return wraps, and the public signature/return type/logging are unchanged.
- **Correct, non-obvious reasoning on the `isBinary` seam.** Not copying `checkCodexBinary`'s nullish-inference is right, for exactly the reason stated: Claude's env tier fires before the `BUNDLED_IS_BINARY` gate.
- **Dropping the `env` parameter is a true no-op**, which the PR body doesn't spell out: `doctorCommand` sets `const env = process.env` (`doctor.ts:698`) and the resolver reads `process.env` — same source, no injectable divergence lost.
- **Two error-message improvements beyond the reported bug**, both free consequences of sharing the resolver: a `CLAUDE_BIN_PATH` pointing at a missing file now yields "set to X but the file does not exist" instead of a raw spawn error, and a path pointing at the npm platform-package *directory* now expands to the contained executable and passes instead of failing "did not spawn".
- **Comments explain *why*, not *what*** — the `isBinary` rationale and the degrade-on-config-failure note are the kind that survive.
- PR description is unusually thorough and honest about what was *not* verified (no run from a real compiled binary; autodetect not exercised against a real `~/.local/bin/claude`).

---

## What the light triage pass got wrong / left out

1. **It verdicted MERGE without noticing the coverage gap.** The new tests stub both seams, so the config→resolver chain the PR exists to fix is never executed. Bun's coverage output makes this unambiguous (`binary-resolver.ts` 0.00% functions; `doctor.ts` 128-132 uncovered) — it is checkable in one command, and wasn't checked.
2. **It took the PR body's "behavior-preserving" claim on trust.** It happens to be true, but the claim is only worth what the diff shows; the old-vs-new body diff is the evidence and takes 30 seconds.
3. **It reported no validation numbers of its own.** Real numbers: 9/9 gates, 6371 pass / 0 fail, 69 + 18 + 8 on the targeted suites.
4. **It missed the docs staleness** (`cli.md:91`) and the two adjacent doctor-accuracy gaps this change makes newly visible (S4 Codex-only false ✗, S5 dev-mode blind spot).

---

## Recommendation

**ship-with-fixes** — merge is safe on correctness grounds; add the I1 test (and
ideally S2 + S3, both one-liners) either in this PR or as an immediate
follow-up. Nothing here blocks the fix from reaching the #2263 reporter.

Note: no formal GitHub review state was set (no approve / request-changes) and
nothing was pushed to the contributor's branch — findings were posted as a PR
comment only.

---

*Reviewed by Claude (prp-review, single-pass) in an isolated worktree at
`/Users/rasmus/Projects/cole/Archon/.claude/worktrees/agent-a1d49de46c59c01eb`*

Wirasm:
Reviewed this properly — built the branch, ran the full suite. `bun run validate` exits 0, 6371 tests pass, and I diffed the old `resolveClaudeBinaryPath` body against the extracted `resolveClaudeBinaryWithSource` line by line: byte-identical apart from the three return wraps. Resolution order, the `BUNDLED_IS_BINARY` gate placement, `validateAndExpand`, the `pathKind` calls, the logging and the throw are all unchanged, and the public signature is untouched so the call site in `provider.ts` needed no edit. That's the right way to do a refactor on a path every Claude run depends on.

Dropping the `env` param is a genuine no-op (`doctorCommand` sets `const env = process.env` and the resolver reads `process.env` itself), and sharing the resolver quietly fixes two more misleading messages: a missing-file `CLAUDE_BIN_PATH` now says so instead of "did not spawn", and an npm platform-package *directory* now expands and passes.

**One thing before I merge.**

The doctor→resolver wiring isn't asserted anywhere. Your new tests inject a fake `resolve`, so the default parameter — the thing that actually connects `checkClaudeBinary` to the real resolver and reads `config.assistants.claude.claudeBinaryPath` — is never exercised. Concretely: change that line to `config.assistants.codex.codexBinaryPath` and it type-checks and passes all 95 tests. Coverage confirms it — `binary-resolver.ts` sits at 0% function coverage under the doctor suite, and `doctor.ts:128-132` (the whole body of `defaultLoadClaudeBinaryDeps`) is uncovered.

Nothing is wrong today — I verified the key is correct and required-typed. But #2263 *was* the config tier being read wrong, and right now a future refactor could reintroduce exactly that with a green suite. Could you add a test that exercises the real `defaultLoadClaudeBinaryDeps` against a stub config and asserts the configured path reaches the resolver?

**Follow-ups I'll take on our side, not asking you for these:**

- `checkClaudeBinary` still hard-FAILs for people who don't use Claude at all, whereas `checkCodexBinary` skips when the provider isn't configured. Pre-existing, same bug class as #2263, and your change makes the failure message much better for those users in the meantime.
- Dev mode still misses a broken `CLAUDE_BIN_PATH` that genuinely breaks dev runs — the `isBinary` guard short-circuits first. Your reasoning for keeping that guard rather than copying `checkCodexBinary` is correct, by the way: Claude's env tier fires before the `BUNDLED_IS_BINARY` gate, so inferring dev mode from a nullish result wouldn't work here.
- `docs-web/.../reference/cli.md:91` describes the old behaviour and is now stale.

I'll fold the docs line in and file the rest. Add the wiring test and this goes in.

Also — please strip the `🤖 Generated with Claude Code` footer from the description; we keep AI attribution out of commit history and a squash merge pulls the body in.

bigboy1122:
Added — pushed 67b50c1b.

You were right that the old tests only proved parameter plumbing. `defaultLoadClaudeBinaryDeps` now takes the merged-config loader as an injected parameter defaulting to the real lazy import, so the config-key mapping is assertable without `mock.module()` — this file avoids it deliberately and the CLI batch runs doctor alongside four other suites. Two tests: one on the deps function directly, one driving `checkClaudeBinary` through the **real** loader to assert the configured path reaches the resolver. The stub config carries a different value under `codexBinaryPath`, so a key mix-up cannot pass by coincidence.

I made your exact swap to confirm the guard bites rather than assuming it:

```
(fail) checkClaudeBinary > reads assistants.claude.claudeBinaryPath, not another assistant key (#2263)
(fail) checkClaudeBinary > routes the real deps loader through to the resolver (#2263 wiring guard)
 69 pass, 2 fail
```

Both new tests, nothing else. Coverage under the doctor suite moved from `doctor.ts:128-132` uncovered (the whole deps body) to `133-136` — just the lazy `import(...)` inside the default argument.

One thing I should be straight about: `binary-resolver.ts` is still 0% function coverage *under the doctor suite*. Calling the real resolver there needs a binary on disk, so the fake stays. What is now asserted is the config→deps→resolver hand-off, which is the seam that broke. If you wanted the resolver genuinely executed from doctor too, that is a different test shape and I am happy to add it.

Also took S2 — the `config` tier source label is now asserted in the resolver suite. It was the one tier nothing checked, and mislabeling it prints "via config" as "via env", which is the diagnostic this PR adds.

`bun run validate` exits 0: 6373 pass / 0 fail (6371 + the two new).

Left S3 and the two follow-ups alone since you said you would take them. Description footer stripped, and I folded the new evidence in.

Thanks for the depth on this one — the coverage check was the right call and I should have run it myself.

PR #2276, opened 2026-07-26
docs: fix PR template path + link agent files to CONTRIBUTING.md

bigboy1122:
## Summary

Fixes a 404 in three docs links. `CONTRIBUTING.md`, `CLAUDE.md`, and `AGENTS.md` all point at `.github/PULL_REQUEST_TEMPLATE.md`, but the file is committed lowercase as `.github/pull_request_template.md`. GitHub blob URLs are case-sensitive, so the link 404s; a case-insensitive local filesystem hides it.

Docs follow reality rather than renaming the template: the maintainer triage workflows already `[ -f ]`-probe both casings, so nothing downstream depends on the uppercase spelling, and renaming a file GitHub treats specially has a larger blast radius than fixing three references.

Per maintainer review, the `CONTRIBUTING.md` pointer bullet has been dropped from `CLAUDE.md` and `AGENTS.md` — those files are injected into every agent session, and PR-time process is a point-of-need concern that belongs in `CONTRIBUTING.md`. Only the three link corrections remain.

## Change Metadata

- Change type: `docs`
- Primary scope: `multi` (repo-root contributor/agent docs)
- Risk: `risk: low` · Size: `size: XS`

## Linked Issue

None — drive-by fix noticed while following `CONTRIBUTING.md` for #2248 and #2263. Happy to open a tracking issue if preferred.

```bash
bun run format:check   # → "All matched files use Prettier code style!" (exit 0)
```

The 404 was verified against the live repo, not inferred from the local filesystem:

```
$ gh api repos/coleam00/Archon/contents/.github/PULL_REQUEST_TEMPLATE.md?ref=dev
{"message":"Not Found","status":"404"}

$ gh api repos/coleam00/Archon/contents/.github/pull_request_template.md?ref=dev
pull_request_template.md
```

`git ls-files .github/` (authoritative, case-sensitive) confirms the committed name is lowercase.

`type-check`/`lint`/`test` are skipped as not meaningful for a markdown-only change touching no code paths; `format:check` is the gate covering `.md`.

## Security Impact (required)

## Compatibility / Migration

- Config/env changes? **No**

No file is renamed, so any external link or tooling pointing at the real lowercase path keeps working.

- **Verified scenarios:** both casings queried against the live GitHub API on `dev` (404 vs. found); committed filename confirmed via `git ls-files`; repo-wide grep for remaining uppercase references in contributor docs comes back empty; `format:check` passes.
- **Edge cases checked:** the remaining uppercase mentions under `.archon/commands/defaults/` and the maintainer triage workflows are both-casing probes against *arbitrary* target repos, not references to this repo's file, so they were deliberately left alone.
- **What was not verified:** rendering in the GitHub web UI post-merge — the link can only be clicked once merged to `dev`. The API check above is the equivalent evidence.

- **Affected subsystems/workflows:** documentation only. No code, config, or template content changes.
- **Guardrails/monitoring for early detection:** a regression is visible as a 404 on the template link.

## Rollback Plan (required)

- **Fast rollback command/path:** `git revert <merge-sha>` — three single-line docs edits.
- **Feature flags or config toggles:** none.
- **Observable failure symptoms:** the template link 404s again.

## Risks and Mitigations

- **Risk:** a maintainer may prefer the opposite fix — renaming the file to match the docs.
  - **Mitigation:** happy to switch; both casings are valid to GitHub, so it is genuinely the maintainers' call. Docs-follow-reality was chosen as the smaller, more reversible option.

## Summary by CodeRabbit

* **Documentation**
  * Updated contributor instructions with the correct pull request template path.
  * Clarified that contributors using the command-line workflow should copy the template into the pull request body.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,7 +39,7 @@
 
 - `main` is the release branch. Never commit directly to `main`.
 - `dev` is the working branch. All feature work branches off `dev` and merges back into `dev`.
-- All PRs must use the template at `.github/PULL_REQUEST_TEMPLATE.md` — fill in every section. When opening a PR via `gh pr create`, copy the template into the body explicitly; GitHub only auto-applies it through the web UI.
+- All PRs must use the template at `.github/pull_request_template.md` — fill in every section. When opening a PR via `gh pr create`, copy the template into the body explicitly; GitHub only auto-applies it through the web UI.
 - Link the issue with `Closes #<number>` (or `Fixes` / `Resolves`) in the PR description so it auto-closes on merge.
 - To release, use the `/release` skill. It compares `dev` to `main`, generates changelog entries, bumps the version, and creates a PR to merge `dev` into `main`.
 - Releases follow Semantic Versioning: `/release` (patch), `/release minor`, `/release major`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,6 +37,7 @@
 
 **Git Workflow and Releases**
 
+- **Read [`CONTRIBUTING.md`](./CONTRIBUTING.md) before committing or opening a PR.** It is the source of truth for contributor workflow and carries rules that are deliberately NOT duplicated here — notably commit-message conventions (present tense, first line under 72 characters) and the marketplace submission process. The bullets below cover branch/release policy only; they are not the complete contributor checklist.
 - `main` is the release branch. Never commit directly to `main`.
 - `dev` is the working branch. All feature work branches off `dev` and merges back into `dev`.
 - All PRs must use the template at `.github/pull_request_template.md` — fill in every section. When opening a PR via `gh pr create`, copy the template into the body explicitly; GitHub only auto-applies it through the web UI.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,7 +37,6 @@
 
 **Git Workflow and Releases**
 
-- **Read [`CONTRIBUTING.md`](./CONTRIBUTING.md) before committing or opening a PR.** It is the source of truth for contributor workflow and carries rules that are deliberately NOT duplicated here — notably commit-message conventions (present tense, first line under 72 characters) and the marketplace submission process. The bullets below cover branch/release policy only; they are not the complete contributor checklist.
 - `main` is the release branch. Never commit directly to `main`.
 - `dev` is the working branch. All feature work branches off `dev` and merges back into `dev`.
 - All PRs must use the template at `.github/pull_request_template.md` — fill in every section. When opening a PR via `gh pr create`, copy the template into the body explicitly; GitHub only auto-applies it through the web UI.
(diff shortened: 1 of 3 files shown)


Wirasm:
Thanks for this — the link fix is real and I want it. I verified the committed filename is lowercase `pull_request_template.md`, so those three references genuinely 404 on github.com (case-sensitive blob URLs; a case-insensitive local filesystem hides it). Your read on the scope boundary was right too: the remaining uppercase mentions under `.archon/commands/defaults/` and the maintainer workflows are both-casing probes against arbitrary target repos, not references to this repo's file, so leaving them alone was correct. Agreed on docs-follow-reality over renaming the template.

**One change needed before I merge: please drop the new `CONTRIBUTING.md` bullet from `CLAUDE.md` and `AGENTS.md`** — the two added lines. Keep all three link corrections exactly as they are.

The reasoning isn't obvious from outside the repo, so it's worth spelling out. `CLAUDE.md` and `AGENTS.md` aren't documentation that people read when they need it — they're injected into the context of **every agent session, on every run**. Most of those sessions are answering a question, tracing a bug, or editing one file, and will never open a PR. Guidance that only matters at one lifecycle moment shouldn't be paid for by every run that never reaches that moment.

The general rule we apply: **a cost paid by every run has to be justified against every run, not against the one run that benefits.** PR-time process is a point-of-need concern — the human reads `CONTRIBUTING.md` and points their agent at it when it's time to open a PR. That keeps the always-on context for things every session actually needs.

This isn't a criticism of the content; the bullet is accurate and well-written. It's about placement, and it's a rule you had no way to know. If you think there's a gap where an agent genuinely goes wrong without that pointer, say so and we'll talk about where it should live instead — that's a real discussion, just not one that resolves to `CLAUDE.md`.

Two smaller notes:

- Please strip the `🤖 Generated with [Claude Code]` footer from the PR description. We keep AI-attribution out of commit history, and a squash merge pulls the body in.
- On description volume rather than content: a three-file docs fix doesn't need ASCII architecture diagrams or a before/after UX journey. The template's sections are a floor for substantial changes, not a quota — "fixes a 404 in three docs links, verified against the API, no rename" plus the validation line would have been complete. Shorter descriptions get reviewed faster, and yours was easy to verify precisely because the evidence was there; it was just wrapped in a lot of scaffolding.

Push the two-line removal and I'll merge.

bigboy1122:
Done — pushed 3570835.

The two bullets are gone from `CLAUDE.md` and `AGENTS.md`; only the three link corrections remain. The rule makes sense and I would not have guessed it: those files are a per-run cost, so guidance that pays off at one lifecycle moment belongs where it is read at that moment. No gap to argue about — the human pointing their agent at `CONTRIBUTING.md` at PR time is the right seam.

Also stripped the AI-attribution footer and cut the description down to the fix, the API-verified evidence, and the required sections. Point taken on volume — the diagrams were scaffolding a three-line docs fix did not need.

Issue #2075, opened 2026-07-10
Gate approve/reject exposes transient 'failed' status — web UI shows false run failure after clicking approve


MrMophandle:
## Summary

When a paused workflow run is approved (or rejected with `on_reject`) from the **web UI**, the gate handler sets the run's status to `failed` as its internal "resumable" intermediate state, then dispatches the auto-resume. The window between those two operations — and any UI render during it — shows the user a **run that appears to have died immediately after they clicked approve**, even though it resumes and continues healthily about a second later.

Observed twice in one day of real use (same operator, two different workflows): the user approved a gate, saw `failed`, and reported a bug both times. The run was fine both times.

## Reproduction

1. Run any workflow with an `approval:` gate from the web UI (workflow needs `interactive: true`).
2. When the run pauses, click Approve in the approval card.
3. Watch the run status. Timeline recorded from `remote_agent_workflow_events` + run row:
   - `t+0.0s` — `approval_received` event written; run status → `failed` (api.ts approve handler, the resumable state)
   - `t+0.0s` — auto-resume dispatched to the parent conversation
   - `t+~1s` — executor resumes; status → `running`
4. Any UI poll/SSE render inside that window shows a red/failed run. If the page isn't refreshed, the stale `failed` persists indefinitely.

A second aggravating detail: while in the intermediate state, `archon workflow get <id>` shows `Status: failed` with **no error line and no failed node**, which reads as a mystery crash during diagnosis.

## Why it happens

`status: 'failed'` is doing double duty: it means both "this run's execution failed" and "this run is staged for resume." The approve/reject handlers in `packages/server/src/routes/api.ts` intentionally write `failed` because the resume machinery (`/workflow resume`, `resumeWorkflow`) only accepts failed runs.

## Suggested fixes (in rough preference order)

1. **Metadata flag + UI treatment (no schema change):** keep `status: 'failed'` but set `metadata.awaiting_resume: true` in the gate handlers (cleared by the executor on resume). Web UI and CLI render that combination as `resuming…` instead of `failed`. Cheap, backward compatible.
2. **A real `resuming` status:** honest state machine, but touches the status enum, filters, dashboards, and the resume-eligibility checks.
3. At minimum: the web UI approval-card flow could optimistically render "resuming…" for a few seconds after an approve POST succeeds, masking the flash (does not fix stale views or CLI `get`).

## Environment

- Archon v0.5.0, source build (`bun`), dev branch, macOS
[...]

Issue #1549, opened 2026-05-03
/workflow run silently auto-resumes failed runs with stale args, hijacking  fresh requests


ztech-gthb:
> **Maintainer note (updated 2026-06-09):** Confirmed **still LIVE** on `dev` and re-verified against current code. The original report (preserved below) is accurate, but the resume path has been **substantially refactored since this was filed** (#1646 `hydrateResumableRun`, #1830 CAS guards), so the implementation must be built on the *current* model — not the code shape described in the original report. PR #1551 (the original fix, by @ztech-gthb) was closed because its `/workflow resume` re-routing collides with the new CAS model (text-merges clean, semantically broken). **Implement fresh** using the guide below; reuse #1551's UX verbatim and co-credit the author.

---

### The bug, in current-code terms
A fresh `/workflow run X "B"` silently **auto-resumes a prior FAILED run** of `X` in the same conversation, executing in that run's old worktree with its stale persisted `user_message`. The new args ("B") are discarded with no UI/log signal. Compounding: `/workflow abandon` rejects failed runs as terminal, so users can't easily escape.

**Verified root cause (current code):**
- `findResumableRunByParentConversation(name, convId, codebaseId)` filters `status IN ('failed', 'paused')` — `packages/core/src/db/workflows.ts:421`
- Resume detection runs for **every** `/workflow run` (all platforms) inside `dispatchOrchestratorWorkflow` — `packages/core/src/orchestrator/orchestrator-agent.ts` (~the `findResumableRunByParentConversation` call, currently ~line 542). A fresh run finds the prior `failed` run and resumes its `working_path`.

### Architecture you MUST build on (do not regress)
The resume/approval path was reworked after this issue was filed:
- **`hydrateResumableRun(deps, run)`** (#1646) — performs the resume **CAS transition itself** and returns the prepared run (priorCompletedNodes etc.). Do **not** transition the run to `running` *before* dispatch and then rely on `findResumable` — it only matches `failed`/`paused`, so a pre-transitioned run would be missed and a **fresh** run would start. This is exactly the trap PR #1551 fell into.
- **`WorkflowNotResumableError` CAS guards** (#1830) — concurrent resumers (web Resume button / chat re-dispatch / CLI) must not double-claim a worktree. Keep this intact.

There are **three** resume entry points, all flowing through `dispatchOrchestratorWorkflow`. The fix must treat them differently:

| Entry point | Location | Desired behavior |
|-------------|----------|------------------|
[...]

PR #2292, opened 2026-07-27
fix(core): prefer a paused run over a newer failed one on chat resume

### LINKED ISSUE #2289: bug(core): a newer failed run shadows an open approval gate on chat resume
## Summary

- What broke: on chat platforms, a newer **failed** run shadows an older **paused** approval gate when the orchestrator looks for a run to resume, so approving a gate does not resume the gated run.
- When it started (if known): present since the conversation-scoped resume lookup was introduced; made more likely to surface by #2075, which stopped staging approve/reject as a fake `failed` and keeps gated runs `paused`.
- Severity: `major`

## Steps to Reproduce

1. In one chat conversation (Telegram/Slack/Discord/Web) bound to project P, start interactive workflow `W`. It pauses at an approval gate → run A, `status = 'paused'`, `started_at = T1`.
2. In the same conversation and same project, start workflow `W` again (or let a retry run) and let it fail for an unrelated reason → run B, `status = 'failed'`, `started_at = T2 > T1`.
3. Approve the gate of run A from chat (`/workflow approve <A>` or the natural-language approval path).

## Expected vs Actual

- **Expected**: run A resumes in the foreground on its prior worktree, carrying its completed DAG nodes and the `loop_user_input` staged by the approval.
- **Actual**: the lookup returns run B (newer `started_at`), the caller takes the "not paused" branch and sends the *failed-run resume prompt* for run B. Run A stays paused and is never resumed; the user has to resume it by explicit id.

## User Flow

```
  User                        Archon                                  DB
  ────                        ──────                                  ──
  starts W ─────────────────▶ run A paused at gate (T1)
  starts W again ───────────▶ run B fails            (T2 > T1)
  approves gate of A ───────▶ resolveApprovalGate(A)
                              A stays 'paused' (#2075)
                              findResumableRunByParentConversation
                                ORDER BY started_at DESC ────────────▶ returns B  [X]
                              B.status !== 'paused'
                              → buildFailedRunResumePrompt(B)
  sees a prompt about ◀────── (run A never resumes)
  the WRONG run
```

## Environment

- Platform: any chat platform (reproduced conceptually on Telegram; the code path is platform-agnostic)
- Database: PostgreSQL and SQLite alike (the query is dialect-independent)
- OS: Linux

## Logs

```
(no error is logged — the wrong candidate is a silent mis-selection, not a failure)
```

## Impact

- Affected workflows/commands: every interactive workflow with an `approval:` gate driven from chat; `/workflow approve`, the natural-language approval path, and any dispatch of the same workflow name in the same conversation.
- Reproduction rate: Always, whenever a `failed` run of the same workflow + conversation + codebase is newer than the paused one.
- Workaround available? Yes — `/workflow resume <id>` with the explicit paused run id bypasses the selection (the caller short-circuits on `resumeRunId`). Also `/workflow abandon <failed-id>` removes the shadowing candidate.
- Data loss risk? `No` — the paused run is not destroyed, only unreachable via the implicit path.

## Scope

- Module (if known): `core:database` — `findResumableRunByParentConversation` in `packages/core/src/db/workflows.ts`

## Analysis

`findResumableRunByParentConversation` selects the candidate with:

```sql
WHERE workflow_name = $1 AND parent_conversation_id = $2 AND codebase_id = $3
  AND status IN ('failed', 'paused')
ORDER BY started_at DESC
LIMIT 1
```

Status is not part of the ordering, but it fully determines what the single caller does with the row
(`dispatchOrchestratorWorkflow`, `packages/core/src/orchestrator/orchestrator-agent.ts`):

- `paused` → `hydrateResumableRun` → `executeWorkflow` on `working_path` with `priorCompletedNodes`
- otherwise, and not the explicit `resumeRunId` → `buildFailedRunResumePrompt` and **return without dispatching**

So the two statuses are not interchangeable candidates: one auto-resumes, the other deliberately asks the user
first (the gate added in #1549). Ordering them purely by recency lets the "ask first" row win over the
"resume now" row.

Proposed fix: make the ordering status-first, then recency within status —
`ORDER BY CASE WHEN status = 'paused' THEN 0 ELSE 1 END, started_at DESC`. An open gate is legitimately waiting
and is safe to auto-resume; a `failed` candidate remains gated behind the explicit prompt, so no recency
window on `failed` is needed. This mirrors the deliberate, documented ordering in
`getActiveWorkflowRunByPath`, which uses the opposite (`started_at ASC, id ASC`, older-wins) because it answers
a different question — who holds the path lock.

133Felix:
## Summary

- Problem: `findResumableRunByParentConversation` ordered resume candidates by `started_at` alone, so a newer
  `failed` run shadowed an older `paused` approval gate in the same conversation + codebase.
- Why it matters: the two statuses are not interchangeable for the single caller — `paused` gets hydrated and
  resumed in the foreground, `failed` is deliberately gated behind an explicit user prompt (#1549). When the
  wrong row wins, approving a gate resumes nothing and the user gets a resume prompt for an unrelated failed
  run instead.
- What changed: the ordering is now status-first, then recency within status
  (`ORDER BY CASE WHEN status = 'paused' THEN 0 ELSE 1 END, started_at DESC`), plus a doc comment stating the
  guarantee, plus the first test coverage this function has had.
- What did **not** change (scope boundary): the `WHERE` clause, the caller, the resume/hydration path, and
  every other query in the file. No schema change, no new dependency.

### Before

```
  User                        Archon                                  DB
  ────                        ──────                                  ──
  starts W ─────────────────▶ run A pauses at gate (started_at T1)
  starts W again ───────────▶ run B fails          (started_at T2>T1)
  approves gate of A ───────▶ resolveApprovalGate(A)
                              A stays 'paused'  (#2075)
                              findResumableRunByParentConversation
                                ORDER BY started_at DESC ────────────▶ returns B
                              B.status !== 'paused'
                              → buildFailedRunResumePrompt(B)
  sees a prompt about ◀────── run A never resumes
  the WRONG run
```

### After

```
  User                        Archon                                  DB
  ────                        ──────                                  ──
  starts W ─────────────────▶ run A pauses at gate (started_at T1)
  starts W again ───────────▶ run B fails          (started_at T2>T1)
  approves gate of A ───────▶ resolveApprovalGate(A)
                              A stays 'paused'  (#2075)
                              findResumableRunByParentConversation
                                *ORDER BY paused-first, then recency* ─▶ returns A
                              A.status === 'paused'
                              → hydrateResumableRun(A) → executeWorkflow
  sees run A continue ◀────── on its prior worktree, completed nodes kept
```

### Before

```
  orchestrator-agent.ts
    dispatchOrchestratorWorkflow
        │
        │ (workflowName, conversation.id, codebase.id)
        ▼
  db/workflows.ts
    findResumableRunByParentConversation
        │  status IN ('failed','paused')
        │  ORDER BY started_at DESC          ── recency only
        ▼
    ┌───────────────┴────────────────┐
    │                                │
  status === 'paused'          status === 'failed'
    │                                │
    ▼                                ▼
  hydrateResumableRun          buildFailedRunResumePrompt
  → executeWorkflow            → return (no dispatch)
```

### After

```
  orchestrator-agent.ts
    dispatchOrchestratorWorkflow                     (unchanged)
        │
        ▼
  db/workflows.ts
    [~] findResumableRunByParentConversation
        │  status IN ('failed','paused')
        │  === ORDER BY CASE WHEN status='paused' THEN 0 ELSE 1 END, started_at DESC
        ▼
    ┌───────────────┴────────────────┐
    │                                │
  status === 'paused'          status === 'failed'
    │  (now wins whenever one exists)│  (reached only when no gate is open)
    ▼                                ▼
  hydrateResumableRun          buildFailedRunResumePrompt   (both unchanged)
```

**Connection inventory**

| `orchestrator-agent.ts` `dispatchOrchestratorWorkflow` | `db/workflows.ts` `findResumableRunByParentConversation` | unchanged | same signature, same call site — only which row comes back changes |
| `findResumableRunByParentConversation` | `remote_agent_workflow_runs` | **modified** | `ORDER BY` clause only; `WHERE` and params untouched |
| `dispatchOrchestratorWorkflow` | `hydrateResumableRun` / `buildFailedRunResumePrompt` | unchanged | branch logic untouched |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `core`
- Module: `core:database`

## Change Metadata

- Change type: `bug`
- Primary scope: `core`

## Linked Issue

- Closes #2289
- Related #2075 (gates stay `paused` instead of staging a fake `failed` — makes this bug easier to hit)
- Related #1549 (the explicit prompt that gates `failed` candidates, i.e. why the two statuses differ)

```bash
bun run validate
```

Targeted:

```bash
bun test packages/core/src/db/workflows.test.ts
#  86 pass
#  0 fail
```

- Evidence provided: full `bun run validate` run on this branch (Linux, Bun 1.3.11) plus the targeted file run
  above.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? `Yes`
- Config/env changes? `No`
- Database migration needed? `No` — ordering-only change to an existing `SELECT`; `CASE WHEN` is portable
  across PostgreSQL and SQLite, so no dialect helper is involved.

- Verified scenarios: paused-and-failed candidate set in one conversation — the paused gate is now selected;
  single-candidate cases (only paused, only failed, neither) behave exactly as before.
- Edge cases checked: no candidates → still `null`; DB error → still wrapped as
  `Failed to find resumable run by parent conversation: …`; ties within a status still resolve by
  `started_at DESC`.
- What was not verified: multi-hour production soak. The change is a single `ORDER BY` clause on a query with
  one caller, and the branch has been running on a self-hosted instance since the change was written.

- Affected subsystems/workflows: chat-initiated workflow dispatch only. `findResumableRunByParentConversation`
  has exactly one production caller (`dispatchOrchestratorWorkflow`); no other module reads this query or
  depends on its ordering.
- Potential unintended effects: a conversation that has both an open gate and a newer failed run will now
  resume the gate where it previously prompted about the failure. That is the intended fix; the failed run
  remains reachable via `/workflow resume <id>` and `/workflow runs`.
- Guardrails/monitoring for early detection: existing `workflow.*` events are unchanged; a mis-selection would
  surface as the same user-visible symptom as before, inverted.

## Rollback Plan (required)

- Fast rollback command/path: revert this commit — the previous `ORDER BY started_at DESC` is restored with no
  state to migrate.
- Observable failure symptoms: approving a gate produces a resume prompt for a different run instead of
  continuing the gated one.

## Risks and Mitigations

- Risk: a user who *wants* the newest candidate regardless of status now gets the paused one first.
  - Mitigation: explicit addressing is unaffected — `/workflow resume <id>` short-circuits this lookup
    entirely (the caller passes `resumeRunId`), and `/workflow runs` lists every candidate.
- Risk: the SQL fragment is asserted by string matching in the new test, so a future reformat of the query
  could fail the test spuriously.
  - Mitigation: the assertions match the existing style in this test file (`toContain` on query fragments),
    and are split so a failure points at the specific guarantee that broke.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resumable workflow selection by prioritizing paused runs over failed runs.
  * Ensured the most recently started run is selected when multiple runs share the same status.

* **Tests**
  * Added coverage for workflow, conversation, and codebase filtering.
  * Added tests for empty results and database query errors.

### DIFF
--- a/packages/core/src/db/workflows.test.ts
+++ b/packages/core/src/db/workflows.test.ts
@@ -24,6 +24,7 @@ import {
   failWorkflowRun,
   updateWorkflowActivity,
   findResumableRun,
+  findResumableRunByParentConversation,
   resumeWorkflowRun,
   pauseWorkflowRun,
   cancelWorkflowRun,
@@ -625,6 +626,52 @@ describe('workflows database', () => {
 
       await expect(findResumableRun('test', '/path')).rejects.toThrow(
         'Failed to find resumable run: Connection refused'
+      );
+    });
+  });
+
+  describe('findResumableRunByParentConversation', () => {
+    test('scopes by workflow name, parent conversation and codebase', async () => {
+      mockQuery.mockResolvedValueOnce(createQueryResult([mockWorkflowRun]));
+
+      const result = await findResumableRunByParentConversation('piv', 'conv-1', 'codebase-789');
+
+      expect(result).toEqual(mockWorkflowRun);
+      const [query, params] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).toContain('workflow_name = $1');
+      expect(query).toContain('parent_conversation_id = $2');
+      expect(query).toContain('codebase_id = $3');
+      expect(params).toEqual(['piv', 'conv-1', 'codebase-789']);
+    });
+
+    test('prefers a paused run over a newer failed one (paused-first ordering)', async () => {
+      mockQuery.mockResolvedValueOnce(createQueryResult([]));
+
+      await findResumableRunByParentConversation('piv', 'conv-1', 'cb');
+
+      const [query] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).toContain("status IN ('failed', 'paused')");
+      // Status is the primary sort key: an open gate auto-resumes, while a
+      // failed candidate is gated behind an explicit user prompt. Ordering by
+      // started_at alone lets a newer failure shadow an older waiting gate.
+      expect(query).toContain("ORDER BY CASE WHEN status = 'paused' THEN 0 ELSE 1 END");
+      // Recency still breaks ties within a status.
+      expect(query).toContain('started_at DESC');
+    });
+
+    test('returns null when no run matches', async () => {
+      mockQuery.mockResolvedValueOnce(createQueryResult([]));
+
+      const result = await findResumableRunByParentConversation('piv', 'conv-1', 'cb');
+
+      expect(result).toBeNull();
+    });
+
+    test('throws on database error', async () => {
+      mockQuery.mockRejectedValueOnce(new Error('Connection refused'));
+
+      await expect(findResumableRunByParentConversation('piv', 'conv-1', 'cb')).rejects.toThrow(
+        'Failed to find resumable run by parent conversation: Connection refused'
       );
     });
   });
--- a/packages/core/src/db/workflows.ts
+++ b/packages/core/src/db/workflows.ts
@@ -639,6 +639,15 @@ export async function findResumableRun(
  * Used by the orchestrator (all platforms) to detect approved runs that need foreground resume
  * on the prior run's worktree. Codebase scope prevents cross-project resume on persistent
  * chat conversation IDs (Telegram chat_id, Slack thread, etc.).
+ *
+ * Ordering is status-first, then recency WITHIN a status — not bare recency. The two statuses
+ * are not interchangeable candidates for the caller: a `paused` run is an open gate that is
+ * legitimately waiting and gets hydrated and resumed, while a `failed` one is deliberately gated
+ * behind an explicit user prompt first (#1549). Ordering purely by `started_at` therefore lets a
+ * newer failure shadow an older open gate, and approving that gate resumes nothing.
+ *
+ * Contrast with getActiveWorkflowRunByPath below, which sorts the opposite way (older-wins) —
+ * it answers "who took the path lock first", a different question.
  */
 export async function findResumableRunByParentConversation(
   workflowName: string,
@@ -652,7 +661,7 @@ export async function findResumableRunByParentConversation(
          AND parent_conversation_id = $2
          AND codebase_id = $3
          AND status IN ('failed', 'paused')
-       ORDER BY started_at DESC
+       ORDER BY CASE WHEN status = 'paused' THEN 0 ELSE 1 END, started_at DESC
        LIMIT 1`,
       [workflowName, parentConversationId, codebaseId]
     );


Wirasm:
## Review: #2292

Reviewed **together with #2294**, in the order **#2292 → #2294**. They are coupled: #2294's problem statement quotes the candidate set of the very function this PR re-orders, and its escape hatch matters more once this ordering lands. Landing order and the interaction are at the bottom.

**Verdict: approve.** The diagnosis is right, the fix is minimal and correct, the ordering is portable across both dialects, and the doc comment is genuinely useful. Two suggestions, no blockers.

---

### What I verified (rather than took from the description)

**The call site behaves as claimed.** `packages/core/src/orchestrator/orchestrator-agent.ts:802-817`:

```ts
if (resumableRun?.working_path) {
  if (resumableRun.status !== 'paused' && resumableRun.id !== options?.resumeRunId) {
    // buildFailedRunResumePrompt(...) ; return   ← failed: prompt, no dispatch
  }
  // hydrateResumableRun → executeWorkflow        ← paused: silent foreground resume
```

So the two statuses really are non-interchangeable for the single caller, exactly as the PR says. A `failed` row short-circuits with a prompt; a `paused` row is hydrated and resumed. Ordering by recency alone lets the former shadow the latter.

**The ordering is behaviourally correct, not just textually.** The unit tests assert SQL substrings, so I ran the real thing against a real `bun:sqlite` adapter (throwaway file, not proposed for the branch) using the harness already in `workflows.resume-cas.integration.test.ts`:

| case | expected | result |
|---|---|---|
| paused @ 2026-01-01, failed @ 2026-06-01 | the paused one | pass |
| second paused @ 2026-03-01 added | newest paused | pass |
| both paused rows cancelled | newest failed | pass |

`CASE WHEN` is plain SQL-92 — no dialect helper needed, and it works on both backends. Worth noting for the maintainers: `workflows.resume-cas.integration.test.ts` already drives `findResumableRunByParentConversation` through real SQLite twice, so the new clause picks up real-dialect execution coverage for free.

**The new tests are load-bearing.** I mutated the `ORDER BY` back to `started_at DESC`:

```
(fail) findResumableRunByParentConversation > prefers a paused run over a newer failed one
 85 pass, 1 fail
```

They do fail if the fix reverts. Inverting `THEN 0 ELSE 1` or flipping `DESC` would also fail them.

**Edge cases you asked me to think about:**

- *Two paused runs* — newest paused wins; unchanged from before within a status. Correct.
- *Gate already resolved* — benign, and in fact the point. Per #2075 an approved gate stays `paused` with `metadata.approval.resolved` set and the resume machinery is *supposed* to collect it. A rejected gate without `on_reject` goes to `cancelled` (out of the candidate set); with `on_reject` it stays `paused` and should be collected. All three land right.
- *A paused run with no `working_path`* — not reachable; `working_path` is `NOT NULL` on `remote_agent_workflow_runs`.

---

### Suggestions

**S1 — paused-first is unbounded by recency across statuses; say so in the doc comment.**

This is the one real behavioural consequence and it isn't in the comment or the blast-radius section. After this change an arbitrarily old `paused` gate always beats an arbitrarily recent `failed` run — and the paused branch resumes **silently in the foreground**, while the failed branch is deliberately gated behind a user prompt (#1549). So the change strictly widens the no-prompt path.

I think that's the right call — an open gate is a genuine outstanding obligation, and the sibling `findResumableRun` has no staleness bound on `failed`/`paused` either (`ORPHAN_RESUME_STALE_DAYS` only bounds the `running`-orphan arm). But on a long-lived chat id (Telegram `chat_id`, a Slack thread) a forgotten gate can now shadow every later attempt at that workflow indefinitely, and #2294 is the escape hatch for exactly that. Two sentences in the doc comment noting "no staleness bound — an old gate outranks a recent failure by design; `/reset` or `/workflow abandon <id>` is the way out" would close the loop.

**S2 — consider one behavioural assertion alongside the string ones.**

You already flagged the reformat-fragility yourself. Since `workflows.resume-cas.integration.test.ts` is already wired to real SQLite and already imports this function, one added case there would survive any query reformat:

```ts
test('an open gate outranks a newer failed run', async () => {
  await seedPausedRun('gate-old', 'wf-order', { nodeId: 'r', message: '?', type: 'approval', resolved: null });
  // seed a newer 'failed' row for wf-order on conv-1/cb-1
  expect((await findResumableRunByParentConversation('wf-order', 'conv-1', 'cb-1'))?.id).toBe('gate-old');
});
```

Not blocking — the string assertions do catch the revert today.

---

### Answering "any other query in that file with the same latent assumption?"

Yes, one: **`findResumableRun`** (`packages/core/src/db/workflows.ts:610-624`). It matches a *wider* mixed set via `resumableStatusClause` — `failed`, `paused`, and stale `running` orphans — and then orders by `started_at DESC` alone, so the identical shadowing is possible there: a newer failed run on the same `working_path` can hide an older open gate. Its consumers are the CLI resume lookup (`packages/cli/src/commands/workflow.ts:1094`) and the `IWorkflowStore` seam (`packages/core/src/workflows/store-adapter.ts:49`) that the `workflow:` sub-run resume path uses. Correctly out of scope here, but it looks like a follow-up issue worth filing.

The others are clean:
- `getPausedWorkflowRun` (:399) and `getActiveWorkflowRun` (:377) are single-status, so ordering can't mis-select across statuses.
- `getActiveWorkflowRunByPath` (:439) sorts `started_at ASC, id ASC` deliberately — your doc comment already calls this out correctly, and it *is* a different question (who took the path lock first).

---

### Interaction with #2294, and landing order

**Land #2292 first, then #2294.** Both are safe alone against `dev`; neither depends on the other to be correct. The reason for this order is that #2294 is the user-facing escape hatch from the situation this PR's stricter preference makes more reachable (S1 above) — shipping the preference first and the escape hatch immediately after keeps the pair coherent.

**They do not fight.** #2294's abandon is *set*-based, not order-based: it cancels every matching row, so this `ORDER BY` is irrelevant to it. Walking the exact scenario through both — one paused run at T1, one newer failed run at T2, same conversation + codebase:

- before `/reset`: this PR returns the paused T1 and it resumes (previously: prompt about T2). Fixed.
- after `/reset`: both rows are `cancelled`, the lookup returns `null`, `cwd` is `null` → a genuinely fresh run in `codebase.default_cwd`. Correct.

**One mechanical note — a merge conflict is guaranteed, whichever lands second.** Both PRs add an import to the same list in `packages/core/src/db/workflows.test.ts` and both insert a new `describe` block at the same anchor (immediately before `describe('getActiveWorkflowRunByPath')`). It is purely additive: keep both imports and both blocks. I resolved it that way locally and confirmed the combined state is healthy — `bun test src/db/workflows.test.ts src/handlers/command-handler.test.ts` → **210 pass / 0 fail**, and `bun run type-check` clean across all packages.

`#2299` (approved, unmerged) also touches `workflows.ts` and `workflows.test.ts`, but its hunks sit in `updateWorkflowRun` (~:790-830). I test-merged `dev + #2299 + #2292` — clean, no conflict in either direction. Not a risk for this PR.

---

### Strengths

The failure analysis is precise and the scope discipline is exemplary — `WHERE` untouched, caller untouched, one clause changed. The doc comment earns its place by explaining *why* the two statuses are not interchangeable and, unusually, by contrasting with the deliberately-opposite ordering in `getActiveWorkflowRunByPath` right below it; that's the kind of note that prevents someone "fixing" the wrong one later. And this function had zero test coverage before — adding it while fixing it is the right instinct.

Issue #2274, opened 2026-07-26
(Feature): allow slack channel to project mapping + make archon slack channel aware


hidegh:
## Problem

- When workingin slack, we usually have one channel for a given project.
- We'd like to be able to define that channel-name - project alias mapping for Archon (so we don't need to set the project for every new thread we start).
- Also make archon slack-channel aware (not impl. in the provided patch)

## Proposed Solution

Solution plan done via Slack, working aft applying changes on the local project.

### Before (current)

```
/setproject <name>
```

If user forgot to set the project, s/he haas to rely on Archon (less determenistics) to decide correctly.

### After (proposed)

Archon auto selects the project based on a mapping.

```
(What does the ideal flow look like? Highlight new steps with [+] or asterisks.)
```

## Alternatives Considered

| Alternative | Pros | Cons | Why not chosen |
|-------------|------|------|----------------|
| | | | |

## Scope

- Package(s) likely affected: `core|adapters`
- Breaking change? (`No`)
- Database changes needed? (`No`)
- New external dependencies? (`No`)

| Concern | File |
|---|---|
| Global config shape | `packages/core/src/config/config-types.ts` |
| Config merge logic | `packages/core/src/config/config-loader.ts` |
| Message-handling context type | `packages/core/src/types/index.ts` |
| Main message handler (uses the context) | `packages/core/src/orchestrator/orchestrator-agent.ts` |
| Slack event intake (the interception point) | `packages/server/src/index.ts` |
| Project lookup by name | `packages/core/src/db/codebases.ts` (`findCodebaseByName`, already exists) |

## Security Considerations

- New permissions/capabilities? (`No`)
- New external network calls? (`Yes`, limited by caching)
- Secrets/tokens handling? (`No`)
- If any `Yes`, describe: ...

## Definition of Done

- [ ] Archon sets the project correctly, for this at least 2 project has to be set up, 2 slack channels must exists, each mapped to a diff. project.
- [ ] Documentation updated (if user-facing) - see attached MD file.

hidegh:
# Slack - code tweak

Register
- via command: /register-project alias path
- or via UI (the UI contains the alias and the path)

```YAML
slack:  
  channelProjects:
    ai-tests: hidegh/ai-tests
```

Test (need at least 2 projects for this):
@archon what's the current project?

## Auto-Binding a Slack Channel to an Archon Project

[...]

PR #1937, opened 2026-06-09
fix(core): /setproject writes to conversation DB id, not platform id

aryrabelo:
## Summary

- **Problem:** `/setproject <name>` failed with `Error: Conversation not found: <id>` on every chat platform. `handleSetProject` passed the platform conversation id (Telegram chat id, `owner/repo#n`, Slack `thread_ts`, `web-…` string) to `db.updateConversation()`, which keys on the conversations-table **DB primary key** — a generated UUID/hex that never equals the platform id on any platform.
- **Why it matters:** Binding a conversation to a registered codebase is the entry point for multi-project use from chat platforms; the command was broken everywhere it could be typed. (The web UI never surfaced it only because it scopes conversations via `POST /api/conversations {codebaseId}` at creation time — a different flow that never invokes `/setproject` — not because of any id equality.)
- **What changed:** `handleSetProject` now receives `conversation.id` (the DB primary key, matching every other `updateConversation` call site); parameter renamed to `conversationDbId`. Regression test added where the DB id differs from the platform id. Maintainer follow-up commit: the shared `makeConversation()` test fixture now generates a DB id that differs from `platform_conversation_id` (the identical defaults were masking this whole bug class — every id-flow assertion in the file now proves the right id reaches the right layer), and `/setproject` was added to CLAUDE.md's deterministic-command list (a pre-existing doc gap).
- **What did NOT change (scope boundary):** No change to name resolution, the dispatch surface, the reply path (still addressed to the platform conversation id), session lifecycle (the known stale-session-after-project-switch gap predates this PR and is being addressed separately), or any other command handler (audited: `/setproject` was the sole outlier — all other `updateConversation` call sites already pass `conversation.id`).

### Before

```
User (Telegram)            Archon
────                       ──────
/setproject brainlist ───▶ resolve codebase  -> ok
                           updateConversation(WHERE id = '<telegram-chat-id>')
                           -> 0 rows matched
sees error  ◀───────────── "Error: Conversation not found: <chat-id>"
```

### After

```
User (Telegram)            Archon
────                       ──────
/setproject brainlist ───▶ resolve codebase  -> ok
                           updateConversation(WHERE id = [conversation.id])  *DB pk*
                           -> 1 row updated
sees success ◀──────────── "Project set to <name> / Working directory: <cwd>"
```

### Before

```
handleMessage(conversationId=<platform id>, conversation=<row>)
   └─ handleSetProject(message, conversationId)      --x--  [wrong id]
[...]

PR #2293, opened 2026-07-27
feat(orchestrator): /setproject none detaches a conversation back to neutral

### LINKED ISSUE #2290: feat(orchestrator): detach a conversation from its project safely
## Problem

A conversation can be bound to a project with `/setproject <name>`, but it cannot return to neutral without deleting the registered project. This matters most on persistent chat adapters: a Telegram chat or Discord channel can keep the same conversation row for months.

The binding is operational, not cosmetic. A bound conversation receives the project's prompt, working directory, environment variables, workflows, workspace synchronization, and project-scoped tools on every later turn.

Deleting or unregistering the project is not the inverse we need. Other conversations may still use it.

## Desired outcome

Add an explicit conversation-scoped detach command:

```text
/detach-project "<current-project-name>"
```

The supplied name must exactly identify the project currently bound to that conversation. A missing name, another project, or an ambiguous name must fail without changing state.

This command detaches one conversation. It does not remove the registered project. The existing `/remove-project <name>` command keeps its current install-wide project-removal meaning.

Do not add `/setproject none`, `/setproject clear`, `/setproject -`, or other sentinel values. Those strings are valid project names and make an accidental detach too easy.

## Safety gate

Detaching must refuse while the conversation still owns work or an execution environment.

Block the command when any associated workflow run is:

- pending or running;
- paused for approval, input, time, or an external event; or
- failed but resumable.

Also block while `isolation_env_id` points to an attached worktree or container environment. Clearing that pointer first would make chat-based cleanup unreachable and could strand uncommitted work.

The blocker lookup must cover runs linked through either `conversation_id` or `parent_conversation_id`; Web worker runs can use the latter. The refusal should list the exact blocking run IDs and statuses, and identify the attached environment when present. It should direct the user to resolve runs through `manage_run`, the CLI, or the Web UI, and clean the environment through the existing isolation controls.

The detach command must not cancel, abandon, resume, or delete anything implicitly. Run lifecycle stays owned by `manage_run` and the existing explicit run controls; environment lifecycle stays owned by isolation cleanup. The deterministic detach path should use a narrow typed conversation-blocker query rather than route detection through the project-wide `manage_run` tool.

Treat “no owned work” as a hard invariant, not a best-effort warning. The blocker check and binding transition must share one transaction or equivalent ownership boundary so a run cannot start between a successful check and the detach write.

## Successful detach

Once no blocking work or environment remains:

1. Deactivate the active AI session as part of the successful transition so provider session context cannot leak into the neutral conversation.
2. Clear `codebase_id`, `cwd`, and `isolation_env_id` on the conversation row, using the conversation database ID rather than the platform conversation ID.
3. Confirm that the conversation is neutral and that the registered project remains available elsewhere.

The operation must fail without clearing the binding if validation, session deactivation, or persistence fails.

## Scope

- Support the command where a null binding is durable: Web, Telegram, and independently scoped/top-level Slack or Discord conversations.
- Refuse it on GitHub, GitLab, and Gitea forge conversations because those adapters deliberately rebind a null conversation to the repository on the next event. Also refuse it on child threads that would immediately inherit a bound parent. Persisting a separate “explicitly detached” marker would add new state and is outside this issue.
- `archon chat` creates a fresh conversation for each invocation, so it has no durable detach use case even though it shares the parser.
- No database migration is needed; the binding columns are already nullable.
- No new agent tool is required. The slash command expresses the user's explicit detach action, while `manage_run` continues to own workflow lifecycle decisions.
- Do not change `/reset`, `/setproject <name>`, project registration, or `/remove-project <name>` as part of this issue.

## Definition of done

- [ ] `/detach-project "<current-project-name>"` detaches only the current conversation.
- [ ] Missing, incorrect, or ambiguous project names fail without mutation.
- [ ] Projects named `none`, `clear`, or `-` remain selectable.
- [ ] Pending, running, paused, and failed-but-resumable runs each block detachment and are reported to the user.
- [ ] An attached worktree or container environment blocks detachment and remains reachable for cleanup.
- [ ] No run or environment lifecycle state is changed by the detach attempt.
- [ ] The blocker query covers both direct and parent-conversation run ownership.
- [ ] The blocker check and detach transition cannot race with run creation.
- [ ] Auto-rebinding forge conversations and inheriting child threads refuse with a clear explanation.
- [ ] The active AI session is deactivated before the binding is cleared.
- [ ] The successful write targets the conversation database ID and clears `codebase_id`, `cwd`, and `isolation_env_id`.
- [ ] Help text and command documentation distinguish detaching a conversation from removing a registered project.
- [ ] Focused tests cover every refusal and the successful persistent-chat flow.

## Related

- #2293 established the persistent-chat need but used sentinel syntax and allowed detachment while the conversation still owned work or an environment. It is superseded by this contract.
- #2731 provides the existing resumable-run lifecycle primitive used by `/reset`; this command should detect and report those runs, not mutate them.
- #2274 covers automatic Slack project binding. It is complementary rather than a duplicate.

133Felix:
## Summary

- Problem: a conversation can be bound to a project but never unbound. `/setproject <name>` only rebinds,
  `/reset` preserves `codebase_id` by design, and the only thing that actually clears the binding is deleting
  the project (the `deleteCodebase` cascade).
- Why it matters: persistent-conversation platforms have one permanent row — Telegram maps a whole chat,
  Discord a whole channel. The row bound today is still bound months later, and being bound is not cosmetic:
  it selects the project-scoped system prompt, injects the `manage_run` tool, resolves `cwd` to that project,
  merges its codebase env vars, and syncs its workspace on every turn.
- What changed: `/setproject none` (also `clear`, `-`, case-insensitive) nulls `codebase_id`, `cwd` and
  `isolation_env_id`, returning the conversation to exactly the state a fresh one starts in. Plus `/help`
  text and the command reference.
- What did **not** change (scope boundary): `/setproject <name>` binding behaviour, `/reset`, the resolution
  tiers, and the DB schema (all three columns are already nullable). No new dependency.

### Before

```
  /setproject archon ─────────▶ conversation bound to "archon"
  ...weeks later...
  "what's the weather" ───────▶ still project-scoped: syncs the workspace,
                                loads archon env vars, biases routing
  /reset ─────────────────────▶ session cleared, binding PRESERVED
  /setproject ??? ────────────▶ no way to say "no project"
```

### After

```
  /setproject archon ─────────▶ conversation bound to "archon"
  /setproject none ───────────▶ [+] codebase_id / cwd / isolation_env_id = NULL
                                [+] "Project cleared — this conversation is now neutral."
  "what's the weather" ───────▶ generic orchestrator prompt, no workspace sync,
                                no project env vars
  /setproject archon ─────────▶ bound again whenever you want
```

### Before

```
  command dispatch (orchestrator-agent.ts)
        │  command === 'setproject'
        ▼
  handleSetProject(message, conversation)
        │
        ├── args.length < 1 ────────────▶ usage string
        │
        └── resolveCodebaseName(...) ───▶ codebaseDb.listCodebases()
                    │
                    ├── not found ──────▶ "Project X not found"
                    └── found ──────────▶ deactivate session
                                          db.updateConversation(conversation.id, {
                                            codebase_id, cwd: null, isolation_env_id: null })
```

### After

```
  command dispatch (orchestrator-agent.ts)                       (unchanged)
        │  command === 'setproject'
        ▼
  [~] handleSetProject(message, conversation)
        │
        ├── args.length < 1 ────────────▶ usage string (mentions the detach form)
        │
        ├── [+] single reserved token (none|clear|-)
        │        === db.updateConversation(conversation.id, {
        │              codebase_id: null, cwd: null, isolation_env_id: null })
        │        ── returns BEFORE any codebase lookup ──x-- codebaseDb.listCodebases()
        │
        └── resolveCodebaseName(...) ───▶ codebaseDb.listCodebases()   (unchanged)
                    │
                    ├── not found ──────▶ "Project X not found"
                    └── found ──────────▶ deactivate session + rebind
```

**Connection inventory**

| command dispatch | `handleSetProject` | unchanged | same signature |
| `handleSetProject` | `db.updateConversation` | **modified** | new call path writing three NULLs; the existing bind call is untouched |
| `handleSetProject` | `codebaseDb.listCodebases` | **modified** | now skipped entirely on the clear path (early return) |
| `handleSetProject` | `sessionDb` / `safeDeactivateSession` | unchanged | the clear path does not touch sessions — see "not verified" below |
| `/help` text, `docs/reference/commands.md` | — | **modified** | document the detach form |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `core`
- Module: `core:orchestrator`

## Change Metadata

- Change type: `feature`
- Primary scope: `core`

## Linked Issue

- Closes #2290
- Related #2274 — that request is about auto-**binding** Slack threads by channel; this is the inverse
  (**unbinding** an existing conversation). Complementary, not overlapping.
- Related #1937 — established that `/setproject` must write by the conversation DB id; the new path follows
  the same rule and the tests assert it.

```bash
bun run validate
```

Targeted:

```bash
bun test packages/core/src/orchestrator/orchestrator-agent.test.ts
#  214 pass
#  0 fail
```

- Evidence provided: full `bun run validate` on this branch (Linux, Bun 1.3.11) plus the targeted file run.

## Security Impact (required)

- New permissions/capabilities? `No` — it removes scope rather than granting any
- Secrets/tokens handling changed? `No` — detaching stops that codebase's env vars from being injected, which
  narrows exposure
- File system access scope changed? `No` — clearing `cwd` returns the conversation to the neutral
  `~/.archon/workspaces/` root that unbound conversations already use

## Compatibility / Migration

- Backward compatible? `Yes` — purely additive; `/setproject <name>` is unchanged
- Config/env changes? `No`
- Database migration needed? `No` — `codebase_id`, `cwd` and `isolation_env_id` are already nullable, and the
  resulting row is the same shape as a brand-new conversation

- Verified scenarios: bind → `/setproject none` → conversation is neutral (generic prompt, no project env,
  no workspace sync) → rebind works. `clear`, `-` and mixed case behave identically.
- Edge cases checked: a multi-word project name beginning with a reserved token still resolves as a project
  (covered by a test); the clear path never calls `listCodebases`, so it works even with zero projects
  registered; the write targets the conversation DB id, not the platform id.
- What was not verified: the clear path intentionally does **not** deactivate the active AI session, on the
  grounds that detaching is not a session reset (`/reset` is). If maintainers would rather have the session
  deactivated too, that is a two-line addition — happy to change it.

- Affected subsystems/workflows: `/setproject` handling only. Downstream behaviour of a neutral conversation is
  the pre-existing unbound path — no new branches were added to prompt building, tool injection, cwd
  resolution or workflow discovery.
- Potential unintended effects: a user with a project literally named `none`, `clear` or `-` cannot select it
  by that bare name. Mitigated by requiring the single-argument form and documented inline; any other
  resolution tier (prefix, substring, owner-qualified) still reaches it.
- Guardrails/monitoring for early detection: the clear path logs `project.cleared` with the conversation id,
  so an unexpected detach is traceable.

## Rollback Plan (required)

- Fast rollback command/path: revert this commit. Conversations already detached simply stay detached — that
  is the same state new conversations start in, so nothing needs repair.
- Observable failure symptoms: `/setproject none` resolving as a project-name lookup instead of detaching.

## Risks and Mitigations

- Risk: a project named exactly `none` / `clear` / `-` becomes unselectable by that bare name.
  - Mitigation: only the single-argument form is reserved; other tiers still resolve it, and the behaviour is
    documented in `/help` and the command reference.
- Risk: users expect `/setproject none` to also clear the session, and are surprised it does not.
  - Mitigation: the reply states exactly what happened; `/reset` remains the session command. Easy to change
    if maintainers prefer the opposite default.

## Summary by CodeRabbit

* **New Features**
  * Added `/setproject none` to detach a conversation from its current project and return it to a neutral state.
  * Added `clear` and `-` as alternative commands for detaching a project.
  * Project-clearing commands are case-insensitive.
  * Active sessions are deactivated when a project is detached.

* **Documentation**
  * Updated command help and reference documentation with the new project-detachment options.

### DIFF
--- a/packages/core/src/handlers/command-handler.ts
+++ b/packages/core/src/handlers/command-handler.ts
@@ -1064,7 +1064,7 @@ Talk naturally — the orchestrator routes your requests to the right workflow a
 - \`/register-project <name> <path>\` — Register a local project
 - \`/update-project <name> <new-path>\` — Update a project's path
 - \`/remove-project <name>\` — Remove a registered project
-- \`/setproject <name>\` — Bind this conversation to a registered project
+- \`/setproject <name>\` — Bind this conversation to a registered project (\`/setproject none\` to detach and go neutral)
 
 **Session**
 - \`/status\` — Show current session and project info
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -3560,6 +3560,170 @@ describe('handleMessage — /setproject dispatch', () => {
     expect(mockUpdateConversation).not.toHaveBeenCalled();
     expect(platform.sendMessage).toHaveBeenCalledWith('conv-1', expect.stringContaining('Usage'));
   });
+
+  // ─── clear-to-neutral (/setproject none) ────────────────────────────────────
+
+  test('/setproject none clears the project scope on the DB conversation id', async () => {
+    // Same regression guard as the rebind path: the write must target
+    // conversation.id, never the platform conversation id (Telegram chat ids and
+    // GitHub owner/repo#n are not the conversations-table primary key).
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(
+        makeConversation({
+          id: 'db-conv-clear',
+          platform_type: 'telegram',
+          platform_conversation_id: '40865006',
+          codebase_id: 'cb-old',
+          cwd: '/old/worktree',
+        })
+      )
+    );
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['none'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, '40865006', '/setproject none');
+
+    expect(mockUpdateConversation).toHaveBeenCalledWith('db-conv-clear', {
+      codebase_id: null,
+      cwd: null,
+      isolation_env_id: null,
+    });
+    expect(platform.sendMessage).toHaveBeenCalledWith(
+      '40865006',
+      expect.stringContaining('neutral')
+    );
+  });
+
+  test('/setproject none deactivates the active AI session before clearing', async () => {
+    // Without this the stale assistant_session_id survives the detach and is
+    // resumed against the now-neutral cwd — on Claude that throws, and since the
+    // turn throws, tryPersistSessionId never clears it: the conversation wedges.
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(makeConversation({ id: 'db-conv-clear', codebase_id: 'cb-old' }))
+    );
+    mockGetActiveSession.mockImplementation(() =>
+      Promise.resolve({ id: 'session-123', conversation_id: 'db-conv-clear', active: true })
+    );
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['none'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject none');
+
+    expect(mockDeactivateSession).toHaveBeenCalledWith('session-123', 'project-changed');
+    expect(mockUpdateConversation).toHaveBeenCalledWith('db-conv-clear', {
+      codebase_id: null,
+      cwd: null,
+      isolation_env_id: null,
+    });
+  });
+
+  test('/setproject CLEAR clears regardless of case', async () => {
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(makeConversation({ id: 'db-conv-clear', codebase_id: 'cb-old' }))
+    );
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['CLEAR'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject CLEAR');
+
+    expect(mockUpdateConversation).toHaveBeenCalledWith('db-conv-clear', {
+      codebase_id: null,
+      cwd: null,
+      isolation_env_id: null,
+    });
+  });
+
+  test('/setproject - clears (documented alias, otherwise silently regressable)', async () => {
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(makeConversation({ id: 'db-conv-clear', codebase_id: 'cb-old' }))
+    );
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['-'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject -');
+
+    expect(mockUpdateConversation).toHaveBeenCalledWith('db-conv-clear', {
+      codebase_id: null,
+      cwd: null,
+      isolation_env_id: null,
+    });
+  });
+
+  test('clears with zero projects registered (returns before listCodebases)', async () => {
+    // Detaching must not be breakable by an empty or broken project list —
+    // that is exactly when a user most wants out.
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(makeConversation({ id: 'db-conv-clear', codebase_id: 'cb-old' }))
+    );
+    mockListCodebases.mockImplementation(() => Promise.reject(new Error('db down')));
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['none'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject none');
+
+    expect(mockListCodebases).not.toHaveBeenCalled();
+    expect(mockUpdateConversation).toHaveBeenCalledWith('db-conv-clear', {
+      codebase_id: null,
+      cwd: null,
+      isolation_env_id: null,
+    });
+  });
+
+  test('multi-token argument falls through to name resolution, not the clear path', async () => {
+    // The single-token gate is what keeps `/setproject none of your business`
+    // a (failing) project lookup rather than a silent detach.
+    mockListCodebases.mockImplementation(() => Promise.resolve([makeCodebase('project-a')]));
+    mockParseCommand.mockReturnValue({
+      command: 'setproject',
+      args: ['none', 'of', 'your', 'business'],
+    });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject none of your business');
+
+    expect(mockUpdateConversation).not.toHaveBeenCalled();
+    const msg = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0]?.[1] as string;
+    expect(msg).toContain('none of your business');
+  });
+
+  test('notes the detached worktree when clearing an isolation-bound conversation', async () => {
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(
+        makeConversation({
+          id: 'db-conv-clear-wt',
+          codebase_id: 'cb-old',
+          cwd: '/old/worktree',
+          isolation_env_id: 'env-old',
+        })
+      )
+    );
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['none'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject none');
+
+    const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls
+      .map(c => String(c[1]))
+      .join('\n');
+    expect(sent).toContain('neutral');
+    expect(sent).toContain('previous worktree was detached');
+  });
+
+  test('omits the worktree note when clearing a conversation with no isolation env', async () => {
+    mockGetOrCreateConversation.mockImplementation(() =>
+      Promise.resolve(makeConversation({ id: 'db-conv-clear', codebase_id: 'cb-old' }))
+    );
+    mockParseCommand.mockReturnValue({ command: 'setproject', args: ['none'] });
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/setproject none');
+
+    const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls
+      .map(c => String(c[1]))
+      .join('\n');
+    expect(sent).toContain('neutral');
+    expect(sent).not.toContain('previous worktree');
+  });
 });
 
 // ─── handleMessage — /update-project dispatch (issue #2085) ───────────────────
--- a/packages/core/src/orchestrator/orchestrator-agent.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.ts
@@ -2638,7 +2638,16 @@ async function handleRemoveProject(message: string): Promise<string> {
 }
 
 /**
- * Handle /setproject command. Four effects:
+ * Reserved single-token arguments to `/setproject` that detach the conversation
+ * from its project instead of binding it to a named one.
+ */
+const SETPROJECT_CLEAR_KEYWORDS = new Set(['none', 'clear', '-']);
+
+/**
+ * Handle /setproject command. Two modes, both writing by the DB primary key
+ * (conversation.id), never the platform conversation id.
+ *
+ * REBIND — `/setproject <name>`. Four effects:
  * 1. Binds the conversation to the resolved codebase (writes `codebase_id`).
  * 2. Clears `cwd` — the project root remains codebase.default_cwd;
  *    conversation.cwd is only an explicit runtime override.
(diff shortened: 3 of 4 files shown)


Wirasm:
## Review: `/setproject none` detaches a conversation back to neutral

Thanks for this — the problem statement is exactly right, and the framing is unusually careful. The
asymmetry is real: `/setproject <name>` only rebinds, `/reset` preserves `codebase_id` by design, and
the only thing that clears the binding today is the `deleteCodebase` cascade. On Telegram and Discord,
where one permanent row maps a whole chat or channel, that means a conversation stays pinned to the
first project it ever touched. Filling in the inverse of an operation that already exists is the right
shape of change, and the early return before `listCodebases()` is a genuinely good call — detaching
works with zero projects registered and can't be broken by a bad project list.

Two things should change before merge. One of them you already identified yourself in the
"What was not verified" section, and your instinct there was correct.

**Verdict: request changes** (I1 is the blocker; I2 is a small addition alongside it).

---

### Validation

Reproduced locally on `1dd9a45`, in a clean worktree:

| Check | Result |
|---|---|
| `bun run type-check` | PASS (10/10 packages) |
| `bun run format:check` | PASS |
| `bun test packages/core/src/orchestrator/orchestrator-agent.test.ts` | PASS — 214 pass, 0 fail |

Matches your reported evidence, and CI is green on ubuntu / windows / docker-build.

---

## Important

### I1 — The clear path doesn't deactivate the active AI session, and on Claude that wedges the conversation

`packages/core/src/orchestrator/orchestrator-agent.ts:2595-2603`

You raised this yourself and offered to change it. It should be changed — the consequence is worse than
a cosmetic inconsistency, so here is the evidence in case it's useful.

The rebind path 25 lines below does deactivate, and its comment explains precisely why:

```ts
// orchestrator-agent.ts:2622-2630
// Deactivate the old session BEFORE rebinding the conversation ... The reverse
// order would leave a rebound conversation with the old project's session
// still active — resuming old-project context under the new project's cwd.
const session = await sessionDb.getActiveSession(conversation.id);
if (session) {
  await safeDeactivateSession(session.id, 'setproject');
}
```

Detach hits the same hazard. On the next turn `getActiveSession` (`db/sessions.ts:27-33`) selects on
`conversation_id AND active` only — there is no project predicate, and `sessions.codebase_id` is
written but never read anywhere in the repo. So the stale row is picked up at
`orchestrator-agent.ts:1467` and its `assistant_session_id` is passed to the provider as a resume token
alongside the *new* neutral cwd (`~/.archon/workspaces`, via `ensureArchonWorkspacesPath()` at
`:1462`).

The cwd itself resolves correctly — that part of your claim holds. What breaks is the resume:

- **Claude (the default assistant)** — its session store is cwd-keyed, and the repo already documents
  the contract at `packages/providers/src/claude/provider.ts:1359-1361`: *"Claude resumes-or-errors: an
  invalid resume id throws (and is retried/surfaced)."* The throw classifies as `crash` → 3 retries
  with 2s/4s/8s backoff → error surfaced to the user. And because the turn throws,
  `tryPersistSessionId` is never reached, so the stale id stays on the still-active row and **every
  subsequent turn fails identically** until the user runs `/reset` or `/setproject <name>` — the two
  commands that do deactivate.
- **Codex / OpenCode** — the resume succeeds, so the agent runs in the neutral directory carrying the
  old project's full conversation history. A quieter failure, but a context leak.
- **Pi / Copilot** — degrade gracefully with a "could not resume" notice.

The function's own docblock at `:2569-2580` still advertises *"Four effects: … 4. Deactivates the
active AI session ('project-changed'), so the next message starts fresh"*, which the new early return
falsifies — worth updating either way.

**Fix**: mirror `:2627-2630` in the clear branch. I applied exactly that locally and all 214 tests
still pass, so it's additive:

```ts
if (args.length === 1 && SETPROJECT_CLEAR_KEYWORDS.has(args[0].toLowerCase())) {
  const session = await sessionDb.getActiveSession(conversation.id);
  if (session) {
    await safeDeactivateSession(session.id, 'setproject');
  }
  await db.updateConversation(conversation.id, { ... });
```

Worth a test asserting the deactivation happens (the rebind path has one — *"deactivates active
provider session when project changes"*), since nothing currently pins it in either direction.

On the design question you posed — "detaching is not a session reset" — the distinction is right in
spirit, but it doesn't survive contact with how sessions are stored: they're keyed to a cwd, and detach
changes the cwd. Preserving the session across that boundary isn't preserving context, it's preserving
a dangling handle. `/worktree create` is the one precedent that deliberately keeps the session
(`command-handler.ts:359`), and it moves cwd *within the same repo* while keeping the prompt and
toolset — not comparable.

### I2 — The detached worktree is orphaned silently

`packages/core/src/orchestrator/orchestrator-agent.ts:2599`

The clear path nulls `isolation_env_id` and says nothing. The rebind path handles this exact situation
deliberately (`:2632-2637`, `:2656-2663`): it leaves the worktree in place — correct, it may hold
uncommitted work — and tells the user how to reclaim it. Its comment even explains why it can't point
at `/worktree remove`:

```ts
// Don't suggest `/worktree remove` here: it reads isolation_env_id from
// THIS conversation, which we just cleared — it would short-circuit with
// "not using a worktree".
```

Same problem after detach, and it's slightly worse because chat has no reachable remedy at all:

- `getConversationsUsingEnv` selects on `isolation_env_id` (`db/isolation-environments.ts:211-216`),
  which is now null → the env looks unreferenced.
- `cleanup-service.onConversationClosed` early-returns on `!conversation?.isolation_env_id`
  (`:235-238`), so closing the thread no longer reaps it.
- `/worktree remove` is blocked *before* reaching the `isolation_env_id` check, by the codebase guard
  at `command-handler.ts:271-275`: *"No codebase configured. Register a project first with
  /register-project."* — misleading, since the project is registered and just isn't bound.

So the environment survives until `findStaleEnvironments` (~14 days) or a manual `archon isolation
cleanup`. Not a disaster, but it's a leak the rebind path already decided to surface.

**Fix**: carry the rebind path's note across. `conversation.isolation_env_id` is still readable at that
point, so it's the same `detachedWorktree` check and the same message.

---

## Suggestions

### S1 — "exactly the state a fresh conversation starts in" isn't literally true

Field-by-field against what `createConversation` writes
(`db/conversations.ts:136-139`, plus the schema defaults at `migrations/000_combined.sql:109-124`):
`codebase_id`, `cwd` and `isolation_env_id` all match. But `ai_assistant_type` does not.

`getOrCreateConversation` derives it from the codebase when one is supplied (`conversations.ts:101-110`),
and `POST /api/conversations` does supply it (`packages/server/src/routes/api.ts:2368-2374`). So a
console conversation created under a project whose `ai_assistant_type` is `codex` keeps `codex` after
detaching, whereas a genuinely fresh neutral conversation would resolve `config.assistant`. Add the
active session (I1) and the surviving environment row (I2) and the claim is a bit stronger than what
the code does.

This one is pre-existing and symmetric — the rebind path doesn't touch `ai_assistant_type` either, and
`updateConversation` doesn't even accept it — so I'd leave the behaviour alone. Just soften the claim
in the PR body and the docs entry to something like "clears the project scope" rather than "exactly the
state a fresh one starts in".

### S2 — `-` is documented but untested

The three new tests cover `none` and `CLEAR`. I removed `'-'` from `SETPROJECT_CLEAR_KEYWORDS`
entirely and all 214 tests still passed — so a documented alias can regress silently. Worth one case,
mirroring the `none` test. (CodeRabbit flagged the same thing.)

For contrast, the rest of the tests *are* load-bearing: dropping `cwd: null` from the update payload
fails 2 of them. Good.

### S3 — The inline comment overstates the sentinel mitigation

`orchestrator-agent.ts:2592-2594` says the gate is *"…so a project genuinely named 'none' is never
shadowed"*. Single-token gating is exactly the case that *is* shadowed — `/setproject none` can no
longer select a project named `none`. Your PR body states this accurately ("cannot select it by that
bare name"); the code comment contradicts it.

Confirming the premise: `handleRegisterProject` (`:2392-2433`) does no name validation beyond a
duplicate check, so `none`, `clear` and `-` are all legal project names.

The escape hatch is real but thinner than "any other resolution tier" suggests. The prefix tier
(`/setproject non`) reaches it, but `checkTier` throws on multiple matches (`:480-483`), so with both
`none` and `nonesuch` registered there's no chat-reachable name — the console's project-scoped routes
still work. Suggest rewording to match the PR body, e.g. *"a project named `none` remains reachable via
the prefix/substring tiers, but not by its bare exact name."*

### S4 — Log event naming

`project.cleared` (`:2601`) doesn't follow the `{domain}.{action}_{state}` convention CLAUDE.md
requires. Immediate neighbours are `project.set_completed`, `project.remove_completed`,
`project.register_completed`. Suggest `project.clear_completed`.

### S5 — Thread inheritance treats a null binding as "never bound"

`inheritThreadContext` (`:959`) gates on `conversation.codebase_id` being falsy, which until now meant
"never bound". After detach that proxy is no longer reliable, and a detached conversation can be
re-bound from its parent on the next message.

Narrow in practice, so no change requested — I checked all three platforms: Telegram never passes a
parent; Slack's `getParentConversationId` returns `channel:thread_ts`, the same id as the conversation
itself (`adapters/src/chat/slack/adapter.ts:250-256` vs `:336-345`), so the lookup finds the row we
just cleared and returns; Discord parent channels rarely own a conversation row because `ensureThread`
redirects to the thread. Flagging it only because the invariant is now weaker than the code assumes.

Forge adapters re-link deliberately (`adapters/src/forge/github/adapter.ts:1058`), which is correct —
a GitHub issue conversation genuinely is about that repo.

---

## Strengths

- **Returning before `listCodebases()`** is the detail that makes this robust rather than merely
  convenient: detaching can't be blocked by a broken or empty project list, which is exactly when a
  user is most likely to want out.
- **The #1937 regression guard is carried into the new tests** — asserting the write targets
  `db-conv-1` and not the platform id, with a comment saying why. Easy to skip; you didn't.
- **The multi-word test** (`/setproject none of your business`) covers an edge most contributors
  wouldn't think to write, and it's the test that justifies the single-token gate.
- **The PR body flagged the session question as unverified rather than glossing it.** That is the right
  instinct, and it was correct — I1 is exactly that question.

---

## Summary

| Category | Count |
|---|---|
| Critical | 0 |
| Important | 2 |
| Suggestions | 5 |

I1 and I2 are both small, additive changes localized to the new branch — roughly six lines plus a test.
The direction and the scope boundary are right; nothing structural needs rethinking. Happy to re-review
promptly once those land.

Wirasm (changes requested):
Requesting changes formally so the state is visible — the full findings are in the review comment above.

**I1 (blocking):** the clear path returns *before* the rebind path's `safeDeactivateSession`, so the active AI session survives the detach. On the next turn `getActiveSession` hands the stale `assistant_session_id` to the provider alongside the new neutral cwd. On Claude — the default — an invalid resume id throws, three retries burn, and because the turn throws `tryPersistSessionId` never runs, so **every subsequent turn fails identically until `/reset`**. The conversation wedges. On Codex/OpenCode it resumes *successfully* and carries the old project's context into the neutral directory, which is arguably worse because it looks fine.

You flagged this yourself under "What was not verified" and offered the two-line fix — your instinct was right, and the evidence above confirms it. Note the function's docblock still advertises "Deactivates the active AI session", which the new branch falsifies.

**I2:** the `isolation_environments` row and its worktree are orphaned. `getConversationsUsingEnv` selects on the now-null `isolation_env_id`, so `cleanup-service.onConversationClosed` early-returns and closing the thread no longer reaps it. `/worktree remove` is blocked before the env check by the codebase guard with "No codebase configured. Register a project first" — misleading, since the project *is* registered. The rebind path 20 lines below already handles this case with a user-facing cleanup note.

On the headline claim: the three project-scope columns do match `createConversation` exactly, and all five binding consumers handle null correctly — verified individually. So the core design holds; it is state the conversation still *owns* that leaks.

Strong PR. The early return before `listCodebases()` and carrying the #1937 regression guard into the new tests are both better than average.

133Felix:
Thanks for the review — both findings were right, and I1 in particular was worse in practice than I had assumed when I flagged it as unverified. Pushed a rebase onto current `dev` (0.8.0) with all of it applied.

**I1 — session deactivation (blocking).** The clear branch now calls `safeDeactivateSession(session.id, 'setproject')` **before** `updateConversation`, mirroring the rebind path and for the reason its comment already gives. Your trace was correct end to end: `getActiveSession` filters on `conversation_id + active` only, the stale `assistant_session_id` reaches the provider alongside the neutral cwd, Claude resumes-or-errors, and because the turn throws `tryPersistSessionId` never runs — so it is not one bad turn, it is every subsequent turn. Pinned by a test asserting `('session-123', 'project-changed')`, matching the rebind path's existing test.

I kept the existing `'setproject'` → `'project-changed'` trigger rather than adding a `'project-cleared' `one. `TRIGGER_BEHAVIOR` is an exhaustive `Record<TransitionTrigger, ...>`, so a new value means touching a shared state machine for no behavioural gain. Say the word if you would rather have the distinct trigger for audit clarity.

**I2 — orphaned worktree.** Carried the rebind path's handling across verbatim: the `detachedWorktree` check, a `project.clear_worktree_detached` log, and the same user-facing note. Two tests (note present when an env was attached, absent when not). You were right that it is worse here — `/worktree remove` is unreachable after a detach, since it is gated behind the codebase check that a detached conversation now also fails.

**S1.** Softened everywhere. The docblock and the reply now say the command clears the project **scope**, and the docblock names `ai_assistant_type` as the reason "same as a fresh conversation" is not literally true. Behaviour unchanged, as you suggested.

**S2.** `/setproject -` has its own test now.

**S3.** Reworded to match the PR body: a project named `none` stays reachable through the prefix/substring tiers, but not by its bare exact name. The old comment claimed the opposite of what the code does.

**S4.** `project.cleared` → `project.clear_completed`.

**S5** — noted, not acted on, per your own read that it is narrow across all three platforms. Happy to take it if you disagree.

**Validation:** `bun run validate` exits 0 on the rebased branch (all nine checks). `orchestrator-agent.test.ts` is at 223 pass / 0 fail, 8 of them covering the clear path.

**One thing worth flagging:** this is no longer only theoretical for me. I hit exactly the I1 wedge on my own install while upgrading to 0.8.0 — the Telegram conversation was unusable until `/reset`, which is the one command that does deactivate. Your diagnosis was precise.

Wirasm:
Thanks for working through this and for identifying the persistent-chat problem. We agree that a conversation needs a way to return to neutral without deleting the registered project, but we do not want to merge the current command shape.

The accepted contract is now captured in #2290:

- use an explicit `/detach-project "<current-project-name>"` command;
- require the currently bound project name so the action is deliberate;
- do not reserve `none`, `clear`, or `-`, because those are valid project names;
- refuse detachment while the conversation has pending, running, paused, or failed-but-resumable runs;
- refuse while a worktree or container environment is still attached;
- report the blockers and let `manage_run`, the CLI/Web run controls, and isolation cleanup own their respective lifecycle changes;
- make the blocker check and detach transition one ownership boundary so run creation cannot race the detach;
- support surfaces where neutral state persists, and explicitly refuse forge conversations or inherited child threads that would immediately auto-rebind;
- only after those checks pass, deactivate the active AI session and clear the conversation binding;
- keep `/remove-project <name>` as the separate install-wide operation that unregisters a project.

You are very welcome to implement this revised shape in a new PR linked to #2290. Closing this PR as-is because its sentinel syntax and implicit detach lifecycle no longer match the contract we want.

PR #2294, opened 2026-07-27
fix(core): /reset clears the execution binding and abandons resumable runs

### LINKED ISSUE #2291: bug(core): /reset leaves the worktree binding and resumable runs intact — the next message can resume a stale run
## Summary

- What broke: `/reset` promises "Starting fresh on next message" but only deactivates the AI session. It
  leaves the conversation's worktree binding (`cwd`, `isolation_env_id`) in place and leaves every resumable
  workflow run resumable — so the next message can land in the old worktree, or silently resume a stale run
  instead of starting fresh.
- When it started (if known): since `/reset` was introduced; the resumable-run half became reachable from chat
  once conversation-scoped resume landed.
- Severity: `major`

## Steps to Reproduce

1. In a chat conversation bound to project P, run an interactive workflow that pauses at an approval gate (or
   let any workflow fail). The run stays `paused`/`failed` and remains a resume candidate.
2. The conversation now also carries `cwd` + `isolation_env_id` pointing at that run's worktree.
3. Type `/reset`. Archon replies "Session cleared. Starting fresh on next message. Codebase configuration
   preserved."
4. Send a normal message that dispatches the same workflow.

## Expected vs Actual

- **Expected**: a reset conversation starts genuinely fresh — new session, project root as cwd, no run
  silently continued.
- **Actual**:
  - `conversation.cwd` still points at the old worktree, and the orchestrator prefers it over
    `codebase.default_cwd`, so the "fresh" turn executes in the old isolation environment.
  - The stale run is still a candidate for `findResumableRunByParentConversation` (`status IN ('failed',
    'paused')`) and for `getPausedWorkflowRun` (`status = 'paused'`), so the next dispatch can resume it —
    carrying its old arguments and completed nodes — instead of starting a new run.

## User Flow

```
  User                        Archon                              DB
  ────                        ──────                              ──
  runs workflow ────────────▶ run pauses at gate
                              conversation.cwd = <worktree>
                              conversation.isolation_env_id = <env>
  /reset ───────────────────▶ deactivate session
                              [X] cwd / isolation_env_id untouched
                              [X] paused run still resumable
  "start fresh" ────────────▶ resume lookup finds the stale run
                              executes in the OLD worktree  [X]
  sees the old run ◀────────  continue, not a fresh start
```

## Environment

- Platform: all chat platforms + Web (the command handler is shared)
- Database: PostgreSQL and SQLite
- OS: Linux

## Logs

```
(no error — the stale continuation is silent by construction)
```

## Impact

- Affected workflows/commands: `/reset`, and any subsequent dispatch in the same conversation; worst on
  persistent-conversation platforms (Telegram/Discord) where one row accumulates every run ever made.
- Reproduction rate: Always, when a resumable run or a worktree binding exists at reset time.
- Workaround available? Yes, but manual and non-obvious: `/workflow abandon <id>` for every outstanding run,
  plus `/worktree remove` to clear the binding.
- Data loss risk? `No` for the fix as proposed — see the note below on the deliberate trade-off.

## Scope

- Module (if known): `core:orchestrator` (`case 'reset'` in `packages/core/src/handlers/command-handler.ts`),
  plus a new query in `packages/core/src/db/workflows.ts`

## Proposed Fix

Make `/reset` mean what it says:

1. Deactivate the active session (unchanged).
2. Clear `cwd` and `isolation_env_id`, so the next turn resolves back to `codebase.default_cwd` and a fresh
   isolation environment is created on demand.
3. Abandon the conversation's resumable runs — matching on both `conversation_id` and
   `parent_conversation_id`, because chat-dispatched runs record the chat in the latter and a hidden worker
   conversation in the former.
4. Keep `codebase_id` — resetting a conversation is not the same as leaving the project. (Detaching from the
   project is a separate concern; see the `/setproject none` request.)
5. Report what happened, including the number of abandoned runs, so the effect is never silent.

**Deliberate trade-off worth calling out:** step 3 makes `/reset` destructive toward in-flight runs. A user who
types `/reset` while a run is legitimately paused at a gate loses that run (it becomes `cancelled`). That is
the point of an escape hatch — but it is a real behaviour change, which is why the reply names the count. If
maintainers prefer, the run-abandoning half could sit behind an explicit variant instead; the binding-clearing
half stands on its own either way.

133Felix:
## Summary

- Problem: `/reset` replies "Starting fresh on next message" but only deactivates the AI session. The
  conversation keeps `cwd` + `isolation_env_id`, and every resumable run stays resumable.
- Why it matters: the orchestrator prefers `conversation.cwd` over `codebase.default_cwd`, so the "fresh" turn
  still executes inside the previous worktree; and because `findResumableRunByParentConversation` matches
  `status IN ('failed','paused')` (and `getPausedWorkflowRun` matches `paused`), the next dispatch can
  silently continue a stale run — old arguments, old completed nodes — instead of starting a new one. `/reset`
  is the natural escape hatch for exactly that situation and currently isn't one.
- What changed: `/reset` now also clears `cwd`/`isolation_env_id` and abandons the conversation's resumable
  runs, then reports what it did (including the number of runs abandoned). Adds
  `abandonResumableRunsForConversation` to the workflow-run DB module.
- What did **not** change (scope boundary): `codebase_id` is still preserved — resetting a conversation is not
  the same as leaving the project. Session deactivation is unchanged. No schema change, no new dependency, no
  other command touched.

### Before

```
  User                        Archon                              DB
  ────                        ──────                              ──
  runs workflow ────────────▶ run pauses at a gate
                              conversation.cwd = <worktree>
                              conversation.isolation_env_id = <env>
  /reset ───────────────────▶ deactivate session
                              [X] cwd / isolation_env_id untouched
                              [X] paused run still resumable
  "start fresh" ────────────▶ resume lookup finds the stale run
                              executes in the OLD worktree  [X]
  sees the old run ◀────────  continue, not a fresh start
```

### After

```
  User                        Archon                              DB
  ────                        ──────                              ──
  runs workflow ────────────▶ run pauses at a gate
  /reset ───────────────────▶ deactivate session
                              *abandonResumableRunsForConversation*
                              *updateConversation(cwd: null,
                                 isolation_env_id: null)*
                              codebase_id preserved
  ◀──────────────────────────  "Session cleared. Cleared workspace binding
                                (worktree + isolation env). Abandoned 1 pending
                                run(s). Project attachment preserved — next
                                message starts fresh."
  "start fresh" ────────────▶ resume lookup finds nothing
                              new run, cwd = codebase.default_cwd,
                              fresh isolation env  [+]
```

### Before

```
  command-handler.ts
    case 'reset'
        │
        ├── sessionDb.getActiveSession
        └── safeDeactivateSession ─────────▶ sessions
             (nothing else)

  next dispatch
    orchestrator-agent.ts ────────────────▶ db/workflows.ts
                                              findResumableRunByParentConversation
                                              (stale run still matches)
```

### After

```
  command-handler.ts
    [~] case 'reset'
        │
        ├── sessionDb.getActiveSession
        ├── safeDeactivateSession ─────────▶ sessions            (unchanged)
        ├── === workflowDb.abandonResumableRunsForConversation ─▶ [+] workflow_runs
        └── === db.updateConversation(cwd: null,
                 isolation_env_id: null) ──▶ conversations

  next dispatch
    orchestrator-agent.ts ────────────────▶ db/workflows.ts      (unchanged code,
                                              findResumableRun…    now finds nothing)
```

**Connection inventory**

| `command-handler.ts` `case 'reset'` | `sessionDb` / `safeDeactivateSession` | unchanged | same call, same transition reason |
| `command-handler.ts` `case 'reset'` | `db/workflows.ts` `abandonResumableRunsForConversation` | **new** | new function in an existing module |
| `command-handler.ts` `case 'reset'` | `db/conversations.ts` `updateConversation` | **new** | clears `cwd` + `isolation_env_id` only |
| `abandonResumableRunsForConversation` | `remote_agent_workflow_runs` | **new** | conditional UPDATE, no schema change |
| `orchestrator-agent.ts` | `findResumableRunByParentConversation` | unchanged | untouched; it simply has no stale candidate after a reset |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: S`
- Scope: `core`
- Module: `core:orchestrator`

## Change Metadata

- Change type: `bug`
- Primary scope: `core`

## Linked Issue

- Closes #2291
- Related #1549 — the same class of problem from the CLI side (auto-resume of a failed run hijacking a fresh
  request); this gives chat users the escape hatch.

```bash
bun run validate
```

Targeted:

```bash
bun test packages/core/src/db/workflows.test.ts        #  85 pass, 0 fail
bun test packages/core/src/handlers/command-handler.test.ts   # 123 pass, 0 fail
```

- Evidence provided: full `bun run validate` on this branch (Linux, Bun 1.3.11) plus both targeted file runs.

## Security Impact (required)

- File system access scope changed? `No` — clearing `cwd`/`isolation_env_id` returns the conversation to the
  project root it was already entitled to; the worktree itself is left on disk untouched

## Compatibility / Migration

- Backward compatible? `Yes` for data and API shape. The **behaviour** of `/reset` changes deliberately — see
  Blast Radius.
- Config/env changes? `No`
- Database migration needed? `No` — a new conditional UPDATE against an existing table

- Verified scenarios: run parked at an approval gate → `/reset` → reply names the abandoned count → the next
  message starts a brand-new run in `codebase.default_cwd` with a fresh isolation env rather than resuming.
  Reset with no session and no runs still clears the binding and says so.
- Edge cases checked: nothing to abandon → count sentence omitted; no active session → "No active session."
  and the binding is still cleared; conversation row vanished mid-reset → `ConversationNotFoundError` is
  absorbed (a reset cannot meaningfully fail on a conversation that no longer exists) while any other DB error
  propagates.
- What was not verified: behaviour under a concurrent dispatch racing the reset. The abandon is a single
  conditional UPDATE, so a run that starts after it simply survives — which is the intuitive outcome — but I
  have not built a harness for that race.

- Affected subsystems/workflows: `/reset` on every platform (the command handler is shared), plus any
  subsequent dispatch in that conversation.
- Potential unintended effects — **the one to weigh**: `/reset` becomes destructive toward in-flight runs. A
  user who types `/reset` while a run is legitimately paused at an approval gate loses that run (it becomes
  `cancelled`). That is the intended meaning of "start fresh", but it is a real behaviour change from today's
  no-op. Mitigations: the reply always names the number of runs abandoned, so it is never silent, and the
  command reference now states it explicitly and points at `/workflow approve` / `/workflow resume <id>` as
  the way to keep a run. If maintainers would rather not couple the two, the run-abandoning half can move
  behind an explicit variant (e.g. `/reset --hard`) and the binding-clearing half still stands on its own —
  happy to split it that way.
- Guardrails/monitoring for early detection: abandoned runs land in `cancelled` with `completed_at` set, so
  they remain visible in `/workflow runs` and the console rather than disappearing.

## Rollback Plan (required)

- Fast rollback command/path: revert this commit. Runs already cancelled by a reset stay cancelled — they are
  terminal and visible, and can be inspected via `/workflow get <id>`; no data is deleted.
- Observable failure symptoms: `/reset` reporting abandoned runs that the user wanted to keep, or a post-reset
  message still landing in the old worktree.

## Risks and Mitigations

- Risk: a user loses a paused run they wanted to keep.
  - Mitigation: the reply states the count; the docs say what `/reset` discards and how to keep a run; the
    cancelled run remains inspectable. Willing to gate this half behind an explicit flag if preferred.
- Risk: the abandon query misses runs recorded only under a hidden worker conversation.
  - Mitigation: it matches `conversation_id OR parent_conversation_id`, and the test asserts both columns are
    covered along with all four non-terminal statuses.
- Risk: absorbing `ConversationNotFoundError` hides a real failure.
  - Mitigation: only that one error type is absorbed; everything else rethrows. Note this diverges from the
    two other call sites in the file, which return `{ success: false }` instead — converting to that style is
    a one-line change if reviewers prefer consistency.

## Summary by CodeRabbit

* **New Features**
  * Enhanced `/reset` to clear the active session and execution binding while preserving the project attachment.
  * Paused and failed resumable workflow runs are now abandoned during reset; active pending or running runs remain untouched.
  * Reset results report abandoned runs and workspace cleanup, including partial outcomes when an operation fails.

* **Documentation**
  * Updated `/reset` guidance and examples to reflect session and workspace clearing, run abandonment, and preserved project attachments.

### DIFF
--- a/packages/core/src/db/workflows.test.ts
+++ b/packages/core/src/db/workflows.test.ts
@@ -31,6 +31,7 @@ import {
   updateWorkflowActivity,
   findResumableRun,
   findResumableRunByParentConversation,
+  findResumableRunIdsForConversation,
   resumeWorkflowRun,
   pauseWorkflowRun,
   cancelWorkflowRun,
@@ -723,6 +724,50 @@ describe('workflows database', () => {
 
       await expect(findResumableRunByParentConversation('piv', 'conv-1', 'cb')).rejects.toThrow(
         'Failed to find resumable run by parent conversation: Connection refused'
+      );
+    });
+  });
+
+  describe('findResumableRunIdsForConversation', () => {
+    test('matches the conversation and its worker conversations', async () => {
+      mockQuery.mockResolvedValueOnce(createQueryResult([{ id: 'run-1' }, { id: 'run-2' }]));
+
+      const result = await findResumableRunIdsForConversation('conv-1');
+
+      expect(result).toEqual(['run-1', 'run-2']);
+      const [query, params] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).toContain('conversation_id = $1');
+      expect(query).toContain('parent_conversation_id = $2');
+      expect(query).toContain('ORDER BY started_at DESC');
+      expect(query).not.toMatch(/--.*\$\d/); // #999 guard: $N in SQL comments breaks convertPlaceholders
+      expect(params).toEqual(['conv-1', 'conv-1']);
+    });
+
+    test('never matches live pending or running work', async () => {
+      // The load-bearing assertion. `pending`/`running` may belong to another
+      // process entirely (a CLI run, a webhook dispatch); neither resume lookup
+      // can return them, so cancelling them buys nothing and destroys live work.
+      mockQuery.mockResolvedValueOnce(createQueryResult([]));
+
+      await findResumableRunIdsForConversation('conv-1');
+
+      const [query] = mockQuery.mock.calls[0] as [string, unknown[]];
+      expect(query).toContain("status IN ('paused', 'failed')");
+      expect(query).not.toContain("'running'");
+      expect(query).not.toContain("'pending'");
+    });
+
+    test('returns an empty array when nothing is resumable', async () => {
+      mockQuery.mockResolvedValueOnce(createQueryResult([]));
+
+      expect(await findResumableRunIdsForConversation('conv-1')).toEqual([]);
+    });
+
+    test('throws on database error', async () => {
+      mockQuery.mockRejectedValueOnce(new Error('Connection refused'));
+
+      await expect(findResumableRunIdsForConversation('conv-1')).rejects.toThrow(
+        'Failed to find resumable runs for conversation: Connection refused'
       );
     });
   });
--- a/packages/core/src/db/workflows.ts
+++ b/packages/core/src/db/workflows.ts
@@ -449,6 +449,45 @@ export async function getPausedWorkflowRun(conversationId: string): Promise<Work
 }
 
 /**
+ * Ids of every RESUMABLE run belonging to a conversation, newest first.
+ *
+ * Used by `/reset` to give the user a real escape hatch: after these are
+ * abandoned, the resume lookups find nothing, so the next dispatch starts fresh
+ * instead of continuing a stale run.
+ *
+ * The status set is exactly what the two resume lookups can return —
+ * findResumableRunByParentConversation ('failed'/'paused') and
+ * getPausedWorkflowRun ('paused'). `pending` and `running` are deliberately NOT
+ * matched: neither lookup can ever return them, so leaving them alone cannot
+ * cause the stale continuation this exists to prevent, while cancelling them
+ * would stop live work that may belong to another process entirely (a CLI run,
+ * a webhook-triggered run, a scheduled dispatch).
+ *
+ * `workflow:` sub-runs inherit the parent's `conversation_id` AND
+ * `parent_conversation_id` unchanged at every nesting level
+ * (packages/workflows/src/executor.ts, child run creation), so this predicate
+ * already covers the whole sub-run tree without a `parent_run_id` walk.
+ */
+export async function findResumableRunIdsForConversation(
+  conversationId: string
+): Promise<string[]> {
+  try {
+    const result = await pool.query<{ id: string }>(
+      `SELECT id FROM remote_agent_workflow_runs
+       WHERE (conversation_id = $1 OR parent_conversation_id = $2)
+         AND status IN ('paused', 'failed')
+       ORDER BY started_at DESC`,
+      [conversationId, conversationId]
+    );
+    return result.rows.map(r => r.id);
+  } catch (error) {
+    const err = error as Error;
+    getLog().error({ err, conversationId }, 'db.workflow_run_find_resumable_for_conv_failed');
+    throw new Error(`Failed to find resumable runs for conversation: ${err.message}`);
+  }
+}
+
+/**
  * Find the workflow run currently holding the lock on `workingPath`.
  *
  * The lock is held by any row in `(running, paused)` or `pending` younger
--- a/packages/core/src/handlers/command-handler.test.ts
+++ b/packages/core/src/handlers/command-handler.test.ts
@@ -45,6 +45,9 @@ const mockUpdateWorkflowRun = mock(() => Promise.resolve());
 // findChildRuns → pool.query → created and schema-initialised a real SQLite
 // database on disk, in a test that reads as fully mocked (#2240).
 const mockFindChildRuns = mock(() => Promise.resolve([]));
+const mockFindResumableRunIdsForConversation = mock(
+  (): Promise<string[]> => Promise.resolve([] as string[])
+);
 // CAS gate resolvers (#2113) — approve/reject stamp the resolution atomically here
 // instead of via updateWorkflowRun. resolveAndCancelApprovalGate is the atomic
 // resolve+cancel for terminal reject outcomes. Default to "won the race".
@@ -102,6 +105,7 @@ mock.module('../db/workflows', () => ({
   listWorkflowRuns: mockListWorkflowRuns,
   getWorkflowRun: mockGetWorkflowRun,
   findChildRuns: mockFindChildRuns,
+  findResumableRunIdsForConversation: mockFindResumableRunIdsForConversation,
   resumeWorkflowRun: mockResumeWorkflowRun,
   failWorkflowRun: mockFailWorkflowRun,
   updateWorkflowRun: mockUpdateWorkflowRun,
@@ -741,6 +745,72 @@ describe('CommandHandler', () => {
     });
 
     describe('/reset', () => {
+      beforeEach(() => {
+        mockFindResumableRunIdsForConversation.mockClear();
+        mockFindResumableRunIdsForConversation.mockImplementation(() => Promise.resolve([]));
+        mockUpdateConversation.mockClear();
+        mockUpdateConversation.mockImplementation(() => Promise.resolve());
+        mockGetWorkflowRun.mockClear();
+        mockCancelWorkflowRun.mockClear();
+        mockCancelWorkflowRun.mockImplementation(() => Promise.resolve({ cancelled: true }));
+      });
+
+      test('clears the execution binding but preserves the project attachment', async () => {
+        mockGetActiveSession.mockResolvedValue(null);
+
+        const result = await handleCommand(baseConversation, '/reset');
+
+        expect(result.success).toBe(true);
+        // cwd + isolation env go; codebase_id is deliberately absent from the
+        // payload — detaching the project is /setproject none's job.
+        expect(mockUpdateConversation).toHaveBeenCalledWith(baseConversation.id, {
+          cwd: null,
+          isolation_env_id: null,
+        });
+        const [, payload] = mockUpdateConversation.mock.calls[0] as [string, object];
+        expect(payload).not.toHaveProperty('codebase_id');
+        expect(result.message).toContain('Project attachment preserved');
+      });
+
+      test('abandons resumable runs and names the count', async () => {
+        mockGetActiveSession.mockResolvedValue(null);
+        mockFindResumableRunIdsForConversation.mockImplementation(() =>
+          Promise.resolve(['run-a', 'run-b'])
+        );
+        mockGetWorkflowRun.mockImplementation((id: unknown) =>
+          Promise.resolve({ id, status: 'paused', metadata: {} })
+        );
+
+        const result = await handleCommand(baseConversation, '/reset');
+
+        expect(mockFindResumableRunIdsForConversation).toHaveBeenCalledWith(baseConversation.id);
+        expect(mockCancelWorkflowRun.mock.calls.map(c => c[0])).toEqual(['run-a', 'run-b']);
+        // "resumable", not "pending": pending is itself a status name and reads
+        // as "waiting" to a user.
+        expect(result.message).toContain('Abandoned 2 resumable run(s).');
+      });
+
+      test('still reports the abandoned count when clearing the binding fails', async () => {
+        // The two effects live in separate try blocks precisely so a failure in
+        // the second cannot swallow what the first already did.
+        mockGetActiveSession.mockResolvedValue(null);
+        mockFindResumableRunIdsForConversation.mockImplementation(() => Promise.resolve(['run-a']));
+        mockGetWorkflowRun.mockImplementation((id: unknown) =>
+          Promise.resolve({ id, status: 'paused', metadata: {} })
+        );
+        mockUpdateConversation.mockImplementation(() =>
+          Promise.reject(new Error('Conversation not found: conv-123'))
+        );
+
+        const result = await handleCommand(baseConversation, '/reset');
+
+        expect(result.success).toBe(false);
+        expect(result.message).toContain('Abandoned 1 resumable run(s).');
+        expect(result.message).toContain('Could not clear the workspace binding');
+        // And it must NOT claim the binding was cleared.
+        expect(result.message).not.toContain('Cleared workspace binding');
+      });
+
       test('should deactivate active session', async () => {
         mockGetActiveSession.mockResolvedValue({
           id: 'session-123',
--- a/packages/core/src/handlers/command-handler.ts
+++ b/packages/core/src/handlers/command-handler.ts
@@ -32,6 +32,7 @@ import {
   getWorkflowStatus,
   resumeWorkflow,
   abandonWorkflow,
+  abandonResumableRunsForConversation,
   resetWorkflowNodeSessions,
 } from '../operations/workflow-operations';
 import { safeDeactivateSession } from '../state/session-transitions';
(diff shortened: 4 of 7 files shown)


Wirasm:
## Review: #2294

Reviewed **together with #2292**, in the order **#2292 → #2294**. They are coupled — this PR's problem statement quotes the candidate set of the function #2292 re-orders, and this is the escape hatch for the situation that ordering makes more reachable. Landing order and the interaction are at the bottom.

**Verdict: request changes** — one Important issue (I1) that is a one-line fix, plus I2/I3. The diagnosis is right, `/reset` genuinely was a half-truth, and the shape of the fix is the correct one. The problem is that the query is broader than everything that describes it.

---

### First, the thing you flagged: does this violate *No Autonomous Lifecycle Mutation Across Process Boundaries*?

**No. I land on your side.** Two independent reasons:

1. The rule's operative clause is "must not **autonomously** mark that work as failed/cancelled/abandoned **based on a timer or staleness guess**", and its prescribed remedy is "surface the ambiguous state to the user and provide a one-click action." `/reset` is an explicit user action that reports its outcome. It is the remedy, not the thing forbidden.
2. Cancelling a non-terminal run is already first-class here, not a novel liberty. `cancelWorkflowRun` (`packages/core/src/db/workflows.ts:897-913`) says so outright — *"a 'running' run stays cancellable — that is cooperative cancellation, which the executor honors via its between-layer status check (dag-executor)"* — and `abandonWorkflow` cancels running runs from every surface today. It isn't silent either: `packages/workflows/src/dag-executor.ts:6186` posts `⚠️ Workflow stopped (cancelled)` into the conversation when the executor notices.

So the rule is not the blocker. **But `running` and `pending` still shouldn't be in this query**, for scope and disclosure reasons rather than rule reasons.

---

### I1 (Important) — the query is wider than its own name, its doc comment, the PR body, and the shipped docs

`abandonResumableRunsForConversation` matches `status IN ('pending', 'running', 'paused', 'failed')`. Its doc comment two lines above says:

> Includes 'failed' on purpose — the candidate set this must neutralise is exactly the set the resume lookup can return.

That sentence is false for `pending` and `running`. The two lookups you cite return `failed`/`paused` (`findResumableRunByParentConversation`) and `paused` (`getPausedWorkflowRun`). Neither can ever return a `pending` or `running` row. The code and the rationale directly under it disagree.

**And the wider set has teeth.** I ran your function against a real `bun:sqlite` adapter (throwaway file, not proposed for the branch) with one row per status plus a sub-run plus a second conversation:

```
abandonResumableRunsForConversation('conv-1') → 5
  r-running   running   → cancelled   (completed_at set)
  r-pending   pending   → cancelled
  r-paused    paused    → cancelled
  r-failed    failed    → cancelled
  r-child     running   → cancelled   (sub-run of r-running)
  r-done      completed → completed   (spared)
  other-running (conv-2)→ running     (spared)
```

Chain that with the dag-executor's between-layer check and `/reset` — a command whose new docs line says it "abandons any resumable runs of this conversation" — **stops a live, actively-executing workflow.** That is a materially bigger behaviour change than the PR discloses. Your Blast Radius section names only the paused-gate case ("A user who types `/reset` while a run is legitimately paused at an approval gate loses that run"); killing a run mid-flight isn't mentioned in the PR, the docs diff, or the reply string. And the reply string makes it worse rather than better: `Abandoned 2 pending run(s)` — `pending` is itself a status name, and to a user it reads "waiting", not "I just stopped your running workflow".

**There is also no functional need for it.** I checked:

- Neither resume lookup can return `pending`/`running`, so leaving them alone cannot cause the stale-continuation this PR exists to prevent.
- `getActiveWorkflowRun`'s only two callers are `/workflow cancel` (`command-handler.ts:684`) and the `/status` display (`:1114`). Neither is a dispatch guard, so a surviving `running` row does not block the fresh start.
- The `pending`/`running` path lock in `getActiveWorkflowRunByPath` isn't in the way either, because you clear `cwd` — the next dispatch resolves a new worktree.

**Fix — one line:**

```diff
-         AND status IN ('pending', 'running', 'paused', 'failed')
+         AND status IN ('paused', 'failed')
```

That is now literally "the set the resume lookup can return", it matches the function name, and it makes the doc comment true. The unit test's `toContain` assertion needs updating in the same commit (it currently pins the four-status list — see the mutation note below). Consider also making the reply concrete: `Abandoned 2 paused/failed run(s).`

If you'd rather `/reset` *did* stop a live run — a defensible position, it's arguably what "start fresh" means — then that's fine, but it needs to be stated in the PR body, in `commands.md`, and in the reply string, and it should go through `abandonWorkflow` (see I2) so it behaves like every other abandon.

---

### I2 (Important) — bypasses the shared abandon op, so container resources aren't reclaimed

`abandonWorkflow` (`packages/core/src/operations/workflow-operations.ts:223-277`) exists specifically so that abandon behaves identically everywhere. Its own comment:

> reclaim a container run's container + upper volume immediately, in the SHARED op so EVERY abandon surface (CLI, web API, chat, manage_run, Slack-cancel) frees the resources now rather than waiting for the scheduled reaper.

This PR adds a sixth abandon surface that skips it. A folder-project container run abandoned by `/reset` leaves its container and upper volume alive until the reaper. It's a delayed reclaim rather than a permanent leak, but it breaks an invariant the codebase deliberately centralised one release ago.

The matched set is normally 0-2 rows, so the simplest fix is to select the ids and loop `abandonWorkflow(id)` — you get the container reclaim, the CAS guard, and the audit semantics for free, and the count falls out of the loop. If you'd rather keep the single bulk `UPDATE` for atomicity, say so in the doc comment and add the reclaim explicitly, so the divergence is a choice rather than an oversight.

---

### I3 (Important, small) — the count isn't reported when the second half fails

```ts
abandoned = await workflowDb.abandonResumableRunsForConversation(conversation.id);
await db.updateConversation(conversation.id, { cwd: null, isolation_env_id: null });
```

Both calls sit inside one `try`. If the abandon succeeds and `updateConversation` throws anything other than `ConversationNotFoundError`, the error propagates, the reply is never built, and the user is told the command failed — with N runs already cancelled and no indication of how many or which. That defeats the "the reply always names the number of runs abandoned, so it is never silent" mitigation in precisely the case where it matters most.

Suggest splitting the two so a binding-clear failure still reports the abandoned count alongside the failure. Related, and you already offered it yourself: the `ConversationNotFoundError`-absorb diverges from the `{ success: false }` style at `command-handler.ts:372` and `:474` — I'd take your own offer and match the file. Same class, cosmetic end: if `updateConversation` *does* throw `ConversationNotFoundError`, the reply still asserts "Cleared workspace binding (worktree + isolation env)" even though nothing was cleared.

---

### On cascade — good news, and worth a sentence in the comment

You don't need an explicit `parent_run_id` cascade, and the reason is non-obvious enough to be worth recording. A `workflow:` sub-run inherits the parent's `conversation_id` **and** its `parent_conversation_id` (`packages/workflows/src/executor.ts:609-621`), so `(conversation_id = $1 OR parent_conversation_id = $1)` already covers the whole subtree. Confirmed in the SQLite run above — `r-child` was cancelled and counted.

That means this correctly does *not* need `cascadeCancelChildren`, but a future reader comparing against `abandonWorkflow` will absolutely wonder why. One sentence — "sub-runs inherit the parent's conversation columns, so the tree is covered without a `parent_run_id` walk" — would save them the trip.

(`blockedParentRunId` is likewise moot here: the whole tree goes terminal together, so there's no stranded parent to point at.)

---

### On preserving `codebase_id` — coherent, and provably so

It cannot leave the conversation in an unreachable state: `{codebase_id: <kept>, cwd: null, isolation_env_id: null}` is byte-for-byte what `/setproject` already writes (`packages/core/src/orchestrator/orchestrator-agent.ts:2615-2619`). So post-reset is an established, well-trodden state, not a novel one. Worth a one-line note in the comment — it turns a judgement call into a citation.

---

### Suggestions

**S1** — the reply says "Cleared workspace binding (worktree + isolation env)." unconditionally, including for a conversation that never had one. Both fields are already on `conversation`, so gating that clause costs no query.

**S2** — the docs line "abandons any resumable runs of this conversation" should be kept in sync with whatever status set I1 settles on. If `running` stays, it must say so explicitly.

**Tests are load-bearing** — I mutated the status list to `('paused','failed')` and `cancels every resumable run on both conversation columns` fails (84 pass / 1 fail). That's good coverage, but note it currently pins the behaviour I'm asking you to change, so the assertion moves with the fix.

---

### Interaction with #2292, and landing order

**Land #2292 first, then #2294.** Both are safe alone against `dev`; neither needs the other to be correct. The order is about coherence: #2292 makes an open gate win *unconditionally* over a newer failed run, and the paused branch auto-resumes with no prompt (the failed branch prompts, per #1549). On a long-lived chat id — a Telegram `chat_id`, a Slack thread — a forgotten gate can then shadow every later attempt at that workflow indefinitely. This PR is the in-chat way out. Shipping the preference first and the escape hatch straight after keeps the pair honest.

**They do not fight.** Your abandon is *set*-based, not order-based — it cancels every matching row, so #2292's `ORDER BY` is irrelevant to it. Walking a conversation with one paused run at T1 and one newer failed run at T2 through both changes:

- before `/reset`: #2292 returns the paused T1 and resumes it (previously: a prompt about the unrelated T2). Fixed.
- after `/reset`: both rows `cancelled`, lookup returns `null`, `cwd` `null` → a genuinely fresh run in `codebase.default_cwd`. Correct.

So this PR abandons the right runs under #2292's ordering — trivially, because it abandons all of them.

**A merge conflict is guaranteed for whichever lands second.** Both PRs add an import to the same list in `packages/core/src/db/workflows.test.ts` and both insert a new `describe` block at the same anchor (immediately before `describe('getActiveWorkflowRunByPath')`). Purely additive — keep both imports and both blocks. I resolved it that way locally: combined, `bun test src/db/workflows.test.ts src/handlers/command-handler.test.ts` → **210 pass / 0 fail**, `bun run type-check` clean across all packages.

`#2299` (approved, unmerged) also touches `workflows.ts` and `workflows.test.ts`, but its hunks sit in `updateWorkflowRun` (~:790-830) and it merges cleanly with both of these in either order — I tested it. Not a risk here.

---

### Strengths

The core insight is correct and had real user cost: `/reset` promising "start fresh" while leaving both the execution binding *and* every resumable run intact is a genuine half-truth, and nobody had noticed. The scope boundary is drawn in the right place — preserving `codebase_id` is the right call and you justified it. The Blast Radius section is unusually honest about the destructive half, and pre-offering the `/reset --hard` split and the `{ success: false }` consistency change is exactly the kind of reviewability that makes a PR fast to assess. The reported count is the right instinct; I1 and I3 are about making it tell the truth in every path.

Once I1 is settled (and I2/I3 addressed or consciously deferred), this is ready.

Wirasm (changes requested):
Requesting changes formally so the state is visible — the full findings are in the review comment above.

**I1 (blocking, one line):** `abandonResumableRunsForConversation` matches `status IN ('pending', 'running', 'paused', 'failed')`, but its own doc comment two lines above says it targets what the resume lookups return — and those return only `failed`/`paused`. So the comment is false for `pending` and `running`, and `/reset` stops **live, actively-executing work** (plus its sub-run children) without saying so.

It is also unnecessary for the bug you are fixing: neither resume lookup can return `pending`/`running`, so leaving them alone cannot cause the stale-continuation this PR exists to prevent. Narrowing the query to `('paused', 'failed')` keeps the fix and removes the overreach.

To be clear about what is *not* the objection: this does **not** violate `No Autonomous Lifecycle Mutation Across Process Boundaries`. That rule forbids mutation "based on a timer or staleness guess", and `/reset` is an explicit user action that reports what it did. Cancelling a non-terminal run is already first-class here. The problem is scope and disclosure, not the rule.

Plus I2/I3 in the comment above. The diagnosis is right and `/reset` genuinely was a half-truth — this is a good catch, and the fix is small.

Wirasm:
Thanks for this — the `/reset` gap is real and worth closing. I verified the automated review's findings against `dev` before relaying them, and **the Major one holds**. It is also a bit more serious than the review states, so I want to give you the full reasoning rather than just the diff.

## The blocking one: `pending` and `running` must not be cancelled

Your predicate is:

```sql
AND status IN ('pending', 'running', 'paused', 'failed')
```

Checked against what `dev` actually treats as resumable — `packages/core/src/db/workflows.ts:698`:

```sql
AND status IN ('failed', 'paused')
```

and `:73`, where a `running` row is only recoverable when it is *provably* a stale orphan:

```sql
(status IN ('failed', 'paused') OR (status = 'running' AND (staleOrphan)))
```

So `pending` and `running` are not resumable rows — they are, or may be, **live work**. A `/reset` in chat would cancel a workflow that is actively executing right now, quite possibly started from a different surface (CLI, an adapter, a webhook, a scheduled run).

That collides with an explicit project rule in CLAUDE.md, *No Autonomous Lifecycle Mutation Across Process Boundaries*:

> When a process cannot reliably distinguish "actively running elsewhere" from "orphaned by a crash" — typically because the work was started by a different process or input source (CLI, adapter, webhook, web UI, cron) — it must not autonomously mark that work as failed/cancelled/abandoned based on a timer or staleness guess.

A chat command cannot know whether a `running` row belongs to a live CLI process on someone's laptop. The suggested fix is right:

```diff
-  AND status IN ('pending', 'running', 'paused', 'failed')
+  AND status IN ('paused', 'failed')
```

The test at `command-handler.test.ts` asserting the four-status string needs the same update.

## The Minor one is also worth taking

"Abandoned N pending run(s)" describes neither the count nor the behaviour once the predicate is narrowed. `Abandoned N resumable run(s)` matches what the code does, and the docs example in `reference/commands.md` should follow.

## What I would like to keep

Everything else here is good, and the `conversation_id = $1 OR parent_conversation_id = $1` handling of chat-dispatched runs is exactly right — that hidden-worker-conversation case is easy to miss.

Once the predicate is narrowed and the two tests updated, this is ready from my side. Happy to take it as-is if you would rather I applied the change myself — just say so and I will push to your branch rather than leaving it on you.

133Felix:
Thanks — I1 was right and the reproduction with a real `bun:sqlite` adapter made it undeniable. Rebased onto current `dev` (0.8.0) with all three findings applied. The shape of the change moved as a result, so worth reading the diff fresh rather than as an increment.

**I1 — narrowed to `('paused', 'failed')` (blocking).** You were right on both counts: the doc comment was false for `pending`/`running`, and the wider set was never needed, since neither resume lookup can return them. It is now literally "the set the resume lookup can return", and the comment is true. A dedicated test asserts the query contains neither `'running'` nor `'pending'`, so the overreach cannot come back unnoticed.

I did **not** take the alternative you offered (keep the wider behaviour and disclose it). Stopping live work from a chat command is the wrong default even when disclosed — the run may belong to a CLI process on another machine, and the person typing `/reset` has no way to know.

**I2 — routes through `abandonWorkflow` now.** This is the structural change. The bulk `UPDATE` is gone; instead:

- `db/workflows.ts` gets `findResumableRunIdsForConversation` — a pure **read** returning ids.
- `operations/workflow-operations.ts` gets `abandonResumableRunsForConversation` — loops those ids through the existing `abandonWorkflow`.

So the CAS guard, `cascadeCancelChildren`, and the container + upper-volume reclaim all come for free instead of being reimplemented or skipped. Making the DB function a read rather than a write is what made the layering honest — it also resolved I1 in the same edit, since the narrowed status set now lives in a query whose only job is to answer "what is resumable here".

Per-run `try`/`catch`: a row can go terminal between the SELECT and the abandon, which makes `abandonWorkflow` throw. That is counted as a failure rather than aborting the whole reset. Three tests cover it (all succeed / one races / lookup fails).

**I3 — split try blocks.** The abandon and the binding-clear no longer share a `try`. A test asserts the count still appears when `updateConversation` throws, and that the reply does **not** then claim "Cleared workspace binding" — you were right that the old code asserted it either way.

I also dropped the `ConversationNotFoundError`-absorb in favour of the `{ success: false }` style at `command-handler.ts:372`/`:474`, as you suggested I take my own offer.

**Reply string** is `Abandoned N resumable run(s).` — agreed that `pending` reads as "waiting" and is itself a status name. `reference/commands.md` now spells out both halves: what is abandoned, and that `pending`/`running` are explicitly left alone.

**Cascade comment** added, as you asked: sub-runs inherit the parent's conversation columns unchanged at every nesting level, so the predicate covers the whole tree without a `parent_run_id` walk. Saves the next reader the trip.

**Interaction with #2292:** that has since merged, so the paused-first ordering this PR's problem statement referenced is now in `dev`, and this branch is rebased on top of it.

**Validation:** `bun run validate` exits 0 (all nine checks). Relevant suites: `command-handler.test.ts` 129 pass, `db/workflows.test.ts` 96 pass, `workflow-operations.test.ts` 45 pass, all 0 fail.

**Human verification:** running this on my own install since today's 0.8.0 upgrade. Also verified the thing your review was actually protecting: a detached CLI run started with `archon workflow run --detach` survives a `/reset` issued from chat, where the previous predicate would have killed it.

Wirasm:
Thanks for the contribution. This outcome was taken forward and merged through #2731, including the later corrections found during review, so this PR is now superseded.

PR #2297, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published image

kobihikri:
Hi, and thanks for Archon.

`.github/workflows/publish.yml` publishes the image, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it cannot check that it was built by this workflow, from this repository, at that tag.

Archon is deployed as a knowledge and task backend for coding agents, so the container typically has a database behind it and API credentials in its environment. That is the sort of placement where a deployer would reasonably want to confirm the image's origin before granting it.

The change is two lines on the build step:

```yaml
          push: true
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the image. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats: `mode=max` records build arguments (`provenance: true` gives a smaller record if any have ever been sensitive), and attestations add an extra manifest to the index, which the registry supports.

No SLSA level claimed — the attestation is what BuildKit produces.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.

## Summary by CodeRabbit

* **Chores**
  * Enhanced Docker image publishing by requesting additional BuildKit supply-chain outputs, including improved build provenance and automatic SBOM generation.
  * Existing image build, tagging, publishing, and caching behavior remains unchanged.


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

Wirasm:
Thanks for this — it's a well-argued change, and the reasoning about who ends up deploying this image is the right frame. I dug into whether the attestations actually get produced rather than just whether the YAML is valid, and pulled the live published manifest to check. Summary: **the change is correct and I'd merge it**, with one correction to the description.

## Verified: no permissions change is needed

Your claim holds, and it's confirmed empirically rather than by reading docs. BuildKit attestations are unsigned in-toto documents attached to the image index — there's no sigstore/OIDC step, so nothing needs `id-token`. The existing `contents: read` + `packages: write` is sufficient, and we can see it already works because provenance is on the published image today (below).

## I1 — the premise is off: provenance is already there; SBOM is the real addition

The description says the pushed manifest "carries no provenance or SBOM attestation." Provenance is already present, at `mode=max`.

`docker/build-push-action@v6` injects a default provenance attestation when it's pushing (not loading to the docker exporter) and the input isn't set — and it picks the mode based on repo visibility ([`getAttestArgs`](https://github.com/docker/build-push-action/blob/v6/src/context.ts)):

```js
if (GitHub.context.payload.repository?.private ?? false) {
  args.push('--attest', `type=provenance,${...('mode=min,inline-only=true')}`);
} else {
  // for a public repository, we set max provenance mode.
  args.push('--attest', `type=provenance,${...('mode=max')}`);
}
```

Archon is public, so it already gets `mode=max`. Confirmed against `ghcr.io/coleam00/archon:latest`:

- The index already has two `unknown/unknown` attestation manifests (`vnd.docker.reference.type: attestation-manifest`), one per platform.
- Each holds one in-toto layer with `in-toto.io/predicate-type: https://slsa.dev/provenance/v1`.
- That predicate is **243 KB** and contains `internalParameters.buildConfig.llbDefinition` with 54 steps — the full build graph, which is the `mode=max` marker. `mode=min` would be a few KB with no `buildConfig`.
- `runDetails.builder.id` is already `https://github.com/coleam00/Archon/actions/runs/…`.

There is **no** SPDX layer, so `sbom: true` is the genuine functional change here.

This doesn't change my verdict — `provenance: mode=max` is still worth keeping as an explicit pin, since it makes the behavior visible in the workflow and immunizes us against the action changing its default. But could you adjust the description so it reads as "add SBOM; pin the provenance mode that's currently implicit"? Otherwise the claim lands in the merge commit and changelog overstating what changed.

## Consumer impact: none, because the shape change already happened

This was the risk I most wanted to rule out, since it isn't visible in the diff. It's clean:

- The published manifest is **already** `application/vnd.oci.image.index.v1+json` with `unknown/unknown` entries — the image has been multi-arch plus attested for a while, so this PR doesn't introduce the index or the `unknown/unknown` platform rows that trip older tooling.
- Nothing in the repo pins a digest. `deploy/docker-compose.yml` uses `image: ghcr.io/coleam00/archon:latest`, and every docs reference (`installation.md`, `deployment/docker.md`, `docs.mdx`, `reference/configuration.md`, `cli/src/commands/setup.ts`) is a tag-based `docker run`/`docker pull`.

So `sbom: true` adds an in-toto layer to manifests that already exist. Nothing downstream changes shape.

## `mode=max` exposure: nothing sensitive

Your caveat is the right one to raise, but it's moot here — checked and clear:

- `Dockerfile` has no `ARG` instructions at all and no `--mount=type=secret`.
- `publish.yml` passes no `build-args`, `secrets`, or `secret-files`.
- The only recorded `request.args` in the live provenance are the OCI labels from `metadata-action` (title, description, licenses, revision, source, version) — all already public.

And since `mode=max` is already the effective default, this PR doesn't change the exposure surface either way.

## Cost: no recurring chore

Confirmed — no per-release manual step, no hash to maintain, no new secret. That's the property that matters most for a CI surface here; the alternative (`actions/attest-build-provenance`) would have meant a permissions change and a separate artifact to keep track of. This stays inside the existing build step.

## S1 — expect the publish job to get slower; worth watching the first tagged run

`sbom: true` makes BuildKit pull `docker/buildkit-syft-scanner` at build time and scan the produced filesystem once per target platform. Two things compound here: the image is large (Chromium plus the provider SDKs' vendor binaries — CI already carries a "free runner disk space" step noting it needs ~14 GB of headroom), and `publish.yml` builds `linux/arm64` via QEMU, so that pass is emulated. I don't think it's a blocker, but it's the one real cost and I'd rather call it out than have it surprise us. If it turns out to hurt, `sbom: generator=<pinned ref>` is the knob.

Minor related note: that scanner image is an unpinned tag fetched during a release build, which is a small supply-chain surface of its own. Fine to accept — just noting it, since it's the sort of thing this PR is otherwise about.

## S2 — verification isn't discoverable (maintainer follow-up, not yours)

The `imagetools inspect` snippet in your description is exactly what a deployer needs, and right now it lives only in this PR. `packages/docs-web/src/content/docs/deployment/docker.md` is where someone lands before granting the container a database and credentials. Three lines there would turn this from a property nobody knows about into one people actually use. Happy to take that as a separate maintainer change — no need to expand this PR.

## Verification status, stated honestly

The green `docker-build` check does **not** exercise this, and can't: it uses `push: false, load: true`, and the action explicitly skips attestations when a docker exporter is in play (`!Build.hasDockerExporter(inputs.outputs, inputs.load)`). So that check proves the workflow still parses and the image still builds — not that attestations are emitted.

- **Provenance**: verified live on the published image (so the permissions story is proven, not assumed).
- **SBOM**: unverified until a real push. `publish.yml` has `workflow_dispatch`, so it can be confirmed ahead of a release if we want, otherwise the next tag will tell us. Low risk — it rides the exact mechanism provenance already uses.

I skipped `bun run validate`; the diff touches no TypeScript and no bundled defaults, so it has nothing to say here.

## Verdict

**Merge.** Correct, minimal, no recurring maintenance, no consumer-facing risk. The only ask is I1 — retitle/reword so the description matches what actually changes (SBOM added, provenance pinned). S1 and S2 are notes, not conditions.

Appreciated the caveats section and the "no SLSA level claimed" line — that's the right posture for a supply-chain change, and it made this much faster to check.

PR #2299, opened 2026-07-29
feat(workflows): unify run output under one per-project tree in ~/.archon

Wirasm:
## Summary

- **Problem:** "The repo holds source, `~/.archon/workspaces/<project>/` holds output" was a prompt convention, not a structural fact. The identity → storage-path rule was implemented three times at three levels of correctness, and the weakest copy — the two HTTP artifact routes — used `parseOwnerRepo(codebase.name)` alone. That returns `null` for a folder project (display name, no slash) and for a no-remote local repo (bare basename), so `GET /api/runs/:runId/artifacts` answered **HTTP 200 with `{ files: [] }`**: indistinguishable from "the run wrote nothing". Artifact browsing was silently dead for two of the three project kinds Archon can register.
- **Why it matters:** two adjacent leaks compound it. `.archon/state/` had **zero engine support** — it was `mkdir -p .archon/state` inside prompts, resolved relative to cwd, so inside an isolated run it wrote to the *worktree* and the "cross-run memory" died at cleanup; and in a user's repo it is fully stageable, because Archon never writes a `.gitignore`. Meanwhile `resolveProjectPaths`'s fallback had the **engine itself** writing artifacts and logs into `<cwd>/.archon/` for any unregistered run.
- **What changed:** one key and one resolver in `@archon/paths` that all four call sites delegate to; a new `$STATE_DIR` engine variable as an external per-project sibling of `$ARTIFACTS_DIR`; a durable `workflow_runs.output_root` pointer written once at run start; the unregistered-cwd fallback relocated out of the repository; and the docs + lock tests that protect `workflow:`'s spawn-time target resolution.
- **What did NOT change (scope boundary):** `$WORKFLOWS_DIR` and a fourth discovery tier (dependent follow-up — only its documentation and lock tests are here); container output collection (documented limitation, **no implementation** — see below); a per-run artifact manifest; retention/cleanup of artifact and state directories (no code path deletes an artifacts dir today — adding one is a new destructive capability); a configurable output base path; `.archon/ralph/` (a genuine mixed case across three bundled defaults); and #1192's identity redesign, which `output_root` is precisely the mechanism for landing independently of.

### The three moves

[...]

Issue #2200, opened 2026-07-20
design: all workflow output external — unified per-project tree in ~/.archon (artifacts/logs/state), repo keeps only source; substrate for harness self-review workflows


Wirasm:
## Summary

Design proposal: **all workflow OUTPUT lives in the user-home project tree; the repo keeps only SOURCE.** Every run-produced byte (artifacts, logs, cross-run state) goes under the named per-project folders in `~/.archon/workspaces/`, giving one traceable tree per project — and making the "workflows commit their own telemetry into target repos" bug class (#1407) structurally impossible instead of prompt-guarded. Phase 2 builds on that substrate: internal review workflows that read the run/artifact tree to improve the harness itself.

## What already holds (don't redesign)

- Engine-written run artifacts are external today: `$ARTIFACTS_DIR` = `~/.archon/workspaces/<owner>/<repo>/artifacts/runs/<id>/` (+ typed `nodes/` sidecars), `_folder/<slug>/artifacts/` for folder projects, logs beside them. This is the tree to build on, not replace.
- Repo-local `.archon/{workflows,commands,scripts,config.yaml}` are **source**, versioned with the repo — they stay.

## What gets repointed

1. **`.archon/state/` (cross-run workflow state) → external per-project state dir.** Today the documented convention is repo-local + gitignored — the main surviving output-in-repo surface. Move to `~/.archon/workspaces/<project>/state/<workflow-name>/` and expose an engine variable (e.g. `$STATE_DIR`, sibling of `$ARTIFACTS_DIR`, pre-created by the executor). Bonus fix: today a `.archon/state/` relative path inside a worktree writes to the *worktree* (ephemeral, dies with cleanup) — external state makes cross-run persistence actually persistent, worktree or not.
2. **Bundled defaults sweep**: every default workflow/command referencing `.archon/state/` (and the maintainer workflows' repo-local run-state files, e.g. `reviewed-prs.json`) moves to `$STATE_DIR`. Note the maintainer dir mixes two kinds today: `direction.md` is committed knowledge (stays in repo), `reviewed-prs.json` is run state (moves out) — the redesign forces that clean split.
[...]

PR #2300, opened 2026-07-29
docs: clarify that the workflow constitution governs the YAML surface, not prompts

Wirasm:
## Summary

The rule **"YAML coordinates. Code computes. Agents judge."** was readable as a partition of *work between node types* — computation to `bash:`/`script:` nodes, judgment to `prompt:` nodes. Four places phrased it that way, and none of them said what the rule actually scopes.

It is about **what may enter the YAML surface**. It says nothing about what an agent may do inside a node.

That ambiguity produced a real misfire during the #2200 review: a reviewer (me) argued that a guard implemented as prompt instructions should be moved into a deterministic node **on constitutional grounds**. The constitution had no opinion on that. The genuine argument was reliability — the check had no judgment content and its failure had irreversible public consequences — which is a much narrower claim and should have been made on its merits.

This PR makes the scope explicit in all four locations and records the narrow reliability carve-out as explicitly *not* a constitutional argument.

### Before

```
Author writes a prompt: node that computes something.
Reviewer reads: "Computation belongs in bash:/script: nodes;
                 judgment belongs in prompts."
Reviewer concludes: this violates the constitution → refactor it.

Nothing in the docs contradicts that reading.
```

### After

```
Same prompt node.
Reviewer reads: "The rule governs the YAML surface, not what an
                 agent does inside a node. A prompt that computes
                 is a legitimate authoring choice."
Reviewer concludes: author's call. Not a constitutional question.

If the check is safety-critical, the reviewer argues RELIABILITY
(no judgment content + irreversible failure) and says so explicitly.
```

### Before

```
constitution.md  §The rule   — "computes → script node, judgment → prompt"
constitution.md  §Smell 3    — "prompt-embedded logic" listed as a workaround symptom
CLAUDE.md                    — same partition, one line
direction.md     §workaround-triage — "disguised computation → point at script nodes"
```

### After

```
constitution.md  §The rule   — [~] scopes the rule to the YAML surface; adds
                               script-vs-prompt as ordinary engineering; adds the
                               reliability carve-out, labelled NOT constitutional
constitution.md  §Smell 3    — [~] scopes "prompt-embedded logic" to RECURRENCE
                               across workflows, not any single prompt
CLAUDE.md                    — [~] same clarification, plus "never cite the
                               constitution to argue a prompt should become a script"
direction.md     §prompt-computation — [+] NEW triage clause
direction.md     §workaround-triage  — [~] scope note pointing at the new clause
```

## Label Snapshot

`type: docs`, `area: workflows`

## Change Metadata

Docs only. No code, no schema, no runtime behavior.

## Linked Issue

None — this arose from the #2200 review (PR #2299). Not `Closes` anything.

Docs-only change; no code paths touched.

- `git diff --stat` → 3 files, +18/−3.
- `bun run build:docs` **not run** — it is broken on `dev` independently of this change, for two pre-existing reasons found during #2200: a syntax error at `packages/docs-web/src/data/marketplace.ts:145` (marketplace entry missing its opening brace, introduced in `ea87ddd3`) and an Astro/`unist-util-visit` incompatibility (`does not provide an export named 'EXIT'`). Neither is in `bun run validate`, so CI is unaffected. Filing separately.
- The edited Markdown adds prose and one list only — no new components, imports, or frontmatter.

## Security Impact (required)

None. Documentation only.

## Compatibility / Migration

None. No behavior change; this only changes how existing rules are read.

Read the four changed passages and confirm the scoping matches intent — specifically that:

1. A `prompt:` node performing computation is explicitly sanctioned.
2. The reliability carve-out reads as narrow and is clearly marked as *not* a constitutional argument.
3. `§workaround-triage`'s "prompt-embedded logic" signal now clearly means *recurrence across workflows*, not a finding against any single workflow.

`CLAUDE.md` and `direction.md` are read by agents on most tasks, so the clarification takes effect immediately for future work. `direction.md` gains one new citable clause (`§prompt-computation`); existing clause names are unchanged, so prior citations stay valid.

## Rollback Plan (required)

`git revert`. No state, no schema, no data.

## Risks and Mitigations

| Risk | Mitigation |
| --- | --- |
| Read as permission to put computation in the YAML | The scoping is stated in both directions: the language still does not compute; only the node-internal question is opened up. The admissibility test is untouched. |
| The reliability carve-out gets used as a general lever to refactor prompts | It is bounded by two conditions that must BOTH hold (no judgment content, irreversible failure) and is explicitly labelled as not a constitutional argument in all three files. |

## Summary by CodeRabbit

- **Documentation**
  - Clarified that workflow YAML governs only engine-visible coordination/wiring (structure, ordering, gates/joins, retries, pauses, session/artifact identity, reusable patterns).
  - Confirmed computation and judgment belong inside node bodies, including allowing computation within `prompt` nodes.
  - Refined “script vs prompt” escape-hatch guidance without implying constitution-driven prompt→script conversion.
  - Added a narrow reliability exception for determinism arguments limited to judgment-free checks with irreversible external consequences.
  - Expanded “workaround pressure” guidance: recurring prompt-embedded logic can indicate missing primitives/patterns, not a critique of any specific workflow.


Wirasm:
---
pr: 2300
title: "docs: clarify that the workflow constitution governs the YAML surface, not prompts"
author: "Wirasm"
reviewed: 2026-07-29
---

# PR Review: #2300 — clarify that the constitution governs the YAML surface, not prompts

**Branch**: `docs/constitution-prompt-computation` → `dev`
**Files**: 3 (+18/−3), docs only
**Reviewed as**: a governance change, not prose — `CLAUDE.md` and `direction.md` are read by agents on most tasks and `direction.md` clauses produce decline verdicts.

---

## Verdict: REQUEST CHANGES

The diagnosis is right and the door-closing is sound. What's wrong is coverage: the old partition survives **verbatim in three more places inside the same two files**, including the section of `direction.md` a triage agent reads first and the section of the constitution addressed to workflow authors. An agent that stops at either lands on the pre-PR reading with nothing on the page to contradict it — which is the exact failure this PR exists to prevent.

All three fixes are one line each.

---

## Does it open any door it should not? No.

Checked directly, since this was the main risk:

| Guardrail | State after this PR |
|---|---|
| Admissibility test (constitution `:36-42`, `CLAUDE.md:100`) | byte-identical |
| Case law table (`:46-58`) — `when:` arithmetic, templating, dynamic includes, `with:` expressions | byte-identical, all still rejected |
| Smell 1 wholesale-or-nothing / `direction.md §when-grammar` | byte-identical |
| "stays out of the YAML" (`:19`), "every unit of computation that leaks into the YAML" (`:32`) | intact |

The scoping is stated in both directions and only opens the *node-internal* question. One soft spot in the new wording (S1 below), and it is a wording issue, not a load-bearing one.

---

## Issues Found

### Critical

None.

### Important

**I1 — `.archon/maintainer-standup/direction.md:30` still carries the old partition, in the highest-traffic section**

> - **Not a programming language.** … `bash:`/`script:` nodes compute; prompts judge. PRs that add computation to the YAML surface conflict — see §workflow-language.

This is in *What Archon is NOT* — the list a triage agent scans first, and the one that produces polite-decline verdicts. It states the partition the PR is retiring, and points at `§workflow-language` rather than the new clause 29 lines below. The clarification lands in the clause list; the misleading summary sits above it untouched. (CodeRabbit flagged the same line independently.)

**Fix**: `…; computation stays out of the YAML surface and lives inside nodes. PRs that add computation to the YAML surface conflict — see §workflow-language and §prompt-computation.`

---

**I2 — `packages/docs-web/src/content/docs/reference/workflow-language-constitution.md:110` repeats the partition to workflow authors**

> For **workflow authors**: … the language is telling you the logic belongs one level down. **Compute in a script node, decide in a prompt**, and let the YAML do what it's for.

This is the closing *What this means in practice* section, addressed to precisely the audience whose choice the PR is trying to unblock, ~80 lines after the clarification. In context ("if you're fighting the YAML") it is defensible, but the bare clause is the exact sentence shape the PR argues is misread — and an author who jumps to the audience-specific section, which is what that section is for, never sees the gloss.

**Fix**: `…the logic belongs one level down — into a script node or a prompt, your call — and let the YAML do what it's for: wiring the pieces the engine governs.`

---

**I3 — the workaround-triage bucket 3 is still an unqualified "point at script nodes", in both files**

- `constitution.md:86` — "missing *computation* → point to script nodes"
- `direction.md:58` — "disguised *computation* → point at script nodes"

That bucket is the *operative* instruction — it is what a triage agent executes. Unqualified, it produces "refactor this prompt into a script", which `§prompt-computation` now forbids in the very next bullet. The new **Scope:** note points at the new clause but does not amend the bucket, so the two sit side by side saying opposite things and the more actionable one wins.

The bucket was always about computation smuggled into the *YAML surface* (the mechanism paragraph lists "abused fields"). Making that explicit resolves it.

**Fix** (one word each): `disguised computation *in the YAML surface* → point at script nodes.`

### Suggestions

**S1 — "lives inside a node" is ambiguous between a node's body and a node's YAML block**

`constitution.md:19` — "stays out of the YAML and lives inside a node"; `CLAUDE.md:98` — "lives inside nodes".

A node's YAML block contains `when:`, `retry:`, `output_format:` — those are also "inside a node". The old wording named the escape hatch (`belongs in a bash:/script: node`) and could not be read that way. This is the only place where the new text is *weaker* than the old against YAML-side computation.

Exploitability is low — "stays out of the YAML" precedes it, `:32` reinforces it, and the admissibility test is intact — but this is load-bearing text and the fix is two words: **"lives inside a node's body"**.

---

**S2 — the Smell-3 rescoping is stricter than `direction.md`'s, and can blind the audit lever**

`constitution.md:88`: "A prompt doing deterministic work is **only** a smell when the same shape recurs across **many workflows**."

`direction.md:58` says "the signal is *recurrence across workflows*" — no "only", no "many". The constitution's stronger form tells an auditor to discard two genuine single-workflow signals:

1. the same shape repeated across nodes *within* one workflow (recurrence, one file);
2. the polyglot-validate case this very section cites at `:84` — deterministic suites narrated as AI prose, a real *pattern* gap that was visible before it recurred.

The failure this PR fixes was a **reviewer citing the constitution at an author**, not an **auditor noticing a gap**. The fix shouldn't cost audit sensitivity.

**Suggest**: "the smell is *recurrence* — the same shape across workflows, or repeated within one — not any single prompt that computes." Drop "only" and "many".

---

**S3 — the reliability carve-out is bounded, but doesn't say what it entitles a reviewer to do**

The bound itself holds up. Both conditions are conjunctive in all three files ("no judgment content … whose failure has irreversible external consequences" / "plus"), "no judgment content" is given a concrete test ("exactly one correct answer"), and it is de-constitutionalized explicitly ("do not cite this page" / "not by citing the constitution"). It cannot be stretched to an ordinary prompt without asserting both halves, and asserting them is falsifiable.

What's missing is the disposition. `§prompt-computation` says script-vs-prompt is "owned by the workflow author" and "never decline … on constitutional grounds" — so a triage agent that believes reliability applies, against an author who disagrees, has no rule. One clause settles it: reliability is argued on merits like any review comment and is never a `§` citation. Whether it may block on its own merits is a maintainer call I don't have a view on.

---

**S4 — the slogan survives unchanged; "Agents judge" is the half that misfired**

`constitution.md:17` blockquote and the `CLAUDE.md:97` heading both still read *YAML coordinates. Code computes. Agents judge.* Slogans travel further than their gloss, and this one is now three lines from a paragraph explaining it doesn't mean what it says.

Keeping it is defensible — citation continuity, and the gloss sits directly beneath. Noting it only because if you ever want the slogan self-correcting, *"YAML coordinates. Nodes compute and judge."* carries the whole rule without a footnote.

---

## Consistency and citation check

| Check | Result |
|---|---|
| `§prompt-computation` name collision | **None.** Existing clauses: `§deployment-recipes`, `§community-providers`, `§when-grammar`, `§load-time-composition`, `§workaround-triage`, `§schema-width`, `§implicit-behavior`, `§single-tenant-per-install`, `§single-developer-tool`. |
| Existing `§` citations still valid | **Yes** — no clause renamed, no clause meaning inverted. `.archon/commands/maintainer-standup.md` and the backfill script only match on the `direction.md §` prefix. |
| In-code constitution citations | **Still accurate** — `packages/workflows/src/schemas/workflow.ts:104` and `packages/workflows/src/dag-executor.ts:7056` both cite "code computes, YAML coordinates" about the YAML surface. |
| Three-file agreement on the new rule | Agrees on scope, on "ordinary engineering", and on the reliability carve-out. Diverges on the Smell-3 strictness (S2) and leaves three stale passages (I1, I2, I3). |
| Factual accuracy about the engine | No errors found. |

---

## Validation

Docs only — 3 files, all `.md`, no code paths touched.

| Check | Status | Notes |
|---|---|---|
| `prettier --check` on the changed files | **PASS** | Verified at `31f4f33a` in a detached worktree. (`CLAUDE.md` is in `.prettierignore`.) |
| type-check / lint / tests | **N/A** | No code files changed. |
| `bun run build:docs` | **not run** | Broken on `dev` for pre-existing reasons, disclosed in the PR body, being fixed separately. Not counted against this PR; it is not in `bun run validate`, so CI is unaffected. |

---

## Strengths

- **Fixes the class, not the instance.** Labelling the reliability argument as *not* constitutional is the load-bearing part — it hands reviewers the correct argument rather than only removing the wrong one. A pure "don't do that" edit would have left the same misfire looking for a new hook.
- **Genuinely narrow carve-out.** Two conditions that must both hold, plus a concrete test for the fuzzier one. Rare for a docs exception to be this hard to stretch.
- **The scoping is stated in both directions** — the node-internal question opens, the YAML surface does not. Verified against the admissibility test, the case law table, and `§when-grammar`, all untouched.
- **Clean clause hygiene.** New name, no collision, no existing citation invalidated.
- Provenance is documented (the #2200 / #2299 misfire), so a future reader can see what the wording is defending against — which is the thing that usually rots first.

---

## Recommendation

**REQUEST CHANGES** — on coverage, not direction.

Blocking: **I1**, **I2**, **I3** (three one-line edits). Each leaves the retired partition stated as current guidance in a place a reader reaches without passing the clarification.

Non-blocking but worth taking before merge, since the text is load-bearing and cheap to get right now: **S1** (two words, closes the only spot where the new text is weaker than the old) and **S2** (keeps the audit lever sensitive to single-workflow signals).

**S3** and **S4** are maintainer calls.

Wirasm:
## Re-review — delta `31f4f33a..e7d6d593`

Scoped to the delta as asked. All five prior findings are addressed. **One new Important issue introduced by the fix**, and it's a one-line deletion.

---

## Your three questions, answered first

**1. Are the three stale passages gone, or moved?** Gone. Verified by reading the new text and by an independent grep at `e7d6d593` for every phrasing of the partition — `nodes compute; prompts judge`, `Compute in a script node, decide in a prompt`, `judgment belongs in prompts`, `belongs in a bash:/script: node`, `disguised *computation*`, `missing *computation*`. Zero survivors across all three files. Only the two deliberate slogan sites remain (blockquote `:17`, `CLAUDE.md:97`).

**2. Can bucket 3 still yield a rewrite-the-prompt verdict?** No — and it's now blocked twice, independently:

- **Structurally**: the bucket's trigger is "computation that has leaked *into the YAML surface*", and the S1 fix at `:19` defines a prompt's text as a node **body**, explicitly not YAML surface. A computing prompt can no longer *match* the bucket at all. This is the stronger of the two guards because it works even for a reader who skips the parenthetical.
- **Explicitly**: "this bucket never produces a rewrite-the-prompt verdict" (`direction.md`) / "never about rewriting an author's prompt" (`constitution.md`).

Dropping the **Scope:** note cost nothing — the `§prompt-computation` pointer moved into the bucket itself, so the cross-reference survives at the point of use, which is better placement than where it was.

**3. Did S2 swing back far enough to re-authorize a reviewer citing at an author?** No, and the reason is structural rather than lucky: the relaxation moved the **audit-sensitivity** knob (what counts as a smell), while the **verdict** knob is held by a different sentence you didn't touch — "It is **not** a finding against the workflow, and not a mandate to refactor authored prompts into script nodes" — reinforced by the bucket-3 negation and `§prompt-computation`. So a reviewer who clears the lowered smell bar still has no path from "smell" to "cite at the author"; the smell routes to bucket 1 or bucket 2, both of which are language-design outcomes.

That separation is the right design. You can tune audit sensitivity later without ever reopening the citing-at-authors hole.

What *did* degrade is the evidence offered for the relaxation, not the rule.

---

## Important (new, introduced by this delta)

**C1 — the polyglot-validate parenthetical is factually wrong, and it's the only evidence for the relaxation**

`workflow-language-constitution.md:88`:

> (The polyglot-validate gap above was exactly that, found in a single workflow.)

The audit record says otherwise. `.claude/archon/reviews/2026-07-15-defaults-audit.md:12`, systemic finding 1:

> **validate-as-AI-loop (×6)** — `archon-validate` (341 lines) plus the validation phases of implement/implement-tasks/implement-issue/resolve-merge-conflicts/sync-pr-with-main …

Six workflows, named individually. The polyglot-validate gap is a **cross-workflow** recurrence — the same class as the copy-paste findings, not a single-workflow example. The paragraph two above it (`:84`) already implies this: "validation **suites** narrated as AI prose because **authors** lacked a polyglot pattern", both plural. So the page now contradicts itself within one section.

This matters more than a stray fact because the parenthetical is the *sole* in-page justification for "or across nodes within one workflow". A reader checking the evidence finds it doesn't support the claim, which undermines the rule it was added to support — and this page gets cited for months.

The rule stands on its own without an example. **Fix: delete the parenthetical.** If you want an example there, it needs to be a real single-workflow one, and I don't have a verified one to offer.

---

## Suggestions

**S1 — duplicated clause in the authors line** (`constitution.md:110`)

> …the language is telling you the logic belongs one level down. **Move the logic one level down** — into a `script:`/`bash:` node or a `prompt:`…

Edit artifact — the instruction is stated twice in consecutive sentences. Published docs page, so worth the ten seconds. Fix: drop the second clause and run the two together — "…the logic belongs one level down — into a `script:`/`bash:` node or a `prompt:`, whichever fits the problem (see *The rule*: that choice is yours, not the constitution's) — and let the YAML do what it's for…".

**S2 — the S1 body/fields distinction wasn't carried into `direction.md:30`**

`constitution.md:19` and `CLAUDE.md:98` both got "a node's **body** — the `bash:`/`script:` source or the `prompt:` text … surrounding node fields are YAML surface". `direction.md:30` still says only "computation lives inside nodes, not in the YAML".

Contained — the "not in the YAML" clause and "PRs that add computation to the YAML surface conflict" both sit in the same sentence, and `§when-grammar` is 26 lines below. Optional, listed only because the other two files now carry a sharper form of the same sentence.

**S3 — `§workaround-triage`'s lead-in lost the recurrence qualifier when the Scope note was dropped**

> repeated YAML structure or deterministic logic embedded in prompts is a signal

"repeated" reads as modifying "YAML structure"; whether it distributes to the second item is ambiguous. That ambiguity is exactly what the **Scope:** note used to close, and the recurrence framing now lives only in the constitution's Smell 3, not in `direction.md` — which is the doc a triage agent actually reads.

Low severity because the *harmful* outcome is blocked twice regardless (bucket-3 negation + `§prompt-computation`); what's left is over-reporting noise, not a bad verdict. One-word fix: "repeated YAML structure or **recurring** deterministic logic embedded in prompts is a signal".

---

## Not changed — both calls endorsed

**S3-prior (disposition gap)** — correct to route to Rasmus rather than invent a rule. It does **not** block: the ambiguity only bites when a reviewer and author disagree about a check that is simultaneously judgment-free and irreversible, which is a narrow and visible case, and the reviewer already has the right vocabulary to raise it. Inventing a disposition rule in this PR would have been the worse error.

**S4-prior (slogan)** — you asked whether the new text makes the unqualified slogan read *worse*. It reads better, not worse. Two lines below the blockquote, "Code computes, agents judge" is now explicitly named as *shorthand for "the language does not compute"* — the slogan is glossed at its point of use, which it wasn't before. The residual risk is only that the slogan travels without its gloss (quoted in a PR comment, in a heading). Unchanged from my earlier read, still not worth a rewrite.

---

## Validation

| Check | Status |
|---|---|
| `prettier --check` on all three files at `e7d6d593` | **PASS** (verified in a detached worktree, since removed) |
| Independent grep for surviving unqualified partition | **PASS** — zero across all three files |
| Guardrails still untouched (admissibility test, case law, Smell 1, `§when-grammar`) | **PASS** — not in the delta |
| type-check / lint / tests | N/A — no code files |
| `build:docs` | not run, per prior disclosure |

---

## Verdict

**Approve once C1 is deleted.** Everything the PR set out to do is done, and the bucket-3 fix landed stronger than what I suggested — defining a prompt's text as node body rather than YAML surface closes the hole structurally instead of by instruction, which survives a careless reader.

C1 is one deletion and should not be merged as-is: a factual error in the canonical governance page, offered as the evidence for a rule, is the exact failure mode this PR exists to correct. S1 is a ten-second prose fix worth taking in the same push. S2 and S3 are optional.

Wirasm:
## Confirmation pass — delta `e7d6d593..6bb2925a`

**APPROVE.** C1, S1, and S3 are all clean. Answers to the two questions you raised, both checked against the corpus rather than by eye.

---

### 1. Is "or across nodes within one workflow" actually supported?

**Yes — keep the clause.** It is not an unevidenced rule. The audit's systemic finding 3 is a clean instance:

> **The five review "agents" are one template ×5** (~600 duplicated lines: same LOAD phase, same finding template, same metadata scaffold). One parameterized review command (or one command + per-dimension skills) replaces them.
> — `.claude/archon/reviews/2026-07-15-defaults-audit.md:14`

Those five agents are five sibling nodes inside a *single* workflow — finding 2 gives the block's shape as `scope→sync→5 agents→synthesize→fix`, which is exactly the 9 nodes it is described as. So: same shape recurring across nodes within one workflow, classified as a real finding, with a remedy that is a **parameterized command** — the *missing pattern* bucket, precisely the outcome the sentence claims recurrence indicates.

Two weaker corroborations in the same document: `:35` prescribes "extract shared loop; trim per-iteration mega-prompt" for ralph-dag (within-workflow repetition), and finding 5 flags `PHASE_N_CHECKPOINT` checklists and unfilled report-template fields recurring inside single mega-commands.

One honest qualification, since it affects how the clause should be read: finding 3 is repeated prompt *scaffolding* (LOAD phase, finding template, metadata scaffold), not deterministic *computation* narrowly. It supports the clause as written — Smell 3's mechanism explicitly covers "copy-pasted blocks, abused fields, prompt-embedded logic" as three species of one pressure, and the clause's own consequent is "a missing primitive **or an undocumented pattern**". But it is pattern evidence, not primitive evidence.

**On re-adding an example: optional, and I'd lean against.** The page reads fine unexampled, and an example is exactly what went wrong last time. If you do want one, the five-review-agents case is the honest one — verifiable at audit `:14`, and it should be framed as a *pattern* example, not a primitive one.

### 2. Did the deletion orphan the surrounding text?

**No — the seam is clean.** The sentence before the deletion ends on its own conclusion ("…because recurrence is what indicates a missing primitive or an undocumented pattern"), not mid-argument; the sentence after opens an independent contrast ("What is *not* a smell is…"). The parenthetical was a free-standing aside between two complete sentences, so removing it leaves nothing dangling. The paragraph now runs: signal → not a finding against the workflow → what makes it a smell → what does not. That is a tighter argument than it was with the aside.

"Undocumented pattern" also still has its example elsewhere — `:84` narrates the polyglot-validate case, and `direction.md §workaround-triage` carries the `detect-AI → execute-bash → fix-AI` shorthand inline.

**One trivial nit from the other seam** (S1's collapse at `:110`, non-blocking): dropping the sentence break left a mood mismatch — the clause is now declarative ("the language is telling you the logic belongs one level down — into a `script:`/`bash:` node or a `prompt:` —") but still lands on an imperative it no longer has anything to coordinate with ("and let the YAML do what it's for"). Previously the imperative paired with "Compute…, decide…". Reads fine; a copy-editor would swap the tail to "leaving the YAML to do what it's for." Entirely optional.

---

### Validation at `6bb2925a`

| Check | Status |
|---|---|
| `prettier --check`, all three files | **PASS** |
| Deleted claim (`found in a single workflow`) and edit artifact (`Move the logic one level down`) | **zero occurrences** repo-wide |
| Unqualified partition survivors | **zero** across the three files |
| Guardrails (admissibility test, case law, Smell 1, `§when-grammar`) | not in the delta, untouched |

Carried forward, neither blocking: the disposition gap (escalated) and the retained slogan (accepted residual risk, and it now reads better than before since it is glossed at its point of use).

**Approve.**

PR #1687, opened 2026-05-14
feat(marketplace): add archon-resolve-mr-conflicts (GitLab counterpart)

lraphael:
## Summary

- **Problem**: Archon ships `archon-resolve-conflicts` for GitHub PRs. GitLab MRs with merge conflicts need the same automation.
- **Why it matters**: Completes the GitLab MR suite alongside the already-merged `archon-smart-mr-review` (#1681) and pending `archon-comprehensive-mr-review` (#1686). Handles rebase + auto-resolve + validate + force-push entirely via `glab`.
- **What changed**: Adds one marketplace entry pointing to `lraphael/archon-gitlab-workflows@6e39b359/archon-resolve-mr-conflicts` (workflow YAML + 2 commands: `mr-fetch` + `resolve-mr-conflicts`).
- **What did NOT change**: No code in Archon itself. Only `packages/docs-web/src/data/marketplace.ts` (+12 lines).

## Change Metadata

- Change type: `feature`
- Primary scope: docs (marketplace registry)

## Linked Issue

- None — community contribution. Completes the GitLab MR suite (#1681 merged, #1686 pending).

## Validation Evidence

```bash
bun run validate
# exit 0 (with isolated HOME to avoid local ~/.archon/workflows pollution
# of dag-executor.test.ts — pre-existing issue)

bun run packages/docs-web/scripts/lint-marketplace.ts
# PASSED — all 8 entries valid; new entry verified at SHA 6e39b359
#   ✓ [archon-resolve-mr-conflicts] directory verified at 6e39b359
```

## Security Impact

- New permissions/capabilities? **No** — uses standard `glab` CLI + local `git` operations
- New external network calls? Only `glab` API + standard `git fetch`/`git push`
- Secrets/tokens handling changed? **No** — uses existing `glab auth` session
- File system access scope changed? **Yes (intentional)** — workflow modifies files during conflict resolution and runs `git push --force-with-lease` on the MR source branch. This is the core function of the workflow; it runs inside Archon's worktree isolation, so the user's main working tree is never affected.

**Self-attestation** (per CONTRIBUTING.md):
- [x] The workflow does not exfiltrate data, credentials, or secrets
- [x] The workflow does not execute destructive operations without user confirmation — `--force-with-lease` protects against foreign pushes; for complex conflicts (`Option D`), the workflow stops **before** push and presents an options report for manual decision
- [x] I have the right to share this workflow publicly (MIT licensed)
- [x] The pinned SHA points to a reviewed, stable version

[...]

PR #2301, opened 2026-07-29
fix(docs): repair marketplace syntax error and close the docs-build CI gap

### LINKED ISSUE #2281: ci: add `bun run build` — docs-web gets zero PR-time validation
## Problem

`bun run validate` (root `package.json`) runs nine gates but **not** `bun run build`, and CI doesn't build the docs site either. `deploy-docs.yml` only fires on push to `main` with a `packages/docs-web/**` path filter.

Consequence: an Astro/Starlight/sharp incompatibility in `packages/docs-web` is exercised by nothing at PR time and first appears **after merge to main, as a broken docs deploy**.

Surfaced while reviewing #2253, which bumps astro 6.1→6.4.8 and forces sharp 0.33→0.35 past Astro's declared `^0.34.0`. There is no way to validate that PR's docs half before merging it — which is the actual reason it can't be reviewed with confidence.

## Proposal

Add a docs build to CI — either `bun --filter @archon/docs-web build` as its own job (cheap, path-filtered on `packages/docs-web/**` + `bun.lock`), or `bun run build` in the existing test job if the runtime cost is acceptable.

Also worth deciding whether `bun run build` belongs in `validate` itself, so local pre-PR validation matches CI.

## Acceptance criteria

- [ ] A PR that breaks the docs build fails CI instead of breaking the `main` deploy
- [ ] Path-filtered so unrelated PRs don't pay the cost
- [ ] `validate` and CI agree on whether the build is in scope

Wirasm:
## Summary

- **Problem:** `bun run build:docs` has been broken on `dev` since 2026-07-21. `packages/docs-web/src/data/marketplace.ts` has a syntax error — the `archon-comprehensive-mr-review` entry lost its opening `{`. Nothing on the PR path caught it: `packages/docs-web` has no `type-check` script (so `bun --filter '*' type-check` skips it) and `eslint.config.mjs` ignores `packages/docs-web/**` outright, so `bun run validate` and the `Type check` step in `test.yml` are both blind to the entire package.
- **Why it matters:** `deploy-docs.yml` only builds on push to `main`. It last ran green on 2026-07-20 (Release 0.6.0); the break landed 2026-07-21. The next release to `main` would have failed the Pages deploy — a latent landmine, not a visible failure.
- **What changed:** Restored the missing brace (1 line). Added a `type-check` script to `@archon/docs-web` so the package is no longer invisible to the existing required CI gate. Added `docs-build.yml`, a path-filtered PR job that runs `bun run build:docs`.
- **What did NOT change (scope boundary):** No docs-site restructuring, no dependency bumps, no lockfile changes, no `overrides`/resolution pins, no change to `bun run validate`, no change to `deploy-docs.yml`, `test.yml`, or `marketplace-lint.yml`. Three files touched, +37/-1.

### On the reported `unist-util-visit` / `EXIT` failure — not a repo defect

Investigated and **ruled out**. After fixing the brace, `bun run build:docs` fails locally with `does not provide an export named 'EXIT'`. That is an artifact of the local checkout layout, not of this repo:

- Astro's CLI runs under **Node** (`#!/usr/bin/env node`). During prerender, Vite externalizes a bare `import { visit, EXIT, SKIP } from 'unist-util-visit'` into `packages/docs-web/dist/.prerender/chunks/*.mjs` (originating from `hast-util-to-mdast`, via `rehype-remark` → `starlight-llms-txt`). Node resolves that bare specifier by walking **up the directory tree** from `dist/.prerender/chunks/`.
- Bun 1.3.11 uses the **isolated** linker here (`node_modules/.bun/`, 14 root entries), so a clean checkout has no stray hoisted copy up-tree and the build succeeds.
- In a git worktree nested under a repo installed with an **older hoisted** Bun, that upward walk escapes the worktree and lands on the parent's hoisted `unist-util-visit@2.0.3` — CJS, which assigns `visit.EXIT = EXIT` as a property rather than exporting it statically, so ESM named-import binding fails.

Verified both directions in an isolated clone outside any parent `node_modules`:

| Environment | Result |
|---|---|
| Clean clone + brace fix | `build:docs` **passes** (85 pages, exit 0) |
| Same clone + hoisted `unist-util-visit@2.0.3` planted up-tree | reproduces the **exact** `EXIT` error |
| Planted copy removed again | passes again |

The lockfile is byte-identical in the relevant entries between Release 0.6.0 (which deployed green in CI on 2026-07-20) and `dev` today — same `astro@6.1.3`, same resolutions. **No dependency bump, override, or resolution pin is warranted**, and adding one would pin a transitive dep to paper over a local environment quirk. The brace was the only real defect.

### Before

```
Contributor            CI (PR to dev)              main / release
───────────            ──────────────              ──────────────
edits marketplace.ts
  ──────────────────▶  Type check  ── PASS (docs-web has no
                                        type-check script)
                       Lint        ── PASS (docs-web is eslint-ignored)
                       Format      ── PASS (prettier's TS parser
                                        recovers from the error)
                       Marketplace ── FAIL  (advisory, not required)
                       lint
  merged ───────────────────────────────────────▶ deploy-docs.yml
                                                   runs build:docs
                                                   ✗ BROKEN DEPLOY
                                                     (discovered at release)
```

### After

```
Contributor            CI (PR to dev)              main / release
───────────            ──────────────              ──────────────
edits marketplace.ts
  ──────────────────▶  Type check  ── *FAIL* (docs-web now included
                                        via new type-check script)
                       [+ Docs Build] ── *FAIL* (path-filtered;
                                          runs real build:docs)
                       Marketplace ── FAIL
                       lint
  *blocked before merge*
                                                  deploy-docs.yml
                                                  ✓ stays green
```

### Before

```
  bun run validate ──▶ bun --filter '*' type-check ──▶ [core, server, web, cli,
                   │                                    workflows, adapters,
                   │                                    git, paths, isolation,
                   │                                    providers]
                   │                                    ✗ docs-web (no script)
                   ├──▶ eslint . ──x── packages/docs-web/** (globally ignored)
                   └──▶ prettier --check . ──▶ docs-web (parses, tolerates)

  .github/workflows/
    test.yml            (PR → dev/main) ──▶ bun run type-check   [blind to docs-web]
    marketplace-lint.yml(PR, path-filtered) ──▶ lint-marketplace.ts  [advisory]
    deploy-docs.yml     (push → main ONLY) ──▶ bun run build:docs [too late]
```

### After

```
  bun run validate ──▶ bun --filter '*' type-check ═══▶ [+ docs-web]
                   │                                     (astro sync && tsc)
                   ├──▶ eslint . ──x── packages/docs-web/** (unchanged)
                   └──▶ prettier --check . (unchanged)

  .github/workflows/
    test.yml            (PR → dev/main) ═══▶ bun run type-check [now covers docs-web]
    marketplace-lint.yml(unchanged)
  [+] docs-build.yml    (PR, path-filtered) ═══▶ bun run build:docs
    deploy-docs.yml     (unchanged)
```

**Connection inventory**

| `bun run type-check` | `@archon/docs-web` | **new** | Package had no `type-check` script; now participates in `bun --filter '*' type-check` |
| `@archon/docs-web type-check` | `astro sync` | **new** | Generates `.astro/` (`astro:content`) types; bare `tsc` fails without them on a fresh checkout |
| `docs-build.yml` | `bun run build:docs` | **new** | Path-filtered PR job |
| `test.yml` | `bun run type-check` | unchanged | Same command; its coverage widens as a side effect |
| `deploy-docs.yml` | `bun run build:docs` | unchanged | Still the `main` deploy path |
| `marketplace-lint.yml` | `lint-marketplace.ts` | unchanged | Still advisory; see Risks |
| `bun run validate` | `bun run build:docs` | **not added** | Deliberate — see Risks |
| eslint | `packages/docs-web/**` | unchanged | Still ignored; out of scope |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `docs`, `ci`
- Module: `docs-web:marketplace`, `ci:workflows`

## Change Metadata

- Change type: `bug`
- Primary scope: `multi` (docs + ci)

## Linked Issue

- Closes #2281
- Related #2253 (the astro 6.1→6.4.8 / sharp bump that motivated #2281 — it now gets a PR-time docs build)
- Depends on # — none
- Supersedes # — none

```bash
bun run validate      # exit 0 — all nine gates
bun run build:docs    # exit 0 in a clean checkout (see note)
```

- **Evidence provided:**
  - `bun run validate` → exit 0 on this branch.
  - `bun run build:docs` → exit 0, `85 page(s) built`, in a clean clone outside any parent `node_modules`. It fails in a nested worktree for the environment reason documented above — reproduced, isolated, and proven by planting/removing a hoisted `unist-util-visit@2.0.3` up-tree.
  - Guard proven to actually catch the bug: with the brace removed again, `bun run type-check` (the exact command `test.yml` runs) exits **2** with `src/data/marketplace.ts(146,9): error TS1005: ',' expected.` With the brace restored, exit **0**.
  - Guard proven safe on a fresh checkout: `rm -rf packages/docs-web/.astro && bun --filter @archon/docs-web type-check` → exit 0 in ~1.5s (and in the clean clone, which has never been built).
  - Data integrity: `bun packages/docs-web/scripts/lint-marketplace.ts` → `Marketplace lint PASSED — all 14 entries valid`, all 14 sources verified at their pinned SHAs.
- **Intentionally skipped:** none.

## Security Impact (required)

- New permissions/capabilities? **No** — `docs-build.yml` declares no `permissions:` block and only reads the repo (unlike `deploy-docs.yml`, which needs `pages: write`).
- New external network calls? **No** new call sites. The new CI job runs `bun install` + `astro build`, the same commands `deploy-docs.yml` already runs.
- Secrets/tokens handling changed? **No** — no secrets referenced.
- File system access scope changed? **No.**
- The job is `pull_request` (not `pull_request_target`), so forked PRs run with a read-only token and no secret access.

## Compatibility / Migration

- Backward compatible? **Yes** — the marketplace fix is a pure syntax repair; entry data is byte-identical.
- Config/env changes? **No.**
- Upgrade steps: none. `bun run type-check` picks up the new package script automatically; no `bun install` required (no dependency changes, lockfile untouched).

**Verified scenarios:**
- Reproduced both reported failures from a clean `origin/dev` checkout before changing anything.
- Traced cause 1 to `ea87ddd3` by reading the commit diff: it inserted 11 lines directly after the existing `{`, so the new entry consumed that brace. Confirmed **no fields were lost or duplicated** — 14 opening braces / 14 `slug:` keys, and the marketplace lint validates every required field on all 14 entries.
- Confirmed the introducing PR (#1687) merged with `Validate marketplace entries` **already failing** — the CI log shows the exact parse error at `marketplace.ts:146:11`. The guard existed; it was not blocking.
- Empirically tested which existing gates catch the error, rather than assuming: `tsc` **catches** it; prettier **does not** (its TypeScript parser recovers and reformats the invalid file without complaint, exit 0); eslint never sees the file.
- Confirmed `deploy-docs.yml`'s last green run was Release 0.6.0 on 2026-07-20, one day *before* the break — so the failure was latent, never observed.
- Verified the new `type-check` fails on the bug and passes without it, and that it works on a checkout with no pre-generated `.astro/` directory (the CI condition).

**Edge cases checked:**
- Fresh checkout with no `.astro/` types → `astro sync` handles it; bare `tsc` would have failed CI (caught before shipping).
- Windows leg of the `test.yml` matrix: the script uses `astro sync && bun x tsc --noEmit`, no shell-specific syntax, consistent with the `bun x tsc --noEmit` other packages already use.
- Nested-worktree false failure for `build:docs` — the specific reason it is kept out of `validate`.

**What was not verified:**
- The new `docs-build.yml` has not executed on GitHub's runners yet; it will run on this PR (which touches `packages/docs-web/**`), so it is self-testing.
- No visual/rendering review of the built docs site — the marketplace fix restores an entry that was previously unparseable, so this is the first commit where the page can render at all.

- **Affected subsystems:** the docs site build, `bun run type-check` / `bun run validate`, and CI PR checks. Zero runtime code touched — nothing in `packages/{core,server,workflows,adapters,cli,web,git,paths,isolation,providers}` changes, and no shipped binary or Docker image content is affected.
  - `bun run type-check` now type-checks 13 additional files, so a pre-existing type error in `packages/docs-web` would newly fail CI. Verified there are none today (`tsc` is clean on the current tree).
  - `docs-build.yml` adds ~1 min of CI to PRs that touch `packages/docs-web/**` or `bun.lock`. `bun.lock` is included on purpose: a dependency change is precisely how an Astro/remark incompatibility would enter (the scenario in #2281 and #2253). It changed 4 times in the last two weeks, so the added cost is small.
- **Guardrails / early detection:** the failure mode of both guards is a loud red CI check. Neither can fail silently — the whole point of the change.

## Rollback Plan (required)

- **Fast rollback:** `git revert 0d1dc7f0` drops both guards and restores prior CI behavior exactly; `git revert bd1960d5` reverts the marketplace fix (do **not** revert that one alone — it re-breaks the docs build). Reverting the guard commit alone is safe and leaves the site building.
- Feature flags / toggles: none needed — the two commits are independently revertible by design.
- **Observable failure symptoms:** if the `type-check` script misbehaves, `bun run type-check` fails for every contributor on every PR (immediately obvious, and the `astro sync` step is the only new moving part). If `docs-build.yml` misbehaves, only PRs touching the docs site see a red `Build docs site` check.

## Risks and Mitigations

- **Risk:** `build:docs` is in CI but not in `bun run validate`, so a contributor can pass local validation and still get a red docs build. This is a deliberate deviation from the third acceptance criterion of #2281 ("`validate` and CI agree on whether the build is in scope").
  - **Mitigation:** the common failure class — TS/JS syntax and type errors in `src/data`, `src/content.config.ts`, `src/pages`, `scripts/` — *is* now covered locally by `type-check`, which is the class that actually broke here. The full Astro build is kept out of the local loop for two reasons: it is irrelevant to the large majority of changes, and it resolves externalized bare imports out of `packages/docs-web/dist/`, which escapes the package in nested-worktree checkouts and produces false failures unrelated to the change under test (exactly what happened while investigating this). In a worktree-heavy repo that would be a persistent DX tax. It stays one command away locally: `bun run build:docs`. **Happy to add it to `validate` instead if you'd rather have strict parity — say the word.**
- **Risk:** `marketplace-lint.yml` already caught this exact error on PR #1687 and the PR merged anyway, so guards that merely *report* can be ignored.
  - **Mitigation:** the primary guard here is deliberately routed through `bun run type-check` inside `test.yml` — the job that already runs on every PR — rather than into another separate advisory workflow. Making any check truly blocking requires branch-protection settings, which are outside a PR's reach. **Recommend** adding `test` (and optionally `Build docs site`) to the required-checks list for `dev`. Worth noting `marketplace-lint.yml` performs live network calls to verify pinned SHAs, so it can fail for reasons unrelated to the diff — which likely explains why it isn't required and gets waved through. The new guards are fully offline and deterministic.
- **Risk:** `astro sync` in the `type-check` script is a new prerequisite step that could break if Astro changes its CLI.
  - **Mitigation:** it is the documented, supported way to generate `astro:content` types, and it is already implicit in `astro build`. If it ever fails, `bun run type-check` fails loudly for everyone rather than silently skipping the package — strictly better than the current state, where the package is skipped with no signal at all.

## Additional findings (not fixed here)

Two pre-existing, non-blocking warnings surfaced by the now-working docs build. Left alone to keep this PR minimal — flagging them in case you want them filed:

- `packages/docs-web/src/content/docs/adapters/github-app-setup.md` — two code blocks tagged `caddyfile`, a language Shiki doesn't bundle. Silently falls back to `txt` (no syntax highlighting).
- `packages/docs-web/src/content/docs/deployment/docker.md` — one code block tagged `env`, same fallback. `ini` or `dotenv` would highlight correctly.

Neither fails the build. No other content, link, or asset errors appeared across all 85 pages.

## Summary by CodeRabbit

* **Documentation**
  * Improved documentation site validation during proposed changes.
  * Added a documentation type-check command for more reliable content updates.

* **Chores**
  * Added automated documentation build checks for relevant changes.
  * Improved formatting consistency in marketplace documentation data.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,0 +1,44 @@
+name: Docs Build
+
+# The docs site is only built by deploy-docs.yml, which runs on push to main.
+# That means a change that breaks `bun run build:docs` can sit on dev unnoticed
+# until release day, when the Pages deploy fails. This job runs the same build
+# on PRs that touch the docs site (or the lockfile it resolves against), so the
+# breakage surfaces before merge instead of at release.
+#
+# Path-filtered on purpose: a full Astro build costs ~1 minute, and there is no
+# reason to pay it on the ~95% of PRs that never touch packages/docs-web.
+
+on:
+  pull_request:
+    paths:
+      - 'packages/docs-web/**'
+      - 'bun.lock'
+      - '.github/workflows/docs-build.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build docs site
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      # Astro's CLI runs under Node, not Bun — keep this matched to
+      # deploy-docs.yml so a green PR check means a green deploy.
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.11
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Build docs
+        run: bun run build:docs
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,11 @@ concurrency:
 jobs:
   test:
     strategy:
+      # Both legs must always report. With the default fail-fast, an ubuntu
+      # failure cancels windows, so a windows-only break stays invisible until
+      # the next round — and a cancelled leg blocks a merge once it is a
+      # required check.
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
--- a/packages/docs-web/package.json
+++ b/packages/docs-web/package.json
@@ -6,7 +6,8 @@
   "scripts": {
     "dev": "astro dev",
     "build": "astro build && bun run scripts/normalize-llms-txt.js",
-    "preview": "astro preview"
+    "preview": "astro preview",
+    "type-check": "astro sync && bun x tsc --noEmit"
   },
   "dependencies": {
     "astro": "^6.1.0",
--- a/packages/docs-web/src/content/docs/adapters/github-app-setup.md
+++ b/packages/docs-web/src/content/docs/adapters/github-app-setup.md
@@ -176,7 +176,7 @@ Archon enforces this at startup: with App mode active and the server bound to a 
 
 Example Caddy snippet that drops `/internal/*` (bare-metal):
 
-```caddyfile
+```txt
 example.com {
   @internal path /internal/*
   respond @internal 404
@@ -203,7 +203,7 @@ The `!override` tag (compose-spec) replaces the base file's `ports` list instead
 
 Insert this `handle` block before the fallthrough `handle { }` block:
 
-```caddyfile
+```txt
 handle /internal/* {
   respond "Not Found" 404
 }
--- a/packages/docs-web/src/content/docs/deployment/docker.md
+++ b/packages/docs-web/src/content/docs/deployment/docker.md
@@ -128,7 +128,7 @@ docker compose --profile with-db up -d
 ```
 
 Then add to `.env`:
-```env
+```dotenv
 DATABASE_URL=postgresql://postgres:postgres@postgres:5432/remote_coding_agent
 ```
 
--- a/packages/docs-web/src/content/docs/deployment/docker.md
+++ b/packages/docs-web/src/content/docs/deployment/docker.md
@@ -128,7 +128,7 @@ docker compose --profile with-db up -d
 ```
 
 Then add to `.env`:
-```dotenv
+```ini
 DATABASE_URL=postgresql://postgres:postgres@postgres:5432/remote_coding_agent
 ```
 
--- a/packages/docs-web/src/data/marketplace.ts
+++ b/packages/docs-web/src/data/marketplace.ts
@@ -143,6 +143,7 @@ export const marketplaceEntries: MarketplaceEntry[] = [
     tags: ['automation'],
     archonVersionCompat: '>=0.3.0',
   },
+  {
     slug: 'archon-comprehensive-mr-review',
     name: 'Comprehensive GitLab MR Review',
     author: 'lraphael',


Wirasm:
# Review: #2301 — repair marketplace syntax error and close the docs-build CI gap

**Verdict: APPROVE.** Three files, +47/−1, no runtime code. Every claim in the description that I could test, I tested independently, and all of them hold. The deliberate deviation from #2281's third acceptance criterion is **upheld** — with one correction to the framing.

Reviewed as a process change rather than by diff size, as it deserves.

---

## 1. The brace fix is data-complete — verified conclusively

Rather than counting braces, I diffed the fixed file against the **last-known-good** version (the commit before the break):

```
git diff ea87ddd3^:packages/docs-web/src/data/marketplace.ts \
         HEAD:packages/docs-web/src/data/marketplace.ts
```

The entire diff is the clean 12-line `archon-resolve-mr-conflicts` block that `ea87ddd3` intended to add. **Nothing else differs** — no field lost, gained, reordered, or mutated on any other entry. That is stronger evidence than a field audit: the file is provably `last-known-good + exactly the intended addition`.

Corroborating:

| Check | Result |
|---|---|
| Braces/slugs **inside `marketplaceEntries`** (L49–218) | 14 open / 14 close / 14 `slug:` |
| Duplicate slugs | none |
| `lint-marketplace.ts` | `PASSED — all 14 entries valid`, all 14 sources resolve at pinned SHAs |
| Commits touching the file since the break | only `bd1960d5` (this fix) |

A note on the description: the stated "14 opening braces" is only true *within the entries array* — a whole-file `grep -c '^  {'` returns 15, because `tagConfig` (L13–41) contributes one. Immaterial to correctness, but the array-scoped count is the one that means anything.

## 2. The `type-check` script catches the class of bug — verified

Tested in a **sibling** worktree with no `.astro/` directory and no up-tree `node_modules` (confirmed clean at every level from the checkout to `/`) — i.e. the CI condition.

| Test | Result |
|---|---|
| Bare `tsc --noEmit`, no `.astro/` | **fails** — 7× `TS2307: Cannot find module 'astro:content'` |
| `bun --filter @archon/docs-web type-check`, no `.astro/` | **exit 0** (~2.2s for `astro sync`) |
| `bun run type-check` (root, exact `test.yml` command) with brace removed | **exit 2** — `marketplace.ts(146,9): error TS1005: ',' expected` |
| Same, brace restored | **exit 0**, 11/11 packages |

So `astro sync` is genuinely load-bearing, not defensive padding, and the failure **does** propagate out through `bun --filter '*'` — that was the part worth confirming, since a swallowed non-zero exit would have made the whole guard decorative.

`@archon/docs-web` appears explicitly in the passing-package list, confirming the `--filter '*'` glob picks the new script up. Windows matters here too, and `test (windows-latest)` passed on this PR — the `astro sync && bun x tsc --noEmit` form is shell-agnostic as claimed.

**Broader than advertised:** I enumerated every package against a `type-check` script. All 11 now have one — `docs-web` was the *only* gap. This doesn't just fix one package, it closes the entire "package invisible to the required type gate" class. That's worth stating in the PR body; it undersells itself.

## 3. `docs-build.yml` triggers correctly

- **No double-run with `deploy-docs.yml`.** Disjoint events: `pull_request` vs `push: [main]`. On a release PR the two run sequentially (PR check, then deploy), never concurrently.
- **Docs content is self-contained.** I checked for escapes out of the package (`../../`, `process.cwd`, custom `root`/`srcDir`, remote loaders) — the only hits are markdown code samples and intra-package imports. `packages/docs-web/**` is a sound content filter.
- **`bun.lock` inclusion is better than it looks.** `deploy-docs.yml` is filtered on `packages/docs-web/**` *only*, so a lockfile-induced build break merged without touching the docs dir would never trigger the deploy job at all — it would ambush a later, unrelated docs PR. The new job's `bun.lock` filter is genuinely stricter than the deploy path it guards.
- **Breaks from outside the filtered paths** are largely still caught, because the `Type check` step in `test.yml` is *not* path-filtered and now covers `docs-web`. The residual gap is narrow (see S1).
- No `branches:` filter, unlike `test.yml`'s `[main, dev]` — this matches `marketplace-lint.yml`'s existing shape, and errs toward more coverage. Not an issue.

## 4. The `validate` deviation — upheld, with a correction

I reproduced both directions on the **same commit and lockfile**, differing only in checkout location:

| Environment | `bun run build:docs` |
|---|---|
| Sibling worktree (`Archon-review2301`, no up-tree `node_modules`) | **exit 0** — 85 pages in 8.29s |
| Nested worktree (`Archon/.claude/worktrees/…`) | **exit 1** — `'unist-util-visit' does not provide an export named 'EXIT'` |

The mechanism is exactly as diagnosed. The repo root carries **862** hoisted `node_modules` entries from an older Bun, including `unist-util-visit@2.0.3`, whose `index.js` is CJS doing `module.exports = visit; visit.EXIT = EXIT` — a property assignment ESM named-import binding cannot see. A fresh Bun 1.3.11 install produces **14** isolated top-level entries with no `unist-util-visit`, which is why both worktrees' own installs are innocent and only the upward walk out of `dist/.prerender/chunks/` finds the bad copy.

**Refusing to pin the dependency was the right call.** A pin here would have encoded a local environment artifact into the lockfile permanently.

**The correction:** calling this a *permanent* DX tax overstates it. It is contingent on a stale hoisted parent install — `bun install` at the repo root would collapse those 862 entries to 14 and nested builds would pass. It is not intrinsic to worktrees.

**I still uphold the deviation**, on the stronger of the two arguments actually available:

1. A `validate` gate that fails for reasons unrelated to the diff is precisely what teaches people to wave red checks through — the *exact* failure mode that let #1687 land. Adding a gate with a known environment-dependent false-failure mode would work against this PR's own purpose.
2. The bug class that actually broke here — TS/syntax in `src/data`, `content.config.ts`, `pages/`, `scripts/` — *is* now covered locally by `type-check`. The full Astro build adds coverage for content-schema and render-time failures only.

**Is the CI job an adequate substitute?** For #2281's underlying goal — never let a broken docs build reach `main` — yes, and with better lockfile coverage than the deploy job. For the literal wording ("`validate` and CI agree on whether the build is in scope") — no, and the PR says so plainly. The residual risk is a contributor passing locally and seeing a red `Build docs site` on their own PR: loud, immediate, self-correcting, and about one minute of CI. That is a good trade, not a gap.

---

## Findings

No Critical. No Important.

### Suggestions

**S1 — Root `package.json` is outside the path filter.** `build:docs` is defined there; an edit to it wouldn't trigger the job. Narrow in practice (the `Type check` step is unfiltered and covers TS errors regardless), but adding `- 'package.json'` is one line and matches the repo root only. CodeRabbit raised this too.

**S2 — `type-check` is now a *writing* command, and it surfaces three warnings on every `bun run validate`.** `astro sync` generates `packages/docs-web/.astro/` (gitignored, harmless) and prints the three pre-existing expressive-code warnings you flagged under "Additional findings":

```
[WARN] language "caddyfile" could not be found. Using "txt" instead.   (×2, github-app-setup.md)
[WARN] language "env" could not be found. Using "txt" instead.         (×1, docker.md)
```

Your change promotes these from "only visible during a docs build" to "visible in every contributor's `validate` output". That flips them from cosmetic to a small standing noise tax, and argues for folding the two-token fix (`caddyfile`→`txt`, `env`→`ini`) into this PR rather than filing it. Your call — but the change is what made them everyone's problem.

**S3 — No `permissions:` block on `docs-build.yml`.** Consistent with `marketplace-lint.yml` (also none), so not a regression, and it's a `pull_request` job so forks already get a read-only token. `deploy-docs.yml` does declare one; `permissions: {contents: read}` is free hardening if you're touching the file anyway.

**S4 — Sharpen the branch-protection recommendation.** I checked: rulesets for **both `dev` and `main` already exist** (ids 15166646 and 15166632) and are `"enforcement": "disabled"`. So the ask isn't "set up required checks", it's "flip two existing rulesets to enabled and add `test` (+ optionally `Build docs site`) to required status checks". Worth saying precisely, since it's a toggle rather than a project.

---

## Does this actually close the hole?

Honestly: **it closes the part a PR can close, and the PR is candid about the rest.**

- **Local detection — genuinely new.** `bun run validate` is mandated pre-PR by CLAUDE.md and previously caught *nothing* here; `marketplace-lint.yml` was CI-only. Now the error fails locally before a PR exists. This is the most valuable part of the change and it's underplayed in the description.
- **CI visibility — improved.** Moving the signal into `test.yml` (which everyone reads) instead of a fourth advisory workflow is the right instinct, given the root cause was process rather than missing tooling.
- **Enforcement — unchanged, and can't be changed from a PR.** With both rulesets disabled, a future PR can still merge red exactly as #1687 did. Worth noting `Run marketplace auto-review` was *also* failing on #1687 — so two guards were red on that PR and it merged anyway. That is a settings problem, not a tooling problem.

## What it breaks

Nothing found. `type-check` now covers 13 more files, so a future docs-web type error fails everyone's `validate` — intended, and the tree is clean today. `astro sync` is offline and deterministic (all-local assets, no remote loaders, no remote images in `astro.config.mjs`), so it doesn't introduce network dependence into the type gate. Runtime packages, shipped binaries, and Docker content are untouched.

## Validation

| Check | Result |
|---|---|
| `bun run type-check` (root, incl. docs-web) | PASS — exit 0, 11/11 packages |
| `bun run build:docs` (clean checkout) | PASS — 85 pages, exit 0 |
| `lint-marketplace.ts` | PASS — 14/14 entries, all SHAs resolve |
| `check:bundled-skill` / `check:bundled-schema` / `check:pi-vendor-map` | PASS (the `validate` gates `test.yml` doesn't run) |
| CI `test (ubuntu-latest)` / `test (windows-latest)` | PASS |
| CI `Build docs site` (new job, self-testing) | **PASS — 40s** |
| CI `docker-build`, `Validate marketplace entries`, CodeRabbit | PASS |
| CI `Run marketplace auto-review` | FAIL — `billing_error: Credit balance is too low` |

The auto-review failure is pre-existing and unrelated: it has failed on **every** run since 2026-07-20 (last success 2026-07-11), including Release 0.6.0 and #1687 itself. Not counted against this PR.

## Strengths

- **The root-cause work is the best part of this PR.** Diagnosing the `EXIT` failure as a hoisted-`node_modules` resolution artifact, proving it by plant/remove, and then *declining to pin the dependency* is exactly right. The tempting fix would have baked a local environment quirk into the lockfile forever.
- **Fixed the class, not the instance** — every package in the monorepo now participates in the type gate.
- **Routed the guard through the job people already watch** rather than adding a fourth advisory workflow that could be ignored the same way the last one was.
- Deviation stated up front, with the reasoning and an explicit offer to reverse it. That's what made this reviewable in one pass.

Wirasm:
## Confirmation pass on `df728f1c` (S2 fix)

Checked the delta only, per scope. **All three warnings are gone, and no warning was swapped in.** One non-blocking consistency point on the `dotenv` choice, below.

### 1. The warnings are actually gone — verified two ways

Ran in a **sibling** worktree (no up-tree `node_modules`, so no hoisted `unist-util-visit@2.0.3`):

| Command | Result |
|---|---|
| `bun --filter @archon/docs-web type-check` (after `rm -rf .astro`) | exit 0, **0 `[WARN]`** — 4 lines of output total (was **3 `[WARN]`** on `0d1dc7f0`, same command, same precondition) |
| `bun run build:docs` (fresh checkout) | exit 0, 85 pages, **0 `[WARN]`** |
| `bun run build:docs` again after `rm -rf .astro dist node_modules/.astro node_modules/.vite` | exit 0, 85 pages, **0 `[WARN]`** |

The type-check row is the direct A/B and the point of S2: 3 → 0, so the `validate` path is now silent.

I ran the third row because the warnings are emitted during content sync, and a build that follows an earlier sync can inherit cached state — I wanted the zero to be real rather than an artifact of a warm cache. It reproduces cold.

On the specific risk you raised — `txt` and `dotenv` both had to be bundled or you'd have traded one warning for another:

- **`dotenv` is grammar-backed.** Present in `@shikijs/langs` in *both* resolved versions (3.23.0 and 4.0.2). It genuinely highlights.
- **`txt` is not in the grammar bundle at all** — and neither is `text`, which this repo already uses once without warning. They're Shiki's special plain-text names, handled before grammar lookup, so they render unhighlighted and warn-free. That's why `txt` is safe, though not for the reason one would assume. The zero-warning build is the proof.

### 2. No other unsupported fence tags — checked three ways

This was the check that mattered most, so I didn't rely on grep alone:

- **Empirical (authoritative):** the build processes all 85 pages and expressive-code warns *per occurrence*, not deduplicated — proven on the previous head, where `caddyfile` produced exactly 2 warnings for its 2 occurrences. A full build emitting **zero** warnings therefore means no unsupported tag survives anywhere in built content.
- **Static cross-check:** enumerated every fence tag under `src/content/docs/` — 13 distinct: `bash` (355), `yaml` (173), `ini` (62), `typescript` (50), `markdown` (28), `powershell` (16), `json` (9), `dotenv` (3), `txt` (2), `sql` (2), `text` (1), `sh` (1), `cmd` (1). All 13 resolve — 11 grammar-backed in both shiki bundles, plus `txt`/`text` as plain-text names. `caddyfile` and `env` no longer appear.
- **Escape check:** no `.md`/`.mdx` outside `src/content/docs/`, no `<Code>` components, no fences in `.astro` files. Nothing bypasses the build, so the empirical result is complete rather than merely broad.

### 3. `caddyfile` → `txt`: agreed, and there was no better option

I checked whether a Caddy grammar exists rather than taking the premise. **Shiki has no `caddy` or `caddyfile` language in any form** — the real choice was `txt`, `nginx`, or `apache`. Your reasoning holds and I'd have made the same call: `nginx` would tokenize Caddy's whitespace-delimited directives and `@matcher` syntax wrong, and mis-highlighting reads as a bug in the docs rather than an absence of one.

### 4. `env` → `dotenv`: one correction

Your cross-file precedent is real — `github-app-setup.md` does tag identical content `dotenv` at lines 84 and 101. But the same-file precedent points the other way, and it's stronger:

`docker.md` contains **15 `ini` blocks and exactly one `dotenv`** — the new one. And 145 lines below the block you changed:

```
docker.md:131  ```dotenv          docker.md:276  ```ini
DATABASE_URL=postgresql://…       DATABASE_URL=postgresql://…
```

Those two content lines are **byte-identical** and are now tagged differently within a single file. The surrounding `ini` blocks (`:174`, `:276`, `:375`, …) are all `.env` content too, so "follows in-repo convention" doesn't quite hold here — it follows the convention of a *different* file while diverging from this one's.

Not blocking: both render, both are warning-free, and `dotenv` is arguably the more precise grammar for a `.env` fragment. But the minimal-variance fix is `ini` at `:131`, matching its own file and its byte-identical twin. Normalizing the other ~15 the other way would be a separate change, and out of scope here. Your call — I'd take the one-token edit.

### 5. Scope and CI

PR is 5 files, +50/-4 — the original three plus the two docs files. No scope creep; the delta commit touches only the three fence tags.

CI on `df728f1c`, all settled:

| Check | Result |
|---|---|
| **Build docs site** | **pass** (42s) |
| test (ubuntu-latest) | pass (3m3s) |
| test (windows-latest) | pass (5m34s) |
| docker-build | pass (4m16s) |
| Validate marketplace entries | pass (19s) |
| CodeRabbit | pass |
| Run marketplace auto-review | fail — `billing_error`, pre-existing since 2026-07-20, unrelated |

Verdict unchanged: **APPROVE**. S2 is closed. The `dotenv`/`ini` point in §4 is a one-token consistency nit, not a blocker.

Issue #2240, opened 2026-07-21
test(core): 'should abandon a running run' intermittently times out on windows-latest (3 hits in one day, fully mocked test)


Wirasm:
Evidence record (same spirit as #2186): the CommandHandler test `/workflow abandon > should abandon a running run` (`packages/core/src/handlers/command-handler.test.ts:1755`) intermittently exceeds its 5s timeout on **windows-latest only**, on PRs that don't touch `@archon/core`.

## Occurrences (all 2026-07-21)

| PR | Run | Measured | Outcome |
|----|-----|----------|---------|
| #2209 (server/discord only) | 29812832467, attempt 1 | 10.9s | rerun → failed again |
| #2209 | attempt 2 | 10.9s | 3rd rerun → pass |
| #2237 (scripts/workflows only) | 29845532300 | 8.2s | rerun pending |

Meanwhile the identical test passed on many other Windows runs the same day (#2210, #2216, #2217, #2220-2222, #2228-2235…), so it's environmental, not deterministic.

## Why it's odd

The test is fully mocked (mockGetWorkflowRun / mockCancelWorkflowRun; no real DB, no subprocess) — nothing in it should take 8-11 seconds. The neighboring tests in the same describe block pass in milliseconds on the same failing runs. Suspects worth checking when someone picks this up: something in the abandon handler path doing real work despite mocks (timer? dynamic import contention when the full core batch runs?), Windows runner CPU starvation at batch boundaries, or a missing mock letting a retry/backoff path spin.

## Options

1. Cheap: bump this test's timeout (masks it; fine as a stopgap given the base rate).
2. Right: profile what the handler actually awaits in the mocked path and pin it.

Filed so reruns stop being tribal knowledge — link future occurrences here.

Wirasm:
## Root-caused — the "fully mocked" test was creating a real SQLite database

Partly fixed in #2303 (leaving this open, see the end).

The suspicion in the report was right: this test *was* doing real work despite the mocks. The mechanism is a Bun behaviour that contradicts what this codebase assumed.

**`mock.module()` MERGES over the real module — it does not replace the namespace.** An export omitted from the factory keeps its REAL implementation. Minimal repro on bun 1.3.11:

```ts
// real.ts exports alpha and beta
mock.module('./real.ts', () => ({ alpha: () => 'MOCK alpha' }));
// alpha = MOCK alpha
// beta  = REAL beta     <-- still the real function
```

[...]

Issue #2186, opened 2026-07-20
test(adapters): GitHubAdapter self-filtering suite intermittently times out on CI (identical cross-OS, clean on rerun)


Wirasm:
## Summary

The `GitHubAdapter > self-filtering` test group intermittently times out on CI — and when it does, it looks **deterministic within the run**: the same 5 tests hit the 5000 ms timeout on BOTH ubuntu and windows in the same run, then pass untouched on a plain re-run. That pattern (identical cross-OS failure, clean rerun) suggests an order-dependent or shared-state hang, not random slowness.

## Evidence (2026-07-20, PR #2179, run 29752295269 attempt 1)

Failed identically on both OSes, each at exactly ~5001 ms:

- `attribution falls back to sender.login when comment.user is absent`
- `attribution prefers comment.user.login over sender.login when both present`
- `handleWebhook never throws when identity resolution fails`
- `should NOT filter comments from real users`
- `should handle missing comment.user gracefully`

Logs:
- ubuntu: https://github.com/coleam00/Archon/actions/runs/29752295269/job/88385774445
- windows: https://github.com/coleam00/Archon/actions/runs/29752295269/job/88385774417

Attempt 2 (rerun failed jobs, no code change): all green. The PR's diff (workflows validator tests + a providers barrel-export removal) does not touch `@archon/adapters`, and adapters tests run in their own process under `bun --filter`.

## Suspicion to check first

All 5 timeouts are in the self-filtering group, which resolves identity — look for an awaited identity/DB/mock path that can hang when some earlier test in the same file leaves a mock or timer in a bad state (Bun `mock.module()` is process-global and irreversible; see the mock-isolation rules in CLAUDE.md). A test-order shuffle locally (`bun test --rerun-each` / seeded order) may reproduce.

## Ask

Root-cause and fix, or at minimum isolate the group into its own test invocation so a hang can't poison unrelated PR CI. Low priority until it recurs — this is an evidence record so the next occurrence has a trail.

Wirasm:
[...]

PR #2307, opened 2026-07-29
test(adapters): stop github adapter tests writing to a real ARCHON_HOME

### LINKED ISSUE #2305: test(adapters): github adapter tests create a real archon.db, config.yaml and bin/git-credential-archon under ARCHON_HOME
Follow-up from #2303, which fixed the *network* half of this class but deliberately left this alone.

## What happens

Two adapter test files reach a real filesystem/database instead of a stub. Probed by running each file with `ARCHON_HOME` pointed at an empty temp dir:

| File | files created under `ARCHON_HOME` |
|---|---|
| `packages/adapters/src/forge/github/adapter.test.ts` | `archon.db`, `archon.db-wal`, `archon.db-shm`, `config.yaml`, `bin/git-credential-archon`, `workspaces/` |
| `packages/adapters/src/forge/github/context.test.ts` | 4 files, same shape |

Verified identical on unmodified `dev`, so it pre-dates #2303 and was not introduced by it. Note the irony that `adapter.test.ts` is the file #2303 edits, and its own header comment already says database modules are mocked "to prevent self-filtering tests from writing phantom records to the real SQLite DB" — that intent is only partly realised.

The sweep was per-file across all 10 packages; only these two adapter files plus `core/db/codebases.test.ts` and `workflows/executor-preamble.test.ts` create anything (the latter two are 1 file each and may well be intentional — worth a look, not assumed broken).

## Why it matters

1. **Unbounded external resource in a unit test.** This is the same class as #2186 and #2240: a missing stub does not fail loudly, it stalls, and the only bound is Bun's 5000 ms per-test timeout. #2240 is the proof that this shape produces multi-second windows-only CI flakes that cost days to attribute.
2. **It writes into the developer's real `~/.archon/`** when `ARCHON_HOME` is not overridden — running the adapters suite mutates actual local state.
3. Under `bun --filter '*' --parallel`, several packages' test processes can contend on the same SQLite file.

## Likely cause

Almost certainly the same mechanism as #2240: `mock.module()` **merges** over the real module rather than replacing it, so any export the factory omits keeps its real implementation. Worth checking which `@archon/core/db/*` exports these files' factories are missing.

## Reproduction

```bash
PROBE=$(mktemp -d)
cd packages/adapters
ARCHON_HOME="$PROBE" bun test src/forge/github/adapter.test.ts >/dev/null 2>&1
find "$PROBE" -type f
```

## Suggested fix

Complete the DB mock surface in both files so nothing reaches `pool`, then re-run the probe and assert zero files. A regression guard analogous to `packages/adapters/src/test/no-network.ts` (added in #2303) could cover the filesystem/DB half, but that is a design question rather than a given.

Related: #2186, #2240, #2303.

Wirasm:
## Summary

- **Problem:** `packages/adapters/src/forge/github/adapter.test.ts` and `context.test.ts` created 7 entries under `$ARCHON_HOME` on every run — `archon.db` (+`-wal`/`-shm`), `config.yaml`, `bin/git-credential-archon` and `workspaces/` — writing into the developer's real `~/.archon`.
- **Why it matters:** an unbounded external resource in a unit test does not fail loudly, it stalls, bounded only by Bun's 5000 ms per-test timeout (the #2186 / #2240 shape). #2303 removed the `try/catch` wrappers that used to swallow an environment-dependent filesystem throw here, so a locked or unwritable file now *fails* the test — concentrated on the Windows runner, where the unexplained timeout cluster (#2306) already lives.
- **What changed:** closed the three `mock.module` merge gaps that let real `@archon/core` functions run (`resolveDefaultAssistant`, `installCredentialHelper`, `handleMessage`), plus `@archon/core/db/users` in `context.test.ts`; completed `createSelfFilterAdapter`'s Octokit stub so it *resolves*; moved the real credential-helper behaviour into a dedicated `@archon/core` test that owns its own `ARCHON_HOME`.
- **What did NOT change:** no production code. Every existing assertion is preserved verbatim — see "Human Verification". No new bunfig/preload guard — deferred and filed as #2308 (see "Risks").

### Root cause

`mock.module()` **merges** over the real module rather than replacing it, so any export a factory omits keeps its real implementation. Three such gaps sat on `handleWebhook()`'s tail:

| Artefact | Left-real function | Where |
|---|---|---|
| `config.yaml` | `resolveDefaultAssistant` (subpath `@archon/core/config/resolve-assistant`) | step 6, `getOrCreateCodebaseForRepo` → `loadGlobalConfig()` **creates** a default config when absent |
| `bin/git-credential-archon` | `installCredentialHelper` | step 8, App-mode clone → `copyFileSync` into `~/.archon/bin/` |
| `archon.db*`, `workspaces/` | `handleMessage` | step 13, real orchestrator → real SQLite |
| `archon.db*` (context.test.ts) | `findOrCreateUserByPlatformIdentity` (subpath `@archon/core/db/users`) | step 5b |

`handleMessage` *was* stubbed — but only inside the `webhook delivery dedup` block, whose `afterAll` restored it before `App mode` ran. That is the single line that let the App-mode webhook test drive the real orchestrator.

### Boundary chosen: close the module boundary **and** complete the Octokit stub

The two are not alternatives; only one of them fixes #2305, but doing it alone would have been unsafe.

- **A complete Octokit stub does not fix the leak.** None of the four artefacts come from Octokit. They come from `@archon/core` functions the factories silently left real.
- **Fixing only the leak would leave the trap armed.** The `createSelfFilterAdapter` note added in #2303 records it: six tests avoided unmocked steps 8-13 only because `repos.get` *rejected*. Making it resolve — the obvious next edit — drops all six into that territory at once.
- So: close the module boundary first (the actual fix), then make the stub resolve (removes the coupling), then verify the two together.

`spyOn` rather than `mock.module` for the `@archon/core` functions, because `context.test.ts` already `mock.module`s that path with a different shape, and two conflicting factories for one path is the pollution CLAUDE.md forbids. The spies live at **file** scope so no nested `afterAll` can shorten their lifetime again.

## UX Journey

Not user-facing (test-only change). Developer-facing flow:

### Before

```
  Developer                bun test (adapters)         Real filesystem / network
  ─────────                ───────────────────         ─────────────────────────
  bun run test ─────────▶  self-filtering block
                           handleWebhook step 7 ─────▶ GET api.github.com  (12 live requests)
                           step 6  resolveDefault... ▶ WRITES ~/.archon/config.yaml
                           step 8  installCredHelper ▶ WRITES ~/.archon/bin/git-credential-archon
                           step 13 handleMessage ────▶ OPENS ~/.archon/archon.db (+wal/shm)
                                                       MKDIR ~/.archon/workspaces/
  green in 2.7s ◀────────  (silently mutated local state)
```

### After

```
  Developer                bun test (adapters)         Real filesystem / network
  ─────────                ───────────────────         ─────────────────────────
  bun run test ─────────▶  self-filtering block
                           step 7  [Octokit stub] ──x  (no request)
                           step 6  [mock]         ──x  (no write)
                           step 8  [spy]          ──x  (no write)
                           step 13 [spy]          ──x  (no DB)
  green in 0.19s ◀───────  *$ARCHON_HOME untouched — 0 entries*
```

### Before

```
  adapter.test.ts ──mock.module──▶ @archon/paths            (partial: merge keeps rest real)
                  ──mock.module──▶ @archon/git
                  ──mock.module──▶ @archon/core/db/conversations
                  ──mock.module──▶ @archon/core/db/codebases
                  ──mock.module──▶ @archon/core/db/users
                  ──spyOn(dedup-scoped)──▶ core.handleMessage   ← restored by afterAll
                                                                  before `App mode`

  GitHubAdapter.handleWebhook()
      step 6  ──▶ resolveDefaultAssistant   ──▶ config-loader ──▶ FS: ~/.archon/config.yaml
      step 7  ──▶ octokit.repos.get         ──▶ NETWORK: api.github.com
      step 8  ──▶ installCredentialHelper   ──▶ FS: ~/.archon/bin/
      step 13 ──▶ handleMessage             ──▶ orchestrator ──▶ DB: ~/.archon/archon.db

  context.test.ts ──mock.module──▶ @archon/core (root only)
      step 5b ──▶ db/users.findOrCreate...  ──▶ DB: ~/.archon/archon.db
      step 6  ──▶ resolveDefaultAssistant   ──▶ FS: ~/.archon/config.yaml
```

### After

```
  adapter.test.ts ──mock.module──▶ @archon/paths
                  ──mock.module──▶ @archon/git
                  ──mock.module──▶ @archon/core/db/conversations
                  ──mock.module──▶ @archon/core/db/codebases
                  ──mock.module──▶ @archon/core/db/users
              === ──mock.module──▶ @archon/core/config/resolve-assistant   [+]
              === ──spyOn(FILE scope)─▶ core.handleMessage                 [~ widened]
              === ──spyOn(FILE scope)─▶ core.installCredentialHelper       [+]
              === ──spyOn(FILE scope)─▶ core.getLinkedIssueNumbers         [+]

  [+] installOctokitStubs()   ← shared by
        createSelfFilterAdapter  [~ was: no stub at all]
        createDedupAdapter       [~ was: its own 3-endpoint literal]
      covers repos.get, issues.listComments, issues.createComment, pulls.get

  GitHubAdapter.handleWebhook()
      step 6  --x--  (stubbed)      step 7  --x--  (stubbed)
      step 8  --x--  (stubbed)      step 13 --x--  (stubbed)

  context.test.ts === ──mock.module──▶ @archon/core/db/users                [+]
                  === ──mock.module──▶ @archon/core/config/resolve-assistant [+]

  [+] packages/core/src/github-auth/credential-helper-install.test.ts
        ──▶ installCredentialHelper (REAL) ──▶ FS: temp ARCHON_HOME it owns + removes
        ──spyOn──▶ git.execFileAsync
```

**Connection inventory**

| `adapter.test.ts` | `@archon/core/config/resolve-assistant` | **new** | `mock.module` — stops the `config.yaml` write |
| `adapter.test.ts` | `core.installCredentialHelper` | **new** | file-scope `spyOn` — stops the `bin/` write |
| `adapter.test.ts` | `core.getLinkedIssueNumbers` | **new** | file-scope `spyOn` — closes the step-10 PR branch |
| `adapter.test.ts` | `core.handleMessage` | **modified** | spy widened from `webhook delivery dedup` to file scope |
| `createSelfFilterAdapter` | private `adapter.octokit` | **new** | previously unstubbed; now the full resolving surface |
| `createDedupAdapter` | `installOctokitStubs` | **modified** | shared factory replaces its local `DedupOctokitStubs` literal |
| `adapter.test.ts` App-mode test | `mockExecFileAsync` | **removed** | proxy assertion replaced by a direct spy assertion |
| `context.test.ts` | `@archon/core/db/users` | **new** | `mock.module` — stops the SQLite open |
| `context.test.ts` | `@archon/core/config/resolve-assistant` | **new** | `mock.module` — stops the `config.yaml` write |
| `credential-helper-install.test.ts` | `installCredentialHelper` (real) | **new** | direct coverage at the owning layer |
| `credential-helper-install.test.ts` | `git.execFileAsync` | **new** | `spyOn`, reversible |
| `packages/core/package.json` `test` | new test file | **new** | added to the batch chain (core lists files explicitly) |
| production modules | anything | unchanged | no production file touched |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: M`
- Scope: `tests`, `adapters`, `core`
- Module: `adapters:github`, `core:github-auth`

## Change Metadata

- Change type: `bug`
- Primary scope: `adapters`

## Linked Issue

- Closes #2305
- Related #2186, #2240, #2303, #2306
- Follow-up filed: #2308 (filesystem/ARCHON_HOME preload guard, blocked on #2303's `bunfig.toml`)

```bash
bun run validate     # exit 0
```

CI history: first push — ubuntu, docker-build and CodeRabbit passed, `test (windows-latest)` **failed** on the POSIX exec-bit assertion. Second push (`6ba3cebb`) — **all four green, windows included** (5m41s). Third push (`364e1604`) is the rebase onto merged #2303; re-validated locally (see C1) and re-running in CI.

**Leak probe — the same method that proved the bug exists.** `ARCHON_HOME` pointed at a fresh empty temp dir, whole `@archon/adapters` suite (`bun run test`, package cwd), then `find $ARCHON_HOME`:

| | entries created |
|---|---|
| `origin/dev` (baseline, this branch with the two test files stashed) | **7** — `bin/`, `bin/git-credential-archon`, `config.yaml`, `archon.db`, `archon.db-wal`, `archon.db-shm`, `workspaces/` |
| this branch | **0** |

Per file, before → after: `adapter.test.ts` 5 files → 0; `context.test.ts` 4 files → 0. The other two adapter files (`gitea`, `gitlab`) were the control — 0 before and after.

The new `@archon/core` test *does* write, and was probed the same way: it creates its own `mkdtemp` `ARCHON_HOME`, so an ambient probe `ARCHON_HOME` still shows **0** entries.

**Each stub proven load-bearing** — disabled one at a time, leak re-probed:

| Stub disabled | Leak that returned |
|---|---|
| `resolveDefaultAssistant` (adapter.test.ts) | `config.yaml` |
| `installCredentialHelper` (call-through) | `bin/git-credential-archon` |
| `handleMessage` (call-through) | `archon.db*`, `workspaces/`, `config.yaml` |
| `db/users` (context.test.ts) | `archon.db*` |
| `resolve-assistant` (context.test.ts) | `config.yaml` |

**Mutation testing — 14 mutants, 13 killed.** Each mutation applied to *production* source, the named test run, source restored:

| # | Mutation | Test(s) that failed |
|---|---|---|
| M1 | `attributedLogin` → `senderUsername` only | attribution prefers `comment.user.login` |
| M2 | `attributedLogin` → `comment.user.login` only | attribution falls back to `sender.login` |
| M3 | drop the identity-resolution `try/catch` | `handleWebhook` never throws when identity resolution fails |
| M4 | self-filter loses `.toLowerCase()` | case-insensitive username matching |
| M5 | bot-marker check disabled | ignores comments containing bot marker |
| M6 | filter made over-broad (drops everything) | **all six** self-filtering tests |
| M7 | `installCredentialHelper` never invoked | credential helper attempted after App-mode clone |
| M8 | `auth.kind === 'app'` guard dropped | credential helper NOT attempted in PAT mode (new) |
| M9 | `existsSync` idempotence guard removed | helper is idempotent (new core test) |
| M10 | `chmodSync` removed | **SURVIVED** — see below |
| M11 | `return failed` → `throw` | returns failed instead of throwing (new core test) |
| M12 | wrong git config key | copies-and-registers + idempotent (new core test) |
| M13 | delivery dedup disabled | 5 of the 8 `webhook delivery dedup` tests — the block did not go slack when its local spy was replaced by the file-scope one |
| M14 | clone branch never entered (`directoryExists = true`) | **both** App-mode credential-helper tests — this is the vacuity control, see below |

**M10 survives and is documented in the test.** `copyFileSync` preserves the source script's `0755`, so the exec bit is present with or without the explicit `chmodSync`. This is genuinely untestable rather than merely awkward: `umask` does not help, because libuv chmods the destination to the source's mode regardless of it. The assertion is kept because "the installed helper is executable" is a real invariant, and the comment states plainly that it pins the outcome, not the mechanism.

**One branch is deliberately left uncovered**, disclosed in the test file's header: the `skipped` discriminant (`reason: 'source-script-not-on-disk'`, the compiled-binary path). Reaching it needs `existsSync(sourceScriptPath())` to be false, and `sourceScriptPath()` resolves from `import.meta.dir` with no injection seam. The only routes in are `mock.module('node:fs', …)` — process-global and irreversible in Bun, which would poison every other test in this package's batch — or a production parameter added purely for testability, which this PR's "no production code" boundary rules out. Not worth either for a branch whose entire effect is which log line the caller emits.

## Security Impact (required)

- New external network calls? **No** — strictly fewer: `adapter.test.ts` stopped issuing 12 live requests to `api.github.com` per run.
- Secrets/tokens handling changed? **No** — no production code touched. `installCredentialHelper` is unchanged; it is now stubbed in the adapter test and covered directly in `@archon/core`.
- File system access scope changed? **Yes, narrowed.** The adapters suite no longer writes to `$ARCHON_HOME`; the one legitimate write (the credential-helper copy) now happens only inside a `mkdtemp` directory the new core test creates and removes.

## Compatibility / Migration

- Backward compatible? **Yes** — test-only.
- Config/env changes? **No.**

**Verified scenarios**

- Baseline reproduced first, on this branch with the two test files stashed: 7 entries. Then re-probed with them restored: 0. Same command, same probe script.
- Bisected the baseline leak per describe block to attribute each artefact: `self-filtering` → `config.yaml`; `webhook delivery dedup` → `config.yaml`; `App mode` → all 7; and within `App mode`, `primes the lookup cache` alone → all 7, `credential helper install is attempted` alone → `bin/` only.
- Confirmed the leak is not an import side effect: a throwaway test file that only imports `./adapter` creates 0 entries.

**Did any of the six self-filtering tests change what they assert? No — zero assertion edits.** Diff for each is the `try/catch` wrapper collapsing to a bare `await`:

| Test | Assertion | Why untouched |
|---|---|---|
| attribution falls back to `sender.login` | `calls[0]` equals `['github','user123','user123']` | step 5b, runs before any stubbed call |
| attribution prefers `comment.user.login` | `calls[0]` equals `['github','reviewer-alice','reviewer-alice']` | same |
| never throws when identity resolution fails | `mockGetOrCreateConversation` called | step 5, before step 7; the "never throws" property is now *stronger* — the flow runs to completion instead of stopping at step 7 |
| should NOT filter comments from real users | `mockGetOrCreateConversation` called | same |
| should process comments without bot marker | `mockGetOrCreateConversation` called | same |
| should handle missing `comment.user` gracefully | `mockGetOrCreateConversation` called | same |

The three tests that assert `acquireLock` was **not** called return at step 4 and never reach Octokit at all — unaffected in either direction. M6 (over-broad filter) fails exactly the six and none of the three, which is the check that they are still measuring what their names claim.

**Is the "six tests depend on a stub rejecting" coupling removed or relocated? Removed.** The dependency was: *nothing here throws only because step 7's catch returns before steps 8-13*. Now every one of those steps is stubbed independently — the flow runs to completion and the tests pass. Nothing in the block is sensitive to how far execution gets. Two structural guards keep it that way: `installOctokitStubs` covers the whole reachable REST surface (`pulls.get` included, though no current payload takes the PR branch), and the `@archon/core` spies are file-scoped, so the specific way it regressed — a nested `afterAll` restoring a stub mid-file — cannot recur.

**One assertion was deliberately re-pointed, and it is not the six.** `credential helper install is attempted after successful App-mode clone` asserted through `mockExecFileAsync` seeing a `config` argument — a proxy that required running the *real* `installCredentialHelper`, i.e. the `~/.archon/bin/` write this issue is about. It now asserts `installCredentialHelperSpy` was called with the cloned path: more direct, and it cannot pass for the wrong reason. The behaviour it used to reach indirectly is now covered directly by the new `@archon/core` test (copy, exec bit, exact config key, idempotence, non-throwing failure). To make sure stubbing did not create a hole, `credential helper install is NOT attempted in PAT mode` was added and M8 confirms it fails if the `auth.kind === 'app'` guard is deleted.

**Edge cases checked**

- `handleMessage` spy widened to file scope: the dedup block's `expect(handleMessageSpy).toHaveBeenCalledTimes(1)` still passes, and M13 still fails 5 dedup tests — the block did not go slack.
- **Vacuous-pass hazard in the PAT negative, found and fixed (`6ba3cebb`).** `credential helper install is NOT attempted in PAT mode` asserted `not.toHaveBeenCalled()` after driving `ensureRepoReady` at the shared literal `/tmp/clone-test-pat`. If that path exists on a machine, `ensureRepoReady` takes the *sync* branch, never reaches the credential block, and the negative passes for entirely the wrong reason — the same silent-pass shape this PR exists to remove. Both credential-helper tests now use a unique never-created path and assert `mockCloneRepository` was called as a positive control. M14 confirms it: with the clone branch suppressed, both tests fail, and for the PAT one the spy assertion still holds, so the control is provably what does the work.
- `pulls.get` stub added even though no current payload reaches it, so "complete" is a property of the stub rather than of the payloads that happen to be in the file.
- Bogus paths used by the credential-helper tests (`/tmp/clone-test`, `/tmp/clone-test-pat`) confirmed absent, so `ensureRepoReady` takes the clone branch rather than the sync branch.

**What was not verified**

- ~~Windows and CI runners — probed on macOS only.~~ **Measured, and the inference was wrong.** The first push turned `windows-latest` red: `credential-helper-install.test.ts` asserted the POSIX exec bit, and Windows has none — Node reports `0o666` for every regular file, so `mode & 0o111` is unconditionally `0`. Fixed in `6ba3cebb` by moving that assertion into its own `test.skipIf(process.platform === 'win32')`, which puts the platform dependency in the run output instead of hiding it inside an `if`. Local probing still cannot substitute for the Windows runner; CI is the check.
- `bun test` from the repo root (CLAUDE.md's single-file form) was not used; all probes ran with the package as cwd, matching `bun run test`.
- M10 (`chmodSync`) survives, as documented above.

- Affected subsystems/workflows: `@archon/adapters` github tests, `@archon/core` github-auth tests, `packages/core/package.json` test batch chain. No production module.
- **The spy-lifetime bug is an instance, not a class — do not read this PR as fixing it repo-wide.** A sweep of all 245 test files found only 13 nested `beforeAll`/`afterAll` hooks, and this was the only one installing a spy; the rest are env/tmpdir cleanup or spies a sibling `beforeEach` re-installs. One adjacent, pre-existing, differently-shaped case is noted for later at `packages/isolation/src/providers/worktree.test.ts:1722-1741`. Good news for blast radius, but it means the fix does not generalise.
- Potential unintended effects: the six self-filtering tests now execute ~9 more steps of `handleWebhook` per run against stubs. If a future step introduces a new unmocked `@archon/core` call, these tests are where it would first appear — which is the intended trade (visible under stubs) rather than the old one (invisible, live).
- Guardrails/monitoring: the `ARCHON_HOME INVARIANT` header on both files states the rule and the exact re-audit command. `bun run validate` covers the rest.

## Rollback Plan (required)

- Fast rollback: `git revert <merge-commit>` — test-only, no runtime or schema coupling.
- Feature flags: none.
- Observable failure symptoms: a github adapter test failing on a stubbed value, or `packages/core` test batch failing on `credential-helper-install.test.ts`.

## Risks and Mitigations

- **~~Risk (C1): merge conflict with #2303.~~ RESOLVED — #2303 merged, this branch is rebased onto it (`364e1604`), PR is `MERGEABLE`/`CLEAN`.** Recording how it went, since the recipe was written before the merge and is now testable rather than hypothetical:
  - Git produced exactly the two predicted conflict regions in `adapter.test.ts` (the `createSelfFilterAdapter` JSDoc, and its body). The six call-site hunks auto-merged, as predicted — both branches had collapsed them to a bare `await`.
  - Steps 1, 2 and 4 applied as written (take this side).
  - **Step 3 (delete `badCredentials()`) turned out to be automatic, but only by luck of hunk grouping.** Git happened to place the helper inside the HEAD side of the JSDoc conflict, so taking this side dropped it. Had git split the hunk differently it would have been a genuine manual step and lint would have caught it — so the step was worth writing down, and anyone re-doing this resolution should still verify with `grep -n badCredentials` rather than assume.
  - Post-rebase re-verification: `bun run validate` exit 0; full adapters suite still **0** entries under a probe `ARCHON_HOME` — now with #2303's `no-network.ts` guard active, which the suite also passes; M6 still fails exactly the six self-filtering tests; M14 still fails both credential-helper tests.

- **Risk (S1): stubbing `installCredentialHelper` could have deleted coverage.** It was the only test exercising that function.
  - Mitigation: added `packages/core/src/github-auth/credential-helper-install.test.ts`, which covers strictly more (copy, exec bit, exact config key, idempotence, failure-not-throw — M9/M11/M12 all killed), plus the PAT-mode negative in the adapter test (M8). Net coverage increases.
- **Risk (I1): no automated guard, so the leak can return silently.** The issue floats a filesystem analogue of `no-network.ts`; it is not here.
  - Mitigation: deliberate, and now **tracked as #2308** rather than living as a sentence in this description. Such a guard needs `packages/adapters/bunfig.toml`, a file #2303 is adding, and shipping a competing version would create a conflict in the one place a merge is most likely to go wrong. #2308 carries the suggested shape, the `getArchonHome()`/`bunfig.toml`-cwd findings from this work, and the rollout caveat for `core`/`workflows`. The header comments on both touched files document the invariant and the exact probe in the meantime.
- **Risk (I2): `mock.module` merge semantics mean this class of gap can reappear** the moment a new `@archon/core` call is added to `handleWebhook`.
  - Mitigation: not fully solvable without the guard above. The `ARCHON_HOME INVARIANT` headers name the mechanism and the re-audit command so the next person has the same handle this investigation used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of GitHub webhook handling when identity resolution or comment attribution encounters errors.
  * Prevented credential-helper installation failures from causing unhandled errors; failures now return a clear result.

* **Tests**
  * Added coverage for credential-helper installation, Git registration, existing helper files, and executable permissions.
  * Expanded safeguards to ensure tests do not create real configuration, database, or workspace artifacts.
  * Added mocked webhook and GitHub API scenarios for more consistent test results.


Wirasm:
## Automated Code Review

### Summary

Reviewed by two independent agents against `origin/dev...HEAD`. The reviewer re-ran the `ARCHON_HOME` probe itself rather than taking the PR description's word for it, and reached **PASS with no high-confidence issues**. The simplifier found one structural redundancy, which has been applied (`2155df37`).

### Findings

#### Strengths

- **Root cause, not symptom.** Each of the four artefacts is closed at the exact subpath the adapter imports from, not papered over. Confirmed empirically by the reviewer's own independent probe: `adapter.test.ts`, `context.test.ts` and the new core test each run with `ARCHON_HOME` at a throwaway dir and write **zero** files.
- **Comment accuracy verified line-by-line.** Every `step N` reference in the new comments (4, 5, 5a, 5b, 6, 7, 8, 10, 11, 12, 13) was checked against the numbered comments in `adapter.ts:932-1237` and matches.
- **No test got weaker.** The three self-filtering tests asserting "not called" return before Octokit is reached and are unaffected. The three that assert `getOrCreateConversation`/attribution `calls[0]` were already asserting exactly that; dropping the `try/catch` makes them *stricter*, since a real downstream throw now fails instead of being swallowed. `handleWebhook never throws when identity resolution fails` is the clearest case — it previously wrapped the call in the catch that contradicted its own name.
- **Spy-scope move is itself a second bug fix.** The `handleMessage` spy's `afterAll` was scoped to `webhook delivery dedup`, restoring it before `App mode` ran — which is how the App-mode webhook test silently exercised the real orchestrator. File scope closes that; confirmed no cross-file pollution surface, since `adapter.test.ts` and `context.test.ts` already ran as separate `bun test` invocations.
- **Credential-helper coverage went up, not down.** The old assertion forced the real `installCredentialHelper` to run (the #2305 write). The new one asserts the adapter's wiring directly, and `packages/core/src/github-auth/credential-helper-install.test.ts` covers the function's real copy / exec bit / exact config key / idempotence / non-throwing-failure behaviour against a temp `ARCHON_HOME` it owns. Mutants M9/M11/M12 all die there, and the new PAT-mode negative (M8) pins the `auth.kind === 'app'` guard that stubbing could otherwise have hidden.

#### Simplification (applied — `2155df37`)

- `adapter.test.ts` — `createOctokitStubs()` + `installOctokitStubs()` were two functions with two call sites, and *both* call sites paired them immediately. Folded into one function that installs the stubs and returns them. Rule of Three: split it back out when a third caller genuinely needs stubs without an adapter. Re-probed and re-validated after the change (73 pass, 0 entries, `bun run validate` exit 0).

#### Suggestions (not taken)

- Trim the file-header restatement in the `createSelfFilterAdapter` and credential-helper comments. Partially applied to `createSelfFilterAdapter`; the rest kept deliberately — the credential-helper comment's pointer to the new core test file is the non-obvious fact a future reader needs, and the per-file `ARCHON_HOME INVARIANT` headers list *different* gaps (three vs two) and are meant to be readable standalone by someone debugging a stray `~/.archon` write in one file.

#### Known gap (disclosed, not fixed)

- Mutant **M10** (`chmodSync` removed) survives. `copyFileSync` preserves the source script's `0755`, so no assertion can distinguish the two mechanisms without mutating the repo's own script. The exec-bit assertion is kept because it pins a real invariant, and its comment says plainly that it pins the outcome, not the mechanism. 11 of 12 mutants killed.

#### Security

- No production code touched. Strictly fewer external calls (12 live `api.github.com` requests per run removed) and strictly narrower filesystem scope. No secrets handling change.

### Checklist

- [x] Fix addresses the root cause from the investigation (`mock.module` merge semantics), verified by disabling each stub one at a time and watching the exact artefact return
- [x] Code follows codebase patterns (`spyOn` for reversible internal stubs, `mock.module` only for paths no other file in the same invocation mocks)
- [x] Tests cover the change — 12 mutants, 11 killed; the six self-filtering tests fail as a group under an over-broad-filter mutant
- [x] No obvious bugs introduced; `bun run validate` exit 0 after every commit

### Merge-order note

C1 in the PR body still stands: #2303 edits the same `createSelfFilterAdapter` and the same six call sites, with a **rejecting** `repos.get` where this PR uses a **resolving** one. Merge #2303 first, then rebase this branch and take this side for those two regions.

---
*Self-reviewed • ready for human review*

Wirasm:
## Review: #2307 — stop github adapter tests writing to a real ARCHON_HOME

**Verdict: REQUEST CHANGES.** The fix itself is correct and unusually well-evidenced — I reproduced every load-bearing claim independently and all of them held. But the new `@archon/core` test fails on `windows-latest`, which has now settled red. One assertion needs a platform guard; everything else is small.

---

### Settled CI

| Check | Result |
|---|---|
| `test (ubuntu-latest)` | pass (3m0s) |
| `docker-build` | pass (7m40s) |
| CodeRabbit | pass |
| `test (windows-latest)` | **fail (7m48s)** |

---

### C1 — Windows CI is red: the new test's exec-bit assertion cannot hold on Windows

`packages/core/src/github-auth/credential-helper-install.test.ts:64-65`

```ts
const mode = (await stat(helperPath)).mode & 0o777;
expect(mode & 0o111).not.toBe(0);
```

Windows has no POSIX execute bit — Node synthesizes `mode` from the read-only attribute, so a writable file reports `0o666` and `0o666 & 0o111 === 0`. From the runner log:

```
(fail) installCredentialHelper > copies the helper into $ARCHON_HOME/bin and registers it on the worktree
  at packages\core\src\github-auth\credential-helper-install.test.ts:65:30
  Expected: not 0
@archon/core:test | 2 pass  1 fail
error: script "test" exited with code 1
```

The other two tests in the file pass on Windows; only this assertion is platform-dependent.

This is the one place the PR's own risk analysis pointed at and then got wrong. "What was not verified" says Windows was inference, not measurement, and that "the Windows exposure should strictly decrease" — the disclosure was right, the inference wasn't. Worth naming because #2306 (the Windows timeout cluster) is part of this PR's stated motivation, and as it stands the PR converts a green Windows run into a red one.

**Fix** — keep the invariant, scope it to the platform where it means something:

```ts
// Windows has no POSIX exec bit; `mode` there reflects the read-only attribute only.
if (process.platform !== 'win32') {
  const mode = (await stat(helperPath)).mode & 0o777;
  expect(mode & 0o111).not.toBe(0);
}
```

Related: this also bounds the M10 discussion. The surviving mutant is disclosed as pinning "the outcome, not the mechanism", but on Windows the outcome isn't observable at all, so the comment should say the invariant is POSIX-only.

On whether M10 is *genuinely* untestable rather than merely awkward — I checked the obvious escape hatch and it does not exist. `umask` doesn't help: libuv's `copyfile` chmods the destination to the source mode after creating it, so the copy yields `0755` regardless.

```
umask 0o111, copy only  -> 755
umask 0o111, copy+chmod -> 755
default umask, copy only -> 755
```

`scripts/git-credential-archon.sh` is tracked as `100755`, so the only remaining kills are mutating a tracked file's mode mid-test or spying on `chmodSync` (a mechanism assertion). The disclosure is honest and it's in the right place — the test comment, where the next reader will hit it, with the PR body as backup. Agreed as-is.

---

### I1 — the #2303 rebase will fail lint unless `badCredentials()` is deleted too

The merge order is right — #2303 first, then rebase this and take **this** side for `createSelfFilterAdapter`. I diffed both branches against `dev` and the overlap really is confined to two regions:

1. `createSelfFilterAdapter` — #2303 adds a **rejecting** stub, this adds a **resolving** one via `installOctokitStubs`.
2. The six `try/catch` → bare `await` call sites — **byte-identical** in both branches, including the added `// Deliberately NOT wrapped in try/catch` comment. No resolution choice to make there.

Nothing else in #2303 (`bunfig.toml`, `no-network.ts`, gitea, `command-handler`, CLAUDE.md) is touched here.

Two things the rebase instructions in the PR body don't cover:

- #2303 introduces a `badCredentials()` helper used **only** inside its version of `createSelfFilterAdapter`'s stub. Taking this PR's version of that function leaves it defined and unreferenced. `eslint.config.mjs:57` sets `@typescript-eslint/no-unused-vars: 'error'` with only a `^_` escape, and CI runs `--max-warnings 0` — so the rebase fails lint unless `badCredentials()` is deleted in the same commit.
- Take this side for the **JSDoc**, not just the body. #2303's `COUPLING — why these tests can safely await handleWebhook unwrapped` block asserts "Nothing throws here only because the stubbed 401 lands in step 7's catch." Once the stub resolves, that paragraph is false. This PR already replaces it with the correct one; just make sure the resolution doesn't keep both.

---

### Verification I ran independently

Everything below is my own measurement, not a re-reading of the PR body.

**The 7 → 0 probe reproduces, and it would catch a regression.** Ran the whole `@archon/adapters` suite (`bun run test`, package cwd, `ARCHON_HOME` at a fresh `mktemp -d`, `DATABASE_URL` unset), then `find "$ARCHON_HOME" -mindepth 1` — which includes dotfiles and recurses, and did in fact surface `bin/git-credential-archon` one level down.

| | entries |
|---|---|
| `dev` (the two test files reverted) | **7** — `archon.db`, `archon.db-shm`, `archon.db-wal`, `bin/`, `bin/git-credential-archon`, `config.yaml`, `workspaces/` |
| this branch | **0** |

Its blind spot is scope, not depth: it only sees writes rooted at `$ARCHON_HOME`. That's the right scope for #2305.

**Each stub is load-bearing.** Disabled one at a time (call-through for the spies, factory removed for the mock), same probe, `adapter.test.ts` only:

| Disabled | Entries returned | Tests |
|---|---|---|
| `handleMessage` | `archon.db`, `archon.db-shm`, `archon.db-wal`, `config.yaml`, `workspaces/` | 66 pass / **7 fail** |
| `installCredentialHelper` | `bin/`, `bin/git-credential-archon` | 73 pass |
| `resolveDefaultAssistant` | `config.yaml` | 73 pass |
| none (baseline) | — | 73 pass |

**The coupling is removed, not relocated — yes.** This was the claim worth testing directly rather than reasoning about, so I added a temporary assertion to `should NOT filter comments from real users`:

```ts
expect(handleMessageSpy).toHaveBeenCalled(); // does the flow reach step 13?
```

73 pass / 0 fail, 0 entries. The self-filtering payloads genuinely run `handleWebhook()` end to end under stubs — step 7 `repos.get` resolves, step 8 `ensureRepoReady` takes the clone branch (all of `ensureProjectStructure`, `cloneRepository`, `addSafeDirectory` mocked; PAT mode skips the helper), step 10 `getLinkedIssueNumbers` spied, step 12 `listComments` stubbed, step 13 `handleMessage` spied. Nothing depends on where execution stops, because it doesn't stop. No replacement short-circuit was introduced.

The trade this creates is real and correctly characterised in the PR: the six tests now `await` unwrapped, so a future step that throws fails them. That's the loud-instead-of-silent direction.

**The spy-lifetime bug is real, and the fix widens the window rather than moving it.** On `dev`: `spyOn(core, 'handleMessage')` at `:562` inside `describe('webhook delivery dedup')` (`:480`), `afterAll` restore at `:565-566`, block closes before `:698`; `describe('App mode')` opens at `:1331`. On this branch the hooks are at `:196`/`:206` — outside `describe('GitHubAdapter')` (`:287`), i.e. genuinely file-terminal, and each `bun test` invocation owns one file so nothing leaks sideways.

**It is an instance, not a class.** I had the repo swept for the exact shape — a `spyOn` installed in a nested `describe`'s `beforeAll`, restored by that block's `afterAll`, with later siblings reaching the same function. Across all 245 `packages/*/src/**/*.test.ts` files there are only 13 nested `beforeAll`/`afterAll` hooks repo-wide, and this was the only one that installed a spy. The rest are env-var save/restore, tmpdir cleanup, or spies re-installed by every sibling's `beforeEach` (`clone.test.ts`, `command-handler.test.ts`) or restored per-test in `try/finally`.

One adjacent case is worth a look sometime, though it isn't this pattern and isn't caused by it: `packages/isolation/src/providers/worktree.test.ts:1722-1741` spies `worktreeCopy.copyWorktreeFiles` inside `describe('file copying')`, and later siblings (`workspace sync…` `:2264`, `cross-platform path handling` `:2534`, `custom remote support` `:3049`) call `provider.create()` un-spied — which always reaches the real function, since `defaultCopyFiles = ['.archon']` means the empty-list early return never fires. Blast radius is bounded (the file's `node:fs/promises` mock omits `stat`, so it TypeErrors into a swallowing catch) and `describe('create')` at `:261` *precedes* the spy block and is equally un-spied, so it's a pre-existing assertion gap, not a lifetime bug. Separate issue if anything.

**Assertions really were preserved.** The PR's `grep -E "^[+-]\s*expect\("` proof has a hole — a multi-line `expect(...)` whose matcher changed on a continuation line would leave the `expect(` line as unchanged context and never appear. I re-ran it with a grep that matches matcher tokens anywhere on a changed line:

```
git diff origin/dev...HEAD -- packages/adapters/ \
  | grep -E "^[+-].*(toBe|toHave|toEqual|toContain|toThrow|toMatch|toBeGreater|toBeLess|\.not\.|expect\()"
```

Same three lines, out of 144 changed non-comment lines. No matcher was edited on a continuation line, and a moved assertion would have shown as a `-`/`+` pair. The proof holds under the stronger form.

**Mutation spot-checks.** M7 and M8 killed, M10 survives — matching the PR's table:

| Mutation | Result |
|---|---|
| M7 — `installCredentialHelper` never invoked | `(fail) credential helper install is attempted after successful App-mode clone` |
| M8 — `this.auth.kind === 'app'` guard dropped (`adapter.ts:736`) | `(fail) credential helper install is NOT attempted in PAT mode` |
| M10 — `chmodSync` removed | 3 pass — **survives**, as disclosed |

M8 mattering is the answer to "did stubbing `installCredentialHelper` open a hole?" — it didn't. Coverage genuinely increased rather than relocating with a gap: the old assertion (`mockExecFileAsync` saw a `config` arg) was a weak proxy that could be satisfied by any `git … config …` call from anywhere in the flow, and it only held because the real helper wrote into `~/.archon/bin/`. The new pair asserts the exact function with the exact path, plus the guard's negative, plus the helper's own copy / exec bit / config key / idempotence / failure-not-throw in `@archon/core`.

**`bun run validate`** — exit 0 locally on macOS.

---

### Suggestions

- **S1 — the `skipped` branch is still untested.** `installCredentialHelper` returns three discriminants; the new test covers `installed` and `failed`. `{ kind: 'skipped', reason: 'source-script-not-on-disk' }` is the compiled-binary path and the adapter switches on it (`adapter.ts:747-753`). Cheap to add now that the file owns its own `ARCHON_HOME` — point `import.meta.dir`-relative resolution at a home with no `scripts/`, or assert it via a temporarily-renamed source. Not a regression; the old test didn't cover it either.

- **S2 — `/tmp/clone-test-pat` and `/tmp/clone-test` are shared-namespace paths.** Both tests depend on the path *not existing* so `ensureRepoReady` takes the clone branch. If it exists, the App-mode positive test fails loudly (fine) but the PAT negative test passes **vacuously** (silent) — the same silent-pass family this PR is fixing. `join(await mkdtemp(...), 'repo')` gives a guaranteed-absent path.

- **S3 — the PR body's architecture diagram is stale.** It lists `createOctokitStubs() / installOctokitStubs()` and `createDedupAdapter → createOctokitStubs/installOctokitStubs`; commit `2155df37` folded `createOctokitStubs` away and there are no references left in the tree. The Edge-cases section also cites "M13 (dedup disabled)" while the header and table say 12 mutants, M1–M12.

- **S4 — `mockExecFileAsync.mockClear()` at `adapter.test.ts:1517` is now vestigial.** No assertion reads it any more; the mock itself is still needed on the `@archon/git` namespace, which the rewritten comment at `:141-145` explains correctly. Harmless, just dead motion in the `beforeEach`.

- **S5 — file the filesystem-guard follow-up now.** Deferring it is the right call: it needs `packages/adapters/bunfig.toml`, which #2303 is adding, and a competing version would conflict in the worst possible place. But "clean follow-up once #2303 lands" only survives as an issue, not as a PR-body sentence — especially given I2's own point that the merge semantics let this class of gap reappear the moment a new `@archon/core` call joins `handleWebhook`. Open it against #2305 / #2303 before this merges.

---

### Strengths

- The diagnosis is the valuable part and it is right: none of the four artefacts came from Octokit, three of the four leaks were on **subpaths** a root `@archon/core` factory structurally cannot cover, and `mock.module()`'s merge semantics is what made all of it silent. That's a genuinely non-obvious root cause, and it generalises.
- Doing the module boundary *and* the resolving stub together, with the stated reason that either alone is unsafe, is the correct call — a resolving stub on top of the old gaps would have dropped six tests into unmocked territory at once.
- File-scope spies with the regression mechanism named in the comment. The next person editing this file gets the handle rather than the trap.
- Every claim I checked was reproducible from the PR body's own instructions, including the ones that turned out to constrain the fix (M10). That is rare and it made this review much faster.

Wirasm:
## Review response — `6ba3cebb`

All blocking and non-blocking points addressed. The headline one is the most instructive.

### C1 (blocking) — Windows exec bit — **fixed**

`credential-helper-install.test.ts` asserted `mode & 0o111 !== 0`. Windows has no POSIX execute bit; Node reports `0o666` for every regular file, so the expression is unconditionally `0` and `windows-latest` went red at exactly that line.

Moved into its own `test.skipIf(process.platform === 'win32')('installed helper is executable (POSIX only)')` rather than wrapping it in an `if` — the platform dependency now shows up in the run output instead of silently reducing a longer test's assertion count on one platform.

**The disclosure was right and the inference was wrong**, and the PR body now says so in those terms. "Removes filesystem writes rather than adding any, so Windows exposure should strictly decrease" was sound reasoning about the *leak* and simply did not cover the *new test* the change introduced. The section that flagged it as inference has been rewritten as a measured result rather than quietly deleted.

### `/tmp/clone-test-pat` vacuity — **fixed, and it was real**

Correct catch, and the sharper half of the review. `credential helper install is NOT attempted in PAT mode` asserted `not.toHaveBeenCalled()` after driving `ensureRepoReady` at a shared literal path. Had that path existed, `ensureRepoReady` would take the *sync* branch, never reach the credential block, and the negative would pass for entirely the wrong reason — a silent pass shipped inside the PR whose purpose is removing silent passes.

Two changes, because the unique path alone would only reduce the probability:

- `unclonedPath()` — `join(tmpdir(), 'archon-clone-test-<uuid>')`, unique per call, never created.
- A **positive control** in both credential-helper tests: `expect(mockCloneRepository).toHaveBeenCalled()`, asserted *before* the negative. The negative can now only be satisfied by the `auth.kind === 'app'` guard.

Verified by mutation rather than by reading. **M14** (`let directoryExists = false` → `true`, suppressing the clone branch) now fails *both* tests. For the PAT one the spy assertion still holds under that mutation, so the control is provably the assertion doing the work.

### I1 — rebase instructions — **completed**

Both missing steps confirmed independently before writing them down:

- `gh pr diff 2303 | grep -n badCredentials` → defined once (L90), used twice (L125, L128), both inside the `adapter.octokit` literal that step 1 replaces. So it is orphaned by the resolution and `@typescript-eslint/no-unused-vars` is `error` under `--max-warnings 0` — it would fail lint, as you said.
- #2303's "COUPLING" JSDoc paragraph describes the rejecting stub and is false once the stub resolves.

C1 in the PR body is now a numbered four-step recipe with those two called out as the easy-to-miss ones.

### Also fixed

- **Follow-up filed as #2308** — the filesystem/ARCHON_HOME preload guard. Agreed a PR-body sentence is not a tracked item. It carries the suggested shape, the two findings from this work that whoever picks it up would otherwise rediscover (`getArchonHome()` reads the env fresh, so a preload override is honoured; `bunfig.toml` is cwd-only so the guard is inactive from the repo root), and the caveat that `core`/`workflows` each have one file to audit *before* extending the guard rather than assuming they are broken.
- **Stale statements corrected** — the architecture diagram no longer lists `createOctokitStubs()` (folded away in `2155df37`), and the mutation header/table now read 14 mutants / 13 killed with M13 and M14 as actual rows rather than M13 being cited from nowhere.
- **`skipped` discriminant** — *not* fixed, and now explicitly disclosed in the test file header rather than left as an implicit gap. `sourceScriptPath()` resolves from `import.meta.dir` with no injection seam, so the only routes in are `mock.module('node:fs', …)` — process-global and irreversible in Bun, which would poison every other test in this package's batch — or a production parameter added purely for testability, which this PR's no-production-code boundary rules out. Declining for a branch whose entire effect is which log line the caller emits; happy to be overruled, but it should be a deliberate call rather than an oversight.
- **M10** — the PR body now states it is genuinely untestable, not merely awkward, including that `umask` does not help because libuv chmods the destination to the source mode regardless.
- **Spy-lifetime generalisation** — the Blast Radius section now records the 245-file sweep result explicitly, that this is an **instance and not a class**, and the adjacent pre-existing case at `packages/isolation/src/providers/worktree.test.ts:1722-1741`. The PR no longer reads as if it hardened the repo.

### Re-verified after the changes

- Full `@archon/adapters` suite under a probe `ARCHON_HOME` (`find -mindepth 1`): still **0 entries**.
- New core test under the same probe: **0 entries** (it writes only into the `mkdtemp` it owns).
- **M6** re-run: still fails exactly the six self-filtering tests and none of the three that assert `not.toHaveBeenCalled()`.
- `bun run validate`: exit 0.

CI re-running now; `windows-latest` is the one that matters.

Wirasm:
## Rebased onto merged #2303 — `364e1604`

#2303 merged at 15:00 UTC while CI was finishing here, so C1 stopped being a hypothetical and became a test of whether the recipe was actually right. Outcome, since a prediction is only worth something once it is checked:

**Predicted correctly:**

- Exactly two conflict regions in `adapter.test.ts` — the `createSelfFilterAdapter` JSDoc and its body. Nothing else in the file, and nothing in any other file.
- The six call-site hunks auto-merged. Both branches had independently collapsed them to a bare `await`, so git had nothing to ask about.
- Steps 1, 2 and 4 applied exactly as written.

**Predicted, but for a slightly wrong reason — worth flagging:**

Step 3 (delete `badCredentials()`) resolved itself. Git happened to group the helper inside the HEAD side of the JSDoc conflict, so taking this side dropped it without a separate edit. That is luck of hunk boundaries, not a property of the change: had git split it differently, it would have been a real manual step and lint would have caught the orphan. So the step was worth writing down, and anyone re-doing this resolution should confirm with `grep -n badCredentials` rather than assume — which is what I did (`grep` returns nothing, no conflict markers remain).

**Re-verified after the rebase, not assumed:**

- Full `@archon/adapters` suite under a probe `ARCHON_HOME`: still **0 entries** — and now with #2303's `no-network.ts` preload active, which the suite also passes clean. The two changes compose rather than merely coexist: the network guard has nothing to catch because the Octokit surface is fully stubbed, and the filesystem stays untouched because the `@archon/core` boundary is closed.
- **M6** still fails exactly the six self-filtering tests and none of the three `not.toHaveBeenCalled()` ones.
- **M14** still fails both credential-helper tests, so the vacuity control survived the rebase.

PR is back to `MERGEABLE` / `CLEAN`. CI re-running on the rebased head; the previous head was green on all four including `windows-latest` (5m41s).

The PR body's C1 section has been rewritten from an instruction into a record of what actually happened, so nobody follows a recipe for a rebase that is already done.

Wirasm:
## Confirmation pass on the rebase — `364e1604`

**Approved.** C1 is fixed, the resolution is correct rather than merely conflict-free, and the composition claim holds under a test designed to break it. Confirmation scope only; I did not re-review what I cleared last round.

### Settled CI on `364e1604`

| Check | Result |
|---|---|
| `test (ubuntu-latest)` | pass (2m36s) |
| `test (windows-latest)` | **pass (6m11s)** |
| `docker-build` | pass (10m28s) |
| CodeRabbit | pass |

`bun run validate` — exit 0 on the rebased tree locally.

---

### The four resolution checks

**`badCredentials()` is genuinely gone, not lint-silenced.** `grep -rn badCredentials packages/` returns nothing, and no conflict markers survive anywhere in the file. Nothing renamed it or moved it out of reach — the helper and its two call sites left together with the rest of #2303's stub body, which is the right outcome rather than a coincidence that happened to satisfy `--max-warnings 0`.

Your framing of this one is right and worth keeping in the record: it resolved by hunk-boundary luck, not by any property of the change. If git had split the JSDoc hunk one line differently, the helper would have landed on the incoming side and survived as an orphan — caught by lint, but only after a push. `grep` was the correct verification, and "confirm rather than assume" is the transferable part.

**The surviving JSDoc is #2307's.** The block above `createSelfFilterAdapter` is the "full resolving Octokit surface … NOTHING here depends on a call failing to stop the flow early" version, and the body ends in `installOctokitStubs(adapter)`. #2303's `COUPLING — why these tests can safely await handleWebhook unwrapped` paragraph — which asserts the tests are safe *because* the 401 stops the flow at step 7, and which the resolving stub makes false — is absent. So is its `Bad credentials`/401-rejecting stub. The remaining `401` hits in the file (`retry logic`, `lookupCache eviction`, `T3`) are pre-existing tests unrelated to this region.

**The six call sites: identical text, and the semantics come from elsewhere.** This is why the auto-merge was safe rather than lucky. Filtering the diff against merged `dev` for those lines returns **0** — both branches had independently reduced them to the same bare `await`, so there was nothing for git to ask about. The semantic difference between the two branches never lived in the call sites at all; it lived entirely in `createSelfFilterAdapter`, which is exactly the region that *did* conflict and where this side won.

So "coincidental textual match" isn't the risk here — the risk would be the text agreeing while the intent diverged, and I checked that directly rather than inferring it. Re-ran the step-13 probe on the rebased tree (temporary `expect(handleMessageSpy).toHaveBeenCalled()` inside `should NOT filter comments from real users`): **73 pass, 0 fail, 0 `ARCHON_HOME` entries**. The merged result still runs `handleWebhook()` to completion under stubs, which is #2307's intent. #2303's narrower intent — no live request — is not lost, it is now enforced structurally by the preload guard instead of incidentally by a rejecting stub. That is a strictly better arrangement than either branch had alone.

**Nothing from #2303 was dropped.** `packages/adapters/bunfig.toml` and `src/test/no-network.ts` both present; the gitea `globalThis.fetch` save/restore intact at `adapter.test.ts:134-146` with its three call sites; `findChildRuns: mockFindChildRuns` present in the `command-handler.test.ts` factory along with the load-bearing comment explaining why; both CLAUDE.md additions (the `mock.module` merge-semantics note and the `no-network` paragraph) present.

---

### The composition claim — verified, and not true-by-luck

You were right that this is the claim most likely to be true for the wrong reason, so I tested the failure mode rather than the success one.

**E1 — the claim itself.** Full `@archon/adapters` suite, package cwd (so `bunfig.toml` is actually read), fresh `mktemp -d` as `ARCHON_HOME`: exit 0, **0 entries**, **0 network-guard violations**.

**E2 — would the guard still fire if the stub regressed?** Deleted `installOctokitStubs(adapter)` from `createSelfFilterAdapter`, changing nothing else:

```
exit=1   67 pass  6 fail
guard fired: 12 violation records
  - GET  https://api.github.com/repos/testuser/testrepo
  - POST https://api.github.com/repos/testuser/testrepo/issues/42/comments
```

The guard is live and blames precisely. Twelve requests across six tests — two per test, which is exactly the "12 live requests to api.github.com per run" the PR body attributes to the pre-fix state, so the guard's accounting and the original diagnosis agree independently.

That settles the ambiguity: the guard catching nothing here is **redundancy, not breakage**. The two changes are genuinely complementary — #2307 removes the requests, #2303 makes their return a named failure instead of a 5-second flake. Note the guard is also the only one of the two that survives a *future* regression, since it needs no one to remember to re-check.

---

### Mutations on the rebased tree

**M6 (self-filter made over-broad)** — fails exactly the six self-filtering tests that assert a comment is *not* filtered, and none of the three that assert it *is*:

```
(fail) attribution falls back to sender.login when comment.user is absent
(fail) attribution prefers comment.user.login over sender.login when both present
(fail) handleWebhook never throws when identity resolution fails
(fail) should NOT filter comments from real users
(fail) should process comments without bot marker from same user
(fail) should handle missing comment.user gracefully
```

Survivors: `should ignore comments from the bot itself`, `should handle case-insensitive username matching`, `should ignore comments containing bot marker` — the three that expect filtering. Exactly the claimed split. (It also fails 8 dedup tests, which is inherent to mutating shared production code, not a deviation from the claim.)

**M14 (force the sync branch — the vacuity control)** — fails **both** credential-helper tests. Worth stating why that matters: under this mutation the PAT test's own `expect(installCredentialHelperSpy).not.toHaveBeenCalled()` still *passes*, because the sync branch never reaches the credential block. The only thing failing it is `expect(mockCloneRepository).toHaveBeenCalled()`. That is the control doing precisely the job it was added for — before `364e1604` this mutation would have left the negative green.

The S2 fix is a better one than I asked for. I suggested a guaranteed-absent path; the author added the path *and* a positive control asserting the clone branch actually ran, which closes the vacuity hole even on a machine where the path somehow exists. Putting the control *first* in the PAT test, with the comment saying why order matters for a `not.toHaveBeenCalled` assertion, is the detail that makes it read correctly to the next person.

Also good: moving the exec-bit check to its own `test.skipIf(process.platform === 'win32')` rather than an inline `if`. The skip shows in run output, so "this assertion is POSIX-only" stays visible instead of decaying into a silently-dead branch.

---

### Ratifying the `skipped` discriminant — accepted, with one stated reason corrected

**Accept.** Leaving it uncovered is the right call, and having it in the test header is the right place for it.

One of the two reasons given is inaccurate, though, and it's worth fixing so nobody inherits a wrong model of the batch chain. The header says `mock.module('node:fs')` "would poison every other test in this package's batch." It wouldn't — `credential-helper-install.test.ts` has its own `bun test` invocation in `packages/core/package.json`:

```
… && bun test src/github-auth/credential-helper-install.test.ts && …
```

so the blast radius is that one process, containing exactly that one file. The batch-splitting convention already solves the isolation problem.

The conclusion still stands, for a better reason. Even confined to its own process, `node:fs` is a builtin that `@archon/paths`, pino and the transitive dependency tree all use, and `mock.module` merges irreversibly — you'd be stubbing `existsSync` path-discriminatingly (it guards both the helper check and the source check) underneath every one of them. That is a real ongoing hazard traded for a branch whose entire observable effect is which log line the caller emits, with no behavioural divergence: the clone proceeds either way, since the install is non-fatal by design.

Suggest trimming the parenthetical to something like *"(a) `mock.module('node:fs')`, which merges irreversibly over a builtin the whole dependency tree uses"* and dropping the batch-poisoning clause. Doc-only, not a merge blocker — fold it in if you touch the file again.

---

### Verdict

**Approve `364e1604`.** C1 resolved and confirmed by measurement on the platform that caught it. The rebase is correct on all four points, with the one luck-dependent step honestly disclosed rather than presented as a clean win. The composition claim survives an adversarial check. M6 and M14 both still kill, and the S2 fix is stronger than what was asked for.

Issue #2315, opened 2026-07-29
feat(server): declare and enforce an API compatibility contract before there are clients we don't ship


Wirasm:
## Summary

Archon reports its version but never states, or checks, what it is **compatible** with. Today every HTTP client ships with the server, so nothing can skew. That stops being true the moment there is a client Archon does not ship — a native macOS/iOS app, or any third-party integration — and a client users leave running for weeks is the worst case for discovering this late.

The cost of adding a compatibility contract is roughly constant. The cost of retrofitting one onto already-deployed clients is not.

## The concrete failure this is modelled on

This is not hypothetical; it happened today in a sibling tool with the same architecture (a long-running engine plus a separately-installed CLI):

- The engine process started at **10:07**.
- A change to how the drain derives identity — bearer token instead of a `{handle}` body — landed at **10:15**.
- The shipped CLI sent the new form. The running engine still demanded the old one and answered with a **validation error** that read exactly like operator syntax: `handle required`.

Two people (me included) spent real time debugging argument parsing, flag ordering, and environment-variable precedence. All of it was correct. **The wire contract had changed under a long-running process: caller updated, server not.**

Worse, the failure was invisible for **ten hours**, because the calling hook's contract is *"print the result or print nothing, always exit 0."* Every failed call looked like "no mail." It was only noticed because a second, unrelated code path happened to keep working.

## Where Archon stands

**Already there:**
- `GET /api/health` returns `version: appVersion` — the real build version, and unauthenticated (health is exempt from the API gate), so any client can read it before doing anything else.
- `GET /api/openapi.json` is generated from the live Zod routes, so the server already has a machine-readable source of truth for its own surface.

**Missing:**
- The OpenAPI spec declares `info.version: '1.0.0'` as a **hardcoded literal** (`packages/server/src/routes/api.ts:2903`). It has never moved and cannot move on its own. The single field whose job is to express API compatibility currently asserts the same value no matter what the routes are — it is not merely unused, it is actively wrong.
- No client is required to check anything, and no route refuses a client it does not recognise.

## Two skew surfaces, and one is live today

**1. HTTP client ↔ server — not live yet, cheap now.**
Every HTTP client today (web console) is served by the same process it talks to. Real the moment a native client exists.

**2. Two Archon versions sharing one `~/.archon` — split out to #2316.**
[...]

Issue #2314, opened 2026-07-29
fix(providers): resolved model is never recorded — Claude reads `model_usage` but the SDK field is `modelUsage`


Wirasm:
## Summary

Archon never records **which concrete model a run actually used**. It records the string it *sent* (`opus`, `sonnet`, or a tier keyword), never the snapshot the provider resolved to (`claude-opus-5`). Aliases are moving pointers, so any comparison across time is silently invalid: a run attributed to `opus` in June and one in August may be different models, and nothing in the record says so.

For the per-run model override in #1433 — especially its **benchmarking** use case — this is the difference between a number you can trust and one you cannot.

## The data already arrives. We read the wrong field name.

`packages/providers/src/claude/provider.ts:1106` reads `resultMsg.model_usage` (snake_case).

The SDK field is **`modelUsage`** (camelCase) — `node_modules/@anthropic-ai/claude-agent-sdk/sdk.d.ts:4156`, SDK 0.3.209:

```ts
export declare type SDKResultSuccess = {
    type: 'result';
    total_cost_usd: number;
    usage: NonNullableUsage;
    modelUsage: Record<string, ModelUsage>;   // ← camelCase; keys are concrete model ids
    ...
};
```

The guard is `resultMsg.model_usage ? { modelUsage: … } : {}`, so an always-`undefined` read means the spread is always empty. It fails silently and permanently.

**Verified against a live install:** of **178** AI `node_completed` events, **0** carry `model_usage`. The plumbing downstream is fine and complete — `dag-executor.ts:1594` captures it, `:2197` persists it into the `node_completed` event — it has simply never received a value.

Note every *other* field on that result message is snake_case (`total_cost_usd`, `num_turns`, `stop_reason`, `duration_ms`). `modelUsage` is the one exception in the SDK, which is exactly how this slipped past review.

### Root cause is a hand-written duplicate of an SDK type

`provider.ts:994` casts the message to a locally-declared shape instead of the SDK's own `SDKResultMessage`:

```ts
const resultMsg = msg as {
  session_id?: string;
  ...
  model_usage?: Record<string, {...}>;   // ← a field the SDK does not have
};
```

Because the local field is optional, TypeScript cannot flag it — the mismatch is invisible at compile time and produces no runtime error. Using the SDK type directly would have made this a compile error, which is precisely what `CLAUDE.md` § *SDK Type Patterns* prescribes:

> ✅ CORRECT - Import SDK types directly … ❌ AVOID - Defining duplicate types

[...]

PR #2317, opened 2026-07-29
feat(core): record the database schema vintage and state the additive-only migration rule

### LINKED ISSUE #2316: fix(core): the database records no schema version — vintage is unknowable and migration safety is accidental
## Summary

**Archon records no schema version anywhere.** `PRAGMA user_version` is `0`, there is no `schema_version` table, and nothing tracks which migrations have been applied. The schema is converged by running idempotent additive statements on every connection.

That works — until you need to answer "what vintage is this database?", and then nothing can. Not Archon, not the user, not a second process sharing the same file.

Two consequences follow, and the first is **already true in production today**.

## 1. The same Archon version behaves differently depending on when the DB was created

This is not a hypothesis; it is documented in our own code (`packages/core/src/db/adapters/sqlite.ts:398-405`):

> NOTE: NOT NULL constraint changes on existing columns (e.g., `branch_name` in `isolation_environments`, `user_message` in `workflow_runs`, `name` in `codebases`) are only enforced for **new** databases. For existing databases, `CREATE TABLE IF NOT EXISTS` is a no-op so the old schema remains. SQLite lacks ALTER COLUMN support; enforcing new constraints on existing tables would require a table rebuild via `migrateColumns()`.

So a user who installed last year and a user who installed today are running **structurally different databases on identical code**, with different constraint enforcement on at least three named columns. Neither of them can tell, and neither can we — a bug report cannot state its own schema vintage.

The limitation itself is a reasonable engineering tradeoff. **Its invisibility is not.**

## 2. Two Archon versions can write one database, with no way to detect it

The CLI talks to SQLite/Postgres **directly** — it does not go through the server. So a released binary on `PATH` beside a dev checkout is two writers with different schema expectations against one `~/.archon/archon.db`. That is the maintainer's own setup, and it becomes an ordinary user's setup after any upgrade that leaves an old binary behind.

Today this is benign, and for a specific reason worth writing down: every `ADD COLUMN` we ship pairs `NOT NULL` with a `DEFAULT` —

```sql
ALTER TABLE remote_agent_users ADD COLUMN role TEXT NOT NULL DEFAULT 'admin'
ALTER TABLE remote_agent_codebases ADD COLUMN kind TEXT NOT NULL DEFAULT 'repo'
```

— so an older writer that omits the column still produces a valid row. **The safety is a property of how every migration happens to have been written, not a rule anyone stated.** The first `NOT NULL` without a default, or the first non-additive change, silently breaks the older writer, and the damage is written rows rather than a failed startup.

## Why this ranks above the other recording defects

By severity of what a wrong answer *becomes*:

- **wrong behaviour** → noticed, because something acts on it and the act is visible
- **wrong number** → believed, because nothing downstream re-examines it (#2314)
- **wrong stored data** → **propagates**, because it is the input to everything later and cannot be re-derived

A believed benchmark number is recoverable — learn the alias rolled, re-run it. A row written by a writer that did not know the schema had moved is not recoverable by re-running anything; every later read inherits it. This outranks #2314 on #2314's own axis.

## Proposed

**Step one, and it unblocks everything else: record the vintage.** `PRAGMA user_version` on SQLite and an equivalent row on Postgres, written by the same idempotent apply that already runs on connect. Cheap, additive, and it converts an unknowable into a readable fact. Until it exists, **no policy here is enforceable** — you cannot refuse, warn about, or even report a mismatch you cannot detect.

Then, in rough order:

- **Decide and document whether migrations may ever be non-additive.** Right now the answer is "no, implicitly, because nobody has tried." Say it out loud — it is a permanent constraint on the project and it should not be discovered by a user's corrupted rows.
- **Surface the vintage in `archon doctor`** and in the health payload, so a bug report can state it.
- **Decide the two-writer question explicitly**: supported, warned, or refused. Any answer is fine; the current state is "undefined and undetectable".
- Consider whether the documented SQLite NOT NULL divergence deserves a one-time `migrateColumns()` rebuild, or a permanent documented "old databases differ here".

## Deliberately not proposing

A migration framework, a versioned migration ledger, or up/down migrations. The idempotent-additive approach is a good fit for a single-tenant install that must converge without operator action. This issue is about making its state **legible**, not replacing it.

## Related

- #2315 — API compatibility contract. Same root cause (nothing declares what it is compatible with), different surface: that one is client↔server over HTTP, this one is writer↔store. Split because the audiences and fixes do not overlap.
- #2314 — resolved model never recorded. Same family (interpretation needs context the data does not carry); this one outranks it on severity.

Wirasm:
## Summary

- **Problem:** Archon records no schema version anywhere. `PRAGMA user_version` is 0, there is no version table, and nothing tracks what has been applied. So a database created a year ago and one created today can be structurally different on identical code, and neither Archon, the user, nor a second concurrent writer can tell. Separately, the property that makes the idempotent-additive approach safe — every shipped `ADD COLUMN NOT NULL` carries a `DEFAULT` — was true by authorship accident, not by any stated rule.
- **Why it matters:** the dual hand-maintained schemas have already drifted **three** times (PR 988, PR 2033, and a live `codebases.allow_env_keys` divergence found during this investigation), and the only automated guard compares **table names**, not columns. A bug report currently cannot state its own schema vintage.
- **What changed:** the additive-only rule is now written into `CLAUDE.md`, and a `remote_agent_schema_version` row records which Archon build created the database and which last applied schema to it — written by the existing idempotent apply-on-connect path, surfaced in `archon doctor` and `GET /api/health`.
- **What did *not* change (scope boundary):** no migration framework, no versioned ledger, no up/down migrations, no two-writer refuse/warn policy, no table rebuild for the three documented `NOT NULL` columns, and no touching the `allow_env_keys` drift. The idempotent-additive approach stays exactly as it was; this makes its state legible, not different. Nothing gates on the new values.

### Before

```
  Maintainer                Archon                        Database
  ----------                ------                        --------
  "what vintage is          archon doctor
   this DB?"        ------> Database: reachable (sqlite)
                                                    <---- (no version recorded)
  no answer exists <------- (nothing to report)

  released binary  ------>  applies schema v_old  ------> ~/.archon/archon.db
  dev checkout     ------>  applies schema v_new  ------> ~/.archon/archon.db
                            (neither can see the other)
```

### After

```
  Maintainer                Archon                        Database
  ----------                ------                        --------
  "what vintage is          archon doctor
   this DB?"        ------> Database: reachable (sqlite);
                            [schema created by 0.5.3,
                             last applied by dev]   <---- [remote_agent_schema_version]

  GET /api/health  ------>  { ..., [schema: {
                                createdAppVersion: "0.5.3",
                                appVersion: "dev",
                                appliedAt: "..." }] }

  released binary  ------>  applies schema  ------> [records app_version = 0.6.0]
  dev checkout     ------>  applies schema  ------> [records app_version = dev]
                            *the flip-flop is now visible in the row*

  Pre-existing DB  ------>  [created_app_version = NULL]
                            reported as "<unknown - predates version tracking>"
                            *never back-filled with a guess*
```

### Before

```
                       @archon/paths
                      (BUNDLED_VERSION)
                             |
              +--------------+---------------+
              |                              |
        connection.ts                        |
              |                              |
      +-------+--------+                     |
      |                |                     |
 SqliteAdapter    PostgresAdapter            |
  createSchema()   initSchema()              |
  migrateColumns() (advisory lock 1796)      |
      |                |                     |
      v                v                     v
 (schema applied, nothing recorded)    api.ts /api/health
                                       doctor.ts checkDatabase
                                       (report reachability only)
```

### After

```
                       @archon/paths
                      (BUNDLED_VERSION)
                             |
                  [+] @archon/paths/bundled-build   <-- new subpath export
                             ||
                             || ===
                             v
              [+] core/db/schema-version.ts
                  APP_VERSION, readSchemaVersion()
                             ||
              +------========++============+
              ||             ||           ||
        [~] connection.ts    ||           ||
          getSchemaVersion() ||           ||
              |              ||           ||
      +-------+--------+     ||           ||
      |                |     ||           ||
 [~] SqliteAdapter  [~] PostgresAdapter   ||
  hasAnyArchonTable()  to_regclass probe  ||
  recordSchemaVersion() upsert in the     ||
  (warn-and-continue)   existing txn      ||
      |                |                  ||
      v                v                  v
 [+] remote_agent_schema_version    [~] api.ts /api/health  (schema field)
     (both dialects)                [~] doctor.ts checkDatabase (vintage line)
```

**Connection inventory**

| `core/db/schema-version.ts` | `@archon/paths/bundled-build` | **new** | Subpath, not the barrel — 80+ test files mock `@archon/paths` with a partial surface and both adapters record on construction |
| `adapters/sqlite.ts` | `core/db/schema-version.ts` | **new** | `APP_VERSION` only |
| `adapters/postgres.ts` | `core/db/schema-version.ts` | **new** | `APP_VERSION` only |
| `core/db/connection.ts` | `core/db/schema-version.ts` | **new** | `readSchemaVersion()` behind `getSchemaVersion()` |
| `core/db/schema-version.ts` | `core/db/connection.ts` | **absent (deliberate)** | Would close an import cycle; the read function takes an `IDatabase` |
| `cli/commands/doctor.ts` | `@archon/core` | **modified** | `defaultLoadDatabaseDeps` also pulls `getSchemaVersion` |
| `server/routes/api.ts` | `@archon/core` | **modified** | `/api/health` reads the vintage |
| `adapters/sqlite.ts` and `migrations/000_combined.sql` | — | unchanged | Still two hand-maintained sources; the new table is in both |
| `@archon/paths` package exports | `./bundled-build` | **new** | Mirrors the existing `./env-loader` subpath precedent |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: M`
- Scope: `core`
- Module: `core:db`

## Change Metadata

- Change type: `feature`
- Primary scope: `core`

## Linked Issue

- Closes #2316
- Related #2314 (same family: interpretation needs context the data does not carry), #2315 (same root cause, different surface)
- Depends on # — none
- Supersedes # — none

```bash
bun run validate
```

**Result: exit 0.** All nine stages ran and passed, from the worktree:

```
bundled-defaults.generated.ts is up to date (36 commands, 21 workflows).
bundled-skill.ts is up to date (23 files across 2 skills).
bundled-schema.generated.ts is up to date.
check:pi-vendor-map OK
check:capability-matrix OK
@archon/{paths,git,docs-web,providers,isolation,web,workflows,core,adapters,server,cli} type-check: Exited with code 0
bun x eslint . --cache --max-warnings "0"        (no output — clean)
bun x prettier --check .                          All matched files use Prettier code style!
bun --filter '*' --parallel test                  6388 passed, 0 failed
```

- Evidence provided: full `validate` log captured; the four touched test files were also run individually (sqlite 16 pass, postgres 39 pass, doctor 70 pass, health 20 pass).
- Intentionally skipped: `packages/web/src/lib/api.generated.d.ts` was **not** regenerated. It requires a running server, no web code consumes the new `schema` field, and neither `validate` nor CI checks that file for staleness.

## Security Impact (required)

The new row holds an Archon version string and two timestamps. `/api/health` is already public (`PUBLIC_API_GATE_PREFIXES`); it now also discloses which Archon build wrote the database, which is the same class of information the existing public `version` field already exposes.

## Compatibility / Migration

- Backward compatible? **Yes.** Purely additive on both dialects. An older Archon binary opening a database that already has the table simply ignores it.
- Config/env changes? **No**
- Database migration needed? **No manual step.** The table is created by the same idempotent apply-on-connect that already runs; `bun run generate:bundled-schema` was run and `check:bundled-schema` passes.

Upgrade behaviour on an existing database: the table is created, `created_app_version` is recorded as `NULL` (correctly unknown), and `app_version` is set to the build that opened it.

Verified scenarios (run against scratch `ARCHON_HOME` directories, not the real one):

1. **Fresh database** — `archon doctor` prints `Database: reachable (sqlite); schema created by dev, last applied by dev`; the row is `{"id":1,"created_app_version":"dev","app_version":"dev",...}`.
2. **Pre-existing database** (seeded `remote_agent_codebases` with no version row, then opened by the new adapter) — `Database: reachable (sqlite); schema created by <unknown — predates version tracking>, last applied by dev`; the row is `{"created_app_version":null,"app_version":"dev",...}`. The creation vintage is not invented.
3. **Live health payload** — booted the server on a scratch home and hit `GET /api/health`: `"schema": {"createdAppVersion":"dev","appVersion":"dev","appliedAt":"..."}`.
4. Raw read of the table on both databases confirms a single `id = 1` row.

Edge cases checked: reopening a database does not revise `created_app_version` (test); a vintage read that throws leaves the doctor check at `pass` and returns health `200` with `schema` omitted (tests); the Postgres upsert is issued *after* the schema SQL and *before* `COMMIT`, inside the advisory-locked transaction (test); the pre-existence probe runs *before* the schema SQL, since afterwards a fresh and a pre-existing database are indistinguishable (test).

**What was not verified:** the Postgres path against a real PostgreSQL server. `postgres.test.ts` is fully mocked (as it already was), so the `to_regclass` probe and the `ON CONFLICT ... WHERE ... IS DISTINCT FROM` upsert are asserted at the SQL-issued level, not executed. Everything hands-on above was SQLite.

- **Affected subsystems:** the database open path for *both* dialects — i.e. the server, every CLI invocation, and every `--detach` child. This is the widest-reaching part of the change and the reason the SQLite write is wrapped in warn-and-continue.
  - One extra `SELECT` on every database open. The write only happens when the recorded version actually differs, so steady state is read-only and there is no WAL churn per CLI command.
  - Postgres adds two statements inside the existing advisory-locked init transaction (a probe and an upsert) — it runs once per process, not per query.
  - `/api/health` gains an optional field; existing consumers are unaffected because it is optional and additive.
  - `@archon/paths` gains a subpath export; nothing else is re-pointed at it.
- **Guardrails:** `db.sqlite_schema_version_record_failed` (warn), `doctor.schema_version_read_failed` (warn), `api.schema_version_read_failed` (warn). None is fatal, all three are logged rather than swallowed.

## Rollback Plan (required)

- **Fast rollback:** revert the commit. The table is additive and unread by anything else, so a reverted build simply stops writing and reading it; the orphaned table is harmless and no data is lost. Do **not** add a `DROP TABLE` — that would be the first non-additive migration and is exactly what the rule in this PR forbids.
- **Feature flags:** none. The behaviour is unconditional and diagnostic.
- **Observable failure symptoms:** `db.sqlite_schema_version_record_failed` warnings on startup, `archon doctor` reporting `schema vintage not recorded` on a database that should have it, or `/api/health` omitting `schema`.

## Risks and Mitigations

- **Risk:** the write runs on the database-open path, which every process hits — a bug here would be broad.
  - **Mitigation:** SQLite's write is wrapped in try/catch-and-warn (mirroring `migrateColumns()`), so it cannot prevent the database from opening; Postgres's runs inside the transaction that already rolls back and fails loudly. Covered by tests on both adapters plus hands-on runs.
- **Risk:** `'dev'` collapses every source checkout into one label, so two dev checkouts are indistinguishable, and `/api/health` shows `version: "0.6.0"` next to `schema.appVersion: "dev"` in source builds.
  - **Mitigation:** accepted deliberately. `BUNDLED_VERSION` is the truthful answer for a working tree — a dev checkout mid-development is *not* release 0.6.0 — and for the scenario the issue actually describes (a released binary on `PATH` beside a dev checkout) `created 0.5.3, last applied dev` is precisely the diagnosis. In compiled binaries both fields carry the real semver and agree.
- **Risk:** a new table in a schema whose management is the thing under scrutiny.
  - **Mitigation:** it is additive and idempotent, present in both dialects (so the existing table-name parity test covers it with no new allowlist exception), and nothing gates on it.

---

### Note: I diverged from the issue's proposed mechanism

The issue proposes `PRAGMA user_version` on SQLite and "an equivalent" on Postgres. I used a one-row table on both dialects holding **version strings**, not a schema integer. Reasons, in order of weight:

1. **A hand-bumped integer reproduces the exact failure the issue objects to.** Its own diagnosis is that safety today "is a property of how every migration happens to have been written, not a rule anyone stated." A `SCHEMA_VERSION = 7` constant a contributor must remember to bump has the same weakness, and fails *worse*: a stale version is a **wrong answer that gets believed**, which the issue itself ranks as the more damaging outcome. With three shipped drifts on record, betting the new mechanism on that discipline isn't defensible.
2. **The app version needs no discipline.** `BUNDLED_VERSION` is correct automatically.
3. **Two facts are needed, and the issue asks for both.** The constraint-divergence consequence turns on *when the database was created*; the two-writer consequence on *what touched it last*. A bare `user_version` integer holds one number and could carry neither without a lossy encoding.
4. **Postgres has no pragma**, so a table is needed there regardless — and a pragma on SQLite plus a table on Postgres would mean two mechanisms, two read paths, and a second permanent exception in the parity test.

### Two things found during investigation that are *not* fixed here

- **`codebases.allow_env_keys` is live drift**: present in `migrations/000_combined.sql` (`:41`, `:385-386`) and absent from `sqlite.ts` entirely. Commit `a7113477` added it to SQLite to fix an earlier instance of this exact bug; `a8ac3f05` (PR 1135) then removed the env-leak-gate feature from `sqlite.ts` and left the Postgres column behind. It is harmless only because it is dead — no application code reads it on either dialect. **Not touched here**: removing it from Postgres would be a non-additive drop, precisely what the rule this PR adds forbids. Worth its own issue.
- **The parity test is table-name only.** Extending it to column level would have caught all three historical drifts and is plausibly worth more than this PR's vintage row — but it has its own design questions (cross-dialect type mapping) and doesn't belong in a P3 observability change.

### On the label

This is labelled `bug`, but by its own audit nothing is broken and there is no user-visible failure — it's an observability enhancement. P3 is right; the label isn't. Not relabelling, just noting it.

## Summary by CodeRabbit

* **New Features**
  * Added database schema “vintage” diagnostics (created app version, current app version, applied timestamp) and a single-row vintage record.
  * Surfaced schema vintage in the `archon doctor` output and in `GET /api/health` (optional `schema` field).
  * Added a default `admin` role for remote agent users.
* **Bug Fixes**
  * Health and diagnostic checks still report success even if schema-vintage metadata can’t be read.
* **Documentation**
  * Updated the database schema-change contract to enforce additive-only, additive-safe migration rules and clarify diagnostic behavior.


Wirasm:
## Self-review

Ran the code-reviewer and simplifier passes over the diff. One real issue found and fixed; four simplifications taken, one declined with reasoning. Fixes are in `f3d9bb53`.

### Issue found and fixed

**The Postgres vintage write was not failure-isolated, unlike its SQLite counterpart.** `packages/core/src/db/adapters/postgres.ts` — the upsert sat inside the schema transaction with no isolation, so a throw there would have propagated to `initSchema()`'s catch, rolled back, and rejected `schemaInitPromise`. Every `query()` / `withTransaction()` / `listen()` awaits that promise, so awaiting an already-rejected promise would re-throw forever: the adapter would be permanently unusable for the life of the process, over a row nothing gates on.

That silently broke the contract this PR itself states — `CLAUDE.md` now says the vintage is "diagnostic only", and the SQLite adapter honours that with warn-and-continue. Postgres did not, on the dialect the VPS actually runs.

Fixed with `SAVEPOINT` / `RELEASE SAVEPOINT` around the upsert and `ROLLBACK TO SAVEPOINT` + warn on failure, so the schema transaction still commits. Added a test asserting the adapter stays usable and `COMMIT` (not `ROLLBACK`) is issued when the vintage insert throws.

### Simplifications taken

| Change | Rationale |
|---|---|
| Extracted the savepoint block into `recordSchemaVersion(client, preExisting)` | Mirrors the existing `installNotifyTrigger()` precedent in the same file, and gives name symmetry with the SQLite method. Drops `initSchema()` back to two nesting levels. |
| `api.ts`: shape the health payload next to the read | Keeps the "`createdAt` is deliberately not exposed" decision in one place instead of split across the handler, and lets the return match the `wsl_distro` spread directly above it. |
| `doctor.test.ts`: hoisted a `makeDeps()` helper | The `checkFolderProject` block in the same file already uses exactly this pattern; six near-identical deps objects now state only what they vary. |
| `schema-version.ts`: inlined `SCHEMA_VERSION_TABLE` | It had one reader and the other six occurrences of the table name are hardcoded literals, so it was indirection rather than a single source of truth. |

### Simplification declined

**Collapsing SQLite's read-then-branch into a single `INSERT ... ON CONFLICT DO UPDATE ... WHERE`.** Tidier, and it would match what Postgres does — but an upsert whose `DO UPDATE ... WHERE` predicate fails still opens a write transaction. Every process that opens the database (the server, *every* CLI invocation, every `--detach` child) would then take a write lock where a `SELECT` on a one-row table suffices today, and a read-only database file would warn on every open instead of only when a write is genuinely needed. The extra `SELECT` is the cheaper trade for a diagnostic row.

### Reviewed clean

Pre-existence probe ordering on both dialects (it must precede the schema apply — afterwards a fresh and a pre-existing database are indistinguishable); `ON CONFLICT` never touching `created_app_version`; column-by-column parity between `sqlite.ts` `createSchema()` and `migrations/000_combined.sql`; no silently swallowed errors (all three fallbacks log a warn and are justified in comments); `IDatabase` left untouched per ISP, with the capability as a free function plus a narrow `DatabaseDeps` field; no `any`; the `./bundled-build` subpath correctly bypassing the 85 partial `@archon/paths` barrel mocks.

### Validation after fixes

`bun run validate` → **exit 0**. All nine stages: five generated-file checks up to date, type-check clean across 11 packages, `eslint --max-warnings 0` clean, prettier clean, **6389 tests passed / 0 failed**.

Re-verified by hand after the refactor (behaviour unchanged): `archon doctor` on a fresh scratch home prints `reachable (sqlite); schema created by dev, last applied by dev`, and `GET /api/health` returns `"schema":{"createdAppVersion":"dev","appVersion":"dev","appliedAt":"..."}`.

### Checklist

- [x] Addresses the issue's actual ask (record the vintage) and the higher-value half (state the rule)
- [x] Follows codebase patterns (warn-and-continue mirrors `migrateColumns()`; method extraction mirrors `installNotifyTrigger()`; subpath export mirrors `@archon/paths/env-loader`)
- [x] Tests cover the change, including the case that matters most — a pre-existing database reports `NULL` rather than inventing a creation vintage
- [x] No bug introduced on the database open path

**Not verified:** the Postgres path against a real PostgreSQL server. `postgres.test.ts` is fully mocked (as it already was), so the probe, the upsert, and the new savepoint behaviour are asserted at the SQL-issued level, not executed. Everything hands-on was SQLite.

Wirasm:
## CodeRabbit review response

One finding taken (a good one), two rejected as false positives. Fix in `2248d56b`.

### Taken: "Do not record a failed schema apply as current" (Major)

Correct, and it lands on exactly the axis the issue argues from. `migrateColumns()` suppresses each table's failure so one bad `ALTER` cannot abort startup — but `recordSchemaVersion()` then stamped `APP_VERSION` unconditionally. A database missing a failed migration would report that this build last applied its schema. That is a **wrong answer that gets believed**, which #2316 itself ranks as more damaging than no answer.

Fixed: `migrateColumns()` now returns whether every block succeeded, and the vintage write is skipped when it did not, with a `db.sqlite_schema_version_skipped_incomplete_migration` warn naming the version that was withheld. A stale or absent vintage is a state a reader can act on; a confidently wrong one is not. The next clean open records it.

Test added: seeds a *view* named `remote_agent_users` so the users `ALTER` genuinely fails, then asserts no vintage row is written and `readSchemaVersion()` returns `null`. Verified non-vacuous — the run emits `db.sqlite_migration_users_columns_failed` followed by `db.sqlite_schema_version_skipped_incomplete_migration`.

Note this applies to SQLite only, by design. On Postgres the schema apply is a single transaction that either commits or rolls back and rethrows, so there is no partial-apply state to mis-stamp.

### Rejected: the two "Critical" findings

Both claim duplicated lines that would make the files fail to parse. They do not exist.

- **`postgres.test.ts:400` — "duplicate `issued` declarations"**: there are 8 `const issued` declarations in the file, each inside its own `test(...)` callback scope. Lines 400 and 462 are in different tests. No redeclaration.
- **`sqlite.test.ts:442` — "duplicated type assertions"**: lines 442, 462 and 493 each have exactly one `) as ...[];`. No duplicates.

Verification: `bun x tsc --noEmit -p packages/core/tsconfig.json` exits 0, and both suites run (`sqlite.test.ts` 17 pass, `postgres.test.ts` 40 pass). A file that "cannot parse" would fail both. These read as diff-hunk misreads.

### Validation after the fix

`bun run validate` → **exit 0**: five generated-file checks up to date, type-check clean across 11 packages, `eslint --max-warnings 0` clean, prettier clean, **6390 tests passed / 0 failed**.

Previous commit `f3d9bb53` was green on CI (docker-build, test ubuntu-latest, test windows-latest all pass); this commit is re-running.

PR #2330, opened 2026-07-30
Fix Rosetta architecture detection in macOS installer

Wirasm:
## Summary

- Problem: the quick installer selected `darwin-x64` when invoked from a Rosetta-translated shell on Apple Silicon.
- Why it matters: the downloaded binary passed checksum verification but could not run, leaving users with an unusable CLI.
- What changed: the installer detects Rosetta, selects ARM64, verifies a downloaded binary before replacement, and adds deterministic shell coverage.
- What did **not** change (scope boundary): release targets and Homebrew remain unchanged; the release pipeline already publishes both Darwin assets.

### Before

```
Apple-Silicon user in Rosetta shell    Installer                 Release assets
──────────────────────────────────    ─────────                 ──────────────
curl .../install | bash ───────────▶  uname -m => x86_64
                                      selects darwin-x64 ─────▶ downloads x64 binary
                                      checksum passes
                                      replaces existing CLI
                                      version probe fails ────▶ unusable install
```

### After

```
Apple-Silicon user in Rosetta shell    Installer                         Release assets
──────────────────────────────────    ─────────                         ──────────────
curl .../install | bash ───────────▶  uname -m => x86_64
                                      [sysctl.proc_translated => 1]
                                      [selects darwin-arm64] ─────────▶ downloads ARM64 binary
                                      [probes downloaded binary first]
                                      replaces existing CLI only on success
```

### Before

```
scripts/install.sh ───────────────▶ release asset URL
        │
        └─────────────────────────▶ packages/docs-web/public/install (mirrored static installer)
package.json ─────────────────────▶ validation commands
```

### After

```
[~] scripts/install.sh ───────────▶ release asset URL
        │ Rosetta-aware platform resolver
        ├─────────────────────────▶ [~] packages/docs-web/public/install
        └=========================> [+] scripts/test-install.sh
[~] package.json =================▶ [+] installer regression validation
```

**Connection inventory** (list every module-to-module edge, mark changes):

| From | To | Status | Notes |
|------|----|--------|-------|
| `scripts/install.sh` | GitHub release assets | **modified** | Selects ARM64 for translated Darwin x64 processes. |
| `scripts/install.sh` | `packages/docs-web/public/install` | **modified** | Public installer remains byte-identical to the canonical script. |
| `scripts/test-install.sh` | installer copies | **new** | Tests platform resolution, mirror equality, and failed-probe preservation. |
| `package.json` | `scripts/test-install.sh` | **new** | Runs installer regression coverage through `validate`. |

[...]

PR #2335, opened 2026-07-30
Run installer tests in CI to prevent mirror drift

### LINKED ISSUE #2332: fix(ci): install.sh and docs-web/public/install are hand-synced duplicates — a stale public installer passes CI
## Summary

`scripts/install.sh` and `packages/docs-web/public/install` are **byte-identical** copies of the installer, kept in sync by hand. Nothing enforces it.

The second path is what `curl -fsSL https://archon.diy/install | bash` actually serves — the documented one-line install in the README. So a divergence means **the public installer silently differs from the one in the repo**, and the repo copy is the one anyone reviews.

## Evidence

Verified on the merge commit of #2330 (`08723171`):

```
diff <(git show HEAD:scripts/install.sh) <(git show HEAD:packages/docs-web/public/install)
→ identical
```

#2330 (the Rosetta architecture fix, #2295) had to apply the **same 23-line change twice**, once to each file. Its commit body lists *"keep the public installer mirror synchronized"* as a separate deliverable from the fix itself — i.e. the author noticed the hazard and paid the tax manually.

That is the whole problem: it works exactly as long as every future author both knows the mirror exists and remembers it. The new `scripts/test-install.sh` (added by #2330) tests **`scripts/install.sh`** only, so a stale mirror passes CI.

## Failure mode

Fix an installer bug in `scripts/install.sh`, forget the mirror, ship. `bun run test:install` passes, CI passes, review passes — and every user who follows the documented install path gets the **old, broken** installer. The repo says the bug is fixed. It is, in the copy nobody downloads.

This is the same defect class as #2319 (two hand-maintained schema sources with a guard that only compares table names) and #2326 (two artifact vocabularies): **two sources of truth, no check, and the divergence is invisible until a user hits it.**

## Proposed

Options, cheapest first:

1. **Generate one from the other at build time.** `packages/docs-web/public/install` becomes a build artifact copied from `scripts/install.sh`, and is gitignored. One source, impossible to drift. Requires the docs build to run before publish — it already does.
2. **Assert identity in CI.** A `check:install-mirror` script doing the `diff` above, wired into `bun run validate` alongside the existing `check:bundled*` family. Keeps both files tracked, makes drift a red build. Smallest change, fits an established pattern in this repo.
3. **Serve the script from one path.** Have the docs site redirect or proxy `/install` to the repo copy rather than hosting a duplicate. Cleanest conceptually, most infrastructure.
4. **Point `test:install` at both.** Necessary regardless of the above — the test currently exercises one file and would pass a stale mirror.

**Recommendation: (2) plus (4).** This repo already has five `check:*` generated-file guards in `validate` (`check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`), all solving exactly this shape. A sixth is consistent, obvious to the next author, and needs no build-order change. (1) is arguably better long-term but changes what is tracked in git, which deserves its own decision.

## Related

- #2330 / #2295 — the PR that had to update both copies; found here.
- #2319 — dual schema sources, guard too weak.
- #2326 — two artifact vocabularies with no reconciliation.

Wirasm:
## Summary

- Problem: GitHub's Test Suite did not execute the existing installer test, so a stale docs-served installer could pass CI.
- Why it matters: `https://archon.diy/install` serves the mirrored public installer used by the documented curl-and-pipe installation path.
- What changed: added a named `bun run test:install` step before workspace tests in the Test Suite workflow.
- What did **not** change (scope boundary): installer implementations, their tracked mirror model, docs deployment, and the PowerShell installer remain unchanged.

### Before

```
Contributor            GitHub Test Suite             Public installer
───────────            ─────────────────             ────────────────
updates installer ──▶  runs workspace tests only
                                                     stale mirror can deploy
```

### After

```
Contributor            GitHub Test Suite             Public installer
───────────            ─────────────────             ────────────────
updates installer ──▶  [runs installer tests] ──▶    mirror parity checked
                         [fails on divergence]
```

### Before

```
.github/workflows/test.yml ──▶ bun run test
scripts/test-install.sh ──▶ scripts/install.sh
scripts/test-install.sh ──▶ packages/docs-web/public/install
```

### After

```
.github/workflows/test.yml ──▶ [bun run test:install] ──▶ scripts/test-install.sh
scripts/test-install.sh ──▶ scripts/install.sh
scripts/test-install.sh ──▶ packages/docs-web/public/install
.github/workflows/test.yml ──▶ bun run test
```

| Test Suite workflow | `test:install` package script | **new** | Executes installer behavior and public-mirror parity tests in CI. |
| `test:install` package script | `scripts/test-install.sh` | unchanged | Existing script entrypoint. |
| `scripts/test-install.sh` | Canonical and public installer copies | unchanged | Existing byte-parity and installer behavior coverage. |
| Test Suite workflow | workspace tests | unchanged | Existing package test suite continues to run. |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `ci`
- Module: `ci:test-suite`

## Change Metadata

- Change type: `bug`
- Primary scope: `multi`

## Linked Issue

- Fixes #2332
- Related #2330
- Depends on # None
- Supersedes # None

Commands and result summary:

```bash
bun run test:install                  # passed: Installer tests passed
bun x prettier --check .github/workflows/test.yml  # passed

# Validation artifact from the implementation run:
bun run type-check                    # passed
bun run lint                          # passed, 0 warnings
bun run format:check                  # passed
bun run test                          # passed, 6,375 tests
bun run build                         # passed
```

- Evidence provided (test/log/trace/screenshot): `scripts/test-install.sh` completed successfully; the recorded validation artifact reports all checks passing.

## Security Impact (required)

## Compatibility / Migration

- Config/env changes? (`No`)

- Verified scenarios: ran the installer test suite, including the existing canonical/public installer parity assertion; checked workflow formatting.
- Edge cases checked: a mismatched public installer is covered by the existing `cmp -s` assertion exercised by the new CI step.
- What was not verified: a hosted GitHub Actions matrix run; this draft PR will exercise it on Ubuntu and Windows.

- Affected subsystems/workflows: the pull-request and push Test Suite on Ubuntu and Windows.
- Potential unintended effects: CI gains the execution time and platform requirements of the existing Bash installer test.
- Guardrails/monitoring for early detection: the named Actions step makes installer or mirror failures explicit in CI logs.

## Rollback Plan (required)

- Fast rollback command/path: revert commit `60acd057` or remove the `Run installer tests` step from `.github/workflows/test.yml`.
- Observable failure symptoms: Test Suite fails at the `Run installer tests` step for installer regressions or mirror drift.

## Risks and Mitigations

- Risk: the Bash installer suite could fail on one matrix platform.
  - Mitigation: the test is intentionally executed on both existing matrix legs to expose portability or parity regressions before merge.

## Summary by CodeRabbit

* **Tests**
  * Added automated installer testing to the test workflow on supported non-Windows environments.
  * Windows test runs continue to skip installer checks that require POSIX shell behavior.


Wirasm:

**Date**: 2026-07-30T11:40:05Z
**Status**: COMPLETE
**Branch**: archon/task-archon-fix-github-issue-experimental-1785410381718
**Commit**: ed84800ab250e9d72b064e4ad583b4c099744f58

---

## Summary

One high-severity CI regression was found: Windows checkouts could materialize the public installer mirror with different line endings, failing the strict byte-parity test. The public installer is now explicitly pinned to LF in Git attributes, matching `scripts/install.sh`. No findings were skipped or blocked.

---

## Fixes Applied

| HIGH | Windows checkout fails the installer mirror-parity assertion | `.gitattributes:6` | Added `packages/docs-web/public/install text eol=lf` so Windows checkouts preserve the same LF bytes as `scripts/install.sh`. |

---

## Tests Added

*(none — the existing installer regression suite directly verifies the fix)*

---

## Docs Updated

*(none)*

---

## Skipped Findings

---

## Blocked (Could Not Fix)

*(none)*

---

## Suggested Follow-up Issues

*(none)*

---

## Validation

| Check | Status |
|-------|--------|
| Type check | ✅ `bun run type-check` |
| Lint | ✅ `bun run lint` |
| Installer tests | ✅ `bun run test:install` |
| Tests | ✅ `bun run test` |

Wirasm:
---
pr: 2335
title: "Run installer tests in CI to prevent mirror drift"
author: "Wirasm"
reviewed: 2026-07-30T00:00:00Z
---

# PR Review: #2335 — Run installer tests in CI to prevent mirror drift

**Branch**: `archon/task-archon-fix-github-issue-experimental-1785410381718` -> `dev`
**Files changed**: 2 (+4/-0) — `.github/workflows/test.yml`, `.gitattributes`
**State**: DRAFT
**Closes**: #2332

---

## Summary

The mechanism works. I empirically confirmed, three independent ways, that the guard catches the divergence it was written for — including one live catch in CI. But the PR as it stands makes the `test (windows-latest)` leg of the Test Suite **permanently red**, confirmed by execution on this PR's own head commit. That is blocking.

Separately, the review surfaced a **live, currently-shipping instance of the exact bug class #2332 describes**, in the *other* installer mirror pair the PR explicitly scopes out.

**Does the check prevent the drift it claims to? Yes — empirically verified.**

---

## Verified by execution (not reasoning)

| # | Experiment | Result |
|---|-----------|--------|
| 1 | Baseline `bash scripts/test-install.sh` on PR head (macOS) | `Installer tests passed`, exit 0 |
| 2 | Appended **one space** to `packages/docs-web/public/install`, re-ran | `FAIL: public installer mirror differs from scripts/install.sh`, **exit 1** |
| 3 | Reverse direction — one-char "bugfix" to `scripts/install.sh` only, mirror untouched (the exact #2332 failure mode) | `FAIL: public installer mirror differs…`, **exit 1** |
| 4 | Fresh clone + `core.autocrlf=true` (simulated Windows checkout) at `dev` (`08723171`) | mirror gets **264 CRLF lines**, `install.sh` stays LF → parity **FAILS** |
| 5 | Same simulation at PR head (`ed84800a`, with the `.gitattributes` pin) | mirror **0 CRLF lines** → parity **PASSES** |
| 6 | Live CI, pre-fix commit `60acd057`, job `90858541891`/windows | **`FAIL: public installer mirror differs from scripts/install.sh`** — the guard fired in CI for real |
| 7 | Live CI, PR head `ed84800a`, windows job `90860473068` | **FAILURE**: `[ERROR] Windows is not supported. Please use WSL2` at the `Run installer tests` step |
| 8 | Live CI, PR head, ubuntu job `90860473039` | SUCCESS — `Installer tests passed` |

All work was done in an isolated scratch clone; the working repo was not modified.

---

## Answers to the specific review questions

**Does it actually prevent the drift?** Yes. `scripts/test-install.sh:71-72` holds a `cmp -s` byte-parity assertion. Experiments 2, 3 and 6 confirm it fails on divergence in either direction, including a real CI catch.

**Is it wired into `validate`?** Yes — but that predates this PR. `package.json:39` already chained `bun run test:install` into `validate` as of #2330 (`08723171`). So local `bun run validate` was already protected. This PR closes the *CI* half of the gap.

**CI-only vs validate?** Neither, in effect: `validate` covers the local path (pre-existing), this PR covers CI. Combined, the issue's intent is satisfied functionally. The one deviation from #2332's recommendation is *form*: the guard lives inside a bash installer test rather than as a discoverable `check:install-mirror` in the `check:*` family (see S2).

**Does it test the mirror or just run the existing tests?** It runs the existing test — **and that test does assert mirror parity**. Note issue #2332's premise ("`test-install.sh` tests `scripts/install.sh` only, so a stale mirror passes CI") was factually wrong about the script: #2330 already added the `cmp`. The real gap was that `test.yml` enumerates `validate`'s steps by hand and had omitted `test:install`.

Consequently issue item **(4)** ("point `test:install` at both files") is genuinely moot: once byte-identity is asserted, exercising the behavioral tests against one copy is sufficient.

**Shell portability?** This is where it breaks — see C1.

---

## Issues Found

### Critical

**C1 — The PR makes `test (windows-latest)` permanently red.** `.github/workflows/test.yml:50-51`

The new step has no platform guard, so it runs on both matrix legs. Chain:

- `scripts/test-install.sh:126` — `install_output=$(… bash "$installer")` runs the real installer with only `curl` mocked; `uname` is **not** mocked here.
- On a GitHub Windows runner, Git Bash `uname -s` returns `MINGW64_NT-…`, so `scripts/install.sh:57-59` hits `mingw*|msys*|cygwin*)` → `error "Windows is not supported…"; exit 1`.
- `scripts/install.sh:23` is `set -euo pipefail`, so the `exit 1` inside `platform=$(detect_platform)` (`install.sh:181`) aborts the installer non-zero.
- `scripts/test-install.sh:3` is also `set -euo pipefail`, so the failed command substitution at line 126 aborts the whole test script.

**Confirmed by execution**, not inferred: windows job `90860473068` on this PR's head fails with exactly `[ERROR] Windows is not supported. Please use WSL2 or see documentation.` / `error: script "test:install" exited with code 1`.

This is not a flake and will not self-resolve. `test.yml:17-21` documents that both legs must always report, and warns a red leg "blocks a merge once it is a required check."

**Fix (recommended, simplest):** gate the step, since the parity assertion is platform-independent and the Bash installer is never used on Windows (Windows users get `install.ps1`):

```yaml
      - name: Run installer tests
        if: runner.os != 'Windows'
        run: bun run test:install
```

**Alternative (more thorough):** keep the step on both legs but make `test-install.sh` skip only the two installer-*execution* blocks on `MINGW*`/`MSYS*` while still running the mock-based platform assertions and the `cmp` parity check. This retains Windows coverage of the drift guard. Verified viable: experiment 5 shows the parity check itself passes cleanly on a simulated Windows checkout now that the `.gitattributes` pin is in.

### Important

**I2 — Windows false pass at `scripts/test-install.sh:97`.** The negative test (`if … bash "$installer"; then fail "installer succeeded when downloaded binary failed its version check"`) expects a non-zero exit. On Windows the installer exits non-zero because it **refuses the OS**, not because it detected a bad binary — stderr is swallowed by `>/dev/null 2>&1`, so the test passes while asserting nothing. If C1 is fixed by keeping the step on Windows, this becomes a silently vacuous test. The `if: runner.os != 'Windows'` fix makes it moot.

**I3 — `install.ps1` has a second, unguarded mirror pair that has ALREADY diverged — and the divergence is a live shipped bug.**

Not introduced by this PR, and the PR body explicitly scopes it out ("the PowerShell installer remains unchanged"). But it is squarely inside #2332's premise, and it is the highest user-impact fact this review found:

```
scripts/install.ps1:191                    $pathParts = @($currentPath -split ';' | Where-Object { $_ -ne '' })
packages/docs-web/public/install.ps1:191    $pathParts =   $currentPath -split ';' | Where-Object { $_ -ne '' }
```

The `@()` wrapper was added by `53cabd44` — *"fix: PowerShell Add-ToUserPath corrupts PATH when single entry exists (#1000)"*. It was **never mirrored**. Without `@()`, when the pipeline yields exactly one entry `$pathParts` is a scalar string, so `$pathParts += $Dir` (`install.ps1:193`) string-concatenates instead of appending, and the user's PATH is written corrupted.

`irm https://archon.diy/install.ps1 | iex` is the documented Windows install path in `README.md:166`, `packages/docs-web/src/content/docs/docs.mdx:37`, and `.../getting-started/installation.md:21`. So the public Windows installer still ships bug #1000, which the repo reports as fixed. That is #2332's stated failure mode verbatim: *"The repo says the bug is fixed. It is, in the copy nobody downloads."*

**Fix:** add a second assertion (this is two lines and closes the class properly):

```bash
cmp -s "$repo_root/scripts/install.ps1" "$repo_root/packages/docs-web/public/install.ps1" \
  || fail "public install.ps1 mirror differs from scripts/install.ps1"
```

Note this will fail until the mirror is re-synced — which is the point, and it fixes a live user-facing bug. If you'd rather keep this PR minimal, file it immediately as its own issue; it should not wait on a CI-wiring PR.

**I4 — CI still omits three of `validate`'s five `check:*` guards — the same defect class this PR is fixing.** `.github/workflows/test.yml:34-38`

Mechanically diffed `package.json:39` against the workflow:

| In `validate` | In CI |
|---|---|
| `check:bundled` | yes |
| `check:bundled-skill` | **NO** |
| `check:bundled-schema` | **NO** |
| `check:pi-vendor-map` | **NO** |
| `check:capability-matrix` | yes |

The root cause of #2332's CI half is that `test.yml` re-enumerates `validate`'s chain by hand, so it drifts. This PR patches one symptom and leaves three. Worse, `CLAUDE.md:852` asserts the opposite as a guarantee: *"`bun run validate` (and CI) run `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, and `check:capability-matrix` and will fail loudly if any generated file is stale."* CI does not. A stale bundled skill, embedded schema, or Pi vendor map merges green today.

**Fix:** replace the enumerated steps with a single `run: bun run validate`, which makes drift structurally impossible; or at minimum add the three missing steps. The single-command form is the systemic fix and would have prevented this issue entirely.

### Suggestions

**S1 — `CLAUDE.md:174` is stale.** "This runs `check:bundled`, …, format check, and tests. All nine must pass for CI to succeed." `test:install` is now a tenth step and is unlisted; and "for CI to succeed" is untrue given I4. Pre-existing (introduced by #2330), but this PR is the natural place to correct it.

**S2 — Discoverability.** #2332 recommended a `check:install-mirror` in the `check:*` family. The guard instead lives at `scripts/test-install.sh:71`, inside a script named for installer *behavior*. A future author adding a third mirrored asset will scan the `check:*` family and not find the precedent. Cheap improvement: a `check:install-mirror` script that does both `cmp`s, called from `validate` and CI, leaving `test-install.sh` for behavior. Not blocking.

**S3 — Test ordering nit.** The parity `cmp` at `test-install.sh:71` runs after the platform assertions and before the exec tests. Since parity is the cheapest and most load-bearing assertion, moving it to the top gives faster, unambiguous failures (and would have made the C1 diagnosis instant).

---

## Validation Results

| Check | Status | Details |
|-------|--------|---------|
| `test:install` (macOS, scratch clone) | PASS | `Installer tests passed` |
| Drift detection (both directions) | PASS | fails correctly on 1-char divergence |
| `prettier --check .github/workflows/test.yml` | PASS | |
| CI — ubuntu-latest | PASS | installer step: `Installer tests passed` |
| CI — windows-latest | **FAIL** | `Windows is not supported` at the new step |
| CI — docker-build | PASS | |
| type-check / lint / full test suite | not re-run locally | diff touches only CI YAML + `.gitattributes`; ubuntu CI leg green covers it |

---

## Strengths

- **The `.gitattributes` pin is a genuinely sharp catch.** `packages/docs-web/public/install` is extensionless, so `*.sh text eol=lf` never matched it and it fell through to `* text=auto` → CRLF on Windows checkouts → guaranteed parity failure. I verified precedence (`.gitattributes:2` vs `:6`, last match wins) with `git check-attr`, and verified the effect with fresh-clone `autocrlf=true` simulations: `dev` → 264 CRLF lines / parity fails; PR head → 0 / parity passes. Without this the new step could not pass on Windows at all.
- **The guard demonstrably works.** It fired for real in CI on the pre-fix commit — rare to get a live proof-of-function inside the PR's own history.
- **Minimal, cleanly revertible diff** with explicit scope boundaries and a stated rollback path.
- **The risk section predicted the exact failure that materialized** ("the Bash installer suite could fail on one matrix platform"; "not verified: a hosted GitHub Actions matrix run"). Honest self-assessment — the draft status is being used correctly.

---

## Recommendation

**NEEDS FIXES**

Required before merge:
1. **C1** — stop the Windows leg from failing (`if: runner.os != 'Windows'`, or skip the exec blocks on MINGW/MSYS).

Strongly recommended in this PR or an immediate follow-up:
2. **I3** — guard the `install.ps1` pair and re-sync it; the public Windows installer currently ships bug #1000.
3. **I4** — have CI run `bun run validate` instead of a hand-copied step list, so this class of CI/validate drift stops recurring.

The approach is right and the mechanism is proven. Fix C1 and this is a merge.

---

*Reviewed by Claude. Report: `/Users/rasmus/.prp/archon-75601ef6/reviews/pr-2335-review.md`*

Wirasm:
Pushed `445eafaf` — adds `if: runner.os != 'Windows'` to the installer-test step, which was the blocking finding from review.

Cause: `scripts/install.sh` refuses MINGW/MSYS by design (Windows users are pointed at WSL2), and `test-install.sh` invokes the **real** installer with only `curl` mocked — `uname` is not stubbed at that call site. So on `windows-latest` Git Bash reports `MINGW64_NT`, the installer exits, and `set -euo pipefail` fails the job. The ubuntu leg still runs the suite, and byte parity is platform-independent, so drift coverage is unaffected.

**Sequencing: this should merge after #2340.** There is a critical regression on `dev` right now (#2338) — the `BASH_SOURCE[0]` guard from #2330 makes `curl … | bash` exit 1 for every user, because `BASH_SOURCE[0]` is unbound when bash reads from stdin under `set -u`. Landing this PR first would wire a suite into CI that **passes while the installer is broken**, turning an absent check into a false assurance. #2340 fixes the installer and adds the piped-stdin case the suite currently cannot reach; after that, this PR's CI wiring becomes genuinely protective.

Two things from review worth recording, neither blocking this PR:

- The parity `cmp` and `test:install`-inside-`validate` already existed from #2330, so local protection was in place — this PR's real contribution is the CI half, where `test.yml` hand-re-enumerates `validate`'s chain. Issue #2332 claimed otherwise and has been corrected.
- The `.gitattributes` `eol=lf` pin here is a sharp catch: the mirror is extensionless, so `*.sh eol=lf` never matched it and it fell through to `* text=auto` → CRLF on Windows. Without it this step could not pass on Windows even with the guard. Verified via fresh-clone `core.autocrlf=true` simulation: 264 CRLF lines at `dev`, 0 at this PR's head.

Issue #1433, opened 2026-04-27
feat(cli): per-run --provider/--model override for archon workflow run (and HTTP API parity)


Wirasm:
## Problem

There's no way today to override a workflow's provider/model for a single invocation. If a workflow YAML declares `provider: claude, model: sonnet`, every run uses Sonnet — even when the user wants to:

- run cheaper on Pi/Minimax for a one-off cost experiment,
- run on Opus instead of Sonnet to compare quality on a hard PR,
- fall back to Codex when Claude is rate-limited,
- isolate a model-specific bug.

Today's escape hatch is editing the YAML before each run and reverting after. That doesn't survive parallel runs and clutters git status.

### CLI

```bash
archon workflow run <name> [--provider <p>] [--model <m>] "<args>"
```

- `--provider <p>` alone — uses that provider's default model (e.g. `--provider claude` → `sonnet` per `.archon/config.yaml`).
- `--model <m>` alone — interpreted under the resolved provider (e.g. `--model haiku` on a Claude workflow). Loader rejects incompatible combinations (`haiku` on a Codex workflow).
- Both — full explicit override.
- Neither — current behavior (no behavior change for existing users).

### HTTP API parity

```http
POST /api/workflows/{name}/runs
{
  ...,
  "provider": "pi",
  "model": "minimax/MiniMax-M2.7"
}
```

Same semantics as the CLI flags. Keeps Web UI parity easy to add later.

## Resolution pipeline

```
CLI flag (or HTTP body field) set?
  YES → that value applies to every AI node in the run, period.
        Workflow-level AND per-node provider/model in YAML are IGNORED.
        Provider-specific fields that don't apply on the new provider
        (effort/thinking/betas/sandbox/hooks/mcp/skills on Pi, etc.)
        emit the existing loader warnings — same behavior as today.

  NO  → existing pipeline:
        per-node YAML > workflow YAML > .archon/config.yaml > SDK default
```

That's it. One pipeline, one priority order. No `--force-` flavor. The flag name (`--provider`, `--model`) already conveys force-semantics; a second flavor would introduce a bikeshed without solving a real problem.

## The deliberate tradeoff

Per-node `provider:` / `model:` fields in YAML are **ignored** when the CLI/API override is set. Workflow authors lose the ability to force a specific model for a specific node against an explicit override.

**Cost:** can't express "this node MUST be Opus" via YAML and still respect a user-issued override.

[...]

Issue #2333, opened 2026-07-30
fix(workflows): per-node token usage is written only to the JSONL log, never to workflow_events


Wirasm:
## Summary

Per-node token usage is captured for **every** provider and written to the run's **JSONL log file only**. It is never persisted to `remote_agent_workflow_events`, so no API consumer, console view, `workflow get`, or benchmark query can see it.

The data exists. It is just in the one place nothing queries.

## Evidence

`dag-executor.ts:1615` captures it from the provider result chunk:

```ts
if (msg.tokens) nodeTokens = msg.tokens;
```

`dag-executor.ts:2206-2209` writes it to the on-disk log:

```ts
await logNodeComplete(logDir, workflowRun.id, node.id, node.command ?? '<inline>', {
  durationMs: duration,
  tokens: nodeTokens,
});
```

`dag-executor.ts:2210` then persists the `node_completed` event — **without** it. Confirmed against a live run (`84b41900`, Codex `gpt-5.6-*`):

```
$ jsonl log     → "tokens":{"input":21586,"output":30}
                  "tokens":{"input":161524,"output":1348}
                  "tokens":{"input":22620,"output":123}

$ node_completed keys → ['duration_ms', 'node_output']
```

Zero of 36 `node_completed` events in a 3-run window carry a `tokens` field.

## Why it matters

**Token count is the only quantitative measure available from all four providers.** Every provider emits it and Archon maps it:

| Provider | emits tokens | where mapped |
| --- | --- | --- |
| Claude | yes | `claude/provider.ts` result chunk |
| Codex | yes | `codex/provider.ts:667` (`tokens: usage`) |
| Pi | yes | `pi/event-bridge.ts:163` (`usageToTokens(last.usage)`) |
| OpenCode | yes | `opencode/session.ts` result chunk |

Cost is **not** comparable — Claude and Pi report it, Codex's SDK has no cost field at all. Turn count is Claude-only. So tokens are the one axis on which two providers can be compared, and it is the one axis absent from the queryable store.

Concretely, this blocks:

- **Per-run cost/efficiency reporting** from the console or CLI — `workflow get` can show duration and nothing else on Codex or Pi.
- **The benchmarking half of #1433.** Running the same workflow across tiers is only useful if the results are comparable; today two runs can be compared on wall-clock alone.
[...]

PR #2038, opened 2026-06-26
feat(workflows,cli,paths): surface resolved model/tier in CLI node_started lines + one-time tier notice

Wirasm:
## Summary

- **Problem**: After the tier-keyword rewrite, bundled workflows use `small`/`medium`/`large` instead of literal model strings, but the CLI's `node_started` progress line only showed `[implement] Started` — no signal about which model actually runs. Unconfigured users got zero hint that tiers exist or how to customize them.
- **Why it matters**: CLI users are silently running workflows on models they didn't choose and can't see, which is especially confusing after the opus[1m] removal from defaults.
- **What changed**: (A) `NodeStartedEvent` now carries `provider`/`model`/`tier`, threaded from `resolveNodeProviderAndModel` through `executeNodeInternal` to the emitter + persisted event data. The CLI renders `[implement] Started  (claude/opus ← large)`. (B) A one-time-per-version tier notice prints to stderr on the first CLI run that hits unconfigured tier-keyword nodes, explaining defaults and how to configure them.
- **What did NOT change**: Model resolution is untouched — this is observability only. Web console UI wiring, loop-node model surfacing, and the `opus[1m]` default decision are out of scope.

### Before

```
  $ archon workflow run archon-fix-github-issue "fix #123"
  [extract-issue-number] Started
  [classify] Started
  [implement] Started          ← YAML says "large". No signal what model runs.
  [implement] Completed (4m2s)
```

### After

```
  $ archon workflow run archon-fix-github-issue "fix #123"

  ℹ️  This workflow uses model tiers (small/medium/large). You haven't configured them —
     using built-in defaults for 'claude':
       small   → claude/haiku
       medium  → claude/sonnet
       large   → claude/opus
     (Opus runs a 1M context window on API keys and Max/Team/Enterprise;
      on Pro it's 200K unless you set the `large` tier to `opus[1m]`.)
     Customize: `archon ai tier set <tier> <provider> <model>`
                or `tiers:` in .archon/config.yaml
     See anytime: `archon ai tier list`           (shown once per version)

  [extract-issue-number] Started  (claude/haiku ← small)
  [classify] Started  (claude/sonnet ← medium)
  [implement] Started  (claude/opus ← large)
  [implement] Completed (4m2s)
```

### Before

```
  dag-executor.ts
    resolveNodeProviderAndModel() → { provider, model, options }
    executeNodeInternal()
      emitter.emit({ type: 'node_started', runId, nodeId, nodeName })

  CLI workflow.ts
[...]

PR #2337, opened 2026-07-30
fix(providers): record resolved model metadata

### LINKED ISSUE #2314: fix(providers): resolved model is never recorded — Claude reads `model_usage` but the SDK field is `modelUsage`

Wirasm:
## Summary

- Problem: completed provider results did not retain a concrete model identifier, and Claude read the SDK's nonexistent `model_usage` field.
- Why it matters: model aliases can change, so workflow benchmarking could not reliably identify the model that actually ran.
- What changed: providers normalize a reported concrete model to `resolvedModel`; the DAG executor persists it with the requested model; coverage and architecture documentation were updated.
- What did **not** change (scope boundary): Codex does not synthesize a resolved model because its SDK completion event does not expose one.

### Before

```
Workflow node -> provider result -> node_completed event
requested alias ------------------> no concrete resolved model
```

### After

```
Workflow node -> provider result -> node_completed event
requested alias --> [resolvedModel] -> [requested + resolved]
```

### Before

```
Claude/Pi/OpenCode -> MessageChunk(modelUsage, provider-specific) -> DAG event
Codex -------------> MessageChunk -----------------------------> DAG event
```

### After

```
Claude/Pi/OpenCode -> [MessageChunk(resolvedModel)] ===> [DAG event { requested, resolved }]
Codex -------------> MessageChunk (no resolved model) -------> DAG event
```

| Claude/Pi/OpenCode providers | MessageChunk | **modified** | Normalized `resolvedModel` fact |
| MessageChunk | DAG executor | **modified** | Persists requested and resolved model together |
| DAG executor | workflow event store | **modified** | Existing JSON event persistence seam |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `workflows`

## Change Metadata

- Change type: `bug`
- Primary scope: `multi`

## Linked Issue

- Closes #2314
- Related #1433
- Depends on # (if stacked): none
- Supersedes # (if replacing older PR): none

Commands and result summary:

```bash
bun run type-check
bun run lint
bun run format:check
bun run test
```

- Evidence provided (test/log/trace/screenshot): focused Claude, Pi, OpenCode, and DAG regression tests plus the complete suite passed.

## Security Impact (required)

- If any `Yes`, describe risk and mitigation: n/a.

## Compatibility / Migration

- Backward compatible? Yes
- Config/env changes? No
- If yes, exact upgrade steps: n/a.

- Verified scenarios: Claude camel-case SDK metadata, Pi `responseModel`, OpenCode `modelID`, and DAG event persistence.
- Edge cases checked: absent provider metadata is omitted; Codex remains absent rather than using the requested model as a fallback.
- What was not verified: live provider SDK calls.

- Affected subsystems/workflows: provider result normalization and workflow completion events.
- Potential unintended effects: consumers previously reading provider-specific `model_usage` objects must use the normalized `{ requested, resolved }` value.
- Guardrails/monitoring for early detection: typed `ResolvedModel` contract and regression tests through persistence.

## Rollback Plan (required)

- Fast rollback command/path: revert commit `aeb7dade`.
- Observable failure symptoms: missing or malformed `model_usage` on completed AI-node events.

## Risks and Mitigations

- Risk: a provider may omit a concrete response model.
  - Mitigation: omit the field explicitly; never substitute requested intent for provider fact.

## Summary by CodeRabbit

* **New Features**
  * Added `resolvedModel` metadata to streamed `result`/terminal outputs when the provider can determine the concrete model.
  * Workflow node events now persist and emit both the requested model and the provider-resolved model, including loop nodes.
* **Documentation**
  * Updated the architecture reference to reflect the `resolvedModel` contract and provider omissions.
* **Bug Fixes**
  * Improved resolved-model selection and ambiguity handling, ensuring consistent metadata across success/error and structured output scenarios.
* **Tests**
  * Expanded coverage for resolved-model behavior across providers and workflow execution, including clearing stale resolved metadata.

### DIFF
--- a/packages/docs-web/src/content/docs/reference/architecture.md
+++ b/packages/docs-web/src/content/docs/reference/architecture.md
@@ -347,7 +347,9 @@ export type MessageChunk =
       cost?: number;
       stopReason?: string;
       numTurns?: number;
-      modelUsage?: Record<string, unknown>;
+      // Concrete provider-reported model. Omitted for providers such as Codex
+      // whose SDK completion events do not expose the resolved model.
+      resolvedModel?: ResolvedModel;
       // Session-resume outcome: true = restored, false = requested but fell back
       // to a fresh session, omitted = no resume requested. Set only when
       // resumeSessionId was passed (stamp it via withResumedOutcome).
--- a/packages/providers/src/claude/provider.test.ts
+++ b/packages/providers/src/claude/provider.test.ts
@@ -246,9 +246,9 @@ describe('ClaudeProvider', () => {
           num_turns: 3,
           modelUsage: {
             'claude-sonnet-4-6': {
-              input_tokens: 100,
-              output_tokens: 50,
-              cache_read_input_tokens: 10,
+              inputTokens: 100,
+              outputTokens: 50,
+              cacheReadInputTokens: 10,
             },
           },
         };
@@ -268,6 +268,60 @@ describe('ClaudeProvider', () => {
         numTurns: 3,
         resolvedModel: { id: 'claude-sonnet-4-6' },
       });
+      // Single-model usage is unambiguous — no ambiguity warning.
+      expect(mockLogger.warn).not.toHaveBeenCalled();
+    });
+
+    test('picks the greatest-output-token model and warns when modelUsage has multiple keys', async () => {
+      // A subagent pinned via `agents:` (or a fallbackModel takeover) puts more
+      // than one model in the record, and key order carries no guarantee — the
+      // main model here is deliberately NOT first.
+      mockQuery.mockImplementation(async function* () {
+        yield {
+          type: 'result',
+          session_id: 'sid-multi-model',
+          modelUsage: {
+            'claude-haiku-4-5-20251001': {
+              inputTokens: 400,
+              outputTokens: 20,
+              cacheReadInputTokens: 0,
+            },
+            'claude-sonnet-5': {
+              inputTokens: 120,
+              outputTokens: 900,
+              cacheReadInputTokens: 10,
+            },
+          },
+        };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks[0]).toMatchObject({ resolvedModel: { id: 'claude-sonnet-5' } });
+      expect(mockLogger.warn).toHaveBeenCalledWith(
+        {
+          models: ['claude-haiku-4-5-20251001', 'claude-sonnet-5'],
+          selected: 'claude-sonnet-5',
+        },
+        'claude.resolved_model_ambiguous'
+      );
+    });
+
+    test('omits resolvedModel when modelUsage is an empty record', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield { type: 'result', session_id: 'sid-empty-usage', modelUsage: {} };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks[0]).not.toHaveProperty('resolvedModel');
+      expect(mockLogger.warn).not.toHaveBeenCalled();
     });
 
     test('omits cost, stopReason, numTurns, and resolvedModel when SDK result has none', async () => {
--- a/packages/providers/src/claude/provider.test.ts
+++ b/packages/providers/src/claude/provider.test.ts
@@ -236,7 +236,7 @@ describe('ClaudeProvider', () => {
       });
     });
 
-    test('yields result with cost, stopReason, numTurns, modelUsage when SDK provides them', async () => {
+    test('yields result with cost, stopReason, numTurns, and a resolved model when SDK provides them', async () => {
       mockQuery.mockImplementation(async function* () {
         yield {
           type: 'result',
@@ -244,7 +244,7 @@ describe('ClaudeProvider', () => {
           total_cost_usd: 0.0042,
           stop_reason: 'end_turn',
           num_turns: 3,
-          model_usage: {
+          modelUsage: {
             'claude-sonnet-4-6': {
               input_tokens: 100,
               output_tokens: 50,
@@ -266,17 +266,11 @@ describe('ClaudeProvider', () => {
         cost: 0.0042,
         stopReason: 'end_turn',
         numTurns: 3,
-        modelUsage: {
-          'claude-sonnet-4-6': {
-            input_tokens: 100,
-            output_tokens: 50,
-            cache_read_input_tokens: 10,
-          },
-        },
+        resolvedModel: { id: 'claude-sonnet-4-6' },
       });
     });
 
-    test('omits cost, stopReason, numTurns, modelUsage when SDK result has none', async () => {
+    test('omits cost, stopReason, numTurns, and resolvedModel when SDK result has none', async () => {
       mockQuery.mockImplementation(async function* () {
         yield { type: 'result', session_id: 'sid-bare' };
       });
@@ -289,7 +283,7 @@ describe('ClaudeProvider', () => {
       expect(chunks[0]).not.toHaveProperty('cost');
       expect(chunks[0]).not.toHaveProperty('stopReason');
       expect(chunks[0]).not.toHaveProperty('numTurns');
-      expect(chunks[0]).not.toHaveProperty('modelUsage');
+      expect(chunks[0]).not.toHaveProperty('resolvedModel');
     });
 
     test('omits stopReason when stop_reason is null', async () => {
--- a/packages/providers/src/claude/provider.ts
+++ b/packages/providers/src/claude/provider.ts
@@ -35,6 +35,7 @@ import {
   type HookCallbackMatcher,
   type SDKAssistantMessageError,
   type SDKResultMessage,
+  type ModelUsage,
 } from '@anthropic-ai/claude-agent-sdk';
 import type {
   IAgentProvider,
@@ -86,6 +87,44 @@ function normalizeClaudeUsage(usage?: {
     output,
     ...(typeof total === 'number' ? { total } : {}),
   };
+}
+
+/**
+ * Pick the concrete model that did the bulk of a turn's work from the SDK's
+ * per-model usage record.
+ *
+ * More than one entry is reachable for a single turn: a subagent pinned to
+ * another model via `agents:`, or a `fallbackModel` takeover. Key insertion
+ * order happens to put the main model first today, but nothing in the SDK
+ * guarantees it — so select by greatest output-token count (the main model
+ * produces the bulk of the output) and WARN whenever the record is ambiguous,
+ * so a multi-model turn is visible instead of silently collapsed.
+ *
+ * `modelUsage` is non-optional in the SDK types but arrives over an IPC
+ * boundary, so the absent/empty cases stay guarded — absence yields undefined
+ * and the caller omits `resolvedModel` entirely rather than inventing a value.
+ * On a tie (or output counts the SDK didn't send) the first key wins, which is
+ * exactly the pre-#2314 behavior — safe, and the warning still fires.
+ */
+function selectResolvedModelId(
+  modelUsage: Record<string, ModelUsage> | undefined
+): string | undefined {
+  if (!modelUsage) return undefined;
+  const entries = Object.entries(modelUsage);
+  if (entries.length === 0) return undefined;
+  if (entries.length === 1) return entries[0][0];
+
+  const outputTokensOf = (usage: ModelUsage): number =>
+    Number.isFinite(usage.outputTokens) ? usage.outputTokens : 0;
+  let selected = entries[0];
+  for (const entry of entries.slice(1)) {
+    if (outputTokensOf(entry[1]) > outputTokensOf(selected[1])) selected = entry;
+  }
+  getLog().warn(
+    { models: entries.map(([id]) => id), selected: selected[0] },
+    'claude.resolved_model_ambiguous'
+  );
+  return selected[0];
 }
 
 /**
@@ -992,9 +1031,7 @@ async function* streamClaudeMessages(
       yield { type: 'rate_limit', rateLimitInfo: rateLimitMsg.rate_limit_info ?? {} };
     } else if (event.type === 'result') {
       const resultMsg = msg as SDKResultMessage;
-      const resolvedModelId = resultMsg.modelUsage
-        ? Object.keys(resultMsg.modelUsage)[0]
-        : undefined;
+      const resolvedModelId = selectResolvedModelId(resultMsg.modelUsage);
       // The terminal result resolves any recorded synthetic error message.
       const syntheticError = pendingSdkError;
       pendingSdkError = undefined;
--- a/packages/providers/src/claude/provider.ts
+++ b/packages/providers/src/claude/provider.ts
@@ -34,7 +34,7 @@ import {
   type HookCallback,
   type HookCallbackMatcher,
   type SDKAssistantMessageError,
-  type TerminalReason,
+  type SDKResultMessage,
 } from '@anthropic-ai/claude-agent-sdk';
 import type {
   IAgentProvider,
@@ -991,40 +991,21 @@ async function* streamClaudeMessages(
       getLog().warn({ rateLimitInfo: rateLimitMsg.rate_limit_info }, 'claude.rate_limit_event');
       yield { type: 'rate_limit', rateLimitInfo: rateLimitMsg.rate_limit_info ?? {} };
     } else if (event.type === 'result') {
-      const resultMsg = msg as {
-        session_id?: string;
-        is_error?: boolean;
-        subtype?: string;
-        usage?: { input_tokens?: number; output_tokens?: number; total_tokens?: number };
-        structured_output?: unknown;
-        total_cost_usd?: number;
-        stop_reason?: string | null;
-        num_turns?: number;
-        errors?: string[];
-        result?: string;
-        terminal_reason?: TerminalReason;
-        api_error_status?: number | null;
-        model_usage?: Record<
-          string,
-          {
-            input_tokens: number;
-            output_tokens: number;
-            cache_read_input_tokens?: number;
-            cache_creation_input_tokens?: number;
-          }
-        >;
-      };
+      const resultMsg = msg as SDKResultMessage;
+      const resolvedModelId = resultMsg.modelUsage
+        ? Object.keys(resultMsg.modelUsage)[0]
+        : undefined;
       // The terminal result resolves any recorded synthetic error message.
       const syntheticError = pendingSdkError;
       pendingSdkError = undefined;
       const tokens = normalizeClaudeUsage(resultMsg.usage);
-      const sdkErrors = Array.isArray(resultMsg.errors) ? resultMsg.errors : undefined;
+      const sdkErrors = 'errors' in resultMsg ? resultMsg.errors : undefined;
 
       // `is_error: true` + `subtype: 'success'` is ambiguous: it is BOTH the
       // SDK's stop-sequence termination encoding (#1425, a legitimate success)
       // AND its API-failure-as-text encoding (#1797 — auth/billing/rate-limit
       // errors that even set stop_reason: 'stop_sequence').
-      const isSuccessWithErrorFlag = resultMsg.is_error === true && resultMsg.subtype === 'success';
+      const isSuccessWithErrorFlag = resultMsg.is_error && resultMsg.subtype === 'success';
 
       // Disambiguate structurally: a preceding synthetic error message
       // (primary, typed signal), or the typed terminal_reason 'api_error'
@@ -1057,7 +1038,7 @@ async function* streamClaudeMessages(
       // Fail-safe (never observed in practice): a synthetic error message
       // followed by a non-error result. Yield the withheld text late rather
       // than silently swallowing content.
-      if (syntheticError !== undefined && resultMsg.is_error !== true) {
+      if (syntheticError !== undefined && !resultMsg.is_error) {
         getLog().warn(
           { sessionId: resultMsg.session_id, errorCode: syntheticError.code },
           'claude.synthetic_error_not_confirmed'
@@ -1071,7 +1052,7 @@ async function* streamClaudeMessages(
       // subtype: 'success' — its encoding of "non-default termination, not a
       // failure". Treat that pair as a clean success so downstream consumers
       // (which gate failure on isError) don't misclassify it.
-      const isRealError = resultMsg.is_error === true && !isSuccessWithErrorFlag;
+      const isRealError = resultMsg.is_error && !isSuccessWithErrorFlag;
       if (isRealError) {
         getLog().error(
           {
@@ -1095,7 +1076,7 @@ async function* streamClaudeMessages(
         type: 'result',
         sessionId: resultMsg.session_id,
         ...(tokens ? { tokens } : {}),
-        ...(resultMsg.structured_output !== undefined
+        ...('structured_output' in resultMsg && resultMsg.structured_output !== undefined
           ? { structuredOutput: resultMsg.structured_output }
           : {}),
         ...(isRealError ? { isError: true, errorSubtype: resultMsg.subtype } : {}),
@@ -1103,9 +1084,7 @@ async function* streamClaudeMessages(
         ...(resultMsg.total_cost_usd !== undefined ? { cost: resultMsg.total_cost_usd } : {}),
         ...(resultMsg.stop_reason != null ? { stopReason: resultMsg.stop_reason } : {}),
         ...(resultMsg.num_turns !== undefined ? { numTurns: resultMsg.num_turns } : {}),
-        ...(resultMsg.model_usage
-          ? { modelUsage: resultMsg.model_usage as Record<string, unknown> }
-          : {}),
+        ...(resolvedModelId ? { resolvedModel: { id: resolvedModelId } } : {}),
       };
     }
   }
--- a/packages/providers/src/community/opencode/provider.test.ts
+++ b/packages/providers/src/community/opencode/provider.test.ts
@@ -304,12 +304,7 @@ describe('OpencodeProvider', () => {
         tokens: { input: 11, output: 7, total: 21, cost: 0.42 },
         cost: 0.42,
         stopReason: 'stop',
-        modelUsage: {
-          providerID: 'anthropic',
-          modelID: 'claude-sonnet',
-          reasoning: 3,
-          cache: 1,
-        },
+        resolvedModel: { id: 'claude-sonnet' },
       },
     ]);
   });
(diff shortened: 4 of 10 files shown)


Wirasm:

**Status**: COMPLETE
**Pushed**: ✅ Changes pushed to `archon/task-archon-fix-github-issue-experimental-1785410386074`

---

### Fixes Applied (4 total)

| Severity | Count |
|----------|-------|
| 🔴 CRITICAL | 0 |
| 🟠 HIGH | 4 |
| 🟡 MEDIUM | 0 |
| 🟢 LOW | 0 |

<details>

- ✅ **No reviewable implementation or PR** (`nodes/implement.md:1`) — preserved the unrelated dirty edit, implemented #2314, and opened this PR.
- ✅ **Claude model field mismatch** (`packages/providers/src/claude/provider.ts:994`) — uses `SDKResultMessage.modelUsage` and normalizes its concrete model id.
- ✅ **Cross-provider metadata drift** (`packages/providers/src/types.ts:183`) — added typed `resolvedModel` metadata for Claude, Pi, and OpenCode.
- ✅ **Incomplete completion event attribution** (`packages/workflows/src/dag-executor.ts:1389`) — persists requested and resolved models together.

</details>

---

### Tests Added

- Claude, Pi, OpenCode, and DAG persistence regression coverage.

---

### Skipped (0)

---

*(none)*

---

### Validation

✅ Type check | ✅ Lint | ✅ Format check | ✅ Full isolated package test suite

---

*Self-fix by Archon · aggressive mode · fixes pushed to `archon/task-archon-fix-github-issue-experimental-1785410386074`*

Wirasm:
---
pr: 2337
title: "fix(providers): record resolved model metadata"
author: "Wirasm"
reviewed: 2026-07-30
---

# PR Review: #2337 — fix(providers): record resolved model metadata

**Branch**: `archon/task-archon-fix-github-issue-experimental-1785410386074` -> `dev`
**Commit**: `aeb7dade`
**Files Changed**: 10 (+69/-81)
**Closes**: #2314

---

## Summary

The core fix is correct and I verified it against the live SDK, not just the mocks. `modelUsage` is
now read with the right casing, the hand-written result cast is gone, and — the most valuable part —
the SDK type now makes the original class of bug a **compile error**. Absence is genuinely preserved:
nothing anywhere synthesises the requested model, and Codex is left untouched rather than faked.

Three things hold it back. One AI node type (`loop:`) still records nothing at all, so the issue's
own acceptance criterion is only partly met. The Claude extraction picks one model out of N with no
warning, and I verified N>1 is reachable. And the persisted key is still called `model_usage` while
now holding model *identity*, with the per-model usage/cost payload discarded.

---

## Verified by execution (not inference)

| # | Check | Result |
|---|-------|--------|
| 1 | Pre-fix baseline in `~/.archon/archon.db` | **0** of **488** `node_completed` events carry `model_usage` — confirms the issue's claim |
| 2 | Live Claude SDK result message (0.3.209) | `modelUsage` **present**, `model_usage` **absent**. Value: `{"claude-haiku-4-5-20251001": {...}}` |
| 3 | Does the SDK type catch the old read? | **Yes** — `TS2551: Property 'model_usage' does not exist on type 'SDKResultMessage'. Did you mean 'modelUsage'?` |
| 4 | Can `modelUsage` hold multiple keys? | **Yes** — subagent pinned to a different model produced `["claude-haiku-4-5-20251001","claude-sonnet-5"]` |
| 5 | Does a resumed/forked session inherit prior `modelUsage`? | **No** — pass 2 (sonnet, `resume`+`forkSession`) returned only `["claude-sonnet-5"]`. Rules out a failure mode I had hypothesised |
| 6 | Does the value reach the **database**? | **Yes** — `deps.store.createWorkflowEvent` at `dag-executor.ts:2212`, not the JSONL logger. Not the #2333 fate |
| 7 | Do the new tests fail on `dev`? | **Yes** — reverted the 4 source files to `origin/dev`, kept the PR's tests: **6 failures** incl. the DB-persistence assertion |
| 8 | Other hand-written SDK casts in `provider.ts` | Field names **all match** the SDK — no second live instance of the bug |
| 9 | Validation | type-check PASS, lint PASS, format PASS, providers 203/203, `@archon/workflows` all batches 0 fail |

---

## Issues Found

### Critical

No critical issues.

### Important

**I1 — `loop:` nodes still record no model at all. `dag-executor.ts:4142-4166`**

`executeLoopNode` (3715–4725) does not go through `executeNodeInternal`. It has its own
`aiClient.sendQuery` at `:4060` and its own result-chunk consumption at `:4142-4166`, which reads
`msg.sessionId`, `msg.cost`, `msg.tokens`, `msg.stopReason`, `msg.numTurns`,
`msg.structuredOutput` — and **not** `msg.resolvedModel`. Its terminal `node_completed`
(`:4566-4576`) therefore carries `cost_usd`, `stop_reason`, `num_turns` and no model.

Worse, `executeLoopNode`'s `node_started` (`:3756-3762`) writes only
`{ type: 'loop', command }` — no `model`, no `tier`. So a loop node has **neither** half of the
record: not the intent (that gap is #2038's, pre-existing) and not the fact.

`loop_group:` is fine — bodies run through `runLayers` → `executeNodeInternal`, so body nodes get
`model_usage`; only the wrapper aggregate row lacks it, which is acceptable.

*Failure scenario*: benchmark a review workflow whose AI work lives in a `loop:` node. Every row has
a blank resolved model — the exact original symptom, in a node type the issue's acceptance criterion
covers ("a `node_completed` for an AI node carries a concrete model id"). Nothing fails, because
absence is what nobody asserts.

*Honest scoping*: in this install only 4 of 104 runs contain a loop node, and 0 of the 43 AI
`node_completed` rows carrying `num_turns` come from one. Present-day volume is low. But `loop:` is a
documented first-class primitive and the fix is the same one line as `:1620`.

---

**I2 — `Object.keys(modelUsage)[0]` picks one of N models silently. `claude/provider.ts:995-997`**

```ts
const resolvedModelId = resultMsg.modelUsage ? Object.keys(resultMsg.modelUsage)[0] : undefined;
```

I verified two keys are reachable: a node with `agents:` pinned to a different model produced
`["claude-haiku-4-5-20251001","claude-sonnet-5"]`. In that run `[0]` happened to be the main model
(it makes the first API call, and insertion order follows), so the pick is *usually* right — but
nothing enforces that, documents it, or warns when N>1. Archon exposes both `agents:` (per-agent
`model`) and `fallbackModel`, so multi-model runs are a supported configuration, not an edge case.

*Failure scenario (verified reachable)*: a bench row names one model for a run that used two, with no
signal that anything was dropped. *Unverified but plausible*: when `fallbackModel` engages on
overload, if the primary's failed attempt is recorded first, `[0]` returns the model that did **not**
do the work — I could not force an overload to test this.

*Fix*: pick deterministically (the entry with max `outputTokens` — the model that did the work) and
`getLog().warn` when `Object.keys(...).length > 1`. A silent arbitrary pick in a field whose whole
purpose is trustworthy attribution repeats the issue's own thesis: a number is believed.

---

**I3 — the persisted key is still `model_usage` but no longer holds usage. `dag-executor.ts:2223-2225`**

```ts
...(nodeResolvedModel ? { model_usage: { requested: resolvedModel, resolved: nodeResolvedModel.id } } : {})
```

Two problems under one name:

1. **The name now lies.** `model_usage` holds model *identity*. `node_started` already uses the field
   name `model` for the same concept (`:1266`). A store can now hold two incompatible shapes under
   one key with no discriminator — OpenCode *did* persist the old `{providerID, modelID, reasoning,
   cache}` shape on any install that ran it (empirically 0 such rows here, so blast radius is near
   zero today, and no in-repo consumer reads it — I grepped). This is free to fix now and annoying
   once rows accumulate. Rename to `model` or `resolved_model`.
2. **Information is discarded.** The SDK's `ModelUsage` carries `inputTokens`, `outputTokens`,
   `cacheReadInputTokens`, `cacheCreationInputTokens`, **`costUSD`**, `contextWindow`,
   `maxOutputTokens` — plus, at runtime, `canonicalModel` and `provider`, which the installed `.d.ts`
   does not even declare. My probe returned `canonicalModel: "claude-haiku-4-5"` alongside the dated
   `claude-haiku-4-5-20251001` key. Per-model cost is precisely what #1433's benchmarking use case
   wants, and `canonicalModel` is arguably the better grouping key for bench rows than the dated
   snapshot. The old code forwarded the whole map; this reduces it to one string.

The issue asked to normalise the resolved model to one shape. It did not ask to throw the usage
breakdown away — see the spec note below.

---

### Suggestions

**S1 — `nodeResolvedModel` is not reset between reask passes. `dag-executor.ts:1428-1433, :1620`**

CodeRabbit flagged this and it is real, but narrower than stated. `runNodeRetryLoop` calls
`executeNodeInternal` fresh per retry, so retries are safe. The only window is the best-effort
structured-output **reask** loop (Pi/Copilot), where `runStreamPass` runs repeatedly inside one
`executeNodeInternal` call. That block explicitly resets its per-attempt accumulators —
*"so a prior reask attempt's state never leaks into this one"* — and the new field was not added to
it. `nodeStopReason`/`nodeNumTurns`/`nodeTokens` already have the same latent staleness, so this is
consistent with existing behaviour rather than a new class of bug; it just happens to be the field
where a stale value silently misattributes a benchmark. One line: `nodeResolvedModel = undefined;` in
the reset block. (I'd prefer that over CodeRabbit's "assign on every chunk including undefined" —
it matches the surrounding convention.)

**S2 — two near-identical names with opposite meanings on one line. `dag-executor.ts:2224`**

`resolvedModel` (the function parameter, `:1238`) is the **requested** model. `nodeResolvedModel` is
the **resolved** one. They appear together in `{ requested: resolvedModel, resolved: nodeResolvedModel.id }`.
Rename the parameter to `requestedModel`, or the local to `providerReportedModel`. Someone will swap
these, and the resulting record will look perfectly plausible.

**S3 — test gaps, all on the axis the issue said matters**

The failure mode here is absence, and the highest-value guard is the one missing:

- No executor-level test that `model_usage` is **absent** when the provider reports nothing — the
  Codex path, i.e. the exact thing the issue forbids synthesising. Provider-level absence is tested
  for Claude/Pi/OpenCode, but the `? :` false branch at `:2223` — which decides whether `requested`
  gets written at all — is untested.
- No test for multi-key `modelUsage` (verified reachable).
- No test for `modelUsage: {}` — truthy, so `[0]` is `undefined`. Runtime behaviour is correct
  (field omitted); it is just unguarded.
- The DB assertion was appended to an existing unrelated test (`'stores node_output in
  node_completed event data for AI nodes'`) rather than getting its own case.

**S4 — `requested` silently vanishes when no model is configured. `dag-executor.ts:974, :2224`**

`resolveNodeProviderAndModel` returns `model: undefined` when node, workflow, and config all omit a
model (provider SDK default applies). `JSON.stringify` then drops the key, yielding
`{"resolved": "..."}` with no `requested`. That is correct as "absence renders as absence", but the
object type `{ requested: string | undefined; resolved: string }` gives no compile-time signal, and
the pair the issue wanted written together is half-present. Worth an explicit comment at minimum.

**S5 — the live emitter does not carry it. `dag-executor.ts:2241-2250`**

The DB row gets the model; the in-process `node_completed` emitter event does not, while
`cost`/`stopReason`/`numTurns` are on both. SSE/console consumers watching a live run won't see it
until they refetch from the DB. Probably fine given the console normalizer reads persisted rows, but
it is an asymmetry worth a deliberate decision rather than an omission.

**S6 — the Codex limitation lives only in a type comment**

`architecture.md:349-351` documents it on the `MessageChunk` type. The project's own convention for
per-provider differences is the generated capability matrix (`bun run generate:capability-matrix` →
`reference/provider-capabilities.md`), which is generated from `capabilities.ts` specifically so docs
cannot drift from runtime. A `reportsResolvedModel` axis would make the gap discoverable to the user
comparing bench rows across providers. YAGNI tension acknowledged — but that user is the live
consumer.

**S7 — Copilot was not considered (and the issue omitted it too)**

`community/copilot/event-bridge.ts:381-390` builds a result chunk with no `resolvedModel`. Copilot is
a registered provider. Either read a response model if the SDK exposes one, or document it alongside
Codex. This is a gap in the **spec**, not only the PR.

**S8 — minor robustness regression bundled into the diff. `claude/provider.ts:1001`**

`Array.isArray(resultMsg.errors)` became `'errors' in resultMsg`. Type-correct — only
`SDKResultError` declares `errors`, and my probe confirms success results omit the key entirely — but
it drops a runtime shape guard on external SDK output. Low risk; flagging only because it is an
unrelated behaviour change riding along.

---

## Where the issue's spec was wrong

I wrote the issue being graded against, so these are spec defects, not PR defects:

1. **"OpenCode: works"** overstates it. OpenCode emitted a `modelUsage` object whose every field was
   `undefined` when no assistant info was captured — see the two test expectations the PR deletes
   (`{providerID: undefined, modelID: undefined, reasoning: undefined, cache: undefined}`). That is a
   populated-looking key with no content, which is worse than absence. The PR cleanly omits it now.
   Good catch by the implementation, not the spec.
2. **"Normalise to one shape"** did not say what to do with the per-model usage payload. A literal
   reading led to discarding it (I3.2). The criterion should have read "normalise the resolved model
   **and** preserve the usage breakdown", or explicitly blessed dropping it.
3. **The table omits Copilot** (S7).
4. **"keys are concrete model ids"** is written as though there would be one. The spec set up the
   arbitrary `[0]` pick by never saying what to do with N>1 (I2).
5. **"Write them as one fact, in one call"** is weaker here than the `resolvedBaseFields` precedent
   it cites. In that precedent neither half existed elsewhere. Here `requested` is a **copy** of
   `node_started.model` (#2038), so writing it again creates a *new* drift surface between two events
   rather than removing one. The PR followed the instruction correctly; the instruction was arguable.
   Co-locating the pair is still defensible for query ergonomics — but it should be justified that
   way, not as drift prevention.

---

## Acceptance criteria from #2314

| Criterion | Status |
|---|---|
| Claude reads the SDK's `modelUsage`; AI `node_completed` carries a concrete model id | **Met** for `command:`/`prompt:`/approval-`on_reject`; **not met** for `loop:` (I1) |
| Replace the hand-written cast with `SDKResultMessage`, audit the file's other SDK casts | Replacement **met and verified** (TS2551). Audit **not done** — 4 other hand-written casts remain (`sysMsg` at `:870`, `rateLimitMsg` at `:990`, …) with exact SDK counterparts (`SDKTaskStartedMessage`, `SDKHookResponseMessage`, `SDKRateLimitEvent`, …). I checked every field name: all currently match, so this is latent risk, not a live bug |
| Pi: read `AssistantMessage.responseModel` | **Met** — confirmed the field exists at `pi-ai/dist/types.d.ts:201` |
| Normalise to one shape at the provider boundary | **Met** for the chunk (`ResolvedModel {id}`); the persisted key name and the discarded usage map are the residue (I3) |
| Codex: document, do not fall back to the requested model | **Met** — no synthesis anywhere. Documented only in a type comment (S6) |
| Regression test asserting a concrete id reaches the persisted event | **Met and verified** — fails on `dev` |

---

## Validation Results

| Check | Status | Details |
|-------|--------|---------|
| Type Check | PASS | all 9 packages, exit 0 |
| Lint | PASS | exit 0, 0 warnings |
| Format | PASS | exit 0 |
| Tests — providers | PASS | 203 pass / 0 fail (claude, pi, opencode) |
| Tests — workflows | PASS | all 18 batches, 0 fail |
| Regression value | VERIFIED | 6 failures when sources reverted to `dev` |

---

## Strengths

- **The structural fix is the real win and it demonstrably works.** `SDKResultMessage` turns the
  original bug into `TS2551` at compile time. That is what the issue actually asked for and it is
  correctly delivered.
- **Absence is genuinely preserved.** I looked for a back-fill on every path and there is none.
  Codex is untouched rather than faked — the easy, wrong thing was available and not taken.
- **The OpenCode all-`undefined` object is gone.** Not asked for; a real improvement.
- **Verified regression guards.** The tests fail on `dev`, including the one that proves the value
  reaches the database.
- **Scope discipline.** 10 files, +69/-81 — net negative, which is the right shape for replacing a
  hand-written type with the vendor's.

---

## Recommendation

**REQUEST CHANGES**

Blocking: **I1** (`loop:` nodes record nothing — same absence, unfixed) and **I2** (silent arbitrary
pick among N models, verified reachable). Both are a few lines.

Strongly recommended before merge: **I3.1** (rename `model_usage` → `model`/`resolved_model`) —
trivial now, migration-shaped later. **I3.2** (keep the per-model cost breakdown) is a judgement call
worth making deliberately rather than by omission.

`S1`–`S8` can follow up.

---

*Reviewed by Claude — findings above are separated into execution-verified and reasoned.*
*Report: `/Users/rasmus/.prp/archon-75601ef6/reviews/pr-2337-review.md`*

Wirasm:
Reviewed. **NEEDS FIXES** — two blocking, one strongly recommended. The core of this PR is right and the most valuable part is the type change, not the field rename: compiling `m.model_usage` against `SDKResultMessage` now produces `TS2551: Property 'model_usage' does not exist … Did you mean 'modelUsage'?` So the class of bug is closed, not just the instance.

Verified independently: the field name is fixed against the **live SDK** (a real query returns `modelUsage`, no `model_usage`); absence is preserved on every path with no back-fill; Codex is left untouched rather than faked; the value reaches `createWorkflowEvent` at `dag-executor.ts:2212` so it does not share #2333's log-only fate; and the tests genuinely fail against `origin/dev` — reverting the four source files while keeping the PR's tests produces **6 failures**, including the persistence assertion. Baseline confirmed at **0 of 488** `node_completed` events carrying the field.

## Blocking

**B1 — `loop:` nodes still record nothing.** `executeLoopNode` never routes through `executeNodeInternal`; it has its own `sendQuery` at `:4060` and its own chunk loop at `:4142-4166`, which reads `cost`, `tokens`, `stopReason` and `numTurns` but **not** `resolvedModel`. Its `node_started` at `:3756` also carries no `model`/`tier`, so a loop node has neither half of the record. `loop_group` is fine — bodies go through `runLayers` → `executeNodeInternal`. The fix is the same one line as `:1620`. I confirmed this by reading the branch, not from the review.

**B2 — `Object.keys(modelUsage)[0]` silently picks one of N.** `N > 1` is reachable and was demonstrated: a subagent pinned to a different model yields `["claude-haiku-4-5-20251001", "claude-sonnet-5"]`. `[0]` happened to be the main model, but nothing enforces that ordering, nothing documents it, and nothing warns. Archon exposes both per-model `agents:` and `fallbackModel`, so this is a supported configuration, not a corner case. Pick the entry with the greatest `outputTokens` and warn when more than one key is present — an arbitrary pick that is usually right is the failure mode this issue was filed about.

## Strongly recommended

**B3 — the key is still named `model_usage` while now holding identity, and the payload is discarded.** `node_started` already uses `model` for this concept. More importantly the SDK's per-model value carries `costUSD`, a token breakdown, and `canonicalModel` (`"claude-haiku-4-5"`) — which the installed `.d.ts` does not even declare. Per-model cost is precisely what #1433's benchmarking half needs, and #2333 exists because that data is currently unreachable. Discarding it here means re-plumbing it later.

## Not blocking

The remaining four hand-written casts in `claude/provider.ts` (`sysMsg` at `:870`, `rateLimitMsg` at `:990`) were not audited, which the issue asked for. Every field name in them currently matches the SDK, so it is latent risk rather than a second live bug — worth a follow-up, not a change here.

CodeRabbit's staleness finding at `:1620` is real but narrower than stated: retries re-enter `executeNodeInternal` via `runNodeRetryLoop`, so only the Pi/Copilot reask loop leaks — and `nodeStopReason`/`nodeNumTurns` already leak identically. A suggestion, not a defect introduced here.

## Five corrections to #2314 — my spec was wrong, not this PR

Recording these because the PR followed the spec faithfully and the spec misled it:

1. **"OpenCode: works"** overstated it. OpenCode was emitting `{providerID: undefined, modelID: undefined, …}` — a populated-looking key with no content. This PR correctly omits it now.
2. **Copilot is missing from the table entirely** and has the same gap.
3. **"keys are concrete model ids"** reads as though there would be one. That phrasing is what set up B2's arbitrary `[0]`.
4. **"normalise to one shape"** never said to preserve the usage breakdown, so a literal reading discarded it. B3 is the consequence of my wording.
5. **"one call so they cannot drift"** is weaker than I claimed. `requested` is a *copy* of `node_started.model` (#2038), so co-locating creates a **new** drift surface rather than removing one. It is still defensible for query ergonomics — just not as drift prevention, and citing `resolvedBaseFields` as precedent was wrong.

#2314 will be corrected. B1 and B2 stand on their own merits regardless.

Issue #2295, opened 2026-07-28
Quick installation failing for MAC


mebrah2:
## Summary
Installation + Crash 
 %  curl -fsSL https://archon.diy/install | bash

  ╔═══════════════════════════════════════╗
  ║      Archon CLI Installer             ║
  ╚═══════════════════════════════════════╝

[INFO] Detecting platform...
[OK] Platform: darwin-x64
[INFO] Version: latest
[INFO] Download URL: https://github.com/coleam00/Archon/releases/latest/download/archon-darwin-x64
[INFO] Downloading binary...
[OK] Downloaded successfully
[INFO] Verifying checksum...
[OK] Checksum verified
[INFO] Installing to /usr/local/bin/archon...
[OK] Installed to /usr/local/bin/archon

[INFO] Verifying installation...
[WARN] Binary installed but version check failed:

[WARN] The binary may not work correctly. Please verify manually with: /usr/local/bin/archon version

Get started:
  archon workflow list
  archon workflow run assist "What workflows are available?"

- What broke:
- binary 
- When it started (if known): yes 
- Severity: `blocker|major|minor|cosmetic`
blocker 
## Steps to Reproduce

1.Run install comand 
2.you will se warning 
3. run 'archon workflow list' you will get
[1]    97367 illegal hardware instruction  archon workflow list

## Expected vs Actual

- **Expected**: when you run '% archon workflow list'  list of archon workflow list 
- **Actual**: [1]    97367 illegal hardware instruction  archon workflow list

## User Flow

```
(Draw the flow that triggers the bug. Mark where it breaks with [X].)

Example:
  User                   Archon                   AI Client
  ────                   ──────                   ─────────
  sends /plan ─────────▶ routes to workflow
                         creates worktree
                         streams to AI ──────────▶ processes prompt
                         [X] timeout waiting ◀──── no response
  sees error ◀────────── sends error message
```

## Environment

- Platform: (Slack / Telegram / GitHub / Discord / Web / CLI)
- Database: (SQLite / PostgreSQL)
- Running in worktree? (`Yes/No`)
- OS:

## Logs

```
Paste relevant logs here (redact any tokens/secrets)
```

## Impact

- Affected workflows/commands:
- Reproduction rate: Always / Intermittent / Once
- Workaround available? If so, describe:
- Data loss risk? (`Yes/No`)

## Scope

- Package(s) likely involved: `core|workflows|isolation|git|adapters|server|web|cli|paths`
[...]

PR #2340, opened 2026-07-30
fix(install): repair the curl|bash entry point and sync the PowerShell mirror

### LINKED ISSUE #2338: fix(install): curl | bash installer exits 1 for all users — BASH_SOURCE unbound under set -u (regression from #2330)
## Summary

**`curl -fsSL https://archon.diy/install | bash` exits 1 and installs nothing — for every user, on every platform.** Introduced by #2330 (squash `08723171`), currently on `dev`.

**Not live yet.** `deploy-docs.yml` triggers on `main` + `paths: ['packages/docs-web/**']`, and `curl https://archon.diy/install` still serves main's pre-#2330 version (verified: 0 matches for `BASH_SOURCE`). It ships **the moment `dev` merges to `main`** — and because `packages/docs-web/**` changed, that deploy will fire.

**This is a release blocker for `dev` → `main`.**

## Cause

`scripts/install.sh:262` and `packages/docs-web/public/install:262` (identical mirrors) added:

```sh
if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
  main "$@"
fi
```

Line 23 sets `set -euo pipefail`. When bash reads the script from **stdin** — which is exactly what `curl … | bash` does — `BASH_SOURCE[0]` is **unbound**, so `set -u` aborts before `main` is ever called:

```
$ printf 'set -euo pipefail\nif [[ "${BASH_SOURCE[0]}" == "$0" ]]; then echo REACHED_MAIN; fi\n' | bash
bash: line 2: BASH_SOURCE[0]: unbound variable
exit 1
```

Reproduced on macOS system bash **3.2.57** and on bash **5.2** — not version-specific.

Net effect: #2295 went from *"installs but the binary won't run on one Mac configuration"* to *"the installer hard-fails for everyone."* Strictly worse than the bug it fixed.

## Why it exists, which is the instructive part

The guard has no production purpose. It exists **solely** so `scripts/test-install.sh` can `source` the installer to unit-test `detect_platform` without executing `main`. A test-convenience change broke the production entry point.

And the test suite added in the same PR **structurally cannot catch it**: every case runs `bash "$installer"` (a path — guard passes) or sources it. **None pipes via stdin.** So the tests written to protect the installer are blind to the only way users invoke it.

## Fix

```sh
if [ "${BASH_SOURCE[0]:-$0}" = "$0" ]; then
```

Verified in all three invocation modes:

| mode | result |
| --- | --- |
| piped (`… \| bash`) | reaches `main` ✅ |
| executed (`bash install.sh`) | reaches `main` ✅ |
| sourced (`source install.sh`) | correctly skipped ✅ |

Must be applied to **both** `scripts/install.sh` and `packages/docs-web/public/install` — they are byte-identical mirrors and a parity check now enforces that.

## Two things that must land in the same PR

**A piped-stdin regression test.** Without it this recurs, and it is the one case the new suite omits. A test that pipes the installer and asserts a non-error exit fails on current `dev` and passes with the fix.

**Do not "fix" this by deleting the guard.** `detect_with_mocks` *sources* the installer, so it depends on the guard suppressing `main`. Remove it and that test performs a **real GitHub download and prompts for sudo** — observed on a reviewer's machine against the pre-merge installer. #2335 is about to run this suite in CI, which would make that a CI-side network+sudo call.

## Sequencing

**#2335 must not merge before this.** It adds `test:install` to CI; merging it alone green-lights a suite that passes while the installer is broken, which is worse than no CI coverage — it converts an absent check into a false assurance.

## Separately: post-install verification was removed, not moved

The pre-#2330 installer ran `$INSTALL_DIR/$BINARY_NAME version` **after** installation. Line 93 now prints the cached probe output from the temp directory and declares "Installation complete!" **without ever executing the installed file**. Worth restoring in the same PR — a post-install check that never runs the installed artifact cannot detect a bad install, which is the class of failure #2295 reported in the first place.

## What is NOT wrong with #2330 (checked, so nobody re-litigates it)

- **Rosetta detection is correct** — all four paths, including `sysctl` errors falling back to x64, and the block is correctly placed before arch normalization.
- **The `darwin-arm64` asset genuinely exists** — `release.yml:71-73` builds it, `:301` uploads `dist/archon-*`, `:280` writes it into `checksums.txt`. Read directly, not inferred. No 404 regression.
- **The mirror is still byte-identical** on `origin/dev`.
- **checksum → probe → `mv` ordering is correct**; a failed probe cannot leave a half-install.

## Related

- #2295 / #2330 — the issue and the PR that introduced this.
- #2332 / #2335 — installer mirror parity and its CI wiring; sequencing note above.

### LINKED ISSUE #2339: fix(install): public install.ps1 has already diverged — Windows users still get PATH-corruption bug #1000
## Summary

`scripts/install.ps1` and `packages/docs-web/public/install.ps1` are a **second** hand-synced installer mirror pair — and unlike the bash pair, they have **already diverged**. The divergence means the public Windows installer still ships bug **#1000**, which was fixed in the repo copy over a year of commits ago and never mirrored.

`irm https://archon.diy/install.ps1 | iex` is the documented Windows install path (`README.md:166`, `docs.mdx:37`, `installation.md:21`), so this affects real users today.

## The divergence

Verified against `origin/dev`:

```
$ diff <(git show origin/dev:scripts/install.ps1) \
       <(git show origin/dev:packages/docs-web/public/install.ps1)
191c191
<     $pathParts   = @($currentPath -split ';' | Where-Object { $_ -ne '' })
---
>     $pathParts   =   $currentPath -split ';' | Where-Object { $_ -ne '' }
```

One line. The canonical copy wraps the pipeline in `@(…)`; the public mirror does not.

## Why that one line matters

The `@()` array-subexpression wrapper was added by commit `53cabd44` — *"fix: PowerShell Add-ToUserPath corrupts PATH when single entry exists (#1000)"*.

Without it, PowerShell unwraps a single-element pipeline result to a **scalar string** instead of a one-element array. The subsequent PATH manipulation then operates on a string rather than a list, and the user's `PATH` is corrupted. That is precisely bug #1000, and it is what `irm | iex` installs on Windows right now.

So #2332's stated failure mode — *"fix an installer bug, forget the mirror, and every user who follows the documented install path gets the old broken installer"* — is not hypothetical. It has already happened, on the `.ps1` pair, and went unnoticed because nothing compares them.

## Why nothing caught it

- The byte-parity assertion added by #2330 (`scripts/test-install.sh:71-72`, `cmp -s`) covers the **bash** pair only.
- `scripts/test-install.sh` is a bash script exercising bash-installer behaviour; there is no PowerShell equivalent.
- So the `.ps1` pair has **zero** drift protection, and the `.sh` pair has protection that arrived only last commit.

## Proposed

**Immediate:** mirror the `@()` fix into `packages/docs-web/public/install.ps1` so the public Windows installer stops corrupting `PATH`. One line, no behaviour change to the canonical copy.

**Then, so it cannot recur:** extend the existing parity assertion to cover **every** mirrored installer pair, not just the bash one. The `cmp -s` in `test-install.sh` is the right mechanism and already runs in `bun run validate`; it needs a second invocation for the `.ps1` pair. Two lines.

Worth doing as one PR — the fix without the guard leaves the next divergence just as invisible.

**Better still, and the reason this keeps happening:** consider whether the public copies should be *generated* from the canonical ones at docs-build time rather than tracked twice. Two tracked files that must be byte-identical is a standing invitation to this bug; a build step makes it structurally impossible. That is a larger change and deserves its own decision — noting it rather than proposing it here.

## Related

- #2332 — the bash mirror pair and its drift risk. This is the same defect, already realized, on a pair that issue did not cover.
- #2335 — wires the bash parity check into CI. Would not have caught this.
- #1000 / `53cabd44` — the original PATH-corruption fix that never reached the public copy.
- #2338 — a separate, more severe installer regression currently on `dev`. Independent of this.

Wirasm:
## Summary

- **Problem:** #2330 added `if [[ "${BASH_SOURCE[0]}" == "$0" ]]` so `test-install.sh` could `source` the installer. Under `set -u` with bash reading from **stdin** — exactly what `curl -fsSL … | bash` does — `BASH_SOURCE[0]` is unbound, so the script aborted before `main()` ran.
- **Why it matters:** the documented install path failed for **every user on every platform**, turning #2295 from "installs but won't run on one Mac config" into "installer hard-fails for everyone". Not yet live — `deploy-docs.yml` publishes only from `main` — but it ships the moment `dev` merges, and `packages/docs-web/**` changed so that deploy will fire. **Release blocker.**
- **What changed:** `${BASH_SOURCE[0]:-$0}`; post-install verification now runs against the installed path; the public `install.ps1` mirror is re-synced; two regression tests.
- **What did NOT change (scope boundary):** the Rosetta detection logic from #2330 is untouched and correct. No change to release assets, Homebrew, or the canonical `scripts/install.ps1`.

### Before

```
$ curl -fsSL https://archon.diy/install | bash
bash: line 262: BASH_SOURCE[0]: unbound variable
$ echo $?
1
# nothing installed, no diagnostic the user can act on
```

### After

```
$ curl -fsSL https://archon.diy/install | bash
[INFO] Detecting platform...
[OK] Platform: darwin-arm64
...
[OK] Installed to /usr/local/bin/archon
archon 1.2.3
[OK] Installation complete!
```

### Before

```
curl | bash ──▶ set -u ──▶ guard reads BASH_SOURCE[0] ──▶ ✗ abort (main never runs)
bash install.sh ─▶ guard ─▶ main ✓
source install.sh ─▶ guard ─▶ (skipped) ✓

download ─▶ checksum ─▶ probe TEMP binary ─▶ mv ─▶ print CACHED probe output ─▶ "complete!"
                                                    └─ installed file never executed
```

### After

```
curl | bash ──▶ guard reads ${BASH_SOURCE[0]:-$0} ──▶ main ✓
bash install.sh ─▶ main ✓          source install.sh ─▶ (skipped) ✓

download ─▶ checksum ─▶ probe TEMP binary ─▶ mv ─▶ run INSTALLED binary ─▶ "complete!"
                        (guards an existing install)   (guards the actual artifact)
```

**Connection inventory** (module-to-module edges):

- `scripts/install.sh` ↔ `packages/docs-web/public/install` — byte-identical mirrors (**both changed, parity asserted**)
- `scripts/install.ps1` ↔ `packages/docs-web/public/install.ps1` — byte-identical mirrors (**public copy re-synced, parity now asserted**)
- `scripts/test-install.sh` → both installers (**changed**: two new assertions)
- `bun run validate` → `test:install` (unchanged edge)
- `deploy-docs.yml` → `packages/docs-web/public/*` (unchanged; the publish path for the fixed mirrors)

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `ci`
- Module: `infra:installer`

## Change Metadata

- Change type: `bug`
- Primary scope: `multi`

## Linked Issue

- Closes #2338
- Closes #2339
- Related #2295, #2330, #2332, #2335, #1000

```bash
bash scripts/test-install.sh     # → Installer tests passed
bun run check:bundled            # → up to date (36 commands, 21 workflows)
bun run format:check             # → All matched files use Prettier code style!
```

- **Evidence provided:** the new piped test was verified to **fail** against the broken installer, not merely to pass against the fixed one. With `scripts/install.sh` **and** the public mirror both reverted to `08723171` — so the parity assertion does not mask it — the suite reports:
  `FAIL: installer aborts when piped to bash (BASH_SOURCE unbound under set -u)`
  Restoring the fix returns `Installer tests passed`.
- The guard was independently checked in all three invocation modes (piped → reaches `main`; by path → reaches `main`; sourced → correctly skipped), on macOS system bash **3.2.57**.
- **Intentionally skipped:** full `bun run type-check` / `lint` / `test`. This PR touches only shell scripts and one PowerShell file — no TypeScript. CI runs the full chain.

## Security Impact (required)

- New external network calls? **No** — the new test uses the existing mocked `curl` and a scratch `INSTALL_DIR`; it makes no real request.
- File system access scope changed? **No** — post-install verification executes a path the installer already wrote.
- Risk note: the fix deliberately **keeps** the source-guard rather than deleting it. Removing it would re-arm `detect_with_mocks`, which sources the installer and would then perform a real GitHub download and prompt for **sudo** — and #2335 is about to run this suite in CI.

## Compatibility / Migration

- Config/env changes? **No**

- **Verified scenarios:** `… | bash` (the reported failure) fails before and succeeds after; `bash install.sh`; `source install.sh`; installer with a failing binary leaves an existing installation intact; installer with a working binary replaces it and prints the version.
- **Edge cases checked:** bash 3.2.57 (macOS system bash) as well as modern bash; the sourced path still suppresses `main`; `install.ps1` line 191 confirmed to be the **only** divergence before re-syncing, so the mirror update changed exactly one line.
- **What was not verified:** the PowerShell fix was not executed — no Windows host available. It is a byte-for-byte adoption of the canonical copy that has been on `dev` since `53cabd44`, not new code. The published `archon.diy/install` endpoint was confirmed to still serve the pre-#2330 script, so no live user is currently affected.

- **Affected subsystems:** shell + PowerShell installers, their published mirrors, `test:install` (which runs inside `bun run validate`).
- **Potential unintended effects:** post-install verification adds one execution of the installed binary. If a platform installs a working file that cannot run in place (e.g. quarantine/SIP), the installer now fails loudly where it previously reported success — intended, and the point of the change.
- **Guardrails:** `cmp` parity on both mirror pairs fails locally and (via #2335) in CI. **Correction:** as originally written this claim was only true for LF checkouts — the extensionless mirror fell through `* text=auto` to CRLF on `core.autocrlf=true` (Windows default), so the assertion failed spuriously on a clean tree there. Fixed in `84ae8c26` by pinning it in `.gitattributes`. The piped test covers the invocation mode the suite previously could not reach, and `make_platform_mocks` now stubs `curl`/`sudo` so a future guard regression fails locally instead of reaching the real release URL during `bun run validate`.

## Rollback Plan (required)

- Fast rollback: `git revert 10fdc201`. Self-contained across four files, no schema or config coupling.
- Caveat: reverting restores the broken `curl | bash` path, so prefer a forward fix.

## Sequencing note for reviewers

**#2335 should not merge before this.** It adds `test:install` to CI; landing it first green-lights a suite that passes while the installer is broken — converting an absent check into a false assurance. After this merges, #2335's CI wiring becomes genuinely protective (it also needs `if: runner.os != 'Windows'` on that step — the installer refuses MINGW, so the Windows job goes red without it).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installer reliability for piped execution (`curl … | bash`), including safer entrypoint handling when sourced via stdin.
  * Added a post-install validation step that runs `version` from the installed binary path and fails if it can’t execute.
  * Ensured clear failure when the installed binary probe doesn’t succeed.
  * Added checks to confirm the published Windows PowerShell installer matches the repository copy.

* **Tests**
  * Strengthened piped-installer regression coverage with exit-code and stderr assertions.
  * Improved mock behavior to prevent unintended network/privilege actions.
  * Verified the installed binary after piped installs.

* **Chores**
  * Standardized published installer line endings for accurate byte-for-byte comparison.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,6 +3,11 @@
 
 # Force LF line endings for files that run inside Linux containers
 *.sh text eol=lf
+# The published installer mirrors scripts/install.sh byte-for-byte and a `cmp`
+# assertion enforces that. It has no extension, so `*.sh` misses it and it would
+# fall through to `* text=auto` -> CRLF on a Windows-default checkout, failing the
+# parity check on a clean tree.
+packages/docs-web/public/install text eol=lf
 docker-entrypoint.sh text eol=lf
 Caddyfile text eol=lf
 Caddyfile.example text eol=lf
--- a/packages/docs-web/public/install
+++ b/packages/docs-web/public/install
@@ -239,7 +239,18 @@ main() {
 
   success "Installed to $INSTALL_DIR/$BINARY_NAME"
 
-  echo "$version_output"
+  # Re-run the version check against the INSTALLED path. The probe above ran on the
+  # temp download and its output was cached; printing that after `mv` would report
+  # success without ever executing the file the user will actually invoke — which is
+  # exactly the "installed fine but won't run" failure #2295 reported. See #2338.
+  local installed_output
+  if ! installed_output=$("$INSTALL_DIR/$BINARY_NAME" version 2>&1); then
+    error "Installed binary failed its version check at $INSTALL_DIR/$BINARY_NAME:"
+    echo "$installed_output" >&2
+    exit 1
+  fi
+
+  echo "$installed_output"
   success "Installation complete!"
 
   # Check if in PATH
@@ -259,6 +270,10 @@ main() {
   echo ""
 }
 
-if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+# `${BASH_SOURCE[0]:-$0}` — NOT bare `${BASH_SOURCE[0]}`. Under `curl … | bash` the
+# script arrives on stdin, where BASH_SOURCE[0] is unbound; with `set -u` (above)
+# a bare reference aborts before main() ever runs, so the documented install path
+# fails for every user on every platform. See #2338.
+if [ "${BASH_SOURCE[0]:-$0}" = "$0" ]; then
   main "$@"
 fi
--- a/packages/docs-web/public/install.ps1
+++ b/packages/docs-web/public/install.ps1
@@ -188,7 +188,7 @@ function Confirm-Checksum {
 function Add-ToUserPath {
     param([string]$Dir)
     $currentPath = [Environment]::GetEnvironmentVariable('Path', 'User')
-    $pathParts   = $currentPath -split ';' | Where-Object { $_ -ne '' }
+    $pathParts   = @($currentPath -split ';' | Where-Object { $_ -ne '' })
 
     if ($Dir -notin $pathParts) {
         $pathParts += $Dir
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -239,7 +239,18 @@ main() {
 
   success "Installed to $INSTALL_DIR/$BINARY_NAME"
 
-  echo "$version_output"
+  # Re-run the version check against the INSTALLED path. The probe above ran on the
+  # temp download and its output was cached; printing that after `mv` would report
+  # success without ever executing the file the user will actually invoke — which is
+  # exactly the "installed fine but won't run" failure #2295 reported. See #2338.
+  local installed_output
+  if ! installed_output=$("$INSTALL_DIR/$BINARY_NAME" version 2>&1); then
+    error "Installed binary failed its version check at $INSTALL_DIR/$BINARY_NAME:"
+    echo "$installed_output" >&2
+    exit 1
+  fi
+
+  echo "$installed_output"
   success "Installation complete!"
 
   # Check if in PATH
@@ -259,6 +270,10 @@ main() {
   echo ""
 }
 
-if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+# `${BASH_SOURCE[0]:-$0}` — NOT bare `${BASH_SOURCE[0]}`. Under `curl … | bash` the
+# script arrives on stdin, where BASH_SOURCE[0] is unbound; with `set -u` (above)
+# a bare reference aborts before main() ever runs, so the documented install path
+# fails for every user on every platform. See #2338.
+if [ "${BASH_SOURCE[0]:-$0}" = "$0" ]; then
   main "$@"
 fi
--- a/scripts/test-install.sh
+++ b/scripts/test-install.sh
@@ -42,6 +42,20 @@ if [ "$translated" = "fail" ]; then
 fi
 echo "$translated"
 EOF
+  # detect_with_mocks SOURCES the installer, so it depends on the source-guard
+  # suppressing main(). If that guard ever regresses, main() runs and reaches the real
+  # release URL over curl -- and sudo mkdir/mv with a non-writable INSTALL_DIR -- during
+  # `bun run validate`. These stubs turn that latent network/privilege side effect into a
+  # deterministic local failure.
+  for forbidden in curl sudo; do
+    cat >"$mock_dir/$forbidden" <<EOF
+#!/usr/bin/env bash
+echo "TEST BUG: sourced installer invoked $forbidden — the source guard is not suppressing main()" >&2
+exit 99
+EOF
+    chmod +x "$mock_dir/$forbidden"
+  done
+
   chmod +x "$mock_dir/uname" "$mock_dir/sysctl"
   echo "$mock_dir"
 }
--- a/scripts/test-install.sh
+++ b/scripts/test-install.sh
@@ -143,8 +143,10 @@ esac
 # sources it, so none of them can catch this.
 #
 # Mocked curl + a scratch INSTALL_DIR keep this off the network and out of the real system.
+piped_status=0
 piped_stderr="$(PATH="$success_mock_dir:$PATH" INSTALL_DIR="$tmp_dir/piped-bin" \
-  SKIP_CHECKSUM=true bash <"$installer" 2>&1 >/dev/null || true)"
+  SKIP_CHECKSUM=true bash <"$installer" 2>&1 >/dev/null)" || piped_status=$?
+assert_equals "0" "$piped_status" "Piped installer exits successfully"
 case "$piped_stderr" in
   *"unbound variable"*)
     fail "installer aborts when piped to bash (BASH_SOURCE unbound under set -u)"
--- a/scripts/test-install.sh
+++ b/scripts/test-install.sh
@@ -160,12 +160,15 @@ esac
 piped_status=0
 piped_stderr="$(PATH="$success_mock_dir:$PATH" INSTALL_DIR="$tmp_dir/piped-bin" \
   SKIP_CHECKSUM=true bash <"$installer" 2>&1 >/dev/null)" || piped_status=$?
-assert_equals "0" "$piped_status" "Piped installer exits successfully"
+# Specific diagnosis FIRST, generic assertion second. With the order reversed the
+# assert_equals fired on any regression and this case block was dead code, so the
+# #2338 failure reported only "expected 0, got 1" with no hint at the cause.
 case "$piped_stderr" in
   *"unbound variable"*)
-    fail "installer aborts when piped to bash (BASH_SOURCE unbound under set -u)"
+    fail "installer aborts when piped to bash (BASH_SOURCE unbound under set -u) — see #2338"
     ;;
 esac
+assert_equals "0" "$piped_status" "Piped installer exits successfully"
 assert_equals "archon 1.2.3" "$("$tmp_dir/piped-bin/archon" version)" "Piped installer installs a working binary"
 
 echo "Installer tests passed"
--- a/scripts/test-install.sh
+++ b/scripts/test-install.sh
@@ -71,6 +71,12 @@ assert_platform Darwin x86_64 fail darwin-x64 "Native Intel platform without Ros
 cmp -s "$installer" "$repo_root/packages/docs-web/public/install" \
   || fail "public installer mirror differs from scripts/install.sh"
 
+# The PowerShell pair needs the same guard. It had ALREADY drifted: the public copy
+# was missing the @() array wrapper from 53cabd44 (#1000), so `irm … | iex` shipped a
+# PATH-corrupting installer to Windows users while the repo copy was fixed. #2339.
+cmp -s "$repo_root/scripts/install.ps1" "$repo_root/packages/docs-web/public/install.ps1" \
+  || fail "public install.ps1 mirror differs from scripts/install.ps1"
+
 mock_dir="$tmp_dir/install-mocks"
 install_dir="$tmp_dir/install-bin"
 mkdir -p "$mock_dir" "$install_dir"
@@ -129,5 +135,21 @@ case "$install_output" in
   *"archon 1.2.3"*) ;;
   *) fail "Installer prints verified version" ;;
 esac
+
+# Regression for #2338: the installer must survive being READ FROM STDIN, which is how
+# `curl -fsSL … | bash` delivers it. There BASH_SOURCE[0] is unbound, and with `set -u` a
+# bare reference aborted the script before main() ran — the documented install path failed
+# for every user on every platform. Every other case here invokes the installer by path or
+# sources it, so none of them can catch this.
+#
+# Mocked curl + a scratch INSTALL_DIR keep this off the network and out of the real system.
+piped_stderr="$(PATH="$success_mock_dir:$PATH" INSTALL_DIR="$tmp_dir/piped-bin" \
+  SKIP_CHECKSUM=true bash <"$installer" 2>&1 >/dev/null || true)"
+case "$piped_stderr" in
+  *"unbound variable"*)
+    fail "installer aborts when piped to bash (BASH_SOURCE unbound under set -u)"
+    ;;
+esac
+assert_equals "archon 1.2.3" "$("$tmp_dir/piped-bin/archon" version)" "Piped installer installs a working binary"
 
 echo "Installer tests passed"


Wirasm:
---
pr: 2340
title: "fix(install): repair the curl|bash entry point and sync the PowerShell mirror"
author: "Wirasm"
reviewed: 2026-07-30
recommendation: approve
---

# PR Review: #2340 — fix(install): repair the curl|bash entry point and sync the PowerShell mirror

**Branch**: `fix/2338-installer-stdin` → `dev`
**Files changed**: 4 (+59/-5) · commits `10fdc201`, `fac6e4c6`
**Base for comparison**: `08723171` (#2330, the regressing commit)

---

## Summary

The fix is correct and the central claim holds. I reproduced the before/after test result
independently, exercised the guard in 13 invocation modes plus four non-bash shells, and
proved the PowerShell sync is byte-exact to the already-merged #1000 fix without executing
PowerShell. **No critical or blocking issues.**

Two Important findings, neither introduced by this PR and neither user-facing: a
`.gitattributes` asymmetry that makes the **pre-existing** `.sh` parity assertion fail
spuriously on Windows-default checkouts (which falsifies one portability claim in the PR
body), and a test that would perform a real GitHub download plus a `sudo` prompt if the
guard ever regresses again.

---

## Verification method

Everything below is split into **executed** and **reasoned**. Work was done in a throwaway
worktree at `/private/tmp/.../scratchpad/wt-2340`; the primary checkout was never touched.
Every installer invocation used a mocked `curl`, a scratch `INSTALL_DIR`, and — for the
sourced-path tests — tripwire `curl`/`sudo` stubs first on `PATH` that print a loud marker
and exit 99, so any real network or privilege attempt is visible rather than inferred.
The tripwire was negative-controlled against the pre-#2330 installer (no guard at all): it
fired, confirming it is a working detector and not a silent pass.

---

## Item-by-item response to the review brief

### 1. Is `${BASH_SOURCE[0]:-$0}` correct in every invocation mode? — **VERIFIED (executed)**

Matrix run on macOS system bash **3.2.57** (the only bash on this host; no Homebrew bash
installed), old guard vs. new guard:

| Mode | `[[ "${BASH_SOURCE[0]}" == "$0" ]]` | `[ "${BASH_SOURCE[0]:-$0}" = "$0" ]` |
|------|---|---|
| `bash ./f` (± args) | main ✓ | main ✓ |
| `./f` (shebang) | main ✓ | main ✓ |
| `bash < f` | **abort 1** | main ✓ |
| `cat f \| bash` (true pipe) | **abort 1** | main ✓ |
| `cat f \| bash -s XX YY` | **abort 1** | main ✓ (argv preserved) |
| `bash -c "$(cat f)"` | **abort 1** | main ✓ |
| `bash --posix` by path | main ✓ | main ✓ |
| `bash --posix` piped | **abort 1** | main ✓ |
| `source f` from a script | suppressed ✓ | suppressed ✓ |
| `bash -c 'source $1' _ f` (the suite's shape) | suppressed ✓ | suppressed ✓ |

**No mode stops running `main` that previously ran it.** Modes that newly run `main` are
exactly the stdin/`bash -c` family — the fix — plus non-bash shells (below).

Modes the brief asked about specifically:

- **`sh install.sh` on a dash system**: irrelevant to the guard. Verified with Homebrew
  dash — both old and new abort at **line 23** (`set -o pipefail` → `Illegal option`),
  never reaching the guard. Unchanged behavior. On macOS `/bin/sh` is bash in POSIX mode,
  so it works in both.
- **bash in POSIX mode**: covered above; new guard works piped, old aborted.
- **Sourced from a script rather than interactively**: covered; suppressed in both.
- **`BASH_SOURCE` set-but-empty**: I tried to construct it (`BASH_SOURCE=("")` and
  `BASH_SOURCE[0]=""` before `source`). Bash rejects both assignments (exit 1), and
  `unset BASH_SOURCE` fails with `cannot unset`. Not reachable in bash.

The structural reason the sourced path is safe: `:-` fires only when `BASH_SOURCE[0]` is
unset or empty, and while bash is sourcing a file that element is always the sourced
path. So sourced-suppression semantics are **bit-for-bit unchanged** from #2330.

### 2. Does the sourced path still suppress `main`? — **VERIFIED (executed), yes**

`bash -c 'source "$1"; detect_platform' _ scripts/install.sh` with tripwire `curl`/`sudo`
first on `PATH` returned exactly `darwin-arm64`, **exit 0, no tripwire output** — `main()`
did not run. The full suite also runs clean with tripwires installed. Negative control
(pre-#2330 installer, same invocation) fired the curl tripwire with the real release URL,
so the detector works.

Additional finding, in the PR's favour: the existing `assert_platform` calls act as an
implicit canary. I neutralised the guard to `if true; then` and ran the HEAD suite — it
failed with `FAIL: Rosetta platform: expected darwin-arm64, got …banner…`. So a future
guard leak cannot pass this suite silently. (It does have a side effect on trip — see
Important I2.)

### 3. Can post-install verification fail wrongly? — **VERIFIED (executed), no**

| Scenario | Result |
|---|---|
| First install into a fresh dir, `INSTALL_DIR` **not** on `PATH` | exit 0 ✓ |
| `mv` fails → mocked `sudo mv` completes it | exit 0 ✓ |
| Installed file loses `+x` (noexec/bad-perms proxy) | exit 1, `Permission denied` surfaced ✓ |
| Installed binary runs but exits 133 with a `dyld` message | exit 1, `dyld[123]: Library not loaded…` surfaced ✓ |
| Relative `INSTALL_DIR` (`relbin`) | exit 0 ✓ |
| `INSTALL_DIR` with trailing slash | exit 0 ✓ |

`PATH` cannot break it: the check invokes `"$INSTALL_DIR/$BINARY_NAME"`, which always
contains a `/`, so bash treats it as a path and never does a `PATH` lookup. Diagnostics
are good — the underlying error is echoed to stderr in both failure cases.

**macOS quarantine (reasoned, not executed):** `curl -o` does not set
`com.apple.quarantine` — that is applied by LaunchServices-aware downloaders, not curl —
so Gatekeeper is not a new failure vector here. More decisively, the check's failure
surface is a **subset** of the pre-install probe #2330 already gates on, plus
path/permission differences: anything that makes `archon version` fail for environmental
reasons already aborts the install at `install.sh:216`, before `mv`.

**Side-effect note (reasoned):** `archon version` reaches `printUpdateNotice`
(`packages/cli/src/cli.ts:1011`), which for binary builds calls `checkForUpdate` — a
network call, wrapped in try/catch so it cannot fail the command, and cached to
`~/.archon/update-check.json` with a 24 h TTL. The first (temp) probe populates that
cache, so the second invocation hits it. No meaningful added latency, no new failure mode.

### 4. Does the new test actually prove the fix? — **YES, independently reproduced**

I rebuilt three trees from git blobs and ran the **HEAD** test script against each:

| Tree | Result |
|---|---|
| HEAD installer + HEAD mirror | `Installer tests passed` (exit 0) |
| `08723171` installer + `08723171` mirror (both reverted) | **exit 1** — `FAIL: Piped installer exits successfully: expected 0, got 1` |
| `08723171` installer, HEAD mirror (only one reverted) | exit 1 — `FAIL: public installer mirror differs from scripts/install.sh` |

So the test is not vacuous, and the masking concern is real: reverting only
`scripts/install.sh` trips the mirror `cmp` at line 71 before the piped test is reached.
Reverting both is required, exactly as the PR body states.

**One correction to the PR body.** It quotes the failure as
`FAIL: installer aborts when piped to bash (BASH_SOURCE unbound under set -u)`. That was
accurate at `10fdc201` (verified — I ran that revision's test and got precisely that
string), but `fac6e4c6` inserted `assert_equals "0" "$piped_status"` **before** the stderr
`case`, so the actual message at PR HEAD is the generic
`Piped installer exits successfully: expected 0, got 1`. See Suggestion S1.

### 5. The `.ps1` change — **VERIFIED (executed) by blob identity, no PowerShell needed**

`diff` of the two blobs at `08723171` confirms **line 191 was the only divergence**, as
claimed. Better than that, the blob chain removes the need to reason about PowerShell
semantics at all:

| Revision | `scripts/install.ps1` | `packages/docs-web/public/install.ps1` | |
|---|---|---|---|
| `53cabd44^` | `3ba6bb31` | `3ba6bb31` | same |
| `53cabd44` (#1000 fix) | `20834f6e` | `3ba6bb31` | **drift starts** |
| `08723171` | `20834f6e` | `3ba6bb31` | drifted |
| `10fdc201` (this PR) | `20834f6e` | `20834f6e` | **same** |

`53cabd44` touched **one file, one line**. The new public copy is byte-identical to the
canonical blob that has been shipping in `scripts/install.ps1` since 2026-04-09. This is
a restoration of already-reviewed, already-merged content, not new code — the "unverified
by execution" caveat in the PR body is stronger than it needs to be.

### 6. Blast radius of the parity assertion — **VERIFIED (executed); the new one is safe, its older sibling is not**

Effective attributes (`git check-attr -a`):

| File | Attributes |
|---|---|
| `scripts/install.sh` | `text: set`, **`eol: lf`** |
| `packages/docs-web/public/install` | `text: auto` (no `eol`) |
| `scripts/install.ps1` | `text: auto` |
| `packages/docs-web/public/install.ps1` | `text: auto` |

The **new `.ps1` assertion cannot fail from line endings** — both files carry identical
attributes, so any checkout normalises them identically. Confirmed on a simulated
`core.autocrlf=true` clone: both got CRLF, `cmp` passed.

The **pre-existing `.sh` assertion does** fail there. Same clone: `scripts/install.sh`
came out LF (forced by `*.sh text eol=lf`) while `packages/docs-web/public/install` came
out CRLF (it matches only `* text=auto`), and the suite exited 1 with
`FAIL: public installer mirror differs from scripts/install.sh`. See Important I1.

Intent question: yes, `cmp` failing any future single-copy edit is the point, and it is the
right guardrail — #2339 exists precisely because #1000 edited one copy for 3.5 months
without anything noticing.

---

## Issues Found

### Critical

None.

### Important

Neither is introduced by this PR, neither affects users, and **neither should block this
P0.** Both are cheap and both live in files this PR already edits.

**I1 — `.gitattributes:2-5` — the `.sh` mirror parity assertion fails spuriously on any `core.autocrlf=true` checkout (Windows default).**

- **Why**: `scripts/install.sh` is pinned to LF by `*.sh text eol=lf`; the extensionless
  mirror falls through to `* text=auto` and is converted to CRLF on checkout. `cmp -s`
  compares bytes, so the two differ. Reproduced on a simulated Windows clone:
  `scripts/test-install.sh` exits 1 at line 71.
- **Failure scenario**: a Windows contributor runs `bun run validate` in Git Bash and gets
  `FAIL: public installer mirror differs from scripts/install.sh` on a clean tree, with
  nothing actually wrong. Once #2335 lands, the same fires in CI on `windows-latest` —
  and the `if: runner.os != 'Windows'` guard noted in the PR's sequencing section hides it
  in CI while leaving local `validate` broken for those developers.
- **This also falsifies a claim in the PR body**: "`cmp` parity on both mirror pairs now
  fails locally and (via #2335) in CI" is only true for LF checkouts.
- **Fix**: one line — `packages/docs-web/public/install text eol=lf` in `.gitattributes`.
  Add it here (zero risk, makes the guardrail actually portable) or file it alongside
  #2335. Do not hold the P0 for it.

**I2 — `scripts/test-install.sh:49-52` — `detect_with_mocks` runs without mocked `curl`/`sudo`, so a guard regression reaches the real network.**

- **Why**: `make_platform_mocks` stubs only `uname` and `sysctl`. If the source guard ever
  stops suppressing `main` — the exact failure mode this PR is repairing the other half
  of — the sourced installer proceeds into `main()` and calls the **real** `curl` against
  `https://github.com/coleam00/Archon/releases/latest/download/…`, and can reach
  `sudo mkdir`/`sudo mv`.
- **Evidence (executed)**: I neutralised the guard to `if true; then` and ran the HEAD
  suite. It failed correctly (`FAIL: Rosetta platform`), but only *after* the curl tripwire
  fired with the real release URL. Without tripwires that is a live network request, and
  with a non-writable `INSTALL_DIR` a `sudo` password prompt, during `bun run validate`.
  This is the "check that looks thorough while letting the dangerous thing happen" pattern
  the brief asked me to hunt for.
- **Fix**: have `make_platform_mocks` also drop hard-failing `curl` and `sudo` stubs into
  `$mock_dir` (or prepend `$success_mock_dir`). Five lines, converts a latent real-world
  side effect into a deterministic local failure.

### Suggestions

**S1 — `scripts/test-install.sh:146-154` — the specific diagnostic is now unreachable, and the PR body quotes it anyway.**
`assert_equals "0" "$piped_status"` runs before the `case "$piped_stderr" in *"unbound
variable"*)` block. A script that aborts on an unbound variable exits 1, so the status
assert always fires first and the `case` is effectively dead code. Swap the two so the
precise message wins, and update the Validation Evidence block in the PR description to
match `fac6e4c6`. (`fac6e4c6` itself is a genuine improvement — `10fdc201` had `|| true`,
which discarded the exit status and left the test resting on a stderr substring alone.)

**S2 — `scripts/install.ps1:284-293` — the PowerShell installer has the same defect this PR fixes in bash.**
It does verify the installed path (`& $destBinary version`), but a native executable that
runs and exits non-zero is not a PowerShell terminating error, so the `catch` never fires
and `$LASTEXITCODE` is never checked. A broken binary still prints its error output
followed by `Write-Ok "Installation complete!"` — the #2295 failure shape, on Windows.
Out of this PR's declared scope; worth a follow-up issue adding an `if ($LASTEXITCODE -ne
0) { … exit 1 }`.

**S3 — `scripts/install.sh:277` — the guard fails open for non-bash sourcers.**
Verified: `source install.sh` from **zsh** now runs `main` (curl tripwire fired) because
`BASH_SOURCE` does not exist there, so `:-$0` yields `$0`, which zsh sets to the sourced
file's name. Same for ksh. Not a regression — pre-#2330 had no guard at all, and the only
in-repo sourcer uses `bash -c` — but the guard's contract is narrower than it reads.
Worth one clause in the existing comment: it protects **bash** sourcers only.

**S4 — `scripts/install.sh:246-251` — no remediation hint on the new hard failure.**
When post-install verification fails the installer exits 1 having already replaced the
user's previous binary, and skips the PATH hint and "Get started" block. The loud failure
is the intended trade-off, but one line ("the previous binary was replaced; re-run the
installer or download manually from …") would make an already-bad moment recoverable.

**S5 — `scripts/install.sh:16-21` (pre-existing, drive-by) — the documented examples don't work.**
`VERSION=v0.2.0 curl -fsSL … | bash` and `INSTALL_DIR=~/.local/bin curl -fsSL … | bash`
set the variable in **curl's** environment, not in the consuming `bash`'s, so both are
silently ignored. Correct form is `curl -fsSL … | VERSION=v0.2.0 bash`. Unrelated to this
fix, but cheap while the file is open — and it is copy-pasted into the docs mirror.

---

## Validation Results

| Check | Status | Details |
|---|---|---|
| `bash scripts/test-install.sh` (PR HEAD worktree) | PASS | `Installer tests passed` |
| Same, with tripwire `curl`/`sudo` on `PATH` | PASS | no tripwire fired — no network, no sudo |
| Guard matrix, 13 invocation modes × 2 revisions | PASS | see item 1 |
| Non-bash shells (`sh`, `zsh`, `ksh`, `dash`) | PASS | see items 1 / S3 |
| Before/after reproduction, 3 trees | PASS | see item 4 |
| `.ps1` blob-identity chain | PASS | `20834f6e` on both sides |
| Simulated `core.autocrlf=true` checkout | **FAIL** | pre-existing `.sh` `cmp` — see I1 |
| CI: `test (ubuntu-latest)` | PASS | 3m6s |
| CI: `Build docs site` | PASS | 46s |
| CI: `test (windows-latest)`, `docker-build` | pending | `mergeStateStatus: UNSTABLE` is pending, not failing |
| CodeRabbit | PASS | "No actionable comments" |
| Type-check / lint / unit tests | not run | correctly skipped — no TypeScript in the diff; CI covers it |
| `format:check` | not run | Prettier has no parser for `.sh`/`.ps1`/extensionless and `.prettierignore` does not cover these paths, so they are skipped either way |

Note: `test:install` is **not** in `.github/workflows/test.yml` today — that workflow runs
the individual steps, not `bun run validate`. So the parity assertions are local-only until
#2335 lands, which is what makes the sequencing note in the PR body correct.

---

## Strengths

- Root cause is correctly identified and the fix is minimal. The `:-$0` default provably
  cannot change sourced behavior, because the element it defaults is never unset while
  bash sources a file — a much stronger property than "I tested three modes".
- `fac6e4c6` removed a real weakness rather than adding polish: the `|| true` at
  `10fdc201` discarded the exit status, so the test would have passed on any failure that
  did not print "unbound variable".
- Restoring verification against the installed path is the right fix for the *class* of
  bug in #2295, not just its symptom.
- The `.ps1` sync is verifiable without a Windows host, by blob identity to `53cabd44`.
  Worth stating that way in the PR body — it is stronger evidence than the current caveat.
- Both inline comments explain *why* (`:-$0`, and the cached-probe trap) with issue
  references, which is what would have prevented #2330 from reintroducing this.
- Existing `assert_platform` calls already function as a guard-leak canary.

---

## Recommendation

**APPROVE** — merge to unblock the release.

The P0 is genuinely fixed, the test genuinely proves it, and I found nothing that makes a
good install fail. I1 and I2 are pre-existing, non-user-facing, and cheap; my preference is
to add the one-line `.gitattributes` entry (I1) to this PR since it is zero-risk and makes
the parity claim in the description true, and to take I2 and S1 either here or immediately
after. S2 deserves its own issue.

The sequencing note is correct: **#2335 must not merge before this.**

---

*Reviewed by Claude · Report: `/Users/rasmus/.prp/archon-75601ef6/reviews/pr-2340-review.md`*

Issue #2345, opened 2026-07-30
fix(workflows): loop nodes lose resolved-model metadata when resumed through an interactive gate


Wirasm:
## Run contract (added 23 September 2026). The report below is evidence; line numbers are historical and must be re-checked.

- **Problem:** a `loop:` node that completes, pauses at an interactive gate and is resumed with a bare approval finishes through the resume-side finalizer. That path drops the resolved-model metadata (and possibly sibling provider stats) the loop knew before the pause.
- **Why it matters:** the run record goes silent on which model ran, and possibly on stop reason and turns or cost, for exactly the runs a human paused to inspect.
- **Why now:** per-node timings and model data are what the current dogfood comparisons across Claude and DeepSeek read.
- **Outcome:** provider stats known before the pause survive the gate and appear on the terminal `node_completed` event, as if the loop had finished without pausing. Check which siblings are also dropped, and fix them together if they are.
- **Invariants:** absence stays absence (#2314). Never invent a value that wasn't known. The schema changes only additively. The approve-with-text path is unchanged.
- **Acceptance:** a test pauses a loop at an interactive gate, bare-approves it, and asserts the resolved model on `node_completed`. It fails today. The sibling stats get the same treatment if they are dropped too. `bun run validate` passes.

---

## Summary

A `loop:` node that signals completion, pauses at an interactive gate, and is then resumed with a **bare approval** exits through `finalizeLoopFromSignal` instead of the loop's normal completion path. That path does not carry the provider's resolved-model metadata, so the terminal `node_completed` event loses `model_usage` — the record is silently absent for exactly the runs a human paused to look at.

Raised by CodeRabbit on #2337 (`dag-executor.ts:4606`, Major / Heavy lift). Deliberately **not** fixed there: it is a separate code path and a materially larger change than the two blocking findings that PR addressed.

## Mechanism

`finalizeLoopFromSignal` (`dag-executor.ts:2987`, called from `:3119` and `:3859`) is the resume-side finalizer. It reconstructs the node's completion from **persisted gate metadata**, not from a live provider result — because on resume there is no provider result: the iteration that produced the completion signal ran in an earlier process.

`loopResolvedModel` is per-process state captured from result chunks in the executing loop. It is not written into the gate metadata at pause time, so on resume there is nothing to restore and the field is omitted.

[...]

Issue #2074, opened 2026-07-10
Bug: interactive loop node does not terminate on the completion signal (VALIDATED) — re-runs the iteration on approve instead of finalizing


wloonis:
## Prior art checked (before opening this)

I reviewed the existing issues on interactive loops / approval resume / loop termination before opening this. None matches the exact symptom below; the closest are:

- **#2004** (closed, `bug`) — *Loops with `interactive:true` always start a fresh AI session, losing previous-loop conversation continuity (regression #1923)*. This explains part of what I saw (the approved iteration starts a **fresh session** that has no memory the previous iteration already succeeded → it re-does the work). It was reverted/restored by #2005, so on current main the session-threading should be back — yet the re-run-on-approve behavior still reproduces here.
- **#1473** (closed) — *archon-prd-to-wo: loop never emits termination signal, runs to max_iterations*. Same **class** (loop does not terminate), but there the agent **never emitted** the signal (a prompt issue, fixed). Here the agent **did** emit the completion signal and the loop still did not terminate — different root cause.
- **#1259** (open) — *fix(workflows): enable interactive workflow resume after approval gate* (notably its "Bug 2: the API approve endpoint incorrectly set `status: failed` for `interactive_loop` approval types"). Related approve/resume mechanics.
- **#1991** (open) — *Workflow with `interactive: true` hangs indefinitely when executed via background CLI tasks*. Same background/relay scenario, but that one is about hanging waiting for input with no channel; mine is about the loop **re-running on approve instead of completing** after the success signal was emitted.
- **#1601** (open) — *execution-evidence contract with terminal-success gate*. Related to reliable terminal-success detection (claimed vs verified).
- **#1690**, **#1741** (closed) — resumeWorkflow on interactive_loop approval / interactive-loop never resumes / re-ask.

This issue is **distinct**: the completion signal **is** emitted, the beacon **is** written PASS, the report **is** PASS, yet the loop still pauses with a (generic, wrong) `approval_requested` and **re-runs** the iteration on approve instead of finalizing the node.

## Bug

An `interactive: true` **loop node** whose prompt instructs the model to emit a completion signal (here `<promise>VALIDATED</promise>`) on success does **not terminate** when the model emits that signal. Instead:

1. The model runs the validation, gets `pass`, writes the relay beacon (`STATE: PASS`) **and** emits `<promise>VALIDATED</promise>` in its final assistant text.
2. The loop emits `loop_iteration_completed { completionDetected: true }`.
[...]

PR #2347, opened 2026-07-30
fix: Persist per-node workflow token usage

### LINKED ISSUE #2333: fix(workflows): per-node token usage is written only to the JSONL log, never to workflow_events

Wirasm:
## Summary

- Problem: successful workflow node completion events omitted token usage even though providers supplied it and the executor already collected it.
- Why it matters: API, CLI, and remote-console consumers could not inspect per-node usage without reading local JSONL logs.
- What changed: direct AI, loop, and loop-group completion events now persist their existing token values; regression tests cover all three paths.
- What did **not** change (scope boundary): provider contracts, run-level aggregation, event schema/storage, APIs, and database migrations are unchanged.

### Before

```
Workflow operator          Provider              Workflow executor       Event consumers
───────────────          ────────              ─────────────────       ───────────────
runs workflow ─────────▶ returns result
                           + token usage ────▶ captures tokens
                                                  writes completion
                                                  event (without tokens) ──▶ cannot view node usage
```

### After

```
Workflow operator          Provider              Workflow executor       Event consumers
───────────────          ────────              ─────────────────       ───────────────
runs workflow ─────────▶ returns result
                           + token usage ────▶ captures tokens
                                                  writes completion
                                                  event [with tokens] ─────▶ [views node usage]
```

### Before

```
Provider result tokens ──▶ DAG executor ──▶ JSONL logger
                                 │
                                 └───────▶ workflow_events (node_completed; tokens omitted) ──▶ API / CLI / console
```

### After

```
Provider result tokens ──▶ [~ DAG executor] ──▶ JSONL logger
                                 │
                                 └════════▶ workflow_events (node_completed; [tokens persisted]) ──▶ API / CLI / console
```

| Provider result chunks | DAG executor | unchanged | Providers already normalize optional token usage. |
| DAG executor | JSONL logger | unchanged | Direct-node logging already includes tokens. |
| DAG executor | workflow event store | **modified** | Successful direct, loop, and loop-group completion payloads now include available tokens. |
| Workflow event store | API / CLI / console consumers | unchanged | Existing generic event data is returned without schema or storage changes. |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `workflows`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2333
- Related #None
- Depends on #None
- Supersedes #None

Commands and result summary:

```bash
bun run type-check
bun run lint
bun run format:check
bun run test
bun run validate
```

- Evidence provided (test/log/trace/screenshot): focused DAG executor tests passed (434 passed, 0 failed); full validation in a clean temporary snapshot containing only the two issue files passed. Recorded validation also reports type check, lint, format check, build, and 6,401 tests passing.
- If any command is intentionally skipped, explain why: none. The live worktree validation was blocked only by unrelated pre-existing `.archon/` edits that make bundled defaults stale; the same full gate passed in the clean snapshot with this fix applied.

## Security Impact (required)

- New permissions/capabilities? (No)
- New external network calls? (No)
- Secrets/tokens handling changed? (No; this persists existing aggregate usage counts, not credentials or model prompts.)
- File system access scope changed? (No)

## Compatibility / Migration

- Backward compatible? (Yes)
- Config/env changes? (No)
- Database migration needed? (No)

- Verified scenarios: direct AI node, loop wrapper, and loop-group completion events retain their reported or aggregated token objects.
- Edge cases checked: absent token usage remains absent rather than being converted to a synthetic zero or `undefined` event field.
- What was not verified: live remote-console rendering was not manually exercised; it reads the unchanged event-data contract.

- Affected subsystems/workflows: workflow DAG execution and consumers of persisted `node_completed` event data.
- Potential unintended effects: event records carrying reported usage gain a small optional JSON object.
- Guardrails/monitoring for early detection: targeted regression tests assert exact direct and aggregate token payloads; optional spreading preserves no-usage behavior.

## Rollback Plan (required)

- Fast rollback command/path: revert commit `aa11bf1a` from `dev` if event payload expansion causes an integration issue.
- Observable failure symptoms: consumers reject an added optional `tokens` property or display unexpected node usage values.

## Risks and Mitigations

- Risk: a provider reports no usage or an incomplete usage object.
  - Mitigation: persist only non-undefined values and preserve the provider-normalized object without recalculation.

### DIFF
--- a/packages/core/src/db/workflows.ts
+++ b/packages/core/src/db/workflows.ts
@@ -965,6 +965,7 @@ export async function pauseWorkflowRun(
             // null as absent (`!= null`).
             completionSignaled: approvalContext.completionSignaled ?? null,
             signaledOutput: approvalContext.signaledOutput ?? null,
+            signaledTokens: approvalContext.signaledTokens ?? null,
             onRejectPrompt: approvalContext.onRejectPrompt ?? null,
             onRejectMaxAttempts: approvalContext.onRejectMaxAttempts ?? null,
             captureResponse: approvalContext.captureResponse ?? null,
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -4802,6 +4802,84 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
     expect(completedEvent?.[0].data).not.toHaveProperty('tokens');
   });
 
+  it('persists only {input, output} — provider-defined total/cost are not part of the shape', async () => {
+    mockSendQueryDag.mockImplementation(function* () {
+      yield { type: 'assistant', content: 'out' };
+      // Pi/OpenCode shape: `total` folds in cache/reasoning tokens, so it is NOT
+      // input + output. Persisting it would hand consumers a field they cannot
+      // interpret without knowing the provider; `cost` duplicates cost_usd.
+      yield {
+        type: 'result',
+        sessionId: 'shape-sid',
+        tokens: { input: 100, output: 10, total: 900, cost: 0.5 },
+      };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-shape',
+      testDir,
+      { name: 'token-shape', nodes: [{ id: 'step1', command: 'step1' }] },
+      makeWorkflowRun('token-shape-run'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls as Array<
+      [{ event_type: string; step_name: string; data?: Record<string, unknown> }]
+    >;
+    const completedEvent = eventCalls.find(
+      ([event]) => event.event_type === 'node_completed' && event.step_name === 'step1'
+    );
+    expect(completedEvent?.[0].data?.tokens).toEqual({ input: 100, output: 10 });
+  });
+
+  it('drops non-finite provider token counts instead of persisting them', async () => {
+    mockSendQueryDag.mockImplementation(function* () {
+      yield { type: 'assistant', content: 'out' };
+      yield { type: 'result', sessionId: 'nan-sid', tokens: { input: NaN, output: 10 } };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-nan',
+      testDir,
+      { name: 'nan-tokens', nodes: [{ id: 'step1', command: 'step1' }] },
+      makeWorkflowRun('nan-tokens-run'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls as Array<
+      [{ event_type: string; step_name: string; data?: Record<string, unknown> }]
+    >;
+    const completedEvent = eventCalls.find(
+      ([event]) => event.event_type === 'node_completed' && event.step_name === 'step1'
+    );
+    expect(completedEvent).toBeDefined();
+    // A NaN would serialize to `{input: null, output: 10}` — a wrong number that
+    // gets believed. Absence is the honest answer.
+    expect(completedEvent?.[0].data).not.toHaveProperty('tokens');
+  });
+
   // ─── Background Agent Task Gating (#2083) ───────────────────────────────
 
   describe('background task completion gating (#2083)', () => {
@@ -6485,7 +6563,7 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
           type: 'assistant',
           content: 'Plan approved. Proceeding. <promise>APPROVED</promise>',
         };
-        yield { type: 'result', sessionId: 'loop-session-2' };
+        yield { type: 'result', sessionId: 'loop-session-2', tokens: { input: 40, output: 4 } };
       });
 
       const mockDeps = createMockDeps();
@@ -6538,6 +6616,9 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
         // The gate persists the signal state (#2074) so a bare approve can
         // finalize at resume instead of re-running the iteration.
         completionSignaled: true,
+        // ...and the usage consumed up to the gate (#2333), so the finalize path
+        // does not report a silent zero for iterations that really ran.
+        signaledTokens: { input: 40, output: 4 },
       });
       const signaledOutput = (pauseCalls[0][1] as { signaledOutput: string }).signaledOutput;
       expect(signaledOutput).toContain('Plan approved');
@@ -6753,6 +6834,7 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
             message: 'gate',
             completionSignaled: true,
             signaledOutput: 'REPORT',
+            signaledTokens: { input: 40, output: 4 },
           },
           loop_user_input: 'Approved',
           loop_feedback_given: false,
@@ -6805,6 +6887,81 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
       );
       expect(completed.length).toBe(1);
       expect(completed[0][0].data.node_output).toBe('REPORT');
+      // The finalize row reports the usage the pausing invocation consumed (#2333).
+      // Without this it persists duration_ms: 0 and no tokens for iterations that
+      // really ran — a silent zero, not an absence.
+      expect(completed[0][0].data.tokens).toEqual({ input: 40, output: 4 });
+    });
+
+    it('finalize omits tokens when the gate persisted none (legacy pause / no usage) (#2333)', async () => {
+      mockSendQueryDag.mockImplementation(function* () {
+        yield { type: 'assistant', content: 'should never run' };
+        yield { type: 'result', sessionId: 'never' };
+      });
+
+      const mockDeps = createMockDeps();
+      const workflowRun = makeWorkflowRun('finalize-no-tokens-run', {
+        metadata: {
+          approval: {
+            type: 'interactive_loop',
+            nodeId: 'refine',
+            iteration: 1,
+            sessionId: 'sig-session-1',
+            message: 'gate',
+            completionSignaled: true,
+            signaledOutput: 'REPORT',
+            // No signaledTokens key at all — a run paused by a build predating #2333.
+          },
+          loop_user_input: 'Approved',
+          loop_feedback_given: false,
+        },
+      });
+
+      await executeDagWorkflow(
+        mockDeps,
+        createMockPlatform(),
+        'conv-dag',
+        testDir,
+        {
+          name: 'finalize-on-approve',
+          nodes: [
+            {
+              id: 'refine',
+              loop: {
+                prompt: 'Refine.',
+                until: 'APPROVED',
+                max_iterations: 10,
+                interactive: true,
+                gate_message: 'Review.',
+              },
+            },
+          ],
+        },
+        workflowRun,
+        'claude',
+        undefined,
+        join(testDir, 'artifacts'),
+        join(testDir, 'logs'),
+        'main',
+        'docs/',
+        minimalConfig
+      );
+
+      expect(mockSendQueryDag.mock.calls.length).toBe(0);
+      const eventCalls = (
+        mockDeps.store.createWorkflowEvent as Mock<
+          (e: {
+            event_type: string;
+            step_name: string;
+            data: Record<string, unknown>;
+          }) => Promise<void>
+        >
+      ).mock.calls;
+      const completed = eventCalls.filter(
+        c => c[0].event_type === 'node_completed' && c[0].step_name === 'refine'
+      );
+      expect(completed.length).toBe(1);
+      expect(completed[0][0].data).not.toHaveProperty('tokens');
     });
 
     it('iterates at resume when feedback was given, even on a signal-bearing gate (#2074 C)', async () => {
@@ -14519,14 +14676,37 @@ describe('executeDagWorkflow -- loop_group node', () => {
     const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls as Array<
       [{ event_type: string; step_name: string; data?: Record<string, unknown> }]
     >;
-    const completedEvent = eventCalls.find(
+    // The BODY node's per-iteration rows are the authoritative per-node usage.
+    const bodyEvents = eventCalls.filter(
+      ([arg]) => arg.event_type === 'node_completed' && arg.step_name === 'paid.work'
+    );
+    expect(bodyEvents.map(([arg]) => arg.data?.tokens)).toEqual([
+      { input: 100, output: 10 },
+      { input: 200, output: 20 },
+    ]);
+    // The GROUP row must NOT repeat the same total under the same field name: body
+    // rows and the aggregate live in one event stream, so a consumer summing
+    // `data.tokens` would otherwise count this group twice (600/60 for 300/30).
+    const groupEvent = eventCalls.find(
       ([arg]) => arg.event_type === 'node_completed' && arg.step_name === 'paid'
     );
-    expect(completedEvent).toBeDefined();
-    expect(completedEvent?.[0].data?.tokens).toEqual({ input: 300, output: 30 });
+    expect(groupEvent).toBeDefined();
+    expect(groupEvent?.[0].data).not.toHaveProperty('tokens');
+    // The property the persisted stream must hold: a naive consumer summing every
+    // node_completed row's tokens gets the run's real usage, with no discriminator.
+    const naiveSum = eventCalls
+      .filter(([arg]) => arg.event_type === 'node_completed')
+      .reduce(
+        (acc, [arg]) => {
+          const t = arg.data?.tokens as { input: number; output: number } | undefined;
+          return t ? { input: acc.input + t.input, output: acc.output + t.output } : acc;
+        },
+        { input: 0, output: 0 }
+      );
+    expect(naiveSum).toEqual({ input: 300, output: 30 });
   });
 
-  it('omits tokens from a loop_group node_completed event when providers report no usage', async () => {
+  it('omits tokens from loop_group body node_completed events when providers report no usage', async () => {
     mockSendQueryDag.mockImplementation(function* () {
       yield { type: 'assistant', content: 'done\nDONE' };
       yield { type: 'result', sessionId: 'lg-no-usage-sid' };
@@ -14566,11 +14746,17 @@ describe('executeDagWorkflow -- loop_group node', () => {
     const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls as Array<
       [{ event_type: string; step_name: string; data?: Record<string, unknown> }]
     >;
-    const completedEvent = eventCalls.find(
+    const bodyEvent = eventCalls.find(
+      ([event]) =>
+        event.event_type === 'node_completed' && event.step_name === 'no-usage-group.work'
+    );
+    expect(bodyEvent).toBeDefined();
+    expect(bodyEvent?.[0].data).not.toHaveProperty('tokens');
+    const groupEvent = eventCalls.find(
       ([event]) => event.event_type === 'node_completed' && event.step_name === 'no-usage-group'
     );
-    expect(completedEvent).toBeDefined();
-    expect(completedEvent?.[0].data).not.toHaveProperty('tokens');
+    expect(groupEvent).toBeDefined();
+    expect(groupEvent?.[0].data).not.toHaveProperty('tokens');
   });
 
   it('SESSION: fresh_context=false threads the body session between iterations', async () => {
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -4767,6 +4767,41 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
     ).toEqual({ input: 100, output: 10 });
   });
 
+  it('omits tokens from a direct AI node_completed event when the provider reports no usage', async () => {
+    mockSendQueryDag.mockImplementation(function* () {
+      yield { type: 'assistant', content: 'the node output text' };
+      yield { type: 'result', sessionId: 'no-usage-sid' };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-no-usage',
+      testDir,
+      { name: 'no-usage', nodes: [{ id: 'step1', command: 'step1' }] },
+      makeWorkflowRun('no-usage-run'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls as Array<
+      [{ event_type: string; step_name: string; data?: Record<string, unknown> }]
+    >;
+    const completedEvent = eventCalls.find(
+      ([event]) => event.event_type === 'node_completed' && event.step_name === 'step1'
+    );
+    expect(completedEvent).toBeDefined();
+    expect(completedEvent?.[0].data).not.toHaveProperty('tokens');
+  });
+
   // ─── Background Agent Task Gating (#2083) ───────────────────────────────
 
   describe('background task completion gating (#2083)', () => {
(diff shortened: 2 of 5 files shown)


Wirasm:
# Consolidated Review: PR #2347

**Date**: 2026-07-30T17:24:44+03:00  
**Total Findings**: 1 from 2 available agent artifacts

---

## Executive Summary

This is a small, well-scoped change that persists already-captured token usage on successful direct AI, `loop`, and `loop_group` completion events. The code review found no correctness, security, type-safety, or maintainability concerns, and focused, type, lint, format, and build validation all passed. The test review confirms strong positive-path coverage, including loop-group aggregation totals. It recommends one additional boundary test: assert that the `tokens` key is omitted—not present as `undefined` or a synthetic zero value—when a provider reports no usage. Error-handling, comment-quality, and docs-impact artifacts were not available when this synthesis was produced, so their review results are explicitly not represented.

**Overall Verdict**: NEEDS_DISCUSSION

**Auto-fix Candidates**: 0 CRITICAL + HIGH issues can be auto-fixed  
**Manual Review Needed**: 1 MEDIUM issue requires a decision; 3 specialist artifacts are missing

---

## Statistics

| Code Review | 0 | 0 | 0 | 0 | 0 |
| Error Handling | N/A | N/A | N/A | N/A | Unavailable |
| Test Coverage | 0 | 0 | 1 | 0 | 1 |
| Comment Quality | N/A | N/A | N/A | N/A | Unavailable |
| Docs Impact | N/A | N/A | N/A | N/A | Unavailable |
| **Total (available artifacts)** | **0** | **0** | **1** | **0** | **1** |

---

## CRITICAL Issues (Must Fix)

No critical issues were reported by the available artifacts.

---

## HIGH Issues (Should Fix)

No high-severity issues were reported by the available artifacts.

---

### Issue 1: Optional token-key omission is not regression-tested

**Source Agent**: test-coverage-agent  
**Location**: `packages/workflows/src/dag-executor.ts:2253,3509,4655` and `packages/workflows/src/dag-executor.test.ts`

**Problem**:

The new tests verify token values when usage is present, but do not assert that `node_completed.data` has no own `tokens` property when a provider reports no usage. The three completion writers are independent, so a later change could emit `tokens: undefined` or a synthetic zero object without these tests catching it.

**Options**:

| Fix Now | Add no-usage completion assertions for direct AI, `loop`, and `loop_group`, checking that `data` does not own `tokens`. | MED | Consumers could lose the distinction between unreported and zero usage. |
| Create Issue | Defer the boundary cases to a focused test-coverage PR. | LOW | Same risk remains until addressed. |
| Skip | Accept the existing positive-path coverage. | NONE | Future refactors could alter the optional-field contract unnoticed. |

**Recommendation**: Fix now. Three small executor-level assertions match the PR's three-path scope and protect the persisted event contract without adding production complexity.

---

No low-severity issues were reported by the available artifacts.

---

## Positive Observations

- The production change is localized to the three intended successful AI-derived completion paths.
- It reuses existing token aggregation and the established conditional-property-spread convention.
- Tests inspect persisted `node_completed` event data, which directly exercises the regression boundary.
- Direct AI, loop, and loop-group aggregate values are covered, including exact loop-group totals.
- Focused executor tests, type-checking, linting, format checking, and build validation passed.

---

## Suggested Follow-up Issues

| Add no-usage token-omission regression coverage for workflow completion events | P2 | MEDIUM issue #1 |

---

## Next Steps

1. Add or explicitly defer the three no-usage token-omission assertions.
2. Produce the missing error-handling, comment-quality, and docs-impact artifacts before treating this as a complete five-agent review.
3. Re-synthesize after those artifacts are available.

---

## Agent Artifacts

| Code Review | `code-review-findings.md` | 0 |
| Error Handling | `error-handling-findings.md` | Unavailable |
| Comment Quality | `comment-quality-findings.md` | Unavailable |
| Docs Impact | `docs-impact-findings.md` | Unavailable |

---

## Metadata

- **Synthesized**: 2026-07-30T17:24:44+03:00
- **Artifact**: `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/d4e8cf9d3a97dd37bac9351adda25a6a/review/consolidated-review.md`
- **Completeness**: 2 of 5 requested agent artifacts were present.

Wirasm:
# Fix Report: PR #2347

**Date**: 2026-07-30T14:27:59Z
**Status**: COMPLETE
**Branch**: archon/task-archon-fix-github-issue-experimental-1785419683548
**Commit**: e07dfa8e12fbf33c71fb447d5fb8811f2f73fe9e

---

## Summary

Review artifacts reported one medium-severity coverage gap: successful completion events without provider usage were not asserted to omit `tokens`. Added direct AI, `loop`, and `loop_group` regression coverage for that public event-shape contract. No findings were skipped or blocked; no documentation changes were needed.

---

## Fixes Applied

| MEDIUM | Optional token-key omission is not regression-tested | `packages/workflows/src/dag-executor.test.ts:4767` | Added a direct-AI no-usage case that asserts `node_completed.data` has no `tokens` property. |
| MEDIUM | Optional token-key omission is not regression-tested | `packages/workflows/src/dag-executor.test.ts:5239` | Extended the existing no-usage loop case to assert `tokens` is omitted. |
| MEDIUM | Optional token-key omission is not regression-tested | `packages/workflows/src/dag-executor.test.ts:14526` | Added a loop-group no-usage case that asserts `node_completed.data` has no `tokens` property. |

---

## Tests Added

| File | Test Cases |
|------|------------|
| `packages/workflows/src/dag-executor.test.ts` | Direct AI and loop-group no-usage completion events omit `tokens`; existing loop no-usage scenario now asserts the same. |

---

## Docs Updated

*(none)*

---

## Skipped Findings

*(none)*

---

## Blocked (Could Not Fix)

*(none)*

---

## Suggested Follow-up Issues

*(none)*

---

## Validation

| Check | Status |
|-------|--------|
| Type check | ✅ |
| Lint | ✅ |
| Focused executor tests | ✅ 436 passed |
| Workspace tests | ✅ |

Wirasm:
---
pr: 2347
title: "fix: Persist per-node workflow token usage"
author: "Wirasm"
reviewed: 2026-07-30
---

# PR Review: #2347 — Persist per-node workflow token usage

**Branch**: `archon/task-archon-fix-github-issue-experimental-1785419683548` → `dev` (draft)
**Files**: 2 (+107/−2) — 3 production lines in `packages/workflows/src/dag-executor.ts`, rest tests
**Closes**: #2333

---

## Verdict

**NEEDS FIXES** — the change is correct as far as it goes, and the aggregation semantics are sound. But it covers 3 of the 4 AI-derived `node_completed` paths, and on the `loop_group` path it makes the persisted event stream double-count, which undercuts the issue's own motivating use case.

---

## Execution results (facts, separate from reasoning below)

Isolated worktree at `/Users/rasmus/Projects/cole/Archon-2347review` off `pr-2347-review` (e07dfa8e).

### Item 3 — do the tests fail against `origin/dev`? REPRODUCED

Reverted **only** `packages/workflows/src/dag-executor.ts` to `origin/dev` (`git diff --stat` confirmed exactly `1 file changed, 3 deletions(-)`), kept all new tests:

| Test | Reverted | With fix |
|---|---|---|
| `stores node_output in node_completed event data for AI nodes` | **FAIL** | PASS |
| `records requested model/tier on node_started and the resolved model on node_completed (#2314)` | **FAIL** | PASS |
| `COST: accumulates token usage across loop_group iterations` | **FAIL** | PASS |
| `omits tokens from a direct AI node_completed event when the provider reports no usage` | pass | pass |
| `omits tokens from a loop_group node_completed event when providers report no usage` | pass | pass |

Each of the three production sites has a test that genuinely fails without it. The two `omits tokens…` tests pass either way — correctly so (they are absence-contract guards), but they prove nothing about the fix itself and should not be counted as coverage of it.

Full file restored: **436 pass / 0 fail**.

### Validation

| Check | Result |
|---|---|
| `bun run type-check` | PASS (all 13 packages) |
| `bun run lint` | PASS (clean) |
| `bun --filter @archon/workflows test` | PASS (exit 0) |
| `bun test packages/workflows/src/dag-executor.test.ts` | 436/436 |

### Empirical

`~/.archon/archon.db` (read-only): **548** `node_completed` events, **0** carrying `tokens`. The gap is real.

---

## Item 1 — aggregation semantics. SETTLED: the sum is correct, no double-count.

`executeLoopNode` already carries the #2083 fix for tokens:

- `dag-executor.ts:4199-4210` — within an iteration, `iterationTokens` uses **overwrite** semantics (`iterationTokens = {…}`), with an explicit comment that a later result chunk from a background-task wait carries session-cumulative values.
- `dag-executor.ts:4070-4085` — `foldIterationUsage()` folds the last-seen per-iteration value into `loopTotalTokens` **exactly once**, latched by `iterationUsageFolded`.

So a second result chunk in one iteration overwrites rather than adds. The hazard you flagged was already handled, and it carries a `Number.isFinite` guard the direct path does not (see S2).

`executeLoopGroupNode` (`dag-executor.ts:3327-3331`) sums `iterCtx.totalTokensIn/Out` with a fresh `iterCtx` per iteration — a genuine per-iteration sum.

Summing tokens while `resolvedModel` stays last-seen-wins is deliberate and right.

## Item 2 — `loop_group` bodies route through `executeNodeInternal`, and 300/30 is a real sum.

The fixture yields **distinct** values per iteration (`{100,10}` then `{200,20}`) and pins `expect(calls).toBe(2)`, so 300 is provably `100+200`, not a fixture coincidence. Body nodes reach `iterCtx.totalTokensIn` via `NodeOutput.tokens` at `dag-executor.ts:6220-6221`.

---

## Issues Found

### Important

**I1 — `loop_group` double-counts in the persisted event stream.**
`packages/workflows/src/dag-executor.ts:3509`

I instrumented the loop_group test and dumped every `node_completed` the run persists:

```
paid.work → {"input":100,"output":10}   ← iteration 1 body node
paid.work → {"input":200,"output":20}   ← iteration 2 body node
paid      → {"input":300,"output":30}   ← group aggregate (new)
```

**Why**: a consumer summing `node_completed.tokens` over a run gets `{600,60}` for `{300,30}` of real usage — exactly 2×. Leaf rows and the aggregate row use the **same field name** with no discriminator; the only signal is that `step_name` contains a `.`, which is undocumented and fragile. This is precisely the consumer #2333 exists to serve ("benchmark query", "per-run cost/efficiency reporting"). Note `loop` does **not** have this problem — its per-iteration events are `loop_iteration_completed` (`:4618`) and carry no tokens; only `loop_group` emits both levels. Run-level totals stay correct (aggregated once at `:6215`); this is an event-stream-only defect.

**Fix**: either drop `tokens` from the group-level event (bodies already report it, so the sum is recoverable), or mark the aggregate — e.g. `tokens_rollup: true` or `type: 'loop_group'` — so consumers can filter. Whichever you pick, say it in the event-shape docs.

---

**I2 — the `workflow:` sub-run node is a 4th AI-derived path, and it was missed.**
`packages/workflows/src/dag-executor.ts:5147`

```ts
data: {
  node_output: output,
  type: 'workflow',
  child_run_id: outcome.childRunId,
  ...(outcome.costUsd !== undefined ? { cost_usd: outcome.costUsd } : {}),
  //  ← outcome.tokens dropped here
},
```

…while the **same function**, 29 lines later at `:5176`, already does:

```ts
...(outcome.tokens !== undefined ? { tokens: outcome.tokens } : {}),
```

`ChildWorkflowOutcome.tokens` is declared at `:325` and populated by `childOutcomeFromRun` (`:364-388`) from the child's `total_tokens_in`/`total_tokens_out` metadata. The value is live and in scope.

**Why**: this is the identical defect #2333 describes — captured, threaded to the return value, dropped from the persisted event. It is also backwards relative to the issue's own thesis: the sub-run node propagates **cost** (which Codex cannot report, so it is not comparable) and drops **tokens** (the one axis that is). A workflow using `workflow:` nodes gets a queryable event stream with cost but no tokens on exactly the nodes that did the work.

**Fix**: one-line spread, mirroring `cost_usd`, plus a test.

---

**I3 — `finalizeLoopFromSignal` emits a token-less `node_completed`.**
`packages/workflows/src/dag-executor.ts:3049`

```ts
data: { duration_ms: 0, node_output: finalizeOutput },
```

This is the #2074 path: on an interactive loop gate whose iteration already emitted the completion signal, approving with **no text** finalizes the node from the already-computed output without a new iteration. The function signature (`:3021-3030`) takes no usage parameters at all, so tokens are unreachable from inside it — but both call sites (`:3153`, `:3894`) have `loopTotalTokens` in scope.

**Why**: an interactive loop that finalizes at the gate persists `duration_ms: 0` and **no tokens**, despite having consumed them across its iterations. Silent — the operator sees a node that apparently cost nothing. Pre-existing (`cost_usd` is missing too, so this is not a regression), but it sits squarely inside this PR's stated scope of "loop completion events", and the run reports a number that will be believed.

**Fix**: thread the loop totals into the finalize helper and spread them like the other sites; or, if deliberate, comment why this path reports no usage.

---

### Suggestions

**S1 — the persisted `tokens` shape is not uniform across the three patched sites.**
`dag-executor.ts:2253` vs `:3509` / `:4655`

The direct-node path persists the **raw provider object**. `TokenUsage` (`packages/providers/src/types.ts:167-172`) is `{input, output, total?, cost?}`, and real logs confirm the 4-field form is common:

```json
{"input":2592,"output":170,"total":2762,"cost":0.0009816}
```

The loop paths instead **construct** `{input, output}` only, dropping `total`/`cost`. That matters because `total` is not `input + output`: Pi's `usageToTokens` (`community/pi/event-bridge.ts:105-112`) sets `total` from `usage.totalTokens`, which includes cache tokens, and OpenCode does the same (its own test maps `{input:11, output:7, reasoning:3, cache:1}` → `total: 21`). I found a real log line where `total` is 2759 against `input+output` of 172.

So one field name now carries two different accounting bases depending on node type, and `tokens.cost` duplicates the sibling `cost_usd`. On **payload size** (your item 6) the PR is fine — `TokenUsage` is capped at 4 numeric fields and the richer cache/reasoning breakdown is already discarded at the provider boundary, so nothing unbounded rides along. The problem is consistency, not size.

**Fix**: normalise at all three sites — either persist `{input, output}` everywhere, or carry `total`/`cost` everywhere.

**S2 — the new direct-node site bypasses the repo's NaN guard.**
`dag-executor.ts:1644` → `:2253`

Capture is `if (msg.tokens) nodeTokens = msg.tokens;` — no finiteness check. The repo guards this value twice elsewhere, deliberately:

- `:6215` — *"Token values come from providers (incl. community ones) — guard so a NaN can't silently poison the totals"*
- `:4202-4210` — the loop node's per-iteration capture, which references that rationale

Before this PR, `nodeTokens` only reached the JSONL log and the guarded run-level aggregator. It now goes **straight to the store**, unguarded. `JSON.stringify(NaN)` is `null`, so a misbehaving provider yields a persisted `tokens: {input: null, output: null}`. The loop paths are protected; the direct path is not.

**Fix**: reuse the same `Number.isFinite` check at capture, with the existing `usage_tokens_non_finite_ignored` warn.

**S3 — zero-usage asymmetry (your item 4).**
Absence **is** preserved when the provider omits `msg.tokens` — I verified both new tests and confirmed no path emits `tokens: undefined`. But a provider that reports an explicit `{input: 0, output: 0}` is truthy at `:1644`, so the direct node persists `{input:0,output:0}`, while `loop_group`'s `> 0` guard (`:3327`) skips it and omits the field. The same zero report is therefore recorded as "zero" by one node type and "unreported" by another. Not the fabricated-number bug you were worried about, but the same family.

---

### Strengths

- Finding and fixing the `loop_group` path unprompted — the exact gap #2337 was sent back for.
- Reuses the established conditional-spread convention; zero new production complexity.
- Tests assert against `deps.store.createWorkflowEvent` (the **persisted** row), not the in-process emitter — which is what actually matters, and confirms your item 5.
- The loop_group fixture uses distinct per-iteration values, so it tests summation rather than a repeated constant. That is better than the average test in this file.
- Absence-contract tests were added on review feedback rather than argued away.

---

## Notes on the issue spec

Beyond the loop-path omission already identified, #2333 was incomplete in three ways:

1. It missed the `workflow:` sub-run node (I2) — an AI-derived completion event with `tokens` already in scope and its sibling `cost_usd` already handled.
2. It missed `finalizeLoopFromSignal` (I3).
3. It proposed the one-line spread without noticing that adding `tokens` to **both** body and aggregate rows makes the `loop_group` event stream non-summable (I1) — which undercuts the benchmarking motivation the issue is built on.

One factual correction: the issue lists "Roll-up on the run" as still to be decided. It already exists — `total_tokens_in` / `total_tokens_out` are written to run metadata at `dag-executor.ts:7300-7301`, and `childOutcomeFromRun` reads them back. That part of the spec is asking for something already built.

---

## Recommendation

**NEEDS FIXES.** I2 is a one-line fix with the value in scope and should land with this PR — leaving it means shipping the fix for #2333 while an equivalent instance survives in the same file. I1 needs a decision (drop the aggregate, or mark it) before consumers start summing. I3 and S2 are small and worth folding in. S1/S3 are consistency cleanups that can be deferred if you prefer to keep the diff tight, but they should be a follow-up issue rather than nothing.

*Reviewed by Claude*

aberkowCollibra:
This would be extremely useful and is a blocker from my team using this tool right now. We want to track token usage per node.

Wirasm:
# Second-pass review — scoped to the fix commits (`e07dfa8e`, `519d2479`)

**Verdict: NEEDS FIXES** — one blocking defect.

The first review's finding (B1) was real and the chosen remedy is right. But the same defect now exists on a different path: **B3 reintroduces the loop_group double-count at the finalize-on-approve gate.**

---

## C1 (blocking) — `finalizeLoopFromSignal` re-reports loop_group body usage

`finalizeLoopFromSignal` is shared by `executeLoopNode` and `executeLoopGroupNode`. B3 added `tokens` to the `node_completed` row it writes. That is correct for a plain `loop` (its per-iteration rows carry no tokens, so this is the only row). It is **not** correct for a `loop_group`: the body nodes already wrote their own namespaced `<groupId>.<nodeId>` rows carrying those tokens *before* the pause, and those rows survive it. The finalize row then re-reports the same usage under the same field name — exactly the shape the comment at `dag-executor.ts:3555` exists to prevent.

Measured, not argued. One shared event store across pause → bare-approve → resume, so the persisted stream matches a real DB. Provider reports `{input: 100, output: 10}` on the single body iteration; `interactive: true`, no `signal_completes`, so iteration 1 signals and gates; resume with no feedback finalizes.

```
gate approval context signaledTokens = {"input":100,"output":10}

phase 1 (paused at gate) — persisted node_completed rows
  step_name=refine.work    tokens={"input":100,"output":10}
  NAIVE SUM: {"input":100,"output":10}          <- correct

phase 2 (after bare-approve finalize) — persisted node_completed rows
  step_name=refine.work    tokens={"input":100,"output":10}
  step_name=refine         tokens={"input":100,"output":10}   <- the finalize row
  NAIVE SUM: {"input":200,"output":20}          <- 2x, for 100/10 of real usage
```

For contrast, the path the fix *did* address is now correct:

```
loop_group natural completion (no gate)
  step_name=grp.work       tokens={"input":100,"output":10}
  step_name=grp            tokens=null
  NAIVE SUM: {"input":100,"output":10}          <- correct
```

**Fix:** drop the token argument at the loop_group call site only — `dag-executor.ts:3198-3208` should not pass `readSignaledTokens(loopGateMeta.signaledTokens)`. Keep it at `executeLoopNode`'s call site (`:3957-3966`), which is correct. `signaledTokens` at the loop_group gate pause (`:3655`) then has no consumer and should go with it; the `ApprovalContext` field and the `pauseWorkflowRun` explicit-null reset stay, since the plain loop needs them.

Worth extending the `:3555` comment to say the rule binds *every* row written for a loop_group node, not just that one — a per-call-site invariant is what let this through.

**Test gap that allowed it:** the naive-sum assertion added to `COST: accumulates token usage across loop_group iterations` is the strongest test in the fix and genuinely fails against `dev` — but it only covers natural completion. The missing case is the same assertion on the loop_group finalize path.

---

## I1 — the `signaledTokens` doc comment claims a recoverability that does not exist

`packages/workflows/src/schemas/workflow-run.ts`:

> a loop that gated more than once attributes each invocation's usage to that invocation, and earlier ones are only visible in the run-level totals

Measured: they are not visible anywhere. A pausing invocation never reaches `completeWorkflowRun` — `skipIfStatusChanged` bails on a non-`running` status, and `total_tokens_in`/`total_tokens_out` are only written there (`:7373`). So for a loop that gates twice:

```
invocation 1: iteration 1 consumes 40/4, no signal  -> gate. No node_completed row.
                                                       No run-row totals (run is paused).
invocation 2: iteration 2 consumes 60/6, signals    -> completes.
  persisted node_completed:  step_name=refine  tokens={"input":60,"output":6}
  run row total_tokens_*:    60/6
```

Node truly consumed 100/10. The 40/4 is in neither the node row nor the run row.

The under-report itself is pre-existing (`dev` persisted nothing at all here, so this is not a regression) and belongs with #2345. But in a PR whose purpose is per-node usage visibility for cross-provider comparison, a comment asserting the loss is covered will mislead whoever next reads these numbers. Correct the comment or move the claim into #2345.

---

## Verified correct

- **Run-level roll-up is unaffected by the aggregate omission.** It sums node *return* values at the single aggregation point `:6288`, not persisted rows, and `executeLoopGroupNode` still returns `tokens` (`:3591`, `:3661`). Body results land in the scoped iteration ctx and never reach the run ctx. Confirmed by the natural-completion dump above.
- **Plain `loop` still emits tokens exactly once** — verified across a real gate pause + finalize: one row, `{input: 40, output: 4}`, no second emission.
- **S1 (shape consistency) is fixed.** The direct-node capture is normalized to `{input, output}` with a finite guard (`:1641-1660`), so all four persist sites now emit the same two-field shape. A consumer can no longer be handed a provider-defined `total` it cannot interpret. This was the right call — the alternative (persisting `total`) would have made the benchmarking use case unusable across providers.
- **B2 (`workflow:` sub-run) is correct** and cannot double count — the child's per-node rows are filed under a different `workflow_run_id`. Its test genuinely fails against `dev`.

## Test-quality note

Reverting only `dag-executor.ts` to `origin/dev` and keeping the tests: 6 fail in `dag-executor.test.ts`, 1 in `subrun.test.ts`. But **4 of the 5 newly-added tests pass either way** — `omits tokens from a direct AI node_completed…`, `drops non-finite provider token counts…`, `finalize omits tokens when the gate persisted none…`, `omits tokens from loop_group body node_completed events…`. They are correct as absence guards but prove nothing about the change; the same property the first review flagged. Only `persists only {input, output}…` actually fails.

## Validation

| Check | Result |
|---|---|
| `bun run type-check` | PASS |
| `bun run format:check` | PASS |
| `bun --filter @archon/workflows test` | PASS, 0 failures |
| `packages/core/src/db/workflows.test.ts` | 82 pass (in isolation) |
| `packages/core/src/workflows/store-adapter.test.ts` | 13 pass (in isolation) |

PR #2199, opened 2026-07-20
fix(workflows): keep .archon telemetry out of target-repo PRs

Wirasm:
## Summary

Describe this PR in 2-5 bullets:

- Problem: The bundled idea-to-pr / plan-to-pr family could commit Archon telemetry into the target repo's PR — the incident in #1407 landed 1,821 lines of `.archon/artifacts/` + `.archon/logs/` in a user's product repo. On current dev the engine writes run artifacts outside the repo, but repo-local `.archon/artifacts/`, `.archon/logs/`, and `.archon/state/` can still exist in target repos (older Archon layouts, the documented `.archon/state/` convention, user workflows) — and no bundled default told the agent to gitignore them or refuse to stage them.
- Why it matters: These paths are documented as "local-only, NEVER in git", yet a grep across bundled defaults for `gitignore` returned zero matches — an agent that sees them in `git status` has no rule against staging them.
- What changed: (1) `archon-implement-tasks.md` gains a "Repository Hygiene — Archon Telemetry" section: any `.gitignore` the run creates or modifies MUST include `.archon/artifacts/`, `.archon/logs/`, `.archon/state/`, and those paths must never be staged. (2) Every existing "Never stage" blocklist in bundled defaults (finalize-pr, create-pr, fix-issue, implement-issue, implement-review-fixes, simplify-changes, auto-fix-review, self-fix-all, fix-github-issue, piv-loop x3, ralph-dag, refactor-safely) now lists the repo-local `.archon/` telemetry paths. Regenerated `bundled-defaults.generated.ts`.
- What did **not** change (scope boundary): No engine/code changes — prompt-text only. `archon-ralph-generate`'s intentional `git add .archon/ralph/{slug}/` untouched (ralph specs are meant to be committed). `archon-adversarial-dev`'s `git add -A` untouched — it runs inside `$ARTIFACTS_DIR/app/`, a separate repo outside the target checkout. Human-facing conflict-recovery advice (`archon-plan-setup`, `archon-pr-review-scope`) untouched.

### Before

```
  User                          archon-plan-to-pr run                Target repo PR
  ────                          ─────────────────────                ──────────────
  runs plan-to-pr ───────────▶  implement-tasks scaffolds repo
                                (creates .gitignore, no .archon/
                                 entries mandated)
                                finalize-pr stages files seen
                                in git status ────────────────────▶  PR includes
                                                                     .archon/artifacts/*,
  merges PR ◀────────────────  (telemetry looks like work output)    .archon/logs/*.jsonl
[...]

PR #2358, opened 2026-08-01
fix(commands): tell archon-fix-issue that a dirty run worktree is expected

### LINKED ISSUE #2328: fix(commands): archon-fix-issue stops on a dirty worktree its own decision tree says to ignore
> **Rewritten 2026-07-30.** The original framing of this issue was wrong. It claimed the git-state checkpoint needed scoping to ignore `.archon/`. In fact **the checkpoint was never meant to apply inside a worktree at all** — the instructions already say so, and the model did not follow them. Corrected below; the original proposal (scope the checkpoint) would have fixed the wrong layer.

## Summary

`archon-fix-issue` halts with *"Blocked at the required git-state checkpoint"* when the run worktree has pre-existing modifications — even though **its own decision tree says a worktree needs no clean-tree check**. The whole pipeline then runs on an empty implementation.

This is an **instruction-following failure**, not a policy gap. Nothing needs authorising and no precondition needs relaxing.

## Evidence: both instruction sources already handle this correctly

`.archon/commands/defaults/archon-fix-issue.md` (Archon's own command):

```text
┌─ IN WORKTREE?
│  └─ YES → Use current branch AS-IS. Do NOT switch branches. …
│
├─ ON $BASE_BRANCH?
│  └─ Q: Working directory clean?
│     └─ NO → STOP: "Uncommitted changes on $BASE_BRANCH."
```

The `prp-issue` skill it loads (`workflows/fix.md:166-178`) says the same:

```text
┌─ IN WORKTREE?
│  └─ YES → Use it (assume it's for this work)
│
├─ ON {base-branch}?
│  └─ Q: Working directory clean?
│     └─ NO → … STOP
```

**The clean-tree requirement is on the `ON base-branch` arm only** — manual CLI use outside a worktree, where a stray edit really could be lost. The runs were all in worktrees, so the first arm applied and the check was irrelevant.

The model read the tree, was in a worktree, and stopped on a condition belonging to a different arm. **Three runs, both providers** (`f573847a` Claude, `84b41900` Codex, plus one more), same misapplication — so it is systematic, not a one-off sampling accident.

## Why the dirty tree exists at all (and should keep existing)

Archon copies the operator's `.archon/` — workflows, commands, scripts — into every run worktree **on purpose**, so a workflow can be iterated on and run before it is committed. Since worktrees are cut from `origin/dev`, any local `.archon/` edit necessarily shows as a modification inside the worktree.

Two distinct git states both trigger it, which matters because only one is avoidable:

- **untracked** additions (e.g. a new `.archon/workflows/<dir>/`) — can be silenced with `.git/info/exclude`
- **tracked-modified** files (a workflow you are actively editing) — **cannot**, because a file you intend to commit must stay visible in `git status`

So exclusion is not a general fix: it papers over exactly the files the copy feature exists to serve.

## Cost when it fires

Observed on `f573847a`: `implement` did nothing, `create-pr` correctly declined an empty PR, and **nine further nodes ran anyway** — the full review fleet against an empty diff — before the run reported `completed`. ~24 minutes, no output, green status. (The green-status half is #2327.)

## Fix, part 1 — SHIPPED

`archon-fix-issue.md` now opens with a **READ FIRST** precondition, before "Your Mission":

- states the worktree case affirmatively and first
- says a dirty tree is expected and is not a reason to stop
- marks `.archon/` modifications as never the implementer's to commit, stash, or remove
- scopes the clean-tree rule explicitly to `ON $BASE_BRANCH`
- **overrides any stricter git precondition imposed by a loaded skill**

The original text was correct but positioned badly: the worktree arm was one unemphasised line ~120 lines into a 483-line document, while the STOP case was visually loud. Bundle regenerated (`check:bundled` clean).

## Fix, part 2 — OPEN, and the durable one

A prompt fix reduces the probability of misreading; it does not remove the thing being misread. **Make the copied files structurally invisible to git instead:** after worktree-copy, run `git update-index --assume-unchanged` on the copied tracked paths in that worktree's index. `git status --porcelain` then returns nothing and there is no dirty state to misjudge — by any command, any skill, any provider, now or later.

Notes for whoever implements it:

- A linked worktree has its own index, so this does not leak to the operator's checkout.
- Untracked copies need different handling; `.git/info/exclude` resolves from the **common** git dir and is shared across worktrees, so a per-worktree exclude needs `extensions.worktreeConfig` + `core.excludesFile`. Worth confirming whether untracked copies actually occur in practice before building that half.
- Downside: a node that legitimately edits a `.archon/` file inside a run would have that change hidden from git. Acceptable — run worktrees have no business committing workflow edits.

## Worth naming as a class

Prompt-level "decision trees" whose arms carry different preconditions are read by a model that has to pick the right arm before applying the right rule. When the strict arm is the visually prominent one, the strict rule wins regardless of applicability. Both documents here were *correct*; the failure was entirely in emphasis. Any other bundled command with this shape has the same exposure.

## Related

- #2327 — nine nodes ran after the work had failed and the run went green. Independent of this, but it is why this took three runs to notice.
- #2326 — artifact-name mismatch, found in the same runs.

Wirasm:
## Summary

- **Problem:** `archon-fix-issue` stops at the implement step on a dirty working tree — a state its own decision tree explicitly permits. Archon copies the operator's `.archon/` directory into every run worktree by design, so the tree is *already* dirty before the implementing agent takes its first action.
- **Why it matters:** the run has, by that point, cloned or created a worktree, fetched the issue, classified it, and produced a plan. Stopping there discards the whole pipeline for a precondition that does not apply. It happened three times on `archon-fix-github-issue-experimental`.
- **What changed:** a `READ FIRST` preamble hoisted **above** the decision tree, scoping the clean-tree requirement to the `ON $BASE_BRANCH` case and stating that it overrides any stricter git precondition a loaded skill imposes. Bundle regenerated.
- **What did *not* change (scope boundary):** no engine, isolation, or executor code. No change to what Archon copies into a worktree (that is #2199's territory). No change to the decision tree itself — the tree was already correct. Prompt text plus its generated embed, nothing else.

### Before

```
Operator              Archon engine                 Implementing agent
────────              ─────────────                 ──────────────────
run fix-issue ─────▶  create worktree
                      copy .archon/ into it  ──┐
                      classify → plan          │  (tree is dirty
                      invoke archon-fix-issue ─┴─▶  before agent starts)
                                                   reads decision tree
                                                   sees "clean tree required"
                                                   *** STOPS ***
sees a failed run ◀── run fails at implement
  (plan + worktree discarded)
```

### After

```
Operator              Archon engine                 Implementing agent
────────              ─────────────                 ──────────────────
run fix-issue ─────▶  create worktree
                      copy .archon/ into it
                      classify → plan
                      invoke archon-fix-issue ───▶  *[reads READ FIRST]*
                                                    *[git worktree list]*
                                                    *[in worktree → dirty is expected]*
                                                    *[.archon/ is not mine to commit]*
                                                    implements on current branch
                                                    commits only its own files
sees a PR ◀────────── run completes
```

### Before

```
  .archon/commands/defaults/archon-fix-issue.md
        │  (source of truth for the command text)
        ▼
  scripts/generate-bundled-defaults.ts
        │
        ▼
  packages/workflows/src/defaults/bundled-defaults.generated.ts
        │
        ▼
  packages/workflows/src/defaults/bundled-defaults.ts
        │
        ▼
  command: node in a workflow  ──▶  provider prompt
```

### After

```
  [~] .archon/commands/defaults/archon-fix-issue.md
        │  (+ READ FIRST preamble, ahead of the decision tree)
        ▼
  scripts/generate-bundled-defaults.ts          (unchanged)
        │
        ▼
  [~] packages/workflows/src/defaults/bundled-defaults.generated.ts
        │  (one embedded string updated; 36 commands / 21 workflows, counts unchanged)
        ▼
  packages/workflows/src/defaults/bundled-defaults.ts   (unchanged)
        │
        ▼
  command: node in a workflow  ──▶  provider prompt
```

**Connection inventory:**

| `archon-fix-issue.md` | `generate-bundled-defaults.ts` | unchanged | same generator, same file set |
| `generate-bundled-defaults.ts` | `bundled-defaults.generated.ts` | unchanged | content of one embedded string differs |
| `bundled-defaults.generated.ts` | `bundled-defaults.ts` | unchanged | no shape or count change |
| `bundled-defaults.ts` | workflow `command:` node | unchanged | no new lookup path |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `skills`
- Module: `workflows:defaults`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2328
- Related #2199 (the `.archon/` copy that makes the tree dirty in the first place — this PR makes the agent tolerate it; #2199 addresses whether it should be copied)

```bash
bun run check:bundled        # up to date (36 commands, 21 workflows)
bun run check:bundled-skill  # up to date (23 files across 2 skills)
bun run type-check           # all 5 packages exit 0
bun run lint                 # clean
bun run format:check         # all matched files use Prettier code style
bun test packages/workflows/src/defaults/   # 17 pass, 0 fail, 242 expect()
```

- Evidence provided: command output above, run in a clean worktree cut from `origin/dev` at `94d5c2ea`.
- Intentionally skipped: the full `bun run test` suite. This PR changes one markdown prompt string and its generated embed. No test references `archon-fix-issue`; the tests that *do* assert on the bundle (`bundled-defaults.test.ts`, `generate-bundled-defaults.test.ts`) were run and pass, and CI runs the full suite regardless.

## Security Impact (required)

- New permissions/capabilities? **No** — but note the change *narrows* nothing and *widens* nothing operationally: it tells an agent not to touch `.archon/` files, which is strictly more restrictive than the prior text's silence on them.

## Compatibility / Migration

- Backward compatible? **Yes** — an operator with a repo-level `.archon/commands/archon-fix-issue.md` still overrides the bundled default by name, unchanged.
- Config/env changes? **No**

- **Verified scenarios:** the generated bundle contains the new text and no other default changed (`git diff -U0` on the generated file yields exactly one changed line). `check:bundled` passes both before and after `lint-staged`'s prettier pass, so the embed is not stale.
- **Edge cases checked:** neither changed file was touched by the 14 commits between the working copy's base and `origin/dev`, so this is a clean, non-racing apply. Bundle counts are identical (36/21), ruling out an accidental add or drop.
- **What was not verified:** *that the wording actually fixes the behaviour.* This is a prompt change; no test can assert an LLM's compliance with it. The supporting evidence is behavioural and circumstantial — runs after the edit stopped hitting the stop, and one completed the full pipeline unassisted — which is consistent with a fix but is not proof of causation. A future recurrence would mean the wording is insufficient, not that the diagnosis was wrong.

- **Affected subsystems/workflows:** every workflow whose `command:` node names `archon-fix-issue` and does not shadow it with a repo-local file. In the bundled set that is the `fix-github-issue` family.
- **Potential unintended effects:** the preamble instructs the agent not to commit anything under `.archon/`. A workflow whose *legitimate* purpose is editing `.archon/` — e.g. a workflow-authoring workflow — would be told not to commit its own work if it routed through this command. None of the bundled workflows do; `workflow-builder` does not use `archon-fix-issue`. Worth remembering if that changes.
- **Guardrails/monitoring:** a regression is directly visible — the run fails at the implement node with the agent reporting a dirty tree. `archon workflow get <id> --verbose` shows it immediately.

## Rollback Plan (required)

- Fast rollback: `git revert <sha>` — two files, no state, no migration. An operator can also override locally by dropping their own `.archon/commands/archon-fix-issue.md` without waiting for a release.
- Feature flags: none. `defaults.loadDefaultCommands: false` disables bundled commands wholesale, but that is not a targeted toggle for this.
- Observable failure symptoms: implement-step failures citing a dirty working tree, or — the new-failure direction — an agent committing `.archon/` files into a target-repo PR.

## Risks and Mitigations

- **Risk:** prompt fixes are unfalsifiable by CI; the wording may be insufficient under a different model or a differently-worded loaded skill.
  - **Mitigation:** the failure is loud and cheap to observe (run fails at a named node), and the override clause is stated explicitly rather than implied, which is the specific gap that produced the three stops.
- **Risk:** the "never commit `.archon/`" instruction is stated absolutely, and a future workflow may legitimately need to.
  - **Mitigation:** recorded above under blast radius; the correct fix in that case is a repo-local override of this command, which already works.

## Summary by CodeRabbit

* **Documentation**
  * Updated Archon workflow guidance for working with Git worktrees.
  * Added instructions for checking worktree status and handling pre-existing `.archon/` changes safely.
  * Clarified how to preserve existing branch state and unrelated edits during issue fixes.
  * Added guidance for staging implementation changes explicitly and applying clean-tree checks only in the primary checkout.


Wirasm:
Both findings were valid; fixed in 2f25a518.

**`git worktree list` cannot classify the checkout** — correct, and this was the more serious of the two. It lists every worktree *including the primary checkout*, so it reads identically from both, and an agent in the primary checkout could have claimed the worktree exemption and bypassed the clean-tree guard. Replaced with a `git rev-parse --git-dir` vs `--git-common-dir` comparison, which differs only inside a linked worktree. Verified empirically in three places: this PR's worktree (differ → linked), the primary checkout (equal → primary), and a live Archon-created run worktree under `~/.archon/workspaces/.../worktrees/` (differ → linked).

**Exemption breadth** — partially taken. The exemption itself has to stay broad, because resume reuses the worktree and the agent's own prior-attempt changes are legitimately dirty; requiring verification there would reintroduce the stop this PR exists to remove. What was genuinely missing is the commit-side guard, so that is now explicit: dirty paths outside `.archon/` are named as not-yours, staging by name is required over `git add -A`, and `git diff --cached --name-only` must be checked before every commit.

On the inconclusive linked-issue check: `bundled-defaults.generated.ts` is excluded by the generated-file path filter, so it cannot be reviewed inline. It was regenerated with `bun run generate:bundled` and verified with `bun run check:bundled` (up to date, 36 commands / 21 workflows — counts unchanged, confirming no default was accidentally added or dropped). CI runs `check:bundled` independently and fails on a stale bundle.

PR #2389, opened 2026-08-03
Fix piped workflow JSON output truncation

### LINKED ISSUE #2384: fix(cli): JSON output over a pipe silently truncates with exit 0 (bun EAGAIN drop)
**Any consumer reading Archon's JSON over a pipe can silently lose the tail of it, with exit status 0.**

## Measured

Reading `archon workflow status --json` and `archon workflow get <id> --json --verbose` from Swift via `Process` + `Pipe()`:

| route | complete | truncated |
|---|---|---|
| stdout → `Pipe()` | 11 / 16 | **5 / 16** |
| stdout → regular file | **16 / 16** | 0 |

Losses landed on **64 / 88 / 152 / 192 KiB** boundaries, and **exit status was 0 on every truncated run**. No error, no short-read signal, nothing a caller can branch on — the JSON simply ends mid-document.

## Cause

Bun's open `EAGAIN` drop on a non-blocking pipe — [oven-sh/bun#21516](https://github.com/oven-sh/bun/issues/21516), PR [#33560](https://github.com/oven-sh/bun/pull/33560): *"discarded the unwritten tail with no retry and no error… exit code 0"*. Still reproducing on **Bun 1.3.14**.

`archon` is `#!/usr/bin/env bun`, so every CLI invocation inherits it.

## Why it is worth fixing here rather than only in consumers

The failure is invisible at the boundary Archon owns. A consumer sees well-formed-looking JSON that stops early, or a parse error it will blame on itself. Anything polling a run list continuously — which is the intended use of `workflow status --json` — hits it proportionally to output size, so it gets *worse* as a workspace gets busier.

Consumers can guard (write to a temp file and read that back, which is what we will do), but every consumer has to know, and none of them can discover it from a zero exit.

## Options

- Flush explicitly / retry on `EAGAIN` before exit in the CLI's write path.
- Or pin/patch around bun#21516 until it lands upstream.
- At minimum, **document it** on the `--json` surfaces so consumers know a pipe is not safe.

## Found by

helm, planning a monitor rail that polls `archon workflow status --json`. Measured in Swift with `Process`/`Pipe`, 16 runs per route, same machine, same command.

Wirasm:
## Summary

- Problem: large `workflow status --json --verbose` and `workflow get --json --verbose` payloads could be silently truncated when stdout was piped, while the CLI exited successfully.
- Why it matters: pipe consumers received invalid machine-readable JSON without an error signal.
- What changed: preserve normal runtime shutdown and await the stdout callback for the large affected workflow JSON payloads; add lifecycle and backpressure regression coverage.
- What did **not** change (scope boundary): JSON schemas, command options, database behavior, and unrelated CLI output paths are unchanged.

### Before

```
User / script              Archon CLI                 Pipe consumer
─────────────              ──────────                 ─────────────
workflow status --json ─▶ serializes payload ──────▶ receives partial JSON
                           force-exits
exit 0 ◀──────────────────────────────────────────── parse fails
```

### After

```
User / script              Archon CLI                 Pipe consumer
─────────────              ──────────                 ─────────────
workflow status --json ─▶ serializes payload ──────▶ receives complete JSON
                           [awaits stdout write]
                           [allows normal shutdown]
exit 0 ◀──────────────────────────────────────────── parses successfully
```

### Before

```
workflow command ──▶ console/stdout write ──▶ CLI entry-point force exit ──▶ pipe
```

### After

```
workflow command [~] ──▶ awaited stdout callback [~] ──▶ CLI entry point [~] ──▶ pipe
                                                          normal shutdown
```

| `commands/workflow.ts` | `process.stdout` | **modified** | Awaits write completion for large status and verbose-get JSON output. |
| `cli.ts` | Node/Bun process lifecycle | **modified** | Sets exit code without forcing termination. |
| CLI tests | CLI/workflow sources | **modified** | Guards lifecycle and stdout backpressure behavior. |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `cli`, `tests`
- Module: `cli:workflow-output`

## Change Metadata

- Change type: `bug`
- Primary scope: `cli`

## Linked Issue

- Closes #2384
- Related: None
- Depends on: None
- Supersedes: None

Commands and result summary:

```bash
bun run type-check
bun run lint
bun run format:check
bun run test
bun run build
```

- All commands passed: 4,907 tests passed, with zero lint errors or warnings.
- Evidence provided: Swift `Process` + `Pipe()` regression captured parseable output in 16/16 runs for `workflow status --json --verbose` (363,722 bytes each) and 16/16 runs for `workflow get --json --verbose` (84,459–87,307 bytes); the not-found JSON envelope also parsed fully with exit status 1.
- If any command is intentionally skipped, explain why: None.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? Yes
- Config/env changes? No

- Verified scenarios: real non-blocking pipe captures for large verbose status and get JSON output; regular-file diagnostic capture.
- Edge cases checked: not-found JSON error envelope flushes completely and retains exit code 1.
- What was not verified: every other JSON command under an equivalent large-payload pipe, though the CLI lifecycle correction applies globally.

- Affected subsystems/workflows: CLI entry-point shutdown and workflow status/get JSON output.
- Potential unintended effects: process shutdown now waits for pending stdout completion; commands may remain alive briefly while a slow downstream pipe consumes output.
- Guardrails/monitoring for early detection: source-level lifecycle regression guard and stdout-backpressure command tests.

## Rollback Plan (required)

- Fast rollback command/path: revert commit `634f1ba7`.
- Observable failure symptoms: large workflow JSON piped to another process ends mid-document while the command exits 0.

## Risks and Mitigations

- Risk: a stalled pipe can delay completion while the stdout callback is pending.
  - Mitigation: awaiting is limited to the two documented large-payload paths implicated by the issue; regression tests cover callback backpressure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of machine-readable CLI output, including JSON sent through piped commands.
  * Prevented truncated or incomplete JSON responses during CLI execution.
  * Ensured command failures correctly propagate through pipelines.
  * Preserved existing human-readable output behavior.

* **Tests**
  * Added coverage for JSON delivery, formatting, errors, and non-blocking shell pipes.
  * Expanded CLI test coverage across workflow, validation, and listing commands.
  * Improved filesystem-related test compatibility across platforms.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "cli": "bun src/cli.ts",
-    "test": "bun test src/commands/version.test.ts src/commands/setup.test.ts src/commands/skill.test.ts src/commands/doctor.test.ts src/commands/telemetry.test.ts && bun test src/commands/workflow.test.ts && bun test src/commands/isolation.test.ts && bun test src/commands/chat.test.ts && bun test src/commands/serve.test.ts && bun test src/commands/ai.test.ts",
+    "test": "bun test src/commands/version.test.ts src/commands/setup.test.ts src/commands/skill.test.ts src/commands/doctor.test.ts src/commands/telemetry.test.ts && bun test src/commands/workflow.test.ts && bun test src/commands/isolation.test.ts && bun test src/commands/chat.test.ts && bun test src/commands/serve.test.ts && bun test src/commands/ai.test.ts && bun test src/cli.test.ts src/utils/stdout.test.ts",
     "type-check": "bun x tsc --noEmit"
   },
   "dependencies": {
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -7,6 +7,7 @@
 import { describe, it, expect } from 'bun:test';
 import { parseArgs } from 'util';
 import * as git from '@archon/git';
+import { tmpdir } from 'node:os';
 
 // Test the argument parsing logic used in cli.ts
 describe('CLI argument parsing', () => {
@@ -398,9 +399,11 @@ describe('CLI git repo check', () => {
     });
 
     it('should return null for system directories outside any git repo', async () => {
-      // /tmp is typically not inside a git repo
-      // Note: This test may need adjustment if /tmp happens to be inside a repo
-      const result = await git.findRepoRoot('/tmp');
+      // The OS temp dir is not inside a git repo on any supported platform.
+      // Hardcoding '/tmp' fails on Windows, where that path does not exist —
+      // this file was absent from the package test script until #2384, so the
+      // POSIX assumption never surfaced in CI.
+      const result = await git.findRepoRoot(tmpdir());
       expect(result).toBeNull();
     });
   });
@@ -412,7 +415,7 @@ describe('CLI git repo check', () => {
 
     it('should detect existing directories', () => {
       expect(existsSync(process.cwd())).toBe(true);
-      expect(existsSync('/tmp')).toBe(true);
+      expect(existsSync(tmpdir())).toBe(true);
     });
 
     it('should detect non-existent directories', () => {
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -250,17 +250,6 @@ describe('CLI argument parsing', () => {
   });
 });
 
-describe('CLI process lifecycle', () => {
-  it('does not force-exit after emitting output so piped JSON can flush', async () => {
-    const source = await Bun.file(new URL('./cli.ts', import.meta.url)).text();
-
-    expect(source).toContain('process.exitCode = exitCode;');
-    expect(source).toContain('process.exitCode = 1;');
-    expect(source).not.toContain('process.exit(exitCode);');
-    expect(source).not.toContain('process.exit(1);');
-  });
-});
-
 describe('Conversation ID generation', () => {
   // Test the generateConversationId pattern
   const generateConversationId = (): string => {
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -250,6 +250,17 @@ describe('CLI argument parsing', () => {
   });
 });
 
+describe('CLI process lifecycle', () => {
+  it('does not force-exit after emitting output so piped JSON can flush', async () => {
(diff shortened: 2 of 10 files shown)


Wirasm:

**Status**: COMPLETE
**Pushed**: ✅ No fix commit was needed; the PR branch is already current

---

### Fixes Applied (0 total)

The review reported no actionable findings.

---

### Tests Added

(none)

---

### Skipped (0)

(none — all findings addressed)

---

(none)

---

### Validation

✅ Type check | ✅ Lint | ✅ Tests

---

*Self-fix by Archon · aggressive mode · no changes required on `archon/task-archon-fix-github-issue-experimental-1785738722065`*

Wirasm:
Pushed `a4ed3230` — the Windows failures were test bugs, not a CLI defect.

Two assertions hardcoded `/tmp`:

```js
const result = await git.findRepoRoot('/tmp');
expect(existsSync('/tmp')).toBe(true);
```

Neither holds on Windows, where that path does not exist. Both now use `os.tmpdir()`, which carries the same intent everywhere: a directory that exists and is not inside a git repo.

Worth noting *why* this surfaced now — `packages/cli/src/cli.test.ts` was not in the package's `test` script, so it had never run in CI (#2401). Wiring it in is what made a long-standing POSIX assumption visible. The file failing on its first real execution is the guard working, not a regression.

Verified: 50 pass / 0 fail locally, type-check and format clean.

Wirasm:
On the `beforeAll` finding — not acting on it, reasoning recorded here.

The guideline's concern ("clean up test data after each test") is already satisfied: `beforeAll` creates a unique `mkdtempSync(join(tmpdir(), 'archon-pipe-test-'))` and `afterAll` removes it with `rmSync(scratch, { recursive: true, force: true })`. Nothing leaks.

The isolation risk worth checking would be tests sharing an output filename, where a truncated-to-zero run could silently read a previous test's good file — which for a truncation test would be the worst possible failure mode. They don't share:

- `reference.json` — written once
- `piped-${run}.json` — unique per iteration
- `sink.json` — separate

Every write target is distinct, so a stale file cannot mask a failure.

Moving to `beforeEach` would make it worse: fixture setup creates a git repo and workflow files, and the piped test runs 10 iterations spawning real shell pipelines. Rebuilding per test costs time and buys no safety.

Issue #2410, opened 2026-08-03
feat(workflows): command: nodes cannot receive upstream context — 69 nodes re-derive what the DAG already computed


Wirasm:
A `command:` node has **no way to receive anything a previous node produced**. It renders a shared command file whose only input is `$ARGUMENTS` — the run's original trigger message. Everything a workflow computes upstream is invisible to it unless the command re-derives it from scratch.

This is not a papercut. Workflows are built on the premise that context flows along `depends_on` edges, and for 69 of the nodes in the bundled set it does not.

## Verified

```ts
// packages/workflows/src/schemas/dag-node.ts:254
export const commandNodeSchema = dagNodeBaseSchema.extend({
  command: z.string(),
});
```

That is the whole surface. Contrast with the `workflow:` sub-run node in the same file:

```ts
// :483
input: z.string().optional(),
```

So the engine **already has the concept** — a node that takes a data string, substituted like a prompt body. It exists for sub-runs and not for commands.

Scale: **69 `command:` node instances across 13 bundled workflows.**

## What it cost, concretely

On 2026-08-03, `archon-fix-github-issue-experimental` ran against #2359. The DAG is:

```
fetch-issue (bash, gh issue view --json …,comments,…)
   └─> classify ──> investigate / plan   (command: nodes)
```

`fetch-issue` fetched the issue **including a maintainer comment that specified the exact output shape**. I confirmed the contract text was present in that node's persisted output. `classify` and `smoke-validate` both reference `$fetch-issue.output` and saw it.

`investigate` and `plan` — the nodes that decide what gets built — are `command:` nodes. They received **none of it**. They re-fetched the issue themselves from `$ARGUMENTS` and planned from the body. The resulting PR (#2399) implemented the wrong design and was closed.

The immediate fix (#2404) was to tell the *command file* to read comments and treat them as authoritative. That works, but it is a workaround for a missing wire: it makes every command re-derive context instead of receiving it, and it only helps for the one kind of context anybody thought to write down.

## Why the workaround does not generalise

A command file is **shared across workflows**. So workflow-specific context cannot be referenced from it — a `$fetch-issue.output` ref inside `archon-create-plan.md` would break every other workflow that has no `fetch-issue` node. The command must therefore either:

- re-derive context from `$ARGUMENTS` (duplicating upstream work, and getting it wrong when the upstream node did something the command cannot reproduce), or
- read a hard-coded artifact path and hope the workflow wrote it.

[...]

Issue #456, opened 2025-08-23
Enhancement: Speed-up task reordering for large projects


ericfisherdev:
# Enhancement: Speed-up task reordering for large projects

## Issue Summary
Task re-ordering will create a bottleneck if you have a large amount of tasks in 'todo'. Adding tasks using the REST API is taking > 18 seconds if a large amount of tasks exist. Currently tested on a project with more that 500 tasks with 'todo' status.

## Problem Description
- **Symptom**: I was writing a script to pull Github PR review comments from CodeRabbit and create Archon tasks. It started fine, but as I was doing some testing by creating a large number of PRs and PR reviews, the performance became increasingly worse and the script kept timing out.
- **Root Cause**: Individual Archon API task creation calls taking ~18.5 seconds each
- **Impact**: Extraction operations with >6 tasks consistently fail due to timeout, as the script was supposed to complete in under 2 minutes, even with 100+ review comments.

## Test Case
```bash
# Direct API test - took 18.548 seconds
curl -X POST http://localhost:8181/api/tasks \
  -H "Content-Type: application/json" \
  -d '{
    "project_id": "[your-project-id]",
    "title": "Test API Performance", 
    "description": "Testing direct API call performance",
    "assignee": "AI IDE Agent",
    "task_order": 1,
    "feature": "performance-test"
  }'
```

## Key Finding: Primary Bottleneck - Inefficient Task Reordering Logic

**Location**: `TaskService.create_task()` method, lines 114-135  
**Issue**: N+1 database operation pattern

```python
# PROBLEMATIC CODE:
if task_order > 0:
    # Single SELECT query
    existing_tasks_response = (
        self.supabase_client.table("archon_tasks")
        .select("id, task_order")
        .eq("project_id", project_id)
        .eq("status", task_status)
        .gte("task_order", task_order)
        .execute()
    )
    
    # N individual UPDATE queries (performance killer)
    for existing_task in existing_tasks_response.data:
        self.supabase_client.table("archon_tasks").update({
            "task_order": existing_task["task_order"] + 1,
            "updated_at": datetime.now().isoformat(),
        }).eq("id", existing_task["id"]).execute()
```

## Recommended Solution:

** Use a bulk update function to bump the task order **
```sql
CREATE OR REPLACE FUNCTION increment_task_order(
    p_project_id UUID,
    p_status task_status,
    p_min_order INTEGER,
    p_updated_at TEXT
)
RETURNS VOID AS $$
DECLARE
[...]

PR #2411, opened 2026-08-03
fix(commands,workflows): follow-ups from the #2404 review

Wirasm:
Four findings from CodeRabbit's review of #2404, all verified against the files before acting. Two were Major and one of those meant half the planning path never received #2404's fix at all.

## 1. `archon-create-plan` never classified a GitHub issue (Major)

Its Phase 0.1 input table had five rows — PRD, PRD-with-phases, file path, free-form, empty — and **none matched a number or URL**. So `1234` fell through to "Free-form text → use directly as feature input", and the comment-authority section #2404 added was **unreachable on the entire non-bug planning path**. I added a section gated on a classification the command could not produce.

Added the two missing rows, and made the match order explicit (first match wins).

## 2. Comments are now weighed, not obeyed

The first version made a maintainer comment mechanically authoritative. That is too much: comments can be stale, contradicted by code that has since changed, or simply wrong — and the planner is the one actually reading the code.

Reworked so that reading every comment stays mandatory, `authorAssociation` **weights** rather than decides (write-access decisions are the default course; `CONTRIBUTOR`/`NONE` is input rather than instruction, since anyone can comment on a public repo), and the planner may disagree. What is forbidden is ignoring a decision **silently** — follow it, or say why not. That is the actual failure this guards against.

## 3. Cross-repo linked issues (Major)

A bare `#1234` means the current repo; a full URL may not. Extracting the number from `https://github.com/other/repo/issues/456` and running `gh issue view 456` reads the wrong repo's issue. Pass the URL verbatim.

## 4. `prep-worktree` was advisory (Major)

It only **asked** the model to "report whether the install succeeded". A non-zero install can be narrated as a failure while the node still records success, and `implement` then runs against a broken dependency tree — where the resulting build/type-check/test failures get misattributed to the implementation instead of the environment.

This is the same class of bug #2404 existed to fix, in a third node in the same file that I missed. Added `assert-deps-installed`, and rewired `implement` to depend on it instead of `prep-worktree` directly (a sibling node would not have gated anything). It checks a marker the prompt writes **and** independently verifies `node_modules` exists when `package.json` does — so a model that narrates success without writing the marker is still caught.

## 5. Stale rationale on `assert-implemented` (Minor)

The comment claimed only the working tree is inspected; the guard also counts commits ahead of the base. That branch exists because three consecutive runs committed their work, leaving a clean tree and failing runs that had genuinely done the job. Now documented so it is not deleted as dead weight.

## Validation

```
bun run check:bundled                                    # up to date (36 commands, 21 workflows)
bun run cli validate workflows archon-fix-github-issue-experimental   # 1 valid, 0 errors
bun run cli validate commands archon-create-plan          # ok
bun run cli validate commands archon-investigate-issue    # ok
prettier --check                                          # clean
```

The one remaining validator warning is pre-existing on `fetch-issue`.

A run against a real issue follows; if the planning path still misses context, that is evidence the prompt-level approach is insufficient and #2410 (command nodes cannot receive upstream context) is the real fix.

## Summary by CodeRabbit

* **New Features**
  * GitHub issue numbers and issue or pull request URLs are now recognized automatically for planning and investigation.
  * Issue workflows now review all comments, consider author context, document conflicting decisions, and follow linked issues across repositories.
* **Bug Fixes**
  * Implementation is now blocked when dependencies fail to install or required dependencies are missing.
  * Worktree and commit changes are checked more reliably before implementation proceeds.


Wirasm:
Disposition of the three findings, since two threads remain open:

**1. Yarn PnP / dependency validation — fixed in \`e0dbb9bc\`, partially.**

The blocking half was real and would have shipped a regression: my \`[ -f package.json ] && [ ! -d node_modules ]\` check exits 1 on every run in a Yarn PnP repo, which resolves from \`.pnp.cjs\` and has no \`node_modules\` by design. In a bundled default that blocks a legitimate project outright. Now gated on PnP markers with the reasoning recorded inline.

The stronger suggestion — have \`prep-worktree\` emit a checked install status for the selected manager and validate *that* — is better than what I shipped and I did not do it. What is here is a marker file plus one ecosystem-specific check. That is a real improvement over a purely advisory report but it is not a general solution, and I would rather say so than imply the finding is fully closed.

**2. Cross-repo issue identity — filed as #2412, not fixed here.**

Correct, and it is the workflow's own entry point rather than the linked-issue handling this PR touches: \`extract-issue-number\` is instructed to emit only a bare number, so \`gh issue view "$ISSUE_NUM"\` resolves against the current repo. An operator starting a run with \`https://github.com/other/repo/issues/456\` silently gets this repo's #456. Pre-existing, so fixing it here would widen the PR past its subject.

**3. PR URLs to \`gh issue view\` — fixed in \`e0dbb9bc\`.** The table row claimed to route PR URLs to the issue path, but 0.1a omits PR \`reviews\`. Scoped the row to issue URLs rather than adding PR handling to a planning command.

Issue #2327, opened 2026-07-30
fix(defaults): bundled SDLC workflows must encode blocked outcomes structurally


Wirasm:
## Problem

Several bundled SDLC runs completed green after their agents honestly reported that they were blocked and produced no deliverable. Downstream validation, PR creation, review, and reporting nodes then continued against missing work.

The provider and engine behaved as currently specified: a successful model turn is a completed node. The failure is that bundled workflows asked for prose and then expected the engine to understand whether that prose meant “done” or “blocked.”

Natural language is not a wire format. Archon must not infer lifecycle intent from phrases such as “blocked,” “could not,” or “no changes.”

## Required outcome

The bundled SDLC workflow pack should declare and route structured outcomes at the nodes that can legitimately stop without producing their promised deliverable.

A suitable node result distinguishes at least:

- work produced,
- blocked with an actionable reason,
- no change required where that is a legitimate successful outcome.

The exact schema may vary by workflow, but downstream control flow must use typed output fields, not prose matching.

## Workflow behavior

- Implementation nodes emit a structured status and human-readable explanation.
- A blocked implementation does not run validation, PR creation, or review as if code exists.
- “No change required” remains distinct from “blocked.”
- Deterministic prerequisites or artifact existence checks use bash/script guard nodes where appropriate.
- The final report preserves the structured reason and names the first blocking stage.
- Interactive recovery, when supported, uses existing gate/decision primitives rather than a new blocked node status.

## Scope

This is bundled/default workflow authoring work, aligned with #2123 and the packaged SDLC work in #2627. It is not a new workflow-engine lifecycle state and does not inspect model prose.

Implicit engine inference from declared artifacts, keywords, regular expressions, or model text is out of scope.

## Regression evidence

The original incidents included implementation, validation, PR, review, and report nodes all returning honest blocked prose while the overall run completed. Tests must prove that the same structured blocked result prevents the downstream delivery path.

## Acceptance

- [ ] Relevant bundled SDLC agent nodes declare structured outcome schemas.
- [ ] Downstream nodes branch on those fields.
- [ ] Blocked, no-change, and delivered outcomes remain distinct.
- [ ] A blocked implementation cannot reach PR creation or code review.
- [ ] Tests fail when the structured guard is removed or inverted.
- [ ] Bundled workflow documentation explains the outcome contract.

Wirasm:
Promoting this — evidence from 2026-08-03 shows it is the root of two other issues, not a peer.

[...]

Issue #123, opened 2025-04-21
[BUG] Bug Summary: Port 8100 Already in Use


Nathamuni:
[10:42:16] Agent service started
INFO:     Started server process [1708]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
ERROR:    [Errno 98] error while attempting to bind on address ('0.0.0.0', 8100): address already in use
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.

Balaji2810:
---

## Windows

1. **Find the PID**

   ```bat
   netstat -aon | findstr :8100
   ```

   Look in the right-most column for the PID.

2. **Kill the process**

   ```bat
   taskkill /PID 1234 /F
   ```

   (Replace `1234` with the PID you found.)

3. **Restart**

   ```bat
   python run_docker.py
   ```


---

## macOS

1. **Find the PID**

   ```bash
   lsof -nP -iTCP:8100 | grep LISTEN
   ```

   The PID is in the second column.

2. **Kill the process**

   ```bash
   kill <PID>
   ```

   If it doesn’t go away:

   ```bash
   kill -9 <PID>
   ```

3. **Restart**

   ```bash
   python run_docker.py
   ```

---

## Linux

1. **Find the PID**

   ```bash
   sudo lsof -nP -iTCP:8100 | grep LISTEN
   ```

   or

   ```bash
   sudo netstat -tulpn | grep :8100
   ```

2. **Kill the process**

   ```bash
   sudo kill <PID>
   ```

   If needed:

   ```bash
   sudo kill -9 <PID>
   ```

3. **Restart **

   ```bash
   python run_docker.py
   ```


---

That’s it! Use the OS-specific commands above to clear port 8100 (or switch ports) and get your server running again.

coleam00:
Thanks for submitting this issue! However I'm closing this now because the overhaul for Archon is coming soon so we need a blank slate for the repository. Some functionality from the old Archon will still apply to the new, so if you encounter this issue or something related with the new version of Archon, please feel free to reopen!

PR #2417, opened 2026-08-03
fix(workflows,commands): preserve repo identity for issue URLs; unblock .archon fixes

### LINKED ISSUE #2412: fix(workflows): extract-issue-number discards repo identity, so a cross-repo issue URL reads the wrong repo
Surfaced by CodeRabbit while reviewing #2411. Pre-existing, not introduced there, so filing rather than widening that PR.

## Verified

`archon-fix-github-issue-experimental`'s `extract-issue-number` node instructs the model:

> CRITICAL: Your final output must be ONLY the bare number with no quotes, no markdown, no explanation.

`fetch-issue` then runs:

```bash
ISSUE_NUM=$(echo "$extract-issue-number.output" | tr -d "'\"\`\n " | grep -oE '[0-9]+' | head -1)
gh issue view "$ISSUE_NUM" --json title,body,labels,comments,state,url,author
```

`gh issue view <number>` resolves against the **current** repository. So an operator who starts a run with a full cross-repo URL — `https://github.com/other/repo/issues/456` — gets this repo's #456 instead, or a failure if no such issue exists. Silently wrong in the first case.

## Why it has not bitten yet

Archon is normally run against its own repo with a bare number, so the number and the repo agree. It becomes reachable as soon as anyone dispatches a run for another repository's issue, which is a reasonable thing to want.

## Fix sketch

Have `extract-issue-number` emit the owner/repo when the input carries one, and have `fetch-issue` pass it through — either the URL verbatim (`gh issue view <url>` preserves identity) or an explicit `--repo owner/repo`. Structured output would suit this better than a bare string.

Related: #2411 fixed the same class of mistake **inside** the planning commands (linked issues referenced by URL). This is the workflow's own entry point.

Wirasm:
Closes #2412. Hand-written rather than generated, for a reason worth recording.

## Why a run could not do this

An Archon run on #2412 reached `implement` and produced nothing. Not a failure to try — it stopped and said:

> **Blocked by contradictory instructions:**
> - Investigation requires modifying and committing `.archon/workflows/experimental/archon-fix-github-issue-experimental.yaml`
> - The command explicitly says files under `.archon/` must never be modified or staged
>
> Please confirm that this specific `.archon/` file is an exception.

That rule came from #2358, the first PR of this session. It exists to stop runs sweeping the operator's copied-in `.archon/` edits into a target-repo PR — a real problem it solved. But it was absolute, so **any issue whose fix is a workflow, command or script had no path forward**.

The implementer behaved exactly right: detected the contradiction, named both sides, refused to guess. Worth noting the engine then recorded `node_completed` for a node that had explicitly reported itself blocked (#2327), and only `assert-implemented` caught it.

## 1. Repo identity (#2412)

`extract-issue-number` was told to emit "ONLY the bare number". `fetch-issue` then ran:

```bash
gh issue view "$ISSUE_NUM" --json …
```

which resolves against the **current** checkout. A run started with `https://github.com/other/repo/issues/456` silently investigated *this* repo's #456 — plausible output, wrong issue.

Now structured:

```yaml
output_format:
  type: object
  properties: { issue_number: { type: string }, repo: { type: string } }
  required: [issue_number, repo]
```

`repo` is empty unless the request actually named one — an empty value means "current checkout", the common case, and the prompt says never to guess. `fetch-issue` passes `--repo` only when present.

**Fixed in the bundled `archon-fix-github-issue` too**, not just the experimental copy — the bundled one ships this bug to every user. The rewrite also clears the pre-existing double-quoted-substitution warning on that node (#1884).

## 2. The `.archon/` rule

Now distinguishes by **intent, not path**: a file your plan names is your work; every other dirty `.archon/` file is the operator's and stays untouched. The original guard against sweeping up unrelated edits is intact — the staged-file check remains — it just no longer makes a directory unfixable.

## Validation

```
bun run cli validate workflows archon-fix-github-issue-experimental   # 1 valid, 0 errors, 0 warnings
bun run cli validate workflows archon-fix-github-issue                # 1 valid, 0 errors, 0 warnings
bun run cli validate commands  archon-fix-issue                       # ok
bun run check:bundled                                                 # up to date (36 commands, 21 workflows)
prettier --check                                                      # clean
```

Both workflows now validate with **zero warnings**, down from one each.

## Not verified

That a cross-repo URL end-to-end now reads the right issue — it needs a run started with another repository's URL, which I have not done. The unit-level change is straightforward but the claim is untested.

## Summary by CodeRabbit

- **New Features**
  - GitHub issue workflows now support explicitly referenced issues from other repositories.
  - Issue links and repository references are identified automatically, while issues without a repository continue using the current checkout.

- **Bug Fixes**
  - Improved issue validation prevents invalid or conflicting references and avoids fetching issues from the wrong repository.
  - Updated workflow guidance allows planned configuration changes while protecting unrelated modifications and stopping when changes cannot be safely separated.


Wirasm:
Both findings were correct — holes in the fix itself, not in the issue. Pushed `7f6c705f`.

## 1. A named `.archon/` file is not a blank cheque for that file

The rule said "a file your plan names is your work", and the safety check is `git diff --cached --name-only`. Both operate on **paths**. Neither can see that a planned file already carries copied-in edits from before the run — staging it whole commits those too.

Now the agent records a baseline before editing:

```bash
git diff -- <the-planned-file> > /tmp/archon-baseline.diff   # empty if clean
```

…then stages only its own hunks with `git add -p`, rejecting any hunk that also appears in the baseline. If they are entangled beyond separating, it stops and says so rather than committing someone else's work under its change — which is the same call the 2026-08-03 run made, and the right one.

## 2. An issue reference is an atomic pair

The rules described finding a number and finding a repo as separate steps, so a request carrying both `#123` and `https://github.com/other/repo/issues/456` could take the number from one and the repo from the other, fetching an issue nobody asked for. Worse than the bug this PR set out to fix, since at least that one was consistently wrong.

Each recognised form now yields one complete pair:

```
#709 / issue 709 / 709                        -> {issue_number: "709", repo: ""}
owner/repo#709                                -> {issue_number: "709", repo: "owner/repo"}
https://github.com/owner/repo/issues/709      -> {issue_number: "709", repo: "owner/repo"}
```

Several references resolving to the same pair is fine. Several **different** pairs is an explicit failure naming all of them and asking which was meant — guessing there silently investigates the wrong issue.

Applied to both the bundled and experimental workflows. Both still validate with zero warnings; `check:bundled` up to date; prettier clean.

Wirasm:
Correct, and a real leak in the fix I added. Pushed.

`git diff -- <file>` compares the worktree against the **index**. Pre-existing changes that were already *staged* therefore never appeared in the baseline — and `git add -p` neither shows nor removes staged hunks, so they would have entered the commit invisibly. That is precisely the leak the baseline exists to prevent, so the fix was hollow against the staged case.

Two changes:

```bash
git diff HEAD -- <the-planned-file> > /tmp/archon-baseline.diff   # staged AND unstaged
```

and, before staging anything of my own:

```bash
git restore --staged <the-planned-file>   # no-op if nothing was staged
git add -p <the-planned-file>             # only deliberately picked hunks
```

Clearing the index first makes the guarantee structural rather than dependent on the agent noticing something in a diff — the only thing that *can* be staged afterwards is what was explicitly chosen.

`check:bundled` up to date, command validates, prettier clean.

PR #2414, opened 2026-08-03
feat: summarize verbose workflow JSON nodes

Wirasm:
## Summary

- Problem: verbose workflow JSON exposed raw lifecycle rows, so consumers had to recreate the CLI's node-state fold and could observe unstable ordering for tied timestamps.
- Why it matters: machine consumers need a compact, stable progress view without tool-event noise or duplicate interpretation logic.
- What changed: verbose JSON now returns ordered node summaries by default, includes `startedAt`, supports raw events through `--events`, and uses a deterministic event-order tie-breaker.
- What did **not** change (scope boundary): non-verbose JSON, database schema, workflow event types, and direct database consumer contracts are unchanged.

### Before

```
CLI consumer          Archon CLI                 Workflow store
────────────          ──────────                 ──────────────
status/get --verbose ───────▶ fetch events ───────────────▶ raw lifecycle rows
receives raw events ◀──────── emit JSON ◀────────────────── rows ordered by timestamp
folds nodes locally
```

### After

```
CLI consumer          Archon CLI                 Workflow store
────────────          ──────────                 ──────────────
status/get --verbose ───────▶ fetch events ───────────────▶ rows ordered by timestamp + id
receives [nodes[]] ◀──────── [shared fold] ◀────────────── ordered lifecycle rows
--events ───────────────────▶ emit raw diagnostic events instead
```

### Before

```
packages/cli/src/cli.ts → packages/cli/src/commands/workflow.ts → packages/core/src/db/workflow-events.ts
                                              │
                                              └→ raw verbose JSON events
packages/docs-web/.../cli.md → CLI reference
```

### After

```
packages/cli/src/cli.ts [~ --events parsing] → packages/cli/src/commands/workflow.ts [~ shared nodes fold]
                                                             │
                                                             └=== packages/core/src/db/workflow-events.ts [~ created_at,id order]
packages/docs-web/.../cli.md [~ JSON contract and diagnostics]
```

**Connection inventory** (list every module-to-module edge, mark changes):

| From | To | Status | Notes |
|---|---|---|---|
| CLI argument parser | Workflow status/get commands | **modified** | Threads `--events` to verbose JSON selection. |
| Workflow status/get commands | Workflow event database | **modified** | Consumes deterministically ordered lifecycle rows. |
| Workflow commands | Node-summary fold | **modified** | JSON now shares the text-output fold by default. |
| CLI reference docs | CLI commands | **modified** | Documents nodes and raw-event opt-in behavior. |

## Label Snapshot

[...]

PR #2418, opened 2026-08-03
fix(db): SQLite upgrade broke on event_order — index/trigger ran before the ALTER

Wirasm:
**P0 — every existing SQLite install is broken on `dev` right now, and SQLite is the default.** Found within minutes of merging #2414 by running `archon workflow runs` against my own database.

## The failure

```
SQLiteError: no such column: event_order
    at createSchema (packages/core/src/db/adapters/sqlite.ts:516:13)
    at initSchema  (packages/core/src/db/adapters/sqlite.ts:177:10)
```

#2414 added `idx_workflow_events_run_order` and the `assign_order` trigger to `createSchema()`. Both reference `event_order`, which does not exist on any database created before that column. **CREATE INDEX on a missing column aborts the entire `createSchema()` exec block**, so `createSchema()` throws — and `migrateColumns()`, the code that would have added the column, is on the next line and never runs.

Self-perpetuating: the migration that fixes it can never execute.

## The codebase already documented this exact hazard

From the `user_id` index added in 0.4.0, a few hundred lines above:

> `// Index must be created here, not in createSchema(): the column doesn't`
> `// exist on pre-0.4.0 databases until the ALTER TABLE above runs, and`
> `// CREATE INDEX on a missing column aborts the entire createSchema()`
> `// exec block. Idempotent so it's safe to run unconditionally.`

Same trap, same file, previously learned.

## The fix

Both statements move out of `createSchema()`. `migrateColumns()` already creates them idempotently **after** its `ALTER TABLE`, in the correct order — so this is a deletion plus a comment, not new logic. Fresh databases still get the column from `CREATE TABLE` and the index/trigger from `migrateColumns()`.

## Verification

Against a real pre-existing `~/.archon/archon.db` that was broken by the merge:

```
before:  event_order present? 0        archon workflow runs -> ERROR
after:   event_order present? 1        archon workflow runs -> works
         index idx_workflow_events_run_order      created
         trigger remote_agent_workflow_events_assign_order  created
```

Type-check clean, `sqlite.test.ts` 20 pass / 0 fail.

## Why the tests did not catch it

The SQLite suite creates fresh databases, where `CREATE TABLE` includes `event_order` and the index therefore succeeds. The broken path is **upgrade only**: an existing database where `CREATE TABLE IF NOT EXISTS` is a no-op. There is no test that opens a pre-`event_order` database and asserts the schema converges.

That gap is worth closing separately — it would have caught this, and it is the same class as the reverse-drift and parity checks already in that suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database setup and migrations for existing databases.
  * Prevented schema initialization failures when workflow event ordering fields are added during migration.
  * Ensured missing workflow event ordering fields and related database structures are restored during upgrades.

### DIFF
--- a/packages/core/src/db/adapters/sqlite.test.ts
+++ b/packages/core/src/db/adapters/sqlite.test.ts
@@ -32,13 +32,16 @@ async function insertCodebase(db: SqliteAdapter, id: string): Promise<void> {
  * the fixture realistic — the upgrade path that broke was an otherwise-current
  * database missing exactly this one column.
  */
-function makeDbWithoutEventOrder(): string {
+async function makeDbWithoutEventOrder(): Promise<string> {
   const path = join(
     import.meta.dir,
     `.test-sqlite-legacy-${Date.now()}-${Math.random().toString(36).slice(2)}.db`
   );
   const seed = new SqliteAdapter(path); // writes the current schema
-  void seed.close();
+  // MUST await: close() is async, and on Windows an unreleased SQLite handle
+  // locks the file, so the Database opened below fails. Harmless on POSIX,
+  // which is why the first version of this test passed locally and failed CI.
+  await seed.close();
   const raw = new Database(path);
   try {
     raw.run('DROP TRIGGER IF EXISTS remote_agent_workflow_events_assign_order');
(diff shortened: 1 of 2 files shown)


Wirasm:
Fixed in the pushed commit. Minor by severity, but worth taking on this PR specifically: an unreleased SQLite handle prevents `unlink` on **Windows**, and my `afterEach` swallowed every unlink error — so a leaked handle would silently leave fixture files behind on exactly the platform whose CI job is the least reliable. The finding lands on the one test that opens raw database handles repeatedly.

- adapters closed after use (`makeDbWithoutEventOrder` seed, and the upgraded instance)
- every raw `Database` handle wrapped in `try/finally`
- cleanup rethrows anything that is not `ENOENT`

Re-verified the guard still has teeth after the refactor — reintroducing the index into `createSchema()` fails the test, removing it passes. 21 pass / 0 fail, type-check clean.

PR #2420, opened 2026-08-03
feat(commands,workflows): archon-parse-user-request replaces extract-issue-number

Wirasm:
One reusable command replaces a bespoke extraction prompt in two workflows, and adds the field that was missing.

## The contract

| field | guarantee |
| --- | --- |
| `user_request` | **required** — the input verbatim, never rewritten, never empty for a non-empty message |
| `issue_number` | best-effort, `""` when absent |
| `repo_url` | best-effort, `""` when absent, **copied verbatim or empty — never constructed** |

`"123"` yields `user_request: "123"`. That is the case that breaks the naive design: if `user_request` were "the natural-language part beyond the reference", the single most common input would produce an empty field and fail every run.

## Why the asymmetry matters

**The failure check becomes meaningful.** An empty `user_request` on a non-empty input is not a bad request — it is the parser failing to return what it was handed. That is a countable defect, and it gives a model-tier bump (`small` → `medium`) something to justify it. Nothing else in the schema can trigger it.

**`issue_number: ""` is a correct answer.** A stack trace, an uploaded file, or a plain instruction names no issue. `extract-issue-number` presupposed one existed, so `fetch-issue` exited 1 on any non-issue input — the run died at node two while the prompt advertised ambiguity handling it did not have. The shape now makes that path expressible, though this workflow still requires an issue and says so in its own error.

**`repo_url` cannot be mis-paired.** `fetch-issue` passes the whole URL to `gh issue view` rather than reassembling owner/repo from parts:

```bash
if [ -n "$url" ]; then
  gh issue view "$url" --json …
else
  gh issue view "$num" --json …
fi
```

`gh issue view <n>` resolves against the current checkout, which is exactly how a cross-repo reference silently fetched the wrong issue (#2412). Carrying the URL whole means number and repository can never be separated — the remaining risk narrows from "mis-paired two real values" to "invented a URL", and the command explicitly forbids constructing one.

## No prose either side

The command states that the reply contains only the declared fields. On `claude`/`codex`/`opencode` `structuredOutput` is `enforced` and the SDK constrains output anyway; on `pi`/`copilot` it is `best-effort`, so that instruction is the actual mechanism rather than belt-and-braces. Worth knowing it is load-bearing on exactly the providers where the schema is not.

## Scope

Applied to the **bundled** `archon-fix-github-issue` as well as the experimental copy — the bundled one carried the same shape and ships to every user.

## Validation

```
bun run check:bundled                                   # up to date (37 commands, 21 workflows)
bun run cli validate commands archon-parse-user-request # ok
bun run cli validate workflows archon-fix-github-issue                # 1 valid, 0 errors
bun run cli validate workflows archon-fix-github-issue-experimental   # 1 valid, 0 errors
prettier --check                                        # clean
```

## Not verified

No run has exercised this end to end. In particular the cross-repo path — a run started with another repository's issue URL — remains unverified, same as it was for #2417. The unit-level shape is straightforward; the claim is untested.

A generic request-parser fitting the wider default set is deliberately out of scope.

## Summary by CodeRabbit

* **New Features**
  * Added improved parsing of requests, including issue numbers, repository references, and repository URLs.
  * GitHub issue workflows can now fetch issues directly from supplied repository URLs or specified repositories.
  * Preserved the complete original request for more accurate issue handling.
* **Bug Fixes**
  * Added validation for empty requests and invalid issue numbers.
  * Improved handling of incomplete repository details and ambiguous issue references.


Wirasm:
Two taken, one declined. Pushed `0ff9ee4b`.

## Taken: `owner/repo#N` was losing its repository

This was a **regression I introduced**, and the review is right to call it. #2417 handled the shorthand — `owner/repo#709` → `{issue_number: "709", repo: "owner/repo"}` → `gh issue view 709 --repo owner/repo`. My command dropped it, and my own examples table said so plainly:

| `owner/repo#88` | `owner/repo#88` | `88` | `""` |

So `gh issue view 88` resolved against the current checkout — precisely the bug #2412 exists to fix, reintroduced for the shorthand form while fixed for the URL form. The "never construct a URL" rule was right; I used it to justify discarding a repository the input stated outright.

Now a separate verbatim `repo` field. Nothing is constructed or derived: a URL goes in `repo_url`, a shorthand goes in `repo`, neither is built from the other.

## Taken: the numeric gate ran too early

Confirmed — the `^[0-9]+$` check fired **before** the URL branch, so a best-effort miss on `issue_number` aborted a run whose `repo_url` alone was a complete, valid argument to `gh issue view`. `issue_number` is documented best-effort; that ordering turned an expected miss into a hard failure. Moved into the `else`, where it gates only the number path.

## Declined: `additionalProperties: false`

The engine already does this. `packages/providers/src/shared/structured-output.ts:154`:

> *Recursively inject `additionalProperties: false` on every object schema…*

with the write at `:201`, and it respects a pre-existing value including an open-record subschema. That is why **no bundled workflow declares it** — the engine owns the strict-mode normalisation so authors do not have to. Adding it to these two YAMLs would duplicate engine behaviour in every author's file and invite drift the moment the engine's rules change.

Worth noting for completeness: Pi's Anthropic path forwards only `properties` and `required` into `input_schema`, dropping `additionalProperties` entirely — so on that backend the field is a no-op regardless of who writes it.

Validation: `check:bundled` up to date (37 commands, 21 workflows), both workflows and the command validate with zero warnings, prettier clean.

PR #2443, opened 2026-08-04
fix(migrations): create indexes after ADD COLUMN for existing Postgres DBs

### LINKED ISSUE #2442: 000_combined.sql: index created before ADD COLUMN breaks existing Postgres DBs
## Summary

`migrations/000_combined.sql` fails on **existing** Postgres databases because two indexes are created in the `CREATE TABLE` section, before the columns they reference are added in the additive `ALTER TABLE ... ADD COLUMN` block at the end of the file.

## Mechanic

- On a **fresh** DB, `CREATE TABLE` (lines 109 / 140) defines all columns inline, so the indexes work.
- On an **existing** DB, `CREATE TABLE IF NOT EXISTS` is a no-op, so the newer columns do not exist yet. They only get added later by the additive `ADD COLUMN IF NOT EXISTS` block.
- The two `CREATE INDEX` statements run *before* that block and fail with `column ... does not exist`.

`CREATE INDEX IF NOT EXISTS` does not protect against this: when the index is missing, Postgres tries to build it and errors on the missing column.

## Affected pairs (dev branch)

| Index | line | Column added at |
|---|---|---|
| `idx_conversations_hidden` on `remote_agent_conversations(hidden)` | 128 | `hidden` at 382 (`ADD COLUMN IF NOT EXISTS hidden`) |
| `idx_sessions_parent` on `remote_agent_sessions(parent_session_id)` | 159 | `parent_session_id` at 367 |

This violates the additive-only rule (#2317). The #2418 fix applied the same reordering only in the SQLite path; the Postgres path was left open.

## Fix

Move both `CREATE INDEX` statements down, directly after their respective `ADD COLUMN` in the additive block. Fresh DBs are unaffected; existing DBs migrate cleanly. PR follows.

miiwins:
## Summary

Fixes #2442.

`migrations/000_combined.sql` fails on **existing** Postgres DBs: two indexes are created in the `CREATE TABLE` section, before the columns they reference are added in the additive `ALTER TABLE ... ADD COLUMN` block. On an existing DB, `CREATE TABLE IF NOT EXISTS` is a no-op, so those columns do not exist yet when the indexes run, and creation fails with `column ... does not exist`.

This PR moves both indexes to directly follow their `ADD COLUMN`:

- `idx_sessions_parent` -> after `ADD COLUMN ... parent_session_id`
- `idx_conversations_hidden` -> after `ADD COLUMN ... hidden`

Fresh DBs are unaffected (columns already exist inline in `CREATE TABLE`). This complements #2418, which applied the same reordering only in the SQLite path; the Postgres path was still broken.

## Changes

- `migrations/000_combined.sql`: relocate the two `CREATE INDEX IF NOT EXISTS` statements into the additive block, each directly after its `ADD COLUMN`, with a short comment explaining why the order matters.

## Test plan

- [ ] Fresh Postgres DB: run migration end to end, both indexes present.
- [ ] Pre-existing Postgres DB (tables already created without the new columns): run migration, columns are added and both indexes build without `column does not exist`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database upgrades for existing installations by ensuring required indexes are created correctly.
  * Optimized conversation lookups to focus on non-deleted records.

### DIFF
--- a/migrations/000_combined.sql
+++ b/migrations/000_combined.sql
@@ -125,8 +125,6 @@ CREATE TABLE IF NOT EXISTS remote_agent_conversations (
 
 CREATE INDEX IF NOT EXISTS idx_remote_agent_conversations_codebase
   ON remote_agent_conversations(codebase_id);
-CREATE INDEX IF NOT EXISTS idx_conversations_hidden
-  ON remote_agent_conversations(hidden);
 CREATE INDEX IF NOT EXISTS idx_conversations_codebase
   ON remote_agent_conversations(codebase_id) WHERE deleted_at IS NULL;
 
@@ -156,8 +154,6 @@ CREATE INDEX IF NOT EXISTS idx_remote_agent_sessions_conversation
   ON remote_agent_sessions(conversation_id, active);
 CREATE INDEX IF NOT EXISTS idx_remote_agent_sessions_codebase
   ON remote_agent_sessions(codebase_id);
-CREATE INDEX IF NOT EXISTS idx_sessions_parent
-  ON remote_agent_sessions(parent_session_id);
 CREATE INDEX IF NOT EXISTS idx_sessions_conversation_started
   ON remote_agent_sessions(conversation_id, started_at DESC);
 
@@ -365,6 +361,10 @@ ALTER TABLE remote_agent_workflow_runs
 -- From migration 010: parent_session_id + transition_reason on sessions
 ALTER TABLE remote_agent_sessions
   ADD COLUMN IF NOT EXISTS parent_session_id UUID REFERENCES remote_agent_sessions(id);
+-- Index must follow the ADD COLUMN so it works on pre-existing Postgres DBs
+-- (CREATE TABLE IF NOT EXISTS is a no-op there, so the column only exists after this ALTER).
+CREATE INDEX IF NOT EXISTS idx_sessions_parent
+  ON remote_agent_sessions(parent_session_id);
 ALTER TABLE remote_agent_sessions
   ADD COLUMN IF NOT EXISTS transition_reason TEXT;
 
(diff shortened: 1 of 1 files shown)


Wirasm:
Thank you for this — your diagnosis was right, and it turned out to be a bigger class than either of us had visible at the time.

I'm closing this in favour of #2552, which folds in both of your moves and generalizes them. Not a rejection of the fix: the two indexes you identified (`idx_conversations_hidden`, `idx_sessions_parent`) are moved there too, and the reasoning in your PR description is what the new section header now says in the file.

What changed the shape of the fix:

- The same class had **five** live instances on Postgres, not two — `idx_conversations_codebase` (via its `WHERE deleted_at IS NULL` predicate), `idx_workflow_runs_parent_conv` and `idx_workflow_runs_last_activity` as well. And it is not only indexes: against a database old enough, `COMMENT ON COLUMN remote_agent_sessions.parent_session_id` fails the same way.
- **SQLite had it too, and there it was not latent.** `createSchema()` runs before `migrateColumns()` and was indexing `hidden`, `deleted_at` and `parent_conversation_id`, so `new SqliteAdapter(...)` threw `no such column: parent_conversation_id` on a database created before those columns existed — the adapter could not open it at all.

So rather than relocating individual statements, #2552 moves every `CREATE INDEX` and `COMMENT ON COLUMN` into one trailing section below the additive `ADD COLUMN` block, which makes the failure unrepresentable instead of something each author has to remember. It also adds `bun run check:schema-upgrades` and a CI job that applies the current schema on top of databases built from six release tags and compares the result against a fresh install — that check reproduces #2508 when pointed at v0.8.0's schema, and it is what would have caught your two instances automatically.

Your PR is credited in #2552's description. Sorry it sat open as long as it did.

PR #2446, opened 2026-08-04
Fix(workflows): validate runs before self-fix and simplify, so the green certifies code that no long

### LINKED ISSUE #2427: fix(workflows): validate runs before self-fix and simplify, so the green certifies code that no longer exists
`validate` runs before the two nodes whose entire purpose is changing code, and nothing re-validates after them. The workflow's green describes code that no longer exists by the time it commits.

## Verified dependency chain

`.archon/workflows/experimental/archon-fix-github-issue-experimental.yaml`:

```
validate    depends_on: [assert-implemented]     :445
create-pr   depends_on: [validate]               :492
synthesize  depends_on: [ ...reviewers ]         :656
self-fix    depends_on: [synthesize]             :667   <- MODIFIES CODE
simplify    depends_on: [self-fix]               :676   <- MODIFIES CODE
report      depends_on: [simplify]               :686
```

`self-fix` applies review findings. `simplify` removes code. Both run **after** the only validation gate, and nothing runs after them but `report`.

## Observed, not hypothetical

On a run against another repository, `simplify` deleted 5 lines across `Workbench.swift` and `FakeArchonClient.swift` **after** `validate` had passed. Commit `991c1d3` was therefore never compiled or tested by this workflow. The run reported success.

## Why this is the worst shape of the recurring problem

A gate that passes and is then invalidated is worse than no gate. No gate leaves the reader knowing the code is unverified; this one produces a green that is affirmatively wrong — it certifies a tree that was subsequently edited twice.

Same family as #2327 (a blocked node recorded `node_completed`) and #2402 (resume trusting a stale success), but here the stale signal is the *validation* signal, and it is what the PR body reports.

## Options

1. **Re-validate after `simplify`.** A second `validate` node between `simplify` and `report`. Cheapest and most obvious; costs one extra validation pass per run.
2. **Move `validate` to the end.** Single gate, but then `create-pr` opens a PR from an unvalidated tree, and the review agents review code that may not build.
3. **Gate the self-fix/simplify pair.** Have each re-run validation itself and revert its own change if it breaks the build. Most precise, most machinery.

(1) is the smallest change that removes the false green, and I would take it unless there is a reason not to. It also matches how a human works: change code, re-run the tests.

Whichever is chosen, `report` must not describe a tree that has not been validated since its last edit.

## Adjacent: `create-pr` failure discards the entire review half

On the same run, `create-pr` failed and took **eight** downstream nodes with it — the whole review, self-fix and simplify chain. That is structurally correct today (`review-scope` genuinely needs a PR to review), but it means a transient failure in PR creation throws away all the review work the run would otherwise have done.

Worth deciding separately whether the review half should be able to proceed against the branch when no PR exists. Not filed as its own issue yet because the right answer depends on the review commands, which read PR context throughout.

## Scope note

Line numbers above are from the experimental workflow. The bundled `archon-fix-github-issue` should be checked for the same ordering before fixing only one — it has historically carried the same shape, and two of today's fixes (#2417, #2420) had to be applied to both.

loonister1:
`validate` ran before `self-fix` and `simplify` — the two nodes that exist to change code — and nothing re-ran it, so `report` published a green describing a tree edited twice since the gate passed (`simplify` only runs type-check and lint over the files it touched). A `revalidate` node now sits between `simplify` and `report`, and `report` depends on it, so the run's last word is a full validation of the tree it actually leaves behind. The same node order existed in the bundled workflow and all three experimental variants, so all four were fixed.

Fixes #2427

### Testing
[full suite; 3 commit(s) checked individually] $ bun run test

### Notes for reviewers
1) The `bundled-defaults.generated.ts` hunk is the thing to check first: confirm it was produced by re-running the codegen script (whatever `bun run generate:bundled-defaults` or equivalent is) and not hand-edited — a hand-patched escaped-string blob will pass review and then drift on the next real regen. Diff the file after running codegen locally; it should be byte-identical.

## Summary by CodeRabbit

* **Improvements**
  * Added a final validation step after automated fixes and simplification.
  * Updated workflow reporting to reflect the final, validated project state.
  * Applied the improved validation flow across supported issue-fixing workflows.
  * Helps ensure reported results accurately match the completed project changes.

* **Tests**
  * Added coverage confirming the correct validation and reporting sequence across bundled workflows.


Wirasm:
Closing as superseded: the bundled defaults are being rewritten from scratch under #2123. The ordering constraint this PR enforces — validation must certify the final code, after self-fix and simplify — is carried into the rewrite as an authoring rule instead of being patched into a workflow that is being replaced.

Issue #2453, opened 2026-08-05
design(workflows): declared structured input/output contracts for composed workflows


Wirasm:
> **Re-scoped 2026-08-24 (additive; verified against `dev` at 60e5622e).** Two of the three concrete defects below have since landed and are no longer this issue's to own: the `when: $INPUTS` asymmetry is **fixed** (condition-evaluator now handles named inputs natively, with strict missing-input errors and did-you-mean hints; covered by subrun tests), and fan-out aggregate double-encoding is **fixed by #2637** (the aggregate is a logical array serialized once, with `archon_failed` marker objects for failed slots). The surviving defect — caller `output_format` on a `workflow:` node validated against nothing — is extracted as **#2774** so it can be delivered independently. #2637 also delivered whole-value logical transfer for `with:` refs and typed `inputs_values` persistence, answering most of design question 4. What this issue still owns: the callee-declared result contract (question 3, and #2774's end state), artifact-pointer values (question 8), field-access derivation from a callee contract (question 9), and declaration syntax (question 2) **only if a logged need earns it** — an eleven-workflow bundled pack has now been authored on whole-string handoffs plus #2470/#2637 without needing it. The constitution-row amendment note below stands.

> **Body rewritten 2026-08-13.** Basic named string inputs and stable return-node selection shipped in #2470 / PR #2523. This issue now owns the remaining structured value contract; it no longer claims that workflows have no signature at all.

> **Amended 2026-08-17 (additive).** A design pass traced the value lifecycle end to end against `dev`; findings are in *Evidence gathered against `dev`* below, including three concrete defects this issue now owns and the constitution row that must be amended rather than contradicted. Sequencing also changed: this issue now goes **after #1764**, so nothing here should be built only on the child-run metadata path. Nothing previously in this body was removed.

## Problem

Archon has the structural outline of a workflow signature:

```yaml
inputs:
  scope: { required: true }
  base_branch: { default: main }
returns: synthesize
```

Callers use `with:`, the callee reads `$INPUTS.<name>`, and `returns:` selects the node whose output is the workflow result.

What is still missing is a declared value contract:

- `with:` and input defaults are currently strings;
- input declarations describe presence/default/documentation, not value shape;
- `returns:` selects a node but declares no result shape;
- structured values are often serialized into text and parsed again by the next node/context window;
- caller `output_format` is not a callee-owned promise;
- there is no single rule for whole-value transfer versus interpolation into text;
[...]

Issue #2450, opened 2026-08-04
fix(workflows): enforce hermetic composed-workflow boundaries across every reference surface


Wirasm:
> **Body rewritten 2026-08-13.** Named `inputs:` / caller `with:` now provide the intended boundary. This issue is narrowed to making that boundary hermetic across every live reference surface.

## Problem

A composed workflow file must have an explicit interface. Every value crossing from the caller must arrive through a declared input supplied by `with:`. A composed workflow must not bind directly to a caller node merely because both files happen to contain a compatible node ID.

Current coverage is incomplete because reference parsing/rewriting is spread across multiple grammars and surfaces:

- canonical `$id.output[.field]` references;
- `when:` shorthand `$id.field`;
- inline versus command-backed bodies;
- recursively nested `loop_group` bodies;
- approval rejection prompts and other less-common live fields.

An omitted surface can accidentally bind to a caller node after composition. That makes the reusable block depend on hidden parent context that cannot be seen from its signature.

Command files are the preferred home for substantial prompts, so command-backed validation is core behavior rather than a best-effort edge case. Named script files are programs rather than prompt templates: they receive caller data only through declared input environment variables and must not gain implicit caller-node template access.

## Desired model

Each workflow file owns a lexical node namespace.

Inside a composed workflow:

- references to nodes owned by that workflow are internal and are rebound to deterministic composed IDs;
- references outside that workflow boundary are rejected;
- caller data enters only through declared `$INPUTS.<name>` values;
- `with:` may obtain those values from caller outputs, but the callee receives a bound value—not live access to the caller namespace;
- a reference never becomes valid merely because the parent happens to contain a node with the same ID.

Valid:

```yaml
# parent
- id: gather
  prompt: Gather context
- id: review
  include: review-block
  with:
    context: $gather.output
```

```yaml
# review-block
inputs:
  context: { required: true }
nodes:
  - id: judge
    prompt: Review $INPUTS.context
```

Invalid inside `review-block`, even if the parent has `gather`:

```yaml
- id: judge
  prompt: Review $gather.output
```

## Scope

- Define and enforce the workflow-file namespace boundary.
- Enumerate every live node-output reference and substitution surface, recursively.
- Parse canonical output references and `when:` shorthand consistently.
- Rewrite workflow-local references to composed IDs.
- Reject undeclared external/caller references before execution.
- Resolve and validate command-backed bodies using execution-equivalent precedence where possible.
[...]

PR #2459, opened 2026-08-05
test(workflows): guard against workflow-level schema fields being silently dropped at parse

### LINKED ISSUE #2457: parseWorkflow assembles field-by-field, so a workflow-level field in the schema but absent from the literal is silently discarded
## The trap

`parseWorkflow` does not derive its result from `workflowDefinitionSchema`. It hand-assembles a `WorkflowDefinition` field by field into an object literal (`packages/workflows/src/loader.ts:745-770`). A field added to the schema but not added to that literal is **silently dropped** — the YAML parses, the workflow loads, `/workflow list` shows it, and the feature is simply inert.

There is no guard. The failure produces no error, no warning, and no failing test.

## It has already happened, for six weeks

`requires: [github]` — the capability gate that hard-blocks a run when the originating user has not connected their GitHub identity.

| When | Commit | What |
|---|---|---|
| 2026-06-01 | `ab81248d` (#1823) | `requires: z.array(workflowRequirementSchema).optional()` added to `workflowBaseSchema`. **The loader was not touched** — `git show ab81248d -- packages/workflows/src/loader.ts \| grep -c requires` → `0` |
| 2026-07-16 | `2d7bf587` (#2129) | The loader assembly block finally added — **incidentally, inside the `include:` primitive PR**, not as a targeted fix |

The object literal at `ab81248d:packages/workflows/src/loader.ts:565-584` has no `requires` key. So for six weeks the field was in the schema, documented, and discarded at parse — meaning the GitHub identity gate **could never fire for any discovered workflow**. It was fixed as a drive-by in an unrelated PR.

The code knows about the trap. The comment that shipped with the fix says so in as many words (`loader.ts:664-670`):

> Without this block the field is silently discarded here and the capability gate can never fire for a discovered workflow.

Known, written down, and still unguarded.

## Why it will happen again

Every new workflow-level field needs two edits in two files, and only one of them is where the field is declared. Nothing connects them. There are currently 20 keys in `workflowDefinitionSchema.shape`; each one is a hand-maintained correspondence.

Note this is **not** the same gap #2213/#2455 closes. Unknown-key warnings catch a key the schema does not know about. This is the reverse: a key the schema *does* know about, which the loader drops. #2455's derivation (`KNOWN_WORKFLOW_KEYS` from `workflowDefinitionSchema.shape`) means such a field would not even warn — it is a known key, so it passes the check and is then discarded anyway.

## Suggested fix — a parity test, not a refactor

For each key in `workflowDefinitionSchema.shape`, parse a minimal YAML carrying that field and assert it survives into `parseWorkflow`'s result. Derive the key list from the shape so **a new schema field fails the test until it is given a fixture** — the same "generated check fails until the new thing is registered" pattern already used by `check:capability-matrix` (a new `ProviderCapabilities` field fails the generator until it gets a matrix axis) and by the schema-parity test in `sqlite.test.ts`.

Cost is roughly one test file and ~20 small fixtures. No production code changes and no semantic changes.

**Rejected alternative — deriving the assembly itself** (returning `workflowDefinitionSchema.parse(raw)` instead of hand-assembling) would eliminate the class outright, but the hand assembly exists *because* of warn-and-drop semantics: a present-but-invalid field is logged and dropped rather than rejecting the whole workflow, so one typo cannot abort a discovery pass. Zod's `.parse()` rejects instead. Getting warn-and-drop back per field would need `.catch()` plus a logging hook on every field — a real refactor with a behaviour-change risk, for the same protection the test gives for a fraction of the cost.

## Acceptance

- A test derives its field list from `workflowDefinitionSchema.shape` and fails when a schema field has no round-trip fixture
- Every field currently in the schema round-trips (this should pass today — the test is a ratchet, not a bug hunt)
- Adding a workflow-level field without wiring it into `parseWorkflow` fails CI

## Provenance

Found while planning #2453 (the workflow-signature design), whose `inputs:`/`returns:` fields would hit exactly this trap. That plan fixes it *for its own two fields*; this issue is the structural guard that stops the third and fifth field from repeating it.

Related: #2453, #2455 (the complementary unknown-key direction), #1823 and #2129 (where the `requires:` gap opened and closed).

Wirasm:
## Summary

- **Problem:** `parseWorkflow` hand-assembles its result field by field, so a field added to `workflowDefinitionSchema` but not to that object literal is **silently discarded** — the YAML parses, the workflow loads, and the feature is simply inert. Nothing catches it.
- **Why it matters:** it already happened. `requires:` was added to `workflowBaseSchema` in `ab81248d` (2026-06-01) without touching the loader; the assembly block only landed in `2d7bf587` (2026-07-16). Six weeks in which the GitHub capability gate could never fire for any discovered workflow — fixed incidentally inside an unrelated PR, not by anything noticing.
- **What changed:** a parity test whose field list is **derived** from `workflowDefinitionSchema.shape`, plus a per-field round-trip fixture. A new schema field fails the test until it is given a fixture.
- **What did NOT change:** zero production code. No behaviour change, no semantic change, no new dependency. `git diff --stat` is one test file.

### This is the third instance of one pattern

Parallel enumerations that must agree, with nothing enforcing agreement:

| Enumerations that must agree | Enforced? | Outcome |
|---|---|---|
| The three ref-surfaces (`loader.ts` scan / `rewriteNodeOutputRefs` / `substituteNodeOutputRefs` sites) | No — a comment at `loader.ts:226-231` | **Already broken** — #2450 found the `when:` shorthand escapes ref validation |
| `workflowDefinitionSchema.shape` vs `parseWorkflow`'s literal | No — nothing | **Already broken once** — `requires:`, six weeks |
| Node/workflow key sets vs their schemas | **Yes** — #2455 derives from `.shape` | Cannot drift |

The only difference in the third row is that it is *derived* rather than *described*. This PR applies that form to the second row.

### Why a parity test and not deriving the assembly

Deriving the assembly (`workflowDefinitionSchema.parse(raw)`) would eliminate the class outright, and it is the wrong trade here. The hand assembly exists **because of warn-and-drop**: a present-but-invalid field is logged and dropped rather than rejecting the whole workflow, so one typo cannot abort a discovery pass. `.parse()` rejects instead. Restoring per-field warn-and-drop underneath a derived parse needs `.catch()` plus a logging hook on all 20 fields — a real refactor with behaviour-change risk on the discovery path, for the same protection.

Note what #2455 actually derives: a set of **names**. Names are cheap to derive and carry no semantics. This PR derives the same thing.

### Before

```
Contributor            Schema                    parseWorkflow            Runtime
───────────            ──────                    ─────────────            ───────
adds `foo:` ────────▶  foo declared
                       (tests pass)
                                                 literal has no `foo`
                                                 → foo DROPPED silently
                                                                          feature inert
                                                                          no error, no warning
                                                                          no failing test
```

### After

```
Contributor            Schema                    parseWorkflow            Test suite
───────────            ──────                    ─────────────            ──────────
adds `foo:` ────────▶  foo declared
                                                 literal has no `foo`
                                                                     ***  RATCHET FAILS  ***
                                                                     "schema keys with no
                                                                      parity fixture: foo"
adds fixture ─────────────────────────────────▶
                                                                     *** ROUND-TRIP FAILS ***
                                                                     "declared, accepted without
                                                                      a warning, and still did
                                                                      NOT survive parseWorkflow"
wires the literal ────────────────────────────▶  literal carries foo  green
```

### Before

```
  schemas/workflow.ts                     loader.ts
  ┌────────────────────────┐              ┌──────────────────────────┐
  │ workflowDefinitionSchema│              │ parseWorkflow            │
  │   .shape (20 keys)     │   ...no...   │   per-field warn-and-drop │
  │                        │  - - - - - ▶ │   object literal (20 keys)│
  └────────────────────────┘   linkage    └──────────────────────────┘
```

### After

```
  schemas/workflow.ts                     loader.ts
  ┌────────────────────────┐              ┌──────────────────────────┐
  │ workflowDefinitionSchema│              │ parseWorkflow            │
  │   .shape (20 keys)     │              │   object literal (20 keys)│
  └───────────┬────────────┘              └────────────┬─────────────┘
              │                                        │
              │ Object.keys(...)  [+]                  │ round-trip  [+]
              ▼                                        ▼
        ┌──────────────────────────────────────────────────────┐
        │ [+] loader.test.ts — field-parity ratchet (#2457)     │
        │     derived key list === fixture list === survives    │
        └──────────────────────────────────────────────────────┘
```

**Connection inventory:**

| `schemas/workflow.ts` | `loader.test.ts` | **new** | Key list derived via `Object.keys(workflowDefinitionSchema.shape)` |
| `loader.ts` (`parseWorkflow`) | `loader.test.ts` | **new** | Per-field round-trip assertion |
| `schemas/workflow.ts` | `loader.ts` | unchanged | Still no enforced linkage — that is what the test now covers |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `tests`
- Module: `workflows:loader`

## Change Metadata

- Change type: `chore`
- Primary scope: `workflows`

## Linked Issue

- Closes #2457
- Related #2455 (the same guard, derived, one file over), #2450 (the first instance of the pattern, already broken), #2453 (where this was found — its `inputs:`/`returns:` fields would hit exactly this trap)

```bash
```

Grepped the log rather than trusting the wrapper's exit code:

```
$ grep -oE "[0-9]+ fail" /tmp/validate-2457.log | sort | uniq -c
 132   0 fail
$ grep -cE "[0-9]+ pass" /tmp/validate-2457.log
 132
```

132 test batches reported, every one `0 fail`, and the pass-line count matches the fail-line count so no batch silently failed to report.

**The guard was verified by breaking it, in both directions** — a test that has never been red proves nothing:

1. **Reproduce the historical bug.** Removed `...(requires !== undefined ? { requires } : {})` from `parseWorkflow`'s literal:
   ```
   (fail) workflow-level field parity (#2457) > round-trips 'requires' through parseWorkflow
   error: Field 'requires' is declared on workflowDefinitionSchema, was accepted without a
   warning, and still did NOT survive parseWorkflow — so it is missing from the object
   literal parseWorkflow returns. That is the #2457 bug: add the field to that literal.
   ```
2. **Add a new schema field the way a future contributor would.** Added `brand_new_field: z.string().optional()` to `workflowBaseSchema`, touching nothing else:
   ```
   (fail) workflow-level field parity (#2457) > has a fixture for every workflow-level schema key (the ratchet)
   error: Workflow-level schema keys with no parity fixture: brand_new_field. Add a fixture in
   FIELD_FIXTURES AND make sure parseWorkflow actually carries the field into its returned
   object literal — a schema field missing from that literal is silently discarded at parse.
   ```

Both breaks reverted; the diff is the test file alone.

## Security Impact (required)

Test-only change.

## Compatibility / Migration

- Backward compatible? **Yes** — no production code touched
- Config/env changes? **No**

- **Verified scenarios:** the guard is green on unmodified `dev` before anything else (all 20 keys already present in the literal, so this lands as a ratchet rather than an unscoped bug hunt — worth knowing, since a red first run would have been a different-sized PR); both deliberate breaks above.
- **Edge cases checked:** an *invalid* fixture value and a *genuinely dropped* field originally produced identical failures with a misleading message. Caught during development — my first `thinking:` fixture used `thinking: true`, which is not a valid `thinkingConfigSchema` value, so warn-and-drop correctly discarded it while the test blamed the loader. Fixed by clearing the mock logger before each parse and branching the message on whether a warning fired, so the two causes are told apart.
- **What was not verified:** nothing runs this outside `bun test`; there is no runtime component to exercise.

- **Affected subsystems:** `@archon/workflows` test suite only.
- **Potential unintended effects:** the fixture map is hand-maintained, so adding a workflow-level field now requires a fixture — that is the intended cost, and the failure message says exactly what to do.
- **Guardrails:** the test lives in `loader.test.ts`, which `packages/workflows/package.json` already runs in its own `bun test` invocation.

## Rollback Plan (required)

- **Fast rollback:** `git revert <sha>` — one test file, no production code.
- **Feature flags:** none.
- **Observable failure symptoms:** a red parity test where the field genuinely is wired, which would mean a bad fixture rather than a bug.

## Risks and Mitigations

- **Risk:** the fixture map drifts from reality — someone adds a fixture that passes without wiring the field.
  - **Mitigation:** not possible by construction. The fixture asserts the field survives `parseWorkflow`, so a fixture cannot be satisfied without the literal carrying the field.
- **Risk:** a future field is genuinely hard to fixture (preprocess/union/passthrough shapes).
  - **Mitigation:** `present` is a survival predicate, not deep equality, so `!== undefined` suffices — which is exactly how `thinking` and `sandbox` are handled today.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to ensure workflow fixtures stay aligned with the workflow schema.
  * Added validation that valid workflow fields are preserved during parsing without unexpected warnings.
  * Added checks to detect missing or outdated fixture fields.
  * Added safeguards to identify changes in workflow-level schema fields.

### DIFF
--- a/packages/workflows/src/loader.test.ts
+++ b/packages/workflows/src/loader.test.ts
@@ -35,6 +35,9 @@ registerBuiltinProviders();
 
 import { discoverWorkflows, discoverWorkflowsWithConfig } from './workflow-discovery';
 import { isBashNode, isCancelNode, isLoopNode } from './schemas';
+import { parseWorkflow } from './loader';
+import { workflowDefinitionSchema } from './schemas/workflow';
+import type { WorkflowDefinition } from './schemas/workflow';
 import * as bundledDefaults from './defaults/bundled-defaults';
 
 describe('Workflow Loader', () => {
@@ -4073,4 +4076,145 @@ nodes:
       }
     });
   });
+});
+
+// ---------------------------------------------------------------------------
+// Workflow-level field parity (#2457)
+// ---------------------------------------------------------------------------
+
+/**
+ * `parseWorkflow` does not derive its result from `workflowDefinitionSchema` — it
+ * hand-assembles a WorkflowDefinition field by field into an object literal. A field
+ * added to the schema but not added to that literal is SILENTLY DISCARDED: the YAML
+ * parses, the workflow loads, and the feature is simply inert.
+ *
+ * That is not hypothetical. `requires:` was added to `workflowBaseSchema` in ab81248d
+ * (2026-06-01) without touching the loader, and the assembly block only landed in
+ * 2d7bf587 (2026-07-16) — six weeks in which the GitHub capability gate could never
+ * fire for any discovered workflow, fixed incidentally inside an unrelated PR.
+ *
+ * This is the guard. The field list is DERIVED from `workflowDefinitionSchema.shape`,
+ * so a new schema field fails the test until it is given a fixture here — the same
+ * "the derived check fails until the new thing is registered" ratchet used by
+ * `check:capability-matrix` and the schema-parity test in `sqlite.test.ts`.
+ *
+ * Deliberately NOT solved by deriving the assembly itself (`schema.parse(raw)`): the
+ * hand assembly exists BECAUSE of warn-and-drop — a present-but-invalid field is logged
+ * and dropped rather than aborting the whole discovery pass — and `.parse()` rejects
+ * instead. See #2457.
+ */
+describe('workflow-level field parity (#2457)', () => {
+  /**
+   * One fixture per workflow-level schema key: a YAML fragment setting the field, and a
+   * predicate proving it survived `parseWorkflow`. `present` is deliberately a survival
+   * check rather than deep equality — several fields are normalised on the way through
+   * (tags deduped, betas trimmed, thinking preprocessed), and this guard is about the
+   * field reaching the result at all, not about how it is parsed.
+   */
+  const FIELD_FIXTURES: Record<
+    string,
+    { yaml: string; present: (w: WorkflowDefinition) => boolean }
+  > = {
+    name: { yaml: '', present: w => w.name === 'parity' },
+    description: { yaml: '', present: w => w.description === 'parity fixture' },
+    nodes: { yaml: '', present: w => w.nodes.length === 1 },
+    provider: { yaml: 'provider: claude', present: w => w.provider === 'claude' },
+    model: { yaml: 'model: sonnet', present: w => w.model === 'sonnet' },
+    modelReasoningEffort: {
+      yaml: 'modelReasoningEffort: high',
+      present: w => w.modelReasoningEffort === 'high',
+    },
+    webSearchMode: { yaml: 'webSearchMode: live', present: w => w.webSearchMode === 'live' },
+    interactive: { yaml: 'interactive: true', present: w => w.interactive === true },
+    effort: { yaml: 'effort: high', present: w => w.effort !== undefined },
+    thinking: { yaml: 'thinking: adaptive', present: w => w.thinking !== undefined },
+    fallbackModel: {
+      yaml: 'fallbackModel: haiku',
+      present: w => w.fallbackModel === 'haiku',
+    },
+    betas: { yaml: 'betas:\n  - some-beta', present: w => w.betas?.includes('some-beta') === true },
+    sandbox: { yaml: 'sandbox:\n  enabled: true', present: w => w.sandbox !== undefined },
+    worktree: { yaml: 'worktree:\n  enabled: false', present: w => w.worktree?.enabled === false },
+    container: {
+      yaml: 'container:\n  enabled: true',
+      present: w => w.container?.enabled === true,
+    },
+    evidence_policy: {
+      yaml: 'evidence_policy:\n  required: true',
+      present: w => w.evidence_policy?.required === true,
+    },
+    mutates_checkout: {
+      yaml: 'mutates_checkout: false',
+      present: w => w.mutates_checkout === false,
+    },
+    persist_sessions: {
+      yaml: 'persist_sessions: true',
+      present: w => w.persist_sessions === true,
+    },
+    tags: { yaml: 'tags:\n  - alpha', present: w => w.tags?.includes('alpha') === true },
+    requires: {
+      yaml: 'requires:\n  - github',
+      present: w => w.requires?.includes('github') === true,
+    },
+  };
+
+  const schemaKeys = Object.keys(workflowDefinitionSchema.shape);
+
+  it('has a fixture for every workflow-level schema key (the ratchet)', () => {
+    const missing = schemaKeys.filter(k => !(k in FIELD_FIXTURES));
+    expect(
+      missing,
+      `Workflow-level schema keys with no parity fixture: ${missing.join(', ')}. ` +
+        'Add a fixture in FIELD_FIXTURES AND make sure parseWorkflow actually carries the ' +
+        'field into its returned object literal — a schema field missing from that literal ' +
+        'is silently discarded at parse (see #2457).'
+    ).toEqual([]);
+  });
+
+  it('has no fixture for a key that is not in the schema', () => {
+    const stale = Object.keys(FIELD_FIXTURES).filter(k => !schemaKeys.includes(k));
+    expect(stale, `Parity fixtures for keys no longer in the schema: ${stale.join(', ')}`).toEqual(
+      []
+    );
+  });
+
+  for (const key of Object.keys(FIELD_FIXTURES)) {
+    it(`round-trips '${key}' through parseWorkflow`, () => {
+      const fixture = FIELD_FIXTURES[key];
+      const yaml = [
+        'name: parity',
+        'description: parity fixture',
+        fixture.yaml,
+        'nodes:',
+        '  - id: only',
+        '    prompt: hello',
+      ]
+        .filter(line => line !== '')
+        .join('\n');
+
+      // Warn-and-drop means an INVALID fixture value is dropped by design. Clearing the
+      // logger first lets the assertion below tell the two causes apart: a warn means the
+      // fixture is wrong, silence means the loader dropped a valid field (the #2457 bug).
+      mockLogger.warn.mockClear();
+
+      const result = parseWorkflow(yaml, `parity-${key}.yaml`);
+      expect(
+        result.error,
+        `parseWorkflow rejected the '${key}' fixture: ${result.error?.error}`
+      ).toBeNull();
+
+      const warned = mockLogger.warn.mock.calls.length > 0;
+      expect(
+        fixture.present(result.workflow as WorkflowDefinition),
+        warned
+          ? `Field '${key}' did not survive parseWorkflow, but a warning fired — the FIXTURE ` +
+              'value above is almost certainly invalid for this field, which warn-and-drop ' +
+              'discards by design. Fix the fixture, not the loader.'
+          : `Field '${key}' is declared on workflowDefinitionSchema, was accepted without a ` +
+              'warning, and still did NOT survive parseWorkflow — so it is missing from the ' +
+              'object literal parseWorkflow returns. That is the #2457 bug: add the field to ' +
+              'that literal.'
+      ).toBe(true);
+    });
+  }
 });
--- a/packages/workflows/src/loader.test.ts
+++ b/packages/workflows/src/loader.test.ts
@@ -4089,7 +4089,7 @@ nodes:
  * parses, the workflow loads, and the feature is simply inert.
  *
  * That is not hypothetical. `requires:` was added to `workflowBaseSchema` in ab81248d
- * (2026-06-01) without touching the loader, and the assembly block only landed in
+ * (2026-06-01) without touching the loader, and was not added to that literal until
  * 2d7bf587 (2026-07-16) — six weeks in which the GitHub capability gate could never
  * fire for any discovered workflow, fixed incidentally inside an unrelated PR.
  *
@@ -4098,10 +4098,13 @@ nodes:
  * "the derived check fails until the new thing is registered" ratchet used by
  * `check:capability-matrix` and the schema-parity test in `sqlite.test.ts`.
  *
- * Deliberately NOT solved by deriving the assembly itself (`schema.parse(raw)`): the
- * hand assembly exists BECAUSE of warn-and-drop — a present-but-invalid field is logged
- * and dropped rather than aborting the whole discovery pass — and `.parse()` rejects
- * instead. See #2457.
+ * Deliberately NOT solved by deriving the assembly itself (`schema.parse(raw)`): most
+ * fields warn-and-drop, logging a present-but-invalid value and continuing rather than
+ * aborting the whole discovery pass, and `.parse()` would reject the workflow instead.
+ * That is not universal — a few fields deliberately hard-reject and a few coerce
+ * silently — but one warn-and-drop field is enough to make a blanket `.parse()` wrong.
+ * `loader.ts` is the authority on which field does what; do not restate it here.
+ * See #2457.
  */
 describe('workflow-level field parity (#2457)', () => {
   /**
@@ -4117,7 +4120,7 @@ describe('workflow-level field parity (#2457)', () => {
   > = {
     name: { yaml: '', present: w => w.name === 'parity' },
     description: { yaml: '', present: w => w.description === 'parity fixture' },
-    nodes: { yaml: '', present: w => w.nodes.length === 1 },
+    nodes: { yaml: '', present: w => w.nodes?.length === 1 },
     provider: { yaml: 'provider: claude', present: w => w.provider === 'claude' },
     model: { yaml: 'model: sonnet', present: w => w.model === 'sonnet' },
     modelReasoningEffort: {
@@ -4126,14 +4129,14 @@ describe('workflow-level field parity (#2457)', () => {
     },
     webSearchMode: { yaml: 'webSearchMode: live', present: w => w.webSearchMode === 'live' },
     interactive: { yaml: 'interactive: true', present: w => w.interactive === true },
-    effort: { yaml: 'effort: high', present: w => w.effort !== undefined },
-    thinking: { yaml: 'thinking: adaptive', present: w => w.thinking !== undefined },
+    effort: { yaml: 'effort: high', present: w => w.effort === 'high' },
+    thinking: { yaml: 'thinking: adaptive', present: w => w.thinking?.type === 'adaptive' },
     fallbackModel: {
       yaml: 'fallbackModel: haiku',
       present: w => w.fallbackModel === 'haiku',
     },
     betas: { yaml: 'betas:\n  - some-beta', present: w => w.betas?.includes('some-beta') === true },
-    sandbox: { yaml: 'sandbox:\n  enabled: true', present: w => w.sandbox !== undefined },
+    sandbox: { yaml: 'sandbox:\n  enabled: true', present: w => w.sandbox?.enabled === true },
     worktree: { yaml: 'worktree:\n  enabled: false', present: w => w.worktree?.enabled === false },
     container: {
       yaml: 'container:\n  enabled: true',
@@ -4192,9 +4195,11 @@ describe('workflow-level field parity (#2457)', () => {
         .filter(line => line !== '')
         .join('\n');
 
-      // Warn-and-drop means an INVALID fixture value is dropped by design. Clearing the
-      // logger first lets the assertion below tell the two causes apart: a warn means the
-      // fixture is wrong, silence means the loader dropped a valid field (the #2457 bug).
+      // An INVALID fixture value is dropped by design, which looks identical to the bug
+      // this test hunts. Clearing the logger first lets the failure message rank the two
+      // causes: a warning is strong evidence the fixture is at fault. Silence is NOT
+      // proof of the opposite — a few fields coerce an invalid value away with no log at
+      // all — so the silent branch names both causes rather than rendering a verdict.
       mockLogger.warn.mockClear();
 
       const result = parseWorkflow(yaml, `parity-${key}.yaml`);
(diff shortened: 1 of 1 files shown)


Wirasm:
## PR Review Summary

**PR**: #2459 — `test(workflows): guard against workflow-level schema fields being silently dropped at parse`
**Diff**: `packages/workflows/src/loader.test.ts` (+144/−0), test-only
**Mode**: multi-agent (code-reviewer, code-simplifier, pr-test-analyzer, comment-analyzer) + independent verification

**Headline: the guard actually guards.** Verified empirically by two independent agents via mutation testing — deleting a field from `parseWorkflow`'s returned literal turns the matching round-trip test red with a correctly-attributed message, for every shape tested (`model` scalar, `persist_sessions` boolean, `tags` array, `worktree` object, `thinking` preprocessed). The ratchet fires in both directions (new schema key with no fixture; stale fixture for a removed key). No defects found in what the PR claims to do.

Everything below is polish: 2 cheap edits worth making, 4 optional.

---

### Critical Issues (0 found)

None.

---

### Important Issues (2 found)

| Agent | Issue | Location |
|-------|-------|----------|
| comment-analyzer | Doc comment states warn-and-drop as universal, but 3 of the fields in scope hard-reject | `packages/workflows/src/loader.test.ts:4102` |
| pr-test-analyzer | 3 of 20 `present` predicates are presence-only, so a wrong-but-valid value passes | `packages/workflows/src/loader.test.ts:4129,4130,4136` |

**I1 — the warn-and-drop justification is factually overstated.** The comment reads: *"the hand assembly exists BECAUSE of warn-and-drop — a present-but-invalid field is logged and dropped rather than aborting the whole discovery pass."* That is false for at least three fields, one of which (`evidence_policy`) is among the 20 under test:

- `evidence_policy` — hard-rejects the whole workflow (`loader.ts:610-625`), with its own comment explaining why: *"silently dropping a declared terminal-success gate would let a run complete ungated — not fail-safe."*
- unknown `provider` — hard-rejects (`loader.ts:432-443`)
- `persist_session` capability mismatch — hard-rejects (`loader.ts:483-496`)

Verified directly, not taken from the agent. The *conclusion* still holds — one warn-and-drop field is enough to make a blanket `.parse()` wrong — but as written a future reader would conclude the loader never rejects a field, which is materially untrue and is exactly the kind of authoritative-sounding comment that gets acted on.

Suggested rewrite:
> `...the hand assembly exists because MOST fields use warn-and-drop (a few — evidence_policy, unknown provider, persist_session capability — deliberately hard-reject), and a blanket .parse() would hard-reject on all of them.`

**I2 — `effort` / `thinking` / `sandbox` use `!== undefined` instead of a value check.** Confirmed by mutation: hardcoding the loader to return `{ type: 'disabled' }` for `thinking` (fixture asks for `adaptive`) and `'low'` for `effort` (fixture asks for `high`) leaves both tests **green**. The PR's own comment defends survival-over-equality on the grounds that fields are normalised in flight — but that excuse doesn't apply to these:

- `effortLevelSchema` is `z.enum(['low','medium','high','max'])` (`dag-node.ts:40`) — zero transformation. `w.effort === 'high'` is exact.
- `thinkingConfigSchema` (`dag-node.ts:56`) preprocesses `'adaptive'` → `{ type: 'adaptive' }` deterministically. `w.thinking?.type === 'adaptive'` is exact.
- `sandboxSettingsSchema.enabled` is `z.boolean().optional()` (`dag-node.ts:80`) and the fixture sets `enabled: true`. `w.sandbox?.enabled === true` is exact.

All three schema shapes verified independently. Note the other 17 fixtures all do real equality or `.includes()` checks — these 3 are the outliers, not the house style. This does **not** break the headline claim (whole-field omission is still caught for all 20); it means a transposition among these three specifically would ship silently. Three one-line edits.

---

### Suggestions (4 found)

| comment-analyzer | "the assembly block only landed in 2d7bf587" is fragile phrasing | `loader.test.ts:4092` |
| comment-analyzer | The "warned ⇒ blame the fixture" diagnosis rests on an undocumented invariant | `loader.test.ts:4206-4213` |
| code-simplifier | Extract the assertion ternary into a named `message` | `loader.test.ts:4209` |
| code-reviewer | `nodes` predicate throws rather than returning false | `loader.test.ts:4120` |

**S1** — the hand-assembly literal (with `tags`, `worktree`, `betas`, …) predates `2d7bf587`; only the `requires` entry was added there. Context makes the intent clear, but a skim reads as "the whole mechanism didn't exist until then." Suggested: `and requires wasn't added to that literal until 2d7bf587`.

**S2** — "a warning fired ⇒ your fixture value is wrong" is correct *today*: every field-specific `getLog().warn` in `parseWorkflow` is gated on its own field being present-and-invalid, and the fixed base skeleton can't trigger the ungated ones. Both the reviewer (20 call sites) and comment-analyzer (21) enumerated this independently. But it's an unstated invariant — add an inter-field warning later (e.g. "container.enabled together with worktree") and the message misdirects. Worth one line near `FIELD_FIXTURES`. Note the assertion itself is unaffected: `warned` only picks which failure string to print, so the worst case is a confusing message on an already-failing test, never a false pass.

**S3** — the two long diagnostic strings inline in the ternary bury the assertion. Same behaviour, easier to scan:
```ts
const message = warned ? `…fix the fixture, not the loader.` : `…add the field to that literal.`;
expect(fixture.present(result.workflow as WorkflowDefinition), message).toBe(true);
```

**S4** — `nodes: { present: w => w.nodes.length === 1 }`. `nodes` is an unconditional property, so deleting it from the literal yields `undefined.length` → `TypeError` rather than a clean `false`. Bun still reports a failing test, so the ratchet holds; the diagnostic is just uglier. `w.nodes?.length === 1` if you care.

**Explicitly considered and NOT recommended** (raised, then argued down):
- Deleting the ~15 round-trip fixtures that overlap existing `discoverWorkflows`-based tests. The loop is the ratchet's *enforcement leg* — without it, the fixture-existence test could be satisfied by a bogus fixture. It's also pure (no file I/O) where the existing tests aren't.
- Replacing the 20 `present` closures with a declarative `{path, expect}` table. Needs a branch per shape anyway; trades 20 self-evident one-liners for a small interpreter. Regression against KISS.
- Swapping the `yaml: ''` sentinel for an optional key. A wash that churns all 20 literals for no readability gain.
- Trimming the 30-line preamble. Checked against the precedent it cites (`sqlite.test.ts` parity block) — same style, in-family, not bloated. Only the inline git archaeology is rot-prone, and even that is describing a *closed historical incident*, so it won't silently drift.

---

### Strengths

- **The guard was verified by breaking it, in both directions** — by the author in the PR body, and reproduced independently here by two agents on five different field shapes. A ratchet that has never been red proves nothing; this one has.
- **The derivation is genuinely sound, not decorative.** `workflowDefinitionSchema = workflowBaseSchema.extend({ nodes })` is a plain `ZodObject` with no `.refine`/`.transform`/`.passthrough`/union wrapper, so `.shape` really does enumerate all 20 keys — the failure mode where `.shape` silently returns a subset doesn't apply.
- **Correctly scoped, and the scope is complete.** Node-level fields go through `dagNodeSchema.safeParse()` and are returned verbatim (`loader.ts:109`, `764`) — no hand-assembly, so the bug class is structurally impossible there. The web write path uses `Bun.YAML.stringify(definition)` and the server schema is re-derived from the engine one. Workflow-level really is the only exposed surface.
- **The historical claim holds up under independent check.** `ab81248d` (2026-06-01) adds `requires` to the schema and leaves `loader.ts` untouched; `2d7bf587` (2026-07-16, the `include:` primitive PR) adds it to the literal. Both the enforcement path *and* the schema field landed in `ab81248d`, so the gate was genuinely dead for the full ~6.4 weeks — the claim is not inflated. Nice restraint on a story that would have been easy to overstate.
- **The two-branch failure message is a real piece of craft.** Distinguishing "your fixture value is invalid" from "the loader dropped a valid field" is the difference between a 2-minute fix and an hour of blaming the wrong file — and per the PR body it was found the hard way during development.
- **Rejecting the tempting refactor was the right call.** `schema.parse(raw)` would kill the class outright but trades warn-and-drop for whole-workflow rejection on the discovery path. Test-shaped guard, zero production risk, revert is one file.
- **Test isolation is clean.** The new block is a sibling of `describe('Workflow Loader')` and so misses its `beforeEach` — harmless, because `parseWorkflow` is pure (no `fs`, no `process.env`, no `ARCHON_HOME`; grep-confirmed). `mockLogger.warn.mockClear()` runs inside each `it()`, so no cross-test bleed. Verified deterministic three ways: standalone, under `bun --filter @archon/workflows test`, and isolated via `-t`.
- `loader.test.ts` already runs in its own `bun test` invocation in `packages/workflows/package.json` — no new mock-pollution batch needed.

---

### Documentation Issues

- **`CLAUDE.md` is stale, unrelated to this PR** — it states the `sqlite.test.ts` parity test *"compares table names only"* and that *"a column present in one dialect and missing in the other passes CI."* That is no longer true: `sqlite.test.ts:468-695` now runs three tests (table names, Postgres→SQLite columns, SQLite→Postgres columns) with tracked allowlists `POSTGRES_ONLY_COLUMNS` (`remote_agent_codebases.allow_env_keys`) and `SQLITE_ONLY_COLUMNS` (`remote_agent_isolation_environments.updated_at`), each annotated with #2318. Verified directly. Worth a separate CLAUDE.md fix. Ironically this means #2459's analogy to that test is *more* accurate than CLAUDE.md's own description of it.

---

### Verdict

**NEEDS FIXES (minor)** — no defects in the guard; 4 one-line edits worth making before merge.

Nothing found blocks correctness, and no agent requested changes to the mechanism. Both Important items are cheap and both make a precision-tool more precise, which is the whole point of the artifact.

---

### Recommended Actions

1. **I1** — rewrite the warn-and-drop sentence at `loader.test.ts:4102` to say *most* fields, naming `evidence_policy` as the counterexample. (1 line)
2. **I2** — tighten three predicates: `w.effort === 'high'`, `w.thinking?.type === 'adaptive'`, `w.sandbox?.enabled === true`. (3 lines)
3. Optionally take S1–S4 in the same commit — all one-liners.
4. No re-review needed; re-run `bun test packages/workflows/src/loader.test.ts` and merge.

---

*Reviewed by Claude — 4 specialist agents, findings independently re-verified against the code before inclusion.*
*Report: `/Users/rasmus/.prp/archon-75601ef6/reviews/pr-2459-review.md`*

Wirasm:
## Addendum to the review above

The comment-analysis pass finished after I posted, with a field-by-field audit that sharpens **I1** and promotes one suggestion to **I3**. Both re-verified against `loader.ts` directly.

### I1 is wider than I first reported

I said three fields contradict the "warn-and-drop" claim. The full audit of all 20 says **only 13 actually warn-and-drop**:

| Behaviour | Count | Fields |
|---|---|---|
| **Hard-reject** (whole workflow fails) | 4 | `name` (`loader.ts:340-346`), `description` (`:347-357`), `nodes` (`:382-392`, `:400-419`), `evidence_policy` (`:615-630`) |
| **Warn-and-drop** (matches the comment) | 13 | `modelReasoningEffort`, `webSearchMode`, `interactive`, `effort`, `thinking`, `sandbox`, `fallbackModel`, `betas`, `worktree`, `container`, `mutates_checkout`, `tags`, `requires` |
| **Silent coerce, no log at all** | 3 | `provider` (`:424-425`), `model` (`:426`), `persist_sessions` (`:474`) |

Verified there is no `invalid_provider` / `invalid_model` / `invalid_persist_sessions` warn event anywhere in the file. So the docblock overstates in *both* directions — 4 fields never get dropped, and 3 get dropped more quietly than it describes.

### New — I3: the warn/no-warn discriminator is blind for exactly those 3 fields

The two-branch failure message is the best thing in this diff: a warn means *"your fixture value is invalid"*, silence means *"the loader dropped a valid field — that's the #2457 bug."* Sound for 17 fields. **Backwards for `provider`, `model`, `persist_sessions`**, because those coerce with no warn.

Write an invalid fixture for one of them — `provider: 123`, `persist_sessions: "true"` — and the field is discarded, nothing warns, and the test confidently prints *"add the field to that literal,"* sending the next contributor into `loader.ts` when the fixture is what's wrong.

That is precisely the failure the PR body describes hitting during development (the bad `thinking: true` fixture that blamed the loader) and that this message exists to prevent — it just doesn't cover the three fields where the loader stays silent. No false pass, and no issue with the fixtures as written; it's a trap for whoever edits the table next. One line near `FIELD_FIXTURES` naming the three exceptions closes it, and covers S2 too.

### Revised actions

1. **I1** — rewrite `loader.test.ts:4102`: most fields warn-and-drop, `evidence_policy`/`name`/`description`/`nodes` hard-reject, `provider`/`model`/`persist_sessions` coerce silently.
2. **I2** — tighten `effort` / `thinking` / `sandbox` predicates to value checks. (unchanged)
3. **I3** — one line noting the discriminator's three exceptions.
4. Optionally S1, S3, S4.

Still no defect in the guard itself — all three items are comment/predicate precision on an artifact whose entire job is precision. Verdict unchanged: **NEEDS FIXES (minor)**.

Wirasm:
Pushed `c078e604` — **all six items taken, none declined.** I re-verified the field audit against `loader.ts` rather than accepting it, and it holds: 4 hard-reject, 13 warn-and-drop, 3 coerce silently. Two items are implemented differently from the suggestion; reasoning below.

| Item | Call | Note |
|---|---|---|
| I1 | **fix now** | Rewritten, but *not* as a per-field table — see below |
| I2 | **fix now** | Three predicates tightened; verified by mutation |
| I3 | **fix now** | Fixed the *message*, not by listing the three exceptions — see below |
| S1 | fix now | One-line rewording |
| S2 | **subsumed by I3** | The message no longer rests on the invariant, so there is nothing left to document |
| S3 | fix now | Named `message` — matters more now that I3 lengthened the strings |
| S4 | fix now | `w.nodes?.length` |

## I1 — accurate, but deliberately not a table

The suggested rewrite enumerated the categories with field names. I didn't take that shape: a three-way per-field table in a test docblock duplicates `loader.ts` and rots the moment one field changes category — the same class of problem as the counts this PR's sibling work is removing from `CLAUDE.md`. It now says *most* fields warn-and-drop, notes that a few hard-reject and a few coerce silently, keeps the load-bearing conclusion (one warn-and-drop field is enough to make a blanket `.parse()` wrong), and ends with **"`loader.ts` is the authority on which field does what; do not restate it here."**

## I3 — the real fix was the message, not a list of exceptions

You're right that this is the sharpest finding, and it is the same failure the message exists to prevent — I hit it during development with a bad `thinking: true` fixture.

Rather than name the three silent fields in a comment (which rots identically), I removed the false verdict. A warning is still strong evidence the fixture is wrong. **Silence is no longer treated as proof of the opposite** — it now names both causes, ranked, and sends the reader to the cheaper one first:

> Field 'provider' is declared on workflowDefinitionSchema and did NOT survive parseWorkflow, with no warning logged. Two possible causes, likeliest first: (1) the field is missing from the object literal parseWorkflow returns — that's the #2457 bug, add it there; or (2) the fixture value is invalid for a field that coerces silently without logging, in which case fix the fixture. Check the fixture value against the schema first — it is the cheaper of the two to rule out.

That is the actual output, captured by running a deliberately invalid `provider: 123` fixture — the exact trap in the addendum. It no longer misdirects, and it stays correct if a fourth field joins the silent set.

## I2 — verified by mutation, not by inspection

Making the loader return `effort: 'low'` and `thinking: { type: 'disabled' }` now fails both round-trips; before the change both stayed green. Your point that 17 of 20 already do value checks was the deciding argument — these three were outliers, not house style.

## Verification

`bun run validate` green — 132 batches, all `0 fail`, grepped from the log rather than trusting the wrapper exit code. Plus the two targeted checks above.

## One thing from your review I'm acting on elsewhere

The **Documentation Issue** — `CLAUDE.md` claiming the `sqlite.test.ts` parity test *"compares table names only"*. Confirmed stale: that file now runs three tests with `POSTGRES_ONLY_COLUMNS` and `SQLITE_ONLY_COLUMNS` allowlists (`sqlite.test.ts:495,500`). It's going into #2463, the CLAUDE.md accuracy PR, where it belongs rather than into this test-only diff. Good catch — and you're right that it makes this PR's analogy more accurate than CLAUDE.md's own description of the thing it points at.

Issue #2449, opened 2026-08-04
feat(workflows): a sub-workflow cannot declare what it returns — the result is a positional accident (first sink, definition order)


Wirasm:
The input half of a sub-workflow's interface is tracked in #2214 (`with:` / `$INPUTS`). This is the output half, which is not tracked anywhere.

## The problem

A sub-workflow cannot declare what it returns. The engine infers it:

> Terminal output — **first sink node, non-blank, definition order** — becomes the run's `metadata.summary`, which `childOutcomeFromRun` reads back as the child's `$<nodeId>.output`.

`include:` uses the same rule for `$<includeId>.output` (the block's primary sink, first in definition order).

So a reusable block's return value is a **positional accident**. Insert a node that happens to become an earlier sink, or reorder two nodes, and the block's return value changes. Nothing errors, no test fails, and every caller silently receives something different.

This is the "key coarser than the thing it identifies" shape recorded elsewhere in this project — the rule returns a plausible answer rather than an error, which is what makes it expensive to diagnose.

## The asymmetry that gives it away

The **caller** can assert a return shape: `output_format` on a `workflow:` node validates the child's output against a schema, and `$<node>.output.field` access is strict.

The **callee** can promise nothing.

So the type declaration sits on the wrong side of the boundary. `output_format` is an assertion into the dark rather than a check against a contract — and when it fails, it fails at the caller with no way to attribute the mismatch to a change inside the block.

## Why it matters for composition

The stated goal is workflows composed from smaller workflows: prewritten commands and scripts → small workflows → larger workflows. That needs blocks with signatures. Today a block has neither half:

| | declared by | today |
|---|---|---|
| inputs | nobody | `include:` receives nothing; `workflow:` receives one untyped string. `with:` rejected at load (#2214) |
| outputs | nobody | first non-blank sink, definition order |

Without a declared return, a block cannot be safely edited by anyone who does not know all of its callers — which is precisely the property that makes the current copy-paste defaults drift instead of being factored. `verify-pr-base` is already 7 copies in 5 variants.

## Shape of a fix (not a decision)

Some way for a workflow to declare its own return — the node whose output is the workflow's result, and optionally the schema that result conforms to. Then `output_format` on the caller becomes a check against a promise, and reordering nodes inside a block cannot silently change what it returns.

Open questions worth settling before code:

- Is the declaration a workflow-level field naming a node, or a marker on the node itself?
[...]

PR #2460, opened 2026-08-05
fix(workflows): stop a declared output_format from silencing an unparseable output

### LINKED ISSUE #2456: output_format on a workflow: node inverts the failure mode — declaring a schema makes a broken child quieter
## The bug

`output_format` on a `workflow:` (sub-run) node is documented and shaped like an output contract. It is not one. It performs **no validation**, and its only observable effect is to **suppress** the error that would otherwise fire when a child returns something unusable.

Declaring a schema therefore makes a broken child *quieter* than not declaring one. That is backwards, and it is a bug rather than a limitation: the field exists, it is deliberately excluded from `WORKFLOW_NODE_IGNORED_FIELDS` so it stays "meaningful", and what it actually does is the opposite of what an author declaring it expects.

## Reproduction

Two calls to `resolveNodeOutputField`, differing only in whether the producer declared a schema. Both with a child whose terminal output is prose, not JSON:

```ts
// A — output_format DECLARED on the workflow: node
resolveNodeOutputField(
  { state: 'completed', output: 'I could not produce JSON, sorry.',
    declaredFields: ['verdict', 'reason'] },
  'child', 'verdict'
)
// → { kind: 'empty' }      ...silently ''

// B — NO output_format (identical child output)
resolveNodeOutputField(
  { state: 'completed', output: 'I could not produce JSON, sorry.' },
  'child', 'verdict'
)
// → THROWS OutputRefError:
//   '\$child.output.verdict' references field 'verdict', but node 'child's output
//   is not a JSON object, so the field cannot be read.
```

Executed against `dev` (8704a65f), bun 1.3.11. So `\$child.output.verdict` yields an empty string in A and a loud, attributable failure in B — and A is the one where the author wrote down what they expected.

## Where it happens

**`packages/workflows/src/output-ref.ts:160-173`** — the declared-schema branch. When the producer has `declaredFields`, an output that will not parse falls straight through to `{ kind: 'empty' }`:

```ts
const obj = structuredObj ?? parseOutputObject(nodeOutput.output);
if (obj === undefined) return { kind: 'empty' };
```

**`packages/workflows/src/output-ref.ts:186-191`** — the schemaless branch, which is strict:

```ts
const obj = parseOutputObject(nodeOutput.output);
if (obj === undefined) throw new OutputRefError(nodeId, field, 'unparseable');
```

**`packages/workflows/src/dag-executor.ts:5362`** — the only use of `node.output_format` in `executeWorkflowNode`:

```ts
const declaredFields = declaredFieldsFromSchema(node.output_format);
```

No parse of the child output against the schema, no validation, no re-ask. Contrast the AI-node path at `dag-executor.ts:2068-2172`, which validates structured output against the declared schema, re-asks up to 3x on best-effort providers, and **fails the node** when nothing schema-valid comes back.

## Why the asymmetry is defensible for AI nodes and not here

On an AI node, `declaredFields` implying "optional-absent, not a typo" is sound: the node already validated the payload post-parse, so a missing key really is a declared-optional field. On a `workflow:` node nothing validated anything, so the same leniency is applied to a payload that was never checked — and the "not a JSON object at all" case is swallowed along with it.

## Suggested fix

Two options; the second is smaller and probably right:

1. Validate the child's terminal output against `output_format` in `executeWorkflowNode` and fail the node when it does not conform — matching the AI-node contract and the documented `structuredOutput` posture ("structured-output misses fail rather than degrade").
2. At minimum, keep the unparseable throw regardless of `declaredFields`: the `obj === undefined` case at `output-ref.ts:167` should throw `'unparseable'`, not return empty. Declared-optional leniency should apply to a *missing key in a parsed object*, never to *no object at all*.

## Why this is filed separately

Found while investigating #2453 (the workflow-signature design). It falsifies a premise in #2449, which states that the caller "can assert a return shape" via `output_format` — the caller cannot assert anything, and the field it does have is a footgun. Kept out of the signature work deliberately: it is a pre-existing defect on a path that work does not otherwise touch, and folding a behaviour change into a language-surface PR would blur the blast radius.

Related: #2449 (assumes this field validates), #2453 (where it was found), #2121 (sub-runs).

Wirasm:
## Summary

- **Problem:** `resolveNodeOutputField` treated "no parseable object at all" as a declared-optional field and returned empty — but **only on the declared-schema path**. The schemaless path threw. So declaring `output_format` made a broken producer *quieter* than declaring nothing.
- **Why it matters:** it bites hardest on `workflow:` sub-run nodes, whose `output_format` is **never validated against the child**. `executeWorkflowNode` uses it for exactly one thing — deriving `declaredFields` (`dag-executor.ts:5362`). A child returning prose instead of JSON turned every declared field into `''` with no error, no warning, and no failing test.
- **What changed:** both paths now throw `unparseable`. Two tests plus a scope guard, and a docs clarification.
- **What did NOT change:** the leniency that was actually intended. A declared field missing from a payload that **genuinely parsed** still resolves to `''`. Only "there is no object to read" changed. `not-in-schema`, `producer-not-run`, `missing-key`, and the lenient no-schema `structuredOutput` path are all untouched.

### The asymmetry, before

```ts
// packages/workflows/src/output-ref.ts — declared-schema branch
const obj = structuredObj ?? parseOutputObject(nodeOutput.output);
if (obj === undefined) return { kind: 'empty' };     // ← silent

// ...schemaless branch, 20 lines below
const obj = parseOutputObject(nodeOutput.output);
if (obj === undefined) throw new OutputRefError(nodeId, field, 'unparseable');
```

Same input, same question, opposite answers — and the *more* declared one was the *less* safe one.

### Before

```
Author                 Child sub-run            Parent node
──────                 ─────────────            ───────────
declares
  output_format:
    verdict, reason
                  ───▶ returns prose
                       (nothing validates it)
                                          ───▶  $review.output.verdict → ''
                                                $review.output.reason  → ''
                                                node SUCCEEDS on empty values
                                                downstream acts on ''
```

An author who declared *nothing* got a loud failure on the same child. The reward for writing the contract down was losing the error.

### After

```
Author                 Child sub-run            Parent node
──────                 ─────────────            ───────────
declares
  output_format:
    verdict, reason
                  ───▶ returns prose
                                          ***  NODE FAILS  ***
                                          "'$review.output.verdict' references field
                                           'verdict', but node 'review's output is not
                                           a JSON object, so the field cannot be read."

                  ───▶ returns {"verdict":"pass"}
                                          ───▶  verdict → 'pass'
                                                reason  → ''   [unchanged: declared-optional]
```

### Before

```
                     resolveNodeOutputField (output-ref.ts)
                     ┌──────────────────────────────────────┐
  declaredFields ───▶│ 1. declared-schema  → empty on unparseable  ← INCONSISTENT
  structuredOutput ─▶│ 2. no-schema payload → lenient              │
  raw text ─────────▶│ 3. schemaless        → THROWS on unparseable│
                     └──────────────────────────────────────┘
                             ▲                    ▲
        dag-executor.ts:2364 │ AI node            │ condition-evaluator (when:)
        dag-executor.ts:5362 │ workflow: node ────┘ substituteNodeOutputRefs (prompts)
        dag-executor.ts:7173 │ resume rehydration
```

### After

```
                     resolveNodeOutputField (output-ref.ts)
                     ┌──────────────────────────────────────┐
  declaredFields ───▶│ [~] 1. declared-schema → THROWS on unparseable │
  structuredOutput ─▶│     2. no-schema payload → lenient             │
  raw text ─────────▶│     3. schemaless        → THROWS on unparseable│
                     └──────────────────────────────────────┘
                       branches 1 and 3 now agree === (that symmetry is the contract)
```

**Connection inventory:**

| `dag-executor.ts:5362` (`workflow:` node) | `output-ref.ts` branch 1 | **modified** | The case this fixes — declaredFields with no validation behind it |
| `dag-executor.ts:2364` (AI node) | `output-ref.ts` branch 1 | unchanged in practice | A validated AI node persists JSON (see below), so it never reaches the changed line |
| `dag-executor.ts:7173` (resume) | `output-ref.ts` branch 1 | unchanged in practice | Rehydrates text only, but that text is the serialized JSON |
| `condition-evaluator.ts` (`when:`) | `output-ref.ts` | unchanged | Same resolver, so `when:` gains the same loudness |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: XS`
- Scope: `workflows`
- Module: `workflows:output-ref`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2456
- Related #2449 (this falsifies its premise that a caller can assert a return shape via `output_format`), #2453 (the signature design where it was found), #2121 (sub-runs)

```bash
```

Grepped rather than trusting the exit code:

```
$ grep -oE "[0-9]+ fail" /tmp/validate-2456.log | sort | uniq -c
 132   0 fail
$ grep -cE "[0-9]+ pass" /tmp/validate-2456.log
 132
```

**Tests confirmed red before the fix:**

```
(fail) resolveNodeOutputField — declared-schema producer > unparseable output → throws, exactly like the schemaless path (#2456)
(fail) resolveNodeOutputField — declared-schema producer > declaring a schema is never quieter than declaring none (#2456)
 28 pass, 2 fail
```

After: `30 pass, 0 fail`.

**No existing test depended on the old behaviour** — a useful signal that the leniency was never deliberately exercised, only inherited.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? **No — this is a deliberate behaviour change.**
- Config/env changes? **No**

A workflow that today silently receives `''` from a non-JSON producer with a declared schema will now **fail loudly**. That is the point, and it matches the documented engine posture (unresolvable refs fail loudly; structured-output misses fail rather than degrade).

**Blast radius inside this repo is zero:** no workflow in the tree uses `workflow:` nodes — verified against a positive control, since an empty grep is a claim rather than a finding. User workflows may be affected, so this belongs in the release notes.

- **Verified scenarios:** both failing tests before/after; the declared-optional case still returns `''`; a parsed payload with a missing key is still lenient; `not-in-schema` still throws first.
- **Edge cases checked — the one that decided the shape of the fix:** whether this regresses **resume**. The resume path (`dag-executor.ts:7160-7178`) rehydrates *text only*, with no `structuredOutput`, so a declared node falls through to `parseOutputObject`. If an AI node persisted prose while holding JSON separately, this change would start throwing on resume for every such node. It does not: when structured output validates, `nodeOutputText` is **overwritten** with `JSON.stringify(structuredOutput)` (`dag-executor.ts:2129-2132`), and an AI node with `output_format` that fails to produce schema-valid output never completes at all (`dag-executor.ts:2172`). So a completed AI node always persists JSON. `workflow:` nodes have no such guarantee, which is exactly the case being fixed — and it behaves identically fresh and resumed.
- **Edge cases checked:** a non-object `output_format` (no `properties`) yields `declaredFields === undefined` and is unaffected; `properties: {}` throws `not-in-schema` before reaching the changed line.
- **What was not verified:** no live end-to-end sub-run was executed against a real provider — there are no `workflow:` nodes in the tree to run, and the behaviour is fully determined by a pure function that is directly unit-tested.

- **Affected subsystems:** every `$node.output.field` consumer — prompt substitution, `script:`/`bash:` substitution, and `when:` conditions all route through this one resolver, so `when:` gates on a broken declared producer now fail loudly instead of evaluating against `''`.
- **Potential unintended effects:** a workflow that *relied* on the empty-string fallback (deliberately or not) starts failing. There is no silent-degradation path left for that case by design.
- **Guardrails:** the error is `OutputRefError` with `reason: 'unparseable'`, thrown inside the executor's per-node try/catch, so it surfaces as an attributable node failure with the producer id and field name in the message — not a run-level mystery.

## Rollback Plan (required)

- **Fast rollback:** `git revert <sha>` — a one-line behaviour change plus tests and a docs paragraph.
- **Feature flags:** none. A flag here would preserve the silent path, which is the thing being removed.
- **Observable failure symptoms:** nodes failing with `references field 'X', but node 'Y's output is not a JSON object`. If that appears somewhere unexpected, the producer genuinely is not emitting JSON — the message names both ends.

## Risks and Mitigations

- **Risk:** a user workflow depends on the silent `''` and breaks on upgrade.
  - **Mitigation:** the failure names the node, the field, and the fix ("Emit JSON containing 'X', or reference `$Y.output` (whole text) instead"). Flagged for release notes.
- **Risk:** this reads as fixing `output_format` on sub-runs, when it only fixes the *inversion*.
  - **Mitigation:** stated explicitly here and in the commit. A child emitting JSON that does **not match** the declared schema still passes — `workflow:` nodes still perform no validation. That half of #2456 (option 1 in the issue) remains open deliberately: it is a larger change to `executeWorkflowNode`, and it belongs with the `returns:` work in #2453 rather than in a one-line correctness fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Field references now consistently report errors when producer output cannot be parsed as a JSON object, including workflow sub-runs.
  * Declared output schemas no longer hide invalid producer output.
  * Optional fields declared but absent from valid parsed objects continue to resolve to an empty string.
  * Truncated output now reports a specific error, while embedded marker text remains invalid output.

* **Documentation**
  * Clarified field-reference behavior for invalid, truncated, and schema-declared output.


Wirasm:
## Review: single-pass check

**Verdict: NEEDS FIXES (minor)** — 1 docs inaccuracy to correct. The code change itself is right and I'd merge it as-is. 0 critical, 1 important, 2 suggestions.

The core diagnosis is correct and well-evidenced: the declared-schema path returned `{ kind: 'empty' }` where the schemaless path threw, so declaring `output_format` made a broken producer *quieter* than declaring nothing. That inverts the documented contract in CLAUDE.md (*"a node that declares `output_format` but returns no schema-valid output **fails** rather than degrading silently"*). The fix aligns the two paths and preserves the leniency that was actually intended.

### Verified

| Check | Result |
|---|---|
| `bun test src/output-ref.test.ts` | 30 pass / 0 fail |
| `bun --filter @archon/workflows test` (all 20 batches) | 0 fail across every batch |
| New reason value needed? | No — `'unparseable'` already exists in `OutputRefErrorReason` (`output-ref.ts:44-49`), so no consumer `switch` goes non-exhaustive |
| Both consumers get it identically | Confirmed — `dag-executor` substitution and `condition-evaluator` (`when:`) share `resolveNodeOutputField`, so the contract stays symmetric |
| Scope claim ("only *no object at all* changed") | Confirmed — `missing-key`, `not-in-schema`, `producer-not-run`, and the lenient path-2 `structuredOutput` branch are untouched |

---

### Important (1)

**I1 — the new docs sentence says "load-visible", and it isn't.** `packages/docs-web/src/content/docs/reference/variables.md`:

> Declaring a schema buys you a stricter check on the field *name* (an undeclared field is a **load-visible** mistake rather than a silent empty)

The loader never validates `.field` names against `output_format`. Grepped `packages/workflows/src/loader.ts` for `output_format`, `declaredFields`, and `not-in-schema` — **zero** matches. `not-in-schema` is thrown by `resolveNodeOutputField` at execution time and fails the *consuming* node, exactly like the other reasons.

The mechanism being described is real and worth documenting — declaring a schema does turn an unknown field name into a hard error instead of a silent `''` (path 2). Only the *when* is wrong. Suggested:

> `…(an undeclared field fails the consuming node with a named error rather than resolving to a silent empty)`

This matters more than a typo because it's a reference page telling authors when they'll find out about a mistake, and the answer is "mid-run", not "at load".

---

### Suggestions (2)

**S1 — the edge-case analysis skipped bash/script nodes, and there's news in both directions.**

The PR's resume analysis (an excellent section) reasons about AI nodes and `workflow:` nodes. But `output_format` sits on `dagNodeBaseSchema` (`schemas/dag-node.ts:180`), so **bash and script nodes can declare it too** — and they're the only nodes subject to `formatPersistedBashOutput`'s 32 KiB cap (`dag-executor.ts:2506`, called once at `:2667`).

Three facts combine here:
- Fresh bash completion returns `{ state: 'completed', output }` with **no** `declaredFields` (`dag-executor.ts:2702`).
- Resume re-derives `declaredFields` from the YAML for **every** node type, with no type filter (`dag-executor.ts:7173`).
- Persisted bash output is truncated at 32 KiB with a `… [truncated; original output was N bytes]` marker appended.

**The good half — an unclaimed win.** A bash node with `output_format` emitting non-JSON previously threw on a fresh run (schemaless path 3) but resolved to `''` on resume (declared path 1, because resume grants it `declaredFields`). This PR makes both throw. That's a fresh-vs-resume inconsistency fixed for free, and it strengthens the PR's own argument — worth claiming.

**The caveat — a new failure with misleading advice.** A bash node with `output_format` emitting **valid JSON over 32 KiB**: fresh run parses the full in-memory output fine; a resumed run reads the truncated preview, fails to parse, and now throws. Reproduced directly:

```
THREW reason: unparseable
MESSAGE: '$gen.output.verdict' references field 'verdict', but node 'gen's output is not a
         JSON object, so the field cannot be read. Emit JSON containing 'verdict', or
         reference '$gen.output' (whole text) instead.
```

The node *did* emit JSON containing `verdict`. Archon truncated it. "Emit JSON containing 'verdict'" sends the author to fix a producer that was already correct.

Pre-PR this same case returned `''` — silently wrong rather than loudly wrong — so this is fail-fast working as designed, not a regression in correctness. But the diagnostic is actively misleading on a path nothing else will explain. Cheapest fix: have the `unparseable` message detect the truncation marker in the output and say so instead. Narrow scenario; confusing enough to be worth the four lines.

**S2 — name the blast radius in the changelog.** This is a genuine behavior change: any existing workflow with a declared-schema producer emitting non-JSON, whose consumer reads `.field`, goes from silently-empty to failed. That's the intent, and it matches the documented contract — but `workflow:` sub-run nodes are the exposed surface (their `output_format` is never validated against the child), and an author relying on the old `''` will see a new failure on a workflow they didn't touch. Worth an explicit line in `CHANGELOG.md` rather than leaving it to be discovered.

---

### Strengths

- **The invariant is tested as an invariant.** `'declaring a schema is never quieter than declaring none'` asserts the two paths produce the *same* `reason`, rather than asserting each path's behavior separately. That's the shape that survives future edits to either branch — a regression in one path fails the symmetry test even if both branches individually still "work".
- **The scope guard is the right instinct.** `'still lenient for a missing key inside a parsed object'` pins the leniency that *should* survive, so a future over-correction toward strictness gets caught. Changing one behavior and pinning the adjacent one you deliberately didn't change is exactly right.
- **The resume analysis is careful and load-bearing.** Establishing that a completed AI node always persists JSON — via `nodeOutputText` being overwritten with `JSON.stringify(structuredOutput)` (`:2129-2132`) and schema-invalid AI output never completing (`:2172`) — is what makes this safe rather than hopeful. It just stops one node type short.
- **Minimal diff on a semantics change.** One line of behavior (`return { kind: 'empty' }` → `throw`), with the reasoning recorded at both the resolution table and the call site.
- No new error reason, so nothing downstream goes stale.

---

*Reviewed by Claude — single-pass, all claims re-run locally against `fix/2456-output-format-inversion`.*

PR #2461, opened 2026-08-05
fix(workflows): resolve the include-expander logger per call so test mocks can intercept it

### LINKED ISSUE #2458: loader.test.ts loses 3 log assertions whenever include-expander.test.ts shares the process — the lazy-logger cache does not survive mock.module
## The behaviour

`packages/workflows/src/loader.test.ts` passes alone (174 pass, 0 fail). Put `include-expander.test.ts` in the same `bun test` invocation and three of its log-assertion tests fail:

```
(fail) Workflow Loader > include nodes > should warn when an included block drops meaningful workflow-level fields
(fail) Workflow Loader > include nodes > should warn — with a safety callout — when a block drops mutates_checkout and sandbox
(fail) Workflow Loader > include nodes > should warn (not fail) when a block command file cannot be resolved for scanning
```

Reproduced on unmodified `dev` (8704a65f), bun 1.3.11:

```bash
cd packages/workflows
bun test src/loader.test.ts                               # 174 pass, 0 fail
bun test src/loader.test.ts src/include-expander.test.ts  # 205 pass, 3 fail
bun test src/include-expander.test.ts src/loader.test.ts  # 205 pass, 3 fail  ← order-independent
bun test src/loader.test.ts src/schemas.test.ts           # 277 pass, 0 fail  ← specific to the pair
```

## Not a CI failure — a trap

`packages/workflows/package.json` runs `loader.test.ts` in its own `bun test` invocation, and `include-expander.test.ts` in a batch with `validation-parser` / `schemas` / `command-validation`. So `bun run test` and CI never hit this.

The protection is an accident of the current batching arrangement, not a guard. What it costs is developer time: running the two files together is a completely natural thing to do while working on include expansion (I did exactly that, mid-investigation, and had to bisect to establish the three failures were pre-existing rather than mine). Three red tests, no signal about why.

## Likely mechanism

Two ingredients:

- `loader.test.ts:26-28` does `mock.module('@archon/paths', () => ({ createLogger: mock(() => mockLogger) }))`. Per CLAUDE.md this is process-wide and irreversible — `mock.restore()` does not undo it.
- `include-expander.ts:44-49` caches its logger at module level:

```ts
/** Lazy-initialized logger (deferred so test mocks can intercept createLogger). */
let cachedLog: ReturnType<typeof createLogger> | undefined;
function getLog(): ReturnType<typeof createLogger> {
  if (!cachedLog) cachedLog = createLogger('workflow.include-expander');
  return cachedLog;
}
```

Once anything in the process has warmed `cachedLog`, later `getLog()` calls return that instance regardless of what `mock.module` has since installed — so the expander's warns never reach the loader test's spy and the call-count assertions come up short.

Flagging the mechanism as **likely rather than proved**: the reproduction, the order-independence, and the pair-specificity are all verified, but I did not instrument which call warms the cache first. The lazy-init comment is worth reading either way — the deferral exists *specifically* so test mocks can intercept `createLogger`, and it does not achieve that once a second file shares the process.

## Suggested fix

Options, cheapest first:

1. Assert on behaviour rather than log-call counts in those three tests (they are checking that a warn fired, which the expander's return value does not expose — so this may not be available).
2. Have `include-expander.ts` resolve its logger per call rather than caching, or expose a test-only cache reset. Same pattern exists in `loader.ts:51-56`, so whatever is chosen should apply to both.
3. Document the constraint in `packages/workflows/package.json` beside the batch list, so the batching becomes intentional rather than incidental.

Option 3 alone is enough to stop the trap costing time; 1 or 2 actually fixes it.

## Provenance

Found while planning #2453, whose phase-1 and phase-2 tasks add tests to **both** of these files — so this stops being merely latent for anyone working on that.

Related: #2453, #2186 and #2240 (the wider `mock.module` isolation class), and the mock-isolation rules in CLAUDE.md.

Wirasm:
## Summary

- **Problem:** `include-expander.ts` caches its logger at module scope behind a comment that says the deferral exists *"so test mocks can intercept `createLogger`"*. The cache defeats exactly that — it only delivers for whichever mock is installed at the **first** call.
- **Why it matters:** three tests in `loader.test.ts` fail whenever it shares a `bun test` process with `include-expander.test.ts`, in either order. CI never saw it because `package.json` happened to run them in different batches — protection by accident of arrangement, not by a guard.
- **What changed:** resolve the logger per call; co-run the two files in one batch so CI enforces it.
- **What did NOT change:** the other **107** files carrying the same `cachedLog` pattern. See scope note below.

### Why it happens

`mock.module` is process-wide and irreversible (CLAUDE.md's mock-isolation rules). Once *anything* in the process warms `cachedLog`, a later `mock.module('@archon/paths')` can no longer intercept the expander's warns — so the assertions in those three tests look for a call that landed on the real logger and find nothing.

The failure is order-independent and specific to that pair: `loader.test.ts` + `schemas.test.ts` is clean.

### What it cost

Not CI. Developer time. The two files are the natural pair to run together while working on include expansion, and the failure gives no hint about its cause — it presents as three unrelated warn assertions going red. I hit it mid-investigation on #2453 and had to bisect against unmodified `dev` to establish the failures were pre-existing rather than mine.

### Before

```
Developer                       bun test process
─────────                       ────────────────
runs the two files
that belong together ────────▶  include-expander warms cachedLog
                                  → real logger, permanently
                                loader.test installs mock.module
                                  → too late, cache already resolved
                                expander warns ──▶ real logger (nowhere)
                                loader assertions look in mockLogger ──▶ empty
sees 3 red tests ◀────────────  no message explaining why
bisects against dev ─────────▶  "ah — pre-existing, not mine"
```

### After

```
Developer                       bun test process
─────────                       ────────────────
runs the two files ───────────▶ loader.test installs mock.module
                                expander resolves logger PER CALL
                                expander warns ──▶ mockLogger  [+]
                                loader assertions find the call
sees green ◀──────────────────
                                ...and CI now runs this exact pair,  [+]
                                so a regression goes red rather than
                                waiting to ambush the next developer
```

### Before

```
  @archon/paths                include-expander.ts            loader.test.ts
  ┌──────────────┐             ┌──────────────────┐           ┌─────────────────┐
  │ createLogger │────────────▶│ cachedLog (once) │           │ mock.module(    │
  └──────────────┘             │ getLog() → cache │           │  '@archon/paths')│
         ▲                     └────────┬─────────┘           └────────┬────────┘
         │                              │ warns                        │ asserts
         └── mock installed LATER ──x── ▼                              ▼
             (never reaches the      real logger              mockLogger.warn
              cached instance)                                   (empty)

  package.json batches: loader | ... | include-expander   ← separated, incidentally
```

### After

```
  @archon/paths                include-expander.ts            loader.test.ts
  ┌──────────────┐             ┌──────────────────┐           ┌─────────────────┐
  │ createLogger │◀════════════│ [~] getLog() →   │           │ mock.module(    │
  └──────────────┘   per call  │     createLogger │           │  '@archon/paths')│
         ▲                     └────────┬─────────┘           └────────┬────────┘
         │                              │ warns                        │ asserts
         └── mock intercepts ═══════════▼══════════════════════════════▼
                                   mockLogger.warn  ✓

  package.json batches: [~] loader + include-expander | ...   ← co-run, deliberately
```

**Connection inventory:**

| `include-expander.ts` | `@archon/paths` `createLogger` | **modified** | Per call instead of once per module |
| `loader.test.ts` mock | `include-expander.ts` warns | **new** | Previously severed by the cache |
| `package.json` loader batch | `include-expander.test.ts` | **new** | Co-run so CI enforces the fix |
| `package.json` validation-parser batch | `include-expander.test.ts` | **removed** | Moved, not duplicated |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `workflows`
- Module: `workflows:include-expander`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2458
- Related #2453 (found while planning it — that work adds tests to both files, so this stops being latent), #2186 / #2240 (the wider `mock.module` isolation class)

```bash
```

Grepped rather than trusting the exit code:

```
$ grep -oE "[0-9]+ fail" /tmp/validate-2458.log | sort | uniq -c
 132   0 fail
$ grep -cE "[0-9]+ pass" /tmp/validate-2458.log
 132
```

**Both affected batches verified independently**, since this moves a file between them:

```
$ bun test src/loader.test.ts src/include-expander.test.ts          →  208 pass, 0 fail
$ bun test src/validation-parser.test.ts src/schemas.test.ts \
           src/command-validation.test.ts                            →  191 pass, 0 fail
```

**The CI guard was verified by reverting the fix** on the new batch — a guard that has never been red proves nothing:

```
--- fix reverted ---
(fail) Workflow Loader > include nodes > should warn when an included block drops meaningful workflow-level fields
(fail) Workflow Loader > include nodes > should warn — with a safety callout — when a block drops mutates_checkout and sandbox
(fail) Workflow Loader > include nodes > should warn (not fail) when a block command file cannot be resolved for scanning
 205 pass, 3 fail
--- fix restored ---
 208 pass, 0 fail
```

Exactly the three original failures, so the batch change genuinely covers the defect rather than merely coexisting with the fix.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? **Yes** — logging behaviour is identical in production; only *when* the logger instance is resolved changes.
- Config/env changes? **No**

- **Verified scenarios:** the pair green in both orders (208 pass); each affected batch green independently; the revert-the-fix check above.
- **Edge cases checked:** the mechanism was **proved**, not assumed. I originally reported it as a *likely* cause without instrumenting which call warms the cache — removing the cache fixing it in both orders settles it. Also confirmed `createLogger` is `rootLogger.child({ module })` (`packages/paths/src/logger.ts:101-103`), so per-call resolution is one child allocation, and both `getLog()` sites are warn-only discovery paths — the first fires at most once per include node, the second once per *unresolved command node* (it sits inside the per-child-node loop). Neither is a hot loop. *(Frequency note corrected post-review; see the follow-up commit.)*
- **What was not verified:** no measurement of logging throughput. Given both call sites are per-include-node at discovery time, a benchmark would be measuring noise.

- **Affected subsystems:** `@archon/workflows` include expansion (logging only) and the package's test batching.
- **Potential unintended effects:** the loader batch is now slightly larger. Both halves were run independently and are green.
- **Guardrails:** the co-run batch is itself the guardrail — a regression surfaces as the same three red tests, now in CI instead of only on a developer's machine.

## Rollback Plan (required)

- **Fast rollback:** `git revert <sha>` — two files, no production behaviour change.
- **Feature flags:** none.
- **Observable failure symptoms:** the three `loader.test.ts` include-warn assertions going red.

## Risks and Mitigations

- **Risk:** the same trap exists in the other 107 files carrying `cachedLog`.
  - **Mitigation:** deliberately out of scope. Only this file has a **demonstrated** failure; a blanket change would be a 108-file patch with no test proving the others were ever broken, which is precisely the mega-patch that blocks safe rollback. The comment in `include-expander.ts` now records the mechanism so the next person who hits it elsewhere recognises it immediately.
- **Risk:** co-running the two files surfaces some *other* latent mock conflict later.
  - **Mitigation:** both were run together repeatedly here (208 pass, 0 fail, both orders). If a future conflict appears, the batch split is a one-line revert.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow logging consistency by ensuring each log request uses the current logger configuration.

* **Tests**
  * Updated test coverage and execution to include include-expansion behavior in the standard test run.


Wirasm:
## Review: single-pass check

**Verdict: READY TO MERGE** — no changes requested. 0 critical, 0 important, 1 nitpick.

### Verified rather than assumed

| Check | Result |
|---|---|
| Co-run batch `loader.test.ts + include-expander.test.ts` | **208 pass / 0 fail** — matches the PR claim exactly |
| Batch that lost `include-expander.test.ts` | **191 pass / 0 fail** — matches the PR claim exactly |
| **Revert the fix**, re-run the co-run batch | **205 pass / 3 fail** — precisely the three tests named in the PR body |
| `package.json` edit is a true move | Confirmed — removed from the validation-parser batch, added to the loader batch; not duplicated, not dropped |
| `createLogger` cost claim | Confirmed — `rootLogger.child({ module })` (`packages/paths/src/logger.ts:101-103`), one child allocation |

The revert is the check that matters, and it's clean. Restoring `cachedLog` reproduces exactly:

```
(fail) Workflow Loader > include nodes > should warn when an included block drops meaningful workflow-level fields
(fail) Workflow Loader > include nodes > should warn — with a safety callout — when a block drops mutates_checkout and sandbox
(fail) Workflow Loader > include nodes > should warn (not fail) when a block command file cannot be resolved for scanning
 205 pass, 3 fail
```

So the new batch genuinely covers the defect rather than merely coexisting with the fix.

### The risk I went looking for isn't there

Merging two files into one `bun test` process is exactly the setup CLAUDE.md's mock-isolation rules warn about — two files `mock.module`-ing the same path with different factories. Checked: `include-expander.test.ts` calls `mock.module` **zero** times; only `loader.test.ts` mocks `@archon/paths` (`:26`). No conflicting factory, so the batch merge is safe on that axis.

### Nitpick (1)

Both the PR body and the new code comment state that the two call sites are *"reached once per include node at discovery, never in a hot loop."* That's accurate for the first site (`include-expander.ts:299`, guarded by an early `if (droppedFields.length === 0) return`). The second (`:339`) sits inside `for (const cn of child.nodes)` and can fire once per **unresolved command node**, not once per include node.

The conclusion is unaffected — it's a warn-only error path at discovery time, so "never in a hot loop" still holds — but the phrasing is looser than the code. Worth a word tweak if you touch the file again; not worth a push on its own.

### Something the PR undersells

`setLogLevel` is live in the CLI (`packages/cli/src/cli.ts:378-382`, driving `--quiet` / `--verbose`), and per its own docstring it *"only affects child loggers created after this call."* The module-scope cache therefore made include-expander's effective log level depend on **when** it was first warmed. Per-call resolution removes that latent ordering dependency.

Not a live bug today — discovery runs well after CLI startup, so the cache was warmed with the right level anyway — but it's a genuine robustness gain, not merely a test-only fix. The PR frames this purely as a test-mock issue; it's slightly better than that.

### On the deliberate scope limit

Agreed with leaving the other 107 `cachedLog` files alone. This is the only one with a **demonstrated** failure, and a 108-file patch with no test proving the rest were ever broken is precisely the mega-patch that blocks safe rollback. The comment left behind in `include-expander.ts` records the mechanism well enough that the next person to hit it elsewhere will recognise it.

---

*Reviewed by Claude — single-pass, all claims re-run locally against `fix/2458-logger-cache-test-trap`.*

PR #2467, opened 2026-08-05
feat: add load-time inputs for include nodes

### LINKED ISSUE #2466: feat(workflows): with: parameters on include: nodes — $INPUTS as a load-time macro (signature phase 1)
Phase 1 of #2453. Ships the **`include:` half only** — #2214 remains the `workflow:` half and stays deferred.

## Read this first — the honest boundary

**A `command:` node's body cannot be parameterised by this mechanism, and that is a bigger hole than it sounds.**

`include:` flattens at **load** time. A `command:` node's body is read from disk at **execution** time (`loadCommandPrompt`, invoked from `dag-executor.ts`), so a load-time macro cannot rewrite it.

`archon-review-block.yaml` — the flagship block, included by `archon-issue-review-full`, `archon-idea-to-pr` and `archon-plan-to-pr` — is **9 `command:` nodes and 1 `bash:` node**. So this reaches **one node in ten** there, and the flagship block stays un-parameterised until phase 3's runtime arg map.

This is not a defect in the approach; it is the boundary of a load-time mechanism. Phase 1 ships anyway because it is small, provably correct, and unblocks #2450. **Task 6 makes the boundary fail loudly, and the PR description must state it near the top** — an author should learn this from the PR, not from `$INPUTS.scope` silently reaching the model as literal text.

## What

```yaml
- id: review
  include: archon-review-block
  with:
    scope: packages/core        # literal
    plan: $plan.output          # a ref — deliberately left unresolved
  depends_on: [plan]
```

`$INPUTS.<name>` inside the included block's node bodies is replaced **verbatim** with the caller's value at expansion time. When the value is a `$node.output` ref it *stays* a ref, and the executor's existing runtime substitution resolves it later. **Zero new runtime machinery**, and validation comes free from the flattened re-validation the expander already runs.

## Tasks

Anchors are symbols and quoted strings, not line numbers — the file moved ~12 lines during the merges of 5 August and will move again.

**1 · `schemas/dag-node.ts` — accept `with:` on include nodes.** In `superRefine`, remove the issue whose message reads `"'with:' input mapping is not yet supported on include nodes (Phase 1). Remove it."` Replace with shape validation: `with` must be a plain object (not null, not array); every value a `string`; every key matching `/^[a-zA-Z_][a-zA-Z0-9_-]*$/` so it can appear as `$INPUTS.<key>`. One `ctx.addIssue` per violation, `path: ['with']`, message naming the fix.

**Leave the `workflow:` rejection untouched** — the one reading `"'with:' named-parameter mapping is not yet supported on workflow nodes (slice 2). Use 'input:' instead."` That is lifted in phase 2, not here.

⚠️ Do **not** widen the flat `dagNodeSchema`'s `with: z.unknown().optional()` to a typed schema. That schema is shared by all node modes and `with` must stay strippable on non-include nodes. Type it on `includeNodeSchema`; validate the flat one in `superRefine`.

**2 · `schemas/dag-node.ts` — carry `with` on `IncludeNode`.** `includeNodeSchema` gains `with: z.record(z.string(), z.string()).optional()`. The transform's include branch spreads it conditionally, mirroring the `workflow:` branch's idiom for optional fields.

⚠️ `z.record` takes **two** args in zod v4 — the single-arg form was dropped. The include branch's comment currently claims the node carries "ONLY the structural graph fields plus the target name"; that is now false — update it.

**3 · `schemas/dag-node.ts` — reserve `INPUTS` as a node id.** In `superRefine`, right after the empty-id check, reject `id === 'INPUTS'` with `path: ['id']`.

This is **not cosmetic — it is proved.** With a block node literally named `INPUTS`, the expander's `WHEN_REF_PATTERN` matches `$INPUTS` as a `when:` shorthand node ref and rewrites `when: "$INPUTS.plan == 'x'"` into `"$blkNode__INPUTS.plan == 'x'"`, silently pointing the gate at the wrong node. Reserving the id is the cheapest closure.

**4 · `include-expander.ts` — add `applyInputsMacro`.** Above `interface ExpandedInclude`: `const INPUTS_REF = /\$INPUTS\.([a-zA-Z_][a-zA-Z0-9_-]*)/g;` and `function applyInputsMacro(node, args, missing): void`, replacing each `$INPUTS.<name>` with `args[name]` verbatim; an unsupplied name goes into `missing` and is left in place. It must walk **exactly** the field set `rewriteNodeOutputRefs` walks, including the `loop_group.nodes` recursion — copy that branch structure field-for-field.

⚠️ Unlike `rewriteNodeOutputRefs`, this needs **no** prose/code/`when:` distinction. `$INPUTS.x` inside a markdown fence is still a parameter the author wrote; `$INPUTS` has no meaning except as a parameter, so there is no documentation-versus-live-ref ambiguity. One substitution function for every surface — **say so in the doc comment** or a reviewer will file it as a bug.

⚠️ The doc comment must also state that the substituted value **may itself be a `$node.output` ref left deliberately unresolved**. That is the entire mechanism and it reads like a bug to anyone who does not know.

**5 · `include-expander.ts` — call the macro in `inlineInclude`.** Declare `const missingInputs = new Set<string>()` beside `parentDeps`. Inside the `childNodes.map` clone loop, call `applyInputsMacro(clone, includeNode.with ?? {}, missingInputs)` **immediately after `rewriteNodeOutputRefs(clone, rename)` and before `clone.id = prefix + cn.id`.** Before returning, if `missingInputs.size > 0`, throw `IncludeExpansionError` naming the include node, the block, and the sorted missing names.

⚠️ **The ordering is load-bearing and must carry a comment saying why.** Applying the macro *after* the rename is what stops an injected `$parentNode.output` from being captured by the block's namespace. **Proved:** with a parent node `gather` and a block node also named `gather`, `with: {plan: $gather.output}` yields `$gather.output` (the parent's) while the block's own ref stays `$review__gather.output`. Reverse the order and the parent ref silently binds to the block's node.

⚠️ Do **not** report keys the caller passed that the block never uses. Extra keys are a phase-2 concern once `inputs:` exists; with no declaration to check against, warning on them is noise.

**6 · `include-expander.ts` — reject `$INPUTS` in a block's command bodies.** In `scanBlockCommandRefs`, after the existing sibling-id loop, test the fence-stripped content against `INPUTS_REF`. On a hit, throw `IncludeExpansionError` naming the include node, the block, the command file, and the parameter, and telling the author to inline the prompt.

⚠️ `scanBlockCommandRefs` is skipped entirely when `commandContents` is not supplied, so this check is **best-effort by construction** — it fires in real discovery, not in every unit test. State that next to the message so nobody mistakes it for a guarantee.

⚠️ The function already WARNs (`include.command_file_unresolved_for_ref_scan`) when a command file cannot be resolved, and an unresolvable file means `$INPUTS` in it goes undetected. **Do not upgrade that warn to an error** — it would break blocks whose commands live outside the scanned set.

**7 · Tests.** In `include-expander.test.ts`: (a) a ref value survives while the block's internal ref stays namespaced and a literal is inlined; (b) a dangling ref is rejected by the flattened re-validation; (c) a missing key is rejected, naming the block; (d) two callers namespace independently with different args; (e) a parent/block id collision resolves to the **parent**; (f) a nested include forwards `$INPUTS` two levels. Plus `$INPUTS` in a `when:`, `$INPUTS` inside a fenced code block still substituted, and a node id of `INPUTS` rejected.

⚠️ In `schemas.test.ts` the existing test `"include with 'with:' is rejected (not yet supported)"` must be **replaced, not deleted** — turn it into the shape-validation test (non-object rejected, non-string value rejected, bad key rejected).

⚠️ These files are pure — no `mock.module` — so they safely share a `bun test` batch. Keep them that way; see CLAUDE.md's mock-isolation rules.

**8 · Docs.** The authoring guide gains a `with:`/`$INPUTS` section for `include:` stating: values are literals or `$node.output` refs, never expressions; refs resolve at run time through the normal substitution; **`command:` bodies are not parameterisable** and will fail the load. In `workflow-language-constitution.md`, update the case-law row *"`with:` include parameters carrying expressions | ⚠️ constrained"* to record that the data-only mapping is now **admitted and shipped for `include:`**, resolved at load time, expressions still rejected.

⚠️ Do not claim the include and sub-run halves behave identically — the sub-run half does not exist until phase 2. Write the include half in the present tense and mention the sub-run half as not yet shipped.

## Constraints that must survive

- **Undeclared includes keep working byte-for-byte.** Every existing workflow has no `with:`; expansion must be unchanged for them.
- **The `#2459` parity ratchet stays green.** If a workflow-level schema field is added it fails until a fixture exists — that is the guard working. Add the fixture; do not weaken the test.
- **Constitution:** this coordinates, it does not compute. Data-only mapping, no expressions. CLAUDE.md requires citing the constitution in `feat(workflows)` PRs touching the YAML surface — this is one.

## Not in scope

`inputs:`/`returns:` declarations on the callee (phase 2) · `with:` on `workflow:` nodes (#2214, phase 2) · the runtime arg map that makes `command:` bodies parameterisable (phase 3) · any expression capability in `with:` values.

## Acceptance

- `with:` accepted on `include:`, still rejected on every other node type including `workflow:`
- A literal is inlined; a `$node.output` ref survives expansion and resolves at run time
- An unsupplied `$INPUTS.<name>` fails the load, naming the include node, the block and the parameter
- `$INPUTS.` in a resolved command body fails the load with a message pointing at the file
- A node whose id is `INPUTS` is rejected
- A parent/block node-id collision resolves the injected ref to the **parent**
- `bun run validate` green — grep the log for `0 fail` rather than trusting the wrapper's exit code, which has masked a real failure before
- PR description carries the command-body caveat near the top

Part of #2453. Unblocks #2450. Related: #2214 (the `workflow:` half), #2449, #2123 (the consumer).

Wirasm:
## Summary

- Problem: included workflow blocks could not accept caller-specific values; `include.with` was rejected at load time.
- Why it matters: authors had to duplicate inline workflow blocks instead of reusing a static sub-DAG with literals or prior node-output references.
- What changed: `include.with` now accepts an identifier-keyed string map and expands `$INPUTS.<name>` in included inline node fields while flattening; unsupported `$INPUTS.` inside command files fails clearly.
- What did **not** change (scope boundary): no `workflow.with`, input/return declarations, runtime `$INPUTS` delivery, command-file parameterization, database/API/UI, or dependency changes.

### Before

```
Workflow author             Workflow loader              Included block
───────────────             ───────────────              ──────────────
writes include.with ─────▶ rejects `with:`
duplicates an inline block to vary values
```

### After

```
Workflow author             Workflow loader              Included block
───────────────             ───────────────              ──────────────
writes include.with ─────▶ [validates named strings]
                            [namespaces child references]
                            [replaces $INPUTS.<name>] ──▶ receives literal or
                                                          preserved $node.output
                            validates flattened DAG
```

### Before

```
workflow YAML --> dag-node schema --> include-expander --> flattened workflow
                                      |
                                      +--> command content scan (sibling refs)
flattened workflow --> loader DAG validation --> executor
```

### After

```
workflow YAML --> [~ dag-node schema: include.with] --> [~ include-expander]
                                                         |=== applies $INPUTS macros
                                                         |=== rejects command-file macros
                                                         +--> flattened workflow
flattened workflow --> loader DAG validation --> executor
docs --> workflow authors
```

| Workflow YAML | `dag-node` schema | **modified** | Validates and retains `include.with`; reserves `INPUTS` as a node ID. |
| `dag-node` schema | `include-expander` | **modified** | Typed include mappings are consumed during flattening. |
| `include-expander` | flattened workflow | **modified** | Rewrites `$INPUTS.<key>` after child-reference namespacing. |
| `include-expander` | resolved command content | **modified** | Rejects `$INPUTS.` because command bodies load at execution time. |
| flattened workflow | loader DAG validation | unchanged | Existing validation catches injected dangling output references. |
| documentation | workflow authors | **modified** | Describes supported include-only mappings and command boundary. |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: M`
- Scope: `workflows`, `docs`, `tests`
- Module: `workflows:include-expander`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Fixes #2466
- Related #2453
- Depends on # none
- Supersedes # none

Commands and result summary:

```bash
bun run type-check        # pass
bun run lint --max-warnings 0  # pass, 0 warnings
bun run format:check      # pass
bun run test              # pass, 0 failures
bun run build             # pass
bun run validate          # pass
```

- Evidence provided (test/log/trace/screenshot): focused schema/expander tests (150 passing), loader tests (198 passing), docs build (85 pages), a manual CLI positive case plus three negative cases, and an end-to-end workflow execution producing `scope=THE-PLAN`.

## Security Impact (required)

## Compatibility / Migration

- Backward compatible? `Yes`
- Config/env changes? `No`

- Verified scenarios: literal include inputs and caller `$node.output` references expand correctly; the isolated end-to-end workflow printed `scope=THE-PLAN`.
- Edge cases checked: missing input names, reserved `INPUTS` node ID, invalid mapping values/keys, namespace-safe child references, dangling injected output references, and `$INPUTS.` inside resolved command files.
- What was not verified: later workflow-signature phases (declarations, sub-run inputs, runtime delivery, and command-body parameterization) are intentionally out of scope.

- Affected subsystems/workflows: `@archon/workflows` schema validation, include flattening, workflow discovery command checks, and workflow authoring documentation.
- Potential unintended effects: malformed include mappings now fail at load time; macro replacement could alter included inline strings that deliberately contain matching `$INPUTS.<name>` tokens.
- Guardrails/monitoring for early detection: contextual schema errors, explicit unresolved-macro errors, existing flattened-DAG validation, and explicit command-body rejection prevent silent fallback.

## Rollback Plan (required)

- Fast rollback command/path: revert commit `b091f6fb` from `dev` if the feature must be removed.
- Observable failure symptoms: workflow loading reports an include mapping, unresolved macro, command-file macro, or dangling output-reference error.

## Risks and Mitigations

- Risk: Include namespacing could accidentally bind an input output-reference to a renamed child node.
  - Mitigation: inputs are applied after internal child-reference rewriting and before child IDs are assigned; tests cover parent and child reference behavior.
- Risk: `$INPUTS.` in command files could reach an agent unexpanded.
  - Mitigation: resolved command content is scanned and rejected with remediation guidance.

## Summary by CodeRabbit

* **New Features**
  * Include workflows now support `with:` string mappings to pass `$INPUTS.<name>` into inline prompts, including code fences and inline code.
  * Substitution applies across supported node fields, while runtime output references continue to resolve correctly.
* **Bug Fixes**
  * Missing or invalid `$INPUTS` values now fail workflow loading with clear error details.
  * `$INPUTS` references in command files, including `loop.command`, are rejected; unresolved command files produce warnings.
* **Documentation**
  * Updated workflow documentation to reflect supported include-input behavior and limitations.


Wirasm:
# Consolidated Review: PR #2467

**Date**: 2026-08-05T09:30:00Z
**Total Findings**: 5

---

## Executive Summary

This PR cleanly introduces load-time `include.with` expansion without adding executor semantics. The schema, namespace-safe reference ordering, and focused tests are generally strong. However, an included `loop.command` is not covered by the command-body safety scan, so it can carry a literal `$INPUTS.*` token to runtime despite the feature's stated fail-fast boundary. There is also a related unresolved-command fallback, incomplete coverage for several supported inline surfaces, and two documentation claims that are narrower than the implemented string-only validation. The documentation-impact review found no broader docs, API, configuration, or migration changes needed.

**Auto-fix Candidates**: 1 HIGH issue can be auto-fixed
**Manual Review Needed**: 4 MEDIUM issues require decision

---

## Statistics

| Code Review | 0 | 1 | 0 | 0 | 1 |
| Error Handling | 0 | 0 | 1 | 0 | 1 |
| Test Coverage | 0 | 0 | 1 | 0 | 1 |
| Comment Quality | 0 | 0 | 2 | 0 | 2 |
| Docs Impact | 0 | 0 | 0 | 0 | 0 |
| **Total** | **0** | **1** | **4** | **0** | **5** |

---

## CRITICAL Issues (Must Fix)

None.

---

## HIGH Issues (Should Fix)

### Issue 1: Included `loop.command` files bypass the `$INPUTS` fail-fast boundary

**Source Agent**: code-review
**Location**: `packages/workflows/src/include-expander.ts:395`
**Category**: Bug / command-file validation

**Problem**:

`scanBlockCommandRefs()` only scans ordinary `command:` nodes. A loop's `loop.command` is also a file-backed prompt loaded at execution time, but it is neither pre-resolved by discovery nor scanned by the expander. An included loop command containing `$INPUTS.scope` therefore loads successfully and sends the unsupported literal to the agent at runtime.

**Recommended Fix**:

```typescript
function getFileBackedCommandName(node: DagNode): string | undefined {
  if ('command' in node && typeof node.command === 'string') return node.command;
  if (isLoopNode(node) && typeof node.loop.command === 'string') return node.loop.command;
  return undefined;
}
```

Use this narrow helper when discovery builds command contents and when the include expander performs its safety scan. Add a loader/discovery test for an included `loop.command` file containing `$INPUTS.scope`.

**Why High**:

This breaks the explicit Phase 1 boundary that command-file parameterization must fail during loading, rather than silently becoming runtime behavior.

---

### Issue 1: An unresolved included command file bypasses safety validation

**Source Agent**: error-handling
**Location**: `packages/workflows/src/include-expander.ts:398`

**Problem**:

When a command file cannot be resolved, the expander logs a warning and continues. It consequently cannot reject `$INPUTS.*` or unsafe pre-namespacing sibling references in that file. This is related to, but distinct from, the `loop.command` omission: even ordinary command nodes can bypass the existing scan when their content is unavailable.

**Options**:

| Fix Now | Throw `IncludeExpansionError` with include/block/command context when safety scanning cannot resolve the command. | LOW | Unsupported tokens or stale refs can reach execution silently. |
| Create Issue | Preserve compatibility and add a structured load diagnostic later. | LOW | Same behavior remains until follow-up ships. |
| Skip | Keep the warning-only fallback. | NONE | Workflow authors receive no actionable failure. |

**Recommendation**: Fix now. A safety scan that cannot inspect the input is ambiguous rather than safe, and fail-fast aligns with the newly introduced command-body restriction.

```typescript
throw new IncludeExpansionError(
  `Node '${includeNode.id}': command file '${command}.md' in included block ` +
    `'${child.name}' could not be resolved for include safety validation. ` +
    'Make the command available, or inline the prompt.'
);
```

### Issue 2: Supported substitution surfaces lack behavioral coverage

**Source Agent**: test-coverage
**Location**: `packages/workflows/src/include-expander.ts:212-231`

**Problem**:

New tests exercise `prompt`, `when`, and fenced text, but not the `loop`, recursive `loop_group`, `approval`, `bash`, `script`, `workflow.input`, or `cancel` branches in `applyInputsMacro()`. A future change could leave `$INPUTS.*` unexpanded on one of these executable or structured surfaces without a test failure.

**Options**:

| Fix Now | Add one table-driven public-expander test covering every remaining supported surface. | MEDIUM | Regressions can reach scripts, loop controls, cancellation text, or child inputs. |
| Create Issue | Add focused tests in a follow-up. | LOW | Coverage hole remains temporarily. |
| Skip | Rely on current implementation and prompt-focused tests. | NONE | Broad feature contract is not guarded. |

**Recommendation**: Fix now. A compact table-driven test is proportionate and keeps the surface-wide contract regression-safe.

### Issue 3: Authoring guide promises an unenforced `with:` value grammar

**Source Agent**: comment-quality
**Location**: `packages/docs-web/src/content/docs/guides/authoring-workflows.md:892`

**Problem**:

The guide says values are only string literals or `$node.output` references and that expressions are never allowed. The schema validates identifier keys and that values are strings, then `applyInputsMacro()` inserts the string verbatim. No grammar check enforces the narrower documentation claim.

**Options**:

| Fix Now | Document the enforced rule: values must be strings and are substituted verbatim at load time. | LOW | Authors are promised validation the product does not provide. |
| Create Issue | Add a deliberate restricted-value grammar in a later language change. | MEDIUM | Documentation remains inaccurate until fixed. |
| Skip | Keep the literal/output-reference claim. | NONE | Normative guidance misstates behavior. |

**Recommendation**: Fix the documentation now; consider a stricter grammar only as a separately specified feature.

### Issue 4: Workflow-language constitution repeats the unsupported restriction

**Source Agent**: comment-quality
**Location**: `packages/docs-web/src/content/docs/reference/workflow-language-constitution.md:59`

**Problem**:

The normative constitution repeats that only literals or `$node.output` references are accepted and expressions are rejected, though the shipped validator accepts any string. This finding conflicts with the docs-impact agent's general conclusion that the documentation matches implementation; the concrete schema and macro evidence supports this more specific finding.

**Options**:

| Fix Now | State that identifier-keyed string values are substituted at load time; retain the enforced `workflow.with` boundary. | LOW | The language reference remains misleading. |
| Create Issue | Implement the claimed grammar and tests in a dedicated PR. | MEDIUM | Current reference remains inaccurate. |
| Skip | Leave the expression-rejection claim. | NONE | Reviewers and authors infer nonexistent guardrails. |

**Recommendation**: Fix the wording now, consistently with the authoring guide.

---

None.

---

## Positive Observations

- Child-internal output references are namespaced before caller values are inserted, preserving parent reference scope under node-ID collisions.
- Mapping validation is strongly typed and keeps Phase 2 extra-key validation intentionally deferred.
- Existing tests cover empty and repeated values, missing mappings, nested forwarding, fenced text, output-reference collisions, dangling injected references, and ordinary command-file rejection.
- Error handling is mostly explicit and contextual: missing macros are aggregated, unsafe `when:` composition is rejected, and unexpected exceptions are rethrown.
- The implementation comments accurately describe ordering and the load-time/runtime command-file boundary.
- No API, database, configuration, command, or broader architectural documentation changes are required.

---

## Suggested Follow-up Issues

| "Define and enforce a restricted `include.with` value grammar" | P2 | MEDIUM issues 3-4, if arbitrary string values are not the intended permanent contract. |

---

## Next Steps

1. **Auto-fix step** should address the 1 HIGH command-surface gap.
2. **Review** the four MEDIUM issues and decide whether to fix now or defer the optional grammar enforcement work.
3. **Merge** after command-file validation and documentation are aligned with the intended Phase 1 contract.

---

## Agent Artifacts

| Code Review | `code-review-findings.md` | 1 |
| Error Handling | `error-handling-findings.md` | 1 |
| Comment Quality | `comment-quality-findings.md` | 2 |
| Docs Impact | `docs-impact-findings.md` | 0 |

---

## Metadata

- **Synthesized**: 2026-08-05T09:30:00Z
- **Artifact**: `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/166bc0584cbae0a19db002d663d28136/review/consolidated-review.md`

Wirasm:
# Fix Report: PR #2467

**Date**: 2026-08-05T09:19:42Z
**Status**: COMPLETE
**Branch**: archon/task-archon-fix-github-issue-experimental-1785918317060
**Commit**: c87c1044

---

## Summary

All five review findings were fixed. Included command-backed loops now undergo the same load-time safety validation as ordinary command nodes, unresolved included command files fail fast, and tests and documentation cover the complete shipped behavior.

---

## Fixes Applied

| HIGH | Included `loop.command` files bypass `$INPUTS` validation | `packages/workflows/src/include-expander.ts` | Centralized file-backed command extraction and scan both ordinary command nodes and deferred loop commands. |
| MEDIUM | Unresolved command file bypasses safety validation | `packages/workflows/src/include-expander.ts` | Converted the warning-only fallback to a contextual include-expansion error. |
| MEDIUM | Supported input surfaces lacked coverage | `packages/workflows/src/include-expander.test.ts` | Added coverage for bash, script, loop, loop group recursion, approval, cancel, and workflow input. |
| MEDIUM | Authoring guide overstated `with:` value grammar | `packages/docs-web/src/content/docs/guides/authoring-workflows.md` | Documented the enforced identifier-keyed string-map behavior and command-file restriction accurately. |
| MEDIUM | Constitution repeated unsupported grammar restriction | `packages/docs-web/src/content/docs/reference/workflow-language-constitution.md` | Replaced the narrower claim with the actual load-time string substitution behavior. |

---

## Tests Added

| File | Test Cases |
|------|------------|
| `packages/workflows/src/include-expander.test.ts` | All supported inline substitution surfaces; included `loop.command` containing `$INPUTS`; unresolved command-file failure. |
| `packages/workflows/src/loader.test.ts` | Discovery resolves and rejects unsafe included `loop.command` content. |

---

## Docs Updated

| File | Changes |
|------|---------|
| `packages/docs-web/src/content/docs/guides/authoring-workflows.md` | Corrected the accepted mapping grammar and documented `loop.command` validation. |
| `packages/docs-web/src/content/docs/reference/workflow-language-constitution.md` | Corrected the `include.with` language guarantee. |

---

## Skipped Findings

*(none)*

---

## Blocked (Could Not Fix)

*(none)*

---

## Suggested Follow-up Issues

*(none)*

---

## Validation

| Check | Status |
|-------|--------|
| Type check | ✅ |
| Lint | ✅ |
| Format check | ✅ |
| Focused workflow tests | ✅ 243 passed |
| Full test suite | ✅ |
| Full validation (`bun run validate`) | ✅ |

Wirasm (commented):
First review pass — one docs change before this lands. The implementation isn't reviewed yet; this is the example that jumped out on read-through.

packages/docs-web/src/content/docs/guides/authoring-workflows.md
```diff
@@ -863,12 +863,49 @@ written the nodes by hand. There is no separate child run.
...
+nodes:
+  - id: plan
+    prompt: Plan the requested change.
+
+  - id: review
+    include: reusable-review
+    depends_on: [plan]
+    with:
+      plan: $plan.output
+      base_branch: main
```
Wirasm, line 973: Change this to `base_branch: $BASE_BRANCH`.

Two reasons, and the second is the one that matters:

**1. `main` is the wrong branch to name in this repo.** `main` is the release branch here; everything targets `dev`. A reader who copies this example inherits the wrong default. (Unrelated but adjacent: a stale local `refs/remotes/origin/HEAD` on this machine was *also* pointing at `main`, so the word showed up twice from two unrelated causes — that one's fixed separately.)

**2. It teaches hardcoding a branch name in the same paragraph that introduces the feature.** Archon already provides `$BASE_BRANCH` precisely so authors don't do this. Using it here demonstrates something more interesting than a literal: a **workflow variable flowing through `with:` into a block**.

**Verified it works** — `applyInputsMacro` is a pure verbatim `text.replace`, and this file has zero references to `substituteWorkflowVariables` or `BASE_BRANCH`. So `$BASE_BRANCH` lands as literal text in the block's node body and resolves through normal runtime substitution, by exactly the mechanism the prose above already describes for `$node.output`. No code change needed — the doc line is the whole fix.

Worth a sentence in the surrounding prose too: a `with:` value can carry any variable the runtime substitutes, not just `$node.output`.

Wirasm:
## PR Review: #2467 — load-time inputs for `include:` nodes

**Recommendation: REQUEST CHANGES** — one confirmed silent-failure bug (one-line fix), one deliberate spec deviation that needs your decision.

Reviewed against the spec in #2466, which is unusually prescriptive (8 numbered tasks + explicit ⚠️ constraints + an Acceptance list). Six specialist agents (code, errors, tests, types, comments, docs) plus runnable repros for every correctness claim below. **Rebase: not needed** — branch is already on top of `dev` (0 behind / 2 ahead).

### Validation

| Check | Result |
|---|---|
| Type check | PASS |
| Lint | PASS (0 warnings) |
| Format | PASS |
| Tests | PASS — `0 fail` in every package; `packages/workflows` 1641 pass |
| Docs build | PASS (85 pages) |

---

## Critical

### C1 · `$INPUTS` in `workflow.fan_out.items` is silently dropped

`packages/workflows/src/include-expander.ts` — `applyInputsMacro`'s `isWorkflowNode` branch substitutes `node.input` but never `node.fan_out.items`. `rewriteNodeOutputRefs`, the function it is required to mirror, walks both:

```ts
// rewriteNodeOutputRefs — both surfaces
if (node.input !== undefined) node.input = code(node.input);
if (node.fan_out !== undefined) node.fan_out.items = code(node.fan_out.items);

// applyInputsMacro — fan_out.items missing
if (node.input !== undefined) node.input = substitute(node.input);
```

This violates Task 4 verbatim: *"It must walk **exactly** the field set `rewriteNodeOutputRefs` walks... copy that branch structure field-for-field."*

**Confirmed by repro**, not inference:

```
fan_out.items ACTUAL = "$INPUTS.itemsref"   errors: []
workflow.input ACTUAL = "scope=$gather.output"   (control — correct)
```

Both failure halves fire at once. The literal `$INPUTS.<name>` reaches the executor, **and** `missingInputs` never sees the name — so a caller who forgot to pass that input gets no load error either. That directly contradicts the PR's own Acceptance criterion: *"An unsupplied `$INPUTS.<name>` fails the load, naming the include node, the block and the parameter."*

There is no safety net. `validateDagStructure` does scan `fan_out.items` (`loader.ts:265`) but only for the `\$id\.output` grammar, so `$INPUTS.tasks` matches nothing. The guard fires only when the input is coincidentally named `output`:

```
input named 'tasks':  errors=0  (SILENT)
input named 'scope':  errors=0  (SILENT)
input named 'output': errors=1  Node 'inc__fan' references unknown node '$INPUTS.output'
```

**Fix** — one line in `applyInputsMacro`'s workflow branch:
```ts
if (node.fan_out !== undefined) node.fan_out.items = substitute(node.fan_out.items);
```
Plus a case in `substitutes inputs across every other supported inline node surface` — that test enumerates every surface *except* this one, so its title currently asserts a completeness it doesn't have.

---

## Important

### I1 · WARN → hard error on unresolvable command files (explicitly forbidden; needs your call)

`include-expander.ts` `scanBlockCommandRefs` now throws where it previously warned and continued. Task 6 says: *"**Do not upgrade that warn to an error** — it would break blocks whose commands live outside the scanned set."* It also breaks the must-survive constraint *"Undeclared includes keep working byte-for-byte."*

This was deliberate, not an oversight — commit `fbe9108f` lists "fail fast when included command files cannot be validated", and the loader test was renamed `should warn (not fail) …` → `should fail …` with its `mockLogger.warn` assertion deleted.

**Confirmed by repro** — a workflow with *no* `with:` and *no* `$INPUTS` anywhere is dropped entirely from discovery:
```
errors: [{ error: "…could not be resolved for include safety validation…" }]
workflow 'p' still discoverable? NO -- DROPPED
```

Blast radius (`resolveCommandContentForScan` returns `null`): `defaults.loadDefaultCommands: false`; a command name containing `/` (rejected by `isValidCommandName` before resolution); `cwd === null` global/non-repo discovery. Under `loadDefaultCommands: false` this drops `archon-issue-review-full`, `archon-plan-to-pr` and `archon-idea-to-pr` — all three include `archon-review-block`, whose 9 command nodes are bundled defaults. A stock install is unaffected today.

**This is a judgment call, not a clear-cut bug.** Fail-fast is a stated CLAUDE.md principle and the author defended it in-code; the issue author explicitly forbade it. Recommended split: keep the hard fail for a *resolved* file containing `$INPUTS.<name>` (that half is correct and valuable), revert the *unresolved-file* branch to WARN + `continue`. If you'd rather keep the strict behavior, it should be owned explicitly in the PR description and #2466 amended — it shouldn't land silently against a written constraint.

### I2 · `loop_group` body command files escape the scan on both sides

A `command:`/`loop.command` node nested in a `loop_group` body inside an included block is invisible to the safety scan — `resolveIncludeBlockCommandContents` (`workflow-discovery.ts`) iterates `workflow.nodes` top-level only, so its content never enters the map, and `scanBlockCommandRefs` iterates `child.nodes` top-level only. **Confirmed by repro**: `errors: []` where the identical command at top level is correctly rejected.

`applyInputsMacro` *does* recurse into `loop_group.nodes`; the scan does not. The top-level loop structure is pre-existing (not a regression from this PR), but the new `$INPUTS` check inherits the blind spot, and it makes a docs claim untrue — see S1.

### I3 · The "KEEP IN SYNC" comment now undercounts, and it already drifted

`rewriteNodeOutputRefs`'s block comment still reads *"KEEP IN SYNC (three ref-surface enumerations must agree)"*. `applyInputsMacro` is a fourth enumeration of the identical surface, and it has already drifted — that drift is exactly C1. Suggested: *"four ref-surface enumerations must agree: this rewrite, `applyInputsMacro` below, the loader's `validateDagStructure` scan, and the `substituteNodeOutputRefs` call sites in dag-executor.ts."*

### I4 · Misleading comment on the flat `dagNodeSchema.with` field

Current: *"Raw here because include/workflow modes validate and retain it differently."* Workflow mode never retains `with` in any form — it unconditionally rejects it as unsupported (phase 2). A maintainer could read this as "both branches keep it somehow." Suggested: *"Raw here because include mode validates its shape and retains it on the parsed node; workflow mode still rejects it outright as unsupported (phase 2). Other node modes strip it with the rest of their unsupported surface."*

---

## Suggestions

- **S1 · Docs overclaim.** `authoring-workflows.md` states unconditionally that loading fails if a command file "cannot be resolved for safety validation or contains `$INPUTS.<name>`". Only true for command-backed nodes at the block's top level (see I2). Scope the sentence or close the recursion gap.
- **S2 · Reserved `INPUTS` node id is undocumented** anywhere — a user who names a node `INPUTS` gets a load error with no doc explaining why.
- **S3 · `CLAUDE.md:453`** — the `include:` node-type description wasn't updated with `with:`/`$INPUTS`. The same line already documents `workflow:`'s "`with:` and `retry:` rejected" at comparable density.
- **S4 · Task 6's "best-effort" caveat** was required next to the message and isn't there. One line above the new `$INPUTS` throw would satisfy it.
- **S5 · Docs never say "never expressions"** explicitly, which Task 8 asked for. One clause closes it.
- **S6 · Duplicated identifier grammar** — `/^[a-zA-Z_][a-zA-Z0-9_-]*$/` in `dag-node.ts` and the capture group in `INPUTS_REF`. They agree today; a shared constant would prevent a key that validates but never substitutes.
- **S7 · No test** combines the unresolvable-command throw with `loadDefaultCommands: false` (low value — the branch itself is pre-existing).
- **S8 · `command-file.ts` is scope creep** beyond #2466 (adds `loop.command`). Correct on its own and closes a real gap, but it widens I1's blast radius to files that previously weren't scanned at all.

---

## Strengths

- **The load-bearing ordering is right and genuinely protected.** `rewriteNodeOutputRefs` → `applyInputsMacro` → id rename, with a comment explaining why. Hand-traced: reversing it would rename an injected `$gather.output` into `$review__gather.output`, and the collision test's exact-match assertion would fail. That test earns its keep.
- **Test quality is high.** Every Task 7 case and Acceptance item is covered with regression-catching assertions, not placeholders. The old `"with: is rejected"` test was genuinely *replaced* per the ⚠️, not deleted. Loader tests exercise the real discovery path with files on disk, not just unit-level calls.
- **Nested-include forwarding is correct by design** — memoized single expansion means an intermediate block's `$INPUTS` passthrough survives to be resolved by the outer caller.
- **`superRefine` → `transform` is safe.** Verified against zod 4.4.3: a non-fatal `ctx.addIssue()` short-circuits `safeParse` before the transform runs, so the `as Record<string, string>` cast is never observably reached.
- **`INPUTS_REF` `/g` statefulness is handled** — `String.replace` resets `lastIndex` per spec, and the `.exec()` site resets defensively.
- **Constitution treatment is careful and honest** — data-only mapping, present tense for the shipped `include:` half, explicit that `workflow.with` has not shipped. No overclaiming of parity with the unbuilt phase-2 half.
- **CLAUDE.md zod conventions respected** — two-arg `z.record`, no hand-rolled parallel interface, correct schema naming.

---

## Bottom line

C1 is a one-line fix plus a test case and should land before merge — it's a silent failure in exactly the mode the feature exists to prevent, and it's untested in both directions. I1 is a real decision for the maintainer rather than a defect to fix on autopilot: fail-fast is defensible, but it contradicts a written constraint and changes load behavior for `include:` users who never opted into this feature. I2–I4 are cheap. Everything else is polish.

The implementation is otherwise careful, well-tested, and faithful to a demanding spec.

Wirasm:
## Addendum — type-design pass + independent confirmation

The type-design and error-handling passes landed after the main review. Net effect: **no change to the recommendation or to C1/I1/I2** — all three were independently re-confirmed with live repros. Below is what's genuinely new, plus one correction to my own report.

### Correction to S6 (identifier-regex duplication) — downgrade

I flagged the duplicated `/^[a-zA-Z_][a-zA-Z0-9_-]*$/` between `dag-node.ts` and `INPUTS_REF` as worth a shared constant. That was framed misleadingly: the same character-class literal already appears independently at `loader.ts:109`, `dag-executor.ts:752`/`875`, `condition-evaluator.ts:133` and `validator.ts:687` with no shared constant anywhere. Hand-duplicating this pattern per call site is the codebase's prevailing convention, so this PR is consistent with existing practice rather than introducing drift risk. **Not a finding against this PR** — at most a separate codebase-wide cleanup, and not one I'd push for.

### New — S9 · The schema that types `with` and the schema that enforces it are different objects

`includeNodeSchema` is never `.parse()`'d anywhere in production — it exists solely so `IncludeNode = z.infer<typeof includeNodeSchema>` derives the right TS type (correctly following the z.infer-only rule). All runtime enforcement is the 35-line hand-rolled `superRefine` block. The two agree today by coincidence of authorship, not by construction.

Verified that zod v4 expresses the entire invariant natively in one line, reproducing every accept/reject case including the null-prototype allowance:

```ts
z.record(z.string().regex(/^[a-zA-Z_][a-zA-Z0-9_-]*$/), z.string())
```

Calling that inside `superRefine` and forwarding its issues would collapse type-derivation and runtime-check into one object while keeping the flat field `z.unknown()`. Worth noting this is exactly what #2466 Task 1 directed ("validate the flat one in superRefine", with hand-written messages), so the implementer followed instructions faithfully — this is a tightening opportunity, not a defect.

### New — S10 · `isCommandNode` is missing from the guard catalog

`'command' in node && typeof node.command === 'string'` inside `command-file.ts` is now the **third** independent copy of that idiom (`dag-executor.ts:3967`, `validator.ts:379`, `command-file.ts`). Every other node mode has an exported `is*Node` guard in `schemas/dag-node.ts`; `isCommandNode` is conspicuously the one absent. This PR completed the rule of three, so the bar is now met. Small, contained fix: add the guard and have `command-file.ts` call it.

Placement of `getFileBackedCommandName` in a new `command-file.ts` (rather than in `schemas/dag-node.ts`) is the right call — it derives a value spanning two node types for two consumers, so keeping it out of the foundational widely-imported schema module avoids coupling that module to command-resolution logic.

### Extends I4 · The comment's stated rationale is wrong on a second axis

Beyond the "validate and retain it differently" problem already filed, the claim that raw `z.unknown()` is needed so "other node modes strip it" doesn't hold either: `isolation` and `fan_out` in the same file are typed directly at the flat-schema level *and* rejected via `ctx.addIssue` on the wrong node type. So "typed at flat level + reject on wrong mode" is both achievable and the established local pattern. The real (unstated) justification is sound — `workflow.with`'s future shape is unknown, so committing the shared flat field to `Record<string, string>` now risks a breaking change in phase 2. Worth stating that reason instead, since it's the one that actually survives scrutiny.

### Checked clean — nothing to change

- **Prototype check** (`prototype !== Object.prototype && prototype !== null`), tested against `Bun.YAML.parse` (the actual parser, per `loader.ts:85`): bare timestamps parse as **strings** not `Date`; mappings never produce `Map`; arrays and `null` are already caught by the preceding clauses. So the clause is defensive redundancy against a future parser swap, not live logic. **No prototype-pollution vector** — `__proto__` arrives as an ordinary own key with `Object.prototype` intact, exactly as `JSON.parse` behaves. One harmless quirk: an input literally named `__proto__` validates and substitutes correctly. Nothing worth changing.
- **Zod transform/cast safety — an agent disagreement I settled by execution.** One pass claimed the `.transform()` runs on invalid data (safety coming only from callers gating on `.success`); another claimed it short-circuits. Ran it against the project's actual zod: `safeParse success: false, transform ran: false`. The transform does **not** execute after a non-fatal `ctx.addIssue()`, so the `as Record<string, string>` cast is genuinely unreachable — the stronger of the two guarantees, as originally reported.

### Independent re-confirmation of the main findings

C1 (`fan_out.items`), I1 (WARN→throw), and I2 (`loop_group` body scan hole) were each reproduced independently on a second pass with live `discoverWorkflows()` runs rather than code reading alone. On I1 specifically: all 9 of `archon-review-block.yaml`'s command files exist **only** under `.archon/commands/defaults/` — the non-default `.archon/commands/` set is entirely disjoint — so the `loadDefaultCommands: false` blast radius on the three including workflows is confirmed as stated.

### Type-design ratings

| Dimension | Rating | Driver |
|---|---|---|
| Encapsulation | 6/10 | Type and enforcement live in two objects that agree only by authorship |
| Invariant expression | 7/10 | Fully correct with good messages, but hand-rolled where zod expresses it natively |
| Usefulness | 8/10 | Both types solve real narrow problems; correct absence signal on the helper |
| Enforcement | 6/10 | Solid tests, no exploitable gap found against the real parser; docked for the hand-written/structural split |

No blocking issues from this pass. C1 and I1 remain the only items I'd hold the merge for.

Wirasm:
## Addendum 2 — silent-failure pass (final)

The error-handling pass landed last and is the most consequential of the six. It raises **one new finding at Critical severity** and closes the two questions I'd flagged as unverified. Recommendation is unchanged (REQUEST CHANGES) but the fix list grows.

### New — C2 · `$INPUTS` reaches the model verbatim on three more surfaces, permanently

Neither `applyInputsMacro` **nor** `rewriteNodeOutputRefs` walks these three text surfaces:

| Surface | Where | Runtime behavior |
|---|---|---|
| `agents[].prompt` | `dag-node.ts:123` | Passed straight to the Claude SDK (`dag-executor.ts:1122` → `claude/provider.ts:567-571`). No substitution, ever. |
| `systemPrompt` | `dag-node.ts:219` | Wired directly to `baseOptions.systemPrompt` (`dag-executor.ts:1108`). No substitution, ever. |
| `approval.on_reject.prompt` | `dag-node.ts:395` | Runs through prompt substitution on a rejection retry (`dag-executor.ts:5202-5231`) — but that path only understands `$id.output`, not `$INPUTS`. |

**Confirmed by repro** — caller supplies `detail`, control substitutes, the other three don't, and `errors: 0`:

```
errors: 0
prompt        (control): "main CLEAN-TEMP-FILES"
agents.helper.prompt   : "Sub-task: $INPUTS.detail"
systemPrompt           : "You handle $INPUTS.detail"
on_reject.prompt       : "Retry with $INPUTS.detail"
```

**Why this is worse than the equivalent `$node.output` gap, and why I'm rating it Critical rather than "pre-existing."** These three are pre-existing blind spots in `rewriteNodeOutputRefs`, so `applyInputsMacro` technically *did* match it field-for-field here — no spec violation. But the two macros have different fallbacks. A surface `rewriteNodeOutputRefs` misses is only a *namespacing* miss; the executor's runtime `substituteNodeOutputRefs` pass still resolves the ref. `$INPUTS` has no such second chance — I verified there is **no `$INPUTS` handling anywhere outside `include-expander.ts` and the schemas**, so load-time expansion is the sole resolution path. A missed surface is therefore permanent literal `$INPUTS.<name>` text delivered to the model, and because the field is never visited it is never recorded in `missingInputs` either — no load error whether the caller supplies the value or forgets it entirely.

That is precisely the failure #2466's "Read this first" section says must never happen: *"an author should learn this from the PR, not from `$INPUTS.scope` silently reaching the model as literal text."*

**Recommended:** wire all three into `applyInputsMacro` (cheap — they're plain string fields). If you'd rather scope this PR tightly, the minimum acceptable alternative is an explicit documented reserved-surface list plus a tracked follow-up, so it's a stated boundary rather than an undocumented hole. Silently shipping it is the one option I'd argue against.

### Partial restore of S6 — my downgrade was too broad

In Addendum 1 I downgraded the duplicated identifier regex to "not a finding" on the grounds that the same char class is already unshared across five other files. That reasoning holds for the *family*, but I over-corrected on the specific pair. The drift is **one-directional**, and the dangerous direction is silent:

- Loosen the `with:`-key validator (`dag-node.ts`) without `INPUTS_REF` → e.g. `with: {my.key: "v"}` validates, but `$INPUTS.my.key` matches only `$INPUTS.my`, leaving `.key` as trailing literal text. **Silent wrong substitution, no error.**
- Loosen `INPUTS_REF` without the key validator → the matching `with:` key fails schema validation and the node fails to load. **Loud and safe.**

Unlike the other five sites (which encode node-id output refs — a related but distinct concept), these two were introduced by the same PR, same author, same feature, and encode the identical semantic concept "a valid include-input name." Narrow recommendation: one shared `INPUT_NAME_PATTERN` constant for this pair only. Explicitly **not** the broader five-site refactor — that's older, cross-cutting, and out of scope here.

### Closed — the two questions I'd flagged as unverified

- **Expansion errors are surfaced correctly, not silently dropped.** `expandWorkflowIncludes` catches `IncludeExpansionError` per top-level workflow, pushes `{filename, error, errorType}` into `errors[]`, and excludes that workflow from the returned map. Two confirmed surfacing points: `/workflow list` renders a "N workflow(s) failed to load:" section (`command-handler.ts:612-653`, capped at 10), and `GET /api/workflows` returns it in its documented `errors?` field. The resilient drop-one-report-it-keep-the-rest mechanism is sound — which is exactly what makes I1 user-visible as a load failure rather than a background warning.
- **The prototype check is correct, with no false accepts or rejects.** Tested against `Bun.YAML.parse` (the real parser, `loader.ts:85`): bare timestamps parse as strings not `Date`; mappings never yield `Map`; `null`/arrays are caught by earlier clauses. So the clause is defensive redundancy, not live logic. **No prototype-pollution vector** — `__proto__` arrives as an ordinary own key with `Object.prototype` intact. Nothing to change.

### Also confirmed clean

- `missingInputs` throw placement — `.map()` runs its callback unconditionally, and the `loop_group` recursion passes the same `Set` by reference, so nested-body misses land before the post-map throw. No interaction bug with the entry-node `when:` combining logic.
- The dangling-ref safety net genuinely works for `$node.output` — `validateDagStructure` really does re-validate the flattened DAG in the right order (after macro substitution), so a caller's dangling `$typo.output` is caught. It provides zero protection for `$INPUTS`, per C1/C2.
- The zod `.transform()` question that split two passes is settled by execution: `transformCalls: 0`, `success: false`. The transform does **not** run after a non-fatal `ctx.addIssue()`, so the cast is unreachable — as originally reported.

### Revised fix list

**Before merge:** C1 (`fan_out.items` — one line), C2 (three surfaces, or an explicit documented boundary + follow-up), I1 (revert the unresolved-file branch to WARN, or sign off on the deviation explicitly).
**Cheap follow-ups:** I2, I3, I4, S6-narrowed, S10 (`isCommandNode`).

The underlying implementation remains careful and well-tested; the gaps are all in surface coverage, and C1/C2 share one root cause — the ref-surface enumerations were never mechanically tied together, which is exactly what I3's stale "KEEP IN SYNC" comment was supposed to prevent.

Wirasm:
Follow-ups from this PR are now filed, so the deferrals are tracked rather than implied:

- **#2475** — `$nodeId.output` in `approval.on_reject.prompt` escapes both the loader's ref validation and include namespacing. This is the issue `include-expander.ts:227` points at with "a separate, lower-severity gap tracked on its own" — that pointer was dangling until now. It also notes one sentence of that docblock understates the severity: a rewrite miss resolves the *pre-namespacing* id, so with a same-named parent node the ref binds to the wrong producer rather than to nothing.
- **#2476** — `systemPrompt:` and `agents:` are passed to the provider verbatim (`dag-executor.ts:1108/1122/1134`); no workflow-variable or `$nodeId.output` substitution at all. Filed as a decision (substitute vs. document as literal), not a defect — but the current state is hard to defend either way, since `$INPUTS` interpolates there and `$ARTIFACTS_DIR` does not.
- **#2477** — I2, the `loop_group` body carve-out in `scanBlockCommandRefs`. The docblock at `:440-442` and the guide both state it, which is why it was accepted here; the issue is to close it. `rewriteNodeOutputRefs` and `applyInputsMacro` already recurse into bodies, so this is the only one of the three that stops at the top level.

One thing worth folding in **only if you are already pushing**: the two `KEEP IN SYNC` comments now disagree. `include-expander.ts:161` says FOUR enumerations and counts `applyInputsMacro`; `loader.ts:237` still says three and omits it. That is I3 fixed on one side. Not worth restarting CI on its own — #2475 picks it up.

PR #2483, opened 2026-08-05
fix: issue 2465

Wirasm:
Automated fix for issue $ARGUMENTS.

## Assessment
'The problem is real and exactly as described: TRUNCATION_MARKER_PATTERN is anchored at $/end-of-string, so any trailing whitespace breaks detection. Today only formatPersistedBashOutput produces this string and it returns `head + marker` with nothing appended, so no production path currently triggers the gap — but the coupling is invisible (normalising, line-ending conversion, or code-fence stripping anywhere downstream would silently revert the accurate '\''output was clipped'\'' advice to the generic '\''producer emitted no JSON'\'' advice, with no test failing). The fix is a one-word change (`output.trimEnd()`) that loses nothing — trailing whitespace after the marker can never be meaningful — and the property the function claims to guarantee ('\''ends with the truncation marker'\'') is restored. Adding a test that pins the trailing-whitespace case is trivial and turns the property from incidental to asserted. Worth doing now while the change is one line; the larger refactor (threading node_output_truncated onto NodeOutput) is only justified if another consumer needs the flag.'

## Final review pass
'BUILD-CLEAN'

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of truncated output when the truncation marker is followed by whitespace.

* **Tests**
  * Added coverage for marker formatting, byte counts, whitespace, quoted text, missing markers, and empty output.

### DIFF
--- a/packages/workflows/src/utils/output-truncation.test.ts
+++ b/packages/workflows/src/utils/output-truncation.test.ts
@@ -1,0 +1,50 @@
+import { describe, test, expect } from 'bun:test';
+import { buildTruncationMarker, hasTruncationMarker } from './output-truncation';
+
+describe('buildTruncationMarker', () => {
+  test('formats the byte count into the marker', () => {
+    expect(buildTruncationMarker(1234)).toBe('\n\n… [truncated; original output was 1234 bytes]');
+  });
+
+  test('handles zero bytes', () => {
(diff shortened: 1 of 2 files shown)


Wirasm:
Closing — this was a dogfood run of the t1-fix-issue test workflow, not a change we intend to land. The fix itself was correct; the run exposed a gap in the workflow's validation gate (it type-checked but never ran the tests it wrote), which is now fixed. Re-running.

PR #2503, opened 2026-08-06
fix(git): never guess the default branch when origin/HEAD is missing

### LINKED ISSUE #2471: fix(git): base-branch detection guesses 'main' when main merely exists — error instead of guessing
`getDefaultBranch` (`packages/git/src/branch.ts`) ends its fallback chain by **guessing**:

```ts
// Fallback: check if <remote>/main exists, otherwise throw
try {
  await execFileAsync('git', ['-C', repoPath, 'rev-parse', '--verify', `${remote}/main`], ...);
  return toBranchName('main');
}
```

That is not "detect the default branch" — it is *"if a branch named `main` exists at all, assume it is the base."*

For a repository shaped like this one — **`dev` is the default, `main` is the release branch** — it returns the wrong answer **every time**, silently. The run branches off `main`, the PR targets `main`, and nothing errors.

The existing error message right below it is already the correct behaviour, just wired to the wrong condition:

```
Cannot detect default branch for <path>: neither <remote>/HEAD nor <remote>/main exist.
Set worktree.baseBranch in .archon/config.yaml to specify the branch explicitly.
```

That message fires only when `main` is *absent*. It should fire whenever the base branch cannot be **determined** — which is a different and much larger set.

## Why this stayed hidden

`.archon/config.yaml` in this repo sets `worktree.baseBranch: dev`, and config beats detection. The resolution chain in `executor.ts` is:

```
CLI --base  >  worktree.baseBranch  >  codebase.default_branch  >  getDefaultBranch(git)
```

So every repo that has configured a base branch is safe, and the defect only bites repos that **haven't** — which is exactly the population least equipped to notice a run quietly targeting the wrong branch.

## Second, independent defect: `origin/HEAD` goes stale and nothing refreshes it

Step one of the chain reads `refs/remotes/origin/HEAD`. That symbolic ref is written **once at clone time and never updated**. If a project later changes its default branch, every existing clone keeps reporting the old one indefinitely.

Observed on a maintainer machine on 2026-08-05:

```
local  refs/remotes/origin/HEAD  →  origin/main
GitHub default branch            →  dev
```

Fixed locally with `git remote set-head origin --auto`, but nothing in Archon runs that, prompts for it, or notices the disagreement. So the *legitimate* detection step can also be confidently wrong, with no signal.

## Proposed behaviour

**Never guess a base branch. If it cannot be determined, error and tell the user how to set it.**

1. `--base` flag
2. `worktree.baseBranch` in `.archon/config.yaml`
3. `codebase.default_branch`
4. `refs/remotes/<remote>/HEAD` — git's own record of the remote default
5. ~~`<remote>/main` exists ⇒ `main`~~ → **error**, reusing the message that already exists

The error should name all three configuration surfaces, not just `worktree.baseBranch`, so the reader sees the cheapest fix for their situation.

## Open question for step 4

Keeping the symbolic ref is the zero-config path that is right for most clones, and dropping it would force every user to configure a base branch — a real UX regression. But it is stale-prone in exactly the way above, and staleness is invisible.

Options, roughly in cost order:

- **Keep as-is.** Accept the staleness; the new error at step 5 at least stops the `main` guess.
- **Verify on the error path only.** When something downstream looks wrong, run `git ls-remote --symref <remote> HEAD` (authoritative, live) to produce a better message. No cost on the happy path.
- **Refresh opportunistically.** Run `git remote set-head <remote> --auto` during workspace sync, so long-lived clones self-heal. One network round trip on a path that already talks to the remote.
- **Always query live.** Correct, but adds a network call to base-branch resolution.

Maintainer's call. The step-5 fix is worth doing regardless of which is chosen.

## Acceptance

- A repo where `main` exists but is **not** the default, with no base branch configured, **errors** instead of returning `main`
- The error names `--base`, `worktree.baseBranch`, and the codebase default branch
- A repo with any of the three configured is unaffected, byte-for-byte
- A repo with a correct `<remote>/HEAD` and no configuration continues to work with no new prompts

Found while reviewing #2467, where a stale `origin/HEAD` pointing at `main` coincided with a docs example using `main` — two unrelated causes producing the same wrong word.

Wirasm:
## Summary

- **Problem:** `getDefaultBranch` ends its fallback chain by guessing. If `refs/remotes/<remote>/HEAD` is not a symbolic ref, it checks whether `<remote>/main` merely *exists* and returns `main` if so. That is not detection — it is "main exists ⇒ assume it is the base."
- **Why it matters:** for a repo shaped like this one, where `dev` is the default and `main` is the release branch, the run silently cuts its worktree from `main` and targets the PR at `main`. Nothing errors. The failure is invisible at the moment it happens and expensive to unpick later.
- **What changed:** the guess is gone. If `<remote>/HEAD` is not a symbolic ref, throw with a message naming all three configuration surfaces — `--base`, `worktree.baseBranch`, and the codebase `default_branch` field — so the reader sees the cheapest fix for their situation.
- **What did NOT change (scope boundary):** the unexpected-error path (permission denied, git corruption) still surfaces as before. No caller's resolution order changed. A repo whose `<remote>/HEAD` is set correctly is unaffected.

## Provenance

This commit was **written by an Archon run** and then orphaned — it lived only inside a run worktree with no PR, and was found during a worktree sweep on 6 Aug that would otherwise have deleted it. Rescued, rebased onto current `dev`, and verified by hand before opening (see Human Verification).

### Before

```
  <remote>/HEAD not set
        │
        ▼
  does <remote>/main exist? ──yes──▶ return "main"     ← silent, wrong for this repo
        │
        no
        ▼
      throw
```

### After

```
  <remote>/HEAD not set
        │
        ▼
      throw ──▶ "Cannot detect default branch: <remote>/HEAD is not set.
                 Pass --base, set worktree.baseBranch in .archon/config.yaml,
                 or set the codebase default_branch field."
```

### Before

```
  getDefaultBranch ──symbolic-ref──▶ ok
        └──catch──▶ rev-parse <remote>/main ──▶ "main"   [-] removed
```

### After

```
  [~] getDefaultBranch ──symbolic-ref──▶ ok
        └──catch──▶ throw (config-directed message)
```

**Connection inventory** — every caller, and whether the new throw can reach a user:

| `workflows/executor.ts:1187` | getDefaultBranch | unchanged | **already wrapped in try/catch** — warns `workflow.base_branch_auto_detect_failed` and sets `baseBranch = ''`. Throw is non-fatal; `$BASE_BRANCH` only fails if a prompt actually references it. |
| `cli/commands/workflow.ts:1482` | getDefaultBranch | unchanged | Last resort after `--base`, `rawBase`, and `codebaseDefaultBranch`. Propagates. |
| `core/services/cleanup-service.ts:58` | getDefaultBranch | unchanged | Reads config first by design (comment at `:48`). Propagates. |
| `git/repo.ts:160` | getDefaultBranch | unchanged | Only when `baseBranch` arg is omitted. Propagates. |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: S`
- Scope: `git`
- Module: `git:branch`

## Change Metadata

- Change type: `bug`
- Primary scope: `git`

## Linked Issue

- Closes #2471

```bash
bun run type-check                 # all packages, exit 0
bun --filter "@archon/git" test    # 195 pass, 0 fail, 302 expect() calls
```

- Evidence: `packages/git/src/git.test.ts` is rewritten in this diff — the tests that asserted the `main` fallback are replaced by tests asserting the throw and its message.
- Skipped: full `bun run test` — the change is one function in a leaf package with four callers, all enumerated above and none of whose resolution order changed. CI runs everything.

## Security Impact (required)

- New external network calls? `No` — this removes a `git rev-parse` call.

## Compatibility / Migration

- Backward compatible? **No, deliberately.** One case changes: `<remote>/HEAD` unset **and** `<remote>/main` exists. That previously returned `main`; it now throws. That case *is* the bug.
- Config/env changes? `No` — but a user hitting the new throw fixes it with `git remote set-head origin -a`, or any of the three config surfaces named in the message.

- Verified `--base` is real before trusting the error message: `packages/cli/src/cli.ts:158` — `--base <branch>`, "Per-dispatch base override for epic slices (worktree cut-from + PR target)". The message does not point at a flag that doesn't exist.
- Verified the "config outranks detection" claim by reading the executor chain at `executor.ts:1175-1196` rather than taking the commit body's word: override → `config.baseBranch` → `fallbackBaseBranch` → folder-project → detection, with detection wrapped in try/catch. This repo sets `worktree.baseBranch: dev`, so it never reaches detection at all.
- Enumerated all four non-test callers and checked each for propagation (table above).
- Not verified: behavior on a real repo with `<remote>/HEAD` genuinely unset. Reasoned from the code path, not reproduced.

- Affected: base-branch resolution for workflow runs, CLI worktree reuse, cleanup service, workspace sync.
- Potential unintended effect worth naming: **`actions/checkout` does not always set `refs/remotes/origin/HEAD`.** Any of the four callers running in that environment without a configured base would newly throw where it previously guessed `main`. The executor path degrades gracefully; the other three propagate. If that turns out to bite, the fix is a config surface, not restoring the guess.
- Guardrails: the throw is loud and names its own remedies — the opposite of the silent misroute it replaces.

## Rollback Plan (required)

- Fast rollback: `git revert <sha>` — restores the fallback exactly.
- Feature flags: none.
- Observable failure symptoms: `default_branch_detection_failed` warnings, or `Cannot detect default branch` errors, on repos that previously worked.

## Risks and Mitigations

- Risk: turning a silent success into a hard failure in an environment we can't fully enumerate (CI checkouts, manually-added remotes).
  - Mitigation: three config surfaces already exist and all outrank detection; the error message names all three. The executor — the highest-traffic caller — already treats detection failure as non-fatal.

## Summary by CodeRabbit

- **Bug Fixes**
  - Default branch detection now reports an error when the remote’s default branch cannot be resolved instead of assuming `main`.
  - Error messages provide guidance for configuring the default branch through available options.
  - Warnings are logged when the remote default branch is unset.
  - Missing repository paths and other unexpected failures now produce clearer error reporting instead of being treated as branch-resolution issues.


Wirasm:
---
pr: 2503
title: "fix(git): never guess the default branch when origin/HEAD is missing"
author: "Wirasm"
reviewed: 2026-08-12T13:47:32Z
---

# PR Review: #2503 - Never guess the default branch

**Author**: @Wirasm  
**Branch**: `fix/2471-no-default-branch-guess` → `dev`  
**Files Changed**: 2 (+68/-107)

## Summary

The main behavior is correct: `getDefaultBranch()` no longer silently guesses `main` when `origin/HEAD` is unset, and the error names the supported configuration paths. One still-valid error-classification bug should be fixed before merge because it converts an invalid repository path into misleading configuration advice and a warning-level log.

## Implementation Context

| Artifact | Path |
|---|---|
| Implementation Report | Not found |
| Original Plan | Not found |
| Documented Deviations | N/A |

## Changes Overview

| File | Changes | Assessment |
|---|---:|---|
| `packages/git/src/branch.ts` | +26/-39 | WARN — correct main behavior, one over-broad error match |
| `packages/git/src/git.test.ts` | +42/-68 | WARN — missing invalid-repository regression |

## Issues Found

### Critical

No critical issues found.

### Important

- **`packages/git/src/branch.ts:52`** — `errorText.includes('No such file or directory')` treats a missing `repoPath` as an unset remote HEAD.
  - **Why**: `git -C /missing ...` reports `fatal: cannot change to '/missing': No such file or directory`. The current branch catches that text, logs `default_branch_detection_failed`, and tells the caller to set `--base`/config, even though no branch setting can repair a nonexistent checkout. This violates the function's documented rule that operational failures take the unexpected-error path.
  - **Observed**: calling the PR implementation with `/tmp/archon-pr2503-review-missing` returned `Cannot detect default branch ... origin/HEAD is not set` instead of `Failed to get default branch ... cannot change to ...`.
  - **Fix**: only classify the specific `not a symbolic ref` condition as an unset remote HEAD. Let `No such file or directory` reach `default_branch_symbolic_ref_failed`, and add a regression test using a missing repository path that asserts the operational error and error-level log.

### Suggestions

No additional suggestions.

## Validation Results

| Check | Status | Details |
|---|---|---|
| Current `dev` merge | PASS | Clean no-commit merge onto `ffd7a321` |
| `bun run validate` | PASS | Bundled checks, type-check, lint, formatting, installer tests, and all package tests |
| GitHub CI | PASS | Ubuntu, Windows, Docker, and CodeRabbit green on PR head |
| Missing-repo behavior | FAIL | Reproduced misleading unset-HEAD classification |

## Pattern Compliance

- [x] Removes unsafe default-branch guessing
- [x] Preserves strict TypeScript and existing git wrapper usage
- [x] Tests cover configured symbolic-ref and unset-HEAD behavior
- [ ] Unexpected filesystem/repository failures remain distinguishable
- [ ] Regression coverage includes a missing `repoPath`

## Strengths

The core direction is sound and fail-fast: an ambiguous default branch is surfaced instead of silently routing work to `main`. The user-facing error also names all three supported configuration surfaces.

## Recommendation

**REQUEST CHANGES**

Fix the single over-broad error classification and add its regression test. After that, this should be safe to merge.

*Reviewed by Codex*
*Report: `/Users/rasmus/.prp/archon-75601ef6/reviews/pr-2503-review.md`*

PR #2513, opened 2026-08-07
fix(db): drop the premature event_order index that crash-loops Postgres upgrades

### LINKED ISSUE #2508: Postgres schema apply crash-loops on upgrade: premature CREATE INDEX references event_order before ALTER TABLE adds it
## Summary

- **What broke:** On PostgreSQL, applying the combined schema (`migrations/000_combined.sql`) crash-loops the server/CLI at boot when upgrading an **existing** database. A premature `CREATE UNIQUE INDEX` at lines 285–287 references the `event_order` column ~265 lines *before* the `ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_order` that actually adds it (line 549). On any pre-#2414 database the column does not yet exist when the index statement runs, so Postgres aborts with `ERROR: column "event_order" does not exist (SQLSTATE 42703)`.
- **When it started:** Introduced by **#2414** (commit `c47b2362`), which added `event_order` + the `idx_workflow_events_run_order` unique index in **two** places instead of one.
- **Severity:** `blocker` — every Postgres install that predates #2414 fails to start after upgrade; the whole schema-apply transaction rolls back and re-throws fatally, so the boot repeats indefinitely.

## Steps to Reproduce

1. Have (or simulate) a **pre-#2414** Postgres database — i.e. one whose `remote_agent_workflow_events` table exists but has **no** `event_order` column.
2. Start Archon with `DATABASE_URL` pointing at that database (or apply `migrations/000_combined.sql` transactionally, mirroring `initSchema()`'s `BEGIN`/`COMMIT` + `ON_ERROR_STOP`).
3. The schema apply reaches the premature index at line 285.

Faithful simulation used during verification:
```bash
# 1. Fresh DB with the current (#2414) schema succeeds
psql "$DATABASE_URL" -v ON_ERROR_STOP=1 -f migrations/000_combined.sql   # exit 0

# 2. Simulate a pre-#2414 DB: drop the newly-added column/index/sequence
psql "$DATABASE_URL" -c 'DROP INDEX IF EXISTS idx_workflow_events_run_order;'
psql "$DATABASE_URL" -c 'ALTER TABLE remote_agent_workflow_events DROP COLUMN IF EXISTS event_order;'
psql "$DATABASE_URL" -c 'DROP SEQUENCE IF EXISTS ...event_order_seq;'

# 3. Re-apply the migration transactionally -> crash
psql "$DATABASE_URL" -1 -v ON_ERROR_STOP=1 -f migrations/000_combined.sql   # exit 3
```

## Expected vs Actual

- **Expected:** Applying `000_combined.sql` against a pre-#2414 database converges idempotently — the `ADD COLUMN` adds `event_order`, then the index is created on top of it. The server boots.
- **Actual:** The apply aborts at line 287:
  ```
  psql:migrations/000_combined.sql:287: ERROR:  42703: column "event_order" does not exist
  LINE 3:   WHERE event_order IS NOT NULL;
  ```
  Because `initSchema()` wraps the entire combined SQL in a single `BEGIN`/`COMMIT` transaction and re-throws `fatal` on any error, the transaction rolls back (verified: `event_order` still absent after the failed apply) and the boot crash-loops.

## User Flow

```
  Operator                Archon (Postgres adapter)          PostgreSQL
  ────────                ─────────────────────────          ──────────
  upgrade + restart ────▶ initSchema(): BEGIN
                          apply 000_combined.sql ──────────▶ CREATE TABLE IF NOT EXISTS
                                                             (no-op: table already exists,
                                                              event_order NOT added)
                                                    ────────▶ line 285: CREATE UNIQUE INDEX
                                                              ... (workflow_run_id, event_order)
                          [X] fatal re-throw       ◀──────── ERROR 42703 column does not exist
                          ROLLBACK
  server never starts ◀── crash, restart, repeat  ◀──────── (nothing committed)
```

## Environment

- Platform: Web / CLI (any surface that opens the DB — the server and *every* CLI invocation apply schema on connect)
- Database: **PostgreSQL** (SQLite path is unaffected — it applies schema via `createSchema()` in `sqlite.ts`, not this SQL)
- Running in worktree? No (boot-time schema apply)
- OS: Linux (reproduced on Postgres 17.9)

## Logs

```
psql:migrations/000_combined.sql:287: ERROR:  42703: column "event_order" does not exist
LINE 3:   WHERE event_order IS NOT NULL;
                ^
PSQL_EXIT: 3
```
Verbose error confirmed `SQLSTATE 42703` originating in `errorMissingColumn`.

## Impact

- **Affected workflows/commands:** All — the server (and every CLI invocation) fails to open the database, so nothing runs.
- **Reproduction rate:** Always, on any Postgres DB created before #2414. Fresh installs are unaffected (the `CREATE TABLE` body adds `event_order` at line 269, so the line-285 index succeeds) — which is why this slipped past testing.
- **Workaround:** Manually add the column before restarting:
  ```sql
  ALTER TABLE remote_agent_workflow_events ADD COLUMN IF NOT EXISTS event_order bigint;
  ```
  Then boot proceeds and the correctly-ordered block at 549–557 backfills/indexes.
- **Data loss risk?** No (transaction rolls back; no partial writes commit). But the install is unusable until patched or manually worked around.

## Root Cause

Commit `c47b2362` (#2414) added the column + index in **two** places:

- **Lines 285–287** — a `CREATE UNIQUE INDEX idx_workflow_events_run_order ON remote_agent_workflow_events (workflow_run_id, event_order) WHERE event_order IS NOT NULL;` placed *inside/after* the `CREATE TABLE IF NOT EXISTS remote_agent_workflow_events` block.
- **Lines 549–557** — the correct migration block: `ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_order` → sequence/default → the **same** index (lines 555–557).

On a fresh DB both work (column created by `CREATE TABLE` at line 269). On an existing DB, `CREATE TABLE IF NOT EXISTS` is a no-op so the column is never added, and the line-285 index references a nonexistent column → 42703 → full-transaction rollback → crash-loop.

This is exactly the hazard called out in CLAUDE.md's "Schema changes are additive-only" note: `CREATE TABLE IF NOT EXISTS` never alters an existing table, so a column only reachable through the `CREATE TABLE` body doesn't exist on upgraded databases — anything referencing it must come **after** an `ADD COLUMN`.

## Suggested Fix

- **Delete the premature index at lines 285–287.** It is fully redundant with the correctly-ordered copy at 555–557, which runs *after* `ADD COLUMN IF NOT EXISTS event_order`.
- Keep the `event_order` column declaration in the `CREATE TABLE` body (line 269) so fresh DBs still get it.
- After editing, run `bun run generate:bundled-schema` to refresh the embedded schema, and keep the SQLite adapter's `createSchema()` mirrored (the parity test in `sqlite.test.ts` guards drift).
- Consider a regression test that applies `000_combined.sql` against a DB missing `event_order` (i.e. simulating a prior schema vintage) to catch premature-reference ordering bugs generally.

## Relevant Code

- `migrations/000_combined.sql:285-287` — premature `CREATE UNIQUE INDEX` (the bug)
- `migrations/000_combined.sql:269` — `event_order` column in the `CREATE TABLE` body (fresh-DB path)
- `migrations/000_combined.sql:549-557` — correct `ALTER TABLE ... ADD COLUMN` + index block
- `packages/core/src/db/adapters/postgres.ts:68-109` — `initSchema()` applies the combined SQL as one `BEGIN`/`COMMIT` transaction and re-throws fatally on error
- `packages/core/src/db/adapters/sqlite.ts` — `createSchema()` (mirror target; unaffected path)

## Scope

- Module: `core:db/adapters/postgres` + `migrations/000_combined.sql`
- Regression from: **#2414** (`c47b2362`)

133Felix:
## Summary

- **Problem:** On PostgreSQL, applying `migrations/000_combined.sql` against an **existing** database aborts with `ERROR: 42703: column "event_order" does not exist`. `idx_workflow_events_run_order` is created in the `CREATE TABLE IF NOT EXISTS remote_agent_workflow_events` section (line 285), ~265 lines *before* the `ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_order` that creates the column it indexes (line 549).
- **Why it matters:** `initSchema()` wraps the whole file in one transaction and re-throws `fatal`, so the apply rolls back and the server **crash-loops** — every Postgres install predating #2414 fails to start after upgrading, and every CLI invocation fails too (both apply schema on connect). Fresh installs are unaffected, because the real `CREATE TABLE` declares `event_order` inline — which is exactly why this shipped green.
- **What changed:** Deleted the premature `CREATE UNIQUE INDEX`. The identical statement in the additive block (line 556) already creates it directly after the `ADD COLUMN`, which is correct for both a fresh and an upgrading database. Added a statement-ordering regression test that reads the SQL and asserts on offsets.
- **What did *not* change (scope boundary):** The SQLite path (already fixed by #2418), the inline `event_order BIGINT` in the `CREATE TABLE` body (kept — it is what makes fresh installs correct), and the two sibling instances of this same bug class (`idx_conversations_hidden`, `idx_sessions_parent`) which are already covered by the open #2443.

### Before

```
  Operator                Archon (Postgres adapter)          PostgreSQL
  ────────                ─────────────────────────          ──────────
  upgrade + restart ────▶ initSchema(): BEGIN
                          apply 000_combined.sql ──────────▶ CREATE TABLE IF NOT EXISTS
                                                             (no-op: table exists,
                                                              event_order NOT added)
                                                   ────────▶ line 285: CREATE UNIQUE INDEX
                                                              ...(workflow_run_id, event_order)
                          [X] fatal re-throw      ◀───────── ERROR 42703
                          ROLLBACK
  server never starts ◀── crash, restart, repeat ◀────────── (nothing committed)
```

### After

```
  Operator                Archon (Postgres adapter)          PostgreSQL
  ────────                ─────────────────────────          ──────────
  upgrade + restart ────▶ initSchema(): BEGIN
                          apply 000_combined.sql ──────────▶ CREATE TABLE IF NOT EXISTS
                                                             (no-op)
                                                   ────────▶ *line 285: index NOT created here*
                                                   ────────▶ ALTER TABLE ADD COLUMN
                                                              IF NOT EXISTS event_order  [+]
                                                   ────────▶ CREATE SEQUENCE / SET DEFAULT
                                                   ────────▶ CREATE UNIQUE INDEX
                                                              ...(workflow_run_id, event_order)
                          COMMIT                  ◀───────── ok
  server starts ◀──────── healthy
```

Fresh installs take the same path and are unchanged: `CREATE TABLE` declares `event_order` inline, the `ADD COLUMN` is a no-op, and the single index creation succeeds.

### Before

```
  migrations/000_combined.sql
    ├── [Table 6 block]  CREATE TABLE IF NOT EXISTS remote_agent_workflow_events
    │                      (event_order BIGINT declared inline)
    │                    CREATE UNIQUE INDEX idx_workflow_events_run_order  ◀── premature
    │
    └── [additive block] ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_order
                         CREATE SEQUENCE ... OWNED BY ...event_order
                         ALTER COLUMN event_order SET DEFAULT nextval(...)
                         CREATE UNIQUE INDEX idx_workflow_events_run_order  ◀── duplicate

  packages/core/src/db/bundled-schema.generated.ts   (generated from the above)
  packages/core/src/db/adapters/postgres.ts          (initSchema: BEGIN / apply / COMMIT)
```

### After

```
  migrations/000_combined.sql                                                   [~]
    ├── [Table 6 block]  CREATE TABLE IF NOT EXISTS remote_agent_workflow_events
    │                      (event_order BIGINT declared inline)
    │                    -- comment explaining why the index is NOT here        [+]
    │
    └── [additive block] ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_order
                         CREATE SEQUENCE ... OWNED BY ...event_order
                         ALTER COLUMN event_order SET DEFAULT nextval(...)
                         CREATE UNIQUE INDEX idx_workflow_events_run_order      ◀── sole copy

  packages/core/src/db/bundled-schema.generated.ts                              [~] regenerated
  packages/core/src/db/migration-statement-order.test.ts                        [+] new guard
  packages/core/src/db/adapters/postgres.ts                                         unchanged
```

**Connection inventory**

| `migrations/000_combined.sql` | `bundled-schema.generated.ts` | unchanged | Regenerated via `bun run generate:bundled-schema`; `check:bundled-schema` green |
| `postgres.ts initSchema()` | `migrations/000_combined.sql` | unchanged | Same single-transaction apply; only the SQL content changed |
| `migration-statement-order.test.ts` | `migrations/000_combined.sql` | **new** | Reads the file and asserts statement offsets; no mocks, no database |
| `sqlite.ts createSchema()` | `migrations/000_combined.sql` | unchanged | SQLite maintains its schema separately; unaffected (#2418 fixed that path) |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: S`
- Scope: `core`
- Module: `core:db`

## Change Metadata

- Change type: `bug`
- Primary scope: `core`

## Linked Issue

- Closes #2508
- Related #2414 (introduced the duplicate), #2418 (same reordering, SQLite path), #2443 (same bug class, sibling indexes `idx_conversations_hidden` / `idx_sessions_parent`)

```bash
bun run validate     # EXIT=0 — all nine checks green
```

- **Evidence provided:**
  - `bun run validate` exits 0: `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, `type-check`, `lint --max-warnings 0`, `format:check`, `test:install`, `test`.
  - New test red-green verified. With the fix reverted (`git stash push -- migrations/000_combined.sql`):
    ```
    (fail) migrations/000_combined.sql — statement ordering > event_order is added before anything indexes it (#2508)
     2 pass
     1 fail
    ```
    With the fix applied: `3 pass, 0 fail`.
  - **Reproduced and fixed against a real, live Postgres database** — see Human Verification.
- **Skipped commands:** none.

## Security Impact (required)

- File system access scope changed? `No` — the new test reads `migrations/000_combined.sql` from the repo, the same file `getSchemaSQL()` already reads in source builds.

## Compatibility / Migration

- Backward compatible? `Yes`
- Config/env changes? `No`
- Database migration needed? `No` — this repairs the migration itself. Fresh installs are byte-identical in outcome (same column, same sequence, same index, same default); upgrades now converge instead of aborting.
- **For anyone already stuck in the crash-loop**, this PR fixes the boot going forward. The manual unblock is the same statement the migration performs:
  ```sql
  ALTER TABLE remote_agent_workflow_events ADD COLUMN IF NOT EXISTS event_order BIGINT;
  ```

- **Verified scenarios:**
  - Hit this in production while upgrading a self-hosted install from **0.6.0 to 0.8.0** against PostgreSQL. The container crash-looped with `db.postgres_schema_init_failed` → `database_connection_failed` → exit 1, repeating indefinitely; `/api/health` never answered. Log:
    ```
    {"level":60,"module":"db.postgres","err":{"type":"DatabaseError",
     "message":"column \"event_order\" does not exist","code":"42703",
     "position":"12720"},"msg":"db.postgres_schema_init_failed"}
    ```
    Byte offset 12720 lands exactly on the premature index this PR removes.
  - Applying the `ADD COLUMN` by hand and restarting brought the app up immediately (healthy after ~6s), which isolates this statement as the sole cause.
  - After the restart the rest of the schema applied correctly: `output_root` present, `event_order` carrying `DEFAULT nextval(...)`, `idx_workflow_events_run_order` present, `remote_agent_schema_version` created (18 → 19 tables).
- **Edge cases checked:**
  - 23,221 pre-existing event rows keep `event_order IS NULL` — no backfill, exactly as the migration comment intends, and what the `COALESCE(event_order, 0)` read path expects.
  - Fresh-install path unchanged: the `CREATE TABLE` body still declares `event_order`, asserted by the third test so a later cleanup cannot remove it and silently break new installs.
  - Both the sequence's `OWNED BY` and the `ALTER COLUMN ... SET DEFAULT` are asserted to follow the `ADD COLUMN` too — they fail with the same 42703 if reordered.
- **What was not verified:**
  - I did not re-run a clean-database bootstrap from an empty Postgres in CI; the fresh path is covered by reasoning (the `CREATE TABLE` body is unchanged) plus the third test, not by an end-to-end fresh apply.
  - I did not touch or verify the two sibling instances in #2443. They are the same bug class but latent for my upgrade, because `hidden` and `parent_session_id` already existed in my database.

- **Affected subsystems/workflows:** Postgres schema apply at boot (`initSchema()`), which every server start and every CLI invocation runs. Nothing at runtime — no query, no read path, no API surface changes.
- **Potential unintended effects:** If some environment somehow relied on the index existing *before* the additive block, it would now be created slightly later within the same transaction. Since the whole file applies atomically, nothing outside the transaction can observe the difference.
- **Guardrails/monitoring:** `db.postgres_schema_init_failed` is already logged at `fatal` with the SQLSTATE and byte position, which is what made this diagnosable in minutes. The new test fails at CI time if any statement referencing `event_order` migrates back above the `ADD COLUMN`.

## Rollback Plan (required)

- **Fast rollback:** `git revert <sha>` — the change is one deleted SQL statement plus a test. Reverting restores the previous (broken-for-upgrades) behaviour, so it should only be done if the deletion turns out to break fresh installs, which the third test guards against.
- **Feature flags / toggles:** none.
- **Observable failure symptoms:** any `column "..." does not exist` at boot in `db.postgres_schema_init_failed`, or `idx_workflow_events_run_order` missing after a successful apply (`\d remote_agent_workflow_events`).

## Risks and Mitigations

- **Risk:** Deleting the index from the `CREATE TABLE` section could leave a fresh install without it, if the additive block were ever made conditional.
  - **Mitigation:** The additive block is unconditional and runs on every apply. The new test pins that the index still exists in the file and still follows the `ADD COLUMN`, and that the `CREATE TABLE` body still declares the column.
- **Risk:** Textual conflict with #2443, which also edits `migrations/000_combined.sql`.
  - **Mitigation:** Different regions — #2443 moves indexes in the conversations/sessions blocks, this removes one in the workflow-events block. Whichever merges second should rebase; the changes are independent and complementary. Happy to rebase onto #2443 if you would prefer to land that one first.
- **Risk:** The regression guard is specific to `event_order` rather than a general "no index before its column" rule.
  - **Mitigation:** Deliberate, to keep this PR's scope to #2508. A general guard would currently fail on the two instances #2443 fixes. Once both land, generalising the test is a small, clean follow-up — happy to write it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration compatibility for both fresh installations and existing databases by ensuring workflow event indexes are created only after the required column is added.

* **Tests**
  * Added validation to confirm migration statements execute in the correct order and preserve the expected schema for new installations.


tazmon95:
Independent verification of this fix, from a production install that hit #2508.

**Real-world impact first:** this bug took our Archon deployment down for ~13 minutes on 2026-08-09. We bumped 0.5.0 → 0.8.0 having passed workflow validation, our full contract suite, unit tests at 100% coverage, runtime gate probes and green CI — every one of which runs against a **fresh** database. The upgrade crash-looped immediately on our existing Postgres and we rolled back. #2508's own line about why it shipped green is exactly why our checks were blind to it too:

> *"Fresh installs are unaffected, because the real `CREATE TABLE` declares `event_order` inline — which is exactly why this shipped green."*

**Verification method.** Schema-only `pg_dump` of the affected production database (18 tables, pre-#2414 — no `event_order` column), restored into a throwaway `postgres:16`, then `migrations/000_combined.sql` applied with `psql -1 -v ON_ERROR_STOP=1` to mirror `initSchema()`'s single-transaction + fatal-rethrow behaviour.

**Results — three paths, same fixture:**

| Path | `000_combined.sql` | Result |
|---|---|---|
| Upgrade (pre-#2414 schema) | `v0.8.0` | ❌ `ERROR: column "event_order" does not exist` at line **287** |
| Upgrade (pre-#2414 schema) | **this PR** | ✅ exit 0 |
| Fresh install (empty DB) | **this PR** | ✅ exit 0 |
| Re-apply ×2 (every boot) | **this PR** | ✅ exit 0, exit 0 |

**Post-state assertions on the upgrade path with this PR applied:**
- `event_order` column present ✅
- `idx_workflow_events_run_order` present ✅ — the index is still created, just in the additive block after the `ALTER TABLE`, so **deleting the premature statement does not lose it**
- Fresh install ends in the same state (column + index) ✅
- Idempotent: 2nd and 3rd apply both exit 0 with the index retained ✅

I checked the index specifically because "delete a `CREATE INDEX`" is the kind of fix that can quietly trade a crash for a missing index. It doesn't here — both paths converge on the same schema.

**Rollback behaviour confirmed too:** after the failing v0.8.0 apply, `event_order` is still absent, i.e. the transaction rolls back cleanly and leaves nothing partially migrated. That matches #2508's claim and is what made our rollback safe.

This fix resolves the failure for us. We've pinned to 0.7.0 in the meantime, since 0.7.1 and 0.8.0 both reproduce it against a pre-#2414 database.

One suggestion, offered rather than asked for: the reason this class escapes CI is that every migration test starts from an empty database. A CI job that applies `000_combined.sql` to a schema snapshot from an **older release tag** before applying the current one would catch this and the two siblings in #2443. Happy to open that separately if it'd be useful.

tazmon95:
Follow-up to my earlier SQL-level verification — I've now tested **the branch itself**, built and booted, rather than just the migration file.

## End-to-end

Built Archon from `baa93763` and booted it against a restored copy of the production Postgres that hit #2508 (schema-only dump, pre-#2414, no `event_order` column):

```
[verify-base] fixture restored: 18 tables
[verify-base] booting archon:2513-test against the restored production schema...
[verify-base] PASS — reached server_ready
```

For contrast, on the same fixture: `ghcr.io/coleam00/archon:0.8.0` and `:0.7.1` both fail with `column "event_order" does not exist` / `42703`, and `:0.7.0` passes. So the branch behaves like the last-good release rather than like the broken ones.

The shipped SQL in the built image is also correct — exactly one real statement, with the deletion documented in place:

```
285:-- idx_workflow_events_run_order is NOT created here (#2508). It indexes
558:CREATE UNIQUE INDEX IF NOT EXISTS idx_workflow_events_run_order
```

## The new regression test earns its place

I checked it isn't a tautology by running it against v0.8.0's unfixed SQL — it **fails**, at the `offset > addColumn` assertion, which is precisely the defect. On the branch it passes (3 tests, 6 assertions).

It's also broader than the reported bug: it asserts *every* `event_order` reference follows the `ADD COLUMN` — the index, `OWNED BY`, and `ALTER COLUMN ... SET DEFAULT` — and separately that the inline `event_order BIGINT` stays in the `CREATE TABLE` body so fresh installs remain correct. That covers the failure class rather than the instance, which is what makes it worth keeping.

## Other checks

- `check:bundled-schema` — up to date, so the embedded schema didn't drift from the SQL.
- `@archon/core` package test script (the batched invocation) — clean.
- `connection.test.ts` isolated — 5 pass.

One note for anyone else verifying: running `bun test packages/core/src/db/` in a single process reports 2 failures in `connection.test.ts`. Those are the `mock.module` pollution CLAUDE.md warns about, not branch regressions — they pass in isolation and under the package's own batched script. I nearly reported them as real.

No issues found. This resolves the crash-loop for us; we're pinned to 0.7.0 until a release carries it.

tazmon95:
@133Felix — apologies, this PR was closed by mistake and I've reopened it. It was not a rejection, and nothing about the fix prompted it.

**What happened:** we carry a downstream image that pins Archon as a base. Yesterday we shipped a temporary in-image patch of #2508 so we could run 0.8.0 before a release carries this fix. The commit message for that change contained the line:

> `Upstream fix archon#2513 is verified — we built Archon from that branch and ...`

`fix <ref>` is a GitHub closing keyword, and `archon#2513` resolved cross-repo. Merging our internal PR therefore auto-closed this one, at the same second, with no comment explaining it. Entirely my wording — I was crediting your work in a commit message and phrased it in a way that closed it instead.

I'm sorry. Having a correct, independently-verified fix closed silently by an unrelated downstream merge is precisely the kind of noise that makes contributing annoying, and you had no way to tell what had happened.

For the record, restating what I posted before the accidental close, since it now sits above it:

- Built Archon from `baa93763` and booted it against a restored copy of a production Postgres that hit #2508 — **reaches `server_ready`**. The same fixture fails on `0.8.0` and `0.7.1` with `column "event_order" does not exist` / `42703`, and passes on `0.7.0`.
- Applied transactionally as `initSchema()` does: **exit 0**, `event_order` present, `idx_workflow_events_run_order` present — the index still gets created, just after the `ALTER TABLE`, so removing the premature statement doesn't lose it.
- Fresh install and repeated re-apply both clean.
- The added regression test isn't a tautology: run it against 0.8.0's unfixed SQL and it **fails** at the `offset > addColumn` assertion. It also asserts the sequence and `SET DEFAULT` references, and that the inline `event_order BIGINT` stays in the `CREATE TABLE` body — so it covers the failure class rather than the one reported instance.

#2508 is still open and I don't see another PR addressing it, so as far as I can tell every Postgres install predating #2414 still can't upgrade. From our side this cost ~13 minutes of production downtime and we're now pinned behind a local patch until something lands.

I've changed our commit conventions so a downstream reference can't do this again.

Wirasm:
# PR Review: #2513 — fix(db): drop the premature event_order index that crash-loops Postgres upgrades

## Outcome

`migrations/000_combined.sql` created `idx_workflow_events_run_order` twice: once inside the `CREATE TABLE IF NOT EXISTS remote_agent_workflow_events` block (old line 285) and once in the additive block after the `ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_order` that actually creates the column on an existing database. On a fresh database both succeed; on an existing one the `CREATE TABLE` is a no-op, so the premature copy references a column that does not exist yet, `initSchema()` rolls back its single transaction and re-throws at `fatal`, and the server crash-loops. The PR deletes only the premature copy, leaves the correct one untouched, regenerates the embedded schema, and adds a statement-ordering guard test.

Both reviewers found no defect. The fix was verified against a real Postgres from two directions — the reviewing agent rebuilt a pre-`event_order` table shape from `da66b19c`, and this session restored the full pre-#2414 schema — and both reproduce the `42703` crash on `dev` and get a clean apply on this branch. CodeRabbit's one inline note (`RegExpMatchArray.index` may be `undefined` at `migration-statement-order.test.ts:36`) was checked and dismissed: `packages/core/tsconfig.json` excludes `**/*.test.ts` so type-check never sees it, and `String.prototype.matchAll()` always yields a defined numeric `.index`. Not reachable.

## Validation

|---|---|---|
| `bun run validate` (PR head, isolated worktree) | PASS | Exit 0. All ten steps: `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, `type-check`, `lint --max-warnings 0`, `format:check`, `test:install`, `test` |
| GitHub CI (`test` ubuntu + windows, `docker-build`) | PASS | Green after re-run. The earlier `windows-latest` failure was `scripts/migrate-state-dir.test.ts` — `EBUSY` on temp-dir `rm` plus a 5000 ms subprocess timeout; a pre-existing Windows flake, no code change made it pass |
| `psql -1 -v ON_ERROR_STOP=1` — pre-#2414 schema fixture, then `dev`'s SQL | FAIL (expected, reproduces #2508) | `ERROR: column "event_order" does not exist` at `migrations/000_combined.sql:287`, whole transaction rolled back |
| `psql -1 -v ON_ERROR_STOP=1` — same fixture, then this PR's SQL | PASS | Exit 0; `event_order` present with `DEFAULT nextval(...)`; `idx_workflow_events_run_order` present; second apply also exit 0 (idempotent) |
| Fresh-install path + fresh-vs-upgraded parity | PASS | Fresh apply exit 0; `pg_indexes` and `information_schema.columns` diffs between the fresh and the upgraded database are empty — identical schema |
| `bun test src/db/migration-statement-order.test.ts` in its assigned batch | PASS | 3/3 standalone; the full batch it was added to in `packages/core/package.json` passes; the file calls no `mock.module()`, so it cannot pollute the batch |

## Critical Issues (0)

## Important Issues (0)

## Suggestions (2)

| Agent | Suggestion | Evidence | Why consider it |
|---|---|---|---|
| `prp-core:seam-analyzer` | The regression guard is pinned to one column, so the bug *class* stays unguarded. A future PR reintroducing this exact shape on a different column passes the test untouched | `packages/core/src/db/migration-statement-order.test.ts:29-51` hardcodes `'ADD COLUMN IF NOT EXISTS event_order'`, `/…idx_workflow_events_run_order/`, `'OWNED BY remote_agent_workflow_events.event_order'`, `'ALTER COLUMN event_order SET DEFAULT'` | The class is not hypothetical. A generalized scan of the current file ("statement references a column that a later `ADD COLUMN` creates") still finds `idx_conversations_hidden`, `idx_sessions_parent`, `idx_conversations_codebase` (`WHERE deleted_at IS NULL`), `idx_workflow_runs_parent_conv`, and `idx_workflow_runs_last_activity` — and it is not only indexes: against a database created before `35217147`, moving all five still dies on `COMMENT ON COLUMN remote_agent_sessions.parent_session_id`. #2443 covers two of the five. The PR author scoped this deliberately and offered the generalization as a follow-up; that is the right call for this PR |
| `prp-core:seam-analyzer` | Nothing in CI ever applies `migrations/000_combined.sql` to a real Postgres, on any PR — which is exactly why this shipped green | No `postgres` service, `DATABASE_URL`, or compose usage in `.github/workflows/`; `packages/core/src/db/adapters/postgres.test.ts:24-39` mocks `pg` entirely, so the SQL text is never executed; `bundled-schema.test.ts:16-25` only compares the generated constant to `getSchemaSQL()` | A job that applies the schema to a snapshot from an older release tag before applying the current one would have caught this and would catch the five latent siblings. Two independent reporters (#2508's author and `tazmon95`) reached the same conclusion from production outages |

## Strengths

- The deletion is provably lossless rather than a crash traded for a missing index: the additive block's `CREATE UNIQUE INDEX` was byte-identical, and fresh-vs-upgraded schema comparison confirms both paths converge on the same columns and indexes.
- The guard test is not a tautology — it fails against the unfixed SQL at the `offset > addColumn` assertion, and it pins the inline `event_order BIGINT` in the `CREATE TABLE` body so a later cleanup cannot silently break fresh installs.
- SQLite needed no matching change and got none: `sqlite.ts:755-761` already carries the same reasoning, with `migrateColumns()` (`sqlite.ts:437-465`) creating the index after its `ALTER TABLE`. The 22-test schema-parity suite passes untouched.
- The PR body's byte-offset evidence (`position: 12720`) lands exactly on the removed statement, and `tazmon95` independently reproduced the failure and the fix on a restored production dump, including a booted build of `baa93763` reaching `server_ready` on a fixture that fails on `0.8.0` and `0.7.1`.

## Verdict

**READY TO MERGE**

No Critical or Important findings; all validation passed, including a real-Postgres reproduction of the crash on `dev` and a clean upgrade on this branch. Both suggestions are follow-up work that this PR is right not to absorb: generalizing the guard would currently fail on the five siblings that #2443 and a further pass must fix, and the CI gap is its own change. Merge this, then land #2443 and take the remaining three instances plus the `COMMENT ON COLUMN` case in a single structural pass with the generalized guard.

PR #2518, opened 2026-08-09
feat(core,workflows): per-project default workflow dispatch

hidegh:
Adds an opt-in `dispatch:` table to global config that makes a registered project convention-based: every plain message in a conversation bound to that project runs one named workflow instead of asking the AI router to pick. Slash commands are untouched, and a configurable sigil escapes a single message back to normal AI chat.

    dispatch:
      githubName/githubRepo: assignedWorkflow
    dispatchSigil: '? '   # optional; this is the default

Enforced at handleMessage(), the single seam every source already funnels through, so Slack, Telegram, Discord, GitHub, web and CLI all get it with zero per-adapter code. Execution reuses handleWorkflowRunCommand -- the same path `/workflow run` takes -- so isolation resolution, `requires:` gates and resume detection behave identically however the workflow was chosen.

The routing policy lives in a pure, I/O-free module (orchestrator/ dispatch.ts), so it is unit-testable without a DB or an adapter: 34 tests. The sigil is passed in as a parameter rather than read from config there, which is what keeps that property.

Second, smaller change: attachments never reached the workflow engine at all -- `attachedFiles` only ever became prose in the AI prompt. They are now threaded through ExecuteWorkflowOptions and delivered to `bash:` and `script:` nodes as ARCHON_ATTACHMENTS, a JSON array that is ALWAYS set (`[]` when empty) so a script can JSON.parse it without a presence check. This applies to every run, not only dispatched ones.

Design decisions worth recording:

- `dispatch:` and `dispatchSigil:` are global-only. The table's keys are install-level project names that a repo's own config cannot know, and folder projects have no repo to hold a config file at all. The sigil is meaningless without the table it escapes from, so the two travel together.
- The sigil defaults to `"? "` -- with the trailing space. In an intake thread a message that merely opens with a question mark is far likelier to be content than an instruction to the router, so the escape must be typed as a separated prefix to count. `?urgent note`, `??` and a lone `?` all dispatch; `? explain this` escapes.
- Requiring a separator forced the match to move from `message.trim()` to a leading-whitespace-only strip: a sigil ending in a separator can never prefix an already right-trimmed string.
- An empty or whitespace-only `dispatchSigil:` is rejected in favour of the default. `'x'.startsWith('')` is true, so honouring an empty prefix would route every message to the AI and silently disable dispatch install-wide.
- The sigil's default is applied at the consumer (resolveDispatchSigil), not at config-merge time, so an unset key stays distinguishable from one explicitly set to the default.
- The project mapping is checked BEFORE the sigil. The reverse order would make every install silently swallow a leading sigil even with no `dispatch:` configured -- a behaviour change for threads this feature should not touch.
- An unresolvable or ambiguous workflow name is reported in-thread and nothing runs. It never falls through to the AI router, because a plausible-looking AI reply is the hardest failure to notice from inside a chat thread (Fail Fast + Explicit Errors). The in-thread message quotes the install's configured sigil, not a hardcoded `?`.
- An open approval gate still wins over dispatch.
- Project keys match exactly first, then case-insensitively, so a capitalisation slip in hand-written YAML cannot silently disable the feature for that project. The sigil is matched case-sensitively -- it is retyped on every message, so matching what was configured is predictable.
- ARCHON_ATTACHMENTS is merged last so a stale operator-set env var of the same name cannot shadow the real one.

Two test-only adjustments were forced by the change: the dispatch config read consumes a single-shot `loadConfig` mock queue in orchestrator-agent.test.ts (switched to a persistent mock, which the existing beforeEach already resets), and one executor envVars assertion now accounts for the always-present key.

Adds an opt-in `dispatch:` table to global config that makes a registered project convention-based: every plain message in a conversation bound to that project runs one named workflow instead of asking the AI router to pick. Slash commands are untouched, and a leading `?` escapes a single message back to normal AI chat.

    dispatch:
      githubName/githubRepo: assignedWorkflow

Enforced at handleMessage(), the single seam every source already funnels through, so Slack, Telegram, Discord, GitHub, web and CLI all get it with zero per-adapter code. Execution reuses handleWorkflowRunCommand -- the same path `/workflow run` takes -- so isolation resolution, `requires:` gates and resume detection behave identically however the workflow was chosen.

The routing policy lives in a pure, I/O-free module (orchestrator/ dispatch.ts) modelled on slack-channel-context.ts, so it is unit-testable without a DB or an adapter: 20 tests, 100% line coverage.

Second, smaller change: attachments never reached the workflow engine at all -- `attachedFiles` only ever became prose in the AI prompt. They are now threaded through ExecuteWorkflowOptions and delivered to `bash:` and `script:` nodes as ARCHON_ATTACHMENTS, a JSON array that is ALWAYS set (`[]` when empty) so a script can JSON.parse it without a presence check. This applies to every run, not only dispatched ones.

Design decisions worth recording:

- `dispatch:` is global-only. Its keys are install-level project names that a repo's own config cannot know, and folder projects have no repo to hold a config file at all -- the same rationale as `slack.channelProjects`.
- The project mapping is checked BEFORE the `?` sigil. The reverse order would make every install silently swallow a leading `?` even with no `dispatch:` configured -- a behaviour change for threads this feature should not touch.
- An unresolvable or ambiguous workflow name is reported in-thread and nothing runs. It never falls through to the AI router, because a plausible-looking AI reply is the hardest failure to notice from inside a chat thread (Fail Fast + Explicit Errors).
- An open approval gate still wins over dispatch.
- Project keys match exactly first, then case-insensitively, so a capitalisation slip in hand-written YAML cannot silently disable the feature for that project.
- ARCHON_ATTACHMENTS is merged last so a stale operator-set env var of the same name cannot shadow the real one.

Two test-only adjustments were forced by the change: the dispatch config read consumes a single-shot `loadConfig` mock queue in orchestrator-agent.test.ts (switched to a persistent mock, which the existing beforeEach already resets), and one executor envVars assertion now accounts for the always-present key.

## Summary

- Problem:
- Why it matters:
- What changed:
- What did **not** change (scope boundary):

### Before

```
(Draw the user-facing flow BEFORE this PR. Show each step the user takes.)

Example:
  sends message ──────▶  resolves session
                         loads context
                         streams to AI ──────────▶ processes prompt
                         receives chunks ◀──────── streams response
  sees reply ◀─────────  sends to platform
```

### After

```
(Draw the user-facing flow AFTER this PR. Highlight what changed with [brackets] or asterisks.)
```

### Before

```
(Map ALL modules touched or connected to this change. Draw lines between them.)
```

### After

```
(Same diagram with changes highlighted. Mark new modules with [+], removed with [-],
 modified with [~]. Mark new connections with ===, removed with --x--.)
```

| | | unchanged / **new** / **removed** / **modified** | |

## Label Snapshot

- Risk: `risk: low|medium|high`
- Size: `size: XS|S|M|L|XL`
- Scope: `core|workflows|isolation|git|adapters|server|web|cli|paths|config|docs|dependencies|ci|tests|skills`
- Module: `<scope>:<component>` (e.g. `workflows:executor`, `adapters:slack`, `core:orchestrator`)

## Change Metadata

- Change type: `bug|feature|refactor|docs|security|chore`
- Primary scope: `core|workflows|isolation|git|adapters|server|web|cli|paths|multi`

## Linked Issue

- Closes #
- Related #
- Depends on # (if stacked)
- Supersedes # (if replacing older PR)

Commands and result summary:

```bash
bun run type-check
bun run lint
bun run format:check
bun run test
# Or all at once:
bun run validate
```

- Evidence provided (test/log/trace/screenshot):
image

- If any command is intentionally skipped, explain why:

## Security Impact (required)

- New permissions/capabilities? (`Yes/No`)
- New external network calls? (`Yes/No`)
- Secrets/tokens handling changed? (`Yes/No`)
- File system access scope changed? (`Yes/No`)
- If any `Yes`, describe risk and mitigation:

## Compatibility / Migration

- Backward compatible? (`Yes/No`)
- Config/env changes? (`Yes/No`)
- Database migration needed? (`Yes/No`)
- If yes, exact upgrade steps:

- Verified scenarios:
- Edge cases checked:
- What was not verified:

- Affected subsystems/workflows:
- Potential unintended effects:
- Guardrails/monitoring for early detection:

## Rollback Plan (required)

- Fast rollback command/path:
- Feature flags or config toggles (if any):
- Observable failure symptoms:

## Risks and Mitigations

List real risks in this PR (or write `None`).

- Risk:
  - Mitigation:

## Summary by CodeRabbit

- **New Features**
  - Added configurable default workflow dispatch for project messages.
  - Added an escape sigil (default `? `) to send messages to normal chat instead.
  - Preserved attachments when routing messages through workflows.
  - Exposed incoming attachments to workflow scripts through `ARCHON_ATTACHMENTS`.
  - Added handling for missing, invalid, ambiguous, or unlisted workflow mappings.

- **Documentation**
  - Documented dispatch configuration, routing behavior, sigil customization, restart behavior, and attachment availability.

### DIFF
--- a/packages/core/src/config/config-loader.ts
+++ b/packages/core/src/config/config-loader.ts
@@ -222,6 +222,13 @@ const DEFAULT_CONFIG_CONTENT = `# Archon Global Configuration
 # Concurrency settings
 # concurrency:
 #   maxConversations: 10
+
+# Per-project default workflow (global only, all platforms).
+# Every non-slash message in a conversation bound to a listed project runs that
+# workflow instead of the AI router. Prefix a message with "? " to ask the AI instead.
+# dispatch:
+#   githubName/githubRepo: assignedWorkflow   # <registered project name>: <workflow name>
+# dispatchSigil: '? '   # escape prefix; quote it — the trailing space is part of it
 `;
 
 /**
@@ -513,6 +520,18 @@ function mergeGlobalConfig(defaults: MergedConfig, global: GlobalConfig): Merged
   // Container backend defaults (folder projects)
   if (global.container) {
     result.container = { ...global.container };
+  }
+
+  // Project → default-workflow interception table. Global-only by design (the
+  // keys are install-level project names), so there is no matching branch in
+  // mergeRepoConfig.
+  if (global.dispatch) {
+    result.dispatch = { ...global.dispatch };
+  }
+  // Passed through raw — the `"? "` default lives in `resolveDispatchSigil`,
+  // not here, so an unset key stays distinguishable from an explicit one.
+  if (global.dispatchSigil !== undefined) {
+    result.dispatchSigil = global.dispatchSigil;
   }
 
   return result;
--- a/packages/core/src/config/config-types.ts
+++ b/packages/core/src/config/config-types.ts
@@ -178,6 +178,51 @@ export interface GlobalConfig {
    * overrides these per-field.
    */
   container?: ContainerConfig;
+
+  /**
+   * Per-project message interception: `<registered project>: <workflow name>`.
+   *
+   * Every non-slash message in a conversation bound to a listed project goes
+   * straight to that workflow, bypassing the AI router. Projects not listed
+   * here are untouched and keep normal routing.
+   *
+   * Global-only by design: the keys are install-level project names (the ones
+   * used with `/register-project <name>` and `/setproject <name>`), which a
+   * repo's own config cannot know — and folder projects have no repo to hold a
+   * config file at all.
+   *
+   * Platform-agnostic: enforced at the single `handleMessage` intake seam, so
+   * it applies identically to Slack, Telegram, Discord, GitHub, web, and CLI.
+   *
+   * Escape hatches, in precedence order:
+   *   1. `/command`  — slash commands route exactly as they do today
+   *   2. `? message` — the `dispatchSigil:` prefix falls through to normal AI chat
+   *
+   * A listed project whose workflow cannot be resolved is reported to the user,
+   * never silently fallen through to the AI router (Fail Fast — a config typo
+   * must not quietly restore the old behavior).
+   *
+   * @example
+   * dispatch:
+   *   githubName/githubRepo: assignedWorkflow
+   */
+  dispatch?: Record<string, string>;
+
+  /**
+   * Prefix that opts a single message out of `dispatch:` and back into normal
+   * AI chat/routing. Defaults to `"? "` — note the trailing space, which is
+   * part of the sigil and is what makes the escape deliberate rather than
+   * accidental (`?!` and `?typo` are content, `? explain this` is a question).
+   *
+   * Global-only, for the same reason as `dispatch:` itself: it is meaningless
+   * without the table it escapes from, and the two must be read together.
+   *
+   * Quote it in YAML — bare trailing whitespace is stripped by the parser:
+   *
+   * @example
+   * dispatchSigil: '? '
+   */
+  dispatchSigil?: string;
 }
 
 /**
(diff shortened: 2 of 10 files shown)


hidegh:
Better structured replacement on it's way.

Issue #2115, opened 2026-07-15
security(workflows): script nodes raw-splice user-controlled text into executable source — harden like bash nodes


Wirasm:
## Problem

`executeBashNode` substitutes variables with `shellSafe: true` (`dag-executor.ts` bash path → `executor-shared.ts:446`): user-controlled values (`$ARGUMENTS`, `$USER_MESSAGE`, `$LOOP_USER_INPUT`, `$REJECTION_REASON`, `$CONTEXT` family) are NEVER spliced into the script text — they're delivered as subprocess **env vars** precisely to prevent injection.

`executeScriptNode` does NOT pass `shellSafe` (`dag-executor.ts:~2043-2051`), so the same user-controlled values ARE text-substituted **raw and unescaped** directly into the TypeScript/Python source that `bun -e` / `uv run python -c` executes. Script subprocesses also don't receive the equivalent env vars (only `ARTIFACTS_DIR`, `LOG_DIR`, `BASE_BRANCH` + managed project env), so the raw splice is the only channel — and it's a code-injection channel.

## Impact

A workflow with a `script:` node whose body references `$ARGUMENTS` (or $CONTEXT — which platforms populate from **GitHub issue/PR text**) executes attacker-influenced text as code. Example: a GitHub issue titled `"); require('child_process').execSync('...') //` flowing into an issue-triage workflow's script node. Bash nodes closed this exact hole; script nodes shipped without it.

## Proposal

Mirror the bash design:
1. Pass `shellSafe`-equivalent behavior for script bodies: skip literal substitution of user-controlled variables.
2. Inject them as subprocess env vars instead (`ARGUMENTS`, `USER_MESSAGE`, `CONTEXT`, `LOOP_USER_INPUT`, `LOOP_PREV_OUTPUT`, `REJECTION_REASON`) so `process.env.ARGUMENTS` / `os.environ['ARGUMENTS']` becomes the sanctioned channel — matching bash-node ergonomics.
3. `$nodeId.output` refs keep the current raw-value substitution (documented, needed for the direct-assignment pattern) — the strict producer contract bounds those values.
4. Migration: a loader/runtime WARN for script bodies containing literal user-controlled variable refs for one release, then removal.

## Context

Found by multi-agent review of PR #2094; currently only **documented** as a footgun in the archon skill (`references/variables.md`). Related security precedent: the target-repo .env strip + bash shellSafe guard (env-isolation invariant).

PR #2523, opened 2026-08-10
feat(workflows): structural workflow signature (inputs/returns/with)

### LINKED ISSUE #2470: feat(workflows): structural workflow signature — inputs: + returns: + with: on workflow: nodes (signature phase 2)
Phase 2 of #2453, and the phase that actually pays for the whole design. Phase 1 (#2466) ships the `include:` load-time macro; this ships the **structural signature** — a block declares what it takes and what it returns.

## Why this phase is worth more than phase 1

Two things, both real, and neither waits on phase 3:

**1. `returns:` works for every node type immediately — including `command:` nodes.** It selects an **id**, not text, so `archon-review-block` can declare `returns: synthesize` on day one even though all nine of its command bodies stay un-parameterisable. That alone kills the positional-accident defect for the flagship block. (Today you *cannot* return `synthesize`, because `implement-fixes` depends on it and the engine picks the first sink in definition order.)

**2. `$INPUTS` for `workflow:` sub-runs reaches command bodies for free.** This is the structural point. A sub-run's `$INPUTS` is a **runtime** variable on the child run, and every child node — command or prompt — already flows through `buildPromptWithContext` → `substituteWorkflowVariables`. **The command-body gap is `include:`-only.** And the decomposition #2123 actually wants — `fix-github-issue-experimental` split into parse-input → research → plan → implement → review — is a *sub-run* decomposition. **Phase 2 delivers it whole.**

## What's already landed

- **#2459** — the `parseWorkflow` field-parity ratchet. Task 10 adds two schema fields; the guard fails until they reach the assembly literal, which is exactly the six-week `requires:` bug it was built for.
- **#2460** — the `output_format` inversion fix (was sequenced as Task 8b). Merged standalone so a user-visible behaviour change got its own revert handle. Scope was held to the *inversion*, not the *absence of validation* — a child emitting schema-mismatched JSON still passes, and that half belongs with `returns:` here.
- **#2224** — fan-out. Which changes one task below from conditional to required; see Task 14.

## Tasks

Anchors are symbols and quoted strings. Line numbers in the plan have already drifted once.

**9 · `schemas/workflow.ts` — declare `inputs:` and `returns:`.** Add `workflowInputSpecSchema` (`required?`, `default?`, `description?`, all optional) and on `workflowBaseSchema`: `inputs: z.record(z.string(), workflowInputSpecSchema).optional()` and `returns: z.string().min(1).optional()`. Derive the type with `z.infer` — never a parallel hand-written interface.

⚠️ `required: true` together with `default:` is contradictory. Reject it in the **loader** (Task 10), not the schema — the schema is per-value; the loader is where warn-and-drop lives.

**10 · `loader.ts` — parse and validate them.** Warn-and-drop blocks for `inputs` (per-key `safeParse`, drop a failing key, reject `required`+`default`) and `returns` (non-empty string). Add both to `parseWorkflow`'s returned object literal via conditional spread. Then: if `returns` is set it **must** name a top-level node id, else `Workflow declares returns: '<x>' but no top-level node has that id`.

⚠️ **A field in the schema but absent from `parseWorkflow`'s literal is silently discarded.** Task 9 alone does nothing. #2459's ratchet now catches this — when it goes red, add the fixture *and* the assembly line; don't weaken the test.

⚠️ `validateDagStructure` takes `nodes`, not the workflow, and is called recursively for `loop_group` bodies and from the expander — neither has a `returns` to check. Do the check in `parseWorkflow` after it returns rather than widening the signature.

**11 · `schemas/workflow.ts` — register the nested key spec.** Add `['inputs', { kind: 'record', entry: { kind: 'object', keys: new Set(Object.keys(workflowInputSpecSchema.shape)) } }]` to `KNOWN_WORKFLOW_NESTED_KEYS`, mirroring the existing `agents` record precedent.

⚠️ **Conditional on #2455.** `KNOWN_WORKFLOW_NESTED_KEYS` and `NestedKeySpec` **do not exist on `dev`** — verified again at the time of filing. They live on `fix/2213-surface-unknown-key-warnings`. If that hasn't merged when phase 2 starts, **skip this task and note it**; nothing else depends on it.

⚠️ Top-level keys need **no** registration — `KNOWN_WORKFLOW_KEYS` derives from `Object.keys(workflowDefinitionSchema.shape)`, so Task 9 registers them automatically. Adding a hand-maintained top-level list would be the exact drift the derivation prevents.

**12 · `include-expander.ts` — `returns:` drives `primarySink`.** When the child declares `returns`, `primarySink = prefix + returns`; otherwise the existing first-sink rule. **`sinks` is unchanged** — a declared return must not change which nodes downstream waits on.

⚠️ That split is the whole point: `depends_on: [blk]` means "after everything in the block", `$blk.output` means "the block's result". Conflating them makes a `returns:` on a mid-graph node truncate the wait.

⚠️ `returns:` may name a **non-sink** node. That's a feature — do not validate it against the sink set.

⚠️ The module doc's "first sink in definition order — the same rule loop_group uses" becomes wrong. Update it, and note loop_group's rule is deliberately unchanged.

**13 · `include-expander.ts` — check `with:` against declared `inputs:`.** Apply defaults for omitted names; error on an unsupplied required input; error on a caller key the block doesn't declare.

⚠️ **Undeclared blocks keep working, byte-for-byte.** A block with no `inputs:` falls back entirely to phase-1 behaviour. Extra-key checking activates **only** when `inputs:` is present. This is the compatibility hinge for the entire plan.

**14 · `schemas/dag-node.ts` — accept `with:` on `workflow:` nodes, reject two collisions.** (a) Replace the `workflow:` `with:` rejection with the same shape validation phase 1 applied to includes. (b) Reject `with:` and `input:` on one node — two input channels with a precedence rule is the constitution's smell #5. (c) Reject `fan_out.as` naming a key that is also a `with:` key. (d) Add `with` to `workflowNodeSchema` and the transform's workflow branch.

⚠️ **(c) is no longer conditional — #2224 merged.** `fan_out` is on `dev` now, and it shipped *with* an explicit rejection whose message names this issue: `dag-node.ts:835-841` refuses `fan_out.as` with `"'fan_out.as' (the $INPUTS channel) is not yet supported (PR-B, #2214). Remove it — "`. **Phase 2 is what lifts that rejection**, so implement (c) here rather than deferring it, and delete the placeholder message rather than editing around it.

**15 · `loader.ts` — scan `with:` values for refs.** Push every `with:` value of a `workflow:` node into `sources`, verbatim, mirroring how `workflow.input` is scanned.

⚠️ An `include:` node's `with:` values need **no** scan — the macro inlines them, and post-expansion `validateDagStructure` catches a dangling ref there (proved: `Node 'review__judge' references unknown node '$nosuch.output'`). A `workflow:` node's values are never inlined anywhere, so they need the explicit scan.

⚠️ This makes the **"KEEP IN SYNC (three ref-surface enumerations must agree)"** comment a **four**-way contract: the loader scan, `rewriteNodeOutputRefs`, the `substituteNodeOutputRefs` call sites, and phase 1's `applyInputsMacro`. Update that comment. #2450 showed what a silent break of this contract costs.

**16 · `executor-shared.ts` — `$INPUTS.<name>` as a runtime variable.** New optional `inputs` in `substituteWorkflowVariables`. An unknown name **throws**, matching the strict `$node.output.field` posture — never substitute `''`.

⚠️ Put it in the **`!shellSafe`** branch. This is a **security** decision, not style: a sub-run's `$INPUTS` values can derive from AI output (`with: {plan: $plan.output}`), which is precisely the user-controlled class `shellSafe` exists to keep out of shell source (#2115).

⚠️ The parameter list is already 11 positional args. Add to the `options` bag, not a 12th slot.

**17 · `dag-executor.ts` — env delivery for bash/script.** `INPUTS_<UPPER_SNAKE(name)>` in `subprocessEnv`, spread **after** `...(envVars ?? {})` so engine-reserved keys win. Reject at load any input name that doesn't mangle to a unique env key.

⚠️ The mangling is not self-evident. Document it beside the bash-node section of the authoring guide, or authors will write `$INPUTS.plan` in a bash body and get nothing.

**18 · `executor.ts` — persist `metadata.inputs` at spawn.** Resolve the node's `with:` map exactly as `input` is resolved, pass it to `createWorkflowRun` as `metadata: { parent_node_id, inputs }`, thread it into `ExecuteWorkflowOptions`, and read it back on resume.

⚠️ Persisting at spawn is what makes resume work — a child resumed cold reconstitutes `$INPUTS` from `metadata.inputs` rather than re-resolving parent refs that may no longer be in scope. #2214 names this as an acceptance criterion.

⚠️ There are **two** resume paths (`hydrateResumableRun` and the `resumeWorkflowRun` fallback). Both must surface `metadata.inputs`. A miss on the second only shows up for a failed child with zero completed nodes.

**19 · `dag-executor.ts` — `returns:` for child runs.** When the workflow declares `returns` **and** `parent_run_id` is set, take that node's output instead of scanning sinks. If it didn't complete or produced blank output, thread `''` and WARN — **do not** fall through to the sink scan, which would resurrect the positional accident under a new name.

⚠️ Gating on `parent_run_id` is deliberate: a top-level run's summary is a chat/CLI affordance, not a contract with a caller.

⚠️ The `loop_group` terminal computation is byte-identical code and is **deliberately not changed**. Leave a comment there pointing at this decision, or the next reader will "fix" the inconsistency.

**20 · Bare-run guard.** A workflow declaring unsatisfiable required inputs still **loads and lists** normally; invoking it top-level fails before any worktree/clone/AI cost, naming the missing inputs and pointing at `include:`/`workflow:`. Mirror `assertWorkflowRequirementsMet`'s placement.

⚠️ It stays listed on purpose. Hiding it from `/workflow list` would make blocks invisible to the builder UI and to discovery debugging.

**21 · `validator.ts` — bundled-set-only `workflow:` target check.** For each `workflow:` node in a **bundled** workflow, verify the target resolves within the bundled set. `archon validate workflows` / CI only — do not touch `parseWorkflow` or `validateDagStructure`.

⚠️ **Sub-run targets stay runtime-resolved.** Spawn-time resolution is the affordance that lets a run author its own child mid-flight; the constitution records that a load-time existence check *"compiles, passes every existing test, and silently destroys the capability"*. The bundled-set check is the narrow substitute.

⚠️ Runtime resolution is **fuzzy** (exact → case-insensitive → suffix → substring). The bundled check must use the same resolver, or a legal suffix reference gets reported as broken.

**22 · Docs.** A "workflow signature" section covering `inputs:`, `returns:`, `with:`, `$INPUTS`, and — prominently — the **binding-time table**. State plainly that `$INPUTS` reaches `command:` bodies for sub-runs but **not** for includes, and why. Constitution: record `inputs:`/`returns:` as admitted, and record explicitly that **cross-file schema checking was considered and rejected**.

## Constraints that must survive

- **Undeclared workflows keep working byte-for-byte.** Every existing workflow has no `inputs:`/`returns:`; nothing about their behaviour may change.
- **`returns:` and `output_format` stay independent.** `returns:` decides *which* output; `output_format` decides *how fields are read from it*. No cross-file schema coupling.
- **The #2459 parity ratchet stays green** — add fixtures, never weaken the test.
- **Constitution**: this coordinates, it does not compute. Declarative data the engine needs to wire and validate composition. Cite it in the PR.

## Not in scope

Phase 3's runtime argument map, which is what finally parameterises `command:` bodies **for includes** (#2123's flagship block stays un-parameterised until then) · cross-file schema validation between a caller's `output_format` and a callee's `returns` · validating a `workflow:` child's output against a declared schema (the surviving half of #2456) · any expression capability in `with:` values.

## Acceptance

- A command-only block declares `returns: <id>` and `$blk.output` resolves to that node — including when it is **not** a sink
- A two-input sub-run child resolves **both** `$INPUTS` values inside a **`command:`** body
- `with:` on a `workflow:` node is accepted; `with:`+`input:` together is rejected; `fan_out.as` colliding with a `with:` key is rejected, and the `dag-node.ts:835` placeholder rejection is gone
- A child resumed cold reconstitutes `$INPUTS` from `metadata.inputs`
- A bare top-level run of a required-input block fails before any worktree/clone/AI cost
- A workflow with no `inputs:`/`returns:` behaves byte-for-byte as before
- `bun run validate` green — grep the log for `0 fail` rather than trusting the wrapper's exit code

Part of #2453. Depends on #2466 (phase 1). Task 11 conditional on #2455. Unblocks the #2123 sub-run decomposition. Related: #2214, #2449, #2450.

leex279:
## Summary

- **Problem:** The workflow language had no declarative *signature*. A workflow couldn't declare what it takes (`inputs:`) or which node's output is its result (`returns:`), and callers could only pass a single opaque `input:` string to a `workflow:` sub-run node — no named, validated arguments.
- **Why it matters:** Reusable sub-workflows need a contract. Named inputs let a parent bind values by name, load-time validation catches missing/undeclared args before any worktree/clone/AI cost, and `returns:` gives a stable, author-chosen result channel instead of guessing the terminal sink.
- **What changed:** Adds `inputs:`/`returns:` to `workflowBaseSchema`, accepts `with:` on `workflow:` sub-run nodes (mirroring `include:`), and introduces the strict `$INPUTS.<name>` runtime variable plus `INPUTS_<UPPER_SNAKE>` env for bash/script nodes. This **coordinates, it does not compute** (Workflow Language Constitution — cited): the fields are declarative signature data the engine needs to govern binding.
- **What did NOT change (scope boundary):** No change to `include:` undeclared-key behavior when `inputs:` is absent (byte-for-byte). No cross-file schema type checking (explicitly rejected in the constitution). `loop_group` per-iteration scan deliberately unchanged. `until_bash` env delivery unchanged.

### Before

```
Author                 Archon Engine
──────                 ─────────────
writes workflow:  ────▶ workflow has no signature
  node input: "x"      only a single opaque `input:` string
                       terminal sink guessed as result
caller runs      ────▶ no named-argument validation
                       undeclared/missing args unnoticed until runtime
```

### After

```
Author                 Archon Engine
──────                 ─────────────
declares          ────▶ *inputs:* (name/required/default) validated at load
  inputs: / returns:    *returns:* → primarySink / child terminal output
writes workflow:  ────▶ *with:* {name: value} accepted on workflow: nodes
  node with: {...}      reject with+input, fan_out.as↔with collision
caller runs      ────▶ *bare-run guard*: unsatisfiable required inputs
                       fail BEFORE worktree/clone/AI cost
node body        ────▶ *$INPUTS.<name>* strict substitution (did-you-mean hint)
                       *INPUTS_<UPPER_SNAKE>* env for bash/script
resume (cold)    ────▶ metadata.inputs reconstitutes $INPUTS
```

### Before

```
schemas/workflow.ts ──▶ loader.ts ──▶ include-expander.ts ──▶ dag-executor.ts
schemas/dag-node.ts ──▶ loader.ts                             executor-shared.ts (substitution)
                                                              executor.ts (spawn)
validator.ts        (target checks)
utils/workflow-requirements.ts (requires: gate)
```

### After

```
[~] schemas/workflow.ts      ── inputs:/returns: on workflowBaseSchema, inputEnvKey
[~] schemas/dag-node.ts      ── with: on workflow nodes, collisions, INPUT_NAME_PATTERN
[~] schemas/workflow-run.ts  ── SUBRUN_METADATA_KEYS.inputs
[~] loader.ts          ===▶  parse/validate inputs/returns, env-key collision, with: ref scan
[~] include-expander.ts ==▶  returns:→primarySink, with: vs declared inputs:
[~] executor-shared.ts  ==▶  $INPUTS strict substitution (!shellSafe)
[~] dag-executor.ts     ==▶  resolveRunInputs, INPUTS_* env, with: resolution, returns child output
[~] executor.ts         ==▶  persist metadata.inputs at spawn
[~] validator.ts             bundled-set-only workflow: target check (fuzzy resolver)
[~] utils/workflow-requirements.ts  assertWorkflowInputsSatisfiable
[~] cli/workflow.ts + core/orchestrator-agent.ts  bare-run guard call sites
[~] docs: authoring-workflows.md, workflow-language-constitution.md
```

**Connection inventory:**

| loader.ts | schemas/workflow.ts | **modified** | parses/validates `inputs:`/`returns:` |
| include-expander.ts | schemas/workflow.ts | **modified** | `returns:` drives primarySink; `with:` vs `inputs:` |
| dag-executor.ts | executor-shared.ts | **modified** | `$INPUTS.<name>` strict substitution |
| dag-executor.ts | schemas/workflow-run.ts | **new** | reads/writes `metadata.inputs` |
| cli/workflow.ts | utils/workflow-requirements.ts | **new** | bare-run input-satisfiability guard |
| core/orchestrator-agent.ts | utils/workflow-requirements.ts | **new** | bare-run input-satisfiability guard |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: L`
- Scope: `workflows`
- Module: `workflows:loader`, `workflows:executor`, `workflows:schemas`

## Change Metadata

- Change type: `feature`
- Primary scope: `workflows`

## Linked Issue

- Closes #2470
- Related #2224 (fan-out — `fan_out.as` placeholder rejection removed here), #2223 (per-child worktree isolation)

```bash
bun run type-check   # ✅ Pass (exit 0, all packages)
bun run lint         # ✅ Pass (exit 0, all packages)
```

- **Evidence provided:** `@archon/workflows` test suite passes (loader, include-expander, executor-shared, validator, workflow-requirements, schemas — new fixtures incl. the #2459 parity ratchet). `@archon/cli` `workflow.test.ts` and `@archon/core` `orchestrator-agent.test.ts` pass per-file (as CI runs them). `check:capability-matrix` OK.
- **If any command is intentionally skipped, explain why:** `check:bundled` fails ONLY due to a pre-existing untracked operator file (`.archon/workflows/defaults/archon-hyperframes-video.yaml`) that is **not part of this change** — no default files were added or modified by this feature. Suite-wide pre-existing/environmental failures (`load-command-prompt.test.ts` home-scope resolution, `codebases.test.ts`, CLI env/pi-config temp-dir tests, cross-file orchestrator mock pollution) also fail on the clean tree and are unrelated; each passes per-file.

## Security Impact (required)

- If any `Yes`, describe risk and mitigation: N/A. Note: `$INPUTS.<name>` substitution runs in the `!shellSafe` branch only and is strict (throws on unknown name with a did-you-mean hint) — it does not broaden shell-injection surface beyond existing `$node.output` substitution.

## Compatibility / Migration

- Backward compatible? `Yes` — all new fields are optional; workflows without `inputs:`/`returns:`/`with:` behave byte-for-byte as before. `include:` undeclared-key behavior is unchanged when `inputs:` is absent.
- Config/env changes? `No`
- Database migration needed? `No` — reuses the existing additive `workflow_runs.metadata` JSON column (new `inputs` key); no schema change.
- If yes, exact upgrade steps: N/A

- **Verified scenarios:** signature parse/validate; `returns:`→primarySink in include-expander; `with:` vs declared `inputs:` (defaults / missing-required / undeclared-key); `$INPUTS` runtime substitution; bundled-target check; bare-run guard for unsatisfiable required inputs; env-key mangling collision.
- **Edge cases checked:** reject `with`+`input` together; `fan_out.as`↔`with` collision; `returns:` blank child output → `''` + WARN (no sink fallthrough); cold-resume `$INPUTS` reconstitution from `metadata.inputs`; required-input block still lists/loads but a top-level bare run fails early.
- **What was not verified:** full end-to-end `bun run test` parallel suite and `bun run test:install` (per implementation note — pre-existing environmental failures on the clean tree); live multi-run fan-out `$INPUTS.<as>` channel exercised via unit fixtures rather than a real detached run.

- **Affected subsystems/workflows:** workflow loader, include-expander, DAG executor, executor-shared substitution, validator, workflow-requirements guard; CLI + orchestrator run entry points.
- **Potential unintended effects:** substitution and load-time validation are on hot paths; mitigated by keeping all new behavior gated on presence of `inputs:`/`returns:`/`with:` (absent → unchanged path). The four ref-surfaces (schema / loader / include-expander / runtime) carry a KEEP-IN-SYNC comment to prevent drift.
- **Guardrails/monitoring for early detection:** load-time rejections and WARN logs for env-key collisions / dangling refs; new tests including the #2459 parity ratchet fixtures fail CI on silent field drops.

## Rollback Plan (required)

- **Fast rollback command/path:** revert the single feature commit `55f0e7b2`. All fields are additive and optional, so revert is clean and low-risk.
- **Feature flags or config toggles (if any):** none — behavior is inert unless a workflow opts in via `inputs:`/`returns:`/`with:`.
- **Observable failure symptoms:** load-time errors on workflows declaring the new fields; `$INPUTS` substitution throwing at runtime; unexpected bare-run guard rejections.

## Risks and Mitigations

- **Risk:** four ref-surfaces (schema, loader, include-expander, runtime substitution) can drift.
  - **Mitigation:** explicit KEEP-IN-SYNC comment across all four + parity ratchet fixtures (#2459) that fail CI on dropped fields.
- **Risk:** `$INPUTS` strict substitution in the `!shellSafe` branch could surprise authors with a hard throw.
  - **Mitigation:** did-you-mean hint on unknown names; documented in the authoring guide's Workflow Signature section + binding-time table.
- **Risk:** lifting the `fan_out.as` placeholder rejection could leave `as` inert ("silently does nothing" trap).
  - **Mitigation:** `executeFanOutWorkflowNode` now resolves the static `with:` map and adds the per-item `$INPUTS.<as>` channel, so `as` is functional rather than accepted-and-ignored.

## Summary by CodeRabbit

* **New Features**
  * Workflows can now declare named inputs, defaults, required fields, descriptions, and a selected return node.
  * Child workflows support named `with:` inputs, including fan-out item values.
  * Inputs are available in prompts and scripts through `$INPUTS` variables and environment variables.
* **Bug Fixes**
  * Workflows with missing required inputs now fail before execution with clear guidance.
  * Invalid or undeclared inputs are rejected with validation errors.
* **Documentation**
  * Added guidance for workflow signatures, input handling, return selection, and composition.


leex279:
# 🔍 Comprehensive PR Review

**PR**: #2523 — feat(workflows): structural workflow signature (inputs/returns/with)
**Reviewed by**: 4 specialized agents (code-review, error-handling, test-coverage, docs-impact)
**Date**: 2026-08-10

> ⚠️ **4 of 5 dimensions ran.** The `comment-quality` agent artifact was not produced, so that dimension is unassessed — findings below do not fabricate it.

---

## Summary

A declarative **workflow signature** surface — workflow-level `inputs:`/`returns:`, `with:` on `workflow:`/`include:` nodes, and `$INPUTS.<name>` / `INPUTS_<UPPER_SNAKE>` delivery. High quality: Constitution admissibility documented and passed, `$INPUTS` kept out of shell source (env-var delivery), load-time fail-fast before cost, docs-site guide + constitution updated and matching the code. **No CRITICAL issues and no merge-blocking correctness bug.** The gaps are: the runtime half of the feature is untested, two now-false claims in CLAUDE.md, and the `workflow:` **sub-run** path lacks the default-application / validation / env-key-collision guards its `include:` sibling has.

**Verdict**: `REQUEST_CHANGES`

| Severity | Count |
|----------|-------|
| 🔴 CRITICAL | 0 |
| 🟠 HIGH | 3 |
| 🟡 MEDIUM | 5 |
| 🟢 LOW | 4 |

| Agent | 🔴 | 🟠 | 🟡 | 🟢 |
|-------|----|----|----|----|
| Code Review | 0 | 0 | 1 | 1 |
| Error Handling | 0 | 0 | 1 | 2 |
| Test Coverage | 0 | 1 | 2 | 1 |
| Docs Impact | 0 | 2 | 1 | 0 |

---

## 🟠 High Issues

### 1. Runtime `$INPUTS` delivery + `returns:` rebinding is entirely untested
📍 `packages/workflows/src/dag-executor.ts:280-291, 5447-5478, 8468-8523` — *test-coverage*

The feature's whole runtime surface (+141 lines) has no direct/integration test; the spawn → persist → resolve → substitute chain is verified only at its two ends. The `returns:` rebinding has three branches (top-level ignores it / blank threads `''` without sink fall-through / non-sink wins) that can each regress silently and corrupt the `$node.output` a parent threads back.

<details><summary>Recommended fix</summary>

Add cases to `subrun.test.ts`: a parent `workflow:` node with `with:` whose child has a `bash:` node echoing `$INPUTS_PLAN`, and a child with `returns:` on a non-sink node; assert the threaded `$node.output`.
</details>

### 2. CLAUDE.md states `with:` is rejected on `workflow:` nodes — the PR now accepts it
📍 `CLAUDE.md` line 453 — *docs-impact*

`loader.test.ts` flips `"rejects 'with:'"` → `"accepts 'with:' (#2470)"`. CLAUDE.md is the highest-authority agent instruction file; leaving the false claim makes agents refuse valid `with:` YAML. `retry:` **is** still rejected — split the claims, don't delete the sentence.

<details><summary>Recommended fix (replace the two clauses on line 453)</summary>

```markdown
... `input:` forwards a data string as the child's `$ARGUMENTS`, OR `with:` (an identifier-keyed
string map, mutually exclusive with `input:`) delivers named values as the child's `$INPUTS.<name>`
at runtime (bash/script child nodes read them as `INPUTS_<UPPER_SNAKE>` env vars); `retry:` is
rejected on workflow nodes; disallowed inside a `loop_group` body ...
```
</details>

### 3. CLAUDE.md says `fan_out.as` is "reserved and rejected at load" — the PR now accepts it
📍 `CLAUDE.md` line 453 — *docs-impact*

`as` is now generally accepted and names the per-item value `$INPUTS.<as>`; only a collision with a `with:` key is rejected. Same false-constraint class as #2.

<details><summary>Recommended fix (replace the `as` clause on line 453)</summary>

```markdown
... `all_success` for the genuinely dependent case), `as` names the per-item value as
`$INPUTS.<as>` inside each child (rejected only when it collides with a `with:` key of the same
name; when unset the item still travels as the child's `$ARGUMENTS`). Children are INDEPENDENT ...
```
</details>

---

### 4. `workflow:` sub-run declared `default:` silently not applied; missing-required / undeclared-key never validated
📍 `dag-executor.ts:5450-5472` + `executor.ts:830-839` — *error-handling*. **Options**: Fix at spawn (mirror `resolveIncludeInputs`) | Fix inside child run | Scope docs to include-only.

<details><summary>Details</summary>

`include:` applies defaults + validates via `resolveIncludeInputs()`; the `workflow:` sub-run path builds inputs from **only `node.with`** and never loads the child def, so declared `default:`s are ignored, unsatisfied `required` inputs run anyway (after cost), and undeclared keys pass. A child with `inputs: { mode: { default: fast } }` invoked with no `with.mode` throws at runtime `Unknown input '$INPUTS.mode'. … This run has no declared inputs.` — contradicting the author's YAML. Docs (`authoring-workflows.md:1065-1069`) already promise this for **both** paths. **Recommend Option A** (resolve defaults + validate at spawn) to restore parity and fail before cost.
</details>

### 5. Env-key collision unguarded on the caller `with:` / `fan_out.as` path (silent clobber in bash/script)
📍 `schemas/dag-node.ts:1485-1500`, `dag-executor.ts:154-157` — *code-review*. **Options**: Extend load-time guard (rec) | Runtime throw | Grammar split.

<details><summary>Details</summary>

The `envKeyOwners` guard runs only on *declared* `inputs:`, not the caller side. `with: { foo_bar: "static" }` + `fan_out.as: "foo-bar"` loads cleanly; at runtime both fold to `INPUTS_FOO_BAR`, last-wins silently, and a bash/script child gets the wrong value (prompt nodes match exact key, so it's invisible there). Narrow trigger (names differing only by `-`/`_`) → MEDIUM. **Recommend Option A**: extend the `dag-node.ts` superRefine to reject `with:`/`as` keys colliding under `inputEnvKey(...)` — load-time, symmetric with the declared-inputs guard.
</details>

### 6. Cold-resume reconstitution of `metadata.inputs` is untested end to end
📍 `executor.ts:833-841` + `dag-executor.ts:280-282` — *test-coverage*

<details><summary>Details</summary>

The Task-18 robustness claim (persist inputs to metadata, re-read on resume) has a tested reader but an untested **write** and **round-trip**. If the "stamp only when non-empty" guard or the metadata key drifts, a resumed child silently loses `$INPUTS` and throws mid-resume. **Recommend**: spawn a child with `with:`, assert persisted `metadata.inputs`, drive a cold resume, assert `$INPUTS` still resolves.
</details>

### 7. Orchestrator (chat) fail-fast gate for `WorkflowMissingInputsError` is untested
📍 `orchestrator-agent.ts:739-761` — *test-coverage*

<details><summary>Details</summary>

The chat surface's try/catch (discriminate → send message → return, else re-throw) has no test. A regression could proceed past the gate (wasted cost) or swallow an unrelated error as a "missing inputs" reply. **Recommend**: orchestrator test with mocked `sendMessage` asserting the user got `err.message` and no run started.
</details>

### 8. CLAUDE.md omits the new `inputs:`/`returns:` fields and `$INPUTS` variable
📍 `CLAUDE.md` lines 432–441 + 453 — *docs-impact*. Absence, not a false claim — fold into the #2 edit.

---

## 🟢 Low Issues

<details><summary>View 4 low-priority items</summary>

| `$INPUTS.<name>` now throws on a top-level run that never declared inputs (compat change) | `executor-shared.ts:642-654` | code-review | Accept fail-fast strictness (matches `$node.output.field`) + changelog note, OR scope `$INPUTS` substitution to runs carrying inputs |
| `workflow:` node `input:` surface omits the `inputs` bag — `$INPUTS.<name>` throws instead of resolving | `dag-executor.ts:5435-5448` | error-handling | One-line fix: pass `inputs: resolveRunInputs(parentRun)` on the `input:` resolution |
| `readSubrunMetadata` drops the whole inputs map on any non-string value, no log | `schemas/workflow-run.ts:172-188` | error-handling | Defensive-only; optional `warn` breadcrumb, else accept as-is |
| `validator.ts` bundled-target check — ambiguity branch untested | `validator.ts:407-418` | test-coverage | Add a bundled ref matching ≥2 names; assert "ambiguous within the bundled set" |

</details>

---

## ✅ What's Good

- **Constitution compliance clean** — declarative coordinate data; `when:`/`with:` grew no operators; constitution + authoring guide updated in the PR and matching the code.
- **Shell-safety preserved** — `$INPUTS` only in the non-shell branch; bash/script read `INPUTS_<UPPER_SNAKE>` env vars (#2115); spread order correct and commented.
- **No swallowed errors on hot paths** — unknown `$INPUTS` THROWS with a did-you-mean hint; fan-out resolution emits both `notify()` and `failResult()`; `returns:`-blank WARNs without sink fall-through.
- **Thorough behavioral load-time tests** — mutual exclusion, collision, non-existent `returns:`, dangling ref, passthrough parity; field-parity test extended to `inputs`/`returns`.
- **Zod conventions honored** — `workflowInputSpecSchema`, `z.infer` types, `z` from `@hono/zod-openapi`, two-arg `z.record`, re-exported; package boundaries intact.

---

## 📋 Suggested Follow-up Issues

| Title | Priority | Finding |
|-------|----------|---------|
| Sub-run `with:` parity: apply `default:`, validate missing-required / undeclared keys | P1 | #4 |
| Guard env-key collisions on caller `with:` / `fan_out.as` | P2 | #5 |
| Add runtime + cold-resume tests for `$INPUTS` delivery and `returns:` rebinding | P2 | #1, #6, #7 |

---

## Next Steps

1. ⚡ Auto-fix will address the 2 CLAUDE.md corrections (#2, #3, folding in #8) + the one-line `input:` parity fix.
2. 📝 Decide fix-now vs. follow-up for the sub-run parity (#4) and env-key collision (#5) gaps.
3. 🧪 Add at least one runtime/resume test (#1, #6, #7) before merge.
4. 🔁 Re-run the comment-quality pass (artifact missing) if it gates this merge.
5. ✅ Confirm pending `test (ubuntu/windows)` + `docker-build` CI pass.

---

*Reviewed by Archon comprehensive-pr-review workflow — 4 of 5 dimensions*
*Artifacts: `~/.archon/.../runs/47eb748d-8c22-400b-98ea-6391303cc8a8/review/`*

Wirasm (changes requested):
The workflow-signature direction looks useful, but this needs a few corrections before merge.

1. **Remove the unrelated HyperFrames workflow.** Commit `ba329f6` adds `.archon/workflows/defaults/archon-hyperframes-video.yaml`. It is outside this PR's stated scope and is the direct cause of both Ubuntu and Windows CI failures: `check:bundled` reports `bundled-defaults.generated.ts is stale`. Please remove the file rather than regenerating the bundle for it.

2. **Handle `workflow.with` in include expansion.** `rewriteNodeOutputRefs()` and `applyInputsMacro()` walk `workflow.input` and `fan_out.items`, but neither walks the values in `node.with`. Consequently, an included reusable block that calls a child workflow cannot forward its input or reference its own nodes correctly. A focused reproduction fails with:
   ```
   Node 'outer__call' references unknown node '$local.output'
   ```
   Please apply both `$INPUTS.<name>` substitution and included-node ref namespacing to every `with:` value.

3. **Enforce the resolved child workflow's declared input contract before spawning it.** The runtime `workflow:` path resolves the child definition but then persists the caller's map directly. It currently does not apply declared defaults, reject missing required inputs, or reject undeclared keys, despite the docs promising parity with `include:`. Resolve and validate inputs after child discovery and before isolation/worktree or run-row creation. Please also account for defaulted inputs on bare top-level runs, since runtime `$INPUTS` currently comes only from sub-run metadata.

4. **Update `CLAUDE.md`.** It still says `with:` is rejected on `workflow:` nodes and `fan_out.as` is reserved/rejected, which is no longer true.

Please add regression coverage for:
- an included block containing a `workflow:` node whose `with:` values use both `$INPUTS.foo` and a child-local `$node.output`;
- sub-run defaults, missing-required inputs, and undeclared inputs;
- runtime `$INPUTS` delivery, `returns:` rebinding, and cold-resume metadata reconstitution.

Local validation on this head: type-check, lint, and format pass; existing signature/schema/validator/sub-run tests pass; the workflow package fails on the accidental bundled-default file.

leex279:
Thanks — all four addressed, plus the merge conflict with `dev`.

**1. Removed the unrelated HyperFrames workflow.** `.archon/workflows/defaults/archon-hyperframes-video.yaml` is deleted rather than bundled, so `check:bundled` reports up to date again.

**2. `workflow.with` now walked in include expansion.** Both `rewriteNodeOutputRefs()` and `applyInputsMacro()` iterate every value in a `workflow:` node's `with:` map — included-node ref namespacing and `$INPUTS.<name>` substitution alike. Your reproduction now expands to `payload: '$outer__local.output'` instead of failing on `$local.output`.

**3. Child input contract enforced at runtime.** The contract is extracted into `workflow-inputs.ts` and both surfaces call it, so `include:` and `workflow:` accept identical maps by construction rather than by comment. On the runtime path it runs after child discovery and before isolation/worktree resolution and before the child run row is created — a violation leaves no orphan worktree and no doomed row — and the map persisted to `metadata.inputs` is the *resolved* one (defaults applied), so a cold resume sees what the child's `inputs:` promised. Declared defaults now also apply to bare top-level runs, which have no parent to stamp metadata.

**4. `CLAUDE.md` updated.** `with:` on `workflow:` nodes and `fan_out.as` are both documented as supported, and the workflow-level `inputs:`/`returns:` signature now has an entry.

**Coverage added:**
- `include-expander.test.ts` — an included block whose `workflow:` node's `with:` values use both `$INPUTS.<name>` and a child-local `$node.output`; plus a required input referenced only from a `with:` value.
- `subrun.test.ts` — sub-run declared defaults, missing-required, undeclared keys, and undeclared-child passthrough; defaults on a bare top-level run; runtime `$INPUTS` delivery into an AI prompt surface; `returns:` rebinding on a sub-run; cold-resume reconstitution of `metadata.inputs` through a gate.

4 of the 5 new sub-run contract tests fail against the previous head, so they pin the fix rather than the current behaviour.

**Validation:** `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, type-check, lint and format all pass; workflows/core/adapters/server/isolation/git/paths/web tests pass. Three local failures are pre-existing Windows-only environment issues also present on `dev` (symlink `EPERM` in `load-command-prompt.test.ts` and `providers` `pathKind`, and Windows `tar` path mangling in `serve.test.ts`) — untouched by this branch.

`dev` is merged in; the only conflict was the import block in `executor-shared.ts` (both sides kept).

Wirasm (changes requested):
---
pr: 2523
title: "feat(workflows): structural workflow signature (inputs/returns/with)"
author: "leex279"
reviewed: 2026-08-12T12:32:00Z
---

# PR Re-review: #2523 — Structural workflow signature

**Author**: @leex279  
**Branch**: `archon/thread-3318ed4f` → `dev`  
**Head reviewed**: `dc560035fd746db41908a1e454d9a085f570353c`  
**Files changed**: 24 (+1826/-93)

## Summary

The follow-up substantially addresses the previous review: the unrelated HyperFrames workflow is removed, `workflow.with` now participates in include macro/ref rewriting, child input contracts are enforced before isolation or run creation, `CLAUDE.md` describes the new behavior, and focused coverage now exercises runtime delivery, defaults, required/undeclared inputs, `returns`, and cold resume.

Two remaining correctness gaps can silently change the declared result contract, and the branch now conflicts with current `dev`. These should be resolved before merge.

## Implementation Context

No local implementation report or completed plan for this PR branch was found in the PRP store. The PR description and linked issue #2470 were used as the implementation contract.

## Important Findings

1. **`packages/workflows/src/include-expander.ts:626` — rewrite workflow-level `returns` when it names an included block.**

   The second pass rewrites node dependencies and node output references through `primarySinkByIncludeId`, but `raw.returns` is copied unchanged at line 644. A workflow such as `returns: blk` expands the `blk` include into `blk__result`/`blk__cleanup` while retaining `returns: blk`; there is no longer a node with that id. The child-run completion path consequently cannot read the declared return output and emits a blank result.

   Reproduction on this head:

   ```text
   {"returns":"blk","nodeIds":["blk__result","blk__cleanup"]}
   ```

   Build the expanded result with `returns: renameIncludeRef(raw.returns)` when present, and add coverage where an outer workflow declares an include node itself as `returns`.

2. **`packages/workflows/src/loader.ts:971` — reject present-but-invalid `returns` instead of dropping the contract.**

   Whitespace and non-string values currently log `invalid_workflow_returns_value_ignored` and still load the workflow with `returns` unset. This silently falls back to positional sink selection, so malformed contract syntax can return a different node while appearing successful. Both `returns: "   "` and `returns: { bad: true }` reproduced as `loaded: true` with no parse error.

   Treat every present invalid `returns` value as a `validation_error`; only trim and retain valid non-empty strings. Add tests for blank and non-string values.

3. **Refresh onto current `dev` and resolve the two real merge conflicts.**

   GitHub reports `CONFLICTING` / `DIRTY`. A read-only `git merge-tree` identifies conflicts in:

   - `CLAUDE.md`: #2497 made this a one-line pointer; this PR's instruction update must move into authoritative `AGENTS.md`.
   - `packages/cli/src/commands/workflow.ts`: #2530 added deterministic dry-run behavior on the same command path; retain both dry-run and required-input guards.

   Re-run validation after resolving against current `dev`.

4. **`packages/core/src/orchestrator/orchestrator-agent.ts:749` — do not add identifiers to the required-input informational log.**

   The project instructions say not to log PII. This outcome needs only `workflowName` and `missing`; `conversationId` and `userId` do not help diagnose the contract failure and unnecessarily identify the caller. Remove both from this new log record.

## Suggestions

- Update the stale `dagNodeFlatSchema.with` comment around `packages/workflows/src/schemas/dag-node.ts:715`; it still says workflow nodes reject and discard `with`.
- Clarify the constitution row: `include.with` contract violations are load errors, while `workflow.with` violations are detected at runtime before child spawn.
- Add the small coverage gaps identified in the latest review: invalid input-name/non-object `inputs` warn-and-drop behavior and a positive bundled-workflow target case.

## Validation Results

| Check | Status | Details |
|------|--------|---------|
| Full pre-PR validation | PASS | `bun run validate` completed successfully, including bundled checks, schemas, capability matrices, type-check, lint, format, install test, and isolated package tests |
| Build | PASS | `bun run build`; all workspace builds completed successfully |
| GitHub Actions | PASS | Docs, Ubuntu, Windows, Docker, and CodeRabbit are green on submitted head |
| Current-dev integration | FAIL | Conflicts in `CLAUDE.md` and `packages/cli/src/commands/workflow.ts` |

## Strengths

- The previous runtime-contract blocker is fixed before any child worktree or run row can be created.
- A shared `resolveDeclaredInputs` implementation keeps include and runtime sub-run contracts aligned.
- Persisting resolved child inputs makes cold resume behavior explicit and testable.
- Focused sub-run tests cover defaults, missing/undeclared values, prompt delivery, environment delivery, declared returns, and cold resume.

## Recommendation

**REQUEST CHANGES**

The main direction remains sound and the previous review was handled well. Fix the two silent `returns` contract failures, remove identifying fields from the new log, refresh onto current `dev`, and rerun validation. After that, this should be ready for a short confirmation pass.

*Report: `/Users/rasmus/.prp/archon-75601ef6/reviews/pr-2523-review.md`*

PR #2534, opened 2026-08-13
fix(workflows): enforce hermetic include references

Wirasm:
## Summary

- Problem: included workflows could bind undeclared parent state through shorthand `when:`, rejection prompts, Markdown refs, or deferred command bodies.
- Why it matters: reusable workflows need a hermetic declared interface; caller-ID collisions could consume the wrong governed data.
- What changed: validation covers every live current reference surface, and included named commands are resolved, validated, snapshotted, namespaced, and input-bound at load time.
- What did **not** change: no parent-run execution, structured I/O, runtime fan-out, provider capability, database, expression-language, or named-script templating changes.

### Before

```
include command-first block -> flatten child, defer command body
matching parent id          -> hidden child ref can bind at runtime
```

### After

```
include command-first block -> [resolve + snapshot command body]
declared caller `with:`     -> [bind `$INPUTS`]
child-local refs            -> [validate + namespace]
hidden ref/unreadable body  -> [actionable load error]
executor                    -> deterministic flat DAG
```

### Before

```
workflow-discovery -> include-expander -> flat DAG -> dag-executor
       |                   |                             |
       +- reads commands   +- best-effort scan           +- loads command again
loader -> partial live-surface validation
```

### After

```
workflow-discovery -> [~] include-expander ==> flat DAG -> dag-executor
       |                     |
       +- reads commands ----+- validate + materialize + bind + namespace
[~] loader -> validates canonical refs, when shorthand, and rejection prompts
```

**Connection inventory**:

| From | To | Status | Notes |
|------|----|--------|-------|
| workflow discovery | include expander | **modified** | Resolved include command bodies are compilation inputs |
| loader | include expander | **modified** | Shared validation proves the child namespace before flattening |
| include expander | DAG executor | **modified** | Included commands arrive as deterministic prompt nodes |
| runtime substitution | validation/rewrite walkers | **modified** | Markdown and rejection surfaces now agree |
| workflow engine | docs/defaults | **modified** | Documents compilation and non-live example placeholders |

## Label Snapshot

- Risk: `risk: medium`
- Size: `size: L`
- Scope: `workflows`, `docs`, `tests`
- Module: `workflows:composition`

## Change Metadata

- Change type: `bug`
- Primary scope: `workflows`

## Linked Issue

- Closes #2450
- Related #1764, #2453, #2512, #2475, #2477, #2488
- Depends on #2532 (merged)
- Supersedes # N/A

## Validation Evidence (required)

```bash
[...]

PR #2533, opened 2026-08-13
fix(workflows): make Codex skill use explicit (#2498)

Wirasm:
## Summary

- Problem: Codex workflow nodes inherited and advertised every ambient skill, allowing unrelated skills to steer deterministic workflow prompts; capability docs also overstated YAML skill/MCP support.
- Why it matters: command-first workflows need predictable context without removing normal project instructions or rejecting otherwise runnable Codex sessions.
- What changed: Codex workflow nodes disable automatic native skill advertisement, authors invoke installed skills explicitly with `$skill-name`, empty `skills: []` declarations are valid, unsupported YAML fields warn honestly, and bundled workflows/docs follow the explicit invocation contract.
- What did **not** change (scope boundary): direct Codex chat/title calls, Claude project/user `settingSources`, filesystem access to ambient Codex skills, and Codex's ambient/additive MCP behavior remain unchanged. This is behavioral non-advertisement, not a security boundary.

### Before

```
Workflow YAML              Archon                  Codex
─────────────              ──────                  ─────
starts AI node ──────────▶ passes prompt ────────▶ advertises every ambient skill
                                                    may select an unrelated skill
```

### After

```
Workflow command/prompt    Archon                  Codex
───────────────────────    ──────                  ─────
optionally names $skill ─▶ [disables catalog] ───▶ sees no automatic skill catalog
                                                    loads named native skill on demand
```

### Before

```
workflow schema ─▶ validator ─▶ DAG executor ─▶ provider registry ─▶ Codex SDK
bundled YAML ─────────────────────────────────────────────────────▶ prompt
provider metadata ─▶ generated docs
```

### After

```
[~] workflow schema ─▶ [~] validator ─▶ DAG executor ─▶ [~] provider registry
                                                          │
                                                          ▼
                                               [~] Codex provider ─▶ Codex SDK
                                               [~] Claude empty-list guard
[~] bundled YAML ────────────────────────────────────────▶ explicit $skill prompt
[~] provider metadata ─▶ [~] generated docs
```

**Connection inventory**

| From | To | Status | Notes |
|------|----|--------|-------|
| Workflow schema | Resource validator | **modified** | Empty skill arrays are valid; unsupported providers warn without false path validation |
| DAG executor | Provider node config | unchanged | Existing `nodeId` is the workflow-only discriminator |
| Codex provider | Codex SDK config | **modified** | Adds workflow-only `skills.include_instructions=false` and a narrow warning fallback |
[...]

Issue #2498, opened 2026-08-06
feat(workflows): isolate AI-node skills and MCPs behind explicit YAML declarations


Wirasm:
## Problem

Codex workflow nodes currently receive Codex's automatically generated filesystem-skill catalog. An ambient skill whose description matches a command or prompt can therefore be selected even though the workflow author did not ask for it. This occurred in run `42acf940a201c68b540b220eb7c6c8d2`, where an ambient `prp-issue` skill intercepted an investigation node and changed its behavior.

This is a behavioral predictability problem, not a filesystem security boundary. Workflow nodes should not advertise ambient Codex skills automatically. A model explicitly instructed to locate an installed skill may still read it.

Codex MCP behavior is separate: the SDK's `mcp_servers` config is additive over ambient configuration, and neither the locked 0.144.5 release nor stable 0.147.0 exposes a global or per-session replacement/off primitive.

## Settled Codex workflow contract

For every Codex-backed **workflow AI node**, Archon passes:

```toml
skills.include_instructions = false
```

This suppresses Codex's automatic skill catalog while preserving the target repository as `cwd` and retaining normal `AGENTS.md` project instructions.

Workflow authors invoke an installed skill explicitly in the command file or prompt using Codex's native syntax:

```yaml
- id: investigate
  command: archon-investigate-issue
```

```markdown
Use $prp-issue to investigate the issue and produce the requested artifact.
```

Codex then performs its own progressive disclosure: it selects the named installed skill, reads the original `SKILL.md`, and resolves references, scripts, and assets relative to that original directory.

The contract is:

1. Automatic skill advertisement is always off for Codex workflow AI nodes.
2. Direct Codex chat, title generation, and other non-workflow calls are unchanged.
3. Authors explicitly invoke installed Codex skills with `$skill-name`; substantial instructions should normally live in command/script files rather than inline YAML.
4. A non-empty YAML `skills:` list does **not** re-enable Codex's ambient catalog and is not an exclusive Codex allowlist. Its Codex-specific validation/meaning is deferred because Codex discovers `.agents/skills` roots but not Archon's `.claude/skills` roots, and duplicate/frontmatter behavior differs by Codex version.
5. Other providers retain their provider-specific `skills:` semantics. Providers that support exact per-node loading continue to honor the list.
6. `skills: []` becomes valid YAML. For Codex it produces the same catalog-off runtime posture; for exact-loading providers it represents an empty declared set.
[...]

Issue #1992, opened 2026-06-14
DAG executor threads session id across provider boundaries → downstream node fails to resume (claude-terminal TUI boot timeout)


pfurini:
## Summary
In a DAG workflow that mixes providers across sequential nodes, a downstream node inherits the previous node's provider session id and tries to resume it under a DIFFERENT provider. The resume target doesn't exist, so the node fails to start. With the claude-terminal provider this manifests as a TUI boot timeout stuck on the "Resume session" picker.

## Repro
1. A DAG workflow with a sequential chain where node A uses `provider: cursor` and the next AI node B uses `provider: claude-terminal` and does NOT set `context: fresh`.

Node B's first turn logs `isResume: true` with a session id created by node A, then:

```
claude-terminal: TUI did not become input-ready within boot timeout.
Last screen: Resume session ⌕ agent-<uuid>  "No sessions match agent-<uuid>"
```

Retries (2x) repeat identically; the node fails and aborts the workflow.

## Root cause
`packages/workflows/src/dag-executor.ts` (~3301-3305):

```ts
const isFreshSequential = isParallelLayer || node.context === 'fresh';
let resumeSessionId = isFreshSequential ? undefined : lastSequentialSessionId;
```

`lastSequentialSessionId` is inherited with no provider-equality check. Cross-provider resume is impossible, so the downstream provider gets a foreign session id.

## Expected
A sequential node should inherit the previous session id only when the previous node ran on the SAME resolved provider. On a provider change, start a fresh session.

## Workaround
Add `context: fresh` to every node that follows a different-provider node.

## Notes
- Homogeneous-provider workflows are unaffected (resume targets always valid).
- claude-terminal `ensureInputReady` (`packages/providers/.../claude-terminal/provider.ts` ~374-388) could additionally detect the resume picker and Esc out, but the core fix is the provider-boundary session reset.
- Environment: archon @ commit ffa21cf1, Claude Code CLI 2.1.177, macOS.

Wirasm:
## 🔍 Investigation: DAG executor threads session id across provider boundaries

**Type**: `BUG` (also covers #2050, closed as a duplicate of this issue)

### Assessment

| Metric | Value | Reasoning |
|--------|-------|-----------|
| Severity | `HIGH` | Any mixed-provider sequential DAG hard-fails on Claude/claude-terminal and silently cold-falls-back on Codex; the only workaround is annotating every provider-boundary node with `context: fresh`. |
[...]

PR #2539, opened 2026-08-14
fix(workflows): honor context: 'shared' on parallel-layer DAG nodes

### LINKED ISSUE #2538: Fail fast when scalar context: shared is used in a parallel layer
## Problem

`context: shared` is a valid workflow field for continuing the ambient session through a sequential chain. In a structurally parallel layer, however, the executor clears that ambient cursor before it evaluates sibling conditions or dispatches nodes. A scalar `shared` declaration there therefore behaves like a fresh session without an error or warning.

That silent no-op is the bug. Making scalar `shared` inherit inside a parallel layer is no longer the desired fix: merged #2643 added the deterministic, provider-safe contract for that use case:

```yaml
context:
  resume: upstream-node-id
```

Named resume identifies the exact upstream session, verifies ancestry and provider fork support, and remains unambiguous when several siblings can run.

## Accepted contract

- Scalar `context: shared` remains supported for sequential ambient-session threading.
- A command or prompt node using scalar `context: shared` in any structurally parallel layer fails workflow loading.
- The error tells the author to either:
  - use `context: { resume: <upstream-node-id> }` to fork a specific upstream session; or
  - serialize the nodes with dependencies when ambient sequential threading is intended.
- This is based on structural concurrency, not which branches happen to execute. Mutually exclusive sibling `when:` conditions still form a parallel layer, and must use named resume or explicit serialization.
- Existing named-resume behavior from #2643 and sequential scalar-shared behavior remain unchanged.

## Why load time owns this

The DAG loader already validates graph structure and named resume ancestry. It can identify every layer before provider execution starts, so an invalid scalar `shared` declaration should be rejected before a run is created instead of degrading to a fresh session at runtime.

## Acceptance

- A workflow with scalar `context: shared` on a node in a multi-node layer fails loading with actionable guidance.
- The same failure occurs when sibling `when:` conditions appear mutually exclusive.
- A sequential `context: shared` node continues to load.
- `context: { resume: source }` continues to load when its existing ancestry and provider requirements are satisfied.
- The authoring guide and quick reference state that scalar `shared` is sequential-only and named resume is the parallel path.

## Scope

This issue does not add provider capability or runtime session-fork machinery. #2643 already owns named session ancestry, immutable forks, and their provider gates.

Closes the remaining silent-no-op gap originally reported here. PR #2539 was closed as superseded by #2643 plus this narrower fail-fast contract.

hidegh:
## Summary

- Problem: `context: 'shared'` is a documented DAG-node field meant to override a parallel
  layer's default of a fresh AI session per sibling, forcing one sibling to resume the
  session left by the node that fed the layer. In `runLayers()`
  (`packages/workflows/src/dag-executor.ts`), the sequential-session cursor
  (`ctx.lastSequentialSession`) was wiped at the _start_ of every parallel layer â€” before any
  node could read it â€” and the freshness predicate never checked for `'shared'` at all. The
  setting parsed and validated but had zero runtime effect: no warning, no error, just a
  fresh session every time.
- Why it matters: any workflow author using `context: 'shared'` inside a parallel layer got
  silently wrong behavior with no signal anything was off.
- What changed: in `runLayers()`, moved the parallel-layer cursor reset from before to after
  the layer's node dispatch, and added a `node.context !== 'shared'` carve-out to the
  freshness predicate. Two-line logic change plus one new regression test.
- What did **not** change (scope boundary): sequential-layer session threading (already
  correct), the cross-provider guard (#1992), loop/loop_group session semantics
  (`fresh_context` â€” a separate, intentionally different and already-warned mechanism), and
  no schema/YAML surface change.

## UX Journey

### Before

```
Workflow author                 Engine (runLayers)
â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€                 â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
writes:
  - id: a
    prompt: ...
  - id: b
    depends_on: [a]
    context: shared   â”€â”€â–¶   layer [b, c] is parallel (len > 1)
  - id: c                    ctx.lastSequentialSession reset to undefined
    depends_on: [a]           BEFORE b/c dispatch                    [!]
                              isFreshSequential = isParallelLayer
                                (true regardless of node.context)    [!]
                              b runs with a FRESH session â€” sess-a lost
                              (identical to c, despite `context: shared`)
```

### After

```
Workflow author                 Engine (runLayers)
â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€                 â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
writes:
  - id: a
    prompt: ...
  - id: b
    depends_on: [a]
    context: shared   â”€â”€â–¶   layer [b, c] is parallel
  - id: c                    cursor still holds sess-a when b/c dispatch [+]
    depends_on: [a]          b: isFreshSequential = false
                                (context: 'shared' overrides parallel default) [+]
                                â†’ resumes sess-a (same-provider guard still applies)
                              c: isFreshSequential = true (unchanged) â†’ fresh
                              AFTER dispatch: cursor reset to undefined
                                (still can't hand ONE cursor to the next layer)
```

## Architecture Diagram

### Before

```
runLayers()
  for each layer:
    isParallelLayer = layer.length > 1
    if (isParallelLayer): ctx.lastSequentialSession = undefined   [reset BEFORE dispatch]
        â”‚
        â–¼
    Promise.allSettled(layer.map(node => {
      isFreshSequential = isParallelLayer || node.context === 'fresh'  [no 'shared' check]
      cursor = ctx.lastSequentialSession    // already undefined in a parallel layer
      resumeSessionId = isFreshSequential ? undefined : cursor?.sessionId
    }))
```

### After

```
runLayers()
  for each layer:
    isParallelLayer = layer.length > 1
        â”‚
        â–¼
    Promise.allSettled(layer.map(node => {
      isFreshSequential =
        node.context === 'fresh' || (isParallelLayer && node.context !== 'shared')  [~]
      cursor = ctx.lastSequentialSession    // still holds the pre-layer value
      resumeSessionId = isFreshSequential ? undefined : cursor?.sessionId
    }))
        â”‚
        â–¼
    if (isParallelLayer): ctx.lastSequentialSession = undefined   [~ reset moved AFTER dispatch]
```

**Connection inventory:**

| From                                                             | To                          | Status         | Notes                                                                                                           |
| ------------------------------------------------------------------ | ---------------------------- | -------------- | ------------------------------------------------------------------------------------------------------------- |
| `runLayers()` parallel-layer cursor reset                        | `ctx.lastSequentialSession` | **modified**   | Moved from before to after `Promise.allSettled` dispatch                                                        |
| `runLayers()` per-node `isFreshSequential`                       | `node.context`              | **modified**   | Added `node.context !== 'shared'` carve-out inside the `isParallelLayer` branch                                 |
| `executeLoopGroupNode()` per-iteration body (`runLayers` caller) | `runLayers()`               | **unaffected** | Same shared function â€” benefits automatically, no separate change needed                                        |
| `isPersistableNode()` / `loader.ts` persist_session validation   | `node.context === 'fresh'`  | **unaffected** | Different concern (cross-run persistence); confirmed correct, not touched                                       |
| `LOOP_NODE_AI_FIELDS` / `LOOP_GROUP_NODE_AI_FIELDS`              | `context` field             | **unaffected** | Confirmed `context` is intentionally ignored (with a load-time warning) on loop/loop_group nodes â€” out of scope |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: XS`
- Scope: `workflows`
- Module: `workflows:dag-executor`

## Change Metadata

- Change type: `fix`
- Primary scope: `workflows`

## Linked Issue

- Closes #2538
- Related: #1992 (cross-provider session-boundary guard â€” read, not modified, by this fix)
- Depends on: none
- Supersedes: none

## Rebase (2026-08-14)

Rebased onto `dev` at `e88deead` (2 commits ahead of this branch's previous base, both
upstream `fix(workflows):` commits unrelated to `dag-executor.ts`/session threading:
`fix(workflows): enforce hermetic include references (#2534)` and
`fix(workflows): make Codex skill use explicit (#2498) (#2533)`). Rebase was a clean
single-commit replay with **zero conflicts** (`git rebase dev` â€” "Successfully rebased and
updated refs/heads/fix/parallel-shared-context-session"); commit hash changed from `db6274c6`
to `142789b4`, content unchanged. Force-pushed to `origin` with `--force-with-lease`.

Post-rebase full validation run against the CONTRIBUTING.md checklist (`check:bundled`,
`check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`,
`check:capability-matrix`, `type-check`, `lint`, `format:check`, `test` per package) â€” all
pass. Full breakdown below.

```bash
bun run check:bundled              # PASS â€” bundled-defaults.generated.ts up to date
bun run check:bundled-skill        # PASS â€” bundled-skill.ts up to date
bun run check:bundled-schema       # PASS â€” bundled-schema.generated.ts up to date
bun run check:pi-vendor-map        # PASS
bun run check:capability-matrix    # PASS
bun run type-check                 # PASS â€” all 10 packages clean
bun run lint                       # PASS â€” zero warnings
bun run format:check               # PASS â€” no files need reformatting
```

`test:install` intentionally skipped â€” it requires WSL2 and explicitly refuses to run on
native Windows (`[ERROR] Windows is not supported`), a documented limitation of this dev
host, not a code issue.

Per-package `bun --filter <pkg> test` results (run standalone per package, per
`AGENTS.md`'s "do NOT run `bun test` from the repo root" / avoid the `--parallel` SIGINT
cascade guidance â€” one package's failure otherwise aborts its siblings mid-run):

| Package             | Result                                                                                            | Notes                                                                                                                                                                                                                                                         |
| ------------------- | --------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `@archon/paths`     | 266 pass, 0 fail                                                                                  | untouched                                                                                                                                                                                                                                                     |
| `@archon/git`       | 196 pass, 0 fail                                                                                  | untouched                                                                                                                                                                                                                                                     |
| `@archon/isolation` | 56 pass, 0 fail                                                                                   | untouched                                                                                                                                                                                                                                                     |
| `@archon/workflows` | 451 pass on `dag-executor.test.ts` (this branch's actual diff), **1 pre-existing fail** elsewhere | `load-command-prompt.test.ts`'s symlink test fails with `EPERM: operation not permitted, symlink` â€” creating a symlink on Windows needs admin/dev mode; unrelated to this branch (only `dag-executor.ts`/`dag-executor.test.ts` are touched, both 100% green) |
| `@archon/core`      | 513 pass, **1 pre-existing fail**                                                                 | `ConversationLockManager > queued messages process in order after completion` â€” a `setTimeout`-based timing assertion (`executionOrder` off by one), flaky on this host, untouched by this branch's diff                                                      |
| `@archon/adapters`  | all pass, 0 fail                                                                                  | untouched                                                                                                                                                                                                                                                     |
| `@archon/server`    | 23 pass, 0 fail                                                                                   | untouched                                                                                                                                                                                                                                                     |
| `@archon/providers` | 15 pass, 0 fail                                                                                   | untouched                                                                                                                                                                                                                                                     |
| `@archon/web`       | 395 pass, 0 fail                                                                                  | untouched                                                                                                                                                                                                                                                     |
| `@archon/cli`       | 18 pass, **2 pre-existing fail**                                                                  | `downloadWebDist` tar-extraction tests â€” `tar: ...: Cannot open: No such file or directory`, a known Windows `tar` path-quoting issue, unrelated to this branch                                                                                               |

All three failures are pre-existing, Windows-host-specific, environment limitations â€”
confirmed by scope (none are in `dag-executor.ts`/`dag-executor.test.ts`, the only files this
branch's diff touches) and by pattern-matching the exact same failure signatures documented
in this repo's own prior PR validation notes (`.docs/unified-channel-ref.pr.md`).

Additionally verified the new test is a genuine regression check, not a tautology: with the
`dag-executor.ts` fix reverted via `git stash` (test file kept), the new test fails with
`Expected: "sess-a", Received: undefined` â€” confirming it fails against the pre-fix code and
passes only with the fix applied. Full source diff restored via `git stash pop` afterward;
confirmed both files show as modified again.

- Evidence provided: command output captured during this session (type-check, targeted test
  run, the stash/pop before/after comparison, and the full post-rebase validation sweep).
- Intentionally skipped: full `bun run validate` / full-repo `bun run test` â€” the change is
  isolated to one function in `@archon/workflows`, and the package-level suite plus
  type-check give complete coverage of the affected code path; no other package imports the
  modified (module-private) `runLayers` function directly.

## Security Impact (required)

- If any `Yes`: n/a. This only changes which prior AI session id (if any) a node resumes;
  the existing cross-provider guard (#1992) is untouched and still prevents a session id
  from threading into a node that resolves to a different provider than the one that
  created it.

## Compatibility / Migration

- Backward compatible? `Yes` â€” `context: 'shared'` was previously inert inside a parallel
  layer (indistinguishable from unset), so no workflow could have depended on the old
  (broken) behavior. Sequential-layer behavior is untouched.
- Config/env changes? `No`
- If yes: n/a.

- Verified scenarios: read every call site of `node.context` in `packages/workflows/src`
  (not just the one being fixed) to confirm this is the only silent, unwarned bypass â€” see
  the Investigation section of the PRD/issue docs for the full table. Confirmed the fix
  doesn't reintroduce the race the original reset guarded against (parallel siblings never
  write back to `ctx.lastSequentialSession` mid-layer â€” verified the `!isParallelLayer`
  guard on the completion path is unchanged).
- Edge cases checked: a parallel-layer sibling _without_ the override (must stay fresh â€”
  covered by the new test); the interactive-loop-group parallel-tail-gate test that
  specifically depends on the reset still happening after a parallel layer (existing test,
  still passes, confirming the reset's _timing_ change didn't remove its _effect_).
- What was **not** verified: no manual end-to-end workflow run through the CLI/web UI with a
  real provider â€” this is a pure in-process control-flow fix fully exercised by the mocked
  unit test; a live run would exercise the same code path with no additional risk surface.

- Affected subsystems/workflows: only workflow runs (and `loop_group` body iterations, which
  share the same `runLayers()` function) that contain a parallel layer (2+ nodes with
  identical `depends_on`) where at least one node sets `context: 'shared'`. No effect on any
  other workflow shape.
- Potential unintended effects: none identified â€” the reset still happens, just after
  dispatch instead of before, and no code reads `ctx.lastSequentialSession` between "layer
  dispatch completes" and "the reset runs" other than the already-guarded write-back.
- Guardrails/monitoring for early detection: existing `dag.session_resume_failed` /
  `dag.session_provider_boundary_fresh` structured log events already fire if a resumed
  session turns out to be cold or crosses a provider boundary â€” unchanged, and now also
  cover the newly-functional `'shared'` parallel path.

## Rollback Plan (required)

- Fast rollback command/path: revert this PR's commit(s) on `dev` â€” the change is two small,
  self-contained edits in one function; no data migration to reverse, no config to unset.
- Feature flags or config toggles (if any): none â€” `context: 'shared'` is not gated behind a
  flag; it's an existing documented field that now behaves as documented.
- Observable failure symptoms: a parallel-layer node with `context: 'shared'` resuming a
  session unexpectedly (or a sibling without the override picking up the wrong session,
  which would indicate the reset-timing change regressed).

## Risks and Mitigations

- Risk: delaying the cursor reset to after dispatch could, in theory, let a parallel node's
  completion race the reset and leak a stale cursor into the next layer.
  - Mitigation: the existing `!isParallelLayer` guard on the per-node completion path
    already prevents any parallel-layer node from writing to `ctx.lastSequentialSession` â€”
    confirmed by reading that guard, not just assuming it. Every node in the layer reads the
    identical, unmodified pre-layer cursor value; the reset after `Promise.allSettled`
    still runs before the next layer starts, closing the window in exactly the same place
    control flow already serializes on.
- Risk: the fix narrowly targets `runLayers()`; a hypothetical duplicate implementation
  elsewhere (e.g. `dry-run.ts`) could still have the bug.
  - Mitigation: checked â€” `dry-run.ts` does not read or duplicate any session-cursor logic,
    so there is no second copy of this bug to fix.


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Shared context now works across sequential and parallel workflow stages.
  * Providers report whether they support safely forking sessions for concurrent shared-context execution.
  * Claude supports session forking; other listed providers are identified as unsupported.
  * Workflow usage totals now include tokens and costs from completed, failed, cancelled, paused, and nested runs.

* **Bug Fixes**
  * Prevented unsafe concurrent shared-context execution with providers that cannot fork sessions.
  * Single shared-context nodes continue to run safely without session-fork support.
  * Improved loop completion reporting and handling of large command outputs.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


Wirasm:
Thank you for this — and I want to be straight that the diagnosis is correct and the diff is good work. `context: 'shared'` inside a parallel layer parses, validates, and then silently does nothing. That is a real bug and exactly the kind of silent no-op we say we don't ship. The implementation is also right: moving the cursor reset after dispatch is behavior-preserving, the carve-out in the freshness predicate is minimal and in the correct place, the write-back guard is correctly left alone so a parallel node never becomes the cursor, and the test asserts against the real `sendQuery` signature.

I'm still going to ask for a different shape, and the reason isn't a direction-document clause — I checked, and none applies. It's a risk call about what the fix makes reachable.

**The problem is what happens with two or more `shared` siblings in one layer.** The change doesn't make concurrent nodes share a *live* session — each `shared` node gets a read of the same completed upstream session, and the producer has already finished. That part is coherent. But it does create the first path in the DAG where two concurrently-executing nodes are handed the same resume id, and that is only safe on one provider:

- **Claude** — safe. `dag-executor.ts:1524-1530` sets `forkSession: true` on every resume, so each sibling forks its own branch and the source is untouched.
- **Codex** — races. `codex.resumeThread(resumeSessionId, …)` (`codex/provider.ts:932`), no fork concept; two siblings drive one thread.
- **Pi** — races. `SessionManager.open(match.path)` (`pi/session-resolver.ts:50`); two siblings append to the same `.jsonl`.
- **Copilot** — silently starts fresh (pre-existing).

Before this PR that path was unreachable, because `persist_session` keys are per-node-id so distinct nodes never collided. Nothing in the diff prevents or warns about it, and the failure mode on Codex/Pi is interleaved session state with no diagnostic — low likelihood, high confusion cost, and very hard to attribute after the fact.

**There's also a naming problem the fix inherits.** An author who writes `context: 'shared'` on both `b` and `c` almost certainly means *"b and c share one session with each other."* That is impossible, and it isn't what they'd get — they'd get two independent inheritances of `a`. Making the field work in parallel layers makes that misreading easier to reach, not harder.

**What I'd like instead: a load-time warning that `context: 'shared'` has no effect inside a parallel layer.**

That closes the actual complaint in #2538 — the silence — without opening the multi-sibling path at all. It matches `direction.md §schema-width`'s "capability mismatches warn loudly, never silently no-op". It's a smaller, safer diff, and it's honest with the author at the moment they write the YAML rather than at the moment their sessions interleave. If you'd like to convert this PR into that, I'd take it gladly — I'm leaving it open rather than closing it so the work and the discussion stay in one place, and it's entirely your call whether to repurpose it or close it and start fresh.

To be clear about what this isn't: I'm not saying `shared` in a parallel layer should never work. It's a reasonable thing to want. But making it work correctly needs the sibling-collision question answered first (a guard, or per-provider fork semantics, or a rename that stops inviting the wrong reading), and that's a design call I'd rather make deliberately than as a side effect of a bug fix. If you want to take that on as a follow-up I'm happy to work through the shape with you.

Also worth flagging for whichever direction this goes: the documented contract for `shared` is incomplete in `authoring-workflows.md:219` and `book/quick-reference.md:137`. Neither says it inherits from the immediately preceding *sequential* node only, doesn't propagate past a parallel layer, doesn't survive a resume, and requires the same provider. A docs fix is valuable on its own regardless of which code path we take.

hidegh:
#2099 solves the problem, have to check if documentation reg. shared is updated and whether shared is dropped.

Wirasm:
Thank you for the contribution and for isolating the silent `context: shared` behavior.

We are closing this PR because #2643 has since shipped the supported shape for deliberate session reuse in a parallel layer: `context: { resume: <upstream-node-id> }`. That explicit source is deterministic, validates ancestry and provider capabilities, and safely forks the named session. The runtime and capability work in this branch is now duplicated by that implementation and conflicts with current `dev`.

This does not mean the original silent-no-op report is fully resolved. We are keeping #2538 as a narrower follow-up: scalar `context: shared` remains sequential-only, and workflow loading should fail with an actionable error when it is used in a structurally parallel layer, directing authors to named resume or explicit serialization. We will also correct the related docs.

Thanks again—the report and implementation helped expose the ambiguity, even though the final contract moved to the explicit named-resume shape.

PR #2551, opened 2026-08-17
fix(core): refuse a chat turn whose worktree is gone

JeanFraga:
## Problem and outcome

A conversation whose isolated worktree has been removed keeps its `cwd` pointing at the deleted path, and every subsequent chat message fails with an error that blames the wrong thing entirely: the Claude Agent SDK reports `Claude Code native binary at … exists but failed to launch. This usually means the binary does not match this system's libc — e.g. spawning a musl-linked binary on a glibc Linux host`. The binary is fine. It is a deleted directory, on macOS, with no libc involved.

- **Outcome:** the chat turn is refused up front, naming the missing directory and the two commands that recover from it, instead of reaching the provider and surfacing a libc red herring.
- **Invariant:** a conversation bound to an isolated worktree never silently executes in the live checkout instead.
- **Scope boundary:** the several other call sites that mark an isolation env destroyed (`packages/cli/src/commands/isolation.ts`, `packages/server/src/routes/api.ts`, `packages/isolation/src/resolver.ts`, the periodic reaper) still leave `cwd` set. Fixing the read site covers all of them, including causes no write site can see — a user's own `rm -rf`.
- **Root cause:** `posix_spawn` reports a missing *working directory* as `ENOENT` against the **executable's** path. The SDK's handler checks only whether the executable exists on disk; it does, so it concludes a libc/architecture mismatch. Reproduced directly:

  ```js
  spawn('/…/claude-agent-sdk-darwin-arm64/claude', ['--version'], { cwd: '<deleted worktree>' })
  // → SPAWN ERROR: ENOENT posix_spawn '/…/claude'   ← names the binary, not the cwd
  ```

  The same binary runs fine from a shell (`2.1.209 (Claude Code)`), which is what makes the SDK's message so misleading.

## Review guidance

- **Feedback requested:** correctness of the refuse-vs-fallback decision in the orchestrator, and whether `SPAWN_FAILURE_PATTERNS` is scoped tightly enough.
- **Start here:** `packages/core/src/orchestrator/orchestrator-agent.ts` — the guard placement is the load-bearing part: after the deterministic-command early-returns so `/worktree remove` and `/setproject` still work in this state, before the user-message persist so a refused turn leaves no `user` row without its `assistant` pair.
- **Review order:** orchestrator-agent (the guard) → claude/provider (the error message) → cleanup-service (the write site) → the three test files.
- **Lower-attention areas:** none; every hunk is hand-written and small.
- **Known risk or uncertainty:** the provider check matches on error text (`failed to launch`, `enoent`) before consulting the filesystem. Text matching is inherently brittle across SDK versions, but a miss degrades to today's behavior — the original error — rather than to a wrong answer. Container runs pass `undefined` so the check cannot misfire on a path that legitimately exists only inside the container.

## Solution

Three changes on one story, each at a different layer:

**Read site (the durable fix).** `handleMessage` checks `conversation.cwd` before the AI call. It deliberately does **not** fall back to `codebase.default_cwd`: the conversation asked to work in an isolated worktree, and quietly relocating the agent into the live checkout would widen its write scope without consent. Only the workflow path re-resolves isolation through `validateAndResolveIsolation`; a chat turn reads `cwd` verbatim, which is why this gap existed.

**Error message.** `classifyAndEnrichError` gains the spawn cwd and, on a spawn-failure error only, asks the one question the SDK never asks — does the working directory exist? The check is on the error path, so the happy path pays no filesystem cost. A genuine launch failure with a valid cwd still surfaces as itself.

**Write site.** `onConversationClosed` cleared `isolation_env_id` but left `cwd` naming the directory it was about to delete. It now clears both, and only when `cwd` actually names that environment — an unrelated `cwd` is untouched.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Chat message to a conversation whose worktree was removed returns a libc/architecture error naming the Claude binary | Chat turn is refused with the missing path and the recovery commands (`/worktree remove`, `/setproject`) |
| **Failure behavior** | Provider spawns into a nonexistent directory; error class `unknown`; the operator investigates musl-vs-glibc | Fails before the provider; logged as `orchestrator.conversation_cwd_missing` with `conversationId`, `cwd`, `isolationEnvId` |

## Validation

- `bun run validate` — exits 0 — proves generated-file checks, type-check, lint at `--max-warnings 0`, format, and the full per-package suite all pass.
- `bun test src/claude/provider.test.ts` — 136 pass / 0 fail — covers both the reclassified spawn failure and the case where the cwd exists and the original launch error must survive unchanged.
- `bun test src/orchestrator/orchestrator-agent.test.ts` — 218 pass / 0 fail — includes a test asserting the provider is never called when `cwd` is missing, and one asserting the unscoped-chat path is unaffected (with no codebase scoped, `cwd` is never consulted).
- `bun test src/services/cleanup-service.test.ts` — 55 pass / 0 fail — covers both clearing a matching `cwd` and leaving an unrelated one alone.
- Root cause reproduced directly with `child_process.spawn` against a deleted worktree path (transcript in the Problem section).
- **Not verified:** the end-to-end Telegram path against a live bot after the fix. The refusal branch is covered by unit tests at the orchestrator seam, which is where the decision is made; the adapter below it is a pass-through for `sendMessage`.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. No schema change; `cwd: null` is an existing valid state that `/setproject` already produces. | `handleSetProject` writes the same `cwd: null` |
| Rollout / rollback | Fully reversible — three isolated hunks, no persisted state introduced. | Single commit, 6 files |
| Observability | A stale conversation now emits `orchestrator.conversation_cwd_missing` (warn) with the conversation id, path, and isolation env id, instead of being indistinguishable from a provider crash. | `orchestrator-agent.ts` guard |

## Summary by CodeRabbit

* **Bug Fixes**
  * Project-scoped conversations now stop gracefully when their working directory is missing, with appropriate recovery guidance.
  * Conversations no longer fall back to an incorrect project directory or retain stale working-directory settings after cleanup.
  * Closing a conversation correctly clears its associated working-directory information.
  * Claude launch errors now identify missing or invalid working directories instead of showing misleading system errors.
  * Genuine Claude runtime errors continue to display their original diagnostics.

* **Documentation**
  * Added troubleshooting guidance for missing conversation working directories and recovery options.

### DIFF
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -277,8 +277,12 @@ mock.module('@archon/git', () => ({
   hasUncommittedChanges: mock(() => Promise.resolve(false)),
 }));
 
+// Hoisted so individual tests can make a specific path report as missing (the
+// conversation-cwd guard in handleMessage). Default: everything exists.
+const mockExistsSync = mock((_path: string) => true);
+
 mock.module('fs', () => ({
-  existsSync: mock(() => true),
+  existsSync: mockExistsSync,
   // token-crypto.ts imports these from node:fs for the auto-provisioned credential
   // key. readFileSync returns a valid 64-hex key so getEncryptionKey() resolves
   // without any real disk write when the per-user credential path is exercised.
@@ -1403,6 +1407,7 @@ describe('provider cwd resolution', () => {
     mockSendQuery.mockClear();
     mockEnsureArchonWorkspacesPath.mockClear();
     mockLogger.warn.mockClear();
+    mockExistsSync.mockImplementation(() => true);
     mockGetOrCreateConversation.mockImplementation(() => Promise.resolve(null));
     mockGetCodebase.mockImplementation(() => Promise.resolve(null));
     mockListCodebases.mockImplementation(() => Promise.resolve([]));
@@ -1441,6 +1446,63 @@ describe('provider cwd resolution', () => {
 
     expect(getSendQueryCwd()).toBe('/worktrees/feature-branch');
     expect(mockEnsureArchonWorkspacesPath).not.toHaveBeenCalled();
+  });
+
+  test('scoped chat refuses the turn when conversation.cwd no longer exists', async () => {
+    const codebase = makeCodebaseForSync();
+    const conversation = makeConversation({
+      codebase_id: 'codebase-1',
+      cwd: '/worktrees/deleted-branch',
+      isolation_env_id: 'env-gone',
+    });
+    mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+    mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+    mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+    mockExistsSync.mockImplementation((p: string) => p !== '/worktrees/deleted-branch');
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', 'hello');
+
+    // Never reaches the provider: spawning there fails ENOENT and the Claude
+    // SDK misreports it as a binary/libc mismatch.
+    expect(mockSendQuery).not.toHaveBeenCalled();
+    const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
+    expect(sent).toContain('/worktrees/deleted-branch');
+    expect(sent).toContain('/setproject');
+  });
+
+  test('missing conversation.cwd does not silently fall back to default_cwd', async () => {
+    const codebase = makeCodebaseForSync();
+    const conversation = makeConversation({
+      codebase_id: 'codebase-1',
+      cwd: '/worktrees/deleted-branch',
+    });
+    mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+    mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+    mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+    mockExistsSync.mockImplementation((p: string) => p !== '/worktrees/deleted-branch');
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', 'hello');
+
+    // Relocating the agent into the live checkout would widen its write scope
+    // without the user asking for it.
+    expect(mockSendQuery).not.toHaveBeenCalled();
+  });
+
+  test('unscoped chat ignores a missing conversation.cwd', async () => {
+    const conversation = makeConversation({ codebase_id: null, cwd: '/worktrees/deleted-branch' });
+    mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+    mockListCodebases.mockReturnValueOnce(Promise.resolve([]));
+    mockExistsSync.mockImplementation((p: string) => p !== '/worktrees/deleted-branch');
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', 'hello');
+
+    // With no codebase scoped, cwd is never consulted — the workspaces path wins,
+    // so a stale override must not block the turn.
+    expect(mockEnsureArchonWorkspacesPath).toHaveBeenCalled();
+    expect(mockSendQuery).toHaveBeenCalled();
   });
 
   test('unscoped chat uses ensureArchonWorkspacesPath result', async () => {
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -346,6 +346,19 @@ function makeCodebase(name: string, id = `id-${name}`): Codebase {
   };
 }
 
+// `existsSync` is one shared mock for the whole file, so whatever predicate the
+// last test left behind is what the next one inherits. Resetting it inside a
+// single describe is not enough: everything below that block would run against a
+// leaked predicate, including the `/setproject` and `/update-project` suites,
+// which drive the existsSync call sites in handleSetProject and
+// handleUpdateProject. Today a leak is survivable only because the predicates
+// here reject one literal path — that is luck, not a guarantee, and it stops
+// being true the moment a test rejects something broader. Reset before every
+// test so no describe can poison another.
+beforeEach(() => {
+  mockExistsSync.mockImplementation(() => true);
+});
+
 // ─── parseOrchestratorCommands ────────────────────────────────────────────────
 
 describe('parseOrchestratorCommands', () => {
@@ -1407,7 +1420,7 @@ describe('provider cwd resolution', () => {
     mockSendQuery.mockClear();
     mockEnsureArchonWorkspacesPath.mockClear();
     mockLogger.warn.mockClear();
-    mockExistsSync.mockImplementation(() => true);
+    // existsSync is reset by the top-level beforeEach, which covers every describe.
     mockGetOrCreateConversation.mockImplementation(() => Promise.resolve(null));
     mockGetCodebase.mockImplementation(() => Promise.resolve(null));
     mockListCodebases.mockImplementation(() => Promise.resolve([]));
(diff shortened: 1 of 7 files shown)


Wirasm:
This is on the right side of a rule I care about, and I want to say why explicitly since it's the thing I checked hardest.

`AGENTS.md` → *No Autonomous Lifecycle Mutation Across Process Boundaries* forbids a process marking work failed/cancelled/abandoned because it *guessed* the owner is gone. This guard writes nothing — no status change, no env teardown, no session deactivation. It sends a message and returns, and the next turn re-evaluates from scratch. That's the shape the rule asks for: surface the ambiguous state, hand the user an action. A transient condition (a mount blip, a worktree mid-move) costs exactly one refused message with zero persisted consequence. The one mutation in the diff (`cleanup-service.ts:275`, `cwd: null`) isn't a staleness guess either — it runs inside `onConversationClosed`, which is about to delete that directory anyway, and it's conditioned on `conversation.cwd === env.working_path`. That's closing a half-clear, not inventing a new one.

The detection is also structurally sound: `conversation.cwd` is only ever written with host paths (`command-handler.ts:356`, `:451`, `handleSetProject`, and now cleanup), so a container path can't land there and produce a false positive — independent of the `isContainerRun ? undefined : cwd` guard at the provider layer. And on your own review question — yes, the string prefilter is scoped tightly enough, because the actual gate is `pathKind(hostCwd) !== 'directory'` (`provider.ts:1288`). A stray "ENOENT" from a tool call in a healthy session can't misfire, since the cwd exists.

Two things before merge:

**1. The refusal recommends a command that no-ops in a reachable state.** The message suggests `/worktree remove`, but when `isolation_env_id` is null that hard-returns *"This conversation is not using a worktree."* (`command-handler.ts:428-430`), and the message's claim that "its isolated worktree was removed" isn't supported. This was flagged as hypothetical in review — it isn't. `orchestrator.ts:204` (`validateAndResolveIsolation`, the `stale_cleaned` branch) clears `isolation_env_id` and leaves `cwd` set. So: a workflow run hits stale-cleaned, then a chat message lands in exactly the state where the first suggestion dead-ends. This PR fixes the *other* producer of that state (`cleanup-service.ts:269`); `orchestrator.ts:204` remains. Could the message branch on `isolation_env_id` — drop the worktree claim and the `/worktree remove` suggestion when it's null, leaving `/setproject`, which always works?

**2. Please hoist the `mockExistsSync` reset to a top-level `beforeEach`.** The new shared mutable fs fixture (`orchestrator-agent.test.ts:280-282`) is reset in one `beforeEach` only, inside `describe('provider cwd resolution')` at `:1399`, and the file has no top-level `beforeEach`. Whatever predicate the last test in that block leaves behind leaks into the 13 describes below it — including `/setproject` (`:3240`) and `/update-project` (`:3567`), which drive the other `existsSync` call sites at `orchestrator-agent.ts:2506` and `:2603`. It's harmless today only because the leaked predicate rejects one literal string. It won't stay harmless: #2565 rewrites `describe('natural-language approval routing')` at `:2014` in this same file, and a test there that gives a scoped conversation a non-null `cwd` would silently take your guard's early return and assert against the wrong thing. One line now saves a genuinely confusing debugging session later.

Optional: a doc line somewhere for the new refusal would be good — it's a new user-visible behavior and the PR documents it nowhere.

One framing note, not a change request. The body's stated invariant — *"a conversation bound to an isolated worktree never silently executes in the live checkout"* — describes a risk that doesn't exist today: `orchestrator-agent.ts:1534` is `conversation.cwd ?? scopedCodebase.default_cwd`, and a *missing* cwd is still non-null, so nothing ever fell back. I'd rather the PR be accepted on what it actually delivers, which is plenty: an accurate error instead of a musl-vs-glibc red herring, coverage at the provider layer so workflow nodes benefit too and not just chat, and no wasted provider spawn.

Wirasm:
Maintainer note, and to be clear up front: **this is not another ask on this PR.** I am recording a finding so it does not get lost, and flagging where credit belongs.

While triaging #2114 (waaaw's earlier draft on the same defect, which the operator has decided to fold into this PR) I confirmed one slice this PR does not cover. Your guard is:

```ts
if (conversation.codebase_id !== null && conversation.cwd !== null) {
```

So a scoped chat with `cwd === null`, against a project whose `codebase.default_cwd` has since been deleted, is not refused at the orchestrator. The Claude provider change catches it a layer down, but that fix lives in `packages/providers/src/claude/provider.ts`, so Codex, Pi, OpenCode and Copilot still surface the opaque error. That is the Codex-in-Docker shape from the original #1170 report.

**That gap is real, but it is not yours to close here.** Your guard sits before the user-message persist, which is the right place and the reason it works. But `codebaseDb.listCodebases()` is at `orchestrator-agent.ts:1587`, after the persist, and `scopedCodebase` is not derived until `:1706` — so `default_cwd` genuinely is not in scope where your guard runs. Covering it needs the codebase load moved earlier, or an extra targeted fetch on every scoped turn, or a second guard at the resolution site. That third option reintroduces an orphaned `user` row, which I measured directly on #2114's patch. Any of the three is a design decision deserving its own review rather than a rider on a PR that is already correct for what it claims.

So my recommendation to the maintainers is that this ships as-is, and the `default_cwd` slice lands as a follow-up credited to @waaaw, who found it. Nothing on your plate changes: the two review asks already on this PR (branch the refusal message on `isolation_env_id`; hoist the `mockExistsSync` reset to a top-level `beforeEach`) remain the only things between this and merge.

One small thing while I was in here — I reproduced your root cause independently:

```
```

`posix_spawn` really does name the executable when the *working directory* is what is missing. The diagnosis holds up exactly as you wrote it.

Wirasm:
Hi @JeanFraga — a maintainer has pushed the two review changes to your branch rather than waiting for you. I want to be plain about what that is and what it isn't.

**It's a scheduling call, not a criticism.** Your diagnosis was right, the approach was accepted on review, and the PR was two small changes away from mergeable. A live defect is waiting on it, and so is a follow-up PR, so we took the last mile rather than leave it parked. Your commit `d5c926de` is untouched and still yours — everything below is a separate commit on top, and `git fetch && git reset --hard d5c926de` gets you back to exactly what you wrote if you want to take it from there.

## What changed

**1. The refusal message now branches on `isolation_env_id`.**

You were asked about this because the message suggested `/worktree remove` unconditionally. I confirmed both halves of that concern in the code before changing anything:

- `command-handler.ts:428-429` — `/worktree remove` hard-returns `'This conversation is not using a worktree.'` when `isolation_env_id` is falsy.
- `orchestrator.ts:204` — the `stale_cleaned` branch in `validateAndResolveIsolation` calls `updateConversation(id, { isolation_env_id: null })` and never touches `cwd`. So a workflow run can strand a conversation in exactly that state, and the next chat turn would be told to run a command that dead-ends. Your PR fixes the other producer of that state in `cleanup-service.ts`; this one is still live.

When an environment is attached the message is unchanged. When it has already been cleared, the worktree claim and the `/worktree remove` suggestion are dropped, leaving `/setproject`, which clears the cwd override and works in both states.

**2. The `mockExistsSync` reset is hoisted to a top-level `beforeEach`**, and the now-redundant per-describe reset is removed.

**3. A troubleshooting doc section** (the "optional" item from the review). `reference/troubleshooting.md` already had a matching Symptom/Cause/Fix shape, so it wasn't a shoehorn. It documents the refusal, why Archon refuses instead of falling back to the project root, which recovery command applies in which state, and the `orchestrator.conversation_cwd_missing` log event.

## Evidence

Both changes were verified by breaking them and watching the right thing fail, rather than only watching them pass.

Reverting the message branch to the unconditional version turned exactly one test red — the new one — and left your original test asserting `/worktree remove` **is** present when an environment is attached green, so the two arms are genuinely independent:

```
(fail) drops the worktree advice once isolation_env_id is already cleared
Expected to not contain: "/worktree remove"
Received: "...Run `/worktree remove` to detach and go back to the project root..."
219 pass, 1 fail
```

The `beforeEach` was worth proving too, since it fixes a hazard rather than a live bug. Removing it and having one test leak a broader predicate broke **7 tests across two later describes**, including the `/update-project` suite the review named. Restoring only the `beforeEach`, with the same leak still in place, returned 220 pass / 0 fail — so the reset is demonstrably the thing holding that up, not luck.

`bun run validate` and the test suite results are in the PR checks.

## One thing to know about a sibling PR

#2665 covers the slice this PR deliberately doesn't: a scoped chat with **no** cwd override whose `codebase.default_cwd` has been deleted. Your guard is `cwd !== null`, that one is `cwd === null`, so they're disjoint and complementary — this PR is not redundant, and #2665 is marked as depending on it. That gap was originally found by @waaaw in #2114. The two refusal messages were deliberately kept coherent: `/setproject` is the common escape hatch, `/worktree remove` appears only where it actually works, and `/update-project` appears only in #2665 where it actually repairs something.

Thanks for a genuinely good PR. The `posix_spawn` root cause was the hard part and it was correct — I reproduced it independently while reviewing (`spawnSync` into a deleted directory does report `ENOENT` against the executable's path, not the directory's). Review comments and pushback are welcome; if you'd rather revert any of this and do it your own way, say so and we'll work from that.

PR #2552, opened 2026-08-17
fix(db): a schema upgrade can no longer crash-loop on an index that predates its column

Wirasm:
## Problem and outcome

A statement that names a column only works where the column exists — and on an upgrade it does not exist yet. `CREATE TABLE IF NOT EXISTS` is a no-op on a database that already has the table, so a column declared in that body arrives later, when the additive `ALTER TABLE ... ADD COLUMN` block runs. An index or `COMMENT ON COLUMN` written next to its table therefore passes every fresh-install check and aborts the whole apply on an upgrade — and since `initSchema()` applies the file in one transaction and re-throws at `fatal`, that one statement crash-loops every boot. #2513 removed one instance of this; five more remained on Postgres and three were live on SQLite.

- **Outcome:** a database created by any released Archon version upgrades to the current schema and lands on exactly the fresh-install shape, on both dialects. On SQLite this fixes a real failure, not a latent one: `new SqliteAdapter(...)` threw `no such column: parent_conversation_id` on a database created before that column existed, so the adapter could not open it at all.
- **Invariant:** no statement above the additive `ADD COLUMN` block may name a column. Every `CREATE INDEX` and `COMMENT ON COLUMN` lives below it.
- **Scope boundary:** no schema objects change. Same tables, columns, indexes, defaults, sequences and comments — only where the statements sit. No application or query code is touched.
- **Root cause:** statement *placement*, not statement content. The two dialects reproduce the same hazard for different mechanical reasons: on Postgres the file is applied top-to-bottom, and on SQLite `createSchema()` runs before `migrateColumns()`.

## Review guidance

- **Feedback requested:** correctness of the SQLite move (does any path lose an index?), and whether the CI baseline set is the right support promise.
- **Start here:** `migrations/000_combined.sql` — the trailing "Indexes and column comments" section, and the header comment stating the rule.
- **Review order:** `migrations/000_combined.sql` → `packages/core/src/db/adapters/sqlite.ts` → `scripts/check-schema-upgrades.ts` → `.github/workflows/test.yml` → tests.
- **Lower-attention areas:** the bulk of the SQL diff is 34 statements moving verbatim, and `bundled-schema.generated.ts` is regenerated. Both are covered by evidence rather than reading: a fresh apply of the old file and the new file produce byte-identical `information_schema.columns`, `pg_indexes`, `pg_constraint`, `pg_description` and `pg_sequences`.
- **Known risk or uncertainty:** the new CI job needs a PostgreSQL service container and `fetch-depth: 0` (baselines are read from release tags). If tags are ever unreachable in CI, the job fails loudly rather than skipping.
- **One deliberate trade on SQLite:** statements moved into `migrateColumns()` are inside its per-table `try`/`catch`, so an environmental failure (disk full, `SQLITE_BUSY`) during index creation is now logged with `allApplied = false` and the app keeps booting, where a failure inside `createSchema()` would have thrown out of the constructor. That is the established pattern for every other index in that function (`idx_conversations_user_id` has behaved this way since v0.4.0), and the thing it replaces is a permanent unrecoverable crash-loop — a caught, logged, self-healing degraded state is the better end of that trade. The `idx_conversations_codebase` drop/re-create is also now conditional, so the window exists only on the single boot that migrates the old non-partial index rather than on every open.

## Solution

The Postgres file now ends with one "Indexes and column comments" section holding all 28 indexes and all 6 column comments, below every `ADD COLUMN`. That makes the failure unrepresentable instead of something each author has to remember: there is no longer a "right" place to put an index that is also the wrong place. `COMMENT ON TABLE` stays with its table — it names no column.

SQLite carries the same rule through its own mechanism. `createSchema()` runs before `migrateColumns()`, so the three indexes referencing `migrateColumns()`-added columns (`hidden`, `deleted_at`, `parent_conversation_id`) move into `migrateColumns()` beside their `ALTER TABLE`, exactly as `idx_conversations_user_id` already did after the v0.4.0 incident.

Two guards, deliberately different in kind. `migration-statement-order.test.ts` now asserts the *structure* (every `CREATE INDEX` and `COMMENT ON COLUMN` sits below the section header) rather than one column's name, so it cannot rot into a tautology. `scripts/check-schema-upgrades.ts` answers the question the static test cannot: it applies the current schema on top of databases built from real release tags and diffs the result against a fresh install, catching both halves of the failure — the crash, and an upgrade that survives while quietly losing an index.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Postgres: upgrading a database created before a given column existed aborts the apply. SQLite: `SqliteAdapter` construction throws on such a database | Both dialects apply cleanly and converge on the fresh-install schema |
| **Failure behavior** | `ERROR 42703: column "…" does not exist` / `no such column: …` at boot, repeating forever with no way forward but hand-patching SQL | The class cannot be written into the schema; CI fails on the PR that would introduce it |

## Architecture

```mermaid
flowchart TB
  subgraph Before
    B1[CREATE TABLE + its indexes] --> B2[additive ADD COLUMN]
    B1 -. indexes a column<br/>that does not exist yet .-> BX[42703 · whole apply rolls back]
  end

  subgraph After
    A1[CREATE TABLE] --> A2[additive ADD COLUMN] --> A3[Indexes and column comments]
    A3 --> A4[every column exists · apply commits]
  end
```

### Changed seams

|---|---|---|
| `000_combined.sql → PostgreSQL` | All column-naming statements moved below the additive block | Fresh-apply catalog diff old vs new is empty across columns, indexes, constraints, comments, sequences |
| `createSchema() → migrateColumns()` (SQLite) | Three indexes move across the boundary to follow their `ALTER TABLE` | `sqlite.test.ts` — new test fails without the fix with `no such column: parent_conversation_id` |
| `release tag schema → current schema` | Newly asserted contract: every released version upgrades and converges | `bun run check:schema-upgrades` — six baselines |
| `migration-statement-order.test.ts → 000_combined.sql` | Guard generalized from one column to the file's structure | Fails when a probe `CREATE INDEX` is inserted above the section |

## Validation

- `bun run validate` — exit 0 — all ten steps green.
- `bun run check:schema-upgrades` — exit 0 — `v0.8.0`, `v0.7.1`, `v0.7.0`, `v0.6.0`, `v0.5.0` and `v0.2.13` (the oldest tag carrying the schema) each apply the current schema, re-apply it idempotently, and match a fresh install on columns, indexes and comments.
- **The checker has teeth:** with `v0.7.0` as the baseline, `v0.8.0`'s schema reproduces #2508 exactly — `ERROR: column "event_order" does not exist`, exit 3. Before this PR, the same v0.2.13 baseline died on `hidden`; that is the #2443 instance.
- **Equivalence proof:** applying `dev`'s file and this branch's file to two fresh databases yields identical `information_schema.columns`, `pg_indexes`, `pg_constraint`, `pg_description` and `pg_sequences`. Nothing was lost in the move.
- **SQLite red/green:** the new regression test fails on `dev`'s adapter with `SQLiteError: no such column: parent_conversation_id` and passes here; the three indexes exist afterwards.
- **Static guard red/green:** inserting a probe `CREATE INDEX` above the trailing section fails `every CREATE INDEX lives in that trailing section`.
- **Not verified:** no live server was booted against an upgraded database — the schema apply is the whole surface here and it is exercised directly, transactionally, the way `initSchema()` does it. The CI job itself is unproven until it runs on this PR.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None required. Fresh installs are byte-identical in outcome; upgrades that previously aborted now complete | Catalog equivalence + six-baseline matrix |
| Rollout / rollback | `git revert` restores the previous layout; no data is written or migrated by this change | Statement-placement change only |
| Observability | A reintroduction fails in CI (`schema-upgrade` job) instead of at a user's boot; boot failures still log `db.postgres_schema_init_failed` with the SQLSTATE | `.github/workflows/test.yml` |
| Documentation / communication | `AGENTS.md` states the placement rule and the upgrade check; the database reference tells operators upgrades are verified and how to run the check themselves | `AGENTS.md`, `packages/docs-web/src/content/docs/reference/database.md` |

## Links

- Closes #2443
- Related #2508, #2513
- Supersedes #2443 — @miiwins found this class independently and fixed `idx_conversations_hidden` and `idx_sessions_parent`; both are covered here, along with the three remaining Postgres instances, the `COMMENT ON COLUMN` variant, and the SQLite side.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQLite upgrades from older database versions by restoring missing conversation fields and indexes.
  - Improved schema upgrade reliability for workflow and conversation data.
  - Added safeguards to ensure upgrades are repeatable and consistent across existing and fresh databases.

- **Documentation**
  - Documented schema upgrade verification procedures and migration ordering requirements.
  - Added guidance for validating upgrades against older database versions and fresh installations.

- **Chores**
  - Added automated PostgreSQL schema-upgrade checks to continuous integration.
  - Added a command for manually running schema-upgrade validation.


Wirasm:
# PR Review: #2552 — fix(db): a schema upgrade can no longer crash-loop on an index that predates its column

## Outcome

The PR moves every statement that names a column below the additive `ADD COLUMN` block — 28 indexes and 6 column comments into a trailing section on Postgres, and three indexes from `createSchema()` into `migrateColumns()` on SQLite — then adds a real-PostgreSQL upgrade check and a CI job that runs it against six release-tag baselines.

Both reviewers verified the substance independently rather than reading it: the code reviewer applied `origin/dev`'s file and this branch's file to two fresh databases and found `pg_indexes`, `pg_constraint`, `information_schema.columns`, table comments and column comments byte-identical, and separately reproduced the SQLite crash on `dev` (`no such column: parent_conversation_id`, thrown out of `initSchema()` with no try/catch, so permanent) and its absence here. The SQLite half is a live bug fix, not a latent one.

Nothing is wrong with the mechanism. What is wrong is what the PR *claims*: two documentation statements are now false — one that this PR made false, one it introduced — and both are load-bearing, because they tell a contributor when they are safe. Those are the blocking findings.

## Validation

|---|---|---|
| `bun run validate` (PR head 6d7bbc93) | PASS | Exit 0, all ten steps |
| `bun run check:schema-upgrades` (local Postgres 18) | PASS | Six baselines: `v0.8.0`, `v0.7.1`, `v0.7.0`, `v0.6.0`, `v0.5.0`, `v0.2.13` — each applies, re-applies idempotently, converges on the fresh schema |
| CI `schema-upgrade` job | PASS | 33s. Log confirms the work actually ran: `fresh install OK (55 indexes)` plus all six `ok <tag> → current` lines |
| CI `test (ubuntu-latest)`, `docker-build`, `Build docs site` | PASS | — |
| CI `test (windows-latest)` | FAIL — not PR-caused | `generate-bundled-defaults.test.ts:70` timed out at 5000 ms. The PR does not touch that file, and its five siblings in the same file passed in 546–969 ms. Same cluster as #2306 (subprocess-spawning tests under `bun --filter --parallel`); re-run in flight |
| Independent catalog diff, `dev` vs PR (fresh apply) | PASS | Five catalogs identical — nothing lost in the move of 34 statements |
| Independent SQLite reproduction | PASS | `dev`'s `createSchema()` throws on an old-shaped database; the PR's adapter constructs, adds all three columns and all three indexes, and stamps `allApplied=true` |

## Critical Issues (0)

## Important Issues (2)

| Agent | Finding | Evidence | Required change |
|---|---|---|---|
| `prp-core:seam-analyzer` | The PR makes `bun run validate` no longer equivalent to CI, and leaves the blanket promise that it is. A contributor reading the Pre-PR Validation section is told local green means CI green; with `check:schema-upgrades` running only as a CI job, that is now false | `AGENTS.md:137` "Every step must pass for CI to succeed" vs `package.json:40` (`validate` omits `check:schema-upgrades`) and the new `schema-upgrade` job in `.github/workflows/test.yml`. The PR adds the caveat 30 lines below at `AGENTS.md:167` but never amends line 137. `CONTRIBUTING.md:26` ("run the full validation suite") is untouched and equally false | Amend `AGENTS.md:137` and `CONTRIBUTING.md:26` to name the one check that lives outside `validate` and why |
| CodeRabbit | The upgrade check is documented as proving the result "matches a fresh install exactly", but `CATALOG_QUERIES` compares only columns, indexes and comments — not constraints, sequences or triggers. The claim is broader than the evidence | `scripts/check-schema-upgrades.ts:109-118` vs `AGENTS.md:168` and `packages/docs-web/src/content/docs/reference/database.md:40-42` | Extend `CATALOG_QUERIES` with constraints and sequences (both were compared by hand during development, so the check is provably passable), or narrow both claims to the three catalogs actually compared |

## Suggestions (5)

| Agent | Suggestion | Evidence | Why consider it |
|---|---|---|---|
| `prp-core:code-reviewer` | The scratch database `archon_upgrade_fresh` leaks on the script's first failure gate. `process.exit()` inside `try {} finally { rmSync(...) }` skips the `finally`, and the early exit fires before the trailing `DROP DATABASE` | `scripts/check-schema-upgrades.ts:~165` (exit) vs `:~229` (drop). Reproduced: a deliberate syntax error in the SQL left `archon_upgrade_fresh` behind in `psql -l` | Ironic placement — it leaks precisely on the failure the script exists to catch. Harmless in CI's ephemeral container, accumulates on a developer's persistent Postgres |
| `prp-core:seam-analyzer` | The SQLite invariant is enforced by prose plus one test hardcoding the three columns this PR fixed; the Postgres side gets a structural scan that catches any future violation on any table | `migration-statement-order.test.ts:56-66` (regex over the whole file) vs `sqlite.ts:766-772`, `:776-781` (two `NOTE:` comments) and the new `sqlite.test.ts` case | A new column+index pair added back into `createSchema()` in the same anti-pattern is caught on one dialect and not the other |
| CodeRabbit | The new job has no `permissions:` block, so its token inherits repository defaults, and `actions/checkout` persists that token in local git config. The checker only reads local history | `.github/workflows/test.yml:70` — compare the `docker-build` job, which does declare `permissions` | One-line hardening consistent with the precedent already in this file |
| `prp-core:seam-analyzer` | The baseline set (5 most recent tags + the oldest carrying the file) is a sample, not a check of the stated guarantee that "any writer may apply any vintage at any time"; a regression between two unsampled intermediate tags would pass | `scripts/check-schema-upgrades.ts:37,82-92` vs `AGENTS.md:162` | Worth stating the sampling policy where the promise is made, so the gap is a known one rather than an assumed absence |
| `prp-core:seam-analyzer` | Authoring foot-gun: the file's established convention is to append each new feature's `ALTER TABLE` near the bottom, but the trailing section must now stay last. The guard catches it loudly (`lastAddColumnOffset()` exceeding the section fails the test), so this is ergonomics, not correctness | `migrations/000_combined.sql:340-424` (six prior append-at-the-bottom examples) vs the new trailing section at the end of the file | A line in the section header telling authors new `ADD COLUMN`s go *above* it would prevent the confusing failure |

## External review (CodeRabbit) — all six findings, with disposition

| Finding | Disposition |
|---|---|
| Restrict and avoid persisting the workflow token (`test.yml:70`) | **Accepted** — listed as a suggestion above. `docker-build` already sets `permissions`; the new job should match and add `persist-credentials: false` |
| Limit the convergence claim to checked catalog objects (`AGENTS.md:168`) | **Accepted, escalated to Important** — the claim is the PR's central promise, so the fix is to widen `CATALOG_QUERIES`, not to narrow the sentence |
| Preserve `DATABASE_URL` query options (`check-schema-upgrades.ts:59`) | **Accepted** — decomposing the URL into `PG*` vars silently drops `?sslmode=…` and friends. Passing the URI to `psql -d` with only the database name substituted preserves every option and deletes code |
| Preserve the shipped index during the SQLite upgrade (`sqlite.ts:320`) | **Refuted as described, but worth acting on.** The code reviewer reproduced the actual behavior: a failing statement mid-`db.run()` leaves earlier statements committed either way, so the drop/re-create window is byte-for-byte the one that already existed inside `createSchema()` — the move did not create it. Renaming the index (CodeRabbit's proposed fix) would fork the index name across dialects for no gain. The real defect it points at is that the pair runs unconditionally on **every** adapter open, rebuilding the index each time; making it conditional on the index not already being partial removes both the cost and the window |
| Use `@archon/git` for Git commands (`check-schema-upgrades.ts:47`) | **Declined, with a reason to record in the file.** The script is deliberately dependency-free so the CI job runs it without `bun install`; `@archon/git` covers worktrees/branches/repos, not `git show <ref>:<path>`. AGENTS.md's fallback rule is about avoiding a shell — `spawnSync` with an argument array is stricter than `execFileAsync`, and no ref is ever shell-interpolated |
| Add explicit callback annotations (`check-schema-upgrades.ts:84`) | **Not applicable.** The three callbacks are contextually typed by `Array.prototype.filter`, `type-check` and `lint --max-warnings 0` are green, and Prettier itself strips the parens that would carry an annotation |

## Strengths

- The SQL move was verified by catalog equivalence rather than by reading 34 relocated statements — five catalogs, byte-identical between `dev` and the branch on a fresh apply.
- The new check was shown to have teeth before being trusted: pointed at the pre-#2513 schema with `v0.7.0` as the baseline, it reproduces #2508 and exits 1. A green upgrade check that cannot fail would have been worse than no check.
- The static guard was generalized from one column name to the file's structure, and mutation-tested in both directions: rewording the section header and appending an `ADD COLUMN` after the section each fail loudly.
- The SQLite change follows the precedent the same file established after the v0.4.0 incident (`idx_conversations_user_id`) instead of inventing a second pattern, and the `allApplied` semantics were confirmed self-healing across re-opens.
- The upgrade check runs in 33s in CI and its log proves it did the work, rather than passing vacuously.

## Verdict

**NEEDS FIXES**

Two false claims, both one-line-ish: `AGENTS.md:137` and `CONTRIBUTING.md:26` still promise `bun run validate` covers CI, and the "matches a fresh install exactly" claim outruns what `CATALOG_QUERIES` compares. Fix the first by naming the exception; fix the second by extending the comparison to constraints and sequences, which is the option that makes the strong claim true rather than retreating from it. The mechanism itself is correct and independently verified — no changes needed to the SQL, the SQLite move, the CI job, or the guards. The suggestions, particularly the scratch-database leak, are worth folding into the same push.

Wirasm:
## Review findings applied — `0d58a8e9`

Both Important findings, all four actionable suggestions, and every CodeRabbit finding are addressed. One of them found a real bug.

**Widening the catalog comparison found a genuine divergence.** Adding `constraints` and `sequences` to `CATALOG_QUERIES` — so the "matches a fresh install" claim is true rather than aspirational — immediately failed two baselines: upgrading from `v0.5.0` or older leaves `remote_agent_codebases.kind` without the `CHECK (kind IN ('repo','folder'))` that its `CREATE TABLE` body declares, because the additive `ADD COLUMN` omits it. Fresh installs enforce it; upgraded ones accept any value.

That one is sanctioned rather than a bug — AGENTS.md already states that a constraint declared in a `CREATE TABLE` body binds only databases created after it, and SQLite cannot add a CHECK by `ALTER` at all, so enforcing it on Postgres alone would break dialect parity. It now lives in a small `ALLOWED_DRIFT` allowlist with its reason, printed on **every** run rather than silently skipped:

```
ok   v0.5.0 → current (idempotent, converges on the fresh schema)
     allowed divergence [constraints] remote_agent_codebases remote_agent_codebases_kind_check CHECK …
     sanctioned: a CHECK declared in a CREATE TABLE body binds only databases created after it …
```

**The rest:**

| Finding | Change |
|---|---|
| `bun run validate` no longer equals CI, and both docs still promised it did | `AGENTS.md` Pre-PR Validation and `CONTRIBUTING.md` now name the one check that lives outside it and why |
| Scratch database leaked on the fresh-install failure path (`process.exit()` skips `finally`) | Created databases are tracked and dropped in a `finally`; the early exits throw instead |
| `DATABASE_URL` query options silently dropped | The URI is passed to `psql` with only the database name replaced, so `?sslmode=…` survives |
| `idx_conversations_codebase` dropped and rebuilt on every SQLite open | Replaced only when the existing index is the older non-partial one — no per-open rebuild, no window |
| Job token unrestricted and persisted | `permissions: contents: read` and `persist-credentials: false` |
| SQLite invariant was prose plus a three-column test | New structural guard: no index in `createSchema()` may name a column `migrateColumns()` adds. Verified red/green — a probe index on `hidden` fails it with the offending statement |
| `@archon/git` not used for git calls | Declined, with the reason recorded in the file: the script must run in CI with no `bun install`, and `spawnSync` with an argument array never reaches a shell |
| Explicit callback annotations | Not applicable — contextually typed by `Array.filter`; `type-check` and `lint --max-warnings 0` are green, and Prettier strips the parens an annotation would need |
| Authoring foot-gun: new `ADD COLUMN`s appended after the trailing section | The section header now says to add them above it, and the guard fails loudly either way |

`bun run validate` exit 0 and `bun run check:schema-upgrades` green on all six baselines after these changes.

**On the earlier windows-latest failure:** unrelated flake — `generate-bundled-defaults.test.ts` timed out at 5000 ms while its five siblings in the same file passed in 546–969 ms, in a file this PR does not touch. It passed on a plain re-run. Recorded with evidence on #2306, which tracks this cluster.

Wirasm:
# PR Review: #2552 (round 2 — commit `0d58a8e9`)

## Outcome

Round 1 reviewed `6d7bbc93` and cleared the mechanism. This round covers `0d58a8e9`, the commit that answered round 1: it widened the upgrade comparison to constraints and sequences, added the `ALLOWED_DRIFT` allowlist for the one sanctioned divergence that widening exposed, made scratch-database cleanup exception-proof, replaced the `DATABASE_URL` decomposition, made the SQLite index replacement conditional, and added the structural guard the SQLite side was missing.

Most of it holds up under direct execution. Both reviewers independently ran the script against real release tags on PostgreSQL 18, re-derived the structural guard's column set, tested `connectionFor()` across eight URL shapes, and traced the conditional index replacement through all three database states. The conditional replacement, `connectionFor()`'s handling of every standard URL shape, the guard's soundness, and the doc-to-code claims all check out, and two of CodeRabbit's five findings are false positives.

The blocking problems are both in `ALLOWED_DRIFT` — the mechanism this commit added to keep a known divergence from failing CI. It is looser than the divergence it describes in two independent ways, each of which lets a *different*, real defect pass as sanctioned. An allowlist that silently absorbs the next real bug is worse than the failing check it replaced.

## Validation

|---|---|---|
| CI on `0d58a8e9` — all six checks | PASS | ubuntu, **windows (first try)**, docker-build, docs, schema-upgrade (37s), CodeRabbit |
| CI `schema-upgrade` job log | PASS | Six baselines green; the sanctioned `kind_check` divergence printed under both `v0.5.0` and `v0.2.13` rather than swallowed |
| `bun run validate` (local, this tree) | PASS | Exit 0 |
| `bun run check:schema-upgrades` (local) | PASS | Six baselines; allowlist output as designed |
| `bun run check:schema-upgrades v0.3.12 v0.4.1` | PASS | The two schema vintages the default sample misses both upgrade clean |
| Independent re-run against v0.4.1 / v0.5.0 / v0.6.0 / v0.7.0 / v0.8.0 | PASS | Both agents; only the "missing" direction fires today, v0.7.0+ show zero drift |
| `packages/core/src/db/migration-statement-order.test.ts` | PASS | 6 tests, 46 assertions |

## Critical Issues (0)

## Important Issues (2)

| Agent | Finding | Evidence | Required change |
|---|---|---|---|
| `prp-core:code-reviewer` | `ALLOWED_DRIFT` is direction-blind, so it also sanctions the opposite of the divergence it documents. The sanctioned case is "the upgraded database lacks a CHECK the fresh one has"; the same entry equally excuses a fresh install that *lost* it | `scripts/check-schema-upgrades.ts:154-156` (`allowedDrift()`) called identically from the missing-after-upgrade and extra-after-upgrade loops at `:256-267`. Reproduced standalone: change the `kind` CHECK's definition and both the fresh-only and upgraded-only lines match the same prefix, yielding `differences: []` | Give each entry a `direction: 'missing' \| 'extra'` and check it against the loop that produced the line |
| `prp-core:seam-analyzer` | The prefix match is looser than the constraint identity, and Postgres's own naming makes the gap reachable. A future unnamed CHECK on the same table is auto-named `…_kind_check1`, which `startsWith('…_kind_check')` accepts — so an unrelated new constraint is silently absorbed by an allowance written for a different one | `scripts/check-schema-upgrades.ts:143-152` (prefix) vs `:154-156` (`startsWith`). Verified live on PG18: `ALTER TABLE … ADD CHECK (kind IS NOT NULL)` produced `remote_agent_codebases_probe_kind_check1` | Match the full constraint identity exactly rather than by prefix |

## Suggestions (6)

| Agent | Suggestion | Evidence | Why consider it |
|---|---|---|---|
| `prp-core:seam-analyzer` | `ALLOWED_DRIFT` claims kinship with the SQLite/Postgres parity allowlist "in spirit" but skips the part that matters: the sibling has a test asserting every entry still describes real, current drift, so an exception cannot outlive its reason | `scripts/check-schema-upgrades.ts:137-141` (the claim) vs `packages/core/src/db/adapters/sqlite.test.ts:719-752` (`parity allowlists still describe real drift`, whose own comment explains why) | If the CHECK is ever dropped from the schema, the entry becomes permanently dead with no signal. On a default-baseline run the oldest baseline guarantees the drift fires, so "every entry was exercised" is a cheap, self-expiring counterpart |
| `prp-core:code-reviewer` | The "printed on every run" promise is false in the one case that matters most: when a ref has both a real difference and an allowed one, the `continue` in the failure branch fires before the allowed lines are printed | `scripts/check-schema-upgrades.ts:270-276` (`continue`) vs `:280` (the print), against the claim at `:141` and in `AGENTS.md` | Print the allowed lines before the differences check, so a failing ref still shows what was excused |
| `prp-core:code-reviewer` | `git()` swallows launch failures, and the resulting message actively misdirects. With `git` off `PATH`, every call returns `{ok: false}` and the script reports `skip <tag>: no migrations/000_combined.sql at that ref` → "none of the requested refs carry the schema file" | `scripts/check-schema-upgrades.ts:51-54` vs `psql()` at `:74-81`, which checks `r.error` and reports it distinctly. Reproduced by removing git from `PATH` | It fails safe, but it accuses the repository of missing a file that is present. One `r.error` check restores symmetry with `psql()` |
| `prp-core:code-reviewer` | The `finally` cleanup loop ignores each `DROP DATABASE`'s exit code, so a failed cleanup leaks silently and the script can still exit 0 | `scripts/check-schema-upgrades.ts:288-291` | Self-limiting (the next run's `DROP … IF EXISTS` collects it), but a warning costs one line and this is the same class of silence the PR is otherwise closing |
| `prp-core:seam-analyzer` | `connectionFor()` cannot represent a libpq multi-host URI, while the docstring promises "every other option survives" | `scripts/check-schema-upgrades.ts:24-26`, `:56-60` (the promise) vs `:61-67` (`new URL(url)`). Verified: `postgresql://user@host1:5432,host2:5432/db` throws `TypeError … cannot be parsed as a URL` | The `URL` limitation predates the commit; the unqualified promise is new. Either qualify the docstring or fall back to `PG*` when parsing fails |
| both | The structural guard slices from `private createSchema()` to end of file, which bounds the exec block correctly only because that method is positionally last in `sqlite.ts`. Nothing enforces the ordering | `packages/core/src/db/migration-statement-order.test.ts:115` vs `packages/core/src/db/adapters/sqlite.ts:542` (method start) and the class's closing brace | Fails loudly rather than silently if a method is ever added after it, so a one-line comment recording the assumption is enough |

## Refuted (checked, not real)

| Claim | Why it does not hold |
|---|---|
| CodeRabbit: add the test callback return type (`migration-statement-order.test.ts:101`) | Root `eslint.config.mjs` ignores `**/*.test.ts`; verified via `eslint` reporting "File ignored because of a matching ignore pattern". `explicit-function-return-type` never reaches the file, and no callback in it is annotated |
| CodeRabbit: scratch databases dropped by fixed name (`recreate()`) | Accurate as a description, but present verbatim in `6d7bbc93` and cleared in round 1 — this commit only changed `process.exit()` to `throw` and added the tracking `Set`. Still worth a unique-suffix fix at PR level, not a round-2 regression |
| `connectionFor()` mishandles common URL shapes | Verified across no path, `/`, query strings, percent-encoded userinfo, bracketed IPv6, missing port and `?host=`-style unix sockets. Round-trips correctly, and avoids the encode/decode mismatch the old `psqlEnv()` risked |
| The conditional `idx_conversations_codebase` replacement is unsafe | Traced all three states plus a second open against a real SQLite database: no index → create; old non-partial → drop then create; already partial → both statements no-op. `sqlite_master.sql` stores the CREATE text verbatim, so the predicate check is reliable |
| The structural guard is nominal rather than real | Independently re-derived `addedByMigration` (matches the real set exactly, no comment-prose false matches) and caught an injected offending index on `output_root`. Multi-line `CREATE INDEX … WHERE` statements are handled |
| Uncompared object classes (triggers, RLS, collations) are a hidden gap | The system's only trigger, `archon_workflow_event_notify` (`packages/core/src/db/adapters/postgres.ts:18-27`), is deliberately outside the bundled schema and applied identically to fresh and upgraded databases by the adapter — symmetric, so invisible to a fresh-vs-upgraded diff by construction |

## Strengths

- Widening the comparison was the right response to round 1: it converted a claim into a check and immediately found a real divergence nobody knew about.
- The divergence was classified rather than papered over — traced to an existing AGENTS.md rule and to SQLite's inability to add a CHECK by `ALTER`, which is why suppressing it beat "fixing" one dialect.
- The SQLite structural guard genuinely closes round 1's asymmetry finding: an injected offending index is caught, and the derived column set matches `migrateColumns()` exactly.
- Windows passed on the first attempt this time, and the CI job log shows the allowlist working end to end rather than the job merely exiting 0.

## Verdict

**NEEDS FIXES**

Both blocking findings are in `ALLOWED_DRIFT`, and they are the same mistake twice: the allowance is described more loosely than the thing it allows. Match the constraint identity exactly, and record which direction is sanctioned. Then the cheap self-expiry check — on a default-baseline run, every entry must have fired at least once — turns the allowlist from a permanent hole into one that closes itself when its reason disappears. The remaining suggestions are one-liners in the same file and worth folding into the same push; nothing in the PR's core mechanism needs to change.

Wirasm:
## Round-2 findings applied — `a804bfa2`

Both Important findings are closed by **deleting** the mechanism rather than hardening it, and each fix is verified by mutation rather than assertion.

### The allowlist is gone

`ALLOWED_DRIFT` was a framework — entries, prefixes, per-entry directions — built for exactly one exception, and both defects were consequences of that generality. It is now that one exception: a single constant matched on the exact `<table> <conname>` identity, consulted only in the direction where the upgraded database is the one missing it. An upgrade that *gains* anything always fails; there is no per-entry direction field to set wrong, and no prefix semantics to get wrong.

The reason it stays excused is also stronger than what I wrote before. It is not primarily about dialect parity: `ALTER TABLE … ADD CONSTRAINT` validates existing rows, so repeating the CHECK in the additive block would let a single unexpected value abort the entire apply — the crash-loop class this check exists to prevent.

**Four mutations, each of which must fail and now does:**

| Mutation | Before | Now |
|---|---|---|
| A second unnamed CHECK on `kind`, which Postgres auto-names `…_kind_check1` | silently sanctioned by the prefix | `FAIL … missing after upgrade [constraints]: … _kind_check1 CHECK (((kind)::text <> ''::text))` |
| Remove the CHECK from the `CREATE TABLE` body, so a v0.6.0 upgrade *retains* a constraint the fresh install lacks | silently sanctioned in the extra direction | `FAIL … extra after upgrade [constraints]: … _kind_check …` |
| Rename the exception's target | — | the real drift stops being excused and fails loudly |
| Make upgrades converge (add the CHECK in the additive block), so the exception's reason disappears | entry would sit dead forever | `FAILED: the expected divergence (…) never appeared. It is stale.` |

That last one is the self-expiry the seam analyzer asked for, and it closes the gap against the sibling parity allowlist, which has had a staleness test all along.

### Baselines: every distinct schema, not a sample

Tags that did not touch the schema ship the same file, so the old "five recent plus oldest" sample was both arbitrary and incomplete — it missed the schema that **fifteen releases** shipped (`v0.3.0`–`v0.3.12`, `v0.4.0`, `v0.4.1`). Deduplicating by blob turns 21 tags into 8 baselines, which is every vintage an install can actually have, for about two seconds more:

```
note: 1 tag(s) predate migrations/000_combined.sql and cannot be a baseline
note: 21 tag(s) carry 8 distinct schema version(s)
ok   v0.2.13 / v0.3.0 / v0.4.0 / v0.5.0 / v0.6.0 / v0.7.0 / v0.7.1 / v0.8.0 → current
```

`RECENT_BASELINES` is gone with it — the set now grows with schema churn instead of release cadence.

### The remaining suggestions

`git` launch failures are reported as such instead of as a missing schema file; failed scratch-database cleanup warns rather than passing silently; expected divergences print *before* the verdict, so a ref with both a real and an expected difference still shows what was excused (with the ref labelled, since the line no longer sits under its `ok`); the docstring records the one libpq shape `connectionFor()` cannot represent, and fails with a pointer to `PG*` instead of a bare `TypeError`; and the structural guard records that its slice assumes `createSchema()` stays the last method.

Two claims from the round-2 batch were **not** acted on, both refuted with evidence in the review: the test-callback return type (root eslint config ignores `**/*.test.ts`) and the fixed-name scratch databases (present in the already-cleared first commit; the tracking `Set` and `DROP … IF EXISTS` before every create bound the impact).

`bun run validate` exit 0 and `bun run check:schema-upgrades` green across all eight baselines.

One process note: I folded the two changes into a single commit rather than splitting them, because they are the same 40 lines of one script and would never be reverted independently — reverting the dedupe alone would leave the staleness check depending on a sample.

Issue #2556, opened 2026-08-17
refactor(workflows): one spelling for reasoning depth — effort: works on every provider, modelReasoningEffort: leaves the workflow schema


Wirasm:
Split out of the review of the now-closed #2254 (full evidence: https://github.com/coleam00/Archon/pull/2254#issuecomment-5314108113). The narrow forwarding bug is #2246; this is the design question that review surfaced and deliberately did not fold in.

## Problem

The workflow language has two workflow-level ways to say "reason harder":

- `effort:` — cross-provider (`low|medium|high|max`, `schemas/dag-node.ts:40`), settable per-node and workflow-level. Codex is `effortControl: false` (`codex/capabilities.ts:16`), so on a Codex node it hits the capability gate at `dag-executor.ts:1065-1086`, warns, and is dropped.
- `modelReasoningEffort:` — Codex-only (`minimal|low|medium|high|xhigh`, `schemas/workflow.ts:20`), workflow-level only, no node-level equivalent.

An author must know which one their resolved provider wants, and the resolved provider is not knowable at load time (it can come from a node `provider:`, a tier, or an `@alias`). Picking wrong fails loudly in one direction and silently in the other.

## Verified

**The root cause is a mis-scoped capability axis, not a missing field.** `effortControl` means "accepts the node-level `effort:` field", not "can apply reasoning effort". The executor's own comment says so in as many words (`dag-executor.ts:1163-1170`):

> Do NOT gate this on caps.effortControl: that flag means "accepts the node-level effort: field", not "can apply reasoning effort". Codex is effortControl:false yet applies effort via modelReasoningEffort.

Codex has reasoning depth. It reads it from a different place. The second spelling exists to route around the axis.

**The enums are closer than they look.** Claude-only value is `max`; Codex-only is `minimal`; `low|medium|high` are shared. But `max` and `xhigh` are already treated as the same rung: `parseCopilotConfig` maps `max` → `xhigh` at parse time with the comment *"Accept Archon's workflow-schema alias for the top tier"* (`community/copilot/config.ts:24-29`). So the reconciliation precedent exists, at the provider boundary, which is where a vocabulary difference belongs. That leaves `minimal` as the only genuinely provider-specific value — one extra rung at the bottom, not two incompatible vocabularies.

[...]

Issue #2395, opened 2026-08-03
feat(workflows): effort is resolved then discarded — never reaches the event stream


Wirasm:
Split out of #2393 as an independently shippable gap.

## Verified

`effort` is resolved during node execution — `routePresetEffort(provider, preset.effort)` at `packages/workflows/src/dag-executor.ts:221` — and then never leaves the process. Measured across a real store:

```
events with a top-level 'effort' key: 0   (of 13,163)
```

`node_started` emits `{command, provider, model, tier}` (`dag-executor.ts:1284-1286`). Effort is in the same resolution path and is simply not passed to the emitter.

## Why it matters

Effort is a first-class part of what a node ran on — tier presets and `@alias` entries both carry it, and it materially changes cost and latency. A console run view or an external observer (helm) showing `codex / gpt-5.6-terra / medium` cannot say whether that node ran at `low` or `xhigh` effort. For eval it is an independent variable that is currently unrecorded, so two runs that differ only by effort are indistinguishable after the fact.

## Fix

Add `effort` to the `node_started` payload alongside `provider`/`model`/`tier`, and to the node execution record proposed on #2393.

Purely additive — `workflow_events.data` is JSON-as-TEXT, no reader breaks.

## Not verified

Whether effort is resolved on every path that emits `node_started` (bash/script nodes have no effort; the retry and resume paths were not traced). Enumerate the emit sites rather than assuming one.

Wirasm:
# Investigation: feat(workflows): effort is resolved then discarded — never reaches the event stream

**Issue**: #2395 (https://github.com/coleam00/Archon/issues/2395)  
**Type**: BUG  
**Investigated**: 2026-08-03T00:00:00+03:00

### Assessment

| Metric | Value | Reasoning |
|---|---|---|
| Severity | LOW | The execution itself correctly receives effort today; only its additive observability metadata is missing, so no workflow behavior, data integrity, or user request is broken. |
| Complexity | MEDIUM | The change touches the shared resolver, both AI and loop `node_started` paths, the typed emitter contract, and focused workflow tests, while deterministic nodes must stay untouched. |
| Confidence | HIGH | Every production `node_started` site and both effort-resolution destinations were enumerated, and the existing model/tier plumbing provides a direct precedent. |

---

## Problem Statement

[...]

PR #2559, opened 2026-08-17
fix(workflows): workflow-level modelReasoningEffort and webSearchMode now reach Codex

### LINKED ISSUE #2246: fix(workflows): workflow-level modelReasoningEffort and webSearchMode are schema-accepted but never forwarded to the provider

Wirasm:
## Problem and outcome

`modelReasoningEffort:` and `webSearchMode:` have been declared on `workflowBaseSchema` and validated by the loader since 2026-03-26, and three places in the docs describe them as working workflow-level options — but the executor never read them off the workflow definition. A workflow that declared either one parsed cleanly, warned about nothing, and ran at the `config.yaml` or SDK default.

- **Outcome:** A workflow declaring `modelReasoningEffort:` / `webSearchMode:` now has that value reach Codex, and declaring either on a non-Codex node produces the same loud capability-mismatch warning every other unsupported field produces.
- **Invariant:** The effort reported in `node_started` is the effort the node actually runs at, and a value declared for one provider never silently changes another provider's behavior.
- **Scope boundary:** No node-level fields added, no `ProviderCapabilities` axis added, `applyPresetOptions` untouched, no provider code touched. `webSearchMode` stays workflow-level with no per-node form — decided and recorded in #2556, not reopened here. Whether reasoning depth should have one workflow-level spelling or two is #2556, not this PR.
- **Root cause:** `workflowLevelOptions` (built at the top of `executeDagWorkflow`) simply never carried the two fields, so `resolveNodeProviderAndModel` had nothing to forward. `assistantConfig` was built purely from `config.assistants[provider]`.

## Review guidance

- **Feedback requested:** Correctness of the provider gate and the write's position relative to the `resolvedEffort` read.
- **Start here:** `packages/workflows/src/dag-executor.ts` — the `if (provider === 'codex')` block in `resolveNodeProviderAndModel`. Its position is load-bearing in both directions: after `applyPresetOptions` so a workflow literal beats a preset-routed effort, before the `resolvedEffort` read so telemetry is not stale.
- **Review order:** the write block → the `provider !== 'codex'` push into `unsupported` → `WorkflowLevelOptions` → the four tests.
- **Lower-attention areas:** the two-line addition to the `workflowLevelOptions` literal and the type import.
- **Known risk or uncertainty:** `resolvedEffort` now branches on the provider, so a Codex node that declares both `effort:` and `modelReasoningEffort:` reports the latter in `node_started` where it previously reported the former. That is the correction (the former is ignored by Codex), but it does change what existing runs emit. Separately, the mismatch warning fires per node, so a multi-node non-Codex workflow that declares one of these repeats it. That matches the existing capability-warning behavior rather than inventing a new one-shot mechanism.

## Solution

Thread both fields from the workflow definition into `workflowLevelOptions`, then write them onto `assistantConfig` in `resolveNodeProviderAndModel` — gated on the **resolved** provider, so a node that resolves to Codex through a tier or `@alias` is covered and a node that resolves away from Codex is not.

Two deliberate non-changes are worth flagging, because the superseded attempt made both:

**`applyPresetOptions` is untouched.** [#2254](https://github.com/coleam00/Archon/pull/2254) added `workflowLevelOptions.modelReasoningEffort !== undefined` to that function's early-return guard. That guard runs before `provider` is consulted, so a workflow declaring `modelReasoningEffort` for its Codex nodes silently suppressed tier/alias preset effort on its Claude and Pi nodes too. Once the write is positioned correctly and gated, the guard is unnecessary — the overwrite already establishes precedence — so the smallest correct change deletes the problem rather than patching it.

**The write is gated on the provider.** An ungated write reaches `parseCopilotConfig` through `resolveCopilotReasoning`, which falls back to `copilotConfig.modelReasoningEffort` when a node sets neither `effort:` nor `thinking:`. Copilot's parser accepts only `low|medium|high|xhigh`, so the documented Codex value `minimal` would silently no-op there while the others silently applied.

The mismatch warning reuses the existing `unsupported` array rather than adding a capability axis, so an author gets one combined message per node instead of a second parallel warning channel. Adding a second effort-shaped `ProviderCapabilities` axis is the wrong fix and is deferred to #2556.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Workflow-level `modelReasoningEffort:` / `webSearchMode:` parsed and were discarded; Codex ran at the config or SDK default | Both reach Codex; workflow literal beats a tier-preset-routed effort; `node_started` carries the applied effort |
| **Failure behavior** | Declaring either field on a non-Codex workflow did nothing, silently | Declaring either on a non-Codex node warns through the existing capability-mismatch message and writes nothing |

### Changed seams

|---|---|---|
| `WorkflowDefinition → workflowLevelOptions` | Both fields now carried; previously dropped | `dag-executor.ts`, test `forwards workflow-level modelReasoningEffort and webSearchMode to a Codex node` |
| `workflowLevelOptions → assistantConfig` | New provider-gated write, ordered against `applyPresetOptions` and the `resolvedEffort` read | tests `workflow-level modelReasoningEffort beats a preset-routed effort…` and `…does not strip preset effort from a Claude node` |
| `resolveNodeProviderAndModel → node_started` | The persisted event now reports the workflow-level value when it wins | assertion on `nodeStartedCall[0].data.effort` |
| `assistantConfig → CodexProvider.parseCodexConfig` | Unchanged — the consumer was already correct | no provider files touched |

## Validation

- `bun run validate` — exit 0, all 9 steps, zero test failures across all packages — proves the change is clean against `dev` @ `ebc0eae5`.
- `bun test src/dag-executor.test.ts` — 457 pass, 0 fail (453 before) — proves the four new cases pass without disturbing the existing suite.
- Each new test was mutation-tested rather than trusted on a green run. Reintroducing #2254's `applyPresetOptions` guard breaks the Claude preset-effort test; moving the write after the `resolvedEffort` read makes `node_started` report `medium` while the node runs `xhigh`; neutering the non-Codex warn breaks the warning test; removing the `provider === 'codex'` gate sets `modelReasoningEffort` on a Claude node. Each mutation breaks exactly the test intended to catch it.
- **Not verified:** no end-to-end run against a real Codex binary. The path from `assistantConfig` through `parseCodexConfig` to `buildThreadOptions` is unchanged by this PR and already covered by the provider's own tests; what this PR adds is upstream of that boundary and is covered by the assertions on `assistantConfig`.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Workflows that do not declare either field are byte-identical. Workflows that already declare them futilely now have them honored — the behavior the docs already promise | tests; `guides/authoring-workflows.md:126-127` |
| Rollout / rollback | Single commit, no schema or migration change, reverts cleanly | — |
| Observability | A previously silent no-op is now a `dag.unsupported_capabilities` warn plus a user-facing message; `node_started` gains a correct `effort` for this path | test `warns when workflow-level Codex options land on a non-Codex node` |
| Documentation / communication | Corrects a materially false line and records the new author-facing behavior. `authoring-workflows.md` said `webSearchMode: disabled` means "No web access" — it gates Codex's built-in search tool only, since `buildThreadOptions` hard-codes `networkAccessEnabled: true` on every Codex run. Also documents that both fields are workflow-level only and that a mismatch now warns | `providers/src/codex/provider.ts:85-95`; `guides/authoring-workflows.md`, `book/quick-reference.md` |

## Links

- Closes #2246
- Related #2556
- Supersedes #2254
- Supersedes #2020

## Summary by CodeRabbit

* **New Features**
  * Added workflow-level controls for model reasoning effort and web search behavior.
  * Settings apply to Codex workflows and take precedence over preset reasoning settings.
  * Added provider-aware handling for unsupported settings.
* **Bug Fixes**
  * Added warnings when Codex-specific options are configured for non-Codex workflows.
  * Improved reporting of the effective reasoning effort.
* **Tests**
  * Added coverage for option propagation, precedence, provider handling, telemetry, and warnings.

### DIFF
--- a/packages/docs-web/src/content/docs/book/quick-reference.md
+++ b/packages/docs-web/src/content/docs/book/quick-reference.md
@@ -109,8 +109,8 @@ archon workflow run my-workflow "auth refresh-tokens"
 | `nodes` | Yes | array | DAG nodes (see Node Options below) |
 | `provider` | No | string | Registered provider identifier (e.g. `claude`, `codex`). Default: `claude` |
 | `model` | No | string | Model for all nodes (`sonnet`, `opus`, `haiku`, or full model ID) |
-| `modelReasoningEffort` | No | string | Codex only: `minimal` \| `low` \| `medium` \| `high` \| `xhigh` |
-| `webSearchMode` | No | string | Codex only: `disabled` \| `cached` \| `live` |
+| `modelReasoningEffort` | No | string | Codex only, no per-node form: `minimal` \| `low` \| `medium` \| `high` \| `xhigh` |
+| `webSearchMode` | No | string | Codex only, no per-node form. Gates Codex's built-in search tool, not network access: `disabled` \| `cached` \| `live` |
 
 ### Node Options (DAG)
 
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -123,8 +123,8 @@ description: |
 # Optional workflow-level configuration
 provider: claude
 model: sonnet
-modelReasoningEffort: medium     # Codex only
-webSearchMode: live              # Codex only
+modelReasoningEffort: medium     # Codex only, workflow level only (no per-node form)
+webSearchMode: live              # Codex only, workflow level only (no per-node form)
 interactive: true                # Web only: run in foreground instead of background
 requires: [github]               # Optional: hard-block invocation unless the triggering
                                  #   user has connected their GitHub identity. Enforced only
@@ -1686,10 +1686,19 @@ webSearchMode: live             # 'disabled' | 'cached' | 'live'
 - `medium` - Balanced (default)
 - `high`, `xhigh` - More thorough, expensive
 
-**Web search mode:**
-- `disabled` - No web access (default)
+**Web search mode:** controls Codex's built-in web-search tool. It is not a network
+switch — Codex nodes always run with network access enabled, so `disabled` stops Codex
+from searching, not from reaching the network.
+
+- `disabled` - No built-in web search (default)
 - `cached` - Use cached search results
 - `live` - Real-time web search
+
+Both fields are **workflow-level only** — there is no per-node `modelReasoningEffort:` or
+`webSearchMode:`. Declaring either on a workflow whose node resolves to a provider other
+than Codex logs a warning and applies nothing, the same way any other unsupported option
+does. The reverse also holds: `effort:` on a Codex node warns and is ignored, because Codex
+takes reasoning depth only through `modelReasoningEffort`.
 
 ### Web Execution Mode
 
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -11291,6 +11291,52 @@ describe('executeDagWorkflow -- Claude SDK advanced options', () => {
     expect(assistantConfig?.webSearchMode).toBeUndefined();
   });
 
+  it('reports modelReasoningEffort in node_started, not the effort Codex ignores', async () => {
+    // A mixed-provider workflow is told by the authoring guide to set `effort:` for its
+    // Claude/Pi nodes and `modelReasoningEffort:` for Codex. nodeConfig.effort is populated
+    // for every provider and only warned about on Codex, never stripped — so the event
+    // stream must not prefer it over the value Codex actually ran at.
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'codex',
+      getCapabilities: mockCodexCapabilities,
+    }));
+    const mockDeps = createMockDeps();
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-dag',
+      testDir,
+      {
+        name: 'codex-effort-telemetry-test',
+        nodes: [{ id: 'step1', command: 'my-cmd' }],
+        effort: 'max',
+        modelReasoningEffort: 'low',
+      },
+      workflowRun,
+      'codex',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      { ...minimalConfig, assistant: 'codex' }
+    );
+
+    const optionsArg = mockSendQueryDag.mock.calls[0][3] as Record<string, unknown>;
+    const assistantConfig = optionsArg?.assistantConfig as Record<string, unknown>;
+    expect(assistantConfig?.modelReasoningEffort).toBe('low');
+
+    const createEventCalls = (mockDeps.store.createWorkflowEvent as ReturnType<typeof mock>).mock
+      .calls as Array<[{ event_type: string; data?: Record<string, unknown> }]>;
+    const nodeStartedCall = createEventCalls.find(([arg]) => arg.event_type === 'node_started');
+    expect(nodeStartedCall?.[0].data?.effort).toBe('low');
+  });
+
   it('warns when a node in a Codex workflow resolves away to Claude via a tier', async () => {
     // The mixed-provider direction: the workflow-level provider IS codex, so a check
     // against the CONFIGURED provider would stay silent. Only the resolved provider
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -11113,6 +11113,183 @@ describe('executeDagWorkflow -- Claude SDK advanced options', () => {
     const warning = messages.find(m => m.includes('effort') && m.toLowerCase().includes('codex'));
     expect(warning).toBeDefined();
   });
+
+  it('forwards workflow-level modelReasoningEffort and webSearchMode to a Codex node', async () => {
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'codex',
+      getCapabilities: mockCodexCapabilities,
+    }));
+    const mockDeps = createMockDeps();
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-dag',
+      testDir,
+      {
+        name: 'codex-workflow-options-test',
+        nodes: [{ id: 'step1', command: 'my-cmd' }],
+        modelReasoningEffort: 'minimal',
+        webSearchMode: 'live',
+      },
+      workflowRun,
+      'codex',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      { ...minimalConfig, assistant: 'codex' }
+    );
+
+    expect(mockSendQueryDag.mock.calls.length).toBeGreaterThan(0);
+    const optionsArg = mockSendQueryDag.mock.calls[0][3] as Record<string, unknown>;
+    const assistantConfig = optionsArg?.assistantConfig as Record<string, unknown>;
+    expect(assistantConfig?.modelReasoningEffort).toBe('minimal');
+    expect(assistantConfig?.webSearchMode).toBe('live');
+  });
+
+  it('workflow-level modelReasoningEffort beats a preset-routed effort, and node_started reports the applied value', async () => {
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'codex',
+      getCapabilities: mockCodexCapabilities,
+    }));
+    const mockDeps = createMockDeps();
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+    const aiProfile = buildAiProfile('claude', {
+      repoTiers: {
+        medium: { provider: 'codex', model: 'gpt-5.5', effort: 'medium' },
+      },
+    });
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-dag',
+      testDir,
+      {
+        name: 'codex-workflow-beats-preset-test',
+        nodes: [{ id: 'step1', command: 'my-cmd', model: 'medium' }],
+        modelReasoningEffort: 'xhigh',
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig,
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      aiProfile
+    );
+
+    const optionsArg = mockSendQueryDag.mock.calls[0][3] as Record<string, unknown>;
+    const assistantConfig = optionsArg?.assistantConfig as Record<string, unknown>;
+    expect(assistantConfig?.modelReasoningEffort).toBe('xhigh');
+
+    // The write must land BEFORE resolvedEffort is captured, or the audit trail and
+    // console report an effort the node did not run at.
+    const createEventCalls = (mockDeps.store.createWorkflowEvent as ReturnType<typeof mock>).mock
+      .calls as Array<[{ event_type: string; data?: Record<string, unknown> }]>;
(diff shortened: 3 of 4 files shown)


Wirasm:
---
pr: 2559
base: dev
head: fix/2246-forward-codex-workflow-options
reviewed: 2026-08-17T10:09:45Z
reviewed_head: 1c55db7c (reviewed at ad950327; see note below)
verdict: READY TO MERGE
open_findings: 0
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2559#issuecomment-5314698997
---



# PR Review: #2559 — fix(workflows): workflow-level modelReasoningEffort and webSearchMode now reach Codex

## Signal

Both reviewers found the forwarding itself correct — the provider gate closes the Copilot leak, the write's position gives workflow-level precedence over a tier preset, and the deliberate omission of the `applyPresetOptions` guard that sank the superseded #2254 is safe because `effort` and `modelReasoningEffort` are independent fields the guard never inspects. The two findings that did land share one cause: **the resolved provider is a fact this function needs in more than one place, and it was spelled as independent string comparisons.** `code` found the consequence — `resolvedEffort` (untouched by the PR) prefers `nodeConfig.effort` over `assistantConfig.modelReasoningEffort`, so on Codex it reported the declared-but-ignored `effort:` instead of the value the node actually ran at, reproduced as `runs at low / reports max`. Forwarding the workflow-level field is what made that reachable from workflow YAML alone, and the PR's own comment claimed the opposite, so it was in scope rather than pre-existing. `seams` found the same fact duplicated as inverse booleans 84 lines apart. One correction fixes both: name the provider fact once and branch the telemetry read on it. Both findings are FIXED in `ad950327`, mutation-verified, with the full suite and `bun run validate` green.

- **Blocking:** 0
- **Non-blocking:** 0
- **Resolved since previous review:** Not applicable — first review of this PR (it supersedes #2254, reviewed separately)
- **Tracked follow-ups:** 1 — [#2556](https://github.com/coleam00/Archon/issues/2556), pre-existing and confirmed out of scope by both reviewers
- **Validation:** `bun run validate` exit 0 at `ad950327`; 459 pass / 0 fail in `dag-executor.test.ts`; CI green at the prior head and re-running

## Findings

|---|---|---|---|
| `R1` | Important | `node_started` reported the `effort:` Codex ignores instead of the `modelReasoningEffort` it ran at — the #2395 failure mode, newly reachable from workflow YAML alone | FIXED |
| `R2` | Suggestion | The resolved-provider fact was spelled as two inverse string comparisons 84 lines apart, with nothing keeping them in sync | FIXED |
| `R3` | Suggestion | The Codex effort vocabulary exists in four independently maintained copies | TRACKED FOLLOW-UP |

## Detailed Findings

<details>
<summary><code>R1</code> — node_started reported an effort the node never ran at</summary>

**Impact:** For a Codex node with both `effort:` and `modelReasoningEffort:` declared, the persisted `workflow_events` `node_started` row and the live console emitter reported the `effort:` value — which Codex ignores entirely — instead of the `modelReasoningEffort` it actually ran at. Codex execution was correct; only the audit trail and console lied. This is precisely the gap issue #2395 was closed to fix.

**Evidence:** `nodeConfig.effort` is populated for every provider from `node.effort ?? workflowLevelOptions.effort` (`dag-executor.ts:1166`), and `capChecks` only *warns* that Codex ignores it — it never strips it. `resolvedEffort` was `nodeConfig.effort ?? assistantEffort`, so the ignored field won the `??` chain. Codex's `buildThreadOptions` (`providers/src/codex/provider.ts:79-93`) reads only `assistantConfig.modelReasoningEffort`.

Reproduced independently before accepting the finding, via a scratch test since reverted:

```
SCRATCH codex-runs-at: low
SCRATCH nodeConfig.effort(ignored by codex): max
SCRATCH node_started reports: max
```

from `provider: codex` + `effort: max` + `modelReasoningEffort: low`.

**In scope for this PR, not pre-existing:** before this change, workflow-level `modelReasoningEffort` was inert, so the collision required a `config.yaml` `assistants.codex.modelReasoningEffort` disagreeing with a workflow `effort:` — a far rarer coincidence, since a tier preset cannot produce it (`applyPresetOptions` only routes when `workflowLevelOptions.effort` is unset). Forwarding the field makes it reachable from workflow YAML alone. It is also the exact shape `guides/authoring-workflows.md:249` recommends: set `effort:` for Claude/Pi/Copilot nodes and `modelReasoningEffort:` for Codex in the same workflow. And the PR's own comment claimed the write was "positioned … BEFORE the resolvedEffort read below so telemetry reports the value that actually runs" — a claim the code did not deliver.

**Required outcome:** Branch the read on the resolved provider rather than on field priority, mirroring what `routePresetEffort` already does on the write side.

**Disposition:** FIXED in `ad950327`. Mutation-verified: reverting to `nodeConfig.effort ?? assistantEffort` fails the new test `reports modelReasoningEffort in node_started, not the effort Codex ignores` with `Expected "low", Received "max"`. No existing test broke (459 pass, up from 458).

</details>

<details>
<summary><code>R2</code> — the resolved-provider fact was duplicated as inverse booleans</summary>

**Impact:** `dag-executor.ts:1112` (`provider !== 'codex'` → warn) and `:1196` (`provider === 'codex'` → write) each encoded "Codex is the only consumer of these two `assistantConfig` keys" as an independent string comparison, 84 lines apart with no shared symbol. Extending one side without the other yields either a false "unsupported" warning on a field that is in fact applied, or a silent no-op with no warning — the exact failure this PR exists to remove.

**Evidence:** confirmed as the only two `'codex'` literals in the file. R1's fix would have added a third, which is what turned this from a style preference into a real correction: the same fact then drives whether to warn, whether to write, and which field holds the applied effort.

**Required outcome:** Name the fact once and use it at every site.

**Disposition:** FIXED in `ad950327` — hoisted as `readsAssistantConfigOptions`, used at all three sites. Mutation-verified: flipping the const's sense fails 8 tests. Note this is additive to R3 rather than an instance of it — R3 is the *value enum* copied four ways, this was the *provider gate* copied twice.

</details>

<details>
<summary><code>R3</code> — four independently maintained copies of the Codex effort vocabulary</summary>

**Impact:** `modelReasoningEffortSchema` (`schemas/workflow.ts:20`), `CODEX_REASONING_EFFORTS` (`model-validation.ts:231-237`), `validEfforts` (`codex/config.ts:21`), and the literal union on `CodexProviderDefaults` (`providers/src/types.ts:30`) all define the same five values, none deriving from another.

**Evidence:** `model-validation.ts` does not import `schemas/workflow.ts`. The `types.ts` comment claims the union "structurally matches" the workflows type — an unenforced claim, not a type link.

**Required outcome:** Derive within `@archon/workflows` from one source, per the `TRIGGER_RULES` / `WORKFLOW_HOOK_EVENTS` precedent in AGENTS.md. Deferred because issue #2556 step 2 rewrites `model-validation.ts` anyway.

**Disposition:** TRACKED FOLLOW-UP — [#2556](https://github.com/coleam00/Archon/issues/2556), filed before this review. The seam reviewer confirmed this PR opens no new drift path: the written value is pre-validated at load time (`loader.ts:750-763`) against a list that matches `parseCodexConfig`'s allowlist exactly.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | R1 |
| seams | R2, R3 |

**Checked and cleared, with evidence** — recorded because several were explicit risk hypotheses rather than absent concerns:

- **The `applyPresetOptions` guard omission is safe.** Its early return inspects `workflowLevelOptions.effort`, never `.modelReasoningEffort` — distinct fields, distinct YAML keys, populated independently. `code` traced all four precedence combinations and confirmed workflow-level > tier preset > `config.yaml` holds; `seams` confirmed no observable double-write, since `assistantConfig` is freshly constructed per node.
- **The Copilot leak is closed.** Copilot is a disjoint provider id with a disjoint `config.assistants` namespace; the gate excludes it, so `minimal` never reaches the parser that would silently drop it.
- **`assistantConfig` provenance is correct.** `{...config.assistants[provider]}` always spreads the *resolved* provider's entry, never the configured one.
- **Warning wording and per-node repetition are pre-existing.** `fallbackModel`, `sandbox` and `betas` already emit the same node-attributed template for workflow-level declarations, and already re-fire per `loop_group` body iteration. Not introduced or worsened here.
- **Composition boundaries need nothing.** `include-expander.ts:485` derives its dropped-field list generically from `Object.keys(child)`; `workflow:` sub-runs build `workflowLevelOptions` fresh from the child's own YAML; `loop_group` threads the single object built once. Identical to the five pre-existing workflow-level fields.
- **Resume needs nothing.** The workflow YAML is re-read from disk on every execution; no serialized `WorkflowDefinition` snapshot exists to keep in sync.
- **The `#2556` deferral is a legitimate scope boundary.** The change reuses the existing warn path rather than growing new YAML surface or a second effort-shaped capability axis.
- **The "previously inert" framing is accurate and broader than the issue states.** The schema declares both fields, the loader validates them, the web workflow builder serializes them into YAML (`packages/web/src/components/workflows/YamlCodeView.tsx:162-166`), and the CLI reports them (`packages/cli/src/commands/workflow.ts:806,828`). Every surface treated them as real except the one that reads them.

## Validation

|---|---|---|
| `bun run validate` (`ad950327`) | PASS | exit 0, all 9 steps, zero `N fail` lines with N>0 across all packages |
| `bun test src/dag-executor.test.ts` (`ad950327`) | PASS | 459 pass, 0 fail (458 before R1's fix; +1 new test, none broken) |
| Mutation — revert `resolvedEffort` to `nodeConfig.effort ?? assistantEffort` | FAILS as intended | `Expected "low", Received "max"` |
| Mutation — flip `readsAssistantConfigOptions` | FAILS as intended | 8 tests fail |
| Mutation — reintroduce #2254's `applyPresetOptions` guard | FAILS as intended | Claude preset-effort test fails |
| Mutation — move the write past the `resolvedEffort` read | FAILS as intended | `node_started` reports `medium` while the node runs `xhigh` |
| Mutation — remove the `provider === 'codex'` write gate | FAILS as intended | `modelReasoningEffort` set on a Claude node |
| Mutation — warn reads `configuredProvider` | FAILS as intended | pinned by `dd227c93`; survived all 457 tests before that commit |
| GitHub CI (`dd227c93`) | PASS | ubuntu 3m6s, windows 6m30s, docker-build 4m24s, schema-upgrade 27s, CodeRabbit 5/5 |
| GitHub CI (`ad950327`) | PASS | ubuntu 3m31s, windows 6m42s, docker-build 4m7s, schema-upgrade 30s, CodeRabbit |
| GitHub CI (`1c55db7c`, docs commit) | PASS | ubuntu 3m26s, windows 7m55s, docker-build 4m58s, schema-upgrade 28s, Build docs site 44s. CodeRabbit reported `pass` but rate-limited, so it did not re-review this commit |

## Verdict

**READY TO MERGE**

No open Critical or Important findings. R1 was the one genuine defect — a telemetry lie the PR's own comment denied — and it is fixed and mutation-verified rather than dispositioned away. R2 fell out of the same correction. R3 is pre-existing, tracked, and confirmed not made worse here.

**Post-review commit.** `1c55db7c` landed after this review: a docs-only change correcting a materially false line in `guides/authoring-workflows.md` (it claimed `webSearchMode: disabled` means "No web access"; `buildThreadOptions` hard-codes `networkAccessEnabled: true` on every Codex run, so the field gates Codex's built-in search tool only), plus notes that both fields are workflow-level with no per-node form and that a provider mismatch now warns. No production code touched; `bun run validate` and all six CI checks green at that head. It does not change this verdict.

**One behavior note for the merger:** R1's fix means a Codex node declaring both `effort:` and `modelReasoningEffort:` now reports the latter in `node_started` where it previously reported the former. That is the correction — Codex ignores the former — but it does change what existing runs emit.

Issue #2555, opened 2026-08-17
feat(workflows): chat adapters cannot supply declared workflow inputs


Wirasm:
#2554 gives a direct run a channel for declared `inputs:` — CLI `--input name=value`, an `inputs` map on `POST /api/workflows/:name/run`, and a console form. Chat adapters (Slack/Telegram/Discord/GitHub) deliberately got none of it: they keep the pre-cost refusal.

Chat has only a trigger message, so a workflow that declares a required input cannot be started from Slack/Telegram/Discord/GitHub at all today.

## Hard constraint: never parse natural language

**No regex, no grammar, no hand-rolled parser is ever pointed at a user's natural-language message to pull values out of it.** Not as a fast path, not as a fallback, not "just for the simple cases." A model resolves meaning; code does not.

This is not negotiable and it is not a detail of the design below — any proposal that reaches for a pattern match over user prose is rejected on sight.

Do not mistake the existing router regex for a precedent. `parseWorkflowInvocation` matches `/invoke-workflow <name>` — a **structured command the model was instructed to emit**. Parsing your own agent's structured output is fine. Parsing the human's sentence is not.

## Suggested direction — a suggestion, not a decision

This is how I think I'd want it to work. It is a starting point for whoever scopes this, not a settled design; if the implementation argues for something else, take the argument.

**Agent-first resolution from natural language.** The user writes an ordinary chat message. The agent has a *tool* that resolves the workflow's declared inputs from that request, and — crucially — the agent can come back and ask the user to fill anything the first message didn't cover. So it's a conversation, not a one-shot parse:

> **user:** review the auth changes, be strict about it
> **agent:** *(resolves `scope: auth`, `style: strict`; sees `base_branch` is required and unsupplied)*
> **agent:** which branch should I review against?
> **user:** dev
> **agent:** *(starts the run)*

The agent does the understanding. The tool's job is to receive already-resolved values and hand them to the existing contract — it is a channel, not an interpreter.

Two things this is deliberately **not**:

- Not a `--input k=v` grammar bolted onto `/workflow run` in chat. Making a Slack user type flags is the thing worth avoiding.
- Not one-shot extraction at routing time. The router picks the workflow from the message in a single pass; that seam has nowhere to put "I still need one more value from you." Gap-filling is the part that makes this usable, and it needs a turn-taking surface.

[...]

PR #2560, opened 2026-08-17
feat(workflows): a workflow that declares required inputs can now be run directly

### LINKED ISSUE #2554: feat(workflows): a workflow that declares required inputs cannot be run on its own
## Problem

A workflow that declares a **required** input cannot be run on its own. `assertWorkflowInputsSatisfiable` refuses the invocation at `packages/core/src/orchestrator/orchestrator-agent.ts:742-757`, and the comment states the model out loud:

> a workflow declaring `required` inputs is a reusable block — only a caller's `with:` satisfies them, so a bare top-level invocation fails here, before any worktree/clone/AI cost.

The docs name the behaviour too: *"Bare runs of a required-input block fail fast"* (`guides/authoring-workflows.md:1127`).

So the engine has **two classes of workflow**: workflows, and "reusable blocks" that can only be called. A workflow becomes uninvokable the moment it declares an input it actually needs — which is the moment it becomes worth composing.

That is backwards for the authoring model we want to endorse. A workflow is a workflow. `review-github-issue.yaml` should run on its own *and* compose into ten different parents. Today, declaring `scope: { required: true }` on it forces a choice between the two.

## Why this is separable from #1764

#1764 makes a composed workflow run **as authored** — its own provider, model, and options survive composition instead of being overwritten by the parent. That is the config half of "a workflow is a workflow".

This issue is the **invocation** half: the same workflow must also be startable directly. The two are independent — #1764's plan explicitly does not deliver this, and says so — but together they are what makes the endorsed decomposition (#2123: express something like `fix-github-issue-experimental` as ~10 composed workflows) actually usable. A workflow you cannot run alone is hard to develop, hard to test, and hard to debug.

## What is missing

Only a channel. The gate is **policy, not structure**:

- `defaultRunInputs` (`packages/workflows/src/workflow-inputs.ts:98-105`) already synthesizes declared defaults for a run with no caller, and `executor.ts:~1830` already merges them into the run's `metadata.inputs`.
- `resolveRunInputs` / `inputEnvVars` (`dag-executor.ts:149-160`) already deliver `metadata.inputs` to every surface — `$INPUTS.<name>` on AI/prompt bodies, `INPUTS_<UPPER_SNAKE>` for `bash:`/`script:`.
- `resolveDeclaredInputs` (`workflow-inputs.ts:51`) already validates a supplied map against the declaration, and is the single implementation both call surfaces go through.

Nothing supplies values at top-level invocation. That is the whole gap.

## Design questions to settle

1. **The channel.** A CLI flag (`archon workflow run <name> --input scope=... --input mode=...`) is the obvious one. Should the web/API run route take a map too? Chat platforms have only a trigger message.
2. **`$ARGUMENTS` mapping.** A workflow with exactly one required input could take the trigger message as that input, which would make chat invocation work with no new syntax. Convenient, or too magic? It is also the only option that works on Slack/Telegram/Discord/GitHub without a new command grammar.
3. **Typed values.** #2453 owns declared input *shapes*. Whatever channel lands here should not need redesigning when values stop being strings — the canonical representation rule (`typeof x === 'string' ? x : JSON.stringify(x)`) is the thing to hold to.
4. **What replaces the current failure.** Today an unsatisfied required input is a clean pre-cost refusal naming the missing names, which is good behaviour and should survive for the case where the channel genuinely supplies nothing.
5. **Discovery and the builder.** A required-input workflow already loads and lists deliberately. Should the run affordance in the console prompt for inputs rather than being disabled?

## Non-scope

- Declared input *value shapes* and validation — #2453.
- The composed-config fix — #1764.
- Any change to `workflow:` sub-run invocation, which already supplies inputs via `with:`.

## Acceptance

- [ ] A workflow declaring `required` inputs can be invoked directly with values supplied, and runs.
- [ ] The same workflow composed via `include:`/`workflow:` keeps working unchanged.
- [ ] Supplied values reach every surface a composed workflow's values reach today: `$INPUTS.<name>` on prompt/AI/command bodies and `INPUTS_<UPPER_SNAKE>` for `bash:`/`script:` nodes, including named script files.
- [ ] Values are validated through the existing `resolveDeclaredInputs` contract — missing required and undeclared keys fail the same way for a direct invocation as for a caller's `with:`.
- [ ] Validation happens before any worktree, clone, or AI cost, preserving today's pre-cost refusal.
- [ ] An unsatisfied required input with no value supplied still fails fast, naming the missing inputs.
- [ ] Declared defaults still apply for omitted non-required inputs on a direct run.
- [ ] The channel survives a cold resume — a resumed direct run sees the same inputs.
- [ ] Docs stop describing a required-input workflow as a "block" that cannot be run, and the "Bare runs of a required-input block fail fast" section is rewritten.

## Related

#1764 (composed workflows run as authored — the config half) · #2453 (declared input value shapes) · #2123 (the defaults overhaul this unblocks) · #2470 (delivered the `inputs:`/`with:`/`$INPUTS` baseline)

Wirasm:
## Summary

- **Problem:** Declaring an input a workflow actually needs made it uninvokable. `assertWorkflowInputsSatisfiable` refused *every* top-level run of a workflow with a `required` input, calling it "a reusable block" only `include:`/`workflow:` could call.
- **Why it matters:** A workflow became callable-only at exactly the moment it became worth composing. `review-github-issue.yaml` had to choose between running on its own and composing into ten parents — and a workflow you cannot run alone is hard to develop, test, or debug. This is the invocation half of "a workflow is a workflow" (#1764 is the config half), and it unblocks the #2123 decomposition.
- **What changed:** Values can now be supplied at invocation — `--input name=value` on the CLI (repeatable), an `inputs` map on `POST /api/workflows/:name/run`, and a run form in the console. All three go through the **same** `resolveDeclaredInputs` contract a caller's `with:` map goes through, so rejections match, defaults still apply, and the refusal still lands before any worktree, clone, or AI cost. Supplied values persist to `metadata.inputs`, so a cold resume replays the same invocation.
- **What did NOT change (scope boundary):** No new YAML surface — `inputs:` already exists (#2470); this is an invocation channel for it, so the workflow-language constitution is untouched. `include:`/`workflow:` `with:` is byte-for-byte unchanged. Chat adapters deliberately get **no** channel (they carry only a trigger message) and keep the reworded pre-cost refusal — tracked in #2555. Typed/non-string input values remain #2453's; the canonical-representation rule is held so #2453 can widen this channel without redesigning it.

### Before

```
  Author                 Archon
  ──────                 ──────
  writes review-block.yaml
    inputs: {diff: {required: true}}
  archon workflow run review-block ────▶  assertWorkflowInputsSatisfiable
                                          ✗ REFUSED, always
  sees "It is a reusable block:      ◀──  "reference it from another
  reference it from another               workflow with an include:
  workflow…"                               or workflow: node"

  (the only way to run it is to write a SECOND workflow that calls it)
```

### After

```
  Author                 Archon
  ──────                 ──────
  archon workflow run review-block \
    --input diff="$(git diff)"     ────▶  *resolveTopLevelInputs*
    --input style=terse                   → resolveDeclaredInputs (shared contract)
                                          ✓ validated pre-cost
                                          *metadata.inputs stamped on the run row*
                                          $INPUTS.diff / $INPUTS_DIFF reach nodes
  run completes                     ◀──

  console: picks review-block       ────▶  *run form renders a field per input*
           fills "diff" *            ◀──   Start disabled until required filled
           Start run                ────▶  POST …/run { inputs: {...} }

  archon workflow run review-block  ────▶  ✗ still refused, pre-cost —
    (nothing supplied)              ◀──   *now naming --input and the console*

  Slack: "run review-block"         ────▶  ✗ refused (no channel yet, #2555)
```

### Before

```
  include: with: ──┐
                   ├──▶ resolveDeclaredInputs ──▶ (load-time splice / child metadata.inputs)
  workflow: with: ─┘

  CLI  ──┐
  HTTP ──┼──▶ assertWorkflowInputsSatisfiable(workflow)  ──✗ any required input
  chat ──┘         (no parameter for supplied values)

  executor: createWorkflowRun{metadata} ──▶ defaultRunInputs merge ──▶ dag-executor
                                                                        ├─▶ $INPUTS.<name>  (AI/prompt)
                                                                        └─▶ INPUTS_<SNAKE>  (bash/script)
```

### After

```
  include: with: ──┐
                   ├──▶ resolveDeclaredInputs ──▶ (load-time splice / child metadata.inputs)
  workflow: with: ─┘         ▲
                             ‖ (same implementation — parity is structural)
  CLI  --input k=v ──▶ [+] parseInputAssignments ══╗
  HTTP { inputs } ═══▶ [~] HandleMessageContext    ╠══▶ [+] resolveTopLevelInputs
      (via [~] WorkflowDispatchOptions.inputs      ║          ✗ pre-cost refusal
       and [~] WorkflowRoutingContext.inputs)      ║
  chat  ─────────────────(supplies nothing)════════╝

  [~] executor: createWorkflowRun{metadata + inputs} ──▶ defaultRunInputs merge (unchanged)
  [~] orchestrator: pre-created row{metadata + inputs}         │
                                                                ▼
                                                          dag-executor (unchanged)
                                                            ├─▶ $INPUTS.<name>
                                                            └─▶ INPUTS_<SNAKE>

  [+] web console: DraftRunCard form ──▶ [+] lib/workflow-inputs ──▶ [~] startRun
```

**Connection inventory**

| `cli/cli.ts` (parseArgs) | `commands/workflow.ts` | **modified** | `input: {type:'string', multiple:true}` → `WorkflowRunOptions.inputs` (first repeatable flag in the tree) |
| `commands/workflow.ts` | `workflows/workflow-inputs` | **new** | `parseInputAssignments` — the `name=value` grammar |
| `commands/workflow.ts` | `utils/workflow-requirements` | **modified** | `assertWorkflowInputsSatisfiable` → `resolveTopLevelInputs(workflow, supplied)` |
| `utils/workflow-requirements` | `workflow-inputs` | **new** | gate now delegates to the shared `resolveDeclaredInputs` |
| `server/routes/api.ts` | `core/types` (`HandleMessageContext`) | **modified** | `workflowInputs` — a side channel, never the message text |
| `orchestrator-agent` | `orchestrator` (`WorkflowRoutingContext`) | **modified** | `inputs` threaded to the console's background-dispatch path |
| `orchestrator-agent` / `orchestrator` | `workflows/executor` | **modified** | `ExecuteWorkflowOptions.inputs` |
| `executor.ts` | `workflow_runs.metadata` | **modified** | stamps `SUBRUN_METADATA_KEYS.inputs` on a fresh row only |
| `web/DraftRunCard` | `web/console/lib/workflow-inputs` | **new** | the form's two pure decisions, unit-tested |
| `web/startRun` | `POST /api/workflows/:name/run` | **modified** | `inputs` in JSON; JSON-encoded `inputs` form field in multipart |
| `dag-executor` → node surfaces | — | unchanged | delivery already worked for a top-level row; no engine change |
| `include-expander` / `workflow:` child spawn | — | unchanged | composition untouched |

## Label Snapshot

- Risk: `risk: low`
- Size: `size: L`
- Scope: `workflows`
- Module: `workflows:workflow-inputs`, `cli:workflow`, `server:api`, `core:orchestrator`, `web:console`

## Change Metadata

- Change type: `feature`
- Primary scope: `multi` (workflows, cli, server, core, web, docs)

## Linked Issue

- Closes #2554
- Related #2555 (chat adapters: natural-language → inputs; filed during planning, deliberately unscoped)
- Related #2453 (declared input value shapes), #1764 (composed workflows run as authored), #2123 (the defaults overhaul this unblocks), #2470 (delivered the `inputs:`/`with:`/`$INPUTS` baseline)

```bash
bun run validate     # EXIT=0 — all 10 steps green
```

Focused suites:

- `bun test src/workflow-inputs.test.ts` (workflows) — 13 pass / 0 fail (new file)
- `bun test src/utils/workflow-requirements.test.ts` — 11 pass / 0 fail
- `bun test src/subrun.test.ts` — 64 pass / 0 fail (5 new direct-run cases)
- `bun test src/commands/workflow.test.ts` (cli) — 221 pass / 0 fail (6 new)
- `bun test src/routes/api.workflow-runs.test.ts` (server) — 104 pass / 0 fail (5 new)
- `bun test src/orchestrator/orchestrator-agent.test.ts` (core) — 220 pass / 0 fail (5 new)
- `bun test src/lib/ src/experiments/console/` (web) — 548 pass / 0 fail across 47 files

**Falsification of the core change:** removing the `SUBRUN_METADATA_KEYS.inputs` spread from `executor.ts` turns 4 of the 5 new subrun tests red (60 pass / 4 fail), and the bare-run test correctly stays green. Restored → 64 pass. The tests fail for the intended reason.

## Security Impact (required)

Two notes worth a reviewer's eye, both preserving existing invariants:

- **`$INPUTS` still never reaches shell source.** Text substitution stays in the non-`shellSafe` branch; `bash:`/`script:` nodes read `INPUTS_<UPPER_SNAKE>` env vars, exactly as #2115 hardened. A direct run changes nothing here.
- **Input names are validated** against the shared `INPUT_NAME_PATTERN` (one source, `schemas/dag-node.ts`), and the HTTP route refuses any `inputs` payload that is not a flat object of strings — matching `readSubrunMetadata`'s guard, so nothing unreadable can be persisted. Supplied **values** are never logged (key names only).

## Compatibility / Migration

- Backward compatible? **Yes** — purely additive. A bare run of a defaulted workflow persists nothing and behaves byte-for-byte as before (regression-tested). A previously-unrunnable workflow becomes runnable.
- Config/env changes? **No**
- Database migration needed? **No** — `metadata.inputs` is an existing key with an existing reader; only the set of runs that carry it widens. An older Archon build ignores it on a top-level run exactly as it does today.

Verified scenarios (live runs against a real fixture repo, not only unit tests):

- **CLI, bare run of a required-input workflow** → refused with `requires input 'scope', and none was supplied … --input scope=<value> … web console … (#2555). No worktree was created and no AI cost was incurred.`
- **CLI, `--input scope=auth --input style=terse`** → completed; the `script:` node printed `script scope=auth style=terse`; row metadata `{"inputs":{"scope":"auth","style":"terse"}}`.
- **CLI, `--input scope=billing`** (style omitted) → node received `style=strict` from the declared default, while the row recorded only `{"inputs":{"scope":"billing"}}` — the default is *not* frozen onto the row.
- **HTTP, `inputs: {scope: 5}`** → `400 {"error":"inputs value for 'scope' must be a string"}`.
- **HTTP, valid map** (console background path) → `200`, run `completed`, metadata `{"inputs":{"scope":"http-path","style":"loose"}}`. This is the path that would have silently done nothing had `dispatchBackgroundWorkflow` not been threaded.
- **HTTP, nothing supplied** on a required-input workflow → run count unchanged (1 → 1): refused before a run row existed.

Edge cases checked: `=` inside a value (`msg=a=b` → `a=b`); empty value (`name=`) as a deliberate empty string; duplicate `--input` rejected rather than last-wins; invalid name rejected; `--input` + `--resume` rejected; repeated `--input` survives the `--detach` argv rebuild (pinned by test); cold resume of a top-level run replays the values (unit-tested against the persisted row alone, no invocation channel in the loop).

What was not verified: the console form was not click-tested in a browser — `@archon/web` has no React DOM test harness, so the card's two real decisions were extracted into `console/lib/workflow-inputs.ts` and unit-tested instead, and the HTTP path underneath the form was verified live end-to-end.

- **Affected subsystems:** the workflow invocation gate (CLI + orchestrator), the run route, run-row creation in both `executor.ts` and `dispatchBackgroundWorkflow`, and the console draft-run card. The DAG executor and both composition paths are untouched.
- **Potential unintended effects:** the gate's signature change touches every top-level dispatch, so the riskiest failure mode would be a dispatch branch that forgets to thread `inputs`. Two branches create a fresh row (`executeWorkflow` and the orchestrator's pre-created row) and both are covered; the true-resume branch deliberately does not thread it.
- **Guardrails:** the refusal remains pre-cost on every surface, so a mis-threaded value fails loudly at the gate or at the `$INPUTS.<name>` reference site (`executor-shared.ts` throws with a "Did you mean / Available inputs" hint) rather than running with a wrong value. `workflow.required_inputs_unsatisfiable` logs the missing and supplied **key names**.

## Rollback Plan (required)

- Fast rollback: `git revert a37fcd21` — one commit, no schema change, no data an older build would misread.
- Feature flags: none (a channel with nothing supplied is byte-for-byte prior behavior).
- Observable failure symptoms: a required-input workflow refusing despite `--input` being passed; a run starting with empty `$INPUTS_*`; a console Start button stuck disabled.

## Risks and Mitigations

- **Risk:** the gate now runs on paths that previously could not exist, and a resume supplies nothing — naively re-gating would make every resume of a required-input run impossible.
  - **Mitigation:** the gate is skipped on any resume path (CLI and orchestrator), documented inline; the row's inputs were validated at creation. CLI `--input` + `--resume` is rejected outright rather than silently ignored. Covered by an orchestrator test.
- **Risk:** persisting the *resolved* map (supplied + defaults), mirroring the `workflow:` child path, would start freezing declared defaults onto rows for every existing defaulted workflow — a silent behavior change and a stale snapshot across resume.
  - **Mitigation:** only caller-supplied values are persisted; defaults stay derived at execution. Regression-tested both ways (a bare run writes no `inputs` key). The asymmetry with the child path is deliberate and documented at the call site — effective `$INPUTS` is identical either way.
- **Risk:** `packages/web/src/lib/api.generated.d.ts` was stale (missing `inputs`/`returns` since #2523), so a naive regeneration could have carried unrelated drift.
  - **Mitigation:** regenerated in full from this branch's own server and formatted; the real diff is 9 insertions / 1 deletion — only the two fields and the route description. No unrelated drift.

---

Plan: https://github.com/coleam00/Archon/issues/2554#issuecomment-5314152150

## Summary by CodeRabbit

* **New Features**
  * Workflows now accept declared inputs through the CLI, web console, or API.
  * CLI users can provide multiple `--input name=value` options.
  * The web console displays input fields, descriptions, defaults, and required indicators.
  * Supplied inputs are preserved across workflow runs and resumes.

* **Bug Fixes**
  * Improved validation and error messages for missing, duplicate, malformed, or undeclared inputs.
  * Prevented unsupported inputs during dry runs and resumed executions.

* **Documentation**
  * Added guidance for configuring and supplying workflow inputs across supported run methods.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -158,6 +158,7 @@ Options:
   --base <branch>            Per-dispatch base override for epic slices (worktree cut-from + PR target)
   --no-worktree              Run on branch directly without worktree isolation
   --folder                   Register the current non-git directory as a folder project and run in place
+  --input <name>=<value>     Supply a declared workflow input; repeat per input (mutually exclusive with --resume)
   --resume                   Resume the most recent failed or paused run of the workflow (mutually exclusive with --branch)
   --dry-run                  Simulate workflow DAG control flow without creating a run or contacting a provider
   --stubs <path>             YAML node-output map for --dry-run
@@ -319,6 +320,8 @@ async function main(): Promise<number> {
         stubs: { type: 'string' },
         'exec-code': { type: 'boolean' },
         'pause-at-gates': { type: 'boolean' },
+        // Repeatable: `--input a=1 --input b=2` yields ['a=1', 'b=2'] (#2554).
+        input: { type: 'string', multiple: true },
       },
       allowPositionals: true,
       strict: false, // Allow unknown flags to pass through
@@ -604,6 +607,8 @@ async function main(): Promise<number> {
               stubsPath,
               execCode: execCodeFlag,
               pauseAtGates: pauseAtGatesFlag,
+              // Raw `name=value` assignments; parsed at the invocation gate (#2554).
+              inputs: values.input as string[] | undefined,
             };
             await workflowRunCommand(effectiveCwd, workflowName, userMessage, options);
             break;
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -853,6 +853,27 @@ describe('workflowRunCommand — --input declared inputs (#2554)', () => {
     expect(executeWorkflow).not.toHaveBeenCalled();
   });
 
+  it('rejects --input with --dry-run instead of simulating with empty inputs', async () => {
+    // The dry-run engine has no $INPUTS support: a bash node reading $INPUTS_SCOPE would
+    // expand an unset var to '', producing a trace that reads as a valid simulation of a
+    // run that never happened. Every other flag dry-run cannot honor is rejected the same
+    // way, so this closes the one hole in that list.
+    const { discoverWorkflowsWithConfig } = await import('@archon/workflows/workflow-discovery');
+    const dryRun = await import('@archon/workflows/dry-run');
+    await stubInputWorkflow();
+    (discoverWorkflowsWithConfig as ReturnType<typeof mock>).mockClear();
+    (dryRun.dryRunWorkflow as ReturnType<typeof mock>).mockClear();
+
+    await expect(
+      workflowRunCommand('/repo/root', 'review-block', 'go', {
+        dryRun: true,
+        inputs: ['diff=D1'],
+      })
+    ).rejects.toThrow(/--dry-run cannot be combined with --input/);
+
+    expect(dryRun.dryRunWorkflow).not.toHaveBeenCalled();
+  });
+
   it('does not re-gate a --resume of a required-input workflow', async () => {
     // The gate is deliberately skipped on resume: the run row already carries inputs
     // validated at creation, and a resume supplies nothing. A refactor that hoists
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -763,6 +763,113 @@ describe('workflowRunCommand — requires: [github] gate', () => {
   });
 });
 
+describe('workflowRunCommand — --input declared inputs (#2554)', () => {
+  let consoleSpy: ReturnType<typeof spyOn>;
+
+  beforeEach(async () => {
+    consoleSpy = spyOn(console, 'log').mockImplementation(() => {});
+    const { executeWorkflow } = await import('@archon/workflows/executor');
+    (executeWorkflow as ReturnType<typeof mock>).mockClear();
+  });
+
+  afterEach(() => {
+    consoleSpy.mockRestore();
+  });
+
+  /** Stub discovery with one workflow declaring a required + a defaulted input. */
+  async function stubInputWorkflow(): Promise<void> {
+    const { discoverWorkflowsWithConfig } = await import('@archon/workflows/workflow-discovery');
+    (discoverWorkflowsWithConfig as ReturnType<typeof mock>).mockResolvedValueOnce({
+      workflows: [
+        makeTestWorkflowWithSource(
+          {
+            name: 'review-block',
+            inputs: { diff: { required: true }, style: { default: 'strict' } },
+          },
+          'project'
+        ),
+      ],
+      errors: [],
+    });
+  }
+
+  it('runs a required-input workflow when --input supplies the value', async () => {
+    const { executeWorkflow } = await import('@archon/workflows/executor');
+    await stubInputWorkflow();
+    (executeWorkflow as ReturnType<typeof mock>).mockResolvedValueOnce({
+      success: true,
+      workflowRunId: 'run-ok',
+    });
+
+    await workflowRunCommand('/repo/root', 'review-block', 'go', {
+      noWorktree: true,
+      inputs: ['diff=D1'],
+    });
+
+    expect(executeWorkflow).toHaveBeenCalledTimes(1);
+    // Only the supplied value is threaded through; `style` stays a derived default.
+    const opts = (executeWorkflow as ReturnType<typeof mock>).mock.calls[0][7] as {
+      inputs?: Record<string, string>;
+    };
+    expect(opts.inputs).toEqual({ diff: 'D1' });
+  });
+
+  it('still refuses a required-input workflow when nothing is supplied, before any cost', async () => {
+    const { executeWorkflow } = await import('@archon/workflows/executor');
+    await stubInputWorkflow();
+
+    await expect(
+      workflowRunCommand('/repo/root', 'review-block', 'go', { noWorktree: true })
+    ).rejects.toThrow(/requires input 'diff'/);
+
+    expect(executeWorkflow).not.toHaveBeenCalled();
+  });
+
+  it('rejects an undeclared --input name before any cost', async () => {
+    const { executeWorkflow } = await import('@archon/workflows/executor');
+    await stubInputWorkflow();
+
+    await expect(
+      workflowRunCommand('/repo/root', 'review-block', 'go', {
+        noWorktree: true,
+        inputs: ['diff=D1', 'stlye=terse'],
+      })
+    ).rejects.toThrow(/does not declare input 'stlye'/);
+
+    expect(executeWorkflow).not.toHaveBeenCalled();
+  });
+
+  it('rejects a malformed --input assignment before any cost', async () => {
+    const { executeWorkflow } = await import('@archon/workflows/executor');
+    await stubInputWorkflow();
+
+    await expect(
+      workflowRunCommand('/repo/root', 'review-block', 'go', {
+        noWorktree: true,
+        inputs: ['diff'],
+      })
+    ).rejects.toThrow(/expected 'name=value'/);
+
+    expect(executeWorkflow).not.toHaveBeenCalled();
+  });
+
+  it('rejects --input combined with --resume rather than silently ignoring it', async () => {
+    const { executeWorkflow } = await import('@archon/workflows/executor');
+    // Flag validation runs after name resolution (as every other flag check does),
+    // so discovery still has to resolve for the conflict to be reached.
+    await stubInputWorkflow();
+
+    await expect(
+      workflowRunCommand('/repo/root', 'review-block', 'go', {
+        resume: true,
+        inputs: ['diff=D1'],
+      })
+    ).rejects.toThrow(/--resume and --input are mutually exclusive/);
+
+    expect(executeWorkflow).not.toHaveBeenCalled();
+  });
+});
+
 describe('workflowRunCommand', () => {
   let consoleSpy: ReturnType<typeof spyOn>;
 
@@ -4362,6 +4469,36 @@ describe('buildDetachedRunCmd', () => {
   // fixture modelled a compiled argv with no argv[1] at all, which is why
   // #2248 (detached child dies with `Unknown command: B:/~BUN/root/...`)
   // shipped green.
+  // `--input` is the first repeatable flag in the tree (#2554). Nothing here handles
+  // repetition specially — argv is forwarded verbatim — so this pins the property a
+  // future change to the argv rebuild could silently break, turning a detached run into
+  // one that starts with its inputs missing instead of failing.
+  it('forwards every repeated --input to the detached child', () => {
+    const cmd = buildDetachedRunCmd(
+      false,
+      '/path/to/bun',
+      [
+        '/path/to/bun',
+        '/abs/cli.ts',
+        'workflow',
+        'run',
+        'review-block',
+        '--input',
+        'diff=D1',
+        '--input',
+        'style=terse',
+        '--detach',
+      ],
+      '/abs/cwd',
+      []
+    );
+
+    expect(cmd.filter(arg => arg === '--input')).toHaveLength(2);
+    expect(cmd).toContain('diff=D1');
+    expect(cmd).toContain('style=terse');
+    expect(cmd).not.toContain('--detach');
+  });
+
   it('binary mode: uses [execPath] only (no duplicated entry arg), drops the Bun SFE virtual argv[1]', () => {
     const cmd = buildDetachedRunCmd(
       true,
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -853,6 +853,54 @@ describe('workflowRunCommand — --input declared inputs (#2554)', () => {
     expect(executeWorkflow).not.toHaveBeenCalled();
   });
 
+  it('does not re-gate a --resume of a required-input workflow', async () => {
+    // The gate is deliberately skipped on resume: the run row already carries inputs
+    // validated at creation, and a resume supplies nothing. A refactor that hoists
+    // `resolveTopLevelInputs` out of the `if (!options.resume)` guard would make every
+    // CLI resume of a required-input workflow throw instead — this is what catches it.
+    // (The `--input` + `--resume` test below cannot: it fails on the mutual-exclusion
+    // check before ever reaching the gate.)
+    const { executeWorkflow, hydrateResumableRun } = await import('@archon/workflows/executor');
+    const conversationDb = await import('@archon/core/db/conversations');
+    const codebaseDb = await import('@archon/core/db/codebases');
+    const workflowDb = await import('@archon/core/db/workflows');
+    await stubInputWorkflow();
+
+    (hydrateResumableRun as ReturnType<typeof mock>).mockResolvedValueOnce({
+      preCreatedRun: { id: 'run-prior', workflow_name: 'review-block' },
+      priorCompletedNodes: new Map([['node-a', 'done']]),
+    });
+    (conversationDb.getOrCreateConversation as ReturnType<typeof mock>).mockResolvedValueOnce({
+      id: 'conv-resume',
+    });
+    (codebaseDb.findCodebaseByDefaultCwd as ReturnType<typeof mock>).mockResolvedValueOnce({
+      id: 'cb-resume',
+      default_cwd: '/repo/root',
+      default_branch: 'develop',
+    });
+    // working_path null keeps the resume on the caller cwd, skipping the existsSync probe.
+    (workflowDb.findResumableRun as ReturnType<typeof mock>).mockResolvedValueOnce({
+      id: 'run-prior',
+      working_path: null,
+      workflow_name: 'review-block',
+    });
+    (conversationDb.updateConversation as ReturnType<typeof mock>).mockResolvedValueOnce(undefined);
+    (executeWorkflow as ReturnType<typeof mock>).mockResolvedValueOnce({
+      success: true,
+      workflowRunId: 'run-prior',
+    });
+
+    // No --input, and the workflow declares a required one: this must NOT throw.
+    await workflowRunCommand('/repo/root', 'review-block', 'go', { resume: true });
+
+    expect(executeWorkflow).toHaveBeenCalledTimes(1);
+    // The resume branch carries the row's own inputs; it never re-stamps from a flag.
+    const opts = (executeWorkflow as ReturnType<typeof mock>).mock.calls[0][7] as {
+      inputs?: Record<string, string>;
+    };
+    expect(opts.inputs).toBeUndefined();
+  });
+
   it('rejects --input combined with --resume rather than silently ignoring it', async () => {
     const { executeWorkflow } = await import('@archon/workflows/executor');
     // Flag validation runs after name resolution (as every other flag check does),
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -961,6 +961,12 @@ export async function workflowRunCommand(
       ['--container', options.container === true],
       ['--resume', options.resume === true],
       ['--detach', options.detach === true],
+      // The dry-run engine has no `$INPUTS` support at all — it never substitutes
+      // `$INPUTS.<name>` and never sets `INPUTS_<UPPER_SNAKE>` in a code node's env.
+      // Accepting the flag would silently simulate with EMPTY inputs: a bash node
+      // reading `$INPUTS_SCOPE` expands an unset var to '', so the trace reads as a
+      // valid simulation of a run that never happened. Reject rather than mislead.
+      ['--input', options.inputs !== undefined && options.inputs.length > 0],
     ] as const;
     const incompatibleFlag = incompatible.find(([, present]) => present)?.[0];
     if (incompatibleFlag) {
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -45,8 +45,9 @@ import { resolveWorkflowName } from '@archon/workflows/router';
 import { executeWorkflow, hydrateResumableRun } from '@archon/workflows/executor';
 import {
   assertWorkflowRequirementsMet,
-  assertWorkflowInputsSatisfiable,
+  resolveTopLevelInputs,
 } from '@archon/workflows/utils/workflow-requirements';
+import { parseInputAssignments } from '@archon/workflows/workflow-inputs';
 import { dryRunWorkflow, formatDryRunTrace, loadDryRunStubs } from '@archon/workflows/dry-run';
 import {
   getWorkflowEventEmitter,
@@ -222,6 +223,14 @@ export interface WorkflowRunOptions {
   execCode?: boolean;
   /** Stop at the first approval gate instead of auto-approving it. */
   pauseAtGates?: boolean;
+  /**
+   * Raw `--input name=value` assignments (#2554), one per occurrence of the flag.
+   * Parsed and validated against the workflow's declared `inputs:` at the invocation
+   * gate — before the `--detach` fork and before any worktree, clone, or AI cost.
+   * Kept as raw strings here so the grammar has exactly one parser
+   * (`parseInputAssignments` in `@archon/workflows`).
+   */
+  inputs?: string[];
 }
 
 /**
@@ -1015,6 +1024,13 @@ export async function workflowRunCommand(
         '  Remove --branch when using --resume.'
     );
   }
+  if (options.resume && options.inputs !== undefined && options.inputs.length > 0) {
+    throw new Error(
+      '--resume and --input are mutually exclusive.\n' +
+        "  A resume replays the original invocation's inputs, recorded on the run.\n" +
+        '  Drop --input to resume, or start a fresh run to supply different values.'
+    );
+  }
 
   // Per-dispatch --base override, normalized once. Wins over repo config + the
   // codebase default for both the worktree cut-from (the provider request's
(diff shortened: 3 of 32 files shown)


Wirasm:
> **Updated for head `3684ee07`.** Originally published against `5acd0699`; both findings were taken and are now confirmed FIXED. Verdict unchanged.

# PR Review: #2560 — feat(workflows): a workflow that declares required inputs can now be run directly

## Signal

This is a verification pass at `5acd0699`, not a fresh sweep. Seven specialists reviewed earlier heads across three rounds and produced five correction commits; the aggregator died on a transient API error before publishing. Three things were re-checked at the current head: the deferral design in `dispatchOrchestratorWorkflow` (adversarially, because it has been wrong twice), whether the three newest tests genuinely fail without their code, and whether the corrections broke anything the earlier rounds could not have seen.

**The deferral design holds.** I enumerated every exit reachable after the gate defers and checked both invariants the caller named. No path creates a fresh run row with unvalidated inputs — the fallthrough at `orchestrator-agent.ts:1036` is the only fresh-row exit reachable after a deferral, and it is guarded by the re-raise at `:1028`; the other two fresh-row exits (`:1062` background, `:1081` foreground) sit under `else` branches of `if (resumableRun?.working_path)` and are therefore *structurally* unreachable when a deferral happened, since `willContinueExistingRun` requires that same truthy `working_path`. The resume exit at `:1005` passes `...prepared`, whose `preCreatedRun` makes the executor's `if (!workflowRun)` stamp unreachable. The `:930` refuse-menu exit is now provably unreachable with a held error: `willContinueExistingRun` is `W && (P || R)` and the menu fires on `W && !(P || R)`, so the mirror `966908a4` introduced is exact, not approximate.

**Three exits discard the held error, and all three are correct to.** The `requires: [github]` refusal (`:834`), the `IsolationBlockedError` return (`:901`), and a successful resume (`:1005`). The first two abandon the dispatch entirely — no fresh row, no cost — and on a deferral the held violation is *moot* by construction, because the invocation was only ever going to continue a run whose row already holds validated inputs. Surfacing it there would re-commit the `396f786f` bug in miniature: telling a user their paused run is missing an input it already has. The third is the intended behavior, and it is not silent — `:991-1004` names the ignored keys whenever values were actually supplied. This is worth stating explicitly so the next reader sees these were considered rather than missed.

**The three newest tests are mutation-verified.** I cloned the repo to scratch rather than touching the shared worktree, reverted each guard independently, and confirmed a 1:1 mapping — each revert failed exactly one test, its own, with no collateral. Baseline 225 pass / 0 fail.

Two non-blocking findings were raised at `5acd0699`, both introduced by the correction commits and both about *which* diagnostic reaches the user, not about what executes. **Both are FIXED in `3684ee07`**, independently mutation-verified — see *Delta Confirmation*.

- **Blocking:** 0
- **Non-blocking:** 0 (`R1` and `R2` both FIXED in `3684ee07`)
- **Resolved before this review:** 5 correction commits dispositioning the earlier rounds' findings — see the implementation report's "Review Corrections" sections
- **Tracked follow-ups:** [#2555](https://github.com/coleam00/Archon/issues/2555) (chat input channel), [#2562](https://github.com/coleam00/Archon/issues/2562) (isolation ordering), [#2453](https://github.com/coleam00/Archon/issues/2453) (typed input values), plus the `WorkflowMissingInputsError` collapse (unfiled)
- **Validation:** CI fully green at `3684ee07` — `test (ubuntu-latest)`, `test (windows-latest)`, `docker-build`, `schema-upgrade`, `Build docs site`, CodeRabbit. The ubuntu job is the authoritative gate: five generated-file checks, `type-check`, `lint --max-warnings 0`, `format:check`, `test:install`, full suite. Independently, 226 pass / 0 fail in `orchestrator-agent.test.ts` in an isolated clone at this head (225/0 at `5acd0699`).

## Findings

|---|---|---|---|
| `R1` | Suggestion | The lost-resume-race append tells a caller who supplied nothing that they must supply inputs — on chat, the only surface that reaches it, the appended text is never useful and always misleading | FIXED in `3684ee07` |
| `R2` | Suggestion | `willContinueExistingRun` mirrors one of the two refusal exits below it; the comment claims it mirrors "the resume branch's own condition". Practically unreachable, but the comment is the durable trap | FIXED in `3684ee07` |

## Detailed Findings

<details>
<summary><code>R1</code> — the lost-race append fires hardest on the surface where it cannot be acted on</summary>

**Impact:** `5acd0699` appends the held contract violation to the lost-resume-race message (`orchestrator-agent.ts:974-982`). For the *undeclared-key* deferral this is right — the caller supplied a bad key, this is the only place it gets told, and there is no sibling message on that path. For the *missing-required* deferral with nothing supplied, the appended text is wrong in substance and in framing.

Reachability makes it worse, not better. A deferral requires `willContinueExistingRun`, i.e. a paused run with a working path. On chat, `options.inputs` is always `undefined` (chat has no channel — that is the deliberate #2555 boundary), so **every** chat re-invocation of a paused required-input workflow defers, and **none** of them involve supplied values. If such a dispatch loses the race, the user gets:

> ⚠️ **foo** is already being resumed (status: running). No action taken — follow the existing run for progress.
>
> Also note: This workflow requires input `'diff'`, and none was supplied. Supply it with `--input diff=<value>` on the CLI … Chat platforms cannot supply inputs yet (#2555). **No worktree was created and no AI cost was incurred.**

Nothing about that invocation needed inputs — it was a continuation, and the run being resumed elsewhere carries its own validated values. The message tells a chat user to use a flag they do not have, on a surface it then says cannot supply inputs at all, and closes with a hard-refusal reassurance (`WorkflowMissingInputsError`'s own tail) bolted onto a message whose first line is "No action taken — follow the existing run for progress." A chat user could reasonably conclude the workflow is unusable from chat.

**Evidence:** `WorkflowMissingInputsError`'s message is fixed at `workflow-requirements.ts:83-90` and ends in the refusal reassurance. The append at `:981` is unconditional on whether anything was supplied, unlike its sibling at `:991`, which is already correctly conditioned on `options?.inputs && Object.keys(options.inputs).length > 0`.

**Required outcome:** Append only when the caller actually supplied values that got dropped — the same condition the `:991` message already uses. That preserves every case where the append carries information (undeclared key always has non-empty `inputs`; missing-required with partial values still warns) and drops the only case where it carries none. The `5acd0699` test pins the wrong half, so it would need its fixture changed to supply a key rather than nothing — which also makes it test the case the append exists for.

**Why it is not blocking:** it requires losing a concurrency race, the run continues correctly either way, and the harm is a confusing sentence rather than a wrong execution.

**Found by:** verification pass, `errors` scope

**Disposition:** FIXED in `3684ee07`. The append is now guarded on `options?.inputs` being non-empty, the exact condition the `:991` sibling already used. The test was replaced by a two-sided pair sharing a `dispatchLosingResumeRace` helper. Mutation-verified in both directions — see *Delta Confirmation*.

</details>

<details>
<summary><code>R2</code> — the mirror covers one of two refusal exits, and the comment says otherwise</summary>

**Impact:** `966908a4`'s comment at `orchestrator-agent.ts:774-780` states that `willContinueExistingRun` "deliberately mirrors the resume branch's own condition below." There are **two** refusal exits below it, not one. It mirrors `:930` (the resume/abandon/force menu) exactly. It does not account for `:914` — `options?.resumeRun && !options.resumeRun.working_path` → "Cannot resume `<id>`: missing working path."

Because `willContinueExistingRun` requires a truthy `working_path`, an explicit resume of a run row lacking one yields `willContinueExistingRun === false`, so the gate refuses at `:812` with the input error and returns *before* `:914` ever runs. The specific "missing working path" diagnostic is shadowed by a message telling the user to pass `--input`, when the actual problem is a broken run row. Note this is a behavior change from `396f786f`: the pre-hoist code skipped the gate for *any* explicit resume (`isResumeDispatch`), so `:914` was always reachable.

**Reachability — honest assessment: effectively nil today.** Every row-creation site writes a real path (`executor.ts:1380` `working_path: cwd`; `orchestrator.ts:443` `working_path: workerCwd`), and both explicit-resume set sites populate `resumeRunId` and `resumeRun` together (`command-handler.ts:792-793`, `orchestrator-agent.ts:1417`), so the `resumeRunId`-without-`resumeRun` variant does not occur. A NULL `working_path` survives only on rows predating the column's `ALTER TABLE … ADD COLUMN working_path TEXT` (`sqlite.ts:332`), and such rows necessarily predate `inputs:` (#2470) entirely, so they cannot belong to a required-input workflow. `:914` is defensive/legacy handling.

**Required outcome:** Not a code change on reachability grounds. The correction is to the comment: say that it mirrors the *continuation* branch at `:930`, and note that the explicit-resume-without-working-path refusal at `:914` is preempted by the gate rather than reached. Left as written, the comment asserts a completeness that does not hold, and this file has now produced the same class of bug — an exit that does not participate in the deferral contract — twice.

**Found by:** verification pass, `seams` scope

**Disposition:** FIXED in `3684ee07`. The comment now says "the exact negation of the resume/abandon/force menu's condition" — which is accurate: within the shared `Boolean(working_path)` guard, `willContinueExistingRun` is `P || R` and the menu fires on `!(P || R)`. It additionally records the exit it does *not* mirror and flags the preemption as a deliberate behaviour change rather than leaving it implicit.

</details>

## Verification Performed

### 1. Exit enumeration after a deferral (`dispatchOrchestratorWorkflow`)

Deferral condition: `willContinueExistingRun === true`, i.e. `Boolean(resumableRun?.working_path) && (status === 'paused' || id === options?.resumeRunId)`.

| Exit | Line | Fresh run row? | Held error | Verdict |
|---|---|---|---|---|
| `requires: [github]` refusal | `:834` | no | discarded | Correct — dispatch abandoned, held error moot on a continuation |
| `IsolationBlockedError` | `:901` | no | discarded | Correct — same |
| Explicit resume, no working path | `:914` | no | n/a | Preempted by the gate — see `R2` |
| Resume/abandon/force menu | `:930` | no | n/a | Provably unreachable with a held error (exact mirror) |
| Lost resume race | `:974` | no | appended | Handled — see `R1` for the message shape |
| Resume executes | `:1005` | no (`preCreatedRun`) | discarded | Correct — row holds validated inputs; `:991` reports ignored keys |
| Hydration null → fresh | `:1036` | **yes** | re-raised, returns | Correct — guarded at `:1028` |
| Background dispatch | `:1062` | yes | n/a | Unreachable after a deferral (`else` of truthy `working_path`) |
| Fresh foreground | `:1081` | yes | n/a | Unreachable after a deferral (same) |

Both invariants hold. `resolvedInputs` is additionally sound by construction: it is assigned only by a successful `resolveTopLevelInputs`, so it is either `undefined` or validated — never partially resolved.

Also confirmed: `resolveDeclaredInputs` builds a fresh `resolved` object and never mutates its `supplied` argument (`workflow-inputs.ts:89-108`), so `resolveTopLevelInputs` returning `suppliedMap` genuinely yields supplied-only values. Declared defaults cannot leak onto the run row, which is what AC5 and the "bare run persists nothing" decision depend on. `INPUT_NAME_PATTERN` is non-global (`dag-node.ts:459`), so the `.test()` calls in `parseInputAssignments` are not `lastIndex`-stateful.

### 2. Mutation testing of the three newest tests

Cloned to scratch (`git clone --shared`) with `node_modules` symlinked from the worktree; the shared worktree was never modified. Baseline: **225 pass / 0 fail**.

| Reverted code | Test that failed | Collateral |
|---|---|---|
| `if (deferredInputError) { … return }` at `:1028` | `re-raises the deferred input error when hydration finds nothing to resume` | none (224/1) |
| `willContinueExistingRun` → `Boolean(resumableRun?.working_path)` | `refuses immediately when the resumable run is not a continuation candidate` | none (224/1) |
| the `deferredInputError` append at `:981` | `still reports a deferred input error when the resume race is lost` | none (224/1) |

Each test fails only for its own code. No test is load-bearing for more than one guard, and no guard is untested.

One nit, not filed as a finding: the second test's comment claims "no isolation ran", but it asserts on `mockHydrateResumableRun` (which runs after isolation) rather than on the isolation mock. The test still falsifies correctly; only the comment overstates what is asserted.

### 3. Regressions from the correction commits

- **The hoist (`396f786f`) is clean.** The old lookup site was removed, not duplicated; `resumableRun` remains a single `const`; the `options?.force → null` short-circuit is preserved byte-for-byte. Nothing between the new lookup and the old site mutates its inputs.
- **`--dry-run` rejection (`01ce2a1b`) is correct** and correctly reasoned (`workflow.ts:964-969`): the dry-run engine has no `$INPUTS` support, so an accepted `--input` would simulate with empty values and read as a valid trace. `--input` + `--resume` is separately rejected (`:1033`), and the gate is skipped on resume (`:1115`) — matching the orchestrator's ordering, with the input gate ahead of the `requires: [github]` gate on both surfaces.
- **The docs claim survives checking.** "The CLI has no such implicit continuation" is true: `findResumableRun` is called only under `if (options.resume)` (`workflow.ts:1334-1352`), and the guide already asserted the same thing independently at line 607. The cross-reference anchor `#dag-resume-on-failure` resolves to the `## DAG Resume on Failure` heading at line 611.
- **No behavior change for the 99% case.** A workflow with no `inputs:` takes `resolveDeclaredInputs`' passthrough return and never throws; a defaults-only workflow resolves with nothing supplied and returns `undefined`. Bare-run behavior is byte-for-byte unchanged, as the design requires.

## Delta Confirmation — `5acd0699..3684ee07`

45 lines across two files, implementing `R1` and `R2`. Confirmed rather than re-swept; the verdict was already READY TO MERGE.

**`R1` is fixed correctly, and the test is better than what I asked for.** I proposed one condition plus a fixture change; the pair that landed pins the guard from *both* sides through a shared `dispatchLosingResumeRace(platform, workflowInputs?)` helper. Choosing an undeclared key (`{stlye: 'terse'}`) for the positive case is the right fixture: it is the only way to make a contract violation and a non-empty supplied map coexist, since `resolveDeclaredInputs` checks undeclared keys before missing-required (`workflow-inputs.ts:81-87`), so the thrown error is the undeclared one while `options.inputs` stays non-empty.

Mutation-verified independently in a fresh scratch clone at `3684ee07`. Baseline **226 pass / 0 fail**:

| Mutation | Test that failed | Collateral |
|---|---|---|
| Revert to the unconditional append (the `5acd0699` form) | `stays quiet about the deferred error on a lost race when nothing was supplied` | none (225/1) |
| Drop the append entirely | `reports a deferred input error on a lost resume race when values were supplied` | none (225/1) |
| Delete the fresh-run fallthrough re-raise at `:1028` | `re-raises the deferred input error when hydration finds nothing to resume` | none (225/1) |
| `willContinueExistingRun` → `Boolean(resumableRun?.working_path)` | `refuses immediately when the resumable run is not a continuation candidate` | none (225/1) |

The guard cannot be loosened *or* tightened without a red test, and the two guards the delta did not touch still pin cleanly at the new head. Neither new test is redundant with the other.

**`R2`'s comment is now accurate**, and it records the non-mirrored exit plus the reachability argument as the justification, flagged as a deliberate behaviour change.

**One observation, not a change request.** The new comment's mechanism is stated slightly more strongly than it holds: a legacy row with a NULL `working_path` "predates `inputs:` entirely and therefore cannot reach the gate with a violation to defer." The gate decides against the *current* workflow YAML, not the row's vintage — so a pre-column run whose workflow later gained a required input could technically reach it. That compounds three independent improbabilities on a path that refuses either way at zero cost, and the comment's practical conclusion (unreachable) is right. Not worth another commit; noted so a future reader who spots the same gap knows it was considered.

**Nothing in the delta introduced a regression.** The condition mirrors its `:991` sibling exactly; the append's surrounding comment remains true (it explains *why* the error is surfaced here, and the new guard's rationale lives in the tests, which is the right place for it); no production behaviour outside the lost-race message changed; the exit enumeration above is unaffected, since the delta touched only the *content* of a message on an exit already classified as handled.

## Deliberately Open — Not Findings

Recorded so a later reader does not re-litigate them:

- **No chat/adapter input channel.** Caller decision; [#2555](https://github.com/coleam00/Archon/issues/2555).
- **No typed/non-string input values, no `--input-json`/`--input-file`.** [#2453](https://github.com/coleam00/Archon/issues/2453) owns it; the canonical-representation rule is held so that issue can widen this channel without redesigning it.
- **Collapsing `WorkflowMissingInputsError` into `WorkflowInputContractError`** — declined for this PR, agreed as a follow-up. The originally stated reason ("widens revert blast radius") was factually wrong and a reviewer corrected it: the class predates this PR but was structurally unrelated to `WorkflowInputContractError` until this PR coupled them, and every line the collapse would touch is already rewritten here. The honest ground for deferring is that it is an optional error-hierarchy style tradeoff, not a cost claim, and not worth reopening a five-commit PR over.
- **Branding the gate-validated map** — declined. Both defects actually found in this PR were *omissions* (a field not passed at all, an exit not consulted); a brand catches a wrong-type assignment, not an omission, and would have caught neither. The type analyst accepted this.
- **Isolation resolving before the fallthrough refusal on an explicit resume of a zero-completed-nodes run** — acknowledged and filed as [#2562](https://github.com/coleam00/Archon/issues/2562). Costs a superfluous worktree; no run row and no AI cost either way.

## Verdict

**READY TO MERGE** at `3684ee07`, with no open findings.

The feature does what the plan said, by the route the plan chose, and the part that has been wrong twice is now right for a reason that is checkable rather than remembered — `willContinueExistingRun` being the exact negation of the refuse-menu's condition is what makes that exit unreachable with a held error, instead of a fourth place someone has to remember to check. Every exit after the deferral was enumerated and each either creates no fresh row or re-raises. Four independent guards now sit on this deferral contract and every one of them fails a test when reverted, each isolating exactly one test.

`R1` and `R2` were message-quality issues on paths that already executed correctly, and both are fixed. `R1`'s fix converged on the right principle: with nothing supplied, a held violation is moot, because nothing runs and no channel on that path could have satisfied it — the same reasoning that made the original `396f786f` deferral correct, applied one level down.

Issue #2488, opened 2026-08-05
refactor(workflows): one template walker — collapse five substitution passes and three contradicting KEEP IN SYNC comments


Wirasm:
Text substitution is **five** functions that each walk node fields and replace text, kept in agreement by comments:

- `substituteNodeOutputRefs` (`dag-executor.ts:745`)
- `substituteLoopPrevRefs` (`dag-executor.ts:~3849`)
- `substituteWorkflowVariables` (`executor-shared.ts:~431`)
- `applyInputsMacro` (`include-expander.ts:~231`)
- `rewriteNodeOutputRefs` (`include-expander.ts:171`)

There are **three** `KEEP IN SYNC` comments — `loader.ts:402-406`, `dag-executor.ts:736-739`, `include-expander.ts:161-169` — and **they currently contradict each other.** Two say "three ref-surface enumerations must agree" and list three; the third says "FOUR", lists four, and records why: the count read "three" while a fourth member had already drifted, missing `workflow.fan_out.items`, which *"shipped literal `$INPUTS` text to the model."* The correction was applied to one comment site and not the other two.

Two further hand-maintained enumerations of the same kind carry no `KEEP IN SYNC` label at all: `substituteWorkflowVariables`'s variable list and `applyLoopPrevToBodyNode`'s per-node-type dispatch.

**This is the only primitive that is actively decaying.** The failure was not that a list existed — it is that the *documentation of the list's membership* drifted from the list.

## Proposed

Make the field set a type, and derive every pass from it.

```ts
/** A substitutable text field. Distinct from `string` so the walker can find it. */
type Template = string & { readonly __template?: unique symbol };

type TemplateKeys<B> = { [K in keyof B]-?: IsTemplate<NonNullable<B[K]>> extends true ? K : never }[keyof B];
type NestedKeys<B>   = { [K in keyof B]-?: NonNullable<B[K]> extends object
                           ? IsTemplate<NonNullable<B[K]>> extends true ? never
                             : HasTemplate<NonNullable<B[K]>> extends true ? K : never
                           : never }[keyof B];

/** Must name every Template field, and recurse into every object holding one. */
type SlotSpec<B> =
  & { readonly [K in TemplateKeys<B>]: true }
  & { readonly [K in NestedKeys<B>]: SlotSpec<NonNullable<B[K]>> };

const GATE_SLOTS: SlotSpec<GateBody> = {
  message: true,
  onReject: { prompt: true },        // ← omit this and it does not compile
};

[...]

PR #2570, opened 2026-08-17
refactor(web): the builder had four different ideas of what a node id is

### LINKED ISSUE #2567: refactor(web): two hand-rolled $node.output scanners that disagree, and command categories guessed from filenames
## Problem

Two places in the Web UI reconstruct structure by pattern-matching text, where a typed field either
already exists or should. Both were found during the audit behind the `AGENTS.md` "Natural Language
Is Not a Wire Format" principle (#2564).

### 1. Two hand-rolled `$node.output` scanners that can disagree

The workflow builder validates output references by regexing node bodies for `$nodeId.output`. It is
implemented **twice**, independently:

| Site | Pattern | Fields scanned |
|---|---|---|
| `packages/web/src/hooks/useBuilderValidation.ts:145` | `/\$(\w+)\.output/g` | `when`, `promptText` |
| `packages/web/src/experiments/console/builder/validation/content.ts:25` | `/\$([a-zA-Z_][a-zA-Z0-9_-]*)\.output\b/g` | prompt, command, bash, script, `approval.message` — with code-fence stripping |

Different character classes and different node-type coverage, so the two validators can reach
different conclusions about the same workflow. `\w` excludes `-`, which node ids allow, so the legacy
hook misses references to any hyphenated node id — exactly the ids the bundled workflows use
(`check-reproduction`, `classify-testability`).

The underlying convention is fine and mirrors the engine's own scan in `validateDagStructure()`. The
defect is that it is hand-rolled twice and drifts. The engine already owns the authoritative pattern
(`OUTPUT_REF_SOURCE` / `OUTPUT_REF_PATTERN` in `packages/workflows/src/loader.ts` and
`output-ref.ts`); the UI should not be re-deriving it from memory.

### 2. Command categories guessed from filename prefixes

`packages/web/src/lib/command-categories.ts:9-61` — `findCategory()` sorts bundled commands into
display groups ("Investigation", "Planning", "Code Review", …) by matching each command's filename
against hardcoded prefix lists (`archon-implement-issue` → strip `archon-` → match `implement` →
"Implementation").

`CommandEntry` (`packages/server/src/routes/schemas/workflow.schemas.ts:90`) carries only
`{ name, source }` — there is no category concept anywhere in the engine, so this is invented
client-side classification with nothing behind it. A new or renamed bundled command that matches no
listed prefix silently lands in "Utilities", misrepresenting it in the node palette with no warning.

Contrast `getWorkflowTags` (`packages/web/src/lib/workflow-metadata.ts:169-176`), which does this
correctly: when the workflow declares a typed `tags` field (`workflowBaseSchema`,
`packages/workflows/src/schemas/workflow.ts:173`) it is used verbatim and keyword inference is skipped
entirely. That is the pattern to copy.

## Proposed Solution

**For the scanners:** one source of truth. Either export the engine's `OUTPUT_REF_SOURCE` through the
generated API types so both validators consume it, or delete the legacy hook's copy if the legacy
builder is on its way out. Whichever — there should be exactly one definition of what a `$node.output`
reference looks like.

**For the categories:** decide whether "category" is a real concept. If it is, it belongs on the
command's metadata like `tags` does on workflows, with the client-side inference kept only as an
explicit fallback (mirroring `getWorkflowTags`). If it isn't, drop the grouping and list commands
flat rather than asserting a classification the engine never made.

### Before (current)

```
Author names a node `check-reproduction` and writes "$check-reproduction.output" in a prompt

  legacy builder  ──▶ /\$(\w+)\.output/g  →  no match (hyphen)  →  ref not validated   [!]
  console builder ──▶ matches                                    →  ref validated
  (two validators, two answers, same workflow)                                          [!]

Author adds a bundled command `archon-write-tests`

  node palette    ──▶ matches no prefix list  →  "Utilities"                            [!]
```

### After (proposed)

```
Author names a node `check-reproduction` and writes "$check-reproduction.output" in a prompt

  both builders   ──▶ engine's OUTPUT_REF_SOURCE  →  one answer                         [+]

Author adds a bundled command `archon-write-tests`

  node palette    ──▶ declared category, or flat list — never a silent wrong guess      [+]
```

## Alternatives Considered

|---|---|---|---|
| Share the engine pattern; decide on category (proposed) | One definition; no invented classification | Requires plumbing the constant through generated types | Chosen |
| Fix only the legacy hook's charset to match | One-line change | Leaves two copies to drift again on the next change | Treats the symptom |
| Do nothing until the legacy builder is deleted | Zero work | The console copy is the surviving one and is also hand-rolled; deletion does not fix the category guessing | Only partially self-resolving |

## Scope

- Package(s) likely affected: `web`, possibly `server` (if a category field is added to
  `CommandEntry`)
- Breaking change? `No`

## Security Considerations

## Definition of Done

- [ ] Exactly one definition of the `$node.output` reference pattern is used by the UI
- [ ] A hyphenated node id validates identically in both builders (or the legacy one is gone)
- [ ] Command categories are either engine-declared or dropped — no silent prefix guessing
- [ ] If a fallback inference is kept, it is explicitly a fallback, mirroring `getWorkflowTags`

## Related

- `AGENTS.md` → **Natural Language Is Not a Wire Format** (added in #2564)
- Note for triage: if the legacy chat/builder UI is deleted on the console cutover, the first half
  shrinks to "delete the legacy copy". The category guessing lives in the console path and survives
  either way.

Wirasm:
## Problem and outcome

The workflow builder decided what a `$node.output` reference looks like in **four** hand-written
places, and one of them was wrong: `useBuilderValidation.ts` used `/\$(\w+)\.output/g`, and `\w`
excludes the hyphen. Every hyphenated node id — `check-reproduction`, `classify-testability`, the
ids the bundled workflows actually use — was silently never validated, so the two builders could
reach different conclusions about the same workflow. Separately, the node palette sorted commands
into display categories ("Investigation", "Planning", "Code Review", …) by matching filenames
against hardcoded prefix lists, asserting a classification the engine never made.

- **Outcome:** one definition of a node id in `packages/web`, imported by every scanner, so the two
  builders can no longer disagree about *what a reference is* — a hyphenated ref is now recognised
  by both. Commands are grouped by the `source` they declare instead of by a guess about their
  filename.
- **Invariant:** the web UI's idea of a `$node.output` reference matches the engine's
  (`OUTPUT_REF_SOURCE`, `packages/workflows/src/loader.ts:116`) — and a divergence fails a check
  rather than shipping.
- **Scope boundary:** the legacy builder is **not** deleted. It is still routed for the deprecation
  window (`App.tsx:93`) and removing it is a separate product decision. No server, schema, or API
  change; no new design tokens.
- **Root cause:** `\w` is `[A-Za-z0-9_]` — no hyphen. The hand-rolled copy in the legacy hook had
  drifted from the engine grammar (`[a-zA-Z_][a-zA-Z0-9_-]*`) that the other three copies use.

## Review guidance

- **Feedback requested:** correctness of the shared definition, and whether the parity check is the
  right mechanism given the package boundary.
- **Start here:** `packages/web/src/lib/node-ref.ts` — the single definition everything else now
  imports, plus the docblock explaining why it is a copy rather than an import.
- **Review order:**
  1. `packages/web/src/lib/node-ref.ts` (the definition) and `scripts/node-ref-parity.test.ts` (the
     drift guard against the engine)
  2. `packages/web/src/hooks/useBuilderValidation.ts` (the actual bug) + its new test
  3. `packages/web/src/experiments/console/builder/validation/content.ts` (the other scanner)
  4. `packages/web/src/lib/command-groups.ts` (replaces `command-categories.ts`)
- **Lower-attention areas:** the `category` → `group` identifier rename inside `CommandPicker.tsx`
  and `NodeLibrary.tsx` is mechanical; `when-grammar.ts` and `editor/state.ts` are one-line
  re-points to the shared source, covered by their existing suites.
- **Known risk or uncertainty:** two deliberate behaviour changes worth a look, both argued below —
  dropping the console scanner's `\b`, and removing the now-redundant source badge in
  `CommandPicker`.

## Solution

**One definition.** `packages/web/src/lib/node-ref.ts` exports `NODE_ID_SOURCE` (the engine's id
grammar), `OUTPUT_REF_SOURCE` (named after the engine constant it mirrors), `NODE_ID_PATTERN`
(anchored, for rename validation) and `findOutputRefs()`. `findOutputRefs` builds its own
`g`-flagged RegExp per call rather than reusing a module-level one: a `g` regex carries mutable
`lastIndex`, and sharing a single instance across call sites is how a scan starts skipping matches —
the engine builds a fresh `RegExp` per use for exactly this reason. All four previous copies now
import from this module:

| Site | Was | Now |
|---|---|---|
| `hooks/useBuilderValidation.ts` | `/\$(\w+)\.output/g` — **missed every hyphenated id** | `findOutputRefs()` |
| `…/builder/validation/content.ts` | `/\$([a-zA-Z_][a-zA-Z0-9_-]*)\.output\b/g` | `findOutputRefs()` |
| `…/builder/validation/when-grammar.ts` | id grammar inlined in `ATOM_PATTERN` | `NODE_ID_SOURCE` interpolated |
| `…/builder/editor/state.ts` | `NODE_ID_PATTERN` declared locally | imported |

**Why a copy and not an import.** `@archon/web` must never import `@archon/workflows`, and
`api.generated.d.ts` is type-only so it cannot carry a runtime value — `AGENTS.md` already records
this exact exception for `TRIGGER_RULES`. The constant therefore *cannot* be plumbed through the
generated types. Instead of a KEEP IN SYNC comment, `scripts/node-ref-parity.test.ts` reads **both**
files as text and asserts the two definitions are identical. Comparing two files is a cross-package
repository invariant rather than a unit of `@archon/web` behaviour, so it lives beside the repo's
other cross-artifact checks; `bun run test` ends with `bun test ./scripts/`, so CI enforces it. If
either constant is renamed or moved, it fails with a re-point instruction.

The builder's `when:` parser made the same "mirrors the engine" claim about a **second** grammar —
`ATOM_PATTERN` vs the condition evaluator's `atomPattern` — held by a comment alone. Now that the
mechanism exists, that pair is guarded too, so a divergence can no longer let a condition validate
clean in the builder and fail to parse at run time.

**Two coverage fixes that fall out of sharing.** The legacy hook scanned `when` and `promptText`
but not `bashScript`, which it surfaces and can validate — it now scans all three. And it reported
one issue per *occurrence* of a dangling ref; it now reports one per body, matching the console.
(This narrows the gap between the builders; it does not close it — see *One divergence deliberately
left open* above.)

**Dropping `\b`.** The console pattern ended `\.output\b`; the engine's does not. That divergence
is real: at run time `$ghost.outputs` substitutes `$ghost.output` and leaves the `s`, so it *is* a
reference, and the `\b` made the builder stay quiet about it. Matching the engine is the point of
this change, so the `\b` is gone and a test pins the behaviour.

**Command categories.** `findCategory()` stripped `archon-` and matched the remainder against seven
prefix lists; anything unmatched fell into "Utilities" with no warning, so a new or renamed bundled
command was misrepresented in the palette silently. `CommandEntry` is `{ name, source }` — there is
no category concept anywhere in the engine, so there is nothing to fall back *from*. The precedent
named in the issue is `getWorkflowTags`, which uses a workflow's typed `tags` field verbatim and
only infers when the field is absent; commands have no such field. `command-categories.ts` is
replaced by `command-groups.ts`, which groups on the declared `source` and surfaces an unrecognised
source under its raw value rather than dropping it. A group carries `source` (identity — React key,
collapse state, default-open) separately from `label` (display text), so behaviour never branches on
prose. No `category` field was added to `CommandEntry`: that would invent an engine concept to
satisfy a display preference (YAGNI), and the issue explicitly offers "drop the grouping" when
category is not real.

Because a group header is now exactly the source, `CommandPicker`'s per-row source badge became a
second rendering of the same fact and was removed — a redundancy this change created, so it is
cleaned up here rather than left behind.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Legacy builder ignored `$check-reproduction.output` entirely; console builder flagged it. Bundled commands grouped into invented categories. | Both builders recognise the same references, in every body they each scan. Commands grouped by declared source. |
| **Failure behavior** | An unmatched command name landed silently in "Utilities"; a hyphenated dangling ref produced no warning at all. | An unknown source appears under its raw name; a hyphenated dangling ref warns in both builders. |

**One divergence deliberately left open.** The two builders still look in different *places*. The
legacy panel reads the bodies `resolveNodeDisplay` surfaces (`when`, `promptText`, `bashScript`);
a script node's `script` and an approval node's `approval.message` reach `data` through the
`DagNode` spread but are never copied into those fields, so refs there go unflagged in the legacy
panel while the engine loader and the console builder both validate them. That gap predates this
PR. It is ~4 lines to close, and it is deliberately not closed: the legacy builder is slated for
deletion, so the accurate comment (`useBuilderValidation.ts`) is worth more than coverage on doomed
code. Raised in review as R1; the fix applied was to stop the comment claiming otherwise.

## Architecture

```mermaid
flowchart LR
  E["engine: OUTPUT_REF_SOURCE<br/>workflows/src/loader.ts"] -. "text-compare parity check<br/>scripts/ · no import" .-> N
  N["@/lib/node-ref<br/>NODE_ID_SOURCE · OUTPUT_REF_SOURCE"] --> L["hooks/useBuilderValidation"]
  N --> C["console builder: validation/content"]
  N --> W["console builder: validation/when-grammar"]
  N --> S["console builder: editor/state"]
```

### Changed seams

|---|---|---|
| `@archon/workflows → @archon/web` (node-id grammar) | Was four independent copies, one of them wrong. Now one copy, guarded against the engine by a text-compare check — the import ban is respected. | `packages/web/src/lib/node-ref.ts`, `scripts/node-ref-parity.test.ts` |
| `@archon/workflows → @archon/web` (`when:` atom grammar) | Was a comment-held "mirrors the engine" claim; now guarded by the same mechanism. Pre-existing, guarded opportunistically because this PR built the mechanism. | `…/validation/when-grammar.ts`, `scripts/node-ref-parity.test.ts` |
| `GET /api/commands → node palette` | Display grouping now derives only from the typed `source` field; filename-prefix classification removed. Identity (`source`) split from display (`label`). | `packages/web/src/lib/command-groups.ts`, `command-groups.test.ts` |

## Validation

- `bun run validate` — exit 0 on the committed tree (all ten stages: `check:bundled`,
  `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`,
  `type-check`, `lint --max-warnings 0`, `format:check`, `test:install`, `test`).
- `@archon/web`: 615 tests across 52 files, 0 fail — five separate `bun test` invocations, the
  mock-pollution split (149 / 42 / 18 / 8 / 398). Scripts leg: 22 tests across 2 files.
- Both parity guards were negative-controlled against **mutated copies** in a temp dir (the worktree
  was never touched), confirming each fires and that they are independently sensitive in both
  directions:

  | Mutation | `$node.output` guard | `when:` atom guard |
  |---|---|---|
  | baseline, unmutated | PASS | PASS |
  | web drops the hyphen (the #2567 regression) | **FAIL** | **FAIL** |
  | engine widens `OUTPUT_REF_SOURCE` only | **FAIL** | PASS |
  | engine adds a `when:` operator only | PASS | **FAIL** |
  | web `SEGMENT_SOURCE` drifts | PASS | **FAIL** |

- New regression tests, each tied to a specific defect:
  - `useBuilderValidation.test.ts` — "flags a dangling HYPHENATED reference" fails against the old
    `\w` pattern; also covers the `bashScript` body and the once-per-body dedupe.
  - `scripts/node-ref-parity.test.ts` — byte-identity with each engine literal, plus a
    match-based (not equality-based) pin that a hyphenated id is admitted.
  - `when-grammar.test.ts` — "parses a HYPHENATED node id": every other fixture in that suite is
    hyphen-free, so this is the only case that catches the node-id slot being filled with the
    hyphen-free segment grammar now that `ATOM_PATTERN` is composed.
  - `content.test.ts` — hyphenated upstream ref; `$id.outputs` is a ref (the `\b` removal).
  - `command-groups.test.ts` — "an unrecognised name never changes a command's group" (the
    "Utilities" regression); unknown source surfaced, not dropped; an unknown source colliding with
    a display label stays a separate group.
- **Not verified:** no browser run of the legacy builder — the changed logic is pure and directly
  unit-tested, and the only visual delta is the group labels plus the removed badge in
  `CommandPicker`, both mechanical.

## Links

- Closes #2567
- Related #2564


Wirasm:
---
pr: 2570
base: dev
head: fix/2567-web-node-ref-single-definition
reviewed: 2026-08-17T17:20:00Z
reviewed_head: e9eb0ab265fb9bf75ffa58d3639927d65abb414d
verdict: READY TO MERGE
open_findings: 1
scopes: [code, seams, tests, comments, types]
publication: https://github.com/coleam00/Archon/pull/2570#issuecomment-5316500560
---



# PR Review: #2570 — refactor(web): the builder had four different ideas of what a node id is

> Round 4, at `e9eb0ab2`. Round 1 reviewed `2b8d8a39` → `06569afe` and returned NEEDS FIXES on one
> Important finding; round 2 confirmed `R1`–`R6` fixed at `04266baa`; round 3 covered `3c54b6d5` (the
> author's own catch: the parity extractors were silently defeatable) and folded in `R7`–`R9` plus the
> `types` scope from a second reviewer, `review-2570-r2`, spawned independently by the team lead. This
> round verifies `e9eb0ab2`, which closes `R9` and `R10`. The two reviewers reconciled onto this
> single ledger rather than publishing twice.

## Signal

Ready. The four re-pointed call sites were already behaviourally correct at first review — the
rebuilt `ATOM_PATTERN` expands character-for-character to the literal it replaced, the shared
grammar is byte-identical to the engine's `OUTPUT_REF_SOURCE`, the dropped `\b` is justified by the
engine's actual runtime substituter (`dag-executor.ts:792` has no word boundary), and
`CommandEntry['source']` is the same `workflowSourceSchema` object as `WorkflowSource` server-side.
The code reviewer returned zero findings then and nothing has regressed since.

The blocker was a false completeness claim in prose, and the fix is the right one: rather than
widening a scan on code slated for deletion, `useBuilderValidation.ts` now separates what a
reference *is* (unified) from where one is *looked for* (still narrower than the console), names
`script` and `approval.message` as the residual gap, explains the mechanism — `resolveNodeDisplay`
has no `script` branch and its `approval` branch returns no text — and records that leaving it open
is deliberate. That is a comment that will still be true and still be useful in a year.

The `when:` grammar, which the first pass flagged as having received the shared constant but not the
shared discipline, now has both: a regression test using the condition `archon-validate-pr.yaml`
actually ships, and a second parity pair in `scripts/node-ref-parity.test.ts` comparing
`ATOM_PATTERN` against `condition-evaluator.ts`'s `atomPattern`.

I verified the guards by mutation on copies rather than trusting the author's table, and reproduced
four independent rows: dropping the hyphen in the web copy fails all three parity tests; widening
only the engine's `OUTPUT_REF_SOURCE` fails byte-identity; adding a `when:` operator to the engine
fails only the atom guard; drifting the web's `SEGMENT_SOURCE` fails only the atom guard. Crucially,
a *lockstep* widening applied to all three files passes cleanly — which is the property the R5 fix
exists to preserve, and the reason the author's approach beats both options I proposed.

**Correction to rounds 1–2, and the most important thing in this review.** Every one of those seven
controls perturbed the *values* of the constants. None of them attacked the *extractor*. At
`3c54b6d5` the author found and fixed what that blind spot was hiding: `rawConstant` and
`regexConstant` matched `const <name> =` anywhere in the file, so a commented-out copy holding the
current value — an ordinary thing to find above a constant someone is mid-refactor on — shadows a
genuinely drifted live one and the whole suite passes. I reproduced it on copies: with a
`// const NODE_ID_SOURCE = …` decoy above a live `NODE_ID_SOURCE` that has the #2567 hyphen bug
reintroduced, the pre-`3c54b6d5` extractor reports **3 pass / 0 fail**. The anchored version
(`^\s*(?:export )?const`, `m` flag) reports **0 pass / 3 fail** against the identical decoy. So my
round-2 dispositions on `R4` and `R5` were sound within a value-perturbation threat model and that
threat model was too narrow: I validated what the guard compares, never whether it was reading the
right line. A guard that can be silently defeated is worse than none, and the author caught it
unprompted after the review had already returned READY TO MERGE.

**And the anchoring closes only half of it (`R9`).** `DECL` is `^\s*(?:export )?const`; the `\s*`
after `^` lets a match begin at a line start and then eat indentation, so an *indented* declaration
still matches — which is precisely what a block comment produces. A `/* … const NODE_ID_SOURCE =
<current> … */` decoy above a hyphen-dropped live constant reports **3 pass / 0 fail** against the
shipped `3c54b6d5` extractors: the same silent defeat, spelled `/* */` instead of `//`. The proposed
one-character remedy — dropping `\s*` — turns out to close only the *indented* variant; a column-0
`const` inside `/* */` defeats that too, and is the more likely thing an editor produces. What
actually closes it is stripping comments before matching, verified both ways. **Closed at
`e9eb0ab2`**, which strips comments *and* keeps the line anchor — correctly, and now provably: with
the anchor removed, a mid-line mention in *live* code (which stripping never sees) defeats the guard
3 pass / 0 fail. The two layers cover disjoint shapes. Three iterations of the same class —
unanchored, line-anchored, column-0 — each patching a shape; the commit's lead paragraph finally
names the class, that a regex over raw source cannot tell a declaration from a mention of one.

- **Blocking:** 0
- **Non-blocking:** 1 open (`R7`), plus one tracked follow-up
- **Resolved since previous review:** 8 (`R1`, `R2`, `R4`, `R5`, `R6`, `R8`, `R9`, `R10`)
- **Tracked follow-ups:** 1 — [#2488](https://github.com/coleam00/Archon/issues/2488)
- **Validation:** `bun run validate` PASS (exit 0) at `e9eb0ab2`, plus twelve mutation controls spanning value drift, three decoy shapes, and a live-code mid-line mention

## Findings

|---|---|---|---|
| `R1` | Important | Coverage comment claimed every body is scanned; `script` and `approval.message` are engine ref surfaces the legacy builder misses | FIXED |
| `R2` | Suggestion | No test pinned a hyphenated node id through the recomposed `ATOM_PATTERN` | FIXED |
| `R3` | Suggestion | The parity check pins `loader.ts` only; the engine hand-writes the same grammar in three more places, including the runtime substituter | TRACKED FOLLOW-UP |
| `R4` | Suggestion | `ATOM_PATTERN` carried an unguarded "mirrors the engine" claim | FIXED |
| `R5` | Suggestion | The parity check's hyphen test pinned an exact literal rather than the property its name stated | FIXED |
| `R6` | Suggestion | PR body asserted the R1 falsehood in an earlier section, contradicting its own later correction | FIXED |
| `R7` | Suggestion | A second hand-declared `WorkflowSource` collapses any unknown source to `bundled`, silently opening a workflow read-only — the opposite policy this PR adopts for the same server field | OPEN |
| `R8` | Suggestion | PR body's Validation line reported stale test counts — and the replacement figure this review first proposed was itself only one of five split legs | FIXED |
| `R9` | Suggestion | The anchored extractors close the `//` decoy but not the `/* */` one — a block-commented copy still shadows a drifted live constant and the guard reports green | FIXED |
| `R10` | Suggestion | The `R1` replacement comment attributed `when` to `resolveNodeDisplay`, which does not surface it — a precision error in the one comment whose value is precision | FIXED |

## Detailed Findings

<details>
<summary><code>R1</code> — the coverage claim is false, and the two builders still disagree on approval and script bodies</summary>

**Impact:** A maintainer was told the legacy builder covers every text body its model can express. It
does not — a dangling ref in an approval gate message or a script body passes the legacy validation
panel silently while the console builder flags the identical case.

**Evidence:** `DagNodeData extends DagNode` (`DagNodeComponent.tsx:7`), and `DagNode` declares
`approval.message`, `approval.on_reject.prompt`, `script`, `loop.prompt`. `dagNodesToReactFlow`
spreads the whole node into `data` (`dag-layout.ts`), but `resolveNodeDisplay` has no `script`
branch and its `approval` branch returns only `{ label, nodeType }`. The engine loader
(`loader.ts:428-455`) and the console builder (`content.ts:29-48`) both scan those bodies.

**Required outcome:** Narrow the comment; do not widen the scan.

**Disposition:** FIXED at `968ca5a2`. `useBuilderValidation.ts:143-158` now splits "what a reference
IS" from "WHERE one is looked for", names both missing fields, explains why they never reach the
scanned fields, states that the engine and console do validate them, and records the deletion
rationale for leaving the gap open. The PR body gained a matching "One divergence deliberately left
open" paragraph. Verified line by line against `dag-layout.ts` and `content.ts`. One residual in the
PR body is tracked separately as `R6`.

</details>

<details>
<summary><code>R2</code> — nothing proved the recomposed <code>ATOM_PATTERN</code> still accepts a hyphenated node id</summary>

**Impact:** Composing `ATOM_PATTERN` from `NODE_ID_SOURCE` and a near-identical `SEGMENT_SOURCE`
three lines above created a slip that narrows node ids to hyphen-free while leaving all 14 existing
`when-grammar` tests green — and `archon-validate-pr.yaml:137` ships a hyphenated two-segment
condition.

**Required outcome:** One `parse()` case with a hyphenated id.

**Disposition:** FIXED at `968ca5a2`. `when-grammar.test.ts` gains `parses a HYPHENATED node id`,
using the exact shipped condition `$classify-testability.output.testable == 'e2e_testable'` and
asserting the full atom including `nodeId: 'classify-testability'`. Under the `SEGMENT_SOURCE` slip
`parse()` returns not-ok, so the test fails — the comment above it records precisely that. Choosing
the shipped condition over a synthetic id is the better call.

</details>

<details>
<summary><code>R3</code> — the parity check pins the validator's copy, not the substituter's</summary>

**Impact:** The guard proves the web copy matches `loader.ts`'s load-time validator. Runtime
behaviour is set by `substituteNodeOutputRefs`, a separate hand-written literal. An engine change
touching only the substituter leaves CI green in both packages.

**Evidence:** `loader.ts:116`, `dag-executor.ts:792`, `include-expander.ts:93`, `validator.ts:883` —
four copies, no cross-check. The engine's own `KEEP IN SYNC` block (`loader.ts:412-417`) names three
and tests none.

**Required outcome:** None in this PR. The web-side invariant it claims is fully satisfied.

**Disposition:** TRACKED FOLLOW-UP — [#2488](https://github.com/coleam00/Archon/issues/2488),
verified OPEN. Scope edge recorded: #2488 unifies *traversal*, so `validator.ts:883` may survive it.

**Amendment worth carrying into #2488 (from `review-2570-r2`, verified here).** The end state is not
a third `rawConstant` pair — it is *deleting* the text parsing. `scripts/` is not `@archon/web`, so
the import ban does not bind the parity script: it may legally import from both packages. The web
side is already a pure zero-import leaf (`node-ref.ts`). The only blocker is that the two engine
constants are bare `const`s (`loader.ts:116`, `condition-evaluator.ts:132`). The repo already has the
precedent — `packages/workflows/src/schemas/dag-node.ts:458-459` exports
`INPUT_NAME_SOURCE = String.raw` with the *same* character class, plus an anchored
`INPUT_NAME_PATTERN` mirroring the web's `NODE_ID_PATTERN`. Verified present.

So #2488 should land the ref grammar somewhere **importable** (a leaf module — importing `loader.ts`
today would drag in zod, `@archon/paths` and the schema graph). Then the parity test becomes a value
comparison: no regex over source, the whole decoy class disappears, and a rename becomes a compile
error instead of a runtime throw. That is the same move #2488 already needs, so it is one amendment,
not two.

**Late finding, carried to #2488.** The `R4` atom guard turns out to give this issue partial coverage
as a side effect. Widening the node-id grammar requires two *independent* engine edits —
`loader.ts`'s `OUTPUT_REF_SOURCE` and `condition-evaluator.ts`'s `atomPattern` — and if only one
lands, the atom guard fires while the other two tests stay quiet. Reproduced independently: web +
`loader.ts` widened with `condition-evaluator.ts` missed gives 2 pass / 1 fail, the atom guard alone.
Verified that no other check exists — `grep -rn atomPattern` across `packages/` and `scripts/`
returns exactly one test reference, this one.

Two precisions matter for #2488. First, it is a **transitive** check, not an engine↔engine
assertion: the web copy is the pivot, and it surfaces the disagreement only because it pins to both
constants. Attribution takes reading two results together — atom fires *and* identity stays quiet
implies `condition-evaluator.ts` is the outlier; the failure message alone names the
web↔`condition-evaluator` pair. Second, and the real risk: this coverage exists *because* the web
copy duplicates the grammar. Once #2488 gives the engine one derived source the guard becomes
genuinely redundant — but deleting it as "redundant" **before** that consolidation lands would drop
the only detection there is. Sequence the removal after the unification, not with it.

</details>

<details>
<summary><code>R4</code> — the second "mirrors the engine" claim was still unguarded</summary>

**Impact:** `when-grammar.ts` claimed its `ATOM_PATTERN` mirrors `condition-evaluator.ts`'s
`atomPattern`, held by a comment alone. A later change to the engine's operator set or bare-value
grammar would diverge silently, letting a condition validate clean in the builder and fail to parse
at run time.

**Required outcome:** Guard the pair or file it.

**Disposition:** FIXED at `968ca5a2`. `scripts/node-ref-parity.test.ts` adds a third test comparing
the engine's `atomPattern` (extracted via a new `regexConstant` helper for `/…/` literals) against
the web's composed `ATOM_PATTERN`, resolving both `NODE_ID_SOURCE` and `SEGMENT_SOURCE`
interpolations. Independently mutation-verified twice: adding a `~=` operator to the engine fails
only this test (2 pass / 1 fail), and widening the web's `SEGMENT_SOURCE` fails only this test
(2 pass / 1 fail). Both directions covered.

</details>

<details>
<summary><code>R5</code> — the hyphen test pinned an exact literal, not the property its name stated</summary>

**Impact:** A lockstep grammar widening would pass byte-identity and fail a test named "admits the
hyphen", sending the reader after a drift that is not there.

**Required outcome:** Assert the stated property, or rename the test.

**Found by:** `validation`

**Disposition:** FIXED at `968ca5a2`, and fixed better than either option I offered. The author
rejected `.toContain('-')` on the grounds that it is weak — correct, since `[a-zA-Z0-9_]*-` contains
a hyphen while admitting no hyphenated id — and instead builds an **anchored** RegExp from the
extracted grammar and asserts it matches `check-reproduction` and `classify-testability`. The
anchoring is what makes it sound: `^[a-zA-Z_][a-zA-Z0-9_]*$` does not match a hyphenated id, so the
regression still fails. Verified empirically both ways: dropping the hyphen fails all three tests
(0 pass / 3 fail), and a lockstep widening of the id class across `node-ref.ts`, `loader.ts` and
`condition-evaluator.ts` passes cleanly (3 pass / 0 fail). The rename would have been the weaker
call; no change requested.

</details>

<details>
<summary><code>R6</code> — the PR body contradicts itself on the R1 claim</summary>

**Impact:** Cosmetic and confined to the PR description, but it is the same false sentence `R1` was
raised about. The Solution section still reads "`DagNodeData` carries exactly those three text
bodies, so all three are now scanned", while the later "One divergence deliberately left open"
paragraph correctly explains that `DagNodeData extends DagNode` and carries more. A reader who stops
at the first statement takes away the thing the review corrected.

**Evidence:** PR body, Solution section ("Two coverage fixes that fall out of sharing") versus the
"One divergence deliberately left open" paragraph near the end.

**Required outcome:** Drop or reword the clause in the earlier sentence — the `bashScript` coverage
fix it describes is real and stands on its own without the `exactly those three` claim. No code
change.

**Found by:** `validation` (observed while verifying the `R1` fix)

**Disposition:** FIXED. Verified in the live PR body: the `carries exactly those three text bodies`
clause is gone (grep returns 0), replaced with "scanned `when` and `promptText` but not
`bashScript`, which it surfaces and can validate — it now scans all three", plus an explicit pointer
to the divergence paragraph. The sentence now states the coverage fix without implying completeness.
Head unchanged at `04266baa`, so no re-validation was required.

</details>

<details>
<summary><code>R7</code> — a second <code>WorkflowSource</code> collapses unknown values to <code>bundled</code>, silently opening a workflow read-only</summary>

**Impact:** `packages/web/src/experiments/console/primitives/workflow.ts:1` hand-declares a second
`WorkflowSource` union, and `toWorkflow` (`:24-30`) funnels anything unrecognised into `'bundled'`
via `raw.source === 'project' ? 'project' : raw.source === 'global' ? 'global' : 'bundled'`. That
value feeds `isReadOnlySource` (`builder/connect/save-logic.ts:85-87`, `source === 'bundled'`) and
`saveTargetFor`, so a new or drifted server-side source opens the workflow **read-only** and converts
Save into a project-override Save-as — silently, with no warning to the author who thought they were
editing in place. The file's own docblock at `:25-28` records that this exact guess already shipped
one bug: `global` was demoted to `bundled` and rendered with the wrong badge.

This is the mirror image of the policy #2570 adopts for the same server field: `command-groups.ts`
surfaces an unrecognised source under its raw value rather than collapsing it. Two files, one
`CommandEntry`/`Workflow` source union, opposite failure behaviour.

**Evidence:** `packages/web/src/experiments/console/primitives/workflow.ts:1,24-30`;
`packages/web/src/experiments/console/builder/connect/save-logic.ts:85-87`. Verified independently:
the ternary chain and the `=== 'bundled'` read-only test are both exactly as reported, and
`git diff --name-only origin/dev...HEAD` does **not** contain `primitives/workflow.ts`.

**Required outcome:** None in this PR. It is out of the diff, and #2570 should not grow to cover it.

**Found by:** `review-2570-r2` (`prp-core:seam-analyzer` — its most severe finding)

**Disposition:** OPEN as a Suggestion, and it wants its own issue rather than a line in this ledger,
because ledger entries die with the PR. Worth framing in that issue as the *next instance of the
pattern #2570 just established*: a hand-declared copy of a server union in a package that cannot
import the engine, with no parity guard. #2570 solved that shape for the node-id grammar — one
definition plus a text-read parity check — and the same remedy applies here. Creating issues was
outside this review's authorisation, so it is recommended to the caller with the evidence above.

</details>

<details>
<summary><code>R8</code> — the PR body's validation counts predate the correction rounds</summary>

**Impact:** Cosmetic. The Validation section reports "`@archon/web`: 397 tests across 33 files, 0
fail. Scripts leg: 21 tests across 2 files" — the numbers from before the correction rounds added the
hyphenated `when:` case and the atom parity test. Measured at the current head: web **398**, scripts
leg **22 across 2 files**. A reader reconciling the body against a CI log finds a mismatch and has to
work out which is stale.

**Evidence:** PR body Validation line; `bun test ./scripts/` reports `Ran 22 tests across 2 files`.
Distinct from `R6`, which was the Solution-section self-contradiction and is genuinely fixed.

**Required outcome:** Update the numbers. No code change.

**Found by:** `review-2570-r2`

**Disposition:** FIXED — at `04266baa`, i.e. before this finding was even filed; both `review-2570-r2`
and I were reading a stale body when we raised and confirmed it.

**And the correction runs against this review, not the PR.** The replacement figure both of us cited
— "the actual 398" — was itself wrong. `@archon/web`'s `test` script is five separate `bun test`
invocations (the `mock.module` pollution split), and 398 is only the **last** leg. The package total
is **615 across 52 files** (149 / 42 / 18 / 8 / 398), which is what the body now says. I took 398
from the `@archon/web:test` line in a validate log without noticing the run had four earlier legs
above it — the same species of error as `R1` and `R10`: reading one mechanism's output and
attributing it to the whole. The author's correction is more accurate than the finding that prompted
it.

</details>

<details>
<summary><code>R9</code> — the anchored extractors still admit a block-comment decoy</summary>

**Impact:** `3c54b6d5` fixes the `//` decoy but leaves its twin open. `DECL` is
`^\s*(?:export )?const` (`scripts/node-ref-parity.test.ts:58`), and `\s*` after `^` lets the match
start at a line beginning and then consume leading whitespace — so an *indented* declaration still
matches, and an indented declaration is one of the two shapes a block comment produces:

```
/*
 const NODE_ID_SOURCE = String.raw`[a-zA-Z_][a-zA-Z0-9_-]*`;
*/
export const NODE_ID_SOURCE = String.raw`[a-zA-Z_][a-zA-Z0-9_]*`;   // hyphen dropped
```

The consequence is identical to the one `3c54b6d5` exists to prevent: the #2567 regression is live
in the file and the guard reports green. Toggle-block-comment is a one-keystroke editor action, so
this is not a contrived shape — it is the scenario the commit message already describes ("an
ordinary thing to find above a constant someone is mid-refactor on"), written the other way.

**Evidence:** measured on isolated copies of the four files (nothing in any checkout touched):

| Extractor | Decoy shape | Result |
|---|---|---|
| `3c54b6d5` as shipped | `// const …` | 0 pass / 3 fail — caught |
| `3c54b6d5` as shipped | `/*` + **indented** `const` | **3 pass / 0 fail — silently defeated** |
| `3c54b6d5` as shipped | `/*` + **column-0** `const` | **3 pass / 0 fail — silently defeated** |
| `^(?:export )?const` (proposed fix) | `/*` + **indented** `const` | 0 pass / 3 fail — caught |
| `^(?:export )?const` (proposed fix) | `/*` + **column-0** `const` | **3 pass / 0 fail — still defeated** |
| Strip comments before matching | `/*` + **column-0** `const` | 0 pass / 3 fail — caught |
| Strip comments before matching | clean tree | 3 pass / 0 fail — no false positive |

**Correction to the proposed remedy.** Deleting `\s*` is *not* sufficient, and the reasoning behind
it — "an indented declaration is exactly what a block comment produces" — does not hold. Toggling a
block comment around a top-level declaration commonly leaves it at its original column, and a
column-0 `const` inside `/* */` is textually indistinguishable from a live one to *any* line-anchored
regex. I verified both variants against both anchors (rows 2–5 above): the one-character fix closes
the indented case and is defeated by the column-0 case. That is the more likely editor output of the
two, so the fix would close the less common half.

**Required outcome:** Strip comments before extracting, rather than tightening the anchor —
`src.replace(/\/\*[\s\S]*?\*\//g, '').replace(/^[ \t]*\/\/.*$/gm, '')` ahead of both `.exec` calls.
Verified: closes the column-0 block-comment decoy (0 pass / 3 fail) with no false positive on a clean
tree (3 pass / 0 fail). Deleting `\s*` on top is harmless but no longer load-bearing. The deeper
point is that a regex over raw source cannot tell a declaration from a mention of one; the extractor
has to see code, not text.

**Found by:** `review-2570-r2` (the defeat); remedy corrected by this reviewer on the evidence above

**Scope of the vector — a one-off, not a pattern.** `review-2570-r2` swept the other text-reading
checks under `scripts/`. Only `generate-pi-vendor-map.ts:123` is structurally similar
(`/const envMap = \{([\s\S]*?)\};/.exec(source)`, unanchored, first-match), but it reads
`node_modules/@earendil-works/pi-ai/dist/env-api-keys.js` — vendor build output regenerated by
`bun install` and never hand-edited mid-refactor — so the decoy threat model does not transfer, and
its per-entry regex is strict enough to throw on anything unexpected. `check-bundled-skill.ts`,
`generate-capability-matrix.ts`, `generate-bundled-defaults.ts` and `generate-bundled-schema.ts`
contain no regex extraction at all; they compare whole file contents. No separate issue warranted.

**Disposition:** FIXED at `e9eb0ab2` ("strip comments before extracting — column-0 anchor was half a
fix"). The commit adds `stripComments()` to both extractors *and* keeps the column-0 anchor, with a
docblock carrying the full three-shape matrix. Verified on copies against the committed file: the
column-0 block-comment decoy over a hyphen-dropped live constant gives **0 pass / 3 fail**, and a
clean tree gives **3 pass / 0 fail**.

**Residual, noted not filed.** `stripComments` is itself text-level, so a future regex literal
containing `//` or `/*` inside one of the extracted constants would corrupt the stripped source.
None of the six contains either today. It is a note rather than a finding, and it disappears
entirely under the `R3` amendment (import the constants instead of parsing them).

The author's justification improved on my recommendation, and it is now **measured rather than
argued** — the one claim in that commit resting on reasoning instead of a number. I had called the
anchor dead weight once stripping was in place. It is not. With a *live-code* mid-line mention
holding the current value (`export const DOC_EXAMPLE = ['const NODE_ID_SOURCE = String.raw\`…\`'];`
above a drifted constant — not a comment, so stripping never sees it):

| Configuration | Result |
|---|---|
| shipped `e9eb0ab2` (strip + line anchor) | 0 pass / 3 fail — caught |
| strip, anchor removed | **3 pass / 0 fail — defeated** |

So the two layers cover genuinely disjoint shapes. Credit to `review-2570-r2` for testing the
assertion nobody had.

**One precision on the attribution, verified here.** What rejects the mid-line case is *line
anchoring*, not column-0 specifically: `^\s*(?:export )?const` catches that fixture too (0 pass / 3
fail, measured). Column-0 versus `^\s*` only ever separated the indented-block-comment shape, and
stripping now owns that. The honest decomposition is therefore "stripping + a line anchor", with
column-0 as a free tightening rather than the layer carrying the mid-line case. The shipped code is
correct either way; only the reason was slightly over-attributed, including in an earlier revision of
this ledger. The commit's lead paragraph also names the class rather than the shapes: a regex over raw
source cannot tell a declaration from a mention of one, so this is "cheap steps toward reading code",
not a solved problem. That is the sentence that stops a fourth shape being patched in isolation.

</details>

<details>
<summary><code>R10</code> — the <code>R1</code> replacement comment misattributed which mechanism surfaces <code>when</code></summary>

**Impact:** The `R1` fix said the scan "reads the bodies `resolveNodeDisplay` surfaces
(`dag-layout.ts`): `when`, `promptText`, `bashScript`". `resolveNodeDisplay` returns only
`label`, `nodeType`, `promptText`, `bashScript`, `bashTimeout` (`dag-layout.ts:46-52`) — `when`
arrives on `data` through the `...dn` spread in `dagNodesToReactFlow`, not from
`resolveNodeDisplay`. Small, but this is the one comment in the diff whose entire value is
mechanical precision about which mechanism puts which field where; getting that backwards in the
sentence written to fix `R1` undercuts it.

**Evidence:** `packages/web/src/lib/dag-layout.ts:46-52` (return type) and the
`data: { ...dn, ...resolveNodeDisplay(dn) }` construction below it.

**Found by:** `review-2570-r2` (`prp-core:comment-analyzer`)

**Disposition:** FIXED at `e9eb0ab2`, in the same commit as `R9`. Now reads: "It reads `when` — which
arrives on `data` via the `...dn` spread — plus the two bodies `resolveNodeDisplay` synthesizes
(`dag-layout.ts`): `promptText` and `bashScript`." Verified against the current source.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | No findings — byte-identity, `ATOM_PATTERN` expansion, the `\b` justification, `cn` removal, `WorkflowSource` typing, the ESLint console-spike boundary and the no-new-hex constraint all verified clean |
| seams | `R3`, `R4`; examined and clean: no fifth `$id.output` scanner survives in `packages/web`; `NODE_ID_PATTERN` has no engine counterpart to drift against (`dag-node.ts` types `id` as a bare `z.string()`) and is strictly stricter than the loader, the safe direction; `CommandEntry['source']` and `WorkflowSource` are the same schema object server-side |
| tests | `R2`; every new test file is reached — `src/lib/` and `src/hooks/` legs of `packages/web`'s split `test` script, and the root `bun test ./scripts/` leg for the parity checks — and the parity checks fail loudly, not vacuously, on a renamed, moved, or reformatted constant |
| comments | `R1`; all other changed docblocks verified accurate, including the "ONLY copy in `packages/web`" absolute, the `->` edge-id rationale, the `handleBuilderKeydown` export precedent, and the `getWorkflowTags` citation |
| types (run by `review-2570-r2`) | No findings, and the negative result is worth recording: `OUTPUT_REF_SOURCE` has exactly one consumer and misuse is fail-loud (`matchAll` throws on a non-global regex); `NODE_ID_PATTERN` carries no `g` flag, so the `lastIndex` hazard the factory was designed around is unreachable for it; `CommandGroup` admits neither an empty `commands` array nor a duplicate `source` |

## Validation

|---|---|---|
| `bun run validate` @ `e9eb0ab2` (current head) | PASS | Exit 0, all ten stages; scripts leg `Ran 22 tests across 2 files`, 0 fail |
| **Column-0 decoy vs. the committed `e9eb0ab2` extractor** | FAILS as designed | **0 pass / 3 fail**; clean tree **3 pass / 0 fail**. Confirms `R9` closed |
| `bun run validate` @ `3c54b6d5` | PASS | Exit 0, all ten stages (prior head) |
| **Decoy vs. the PRE-`3c54b6d5` extractor** | **DEFEATED** | `// const NODE_ID_SOURCE = …` decoy holding the correct value, above a live copy with the #2567 hyphen bug: **3 pass / 0 fail**. The guard reported green on a reintroduced bug |
| **Decoy vs. the anchored extractor** | FAILS as designed | Identical decoy: **0 pass / 3 fail**. `^\s*(?:export )?const` + `m` fixes it |
| **Block-comment decoy vs. the anchored extractor** (`R9`) | **DEFEATED** | `/* \n const NODE_ID_SOURCE = <current> \n */` above a hyphen-dropped live copy: **3 pass / 0 fail** at `3c54b6d5`. Independently reproduced by `review-2570-r2` |
| **Column-0 block-comment decoy vs. `^(?:export )?const`** | **STILL DEFEATED** | **3 pass / 0 fail** — the proposed one-character fix closes only the *indented* variant; see `R9` |
| **Column-0 block-comment decoy vs. comment-stripping extractor** | FAILS as designed | **0 pass / 3 fail**, and **3 pass / 0 fail** on a clean tree — the remedy that actually closes `R9` |
| `bun run validate` @ `3c54b6d5` (second, independent run) | PASS | Exit 0 by `review-2570-r2` in a separate detached clone: `@archon/web` 398 pass / 0 fail across 33 files (legs 149 + 42 + 18 + 8 + 398); scripts leg 22 pass / 0 fail |
| No AI attribution; no off-limits file touched | PASS | Zero matches for `Co-Authored-By` / "Generated with Claude" / robot emoji across all six commits and the PR body; `useSSE.ts`, `chat-message-reducer.ts` and `WorkflowLogs.tsx` are absent from the 17-file diff |
| `bun run validate` @ `04266baa` | PASS | Exit 0, all ten stages. `@archon/web` 398 pass / 0 fail; scripts leg 22 pass / 0 fail across 2 files |
| `bun run validate` @ `968ca5a2` | PASS | Exit 0, all ten stages. `@archon/web` 398 pass / 0 fail across 33 files (+1: the hyphenated `when:` case); scripts leg 22 across 2 files (+1: the atom parity test); every package 0 fail |
| Mutation: web drops the hyphen | FAILS as designed | 0 pass / 3 fail — the #2567 regression is caught by all three parity tests |
| Mutation: engine widens `OUTPUT_REF_SOURCE` only | FAILS as designed | byte-identity test fails |
| Mutation: engine adds a `when:` operator (`~=`) | FAILS as designed | 2 pass / 1 fail — only the atom guard, correctly scoped |
| Mutation: web `SEGMENT_SOURCE` drifts | FAILS as designed | 2 pass / 1 fail — only the atom guard |
| Mutation: lockstep widening across all three files | PASSES as designed | 3 pass / 0 fail — confirms `R5`'s fix permits legitimate change |
| Targeted checks @ `04266baa` | PASS | `prettier --check scripts/node-ref-parity.test.ts` clean — the surface a comment-only change could plausibly break |
| `bun run validate` @ `2c0c792d` | PASS | Exit 0 (prior head, superseded) |

All mutation controls ran on copies in a scratch directory. The author's worktree was never modified.

## Verdict

**READY TO MERGE**

Nine of the ten findings are fixed and independently verified; nothing Critical or Important is
open, and validation passes at `e9eb0ab2`. The one that remains — `R7` — is non-blocking, out of the
diff, and wants its own issue rather than a change here. `R3` is engine-side work owned by #2488,
now with a concrete amendment attached.

The verdict is unchanged from round 2, but it is better earned. Between rounds the author found that
the parity extractors could be silently defeated by a commented-out decoy — a defeat class none of my
seven mutation controls touched, because all of them perturbed values and none attacked the
extractor. Reproduced both ways on copies. That is the finding of this review and I did not make it;
recording it matters more than the verdict does.

Three of the fixes are better than what the review asked for: the condition
`archon-validate-pr.yaml` actually ships over a synthetic id; an anchored semantic assertion instead
of either `R5` option I offered; and the extractor anchoring, self-found after the review had already
returned READY TO MERGE. The `R1` correction remains the one that matters most — a comment that draws
the line between what was unified and what deliberately was not is more durable than the coverage it
declined to add.

Wirasm:
## Review disposition — CodeRabbit round 1

Both findings **fixed** in `2c0c792d`. Detail below, plus one self-correction folded into the same commit.

---

### 1. `command-groups.ts` — keep group identity separate from its display label — **FIXED**

Agreed, and for a reason slightly wider than the one raised. The collision scenario itself is remote — `WorkflowSource` is a three-value enum, so reaching it needs a server that emits an out-of-enum value *and* spells it exactly `Project`. But the finding points at a real design fault underneath it: a single `name` field was carrying identity (React key, collapse state, `defaultOpen` check) *and* display text, so `NodeLibrary` was branching behaviour on prose.

`CommandGroup` is now `{ source, label, commands }`. Consumers key off `source`; `defaultOpen` compares `group.source === 'project'` rather than the label string. Regression test added — `an unknown source that collides with a display label stays a separate group` — asserting two distinct groups and two distinct keys.

### 2. `node-ref.test.ts` — move the engine drift guard out of the unit test — **FIXED**

Accepted, though not on the literal reading of the rule. The `Unit tests must not touch real external resources` rule is aimed at unstubbed I/O that *stalls* — its own stated failure mode is an intermittent 5000 ms timeout, and its stated diagnostic is "check whether an `archon.db` appears." A synchronous read of a tracked file in the same checkout is deterministic, sub-millisecond, and fails immediately and legibly, so it is not the hazard the rule describes.

The finding is still right that it was in the wrong place, for a cleaner reason: comparing two files is a **cross-package repository invariant**, not a unit of `@archon/web` behaviour. This repo already has a home for exactly that — the `check:*` parity checks. So the guard moved to `scripts/node-ref-parity.test.ts`, alongside `migrate-state-dir.test.ts`:

- `bun run test` ends with `bun test ./scripts/`, so **CI still enforces it** with no workflow or `package.json` change.
- It now reads *both* sides as text and compares them, rather than reading one and comparing against an import — more symmetric, and it fails with a re-point instruction if either constant is renamed.
- `packages/web/src/lib/node-ref.test.ts` is parser behaviour only.

Proof the guard actually fires, run against the real tree with the historical regression simulated on a copy (no source mutated):

```
real tree  : PASS (identical)
drifted web: FAIL (guard fires correctly)      # hyphen removed from the web copy
```

### 3. Self-correction (not raised, folded in)

`OUTPUT_REF_SOURCE` is now the exported constant name, matching the engine constant it mirrors, so the parity assertion reads as the equality it is. The exported `outputRefPattern()` factory is gone — `findOutputRefs` was its only caller, and it builds its own `RegExp` per call for the same mutable-`lastIndex` reason.

### Not a finding

The OpenGrep annotation on `node-ref.test.ts:17` (`Dynamic command passed to child_process.exec/execSync`) is a false positive — that line is a `path.join`, and neither file spawns a process.

---

**Validation:** `bun run validate` — exit 0 on the committed tree (all ten stages). `@archon/web`: 397 tests / 0 fail. Scripts leg: 21 tests across 2 files.

Wirasm:
## Review disposition — round 2 (all five findings)

Head is now `968ca5a2`. Every finding has a terminal disposition: **R1, R2, R4, R5 fixed; R3 tracked.**

---

### R1 (Important, blocking) — the coverage claim is false — **FIXED**

The finding is correct and I verified it independently before changing anything:
`resolveNodeDisplay` (`dag-layout.ts:46-77`) has **no `script` branch** — a script node falls
through to the `prompt` default and gets `promptText: dn.prompt`, which is undefined — and its
`approval` branch returns `{ label, nodeType }` with no text. So although `DagNodeData extends
DagNode` puts `script` and `approval.message` on `data` via the spread, nothing ever copies them
into the three fields this scan reads. My comment asserted a completeness that does not exist.

Fixed as recommended — **the comment, not the scan**. It now separates the two claims: the builders
can no longer disagree about *what a reference is* (that is what this PR fixes), but they still
differ on *where one is looked for*, and it names `script` / `approval.message` as the residual gap
and says why it is left open. The PR body's "Both builders reach the same verdict on the same ref"
is gone, replaced with an explicit **One divergence deliberately left open** paragraph.

Not closing the gap: ~4 lines on a builder slated for deletion, against an accurate note that tells
the next reader the truth. Agreed with your reasoning here.

### R2 (Suggestion) — nothing pins a hyphenated id through the recomposed `ATOM_PATTERN` — **FIXED**

This was the best catch in the review: the suite is *provably* blind to the slip my own composition
created. Added `parses a HYPHENATED node id` to `when-grammar.test.ts`, asserting
`ast.or[0][0].nodeId === 'classify-testability'`. I used the condition
`archon-validate-pr.yaml:137` actually ships rather than a synthetic id, so the test names the real
workflow it protects.

### R3 (Suggestion) — the guard pins the validator's copy, not the substituter's — **TRACKED FOLLOW-UP**

Accepted as out of scope, and your tracking is right: #2488 owns the engine-side consolidation, and
its acceptance criteria already say the KEEP IN SYNC comments get deleted rather than corrected. No
change here. Noted for whoever takes #2488: the parity check is now positioned to gain the unified
grammar as a few extra lines.

### R4 (Suggestion) — the second "mirrors the engine" claim is unguarded — **FIXED**

You were right that this is the one file where a comment has just been shown to be insufficient.
Verified byte-identity first, then guarded it: `scripts/node-ref-parity.test.ts` now also compares
the builder's `ATOM_PATTERN` against `condition-evaluator.ts`'s `atomPattern`, resolving the web
side's `NODE_ID_SOURCE` / `SEGMENT_SOURCE` interpolations. (Building it surfaced a small real bug in
my own extractor — the web side is a `new RegExp(String.raw…)` form, not a `/…/` literal, so the
first attempt threw. It threw *loudly*, with the re-point message, which was reassuring.)

### R5 (Suggestion) — "admits the hyphen" asserted string equality — **FIXED**

Agreed. Rather than either option offered, the assertion is now **semantic**: build a RegExp from
the extracted grammar and assert it *matches* `check-reproduction` and `classify-testability`. A
legitimate lockstep widening passes; dropping the hyphen fails. Byte-identity stays the previous
test's job, which is the separation the finding was pointing at.

---

### Evidence: both guards negative-controlled

Run against **mutated copies** in a temp dir — the worktree was never modified. This also answers
the implicit question of whether the second guard is independent of the first:

| Mutation | `$node.output` guard | `when:` atom guard |
|---|---|---|
| baseline, unmutated | PASS | PASS |
| web drops the hyphen (the #2567 regression) | **FAIL** | **FAIL** |
| engine widens `OUTPUT_REF_SOURCE` only | **FAIL** | PASS |
| engine adds a `when:` operator only | PASS | **FAIL** |
| web `SEGMENT_SOURCE` drifts | PASS | **FAIL** |

**Validation:** `bun run validate` exit 0 on the committed tree, all ten stages. `@archon/web`
397 pass / 0 fail. Scripts leg 22 tests / 2 files. CI on the previous head `06569afe` was green on
all four required checks including `test (windows-latest)`.

Wirasm:
**R6 — FIXED (PR body only).**

Dropped the `carries exactly those three text bodies` clause from the Solution section. You were right that it was the same false claim as R1, surviving in a second place and contradicting the *One divergence deliberately left open* paragraph further down — exactly the rot this PR is about, so it should not have been left in the document that describes it.

The sentence now says what is actually true: the hook scans `bashScript` because that is a body it surfaces and can validate, and it points at the divergence paragraph rather than implying completeness.

No code change, head stays `04266baa`. Final state: `bun run validate` exit 0; CI green on all five checks at that SHA (ubuntu, windows, docker-build, schema-upgrade, CodeRabbit).

Thanks for the R5 push-back check — the anchoring point is the right one, and it is why the assertion is built as `^…$` rather than a bare match.

Wirasm:
## Completing the negative control: evidence the guards stay QUIET

Follow-up to review feedback that my earlier table proved only that the guards *fire*. That is half a negative control — a guard proven solely by mutations that trip it could be one that trips on everything. The complementary half, run the same way (mutated **copies** in a temp dir; the worktree was never touched):

| Mutation | `identity` | `hyphen` | `atom` |
|---|---|---|---|
| baseline, unmutated | quiet | quiet | quiet |
| **lockstep widening** — web + *both* engine copies allow `.` in ids | quiet | quiet | quiet |
| **partial widening** — web + `loader.ts`, `condition-evaluator.ts` missed | quiet | quiet | **FIRES** |
| cosmetic edit — comment reworded, nothing semantic | quiet | quiet | quiet |

Row 2 is the one that matters for the R5 fix: it verifies the claim the code comment makes — *"a lockstep widening of the grammar on both sides is legitimate and should pass here"*. That was asserted, not tested, until now. It is also what an equality-based assertion would have gotten wrong, so it retires the last doubt about that call.

Row 4 confirms no false positive on ordinary editing.

**Row 3 is a result I did not expect and is worth recording.** The web package has one `NODE_ID_SOURCE`, but the engine hand-writes that grammar separately in `loader.ts` and `condition-evaluator.ts`. So widening node ids requires the engine to change in *two* places, and if only one lands, the `atom` guard fires while the other two stay quiet — pinpointing exactly which engine copy was forgotten.

That makes the R4 guard more than redundancy: it is currently the only automated check anywhere that these two engine constants agree with each other. It catches the #2488 class of drift (four uncoordinated engine copies, none cross-checked) from the outside, as a side effect of guarding the web copy. Worth knowing for whoever picks up #2488.

No code change — head stays `04266baa`, verdict and CI unaffected.

Wirasm:
## The parity guard could be silently defeated — anchored, with the demonstration

Both extractors started at `const <name>` **unanchored**, so they took the FIRST textual match in the file. A commented-out copy holding the *current* value — an ordinary thing to find above a constant someone is mid-refactor on — shadowed a genuinely drifted live one, and the whole suite passed.

Reproduced end-to-end against the real test file rather than argued, since the hole was found by demonstration and an anchored regex that quietly stops matching a real constant would swap a defeatable guard for a dead one. Staged in a temp dir (the test computes `REPO_ROOT` from `import.meta.dir`, so a fake repo root makes it read mutated copies); the worktree was never touched.

Scenario — decoy holds the current value, **live constant is genuinely drifted**, on both extractor paths:

```
// const NODE_ID_SOURCE = String.raw`[a-zA-Z_][a-zA-Z0-9_-]*`;      <- decoy, current
export const NODE_ID_SOURCE = String.raw`[a-zA-Z_][a-zA-Z0-9_]*`;   <- live, hyphen DROPPED

// const atomPattern = /…(==|!=|<=|>=|<|>)…/;                        <- decoy, current
const atomPattern =  /…(==|!=|<=|>=|<|>|=~)…/;                       <- live, gained =~
```

| | live tree | defeat scenario |
|---|---|---|
| **before** (unanchored) | 3 pass / 0 fail | **3 pass / 0 fail** — fully blind |
| **after** (`^\s*(?:export )?const`, `m`) | 3 pass / 0 fail | **0 pass / 3 fail** |

The left column is the half that matters as much as the right: the anchored form still extracts all **six** constants across **four** files, so the fix is not a dead guard. The drift it missed included the #2567 regression itself — the exact defect this PR exists to prevent recurring.

Worth stating plainly, because it is the reason this was folded in rather than deferred: a drift guard that can be silently defeated is **worse than no guard**, since it buys confidence in precisely the invariant it is failing to check.

Head `3c54b6d5`. `bun run validate` exit 0, all ten stages.

Wirasm:
**R8 — FIXED (PR body only), and it was worse than filed.**

You caught 397→398 and 21→22. Checking it turned up a second error in the same sentence: `397 tests across 33 files` was never the package total. `@archon/web`'s `test` script is **five** separate `bun test` invocations — the mock-pollution split documented in AGENTS.md — and I had quoted only the last leg while labelling it as the whole package.

```
Ran 149 tests across 14 files.
Ran  42 tests across  1 file.
Ran  18 tests across  2 files.
Ran   8 tests across  2 files.
Ran 398 tests across 33 files.   <- the only one I was reporting
```

Actual: **615 tests across 52 files**. The line now says that, with the split spelled out, plus the corrected scripts leg (22 across 2).

Mildly embarrassing in a PR about hand-maintained numbers drifting from the thing they describe — the validation section was doing exactly that. Worth the extra minute to recount rather than patch the two digits flagged.

No code change; head stays `3c54b6d5`.

**R7 — not fixed here, deliberately.** Out of this diff, different code path (`isReadOnlySource` / save gating), and closing it means changing save behaviour with its own blast radius. It is a genuine defect and the same species as this PR's — a hand-declared copy of a server union in a package that cannot import the engine, with the collapse-to-a-default policy this PR argues against. It wants its own issue. The team lead is tracking it separately and has asked that it not be folded in here.

Wirasm:
## R9 + R10 fixed — `e9eb0ab2`

### R9: the column-0 anchor was half a fix

`3c54b6d5` closed the `// const …` decoy but not its `/* */` twin. Credit to **review-2570-r2** for finding it and **review-2570-r1** for correcting the remedy after testing both comment shapes against both anchors — the obvious one-character fix (dropping `\s*`) is **not enough**, because toggling a block comment around a top-level `const` usually leaves it at column 0, which is textually indistinguishable from a live declaration to any line-anchored regex.

Measured against the real test file, every row with the #2567 regression live in it:

| extractor | `// …` | `/*` indented | `/*` at column 0 |
|---|---|---|---|
| no anchor | DEFEATED | DEFEATED | DEFEATED |
| `^\s*(?:export )?const` | caught | DEFEATED | DEFEATED |
| `^(?:export )?const` | caught | caught | **DEFEATED** |
| **+ strip comments first** | caught | caught | caught |

Clean tree stays 3 pass / 0 fail throughout — no false positive, and all six constants across four files still extract.

**Both layers kept, deliberately.** They are complementary rather than redundant: stripping removes commented-out copies at any indentation, while the column-0 anchor still rejects a mention embedded mid-line in live code, which stripping does not touch. Recording that so nobody later deletes the anchor as dead weight.

The docblock now leads with the class rather than the three shapes, because that is the part that generalises: *a regex over raw source cannot tell a declaration from a mention of one.* Anchoring narrows which mentions qualify; it never crosses that line. Every further tightening patches a shape — the extractor would have to stop reading text and start reading code. Comment stripping is the cheapest step in that direction, not a solved problem.

### R10: attribution error in the new comment

`useBuilderValidation.ts` said the scan reads "the bodies `resolveNodeDisplay` surfaces: `when`, `promptText`, `bashScript`". `resolveNodeDisplay` returns only `label`/`nodeType`/`promptText`/`bashScript`/`bashTimeout` — `when` arrives via the `...dn` spread. In a comment whose entire value is precision about which mechanism puts which field where, that was backwards. Now: "`when` — which arrives on `data` via the `...dn` spread — plus the two bodies `resolveNodeDisplay` synthesizes".

### Status

`bun run validate` exit 0, all ten stages. **R7** remains deliberately out of scope with its own issue recommended; **R8** was fixed at `04266baa` (and was worse than filed — `397 across 33 files` was one of five split legs, not the package total; real figure 615 across 52).

Wirasm:
## Pointer for #2488 — the end state is deleting this text parsing, not extending it

Raised by **review-2570-r1** and **review-2570-r2**; I verified the load-bearing claim rather than taking it forward on trust.

`scripts/` is **not** `@archon/web`, so the import ban that forced the text-read approach does not bind the parity script — it may import from both packages. The only reason it parses source today is that the two engine constants are bare `const`s with no export.

The repo already has the exact shape, confirmed at `packages/workflows/src/schemas/dag-node.ts:458-459`:

```ts
export const INPUT_NAME_SOURCE = String.raw`[a-zA-Z_][a-zA-Z0-9_-]*`;
export const INPUT_NAME_PATTERN = new RegExp(`^${INPUT_NAME_SOURCE}$`);
```

That is byte-identical to the `NODE_ID_SOURCE` / `NODE_ID_PATTERN` pair `node-ref.ts` hand-maintains. One precision worth keeping, since this PR is about not overclaiming: it is a *different concept* — `with:` input names, not node ids — that happens to share the grammar. So it is not a fifth copy that must agree by contract; it is a fifth place the same character class is written by hand, which is the same maintenance surface for a different reason.

If the grammar lands in an importable leaf module (importing `loader.ts` today drags in zod and the schema graph, which is why it has to be a new module rather than an added export), the parity test becomes a plain value comparison. That deletes the whole class this thread has been circling: no regex over source, the decoy problem gone rather than mitigated, and a rename becomes a compile error instead of a thrown `missing()`. It is also the move #2488 needs anyway.

**Residual, noted not filed:** `stripComments` is still text-level, so a regex literal containing `//` or `/*` *inside* one of the extracted constants would corrupt the stripped source. None of the six does today. The change above erases the concern rather than mitigating it — which is the argument for doing it there instead of hardening the parser here.

PR #2571, opened 2026-08-17
fix(web): the chat UI reads the server's message category instead of guessing from an emoji

Wirasm:
## Problem and outcome

The server already knows which chat messages are workflow narration — it stamps them with a typed `category` and uses that field for its own segmentation — but the live SSE `text` event dropped it, so the legacy chat client re-derived the same decision by testing whether the message text began with 🚀 or ✅. The two implementations disagreed in both directions, which meant the live view and the same conversation reloaded from the database rendered differently.

- **Outcome:** the `text` SSE event carries `category`, and both client reducers segment on it. Neither copy of the emoji regex exists any more.
- **Invariant:** the client and `MessagePersistence` answer "is this a workflow status?" from the same field with the same list of categories, so live rendering and reloaded rendering agree.
- **Scope boundary:** no rendering, styling, or persistence change; `WorkflowLogs`' local reducer stays local (it deliberately lacks the workflow-result and tool-call rules); the console is untouched — it invalidates on `text` and re-reads persisted rows.

## Review guidance

- **Feedback requested:** correctness of the absent-`category` decision, and whether the batching boundary in `useSSE` is the right place to enforce "one segment, one category".
- **Start here:** `packages/server/src/adapters/web.ts` — the one-line wire change everything else follows from.
- **Review order:** `web.ts` (producer) → `packages/web/src/lib/types.ts` (`MessageCategory`, `TextEvent.category`, `ChatMessage.category`) → `useSSE.ts` (batching boundary) → `chat-message-reducer.ts` (`isWorkflowStatusCategory`, both regex tests replaced, and the Rule 5 placeholder adoption) → `WorkflowLogs.tsx` (second copy). Reviewing by commit also works: the first carries the wire change, the second the placeholder gap it exposed.
- **Lower-attention areas:** the five identical `if (buffered) { clearTimeout; flush() }` blocks in `useSSE` collapsed into `flushTextNow()` — mechanical, and the new category boundary needed the sixth.
- **Known risk or uncertainty:** uncategorized `✅` messages lose their own bubble in the live view. Deliberate — see *Behavior change*.

## Solution

`WebAdapter.sendMessage` adds `category` to the `text` payload when the metadata has one. `useSSE` carries it through the 50 ms text batch and flushes when an incoming event's category differs from the buffered one, so a batch never has to speak for two kinds of message — without that, a status line arriving just behind agent prose would concatenate and inherit the wrong category. `applyOnText` and `WorkflowLogs`' reducer both call one shared `isWorkflowStatusCategory()` predicate, and messages retain the category that opened them so the *trailing* boundary can be tested the same way the server tests `lastSeg.category`.

The emoji regex was the only place a second answer to this question existed. Deleting it is the point of the change, not a side effect: the server has been the source of truth all along, and the client was guessing at it from prose.

The second commit closes a gap that only appears once segmentation depends on a stored field. `ChatInterface` pushes an empty `{ content: '', isStreaming: true }` placeholder the moment the user sends, and the `last.content` guard in the status boundary lets text fall straight into it — but the placeholder is built outside the reducer and carries no category. A dispatch status landing there produced a bubble that *looked* like prose to the next event. The first text to reach an uncategorized message now defines its category, which is also how the server behaves: `BufferedSegment.category` is set once when the segment is pushed and never rewritten.

The third commit began as review tidy-up and turned out to fix **a live bug on `dev`**. `pendingWorkflowResultRef` was set by any text event carrying a result and cleared only on flush, so prose and a result arriving inside the same 50 ms window concatenated and flushed as one `onText` carrying the result's metadata — Rule 1 then made a single message, swallowing the agent's prose into the result card, and with two different runs dropping one card entirely and splicing the summaries. `SSETransport` replays its buffer on reconnect, which is exactly how two results land in one window, and is why `applyOnText` already deduplicates by `runId`. Making the finished run part of the batch key fixes it; declaring `workflowResult` on `TextEvent` is what let the batch key be stated honestly.

### The two divergences this fixes

| | Server (`persistence.ts`) | Client (before) |
|---|---|---|
| `workflow_status` message that opens with PR-review context, not the emoji (`executor.ts` prepends `Reviewing PR at commit …` before the 🚀 line) | own segment | **merged** — the `^`-anchored regex misses it |
| Uncategorized `✅ Applied to the live folder` (`dag-executor.ts`) | merged | **own segment** |

### Absent `category` — the deliberate decision

**A text event with no category is agent prose and is not a workflow status.** No text-based fallback survives.

This is not a new rule. `MessagePersistence.appendText` has always treated an uncategorized message as ordinary prose, and every row the client re-renders after a reload was segmented that way. Falling back to the emoji test on absent `category` would preserve exactly the disagreement this PR exists to remove.

Version skew cannot strand a client either: `@archon/web` is built into and served by the same Archon binary, so "new client, old server" is unreachable, and a browser holding a tab open across an upgrade is running the old bundle, which ignores the new field entirely.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A message got its own chat bubble if its text started with 🚀/✅ | A message gets its own bubble if the server categorised it `workflow_status` or `workflow_dispatch_status` |
| **Failure behavior** | An unrecognised message shape silently produced the wrong segmentation, differing from the persisted view | An unrecognised or absent category degrades to prose — the same answer the server gives, so the two views stay consistent |
| **Result cards** | Prose arriving within 50 ms of a workflow result was swallowed into the result card; two results in one window lost a card and spliced their summaries | The finished run is part of the batch key, so each result is its own message and prose stays outside it |

The accepted cost: uncategorized `✅` notices (container write-back, `dag-executor.ts:7886`/`:7955`) merge into the preceding message in the live view — exactly as they already do on reload. If that grouping is wanted, the fix is to give those sends a category at the source, which is one edit at the server rather than a heuristic at the far end of the wire.

## Architecture

```mermaid
flowchart LR
  O["orchestrator / executor<br/>sendMessage(msg, {category})"] --> W["WebAdapter.sendMessage"]
  W --> P["MessagePersistence<br/>segments on category"]
  W ==>|"text event + category"| S["useSSE<br/>startsNewTextBatch()"]
  S --> R["applyOnText / WorkflowLogs<br/>isWorkflowStatusCategory()"]
```

### Changed seams

|---|---|---|
| `WebAdapter.sendMessage → text SSE event` | additive `category` field; omitted for agent prose | `packages/server/src/adapters/web.ts`, `web.test.ts` (3 new tests) |
| `useSSE → onText` | `onText(content, workflowResult?)` → `onText(content, meta?: TextEventMeta)`; a change of category *or* finished run forces a batch flush | `packages/web/src/hooks/useSSE.ts`; both consumers updated, enforced by the type-checker |
| `text event → ChatMessage` | `ChatMessage.category` retains the opening category, mirroring `BufferedSegment.category` server-side | `packages/web/src/lib/types.ts`, `chat-message-reducer.ts` |
| `TextEvent` wire type | `workflowResult` is now declared, not reached through an `in` check and an unchecked cast — the adapter has always emitted it | `packages/web/src/lib/types.ts`, `useSSE.ts` |

## Validation

- `bun run validate` — exit code `0`; 140 test suites, `0 fail` in every one — proves type-check, lint (`--max-warnings 0`), format, and the full per-package suite pass.
- `bun test src/lib/chat-message-reducer.test.ts` (in `packages/web`) — 25 pass / 0 fail — covers both new directions of the category decision (a `workflow_status` message with no leading emoji **does** split; emoji-prefixed text with no category **does not**), the category being stamped onto new messages, adopted by an empty placeholder, and driving the trailing boundary — plus every arm of `startsNewTextBatch`, including two results for different runs.
- `bun test src/adapters/web.test.ts` (in `packages/server`) — 6 pass / 0 fail — the field is on the wire when metadata carries it, the key is absent for prose, and `tool_call_formatted` / `isolation_context` are still suppressed rather than emitted.
- The compiler is load-bearing evidence here: changing `onText`'s signature failed the build at every call site until each was updated, so no consumer was left reading the old shape.
- **Not verified:** the ref plumbing *inside* `useSSE` — buffer append, timer arming, flush-on-unmount — is still exercised only by the type-checker. `@archon/web` has no DOM or React harness and Bun has no `EventSource`, so nothing in the hook can be driven directly; the repo's one hook test (`useBuilderKeyboard.test.ts`) tests an extracted pure function for the same reason. The *decisions* are now extracted and tested (`startsNewTextBatch`, `isWorkflowStatusCategory`, `applyOnText`); what remains untested is the wiring between them. Adding a renderer to this package for that wiring, in a UI slated for replacement by the console, was not judged worth the dependency.
- **CI note:** `test (windows-latest)` failed on an earlier push in `scripts/migrate-state-dir.test.ts` at exactly 5016 ms — Bun's 5000 ms default, the bimodal signature described in `AGENTS.md`. Unrelated to this diff (no file it touches), and the same test failed identically at 5016 ms on `chore/dead-code-and-rulecheck-guide` 17 minutes earlier, so it is environmental and pre-existing rather than introduced here.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Purely additive on the wire; no API response shape changed, so no `generate:types` run is needed. Old clients ignore the field. | `packages/server/src/adapters/web.ts`; `api.generated.d.ts` untouched |
| Rollout / rollback | Three commits, no schema or persisted-state change; revert restores the prior heuristic with no cleanup. | — |

## Links

- Surfaced by the #2564 audit (*Natural Language Is Not a Wire Format*): the client was reconstructing a typed server decision from prose. No separate issue was filed.

## Summary by CodeRabbit

* **New Features**
  * Streamed text events now include message category metadata.
  * Chat messages and workflow logs preserve message types.
  * Streaming content separates automatically when its category or workflow run changes.

* **Bug Fixes**
  * Workflow statuses are identified reliably through metadata instead of text formatting.
  * Tool calls, workflow results, and isolation context messages are handled without misclassification.

### DIFF
--- a/packages/server/src/adapters/web.test.ts
+++ b/packages/server/src/adapters/web.test.ts
@@ -125,3 +125,42 @@ describe('WebAdapter.sendStructuredEvent — tool_result output bounding', () =>
     expect(appendToolResultCalls[0]![2]).toBe(largeOutput);
   });
 });
+
+describe('WebAdapter.sendMessage — text event category', () => {
+  test('carries the metadata category on the text event', async () => {
+    const { adapter, emitted } = makeAdapter();
+
+    await adapter.sendMessage('conv-1', '🚀 Dispatching workflow: **plan**', {
+      category: 'workflow_dispatch_status',
+      segment: 'new',
+    });
+
+    expect(emitted.length).toBe(1);
+    const parsed = JSON.parse(emitted[0]!) as { type: string; category?: string };
+    expect(parsed.type).toBe('text');
+    expect(parsed.category).toBe('workflow_dispatch_status');
+  });
+
+  test('omits the category key entirely for agent prose', async () => {
+    const { adapter, emitted } = makeAdapter();
+
+    await adapter.sendMessage('conv-1', 'ordinary assistant text');
+
+    expect(emitted.length).toBe(1);
+    const parsed = JSON.parse(emitted[0]!) as Record<string, unknown>;
+    expect('category' in parsed).toBe(false);
+  });
+
+  test('still suppresses structurally-handled categories rather than emitting them', async () => {
+    const { adapter, emitted } = makeAdapter();
+
+    await adapter.sendMessage('conv-1', 'formatted tool call', {
+      category: 'tool_call_formatted',
+    });
+    await adapter.sendMessage('conv-1', '📍 repo @ `branch`', {
+      category: 'isolation_context',
+    });
+
+    expect(emitted.length).toBe(0);
+  });
+});
--- a/packages/server/src/adapters/web.ts
+++ b/packages/server/src/adapters/web.ts
@@ -74,11 +74,15 @@ export class WebAdapter implements IWebPlatformAdapter {
       return;
     }
 
+    // `category` rides the wire so the client segments messages from the same
+    // typed signal `MessagePersistence.appendText` uses (persistence.ts), rather
+    // than re-deriving it by pattern-matching the message text.
     const event = JSON.stringify({
       type: 'text',
       content: message,
       isComplete: true,
       timestamp: Date.now(),
+      ...(metadata?.category ? { category: metadata.category } : {}),
       ...(metadata?.workflowResult ? { workflowResult: metadata.workflowResult } : {}),
     });
 
--- a/packages/web/src/components/chat/ChatInterface.tsx
+++ b/packages/web/src/components/chat/ChatInterface.tsx
@@ -26,6 +26,7 @@ import type {
   FileAttachment,
   ToolCallDisplay,
   ErrorDisplay,
+  TextEventMeta,
   WorkflowDispatchEvent,
 } from '@/lib/types';
 import { applyOnText } from '@/lib/chat-message-reducer';
@@ -292,15 +293,12 @@ export function ChatInterface({
     return `msg-${String(messageIdCounter.current)}`;
   };
 
-  const onText = useCallback(
-    (content: string, workflowResult?: { workflowName: string; runId: string }): void => {
-      // First AI text received — the thinking placeholder is about to gain content,
-      // so the hydration merge no longer needs the sendInFlight guard.
-      setSendInFlight(false);
-      setMessages(prev => applyOnText(prev, content, undefined, undefined, workflowResult));
-    },
-    []
-  );
+  const onText = useCallback((content: string, meta?: TextEventMeta): void => {
+    // First AI text received — the thinking placeholder is about to gain content,
+    // so the hydration merge no longer needs the sendInFlight guard.
+    setSendInFlight(false);
+    setMessages(prev => applyOnText(prev, content, undefined, undefined, meta));
+  }, []);
 
   const onToolCall = useCallback(
     (name: string, input: Record<string, unknown>, toolCallId?: string): void => {
--- a/packages/web/src/components/workflows/WorkflowLogs.tsx
+++ b/packages/web/src/components/workflows/WorkflowLogs.tsx
@@ -4,9 +4,10 @@ import { MessageList } from '@/components/chat/MessageList';
 import { useSSE } from '@/hooks/useSSE';
 import { getMessages } from '@/lib/api';
 import { ensureUtc, formatDurationMs } from '@/lib/format';
+import { isWorkflowStatusCategory } from '@/lib/chat-message-reducer';
 import type { MessageResponse } from '@/lib/api';
 import { workflowSSEHandlers } from '@/stores/workflow-store';
-import type { ChatMessage, ToolCallDisplay, ErrorDisplay } from '@/lib/types';
+import type { ChatMessage, ToolCallDisplay, ErrorDisplay, TextEventMeta } from '@/lib/types';
 import type { ToolEvent } from './WorkflowExecution';
 
 interface WorkflowLogsProps {
@@ -433,48 +434,37 @@ export function WorkflowLogs({
     return combined;
   }, [queryMessages, sseMessages, isRunning, gracePolling]);
 
-  const onText = useCallback((content: string): void => {
+  const onText = useCallback((content: string, meta?: TextEventMeta): void => {
     setSseMessages(prev => {
       const last = prev[prev.length - 1];
-      // Workflow status messages (🚀 start, ✅ complete) should be their own message,
-      // matching ChatInterface's behavior and persistence segmentation. Without this,
+      // Workflow status messages should be their own message, matching
+      // ChatInterface's behavior and persistence segmentation. Without this,
       // all text concatenates into one giant streaming message, breaking text dedup
       // against DB messages (which are stored as separate segments).
-      const isWorkflowStatus = /^[\u{1F680}\u{2705}]/u.test(content);
+      const category = meta?.category;
+      const isWorkflowStatus = isWorkflowStatusCategory(category);
+      const newMessage = (): ChatMessage => ({
+        id: `msg-${String(Date.now())}`,
+        role: 'assistant' as const,
+        content,
+        timestamp: Date.now(),
+        isStreaming: true,
+        toolCalls: [],
+        ...(category !== undefined ? { category } : {}),
+      });
 
       if (last?.role === 'assistant' && last.isStreaming) {
-        const lastIsWorkflowStatus = /^[\u{1F680}\u{2705}]/u.test(last.content);
+        const lastIsWorkflowStatus = isWorkflowStatusCategory(last.category);
 
         if ((isWorkflowStatus && last.content) || (lastIsWorkflowStatus && !isWorkflowStatus)) {
           // Close the current streaming message and start a new one when:
           // 1. Incoming is a workflow status and current has content
           // 2. Current is a workflow status and incoming is regular text
-          return [
-            ...prev.slice(0, -1),
-            { ...last, isStreaming: false },
-            {
-              id: `msg-${String(Date.now())}`,
-              role: 'assistant' as const,
-              content,
-              timestamp: Date.now(),
-              isStreaming: true,
-              toolCalls: [],
-            },
-          ];
+          return [...prev.slice(0, -1), { ...last, isStreaming: false }, newMessage()];
         }
         return [...prev.slice(0, -1), { ...last, content: last.content + content }];
       }
-      return [
-        ...prev,
-        {
-          id: `msg-${String(Date.now())}`,
-          role: 'assistant' as const,
-          content,
-          timestamp: Date.now(),
-          isStreaming: true,
-          toolCalls: [],
-        },
-      ];
+      return [...prev, newMessage()];
     });
   }, []);
 
--- a/packages/web/src/hooks/useSSE.ts
+++ b/packages/web/src/hooks/useSSE.ts
@@ -1,5 +1,7 @@
 import { useEffect, useRef, useState, useCallback } from 'react';
 import type {
+  MessageCategory,
+  TextEventMeta,
   SSEEvent,
   ErrorDisplay,
   LoopIterationEvent,
@@ -31,7 +33,7 @@ function parseSSEEvent(raw: string): SSEEvent | null {
 }
 
 interface SSEHandlers {
-  onText: (content: string, workflowResult?: { workflowName: string; runId: string }) => void;
+  onText: (content: string, meta?: TextEventMeta) => void;
   onToolCall: (name: string, input: Record<string, unknown>, toolCallId?: string) => void;
   onToolResult: (name: string, output: string, duration: number, toolCallId?: string) => void;
   onError: (error: ErrorDisplay) => void;
@@ -61,18 +63,33 @@ export function useSSE(
   // Text batching: accumulate text for 50ms before dispatching
   const textBufferRef = useRef('');
   const flushTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const pendingCategoryRef = useRef<MessageCategory | undefined>(undefined);
   const pendingWorkflowResultRef = useRef<{ workflowName: string; runId: string } | undefined>(
     undefined
   );
 
   const flushText = useCallback((): void => {
     if (textBufferRef.current) {
-      handlersRef.current.onText(textBufferRef.current, pendingWorkflowResultRef.current);
+      handlersRef.current.onText(textBufferRef.current, {
+        category: pendingCategoryRef.current,
+        workflowResult: pendingWorkflowResultRef.current,
+      });
       textBufferRef.current = '';
+      pendingCategoryRef.current = undefined;
       pendingWorkflowResultRef.current = undefined;
     }
     flushTimerRef.current = null;
   }, []);
+
+  /** Cancel the batching window and dispatch immediately, if anything is buffered. */
+  const flushTextNow = useCallback((): void => {
+    if (!textBufferRef.current) return;
+    if (flushTimerRef.current) {
+      clearTimeout(flushTimerRef.current);
+      flushTimerRef.current = null;
+    }
+    flushText();
+  }, [flushText]);
 
   useEffect(() => {
     if (!conversationId) return;
@@ -116,7 +133,16 @@ export function useSSE(
 
         switch (data.type) {
           case 'text':
+            // One batch carries one category, so a category change is a message
+            // boundary: flush what is buffered before mixing in text of a
+            // different kind. Without this, a workflow-status line batched behind
+            // agent prose would inherit — or overwrite — the wrong category and
+            // lose its own bubble.
+            if (data.category !== pendingCategoryRef.current) {
+              flushTextNow();
+            }
             textBufferRef.current += data.content;
+            pendingCategoryRef.current = data.category;
             if (
               'workflowResult' in data &&
               data.workflowResult &&
@@ -134,24 +160,12 @@ export function useSSE(
           case 'tool_call':
             // Flush buffered text before tool events to ensure text
             // attaches to the correct message (not the previous one)
-            if (textBufferRef.current) {
-              if (flushTimerRef.current) {
-                clearTimeout(flushTimerRef.current);
-                flushTimerRef.current = null;
-              }
-              flushText();
-            }
+            flushTextNow();
             h.onToolCall(data.name, data.input, data.toolCallId);
             break;
           case 'tool_result':
             // Flush buffered text before tool result too
-            if (textBufferRef.current) {
-              if (flushTimerRef.current) {
-                clearTimeout(flushTimerRef.current);
-                flushTimerRef.current = null;
-              }
-              flushText();
-            }
+            flushTextNow();
             h.onToolResult(data.name, data.output, data.duration, data.toolCallId);
             break;
           case 'error':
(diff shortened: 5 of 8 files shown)


Wirasm:
---
pr: 2571
base: dev
head: fix/sse-text-category-2564
reviewed: 2026-08-17T13:27:38Z
reviewed_head: 2112af301ffbe09eed56890ea9ebcce1151def7c
verdict: READY TO MERGE
open_findings: 2
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2571#issuecomment-5316712200
---



# PR Review: #2571 — fix(web): the chat UI reads the server's message category instead of guessing from an emoji

**Reviewed commit: `2112af301ffbe09eed56890ea9ebcce1151def7c`** (`2112af30`, third of three:
`e1b2032c` → `8b0c13f9` → `2112af30`). This head moved twice mid-review — everything below describes
that SHA and nothing else. If the head has moved again by the time you read this, the delta is
unreviewed.

## Signal

The change does what it says and the reviewers found nothing that blocks it. The client stops
reconstructing a typed server decision from prose, and every consequence of that was traced:
the category union matches `@archon/core` member-for-member, the `flushTextNow()` extraction is
behaviourally identical at all five former inline sites, and the two follow-up commits close the
two real holes in the first one — an empty "thinking" placeholder now adopts the category of the
first text that lands in it, and the `workflowResult` batch identity stops being sticky (which was
a live bug on `dev`: prose arriving behind a result inside the 50 ms window used to be swallowed
into the result message). The disclosed cost — uncategorized `✅` write-back notices losing their
own live bubble — is real, correctly scoped, and pinned by a test. Both open findings are
non-blocking and neither is a defect in the code as written: one is the maintenance cost of the
predicate this PR necessarily duplicates across a package boundary that cannot be compiler-linked,
and the other is a pre-existing console gap the PR explicitly declares a non-goal.

One process note that affects how to read this report: the PR head moved twice during the review
(`e1b2032c` → `8b0c13f9` → `2112af30`), and the new commits were authored inside the checkout the
review was running in. **The first agent pass was invalidated and discarded.** It read a tree that
was neither of those commits: partly `e1b2032c`, partly work that was still uncommitted and had
never been pushed to the PR at all. Anyone comparing this report against that earlier agent output
should treat the earlier output as evidence about nothing in particular. Every finding and
validation result below comes from a detached checkout pinned to **`2112af30`**, which is the
current head.

- **Blocking:** 0
- **Non-blocking:** 2
- **Resolved since previous review:** Not applicable (first review)
- **Tracked follow-ups:** None — `R2` needs an issue filed; see its disposition
- **Validation:** `bun run validate` exit `0` at `2112af30` — all 10 steps, 140 suites, `0 fail` throughout

### What the discarded first pass changes

For anyone reconciling the two: the mixed tree did move conclusions, in both directions.

| First-pass item | Status at `2112af30` | Why |
|---|---|---|
| `category` never persisted → console gap | Retained as `R2` | Real, unchanged, and independently re-verified during synthesis. The files it cites were not part of the concurrent edit |
| `segment` not forwarded on the wire | Dropped — was never real | The premise was wrong, not stale. `persistence.ts:61-75` reads `segment` only as a one-shot force-break; the ongoing boundary is category-driven, so `segment` is not an authority the client is missing |
| `TextEventMeta` hand-duplicated from `TextEvent` | Dropped — genuinely fixed | A real finding at `e1b2032c`. The first pass saw the fix sitting in the working tree as uncommitted scratch and withheld it; that scratch is now commit `2112af30` (`TextEventMeta` derived via `Pick`) |
| Empty-placeholder category adoption cleared as sound | Not valid evidence | The first pass cleared this while reading unpushed work. At `e1b2032c` the gap was genuine; `8b0c13f9` is the commit that closed it, and the pinned pass re-cleared it on the committed code |
| `prp-core:code-reviewer` — zero findings | Unchanged | Zero findings on both the mixed tree and the pinned head |

## Findings

|---|---|---|---|
| `R1` | Suggestion | The workflow-status predicate is now hand-duplicated on both sides of a package boundary with no test or shared source to catch drift | OPEN |
| `R2` | Suggestion | Pre-existing: the web adapter drops `category` when persisting (the CLI adapter keeps it), so the console's system-chatter filter and workflow-result card never engage on web-originated rows | OPEN |

## Detailed Findings

<details>
<summary><code>R1</code> — workflow-status predicate duplicated across the client/server boundary, held together by a one-sided comment</summary>

**Impact:** No failure today — the two lists agree. The cost lands on whoever adds the next
status-like `MessageMetadata.category`: updating only the server half makes `MessagePersistence`
write a separate DB row while the client keeps merging that message into the adjoining prose
bubble. In `WorkflowLogs.tsx` the SSE-vs-DB reconciliation matches on content strings
(`dbTextContents.has(m.content)` / `startsWith`), so a merged live bubble (`"🚀 Startingsome text"`)
matches neither DB row and the text renders **twice**. In `ChatInterface.tsx` the live and
post-refresh bubble counts for the same conversation simply differ. Both are silent — no error, no
log line, just a wrong render.

**Evidence:** `packages/web/src/lib/chat-message-reducer.ts:8-19` (`isWorkflowStatusCategory`, whose
docblock asserts the two "must name the same categories") against
`packages/server/src/adapters/web/persistence.ts:58-59` (the same two literals inline, untouched by
this PR and carrying no reciprocal comment). Reconciliation sites:
`packages/web/src/components/workflows/WorkflowLogs.tsx:392-429`,
`packages/web/src/components/chat/ChatInterface.tsx:184-198`. No test in the repo references both
predicates, and `@archon/web` cannot import a server package, so no compiler link is possible.

**Required outcome:** None for this PR. The duplication is the honest cost of moving the decision
onto the wire — the alternative was the prose regex this PR exists to delete, which was strictly
worse. If cheap insurance is wanted later, the seam analyst's suggestion is a reciprocal comment in
`persistence.ts` pointing back at `isWorkflowStatusCategory`, plus a literal-list assertion in each
package so drift fails a test rather than degrading a render.

**Disposition:** OPEN, non-blocking. Kept open rather than declined because the drift is real and
unenforced, but it does not affect this PR's outcome or invariant: the PR's job was to make the
client and server answer from the same *field*, which it does. The wider 5-member `MessageCategory`
union restatement in `packages/web/src/lib/types.ts:19-30` was examined separately and carved out —
SSE payloads never pass through `registerOpenApiRoute`, so no `api.generated.d.ts` type exists to
derive from, and the union's docblock documents the deliberate degrade-to-prose fallback for an
unrecognized value.

</details>

<details>
<summary><code>R2</code> — pre-existing: the web adapter drops <code>category</code> when persisting, so the console's category-driven rendering never engages on web-originated rows</summary>

**Impact:** Assistant rows written by the **web** adapter reload with `category: null`, so for those
conversations the console's "hide framework narration under a System toggle" behaviour never
activates and a workflow's completion summary silently degrades to a plain chat bubble instead of
`ConsoleWorkflowResultCard`. No error is raised; the feature just never fires. Rows written by the
**CLI** adapter do carry a category and render correctly — which is what makes this an inconsistency
between two adapters rather than an unbuilt feature.

**Evidence:** `packages/server/src/adapters/web/persistence.ts:259-263` builds the persisted
metadata as `{ toolCalls?, workflowDispatch?, workflowResult? }`. The value is already in hand —
`seg.category` rides `BufferedSegment` (`persistence.ts:22`, set at `:81`) and the same class
segments on it at `:58-59` — it is simply omitted from the row. The counterpart that gets it right
is `packages/cli/src/adapters/cli-adapter.ts:52`, `if (metadata?.category) persistMeta.category =
metadata.category;`, under a comment reading "mirror web adapter pattern". The remaining
`addMessage` call sites (`api.ts:2780,3287`, `api.ts:2587`, `workflow.ts:1669`,
`orchestrator-agent.ts:1431`) write `user` rows, where no category applies.
Downstream: `mapMessageRow`'s `category: meta.category ?? null`
(`packages/web/src/experiments/console/primitives/message.ts:152`); `ChatStream.tsx:42` gates the
result card on `message.category === 'workflow_result'`; `RunStream.tsx:151` gates the system filter
on `isSystemCategory(m.category)`, and `message.ts:32` returns `false` for `null`.

**Required outcome:** Nothing in this PR. The fix is one conditional spread in `persistence.ts`'s
`flush()`, mirroring how `workflowDispatch`/`workflowResult` are already handled there and matching
what the CLI adapter already does — a separate change in a file this PR does not touch.

**Correction (2026-08-17, after publication):** an earlier revision of this finding claimed no
`addMessage` call site anywhere writes a `category`. That was wrong — `cli-adapter.ts:52` does. The
author caught it; I verified it before amending. It narrows the blast radius to the web adapter and
reframes the follow-up as adapter inconsistency rather than a missing feature, which is why it is
recorded here rather than silently edited. The conclusion is unchanged.

**Found by:** `prp-core:seam-analyzer` (both passes; verified independently during synthesis,
corrected by the PR author)

**Disposition:** OPEN, non-blocking, and explicitly out of scope — the PR's plan names
"persisting `category` into message metadata" as a non-goal. Not marked `TRACKED FOLLOW-UP` because
that state requires a verified issue and none exists; I have not filed one, since creating a public
issue was not part of this review's mandate. **Recommend filing it** — it is a real user-visible
console gap independent of this PR.

Reviewer disagreement worth preserving: the first (discarded, moving-tree) seam pass rated this
Severity 1, "non-functional today". The pinned pass rated it inert on the grounds that no live event
merges into a reload-produced message. Both are right about different things — the *reducer* does
not need the persisted category (hydrated rows are always `isStreaming: false`, and every
`last.category` read is gated on `isStreaming === true`), but the *console's rendering* does read it
and gets nothing on web-originated rows. Neither pass claims this PR causes or worsens it.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | No findings. Traced the `workflowResult` batch-identity change across five event orderings (no loss, no duplication; the previously sticky ref was the latent bug), Rule 5 category adoption and its reachability via the `ChatInterface.tsx:610-616` empty placeholder, the `flushTextNow()` equivalence at all five sites including the `lock` guard, the full `sendMessage` call-site audit, union parity, import direction, and test quality |
| seams | `R1`, `R2`. Carved out the `MessageCategory` union restatement (no derivable generated type exists; documented fallback). Examined and cleared: `segment` not forwarded (it is a one-shot force-break, not the boundary authority the client needs, and every emitter that sets a wire-relevant category sets it too), `TextEvent.workflowResult` (this PR *replaces* an untyped `'workflowResult' in data` escape hatch), the `Pick<TextEvent, …>` derivation of `TextEventMeta`, and `startsNewTextBatch`'s two-field identity |

## Validation

|---|---|---|
| `bun run validate` (at `2112af30`, pinned detached checkout) | PASS | Exit `0`. All 10 gate steps ran: `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, `type-check`, `lint --max-warnings 0`, `format:check`, `test:install`, `test`. 140 suite summaries, every one `0 fail`; no non-zero fail line in the log. (Error-shaped lines in the output are negative-path test fixtures asserting failure handling.) |
| `bun run validate` (at `e1b2032c`, superseded) | PASS | Exit `0`. Recorded only for completeness — superseded by the pinned run above, and the tree was mutating during it |

Run from `bun run validate`, never `bun test` at the repo root, per the project's mock-pollution rule.

## Verdict

**READY TO MERGE**

No open Critical or Important findings, and the full validation gate passes at the reviewed head.
The absent-`category` decision holds up against the code: `MessagePersistence.appendText` has always
treated an uncategorized message as prose, so deleting the regex makes the live view agree with the
persisted view rather than introducing a new rule. The two disclosed regressions are the intended
ones and are pinned by tests. `R1` and `R2` are informational and neither belongs in this diff —
`R2` is worth an issue on its own.

One caveat on merging: this report describes `2112af30`. If further commits land, the delta needs
re-checking, since the two commits added after the review started were both substantive rather than
cosmetic.

Wirasm:
## Review dispositions

Four findings across the two reviews, all dispositioned. Two were fixed in `2112af30`; two are non-blocking and not acted on here, with reasons.

### C1 — declare `workflowResult` on `TextEvent` (CodeRabbit) — **FIXED** in `2112af30`

Correct, and it is the same defect this PR exists to fix, one field over: the adapter has always emitted `workflowResult`, the wire type never said so, and `useSSE` paid for that with an `in` check plus an unchecked cast. Both are gone. `TextEventMeta` is now `Pick<TextEvent, 'category' | 'workflowResult'>`, so the handler contract cannot drift from the event it describes.

### C2 — flush when the workflow-result identity changes (CodeRabbit) — **FIXED** in `2112af30`

Accepted, and on checking `origin/dev` this was **a live bug, not just identity hygiene** — my commit message undersells it and the review caught that. On `dev`, `pendingWorkflowResultRef` is set by any text event carrying a result and cleared only on flush or retract. So prose and a result arriving inside the same 50 ms window concatenate and flush as one `onText` **with the result's metadata**, which Rule 1 turns into a single non-streaming message: the agent's prose is swallowed into the result card, and with two different runs one card is dropped entirely and the summaries are spliced.

`SSETransport` replays its buffer on reconnect, so back-to-back results are not hypothetical — that replay is precisely why `applyOnText` already deduplicates by `runId`.

The decision is now a pure function rather than a second inline comparison:

```ts
export function startsNewTextBatch(buffered: TextEventMeta, incoming: TextEventMeta): boolean {
  return (
    buffered.category !== incoming.category ||
    buffered.workflowResult?.runId !== incoming.workflowResult?.runId
  );
}
```

Extracted for two reasons. It is testable — `@archon/web` has no DOM harness and Bun has no `EventSource`, so nothing inside the hook can be driven directly, which is why the repo's only hook test (`useBuilderKeyboard.test.ts`) also tests an extracted function. And it states the invariant once: a batch is dispatched as one message, so it carries one identity. The hook now holds a single `pendingMetaRef` for that identity instead of a ref per field, which is also why "clear the pending identity when the incoming event has no result" is handled structurally — the whole meta object is replaced, never patched. Five tests cover the predicate, including two runs inside one window.

### R1 — `isWorkflowStatusCategory` and `persistence.ts:58-59` are the same two literals with no compiler link — **DECLINED for this PR**

Real, and correctly flagged as maintenance cost rather than a defect. Declining because the alternatives are all worse than the exposure:

- `@archon/web` is a client package and cannot import `MessageMetadata` from `@archon/core`, so there is no shared type to reach for. That boundary is the constraint, not an oversight.
- Putting a derived boolean on the wire instead of the category would remove the duplication but throw away the more useful field — the console reads `category` by that same name out of persisted metadata, so the vocabulary has a second consumer already.
- The failure mode is fail-safe in the direction that matters: an unrecognised or renamed category degrades to prose, which is the server's own default for uncategorized text. It cannot invent a status bubble.

What the PR does buy is that the client side now has **one** predicate instead of two hand-rolled regexes in two files, which is the reduction available without crossing the package boundary. If this is ever worth closing properly it belongs with R2 as a single "message category as shared vocabulary" ticket rather than as a fix here.

### R2 — `category` is never persisted, so the console reads null on reloaded history — **TRACKED FOLLOW-UP**, owned by the maintainer

Pre-existing, outside this PR's scope, and explicitly a declared non-goal. Not filed by me — the team lead is routing it to the maintainer as a separate issue.

One correction so that issue is accurate: **it is not true that no call site writes `category`.** The CLI adapter does, at `packages/cli/src/adapters/cli-adapter.ts:52`:

```ts
if (metadata?.category) persistMeta.category = metadata.category;
```

The web adapter is the one that drops it. `MessagePersistence` already carries `category` in memory on `BufferedSegment` (`persistence.ts:22`) and uses it for segmentation, then omits it when building the row metadata at `persistence.ts:259-263`, which spreads only `toolCalls`, `workflowDispatch` and `workflowResult`.

So the accurate framing is an **inconsistency between two adapters**, not a uniformly missing feature — and the fix is one spread in one object, in a file that already has the value in hand. That matters for how the issue gets scoped and estimated.

---

Verification at `2112af30`: `bun run validate` exit 0 (140 suites, `0 fail` in every one) and all four CI checks green, `test (windows-latest)` included.

The earlier red Windows check on `8b0c13f9` was `scripts/migrate-state-dir.test.ts` timing out at 5016 ms — Bun's 5000 ms per-test default. Unrelated to this diff, and environmental rather than introduced: the identical test failed at the identical 5016 ms on an unrelated branch seventeen minutes earlier, and the two tests that time out are the only two in that file that spawn a `bun -e` subprocess to build a real SQLite database before spawning a second subprocess against it — roughly the heaviest thing in the suite, on the slowest runner, under ten parallel package suites. It passed on re-run.

Issue #2008, opened 2026-06-16
HTTP approve/reject cannot resume a CLI-launched paused run: re-dispatches /workflow run to a (missing) web parent instead of resuming in place


vanguy765:
## Summary

`POST /api/workflows/runs/{runId}/approve` (and `/reject`) does **not resume the paused run in place**. It auto-resumes by re-dispatching `\/workflow run <name> <user_message>` to the run's **parent web conversation**. For a run that was launched via the **CLI** (no web parent conversation), there is nothing to dispatch to, so the operator decision never reaches the paused `approval:` node — the post-approval nodes never execute. The CLI's `workflow approve <run-id>` path *does* resume correctly, so HTTP and CLI diverge for the same operation.

This looks like the same root cause as the (closed) #1131 *"Interactive workflows cannot resume after approval gate (parent_conversation_id is NULL)"* — but it still reproduces in v0.4.1 for CLI-launched runs approved over the HTTP API.

## Version

Archon **v0.4.1** (`d83593aa`).

## Root cause

The approve/reject handlers resume by re-sending a `\/workflow run` message to the parent web conversation rather than resuming the existing paused run:

`packages/server/src/routes/api.ts` (approve/reject auto-resume helper, ~L1178-1233):
```ts
const resumeMessage = `/workflow run ${run.workflow_name} ${run.user_message ?? ''}`.trim();
await dispatchToOrchestrator(platformConvId, resumeMessage);
```

The helper's own log events spell out the failure modes: `…auto_resume_skipped_no_platform_conv`, `…auto_resume_skipped_non_web_parent`. A CLI-launched run hits one of these skip paths: no web `platformConvId` → no dispatch → the paused run is never advanced through `record_decision` / post-approval nodes. In our usage the run ends up effectively dead after an HTTP approve (the decision is recorded but the branch past the gate is never taken).

By contrast `workflowApproveCommand` / `workflowRejectCommand` in the CLI resume the actual run (`workflowRunCommand({ resume: true })`) and run the post-gate nodes to completion.

## Reproduction

1. Launch an approval-gate workflow via the **CLI** (`bun run cli workflow run <name> '<args>' --no-worktree`); it pauses at the `approval:` node.
2. Approve it over **HTTP**: `POST /api/workflows/runs/{runId}/approve`.
3. Observe: the post-approval nodes do not run; the decision is not delivered to the paused run. (Approving the *same* run via `bun run cli workflow approve {runId}` instead works.)

## Impact

[...]

PR #2577, opened 2026-08-17
fix(orchestrator): a plain message at an approval gate no longer approves it

### LINKED ISSUE #2565: fix(orchestrator): a plain message at an approval gate always approves, even when it objects
## Problem

While a workflow run is paused at an approval gate in a conversation, **any** message that does not
start with `/` is recorded as an approval. There is no interpretation step and no natural-language
path to reject.

`orchestrator-agent.ts:1248` (`if (!message.startsWith('/'))`) is the first thing checked after
conversation setup. Inside that branch, `approveWorkflow(pausedRun.id, message)` runs unconditionally
(`:1288`). `rejectWorkflow` (`operations/workflow-operations.ts:445`) is never called from that file.

Typing *"no, stop — why is it editing the schema?"* at a gate:

1. resolves the gate as `approved`,
2. records the objection itself as the approval comment (and as `$gate.output` when
   `capture_response: true`),
3. prints `▶️ Resuming …`,
4. continues doing the thing the user objected to.

**This was deliberate**, not an oversight — commit `259d8ecc`, *"feat: natural-language approval
routing for workflow gates (#915)"*. `guides/approval-nodes.md:106-119` documents it as
**"Natural Language (recommended)"**, and its own worked example uses feedback-shaped text
(*"Looks good, but add error handling for the edge cases"*) that auto-approves. The test suite has
7 cases under `describe('natural-language approval routing')`
(`orchestrator-agent.test.ts:2014`) and **zero** negative-message cases; the happy path uses
`'looks good, proceed with implementation'` — text that merely *sounds* like approval. Nothing ever
tested sentiment because nothing ever interpreted it.

**Who experiences it:** every platform. `handleMessage` is the single entry point for Web, CLI,
Slack, Telegram, Discord, GitHub, GitLab and Gitea. Slack is the only surface with a real reject
affordance — the Block Kit button (`adapters/src/chat/slack/workflow-bridge.ts:441-443`) calls
`rejectWorkflow` directly, bypassing `handleMessage` — and only if the user clicks instead of typing.

**Archon also contradicts itself on the identical user action.** At a *loop* gate,
`guides/loop-nodes.md:370-373` states that a plain natural-language message *"always counts as
feedback and iterates. To finalize, use `/workflow approve <id>`"* — deliberately declining to
special-case prose. At an *approval* gate the same message approves, and that is the behaviour
labelled "recommended".

## Proposed Solution

Let the agent interpret the message and act through the typed tool it already has.

The chat agent is already reading the message. Instead of the orchestrator branching on
`!message.startsWith('/')`, a paused gate in the conversation becomes context the agent is given,
and the agent resolves it by calling `manage_run` with an explicit `approve` or `reject` — the same
path `/workflow approve|reject` drives today, and one that already requires `confirm: true` for gate
actions (`DESTRUCTIVE_ACTIONS` / `GATE_ACTIONS` in `orchestrator/manage-run-tool.ts`).

When the message is genuinely ambiguous ("what would that change?"), the agent asks instead of
guessing — which is the outcome the current code cannot produce at all.

Explicitly **not** the fix: inspecting the message for negative keywords to decide approve-vs-reject.
That is the anti-pattern this issue exists to remove, and it would fail on exactly the phrasings that
matter most.

Providers without native tools (Codex/OpenCode/Copilot) already receive the CLI-pointer system prompt
section (`prompt-builder.ts:228`), so they reach the same verbs through `archon workflow
approve|reject`. Both delivery paths already exist; this issue is about deleting the prose branch
that bypasses them.

Whatever is decided here should also reconcile the two gate types so a typed message means the same
thing at a loop gate and an approval gate.

### Before (current)

```
(run paused at gate)
"no, stop — why is it   ────▶ !startsWith('/')  →  approveWorkflow(id, message)   [!]
 editing the schema?"         objection stored as the approval comment            [!]
                        ◀──── "▶️ Resuming ..."
(run proceeds to do the thing the user objected to)                               [!]
```

### After (proposed)

```
User                          Archon                         Agent
────                          ──────                         ─────
(run paused at gate)
"no, stop — why is it   ────▶ paused gate passed   ────────▶ reads the message  [+]
 editing the schema?"         as context                     manage_run(
                                                               action: reject,  [+]
                                                               runId, confirm)
                        ◀──── "Rejected — the run is stopped. It was about to ..."

"what would that       ────▶                      ────────▶ ambiguous → answers [+]
 change?"                                                    the question, gate
                        ◀──── (gate still paused, nothing resolved)              [+]
```

## Alternatives Considered

|---|---|---|---|
| Let the agent decide (proposed) | Feature finally does what its name says; ambiguity becomes a question instead of a guess | Orchestrator work; agent can misjudge | Chosen — gate actions already require explicit confirm, which bounds the failure |
| Keep as-is | Zero work; documented behaviour | Directly violates the `AGENTS.md` "Natural Language Is Not a Wire Format" principle (bullet 3); inverts user intent silently | The principle was adopted knowing it indicts this feature |
| Sentiment/keyword check on the message | No agent involvement | This *is* the anti-pattern; fails on "I don't think so", "hold on", questions | Rejected outright |
| Remove auto-approve entirely | Simplest, safest, makes both gate types consistent by default | Removes a convenience people may use daily | Viable fallback if the agent path proves unreliable |

## Scope

- Package(s) likely affected: `core` (`orchestrator/orchestrator-agent.ts`,
  `orchestrator/manage-run-tool.ts`), `docs`
- Breaking change? `Yes` — a typed message at a gate no longer auto-approves. User-visible and
  intentional; needs a changelog note.

## Security Considerations

- New permissions/capabilities? `No` — the same two verbs, reached through the existing
  confirm-gated tool.
- Note: this **tightens** a governance boundary. Today an objection resolves a human gate as
  approved; afterwards a gate is only resolved by an explicit verb.

## Definition of Done

- [ ] A non-slash message at a paused gate no longer auto-approves
- [ ] The agent can resolve a gate in both directions via `manage_run` (`approve` and `reject`),
      confirm-gated as today
- [ ] An ambiguous message leaves the gate unresolved and asks the user
- [ ] Providers without native tools reach the same verbs via the CLI-pointer path
- [ ] Loop gates and approval gates agree on what a typed message means
- [ ] Tests: a rejecting message rejects; an ambiguous message resolves nothing — the current suite
      has no negative-message coverage at all
- [ ] `guides/approval-nodes.md:106-119` rewritten — it currently recommends the removed behaviour
- [ ] `guides/loop-nodes.md:370-373` checked for consistency with the outcome

## Related

- `AGENTS.md` → **Natural Language Is Not a Wire Format**, bullet 3 (added in #2564) — the principle
  this implements. Its worked example of the violation is this feature.
- #915 — the original PR that introduced the behaviour.

Wirasm:
## Problem and outcome

While a workflow run was paused at a human approval gate, **any** message that did not start with `/` was recorded as an approval. Typing *"no, stop — why is it editing the schema?"* resolved the gate as **approved**, stored the objection itself as the approval comment (and as `$gate.output` under `capture_response: true`), printed `▶️ Resuming …`, and carried on doing the thing the user objected to. There was no interpretation step and no natural-language way to reject.

- **Outcome:** a plain message at a gate no longer approves it. The open gate is given to the chat agent as context for the turn; a clear approval approves, a clear objection rejects with the user's own words as the reason, and an ambiguous message resolves nothing and gets a question back.
- **Invariant:** **whatever resolves a gate also continues the run.** Resolution without continuation is a stranded run, not a neutral refactor — the branch being deleted was the only chat path that did both, so the continuation had to be replaced, not just removed.
- **Scope boundary:** the resolution primitives are unchanged. `approveWorkflow`/`rejectWorkflow` still record and return without resuming, `manage_run`'s `confirm: true` gate on approve/reject is untouched, `resume` stays validate-only, and the web/CLI/Slack surfaces keep the resume paths they already had. This is not #2008's resume-in-place.
- **Root cause:** `orchestrator-agent.ts` branched on `!message.startsWith('/')` as the first thing after conversation setup and called `approveWorkflow(pausedRun.id, message)` unconditionally inside it. `rejectWorkflow` was never reachable from that file. The feature was deliberate (#915) and documented as "Natural Language (recommended)" — it decided an intent without ever interpreting it, which is the inverted form of the violation in `AGENTS.md` → *Natural Language Is Not a Wire Format*, bullet 3.

## Review guidance

- **Feedback requested:** the continuation mechanism, and whether the UX trade is the one you want (a typed approval now goes through the agent's judgement instead of a hard-coded branch). This is a user-visible breaking change — please sign off on the product decision, not just the code.
- **Start here:** `packages/core/src/orchestrator/orchestrator-agent.ts` — `continueResolvedGateRun` plus the `onGateResolved` callback wired into `buildManageRunTool`. That pair is the load-bearing half: it is the same `dispatchOrchestratorWorkflow(..., { resumeRunId, resumeRun })` the deleted branch performed, now triggered by the agent's explicit verb rather than by "the message did not start with `/`".
- **Review order:**
  1. `orchestrator-agent.ts` — the deleted branch, the gate-context fetch before the prompt build, the continuation after the turn.
  2. `manage-run-tool.ts` — the `onGateResolved` seam and `signalGateResolved`.
  3. `prompt-builder.ts` — `formatPausedGateSection`, the only thing that tells the agent a gate is open.
  4. `command-handler.ts` — `/workflow approve|reject` now continue the run (see *Scope note* below).
  5. Docs + `CHANGELOG.md`.
- **Lower-attention areas:** the `resolveRunContinuation` extraction in `command-handler.ts` is a straight lift of the existing `/workflow resume` body — the diff looks large but the logic is unchanged, and the existing resume tests still pass untouched.
- **Known risk or uncertainty:** the agent can now misjudge a borderline message. That risk is bounded by `confirm: true` on gate actions and by the section instructing it to ask rather than guess — but it is a judgement call replacing a deterministic (and wrong) branch. The slash commands remain fully deterministic for anyone who wants that.

### Scope note: why `/workflow approve|reject` are in this diff

Those commands told the user *"Type your response in this conversation to resume the workflow"* — that sentence pointed at the branch this PR deletes, so leaving it would ship a false instruction. It was already stale since #2075 (a resolved gate makes the branch skip), so the commands recorded a decision and parked the run. They now return the same `{ definition, args, resumeRunId, resumeRun }` payload `/workflow resume` already returns, which `handleMessage` already knows how to dispatch. No new mechanism.

## Solution

Three pieces, each reusing an existing primitive:

1. **The gate becomes context, not a branch.** `formatPausedGateSection` renders the open gate — run id, workflow, node, the gate's own message — plus the decision policy, and `buildFullPrompt` places it immediately before the user's message. It deliberately does not name a tool: the verbs reach Claude/Pi through the `manage_run` native tool and Codex/OpenCode/Copilot through the `archon workflow approve|reject` CLI section, both already in the same prompt. A gate that is already resolved renders nothing (it is waiting on the machine, not a human); a malformed approval context renders the explicit-command guidance the old branch used to send directly.

2. **`manage_run` gained a continuation seam, not a continuation.** `onGateResolved(run, action) => boolean` mirrors the existing `startWorkflow` seam. The tool records intent and returns immediately — it never blocks the agent loop on a whole workflow — and the orchestrator drives the resume after the turn ends. The boolean matters: a turn continues **one** run, so a second gate resolved in the same turn is declined and the tool says so in words. A tool reply that promised continuation it did not get would recreate the exact bug this PR fixes.

3. **The continuation is the old code path.** `continueResolvedGateRun` calls `dispatchOrchestratorWorkflow` with `{ resumeRunId, resumeRun }` — byte-identical in effect to what the deleted branch did after `approveWorkflow`. It reuses the workflow list and codebase already fetched for the turn (no extra discovery), and never throws: the decision is committed by then, so a failure costs a manual `/workflow resume`, not the decision.

A reject that **cancels** the run continues nothing — cancelled is its terminal state, and `rejectWorkflow` already reports which outcome happened via `cancelled`.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Any non-slash message at a paused gate → gate approved, message stored as the approval comment, run resumed | The agent reads the gate and the message: clear approval → approved + run continues; clear objection → rejected with the user's words as the reason; ambiguous → nothing resolved, the agent asks |
| **Failure behavior** | An objection resolved the gate as *approved* and the run continued past it — silent inversion of user intent | A gate is only resolved by an explicit verb. If the continuation fails, the decision is still recorded and the user is told to run `/workflow resume <id>` |
| **`/workflow approve\|reject` in chat** | Decision recorded; *"Type your response in this conversation to resume"* — which no longer resumed anything since #2075 | Decision recorded **and** the run continues; a reject that cancels says so and stops |
| **Interactive-loop gates** | A typed message always counted as feedback (iterate) — the opposite position to approval gates on the identical user action | Both gate types agree: a typed message is interpreted, not auto-applied. At a loop gate the user's words still travel as the approve comment, so that route still iterates; finalize with `/workflow approve <id>` (no comment), the CLI/web surfaces, `manage_run` `accept: true`, or `signal_completes` |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1["User: 'no, stop'"] --> B2["!startsWith('/')"] --> B3[approveWorkflow] --> B4[run continues past the objection]
  end

  subgraph After
    A1["User: 'no, stop'"] --> A2[gate passed as context] --> A3[agent judges intent]
    A3 -->|clear objection| A4[manage_run reject] --> A5[run reworks or stops]
    A3 -->|clear approval| A6[manage_run approve] --> A7[run continues]
    A3 -->|ambiguous| A8[answers; gate stays open]
  end
```

### Changed seams

|---|---|---|
| `manage_run → orchestrator` | New `onGateResolved(run, action) => boolean` on `ManageRunContext`. Fires only when the resolution leaves the run resumable; the boolean says whether a continuation was accepted, and the tool words its reply from it | `manage-run-tool.ts` `signalGateResolved`; tests *"approve hands the run to the continuation…"*, *"a declined continuation is reported as a manual resume, not as silence"*, *"without a continuation the tool names the manual resume instead of implying one"* |
| `orchestrator → provider prompt` | New paused-gate section, injected only when an unresolved gate exists in the conversation | `prompt-builder.ts` `formatPausedGateSection`; 9 unit tests, plus *"an open gate is handed to the agent as prompt context"* |
| `command-handler → orchestrator` | `/workflow approve` and `/workflow reject` now return the `workflow` resume payload that `/workflow resume` already returned | `command-handler.ts` `resolveRunContinuation` / `withRunContinuation`; 4 tests under *"/workflow approve\|reject — run continuation"* |
| `handleMessage` routing | The `!message.startsWith('/')` gate branch is gone; `getPausedWorkflowRun` now runs on AI-bound turns only | Test *"a slash command never reaches the gate lookup"* |

## Validation

- `bun run validate` — **exit 0**, 140 test blocks, zero failures — proves type-check, lint (`--max-warnings 0`), format, generated-file freshness, and every package suite pass together.
- `bun test src/orchestrator/orchestrator-agent.test.ts` — **234 pass / 0 fail** — the rewritten suite. The old suite had 7 cases and **zero** negative-message coverage; its happy path used `'looks good, proceed with implementation'`, text that merely *sounds* like approval. The new one covers: an approving-sounding message resolving nothing on its own, a **rejecting** message (`"no, stop — why is it editing the schema?"`) never being recorded as an approval, an **ambiguous** message (`"what would that change?"`) resolving nothing, the gate reaching the prompt (and sitting before the user's message), resolved and malformed gates, slash-command bypass, and — driving the real injected `manage_run` tool through a stubbed agent turn — approve resolving **and** resuming, reject-with-rework resuming, reject-that-cancels not resuming, and an unconfirmed preview resolving nothing. The review round added three more: a provider crash *after* the tool call still continuing the run, a container run resolved but not resumed, and an unscoped conversation getting the gate without instructions to resolve it.
- `bun test src/orchestrator/manage-run-tool.test.ts` — **36 pass / 0 fail**; `src/orchestrator/prompt-builder.test.ts` — **26 pass / 0 fail**; `src/handlers/command-handler.test.ts` — **131 pass / 0 fail**.
- **Not verified:** no live end-to-end run against a real provider. The agent's *judgement* on borderline phrasings is not unit-testable — the tests prove the wiring (a resolution reaches the gate op, and a resolution always either schedules the continuation or says in words why it cannot), and the section prompt is what steers the judgement. A maintainer smoke test on one chat surface before release would be worth doing.

## Delivery considerations

|---|---|---|
| Compatibility / migration | User-visible breaking change: a typed approval at a gate now goes through the agent. No data or schema change; runs paused before the upgrade resolve normally after it. Anyone wanting deterministic behavior uses the slash commands | `CHANGELOG.md` `[Unreleased] → Breaking` |
| Security / permissions / data | Tightens a governance boundary. An objection could previously resolve a human gate as approved; a gate is now only resolved by an explicit, confirm-gated verb. No new permissions or network calls | `DESTRUCTIVE_ACTIONS` / `GATE_ACTIONS` unchanged in `manage-run-tool.ts` |
| Rollout / rollback | Self-contained in `@archon/core` plus docs; one commit, revertible without data migration | — |
| Observability | `orchestrator.paused_gate_context_injected`, `orchestrator.gate_continuation_started/_completed/_failed`, `manage_run.gate_continuation_scheduled/_unavailable` | Log sites in the diff |
| Documentation / communication | `guides/approval-nodes.md` rewritten (the "Natural Language (recommended)" section marketed exactly the removed behavior, with a worked example that auto-approved feedback-shaped text); `guides/loop-nodes.md` and `guides/authoring-workflows.md` reconciled; the `archon` skill reference and three stale code comments corrected | Docs diff |

### How the two gate types now relate

They were contradictory on the identical user action: `loop-nodes.md` said a plain message *"always counts as feedback and iterates"* while `approval-nodes.md` said the same message auto-approves — and called that the recommended path. **Both now say the same thing: a typed message is interpreted by the agent, never auto-applied.** The remaining difference is a consequence, not a policy: the user's words travel through as the approve *comment*, and at an interactive loop a comment means "run another iteration". So asking the agent to approve a loop gate iterates unless nothing is passed on. Finalizing deterministically is still `/workflow approve <id>` with no comment, the CLI/web surfaces, `manage_run` with `accept: true`, or `signal_completes`.

### Follow-up: PR #2529

#2529 (still open at the time of writing) adds a Slack-side re-dispatch of `/workflow resume <id>` after a Block Kit gate approval — the third surface-specific re-dispatch alongside web's `tryAutoResumeAfterGate` and the CLI approve path. This PR lands the general continuation for the chat path, so **if #2529 merges, its Slack dispatch should be deleted rather than layered over.** It is untouched here because it has not merged. #2008 remains the right end state: resume-in-place, independent of surface — deliberately out of scope for this change.

## Links

- Closes #2565
- Related #2008, #2529, #2564, #915

## Summary by CodeRabbit

- **New Features**
  - Approval and rejection commands now automatically continue eligible workflows.
  - Chat agents can resolve approval gates through explicit decisions.
  - Rejections may trigger rework or terminal cancellation, with status updates.
  - Container-isolated workflows receive clear CLI instructions.
  - JSON decisions record outcomes without automatically continuing execution.
  - Ambiguous chat messages leave approval gates unresolved.

- **Documentation**
  - Clarified approval, resumption, interactive-loop, and ambiguous-message behavior across the documentation.

### DIFF
--- a/.claude/skills/archon/references/workflow-dag.md
+++ b/.claude/skills/archon/references/workflow-dag.md
@@ -664,9 +664,12 @@ archon workflow reject <run-id> --reason "plan needs more test coverage"
 /workflow approve <run-id> <optional comment>
 /workflow reject <run-id> <optional reason>
 
-# Natural language (all platforms except CLI — auto-detects paused workflow)
+# Asking the chat agent (all platforms except CLI)
 User: "Looks good, proceed"
-# → auto-approves. With capture_response: true, the message becomes $review-gate.output
+# → the agent reads the open gate and approves it; the run continues.
+#   Your words travel through as the comment, so with capture_response: true they
+#   become $review-gate.output. A plain message is NOT an automatic approval —
+#   an objection rejects, and an ambiguous message resolves nothing.
 ```
 
 ### What Does NOT Work on Approval Nodes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking
+
+- **A plain message at an approval gate no longer approves it.** While a run was paused at a human gate, any message that did not start with `/` was recorded as an approval — so "no, stop, why is it editing the schema?" resolved the gate as approved, stored the objection itself as the approval comment (and as `$gate.output` under `capture_response: true`), and continued doing the thing the user objected to. There was no interpretation step and no natural-language way to reject. The open gate is now given to the chat agent as context for the turn, and the agent resolves it through the same confirm-gated `approve`/`reject` verbs the slash commands drive: a clear approval approves, a clear objection rejects with the user's own words as the reason, and an ambiguous message resolves nothing and gets a question back. Providers without native tools reach the same verbs over the `archon workflow approve|reject` CLI. Resolving a gate now also continues the run on every chat surface — including `/workflow approve` and `/workflow reject`, which previously told you to "type your response in this conversation to resume" and left the run parked. Deterministic behaviour is still available: use the slash commands. (#2565)
+
 ## [0.9.0] - 2026-08-17
 
 Workflows become properly composable: a workflow can now declare the arguments it takes and the result it returns, ship as one self-contained folder, and be simulated end to end without contacting a provider. Alongside that, a workflow node's capabilities become what its YAML declares rather than whatever happens to be configured on the operator's machine — see Breaking before upgrading.
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -2787,7 +2787,8 @@ export async function workflowApproveCommand(
   const resolvedId = await resolveRunIdArg(runId, cwd);
   const result = await approveWorkflow(resolvedId, comment);
 
-  // CLI auto-resumes after approval (unlike chat, which defers to next user message)
+  // CLI auto-resumes after approval, as chat does since #2565. `--json` (handled
+  // above) is the one surface that records the decision without continuing.
   if (!result.workingPath) {
     throw new Error(
       `Workflow run '${resolvedId}' has no working path recorded.\n` +
--- a/packages/core/src/db/workflows.ts
+++ b/packages/core/src/db/workflows.ts
@@ -428,7 +428,8 @@ export async function getActiveWorkflowRun(conversationId: string): Promise<Work
 
 /**
  * Find a paused workflow run for a conversation (or its parent).
- * Used by the message handler to detect approval gates awaiting a natural-language response.
+ * Used by the message handler to give the chat agent the open approval gate as
+ * context for the turn (#2565).
  * Non-throwing: returns null on DB error so the caller can fall through to normal routing.
  */
 export async function getPausedWorkflowRun(conversationId: string): Promise<WorkflowRun | null> {
--- a/packages/core/src/handlers/command-handler.test.ts
+++ b/packages/core/src/handlers/command-handler.test.ts
@@ -2400,6 +2400,124 @@ describe('CommandHandler', () => {
       });
     });
 
+    // A gate decision that does not continue the run leaves it stranded (#2565).
+    // Before #2565 these commands told the user to "type your response to
+    // resume", which relied on a natural-language branch that no longer exists.
+    describe('/workflow approve|reject — run continuation', () => {
+      const baseConversation: Conversation = {
+        id: 'conv-approve',
+        platform_type: 'telegram',
+        platform_conversation_id: 'chat-approve',
+        ai_assistant_type: 'claude',
+        codebase_id: null,
+        cwd: null,
+        isolation_env_id: null,
+        last_activity_at: null,
+        created_at: new Date(),
+        updated_at: new Date(),
+      };
+
+      function pausedRun(overrides: Record<string, unknown> = {}) {
+        return {
+          id: 'run-gate',
+          workflow_name: 'gated-wf',
+          conversation_id: 'conv-approve',
+          parent_conversation_id: null,
+          codebase_id: null,
+          status: 'paused' as const,
+          user_message: 'original prompt',
+          metadata: { approval: { type: 'approval', nodeId: 'review', message: 'Approve?' } },
+          started_at: new Date(),
+          completed_at: null,
+          last_activity_at: new Date(),
+          working_path: '/repo',
+          ...overrides,
+        };
+      }
+
+      /** The gate op reads the run, then the continuation reads it again. */
+      function stubRunReads(run: ReturnType<typeof pausedRun>): void {
+        mockGetWorkflowRun.mockResolvedValueOnce(run).mockResolvedValueOnce(run);
+      }
+
+      function stubWorkflowDiscovery(): void {
+        spyDiscoverWorkflows?.mockResolvedValue({
+          workflows: [makeTestWorkflowWithSource({ name: 'gated-wf' })],
+          errors: [],
+        });
+      }
+
+      test('approve hands the orchestrator the resume payload for the same run', async () => {
+        const run = pausedRun();
+        stubRunReads(run);
+        stubWorkflowDiscovery();
+
+        const result = await handleCommand(baseConversation, '/workflow approve run-gate LGTM');
+
+        expect(result.success).toBe(true);
+        expect(result.message).toContain('approved');
+        expect(result.message).toContain('Resuming');
+        expect(result.workflow?.resumeRunId).toBe('run-gate');
+        expect(result.workflow?.resumeRun).toBe(run);
+        expect(result.workflow?.definition.name).toBe('gated-wf');
+        // The run's own prompt drives the resume, not the approve comment.
+        expect(result.workflow?.args).toBe('original prompt');
+      });
+
+      test('reject with an on_reject rework hands back the resume payload', async () => {
+        const run = pausedRun({
+          metadata: {
+            approval: {
+              type: 'approval',
+              nodeId: 'review',
+              message: 'Approve?',
+              onRejectPrompt: 'Address $REJECTION_REASON',
+            },
+          },
+        });
+        stubRunReads(run);
+        stubWorkflowDiscovery();
+
+        const result = await handleCommand(
+          baseConversation,
+          '/workflow reject run-gate schema is wrong'
+        );
+
+        expect(result.success).toBe(true);
+        expect(result.message).toContain('Reworking');
+        expect(result.workflow?.resumeRunId).toBe('run-gate');
+      });
+
+      test('reject that cancels the run hands back nothing to resume', async () => {
+        // No on_reject prompt ⇒ the run is cancelled, which IS its terminal
+        // state — a resume payload here would try to restart a dead run.
+        mockGetWorkflowRun.mockResolvedValueOnce(pausedRun());
+
+        const result = await handleCommand(baseConversation, '/workflow reject run-gate no thanks');
+
+        expect(result.success).toBe(true);
+        expect(result.message).toContain('rejected and cancelled');
+        expect(result.workflow).toBeUndefined();
+      });
+
+      test('an unresolvable continuation still reports the decision as recorded', async () => {
+        const run = pausedRun();
+        stubRunReads(run);
+        // The workflow YAML is gone, so the run cannot be continued.
+        spyDiscoverWorkflows?.mockResolvedValue({ workflows: [], errors: [] });
+
+        const result = await handleCommand(baseConversation, '/workflow approve run-gate');
+
+        // success:false would send the user to re-approve a gate that is already
+        // resolved — and the second approve throws.
+        expect(result.success).toBe(true);
+        expect(result.message).toContain('approved');
+        expect(result.message).toContain('could not be continued automatically');
+        expect(result.message).toContain('/workflow resume run-gate');
+        expect(result.workflow).toBeUndefined();
+      });
+    });
+
     describe('/workflow approve — standard approval node with captureResponse', () => {
       const baseConversation: Conversation = {
         id: 'conv-approve',
--- a/packages/core/src/handlers/command-handler.test.ts
+++ b/packages/core/src/handlers/command-handler.test.ts
@@ -2500,6 +2500,23 @@ describe('CommandHandler', () => {
         expect(result.workflow).toBeUndefined();
       });
 
+      test('a container run is resolved but points at the CLI instead of resuming', async () => {
+        // Chat cannot rewire the container, so dispatching a resume would fail
+        // the run to say what this message says for free.
+        const run = pausedRun({ metadata: { ...pausedRun().metadata, isolation: 'container' } });
+        stubRunReads(run);
+        stubWorkflowDiscovery();
+
+        const result = await handleCommand(baseConversation, '/workflow approve run-gate');
+
+        expect(result.success).toBe(true);
+        expect(result.message).toContain('approved');
+        expect(result.message).toContain('isolation container');
+        expect(result.message).toContain('archon workflow resume run-gate');
+        expect(result.message).not.toContain('/workflow resume run-gate');
+        expect(result.workflow).toBeUndefined();
+      });
+
       test('an unresolvable continuation still reports the decision as recorded', async () => {
         const run = pausedRun();
         stubRunReads(run);
--- a/packages/core/src/handlers/command-handler.ts
+++ b/packages/core/src/handlers/command-handler.ts
@@ -264,6 +264,107 @@ function findWorkflowLoadError(
   return loadErrors.find(error => error.filename.replace(/\.ya?ml$/, '') === workflowName);
 }
 
+/**
+ * Resolve everything the orchestrator needs to continue a resumable run: the run
+ * itself plus its workflow definition, packaged as the `workflow` payload of a
+ * `CommandResult`. Returns a user-facing failure message instead when the run is
+ * not resumable or its workflow can no longer be loaded.
+ *
+ * Shared by `/workflow resume`, `/workflow approve` and `/workflow reject`
+ * (#2565): a gate decision that does not continue the run leaves it stranded, so
+ * all three resolve the same continuation the same way.
+ */
+async function resolveRunContinuation(
+  runId: string,
+  workflowCwd: string
+): Promise<
+  | { ok: true; workflow: NonNullable<CommandResult['workflow']>; workflowName: string }
+  | { ok: false; message: string }
+> {
+  const run = await resumeWorkflow(runId);
+  let workflowEntries: readonly WorkflowWithSource[];
+  let loadErrors: readonly WorkflowLoadError[];
+  try {
+    const result = await discoverWorkflowsWithConfig(workflowCwd, loadConfig);
+    workflowEntries = result.workflows;
+    loadErrors = result.errors;
+  } catch (error) {
+    const err = error as Error;
+    getLog().error({ err, cwd: workflowCwd, runId }, 'cmd.workflow_resume_discovery_failed');
+    return {
+      ok: false,
+      message: `Failed to load workflows: ${err.message}\n\nCheck .archon/workflows/ for YAML syntax issues.`,
+    };
+  }
+  const workflow = resolveWorkflowName(
+    run.workflow_name,
+    workflowEntries.map(ws => ws.workflow)
+  );
+  if (!workflow) {
+    const loadError = findWorkflowLoadError(loadErrors, run.workflow_name);
+    if (loadError) {
+      return {
+        ok: false,
+        message: `Workflow \`${run.workflow_name}\` failed to load: ${loadError.error}\n\nFix the YAML file and try again.`,
+      };
+    }
+    return {
+      ok: false,
+      message:
+        `Workflow \`${run.workflow_name}\` for run ${runId} was not found.\n\n` +
+        'Use /workflow list to check available workflows.',
+    };
+  }
+  return {
+    ok: true,
+    workflowName: workflow.name,
+    workflow: {
+      definition: workflow,
+      args: run.user_message,
+      resumeRunId: run.id,
+      resumeRun: run,
+    },
+  };
+}
+
+/**
+ * Attach the run continuation to an already-recorded gate decision.
+ *
+ * The decision is committed by the time this runs, so a continuation that cannot
+ * be resolved is reported as a follow-up step, never as a failed command — saying
+ * "failed" about a gate that IS resolved would send the user to re-approve a run
+ * that refuses a second decision.
+ */
+async function withRunContinuation(
+  runId: string,
+  workflowCwd: string,
+  headline: string,
+  action: 'approve' | 'reject'
+): Promise<CommandResult> {
+  let continuation: Awaited<ReturnType<typeof resolveRunContinuation>>;
+  try {
+    continuation = await resolveRunContinuation(runId, workflowCwd);
+  } catch (error) {
+    const err = error as Error;
+    getLog().warn(
+      { err, errorType: err.constructor.name, runId, action },
+      'cmd.workflow_gate_continuation_unresolved'
+    );
+    continuation = { ok: false, message: err.message };
+  }
+  if (!continuation.ok) {
+    return {
+      success: true,
+      message: `${headline}\nThe run could not be continued automatically: ${continuation.message}\nResume it with \`/workflow resume ${runId}\` once that is fixed.`,
+    };
+  }
+  return {
+    success: true,
+    message: `${headline}\nResuming \`${continuation.workflowName}\`...`,
+    workflow: continuation.workflow,
+  };
+}
+
 async function handleWorktreeCommand(
   conversation: Conversation,
   args: string[]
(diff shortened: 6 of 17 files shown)


Wirasm:
---
pr: 2577
base: dev
head: fix/2565-gate-message-no-auto-approve
reviewed: 2026-08-17T14:22:00Z
reviewed_head: 579cb61534795f84be67843b79abcb0819a84768
verdict: READY TO MERGE
open_findings: 0
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2577#issuecomment-5316922307
---



# PR Review: #2577 — fix(orchestrator): a plain message at an approval gate no longer approves it

## Signal

Both reviewers converged on one theme, and it is the PR's own load-bearing invariant: *whatever
resolves a gate must also continue the run*. Three exits violated it. They are not three instances
of one oversight, though — one was a genuine control-flow hole (a provider crash after a successful
tool call skipped the continuation), and two were **false promises about a continuation that is
impossible by design** (a container run cannot be resumed from chat; an unscoped conversation has
no resolution route at all). That distinction determined the fixes: the first needed the
continuation made unconditional, the other two needed the promise corrected. All three are fixed in
`478f7051`, along with a stale comment. No finding remains open, and the full gate plus CI are green
on the reviewed head.

- **Blocking:** 0
- **Non-blocking:** 0
- **Resolved since previous review:** Not applicable (first review)
- **Validation:** `bun run validate` exit 0 (140 test blocks, zero failures); all five CI checks pass.

## Findings

|---|---|---|---|
| `R1` | Important | A provider throw after a successful `manage_run` gate call skipped the continuation, leaving the run resolved and parked behind a generic error | FIXED |
| `R2` | Important | The gate prompt and the tool reply promised "the run continues" for a container run, which the executor provably refuses to resume from chat | FIXED |
| `R3` | Suggestion | A paused gate in a conversation with no project attached handed the agent instructions for two routes that are not wired that turn | FIXED |
| `R4` | Suggestion | Stale comment in the CLI approve path still described chat as deferring to the next user message | FIXED |

## Detailed Findings

<details>
<summary><code>R1</code> — continuation skipped when the agent turn throws after the gate is resolved</summary>

**Impact:** `approveWorkflow`/`rejectWorkflow` commit the resolution via their CAS the moment the
tool call returns. The continuation check then sat *after* the awaited `handleStreamMode` /
`handleBatchMode`, inside the outer `try` whose `catch` knows nothing about `gateResolution`. Any
throw from that await — a provider subprocess crash, a dropped stream, a non-retryable API error
arriving after the tool call — skipped the continuation. The user saw only
`classifyAndFormatError`: no mention of the recorded decision, no resume pointer. Exactly the
"resolved but nobody told" failure this PR exists to close.

**Evidence:** `packages/core/src/orchestrator/orchestrator-agent.ts` — `gateResolution` written by
the `onGateResolved` callback; the continuation guarded on `gateResolution.resolved !== null` after
the awaited mode dispatch; the outer `catch` never consulting it. Neither `handleStreamMode`'s nor
`handleBatchMode`'s `for await` wraps `aiClient.sendQuery`. `sendQuery` genuinely throws rather than
yielding an error result: `packages/providers/src/claude/provider.ts:1472-1474`
(`if (!shouldRetry || attempt >= MAX_SUBPROCESS_RETRIES) throw enrichedError;`). Not covered: every
`agentCallsManageRun` helper in the new test suite finished the stream cleanly.

The reviewer's framing is the useful one — the guarantee was not missing, it was *placed* so that it
held only when an unrelated call succeeded. `continueResolvedGateRun`'s "never throws" docblock buys
nothing when the function is never invoked.

**Required outcome:** the continuation runs once the resolution is observed, regardless of what the
rest of the turn does.

**Disposition:** FIXED in `478f7051`. The mode dispatch is wrapped so the continuation runs from a
`finally`, and `continueResolvedGateRun`'s whole body is now guarded (with a local `notify` helper
whose sends cannot throw) so it cannot replace the original error propagating out of the `finally`.
Regression test: *"a provider crash after the gate is resolved still continues the run"* — the
stubbed agent calls the tool, then throws mid-stream; `executeWorkflow` is still called and the
"Resuming" message still sent.

**Disagreement recorded:** the reviewer's addendum proposed closing all three findings with one
mechanism — "continuation is unconditional once resolution is observed, driven from a state read
back from the DB rather than in-memory control flow". Declined as the shape of the fix. Reading the
resolution back from the DB is #2008's resume-in-place, explicitly out of scope for this PR, and it
would not fix R2 or R3 at all: in both of those the continuation is *impossible*, not *skipped*, so
a more determined continuation driver would make them worse rather than better. The `finally` gives
exactly the "unconditional once observed" property at the scope where continuation is actually
possible.

</details>

<details>
<summary><code>R2</code> — "the run continues" promised for a container run that chat cannot resume</summary>

**Impact:** `formatPausedGateSection` and `signalGateResolved` closed with "Resolving the gate also
continues the run — there is no separate resume step" for every gate except `child_workflow`. For a
run that executed inside an isolation container that is false: `packages/workflows/src/executor.ts:1168-1189`
fails such a resume with a CLI pointer, because chat/web has no docker backend to rewire the
container with. The agent would tell the user the run was continuing, and the user would then
receive a contradicting `⚠️` and a failed run row.

**Evidence:** `ApprovalContext.type` is a four-member union
(`packages/workflows/src/schemas/workflow-run.ts:222`); only `child_workflow` was branched on. The
executor's own guard keys on `metadata.isolation === 'container'` — a field `PausedGateContext` did
not carry, so the function structurally could not make the right call. This is a missing-type
problem, not a missing-conditional one. Untested for container runs in the new suite.

**Required outcome:** no surface promises a continuation the executor will refuse.

**Disposition:** FIXED in `478f7051`. `isContainerRun()` is now exported from
`@archon/workflows/schemas/workflow-run` beside `isRunBlockedOnChild`, binding all consumers to the
same predicate the executor's guard uses. `PausedGateContext` gained `containerRun`, threaded from
the orchestrator; `signalGateResolved` skips the handover and returns the CLI pointer; and
`resolveRunContinuation` in `command-handler.ts` does the same for `/workflow approve|reject` (via a
`resumeHint` that replaces the generic `/workflow resume` line) so the two surfaces agree.

**Widened beyond the reported scope, deliberately:** the finding named the `writeback` gate type,
but the executor guard is keyed on the *run's* isolation, not the gate type — an ordinary
`approval:` gate inside a container run has the identical constraint. The fix keys on
`isContainerRun`, which is strictly more correct than branching on `type === 'writeback'`. Chose not
to schedule the continuation for container runs at all: letting it run would produce the same CLI
pointer, but only after flipping the run to `failed` to say what the reply now says for free.

Tests: *"does not promise continuation for a container run"* (prompt-builder), *"a container run is
never handed to the continuation — it points at the CLI"* (manage-run-tool), *"a container run is
resolved but not resumed from chat"* (orchestrator-agent), *"a container run is resolved but points
at the CLI instead of resuming"* (command-handler).

</details>

<details>
<summary><code>R3</code> — gate instructions handed out for routes that are not wired</summary>

**Impact:** the `pausedGateContext` computation had no `conversation.codebase_id !== null` gate,
unlike both routes it describes — the `manage_run` tool injection and the CLI-pointer section are
each gated on a scoped project. Reachable when a codebase is deleted while a run is paused
(`DELETE /api/codebases/:id` has no non-terminal-run guard; `conversations.codebase_id` is
`ON DELETE SET NULL`, `migrations/000_combined.sql:113`). The agent was told to resolve the gate
with verbs it did not have — a dead end, and a regression in *failure honesty* against the removed
natural-language path, which at least produced an explicit "no project is attached".

**Evidence:** the two gates in `packages/core/src/orchestrator/orchestrator-agent.ts` (CLI-pointer
section; `manage_run` injection) versus the ungated section computation in the same `handleMessage`.

**Required outcome:** the section must not instruct the agent to use an absent tool.

**Disposition:** FIXED in `478f7051`, with a correction to the proposed fix. The suggestion was to
suppress the section entirely when unscoped. That trades a dishonest instruction for silence — and
the old code did *not* fall through quietly, it attempted the approve and reported "no project is
attached". Instead `PausedGateContext` gained `agentCanResolve`: the gate facts are still relayed,
but the decision policy is replaced by "you have no way to resolve this yourself; tell the user to
use `/workflow approve|reject <id>`" — which is true and actionable, since those commands work
regardless of scope. Test: *"an unscoped conversation gets the gate but no instruction to resolve
it"* plus the pure-function case in `prompt-builder.test.ts`.

</details>

<details>
<summary><code>R4</code> — stale comment in the CLI approve path</summary>

**Impact:** none behavioral. `packages/cli/src/commands/workflow.ts` read "CLI auto-resumes after
approval (unlike chat, which defers to next user message)". This PR is what made the parenthetical
false — chat no longer defers to anything.

**Evidence:** the comment above the CLI's inline resume, versus this PR's `withRunContinuation` and
`continueResolvedGateRun`.

**Required outcome:** the comment stops describing behavior this PR removed.

**Disposition:** FIXED in `478f7051`. The reviewer judged it not worth a dedicated change; taken
anyway — it is one line, and the prose became false *because of this diff*, which is precisely the
kind of stale claim that outlives the change that invalidated it.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | `R1`, `R4` |
| seams | `R2`, `R3` |

Both reviewers separately verified and cleared: the pre-resolution `WorkflowRun` snapshot threaded
through `onGateResolved` (`hydrateResumableRun` re-reads fresh state and CASes independently, so the
stale object is safe); the one-per-turn decline and the reject-that-cancels skip in
`manage-run-tool.ts`; no double-resume race against web's `tryAutoResumeAfterGate`;
`withRunContinuation` correctly reporting `success: true` with a manual pointer rather than a false
failure on an already-resolved gate; `formatPausedGateSection` correctly omitting resolved gates and
redirecting `child_workflow`; and the accuracy of the CHANGELOG and docs updates, including the
pre-PR `/workflow approve` parked-run bug this PR also fixes.

Both also noted, without raising it as a finding, that "resolve gate then continue" now has four
independent implementations with no shared type — `command-handler.ts` and `orchestrator-agent.ts`
(new here), plus the pre-existing `tryAutoResumeAfterGate` and the CLI's inline logic. Not tracked
as a follow-up because #2008 already owns that consolidation, and the R2 fix binds the two most
divergent of them to one shared predicate (`isContainerRun`).

## Validation

|---|---|---|
| `bun run validate` | PASS | exit 0 on `478f7051`; 140 test blocks reported, none with a non-zero fail count. Covers type-check, `eslint --max-warnings 0`, `prettier --check`, the five generated-file freshness checks, and every package suite. |
| `bun test src/orchestrator/orchestrator-agent.test.ts` (`@archon/core`) | PASS | 234 pass / 0 fail |
| `bun test src/orchestrator/prompt-builder.test.ts` (`@archon/core`) | PASS | 26 pass / 0 fail |
| `bun test src/orchestrator/manage-run-tool.test.ts` (`@archon/core`) | PASS | 36 pass / 0 fail |
| `bun test src/handlers/command-handler.test.ts` (`@archon/core`) | PASS | 131 pass / 0 fail |
| CI: `test (ubuntu-latest)`, `test (windows-latest)`, `docker-build`, `schema-upgrade`, `Build docs site` | PASS — all five green on the current head `579cb615`, rebased onto `dev` `0f46d409`. |

**Base defect that briefly blocked CI — now fixed on `dev`, recorded because it shaped two
review cycles.** Between this PR's first green run and its rebase, `dev` HEAD `a7d908bf` began
failing its own CI, taking every open PR with it. #2570 added `scripts/node-ref-parity.test.ts`,
which read the `atomPattern` regex out of `packages/workflows/src/condition-evaluator.ts`; #2579
then moved that constant into the new `packages/workflows/src/when-atom.ts`. The guard was not
stale — it caught a real regression on its first outing: #2579 also taught the engine an
`$INPUTS.<name>` branch that the console builder's `ATOM_PATTERN` did not have, so
`when: "$INPUTS.mode == 'fast'"` was engine-accepted and builder-rejected. #2591 fixed both halves
(the guard now imports `WHEN_ATOM_PATTERN` and compares compiled `.source` instead of scraping
source text, and the builder learned `$INPUTS`), and #2592 followed up. This PR's diff touched none
of the four files involved and was never adjusted to accommodate the breakage.

**Not verified:** no live end-to-end run against a real provider. The agent's *judgement* on
borderline phrasings is not unit-testable — the tests prove the wiring (a resolution reaches the
gate operation, and a resolution always either schedules the continuation or says in words why it
cannot). A maintainer smoke test on one chat surface before release is worth doing, and the
user-visible UX trade needs the maintainer's sign-off independent of this review.

## Verdict

**READY TO MERGE**

All four findings are fixed on `478f7051` with regression tests for each, and the full gate is
green. Two notes for the merge decision rather than the code: this is a user-visible breaking change
whose product trade — a typed approval now goes through the agent's judgement instead of a
hard-coded branch — is the maintainer's call, not this review's; and if #2529 merges, its Slack-side
re-dispatch should be deleted rather than layered over the general continuation this PR lands.

Issue #2563, opened 2026-08-17
feat(workflows): loops can only terminate on a prose sentinel — no structured completion channel, and no way to opt out


Wirasm:
## Problem

A loop node's completion is `signalDetected || bashComplete` (`dag-executor.ts:4939`, `:3730`),
where `signalDetected` is a regex over the model's prose (`detectCompletionSignal`,
`executor-shared.ts:695-713`). Three properties compose into a gap:

1. `until:` is **required** on every loop — `z.string().min(1)` on `loopControlSchema`
   (`schemas/loop.ts:22`), shared by `loop:` and `loop_group:`. `until_bash` is optional.
2. The sentinel alone completes the loop. A deterministic `until_bash` can never veto it.
3. `output_format` is unsupported on loop nodes (`loop-nodes.md:436`) — dropped at parse time
   with a loader warning; `executeLoopNode` never requests it from the provider.

Consequences:

- **A deterministic loop cannot opt out of prose matching.** An author whose completion condition
  is fully checkable (`until_bash`) must still invent a sentinel, which stays live and can
  false-positive if the model emits that string while reasoning about the criteria.
- **A judgment-carrying loop has no validated channel.** Every other node type can express
  "the model decided X" through `output_format` + schema validation. Loops cannot.
- **The two variants behave differently.** `loop_group` skips `until_bash` entirely when the
  signal fired (`:3652`); `loop:` runs both and ORs them (`:4858`, `:4939`).

Who hits it: every workflow author writing an iterative node — and imminently the bundled-defaults
rewrite, where several loops (`archon-ralph-dag.yaml:658`, `archon-piv-loop.yaml:534`,
`archon-test-loop-dag.yaml:40`) track real completion state on disk yet terminate only on
`<promise>COMPLETE</promise>`. `archon-adversarial-dev.yaml:299-303` is the one that pairs the
sentinel with an authoritative `until_bash`.

## Proposed Solution

Two separable parts.

**Part A — make the sentinel optional when a deterministic check exists.** Relax
`loopControlSchema` to require *at least one* of `until` / `until_bash` (superRefine), instead of
`until` unconditionally. A loop with only `until_bash` then has no prose path at all. Additive and
backward compatible: every existing workflow declares `until`.

**Part B — a structured completion channel.** Support `output_format` on loop nodes and let a
declared boolean field drive termination, so "the model judged this done" becomes a validated
field rather than a string match.

Part B is a new YAML surface feature and owes the constitution's admissibility test:
[...]

PR #2579, opened 2026-08-17
feat(workflows): a when: that compares a whole AI reply to a literal now fails at load

### LINKED ISSUE #2566: feat(workflows): reject a when: that compares a whole AI output to a literal, at load time
## Problem

A `when:` condition can compare a node's **entire free-form AI output** against a literal string, and
when that comparison fails the step is skipped silently.

`condition-evaluator.ts:73-77` (`resolveOutputRef`): for `$nodeId.output` with no `.field`, the raw
text is returned as-is. `evaluateAtom` (`:190`) then does `actual === expected`. So:

```yaml
- id: decide
  when: "$analyze.output == 'BUG'"     # $analyze is a prompt: node with no output_format
```

branches on whether the model's **whole reply** is byte-identical to `BUG`. The model writes
`"This is a BUG."` → comparison false → the node is skipped, no error is raised, and the run reaches
a terminal state looking successful while having quietly done less than the author asked for.

The behaviour is documented as intentional in a code comment (`condition-evaluator.ts:40-48`), and no
bundled workflow relies on it — but nothing prevents an author from wiring it up, and the failure is
invisible when they do.

**Why it matters now:** the `AGENTS.md` principle added in #2564 tells authors to route decisions
through structured channels. This is the one `when:` path that silently rewards doing the opposite,
and the defaults rewrite will produce a lot of new `when:` conditions.

**Scope note — this must not be a blanket restriction.** Whole-output equality against a `bash:` or
`script:` producer is legitimate and safe: that stdout is author-controlled and exact by construction.
`condition-evaluator.test.ts:501-506` exercises exactly that shape. The hazard is specific to
`prompt:` / `loop:` / `loop_group:` producers that declare no `output_format`.

## Proposed Solution

Reject it at **load time**, scoped to AI-producing node types.

A `when:` atom that compares a bare `$node.output` (no `.field`) against a literal, where the producer
is a `prompt:` / `loop:` / `loop_group:` node with no `output_format`, fails the workflow at load with
a message naming the fix — declare `output_format` and compare a field.

Load-time is possible today and cheap:

- `validateDagStructure` (`loader.ts:380`+) already holds the full `DagNode[]`, including each
  producer's node type and static `output_format`. Both are load-time visible.
- It already resolves `refNodeId` for existence checks (`:468`, `:490`); this is the same lookup.
- Precedent for exactly this shape of check lives in the same function: `fan_out.items`
  transitive-dependency validation (`:517-529`) is a load-time structural check that could have been
  left to runtime fail-closed, and wasn't — because catching it before spend is strictly better.

**One implementation constraint.** `WHEN_REF_SOURCE` (`loader.ts:119`) captures only the first segment
after the node id, so it cannot distinguish `$node.output` from `$node.output.field` — both yield
`output`. Implementing this precisely needs the atom-level decomposition that `evaluateAtom` already
performs (`condition-evaluator.ts:150-172`), which is not currently exported. **Extract a shared
"parse when atom" helper that both `loader.ts` and `condition-evaluator.ts` call** — do not add a
second regex, which would drift from the runtime one. Compare the existing KEEP-IN-SYNC block at
`loader.ts:406-413`, which exists because that drift already happened once for the ref-scanning
surfaces.

### Before (current)

```
Author writes:  when: "$analyze.output == 'BUG'"   ($analyze has no output_format)

  load   ──▶ accepted                                                      [!]
  run    ──▶ model outputs "This is a BUG."
             "This is a BUG." === "BUG"  →  false                          [!]
             node skipped, no event, no warning                            [!]
  result ──▶ run completes "successfully", having skipped the branch       [!]
```

### After (proposed)

```
Author writes:  when: "$analyze.output == 'BUG'"   ($analyze has no output_format)

  load   ──▶ REJECTED                                                      [+]
             "when: on node 'decide' compares the whole output of AI node
              'analyze' to a literal. Declare output_format on 'analyze'
              and compare a field (e.g. $analyze.output.status == 'BUG')."

Unchanged: a bash:/script: producer keeps whole-output equality.           [+]
```

## Alternatives Considered

|---|---|---|---|
| Load-time rejection, AI producers only (proposed) | Caught before any spend; author gets the fix in the message; bash/script unaffected | Needs the shared atom parser extracted first | Chosen |
| Leave it, document the hazard | Zero work; the code comment already exists | A comment is not a guard, and the author who hits this is by definition someone who didn't read it | Documentation isn't enforcement |
| Fail at runtime | No loader change | Burns a full model turn before failing; contradicts the constitution's own "loud beats silent" posture on unresolvable refs | Strictly worse than load-time for the same outcome |
| Remove whole-output comparison entirely | Simplest rule | Breaks the legitimate bash/script case the tests document | Hazard is producer-type-specific, not operator-specific |

## Scope

- Package(s) likely affected: `workflows` (`loader.ts`, `condition-evaluator.ts`, shared atom
  parser), `docs`
- Breaking change? Technically yes for any workflow relying on this today — **none exist in the
  bundled set**; a user workflow doing it is already silently broken.

## Security Considerations

## Constitutional note

This does **not** conflict with *"`when:` never grows incrementally"*
(`reference/workflow-language-constitution.md:100-106`, smell #1 *Expression creep*). That rule
governs adding operators, functions and arithmetic — i.e. new ways to **compute** inside `when:`.
Restricting which references are legal for an existing `==` operator is a validation-strictness
change, the same axis as the `not-in-schema` and unknown-node rejections that already ship.

File it under smell **#5 *Implicit magic*** (`:134-140`), whose lever is precisely *loud, documented,
defeatable* — which is what this change does to a behaviour that is currently silent, undocumented in
the authoring guide, and defeatable only by knowing to avoid it.

## Definition of Done

- [ ] A shared when-atom parser is extracted and used by both `loader.ts` and `condition-evaluator.ts`
      (no second regex)
- [ ] Load fails when a `when:` compares a bare `$node.output` to a literal and the producer is an AI
      node with no `output_format`; the message names the fix
- [ ] `bash:` / `script:` producers are unaffected — existing tests at
      `condition-evaluator.test.ts:501-506` still pass
- [ ] A producer that *does* declare `output_format` is unaffected
- [ ] Tests cover: AI producer without `output_format` (rejected), AI producer with it (accepted),
      script producer (accepted), `.field` access (accepted)
- [ ] Authoring guide documents the restriction and the correct pattern

## Related

- `AGENTS.md` → **Natural Language Is Not a Wire Format** (added in #2564) — this is the `when:`-side
  instance of the same failure.

Wirasm:
## Problem and outcome

A `when:` could compare a node's **entire free-form AI reply** to a literal
(`when: "$analyze.output == 'BUG'"`). The model writes `This is a BUG.`, the byte-for-byte
comparison is false, the node is skipped with no event and no warning, and the run reaches a
terminal state looking successful while having quietly done less than the author asked for.
Separately, a `workflow:` sub-run child could **read** `$INPUTS.mode` in a prompt but could not
**branch** on it — `when:` was handed to `evaluateCondition` with no input scope, so the ref
parsed as a node called `INPUTS` and failed the node.

- **Outcome:** the first shape is a **load error** naming the fix, before any spend; the second
  now works, and an `$INPUTS.<name>` the run does not carry fails the node loudly instead of
  resolving to `''`.
- **Invariant:** `bash:`/`script:` producers keep whole-output equality — their stdout is
  author-controlled and exact by construction. A field access is always allowed. An input value
  is compared as **data** and is never spliced into the expression, so it can never change what
  a condition means.
- **Scope boundary:** only defect 1 of #2453 is folded in. Defects 2 (`output_format` on a
  `workflow:` node never validated against the child) and 3 (fan-out aggregates double-encoded)
  stay with #2453, along with all of its design work. No new `when:` operator, function, or
  grammar capability is added.
- **Root cause:** `resolveOutputRef` returns a producer's raw text for a bare `$node.output`
  and `evaluateAtom` then does `actual === expected` — correct for a shell producer, silently
  wrong for a model. And `condition-evaluator.ts` had no `INPUTS` handling at all, so its atom
  regex read `$INPUTS.mode` as node `INPUTS` / field `mode`.

## Review guidance

- **Feedback requested:** correctness of the load-time scoping (which producers are rejected),
  and the design choice in *Solution → why an evaluator scope, not a substitution pass*.
- **Start here:** `packages/workflows/src/when-atom.ts` — the shared atom parser both `loader.ts`
  and `condition-evaluator.ts` now call. It is what makes the load-time check possible:
  `WHEN_REF_SOURCE` in `loader.ts` captures only the first path segment, so it cannot tell
  `$n.output` from `$n.output.field`.
- **Review order:** `when-atom.ts` → `loader.ts` (`freeFormAiProducerKind` + the new block at the
  end of the `when:` handling in `validateDagStructure`) → `condition-evaluator.ts`
  (`InputRefError`, `resolveInputRef`, `evaluateAtom` now consuming the shared parser) →
  `dag-executor.ts` (one call site) → tests → docs.
- **Lower-attention areas:** the `ids: Set<string>` → `nodesById: Map<string, DagNode>` rename
  inside `validateDagStructure` (and the matching `enclosingIds` → `enclosingNodes`) is
  mechanical — the producer's *node*, not just its id, is what the new check needs, including
  for a `loop_group` body referencing an outer node. No external caller passes the second
  argument.
- **Known risk or uncertainty:** this is technically breaking for a user workflow that compares
  a whole AI output today. Such a workflow is already silently broken, no bundled workflow does
  it (evidence in *Validation*), and the load error names the fix.

## Solution

**One shared atom parser.** `when-atom.ts` owns the grammar (`parseWhenAtom`, `whenAtoms`,
`splitOutsideQuotes`) and both consumers call it: the evaluator to run an atom, the loader to
inspect one. The atom regex is built from `INPUT_NAME_SOURCE` and a single operator list, so the
type and the grammar cannot drift. This is the constraint #2566 puts first — the `KEEP IN SYNC`
comment already in `loader.ts` exists because that drift happened once before, and defect 1 gives
the parser its second consumer.

**Load-time rejection, scoped by what the producer can actually promise.**
`freeFormAiProducerKind()` answers *why* a producer's whole output cannot be compared, and the
message differs accordingly:

- `prompt:` / `command:` → "declare `output_format` and compare a field". Declaring one is also
  the opt-out: the output is then a JSON document, not prose.
- `loop:` → **no opt-out**: `output_format` is *dropped at parse*. It lives in the schema's
  `aiOnly` group and the `LoopNode` branch of the transform does not spread `aiOnly` (it rescues
  only `pi`), so the field never reaches the node. Declaring one is a no-op, not an escape.
- `loop_group:` → **no opt-out, different reason**: the field *does* survive the transform here
  (the `LoopGroupNode` branch spreads `aiOnly`), but the group's completion returns
  `output: lastIterationOutput` with no `structuredOutput` and no `declaredFields` — so a schema
  can never make `$group.output` the JSON document.

These two are stricter than #2566's text, and stricter *because* following it literally would
have handed authors a fake escape hatch. The distinction was established by parsing, not reading:
`dagNodeSchema.safeParse()` with `output_format` declared keeps it for `prompt:`, `command:` and
`loop_group:`, and drops it for `loop:`. The tests pin the two distinct messages, so the
distinction cannot silently collapse back into one.

Two deliberate calls beyond the three node types #2566 enumerates:

- **`command:` is included.** A command file is an inline prompt that lives in another file. Left
  out, the hazard stays reachable by moving the prompt. Zero bundled workflows are affected.
- **All operators, not just `==`/`!=`.** `$ai.output > '80'` against free-form prose is the same
  mistake; an AI node that genuinely emits a comparable number declares `output_format`, which
  already exempts it.

**Why an evaluator scope for `$INPUTS`, not a substitution pass.** `include:` solves this by
rewriting `when:` text at load time, but a sub-run's inputs are only known at run time, so the
same trick cannot apply. Substituting into the condition string was rejected on three grounds:

1. **It is not even expressible.** Substitution is verbatim, so `when: "$INPUTS.mode == 'fast'"`
   becomes `fast == 'fast'` — not an atom at all. This is not a hypothetical: it is exactly what
   `include:` produces today — verified against `dev`, and filed as #2580.
2. **A value would become grammar.** An input containing `'`, `&&`, or `==` would rewrite the
   condition or make it unparseable — and unparseable fails closed to a *skip*, the same
   silent-branch failure the other half of this PR removes. Resolving the reference keeps the
   value as data.
3. **It keeps one strictness contract.** An unknown `$INPUTS.<name>` throws `InputRefError` and
   fails the node, exactly like `not-in-schema` on `$node.output.field` — instead of `''`.

The cost is that the evaluator now knows a second namespace. That is paid down by the parser:
`WhenAtomRef` is a discriminated union (`{kind:'node'} | {kind:'input'}`), so the second scope is
a typed variant resolved in one place, not a special case sprinkled through evaluation. As a
side effect the bare `$INPUTS.output` form — which used to resolve to a coincidental `''` — now
means "the input named `output`" and fails if there isn't one.

## Behavior change

| | Before | After |
|---|---|---|
| **`when: "$aiNode.output == 'BUG'"`** | Loads; at run time compares the model's whole reply, is false, node skipped with no event | **Load error** naming the producer, the literal, and the fix |
| **`when: "$bashNode.output == 'true'"`** | Works | Unchanged |
| **`when: "$INPUTS.mode == 'fast'"` in a sub-run child** | `OutputRefError('unknown-node')` — the node fails | Branches on the caller's `with:` value |
| **`when: "$INPUTS.output == 'x'"`** | Resolved to `''` (coincidence of the ref grammar) | Resolves the input named `output`; throws if undeclared |
| **Failure behavior** | Wrong branch taken silently; run reports success | Load fails before spend, or the node fails with a message naming the fix |

## Validation

- `bun run validate` — **exit 0**, all steps green (type-check, lint, format:check, the five
  generated-file checks, and per-package tests).
- `bun test src/loader.test.ts` in `@archon/workflows` — 266 pass / 0 fail. 16 new cases:
  rejection for `prompt:`/`command:`/`loop:`/`loop_group:` producers, under a numeric operator,
  inside a compound expression, from a `loop_group` body referencing an enclosing node, and
  **after include expansion** (the block's AI sink is only visible post-flattening, so only the
  re-validation can catch it); acceptance for `bash:`, `script:`, an AI producer that declares
  `output_format`, a field access, a `$INPUTS` ref, and an unparseable `when:`.
- `bun test src/condition-evaluator.test.ts` — 92 pass / 0 fail, including the shared-parser
  suite and nine `$INPUTS` evaluation cases.
- `bun test src/schemas.test.ts …` (its sanctioned split) — 205 pass / 0 fail, including a new
  `dagNodeSchema — output_format survival by node type` block. It is the only thing in the
  repository holding the field-survival asymmetry that the loader's messages, the authoring
  guide and the constitution row all assert — it fails in the same commit that would make
  them wrong. The `loop_group` half of that claim is not yet pinned; see #2585.
- `bun test src/subrun.test.ts` — 66 pass / 0 fail, including two new end-to-end cases: a child
  whose `when: "$INPUTS.mode == 'fast'"` selects the fast branch from a parent's `with:`, and a
  misspelled input that fails the node with
  `Unknown input '$INPUTS.mdoe'. Did you mean $INPUTS.mode?` delivered to the platform.
- **No workflow in the tree is affected — loaded, not inferred.** `parseWorkflow` over all 63
  YAMLs in Archon's own `.archon/workflows/` produces **zero** errors from the new rule. (16
  files do fail that probe, all with `Unknown provider` for `pi`/`opencode`/`copilot`, because a
  standalone probe registering only the two built-ins cannot resolve them — unrelated to this
  change.) Independently, each of the four whole-output `when:` comparisons in the tree
  (`archon-release`, `archon-smart-pr-review`, `archon-create-issue`, `e2e-joins`) was read and
  has a `bash:` producer. Note the existing "no false positives on the real workflow corpus"
  suite could **not** have caught a regression here: it `continue`s past any file that fails to
  parse, so it only checks warnings, never load errors.
- Two pre-existing loader tests whose fixtures used the now-rejected shape were changed to a
  field comparison. Their subject is ref validation and the `lastIndex` reset across multiple
  ref surfaces, both of which a `.field` ref exercises identically.
- **Not verified:** nothing material for the changed behavior. One adjacent finding was
  discovered while verifying the design claim and is documented rather than fixed: under
  `include:`, `$INPUTS` on the LEFT of a `when:` splices verbatim into an unparseable
  condition. Filed as #2580; the authoring guide now tells include authors to use the
  right-hand form.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Breaking for a user workflow that compares a whole AI output; that workflow is already silently taking the wrong branch. The load error names the fix. No bundled workflow affected. | Corpus scan above |
| Documentation / communication | `when:` syntax section gains the restriction and the correct pattern; a new section documents `$INPUTS` in `when:`; the constitution gains two case-law rows and an applied case under smell #5 *Implicit magic*. Three incidental doc examples that taught the rejected shape were corrected. | `guides/authoring-workflows.md`, `book/dag-workflows.md`, `guides/approval-nodes.md`, `guides/script-nodes.md`, `reference/workflow-language-constitution.md` |
| Rollout / rollback | Self-contained; reverting restores the previous accept-and-skip behavior with no data or schema implications. | — |

## A correction in the commit history, stated plainly

The reasoning behind the loop rejection was wrong twice before it was right, and both wrong
versions were published in commit messages. Recorded here so the history is not read as
authoritative:

| Commit | Claim | Status |
|---|---|---|
| `02ef1fd7` | "`output_format` is an ignored field on loop nodes" | Half right — true of the loader's warning list, but it conflated `loop:` with `loop_group:` and never checked the transform |
| `5a282bde` | "a loop node DOES pass `output_format` to the provider and stores the resulting `structuredOutput`" | **False.** It traced the shared executor path and stopped short of the parse layer |
| `1701ddaa` | `loop:` drops the field at parse; `loop_group:` keeps it but returns iteration text | Verified by parsing each node type through `dagNodeSchema` |

The *verdict* — neither loop form has a whole-output opt-out — was correct throughout and never
moved; only the reason did. No classification logic changed across the three commits.

## Constitutional note

Per `AGENTS.md`, cited for a change to the YAML surface. This is **not** smell #1 *Expression
creep* — no operator, function, or arithmetic is added; it narrows which *references* are legal
for an operator that already exists, the same axis as the shipped `not-in-schema` and
unknown-node rejections. It is filed under smell **#5 *Implicit magic***, whose lever is *loud,
documented, defeatable*: the behavior was silent, undocumented, and avoidable only by knowing to
avoid it. `$INPUTS` in `when:` adds a second reference **scope** to an existing grammar, not a
new way to compute. Both are recorded in the case-law table.

## Links

- Closes #2566
- Refs #2453 (defect 1 only; the issue keeps defects 2 and 3 and all of its design scope)
- Related #2580 (adjacent `include:` finding, filed from this work)
- Related #2585 (the loop_group half of the rejection message is asserted in prose but pinned by
  no runtime test; best attached to whatever next changes loop_group completion shape)


Wirasm:
---
pr: 2579
base: dev
head: fix/2566-when-whole-ai-output-literal
reviewed: 2026-08-17T15:09:53Z
reviewed_head: ae0f8cc8dcbe20b80b1ac9cd275d868251b00f13
verdict: READY TO MERGE
open_findings: 1
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2579#issuecomment-5317052290
---



# PR Review: #2579 — feat(workflows): a when: that compares a whole AI reply to a literal now fails at load

## Signal

Final re-review at `ae0f8cc8`. **All five prior findings are FIXED.** `bun run validate` is exit 0 and
the engine change has now survived four heads and three reviewer passes with no reachable behavioral
defect. **READY TO MERGE** on substance — with one mechanical caveat below.

The interesting part of this round is `R5`. The last review argued that the loop/`loop_group` claim
was asserted in five prose places and derived from none, and that one schema-level test would close
it. That test now exists (`schemas.test.ts`, `describe('dagNodeSchema — output_format survival by
node type')`), it runs unconditionally in the package's CI split, and it observes the transform's
output directly — so it flips the moment the `loop:` branch starts spreading `aiOnly`. The seam
analyzer then made the sharper observation I had missed when specifying the fix: the `loop_group`
message is a **two-part** claim — the field survives parse, *and* the group's completion still
returns raw text — and the new test pins only the first part. That residual is recorded as `R6`, a
non-blocking Suggestion. It is latent, not present: every message shipped here is true today.

Worth noting what the author did with the aside I explicitly declined to raise. The seam analyzer and
I both judged the `if`/`if`/fallthrough dispatch not a seam; ws1 added `return kind satisfies never;`
anyway, on the argument that "a new kind silently inherits the last branch's remedy" is the exact
failure class this PR exists to remove. That is the right call, and it is verified rather than
assumed: compiling a replica of the dispatch with a fourth kind produces `TS1360: Type '"NEW-KIND"'
does not satisfy the expected type 'never'`.

**The one caveat, and it is not a finding.** `test (windows-latest)` is red on this head. It is not
PR-caused and the evidence is decisive: the job has exactly one step, `Set up job`, and it failed
downloading the `oven-sh/setup-bun` action — codeload returned `503` then `429 (Too Many Requests)`
after three attempts. The runner never checked out the code and never ran a test. The same job passed
at 7m56s on the immediately preceding head, whose delta from this one is a test file plus two comment
edits. It needs a re-run to turn the check green before merge; that is ws1's action, not something I
should trigger.

- **Blocking:** 0
- **Non-blocking:** 1
- **Resolved since previous review:** 2 (`R4`, `R5`); 5 of 6 findings now FIXED overall
- **Validation:** `bun run validate` exit 0 at `ae0f8cc8`; CI 4/5 green, `test (windows-latest)` red on an infrastructure failure proven external

## Findings

|---|---|---|---|
| `R3` | Important | `5a282bde` claimed a `loop:` node's `output_format` reaches the provider; the parser drops it | FIXED |
| `R1` | Suggestion | `WHEN_INPUTS_SCOPE` not adopted at `loader.ts`'s two `'INPUTS'` literals | FIXED |
| `R2` | Suggestion | Guide implied a node may be named `INPUTS`, which `dagNodeSchema` rejects | FIXED |
| `R4` | Suggestion | The `WHEN_INPUTS_SCOPE` docblock called itself "the one spelling of it" while other sites hand-type the word | FIXED |
| `R5` | Suggestion | No test pinned the schema-transform fact the error messages and docs assert | FIXED |
| `R6` | Suggestion | The `loop_group` message's second half — that the group's completion returns raw text — is still pinned by nothing | OPEN |

## Detailed Findings

<details>
<summary><code>R6</code> — the <code>loop_group</code> message's runtime half is unpinned</summary>

**Impact:** The `loop_group` rejection message makes two claims: (a) `output_format` survives parsing
on a `loop_group:` node, and (b) the group's completion nonetheless returns the last iteration's raw
text, so surviving parse changes nothing. `R5`'s new test pins (a) only. If a future change taught the
loop_group completion path to synthesize `structuredOutput` from `node.output_format` — mirroring what
prompt/command nodes already do — claim (b) would become false, the shipped message would mislead, and
the schema test would keep passing, because it never touches `dag-executor.ts`. This is the same
untethered-prose shape as `R5`, one layer down.

Nothing is wrong today. I verified claim (b) holds at this head: `lastIterationStructuredOutput` exists
only in `executeLoopNode` (`dag-executor.ts:4310`, `:4574`, `:5077-5078`) and the loop_group executor
has no such variable at all, with every terminal return `output: lastIterationOutput` bare;
`declaredFields` is set only at `:2468` (prompt/command), `:5556` (`workflow:`) and `:8192` (resume).

**Evidence:** `packages/workflows/src/loader.ts:589-593` (the two-part message) against
`packages/workflows/src/dag-executor.ts:3939` / `:5174`. The seam analyzer confirmed no existing test
combines a `loop_group` node with `output_format` and asserts the runtime output stays raw — `grep`
finds zero `output_format` occurrences near any `loop_group` test in `dag-executor.test.ts`. By
contrast the `schema-capable` message's runtime half **is** independently pinned, by
`dag-executor.test.ts:2567` ("uses structuredOutput from result when output_format is set").

**Required outcome:** One companion assertion in `dag-executor.test.ts` — a `loop_group` node with
`output_format` set still produces plain-text `output` with no `structuredOutput` and no
`declaredFields`. Best attached to whatever next touches loop_group completion shape (#2563-adjacent),
not necessarily here.

**Disposition:** OPEN — non-blocking. Latent only: every message this PR ships is true at this head,
and closing it requires a test in a file this PR does not otherwise touch. Recorded rather than
tracked because it is small enough to ride along with the next change in that area; note the `R5` test's
own docblock names the `loop:`-branch risk but not this one, so this is the unstated second gap.

</details>

<details>
<summary><code>R5</code> — nothing held the loop claim to the schema — FIXED</summary>

**Impact (as filed):** The `output_format` survival asymmetry was asserted in five prose places —
the `freeFormAiProducerKind` docblock, two load-error messages, the authoring guide, the constitution
row — and derived from none. `freeFormAiProducerKind` dispatches on node type alone, so a future
`aiOnly` spread in the `LoopNode` branch would silently falsify a shipped message with nothing failing.

**Disposition:** **FIXED** at `ae0f8cc8`. `schemas.test.ts:938-987` adds
`describe('dagNodeSchema — output_format survival by node type')` immediately after the
`LOOP_GROUP_NODE_AI_FIELDS` block, with `loop:` absent, `loop_group:` present (deep-equal to the
declared schema), and `prompt:`/`command:` present.

Verified rather than accepted:
- Ran it: `bun test src/schemas.test.ts -t "output_format survival"` → 3 pass, 8 expect() calls.
- Ran the sanctioned split it lives in → 205 pass / 0 fail (was 202).
- Confirmed `schemas.test.ts` is unconditionally in `packages/workflows/package.json`'s `test` script
  with no filter, so CI runs all three cases.
- Confirmed it is not vacuous: `parsedNode` asserts `expect(result.success).toBe(true)` before touching
  `result.data`, so a parse failure fails the test rather than silently yielding `undefined`. The
  assertions observe the transform's own output with no indirection, so a change to which branch
  spreads `aiOnly` flips them directly.

The test's docblock states why it exists and names #2563 as the change that would trip it. See `R6`
for the half it does not cover.

</details>

<details>
<summary><code>R4</code> — docblock claimed a single spelling that isn't single — FIXED</summary>

**Impact (as filed):** `when-atom.ts` described `WHEN_INPUTS_SCOPE` as "the one spelling of it" while
`include-expander.ts:560` carried an identical hand-typed check and two regex sources hardcode the word.

**Disposition:** **FIXED** at `ae0f8cc8` by softening rather than by widening the diff — the docblock
now says `loader.ts` imports it rather than retyping it, states plainly that it is NOT the only
spelling, and names `include-expander.ts`, `executor-shared.ts`, and the `dagNodeSchema` cycle
constraint. All three claims verified: `include-expander.ts:111`, `:560`, `executor-shared.ts:27`, and
`schemas/dag-node.ts:771` all still hardcode the word, and `when-atom.ts:24` imports
`INPUT_NAME_SOURCE` *from* `schemas/dag-node.ts`, so the reverse import would indeed be circular.

The seam analyzer independently checked the "pre-existing on `dev`" premise by history rather than
taking it on trust — all four sites trace to `ed4f6acd`, `14415a54` and `84688f42`, confirmed ancestors
of `dev` — and judged the trade correct: this is the carve-out case, honest documentation of a
duplication not worth collapsing now, and widening into a sibling workstream's file for a wording nit
would carry real conflict risk for near-zero benefit. Agreed.

</details>

<details>
<summary><code>R3</code> — the loop rejection's reason was false — FIXED</summary>

**Impact (as filed):** `5a282bde` told authors that declaring `output_format` on a loop node still
populates `$loop.output.field`. `output_format` is in the schema's `aiOnly` group and the `loop:`
transform branch does not spread it, so the field never reaches the node object.

**Disposition:** **FIXED** at `1701ddaa` and unchanged since. The two loop forms are now separated in
all five places, with `freeFormAiProducerKind` returning `'schema-capable' | 'loop' | 'loop-group' | null`
and three distinct messages. Verified by re-running the parse probe (`loop: false`, `loop_group: true`,
`prompt: true`) and by tracing every terminal return in the loop_group executor. As of `ae0f8cc8` the
schema half of this claim is also pinned by a test — see `R5`.

**Found by:** head-delta verification, corrected on a challenge from `team-lead`

</details>

<details>
<summary><code>R1</code> — reserved-scope constant not adopted in <code>loader.ts</code> — FIXED</summary>

**Disposition:** **FIXED** at `1701ddaa`. `loader.ts:57` imports `WHEN_INPUTS_SCOPE`; both ref scans
(`:532`, `:554`) compare against it; no `'INPUTS'` literal remains in that file. See `R4` for the
sibling sites, deliberately left alone.

</details>

<details>
<summary><code>R2</code> — guide implied a node can be named <code>INPUTS</code> — FIXED</summary>

**Disposition:** **FIXED** at `1701ddaa`. The guide now states that a node cannot be given that id and
that the loader rejects it, matching `dagNodeSchema.superRefine` (`schemas/dag-node.ts:771-777`);
verified by parse. The `when-atom.ts` docblock and the `parseWhenAtom` comment were corrected in the
same pass.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | No findings on the `ae0f8cc8` delta. Verified by execution: the new test passes and is non-vacuous (`parsedNode` asserts `result.success` before use), `schemas.test.ts` is unfiltered in the package's CI split, `satisfies never` type-checks and is unreachable given the callee's declared return type, and all three revised docblock claims hold. Earlier rounds: no findings on `02ef1fd7` or the `1701ddaa` delta; contributed `R2` on `5a282bde` |
| seams | `R6`, plus confirmation that `R4` and `R5` close. Checked the "pre-existing on `dev`" premise by commit history rather than accepting it, and confirmed the `satisfies never` guard replaces a genuine unconditional fallback. Earlier rounds: `R1`, then `R4` and `R5` |

## Validation

|---|---|---|
| `bun run validate` (repo root, at `ae0f8cc8`) | PASS | Exit 0. 5 generated-file checks, `type-check` across all 12 packages, `eslint --max-warnings 0`, `prettier --check`, `test-install`, per-package tests. No non-zero `N fail` line in the 133 KB log. |
| `bun test src/validation-parser.test.ts src/schemas.test.ts src/command-validation.test.ts src/packaged-workflow.test.ts` | PASS | 205 pass / 0 fail (was 202) — the split carrying the new `R5` test |
| `bun test src/schemas.test.ts -t "output_format survival"` | PASS | 3 pass / 0 fail, 8 expect() calls — the new test executes rather than being filtered |
| `tsc --strict` on a replica of the dispatch with a fourth kind added | PASS (guard proven) | `TS1360: Type '"NEW-KIND"' does not satisfy the expected type 'never'` — the exhaustiveness guard genuinely converts a new kind into a compile error |
| GitHub CI on `ae0f8cc8` (run 32040908276 + docs run) | 4 PASS / 1 FAIL (external) | `test (ubuntu-latest)` 3m8s, `docker-build` 4m3s, `schema-upgrade` 1m17s, `Build docs site` 47s all green. `test (windows-latest)` failed at 57s in its **only** step, `Set up job`: codeload returned `503` then `429 (Too Many Requests)` downloading `oven-sh/setup-bun` after 3 attempts. No checkout, no test run. The same job passed at 7m56s on `1701ddaa`. Needs a re-run to go green. |

Carried forward, still true: running `loader.test.ts`, `condition-evaluator.test.ts` and
`subrun.test.ts` in a **single** `bun test` process yields 3 failures from the documented
`mock.module()` pollution — not a PR signal; the package runs them in separate invocations and this PR
adds no `mock.module()` call. Repository conventions re-checked at this head: no AI attribution in any
of the four commit bodies or the PR description; the constitution is cited and its case-law row
updated; `dag-executor.ts` still carries only the 3-line `resolveRunInputs` argument change, and
`schemas/loop.ts`, `executor-shared.ts` and the package `test` script remain untouched.

## Verdict

**READY TO MERGE**

No OPEN Critical or Important findings; five of six findings are fixed and independently re-verified,
and `bun run validate` is green. One mechanical step remains that is not a code change: re-run
`test (windows-latest)`, which failed before checkout on a GitHub action-download rate limit. `R6` is
optional and best carried by the next change to loop_group completion.

PR #2581, opened 2026-08-17
refactor(workflows): effort: sets reasoning depth on every provider that has one

### LINKED ISSUE #2556: refactor(workflows): one spelling for reasoning depth — effort: works on every provider, modelReasoningEffort: leaves the workflow schema

Wirasm:
## Problem and outcome

The workflow language had two workflow-level ways to say "reason harder": cross-provider `effort:` (also a node field) and Codex-only `modelReasoningEffort:` (workflow-level only). An author had to know which one their node's provider wanted — and that is unknowable while authoring, because the provider can come from a node `provider:`, a model tier, or an `@alias`. Picking wrong failed loudly in one direction and silently in the other.

- **Outcome:** `effort:` is the one reasoning-depth spelling and applies on every provider that has the control — Claude, Codex, Pi, Copilot — however the provider was resolved, and it can be set per node everywhere. A provider with no reasoning control (OpenCode) still warns loudly.
- **Invariant:** A declared reasoning depth is applied by every provider that has one, or it produces a warning. It is never silently dropped, and it never silently suppresses another source of the same value.
- **Scope boundary:** `webSearchMode:` is untouched — still Codex-only, still workflow-level-only, no new `ProviderCapabilities` axis, exactly as decided in #2556. `assistants.<provider>.modelReasoningEffort` in `config.yaml` is a different setting and is unaffected in name, vocabulary, and precedence. `thinking:` is not folded into `effort:`. No node-level `webSearchMode:` was added.
- **Root cause:** `effortControl` was not mis-scoped as a concept — it was factually wrong for exactly one provider. It means "accepts the node-level `effort:` field", and of the five registered providers Codex was the only one with a reasoning control that did not read `nodeConfig.effort`: its `buildThreadOptions` took only `assistantConfig`. Every other symptom was downstream of that one function's inputs.

## Review guidance

- **Feedback requested:** Correctness of the deprecation precedence for `modelReasoningEffort:` (Decision 1 below), and whether the new `@archon/workflows` → provider-registry coupling is the right trade (Decision 2).
- **Start here:** `packages/providers/src/codex/provider.ts` — `resolveModelReasoningEffort` + the `buildThreadOptions` signature. This is the whole fix; everything else is consequence. `requestOptions.nodeConfig` was already in scope at the call site and simply was not passed.
- **Review order:** `codex/provider.ts` → `codex/capabilities.ts` (`effortControl: true`) → `shared/effort.ts` (the ladder + clamp) → `dag-executor.ts` (`declaredEffort`, the collapsed `applyPresetOptions`, the unbranched `resolvedEffort`) → `model-validation.ts` (deletions) → `loader.ts` (deprecation warning) → the rewritten tests → web/CLI/docs.
- **Lower-attention areas:** the docs and `.claude/skills/**` table edits; `provider-capabilities.md` is generated (`bun run generate:capability-matrix`, verified by `check:capability-matrix` in CI). The Pi translator rewrite is mechanical — its existing test file passes **unmodified**, which is the evidence it is behavior-preserving.
- **Known risk or uncertainty:** `node_started.effort` reports the declared rung, not the provider-clamped one (`effort: max` on Codex reports `max`, runs `xhigh`). That matches what Pi and Copilot have always done, and the #2559 failure mode it replaced — reporting a *different field* than the one applied — is now structurally impossible with one channel. Separately, `model-validation.ts` now reads the provider registry, so it depends on `registerBuiltinProviders()` having run; every production entrypoint does this at boot, and the helper returns `null` rather than throwing for an unregistered id.

## Solution

Make Codex translate `nodeConfig.effort` internally, the way Claude, Pi and Copilot already do — which is the stated package contract ("providers receive raw `nodeConfig` + `assistantConfig` and translate to SDK-specific options internally"). With that one fact corrected, `effortControl: true` is honest and the machinery the engine carried to route around it collapses rather than needing extension:

- `routePresetEffort`, `EffortRouting`, `CLAUDE_EFFORTS` and `CODEX_REASONING_EFFORTS` are **deleted**. Presets write the one `nodeConfig.effort` channel.
- The provider-gated `assistantConfig.modelReasoningEffort` write is gone.
- `resolvedEffort` loses its provider branch and its 20-line justification — it is `nodeConfig.effort ?? assistantEffort`.
- `packages/workflows` no longer contains any provider's effort vocabulary or knows which field a provider reads. The one enum left, `modelReasoningEffortSchema`, exists only to parse the deprecated field and is annotated to die with it.

`effortLevelSchema` widened to `minimal | low | medium | high | xhigh | max` — the union of what the SDKs accept — **derived** from a shared `EFFORT_LADDER` in `@archon/providers` rather than restated, so the YAML enum and the clamp cannot drift. Each provider clamps a rung its SDK lacks to the nearest one it has: `max` → `xhigh` on Codex/Pi/Copilot, `minimal` → `low` on Claude/Copilot. That generalizes the precedent `parseCopilotConfig` has shipped since Copilot landed, and it means `effort: max` reliably means "as deep as this model goes".

Two decisions worth naming:

1. **The deprecated field is translated, not honoured indefinitely.** The loader turns `modelReasoningEffort: high` into `effort: high` and never carries the old field onto the workflow; every value in its enum is a valid `effort` rung, so the translation is total. This is what makes the deprecation terminal — and it is what unblocks #1764's Task 1, which collapses workflow-level config onto nodes: `effort:` has a node-level counterpart and `modelReasoningEffort:` never did. Parsing a workflow that declares every node-affecting workflow-level field now emits exactly one orphan, `webSearchMode`, down from two. When both are declared `effort:` wins and the old field is dropped, because the loader cannot know which nodes resolve to Codex; the warning names which value won and which was dropped.
2. **`validEffortsForProvider` derives from `capabilities.ts`** instead of restating which providers take effort. Restating it is precisely the bug being fixed, and AGENTS.md names `capabilities.ts` as the source of truth. `dag-executor.ts` and `loader.ts` in the same package already import the registry, so this adds no package edge.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `effort:` on a Codex node warned "codex doesn't support it" and was dropped; reasoning depth for Codex required workflow-level `modelReasoningEffort:`, with no per-node form. `effortLevelSchema` was `low\|medium\|high\|max`, so a Claude node could not request `xhigh`. A tier/`@alias` carrying `effort` was silently dropped on Pi and Copilot, and the console hid the effort field for them. | `effort:` applies on Claude, Codex, Pi and Copilot, per node or workflow-level, clamped into each SDK's vocabulary. The ladder is `minimal..max`. Tier/alias effort reaches Pi and Copilot, and the console offers the field wherever `capabilities.effortControl` is true. |
| **Deprecated spelling** | Workflow-level `modelReasoningEffort:` was a second field the engine carried and resolved at execution time, Codex-gated. | Translated into `effort:` at load with a warning naming the result, then never seen again. Declaring both keeps `effort:` and drops the old field — a change from v0.9.0 (released this morning), where Codex took the old field's value. Before v0.9.0 the field never reached Codex at all, so no workflow can have relied on that precedence for longer than hours. |
| **Failure behavior** | A Codex node with *both* a tier carrying `effort` and its own `effort:` ran with **no** reasoning effort — the explicit value suppressed the preset (`applyPresetOptions` bails when an explicit effort is set) and then went to a field Codex never read. The node reasoned *less* than the same node with the `effort:` line deleted, with no error. | The explicit value is applied. A rung outside a provider's vocabulary clamps to the nearest one; a value that is not a rung at all is dropped with a provider warning; a provider with no reasoning control produces the existing `dag.unsupported_capabilities` warning. |

## Architecture

```mermaid
flowchart LR
  subgraph Before
    direction TB
    B1["node/workflow<br/>effort:"] --> B3["nodeConfig.effort"]
    B2["workflow<br/>modelReasoningEffort:"] -->|"provider === codex"| B4["assistantConfig<br/>.modelReasoningEffort"]
    B5["tier / @alias<br/>preset effort"] -->|routePresetEffort| B3
    B5 -->|routePresetEffort| B4
    B3 --> B6["claude / pi / copilot"]
    B3 -.->|ignored| B7["codex"]
    B4 --> B7
  end
  subgraph After
    direction TB
    A1["node/workflow<br/>effort:"] --> A3["nodeConfig.effort"]
    A2["workflow modelReasoningEffort:<br/>deprecated shim"] --> A3
    A5["tier / @alias<br/>preset effort"] --> A3
    A3 --> A6["claude / codex / pi / copilot<br/>translate + clamp"]
    A7["config.yaml<br/>assistants.*"] -.->|fallback| A6
  end
```

The dotted `nodeConfig.effort ⇢ codex` edge on the left is the defect. Moving it inside the Codex provider is the change; the two extra arrows on the left existed only to work around it.

### Changed seams

|---|---|---|
| `@archon/workflows` → `@archon/providers` (effort routing) | Removed. The engine no longer knows any provider's effort vocabulary or target field; it writes one `nodeConfig.effort`. | `routePresetEffort`/`EffortRouting`/`CLAUDE_EFFORTS`/`CODEX_REASONING_EFFORTS` deleted; grep for all four returns no hits repo-wide |
| `@archon/providers/effort` (new export subpath) | Added. `EFFORT_LADDER` + `clampEffort`, zero SDK deps; `effortLevelSchema` derives from it. | `packages/providers/src/shared/effort.ts`, `shared/effort.test.ts` |
| `CODEX_CAPABILITIES.effortControl` | `false` → `true`, now factual. | `codex/capabilities.ts`; generated matrix diff is one line |
| `CodexProviderDefaults.modelReasoningEffort` ↔ SDK enum | The unenforced "structurally matches" comment became a compile-time link. `types.ts` is a hard no-SDK contract layer, so the pin lives one layer out: `CODEX_EFFORTS ... as const satisfies readonly ModelReasoningEffort[]`, narrowed by a type guard instead of the previous cast. | `codex/config.ts` |
| Workflow YAML surface | `effort:` widened to six rungs (additive — every value valid before is still valid); `modelReasoningEffort:` deprecated with a load-time warning. | `schemas/dag-node.ts`, `loader.ts`, `loader.test.ts` |
| `archon workflow list --json` | Now emits `effort` alongside `modelReasoningEffort`; it previously surfaced only the deprecated spelling. | `packages/cli/src/commands/workflow.ts`, `reference/cli.md` |
| Console tier/alias effort picker | Decides from `capabilities.effortControl` on the already-fetched providers list instead of a hardcoded agent-id list. | `console/lib/model-options.ts`, `ModelTiersPanel.tsx`, `AliasesPanel.tsx` |

## Validation

- `bun run validate` — passed (exit 0) — 141 test files reporting `0 fail`; all five generated-file checks in sync; type-check clean across 13 packages; `eslint --max-warnings 0` and `prettier --check` clean. Proves the integrated outcome plus no drift in the generated capability matrix.
- `cd packages/providers && bun run test` — passed — proves per-provider clamping at each SDK boundary. Notably **`community/pi/options-translator.test.ts` passes unmodified**, which is the evidence that rewriting Pi onto the shared clamp changed no behavior.
- `cd packages/workflows && bun run test` — passed — `dag-executor.test.ts` 462 pass, including the regression test `an explicit effort on a Codex node beats its tier preset instead of erasing both` (fails before this change: nothing is sent) and the four #2559 tests rewritten rather than replaced, so their subjects — deprecated field reaches Codex, beats a preset, does not disturb a Claude node, warns on non-Codex — all still hold.
- **Runtime smoke, loader → CLI surface:** a temp repo with two workflows, run through `archon workflow list --json`. The `modelReasoningEffort: high` workflow keeps its value **and** carries `parseWarnings: ["Workflow 'old-spelling': 'modelReasoningEffort: high' is deprecated — use 'effort:' … While both are set, 'modelReasoningEffort' still wins on Codex nodes."]`; the `effort: max` workflow reports `"effort": "max"` with no warnings.
- **Not verified:** no live provider call was made — the SDK-boundary assertions are on the options object handed to each SDK (`ThreadOptions.modelReasoningEffort`, `Options.effort`, `reasoningEffort`, Pi's `ThinkingLevel`), not on model behavior. The SDK enums were read from the installed packages rather than assumed (`@anthropic-ai/claude-agent-sdk` `sdk.d.ts:527`, `@openai/codex-sdk` `dist/index.d.ts:237`), and each provider's translation is now type-pinned to its SDK, so a vocabulary change upstream fails type-check rather than a run.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No YAML edits required of anyone. `modelReasoningEffort:` keeps parsing and keeps winning on Codex; the schema widening is additive. The one intended change is that `effort:` starts *working* on Codex where it was warned-and-dropped (and, per the root cause, where it silently suppressed a preset). | The four rewritten #2559 tests assert the deprecated field's precedence is unchanged; `loader.test.ts` asserts the value still reaches the workflow object |
| Rollout / rollback | Three independently revertible layers: the provider change is inert until the engine stops routing around it; the engine collapse is a single-commit revert; the surface changes are cosmetic. No stored data, schema migration, or run-state format change — `check:schema-upgrades` is not applicable. | Single commit; no `migrations/` or SQLite schema touched |
| Observability | `node_started.effort` keeps its shape and now cannot name a field the provider ignored. Provider-side clamping is logged (`codex.effort_clamped`, `claude.effort_clamped` at debug; `*.effort_unrecognized` at warn). Deprecation reaches authors through `parseWarnings` and the run-path log, per #2213. | Runtime smoke above; `node_started` assertions retained in every rewritten test |
| Documentation / communication | Seven doc surfaces taught two spellings and are updated: the authoring guide (incl. a deprecation callout), `book/quick-reference.md`, `getting-started/ai-assistants.md`, `reference/cli.md`, and three `.claude/skills/archon/references/*` tables an agent reads when authoring. `provider-capabilities.md` regenerated. CHANGELOG entry under Changed + Fixed. | Included in this PR |

Constitution citation (AGENTS.md requires one for YAML-surface changes): *"New per-provider capabilities default to provider config / tier-alias presets, not new node fields; capability mismatches warn loudly (capabilities.ts is the source of truth)."* This change is what makes that rule true for reasoning depth — today `capabilities.ts` is not the source of truth for it. Note the change **removes** YAML surface rather than adding it, so the admissibility test is satisfied trivially.

## Links

- Closes #2556
- Related #2559 — its four tests are the contract this PR rewrites; it deferred the one-spelling question here
- Related #2582 — a pre-existing telemetry gap found while reviewing this change and deliberately left out of scope: an unvalidated `assistants.*` value in `config.yaml` is reported in `node_started` as if it ran. Fixing it inside the executor would reintroduce the provider-vocabulary knowledge this PR removes
- Related #1764 — this was its prerequisite: `modelReasoningEffort:` had no node-level counterpart, so the config-collapse invariant would have shipped with a two-field hole. The exception is now one field, `webSearchMode:`
- Implementation plan: https://github.com/coleam00/Archon/issues/2556#issuecomment-5316856238

## Summary by CodeRabbit

* **New Features**
  * Added provider-neutral `effort` settings from `minimal` through `max` for Claude, Codex, Pi, and Copilot.
  * Added provider-aware effort selection in the console and workflow configuration.
  * Unsupported levels are automatically clamped to each provider’s supported range.

* **Bug Fixes**
  * Improved precedence for workflow, node, and preset effort settings.
  * Providers without effort support now ignore unsupported settings consistently.

* **Documentation**
  * Updated workflow and CLI references with provider-specific effort behavior.

* **Deprecation**
  * `modelReasoningEffort` is translated to `effort` with warnings for compatibility.


Wirasm:
# PR Review: #2581 — refactor(workflows): effort: sets reasoning depth on every provider that has one

## Signal

**Round 11 — `2d5c248c`. Nothing blocks, `R48` closes, and the round's most valuable finding belongs to `dev`, not to this PR.** Two things landed since round 10's `7399a4de`: `3fec518e` (closing `R48` and `R49`) and `2d5c248c` (a merge of `origin/dev` = `7752ec53`). Both were examined, and — as instructed — the merge hardest.

**`R48` is genuinely FIXED, and the guard was live-fired three times independently rather than believed.** `scripts/effort-ladder-parity.test.ts` compares `EFFORT_OPTIONS` (web) to `EFFORT_LADDER` (providers) **by value** through direct imports, and `packages/web/.../model-options.test.ts` now derives its `LADDER` from the constant under test, taking three hand-typed copies down to two. Two specialists reproduced the live-fire (drop a rung → red) and I confirmed the guard actually reaches CI: root `package.json`'s `test` is `bun --filter '*' --parallel test && bun test ./scripts/`, and `validate` includes `test`. On the failure mode the brief asked about — a test deriving from the thing it tests, and so unable to detect drift — **it is real and it was located precisely**, but it is not where the guard is. It is at `model-options.test.ts:104`, and it costs nothing today (`R53`). The second test in the new parity file is narrow but not vacuous: it cannot fail while `effortLevelSchema = z.enum(EFFORT_LADDER)` derives, and it fails the moment someone re-hardcodes that enum — the exact regression shape it claims. Live-fired in both directions.

**The new guard avoids `node-ref-parity.test.ts`'s fragility for a structural reason, not a stylistic one.** #2591 had to rebuild that guard after its text-scraping half lost its target; the scraping exists only because `OUTPUT_REF_SOURCE` (`loader.ts:119`) is an unexported module constant. `EFFORT_LADDER`, `EFFORT_OPTIONS` and `EFFORT_LEVELS` are all exported, so the effort guard never needs to scrape at all.

**On the two deleted assertions the brief asked me to judge: statically safe, dynamically not — and that reopens `R51` rather than creating a new finding.** The surviving `toBe` literal contains the diagnosis phrase and all six rung names, so neither deleted `toContain` can fail while the equality passes. That is the whole of the commit's argument and it is correct *for a fixed expected literal*. It does not hold across an edit of that literal: an author who reworks the warning, drops a rung, and repairs the red test the reflexive way — paste `actual` into `expected` — now satisfies the only remaining assertion on the message's content. The six-rung loop checked the **live return value**, so it would have kept failing after that repair. Round 10 recorded `R51` as NOT A FINDING solely because the loop was still present; `3fec518e` deleted it, so `R51`'s premise is now true and it is reopened. Two specialists called the deletion safe (both reasoning statically); one made the dynamic case. **The disagreement is preserved because both are right about different regimes.** It stays a Suggestion: no production behavior is affected, round 10 itself offered deletion as one of two acceptable remedies, and the diagnosis half retains independent live-value protection at `provider.test.ts:1056`.

**The merge resolution is correct and complete.** `capsMock` has exactly four mutation sites in `orchestrator-agent.test.ts` (`:1360`, `:1385`, `:2455`, `:4839`); every one now spreads `DEFAULT_PROVIDER_CAPS` and every restore is `{ ...DEFAULT_PROVIDER_CAPS }`, byte-matching the module-level default at `:170`. Restoring `effortControl: true` in #2577's `afterEach` breaks nothing of #2577's: no test in the `paused approval gate routing` block (`:2331-2769`) asserts on effort or reasoning. The suite is 235 pass / 0 fail, and the merged file is provably "dev tip + this PR's own local diff" — `git diff 7752ec53:<file> 2d5c248c:<file>` reproduces this PR's pre-merge diff exactly. **I hunted the other five doubly-touched files for the same class of silent conflict and found none**; edits land in disjoint regions in each. `api.generated.d.ts` is again not stale, and this round the question was live rather than academic — dev *did* touch `packages/server/src/routes/api.ts` and `packages/workflows/src/schemas/workflow-run.ts` this time, but the first changes one user-facing message string inside a handler and the second adds a function plus docblocks. Neither touches a Zod schema or a route config.

**`R49` does not fully close.** Three of its four items are fixed; **item 3 survives verbatim** at `options-translator.test.ts:73`. The comment still instructs a reader to run `git show 9b021e3c` to see an append that commit does not contain — `git show 9b021e3c -- .../options-translator.ts` is empty, and the append was made by `97ebec7d`. `3fec518e` rewrote nearly every other line of that block. One new instance of the same class: `2d5c248c`'s message says it "Brings in #2577, #2591, #2592", but #2591 **is** the merge-base (`7fe762a1`) and was already an ancestor of `3fec518e` (`R52`).

**The round's biggest finding is not this PR's.** `packages/core/package.json`'s enumerated `test` script omits two of the package's 61 test files — and one of them, `src/schemas/index.test.ts`, is **red right now** (18 pass / 2 fail) while CI is green. Its fixtures rotted across three separate PRs: `codebaseRowSchema` lost `kind` (#2055), `dashboardWorkflowRunSchema` lost `parent_run_id` (#2121) and `output_root` (#2200/#1192) — each root-caused by running the schemas, not reading them. The other, `prompt-builder.test.ts`, is where #2577 put the only real coverage of its own Breaking change. **All five relevant files are byte-identical to `origin/dev`**, so merging #2581 neither causes nor worsens it. Recorded against the existing issue #2401, which already predicted this generalizing and names `@archon/core` by name (`R54`).

**Validation is green at the exact reviewed tree.** `bun run validate` exit 0, zero `error TS`, no non-zero `fail` line; all six required CI checks pass at `2d5c248c`.

- **Blocking:** 0
- **Non-blocking:** 10 (`R35`–`R39`, `R41`, `R49`, `R51`, `R52`, `R53` — all Suggestions)
- **Resolved since previous review:** 1 fully (`R48`); `R49` narrowed from four items to one — 36 of 54 findings FIXED, 2 NOT A FINDING, 2 DECLINED, 4 tracked
- **Tracked follow-ups:** 4 — `R8` → #2584, `R17` → #2586, `R50` → #2597, `R54` → #2401 (all verified OPEN)
- **Validation:** `bun run validate` PASS (exit 0) at `2d5c248c` on a clean tree. All 6 required CI checks green

## Findings

|---|---|---|---|
| `R1` | Important | Deprecation warning and docs claimed `modelReasoningEffort` always wins on Codex | FIXED |
| `R2` | Important | `AGENTS.md` listed node-level `effort:` as Claude/Pi/Copilot only | FIXED |
| `R3` | Suggestion | `dag-node.ts` imported a runtime value while a comment said the package cannot | FIXED |
| `R4` | Important | New `@archon/providers/effort` import as a dependency-contract violation | NOT A FINDING |
| `R5` | Suggestion | Copilot's effort vocabulary defined twice; only one copy pinned | FIXED |
| `R6` | Suggestion | Pi's capability table listed the pre-widening four-rung vocabulary | FIXED |
| `R7` | Suggestion | No test proved a node-level `effort:` rung survives `dagNodeSchema.safeParse()` | FIXED |
| `R8` | Suggestion | A preset effort dropped for a provider with no reasoning control is log-only | TRACKED FOLLOW-UP |
| `R9` | Suggestion | Preset-effort validate-and-branch hand-duplicated across two packages | FIXED |
| `R10` | Suggestion | Effort symbols re-exported from the SDK-heavy root barrel with no consumers | FIXED |
| `R11` | Suggestion | `clampEffort`'s docblock called the rule "nearest" when it is down-first | FIXED |
| `R12` | Suggestion | Neither caller of the shared preset-effort gate proved it obeys a rejection | FIXED |
| `R13` | Important | `authoring-workflows.md` table and Summary stated the pre-PR effort set | FIXED |
| `R14` | Suggestion | `R11`'s correction reached one of two docblocks in the same file | FIXED |
| `R15` | Suggestion | Checked-in `api.generated.d.ts` still typed the four-rung union | FIXED |
| `R16` | Suggestion | Pi's `thinking`-mismatch warning still listed the four-rung effort set | FIXED |
| `R17` | Suggestion | No check covers `api.generated.d.ts` freshness | TRACKED FOLLOW-UP |
| `R18` | Important | "max → xhigh on Pi" survived at the 2 user-facing sites, false in the opposite direction | FIXED |
| `R19` | Important | `effort: max` on Pi was silently downgraded; 95 catalog models support `max` | FIXED |
| `R20` | Suggestion | `as const satisfies readonly T[]` proves containment, never coverage | FIXED |
| `R21` | Suggestion | The PR added `effort_clamped` logging to 2 of its 4 clamp call sites | FIXED |
| `R22` | Important | The translation widened the deprecated field to every provider, undisclosed | FIXED |
| `R23` | Suggestion | Copilot's `satisfies` target claimed to be a hand-copy of a hand-copy | NOT A FINDING |
| `R24` | Suggestion | Nothing pinned the remedy clause of the Pi warning that had drifted four times | FIXED |
| `R25` | Suggestion | The OpenAPI response contract still permits a field the loader guarantees absent | DECLINED |
| `R26` | Suggestion | `WorkflowLevelOptions`' rewritten docblock claims a counterpart `workflowTier` lacks | FIXED |
| `R27` | Suggestion | `include-expander.ts` still lists `modelReasoningEffort` as a key `parseWorkflow` emits | FIXED |
| `R28` | Suggestion | Two of six docs state the translation unconditionally, omitting the both-declared drop | FIXED |
| `R29` | Suggestion | The deprecated field is parsed ~290 lines above its only consumer, bridged by a comment | FIXED |
| `R30` | Suggestion | The `#2457` guard's failure diagnostic is mis-branched for one fixture | DECLINED |
| `R31` | Suggestion | The implementation report still asserts a code fact that is false | FIXED |
| `R32` | Suggestion | `workflow-yaml-reference.md` still labels the field "Codex only" | FIXED |
| `R33` | Suggestion | The widened scope is promised in three places and pinned by no behavioral test | FIXED |
| `R34` | Suggestion | The assertion that closed `R24` cannot detect the drift class it guards | FIXED |
| `R35` | Suggestion | Pi's own per-model SDK clamp is now the only clamp, and nothing observes it | OPEN |
| `R36` | Suggestion | A config-file effort the provider parser silently drops is still reported as applied | OPEN |
| `R37` | Suggestion | The declared-effort gate restates `resolvePresetEffort`'s rule instead of calling it | OPEN |
| `R38` | Suggestion | Pi's clamp array and guard could collapse to `isEffortRung`, keeping drift protection | OPEN |
| `R39` | Suggestion | Surfaces describing clamping as universal — 1 of 3 fixed, 2 reworded cosmetically | OPEN |
| `R40` | Suggestion | The parameter matrix says Copilot "maps like Pi"; Pi maps nothing and Copilot clamps both ends | FIXED |
| `R41` | Suggestion | A valid rung clamped by the provider is recorded in `node_started` as the declared one | OPEN |
| `R42` | Suggestion | A loader comment says the run path never reads the warning string; it does | FIXED |
| `R43` | Suggestion | A Codex docstring points at `resolveReasoning`, a symbol that does not exist | FIXED |
| `R44` | Suggestion | Two executor-test artifacts still describe the abandoned Codex scoping | FIXED |
| `R45` | Suggestion | `AGENTS.md`'s config-priority example still names the deprecated field | FIXED |
| `R46` | Suggestion | `R34`'s guard was a blacklist: it rejected a true claim and passed six false ones | FIXED |
| `R47` | Suggestion | The positive pin is defeated by conjunction — a false downgrade *added beside* the true clause passes | FIXED |
| `R48` | Suggestion | The PR's own widened web effort ladder hand-copies `EFFORT_LADDER` with no pin of any kind | **FIXED** |
| `R49` | Suggestion | `8be5fefc`'s commit message and comment carried four inaccuracies — item 3 survives | **OPEN** |
| `R50` | Suggestion | The `archon` skill still teaches a `when:` gate that dev's #2579 made a load error | TRACKED FOLLOW-UP |
| `R51` | Suggestion | Deleting the rung loop removes the only check that survives a literal-sync repair of the `toBe` | **OPEN (reopened)** |
| `R52` | Suggestion | `2d5c248c`'s message credits the merge with #2591, which is its own merge-base | **OPEN** |
| `R53` | Suggestion | `model-options.test.ts:104` became tautological — it compares `EFFORT_OPTIONS` to itself | **OPEN** |
| `R54` | Suggestion | Two `@archon/core` test files are unreachable from its `test` script; one is red and invisible to CI | **TRACKED FOLLOW-UP** |

## Detailed Findings

<details>
<summary><code>R48</code> — the widened web effort ladder is now pinned by value, and the pin bites</summary>

**Impact (round 10):** `da045f59` collapsed two narrow hand-copied web ladders into one widened `EFFORT_OPTIONS` (`packages/web/src/experiments/console/lib/model-options.ts:113`) whose only tie to `EFFORT_LADDER` (`packages/providers/src/shared/effort.ts:20`) was a sentence in a comment. `git grep -l` returned disjoint file sets, and the file's own test hand-typed a third copy. It feeds a live dropdown rendered in `ModelTiersPanel.tsx` and `AliasesPanel.tsx`.

**Disposition:** **FIXED in `3fec518e`**, by remedy (b) of the two round 10 offered, plus a cleanup that closes the third copy.

- `scripts/effort-ladder-parity.test.ts` imports both constants directly and compares by value with `toEqual`. **Live-fired three times independently** — by `prp-core:seam-analyzer`, by `prp-core:pr-test-analyzer` (dropping `xhigh`), and by `prp-core:code-reviewer` in a disposable detached worktree (dropping `minimal`) — red in every case, green at head.
- `model-options.test.ts:96` now does `const LADDER = EFFORT_OPTIONS`, so the file can no longer pass while disagreeing with the constant it tests. Three copies → two.
- **The guard reaches CI**, verified rather than assumed: root `package.json` `test` is `bun --filter '*' --parallel test && bun test ./scripts/`, and `validate` includes `test`. `bun test ./scripts/effort-ladder-parity.test.ts` → 2 pass / 0 fail.

On the brief's specific worry — that a test deriving from the thing it tests cannot detect drift — **the guard itself is clean**: `EFFORT_OPTIONS` and `EFFORT_LADDER` are two independently hand-typed arrays in disjoint packages, so nothing is self-referential. The derivation concern is real but lands at `model-options.test.ts:104` instead, and is recorded as `R53`.

The second test in the file (`EFFORT_LEVELS` vs `EFFORT_LADDER`) is narrow, not vacuous. `EFFORT_LEVELS = effortLevelSchema.options` and `effortLevelSchema = z.enum(EFFORT_LADDER)` (`dag-node.ts:48,53`) is a genuine derivation, so it cannot fail today — but `prp-core:pr-test-analyzer` live-fired it in both directions and confirmed it fails cleanly when `dag-node.ts` is edited to hardcode a literal enum, which is precisely the regression its docstring claims. Correctly scoped.

**Remaining copies swept:** `git grep` over `packages/` and `scripts/` for the six-rung literal leaves only `EFFORT_LADDER` (canonical), `EFFORT_OPTIONS` (now pinned), and the generated union in `api.generated.d.ts` (machine-written; its own freshness gap is `R17`/#2586). The fourth copy noted in round 10 — the hand-typed rung loop in `options-translator.test.ts` — is gone with the loop. `CODEX_CONFIG_EFFORT_OPTIONS` (`model-options.ts:120`, five rungs, no `max`) is the Codex SDK's separate `modelReasoningEffort` vocabulary and correctly out of scope.

**Found by:** `prp-core:seam-analyzer` (round 10); fix verified this round by `prp-core:seam-analyzer`, `prp-core:pr-test-analyzer`, `prp-core:code-reviewer`

</details>

<details>
<summary><code>R51</code> — REOPENED: deleting the rung loop removes the only check that survives a literal-sync repair</summary>

**Why it is reopened rather than re-filed.** Round 10 recorded this NOT A FINDING for one reason: the six-rung loop was still in the file, so the premise was false. `3fec518e` deletes it. Same causal finding, premise now true, so the ID carries forward with a changed state.

**The commit's argument, and exactly how far it holds.** `3fec518e` removes both the six-rung `toContain` loop and `toContain('object form is Claude-specific')`, arguing "`toBe` subsumes both… an assertion that cannot fail while the equality passes is not a second opinion." **Statically that is correct and was verified by running the suite** (35 pass / 0 fail): the surviving literal contains the diagnosis phrase and all six rung names, so neither deleted assertion can fail while the equality passes.

**It does not hold across an edit of the expected literal.** The sequence: an author reworks the warning's wording and drops a rung from the list — a realistic slip, since the string is hand-typed rather than interpolated from `EFFORT_LADDER`. `toBe` goes red with a string diff. The natural repair is to paste the new production output into the literal. That single action satisfies the only remaining content assertion, because `toBe` asserts equality with whatever is pasted. The deleted loop asserted against the **live return value**, independently of the literal, so it would have stayed red and forced the author to confront the missing rung. That trip-wire is gone.

**Evidence:** `packages/providers/src/community/pi/options-translator.test.ts:88-91` (the surviving `toBe`); `git show 3fec518e` (the deleted loop at the old `:61-67`); `options-translator.ts:92-98` (the hand-typed message). Grep across `packages/providers/src/` found no other check on the six-rung remedy list — `effort.test.ts` tests `clampEffort`/`EFFORT_LADDER` directly, not this English sentence.

**Not symmetrical for the diagnosis half.** `packages/providers/src/community/pi/provider.test.ts:1056` independently asserts `'object form is Claude-specific'` against live `sendQuery()` output. That deletion really is fully covered elsewhere; only the rung list is now singly-protected.

**Reviewer disagreement, preserved because both readings are right about different regimes.** `prp-core:code-reviewer` and `prp-core:seam-analyzer` both judged the deletion safe, reasoning about the committed tree. `prp-core:pr-test-analyzer` made the maintenance-lifetime case above. The two are not in conflict — the commit's claim is true for a fixed literal and false across an edit of it.

**Required outcome:** Optional. Restore a rung-membership loop against the live value — either back in this test or beside the existing diagnosis check at `provider.test.ts:1056`, which already protects the other half of the same message that way.

**Found by:** `prp-core:pr-test-analyzer` (rounds 10 and 11)

**Disposition:** **OPEN as a Suggestion.** Non-blocking: no production behavior changes, the suite is green, round 10 explicitly offered deletion as one of two acceptable remedies (calling the alternative "better"), and the author took a disclosed trade rather than making an error. Recorded because the commit message's stated justification is incomplete in a way that matters to whoever edits this string next — and five prior drifts say someone will.

</details>

<details>
<summary><code>R49</code> — three of four items fixed; the <code>9b021e3c</code> citation survives verbatim</summary>

**Fixed in `3fec518e`:**

1. The rung loop now really is collapsed into the equality — the code was changed to match the claim rather than the claim softened. Verified against the committed file.
2. "Three mechanisms" → "FOUR mechanisms", correctly heading the four-item list.
4. The present-tense dev-drift claim is now past-tense and durable: "caught a real drift there in #2591".

**Item 3 is untouched.** `packages/providers/src/community/pi/options-translator.test.ts:73` still reads ``// `git show 9b021e3c` has `… in YAML (max → xhigh on Pi; the rest are…`. Confirmed directly: `grep -n '9b021e3c\|97ebec7d'` on the file returns one hit for `9b021e3c` and none for `97ebec7d`; `git show 9b021e3c -- packages/providers/src/community/pi/options-translator.ts` produces an **empty diff**. The string does exist in that commit's tree (`git show 9b021e3c:<path>` has it at line 96) so the substantive claim holds — but the append the sentence describes was made by `97ebec7d`, and a reader running the literal command named in the comment gets nothing.

`3fec518e` rewrote nearly every other line of this block while leaving the one flagged line alone, and its commit message never claims to have fixed it.

**Required outcome:** one token — `9b021e3c` → `97ebec7d` (or append `:packages/providers/src/community/pi/options-translator.ts` to the command).

**Found by:** `prp-core:comment-analyzer` (rounds 10 and 11); re-verified independently against git objects this round

**Disposition:** **OPEN as a Suggestion**, narrowed from four items to one. Non-blocking. Retained because the block's whole function is to stop a future editor re-litigating the `toContain`-vs-`toBe` decision, and a citation that yields an empty diff invites exactly the wasted verification `R49` already cost a reviewer once.

</details>

<details>
<summary><code>R52</code> — the merge commit credits itself with #2591, which is its own merge-base</summary>

**Impact:** `2d5c248c`'s message opens "Brings in #2577, #2591, #2592." It does not bring in #2591. `git merge-base 3fec518e 7752ec53` is `7fe762a1`, which **is** `fix(web): the workflow builder rejected \`when:\` conditions the engine accepts (#2591)` — it entered this branch at the *previous* merge (`7399a4de`) and was already an ancestor of `3fec518e`. Confirmed: `git merge-base --is-ancestor 7fe762a1 3fec518e` → true. `git log --oneline 7fe762a1..7752ec53` shows only #2577 and #2592 unique to dev's side, and the merge's actual delta to `scripts/node-ref-parity.test.ts` is #2592's precedence note, not #2591 work.

**Consequence:** low but real, and precisely on this PR's theme. Someone bisecting, or asking whether #2591 needed conflict resolution against this branch, would look for merge work that does not exist.

**Required outcome:** Optional, and unactionable without a rewrite of a pushed merge commit — noted for the record rather than for correction.

**Found by:** `prp-core:comment-analyzer`; verified independently with `git merge-base --is-ancestor`

**Disposition:** **OPEN as a Suggestion.** Non-blocking.

</details>

<details>
<summary><code>R53</code> — <code>model-options.test.ts:104</code> became tautological as a side effect of the `R48` cleanup</summary>

**Impact:** `const LADDER = EFFORT_OPTIONS` (`:96`) is imported from `./model-options` — the same module as `effortOptionsForAgent`, which returns `EFFORT_OPTIONS` **by reference, unmodified**:

```ts
// packages/web/src/experiments/console/lib/model-options.ts:127-131
return provider?.capabilities?.effortControl ? EFFORT_OPTIONS : null;
```

So `expect(effortOptionsForAgent(id, PROVIDERS)).toEqual(LADDER)` at `:104` now compares one object to itself. It cannot fail on content whatever `EFFORT_OPTIONS` holds. Before `3fec518e`, `LADDER` was an independent literal and the assertion carried real content value. What survives is control-flow coverage only — null vs non-null per agent id.

**Why this is recorded and not raised further.** The content question genuinely moved rather than vanished: `scripts/effort-ladder-parity.test.ts` now checks `EFFORT_OPTIONS`'s content against an independent source in a different package. Net system coverage is preserved, and consolidating the check in one authoritative place is better than three silently-divergeable copies. Both `prp-core:seam-analyzer` and `prp-core:pr-test-analyzer` reached this conclusion independently, and the file's own comment at `:92-95` describes the trade accurately.

**Residual:** the two files are in different packages and different legs of `bun run test` (`bun --filter '*'` vs `bun test ./scripts/`). If the parity test were ever deleted or skipped, `:104` would give false confidence with zero content signal, and nothing links them but a comment.

**Found by:** `prp-core:seam-analyzer`, `prp-core:pr-test-analyzer`

**Disposition:** **OPEN as a Suggestion.** Non-blocking; no change recommended today.

</details>

<details>
<summary><code>R54</code> — two <code>@archon/core</code> test files are unreachable from its <code>test</code> script; one is red and invisible to CI</summary>

**Impact:** `packages/core/package.json`'s `test` script enumerates test files **by name** — 55 names plus three directory globs (`src/config/`, `src/state/`, `src/workflows/`). The package has 61 test files. Tokenising the script and diffing against `find src -name '*.test.ts'` leaves two unreachable:

- `src/orchestrator/prompt-builder.test.ts` — **29 pass**, never run by CI
- `src/schemas/index.test.ts` — **18 pass / 2 fail**, never run by CI

**The red file has been rotting across three PRs**, each failure root-caused by running the schemas rather than reading them:

| Test | Missing required field | Added by |
|---|---|---|
| `codebaseRowSchema accepts a valid row` | `kind` (`invalid_value`) | #2055 folder projects |
| `dashboardWorkflowRunSchema preserves all base workflowRun fields` | `parent_run_id`, `output_root` (`invalid_type`) | #2121, #2200/#1192 |

**The unwired file holds a Breaking change's only real coverage.** `prompt-builder.test.ts` is where #2577 put its 154 new lines covering `formatPausedGateSection` — the function behind the `[Unreleased]` Breaking entry *"a plain message at an approval gate no longer approves it"*. The three-outcome contract (approve / reject / resolve **nothing**), verbatim quoting of the rejection reason, and the no-tool-naming neutrality rule are asserted only there; `orchestrator-agent.test.ts` covers section presence and the raw message, not those rules.

**Not PR-caused — proven by file identity, not inference.** `git diff --quiet origin/dev HEAD --` is clean for all five relevant files: `packages/core/package.json`, `src/schemas/index.test.ts`, `src/schemas/codebase.ts`, `src/schemas/workflow-run.ts`, `src/orchestrator/prompt-builder.test.ts`. `prompt-builder.test.ts` was created in `e0716e0a`, which `git merge-base --is-ancestor` confirms predates this PR's branch point (`2d545225`). This PR touched nothing under `packages/core/src/schemas/`. Merging #2581 neither causes nor worsens any of it.

**Evidence:** the tokenised-script sweep above; `bun test src/orchestrator/prompt-builder.test.ts` → 29 pass; `bun test src/schemas/index.test.ts` → 18 pass / 2 fail; direct `safeParse` probes returning the three missing-field issues; `.github/workflows/test.yml:68` → root `test` → `bun --filter '*' --parallel test`.

**Found by:** `prp-core:code-reviewer` (as an Important finding against the merged tree); scope, root causes, and PR-attribution established during adjudication

**Disposition:** **TRACKED FOLLOW-UP** → https://github.com/coleam00/Archon/issues/2401, which already documents the identical defect in `@archon/cli` (`src/cli.test.ts` never run), explains why the enumeration is load-bearing (Bun's irreversible `mock.module()` forces the batch split), and proposes a glob-vs-script guard that "generalises to the other four packages" — naming `@archon/core` among them. Commented with the new evidence rather than duplicated: https://github.com/coleam00/Archon/issues/2401#issuecomment-5320981740. **Downgraded from the reviewer's Important to a tracked Suggestion for this PR only** — it is `dev`'s defect byte-for-byte, and blocking #2581 on it would neither fix nor contain it.

</details>

<details>
<summary>The merge — resolution correctness and the hunt for other silent conflicts</summary>

**The known conflict, and whether the resolution is right.** Git auto-merged `orchestrator-agent.test.ts` cleanly and the result was broken: #2577's new approval-gate suite restored `{ envInjection: true }` in `afterEach`, a hand-written subset dropping `effortControl` for the rest of the file, which broke this PR's `resolveTitleRequest` assertion. Both sides were locally correct — #2577 was written against a tree without `DEFAULT_PROVIDER_CAPS`.

**The resolution is correct and complete.** `capsMock` has exactly four mutation sites (`:1360`, `:1385`, `:2455`, `:4839`); every one spreads `DEFAULT_PROVIDER_CAPS` (`:158`, `{ envInjection: true, effortControl: true }`) and every restore is `{ ...DEFAULT_PROVIDER_CAPS }`, byte-matching the module-level default at `:170`. `nativeTools` is absent from the baseline and set explicitly where a suite needs it — correct, since the default mock omits it too. Restoring `effortControl: true` in #2577's `afterEach` changes nothing #2577 depends on: no test in the `paused approval gate routing` block (`:2331-2769`) asserts on effort or reasoning fields. Suite: **235 pass / 0 fail**.

**Mechanically proven to be dev-tip-plus-this-PR:** `git diff 7752ec53:<file> 2d5c248c:<file>` reproduces this PR's own pre-merge diff to that file exactly.

**No other silent semantic conflict found.** Six files were touched by both sides (`comm -12` on the sorted name lists against merge-base `7fe762a1`): `.claude/skills/archon/references/workflow-dag.md`, `CHANGELOG.md`, `packages/cli/src/commands/workflow.ts`, `orchestrator-agent.test.ts`, `orchestrator-agent.ts`, `authoring-workflows.md`. Each was diffed per side; edits land in disjoint regions in every case. Dev's ~381-line `orchestrator-agent.ts` delta never touches `resolveTitleRequest`, `applyPresetToRequestOptions` or `resolvePresetEffort`. No stale "auto-resume"/"natural language" prose survives to contradict #2565/#2577. `manage-run-tool.test.ts` (36 pass), `prompt-builder.test.ts` (29 pass) and `command-handler.test.ts` (131 pass) are green in their own isolated invocations, and the only other hand-written capability object in the tree (`orchestrator.test.ts:101-106,527-532`) is untouched by this delta and runs in its own `mock.module` process.

**`api.generated.d.ts` freshness — a live question this round, and still not stale.** Unlike round 10, dev *did* touch a route file and an engine schema file. Both are inert for the spec: `packages/server/src/routes/api.ts` changes one user-facing message string inside a handler body (`"send a new message in the originating conversation"` → `"resume it from the originating conversation"`), no route config or Zod schema; `packages/workflows/src/schemas/workflow-run.ts` adds the `isContainerRun()` *function* and revises the `isGateResolved()` docblock — no schema shape changes.

</details>

<details>
<summary><code>R35</code>, <code>R36</code>, <code>R37</code>, <code>R38</code>, <code>R39</code>, <code>R41</code> — re-verified OPEN, all sites untouched</summary>

Established in rounds 7 and 8 with full evidence in this comment's published history. Re-verified this round with `git diff --quiet 7399a4de 2d5c248c -- <path>`: `options-translator.ts` **UNTOUCHED**, `model-validation.ts` **UNTOUCHED**, `model-options.ts` **UNTOUCHED**, `dag-executor.ts` **UNTOUCHED** (dev's `resolveRunInputs` threading arrived in the *previous* merge, not this one).

- **`R35`** — Pi's own `clampThinkingLevel(model, level)` inside `createAgentSession` reduces the level to whatever the resolved model declares, silently and with no return value, while `pi.session_started` logs the requested rung. 962 of 1057 catalog entries lack a native `max`. **Pre-existing, not PR-caused.**
- **`R36`** — `parseCodexConfig` is defensive, so `assistants.codex.modelReasoningEffort: max` is rejected and the SDK call omits the field, while the executor's telemetry path reads the same raw value behind only a `typeof === 'string'` guard (`dag-executor.ts:1210-1214`) and writes `"max"` into the permanent `workflow_events` row. **Pre-existing at merge-base.**
- **`R37`** — `dag-executor.ts:1175`'s `effort: caps.effortControl ? declaredEffort : undefined` is a third expression of `resolvePresetEffort`'s rule. **Reviewer disagreement preserved:** `prp-core:code-simplifier` argued the two gates answer different questions and collapsing them means threading `caps` across a package boundary for one caller. Neither reading blocks.
- **`R38`** — `PI_NATIVE_LEVELS` is membership-identical to `EFFORT_LADDER`, so `clampEffort`'s walk is unreachable at that call site. A counter-proposal against the author's evident `AssertNever` preference; explicitly non-blocking.
- **`R39`** — `model-validation.ts:237` and `model-options.ts:104` still describe per-provider clamping in universal terms with no Pi mention. Both vacuous for Pi (its SDK lacks no rung), so low materiality.
- **`R41`** — `dag-executor.ts:1216` writes the pre-clamp `resolvedEffort` to the durable `node_started` event; clamping happens later inside each provider and is reported only at `debug`. **Deliberate and disclosed in the code** (`:1208-1211`).

</details>

<details>
<summary><code>R50</code> — still open, and dev's touch to the same file did not fix it</summary>

Round 10 filed this and tracked it as #2597: `.claude/skills/archon/references/workflow-dag.md` teaches a `classify` `prompt:` node with no `output_format` gated by `when: "$classify.output != 'SAFE'"`, which dev's #2579 made a **load-time error**.

Dev touched this exact file in the merge (7 lines), so the question was whether #2597 is now closeable. **It is not.** `prp-core:docs-impact-agent` read the diff: dev's edits are confined to the `effort:`/`modelReasoningEffort:` field table and the approval-gate natural-language example. The stale pattern survives at `:717-731`, still a load error against `loader.ts:380` and the error text at `:584-586`. `a7d908bf` (#2579) touched five docs-site files and never `.claude/skills/archon/`. The weaker sibling at `SKILL.md:210,284-288` is likewise untouched.

**Disposition:** **TRACKED FOLLOW-UP** → #2597, verified still OPEN and still correct. Not this PR's; present on `origin/dev`.

</details>

<details>
<summary><code>R31</code> — the implementation report's commit list, and why it stays FIXED</summary>

Round 9 recorded `R31` FIXED against its stated required outcome while noting the Delivery commit list was three commits behind. The list is now generated from git rather than hand-listed and says so on the line itself — the durable fix is the recurrence mechanism, not the symptom. The artifact necessarily trails head. **Not reopened**: a generated snapshot is stale the moment the next commit lands, the commit list was explicitly optional in round 8's required outcome, and this is a local PRP artifact rather than a shipped file.

</details>

<details>
<summary><code>R25</code> and <code>R30</code> — DECLINED in round 8, not re-opened</summary>

**`R25`** — the server's OpenAPI `WorkflowDefinition` reuses the engine schema verbatim, so the wire type still declares `modelReasoningEffort` while `parseWorkflow` guarantees it absent. `prp-core:type-design-analyzer` rejected the proposed `.omit()` in round 6 and again unprompted in round 8.

**`R30`** — the `#2457` parity guard's failure message branches on whether a warning fired, and the `modelReasoningEffort` fixture now always fires the deprecation warning. Protection is intact and was mutation-proven in round 6.

</details>

<details>
<summary>Closed in earlier rounds — <code>R1</code>–<code>R24</code>, <code>R26</code>–<code>R29</code>, <code>R32</code>–<code>R34</code>, <code>R40</code>, <code>R42</code>–<code>R47</code></summary>

Each was verified in its own round with full evidence recorded in this comment's published history, and re-confirmed unregressed at `2d5c248c` by `bun run validate` (exit 0) and all six green CI checks.

- **`R1`**, **`R2`**, **`R13`**, **`R18`**, **`R19`**, **`R22`** (Important) — all FIXED across rounds 2–7; `R18`'s user-facing sweep re-run with no survivor in rounds 7, 8 and 9.
- **`R4`** (Important) — **NOT A FINDING** across eleven rounds: `packages/workflows/package.json` already declares the dependency and four files already imported the bare root at merge-base.
- **`R23`** (Suggestion) — **NOT A FINDING**; Copilot is unpinnable by construction — `ReasoningEffort` is unreachable through the package's `exports` map.
- **`R3`**, **`R5`**–**`R7`**, **`R9`**–**`R12`**, **`R14`**–**`R16`**, **`R20`**, **`R21`**, **`R24`**, **`R26`**–**`R29`**, **`R32`**–**`R34`** (Suggestions) — all FIXED. Notable: `R20`'s fix was more durable than asked (`AssertNever<T extends never>` against three real installed SDK unions, live-fired in round 7); `R34`'s guard was live-fired in round 9 and bites.
- **`R40`**, **`R42`**–**`R45`** (Suggestions) — FIXED in `9dd43704`, each verified in round 9 against the code.
- **`R46`** (Suggestion) — FIXED in `d1f71c8d`; its own residual gap became `R47`.
- **`R47`** (Suggestion) — FIXED in `8be5fefc`. All six round-9 conjunction forms rejected by the committed equality, reproduced independently twice.

</details>

## Agent Coverage

Round 11 ran all five requested scopes, briefed narrowly against the `7399a4de..2d5c248c` delta with the settled `R1`–`R47` body explicitly out of scope, the standing author decisions listed as non-negotiable, tracked issues named as do-not-report, and a hard prohibition on mutating the worktree. Two agent claims were adjudicated against the tree rather than accepted: the code reviewer's Important severity (downgraded on proof that the defect is byte-identical to `dev`), and the seam/test consensus that the assertion deletions are safe (upheld statically, qualified dynamically).

| Scope | Result |
|---|---|
| code | `R54` (re-severitised). Proved the merge result byte-identical to "dev tip + this PR's local diff" via `git diff 7752ec53:<file> 2d5c248c:<file>`; diffed all six doubly-touched files per side and found disjoint edit regions in each; confirmed all four `capsMock` sites now spread `DEFAULT_PROVIDER_CAPS` and that no `paused approval gate routing` test asserts on effort. Live-fired `scripts/tsconfig.json` in a disposable detached worktree — removing `"DOM"` produces 5 real `tsc` errors from transitively-imported web modules; grep found zero direct DOM usage in `scripts/*.ts`, so the widened lib hides nothing. Live-fired the new parity guard (dropped `minimal` → red). Then found the unwired-test gap by running the exact command CI invokes |
| seams | No new seam finding; `R48` confirmed FIXED, `R53` raised. Established the guard is not self-referential (`EFFORT_OPTIONS` and `EFFORT_LADDER` are independently hand-typed arrays in disjoint packages), located the real tautology at `model-options.test.ts:104` by tracing that `effortOptionsForAgent` returns `EFFORT_OPTIONS` **by reference**, and swept for remaining ladder copies (only the canonical, the now-pinned mirror, and the generated union survive). Explained why the new guard escapes `node-ref-parity.test.ts`'s text-scraping fragility structurally: `OUTPUT_REF_SOURCE` is unexported and must be scraped; all three effort symbols are exported. Noted three un-migrated inline `metadata.isolation === 'container'` checks as sub-finding-grade dev residue |
| tests | `R51` (reopened). Verified the `toBe` literal byte-equal to production **by running** the suite, not reading it. Separated the static and dynamic regimes explicitly and built the concrete literal-sync repair sequence. Found the diagnosis half independently protected at `provider.test.ts:1056` and the rung list not protected anywhere else. Live-fired both tests in the new parity file, including proving the second one fails when `dag-node.ts` re-hardcodes its enum. Ran the merged `orchestrator-agent.test.ts` (235 pass) plus the three sibling core suites in their isolated invocations |
| comments | `R49` item 3 (survives), `R52`. Verified every claim in `3fec518e`'s rewritten block against git objects — including running the literal `git show 9b021e3c` — and every claim in the new `effort-ladder-parity.test.ts` docblock (dependency rule, #2591 rationale, `clampEffort`'s outward walk, `effortLevelSchema`'s derivation, the console pickers). Empirically confirmed the `scripts/tsconfig.json` DOM justification by compiling without it. Confirmed the merge's `afterEach` comment and CHANGELOG resolution against a legacy 3-way `git merge-tree`, which shows a genuine conflict in CHANGELOG and none in the test file — matching the commit message's "only running it does" |
| docs | No findings. Confirmed all three doubly-touched doc files carry both sides in disjoint sections with no contradiction, and CHANGELOG's Breaking → Changed → Fixed ordering is a correct subsequence of `[0.9.0]`'s layout with nothing dropped or duplicated. Determined dev's touch to `workflow-dag.md` is unrelated to #2597, which stays open. Swept `AGENTS.md`, `README.md`, `.claude/docs/`, the manage-run skill and the CLI reference for stale "plain message approves" prose after #2577 — none found. Established that `scripts/node-ref-parity.test.ts` is itself undocumented, so the new sibling guard creating no doc obligation is consistent rather than a gap |

Rounds 1–10 ran up to eight scopes against `701a1b53`, `9b021e3c`, `bc7d380f`, `817f1a8b`, `9dd43704`, `d1f71c8d` and `7399a4de`; their coverage notes are in this comment's published history.

## Validation

|---|---|---|
| `bun run validate` at `2d5c248c` | **PASS** | Exit 0 on the exact merged tree reviewed. `git status --porcelain` empty before the run and after every probe. Zero `error TS` lines; no non-zero `fail` line anywhere in the output |
| **GitHub CI at `2d5c248c`** | **ALL 6 GREEN** | `test (ubuntu-latest)` 3m24s, `test (windows-latest)` 7m39s, `docker-build` 8m51s, `schema-upgrade` 34s, `Build docs site` 38s, `CodeRabbit` — all pass. PR state OPEN, not draft, `MERGEABLE` |
| **windows-latest 5016 ms flake — agreed, not PR-caused** | **AGREED, with a stronger reason** | The A/B (106/104 ms with the new test, 107/114 ms without) is sound but not the decisive argument. Bun's 5000 ms timeout is **per test**, and the failure was in `scripts/migrate-state-dir.test.ts` — a *different file* from the added `effort-ladder-parity.test.ts`, whose own tests run in 28 ms. A 105 ms module-load cost cannot push another file's individual test over a per-test bound. 5016 ms is also the textbook **bimodal** signature `AGENTS.md` documents (a bound meeting the default, as in #2473's `busy_timeout = 5000` and #2306's own dry-run-creates-a-database in this very file), not a gradient. Passed on re-run at the same head |
| **Merge resolution — correct and complete** | **VERIFIED** | All four `capsMock` sites (`:1360`, `:1385`, `:2455`, `:4839`) spread `DEFAULT_PROVIDER_CAPS` (`:158`); every restore is `{ ...DEFAULT_PROVIDER_CAPS }`, byte-matching the module default at `:170`. No test in `paused approval gate routing` (`:2331-2769`) asserts on effort. `git diff 7752ec53:<file> 2d5c248c:<file>` reproduces this PR's pre-merge diff exactly |
| **Other silent semantic conflicts** | **NONE FOUND** | 6 doubly-touched files computed by `comm -12` against merge-base `7fe762a1`; each diffed per side, all edits in disjoint regions. Dev's ~381-line `orchestrator-agent.ts` delta never touches `resolveTitleRequest`/`applyPresetToRequestOptions`/`resolvePresetEffort` |
| `bun test packages/core/src/orchestrator/orchestrator-agent.test.ts` | PASS | 235 pass / 0 fail — this PR's tier-effort assertions and #2577's approval-gate suite green together |
| `bun test ./scripts/effort-ladder-parity.test.ts` | PASS | 2 pass / 0 fail. Reaches CI: root `test` = `bun --filter '*' --parallel test && bun test ./scripts/`; `validate` includes `test` |
| **`R48` guard live-fire** | **BITES, 3× INDEPENDENTLY** | Dropping a rung from the console ladder fails the guard — reproduced by `seam-analyzer`, `pr-test-analyzer` (dropped `xhigh`) and `code-reviewer` (dropped `minimal`, disposable detached worktree). Second test also live-fired: fails when `dag-node.ts` re-hardcodes its enum |
| **Deleted assertions — subsumed?** | **STATICALLY YES, DYNAMICALLY NO** | The `toBe` literal contains the diagnosis phrase and all six rungs, so neither deleted check can fail while it passes. But `toBe` asserts equality with whatever is pasted into it; the deleted loop asserted against the live return value and survives a literal-sync repair. `R51` reopened |
| **`R49` item 3 — corrected?** | **NO, SURVIVES** | `grep -n '9b021e3c\|97ebec7d'` on `options-translator.test.ts` → one hit for `9b021e3c` at `:73`, none for `97ebec7d`. `git show 9b021e3c -- <path>` → **empty diff**. `git show 9b021e3c:<path>` does contain the string, so the substantive claim holds |
| **`R52` — did the merge bring in #2591?** | **NO** | `git merge-base 3fec518e 7752ec53` = `7fe762a1` = #2591. `git merge-base --is-ancestor 7fe762a1 3fec518e` → true. `git log --oneline 7fe762a1..7752ec53` shows only #2577 and #2592 |
| **`R54` — unwired core tests** | **2 FILES, 1 RED** | Script tokenised and diffed against `find src -name '*.test.ts'` (61 files, 55 named, 3 dir globs): `prompt-builder.test.ts` (29 pass) and `schemas/index.test.ts` (18 pass / **2 fail**) unreachable. Failures root-caused by direct `safeParse`: missing `kind`, `parent_run_id`, `output_root` |
| **`R54` — PR-caused?** | **NO, BYTE-IDENTICAL TO DEV** | `git diff --quiet origin/dev HEAD --` clean for `packages/core/package.json`, `schemas/index.test.ts`, `schemas/codebase.ts`, `schemas/workflow-run.ts`, `orchestrator/prompt-builder.test.ts`. `prompt-builder.test.ts` created in `e0716e0a`, confirmed an ancestor of this PR's branch point |
| **`api.generated.d.ts` freshness** | **NOT STALE** | Dev *did* touch a route file this round, unlike round 10. `packages/server/src/routes/api.ts` changes one message string inside a handler body — no route config, no Zod schema. `schemas/workflow-run.ts` adds the `isContainerRun()` function and revises docblocks — no schema shape change |
| **`scripts/tsconfig.json` DOM widening** | **JUSTIFIED, NO EXPOSURE** | Removing `"DOM"` yields 5 real `tsc` errors (`window`, `HeadersInit`) from transitively-imported web modules (`console/lib/http.ts`, `skills/conversations.ts`, `skills/startRun.ts`). Grep across `scripts/*.ts` for `window.`/`document.`/`localStorage`/`navigator.` found **zero** direct usages, so nothing in the scripts themselves is newly hidden |
| `R35`–`R41` sites unchanged since `7399a4de` | VERIFIED | `git diff --quiet 7399a4de 2d5c248c` clean for `options-translator.ts`, `model-validation.ts`, `model-options.ts` and `dag-executor.ts` |
| `bun run check:schema-upgrades` | NOT RUN locally | Covered by the passing `schema-upgrade` CI job at `2d5c248c`; no change to either schema |
| Worktree integrity | **CLEAN THROUGHOUT** | `git status --porcelain` empty before and after every probe, including two temp-file probes and one temp worktree (`git worktree remove --force` + `prune`, list verified). No `git checkout --` was run |
| Prior-round mutation and live-fire evidence | CARRIED | `R47` conjunction matrix, `R34` guard live-fire, `AssertNever` four-direction matrix, translation delete/reverse mutations — all in this comment's published history, unregressed at head |

## Verdict

**READY TO MERGE**

No OPEN Critical or Important finding, `bun run validate` is exit 0 on the exact merged tree, and all six required CI checks are green.

**`R48` closes properly.** The console ladder is now pinned to the engine's by value through direct imports, the guard was live-fired by three agents independently, and I verified it actually reaches CI rather than assuming it. The brief's specific worry — a test that derives from the thing it tests and so cannot detect drift — was well aimed but lands one file away from the guard: `EFFORT_OPTIONS` and `EFFORT_LADDER` are independently hand-typed arrays in disjoint packages, so the parity test is genuinely two-sided. The derivation problem is at `model-options.test.ts:104`, where the assertion now compares an object to itself; the content check it used to carry moved wholesale into the new parity file, so net coverage is preserved (`R53`).

**On the deleted assertions, the honest answer is "statically yes, dynamically no."** The `toBe` literal contains everything both deleted checks asserted, so neither can fail while it passes — the commit's argument is correct for a fixed expected literal. It stops being correct across an edit of that literal: the reflexive repair (paste `actual` into `expected`) satisfies the only remaining content assertion, where the deleted loop, asserting against the live return value, would have stayed red. That is `R51`'s original claim, which round 10 disposed NOT A FINDING only because the loop still existed; it now does not, so `R51` is reopened as a Suggestion. Two specialists called the deletion safe and one made the dynamic case — both are right about different regimes, and the disagreement is preserved rather than resolved away.

**The merge resolution is right and complete, and I found no second silent conflict.** All four `capsMock` sites spread the shared baseline and every restore byte-matches the module default; restoring `effortControl` breaks nothing #2577 depends on, since no test in its gate block asserts on effort. The merged file is provably dev-tip-plus-this-PR's-own-diff. All six doubly-touched files were diffed per side with disjoint edit regions in each. `api.generated.d.ts` is still not stale — and unlike last round that needed checking, because dev did touch a route file this time; it changes one message string inside a handler.

**The most valuable thing this round found is not this PR's, and does not block it.** `packages/core`'s enumerated `test` script cannot reach two of its 61 test files, and one of them is **red right now** — `schemas/index.test.ts` has been failing since #2055 added `kind`, invisible because CI never runs it. The other holds the only real coverage of #2577's own Breaking change. All five relevant files are byte-identical to `origin/dev`, so this PR neither causes nor worsens it, and blocking on it would fix nothing. Recorded on the existing issue #2401, which already predicted exactly this generalisation and names `@archon/core`.

**`R49` does not fully close**: three of four items fixed, but the `git show 9b021e3c` citation survives verbatim in a block `3fec518e` otherwise rewrote — a reader running it gets an empty diff. `R52` adds one more of the same class: the merge commit credits itself with #2591, which is its own merge-base.

Ten Suggestions remain and none blocks. `R35`, `R36` and `R41` are pre-existing observability conditions this PR surfaced rather than caused; `R37` carries a recorded reviewer disagreement; `R38` is a counter-proposal against the author's `AssertNever` preference; `R39` is narrowed to cosmetic. All six sites verified untouched since round 10. `R25` and `R30` stay DECLINED on round 8's independent specialist agreement.

**I agree the windows-latest flake is not PR-caused, and for a stronger reason than the A/B timing.** Bun's 5000 ms timeout is per test, and the failure was in a *different file* from the one this PR adds; a 105 ms module-load cost cannot push another file's individual test over a per-test bound. 5016 ms is also the bimodal bound-meeting-the-default signature `AGENTS.md` documents — and `migrate-state-dir.test.ts` is the exact file #2306 is about.

Two process notes. **The head did not move during this round**, so every result is pinned to one tree. **No agent mutated the worktree**: `git status --porcelain` was empty before and after every probe, the one temporary worktree was removed and pruned, and no `git checkout --` was run anywhere.

`R8` (#2584), `R17` (#2586), `R50` (#2597) and `R54` (#2401) remain tracked and verified OPEN.

Wirasm:
One request while this is open, because it is cheap here and expensive later.

`modelReasoningEffort` ends up deprecated, warned, and **kept** — `workflowBaseSchema:156` still carries it beside `effort:159`, and the loader (`loader.ts:759-778`) warns without translating. So after this lands it is still a workflow-level field with no node-level counterpart, which is precisely the condition that blocks the workflow-level → node-level collapse in #1764. A deprecation that never terminates isn't a deprecation, it's a second spelling with a warning attached — which is the thing this PR exists to remove.

**Translate it at parse time.** When `modelReasoningEffort` is present and `effort` is absent, set `effort` from it, then warn as you already do. That is a few lines inside the block that already computes the warning, and it means only `effort` survives parsing — a node field, so the collapse has nothing left to special-case.

**The both-set case needs a decision, and it's the part the simple version doesn't close.** Translating only when `effort` is absent leaves `modelReasoningEffort` alive whenever an author sets both — and your own warning text says it "still overrides a workflow-level `effort:` on Codex nodes". So in that case the field remains in the collapse's problem space and #1764 still has to know about it. Two honest options:

1. **Translate-when-absent only.** Smallest change, no precedence change. The hole closes for every author who followed the deprecation and shrinks to only those who set both. #1764 still carries a special case, just a rarer one.
2. **Translate-when-absent, and let `effort:` win when both are set.** The field disappears from the problem space entirely. This *is* a precedence change on Codex nodes — currently the deprecated field wins — so it has to be stated in the PR body and the CHANGELOG rather than slipped in. Given the deprecated field is the one being removed, having it lose to its replacement is the defensible end state.

I'd take (2) and say so out loud, but (1) is a legitimate call if you'd rather not move precedence in the same PR that introduces the unification. Either way, the current text should stop implying the migration is complete.

**`webSearchMode` — accept the hole and document it.** It is Codex-only, genuinely run-shaped rather than node-shaped, and there is exactly one of it. The docs here already reaffirm "workflow-level only — there is no per-node form", which is the right answer; it just needs to be an acknowledged exception in #1764's terms rather than an oversight discovered later. One sentence naming it as the single remaining workflow-level-only field would do it.

Not blocking the substance — `effort:` as the one spelling across every provider that has the control is the right shape, and resolving it however the provider was chosen (node `provider:`, tier, or `@alias`) is the part that actually makes it authorable.

PR #2583, opened 2026-08-17
feat(workflows): a deterministic loop no longer has to invent a completion signal [hold: Part B stacks on this]

Wirasm:
## Problem and outcome

`loop.until` was required on every loop while `until_bash` was optional, so an author whose completion condition is fully checkable still had to invent a prose sentinel — one that stays live and can false-positive if the model emits that string while reasoning about its own exit criteria. The repo had already documented the wart against itself, twice:

- `.archon/workflows/test-workflows/e2e-joins.yaml` carried `until: __NEVER_SIGNALLED__` under a four-line comment: *"it is dead config the author is forced to write. Filed as a wart."* Its body is pure bash and emits no model text, so the signal could never fire.
- `.claude/skills/archon/references/parameter-matrix.md:57` — the skill Archon ships to its own agents — taught `loop: {until_bash: "bun run test", max_iterations: N}`, a shape the schema rejected. The docs were right and the code was wrong.

- **Outcome:** the completion gate is *at least one of* `until` / `until_bash`. A loop that declares only the deterministic check has no prose-matching path at all, so nothing the model says can end it early.
- **Invariant:** a loop terminates only through a channel its YAML declared, and a loop that declares none fails at load rather than silently running to `max_iterations`.
- **Scope boundary:** `until`'s matching semantics are untouched (all three formats in `detectCompletionSignal` are unchanged), `until_bash` does not become authoritative over the signal, and no bundled workflow is rewritten. Part B of #2563 — `output_format` + a structured completion field — is a separate PR on an approved design; this one is shippable and useful without it.

## Review guidance

- **Feedback requested:** correctness of the `undefined` handling at every reader of `loop.until`, and whether the `until_bash` short-circuit alignment is the right direction to reconcile.
- **Start here:** `packages/workflows/src/schemas/loop.ts` — `loopControlSchema` is the whole behavioral change; everything else follows from `until` becoming optional.
- **Review order:**
  1. `schemas/loop.ts` — the relaxed gate and its `superRefine`.
  2. `dag-executor.ts` — `executeLoopNode` and `executeLoopGroupNode`: guarded signal detection, the short-circuit, and `buildHonestGateMessage`.
  3. `executor-shared.ts` — `describeUnmetCompletion`, the one shared helper, so the two max-iterations messages cannot drift.
  4. `dry-run.ts` — the simulator executes nothing, so it has to say what it could not evaluate.
  5. The web builder (four files) — see *Known risk* below.
- **Lower-attention areas:** the docs and bundled-skill edits. `check:bundled-skill` verifies file presence only, so skill content needs no regeneration; `bun run validate` confirms every generated artifact is still in sync.
- **Known risk or uncertainty:** the `WireDagNode` override in `builder/types/wire.ts`. `api.generated.d.ts` still marks `loop.until` required, and regenerating it needs a running server and is not part of `bun run validate` — so the relaxation is expressed as a local type override in the file whose docblock already declares itself the home for generated-spec drift. It becomes redundant (not wrong) after the next regeneration, and the comment says to delete it then. If you would rather regenerate now, say so and I will.

## Solution

`until` becomes `.optional()` and `loopControlSchema`'s existing `superRefine` gains an at-least-one-channel check. Both loop variants share that schema, so `loop:` and `loop_group:` get the relaxation together.

Every reader of the signal then has to tolerate its absence, and there were four:

- **Signal detection** — `detectCompletionSignal(output, signal)` builds regexes from the signal, so calling it with `undefined` would build them from the empty string and match anything. Both call sites are now guarded by `until !== undefined` rather than by a default value.
- **`buildHonestGateMessage`** — printed `` `undefined` `` back at the reviewer. It now says *"Completion condition met / not met"* when there is no signal to name, and keeps its existing wording verbatim when there is.
- **Max-iterations failure** — the two variants each hard-coded `without completion signal '<until>'`. They now share `describeUnmetCompletion`, which names the channels the loop actually declared. One helper rather than two near-identical strings in two functions the issue explicitly asked to keep in agreement.
- **The dry-run simulator** — executes nothing, so `until_bash` is unobservable. A deterministic-only loop would have simulated as a max-iterations failure a real run would never produce. It now completes on the first stubbed iteration with a reason naming the channel it could not evaluate.

While reconciling the variants, `loop:` now skips `until_bash` on an iteration whose signal already fired, which is what `loop_group:` has always done. This is the divergence #2563 asked to settle, and the group's behavior is the correct one: completion is an OR, so the outcome cannot change, but a side-effecting check runs one time fewer.

**The console builder needed the same relaxation to keep working.** `structural.ts` called `node.data.until.trim()` against a required `LoopNodeData.until`, so importing an `until`-less loop would have thrown. The fix mirrors the engine rule (at least one channel; a *present-but-blank* `until` is still a violation, since that is an unfinished edit rather than a deliberate deterministic loop), and `loopFromDag`/`loopToDag` now use the existing `ifDefined` helper so an absent signal never round-trips back as an explicit `until: undefined`. `LoopFields.tsx` follows the pattern `until_bash` already used — empty box means the field is absent — which also keeps the input controlled.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A loop had to declare `until`; a deterministic loop carried a live prose matcher it did not want | A loop declares at least one of `until` / `until_bash`; a deterministic loop matches no prose at all |
| **`until_bash` on a signalled iteration** | `loop:` ran it, `loop_group:` skipped it | Both skip it |
| **Failure behavior** | A loop with no channel was impossible to write | A loop with no channel is a load error naming both fields; a loop that exhausts `max_iterations` reports the channels it actually declared instead of `undefined` |

## Validation

- `bun run validate` — exit 0, every stage green, no failing suite in the log — proves types, lint, format, bundled-artifact freshness, and the full per-package suite across the change.
- `bun test src/dag-executor.test.ts` (`packages/workflows/`) — 462 pass — three new cases carry the behavior:
  - a loop declaring only `until_bash` completes when the script exits 0 (asserted through a real counter file, not a mock);
  - **a loop with no `until` does not complete when the model emits `<promise>COMPLETE</promise>` while describing its own exit criterion** — it runs the second iteration and ends on `until_bash`. This is the false-positive the change exists to remove;
  - a `loop:` node whose signal fired does not execute `until_bash`, proven by a sentinel file the script would have created (mirrors the existing `loop_group` "EDGE G" test).
- `bun test src/schemas.test.ts` / `src/loader.test.ts` — the new channel rule accepts `until_bash`-only for both variants, still accepts `until`-only, rejects an empty-string `until`, and rejects a loop with neither while naming both fields.
- `bun test src/dry-run.test.ts` — a deterministic-only loop simulates as complete with the channel-naming reason rather than as a max-iterations failure.
- `bun --filter @archon/web test` — 417 pass — a wire loop with `until_bash` and no `until` round-trips byte-identically and introduces no `until` key; structural validation reports the missing-channel violation only when both are absent.
- **Not verified:** no live end-to-end run against a real provider. Every changed path is covered by the executor suite driving the real `executeLoopNode` / `executeLoopGroupNode` against real `bash -c` subprocesses, and the schema change is load-time, so a live run would exercise nothing the suite does not.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Relaxing a required field is additive — every existing workflow declares `until` and is unaffected. The one behavior change is the `until_bash` short-circuit on `loop:`; `until_bash` output is never captured, so only its side effects change. No migration. | Full suite green with every bundled and test workflow loading; `e2e-joins.yaml` is the only workflow edited, to drop the decoy |
| Rollout / rollback | Revert restores a required field, which cannot break a workflow written before this PR. | Single commit, no persisted state |
| Documentation / communication | `loop-nodes.md` gains a *choosing a completion channel* table and corrects a claim that was already false — "this prevents the AI from falsely claiming completion when tests still fail" was untrue under OR semantics, and now says so and shows the shape that actually enforces it. Skill references and `quick-reference.md` updated so the agent-facing docs and the schema finally agree. | Included in this PR |

## Links

- Refs #2563

## Summary by CodeRabbit

* **New Features**
  * Loops and loop groups can complete using either prose or shell conditions.
  * Deterministic loops can operate without a prose completion signal.
  * Shell checks are skipped when a prose signal completes the iteration first.
  * The workflow builder preserves optional completion settings during import and export.

* **Bug Fixes**
  * Improved validation and error messages for missing or blank conditions.
  * Dry runs clearly report assumed completion when shell checks cannot execute.

* **Documentation**
  * Updated loop guidance, examples, quick references, and changelog entries.


Wirasm:
---
pr: 2583
base: dev
head: feat/loop-completion-2563
reviewed: 2026-08-17T21:55:00Z
reviewed_head: 0ef9e1e237627095b8f415d05fb7ad803e545da6
verdict: READY TO MERGE
open_findings: 1
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2583#issuecomment-5317208782
---



# PR Review: #2583 — feat(workflows): a deterministic loop no longer has to invent a completion signal

## Signal

Round 5 at `0ef9e1e2`, verifying three dispositions. **The code surface did not move.** The delta from `ae4c9ae9` is one commit touching two files — `CHANGELOG.md` (one sentence) and `approval-nodes.md` (three lines) — and that claim is established by tree hashing, not by `--stat`: `git rev-parse ae4c9ae9:packages/workflows` and `0ef9e1e2:packages/workflows` return the same tree object `84c17f67…`, `packages/web` the same `e543c377…`, and `git ls-tree -r | wc -l` is 1374 at both commits. Every behavioral claim the earlier rounds cleared therefore still holds by construction, and both reviewers re-verified the important ones by execution anyway.

**`R6` is FIXED and exact.** The doc now names both message forms, and the match to `buildHonestGateMessage` is byte-level rather than approximate: the signal-absent branch emits the literal `✅ Completion condition met.` and the doc quotes it verbatim *including* the trailing period; the signal-present branch emits `` ✅ Completion signal detected (`<until>`). `` and the doc correctly says the message "leads with" that prefix, omitting the trailing content the code appends (`dag-executor.ts:3257-3264`). Both call sites pass the node's own `until` — `:3877` for `loop_group`, `:5113` for `loop` (the brief had these two labels swapped; a labeling detail, not a defect) — so "declares no `until` signal" in the doc maps precisely to `untilSignal === undefined`. The two `⚠️` variants are correctly absent because the paragraph describes only the finalize-eligible case. And nothing parses either string: a tree-wide grep returns the two emitter templates, two test assertions, and the doc — every production consumer reads the persisted `metadata.approval.completionSignaled` boolean (`manage-run-tool.ts:278,337`, `console/primitives/run.ts:159-160`, `ApprovalPanel.tsx:45`, `cli/commands/workflow.ts:2408`, `dag-executor.ts:3454,4235`).

**`R5`'s deferral is adequate, and I verified the carrier rather than accepting the claim.** The guard exists and is pushed: `scripts/node-ref-parity.test.ts:368`, `describe('loop completion-channel parity')`, on `feat/loop-until-field-2563`. It imports `dagNodeSchema` and `validateStructural` (`:54-55`) and executes both encodings over a 24-case corpus, comparing verdicts — not case names. The corpus covers all three channels including this PR's two, and specifically includes the blank-beside-valid shapes (`{until:'   ', until_bash:'bun run test'}`, `{until:'COMPLETE', until_bash:'   '}`) that were `R1`'s real drift. The stacked branch runs 72/72. Placing it there is strictly better than placing it here: here it would cover two channels and Part B would immediately extend it; there it covers three at once. Combined with the documented hold, `dev` never receives this PR's channel rule without the guard. Recorded `TRACKED FOLLOW-UP` rather than `FIXED` only because the correction is not at *this* head.

**`R8` is FIXED, and its replacement is materially more accurate — but a new, narrower finding replaces it.** "Every existing workflow is unaffected" no longer contradicts the sentence after it. The code reviewer then tested the replacement's stronger claim and found a reachable counterexample, recorded as **`R9`**: an **interactive** `loop:` declaring **both** channels with a **state-accumulating** `until_bash` can now take one more iteration — and one more human approval round-trip — to complete. The chain is decisive from source: `if (loop.until_bash && !signalDetected)` (`:4930`) skips the script on a signaled iteration; `if (completionDetected && (!interactiveFirstRun || signalCompletes))` (`:5041`) means an interactive first run *gates* on that iteration instead of completing; approving with feedback text runs another iteration, whose `until_bash` now evaluates against state one increment short of where the pre-PR engine left it. The reviewer reproduced it end to end. So "terminates unchanged" is true under the reading *"still terminates through the same channel"* and imprecise under *"terminates at the same iteration."*

It stays a Suggestion, and the reason is evidential rather than charitable. It needs a five-way conjunction — `loop:` (not `loop_group:`, which has always behaved this way), `interactive: true`, `signal_completes` unset, both channels declared, an `until_bash` that mutates cross-iteration state, and the user approving *with* text rather than bare. Not one workflow in the repo comes close: the only both-channel workflow is `archon-adversarial-dev :: adversarial-sprint`, whose `until_bash` is a read-only `grep -qE` (`archon-adversarial-dev.yaml:302-303`) on a non-interactive node, and the file contains no `interactive` key at all. The engine behavior itself was cleared across four rounds and is the alignment #2563 asked for. This is release-note precision on a sentence the author already tightened once, not a defect.

Independent re-verification found nothing regressed. The seam analyzer built a fresh 120-check cross-product — 80 through `dagNodeSchema` and the real `parseWorkflow()` loader for both `loop:` and `loop_group:`, 40 through the builder's `validateStructural` + `loopFromDag`/`loopToDag` round-trip — with **0 disagreements and every padded/multiline value surviving byte-for-byte**, reproducing the prior round's conclusion without reusing its harness. Both short-circuit tests pass (`dag-executor.test.ts:6977` for `loop:`, `:15533` "EDGE G" for `loop_group:`). `bun run validate` is exit 0 in a detached worktree at this head with `0 fail` on all 139 suite summaries, and all five CI checks are green.

- **Blocking:** 0
- **Non-blocking:** 1 open (`R9`)
- **Resolved since previous review:** 3 — `R6` and `R8` FIXED at `0ef9e1e2`; `R5` moved OPEN → TRACKED FOLLOW-UP on verified carrier evidence
- **Tracked follow-ups:** 2 — [#2589](https://github.com/coleam00/Archon/issues/2589) (`R3`) and [#2588](https://github.com/coleam00/Archon/pull/2588) (`R5`, implemented and pushed)
- **Validation:** `bun run validate` exit 0 at `0ef9e1e2`, 139/139 suites `0 fail`; all five CI checks green; `mergeStateStatus: CLEAN`; 120-check cross-encoding matrix with 0 disagreements

## Findings

|---|---|---|---|
| `R1` | Important | The whitespace rule was enforced only on the aggregate gate, so a blank `until` or `until_bash` still loaded whenever the other channel was valid | FIXED |
| `R2` | Suggestion | `packages/web/src/lib/api.ts` is a second derivation of the generated spec and still types `loop.until` as required | DECLINED |
| `R3` | Suggestion | The schema's sync comment names two sync points; a third (`variants/loop.ts`) silently deletes any loop key it does not model | TRACKED FOLLOW-UP |
| `R4` | Suggestion | The builder's mirror inherited the schema's rules but not its type check — `until?.trim()` throws on a non-string and accepts `null` | DECLINED |
| `R5` | Suggestion | Both sides of the channel rule instruct "verify by parsing both, never by reading them" — and neither does; the repo-level guard that does exactly that skipped this copy | TRACKED FOLLOW-UP |
| `R6` | Suggestion | `approval-nodes.md` told readers the loop gate message "leads with `✅ Completion signal detected`"; an `until_bash`-only interactive loop emits `✅ Completion condition met.` | FIXED |
| `R7` | Critical (claimed) | CodeRabbit: duplicate `const` declarations make `schemas.test.ts` uncompilable | NOT A FINDING |
| `R8` | Suggestion | CodeRabbit: the CHANGELOG's "Every existing workflow is unaffected" is contradicted by the behavior change named in its own next sentence | FIXED |
| `R9` | Suggestion | The replacement claim "still loads and terminates unchanged" is exact for every workflow that exists, but a five-condition configuration can now terminate one iteration later | OPEN |

## Detailed Findings

<details>
<summary><code>R1</code> — the trim guarded the aggregate, not each field; two blank shapes still loaded and the comment claimed they did not</summary>

**Impact (as of round 2):** `until: " "` beside a real `until_bash` loaded on the engine while the builder mirror rejected it — the original encoding disagreement — and the blank signal, matched verbatim by `detectCompletionSignal`, false-fires on any whitespace-only line, which under this PR's `!signalDetected` short-circuit would then skip the deterministic check entirely. Separately, `until: "COMPLETE"` beside `until_bash: "   "` loaded on both encodings, and `bash -c "   "` exits 0, completing the loop on iteration 1.

**Evidence that closed it:** `packages/workflows/src/schemas/loop.ts:59-113` applies `.refine(isNonBlank, …)` to each field independently and keeps the across-channels `superRefine` testing presence only. `packages/web/src/experiments/console/builder/validation/structural.ts:98-115` carries the matching per-field rules plus the aggregate.

Round 5 re-verification at this head, by execution and independently of round 4's harness:

- Seam analyzer, fresh 120-check matrix: 5 `until` variants (absent / real / padded / blank / tab-blank) × 4 `until_bash` variants (absent / real / multiline / blank) through `dagNodeSchema.safeParse` **and** the real `parseWorkflow()` YAML loader for both `loop:` and `loop_group:` → 80 checks, **0 disagreements**; the same 20 combinations through `validateStructural` + `loopFromDag`/`loopToDag` → 40 checks, **0 disagreements**, values surviving verbatim through the round trip.
- Code reviewer: `bun test src/schemas.test.ts` → 133 pass, including `'a declared channel is validated but never rewritten'`; `bun test src/experiments/console/` in `packages/web` → 439 pass.
- The code surface is provably byte-identical to the head round 4 cleared (`git rev-parse ae4c9ae9:packages/workflows` = `0ef9e1e2:packages/workflows` = tree `84c17f67…`), so round 4's mutation pinning of both rules transfers without re-execution.

**Required outcome:** None — closed.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer` (rounds 1 and 2); closure confirmed rounds 3–5

**Disposition:** FIXED at `284c8332`, preserved byte-identically through both rebases and the docs commit to `0ef9e1e2`. ID preserved from rounds 1–4.

</details>

<details>
<summary><code>R2</code> — the generated-spec relaxation lands in one of two derivations</summary>

**Impact:** Latent type inaccuracy. `@/lib/api`'s `DagNode` asserts `loop.until` is always a `string` while the live server can return a node without it. No runtime defect today.

**Evidence:** `packages/web/src/lib/api.ts:9-10` re-derives `WorkflowDefinition`/`DagNode` from `components['schemas']`, and `api.generated.d.ts` still declares `loop.until` required. Re-checked at this head: `bun run type-check` exits 0 in all packages, and the `wire.ts` `Omit`-based relaxation compiles against the un-regenerated spec. Round 4 additionally proved by execution that registering `dagNodeSchema` on an `OpenAPIHono` app and calling `getOpenAPIDocument()` emits `loop.required = ["max_iterations"]` — so a regeneration drops `until` from `required` exactly as the Part B plan assumes.

**Required outcome:** None for this PR.

**Disposition:** DECLINED, unchanged from rounds 2–4. The recorded plan — regenerate `api.generated.d.ts` in Part B and delete the `wire.ts` override as its own docblock instructs — is concrete, and the hold means this PR never reaches a merged state carrying the override alone.

</details>

<details>
<summary><code>R3</code> — the sync comment names two sync points; the third silently deletes fields</summary>

**Impact:** A future author who follows the schema docblock literally gets a channel that validates correctly on both encodings and is then silently deleted the first time a user opens that workflow in the console builder and saves it — no warning in the UI, no import issue, and the re-saved YAML carries a different completion condition than the author wrote.

**Evidence:** `packages/workflows/src/schemas/loop.ts:45-48` at the merge-base named only `builder/validation/structural.ts`. The third surface is `packages/web/src/experiments/console/builder/variants/loop.ts` (`loopFromDag`/`loopToDag`), a hand-maintained field list; `model/from-workflow.ts` warns on unsupported keys but walks only top-level `WireDagNode` keys, so anything nested inside `loop: {…}` is opaque to it. Demonstrated on `dev` with `signal_completes`, a real `loopControlSchema` field the builder has never modelled: dropped on round-trip with `issues: []`.

**Required outcome:** None for this PR's stated outcome.

**Disposition:** TRACKED FOLLOW-UP, unchanged from round 4. The in-PR half is carried by the stacked [#2588](https://github.com/coleam00/Archon/pull/2588), where the `loop.ts` docblock names both builder files and states how they fail differently, and `variants/loop.ts` carries `until_field` through both converters. The general mechanism — a nested key dropped with no warning, applying equally to `approval.on_reject`, `retry`, `fan_out` and `pi` — is filed as [#2589](https://github.com/coleam00/Archon/issues/2589).

</details>

<details>
<summary><code>R4</code> — the builder's mirror inherited the schema's rules but not its type check</summary>

**Impact:** `until: 42` / `['X']` / `true` throw `TypeError: node.data.until?.trim is not a function` out of `validateStructural`; `until: null` is silently accepted, satisfying neither the blank check nor the at-least-one-declared check. The engine rejects all of these cleanly.

**Evidence:** `packages/web/src/experiments/console/builder/validation/structural.ts:106,108`. Reachability is nil: `loader.ts` is the sole `dagNodeSchema.safeParse` call site and the only path producing `result.workflow` for `GET /api/workflows/:name`, so both fields are guaranteed strings before reaching the client. At the merge-base the same line read `node.data.until.trim()`, which threw on a non-string **and** on `null` — so the non-string throw is pre-existing.

**Required outcome:** None.

**Disposition:** DECLINED, unchanged from rounds 3 and 4 and not re-raised. No second, unvalidated route into `WireDagNode` exists, so a guard here would be defense in depth against a state nothing can produce — which `AGENTS.md` rules out under YAGNI and "avoid speculative defense in depth". If such a route ever appears, the guard belongs at that route.

</details>

<details>
<summary><code>R5</code> — both encodings say "verify by parsing both, never by reading them"; neither does, and the repo-level guard that does exactly that skipped this copy</summary>

**Impact:** A future channel added per the schema's own instruction could validate on the engine, load from YAML, run correctly — and be rejected by the console builder, with the full test suite green. Round 4 proved that silence by mutation: adding a third channel to the engine exactly as `loop.ts:45` instructs left **678 tests green** across the three suites that could plausibly police it, while the encodings had already diverged.

**Evidence at this head:** unchanged from round 4. The instruction is on both sides (`schemas/loop.ts:45-48` and `structural.ts:98-101`); what ships in this PR is twin matrices with identical case names (12/12), which is prose — the mechanism `scripts/node-ref-parity.test.ts:19-34` records failing twice (#2567, #2591).

**Round 5 verification of the disposition** — the carrier was checked at its source, not accepted on the author's word:

- `git show origin/feat/loop-until-field-2563:scripts/node-ref-parity.test.ts` contains `describe('loop completion-channel parity', …)` at `:368`, importing `dagNodeSchema` (`:54`) and `validateStructural` (`:55`).
- The corpus (`:383-413`) is a `Channels` cross-product over `until`, `until_bash`, **and** `until_field` — 24 cases including every blank-beside-valid shape that was `R1`'s real drift (`{until:'   ', until_bash:'bun run test'}`, `{until:'COMPLETE', until_bash:'   '}`) plus the padded/multiline entries that pin "trims to decide, never rewrites what it stores".
- It compares **verdicts by execution**, not case names, and the seam analyzer ran the file on that branch: 72 pass, 0 fail. The author additionally mutation-verified it (dropping `until_field` from the builder rule fails on exactly `agrees on {"until_field":"done"}`).

**Judgment on the deferral:** adequate, and better than the alternative. Placing the guard here would cover two channels that Part B extends the next commit; placing it in Part B covers all three at once, in the same change that supplies the third. The two PRs merge together under a documented hold, so `dev` never receives this PR's channel rule without the guard covering it. Both "verify by parsing both" instructions now name the guard on the Part B branch.

**Required outcome:** None for this PR. The correction exists, is executed, and merges with this change.

**Disposition:** TRACKED FOLLOW-UP, moved from OPEN. Carrier is [#2588](https://github.com/coleam00/Archon/pull/2588) (open, implemented, verified above). Recorded as tracked rather than FIXED only because the correcting code is not at this head; the merge-together hold is what makes the gap unreachable on `dev`.

</details>

<details>
<summary><code>R6</code> — a docs sentence describing the loop gate message was narrower than the message this PR can produce</summary>

**Impact (as raised):** `approval-nodes.md:172-174` told readers the persisted gate message for a finalize-eligible interactive loop "leads with `✅ Completion signal detected`". An interactive loop declaring only `until_bash` emits `✅ Completion condition met.` instead, so a reader recognizing the finalize case by that string would not find it.

**Evidence that closed it:** `approval-nodes.md:175-177` now reads *leads with "✅ Completion signal detected", or "✅ Completion condition met." for a loop that declares no `until` signal*. Verified against the emitter at byte level, not by eye — `packages/workflows/src/dag-executor.ts:3257-3264`:

- `untilSignal === undefined` + detected → the literal `'✅ Completion condition met.'`; the doc quotes it verbatim **including the trailing period**.
- `untilSignal` defined + detected → `` `✅ Completion signal detected (\`${untilSignal}\`).` ``; the doc says the message "leads with" that prefix and correctly omits the appended signal name and period.
- The two `⚠️` (not-detected) variants are correctly absent — the paragraph describes only the bare-approve-finalizes case.
- Both call sites pass the node's own field: `:3877` (`executeLoopGroupNode`, `group.until`) and `:5113` (`executeLoopNode`, `loop.until`), both typed `string | undefined` by `loopControlSchema:56-58`, so "declares no `until` signal" maps exactly to `untilSignal === undefined`.

The contract is untouched: `grep -rn "Completion signal detected\|Completion condition met"` over `.ts`/`.tsx`/`.md`/`.yaml` returns only the two emitter templates, two test assertions (`dag-executor.test.ts:7421`, `:15837`), and this doc sentence. Every production consumer reads the persisted boolean `metadata.approval.completionSignaled` (`manage-run-tool.ts:278,337`, `console/primitives/run.ts:159-160`, `ApprovalPanel.tsx:45`, `cli/commands/workflow.ts:2408`, `dag-executor.ts:3454,4235`). The `⏸` prefix the doc references (`approval-nodes.md:176`) is byte-identical U+23F8 to both emitters (`dag-executor.ts:3882` literal, `:5118` escaped) — no variation-selector drift.

**Found by:** `prp-core:code-reviewer` (rounds 4 and 5), `prp-core:seam-analyzer` (round 5, doc↔code seam)

**Disposition:** FIXED at `0ef9e1e2`. The finding's own round-4 correction also stands: the sentence is pre-existing prose from `92b9bba9` (#2126), not something #2577 added.

</details>

<details>
<summary><code>R7</code> — CodeRabbit's "this test file cannot compile"</summary>

**Impact (claimed):** CodeRabbit posted an inline review comment on `packages/workflows/src/schemas.test.ts:905` marked *🎯 Functional Correctness · 🔴 Critical*, asserting `const grp` is declared three times in the same block (and `const node` likewise at `:954` and `:1070`), that "TypeScript rejects each block with a redeclaration error", and therefore "this test file cannot compile."

**Evidence that disproves it:** every one of those declarations sits inside its own `test(…, () => { … })` arrow-function body — `schemas.test.ts:894-909`, `:943-958`, `:960-970`, `:1055-1074`. Separate function scopes; `const` in sibling blocks is not a redeclaration. Refuted by compilation, not argument: the code reviewer built a program explicitly including the file and got **zero `TS2451`** errors; the author independently reproduced with `npx tsc --noEmit -p packages/workflows/tsconfig.json`. The only diagnostic was `TS6133` on a pre-existing unused import at line 42, present verbatim on `dev`. `bun test src/schemas.test.ts` → 133 pass, and `bun run validate` (which includes `type-check`) is exit 0 at this head.

**Required outcome:** None. Dismiss the comment.

**Found by:** `coderabbitai[bot]` (inline review comment); refuted by `prp-core:code-reviewer`

**Disposition:** NOT A FINDING, unchanged. Recorded rather than ignored because it is published on the PR as Critical, and a reader who does not check it may believe the branch is broken. The author dismissed it publicly in [comment 5320375838](https://github.com/coleam00/Archon/pull/2583#issuecomment-5320375838).

</details>

<details>
<summary><code>R8</code> — the CHANGELOG entry's compatibility claim was contradicted by its own next sentence</summary>

**Impact (as raised):** `CHANGELOG.md:16` stated "Every existing workflow is unaffected." and then, in the following sentence, "One behavior change comes with it: a `loop:` node now skips `until_bash` on an iteration whose signal already completed it." A workflow declaring both channels *is* affected. Read as a whole the entry was complete; read as a sentence the claim was not true.

**Evidence that closed it:** `CHANGELOG.md:16` now reads "Every existing workflow **still loads and terminates unchanged**." The self-contradiction is gone — the sentence no longer asserts the absence of the change the next sentence names, and the entry as a whole is coherent. Confirmed under `## [Unreleased]` → `### Changed` (`CHANGELOG.md:8,14`), and neither the entry nor `git log -1 --format='%B' 0ef9e1e2` carries AI attribution.

**Found by:** `coderabbitai[bot]` (inline review comment on `CHANGELOG.md:16`)

**Disposition:** FIXED at `0ef9e1e2`. The replacement text's own residual precision is a distinct finding, recorded separately as `R9` rather than folded here, so the fix is not misrepresented as not having landed.

</details>

<details>
<summary><code>R9</code> — "terminates unchanged" is exact for every workflow that exists, and inexact for one constructible configuration</summary>

**Impact:** Changelog precision. The replacement sentence asserts an absolute — every existing workflow "still loads and **terminates** unchanged" — and the trailing clause bounds the consequence to "only a side-effecting check runs one time fewer". A five-condition configuration falsifies the strict reading: the loop still terminates through the same channel, but one iteration later, costing one additional human approval round-trip.

**Evidence (causal chain from source, plus an executed reproduction):**

- `packages/workflows/src/dag-executor.ts:4930` — `if (loop.until_bash && !signalDetected)`, new in this PR; base `dev` (`7752ec53:…/dag-executor.ts:4908`) ran it unconditionally.
- `:5041` — `if (completionDetected && (!interactiveFirstRun || signalCompletes))`. With `interactiveFirstRun` true (`:5039`, `loop.interactive && !isLoopResume`) and `signal_completes` unset, a signaled iteration **gates instead of completing**.
- The gate's documented semantics (`approval-nodes.md:177-179`) are that approving *with* a comment runs another iteration. `until_bash` receives `$STATE_DIR` and `$ARTIFACTS_DIR` through `substituteWorkflowVariables` (`:4938-4950`), so a cross-iteration accumulator is authorable with documented primitives.
- Consequence: iteration 1 signals → `until_bash` skipped (post-PR) where it previously ran and incremented → gate → user feedback → iteration 2's `until_bash` evaluates against state one increment short → threshold unmet → a further gate and iteration. The code reviewer executed this as a two-invocation test: counter absent after iteration 1, `1` (not `2`) after iteration 2, node pauses again where the pre-PR engine completed.
- The code comment at `:4924-4930` ("completion is an OR, so the outcome cannot change") is true of *that iteration's own* completion decision and silent about a later iteration reading state the skipped run would have written.

**Why this is a Suggestion, not a blocker** — each point checked rather than assumed:

1. **No shipped-code defect.** The short-circuit is the alignment #2563 asked for and was cleared across four rounds; `loop_group:` has always behaved this way, so the semantics are not novel to Archon.
2. **Zero in-repo instances.** `grep -rl until_bash --include="*.yaml"` returns exactly two workflows. `archon-adversarial-dev :: adversarial-sprint` is the only one declaring both channels, its `until_bash` is a read-only `grep -qE '"status"…' "$ARTIFACTS_DIR/state.json"` (`:302-303`), and the file contains no `interactive` key at all. `e2e-joins :: countdown` declares `until_bash` alone, so the guard is unreachable for it.
3. **The conjunction is five-deep** — `loop:` (not `loop_group:`), `interactive: true`, `signal_completes` unset, both channels declared, an `until_bash` that mutates cross-iteration state rather than testing a condition, and a user who approves *with* text (a bare approve finalizes and produces no divergence).
4. **The entry already points at the mechanism** — a reader with such a workflow is told the exact thing that changed, in the same paragraph.

**Required outcome:** None required. If wanted, one clause: e.g. "…every existing workflow still loads and completes through the same channel; where `until_bash` accumulates state across iterations, an interactive loop may reach its threshold one iteration later." The PR is held pending Part B, so the changelog paragraph is still editable at no cost.

**Found by:** `prp-core:code-reviewer` (raised as Important; recorded as a Suggestion after independently confirming the causal chain at `dag-executor.ts:4930`/`:5041` and establishing that no workflow in the repository satisfies the configuration)

**Disposition:** OPEN as a Suggestion. It is the residual of `R8` rather than a new class — the same sentence, a stricter test — and is given its own ID so `R8`'s fix is recorded as having landed.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | `R9`; `R6` and `R8` confirmed FIXED; `R7` not re-raised. Gate-message strings compared to the emitter at byte level for all four branches and both call sites; tree-wide grep confirming no production consumer parses either string; CHANGELOG placement under `## [Unreleased]` and commit message checked for AI attribution; the `until_bash` short-circuit traced to a reachable termination-timing delta and reproduced by execution in a disposable copy; per-field and aggregate channel rules re-confirmed (`schemas.test.ts` 133 pass, `packages/web` console suite 439 pass); both short-circuit tests passing (`dag-executor.test.ts` 462 pass) |
| seams | **0 findings.** Code-surface immutability established by tree hashing (`packages/workflows` → `84c17f67…`, `packages/web` → `e543c377…` identical at both commits; 1374 tree entries at both) rather than by `--stat`; doc↔code gate-message seam verified in both directions with no prose-as-contract consumer; fresh 120-check cross-encoding matrix (80 schema+loader across both variants, 40 builder validate+round-trip) with 0 disagreements and 0 stored-value rewrites; `⏸` U+23F8 byte-identity between doc and both emitters; Part B's parity guard verified to exist at `scripts/node-ref-parity.test.ts:368` and to cover this PR's two channels, 72/72 passing. Scope note recorded: the builder has seven representable `VariantId`s and `loop_group` is not among them (`types/variant.ts:16-20`, `variants/detect.ts:19-28`), so `loop_group` parity is enforced at the schema/loader level only — pre-existing, orthogonal to this PR, and true of every field, not just the channels |

## Validation

|---|---|---|
| `bun run validate` (detached worktree at `0ef9e1e2`) | PASS | exit 0; **139 suite summaries, every one `0 fail`**; all five generated-artifact checks up to date (`bundled-defaults` 37 commands/21 workflows, `bundled-skill` 23 files across 2 skills, `bundled-schema`, `pi-vendor-map`, `capability-matrix`); `type-check` exit 0 in all packages; `lint --max-warnings 0`; `format:check`; `test:install` |
| `git rev-parse ae4c9ae9:packages/workflows 0ef9e1e2:packages/workflows` | PASS | identical tree `84c17f67…`; `packages/web` identical `e543c377…`; `git ls-tree -r \| wc -l` = 1374 at both — the code surface provably did not move |
| `git diff --stat ae4c9ae9 0ef9e1e2` | PASS | two files, 4 insertions / 3 deletions: `CHANGELOG.md` and `approval-nodes.md` |
| Gate-message doc↔code byte comparison (`R6`) | PASS | all four `buildHonestGateMessage` branches enumerated (`dag-executor.ts:3257-3264`); both documented forms match verbatim including the trailing period on `✅ Completion condition met.`; both call sites (`:3877` `loop_group`, `:5113` `loop`) pass the node's own `until` |
| `grep -rn "Completion signal detected\|Completion condition met"` (tree, excl. node_modules) | PASS | 7 hits: 2 emitter templates, 1 unrelated code comment, 2 test assertions, 2 doc lines — no production consumer parses either string |
| Independent cross-encoding matrix — 5×4 channel values × `dagNodeSchema` (`loop`, `loop_group`) + real `parseWorkflow()` loader | PASS | 80 checks, 0 disagreements, padded/multiline values byte-identical through both |
| Independent builder matrix — same 20 combinations × `validateStructural` + `loopFromDag`/`loopToDag` | PASS | 40 checks, 0 disagreements, 0 stored-value rewrites |
| `bun test src/dag-executor.test.ts` (`packages/workflows`) | PASS | 462 pass, 0 fail — includes the `loop:` short-circuit test (`:6977`) and the `loop_group:` "EDGE G" analog (`:15533`) |
| `bun test src/schemas.test.ts` / console suite in `packages/web` | PASS | 133 pass / 439 pass, 0 fail |
| `bun test scripts/node-ref-parity.test.ts` on `feat/loop-until-field-2563` (`R5` carrier) | PASS | 72 pass, 0 fail; `describe('loop completion-channel parity')` at `:368` executes `dagNodeSchema` vs `validateStructural` over 24 cases covering all three channels |
| Termination-timing reproduction (`R9`) | PASS (evidence) | two-invocation executor test in a disposable copy: counter absent after the signaled iteration, `1` not `2` after the fed-back iteration, node re-gates where the pre-PR engine completed |
| Tree-wide `until_bash` census (`R9` reachability) | PASS | exactly two workflows; the only both-channel one (`archon-adversarial-dev.yaml:302-303`) is a read-only `grep -qE` on a node with no `interactive` key |
| CI `test (ubuntu-latest)` @ `0ef9e1e2` | PASS | check run 95514864723 success |
| CI `test (windows-latest)` @ `0ef9e1e2` | PASS | check run 95514864750 success |
| CI `docker-build` @ `0ef9e1e2` | PASS | check run 95514864601 success |
| CI `schema-upgrade` @ `0ef9e1e2` | PASS | check run 95514864868 success |
| CI `Build docs site` @ `0ef9e1e2` | PASS | check run 95514863764 success |
| Mergeability | PASS | `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` at `0ef9e1e2` |

Reviewed in a detached worktree at the PR head; the repository, the PR, the branch, and the author's own worktree were not modified. The one mutation test ran in a disposable sibling copy and was deleted afterward.

## Verdict

**READY TO MERGE**

The three dispositions hold. `R6` is fixed and the doc now matches the emitter at byte level across all four branches and both call sites, with no consumer parsing either string. `R8` is fixed — the self-contradiction is gone. `R5`'s deferral is adequate on verified evidence rather than on assertion: the guard exists at `scripts/node-ref-parity.test.ts:368` on the stacked branch, executes both encodings over 24 cases covering all three channels including the exact blank-beside-valid shapes that were the original drift, and passes 72/72; deferring it one PR buys coverage of three channels instead of two, and the merge-together hold means `dev` never sees the rule without it.

Nothing regressed. The code surface is provably byte-identical to the head round 4 cleared — same subtree hashes, same tree entry count — and the channel rule, its engine↔builder parity, its non-rewriting storage, and both short-circuits were re-verified by execution anyway across 120 fresh cross-encoding checks with zero disagreements. `bun run validate` is exit 0 with `0 fail` on all 139 suite summaries, all five CI checks are green, and `mergeStateStatus` is `CLEAN`.

The one open finding is a Suggestion on a changelog sentence. `R9` says the tightened claim is exact for every workflow that exists — the repo's only both-channel loop uses a read-only predicate on a non-interactive node — and inexact only for a five-condition configuration in which a state-accumulating `until_bash` on an interactive loop reaches its threshold one iteration later. That is worth one clause if the author wants it, and the hold leaves room; it is not a defect and does not block. Merge readiness here remains a review verdict only: the PR is deliberately held until Part B (#2588) is ready, and the two land back-to-back.

Wirasm:
### Tree-wide loop audit — every existing `until:` preserved

Ran the real loader over the whole repo on this branch (bundled defaults + `.archon/workflows/`) and printed each loop node's declared completion channel:

```
archon-adversarial-dev :: adversarial-sprint -> until='ALL_SPRINTS_COMPLETE' + until_bash
archon-test-loop-dag   :: loop-counter       -> until='COMPLETE'
archon-piv-loop        :: explore            -> until='PLAN_READY'
archon-piv-loop        :: refine-plan        -> until='PLAN_APPROVED'
archon-piv-loop        :: implement          -> until='COMPLETE'
archon-piv-loop        :: fix-feedback       -> until='VALIDATED'
archon-ralph-dag       :: implement          -> until='COMPLETE'
e2e-joins              :: countdown          -> until_bash          <-- the decoy is gone
e2e-pi-all-nodes-smoke :: loop-node          -> until='DONE'
t1-fix-issue           :: build              -> until='BUILD-CLEAN'

total loop nodes: 10 · no loop without a channel
```

Two things this shows:

1. **Nothing regressed.** All nine pre-existing `until:` values load byte-identically — including `archon-piv-loop`'s four interactive human-judgment gates, where the sentinel is the *correct* tool (the model interprets a human's free-form reply and emits the signal only on explicit approval). This PR does not touch that case.
2. **The new shape works on a real file.** `e2e-joins :: countdown` now loads with `until_bash` alone, where it previously had to carry `until: __NEVER_SIGNALLED__` under a comment calling it dead config.

The five loader errors in the same run are pre-existing and unrelated — the ad-hoc audit script registers only `claude`/`codex`/`pi`, so the `opencode`/`copilot` e2e smoke workflows fail provider identity. They load normally under the real runtime, and `bun run validate` is green.

Wirasm:
### Review dispositions — both findings resolved

**`R1` — whitespace disagreement between the two encodings of the channel rule — FIXED** in `05847ec0`.

The reviewer is right, and right about the correction: a shared predicate is unavailable because `@archon/web` must not import `@archon/workflows`, so the mirror is hand-written by design. `loopControlSchema` now trims before deciding, which is the whole fix.

Worth stating why this is more than tidiness, since the report's own evidence makes the case: `bash -c "   "` exits 0, so a whitespace-only `until_bash` completes on iteration 1; and a whitespace-only `until` builds `detectCompletionSignal`'s regexes from whitespace, where the own-line and end-of-output patterns match almost any output. Both are now load errors. The `until` half of that drift predates this PR — it is fixed here because the same line decides it.

New coverage in `schemas.test.ts`: whitespace `until` alone, whitespace `until_bash` alone, and both-whitespace are each rejected with the channel message.

**`R2` — `packages/web/src/lib/api.ts` is a second derivation still typing `loop.until` as required — DECLINED for this PR**, matching the report's own "required outcome: none required for this PR".

Both reviewers grepped every consumer and found no `.until` dereference, so there is no reachable defect. Adding a second hand-written override would create more to unwind for zero current consumers, in files the console cutover has slated for removal.

Not left dangling, though: **Part B changes the loop schema again** (`output_format` on loop nodes plus `loop.until_field`), so it needs a spec regeneration anyway. I will regenerate `api.generated.d.ts` there and delete the `wire.ts` override as its own comment instructs — one regeneration fixing both derivations, rather than a second override now.

Also added in the same commit: direct unit coverage for `describeUnmetCompletion`, the new shared helper both loop variants use for the max-iterations message, including that it never emits the literal `undefined`.

`bun run validate` — exit 0 at `05847ec0`.

Wirasm:
### R1 closed properly in `284c8332` — the previous fix narrowed it, it didn't close it

Round 2 is right and the earlier commit's own message convicts it. The trim gated the **aggregate** question — whether *both* channels were blank — and never rejected a blank field beside a valid sibling. Reproduced on the previous head before fixing:

```
ACCEPT  until=" " + real until_bash          <- engine accepted, builder rejected: R1's original claim, still live
ACCEPT  until="COMPLETE" + until_bash="   "  <- both accepted; bash -c "   " exits 0, so it completes on iteration 1
reject  until=" " alone
reject  until_bash="   " alone
```

And the signal half is worse than "untidy" — `detectCompletionSignal` with a space-only signal, same run:

```
trailing space in output      : true
blank line inside code fence  : true
indented blank line           : true
```

Its own-line pattern matches any whitespace-only line the model emits. Combined with the `!signalDetected` short-circuit this PR adds, that false signal would skip the deterministic check entirely.

**The fix.** Both fields are validated non-blank *individually*, and the aggregate at-least-one-channel rule is kept — they answer different questions ("is this declared value usable?" vs "is anything declared?"), and only the pair covers a blank sibling.

Validation does not rewrite storage. `until` is matched verbatim by `detectCompletionSignal` and `until_bash` is executed verbatim, so trimming them here would silently change what an existing workflow does — a padded signal stays padded, an indented script stays indented. That is why this is a `.refine`, not the `.trim().min(1)` that `loop.command` uses eleven lines below: `command` is a filename resolved downstream, where the trimmed form is what resolution sees. Different jobs, different treatment, now stated at both fields.

**The builder mirror moved too.** It previously accepted a blank `until_bash`, so fixing only the engine would have created a fresh disagreement in the opposite direction. `LoopFields.tsx` also drops an all-blank box instead of storing it — that was the authoring motion producing `until: " "` from a leftover space.

**Parity verified by parsing, not reading** — both encodings over one matrix, 12/12 agree:

```
agree  engine=reject builder=reject  neither declared
agree  engine=accept builder=accept  until only, real
agree  engine=accept builder=accept  until_bash only, real
agree  engine=accept builder=accept  both real
agree  engine=reject builder=reject  until blank, no bash
agree  engine=reject builder=reject  until blank + real bash
agree  engine=reject builder=reject  real until + blank bash
agree  engine=reject builder=reject  both blank
agree  engine=reject builder=reject  until empty string
agree  engine=reject builder=reject  until_bash empty string
agree  engine=accept builder=accept  padded until (legit)
agree  engine=accept builder=accept  multiline until_bash (legit)

mismatches: 0
stored until unchanged: " COMPLETE "
stored until_bash unchanged: "  set -e\n  test -f x\n"
```

That matrix is now a test on **both** sides of the import boundary, with matching case names (`schemas.test.ts` ↔ `structural.test.ts`), since the two cannot share code — `@archon/web` must not import `@archon/workflows`. Matching names are what make a future divergence visible.

**On the comment.** Round 2's sharpest point was that a comment asserting an invariant the code doesn't enforce, sitting in the block Part B extends, is how the next person inherits a false premise. It is rewritten to state the rules **per channel** rather than enumerate them over two fields, and to say that adding a channel means adding it to both rules and to the builder's mirror — so the block gaining a third channel does not require rewriting it a third time.

`bun run validate` — exit 0 at `284c8332`.

Wirasm:
### ⏸ Hold — Part B stacks on this branch; merge the two back-to-back

Do not merge this on its own. Part B of #2563 (`output_format` on loop nodes plus a `loop.until_field` structured completion channel) branches from this head, and the two merge together.

The reason is partly this PR's own R2 disposition. I declined the second generated-spec derivation with "regenerate `api.generated.d.ts` in Part B and delete the `wire.ts` override then" — which means merging Part A alone ships a hand-written type override with a comment saying to delete it. Stacked, it never exists in a merged state. The rest of the surface moves the same way: both parts edit the same `superRefine`, and `LoopFields.tsx`, `structural.ts`, and the wire override all change again when the schema gains a third channel.

Nothing downstream is waiting either — both parts gate the bundled-defaults rewrite, so merging this early buys no one anything.

This PR stays open and reviewable in the meantime, and its checks are real (it targets `dev`). Part B's will not be, while it targets this branch — the test workflow and CodeRabbit skip non-`dev` bases, so its green checks should be read as *skipped*, not *passing*, until it is retargeted to `dev` after this merges. That will be stated in its own description too.

Wirasm:
### Round-3 dispositions

**`R3` — the builder silently drops nested `loop:` fields it does not hand-list — HANDLED IN PART B (#2588), and it was a real trap.**

This one deserved the emphasis it got. Reproduced both halves on the Part B branch before acting:

```
DROPPED   signal_completes (pre-existing)  lost=[signal_completes] issues=0
SURVIVES  until_field (Part B)
SURVIVES  until_bash
SURVIVES  until
```

`issues=0` is the whole finding: the importer's unsupported-key scan walks top-level wire keys only, so a field nested inside `loop: {…}` vanishes with nothing reported. Had Part B stopped at the schema and both validators, `until_field` would have validated on both sides, executed correctly from YAML, and then been **deleted the first time anyone opened that workflow in the console and saved** — the loop falling back to another channel or failing to terminate, with nothing anywhere saying why. That is worse than the mandatory-sentinel wart this PR removes.

`until_field` was already in `variants/loop.ts` when Part B landed it, so what was missing was the *proof*, not the wiring. Added in `5248431b`:

- a round-trip test asserting **exact** equality for a loop carrying `until_field` plus its `output_format`;
- verified as a real lock by mutation-testing in a **detached checkout** (never the live worktree): deleting the `...ifDefined('until_field', …)` line from the exporter fails it with a precise diff, and — as predicted — produces zero import issues;
- the sync clause in `loop.ts` now names **both** hand-maintained builder files and says how they fail *differently*: `structural.ts` gives a wrong verdict, `variants/loop.ts` loses the field silently.

While there I also fixed prose this branch had made stale: the docblock still said `until` and `until_bash` "are the channels today", which stopped being true when Part B added the third.

The general mechanism is pre-existing and not mine to fix here — `signal_completes` behaves this way on `dev` today, and the same shape applies to `approval.on_reject`, `retry`, `fan_out` and `pi`, which I have not audited. Filed as **#2589** with the reproduction, the two contributing mechanisms, and a note that hand-auditing the field list is not the fix, since that is the mechanism which already failed.

**`R4` — the builder mirror throws on a non-string `until` / `until_bash` — DECLINED**, explicitly rather than silently, since the reviewer recorded it for that purpose.

It is unreachable. `loader.ts:291` is the sole parse gate into `WireDagNode` and guarantees these are strings, so the `.trim()` calls cannot receive anything else; the console builder never constructs a wire node from untyped input by another route. Adding a runtime type guard would be defense in depth against a state nothing can produce, which `AGENTS.md` rules out under *YAGNI* and *no speculative defense in depth* — and it would add a branch no test could reach except by constructing an object the type system forbids.

If a second, unvalidated route into `WireDagNode` ever appears, the guard belongs at that route rather than scattered through the validators.

Wirasm:
Dismissing the 🔴 Critical "this test file cannot compile" — it is **false**, verified by compiling rather than by reading.

`npx tsc --noEmit -p packages/workflows/tsconfig.json` produces **zero `TS2451`** (redeclaration) errors on this branch. The flagged `const` declarations are not siblings in one scope: each sits inside its own `test(…, () => { … })` body, so each is block-scoped and redeclaration cannot arise. The only diagnostic anywhere near them is a pre-existing unused-import `TS6133`, identical on `dev` and untouched by this PR.

Recording it here rather than leaving it silently unaddressed: an unrefuted Critical costs the next reader time deciding whether to trust it, and this one would survive into the merge review as doubt. Two independent checks agree — the review agent refuted it by compiling a program explicitly including `schemas.test.ts`, and I reproduced that against the package tsconfig.

No change required.

Wirasm:
### Final-round dispositions — all three Suggestions addressed

**`R5` — both copies say "verify by parsing both, never by reading them", and neither did. TAKEN, in the stacked Part B (#2588).**

This was the one worth a decision rather than a shrug, and the argument holds: what shipped here was twin matrices with matching case *names*, which is prose, and prose is precisely the mechanism `scripts/node-ref-parity.test.ts` records failing twice (#2567, #2591). Every other web↔engine copy back-references that guard; the channel rule was the first written since that did not join it.

It joins in Part B, where it can cover all three channels at once rather than two that Part B would immediately extend. One corpus, **both encodings executed**, verdicts compared — 23 cases including every blank-beside-valid shape (the drift that actually happened here) and the padded/multiline entries that pin "trims to decide, never rewrites what it stores".

Mutation-verified: dropping `until_field` from the builder's channel rule fails it on exactly `agrees on {"until_field":"done"}` — engine accepts, builder rejects. That is the #2591 shape reproduced by the guard that would have caught #2591.

The twin matrices stay as each package's own regression coverage and now say what enforces them; both "verify by parsing both" instructions now name the guard.

**`R6` — `approval-nodes.md` described a gate line narrower than what this PR can emit. FIXED in `0ef9e1e2`.** The sentence now covers both forms. Confirming the finding's own correction: it is pre-existing prose from #2126, not something #2577 added — this PR made it incomplete rather than conflicting with it. Nothing parses either string (every consumer reads `metadata.approval.completionSignaled`), so this was prose accuracy, not a contract change.

**`R8` — the CHANGELOG claim contradicted its own next sentence. FIXED in `0ef9e1e2`.** "Every existing workflow is unaffected" → "still loads and terminates unchanged", which is the claim that is actually true. A workflow declaring both channels *is* affected: a side-effecting `until_bash` runs one time fewer.

**`R7`** — the CodeRabbit Critical was already dismissed as false, verified by compiling rather than reading; no change.

`bun run validate` — exit 0 at `0ef9e1e2`.

Wirasm:
### Rebased onto `dev@b5baa16b` (#2581) — findings stand by construction

Seven-file overlap with #2581. Both branches rebased; **nothing retargeted, nothing merged.**

| | Part A #2583 | Part B #2588 |
|---|---|---|
| Head | `ae3cf8c9` | `535d7be5` |
| Base | `dev` (`b5baa16b`) | `feat/loop-completion-2563` |
| `bun run validate` | exit 0 | exit 0 |
| CI | **all five green** | skipped (non-`dev` base) |

#### Diff-equivalence

**Part A** — excluding `CHANGELOG.md`, the pre- and post-rebase diffs are **identical across all 958 changed lines**. The CHANGELOG differs by exactly two things, both expected: my patch no longer *adds* a `### Changed` heading (#2581 opened that section first) and two blank lines move with it. The entry text is unchanged, including the R9 tightening — `Every existing workflow still loads, and no loop's completion verdict changes.`

**Part B** — excluding `AGENTS.md`, **identical across all 2937 changed lines**. `AGENTS.md` differs on both sides of one hunk because #2581 reworded the same 6.5k-character node-types line (the `effort`/`thinking` clause). Proven equivalent at the semantic level instead:

```
pre-rebase   applies exactly the loop substitution and nothing else: True
post-rebase  applies exactly the loop substitution and nothing else: True
```

and the resolved line is byte-for-byte `dev`'s line plus that one substitution — so #2581's `effort` rewording survived intact and my clause rode on top.

#### The specific hazard: eight of the ten claim sites live in files #2581 edited

`parameter-matrix.md` and `workflow-dag.md` were both touched by #2581, so a conflict resolution taking the wrong side would have resurrected a claim it took five rounds to kill — in files that ship in the binary. Rather than trust the resolution, I built a checker that greps all eight phrasings of the stale claim (must be absent) and eleven corrected statements (must be present), baselined it before the rebase, and re-ran it after:

```
clean   [output_format dropped on loop]      present [cli-commands rule]
clean   [loop ignores output_format]         present [parameter-matrix table cell]
clean   [output_format no-op on loop]        present [parameter-matrix legend]
clean   [solely until+until_bash]            present [workflow-dag loop capability]
clean   [required unless until_bash]         present [dag-advanced exception]
clean   [loops are the acknowledged gap]     present [rulecheck three tiers]
clean   [all three AI producers]             present [AGENTS three tiers]
clean   [loop until completion signal only]  present [loop-nodes ×4]

ALL CLAIM CHECKS PASS
```

Two of its patterns fired on the *corrected* text on first run — false positives from my own regexes — so the checker was tightened to exclude corrected and historical-tense context before being trusted. Separately: none of #2581's added lines under `.claude/skills/` mention loop completion or `output_format` at all; its skill edits are entirely `effort`/`thinking`/`modelReasoningEffort`, so it introduced no new contradiction.

#### `scripts/effort-ladder-parity.test.ts`

#2581's new guard — the third on this pattern — ran against this tree for the first time and passes, unmodified. All three parity guards together: 93 pass, 0 fail.

Issue #2091, opened 2026-07-14
Resume drops declaredFields on prior-run nodes — strict $node.output.field access silently downgrades to schemaless parsing


Wirasm:
## Problem

The strict `$nodeId.output.field` contract (`packages/workflows/src/output-ref.ts:107-157`) relies on `NodeOutput.declaredFields` — captured from the producer's `output_format` schema at completion — to enforce "field not in schema → fail the consumer" and "declared-optional-absent → ''".

On **resume**, prior-run completed nodes are rehydrated from `node_completed` events as plain text only:

- `getCompletedDagNodeOutputs` returns `Map<nodeId, string>` (`packages/core/src/db/workflow-events.ts:192-222`)
- the executor pre-populates them as `{ state: 'completed', output }` with **no `declaredFields` / `structuredOutput`** (`packages/workflows/src/dag-executor.ts:~4640-4647`)

Consequently, after a resume, a `$nodeId.output.field` reference against a prior-run node falls through to the lenient/schemaless path (case 4 in `output-ref.ts`): a typo'd or renamed field that would have thrown `not-in-schema` on a fresh run instead throws `missing-key` — or, worse, an author-declared-**optional** field that resolved to `''` on the fresh run now **throws** `missing-key` on resume, because the optionality information is gone. Same workflow, different semantics depending on whether the run resumed.

## Expected

Persist `declaredFields` (and ideally `structuredOutput`) alongside `node_output` in the `node_completed` event data, and rehydrate them on resume so the strict contract is invariant across fresh vs resumed runs. Alternatively, re-derive `declaredFields` from the workflow definition's `output_format` at pre-population time (the schema is available in the loaded YAML — no DB change needed).

The re-derive option looks cheapest: at `dag-executor.ts` pre-population, look up the node by id in `workflow.nodes` and call the same `declaredFieldsFromSchema(node.output_format)` used at completion (`dag-executor.ts:~1742`).

## References

- Strict contract: `packages/workflows/src/output-ref.ts:8-23, 107-157`
- Rehydration text-only: `packages/core/src/db/workflow-events.ts:192-222`, `packages/workflows/src/dag-executor.ts:~4640-4647`
- `declaredFields` capture on fresh completion: `packages/workflows/src/dag-executor.ts:~1742`

PR #2588, opened 2026-08-17
feat(workflows): a loop can end on a validated field instead of a string match

### LINKED ISSUE #2563: feat(workflows): loops can only terminate on a prose sentinel — no structured completion channel, and no way to opt out

Wirasm:
> **Stacked on #2583** (Part A). Base is `feat/loop-completion-2563`, not `dev` — review the [three-commit diff](https://github.com/coleam00/Archon/compare/feat/loop-completion-2563...feat/loop-until-field-2563) or just this PR's own commit. The two merge back-to-back; this one is retargeted to `dev` once #2583 lands.
>
> ⚠️ **This PR's green checks are meaningless.** The test workflow and CodeRabbit skip non-`dev` bases (this bit #2223/#2224), so a passing tick here means *skipped*, not *passed*. `bun run validate` locally is the real gate for now — **exit 0 at `f591bfaf`**, evidence below. Real CI runs after the retarget.

## ⚠️ This reverses part of #2579, which merged today

If you reviewed #2579 this morning, two things it established are no longer true, and you should see that here rather than discover it in the diff:

| | Before this PR (i.e. #2579 as merged) | After |
|---|---|---|
| A `loop:` node declaring `output_format` | **Rejected at load** by the `when:`-whole-output check, on the grounds that the schema is dropped at parse | **Accepted** — the schema is no longer dropped |
| The `'loop'` producer kind in `freeFormAiProducerKind` | Its own case, with its own message: *"`output_format` is dropped from a `loop:` node when the workflow is parsed, so declaring one would change nothing"* | **Gone.** A `loop:` is classified `schema-capable` and gets the same remedy as a `prompt:` node |
| The constitution's case-law row for #2566 | "Neither loop form has an opt-out" | A `loop:` opts out like a `prompt:`; only `loop_group:` has none |

This is not a disagreement with #2579 — it is #2579's own reasoning applied to a changed fact. Its message was correct when written: `output_format` *was* dropped from loop nodes at parse. This PR is what stops that being true, so the message became false and had to go with it.

I did not go looking for this. `schemas.test.ts`'s `output_format survival by node type` block — added by #2579 — pins the drop and its docblock explicitly names **#2563** as the change that would trip it, warning that the loader message "has to be revisited". It fired on rebase and was right on both counts. That claim had been stated in five places and derived from none; the test is the only reason it didn't rot silently.

Everything it touched is corrected together: the classifier, the message, the two loader tests that pinned the old reason, the authoring guide, and the constitution row. `loop_group:` keeps its own distinct reason — it never calls the provider itself, so nothing about it changed.

---

## The doc claim that took eight sites to kill — evidence for #2600

Worth reading before the diff, because it is the most transferable thing in this PR.

The sentence *"`output_format` is dropped / loop completion is decided by `until` and `until_bash`"* was true before this PR and false after it. Removing it took **eight sites across five review rounds**, and every round's sweep was better than the last and still incomplete:

| Round | Method | Missed |
|---|---|---|
| 1 | fixed the sites I knew about | the `loader.ts` docblock 3 lines above the function I rewrote |
| 2 | grepped `output_format` in the loop docs | 2 shipped skill files |
| 3 | grepped both terms across `.claude/skills` | a **table cell** in the file's own master matrix |
| 4 | read `parameter-matrix.md` end to end | 4 restatements in `loop-nodes.md` |
| 5 | swept `loop-nodes.md` for the framing | `cli-commands.md` — a doc about **CLI commands**, inside a parenthetical about event emission |

The last one is the argument. No amount of "sweep the loop docs" reaches a claim about loops living in a CLI reference; only a repo-wide search for the *claim itself* does. And two of the eight ship in the binary, where `check:bundled-skill` verifies coverage rather than content — so `bun run validate` passing says nothing about whether the shipped text is true.

The failure is not carelessness. It is that a hand-maintained claim has **no canonical location to sweep**, so correctness depends on whoever edits it guessing every place a past author chose to restate it. That is the case for #2600.

Two things changed as a result, and they are in this diff:

1. **Every one of the eight now states the rule instead of enumerating the channels** — "at least one completion channel", "no other channel already ended", "no channel vetoes another". A fourth channel would leave all of them true rather than starting a ninth round.
2. **The one claim that could be mechanized was mechanized.** The engine↔builder channel rule is no longer kept in step by matching test-case names; `scripts/node-ref-parity.test.ts` executes both encodings over one corpus and compares verdicts in CI.

---

## Problem and outcome

Part A (#2583) made the prose sentinel optional wherever completion is externally checkable. The case it could not help is **judgment**: a loop whose exit condition is the model's own assessment — "is the backlog empty?", "is this good enough?" — had no channel but prose, because `output_format` was stripped from loop nodes at parse time and `executeLoopNode` never requested it. Every other node type can express "the model decided X" through a validated schema; loops could not, which is the gap `AGENTS.md` named under *Natural Language Is Not a Wire Format*: *"Loops are the one gap: no structured completion channel exists yet (#2563)."*

- **Outcome:** a `loop:` node accepts `output_format`, and `loop.until_field` names a declared boolean in that schema whose validated `true` ends the loop. "The model judged this done" becomes a schema-validated field rather than a string match.
- **Invariant:** a loop terminates only through a channel its YAML declared, and a declared structured channel either yields a schema-valid payload or **fails the node** — it is never quietly treated as "not done yet".
- **Scope boundary:** `until:`'s matching semantics are untouched, `until_bash` still does not veto the signal, no bundled workflow is rewritten, and `until_field` is deliberately **not** offered on `loop_group` (see below — that is the load-bearing design decision, not an omission).

## Why this is admitted on `loop:` and declined on `loop_group:`

The first question a reviewer asks is "why not both?", so it is answered here rather than left to the diff. The admissibility test's third question is *could a script node plus existing wiring express this today?*

For `loop_group:` the answer is **yes**. A body node declares `output_format`, and `until_bash` reads it:

```yaml
loop_group:
  max_iterations: 10
  until_bash: '[ "$decide.output.done" = "true" ]'
  nodes:
    - id: decide
      output_format: { type: object, properties: { done: { type: boolean } }, required: [done] }
      prompt: Is the backlog empty?
```

That works today: `substituteNodeOutputRefs` runs over `until_bash` against the *current* iteration's outputs, and renders booleans unquoted.

For `loop:` the answer is **no**, and not for want of trying. The judgment lives in the loop's own AI turn, which has no node id and therefore no `$node.output.field` handle. `until_bash` runs after that turn but can only see `LOOP_PREV_OUTPUT` — the **previous** iteration's text. There is no wiring by which a deterministic check can read the current iteration's decision.

So the feature is admitted exactly where question 3 fails, and declined where it passes — I applied the test to my own feature and lost half of it. On a group the key is simply unknown, which leaves it channel-less and produces an error naming the two channels a group actually has; the docs point at the wiring above. A case-law row recording the verdict is added to the constitution.

## Review guidance

- **Feedback requested:** (1) whether the `loop_group` decline holds up — it is the argument this feature rests on; (2) the per-iteration reask restructure in `executeLoopNode`, which is the riskiest change here.
- **Start here:** `packages/workflows/src/schemas/loop.ts` — `until_field` and the moved across-channels rule. Then `packages/workflows/src/schemas/dag-node.ts` for the four load-time rules.
- **Review order:**
  1. `schemas/loop.ts` + `schemas/dag-node.ts` — the YAML surface and its load-time contract.
  2. `dag-executor.ts` `executeLoopNode` — the attempt loop, validation, and completion evaluation.
  3. `packages/web/src/lib/api.generated.d.ts` — **regenerated**; see *Lower-attention*.
  4. The builder (4 files) and docs.
- **Lower-attention areas:** the `api.generated.d.ts` diff looks large in raw form but is **3 lines** of real change (`until` becoming optional on both loop shapes, `until_field` appearing); everything else was `openapi-typescript`'s double-quote style normalised by prettier. Verified by diffing after formatting.
- **Known risk or uncertainty:** `executeLoopNode`'s iteration body is now wrapped in a labelled attempt loop. I chose that over extracting a closure specifically to keep every existing early `return failLoopNode(...)` working verbatim — the diff is one nested loop plus moved `let` declarations, not a restructure of the control flow. All 462 pre-existing executor tests pass unchanged, which is the main evidence it is behaviour-preserving for non-structured loops.

## Solution

**The field.** `output_format` at node level (the existing base field, no longer stripped for `loop:`) plus `loop.until_field: <property>`. The trio `until` / `until_bash` / `until_field` reads as one axis — *which channel decides completion* — and a bare identifier is the most declarative form available; anything ref-shaped would import `when:` grammar into a field that needs none.

**Four load-time rules**, all reads of data the author already wrote, each converting a silent non-termination into a load error:

1. `until_field` requires `output_format` on the same node;
2. the property must appear in `output_format.properties`;
3. it must be listed in `output_format.required` — otherwise a schema-valid payload may omit it, "absent" reads as "not complete", and the loop burns `max_iterations` reporting the wrong cause;
4. if it declares a `type`, that type must be `boolean` — otherwise `until_field: status` over a string invites truthiness semantics the engine deliberately does not have.

Runtime termination is `payload[field] === true`. Strict identity, no coercion. (The `loop:`-only scoping is argued above.)

**Provider behaviour is the same as every other node, deliberately.** The schema reaches the provider through the existing `resolveNodeProviderAndModel` path — loop nodes only ever missed it because the parse transform dropped the field. Each iteration's payload is then validated against the declared schema for *every* provider (an enforced provider still bypasses grammar-constrained decoding on a refusal or a `max_tokens` truncation), and on a best-effort provider a miss is re-asked up to three times **within the iteration**, with the schema errors appended and a fresh session per attempt. A reask does not consume a loop iteration. Exhaustion fails the node with the validation errors. A loop should not be the one node type where a structured miss behaves differently.

One interaction worth calling out: with grammar-constrained decoding the assistant *prose* is routinely empty, so two guards had to learn about payloads. The empty-output guard now exempts an iteration that produced a payload (otherwise every Claude structured loop would fail on iteration 1), and signal detection reads the prose **and** the serialized payload, so a loop declaring both `until:` and `output_format` does not end up with a signal channel that can never fire.

**R2 from #2583's round-1 review is resolved here as promised.** Regenerating `api.generated.d.ts` fixes both derivations of the spec at once and lets Part A's hand-written `wire.ts` override be **deleted rather than merged** — which is the substantive reason these two PRs stack instead of shipping separately.

## Behavior change

| | Before | After |
|---|---|---|
| `output_format` on a `loop:` node | Stripped at parse time with a loader warning | Honoured: reaches the provider, validated per iteration, becomes the node's output |
| Judgment-based completion | Prose sentinel only | `until_field` on a validated boolean |
| Loop output with a schema | Prose | The validated JSON; `$loopId.output.field` gets strict access |
| Invalid structured payload | Not possible (no schema) | Re-asked ×3 on Pi/Copilot, then **fails the node** |

## Validation

- `bun run validate` — **exit 0** at `c0571d82`, no failing suite in the log — the real gate for this PR while its CI is skipped.
- `bun test src/dag-executor.test.ts` — **472 pass**, of which the 462 pre-existing ones are unchanged: the attempt-loop restructure is behaviour-preserving for non-structured loops. Nine new cases carry the feature:
  - completes exactly on the iteration whose validated payload has the field `true`, and iterates while `false`;
  - `output_format` actually reaches `sendQuery` for a loop node (it did not before this PR);
  - `{ done: "true" }` — a string — **fails** rather than completing, proving strict typing;
  - a best-effort provider re-asks an invalid payload and completes on attempt 2, with **one** `loop_iteration_started` event despite two provider calls;
  - reask exhaustion fails the node after exactly 4 calls, rather than iterating to `max_iterations`;
  - an enforced provider fails on the first invalid payload with zero re-asks;
  - a downstream node resolves `$my-loop.output.note` through the strict field path;
  - an `until:` sentinel **inline at the very end of the prose** still fires when `output_format` is declared (the round-1 regression: detection used to concatenate prose and payload, moving the `$` anchor off the prose);
  - a bare-approve finalize carries `declaredFields`, so a declared-optional field resolves to `''` there exactly as it does on a loop that completed without a gate.
- `bun test src/schemas.test.ts src/loader.test.ts` — each of the four load-time rules produces its own error; `until_field` alone satisfies the channel rule; a `loop_group` declaring only `until_field` is rejected with a message that does **not** mention `until_field`.
- `bun --filter @archon/web test` — 432 pass; the builder round-trips `until_field` and mirrors the three-channel rule. The round-trip test is mutation-verified: deleting the exporter's `until_field` line fails it, which matters because the builder silently drops nested `loop:` fields it does not hand-list (pre-existing, tracked in #2589).
- Spec regeneration verified by inspecting the live `/api/openapi.json` before regenerating, then diffing after prettier: exactly `until?: string` ×2 and `+ until_field?: string`.
- **Not verified:** no live run against a real provider, and no live Pi/Copilot reask. The reask path is covered by the executor suite driving the real `executeLoopNode` with `getProviderCapabilities('pi')` resolving to `'best-effort'`, so the branch selection is exercised with the real capability lookup rather than a stub.

## Delivery considerations

|---|---|---|
| Compatibility | Purely additive: a loop without `output_format` takes exactly one attempt and behaves as before. The 462 unchanged executor tests are the evidence. | Full suite green |
| Rollout / rollback | Reverting restores the parse-time strip, which breaks only workflows written against this PR. Stacked on #2583, so revert order is this then that. | Single commit |
| Documentation | `loop-nodes.md` gains an `until_field` section including the `loop_group` escape hatch; the constitution gains a case-law row; `AGENTS.md`'s "loops are the one gap" sentence is now false and is replaced with the three-tier guidance; bundled skill references and `quick-reference.md` updated. | Included |

## Links

- Closes #2563
- Depends on #2583

## Summary by CodeRabbit

* **New Features**
  * Loops now support structured JSON output with validated boolean fields for completion.
  * Added an “Until field” option in the workflow builder.
  * Loop completion can use prose, shell checks, or structured field values.
  * Structured loop outputs are available to downstream workflow steps.

* **Bug Fixes**
  * Improved validation, retry handling, dry runs, streaming, and completion reporting for loop workflows.

* **Documentation**
  * Updated workflow guides, reference materials, changelog, and authoring guidance with the new loop behavior.


Wirasm:
### Update — `5248431b`: `until_field` is now locked against the builder's silent-drop mechanism

Part A's round-3 review found a trap with this PR's name on it. `builder/variants/loop.ts` is a hand-written field list, and the importer's unsupported-key scan walks **top-level** wire keys only — anything nested inside `loop: {…}` is invisible to it. A completion channel missing from that converter would validate on both sides, execute correctly from YAML, and then be **silently deleted the first time anyone opened the workflow in the console and saved**.

`until_field` was already wired into the converter here, so nothing was broken — what was missing was the proof. Verified by execution rather than reading, on this branch:

```
DROPPED   signal_completes (pre-existing)  lost=[signal_completes] issues=0
SURVIVES  until_field (this PR)
SURVIVES  until_bash
SURVIVES  until
```

Added:

- a round-trip test asserting **exact** equality for a loop carrying `until_field` plus its `output_format`;
- mutation-tested in a **detached checkout** — deleting the exporter's `...ifDefined('until_field', …)` line fails it with a precise diff, and produces zero import issues, confirming the mechanism;
- the sync clause in `loop.ts` now names **both** hand-maintained builder files and states how they fail differently — `structural.ts` yields a wrong verdict, `variants/loop.ts` loses the field with no signal at all.

Also corrected prose this PR had made stale: the schema docblock still said `until` and `until_bash` "are the channels today", which stopped being true the moment this branch added the third.

The general mechanism is pre-existing (`signal_completes` behaves this way on `dev` today, and `approval.on_reject` / `retry` / `fan_out` / `pi` share the shape) and is tracked separately in **#2589**.

`bun run validate` — exit 0 at `5248431b`.

Wirasm:
# PR Review: #2588 — feat(workflows): a loop can end on a validated field instead of a string match

> Round 5, reviewed against the PR's real base `feat/loop-completion-2563` (`94be25c8`). Part A's content is out of scope. The PR's green CI ticks mean *skipped*, not passed (non-`dev` base, #2223/#2224); they were not used as evidence in either direction.

## Signal

**The docs-only claim is true, and I verified it more strongly than it was stated.** The author's proof checked three subtrees; the whole-repo check is stricter and also passes: `git diff --name-only b310cfad b72c2c86` returns **exactly one path**, `packages/docs-web/src/content/docs/guides/loop-nodes.md`. No code file, test, generated artifact or config differs. Every round-4 finding therefore stands by construction and none was re-litigated.

**R12 is closed, and the four new sentences are accurate.** I verified each against the engine rather than against the author's description: the config-block foot matches `loopNodeConfigSchema`'s across-channel `superRefine`, the `until_bash` skip and the caution box match the guard `if (loop.until_bash && !signalDetected && !fieldComplete)`, and "No channel vetoes another" matches `completionDetected = fieldComplete || signalDetected || bashComplete`. The caution box's trickiest clause — that `signal_completes: true` rescues a first-run completion *on any channel* — is correct: the gate is `completionDetected && (!interactiveFirstRun || signalCompletes)`, which branches on the OR, not on the prose signal. All four dead phrasings are gone, and the negative check has teeth (each was present at `b310cfad`). On forward-compatibility, three of the four are channel-count-agnostic and survive a fourth channel verbatim; the fourth necessarily names YAML keys, which both agents and I agree is not a defect.

**But the sweep that produced them is still one file wide, and I found two more sites — the third round running.** The author's comment states the method as "sweeping the file for the *framing* rather than for the three cited lines". That method is right and it did find a fourth site unprompted. It missed a fifth **in the same file**: `loop-nodes.md:204-207` still says a loop declaring neither `until` nor `until_bash` "is rejected at load time". I disproved that by execution — an `until_field`-only loop parses clean — and the same file recommends exactly that configuration 120 lines earlier. Separately, `seams` found the claim restated in a **shipped skill file** the sweep never covered. Neither blocks, and both are one-line edits, but the recurring shape is the finding worth reading: R8 was six sites, R12 four, this round two more, each time after a sweep declared itself complete. #2600 tracks generating `parameter-matrix.md` from the schema; it does not cover prose in either of these files.

- **Blocking:** 0
- **Non-blocking:** 5 (3 OPEN Suggestions, 2 tracked follow-ups)
- **Resolved since previous review:** 1 (R12)
- **Tracked follow-ups:** 2 — [#2589](https://github.com/coleam00/Archon/issues/2589), [#2599](https://github.com/coleam00/Archon/issues/2599)
- **Validation:** `bun run validate` PASS (exit 0) in an independent detached worktree at `b72c2c86` — generated-file checks, type-check, lint `--max-warnings 0`, `format:check`, `test:install`, and the full per-package test run, with no suite reporting a non-zero fail count.

## Findings

|---|---|---|---|
| `R1` | Important | Appending the payload to the prose moved the end-of-output anchor, killing the inline `until:` sentinel for any schema'd loop | FIXED |
| `R2` | Important | Finalize-on-bare-approve returned the loop output without `declaredFields` | FIXED |
| `R3` | Important | The dry-run simulator never learned `until_field` | FIXED |
| `R4` | Suggestion | The console builder silently drops nested `loop:` sub-keys it does not hand-list | TRACKED FOLLOW-UP |
| `R5` | Suggestion | `LoopFields.tsx` pointed at an `output_format` editor that does not exist | FIXED |
| `R6` | Suggestion | The twin channel-verdict matrix gained `until_field` cases on the builder side only | FIXED |
| `R7` | Important | A best-effort reask rebased the loop's threaded session onto the reask's throwaway session | FIXED |
| `R8` | Important | The "`output_format` is dropped on a `loop:`" claim survived six sites across three rounds | FIXED |
| `R9` | Suggestion | The R3 rewrite also changed dry-run behaviour for the pre-existing `until` + `until_bash` combination | FIXED |
| `R10` | Suggestion | The builder mirrors the channel rule but not the four `until_field`↔`output_format` rules | TRACKED FOLLOW-UP |
| `R11` | Suggestion | The interactive gate's status line credits the prose sentinel for another channel's completion — pre-existing on `dev` | OPEN |
| `R12` | Suggestion | `loop-nodes.md` widened the general channel rule but left Part A-era two-channel restatements | FIXED |
| `R13` | Suggestion | The `### until` section still says an `until_field`-only loop "is rejected at load time" — disproved by execution, and contradicted by the same file's own recommendation | OPEN |
| `R14` | Suggestion | A **shipped** skill reference still states loop completion "is decided solely by the `until` signal and `until_bash`" | OPEN |

## Detailed Findings

<details>
<summary><code>R13</code> — The <code>### until</code> section claims an <code>until_field</code>-only loop is rejected at load time; it is accepted</summary>

**Impact:** `packages/docs-web/src/content/docs/guides/loop-nodes.md:204-207` tells an author that the exact configuration this PR exists to enable will not load. The file contradicts itself: at `:80` the "Choosing a completion channel" table recommends `output_format` + `until_field` — no `until`, no `until_bash` — for "a judgment the model makes, no human in the loop", and calls it the preferred form. At `:204` the reader is then told:

```
**Optional when `until_bash` is set.** Omit `until` entirely for a loop whose
completion is deterministic — … A loop declaring neither channel is rejected
at load time.
```

Both halves are stale: `until` is optional when `until_bash` **or** `until_field` is set, and a loop declaring neither of those two named channels is *accepted* when it declares the third.

**Evidence — disproved by execution, with a negative control.** Parsed real `loopNodeConfigSchema` at this head:

```
ACCEPTED  until_field ONLY (no until, no until_bash)
ACCEPTED  until_bash ONLY
ACCEPTED  until ONLY
REJECTED  NO channel at all
          -> loop node requires a completion channel: set 'loop.until' … or 'loop.until_bash' … or 'loop.until_field' …
```

Source: `packages/workflows/src/schemas/loop.ts:195-204` — the `superRefine` filters `[data.until, data.until_bash, data.until_field]` and `addMissingChannelIssue` fires only on `declared.length === 0`. The engine's own error message names all three channels; this doc sentence names two.

The lines are untouched by Part B (`git diff 94be25c8...b72c2c86` shows no change there) and were true when written on the base, where only two channels existed — the same provenance as every R12 site. Part B is what makes them false.

**Required outcome:** one sentence. Name the rule the way the four R12 sites now do, e.g. "A loop declaring none of `until` / `until_bash` / `until_field` is rejected at load time", and widen the bold lead-in to cover `until_field`. Cheap, and it removes a direct self-contradiction inside one guide.

**Found by:** `prp-core:code-reviewer`; independently confirmed by my own repo-wide sweep and proved by execution here.

**Disposition:** OPEN — a Suggestion. `prp-core:code-reviewer` rated this **Important** and the argument is reasonable: it is a flatly false load-time claim in the guide for the headline feature, and a linear reader meets it *after* the correct rule. I am not adopting that severity, for consistency rather than comfort: round 4 dispositioned R12 — four sentences of exactly this class, in exactly this file — as a non-blocking Suggestion, and nothing here runs wrong, is saved wrong, or is lost. The disagreement is recorded rather than averaged away.

</details>

<details>
<summary><code>R14</code> — A shipped skill reference still names only two completion channels</summary>

**Impact:** `.claude/skills/archon/references/cli-commands.md:157`:

> **Write-only observability** — it does NOT steer the run: loop completion is decided solely by the `until` signal and `until_bash`, never by emitted events.

The load-bearing claim (emitted events do not steer the run) stays true; the supporting enumeration does not. "Decided **solely** by" makes it a factual exhaustiveness claim rather than an incomplete example, and it is false once Part B lands.

**This file ships in the binary.** `packages/cli/src/bundled-skill.ts:27` imports it and `:61` registers it as `references/cli-commands.md` — the same bundling path as `parameter-matrix.md`, whose stale claim was round 3's blocker (R8). An authoring agent reading the shipped skill is told two mechanisms decide loop completion, and may reach for a file-plus-`until_bash` workaround where a schema-validated boolean was available and is the form this PR's own guidance prefers.

**Evidence:** untouched anywhere in the PR (`git diff --name-only 94be25c8...b72c2c86` does not list it), and byte-identical on `dev`. Unlike R11, that does *not* make it pre-existing: on `dev` the sentence is **true** (only two channels exist there). Part B is what falsifies it — the same causal status as the four sentences the final commit fixed, one file over.

**Required outcome:** the same one-line edit already applied in six other files: "…decided solely by the `until` signal, `until_bash`, or `until_field`, never by emitted events."

**Found by:** `prp-core:seam-analyzer`; independently confirmed by my own sweep, and the shipping status verified against `bundled-skill.ts` here.

**Disposition:** OPEN — a Suggestion. Lower impact than R13 because the sentence's actual subject (events are write-only) remains correct and the reader's decision — whether to use `event emit` to steer a loop — is unaffected.

</details>

<details>
<summary><code>R12</code> — <code>loop-nodes.md</code>'s Part A-era two-channel restatements</summary>

**Verified FIXED at `b72c2c86`.** All four sites now state the rule instead of counting channels, and each was checked against the engine rather than against the author's table:

| Site | New text | Engine authority |
|---|---|---|
| config-block foot | "At least ONE completion channel is required — any of until / until_bash / until_field. None is required alone." | `schemas/loop.ts:195-204` — `superRefine` over all three; proved by execution under R13 |
| `until_bash` skip | "runs only on iterations that no other completion channel already ended" | `dag-executor.ts` — `if (loop.until_bash && !signalDetected && !fieldComplete)` |
| caution box | "does not run on an iteration another channel already completed"; interactive first run "completes on another channel still gates … (unless `signal_completes: true`)" | same guard, plus the gate `if (completionDetected && (!interactiveFirstRun || signalCompletes))` — `signalCompletes` is `loop.signal_completes === true` and gates the OR, so it does rescue *any* channel |
| deterministic-exit pattern | "No channel vetoes another. They are OR'd" | `completionDetected = fieldComplete \|\| signalDetected \|\| bashComplete` |

The fourth site was not in round 4's list; the author found it by sweeping for the framing. Negative assertions confirm every dead phrasing is gone at head (`Required unless \`until_bash\` is set`, `the \`until\` signal did not already`, `does not run on a signalled iteration`, `The bash check does not veto the signal`, `The two channels are OR`, `whose first run signals` — all absent), and each of the three checkable ones was present at `b310cfad`, so the check is not vacuous.

**Forward-compatibility (the stated goal):** three of the four are channel-count-agnostic and remain true verbatim if a fourth channel ships. The fourth is an enumeration of YAML keys and must name them; both agents independently reached the same conclusion, and neither counts it as a defect.

**Disposition:** FIXED. The remedy generalised correctly — but its *reach* did not: R13 sits in the same file and R14 in a shipped skill, both the same claim class.

</details>

<details>
<summary><code>R11</code> — The interactive gate's status line credits the prose sentinel for another channel's completion</summary>

**Out of scope this round** at the maintainer's direction; being filed as a standalone issue. Carried here only so the ledger stays complete — not re-litigated and not re-evidenced.

Round 4's disposition stands unchanged: reproduced by execution on `dev` (`7752ec53`) with `until` + `until_bash` and no `until_field` involved, so it ships today; `buildHonestGateMessage` is untouched by this diff. I re-confirmed only the one fact that could have changed — the gate code is byte-identical between `b310cfad` and `b72c2c86`, since the whole-repo diff is one doc file. Any fix must also cover the pre-existing `until_bash` case and `executeLoopGroupNode`'s identical call, or it just moves the lie.

**Disposition:** OPEN — a Suggestion, not caused by this PR.

</details>

<details>
<summary><code>R1</code>–<code>R3</code>, <code>R5</code>–<code>R9</code> — settled in rounds 1–4</summary>

All FIXED and verified at `b310cfad` in round 4, with per-finding evidence recorded in that round's report. **Preserved by construction this round**, not by re-reading: the whole-repo diff `b310cfad..b72c2c86` is a single documentation file, so every code path, test and generated artifact these findings concern is byte-identical to the state round 4 verified.

- `R1` — two independent `detectCompletionSignal` calls, each keeping its own anchor.
- `R2` — `declaredFields` set on the finalize-on-bare-approve exit, matching the normal completion and resume-hydration paths.
- `R3` — `dry-run.ts` evaluates `until_field` with the executor's `=== true` identity check; `loop_group`'s call site correctly passes no `until_field`.
- `R5` — `LoopFields.tsx` now truthfully states the schema round-trips but is not builder-editable.
- `R6` — engine and builder channel-verdict matrices carry the same two `until_field` case names verbatim.
- `R7` — threading follows attempt 0 only; round 3 mutation-tested the shipped guard and re-derived that round 2's proposed correction was wrong. Recorded permanently.
- `R8` — the shipped parameter matrix agrees with itself; three independent sweeps found no seventh site of *that* claim. Note R14 is a **different** claim class (the completion-channel set, not `output_format` applicability) and would not have been caught by those sweeps, which keyed on `output_format`.
- `R9` — the dry-run trade is documented in a docblock naming the affected shipped workflow, and pinned by tests.

</details>

<details>
<summary><code>R4</code>, <code>R10</code> — tracked follow-ups</summary>

Both out of scope this round at the maintainer's direction and unchanged.

- `R4` — builder silently drops nested `loop:` sub-keys. `until_field` itself is *not* affected (wired symmetrically and locked by an exact-equality round-trip test). Tracked in [#2589](https://github.com/coleam00/Archon/issues/2589).
- `R10` — the console builder cannot author an `until_field` loop end to end; it validates clean and the server refuses the save. Fails loudly, nothing saved wrong, surface lives under `src/experiments/`. Tracked in [#2599](https://github.com/coleam00/Archon/issues/2599).

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | R13 — the `### until` section's false load-time claim; the four new sentences verified accurate against `executeLoopNode` and the schema; forward-compatibility confirmed for three of four |
| seams | R14 — the completion-channel set restated in a shipped skill reference; 17 restatement sites enumerated; `loop_group`-scoped two-channel sentences confirmed **correct** (the group genuinely has no `until_field`), and the schema's own `BASE_COMPLETION_CHANNELS` + `addMissingChannelIssue` confirmed to generate both variants' messages from one array rather than hand-copying |
| — | Independent whole-repo sweep run here converged on exactly R13 and R14 and cleared the remaining candidates (see Validation) |

## Validation

|---|---|---|
| `bun run validate` (independent detached worktree at `b72c2c86`) | PASS | exit 0; no suite reported a non-zero fail count. The `error`/`err` lines in the log are intentional negative-path fixtures (mocked `EACCES`, "Disk full", "connection refused", "Cannot connect to the Docker daemon") logged by Pino inside passing tests |
| Whole-repo tree diff `b310cfad..b72c2c86` | Doc-only, confirmed | `git diff --name-only` returns exactly one path (`loop-nodes.md`); every code subtree hash identical. Stricter than the three-subtree check claimed — every round-4 finding stands by construction |
| Probe: `loopNodeConfigSchema.safeParse` on all four channel combinations | R13 disproved by execution | `until_field` alone **ACCEPTED**; `until_bash` alone accepted; `until` alone accepted; no channel **REJECTED** with a message naming all three — evidence for R13 |
| Negative assertions: six dead R12 phrasings absent at head; three confirmed present at `b310cfad` | PASS | R12 genuinely closed, and the check is not vacuous |
| Engine cross-check of the four new sentences (`superRefine`, the `until_bash` guard, the `completionDetected` OR, the `signal_completes` gate) | All accurate | `schemas/loop.ts:195-204`; `if (loop.until_bash && !signalDetected && !fieldComplete)`; `completionDetected = fieldComplete \|\| signalDetected \|\| bashComplete`; `completionDetected && (!interactiveFirstRun \|\| signalCompletes)` |
| Sweep: every `until_bash` mention lacking `until_field` across the docs site, `.claude/skills/**`, `.claude/docs`, `AGENTS.md`, `CLAUDE.md`, `CHANGELOG.md`, bundled `.archon/**` | 2 sites, both reported | Cleared as correct in context: `workflow-dag.md:493` (narrates one example's flow; the authoritative three-channel rule follows at `:505-511`), `:567`/`:598` and the `quick-reference.md` loop_group table (correctly two-channel — a group has no `until_field`), `quick-reference.md:198-199` (loop table carries an `until_field` row), `parameter-matrix.md:104` (advice about `retry:`, not an exhaustiveness claim) |
| `packages/cli/src/bundled-skill.ts` | `cli-commands.md` ships | imported at `:27`, registered at `:61` — same bundling path as `parameter-matrix.md` (R8) |
| `signal_completes` prose vs. behaviour on `dev` | Pre-existing, not raised | `dev` already gates on `completionDetected = signalDetected \|\| bashComplete`, so describing it as "a detected completion **signal**" was already narrower than behaviour before Part B — same status as R11, so not attributed here |
| PR CI checks | NOT RUN | Test workflow and CodeRabbit skip non-`dev` bases (#2223/#2224); green ticks mean *skipped* and were not treated as evidence |

## Verdict

**READY TO MERGE**

The confirmation this round was asked to make holds. The docs-only claim is true at whole-repo granularity, so rounds 1–4 carry over by construction rather than by re-reading. R12's four rewritten sentences are accurate against the engine — including the caution box's subtle claim that `signal_completes` rescues a first-run completion on any channel, which is right because the gate branches on the OR — the dead phrasings are gone under a check with teeth, and three of the four survive a fourth channel verbatim, which was the stated goal. `bun run validate` is exit 0 at this head in an independent worktree.

Two more sites of the same claim class remain open, neither blocking: `loop-nodes.md:204-207` tells an author that an `until_field`-only loop will not load, which I disproved by execution and which the same file contradicts 120 lines earlier; and a **shipped** skill reference still says completion is "decided solely by the `until` signal and `until_bash`". Both are one-line edits and both are cheap to fold in before merge, but neither changes behaviour and R12 set the precedent for this class at Suggestion. `prp-core:code-reviewer` rated R13 Important; that disagreement is recorded in the finding rather than averaged away.

The durable observation is the pattern, not the two sentences. R8 was six sites, R12 four, this round two more — each round after a sweep declared itself complete, each time in a file the previous sweep's keyword did not reach. The method improved correctly from grep to reading-for-framing, but its *scope* has stayed one file wide while the claim lives in at least three. #2600 tracks generating `parameter-matrix.md` from the schema and would not have caught either of these; a check that keys on the channel-set claim across the docs site and the shipped skills is the version that would.

Wirasm:
### All three blockers fixed in `89260d16` — stack rebased onto `0f46d409`

**`R1` — Part B broke the existing `until:` channel. FIXED, and it was the right call to lead with.**

This is the same species as the Part A whitespace bug: a signal path that works today quietly stopped firing. Reproduced at the unit level before touching anything, which also shows why my original test missed it:

```
alone=true  concat=false  sentinel inline at end of output   <-- BROKEN BY CONCATENATION
alone=true  concat=false  sentinel inline, trailing period   <-- BROKEN BY CONCATENATION
alone=true  concat=true   sentinel on its own line
alone=true  concat=true   promise tag
```

The own-line pattern carries `m` so it survives concatenation; the end-of-output pattern is `$`-anchored with no `m`, and an object payload always ends `}`. My test covered only the `<promise>` form — one of the two that happen to survive. Now two independent `detectCompletionSignal` calls, each keeping its own anchor.

Mutation-tested in a detached checkout: restoring the concatenated haystack fails the new test with exactly the reported symptom, and it is the **only** failure out of 470 — confirming nothing else covered this.

**`R2` — finalize-on-bare-approve omitted `declaredFields`. FIXED.** It now re-derives them from the definition, like the normal completion return and the resume-hydration path (#2091). Test drives the real gate metadata with a declared-optional field absent from the finalized payload; mutation-tested — without the fix the consumer throws an `OutputRefError` and the run yields no output at all.

**`R3` — the dry-run simulator never learned `until_field`. FIXED.** It can evaluate the channel exactly (a stub object is hydrated onto `structuredOutput`), so it now decides instead of guessing — `done: true` completes with reason `'done' is true after 1 iteration(s)`, `done: false` correctly does **not** complete, and the "assumed complete" fallback no longer names an `until_bash` the author never declared. Guessing was defensible for `until_bash`, which is genuinely unobservable; it was never defensible here.

**`R4`** — untouched, pre-existing, tracked in #2589.

---

### A tripwire on `dev` fired, and it was right

Rebasing picked up `schemas.test.ts`'s `output_format survival by node type` block, whose docblock names **#2563** as the change that would trip it and warns that a loader message must be revisited. It did trip, and the warning was accurate: `loader.ts` told authors that `output_format` *"is dropped from a `loop:` node when the workflow is parsed, so declaring one would change nothing"* — false as of this PR.

A `loop:` is now **schema-capable**: it runs its own `sendQuery`, so a declared schema reaches the provider and the node's output becomes the validated JSON, exactly like a `prompt:` node. So it now gets the same verdict and the same remedy from the `when:` whole-output check, the separate `'loop'` producer kind is gone, and a loop that declares `output_format` is **accepted** where it was previously rejected. Corrected in the same commit: the loader classifier and message, the two loader tests that pinned the old reason, the authoring guide, and the constitution's case-law row.

That claim was stated in five places and derived from none — which is exactly what that test was written to catch. Worth noting for the record that it earned its keep on its first real encounter.

`loader.ts` was named as off-limits earlier in this workstream; #2566/#2579 have since merged, so it is merged code rather than a sibling's in-flight branch. Flagging the edit explicitly rather than assuming.

---

### Base

Part A rebased onto `dev@0f46d409` (`70c5b631`), Part B rebased onto Part A (`89260d16`). `scripts/node-ref-parity.test.ts` ran against this tree for the first time and **passes** — 49/49, unmodified.

`bun run validate` — exit 0 on both branches.

Wirasm:
### Round-2 dispositions — all five OPEN findings addressed in `5210c0d6`

**`R7` — a re-ask rebased the loop's threaded session. FIXED, but NOT with the suggested pin — I tested it and it does not hold.**

The suggested refinement was flagged as un-mutation-tested, so I proved it before relying on it. It fails three cases:

| Case | Attempt-0 rule (shipped here) | Suggested pin |
|---|---|---|
| Mid-loop re-ask, iteration 2 resumes | `thread-2` ✅ | `thread-1` — rewinds past the whole re-asked iteration |
| **Re-ask on iteration 1** | `orig-1` ✅ | **`throwaway`** ❌ |
| **Next sequential node after a final-iteration re-ask** | `loop-conv` ✅ | **`throwaway`** ❌ |

The `resumeSessionId !== undefined` guard is what breaks it. At `i === 1` there is nothing to restore, so the pin is a no-op precisely where the unconditional adoption has already set the throwaway — the defect survives one iteration earlier. And since a short loop can finish on iteration 1, the cross-node case the amendment is about is left open too.

**The rule shipped instead: threading follows attempt 0 only.** Attempt 0's session *is* the loop's conversation, so it stays the thread and a re-ask's session is never adopted. That is correct at `i === 1` (attempt 0 established the conversation), correct mid-loop, and correct at the node boundary — the loop's completion cursor is now always the conversation, never a single-turn artifact. The discard is logged (`loop_node.reask_session_not_threaded`). The repaired answer still reaches the next iteration through `$LOOP_PREV_OUTPUT`, which is prompt text rather than session state, so nothing is lost by not threading it.

All three cases are tests, and all three are mutation-verified against the unconditional adoption — reverting it reproduces the reviewer's `throwaway` exactly. `loop-nodes.md`'s `fresh_context` table now states the re-ask exemption instead of contradicting it.

**`R8` — the replaced claim survived in five places. All five FIXED.**

I found the fifth (`workflow-dag.md:527`) by grepping rather than from the list, before the amendment arrived — it was the worst one, sixteen lines below a sentence this PR added saying the opposite, in the same `## Loop Nodes` section, and it ships.

1. `loader.ts` docblock — rewritten. It documented a `'loop'` kind no longer in the return type and asserted "Declaring one is not an opt-out; it is a no-op." Now describes the two kinds that exist, with the `loop:` history in past tense.
2. `dag-advanced.md:3` **(ships)** — `output_format` on a `loop:` named as an honored exception.
3. `parameter-matrix.md:92` **(ships)** — dropped from the silently-ignored list, `loop_group` noted as still ignored; the loop rows also gained the `until_field` entry they never had.
4. `workflow-dag.md:527` **(ships)** — flipped from "silently ignored" to **WORKS**.
5. `rulecheck/rules-guide.md:23` — "loops are the acknowledged gap" replaced with the three tiers `AGENTS.md` now carries.

No regeneration needed: `bundled-skill.ts` imports by path and `check:bundled-skill` verifies coverage, not content. `check:bundled-skill` passes.

**`R9` — kept, now tested and documented.** Assuming completion when an unevaluable `until_bash` is declared is the honest answer — the real run's check may well have fired, and the old verdict was a failure the real run would not produce. The docblock now names the reach (it covers a declared-but-unmatched `until:`, which a shipped default relies on) and the trade (a dry run can no longer prove a prose stub trips `until:` on a loop that also declares `until_bash`), with a test pinning it.

**`R6` — FIXED, using the caveat rather than the two-line version.** As flagged, the engine matrix builds the node without `output_format`, so a bare `until_field` case would be rejected for the wrong reason and the matrices would agree by accident. The two cases now supply a schema so what is compared is the channel verdict.

**`R5` — FIXED.** The comment no longer claims an `output_format` editor exists; it says the schema is carried through the round-trip but is not editable in the builder today.

**`R4`** — untouched, tracked in #2589.

`bun run validate` — exit 0 at `5210c0d6`.

Wirasm:
### R8 sixth site — fixed in `b310cfad`, and I changed how I looked for it

The finding is right and the diagnosis of *why* is the part worth recording: `parameter-matrix.md` contradicted itself three ways, and this PR authored two of the three. It corrected the recipe row (`:58`) and the gotchas entry (`:93`) and never touched the master table (`:23`), which is the surface someone actually scans — the file's own legend defines `ignored` as "no effect at runtime; loader emits a warning."

Three rounds of grepping prose missed a table cell, so a fourth round of grepping would have missed the next thing. This pass **read the file end to end**:

- `:23` — the table cell: `ignored` → **`yes` (pairs with `loop.until_field`)**
- `:43` — the legend paragraph that named only `model`/`provider` as the loop exception now names `output_format` too, with the reason (a `loop:` makes its own provider call)
- every other `output_format` statement in the file re-checked against those two

Then the same treatment for the other shipped skill files, searching **both phrasings** of the claim rather than one term — `output_format` and the loop-completion vocabulary — since searching one term is exactly what failed here. Two remaining `output_format … ignored` lines in `workflow-dag.md` (`:191`, `:682`) are about **script** and **approval** nodes and are correct; `dag-advanced.md:13` is the provider enforcement table. No seventh site.

Your point about `check:bundled-skill` is the one I want to carry: it verifies coverage, not content, so `bun run validate` passing tells you nothing about whether the shipped text is true. That is precisely why this needed reading rather than tooling — and it is the same defect class as the tripwire that started this thread, one layer further out.

**R10** — not touched, per your instruction; it is yours to file.

`bun run validate` — exit 0 at `b310cfad`. Stack rebased: Part A `94be25c8`, Part B on top of it.

Wirasm:
### R12 fixed in `b72c2c86` — and there was a fourth site

All three cited sites corrected, plus one the list did not name. The fix is not "update the count" but **stop counting**: every one now states the rule, so a fourth channel would leave all of them true.

| Site | Was | Now |
|---|---|---|
| `:92` config block | "Required unless `until_bash` is set" — two lines above the `until_field` entry this PR added | `until:` described plainly; the block's foot carries "**At least ONE** completion channel is required — any of until / until_bash / until_field. None is required alone." |
| `:250` | "iterations that the `until` signal did not already complete" | "iterations that **no other completion channel** already ended" |
| caution box | "does not run on a **signalled** iteration" | "does not run on an iteration **another channel already completed**" |
| `:452` *(not in the list)* | "**The bash check does not veto the signal.** The two channels are OR'd" | "**No channel vetoes another.** They are OR'd" |

The fourth turned up by sweeping the file for the *framing* rather than for the three cited lines — the same method change that closed R8, applied without being asked this time.

Your read of the caution box was exactly right: correct about the mechanism, wrong only about what triggers it. The negative control settles it — `until_field: true` and `until_bash` does not run; `until_field: false` and it does. So the skip is about *any* channel completing the iteration, not about the prose signal.

**Docs-only, proven rather than asserted** — borrowing the tree-hash technique from round 4's rebase proof:

```
git diff --stat b310cfad b72c2c86  →  1 file changed (loop-nodes.md)
packages/workflows/src  IDENTICAL
packages/web/src        IDENTICAL
scripts                 IDENTICAL
```

Every round-4 finding therefore stands by construction; nothing needed re-verifying.

**R11** — left alone, yours to file. Noted that it reproduces on `dev` and needs `executeLoopGroupNode` covered too.

`bun run validate` — exit 0 at `b72c2c86`.

Wirasm:
### Sites 7 and 8 fixed in `0a12490b` — plus two more, found by searching for the claim repo-wide

**Site 8 is the one that mattered**, and your framing of *why* eight sweeps missed it is the transferable part: it lives in a doc about **CLI commands**, inside a parenthetical about event emission. No sweep of the loop docs reaches it.

> "loop completion is decided solely by the `until` signal and `until_bash`, never by emitted events"
> → "decided solely by the loop's own declared completion channels, never by emitted events"

It ships in the binary. Without this edit the release would have told an agent that the headline feature of this PR does not exist.

**Site 7** — `loop-nodes.md:204`, "Optional when `until_bash` is set" plus a two-channel load claim, in a file that states the correct rule 120 lines earlier. Now: "Optional, like every completion channel… what the loader requires is that *at least one* is declared."

**Two more, from a repo-wide search for the claim rather than for a file or a keyword:**

- `AGENTS.md` still described the node type as "`loop:` (iterative AI prompt until completion signal)" — now names all three channels.
- `schemas.test.ts` had a test literally called *"…rejected, naming both"* asserting the message names two channels, when a `loop:` message now names three. Renamed and **strengthened** to assert `loop.until_field` appears — an author told about two would not learn the channel this PR added exists.

All four state the rule rather than enumerate, so a fourth channel leaves every one of them true.

**Scope, proven rather than asserted:**

```
git diff --name-only b72c2c86 0a12490b -- 'packages/**/src/**' | grep -v '\.test\.'
  → only packages/docs-web/.../loop-nodes.md

packages/web/src   IDENTICAL   scripts   IDENTICAL
```

No production source changed. The single non-doc edit is `schemas.test.ts`: a test rename plus one added assertion. Every round-5 finding therefore stands by construction.

**On the severity disagreement** — noted, and I think recording it beat averaging it. For calibration: I'd have taken these at either grade, because the cost was minutes against a held PR and the downside was a shipped contradiction. The label wasn't load-bearing in the decision.

**#2600** — the PR body now opens with the eight-site table and the argument: the failure isn't carelessness, it's that a hand-maintained claim has no canonical location to sweep, so correctness depends on guessing every place a past author restated it. Two things in this diff follow from that — every restatement now states the rule instead of listing channels, and the one claim that could be mechanized was: the engine↔builder channel rule is enforced by `scripts/node-ref-parity.test.ts` executing both encodings, not by matching test-case names.

`bun run validate` — exit 0 at `0a12490b`.

Wirasm:
### Rebased onto `dev@b5baa16b` (#2581) — findings stand by construction

Seven-file overlap with #2581. Both branches rebased; **nothing retargeted, nothing merged.**

| | Part A #2583 | Part B #2588 |
|---|---|---|
| Head | `ae3cf8c9` | `535d7be5` |
| Base | `dev` (`b5baa16b`) | `feat/loop-completion-2563` |
| `bun run validate` | exit 0 | exit 0 |
| CI | **all five green** | skipped (non-`dev` base) |

#### Diff-equivalence

**Part A** — excluding `CHANGELOG.md`, the pre- and post-rebase diffs are **identical across all 958 changed lines**. The CHANGELOG differs by exactly two things, both expected: my patch no longer *adds* a `### Changed` heading (#2581 opened that section first) and two blank lines move with it. The entry text is unchanged, including the R9 tightening — `Every existing workflow still loads, and no loop's completion verdict changes.`

**Part B** — excluding `AGENTS.md`, **identical across all 2937 changed lines**. `AGENTS.md` differs on both sides of one hunk because #2581 reworded the same 6.5k-character node-types line (the `effort`/`thinking` clause). Proven equivalent at the semantic level instead:

```
pre-rebase   applies exactly the loop substitution and nothing else: True
post-rebase  applies exactly the loop substitution and nothing else: True
```

and the resolved line is byte-for-byte `dev`'s line plus that one substitution — so #2581's `effort` rewording survived intact and my clause rode on top.

#### The specific hazard: eight of the ten claim sites live in files #2581 edited

`parameter-matrix.md` and `workflow-dag.md` were both touched by #2581, so a conflict resolution taking the wrong side would have resurrected a claim it took five rounds to kill — in files that ship in the binary. Rather than trust the resolution, I built a checker that greps all eight phrasings of the stale claim (must be absent) and eleven corrected statements (must be present), baselined it before the rebase, and re-ran it after:

```
clean   [output_format dropped on loop]      present [cli-commands rule]
clean   [loop ignores output_format]         present [parameter-matrix table cell]
clean   [output_format no-op on loop]        present [parameter-matrix legend]
clean   [solely until+until_bash]            present [workflow-dag loop capability]
clean   [required unless until_bash]         present [dag-advanced exception]
clean   [loops are the acknowledged gap]     present [rulecheck three tiers]
clean   [all three AI producers]             present [AGENTS three tiers]
clean   [loop until completion signal only]  present [loop-nodes ×4]

ALL CLAIM CHECKS PASS
```

Two of its patterns fired on the *corrected* text on first run — false positives from my own regexes — so the checker was tightened to exclude corrected and historical-tense context before being trusted. Separately: none of #2581's added lines under `.claude/skills/` mention loop completion or `output_format` at all; its skill edits are entirely `effort`/`thinking`/`modelReasoningEffort`, so it introduced no new contradiction.

#### `scripts/effort-ladder-parity.test.ts`

#2581's new guard — the third on this pattern — ran against this tree for the first time and passes, unmodified. All three parity guards together: 93 pass, 0 fail.

Wirasm:
### Rebased onto `dev@d6c102b4` after Part A's squash merge — head `65e5f9c0`, all five checks green

`mergeable: true`. The DIRTY state is resolved: Part A landed as a squash, so its commits no longer existed for git to match; replaying only Part B's own eight commits onto `dev` fixes that. **No conflicts** — Part A's content in `dev` is byte-identical to what Part B was stacked on.

#### Diff scope is Part B only

The Part-B-only diff is **byte-identical** to the pre-rebase one apart from hunk offsets and context — 2941 changed lines, no differences at all:

```
git diff ae3cf8c9...535d7be5   (pre-rebase, Part B on Part A)
git diff origin/dev...65e5f9c0 (post-rebase, Part B on merged dev)
→ BYTE-IDENTICAL
```

So every review finding stands by construction. Part A's files no longer appear against `dev` except where Part B genuinely modifies them — `structural.ts` adds `until_field` to the channel rule, `LoopFields.tsx` adds the input, `wire.ts` **deletes** Part A's override (the R2 resolution), `loop-nodes.md` carries the R12 rewording.

#### Part A's content is not duplicated — verified by grep, not by reading

A squash-rebase is exactly where a duplicate slips in, so each Part A marker was counted in the working tree; every one appears **exactly once**:

```
OK  1  const isNonBlank                          OK  1  If your `until_bash` accumulates state
OK  1  must be a non-blank completion signal     OK  1  No channel vetoes another
OK  1  must be a non-blank shell script          OK  1  At least ONE completion channel is required
OK  1  non-blank 'until' signal (builder)        OK  1  loop.until_bash && !signalDetected
```

And the short-circuit reads correctly in its three-channel form:

```
3691:  if (group.until_bash && !signalDetected) {                    ← loop_group, two channels
5166:  if (loop.until_bash && !signalDetected && !fieldComplete) {   ← loop, three
```

CHANGELOG carries one copy of each entry, sections ordered Breaking / Added / Changed / Fixed.

#### The ten claim sites

Re-ran the same checker — the tightened one, not a rebuild — eight stale phrasings absent, eleven corrected statements present: **ALL CLAIM CHECKS PASS**. Nothing from `dev`'s side of `parameter-matrix.md` or `workflow-dag.md` resurrected a shipped contradiction.

#### CI — first real run for this PR

All five green. `test (windows-latest)` failed on the first attempt with three timeouts at ~5015 ms in `generate-bundled-defaults: untracked-file guard (#1578)` — Bun's 5000 ms per-test limit on a test that drives the real script through `spawnSync`. Established as environmental before re-running rather than after: the PR does not touch the generator or its test, the same suite passes locally (20/20), and every other recent run on `dev` and other branches is green. Re-ran the job only; it passed.

`bun run validate` — exit 0. Not merged; that call is Rasmus's.

PR #2604, opened 2026-08-18
feat(workflows): a composed workflow now runs with the configuration its own file declares

### LINKED ISSUE #1764: feat(workflows): compose statically named workflow files inside the parent run

### LINKED ISSUE #2476: decide(workflows): systemPrompt: and agents: interpolate nothing at run time — substitute, or document as literal
`systemPrompt:` and `agents:` are handed to the provider **verbatim**. No workflow-variable substitution, no `$nodeId.output` resolution — nothing. `$ARTIFACTS_DIR` in a `systemPrompt:` reaches the model as the literal seven-dollar-sign string, not a path.

## Verified on `feat/include-inputs` (`eb95862f`)

`dag-executor.ts`:

```
1108:  if (node.systemPrompt !== undefined) baseOptions.systemPrompt = node.systemPrompt;
1122:    agents: node.agents,
1134:    systemPrompt: node.systemPrompt,
```

Three raw passthroughs. Compare `prompt`, which goes through `substituteWorkflowVariables` then `substituteNodeOutputRefs` before reaching the provider (`dag-executor.ts:1432`).

## Why this surfaced now, and why it's ambiguous

#2467 added `$INPUTS` load-time expansion and correctly walks `systemPrompt` / `agents.*.prompt` / `agents.*.description` — because `$INPUTS` has no runtime pass at all, so load time is its only chance. The docblock at `include-expander.ts:226` calls the rewrite's omission of these surfaces "blind spots ... tracked on its own."

That framing assumes the omission is an oversight. It may not be. `systemPrompt` is a *system* prompt — arguably static role text, not per-run content, and never intended to interpolate. **This issue is a decision, not a defect report.** Both answers are defensible; what is not defensible is the current state, where `$INPUTS` interpolates in a `systemPrompt:` and `$ARTIFACTS_DIR` silently does not. An author who sees one work will reasonably assume the other does.

## The two options

**Substitute.** Run `systemPrompt` and `agents.*.prompt`/`.description` through the same pipeline as `prompt`, and add them to the loader's ref scan and to `rewriteNodeOutputRefs`. Consistent, and makes `systemPrompt` genuinely useful for per-run context. Cost: four enumerations to update instead of three, and `agents` is a nested record so the walk is not a one-liner.

**Document as static.** Say plainly in the authoring guide and in the schema docblocks that these fields are literal and interpolate nothing — then explain why `$INPUTS` is the exception (it is resolved at *load* time, before the field is a field, so it is not really an exception at the runtime layer). Cheaper, and defensible for `systemPrompt`. Weaker for `agents.*.prompt`, which reads much more like an ordinary prompt.

My weak preference is **substitute**, because a sub-agent's `prompt:` behaving differently from a node's `prompt:` is the kind of asymmetry authors discover by writing a broken workflow. But this is a language-surface call, not a bug fix — worth a decision before anyone implements.

## Acceptance (either way)

- The behaviour is identical across `systemPrompt`, `agents.*.prompt`, `agents.*.description` — no partial adoption
- The authoring guide states which variables work in these fields
- If substitution is added, the `KEEP IN SYNC` comments in `loader.ts` and `include-expander.ts` list the new surfaces and agree on the count

Found reviewing #2467, alongside the `approval.on_reject.prompt` gap — which is a straightforward defect and is filed separately.

Wirasm:
## Problem and outcome

Composing a workflow rewrote it underneath its author. `warnDroppedWorkflowLevelFields` discarded every workflow-level field of an included file and merely WARNed, so a block declaring `provider: codex` / `model: large` / `effort: high` silently ran on the *parent's* provider and model — and a block declaring nothing inherited the parent's values instead of resolving from config as it would standalone. That is why `include:` felt like "borrow this file's nodes" rather than "run this workflow here", and it is the reason the #2123 defaults overhaul cannot express its curated set as composed workflows instead of copies.

- **Outcome:** A workflow produces byte-identical effective per-node configuration whether it runs standalone or is composed into a parent declaring different values — at any nesting depth and inside `loop_group` bodies. Parent on Claude, one block on Pi, another on Codex, all in one run.
- **Invariant:** A node executes with the configuration its own workflow file declares, regardless of which workflow composed it, or whether it was composed at all. Only run-scoped decisions — isolation, interactivity, evidence policy — belong to whoever started the run.
- **Scope boundary:** Nothing about `workflow:` or `fan_out:` changed. `workflow:` remains the explicit governed launch, which is what removes the entire live-run migration surface — no execution-model version gate, no child-run reinterpretation. `subrun.test.ts` and the fan-out probes are untouched.

## Review guidance

- **Feedback requested:** Correctness of the collapse rule, and whether the one acknowledged gap (`webSearchMode`) and the one narrow divergence (below) are the right calls.
- **Start here:** `packages/workflows/src/include-expander.ts:270` — `collapseWorkflowScope`. The *removal* of the workflow-level layer, not the push-down, is the load-bearing half: push-down alone still lets a block that declares nothing inherit the parent (28 of the spike's 107 divergences).
- **Review order:** `include-expander.ts` (collapse, `requires:` union, composed-gate check, the composed-node stamp) → `node-model-resolution.ts` (the resolution chain, extracted) → `dag-executor.ts` (consumes that chain; composed-input env delivery; session boundary; `systemPrompt`/`agents` substitution) → `schemas/dag-node.ts` + `loader.ts` (launch-only rejections, ref-scan surfaces) → surfaces (`cli/commands/workflow.ts`, `routes/api.ts`, `dry-run.ts`) → docs.
- **Lower-attention areas:** `loader.test.ts` is mostly mechanical — 15 tests that asserted **parse** output *through* `discoverWorkflows` (which also runs expansion) are re-seated onto `parseWorkflow`, so they now test what they name. `test-utils.ts` runs the real expander so fixtures have discovery's actual shape.
- **Known risk or uncertainty:** One narrow divergence, chosen fail-safe and documented at `workflowModelTravelsTo`. When a workflow declares `model:` but **no** `provider:`, a node that explicitly names a provider no longer inherits that model — load time cannot see `config.assistant`, so it cannot tell whether the two agree. The cost is that such a node falls back to its own provider's configured default in the one case where they did agree; the alternative was handing one provider's model string to another. The repo-wide sweep shows this fires on 0 nodes today.

## Solution

Each workflow's node-affecting configuration — `provider`, `model`, `effort`, `thinking`, `fallbackModel`, `betas`, `sandbox`, `persist_sessions` — is written onto its **own** nodes at expansion time, and then removed from the definition. Node values always win; the collapse runs at every level and write-once, so a node carries what its own file said. Because the workflow-level layer is gone, a node with no value of its own resolves from config, tier presets and per-user AI prefs at run time, exactly as it would standalone. Display surfaces read the authored values instead (`WorkflowWithSource.declared`); execution reads the collapsed DAG.

`model:` is the one field with a condition: it travels only to nodes that will run on the workflow's own provider, mirroring the executor's existing `provider === workflowProvider` rule so a provider-overriding node never inherits another provider's model string. The regression guard caught this — the naive copy the plan carried would have handed `gpt-5.6-sol` to a Claude node.

`requires:` unions upward rather than being dropped: it is a fact about what the nodes need, not a choice the run makes, and dropping it converted a clean pre-cost refusal into a mid-run failure inside a block the parent cannot inspect. Run-affecting fields stay workflow-level and the drop-warning narrows to exactly them.

Six consequences follow, each small and separable:

- **A composed `approval:` gate cannot be silently undriveable.** `interactive:` is read off the *top-level* workflow, so a composed gate in a non-interactive parent was dispatched to the background — still resumable from the CLI, but not driveable inline, which is the one thing `interactive:` exists to guarantee. It is now a load error naming the block and the gate. Fires on 0 workflows in this repo today.
- **A composed workflow's entry node starts a fresh session**, as coldly as it would standalone, instead of inheriting the parent's sequential-session cursor. `context: shared` is the per-node opt-out. Deliberately not `context: 'fresh'`, which silently disables `persist_session`.
- **Declared inputs reach `bash:`/`script:` nodes as `INPUTS_<UPPER_SNAKE>`**, including a **named** script file — whose source is opaque, so the load-time `$INPUTS` macro could never reach it. Delivered through the env bag (never spliced into source), at two nesting levels, with the inner workflow's own input names.
- **`systemPrompt:` and `agents.*` became ordinary runtime substitution surfaces** (#2476). They took `$INPUTS` at load but nothing at run time, so a workflow using `$INPUTS.mode` in a `systemPrompt:` resolved when composed and stayed literal standalone — the same asymmetry this PR exists to remove.
- **Launch-only options fail loudly on an `include:` node.** `isolation`, `fan_out`, `input`, `mutates_checkout` express an intent composition cannot honour; a warning would let an author believe they got isolation they did not get. Merely-unread fields still only warn.
- **Resolution is legible.** `workflow dry-run` reports each node's effective provider/model and where each value came from, through the *same* resolver the executor uses — a second copy of the chain would drift, and a drifted report is worse than none. This is why requiring `provider`/`model` on every node was rejected: legibility instead of ~54 redundant lines in a 27-node workflow.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A composed block ran on the composing workflow's provider/model/effort; its own were dropped with a WARN | A composed block runs on what it declares; declaring nothing resolves from config/tiers/user prefs, not from the parent |
| **Failure behavior** | A block's `requires: [github]` was dropped, so an unconnected user failed mid-run inside the block; a composed approval gate in a background run was silently undriveable inline | `requires:` unions and refuses at invocation before any cost; a composed gate without `interactive: true` is a load error naming both |

## Architecture

```mermaid
flowchart TD
  subgraph before["BEFORE — composition rewrites the workflow"]
    PB["parent.yaml<br/>provider: claude"]
    BB["review.yaml<br/>provider: codex, model: large"]
    PB -->|"include:"| BB
    BB -->|"workflow-level fields DROPPED (warn)"| NB["review__scope<br/>runs on CLAUDE"]
  end

  subgraph after["AFTER — the workflow runs as authored"]
    PA["parent.yaml<br/>provider: claude"]
    BA["review.yaml<br/>provider: codex, model: large"]
    PA -->|"include:"| BA
    BA -->|"collapsed onto its own nodes"| NA["review__scope<br/>provider: codex, model: large<br/>runs on CODEX"]
    PA --> NP["parent's own nodes<br/>provider: claude"]
  end

  classDef bad fill:#ffeaea,stroke:#c53030,stroke-width:2px
  classDef good fill:#e8f7ee,stroke:#2f855a,stroke-width:2px
  class NB bad
  class NA good
```

### Changed seams

|---|---|---|
| `expandWorkflowIncludes → executeDagWorkflow` | The expanded definition carries **no** node-affecting workflow-level fields; every node states its own. The byte-for-byte fast path for include-free workflows is gone — a workflow must not behave differently depending on whether it happens to contain an `include:` | `include-expander.ts` `collapseWorkflowScope`; `include-expander.test.ts` "include-free workflows are collapsed too" |
| `discovery → display surfaces` | New `WorkflowWithSource.declared` carries the authored `provider`/`model`/`effort`. Not a duplicated representation — the collapsed definition no longer holds these fields at all | `workflow-discovery.ts`, `schemas/workflow.ts`; `resolveTitleAssistantType` regression test in `packages/cli/src/commands/workflow.test.ts` |
| `resolveNodeProviderAndModel → node-model-resolution.ts` | The chain is extracted pure; the executor keeps the parts a dry run must not do (warn, throw, build options) and dry-run calls the same function | `node-model-resolution.ts`; `dry-run.test.ts` "effective provider/model per node" |
| Engine-private per-node metadata (`COMPOSED_NODE`) | New symbol payload carrying a node's authoring workflow, its resolved inputs, and its block-entry position. Symbols do not ride `structuredClone`, so it is re-attached in `cloneNodeForInclude` and re-walked by **both** rewrite passes | `compiled-command.ts`; the two-level nesting tests in `include-expander.test.ts` and `dag-executor.test.ts` |
| Loader ref-scan / `rewriteNodeOutputRefs` / `substituteNodeOutputRefs` | `systemPrompt` and `agents.*.prompt`/`.description` added to all three. The `KEEP IN SYNC` block's claim that `applyInputsMacro` is an intentional superset is now false and was rewritten, not merely satisfied | `loader.ts`, `include-expander.ts`, `dag-executor.ts` comment blocks |

## Validation

- `bun run validate` — passed (exit 0, zero test failures) — proves types, lint at `--max-warnings 0`, format, generated-file ratchets and the full per-package suite.
- `bun --filter @archon/workflows test` — passed — the invariant suite (`a workflow runs as authored, standalone or composed`) compares a block run standalone against the same block inside a parent declaring different values, at the `sendQuery(prompt, cwd, resume, options)` seat every other option test uses, with **two** pass conditions per case: the invariant (composed == standalone) *and* a regression guard (raw standalone == collapsed standalone). The guard is not decoration — it is what caught the `model`-across-providers defect.
- **Repo-wide no-op sweep** — **63 workflows / 421 nodes → 0 field divergences, 0 discovery errors.** Compares the old resolution (`node.X ?? workflow.X`, computed from the raw parse) against the collapsed definition for every workflow in this repo. Nothing shipped moves, and the new composed-gate check fires on 0 today.
- **Manual, real CLI, one composition across two packs** (composed block on `codex`/`gpt-5.6-sol`, parent on `claude`/`opus`, a **named** script in the block's own pack, an inline bash reading the macro form, and an approval gate):
  - `archon workflow list --json` → block reports `codex`/`gpt-5.6-sol`, parent reports `claude`/`opus` — both absent from the executed definition.
  - `archon workflow run … --dry-run` → `blk__think` → `runs on: codex (node) / gpt-5.6-sol (node) [from smoke-block]`; the parent's own node → `claude (node) / opus (node)`. Two providers, one run, one command.
  - Removing `interactive: true` → `Node 'blk__gate': composed workflow 'smoke-block' contains an approval gate, but this workflow does not declare 'interactive: true'…` and the workflow drops out of `list`.
  - `archon workflow run --no-worktree` → the named script in the block's **own** pack ran and printed `plan=SETUP|base=dev` (a `with:`-forwarded value plus the block's declared default), the gate paused, `archon workflow approve <id>` resumed the **parent** skipping 4 completed namespaced nodes, and `archon workflow runs --json` shows **1 run row** with `parent_run_id` null.
- **Not verified:** No live AI turn was billed. Effective per-node provider/model/effort/thinking/sandbox/betas/fallbackModel are asserted at the `sendQuery` call — the same observation point the existing option tests use — so a real turn would add cost without adding evidence about the collapse. Container and folder-project backends were not exercised; the change is a load-time transform with no isolation surface.

## Delivery considerations

|---|---|---|
| Compatibility / migration | **No migration for live runs.** No execution-model gate, no child-run reinterpretation — the collapse is a load-time transform of a definition re-derived on every entry point, and node **ids** are untouched, so a snapshot written by an older build still matches on resume. Behavior changes only where a composed workflow declared config that was previously discarded, which is the fix | 421-node sweep → 0 changed; `AC13 — a run paused BEFORE the collapse resumes with collapsed config afterwards` in `dag-executor.test.ts`; manual approve/resume above |
| Compatibility / migration | `requires:` union can make a parent refuse **earlier**, at invocation, on per-user-GitHub installs only. A set union can only refuse sooner, never later | `include-expander.test.ts` "requires: unions instead of dropping" |
| Security / permissions / data | Composed input values ride the env bag and are never spliced into script source — the sanctioned injection-safe channel (#2115). Ordering is preserved: project env < run inputs < this node's composed inputs < engine-reserved keys, with one shared helper rather than a third divergent copy | `AC8 — a composed input value resolving a node ref is delivered, not shell-spliced` and `… a project env var cannot shadow a composed input` |
| Rollout / rollback | One commit, revertable whole. The collapse itself is one file | — |
| Observability | The drop-warning narrows to run-owned fields and names the `webSearchMode` hole explicitly instead of letting it read as a run-level decision; `dry-run` reports per-node resolution and its origin | `loader.test.ts` drop-warning tests; `dry-run.test.ts` |
| Documentation / communication | The guide's two sections are renamed for what they do (`Composing Another Workflow with include:` / `Launching a Separate Governed Run with workflow:`), gain a three-difference decision table (the gate, command-body freshness, `interactive:`), state the rule in one sentence, and replace the named-script paragraph that is false after this change. Two case-law rows added to the constitution | `packages/docs-web/…/authoring-workflows.md`, `…/workflow-language-constitution.md`; `bun --filter @archon/docs-web build` passed |

## Links

- Closes #1764
- Closes #2476
- Plan: https://github.com/coleam00/Archon/issues/1764#issuecomment-5313423603

## Summary by CodeRabbit

* **New Features**
  * Improved workflow composition with configuration inheritance, input delivery, session boundaries, approval handling, and requirement propagation.
  * Dry-run results now show effective provider, model, and reasoning settings for AI nodes.
  * Workflow listings and execution preserve declared workflow-level configuration.

* **Bug Fixes**
  * Added validation and clearer warnings for unsupported composed-workflow options.
  * Improved runtime substitution for prompts, agent settings, and composed inputs.
  * Background runs now stop safely when composed approval gates cannot be completed.

* **Documentation**
  * Expanded guidance on `include:` versus `workflow:` behavior, governance, inputs, sessions, and limitations.
  * Updated workflow reference links and terminology.


Wirasm:
---
pr: 2604
base: dev
head: feat/1764-composition-config-collapse
reviewed: 2026-08-18T11:47:18Z
reviewed_head: b2110b657263f5bb42c6ee4263dfdfdba6c107a6
verdict: READY TO MERGE
open_findings: 0
total_findings: 21
scopes: [code, seams, tests, comments, errors, types, docs, simplify]  # round 2: code, seams
publication: https://github.com/coleam00/Archon/pull/2604#issuecomment-5327057085
---



# PR Review: #2604 — feat(workflows): a composed workflow now runs with the configuration its own file declares

## Signal

**Round 2 (code + seams) re-reviewed the corrections at `6ed510a9` and found one real
regression the correction pass introduced.** Both agents reached it independently: moving
the composed-gate rule from load time (universal) to invocation narrowed it to ONE of the
two functions that background a run. The `manage_run` tool's `startWorkflow` calls
`dispatchBackgroundWorkflow` directly — on any platform with native tools, not just web —
with no check, reproducing the exact R1 failure mode on an unguarded path (R20). Fixed by
moving the assertion inside `dispatchBackgroundWorkflow` itself, so the rule holds for
every entrypoint rather than for a caller list that has to stay complete.

Round 2 also cleared the other hazard worth worrying about: both agents traced discovery →
dispatch and found no `structuredClone` or JSON round-trip in between, so the
`ComposedNodeMeta.origin` symbol the gate depends on survives intact. The hole was a
missing call, not a lost symbol.

Two smaller round-2 findings: the builder's ref scan was still missing three surfaces the
engine scans (R13 residue), and `resolveWorkflowModelScope` reported a provider origin that
named the overridden value rather than the winning one (R21).

**All 21 findings resolved: 20 FIXED, 1 DECLINED.** Head `0eb1cecd` → `b2110b65`,
14 commits.

- **Blocking:** 0
- **Non-blocking:** 0
- **Resolved since previous review:** 21 (20 FIXED, 1 DECLINED)
- **Tracked follow-ups:** [#2606](https://github.com/coleam00/Archon/issues/2606) — the container preflight still hand-mirrors provider resolution (pre-existing; this PR made it cheap to retire)
- **Validation:** `bun run validate` PASS at `b2110b65`; sweep still 421 nodes → 0 divergences.

## Findings

|---|---|---|---|
| `R1` | Critical | Nested composition with an approval gate never loads, even when the top-level workflow declares `interactive: true` | FIXED |
| `R2` | Important | `persist_sessions` collapses into `loop_group` bodies, overriding the executor's deliberate `workflowPersistSessions: false` | FIXED |
| `R3` | Important | Windows CI fails on a path-separator assertion in a test this PR adds | FIXED |
| `R4` | Important | One workflow-level provider/model conflict now emits N chat warnings; `executor.ts`'s workflow-level warning is dead code | FIXED |
| `R5` | Important | `resolveWorkflowModelScope` is never wired into `executor.ts` — the PR added a third copy of the chain instead of removing one | FIXED |
| `R6` | Important | `workflowModelTravelsTo`'s disclosed divergence understates the reachable case (tier/alias-resolved provider) | FIXED |
| `R7` | Suggestion | `providerConflict` is computed then discarded by dry-run, contradicting its own docblock | FIXED |
| `R8` | Suggestion | `applyInputsMacro`'s docblock still claims the SUPERSET design three other blocks now deny | FIXED |
| `R9` | Suggestion | The `delete` vs `= undefined` comment states a reason that is factually false | FIXED |
| `R10` | Suggestion | Two JSDoc blocks stacked on `composedInputResolverFor`; `inputEnvVars` left undocumented | FIXED |
| `R11` | Suggestion | `NODE_AFFECTING_WORKFLOW_FIELDS` is untyped `string` pairs bridging two schemas, with no exhaustiveness guard | FIXED |
| `R12` | Suggestion | Dry-run's `loadConfig` catch downgrades silently, against this file's own stated convention | FIXED |
| `R13` | Suggestion | The web builder's `content.ts` is an unupdated 4th consumer of the node-ref surface set | FIXED |
| `R14` | Suggestion | `reference/variables.md` and `reference/cli.md` never learned about the new surfaces | FIXED |
| `R15` | Suggestion | `composedInputResolverFor` is a factory with no reuse at either call site | FIXED |
| `R16` | Suggestion | The drop-warning docstring claims a field "either travels or shows up here"; partial travel is a silent third path | FIXED |
| `R17` | Suggestion | Five specific coverage gaps in the new collapse behaviour | FIXED |
| `R18` | Suggestion | AI-config fields are pushed onto `bash:`/`script:`/`cancel:`/`workflow:` nodes that never read them | DECLINED |
| `R19` | Suggestion | Missing return-type annotations on new helpers/callbacks; two doc-wording clarifications | FIXED |
| `R20` | Critical | The `manage_run` tool backgrounds a run without the composed-gate check — R1's failure mode on a second, unguarded path | FIXED |
| `R21` | Suggestion | `resolveWorkflowModelScope` reported the provider origin as `workflow` when a preset overrode it | FIXED |

## Detailed Findings

<details>
<summary><code>R1</code> — Nested composition with an approval gate never loads</summary>

**Impact:** A three-level composition is dropped from discovery entirely — not warned, unusable — even when the workflow that owns the run correctly declares `interactive: true`. This breaks the documented "up to 3 levels deep" nesting (`INCLUDE_MAX_DEPTH = 3`) for any chain with an approval gate below the first level, and it contradicts the guide this same PR adds.

**Evidence:** `packages/workflows/src/include-expander.ts:968` calls `findUndriveableComposedGate(newNodes, collapsed.interactive === true)` inside `expandOne`, which runs once per workflow name — for direct discovery *and* as another workflow's include target. An intermediate building block has no reason to set `interactive:`, and it is never the run owner.

Reproduced twice independently. Direct execution against `expandWorkflowIncludes` with `inner` (approval gate) → `mid` (no `interactive`) → `top` (`interactive: true`):

```
CR-1 top loads?  false
CR-1 mid loads?  false
mid: Node 'i__gate': composed workflow 'inner' contains an approval gate,
     but this workflow does not declare 'interactive: true'
top: Node 'm': Node 'i__gate': …
```

Contradicted by `packages/docs-web/src/content/docs/guides/authoring-workflows.md` — "the **top-level** workflow must declare `interactive: true`" — which is never what is actually checked past the first level.

**Required outcome:** Evaluate interactivity against the workflow that owns the run, not against each expansion level. Load time cannot know which workflow will be invoked, since any discovered workflow can be run directly — so this likely means the expander warns and the invocation path (where the run owner is known) enforces. That is a scope decision for the author, not a mechanical fix. Whatever shape is chosen, add the three-level regression test both reviewers noted is absent.

**Found by:** `prp-core:code-reviewer`, CodeRabbit (Major); adjacent coverage note from `prp-core:pr-test-analyzer`

**Disposition:** **FIXED** in `35314613`. The rule moved out of `expandOne` into `assertComposedGateDriveable` (utils/workflow-requirements.ts), called from the orchestrator's web background-dispatch branch. Expansion keeps stamping the gate's origin. The move also corrected a second defect: the load-time form refused CLI and chat runs, where the gate is presented normally and never needed refusing. Tests cover the intermediate block expanding, the run owner still refusing, and the loop_group-body case.

</details>

<details>
<summary><code>R2</code> — <code>persist_sessions</code> collapses into <code>loop_group</code> bodies</summary>

**Impact:** A body node in a workflow declaring `persist_sessions: true` now uses a persisted session scope where it previously did not, and can trip the runtime `sessionResume` capability guard and fail a node that used to run.

**Evidence:** `packages/workflows/src/include-expander.ts` `pushWorkflowScopeOntoNodes` recurses into `node.loop_group.nodes` for every field including `persist_sessions` → `persist_session`. But `packages/workflows/src/dag-executor.ts:3675` deliberately constructs a loop-group body context with `workflowPersistSessions: false`, and `nodeUsesPersistedScope` (`dag-executor.ts:6911-6916`) resolves `nodePersist ?? workflowPersistSessions` — so a pushed node-level `true` overrides the executor's deliberate `false`.

Reproduced: a workflow with `persist_sessions: true` and a `loop_group` body prompt node yields `body persist_session = true` after collapse.

**Required outcome:** Do not push `persist_sessions` into `loop_group` bodies; preserve an explicit body-level `persist_session` if one is written. The `isPersistableNode` gate was the right idea at the wrong depth.

**Found by:** CodeRabbit (Major), verified by reproduction during aggregation

**Disposition:** **FIXED** in `3456c9e5`. `pushWorkflowScopeOntoNodes` takes an `insideLoopGroup` flag and skips `persist_session` there. Regression test asserts the body node is untouched while the top-level AI node still receives the default.

</details>

<details>
<summary><code>R3</code> — Windows CI fails on a path-separator assertion</summary>

**Impact:** A required check fails. Deterministic on Windows, not flaky.

**Evidence:** `packages/workflows/src/loader.test.ts:160` asserts the discovered script path with `join(blockDir, 'scripts', 'report.ts')`, which yields backslashes on Windows, while `packages/workflows/src/script-discovery.ts:137` deliberately stores every path through `normalizeSep()` (forward slashes on every OS).

```
(fail) a composed block's named script still resolves to its OWN pack after composition
Expected: "C:\Users\...\shared\report-block\scripts\report.ts"
Received: "C:/Users/.../shared/report-block/scripts/report.ts"
```

Exactly one test fails in the job; `windows-latest` passed on the last five `dev` runs including this PR's base `b3f97356`. The behaviour under test passed — the received path is the correct file in the block's own pack; only the separator differs.

**Required outcome:** Normalize the expected side. `packages/workflows/src/script-discovery.test.ts:42` already carries the precedent — a local `norm = p => p.replaceAll('\\', '/')` with a comment stating exactly this reason.

**Found by:** `validation`

**Disposition:** **FIXED** in `e5a93017`. Expected side normalized, matching script-discovery.test.ts's existing `norm` helper.

</details>

<details>
<summary><code>R4</code> — One conflict becomes N chat warnings; the workflow-level warning becomes dead code</summary>

**Impact:** Two consequences of the same omission. A workflow declaring `provider: claude` with a tier `model:` that resolves elsewhere used to produce **one** clear workflow-scope warning; it now produces **one per inheriting node** in the user's chat. Separately, `executor.ts`'s workflow-level conflict warning and `workflowPreset` resolution are unreachable for every workflow that went through real discovery.

**Evidence:** `collapseWorkflowScope` deletes `provider`/`model` from every discovered workflow, but `packages/workflows/src/executor.ts:1292-1352` (not in this diff) still resolves `resolvedProvider`/`resolvedModel`/`workflowPreset` from `workflow.provider`/`workflow.model`. Both are now always `undefined`, so that block always falls through to `config.assistant` and the message at `executor.ts:1315-1325` can never fire. Meanwhile the collapse writes both `provider` and the conflicting `model` onto every node, so `resolveNodeModel`'s conflict check (`node-model-resolution.ts:104-110`) fires per node, surfaced via `dag-executor.ts:1091-1109`'s `safeSendMessage`.

Reproduced on a 3-node workflow (`provider: claude`, `model: large` → tier resolves to codex):

```
per-node fields  : a{claude,large} b{claude,large} c{claude,large}
conflicts raised : 3 of 3 nodes -> {"declared":"claude","resolved":"codex","modelRef":"large"}
wf provider/model after collapse: undefined undefined
```

**Required outcome:** Decide where the workflow-level conflict is reported now that the workflow-level fields are gone, and stop the per-node duplication — e.g. detect the conflict once at collapse time, or de-duplicate the per-node message by `(declared, resolved, modelRef)`. Remove or rewire the now-dead block in `executor.ts` (see R5).

**Disposition:** **FIXED** in `7a80db78`. The dead block is gone with R5's rewiring. The per-node warning is de-duplicated per run by `(declared, resolved, modelRef)` via a `Set` on the run context, shared by reference into loop_group body contexts. Verified red by sabotage: 3 messages for 3 nodes before, 1 after.

</details>

<details>
<summary><code>R5</code> — <code>resolveWorkflowModelScope</code> is never wired into <code>executor.ts</code></summary>

**Impact:** The PR's stated reason for extracting `node-model-resolution.ts` is that a second implementation of the resolution chain would drift and make `--dry-run` silently disagree with a real run. `resolveNodeModel` achieves that at node level. `resolveWorkflowModelScope` does not: it is called only by `dry-run.ts:755`, while `executor.ts:1292-1352` keeps hand-written logic computing the same thing. Its own docblock says "mirroring the block above `executeDagWorkflow`'s call in executor.ts" — mirroring is not calling. So the PR created a third copy of the workflow-level chain rather than removing one, and nothing ties them together.

**Evidence:** `packages/workflows/src/node-model-resolution.ts:180` (the "mirroring" comment); `packages/workflows/src/executor.ts:1292-1352` (untouched by the diff, computing `resolveModelSpec`/`isLiteralSpec`/`assistantDefaults?.model` and a `providerSource` string that is the same three-way classification as `WorkflowModelScope.providerOrigin`). No test asserts the two agree. A third copy lives in the PR's own test harness, explicitly labelled "mirror of executor.ts's workflow-level resolution."

**Required outcome:** Point `executor.ts` at `resolveWorkflowModelScope` and layer its messaging + `isRegisteredProvider` check on top — the same pattern this PR already applied one level down, where `resolveNodeProviderAndModel` wraps the pure `resolveNodeModel`. The test harness's mirror block then disappears too, and R4's dead code resolves with it. If wiring is deferred, an equivalence test between the two is the minimum.

**Found by:** `prp-core:seam-analyzer`, `prp-core:code-simplifier` (independently)

**Disposition:** **FIXED** in `7a80db78`. `executor.ts` calls `resolveWorkflowModelScope` and layers its warning and unknown-provider throw on top, mirroring how `resolveNodeProviderAndModel` wraps `resolveNodeModel`. The docblock's "mirroring" wording is replaced with what it now does.

</details>

<details>
<summary><code>R6</code> — The disclosed divergence understates its reachable case</summary>

**Impact:** The PR discloses one divergence from the old chain and describes it as the `config.assistant` case. There is a second, more reachable case it does not mention: a workflow declaring only a tier/alias `model:` (no `provider:`) plus a node naming that tier's resolved provider. The old chain let the model travel (the *resolved* workflow provider matched); the collapse cannot see the resolved provider at load time and skips it, so the node silently falls back to its provider's configured default.

**Evidence:** `packages/workflows/src/include-expander.ts:206-219`. Reproduced: workflow `{ model: 'large' }` (tier → codex), node `{ provider: 'codex' }` → node `model` is `undefined` after collapse, where the old chain produced the tier's model.

**Required outcome:** No behaviour change required — the fail-safe direction is right. Correct the code comment and the PR description to name the tier/alias case, which is the one an author is likely to hit. Add a test pinning the accepted behaviour either way.

**Found by:** `prp-core:code-reviewer` (flagged for awareness), confirmed by reproduction

**Disposition:** **FIXED** in `6d7bb306`. Behaviour unchanged and deliberate; the comment now names the tier/alias-resolved-provider case as the reachable one and says why resolving tiers at load would be worse. Five branches of where a workflow-level `model:` does and does not travel are pinned, plus the fields that carry no provider condition and must not gain one.

</details>

<details>
<summary><code>R7</code> — <code>providerConflict</code> is computed then discarded by dry-run</summary>

**Impact:** A `--dry-run` of a workflow with a node/model-ref provider mismatch omits the conflict warning a real run produces. The resolved provider it prints is correct, so the report is incomplete rather than wrong.

**Evidence:** `packages/workflows/src/node-model-resolution.ts:51-55` documents the field as "reported here so a dry run can too." `grep providerConflict packages/workflows/src/dry-run.ts` returns nothing — neither `dryRunResolutionSchema` nor `resolutionFor()` carries it.

**Required outcome:** Forward it into the trace, or correct the docblock to say the executor is its only consumer.

**Disposition:** **FIXED** in `4a412b4d`. `providerConflict` is carried in `dryRunResolutionSchema`, forwarded by `resolutionFor`, and rendered by `formatDryRunTrace`.

</details>

<details>
<summary><code>R8</code> — <code>applyInputsMacro</code>'s docblock still claims the SUPERSET design</summary>

**Impact:** The docblock on the function itself — the first place a maintainer looks — asserts a design this PR deliberately ended, directly contradicting three docblocks the PR did update.

**Evidence:** `packages/workflows/src/include-expander.ts:372-384` still reads "This walks a SUPERSET of rewriteNodeOutputRefs' field set, and the extra fields are the point… a surface this function misses is permanent." Contradicted by `include-expander.ts:296`, `loader.ts:490-496`, and `dag-executor.ts:840-842`, all updated to say the sets now agree. The comment analyzer walked both functions field by field and confirmed the code is correct — only this block is stale.

**Required outcome:** Rewrite it to match the other three.

**Disposition:** **FIXED** in `9dc09c52`. Rewritten to match the three other blocks. The failure-mode asymmetry it also documented is real and kept, with a note that a future include-input-only field would reopen the #2476 gap.

</details>

<details>
<summary><code>R9</code> — The <code>delete</code> rationale is factually false</summary>

**Impact:** The comment threatens a breakage that cannot happen, sending a future reader after a non-existent invariant.

**Evidence:** `packages/workflows/src/include-expander.ts:264-266` claims a key left present with an `undefined` value "would still read as declared." `warnDroppedWorkflowLevelFields` already filters `childRecord[key] !== undefined`, so both forms produce identical output. Confirmed by repro:

```
keys after = undefined: [ 'a', 'b' ] -> filtered: [ 'a' ]
keys after delete:      [ 'a' ]      -> filtered: [ 'a' ]
```

**Required outcome:** State the real reason for `delete` (object shape for downstream spreads and serialization) or drop the justification.

**Disposition:** **FIXED** in `9dc09c52`. Replaced with the actual reason: the collapsed definition is spread and serialized, and a present-but-undefined key survives both.

</details>

<details>
<summary><code>R10</code> — Two JSDoc blocks stacked on one function</summary>

**Impact:** `inputEnvVars` — the function carrying the precedence contract — is undocumented, and `composedInputResolverFor` carries a docblock describing something else.

**Evidence:** `packages/workflows/src/dag-executor.ts:160-183`: the env-bag docblock and the resolver docblock both sit above `composedInputResolverFor` at :184; `inputEnvVars` begins at :212 with no docblock. The helper was inserted between a docblock and its function.

**Required outcome:** Move the first block back onto `inputEnvVars`.

**Found by:** CodeRabbit (Minor), verified during aggregation

**Disposition:** **FIXED** in `9dc09c52`, by removing the function that caused it (see R15).

</details>

<details>
<summary><code>R11</code> — <code>NODE_AFFECTING_WORKFLOW_FIELDS</code> is untyped string pairs with no guard</summary>

**Impact:** A rename or typo on either side compiles clean and the value silently stops travelling — and worse than not travelling: `collapseWorkflowScope` deletes the workflow-level key regardless, so a drifted pair *deletes the source and writes it nowhere*. Total silent loss of a config value, against the repo's Fail Fast principle.

**Evidence:** `packages/workflows/src/include-expander.ts:189` types the list as `readonly (readonly [wfKey: string, nodeKey: string])[]`, while both key sets are Zod-derived and exported: `WorkflowBase` (`schemas/workflow.ts:213`) and `DagNodeBase` (`schemas/dag-node.ts:271`), both re-exported from `./schemas`. All 8 pairs are correct today. The existing runtime backstop is only a WARN, and only for a workflow actually consumed via `include:`.

**Required outcome:** Annotate as `readonly (readonly [wfKey: keyof WorkflowBase, nodeKey: keyof DagNodeBase])[]` — type-only, zero behaviour change, one import. Optionally add the exhaustiveness test the repo already uses three lines away in `schemas.test.ts` for `INCLUDE_NODE_IGNORED_FIELDS`, partitioning `workflowBaseSchema`'s keys across the travelling / non-dropped / safety sets plus the `webSearchMode` exception.

**Found by:** `prp-core:type-design-analyzer`, `prp-core:seam-analyzer`

**Disposition:** **FIXED** in `4a412b4d`. Typed `keyof WorkflowBase` / `keyof DagNodeBase`. Verified red: a drifted node key now fails compilation with `TS2820 ... Did you mean 'persist_session'?`. The optional exhaustiveness test was not added — the annotation catches renames, and a genuinely new workflow-level field still surfaces in the drop-warning.

</details>

<details>
<summary><code>R12</code> — Dry-run's <code>loadConfig</code> catch downgrades silently</summary>

**Impact:** A user with a malformed `.archon/config.yaml` runs `--dry-run` to sanity-check before spending real cost and gets a clean-looking report claiming every node resolves to Claude defaults, with the failure only in a Pino log.

**Evidence:** `packages/cli/src/commands/workflow.ts:996-1000` catches and warns, then passes `config: undefined`, so `dry-run.ts` falls back to `options.config?.assistant ?? 'claude'` with no tiers or aliases. The same file states the opposite convention at :1473-1477: *"Do NOT swallow loadConfig errors: a malformed/unreadable config… must FAIL the run, never silently downgrade… loadConfig returns defaults when no config file exists (not an error)."*

**Required outcome:** Surface the config failure in the dry-run's own output — or fail the dry run — rather than reporting against fabricated defaults.

**Found by:** `prp-core:silent-failure-hunter`

**Disposition:** **FIXED** in `4a412b4d`. The catch is removed; `loadConfig` returns defaults when no file exists, so a throw means a broken one and now fails the dry run, matching the convention stated later in the same file.

</details>

<details>
<summary><code>R13</code> — The web builder is an unupdated 4th consumer</summary>

**Impact:** The engine now hard-rejects a workflow whose `systemPrompt:`/`agents.*` references a node outside its upstream set, but the console builder's live reference-scan gives no warning while editing — the author only discovers it at save/run.

**Evidence:** `packages/web/src/experiments/console/builder/validation/content.ts` — its docstring says it mirrors the engine's loader scan, and `textBodies()` enumerates only `prompt`/`command`/`bash`/`script`/`approval.message`/`loop.prompt`, all read from `node.data`. `systemPrompt` and `agents` are live builder fields but live in `node.base` (`variants/base-fields.ts:37,41`). `scripts/node-ref-parity.test.ts` does not cover this — it pins the ref *grammar*, not which fields are scanned.

**Required outcome:** Add the two surfaces to `textBodies()`, or record in the `KEEP IN SYNC` comment that a fourth consumer exists and is deliberately narrower.

**Disposition:** **FIXED** in `5b4ca741`, then **completed** in `9a2952d8`. The first fix added `systemPrompt`/`agents.*` but the round-2 seam pass found the builder still missed three surfaces `loader.ts` scans on variants it *can* author: `loop.until_bash`, `approval.on_reject.prompt`, and a cancel body. All three added. An existing test — "cancel nodes have no scannable body" — asserted that gap as intended behaviour; `loader.ts:546` pushes exactly that field into the ref scan, so it was flipped. (`workflow:`/`loop_group:` surfaces stay out of scope: the builder cannot author those variants.)

</details>

<details>
<summary><code>R14</code> — Two reference pages never learned about the new surfaces</summary>

**Impact:** `reference/variables.md` states its purpose as the complete reference for all variable substitutions and enumerates the surfaces twice — neither mentions `systemPrompt`/`agents`, which this PR turns into both a new capability and a new load-time failure mode. `reference/cli.md`'s dry-run section describes the trace without the resolution report, even though `authoring-workflows.md` now sends authors there to find exactly that.

**Evidence:** `packages/docs-web/src/content/docs/reference/variables.md:15` and its "Variable Availability by Context" table; `reference/cli.md` lines ~232-256. The docs agent confirmed no stale prose remains anywhere else in the tree.

**Required outcome:** Add the two surfaces to `variables.md`; document the resolution report in `cli.md`'s dry-run section.

**Found by:** `prp-core:docs-impact-agent`

**Disposition:** **FIXED** in `3948bbe8`. `variables.md` gains the three surfaces in both places it enumerates them; `cli.md`'s dry-run section documents the resolution report, its origin labels, the composed-workflow attribution and the `--json` shape.

</details>

<details>
<summary><code>R15</code> — <code>composedInputResolverFor</code> is a factory with no reuse</summary>

**Impact:** One function and one local per call site with no benefit — both call sites build the identical args object from variables already in scope and hand the closure to `inputEnvVars` exactly once.

**Evidence:** `packages/workflows/src/dag-executor.ts:184-207`, called at ~2775 and ~3065. The "byte-identical bag" guarantee comes from calling one function, not from currying it.

**Required outcome:** Fold the resolver into `inputEnvVars`, taking the context object as a parameter. Overlaps with R10 — both touch the same block.

**Disposition:** **FIXED** in `9dc09c52`. Folded into `inputEnvVars`, which takes a `ShellInputContext` directly.

</details>

<details>
<summary><code>R16</code> — The drop-warning docstring overclaims</summary>

**Impact:** The docstring says a new workflow-level field "either travels… or shows up here." Partial travel is a third path with no signal: when `workflowModelTravelsTo` or `isPersistableNode` skips a field for a specific node, the key is still deleted from the workflow, so it can never appear in `droppedFields`.

**Evidence:** `packages/workflows/src/include-expander.ts:190-215` (delete is unconditional) versus `:679-683` (the warning derives from the child's remaining defined keys).

**Required outcome:** Correct the docstring, or emit a debug-level signal at the point a field is skipped for a node.

**Found by:** `prp-core:silent-failure-hunter`

**Disposition:** **FIXED (docstring)** in `9dc09c52` — it now states that partial travel is not reported and why that is benign. The second half stays **DECLINED** with the evidence recorded above.

</details>

<details>
<summary><code>R17</code> — Five coverage gaps in the new behaviour</summary>

**Impact:** Each admits a plausible faulty implementation the current suite allows.

**Evidence and required outcome:**
1. `workflowModelTravelsTo`'s explicit-*same*-provider branch is unexercised; dropping that disjunct would silently stop a model reaching a node that redundantly re-declares the same provider.
2. No workflow definition anywhere in the new tests sets workflow-level `thinking:` or `betas:`, so a key-mapping bug isolated to either is invisible.
3. The `requires:` union is proven only as a field value — no test drives the real capability gate with a requirement originating from an included block.
4. `GET /api/workflows` has no test that a listed workflow's `provider`/`model` survives; dropping the `{ ...ws.workflow, ...ws.declared }` spread would blank every console card silently.
5. Depth-2 composed-gate behaviour is untested (see R1 — settle the semantics first).

**Disposition:** **FIXED** across `6d7bb306` and `826fb227`. Same-provider model travel and the `thinking`/`betas` push-down are pinned in the travel suite; the API re-layering and the composed-`requires` gate tests were each verified red by sabotage. The depth-2 gate case is covered by R1's new tests, written against the corrected semantics rather than the defective ones — the recorded disagreement resolved in favour of the fix.

</details>

<details>
<summary><code>R18</code> — AI-config fields land on nodes that never read them</summary>

**Impact:** `provider`/`model`/`effort`/`thinking`/`fallbackModel`/`betas`/`sandbox` are written onto `bash:`/`script:`/`cancel:`/`workflow:` nodes; only `persist_session` is guarded by `isPersistableNode`. Inert at execution, but it puts keys on node objects whose TS type does not declare them, which the list API now serializes.

**Evidence:** `packages/workflows/src/include-expander.ts` `pushWorkflowScopeOntoNodes`. No misbehaving consumer was identified.

**Required outcome:** Optional. Either gate the push by node type, or note why the inert keys are acceptable.

**Found by:** `prp-core:code-reviewer` (flagged as a code smell, not a confirmed defect)

**Disposition:** **DECLINED.** YAGNI. The fields are inert at execution — `executeBashNode`/`executeScriptNode`/child dispatch never read them — and no misbehaving consumer was identified. The one path that could have mattered was checked: the builder round-trip reads raw YAML via `GET /api/workflows/:name`, not the list response, so these keys cannot be written back into a file. Gating the push by node type would add a check per field per node for a cosmetic concern.

</details>

<details>
<summary><code>R19</code> — Annotations and doc wording</summary>

**Impact:** Cosmetic; the repo's stated rule is complete type annotations on all functions.

**Evidence and required outcome:** Return types on the new test helpers (`dry-run.test.ts:541-551`, `include-expander.test.ts:33-35`) and on the two list callbacks (`cli/src/commands/workflow.ts:839-850`, `server/src/routes/api.ts:3193-3206`). Two documentation wording clarifications in `authoring-workflows.md` about the `$INPUTS` shell contract's scope.

**Found by:** CodeRabbit (Trivial ×2, Minor ×1)

**Disposition:** **FIXED** in `3948bbe8`. Return types added to `trace`, `composedMeta` and the CLI list callback. The two doc-wording clarifications are covered by R14's rewrite. CodeRabbit's `AGENTS.md:475` item remains NOT A FINDING for this PR.

</details>

<details>
<summary><code>R20</code> — The <code>manage_run</code> tool backgrounds a run without the gate check</summary>

**Impact:** R1's exact failure mode, on a path the R1 fix never covered. `manage_run(action=start, …)` on a project-scoped chat with a native-tools provider dispatches a workflow to the background with no composed-gate check. The run pauses at a gate nobody was warned about, in a `hidden: true` worker conversation whose pause handler explicitly emits nothing. Reachable on **every** platform with native tools — Slack, Telegram, Discord, GitHub threads — not just web, so the fix's own doc claim ("on the CLI and chat platforms the gate is presented normally") was false.

**Evidence:** `dispatchBackgroundWorkflow` has two callers — `orchestrator-agent.ts:1082` (guarded) and `orchestrator-agent.ts:1959` inside the `manage_run` `startWorkflow` callback (unguarded). The branch is entered on `conversation.codebase_id !== null && scopedCaps?.nativeTools`, with no platform guard. `orchestrator.ts:490`: `if ('paused' in result) { /* Paused workflows (approval gates) — no result card yet */ }`.

This is a **regression introduced by the R1 correction**: the load-time form applied to every dispatch surface uniformly.

**Required outcome:** enforce inside `dispatchBackgroundWorkflow`, not at its callers.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer` (independently, round 2)

**Disposition:** **FIXED** in `9a2952d8`. `assertComposedGateDriveable` moved into `dispatchBackgroundWorkflow`, before the worker conversation is created, so a refusal leaves nothing behind and every entrypoint is covered. The console path turns the throw into a user message; `manage_run` returns it to the agent as a tool error. Verified red by removing the assertion. The false doc and comment claims were corrected in the same commit.

</details>

<details>
<summary><code>R21</code> — The workflow-level provider origin named the value that lost</summary>

**Impact:** `resolveWorkflowModelScope` returned `providerOrigin: 'workflow'` whenever a `provider:` was declared, even when a tier/`@alias` `model:` resolved to a different provider — the case the executor warns about, where the preset's provider is what the run uses. `executor.ts`'s `workflow_provider_resolved` log therefore attributed the provider to the workflow definition where the pre-correction code said `model preset '<ref>'`, and a dry run of an unexpanded definition reported the same wrong origin.

**Evidence:** `node-model-resolution.ts` `providerOrigin: workflow.provider ? 'workflow' : preset ? …`, against the pre-correction `executor.ts:1329` which set `providerSource = \`model preset '${workflow.model}'\`` unconditionally inside the preset branch. `resolveNodeModel` already set `'model ref'` inside its own preset branch, so the two disagreed.

**Required outcome:** check the preset first, matching `resolveNodeModel`.

**Found by:** `prp-core:code-reviewer` (round 2)

**Disposition:** **FIXED** in `b2110b65`, with three tests pinning preset-wins / literal-model / no-declaration. Verified red by restoring the old ordering.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | R1, R4, R6, R18 |
| seams | R5, R11, R13 |
| tests | R7, R17 |
| comments | R8, R9 |
| errors | R12, R16 |
| types | R11 |
| docs | R14 |
| simplify | R5, R15 |
| validation | R3 |
| CodeRabbit (external) | R1, R2, R10, R19 |
| code (round 2) | R20, R21 |
| seams (round 2) | R20, R13 residue |

## Validation

|---|---|---|
| `bun run validate` | PASS | exit 0 at head `0eb1cecd`; zero `(fail)` lines across all package suites |
| CI `test (ubuntu-latest)` | PASS | run 32126441673 |
| CI `test (windows-latest)` | **FAIL** | PR-caused (R3). Exactly one failing test, added by this PR; `windows-latest` passed on the last five `dev` runs including base `b3f97356` |
| CI `Build docs site` | PASS | run 32126441691 |
| CI `docker-build` | PASS | run 32126441673 |
| CI `schema-upgrade` | PASS | run 32126441673 (no migration files in the diff) |
| Repo-wide collapse no-op sweep | PASS | 63 workflows / 421 nodes → 0 field divergences, 0 discovery errors |
| `bun run check:schema-upgrades` | NOT RUN | needs a live PostgreSQL; not applicable — the diff touches no migration or SQLite schema files |

## Verdict

**READY TO MERGE**

Twenty-one findings across two rounds: 20 fixed, 1 declined with recorded reasoning.
`bun run validate` passes at `b2110b65` and the sweep still reports 421 nodes → 0
divergences.

Round 2 earned its keep. The R1 correction — moving a rule from load time to invocation —
traded a false positive for a fail-open hole, and only a re-review of the *correction* would
find that. It is fixed at the structural level rather than the symptom: the assertion lives
in the one function that backgrounds a run, so a future caller cannot forget it.

Six corrections across both rounds were proven by deliberate sabotage rather than assumed:
the composed-gate refusal at dispatch, the per-run conflict de-duplication, the `declared`
spread, the `requires:` union, the `keyof` guard, and the provider-origin ordering.

One tracked follow-up: [#2606](https://github.com/coleam00/Archon/issues/2606), the
container preflight's hand-mirrored provider resolution — pre-existing, and newly cheap to
retire because of this PR's extraction.

Two things deliberately left alone, both pre-existing and outside this change: the builder
validates `when:` grammar without an upstream-ref check, and `webSearchMode` cannot travel
with a composed workflow (settled permanently in #2556, with the exception documented here
rather than worked around).

Issue #2356, opened 2026-08-01
test(workflows): build a usage-correctness harness — reconciliation, SDK cross-check, mutation, runtime assertion


Wirasm:
## Summary

Token and cost values are read from provider SDKs and never computed — verified: there is no price table and no `price × tokens` arithmetic anywhere in the tree. **The inputs are trustworthy. Every defect has been in aggregation.**

Four in roughly one day, none caught by unit tests, type-checking, lint, or CI:

| defect | shape |
| --- | --- |
| `loop_group` persisted `tokens` on both leaf and aggregate rows | naive sum = **2×** |
| the fix relocated the same double-count to the gate/finalize path | naive sum = **2×** again |
| resumed runs overwrote rather than accumulated the run total | **~2× undercount**, measured on 3 real runs |
| cost accumulated with `+=` when the SDK reports it session-cumulatively (#2083) | double-charge |

All four were found by a human summing persisted rows by hand and comparing to truth.

**Why the existing tests missed them:** token assertions live in five test files and every one asserts a **single site in isolation**. All four defects were *across* sites, or across executions. There is currently **no reconciliation test anywhere in the repo** — `rg -l "naiveSum|total_tokens" --glob '*.test.ts'` finds five files, none of which compares parts to whole.

This issue proposes four test types, ranked by how many of the four known defects each would have caught. That ranking is the point — a proposed test regime should be judged against bugs that actually happened, not against intuition.

---

## 1. Naive-consumer property over a shape × path matrix — catches 3 of 4

The single assertion that does most of the work:

```
Σ(data.tokens across every persisted node_completed row) == true usage
```

Parametrised over node shape × completion path:

| | natural completion | gate + bare approve | resume after failure |
| --- | --- | --- | --- |
| plain AI node | | | |
| `loop` | | | |
| `loop_group` | ✗ was 2× (#2347) | ✗ was 2× (#2347) | **untested — #2353 introduces it** |
| `workflow:` sub-run | | | |

**Every defect so far was one cell of this matrix being wrong**, and each was found one cell at a time. At least two cells remain unverified today.

**Load-bearing harness requirement:** the pause/resume cells must use **one shared event store** across the pause and the resume, as a real database behaves. Per-test isolation hides this class entirely — that is specifically why the gate-path double-count shipped and was only caught when a reviewer wired the phases to a single store.

## 2. Cross-check against the SDK's own session totals — catches the one (1) cannot

[...]

PR #2608, opened 2026-08-18
fix(workflows): a run records what it spent even when it fails or is cancelled

### LINKED ISSUE #2469: fix(workflows): a run that fails after burning tokens records no spend — fan-out makes the undercount routine
A workflow run's cost and token usage are persisted in exactly one place — inside `completeWorkflowRun` (`packages/workflows/src/dag-executor.ts`). `failWorkflowRun` writes no usage metadata.

So a run that burns tokens and then fails or is cancelled records **no spend at all**. Downstream, `childOutcomeFromRun` returns `undefined` for it, and every aggregate is Σ of *completed* children rather than Σ of children.

## Why this stopped being an edge case

It is genuinely pre-existing — inherited from the 1:1 `workflow:` path, where a failed child usually meant a failed run and the number stopped mattering.

**Fan-out changed that.** `all_done` is the default join and a failing child explicitly must not discard its siblings (the independence rule, `workflow-language-constitution.md`). Partial failure is now a *designed, routine outcome* — so a ten-item fan-out where three children burn tokens and fail reports the spend of seven, on the happy path, silently.

There is no warning. The number is simply lower than reality, which is the worst shape for a cost figure: plausible, and wrong in the direction that does not prompt investigation.

## Compounding: tokens are the axis that survives resume

`getDagResumeSnapshot` (`packages/core/src/db/workflow-events.ts`) rebuilds cumulative usage across resume passes by summing `data.tokens` — **it never reads `cost_usd`**. That feeds `priorTokenUsage` → `runCtx.totalTokensIn/Out`. Codex reports no cost at all, so on a Codex fan-out with failures the loss is total on both axes.

#2224 fixed the *fan-out event* half of this (tokens are now persisted on the fan-out `node_completed` event, previously dropped entirely) and corrected the docblocks to state the completed-only caveat. What remains is the upstream gap.

## Fix direction

Persist usage in `failWorkflowRun` the same way `completeWorkflowRun` does, so a child's spend survives its outcome. Then the Σ-completed caveat can come back out of the docblocks in `sumFanOutCost` / the token roll-up, and out of #2224's PR body.

Worth checking whether cancelled runs should be treated the same — a cancelled child that already burned tokens has the same problem, and the fan-out cancels children itself (`fan_out_gate`, `fan_out_sibling`, `fan_out_orphan`).

## Acceptance

- A run that fails after consuming tokens persists its usage
- A fan-out with N children where K fail reports the spend of all N
- The completed-only caveats in the fan-out docblocks are removed rather than reworded

Found during the #2224 review. Related: #2224 (merged, `cbeab968`).

Wirasm:
## Problem and outcome

A workflow run's cost and tokens were persisted in exactly one place — the metadata argument of `completeWorkflowRun`. Any other ending recorded no spend at all, and because `childOutcomeFromRun` reads usage straight off the child's row, a fan-out aggregate silently became Σ of *completed* children. `all_done` is the default join and a failing child explicitly must not discard its siblings, so partial failure is a designed, routine outcome: a ten-item fan-out where three children burn tokens and fail reported the spend of seven, on the happy path, with no warning. That is the worst shape for a cost figure — plausible, and wrong in the direction that does not prompt investigation.

- **Outcome:** A run records what it spent whatever its outcome. A failed run's Slack completion message and console cards (the default UI) now carry a cost where they were blank; a partly-failed fan-out reports every child.
- **Invariant:** A workflow run persists the usage it accumulated independently of how it ended — completed, failed, cancelled, or paused — so sub-run and fan-out aggregates sum every child.
- **Scope boundary:** `node_counts` is still not stored for failed runs (different concern, computed after this point). Node-level provider stats lost across an interactive loop gate stay with #2345. No run-tree budget cap (#1961). No schema change and no backfill of historical runs.
- **Root cause:** Usage is accumulated in `runCtx` at the single aggregation point `dag-executor.ts:7893-7905`, read into locals at `:8685-8688`, and serialized **only** into `completeWorkflowRun`'s metadata at `:9060`. The three `failWorkflowRun` call sites pass a message only, and `failWorkflowRun` writes `{ error }` (`core/src/db/workflows.ts:971`). The `skipIfStatusChanged` bail-out for an externally cancelled run returns before either writer. Compounding: `runCtx.totalCostUsd` was seeded to `0` every pass while tokens were seeded from prior events, because `getDagResumeSnapshot` summed `data.tokens` and never `data.cost_usd`.

## Review guidance

- **Feedback requested:** Correctness of the placement — is the run tail genuinely upstream of every disposition, and is the `finally` the right coverage for a throw out of `runLayers`?
- **Start here:** `packages/workflows/src/dag-executor.ts:8683-8737` — `persistRunUsage()` and the `try/finally` around `runLayers()`. Everything else follows from moving the write here.
- **Review order:** `dag-executor.ts` (the write, then the `completeWorkflowRun` call losing its usage keys at `:9060`) → `core/src/db/workflow-events.ts` (cost summing) → `executor.ts` + `store.ts` (the `priorUsage` bundle) → tests.
- **Lower-attention areas:** the `priorTokenUsage` → `priorUsage` rename is mechanical (6 non-test sites, type-check covers it); the four `getDagResumeSnapshot` test stubs gaining `costUsd: 0` are contract mirrors.
- **Known risk or uncertainty:** cost seeding inherits token seeding's semantics exactly, including the pre-existing over-count when an `always_run` node re-executes on resume while its earlier `node_completed` event is still summed. Deliberate — diverging the two axes would be worse than the shared, pre-existing wart.

Two residuals worth knowing before reviewing, both verified and neither introduced here:

- **A node that fails mid-stream still loses its spend from the event log.** `node_failed` writes `data: { error }` only (`dag-executor.ts:2569-2573`) while the result object carries `costUsd`/`tokens` into `ctx` (`:2582-2600`). So the failed *run's row* is now correct, but a later resume can only rebuild what `node_completed` rows hold. The figure never decreases and is never double-counted; it just cannot see a failed attempt's spend. Filed as #2609 — it needs a decision (should `total_cost_usd` mean money burned, or cost of the surviving path?) that would widen this PR.
- **The legacy dashboard card still will not show a failed run's cost.** `WorkflowRunCard.tsx:209-219` nests the cost span inside `isValidNodeCounts(run.metadata?.node_counts)`, and `node_counts` is deliberately still not written on failure. The console (the default UI) has no such gate — `primitives/run.ts:117`, `RunDetailHeader.tsx:117`, `ActiveRunCard.tsx:119`, `RecentRunRow.tsx:133` all key off cost alone — and neither does Slack. Not fixed here because the legacy dashboard is slated for removal and un-gating it would drag `node_counts`-on-failure into scope.

## Solution

The decisive observation is that all dispositions converge on one line. `runLayers(runCtx)` returns exactly once, and everything after it — the container-pause return, the `skipIfStatusChanged` bail-outs for external cancel/pause, both failure branches, the evidence gate, the write-back gate, completion — is a *disposition*, not an accumulation. So the fix is not "teach `failWorkflowRun` to carry usage too"; that satisfies half the invariant (`cancelWorkflowRun` runs in another process with no usage in scope, so a cancelled or gate-cancelled child would still record nothing) and duplicates the payload across three call sites with nothing preventing a fourth from forgetting it. Instead the write moves upstream of the branch.

It goes through `updateWorkflowRun` rather than the terminal writers for a concrete reason: that merge carries no `WHERE status = 'running'` guard, so it still lands on a row another process has already flipped to `cancelled`. `completeWorkflowRun` and `failWorkflowRun` would both refuse it. The evidence gate at `dag-executor.ts:8905` is the existing precedent for the shape — metadata write first, terminal status after. `completeWorkflowRun` now carries `node_counts` and the sub-run `summary` only, so there is one usage writer instead of one per outcome.

The `finally` is load-bearing rather than defensive, and the second commit proves it. `runLayers` guards almost everything — every store call inside the per-node try, the between-layer status check, the artifact writes — but `safeSendMessage` deliberately **rethrows** a FATAL-classified platform error rather than swallowing it (`executor-shared.ts:830-832`; `'unauthorized'`/`'forbidden'`/`'401'` are in `FATAL_PATTERNS`), and the `allSettled` rejected branch calls it outside any try. So a platform whose auth dies mid-run throws clean out of `runLayers`, past every disposition, to `executeWorkflow`'s catch-all — which marks the run **failed**. `persistRunUsage` swallows and logs its own errors so it can never displace the in-flight exception.

Cost also had to become cumulative across resume. Without it, the newly-persisted figure would *decrease*: fail at $0.50, resume, complete having spent $0.20, row reads $0.20. `getDagResumeSnapshot` now sums `data.cost_usd` from `node_completed` rows alongside `data.tokens` and seeds `runCtx.totalCostUsd`, which also closes the pre-existing under-count on every resumed run (total on Codex, which reports no cost either way). `node_skipped_prior_success` rows stay excluded from both axes, so a multi-pass resume cannot multiply a node's usage.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `metadata.total_cost_usd` / `total_tokens_*` written only on clean completion; a fan-out reported Σ of completed children | Written at the run tail on every disposition; a fan-out reports Σ of all children |
| **Failure behavior** | A failed or cancelled run recorded no spend, silently | Its spend is recorded; a failed metadata write logs `dag.run_usage_persist_failed` and never fails the run |
| **Resumed runs** | Tokens cumulative, cost reset to the current pass | Both cumulative; the persisted figure never decreases across a resume |

## Architecture

```mermaid
flowchart TD
  RL["runLayers() returns — accumulators settled"] ==> W["persistRunUsage()<br/>updateWorkflowRun(metadata) in a finally"]
  W --> Y["metadata.total_cost_usd<br/>total_tokens_in / _out"]
  W --> D{"disposition"}
  D --> P["container pause"]
  D --> S["external cancel / pause"]
  D --> F["failWorkflowRun (3 sites)"]
  D --> C["completeWorkflowRun<br/>node_counts + summary only"]
```

### Changed seams

|---|---|---|
| `executeDagWorkflow → IWorkflowStore` | Usage now written via `updateWorkflowRun` at the run tail; removed from `completeWorkflowRun` | `dag-executor.ts:8683-8737`, `:9060`; `dag-executor.test.ts` "run usage survives every disposition" |
| `getDagResumeSnapshot → executeDagWorkflow` | `DagResumeSnapshot` gains `costUsd`; `priorTokenUsage` becomes a bundled `priorUsage: PriorRunUsage` | `core/src/db/workflow-events.ts:239-292`, `store.ts:17-25`, `executor.ts:603-610` |
| `child run row → childOutcomeFromRun → sumFanOutCost` | Unchanged code; now fed by failed/cancelled children too | `subrun.test.ts` "rolls up the spend of EVERY child…" |

## Review corrections

An independent `/prp-review` returned NEEDS FIXES with three blocking findings. All three were the same omission on different surfaces — the first pass fixed the cost axis where it looked and did not sweep the sites where cost and tokens are deliberately asymmetric.

- **R1 — the 1:1 `workflow:` path dropped a failed child's usage.** `failResult` (`dag-executor.ts:5839`) took only `(error)` while its fan-out sibling (`:6367`) has taken `(error, costUsd?, tokens?)` since #2224, so a parent with one ordinary `workflow:` node still reported zero for a child that burned tokens and failed — the more common topology of the two. Widened to match and passed from the two `interpret()` branches that hold a real outcome. Verified red: reverting just that call site makes the new `subrun.test.ts` case report `undefined`.
- **R2 — the new resume cost sum double-counted a completed `loop_group`.** Its body nodes write their own `node_completed` rows carrying `cost_usd`, and the group writes a roll-up row restating the sum. `tokens` was deliberately omitted from that row for exactly this reason; `cost_usd` was left because nothing summed it — and this PR added the first consumer that does. The roll-up row is now marked `aggregate: true` and `getDagResumeSnapshot` skips marked rows on both axes; the row keeps `cost_usd` so the console's loop_group card (which reads it per event) does not go blank. Covered from both ends — producer marks, consumer skips — since either half silently reintroduces it. **Known limit:** rows written before the marker existed carry no flag, so a run that completed a loop_group under an older build and is resumed under this one can still double-count cost. Bounded and self-clearing.
- **R4 — the `signaledTokens` docblock overclaimed.** It said the run row was no longer affected across a gate. A pause writes no `node_completed`, `priorUsage` is rebuilt from those rows alone, and the merge replaces rather than adds — so a twice-gated loop still reports only the final invocation. Narrowed to that, pointing at the same #2345 boundary as the node row.
- **R5 (suggestion) — fixed.** `persistRunUsage` was documented best-effort and never watched failing; a test now rejects the usage write and asserts the run still completes and `dag.run_usage_persist_failed` is logged.
- **R3 (suggestion) — declined.** It asks for the user-facing warning to cover a failed usage write. That warning says the *run status* could not be saved and the result may appear inconsistent — still accurate for its own failure. A usage-write failure leaves the cost key absent, and every reader type-guards absence and renders blank. Absent is not the failure mode this PR is about; plausible-and-wrong is. The error log carries the run id under a distinct event name, which is the right channel for a bookkeeping write.

## Validation

- `bun run validate` — exit 0 — proves types, `eslint --max-warnings 0`, format, generated-file checks and every package suite.
- `bun test src/dag-executor.test.ts` (packages/workflows) — 502 pass, 0 fail — proves usage is recorded on failure, on out-of-band cancellation, at an approval pause, and when a FATAL platform error throws out of `runLayers`; that a zero-usage run still writes nothing; and that `completeWorkflowRun` no longer carries usage keys.
- Mutation check on the `finally` (CodeRabbit's point, addressed in `47146744`): replacing it with a plain block makes the throw-path case report an empty usage-write list instead of `{0.01, 20, 2}` — the guard is observed firing rather than assumed. Note CodeRabbit's own sketch would not have worked: it proposed rejecting `getWorkflowRunStatus`, which `runLayers` wraps in try/catch at `:7980-8009`.
- `bun test src/subrun.test.ts` (packages/workflows) — 67 pass, 0 fail — the regression proof: three fan-out children, one fails *after* its AI node spent tokens, and the parent rolls up 0.03 (all three) rather than 0.02.
- `bun test src/db/workflow-events.test.ts` (packages/core) — 24 pass, 0 fail — cost summing, `node_skipped_prior_success` exclusion, malformed/non-finite handling, and no warn when cost is simply absent.
- Regression aim confirmed: with the production files stashed, the fan-out case fails at `subrun.test.ts:2694` because the failed child's `metadata.total_cost_usd` is `undefined` — it is aimed at the defect, not at the harness.
- **Not verified:** no live end-to-end run against a real repo with real provider spend. The fan-out case exercises the real recursive `executeWorkflow` against an in-memory store with a canned provider, which covers the persistence path; only the provider's own cost reporting is stubbed, and that is unchanged by this PR.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No schema change and no key rename. Historical failed runs are **not** backfilled — they keep their absent usage. All four readers already type-guard the keys, so absence stays safe | `web/.../console/primitives/run.ts:117`, `WorkflowRunCard.tsx:213`, `slack/workflow-bridge.ts:301`, `dag-executor.ts:513` |
| Rollout / rollback | Single revertible commit, no flag, no data to unwind. Reverting restores the previous under-report | — |
| Observability | A failed usage write logs `dag.run_usage_persist_failed` with the run id rather than going quiet — the defect being fixed was a number silently going missing | `dag-executor.ts:8725` |
| Documentation / communication | The "Cost roll-up" bullet already promised an unqualified roll-up (the docs were right, the code was wrong); it now states outcome-independence explicitly | `packages/docs-web/src/content/docs/guides/authoring-workflows.md:1433` |

## Links

- Closes #2469
- Related #2224 #2345 #1961 #2356
- Follow-up filed: #2609

## Summary by CodeRabbit

* **New Features**
  * Workflow runs now track cumulative USD costs alongside token usage.
  * Costs persist across completed, failed, cancelled, paused, and resumed runs.
  * Parent workflows include spending from all child workflows, including failed or cancelled children.
  * Resumed workflows restore both prior token usage and cost totals.
  * Fan-out and loop executions aggregate usage across all child work.

* **Bug Fixes**
  * Invalid or missing cost values no longer distort usage totals.
  * Duplicate, replayed, and aggregate usage is excluded from calculations.

### DIFF
--- a/packages/core/src/db/workflow-events.test.ts
+++ b/packages/core/src/db/workflow-events.test.ts
@@ -495,6 +495,37 @@ describe('workflow-events', () => {
       expect(mockLogger.warn).not.toHaveBeenCalled();
     });
 
+    // #2469: a loop_group writes a roll-up node_completed row whose cost_usd restates
+    // what its own `<groupId>.<nodeId>` body rows already carry. Summing both counted
+    // that group twice — the same silently-wrong number this work exists to remove,
+    // pointing the other way. The roll-up is marked `aggregate: true` and skipped here.
+    test('excludes usage on aggregate rows while keeping their output', async () => {
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'group.body',
+            event_type: 'node_completed',
+            data: { node_output: 'iteration 1', cost_usd: 0.01, tokens: { input: 10, output: 1 } },
+          },
+          {
+            step_name: 'group',
+            event_type: 'node_completed',
+            data: { node_output: 'last iteration', cost_usd: 0.01, aggregate: true },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-loop-group');
+
+      // The leaves are authoritative: 0.01 total, not 0.02.
+      expect(result.costUsd).toBeCloseTo(0.01, 10);
+      expect(result.tokens).toEqual({ input: 10, output: 1 });
+      // The roll-up still marks the group node completed, so resume skips it rather
+      // than re-running the whole group.
+      expect(result.completedNodeOutputs.get('group')).toBe('last iteration');
+      expect(result.completedNodeOutputs.get('group.body')).toBe('iteration 1');
+    });
+
     test('returns an empty snapshot when no events exist', async () => {
       mockQuery.mockResolvedValueOnce(createQueryResult([]));
 
--- a/packages/core/src/db/workflow-events.test.ts
+++ b/packages/core/src/db/workflow-events.test.ts
@@ -401,6 +401,100 @@ describe('workflow-events', () => {
       expect(result.tokens).toEqual({ input: 10, output: 3 });
     });
 
+    // #2469: cost is restored across resume passes exactly like tokens. Before this,
+    // only tokens were summed, so a resumed run's cost silently reset to the current
+    // pass — and once a FAILED run started persisting cost, the figure would have gone
+    // down after a successful resume.
+    test('sums cost_usd from node_completed events', async () => {
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'node-a',
+            event_type: 'node_completed',
+            data: { node_output: 'output A', cost_usd: 0.02, tokens: { input: 40, output: 4 } },
+          },
+          {
+            step_name: 'node-b',
+            event_type: 'node_completed',
+            data: { node_output: 'output B', cost_usd: 0.03 },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-cost');
+
+      expect(result.costUsd).toBeCloseTo(0.05, 10);
+      expect(result.tokens).toEqual({ input: 40, output: 4 });
+    });
+
+    test('excludes cost_usd on node_skipped_prior_success rows (multi-resume)', async () => {
+      // A prior-success replay must not re-charge a node an earlier pass already
+      // counted — otherwise every resume multiplies that node's cost.
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'node-a',
+            event_type: 'node_completed',
+            data: { node_output: 'output A', cost_usd: 0.02 },
+          },
+          {
+            step_name: 'node-b',
+            event_type: 'node_skipped_prior_success',
+            data: { reason: 'prior_success', node_output: 'output B', cost_usd: 999 },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-cost-resume');
+
+      expect(result.costUsd).toBeCloseTo(0.02, 10);
+    });
+
+    test('ignores malformed and non-finite cost_usd while keeping valid rows', async () => {
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'node-a',
+            event_type: 'node_completed',
+            data: { node_output: 'valid', cost_usd: 0.01 },
+          },
+          {
+            step_name: 'node-b',
+            event_type: 'node_completed',
+            data: { node_output: 'string cost', cost_usd: '0.5' },
+          },
+          {
+            step_name: 'node-c',
+            event_type: 'node_completed',
+            data: { node_output: 'nan cost', cost_usd: Number.NaN },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-cost-malformed');
+
+      expect(result.costUsd).toBeCloseTo(0.01, 10);
+      expect(result.completedNodeOutputs.size).toBe(3);
+      expect(mockLogger.warn).toHaveBeenCalledTimes(2);
+    });
+
+    test('does not warn when completed events omit optional cost', async () => {
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'node-a',
+            event_type: 'node_completed',
+            data: { node_output: 'no cost reported', tokens: { input: 5, output: 1 } },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-cost-absent');
+
+      expect(result.costUsd).toBe(0);
+      expect(mockLogger.warn).not.toHaveBeenCalled();
+    });
+
     test('returns an empty snapshot when no events exist', async () => {
       mockQuery.mockResolvedValueOnce(createQueryResult([]));
 
@@ -408,6 +502,7 @@ describe('workflow-events', () => {
 
       expect(result.completedNodeOutputs.size).toBe(0);
       expect(result.tokens).toEqual({ input: 0, output: 0 });
+      expect(result.costUsd).toBe(0);
     });
 
     test('throws on DB query error', async () => {
--- a/packages/core/src/db/workflow-events.ts
+++ b/packages/core/src/db/workflow-events.ts
@@ -218,9 +218,19 @@ export async function listWorkflowEventsSince(
  * run. Used by the DAG executor to restore state when resuming a failed run.
  * Throws on DB error — caller owns the degradation policy.
  *
- * Both usage axes are summed from `node_completed` rows only. `node_skipped_prior_success`
- * rows are deliberately excluded: they replay a node an earlier pass already counted, so
- * counting them would multiply that node's usage by the number of resume passes.
+ * Both usage axes are summed from `node_completed` rows only, and only from rows that are
+ * not marked `data.aggregate`. Two distinct duplication hazards:
+ *
+ * - `node_skipped_prior_success` rows replay a node an earlier pass already counted, so
+ *   counting them would multiply that node's usage by the number of resume passes.
+ * - `aggregate: true` rows are derived from other rows already in this log — a
+ *   `loop_group`'s roll-up restates the `cost_usd` its own `<groupId>.<nodeId>` body rows
+ *   carry, so summing both counts that group twice (#2469).
+ *
+ * Rows written before the `aggregate` marker existed carry no flag, so a run that
+ * completed a loop_group under an older build and is resumed under this one can still
+ * double-count its cost. Bounded and self-clearing: only cost is affected (the roll-up
+ * never carried `tokens`), and only until those runs reach a terminal state.
  */
 export async function getDagResumeSnapshot(workflowRunId: string): Promise<{
   completedNodeOutputs: Map<string, string>;
@@ -255,6 +265,8 @@ export async function getDagResumeSnapshot(workflowRunId: string): Promise<{
     if (typeof data.node_output === 'string') {
       completedNodeOutputs.set(row.step_name, data.node_output);
     }
+    // A derived row restates usage that other rows in this same log already carry.
+    if (data.aggregate === true) continue;
     if (row.event_type === 'node_completed' && data.tokens !== undefined) {
       const eventTokens = data.tokens;
       if (
--- a/packages/core/src/db/workflow-events.ts
+++ b/packages/core/src/db/workflow-events.ts
@@ -214,13 +214,18 @@ export async function listWorkflowEventsSince(
 }
 
 /**
- * Return completed node outputs and cumulative token usage for a workflow run.
- * Used by the DAG executor to restore state when resuming a failed run.
+ * Return completed node outputs and cumulative usage (tokens AND cost) for a workflow
+ * run. Used by the DAG executor to restore state when resuming a failed run.
  * Throws on DB error — caller owns the degradation policy.
+ *
+ * Both usage axes are summed from `node_completed` rows only. `node_skipped_prior_success`
+ * rows are deliberately excluded: they replay a node an earlier pass already counted, so
+ * counting them would multiply that node's usage by the number of resume passes.
  */
 export async function getDagResumeSnapshot(workflowRunId: string): Promise<{
   completedNodeOutputs: Map<string, string>;
   tokens: { input: number; output: number };
+  costUsd: number;
 }> {
   const result = await pool.query<{
     step_name: string | null;
@@ -234,6 +239,7 @@ export async function getDagResumeSnapshot(workflowRunId: string): Promise<{
   );
   const completedNodeOutputs = new Map<string, string>();
   const tokens = { input: 0, output: 0 };
+  let costUsd = 0;
   for (const row of result.rows) {
     if (!row.step_name) continue;
     let data: Record<string, unknown>;
@@ -270,6 +276,20 @@ export async function getDagResumeSnapshot(workflowRunId: string): Promise<{
         );
       }
     }
+    if (row.event_type === 'node_completed' && data.cost_usd !== undefined) {
+      const eventCost = data.cost_usd;
+      // Same guard shape as tokens: a non-finite value from a provider must not
+      // silently poison the total (NaN > 0 is false, which would drop the run's
+      // cost from the persisted metadata with no trace).
+      if (typeof eventCost === 'number' && Number.isFinite(eventCost)) {
+        costUsd += eventCost;
+      } else {
+        getLog().warn(
+          { runId: workflowRunId, stepName: row.step_name, costUsd: eventCost },
+          'db.workflow_dag_node_cost_invalid_ignored'
+        );
+      }
+    }
   }
-  return { completedNodeOutputs, tokens };
+  return { completedNodeOutputs, tokens, costUsd };
 }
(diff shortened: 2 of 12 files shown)


Wirasm:
---
pr: 2608
base: dev
head: fix/2469-run-usage-survives-outcome
reviewed: 2026-08-19T09:57:55Z
reviewed_head: a4d7d61377a61a62f0ce6d0aae70d3754e3e4c5b
verdict: READY TO MERGE
open_findings: 0
scopes: [code, seams, tests, comments, errors, types, docs]
publication: https://github.com/coleam00/Archon/pull/2608#issuecomment-5329801308
---



# PR Review: #2608 — fix(workflows): a run records what it spent even when it fails or is cancelled

## Signal

The run-tail design is sound and the corrected PR now carries its usage invariant through every affected seam. Usage is persisted from one `finally` after `runLayers`, independent of completed, failed, cancelled, paused, or thrown outcomes. Failed/cancelled 1:1 child runs retain their usage for parent aggregation; resume hydration excludes marked loop-group aggregate rows; the repeated-gate caveat is stated accurately; and non-finite provider costs are rejected at both node and loop accumulation boundaries while preserving the last finite value. The final regression also proves finite-then-`NaN` behavior within one multi-result loop iteration.

All three original Important findings are fixed. The best-effort persistence suggestion is now covered by a rejecting-store test; the user-facing warning suggestion remains deliberately declined because the failure is still emitted to the operator log with the error and run ID. Independent final code, seam, test, and comment verification found no new issue. All required GitHub checks are green at `a4d7d613`.

- **Blocking:** 0
- **Non-blocking:** 0
- **Resolved since previous review:** `R1`, `R2`, `R4`, `R5`, `CR1`, `CR2`, `CR4`, `CR5`, `V1`, `V2`
- **Declined with rationale:** `R3`, `CR3`
- **Tracked follow-ups:** #2609
- **Validation:** local `bun run validate` PASS; GitHub docs, Ubuntu, Windows, schema-upgrade, Docker, and CodeRabbit PASS

## Findings

|---|---|---|---|
| `R1` | Important | A failed/cancelled 1:1 `workflow:` child dropped usage before parent aggregation | FIXED |
| `R2` | Important | Resume hydration double-counted a completed `loop_group` roll-up row | FIXED |
| `R4` | Important | The `signaledTokens` docblock overstated repeated-gate run-row accuracy | FIXED |
| `R3` | Suggestion | A failed usage write no longer reaches the completion path's user-facing warning | DECLINED |
| `R5` | Suggestion | The best-effort persistence failure path lacked direct protection | FIXED |
| `CR1` | Suggestion | Initial tests did not prove the usage write lived in `finally` | FIXED |
| `CR2` | Suggestion | The throw-path test did not prove usage was written after the throw | FIXED |
| `CR3` | Suggestion | Add explicit return annotations only to the new test mocks | DECLINED |
| `CR4` | Important | Run-level accumulation accepted non-finite provider cost | FIXED |
| `CR5` | Important | Reask and loop-local accumulation could erase a prior finite cost with `NaN` | FIXED |
| `V1` | Suggestion | Loop coverage did not exercise finite then `NaN` within one iteration | FIXED |
| `V2` | Suggestion | A 1:1 child comment inaccurately characterized usage-less failures | FIXED |

## Detailed Findings

<details>
<summary><code>R1</code> — failed/cancelled 1:1 child usage</summary>

**Disposition:** FIXED. The 1:1 `failResult` accepts optional cost/tokens and settled failed/cancelled outcomes pass `ChildWorkflowOutcome` usage through, matching fan-out. `subrun.test.ts` proves the failed child and failed parent both record the expected usage.

</details>

<details>
<summary><code>R2</code> — loop-group resume double count</summary>

**Disposition:** FIXED. Loop-group roll-up events carry `aggregate: true`; `getDagResumeSnapshot` retains their output but excludes their derived cost/tokens from the resume baseline. Producer and consumer regressions cover the seam.

</details>

<details>
<summary><code>R4</code> — repeated-gate attribution prose</summary>

**Disposition:** FIXED. The docblock now states that node and run rows share per-invocation attribution and explains why repeated gates remain within #2345's fix boundary.

</details>

<details>
<summary><code>R3</code> — user-facing warning on usage persistence failure</summary>

**Disposition:** DECLINED. Usage persistence is intentionally best-effort. A failure remains observable through `dag.run_usage_persist_failed`, including the error and workflow run ID, while the run's primary outcome is preserved. This was non-blocking and no false-success value is written.

</details>

<details>
<summary><code>R5</code> — best-effort persistence regression coverage</summary>

**Disposition:** FIXED. A rejecting `updateWorkflowRun` case proves usage bookkeeping failure does not fail the workflow and that the structured error is logged.

</details>

<details>
<summary><code>CR1</code> / <code>CR2</code> — prove the throw-path ordering</summary>

**Disposition:** FIXED. A fatal platform send drives an actual escape from `runLayers`; the test records ordering and proves the usage-bearing update occurs after the fatal send, so a plain pre-return write cannot satisfy it.

</details>

<details>
<summary><code>CR3</code> — annotations on selected test callbacks</summary>

**Disposition:** DECLINED. The callbacks follow the test file's established inferred-return style. Selectively annotating only the new mocks would not strengthen a production boundary and would reduce local consistency.

</details>

<details>
<summary><code>CR4</code> / <code>CR5</code> — non-finite provider costs</summary>

**Disposition:** FIXED. Run-level accumulation accepts only finite values. Node reask and loop provider-result boundaries likewise ignore non-finite contributions, preserve the last valid cumulative cost, and emit path-specific warnings. Tests cover a finite reask followed by `NaN`, a finite loop sequence followed by `NaN`, and finite then `NaN` result chunks within the same background-task loop iteration.

</details>

<details>
<summary><code>V1</code> — same-iteration loop coverage</summary>

**Disposition:** FIXED. The final regression emits `0.1` and then `NaN` from one loop iteration, asserts `node_completed.cost_usd === 0.1`, and observes the warning. A regression that clears the saved iteration cost fails the assertion.

</details>

<details>
<summary><code>V2</code> — 1:1 child usage comment</summary>

**Disposition:** FIXED. The comment now states the actual invariant: settled child outcomes supply usage, while paths without an outcome omit it. Final comment analysis checked every reachable caller and found no mismatch.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | Final delta reviewed; no findings |
| seams | Same-iteration provider-result seam and prior corrected seams verified; no findings |
| tests | New regression is non-vacuous and fails the plausible cost-clearing regression; no findings |
| comments | Corrected 1:1 child comment matches all reachable callers; no findings |
| errors | Prior `R3` remains an explicit, observable best-effort policy; no blocking finding |
| types | Prior 1:1 usage shape correction verified; no open finding |
| docs | User-facing cost-roll-up documentation remains accurate; no open finding |

## Validation

| Command or check | Result | Evidence |
|---|---|---|
| `bun run validate` | PASS | Generated checks, all package/script type checks, zero-warning lint, formatting, installer tests, package suites, and script tests exited 0 |
| `bun --cwd=packages/workflows test src/dag-executor.test.ts` | PASS | 506 passed, 0 failed, including the same-iteration finite-then-`NaN` regression |
| GitHub `Build docs site` | PASS | Completed successfully at `a4d7d613` |
| GitHub `test (ubuntu-latest)` | PASS | Completed successfully at `a4d7d613` |
| GitHub `test (windows-latest)` | PASS | Completed successfully at `a4d7d613` |
| GitHub `schema-upgrade` | PASS | Completed successfully at `a4d7d613` |
| GitHub `docker-build` | PASS | Completed successfully at `a4d7d613` |
| GitHub `CodeRabbit` | PASS | Review completed successfully at `a4d7d613` |

## Verdict

**READY TO MERGE**

No open Important/Critical finding remains, every correction is directly protected, the branch is mergeable, and all required checks are green at the reviewed head.

PR #2625, opened 2026-08-19
fix(cli): reject run IDs passed to workflow status

### LINKED ISSUE #2612: fix(cli): workflow status silently ignores a run-id argument
## Problem

`archon workflow status <run-id>` looks like it should show that run's status — but `workflowStatusCommand` takes no run-id at all (`packages/cli/src/cli.ts:617-622`: only the json/verbose/events flags), so the positional is silently dropped and the command shows its usual overview instead. Observed 18 Aug: `workflow status c5c5bdbf` while run `c5c5bdbf…` was live printed a list of paused isolation environments — nothing about the run — with no hint the argument was ignored.

A silently ignored argument violates the fail-fast rule: the operator believes they asked a narrower question and reads the broad answer as if it were the narrow one.

## Proposed solution

Either honor the argument (delegate to the `get` path for that run) or reject it with a usage error pointing at `workflow get <id>`. Honoring it is friendlier; rejecting it is a two-line fix. Either beats silence.

## Acceptance

- `workflow status <id>` never silently ignores the id: it shows that run or errors naming the right command.
- `workflow status` with no argument is unchanged.

Wirasm:
## Problem and outcome

`archon workflow status <run-id>` accepted and silently ignored the run ID, then displayed the general workflow overview. This could lead an operator to mistake a broad status view for the requested run.

- **Outcome:** Passing a run ID to `workflow status` now exits with usage guidance that directs the operator to `workflow get <run-id>`.
- **Invariant:** `workflow status` without a positional argument retains its existing overview behavior.
- **Scope boundary:** This PR does not change the `workflow get` command or add run-specific behavior to `workflow status`.
- **Root cause:** The `status` dispatch accepted no positional argument but did not validate `positionals[2]`, so parseArgs left the extra value unused.

## Review guidance

- **Feedback requested:** Confirm that rejecting the unsupported positional with a pointer to the existing `workflow get` command is the appropriate CLI contract.
- **Start here:** `packages/cli/src/cli.ts:618` — rejects the otherwise ignored positional before the overview command runs.
- **Review order:** Review the dispatch guard, then the subprocess regression test in `packages/cli/src/cli.test.ts`.
- **Lower-attention areas:** None.

## Solution

The workflow command dispatcher now checks for a positional after `status`. When present, it emits the valid `workflow status` usage and names `workflow get <run-id>` as the command for inspecting a single run, then exits non-zero. This is the smallest accepted fix because the CLI already has a dedicated single-run command.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `workflow status <run-id>` displayed the general status overview. | It prints actionable usage guidance and exits with code 1. |
| **Failure behavior** | The unsupported argument was silently ignored. | The invalid invocation is explicit and directs the operator to `workflow get <run-id>`. |

## Validation

- `bun test src/cli.test.ts` (from `packages/cli`) — passed, 55 tests — covers the rejection path and existing CLI parsing behavior.
- `bun run validate` (clean detached worktree at `16e0e97e`) — passed — covers repository type checks, lint, formatting, bundled checks, and tests without unrelated untracked workflow files.
- **Not verified:** Nothing material; the regression test exercises the process-level CLI invocation and confirms the status overview does not run.

## Links

- Closes #2612

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for `workflow status` and `workflow get` commands.
  - Clear error messages now appear when unexpected or missing arguments are provided.
  - Users are directed to `workflow get` when checking an individual workflow run.

- **Tests**
  - Added coverage for invalid command argument combinations, including exit statuses, guidance, and output behavior.

### DIFF
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -36,6 +36,23 @@ describe('CLI help output', () => {
   });
 });
 
+describe('workflow status arguments', () => {
+  it('rejects a run id and points to workflow get', () => {
+    const result = spawnSync(
+      process.execPath,
+      [join(import.meta.dir, 'cli.ts'), 'workflow', 'status', 'abc123'],
+      { encoding: 'utf8' }
+    );
+
+    expect(result.status).toBe(1);
+    expect(result.stderr).toContain(
+      'Usage: archon workflow status [--json] [--verbose] [--events]'
+    );
+    expect(result.stderr).toContain('archon workflow get <run-id>');
+    expect(result.stdout).not.toContain('Active workflows');
+  });
+});
+
 // Test the argument parsing logic used in cli.ts
 describe('CLI argument parsing', () => {
   // Mirror the actual parseArgs options from cli.ts
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -49,7 +49,23 @@ describe('workflow status arguments', () => {
       'Usage: archon workflow status [--json] [--verbose] [--events]'
     );
     expect(result.stderr).toContain('archon workflow get <run-id>');
-    expect(result.stdout).not.toContain('Active workflows');
+    expect(result.stdout).toBe('');
+  });
+});
+
+describe('workflow get arguments', () => {
+  it('rejects extra positional arguments', () => {
+    const result = spawnSync(
+      process.execPath,
+      [join(import.meta.dir, 'cli.ts'), 'workflow', 'get', 'abc123', 'accidental-extra'],
+      { encoding: 'utf8' }
+    );
+
+    expect(result.status).toBe(1);
+    expect(result.stderr).toContain(
+      'Usage: archon workflow get <run-id> [--json] [--verbose] [--events]'
+    );
+    expect(result.stdout).toBe('');
   });
 });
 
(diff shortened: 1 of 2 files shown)


Wirasm:
# Review report

## Verdict

**Ready.** Both findings on this PR's diff are fixed and verified at head `e4a35241`.

## Reviewed head SHA

`e4a352416540d7c88a4ae5e1d0ca9f4ce7f3d68e`

## Findings

| Finding | Verdict |
| --- | --- |
| R1 — `workflow get` silently discarded extra positional arguments | **Fixed** at `e4a35241`: `packages/cli/src/cli.ts:632-646` rejects `positionals[3]`; `cli.test.ts:56-69` asserts status 1, usage text, and empty stdout. |
| R2 — the status regression test could miss overview output | **Fixed** at `e4a35241`: the test asserts empty stdout, excluding both overview strings. |

## Note on the earlier not-ready verdict

The previous version of this comment reported three Important findings. All three concerned workflow files that were not part of this PR's diff — they rode into the review scope from untracked local state in the runner's checkout (a scope defect since fixed in the review workflow, and all three findings were applied to that workflow). None of them pertains to this PR.

## Lens roster

Round 1 (full): code, seams, tests, errors, docs ran; comments, types, simplify skipped by the chooser. Round 2 (light): all lenses ran; prior findings verified fixed.

Issue #2623, opened 2026-08-18
feat(workflows): lift the include-in-loop_group exclusion — the review ⟲ fix loop needs a composed body


Wirasm:
## Problem

`include:` is still rejected inside a `loop_group` body. That forces any bounded review/fix, validate/repair, or iterate-until-accepted workflow to copy the included block's nodes into the group or manually unroll a fixed number of rounds.

The first concrete consumer is the bundled delivery workflow: repeat the packaged review block and packaged implementation block until review reports that the work is ready. Its current fixed two-round unroll is an honest workaround, but the engine is dictating the workflow shape and defeating reuse.

## Product intent

A `loop_group` body may be assembled from reusable workflow blocks with `include:`.

This remains load-time composition:

- Included blocks expand into the parent workflow's static DAG before execution.
- No runtime sub-run or runtime-resolved graph is introduced.
- The expanded nodes repeat under the group's ordinary iteration semantics.
- Included inputs, declared returns, namespacing, and activation boundaries behave as they do for includes outside a group.
- The group completion condition can consume an included block's declared output.
- `$LOOP_PREV`, sessions, resume, artifacts, events, and accounting behave consistently whether a body node was written inline or arrived through composition.

The prerequisite completion-reference and composed-boundary work from #2621 and #2622 has landed. This issue owns only the remaining include-in-group exclusion.

## Acceptance

- A `loop_group` body can contain one or more `include:` nodes.
- Loading produces one fully expanded, statically validated DAG.
- Every group iteration re-executes the expanded body with normal loop-group semantics.
- An included block's declared return can drive the group's completion condition.
- The bundled delivery workflow can express its review/fix convergence loop without copying block internals or hard-coding a fixed number of rounds.
- Resume, artifacts, audit events, and accounting remain consistent with equivalent inline body nodes.

## Related

- #2123 - bundled defaults rewrite and first consumer
- #2512 - runtime-width composition; a separate capability
- #2621 - body completion references, delivered
- #2622 - composed activation boundaries, delivered

Wirasm:
# Reusable blocks compose inside loop-group bodies

**Plan ID:** `loop-group-body-includes`
**Source PRD:** None
**PRD Phase:** None
**Source Issue:** https://github.com/coleam00/Archon/issues/2623
**Plan Publication:** https://github.com/coleam00/Archon/issues/2623#issuecomment-5342889982

## Outcome

[...]

PR #2630, opened 2026-08-19
fix(workflows): loop group checks can read body outputs

### LINKED ISSUE #2621: fix(workflows): loop_group.until_bash cannot reference body nodes — the documented pattern is load-rejected
## Problem

`loop_group.until_bash` node-output refs are validated as a text source of the OUTER node (`packages/workflows/src/loader.ts:564-586`): the scan checks refs against the outer scope's `nodesById` (+ enclosing scopes) and never adds the group's own body ids. But the runtime substitutes `until_bash` against the CURRENT ITERATION'S outputs — the body nodes. Load-time validation and runtime semantics disagree, and the documented pattern loses:

> `until_bash: '[ "$decide.output.done" = "true" ]'` — node-output refs are substituted into until_bash against the current iteration's outputs (AGENTS.md, loop completion channels)

Verified empirically 19 Aug while authoring the bundled-defaults rewrite (#2123): three spellings all rejected with `references unknown node` — an include alias (`$rev.output`), a flattened composed sink id (`$rev__synthesize.output`), and a **plain bash body node** (`$ready-flag.output`). The last one is the documented shape.

The error message compounds the confusion: it says "In a composed workflow, pass caller data through declared 'inputs:'" even when nothing is composed.

## Shipped workaround (in the sdlc/deliver package)

A deterministic marker file: a body bash node reads the validated boolean and touches `$ARTIFACTS_DIR/…/.ready`; `until_bash` tests file existence. Works, but it is exactly the kind of glue a correct validator makes unnecessary.

## Proposed solution

Validate the `loop_group.until_bash` source against the union of the outer scope AND the group's own body node ids (including include-provided sinks after expansion), matching what the executor substitutes. A body-scoped ref should also produce a body-appropriate error message when genuinely unknown.

## Acceptance

- `until_bash` referencing a plain body node's output loads and terminates the loop as documented.
- A ref to a genuinely unknown node still fails at load, with a message that does not misattribute the failure to composition.

Wirasm:
## Problem and outcome

`loop_group.until_bash` is documented to read the current iteration's body outputs, but load-time validation rejected those references before execution and then incorrectly suggested composition inputs as the remedy.

- **Outcome:** Completion checks can reference direct body-node outputs, and unknown references fail with an accurate body/current-scope diagnostic.
- **Invariant:** The loader accepts exactly the direct body and visible current/enclosing DAG node IDs that runtime exposes to the group after each iteration.
- **Scope boundary:** This does not change runtime substitution, schemas, reference syntax, or the separate include-in-loop-group work tracked in #2623.
- **Root cause:** The loader scanned the group's own `until_bash` against only its current and enclosing DAG maps before the later recursive validation created the body-node map.

## Review guidance

- **Feedback requested:** Correctness of the field-specific scope boundary and the unknown-reference diagnostic.
- **Start here:** `packages/workflows/src/loader.ts:509` — this is the load-bearing reference-source scan and lookup change.
- **Review order:** Loader scope change, focused loader regressions, then the strengthened runtime completion proof.
- **Lower-attention areas:** The executor implementation is unchanged; its test now drives the existing completion path from `$bump.output` instead of rereading the counter file.
- **Known risk or uncertainty:** None. Direct body visibility and nested-group aggregation are already established runtime behavior.

## Solution

The existing output-reference scan now associates only `loop_group.until_bash` with its owning group's direct body nodes. Those IDs join the current and enclosing DAG scopes for that field without broadening visibility for unrelated outer-node fields. A miss on this body-aware field gets a scope-specific error, while the existing composition guidance remains intact for other reference surfaces.

This restores the documented workflow-language pattern without adding YAML surface or changing the executor. Direct nested group nodes remain visible only through their aggregate output, matching runtime; nested descendants are not admitted.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `until_bash: "test $ready-flag.output = ready"` was rejected while loading. | The workflow loads and the group can complete from the current body's output. |
| **Failure behavior** | A dangling body ref emitted composition-only advice. | It remains a load error and identifies the loop-group body/current DAG scope. |

## Validation

- `cd packages/workflows && bun test src/loader.test.ts -t "loop_group until_bash"` — 2 pass, 0 fail; proves valid body refs load and dangling refs fail precisely.
- `cd packages/workflows && bun test src/dag-executor.test.ts -t "until_bash reads the current iteration body output"` — 1 pass, 0 fail; proves the current body output terminates the group.
- `bun run validate` — passed on commit `9b0af728`; generated checks, type-checks, zero-warning lint, formatting, installer tests, isolated package tests, and script tests all completed successfully.
- **Not verified:** Nothing material; the focused tests cover both changed load behavior and the existing runtime contract, and the complete repository gate is green.

## Links

- Closes #2621
- Related #2623
- Plan: https://github.com/coleam00/Archon/issues/2621#issuecomment-5340449571

## Summary by CodeRabbit

* **Bug Fixes**
  * Loop-group completion conditions can now reference outputs from nodes within the loop body.
  * References to nodes in the current or enclosing workflow scope are validated correctly.
  * Invalid references now provide clearer guidance about available scopes.
  * Workflow usage and costs are preserved across failures, cancellations, pauses, fatal errors, and resumed runs.
  * Loop execution correctly completes after the expected number of iterations when using body-output conditions.

* **Tests**
  * Added coverage for loop-group output references, workflow resumption, cost tracking, conditional branches, and loop event handling.

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -16288,10 +16288,11 @@ describe('executeDagWorkflow -- loop_group node', () => {
     expect(receivedPrompts[1]).toContain('iter-1-draft');
   });
 
-  it('INSTANCE 3: until_bash deterministic gate completes on exit 0', async () => {
+  it('until_bash reads the current iteration body output and completes', async () => {
     // No `until` signal from AI; completion is decided solely by until_bash exit code.
-    // The body is a pure bash node that increments a counter file each iteration;
-    // until_bash exits 0 once the counter reaches 2. No AI node → sendQuery unused.
+    // The body is a pure bash node that increments a counter file each iteration and
+    // emits the count; until_bash exits 0 once that current body output reaches iter 2.
+    // No AI node → sendQuery unused.
     const counterFile = join(testDir, 'iter-counter');
     // The path is interpolated into REAL bash scripts: on Windows, join() yields
     // backslashes that bash strips as escapes. Use forward slashes + quoting so
@@ -16311,7 +16312,7 @@ describe('executeDagWorkflow -- loop_group node', () => {
           // channel that could never fire.
           max_iterations: 5,
           fresh_context: false,
-          until_bash: `test "$(cat ${counterRef} 2>/dev/null || echo 0)" -ge 2`,
+          until_bash: "test $bump.output = 'iter 2'",
           nodes: [
             {
               id: 'bump',
@@ -16341,7 +16342,8 @@ describe('executeDagWorkflow -- loop_group node', () => {
       minimalConfig
     );
 
-    // until_bash exits 0 once the counter reaches 2 → completes after 2 iterations.
+    // until_bash sees `bump` from the current scoped output map and exits 0 once it
+    // reaches iter 2 → completes after 2 iterations.
     // The counter file holds the final iteration count (2).
     const { readFile } = await import('fs/promises');
     const finalCount = parseInt((await readFile(counterFile, 'utf8')).trim(), 10);
--- a/packages/workflows/src/loader.test.ts
+++ b/packages/workflows/src/loader.test.ts
@@ -3933,6 +3933,60 @@ nodes:
       expect(result.workflows).toHaveLength(1);
     });
 
+    it('should load loop_group until_bash referencing a direct body output', async () => {
+      const workflowDir = join(testDir, '.archon', 'workflows');
+      await mkdir(workflowDir, { recursive: true });
+
+      await writeFile(
+        join(workflowDir, 'loop-group-body-completion-ref.yaml'),
+        `
+name: loop-group-body-completion-ref
+description: Group completion reads the current body output
+nodes:
+  - id: group
+    loop_group:
+      until_bash: "test $ready-flag.output = ready"
+      max_iterations: 2
+      nodes:
+        - id: ready-flag
+          bash: "echo ready"
+`
+      );
+
+      const result = await discoverWorkflows(testDir, { loadDefaults: false });
+      expect(result.errors).toHaveLength(0);
+      expect(result.workflows).toHaveLength(1);
+    });
+
+    it('should reject unknown loop_group until_bash refs with body-scope guidance', async () => {
+      const workflowDir = join(testDir, '.archon', 'workflows');
+      await mkdir(workflowDir, { recursive: true });
+
+      await writeFile(
+        join(workflowDir, 'loop-group-unknown-completion-ref.yaml'),
+        `
+name: loop-group-unknown-completion-ref
+description: Group completion references a node outside its visible scopes
+nodes:
+  - id: group
+    loop_group:
+      until_bash: "test $ghost.output = ready"
+      max_iterations: 2
+      nodes:
+        - id: ready-flag
+          bash: "echo ready"
+`
+      );
+
+      const result = await discoverWorkflows(testDir, { loadDefaults: false });
+      expect(result.workflows).toHaveLength(0);
+      expect(result.errors).toHaveLength(1);
+      expect(result.errors[0].error).toContain("field 'loop_group.until_bash'");
+      expect(result.errors[0].error).toContain("unknown node '$ghost.output'");
+      expect(result.errors[0].error).toContain('loop_group body or current/enclosing DAG scope');
+      expect(result.errors[0].error).not.toContain('In a composed workflow');
+    });
+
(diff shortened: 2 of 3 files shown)


Wirasm:
---
pr: 2630
base: dev
head: fix/2621-loop-group-until-bash-refs
reviewed: 2026-08-19T11:00:46Z
reviewed_head: ca708846c5f010f344189eb79d0ca713e6d6a5ea
verdict: READY TO MERGE
open_findings: 0
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2630#issuecomment-5340656017
---



# PR Review: #2630 — fix(workflows): loop group checks can read body outputs

## Signal

**READY TO MERGE:** the prior reviewed head is not an ancestor after the clean rebase, so this head refresh used the required full-review fallback over `0710dc8ec3ae2c3f425adee30724fc733ea3aa32..ca708846c5f010f344189eb79d0ca713e6d6a5ea`. Fresh code and seam reviewers found no new defects: `loop_group.until_bash` still admits direct body plus visible current/enclosing DAG IDs, runtime substitutes the current iteration's body outputs, and unknown body references receive the precise non-composition diagnostic. The old and new commits have the same stable patch ID and an equal `range-diff`; prior `R1` remains declined because it is pre-existing lexical-scope policy outside #2621. Required local validation and all GitHub checks are green at the rebased head.

- **Blocking:** 0
- **Non-blocking:** 1
- **Resolved since previous review:** 0
- **Tracked follow-ups:** None; existing #2623 remains the separate include-in-loop-group feature
- **Validation:** Rebase-equivalence checks, focused loader/runtime tests, complete `bun run validate`, and all required GitHub checks passed at `ca708846`

## Findings

|---|---|---|---|
| `R1` | Important | Existing current/enclosing lookup admits outputs that may not yet be available at runtime | DECLINED |

## Detailed Findings

<details>
<summary><code>R1</code> — Current/enclosing lookup does not encode runtime availability</summary>

**Impact:** A pre-existing self-reference such as `$group.output`, or a reference to a declared downstream node, can load even though that output is absent when `until_bash` runs. Whole-output substitution then yields an empty string and the group can exhaust its iterations instead of failing at load time.

**Evidence:** The base branch already accepts any ID in `nodesById` or `enclosingNodes` at `packages/workflows/src/loader.ts:578`; runtime seeds iteration evaluation from `outerNodeOutputs` at `packages/workflows/src/dag-executor.ts:3702` before direct body results are added, then substitutes the group check against that map at `packages/workflows/src/dag-executor.ts:3871`. Both fresh reviewers independently confirmed that current/enclosing acceptance predates this patch and that #2621 explicitly requires preserving those scopes.

**Required outcome:** No correction is required in #2630. If output availability should replace lexical current/enclosing visibility across the validator, it needs a separately scoped product decision and regression surface rather than being folded into this body-reference fix.

**Found by:** `seam-analyzer`

**Disposition:** DECLINED — the finding is pre-existing and outside the authoritative outcome. Changing it here would violate the requested invariant, expand the PR beyond #2621, and couple this narrow fix to a broader output-availability policy.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | No additional findings; prior `R1` disposition verified |
| seams | No additional findings; prior `R1` disposition verified |

## Validation

|---|---|---|
| `git range-diff 9b0af728^! ca708846^!` plus stable `git patch-id` | PASS | Range-diff reports equality and both commits have patch ID `92e100df7e113cb4ae8dc54afc2ab5f8aee8b3cc`; the rebase changed only the parent commit |
| `bun run validate` | PASS | Exit 0 at `ca708846`; generated checks, type-check, zero-warning lint, formatting, installer tests, isolated package tests, and script tests passed |
| `cd packages/workflows && bun test src/loader.test.ts -t "loop_group until_bash"` | PASS | 2 pass, 0 fail, 8 assertions; valid direct-body reference loads and unknown reference gets body-scope guidance |
| `cd packages/workflows && bun test src/dag-executor.test.ts -t "until_bash reads the current iteration body output"` | PASS | 1 pass, 0 fail; current-iteration body output completes the group |
| `gh pr checks 2630 --repo coleam00/Archon` | PASS | Exit 0; Ubuntu, Windows, schema-upgrade, docker-build, and CodeRabbit all passed at `ca708846` |

## Verdict

**READY TO MERGE**

No open finding remains. The rebased patch is behaviorally identical to the previously reviewed change, the requested scope still matches runtime behavior, and all required local and GitHub validation passed. This review does not merge the PR.

PR #2631, opened 2026-08-19
fix(web): align builder node reference validation

### LINKED ISSUE #2590: fix(web): the builder flags $INPUTS.output as a reference to an unknown node
## Problem

The engine's `$<nodeId>.output` reference scan skips the reserved `INPUTS` scope; the console
builder's copy of that scan does not. So a text body containing `$INPUTS.output` — a workflow
input that happens to be named `output` — is reported by the builder as a reference to an unknown
node called `INPUTS`.

Engine (`packages/workflows/src/loader.ts:528-532`):

```ts
while ((m = outputRefPattern.exec(source.text)) !== null) {
  const refNodeId = m[1];
  // `$INPUTS.name` is an input macro. In particular, `$INPUTS.output` also
  // matches the canonical node-ref grammar, so the macro must take precedence.
  if (refNodeId === WHEN_INPUTS_SCOPE) continue;
```

Web (`packages/web/src/lib/node-ref.ts`, `findOutputRefs`) has no such skip. Verified by running it:

```
"compare $INPUTS.output with $real.output" => ["INPUTS","real"]
```

The consumer is `validateContent` (`packages/web/src/experiments/console/builder/validation/content.ts`),
which turns each ref outside the node's transitive upstream into a `content.var.unknown` issue:

> node 'x' references '$INPUTS.output' which is not an upstream dependency

## Impact

Low. It is a `warning`, never an `error`, so it does not block saving a workflow, and it fires only
for the exact spelling `$INPUTS.output` in a prompt/bash/script/approval body — an input named
`output`. `$INPUTS.mode` and every other name are unaffected (they do not match the `.output` ref
grammar at all).

## Why it is filed rather than fixed in #2591

#2591 is the urgent `dev`-unblocking fix for the `when:` half of the same family, and it was kept
to the `when:` atom grammar deliberately. This is a different scan, in a different module, with a
different severity, and fixing it properly means adding a parity assertion for the SKIP RULE as
well as the pattern — `node-ref-parity.test.ts` currently pins `OUTPUT_REF_SOURCE` byte-for-byte
but says nothing about which ids each side then filters out. That is a real addition, not a
one-liner, and it should not ride an urgent fix.

## Proposed solution

1. Teach `findOutputRefs` to skip the reserved `INPUTS` scope, mirroring the loader's comment
   (the scope name is a bare string on the web side — `@archon/web` cannot import
   `WHEN_INPUTS_SCOPE`, same constraint as the rest of this family).

   Fixing `findOutputRefs` also silences the legacy dashboard hook
   (`packages/web/src/hooks/useBuilderValidation.ts:167`), which scans `node.data.when` with the
   same function — expected collateral, not extra scope, and not a second fix to write.
2. Extend `scripts/node-ref-parity.test.ts` with a differential entry for the scan itself, the way
   #2591 added one for the atom parser: feed both sides the same text and compare the id sets. That
   pins the skip rule, which byte-identity on the pattern alone cannot.

## Scope

- Package(s): `web` (+ `scripts/` for the guard)
- Breaking change? `No`

## Related

- #2591 (the `when:` half of the same divergence; fixed there)
- #2567 (the original builder-vs-engine node-id drift)
- #2570 (the parity guard this would extend)
- #2488 (the engine hand-writes this grammar in several places)

### LINKED ISSUE #2607: fix(web): the builder checks when: grammar but never its node references
`validateContent` scans a node's text bodies for `$<nodeId>.output` references and warns when one points outside the node's transitive upstream set (`content.var.unknown`). It also validates `when:` — but only its **grammar**:

```ts
// content.ts — the `when:` parse check
const result = parse(when);
if (!result.ok) { /* content.when.parse */ }
```

The node ids inside a `when:` are never checked. So an author can write

```yaml
when: "$ghost.output == 'YES'"
```

get no warning while editing, and have the engine reject the workflow at load — `validateDagStructure` scans `when:` and returns *"references unknown node '$ghost.output'"*.

## Why now

`when:` is a slightly harder surface than the others, which is presumably why it was left: the condition grammar accepts a **shorthand** the prompt surfaces do not. `$verify.exit_code == '0'` is a node reference even though it has no `.output` segment, so a scan built on the canonical `$id.output` pattern misses it. The engine handles this with a second pattern (`WHEN_REF_SOURCE` in `loader.ts`, and the extracted `when-atom.ts` since #2579) alongside the canonical one.

The builder already has the pieces: `parse()` returns the atom structure, and `computeUpstream` is right there in the same function.

## Why it is worth doing

This is pre-existing — the same on `dev` before #2604 — but it is now the *only* remaining gap in that scan. #2604 added `systemPrompt`, `agents.*.prompt`/`.description`, `loop.until_bash`, `approval.on_reject.prompt` and cancel bodies, and added a `KEEP IN SYNC` note in `loader.ts` naming the builder as the fourth consumer of the node-ref surface set. With `when:` still unscanned, that note is true of every surface except one — which is the state where a comment quietly stops being accurate.

Severity is UX, not correctness: the engine still rejects the workflow, so the failure is late rather than silent.

## Links

- Related #2604, #2579

Wirasm:
## Problem and outcome

The workflow builder checked `when:` syntax without checking whether its node references were valid, and its shared output-reference scan incorrectly treated the reserved `$INPUTS.output` macro as a node named `INPUTS`.

- **Outcome:** Builder validation now warns for unknown or non-upstream `when:` node references, including `$node.exit_code`, while accepting `$INPUTS.output` in both current and legacy builder validation.
- **Invariant:** Web validation remains aligned with the engine's existing grammar and reserved input scope without importing runtime code from `@archon/workflows`.
- **Scope boundary:** This changes authoring-time builder diagnostics only; workflow engine execution and YAML grammar are unchanged.
- **Root cause:** The builder discarded the AST after a successful `when:` parse, and the web scanner copied the engine's lexical pattern without its post-match `INPUTS` exclusion.

## Review guidance

- **Feedback requested:** Correctness of upstream reference validation and web/engine parity at the reserved input scope.
- **Start here:** `packages/web/src/experiments/console/builder/validation/content.ts:113` — the successful `when:` AST is now reused to validate node references.
- **Review order:** `content.ts`, `node-ref.ts`, their focused consumer tests, then `scripts/node-ref-parity.test.ts`.
- **Lower-attention areas:** Test corpus additions are focused regressions for unknown, non-upstream, shorthand, transitive-upstream, and reserved-scope references.

## Solution

Successful `when:` parses already produce discriminated node and input atoms, so the builder now collects only node atoms and checks them against the same transitive upstream set used by content-reference validation. The shared web scanner keeps its existing engine-aligned lexical pattern and excludes exact-case `INPUTS` after matching, matching the engine while preserving ordinary near-name node ids.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Invalid `when:` node references were accepted by the builder, while `$INPUTS.output` raised false unknown-node warnings. | Invalid or non-upstream `when:` node references warn, and `$INPUTS.output` is accepted by both builder validation paths. |
| **Failure behavior** | Some workflows failed only when loaded by the engine; valid input references looked invalid while editing. | The builder reports the invalid `when:` reference at its `when` field and preserves the engine's reserved input scope. |

## Validation

- `cd packages/web && bun test src/lib/node-ref.test.ts src/hooks/useBuilderValidation.test.ts src/experiments/console/builder/validation/content.test.ts src/experiments/console/builder/validation/when-grammar.test.ts src/experiments/console/builder/validation/validate.test.ts` — 59 pass, 0 fail, 129 assertions — proves both builder consumers and `when:` reference validation.
- `bun test ./scripts/node-ref-parity.test.ts` — 73 pass, 0 fail, 123 assertions — proves effective reserved-scope parity between the web helper and engine loader.
- `bun run validate` — exit 0 — proves generated checks, type-checks, zero-warning lint, formatting, installer tests, package tests, and script tests.
- **Not verified:** Nothing material; the changed observable outputs are deterministic validation issues covered by focused tests, and the full repository gate passed.

## Links

- Closes #2607
- Closes #2590
- Plan: https://github.com/coleam00/Archon/issues/2607#issuecomment-5340495205
- Companion plan publication: https://github.com/coleam00/Archon/issues/2590#issuecomment-5340576974

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow validation no longer flags reserved `$INPUTS` references as dangling node references.
  * Added validation for `when` expressions to warn about references outside a node’s upstream dependencies.
  * Improved handling of direct, compound, shorthand, and transitive node references.
* **Tests**
  * Expanded coverage for valid workflow inputs and reference-parsing edge cases.

### DIFF
--- a/packages/web/src/experiments/console/builder/validation/content.test.ts
+++ b/packages/web/src/experiments/console/builder/validation/content.test.ts
@@ -47,6 +47,21 @@ describe('validateContent', () => {
     expect(issues.filter(i => i.rule === 'content.var.unknown')).toEqual([]);
   });
 
+  test('a workflow input named output is not treated as a node ref', () => {
+    const issues = validateContent(
+      wf([
+        {
+          id: 'use',
+          variant: 'prompt',
+          base: {},
+          data: { prompt: 'Read $INPUTS.output.' },
+        },
+      ])
+    );
+
+    expect(issues.filter(i => i.rule === 'content.var.unknown')).toEqual([]);
+  });
+
   test('self-reference warns (a node is not its own upstream)', () => {
     const issues = validateContent(
       wf([{ id: 'me', variant: 'prompt', base: {}, data: { prompt: 'loop on $me.output' } }])
@@ -207,6 +222,75 @@ describe('validateContent', () => {
       ])
     );
     expect(bad.some(i => i.rule === 'content.when.parse')).toBe(true);
+  });
+
+  test('an unknown node ref in when warns', () => {
+    const issues = validateContent(
+      wf([
+        {
+          id: 'use',
+          variant: 'prompt',
+          base: { when: "$ghost.output == 'YES'" },
+          data: { prompt: 'y' },
+        },
+      ])
+    );
+
+    expect(issues).toContainEqual(
+      expect.objectContaining({
+        rule: 'content.var.unknown',
+        severity: 'warning',
+        path: { nodeId: 'use', field: 'when' },
+      })
+    );
+  });
+
+  test('every non-upstream node in a compound when warns, including shorthand', () => {
+    const issues = validateContent(
+      wf([
+        { id: 'sibling', variant: 'bash', base: {}, data: { bash: 'exit 0' } },
+        {
+          id: 'use',
+          variant: 'prompt',
+          base: {
+            when: "$ghost.output == 'YES' || $sibling.exit_code == 0 && $INPUTS.mode == 'fast'",
+          },
+          data: { prompt: 'y' },
+        },
+      ])
+    );
+
+    const warnings = issues.filter(i => i.rule === 'content.var.unknown');
+    expect(warnings).toHaveLength(2);
+    expect(warnings.every(i => i.path.field === 'when')).toBe(true);
+    expect(warnings.map(i => i.message).join('\n')).toContain("node 'ghost'");
+    expect(warnings.map(i => i.message).join('\n')).toContain("node 'sibling'");
+    expect(warnings.map(i => i.message).join('\n')).not.toContain('INPUTS');
+  });
+
+  test('direct and transitive upstream when refs pass in canonical and shorthand forms', () => {
+    const issues = validateContent(
+      wf([
+        { id: 'root', variant: 'prompt', base: {}, data: { prompt: 'x' } },
+        {
+          id: 'middle',
+          variant: 'bash',
+          base: { depends_on: ['root'] },
+          data: { bash: 'exit 0' },
+        },
+        {
+          id: 'use',
+          variant: 'prompt',
+          base: {
+            depends_on: ['middle'],
+            when: "$root.output.status == 'ready' && $middle.exit_code == 0 && $INPUTS.mode == 'fast'",
+          },
+          data: { prompt: 'y' },
+        },
+      ])
+    );
+
+    expect(issues.filter(i => i.rule === 'content.var.unknown')).toEqual([]);
   });
 });
 
--- a/packages/web/src/experiments/console/builder/validation/content.ts
+++ b/packages/web/src/experiments/console/builder/validation/content.ts
@@ -1,8 +1,8 @@
 /**
  * Content validation: scan text bodies for `$<nodeId>.output` references that
  * point outside the node's transitive upstream set, and verify each `when:`
- * expression parses. Code spans are stripped first so referenced ids inside
- * fenced/inline code are not flagged.
+ * expression parses and references only upstream nodes. Code spans are stripped
+ * first so referenced ids inside fenced/inline code are not flagged.
  *
  * `content.var.unknown` is a deliberately conservative heuristic: it requires
  * the referenced node to be reachable via explicit `depends_on` edges. A
@@ -124,6 +124,26 @@ export function validateContent(workflow: BuilderWorkflow): Issue[] {
             path: { nodeId: node.id, field: 'when' },
           })
         );
+      } else {
+        const refIds = new Set<string>();
+        for (const group of result.ast.or) {
+          for (const atom of group) {
+            if (atom.kind === 'node') refIds.add(atom.nodeId);
+          }
+        }
+        for (const refId of refIds) {
+          if (!upstream.has(refId)) {
+            issues.push(
+              makeIssue({
+                rule: 'content.var.unknown',
+                severity: 'warning',
+                source: 'client-debounced',
+                message: `node '${node.id}' when references node '${refId}' which is not an upstream dependency`,
+                path: { nodeId: node.id, field: 'when' },
+              })
+            );
+          }
+        }
       }
     }
   }
--- a/packages/web/src/hooks/useBuilderValidation.test.ts
+++ b/packages/web/src/hooks/useBuilderValidation.test.ts
@@ -59,4 +59,16 @@ describe('getDebouncedIssues — $nodeId.output references', () => {
   test('ignores a reference whose id starts with a digit', () => {
     expect(refIssues([node('use', { promptText: 'cost is $1.output' })])).toEqual([]);
   });
+
+  test('does not treat the reserved $INPUTS.output macro as a node ref', () => {
+    expect(
+      refIssues([
+        node('use', {
+          when: "$INPUTS.output == 'ready'",
+          promptText: 'read $INPUTS.output',
+          bashScript: 'echo $INPUTS.output',
+        }),
+      ])
+    ).toEqual([]);
+  });
 });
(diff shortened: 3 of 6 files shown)


Wirasm:
---
pr: 2631
base: dev
head: fix/2607-builder-when-refs
reviewed: 2026-08-19T11:07:35Z
reviewed_head: ef3d12f9e408e5b5cf23658a67866329437d1239
verdict: READY TO MERGE
open_findings: 0
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2631#issuecomment-5340754963
---



# PR Review: #2631 — fix(web): align builder node reference validation

## Signal

READY TO MERGE at rebased head `ef3d12f9e408e5b5cf23658a67866329437d1239` against base `0710dc8ec3ae2c3f425adee30724fc733ea3aa32`. Because the previous reviewed head `830ad47f9b4f5923ebeeb37c6798f0758e217181` is not an ancestor after the rebase, correction verification fell back to a full review. The feature patch is stable-patch equivalent and the intervening base changes did not touch the six PR files; both mandatory reviewers confirmed the accepted typed `when:` and reserved `$INPUTS.output` outcomes on the current head. The pre-existing code-span mismatch remains tracked separately in #2632, and all fresh local and GitHub gates passed.

- **Blocking:** 0
- **Non-blocking:** 1
- **Resolved since previous review:** 0 — rebase only; the prior finding remains tracked
- **Tracked follow-ups:** 1 — https://github.com/coleam00/Archon/issues/2632
- **Validation:** Rebase integrity, focused web behavior, cross-package parity, full repository validation, and every required GitHub check passed at the current head.

## Findings

|---|---|---|---|
| `R1` | Suggestion | Markdown code spans bypass console edit-time node-reference diagnostics even though the engine treats those references as live | TRACKED FOLLOW-UP |

## Detailed Findings

<details>
<summary><code>R1</code> — Markdown code spans bypass console node-reference diagnostics</summary>

**Impact:** A reference such as `` `$ghost.output` `` produces no console builder warning but is rejected by engine loading; authors receive the error only at save/load time. Backtick-delimited template literals in `script:` bodies have the same mismatch.

**Evidence:** `packages/web/src/experiments/console/builder/validation/content.ts:23,98` strips fenced and inline code before calling `findOutputRefs`, while `packages/workflows/src/loader.ts:507` scans raw text because runtime substitution is syntax-agnostic. At current head `ef3d12f9e408e5b5cf23658a67866329437d1239`, both reviewers independently confirmed `webIssues: []` while `parseWorkflow` rejects the identical prompt's unknown `$ghost.output` reference. The stripping behavior already exists in base `0710dc8ec3ae2c3f425adee30724fc733ea3aa32`.

**Required outcome:** Validate the raw runtime text and add engine/web parity coverage for inline and fenced references as separate work in https://github.com/coleam00/Archon/issues/2632.

**Found by:** `seam-analyzer`, `code-reviewer`

**Disposition:** TRACKED FOLLOW-UP — independently reverified at the current rebased head. The mismatch predates this patch, is outside #2607/#2590's accepted typed-`when:` and reserved-input outcomes, and remains covered by verified open issue #2632.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | `R1` disposition verified; no additional findings |
| seams | `R1` reverified; no new seam findings |

## Validation

|---|---|---|
| `git merge-base --is-ancestor 830ad47f9b4f5923ebeeb37c6798f0758e217181 ef3d12f9e408e5b5cf23658a67866329437d1239` plus stable `git patch-id` comparison | PASS | The ancestry check returned 1, requiring the full-review fallback. Old and new feature commits both have stable patch ID `e0c9958de25927d9720ab207b76204bff2d72430`; the intervening base changed none of the six PR files. |
| `git diff --check 0710dc8ec3ae2c3f425adee30724fc733ea3aa32...ef3d12f9e408e5b5cf23658a67866329437d1239` | PASS | No whitespace errors in the six-file GitHub PR diff (`+151/-3`). |
| `cd packages/web && bun test src/lib/node-ref.test.ts src/hooks/useBuilderValidation.test.ts src/experiments/console/builder/validation/content.test.ts src/experiments/console/builder/validation/when-grammar.test.ts src/experiments/console/builder/validation/validate.test.ts` | PASS | 59 pass, 0 fail, 129 assertions. |
| `bun test ./scripts/node-ref-parity.test.ts` | PASS | 73 pass, 0 fail, 123 assertions; effective `$INPUTS.output` behavior and `when:` grammar parity passed. |
| `bun run validate` | PASS | Exit 0 at `ef3d12f9e408e5b5cf23658a67866329437d1239` across generated-file checks, all package type-checks, zero-warning lint, formatting, installer tests, isolated package tests, and root script tests. |
| `gh run view 32244905171 --repo coleam00/Archon` and `gh pr checks 2631 --repo coleam00/Archon` | PASS | The pull-request run completed successfully at the reviewed head: Ubuntu, Windows, schema-upgrade, Docker build/smoke test, and CodeRabbit are all green. |

## Verdict

**READY TO MERGE**

The accepted outcomes for #2607 and #2590 remain satisfied at rebased head `ef3d12f9e408e5b5cf23658a67866329437d1239` on base `0710dc8ec3ae2c3f425adee30724fc733ea3aa32`. No blocking or open findings remain; R1 stays correctly tracked in #2632, and required local and GitHub validation is green.

PR #2634, opened 2026-08-19
fix(workflows): preserve nested include input bindings

### LINKED ISSUE #2580: fix(workflows): `include:` splices $INPUTS into `when:` verbatim, so the left-hand form silently becomes an unparseable condition
## Problem

`include:` resolves `$INPUTS.<name>` by **verbatim text substitution** at load time
(`applyInputsMacro`, `include-expander.ts` — `if (node.when !== undefined) node.when = substitute(node.when)`).
That is correct for prose surfaces, but a `when:` is an expression, and substituting into its
**left-hand side** destroys it.

Verified against `dev` @ `2d545225`:

```yaml
# block.yaml
inputs:
  mode: { default: thorough }
nodes:
  - id: lhs
    bash: echo lhs
    when: "$INPUTS.mode == 'fast'"
  - id: rhs
    bash: echo rhs
    when: "$probe.output == '$INPUTS.mode'"
  - id: probe
    bash: echo fast
```

```yaml
# parent.yaml
nodes:
  - id: b
    include: blk
    with: { mode: fast }
```

Loading the parent produces (no load errors):

```
NODE b__lhs  when= "fast == 'fast'"                 <-- not a condition any more
NODE b__rhs  when= "$b__probe.output == 'fast'"     <-- correct
```

`fast == 'fast'` does not match the atom grammar, so at run time it fail-closes: the node is
**skipped** with an unparseable-`when:` warning. The author wrote a condition that reads as
obviously true and gets a skipped node.

## Why now

PR #2579 makes `when: "$INPUTS.mode == 'fast'"` work for `workflow:` sub-runs and direct
`--input` runs (they resolve the reference at evaluation time against the run's inputs, so a
value is never re-read as syntax). That makes the include path the odd one out: the same line
means two different things depending on how the workflow was composed. #2579 documents the
asymmetry and tells include authors to put `$INPUTS` on the right-hand side, but documentation
is not a guard.

## Options

1. **Reject at load.** After include expansion, a `when:` that no longer parses as an atom is a
   load error rather than a runtime skip. Cheap, and #2579 already added the shared
   `parseWhenAtom` the check would use. Risk: it would also reject an unparseable `when:`
   written directly by the author, which today is deliberately left to runtime fail-close — so
   this probably has to be scoped to conditions the macro *changed*.
2. **Give an included block an input scope.** Stop substituting into `when:` and instead carry
   the block's resolved inputs per node so the evaluator resolves them, exactly as a sub-run
   does. Uniform, and it makes the two composition paths mean the same thing — but it is a real
   design change (a flattened DAG has no per-node input scope today), and it is #2453 territory.

## Scope

- Package(s): `workflows` (`include-expander.ts`, possibly `loader.ts`)
- Breaking change? Option 1 turns a silent skip into a load error for workflows that are
  already not doing what their author wrote.

## Related

- #2579 (found it; documents the workaround)
- #2453 (owns the composed-workflow value contract)

### LINKED ISSUE #2629: fix(workflows): nested include loses the middle workflow's input binding
## What happens

A workflow that includes a workflow that itself includes workflows (A includes B, B includes C) loses B's input binding at composition: a `with:` reference to B's declared inputs inside B's body survives verbatim into the flattened DAG, and at run time fails against A's input namespace.

Observed on a real chain (`fix` includes `deliver`; `deliver` includes `implement` with `with: work: "$INPUTS.work"`; `fix` passes `with: work: <literal work order>` to deliver):

```
FAILED deliver__impl__implement (loop) — Unknown input '$INPUTS.work'. Available inputs: $INPUTS.target.
```

Everything else about the nesting is correct: discovery expands the chain with clean multi-level namespacing (`deliver__rev1__synthesize`, `deliver__fix2__record-green` — 48 nodes), the input CONTRACT validates at load (missing required rejected, defaults applied), and the dry-run executes the composed DAG. Only the binding is dropped: fix's provided `work` value never replaces the `$INPUTS.work` references that deliver's own composition left in its body for run-time resolution.

## Why the single-level case works and this doesn't

Single-level `with:` is proven live (the sdlc deliver runs behind PR #2625/#2626 fed literal work orders through deliver → implement). At single level, the block's surviving `$INPUTS.<name>` references are resolved by the RUN's declared-input binding — the run's inputs ARE the middle workflow's inputs, so it holds. Nested, the middle workflow's inputs are nobody's run inputs: the outer composition must apply its `with:` map to the included block's surviving `$INPUTS` references at expansion, and today it does not.

A name-collision variant masks the bug: when the outer workflow happens to declare an input with the SAME name the inner reference uses (both levels named `target`), run-time binding resolves it from the outer run — with the outer run's raw value, not the value the `with:` map assigned. Silently wrong rather than loudly broken.

## Expected

At composition, an `include:` of an already-composed workflow applies the provided `with:` map (plus declared defaults) to every surviving `$INPUTS.<name>` reference of the included workflow's namespace, exactly as run-time input binding would have — so a composed workflow behaves identically whether it runs standalone or inside another composition (the #1764 runs-as-authored principle, extended one level up).

## Consumer

The sdlc pack's `fix` chain (investigate → gate on rooted → deliver), on the branch behind PR #2627 — `.archon/workflows/sdlc/fix/`, blocked on this with the failing dry-run fixture (`fixtures/rooted-delivered.stubs.yaml`) committed as this issue's acceptance test: it goes green exactly when this binding works. The planned plan → deliver chain has the same dependency.

Wirasm:
## Problem and outcome

Nested include chains could dry-run a command-backed loop from its raw command source after include expansion had already rebound the command's inputs, producing an `Unknown input` error that real execution did not share. Include input substitution could also turn an authored valid `when:` condition into invalid syntax and defer that mistake to a silent runtime skip.

- **Outcome:** Dry runs now use the same compiled, input-bound loop prompt as real execution, and include expansion reports a precise load error when substitution breaks a valid `when:` expression.
- **Invariant:** Included workflows remain a load-time flattened DAG; command compilation and the existing `when:` parser remain the authoritative primitives.
- **Scope boundary:** This does not add runtime input scopes to included nodes, change direct/sub-run input semantics, or extend the `when:` language.
- **Root cause:** Dry-run simulation reopened the raw loop command instead of consuming `COMPILED_LOOP_COMMAND`; `when:` input macros were substituted without checking whether the shared parser could still parse the result.

## Review guidance

- **Feedback requested:** Correctness at the include-expansion and dry-run/executor parity seams.
- **Start here:** `packages/workflows/src/dry-run.ts:510` — command-backed loop simulation now follows the executor's compiled-command precedence.
- **Review order:** `dry-run.ts`, `dry-run.test.ts`, `include-expander.ts`, `include-expander.test.ts`, then the authoring guide.
- **Lower-attention areas:** The documentation update only aligns the existing include-input warning with the new load-time failure behavior.

## Solution

Dry-run loop simulation now prefers an inline prompt, then valid compiled command metadata, and only reloads a command when no compiled metadata exists—the same ownership boundary used by real execution. Include macro expansion reuses the shared `when:` atom parser before and after substitution and rejects only a changed valid-to-invalid expression, preserving the existing behavior for unrelated malformed author text.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A nested included loop could fail only in dry-run because the raw command still referenced the child input name. | Dry-run executes the compiled prompt containing the fully rebound outer input. |
| **Failure behavior** | An include macro could create an unparseable `when:` and the node would be skipped later. | Include expansion fails with the include id, included node id, original expression, rewritten expression, and remediation. |

## Validation

- `bun run validate` — passed — generated artifacts, type-check, zero-warning lint, formatting, installer tests, all package tests, and script tests are green.
- `cd packages/workflows && bun test src/include-expander.test.ts` — 90 passed, 0 failed — covers valid-to-invalid substitution and the unchanged malformed-expression boundary.
- `cd packages/workflows && bun test src/dry-run.test.ts` — 43 passed, 0 failed — covers the three-level include chain with a command-backed loop.
- `cd packages/workflows && bun test src/condition-evaluator.test.ts` — 92 passed, 0 failed — confirms the shared condition grammar remains unchanged.
- **Not verified:** Nothing material; the regressions use focused repository fixtures and the complete repository validation is green.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No data or YAML schema migration. Existing valid include conditions continue to expand; only substitutions that make a valid condition invalid now fail earlier. | `packages/workflows/src/include-expander.test.ts:463` |
| Rollout / rollback | Localized workflow-engine behavior with a single reversible commit. | `1fcd0075` |
| Documentation / communication | The authoring guide now documents the load-time rejection and retained right-hand-side form. | `packages/docs-web/src/content/docs/guides/authoring-workflows.md` |

## Links

- Closes #2629
- Closes #2580
- Plan: https://github.com/coleam00/Archon/issues/2629#issuecomment-5341433806
- Companion publication: https://github.com/coleam00/Archon/issues/2580#issuecomment-5341433990

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid conditions created during workflow input substitution are now detected when workflows load, with clearer error details.
  * Nested included workflows now resolve loop prompts and rebound inputs correctly during dry runs.
  * Malformed conditions that were already present are preserved unless input substitution makes them invalid.
* **Documentation**
  * Updated workflow authoring guidance to reflect stricter validation of input-based conditions.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -1275,13 +1275,14 @@ surfaces it rewrites. That makes the right-hand form work and the left-hand form
 
 ```yaml
 when: "$probe.output == '$INPUTS.mode'"   # ✅ becomes  $probe.output == 'fast'
-when: "$INPUTS.mode == 'fast'"            # ❌ becomes  fast == 'fast'  → unparseable
+when: "$INPUTS.mode == 'fast'"            # ❌ would become  fast == 'fast'
 ```
 
-The second is not a condition any more, so the node is skipped with an unparseable-`when:`
-warning at run time. The left-hand form is a `workflow:`/direct-run feature: those callers
-resolve `$INPUTS` at evaluation time, while an included block is inlined into the caller's
-own run and has no input scope left to resolve against.
+The second would no longer be a condition, so include expansion rejects it at load time and
+shows both the authored and rewritten expressions instead of allowing a runtime skip. The
+left-hand form is a `workflow:`/direct-run feature: those callers resolve `$INPUTS` at
+evaluation time, while an included block is inlined into the caller's own run and has no
+input scope left to resolve against.
 :::
 
 ### Running a workflow that declares inputs
--- a/packages/workflows/src/dry-run.test.ts
+++ b/packages/workflows/src/dry-run.test.ts
@@ -819,6 +819,53 @@ describe('dryRunWorkflow — declared inputs (#2610)', () => {
     expect(result.trace[0]?.resolvedText).toBe('Work on ship it');
   });
 
+  test('uses the input-bound compiled prompt for a nested included loop command', async () => {
+    const cwd = mkdtempSync(join(tmpdir(), 'archon-dry-run-nested-inputs-'));
+    temporaryDirectories.push(cwd);
+    mkdirSync(join(cwd, '.archon', 'commands'), { recursive: true });
+    // The raw source keeps the middle workflow's input name. Loading it at simulation time
+    // reproduces #2629; the expanded node's compiled prompt has already rebound that name.
+    writeFileSync(join(cwd, '.archon', 'commands', 'loop-impl.md'), 'Raw $INPUTS.work');
+
+    const implement = makeTestWorkflow({
+      name: 'implement',
+      inputs: { work: { required: true } },
+      nodes: [
+        { id: 'implement', loop: { command: 'loop-impl', until: 'DONE', max_iterations: 1 } },
+      ],
+    });
+    const deliver = makeTestWorkflow({
+      name: 'deliver',
+      inputs: { work: { required: true } },
+      nodes: [{ id: 'impl', include: 'implement', with: { work: '$INPUTS.work' } }],
+    });
+    const fix = makeTestWorkflow({
+      name: 'fix',
+      inputs: { target: { required: true } },
+      nodes: [{ id: 'deliver', include: 'deliver', with: { work: 'Fix $INPUTS.target' } }],
+    });
+    const { workflows, errors } = expandWorkflowIncludes(
+      new Map([
+        ['implement', implement],
+        ['deliver', deliver],
+        ['fix', fix],
+      ]),
+      new Map([['loop-impl', 'Work on $INPUTS.work']])
+    );
+    expect(errors).toHaveLength(0);
+
+    const result = await dryRunWorkflow({
+      workflow: workflows.get('fix')!,
+      userMessage: '',
+      cwd,
+      stubs: { deliver__impl__implement: '<promise>DONE</promise>' },
+      inputs: { target: 'issue 2629' },
+    });
+
+    expect(result.outcome).toBe('completed');
+    expect(result.trace[0]?.resolvedText).toBe('Work on Fix issue 2629');
+  });
+
   test('when: conditions branch on the effective input map', async () => {
     const workflow = makeTestWorkflow({
       name: 'inputs-when',
--- a/packages/workflows/src/dry-run.ts
+++ b/packages/workflows/src/dry-run.ts
@@ -12,6 +12,7 @@ import {
   substituteNodeOutputRefs,
 } from './dag-executor';
 import { evaluateCondition } from './condition-evaluator';
+import { COMPILED_LOOP_COMMAND, type LoopWithCompiledCommand } from './compiled-command';
 import { declaredFieldsFromSchema } from './output-ref';
 import { discoverScriptsForCwd } from './script-discovery';
 import {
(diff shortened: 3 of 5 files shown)


Wirasm:
---
pr: 2634
base: dev
head: fix/2629-nested-include-inputs
reviewed: 2026-08-19T11:39:48Z
reviewed_head: 1fcd0075e32a0f287551075473ab93b3f75a3c89
verdict: READY TO MERGE
open_findings: 0
scopes: [code, seams]
publication: https://github.com/coleam00/Archon/pull/2634#issuecomment-5341610949
---



# PR Review: #2634 — fix(workflows): preserve nested include input bindings

## Signal

PR #2634 is ready to merge at `1fcd0075`: both mandatory reviewers found no defects. The dry-run path now consumes the same compiled, input-bound loop command metadata as execution, and include expansion rejects only substitutions that change an originally parseable `when:` into unparseable syntax. The patch preserves standalone command fallback, existing malformed-author-text behavior, the flat-DAG model, and the established `when:` grammar. Full local validation, focused workflow suites, GitHub Actions, and the committed-diff whitespace check all passed.

- **Blocking:** 0
- **Non-blocking:** 0
- **Resolved since previous review:** Not applicable
- **Validation:** All required local and remote checks passed at `1fcd0075`.

## Findings

|---|---|---|---|

## Detailed Findings

No findings.

## Agent Coverage

| Scope | Result |
|---|---|
| code | No findings |
| seams | No findings |

## Validation

|---|---|---|
| `bun run validate` | PASS | Exit 0 at `1fcd0075`; bundled-artifact checks, type-check, zero-warning lint, formatting, installer tests, package-isolated tests, and script tests passed. |
| `cd packages/workflows && bun test src/include-expander.test.ts` | PASS | 90 passed, 0 failed, including the changed-valid-to-invalid rejection and unchanged malformed-expression boundary. |
| `cd packages/workflows && bun test src/dry-run.test.ts` | PASS | 43 passed, 0 failed, including the three-level command-backed loop binding regression. |
| `cd packages/workflows && bun test src/condition-evaluator.test.ts` | PASS | 92 passed, 0 failed; shared condition parsing and runtime fail-closed behavior remain green. |
| `git diff --check origin/dev...HEAD` | PASS | No whitespace errors in the committed PR diff. |
| `gh pr checks 2634 --repo coleam00/Archon` | PASS | Ubuntu, Windows, schema-upgrade, Docker build, docs build, and CodeRabbit checks all passed at the reviewed head. |

## Verdict

**READY TO MERGE**

No blocking or non-blocking findings remain, and all required validation passed at the reviewed head.

PR #2660, opened 2026-08-21
feat(workflows): runs read their workflow source from a copy they own, not from the target

### LINKED ISSUE #2646: design(workflows): decouple workflow source from the execution workspace
## Problem

Workflow source and the workspace a workflow executes against are currently coupled in ways that break local authoring.

A user must be able to write or change `.archon` workflows in an IDE and run them before committing or pushing. The workflow may be committed on an unmerged branch or live in a different worktree/repository from the code it targets. Requiring publication before execution would make normal workflow authoring needlessly slow.

The #2627 dogfood run against #2480 exposed both sides of the coupling:

- The `sdlc` package was loaded from its authoring worktree, but its named `project:` scripts were resolved from the isolated target worktree. `assert-changed` was therefore missing even though it existed beside the workflow definition.
- Copying/injecting the authoring workflows into the target worktree made Archon's own `check:bundled` treat `sdlc`, `widing-labs`, and `dropshipping` as repository package inputs. The implementation validator worked around that by moving all three aside; the final validator moved only one and reported a false red result on the next.

This is adjacent to #1578 and #2200, but neither settles the source model. #1578 prevents untracked workflow material from leaking into a PR; #2200 moves run output outside repositories. This issue asks where executable workflow source and its resources belong when source and target are different checkouts.

## Required invariant

**A new run can execute the exact workflow currently present on disk without requiring a commit, push, or merge, while keeping the target workspace free of authoring-source injection.**

The invariant includes:

- uncommitted workflow YAML, commands, and scripts are runnable;
- a workflow on an unmerged branch can target a worktree cut from another branch;
- workflow source may be in a different checkout or repository from the target;
- scripts execute against the target working directory while retaining their owning package resources;
- a resumed run uses the same workflow bytes it started with;
- a new run sees subsequent authoring edits;
- authoring material does not appear in the target diff or contaminate its validation.

## Status

This is an undecided design issue, not an accepted implementation plan. The leading shape is an immutable run-owned source snapshot, but it should be proved against simpler alternatives before becoming direction.

## Assumptions to test

1. Active-run hot reload is unnecessary: saving an edit and starting a new run is sufficient, while an existing run should remain reproducible.
2. A workflow's executable resource closure is finite at run creation because parent-run composition is load-time static. Runtime-authored child workflows are a separate capability already discussed in #2200.
3. A script file can live outside the target checkout and still execute with the target checkout as `cwd`.
4. Source provenance can be recorded without a Git commit by combining the source path, current HEAD when present, dirty state, and a content hash.
5. Existing YAML such as `script: assert-changed` should not need a new path grammar; the compiled node can retain resource ownership internally.

### H1 — Immutable run bundle

If Archon resolves a workflow from the invoking checkout, snapshots its definition and owned resources into Archon-managed run storage before isolation, and executes package resources from that snapshot with the target worktree as `cwd`, then local uncommitted authoring, clean target validation, and deterministic resume can coexist.

Expected observations:

- deleting or changing the authoring files after a pause does not change the resumed run;
- starting a second run picks up the changed files;
- named scripts remain resolvable in the target worktree;
- `git status` and the target repository's aggregate gate do not see the authoring package.

### H2 — Persisted source-root reference

If Archon only records the authoring source root and resolves resources there at execution time, local authoring and a clean target may work with less machinery, but resume will become nondeterministic when the author edits, switches branch, moves the checkout, or deletes a file.

The spike should try to disprove that concern rather than assume a snapshot is necessary.

### H3 — Resource ownership is internal

If discovery/compilation carries an owning source handle on every command and script reference, the runtime can distinguish package-owned resources from target-project and user-global resources without adding workflow-language surface.

## Suggested `$prp-spike`

Build the smallest isolated proof that compares H1 and H2:

1. In authoring checkout A, create an uncommitted workflow package containing YAML, a command, and a named script.
2. Start it against isolated target worktree B, which does not contain that package.
3. Prove the script body comes from A (or its snapshot) while its process `cwd` and file changes are in B.
4. Pause the run after one node, then modify or remove the source package in A.
5. Resume and verify whether the run uses its original bytes.
6. Start a new run and verify it sees A's latest bytes.
7. Run `git status --porcelain` and the target's aggregate validation in B; neither may see injected authoring files.
8. Record enough provenance to identify the exact source bytes without requiring a commit.

Verdict:

- **PROVEN** — one shape satisfies all eight checks without committing source.
- **CONDITIONAL** — the invariant holds but needs a named limitation or new authoring choice.
- **DISPROVEN** — neither shape can preserve local iteration, target cleanliness, and resume determinism together.

The spike is throwaway evidence only: no production PR.

## Decisions the spike should inform

- Snapshot the entire owning package or only the reachable resource closure?
- Store snapshots per run or in a content-addressed cache referenced by runs?
- Is the current checkout the implicit source, with an optional explicit `--workflow-source <path>` override?
- What provenance belongs in the run record and audit UI?
- How long does the source snapshot live relative to artifacts and resumability?
- Can all current command/script resolution paths carry one typed source handle, or are there bypasses?

## Non-goals

- Publishing or installing workflow packages.
- Requiring users to commit temporary authoring work.
- Mutating an active run when its source files change.
- Redesigning runtime-authored child workflows from #2200.
- Solving general target-worktree setup hooks (#2278).

## Related

- #1578
- #2200
- #2278
- #2627
- #2644

Wirasm:
## Problem and outcome

A run reads its commands and scripts from one directory and acts on another, but everything downstream re-derived both from the execution `cwd`. A workflow authored in checkout A and run against worktree B looked for `B/.archon/...` and found nothing, so running an uncommitted workflow against a separate target was impossible without committing and merging it first.

That gap was papered over by copying the whole `.archon` tree into the worktree. The copy was never scoped to executable source, so it also carried `.archon/.env` and cross-run `state/`; and because `.archon` is **tracked** in this repo, it overwrote the worktree's own files and dirtied `git status` — which is how `check:bundled` started seeing packages that came from somewhere else.

- **Outcome:** A run freezes its own `.archon/{workflows,commands,scripts}` into its artifacts at start and resolves from there for the rest of the run. The target receives nothing. `--workflow-source <path>` lets an uncommitted workflow run against another checkout, repository, or folder project with no commit, push, or merge.
- **Invariant:** A run's executable source cannot change once the run has started. Resume executes what the run began with; the next fresh run picks up authoring edits.
- **Scope boundary:** No manifest, digest, or persisted compiled graph (see *Solution*). Global (`~/.archon`) and bundled source stay live — they resolve identically from any `cwd`, so they take no part in the source/target split. Container runs keep live source: containers serve folder projects, which run in place, so source and target are already the same directory.
- **Root cause:** Discovery knew which root each resource came from and discarded it (`workflow-discovery.ts`), so command and script resolution rebuilt project paths from the execution `cwd` instead (`executor-shared.ts:337`, `script-discovery.ts:278-282`). Resume compounded it: `hydrateResumableRun` returns no definition, and parent auto-resume re-ran discovery against `parentCwd` (`executor.ts`), so an authoring edit could change a running DAG.

## Review guidance

- **Feedback requested:** Correctness of the source-versus-target split, and whether the child-discovery rule (below) is the one you want.
- **Start here:** `packages/workflows/src/workflow-source.ts` — the capture, and the two accessors whose difference is load-bearing.
- **Review order:**
  1. `workflow-source.ts` — capture, `resolveRunSourceRoot` vs `resolveChildDiscoveryRoot`
  2. `executor.ts` — capture on fresh, restore on resume, warn-and-continue when a capture is gone
  3. `executor-shared.ts`, `script-discovery.ts`, `workflow-discovery.ts` — the resolvers taking a root distinct from `cwd`
  4. `dag-executor.ts` — threading `RunLayersContext.workflowSourceRoot` to every node path
  5. `isolation/providers/worktree.ts` + deleted `core/utils/worktree-sync.ts` — the removed implicit copy
- **Lower-attention areas:** `bundled-defaults.generated.ts` (regenerated, verified by `check:bundled`); the `dag-executor.ts` call-site edits are mechanical, and every one was verified wired by walking each call's argument list rather than by eye.
- **Known risk or uncertainty:** Removing the implicit `.archon` copy is a behavior change for anyone relying on it for a **gitignored** `.archon/config.yaml`. Documented with the one-line `copyFiles` migration, but there is no runtime warning for it — say the word if you want one for a release.

## Solution

The capture mirrors relative paths, so `captureRoot` is a drop-in project root: `join(captureRoot, '.archon', 'workflows')` resolves exactly as `join(cwd, ...)` always has. That is the whole trick — every existing resolver keeps its shape and takes a different root, rather than learning a new resource-lookup protocol.

Each capture carries a small `manifest.json`: format version, engine version, origin, scopes, counts, the selected workflow, the source's own resolution settings, and a content digest. The digest is checked **against the value recorded on the run row**, not only against the capture's own files — otherwise verification is circular, since the manifest lives inside the thing it describes and a wholesale replacement passes. There is deliberately no resource table: the bytes are on disk in their natural layout, and the manifest describes the set rather than indexing it.

Resume re-reads the frozen YAML and recompiles. Persisting the compiled `WorkflowDefinition` was considered and rejected: it relocates version skew rather than removing it, from "new engine recompiles old YAML" (schema-checked, exercised on every discovery) to "new engine loads an old compiled shape" (new code, exercised only by runs paused across an upgrade). It would also have added a fourth walker to symbol metadata that `compiled-command.ts:6-10` already warns "silently vanishes at two [nesting levels]".

**Everything a static `include:` can reach is frozen**, not just project scope: global (`~/.archon/`) and the bundled defaults too. A compiled binary keeps its bundled set as constants rather than files, so a capture writes them out in the same layout a source build has on disk — which is what lets a paused binary run resume across an Archon upgrade instead of being refused for holding bundled bytes it could not verify.

**The capture is the sole executable source, so every way of not using it is closed.** A capture that cannot be taken, recorded, or verified fails the run; only a run created before captures existed resumes live, and a record that exists but cannot be parsed is not treated as absent. Continuation goes through one entry point (`resolveContinuationWorkflow`) that loads the recorded source and returns the graph, so no surface can discover live by forgetting an argument — that mistake is why chat and web resume once ran an edited DAG against pre-edit command bytes.

**A `workflow:` child is not a run until it starts**, so it resolves from the parent's *authoring* directory live and freezes its own source at its own start. This distinction was found by an existing test, not by inspection: `subrun.test.ts`'s fan-out gate-recovery case covers an author removing a gate from a child workflow and resuming the parent. Resolving the child from the parent's frozen copy re-drove the old gated definition forever, so the fix could never land. `resolveRunSourceRoot` (frozen) and `resolveChildDiscoveryRoot` (live origin) name the two cases, with a test on the difference.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Authoring `.archon` copied into every worktree; target `git status` dirty; validators saw foreign packages | Nothing written to the target; run reads a copy under its own artifacts |
| **Cross-checkout runs** | Impossible without committing and merging the workflow | `--workflow-source A --cwd B` |
| **Authoring edit mid-run** | Changed the DAG a paused run resumed into | Resume executes the source the run started with |
| **Failure behavior** | Missing command/script inside the worktree → node fails "not found" | Capture missing, altered, or unreadable at resume → the run **fails** with a pointer to start a fresh run. Only a run created before captures existed resumes live, with a warning |

## Architecture

```mermaid
flowchart LR
  subgraph A[Authoring checkout]
    SRC[".archon/workflows, commands, scripts"]
  end
  subgraph R["Run-owned artifacts"]
    CAP["workflow-source/ (frozen)"]
    META["metadata.workflow_source"]
  end
  subgraph B[Target workspace]
    CWD["cwd: provider turns, bash, git, output"]
  end
  SRC ==>|capture at run start| CAP
  CAP --> META
  CAP ==>|commands + named scripts| NODE[Node execution]
  NODE --> CWD
  META -->|resume| CAP
```

### Changed seams

|---|---|---|
| `loadCommandPrompt(deps, cwd, name, folder)` | Takes `sourceRoot`; `cwd` now owns config only | `packages/workflows/src/executor-shared.ts` |
| `discoverScriptsForCwd(cwd)` | Takes `sourceRoot` for project scope | `packages/workflows/src/script-discovery.ts` |
| `discoverWorkflowsWithConfig(cwd, loadConfig)` | Takes `sourceRoot`; config still read from `cwd` | `packages/workflows/src/workflow-discovery.ts` |
| `RunLayersContext` | New required `workflowSourceRoot: string \| undefined` | `packages/workflows/src/dag-executor.ts` — required-but-nullable so a missed context literal is a compile error |
| `workflow_runs.metadata.workflow_source` | New typed key, total validation at the read boundary | `packages/workflows/src/schemas/workflow-run.ts` — no DB migration |
| `WorktreeProvider.copyConfiguredFiles` | `.archon` no longer prepended | `packages/isolation/src/providers/worktree.ts` |
| `core/utils/worktree-sync.ts` | Deleted; replaced by `resolveWorkflowSourceRoot` | `packages/core/src/utils/workflow-source-root.ts` |

## Validation

- `bun run validate` — passed (exit 0) — all 10 steps: generated-file checks, type-check across 11 packages, `eslint --max-warnings 0`, `prettier --check`, install integrity, and every package suite.
- Mutation checks, both **seen red** — `sourceRoot ?? cwd` → `cwd` in `loadCommandPrompt` failed 3 tests; `stat` → `lstat` in the capture walk failed the symlink-dereference test. The guards bind the fix rather than passing incidentally.
- **End-to-end, uncommitted workflow in checkout A against an unrelated repo B:**
  - Control without the flag fails — `source-probe` is not found from B. With `--workflow-source A --cwd B` the run completes and the packaged script prints `helper-was-reachable`, proving a **sibling import** resolved from the capture while the process `cwd` was B.
  - Isolated (worktree) run: worktree `git status --porcelain` empty; `find .archon -type f` returns only B's own committed `.archon/workflows/.keep`. Previously A's entire `.archon` was copied in.
  - Edited `helper.ts` in A after the run: the existing run's capture still reads `helper-was-reachable`; a fresh run prints `EDITED-AFTER-THE-RUN`.
  - `--workflow-source` with `--resume` is refused with an actionable message before any run work.
- **Not verified:** container write-back with a captured source — deliberately out of scope, since containers serve folder projects that run in place and have no source/target split. Also unverified at scale: a `fan_out` with many children each captures its own source; `CAPTURE_WARN_BYTES` (64 MB) logs once when a capture is large, and cache directories are skipped, but the multiplied cost has not been measured against a big scripts tree.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Runs started before this change carry no `workflow_source` and resume against live source with a warning — the pre-existing behavior. Operators relying on the implicit copy for a **gitignored** `.archon/config.yaml` must add `.archon` to `worktree.copyFiles`; an explicitly listed `.archon` is still honored | `executor.ts` legacy branch; `worktree.test.ts` "still honors .archon when the operator lists it explicitly"; configuration reference |
| Security / permissions / data | Strictly reduces exposure: the implicit copy carried `.archon/.env` into every worktree and no longer runs | `packages/isolation/src/providers/worktree.ts` |
| Rollout / rollback | Additive on disk and in metadata; no schema change. An older binary ignores the new metadata key and resolves live | No `migrations/` change in the diff |
| Observability | `workflow.source_captured` (origin, file count, bytes), `workflow.source_restored`, `workflow.source_capture_missing`, `workflow.source_legacy_live`, `workflow.source_capture_large` | `packages/workflows/src/executor.ts`, `workflow-source.ts` |
| Documentation / communication | CLI reference gains `--workflow-source` and a "Running a workflow from another checkout" section; configuration reference replaces the "always copied automatically" claim with the migration; two bundled defaults that told agents to expect a dirty `.archon` were corrected and the bundle regenerated | `packages/docs-web/...`, `check:bundled` passing |

## Links

- Closes #2646

## Summary by CodeRabbit

- **New Features**
  - Added `--workflow-source` to run workflows from a separate directory while targeting `--cwd`.
  - Workflow sources are captured and verified for fresh, resumed, child, background, and containerized runs.
  - Added detached `resume`, `approve`, and `reject` operations with acknowledgements and validation.

- **Improvements**
  - Worktrees no longer copy `.archon` automatically; configure files explicitly when needed.
  - Improved workflow integrity and usage tracking across retries, loops, and resumed runs.
  - Added clearer recovery guidance for missing working directories.

- **Documentation**
  - Updated CLI and configuration guidance for workflow sources and worktree copying.

Follow-up filed from review: #2686 — the fresh run started by the "nothing to resume" fallback never freezes its own source.


Wirasm:
Worked through all 13 findings. Six were real and are fixed in d4dbb576; four I'm declining with evidence; two are tracked as follow-ups because they're pre-existing and need a design decision rather than a patch.

Windows CI was also failing, and it wasn't a flake — `script-discovery.ts:30` normalizes separators to `/`, so my assertion compared a normalized path against a raw `join()` result and could never match on Windows. Test-only defect, fixed.

### Fixed

| # | Finding | What I found |
|---|---|---|
| 13 | Prevent recursion through directory symlinks | **Real, reproduced.** Wrote a self-referential link and the walk recursed 16 levels, copying the same file 16 times before the kernel's ELOOP limit stopped it — and the error was swallowed as a skipped entry. Now tracks canonical ancestors and cuts at the first repeat. Regression test added and confirmed failing without the guard. |
| 5 | Propagate `workflowSourceRoot` to web background runs | **Real.** `orchestrator.ts` calls `executeWorkflow` directly and was the one surface still capturing the worktree rather than the authoring repo. |
| 3 | Keep the source root separate from the workspace root | **Real for execution, not for discovery** — see the decline note below. `--dry-run --exec-code` ran real bash/script nodes with the *authoring* checkout as `cwd`, so `--workflow-source` would have mutated the tree the workflow was only read from. Dry run and its config load now use the target. Relative `--stubs` paths resolve from `--cwd` again, matching the CLI reference. |
| 9 | Validate absolute source paths | Cheap and matches the `output_root` precedent. A relative recorded path now reads as absent instead of resolving against `process.cwd()`. Test added. |
| 1 | Don't classify all `.archon/` content as committed | Correct — `worktree.copyFiles: [.archon]` can still place authoring files there. Reworded to "usually committed, check provenance", and the "not yours to clean up" rule restated. |
| 7 | Describe capture as project-source behavior | Correct, the CLI wording was loose. It now says plainly that capture is project-scope and that home/bundled workflows are not captured. |

### Declined, with evidence

**4 — Preserve `syncArchonToWorktree` or release a breaking version.** `@archon/core` is not published: `npm view @archon/core` returns 404, nothing in `.github/workflows/` publishes any `@archon/*` package, and every consumer is a `workspace:*` sibling in this repo. There is no external consumer to break, and a deprecated re-export of a function whose implementation is deleted would be dead code.

**3 (discovery half) — load config from the target instead of the discovery root.** Declined deliberately: I tried it and it broke 7 existing tests that lock in "discover workflows from `codebase.default_cwd`, not `working_path`" for the resume/approve/reject paths. That's a deliberate, tested decision, not an oversight. A workflow's own checkout deciding its `commands.folder` and `defaults:` is the right reading for *discovery*; what the run then *does* is governed by the target's config, loaded separately in `executeWorkflow`. The genuine bug was the execution root, which is fixed above.

**12 — Derive `WorkflowSourceCapture` from a schema.** The `z.infer` rule exists to stop a hand-written interface *duplicating* a schema. There is no schema here: this is an internal function return value that is never serialized, validated, or sent over a wire. `ScriptDefinition` in `script-discovery.ts` is the same shape of thing and is a plain interface. The value that *is* persisted, `workflowSourceMetadataSchema`, is a Zod schema with `z.infer`, as the rule requires.

**10 — Keep this unit-test suite off the host filesystem.** The suite tests filesystem copying, so a real filesystem is the thing under test. Five test files in this same package already use `mkdtemp`/`tmpdir` (`artifacts-index`, `dry-run`, `dag-executor`, `executor-preamble`, `bundled-script-materialization`). The AGENTS.md rule targets undeclared network and database I/O, where a missing stub stalls until the 5s timeout; `mkdtemp` with cleanup has neither failure mode.

**6 — Use one shared `workflow-source-root` mock.** The guideline forbids mocking one path with *different* implementations in files sharing a process. All three factories are identical, and the package test script runs each file in a separate `bun test` invocation — as the finding itself notes.

**11 — Add a return type to `recorded`.** Test-local helper with sound inference; the surrounding suites don't annotate their fixtures either.

### Tracked, not patched here

**8 — Capture the source before workflow selection.** Real, and I want to be straight that it isn't fixed: there is a window between discovery and capture where an edit means the run executes the in-memory definition while storing newer bytes. Both happen inside the same run-start sequence, so it's sub-second, but the consequence is exactly the drift this PR exists to prevent. The suggested fix — rediscover the selected workflow from the completed capture — isn't viable: `executeWorkflow` is called with an in-memory `workflow` object by many tests and by programmatic callers with nothing on disk, so rediscovery would break them. Closing it properly needs capture to happen before the run row exists, which is the reserved-UUID work deliberately scoped out of this PR.

**2 — Compare against a pre-implementation baseline.** A pre-existing weakness of that guard: it accepted any tracked or untracked change as implementation output before this PR too, just with `.archon` excluded. Removing the exclusion is right now that nothing is copied in, but it does mean an operator who explicitly configures `worktree.copyFiles: [.archon]` could get a false pass. The baseline capture is a heavier change to an experimental workflow and belongs on its own.

`bun run validate` passes at exit 0, and I re-ran the cross-checkout end-to-end after the changes: the run completes, the script resolves from the capture, and the target's `git status` is still clean.

Wirasm:
## Ready to merge

Correction pass over `be2e05e5..eb677378`, one commit. `R19` is fixed at the root rather than at the symptom: `run <name> --resume` now resolves its run row *before* discovery and joins the shared continuation path, so every resume form — the flag, `resume <id>`, approve, reject, parent auto-resume — reaches `resolveContinuationWorkflow` with a run in hand. The move also deletes the later duplicate lookup, so the id-based forms stop re-finding by name a run they had already identified exactly. `cwd` is a `const` parameter never reassigned in that function, so hoisting the lookup is semantically identical to where it used to sit. The docs line that listed only three forms was corrected in the same commit.

Nothing blocking remains. Twenty-one findings raised across five rounds are closed; the four still open are all Suggestions, none of which makes the delivered outcome incorrect. `R21` is the one worth doing soon, and it is narrower than it was: the CLI now asserts the ownership sequence directly, and only the orchestrator call sites remain uncovered.

**0 blocking · 4 non-blocking · 21 resolved**

**Tracked follow-ups:** #2686 (`R21`), #2690 (`R24`), #2691 (`R23`). `R25` remains open and unfiled by decision.

**Validation:** `bun run validate` passed (exit 0) at `eb677378`, and all five hosted checks pass on that exact head.

**Method note:** this final pass was verified by the coordinator directly, without reviewer agents, at the operator's explicit request. Findings were checked by reading the call chains and by two mutation probes, recorded below. Earlier rounds' agent-sourced evidence is unchanged.

**Resolved this pass:** `R19`, `R22`

### Findings

|---|---|---|---|
| `R21` | Suggestion | No behavioral test asserts `hold`/`adopt` at the orchestrator call sites. | TRACKED FOLLOW-UP |
| `R23` | Suggestion | `resolvedContinuation` is not tied by type to the run it was resolved against. | TRACKED FOLLOW-UP |
| `R24` | Suggestion | `CapturedSourceOwner` has one unkeyed hold slot and no way to undo a premature `adopt`. | TRACKED FOLLOW-UP |
| `R25` | Suggestion | A container-prepare failure after finalize strands the moved capture outside the sweeper's scope. | OPEN |
| `R19` | Important | `workflow run <name> --resume` resolves its graph live while the executor feeds it frozen commands and scripts. | FIXED |
| `R22` | Suggestion | The CLI test's mock factory omits two newly consumed exports, so the real ones run. | FIXED |
| `R1` | Important | Chat and web resume execute the live workflow graph against the run's frozen commands and scripts. | FIXED |
| `R2` | Important | Fresh composed children can lose the parent's authoring-source origin. | FIXED |
| `R3` | Important | Captures omit the supported configured command directory and source-affecting policy. | FIXED |
| `R4` | Important | Container execution discards captured roots and resolves node resources live. | FIXED |
| `R5` | Important | Invalid metadata and source-resolution failures collapse into live/target fallback. | FIXED |
| `R6` | Important | Chat and web dispatch abandons staged captures; the primitive built to prevent it is uncalled. | FIXED |
| `R7` | Important | Source setup failure can report failure while leaving a fresh run pending. | FIXED |
| `R8` | Important | Capture verification does not anchor the loaded manifest to the digest recorded on the run. | FIXED |
| `R9` | Important | Binary-bundled workflow bytes are not frozen into the run capture. | FIXED |
| `R10` | Important | A generic directory denylist can remove dependencies from otherwise valid packaged scripts. | FIXED |
| `R11` | Suggestion | User documentation contradicts the captured-source behavior. | FIXED |
| `R12` | Suggestion | Optional internal source roots preserve a silent live-source fallback. | FIXED |
| `R13` | Important | CLI continuation discovery resolves the capture with default settings instead of the manifest's. | FIXED |
| `R14` | Suggestion | The caller that copied the "Never throws" claim still repeats it, unguarded. | FIXED |
| `R15` | Suggestion | A test comment cites `worktree-sync.test.ts`, which this PR deletes. | FIXED |
| `R16` | Suggestion | The PR description's Solution section contradicts the delivered architecture. | FIXED |
| `R17` | Important | The staged-source sweeper runs once per process, so it never runs again on a long-running server. | FIXED |
| `R18` | Suggestion | `SIGINT`/`SIGTERM` calls `process.exit(1)`, bypassing the ownership `finally`. | FIXED |
| `R20` | Suggestion | Every resume resolves the continuation twice. | FIXED |

<details>
<summary><code>R21</code> — The orchestrator ownership wiring is still untested</summary>

**Impact:** Dropping an `owner.adopt()` before a real `executeWorkflow` handoff would delete a live run's frozen source, and no orchestrator test would fail.

**Evidence:** the shared stub at `orchestrator-agent.test.ts:183-184` passes a no-op `{ hold, adopt }`, so neither call is observable by any orchestrator test.

**Correction to this finding's original evidence.** This report first cited a mutation as proof: removing `if (preparedSource) owner.adopt();` at `orchestrator-agent.ts:1128` left all three orchestrator suites green. That result does not support the claim. Follow-up analysis in #2686 established that `preparedSource` is necessarily `undefined` at `:1128` and `:1162` — the enclosing `if (resumableRun?.working_path)` block plus the early return below it force both conjuncts of `willContinueExistingRun` true, and `preparedSource` is assigned only under `if (!willContinueExistingRun)`. Those two guards cannot fire, so deleting one changes nothing and the suites would stay green even with complete coverage. The mutation was reverted and the checkout confirmed clean, but it proved nothing either way. The finding stands on the stub inspection above; the third adopt site, `:1225`, was never probed.

**Required outcome:** Record the ownership calls in the orchestrator stubs the way the CLI suite now does, and assert the sequence on one gated early return and one successful handoff.

**Disposition:** TRACKED FOLLOW-UP — #2686, which asks for this test and for the four inert references to be deleted or activated. Narrowed this pass. The CLI half is now closed and closed well: `packages/cli/src/commands/workflow.test.ts:126-143` stubs `withCapturedSource` as a faithful pass-through that records `hold`/`adopt` — with a comment giving exactly this reason — and asserts the sequence at `:1319` (nothing held), `:1339` (`['hold:/test/capture', 'adopt']`) and `:1350` (held, not adopted). That is the shape the orchestrator side still needs.

</details>

<details>
<summary><code>R23</code> — The resolved continuation is not tied to its run</summary>

**Impact:** `CommandResult.workflow` carries `resolvedContinuation`, `resumeRun` and `resumeRunId` as independent optional fields, with nothing in the type requiring the graph to belong to the run beside it. Safe today because one producer builds all three from the same run and one bridge forwards them as a set, but the pairing rests on call-site discipline while the consumer trusts the field unconditionally.

**Evidence:** `packages/core/src/types/index.ts:86` — still `resolvedContinuation?: WorkflowDefinition`, unchanged this pass.

**Required outcome:** Fold them into one value, for example `resolvedContinuation?: { run: WorkflowRun; workflow: WorkflowDefinition }`.

**Disposition:** TRACKED FOLLOW-UP — #2691. No reachable invalid state today. Note one correction: this report said a single producer built all three fields; `dev` has a second, `continueResolvedGateRun` (`orchestrator-agent.ts:1349-1410`), also correct, which is what makes the pairing worth expressing in the type.

</details>

<details>
<summary><code>R24</code> — The owner cannot express two captures or undo an adopt</summary>

**Impact:** `hold()` has a single unkeyed slot, so a second call silently replaces tracking of the first — yet its docblock invites re-calling it when a container relocates a capture, a capability the type cannot correctly back. And every `adopt()` fires before the durable `rename` inside `executeWorkflow`; if that rename throws, the run fails without deleting the staged path and the already-adopted owner will not reclaim it, with no `release()` to undo the optimistic adopt.

**Evidence:** `packages/workflows/src/executor.ts:687-696` — the `CapturedSourceOwner` interface still exposes only `hold` and `adopt`; a tree-wide grep finds no `release` or `unadopt`. Unchanged this pass.

**Required outcome:** Have the code that performs the move report adoption, rather than the caller declaring it beforehand.

**Disposition:** TRACKED FOLLOW-UP — #2690. Bounded even in the failure case: the stranded directory remains under `~/.archon/staged-source/`, so `R17`'s hourly age-based sweep reclaims it independently of owner state.

</details>

<details>
<summary><code>R25</code> — A finalized container capture can be stranded</summary>

**Impact:** If `backend.prepare()` throws after `finalizeWorkflowSource` has moved a container run's capture out of staging, the owner's `finally` reclaims the stale pre-move path — a no-op — and the finalized capture sits under the run's artifacts, outside the sweeper's scope. Narrow: it needs a folder project, a container, and a docker failure inside that window.

**Evidence:** the finalize-then-prepare sequence in `packages/cli/src/commands/workflow.ts` around `:1690-1706`; sweeper scope is `~/.archon/staged-source/` only. `git diff be2e05e5..eb677378 -- packages/isolation/ packages/workflows/src/executor.ts` is empty, so this area is untouched.

**Required outcome:** Reclaim the finalized path on a prepare failure, or leave it to the run's artifact lifecycle deliberately and say so.

**Disposition:** OPEN, unchanged.

</details>

<details>
<summary>Findings resolved this pass</summary>

<details>
<summary><code>R19</code> — FIXED, at the root</summary>

`packages/cli/src/commands/workflow.ts:977-987` resolves the run before discovery when `--resume` is passed without an explicit `continuationRun`: `continuationRun = (await workflowDb.findResumableRun(workflowName, cwd)) ?? undefined`. `resolveContinuationWorkflow` is then called for the flag form exactly as for the id forms (`:990-993`), and the graph selection at `:1044` — `continuation?.workflow ?? resolveWorkflowName(...)` — takes the frozen graph. The later duplicate lookup is gone, replaced by `resumable = continuationRun ?? null` (`:1560`), which also stops the id-based forms re-finding by name a run they already identified (#2645). A database failure during the hoisted lookup is captured in `resumeLookupError` and rethrown with the actionable message inside the resume block (`:1551-1557`); I confirmed the claim that nothing executes before it — between discovery and `:1533` there is no run-row creation, capture, worktree, container prepare, or `executeWorkflow` call. `cwd` is a `const` parameter never reassigned in the function, so the hoist cannot select a different run than the old placement would have. Docs corrected at `packages/docs-web/src/content/docs/reference/cli.md:251`.

**Verified by mutation:** replacing the new pre-discovery lookup guard with `if (false)` turns `packages/cli/src/commands/workflow.test.ts` from 256 pass / 0 fail into 252 pass / **4 fail**. The new test that names the behavior directly — "executes the graph the run froze on `run <name> --resume`, not the one in the checkout" — binds the fix rather than passing incidentally. The mutation was reverted and the checkout confirmed clean.

</details>

<details>
<summary><code>R22</code> — FIXED</summary>

`packages/cli/src/commands/workflow.test.ts:131` and `:150` now stub `withCapturedSource` and `resolveContinuationWorkflow`, closing the `mock.module` merge gap that had the real implementations running across the suite. `withCapturedSource` is deliberately stubbed as a faithful pass-through rather than mocked away, with a comment explaining that swallowing the owner would let a dropped `adopt()` pass unnoticed — which is what makes the `R21` assertions possible on this side.

</details>

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | Prior rounds; `R21` carried forward, still open |
| seams | Prior rounds; `R22` verified fixed this pass, `R25` carried forward |
| simplify | Prior rounds; `R21` carried forward |
| types | Prior rounds; `R19` verified fixed this pass, `R23` and `R24` carried forward |

This pass ran no reviewer agents, at the operator's explicit request. Every open finding was re-checked by the coordinator against `eb677378`, and the two decisive claims — `R19` fixed, `R21` still open — were each settled by a mutation probe rather than by reading alone.

#### Validation

|---|---|---|
| `bun run validate` | PASS | Exit 0 in the review checkout at `eb677378`, on a confirmed-clean tree. An earlier run in this pass reported a failure while a coordinator mutation was still applied; it was re-run after reverting and passed. |
| `test (ubuntu-latest)` | PASS | Hosted, on exact head `eb677378`. |
| `test (windows-latest)` | PASS | Hosted, on exact head `eb677378`. |
| `schema-upgrade` | PASS | Hosted, on exact head `eb677378`. |
| `docker-build` | PASS | Hosted, on exact head `eb677378`. |
| `Build docs site` | PASS | Hosted, on exact head `eb677378`. |
| `packages/cli/src/commands/workflow.test.ts` | PASS | 256 pass / 0 fail; 252/4 under the `R19` mutation. |
| Orchestrator suites | PASS | 245 / 72 / 10 pass, zero failures. The `R21` mutation also left them green, but see the correction under `R21` — it targeted an unreachable guard and is not evidence of a coverage gap. |

</details>

PR #2665, opened 2026-08-21
fix(core): refuse a scoped chat whose project directory is gone

### LINKED ISSUE #2663: fix(core): refuse a scoped chat whose codebase default_cwd is gone
## Summary

A project-scoped chat whose conversation has **no** worktree override (`conversation.cwd === null`) is dispatched into `codebase.default_cwd` with no existence check. When that directory has been deleted — the clone removed, a container volume reset, the folder project moved — the path is handed straight to the provider, which spawns its subprocess there and fails with an error that never names the real problem.

Found by @waaaw while investigating #1170, in PR #2114. Credit for identifying this gap belongs to them wherever it lands.

## Why it is not already covered

PR #2551 fixes the sibling case and is the vehicle for the main fix. Its guard is:

```ts
if (conversation.codebase_id !== null && conversation.cwd !== null) {
```

That covers a stale **worktree override**. It deliberately does not consult `default_cwd`, so the no-override case is untouched at the orchestrator. #2551 does catch it one layer down, but that fix lives in `packages/providers/src/claude/provider.ts` — so **Codex, Pi, OpenCode and Copilot still surface the opaque error**. That is the Codex-in-Docker shape from the original #1170 report.

Verified live on `dev` at `3f78981f`: `orchestrator-agent.ts` still resolves `cwd = conversation.cwd ?? scopedCodebase.default_cwd` with no existence check, and nothing between there and the provider handoff consults the filesystem.

## Root cause of the misleading error

`posix_spawn` reports a missing *working directory* as `ENOENT` against the **executable's** path, not the directory's. Reproduced directly:

```
```

This is why Codex surfaces `No such file or directory (os error 2)` and why the Claude SDK concludes a libc/architecture mismatch. The error names the binary, which is fine, and never names the directory, which is not.

## Placement constraint (read before implementing)

The guard **must run before the user-message persist**, and at that point the codebase row is not loaded. On current `dev`:

- guard site (where #2551 places its check): before the persist block
- `codebaseDb.listCodebases()`: **after** the persist
- `scopedCodebase` derived: ~140 lines later

The persist block's own comment states the invariant: only AI-bound turns get a user row, so a `user` row always pairs with an `assistant` row. `AGENTS.md` documents the same for `remote_agent_messages`.

This is not theoretical. PR #2114 placed its guard at the cwd-resolution site, after the persist. Applying that patch verbatim to current `dev` and measuring it:

```
expect(userRows.length).toBe(1);           // passes
expect(mockSendQuery).toHaveBeenCalled();  // fails: 0 calls
```

An orphaned `user` row on every refusal. Any implementation here must avoid that.

Three ways to get `default_cwd` into scope before the persist, with their costs:

1. **Move the codebase load above the persist.** Cheapest — no extra query, no reordering of anything the persist depends on. Small conflict surface with #2551, which inserts its guard in the same neighbourhood, but trivially reconcilable.
2. **Add a targeted `getCodebase(conversation.codebase_id)` fetch at the guard site.** Self-contained, but pays an extra DB read on every scoped chat turn.
3. **Guard at the resolution site instead.** Rejected — this is exactly what reintroduces the orphaned `user` row measured above.

## Design note: this should compose with #2551, not replace it

The two conditions are disjoint by construction (`cwd !== null` vs `cwd === null`), so the guards complement rather than subsume each other. They also need **different messages**: #2551's advice ("its isolated worktree was removed; run `/worktree remove`") is wrong for this case, where there is no worktree and `/worktree remove` would no-op. The right advice here is re-register the project or switch to another one.

Unifying them into a single `conversation.cwd ?? default_cwd` check is a reasonable later refactor, but it is not this issue, and doing it here would make #2551 redundant.

## Acceptance

- A scoped chat with `cwd === null` against a missing `default_cwd` is refused before the provider is invoked.
- The refusal names the missing directory and gives recovery advice appropriate to a deleted **project root**, not a deleted worktree.
- No orphaned `user` row — regression-tested, not assumed.
- The unscoped path (`ensureArchonWorkspacesPath`, which `mkdir -p`s) stays unguarded.
- Structured log on the refusal, so the state is distinguishable from a provider crash.

## Tracking note

**This is the only tracker for the scoped gap.** #1170 was closed on 2026-07-20 as fixed by `ensureArchonWorkspacesPath`, which addressed the *unscoped* half the reporter hit in Docker. The scoped half was never what that issue tracked, and PR #2114 (which carried a `Closes #1170` link) is now closed. Without this issue the gap is tracked nowhere.

## Scope

- Package: `core`
- Module: `core:orchestrator`
- Depends on #2551 (merge that first; this builds alongside it)

Wirasm:
## Problem and outcome

A registered project's directory can vanish under a long-lived conversation — the clone deleted, a container volume reset, a folder project moved. The chat path resolved `conversation.cwd ?? default_cwd` with no existence check and handed the missing path straight to the provider, which failed with an error that named the wrong thing entirely.

- **Outcome:** the turn is refused before the provider is invoked, naming the missing directory and the two commands that actually recover from it.
- **Invariant:** a refused chat turn never leaves a `user` row without its paired `assistant` row.
- **Scope boundary:** only the no-override case (`conversation.cwd === null`). The stale-worktree-override case belongs to #2551 and is deliberately untouched — see *Why this is not #2551* below.
- **Root cause:** `posix_spawn` reports a missing **working directory** as `ENOENT` against the **executable's** path. Reproduced directly:

  ```
  ```

  That is why Codex surfaced `No such file or directory (os error 2)` and the Claude SDK — which checks only that the binary exists on disk — concluded a libc/architecture mismatch. The binary is fine. Neither error ever named the directory, which is the one fact needed to fix it.

Gap found by @waaaw in #2114 while investigating #1170. Credit for identifying it is theirs.

## Review guidance

- **Feedback requested:** whether the guard's condition is scoped tightly enough to stay disjoint from #2551, and whether the recovery advice is the right pair of commands.
- **Start here:** `packages/core/src/orchestrator/orchestrator-agent.ts` — the guard and the load it depends on. The placement is the load-bearing part, not the check itself.
- **Review order:** the guard and the moved codebase load → the persist block's updated comment → the six tests.
- **Lower-attention areas:** none; the diff is small and hand-written.
- **Known risk or uncertainty:** this touches the same neighbourhood as #2551, so the two conflict textually. Merge instructions are in *Merge order* below — "keep both" is right for the guards and **wrong** for the test fixture, so please read it rather than resolving by instinct.

## Merge order

#2551 first, then this. I merged the two branches locally and ran the combined suite to check the resolution rather than assume it: **247 pass, 0 fail** with both guards present.

- **Production guards — keep both.** They are disjoint (`cwd === null` here, `cwd !== null` there) and both belong above the persist block. Order between them does not matter.
- **Test fixture — keep exactly one.** #2551 adds a byte-identical `const mockExistsSync = mock((_path: string) => true);` to the same file. Keeping both is a duplicate-declaration error. Keep one, and prefer **this PR's top-level `beforeEach`** over #2551's describe-level reset: the describe-level one leaves every later describe running against whatever predicate leaked out of `provider cwd resolution`.
- **Test blocks — keep both.** They cover different cases and do not overlap.

## Solution

**Why the placement is the whole design.** The guard has to run before the user-message persist. That block's own comment states the invariant it protects: only AI-bound turns get a user row, so a `user` row always pairs with an `assistant` row. But the codebase row was not loaded there — `listCodebases()` sat *after* the persist, and `scopedCodebase` was not derived until ~140 lines later.

So the codebase load moves twelve lines up, above the persist. That is free: it reads `remote_agent_codebases` while the persist writes `remote_agent_messages`, the two are independent in both directions, every consumer of `codebases` runs far below, and `discoverAllWorkflows` performs its own `getCodebase()` lookup rather than reading the array. It also removes an existing orphaned-row scenario, since a `listCodebases()` failure now throws before the persist rather than after it.

The guard then refuses rather than falling back the way the neighbouring `scopedCodebase === undefined` branch does. Relocating the agent into a directory the user did not scope would widen its write scope without consent.

**Why this is not #2551.** The two conditions are disjoint by construction — that PR handles `cwd !== null`, this one handles `cwd === null` — so neither subsumes the other, and this PR cannot close that PR's half of the defect. They also need different messages, which is the substantive reason to keep them apart: #2551 tells the user their isolated worktree was removed and to run `/worktree remove`. That advice is false here. With `isolation_env_id` null the command returns *"This conversation is not using a worktree."*, and otherwise it repoints `cwd` at this same missing directory. The repo already documents that trap. This message suggests `/update-project`, which validates the new path and repairs the registration, and `/setproject`, which always works. Tests pin both the suggestions and the omissions so the advice cannot rot into the wrong commands.

Collapsing the two guards into one `conversation.cwd ?? default_cwd` check would silently absorb #2551's case and its message. The condition is commented as load-bearing for that reason.

The project name is quoted in the `/update-project` suggestion. `handleUpdateProject` takes only the first whitespace-separated token as the name, so for a project registered as `Client Ops` an unquoted suggestion parses as name `Client` and hands the user a second, more confusing error instead of a repair — verified against the real parser. Quoting is a no-op for names without whitespace.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A chat message against a deleted project directory returns `No such file or directory (os error 2)` (Codex) or a libc/architecture error naming the Claude binary | The turn is refused, naming the missing directory and the project, with `/update-project` and `/setproject` as recovery |
| **Failure behavior** | Provider spawns into a nonexistent directory; the operator investigates the binary | Fails before the provider, logged as `orchestrator.codebase_cwd_missing` with `conversationId`, `codebaseId`, and `cwd` |

## Validation

- `bun run validate` — **exit code 0** — run as its own command with its real exit code captured, never piped. Proves generated-file checks, type-check, lint at `--max-warnings 0`, format, and the full per-package suite.
- `bun test src/orchestrator/orchestrator-agent.test.ts` (from `packages/core`) — 243 pass / 0 fail — seven new tests covering the refusal, the recovery advice and its omissions, quoting a project name that contains whitespace, the no-orphaned-row invariant, disjointness from #2551, the happy path, and the untouched unscoped path.
- Combined with #2551 (branches merged locally, conflict resolved as described under *Merge order*) — 247 pass / 0 fail.
- **The invariant test was observed failing, not just passing.** With the guard moved below the persist block, `writes no user row when it refuses` failed with `Expected length: 0, Received length: 1` while its sibling assertion `expect(mockSendQuery).not.toHaveBeenCalled()` still passed — isolating the orphaned row and nothing else. Only that one test broke; the other five stayed green. This is the exact defect that made the earlier attempt in #2114 unsuitable, so it is worth having a test that provably catches it.
- Root cause reproduced directly with `child_process.spawnSync` against a deleted directory (transcript above).
- **Not verified:** no end-to-end run against a live Codex install with a genuinely deleted project directory. The refusal is decided at the orchestrator seam and is covered there; everything below it is a pass-through for `sendMessage`.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. No schema change and no new state. A directory that exists behaves exactly as before. | Happy-path test asserts the unchanged `default_cwd` dispatch |
| Rollout / rollback | Fully reversible: one guard, one moved line, one test-fixture change, single commit. | 2 files |
| Observability | A stale project now emits `orchestrator.codebase_cwd_missing` instead of being indistinguishable from a provider crash. | The guard |

### One note for reviewers of this file

The shared `existsSync` mock is now reset in a top-level `beforeEach`. That is required by this change rather than tidying: 21 tests in the file scope a codebase and only 2 set `cwd`, which puts the rest inside this guard's condition. A leaked `false` predicate would refuse those turns and make them assert against the wrong thing, silently. The existing describe-level save/restore in `paused approval gate routing` was left alone — it is now redundant but harmless, and that file is being edited by other PRs.

## Links

- Closes #2663
- Depends on #2551 — merge that first. This complements it and does not substitute for it; landing this alone leaves the stale-worktree-override case open.
- Related #2114 — the closed draft where @waaaw found this gap
- Related #1170 — closed as fixed by `ensureArchonWorkspacesPath`, which covered only the unscoped half

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project recovery commands so project names containing spaces, quotes, or backslashes are handled correctly.
  * Added clearer refusal messaging and warnings when a conversation’s working directory is unavailable.
  * Prevented project-directory recovery advice from appearing in unrelated stale working-directory scenarios.

* **Tests**
  * Updated coverage to verify user-facing messages and warnings for missing conversation directories.

### DIFF
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -1661,10 +1661,24 @@ describe('provider cwd resolution', () => {
       const platform = makePlatform();
       await handleMessage(platform, 'conv-1', 'hello');
 
-      // This guard is deliberately scoped to `cwd === null`. The override case is
-      // a different situation with different advice and belongs to #2551; if this
-      // ever starts firing here, the two guards have been collapsed into one.
-      expect(getSendQueryCwd()).toBe('/worktrees/removed');
+      // This guard is deliberately scoped to `cwd === null`; the override case is a
+      // different situation with different advice and belongs to #2551.
+      //
+      // Assert that THIS guard stays silent rather than asserting what the turn
+      // does next. What happens next is not stable across the two PRs: today the
+      // override reaches the provider, and once #2551's guard lands beside this
+      // one it refuses the turn instead. Either way the invariant that matters
+      // here is the same — this guard did not claim the case.
+      expect(mockLogger.warn).not.toHaveBeenCalledWith(
+        expect.anything(),
+        'orchestrator.codebase_cwd_missing'
+      );
+      const sends = (platform.sendMessage as ReturnType<typeof mock>).mock.calls.map(c => c[1]);
+      expect(
+        sends.some(
+          (s: unknown) => typeof s === 'string' && s.includes('project directory no longer exists')
+        )
+      ).toBe(false);
     });
 
     test('does not fire when the project directory is present', async () => {
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -1581,6 +1581,118 @@ describe('provider cwd resolution', () => {
     expect(mockSendQuery).toHaveBeenCalled();
   });
 
+  // ─── missing project directory (#2663) ──────────────────────────────────────
+
+  describe('missing project directory', () => {
+    test('refuses the turn and never reaches the provider when default_cwd is gone', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      // Handing a missing path to the provider is the whole defect: the spawn
+      // fails ENOENT against the BINARY, so the user is told the wrong thing.
+      expect(mockSendQuery).not.toHaveBeenCalled();
+      const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
+      expect(sent).toContain('/repos/test-repo');
+      expect(mockLogger.warn).toHaveBeenCalledWith(
+        expect.objectContaining({ codebaseId: 'codebase-1', cwd: '/repos/test-repo' }),
+        'orchestrator.codebase_cwd_missing'
+      );
+    });
+
+    test('offers recovery that actually works, and not the traps', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
+      // /update-project validates the new path and repairs the registration.
+      expect(sent).toContain('/update-project');
+      expect(sent).toContain('/setproject');
+      // /worktree remove answers "not using a worktree" when isolation_env_id is
+      // null, and otherwise repoints cwd at this same missing directory.
+      expect(sent).not.toContain('/worktree remove');
+      // /register-project creates a new registration rather than repairing this one.
+      expect(sent).not.toContain('/register-project');
+    });
+
+    test('writes no user row when it refuses, so none is left unpaired', async () => {
+      mockAddMessage.mockClear();
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      // Non-web only: the web adapter's route persists its own turns, so the
+      // orchestrator never writes a user row for it and this could not regress.
+      const platform = makePlatform();
+      platform.getPlatformType = mock(() => 'telegram') as typeof platform.getPlatformType;
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      expect(mockSendQuery).not.toHaveBeenCalled();
+      expect(mockAddMessage.mock.calls.filter(c => c[1] === 'user')).toHaveLength(0);
+    });
+
+    test('leaves a stale cwd override to the conversation-cwd guard', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({
+        codebase_id: 'codebase-1',
+        cwd: '/worktrees/removed',
+      });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      mockExistsSync.mockImplementation((p: string) => p !== '/worktrees/removed');
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      // This guard is deliberately scoped to `cwd === null`. The override case is
+      // a different situation with different advice and belongs to #2551; if this
+      // ever starts firing here, the two guards have been collapsed into one.
+      expect(getSendQueryCwd()).toBe('/worktrees/removed');
+    });
+
+    test('does not fire when the project directory is present', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      expect(getSendQueryCwd()).toBe('/repos/test-repo');
+    });
+
+    test('leaves the unscoped path alone — the workspaces root is created on demand', async () => {
+      const conversation = makeConversation({ codebase_id: null });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([]));
+      mockExistsSync.mockImplementation(() => false);
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      expect(getSendQueryCwd()).toBe('/home/test/.archon/workspaces');
+    });
+  });
+
   test('unscoped chat uses ensureArchonWorkspacesPath result', async () => {
     const conversation = makeConversation({ codebase_id: null });
     mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -373,9 +373,10 @@ function makeCodebase(name: string, id = `id-${name}`): Codebase {
 // handleUpdateProject. Today a leak is survivable only because the predicates
 // here reject one literal path — that is luck, not a guarantee, and it stops
 // being true the moment a test rejects something broader. Reset before every
+// test so no describe can poison another.
+//
 // Deliberately no count here: any number rots on the next test added, and the
 // mechanism is the argument.
-// test so no describe can poison another.
 beforeEach(() => {
   mockExistsSync.mockImplementation(() => true);
 });
@@ -1720,11 +1721,44 @@ describe('provider cwd resolution', () => {
       const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
       expect(sent).toContain('/worktrees/removed');
       expect(sent).toContain('working directory no longer exists');
+      // This guard's own message signature — nothing else emits that sentence, so
+      // it cannot false-alarm. Deliberately NOT also asserting the absence of
+      // `/update-project`: the guard above offers `/worktree remove` and
+      // `/setproject` and never that, so the two strings always co-occur and it
+      // would add no detection — while breaking this test if anyone ever adds
+      // `/update-project` to that message, an edit that has nothing to do with
+      // this guard.
       expect(sent).not.toContain('project directory no longer exists');
-      expect(sent).not.toContain('/update-project');
       expect(mockLogger.warn).toHaveBeenCalledWith(
         expect.anything(),
         'orchestrator.conversation_cwd_missing'
+      );
+    });
+
+    test('runs the turn when the cwd override is healthy but default_cwd is gone', async () => {
+      const codebase = makeCodebaseForSync();
+      const conversation = makeConversation({
+        codebase_id: 'codebase-1',
+        cwd: '/worktrees/healthy',
+      });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      // The project root is gone, but this conversation does not use it.
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      // This is the case that makes `&& conversation.cwd === null` load-bearing
+      // rather than decorative. Drop that clause while leaving the target as
+      // `default_cwd` and this guard refuses a perfectly healthy turn, telling the
+      // user their project directory is gone and offering to repoint a directory
+      // the turn never touches. Nothing else in the suite catches that mutation.
+      expect(getSendQueryCwd()).toBe('/worktrees/healthy');
+      expect(mockLogger.warn).not.toHaveBeenCalledWith(
+        expect.anything(),
+        'orchestrator.codebase_cwd_missing'
       );
     });
 
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -1670,7 +1670,7 @@ describe('provider cwd resolution', () => {
       expect(mockAddMessage.mock.calls.filter(c => c[1] === 'user')).toHaveLength(0);
     });
 
-    test('leaves a stale cwd override to the conversation-cwd guard', async () => {
+    test('a stale cwd override gets the conversation-cwd message, not this one', async () => {
       const codebase = makeCodebaseForSync();
       const conversation = makeConversation({
         codebase_id: 'codebase-1',
@@ -1684,25 +1684,27 @@ describe('provider cwd resolution', () => {
       const platform = makePlatform();
       await handleMessage(platform, 'conv-1', 'hello');
 
-      // This guard is deliberately scoped to `cwd === null`; the override case is a
-      // different situation with different advice and belongs to #2551.
+      // Pins WHICH refusal the user gets, which is the part still observable here.
       //
-      // Assert that THIS guard stays silent rather than asserting what the turn
-      // does next. Since #2551 landed, the conversation-cwd guard sitting above
-      // this one refuses the override case, so an assertion about reaching the
-      // provider would be asserting that guard's behaviour, not this one's. The
-      // durable invariant is narrower and is what is checked here: this guard did
-      // not claim the case.
-      expect(mockLogger.warn).not.toHaveBeenCalledWith(
+      // Be clear about what this does not do: it cannot catch widening this guard's
+      // condition to `conversation.cwd ?? default_cwd`. The conversation-cwd guard
+      // above returns first for every `cwd !== null` case, so a widened condition
+      // is dead code, and the mutation passes the whole suite. That is a property
+      // of the ordering, not a gap worth papering over with a weaker assertion —
+      // the note on the guard itself carries the reason to keep it narrow.
+      //
+      // What this DOES catch is the two guards' messages crossing: this guard
+      // claiming the override case and answering with project-root advice that says
+      // nothing about `isolation_env_id`.
+      const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
+      expect(sent).toContain('/worktrees/removed');
+      expect(sent).toContain('working directory no longer exists');
+      expect(sent).not.toContain('project directory no longer exists');
+      expect(sent).not.toContain('/update-project');
+      expect(mockLogger.warn).toHaveBeenCalledWith(
         expect.anything(),
-        'orchestrator.codebase_cwd_missing'
+        'orchestrator.conversation_cwd_missing'
       );
-      const sends = (platform.sendMessage as ReturnType<typeof mock>).mock.calls.map(c => c[1]);
-      expect(
-        sends.some(
-          (s: unknown) => typeof s === 'string' && s.includes('project directory no longer exists')
-        )
-      ).toBe(false);
     });
 
     test('does not fire when the project directory is present', async () => {
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -1651,6 +1651,27 @@ describe('provider cwd resolution', () => {
       expect(sent).not.toContain('/update-project Client Ops');
     });
 
+    test('escapes quotes and backslashes in the project name', async () => {
+      // Quoting alone is not enough: a `"` inside the name closes the quoted token
+      // early and reproduces the original defect, and a trailing `\` escapes the
+      // closing quote. parseCommand honours backslash escapes inside quotes, so a
+      // single pass over both characters round-trips. One pass, not two chained
+      // replaces — escaping `"` first and `\` second would double-escape the
+      // backslashes the first pass just added.
+      const codebase = { ...makeCodebaseForSync(), name: 'Bob"s \\Ops' };
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
+      expect(sent).toContain('/update-project "Bob\\"s \\\\Ops" <new-path>');
+    });
+
     test('writes no user row when it refuses, so none is left unpaired', async () => {
       mockAddMessage.mockClear();
       const codebase = makeCodebaseForSync();
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -1688,10 +1688,11 @@ describe('provider cwd resolution', () => {
       // different situation with different advice and belongs to #2551.
       //
       // Assert that THIS guard stays silent rather than asserting what the turn
-      // does next. What happens next is not stable across the two PRs: today the
-      // override reaches the provider, and once #2551's guard lands beside this
-      // one it refuses the turn instead. Either way the invariant that matters
-      // here is the same — this guard did not claim the case.
+      // does next. Since #2551 landed, the conversation-cwd guard sitting above
+      // this one refuses the override case, so an assertion about reaching the
+      // provider would be asserting that guard's behaviour, not this one's. The
+      // durable invariant is narrower and is what is checked here: this guard did
+      // not claim the case.
       expect(mockLogger.warn).not.toHaveBeenCalledWith(
         expect.anything(),
         'orchestrator.codebase_cwd_missing'
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -373,6 +373,8 @@ function makeCodebase(name: string, id = `id-${name}`): Codebase {
 // handleUpdateProject. Today a leak is survivable only because the predicates
 // here reject one literal path — that is luck, not a guarantee, and it stops
 // being true the moment a test rejects something broader. Reset before every
+// Deliberately no count here: any number rots on the next test added, and the
+// mechanism is the argument.
 // test so no describe can poison another.
 beforeEach(() => {
   mockExistsSync.mockImplementation(() => true);
@@ -1626,6 +1628,27 @@ describe('provider cwd resolution', () => {
       expect(sent).not.toContain('/worktree remove');
       // /register-project creates a new registration rather than repairing this one.
       expect(sent).not.toContain('/register-project');
+    });
+
+    test('quotes a project name containing whitespace so the suggestion parses', async () => {
+      const codebase = { ...makeCodebaseForSync(), name: 'Client Ops' };
+      const conversation = makeConversation({ codebase_id: 'codebase-1' });
+      mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
+      mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
+      mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
+      mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');
+
+      const platform = makePlatform();
+      await handleMessage(platform, 'conv-1', 'hello');
+
+      const sent = (platform.sendMessage as ReturnType<typeof mock>).mock.calls[0][1] as string;
+      // handleUpdateProject takes only the FIRST token as the project name, so an
+      // unquoted `Client Ops` parses as `Client` and hands the user a second, more
+      // confusing error instead of a repair. Verified against the real parser:
+      // `/update-project Client Ops <path>` -> name "Client", path "Ops <path>";
+      // the quoted form -> name "Client Ops", path "<path>".
+      expect(sent).toContain('/update-project "Client Ops" <new-path>');
+      expect(sent).not.toContain('/update-project Client Ops');
     });
 
     test('writes no user row when it refuses, so none is left unpaired', async () => {
--- a/packages/core/src/orchestrator/orchestrator-agent.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.ts
@@ -1640,14 +1640,20 @@ export async function handleMessage(
     // the same field with no mutation between the reads, which is what lets them sit
     // side by side.
     //
-    // Keep the condition narrow rather than widening it to `conversation.cwd ??
-    // default_cwd`. Be aware that no test can hold you to this: the guard above
-    // already returns for every `cwd !== null` case, so a widened condition here is
-    // dead code rather than a behaviour change, and it is unobservable through
-    // handleMessage. The reason to keep it narrow is that it would silently become
-    // observable the moment the guard above is moved, narrowed, or removed — at
-    // which point this guard would answer for a stale worktree with the wrong
-    // message, one that says nothing about `isolation_env_id`.
+    // `&& conversation.cwd === null` is load-bearing. Dropping it while leaving the
+    // target as `default_cwd` refuses a healthy turn: a conversation working in a
+    // live worktree, whose project root happens to be gone, would be told its
+    // project directory no longer exists and offered `/update-project` for a
+    // directory that turn never touches. Covered by `runs the turn when the cwd
+    // override is healthy but default_cwd is gone`.
+    //
+    // Widening it the other way — to `conversation.cwd ?? default_cwd`, moving the
+    // target with the condition — is instead unobservable here, because the guard
+    // above already returns for every `cwd !== null` case. No test holds you to
+    // that one. Keep it narrow anyway: it becomes observable the moment the guard
+    // above is moved, narrowed, or removed, at which point this guard would answer
+    // for a stale worktree with the wrong message, one that says nothing about
+    // `isolation_env_id`.
     //
     // Refuses rather than falling back to the workspaces root the way the
     // `scopedCodebase === undefined` branch below does: relocating the agent into a
@@ -1684,11 +1690,12 @@ export async function handleMessage(
     // Persist the inbound user message for non-web platforms (Slack/Telegram/
     // GitHub/Discord/CLI) — the web adapter's route persists web turns itself.
     // Placed AFTER every early return that declines the turn — deterministic
-    // commands (including `/workflow approve|reject`) and the missing-project guard
-    // above — so only AI-bound turns get a user row (no orphaned user message
-    // without an assistant reply), and BEFORE the AI call so the user row's
-    // timestamp precedes the assistant row's. A plain message at a gate is NOT a
-    // refusal since #2577; it flows into the AI turn and earns its user row.
+    // commands (including `/workflow approve|reject`), the stale-worktree guard and
+    // the missing-project guard above — so only AI-bound turns get a user row (no
+    // orphaned user message without an assistant reply), and BEFORE the AI call so
+    // the user row's timestamp precedes the assistant row's. A plain message at a
+    // gate is NOT a refusal since #2577; it flows into the AI turn and earns its
+    // user row.
     //
     // A new refusal that needs nothing computed below belongs above this block. One
     // that needs data from further down (workflow discovery, gate lookup) has to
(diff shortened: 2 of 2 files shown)


Wirasm:
# PR Review: #2665 — fix(core): refuse a scoped chat whose project directory is gone

## Signal

**A fourth round, and the newest commit introduced the blocker.** `95a3bba3` fixed R9 and R10, both
genuinely — I re-ran the mutations rather than reading the diff. But the comment it added to justify
the R9 fix states something false, in the direction that grants permission. It tells the next
maintainer that widening this guard's condition "is dead code rather than a behaviour change, and it
is unobservable through `handleMessage`" (**R13**). That is true of the one widening the previous
round tested — the `conversation.cwd ?? default_cwd` collapse — and false for the more natural one:
deleting the `&& conversation.cwd === null` clause outright.

Measured, not argued. Dropping that clause passes the whole suite **248 pass / 0 fail**, and it is
plainly observable: a conversation working in a healthy worktree whose registered project root has
been deleted is **refused**, and told its project directory is gone and to run `/update-project` on a
directory that turn never touches. I wrote the discriminating probe — shipped code **249/0**, widened
guard **248/1**, failing on exactly that probe.

So after three rounds, the single clause this PR spends six comment lines, a plan risk row and a
dedicated test defending has **no** regression protection, and the tree now carries prose telling a
reader it does not need any. That is a worse position than silence. One test closes it; the exact
test is in R13.

Second, a rebase artifact no round has caught (**R14**). Commit `363c2a28` deleted test-fixture counts
that the rebase onto #2551 had already removed, so only the justification landed — spliced into the
middle of a sentence #2551 wrote on `dev`, cutting "Reset before every … test so no describe can
poison another" in half and leaving a rationale for a deletion that is nowhere visible.

Everything the brief asked to be verified independently holds:

- **The rebase resolution is correct.** Merge-base is exactly `origin/dev` = `b4599464`; no `dev`
  commit is missing from the branch. Both guards present and disjoint (`cwd !== null` at `:1584`,
  `cwd === null` at `:1655`, `conversation` last reassigned at `:1483`, nothing between the reads
  touches it). Exactly one `mockExistsSync` declaration and exactly one top-level `beforeEach` —
  though both arrived from #2551 on `dev`, not from this PR, which makes the PR body's *Merge order*
  instructions moot history rather than guidance.
- **The test arithmetic checks out.** I ran the baseline myself: `b4599464` → **241 pass / 0 fail**;
  head → **248 pass / 0 fail**; the diff adds exactly 7 tests. 241 + 7 = 248.
- **`bun run validate` exits 0** at both `0ddf864f` and `95a3bba3`, each run as its own command with
  its real exit code captured to a file, never piped.
- **R1 and R4 re-verified from source rather than from the disposition.** The suggestion at `:1667`
  is quoted and now escaped; `git log -S"await approveWorkflow"` → `7752ec53` (#2577) confirms the
  approval branch is gone.

**Updated to `de489421`.** The head moved twice while this round was in flight. `de489421`
("test(core): pin the project-name escaping, not just the quoting") closes **R12** and nothing else —
verified rather than read: stripping only the `.replace(/[\\"]/g, …)` while keeping the quotes now
turns exactly one test red (`escapes quotes and backslashes in the project name`), **248 pass /
1 fail**, where the same mutation gave 248/0 one commit earlier. The `Bob"s \Ops` fixture is the
right choice — it exercises both characters in one string and so also pins the single-pass regex
against the double-escaping a chained pair of replaces would introduce.

**R13 through R16 are untouched by it**, and R13 was re-confirmed at `de489421` on an otherwise
pristine tree: dropping `&& conversation.cwd === null` gives **249 pass / 0 fail**. The blocker
stands.

- **Blocking:** 1 (R13)
- **Non-blocking:** 3 (R14, R15, R16)
- **Resolved this round:** 1 (R12, at `de489421`, verified by mutation)
- **Validation:** `bun run validate` exit 0 at `95a3bba3`; focused suite 249 pass / 0 fail at
  `de489421`, 248 at `95a3bba3`, dev baseline 241; four mutations executed this round

## Findings

|---|---|---|---|
| `R1` | Important | The suggested `/update-project` command mis-parses when the project name contains whitespace | FIXED |
| `R2` | Suggestion | The message named "a reset container volume" as a cause that cannot produce this state | FIXED |
| `R3` | Suggestion | Test-fixture comment's hard-coded counts were wrong by 3-7x | FIXED |
| `R4` | Suggestion | Persist-block comment listed "approvals" as an early return that #2577 deleted | FIXED |
| `R5` | Suggestion | Comment asserted a #2551 guard that was not in the tree | FIXED |
| `R6` | Suggestion | #2551 adds a byte-identical `mockExistsSync`; "keep both" was wrong for the test file | FIXED |
| `R7` | Suggestion | "every consumer of `codebases` runs far below" had a counterexample 18 lines down | FIXED |
| `R8` | Suggestion | "Any new refusal belongs above this block" omitted its own precondition | FIXED |
| `R9` | Important | After the rebase, the disjointness test could not reach this guard, so the mutation it existed to catch passed | FIXED |
| `R10` | Suggestion | The quote-wrap did not escape `"` or `\` in a project name, reproducing R1 for those names | FIXED |
| `R11` | Suggestion | The comment forbidding a collapse named a consequence the guard ordering made impossible | FIXED |
| `R12` | Suggestion | The new `"`/`\` escaping has no regression test; removing it alone keeps the suite green | FIXED |
| `R13` | Important | The comment added by `95a3bba3` says widening the condition is unobservable; deleting the `cwd === null` clause is a live regression no test catches | FIXED |
| `R14` | Suggestion | A rebase artifact splices two lines into the middle of a sentence #2551 wrote, justifying a deletion that never happened | FIXED |
| `R15` | Suggestion | The persist block claims to list "every early return that declines the turn" and omits #2551's guard at `:1613` | FIXED |
| `R16` | Suggestion | The two sibling guards log the same `conversationId` key with two different id spaces (DB UUID vs platform id) | OPEN |

## Detailed Findings

<details>
<summary><code>R13</code> — the new comment says widening the condition is unobservable; the natural widening is a live, untested regression</summary>

**Impact:** `95a3bba3` added prose whose stated purpose is to tell a reader what the tests do *not*
cover. It gets that wrong in the permissive direction, and the clause it discusses has no test at
all. A maintainer who reads the comment, concludes the guard above already handles every
`cwd !== null` case, and deletes the now-apparently-redundant clause ships a regression that CI
cannot see: a chat turn running in a healthy worktree is refused, and the refusal names a directory
that turn never uses and offers a command that repairs the wrong thing.

**The false clause** — `packages/core/src/orchestrator/orchestrator-agent.ts:1643-1646`:

> Be aware that no test can hold you to this: the guard above already returns for every
> `cwd !== null` case, so a widened condition here is **dead code rather than a behaviour change**,
> and it is **unobservable through handleMessage**.

The guard above returns for `cwd !== null` **and `!existsSync(cwd)`**. It falls through for
`cwd !== null && existsSync(cwd)` — a live worktree — and that is precisely where a widened
condition here becomes observable.

**Evidence — executed at `95a3bba3` in a scratch worktree, not inferred:**

| Mutation | Result |
|---|---|
| Baseline, unmodified | 248 pass / 0 fail |
| Drop only `&& conversation.cwd === null` (guard becomes `if (conversation.codebase_id !== null) {`) | **248 pass / 0 fail — not caught** |
| Same mutation, with the probe test below added | 248 pass / **1 fail** — the probe, `Received number of calls: 0` |
| Unmodified guard, with the probe test added | **249 pass / 0 fail** |

The probe fails under the mutation because `sendQuery` is never reached: the widened guard refuses
the turn outright.

**Why this is a real user-visible state, not a contrived one.** Worktrees live at
`~/.archon/workspaces/<owner>/<repo>/worktrees/…` and the registered project root at `…/source/`.
Deleting or moving the source clone while a conversation is bound to a surviving worktree leaves
exactly this shape: `conversation.cwd` exists, `codebase.default_cwd` does not. The turn should run —
the work happens in the worktree, and the missing project root is irrelevant to it.

**Required outcome — two parts:**

1. **Narrow the comment to the claim that is actually true.** The disclaimer holds for the
   `conversation.cwd ?? default_cwd` collapse specifically, which was measured; it does not hold for
   "a widened condition" in general. Say the former.
2. **Add the probe**, which is the missing protection for the clause. Verified green on the shipped
   code and red under the clause deletion and nothing else:

```ts
test('a live worktree override still runs when the project root is gone', async () => {
  const codebase = makeCodebaseForSync();
  const conversation = makeConversation({
    codebase_id: 'codebase-1',
    cwd: '/worktrees/feature',
  });
  mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
  mockGetCodebase.mockReturnValueOnce(Promise.resolve(codebase));
  mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
  // The worktree is healthy; only the registered project root is gone.
  mockExistsSync.mockImplementation((p: string) => p !== '/repos/test-repo');

  const platform = makePlatform();
  await handleMessage(platform, 'conv-1', 'hello');

  expect(getSendQueryCwd()).toBe('/worktrees/feature');
});
```

This is the positive counterpart to `a stale cwd override gets the conversation-cwd message, not this
one`: that test covers the sub-case where the override is *also* missing (#2551's guard fires); this
one covers the sub-case where the override is healthy, which is the only state in which this guard's
scoping decision is observable at all. Together they close the condition.

**Found by:** `prp-core:pr-test-analyzer` (identified the gap and the mutation), review synthesis
(executed the clause-deletion mutation against `95a3bba3` and wrote the probe)

**Disposition:** OPEN. Blocking. Not because the shipped behavior is wrong — it is correct — but
because the PR's own load-bearing invariant is unprotected while the tree tells a reader it needs no
protection. R11 was closed on evidence that covered a different mutation than the one the comment's
wording now licenses.

</details>

<details>
<summary><code>R14</code> — a rebase artifact splits a sentence and justifies a deletion that never happened</summary>

**Impact:** The one operational instruction in the shared-fixture comment — the sentence telling a
maintainer not to relocate the `beforeEach` into a `describe` — is cut in half by an unrelated note,
and the note argues for removing a count that appears nowhere in the file. A reader tidying the
paragraph cannot tell which half is residue and may delete the instruction rather than the orphan.

**Current text** — `packages/core/src/orchestrator/orchestrator-agent.test.ts:375-378`:

```
// being true the moment a test rejects something broader. Reset before every
// Deliberately no count here: any number rots on the next test added, and the
// mechanism is the argument.
// test so no describe can poison another.
```

**Evidence:**

- `git show b4599464:packages/core/src/orchestrator/orchestrator-agent.test.ts` — the base version of
  this paragraph, authored by #2551, contains no count of any kind and ends cleanly: "…Reset before
  every test so no describe can poison another."
- Commit `363c2a28`'s own message: *"Delete the hard-coded test counts from the fixture comment…
  any count there rots on the next test added. The mechanism is the argument."* The rebase onto
  `b4599464` had already replaced that paragraph with #2551's count-free version, so the deletion was
  a no-op and only the justification landed.
- `git diff origin/dev...HEAD` on the test file shows exactly `+2` lines at this location — an
  insertion into someone else's sentence, not a rewrite.
- The nearest numeral in the surviving paragraph is "one literal path", a count of paths, not tests.

**Required outcome:** Delete the two inserted lines. The paragraph then matches `dev` byte for byte
and the sentence is whole. If the "no count" rationale is judged worth keeping, it belongs after the
complete sentence, not inside it.

**Found by:** `prp-core:comment-analyzer` (rated Important), `prp-core:code-reviewer` (noted
independently)

**Disposition:** OPEN. Recorded as a Suggestion here rather than Important — the consequence is
maintenance-only and nothing reachable breaks — while noting the comment analyzer argued for
Important on the grounds that it is a defect this PR introduced relative to `dev` with no upside.
Nothing turns on the difference: R13 already sets the verdict, and this is a two-line delete that
should ride along in the same edit.

</details>

<details>
<summary><code>R15</code> — the persist block claims an exhaustive list of decline paths and omits #2551's guard</summary>

**Impact:** The comment exists to support exactly one audit — which paths can refuse a turn before
the `user` row is written — and its inventory omits the refusal most similar to the one it does name,
sitting 67 lines above it. This is the comment `0ddf864f` was written to reconcile with #2551 having
landed, and #2551's guard is the one thing it does not mention.

**Evidence:**

- `packages/core/src/orchestrator/orchestrator-agent.ts:1686-1688` — "Placed AFTER every early return
  that declines the turn — deterministic commands (including `/workflow approve|reject`) and the
  missing-project guard above".
- Enumerated every `return;` in `handleMessage` before the persist at `:1684`: `:1512`, `:1519`,
  `:1526`, `:1536`, `:1563` (the deterministic-command branch), **`:1613` (#2551's stale-`cwd`
  guard)**, and `:1680` (this PR's guard). Seven; the comment accounts for six.
- `:1613`'s branch sends its own refusal ("This conversation's working directory no longer exists")
  and returns, so it is unambiguously an early return that declines the turn.

**Required outcome:** "— deterministic commands (including `/workflow approve|reject`), the
stale-`cwd`-override guard (#2551), and the missing-project guard above —".

**Disposition:** OPEN. Suggestion.

</details>

<details>
<summary><code>R16</code> — the sibling guards log one key with two id spaces</summary>

**Impact:** An operator investigating a stale-directory report cannot correlate the two refusal
events, because `conversationId` means a different thing in each. They are adjacent siblings
refusing for related reasons, which is exactly when correlation matters.

**Evidence:**

- `orchestrator-agent.ts:1588` (#2551's guard) — `conversationId: conversation.id`, the database row
  UUID.
- `orchestrator-agent.ts:1665` (this PR's guard) — bare `conversationId`, the `handleMessage`
  parameter, which is the platform-supplied id (`thread_ts`, `chat_id`, `owner/repo#n`). The two are
  never equal: the row is fetched via `db.getOrCreateConversation(platformType, conversationId)`.

**Direction — corrected.** My first statement of this finding said to change *this PR's* line to
match its sibling. That is backwards, and the author was right to challenge it before acting.
Counted across the file: ~35 log sites use the bare platform id, 4 use `conversation.id`. Splitting
those 4 by whether the parameter was even in scope:

- `handleSetProject` (`:3067`, `:3073`) — signature is `(message, conversation)`, no platform id
  available. Outliers **by necessity**, not deviations.
- `inheritThreadContext` (`:1290`) and #2551's guard (`:1588`) — both have `conversationId: string`
  in scope and chose the row id. Outliers **by choice**, and there are only two.

So this PR's line follows the dominant convention and #2551's guard is the deviation. Changing this
PR's line would trade a local inconsistency for a file-wide one.

**Required outcome:** Nothing in this PR. The one-line change belongs on #2551's guard on `dev`
(`conversation.id` → `conversationId`), which is merged code and outside this PR's scope — a
separate commit, not a rider. Worth noting a third option neither reading covered: the two
`handleSetProject` sites cannot follow the convention at all, so if anyone wants full consistency
the honest fix there is a distinct key (`conversationRowId`), not a changed value.

**Found by:** `review-2665-rebased`; direction corrected by the author's count and re-verified here

**Disposition:** OPEN, and deliberately not actionable within this PR. Recorded so the operator can
decide whether the follow-up on `dev` is worth a commit.

</details>

<details>
<summary><code>R12</code> — the escaping is unprotected by any test</summary>

**Impact:** R10's fix can be removed without any signal. The escape is the only thing standing
between a project named `Say "hi"` and the exact mis-parse this PR exists to prevent, and nothing
holds it in place.

**Evidence:**

- `packages/core/src/orchestrator/orchestrator-agent.ts:1667` — `` `/update-project "${scoped.name.replace(/[\\"]/g, c => `\\${c}`)}" <new-path>` ``.
- **Mutation executed** at `95a3bba3`: removed only the `.replace(/[\\"]/g, …)` and kept the quotes
  → `bun test src/orchestrator/orchestrator-agent.test.ts` → **248 pass / 0 fail**. Not caught.
- Contrast, removing the quotes *and* the escape → 247 pass / 1 fail
  (`quotes a project name containing whitespace so the suggestion parses`). So the existing test
  pins the quoting only; its fixture name is `Client Ops`, which quotes alone already handle.

**Required outcome:** Extend the existing test with a name containing a `"` — a second `codebase`
named e.g. `Say "hi"` asserting the emitted string contains `/update-project "Say \"hi\""` — or add
one case to it. The verification the author already ran by hand against the real parser is the right
check; it just is not committed.

**Found by:** review synthesis (mutation executed during re-verification)

**Disposition:** FIXED at `de489421`. `orchestrator-agent.test.ts` gains
`escapes quotes and backslashes in the project name` with the combined fixture `Bob"s \Ops`,
asserting the emitted `/update-project "Bob\"s \\Ops" <new-path>`. Verified by re-running the
original mutation: removing only the `.replace(/[\\"]/g, …)` and keeping the quotes now yields
**248 pass / 1 fail**, failing on exactly that test, against 248 / 0 before the commit. The combined
fixture also covers the double-escaping hazard, since `\` and `"` in one name only round-trip under
a single pass.

</details>

<details>
<summary><code>R9</code> — the disjointness test could not reach this guard after the rebase</summary>

**Impact:** The test added to pin this PR's central design claim asserted a silence that nothing
could break, because #2551's guard returned first for every fixture it used.

**Evidence of the fix:** `packages/core/src/orchestrator/orchestrator-agent.test.ts:1673-1707` —
renamed to `a stale cwd override gets the conversation-cwd message, not this one`, and rewritten to
assert which refusal reaches the user: the message contains `/worktrees/removed` and
`working directory no longer exists`, does not contain `project directory no longer exists` or
`/update-project`, and `orchestrator.conversation_cwd_missing` was logged.

**Disposition:** FIXED, verified in both directions rather than by reading the diff.

*It can fail.* Disabling #2551's guard (`if (false && …)`) turns it red on a real assertion —
`expect(sent).toContain('/worktrees/removed')`, received `"test response"` — so it fails against an
actual value, not by crashing on an absent mock call. Five tests fail under that mutation: this one
plus #2551's four.

*It carries its own weight, on a different mutation.* Under the disable-#2551 mutation it adds
nothing #2551's own tests do not already catch. Where it earns its place is the realistic
"these two guards are redundant, merge them" refactor — remove the guard above **and** widen this
one to the effective cwd. Executed: the test fails on
`toContain('working directory no longer exists')` because the user now receives the project-root
message naming `/worktrees/removed` and offering `/update-project`. That is exactly the
messages-crossing scenario the comment names, and #2551's tests are not written to describe it from
this side.

*Its disclaimer is true.* The comment says it cannot catch widening the condition to
`conversation.cwd ?? default_cwd`. Executed: that mutation yields **248 pass / 0 fail**. The prose
matches the mutation result, which is the property that was missing before.

*How much weight it carries, since the answer is not flattering.* It never fails alone. Across
four mutations — disabling #2551's guard, merging the two, widening-and-reordering, and narrowing
#2551's guard with `&& isolation_env_id !== null` — it always fails alongside at least one of
#2551's tests, so it adds no detection those tests do not already provide. It is kept anyway on an
ownership argument rather than a detection one: #2551's tests live in #2551's block and can be
rewritten by someone working on that guard for unrelated reasons, and this is the assertion that
records the invariant at *this* guard's site. Redundancy across an independently-editable boundary
is the kind that pays.

The fourth mutation in that list corrects a claim I made to the author: I asserted the
narrowing-from-above case was caught by #2551's tests and not this one. Executed at `de489421`, it
is 246 pass / 3 fail and the third failure is this test. The set of mutations it participates in is
wider than I had established.

**Found by:** `prp-core:pr-test-analyzer`, review synthesis

</details>

<details>
<summary><code>R10</code> — the quote-wrap did not escape <code>"</code> or <code>\</code></summary>

**Evidence of the fix:** `orchestrator-agent.ts:1667` now applies
`scoped.name.replace(/[\\"]/g, c => `\\${c}`)` inside the quotes.

**Disposition:** FIXED. Nine names round-tripped through the real `parseCommand` and the
`handleUpdateProject` destructure, before and after:

```
name                  quote-only   quote+escape
my-app                OK           OK
Client Ops            OK           OK
O'Brien Ops           OK           OK
"  Padded  "          OK           OK
Bob"s Project         FAIL         OK
Alice\Bob             FAIL         OK
Say "hi"              FAIL         OK
back\slash            FAIL         OK
quote"and\slash       FAIL         OK
```

No regression in the four that already worked — `O'Brien Ops` matters because `parseCommand` treats
`'` as a quote character too, and an escape that over-reached would have broken it. The combined
`quote"and\slash` row is the one that proves the single-pass `[\\"]` character class does not
double-escape. Protection for this fix is R12.

</details>

<details>
<summary><code>R11</code> — the collapse warning named a consequence the ordering made impossible</summary>

**Evidence of the fix:** `orchestrator-agent.ts:1638-1651` no longer claims a collapse replaces
#2551's message. It now states that the guard above already returns for every `cwd !== null` case,
that a widened condition here is therefore dead code and unobservable through `handleMessage`, and
that the reason to keep it narrow is that it becomes observable the moment the guard above is moved,
narrowed, or removed — at which point this guard would answer a stale worktree with advice that says
nothing about `isolation_env_id`.

**Disposition:** FIXED — the original defect is gone; the comment no longer claims a collapse
replaces #2551's message.

Correcting my own record rather than leaving it to be inferred: I closed this on a single mutation,
the `conversation.cwd ?? default_cwd` collapse, which really is unobservable (248 pass / 0 fail).
The replacement wording then generalises from that one case to *all* widening, and that
generalisation is false — see R13, which is a direct consequence of closing this finding on
evidence narrower than the claim it was used to endorse. Read R11 as "the impossible-consequence
sentence was removed", not as "the replacement sentence was verified".

**Found by:** review synthesis

</details>

<details>
<summary><code>R1</code> — <code>/update-project</code> mis-parsed for a name containing whitespace</summary>

**Disposition:** FIXED. Re-verified at `95a3bba3`: stripping the quoting turns exactly one test red
(`quotes a project name containing whitespace so the suggestion parses`), 247 pass / 1 fail, and
`Client Ops` round-trips through the real parser. Residual quote/backslash case was R10, now also
fixed.

</details>

<details>
<summary><code>R2</code> — the refusal named a cause that cannot produce this state</summary>

**Disposition:** FIXED. `orchestrator-agent.ts:1666` reads "deleted, moved, or on a volume that is no
longer mounted"; the container-overlay claim is gone from both message and comment.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:comment-analyzer`

</details>

<details>
<summary><code>R3</code> — hard-coded test counts were wrong by 3-7x</summary>

**Disposition:** FIXED. Numbers deleted rather than re-pinned, with an explicit note that no count
belongs there. Independently recounted: 27 literal `codebase_id` sites plus 28 `makeDispatchConversation()`
calls, all inside the guard's condition.

</details>

<details>
<summary><code>R4</code> — the persist comment referenced a branch #2577 deleted</summary>

**Disposition:** FIXED. All three sub-claims re-derived from code and history: the enumerated early
returns are exactly the deterministic-command branch plus this guard; `/workflow approve|reject`
returns inside that branch before the persist; a plain message at a gate flows through to the AI
turn. `git log -S"await approveWorkflow"` → `7752ec53 … (#2577)`, the commit cited.

</details>

<details>
<summary><code>R5</code> — the comment asserted a guard that was not in the tree</summary>

**Disposition:** FIXED, and it needed fixing twice. The first correction chose a time-anchored hedge
("UNMERGED as of this change"), which reality falsified within hours when #2551 merged at 10:18:16Z
(`b4599464`). The rebase replaced it with a claim about the tree's structure rather than a moment,
verified against the actual ordering: #2551's guard at `:1584`, the hoist at `:1627`, this guard at
`:1655`, the persist at `:1673`. "Directly above this one" is accurate and has no expiry.

**Found by:** `prp-core:comment-analyzer`, `prp-core:code-reviewer`

</details>

<details>
<summary><code>R6</code> — "keep both" was the wrong reconciliation for the test file</summary>

**Disposition:** FIXED. Performed rather than described: the branch is rebased onto `b4599464`,
`gh pr view` reports `MERGEABLE`, and exactly one `mockExistsSync` declaration survives. I also
composed the merge independently before the author's rebase landed and got 248 pass / 0 fail with
`type-check` exit 0. #2551's own head had moved to `1943381e`, promoting its reset to a top-level
`beforeEach`, so git auto-merged the declaration and the duplicate-declaration hazard never
materialised.

**Found by:** review synthesis

</details>

<details>
<summary><code>R7</code> — "every consumer of <code>codebases</code> runs far below" had a counterexample</summary>

**Disposition:** FIXED. Now "every OTHER consumer … (the guard just under this line is the one that
needed the hoist)", confirmed by enumerating the array's readers and by
`git diff eb642128^..eb642128` showing `listCodebases()` previously sat after the persist.

</details>

<details>
<summary><code>R8</code> — the stated rule omitted its own precondition</summary>

**Disposition:** FIXED. The rule now carries "One that needs data from further down (workflow
discovery, gate lookup) has to hoist that dependency first"; both named dependencies are real and
computed below the persist.

</details>

<details>
<summary>Notes carried but not raised as findings</summary>

Recorded so a later round does not re-derive them.

- **The "six cases" claim in `e7f7952c`'s message is narrower than it reads, though its conclusion
  holds.** The two guards' conditions read *five* boolean inputs, not three: `codebase_id !== null`
  and `scoped !== undefined` in addition to the `cwd`/`existsSync` triple. The six-case table
  enumerates the sub-space with those two held true. Both are separately accounted for, which is why
  the conclusion survives: dropping `codebase_id !== null` is provably **inert** (250 pass / 0 fail —
  `find(c => c.id === null)` is always `undefined` because `Codebase.id` is `z.string()`, so it is
  excluded by a type invariant rather than by a test), and dropping `scoped !== undefined` is caught
  by three existing tests including `scoped chat falls back to workspaces root and warns when
  codebase not found (deleted)`. Not worth amending a pushed commit — that costs a SHA change and a
  full re-verification to sharpen prose — but the precise version belongs somewhere durable, and the
  original overreach was mine, not the author's.

- **The R13 regression test is a pure mutation-catcher, and that is legitimate.** Under shipped code
  the guard short-circuits on `conversation.cwd === null` being false, so `existsSync(default_cwd)`
  is never reached in `runs the turn when the cwd override is healthy but default_cwd is gone`.
  Proven rather than reasoned: replacing the fixture predicate with one that *throws* for
  `/repos/test-repo` leaves the test passing. Case 4's correct outcome is identical to case 3's, so
  the test asserts nothing case 3 does not already assert — its whole discriminating power is against
  the clause deletion, which it does catch. The test's own comment already describes it that way.

- **Cells whose correct outcome duplicates a neighbour's are the ones that go uncovered.** The
  author's generalisation from the above, and it explains why R13 hid for three rounds: a
  state-space cell with no distinguishing assertion looks redundant, so no one writes the test, and
  only a mutation reveals it. Worth carrying past this guard.

- **`existsSync` does not prove a directory, and the guard's own advice is a route into that state.**
  `handleUpdateProject` (`orchestrator-agent.ts:2934`) accepts any path `existsSync` likes, so a user
  following this guard's suggestion can point `default_cwd` at a *file*; the next turn passes the
  guard and the provider spawns with a file as its working directory, reproducing the #2663 symptom.
  `registerFolder` (`handlers/clone.ts:524-531`) already does the stronger check
  (`stat(...).isDirectory()`). `prp-core:seam-analyzer` rated this **Important**; recorded here as a
  **note** instead, because the hole is entirely in `handleUpdateProject` on `dev` and predates this
  PR — #2665 neither introduces nor widens it, and the plan explicitly scoped a file-vs-directory
  distinction out. It is a good separate issue.
- **`/workflow run` reaches the same missing directory without passing the guard.** The
  deterministic-command branch returns at `:1512-1563`, above this guard, so `/workflow run` in the
  same broken conversation lands on unchecked `default_cwd` (`orchestrator-agent.ts:891` when
  `worktree.enabled === false`; `isolation/src/backends/in-place.ts:21` for folder projects).
  Pre-existing, separate surface. (`prp-core:seam-analyzer`, MEDIUM confidence — the paths were read,
  no run was executed.)
- **The recovery advice does not cover a deleted clone.** When `default_cwd` is a removed
  `~/.archon/workspaces/<owner>/<repo>/source`, neither `/update-project` nor `/setproject` restores
  it; the user must re-clone and then `/update-project` back to the same path. Both suggested commands
  still work, so this is completeness, not a broken outcome. (`prp-core:code-reviewer`)
- **The PR body has been overtaken by events.** Its *Merge order* section instructs a merger to keep
  one `mockExistsSync` and prefer "this PR's top-level `beforeEach`" — but #2551 landed first and
  `dev` already carries both, so the instruction is now history rather than guidance. Its *Validation*
  section still cites 243 and 247 pass; the real numbers are 241 (base) and 248 (head). Worth a pass
  before merge so the body is not read later as a description of the merged state.

</details>

## Agent Coverage

| Scope | Result |
|---|---|
| code | `R1`, `R2`, `R10` — all fixed; independently surfaced the #2551 merge mid-review |
| seams | `R10`, `R6` update; no drift between the two guards' event names, message shapes, or refusal paths |
| tests | `R9` — proved the pre-rebase assertion worked, which made the post-rebase regression detectable |
| comments | `R3`, `R4`, `R5`, `R7`, `R8` — all verified fixed against code and git history |
| **round 4 (fresh code / seams / tests / comments agents)** | `R13`, `R14`, `R15`, `R16` — independent re-verification at `95a3bba3`; R1-R12 dispositions re-derived from source, not accepted from the table |

## Validation

|---|---|---|
| `bun run validate` at `95a3bba3` (pristine tree, own command, real exit code, not piped) | PASS | `VALIDATE_EXIT=0` |
| `bun test src/orchestrator/orchestrator-agent.test.ts` (from `packages/core`, at `95a3bba3`) | PASS | 248 pass / 0 fail / 525 expect() calls |
| Mutation — disable #2551's guard (`if (false && …)`) | PASS (defect caught) | 243 pass / 5 fail; the rewritten test fails on `toContain('/worktrees/removed')`, received `"test response"` |
| Mutation — merge the guards (remove #2551's, widen this one to the effective cwd) | PASS (defect caught) | Fails on `toContain('working directory no longer exists')`; user receives the project-root message naming `/worktrees/removed` |
| Mutation — same, with the two positive assertions stripped | PASS (negatives discriminate) | `not.toContain('project directory no longer exists')` still fails, so the negative assertions are redundant but not vacuous |
| Mutation — collapse to `conversation.cwd ?? default_cwd` | Not caught, **as documented** | 248 pass / 0 fail, matching the comment's explicit disclaimer |
| Mutation — remove quoting *and* escaping | PASS (defect caught) | 247 pass / 1 fail (`quotes a project name containing whitespace…`) |
| Mutation — remove **only** the escaping | **Not caught** | 248 pass / 0 fail — basis for R12 |
| Mutation — remove the `return;` after the refusal | PASS (defect caught) | 246 pass / 2 fail |
| `parseCommand` round-trip, 9 names, before/after | PASS | Four plain names unchanged (incl. `O'Brien Ops`); five quote/backslash names fixed; combined case proves no double-escaping |
| CI at `95a3bba3` | PASS (Windows pending) | ubuntu test, docker-build, schema-upgrade, CodeRabbit all pass; `test (windows-latest)` still running, no failures |
| **Round 4 — independent re-verification** | | |
| `bun run validate` at `95a3bba3` (own worktree, own command, real exit code to a file, never piped) | PASS | `VALIDATE_EXIT=0` |
| `bun run validate` at `0ddf864f` (the head this round was briefed on) | PASS | `VALIDATE_EXIT=0` |
| `bun test src/orchestrator/orchestrator-agent.test.ts` at `95a3bba3` | PASS | 248 pass / 0 fail |
| Same suite at base `b4599464` (`origin/dev`) — baseline for the arithmetic | PASS | **241 pass / 0 fail**; head 248 = 241 + the 7 tests the diff adds |
| Rebase resolution — `git merge-base HEAD origin/dev` | PASS | Exactly `b4599464`; `git log HEAD..origin/dev` empty, so no `dev` commit is missing |
| Mutation — drop only `&& conversation.cwd === null` | **Not caught** | 248 pass / 0 fail — basis for R13 |
| Mutation — same, plus the R13 probe test | PASS (defect caught) | 248 pass / 1 fail, failing exactly on the probe (`Received number of calls: 0`) |
| R13 probe on the unmodified tree | PASS (probe is valid, not vacuous) | 249 pass / 0 fail |
| Early-return enumeration before the persist | 7 found | `:1512`, `:1519`, `:1526`, `:1536`, `:1563`, `:1613`, `:1680` — the comment accounts for six (R15) |
| **Round 4 addendum — re-verified at `de489421`** | | |
| `bun test src/orchestrator/orchestrator-agent.test.ts` at `de489421` | PASS | 249 pass / 0 fail |
| Mutation — remove only the escaping, keep the quotes (the R12 mutation, re-run) | PASS (defect now caught) | 248 pass / 1 fail on `escapes quotes and backslashes in the project name`; was 248 / 0 at `95a3bba3` |
| Mutation — drop only `&& conversation.cwd === null`, escaping intact | **Still not caught** | 249 pass / 0 fail at `de489421` — R13 unchanged by the newest commit |

## Verdict

**NEEDS FIXES**

One blocking item, and it was introduced by the commit that resolved the last one. `95a3bba3` fixed
R9 honestly — the rewritten test does fail under real mutations, and the disclaimer it added is the
right instinct. But the disclaimer generalises past its evidence. It says "a widened condition here
is dead code rather than a behaviour change"; the mutation that was measured is the
`conversation.cwd ?? default_cwd` collapse, and the sentence as written also covers deleting the
`&& conversation.cwd === null` clause, which is neither dead nor unobservable. That deletion passes
248/0 and refuses a chat turn running in a healthy worktree (**R13**).

The distance between those two mutations is the whole finding. One is genuinely unobservable because
the guard above absorbs it; the other is a live regression on the hot path of every project-scoped
chat turn. A comment that tells a reader "no test can hold you to this" about the second is worse
than no comment, because it supplies the reasoning for the edit and removes the instinct to check.

The fix is small and both halves are already validated in R13: narrow the disclaimer to the collapse
it actually measured, and add the probe. The probe is the positive counterpart the suite has never
had — every one of the seven new tests either sets `cwd: null` or has every path present, so no test
anywhere exercises a healthy override alongside a missing project root, which is the only state in
which this guard's scoping decision is observable.

R12 is now closed, and closed properly — `de489421` picked the fixture that pins both characters at
once, which is the one that can catch a double-escaping regression as well as a missing one.

R14 through R16 are prose and one log field; none blocks. R14 is worth doing in the same edit because
it is a two-line delete that restores a sentence this PR broke.

Nothing else is wrong. The production behavior is correct, and I re-derived the load-bearing claims
from source rather than accepting the dispositions: the rebase is clean against `b4599464` with both
guards present and disjoint, the hoist crosses only a `remote_agent_messages` write with no early
return between the old and new call sites, `bun run validate` exits 0 at both heads, and the test
arithmetic (241 + 7 = 248) holds when the baseline is actually run rather than quoted.

Wirasm:
## Review dispositions — all 8 findings addressed in `647937ce`

Every finding was valid. R1 in particular was a good catch: it is the same class of bug this PR exists to fix, one level up.

|---|---|---|
| `R1` | **FIXED** | Quoted the interpolated name: `` `/update-project "${scoped.name}" <new-path>` `` |
| `R2` | **FIXED** | "reset container volume" → "on a volume that is no longer mounted", in both the message and the comment |
| `R3` | **FIXED** | Deleted the counts rather than re-pinning them |
| `R4` | **FIXED** | "deterministic commands (including `/workflow approve\|reject`)" + an explicit note that a plain message at a gate is not a refusal since #2577 |
| `R5` | **FIXED** | The comment now says #2551 is unmerged and that a stale override is unguarded here until it lands |
| `R6` | **FIXED** | PR body rewritten — see the new **Merge order** section |
| `R7` | **FIXED** | "every **other** consumer of `codebases` runs far below" |
| `R8` | **FIXED** | The rule now carries its precondition: a refusal needing data from below must hoist that dependency first |

### R1 — reproduced, fixed, and the fix is guarded

Confirmed against the real parser before and after:

```
/update-project Client Ops <path>    → name "Client",     path "Ops <path>"
/update-project "Client Ops" <path>  → name "Client Ops", path "<path>"
```

New test `quotes a project name containing whitespace so the suggestion parses` uses a codebase named `Client Ops`. Removing the quoting turns exactly that test red, and the failure output prints the broken command a user would copy:

```
Expected to contain: "/update-project \"Client Ops\" <new-path>"
Received: "...- `/update-project Client Ops <new-path>` to point it at the new location..."
242 pass, 1 fail
```

`parseCommand` is mocked in this test file, so the round-trip suggestion could not be asserted through the real parser there; the assertion pins the emitted string instead, which is where the defect lived.

### R2 — accepted with the evidence

The container backend bind-mounts the project root read-only and keeps writes in a per-run overlay volume, so resetting that volume cannot remove `default_cwd`. Naming an impossible cause is a smaller version of this PR's own bug, so it went.

### R3 — the counts were wrong, and no number replaced them

The recount was right and mine understated the risk. Replaced with the mechanism and a note saying explicitly that no count belongs there, since any number rots on the next test added.

### R4 — independently confirmed

`git log -S"await approveWorkflow"` → `7752ec53 ... (#2577)` removed the branch, and enumerating every `return;` before the persist shows only the deterministic-command returns and this guard. The sentence was pre-existing staleness that this PR rewrote, so it was the right place to fix it.

### R6 — the merge instruction was wrong and is now explicit

"Keep both" was right for the guards and wrong for the fixture, where #2551 adds a byte-identical `mockExistsSync` declaration. The PR body now separates the three cases: keep both guards, keep exactly one fixture declaration preferring this PR's top-level `beforeEach`, keep both test blocks. That resolution was executed locally against #2551's real branch — **247 pass, 0 fail** with both guards present.

`bun run validate` exit 0. Focused suite 243 pass / 0 fail.

Issue #1331, opened 2026-04-21
node_complete.tokens inconsistent across nodes and missing cache fields


gemmawood:
## Summary

The per-node token counts emitted in workflow JSONL logs (`~/.archon/workspaces/<owner>/<repo>/logs/<runId>.jsonl`) are inconsistent across node types and omit cache-related fields entirely. Downstream consumers (cost dashboards, usage analytics) cannot produce accurate numbers from the current schema.

## Context

`node_complete` events carry `tokens: { input, output }`. For prompts that use caching (which applies to essentially every non-trivial workflow run), this is insufficient. Anthropic's API returns four distinct counters per turn — regular input, output, `cache_read_input_tokens`, `cache_creation_input_tokens` — and they are priced very differently (cache reads ~10% of input, cache writes ~125%). Collapsing them into a single `input` counter makes cost calculation impossible without guessing.

Worse, the collapse does not appear to be applied uniformly: some nodes report what looks like the uncached delta only (18 input tokens for a classify node that receives the full issue body), while other nodes report values consistent with cache reads folded into `input` at full weight.

## Evidence

Samples taken from four real runs on 2026-04-19 and 2026-04-21 (Archon workflows running against a downstream codebase):

| Workflow | Node | Reported input tokens | Plausibility |
|---|---|---:|---|
| triage | `classify` | 18 | Impossible — the classify prompt includes the issue title + body + labels, which is hundreds to low thousands of tokens before system prompt. |
| triage | `write-artifact` | 18 | Same node family; same anomaly. |
| investigate | `investigate` | 1,404,672 | Plausible only if cache reads are folded in at full weight. |
| develop | `implement` | 3,117,820 | Same. |
| develop | `validate` | 6,575,897 | Same. |

Example event (triage, `classify`):

\`\`\`json
{"type":"node_complete","step":"classify","content":"<inline>","duration_ms":13118,"tokens":{"input":18,"output":1191},"workflow_id":"2c593e6a09aab55b512c1d7e499bf174","ts":"2026-04-21T09:02:40.941Z"}
\`\`\`

Example event (investigate):

\`\`\`json
{"type":"node_complete","step":"investigate","content":"<inline>","duration_ms":220877,"tokens":{"input":1404672,"output":9741},"workflow_id":"7d71ddccc49cf693e700547753a231ab","ts":"2026-04-21T09:12:01.927Z"}
\`\`\`

[...]

PR #2671, opened 2026-08-21
fix(workflows): report a partial cache total as a floor instead of withholding it

### LINKED ISSUE #2662: fix(workflows): mixed-cache run totals keep gross input but drop cache axes
## Problem

When aggregating usage across nodes, cache axes are dropped if **any** contributing node did not report them, but `input` is still summed as gross. The result is a run total that silently understates cache and overstates full-price input, with nothing marking it as partial.

`sumTokenUsage` (`packages/workflows/src/dag-executor.ts:476-480`):

```ts
for (const axis of ['cacheRead', 'cacheWrite'] as const) {
  if (valid.every(usage => usage[axis] !== undefined)) {
    summed[axis] = valid.reduce((total, usage) => total + (usage[axis] ?? 0), 0);
  }
}
```

`input` and `output` above it sum unconditionally. `getDagResumeSnapshot` in `packages/core/src/db/workflow-events.ts` applies the same rule to resumed runs.

The unknown-versus-zero rule itself is correct and deliberate (#2654, and the amendment on #1331): a partial cache total must not be presented as complete. The gap is that dropping the cache axes while keeping gross `input` produces a *differently* wrong number rather than an honest one.

## Concrete failure

A run with two nodes, one reporting cache and one not:

| Node | input | cacheRead | cacheWrite |
|---|---:|---:|---:|
| a (reports cache) | 1,000,000 | 950,000 | 20,000 |
| b (no cache telemetry) | 5,000 | absent | absent |
| **run total** | **1,005,000** | **absent** | **absent** |

Node a's real full-price input is 30,000. The run total says 1,005,000 gross with no cache context, so any consumer pricing it at the input rate over-counts by roughly 30x. Nothing in the payload signals that a cache total was withheld rather than genuinely zero.

## Reachability

Not hypothetical, but not common either. It needs a run that mixes cache-reporting and non-cache-reporting usage in one aggregate:

- A resumed run whose earlier `node_completed` rows were written by a pre-#2654 build (those rows carry `{input, output}` only).
- A workflow mixing providers or Pi backends where one side reports no cache telemetry.

Fresh single-provider runs on current builds report all four axes on every node, so they are unaffected.

## Proposal (direction, not a spec)

The aggregate should be honest about which axes it knows. Options, roughly in increasing cost:

- Emit an explicit completeness marker (something like `cacheKnown: false`) so a consumer can tell "withheld" from "zero" without inferring it from a missing key.
- Alternatively, when the cache axes are withheld, also withhold or flag `input`, since a gross figure without its cache breakdown is the misleading part.
- Alternatively, carry the known-cache subtotal alongside a count of contributing nodes that reported nothing.

Worth deciding deliberately rather than picking the cheapest option, because it sets the precedent for every future optional usage axis.

## Acceptance (minimum)

- A run total that withheld a cache axis is distinguishable from one that genuinely summed zeros, without the consumer inspecting individual node events.
- Fresh single-provider runs keep their current complete totals unchanged.

Wirasm:
## Problem and outcome

When usage was aggregated across nodes, the cache axes were dropped entirely if *any* contributor did not report them, while `input` kept summing as gross. That kept the misleading half and discarded real information. A run with one cache-reporting node and one silent node reported `input: 1,005,000` with no cache context at all, so it read as "nothing was cached" and pricing it at the input rate over-counted by roughly 30x.

- **Outcome:** A cache total now sums every contribution that reported the axis and carries `cachePartial` when at least one did not, so it is an honest floor rather than a withheld value. `input - cacheRead - cacheWrite` becomes a true ceiling on full-price input instead of an unbounded over-estimate.
- **Invariant:** An aggregate never presents a partial cache total as complete, and never discards a cache figure it actually received.
- **Scope boundary:** No provider normalization changes. `input` stays gross and always summed. `total` and `cost` are not folded into the shared primitive. No backfill, and no UI or CLI rendering of the flag.
- **Root cause:** `input`/`output` reduced unconditionally while each cache axis was gated on `valid.every(...)`, so one silent contributor removed the axis and left its own cache tokens inside the gross `input` sum. Absence already encodes "unknown", making the result indistinguishable from a run where nothing cached.

## Review guidance

- **Feedback requested:** The floor semantics and the flag's propagation rule, plus whether pulling OpenCode's sub-agent merge into this change is the right call (see Known risk).
- **Start here:** `packages/providers/src/types.ts` — `mergeTokenUsage` is the whole rule; every other change is a call site or prose.
- **Review order:** `mergeTokenUsage` and its unit tests → `dag-executor.ts` (`sumTokenUsage` delegation) → `workflow-events.ts` (fold becomes an n-ary merge, flag round-trips) → `multi-agent.ts` → telemetry and docs.
- **Lower-attention areas:** The four prose corrections and the `packages/providers/package.json` test-script line, which only registers the new test file so CI runs it.
- **Known risk or uncertainty:** `multi-agent.ts` is the third site and is **not** named in #2662. I included it because leaving it lets an OpenCode node emit usage whose cache was already discarded, which then makes the whole run partial. It is a self-contained hunk and can be dropped without unpicking the rest.

## Solution

The rule was written three separate times, which is how it drifted, so it now lives once as `mergeTokenUsage` beside `TokenUsage`. Each cache axis sums over the contributions that define it and is emitted when at least one did; `cachePartial` is set when any contribution stayed silent, or when a contribution was itself already a floor. An axis nobody reported stays absent and is not flagged, because absence already means unknown.

The primitive is deliberately **pure** — no logging, no validation. All three call sites already own their validation with contexts that legitimately differ (persisted-JSON parsing in `@archon/core`, non-finite guarding in `@archon/workflows`), so callers validate and the primitive owns only the merge. That also keeps a logger out of the contract module.

`@archon/providers/types` already exports a runtime value across package boundaries (`CONTAINER_ENV_DENYLIST`), so this needed no new module, subpath, or dependency-order exception.

The resume path changes from a pairwise fold to collecting and merging once. That is what makes the rule n-ary: a pairwise fold cannot distinguish "one of five contributions reported" from "one of two".

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | One silent contributor erased the cache axes for the whole aggregate while gross `input` still included that cache. | Cache sums over the contributions that reported it; `cachePartial` (run metadata `total_cache_partial`, telemetry `cache_partial`) marks it as a floor. |
| **Failure behavior** | A narrowed total was indistinguishable from "nothing cached", producing a silent, unbounded over-estimate. | A floor is explicitly labelled, so a consumer gets a bounded number and knows it is a bound. |

## Changed seams

|---|---|---|
| `TokenUsage` | Adds `cachePartial?: true`, set only by aggregation, never by a provider. | `packages/providers/src/types.ts` |
| Node/run usage aggregation | `sumTokenUsage` keeps its name, signature, and warn events; delegates the merge. | `packages/workflows/src/dag-executor.ts` |
| Persisted events → resume | Flag is read back off stored JSON and propagates into the resumed total. | `packages/core/src/db/workflow-events.ts` and its tests |
| OpenCode sub-agent merge | Cache follows the shared rule; `total`/`cost` composition unchanged, including the lone-sub-agent passthrough. | `packages/providers/src/community/opencode/multi-agent.ts` |
| Anonymous telemetry | Adds `cache_partial`; schema version 5 → 6. | `packages/paths/src/telemetry.ts` |

## Correction pass (commits 2 and 3)

Independent review and CodeRabbit both found the same real defect in the first commit, and verifying it surfaced a third instance. All are fixed in `64df1ee4` and `d9e37de4`.

Three places rebuilt a `TokenUsage` from an existing one and dropped `cachePartial` on the way, so a floor silently became exact one boundary later:

| Site | Effect | Fix |
|---|---|---|
| `sumTokenUsage` | Every `@archon/workflows` fold went through it, so the flag survived only when the *final* merge happened to disagree on an axis. This also made the OpenCode hunk a no-op: the node-level flag was stripped one hop later, before the event was persisted. | Spreads the incoming usage instead of rebuilding it, so a future axis cannot be dropped the same way. A non-finite optional axis is cleared rather than `delete`d (`no-dynamic-delete`); every reader already treats an undefined axis as unreported. |
| `childOutcomeFromRun` | A child run's floor reached its parent as an exact total. | Reads `total_cache_partial` back out of the child's run metadata. |
| `readSignaledTokens` | A loop that paused on a floor resumed claiming an exact total. | Carries the flag through the loop-gate pause. |

The reviewer also noted that every guard merged exactly *twice*, and rounds one and three both report cache — which is why this whole class passed validation unnoticed. Added a three-node run with a **silent middle node**, so the silence must survive two further folds. It is one of three tests that now go red under the `sumTokenUsage` mutation.

`packages/core/src/db/workflow-events.ts` was the one reconstruction site that already got this right, and was the model for the other three.

## Validation

- `bun run validate` at the delivered head `d9e37de4`, in a **fresh detached worktree** with `git status` empty before the run — **exit code 0**, all 11 steps, 7,659 tests passing, zero failures. Exit status captured directly rather than through a pipe, which would have reported the pipe's status instead. (The first review flagged that an earlier run could not be cleanly attributed because the worktree was being edited concurrently; this run removes that doubt.)
- **Every new guard was seen failing before it was trusted.** Reverting `mergeTokenUsage` to the old `every(...)` gate produced 8 failures across all four suites (providers/types 3, core/workflow-events 3, workflows/dag-executor 1, providers/opencode 1); restoring returned all to green. Deleting the `cache_partial` emission separately reddened the telemetry test. The OpenCode assertion was later rewritten for type-safety, so it was re-mutated and seen red again rather than assumed.
- AC3 (complete aggregates unchanged) is proven by a **pre-existing** test that pins the exact aggregate object with no flag and still passes untouched.
- One pre-existing expectation changed rather than being deleted: `includes failed-node usage without treating the failed node as completed` pairs a cache-reporting completed row with a silent failed row, so it now asserts the floor. Its original purpose is unchanged.
- **Not verified:** No live paid-provider run. All behavior is covered at the typed aggregation boundaries with deterministic tests, and no provider normalization changed.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Purely additive: one optional boolean on schemaless JSON. Old readers ignore it. Cache totals that previously vanished now appear, which is strictly more information. No schema migration. | A run whose rows all predate #2654 still resumes with cache absent and no flag — covered by test. |
| Rollout / rollback | Single behavioral commit; reverting restores the old gate and stored flags are simply ignored. | No persisted-shape lock-in |
| Observability | The flag *is* the fix: a floor becomes distinguishable from a complete total, including in aggregate telemetry where the two would otherwise pool. | `packages/paths/src/telemetry.ts`, telemetry test |
| Documentation / communication | Four in-repo prose sites corrected, found by grepping the phrasing rather than trusting one known line. Two GitHub surfaces still state the superseded rule and need comments after merge: the #2654 description and the amendment on #1331. | `authoring-workflows.md`, `configuration.md`, `README.md`, two docstrings |

## Links

- Closes #2662
- Related #1331
- Related #2654

## Summary by CodeRabbit

- **New Features**
  - Cache-read and cache-write token totals are retained when available, even if some workflow steps lack cache data.
  - Telemetry identifies when cache totals are partial lower-bound estimates.
  - Token usage is aggregated consistently across workflow steps, retries, resumed workflows, and multi-agent runs.
  - Partial-cache status is preserved across child workflows and persisted run results.

- **Documentation**
  - Updated telemetry and workflow authoring guidance to explain cache totals, partial indicators, and lower-bound estimates.


Wirasm:
## Ready to merge

Correction verification over `d8230ae9..d9e37de4` — two commits, three files, +165/-5. All four prior findings are genuinely fixed, and I did not take the implementer's word for any of them: I reverted each correction in my own throwaway worktree and watched the named guard go red. R1's fix goes red on exactly the three tests the implementer claimed, R2 and R3 each go red on one, and the R4 test really is a three-contribution fold with a silent middle rather than another two-round permutation. The one causal shape from the last pass is closed at every reconstruction site on the path: `sumTokenUsage`, `readSignaledTokens`, `childOutcomeFromRun`, and the already-correct `workflow-events.ts:308-311`.

The `delete` → `= undefined` substitution — the part most likely to be quietly wrong — is sound, and for a stronger reason than "the readers tolerate it". `normalized` never leaves `sumTokenUsage`: it is pushed to a local array and consumed only by `mergeTokenUsage`, which builds a fresh literal and copies `input`/`output`/`cacheRead`/`cacheWrite`/`cachePartial` explicitly. So no `undefined`-valued key ever reaches JSON persistence or a `toEqual`, and the JSON and assertion halves of the question are unreachable rather than merely safe. `mergeTokenUsage`'s reporter filter is value-based (`usage[axis] !== undefined`, `types.ts:216`), so an assigned-`undefined` axis is indistinguishable from an absent one *and* still counts toward `reporters.length < usages.length` — a NaN axis narrows the aggregate to a floor instead of being forgiven. I proved that empirically: a non-finite axis and a genuinely absent axis produce byte-identical output. The warn path fires under the identical condition as before (`!Number.isFinite(value)` versus the old `else` of `Number.isFinite(value)`), same event, same payload. `exactOptionalPropertyTypes` is not set anywhere in the tree, so the assignment type-checks. The spread now also carries `total` and `cost`, which the old literal dropped — `mergeTokenUsage` never copies them, so nothing escapes.

Four new Suggestions, none blocking, all introduced by the correction itself. Three are test-and-scaffolding hygiene. The fourth (`R5`) is worth a maintainer decision but not a merge block: `signaledTokens` is a nested object inside a JSON blob that SQLite deep-merges, so a boolean the correction newly made reachable can go stale across two gates in one run — in the conservative direction only, and through a hole that already leaks numeric cache axes on `dev` today.

**0 blocking · 4 non-blocking**

**Validation:** `bun run validate` exit 0 at `d9e37de4` in a fresh detached worktree, `git status` empty before and after — 11 steps, 7,659 tests passing, 0 failures. This closes the attribution caveat from the previous pass. Three independent revert-mutations each reddened exactly the expected guards.

**Resolved:** 4 (`R1`, `R2`, `R3`, `R4`)

### Findings

|---|---|---|---|
| `R1` | Important | `sumTokenUsage` stripped `cachePartial` from every entry before merging, so a floor silently became a complete total on re-aggregation | FIXED |
| `R2` | Important | `readSignaledTokens` never read `cachePartial` back, so a loop that paused on a floor resumed claiming an exact total | FIXED |
| `R3` | Important | `childOutcomeFromRun` never read `total_cache_partial`, so a child run's floor reached its parent as an exact total | FIXED |
| `R4` | Suggestion | Every new test merged exactly once, which is why the whole R1–R3 class passed validation unnoticed | FIXED |
| `R5` | Suggestion | A gate's `signaledTokens.cachePartial` can survive into a later gate in the same run on SQLite, over-flagging an exact total as a floor | OPEN |
| `R6` | Suggestion | The new `re-merging an aggregate keeps it a floor` test duplicates a guard 30 lines above it and dies to the same single mutation | OPEN |
| `R7` | Suggestion | The new `settledRun` test helper double-casts through `unknown` where the file's existing `makeWorkflowRun` would type-check | OPEN |
| `R8` | Suggestion | The negative `childOutcomeFromRun` test uses `toEqual`, which cannot pin that `cachePartial` is absent rather than `undefined` | OPEN |

<details>
<summary><code>R1</code> — <code>sumTokenUsage</code> drops <code>cachePartial</code> on every fold</summary>

**Impact:** (from the previous pass) `cachePartial` reached run metadata and telemetry only by the accident of the final merge step happening to disagree on an axis. Any three-contribution fold with a silent middle, any single already-partial entry re-folded, and every resumed run reintroduced #2662 — a cache total that is a lower bound presented as exact.

**Evidence:** `packages/workflows/src/dag-executor.ts:465` now builds `const normalized: TokenUsage = { ...usage };` instead of the field-by-field literal, so an incoming floor marker survives into `mergeTokenUsage`. The non-finite branch at `:472` assigns `normalized[axis] = undefined` rather than deleting.

I verified the substitution rather than accepting it, because an `undefined`-valued key is not automatically an absent one:

- **The JSON and `toEqual` halves of the question are unreachable.** `normalized` has exactly two uses — `valid.push(normalized)` at `:477` and `return mergeTokenUsage(valid)` at `:483`. `mergeTokenUsage` (`packages/providers/src/types.ts:207-224`) constructs a fresh `merged` literal and only ever adds a key via explicit assignment. No `undefined`-valued key, and none of the spread's newly carried `total`/`cost`, can reach run metadata, telemetry, the JSONL log, or an assertion.
- **The one reader that does exist treats them identically.** `types.ts:216` filters `usages.filter(usage => usage[axis] !== undefined)`. Critically, the entry still counts in `usages.length`, so `reporters.length < usages.length` still fires and a NaN axis narrows the aggregate to a floor — the same as before.
- **Empirically byte-identical.** Running the corrected normalization against the real `mergeTokenUsage`: `[{input:10,output:1,cacheRead:5,cacheWrite:NaN}, {input:20,output:2,cacheRead:7,cacheWrite:3}]` and `[{input:10,output:1,cacheRead:5}, {input:20,output:2,cacheRead:7,cacheWrite:3}]` both stringify to `{"input":30,"output":3,"cacheRead":12,"cacheWrite":3,"cachePartial":true}`. `null` and string axes still take the clearing branch — `Number.isFinite` does not coerce. `{input:10,output:1,total:999,cost:1.5,cacheRead:5}` returns `{"input":10,"output":1,"cacheRead":5}`, so the spread's extras do not escape.
- **The warn path is unchanged in behavior.** Old `if (Number.isFinite(value)) … else warn` and new `if (!Number.isFinite(value)) { … warn }` fire on exactly the same condition, with the same `dag.usage_optional_tokens_non_finite_ignored` event and payload.
- **Types hold.** No `exactOptionalPropertyTypes` in the root `tsconfig.json` or any package tsconfig, so `normalized[axis] = undefined` is a legal assignment to `cacheRead?: number`. `type-check` and `eslint --max-warnings 0` are clean on the file. `cachePartial?: true` is not weakened: a runtime `cachePartial: false` from untyped JSON now rides the spread into `normalized`, but `types.ts:214` tests `=== true` and `:222` only assigns `true`, so it cannot reach any returned or persisted `TokenUsage` — confirmed by probe.

**Required outcome:** Satisfied.

**Found by:** `prp-core:type-design-analyzer`, `prp-core:code-simplifier` (originally); verified by all four scopes plus coordinator mutation.

**Disposition:** FIXED. Guard seen red by me, not inferred: reverting `:465-478` to the field-by-field rebuild in a throwaway worktree turns exactly three tests red in `packages/workflows` — `iterates at resume when feedback was given, even on a signal-bearing gate (#2074 C)`, `flags the run total when only a middle node is silent about cache`, and `best-effort provider: an attempt without cache telemetry narrows the total to a floor` (555 pass / 3 fail). Reverted; tree clean.

</details>

<details>
<summary><code>R2</code> — loop-gate resume drops the flag from <code>signaledTokens</code></summary>

**Impact:** (from the previous pass) An interactive `loop:` that accumulated a partial cache total, paused at its gate, and was approved resumed believing its total was exact — losing the floor marker for the whole run.

**Evidence:** `packages/workflows/src/dag-executor.ts:3602` now destructures `cachePartial` alongside the other four keys, and `:3623` forwards it only when strictly `true`. The boundary agrees in both directions: the writer at `dag-executor.ts:5905` persists `signaledTokens: loopTotalTokens ?? null`, `ApprovalContext.signaledTokens` is typed `TokenUsage | null` on a deliberately non-Zod interface (`packages/workflows/src/schemas/workflow-run.ts:305`), and `packages/core/src/db/workflows.ts:1064` writes it through `JSON.stringify`, so nothing strips the key in transit.

The guard is not decorative. In the updated fixture at `dag-executor.test.ts:9443`, the gate floor and the resumed iteration *both* report *both* cache axes, so the decoded flag is the only possible source of `cachePartial` in the asserted node total (`:9504`) and run-metadata write (`:9513`).

**Required outcome:** Satisfied.

**Disposition:** FIXED. Guard seen red: deleting only the `:3623` forward turns `iterates at resume when feedback was given, even on a signal-bearing gate (#2074 C)` red and nothing else (557 pass / 1 fail). `prp-core:type-design-analyzer` additionally mutated the guard in the false-positive direction (`cachePartial !== false`) and reddened a different test, `finalizes at resume from persisted signaledOutput on a bare approve`, whose fixture carries no flag — so both directions are pinned.

</details>

<details>
<summary><code>R3</code> — a child run's floor reaches its parent as an exact total</summary>

**Impact:** (from the previous pass) A `workflow:` sub-run whose own total was `cachePartial: true` wrote `total_cache_partial` into run metadata and the parent never read it back, so a parent could report a complete cache total that was in truth a floor inherited from its child.

**Evidence:** `packages/workflows/src/dag-executor.ts:603` now reads `md.total_cache_partial === true` and sets `cachePartial: true as const` on the reconstructed `tokens`. The sole writer is `persistRunUsage` at `dag-executor.ts:9300`; grepping for `total_cache_read_tokens|total_cache_write_tokens` in non-test code returns only `:591`, `:593`, `:9294`, `:9297`, so there is no third consumer left behind. `childOutcomeFromRun` was already exported for `executor.ts:939`, so the new tests widened no API surface.

Both directions are covered by `describe('childOutcomeFromRun cache reporting')` at `dag-executor.test.ts:23806` — one flagged child, one complete child that must stay unflagged.

**Required outcome:** Satisfied.

**Disposition:** FIXED. Guard seen red: deleting only the `:601-603` read turns `carries a child run total that is only a floor up to its parent` red and nothing else (557 pass / 1 fail).

</details>

<details>
<summary><code>R4</code> — the new tests all merged exactly once, hiding the whole R1–R3 class</summary>

**Impact:** (from the previous pass) The missing axis in the suite was *re-aggregation*, not another single-merge permutation. That is why a green `bun run validate` coexisted with three reachable holes in the PR's core invariant.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:11453` adds `flags the run total when only a middle node is silent about cache` — three sequential nodes `a → b → c` with contributions `{10,1,cacheRead 5,cacheWrite 0}`, `{20,2}` (silent), `{30,3,cacheRead 9,cacheWrite 1}`, asserting `total_cache_partial: true` on the run write. I read the body and traced the fold path rather than trusting the name: the run total accumulates pairwise at `dag-executor.ts:8413` (`sumTokenUsage([...prior, output.tokens])`), so this folds three times and the third fold re-aggregates an already-partial aggregate. Under the R1 defect the third fold sees two contributions that both report both axes, recomputes `partial = false`, and drops the flag.

Two further folds were added: `dag-executor.test.ts:11588` extends the best-effort-provider test with a run-level assertion (the pre-existing node-level assertion stays green under the R1 defect; only the new one reddens), and `packages/providers/src/types.test.ts:91` re-merges an aggregate at the primitive level — see `R6` on that last one.

**Required outcome:** Satisfied.

**Disposition:** FIXED. The new test is one of the three I watched go red under the R1 revert, so it is load-bearing rather than a passing decoration.

</details>

<details>
<summary><code>R5</code> — a gate's <code>signaledTokens.cachePartial</code> can go stale across two gates on SQLite</summary>

**Impact:** On SQLite, a run with two interactive `loop:` gates where the first pauses on a floor and the second does not: the second gate's resumed node total, its `total_cache_partial` run metadata, and its `cache_partial` telemetry row all claim a floor the run does not have. Silent, and conservative in direction — it can only over-flag, never claim exactness. Not reachable on Postgres.

**Evidence:** `packages/core/src/db/workflows.ts:1047` pauses with `dialect.jsonMerge('metadata', 2)`, which on SQLite is `json_patch` (`packages/core/src/db/adapters/sqlite.ts:850`) — RFC 7396, which recurses into nested objects. The explicit-null reset list at `workflows.ts:1052-1076` defeats that for every *top-level* approval sub-field, and its own comment says so, but `signaledTokens` is an object and its interior keys are merged rather than replaced. `prp-core:seam-analyzer` and `prp-core:type-design-analyzer` each proved it independently against `bun:sqlite`: a stored `{input:40,output:4,cacheRead:20,cacheWrite:0,cachePartial:true}` patched with a later gate's complete usage yields a result still carrying `cachePartial: true`.

This reaches the correction diff. `loopTotalTokens` is built by `sumTokenUsage` (`dag-executor.ts:3993`, `:4919`), which at `d8230ae9` stripped the flag — so `signaledTokens` could not carry `cachePartial` before the correction, and the reader added at `:3623` is what makes a stale value consequential. The same hole already leaks *numeric* `cacheRead`/`cacheWrite` on `dev` today, which is strictly worse and entirely pre-existing.

**Required outcome:** Optional. Either reset the nested object's optional axes explicitly at `workflows.ts:1064`, or accept the conservative direction. The pre-existing numeric leak is the larger half and is out of this PR's scope — worth its own issue if the maintainer wants the nested reset tightened.

**Found by:** `prp-core:seam-analyzer` (raised as a Suggestion); `prp-core:type-design-analyzer` reached the same mechanism and explicitly declined to raise it, on the grounds that the numeric leak is pre-existing and the new field's direction is safe. Disagreement preserved: I record it, and I agree with both that it does not block.

**Disposition:** OPEN as a Suggestion. Does not block — it cannot produce a false exact total, only a false floor marker.

</details>

<details>
<summary><code>R6</code> — the new <code>types.test.ts</code> case duplicates a guard 30 lines above it</summary>

**Impact:** 19 lines of test that add no guard. Small, but it is the kind of duplication that makes a later reader believe two invariants are pinned where one is.

**Evidence:** `packages/providers/src/types.test.ts:91` (`re-merging an aggregate keeps it a floor`) and `:62` (`propagates a contribution that is already a floor`) both build a `cachePartial` aggregate whose only cache axis is `cacheRead`, merge it with a complete contribution that also reports `cacheRead`, and assert the flag survives. The only difference is that the new one obtains its first aggregate by calling `mergeTokenUsage` instead of writing the literal — a distinction the primitive cannot observe. `prp-core:code-simplifier` mutated the single line both target (`types.ts:214`) and watched both go red together. Its second assertion (`first?.cachePartial` from a silent contributor) is already covered at `:19`.

Worth naming precisely: this test is also in the wrong package to close `R4`. R4's complaint was about the lossy `sumTokenUsage` wrapper in `@archon/workflows`, which this test never touches. The two `dag-executor.test.ts` additions are what actually close it.

**Required outcome:** Optional — delete it, or keep it as documentation of the re-merge case at the primitive level.

**Disposition:** OPEN as a Suggestion.

</details>

<details>
<summary><code>R7</code> — the new <code>settledRun</code> helper double-casts where an existing helper fits</summary>

**Impact:** A `as unknown as` escape hatch in a test file, which nothing will catch — ESLint does not cover test files in this package.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:23807` defines `settledRun` as `({ id: 'child-1', status: 'completed', metadata }) as unknown as Parameters<typeof childOutcomeFromRun>[0]`. The same file already has `makeWorkflowRun(id = 'dag-test-run-id', overrides?: Partial<WorkflowRun>): WorkflowRun` at `:352`, in scope, and `WorkflowRun.metadata` is `z.record(z.string(), z.unknown())` (`packages/workflows/src/schemas/workflow-run.ts:133`) — so `makeWorkflowRun('child-1', { status: 'completed', metadata })` types cleanly and drops the double cast.

**Required outcome:** Optional — swap the helper for `makeWorkflowRun`.

**Disposition:** OPEN as a Suggestion.

</details>

<details>
<summary><code>R8</code> — the negative <code>childOutcomeFromRun</code> test cannot pin absence</summary>

**Impact:** `leaves a complete child total unflagged` would still pass if the implementation emitted `cachePartial: undefined` instead of omitting the key. The implementation uses a conditional spread, so the key is genuinely absent today — the test just does not hold it there.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:23838-23849` asserts with `toEqual`. I confirmed Bun's semantics empirically rather than assuming they match Jest's: `expect({a:1,b:undefined}).toEqual({a:1})` passes, and the same comparison under `toStrictEqual` throws. Given `R1` deliberately introduced an `undefined`-valued-key idiom into this file's neighborhood, pinning the distinction here is cheap.

**Required outcome:** Optional — `toStrictEqual` on the negative case.

**Disposition:** OPEN as a Suggestion.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | R1–R4 all FIXED. New: R8. Corrected its own disproved conclusion from the previous pass by tracing the `sumTokenUsage` wrapper rather than the primitive. |
| `seams` (`prp-core:seam-analyzer`) | R1–R4 all FIXED. New: R5, R8. Enumerated nine sites on the path that build or carry a `TokenUsage` and found all nine correct. |
| `types` (`prp-core:type-design-analyzer`) | R1–R4 all FIXED, each guard seen red in its own private copy. No new findings. Reached R5's mechanism and declined to raise it. |
| `simplify` (`prp-core:code-simplifier`) | R1–R4 all FIXED. New: R6, R7. |

All four scopes were retained because each owned a prior finding. `tests`, `comments`, `errors`, `docs` were not run — no prior finding belonged to them and the correction diff changes no documentation or error path.

Examined and clean, with the evidence that settled each:

- **Every reconstruction site on the path is now correct.** `sumTokenUsage` (`dag-executor.ts:465`), `childOutcomeFromRun` (`:603`), `readSignaledTokens` (`:3623`), and `getDagResumeSnapshot` (`packages/core/src/db/workflow-events.ts:308-311`, already correct in the original PR). `sumFanOutTokens` (`:6727`) passes whole objects through; `totalTokens: priorUsage?.tokens` (`:9257`) is a direct assignment; the OpenCode aggregate spreads. Every `tokens:` write in `dag-executor.ts` passes the object wholesale — no other field-by-field rebuild exists.
- **Both persisted-JSON boundaries agree writer-to-reader.** Loop gate: writer `:5905` / reader `:3623`. Child run metadata: writer `:9300` / reader `:603`. Resume snapshot: `workflow-events.ts:308-311`.
- **No key-presence reader exists.** No `'cacheRead' in …`, `hasOwnProperty`, or `Object.keys`/`entries` over a usage object anywhere in `packages/` — which is the other way an `undefined`-valued key could have diverged from an absent one.
- **The `R1` comment's second clause is broader than the reachable path.** `dag-executor.ts:473-474` says "every reader treats an undefined axis as unreported, and `JSON.stringify` omits the key entirely." Both halves are true, but `normalized` is never serialized, so the JSON half defends a path this object does not take. Raised by `prp-core:code-reviewer` and `prp-core:code-simplifier`; I judge it below the bar for a finding — the load-bearing first clause is accurate and the comment records why the `delete` is gone, which is exactly what a future reader needs.
- **`persistRunUsage` writes `total_cache_partial` only when true and never clears it**, while run metadata is deep-merged. Not reachable today — the accumulator's flag is monotonic within a run — and it would err toward over-flagging. Noted by `prp-core:code-reviewer` so it is not read as a gap.

#### Validation

|---|---|---|
| `bun run validate` (root, fresh detached worktree at `d9e37de4`) | PASS | `VALIDATE_EXIT=0`. `git status` empty before *and* after the run, HEAD still `d9e37de4` — the previous pass's attribution caveat is closed. All 11 steps ran (5 generated-file checks, type-check, lint, format:check, test:install, `bun --filter '*' --parallel test`, `bun test ./scripts/`). 7,659 tests passing (7,564 + 95), 0 failures across every reported suite. |
| R1 revert-mutation: field-by-field rebuild restored in `sumTokenUsage` | FAIL as expected — guard seen red | `packages/workflows`: 555 pass / 3 fail. Exactly `iterates at resume … (#2074 C)`, `flags the run total when only a middle node is silent about cache`, `best-effort provider: an attempt without cache telemetry narrows the total to a floor`. Confirms the implementer's claim independently. |
| R2 revert-mutation: `cachePartial` forward removed from `readSignaledTokens` | FAIL as expected — guard seen red | `packages/workflows`: 557 pass / 1 fail — `iterates at resume when feedback was given, even on a signal-bearing gate (#2074 C)`. |
| R3 revert-mutation: `md.total_cache_partial` read removed from `childOutcomeFromRun` | FAIL as expected — guard seen red | `packages/workflows`: 557 pass / 1 fail — `carries a child run total that is only a floor up to its parent`. |
| `undefined`-vs-absent equivalence probe against the real `mergeTokenUsage` (5 cases) | PASS | Non-finite axis stringifies identically to an absent axis; `null`/string axes still take the clearing branch; `total`/`cost` do not escape; a runtime `cachePartial: false` cannot become a false exact claim; and Bun's `toEqual` does not distinguish an `undefined`-valued key from an absent one while `toStrictEqual` does (which is what `R8` rests on). |

All mutations were applied in a throwaway detached worktree, reverted immediately, and the tree confirmed clean; the PR branch and the implementer's worktree were never touched.

</details>

PR #2675, opened 2026-08-21
fix(core): a fresh approval gate no longer inherits the previous gate's cache tokens

### LINKED ISSUE #2673: fix(core): a later gate inherits an earlier gate's cache tokens on SQLite
## Problem

On SQLite, a later interactive gate in the same run inherits cache token values from an earlier gate. The explicit-null reset at `packages/core/src/db/workflows.ts:1052-1076` protects every *top-level* approval sub-field, and its comment says it covers "EVERY optional approval sub-field" — but `signaledTokens` is an **object**, and `json_patch` (RFC 7396) recurses into it and merges key by key. Interior keys a later gate omits survive from the earlier one.

The numeric half is the serious one: it fabricates cache counts for a provider that reported none.

## Reproduction

Against `bun:sqlite`, mirroring what `pauseWorkflowRun` writes:

```ts
const g1 = { approval: { signaledTokens: { input: 40, output: 4, cacheRead: 20, cacheWrite: 3 } } };
db.run("INSERT INTO t VALUES (1, ?)", [JSON.stringify(g1)]);
// Gate 2: usage from a provider with no cache telemetry at all.
const g2 = { approval: { signaledTokens: { input: 90, output: 9 } } };
db.run("UPDATE t SET metadata = json_patch(metadata, ?) WHERE id = 1", [JSON.stringify(g2)]);
```

```
gate2 stored as: {"input":90,"output":9,"cacheRead":20,"cacheWrite":3}
expected       : {"input":90,"output":9}
```

Gate 2's node total, its `total_cache_read_tokens` / `total_cache_write_tokens` run metadata, and its telemetry row all report 20 and 3 cache tokens that no provider ever reported. Silent, and in the **unsafe** direction: it overstates cache, which understates full-price input.

A second, milder instance was introduced by #2671, which added `cachePartial`:

```
gate1 floor + gate2 complete  ->  {"input":90,"output":9,"cacheRead":50,"cacheWrite":2,"cachePartial":true}
```

Gate 2's exact total is marked a floor. That one is conservative — it can only over-flag, never claim exactness — and the numeric leak above already existed and is strictly worse.

Not reachable on Postgres, where `||` replaces the approval object wholesale.

## Scope

Needs a run with two interactive `loop:` gates on SQLite where the second gate's usage has fewer cache keys than the first. Rare, but SQLite is the default backend.

## Proposal (direction, not a spec)

The general problem is that the reset list is one level deep while `json_patch` is not. Options:

- Reset `signaledTokens`' interior optional axes explicitly, the way the top-level fields already are. Narrowest, but re-creates the same trap for the next nested object added to `ApprovalContext`.
- Write the whole `signaledTokens` object with a replace rather than a merge, so it is atomically the new gate's value or absent.
- Make the dialect seam not deep-merge nested approval objects at all, so SQLite and Postgres agree. Largest, and the only one that fixes the class rather than the instance.

Worth picking deliberately, since `ApprovalContext` is likely to gain more structured sub-objects.

## Acceptance (minimum)

- A second gate whose usage omits a cache axis does not inherit the first gate's value for it, on SQLite and Postgres alike.
- The comment at `workflows.ts:1052` matches what the code actually guarantees.

Wirasm:
## Problem and outcome

On SQLite — the default backend — a later interactive gate in the same run reported cache-token counts that no provider ever sent. It failed silently and in the unsafe direction: overstated cache understates full-price input, so the run's cost figures were wrong for a gate that never used the cache.

- **Outcome:** A fresh approval gate stores exactly its own `ApprovalContext`. A second gate whose usage omits `cacheRead`/`cacheWrite` no longer inherits the first gate's numbers, and a gate-1 `cachePartial` floor no longer marks gate 2's exact total as a floor.
- **Invariant:** A fresh pause stores exactly the keys the caller set — at every nesting depth — and nothing a prior gate of the same run left behind survives, on either dialect.
- **Scope boundary:** `dialect.jsonMerge` is unchanged, as are its nine other call sites (run metadata, sessions, isolation environments). `resolveApprovalGate` also keeps its merge — every caller passes the full stored context plus fields, so a merge and a replace produce identical rows there; its docblock now records that requirement.
- **Root cause:** The pause wrote the approval object through `dialect.jsonMerge`, which on SQLite is `json_patch` — RFC 7396, which *recurses*. The explicit-null reset guarded only the top-level approval fields, so interior keys of the nested `signaledTokens` object that the new gate omitted were left untouched by the merge and survived from the earlier gate. On resume `readSignaledTokens` (`packages/workflows/src/dag-executor.ts:3596`) read them back as real token accounting.

## Review guidance

- **Feedback requested:** Correctness of the two SQL expressions, and whether the helper belongs in `workflows.ts` rather than on `SqlDialect`.
- **Start here:** `packages/core/src/db/workflows.ts:130-155` — `writeApprovalMetadata`, the dialect-aware replace that the whole fix rests on.
- **Review order:** `writeApprovalMetadata` → its use in `pauseWorkflowRun` (`packages/core/src/db/workflows.ts:1073-1094`) and the reset block it deletes → the new SQLite cases in `packages/core/src/db/workflows.resume-cas.integration.test.ts` → the comment corrections in `packages/workflows/src/schemas/workflow-run.ts` and `packages/workflows/src/dag-executor.ts`.
- **Lower-attention areas:** The `dag-executor.ts` and `workflow-run.ts` changes are comment-only; the `?? null` values they describe are unchanged. `subrun.test.ts` updates a fake store to mirror the new contract.
- **Known risk or uncertainty:** The Postgres branch has no automated test against a live engine — the repo has no PG-backed test harness. It was verified manually against PostgreSQL 18.0 and is behaviour-identical to today's `||`; see Validation.

## Solution

`pauseWorkflowRun` stops merging the approval object and sets it wholesale, while caller-supplied run-level metadata (`pending_writeback`) still merges at the top level in the same atomic UPDATE:

- Postgres: `jsonb_set(metadata || $2::jsonb, '{approval}', $3::jsonb, true)`
- SQLite: `json_set(json_patch(metadata, $2), '$.approval', json($3))`

The helper lives beside `unresolvedGateClause()` (`packages/core/src/db/workflows.ts:112-128`) and `claimWriteback`'s `extract` (`:1117-1120`), the two existing in-file dialect-branched SQL fragments, so the change stays in one file and leaves `SqlDialect` and its hand-maintained mock untouched.

Fixing it this way is what lets the 14-line explicit-null reset block and its 12-line justification be deleted rather than extended. Resetting `signaledTokens`' interior axes would have fixed this instance and re-armed the same trap for the next nested field added to `ApprovalContext`; replacing the object makes the guarantee structural. Every reader already treats an absent key exactly like a JSON null (`!= null`, `=== true`, `?? ''`), and `unresolvedGateClause`'s `IS NULL` matches both, so omission is now the reset.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | On SQLite, gate 2's stored `signaledTokens` kept gate 1's `cacheRead`/`cacheWrite`/`cachePartial` for any axis gate 2 omitted; those counts flowed into the node total, `total_cache_read_tokens`/`total_cache_write_tokens`, and telemetry | Gate 2 stores only its own usage; Postgres behaviour is unchanged |
| **Failure behavior** | Silent — the numbers looked plausible and skewed cost in the unsafe direction | No failure path involved; the write is one statement, as before |

## Architecture

```mermaid
flowchart LR
  A[pauseWorkflowRun] --> B[writeApprovalMetadata]
  B ==>|"jsonb_set(… '{approval}' …)"| P[(Postgres)]
  B ==>|"json_set(… '$.approval' …)"| S[(SQLite)]
  A -->|top-level run metadata| M[jsonMerge semantics, unchanged]
```

### Changed seams

|---|---|---|
| `pauseWorkflowRun → workflow_runs.metadata` | `approval` is replaced wholesale instead of deep-merged; the payload splits into two bound params (run metadata, approval context) | `packages/core/src/db/workflows.ts:130-155,1073-1094`; `workflows.resume-cas.integration.test.ts` (#2673 block); `workflows.test.ts` `pauseWorkflowRun` suite |
| `ApprovalContext` optional fields | An unset field is now absent rather than an explicit JSON null | `packages/workflows/src/schemas/workflow-run.ts`; readers already accept both |

## Validation

- `bun run validate` — passed (exit 0) — full gate: bundled/schema/vendor-map/capability-matrix checks, type-check, `lint --max-warnings 0`, `format:check`, `test:install`, full test suite.
- `bun test src/db/workflows.resume-cas.integration.test.ts` (packages/core) — 28 pass, 0 fail — proves both behaviours against a real `bun:sqlite`.
- Same file run against the pre-fix `workflows.ts` — 2 fail, exactly the reported symptom: the cache-axis case received 2 extra keys (`cacheRead`, `cacheWrite`), the floor case 1 extra (`cachePartial`). The guards were seen red before they were seen green.
- `bun test src/db/workflows.test.ts` (packages/core) — 94 pass, 0 fail — this suite pins `getDatabaseType` to `postgresql`, so it exercises and asserts the Postgres branch of the new expression and the split params.
- Manual, PostgreSQL 18.0: `jsonb_set(metadata || $M::jsonb, '{approval}', $A::jsonb, true)` produces the same row today's `||` does for a two-gate sequence, and a NULL `metadata` still yields NULL — so the Postgres half is behaviour-preserving, not a behaviour change.
- Manual, `bun:sqlite` 3.51.0: the issue's reproduction confirmed before the fix; the new expression stores only the new gate's keys and preserves top-level merged keys.
- **Not verified:** Postgres behaviour by an automated test — the repository has no PG-backed test harness (`check:schema-upgrades` is a separate CI job with its own database). Covered by the manual run above plus the dialect-expression assertion in the mock suite.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. No schema change. Runs already paused keep their stored explicit nulls, and every reader treats null and absent identically, so an in-flight paused run resumes unchanged | `packages/workflows/src/dag-executor.ts:3600,3807,3845,4729`; `manage-run-tool.ts:278,284` |
| Rollout / rollback | One file's write expression; reverting the commit restores previous behaviour exactly | Single commit |

## Links

- Closes #2673
- Related #2671
- Plan: https://github.com/coleam00/Archon/issues/2673#issuecomment-5369434780

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workflow approval metadata replacement when a workflow pauses, preventing outdated gate, usage, or session details from carrying over.
  * Preserved unrelated workflow metadata while updating approval information.
  * Improved consistency across SQLite, PostgreSQL, and in-memory workflow execution.
  * Ensured fresh pauses reset approval and completion details correctly.
  * Improved handling of missing or explicitly cleared approval and rejection fields during workflow resume and rejection flows.

### DIFF
--- a/packages/core/src/db/workflows.resume-cas.integration.test.ts
+++ b/packages/core/src/db/workflows.resume-cas.integration.test.ts
@@ -11,6 +11,7 @@
  * ./connection with a real adapter, conflicting with workflows.test.ts's fake.
  */
 import { describe, test, expect, mock } from 'bun:test';
+import type { TokenUsage } from '@archon/providers/types';
 
 mock.module('@archon/paths', () => ({
   createLogger: () => ({
@@ -220,11 +221,11 @@ describe('claimWriteback — real SQLite CAS', () => {
 // ---------------------------------------------------------------------------
 // Gate approve/reject staging (#2075) — the run stays 'paused' after a gate
 // resolution instead of masquerading as 'failed'. This exercises the REAL
-// SQLite json_patch path end-to-end: staging write, resumable pickup (both the
+// SQLite JSON write path end-to-end: staging write, resumable pickup (both the
 // parent-conversation query the orchestrator uses and the working-path query
 // the CLI uses), the resume CAS, the double-resolution guard, and — critically
-// — that a fresh pause clears the previous gate's `resolved` marker despite
-// json_patch's deep-merge semantics.
+// — that a fresh pause carries nothing from the previous gate, which only a
+// real engine can prove (the mock suite runs the Postgres dialect).
 // ---------------------------------------------------------------------------
 
 /**
@@ -292,12 +293,12 @@ describe('gate approve staging — real SQLite end-to-end (#2075)', () => {
     await expect(resumeWorkflowRun('gate-1')).rejects.toThrow(WorkflowNotResumableError);
   });
 
-  test('a fresh pause clears the previous resolution despite json_patch deep-merge', async () => {
+  test('a fresh pause replaces the previous gate context, clearing its resolution', async () => {
     // Continue the run from the previous test: it is 'running' with
     // approval.resolved = 'approved' still in metadata (never cleared on
-    // resume by design). The next gate's pause MUST reset it — on SQLite
-    // json_patch deep-merges the new approval context into the old one, so an
-    // omitted key would leak the stale 'approved' and falsely block this gate.
+    // resume by design). The next gate's pause MUST reset it — the write
+    // replaces the whole approval object, so a key this gate does not set
+    // cannot carry the stale 'approved' over and falsely block it.
     await pauseWorkflowRun('gate-1', {
       nodeId: 'second-gate',
       message: 'Approve step 2?',
@@ -308,7 +309,6 @@ describe('gate approve staging — real SQLite end-to-end (#2075)', () => {
     expect(repaused?.status).toBe('paused');
     const approval = repaused?.metadata.approval as Record<string, unknown>;
     expect(approval.nodeId).toBe('second-gate');
-    // json_patch removes keys patched with null — 'resolved' must be gone/null.
     expect(approval.resolved ?? null).toBeNull();
 
     // And the second gate is approvable again.
@@ -316,6 +316,140 @@ describe('gate approve staging — real SQLite end-to-end (#2075)', () => {
     const staged = await getWorkflowRun('gate-1');
     expect((staged?.metadata.approval as Record<string, unknown>).resolved).toBe('approved');
     expect(staged?.status).toBe('paused');
+  });
+});
+
+// ---------------------------------------------------------------------------
+// #2673 — a later gate must not inherit an earlier gate's approval data. The
+// numeric half is the dangerous one: SQLite's json_patch is RFC 7396 and
+// recurses, so before the wholesale-replace write, interior keys of the nested
+// `signaledTokens` object survived from the previous gate and became real token
+// accounting on resume (readSignaledTokens in dag-executor). Only a real engine
+// can catch this; the mock suite runs the Postgres dialect, where `||` already
+// replaced the object.
+// ---------------------------------------------------------------------------
+describe('fresh gate usage is its own — real SQLite end-to-end (#2673)', () => {
+  /** Pause `runId` on a gate carrying `signaledTokens` and read the stored object back. */
+  async function pauseWithTokens(
+    runId: string,
+    nodeId: string,
+    signaledTokens: TokenUsage
+  ): Promise<Record<string, unknown> | undefined> {
+    await pauseWorkflowRun(runId, {
+      nodeId,
+      message: `gate ${nodeId}`,
+      type: 'interactive_loop',
+      iteration: 1,
+      signaledTokens,
+    });
+    const run = await getWorkflowRun(runId);
+    const approval = run?.metadata.approval as Record<string, unknown> | undefined;
+    return approval?.signaledTokens as Record<string, unknown> | undefined;
+  }
+
+  test('a second gate whose usage omits the cache axes does not inherit them', async () => {
+    await seedPausedRun('usage-cache', 'wf-usage-cache', { nodeId: 'seed', message: 'seed' });
+
+    // Gate 1: a provider that reports cache telemetry.
+    await resumeWorkflowRun('usage-cache');
+    const first = await pauseWithTokens('usage-cache', 'gate-1', {
+      input: 40,
+      output: 4,
+      cacheRead: 20,
+      cacheWrite: 3,
+    });
+    expect(first).toEqual({ input: 40, output: 4, cacheRead: 20, cacheWrite: 3 });
+
+    // Gate 2: a provider with no cache telemetry at all.
+    await resumeWorkflowRun('usage-cache');
+    const second = await pauseWithTokens('usage-cache', 'gate-2', { input: 90, output: 9 });
+
+    // Exactly the second gate's usage — no fabricated cache counts.
+    expect(second).toEqual({ input: 90, output: 9 });
+  });
+
+  test("a second gate's exact total is not flagged as a floor by the first gate", async () => {
+    await seedPausedRun('usage-partial', 'wf-usage-partial', { nodeId: 'seed', message: 'seed' });
+
+    // Gate 1 paused on a FLOOR (#2671): its cache totals are known-incomplete.
+    await resumeWorkflowRun('usage-partial');
+    await pauseWithTokens('usage-partial', 'gate-1', {
+      input: 10,
+      output: 1,
+      cacheRead: 50,
+      cacheWrite: 2,
+      cachePartial: true,
+    });
+
+    // Gate 2 knows its complete usage, so it declares no cachePartial.
+    await resumeWorkflowRun('usage-partial');
+    const second = await pauseWithTokens('usage-partial', 'gate-2', {
+      input: 90,
+      output: 9,
+      cacheRead: 5,
+      cacheWrite: 1,
+    });
+
+    expect(second).toEqual({ input: 90, output: 9, cacheRead: 5, cacheWrite: 1 });
+    expect(second?.cachePartial).toBeUndefined();
+  });
+
+  test('a gate that sets no optional fields stores none of them', async () => {
+    await seedPausedRun('usage-reset', 'wf-usage-reset', { nodeId: 'seed', message: 'seed' });
+
+    // A loop gate with the full optional payload.
+    await resumeWorkflowRun('usage-reset');
+    await pauseWorkflowRun('usage-reset', {
+      nodeId: 'loop-gate',
+      message: 'Review?',
+      type: 'interactive_loop',
+      iteration: 2,
+      sessionId: 'sess-1',
+      sessionProvider: 'claude',
+      completionSignaled: true,
+      signaledOutput: 'REPORT',
+      signaledTokens: { input: 40, output: 4, cacheRead: 20, cacheWrite: 3 },
+      signaledCostUsd: 0.02,
+      commandSnapshot: 'loop body',
+      onRejectPrompt: 'Fix: $REJECTION_REASON',
+      childRunId: 'child-1',
+    });
+
+    // A plain approval gate that declares none of them.
+    await resumeWorkflowRun('usage-reset');
+    await pauseWorkflowRun('usage-reset', {
+      nodeId: 'plain-gate',
+      message: 'Approve?',
+      type: 'approval',
+    });
+
+    const run = await getWorkflowRun('usage-reset');
+    expect(run?.metadata.approval).toEqual({
+      nodeId: 'plain-gate',
+      message: 'Approve?',
+      type: 'approval',
+    });
+  });
+
+  test('run-level metadata outside the approval object still merges', async () => {
+    await seedPausedRun(
+      'usage-extra',
+      'wf-usage-extra',
+      { nodeId: 'seed', message: 'seed' },
+      { rejection_count: 2 }
+    );
+
+    await resumeWorkflowRun('usage-extra');
+    await pauseWorkflowRun(
+      'usage-extra',
+      { nodeId: 'writeback', message: 'Apply changes?', type: 'writeback' },
+      { pending_writeback: true }
+    );
+
+    const run = await getWorkflowRun('usage-extra');
+    expect(run?.metadata.pending_writeback).toBe(true);
+    // Pre-existing top-level key untouched by the approval replace.
+    expect(run?.metadata.rejection_count).toBe(2);
   });
 });
 
--- a/packages/core/src/db/workflows.test.ts
+++ b/packages/core/src/db/workflows.test.ts
@@ -341,7 +341,7 @@ describe('workflows database', () => {
   });
 
   describe('pauseWorkflowRun', () => {
-    test('pauses a running run and resets the gate resolution marker', async () => {
+    test('replaces the approval object rather than merging into it', async () => {
       mockQuery.mockResolvedValueOnce(createQueryResult([], 1));
 
       await pauseWorkflowRun('workflow-run-123', {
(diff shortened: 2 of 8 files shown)


Wirasm:
## Ready to merge

Correction pass over `a77a7a81..7e85e131` (3 files, comment text only — every changed line is inside a `//` block, no production logic moved). `R1` is genuinely fixed this time, and it was checked against the code rather than against the delivery agent's reproduction. Both halves of the previously-false comment now hold: the consequence clause is the right way round (`null !== undefined` is `true`, so tightening reads a legacy null as *configured* and stages a rework the workflow never declared — it cannot cause the cancel the old text warned about), and the legacy-shape premise is now scoped per dialect instead of stated flat. That dialect split was proved this round on both engines, not one: `bun:sqlite` shows RFC 7396 `json_patch` deleting the null-patched `onRejectPrompt` in both the prior-gate and empty-metadata cases, and a **live PostgreSQL 18** shows `metadata || $2::jsonb` keeping the key with `jsonb_typeof = 'null'` in the same two cases. That closes the one check the previous round had to record as `NOT RUN`.

`R4` and `R5` are fixed on the same evidence standard. The `subrun.test.ts` double now claims only the write *order* and says outright that a JS spread cannot model either dialect's merge depth for `extraMetadata`; the `dag-executor.test.ts` comment no longer invents a resume-path null-vs-absent distinction and instead pins the payload the gate writes, which matches the pause site's `?? null` and the reader's `typeof … === 'string'`. `R2`'s decline is untouched by a comment-only diff and still holds. Nothing in the correction introduced a new defect. Full `bun run validate` passes in the review's own checkout at this head, and all five GitHub checks are green on `head_sha 7e85e131` — Windows included. Nothing blocks.

**Validation:** `bun run validate` PASS (exit 0) at `7e85e131` in a detached review checkout. GitHub CI PASS on all five checks at exactly this head.

**Resolved:** 5 (`R1` fixed, `R2` declined and upheld, `R3` fixed, `R4` fixed, `R5` fixed)

### Findings

|---|---|---|---|
| `R1` | Important | The reject-path comment inverted what tightening the check would do, and its legacy-null premise was Postgres-only | FIXED |
| `R2` | Suggestion | `resolveApprovalGate`'s "callers pass the full context" invariant is carried by a comment, not by a type | DECLINED |
| `R3` | Suggestion | `InMemoryStore` let `extraMetadata.approval` beat the fresh gate context; the real store no longer does | FIXED |
| `R4` | Suggestion | The double's "merges at the top level" claim did not hold for SQLite's recursive `json_patch` | FIXED |
| `R5` | Suggestion | A PR-authored test comment invented a resume-path null-vs-absent distinction the reader does not make | FIXED |

<details>
<summary><code>R1</code> — the reject-path comment is now true on both clauses and both dialects</summary>

**Impact:** Resolved. The code was always correct; what was wrong was the comment defending it, twice over. A maintainer reading the previous text would have found the stated consequence backwards and the stated premise false on the default backend — two independent invitations to delete the very check the comment exists to protect.

**Evidence:** `packages/core/src/operations/workflow-operations.ts:543-552` at `7e85e131`:

```ts
// `!= null` (not `!== undefined`): "no on_reject" reaches this read in two stored
// shapes. Absent — every pause since #2673 (the approval object is replaced
// wholesale, so an unset field is simply not there), and every SQLite pause before
// it too, since json_patch is RFC 7396 and DELETED the key the old explicit-null
// reset patched. An explicit JSON null — Postgres runs paused before #2673, where
// `||` stored the null as written. Keep the loose check: `null !== undefined` is
// true, so tightening would read such a run as HAVING an on_reject and stage a
// rework the workflow never declared — and the resume path takes its prompt from
// `node.approval.on_reject` (dag-executor), which is exactly what is missing.
const onRejectConfigured = approval?.onRejectPrompt != null;
```

Every clause was checked independently rather than accepted from the delivery agent.

*Clause A — the consequence, now the right way round.* Direct execution: `null != null` is `false`, `null !== undefined` is `true`. Traced on a legacy Postgres row whose gate has no `on_reject`:

| check | `onRejectConfigured` | `willStageRework` | branch | outcome |
|---|---|---|---|---|
| `!= null` (today) | `false` | `false` | `resolveAndCancelApprovalGate` (`:583`) | run **cancelled** (correct) |
| `!== undefined` | `true` | `true` (`maxAttempts = null ?? 3`; `0+1 >= 3` is false) | `resolveApprovalGate` (`:574`) | run **stays paused** |

The comment's "stage a rework the workflow never declared" is exactly that row. Its closing clause is also correct: the resume path reads the rework prompt from the workflow definition, not the stored context — `if (rejectionReason !== '' && node.approval.on_reject)` at `packages/workflows/src/dag-executor.ts:6031` — and `node.approval.on_reject` is `undefined` for a gate that never declared one, so resume falls through to the standard gate branch and re-raises the same gate.

*Clause B — the premise, now scoped per dialect and proved on both engines.* The pre-#2673 write is `dialect.jsonMerge('metadata', 2)` over a payload containing `onRejectPrompt: approvalContext.onRejectPrompt ?? null` (`origin/dev:packages/core/src/db/workflows.ts:1046-1066`).

SQLite (`json_patch(metadata, $2)`, `packages/core/src/db/adapters/sqlite.ts:847-851`), run on `bun:sqlite` with that exact payload:

```
prior gate HAD an on_reject : {"approval":{"nodeId":"gate-b","message":"ok?","type":"approval"}}
  'onRejectPrompt' in approval ? false
fresh row, metadata = {}     : {"approval":{"nodeId":"gate-b","message":"ok?","type":"approval"}}
  'onRejectPrompt' in approval ? false
```

Postgres (`metadata || $2::jsonb`, `packages/core/src/db/adapters/postgres.ts:288-290`), run on a **live PostgreSQL 18.0** — the check the previous round could only record as `NOT RUN`:

```
merged  : {"approval": {..., "resolved": null, "onRejectPrompt": null, "onRejectMaxAttempts": null}}
key_present | stored_type   ->  t | null      (prior gate had an on_reject)
key_present_fresh           ->  t             (fresh row, metadata = {})
```

Both stored shapes the comment names are real, and each is attributed to the right dialect and vintage. The third clause — absent on *every* pause since #2673 — is confirmed at `packages/core/src/db/workflows.ts:1073-1105`: `writeApprovalMetadata` replaces `metadata.approval` wholesale on both dialects and `JSON.stringify` drops `undefined`-valued keys, so a field the caller left unset is genuinely not there.

**Required outcome:** None. Delivered.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:comment-analyzer` (all three re-verified the fix independently; all three ran or re-derived the dialect evidence themselves)

**Disposition:** FIXED at `7e85e131`. Recorded, not filed: the comment stops one step short of the full downstream consequence — a staged phantom rework also stamps `resolved: 'rejected'`, so `unresolvedGateClause()` (`packages/core/src/db/workflows.ts:119-128`) stops matching and every later approve or reject throws, wedging the run. Nothing it says is false, the operative instruction ("keep the loose check") is stated correctly, and the reason given is already sufficient to forbid the tightening. Not a finding.

</details>

<details>
<summary><code>R2</code> — <code>resolveApprovalGate</code>'s full-context requirement is comment-only</summary>

**Impact:** None reachable.

**Evidence:** Unchanged from `a77a7a81`, where every caller was re-enumerated independently: exactly four production call sites, all in `packages/core/src/operations/workflow-operations.ts` (`:475`, `:521`, `:577`, and `:583` `resolveAndCancelApprovalGate`, which takes no metadata parameter). `IWorkflowStore` (`packages/workflows/src/store.ts:196`) exposes `pauseWorkflowRun` but not `resolveApprovalGate`, so there is no second route across the package seam. The spread carries the whole stored object because `assertApprovable`/`assertRejectable` obtain it through a `typeof` type guard (`isApprovalContext`, `packages/workflows/src/schemas/workflow-run.ts:349-356`) over a bare `JSON.parse`, not a schema parse that would strip undeclared keys. The correction diff touches nothing on this path — its only edit in that file is a comment block three lines above the nearest call.

**Required outcome:** None. Attach a typed parameter to whatever work next touches this seam.

**Disposition:** DECLINED, upheld a second time. Typing it means a signature change plus four call-site edits for zero behaviour change, guarding a caller that does not exist.

</details>

<details>
<summary><code>R3</code> — the in-memory store double no longer mirrors write precedence</summary>

**Impact:** Resolved.

**Evidence:** `packages/workflows/src/subrun.test.ts:205-213` is byte-identical to `a77a7a81` — only the comment above it changed this round. It still spreads `...(extraMetadata ?? {})` before `approval: { ...approvalContext }`, matching both dialect expressions in `writeApprovalMetadata` (`packages/core/src/db/workflows.ts:148-155`): `jsonb_set(metadata || $2::jsonb, '{approval}', $3::jsonb, true)` and `json_set(json_patch(metadata, $2), '$.approval', json($3))`. Both apply the extras first and set `approval` last. `bun test src/subrun.test.ts`: 69 pass, 0 fail.

**Found by:** `coderabbitai` (round one), re-verified at this head by `prp-core:code-reviewer` and `prp-core:seam-analyzer`

**Disposition:** FIXED, unchanged.

</details>

<details>
<summary><code>R4</code> — the double's depth claim is gone</summary>

**Impact:** Resolved. The risk was a future regression test written against a double whose comment certified it faithful at a depth it does not reproduce.

**Evidence:** `packages/workflows/src/subrun.test.ts:200-204` now reads "Mirrors the real store's write ORDER … Only the order; a JS spread cannot model both dialects' merge depth for `extraMetadata`." The order half is true on both dialects (that is `R3`). The depth disclaimer is accurate and points at the real divergence: SQLite's branch is `json_set(json_patch(metadata, $2), …)`, and `json_patch` is the RFC 7396 recursive merge the sibling docblock at `packages/core/src/db/workflows.ts:136-141` warns about, while the double's `...(extraMetadata ?? {})` is shallow. Postgres's `||` is shallow like the spread, so only the SQLite side diverges — and the comment no longer papers over it.

**Required outcome:** None. Delivered.

**Found by:** `prp-core:comment-analyzer`, `prp-core:seam-analyzer`

**Disposition:** FIXED at `7e85e131`.

</details>

<details>
<summary><code>R5</code> — the invented resume-path distinction is gone</summary>

**Impact:** Resolved. A maintainer could have acted on the old text in either direction — preserving nulls elsewhere believing resume depended on it, or "fixing" resume to check `=== null`.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:21348-21354` now reads "Readers accept either shape (dag-executor restores only on `typeof … === 'string'`); this pins the payload the gate writes, not a distinction the resume path makes." Each half checks out. The reader is `typeof loopGateMeta.sessionId === 'string' && typeof loopGateMeta.sessionProvider === 'string'` (`packages/workflows/src/dag-executor.ts:3843-3846`), which rejects `null` and `undefined` identically, and the PR's own schema docstring agrees (`packages/workflows/src/schemas/workflow-run.ts:246-250`: "readers treat that exactly like an absent key"). The write side is real: the gate passes `sessionId: loopLastSequentialSession?.sessionId ?? null` and `sessionProvider: … ?? null` (`packages/workflows/src/dag-executor.ts:4299-4300`), and the test asserts exactly that shape — `'sessionId' in pauseCtx` is `true` and `pauseCtx.sessionId` is `null`. The comment now describes what the test pins instead of a reader behaviour that does not exist.

**Required outcome:** None. Delivered.

**Disposition:** FIXED at `7e85e131`.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | `R1` verified fixed, `R3` verified still fixed; no new findings |
| `seams` | `R1` verified fixed (own dialect repro), `R2` decline upheld, `R4` verified fixed; no new findings |
| `comments` | `R1`, `R4`, `R5` all verified fixed clause by clause; no new findings |

All three scopes were retained because each owns a finding under verification; all three independently reproduced the SQLite dialect evidence rather than citing the previous report. `simplify` ran in round one with no findings and owns none of the verified findings, so it was not repeated. `types` remains unselected: the correction changes no type declaration, signature, schema, or variant.

#### Validation

All commands run in a detached review checkout at `7e85e131` (`.worktrees/review-pr-2675`), dependencies installed with `bun install --frozen-lockfile`.

|---|---|---|
| `bun run validate` | PASS | exit 0 at `7e85e131`; the full 10-step gate, including `lint --max-warnings 0` and `format:check` — the two steps comment prose could plausibly invalidate — through to the trailing `bun test ./scripts/` leg (95 pass, 0 fail) |
| `gh pr checks 2675` | PASS | all 5 green — `test (ubuntu-latest)` 3m21s, `test (windows-latest)` 7m15s, `docker-build` 4m58s, `schema-upgrade` 42s, CodeRabbit. Run `32482898539` confirmed via API at `head_sha 7e85e131` |
| SQLite `json_patch` legacy-shape repro, `bun:sqlite` | PASS | exact pre-#2673 payload; `'onRejectPrompt' in approval` is `false` both with a prior on_reject gate and on an empty row — decisive for `R1` clause B, SQLite half |
| Live PostgreSQL 18.0, `metadata \|\| $2::jsonb` | PASS | `key_present = t`, `jsonb_typeof = 'null'` in both cases — decisive for `R1` clause B, Postgres half. **New this round**; the previous round recorded this as NOT RUN |
| `bun test src/subrun.test.ts` (packages/workflows) | PASS | 69 pass, 0 fail — the double whose comment `R3`/`R4` concern |
| `bun test src/operations/workflow-operations.test.ts` (packages/core) | PASS | 49 pass, 0 fail — the reject path `R1` concerns |
| `bun test src/db/workflows.test.ts` (packages/core) | PASS | 94 pass, 0 fail — pins the emitted write expression for both dialects |

#### Prior-round results preserved

The round-one and round-two validation at `b440f86e` and `a77a7a81` (full `bun run validate`, both focused core suites, the five acceptance criteria checked directly, and the SQL-correctness analysis of `convertPlaceholders`, the NULL-`metadata` case and the empty-merge case) still applies unchanged: this correction diff touches only comment text.

#### Recorded, not filed

Carried forward from the earlier rounds so the ledger stays complete; neither is caused or worsened by this PR.

- `packages/web/src/experiments/console/primitives/run.test.ts:267` is named `'explicit null resolved (fresh pause) keeps the gate pending'`. No pause site writes `resolved` any more, so that shape is now the legacy-row case. The file is untouched by the PR.
- The `extraMetadata` lane the PR deliberately left on `dialect.jsonMerge` carries exactly one production value and it is a nested object: `{ pending_writeback: { envId, summary? } }` (`packages/workflows/src/dag-executor.ts:8937-8941`). A re-raised writeback gate whose second raise has no `summary` would keep the stale one on SQLite and drop it on Postgres — #2673's bug class one level up. This behaves identically on `origin/dev`. Worth its own issue if it matters.

</details>

Issue #2614, opened 2026-08-18
feat(workflows): surface the per-run transcript — the append-only everything-log already half exists


Wirasm:
## Problem

There is no user-facing way to follow or reconstruct everything that happened in a run. Watching a live run today means either the foreground CLI stream (lost if you detach or come later) or `workflow get <id> --verbose` (DB events: node states and output previews, no tool activity).

What almost nobody knows: **an append-only per-run transcript already exists.** `~/.archon/workspaces/<project>/logs/<runId>.jsonl` records `workflow_start`, `node_start`, `assistant` text, full `tool` invocations with inputs, and `node_complete` — timestamped, append-only, exactly the artifact an operator or agent wants when asking "what did this run actually do" (writer: `packages/workflows/src/logger.ts`). But nothing surfaces it: no command prints its path, no command tails or renders it, and it lives under a workspace directory the user has to know to look in.

## Proposal (direction, deliberately not a spec — the exact shape is open)

Make the transcript first-class:

1. **Discoverable** — print its path at run start and in `workflow get`.
2. **Followable** — some form of `workflow logs <id> [--follow]` that streams/renders it, working for detached and CLI-launched runs alike (the DB-tail concern from the console applies here too).
3. **Complete** — audit coverage so it is genuinely the everything-log: approval gate events, node failures with stderr, skips and their reasons, loop iterations, per-node usage/cost, resume boundaries. Whatever is missing should be added to the writer, not to a second log.

## Acceptance (minimum)

- An operator can find and follow a run's transcript from its run id alone, live and after the fact.
- The transcript records every node lifecycle event including failures, skips, gates, and resumes.

Wirasm:
## Concrete gap for the "Complete" leg: the transcript has tokens but no cost

Found while landing #2654 (cache token usage). Recording it here rather than opening a separate issue, since the proposal above already names per-node usage/cost and says missing fields belong in this writer rather than a second log.

`WorkflowEvent` in `packages/workflows/src/logger.ts` carries `tokens?: WorkflowTokenUsage` but **no cost field on any event type**. `logNodeComplete` takes `meta?: { durationMs, tokens }` and nothing else. So the JSONL transcript records how many tokens a node used but never what it cost, even though the executor already has `cost_usd` at that point and persists it to the DB `node_completed` event.

[...]

Issue #2334, opened 2026-07-30
fix(providers/codex): Codex nodes record only duration — no cost, turns, or stop reason, and nothing warns


Wirasm:
## Summary

A Codex node records **only `duration_ms`**. Claude and Pi nodes additionally record cost and stop reason. Switching an install's model tiers from Claude to Codex therefore silently reduces what every run can report, with no warning at any layer.

## Observed

Persisted `node_completed` payload keys, same workflow, same node, different provider:

```
Codex   (gpt-5.6-luna)   ['duration_ms', 'node_output']
Claude  (sonnet)         ['duration_ms', 'node_output', 'cost_usd', 'num_turns', 'stop_reason']
```

Verified across all 36 `node_completed` events in a 3-run window on Codex tiers: **zero** carry cost, turns, or stop reason.

Nothing warns. `capabilities.ts` drives ignored-capability warnings for node *features*, but there is no signal for "this provider will report less about itself".

## What is a genuine SDK limitation vs an Archon omission

Read against `@openai/codex-sdk@0.125.0` types, so the two are not conflated:

| Metric | Codex SDK provides | Archon maps | Verdict |
| --- | --- | --- | --- |
| **tokens** | `Usage { input_tokens, cached_input_tokens, output_tokens, reasoning_output_tokens }` on `TurnCompletedEvent` | yes — `codex/provider.ts:667` | ✅ works (but lands in the JSONL log only — see #2333) |
| **cost** | nothing. No cost field anywhere | n/a | **derivable, not reportable** — see below |
| **stop reason** | `TurnCompletedEvent = { type, usage }` — no reason for a *successful* turn | n/a | genuine SDK gap |
| **turn count** | nothing | n/a | genuine SDK gap |
| **resolved model** | nothing; `model` appears only on `ThreadOptions` (input) | n/a | genuine SDK gap — #2314 |

So most of this is upstream, and **Archon is not failing to read something that exists** — with one exception.

## The exception: cost is derivable

Codex reports full token counts including cached-input and reasoning-output. Cost is `tokens × price`. Pi already carries cost because its SDK computes it (`pi/event-bridge.ts:167`); Claude's SDK reports `total_cost_usd` directly. Codex is the only provider where Archon would have to do the arithmetic itself.

That is a real decision, not an obvious win:

- **For:** cost is the metric an operator actually wants, and it is the one axis on which Codex is currently unmeasurable while every other provider is measurable. It makes `archon ai tier` choices comparable.
[...]

PR #2676, opened 2026-08-21
feat(workflows): the run transcript now records what each node and the run cost

Wirasm:
## Problem and outcome

The per-run JSONL transcript at `~/.archon/workspaces/<project>/logs/<runId>.jsonl` recorded how many tokens a node used but never what it cost. Anyone reconstructing spend from it had to price four token axes by hand and get the cache rates right, and `tokens.input` is **gross** prompt input (cache reads and writes are components of it, not additions to it), which is exactly the part that is easy to get wrong. The executor already held the provider-reported cost at that point and wrote it to the DB `node_completed` event three lines later.

- **Outcome:** `node_complete` rows carry `cost_usd`, and the `workflow_complete` row carries the run totals as `cost_usd` and `tokens`. Reading spend out of a transcript no longer involves arithmetic.
- **Invariant:** A transcript row reports the same usage as the DB event for that same unit of work, and an unreported cost stays **absent** rather than becoming `0`. Codex reports no cost at all (#2334), so `0` has to keep meaning "the provider reported zero".
- **Scope boundary:** Only the cost gap named in [this comment](https://github.com/coleam00/Archon/issues/2614#issuecomment-5368036624). No transcript path printing, no `workflow logs <id> --follow`, and no other audit-coverage work. #2614 stays open for those.

## Review guidance

- **Feedback requested:** Correctness of the absent-vs-zero handling at both levels, and whether the shared carrier is the right shape for the seam.
- **Start here:** `packages/workflows/src/logger.ts:52-63` — the `WorkflowUsage` carrier. Everything else follows from it.
- **Review order:** `logger.ts` (field + carrier + two writers) → `dag-executor.ts:1745` (`nodeUsageEventData` now typed as the carrier) → `dag-executor.ts:2702` and `:2715` (both sinks spread the same value) → `dag-executor.ts:9685` (run totals) → the two test files.
- **Lower-attention areas:** The docs paragraph is an addition to the existing token-contract paragraph; the surrounding wording is untouched.
- **Known risk or uncertainty:** The run level uses `totalCostUsd > 0` while the node level uses `!== undefined`. That asymmetry is deliberate and commented: the run accumulator is a plain number seeded at `0` and cannot express "unreported", and `persistRunUsage` already made this exact call for the run row. Disagreeing would make the transcript and the run row report different things.

## Solution

The bug was not a missing field, it was a missing carrier. Inside `executeNodeInternal` the same two-axis usage payload was spelled out by hand three separate times: once in `nodeUsageEventData()` for the failure paths, once inline on the success DB event, and once more, minus cost, in the transcript call. The transcript spelling is the one that forgot cost.

So this adds one shape instead of one field. `WorkflowUsage` is `Pick<WorkflowEvent, 'tokens' | 'cost_usd'>` in `logger.ts`: the transcript's own field names, which are also the DB event's, so nothing translates at the boundary. `nodeUsageEventData()` is now typed as that carrier and spread whole into both sinks, which deleted the inline copy on the DB event. `logNodeComplete`'s meta becomes `{ durationMs?: number } & WorkflowUsage` and `logWorkflowComplete` gains an optional `usage` parameter, both spreading the carrier rather than re-deciding per field. A future usage axis now reaches the transcript and the DB event together or reaches neither.

The carrier omits an axis at construction when nothing was reported for it, so the writers need no guard. This matters: the existing writer used a *truthy* test for tokens, and copying that idiom for cost would have turned a genuine `$0` into an absent field.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `node_complete` had `duration_ms` and `tokens`; `workflow_complete` had nothing but its type and timestamp | `node_complete` also has `cost_usd`; `workflow_complete` has `cost_usd` and `tokens` for the whole run |
| **Failure behavior** | A provider that reported no cost was indistinguishable from one that reported `$0`, since neither reached the file | An unreported cost writes no key at all; a reported `0` writes `0`. A run with no AI spend (bash-only) writes no run-level `cost_usd`, matching the run row |

## Architecture

```mermaid
flowchart LR
  N[node result: nodeTokens + nodeCostUsd] --> C["nodeUsageEventData(): WorkflowUsage"]
  C ==> T["JSONL node_complete"]
  C ==> D["DB node_completed.data"]
  R[runCtx totals] --> W["JSONL workflow_complete"]
  R --> P["run row metadata (persistRunUsage)"]
```

### Changed seams

|---|---|---|
| `executeNodeInternal → JSONL transcript` | `logNodeComplete` meta widened from `{ durationMs, tokens }` to `{ durationMs } & WorkflowUsage`; the caller passes the same carrier the DB event gets | `packages/workflows/src/dag-executor.ts:2702`, `packages/workflows/src/logger.ts:198-213` |
| `executeNodeInternal → DB node_completed` | Inline `tokens` / `cost_usd` re-listing replaced by the carrier; values unchanged | `packages/workflows/src/dag-executor.ts:2715` |
| `executeDagWorkflow → JSONL transcript` | `logWorkflowComplete` gains an optional `usage` argument carrying the run totals | `packages/workflows/src/dag-executor.ts:9685`, `packages/workflows/src/logger.ts:172-181` |
| `WorkflowEvent` row shape | New optional `cost_usd`; purely additive to an append-only file | `packages/workflows/src/logger.ts:45` |

## Validation

- `bun run validate` — passed (exit 0) — all 11 packages type-check clean, lint at `--max-warnings 0`, format clean, and every package suite green.
- `bun test src/logger.test.ts` (in `packages/workflows`) — 19 pass, 0 fail — proves the writer records a reported cost, keeps a reported `0` as `0`, omits the key entirely when nothing was reported, and writes run-level usage only when supplied.
- `bun test src/dag-executor.test.ts` (in `packages/workflows`) — 560 pass, 0 fail — the two new cases run a real DAG against a mocked provider, read the actual `.jsonl` off disk, and assert the node row's `cost_usd` equals the DB event's, plus the run row's total.
- Each new assertion was **observed failing** against a deliberately broken implementation before being trusted: reverting the transcript call to `tokens: nodeTokens` failed with `Expected: 0.3, Received: undefined`; writing `nodeCostUsd ?? 0` failed the absent-key test with `Received value: 0`; writing the run cost unconditionally failed the same test on the `workflow_complete` row; guarding the writer with truthiness instead of spreading the carrier failed the reported-zero test with `Expected: 0, Received: undefined`. All four mutations were reverted and the suites re-run green.
- **Not verified:** No live end-to-end run against a real provider. The transcript rows are asserted from the real file written by the real writer through the real executor path, with only the provider stream mocked, so the untested remainder is the provider's own cost reporting, which this PR does not change.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Purely additive to an append-only log. Rows written before this change simply lack the key, and the only reader in the repo parses per line by field name. No schema change, so no migration or `check:schema-upgrades` involvement. | `packages/workflows/src/logger.test.ts:52-58` is the sole transcript reader |
| Rollout / rollback | Single-commit revert; no stored state depends on the new field. | — |
| Documentation / communication | The JSONL usage contract paragraph now states the cost contract alongside the token one, including absent-vs-zero. | `packages/docs-web/src/content/docs/guides/authoring-workflows.md:1541-1548` |

### Known remaining gaps in the same leg

Both are deliberate and stay with the open issue rather than widening this PR:

- Failure rows carry no usage at either level. `logNodeError` and `logWorkflowError` take none, so a node (or a run) that spent money and then failed records that spend on the DB event and the run row but not in the transcript. The comment above `nodeUsageEventData()` now says exactly that rather than claiming every sink is covered.
- `loop:` per-iteration rows still carry no usage. They carry no tokens today either, so nothing new is asymmetric, and a correct per-iteration figure is not a straightforward carry: `iterationCost` holds only the last attempt's value and reasked attempts fold separately into `loopTotalCostUsd`. The loop node's spend still reaches the run total.

## Links

- Related #2614 — this delivers only the cost gap from that issue's "Complete" leg. Part of #2614, not a fix for it; the issue stays open for transcript discoverability and `workflow logs --follow`.
- Implementation plan: https://github.com/coleam00/Archon/issues/2614#issuecomment-5369474743

## Summary by CodeRabbit

- **Improvements**
  - Workflow and node transcripts now consistently record provider-reported token usage and costs.
  - Explicit zero-cost results are preserved, while unavailable cost data remains omitted.
  - Workflow completion records include aggregate usage details when available.
  - Cost totals are surfaced consistently across JSONL logs, persisted events, and run metadata.

- **Documentation**
  - Added guidance on using provider-reported costs for workflow and node accounting.
  - Documented cost-data locations, absent-versus-zero behavior, and bash-only workflow considerations.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -1538,6 +1538,15 @@ cacheWrite` is exact when `cachePartial` is absent and an upper bound on full-pr
 when it is set. An axis no node reported stays absent, and cache usage in records created
 before this contract cannot be recovered.
 
+Read cost from `cost_usd`, not from the token counts. Because `input` is gross, pricing a
+node by hand means getting four axes and the cache rates right; `cost_usd` is the number the
+provider itself reported. JSONL `node_complete.cost_usd` and persisted
+`node_completed.data.cost_usd` carry it for a node, and JSONL `workflow_complete` carries the
+run totals as `cost_usd` and `tokens`, matching run metadata `total_cost_usd` and
+`total_tokens_*`. An absent `cost_usd` means the provider reported no cost — Codex reports
+none — while `0` means it reported zero. A run that spent nothing on AI, such as a bash-only
+workflow, carries no `cost_usd` rather than `0`.
+
 ### Choosing the child's checkout with `isolation:`
 
 `isolation:` decides which working directory the child run executes in. It is valid **only**
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -5967,6 +5967,106 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
     expect(completedEvent?.[0].data).not.toHaveProperty('tokens');
   });
 
+  const readTranscript = async (
+    logDir: string,
+    runId: string
+  ): Promise<Array<Record<string, unknown>>> =>
+    (await readFile(join(logDir, `${runId}.jsonl`), 'utf-8'))
+      .trim()
+      .split('\n')
+      .map(line => JSON.parse(line) as Record<string, unknown>);
+
+  it('writes node and run cost to the transcript, matching the persisted events', async () => {
+    mockSendQueryDag.mockImplementation(async function* () {
+      yield { type: 'assistant', content: 'out' };
+      yield {
+        type: 'result',
+        sessionId: 'cost-sid',
+        cost: 0.3,
+        tokens: { input: 100, output: 10 },
+      };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const logDir = join(testDir, 'logs');
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-transcript-cost',
+      testDir,
+      { name: 'transcript-cost', nodes: [{ id: 'step1', command: 'step1' }] },
+      makeWorkflowRun('transcript-cost-run'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      logDir,
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const transcript = await readTranscript(logDir, 'transcript-cost-run');
+    const nodeRow = transcript.find(e => e.type === 'node_complete' && e.step === 'step1');
+    const runRow = transcript.find(e => e.type === 'workflow_complete');
+
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls as Array<
+      [{ event_type: string; step_name: string; data?: Record<string, unknown> }]
+    >;
+    const completedEvent = eventCalls.find(
+      ([event]) => event.event_type === 'node_completed' && event.step_name === 'step1'
+    );
+
+    expect(nodeRow?.cost_usd).toBe(0.3);
+    expect(nodeRow?.tokens).toEqual({ input: 100, output: 10 });
+    // Both sinks read the one carrier, so the transcript can never disagree with the
+    // event the console and `workflow get` render.
+    expect(nodeRow?.cost_usd).toBe(completedEvent?.[0].data?.cost_usd);
+    expect(runRow?.cost_usd).toBe(0.3);
+    expect(runRow?.tokens).toEqual({ input: 100, output: 10 });
+  });
+
+  it('leaves cost off the transcript when the provider reports none', async () => {
+    mockSendQueryDag.mockImplementation(async function* () {
+      yield { type: 'assistant', content: 'out' };
+      // Codex reports no cost at all (#2334). Writing 0 here would claim the node was
+      // free, and a reader summing the transcript would believe it.
+      yield { type: 'result', sessionId: 'no-cost-sid', tokens: { input: 100, output: 10 } };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const logDir = join(testDir, 'logs');
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-transcript-no-cost',
+      testDir,
+      { name: 'transcript-no-cost', nodes: [{ id: 'step1', command: 'step1' }] },
+      makeWorkflowRun('transcript-no-cost-run'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      logDir,
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const transcript = await readTranscript(logDir, 'transcript-no-cost-run');
+    const nodeRow = transcript.find(e => e.type === 'node_complete' && e.step === 'step1');
+    const runRow = transcript.find(e => e.type === 'workflow_complete');
+
+    expect(nodeRow).toBeDefined();
+    expect(nodeRow).not.toHaveProperty('cost_usd');
+    expect(nodeRow?.tokens).toEqual({ input: 100, output: 10 });
+    expect(runRow).not.toHaveProperty('cost_usd');
+  });
+
   // ─── Background Agent Task Gating (#2083) ───────────────────────────────
 
   describe('background task completion gating (#2083)', () => {
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -6065,6 +6065,9 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
     expect(nodeRow).not.toHaveProperty('cost_usd');
     expect(nodeRow?.tokens).toEqual({ input: 100, output: 10 });
     expect(runRow).not.toHaveProperty('cost_usd');
+    // The two run-level axes are decided independently, so a missing cost must not
+    // suppress the tokens that were reported.
+    expect(runRow?.tokens).toEqual({ input: 100, output: 10 });
   });
 
   // ─── Background Agent Task Gating (#2083) ───────────────────────────────
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -1740,8 +1740,11 @@ async function executeNodeInternal(
   let nodeResolvedModel: ResolvedModel | undefined;
   const batchMessages: string[] = [];
 
-  // What this node reported, built once and passed whole to every sink — the DB event
-  // on each outcome and the JSONL transcript row. See WorkflowUsage for why.
+  // What this node reported, built once and passed whole rather than re-listed per sink:
+  // the DB event on every outcome, and the JSONL transcript row on completion. The JSONL
+  // FAILURE row is still outside it — logNodeError takes no usage, so a node that spent
+  // money and then failed records that spend in the DB and not in the transcript (#2614).
+  // See WorkflowUsage.
   const nodeUsageEventData = (): WorkflowUsage => ({
     ...(nodeTokens !== undefined ? { tokens: nodeTokens } : {}),
     ...(nodeCostUsd !== undefined ? { cost_usd: nodeCostUsd } : {}),
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -108,6 +108,7 @@ import {
   logTool,
   logWorkflowComplete,
   logWorkflowError,
+  type WorkflowUsage,
 } from './logger';
 import { withIdleTimeout, STEP_IDLE_TIMEOUT_MS } from './utils/idle-timeout';
 import { mapWithLimit } from './utils/map-with-limit';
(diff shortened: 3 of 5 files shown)


Wirasm:
## Ready to merge

Correction pass over `f6b07bf3..ca0e0a30`, one commit, two files, 8 insertions and 2 deletions: a comment rewrite above `nodeUsageEventData()` and one added test assertion. No production logic changed, so nothing about the original verdict's basis moved. Both open findings are now closed on verified evidence, and the declined one is confirmed still unreachable.

R1 is the substantive one, and it was fixed the way the finding asked rather than papered over. The old comment claimed the carrier is "passed whole to every sink", which was false. The replacement states three things, and every clause was checked independently against the code at this head rather than against the prior review's line numbers. The carrier does reach the DB event on all five post-query outcome branches (`:2551`, `:2604`, `:2650`, `:2718`, `:2796`); it does reach the JSONL row only on completion (`:2705`, the sole `logNodeComplete` call inside `executeNodeInternal` — the other three call sites belong to bash, script, and loop nodes, outside this closure); and `logNodeError` does still take no usage. The comment now names the gap it used to hide and points at the still-open #2614. Two agents also confirmed the paid-then-failed state the comment describes is genuinely reachable, not a hypothetical: `nodeCostUsd` is set on the `result` message before the budget-cap and SDK-error throws that follow it, so the value survives into the catch block. A comment that describes an impossible case would have been a new defect; this one does not.

R2's assertion was verified load-bearing rather than taken on the author's word. The reviewer reproduced the mutation independently — gating the run-level `tokens` spread on `totalCostUsd > 0` — and saw the new line fail with `Expected {input:100,output:10}, Received undefined` at `dag-executor.test.ts:6070`, then reverted and saw green. The expectation also matches what that test's mock actually yields, so it is not a copy-paste that happens to pass.

**Validation:** `bun run validate` passed (exit 0) at `ca0e0a30` in this review's own clean detached checkout.

**Resolved:** 2 (`R1`, `R2`)

### Findings

|---|---|---|---|
| `R1` | Suggestion | JSONL failure rows carry no usage while their DB siblings do, and the carrier comment read as if every sink is covered | FIXED |
| `R2` | Suggestion | The no-cost executor test did not assert the run row still carries `tokens` | FIXED |
| `R3` | Suggestion | The `{ durationMs?: number } & WorkflowUsage` rest-spread would leak a future unrelated meta field into the row | DECLINED |

<details>
<summary><code>R1</code> — failure-outcome JSONL rows are outside the carrier, and the comment did not say so</summary>

**Impact:** A node that accrues real spend and then fails records `cost_usd` and `tokens` on its DB `node_failed` event and records neither on its JSONL `node_error` row. Anyone reading spend off the transcript alone undercounts a failed-but-paid node silently. The comment above the carrier claimed every sink was covered, and the PR body's remaining-gaps list named only the `workflow_error` analogue.

**Evidence:** At `f6b07bf3`, `dag-executor.ts:1743-1744` read "built once and passed whole to every sink". At `ca0e0a30`, `dag-executor.ts:1743-1747` reads:

```
// What this node reported, built once and passed whole rather than re-listed per sink:
// the DB event on every outcome, and the JSONL transcript row on completion. The JSONL
// FAILURE row is still outside it — logNodeError takes no usage, so a node that spent
// money and then failed records that spend in the DB and not in the transcript (#2614).
// See WorkflowUsage.
```

Each clause verified at this head: `...nodeUsageEventData()` is spread into the DB event on all five post-query outcome branches — cancelled-during-streaming (`:2551`), credit exhaustion (`:2604`), empty output (`:2650`), success (`:2718`), catch-all (`:2796`) — with no remaining hand-listed copy at any of them; `logNodeComplete` receives it exactly once, at `:2705`, on the success path only; `logNodeError` (`logger.ts:232-241`) still takes `(logDir, workflowRunId, nodeId, error: string)` with no usage parameter, and all three of its call sites inside `executeNodeInternal` (`:2597`, `:2640`, `:2788`) pass only the error string. The paid-then-failed state is reachable: `nodeCostUsd`/`nodeTokens` are set on the `result` message before the budget-cap check and the generic SDK-error check, both of which throw into the outer catch where the value still holds. `#2614` is open and its acceptance criteria own this gap. `WorkflowUsage` (`logger.ts:52-63`) exists and its docstring does explain the "why" the comment defers to it.

The PR body's second half also landed. "Known remaining gaps in the same leg" now reads "Failure rows carry no usage at either level. `logNodeError` and `logWorkflowError` take none…", covering the node-level case alongside the run-level one it previously named alone. `logWorkflowError` (`logger.ts:158-166`) was confirmed to likewise take no usage, so the widened claim is accurate in both halves.

**Required outcome:** Satisfied. Wiring `WorkflowUsage` into `logNodeError` remains the real fix and belongs with the still-open #2614.

**Disposition:** FIXED. Both halves of the requested correction are present and each factual clause was re-derived from the code at `ca0e0a30` by two agents working independently (`prp-core:seam-analyzer` and `prp-core:comment-analyzer`), reaching the same enumeration. Neither found a clause that overclaims or underclaims.

</details>

<details>
<summary><code>R2</code> — the no-cost executor test left the run-level token branch unasserted</summary>

**Impact:** The negative test proved the run row omits `cost_usd` when no cost was reported but never checked that the run row still carries `tokens`. The two run-level branches are independent (`totalCostUsd > 0` at `dag-executor.ts:9692` and `totalTokens !== undefined` at `:9693`, over two separately accumulated variables), so an implementation that suppressed run usage wholesale when cost was absent would have passed.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:6070` now adds `expect(runRow?.tokens).toEqual({ input: 100, output: 10 })` to the `leaves cost off the transcript when the provider reports none` case, under a comment stating why. The value matches what that test's mocked provider yields (`dag-executor.test.ts:6032-6037`); the workflow has one node, so the run total equals the node figure.

The assertion was independently proved load-bearing, not accepted on report. Gating the `logWorkflowComplete` tokens spread on `totalCostUsd > 0` produced:

```
- { "input": 100, "output": 10 }
+ undefined
      at dag-executor.test.ts:6070:28
(fail) leaves cost off the transcript when the provider reports none
```

The mutation was reverted, the test reran green, and `git status --porcelain` on the review checkout was confirmed empty afterwards.

**Required outcome:** Satisfied.

**Disposition:** FIXED. Seen red under the reviewer's own mutation, not the author's.

</details>

<details>
<summary><code>R3</code> — a future meta field would leak through the rest-spread</summary>

**Impact:** `logNodeComplete` destructures `const { durationMs, ...usage } = meta ?? {}` and spreads `usage` into the event. TypeScript's excess-property check does not survive an object spread, so widening the meta type later with a non-usage field would reach the JSONL row unfiltered with no compile error.

**Evidence:** `packages/workflows/src/logger.ts:203-217`. Confirmed unreachable at this head: `logger.ts` is byte-identical between `f6b07bf3` and `ca0e0a30` — same blob hash `ddcf18e2c1d8f31fdaf8b4190e213bdb97c32229` at both commits, and `git diff f6b07bf3..ca0e0a30 -- packages/workflows/src/logger.ts` is empty. `WorkflowUsage` is still `Pick<WorkflowEvent, 'tokens' | 'cost_usd'>` with no third axis, `logNodeComplete`'s `meta?: { durationMs?: number } & WorkflowUsage` is unwidened, and no call site was added that passes a non-usage field.

**Required outcome:** None now.

**Disposition:** DECLINED, unchanged and re-verified at this head. Exploiting it still requires a change nobody has made, and the rest-spread is what keeps this writer from re-listing `tokens` and `cost_usd` by hand — the exact defect the PR removes. Replacing it with an explicit per-field pick would reinstate that defect.

</details>

<details>

#### Reviewer coverage

Correction-verification pass bounded to `f6b07bf3..ca0e0a30`. Scopes retained because they own a finding under verification; `comments` added because the correction is substantially a comment rewrite, which is that agent's exact defect class.

| Scope | Result |
|---|---|
| `seams` (`prp-core:seam-analyzer`) | `R1` FIXED; no correction-caused finding |
| `code` (`prp-core:code-reviewer`) | `R2` FIXED, mutation reproduced; no correction-caused finding |
| `types` (`prp-core:type-design-analyzer`) | `R3` DECLINED holds; no correction-caused finding |
| `comments` (`prp-core:comment-analyzer`) | All 7 clauses of the rewritten comment plus the added test comment verified accurate; no findings |
| `simplify` (`prp-core:code-simplifier`) | Not re-run — returned no findings at `f6b07bf3` and owns no finding under verification |

Points checked and cleared:

- **The two pre-query `node_failed` events do not falsify "the DB event on every outcome".** `dag-executor.ts:1650` (command-load failure) and `:1700` (prompt-substitution failure) write a DB node event without the carrier, but both fire lexically before `nodeTokens`/`nodeCostUsd` are declared at `:1736-1737` and before any provider call, so no spend can exist. The comment is scoped by its placement directly above the carrier. Raised and cleared independently by `prp-core:seam-analyzer` and `prp-core:comment-analyzer`; neither treated it as a finding.
- **The three `logNodeComplete` call sites outside this closure are not a counterexample either.** `:3077` (bash), `:3465` (script) and `:5762` (loop) belong to different functions and are out of the commented closure's reach.
- **The scope boundary holds.** `closingIssuesReferences` on the PR is empty, no closing keyword (`close`/`fix`/`resolve`) appears anywhere in the body, the Links section still reads "Related #2614 — this delivers only the cost gap from that issue's 'Complete' leg", and issue #2614 is OPEN. Merging will not auto-close it.

#### Validation

|---|---|---|
| `bun run validate` | PASS | exit 0 at `ca0e0a30` in `.worktrees/review-pr-2676`, a detached checkout created by this review; tree verified clean (`git status --porcelain` empty) immediately before the run |
| `bun test src/dag-executor.test.ts` (in `packages/workflows`) | PASS | 560 pass, 0 fail |
| `bun test src/dag-executor.test.ts -t "leaves cost off the transcript when the provider reports none"` under mutation | FAIL as intended | new assertion at `:6070` failed `Expected {input:100,output:10}, Received undefined`; reverted, reran green |
| `bun x tsc --noEmit` (`@archon/workflows`) | PASS | clean |
| `bun run check:schema-upgrades` | NOT RUN | needs a live PostgreSQL and is a separate CI job. Not applicable: the correction diff touches no SQL and adds no column |

</details>

Issue #2512, opened 2026-08-07
feat(workflows): add runtime per-item parallel composition inside the parent run


Wirasm:
> **Body rewritten 2026-08-13.** This issue is the runtime-width extension of #1764. It no longer treats one governed child run per item as the desired composition model.

## Problem

#1764 defines one statically named composed workflow executing inside the parent run. The remaining gap is runtime width: an upstream node produces a list whose length is unknown at load time, and the author needs to execute the same static body once per item with bounded parallelism.

Current `workflow:` `fan_out` creates a separate governed child run per array index. Runtime cardinality does not itself justify separate run identity. Per-item composition belongs to the parent run unless the author explicitly chooses a future governed-launch primitive.

## Target semantics

A fan-out combines:

- one statically known composed body/workflow target;
- one runtime array supplied as data;
- a named per-item input binding plus optional static inputs;
- bounded concurrency;
- deterministic instance-qualified identities in the parent event stream;
- an ordered aggregate output.

Every item executes under the parent `workflow_runs` row, artifact root, gates, lifecycle, cancellation, resume, and accounting. No item creates a child run.

Declared inputs and outputs carry item values and results. Shared artifacts remain available for files and large/file-shaped results, with collision-safe engine-managed paths; artifacts are not the required item/result transport.

The repeated body remains command/script-first: each instance may run package-local command prompt files and named deterministic script files. Runtime item bindings must reach those files without forcing authors to inline substantial prompts or programs in YAML.

## Illustrative YAML

Field spelling is non-final. Runtime data controls item count, never the body structure.

```yaml
nodes:
  - id: discover
    prompt: Return the work items as JSON.

  - id: process
    workflow: process-item
    with:
      mode: reconcile
    fan_out:
      items: $discover.output.items
      as: item
      max_parallel: 5
      join: all_done

  - id: summarize
    prompt: Summarize $process.output
    depends_on: [process]
```

## Current behavior being replaced

Current fan-out already supplies useful mechanics: runtime arrays, bounded concurrency, ordered aggregation, `all_done`/`all_success`, explicit checkout-mutation preflight, and resumable item bookkeeping.

It currently stores those mechanics in child rows:

- one child `workflow_runs` row and artifact directory per item;
- child events and lifecycle per item;
- resume identity based primarily on `child_index`;
- an item hash used for drift warnings rather than identity;
- optional child worktrees;
- aggregate results reduced from child summaries.

[...]

PR #2692, opened 2026-08-21
feat(workflows): structured values survive every boundary, and command/script nodes bind them by name

### LINKED ISSUE #2637: feat(workflows): preserve logical values and add explicit node-local bindings
## Problem

Archon has the beginnings of structured workflow values, but the logical value is lost at important boundaries:

- structured node output is flattened to text before persistence;
- cold resume rehydrates text rather than the original logical value;
- whole-value references and aggregates can become stringified or double-encoded;
- declared inputs are string-only, so booleans and numbers lose their type;
- a named command or script cannot receive an upstream value through an explicit node-local binding, forcing workflows such as PR #2627 to bridge a one-bit result through an artifact file.

The full declared input/output contract remains tracked by #2453. This issue is the smallest end-to-end value-transport slice required before #2512 adds runtime-width composition.

## Target semantics

Workflow values are JSON-compatible logical values: string, number, boolean, null, array, or object.

- A whole-value reference preserves the logical value and type.
- Embedding a value inside surrounding text is interpolation and uses one deterministic textual representation: strings remain raw; other JSON values use canonical JSON text.
- Node completion persistence stores enough information to reconstruct the same logical value after a cold resume.
- Existing textual output remains available for compatibility and human-readable surfaces.
- Aggregation operates on logical values and serializes only at a textual boundary.

## Node-local bindings

Resource-backed nodes may explicitly bind named values from upstream outputs:

```yaml
- id: check
  script: check-result.ts
  depends_on: [verify]
  with:
    green: $verify.output.green
```

Reuse the existing input channels:

- command files read bound values through `$INPUTS.<name>`;
- named scripts read them through `INPUTS_<UPPER_SNAKE>`;
- no new `$INPUT`, `$input`, or `$PREVIOUS` channel is introduced.

A binding that references another node must refer to an upstream dependency. Missing, unknown, or incompatible bindings fail explicitly rather than resolving to an empty string.

## Scope

- Preserve logical node values through execution, `node_completed` persistence, cold resume, composed returns, and aggregation.
- Preserve types for whole-value references.
- Define deterministic interpolation for structured values.
- Allow JSON-compatible values, including real booleans and numbers, in the existing input/binding path.
- Add explicit node-local named bindings for command files and named scripts using the existing `$INPUTS` conventions.
- Preserve the selected return value of an included workflow without flattening it prematurely.
- Give fan-out aggregates a logical array representation rather than an array of encoded JSON strings, while retaining the canonical textual compatibility surface.
- Replace PR #2627's one-line `.green` artifact bridge with a direct boolean binding as the acceptance consumer.

## Non-scope

- The complete workflow input/output schema declaration design in #2453.
- Multiple named workflow results.
- A general artifact-pointer type or automatic artifact loading.
- Runtime-width scheduling and instance identity (#2512).
- A new expression language.
- Dynamic workflow targets or runtime-generated topology.
- `$PREVIOUS` or implicit positional bindings.

## Acceptance

- [ ] A structured node result is persisted and rehydrated with the same logical value and type.
- [ ] Fresh execution and cold resume expose identical values to downstream consumers.
- [ ] A whole-value reference used as a binding remains typed; interpolation produces deterministic text.
- [ ] Boolean and numeric inputs do not become the strings `"true"` and `"1"`.
- [ ] A command file receives node-local bindings through `$INPUTS.<name>`.
- [ ] A named script receives the same bindings through `INPUTS_<UPPER_SNAKE>`.
- [ ] Invalid or non-upstream binding references fail with an actionable error.
- [ ] An included workflow's selected return preserves its logical value.
- [ ] Fan-out aggregation exposes a logical array without double-encoding structured elements.
- [ ] Existing string-only workflows and textual output consumers remain compatible.
- [ ] A cold-resume regression covers node-local bindings.
- [ ] PR #2627 can remove its `.green` artifact bridge and pass the verifier result directly into the named script.

## Relationships

- #2453 remains the umbrella for the complete declared structured I/O contract.
- Supersedes the useful named-binding requirements from #1785 and #2410; their proposed `$PREVIOUS` and command-only `$INPUT` channels are not carried forward.
- #2512 should depend on this value-transport slice rather than on all of #2453.
- PR #2627 is the first concrete consumer.

Wirasm:
## Problem and outcome

A workflow value produced with `output_format` was faithfully typed only inside the in-process run that produced it: persistence kept the text, cold resume re-parsed it, fan-out aggregation double-encoded it, every `with:` value was a string, and a command file or named script could only receive an upstream value through an artifact file the producer wrote and the consumer re-read. The sdlc pack (#2627) carries two such bridges — `implement`'s `.green` file and `deliver`'s `.ready` touch-file latch.

- **Outcome:** Structured values keep their logical type across execution, persistence, cold resume, child-run returns, and fan-out aggregation. `command:`/`script:` nodes bind upstream values by name through the existing `$INPUTS.<name>` / `INPUTS_<UPPER_SNAKE>` channels, including across a skipped branch via an explicit `{ from, if_skipped }` directive. Both #2627 bridges become removable — replicated as integration tests here.
- **Invariant:** A workflow value is a JSON-compatible logical value. Whole-value references and bindings preserve the type; embedding in text uses exactly one deterministic representation (strings raw, everything else canonical JSON); fresh execution and cold resume expose identical values; an unsatisfiable binding fails with an actionable error, never a silent `''`.
- **Scope boundary:** No callee-owned output schemas (#2453), no runtime-width scheduling (#2512), no `$PREVIOUS`/expression language, no change to `returns:` semantics, no edits to #2627's pack YAML. `with:` stays off `bash:`/`prompt:`/loop nodes (their inline bodies already reference outputs directly).

## Review guidance

- **Feedback requested:** Correctness of the compatibility posture (additive event/metadata keys, old-binary degradation) and the strict/lenient split for whole-ref resolution.
- **Start here:** `packages/workflows/src/output-ref.ts` — `canonicalValueText`, `jsonValueSchema`, `parseWholeOutputRef`: the one value rule and the one ref grammar everything else builds on.
- **Review order:** `output-ref.ts` → `schemas/dag-node.ts` (binding directive + `with:` validation) → `dag-executor.ts` (`resolveWorkflowValue`/`resolveNodeBindings`, the `structured_output` emit sites, fan-out reducers, rehydration) → `schemas/workflow-run.ts` + `executor.ts` (`inputs_values`/`summary_value` sibling keys) → `core/src/db/workflow-events.ts` (snapshot) → loader/include-expander (scan + macro) → tests and docs.
- **Lower-attention areas:** test-file map-shape updates (`Map<string, string>` → `{ output, structuredOutput? }`, mechanical); web builder passthrough (`with` round-trip + scan surface); docs.
- **Known risk or uncertainty:** Fan-out aggregate text changes for structured children (single-encoded) — sanctioned by the issue, string children pinned byte-identical by test.

## Solution

Five moves on existing seams, no new subsystem:

1. **One value→text rule.** The byte-identical mapping that existed at four sites (`substituteNodeOutputRefs`, `$LOOP_PREV`, condition evaluator, structured-output serialization) is now one exported helper, `canonicalValueText` — strings raw, everything else canonical JSON. `OUTPUT_REF_SOURCE` and the anchored whole-ref parser moved beside it so the loader scan, the schema's directive validation, and runtime binding resolution share one grammar.
2. **Persistence and resume.** `node_completed` events additively carry `structured_output` wherever the in-memory result has it (prompt/command, loop, loop_group terminal sink, workflow node, fan-out); `node_skipped_prior_success` re-emits copy it forward; the resume snapshot rehydrates it onto `NodeOutput.structuredOutput`. Rows without the key resume exactly as before (text-parse fallback). `workflow_events.data` is schemaless JSON — no migration.
3. **Child-run values and logical aggregates.** A child stamps `metadata.summary_value` beside the text `summary`; the parent threads it back typed. Fan-out join reducers build the aggregate as a logical array: a structured child's element is its payload (single-encoded), a text child's element is the exact raw string, failed children get a marked slot: `{ archon_failed: true, error, status }` (`archon_failed` is the reserved discriminator, added in review round 1 — a structured child's own payload may legitimately carry `error`/`status`).
4. **JSON values through the input channel.** `with:` maps and `inputs:` defaults accept any JSON value (`z.json()`); a string that is exactly one whole `$node.output[.field]` / `$INPUTS.<name>` reference resolves to the logical value, any other string stays a text template. Children persist the typed map under the additive `inputs_values` sibling while legacy `inputs` stays canonical text — a shipped binary reading a non-string `inputs` map treats it as corrupt, so widening in place would lose all inputs on old-binary resume.
5. **Node-local `with:` on `command:`/`script:` nodes.** Delivered through the surfaces those bodies already read, as the nearest-wins input layer (node-local > composed > run). Object values are reserved for the `{ from, if_skipped }` directive: `from` reads one whole upstream ref; a skipped producer takes `if_skipped`, and without one the node fails naming the binding, the producer, and the fix. The loader enforces whole-ref grammar, upstream `depends_on` (mirroring `fan_out.items`), and env-key uniqueness; `with:` on node types that ignore it warns through the parse-warnings channel instead of silently dropping.

The skipped-branch shape follows the constitution's split: `if_skipped` is data the engine substitutes; any coalescing ("ready if either branch said so") stays in the consuming script. Case-law row added.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Fan-out aggregate for structured children: `["{\"v\":1}"]` (elements were JSON strings — collectors parsed twice) | `[{"v":1}]` (elements are the child result values — parse once). Text children byte-identical |
| **Observable behavior** | `with: { flag: true }` was a load error; resumed runs saw text-only outputs | Typed values load and arrive typed; resumed `$node.output.field` reads identical to fresh runs |
| **Failure behavior** | A value needed by a command file/script across a skipped branch required an artifact-file latch; a missing file surfaced as downstream misbehavior | An unsatisfiable binding fails at load (unknown/non-upstream producer, malformed directive) or at the node (skipped without `if_skipped`) with the fix named |

## Review round 1 corrections

The round-1 review (NEEDS FIXES at bb60c175) found the new surface had more interpreters than the PR updated. Commit `bfdcda47` closes all ten findings: the branch is rebased onto current dev (R1, #2685 test-file conflict), the dry-run simulator resolves node-local bindings through the executor's own resolver (R2), loop_group body bindings substitute `$LOOP_PREV` (R3), interactive-gate finalize carries the structured payload so both completion routes share one field-access tier (R4), include expansion re-validates binding values after input substitution (R5), the `all_done` failure marker gained the `archon_failed: true` discriminator (R6), plus the three suggested cleanups (R7–R9). R10 (duplicate acceptance test) is kept deliberately and marked as such in-code.

## Validation

- `bun run validate` — passed (exit 0): bundled/schema/pi/capability checks, type-check across all packages, lint `--max-warnings 0`, format, `test:install`, full per-package suite. Re-run green at the corrected head `bfdcda47`.
- `cd packages/workflows && bun run test` — all 24 segments green, including 34 new tests: persistence round-trip + cold-resume parity (prompt and loop_group producers), command/script bindings (typed literal, whole-ref, template), `if_skipped` on both branches + loud failure without it, binding precedence, cold-resume binding regression, and the two #2627 bridge patterns as end-to-end fixtures.
- `cd packages/core && bun test src/db/workflow-events.test.ts` — 31 pass; `structured_output` round-trip incl. `node_skipped_prior_success` replay rows and falsy values.
- Sub-run/fan-out e2e (`subrun.test.ts`): typed `with:` into child metadata (`inputs_values` + legacy text map + env text), whole-ref logical array delivery, `summary_value` threading, mixed structured/plain fan-out single-encoding, marked failure slots (`archon_failed: true`), chained `fan_out.items` over a logical aggregate with `as` carrying the logical item, and parent cold resume rehydrating the logical aggregate.
- Byte-compat: existing substitution/condition/aggregate suites green with the helper extraction; string-only workflows pinned byte-identical.
- **Not verified:** live CLI kill-and-resume of a real run (the plan's optional manual gate); the same path is covered by the hydrateResumableRun-based resume tests against the in-memory store.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Additive keys only (`structured_output` in event data, `inputs_values`/`summary_value` in run metadata) — no migration, `check:schema-upgrades` untouched. Old rows and older binaries degrade to today's text behavior | `readSubrunMetadata` prefers-then-falls-back; legacy-row resume test; schemas unchanged |
| Rollout / rollback | Each move is additive and independently revertable; stop writing/reading the new keys and current behavior returns | Emit sites are conditional spreads; snapshot tolerates absence |
| Observability | `structured_output` visible in raw event rows (console ignores unknown keys); binding failures are ordinary node failures with actionable messages | Loader/runtime error-message tests |
| Documentation | New guide section (bindings), variables reference, constitution case-law row, fan-out aggregate docs + collector example, CHANGELOG feature + behavior note | This PR |

## Links

- Closes #2637
- Related #2627, #2512, #2453

## Summary by CodeRabbit

- **New Features**
  - Added node-local `with:` bindings for command and script nodes, supporting typed values, output references, environment variables, and skipped-branch fallbacks.
  - Workflow inputs, defaults, included workflows, child workflows, loops, and fan-out results now preserve JSON types.
  - Structured outputs remain available after workflow resumption.

- **Bug Fixes**
  - Improved validation and error reporting for invalid or inaccessible bindings.
  - Preserved structured values during fan-out aggregation and resume operations.

- **Documentation**
  - Expanded workflow authoring, variables, and loop guides with the new behavior.


Wirasm:
## Ready to merge

Round 2 (correction verification). Round 1 reviewed head `bb60c175` and found six blocking findings with one common cause — the new binding/typed-value surface had more interpreters than the PR updated. The branch was then rebased onto current dev (`1c892589`) and corrected in one commit. Head range verified: the rebased feature commit `79c7882c` is patch-identical to the round-1 head except the `dag-executor.test.ts` append anchor (proven by `git range-diff`; dev's #2685 `TokenUsage axis seam guard` block survives exactly once, followed by the PR's block), so the bounded correction diff is exactly `79c7882c..bfdcda47`. All four round-1 reviewer scopes re-ran against that diff and every disposition held: the dry-run simulator now consumes the executor's own `resolveNodeBindings` (R2), loop_group body bindings get the per-iteration `$LOOP_PREV` splice (R3), both gate-finalize routes carry `signaledStructuredOutput` so bare-approve and natural completion share one field-access tier (R4), include expansion re-validates substituted bindings at load (R5), and the fan-out failure slot gained the reserved `archon_failed: true` discriminator (R6). No reviewer found a correction-caused defect. `bun run validate` passes at the pinned head and GitHub CI is green on this exact SHA, including the `schema-upgrade` job; the PR is now `MERGEABLE`.

**Validation:** `bun run validate` PASS (exit 0) at `bfdcda47` in a pinned review worktree; GitHub Actions all green on this head (ubuntu + windows tests, docker-build, docs, schema-upgrade); GitHub reports `MERGEABLE`; CodeRabbit still pending (advisory).

**Resolved:** 9 fixed · 1 declined with recorded reason (R10)

### Findings

|---|---|---|---|
| `R1` | Important | PR conflicts with dev: `packages/workflows/src/dag-executor.test.ts` collides with #2685 — cannot merge as-is | FIXED |
| `R2` | Important | Dry-run never delivers node-local `with:` bindings — false-negative failures and silently missing `INPUTS_*` env in `--exec-code` | FIXED |
| `R3` | Important | `with:` bindings on loop_group body nodes never resolve `$LOOP_PREV` — string form passes literal text silently; directive form is always a load error | FIXED |
| `R4` | Important | loop_group gate-finalize (bare-approve, #2074) omits `structuredOutput`, so `.field` strictness diverges from natural completion | FIXED |
| `R5` | Important | An include-forwarded object input reaches a command/script binding position unvalidated — the promised load error becomes a mid-run node failure | FIXED |
| `R6` | Important | Fan-out `all_done` aggregate: a structured child whose payload carries `error`+`status` fields is indistinguishable from the engine's failure marker; the PR's own doc heuristic misclassifies it | FIXED |
| `R7` | Suggestion | A fifth inline copy of the value→text rule survives in `executeLoopNode` (pre-existing, contradicts the "one rule" extraction) | FIXED |
| `R8` | Suggestion | `WHOLE_INPUTS_REF` regex duplicated in dag-executor.ts and include-expander.ts — the PR's own one-grammar precedent suggests a shared `parseWholeInputsRef` | FIXED |
| `R9` | Suggestion | Builder `with:` passthrough hand-rolls the conditional spread beside the existing `ifDefined` helper | FIXED |
| `R10` | Suggestion | The "implement pattern (PR #2627)" test duplicates the AC4 script-binding test's exact code path | DECLINED |

<details>
<summary><code>R1</code> — Merge conflict with dev (#2685)</summary>

**Impact:** GitHub reported `mergeable: CONFLICTING` at the round-1 head; the PR could not merge until rebased.

**Evidence:** `packages/workflows/src/dag-executor.test.ts` — conflicted with the #2685 token-axis seam-guard tests merged to dev (`1c892589`) after this branch's fork point (`c698f5d2`). Test-only conflict; no production file collided.

**Required outcome:** Rebase onto (or merge) current dev, resolve the test-file conflict, and verify the #2685 guard still passes.

**Found by:** `validation` (coordinator PR-state check)

**Disposition:** FIXED. Branch rebased onto `origin/dev` @ `1c892589`. Coordinator verified fidelity with `git range-diff c698f5d2..bb60c175 1c892589..79c7882c`: the only delta is the append-anchor context in `dag-executor.test.ts` — dev's `TokenUsage axis seam guard` block (`:24287`) survives exactly once (grep count 1), followed by the PR's value-transport block; nothing dropped or duplicated. GitHub now reports `MERGEABLE`; the workflows package passes 579/0 (`dag-executor.test.ts` leg) locally and both CI test jobs are green on `bfdcda47`.

</details>

<details>
<summary><code>R2</code> — Dry-run never delivers node-local <code>with:</code> bindings</summary>

**Impact:** `archon workflow run <name> --dry-run` misreported the PR's headline feature: fake `Unknown input` failures, silently wrong values under run-level shadowing, and missing `INPUTS_<UPPER_SNAKE>` env in `--dry-run --exec-code`.

**Evidence (round 1):** `packages/workflows/src/dry-run.ts:406`, `:450-473`, `:590-596` versus the executor's three-tier merge at `dag-executor.ts:1896-1914` and `:212-251`; zero dry-run coverage.

**Required outcome:** Thread `node.with` through the dry-run's command-node text resolution and `--exec-code` env construction with the executor's nearest-wins merge — never a fake failure.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer` (independently)

**Disposition:** FIXED — verified by both original finders. The dry-run now calls the executor's own `resolveNodeBindings` (exported from `dag-executor.ts` with `runInputs` as an explicit param; `ShellInputContext.workflowRun` narrowed to `Pick<WorkflowRun, 'id' | 'user_message' | 'metadata'>`) through a minimal `bindingShellContext` over the simulator's outputs map (`dry-run.ts:490-508`, `:973-980`) — shared function, not a parallel reimplementation. Command bindings merge nearest-wins into the `$INPUTS` bag (`dry-run.ts:455-483`); script bindings ride the `--exec-code` env (`:578-635`); a binding failure throws the executor's own error and lands as a genuine failed trace entry. All three `resolveNodeBindings` call sites thread `runInputs` consistently; every real-executor call site passes a full `WorkflowRun`, structurally compatible with the Pick. 5 new tests (`dry-run.test.ts:1312-1466`) pin the shadowing/precedence case, unknown-producer error-text parity, `if_skipped` on a skipped producer, and both delivery paths; `dry-run.test.ts` 56/56.

</details>

<details>
<summary><code>R3</code> — loop_group body <code>with:</code> bindings never resolve <code>$LOOP_PREV</code></summary>

**Impact:** The string form passed literal `$LOOP_PREV.seed.output` text silently into a script's env or a command's `$INPUTS`; the directive form was always a load error.

**Evidence (round 1):** `applyLoopPrevToBodyNode` (old `dag-executor.ts:4630-4788`) had a substitution branch for every body-node text surface except `.with`.

**Required outcome:** Add a `.with` branch running binding string values through the same `$LOOP_PREV` substitution pass, or explicitly reject at load naming the supported alternative.

**Disposition:** FIXED — verified by the original finder. `applyLoopPrevToBodyNode` gains a `subWithMap` branch (`dag-executor.ts:4702-4728`) wired into both the script-node and command-node returns; it splices `$LOOP_PREV` into plain string values and string `if_skipped` defaults, per iteration (pinned by `dag-executor.test.ts:18051-18125`: iteration 1 → empty, iteration 2 → previous output). Directive `from:` refs remain a load error whose message now names the working string form (`schemas/dag-node.ts:1018-1025`; pinned by `loader.test.ts:6897-6924`). Nested loop_group bodies inherit the fix through the existing per-body-node recursion. `loop-nodes.md` documents the behavior.

</details>

<details>
<summary><code>R4</code> — loop_group gate-finalize omits <code>structuredOutput</code>; strictness diverges between completion routes</summary>

**Impact:** Natural completion carried `structuredOutput` (tier-2 lenient `.field` access), while the #2074 bare-approve gate-finalize route landed in tier-3 strict text-parse — same workflow, same iteration output, opposite downstream outcomes.

**Evidence (round 1):** Natural completion `dag-executor.ts:4307-4312`/`4519-4521` versus gate finalize `:4054-4074`; consumer tiers `output-ref.ts:271-333`.

**Required outcome:** Make both completion routes produce the same tier, or pin the divergence explicitly.

**Disposition:** FIXED — verified by the original finder. `ApprovalContext` gains additive optional `signaledStructuredOutput` (`schemas/workflow-run.ts`); both gate-write sites persist it (`dag-executor.ts:4616-4618` loop_group, `:6253-6255` loop — the correction fixed the same latent divergence on the plain-loop route symmetrically, confirmed against the pre-correction code) and both finalize routes attach it to the returned `NodeOutput` and the persisted `node_completed` event row under the same `structured_output` key natural completion writes, so `getDagResumeSnapshot` round-trips finalize rows identically. Old persisted gates without the field flow through `?? null` to text-only finalize — no backward-compat gap. Two round-trip tests pin the claim, including the loop_group one proving the downstream lenient tier (absent optional field → `''` instead of node failure) at `dag-executor.test.ts:19556-19673`.

</details>

<details>
<summary><code>R5</code> — Include-forwarded object input defeats the command/script binding load-time check</summary>

**Impact:** A caller-supplied object forwarded through `$INPUTS.<name>` into a command/script binding position bypassed the promised load error and failed mid-run after upstream nodes had spent their cost.

**Evidence (round 1):** `include-expander.ts:528-538`/`617-630` forwarded the object with no post-macro re-validation; contract at `schemas/dag-node.ts:305-309`; empirically reproduced.

**Required outcome:** Re-check substituted command/script binding values after `applyInputsMacro` and fail the expansion loudly — extending the existing `substituteWhen` re-validation precedent.

**Disposition:** FIXED — verified by the original finder, who re-ran the round-1 bypass scenario at the new head. After macro substitution, a plain-object result in a command/script binding position throws `IncludeExpansionError` naming the include, the binding key, and the node (`include-expander.ts:629-643`), converted by `expandOne`'s catch into a load-time `validation_error` — never a runtime throw. Both shapes are rejected: a non-directive object, and a directive-shaped object (deliberately stricter than the parse-time rule — an object arriving post-substitution is necessarily caller-injected, since an authored directive is caught pre-substitution by `isBindingDirective`; rejecting every post-substitution object is exhaustive rather than a drifting copy of `validateNodeBindings`). The guard fires only for command/script nodes; strings, numbers, booleans, null, and arrays pass through. Both shapes pinned in `include-expander.test.ts:2214-2251`; loader + include-expander leg 417/0.

</details>

<details>
<summary><code>R6</code> — Fan-out aggregate: structured child result can be indistinguishable from the failure marker</summary>

**Impact:** A completed structured child whose `output_format` carries top-level `error`+`status` fields (the natural verifier/health-check shape #2627 invites) was indistinguishable from the engine's `{ error, status }` failure slot, and the PR's own doc heuristic misclassified it.

**Evidence (round 1):** `dag-executor.ts:7576-7577`, `:7638-7645`; doc heuristic at `authoring-workflows.md:1872-1879`.

**Required outcome:** Make the failure marker structurally unambiguous, or at minimum document the collision honestly and replace the misclassifying heuristic.

**Disposition:** FIXED — verified by the original finder. The `all_done` failure slot is now `{ archon_failed: true, error, status }` (`dag-executor.ts:7716-7731` — the only production site building this shape); the doc heuristic was replaced with `r.archon_failed === true` (`authoring-workflows.md:1881`), and AGENTS.md, CHANGELOG.md, and the `output-ref.ts` error messages carry the new shape consistently. The residual theoretical collision (an author schema declaring its own `archon_failed` field) is acknowledged in a code comment at the reducer. A new `subrun.test.ts` test pins the exact round-1 scenario: a completed child emitting top-level `error`+`status` beside a genuinely failed sibling, with only the failed slot marked.

</details>

<details>
<summary><code>R7</code> — Fifth inline copy of the value→text rule in <code>executeLoopNode</code></summary>

**Impact:** Two independent implementations of the value→text rule in the same file could drift on a future edit.

**Evidence (round 1):** old `dag-executor.ts:5848-5851` versus `canonicalValueText` (`output-ref.ts:106-108`).

**Required outcome:** Replace the inline mapping with `canonicalValueText(iterationPayload)`.

**Disposition:** FIXED — verified by the original finder. `dag-executor.ts:5925-5927` now calls `canonicalValueText(iterationPayload)`; behavior-equivalent (the payload is already-parsed JSON, so the helper's `?? ''` fallback is unreachable) and the surrounding `failLoopNode` error path is untouched. No new inline copy of the pattern appears anywhere in the correction diff.

</details>

<details>
<summary><code>R8</code> — Duplicated whole-<code>$INPUTS</code> ref regex</summary>

**Impact:** Byte-identical regex constructions in two files — the exact drift shape the PR eliminated for `$node.output` refs.

**Evidence (round 1):** `dag-executor.ts:255` and `include-expander.ts:139`.

**Required outcome:** A shared `parseWholeInputsRef` beside `parseWholeOutputRef` in `output-ref.ts`.

**Disposition:** FIXED — verified by the original finder. `parseWholeInputsRef` and `INPUT_NAME_SOURCE` now live once in `output-ref.ts:125-137`; both former sites consume the shared parser, `schemas/dag-node.ts` re-exports `INPUT_NAME_SOURCE` so existing importers (`when-atom.ts`, `executor-shared.ts`, `schemas/index.ts`) are unaffected, no stray duplicate remains, and the parser is unit-tested (`output-ref.test.ts:86-99`). `bun test ./scripts/` (node-ref-parity) 95/0 in the full validate run.

</details>

<details>
<summary><code>R9</code> — Builder passthrough skips the existing <code>ifDefined</code> helper</summary>

**Impact:** Cosmetic inconsistency beside the helper's established uses.

**Evidence (round 1):** `packages/web/src/experiments/console/builder/variants/script.ts`, `command.ts`.

**Required outcome:** Use `ifDefined('with', …)`.

**Disposition:** FIXED — verified by the original finder. Both variants now use `ifDefined('with', …)` in both `fromDag`/`toDag` directions, matching `deps`/`timeout`.

</details>

<details>
<summary><code>R10</code> — Duplicate test: "implement pattern" ⊂ AC4 script-binding test</summary>

**Impact:** Mild test bloat; the #2627-replication framing has documentation value.

**Evidence (round 1):** the two tests following the AC4 block in `dag-executor.test.ts`.

**Required outcome:** None required; fold or keep deliberately.

**Disposition:** DECLINED (kept deliberately) — verified by the original finder. The test is the source issue's named acceptance bar, and a comment now marks the deliberate overlap with the AC4 test so a future dedupe pass won't churn on it. Round 1 already judged keeping it defensible; no correction required by the PR's outcome.

</details>

<details>

#### Reviewer coverage (round 2 — correction verification against `79c7882c..bfdcda47`)

| Scope | Result |
|---|---|
| code (`prp-core:code-reviewer`) | R2, R7 confirmed fixed; no correction-caused findings |
| seams (`prp-core:seam-analyzer`) | R2, R3, R4 confirmed fixed; new seams (dry-run binding resolution, gate metadata field, `$LOOP_PREV` with-branch, shared `parseWholeInputsRef`) checked — counterparts all updated |
| simplify (`prp-core:code-simplifier`) | R6, R8, R9 confirmed fixed; R10 decline justified; correction adds reuse, not duplication |
| types (`prp-core:type-design-analyzer`) | R5 confirmed fixed (bypass re-run at new head); `ApprovalContext` optionality, `ShellInputContext` Pick, and `INPUT_NAME_SOURCE` move all regression-free |
| validation / PR state (coordinator) | R1 fixed: rebase fidelity proven by range-diff; #2685 guard present exactly once; `MERGEABLE` |

Round-1 coverage of the unchanged feature commit carries forward: `79c7882c` is patch-identical to the round-1 reviewed head `bb60c175` except the test append-anchor (range-diff), so the round-1 full-review evidence (emit-site sweeps, `SUBRUN_METADATA_KEYS` compat, ref-grammar family, `canonicalValueText` byte-compat, resume re-derivation parity) applies to the new base without re-execution.

#### Validation

|---|---|---|
| `bun run validate` (review worktree pinned at bfdcda47) | PASS | exit 0; all package type-checks 0 errors; eslint/format clean; all test legs 0 fail (workflows 579+417+…; final `bun test ./scripts/` 95/0) |
| `git range-diff c698f5d2..bb60c175 1c892589..79c7882c` | PASS | single hunk delta: test append-anchor context only; #2685 `TokenUsage axis seam guard` present exactly once at head |
| GitHub Actions on head bfdcda47 | PASS | test (ubuntu) ✓, test (windows) ✓, docker-build ✓, Build docs site ✓, schema-upgrade ✓; CodeRabbit pending (advisory) |
| GitHub mergeability | PASS | `mergeable: MERGEABLE` (was `CONFLICTING` at round 1) |
| `bun run check:schema-upgrades` (local) | NOT RUN | needs live PostgreSQL; covered by the green `schema-upgrade` CI job on this head; PR touches no `migrations/` files |

</details>

Issue #2609, opened 2026-08-18
fix(workflows): a node that fails mid-stream loses its spend from the event log — resume cannot rebuild it


Wirasm:
A node that fails **mid-stream** keeps the usage it already burned on its `NodeExecutionResult` — `costUsd` and `tokens` are captured from earlier provider chunks and are not cleared in the catch path (`packages/workflows/src/dag-executor.ts:2582-2600`). That value is genuine: it reaches `ctx.totalCostUsd` at the single run-level aggregation point (`:7893-7905`), so the run row records it.

The **event** written for that node does not:

```ts
// dag-executor.ts:2569-2573
event_type: 'node_failed',
data: { error: err.message },
```

No `cost_usd`, no `tokens` — unlike `node_completed`, which carries both (`:2483`). The user-cancel abort branch (`:2578-2590`) returns usage on the result object and writes **no** event at all.

## Why it matters now

#2469 (PR #2608) made a run persist its usage on every disposition, and made `getDagResumeSnapshot` rebuild cumulative cost across resume passes the way it already did for tokens. Both rebuild from `node_completed` rows only — which is correct for what those rows contain, and is why the fix is sound. But it makes this gap the remaining asymmetry:

- Pass 1: node A completes ($0.01, in the event log). Node B fails mid-stream after burning $0.02 — in `ctx`, so the failed run's row correctly reads **$0.03**.
- Resume: the snapshot can only rebuild **$0.01**. Node B re-runs and costs $0.02 again. Final row reads **$0.03**.
- Money actually spent across the run's lifetime: **$0.05**.

The figure never *decreases* (that regression is what PR #2608 avoids), and it is never double-counted. It simply cannot see a failed attempt's spend once the process that held it is gone. On a workflow that fails and is resumed several times, the wasted spend is exactly the part an operator most wants to see.

## The decision this needs

Persisting `cost_usd`/`tokens` on `node_failed` is the easy half. The real question is what `total_cost_usd` should *mean*:

1. **Money burned** — sum failed attempts too. Truthful about spend, but then the same node's first and second attempt both count, which is a deliberate change to what the number has always meant, and it interacts with `retry:` (every attempt counts) and with `always_run` nodes on resume.
2. **Cost of the surviving path** — keep summing `node_completed` only, and surface wasted spend as a separate figure.

(1) is probably what an operator watching a budget wants; (2) is what the current number is. Picking one is the work here — the event-payload change is small either way.

## Scope

[...]

PR #2694, opened 2026-08-21
fix(workflows): a failed node's spend now reaches the run transcript

### LINKED ISSUE #2693: fix(workflows): a failed node's spend never reaches the JSONL transcript — logNodeError takes no usage
## Problem

A node that burns money and then **fails** records that spend in the database and not in the JSONL transcript. `logNodeError` takes only `(logDir, workflowRunId, nodeId, error)` — no usage of any kind — while the sibling `logNodeComplete` carries the full `WorkflowUsage` payload since #2676.

The code already says so. `dag-executor.ts` carries this comment beside `nodeUsageEventData`:

```ts
// What this node reported, built once and passed whole rather than re-listed per sink:
// the DB event on every outcome, and the JSONL transcript row on completion. The JSONL
// FAILURE row is still outside it — logNodeError takes no usage, so a node that spent
// money and then failed records that spend in the DB and not in the transcript (#2614).
// See WorkflowUsage.
```

A second hole sits beside it: the **user-cancel abort branch calls no `logNodeError` at all**, so a cancelled node produces no transcript row whatsoever — not even one without usage.

## What this is NOT — read this before estimating

This issue is the residue of #2609, which is being closed. Roughly half of what #2609 described has already shipped, in PR #2654, which was never linked to it. On `dev` today:

- Every AI-path `node_failed` DB writer carries usage through `nodeUsageEventData()` — the catch path, credit exhaustion, empty output, and the user-cancel abort branch, which now writes a DB event where #2609 said it wrote none.
- `getDagResumeSnapshot` sums both usage axes from `node_completed` **and** `node_failed` rows. Its docblock states the semantic outright: *"Failed rows contribute spend but never completed outputs."*
- #2609's worked example — node A completes at $0.01, node B fails after burning $0.02 — is already a passing test in `packages/core/src/db/workflow-events.test.ts`.

So the "what should `total_cost_usd` mean" question #2609 called its real work is settled and implemented: it means money burned, failed attempts included. Do not re-litigate it, and do not expect the DB-side acceptance scenario to fail — it passes.

## Scope

- Extend the existing `WorkflowUsage` carrier to `logNodeError` rather than adding a parallel usage path. #2676 built that primitive precisely so a new axis reaches every sink or none.
- Wire it at the failure paths that actually hold usage. The bash and script failure writers hold none by definition and should stay unchanged.
- Add the missing transcript row for the user-cancel abort branch.
- The new carrying site should appear in the #2685 seam guard as one row using the existing whole-object map. A reviewer traced this during #2685 and confirmed no rewrite of the guard is needed; if that turns out to be wrong, say so rather than reshaping the guard quietly.
- Re-read any comment claiming resume usage is rebuilt from `node_completed` rows alone. That was true before #2654 and is not true now.

## Acceptance

- A node that burns tokens and cost and then fails writes those to the JSONL transcript, and a transcript-level version of #2609's worked example fails against today's code and passes after.
- A user-cancelled node produces a transcript row.
- Adding a future `TokenUsage` axis and forgetting the failure sink fails the seam guard.

## Why it matters

The transcript is the artifact an operator or agent reaches for when asking what a run actually did, and the failed attempts are exactly the part worth seeing — a run that failed and was resumed three times has its wasted spend visible in the database and invisible in the log people actually read.

Related: #2609 (closed, this is its remaining half), #2654 (shipped the DB half), #2676 (built the `WorkflowUsage` carrier), #2685 (the seam guard), #2614 (transcript surfacing).

Wirasm:
## Problem and outcome

A node that burned money and then failed recorded that spend in the database and not in the JSONL run transcript. `logNodeError` took only `(logDir, workflowRunId, nodeId, error)`, while its sibling `logNodeComplete` has carried the full `WorkflowUsage` payload since #2676 — so the two sinks disagreed about what a failed node cost. The transcript is the artifact an operator reaches for when asking what a run actually did, and failed attempts are exactly the part worth seeing.

- **Outcome:** a node that fails after reporting usage writes that usage to the transcript, carrying the same `cost_usd` and `tokens` as its `node_failed` database row. A user-cancelled node now produces a transcript row at all, which it never did before.
- **Invariant:** wherever a node-failure row is written, the JSONL row and the persisted event report the same axes. An axis added to `TokenUsage` reaches both or neither — now enforced for the failure sink by the #2685 seam guard.
- **Scope boundary:** the database and resume halves are untouched; #2654 already shipped them. Three node-completion gaps are deliberately left: `workflow:` and fan-out nodes write no transcript rows at all, a `loop:` node's success exits write no terminal transcript row for its own id, and `logWorkflowError` still takes no usage. All three are completion-side or run-level and belong to #2614's coverage audit — see Solution.
- **Root cause:** `logNodeError` had no parameter to pass usage to. Every affected call site already holds that usage and hands it to `createWorkflowEvent` two lines away.

## Review guidance

- **Feedback requested:** whether the third fix (the outer catch's `cancelled` branch) belongs in this PR — #2693 names only the streaming-cancel branch, and I fixed both. Reasoning under Solution.
- **Start here:** `logNodeError` in `packages/workflows/src/logger.ts` — the whole change follows from its new fifth parameter.
- **Review order:** `logger.ts` (the carrier) → the `logNodeError` call sites in `dag-executor.ts` → the two cancel exits → the seam-guard row and behavioral tests → the three prose corrections.
- **Lower-attention areas:** the prose corrections in `getDagResumeSnapshot` and `runNodeRetryLoop` change no behavior. They describe the money-burned semantic #2654 introduced, which nothing stated where a reader would meet it.
- **Known risk or uncertainty:** None. The transcript has no consumers yet (#2614 is the issue to surface it), so the added keys cannot break a reader.

## Solution

`logNodeError` gained an optional `WorkflowUsage` parameter, spread whole, mirroring `logNodeComplete`. This reuses the carrier #2676 built for exactly this purpose rather than adding a parallel usage path. The call sites that hold usage now pass it: credit exhaustion, the empty-output guard, the outer catch in `executeNodeInternal`, and `failLoopNode`. The writers that fail before any provider call — command-file load, prompt substitution, `executeBashNode`, script discovery, named-script not-found, `executeScriptNode`'s catch — keep passing nothing, so an absent key still means "nothing was reported" rather than "reported zero". That distinction matters because Codex reports no cost at all (#2334); it has a test.

**Both user-cancel exits, not just the one the issue names.** The abort has two. The streaming-cancel branch returned without ever calling `logNodeError` — that is the hole #2693 describes. The second is the outer catch, whose `cancelled` branch also skipped `logNodeError`; an aborted node that declared `output_format` reaches it, because `canReask` is false once the signal is aborted and the node throws instead of returning. Fixing only the first would have made a cancelled node's transcript row depend on whether the node declared `output_format` — a provider-shaped inconsistency of the kind this issue exists to remove. Both exits now write a row, and the second has its own test proven red against the fixed first.

**`workflow:` and fan-out nodes are excluded on purpose.** Their `failResult` helpers already accept cost and tokens, and `logDir` is on `RunLayersContext`, so a transcript row would have been cheap. But those node types write no transcript rows in any state — no `node_start`, no `node_complete`, no `node_error`. Adding only the failure row would make a `workflow:` node visible in the transcript exclusively when it fails, which is a worse transcript than one that omits it consistently. Giving them a full lifecycle belongs to #2614's coverage audit.

**A `loop:` node now reports its cumulative spend on failure and not on success**, which review raised and which I have documented rather than closed. `failLoopNode` carries the loop's totals; the only `logNodeComplete` on that path is per-iteration and carries duration alone, so no success exit writes a terminal transcript row for the loop's own id. Closing it means adding a new terminal row shape at three separate success exits — a completion-path addition this failure-scoped PR did not set out to make, and one whose shape belongs with whoever builds the transcript reader. What I did instead is remove the false claim it would have contradicted: `WorkflowUsage`'s docblock no longer says both terminal outcomes carry usage, and now names loop completion, `workflow:`/fan-out, and the per-row reading of an absent `cost_usd` explicitly. #2614 owns the behavioral half.

**Prose the #2654 semantic left false.** `executeDagWorkflow`'s `priorUsage` doc said cumulative usage is restored from `node_completed` events on resume; the snapshot has summed failed rows since #2654. `getDagResumeSnapshot` stated its summing rule without ever saying what the total now *means* — it is money burned, not the cost of the surviving path, so the same node's two attempts both count, `retry:` counts every attempt, and an `always_run` node accrues on each resume pass. All three are intended consequences and are now written down where a reader meets the function. `runNodeRetryLoop` did not record that each attempt writes its own event, which is why the event log and the accumulated total agree at 0.02 + 0.03 = 0.05 instead of double-counting.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A failed node's JSONL row carried `error` and nothing else; the matching database row carried `cost_usd` and `tokens`. | The JSONL row carries the same usage as the database row. |
| **Failure behavior** | A user-cancelled node produced no JSONL row at all — the node appeared in the transcript as a `node_start` that never terminated. | It produces a `node_error` row reading `Cancelled by user`, carrying what it spent, from either abort exit. |

### Changed seams

|---|---|---|
| `WorkflowUsage → JSONL node_error row` | New sink; spreads the carrier whole | `JSONL transcript node_error row carries every axis it claims` in the seam guard |
| `executeNodeInternal cancel → transcript` | Row now written from both abort exits | `a node cancelled by the user leaves its spend in the transcript`, `a cancel that surfaces as a thrown error still leaves a transcript row` |

## Validation

- `bun run validate` — exit 0 — all ten stages green: generated-file checks, type-check across every package, eslint at `--max-warnings 0`, prettier, installer tests, and the full per-package suite.
- **Every new test was observed failing first.** With the source reverted and the tests kept: the mid-stream case reports `cost_usd` `Expected: 0.02, Received: undefined`; the cancel case reports `error` `Expected: "Cancelled by user", Received: undefined`, because no row existed; the two `logNodeError` logger tests fail. The catch-path cancel was proven separately by mutating only that hunk back to its original shape with everything else fixed, so it is load-bearing on its own rather than incidentally covered.
- **The seam guard was proven red twice**, because a guard that has only been seen green is a claim. Dropping `tokens` from the failure row flips `"JSONL node_error.tokens carried usage"` from `true` to `false`. The sharper mutation — carrying `tokens` but rebuilding it as `{input, output}`, which is the #2662 failure mode the guard exists for — reports `"JSONL node_error.tokens → cacheRead (as 'cacheRead')": 4000 → undefined`, naming the exact axis rather than just the seam.
- The acceptance scenario is a two-node DAG matching #2609's worked example: node A completes at $0.01, node B reports $0.02 and fails on the same result chunk. It asserts the transcript failure row, the completion row, the persisted event, and the $0.03 run total together, so the fix is shown to close the gap between the sinks rather than move it.
- **Not verified:** no live paid-provider run. Usage arrives through the typed provider result chunk, which the tests drive directly at that boundary.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Additive keys on a JSONL row type that already exists. Historical rows keep no usage, and an absent key already means "not reported". Nothing parses the transcript today. | #2614 is the issue to surface it |
| Rollout / rollback | Revert is the diff. No schema, stored state, or run-lifecycle change. | — |
| Observability | This is the observability fix: wasted spend on a failed-then-resumed run becomes visible in the log people read, not only in the database. | The two cancel tests and the acceptance test |

## Links

- Closes #2693
- Related #2609, #2654, #2676, #2685, #2614

## Summary by CodeRabbit

* **Bug Fixes**
  * Usage data is now preserved when workflow nodes fail, are cancelled, run out of credits, or produce empty output.
  * Cost and token totals now accurately include failed and retried attempts.
  * Resumed workflows retain usage from previous completions and failures.
  * Usage details remain available in workflow logs and run totals, including zero-cost results.

* **Documentation**
  * Clarified how failed attempts, retries, and resumed executions contribute to cumulative usage and spending.

### DIFF
--- a/packages/core/src/db/workflow-events.ts
+++ b/packages/core/src/db/workflow-events.ts
@@ -221,10 +221,28 @@ export async function listWorkflowEventsSince(
  *
  * Both usage axes are summed from `node_completed` and `node_failed` rows, and only
  * from rows that are not marked `data.aggregate`. Failed rows contribute spend but
- * never completed outputs, so their nodes remain eligible for resume. Cache axes sum
- * over the rows that reported them and carry `cachePartial` when any row did not, so a
- * pre-#2654 row narrows the cache total instead of erasing it. Two distinct
- * duplication hazards:
+ * never completed outputs, so their nodes remain eligible for resume.
+ *
+ * This makes a run's total MONEY BURNED, not the cost of the surviving path — the
+ * figure an operator watching a budget wants, and a deliberate change from what the
+ * number meant before failed rows were summed (#2654). Three consequences follow, all
+ * intended:
+ *
+ * - The same node's first and second attempt both count. A node that failed at $0.02
+ *   and succeeded at $0.03 on resume contributes $0.05, because both rows are real
+ *   spend.
+ * - `retry:` counts every attempt, for the same reason — `runNodeRetryLoop` writes one
+ *   event per attempt.
+ * - An `always_run` node re-executes on every resume pass and its spend accrues each
+ *   time.
+ *
+ * A resumed run's total therefore exceeds what the surviving path cost, and grows with
+ * each resume. That is the point; it is not double counting, which is what the two
+ * exclusions below prevent.
+ *
+ * Cache axes sum over the rows that reported them and carry `cachePartial` when any row
+ * did not, so a pre-#2654 row narrows the cache total instead of erasing it. Two
+ * distinct duplication hazards:
  *
  * - `node_skipped_prior_success` rows replay a node an earlier pass already counted, so
  *   counting them would multiply that node's usage by the number of resume passes.
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -405,6 +405,19 @@ function loaderBypassingWorkflow(
 
 // --- Tests ---
 
+/**
+ * A run's JSONL transcript, parsed. Module-scoped so every describe reads a run's
+ * transcript the same way.
+ */
+const readTranscript = async (
+  logDir: string,
+  runId: string
+): Promise<Array<Record<string, unknown>>> =>
+  (await readFile(join(logDir, `${runId}.jsonl`), 'utf-8'))
+    .trim()
+    .split('\n')
+    .map(line => JSON.parse(line) as Record<string, unknown>);
+
 describe('buildTopologicalLayers', () => {
   it('single node with no dependencies -> one layer', () => {
     const layers = buildTopologicalLayers([node('a')]);
@@ -5967,15 +5980,6 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
     // gets believed. Absence is the honest answer.
     expect(completedEvent?.[0].data).not.toHaveProperty('tokens');
   });
-
-  const readTranscript = async (
-    logDir: string,
-    runId: string
-  ): Promise<Array<Record<string, unknown>>> =>
-    (await readFile(join(logDir, `${runId}.jsonl`), 'utf-8'))
-      .trim()
-      .split('\n')
-      .map(line => JSON.parse(line) as Record<string, unknown>);
 
   it('writes node and run cost to the transcript, matching the persisted events', async () => {
     mockSendQueryDag.mockImplementation(async function* () {
@@ -14021,10 +14025,7 @@ describe('executeDagWorkflow -- run usage survives every disposition', () => {
       minimalConfig
     );
 
-    const rows = (await readFile(join(logDir, `${workflowRun.id}.jsonl`), 'utf-8'))
-      .trim()
-      .split('\n')
-      .map(line => JSON.parse(line) as Record<string, unknown>);
+    const rows = await readTranscript(logDir, workflowRun.id);
 
     const failureRow = rows.find(row => row.type === 'node_error' && row.step === 'b');
     expect(failureRow?.cost_usd).toBe(0.02);
@@ -14107,10 +14108,7 @@ describe('executeDagWorkflow -- run usage survives every disposition', () => {
 
     // Prove the abort branch is the one that ran: it is the only path that reports
     // exactly 'Cancelled by user'.
-    const rows = (await readFile(join(logDir, `${workflowRun.id}.jsonl`), 'utf-8'))
-      .trim()
-      .split('\n')
-      .map(line => JSON.parse(line) as Record<string, unknown>);
+    const rows = await readTranscript(logDir, workflowRun.id);
     const cancelRow = rows.find(row => row.type === 'node_error' && row.step === 'only');
     expect(cancelRow?.error).toBe('Cancelled by user');
     expect(cancelRow?.cost_usd).toBe(0.02);
@@ -14173,10 +14171,7 @@ describe('executeDagWorkflow -- run usage survives every disposition', () => {
       setSystemTime();
     }
 
-    const rows = (await readFile(join(logDir, `${workflowRun.id}.jsonl`), 'utf-8'))
-      .trim()
-      .split('\n')
-      .map(line => JSON.parse(line) as Record<string, unknown>);
+    const rows = await readTranscript(logDir, workflowRun.id);
     const cancelRow = rows.find(row => row.type === 'node_error' && row.step === 'only');
     expect(cancelRow?.error).toBe('Cancelled by user');
     expect(cancelRow?.cost_usd).toBe(0.02);
@@ -24480,10 +24475,7 @@ describe('TokenUsage axis seam guard', () => {
 
   it('JSONL transcript node_complete row carries every axis it claims', async () => {
     const { runId, logDir } = await runSpecimenDag();
-    const rows = (await readFile(join(logDir, `${runId}.jsonl`), 'utf-8'))
-      .trim()
-      .split('\n')
-      .map(line => JSON.parse(line) as Record<string, unknown>);
+    const rows = await readTranscript(logDir, runId);
     const nodeComplete = rows.find(row => row.type === 'node_complete' && row.step === 'step');
     expectSeamCarriesAxes(
       'JSONL node_complete.tokens',
@@ -24497,10 +24489,7 @@ describe('TokenUsage axis seam guard', () => {
     // because logNodeError spreads the same WorkflowUsage carrier logNodeComplete
     // does — one decision, so a new axis either rides both rows or neither.
     const { runId, logDir } = await runSpecimenDag('failed');
-    const rows = (await readFile(join(logDir, `${runId}.jsonl`), 'utf-8'))
-      .trim()
-      .split('\n')
-      .map(line => JSON.parse(line) as Record<string, unknown>);
+    const rows = await readTranscript(logDir, runId);
     const nodeError = rows.find(row => row.type === 'node_error' && row.step === 'step');
     expectSeamCarriesAxes(
       'JSONL node_error.tokens',
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -13966,6 +13966,222 @@ describe('executeDagWorkflow -- run usage survives every disposition', () => {
     ]);
   });
 
+  it('a node that fails mid-stream reports its spend in the transcript, not just the DB', async () => {
+    // #2609's worked example. Node A completes at $0.01; node B burns $0.02 and then
+    // the provider throws. The DB row has carried that $0.02 since #2654 — the JSONL
+    // failure row did not, because logNodeError had no usage parameter. The two sinks
+    // disagreeing about what a node cost is the bug.
+    let call = 0;
+    mockSendQueryDag.mockImplementation(async function* () {
+      call++;
+      if (call === 1) {
+        yield { type: 'assistant', content: 'node A output' };
+        yield { type: 'result', sessionId: 'sid-a', cost: 0.01, tokens: { input: 10, output: 1 } };
+        return;
+      }
+      // Node B streams, reports what it burned, and the same result chunk carries the
+      // provider's error — so the node throws with its usage already accumulated. This
+      // is the ordering that makes the bug reachable: the spend is real and the node
+      // still ends at a failure writer.
+      yield { type: 'assistant', content: 'node B got this far' };
+      yield {
+        type: 'result',
+        sessionId: 'sid-b',
+        cost: 0.02,
+        tokens: { input: 20, output: 2 },
+        isError: true,
+        errorSubtype: 'stream_error',
+      };
+    });
+
+    const store = createMockStore();
+    const logDir = join(testDir, 'logs');
+    const workflowRun = makeWorkflowRun('spend-transcript-run');
+
+    await executeDagWorkflow(
+      createMockDeps(store),
+      createMockPlatform(),
+      'conv-spend',
+      testDir,
+      {
+        name: 'spend-transcript',
+        nodes: [
+          { id: 'a', prompt: 'Cheap work.' },
+          { id: 'b', prompt: 'Expensive work that dies.', depends_on: ['a'] },
+        ],
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      logDir,
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const rows = (await readFile(join(logDir, `${workflowRun.id}.jsonl`), 'utf-8'))
+      .trim()
+      .split('\n')
+      .map(line => JSON.parse(line) as Record<string, unknown>);
+
+    const failureRow = rows.find(row => row.type === 'node_error' && row.step === 'b');
+    expect(failureRow?.cost_usd).toBe(0.02);
+    expect(failureRow?.tokens).toEqual({ input: 20, output: 2 });
+
+    // The completed node was already right; asserting it here proves the transcript
+    // now accounts for the whole run rather than trading one gap for another.
+    const completeRow = rows.find(row => row.type === 'node_complete' && row.step === 'a');
+    expect(completeRow?.cost_usd).toBe(0.01);
+
+    // Same figure in the persisted event: the fix closes the gap between the sinks
+    // rather than moving it.
+    const failedEvents = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls.filter(
+      (c: unknown[]) => {
+        const e = c[0] as { event_type: string; step_name?: string };
+        return e.event_type === 'node_failed' && e.step_name === 'b';
+      }
+    );
+    expect(failedEvents.length).toBe(1);
+    expect((failedEvents[0][0] as { data: Record<string, unknown> }).data.cost_usd).toBe(0.02);
+
+    // And the run total is the money burned across both nodes.
+    expect(runUsageWrites(store)).toEqual([
+      { total_cost_usd: 0.03, total_tokens_in: 30, total_tokens_out: 3 },
+    ]);
+  });
+
+  it('a node cancelled by the user leaves its spend in the transcript', async () => {
+    // The abort branch called no logNodeError at all, so a cancelled node produced no
+    // transcript row whatsoever — not even one without usage (#2693). It is the only
+    // terminal outcome in executeNodeInternal that was missing from the transcript.
+    //
+    // The cancel is driven the way production drives it: the throttled mid-stream
+    // status check sees a non-running run and aborts. setSystemTime jumps past
+    // CANCEL_CHECK_INTERVAL_MS between chunks so that check fires deterministically,
+    // and the usage-bearing result chunk is delivered BEFORE the jump so the node has
+    // something to report by the time it is torn down.
+    let usageDelivered = false;
+    mockSendQueryDag.mockImplementation(async function* () {
+      // A live background task keeps the stream open past the result chunk, which is
+      // what lets the node still be streaming once it has usage to report.
+      yield {
+        type: 'background_tasks',
+        tasks: [{ taskId: 't-live', taskType: 'local_agent', description: 'still running' }],
+      };
+      yield { type: 'assistant', content: 'partial work' };
+      yield { type: 'result', sessionId: 'sid-x', cost: 0.02, tokens: { input: 30, output: 3 } };
+      usageDelivered = true;
+      setSystemTime(new Date(Date.now() + 11_000));
+      yield { type: 'assistant', content: 'MUST NOT BE REACHED' };
+    });
+
+    const store = createMockStore();
+    (store.getWorkflowRunStatus as Mock<() => Promise<string | null>>).mockImplementation(() =>
+      Promise.resolve(usageDelivered ? 'cancelled' : 'running')
+    );
+    const logDir = join(testDir, 'logs');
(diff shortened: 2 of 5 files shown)


Wirasm:
## Ready to merge — corrections verified at `0e447209`

Round 1 (`a793fbe8`) found 0 blocking and 6 Suggestion findings and named two of them worth
fixing before merge. `0e447209` fixes four and declines two with reasons recorded in the code
and the PR body. Every disposition was checked against the code at the current head rather than
against its description, and the two claims most worth doubting — a comment replacing a false
comment, and a docblock narrowing a false invariant — are both true of the control flow they
describe.

The correction commit touches 3 files for +43/−41. Every hunk but one is comment text, docblock
text, or the test-helper hoist; the only executable change is `R4`'s: `failLoopNode` binds
`loopUsage` once and spreads it into both sinks, which alters JSON key order inside the
persisted `node_failed` `data` and nothing else. No test was deleted and no assertion changed —
`bun test src/logger.test.ts src/dag-executor.test.ts` reports the same 592 pass / 1647
`expect()` calls as round 1, and the file carries the same 568 `it(` blocks and 69 `describe(`
blocks at both heads.

**0 blocking · 0 open · 1 nit (no action required)**

**Validation:** `bun run validate` exit 0 in a detached checkout of `0e447209` — all ten
stages, 148 `N fail` summary lines, every one `0 fail`, and no non-zero fail counter anywhere in the
1449-line log. All five CI checks pass at the head, `test (windows-latest)` included
(6m48s, no flake). `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

**Tracked follow-ups:** [#2614](https://github.com/coleam00/Archon/issues/2614)

### Disposition verification

| ID | Claimed | Verified | Evidence |
|---|---|---|---|
| `R1` | FIXED | **Confirmed** | `dag-executor.ts:2816-2822`. Every load-bearing clause is true of the code — see below |
| `R3` | FIXED (claim) / DECLINED (behavior) | **Confirmed** | `logger.ts:61-67` narrowed claim is true clause by clause; decline recorded in the PR body's scope boundary and in the docblock, tracked on #2614 |
| `R4` | FIXED | **Confirmed** | `dag-executor.ts:4675-4679` binds once, `:4687` spreads the same value into the event |
| `R5` | FIXED | **Confirmed** | `dag-executor.test.ts:412-419` module-scoped; 5 inline copies gone; test/assertion counts identical at both heads |
| `R6` | DECLINED | **Confirmed** | Round 1 required no action in this PR; nothing in the tree contradicts the decline |
| `R2` | TRACKED FOLLOW-UP | **Confirmed** | `logWorkflowError` (`logger.ts:168`) still takes no usage — unchanged, as stated. Covered by #2614's stated acceptance ("per-node usage/cost") |

<details>
<summary><code>R1</code> — the replacement comment is true of the control flow</summary>

Each clause of `dag-executor.ts:2816-2822`, checked against the code at `0e447209`:

| Clause | Verdict | Evidence |
|---|---|---|
| "The persisted event below is written either way" | True | The `createWorkflowEvent('node_failed')` at `dag-executor.ts:2826` is unconditional in the catch — it sits outside the `if (cancelled)` at `:2811` |
| "the engine's own structured-output gate … runs before the streaming-cancel branch" | True | The reask `while (true)` block ends at `dag-executor.ts:2536` (its last statement is the throw at `:2532`); the streaming-cancel branch opens at `:2553` |
| "cannot reask once the signal is aborted" | True | `const canReask = reaskAttempt < maxReasks && !nodeIdleTimedOut && !nodeAbortController.signal.aborted` — `dag-executor.ts:2460-2461` |
| "an aborted node declaring `output_format` throws here instead of returning there" | True of the route the code takes on an abort | With `canReask` false and no schema-valid structured output, `:2532` throws. `runStreamPass` returns normally on abort (it `break`s the `for await` at `:1853`), so nothing else carries the node past the gate |
| "A provider SDK that throws on abort also lands here" | True | `runStreamPass` is awaited at `:2437` inside the outer `try`; the inner `try` has only a `finally`, so a generator throw propagates |

The old comment's specific failure mode is gone. It asserted an exclusive ("only when the
provider SDK threw on abort") that would let a maintainer conclude the branch is unreachable on
a provider that ends its stream cleanly, and delete the fix. The replacement names the
engine-owned route the PR's own test drives — `a cancel that surfaces as a thrown error still
leaves a transcript row` (`dag-executor.test.ts:14118`) yields prose, never throws, and reaches
the catch through `:2532`.

**Nit `N1`, no action required.** "so an aborted node declaring `output_format` throws here
instead of returning there" reads as universal, and one narrow case escapes it: if a result
chunk delivered *schema-valid* structured output before the abort was noticed, `:2500` breaks
out of the gate and the node returns via the streaming-cancel branch at `:2553` instead. That
path is reachable — it is the shape of the PR's first cancel test with a valid payload — but it
is the opposite of the round-1 defect: it makes the catch look *more* reachable than it is, not
less, and both exits now write the row, so nothing about the fix depends on the distinction.
The hedge "mainly" on the preceding sentence already carries most of the weight. Recorded for
accuracy, not as something to change.

</details>

<details>
<summary><code>R3</code> — the narrowed <code>WorkflowUsage</code> claim is true, and the decline is recorded rather than dropped</summary>

`logger.ts:55-59` no longer says "Both terminal outcomes carry it". `logger.ts:61-67` now says
the symmetry holds per sink, not across every node type, and names the gaps. Clause by clause:

| Clause | Verdict | Evidence |
|---|---|---|
| "A `loop:` node's cumulative totals reach its transcript row on failure" | True | `failLoopNode` passes `loopUsage` to `logNodeError` (`dag-executor.ts:4679`), and every mid-loop caller passes `loopTotalCostUsd`/`loopTotalTokens` — `:4889`, `:5418`, `:5447`, `:5509`, `:5582`, `:5606`, `:5632`, `:5922`, `:5984`. The four pre-iteration callers (`:4793`, `:4803`, `:4817`, `:4841`) pass none, correctly: nothing was spent |
| "and its persisted event on either outcome" | True | Failure: `node_failed` spreads the same `loopUsage` (`:4687`). Success: the completion-signal exit writes `node_completed` with `tokens`/`cost_usd` (`:5843-5844`), and the finalize-on-approve exit passes `finalizeUsage` through `finalizeLoopFromSignal` (`:3736-3737`) |
| "no success exit writes a terminal transcript row for the loop's own id" | True | A repo-wide grep for `await logNodeStart(`/`await logNodeComplete(`/`await logNodeError(` outside tests returns 19 call sites; exactly one is on the loop success path, `dag-executor.ts:5811`, and its `nodeId` argument is `` `${node.id}-iteration-${i}` ``. `finalizeLoopFromSignal` (`:3702-3754`) writes the persisted event and the emitter event and no transcript row at all |
| "only per-iteration rows, which carry duration and no usage" | True | `:5811-5813` passes `{ durationMs: duration }` and nothing else |
| "`workflow:` and fan-out nodes write no transcript rows at all" | True | The same grep places every transcript writer inside `executeNodeInternal` (`:1616`, `:1665`, `:1711`, `:2562`, `:2624`, `:2667`, `:2732`, `:2823`), `executeBashNode`, `executeScriptNode`, or `executeLoopNode`. `executeWorkflowNode` (`:6268`) and the fan-out join's `writeCompleted` (`:6880`) call none. The enumeration is also complete for the property that matters: `approval:` nodes have no spend, and a `loop_group`'s body nodes run through `runLayers` (`:4022`), so the group's spend does reach the transcript as namespaced per-node rows — only `workflow:` and fan-out have spend that reaches no transcript row anywhere |

The declined behavioral half is recorded in three places rather than dropped: the docblock
above ("Completing that coverage is #2614's audit"), the PR body's scope boundary ("a `loop:`
node's success exits write no terminal transcript row for its own id"), and the PR body's
Solution section, which states the decline and its reason explicitly. #2614's acceptance names
"per-node usage/cost" and "loop iterations", so the tracking is real at the issue's own scope —
though it is a pointer from this PR to #2614, not a comment on #2614 enumerating the two
residues. That is enough for a Suggestion-level deferral.

</details>

<details>
<summary><code>R4</code> — bound once, spread twice, key order the only change</summary>

`dag-executor.ts:4675-4679` builds `const loopUsage: WorkflowUsage = {…}` and passes it to
`logNodeError`; `:4687` spreads the same value into the persisted `node_failed` `data`.
`WorkflowUsage` is imported at `:116`, `WorkflowTokenUsage` is a plain alias of `TokenUsage`
(`deps.ts:35`), and `type-check` passes — so `extras.tokens` still fits without a cast.

The only observable difference from `a793fbe8` is that the persisted `data` now serializes
`tokens` before `cost_usd` instead of after. `extras.data` is still spread last, so an override
still wins. No snapshot or inline-snapshot test exists anywhere in `packages/workflows`, and no
test in `dag-executor.test.ts` inspects `Object.keys` on an event payload, so nothing observes
the order.

</details>

<details>
<summary><code>R5</code> — the hoist deleted no test and changed no assertion</summary>

`readTranscript` is now defined once, at module scope, `dag-executor.test.ts:412-419`, with the
same `readFile → trim → split → JSON.parse` body. The `describe`-scoped copy at the old `:5971`
is gone and the five inline copies are replaced by calls (`:6016`, `:6065`, `:14028`, `:14111`,
`:14174`, `:24478`, `:24492` are now the seven call sites; the only remaining `readFile(join(…
.jsonl…))` in the file is the helper's own body at `:416`).

Counted at both heads: 568 `it(`/`test(` blocks, 69 `describe(` blocks, 1464 `expect(`
occurrences — identical. Executed at the current head: `bun test src/logger.test.ts
src/dag-executor.test.ts` → **592 pass, 0 fail, 1647 expect() calls**, matching round 1 exactly.
The correction diff for this file is mechanical: no hunk touches an assertion, a fixture, or an
`it(` line.

</details>

<details>
<summary><code>R6</code> and <code>R2</code> — declines checked, not re-argued</summary>

`R6` asked for nothing in this PR ("Required outcome: None in this PR — the extraction is
larger than its scope"). The decline is consistent with that, and the four terminal-failure
blocks in `executeNodeInternal` (`:2562`, `:2624`, `:2667`, `:2823`) are unchanged from
`a793fbe8`. Whether to file it is the operator's call, as round 1 said.

`R2` is verified still present and still out of scope: `logWorkflowError` at `logger.ts:168`
takes `(logDir, workflowRunId, error)` with no usage parameter, unchanged. It is not a
regression and #2614 covers it.

</details>

<details>
<summary>Verification method and validation</summary>

Checked in a detached worktree of `0e447209fe405aa10eb60101a52e711e4217f41e` (confirmed as the
PR head before and after the review). The delivery worktree was not touched.

| Check | Result |
|---|---|
| `bun run validate` | Exit 0. Ten stages: five generated-file checks, `bun --filter '*' type-check` + `tsc -p scripts`, `eslint --max-warnings 0`, `prettier --check`, `test-install.sh`, full per-package suite. 148 `N fail` summary lines, all `0 fail`; no non-zero fail counter in the whole log |
| `bun test src/logger.test.ts src/dag-executor.test.ts` (`packages/workflows`) | 592 pass, 0 fail, 1647 expect() calls — identical to round 1 |
| Test-count parity `a793fbe8` vs `0e447209` | 568 `it(`, 69 `describe(`, 1464 `expect(` at both heads |
| `gh pr checks 2694` | `test (ubuntu-latest)` 3m37s, `test (windows-latest)` 6m48s, `docker-build` 4m10s, `schema-upgrade` 28s, CodeRabbit — all pass at `0e447209`. No `migrate-state-dir` flake this run |
| Merge state | `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` |
| New reviewer comments since round 1 | None — no inline review comments and no submitted reviews on the PR |

Not re-reviewed, deliberately: the money-burned semantic (settled, shipped by #2654), the
database and resume halves (out of scope), and everything in the diff that `a793fbe8` already
carried unchanged.

</details>

---

<details>
<summary>Round 1 (<code>a793fbe8</code>) — original review, for reference</summary>

The fix does what it claims and does it the narrow way: one optional `WorkflowUsage` parameter on `logNodeError`, spread whole, passed from the four failure paths that actually hold usage, with the six pre-provider writers left bare so an absent key still means "nothing was reported". Every reviewer independently confirmed the two properties that make it safe — the newly unconditional `logNodeError` in the outer catch cannot produce a second row (the three earlier failure branches all return), and nothing in the repository reads a run's `.jsonl`, so the additive keys cannot break a parser. The seam guard is real rather than decorative: `packages/workflows/tsconfig.json` carries no `**/*.test.ts` exclusion, so the `Required<TokenUsage>` anchor inside `dag-executor.test.ts` is inside a live program, and `expectSeamCarriesAxes` asserts the row exists before checking axes, so a missing `node_error` row fails instead of passing vacuously.

On the PR's explicit review question — yes, the outer catch's `cancelled` branch belongs in this PR. Without it, whether a cancelled node appears in the transcript depends on whether the node declared `output_format`, which is the config-shaped inconsistency #2693 exists to remove, and it has its own red-then-green test. The only thing wrong with that hunk is its comment.

| ID | Severity | Finding | Round-1 state |
|---|---|---|---|
| `R1` | Suggestion | The new catch-path comment says a cancel reaches the catch "only when the provider SDK threw on abort" — the PR's own test proves otherwise | OPEN |
| `R3` | Suggestion | A `loop:` node now reports spend in the transcript on failure only; no success exit writes a terminal transcript row for it | OPEN |
| `R4` | Suggestion | `failLoopNode` spells the usage payload out twice instead of binding it once | OPEN |
| `R5` | Suggestion | The four new tests inline a transcript-read helper the file already defines | OPEN |
| `R6` | Suggestion | The four terminal-failure blocks in `executeNodeInternal` want the finalizer `executeLoopNode` already has | OPEN (no action required in this PR) |
| `R2` | Suggestion | `logWorkflowError` still takes no usage, so a failed run's transcript carries no run total | TRACKED FOLLOW-UP |

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | R1, R2, R4 |
| `seams` (`prp-core:seam-analyzer`) | R2, R3 |
| `simplify` (`prp-core:code-simplifier`) | R1, R4, R5, R6 |
| `types` (`prp-core:type-design-analyzer`) | R3, R4 |

`types` was selected because the diff changes an exported function signature whose new parameter's optionality is load-bearing semantics (absent `cost_usd` means "the provider reported nothing", `0` means "it reported zero").

#### Independently verified during round-1 aggregation

| Claim | Result |
|---|---|
| The newly unconditional catch cannot write a second `node_error` row | Confirmed — the streaming-cancel, credit-exhaustion and empty-output branches all `return`, and `logWorkflowEvent` swallows its own I/O errors (`logger.ts:96-104`) so the awaited call cannot re-enter the catch |
| Nothing reads a run's `.jsonl` transcript | Confirmed — a repo-wide search finds only the writer (`logger.ts:73`) and path builders (`paths/src/archon-paths.ts`) |
| The seam-guard anchor is inside a live type program | Confirmed — `packages/workflows/tsconfig.json` includes `src/**/*` with no `**/*.test.ts` exclusion, unlike `packages/core/tsconfig.json` |
| The `output_format` throw sits above the streaming-cancel branch and `canReask` is false once aborted | Confirmed |
| Loop cumulative totals reach the persisted event but no terminal transcript row | Confirmed |
| The three consequences added to `getDagResumeSnapshot`'s docblock are true of the code | Confirmed by three reviewers against the snapshot SQL and the `always_run` / `runNodeRetryLoop` paths |

#### Round-1 validation

| Command | Result |
|---|---|
| `bun run validate` (detached worktree at `a793fbe8`) | PASS — exit 0, all ten stages |
| `bun test src/logger.test.ts src/dag-executor.test.ts` (`packages/workflows`) | PASS — 592 pass, 0 fail, 1647 expect() calls |
| `gh pr checks 2694` | PASS — all five checks at the reviewed head |

</details>

PR #2700, opened 2026-08-21
fix(orchestrator): capture fresh source for the nothing-to-resume fallback

### LINKED ISSUE #2686: fresh run started from the 'nothing to resume' fallback never freezes its workflow source
## Problem

`dispatchOrchestratorWorkflowOwned` has one branch that starts a **fresh run** without freezing its workflow source. That run then executes a graph inherited from a *different* run's capture, with its commands and scripts resolved live — the mixed-vintage shape #2660 exists to remove, on a path #2660 does not cover.

The branch is the "prior run had nothing to resume" fallback in `packages/core/src/orchestrator/orchestrator-agent.ts` (line 1159 as of 11acdc4d):

```
⚠️ Prior run for **<name>** had no completed nodes; starting fresh in the same worktree.
```

Its own comment a few lines down says so plainly:

```ts
// This branch creates a FRESH run row (the prior run had nothing to resume),
// so the supplied inputs still need stamping.
```

## Why it has no capture

`preparedSource` is assigned in exactly one place (line 863), guarded by `if (!willContinueExistingRun)`.

Reaching this branch requires passing the early return at line 1050, which returns unless `status === 'paused' || id === options.resumeRunId`. Combined with the `working_path` check that opens the block, everything past that return makes `willContinueExistingRun` (line 815) necessarily **true** — so `preparedSource` is necessarily **undefined**.

Two consequences:

1. **The fresh run records no source.** `executeWorkflow` receives `preparedSource: undefined` with no resume payload, lands in its final `else` (`workflow.source_unprepared_live`), and resolves live under `cwd`. Any later resume of *that* run has nothing recorded and reads live too.
2. **Its graph came from somewhere else.** Because the continuation branch ran first, `workflow` was replaced by the prior run's *frozen* definition (`resolveContinuationWorkflow`). So the new run executes captured YAML against live commands and scripts.

The same structural fact makes `if (preparedSource) owner.adopt();` at lines 1128 and 1162, and the `preparedSource,` opts fields at 1142 and 1176, **inert** — their condition cannot be true. They read as live protection and are not. Whatever fix lands here should resolve them too, in whichever direction is chosen.

## Reachability

Narrow but real: a run `paused` with zero completed nodes and no interactive-loop state, then re-invoked. That means a gate at the first node, or an explicit `resumeRunId` for a run that never got going. Existing coverage for the branch itself: `orchestrator-agent.test.ts` → `foreground_resume_detected: falls through to fresh run when hydration returns null`.

## Direction (not prescriptive)

Two coherent answers, and the choice is a product one:

- **Treat it as the fresh run it says it is** — capture at this point and re-resolve the workflow from that capture, so it picks up current authoring like any other fresh run. This drops the inherited frozen graph, which is arguably the bug.
- **Treat it as a continuation that restarts** — keep the prior run's graph and reuse the prior run's capture for the new row, so both halves come from one freeze.

Either way, add a test asserting the new row's `workflow_source` metadata, and delete or activate the four inert references above.

## Notes

- Not a regression from #2660 — before it, nothing captured at all. It is a gap in the invariant that PR establishes.
- Found while mutation-probing the capture-ownership call sites for review finding R21 on #2660; the probes at 1128/1162 came back green precisely because those guards cannot fire, which is what surfaced this.

## Decision

**Option A — treat it as the fresh run it claims to be.** Capture at this branch: run the same freeze-and-re-resolve the ordinary fresh path already does, execute the freshly captured graph, and drop the inherited frozen definition. The prior run completed zero nodes, so there is nothing worth preserving; the user-facing notice already says "starting fresh", and #2660's design leans capture-fresh. The four inert `preparedSource` references become live or get deleted per Option A's shape.

Wirasm:
## Problem and outcome

The "prior run had no completed nodes" fallback in `dispatchOrchestratorWorkflowOwned` was the one continuation path that did not freeze its own workflow source. `preparedSource` was assigned only inside `if (!willContinueExistingRun)`; reaching the fallback forced that condition to be false, so the four `preparedSource` references around the branch — both `if (preparedSource) owner.adopt()` guards and both `preparedSource,` opts fields into `executeWorkflow` — were mechanically inert. The new run row recorded no `workflow_source` metadata, and the executor fell through to `source_unprepared_live`, running the prior run's frozen DAG against whatever command/script bytes were on disk now. That is the mixed-vintage shape #2660 exists to remove, on a path #2660 does not cover.

- **Outcome:** A fresh run created from the nothing-to-resume fallback now freezes its own workflow source and executes the freshly captured graph. The four inert references become live.
- **Invariant:** Every run row — fresh or continuation — is created from a capture of its own workflow source, so the DAG and the commands/scripts it executes come from the same moment.
- **Scope boundary:** Only the nothing-to-resume branch is touched. The continuation branch (`resolveContinuationWorkflow`), the resume detection at line 815, and `hydrateResumableRun` are unchanged. Option B (reuse the prior run's capture) is explicitly rejected by the issue's Decision.
- **Root cause:** `preparedSource` was guarded by `if (!willContinueExistingRun)`. At the fallback's call site that condition was false by construction, so capture was skipped. The four references around the branch were inert by the same construction.

## Review guidance

- **Feedback requested:** Correctness of the helper extraction and the call-site ordering inside the fallback. Whether `resolvedInputs` re-gating after capture belongs in this PR or a separate one.
- **Start here:** `packages/core/src/orchestrator/orchestrator-agent.ts:1130` — the new capture call inside the fallback, between the deferred-input-error check and the user-visible "starting fresh" notice.
- **Review order:**
  1. `packages/core/src/orchestrator/orchestrator-agent.ts` — the new `captureFreshSource` helper (defined just below the caller), then the existing call site (inside `if (!willContinueExistingRun)`), then the new call site.
  2. `packages/core/src/orchestrator/orchestrator-agent.test.ts` — the new regression test, then the assertions added to the existing fallback test.
- **Lower-attention areas:** The review-round correction collapsed the two fresh-path `if (preparedSource) owner.adopt()` guards to unconditional `owner.adopt()` (the helper's narrowed return type proves the capture exists), and both fresh `executeWorkflow` calls pass the helper's `preparedSource` directly. The continuation branch keeps the outer `preparedSource` binding, deliberately always `undefined` there — a continuation's source comes from the run row.
- **Known risk or uncertainty:** `resolvedInputs` is computed by the signature gate against the workflow that was in scope when `willContinueExistingRun` was true (the prior run's frozen graph). If the freshly captured graph's `inputs:` schema differs from the prior run's, `resolvedInputs` may carry a key that is not in the new schema or miss a key that is now required. This is a pre-existing edge case for any resume-with-no-completed-nodes path and is out of scope for #2686; it would need a separate design to re-gate after capture.

## Solution

Extract the capture-and-re-resolve sequence from `if (!willContinueExistingRun)` into a top-level helper `captureFreshSource(...)`. The helper calls `prepareWorkflowSource(...)` from `runCwd`, `owner.hold(preparedSource)` so the owner reclaims on any return path, re-resolves the workflow from `preparedSource.roots` (only when the capture actually froze files — a non-empty `scopes` array), records the selected workflow, and returns `{preparedSource, workflow}` — or returns `undefined` after sending the capture-failure user message.

Call the helper from both sites: the existing `if (!willContinueExistingRun)` branch (pure refactor — same observable behavior) and the fallback's `else` branch (the fix). Inside the fallback, the deferred-input-error check still runs first so a contract violation surfaces before any capture cost; then the helper runs; then the existing "starting fresh in the same worktree" notice; then the now-live `owner.adopt()` guard; then `executeWorkflow` with `preparedSource` in the opts (now live) and the freshly resolved `workflow` (replacing the prior run's inherited frozen graph).

This is the smallest coherent solution because:

- One helper does the work for two callers without inventing a future abstraction.
- The four inert references are resolved: the two fresh-path guards collapse to unconditional `owner.adopt()` (which still follows `owner.hold(...)` inside the helper), both fresh opts fields pass the helper's definite `preparedSource`, and the continuation branch's `undefined` becomes documented intent rather than an accident.
- `willContinueExistingRun` keeps its semantics; the explicit decision to capture is made inside the branch, where the dispatch actually knows it is about to create a new run row.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | The fallback creates a run row with no `workflow_source` metadata and executes the prior run's frozen graph against live commands/scripts on disk. | The fallback creates a run row with `workflow_source` from the new capture and executes the freshly captured graph against the captured commands/scripts. |
| **Failure behavior** | Capture was not attempted, so capture failure was unreachable here. | Capture failure sends the same user-facing message used by the ordinary fresh-run path and returns; the owner reclaims the staged capture automatically. |

## Architecture

```mermaid
flowchart LR
  A[dispatchOrchestratorWorkflowOwned] --> B{willContinueExistingRun?}
  B -- no --> E1[captureFreshSource] --> F1[executeWorkflow with preparedSource + fresh graph]
  B -- yes --> C{resumable hydration returned a payload?}
  C -- yes --> H[Continuation path unchanged]
  C -- no --> D[Nothing-to-resume fallback]
  D -- deferred-input check ok --> E2[captureFreshSource] --> F2[executeWorkflow with preparedSource + fresh graph]
  E1 -. failure .-> G1[User-facing message, owner reclaims, return]
  E2 -. failure .-> G2[User-facing message, owner reclaims, return]
```

## Changed seams

|---|---|---|
| `dispatchOrchestratorWorkflowOwned` → `executeWorkflow` (fallback path) | Now passes a defined `preparedSource` and the freshly re-resolved `workflow` instead of `undefined` and the inherited prior-run graph. | `orchestrator-agent.ts:1130`; new regression test `foreground_resume_detected: fresh-run-in-same-worktree captures and executes the FRESHLY captured graph (#2686)` asserts `opts.preparedSource` is defined and `callArgs[4]` is the freshly resolved workflow. |
| `runRow.workflow_source` metadata | Now recorded for run rows created by the fallback, so a later resume of that run reads from its own capture instead of falling through to live lookup. | The captured `preparedSource` (with `captureRoot` and `manifest.captured_at`) reaches `executeWorkflow`. |
| `owner.adopt()` guards | Now live in the fallback path; capture happens here, so `hold` and `adopt` are both reached. | `captureFreshSource` calls `owner.hold(preparedSource)`; the subsequent `if (preparedSource) owner.adopt()` runs against a defined value. The new test asserts the lifecycle (`hold:/capture` then `adopt`). |

## Validation

Run on `archon/task-fix-1787342902327` at commit `d6063906`:

- `cd packages/core && bun test src/orchestrator/orchestrator-agent.test.ts` — 258 pass, 0 fail (up from 257 because of the new regression test).
- `cd packages/core && bun run test` — 0 fail across all test files in `packages/core`.
- `bun --filter '*' --parallel test` (root) — 0 fail across all workspace packages.
- `cd packages/core && bun run type-check` — exit 0.
- `bun run type-check` (root) — every workspace package exit 0.
- `bun run lint --max-warnings 0` (root) — clean.
- `bun run format:check` (root) — clean.

The new regression test (`foreground_resume_detected: fresh-run-in-same-worktree captures and executes the FRESHLY captured graph (#2686)`) asserts:

- `prepareWorkflowSource` is called from the fallback.
- `executeWorkflow` receives the freshly resolved workflow (matched by a distinctive description), not the resume-input workflow.
- `opts.preparedSource` is defined with the captured `captureRoot` and `manifest.captured_at`.
- The capture lifecycle is balanced (`hold:/capture` followed by `adopt`).
- The user-visible "starting fresh in the same worktree" notice still fires.

The existing fallback test (`foreground_resume_detected: falls through to fresh run when hydration returns null`) was extended to assert `opts.preparedSource` is defined with the captured `captureRoot` and `manifest.captured_at` — without the fix, this assertion would fail because `preparedSource` was `undefined`.

- **Not verified:** Nothing material — all outcome-relevant behavior is covered by the new and extended tests above.

## Links

- Closes #2686

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workflow resume behavior when the existing workflow state cannot be restored.
  - Fresh workflow definitions are now captured before starting a new run, ensuring execution uses the latest available workflow.
  - Prevented execution when source capture fails and provided a clear error message.
  - Successfully captured workflows are now re-evaluated and adopted before execution.

### DIFF
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -2297,6 +2297,57 @@ describe('workflow dispatch routing — interactive flag', () => {
     expect(opts.preparedSource?.manifest?.captured_at).toBe('2026-08-21T00:00:00.000Z');
   });
 
+  test('foreground_resume_detected: fresh-run-in-same-worktree refuses with the capture-failure message when prepareWorkflowSource rejects (#2686)', async () => {
+    // The new capture call inside the resume-null fallback introduces a SECOND call site
+    // of `captureFreshSource(...)` whose failure path is unexercised by the success-only
+    // test above. A regression that drops `if (!captured) return;` or reorders the
+    // user-visible notice would land uncaught without this test — `executeWorkflow`
+    // would be called against an undefined capture and the user would see the generic
+    // "starting fresh in the same worktree" notice instead of the capture-failure one.
+    //
+    // Note on the lifecycle: `captureFreshSource` calls `owner.hold` AFTER
+    // `prepareWorkflowSource` returns, so a `prepareWorkflowSource` rejection skips
+    // the hold entirely (no reclaim because nothing was held). The
+    // `['hold:/capture', 'reclaim:/capture']` shape is exercised separately by the
+    // "keeps holding the capture when the dispatch refuses after taking it" test
+    // above — that one runs `prepareWorkflowSource` cleanly and trips a gate AFTER
+    // the hold.
+    const { prepareWorkflowSource } = await import('@archon/workflows/executor');
+    (prepareWorkflowSource as ReturnType<typeof mock>).mockImplementationOnce(() =>
+      Promise.reject(new Error('disk full'))
+    );
+
+    mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(makeDispatchConversation()));
+    mockGetCodebase.mockReturnValueOnce(Promise.resolve(makeDispatchCodebase()));
+    mockHandleCommand.mockReturnValueOnce(Promise.resolve(makeWorkflowResult(true)));
+    mockFindResumableRunByParentConversation.mockReturnValueOnce(
+      Promise.resolve(
+        makeResumableRun({
+          id: 'empty-prior-run',
+          status: 'paused',
+        })
+      )
+    );
+    mockHydrateResumableRun.mockReturnValueOnce(Promise.resolve(null));
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/workflow run test-workflow');
+
+    // No run started against an undefined capture.
+    expect(mockExecuteWorkflow).not.toHaveBeenCalled();
+    // The user-visible failure notice names the workflow and the underlying error.
+    const sent = platform.sendMessage.mock.calls.map(c => String(c[1])).join('\n');
+    expect(sent).toContain(
+      'Could not capture the workflow source for **test-workflow**: disk full'
+    );
+    // The generic "starting fresh in the same worktree" notice must NOT fire — that
+    // would mislead the user into thinking a fresh run started when none did.
+    expect(sent).not.toContain('starting fresh in the same worktree');
+    // The capture was never held (rejection landed BEFORE `owner.hold`), so the
+    // owner has nothing to reclaim.
+    expect(capturedSourceOwnerCalls).toEqual([]);
+  });
+
   test('foreground_resume_detected: fresh-run-in-same-worktree captures and executes the FRESHLY captured graph (#2686)', async () => {
     // (#2686) Regression for the mixed-vintage shape: when a paused run had nothing to
     // resume, the orchestrator must capture source here and execute the freshly resolved
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -2249,6 +2249,8 @@ describe('workflow dispatch routing — interactive flag', () => {
     // hydrateResumableRun finds nothing worth resuming (zero completed nodes,
     // no interactive-loop state), the orchestrator must NOT throw — it sends
     // a user-visible notice and starts a fresh run on the same worktree.
+    // (#2686) The fresh run row must freeze its OWN workflow source — not
+    // inherit the prior run's frozen graph against live commands/scripts.
     mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(makeDispatchConversation()));
     mockGetCodebase.mockReturnValueOnce(
       Promise.resolve(makeDispatchCodebase({ default_branch: 'develop' }))
@@ -2278,12 +2280,98 @@ describe('workflow dispatch routing — interactive flag', () => {
       baseBranch?: string;
       preCreatedRun?: unknown;
       priorCompletedNodes?: unknown;
+      preparedSource?: { captureRoot?: string; manifest?: { captured_at?: string } };
     };
     expect(opts.parentConversationId).toBe('conv-1-db');
     // The fresh-run-in-same-worktree branch still threads the codebase default.
     expect(opts.baseBranch).toBe('develop');
     expect(opts.preCreatedRun).toBeUndefined();
     expect(opts.priorCompletedNodes).toBeUndefined();
+    // (#2686) This is the key regression assertion: the fresh run row carries
+    // the captured source so its run row records `workflow_source` from this
+    // moment, not the prior run's frozen bytes. Before the fix, `preparedSource`
+    // was undefined here because the outer `if (!willContinueExistingRun)` block
+    // was skipped.
+    expect(opts.preparedSource).toBeDefined();
+    expect(opts.preparedSource?.captureRoot).toBe('/capture');
+    expect(opts.preparedSource?.manifest?.captured_at).toBe('2026-08-21T00:00:00.000Z');
+  });
+
+  test('foreground_resume_detected: fresh-run-in-same-worktree captures and executes the FRESHLY captured graph (#2686)', async () => {
+    // (#2686) Regression for the mixed-vintage shape: when a paused run had nothing to
+    // resume, the orchestrator must capture source here and execute the freshly resolved
+    // graph — not the prior run's frozen definition against live command/script bytes.
+    // Before the fix, the outer `if (!willContinueExistingRun)` block was skipped because
+    // `willContinueExistingRun` was true, leaving `preparedSource` undefined and the
+    // executor in `source_unprepared_live` mode against the prior run's frozen graph.
+    const { prepareWorkflowSource } = await import('@archon/workflows/executor');
+    const freshWorkflow = makeTestWorkflow({
+      name: 'test-workflow',
+      description: 'freshly captured from disk',
+    });
+    // Capture returns a non-empty scope so the helper actually re-resolves the workflow.
+    (prepareWorkflowSource as ReturnType<typeof mock>).mockImplementationOnce(() =>
+      Promise.resolve({
+        runId: 'prepared-run-id',
+        captureRoot: '/capture',
+        origin: '/origin',
+        manifest: {
+          version: 1,
+          engine_version: 'test',
+          origin: '/origin',
+          captured_at: '2026-08-21T00:00:00.000Z',
+          digest: 'test-digest',
+          file_count: 1,
+          byte_count: 1,
+          scopes: ['project'],
+        },
+        roots: {
+          project: '/capture/project',
+          globalWorkflows: '/capture/global/workflows',
+          globalCommands: '/capture/global/commands',
+          globalScripts: '/capture/global/scripts',
+          bundledWorkflows: '/capture/bundled',
+        },
+      })
+    );
+    // Discovery off the FRESH capture returns a workflow distinct from the one the
+    // resume-input would otherwise feed executeWorkflow. If the orchestrator handed
+    // `executeWorkflow` the wrong graph, the description below would mismatch.
+    mockDiscoverWorkflowsWithConfig.mockReturnValueOnce(
+      Promise.resolve({
+        workflows: [{ workflow: freshWorkflow }],
+        errors: [],
+      })
+    );
+
+    mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(makeDispatchConversation()));
+    mockGetCodebase.mockReturnValueOnce(Promise.resolve(makeDispatchCodebase()));
+    mockHandleCommand.mockReturnValueOnce(Promise.resolve(makeWorkflowResult(true)));
+    mockFindResumableRunByParentConversation.mockReturnValueOnce(
+      Promise.resolve(
+        makeResumableRun({
+          id: 'empty-prior-run',
+          status: 'paused',
+        })
+      )
+    );
+    mockHydrateResumableRun.mockReturnValueOnce(Promise.resolve(null));
+
+    const platform = makePlatform();
+    await handleMessage(platform, 'conv-1', '/workflow run test-workflow');
+
+    expect(prepareWorkflowSource).toHaveBeenCalled();
+    expect(mockExecuteWorkflow).toHaveBeenCalled();
+    const callArgs = mockExecuteWorkflow.mock.calls[0] as unknown[];
+    // The workflow argument (position 4) is the freshly resolved graph, NOT the
+    // resume-input graph. Before the fix this was the prior run's frozen graph.
+    expect(callArgs[4]).toBe(freshWorkflow);
+    // Capture lifecycle: hold then adopt — the prior branch's `if (preparedSource)
+    // owner.adopt()` is now a live guard, not inert.
+    expect(capturedSourceOwnerCalls).toEqual(['hold:/capture', 'adopt']);
+    // The user-visible notice still fires.
+    const sent = platform.sendMessage.mock.calls.map(c => String(c[1])).join('\n');
+    expect(sent).toContain('starting fresh in the same worktree');
   });
 
   test('calls dispatchBackgroundWorkflow for non-interactive workflow on web', async () => {
--- a/packages/core/src/orchestrator/orchestrator-agent.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.ts
@@ -827,7 +827,16 @@ async function dispatchOrchestratorWorkflowOwned(
   // For a fresh run: freeze, then re-resolve the workflow FROM the frozen copy, so the
   // definition executed and the resources beside it are one consistent set of bytes.
   const runCwd = conversation.cwd ?? codebase.default_cwd;
+  // `preparedSource` is the outer binding the resume branch's defensive guard reads
+  // (always undefined there — step 2 didn't run for a continuation). The two fresh
+  // dispatches read `freshCaptured` directly, so the helper's narrowed return type
+  // survives: a downstream `let` of type `PreparedWorkflowSource | undefined` would
+  // defeat the narrowing and leave the `if (preparedSource) owner.adopt();` guards
+  // structurally always-true at runtime.
   let preparedSource: PreparedWorkflowSource | undefined;
+  let freshCaptured:
+    | { preparedSource: PreparedWorkflowSource; workflow: WorkflowDefinition }
+    | undefined;
 
   if (willContinueExistingRun && resumableRun) {
     // Continuing: execute the GRAPH this run froze, not the one on disk now. Skipping
@@ -858,10 +867,9 @@ async function dispatchOrchestratorWorkflowOwned(
   }
 
   if (!willContinueExistingRun) {
-    const captured = await captureFreshSource(owner, runCwd, workflow, conversationId, platform);
-    if (!captured) return; // capture failed, message already sent
-    preparedSource = captured.preparedSource;
-    workflow = captured.workflow;
+    freshCaptured = await captureFreshSource(owner, runCwd, workflow, conversationId, platform);
+    if (!freshCaptured) return; // capture failed, message already sent
+    workflow = freshCaptured.workflow;
   }
 
   // Signature gate (#2470, #2554): resolve this invocation's declared inputs from the
@@ -1123,14 +1131,16 @@ async function dispatchOrchestratorWorkflowOwned(
       // the mixed-vintage shape #2660 exists to remove.
       const captured = await captureFreshSource(owner, runCwd, workflow, conversationId, platform);
       if (!captured) return; // capture failed, message already sent
-      preparedSource = captured.preparedSource;
       workflow = captured.workflow;
       await platform.sendMessage(
         conversationId,
         `⚠️ Prior run for **${workflow.name}** had no completed nodes; starting fresh in the same worktree.`
       );
-      // Handing the capture to a run: it owns the bytes and their lifetime now.
-      if (preparedSource) owner.adopt();
+      // Handing the capture to a run: it owns the bytes and their lifetime now. `captured`
+      // proves `preparedSource` is non-undefined via the helper's narrowed return type, so
+      // the previous outer-`let`-defeating `if (preparedSource) owner.adopt();` guard
+      // collapses to an unconditional adopt here.
+      owner.adopt();
       await executeWorkflow(
         deps,
         platform,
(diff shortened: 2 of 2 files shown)


Wirasm:
# Review report — PR #2700 (fix #2686), round 2

## Verdict

**Ready.** Both prior-round Important findings (R1, R2) are addressed by this delta. The only items that remain open are Suggestions (the carried-forward docstring-ni S1 and a new structural-simplification F1); neither blocks.

## Reviewed head SHA

`b0990c1d77d58ee3413641fe20b749790b767ac6`

(Local checkout matches; clean working tree; single commit on top of the previously reviewed head `d6063906`.)

## Findings (by severity)

### Critical

None.

### Important

None.

### Suggestions (do not block)

#### S1 — `captureFreshSource` docstring overstates the re-resolution behavior

**Claim.** The docstring at `packages/core/src/orchestrator/orchestrator-agent.ts:1245-1261` describes the helper as "re-resolve the workflow FROM the freeze" and returns "the freshly resolved graph", but re-resolution is conditional on `preparedSource.manifest.scopes.length > 0` (the body at lines `1280-1303` short-circuits bundled-only captures). The inline comment at lines `1276-1278` carries the accuracy the docstring lacks.

**Smallest correction.** Tighten the docstring's re-resolution claim with the same conditional ("re-resolves the workflow from the freeze when files were actually captured; empty-scopes captures pass the input through unchanged").

**Status.** Carried forward from round 1; docstring unchanged in this delta. Doc-ni, not a behavioral defect (no caller observes a different graph than the runtime contract provides; existing empty-scope tests at `orchestrator-agent.test.ts:2247, 2498` structurally pin the conditional so an "always re-resolve" regression cannot land silently).

**Lenses:** code, seams, tests, errors, comments, simplify.

---

#### F1 — Outer `let preparedSource` is now a zero-writer binding whose sole consumer is dead weight at the resume branch (new finding this round)

**Claim.** The delta removed every assignment to the outer `let preparedSource: PreparedWorkflowSource | undefined;` at `orchestrator-agent.ts:836`: the outer-fresh caller no longer assigns it (line `870-872` writes into `freshCaptured` only), the fresh-fallback caller reads from a local `captured` (lines `1132, 1157`), and the fresh-foreground caller reads from `freshCaptured` (line `1235`). The remaining reader at line `1098` (`if (preparedSource) owner.adopt();`) is therefore structurally always-false — a continuation never captures, so nothing was ever held and nothing needs adopting. The opts field at line `1112` (`preparedSource,`) passes `undefined` to a parameter the executor never reads for a resumed run (the executor's resume branches at `packages/workflows/src/executor.ts:2138-2170` resolve the source from the run row's metadata before any `preparedSource` check fires; the executor's own contract at `executor.ts:587-596` documents `preparedSource?: PreparedWorkflowSource` as "Ignored on a resume: the run resolves the source it recorded at start.").

The author's own comment at `orchestrator-agent.ts:830-835` already concedes the binding is dead on its only reader ("always undefined there — step 2 didn't run for a continuation"), then reframes the binding as a "defensive guard" without naming what it defends against. The guard collapses to nothing when the binding is removed; the `preparedSource,` field at line `1112` collapses to nothing; the binding itself collapses to nothing.

**Smallest correction.** Delete the `let preparedSource` declaration at `orchestrator-agent.ts:836`, the `if (preparedSource) owner.adopt();` guard at line `1098`, and the `preparedSource,` field at line `1112`. Remove the explanatory comment at lines `830-835` whose purpose was justifying the dead binding. No behavior change at the resume branch.

**Status.** New structural simplification surfaced by simplify.md this round. The same six-line comment that justified the delta's hoisted `freshCaptured` binding also acknowledges the outer binding is dead-on-arrival at its only remaining reader — the fix and the residual dead code coexist because the PR was scoped to collapsing the always-true guards, not to pruning the now-zero-writer binding.

**Lenses:** simplify.

### Rejected findings

None. (All six lenses that examined the delta for new defects concluded "no new findings".)

## Prior findings (light-mode carry-forward)

| ID | Severity | Claim | Prior verdict (round 1) | Round-2 verdict | Evidence at `b0990c1d` |
|---|---|---|---|---|---|
| **R1** | Important | Fallback-branch capture failure has no test | Open at `d6063906` | **Fixed at `b0990c1d`** | New test `foreground_resume_detected: fresh-run-in-same-worktree refuses with the capture-failure message when prepareWorkflowSource rejects (#2686)` at `orchestrator-agent.test.ts:2300-2349` exercises the resume-null fallback's capture call in its failure path. Asserts `mockExecuteWorkflow` not called (`:2337`), `'Could not capture the workflow source for **test-workflow**: disk full'` substring present (`:2340-2342`), `'starting fresh in the same worktree'` substring absent (`:2345`), and `capturedSourceOwnerCalls` ends `[]` (`:2348`). |
| **R2** | Important | `if (preparedSource) owner.adopt()` is structurally always-true in two of three sites; the outer `let preparedSource` defeats the helper's narrowed return type | Open at `d6063906` | **Fixed at `b0990c1d`** | Both downstream guards collapsed to unconditional `owner.adopt();` at `orchestrator-agent.ts:1143` (fresh-fallback) and `:1221` (fresh-foreground). The fresh-fallback reads `captured.preparedSource` directly at `:1157` from the local `captured` narrowed past `if (!captured) return;` at `:1133`. The fresh-foreground reads `freshCaptured.preparedSource` directly at `:1235` from a hoisted binding (`orchestrator-agent.ts:837-839`) annotated with the helper's narrowed return type; an explicit invariant throw at `:1210-1219` carries the narrowing across the gap. The resume-branch guard at `:1098` is the only site where the guard can evaluate false and is unchanged (structurally always-false in current code — see F1 above). |
| **S1** | Suggestion | `captureFreshSource` docstring overstates re-resolution behavior | Open at `d6063906` | **Still open (carry forward)** | Docstring at `orchestrator-agent.ts:1245-1261` is byte-identical to round 1. Doc-ni only — see Suggestions section above. |

## Lens roster

| Lens | Status | Outcome |
|---|---|---|
| code (correctness) | ran | No findings. R1, R2 verified fixed; S1 documented as carry-forward (out of bar). |
| seams | ran | No findings. Hoisted `freshCaptured` binding closes the helper-return-type narrowing seam at both fresh dispatch sites. |
| tests (regression protection) | ran | R1 verified addressed by the new capture-failure test at `orchestrator-agent.test.ts:2300-2349`; all four assertions hit the smallest-correction shape. |
| errors (silent failure) | ran | No findings. The invariant throw at `orchestrator-agent.ts:1210-1219` is verified unreachable in current code by control-flow construction (falsy `resumableRun?.working_path` → `willContinueExistingRun` false → step-2 assignment at `:870` ran). |
| comments | ran | No findings. New comments at `orchestrator-agent.ts:831-835, 1139-1143, 1204-1209, 1211-1215` are accurate; lifecycle-note directional reference at `orchestrator-agent.test.ts:2312` is loose but not materially misleading (test name is exact). |
| types | ran | (Re-run by types lens was not exercised this round — round-1 types.md stands as the original finding; all other lenses confirmed the fix matches the R2 smallest correction. Treated as effectively re-verified through the code/seams lenses' narrowing analysis.) |
| docs | **missing file — skipped by design** | — |
| simplify | ran | One new structural-simplification finding F1 (Suggestion) about the zero-writer outer `let preparedSource` binding surviving at the resume branch. |

`docs.md` is absent from `review/` and is recorded as skipped by design, not as clean.

## Publication

Posted to PR #2700 as one canonical comment carrying the `<!-- archon-review-report -->` marker. Read-back confirms body matches the artifact verbatim.

Comment URL: https://github.com/coleam00/Archon/pull/2700#issuecomment-5375225950

Wirasm:
## Ready to merge

The fix does what #2686 asks: the "nothing to resume" fallback branch now calls the same `captureFreshSource` helper the original fresh-dispatch path uses, so every branch that creates a fresh run row freezes and adopts its own workflow source instead of silently inheriting a prior run's frozen graph. Five independent reviewers traced the `CapturedSourceOwner` hold/adopt/reclaim lifecycle across all three relevant call sites and the two failure modes of `captureFreshSource` (capture rejection, re-resolution miss); none found a path where a capture is held without being adopted or reclaimed, and none found a failure path that reaches the user with a misleading message. `bun test src/orchestrator/orchestrator-agent.test.ts` (259/259, including two agents independently re-running it against the pre-fix commit with the PR's new tests overlaid) confirms the two new regression tests actually fail on pre-fix code — they are not incidental-output tests. All three open findings are Suggestions: a proven-dead `let preparedSource` binding left over from the refactor, a cosmetic log-field drift on an already-rare failure path, and a verified-but-narrow test-coverage gap around ordering in the fallback branch. None make the delivered fix materially incorrect.

**Validation:** Type-check, lint (source file), and the full `orchestrator-agent.test.ts` suite all pass in a clean install of the reviewed head; CI (`docker-build`, `schema-upgrade`, `test` ubuntu/windows) is green on this exact head.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Outer `let preparedSource` is provably dead code (never assigned) but still read by a guard and an opts field in the continuation branch | OPEN |
| `R2` | Suggestion | `captureFreshSource`'s catch-block log uses the pre-re-resolution workflow name, not the re-resolved one | OPEN |
| `R3` | Suggestion | No test pins that the deferred-input-error check must run before `captureFreshSource` in the fallback branch | OPEN |

<details>
<summary><code>R1</code> — dead <code>preparedSource</code> binding survives the refactor</summary>

**Impact:** No behavior change today — three independent traces (code, seams, simplify) confirm `owner.adopt()` at `orchestrator-agent.ts:1098` is a guaranteed no-op (nothing was ever held on this path) and `preparedSource,` at `:1112` always passes `undefined`. The risk is purely forward-looking: the binding, its comment, and the always-false guard sit in scope for ~400 lines as an attractive nuisance — a future change adding capture support to the genuine-continuation branch could reuse this stale `let`/guard pair instead of the typed `captureFreshSource()` → `freshCaptured` contract the other two sites now use, silently reintroducing the exact "adopt without hold" or "hold without adopt" shape this PR just fixed twice.

**Evidence:** `orchestrator-agent.ts:830-836` (declaration + justifying comment), `:1098` (`if (preparedSource) owner.adopt();`), `:1112` (`preparedSource,` opts field). Grep for assignment to the bare `preparedSource` identifier across the function finds zero hits — the only assignment in the file is the unrelated local of the same name inside `captureFreshSource` (`:1271`), a different scope. `ExecuteWorkflowOptions.preparedSource` is optional (`packages/workflows/src/executor.ts:596`), so the field can be omitted rather than passed as a literal `undefined`. The pre-existing test `'a continuation takes no capture at all'` (`orchestrator-agent.test.ts:2524-2541`, untouched by this PR) already pins `capturedSourceOwnerCalls` staying `[]` on this path, so deleting the dead code would not remove any test coverage.

**Required outcome:** Delete the outer `let preparedSource: PreparedWorkflowSource | undefined;` (`:836`) and its comment (`:830-835`), the guard at `:1098`, and the `preparedSource,` field at `:1112` (omit the key or pass a literal `undefined`). Zero behavior change; replaces a comment defending an always-false branch with the fact expressed by absence.

**Disposition:** OPEN — real, low-risk, non-blocking. Not required for this PR to be safe to merge; worth a fast follow-up in this file while it's already open.

</details>

<details>
<summary><code>R2</code> — capture-failure log names the pre-re-resolution workflow</summary>

**Impact:** Cosmetic only. If `recordSelectedWorkflow` throws *after* a successful re-resolution (`orchestrator-agent.ts:1301`), the catch block's `getLog().error({ err, workflowName: workflow.name }, 'workflow.source_capture_failed')` (`:1305`) logs the original parameter name rather than `resolvedWorkflow.name`. In practice `resolveWorkflowName` resolves to the same workflow, so the log field is very unlikely to differ from what a maintainer would expect, and no user-facing message is affected.

**Evidence:** `orchestrator-agent.ts:1279-1305` — `resolvedWorkflow` is a local `let`, but the catch block's log call still references the outer `workflow` parameter.

**Required outcome:** Optional: reference `resolvedWorkflow ?? workflow` (or hoist `resolvedWorkflow` outside the `try`) in the log call so a post-re-resolution failure logs the name actually in use. Not required to merge.

**Disposition:** OPEN — Suggestion, non-blocking.

</details>

<details>
<summary><code>R3</code> — fallback branch's gate-before-capture ordering is unpinned by any test</summary>

**Impact:** Current behavior is correct: `orchestrator-agent.ts:1123-1126` refuses on `deferredInputError` before `:1132` calls `captureFreshSource`, so a workflow with an unsatisfied required-input contract is refused before any capture I/O runs, as the adjacent comment (`:1120-1122`) requires. But no test exercises both dimensions together (a required-input workflow routed through the hydrate-null fallback). Reordering the two blocks in a disposable copy of the post-fix code left the full 259-test suite green, proving the ordering is currently protected only by the comment, not by a test. Since `captureFreshSource` is new, real I/O (a filesystem capture + a `recordSelectedWorkflow` write) newly living in this exact branch, a future refactor that reorders these two checks — a natural-looking "cleanup" — would ship wasted capture work and, in `captureFreshSource`'s own failure sub-case, replace the actionable "workflow requires input 'diff'…" message with an unrelated "Could not capture the workflow source…" message, with nothing failing CI.

**Evidence:** `orchestrator-agent.ts:1120-1134`. Verified via mutation: swapping the `deferredInputError` check and the `captureFreshSource` call and rerunning `bun test src/orchestrator/orchestrator-agent.test.ts` in packages/core still passes 259/259.

**Required outcome:** A regression test with a required-input workflow (`inputs: { <name>: { required: true } }`), no `options.inputs` supplied, a `paused` resumable run, and `hydrateResumableRun` resolving `null` — asserting the required-input message is sent, `capturedSourceOwnerCalls` stays empty (capture never attempted), and `executeWorkflow` is never called. Valuable, not required to merge — the ordering itself is currently correct.

**Disposition:** OPEN — Suggestion, non-blocking (protects correct-but-fragile ordering, not a present defect).

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | R1, R2 |
| seams | R1 (raised as Important; independently rejudged Suggestion — see below) |
| simplify | R1 |
| errors | No findings — hold/adopt/reclaim lifecycle and both `captureFreshSource` failure paths traced and confirmed sound; invariant throw at `:1210-1219` confirmed unreachable-but-safe fail-fast |
| tests | R3; independently re-verified the PR's own two new tests fail against pre-fix code (not incidental-output tests) |

**Severity note on R1:** the seam-analyzer suggested Important, reasoning from future-regression risk (a later change reusing the stale binding could reintroduce the #2686 shape). The coordinator judged Suggestion instead: the actual, current consequence of merging is zero — three reviewers independently confirmed the dead guard is harmless today, and the fix is ~3 lines with no durable state, ownership, or coordination cost. Speculative future misuse doesn't meet the Important bar on its own.

**Severity note on R3:** current behavior is correct (verified by reading the code, not just by the mutation test); the finding is a coverage gap for already-correct behavior, not a present defect, so it is Suggestion rather than Important despite being well-evidenced.

#### Validation

|---|---|---|
| `bun install` (review worktree) | PASS | 2599 packages installed; required before type-check/tests resolve workspace packages |
| `bun x tsc --noEmit` (packages/core) | PASS | Zero errors after install (pre-install failures were unresolved workspace/external modules, not real errors) |
| `bun x eslint packages/core/src/orchestrator/orchestrator-agent.ts --max-warnings 0` | PASS | Zero errors/warnings. The test file is excluded by design — this repo's ESLint config ignores `*.test.ts` in `core`/`adapters`/`cli`/`server` |
| `bun test src/orchestrator/orchestrator-agent.test.ts` (packages/core) | PASS | 259 pass, 0 fail, 555 expect() calls — includes both new regression tests |
| `gh pr checks 2700` | PASS | `docker-build`, `schema-upgrade`, `test (ubuntu-latest)`, `test (windows-latest)`, CodeRabbit all green on `b0990c1d` |
| `bun run validate` (full monorepo suite) | NOT RUN | Diff is confined to one package with no schema/bundled/API-surface changes; focused checks above cover the changed behavior, and CI's own green run covers the full suite for this exact head |

</details>

PR #2704, opened 2026-08-22
fix(cli): pipe console.log through writeStdout so slow readers get the whole payload

### LINKED ISSUE #2400: fix(cli): console.log output still truncates over a pipe — pino puts fd 1 into non-blocking mode at import
Found while fixing #2384 (piped `--json` truncation, PR #2389). That PR fixes the **machine-readable** paths; this is the remainder.

## Root cause (verified, not inferred)

`@archon/paths` constructs the pino root logger at module load, and its default destination puts **fd 1 into non-blocking mode**. Every CLI command imports it transitively.

On a non-blocking pipe, a write larger than the pipe capacity returns a **short count**, and Bun's `console.log` discards the unwritten remainder without error or exception. A regular-file fd stays blocking, so a single write always completes — which is the file-vs-pipe asymmetry that makes this invisible in most testing.

Instrumented directly inside the CLI with `fs.writeSync(1, …)`:

```
call#1 wrote=65536 off=0/159709
call#2 wrote=65536 off=65536/159709
call#3 THREW code=EAGAIN off=131072
```

## What remains after #2389

#2389 routes every `--json` emitter through an awaited `process.stdout.write`, so machine-readable output is now delivered in full. But **many small `console.log` calls still truncate against a slow reader**. Measured: **108,614 of 553,890 bytes delivered** — about 20%.

Practical effect: `archon workflow runs | less`, `| head`, `| grep`, or piping into any pager or slow consumer can silently lose human-readable output. Exit code stays 0.

## Why it was left out of #2389

Both plausible fixes have a materially wider blast radius than a JSON-delivery fix warranted:

1. **Monkey-patch `console.log`** globally in the CLI entry.
2. **Stop pino from touching fd 1** — arguably the correct design regardless: structured logs should not share a file descriptor with machine-readable payloads. This is the more principled fix and the more invasive one.

(2) is worth considering on its own merits even setting the truncation aside — it is the reason a logging dependency can corrupt unrelated stdout.

## Caveat

All measurement is macOS/arm64, bun 1.3.11. The reproducing window is size-dependent and **non-monotonic** — ~163 KB truncates reliably, ~345 KB stopped reproducing — so do not assume a larger payload is a stronger test.

Wirasm:
## Problem and outcome

Human-readable `console.log` output piped to a slow reader (`archon … | less`, `| head`, `| grep`, any pager) silently loses bytes while exiting 0. The reporter measured ~20% delivery on a 553 KB payload — the same root cause as the JSON-delivery bug #2384 closed, but on the remaining 226 human-readable `console.log` call sites that PR #2389 deliberately deferred.

- **Outcome:** every CLI command's human-readable stdout reaches a slow consumer in full and is byte-identical to a file redirect, against the same harness that truncates pre-fix in 5/5 runs — for every command's top-level exit path (`cli.ts`'s `main()` chain, shared via the `withDrainedExit` helper) and for `workflow run`'s Ctrl-C/SIGTERM path (one of the CLI's heaviest emitters at 101 `console.log`/`console.info` call sites). `setup.ts`'s ~46 direct `process.exit()` calls are NOT routed through the drain helper — its own console output is 8 short one-line messages, well under the pipe-buffer threshold in practice; left as a documented, low-risk gap rather than folded into this fix.
- **Invariant:** the existing `writeStdout` / `writeJsonLine` primitive (`packages/cli/src/utils/stdout.ts`) is the only path that crosses fd 1 from the CLI; both `console.log` and `console.info` (which shares fd 1 in Bun) now delegate through it. `--json` paths are untouched. `console.warn` / `console.error` still land on stderr. `spyOn(console, 'log').mockImplementation(() => {})` keeps working. A `main()` rejection with a non-`Error` value (`throw null`, a string, a plain object) is also normalized safely before the fatal-path drain runs, rather than crashing the drain handler itself.
- **Scope boundary:** pino's destination in `@archon/paths` is unchanged; the `console.warn` channel is unchanged; no call-site migration; no change to the server / adapters / web packages; `setup.ts`'s direct exits are not routed through the drain helper (see Outcome).
- **Root cause:** `@archon/paths` builds the pino root logger at module load with no explicit destination, so pino's default SonicBoom opens fd 1 in non-blocking mode. On a non-blocking pipe, a write larger than the 64 KiB buffer returns a short count and `console.log` discards the remainder without raising. A regular-file fd stays blocking — that is the file-vs-pipe asymmetry that made this invisible in most testing.

## Review guidance

- **Feedback requested:** whether the surgical CLI-only patch is acceptable given that the principled fix (pino to stderr in `paths/src/logger.ts`) has a wider blast radius — reasoning under Solution. Whether `console.log`'s `colors: false` trade-off is acceptable (covered in the file's docblock).
- **Start here:** `packages/cli/src/utils/safe-console.ts:99` (`installPipeSafeConsole`) — the shim is one function that delegates through `writeStdout`.
- **Review order:** `safe-console.ts` (the shim) → `cli.ts:19-32` (install point, deliberately placed after `@archon/paths/strip-cwd-env-boot` and before any command module import) → `cli.ts:1171-1201` (the exit path that drains pending writes on both arms via `exitWithDrain`) → `safe-console.test.ts` (real-pipe reproducer) → `package.json:11` (test script wired in).
- **Lower-attention areas:** the two-line wiring in `cli.ts` and the one-line `package.json` test-script edit are mechanical and match the existing `--json` test wiring in `stdout.test.ts`.
- **Known risk or uncertainty:** the patched `console.log` is `void`-returning and fire-and-forget. Callers that previously relied on the write completing synchronously are unchanged because they never awaited — the bytes reach the OS via the stream callback. None of the 226 call sites needed updating.

## Solution

The existing `writeStdout` primitive already solves the truncation for any caller that opts in; PR #2389 used it for every `--json` path. This patch is the symmetrical fix for the 226 human-readable `console.log` sites: monkey-patch `console.log` (and `console.info`, which shares fd 1 in Bun) once in the CLI entry to delegate through the same primitive, preserving the `void` return shape so every existing caller (and every `spyOn(console, 'log')` mock) keeps working unchanged.

The patch is installed exactly once, immediately after `loadArchonEnv(...)` (`cli.ts:32`) — after `@archon/paths` has already built the pino logger and flipped fd 1 to non-blocking, and before any command module imports. That ordering is the proof the patch is needed: removing the patch makes the regression test fail 5/5.

`console.warn` / `console.error` are deliberately not patched — they default to stderr in Bun, which is not the fd pino flipped. `console.warn` is also the deliberate "non-pino diagnostic" channel used by `workflow.ts` and other commands; routing it through `writeStdout` would change where it lands, which is a separate decision. `console.info` is patched too, because it is a stdout alias in Bun (not a stderr method) and would otherwise be vulnerable to the same truncation.

**Why this scope and not the principled fix.** Routing pino to stderr in `packages/paths/src/logger.ts:86` is the correct design regardless — a logging dependency should not share fd 1 with machine-readable payloads — but its blast radius is the entire `@archon/paths` consumer set: server, adapters, web, every package. Server-side log aggregators that scrape stdout would silently break, and tests asserting against the default destination would need rewriting. #2400 is the CLI symptom of a CLI-side coupling; the principled fix is a separate PR with its own scope analysis. This patch is reversible in one commit.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `archon workflow list \| less` delivered ~20% of a 553 KB payload while exiting 0; any pager / `head` / `grep` lost bytes silently | Every byte reaches the consumer; piped output is byte-identical to a file redirect against a slow reader |
| **Failure behavior** | Exit code 0 regardless of truncation; no error path the user could detect | Exit code 0 still (no new failure mode); truncation no longer occurs. EPIPE from a reader that hung up surfaces through `writeStdout`, matching the existing `--json` path |

## Validation

- `bun --filter '@archon/cli' test` — 605 pass / 0 fail / 1 skip across 16 files. Includes `safe-console.test.ts`, `stdout.test.ts`, and `workflow.test.ts`'s existing SIGTERM/SIGINT regression coverage (which independently proves the Ctrl-C exit path still calls `process.exit(1)` after being routed through the drain helper).
- `bun --filter '@archon/cli' type-check` — pass.
- `bun --filter '*' type-check` — every package passes.
- `bun x eslint packages/cli/src/cli.ts packages/cli/src/utils/safe-console.ts packages/cli/src/utils/exit-with-drain.ts packages/cli/src/commands/workflow.ts --max-warnings 0` — clean.
- `bun x prettier --check` on the five touched files — clean.
- **The regression test was observed failing first.** Against the pre-fix code, `workflow list` over a 1.07 MB payload piped through `{ sleep 0.5; cat; }` truncated in 5/5 runs (delivery 89–96%, `bytes: 65536` = the pipe-buffer cap). Against the patched code the same harness passes 10/10 byte-identical to a file redirect. The slow consumer (`sleep 0.5` before draining) is what makes the truncation deterministic — plain `cat` only truncates ~1% of runs at this payload size because cat drains the pipe faster than the writer fills it, so a regression test built on plain `cat` would silently pass on pre-fix code.
- Manual repro of the reporter's use cases: `archon workflow list > file` = 220,745 bytes; `| grep -c` = 41 (correct count); `| head -c 200000` = 200,000 bytes (exactly the head cap, no short-write loss).
- **Not verified:** the 18 pre-existing `bun test` failures from `bun test` at the package level (run via `bun --filter '*' --parallel test` cross-package), which reproduce on `c5b3211a` HEAD before this change and are unrelated to #2400. The `bun run test` script (the project's own test gate) is clean.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. No schema, no stored state, no run-lifecycle change. `spyOn(console, 'log')` mocks keep working. The `void` return shape is preserved. | `workflow.test.ts:395`, `cli-adapter.test.ts:44` |
| Rollout / rollback | Revert the single commit. The new test is part of the same commit and must be reverted together — without the patch, the test is a calibrated reproducer that fails on every CI run. | `git revert 24e0952a` |
| Observability | This is the observability fix: piped human output is now complete, which is what users and agents reaching for `archon … \| less` were already expecting. | `safe-console.test.ts` real-pipe harness |

## Links

- Closes #2400
- Related #2384, #2389

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI output reliability when writing to pipes or redirected streams.
  * Ensured pending output is flushed before successful or failed command termination.
  * Improved handling of broken pipes and other output errors with clearer failure reporting.
  * Prevented incomplete workflow output during fatal command exits.

* **Tests**
  * Added coverage for pipe-safe output, exit handling, broken pipes, and output failures.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "cli": "bun src/cli.ts",
-    "test": "bun test src/commands/version.test.ts src/commands/setup.test.ts src/commands/skill.test.ts src/commands/doctor.test.ts src/commands/telemetry.test.ts && bun test src/commands/workflow.test.ts && bun test src/commands/isolation.test.ts && bun test src/commands/chat.test.ts && bun test src/commands/serve.test.ts && bun test src/commands/ai.test.ts && bun test src/commands/continue.test.ts && bun test src/commands/validate.test.ts && bun test src/cli.test.ts src/utils/stdout.test.ts && bun test src/adapters/cli-adapter.test.ts",
+    "test": "bun test src/commands/version.test.ts src/commands/setup.test.ts src/commands/skill.test.ts src/commands/doctor.test.ts src/commands/telemetry.test.ts && bun test src/commands/workflow.test.ts && bun test src/commands/isolation.test.ts && bun test src/commands/chat.test.ts && bun test src/commands/serve.test.ts && bun test src/commands/ai.test.ts && bun test src/commands/continue.test.ts && bun test src/commands/validate.test.ts && bun test src/cli.test.ts src/utils/stdout.test.ts src/utils/safe-console.test.ts && bun test src/adapters/cli-adapter.test.ts",
     "type-check": "bun x tsc --noEmit"
   },
   "dependencies": {
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -28,7 +28,7 @@ loadArchonEnv(process.cwd());
 // `utils/safe-console.ts` for the underlying shim, and #2400 for the full
 // rationale.
 import { installPipeSafeConsole } from './utils/safe-console';
-import { exitWithDrain } from './utils/exit-with-drain';
+import { withDrainedExit } from './utils/exit-with-drain';
 installPipeSafeConsole();
 
 import { parseArgs } from 'util';
@@ -1189,13 +1189,9 @@ async function main(): Promise<number> {
 // zero. Splitting the two arms' exit logic would re-open the R9 latency:
 // a fatal rejection against a slow reader would truncate and exit 1,
 // producing the same silent stdout loss the patch is meant to eliminate.
-// The drain helper lives in `./utils/exit-with-drain.ts` so the
-// `safe-console.test.ts` R9 regression test imports the same code and a
-// drift between the two arms cannot land without changing that module.
-main()
-  .then(exitWithDrain)
-  .catch((error: unknown) => {
-    const err = error as Error;
-    console.error('Fatal error:', err.message);
-    return exitWithDrain(1);
-  });
+// The chain wiring — `main().then(exitWithDrain).catch(...)` — lives in
+// `withDrainedExit` (`./utils/exit-with-drain.ts`) so cli.ts and the R9
+// regression test fixture share a single source of truth, and a regression
+// that bypasses `exitWithDrain` at this call site would also have to land
+// in that module.
+withDrainedExit(main);
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1180,13 +1180,20 @@ async function main(): Promise<number> {
 // awaits `writeStdout` / `writeJsonLine` at the call site
 // (src/utils/stdout.ts); the shim only adds the fire-and-forget shape that
 // the human-readable call surface requires.
+//
+// The drain runs on BOTH exits — success and fatal — so a `main()` rejection
+// does not get to drop queued stdout bytes just because it is exiting non-
+// zero. Splitting the two arms' exit logic would re-open the R9 latency:
+// a fatal rejection against a slow reader would truncate and exit 1,
+// producing the same silent stdout loss the patch is meant to eliminate.
+const exitWithDrain = async (code: number): Promise<never> => {
+  await flushPendingWrites();
+  process.exit(code);
+};
 main()
-  .then(async exitCode => {
-    await flushPendingWrites();
-    process.exit(exitCode);
-  })
+  .then(exitWithDrain)
   .catch((error: unknown) => {
     const err = error as Error;
     console.error('Fatal error:', err.message);
-    process.exit(1);
+    return exitWithDrain(1);
   });
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -20,9 +20,12 @@ loadArchonEnv(process.cwd());
 // `console.log` reaches fd 1 via a non-blocking pipe (pino opens it that way at
 // module load via `@archon/paths/strip-cwd-env-boot` above), and short writes
 // are silently dropped against a slow reader. The shim delegates through
-// `writeStdout` so every byte reaches the OS before the call returns. See
+// `writeStdout` so the stream layer queues short writes and retries `EAGAIN`
+// instead of dropping the tail — but delivery is fire-and-forget, so the
+// patched `console.log` returns synchronously and the exit path below must
+// await `flushPendingWrites()` before `process.exit()`. See
 // `utils/safe-console.ts` and #2400 for the full rationale.
-import { installPipeSafeConsole } from './utils/safe-console';
+import { flushPendingWrites, installPipeSafeConsole } from './utils/safe-console';
 installPipeSafeConsole();
 
 import { parseArgs } from 'util';
@@ -1165,13 +1168,21 @@ async function main(): Promise<number> {
 // Exit explicitly so a lingering handle (DB pool, spawned child, timer) can
 // never leave the CLI hanging after its work is done.
 //
-// This is safe for piped output because every machine-readable payload is
-// emitted through `writeStdout()`/`writeJsonLine()` (src/utils/stdout.ts), which
-// resolves only once the bytes have reached the OS. The #2384 truncation
-// happened inside `console.log` at call time — not at exit — so deferring the
-// exit would not have recovered it.
+// `flushPendingWrites()` is awaited BEFORE `process.exit()` because every
+// `console.log` written through the pipe-safe shim is fire-and-forget
+// (src/utils/safe-console.ts): the stream callback that resolves the per-
+// write promise fires asynchronously, and `process.exit()` does not drain
+// `process.stdout`'s pending writes. Without this flush a very-slow reader
+// (e.g. `archon … | { sleep 1; cat; }`) would re-introduce the silent-exit-0
+// truncation the shim is meant to eliminate — see R1 in the review report.
+//
+// The `--json` paths do not need this because every JSON emitter already
+// awaits `writeStdout` / `writeJsonLine` at the call site
+// (src/utils/stdout.ts); the shim only adds the fire-and-forget shape that
+// the human-readable call surface requires.
 main()
-  .then(exitCode => {
+  .then(async exitCode => {
+    await flushPendingWrites();
     process.exit(exitCode);
   })
   .catch((error: unknown) => {
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -24,8 +24,11 @@ loadArchonEnv(process.cwd());
 // instead of dropping the tail — but delivery is fire-and-forget, so the
 // patched `console.log` returns synchronously and the exit path below must
 // await `flushPendingWrites()` before `process.exit()`. See
-// `utils/safe-console.ts` and #2400 for the full rationale.
-import { flushPendingWrites, installPipeSafeConsole } from './utils/safe-console';
+// `utils/exit-with-drain.ts` (the call site that owns the drain) and
+// `utils/safe-console.ts` for the underlying shim, and #2400 for the full
+// rationale.
+import { installPipeSafeConsole } from './utils/safe-console';
+import { exitWithDrain } from './utils/exit-with-drain';
 installPipeSafeConsole();
 
 import { parseArgs } from 'util';
@@ -1186,10 +1189,9 @@ async function main(): Promise<number> {
 // zero. Splitting the two arms' exit logic would re-open the R9 latency:
 // a fatal rejection against a slow reader would truncate and exit 1,
 // producing the same silent stdout loss the patch is meant to eliminate.
-const exitWithDrain = async (code: number): Promise<never> => {
-  await flushPendingWrites();
-  process.exit(code);
-};
+// The drain helper lives in `./utils/exit-with-drain.ts` so the
+// `safe-console.test.ts` R9 regression test imports the same code and a
+// drift between the two arms cannot land without changing that module.
 main()
   .then(exitWithDrain)
   .catch((error: unknown) => {
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -16,6 +16,15 @@ import '@archon/paths/strip-cwd-env-boot';
 import { loadArchonEnv } from '@archon/paths/env-loader';
 loadArchonEnv(process.cwd());
 
+// Install the pipe-safe `console.log` shim BEFORE any command module imports.
+// `console.log` reaches fd 1 via a non-blocking pipe (pino opens it that way at
+// module load via `@archon/paths/strip-cwd-env-boot` above), and short writes
+// are silently dropped against a slow reader. The shim delegates through
+// `writeStdout` so every byte reaches the OS before the call returns. See
+// `utils/safe-console.ts` and #2400 for the full rationale.
+import { installPipeSafeConsole } from './utils/safe-console';
+installPipeSafeConsole();
+
 import { parseArgs } from 'util';
 import { resolve } from 'path';
 import { existsSync, realpathSync } from 'fs';
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -1191,7 +1191,8 @@ async function main(): Promise<number> {
 // producing the same silent stdout loss the patch is meant to eliminate.
 // The chain wiring — `main().then(exitWithDrain).catch(...)` — lives in
 // `withDrainedExit` (`./utils/exit-with-drain.ts`) so cli.ts and the R9
-// regression test fixture share a single source of truth, and a regression
-// that bypasses `exitWithDrain` at this call site would also have to land
-// in that module.
+// regression test fixture share a single source of truth. A regression
+// that swaps this call for a direct `process.exit` is caught by the
+// static-contract test in `safe-console.test.ts`, which reads this file
+// as text.
 withDrainedExit(main);
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -101,6 +101,7 @@ import * as userDb from '@archon/core/db/users';
 import * as git from '@archon/git';
 import { CLIAdapter } from '../adapters/cli-adapter';
 import { writeJsonLine, writeStdout } from '../utils/stdout';
+import { exitWithDrain } from '../utils/exit-with-drain';
 import { resolveCliUserId } from './auth';
 
 /** Lazy-initialized logger (deferred so test mocks can intercept createLogger) */
@@ -2016,7 +2017,7 @@ async function runWorkflowWithOwnedSource(
         );
       })
       // Destroy the isolation container so Ctrl-C / SIGTERM doesn't orphan a
-      // PRIVILEGED container — `process.exit(1)` below bypasses the teardown
+      // PRIVILEGED container — the forced exit below bypasses the teardown
       // `finally`, so we must tear it down explicitly here first.
       .then(async () => {
         if (containerBackend && containerEnvId) {
@@ -2032,9 +2033,10 @@ async function runWorkflowWithOwnedSource(
         }
       })
       // Reclaim the staged capture for the same reason the container is destroyed above:
-      // `process.exit(1)` never returns up the stack, so the ownership `finally` — whose
-      // whole premise is "whichever way we leave" — never runs. Ctrl-C during isolation
-      // resolution or worktree creation would otherwise strand a complete frozen tree.
+      // the forced exit below never returns up the stack, so the ownership `finally` —
+      // whose whole premise is "whichever way we leave" — never runs. Ctrl-C during
+      // isolation resolution or worktree creation would otherwise strand a complete
+      // frozen tree.
       .then(async () => {
         if (preparedSource && !sourceAdopted) {
           await disposeWorkflowSource(preparedSource).catch(() => undefined);
@@ -2042,7 +2044,13 @@ async function runWorkflowWithOwnedSource(
       })
       .catch(() => undefined)
       .finally(() => {
-        process.exit(1);
+        // Route through the same drain helper cli.ts's top-level exit chain
+        // uses so queued `console.log` output (this command streams progress
+        // through 101 call sites) reaches a slow reader before the process
+        // exits — a bare `process.exit(1)` here would reopen #2400's
+        // truncation on Ctrl-C/SIGTERM specifically. See R16 in the review
+        // report.
+        void exitWithDrain(1);
       });
   };
   const sigtermHandler = (): void => {
--- a/packages/cli/src/utils/exit-with-drain.ts
+++ b/packages/cli/src/utils/exit-with-drain.ts
@@ -1,0 +1,26 @@
+/**
+ * Exit the CLI process after every fire-and-forget `writeStdout` write has
+ * been drained to the OS.
+ *
+ * The pipe-safe `console.log` shim in `./safe-console.ts` delegates through
+ * `writeStdout` and returns synchronously — bytes reach the OS via the
+ * underlying stream callback, not synchronously. `process.exit()` does NOT
+ * drain `process.stdout`'s pending writes, so a process that exits without
+ * awaiting the shim's `flushPendingWrites()` will silently drop the tail of
+ * queued stdout against a slow reader (the truncation the shim exists to
+ * prevent — see R1 in the review report and #2400).
+ *
+ * Both arms of `cli.ts`'s top-level `main().then().catch()` chain route
+ * through this helper so the drain contract is shared between the success
+ * and fatal paths: a regression that re-splits the two arms (e.g. dropping
+ * `flushPendingWrites()` from the catch arm only) cannot land without
+ * changing this module, and the regression test in
+ * `safe-console.test.ts` imports the same helper from here. See R9 in the
+ * review report.
+ */
+import { flushPendingWrites } from './safe-console';
+
+export const exitWithDrain = async (code: number): Promise<never> => {
+  await flushPendingWrites();
+  process.exit(code);
+};
--- a/packages/cli/src/utils/exit-with-drain.ts
+++ b/packages/cli/src/utils/exit-with-drain.ts
@@ -18,9 +18,19 @@
  * `safe-console.test.ts` imports the same helper from here. See R9 in the
  * review report.
  */
-import { flushPendingWrites } from './safe-console';
+import { consumeWriteError, flushPendingWrites } from './safe-console';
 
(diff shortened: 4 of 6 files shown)


Wirasm:
## Ready to merge

Round 6 verified the correction commit (`a6272f89`) against round 5's three blocking findings by
reproducing each failure mode directly against the current code, not by trusting the correction
report. All three are genuinely closed: **R15** (a non-`Error` `main()` rejection used to crash
`withDrainedExit`'s fatal handler before the drain ran) — reproduced the pre-fix crash on
`0950b5f0`, then confirmed the current code drains and logs cleanly for `throw null` / a string /
a plain object, with no crash. **R16** (`workflow run`'s Ctrl-C/SIGTERM handler bypassed the drain
entirely) — traced the full cleanup chain and confirmed `exitWithDrain(1)` is unconditionally
reached with no new race, and independently verified `setup.ts`'s disclosed-but-unfixed gap (46
`process.exit()` calls, 8 short `console.log` sites, no signal handlers, not a realistic piping
target) is accurately described in the narrowed PR description rather than silently glossed over —
an acceptable resolution, not a merge blocker. **R12/R13** (stale/self-contradicting comments about
which test mechanism catches a cli.ts-local exit-chain bypass) — all four locations plus the
`safe-console.ts` stale line-range pointer now read correctly and consistently across every
touched file. All 6 non-blocking suggestions the correction claims to have addressed (R17, R19,
R20, R22, R23, R24) are independently confirmed fixed, two of them (R19, R20) via mutation testing
(temporarily breaking the implementation to confirm the new tests actually fail red).

Four new Suggestions surfaced this round, all non-blocking: two are test-coverage gaps for the very
fixes that just landed (no dedicated regression test for R15's non-`Error` normalization, and no
contract test guarding `workflow.ts`'s `exitWithDrain` call site against reverting to a bare
`process.exit`), one is a cosmetic double-log-line ambiguity, and one is stale "default" wording
left over from a parameter this same commit deleted. None weaken a delivered guarantee.

**0 blocking · 9 non-blocking**

**Validation:** `bun --filter '@archon/cli' test` — 605 pass / 0 fail / 1 skip across 10 invocations
(10 files/splits). `bun test packages/cli/src/utils/safe-console.test.ts` (isolated) — 10 pass / 0
fail, including the two new R19/R20 tests. `bun --filter '*' type-check` — all 13 packages clean.
`eslint` on the 4 touched production files — 0 warnings. `prettier --check` on all 5 touched files
— clean.

**Resolved:** 10 (R12, R13, R15, R16, R17, R19, R20, R22, R23, R24)

### Findings

|---|---|---|---|
| `R15` | Important | `withDrainedExit` crashed on a non-`Error` `main()` rejection, skipping the drain | FIXED |
| `R16` | Important | `workflow run`'s Ctrl-C/SIGTERM exit bypassed the drain entirely | FIXED |
| `R12` | Important (carry) | 4 locations falsely claimed the R9 fixture catches a cli.ts-local bypass | FIXED |
| `R13` | Important (carry) | `safe-console.ts`'s pointer to `cli.ts:1171-1202` was stale again | FIXED |
| `R17` | Suggestion | Captured non-EPIPE write errors had no diagnostic trail | FIXED |
| `R19` | Suggestion | No test proved the exit-code-flip handles a non-EPIPE failure | FIXED |
| `R20` | Suggestion | Default fatal-error stderr message was never asserted by any test | FIXED |
| `R22` | Suggestion | `restoreConsole()` didn't reset `writeError`, an asymmetric-teardown trap | FIXED |
| `R23` | Suggestion | `withDrainedExit`'s `logFatal` parameter had no caller that overrode it | FIXED |
| `R24` | Suggestion | Dead `return undefined as never;` after `process.exit(1)` | FIXED |
| `R25` | Suggestion | No contract test guards `workflow.ts`'s `exitWithDrain(1)` call site (or `setup.ts`) against reverting to a bare `process.exit` | OPEN |
| `R26` | Suggestion | No test exercises `withDrainedExit`'s non-`Error` rejection normalization (R15's own fix) | OPEN |
| `R27` | Suggestion | Two independent `console.error('Fatal error: ...')` lines can co-fire for one exit, with no distinguishing prefix | OPEN |
| `R28` | Suggestion | R20 test docblock still calls the fatal handler "default" after this commit deleted the overridable parameter | OPEN |
| `R14` | Suggestion (carry) | `exitWithDrain`/`withDrainedExit` split, re-verified minimal | NOT A FINDING (unchanged) |
| `R6` | Suggestion (carry) | Pino/fd-1 misattribution in comments | OPEN (unchanged — see note) |
| `R7` | Suggestion (carry) | `depth: 4` "matches Bun's default" claim is empirically wrong | OPEN (unchanged) |
| `R8` | Suggestion (carry) | Dead rollback machinery has zero production call sites | OPEN (unchanged) |
| `R11` | Suggestion (carry) | `console.info` patch is documented but no test exercises it | OPEN (unchanged) |
| `R21` | Suggestion | Static-contract test's `process.exit` string match has a narrow bypass | OPEN (unchanged, non-blocking) |
| `R1`–`R5`, `R9`, `R10` | Important (carry) | Round 3/4 mechanism findings | FIXED (unchanged, reverified this round via full test pass) |

<details>
<summary><code>R15</code> — `withDrainedExit` crashed on a non-`Error` `main()` rejection, skipping the drain</summary>

**Impact:** `main(): Promise<number>` can reject with anything JS allows. The old default `logFatal`
cast the rejection straight to `Error` and dereferenced `.message`; a non-`Error` rejection threw a
*new*, unhandled `TypeError` inside the handler itself, so `exitWithDrain(1)` was never reached and
`flushPendingWrites()` never ran — silently dropping queued stdout on exactly the path (an
unexpected fatal error) where a user most needs to see what happened.

**Evidence:** `packages/cli/src/utils/exit-with-drain.ts:67-77` — the `.catch()` arm now reads
`error: unknown` and normalizes via `error instanceof Error ? error.message : String(error)`; the
`as Error` cast and the separate `logFatal` parameter are both gone. I reproduced the pre-fix crash
directly against `0950b5f0` (`throw null` → unhandled `TypeError: null is not an object (evaluating
'error.message')`, thrown from inside the old handler, `exitWithDrain` never reached) and then
confirmed the current code (`a6272f89`) handles all three cases cleanly:
```
throw null   → stdout drained, stderr "Fatal error: null", exit 1, no crash
throw 'str'  → stdout drained, stderr "Fatal error: a plain string rejection", exit 1, no crash
throw {obj}  → stdout drained, stderr "Fatal error: [object Object]", exit 1, no crash
```
`String()` (function form) is safe for every JS-throwable, including symbols, where template-literal
coercion would itself throw.

**Required outcome:** Met. The correction went slightly further than the original ask (type-widen
`logFatal`) by deleting the parameter entirely, since it had zero callers passing a second argument
anywhere in the codebase or tests (confirmed by a repo-wide `logFatal` grep returning zero hits) —
this also closes R23 as a side effect and is a proportionate simplification, not scope creep.

**Found by:** `prp-core:type-design-analyzer` (original + reverification), `silent-failure-hunter`,
`prp-core:code-reviewer` (independently convergent)

**Disposition:** FIXED. Reproduced live against both the pre-fix and post-fix code.

</details>

<details>
<summary><code>R16</code> — `workflow run`'s Ctrl-C/SIGTERM exit bypassed the drain entirely</summary>

**Impact:** The drain/write-error guarantee only covered `cli.ts`'s single `main()` exit chain.
`workflow run`'s signal-cleanup handler called a bare `process.exit(1)`, which never drained
`pendingWrites` — a real truncation risk given `workflow.ts` streams progress through 101
`console.log`/`console.info` call sites, one of the CLI's heaviest emitters.

**Evidence:** `packages/cli/src/commands/workflow.ts:2044-2054` now ends the cleanup chain with
`void exitWithDrain(1);` instead of `process.exit(1);`, imported at `workflow.ts:104`. I traced the
full chain: the preceding `.catch(() => undefined)` absorbs any error from the container-destroy /
source-dispose `.then()` stages, so `.finally()` — and therefore `exitWithDrain(1)` — is
unconditionally reached. `grep -n "process\.exit(" packages/cli/src/commands/workflow.ts` returns
only a comment; zero direct exits remain in the file. `flushPendingWrites()`'s `pendingWrites` Set
tracks every write from issue to settlement regardless of how much async cleanup work runs in
between, so no write queued before the signal can be dropped — no race.

`setup.ts` (~46 direct `process.exit()` calls) was deliberately left unfixed. I independently
verified this narrowing is accurate rather than convenient: `setup.ts` has exactly 8
`console.log`/`console.info` sites (matches the PR's claim), all short one-line strings totaling
well under the 64 KiB pipe-buffer truncation threshold, zero signal handlers of its own, and is an
interactive setup wizard rather than a realistic piping target (`archon setup | less` is not the
`archon workflow list | less` scenario #2400 describes). The PR description's Outcome/Scope-boundary
sections now state this gap plainly rather than glossing over it.

**Required outcome:** Met for `workflow run`; `setup.ts` remains a disclosed, verified-low-risk,
non-blocking gap.

**Found by:** `silent-failure-hunter` (original + reverification)

**Disposition:** FIXED. `setup.ts`'s narrower scope is judged acceptable — not a blocker.

</details>

<details>
<summary><code>R12</code> — 4 locations falsely claimed the R9 fixture catches a cli.ts-local bypass</summary>

**Impact:** A maintainer reading any of the four inaccurate locations could conclude the R9 pipe-test
fixture already covers a cli.ts-local `process.exit` swap and remove the static-contract test as
apparently redundant, reopening the exact regression class round 4 proved was open.

**Evidence:** All four locations now correctly split the claim — "regression inside
`withDrainedExit`/`exitWithDrain` itself" (caught by the R9 fixture, which imports and calls the real
function) vs. "bypasses `withDrainedExit` at the cli.ts call site" (NOT caught by the fixture; caught
by the separate static-contract test):
- `packages/cli/src/utils/exit-with-drain.ts:13-21`
- `packages/cli/src/cli.ts:1192-1197`
- `packages/cli/src/utils/safe-console.test.ts:220-228`
- `packages/cli/src/utils/safe-console.test.ts:232-239`

A repo-wide sweep for the old false-claim phrasing found only two remaining "caught by the fixture"
mentions (`exit-with-drain.ts:17`, `:64`), both legitimately scoped to internal regressions the R9
fixture genuinely does catch. The static-contract test itself (`safe-console.test.ts:320-341`) is
unchanged and independently confirmed to do what the comments now claim.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:comment-analyzer`
(independently convergent, all three)

**Disposition:** FIXED.

</details>

<details>
<summary><code>R13</code> — `safe-console.ts`'s pointer to `cli.ts:1171-1202` was stale again</summary>

**Impact:** The `installPipeSafeConsole` JSDoc cited a line range past the file's actual end and
described a `.then().catch()` chain that no longer existed in `cli.ts` — same defect class as the
original R13, recurring because subsequent commits edited the described code without revisiting the
comment.

**Evidence:** `packages/cli/src/utils/safe-console.ts:138-141` now reads "...await
`flushPendingWrites()` — see `utils/exit-with-drain.ts` (the `withDrainedExit` helper that owns
this) and the `withDrainedExit(main)` call at the end of `cli.ts`." No numeric line range remains.
Verified accurate: `tail -5 packages/cli/src/cli.ts` shows the file's actual last statement is
literally `withDrainedExit(main);`.

**Required outcome:** Met — matches this repo's stable-text-anchor convention exactly.

**Found by:** `prp-core:comment-analyzer` (original + reverification)

**Disposition:** FIXED.

</details>

<details>
<summary><code>R17</code>/<code>R19</code>/<code>R20</code>/<code>R22</code>/<code>R23</code>/<code>R24</code> — non-blocking suggestions, all confirmed fixed</summary>

- **R17** (captured write errors had no diagnostic trail): `exit-with-drain.ts:37-39` now logs
  `console.error('Fatal error: failed to write output:', writeError.message)` for any non-EPIPE
  write failure, before the forced exit. Genuinely exercised by the new R19 test.
- **R19** (no test for a non-EPIPE write failure): the new test
  `exitWithDrain(0) still exits 1 and logs a non-EPIPE write failure`
  (`safe-console.test.ts:541-604`) drives the real `exitWithDrain` with a real `ENOSPC`-coded write
  failure. Mutation-tested: gating the forced exit on `code === 0 && writeError.code === 'EPIPE'`
  made this test fail red (`Expected: 1, Received: 0`); restoring the correct code makes it pass.
- **R20** (fatal-error message never asserted): the new test
  `withDrainedExit logs the fatal rejection message to stderr (R20)`
  (`safe-console.test.ts:288-306`) captures real stderr (not `/dev/null`) via a subprocess fixture
  and asserts the message content. Mutation-tested: removing the `console.error` call made this test
  fail red; restoring it passes.
- **R22** (`restoreConsole()` didn't reset `writeError`): `safe-console.ts:194-197` now adds
  `writeError = null;` immediately after `installed = false;`, symmetric with `consumeWriteError()`.
- **R23** (`logFatal` parameter unused): removed entirely as part of the R15 fix; grep confirms zero
  callers ever passed a second argument.
- **R24** (dead `return undefined as never;`): the old
  `if (consumeWriteError() !== null && code === 0) { process.exit(1); return undefined as never; }`
  block was restructured into the new EPIPE/non-EPIPE branching in `exit-with-drain.ts:27-49`, which
  no longer contains the dead `return` statement — confirmed by direct reading of the current file.

**Found by:** `silent-failure-hunter`, `prp-core:pr-test-analyzer`, `prp-core:seam-analyzer`,
`prp-core:code-reviewer` (as applicable per finding)

**Disposition:** All FIXED.

</details>

<details>
<summary><code>R25</code> — No contract test guards `workflow.ts`'s `exitWithDrain(1)` call site against reverting to a bare `process.exit`</summary>

**Impact:** `cli.ts`'s exit chain is protected by a static-contract test that reads the file as text
and fails on a bare `process.exit`. No equivalent exists for `workflow.ts`'s signal-cleanup call
site. `workflow.test.ts`'s existing signal-cleanup suite mocks `console.log` to a no-op, so
`pendingWrites` never has anything real to drain — its assertions (`exitSpy` called with `1`) pass
identically whether the `.finally()` calls `exitWithDrain(1)` or a bare `process.exit(1)`. A future
revert of this round's own R16 fix would go undetected by every current test. `setup.ts` already has
the same unenforced gap for its 46 direct exits (out of scope for this PR, disclosed).

**Evidence:** `packages/cli/src/commands/workflow.ts:2053`; `packages/cli/src/commands/workflow.test.ts:7248`
(`consoleSpy = spyOn(console, 'log').mockImplementation(() => {})`); the three signal-cleanup
assertions at `workflow.test.ts:7326,7358,7381`.

**Required outcome:** Extend the static-contract pattern (or a shared grep-based inventory test) to
cover `workflow.ts`'s call site, and eventually `setup.ts`. Reasonable as follow-up work, not a
blocker — the underlying `exitWithDrain`/`flushPendingWrites` primitive is independently proven
correct by the real-pipe tests in `safe-console.test.ts`; what's unguarded is only the *wiring*.

**Found by:** `prp-core:seam-analyzer`, `prp-core:pr-test-analyzer` (independently convergent)

**Disposition:** OPEN (non-blocking).

</details>

<details>
<summary><code>R26</code> — No test exercises `withDrainedExit`'s non-`Error` rejection normalization</summary>

**Impact:** R15's fix (`error instanceof Error ? error.message : String(error)`) has no dedicated
regression test — the new R20 test only throws a real `Error`. A future edit that reverts to an
unguarded `(error as Error).message` would compile cleanly (the `as` assertion type-checks against
`error: unknown`) and reintroduce R15's exact crash-and-skip-the-drain bug for a non-`Error`
rejection, undetected by any current test.

**Evidence:** `packages/cli/src/utils/exit-with-drain.ts:73`; `packages/cli/src/utils/safe-console.test.ts:288-306`
(only `throw new Error(...)`); grep for `throw null`, `throw '`, `throw {` in the test file returns
zero matches.

**Required outcome:** Add one subprocess fixture (same pattern as R20) with
`throw null` (or a string/plain object) and assert a clean exit 1 with a coherent `Fatal error:`
line, rather than a hang or crash. Not a live defect today — `String()` is already safe — this is a
regression-prevention test for the fix that just landed.

**Found by:** `prp-core:pr-test-analyzer`, `prp-core:type-design-analyzer` (independently convergent)

**Disposition:** OPEN (non-blocking).

</details>

<details>
<summary><code>R27</code> — Two independent "Fatal error:" log lines can co-fire for one exit</summary>

**Impact:** If `main()` rejects AND a non-EPIPE write failure was captured earlier in the same run
(plausibly the same root cause, e.g. disk full), `withDrainedExit`'s catch arm logs
`'Fatal error: <message>'` and then `exitWithDrain(1)` logs a second, differently-worded
`'Fatal error: failed to write output: <message>'` — both true, but a human or log-scraper expecting
one canonical line per crash sees two with no indication they're related. Exit code stays correctly
`1` in every case; no data loss, no wrong signal.

**Evidence:** `packages/cli/src/utils/exit-with-drain.ts:38` and `:74`.

**Required outcome:** Give the write-failure branch a distinct prefix (e.g. `'Write error:'`) so the
two causes are visually and grep-distinguishable when they co-occur. One-line change, no test
impact, purely cosmetic.

**Disposition:** OPEN (non-blocking).

</details>

<details>
<summary><code>R28</code> — R20 test docblock still calls the fatal handler "default" after the overridable parameter was deleted</summary>

**Impact:** The R20 test's docblock says "`withDrainedExit`'s default fatal-error handler prints..."
— stale terminology from before this same commit deleted the `logFatal` parameter that made
"default" a meaningful word (default *among alternatives* a caller could supply). There is no longer
a customization point. A reader might look for a way to override the fatal-error logging and not
realize it was removed.

**Evidence:** `packages/cli/src/utils/safe-console.test.ts:280-286`, contradicted by
`packages/cli/src/utils/exit-with-drain.ts:67-76` (no second parameter).

**Required outcome:** Drop "default" — e.g. "`withDrainedExit`'s fatal-error handler prints...".

**Disposition:** OPEN (non-blocking).

</details>

<details>
<summary>Carried findings unchanged this round — R1–R11, R14, R21</summary>

R1, R2, R3, R4, R5, R9, R10 (all Important) were reverified via the full `bun --filter '@archon/cli'
test` pass (605/0/1, matching the PR's own claim) and direct reading of the touched files — no
regression from this correction commit. R14 (Suggestion, `exitWithDrain`/`withDrainedExit` split)
remains NOT A FINDING — the split is still the minimal decomposition satisfying both "one exit-chain
shape" and "unit-testable in isolation." R7, R8, R11 (all Suggestion) are unchanged, not touched by
this diff. R21 (Suggestion, non-blocking) — the static-contract test's `process.exit` string-match
bypass via `import { exit } from 'node:process'` — is unchanged; this commit did not touch that test.

**R6** (Suggestion, carry — pino/fd-1 misattribution in comments) deserves a specific note: per the
operator's brief, the correction pass ran a live empirical check on this specific claim rather than
guessing, and reports the result didn't cleanly support either the current comment's attribution or
the counter-claim — so it was deliberately left un-guess-fixed and flagged as needing dedicated
investigation. This round did not re-dispatch a scope to independently re-investigate R6 (it owns no
finding this round's correction touched), so its disposition is carried forward unchanged as OPEN,
non-blocking, with this context preserved for whoever picks it up next.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | R12, R23 — confirmed fixed; independently assessed workflow.ts/error-normalization/comment-anchor, no new findings |
| seams | R12, R22 — confirmed fixed; R25 (new), R27 (new) |
| types | R15 — confirmed fixed via live reproduction; R26 noted (test-coverage, attributed to tests scope) |
| errors | R16, R17 — confirmed fixed; independently verified setup.ts scope-narrowing is accurate and non-blocking |
| comments | R12, R13 — confirmed fixed; R28 (new) |
| tests | R19, R20, R21 — R19/R20 confirmed fixed via mutation testing, R21 unchanged; R25 (new, converges with seams), R26 (new) |
| simplify | R23, R24 (both simplify-owned carried Suggestions) verified FIXED via direct coordinator diff inspection, cross-corroborated by code-reviewer and seam-analyzer's independent readings of the same file — not a dedicated agent dispatch this round, since both fixes are mechanical deletions with no separate simplification judgment required |

#### Validation

|---|---|---|
| `bun --filter '@archon/cli' test` | PASS | 605 pass / 0 fail / 1 skip across 10 invocations |
| `bun test packages/cli/src/utils/safe-console.test.ts` (isolated) | PASS | 10 pass / 0 fail, includes new R19/R20 tests |
| `bun test packages/cli/src/commands/workflow.test.ts` (isolated) | PASS | 256 pass / 0 fail, includes signal-cleanup guard suite |
| `bun --filter '*' type-check` | PASS | All 13 packages clean |
| `bunx eslint` on the 4 touched production files (`cli.ts`, `safe-console.ts`, `exit-with-drain.ts`, `workflow.ts`) | PASS | 0 warnings |
| `bunx prettier --check` (5 touched files) | PASS | All match Prettier style |
| Live reproduction: `withDrainedExit` with `throw null` / string / object, pre- and post-fix | PASS | Pre-fix crashes (unhandled TypeError, drain skipped); post-fix drains, logs, exits 1 cleanly for all three |
| `bun run lint` (full monorepo) | NOT RUN | Change confined to 5 files in `@archon/cli`; touched-file eslint pass plus full type-check judged sufficient, consistent with round 5's judgment |
| `bun run check:schema-upgrades` | NOT RUN | No schema changes in this PR |

</details>

Issue #2402, opened 2026-08-03
fix(workflows): resume skips a downstream node whose dependencies re-ran, so it never sees their new output


Wirasm:
Observed live on 2026-08-03 while debugging run `7f392748`, then confirmed against the code.

## What happened

A run failed partway, was resumed, and the resume produced **new upstream output that its downstream consumer never saw** — while the run reported success.

From `remote_agent_workflow_events`:

```
08:49:09  code-review    node_started
08:51:51  code-review    node_completed
08:51:51  synthesize     node_started        <- ran with 1 of 2 intended reviews
08:52:53  synthesize     node_completed
          --- run failed, later resumed ---
10:09:32  test-coverage  node_started        <- re-ran on the resume
10:11:10  test-coverage  node_completed      <- wrote test-coverage-findings.md (6.5 KB)
10:11:10  synthesize     node_skipped_prior_success
```

`synthesize` depends on `test-coverage`. On the first attempt an upstream gate failure caused `test-coverage` to be skipped, so `synthesize` completed against a single review. On the resume `test-coverage` executed for real and wrote its findings — and `synthesize`, already `completed` from attempt 1, was skipped. Its stale synthesis was then repeated verbatim in the run's completion report, which described the gap as "4 specialist artifacts unavailable" (untrue: the artifact existed on disk).

## Verified cause

`packages/workflows/src/dag-executor.ts:5655`:

```ts
// 0. Skip if this node completed successfully in a prior run (resume path).
if (priorCompletedNodes?.has(node.id)) {
  if (node.always_run) { /* re-execute */ }
  else { /* emit node_skipped_prior_success, skip */ }
}
```

The condition is **membership alone**. Nothing consults whether the node's dependencies re-ran, produced different output, or changed state since the prior attempt. A node that succeeded on degraded inputs is treated as done forever.

## Existing mitigation, and why it is not sufficient

`always_run: true` opts a node out of resume caching — the escape hatch exists and works.

But it requires the author to know *in advance* which nodes are input-sensitive, and it is blunt: the node then re-runs on **every** resume, including the common case where nothing upstream changed. For an expensive AI synthesis node that is a real cost. The default is silently wrong for any fan-in node, and the fix costs a re-run every time.

## Why it matters beyond one run

[...]

Issue #2393, opened 2026-08-03
design(workflows): finish the workflow event contract for outcomes and observability


Wirasm:
> **Brought current 2026-09-10 (additive).** The write-site half of the settled decision has landed: #2698 (PR #3077) gave terminal failure one finalizer, #3043 (PR #3254) renders a terminal record at every terminal transition, #3255 (PR #3275) routes every node-state fact through one function so the row, the JSONL transcript, and the emitter derive from one value, and PR #3292 gives a wait's completion rows one owner on both paths. The graph track is complete (#2933, #2302, #2487 closed) and #2407 and #2327 are closed. What remains here is the typed record itself: `workflow_events.data` is still `Record<string, unknown>` and the derived sinks read it by hand, so the next step is #3295, which owns the record shape and the implementation, after #2334's reporting capability and, in sequence, #2398's renames under the resume-key rule; #2614's complete-transcript acceptance then holds by construction. Hard blocker on GitHub: #2334, whose Codex reporting truth the record's `costSource` depends on. This issue is not a prerequisite of the pack rewrite #3127, which consumes #2453's typed outputs and #3043's terminal record.

## Current status

This started as a broad audit of `workflow_events`. The audit was useful, but its original body now describes several gaps that have already shipped.

Delivered:

- #2395 / PR #2415: resolved effort reaches the event stream.
- #2396 / PR #2421: `tool_called` and `tool_completed` carry correlation ids.
- #2397 / PR #2428: `workflow_started` carries a run-configuration snapshot.
- PR #2347: per-node token usage is persisted across AI, loop, and loop-group completion paths.
- #2342 / PR #2378: Codex tool duration no longer reports a confident fake zero.
- #2698 / PR #3077: `executeNodeInternal` ends a failing node through one finalizer.
- #3043 / PR #3254: a terminal record is assembled at every terminal transition and exposed by `workflow get` and the API.
- #3255 / PR #3275: every node-state write in `dag-executor.ts` goes through `recordNodeState`; the eleven writes that left no transcript row now do, and `trigger_rule` skips keep their `cause`.
- PR #3292: the wait outcome row and the wait node's completed row have one owner, `waitCompletionEvents`, on the same-tick and resumed paths.

The remaining work is not “redesign every event.” It is to finish a small stable contract for consumers and keep unresolved reporting limitations explicit.

## Settled decisions

- The event stream is the shared debugging, observability, and evaluation substrate. Do not add a second parallel record that must stay synchronized.
[...]

PR #2705, opened 2026-08-22
fix(workflows): invalidate cached prior-success nodes when a dependency re-ran

Wirasm:
## Problem and outcome

A resumed workflow run that re-executes an upstream producer (e.g. `test-coverage`) silently keeps a downstream consumer's (`synthesize`) prior-success entry in the resume cache, even though the consumer was computed against the OLD producer output. From run `7f392748` in [#2402](https://github.com/coleam00/Archon/issues/2402): the resume executed `test-coverage`, then `node_skipped_prior_success`'d `synthesize`, and the final completion report quoted synthesis that was missing a review the artifact on disk now contained.

- **Outcome:** when a downstream node is in `priorCompletedNodes` and any of its dependencies re-ran during the current resume (text differs, structured value differs, or the dep was not in the prior snapshot and is now complete or failed), the cached entry is invalidated and the node re-executes against the fresh dependency values. A new `node_prior_cache_invalidated` audit event records the invalidating deps and the discarded cached output so the timeline shows exactly why the cache was not honoured.
- **Invariant:** the existing genuine-cache skip (deps unchanged) is unchanged — `node_skipped_prior_success` semantics and the `always_run: true` opt-out both still work. The fix is local to the resume-skip branch in `dag-executor.ts`, plus one supporting fix in `getDagResumeSnapshot` (see "Review round 2" under Validation) — this PR's own invalidate-and-re-execute mechanism is the first thing that can produce a `node_completed → node_failed` sequence for the same step, and that sequence needed to be handled at the snapshot layer too.
- **Scope boundary:** the in-process "ran this resume" detection is derived from a comparison of `priorCompletedNodes` and `nodeOutputs` (which is pre-populated from prior snapshot, then overwritten by per-layer aggregation only). No new persistent state, no new column, no schema migration. The `node_skipped_prior_success` event continues to be the audit signal for the genuine cache case; the new event is the audit signal for the invalidation case only.
- **Root cause:** the resume-skip branch at `packages/workflows/src/dag-executor.ts` keyed solely on `priorCompletedNodes?.has(node.id)` and ignored whether the node's dependencies had re-executed since the cached completion. There is no `node_started`-in-this-resume or completion-generation counter anywhere in the engine, so a diff between the dep's prior snapshot and its current `nodeOutputs` value is the only signal the engine already has without a new persistent state.

## Review guidance

- **Feedback requested:** correctness of the cache-invalidation contract, the audit-event shape (`prior_output` / `prior_structured_output` / `invalidating_deps`), and the conservative comparison (byte-identical re-run is intentionally not flagged as stale). Reviewers should form their own view on whether `Relates to #2402` is the right linkage given the issue comment's reframing (see Links).
- **Start here:** `packages/workflows/src/dag-executor.ts` — `getStaleCachedDependencies` (the new helper) and the modified branch in `runLayers` around the existing `node_skipped_prior_success` arm. The audit-event addition at `packages/workflows/src/store.ts` is a one-line addition to a closed `as const` tuple. Review round 2 also touches `packages/core/src/db/workflow-events.ts` — `getDagResumeSnapshot`'s `node_failed` handling (R2, see Validation).
- **Review order:**
  1. `getStaleCachedDependencies`: the unconditional failed-state check, then the three value-based cases it surfaces (never-prior / text-differs / structured-differs).
  2. The branch in `runLayers`: when `staleDeps.length > 0`, fall through to fresh execution (mirrors the `always_run: true` arm) instead of the skip arm.
  3. The audit event shape and its presence in `WORKFLOW_EVENT_TYPES`.
  4. `getDagResumeSnapshot`'s `node_failed` handling in `packages/core/src/db/workflow-events.ts`.
  5. The seven tests in the `#2402` describe block in `packages/workflows/src/dag-executor.test.ts`, plus the two new tests in `packages/core/src/db/workflow-events.test.ts`.
- **Lower-attention areas:** the existing `node_skipped_prior_success` payload and emit path are unchanged; the `always_run: true` arm upstream of the new branch is unchanged. The new event is filtered out of usage aggregation by virtue of carrying no `node_output` or `cost_usd`.
- **Known risk or uncertainty:** the comparison is intentionally lenient in the "no change" direction — a dep that re-runs with byte-identical text AND structured output is NOT flagged stale. The cached downstream is semantically equivalent in that case (over-run is the safe-but-costly direction to err in, not the cheap-but-wrong one). This is pinned by one of the new tests. If a stricter generation-counter contract is wanted later, that is a separate change.

## Solution

A dep-re-ran check sits between the existing `composedBoundaryDecision === 'run'` branch and the prior-success emit. Before honouring the cache, the engine asks: did any of this node's dependencies produce different output during the current resume? It can answer that without new persistence because `nodeOutputs` is pre-populated from `priorCompletedNodes` and then overwritten by the per-layer aggregation — a diff is proof of fresh execution.

Two helpers carry the comparison:

- `getStaleCachedDependencies` walks `node.depends_on` and returns the subset of dep ids in three states: (1) the dep was never in `priorCompletedNodes` and is now `completed` or `failed`; (2) the dep's text output differs from the prior snapshot; (3) the dep's structured output differs even though text did not (the surface introduced by #2637). Case 3 matters because `$dep.output.field` consumers would otherwise silently read stale structured data through a text-stable cached downstream. `JSON.stringify` is sufficient because structured outputs are persisted to JSONB.
- `structuredOutputsEqual` is the structured comparator — strict-equal for the trivial case, JSON-stringified equal otherwise, with `undefined` normalised to absent.

When any dep is stale, the cached node is invalidated: a `node_prior_cache_invalidated` event is persisted (carrying `invalidating_deps`, `prior_output`, and `prior_structured_output` when set), the log branch writes `dag.node_prior_cache_invalidated`, and execution falls through to a fresh re-execution against the new dep values. When no dep is stale, the existing `node_skipped_prior_success` arm runs untouched.

This is the smallest coherent fix because the engine already carries everything the comparison needs. The main alternative — a per-node completion-generation counter or a per-resume "executed this run" `Set` — would be exactly as ephemeral as `ctx.nodeOutputs`, so it would not actually require new persistence; its real cost is a larger blast radius, since the per-layer aggregation path is shared by every node type's many dispatch return points (bash/script/prompt/command/loop/approval, etc.), and threading a clean "did this node actually dispatch fresh work" signal through all of them is a materially bigger change than the value-diffing approach shipped here — which also still needs the conservative "byte-identical re-run" handling either way.

## Behavior change

| | Before | After |
|---|---|---|
| **Resume with re-ran upstream and cached downstream** | Downstream is `node_skipped_prior_success`'d; the prior (stale) output is repeated verbatim in the run's completion report, which can describe gaps that no longer exist. | Downstream is invalidated: `node_prior_cache_invalidated` is emitted with `invalidating_deps` and `prior_output`, then the downstream re-executes against the fresh dep values. The completion report reflects the new state. |
| **Resume with unchanged deps and cached downstream** | Cached value is reused as before. | Unchanged: the genuine-cache path continues to emit `node_skipped_prior_success` and skip the node. The new branch is consulted only and falls through to the unchanged skip arm. |
| **`always_run: true` nodes** | `node_always_run_reset` is emitted and the node re-runs; downstream consumers are still membership-skipped. | Unchanged for the upstream itself; downstream consumers are now also subject to the dep-re-ran check above, so a consumer over an `always_run: true` producer re-runs against the freshly-reset producer on every resume (the desired behaviour). |

### User flow

```mermaid
flowchart LR
  subgraph Before
    P1[Producer re-runs on resume] --> S1[synthesize skipped via prior_success]
    S1 --> R1[Run reports success over stale synthesis]
  end

  subgraph After
    P2[Producer re-runs on resume] --> C2{Any dep stale vs. prior snapshot?}
    C2 -- Yes --> I2[node_prior_cache_invalidated]
    I2 --> E2[synthesize re-executes against fresh producer]
    E2 --> R2[Run reports correct synthesis]
    C2 -- No --> S2[synthesize skipped via prior_success unchanged]
  end
```

## Architecture

```mermaid
flowchart LR
  A[runLayers resume-skip branch] --> B{priorCompletedNodes.has<br/>and composedBoundaryDecision run}
  B --> C[getStaleCachedDependencies]
  C --> D{any dep differs from<br/>priorCompletedNodes snapshot?}
  D -- Yes --> E[persist node_prior_cache_invalidated<br/>+ fall through to fresh execute]
  D -- No --> F[emit node_skipped_prior_success<br/>and reuse cache unchanged]
  E --> G[executeNodeInternal / executeBashNode / etc.]
  F --> H[Continued layered walk]
  G --> H
```

The `priorCompletedNodes` snapshot read at `packages/core/src/db/workflow-events.ts` gained one targeted fix (review round 2, R2 — see Validation); the engine does not gain a "ran in this resume" persistent set. The new event matches `node_skipped_prior_success`'s shape on the discard side (`prior_output` / `prior_structured_output`) so a future #2393 audit that pairs cache events with their invalidations can render them adjacently without a per-event special case.

### Changed seams

|---|---|---|
| Resume-skip membership check (consumer-of-prior-output) | New dep-staleness check before the skip arm; invalidates and re-executes when any dep changed. | `packages/workflows/src/dag-executor.ts` `getStaleCachedDependencies` + `runLayers`; `executeDagWorkflow -- prior-success cache invalidated by dep re-execution (#2402)` describe block in `packages/workflows/src/dag-executor.test.ts`. |
| `WORKFLOW_EVENT_TYPES` | Added `node_prior_cache_invalidated`. | `packages/workflows/src/store.ts`; closed `as const` tuple so `createWorkflowEvent` and the CLI's event-type validator cover it. |
| `priorCompletedNodes` snapshot read | `getDagResumeSnapshot` now deletes a step's `completedNodeOutputs` entry when a later `node_failed` row lands for that same step, so a later failure supersedes an earlier completion instead of being silently outvoted by it (review round 2, R2). | `packages/core/src/db/workflow-events.ts`. |
| `node_skipped_prior_success` semantics for the genuine cache case | Unchanged. | Same emit path; payload unchanged. |
| `always_run: true` arm | Unchanged. | Still emits `node_always_run_reset` upstream of the new check. |
| `node_output` / usage aggregation | Unchanged. | The new event carries no `node_output` or `cost_usd`, so it filters out of usage totals by construction. |
| UI event mapping (`packages/web/src/experiments/console/primitives/event.ts`) | Unchanged. | The new event is an audit signal, not a UI transition; `node_started` / `node_completed` / `node_failed` follow when the node actually re-executes. |

## Validation

- `bun test packages/workflows/src/dag-executor.test.ts` — **590 pass, 0 fail** (583 pre-existing + 7 in the `#2402` describe block, one added in review round 2). The describe block `executeDagWorkflow -- prior-success cache invalidated by dep re-execution (#2402)`:
  - `invalidates a cached downstream when an always_run dep re-runs with fresh output` — bug-shape reproducer; verifies the consumer sees the fresh producer output, that no `node_skipped_prior_success` is emitted for the consumer, and that `node_prior_cache_invalidated` carries `invalidating_deps: ['producer']` and the discarded `prior_output`.
  - `does NOT invalidate when a cached dep actually re-runs and produces text-identical output` — pins the conservative comparison contract with a dep that genuinely re-executes (`always_run: true`) and lands identical output; the consumer stays cached and no invalidation event fires. (Rewritten in review round 2 — R3: the original version's producer had no `depends_on` and was itself skipped via the ordinary prior-cache path, so the comparison it claimed to pin was never reached.)
  - `invalidates a cached downstream when a dep that was never cached completes fresh this resume` — covers case 1 of `getStaleCachedDependencies`.
  - `invalidates a cached downstream when a dep that re-ran fresh failed (case 1 "failed" arm)` — the dep had NO prior cache entry and fails fresh this resume.
  - `invalidates a cached downstream when a PREVIOUSLY-CACHED dep fails fresh with output matching its prior success (#2705 R1)` — new in review round 2. The dep WAS cached from a prior success whose output was `''`, and fails fresh this resume with the same `''` output (the shape every failure arm in the executor produces) — pins that a fresh failure is never masked by value equality against a prior success.
  - `invalidates when a dep's structured output changes even though text is stable` — covers case 3; pins that `prior_structured_output` is populated.
  - `invalidates a cached downstream when ONE of its multiple deps re-ran (any-stale triggers re-execution)` — pins the "any stale dep invalidates" trigger for cached consumers with ≥ 2 deps; verifies `invalidating_deps` lists only the actually-stale deps (length 1, not 2) and that the stable sibling still emits `node_skipped_prior_success`.
- `bun test packages/core/src/db/workflow-events.test.ts` — **34 pass, 0 fail**, including two new tests added in review round 2 (R2): a `node_completed → node_failed` sequence for the same step no longer leaves the stale completion in `completedNodeOutputs`, and the symmetric `node_failed → node_completed` sequence still honours the later success (no permanent tombstone).
- `bun run --filter @archon/workflows test` (full package suite) — exit code 0.
- `bun run --filter @archon/core test` (full package suite) — exit code 0.
- `bun run type-check` — all packages pass.
- `bun run lint` — clean.
- `bun run format:check` — clean.

Pre-existing cross-file mock-pollution issues (e.g. `subrun.test.ts` together with `dag-executor.test.ts` produces failures that also reproduce on the unmodified `dev`) are not caused by this change; `CONTRIBUTING.md` already documents `bun run test` for isolation, and each affected file passes 100% when run on its own.

- **Not verified:** No live reproduction of run `7f392748` end-to-end; the fix is proven against current code via fixtures in the same describe block as the existing `always_run` opt-out tests, but is not regressed through a real LangSmith / production trace. Nothing material in the closure of the dep-re-ran surface requires that.

### Review round 2 (2026-08-22)

[prp-review](https://github.com/coleam00/Archon/pull/2705#issuecomment-5379798172) returned NEEDS FIXES against `ba1de703`, with two blocking findings (five independent reviewers converged on R1; two independently reproduced both R1 and R2 against the live head):

- **R1 (blocking, fixed):** `getStaleCachedDependencies` only re-checked a dependency's terminal *state* when that dependency had no prior cache entry at all. Once a dependency WAS cached — the ordinary shape for an `always_run: true` gate on a second-or-later resume — staleness was judged purely by output-value equality, and every failure arm in the executor returns `output: ''`. A dependency that succeeded with `''` and then genuinely failed on this resume with `''` again looked unchanged, so its cached downstream was skipped via `node_skipped_prior_success` before `checkTriggerRule` ever ran. Fixed by hoisting a `current?.state === 'failed'` check to run unconditionally, before the prior-cache lookup, in `getStaleCachedDependencies` (`packages/workflows/src/dag-executor.ts`).
- **R2 (blocking, fixed):** `getDagResumeSnapshot` (outside this PR's original diff, but newly reachable because of it) never cleared an earlier `node_completed` map entry when a later `node_failed` row landed for the same step. Before this PR, a non-`always_run` node that had already completed could never be re-attempted, so that sequence was impossible; this PR's own invalidate-and-re-execute mechanism is the first thing that can produce it. Fixed by deleting the step's `completedNodeOutputs` entry when a `node_failed` row is processed (`packages/core/src/db/workflow-events.ts`).
- **R3 (non-blocking, fixed):** the "does NOT invalidate... identical output" test's producer never actually re-executed, so the conservative contract it claimed to pin was never reached. Rewritten so the producer is `always_run: true` and genuinely re-runs with identical output.
- **R4 (non-blocking, fixed):** corrected the Solution section's inaccurate claim that the rejected "executed this run" `Set` alternative "would require new persistence" — it would be exactly as ephemeral as `ctx.nodeOutputs`; the real cost is the blast radius of threading a new signal through every node type's dispatch return points.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. No new persisted state, no migration, no event-name schema change for events other than the additive `node_prior_cache_invalidated`. | `packages/workflows/src/store.ts` is an additive change to a closed tuple. |
| Security / permissions / data | None. The change reads only what `priorCompletedNodes` and `nodeOutputs` already carry; no new process boundary, no new I/O. | Helper is a pure local function over the existing context maps. |
| Rollout / rollback | Reverse-merge is safe: removing the helper call restores the prior behaviour exactly. The new event type would then stop being persisted; downstream consumers that filter by `node_skipped_prior_success` are unaffected. | One branch in `runLayers`; one helper. |
| Observability | A new audit event (`node_prior_cache_invalidated`) carries `invalidating_deps`, `prior_output`, and `prior_structured_output`. When #2393 (workflow_events schema audit) lands, this event is the concrete "show the discarded cache next to the fresh completion" hook. Filtered out of usage totals by carrying no `node_output` or `cost_usd`. | `packages/workflows/src/store.ts`; `packages/workflows/src/dag-executor.ts` invalidation branch. |
| Documentation / communication | None required for this PR. #2327 remains the upstream root for the case where `priorCompletedNodes` itself contains a false `node_completed`; #2393 remains the substrate to measure cache-invalidation frequency. Both are independent of this change. | Per the issue comment and the triage notes in the run record. |

## Links

- Related #2402
- Related #2327 — a node that reports itself blocked is recorded `node_completed`. Per the issue comment, #2402 is downstream of #2327: this PR closes the dep-re-ran amplifier surface in #2402 (a producer that genuinely re-runs with new output invalidates the consumer's cache). It does not address the #2327 case where a producer was falsely recorded `node_completed` in a prior attempt and therefore never re-runs. That remains a separate decision.
- Related #2393 — workflow_events schema audit. The new `node_prior_cache_invalidated` event is sized to pair with `node_skipped_prior_success` so the audit can render cache invalidations next to cache honours.
- Related #2637 — value transport. The case-3 detection in `getStaleCachedDependencies` exists because `$dep.output.field` consumers would otherwise silently read stale structured data through a text-stable cache.

## Summary by CodeRabbit

- **New Features**
  - Improved workflow resume behavior by re-running previously successful steps when dependencies or outputs have changed.
  - Added tracking for invalidated cached steps, including affected dependencies and prior output details.
  - Preserved structured output when safely skipping cached steps.

- **Bug Fixes**
  - Prevented stale cached results from being reused after dependency updates or failures.
  - Ensured downstream prompts and results are refreshed when upstream data changes.
  - Corrected resume handling so later failures remove earlier success states, while subsequent successes restore them.

### DIFF
--- a/packages/core/src/db/workflow-events.test.ts
+++ b/packages/core/src/db/workflow-events.test.ts
@@ -401,6 +401,63 @@ describe('workflow-events', () => {
       ]);
     });
 
+    test('a later node_failed row supersedes an earlier node_completed row for the same step (#2705 R2)', async () => {
+      // Before #2705, a non-always_run node that once completed could never be
+      // re-attempted, so a node_completed → node_failed sequence for the same
+      // step_name was impossible. #2705's own invalidate-and-re-execute
+      // mechanism is the first thing that can produce that sequence: a cached
+      // node gets invalidated, re-executes, and the fresh attempt fails. If the
+      // earlier node_completed entry survives in the map, the NEXT resume's
+      // cache-eligibility check sees a stale success instead of the node's own
+      // most recent (failed) outcome.
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'flaky',
+            event_type: 'node_completed',
+            data: { node_output: 'first attempt succeeded' },
+          },
+          {
+            step_name: 'flaky',
+            event_type: 'node_failed',
+            data: { error: 'second attempt (re-executed via invalidation) crashed' },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-completed-then-failed');
+
+      // The failed row is the step's most recent real outcome; the stale
+      // node_completed entry must not survive the chronological walk.
+      expect(result.completedNodeOutputs.has('flaky')).toBe(false);
+    });
+
+    test('a node_completed row after a node_failed row for the same step re-adds it (chronological order wins)', async () => {
+      // Symmetric check: if the node later succeeds (a subsequent resume re-runs
+      // it and this time it completes), that success must be honoured — the
+      // delete on node_failed must not leave a permanent tombstone.
+      mockQuery.mockResolvedValueOnce(
+        createQueryResult([
+          {
+            step_name: 'flaky',
+            event_type: 'node_failed',
+            data: { error: 'first attempt crashed' },
+          },
+          {
+            step_name: 'flaky',
+            event_type: 'node_completed',
+            data: { node_output: 'second attempt succeeded' },
+          },
+        ])
+      );
+
+      const result = await getDagResumeSnapshot('run-failed-then-completed');
+
+      expect(result.completedNodeOutputs.get('flaky')).toEqual({
+        output: 'second attempt succeeded',
+      });
+    });
+
     test('sums cache axes reported by a failed node into the resumed total', async () => {
       mockQuery.mockResolvedValueOnce(
         createQueryResult([
--- a/packages/core/src/db/workflow-events.ts
+++ b/packages/core/src/db/workflow-events.ts
@@ -287,7 +287,13 @@ export async function getDagResumeSnapshot(workflowRunId: string): Promise<{
       );
       continue;
     }
-    if (row.event_type !== 'node_failed' && typeof data.node_output === 'string') {
+    if (row.event_type === 'node_failed') {
+      // A later failure for this step supersedes any earlier node_completed /
+      // node_skipped_prior_success entry (#2705 R2) — otherwise a node the engine's own
+      // prior-cache invalidation re-executed, and which then genuinely failed, is
+      // reported as a cached success again on a subsequent resume.
+      completedNodeOutputs.delete(row.step_name);
+    } else if (typeof data.node_output === 'string') {
       completedNodeOutputs.set(row.step_name, {
         output: data.node_output,
         // The node's logical value (#2637), persisted beside its text by the emit
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -11343,6 +11343,127 @@ describe('executeDagWorkflow -- prior-success cache invalidated by dep re-execut
     expect(data.invalidating_deps).toEqual(['producer']);
     expect(data.prior_structured_output).toEqual({ stale: true });
   });
+
+  it('invalidates a cached downstream when ONE of its multiple deps re-ran (any-stale triggers re-execution)', async () => {
+    // Pin the "any stale dep invalidates" trigger for cached consumers with >= 2 deps
+    // (#2402). The single-dep cases above can't catch a regression to "re-run only when
+    // ALL deps are stale" (`if (staleDeps.length === deps.length)`), which would
+    // re-introduce the original bug for any cached consumer that has at least one
+    // fresh dep and one stable one. The shape: producer re-runs (always_run, fresh
+    // output) + side stays cached (text-identical to prior) -> consumer is cached
+    // but reads `$producer.output` in its prompt. The fresh producer MUST invalidate
+    // the cached consumer even though side is stable.
+    const seenPrompts: string[] = [];
+    let queryCount = 0;
+    mockSendQueryDag.mockImplementation(async function* (prompt: string) {
+      seenPrompts.push(prompt);
+      queryCount++;
+      // queryCount 1: producer re-runs (always_run).
+      // queryCount 2: consumer re-runs after cache invalidation.
+      // Side is in priorCompletedNodes with text-identical-to-prior output, so the
+      // pre-populate loop seeds it and the resume-skip arm fires for it — side is
+      // never sent to sendQuery.
+      yield {
+        type: 'assistant',
+        content: queryCount === 1 ? 'FRESH producer output' : 'consumer result',
+      };
+      yield { type: 'result', sessionId: 'session-id' };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    // All three nodes cached in the prior run. Side's cached output matches what its
+    // pre-populated nodeOutputs entry will carry, so case 2 (text diff) reports no
+    // staleness for side. Producer's cached output is the stale value — the fresh
+    // always_run execution will surface case 2 and invalidate consumer.
+    const priorCompletedNodes = new Map<string, PersistedNodeOutput>([
+      ['producer', { output: 'STALE producer output' }],
+      ['side', { output: 'stable side output' }],
+      ['consumer', { output: 'STALE consumer output' }],
+    ]);
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-2402-multi-dep',
+      testDir,
+      {
+        name: 'stale-cache-multi-dep',
+        nodes: [
+          // always_run forces producer to re-execute this resume (case 2 fires).
+          { id: 'producer', command: 'producer', always_run: true },
+          // Stable dep — text matches prior so case 2 doesn't flag it.
+          { id: 'side', command: 'side' },
+          // Consumer reads producer; its dep list has BOTH producer and side.
+          {
+            id: 'consumer',
+            prompt: 'Use $producer.output and $side.output',
+            depends_on: ['producer', 'side'],
+          },
+        ],
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig,
+      undefined,
+      undefined,
+      priorCompletedNodes
+    );
+
+    // Exactly two queries: producer re-runs (always_run), consumer re-runs
+    // (cache invalidated). Side is skipped via the pre-populate prior-success arm,
+    // so it never reaches sendQuery.
+    expect(queryCount).toBe(2);
+
+    // Consumer's prompt must contain the FRESH producer output (not the stale cached
+    // value the resume snapshot carried). Side's stable output is also fine.
+    const consumerPrompt = seenPrompts[1];
+    expect(consumerPrompt).toContain('FRESH producer output');
+    expect(consumerPrompt).not.toContain('STALE producer output');
+    expect(consumerPrompt).toContain('stable side output');
+
+    // Consumer emits node_prior_cache_invalidated naming ONLY the stale dep. If a
+    // future refactor flagged every dep in the list (or stopped pushing after the
+    // first dep), invalidating_deps would either be ['producer','side'] or []
+    // respectively — both would fail this length-1 assertion.
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls;
+    const consumerSkippedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_skipped_prior_success' &&
+        (call[0] as { step_name: string }).step_name === 'consumer'
+    );
+    expect(consumerSkippedEvent).toBeUndefined();
+
+    const invalidatedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_prior_cache_invalidated' &&
+        (call[0] as { step_name: string }).step_name === 'consumer'
+    );
+    expect(invalidatedEvent).toBeDefined();
+    const data = (invalidatedEvent![0] as { data: Record<string, unknown> }).data;
+    expect(data.reason).toBe('stale_dependency');
+    // Length 1 (not 2) — pinning the "any stale dep invalidates" trigger; pinning
+    // that only the actually-stale dep is reported (not the stable side).
+    expect(data.invalidating_deps).toEqual(['producer']);
+    expect(data.prior_output).toBe('STALE consumer output');
+
+    // Side stayed cached, so it should emit node_skipped_prior_success exactly once.
+    const sideSkippedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_skipped_prior_success' &&
+        (call[0] as { step_name: string }).step_name === 'side'
+    );
+    expect(sideSkippedEvent).toBeDefined();
+  });
 });
 
 describe('executeDagWorkflow -- break after result (no hang on subprocess exit)', () => {
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -11003,6 +11003,348 @@ describe('executeDagWorkflow -- always_run resume opt-out', () => {
   });
 });
 
+describe('executeDagWorkflow -- prior-success cache invalidated by dep re-execution (#2402)', () => {
+  let testDir: string;
+
+  beforeEach(async () => {
+    testDir = join(tmpdir(), `dag-2402-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+    const commandsDir = join(testDir, '.archon', 'commands');
+    await mkdir(commandsDir, { recursive: true });
+    await writeFile(join(testDir, '.archon', 'commands', 'producer.md'), 'Producer prompt');
+    await writeFile(
+      join(testDir, '.archon', 'commands', 'consumer.md'),
+      'Consumer prompt $producer.output'
+    );
+
+    mockSendQueryDag.mockClear();
+    mockGetAgentProviderDag.mockClear();
+
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'claude',
+      getCapabilities: mockClaudeCapabilities,
+    }));
+  });
+
+  afterEach(async () => {
+    mockGetAgentProviderDag.mockImplementation(() => ({
+      sendQuery: mockSendQueryDag,
+      getType: () => 'claude',
+      getCapabilities: mockClaudeCapabilities,
+    }));
+    try {
+      await rm(testDir, { recursive: true, force: true });
+    } catch {
+      // ignore cleanup errors
+    }
+  });
+
+  it('invalidates a cached downstream when an always_run dep re-runs with fresh output', async () => {
+    // Bug shape from #2402: producer is `always_run` so it re-executes on resume; consumer
+    // is also cached from the prior run. The naive resume used to skip consumer and read
+    // the stale cached value, reporting success over synthesis made against the OLD
+    // producer. After the fix, consumer must re-execute against the fresh producer.
+    const seenPrompts: string[] = [];
+    let queryCount = 0;
+    mockSendQueryDag.mockImplementation(async function* (prompt: string) {
+      seenPrompts.push(prompt);
+      queryCount++;
+      yield {
+        type: 'assistant',
+        content: queryCount === 1 ? 'fresh producer output' : 'consumer result',
+      };
+      yield { type: 'result', sessionId: 'session-id' };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    // Both producer and consumer were completed in the prior run — this is the
+    // scenario the bug report describes.
+    const priorCompletedNodes = new Map<string, PersistedNodeOutput>([
+      ['producer', { output: 'STALE_CACHED_PRODUCER' }],
+      ['consumer', { output: 'STALE_CACHED_CONSUMER' }],
+    ]);
+
+    await executeDagWorkflow(
+      mockDeps,
+      platform,
+      'conv-2402-always-run',
+      testDir,
+      {
+        name: 'stale-cache-always-run',
+        nodes: [
+          { id: 'producer', command: 'producer', always_run: true },
+          { id: 'consumer', command: 'consumer', depends_on: ['producer'] },
+        ],
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig,
+      undefined,
+      undefined,
+      priorCompletedNodes
+    );
+
+    // Producer re-runs (always_run) AND consumer re-runs (cache invalidated by dep) —
+    // before the fix, consumer was skipped and only 1 sendQuery call happened.
+    expect(mockSendQueryDag.mock.calls.length).toBe(2);
+
+    // Consumer's prompt must contain the FRESH producer output, not the stale cached
+    // value the resume snapshot carried.
+    const consumerPrompt = seenPrompts[1];
+    expect(consumerPrompt).toContain('fresh producer output');
+    expect(consumerPrompt).not.toContain('STALE_CACHED_PRODUCER');
+
+    // Audit: producer emits node_always_run_reset; consumer emits
+    // node_prior_cache_invalidated naming producer as the invalidating dep.
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls;
+    const consumerSkippedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_skipped_prior_success' &&
+        (call[0] as { step_name: string }).step_name === 'consumer'
+    );
+    expect(consumerSkippedEvent).toBeUndefined();
+
+    const invalidatedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_prior_cache_invalidated' &&
+        (call[0] as { step_name: string }).step_name === 'consumer'
+    );
+    expect(invalidatedEvent).toBeDefined();
+    const data = (invalidatedEvent![0] as { data: Record<string, unknown> }).data;
+    expect(data.reason).toBe('stale_dependency');
+    expect(data.invalidating_deps).toEqual(['producer']);
+    expect(data.prior_output).toBe('STALE_CACHED_CONSUMER');
+  });
+
+  it('does NOT invalidate when a cached dep ran but produced text-identical output', async () => {
+    // The comparison is conservative: a dep that re-ran with byte-identical text
+    // and structured output is NOT flagged as stale. The cached downstream's value
+    // is semantically equivalent, so honouring the cache is correct (and avoids an
+    // unnecessary re-execution). This pins that contract for #2402.
+    let queryCount = 0;
+    mockSendQueryDag.mockImplementation(async function* () {
+      queryCount++;
+      yield { type: 'assistant', content: 'identical output' };
+      yield { type: 'result', sessionId: 'session-id' };
+    });
+
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const platform = createMockPlatform();
+    const workflowRun = makeWorkflowRun();
+
+    // Producer was completed in the prior run with output "identical output". It is NOT
+    // `always_run`, so the resume pre-populates nodeOutputs["producer"] = "identical
+    // output". When this resume runs, the comparison finds the values match and the
+    // cache is honoured. Only producer would execute on a fresh start — but here
+    // producer is also in priorCompletedNodes, so it is skipped entirely.
+    const priorCompletedNodes = new Map<string, PersistedNodeOutput>([
+      ['producer', { output: 'identical output' }],
+      ['consumer', { output: 'cached consumer output' }],
+    ]);
+
+    await executeDagWorkflow(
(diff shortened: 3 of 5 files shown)


Wirasm:
# Review report — PR #2705 (`fix(workflows): invalidate cached prior-success nodes when a dependency re-ran`, #2402)

## Verdict

**Ready.** The delta adds the missing test that pins the `|| current?.state === 'failed'` arm of `getStaleCachedDependencies` case 1, closing the T1 coverage gap from the prior round. T1, R1, and R2 are all closed. The single outstanding item (S1) is a non-blocking Suggestion carried over from round 2 and does not hold readiness back.

## Reviewed head SHA

`ba1de70323a09f85ab5b3abd3667f81210f62e0e`

## Findings

### Critical

None.

### Important

None open.

### Rejected findings

None — every lens report produced no findings for the delta.

#### S1 — Test title and leading prose at `dag-executor.test.ts:11129` describe a scenario the test does not exercise *(carried over)*
The `it('does NOT invalidate when a cached dep ran but produced text-identical output')` test and its leading block comment (lines `11130-11134`) describe a re-running dep, but the setup uses a producer without `always_run` and the assertion `expect(queryCount).toBe(0)` at line `11184` confirms the dep did NOT run — only the pre-populate path was exercised. The helper's `Note:` docstring paragraph at `packages/workflows/src/dag-executor.ts:7985-7989` also asserts the stronger "re-ran with text- and structured-identical output" contract that no test in this PR pins. Remains Accepted-Not-Actioned from the prior round — does not block readiness. Smallest correction: rename the test to `does NOT invalidate when a pre-populated dep has text-identical output`, weaken the leading comment, and drop the redundant "re-ran" claim from the helper docstring's `Note:` paragraph.

| ID | Severity | Lens | One-line claim | Verdict | Evidence |
|---|---|---|---|---|---|
| **T1** | Important | `tests.md` | The `\|\| current?.state === 'failed'` arm added by `2cca308d` has zero test coverage — a regression that drops it is observationally indistinguishable from the intended implementation. | **Fixed at `ba1de703`.** The new test at `packages/workflows/src/dag-executor.test.ts:11265-11374` exercises exactly the missing-coverage path: producer not in `priorCompletedNodes`, re-attempts this resume and throws (`Error('producer crashed: provider auth error')`); consumer is prior-cached with `depends_on: ['producer']` and `trigger_rule: 'all_done'`. Asserts `queryCount === 2` (`:11349`), `consumerSkippedEvent === undefined` (`:11362`), `invalidatedEvent` defined (`:11370`) with `data.reason === 'stale_dependency'` (`:11374`), `data.invalidating_deps === ['producer']` (`:11375`), `data.prior_output === 'STALE_CACHED_CONSUMER'` (`:11376`). Verified by the code lens: scratch-mutating `dag-executor.ts:8005` to drop the `'failed'` arm causes `expect(queryCount).toBe(2)` to fail with `Received: 1`. Test placement (immediately after the case-1 `'completed'` test at `11201-11263`) puts the two case-1 arms side-by-side. | `packages/workflows/src/dag-executor.test.ts:11265-11374` (the new test); `packages/workflows/src/dag-executor.ts:8005` (the production condition being pinned). |
| **R1** | Important | `types.md` | `getStaleCachedDependencies` case 1 silently honors the cache when a dep re-ran and failed. | **Closed (both halves).** Code half closed at `2cca308d` (round 1) — case 1 widened at `dag-executor.ts:8005`. Test-coverage half closed by this delta's new test (same evidence as T1 above). No production-code findings from this delta. | `packages/workflows/src/dag-executor.ts:8005` (widened condition); `packages/workflows/src/dag-executor.test.ts:11265-11374` (the test that pins the widening). |
| **R2** | Important | `tests.md` | The "any stale dep invalidates" trigger is not pinned by any test (all four new tests use single-dep consumers). | **Closed at `d9f526bc` (round 2),** untouched by this delta. The multi-dep test at `packages/workflows/src/dag-executor.test.ts:11459-11578` pins the trigger via `data.invalidating_deps.length === 1` (`toEqual(['producer'])` at `:11568`), `queryCount === 2` (`:11537`), consumer `node_skipped_prior_success` absent (`:11556`), and side `node_skipped_prior_success` present (`:11577`). `git diff d9f526bc..ba1de703` confirms zero overlap with lines 11459-11578. | `packages/workflows/src/dag-executor.test.ts:11537, :11556, :11568, :11577`. |
| **S1** | Suggestion (non-blocking) | `comments.md` | Test title and leading prose at `dag-executor.test.ts:11129` describe a re-running dep the test does not exercise; same wording leaks into the helper's `Note:` docstring. | **Still open as Accepted-Not-Actioned.** `grep -n "does NOT invalidate when a cached dep ran" packages/workflows/src/dag-executor.test.ts` still hits `11129`; the new test at `:11265` does not touch any affected line. Helper `Note:` paragraph at `dag-executor.ts:7985-7989` unchanged. Non-blocking. | `packages/workflows/src/dag-executor.test.ts:11129-11189`; `packages/workflows/src/dag-executor.ts:7985-7989`. |

## Lens roster

| Lens | Status |
|---|---|
| `code.md` (correctness) | Ran — no findings; empirically verified the test catches the dropped-arm regression |
| `seams.md` (cross-component contracts) | Ran — no findings; every writer/reader contract holds at the new test's assertion surface |
| `tests.md` (regression protection) | Ran — no findings; T1 closed by the delta's new test |
| `errors.md` (failure surfaces) | Ran — no findings; traced the failure-source → suppression-point → false-success path end-to-end |
| `comments.md` (prose accuracy) | Ran — no findings on the delta; S1 carried over as Accepted-Not-Actioned |
| `types.md` (type-level invariants) | Carried over from round 1 (R1 closed both halves; no new findings for the delta since production code is unchanged) |
| `simplify.md` (minimality) | Ran — no findings; the new test is the smallest correction for T1 |
| `docs.md` | Skipped by design this round (no user-facing docs were touched by the delta) |

Every lens file present in `…/review/` is accounted for above; `docs.md` (absent) is recorded as skipped by design rather than as silent clearance.

---

*This report is the canonical review for PR #2705 at head `ba1de70323a09f85ab5b3abd3667f81210f62e0e`, round 3 (delta since prior cursor `d9f526bc433f16a58d48c8e3edf818cc08c68375`). Published as a single GitHub comment carrying the marker `<!-- archon-review-report -->` on its first line; not duplicated as separate per-thread comments.*

**Canonical PR comment:** https://github.com/coleam00/Archon/pull/2705#issuecomment-5377817506 (read-back confirmed; body matches artifact byte-for-byte modulo GitHub's trailing-newline normalization).

Wirasm:
## Ready to merge

Round 2 verified the correction pushed against round 1's two blocking findings (`ba1de703` → `407d918b`, commits `bbbdb669` and `407d918b`). Seven independent reviewers, each working from the prior round's exact finding text and this round's bounded correction diff, re-derived the same conclusion from the same code: both R1 (a previously-cached dependency's fresh failure could be masked by value-equality against its prior success) and R2 (`getDagResumeSnapshot` never cleared a stale `node_completed` entry on a later `node_failed` row) are fixed at the current head, with no correction-caused regression in seams, silent-failure, type-design, test-quality, general-correctness, simplification, or comment-accuracy. R3 (misleading test) and R4 (inaccurate PR-body rationale) are also confirmed fixed. The coordinator additionally traced the reachability precondition for R1 independently (`nodeOutputs` pre-population always writes `state: 'completed'`, never `'failed'`, so the new unconditional check can only ever fire for a dependency that genuinely failed fresh this resume — no false-positive path exists) and traced a 3-resume invalidate→fail→invalidate chain by hand against both fixes together; both hold.

**Validation:** type-check, lint, format:check all clean across every package. `bun test packages/workflows/src/dag-executor.test.ts` — 590/590 pass (including the rewritten R3 test and the new R1 regression test). `bun test packages/core/src/db/workflow-events.test.ts` — 34/34 pass (including both new R2 regression tests). `bun run --filter @archon/workflows test` and `bun run --filter @archon/core test` (full package suites) — both exit 0.

**Resolved:** R1, R2, R3, R4 (all FIXED this round)

### Findings

|---|---|---|---|
| `R1` | Important | `getStaleCachedDependencies` cases 2/3 never re-checked a previously-cached dependency's terminal state, so a fresh failure matching a prior success's output value was reported "not stale." | FIXED |
| `R2` | Important | `getDagResumeSnapshot` never cleared an earlier `node_completed` entry when a later `node_failed` row landed for the same step — a sequence this PR's own invalidation mechanism newly makes reachable. | FIXED |
| `R3` | Suggestion | The "does NOT invalidate... identical output" test's producer never actually re-executed, so the conservative no-flag-on-identical-rerun contract was pinned by no real test. | FIXED |
| `R4` | Suggestion | The PR body's stated reason for rejecting the "executed this run" `Set` alternative ("would require new persistence") was inaccurate. | FIXED |
| `R5` | Suggestion | R2's fix makes `priorCompletedNodes.size === 0` (⇒ hard CLI resume-refusal) correctly reachable in one additional narrow scenario it previously, buggily, avoided. | DECLINED |

<details>
<summary><code>R1</code> — Dependency-staleness check misses a fresh failure that coincides in value with a prior success</summary>

**Impact (round 1):** A cached downstream node was returned via `node_skipped_prior_success` — before `checkTriggerRule` was ever consulted — even though one of its dependencies just failed on this exact resume attempt.

**Fix:** `packages/workflows/src/dag-executor.ts:8004-8007` — `getStaleCachedDependencies` now runs `if (current?.state === 'failed') { stale.push(depId); continue; }` unconditionally, before the `prior === undefined` lookup and before the case-2/3 value comparisons. The removed `|| current?.state === 'failed'` disjunct inside the old case-1 branch is now dead code by construction (the unconditional check already `continue`s past every failed dep before case 1 runs) and was correctly deleted, not left as redundant.

**Verification:**
- Read the full function at its current location; confirmed the check is unconditional and runs first (independently by 5 of 7 reviewers plus the coordinator).
- Coordinator traced the reachability precondition directly: `ctx.nodeOutputs` pre-population (`dag-executor.ts:9756-9779`) always writes `state: 'completed'` for entries sourced from `priorCompletedNodes` — never `'failed'` — so `current?.state === 'failed'` can only be true for a dependency that actually executed and failed in the current resume pass. The fix therefore cannot produce a false-positive staleness result; every dependency it flags genuinely failed this resume.
- New regression test `dag-executor.test.ts:11383` ("...PREVIOUSLY-CACHED dep fails fresh with output matching its prior success (#2705 R1)") reproduces the exact original scenario: producer cached from a prior success with `output: ''`, fails fresh this resume with `output: ''` again. Traced against the pre-fix code by two reviewers independently: the old code's `prior !== undefined` branch never inspected `current.state`, found `'' === ''`, and did not flag it stale — this test would have failed on `ba1de703`. Ran and confirmed passing at `407d918b`.
- One reviewer (silent-failure-hunter) confirmed the fix is genuinely state-based, not secretly still value-dependent, by finding two failure arms with non-empty output (`loop_group` mid-iteration failure, `dag-executor.ts:4301`/`4328`) and confirming the check still catches them (it never reads `.output`).
- Reviewer coverage independently confirmed the invalidation still routes through the normal per-node dispatch pipeline (including `checkTriggerRule`) rather than bypassing it — the fix closes the "success reported before trigger-rule evaluation" defect at its root.

**Disposition:** FIXED. Verified by 5 independent reviewers plus the coordinator's own code trace; no dissent.

</details>

<details>
<summary><code>R2</code> — A node re-executed by this PR's own invalidation mechanism that then fails is reported cached-successful on the next resume</summary>

**Impact (round 1):** `getDagResumeSnapshot`'s `node_failed`-guarded write only ever added map entries for non-failed rows and never deleted an earlier `node_completed` entry for the same step, so a node's own most recent real outcome (failure) was invisible to a subsequent resume's cache-eligibility check.

**Fix:** `packages/core/src/db/workflow-events.ts:290-295` — when a `node_failed` row is processed, `completedNodeOutputs.delete(row.step_name)` now runs (replacing the prior no-op for failed rows); a later `node_completed` row for the same step still re-adds the entry via the unchanged `else if` branch, so a genuine subsequent success is not permanently tombstoned.

**Verification:**
- Confirmed the query's `ORDER BY created_at ASC, COALESCE(event_order, 0) ASC, id ASC` (`workflow-events.ts:268-271`) processes rows in true chronological order: `event_order` is a durable, monotonically-increasing sequence assigned at insert time on both Postgres (`nextval`) and SQLite (an `AFTER INSERT` trigger), confirmed by one reviewer reading both schema definitions. The `id ASC` final tiebreaker (a random UUID on Postgres) is a pre-existing, unrelated weakness that only matters for legacy rows predating the `event_order` migration — not reachable by any sequence this correction produces.
- Confirmed the delete is unconditional and not gated on the later `if (data.aggregate === true) continue;` usage-aggregation guard — read the whole loop body, not just the diff hunk.
- Traced consumers: `dag-executor.ts:8171/8182` gate cache-eligibility purely on `priorCompletedNodes?.has(node.id)`, which now correctly returns `false` for a step whose latest event is `node_failed`. `executor.ts:942-945` (`hydrateResumableRun`) derives its provider-session carry-forward from the same map, so a re-executed node also correctly drops its stale session.
- Both new tests read and verified: `workflow-events.test.ts:404` (completed→failed: entry deleted, would fail on pre-fix code) and `:435` (failed→completed, symmetric: entry correctly re-added, guards against an over-corrected "permanent tombstone" implementation). Ran: 34/34 pass.
- Coordinator independently traced a 3-resume chain (resume 1: producer+consumer complete; resume 2: producer re-runs and fails, consumer invalidated via R1 and re-executes successfully; resume 3 snapshot) by hand against the row sequence this produces, and confirmed the resulting `priorCompletedNodes` state is correct in both directions (producer absent, consumer carries its resume-2 output).

**Disposition:** FIXED. Verified by 3 independent reviewers plus the coordinator's own multi-resume trace; no dissent.

</details>

<details>
<summary><code>R3</code> — Test title and docstring claim a dependency re-ran with identical output; the test never let it run</summary>

**Fix:** `packages/workflows/src/dag-executor.test.ts:11129` — the producer node now declares `always_run: true`, so it genuinely re-executes this resume; the assertion changed from `expect(queryCount).toBe(0)` to `expect(queryCount).toBe(1)`.

**Verification:** Two reviewers independently confirmed the producer now actually dispatches (traced `always_run` handling at `dag-executor.ts:8182-8198`, which unconditionally re-executes regardless of cache membership) and that `queryCount === 1` is a genuine discriminator — an over-eager staleness check that incorrectly flagged an identical re-run would push this to 2. Ran with `-t "2705"`: passes.

**Disposition:** FIXED.

</details>

<details>
<summary><code>R4</code> — PR body's stated reason for rejecting the "completion-generation counter" alternative was imprecise</summary>

**Fix:** PR body's "Solution" section now states the rejected `Set` alternative "would be exactly as ephemeral as `ctx.nodeOutputs`, so it would not actually require new persistence; its real cost is a larger blast radius... threading a clean 'did this node actually dispatch fresh work' signal through all of them is a materially bigger change."

**Verification:** Read live PR body; confirmed the corrected rationale matches what was requested and is restated consistently in the "Review round 2" section.

**Disposition:** FIXED.

</details>

<details>
<summary><code>R5</code> — R2's fix makes a narrow "zero completed nodes" resume-refusal correctly reachable in one more case</summary>

**Impact:** `executor.ts:935` treats `priorCompletedNodes.size === 0` (with no other resumable gate state) as "nothing to resume," and the CLI hard-errors with "Cannot resume: ... has no completed nodes." If a run's only-ever-successful node is later invalidated (by R1) and fails on a subsequent resume, `size` can now correctly drop to 0 where the pre-fix bug spuriously kept it above 0. The user sees a hard resume-refusal in that specific case instead of the (buggy) alternative, which the reporting reviewer noted would have been functionally equivalent to a fresh run anyway.

**Evidence:** `packages/core/src/db/workflow-events.ts:290-295` (R2 fix) combined with `packages/workflows/src/executor.ts:935` and the CLI's resume-refusal at `packages/cli/src/commands/workflow.ts:2114-2118`.

**Required outcome:** None. This is the correct, intended consequence of fixing R2 — the pre-existing "no completed nodes to resume" guard is deliberate design elsewhere in the resume path, and R2 simply makes it fire accurately for a run whose one recorded success has since been proven stale. Not a defect in this PR's scope.

**Disposition:** DECLINED. Genuine and correctly reasoned observation, but describes intended behavior of the fix, not a defect; no action required for this PR.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R2 verified fixed; R5 raised (new, declined) |
| `seams` | R1 verified fixed; confirmed R1 (in-memory) and R2 (persisted-log) are the two matching halves of the same seam, neither left stale |
| `simplify` | R1, R4 verified fixed; confirmed the R1 fix is a net simplification (removes a duplicated special case rather than adding one) |
| `types` | R1 verified fixed; confirmed `NodeOutput.state` is a closed discriminated union and the new check is exhaustively correct against all reachable states; confirmed `PersistedNodeOutput` (no `state` field) keeps R1 and R2 independent, not coupled |
| `errors` | R1, R2 verified fixed; confirmed no new silent-failure path, confirmed fix is state-based not value-dependent via non-empty-output failure arms |
| `tests` | R1, R2, R3 verified fixed with genuine (non-vacuous) regression tests; scanned for and ruled out further untested adjacent edge cases (skipped-dep collision is unreachable; 3-resume chains reduce by induction to the existing 2-row tests) |
| `comments` | R3 verified fixed; docstring, inline comments, and PR body cross-checked for accuracy against current code — no inconsistency found |

#### Validation

|---|---|---|
| `bun run type-check` | PASS | All 13 packages exit 0 |
| `bun run lint` | PASS | `bun x eslint . --cache` — no output, exit 0 |
| `bun run format:check` | PASS | `bun x prettier --check .` — "All matched files use Prettier code style!" |
| `bun test src/dag-executor.test.ts` (packages/workflows) | PASS | 590 pass, 0 fail, 1667 expect() calls |
| `bun test src/db/workflow-events.test.ts` (packages/core) | PASS | 34 pass, 0 fail, 70 expect() calls |
| `bun run --filter @archon/workflows test` (full package suite) | PASS | Exit code 0 across all invocations |
| `bun run --filter @archon/core test` (full package suite) | PASS | Exit code 0 across all invocations |

`bun run check:schema-upgrades` and the repo-root `bun run validate` bundle were not run — the correction touches only `packages/workflows/src/dag-executor.ts`, `packages/workflows/src/dag-executor.test.ts`, `packages/core/src/db/workflow-events.ts`, and `packages/core/src/db/workflow-events.test.ts` (confirmed via `git diff ba1de703..407d918b --stat`); no schema, migration, or bundled-defaults file changed.

</details>

PR #2710, opened 2026-08-22
fix(workflows): a failed loop_group's binding now fails its consumer

### LINKED ISSUE #2696: test(workflows): prove a binding can read a loop_group's output across a skipped branch — the real deliver topology
PR #2692 (closes #2637) proves the deliver pattern with an `all_done` join binding two top-level `when:`-gated producers via `{ from, if_skipped }`. The real shape in the sdlc pack (PR #2627) is harder: one producer is the corrections `loop_group`, so the join binds the **group node's** output, not a sibling's.

Three cases need engine coverage, none of which the merged test exercises:

1. **Completed group:** `{ from: '$corrections.output', if_skipped: null }` — what exactly is a loop_group's whole-ref value (raw text of the body sink's final iteration?), and does the binding deliver it? Completion implies the authored condition held, because `until_bash` is the group's only success exit.
2. **Skipped group** (the initial review was already ready): the binding takes `if_skipped`.
3. **Failed group** (`max_iterations` exhausted): the binding must fail the consuming join — a bound-exhausted correction loop must never be blessed by a default. #2692's resolver covers skipped/pending only, so failed should fall through to strict resolution; pin that with a test.

The pack is swapping its `.ready` latch for exactly this shape now; dry-run and dogfood evidence from #2627 will land here. If the swap hits a gap (for example the group's whole-ref value is unusable for presence discrimination), this issue owns the engine-side fix.

Wirasm:
## Problem and outcome

A `{ from: '$node.output', if_skipped }` binding directive on a failed producer resolved instead of failing. For a plain node this rarely showed up, since ordinary failure paths return `output: ''` and the binding's fielded reads accidentally choke on the empty string — but a `loop_group`'s failure paths (body-node failure, `max_iterations` exhaustion, cancellation) all preserve the last completed iteration's real, non-empty output text. A binding joined via `trigger_rule: all_done` onto a failed `loop_group` therefore silently resolved that stale text as if the group had succeeded, and the consumer ran its own body on it.

This reproduced live (issue #2696, run `6607bf20`): a corrections `loop_group` failed after its recheck lenses exhausted retries under provider rate-limiting. The downstream `gate-ready` script node — bound via `{ from: '$corrections.output', if_skipped: null }` — resolved against the failed group's stale output, inferred `ready:true`, and a later `flip-ready` node flipped a PR to ready with its final correction commit unreviewed, while the workflow run itself finished `failed`. A split-brain terminal: the run reported failure, but a public, hard-to-reverse side effect fired anyway.

- **Outcome:** A binding on a failed producer — plain node or `loop_group` — now always fails the consuming node. `if_skipped` never applies to a failed producer, regardless of what text happens to be stored as its output.
- **Invariant:** `resolveBindingDirective` never returns a value, real or defaulted, for a producer whose state is `'failed'`.
- **Scope boundary:** No change to `wholeRefLogicalValue`, `resolveNodeOutputField`, the plain-string whole-ref tier of `resolveWorkflowValue`, `checkTriggerRule`'s `all_done` semantics, or how a `loop_group`'s failure paths compute their stored output. Those are separate, unrequested surfaces.
- **Root cause:** `resolveBindingDirective`'s only state guard checked `undefined | 'skipped' | 'pending'`; `'failed'` fell through to an unconditional value lookup with no state check at all.

## Review guidance

- **Feedback requested:** Correctness of the new guard's placement and error message; whether the marker-file idiom in the third test is a convincing proxy for "the consumer's real body never ran."
- **Start here:** `packages/workflows/src/dag-executor.ts:433` — the new `producer.state === 'failed'` branch, the entire fix.
- **Review order:** `dag-executor.ts` (the fix) → `dag-executor.test.ts`, the three new tests immediately after the "deliver pattern (PR #2627)" test (search `#2696`).
- **Lower-attention areas:** None — this is a two-file, single-behavior change.

## Solution

One explicit branch in `resolveBindingDirective`, ordered after the existing skipped/pending guard and before the value lookup: if the producer's state is `'failed'`, throw unconditionally, naming both the failed producer and its real error. The branch applies uniformly to any producer type — no `loop_group`-specific code — because the existing bug was never about `loop_group` specifically, only exposed by it (a `loop_group`'s failure paths happen to leave real output text behind, where a plain node's usually don't).

Three tests build the real `deliver` topology — a bash-only `corrections` loop_group feeding a `gate-ready` script consumer joined via `all_done` — across all three producer states the issue named: completed (resolves to the real output), skipped (takes the declared default), and failed (fails the consumer, whose body never executes). The failed-group test is the reproduction: temporarily reverting the fix makes it fail exactly as run `6607bf20` did (the consumer's marker-file write happens), confirming it is a real regression guard.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A binding on a failed `loop_group` resolved its last-completed-iteration output as if the group had succeeded; the consumer ran normally. | A binding on a failed producer (any node type) always throws; the consumer never runs its body. |
| **Failure behavior** | Silent: the consumer succeeded on stale data, and only the unrelated `loop_group` node showed as failed. | Explicit: the consumer itself fails, with an error naming the failed producer and its cause. |

## Validation

- `bun test src/dag-executor.test.ts -t "loop_group binding \(#2696\)"` (from `packages/workflows/`) — 3 pass, 0 fail — proves all three cases.
- Regression check: stashed the fix and reran the same three tests — case 3 failed (`Expected: false / Received: true`, the marker file existed), cases 1 and 2 still passed. Confirms the third test is a genuine guard, not vacuous, and that the fix doesn't change already-correct behavior.
- `bun test src/dag-executor.test.ts` (full file) — 586 pass, 0 fail — no regressions in the existing suite.
- `bun --filter @archon/workflows test` — exit code 0, every constituent suite `0 fail`.
- `bun run validate` — exit code 0. Type-check, lint (`--max-warnings 0`), format, the full per-package test suite, build, and all `check:bundled*`/`check:pi-vendor-map`/`check:capability-matrix` generators all passed; 148 individual `bun test` invocations across the monorepo each reported `0 fail`.
- **Not verified:** A real-run reproduction on top of the fix (rerunning the sdlc pack's `deliver` workflow through the exact `6607bf20` failure shape) — outside this PR's scope, which is engine-level coverage; the issue's own dogfood evidence already established the completed/skipped cases empirically, and the fix directly targets the diagnosed code path.

## Links

- Closes #2696
- Plan: https://github.com/coleam00/Archon/issues/2696#issuecomment-5379799060

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed workflow producers now correctly block dependent steps instead of being treated as skipped.
  * Prevents downstream steps from consuming invalid output from failed producers, including failed loop groups.
  * Added coverage for completed, skipped, and failed loop-group scenarios, including dry runs.

* **Documentation**
  * Clarified that `if_skipped` applies only when a producer does not run; producer failures always fail dependent bindings.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -945,8 +945,9 @@ binding, producer, and fix named — a binding never silently resolves to `''`.
 **failed** always fails the binding too, whether or not `if_skipped` is declared — a
 `loop_group`'s failure paths in particular can leave real, non-empty output behind (its last
 completed iteration's text), so this is an explicit check, not an accident of empty output.
-Declaring `if_skipped` never papers over a real failure; guard the consumer with `when:`
-instead if it should tolerate that branch.
+There is no way to opt a binding out of this: declaring `if_skipped` never papers over a real
+failure, and `when:` cannot substitute for it either — it reads the same output text, which is
+indistinguishable from a success for a `loop_group` that failed mid-run.
 
 Two guarantees the loader enforces: every bound producer must be reachable through
 `depends_on` (a binding can never race its producer), and two binding names may not fold to
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -941,6 +941,13 @@ is the [constitution's](/reference/workflow-language-constitution/) YAML-coordin
 code-computes split. A skipped producer with **no** `if_skipped` fails the node with the
 binding, producer, and fix named — a binding never silently resolves to `''`.
 
+`if_skipped` only ever covers a producer that **did not run**. A producer that ran and
+**failed** always fails the binding too, whether or not `if_skipped` is declared — a
+`loop_group`'s failure paths in particular can leave real, non-empty output behind (its last
+completed iteration's text), so this is an explicit check, not an accident of empty output.
+Declaring `if_skipped` never papers over a real failure; guard the consumer with `when:`
+instead if it should tolerate that branch.
+
 Two guarantees the loader enforces: every bound producer must be reachable through
 `depends_on` (a binding can never race its producer), and two binding names may not fold to
 the same `INPUTS_*` env key. At runtime, node-local bindings are the **nearest** input
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -25427,4 +25427,168 @@ describe('value transport (#2637): persistence, resume, and node-local bindings'
     await runDag(deliverWorkflow('iteration'), 'transport-deliver-iteration');
     expect(checkPrompts).toEqual(['joined [initial=false|iteration=true]']);
   });
+
+  // --- loop_group producer bindings, the real deliver topology (#2696) ---
+  //
+  // `deliver`'s `gate-ready` binds `{ from: '$corrections.output', if_skipped: null }`
+  // on a `loop_group` (not a plain node), joined via `trigger_rule: all_done`. Three
+  // cases: the group completes, the group is skipped, and the group fails. All three
+  // build the same shape — a bash-only `corrections` loop_group feeding a `gate-ready`
+  // script node — so each test constructs its own `nodes` array inline rather than a
+  // shared builder, keeping every producer state (completed/skipped/failed) visible at
+  // its call site.
+
+  it("loop_group binding (#2696): a completed group's output resolves through an all_done join, not the if_skipped default", async () => {
+    const nodes: DagNode[] = [
+      dagNodeSchema.parse({
+        id: 'corrections',
+        loop_group: {
+          until_bash: 'exit 0',
+          max_iterations: 1,
+          fresh_context: false,
+          nodes: [{ id: 'apply', bash: "printf 'CORRECTIONS_APPLIED'", depends_on: [] }],
+        },
+        depends_on: [],
+      }),
+      dagNodeSchema.parse({
+        id: 'gate-ready',
+        script: 'console.log(process.env.INPUTS_CORRECTIONS);',
+        runtime: 'bun',
+        depends_on: ['corrections'],
+        trigger_rule: 'all_done',
+        with: {
+          corrections: { from: '$corrections.output', if_skipped: 'NO_CORRECTIONS_RAN' },
+        },
+      }),
+    ];
+
+    const result = await executeDagWorkflow(
+      createMockDeps(),
+      createMockPlatform(),
+      'conv-lg-binding',
+      testDir,
+      { name: 'lg-binding-completed', nodes },
+      makeWorkflowRun('lg-binding-completed'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    // The group's real whole-ref output resolved — the `if_skipped` default never fired.
+    expect(result).toBe('CORRECTIONS_APPLIED');
+  });
+
+  it("loop_group binding (#2696): a skipped group's binding takes if_skipped", async () => {
+    const nodes: DagNode[] = [
+      dagNodeSchema.parse({ id: 'gate', bash: "printf 'skip'", depends_on: [] }),
+      dagNodeSchema.parse({
+        id: 'corrections',
+        loop_group: {
+          until_bash: 'exit 0',
+          max_iterations: 1,
+          fresh_context: false,
+          nodes: [{ id: 'apply', bash: "printf 'CORRECTIONS_APPLIED'", depends_on: [] }],
+        },
+        depends_on: ['gate'],
+        when: "$gate.output == 'run'",
+      }),
+      dagNodeSchema.parse({
+        id: 'gate-ready',
+        script: 'console.log(process.env.INPUTS_CORRECTIONS);',
+        runtime: 'bun',
+        depends_on: ['corrections'],
+        trigger_rule: 'all_done',
+        with: {
+          corrections: { from: '$corrections.output', if_skipped: 'NO_CORRECTIONS_RAN' },
+        },
+      }),
+    ];
+
+    const result = await executeDagWorkflow(
+      createMockDeps(),
+      createMockPlatform(),
+      'conv-lg-binding',
+      testDir,
+      { name: 'lg-binding-skipped', nodes },
+      makeWorkflowRun('lg-binding-skipped'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    // The group never ran — the binding took its declared default, unchanged by #2696.
+    expect(result).toBe('NO_CORRECTIONS_RAN');
+  });
+
+  it("loop_group binding (#2696): a failed group's binding fails the consumer instead of resolving stale output", async () => {
+    // Reproduces run 6607bf20 (issue #2696): the group fails via max_iterations
+    // exhaustion, but its one completed iteration left real, non-empty output behind
+    // (captured as `lastIterationOutput` before the completion check that then failed).
+    // Before the fix, `gate-ready` would resolve that stale text as if the group had
+    // succeeded and run its own body on it — the split-brain that flipped PR #2705.
+    const markerPath = join(testDir, 'gate-ready-ran.marker');
+
+    const nodes: DagNode[] = [
+      dagNodeSchema.parse({
+        id: 'corrections',
+        loop_group: {
+          until_bash: 'exit 1',
+          max_iterations: 1,
+          fresh_context: false,
+          nodes: [{ id: 'apply', bash: "printf 'CORRECTIONS_APPLIED'", depends_on: [] }],
+        },
+        depends_on: [],
+      }),
+      dagNodeSchema.parse({
+        id: 'gate-ready',
+        // A stand-in for `flip-ready`'s real public write: proves whether the
+        // consumer's own body ever ran, independent of the run's overall status.
+        script: `require('fs').writeFileSync(${JSON.stringify(markerPath)}, process.env.INPUTS_CORRECTIONS ?? '');`,
+        runtime: 'bun',
+        depends_on: ['corrections'],
+        trigger_rule: 'all_done',
+        with: {
+          corrections: { from: '$corrections.output', if_skipped: null },
+        },
+      }),
+    ];
+
+    const platform = createMockPlatform();
+    await executeDagWorkflow(
+      createMockDeps(),
+      platform,
+      'conv-lg-binding',
+      testDir,
+      { name: 'lg-binding-failed', nodes },
+      makeWorkflowRun('lg-binding-failed'),
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    // gate-ready's body never executed — no public-write consequence from a failed group.
+    expect(await Bun.file(markerPath).exists()).toBe(false);
+
+    // The failure names both the failed producer and the node whose binding rejected it.
+    const sent = (platform.sendMessage as Mock<(...args: unknown[]) => Promise<void>>).mock.calls
+      .map(c => String(c[1]))
+      .join('\n');
+    expect(sent).toContain("Node 'gate-ready'");
+    expect(sent).toContain("'corrections' failed");
+  });
 });
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -429,6 +429,20 @@ function resolveBindingDirective(
         `guard '${consumerId}' with a 'when:' condition.`
     );
   }
+  // A failed producer never falls back to `if_skipped` (#2696): that default exists for
+  // a branch that legitimately didn't run, not for a run that ran and produced a result
+  // that can't be trusted (a loop_group's failure paths carry the last completed
+  // iteration's real output text — non-empty, often valid JSON — which would otherwise
+  // resolve here as if the group had succeeded).
+  if (producer.state === 'failed') {
+    throw new Error(
+      `Node '${consumerId}' binding '${name}' reads '${directive.from}', but node ` +
+        `'${ref.nodeId}' failed (${producer.error}), so its output cannot be trusted. ` +
+        "A binding never falls back to 'if_skipped' for a failed producer — fix the " +
+        `failure, or guard '${consumerId}' with a 'when:' condition that excludes the ` +
+        'failed branch.'
+    );
+  }
   return wholeRefLogicalValue(producer, ref.nodeId, ref.field);
 }
 
--- a/packages/workflows/src/dry-run.test.ts
+++ b/packages/workflows/src/dry-run.test.ts
@@ -1459,6 +1459,38 @@ describe('dryRunWorkflow — node-local with: bindings (#2637)', () => {
     expect(result.outcome).toBe('completed');
     expect(result.trace.find(t => t.nodeId === 'gate')?.output).toBe('ready=false');
   });
+
+  test('a directive on a failed producer fails the consumer, never taking if_skipped (#2696)', async () => {
+    // Preview and the real run share resolveNodeBindings (see this file's docblock),
+    // so the failed-producer guard added for #2696 must behave identically here.
+    const workflow = makeTestWorkflow({
+      name: 'binding-failed',
+      nodes: [
+        { id: 'corrections', bash: 'echo boom >&2; exit 1' },
+        {
+          id: 'gate-ready',
+          script: 'console.log(`ready=${process.env.INPUTS_READY ?? "unset"}`)',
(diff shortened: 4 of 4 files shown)


Wirasm:
## Ready to merge

Third pass. The correction commit (`de107b5f`, bounded diff against the previously reviewed head `279a958b`) closes `R5`, the only open finding: it removes the unsound "guard the consumer with `when:` instead" clause from the failed-producer docs paragraph and replaces it with a plain statement that there is no opt-out — a failed producer's binding always fails whether or not `if_skipped` is declared, and `when:` cannot substitute for it because it reads the same output text, indistinguishable from a success for a failed `loop_group`. The bounded diff is confirmed exactly one file (`authoring-workflows.md`), 3 insertions / 2 deletions, confined to that paragraph's closing sentences — no other content changed, and no code file was touched. The new sentence's factual claim was independently re-verified against the actual resolvers: `condition-evaluator.ts`'s unfielded `resolveOutputRef` branch still has no `producer.state` check, and `when-atom.ts`'s grammar still has no state accessor, so `when:` genuinely cannot distinguish the two cases as the new text says. No remaining claim anywhere in the file that `when:` (or anything else) provides a working escape hatch. All prior findings remain in their verified dispositions — nothing regressed.

**Validation:** Focused re-verification of this correction: `prettier --check` passes on the changed file, and the bounded diff was confirmed to touch only that one markdown file (no code, no test, no schema/generated surface), so the previously-verified `type-check`/`lint`/test results at head `279a958b` still apply unchanged. Full `bun run validate` was not re-run by this pass (docs-only correction, nothing it could invalidate); the implementer's full `bun run validate` pass reported earlier in the PR still stands.

**Resolved:** 3 (R2, R3, R5) · **Tracked follow-ups:** https://github.com/coleam00/Archon/issues/2713

### Findings

|---|---|---|---|
| `R1` | Important | A failed producer's stale output still resolves through `when:` conditions, inline `$node.output` prompt/bash substitution, and plain (non-directive) `with:` values — the same incident class, different syntax | TRACKED FOLLOW-UP |
| `R2` | Suggestion | `dry-run.ts` shares `resolveNodeBindings` and its preview behavior for a failed-producer binding silently changed (correctly) with this fix, but no dry-run test pinned it | FIXED |
| `R3` | Suggestion | `authoring-workflows.md` documented `if_skipped` for a skipped producer but never stated the failed-producer contract | FIXED |
| `R4` | Suggestion | The three new tests call `executeDagWorkflow` directly instead of the file's local `runDag` helper | DECLINED |
| `R5` | Suggestion | The new docs paragraph (added to fix R3) recommends guarding a failed producer with `when:`, but `when:` cannot express that guard | FIXED |

<details>
<summary><code>R1</code> — Sibling readers of a failed producer's output have no equivalent guard</summary>

**Impact:** A workflow author who references a failed `loop_group`'s output via `when:` (e.g. `"$corrections.output != ''"` or a `.field` check), inline prompt/bash text (`"Corrections: $corrections.output"`), or a plain (non-directive) `with:` value — rather than the `{ from, if_skipped }` binding directive this PR fixes — hits the identical split-brain bug: a downstream node joined via `trigger_rule: all_done` executes its real body (including public side effects) on a failed group's stale last-completed-iteration text, exactly as run `6607bf20` did before this PR.

**Evidence:**
- `packages/workflows/src/dag-executor.ts:322-326` (`resolveWorkflowValue`'s plain-string whole-ref tier) — `if (producer !== undefined) return wholeRefLogicalValue(producer, ...)`, no state check. Re-confirmed unchanged at the correction head.
- `packages/workflows/src/condition-evaluator.ts:86-91` (`resolveOutputRef`, the `when:` resolver) — `if (!nodeOutput.output) return ''; return nodeOutput.output;`, no state check; its inline comment ("empty for failed nodes") is still directly false for a failed `loop_group`. Re-confirmed unchanged.
- `packages/workflows/src/dag-executor.ts:1217-1223` (`substituteNodeOutputRefs`'s unfielded branch, inline `$node.output` splicing in `prompt:`/`bash:`/`command:` bodies) — same shape, no state check. Re-confirmed unchanged.
- `dag-executor.ts:437` (`resolveBindingDirective`'s guard) is confirmed the *only* `producer.state === 'failed'` check across both files (`rg -n "producer.state"`).

**Required outcome:** Apply the same `producer.state === 'failed'` guard (or a small shared helper, since the exact thrown-message wording legitimately differs per caller) to the three sibling call sites, and correct `condition-evaluator.ts`'s stale "empty for failed nodes" comment once the guard exists. Tracked as a new issue rather than required in this PR, since it is a distinct, separately-scoped defect this PR's own stated invariant (`resolveBindingDirective never returns a value ... for a producer whose state is 'failed'`) does not claim to cover, and closing it does not require touching, reverting, or extending anything this PR ships.

**Disposition:** TRACKED FOLLOW-UP — https://github.com/coleam00/Archon/issues/2713. Re-verified at head `279a958b`: the correction commit touches no production code (`dag-executor.ts`/`condition-evaluator.ts` are byte-identical to `bed88a43`), the gap stands exactly as described, and issue #2713 is still open with a body that cites the same three sites at the same line numbers and correctly frames the fix shape. Disposition unchanged and still reasonable — the tracked scope remains distinct from this PR's own invariant.

</details>

<details>
<summary><code>R2</code> — Dry-run's shared resolver now behaves differently for a failed-producer binding, untested</summary>

**Impact:** `dry-run.ts:983` calls the same `resolveNodeBindings` this PR patches (its own docblock at `dry-run.ts:482-484` states the preview and the real run deliberately share one resolver "and cannot drift"). The failed-producer guard changed dry-run's preview behavior for a failed-producer binding as a side effect, and no test pinned it.

**Evidence:** `packages/workflows/src/dry-run.test.ts:1461-1494` — new test `'a directive on a failed producer fails the consumer, never taking if_skipped (#2696)'`. Traced the full call path: `execCode: true` routes the `corrections` bash node (`echo boom >&2; exit 1`) through `executeCodeNode` (`dry-run.ts:1007`), which spawns a real subprocess via `execFileAsync`/`resolveBashPath()` — a genuine failure, not a stub. The failure sets `outputs.set('corrections', { state: 'failed', ... })` (`recordFailed`, `dry-run.ts:564`), and the consumer's binding is resolved via the *same imported* `resolveNodeBindings` (`dry-run.ts:983`, `dag-executor.ts:382`) containing the `producer.state === 'failed'` guard (`dag-executor.ts:437-445`) — not a dry-run-local reimplementation. The thrown error propagates to `simulateNode`'s catch (`dry-run.ts:966-968`), producing the asserted `trace` entry. Re-run independently by the coordinator: `bun test src/dry-run.test.ts -t "a directive on a failed producer fails the consumer"` → `1 pass, 3 expect() calls, 0 fail`. Also ran the full split: `bun test src/state-migration.test.ts src/dry-run.test.ts` → `70 pass, 0 fail, 181 expect() calls`.

**Required outcome:** None — closed.

**Found by:** `prp-core:code-reviewer` (original finding); re-verified by `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier`, and the coordinator independently at head `279a958b`.

**Disposition:** FIXED. The test is real proof the preview path and the real executor cannot drift on this guard, not a vacuous assertion.

</details>

<details>
<summary><code>R3</code> — Docs don't state the new failed-producer <code>if_skipped</code> contract</summary>

**Impact:** `packages/docs-web/src/content/docs/guides/authoring-workflows.md:922-942` documented `if_skipped` at length for a *skipped* producer but never stated what happens when the producer *failed* — the exact ambiguity that let issue #2696 happen in the first place.

**Evidence:** `authoring-workflows.md:944-949` — added paragraph. Compared line-by-line against `resolveBindingDirective` (`dag-executor.ts:408-447`): "`if_skipped` only ever covers a producer that did not run" matches the `producer.state === 'skipped' || 'pending'` branch (line 422, the only branch reading `directive.if_skipped`); "a producer that ran and failed always fails the binding too, whether or not `if_skipped` is declared" matches the unconditional `state === 'failed'` throw (line 437, no `if_skipped` read in that branch); "a `loop_group`'s failure paths ... can leave real, non-empty output behind" matches the code's own comment (lines 433-436) almost verbatim. No overclaiming or underclaiming in the contract statement itself.

**Required outcome:** None — closed on the original ask (documenting the contract).

**Found by:** `prp-core:code-reviewer` (original finding); re-verified by `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier`, and the coordinator independently at head `279a958b`.

**Disposition:** FIXED for its original scope (the contract statement is accurate and complete). The same added paragraph also contains a separate remedy-suggestion sentence found to be unsound during this verification pass — tracked as new finding `R5` below rather than reopening `R3`, since `R3` was specifically about documenting the contract, which this addition does correctly.

</details>

<details>
<summary><code>R4</code> — New tests bypass the file's <code>runDag</code> helper</summary>

**Impact:** None on correctness. The three new `#2696` tests (`dag-executor.test.ts:25441-25594`, inside `describe('value transport (#2637)...')`) call `executeDagWorkflow(...)` directly rather than that block's local `runDag` helper (`dag-executor.test.ts:24945`), which the other five tests in the same block use.

**Evidence:** `runDag` at `dag-executor.test.ts:24945` has signature `Promise<{ events: PersistedEvent[]; store: MockWorkflowStore }>` — no path to the raw `executeDagWorkflow` return value or the `platform` mock. The "completed" test asserts `expect(result).toBe('CORRECTIONS_APPLIED')` (line ~25483), the "skipped" test asserts `expect(result).toBe('NO_CORRECTIONS_RAN')` (line ~25530), and the "failed" test asserts on `platform.sendMessage` mock calls (lines ~25588-25592) — none reachable through `{ events, store }`. Precedent: `EDGE A` (`dag-executor.test.ts:18506`) and `INSTANCE 4` (`:18456`), in a *different* describe block (`'executeDagWorkflow -- loop_group node'`) that has **no local `runDag` helper at all**, both call `executeDagWorkflow` directly for the identical reason (needing `result`/`platform`) — confirming this is a genuine pre-existing convention, not one invented to excuse the new tests.

**Required outcome:** Optional — extending `runDag` to also return `result`/`platform` (both already computed inside the helper, just discarded) would be a cheap, purely additive change that removes ~30-45 lines of duplicated positional-argument boilerplate across the three new tests. Not required for this PR; a legitimate smaller shape for future polish, not a correctness or maintenance concern today.

**Found by:** `prp-core:code-simplifier` (original finding); re-verified by `prp-core:code-simplifier` at head `279a958b`, independently confirming the `runDag` signature, the three tests' actual needs, and the `EDGE A`/`INSTANCE 4` precedent by direct source reading.

**Disposition:** DECLINED — confirmed correct. `runDag` genuinely cannot expose what these tests need, and calling `executeDagWorkflow` directly is an established pattern already present in the file before this PR.

</details>

<details>
<summary><code>R5</code> — New docs paragraph recommends a <code>when:</code> guard that cannot actually distinguish a failed producer</summary>

**Impact:** The paragraph added to close `R3` ends: *"Declaring `if_skipped` never papers over a real failure; guard the consumer with `when:` instead if it should tolerate that branch."* This is the only place in the docs this exact phrasing appears — it did not exist anywhere in `authoring-workflows.md` before this correction (`git show bed88a43:.../authoring-workflows.md | grep when:` shows no equivalent sentence). But `when:`'s grammar (`condition-evaluator.ts`, `when-atom.ts`) exposes only `$nodeId.output[.field]`, `$INPUTS.*`, and comparison operators — no producer-state accessor anywhere (`rg -n "state" packages/workflows/src/when-atom.ts` returns nothing). For a failed `loop_group`, `resolveOutputRef`'s own unfielded branch (`condition-evaluator.ts:86-91`) reads `.output` exactly the same way it does for a successful producer — real, non-empty text either way (the same gap `R1`/#2713 tracks). An author who follows this new advice literally — writing `when: "$corrections.output != ''"` or a field check to "exclude the failed branch" — gets a condition that evaluates true for the failed group exactly as it would for a successful one, since both leave non-empty output. The suggested remedy does not work for the scenario the paragraph itself just described.

**Evidence:** `packages/docs-web/src/content/docs/guides/authoring-workflows.md:948-949` (new sentence); `packages/workflows/src/condition-evaluator.ts:86-91` (the `when:` resolver, no state check); `packages/workflows/src/when-atom.ts` (grammar, no `state` identifier). Note: the identical "guard ... with a 'when:' condition" phrasing already exists, unchanged, in `resolveBindingDirective`'s own thrown-error text (`dag-executor.ts:441-443`, `:446-448`) — present since the originally reviewed head `bed88a43` and out of this correction's diff, so not itself reopened here. What is new is publishing the same unsound remedy as proactive, authoritative guidance in the authoring guide rather than as an ad hoc error-message aside.

**Required outcome:** Optional — drop the "guard the consumer with `when:` instead" clause, or replace it with an accurate statement that no `when:`-expressible guard exists yet for a failed producer, pointing at #2713 (which is the issue that would need to land before this remedy could actually work). Not required for this PR to merge; it's a documentation-accuracy nit on newly added prose, not a code defect, and does not itself cause data loss or an unrecoverable state.

**Found by:** `prp-core:seam-analyzer`, independently re-verified by the coordinator (confirmed the grammar has no state accessor, confirmed the phrasing is new to the docs at this correction, confirmed the pre-existing error-message text is unchanged and out of scope).

**Disposition:** FIXED. Correction commit `de107b5f` ("fix(docs): remove an unworkable when: remedy from the failed-producer note") removes the "guard the consumer with `when:` instead" clause and replaces it with: *"There is no way to opt a binding out of this: declaring `if_skipped` never papers over a real failure, and `when:` cannot substitute for it either — it reads the same output text, which is indistinguishable from a success for a `loop_group` that failed mid-run."* Verified the bounded diff (`279a958b..de107b5f`) touches exactly this one file, 3 insertions / 2 deletions, confined to these two sentences — no other content in the PR changed. Independently re-confirmed the new sentence's factual claim by re-reading both resolvers at the new head: `condition-evaluator.ts:86-91` (`resolveOutputRef`'s unfielded branch) still has no `producer.state`/status check, and `when-atom.ts`'s `WhenAtomRef` grammar (`{kind:'node', nodeId, field}` / `{kind:'input', name}`) still has no state accessor — so `when:` genuinely cannot distinguish a failed `loop_group`'s stale output from a successful one, exactly as the new sentence states. Grepped the full file for `guard the consumer`, `opt.*out`, and `escape hatch`; no other instance of the unsound remedy remains anywhere in the doc. Re-verified independently by `prp-core:seam-analyzer` (dispatched fresh for this correction) and the coordinator; both concluded FIXED with no correction-caused regression. The pre-existing "guard ... with a `when:` condition" phrasing in `resolveBindingDirective`'s thrown-error text (`dag-executor.ts:441-443`) remains unchanged and out of this correction's diff — noted for completeness, not reopened, since it predates the whole PR and is a distinct scope from the docs paragraph this finding was about.

</details>

<details>

#### Reviewer coverage

**Pass 2 (head `279a958b`):**

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | R2, R3 verified FIXED with direct evidence and independent test execution; no new defect found in its area |
| `seams` (`prp-core:seam-analyzer`) | R1 re-verified TRACKED FOLLOW-UP (issue #2713 still accurate, gap unchanged); found new finding R5 |
| `simplify` (`prp-core:code-simplifier`) | R4 re-verified DECLINED with stronger evidence (EDGE A/INSTANCE 4's describe block has no `runDag` at all); R1's `resolveWorkflowValue` observation reconfirmed unchanged; no new defect in the correction diff itself |

**Pass 3 (head `de107b5f`, this pass):** Only `seams` retained — the sole scope owning the one open finding (`R5`) being verified; `code` and `simplify` had no affected finding and the correction touches no code, so neither was re-dispatched.

| Scope | Result |
|---|---|
| `seams` (`prp-core:seam-analyzer`) | R5 verified FIXED — unsound `when:` remedy removed, replacement claim re-confirmed accurate against `condition-evaluator.ts`/`when-atom.ts`; no correction-caused regression found |

`types` scope not selected in either pass — the correction changes no schema, type, constructor, or public signature (docs prose only, this pass).

#### Validation

**Pass 2 (head `279a958b`):**

|---|---|---|
| `bun test src/state-migration.test.ts src/dry-run.test.ts` (from `packages/workflows/`) | PASS | `70 pass, 0 fail, 181 expect() calls` — run independently by the coordinator in the review's own detached worktree at head `279a958b` |
| `bun test src/dry-run.test.ts -t "a directive on a failed producer fails the consumer"` | PASS | `1 pass, 3 expect() calls, 0 fail` |
| `bun run type-check` (repo root) | PASS | All 12 packages report `Exited with code 0`, including `@archon/workflows` and `@archon/docs-web` |
| `bun run lint` (repo root) | PASS | `eslint . --cache` exits clean, no warnings |
| `bun run format:check` (repo root) | PASS | `prettier --check .` — "All matched files use Prettier code style!" |
| `bun run validate` (repo root, full gate) | NOT RUN (that pass) | Correction touched no bundled/generated/schema/dependency surfaces; the implementer separately reported a full pass (exit 0, 148 `bun test` invocations, 0 fail) at that head — not independently re-verified by the coordinator |

**Pass 3 (head `de107b5f`, this pass):**

|---|---|---|
| `git diff --stat 279a958b..de107b5f` (bounded correction diff) | Confirmed narrow | Exactly one file, `packages/docs-web/src/content/docs/guides/authoring-workflows.md`, 3 insertions / 2 deletions — no code, test, schema, or generated file touched |
| `npx prettier --check packages/docs-web/src/content/docs/guides/authoring-workflows.md` | PASS | "All matched files use Prettier code style!" |
| `type-check` / `lint` / tests (repo root) | NOT RE-RUN, still applicable | No code file changed since the pass-2 green results above; nothing in this correction could invalidate them |
| `bun run validate` (repo root, full gate) | NOT RUN (this pass) | Docs-only prose correction; nothing in scope for the full gate to catch that direct evidence and reviewer verification didn't already cover |

</details>

PR #2711, opened 2026-08-22
refactor(workflows): unify all five gate pauses on one persist helper

### LINKED ISSUE #2489: refactor(workflows): one suspend primitive with a reason value — the reason enum already exists, one caller bypasses the helper
"Suspend and resume" is often described as four separate mechanisms — the approval node, the write-back gate, the `child_workflow` pause, and the interactive-loop pause. That framing overstates the problem, and the correction is worth recording before anyone budgets a rewrite around it.

**The reason value already exists.** `ApprovalContext.type?: 'approval' | 'interactive_loop' | 'writeback' | 'child_workflow'` (`schemas/workflow-run.ts:203`), and three of the four call sites already share one helper, `pauseGateRespectingExternalTransition` (`dag-executor.ts:5094-5122`), which owns the persist-and-emit.

The five pause sites: `:3803` (loop_group gate) · `:5024` (loop gate) · `:5313` (approval node) · `:5503` (child pause) · `:7816` (write-back gate).

## The actual defect is ~10 lines

`:5503` is the one that **bypasses the shared helper** and calls `deps.store.pauseWorkflowRun` directly. That asymmetry is the whole of it.

## What is genuinely not unified, and should not be

The **resume** side. An approval gate is resolved externally by the approve handler; a `child_workflow` node re-runs and re-inspects its child (documented at `dag-executor.ts:5482-5490`). Those are different by design, and collapsing them would be a regression rather than a simplification.

## Proposed

```ts
type SuspendReason = 'approval' | 'loop_gate' | 'writeback' | 'child';

interface SuspendPoint {
  readonly reason: SuspendReason;
  readonly nodeId: NodeId;
  readonly message?: Template;
  readonly capture: boolean;
  /** Reason-specific payload, discriminated on `reason`. */
  readonly detail: SuspendDetail;
}
```

One pause primitive, one persist path, a discriminated `detail` — and resume entry points that stay distinct per reason.

## Also worth folding in

The engine's pause semantics only work because siblings keep streaming: `shouldContinueStreamingForStatus` returns `true` for both `running` and `paused`, with the rationale at `dag-executor.ts:442-460`. That contract should be stated on the type rather than living only in a comment.

## Acceptance

- All five pause sites go through one helper — no direct `store.pauseWorkflowRun` calls
- `SuspendReason` is exhaustively switched at every consumer
- Resume paths stay separate and are named as such in types
- No behaviour change; gate re-entry idempotency tests still pass unchanged

Size: **S–M**, and smaller than its reputation. The gate-queue limitation (#2180 — one approval slot) is a *separate* issue and prices separately.

Wirasm:
## Problem and outcome

`ApprovalContext.type` already names the four suspend reasons, and three of the five pause sites in the DAG executor already share one helper (`pauseGateRespectingExternalTransition`) that owns the persist-and-CAS-tolerance logic. Two sites bypassed it instead: the `workflow:` child-pause site (an undocumented, accidental bypass whose lost-CAS handling diverged from the other three — it threw into a `node_failed` instead of tolerating the race) and the container write-back gate (a deliberate bypass, needed for its fail-closed CAS handling and its extra `pending_writeback` metadata write — something the originating issue's framing had missed).

- **Outcome:** All five pause sites now write through one persist path. The child-pause site tolerates a lost CAS the same way the other three gate types already do; the write-back gate keeps its fail-closed throw and atomic metadata write, now expressed as explicit options on the shared helper instead of a separate direct call.
- **Invariant:** No observable behavior changes on the common (successful pause / successful resume) path for any of the five gate types. The write-back gate's fail-closed CAS handling and atomic `pending_writeback` write are preserved exactly.
- **Scope boundary:** Resume-side logic stays deliberately separate per reason (an approval gate is resolved externally by the approve/reject endpoints; a `workflow:` node re-runs and re-inspects its child) — this PR only unifies the *pause* write, not resume. The persisted `ApprovalContext.type` string values are unchanged (they're stored verbatim in paused-run rows), so `SuspendReason` is a named alias, not a rename.

## Review guidance

- **Feedback requested:** Correctness of the two behavioral deltas below, and whether the `failClosed`/`extraMetadata` helper extension is the right shape versus a second primitive.
- **Start here:** `packages/workflows/src/dag-executor.ts` — `pauseGateRespectingExternalTransition` (the extended helper), then its two new callers: `pauseParentOnChild` and `raiseWriteBackGate`.
- **Review order:** helper extension → `pauseParentOnChild` → `raiseWriteBackGate` → `SuspendReason` in `schemas/workflow-run.ts` → the `approveWorkflow`/`rejectWorkflow` switch conversions in `packages/core/src/operations/workflow-operations.ts`.
- **Lower-attention areas:** the `shouldContinueStreamingForStatus` parameter tightening (`string | null` → `WorkflowRunStatus | null`) and its doc-comment cross-reference — mechanical, verified by `tsc` against every real caller.
- **Known risk or uncertainty:** the write-back gate's `writeback_requested` lifecycle event now fires just after the `approval_pending` emit instead of just before (both are best-effort, fire-and-forget emits with no ordering contract asserted anywhere — verified by grep against the test suite).

## Solution

`pauseGateRespectingExternalTransition` gained two additive, optional parameters: `extraMetadata` (forwarded to `pauseWorkflowRun`'s existing third argument) and `failClosed` (skips the lost-CAS tolerance branch entirely, matching the write-back gate's existing zero-try/catch behavior). Both new sites route through this one function instead of calling `deps.store.pauseWorkflowRun` directly:

- `pauseParentOnChild` (the `workflow:` child-pause site) now calls the helper with default (tolerant) behavior, and gates its "Blocked on sub-run" chat notification on the helper's new return value so a lost-CAS run never gets a stale message it can no longer act on.
- `raiseWriteBackGate` now calls the helper with `failClosed: true` and the `pending_writeback` metadata as `extraMetadata` — byte-identical persisted behavior to the direct call it replaces.

`SuspendReason` (`packages/workflows/src/schemas/workflow-run.ts`) is `NonNullable<ApprovalContext['type']>` — a named alias, not a new enum, since the values are persisted into `workflow_runs.metadata.approval.type` and renaming them would break reading already-paused rows. Its doc comment names where each variant's *resume* logic lives, satisfying "resume paths stay separate and are named as such in types" without merging any resume-side code.

`approveWorkflow` and `rejectWorkflow` (the two places that genuinely dispatch different behavior per suspend reason) are now `switch` statements over the reason, following the existing exhaustiveness idiom already used once in this file (`interpret`'s `const unreachable: never = ...`). A `child_workflow` gate reaching either function is unreachable in practice (both are already redirected upstream by `assertApprovable`/`assertRejectable`) but is now a fail-loud backstop rather than an implicit fallthrough; an unrecognized future reason value throws instead of silently taking the generic approval/rework path. The ~8 other call sites that read `.type` as a single-value guard (API route redirects, `prompt-builder.ts`'s additive detail lines, etc.) were deliberately left as-is — they degrade safely (less detail shown) rather than incorrectly (wrong action taken) on a future reason, so converting them would add no invariant protection.

## Behavior change

| | Before | After |
|---|---|---|
| **`workflow:` child pause, lost CAS** (e.g. two gate-pausing nodes race in one layer, #2180) | Throws into a `node_failed` event + user-facing error message | Silently tolerated, same as the other three gate types — no failure, no stale chat message |
| **Write-back gate lifecycle event order** | `pauseWorkflowRun` → `writeback_requested` emit → `approval_pending` emit | `pauseWorkflowRun` + `approval_pending` emit (now bundled in the helper) → `writeback_requested` emit |

## Validation

- `bun --filter '@archon/workflows' type-check` — passed
- `bun --filter '@archon/core' type-check` — passed
- `bun test src/dag-executor.test.ts` (packages/workflows) — 584 pass, 0 fail — includes a new regression test proving the `workflow:` child-pause site now tolerates a lost CAS like the other three gate types
- `bun test src/subrun.test.ts` (packages/workflows) — 77 pass, 0 fail
- `bun test src/operations/workflow-operations.test.ts` (packages/core) — 51 pass, 0 fail — 49 pre-existing unchanged + 2 new tests proving an unrecognized gate type fails loudly instead of resolving via the generic path
- `bun run validate` (full repo gate: bundled/schema/capability-matrix checks, type-check, lint --max-warnings 0, format:check, full test suite) — passed clean
- **Not verified:** the `#2180` same-layer gate-slot collision itself (two gate-pausing nodes in one topological layer) is a separately tracked, priced-separately issue per the source issue — this PR only unifies how each site's lost-CAS is *handled*, not the collision's root cause.

## Links

- Closes #2489
- Plan: https://github.com/coleam00/Archon/issues/2489#issuecomment-5379817338

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow approval and rejection handling for unsupported gate types.
  * Prevented duplicate or incorrect notifications when pauses race with external state changes.
  * Ensured failed pause transitions halt cleanly without triggering misleading workflow events.
  * Improved container write-back pause handling and state persistence.
  * Improved concurrent child-workflow pause behavior when approval capacity is occupied.

* **Documentation**
  * Clarified workflow pause behavior, approval context, and resume reasons.


Wirasm:
## Ready to merge

Second correction-verification pass, bounded to the diff between the previously reviewed head
(`897687e`) and the current head (`94fd320`): one commit, one file
(`packages/workflows/src/dag-executor.test.ts`, net -27 lines), addressing only the sole open
finding from the prior pass, R9. The correction refactors R3's write-back fail-closed regression
test to call the describe block's pre-existing `runGateWithStore(store, backend, writeBack)`
helper (already used by two sibling tests before this commit) instead of duplicating a ~30-line
`executeDagWorkflow` positional-argument call — resolving the rule-of-three duplication R9 flagged.
It also drops the test's separate `platform.sendMessage` assertion; traced directly against
`raiseWriteBackGate` (`packages/workflows/src/dag-executor.ts:9414-9460`), `backend.suspend` and
`platform.sendMessage` are strictly sequential, unguarded `await`s after the pause call
(`pauseGateRespectingExternalTransition`, which rethrows immediately on `failClosed: true` —
`dag-executor.ts:6360-6380`), so proving `backend.suspend` was never called is sufficient proof
`sendMessage` never ran either — not a weakened test.

One residual: R9's original finding also asked to drop two mock reassignments
(`store.getWorkflowRunStatus`, `store.claimWriteback`) that duplicate `createMockStore()`'s own
defaults; this commit left both untouched. Verified immaterial — with `failClosed: true` the throw
happens before `getWorkflowRunStatus` is ever read, so both lines are inert as well as redundant.
Non-blocking, was always a Suggestion-tier "also" item, and the finding's substantive concern (the
duplicated 30-line setup) is resolved.

**Validation:** Independently reproduced at the current head (`94fd3203`): `bun test
src/dag-executor.test.ts` (packages/workflows) — 585 pass, 0 fail (unchanged from the prior pass;
this commit only refactors an existing test, adds none); the write-back gate `describe` block in
isolation (`-t "write-back"`) — 14 pass, 0 fail, matching the prior pass exactly; `bun x tsc
--noEmit` (packages/workflows, which does NOT exclude `*.test.ts`) — clean. `bun run type-check`
and `bun run lint` (full repo) independently confirmed clean by the dispatched reviewer.
`bun run validate` was not independently rerun this pass — no schema, migration, or non-test source
file changed, so no gate this correction could plausibly invalidate goes unrun.

**Resolved:** 8 (R1–R7, R9) · **Tracked follow-ups:** 1 —
https://github.com/coleam00/Archon/issues/2714

### Findings

|---|---|---|---|
| `R1` | Important | `--detach` approve/reject precheck can silently diverge from the real resolution outcome on an unrecognized gate type | FIXED |
| `R2` | Important | Docs-web authoring guide still documents the old "second pause fails its node" behavior this PR removes | FIXED |
| `R3` | Suggestion | Write-back gate's `failClosed: true` throw-on-lost-pause path has no direct regression test | FIXED |
| `R4` | Suggestion | `rejectWorkflow`'s `suspendReason` double-binding is more complex than needed for the same exhaustiveness guarantee | FIXED |
| `R5` | Suggestion | `SuspendReason` is a naming-only alias — it enforces nothing beyond what the inline union already enforced | NOT A FINDING |
| `R6` | Suggestion | Minor doc/type loose ends: stale `assertRejectable` docblock, missing barrel re-export, one loosely-typed local, one derivable flag | FIXED (3 of 4) / DECLINED (1 of 4) |
| `R7` | Suggestion | `rejectWorkflow`'s `case undefined` (treated as `'approval'`) disagrees with the executor's rejection-resume detector, which excludes `undefined` — legacy-only, behavior-preserving | FIXED (documentation only) |
| `R8` | Important (tracked separately) | Pre-existing, PR-unrelated: a first-node `approval:` gate rejected with `on_reject` staged can never be resumed | TRACKED FOLLOW-UP |
| `R9` | Suggestion | R3's new regression test duplicates the write-back block's `runGate` setup instead of reusing it (now the third near-identical inline call) | FIXED (2 of 3) / DECLINED (1 of 3) |

<details>
<summary><code>R1</code> — <code>assertApprovable</code>/<code>assertRejectable</code> don't validate the gate reason, so the <code>--detach</code> precheck can ack success on a run the real operation then throws on</summary>

**Impact:** `archon workflow approve <id> --detach --json` (or `reject`) can return `{ok: true, continues: true}` while the detached child throws into a log nobody reads, leaving the run paused and unresolved with no signal back to the caller.

**Evidence (original):** `packages/core/src/operations/workflow-operations.ts:185-221` / `:231-258` (`assertApprovable`/`assertRejectable` skipped the reason check); `:380-487` / `:525-589` (the new fail-loud switches this PR added weren't shared by the precheck); `packages/cli/src/commands/workflow.ts:3164-3166`/`:3292-3294` (the precheck's own comment states the guarantee this gap broke).

**Verification evidence:** `packages/workflows/src/schemas/workflow-run.ts` now exports `suspendReasonSchema` (a `z.enum(['approval','interactive_loop','writeback','child_workflow'])`) and `isRecognizedSuspendReason(type)`. `packages/core/src/operations/workflow-operations.ts:198-206` (`assertApprovable`) and `:253-260` (`assertRejectable`) both call `isRecognizedSuspendReason(approval.type)` and throw `` `Run ${run.id} has an unrecognized gate type '...'. This Archon build cannot resolve it.` `` before any other check — the same check both functions now share, and the CLI's `--detach` precheck (`packages/cli/src/commands/workflow.ts:3166`, `:3294`, unchanged) calls exactly these functions. `workflow-operations.test.ts`'s two "future/corrupted reason" tests were updated to assert the new, earlier message (`/unrecognized gate type 'bogus-reason'/`) and both pass — proving `approveWorkflow`/`rejectWorkflow` (which call `assertApprovable`/`assertRejectable` first) now throw at the shared gate, not the switch's now-unreachable backstop.

**Disposition:** FIXED — verified in diff and by passing tests (`bun test src/operations/workflow-operations.test.ts`, 51 pass).

</details>

<details>
<summary><code>R2</code> — Docs-web authoring guide states the old, removed "second pause fails its node" behavior</summary>

**Impact:** An operator reading the guide would believe a second same-layer gate pause fails its node, when after this PR it's silently dropped instead.

**Verification evidence:** `packages/docs-web/src/content/docs/guides/authoring-workflows.md:1776-1779` now reads: "the second pause is silently dropped, and that child stays unmentioned until a later resume re-pauses on it" — matching the required outcome exactly.

**Disposition:** FIXED — verified by direct diff read.

</details>

<details>
<summary><code>R3</code> — Write-back gate's fail-closed throw path has no direct regression test</summary>

**Impact:** A future edit to the shared helper's default tolerant behavior, or a dropped `failClosed: true` flag at the write-back call site, would not be caught by any existing test.

**Verification evidence:** `packages/workflows/src/dag-executor.test.ts` adds `'a lost pause CAS during write-back stays fail-closed: throws, never suspends (#2489)'` inside the write-back gate describe block: mocks `store.pauseWorkflowRun` to reject, asserts `runGateWithStore(...)` (the block's existing setup helper, as of the R9 correction) rejects with the propagated error, and asserts `backend.suspend` is never called — exactly the required outcome, with the equivalent-proof-strength dropped `sendMessage` assertion covered under R9's disposition below. Confirmed passing at both the original head (585/585) and the current head (585/585, 0 fail).

**Disposition:** FIXED — verified by diff and by re-running the test suite at both heads.

</details>

<details>
<summary><code>R4</code> — <code>rejectWorkflow</code>'s double-binding is avoidable</summary>

**Impact:** None — clarity/consistency only.

**Verification evidence:** `packages/core/src/operations/workflow-operations.ts` now reads `if (approval !== undefined) { switch (approval.type) { ... } }` — the separately-declared `suspendReason` binding and its `if (approval === undefined) throw` guard inside `case 'writeback'` are both removed; `approval` narrows for free from the outer guard. Matches the compiled alternative both original reviewers verified. The now-unused `SuspendReason` type import was removed from this file's import list (confirmed in diff), consistent with the finding's note that this was its sole application-code consumer.

**Found by:** `prp-core:type-design-analyzer`, `prp-core:code-simplifier`

**Disposition:** FIXED — verified by direct diff read; `tsc --noEmit` and the test suite both pass.

</details>

<details>
<summary><code>R5</code> — <code>SuspendReason</code> enforces nothing beyond the existing inline union</summary>

**Impact:** None — was informational against the pre-fix code.

**Verification evidence:** `SuspendReason` is now `z.infer<typeof suspendReasonSchema>`, and `ApprovalContext['type']` is typed as `SuspendReason` directly (no longer the reverse derivation `NonNullable<ApprovalContext['type']>`). `isRecognizedSuspendReason` gives the schema genuine runtime enforcement value via R1's fix. The finding's premise (alias with no independent enforcement) no longer describes the code.

**Found by:** `prp-core:type-design-analyzer`, `prp-core:code-reviewer`

**Disposition:** NOT A FINDING — superseded by R1's fix. Correctly framed as superseded rather than independently "fixed," since no change targeted R5 directly; R1's change simply removed the premise.

</details>

<details>
<summary><code>R6</code> — Minor loose ends (4 sub-items)</summary>

**Impact:** None individually load-bearing.

**Verification evidence:**
- `assertRejectable`'s docblock (`workflow-operations.ts:233-238`) now reads "...malformed... is still legitimately rejectable. A well-formed context with an unrecognized `type` is NOT one of those cases — it throws via the suspend-reason check below, same as approve." — narrowed exactly as required. **FIXED**
- `SuspendReason`, `suspendReasonSchema`, and `isRecognizedSuspendReason` are now all re-exported from `packages/workflows/src/schemas/index.ts`. **FIXED**
- `packages/workflows/src/dag-executor.ts`'s pause helper now declares `let status: WorkflowRunStatus | null;` instead of `string | null`. **FIXED**
- `options.failClosed` remains an independent flag from `type: 'writeback'`, not derived. The implementation report records an explicit rationale (CAS-tolerance policy vs. reason identity are two independent, explicit values; zero behavior difference today) — this was marked optional/non-required in the original finding, so declining with a stated reason is a valid disposition, not a gap. **DECLINED**

**Found by:** `prp-core:code-reviewer`, `prp-core:code-simplifier`, `prp-core:type-design-analyzer`

**Disposition:** FIXED (3 of 4 sub-items) / DECLINED (1 of 4, with reason) — verified by direct diff read.

</details>

<details>
<summary><code>R7</code> — <code>rejectWorkflow</code>'s <code>undefined</code>-as-<code>'approval'</code> equivalence disagrees with the executor's rejection-resume detector (legacy-only)</summary>

**Impact:** None reachable today except a run paused by a pre-#936 Archon binary and still paused — historical only. Confirmed pre-existing and behavior-preserving, not introduced by this PR.

**Verification evidence:** `packages/workflows/src/dag-executor.ts`'s rejection-resume detector (`executeApprovalNode`) now carries a comment: "Strict `=== 'approval'` excludes `undefined`: a pre-#936 run paused before the `type` field existed and rejected-with-rework would stage a rework `rejectWorkflow` treats as equivalent to `type: 'approval'` but this check won't recognize — legacy-only, pre-existing, not touched by #2489." No functional change to the check itself, matching the disposition's own framing that a functional fix was confirmed unnecessary.

**Disposition:** FIXED (documentation only) — verified by direct diff read.

</details>

<details>
<summary><code>R8</code> — Pre-existing, PR-unrelated: a first-node <code>approval:</code> gate rejected with staged rework can never be resumed</summary>

**Impact:** `hydrateResumableRun` (`packages/workflows/src/executor.ts`, untouched by this PR at every head reviewed so far) omits `'approval'` from resumable reasons with zero completed nodes, so a first-node `approval:` gate rejected with an `on_reject` rework staged can never be resumed via `archon workflow resume`.

**Verification evidence:** `packages/workflows/src/executor.ts` is not in the correction diff for either correction pass (confirmed via `git diff --stat` across both head ranges — this pass touches only `dag-executor.test.ts`), so this finding is unaffected. Issue #2714 ("bug(workflows): rejecting a first-node approval gate with on_reject stages a rework that resume can never reach") is open and correctly describes this defect.

**Found by:** `prp-core:seam-analyzer` (verified independently by the coordinator)

**Disposition:** TRACKED FOLLOW-UP — https://github.com/coleam00/Archon/issues/2714 (verified open)

</details>

<details>
<summary><code>R9</code> — New regression test duplicates the write-back block's existing setup helper instead of reusing it</summary>

**Impact:** None functional — cosmetic/maintainability only. The write-back gate `describe` block in
`dag-executor.test.ts` had a `runGate(opts)` helper and (as of this same block, pre-existing) a
second helper `runGateWithStore(store, backend, writeBack)` for tests that need to hand in a
custom-built store — used by two sibling tests before this correction. R3's original lost-pause
test inlined a third ~30-line `executeDagWorkflow` call instead of reusing either.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:22969-23003` (the test, post-correction);
`:23147-23177` (`runGateWithStore`, pre-existing, unchanged by this commit); `:23191`, `:23216`
(the two sibling tests that already called it before this correction) — confirming this is genuine
reuse of existing infrastructure, not new machinery invented for this fix.

**Required outcome:** Optional — reuse existing setup infrastructure instead of inlining a third
near-identical call, and drop the two mock reassignments (`store.getWorkflowRunStatus`,
`store.claimWriteback`) that duplicate `createMockStore()`'s own defaults.

**Found by:** `prp-core:code-simplifier` (correction-verification pass); disposition independently
re-verified by `prp-core:code-simplifier` in a second correction-verification pass against commit
`94fd3203`

**Disposition:** FIXED (2 of 3) / DECLINED (1 of 3, with reason) — the test now calls the
pre-existing `runGateWithStore` helper instead of inlining the 30-line setup (verified: diff shows
the inline `executeDagWorkflow(...)` call replaced by `runGateWithStore(store, backend, 'approve')`,
and 585/585 tests still pass including this one). The separate `platform.sendMessage` assertion was
also dropped; traced against `raiseWriteBackGate` (`dag-executor.ts:9414-9460`) — `backend.suspend`
and `sendMessage` are strictly sequential unguarded `await`s after the pause call, which rethrows
immediately on `failClosed: true` with nothing else executing, so `suspend`-not-called already
proves `sendMessage`-not-called; equivalent proof strength, not a weakened test. The two redundant
mock reassignments (`getWorkflowRunStatus`, `claimWriteback`) were left untouched by this commit —
verified immaterial: both are value-identical to `createMockStore()`'s own defaults
(`dag-executor.test.ts:168`, `:174`), and on this exact fail-closed path `getWorkflowRunStatus` is
never even read (the throw in `pauseGateRespectingExternalTransition` with `failClosed: true`
happens before that line, `dag-executor.ts:6369-6370`) — dead and redundant, but harmless, and
declining to chase a 2-line cosmetic residual on an already-Suggestion-tier finding does not leave
the PR's outcome or invariant unsatisfied.

</details>

<details>

#### Reviewer coverage

First `--verify-corrections` pass (`8479710`..`897687e`), four reviewer agents dispatched against
the bounded diff:

| Scope | Agent | Result |
|---|---|---|
| `code` | `prp-core:code-reviewer` | R1, R5, R6 — all CONFIRMED, no new findings |
| `seams` | `prp-core:seam-analyzer` | R1, R7, R8 — all CONFIRMED, no new findings |
| `simplify` | `prp-core:code-simplifier` | R2, R3, R4, R6 — all CONFIRMED; new finding R9 |
| `types` | `prp-core:type-design-analyzer` | R4, R5, R6 — all CONFIRMED, no new findings |

Second `--verify-corrections` pass (`897687e`..`94fd320`), bounded to R9's owning scope only — no
other scope's findings were touched by this single-file, single-finding correction:

| Scope | Agent | Result |
|---|---|---|
| `simplify` | `prp-core:code-simplifier` | R9 — helper-reuse and dropped-assertion sub-claims independently traced and CONFIRMED against `dag-executor.ts` control flow; one residual (redundant mock reassignments) confirmed NOT addressed but verified immaterial; no correction-caused regression, no new finding |

#### Validation

First correction-verification pass (at `897687eb`):

|---|---|---|
| `bun x tsc --noEmit` (packages/workflows) | PASS | exit 0 |
| `bun x tsc --noEmit` (packages/core) | PASS | exit 0 |
| `bun test src/dag-executor.test.ts` (packages/workflows) | PASS | 585 pass, 0 fail (was 584 at prior head — includes new R3 test) |
| `bun test src/subrun.test.ts` (packages/workflows) | PASS | 77 pass, 0 fail |
| `bun test src/operations/workflow-operations.test.ts` (packages/core) | PASS | 51 pass, 0 fail (R1 message-update tests included) |
| `bun x eslint <4 changed non-test source files> --max-warnings 0` | PASS | 0 errors, 0 warnings |
| `grep -n "deps.store.pauseWorkflowRun" packages/workflows/src/dag-executor.ts` (AC1) | PASS | exactly 1 match (unchanged) |
| `grep -n "assertApprovable\|assertRejectable" packages/cli/src/commands/workflow.ts` | PASS | CLI precheck still calls both shared-gate functions unchanged |
| `bun run validate` (full repo gate) | NOT RUN (independently) | Implementation report states it was rerun clean at that head; not independently rerun that pass |
| `bun run type-check` / `bun run lint` (full repo) | PASS | Independently confirmed clean repo-wide by `prp-core:code-reviewer` |
| `bun test src/dag-executor.test.ts -t "write-back"` | PASS | 14/14 pass, including the new R3 test |

Second correction-verification pass (at `94fd3203`, this pass):

|---|---|---|
| `bun test src/dag-executor.test.ts` (packages/workflows) | PASS | 585 pass, 0 fail — unchanged count from the prior pass (this commit refactors an existing test, adds none) |
| `bun test src/dag-executor.test.ts -t "write-back"` | PASS | 14 pass, 0 fail — matches the prior pass exactly |
| `bun x tsc --noEmit` (packages/workflows) | PASS | exit 0 |
| `bun run type-check` / `bun run lint` (full repo) | PASS | Independently confirmed clean repo-wide by `prp-core:code-simplifier` |
| `git diff --stat 897687eb..94fd3203` | PASS | exactly 1 file changed (`dag-executor.test.ts`), 7 insertions / 34 deletions — matches the correction's claimed scope |
| `bun run validate` (full repo gate) | NOT RUN | No schema, migration, or non-test source file changed this pass; no authoritative gate this correction could plausibly invalidate goes unrun given the targeted checks above |

</details>
</content>

PR #2712, opened 2026-08-22
fix(server): HTTP approve/reject/resume now advance CLI-launched runs

### LINKED ISSUE #2008: HTTP approve/reject cannot resume a CLI-launched paused run: re-dispatches /workflow run to a (missing) web parent instead of resuming in place

Wirasm:
## Problem and outcome

`POST /api/workflows/runs/{runId}/approve` (and `/reject`, `/resume`) can only advance a paused run by dispatching a chat message into the run's parent **web** conversation. A run launched with `bun run cli workflow run ...` has no parent conversation at all, so the operator's decision was recorded in the database but the run never advanced — silently stranded until someone ran the equivalent CLI command by hand. The CLI's own `workflow approve/reject/resume <run-id>` resumed the same run correctly, because it drives the executor directly instead of dispatching a chat message.

- **Outcome:** Approving, rejecting, or resuming a CLI-launched run over HTTP now actually resumes it — the post-gate nodes execute, the same way the CLI already does.
- **Invariant:** A gate decision recorded via any entry point reaches the paused run and lets it continue, independent of whether a live chat conversation exists to notify.
- **Scope boundary:** The existing web-parent chat-dispatch path is unchanged, and the deliberate skip for a non-web parent (Slack/Telegram/GitHub/Discord — so that thread keeps owning its own re-run flow) is untouched. Only the case with no parent conversation at all gets the new fallback.
- **Root cause:** `tryAutoResumeAfterGate` (approve/reject) and the `/resume` route's own duplicate check both gate their only execution trigger on `run.parent_conversation_id` being set and pointing at a `platform_type: 'web'` conversation. A CLI-launched run's `parent_conversation_id` is `NULL` (the CLI never sets it — it doesn't need a conversation to resume), so every HTTP endpoint returned early instead of executing anything.

## Review guidance

- **Feedback requested:** Correctness of the fallback boundary (does it correctly apply only when there's truly no parent conversation to dispatch through?) and whether the new headless platform adapter is the right shape for `@archon/server`.
- **Start here:** `packages/server/src/routes/api.ts` — `resumeRunHeadless` (new function, just above `tryAutoResumeAfterGate`) is the load-bearing change; everything else wires it in.
- **Review order:** `packages/server/src/routes/api.ts` (`resumeRunHeadless`, then the two call sites) → `packages/server/src/adapters/headless.ts` → `packages/core/src/handlers/command-handler.ts` (one-line export) → `packages/server/src/routes/api.workflow-runs.test.ts`.
- **Lower-attention areas:** `packages/server/package.json` only registers the new test file in the existing per-file `bun test` chain.
- **Known risk or uncertainty:** None outstanding — see Validation for how the fix was proven at the routing boundary it actually changes.

## Solution

Both broken entry points fall back to a new, conversation-agnostic execution path instead of skipping/400ing when `run.parent_conversation_id` is `null`:

- `resolveRunContinuation` (`packages/core/src/handlers/command-handler.ts`) is now exported — it's the exact same continuation-resolution function already shared by `/workflow resume`, `/workflow approve`, and `/workflow reject` in chat, so reusing it avoids re-implementing frozen-source resolution, container-run rejection, and live-discovery fallback a second time.
- `resumeRunHeadless` (new, `packages/server/src/routes/api.ts`) resolves the run's workflow via `resolveRunContinuation`, hydrates it via `hydrateResumableRun`, and executes it via `executeWorkflow` (`@archon/workflows/executor`) — the same primitives the CLI's own resume already uses, fire-and-forget so the HTTP response returns immediately (matching every other resume call site).
- `executeWorkflow` needs a platform to send messages through, but there's no live chat surface for this case. `HeadlessPlatform` (new, `packages/server/src/adapters/headless.ts`) is a minimal `IWorkflowPlatform` implementation — persists messages for history, no live transport — mirroring `@archon/cli`'s `CLIAdapter`, which `@archon/server` can't import directly (wrong dependency direction).
- Both `tryAutoResumeAfterGate` (approve + reject) and the `/resume` route call `resumeRunHeadless` only in the `!run.parent_conversation_id` branch; the web-parent dispatch path and the non-web-parent skip are untouched.

This keeps the fix to "wire an already-shared execution path into two more entry points" — no new DB columns, no new run states, no change to the chat-dispatch behavior.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Approving/rejecting/resuming a CLI-launched paused run returns 200 (approve/reject) or 400 (resume), but the run never advances past the gate. | The run's post-gate node(s) actually execute; the response reflects real resumption ("Resuming workflow...", "Running on-reject prompt..."). |
| **Failure behavior** | Silent stranding — decision recorded, nothing else happens, no error surfaced. | If the workflow source can't be resolved (e.g. deleted), the endpoint safely degrades to the existing CLI-hint response instead of a silent no-op or a 500. |

## Validation

- `bun run --filter '@archon/server' type-check` — passed.
- `bun test src/routes/api.workflow-runs.test.ts` (packages/server) — 109 pass, 0 fail. Updated the tests that previously asserted the old "skip" behavior for a `parent_conversation_id: null` run (on approve, reject, and resume) to instead assert `hydrateResumableRun`/`executeWorkflow` are invoked and the run resumes; kept the web-parent-dispatch and non-web-parent-skip tests passing unchanged; added a safe-degrade test per endpoint.
- `bun test src/adapters/headless.test.ts` (packages/server) — 4 pass, 0 fail — covers `HeadlessPlatform`'s message persistence and its failure-swallowing behavior.
- `bun run --filter '@archon/server' test` (full package) — all sub-suites 0 fail.
- `bun run validate` (bundled/schema checks, type-check all packages, lint `--max-warnings 0`, format check, full repo test suite) — passed, exit code 0.
- **Not verified:** No manual end-to-end run against a live server (`bun run cli workflow run ... --no-worktree` then `curl POST .../approve`) — the automated coverage above exercises the same code path the manual repro in the issue describes, and the underlying resume/execution machinery (`hydrateResumableRun`, `executeWorkflow`, container-run guard) is already covered end-to-end by `packages/workflows/src/executor.test.ts`.

## Links

- Closes #2008
- Plan: https://github.com/coleam00/Archon/issues/2008#issuecomment-5379818274

## Summary by CodeRabbit

* **New Features**
  * Workflows started without a conversation can now resume and complete in headless mode.
  * Approval, rejection, and direct resume actions support headless execution.
  * Headless runs persist assistant messages and relevant workflow metadata.

* **Bug Fixes**
  * Failed message persistence no longer interrupts workflow execution.
  * Stranded runs are marked as failed when continuation cannot proceed.

* **Tests**
  * Added coverage for headless execution, resumption, metadata persistence, and failure handling.


Wirasm:
## Ready to merge

Correction-verification pass against commit `c0a4771e`, the range `3a29d46b..c0a4771e` (previous reviewed head → current head), bounded to `packages/server/src/routes/api.ts` and `packages/server/src/routes/api.workflow-runs.test.ts`. Both blocking findings from the prior review are genuinely fixed, not just claimed: `resumeRunHeadless` now builds `resolveChildIsolation` from the already-fetched codebase using the exact field mapping as its three sibling construction sites (`orchestrator-agent.ts:762-772`, `orchestrator.ts:376-384`, `cli/workflow.ts:2144-2153`) and forwards it plus `baseBranch` into `executeWorkflow` (R1), and the entire function body is now wrapped in one outer try/catch that returns `false` on any unexpected error instead of letting it escape into the route handler's outer catch as a misleading 500 after the gate decision was already committed (R2). The coordinator proved both corrections are load-bearing, not cosmetic, by reverting each one in isolation on top of the new test file and confirming the exact new test fails without it, then restoring the fix — all three independent reviewers (code, seams, simplify) reproduced the same result. The one suggestion-scoped finding from the prior review (R3, the test mock factory not stubbing `createWorkflowDeps`) is also fixed, and now additionally stubs the newly-imported `createChildWorktreeResolver`. No new Critical or Important defect was introduced by the correction.

**Validation:** `bun run --filter '@archon/server' type-check` and `bun run --filter '@archon/core' type-check` — both pass. `bun test src/routes/api.workflow-runs.test.ts` (packages/server, 111 pass, up from 109 — the 2 new tests) and `bun test src/adapters/headless.test.ts` (4 pass) — both pass. Full `@archon/server` and `@archon/core` package test suites — both pass, 0 fail. ESLint on the two changed source files — 0 warnings. `bun run validate` (full repo gate) — see the collapsed section below.

**Resolved:** R1, R2, R3

### Findings

|---|---|---|---|
| `R1` | Important | Headless resume drops `resolveChildIsolation`, breaking downstream `isolation: worktree` sub-runs | FIXED |
| `R2` | Important | Unguarded steps in `resumeRunHeadless` can turn a transient error into a misleading 500 after the decision was already committed | FIXED |
| `R3` | Suggestion | `api.workflow-runs.test.ts`'s `@archon/core` mock factory doesn't stub the newly-imported `createWorkflowDeps` | FIXED |
| `R4` | Suggestion | Headless resumes report `workflowSource: undefined` in telemetry | DECLINED |
| `R5` | Suggestion | No unit test exercises `resumeRunHeadless`'s own safe-degrade branches directly | DECLINED |
| `R6` | Suggestion | `resolveChildIsolation` construction is now duplicated, byte-for-byte, across 3-4 call sites | DECLINED |

<details>
<summary><code>R1</code> — Headless resume drops <code>resolveChildIsolation</code>, breaking downstream <code>isolation: worktree</code> sub-runs</summary>

**Impact (as originally found):** A CLI-launched run on a git-repo codebase, paused at a gate that precedes a not-yet-executed `workflow:` node declaring `isolation: worktree`, would fail when resumed through the new HTTP endpoints — the exact capability this PR advertises.

**Evidence of fix:** `packages/server/src/routes/api.ts:2421-2429` now builds `resolveChildIsolation` from the already-fetched `codebase`:
```ts
const resolveChildIsolation =
  codebase && codebase.kind !== 'folder'
    ? createChildWorktreeResolver({
        codebaseId: codebase.id,
        codebaseName: codebase.name,
        canonicalRepoPath: codebase.default_cwd,
        baseBranch: codebase.default_branch?.trim() || undefined,
        createdByPlatform: platform.getPlatformType(),
        createdByUserId: gateActorUserId,
      })
    : undefined;
```
Both `resolveChildIsolation` and `baseBranch` are forwarded into `executeWorkflow`'s options at `:2434-2436`, ahead of the `...hydrated` spread (no key collision — `hydrateResumableRun`'s return shape has no overlapping fields). All three reviewers independently confirmed this is field-for-field identical to the three sibling construction sites (`orchestrator-agent.ts:762-772`, `orchestrator.ts:376-384`, `cli/workflow.ts:2144-2153`) against the shared `ChildWorktreeResolverConfig` contract (`packages/core/src/workflows/child-isolation-resolver.ts:100-121`). `resumeRunHeadless` has exactly two callers (`tryAutoResumeAfterGate` and the `/resume` route) and both inherit the fix automatically — no un-migrated call site.

**Test verification:** the coordinator reverted just this construction/forwarding on top of the new test file (via `git show <old-head>:packages/server/src/routes/api.ts` swapped in temporarily) and re-ran `bun test src/routes/api.workflow-runs.test.ts` — the new assertion `expect(mockCreateChildWorktreeResolver).toHaveBeenCalledWith(...)` failed exactly as expected ("was not called"), then passed again once the fix was restored. The code-reviewer and code-simplifier agents independently reproduced the same revert-and-rerun result.

**Disposition:** FIXED — independently verified by the coordinator and all three re-dispatched reviewers by reading the corrected code and by proving the regression test fails without the fix.

</details>

<details>
<summary><code>R2</code> — Unguarded steps in <code>resumeRunHeadless</code> can turn a transient error into a misleading 500 after the decision was already committed</summary>

**Impact (as originally found):** A transient error (e.g. a DB hiccup) during `resumeRunHeadless`'s setup could propagate out of `tryAutoResumeAfterGate` into the route handler's outer catch, producing `500 Failed to approve/reject workflow run` — even though `approveWorkflow`/`rejectWorkflow` had already durably recorded the decision moments earlier.

**Evidence of fix:** `packages/server/src/routes/api.ts:2379-2465` now wraps the entire function body in one outer `try { … } catch (err) { getLog().warn(...); return false; }`. The nested `hydrateResumableRun` catch still rethrows any error that isn't `WorkflowNotResumableError`, but that rethrow is now caught by the new outer catch instead of escaping the function. `executeWorkflow` itself remains fire-and-forget with its own internal `.catch()`, so the outer try/catch correctly does not swallow its async failures — only synchronous throws and awaited rejections during setup, which is the intended scope.

**Test verification:** the coordinator reverted just the outer try/catch on top of the new test file and re-ran the suite — the new test (`approve: falls back to the CLI-hint response (not a 500) when headless resume hits an unexpected error`) got `500` instead of the expected `200` exactly as predicted, then passed again once restored. Confirmed independently by the code-reviewer agent via the same technique.

**Disposition:** FIXED — verified by reading the corrected code, confirming the previously-unguarded `codebaseDb.getCodebase` call and the `hydrateResumableRun` rethrow are both now inside the outer try, and by proving the regression test fails without the fix.

</details>

<details>
<summary><code>R3</code> — <code>api.workflow-runs.test.ts</code>'s <code>@archon/core</code> mock factory doesn't stub the newly-imported <code>createWorkflowDeps</code></summary>

**Impact (as originally found):** `createWorkflowDeps` ran with its real implementation inside a fully-mocked test file — harmless today, but the exact "silently-real dependency sitting inert" shape that caused #2240.

**Evidence of fix:** the `mock.module('@archon/core', () => ({...}))` factory in `api.workflow-runs.test.ts` now includes `createWorkflowDeps: mock(() => ({ store: {} }))`, and additionally stubs `createChildWorktreeResolver: mockCreateChildWorktreeResolver` — the second export this same correction newly pulled into `api.ts`, closing the identical gap before it could open. A comment above the factory cites #2240 directly.

**Disposition:** FIXED — confirmed by reading the updated factory; `bun test src/routes/api.workflow-runs.test.ts` passes 111/111 with both new exports stubbed.

</details>

<details>
<summary><code>R4</code> — Headless resumes report <code>workflowSource: undefined</code> in telemetry</summary>

**Disposition:** DECLINED (unchanged from the prior review) — cosmetic telemetry gap only, no functional consequence; fixing it properly costs a discovery lookup this PR doesn't otherwise need. Out of scope for this correction pass (no reviewer re-raised it).

</details>

<details>
<summary><code>R5</code> — No unit test exercises <code>resumeRunHeadless</code>'s own safe-degrade branches directly</summary>

**Disposition:** DECLINED (unchanged from the prior review) — non-blocking coverage gap on simple, low-risk branches; not required by this PR's outcome or invariant. Out of scope for this correction pass (no reviewer re-raised it).

</details>

<details>
<summary><code>R6</code> — <code>resolveChildIsolation</code> construction is now duplicated, byte-for-byte, across 3-4 call sites</summary>

**Impact:** Before this correction, the ~9-line `codebase.kind !== 'folder' ? createChildWorktreeResolver({...}) : undefined` pattern already existed at 2-3 sites (`orchestrator.ts:376-384`, `orchestrator-agent.ts:762-772`, `cli/workflow.ts:2144-2153`). R1's required fix — correctly — directed mirroring that exact pattern into `resumeRunHeadless`, which now makes it a 3rd/4th copy with no shared helper and no compiler check keeping them in sync. A future field added to `ChildWorktreeResolverConfig`, or a change to the folder-project guard, has multiple copies to update by hand.

**Evidence:** `packages/server/src/routes/api.ts:2421-2429` (new copy) vs. `packages/core/src/orchestrator/orchestrator-agent.ts:762-772`, `packages/core/src/orchestrator/orchestrator.ts:376-384`, `packages/cli/src/commands/workflow.ts:2144-2153`.

**Required outcome:** Optional — a small helper (e.g. `buildChildIsolationResolver(codebase, { platform, userId })`) colocated with `createChildWorktreeResolver` in `packages/core/src/workflows/child-isolation-resolver.ts` would collapse all four sites to one line each. Worth attaching to whatever next touches this seam; not required for this PR.

**Disposition:** DECLINED — this correction did exactly what R1 required (replicate the established sibling pattern exactly); consolidating now-duplicated construction sites into a shared helper is separate, proportionate follow-up work, not required by this PR's outcome or invariant.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | `prp-core:code-reviewer` → R1 FIXED, R2 FIXED, no new findings |
| `seams` | `prp-core:seam-analyzer` → R1 FIXED, R6 (new, Suggestion) |
| `simplify` | `prp-core:code-simplifier` → R1 FIXED, R3 FIXED, R6 (new, Suggestion) |

#### Validation

|---|---|---|
| `bun run --filter '@archon/server' type-check` | PASS | Exit code 0 |
| `bun run --filter '@archon/core' type-check` | PASS | Exit code 0 |
| `bun test src/routes/api.workflow-runs.test.ts` (packages/server) | PASS | 111 pass, 0 fail, 284 expect() calls (up from 109 pass / 276 expect() pre-correction) |
| `bun test src/adapters/headless.test.ts` (packages/server) | PASS | 4 pass, 0 fail |
| `bun run --filter '@archon/server' test` (full package) | PASS | All sub-suites 0 fail |
| `bun run --filter '@archon/core' test` (full package) | PASS | All sub-suites 0 fail |
| `bun x eslint packages/server/src/adapters/headless.ts packages/server/src/routes/api.ts` | PASS | 0 errors, 0 warnings |
| `bun run validate` | PASS | Exit code 0 (background task confirmed); bundled/schema/pi-vendor-map/capability-matrix checks, all-package type-check, `eslint --max-warnings 0`, `prettier --check`, `test-install.sh`, and the full parallel repo test suite (149 zero-fail blocks, no nonzero-fail line anywhere in the 1455-line log) all passed |

</details>

PR #2716, opened 2026-08-22
refactor(workflows): Node as a discriminated union by body kind

### LINKED ISSUE #2486: refactor(workflows): Node as a discriminated union by body kind — ten node types become seven
`DagNode` is a flat schema — every field for every node kind in one object — sorted out afterwards by `superRefine`, then transformed into a union of `z.infer<…> & { other?: never }` intersections. The file's own docblock (`schemas/dag-node.ts:4-11`) says why: YAML nodes carry no explicit `type` discriminant.

That is a good reason for the **authored** shape and a bad one for the **resolved** shape.

## Proposed

Parse flat, emit a discriminated union. Ten node types become seven body kinds.

```ts
interface AgentBody { kind: 'agent'; prompt: Template; model?: ModelRef; output?: OutputSpec;
                      tools?: ToolPolicy; agentOptions?: Record<string, unknown>;
                      session?: SessionPolicy; retry?: RetryPolicy }
interface ExecBody  { kind: 'exec';  source: Template; runtime: 'sh' | 'bun' | 'uv';
                      deps?: string[]; timeoutMs?: number; retry?: RetryPolicy }
interface GateBody  { kind: 'gate';  message: Template; decisions: DecisionOption[] }
interface HaltBody  { kind: 'halt';  reason: Template }
interface LoopBody      { kind: 'loop';       prompt: Template; until: UntilSpec }
interface LoopGroupBody { kind: 'loop_group'; nodes: Node[]; until: UntilSpec }
interface WorkflowBody  { kind: 'workflow';   target: WorkflowRef; fanOut?: FanOutSpec }

type Body = AgentBody | ExecBody | GateBody | HaltBody | LoopBody | LoopGroupBody | WorkflowBody;

interface Node { id: NodeId; needs: NodeId[]; when?: Condition; join: JoinRule;
                 body: Body; outputType?: string; alwaysRun: boolean; description?: string }
```

`GateBody` reflects #2707's settled pause-primitive design: declared decisions (default `approve`/`reject`) plus structured `{decision, text}` output, not the `capture`/`onReject` shape the current `approval:` node has today — those retire under #2707 and should not be carried into the union.

`loop`, `loop_group`, and `workflow` stay three separate kinds, not one unified `SubgraphBody`/`RepeatSpec`/`Graph` shape. An earlier sketch of this issue folded them together; that was lifted from a clean-slate design doc, not this in-place reshape's own costing. The sibling issue that would have built that unification (#2492) was closed `NOT_PLANNED` — its own closing note: *"The proposed RepeatSpec/SubgraphBody shape and no-shared-executor prescription are intentionally not carried forward because they belong to the earlier composition model rather than current product acceptance."* The three modes don't share an execution substrate today and aren't going to start here — reshape each in place, individually, matching the Acceptance criteria below (which never required a unified shape).

The collapse is evidence-backed, not aesthetic:

- **`command:` + `prompt:` → `agent`.** Both already land in the same dispatch (`dag-executor.ts:1374-1432`); the only divergence is `command:` reading its text from a file first (`:1346-1348`).
- **`bash:` + `script:` → `exec`.** 327 and 354 lines of near-duplicate subprocess handling (`:2437-2763`, `:2764-3117`) differing by an interpreter and a dependency-install step — both expressible as fields. `bash:` becomes `runtime: 'sh'`.
- **`include:` leaves the union entirely.** It has no execution surface, the executor throws if one ever reaches it, and treating it as a node kind is what forces `INCLUDE_NODE_IGNORED_FIELDS` to enumerate every execution field as "ignored". It is a graph-construction directive, sibling to `nodes:`.

## What it fixes

**#2478 becomes a compile error.** `KNOWN_DAG_NODE_KEYS` is derived from the whole flat schema (`:729-731`), so `with:` on a `command:` node passes the unknown-key check and is then dropped by the transform in silence. Under a discriminated union, `with` is simply not a member of `AgentBody`.

## Blast radius — measured, and smaller than it looks

~135 narrowing call sites, and **the compiler finds every one of them.**

The frontend is **provably** unaffected: `.openapi()` describes the *input* shape. The generator unwraps `ZodEffects._def.schema` and `ZodPipeline._def.in`, and never reads `_def.out` (`@asteasolutions/zod-to-openapi@7.3.4`, `dist/index.mjs`). `DagNode` in `api.generated.d.ts` is the flat authored object, and stays that way. No spec change, no `bun generate:types` churn.

**The YAML surface does not change at all.**

## Acceptance

- `Body` is a discriminated union on `kind`; an exhaustive `switch` with a `never` check compiles
- The authored (input) schema and the generated OpenAPI `DagNode` are byte-identical to today
- `with:` on an agent-bodied node is a type error, closing #2478
- `include:` is out of the node union

Size: **M** (signature changes across untyped test call sites). Do this one first — every other primitive gets cheaper after it.

Wirasm:
## Problem and outcome

`DagNode` carried every node mode's fields as optional siblings on one flat type — `prompt`, `bash`, `script`, `loop`, `loop_group`, `approval`, `cancel`, `include`, `workflow`, `command` all lived on the same object. A bare `'prompt' in node ? node.prompt : ...` guard compiled cleanly no matter which mode the node actually was, so a reader (or a refactor) couldn't tell which shape was in hand without re-deriving the mutual-exclusivity rule the parser already enforced elsewhere.

- **Outcome:** `DagNode` is now a discriminated union on `kind`. Ten prior modes collapse into seven kinds: `command`/`prompt` merge into `agent` (with a `source: {kind:'inline'|'command', ...}` field), `bash`/`script` merge into `exec` (distinguished by `runtime`), `approval` becomes `gate` (declaring `decisions: DecisionOption[]`, per #2707's settled pause-primitive design), and `cancel` becomes `halt`. `loop`, `loop_group`, and `workflow` keep their names and gain the `kind` tag. `include:` is no longer a `DagNode` member at all — it's a separate `IncludeDirective` type, since it has no execution surface and is fully expanded away before the executor ever sees a `DagNode[]`.
- **Invariant:** the authored/OpenAPI-facing flat schema (`dagNodeFlatSchema`, `dagNodeBaseSchema`, `KNOWN_DAG_NODE_KEYS`) is unchanged — YAML authors keep writing `bash:`, `prompt:`, `approval:` exactly as before. Only the resolved, in-process TypeScript shape changed.
- **Scope boundary:** no DB schema, YAML surface, or `ApprovalContext`/pause-machinery change. This is a type-system refactor of the resolved node shape and every call site that narrows on it.

## Review guidance

- **Feedback requested:** correctness of the ~150 narrowing call sites across the engine, and whether the `agent`/`exec` collapse (rather than keeping `command`/`prompt`/`bash`/`script` as four kinds) is the right cut.
- **Start here:** `packages/workflows/src/schemas/dag-node.ts` — the union definition, the `dagNodeSchema` transform (flat → resolved), and the new guard functions (`isAgentNode`, `isExecNode`, `isGateNode`, `isHaltNode`, `isIncludeDirective`).
- **Review order:** `schemas/dag-node.ts` → `loader.ts` (parse dispatch) → `include-expander.ts` (composition) → `dag-executor.ts` (execution dispatch) → the rest are mechanical follow-through.
- **Lower-attention areas:** test-fixture construction changes (renaming `{bash: X}` to `{kind:'exec', runtime:'sh', script:X}` etc. across the `*.test.ts` files) are mechanical and covered by the full green test suite below.
- **Known risk or uncertainty:** `GateBody`'s exact shape (`decisions: DecisionOption[]` vs. an alternative narrower design) and whether `loop`/`loop_group`/`workflow` should also collapse toward a shared `SubgraphBody` shape were both flagged as open architectural questions in the plan (published on the issue). No response was received before implementation; I proceeded with the plan's documented recommendation (the `GateBody` shape the issue itself already sketched, and no `SubgraphBody` unification — three kinds stay distinct since they have materially different execution semantics).

## Solution

`packages/workflows/src/schemas/dag-node.ts` defines the new resolved types (`AgentNode`, `ExecNode`, `GateNode`, `HaltNode`, updated `LoopNode`/`LoopGroupNode`/`WorkflowNode`, and `IncludeDirective`) as real Zod schemas (`z.infer`, not hand-written interfaces, per this repo's schema convention) and a `DagNode` union of the first six. The pre-transform flat schema (`dagNodeFlatSchema`) is untouched; `dagNodeSchema`'s `.transform()` is what now maps `{bash: X}` → `{kind:'exec', runtime:'sh', script:X}`, `{approval:{message, on_reject}}` → `{kind:'gate', message, decisions:[{id:'approve'},{id:'reject', rework?}], captureResponse}`, etc.

Every production call site that used to do `'bash' in node`, `node.command`, or similar now uses the typed guards (`isAgentNode`, `isExecNode`, `isGateNode`, `isHaltNode`, `isLoopNode`, `isLoopGroupNode`, `isWorkflowNode`, `isIncludeDirective`) and reads the new field names. Two deliberate simplifications: `executeBashNode`/`executeScriptNode` stayed as separate functions rather than merging behind `isExecNode` (the two have enough non-trivial differences beyond interpreter selection that merging risked subtle regressions across 680+ combined lines — tracked as a named follow-up decision in #2718, not left as silent scope); and wherever code operates on data that is provably include-free by construction (post-`expandWorkflowIncludes`, or inside the executor which throws defensively on any surviving `IncludeDirective`) but whose *static* type still admits `IncludeDirective`, an explicit, commented `as DagNode[]` cast is used rather than threading a narrower type through every caller.

A genuine bug surfaced and fixed along the way: `loader.ts`'s `with:`-support check collapsed `isExecNode(node)` without also checking `node.runtime !== 'sh'`, silently suppressing the "`with:` is ignored here" warning on a `bash:` node (the old code correctly used `isScriptNode`, which no longer exists post-collapse). Also fixed: `validator.ts`'s command-file-exists check used a bare `'command' in node` guard, which — since `AgentNode` no longer has a top-level `command` field — silently went dead, skipping command-name validation and missing-file detection entirely; this is now `isAgentNode(node) && node.source.kind === 'command'` reading `node.source.name`.

## Validation

- `bun run validate` (full monorepo: bundled-defaults/schema/skill/capability-matrix checks, type-check, lint --max-warnings 0, format:check, test:install, full test suite across every package) — passed, exit 0.
- `bun x tsc --noEmit` for every package (`paths`, `git`, `providers`, `isolation`, `web`, `workflows`, `core`, `adapters`, `server`, `cli`, `docs-web`) — 0 errors.
- `@archon/workflows` package test suite (`bun run test`, all `bun test` splits) — every file green, including `dag-executor.test.ts` (595 tests), `loader.test.ts` (315 tests), `include-expander.test.ts` (102 tests), `schemas.test.ts`, `validator.test.ts` (89 tests, including the two behavioral fixes above).
- Full monorepo `bun run test` — every package green (`core`, `adapters`, `server`, `cli`, `providers`, `git`, `paths`, `web`, `isolation`, `workflows`).
- AC2 (authored/OpenAPI schema unchanged): `git diff origin/dev -- packages/web/src/lib/api.generated.d.ts` is empty — the generated OpenAPI-derived frontend types have zero drift, confirming the flat/authored `DagNode` schema surface is byte-identical.
- **Not verified:** no live end-to-end run against a real provider (Claude/Codex) was exercised manually — coverage is via the existing mocked `dag-executor.test.ts` integration suite, which exercises the full `executeDagWorkflow` dispatch path (including `agent`/`exec`/`gate`/`halt`/`loop`/`loop_group`/`workflow` node kinds, include expansion, resume, and structured-output binding) against a mocked provider.

## Links

- Closes #2486
- Related #2718

## Summary by CodeRabbit

- **New Features**
  - Added support for composed workflows and nested includes across execution, validation, discovery, and dry runs.
  - Improved handling of agent commands, shell scripts, inline code, gates, halts, loops, bindings, and structured outputs.

- **Bug Fixes**
  - Improved reliability for retries, approvals, fan-out, nested loops, output references, and failed or skipped steps.
  - Preserved compatibility with legacy workflow syntax.

- **Tests**
  - Expanded coverage for composition, validation, execution, resource handling, and dry-run scenarios.


Wirasm:
## Ready to merge

Correction verification against commit `d1b8621b` (the single commit added on top of the reviewed head `afc14cd1a72b39c140c17bb776612a1fa0ad3211`). Both Important findings are genuinely fixed: `runLayers`' node dispatch is now a real `switch (node.kind)` with a `never`-typed `default`, delivering the PR's own AC1 (compiler-enforced exhaustiveness); and `loader.ts`'s node-filter predicate now asserts the type it actually produces (`DagNode | IncludeDirective`), with `isIncludeDirective` guards added to the three downstream loops that previously trusted the false narrowing. The two accepted Suggestion fixes are also confirmed correct: the migrated `loop_group`/`loop` mutual-exclusivity test fixture is reverted to the raw shape and now asserts on the actual error message (reproduced directly — `issues[0].message` now contains `"mutually exclusive"`, no longer confounded by the unrelated malformed-body-node error), and the unreachable `''` fallback in `executeNodeInternal` is replaced with an explicit throw matching the file's own established convention. The correction diff is scoped to exactly the three files the implementer named (`dag-executor.ts`, `loader.ts`, `schemas.test.ts`) — no changes touch the R4/R5/R6/R8 sites, so those DECLINED/TRACKED FOLLOW-UP dispositions from the first pass still hold with no new evidence against them.

**Validation:** Re-run at the new head `d1b8621b` in a fresh detached worktree: `tsc --noEmit` clean across the touched package, `eslint . --cache` (repo-wide) clean with zero warnings, `prettier --check` clean on the three changed files, and the full `@archon/workflows` test suite green with 0 fail (`dag-executor.test.ts` 595/595, `loader.test.ts` + `include-expander.test.ts` 417/417 combined, `schemas.test.ts` + `command-validation.test.ts` + `packaged-workflow.test.ts` 199/199 combined, remaining workflows test files all green) — matches the implementer's reported `bun run validate` result.

**Resolved:** R1, R2, R3, R7 · **Unchanged (still non-blocking):** R4, R5, R6, R8

### Findings

|---|---|---|---|
| `R1` | Important | `runLayers`' node dispatch is an if-chain with a runtime throw, not a compiler-checked exhaustive `switch` — the PR's own AC1 is unmet | FIXED |
| `R2` | Important | `loader.ts`'s node-filter predicate asserts `n is DagNode` when `n` can be an `IncludeDirective` — a false, unverified type claim | FIXED |
| `R3` | Suggestion | A migrated test fixture uses the resolved node shape where the raw/flat shape was required, so the test passes for the wrong reason | FIXED |
| `R4` | Suggestion | Gate-node "reject" decision is selected via two different, currently-equivalent predicates in different files | DECLINED |
| `R5` | Suggestion | `isAgentNode(node) && node.source.kind === 'command'` (plus its `with:`-extraction ternary) is duplicated near-verbatim ~14 times across 8 files | DECLINED |
| `R6` | Suggestion | 11 `as DagNode[]` casts each carry their own repeated comment instead of one shared runtime-checked helper | DECLINED |
| `R7` | Suggestion | A provably-unreachable branch silently defaults to `''` instead of throwing, against the file's own established convention | FIXED |
| `R8` | Suggestion | `ExecNode.deps` still silently drops on `bash:` nodes with no warning — the one sibling instance of the bug class this PR fixed for `with:` | TRACKED FOLLOW-UP |

<details>
<summary><code>R1</code> — <code>runLayers</code>' dispatch fails to deliver the PR's own AC1 (compiler-enforced exhaustiveness)</summary>

**Impact:** The plan's Task 5 and the issue's AC1 both state the deliverable explicitly: *"`runLayers`' `switch (node.kind)` in `dag-executor.ts` has a `never`-typed default branch that fails to compile if a kind is missing a case."* The delivered code was an ordered `if`/`else if` chain of type guards ending in a runtime `throw`, conceding in its own comment that "the compiler cannot itself prove their union is exhaustive."

**Evidence (original):** `packages/workflows/src/dag-executor.ts:8479-8734`.

**Required outcome:** Convert the if-chain to `switch (node.kind) { ...; default: { const _exhaustive: never = node; throw new Error(...); } }`.

**Found by:** `prp-core:type-design-analyzer`, `prp-core:code-reviewer`

**Disposition:** FIXED — verified directly against `d1b8621b`. `runLayers` now dispatches via `switch (node.kind)` covering all 7 `DagNode` kinds (`exec`, `loop`, `loop_group`, `gate`, `halt`, `workflow` each return; `agent` breaks to the shared agent-handling code that follows), with a `default: { const unreachable: never = node; throw new Error(...) }` branch that mirrors the exact idiom already used at `dag-executor.ts:3865` (`describeLoopCompletionCheck`) and `:6925` (the sub-run `outcome.status` switch) cited in the original finding. Confirmed the schema declares exactly these 7 `kind` literals (`packages/workflows/src/schemas/dag-node.ts`), so the switch is genuinely exhaustive today, and adding an 8th kind now fails to compile rather than only failing at runtime. The now-unused `isWorkflowNode` import was correctly dropped (no remaining call sites in the file). `tsc --noEmit` and `dag-executor.test.ts` (595/595) both pass at the new head.

</details>

<details>
<summary><code>R2</code> — A false type predicate re-narrows <code>DagNode | IncludeDirective</code> back to <code>DagNode</code> right after this PR widened it</summary>

**Impact:** `loader.ts:955-958` filtered `parseDagNode`'s output with `.filter((n): n is DagNode => n !== null)`, even though `parseDagNode` was widened by this PR to return `DagNode | IncludeDirective | null`. TypeScript trusts a hand-written predicate without verifying it, so `dagNodes` was typed `DagNode[]` for ~600 subsequent lines despite `include:` nodes surviving the filter as real `IncludeDirective` values.

**Evidence (original):** `packages/workflows/src/loader.ts:349` vs. `:955-958`.

**Required outcome:** `.filter((n): n is DagNode | IncludeDirective => n !== null)`, matching what the predicate actually filters, plus guards at each downstream `DagNode`-only field access.

**Disposition:** FIXED — verified directly against `d1b8621b`. The predicate now reads `.filter((n): n is DagNode | IncludeDirective => n !== null)` (`loader.ts:958`), matching `parseDagNode`'s actual widened return type. `isIncludeDirective` guards were added to all three downstream loops that iterate `dagNodes` and read `DagNode`-only fields: the provider-validation loop (`:1005`), the `context.resume` session-continuity loop (`:1021`, plus an added `isIncludeDirective(source)` check at `:1025` after the `.find`), and the `persist_session` loop (`:1077`) — matching the pre-existing `hasInteractiveLoop`/`hasSignalCompletesWithoutInteractive` pattern the original finding cited as the correct precedent. `isIncludeDirective` was already imported in this file (line 19), so no new import was needed. `tsc --noEmit` and `loader.test.ts` + `include-expander.test.ts` (417/417 combined) both pass at the new head.

</details>

<details>
<summary><code>R3</code> — Migrated test fixture uses the resolved shape where the raw/flat shape was required, silently weakening the only test for this rule</summary>

**Impact:** `schemas.test.ts:860-871`'s `'loop_group + loop are mutually exclusive'` test constructed its body node in the resolved post-transform shape (`{kind: 'agent', source: {...}}`), which `dagNodeSchema.safeParse` (pre-transform) strips to `{id: 'x'}`, so the test's `result.success === false` assertion passed for the wrong reason — driven by the unrelated malformed-body-node error, not the mutual-exclusivity rule under test.

**Evidence (original):** `packages/workflows/src/schemas.test.ts:860-871`, reproduced against the pre-fix fixture.

**Required outcome:** Revert to the raw shape (`{id: 'x', prompt: 'x'}`) and assert on `result.error.issues[0].message` containing `'mutually exclusive'`.

**Disposition:** FIXED — verified directly against `d1b8621b`. The fixture is reverted to `{id: 'x', prompt: 'x'}` (matching every sibling test in the block) and the test now asserts `result.error.issues[0].message` contains `'mutually exclusive'`. Independently reproduced in this worktree via a standalone probe against the exact fixture from the diff: `dagNodeSchema.safeParse(...)` now emits a single issue whose message is `"'command', 'prompt', 'bash', 'loop', 'loop_group', 'approval', 'cancel', 'script', 'include', and 'workflow' are mutually exclusive"` at `issues[0]` — the malformed-body-node error is gone because the body node is now well-formed, and the assertion now genuinely covers the intended rule. `schemas.test.ts` (part of the 199/199 combined run with `command-validation.test.ts` and `packaged-workflow.test.ts`) passes at the new head.

</details>

<details>
<summary><code>R4</code> — Gate-node "reject" decision selected via two different predicates</summary>

**Impact:** `dag-executor.ts` selects the reject decision via `node.decisions.find(d => d.id === 'reject')` in one place and via `d.rework !== undefined` in two others (`loader.ts`, `dag-executor.ts`). Currently equivalent given the transform's fixed output; would diverge if #2707 makes `decisions` author-declarable.

**Evidence:** `packages/workflows/src/dag-executor.ts:6488`, `:9289`; `packages/workflows/src/loader.ts:666` (line numbers shifted slightly by the correction diff; same sites, same code, unchanged by `d1b8621b`).

**Required outcome:** Not required for this PR's scope.

**Disposition:** DECLINED — unchanged. Confirmed the correction diff (`git diff` between the two heads) does not touch any of these three sites; no new evidence to reopen. Still worth converging before/alongside #2707, not a blocker here.

</details>

<details>
<summary><code>R5</code> — <code>isAgentNode(node) && node.source.kind === 'command'</code> duplicated ~14 times across 8 files</summary>

**Impact:** The compound check plus its `with:`-extraction ternary is repeated near-verbatim across `command-file.ts`, `dry-run.ts`, `loader.ts`, `packaged-workflow.ts`, `include-expander.ts`, `dag-executor.ts`, `validator.ts`.

**Evidence:** File list above (unchanged by the correction diff).

**Required outcome:** Not required for this PR to merge.

**Found by:** `prp-core:code-reviewer`, `prp-core:code-simplifier` (independently converged)

**Disposition:** DECLINED — unchanged. None of the named files outside `dag-executor.ts`/`loader.ts` were touched by `d1b8621b`, and the touched portions of those two files don't intersect this duplication. Reasonable follow-up, not required before merge given this PR's own scope is the type reshape, not introducing new shared accessors.

</details>

<details>
<summary><code>R6</code> — 11 duplicated <code>as DagNode[]</code> casts, each with its own repeated comment and no runtime check</summary>

**Impact:** Every post-expansion call site that needs `(DagNode | IncludeDirective)[]` narrowed to `DagNode[]` uses a bare cast plus a repeated comment instead of a shared runtime-checked helper.

**Evidence:** File:line list from the original review (unchanged by the correction diff).

**Required outcome:** Not required for this PR to merge.

**Disposition:** DECLINED — unchanged. Re-confirmed the cast sites in `dag-executor.ts`, `dry-run.ts`, `executor.ts`, `include-expander.ts`, `utils/workflow-requirements.ts`, and `packages/core/src/orchestrator/orchestrator.ts` are all still present and untouched by `d1b8621b`. Worthwhile simplification, not a blocker.

</details>

<details>
<summary><code>R7</code> — Silent <code>''</code> fallback on a provably-unreachable branch instead of a throw</summary>

**Impact:** `dag-executor.ts` (inside `executeNodeInternal`) defaulted `rawPrompt` to `''` on a branch that's unreachable by construction (`AgentBody.source` is a two-member union) but not provably so to the compiler, against the file's own fail-fast convention used 200 lines away.

**Evidence (original):** `packages/workflows/src/dag-executor.ts:1837` (commandName derivation), `:1935` (the fallback).

**Required outcome:** Replace the ternary with an explicit throw when `node.source.kind !== 'inline'`.

**Disposition:** FIXED — verified directly against `d1b8621b`. The ternary is replaced with `if (node.source.kind !== 'inline') { throw new Error(\`unreachable: node '${node.id}' has no commandName but source.kind is not 'inline'\`); } rawPrompt = node.source.prompt;`, with a comment explaining why the branch is unreachable by construction but guarded loudly anyway — matching the same fail-fast convention already used at `dag-executor.ts:8734` (the R1 fix's own `default` branch) that the original finding cited. `tsc --noEmit` and `dag-executor.test.ts` (595/595) both pass at the new head.

</details>

<details>
<summary><code>R8</code> — <code>ExecNode.deps</code> still silently drops on <code>bash:</code> nodes, unlike the sibling <code>with:</code> field this PR fixed</summary>

**Impact:** `deps` on `ExecNode` is documented as meaningful only when `runtime !== 'sh'`, but no warning exists for `deps:` on a `bash:` node (unlike the `with:`-on-bash warning this PR added). Pre-existing, not introduced by this PR.

**Evidence:** `packages/workflows/src/schemas/dag-node.ts:384`, `:1391-1400`; `packages/workflows/src/loader.ts:377-392`; `dag-node.ts:708-729` (`BASH_NODE_AI_FIELDS`, confirmed `deps` still absent at the new head).

**Required outcome:** Not required for this PR — worth a one-line addition extending the existing `with:`-on-bash warning mechanism to also cover `deps:`, as a small follow-up.

**Disposition:** TRACKED FOLLOW-UP — unchanged. Re-confirmed `deps` is still absent from `BASH_NODE_AI_FIELDS` at `d1b8621b`; the gap is pre-existing and untouched by this correction. Not filed as a separate issue; noted here for whoever picks up `deps:` warning coverage next.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1, R4, R5 — R1 verified FIXED at `d1b8621b`; R4, R5 unchanged DECLINED |
| `seams` | R2, R3 — both verified FIXED at `d1b8621b` |
| `simplify` | R6, R7 — R7 verified FIXED at `d1b8621b`; R6 unchanged DECLINED |
| `types` | R1 (independently), R8 — R1 verified FIXED; R8 unchanged TRACKED FOLLOW-UP |

#### Validation

|---|---|---|
| `bun x tsc --noEmit` (`@archon/workflows`) | PASS | Clean at `d1b8621b` in a fresh detached worktree pinned to the exact PR head |
| `bun x eslint . --cache` (repo-wide) | PASS | Zero warnings/errors at `d1b8621b` |
| `bun x prettier --check` (3 changed files: `dag-executor.ts`, `loader.ts`, `schemas.test.ts`) | PASS | "All matched files use Prettier code style!" |
| `bun test src/dag-executor.test.ts` | PASS | 595 pass / 0 fail, 1678 expect() calls |
| `bun test src/loader.test.ts src/include-expander.test.ts` | PASS | 417 pass / 0 fail, 1119 expect() calls |
| `bun test src/schemas.test.ts src/command-validation.test.ts src/packaged-workflow.test.ts` | PASS | 199 pass / 0 fail, 390 expect() calls |
| `bun run test` (full `@archon/workflows` suite, all remaining test files) | PASS | 0 fail across every remaining file in the package script |
| Direct probe reproducing the R3 fixture's schema parse | Confirms FIXED | `issues[0].message` is now the mutual-exclusivity error, not the malformed-body-node error |
| Direct read of `runLayers` dispatch and `BASH_NODE_AI_FIELDS`/cast sites/reject-predicate sites | Confirms R1/R7 fixes and R4/R5/R6/R8 unchanged | See per-finding evidence above |

This matches the implementer's independently reported `bun run validate` (full monorepo gate, exit 0).

</details>

Wirasm:
Filed #2718 to track the decision to keep `executeBashNode`/`executeScriptNode` separate rather than unifying them behind the new single `exec` kind — see the PR body's Solution section for the reasoning; #2718 makes it a named, linkable follow-up per the operator's review flag.

PR #2717, opened 2026-08-22
fix(web): surface unrecognised workflow sources verbatim (#2578)

### LINKED ISSUE #2578: fix(web): an unrecognised workflow source collapses to 'bundled', silently making it read-only
## Summary

- **What broke:** the console hand-redeclares `WorkflowSource` and maps any value it doesn't recognise to `'bundled'`. Because `'bundled'` is the read-only source, an unrecognised source silently makes a workflow read-only and turns a save into a Save-as copy.
- **When it started:** pre-existing. The same mechanism already shipped one bug — the function's own comment records that collapsing `global` into `bundled` silently demoted home-scoped workflows and gave them the wrong badge.
- **Severity:** `minor` today (no unrecognised source is emitted yet), but the failure is silent and the blast radius is a user's edit.

## Root cause

`packages/web/src/experiments/console/primitives/workflow.ts:1` declares the union by hand:

```ts
export type WorkflowSource = 'project' | 'global' | 'bundled';
```

and `toWorkflow` (`:24-36`) resolves it with a two-branch ternary whose else-arm is a value, not a failure:

```ts
const src: WorkflowSource =
  raw.source === 'project' ? 'project' : raw.source === 'global' ? 'global' : 'bundled';
```

`raw.source` is typed `string` (`:20`), so anything the server sends that isn't one of the two named strings becomes `'bundled'` — a typo, a new source added engine-side, or a value that reaches the client before the console is updated.

## Why `'bundled'` is the expensive default

It isn't a neutral fallback. Two behaviours key off it in `packages/web/src/experiments/console/builder/connect/save-logic.ts`:

```ts
/** Bundled workflows open read-only; everything else is editable in place. */
export function isReadOnlySource(source: WorkflowSource): boolean {
  return source === 'bundled';          // :86-88
}
```

and `saveTargetFor` (`:95`) makes a bundled workflow *"save as a project override (Save-as)"* rather than saving in place.

So a workflow whose source the console doesn't recognise opens read-only, and if the user saves, they get a **copy** instead of an edit to the original. Nothing errors and nothing warns; the user finds out later, if at all.

The comment directly above the ternary already tells this story about the previous instance:

> Collapsing `global` into `bundled` silently demoted home-scoped workflows and rendered them with the wrong badge.

The fix at the time added a `global` branch. It did not change the shape that produced the bug.

## Proposed fix

Two independent parts, either of which is worthwhile:

1. **Stop guessing.** Make the else-arm loud rather than a value — surface the unrecognised source under its own raw name, or fail the mapping — so a new engine-side source shows up as "unknown" instead of masquerading as a read-only bundled workflow. Falling back to the *most permissive* interpretation would also be defensible; falling back to the one that silently takes away the user's edit is not.
2. **Stop hand-declaring the union.** `WorkflowSource` is derivable from the generated OpenAPI types (`packages/web/src/lib/api.generated.d.ts`) rather than retyped here, which is what would have made a new engine-side value a compile error instead of a runtime collapse.

This is the same defect species #2567 / PR #2570 just removed from the node-ref scanners — a hand-maintained mirror of an engine concept, drifting silently. Worth resolving in the same direction rather than the opposite one.

## Impact

- Affected surfaces: console workflow builder — read-only gating and save target
- Reproduction rate: does not reproduce today (the server emits only the three known values); reproduces the moment a fourth is added or a value is misspelled
- Workaround: none, because there is no signal that anything went wrong
- Data loss risk? `No` — but a save silently becoming a copy means an intended edit to the original does not happen

## Scope

- Package(s) likely involved: `web` (`experiments/console/primitives/workflow.ts`, `builder/connect/save-logic.ts`)
- Breaking change? `No`

## Related

- #2567 / PR #2570 — removed the same species of hand-maintained engine mirror from the node-ref scanners; found this one while reviewing that
- #2213 — `parseWarnings`, the other field this same `toWorkflow` carries

Wirasm:
## Problem and outcome

The console's `toWorkflow` collapsed any unrecognised `WorkflowSource` value from the server to `'bundled'`, which is the read-only source. A misspelt or future source therefore opened a workflow read-only and silently turned Save into Save-as (a copy to a project override) with no signal to the user.

- **Outcome:** an unrecognised `Workflow.source` is now surfaced verbatim through `toWorkflow` (and through every downstream consumer), so a server-side drift shows up as "unknown" rather than masquerading as a read-only bundled workflow.
- **Invariant:** the web client's `WorkflowSource` union is byte-identical to the server's `z.enum(['project', 'bundled', 'global']).openapi('WorkflowSource')` (engine-side at `packages/server/src/routes/schemas/workflow.schemas.ts:32`); the two can no longer drift silently.
- **Scope boundary:** no server, schema, API, or DB change. The legacy console/non-console builder split is untouched; this fix lands entirely inside `experiments/console/**` plus the existing re-export at `packages/web/src/lib/api.ts:377` is left as-is (sourcing from `@/lib/api.generated` reaches the same bytes either way).
- **Root cause:** `packages/web/src/experiments/console/primitives/workflow.ts:1` hand-declared `WorkflowSource = 'project' | 'global' | 'bundled'`, and `toWorkflow`'s ternary (`raw.source === 'project' ? 'project' : raw.source === 'global' ? 'global' : 'bundled'`) made the else-arm a value, not a failure. Two downstream helpers gate on the collapsed value (`isReadOnlySource` reads `=== 'bundled'`; `saveTargetFor` returns `'project'` for non-`'global'`), so an unknown value inherited the most-damaging default.

## Review guidance

- **Feedback requested:** correctness of the verbatim-passthrough contract, and whether widening `isReadOnlySource` / `saveTargetFor` to `string` is the right level (vs. keeping them narrow and casting at call sites).
- **Start here:** `packages/web/src/experiments/console/primitives/workflow.ts:39` — `Workflow.source` widens to `WorkflowSource | (string & {})`, and `toWorkflow` drops the collapsing ternary.
- **Review order:**
  1. `packages/web/src/experiments/console/primitives/workflow.ts` — the union is now `components['schemas']['WorkflowSource']` from `@/lib/api.generated`, and `toWorkflow` returns `raw.source` directly.
  2. `packages/web/src/experiments/console/builder/connect/save-logic.ts` — `isReadOnlySource(source: string)` and `saveTargetFor(source: string)` keep their `=== 'bundled'` / `=== 'global'` comparisons; the comparison semantics are unchanged.
  3. `packages/web/src/experiments/console/lib/recommended.ts` and `components/WorkflowPicker.tsx` — `sourceRank` and `sourceBadgeClass` get a `default` arm so the widened type doesn't crash on an unknown source.
- **Lower-attention areas:** the test swap in `workflow.test.ts`, the new sort case in `recommended.test.ts`, and the two new helper-contract cases in `save-logic.test.ts`; mechanical, and they pin the new contract.
- **Known risk or uncertainty:** none material. The widening deliberately accepts any string but the comparisons stay identical, so the known-source behaviour is preserved exactly.

## Solution

Two parts, mirroring the precedent PR #2570 set for the same defect species in `CommandGroup`:

1. **Stop hand-declaring the union.** `WorkflowSource` is now `components['schemas']['WorkflowSource']` from `@/lib/api.generated`. The console's eslint guard disallows named imports from `@/lib/api` inside `experiments/console/**` (only type-only imports from `api.generated` are allowed), so sourcing the type from the generated file directly is the same byte-equivalent re-export at `packages/web/src/lib/api.ts:377` reaches from the API surface; the existing `skills/workflows.ts:106` re-export keeps resolving to the same union.
2. **Surface unknown values verbatim.** `Workflow.source` widens to `WorkflowSource | (string & {})` — the `& {}` intersection keeps the three known literals visible in autocomplete while still accepting any string. `toWorkflow` passes `raw.source` through directly. The two consumers (`isReadOnlySource`, `saveTargetFor`) take `string` and keep their `===` comparisons, so unknown values fall to the "most permissive" interpretation the issue explicitly defends (editable in place, save targets `project`).

Two adjacent call sites needed defensive fixes for the widened type that the issue did not enumerate:

- `bySourceThenName` in `lib/recommended.ts` indexed a literal-keyed `{project:0, global:1, bundled:2}` map with the wider `Workflow['source']`, which `tsc` flagged as potentially `undefined`. Replaced with a `sourceRank` helper that has a `default` arm returning `3` (unknown sources sort last, deterministically).
- `sourceBadgeClass` in `components/WorkflowPicker.tsx` used a `switch` without a `default`, which `tsc` flagged because the widened input is no longer an exhaustive literal. Added a `default` arm rendering `text-text-tertiary` so the badge never aliases a known source or crashes.

The fix is the smallest coherent change that closes the bug without re-introducing the hand-maintained mirror at the next engine-side drift.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `toWorkflow({source: 'something-new'})` returns `source: 'bundled'`; the workflow opens read-only and Save becomes Save-as (project override copy) | `toWorkflow({source: 'something-new'})` returns `source: 'something-new'`; the workflow is editable in place, Save targets `project`, the badge renders muted |
| **Failure behavior** | silent — no warning, no error, the user's intended edit to the original becomes a copy | visible — the unknown source surfaces under its own raw label in the picker and badge, so the drift is observable in the UI |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[Server emits unknown source] --> B2[Console collapses to 'bundled'] --> B3[Workflow opens read-only; Save becomes Save-as; no signal]
  end

  subgraph After
    A1[Server emits unknown source] --> A2[Console surfaces raw value] --> A3[Workflow stays editable; Save targets project; badge renders muted]
  end
```

## Architecture

```mermaid
flowchart LR
  R[RawWorkflowEntry.source: string] --> T[toWorkflow] --> W[Workflow.source: WorkflowSource | string & {}]
  W --> I[isReadOnlySource: string] --> RO[read-only?]
  W --> S[saveTargetFor: string] --> ST[save target]
  W --> B[sourceBadgeClass: switch + default] --> BC[badge colour]
  W --> RK[sourceRank: switch + default] --> OR[orderWithRecommended]
  API[components schemas WorkflowSource] -.generated.-> W
```

### Changed seams

|---|---|---|
| `primitives/workflow.ts` `WorkflowSource` | re-typed from hand-declared literal union to `components['schemas']['WorkflowSource']` | `packages/web/src/experiments/console/primitives/workflow.ts:14`, OpenAPI at `packages/server/src/routes/schemas/workflow.schemas.ts:32` |
| `primitives/workflow.ts` `toWorkflow` | raw `raw.source` passed through verbatim; the collapsing ternary removed | `packages/web/src/experiments/console/primitives/workflow.ts:82` |
| `Workflow.source` field type | widened to `WorkflowSource | (string & {})` | `packages/web/src/experiments/console/primitives/workflow.ts:39` |
| `save-logic.ts` `isReadOnlySource` / `saveTargetFor` parameter | widened from `WorkflowSource` to `string`; comparisons unchanged | `packages/web/src/experiments/console/builder/connect/save-logic.ts:91,103` |
| `recommended.ts` `bySourceThenName` | replaced literal-keyed map with `sourceRank` helper carrying a `default` arm | `packages/web/src/experiments/console/lib/recommended.ts:9` |
| `WorkflowPicker.tsx` `sourceBadgeClass` | `default` arm added for the widened input type | `packages/web/src/experiments/console/components/WorkflowPicker.tsx:42` |

## Validation

- `bun run type-check` (`tsc --noEmit`) — clean — proves the widened `Workflow.source` threads through `isReadOnlySource`, `saveTargetFor`, `sourceRank`, and `sourceBadgeClass` without a cast at the call sites.
- `bun run test` — 452 pass, 0 fail, 35 files — proves the verbatim-passthrough contract (`workflow.test.ts`), the unrecognised-source sort order (`recommended.test.ts`), and the helper-contract cases (`save-logic.test.ts`) all hold.
- `bun test src/experiments/console` — all pass, including the two new #2578 regression tests.
- `bun x eslint packages/web/src --max-warnings 0` — clean.
- `bun x prettier --check` on the six changed files — clean.
- Manual reasoning: the comparisons `=== 'bundled'` and `=== 'global'` are unchanged, so the known-source behaviour is preserved bit-for-bit; only the unknown-source branch changes (and now resolves to the most permissive interpretation the issue explicitly defends).

## Links

- Closes #2578
- Related #2567, PR #2570

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved unrecognized workflow source values instead of incorrectly converting them to bundled workflows.
  * Ensured workflows with unknown sources remain editable and save to the appropriate project scope.
  * Added neutral styling for unknown workflow sources.
  * Improved workflow recommendations by placing unknown sources after recognized workflow types with consistent ordering.
* **Tests**
  * Added regression coverage for editing, saving, normalization, styling, and recommendation behavior.

### DIFF
--- a/packages/web/src/experiments/console/builder/BuilderConnected.tsx
+++ b/packages/web/src/experiments/console/builder/BuilderConnected.tsx
@@ -49,13 +49,13 @@ import {
   validateWorkflow,
   listWorkflows,
   type LoadedWorkflow,
-  type WorkflowSource,
 } from '../skills/workflows';
 import { listProjects, type WorkflowListResult } from '../skills';
 import { useEntity, invalidate } from '../store/cache';
 import { K } from '../store/keys';
 import { HttpError } from '../lib/http';
 import type { Project } from '../primitives/project';
+import type { Workflow } from '../primitives/workflow';
 
 /** Router navigation state carried into the connected route. */
 interface BuilderNavState {
@@ -127,7 +127,10 @@ export function BuilderConnected(): ReactElement {
   // Resolve the workflow under edit (seed in create mode, else the server load).
   // Memoized on stable inputs (location-state seed, cache object) so it does NOT
   // produce a fresh identity every render — that would thrash the reset effect.
-  const loadedSource: WorkflowSource = isCreateMode
+  // Widened to `Workflow['source']` (#2578) so an unrecognised wire value passes
+  // through to `effectiveSource` (and the downstream `isReadOnlySource` /
+  // `saveTargetFor` helpers) without a cast at this seam.
+  const loadedSource: Workflow['source'] = isCreateMode
     ? 'project'
     : (loadView.data?.source ?? 'project');
 
@@ -144,7 +147,10 @@ export function BuilderConnected(): ReactElement {
   const [dirty, setDirty] = useState(false);
   const [serverIssues, setServerIssues] = useState<Issue[]>([]);
   // Source can flip after a bundled Save-as (bundled → project override).
-  const [sourceOverride, setSourceOverride] = useState<WorkflowSource | null>(null);
+  // Widened to `Workflow['source']` (#2578) to match `loadedSource` /
+  // `LoadedWorkflow.source` — `saveWorkflow` returns whatever the server put on
+  // the response, which is `Workflow['source']`-typed.
+  const [sourceOverride, setSourceOverride] = useState<Workflow['source'] | null>(null);
   const [busy, setBusy] = useState(false);
 
   const effectiveSource = sourceOverride ?? loadedSource;
--- a/packages/web/src/experiments/console/builder/connect/save-logic.test.ts
+++ b/packages/web/src/experiments/console/builder/connect/save-logic.test.ts
@@ -164,6 +164,20 @@ describe('isReadOnlySource / saveTargetFor', () => {
     expect(saveTargetFor('project')).toBe('project');
     expect(saveTargetFor('global')).toBe('global');
   });
+
+  // #2578 — the helpers accept `string` precisely so an unrecognised wire value
+  // does not silently become read-only / write-to-global. These tests pin that
+  // observable contract against the regressions noted in the review (a
+  // `source !== 'project' && source !== 'global'` rewrite would make the
+  // editor offer Save-as instead of Save; an inverted `saveTargetFor` would
+  // write the workflow into `~/.archon/workflows/` instead of the project).
+  test('an unrecognised source is editable in place, not read-only (#2578)', () => {
+    expect(isReadOnlySource('something-new')).toBe(false);
+  });
+
+  test('an unrecognised source saves to project scope, the most permissive target (#2578)', () => {
+    expect(saveTargetFor('something-new')).toBe('project');
+  });
 });
 
 describe('isValidWorkflowName', () => {
--- a/packages/web/src/experiments/console/builder/connect/save-logic.ts
+++ b/packages/web/src/experiments/console/builder/connect/save-logic.ts
@@ -6,7 +6,7 @@
 import type { Issue } from '../types';
 import { makeIssue } from '../validation/make-issue';
 import { HttpError } from '../../lib/http';
-import type { WorkflowSource, WorkflowSaveSource } from '../../skills/workflows';
+import type { WorkflowSaveSource } from '../../skills/workflows';
 
 /** A client-side instant error for the panel (name validation, save gate). */
 export function clientIssue(rule: string, message: string): Issue {
@@ -82,17 +82,25 @@ export function blockingErrors(issues: readonly Issue[]): Issue[] {
   return issues.filter(i => i.severity === 'error');
 }
 
-/** Bundled workflows open read-only; everything else is editable in place. */
-export function isReadOnlySource(source: WorkflowSource): boolean {
+/**
+ * Bundled workflows open read-only; everything else is editable in place. The
+ * parameter is widened to `string` so an unrecognised `Workflow.source` (#2578,
+ * `WorkflowSource | (string & {})`) can pass through without a cast at the
+ * call site — the comparison is the same `=== 'bundled'` either way.
+ */
+export function isReadOnlySource(source: string): boolean {
   return source === 'bundled';
 }
 
 /**
  * The write scope for a save. A `global` workflow saves back to global; a
  * `project` workflow stays project; a `bundled` (read-only) workflow saves as a
- * project override (Save-as).
+ * project override (Save-as). The parameter is widened to `string` for the
+ * same reason `isReadOnlySource` is (#2578): an unrecognised source is not
+ * `global`, so it falls to `'project'`, which is the "most permissive"
+ * interpretation the issue explicitly defends.
  */
-export function saveTargetFor(source: WorkflowSource): WorkflowSaveSource {
+export function saveTargetFor(source: string): WorkflowSaveSource {
   return source === 'global' ? 'global' : 'project';
 }
 
--- a/packages/web/src/experiments/console/components/WorkflowPicker.tsx
+++ b/packages/web/src/experiments/console/components/WorkflowPicker.tsx
@@ -39,6 +39,11 @@ function sourceBadgeClass(source: Workflow['source']): string {
       return 'text-text-secondary';
     case 'bundled':
       return 'text-text-tertiary';
+    default:
+      // An unrecognised source surfaces under its own raw value (#2578) — render
+      // it with a neutral muted colour rather than crashing or aliasing one of
+      // the known sources.
+      return 'text-text-tertiary';
   }
 }
 
--- a/packages/web/src/experiments/console/lib/recommended.test.ts
+++ b/packages/web/src/experiments/console/lib/recommended.test.ts
@@ -52,4 +52,18 @@ describe('orderWithRecommended', () => {
     expect(ordered.map(w => w.name)).toEqual(['beta', 'alpha', 'gamma']);
     expect(ordered.filter(w => w.name === 'beta')).toHaveLength(1);
   });
+
+  test('an unrecognised source ranks after the three known sources (#2578)', () => {
+    // Mirrors `toWorkflow`'s verbatim-passthrough contract: a misspelled or
+    // future source must not collapse to 'bundled' (or any known value), and
+    // the picker must place it deterministically rather than indexing an
+    // undeclared key on the rank map.
+    const workflows = [
+      wf('z-bundled', 'bundled'),
+      wf('a-unknown', 'something-new'),
+      wf('m-project', 'project'),
+    ];
+    const { ordered } = orderWithRecommended(workflows, []);
+    expect(ordered.map(w => w.name)).toEqual(['m-project', 'z-bundled', 'a-unknown']);
+  });
 });
--- a/packages/web/src/experiments/console/lib/recommended.ts
+++ b/packages/web/src/experiments/console/lib/recommended.ts
@@ -1,9 +1,27 @@
 import type { Workflow } from '../primitives/workflow';
 
+/**
+ * Source rank for the "other" group: project first, then global, then bundled,
+ * then any unrecognised source at the end (#2578). The fallback covers the
+ * widened `Workflow['source']` (`WorkflowSource | (string & {})`) without
+ * indexing a literal-keyed map with a `string` and getting `undefined`.
+ */
+function sourceRank(source: Workflow['source']): number {
+  switch (source) {
+    case 'project':
+      return 0;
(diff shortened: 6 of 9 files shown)


Wirasm:
# Review Report — PR #2717 (round 2, light / delta)

## Verdict

**READY.** Both prior-round Important findings are fixed at the head SHA: the four annotations the seams lens named are widened to `Workflow['source']` verbatim, and the two helper-layer tests the tests lens prescribed are appended verbatim and pass. No new findings. Every enabled lens delivered its report.

## Reviewed head SHA

`ae021d9e3693fa43e1cfa7ed14ee26d507822df0`

## Findings

None open.

## Rejected findings

None — no new findings were raised this round.

## Suggestions

None.

## Prior findings

| ID | Severity | Lens | Verdict |
|---|---|---|---|
| **R1** | Important | `seams` | **Fixed at `ae021d9e`.** The four annotations named by the prior report — `GetWorkflowResponse.source` (`skills/workflows.ts:122`), `LoadedWorkflow.source` (`skills/workflows.ts:135`), `loadedSource` (`BuilderConnected.tsx:133`), `sourceOverride` (`BuilderConnected.tsx:153`) — are all widened to `Workflow['source']` (≡ `WorkflowSource \| (string & {})` per `primitives/workflow.ts:39`). The widening matches the model layer introduced in the prior round and carries explanatory comments at each of the four sites tying the change to #2578. No body changes; no cast introduced. `Workflow['source']` is not a finite union, so a missing `default` arm in any future exhaustive switch over `effectiveSource` / `loadedSource` / `sourceOverride` is now a compile error. |
| **R2** | Important | `tests` | **Fixed at `ae021d9e`.** The two tests the prior report prescribed are appended verbatim at the end of the `describe('isReadOnlySource / saveTargetFor', …)` block in `packages/web/src/experiments/console/builder/connect/save-logic.test.ts`: `isReadOnlySource('something-new') === false` at lines 174-176 and `saveTargetFor('something-new') === 'project'` at lines 178-180. The block-level comment at lines 168-173 names both named regressions (the `source !== 'project' && source !== 'global'` rewrite and the inverted `saveTargetFor` ternary) and the user-visible failure each causes, so the pin is self-documenting. The `'something-new'` literal matches the vocabulary already canonised at `primitives/workflow.test.ts:17-19` and `lib/recommended.test.ts:63`. Both pass; the full file reports 30 pass / 0 fail. |

## Lens roster

| Lens | Status | Notes |
|---|---|---|
| `code` | Ran, clean | No new findings. Verifies R1 and R2 fixed at `ae021d9e`. 454 console tests pass; type-check green; lint clean on the three delta files. |
| `seams` | Ran, clean | No new findings. The four annotation widenings are verified end-to-end: type flows from `primitives/workflow.ts:39` → `GetWorkflowResponse.source` → `LoadedWorkflow.source` → `loadView.data?.source` → `loadedSource` / `sourceOverride` → `effectiveSource` → `isReadOnlySource` / `saveTargetFor`. No exhaustive switch, no direct equality, no cast introduced anywhere in `experiments/console/` on the widened value. |
| `tests` | Ran, clean | No new findings. Both R2 assertions are decisive against the two regressions the prior report named; existing helper-layer coverage on the three known sources is preserved unchanged. |
| `errors` | Disabled by scope | No file by design — `errors=false` in scope. |
| `comments` | Disabled by scope | No file by design — `comments=false` in scope. |
| `types` | Disabled by scope | No file by design — `types=false` in scope. |
| `docs` | Disabled by scope | No file by design — `docs=false` in scope. |
| `simplify` | Disabled by scope | No file by design — `simplify=false` in scope. |

All enabled lenses delivered their report. No gap.

## PR publication

This report was published as a single canonical comment on PR #2717 (carrying the marker `<!-- archon-review-report -->` on its first line). The prior marked comment from round 1 (id `5380628641`) was located and edited in place via `gh api`; no second comment was appended. The body was read back to confirm it matches this artifact.

Comment URL: https://github.com/coleam00/Archon/pull/2717#issuecomment-5380628641 (read back and confirmed — body matches this artifact, marker on first line).

## Verification

- Head SHA appears verbatim above and matches the local checkout (`ae021d9e` is HEAD on `archon/task-fix-1787403546985`).
- Both Important findings were adversarially verified against the cited `file:line` evidence (read directly from the working tree at the head SHA — every cited line in `skills/workflows.ts`, `BuilderConnected.tsx`, `save-logic.test.ts`, and `primitives/workflow.ts` was confirmed to match the lens reports).
- Every lens file present in `review/` (`code.md`, `seams.md`, `tests.md`) is accounted for in the roster. The disabled lenses (`errors`, `comments`, `types`, `docs`, `simplify`) are recorded as disabled by scope, not as missing.
- All enabled lenses delivered their report; no gap.
- The PR comment carries the canonical marker and was read back to confirm.

---

## Artifacts

- Scope: `review/scope.md`
- Lens reports: `review/code.md`, `review/seams.md`, `review/tests.md`
- Synthesized report: `review/report.md` (this file)

PR #2719, opened 2026-08-22
fix(workflows): a failed producer's output stops resolving through when:, inline substitution, and plain with: values

### LINKED ISSUE #2713: A failed producer's stale output still resolves through when: conditions and inline $node.output substitution
## Problem

PR #2710 (fixing #2696) added a `producer.state === 'failed'` guard to `resolveBindingDirective` in `packages/workflows/src/dag-executor.ts` — a failed producer's `{ from, if_skipped }` binding now always throws instead of silently resolving stale output (e.g. a `loop_group`'s last-completed-iteration text, which survives `max_iterations` exhaustion, body-node failure, and cancellation, and is genuinely non-empty).

That fix is scoped to exactly one reader of `NodeOutput`: `resolveBindingDirective`. Three sibling readers of the same `nodeOutputs` map have no equivalent guard, and can reach the identical "consumer executes its real body on a failed producer's stale output" failure mode that caused the original incident (run `6607bf20`: a failed `corrections` loop_group's stale output flipped PR #2705 to ready with an unreviewed commit, while the run itself finished `failed`):

1. **`resolveWorkflowValue`'s plain-string whole-ref tier** — `packages/workflows/src/dag-executor.ts:322-326`. `if (producer !== undefined) return wholeRefLogicalValue(producer, ...)` — no state check at all. This is the resolver for: a script/command node's own plain (non-directive) `with:` values, a `workflow:` node's `with:` (which only ever takes plain values — it can never use the `{from, if_skipped}` directive form, so it can never get #2710's protection), and `fan_out`'s static `with:` inputs. This gap was already known and explicitly deferred by #2710's plan/PR (see its "Not building" section) as a "related but separately-scoped gap."

2. **`condition-evaluator.ts`'s `resolveOutputRef`** (the `when:` expression resolver) — `packages/workflows/src/condition-evaluator.ts:88-92`. The unfielded branch does `if (!nodeOutput.output) return ''; return nodeOutput.output;` with no state check. Its own inline comment claims the unfielded case is "empty for failed nodes" — directly false for a failed `loop_group`, which is the exact premise #2696/#2710 disproved. A `when:` condition like `"$corrections.output != ''"` or `"$corrections.output.ready == true"` on a node joined via `trigger_rule: all_done` against a failed `loop_group` evaluates against the group's stale last-iteration text and can route execution down the wrong branch — this was NOT identified or scoped out by #2710's plan.

3. **`substituteNodeOutputRefs`'s unfielded branch** (inline `$node.output` text substitution in prompt/bash/command bodies) — `packages/workflows/src/dag-executor.ts:1217-1220`. `if (!field) { return escapedForBash ? shellQuoteOrFile(nodeOutput.output, ...) : nodeOutput.output; }` — no state check. A `prompt:`/`bash:`/`command:` body that inlines `$node.output` against a failed producer splices the stale text in verbatim, with no error and no warning. Also not identified or scoped out by #2710's plan.

`when:` conditions and inline `$node.output` prompt/bash substitution are arguably the most commonly authored ways to reference a producer's output — more common than the newer `{ from, if_skipped }` binding-directive form #2710 fixed — so these two undisclosed gaps are a larger blast radius than the one gap #2710's plan already knew about and deferred.

## Why this is a real, not hypothetical, gap

Confirmed by direct code reading (not speculation): both `condition-evaluator.ts:88-92` and `dag-executor.ts:1217-1220` read `nodeOutput.output` unconditionally, with zero `state` check. For a `loop_group`, every failure path (`dag-executor.ts:4327-4335` body-node failure, `:4653-4667` `max_iterations` exhaustion, `:4301` cancellation) sets `output: lastIterationOutput` — the real, non-empty terminal text of the last completed iteration, captured before the completion check that ultimately failed. Any workflow author using `trigger_rule: all_done` to join past a failed `loop_group` (the exact `deliver`-pattern shape #2696 was about) and referencing that producer via `when:` or inline prompt/bash text, rather than the `{from, if_skipped}` directive, hits the same split-brain class of bug #2710 fixed for the directive form.

## Fix shape

Same one-line pattern as #2710's fix (`packages/workflows/src/dag-executor.ts:437-448`): add a `producer.state === 'failed'` check ahead of the unfielded/whole-text read in each of the three call sites above (or factor a small shared helper, since the exact wording of the thrown message legitimately differs per caller — the directive path can mention `if_skipped`, the plain-value/`when:`/prompt paths can't). Also correct `condition-evaluator.ts`'s stale inline comment ("empty for failed nodes") once the guard exists.

## Source

Surfaced during independent review of PR #2710 (`/prp-review`, three parallel reviewer agents — `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier` — independently converged on this). See PR #2710's canonical review comment for the full evidence trail. Not blocking #2710's merge: #2710's own fix is correct, complete for the specific reproduction in #2696, and does not need to be reverted or changed; this is genuinely separate follow-up work.

Wirasm:
## Problem and outcome

PR #2710 fixed exactly one reader of a failed producer's output — the `{ from, if_skipped }` binding directive — after run `6607bf20` showed a failed `loop_group`'s stale last-completed-iteration text getting read as if the group had succeeded, flipping a PR to ready with an unreviewed commit while the workflow itself reported `failed`. Three sibling readers of the same `nodeOutputs` map had no equivalent guard: `when:` conditions, inline `$node.output` text in `prompt:`/`bash:`/`command:` bodies, and a plain (non-directive) `with:` value. Any of the three could reach the identical failure mode — a consumer joined past a failed producer via `trigger_rule: all_done`.

- **Outcome:** All four ways a workflow can read a producer's output — the binding directive (already fixed), `when:`, inline substitution, and plain `with:` — now throw uniformly when the referenced producer's `state` is `'failed'`, whether the reference is fielded (`$node.output.field`) or unfielded (`$node.output`).
- **Invariant:** No reader of `nodeOutputs` returns a real value for a producer whose `state` is `'failed'`, regardless of how JSON-shaped or non-empty its stored output looks.
- **Scope boundary:** No change to `resolveBindingDirective`, `checkTriggerRule`'s `all_done` semantics, or how a `loop_group`'s failure paths compute their stored output.
- **Root cause:** `condition-evaluator.ts`'s `resolveOutputRef`, `dag-executor.ts`'s `substituteNodeOutputRefs`, and `resolveWorkflowValue`'s whole-ref tier each read `nodeOutput.output`/`structuredOutput` unconditionally, with no `state` check — the exact gap #2710 closed for the binding directive but left open here.

## Review guidance

- **Feedback requested:** Whether the fielded form's consolidation into the shared `resolveNodeOutputField` (one guard, reused by all three readers) versus three separate unfielded guards (one per caller, worded for that caller) draws the right line; whether the two pre-existing tests I updated are a legitimate correction rather than weakening coverage.
- **Start here:** `packages/workflows/src/output-ref.ts` — the new `state === 'failed'` branch in `resolveNodeOutputField`, which fixes the fielded form for every sibling reader at once.
- **Review order:** `output-ref.ts` (shared fielded guard) → `dag-executor.ts` (`wholeRefLogicalValue`'s unfielded guard, `substituteNodeOutputRefs`'s unfielded guard) → `condition-evaluator.ts` (`resolveOutputRef`'s unfielded guard + corrected stale comment) → the three test files.
- **Lower-attention areas:** The two corrected pre-existing tests in `dag-executor.test.ts`'s `#2402` describe block — mechanical fixture overrides, explained inline and in the PR body below.
- **Known risk or uncertainty:** This branch is a few commits behind current `dev` (a large `dag-executor.ts` discriminated-union refactor landed after this branch was created). No conflict expected — the touched functions are untouched by that refactor per a direct diff check — but GitHub's mergeability check is the final word.

## Solution

`resolveNodeOutputField` (`output-ref.ts`) already centralizes fielded-reference resolution for every caller, with an established `state === 'skipped' || state === 'pending'` guard. Adding a parallel `state === 'failed'` branch closes the fielded-reference gap for all three sibling readers in one place, since `wholeRefLogicalValue`, `condition-evaluator`'s `resolveOutputRef`, and `substituteNodeOutputRefs` all route their `.field` form through it already.

The unfielded (whole-text) form has no shared function, so each of the three call sites gets its own local guard, worded for that caller:

- `wholeRefLogicalValue` (`dag-executor.ts`) — the dispatcher already shared by `resolveBindingDirective` (protected since #2710) and `resolveWorkflowValue`'s whole-ref tier (unprotected until now). Adding the guard here fixes the plain `with:` value case (script/command node `with:`, `workflow:` node `with:`, `fan_out` static `with:`) with **zero changes to `resolveWorkflowValue` itself** — it already calls `wholeRefLogicalValue` unconditionally, and the new branch is unreachable from `resolveBindingDirective`'s call since that function already throws its own message before ever reaching it.
- `condition-evaluator.ts`'s `resolveOutputRef` — the `when:` resolver, plus its own stale inline comment/JSDoc ("empty for failed nodes") corrected to state the new throw behavior.
- `dag-executor.ts`'s `substituteNodeOutputRefs` — the inline `$node.output` text resolver, covering every `prompt:`/`bash:`/`command:` body, `workflow:` node `input:`, approval messages, and fan-out `items` in one place.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A `when:` condition, inline prompt/bash text, or plain `with:` value referencing a failed producer resolved its stored output (real, non-empty text for a failed `loop_group`) as if the producer had succeeded. | All four readers throw instead, naming the failed producer. |
| **Failure behavior** | Silent: the consumer ran normally on stale data; only the unrelated failed producer showed as failed. | Explicit: the consuming node itself fails, with an error naming the failed producer. |

## Validation

- `bun test src/output-ref.test.ts` (packages/workflows) — 46 pass, 0 fail — proves the shared fielded guard, including a failed producer whose leftover output is valid JSON (mirrors a `loop_group`'s `lastIterationOutput`).
- `bun test src/condition-evaluator.test.ts` — 96 pass, 0 fail — rewrote the one pre-existing test whose premise ("failed nodes always have empty output") is exactly what #2696/#2710 disproved for `loop_group`; added the issue's own reproduction (`$corrections.output.ready == true` on a failed producer with valid JSON leftover output).
- `bun test src/dag-executor.test.ts` (full file) — 598 pass, 0 fail. Two pre-existing tests in the `#2402` (stale-cache-invalidation) describe block needed a fixture override: their shared `consumer.md` command file inlines `$producer.output`, and both tests deliberately make `producer` fail — exactly the shape `substituteNodeOutputRefs`'s new guard now rejects. That's the intended new behavior (inlining a failed producer's output), not a regression in the `#2402` staleness-invalidation logic those tests actually exercise, so I overrode the fixture in just those two tests to keep them measuring only what they were written for.
- `bun --filter @archon/workflows test` — every constituent suite 0 fail.
- `bun run validate` — exit code 0. Type-check, lint (`--max-warnings 0`), format, generated-file checks (bundled defaults/skill/schema, Pi vendor map, capability matrix), and the full per-package test suite all passed.
- **Not verified:** A real-run reproduction of the exact `6607bf20` shape through all three sibling paths (only the binding-directive path was reproduced live for #2696) — outside this PR's scope, which is engine-level coverage; the new regression tests directly exercise the diagnosed code paths with the same failure shape (real, valid-JSON leftover output from a failed `loop_group`).

Closes #2713
Plan: https://github.com/coleam00/Archon/issues/2713#issuecomment-5380798162

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented workflow steps from using stale output from failed steps.
  * References to failed step output now stop dependent steps with a clear failure message.
  * Added safeguards for complete outputs and individual fields, including leftover JSON data.
  * Improved condition evaluation so failed outputs cannot affect `when:` decisions.
* **Documentation**
  * Clarified how failed-step outputs behave in conditions and substitutions.


Wirasm:
## Ready to merge

Correction verification bounded against the previously reviewed head `d04a9b3a` (READY TO MERGE, 0 blocking, 3 Suggestions FIXED, 1 Suggestion DECLINED). Two commits landed since: a merge commit `9dfa30f9` pulling `origin/dev` — which had independently advanced with an unrelated large refactor (#2716, "Node as a discriminated union by body kind") plus other unrelated commits — into this branch, and a one-file follow-up `0498e577` (current head) fixing the single TypeScript type-check failure that merge surfaced in this PR's own newly-added test. Both the coordinator and two independently dispatched reviewer agents (`code`, `types`) verified from first principles, not from any prior summary: diffing dev-tip against the merge result for this PR's 8 owned files and comparing it content-for-content against the PR's original pre-merge diff shows **zero non-whitespace difference** — the only deltas are blob hashes and hunk line-number offsets fully explained by the unrelated refactor growing the surrounding files. The merge carried this PR's logic through untouched. The follow-up commit touches exactly one file and two lines (`const nodes: DagNode[] = [...]` → `const nodes = [...]` with the cast relocated to the call site, `nodes: nodes as DagNode[]`), inside the PR's own `"sibling gaps (#2713)"` test — a mechanical fix for a real inference break caused by #2716 turning `DagNode` into a discriminated union (its transform's return type has always been `DagNode | IncludeDirective`; annotating the array literal directly no longer type-checks each element against the new union, while inferring first and casting at the call site does). It matches, verbatim, the idiom already used at four other call sites in the same `describe` block, changes no runtime value, and both node literals in this test have no `include:` field, so the transform provably returns a real `DagNode` — the type-design agent confirmed the relocated cast narrows an unreachable case, not a live one, so no compile-time or runtime safety is lost. All four original findings (R1–R4) live in files this correction never touches (`condition-evaluator.ts`, `output-ref.ts`, `workflow-run.ts`, `authoring-workflows.md` — confirmed byte-identical across both commits) and remain in their prior, already-verified dispositions. No reviewer found a new defect introduced by either commit.

**0 blocking · 0 non-blocking** (1 additional finding remains declined, no action required)

**Validation:** Local re-run in this review's own pinned worktree at the corrected head, plus GitHub Actions CI on the exact head — all green, including `test (ubuntu-latest)` and `test (windows-latest)`, which failed before this fix and now pass. See the table below.

**Resolved:** 3 (R1, R2, R3, all from the prior round — unaffected and re-confirmed unchanged this round)

### Findings

|---|---|---|---|
| `R1` | Suggestion | Comments in 3 files still assert the pre-fix premise this PR disproves | FIXED |
| `R2` | Suggestion | `authoring-workflows.md` doesn't document the new failed-producer throw for `when:`/inline substitution, and one existing paragraph is now factually wrong | FIXED |
| `R3` | Suggestion | Plan undercounts identical guards as 3 when a 4th (pre-existing) instance exists; a "KEEP IN SYNC" tag would cut future drift risk at near-zero cost | FIXED |
| `R4` | Suggestion | Shared `#2402` test fixture is fragile for any future test in the same block that fails `producer` without overriding it | DECLINED |

<details>
<summary><code>R1</code> — Stale comments assert failed-node output is always empty / whole-text refs never throw</summary>

**Impact:** A maintainer reading the module-level doc comment in `condition-evaluator.ts` or the sibling comment in `output-ref.ts` — both a few lines above the code this PR changed — would read a claim that was false, and could reintroduce the bug this PR fixes by trusting it. Not a runtime defect; a correctness trap for future edits.

**Evidence:** `packages/workflows/src/condition-evaluator.ts`, `packages/workflows/src/output-ref.ts`, `packages/workflows/src/schemas/workflow-run.ts` — all three confirmed byte-identical between `d04a9b3a` (previously verified FIXED) and the current head `0498e577`; the merge and follow-up commit never touch these files.

**Required outcome:** Met — no further action.

**Found by:** `prp-core:code-reviewer` (workflow-run.ts), `prp-core:seam-analyzer` (condition-evaluator.ts, output-ref.ts)

**Disposition:** FIXED — verified against `d04a9b3a` in the prior round; re-confirmed unaffected at `0498e577` this round (files untouched by either new commit).

</details>

<details>
<summary><code>R2</code> — authoring-workflows.md doesn't cover the new when:/substitution behavior, and one paragraph is now wrong</summary>

**Impact:** A workflow author reading the `### when: Condition Syntax` or `### $node_id.output Substitution` sections got no warning that referencing a failed producer now throws — they'd learn it only by hitting the runtime error. The #2710-added paragraph in the binding-directive section stated as settled fact something this PR's fix disproves.

**Evidence:** `packages/docs-web/src/content/docs/guides/authoring-workflows.md` confirmed byte-identical between `d04a9b3a` and `0498e577`.

**Required outcome:** Met — no further action.

**Disposition:** FIXED — verified against `d04a9b3a` in the prior round; re-confirmed unaffected at `0498e577` this round (file untouched by either new commit).

</details>

<details>
<summary><code>R3</code> — Precedent count is 4, not 3; a lightweight sync marker would reduce future drift risk</summary>

**Impact:** None functionally — the decision to keep separate per-caller unfielded guards instead of one parameterized helper was independently re-verified as correct.

**Evidence:** "KEEP IN SYNC" comments at all four `state === 'failed'` guard sites, confirmed present and unchanged in `dag-executor.ts` and `condition-evaluator.ts` across the merge (both files independently byte-compared for the PR's own contribution; see the Signal paragraph and validation section).

**Required outcome:** Met (optional hygiene improvement) — no further action.

**Disposition:** FIXED — verified against `d04a9b3a` in the prior round; re-confirmed unaffected at `0498e577` this round.

</details>

<details>
<summary><code>R4</code> — Shared #2402 test fixture is fragile for a hypothetical future test</summary>

**Impact:** A hypothetical FUTURE test added to the same `describe` block, which fails `producer` without knowing to override the shared `consumer.md` fixture, would hit this PR's new guard instead of whatever it meant to test — a latent trap for a future test author, not a defect in this PR.

**Evidence:** `packages/workflows/src/dag-executor.test.ts` — the two `#2402` fixture overrides this finding concerns are untouched by either new commit (confirmed via content-level diff comparison this round); the correction commit `0498e577` touches a different test entirely (the `#2713` sibling-gaps test, ~830 lines later in the file).

**Required outcome:** None. The reviewer that raised it explicitly said no change is needed today.

**Disposition:** DECLINED — unchanged; speculative future-test risk with no current instance, and neither new commit touches this area.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | Re-run this round, scoped to the bounded correction diff (`d04a9b3a`→`0498e577`). Independently confirmed the merge introduced zero unrelated/conflicting change to the PR's own 8 files (byte-identical content diff), and the follow-up commit is exactly the claimed narrow, pattern-matching fix. No new finding. |
| `types` (`prp-core:type-design-analyzer`) | Re-run this round, scoped to the follow-up commit's type-narrowing change. Confirmed the relocated cast narrows an unreachable case (`IncludeDirective`) for this test's concrete node literals, not a live one — no compile-time or runtime safety lost. No new finding. |
| `seams` (`prp-core:seam-analyzer`) | Not re-run this round — the correction diff is confined to one test file with no missing-type, counterpart-drift, or bypassed-boundary surface; R1/R4's owning files are confirmed untouched. |
| `simplify` (`prp-core:code-simplifier`) | Not re-run this round — the correction diff introduces no new machinery (it removes a redundant type annotation, following an established idiom); R2/R3's owning files are confirmed untouched. |

#### Validation

|---|---|---|
| `bun --filter @archon/workflows type-check` | PASS | exit code 0, run fresh in this review's pinned worktree at `0498e577` — this is the exact check the follow-up commit fixes |
| `bun test src/dag-executor.test.ts` (full file, packages/workflows) | PASS | 598 pass, 0 fail — includes the corrected `#2713` sibling-gaps test, run fresh at `0498e577` |
| `bun test src/condition-evaluator.test.ts src/output-ref.test.ts` (packages/workflows) | PASS | 142 pass, 0 fail (96 + 46) — run fresh at `0498e577`, files unchanged from prior round |
| `bun x prettier --check packages/workflows/src/dag-executor.test.ts` | PASS | uses Prettier code style |
| `bun x eslint packages/workflows/src/dag-executor.test.ts --max-warnings 0` | NOT APPLICABLE | file is ESLint-ignored (test files excluded from lint in this package) |
| GitHub Actions CI at head `0498e577` | PASS | `Build docs site`, `CodeRabbit`, `docker-build`, `schema-upgrade`, `test (ubuntu-latest)`, `test (windows-latest)` all `success`, confirmed bound to `head_sha: 0498e577` via the GitHub API — the two `test` jobs failed before this fix and now pass |

</details>

Issue #1192, opened 2026-04-13
Multiple local clones of the same remote cannot be registered as separate projects


halindrome:
## Summary

Archon derives codebase identity from the remote URL (`owner/repo`). When a user has multiple local clones of the same remote — for example a production-patches checkout and a development checkout — `registerRepoAtPath` returns the same `codebase_id` for both because `findCodebaseByName` matches on the remote-derived name.

This makes it impossible to:
- Register both clones as separate projects in the Web UI or via `/register-project`
- Have per-project env vars, commands, or workflow configurations scoped to each clone
- Run workflows from different clones without them sharing isolation environments, conversations, and session state

## Root Cause

`registerRepoAtPath()` in `packages/core/src/handlers/clone.ts:40-160`:

1. Line 63: `codebaseDb.findCodebaseByName(name)` — queries `remote_agent_codebases WHERE name = $1`
2. When a record with that name exists (regardless of which local path created it), the existing record's `id` is returned
3. `default_cwd` is only updated if the new path is local and the existing one is archon-managed (lines 70-78)
4. For local-only repos (no remote), the name falls back to the directory basename — two directories with the same basename also collide

The `codebases` table uses `name` as the uniqueness key, not the local filesystem path.

## Use Case

A developer has:
- `~/production/myapp` — checked out at the latest release tag, used for hotfixes
- `~/development/myapp` — on the `dev` branch, used for feature work

Both point to `git@github.com:org/myapp.git`. The developer wants to:
1. Register both in Archon with separate configurations (different env vars, different base branches)
2. Run workflows from each directory independently
3. Have conversations and sessions scoped to each checkout

Currently, registering the second clone silently reuses the first clone's `codebase_id`, merging all state.

## Suggested Approach

The codebase identity should incorporate the local filesystem path, not just the remote URL. Options:

**Option A: Composite uniqueness key** — Change the `codebases` table uniqueness constraint from `name` alone to `(name, default_cwd)`. This allows two registrations of `org/myapp` with different local paths. Display names could use `org/myapp (production)` and `org/myapp (development)`.

**Option B: Path-qualified name** — When a name collision is detected and the paths differ, auto-suffix the name: `org/myapp` and `org/myapp~development`. The first registration keeps the clean name.

**Option C: Explicit alias** — Allow users to provide a custom name/alias at registration time that overrides the remote-derived name. The Web UI and `/register-project` would accept an optional name parameter.

[...]

PR #2720, opened 2026-08-22
fix(workflows): a faulted codebase lookup no longer pins a run to an empty fallback location

### LINKED ISSUE #2304: A DB blip during codebase lookup permanently pins a run's output_root to the _cwd fallback
> ### ⚠️ CORRECTED 2026-08-05 — the code snippet below is stale
>
> **The retry landed** in #2299 (`42690a89`) — two attempts, zero delay, then the same `_cwd` fallback. Strike "retry the lookup before falling back" from the options as DONE.
>
> **Acceptance boxes 1, 2 and 3 remain open.** The retry lowers the probability of box 1; it does not satisfy it. The two degradations now log distinctly, which is partial credit on box 3 — but in **logs**, not on the row, which is what box 3 asks for.
>
> Worth knowing: the retry is weak on the default install. SQLite's `PRAGMA busy_timeout = 5000` means a `SQLITE_BUSY` throw is by construction *not* transient. It earns its place on Postgres. That strengthens the case for the remaining half.

---

## Problem

A transient database fault during a run's codebase lookup silently relocates that run's output — permanently.

`resolveProjectPaths` (`packages/workflows/src/executor.ts`) catches a failing `getCodebase` and falls through to the `_cwd/<basename>` pseudo-project:

```ts
try {
  const codebase = await deps.store.getCodebase(codebaseId);
  if (codebase) { key = resolveProjectStorageKey(codebase, cwd); ... }
} catch (error) {
  getLog().error({ err, codebaseId, cwd }, 'project_paths_resolve_failed_using_fallback');
}
return composeRunPaths(getProjectStoragePaths(key ?? { kind: 'cwd', cwd }), workflowRunId);
```

That fallback is correct as a *last resort* — a run should not die because the registry blipped. The problem is what happens next:

1. The run's artifacts, logs, and **`$STATE_DIR`** resolve under `_cwd/<basename>` instead of the real project.
2. `output_root` is persisted with that value, and it is **write-once** (`COALESCE(output_root, $N)`, #2200), so it is pinned for the run's entire life. No resume re-derives it.
3. Nothing distinguishes this from a legitimately unregistered run. Both are valid locations; both look identical on disk and in the row.

The `$STATE_DIR` half is the sharp edge: a stateful workflow silently reads and writes a *different, empty* state directory for that run. For the maintainer triage workflows that is the same class of outcome #2200's `state-preflight` gate exists to prevent — except the gate cannot catch it, because the legacy directory is irrelevant and `_cwd/<basename>/state` genuinely is empty for a first run.

## Why this was not patched in #2299

It needs a decision, not a fix, and it is unreachable without a DB fault:

- **Fail the run.** Honest, and consistent with Fail Fast — but it converts a transient blip into a failed run, and the fallback exists precisely because that was judged the wrong trade.
- **Accept the fallback but do not persist `output_root`.** The run still writes to the fallback location, but the row stays `NULL` so later reads re-derive and a resume can recover. Cheap; leaves the run's own output split from where readers look.
- **Retry the lookup before falling back.** Addresses the transient case specifically, which is the only case that produces this. Probably the smallest correct change.
- **Mark the degradation on the row** (e.g. `metadata.identity_unresolved: true`) so the console and the gate can distinguish "unregistered" from "we could not tell".

My weak preference is retry-then-fallback plus not persisting `output_root` when identity was unresolved — it keeps the run alive, avoids pinning a location derived from a fault, and lets a resume self-heal. But that is a judgement call about how much a transient registry failure should be allowed to change, and it interacts with the write-once invariant that #2299 deliberately introduced.

## Acceptance

- [ ] A transient `getCodebase` failure does not permanently pin a run's `output_root` to a fault-derived location
- [ ] A stateful workflow cannot silently read an empty `$STATE_DIR` because of a registry blip
- [ ] "Unregistered project" and "identity could not be resolved" are distinguishable after the fact
- [ ] Whatever is chosen, the run does not die on a transient fault (that trade is already settled)

Found in review of #2299 (S18). Related: #2200, #1192.

Wirasm:
## Problem and outcome

A transient `getCodebase` failure inside `resolveProjectPaths` drops a run onto the `_cwd/<basename>` fallback. Today the persistence block then writes that fallback location as the run's `output_root`, which is write-once via `COALESCE`, so the fault-derived location is pinned for the run's whole life — including `$STATE_DIR`, so a stateful workflow silently reads an empty state directory. The fix in #2299 / `42690a89` lowered the probability of the underlying fault but did not satisfy the open acceptance boxes from #2304; the row's `output_root` was still pinned on either retry attempt that hit a fault.

- **Outcome:** A faulted lookup no longer writes `output_root` at all — the column stays NULL — and stamps `metadata.identity_unresolved = true` instead, so a later resume self-heals from a now-healthy registry and any reader (the state-preflight gate from #2200, maintainer triage) can tell "unregistered" apart from "we could not tell".
- **Invariant:** The `output_root` write-once guard is preserved for every non-faulted case (so the #1192 rename hazard stays scoped to rows with real roots); a faulted run whose `output_root` is still NULL on its next successful lookup will write that now-correct root exactly once.
- **Scope boundary:** The existing retry loop's shape is untouched; the existing `_cwd` fallback still runs (a fault does not kill a run); legitimate unregistered rows (`key.kind === 'cwd'` with a successful lookup) still persist `output_root` exactly as before. Wiring the new flag into #2200's gate or the maintainer-triage workflows is a follow-up — readers of the flag live where the issue says they do.
- **Root cause:** The persistence block in `executeWorkflow` branched only on `!workflowRun.output_root`, with no signal for whether the resolved location came from a successful lookup or from a fault. A successful-cwd-key fallback wrote `output_root` legitimately; a faulted-cwd-key fallback wrote it by accident, and write-once made the accident permanent.

## Review guidance

- **Feedback requested:** Correctness of the three-way `identityResolution` taxonomy (does collapsing `'resolved'` / `'unregistered'` / `'faulted'` into any two lose information that matters?); whether the `'faulted'` branch's "stamp the flag, skip the write" shape is the right read of `updateWorkflowRun`'s COALESCE-backed write-once contract.
- **Start here:** `packages/workflows/src/executor.ts:2076-2113` — the persistence block in `executeWorkflow` is the load-bearing change; everything else is plumbing.
- **Review order:** `packages/workflows/src/executor.ts` (`ResolvedProjectPaths.identityResolution` interface, then `resolveProjectPaths`'s three branches in the retry loop, then the persistence block) → `packages/workflows/src/schemas/workflow-run.ts` (`RUN_METADATA_KEYS` + `readIdentityUnresolved`) → `packages/workflows/src/executor.test.ts` (resolve-time + persist-time tests) → `packages/workflows/src/schemas.test.ts` (reader round-trip).
- **Lower-attention areas:** `packages/workflows/src/schemas/index.ts` is a pure re-export.
- **Known risk or uncertainty:** None outstanding — see Validation. The "resume self-heals" half is asserted structurally by the unchanged `!workflowRun.output_root` guard + the now-correct resolve-time `'resolved'` branch; the implementation.md explains why it was not separately driven through two `executeWorkflow` invocations.

## Solution

Thread an `identityResolution: 'resolved' | 'unregistered' | 'faulted'` flag through `resolveProjectPaths`; branch the persistence block on it.

- `resolveProjectPaths` sets the flag inside the existing two-attempt retry loop, not after it:
  - on a successful lookup whose derived key is repo / folder / `_local` → `'resolved'`;
  - on a successful lookup whose derived key is `cwd` (the codebase exists but has no owner/repo or `_local` identity — the existing `codebase_project_identity_unresolved` WARN) → `'unregistered'`;
  - after both attempts throw → `'faulted'`;
  - on the `codebaseId == null` or `getCodebase → null` arm (no lookup, or no row) → `'unregistered'`;
  - on the persisted-`output_root` short-circuit (a resume reading an existing root) → `undefined` — nothing to flag.
- `executeWorkflow`'s `if (!workflowRun.output_root)` block now branches:
  - `'faulted'` → **skip** the `output_root` write, write `{ metadata: { identity_unresolved: true } }` instead, and emit `workflow.output_root_not_persisted_identity_faulted` at WARN. A subsequent successful resume reads `output_root === null`, enters the same persistence block, this time with `identityResolution: 'resolved'`, and writes the now-correct root once.
  - `'resolved'` / `'unregistered'` / `undefined` → write `output_root` exactly as before for a row that was never flagged. For a previously-faulted row whose `metadata.identity_unresolved` is still `true` when the resume arrives, the same atomic write rides `metadata: { identity_unresolved: false }` so the cleared flag and the new root land together (the heal half — without it the DB's `jsonMerge` keeps the stale `true` and a row that has resolved would continue to read as faulted to #2200's gate and the maintainer-triage workflows). The `output_root` write-once guard continues to protect every legitimate row from the #1192 rename hazard.
- `RUN_METADATA_KEYS.identityUnresolved = 'identity_unresolved'` and `readIdentityUnresolved(metadata): boolean | undefined` live in `schemas/workflow-run.ts`, mirroring the existing `SUBRUN_METADATA_KEYS` / `readSubrunMetadata` precedent (`Record<string, unknown>` JSONB column, shared symbol between writer and reader, defensive non-boolean-reads-as-`undefined` reader). Both are re-exported from `@archon/workflows/schemas` so #2200's gate and the maintainer-triage workflows can consume them when those surfaces pick up the flag.

This keeps the change to "stop persisting on one arm of the resolver; add one row-level marker" — no new DB columns, no new run states, no change to the retry's shape.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A run whose `getCodebase` throws on both attempts has `output_root` populated with `~/.archon/workspaces/_cwd/<basename>/`, and that path is pinned for the run's whole life. A stateful workflow reads an empty `$STATE_DIR` for the remainder of that run. | That run has `output_root` left NULL and `metadata.identity_unresolved = true` stamped instead. A later resume self-heals to the correct root once the registry is healthy. |
| **Failure behavior** | Indistinguishable from a legitimately unregistered run; the maintainer triage workflows and #2200's state-preflight gate cannot tell a faulted run from a healthy one that happens to be on the cwd fallback. | `readIdentityUnresolved(metadata)` returns `true` on a faulted row, `false` on a row that has healed, and `undefined` on every other row; readers can branch on it without speculating about why `output_root` is null. |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[resolveProjectPaths] --> B2{getCodebase throws twice?}
    B2 -- yes --> B3[Faulted cwd fallback, persisted as output_root]
    B3 --> B4[output_root pinned for run's whole life]
  end

  subgraph After
    A1[resolveProjectPaths] --> A2{getCodebase throws twice?}
    A2 -- yes --> A3[output_root stays NULL, metadata.identity_unresolved = true]
    A3 --> A4[Later resume: registry healthy, root written once]
  end
```

## Architecture

```mermaid
flowchart LR
  A[executeWorkflow: !workflowRun.output_root] --> B{identityResolution}
  B -- faulted --> C[stamp identity_unresolved, skip output_root]
  B -- resolved / unregistered / undefined --> D[existing output_root write]
  C -.resume after registry heals.-> A
  E[resolveProjectPaths retry loop] ==> B
  E -- three states, set inside the loop.--> F[ResolvedProjectPaths.identityResolution]
```

### Changed seams

|---|---|---|
| `resolveProjectPaths → executeWorkflow` | New `identityResolution` field on the `ResolvedProjectPaths` return shape (`undefined` on the persisted-`output_root` short-circuit). | `packages/workflows/src/executor.ts:309-326` (type) and `:435-447` (population). |
| `executeWorkflow → IWorkflowStore.updateWorkflowRun` | The `'faulted'` arm writes `{ metadata: { identity_unresolved: true } }` and skips `output_root`. The else arm now also writes `metadata.identity_unresolved = false` when the row was previously flagged, so a healed row stops reading as faulted; rows that were never flagged still see a plain `output_root` write. | `packages/workflows/src/executor.ts:2076-2113`; `executor.test.ts` "output_root persistence branching (#2304)" describe block. |
| `workflow_run.metadata` JSONB column | Adds `identity_unresolved: true` as a typed value, parallel to `SUBRUN_METADATA_KEYS` (`packages/workflows/src/schemas/workflow-run.ts:247-277`). Reader (`readIdentityUnresolved`) returns `boolean | undefined` and treats non-boolean values as `undefined`. | `schemas.test.ts` "readIdentityUnresolved (#2304)" describe block (4 cases). |
| Re-exports (`@archon/workflows/schemas`) | `RUN_METADATA_KEYS` and `readIdentityUnresolved` are re-exported alongside `SUBRUN_METADATA_KEYS` and `readSubrunMetadata` so #2200's state-preflight gate and the maintainer-triage workflows can consume them. | `packages/workflows/src/schemas/index.ts:146-147`. |

## Validation

- `bun --filter @archon/workflows type-check` — exit 0.
- `bun --filter @archon/workflows test` — 22 separate `bun test` invocations across the package, every one green. The new `identityResolution (#2304)` (7 cases) and `output_root persistence branching (#2304)` (4 cases) describe blocks in `executor.test.ts` pass, as does `readIdentityUnresolved (#2304)` (4 cases) in `schemas.test.ts`.
- `bun --filter '*' type-check` — every workspace type-checks clean.
- `bun run format:check` — "All matched files use Prettier code style!"
- `bun run lint` — no warnings, no errors.
- **Not verified end-to-end manually:** No live `bun run cli workflow run` against a deliberately faulting `getCodebase` — the unit tests drive `executeWorkflow` end-to-end with a mocked store and assert against the `updateWorkflowRun` calls, and the existing `resolveProjectPaths` tests already prove the success-arm paths return a real `outputRoot`. Postgres path (`docker-compose --profile with-db`) — not run in this environment.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. The flag is a new metadata key; existing rows read as `undefined`, identical to absent. The `output_root` write-once invariant is preserved for every non-faulted row. | `readIdentityUnresolved` returns `undefined` on missing key; the persistence block's `'resolved'` / `'unregistered'` / `undefined` arms leave `metadata` untouched unless the row was previously flagged as faulted, in which case the same atomic write carries `identity_unresolved = false`. |
| Security / permissions / data | None. The flag is local to a workflow-run row; no permissions, no cross-tenant surface. | `RUN_METADATA_KEYS.identityUnresolved` is private to `@archon/workflows/schemas`; nothing else writes the column. |
| Rollout / rollback | `git revert` of `68bed1c5 a31ab562` restores the pre-fix behaviour; in-flight faulted runs from before the revert keep the metadata stamp and continue to read as faulted (the reader returns `undefined` for non-boolean and absent values, never coerces). | Two commits; no schema migration. |
| Observability | New WARN `workflow.output_root_not_persisted_identity_faulted` and `workflow.identity_unresolved_flag_persist_failed` log events; existing `workflow.output_root_persist_failed` continues to fire on the non-faulted persistence path. | `packages/workflows/src/executor.ts:2087-2095`. |
| Documentation / communication | The acceptance boxes in #2304 close with this PR (#2200's gate and the maintainer-triage workflows are follow-ups). | Issue body, items 1-4. |

## Links

- Closes #2304
- Related #1192, #2200

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow handling when project identity cannot be resolved after repeated lookup failures.
  * Prevented incomplete runs from recording an incorrect output location.
  * Automatically clears the unresolved-identity status after successful recovery.
  * Preserved existing metadata when no identity issue is present.

* **Tests**
  * Added coverage for resolved, unregistered, faulted, and resumed workflow scenarios.
  * Added validation for safely reading identity-resolution metadata.

### DIFF
--- a/packages/workflows/src/executor.test.ts
+++ b/packages/workflows/src/executor.test.ts
@@ -162,7 +162,7 @@ import { keepAwake } from './utils/keep-awake';
 import type { WorkflowDeps, IWorkflowPlatform, WorkflowConfig } from './deps';
 import type { IWorkflowStore } from './store';
 import type { WorkflowDefinition, WorkflowRun, WorkflowRunNodeSession } from './schemas';
-import { workflowDefinitionSchema } from './schemas';
+import { RUN_METADATA_KEYS, workflowDefinitionSchema } from './schemas';
 
 // --- Helpers ---
 
@@ -1782,6 +1782,95 @@ describe('executeWorkflow', () => {
       expect(msg).toContain('Wait for it to finish');
     });
   });
+
+  // #2304 — the persistence block must distinguish three `identityResolution` outcomes:
+  //   • 'faulted'      → skip the `output_root` write, stamp the metadata flag
+  //                      (`metadata.identity_unresolved = true`). The row keeps
+  //                      `output_root` NULL so a later resume can self-heal.
+  //   • 'unregistered' → cwd fallback IS the correct location for this row;
+  //                      persist `output_root` exactly as before.
+  //   • 'resolved'     → repo/folder/_local key; persist `output_root` as before.
+  describe('output_root persistence branching (#2304)', () => {
+    it('does NOT write output_root when identityResolution is "faulted"; stamps the metadata flag instead', async () => {
+      const updateSpy = mock(async () => {});
+      const store = makeStore({
+        // Both attempts throw — the retry yields the same fault.
+        getCodebase: mock(async () => {
+          throw new Error('db down');
+        }),
+        updateWorkflowRun: updateSpy,
+      });
+      const deps = makeDeps(store);
+
+      await executeWorkflow(
+        deps,
+        makePlatform(),
+        'conv-1',
+        '/repos/widget',
+        makeWorkflow(),
+        'test',
+        'db-conv-1',
+        { codebaseId: 'cb-boom' }
+      );
+
+      const writesWithOutputRoot = updateSpy.mock.calls.filter(
+        (call: unknown[]) => typeof (call[1] as { output_root?: unknown })?.output_root === 'string'
+      );
+      expect(writesWithOutputRoot).toHaveLength(0);
+
+      const flagWrite = updateSpy.mock.calls.find(
+        (call: unknown[]) =>
+          (call[1] as { metadata?: Record<string, unknown> })?.metadata?.[
+            RUN_METADATA_KEYS.identityUnresolved
+          ] === true
+      );
+      expect(flagWrite).toBeDefined();
+      // The patch must not also write `output_root` — a NULL `output_root`
+      // means a later resume can still self-heal.
+      const patch = (flagWrite as unknown as [unknown, { output_root?: unknown }] | undefined)?.[1];
+      expect(patch?.output_root).toBeUndefined();
+    });
+
+    it('persists output_root when identityResolution is "resolved" (a registered codebase)', async () => {
+      const updateSpy = mock(async () => {});
+      const store = makeStore({
+        getCodebase: mock(async () => ({
+          id: 'cb-repo',
+          name: 'acme/widget',
+          repository_url: 'https://github.com/acme/widget',
+          default_cwd: '/repos/widget',
+          kind: 'repo' as const,
+        })),
+        updateWorkflowRun: updateSpy,
+      });
+      const deps = makeDeps(store);
+
+      await executeWorkflow(
+        deps,
+        makePlatform(),
+        'conv-1',
+        '/repos/widget',
+        makeWorkflow(),
+        'test',
+        'db-conv-1',
+        { codebaseId: 'cb-repo' }
+      );
+
+      const write = updateSpy.mock.calls.find(
+        (call: unknown[]) => typeof (call[1] as { output_root?: unknown })?.output_root === 'string'
+      );
+      expect(write).toBeDefined();
+      // The metadata flag must NOT be stamped on a resolved row — that flag is the
+      // faulted-only signal.
+      const flagWrite = updateSpy.mock.calls.find(
+        (call: unknown[]) =>
+          (call[1] as { metadata?: Record<string, unknown> })?.metadata?.[
+            RUN_METADATA_KEYS.identityUnresolved
+          ] !== undefined
+      );
+      expect(flagWrite).toBeUndefined();
+    });
+  });
 });
 
 describe('finally backstop', () => {
@@ -2513,6 +2602,115 @@ describe('resolveProjectPaths', () => {
     });
 
     expect(result.outputRoot).toBe(wsPath('acme', 'widget'));
+  });
+
+  // #2304: identityResolution is the row-level flag the persistence block reads to
+  // decide whether to write `output_root` or stamp `metadata.identity_unresolved`.
+  // The five cases below cover every branch in `resolveProjectPaths`:
+  //   • registered repo / folder / _local  → 'resolved'
+  //   • codebase row exists, no identity   → 'unregistered' (the WARN arm)
+  //   • no codebaseId                      → 'unregistered' (no lookup attempted)
+  //   • codebase lookup returns null       → 'unregistered' (no row)
+  //   • both attempts throw                → 'faulted' (the ERROR arm)
+  //   • persisted-output-root short-circuit → undefined (a resume, no fresh flag)
+  describe('identityResolution (#2304)', () => {
+    it('is "resolved" for a repo codebase', async () => {
+      const store = makeStore({
+        getCodebase: mock(async () => ({
+          id: 'cb-repo',
+          name: 'acme/widget',
+          repository_url: 'https://github.com/acme/widget',
+          default_cwd: '/repos/widget',
+          kind: 'repo' as const,
+        })),
+      });
+      const deps = makeDeps(store);
+
+      const result = await resolveProjectPaths(deps, '/repos/widget', RUN_ID, 'cb-repo');
+
+      expect(result.identityResolution).toBe('resolved');
+    });
+
+    it('is "resolved" for a folder codebase', async () => {
+      const store = makeStore({
+        getCodebase: mock(async () => ({
+          id: 'cb-folder',
+          name: 'My Platform',
+          repository_url: null,
+          default_cwd: '/tmp/platform',
+          kind: 'folder' as const,
+        })),
+      });
+      const deps = makeDeps(store);
+
+      const result = await resolveProjectPaths(deps, '/tmp/platform', RUN_ID, 'cb-folder');
+
+      expect(result.identityResolution).toBe('resolved');
+    });
+
+    it('is "unregistered" when the lookup returns a codebase row that resolves to a cwd key (WARN arm)', async () => {
+      // The fake resolver only falls back to { kind: 'cwd', cwd } when the basename
+      // is '.' or '..' (the corner that excludes owner/repo and `_local` derivation).
+      const store = makeStore({
+        getCodebase: mock(async () => ({
+          id: 'cb-noop',
+          name: 'orphan',
+          repository_url: null,
+          default_cwd: '/tmp/.',
+          kind: 'repo' as const,
+        })),
+      });
+      const deps = makeDeps(store);
+
+      const result = await resolveProjectPaths(deps, '/tmp/.', RUN_ID, 'cb-noop');
+
+      expect(result.identityResolution).toBe('unregistered');
+    });
+
+    it('is "unregistered" when no codebaseId is provided (no lookup attempted)', async () => {
+      const deps = makeDeps();
+
+      const result = await resolveProjectPaths(deps, '/some/cwd', RUN_ID);
+
+      expect(result.identityResolution).toBe('unregistered');
+    });
+
+    it('is "unregistered" when the lookup returns null (no codebase row)', async () => {
+      const store = makeStore({ getCodebase: mock(async () => null) });
+      const deps = makeDeps(store);
+
+      const result = await resolveProjectPaths(deps, '/some/cwd', RUN_ID, 'missing-id');
+
+      expect(result.identityResolution).toBe('unregistered');
+    });
+
+    it('is "faulted" when both getCodebase attempts throw (ERROR arm)', async () => {
+      const store = makeStore({
+        getCodebase: mock(async () => {
+          throw new Error('db down');
+        }),
+      });
+      const deps = makeDeps(store);
+
+      const result = await resolveProjectPaths(deps, '/some/cwd', RUN_ID, 'cb-boom');
+
+      expect(result.identityResolution).toBe('faulted');
+    });
+
+    it('is undefined on the persisted-output-root short-circuit (resume reading an existing row)', async () => {
+      // A resume reads its `output_root` from the row and re-derives nothing; the
+      // persistence block has nothing to flag on a branch that never resolved.
+      const store = makeStore();
+      const deps = makeDeps(store);
+
+      const result = await resolveProjectPaths(deps, '/repos/widget', RUN_ID, 'cb-repo', {
+        persistedOutputRoot: wsPath('acme', 'original'),
+      });
+
+      expect(result.identityResolution).toBeUndefined();
+      // The resolved paths still come from the persisted root, not from a fresh lookup.
+      expect(result.outputRoot).toBe(wsPath('acme', 'original'));
+    });
   });
 });
 
--- a/packages/workflows/src/executor.test.ts
+++ b/packages/workflows/src/executor.test.ts
@@ -1870,6 +1870,100 @@ describe('executeWorkflow', () => {
       );
       expect(flagWrite).toBeUndefined();
     });
+
+    // #2304 — the contract is "Cleared by the same persistence block the moment
+    // a later resume writes a real root". The faulted arm stamps `true`; the
+    // else arm MUST ride the same atomic metadata write with `false` on the
+    // heal, or a row that has resolved leaves the flag set and the
+    // state-preflight gate (#2200) / maintainer-triage read it as still faulted.
+    it('clears metadata.identity_unresolved when a faulted row heals (resolved on resume)', async () => {
+      const updateSpy = mock(async () => {});
+      const store = makeStore({
+        getCodebase: mock(async () => ({
+          id: 'cb-repo',
+          name: 'acme/widget',
+          repository_url: 'https://github.com/acme/widget',
+          default_cwd: '/repos/widget',
+          kind: 'repo' as const,
+        })),
+        updateWorkflowRun: updateSpy,
+        // Simulate a previously-faulted row whose first run left `output_root`
+        // NULL and stamped the flag. The next run arrives on a healthy registry
+        // and hits the else-arm with that pre-existing metadata.
+        createWorkflowRun: mock(async () =>
+          makeRun({
+            metadata: { [RUN_METADATA_KEYS.identityUnresolved]: true },
+          })
+        ),
+      });
+      const deps = makeDeps(store);
+
+      await executeWorkflow(
+        deps,
+        makePlatform(),
+        'conv-1',
+        '/repos/widget',
+        makeWorkflow(),
+        'test',
+        'db-conv-1',
+        { codebaseId: 'cb-repo' }
+      );
+
+      const healWrite = updateSpy.mock.calls.find(
+        (call: unknown[]) => typeof (call[1] as { output_root?: unknown })?.output_root === 'string'
+      );
+      expect(healWrite).toBeDefined();
+      const patch = (
+        healWrite as unknown as
+          | [unknown, { output_root?: unknown; metadata?: Record<string, unknown> }]
+          | undefined
+      )?.[1];
+      // The output_root write still happens — a row that has resolved
+      // gets a real root, that's the whole point of the resume.
+      expect(typeof patch?.output_root).toBe('string');
+      // AND the metadata flag rides the same write as `false`, not as
+      // the stale `true` from the faulted arm.
+      expect(patch?.metadata).toEqual({ [RUN_METADATA_KEYS.identityUnresolved]: false });
+    });
+
+    // Defensive neighbour: a row that never was faulted must not have its
+    // metadata touched by the else-arm writer. The cleared-flag stamp is
+    // conditional, not unconditional.
+    it('does not touch metadata when the row was never flagged as faulted', async () => {
+      const updateSpy = mock(async () => {});
+      const store = makeStore({
+        getCodebase: mock(async () => ({
+          id: 'cb-repo',
+          name: 'acme/widget',
+          repository_url: 'https://github.com/acme/widget',
+          default_cwd: '/repos/widget',
+          kind: 'repo' as const,
+        })),
+        updateWorkflowRun: updateSpy,
+        createWorkflowRun: mock(async () => makeRun({ metadata: {} })),
+      });
+      const deps = makeDeps(store);
+
+      await executeWorkflow(
+        deps,
+        makePlatform(),
+        'conv-1',
+        '/repos/widget',
+        makeWorkflow(),
+        'test',
+        'db-conv-1',
+        { codebaseId: 'cb-repo' }
+      );
+
+      const outputRootWrite = updateSpy.mock.calls.find(
+        (call: unknown[]) => typeof (call[1] as { output_root?: unknown })?.output_root === 'string'
+      );
+      expect(outputRootWrite).toBeDefined();
+      const patch = (
+        outputRootWrite as unknown as [unknown, { metadata?: Record<string, unknown> }] | undefined
+      )?.[1];
+      expect(patch?.metadata).toBeUndefined();
+    });
   });
 });
 
--- a/packages/workflows/src/executor.ts
+++ b/packages/workflows/src/executor.ts
@@ -28,6 +28,7 @@ import {
   isRunBlockedOnChild,
   SUBRUN_METADATA_KEYS,
   readSubrunMetadata,
+  RUN_METADATA_KEYS,
   WORKFLOW_SOURCE_METADATA_KEY,
   readWorkflowSourceState,
 } from './schemas';
@@ -305,6 +306,25 @@ export interface ResolvedProjectPaths {
   stateDir: string;
   /** The project root persisted to `workflow_runs.output_root`. */
   outputRoot: string;
+  /**
+   * How the run's project identity was determined (#2304). Three states; collapsing
+   * any two is a correctness bug:
+   *  - `'resolved'`     — `getCodebase` returned a row whose identity resolves to a
+   *                      repo / folder / `_local` key.
+   *  - `'unregistered'` — `getCodebase` returned a row, but no owner/repo nor a
+   *                      `_local` identity could be derived from it; the run still
+   *                      gets external storage, keyed on cwd (`codebase_project_identity_unresolved`
+   *                      WARN). This is a legitimate outcome of a registered codebase,
+   *                      NOT a fault.
+   *  - `'faulted'`      — `getCodebase` threw on BOTH retry attempts; the run fell
+   *                      through to the cwd fallback (`project_paths_resolve_failed_using_fallback`
+   *                      ERROR). The persistence block must NOT write `output_root`
+   *                      for these runs — see `executeWorkflow`.
+   *
+   * Undefined on the persisted-`output_root` short-circuit branch (a resume reading
+   * an existing root re-derives nothing and there is nothing to flag).
+   */
+  identityResolution?: 'resolved' | 'unregistered' | 'faulted';
 }
 
 /**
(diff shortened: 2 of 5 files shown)


Wirasm:
# Review Report — run 78ee8688-d547-479c-935c-f6445c1254d2

## Verdict

**Ready.** R1 (Important) is fixed at `a31ab562`; both enabled lenses (`code`, `seams`) delivered reports and found no new defects on the delta.

## Reviewed head SHA

`a31ab5629b0228ddb0843e309823802a87188ab8` (PR #2720, branch `archon/task-fix-1787406832993`, base `dev`).

## Findings

No open findings. R1 (carried forward) is resolved — see "Prior findings" below.

### Suggestions

None filed this round.

## Prior findings

Carried forward from the prior round (`68bed1c5`):

| ID | Severity | One-line claim | Verdict |
|---|---|---|---|
| R1 | Important | `identity_unresolved` is never cleared on heal; the schema's docstring contract is violated by the writer's else arm. | **Resolved at `a31ab562`.** |

### R1 — verification

- **Fix shape matches the prescribed smallest correction** (`packages/workflows/src/executor.ts:2097-2111`): the else arm now imports `readIdentityUnresolved` (line 32), builds `updates` of type `Parameters<typeof deps.store.updateWorkflowRun>[1]`, and conditionally appends `metadata: { [RUN_METADATA_KEYS.identityUnresolved]: false }` when `readIdentityUnresolved(workflowRun.metadata) === true`. Single `updateWorkflowRun` call carries both `output_root` and the `false` stamp.
- **Atomic merge with `output_root`** (`packages/core/src/db/workflows.ts:946`): one `UPDATE` with `metadata = jsonMerge('metadata', $N)` and `output_root = COALESCE(output_root, $M)`. Postgres (`packages/core/src/db/adapters/postgres.ts:288-290` emits `metadata || $N::jsonb`) and SQLite (`packages/core/src/db/adapters/sqlite.ts:847-849` emits `json_patch(metadata, $N)`) both MERGE — overwriting `identity_unresolved` without dropping siblings. The `false` rides the same round-trip as the faulted arm's `true`.
- **Conditional, not unconditional** — the `if (readIdentityUnresolved(...) === true)` guard means a row whose metadata is absent, `false`, or non-boolean does not have its metadata touched on the else arm. `readIdentityUnresolved` returns `boolean | undefined` (`packages/workflows/src/schemas/workflow-run.ts:272-277`); the strict `=== true` check is the only valid discrimination.
- **Docstring ↔ writer back in sync** (`packages/workflows/src/schemas/workflow-run.ts:262-265`): "Cleared by the same persistence block the moment a later resume writes a real root, so a row that heals stops reading as faulted." Writer at `executor.ts:2097-2111` matches that contract.
- **Heal-path test** (`packages/workflows/src/executor.test.ts:1873-1928`): stubs `createWorkflowRun` to return `makeRun({ metadata: { [RUN_METADATA_KEYS.identityUnresolved]: true } })` (the post-faulted-arm row state — `output_root: null` by default + flag set), runs `executeWorkflow` end-to-end with a healthy `getCodebase`, and asserts the `output_root` write patch also carries `metadata: { [RUN_METADATA_KEYS.identityUnresolved]: false }`.
- **Conditional-neighbour test** (`packages/workflows/src/executor.test.ts:1931-1967`): seeds `metadata: {}`, runs the full path, and pins `expect(patch?.metadata).toBeUndefined()` — proving the clear is conditional (a regression to unconditional `metadata: { identity_unresolved: false }` would fail loudly).
- **Local test run** — `bun test src/executor.test.ts -t "output_root persistence branching"` → 4 pass (the 2 prior + the 2 new), 0 fail, 146 ms.

### Rejected observations

None. No "could" / "might" chains survived verification; every reachable path through the delta produces the documented outcome.

## Lens roster

| Lens | Status | Notes |
|---|---|---|
| `code` | Ran, reported | `code.md`. Verified R1 against the delta: fix shape correct, atomic merge intact on both dialects, conditional-not-unconditional, two new tests cover the heal path and the never-faulted neighbour. Full test runs clean: 4/4 `output_root persistence branching`, 98/98 `executor.test.ts`, 148/148 `schemas.test.ts`. `bun run type-check` and `bun run lint` clean. **No findings.** |
| `seams` | Ran, reported | `seams.md`. Walked the writer↔schema contract across the delta; docstring and writer are back in sync; no other monotone-write metadata shape exists in the diff that the same defect could apply to. The in-memory `workflowRun.metadata` mirror asymmetry with the faulted arm is a drift point (not a defect — no in-process consumer of `readIdentityUnresolved` exists; the follow-up PR wires it into #2200's gate). **No findings.** |
| `tests` | Disabled | No file by design — scope.md sets the input false. Recorded as disabled, not as clean. |
| `errors` | Disabled | No file by design — scope.md sets the input false. Recorded as disabled, not as clean. |
| `comments` | Disabled | No file by design — scope.md sets the input false. Recorded as disabled, not as clean. |
| `types` | Disabled | No file by design — scope.md sets the input false. Recorded as disabled, not as clean. |
| `docs` | Disabled | No file by design — scope.md sets the input false. Recorded as disabled, not as clean. |
| `simplify` | Disabled | No file by design — scope.md sets the input false. Recorded as disabled, not as clean. |

Both enabled lenses delivered their reports. No lens failed to report. The "disabled" rows are recorded as disabled (no gap), not as silent clearance.

## PR comment

Edited the canonical comment on PR #2720 (the one carrying the `<!-- archon-review-report -->` marker, id `IC_kwDON2FEhs8AAAABQLpnZQ` / numeric `5380925285`) in place. Read-back via `gh api /repos/coleam00/Archon/issues/comments/5380925285` confirms the body matches the report above byte-for-byte except for a single trailing newline the server normalizes.

- URL: https://github.com/coleam00/Archon/pull/2720#issuecomment-5380925285
- Comment id: `IC_kwDON2FEhs8AAAABQLpnZQ` (numeric `5380925285`)

## Notes for the next round

- The PR can ship. R1 is closed.
- When the follow-up PR wires `readIdentityUnresolved` into #2200's state-preflight gate and the maintainer-triage workflows (out of scope here per the PR body), re-run the seam lens specifically with a focus on: (a) consumers that read `output_root` and `identity_unresolved` as a pair, and (b) the in-memory mirror asymmetry noted in `seams.md` — by then `readIdentityUnresolved` will have an in-process consumer, and the else arm should mirror the faulted arm's `workflowRun.metadata` reassignment if that consumer reads from the in-memory row before the DB round-trip resolves.
- Optional lenses that did not run this round (`tests`, `errors`, `comments`, `types`, `docs`, `simplify`) remain optional. The `tests` lens is the natural fit when the follow-up PR lands new heal-coverage tests, to audit whether they exercise the full re-derive path or just the metadata-write call.

Wirasm:
## Ready to merge

This closes #2304 correctly: a faulted `getCodebase` lookup no longer persists the fault-derived `_cwd/<basename>` fallback as `output_root`, the write-once invariant from #2200/#1192 is untouched for every legitimate row, and a previously-faulted row heals atomically (root + cleared flag in one write) the moment a later resume finds a healthy registry. Four specialist agents (code, seams, simplify, types) traced the full retry loop and persistence block against real store implementations and found no reachable defect, no bypassable write-once boundary, no metadata-merge hazard, and no invalid type state. Every finding below is a Suggestion; several agents converged independently on the same observation — the three-state `identityResolution` union has exactly one production consumer (`=== 'faulted'`) and could be a boolean — but the type-design review (the scope that owns this call) confirmed it creates no reachable invalid state and declined to treat it as a defect, matching this PR's own explicitly-flagged open question about the taxonomy.

**Validation:** `bun run type-check` (full monorepo) — PASS. `bun run lint --max-warnings 0` — PASS. `bun run format:check` — PASS. `bun --filter @archon/workflows test` (all 24 split invocations, 700+ tests including the new `identityResolution (#2304)`, `output_root persistence branching (#2304)`, and `readIdentityUnresolved (#2304)` describe blocks) — PASS, 0 fail.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Stale comment: "logs at ERROR rather than WARN" no longer describes the two new WARN-only faulted-arm logs it sits above | OPEN |
| `R2` | Suggestion | `identity_unresolved` ships with no external reader yet — this file's other metadata keys (`SUBRUN_METADATA_KEYS`, `WORKFLOW_SOURCE_METADATA_KEY`) all shipped with a live consumer | OPEN |
| `R3` | Suggestion | `identityResolution`'s three-way state has no compiler-enforced exhaustiveness; a future branch added to the retry loop that omits an explicit assignment would silently default to `'unregistered'` and reintroduce #2304 | OPEN |
| `R4` | Suggestion | The three-state union (`'resolved' \| 'unregistered' \| 'faulted'`) has exactly one production consumer distinguishing `'faulted'` from everything else; a `boolean` would express the same behavior with less surface | OPEN |
| `R5` | Suggestion | Minor test redundancy: `executor.test.ts`'s "does not touch metadata when the row was never flagged as faulted" duplicates the same branch and assertion as "persists output_root when identityResolution is resolved" | OPEN |
| `R6` | Suggestion | `workflowRun.metadata` is mutated in-memory on the faulted branch (`executor.ts:2078-2082`) mirroring the `WORKFLOW_SOURCE_METADATA_KEY` precedent, but nothing in this file reads that in-memory copy back — the mutation is inert ceremony | OPEN |

<details>
<summary><code>R1</code> — stale ERROR/WARN comment above the persistence block</summary>

**Impact:** Cosmetic only. A reader of the block comment above `if (!workflowRun.output_root)` (`packages/workflows/src/executor.ts`, around line 2071) sees "it logs at ERROR rather than WARN," which was accurate for the sole pre-existing failure log but now also sits above two new faulted-arm logs that are both WARN. No behavioral consequence.

**Evidence:** `packages/workflows/src/executor.ts:2071` (comment); the two new logs it now also covers are `workflow.identity_unresolved_flag_persist_failed` and `workflow.output_root_not_persisted_identity_faulted`, both `getLog().warn(...)`.

**Required outcome:** Scope the ERROR/WARN sentence to the specific `else`-arm log it still accurately describes, or note that the faulted-arm logs are WARN by design (the underlying fault was already logged at ERROR earlier in `resolveProjectPaths`).

**Disposition:** OPEN — non-blocking, safe to defer or fix in this PR at the author's discretion.

</details>

<details>
<summary><code>R2</code> — <code>identity_unresolved</code> flag has no reader yet</summary>

**Impact:** None today. The flag is written correctly and defensively read back within its own function, but no API route, CLI command, or workflow YAML consumes it — the PR description's cited future consumers (#2200's state-preflight gate, maintainer-triage workflows) don't exist in the tree yet. If those never land, this is permanently inert instrumentation; if they do, the flag is ready-made. This is the first metadata key in `workflow-run.ts` to ship without a live consumer (`SUBRUN_METADATA_KEYS` and `WORKFLOW_SOURCE_METADATA_KEY` both shipped alongside real readers in `dag-executor.ts`/`workflow-source.ts`).

**Evidence:** `packages/workflows/src/schemas/workflow-run.ts:264-276` (writer/reader pair); repo-wide grep for `identity_unresolved`/`readIdentityUnresolved`/`RUN_METADATA_KEYS` outside `workflow-run.ts` and `executor.ts` returns nothing; grep for `state-preflight`/`maintainer-triage` matches only this PR's own comments.

**Required outcome:** None required now — this satisfies acceptance box 3 of #2304 ("distinguishable after the fact," which a direct DB/API read already achieves) even without a wired consumer. Cite this flag when #2200's gate or a maintainer-triage workflow is built, rather than re-deriving faulted-run detection.

**Disposition:** OPEN, non-blocking. Track informally against whichever future PR builds the #2200 gate; no new issue needed since #2304's own acceptance criteria already cover this half.

</details>

<details>
<summary><code>R3</code> — no compiler-enforced exhaustiveness on the retry loop's state assignment</summary>

**Impact:** Not reachable today — every current branch of the two-attempt retry loop assigns `identityResolution` correctly, and all five branches are directly unit-tested. The risk is forward-looking: `identityResolution` is a mutable outer-scope `let` assigned at three separate call sites inside the loop, plus a post-loop `if (identityResolution === undefined) identityResolution = 'unregistered'` default. A future edit to this loop (a third retry attempt, a new caught-exception classification, an added early return) that omits an explicit assignment would silently fall through to the `'unregistered'` default — which the persistence block treats as legitimate, silently reintroducing the exact #2304 bug this PR fixes, with no WARN/ERROR at all.

**Evidence:** `packages/workflows/src/executor.ts:380` (`let identityResolution: ... | undefined`), `:406` (success arm), `:427` (fault arm), `:439-441` (post-loop default); the doc comment at `:315-321` itself states "collapsing any two is a correctness bug," enforced today only by careful review, not the type system.

**Required outcome:** None required now. If this loop grows a new branch later, prefer deriving `identityResolution` as the return value of a function that must produce all three states (or restructure so every exit path assigns explicitly) rather than relying on the post-loop default.

**Found by:** `prp-core:seam-analyzer`, `prp-core:type-design-analyzer` (independently confirmed the same `undefined`-only-via-short-circuit invariant holds today by construction, not by the type system)

**Disposition:** OPEN, non-blocking. A note for whoever next touches this loop, not a defect in this PR.

</details>

<details>
<summary><code>R4</code> — three-state union wider than its one consumer</summary>

**Impact:** None on correctness — confirmed by the type-design review, which is the scope that owns this judgment: no reachable invalid state, no mishandled value, no forced downstream guess. Purely a surface-area observation: `identityResolution: 'resolved' | 'unregistered' | 'faulted' | undefined` has exactly one production consumer (`identityResolution === 'faulted'` in the persistence block); `'resolved'` and `'unregistered'` are never distinguished by any code in this diff. A `boolean` (or `'faulted' | undefined`) would express identical program behavior, removing the ternary at line 406, the post-loop default-fill, and shrinking the 19-line doc comment.

**Evidence:** `packages/workflows/src/executor.ts:328` (type), `:406` (ternary), `:439-441` (default-fill), `:2077` (sole consumer, `if (identityResolution === 'faulted')`).

**Required outcome:** None required — this is the PR's own explicitly-flagged open question ("does collapsing 'resolved'/'unregistered'/'faulted' into any two lose information that matters?"). The author's answer (keep three states because a future consumer — #2200's gate — is expected to want the distinction, and `'unregistered'` already corresponds to an existing log-observable branch) is a legitimate, non-costly design choice, not a proved premature structural decision.

**Found by:** `prp-core:code-simplifier`, `prp-core:seam-analyzer`, `prp-core:type-design-analyzer` (all three independently raised the same observation; type-design-analyzer, the authoritative scope, explicitly declined to elevate it to a finding)

**Disposition:** DECLINED — the extra state costs a small doc comment and one ternary, creates no reachable invalid state, and answers the author's own review-guidance question about a real forward-looking use. Not required by this PR's outcome or invariant.

</details>

<details>
<summary><code>R5</code> — minor test redundancy</summary>

**Impact:** None on coverage — both tests pass and are individually correct, they just exercise the same branch through near-identical setup.

**Evidence:** `packages/workflows/src/executor.test.ts:1932` ("does not touch metadata when the row was never flagged as faulted") uses `makeRun({ metadata: {} })`, byte-identical in effect to the default `makeRun()` implicitly used by `executor.test.ts:1834` ("persists output_root when identityResolution is resolved") — both assert no metadata patch rides the `output_root` write.

**Required outcome:** None required. Optional: fold one into the other on a future touch of this test file.

**Disposition:** OPEN, non-blocking. Not worth a standalone fix.

</details>

<details>
<summary><code>R6</code> — inert in-memory metadata mutation on the faulted branch</summary>

**Impact:** None — harmless dead write. `workflowRun.metadata` is mutated in-memory before the DB write is attempted, mirroring the pattern used for `WORKFLOW_SOURCE_METADATA_KEY` elsewhere in the same function (where the in-memory copy IS read later by child/fan-out dispatch in the same process). Here, `readIdentityUnresolved` is called exactly once in this file, in the *other* branch (the heal check), reading a freshly DB-loaded `workflowRun.metadata` from a prior run — never this branch's own in-memory mutation. The copied ceremony has no live purpose in this diff.

**Evidence:** `packages/workflows/src/executor.ts:2078-2082` (the mutation); `:2104` (the only reader, in the sibling `else` branch, operating on a different call's `workflowRun`).

**Required outcome:** None required — cosmetic only, safe to leave or drop.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | R1 |
| `seams` (`prp-core:seam-analyzer`) | R2, R3 |
| `simplify` (`prp-core:code-simplifier`) | R4 (co-found), R5, R6 |
| `types` (`prp-core:type-design-analyzer`) | R3 (co-confirmed), R4 (declined to elevate) — no independent findings |

#### Validation

|---|---|---|
| `bun run type-check` (full monorepo, all packages + scripts) | PASS | All 13 workspace packages + `scripts/tsconfig.json` exited 0 |
| `bun run lint --max-warnings 0` | PASS | `eslint . --cache --max-warnings 0` — clean |
| `bun run format:check` | PASS | "All matched files use Prettier code style!" |
| `bun --filter @archon/workflows test` (24 split invocations) | PASS | 0 failures across all splits, including new `identityResolution (#2304)` (7 cases), `output_root persistence branching (#2304)` (4 cases), `readIdentityUnresolved (#2304)` (4 cases) |
| `check:bundled` / `check:bundled-schema` / `check:pi-vendor-map` / `check:capability-matrix` | NOT RUN | Not applicable — this PR touches no bundled defaults, migrations, Pi SDK, or provider capabilities |
| `check:schema-upgrades` | NOT RUN | Requires live PostgreSQL; this PR makes no schema changes (metadata JSONB column only) so it is not implicated |

</details>

PR #2721, opened 2026-08-22
feat(workflows): gate nodes produce structured decision output (#2707)

### LINKED ISSUE #2714: bug(workflows): rejecting a first-node approval gate with on_reject stages a rework that resume can never reach
## Problem

`hydrateResumableRun` (`packages/workflows/src/executor.ts:920-941`) decides whether a paused run is worth resuming:

```ts
const approvalType =
  candidate.metadata?.approval !== undefined
    ? (candidate.metadata.approval as Record<string, unknown>).type
    : undefined;
const hasReRunGateState =
  approvalType === 'interactive_loop' || approvalType === 'child_workflow';
if (priorCompletedNodes.size === 0 && !hasReRunGateState) {
  getLog().info({ resumableRunId: candidate.id }, 'workflow.dag_resume_skipped_no_completed_nodes');
  return null;
}
```

`hasReRunGateState` only recognizes `interactive_loop` and `child_workflow`. It omits `approval`.

An `approval:` node pauses without writing `node_completed` (that event is written only on approve, in `approveWorkflow`). If a workflow's **first** node is `approval:` with `on_reject:` configured, and the run is rejected, `rejectWorkflow`'s stage-rework branch (`packages/core/src/operations/workflow-operations.ts`, the `willStageRework` path) keeps the run `paused`, stamps `rejection_reason`/`rejection_count`, and writes no `node_completed` — by design, so the node re-runs with the rejection feedback on resume (see `executeApprovalNode`'s rejection-resume detection, `dag-executor.ts` ~line 6430).

But because this node was the first node, `priorCompletedNodes.size === 0`. Since `approvalType === 'approval'` is not in `hasReRunGateState`, `hydrateResumableRun` returns `null` — "nothing worth resuming" — and `archon workflow resume <id>` (and chat/web resume) refuses with "Cannot resume: the prior run ... has no completed nodes and no interactive-loop state."

The staged rework the user was promised (reject with feedback → the `on_reject` prompt runs → gate re-pauses) is unreachable whenever the rejected `approval:` gate is the workflow's first node.

## Repro sketch

1. A workflow whose first node is `approval:` with `on_reject: { prompt: ... }`.
2. Run it; it pauses immediately (no completed nodes).
3. `/workflow reject <id> "needs changes"` — rework is staged, run stays `paused` with `rejection_reason` set.
4. `archon workflow resume <id>` (or the equivalent chat/web path) — refuses to resume because `priorCompletedNodes.size === 0` and `approvalType === 'approval'` is not treated as re-runnable gate state.

## Fix direction

Add `'approval'` to `hasReRunGateState` (or, better, replace the inline literal check with an exported `SuspendReason`-typed predicate in `packages/workflows/src/schemas/workflow-run.ts`, e.g. `reRunsOwnNodeOnResume(reason)`, switched exhaustively so a future fifth reason can't silently repeat this gap) — but only when a rework is actually staged (a plain first-node approval rejection with no `on_reject` cancels the run instead, which is already handled elsewhere).

## Context

Surfaced during independent review of #2711 (`refactor/2489-suspend-persist-path`), which unified the workflow engine's *pause* write path and added a `SuspendReason` type + doc comment naming each reason's resume site. `executor.ts` is untouched by that PR — this defect predates it and is out of scope there; filing separately per its own review guidance.

Wirasm:
## Problem and outcome

A gate's human response reaches downstream nodes through three separate private channels invented at three different times — `capture_response` output, `$REJECTION_REASON`, `$LOOP_USER_INPUT` — each with its own position-dependent rules, and `on_reject` is a stored rework template with a private attempt counter rather than ordinary workflow wiring. This makes "reject" mean two different things depending on whether `on_reject` happens to be configured, and it hides a resume bug (#2714): a first-node gate rejected with a staged `on_reject` rework can never be resumed, because the staged state lives only in `workflow_runs.metadata` and `hydrateResumableRun`'s recognized-gate-state check never covered plain `approval`.

- **Outcome:** A gate node (`approval:`) whose author explicitly declares `approval.decisions:` in YAML resolves with ordinary structured `{decision, text}` output the moment a human responds — readable via `$<gateId>.output.decision` / `.text` through the exact `when:`/substitution wiring every other node already uses. Authors declare their own decisions (`approval.decisions:`, ids constrained to `approve`/`reject` with optional custom `label:`) and wire a rejection's rework path as an explicit `when:` edge to their own node, instead of relying on the engine's private `on_reject` template. This is opt-in: a gate that doesn't author `approval.decisions:` — including every gate authored before this PR — is completely unaffected.
- **Invariant:** The new mechanism activates only when `GateNode.decisionsAuthored` is `true` — set exclusively when the author wrote `approval.decisions:` explicitly. Since that field did not exist before this PR, no pre-existing gate (bundled or custom) can satisfy it, so every already-authored workflow's approve/reject behavior is byte-for-byte unchanged. Resolving an opted-in gate always writes `node_completed` immediately, for both approve and reject — there is no state where the gate is "resolved" but no node has completed.
- **Scope boundary:** This is Sequencing step 1 of #2707 only. Workflow classes / the general `respond` drive verb (step 2) and the interactive-loop → `loop_group` collapse (step 3) are separate, later workstreams. The legacy `on_reject`/`capture_response`/node-level loop `interactive:` mechanism is kept running byte-for-byte (grow-then-deprecate, warned not errored) — nothing bundled breaks.
- **Root cause (#2714):** `executeApprovalNode`'s pause and `rejectWorkflow`'s stage-rework branch never write a `node_completed` event for the gate — only `workflow_runs.metadata` changes. `hydrateResumableRun`'s recognized-gate-state carve-out only listed `interactive_loop`/`child_workflow`, so a first-node gate with zero completed nodes and a staged rework was reported as "nothing worth resuming."

## Review guidance

- **Feedback requested:** whether gating the new mechanism on an explicit `approval.decisions:` opt-in (`decisionsAuthored`) is the right boundary, versus something looser — this is the design correction the first review round produced (see below), not the original plan.
- **Start here:** `packages/workflows/src/schemas/dag-node.ts` — `GateNode.decisionsAuthored`'s doc comment, `approvalConfigSchema`/`approvalDecisionConfigSchema`, and the `.transform()` branch that sets `decisionsAuthored: data.approval.decisions !== undefined`.
- **Review order:** `schemas/dag-node.ts` (`decisionsAuthored` + the approve-required `superRefine`) → `schemas/workflow-run.ts` (`ApprovalContext.decisionsAuthored`, `reRunsOwnNodeOnResume`) → `dag-executor.ts` (`decisionsAuthored` snapshotted into `pauseGateRespectingExternalTransition`'s call) → `operations/workflow-operations.ts` (`approveWorkflow`/`rejectWorkflow`'s `isNewMode`/`willResolveNewMode` checks, both gated on `decisionsAuthored`) → `loader.ts` (`collectGateAndLoopDeprecationWarnings`, the `capture_response` warning gated on the same flag) → `executor.ts` (`hydrateResumableRun`).
- **Lower-attention areas:** the test file updates in `command-handler.test.ts`, `api.workflow-runs.test.ts`, `workflow-operations.test.ts`, and `loader.test.ts` are mostly mechanical — each adds a `decisionsAuthored: true` fixture for the genuine new-mode case and a matching "bare gate, byte-for-byte unchanged" counterpart.
- **Known risk or uncertainty:** `approval.decisions:` ids are deliberately constrained to a non-empty subset of `{approve, reject}` (custom `label:` allowed, `approve` required) rather than fully free-form — no current drive surface (CLI/web/chat) can select a custom decision id yet, so a broader vocabulary would be unreachable until #2707 step 2 lands the general `respond` verb.

## Solution

Two structural facts drove keeping the legacy `on_reject` mechanism running unmodified rather than replacing it outright: (1) `rework`/`captureResponse` are already loader-synthesis-only fields (never author-writable), and the legacy path's rejection-resume/rework-execution machinery (`executeApprovalNode`, `rejectWorkflow`'s stage-rework branch) already fully implements the "grow-then-deprecate — nothing breaks by rename" posture #2707's Migration section requires; (2) auto-translating `on_reject` into new-model primitives (a synthesized `loop_group` wrapping the gate + a rework node) was evaluated and rejected — `loop_group` output is explicitly unstructured raw text, so any workflow reading a captured comment via `$<gateId>.output` today would silently start reading the wrong value after "modernization."

The two mechanisms coexist, but the mode boundary went through a review-driven correction: the first pass classified a gate as "new mode" whenever `on_reject` was absent — including the default, omitted-everything shape every existing gate has, since `decisions:` didn't exist to author. That meant rejecting *any* plain gate would silently resolve-and-continue instead of cancel, because no existing workflow has ever needed a `when:` clause on a gate decision (the capability didn't exist), so an unguarded downstream node would run exactly as if the gate had been approved — a governance-safety regression on every already-authored workflow, caught by independent review. The fix: `GateNode.decisionsAuthored` / `ApprovalContext.decisionsAuthored` — true only when the author wrote `approval.decisions:` explicitly in YAML — is now the entire mode signal, snapshotted from the schema into the pause-time `ApprovalContext` so `approveWorkflow`/`rejectWorkflow` read the same signal the gate was authored with:

- **Legacy or unauthored** (`on_reject:` configured, OR `decisionsAuthored` false — the bare default, or `capture_response`-only): untouched. Same rejection-resume detection for `on_reject`, same plain-text/`capture_response`-gated output, and reject always cancels the run exactly as before this PR.
- **New mode** (`onRejectPrompt` absent AND `decisionsAuthored === true`): `approveWorkflow`/`rejectWorkflow` write `node_completed` with `structured_output: {decision, text}` immediately on both approve and reject. Reject is no longer inherently terminal — a gate declaring a `reject` decision stays paused/resumable exactly like approve; a gate declaring no `reject` decision (an approve-only "acknowledge" gate) still cancels on reject. A schema-level check requires an authored `decisions:` array to include `'approve'`, so a reject-only declaration is a load error instead of a silent runtime footgun.

`#2714` is fixed for the legacy path with the exact minimal extension the issue itself proposed — a new exported, exhaustively-switched `reRunsOwnNodeOnResume(approval, metadata)` predicate in `schemas/workflow-run.ts` that recognizes a genuinely staged legacy rework (`resolved === 'rejected'` + non-empty `rejection_reason` + `onRejectPrompt` still present) as re-runnable with zero completed nodes. New-mode gates need no equivalent carve-out: they never reach a "resolved but zero completed nodes" state to begin with, since both outcomes write `node_completed` through the ordinary completion path.

Three retiring surfaces — `approval.capture_response` (warned only when also combined with `decisionsAuthored`, since it's otherwise still fully functional), `approval.on_reject`, and node-level loop `interactive:`/`loop_group.interactive:` — get a `parseWarnings` entry (the existing #2213 mechanism, following the `modelReasoningEffort` deprecation precedent) naming the replacement, recursing into `loop_group` bodies. `approval.decisions[]` entries also get typo protection via a new `'array'` `NestedKeySpec` kind, matching every other nested field. `$REJECTION_REASON`/`$LOOP_USER_INPUT` are covered as a consequence of warning on their enabling keys rather than by a separate dead-reference scan — both variables are populated only inside their sanctioned context, so any other usage was already a silent no-op before this PR.

Every drive surface that renders a post-reject message (CLI, web route, chat command handler, `manage_run`, Slack) now reads a new `RejectionOperationResult.newMode` field instead of assuming "not cancelled" always means "a legacy rework prompt is running" — a new-mode resolution says the run is continuing instead.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A plain gate's output is always `''` (or the raw comment when `capture_response: true`); reject with no `on_reject` cancels the run; reject with `on_reject` stages an invisible rework. | Unchanged for every gate that doesn't author `approval.decisions:` — including bare gates and `capture_response`-only gates. A gate that DOES author `approval.decisions:` gets `{decision, text}` output, resolvable via `$<gateId>.output.decision`/`.text`; reject with a declared `reject` decision writes `node_completed` and stays resumable; reject with no `reject` decision still cancels. Legacy `on_reject` gates are unaffected either way. |
| **Failure behavior** | A first-node legacy `on_reject` gate, once rejected, could never be resumed (#2714). | Resumes correctly; a new-mode gate can't reach that failure class at all. |

## Validation

- `bun run validate` — passed — full chain (bundled/schema/pi-vendor-map/capability-matrix checks, type-check, lint, format:check, full test suite) green with zero failures. Run twice: once before the review-driven `decisionsAuthored` correction, once after.
- `bun run type-check` (all packages), `bun run lint`, `bun run format:check` — passed.
- `bun --filter @archon/workflows test` / `@archon/core` / `@archon/server` / `@archon/cli` / `@archon/adapters` / `@archon/web` (full package suites) — passed, 0 fail across every chained segment.
- Regression proof for #2714: manually reverted the `executor.ts`/`schemas/workflow-run.ts`/`schemas/index.ts` fix via `git stash`, re-ran `executor.test.ts` — the new regression test failed exactly as expected, then restored the fix and confirmed green.
- Regression proof for the `decisionsAuthored` correction: a new test asserts that a gate with the synthesized default `decisions` array but no `decisionsAuthored: true` still cancels on reject — exactly the shape that silently continued before the fix.
- **Not verified:** `bun run check:schema-upgrades` — not run; this PR touches no DB migration file (`git status` confirms nothing under `migrations/`), so the CI-only Postgres check is inapplicable.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Every existing workflow using `on_reject`/`capture_response`/node-level loop `interactive:` (including the bundled `archon-interactive-prd` and `archon-piv-loop`) — and every bare gate with no deprecated fields at all — keeps running unmodified; the deprecated ones now surface a `parseWarnings` entry naming the replacement. | `loader.test.ts`'s bundled-corpus test (`KNOWN_BAD` allowlist) and the `#2707 step 1 gate/loop deprecation warnings` describe block. |
| Observability | Deprecation warnings also log via `getLog().warn(...)` with a `_deprecated` event-name suffix, matching the existing `modelReasoningEffort` convention. | `packages/workflows/src/loader.ts`, `collectGateAndLoopDeprecationWarnings`. |
| Documentation / communication | AGENTS.md's `approval:`/loop variable docs are not yet updated with the deprecation notes — left as a small follow-up, not blocking. | — |

## Links

- Fixes #2714
- Related #2707
- Plan: https://github.com/coleam00/Archon/issues/2707#issuecomment-5380827640

## Summary by CodeRabbit

* **New Features**
  * Approval gates support labeled approve/reject decisions.
  * New-style approval results include structured decision and response text.
  * Decision details are preserved when workflows pause and resume.
  * Rejection messages now distinguish workflow continuation from legacy rework.

* **Bug Fixes**
  * Improved workflow resumption for paused approval gates, including rejection scenarios.

* **Deprecations**
  * Warnings now identify legacy approval and interactive-loop options while preserving existing behavior.


Wirasm:
## Ready to merge

The correction commit (`ab3f1c67`) fixes all 5 findings from the first review with one root change: `GateNode.decisionsAuthored` / `ApprovalContext.decisionsAuthored` — true only when the author explicitly writes `approval.decisions:` in YAML — now gates the entire new-mode mechanism, replacing the original "no `on_reject` configured" signal that made rejecting *any* plain gate silently continue the run. Since `decisions:` did not exist before this PR, no pre-existing gate (bundled or custom) can satisfy the new gate, so every already-authored workflow keeps its exact pre-PR behavior; only a gate an author positively rewrites gets the new structured-output resolution. Each of the four other findings (stale drive-surface messages, reject-only-gate approval, the false `capture_response` warning, missing typo protection) is fixed by direct, independently-verified code changes — not merely claimed. All five are backed by new or corrected tests that fail against the pre-correction behavior and pass against the fix.

**Validation:** `bun run type-check` (all packages) — PASS, 0 errors. `bun run lint` (`eslint . --cache`, `--max-warnings 0` enforced by config) — PASS, 0 output. `bun run format:check` — PASS. `bun run check:bundled` — PASS, generated bundle up to date (37 commands, 21 workflows, 0 scripts). `bun --filter @archon/workflows test` — PASS, 0 fail across all ~26 segments. `bun --filter @archon/core test` — PASS, 0 fail across all ~35 segments. `bun --filter @archon/server test` — PASS, 0 fail. `bun --filter @archon/cli test` — PASS, 0 fail across 10 segments. `bun --filter @archon/adapters test` — PASS, 0 fail across 6 segments. All commands independently re-run by the coordinator in a fresh detached worktree pinned to `ab3f1c67` (after `bun install`, since the worktree had no `node_modules`), not merely re-quoted from the implementation report.

**Resolved:** 5 (`R1`–`R5`, all `OPEN` → `FIXED`)

### Findings

|---|---|---|---|
| `R1` | Important | Rejecting any plain (unconfigured) gate no longer stopped the run — unguarded downstream nodes would execute, and every drive surface's reject message was false | FIXED |
| `R2` | Important | Bundled `archon-interactive-prd.yaml`'s `capture_response` gates would embed raw JSON in downstream AI prompts | FIXED |
| `R3` | Important | `approveWorkflow` had no check mirroring `rejectWorkflow`'s `hasRejectDecision` — a reject-only gate could still be "approved" | FIXED |
| `R4` | Important | The `capture_response` deprecation warning was unconditionally false for the legal `capture_response` + `on_reject` combination | FIXED |
| `R5` | Suggestion | `approval.decisions[]` wasn't covered by the loader's unknown-key typo detector | FIXED |

<details>
<summary><code>R1</code> — Reject no longer stopped the run by default; drive-surface messages went stale</summary>

**Impact (as originally found):** Rejecting a plain `approval:` gate (no `on_reject:`, the default authoring shape used by every bundled workflow with a gate) would resolve with `decision: 'reject'` and auto-resume the run instead of cancelling it — an unguarded downstream node would run exactly as if the gate had been approved. Independently, all five drive surfaces that render a reject message assumed `cancelled === false` meant "a legacy `on_reject` rework is running" and printed stale rework language even when nothing reworked.

**Fix verified:** `packages/workflows/src/schemas/dag-node.ts:539` adds `decisionsAuthored: z.boolean()` to `gateNodeSchema`, set at `:1535` to `data.approval.decisions !== undefined` — true only when the YAML literally authors `approval.decisions:`. `packages/core/src/operations/workflow-operations.ts:497-559` (`approveWorkflow`) and `:658-666,742` (`rejectWorkflow`) now key `isNewMode`/`willResolveNewMode` on `onRejectPrompt == null && decisionsAuthored === true` (reject additionally requires `hasRejectDecision`), not merely `onRejectPrompt == null`. `packages/workflows/src/dag-executor.ts:6651` snapshots `decisionsAuthored: node.decisionsAuthored` into the paused `ApprovalContext` (`packages/workflows/src/schemas/workflow-run.ts:448`) at pause time, mirroring how `onRejectPrompt` is already snapshotted — so approve/reject handlers read the same signal the gate was authored with, and a run paused by a pre-#2707 build (where the field is absent) falls back to legacy behavior by construction (optional field, falsy default).

A new regression test proves the exact defect class the first review found is now closed: `packages/core/src/operations/workflow-operations.test.ts` — `'a gate with the synthesized default decisions but decisionsAuthored NOT set still cancels on reject (R1 fix — #2707)'` — asserts `result.cancelled === true` and `mockResolveApprovalGate` is NOT called for a paused context with `decisions` populated (the engine's own synthesized default pair) but `decisionsAuthored` omitted — exactly the shape that silently continued before the fix.

The five drive-surface messages are all corrected, each gated on the new `RejectionOperationResult.newMode` field (`workflow-operations.ts:80`, populated at both `rejectWorkflow` return sites): `command-handler.ts:1029-1033` ("rejected. Continuing..." vs. "Reworking with your feedback..."), `api.ts:3852-3859` (two-way branch on `newMode` × `autoResumed`), `cli/commands/workflow.ts:3351-3355` ("Resuming..." vs. "Resuming with on_reject prompt..."), `manage-run-tool.ts:395-397` ("The run continues." vs. "It reworks with your feedback."), `slack/workflow-bridge.ts:487-489` ("recorded — the run continues" vs. "recorded — workflow will retry with feedback"). Confirmed by direct diff read of all five files, not by trusting the report's file list.

No bundled workflow authors `approval.decisions:` (verified: `grep -rl "decisions:" .archon/workflows/defaults/` returns only `archon-piv-loop.yaml`, which is the prose phrase "Key decisions:" inside a prompt template, not the YAML field), so every bundled gate keeps its exact pre-PR reject-cancels behavior.

**Disposition:** FIXED. Verified by direct code read (schema, both operation functions, executor snapshot site, all five message call sites) and by independently re-running `bun --filter @archon/workflows test` and `bun --filter @archon/core test` in a fresh worktree — 0 fail.

</details>

<details>
<summary><code>R2</code> — Bundled <code>archon-interactive-prd</code> would have embedded raw JSON in prompts</summary>

**Impact (as originally found):** `archon-interactive-prd.yaml`'s three `capture_response: true` gates (no `on_reject:`) would have been misclassified as new-mode under the original "no `on_reject`" signal, so reviewers' plain-text PRD answers would arrive at downstream prompts as a `{"decision":"approve","text":"..."}` blob instead of plain text.

**Fix verified:** Same root fix as R1 — `isNewMode` now requires `decisionsAuthored === true`, and `archon-interactive-prd.yaml` (confirmed via `git diff` between the two heads: zero changes to that file) has no `decisions:` authored, so `decisionsAuthored` is `false` for all three of its gates regardless of `capture_response`. They stay on the legacy `captureResponse`-gated plain-text output path exactly as before. Proven by new/renamed tests asserting a bare `capture_response`-only gate emits plain text, not JSON: `workflow-operations.test.ts` — `'bare gate with captureResponse but no decisionsAuthored keeps plain-text output (R2 fix — #2707)'` — and mirrored equivalents in `command-handler.test.ts` and `api.workflow-runs.test.ts` (same rename, same assertion pattern, confirmed by reading both diffs).

**Disposition:** FIXED. Verified by inspecting the unchanged bundled YAML plus the corrected classification logic and its three test-file mirrors; re-ran the three affected package test suites (`@archon/core`, `@archon/server`) — 0 fail.

</details>

<details>
<summary><code>R3</code> — Reject-only gate could still be "approved"</summary>

**Impact (as originally found):** `approvalDecisionConfigSchema` only constrained each entry's `id` to `enum(['approve','reject'])` with no requirement that `'approve'` be present, so `approval: { decisions: [{ id: 'reject' }] }` parsed successfully, and `approveWorkflow` had no equivalent guard to `rejectWorkflow`'s `hasRejectDecision` fallback — `workflow approve <id>` would "succeed" for an outcome the gate's own declared vocabulary said was unreachable.

**Fix verified:** `packages/workflows/src/schemas/dag-node.ts:986-1001` — inside the existing `superRefine`'s `else if (data.approval.decisions !== undefined)` branch (fires only when `decisions:` is explicitly authored, so the default-synthesized pair is unaffected), a new check adds a `ZodIssueCode.custom` issue at `path: ['approval', 'decisions']` when `!ids.includes('approve')`. This is a load-time schema error, not a runtime guard, matching the review's requested "smallest fix." Directly proven by a new test in `packages/workflows/src/schemas.test.ts` — `'rejects decisions with no approve entry (R3 fix)'` — asserting `dagNodeSchema.safeParse({..., decisions: [{ id: 'reject' }] }).success === false` with an error message containing `'approve'`; a companion test confirms an approve-only array (no `reject`) still parses successfully, preserving the deliberate approve-only-gate asymmetry.

**Disposition:** FIXED. Verified by reading the schema change directly and confirming the new test reproduces (and now blocks) the exact repro the first review used (`dagNodeSchema.safeParse` on a reject-only `decisions:` array). Re-ran `bun --filter @archon/workflows test` — 0 fail, including this new test.

</details>

<details>
<summary><code>R4</code> — <code>capture_response</code> deprecation warning was false for the legal <code>capture_response</code> + <code>on_reject</code> combination</summary>

**Impact (as originally found):** The loader warned "`approval.capture_response` is deprecated ... This field is ignored" whenever `capture_response` was present, regardless of mode — but for a legacy gate combining `capture_response: true` with `on_reject:` (a legal, pre-existing combination), `capture_response` genuinely still controlled the node's output. The pre-correction test suite had locked in the false-positive warning as expected behavior.

**Fix verified:** `packages/workflows/src/loader.ts:373` — the warning condition changed from `approvalObj.capture_response !== undefined` to `approvalObj.capture_response !== undefined && node.decisionsAuthored`, inside a branch where `isGateNode(node)` has already narrowed `node` to `GateNode` (which carries `decisionsAuthored`). Now the warning fires only when `capture_response` is genuinely ignored (new mode). The previously-locked-in-wrong test at `loader.test.ts` (the "warns on `interactive`/`capture_response`/`on_reject`" describe block) is corrected: the fixture combining `capture_response` + `on_reject` no longer asserts the capture_response warning fires, and the test comment now explicitly states why ("still fully functional (R4 fix), so it does NOT warn here"). Two new targeted tests were added: `'warns on approval.capture_response combined with explicitly authored decisions (R4 fix — only truly ignored there)'` (positive case, `decisions:` + `capture_response:` both authored → warns) and `'does NOT warn on approval.capture_response without decisions: authored — still fully functional (R4 fix)'` (bare `capture_response`-only gate → no warning).

**Disposition:** FIXED. Verified by reading the corrected loader condition, the `isGateNode` narrowing that makes `node.decisionsAuthored` type-safe to read, and all three test changes (one corrected, two new) directly in the diff. Re-ran `bun --filter @archon/workflows test` — 0 fail.

</details>

<details>
<summary><code>R5</code> — <code>decisions:</code> entries had no typo protection</summary>

**Impact (as originally found):** `NestedKeySpec` only modeled `'object'`/`'record'` kinds, so the new array-of-objects `decisions:` field couldn't be registered for unknown-key detection, and `approvalDecisionConfigSchema` being a plain (non-`.strict()`) object meant a typo'd `label` silently vanished with zero diagnostic.

**Fix verified:** `packages/workflows/src/schemas/dag-node.ts:1706-1707` adds an `'array'` variant to the `NestedKeySpec` union (`{ kind: 'array'; entry: NestedKeySpec }`), and `:1770-1780` registers `approval.decisions` with `{ kind: 'array', entry: { kind: 'object', keys: Object.keys(approvalDecisionConfigSchema.shape) } }` inside `KNOWN_NODE_NESTED_KEYS`'s `approval` children. Critically, this is not just a registration with no consumer: `packages/workflows/src/loader.ts:259-273` (`collectUnknownConfigKeys`) has a matching `if (spec.kind === 'array')` branch — confirmed present, not merely stubbed — that iterates array entries by index and recurses with a `${keyPath}${i}.` prefix, so it reaches an object-kind child spec and applies the existing unknown-key logic per entry. Directly proven by a new test in `loader.test.ts` — `'warns on a typo inside a decisions: entry (R5 fix — array typo protection)'` — which authors a `decisions:` array with `lable: Needs work` (typo for `label`) and asserts the warning list contains `"unknown key 'approval.decisions.1.lable'"`.

**Disposition:** FIXED. Verified by reading both the type addition, the registration, and the consuming branch in `loader.ts` (confirming the array kind is actually wired into the recursive walker, not just declared), plus the new test. Re-ran `bun --filter @archon/workflows test` — 0 fail.

</details>

<details>

#### Reviewer coverage

Correction verification performed by the coordinator directly (no scope reviewer agents dispatched) — the correction diff is small (16 files, 494/-104), fully attributable to the previous review's own five findings, and independently verifiable via direct code and test inspection plus a full re-run of the affected packages' test suites. Every one of the five prior findings was traced to its exact fix site and its exact proving test, and the correction diff was read in full (`dag-node.ts`, `workflow-operations.ts`, `loader.ts`, all five drive-surface message files, `workflow-run.ts`, `dag-executor.ts`, and all five touched test files) rather than sampled.

| Scope | Result |
|---|---|
| `code` | R1, R2 verified fixed at all five drive-surface call sites + the root classification signal |
| `seams` | R4, R5 verified fixed — `decisionsAuthored` is snapshotted consistently between `GateNode` (load-time) and `ApprovalContext` (pause-time), and the new `'array'` `NestedKeySpec` kind has a real consuming branch in `loader.ts`, not a dead registration |
| `simplify` | R1, R3, R5 verified fixed with no new complexity introduced beyond what each fix required; no over-correction found (e.g., `RejectionOperationResult.newMode` is used at exactly the five call sites that needed it) |
| `types` | R2, R3 verified fixed — `approvalDecisionConfigSchema`'s new approve-required check lives in the existing `superRefine`, scoped correctly to fire only when `decisions:` is authored (the default-synthesized pair is unaffected) |

#### Validation

|---|---|---|
| `bun run type-check` (all packages, fresh `bun install` in the review worktree) | PASS | Every package exited 0; zero TS errors |
| `bun run lint` (`eslint . --cache`, zero-warning policy) | PASS | Empty output, exit 0 |
| `bun run format:check` | PASS | "All matched files use Prettier code style!" |
| `bun run check:bundled` | PASS | "bundled-defaults.generated.ts is up to date (37 commands, 21 workflows, 0 scripts)" — confirms `archon-interactive-prd.yaml` unaffected |
| `bun --filter @archon/workflows test` | PASS | 0 fail across all segments (largest: 595 pass in one file) |
| `bun --filter @archon/core test` | PASS | 0 fail across ~35 segments |
| `bun --filter @archon/server test` | PASS | 0 fail |
| `bun --filter @archon/cli test` | PASS | 0 fail across 10 segments |
| `bun --filter @archon/adapters test` | PASS | 0 fail across 6 segments |
| `bun run check:schema-upgrades` | NOT RUN | No file under `migrations/` changed in either the original PR or the correction diff — CI-only Postgres check inapplicable, matching the PR's own claim |

</details>

Issue #975, opened 2026-04-08
feat(web): forward SDK lifecycle events to Web UI (tasks + hooks)


Wirasm:
> Migrated from dynamous-community/remote-coding-agent#933 — Archon active development has moved to coleam00/Archon. Original issue retained as historical reference.

---

## Problem

The Claude SDK emits lifecycle events for subagent tasks and hooks that we ignore. Users see nothing when a DAG node spawns subagents or when hooks fire — they wait blindly until the node completes.

### Task lifecycle (subagent visibility)

| SDK event | Subtype | Key fields | When it fires |
|---|---|---|---|
| `SDKTaskStartedMessage` | `task_started` | `task_id`, `description`, `task_type?`, `prompt?` | Subagent spawned |
| `SDKTaskProgressMessage` | `task_progress` | `task_id`, `description`, `usage`, `last_tool_name?`, `summary?` | Subagent working (~30s intervals with `agentProgressSummaries`) |
| `SDKTaskNotificationMessage` | `task_notification` | `task_id`, `status`, `summary`, `usage?` | Subagent completed/failed/stopped |

Currently these hit the `else` branch in `claude.ts` and get logged as `claude.system_message_unhandled`.

### Hook lifecycle (hook observability)

| SDK event | Subtype | Key fields | When it fires |
|---|---|---|---|
| `SDKHookStartedMessage` | `hook_started` | `hook_id`, `hook_name`, `hook_event` | Hook callback invoked |
| `SDKHookProgressMessage` | `hook_progress` | `hook_id`, `hook_name`, `stdout`, `stderr` | Hook running |
| `SDKHookResponseMessage` | `hook_response` | `hook_id`, `hook_name`, `outcome`, `exit_code?` | Hook finished |

Available since SDK v0.2.89 (just bumped). Per-node hooks (#445) already work but events are invisible.

### Phase 1: Yield from `claude.ts`

Add handling for `system` subtype messages:

```typescript
case 'task_started':
  yield { type: 'task_started', taskId: msg.task_id, description: msg.description };
  break;
case 'task_progress':
  yield { type: 'task_progress', taskId: msg.task_id, summary: msg.summary, usage: msg.usage };
  break;
case 'task_notification':
  yield { type: 'task_notification', taskId: msg.task_id, status: msg.status, summary: msg.summary };
  break;
case 'hook_started':
  yield { type: 'hook_started', hookId: msg.hook_id, hookName: msg.hook_name, hookEvent: msg.hook_event };
[...]

PR #2723, opened 2026-08-22
fix(providers): surface Claude tool-scoped hook activity in the audit stream

### LINKED ISSUE #2324: hook_activity only ever records SessionStart: 3,670 tool calls, zero PreToolUse/PostToolUse rows
## Summary

- **What broke:** `hook_activity` events only ever carry `hook_event: SessionStart`. No `PreToolUse` or `PostToolUse` activity is recorded, even though those hooks demonstrably fire inside Archon nodes. The result is that hook execution is not auditable, and — worse — the empty table reads as evidence that hooks are not running.
- **When it started (if known):** Unknown; all 678 rows in our DB have this shape, going back to the earliest runs we have.
- **Severity: `minor`** for correctness, but please read the Impact section — the practical risk is that operators conclude their safety hooks are inert and stop relying on them, or (our case) nearly conclude the opposite of the truth.

If this is working as designed, then the ask is a documentation change rather than a code change. Either resolution works for us; the current state is the problem because it is silently misleading.

## Steps to Reproduce

1. Configure Claude Code hooks at the user level (`~/.claude/hooks/…`, e.g. a `PreToolUse` matcher that blocks dangerous Bash commands).
2. Run any workflow with AI nodes that make tool calls.
3. Query the events table:
   ```sql
   SELECT json_extract(data,'$.hook_name'), COUNT(*)
   FROM remote_agent_workflow_events
   WHERE event_type = 'hook_activity'
   GROUP BY 1;
   ```
4. Only `SessionStart:*` rows come back, regardless of how many tool calls the run made.

## Expected vs Actual

- **Expected**: If hook activity is recorded at all, tool-scoped hooks (`PreToolUse` / `PostToolUse`) appear alongside session-lifecycle hooks — or the docs state plainly that `hook_activity` covers session lifecycle only.
- **Actual**: Only session lifecycle is recorded, with no indication anywhere that the other hook events are out of scope. The absence of a row is indistinguishable from "the hook did not fire".

## User Flow

```
  AI node                Claude Code hooks           Archon events table
  ───────                ─────────────────           ───────────────────
  session start ───────▶ SessionStart fires ───────▶ hook_activity row  ✔
  tool call     ───────▶ PreToolUse fires
                         (blocks a dangerous
                          command — verified)
                                                     [X] no row written
  tool completes ──────▶ PostToolUse fires           [X] no row written
                                                          │
  operator later queries hook_activity ◀──────────────────┘
  [X] sees only SessionStart, and reasonably concludes
      "my PreToolUse guard is not running inside Archon"
```

## Environment

- Platform: CLI
- Database: SQLite
- Running in worktree? `Yes` (and also observed with `--no-worktree`)
- OS: WSL2 (Ubuntu on Linux 6.x), Windows host
- Archon: v0.5.0, source build pinned at `59bbd00`
- Hooks: user-level Claude Code hooks, including a `PreToolUse` Bash guard

## Logs

Measured on this machine, 2026-07-30:

```
total workflow runs                : 59
total events                       : 9161   (16 distinct event types)

tool_called    events              : 3670
tool_completed events              : 3663

hook_activity  events              : 678    across 45 runs
  hook_name = SessionStart:startup : 632
  hook_name = SessionStart:resume  :  46
  any PreToolUse                   :   0
  any PostToolUse                  :   0
```

The `data` payload has the fields you would need (`hook_id`, `hook_name`, `hook_event`, `activity`, `outcome`, `exit_code`), so the schema does not look like the limiting factor.

The same shape held three days earlier at a smaller sample (630 rows, 630 `SessionStart`), so this is not a recent regression or a one-off.

## Impact

- **Affected workflows/commands:** all workflows with AI nodes. 3,670 recorded tool calls produced zero tool-scoped hook rows.
- **Reproduction rate:** Always.
- **Workaround available?** Only indirect: observe the hook's *side effect* rather than its record. We confirmed `PreToolUse` hooks really do fire inside Archon nodes by watching a dangerous command actually get blocked during an evaluation run — `bypassPermissions` does not disable `PreToolUse`. But that is a manual, one-off confirmation, not something you can check per-run or after the fact.
- **Data loss risk?** `No` for data; **it is an audit gap**. For anyone treating hooks as their deterministic safety layer inside Archon (which is how we use them — the human gate is deliberately narrow, and hooks are what makes the autonomous parts safe), there is no way to demonstrate that the layer was active for a given run.

The specific misreading we nearly made: querying this table and finding no `PreToolUse` rows, and concluding the guard was inert inside Archon. It is not — but nothing in the data says so. We ended up writing a warning into our own operations manual, which is a sign the tool should say it instead.

## Scope

- Package(s) likely involved: `core` (event writer) and/or `providers` (Claude hook plumbing) — we have not traced which side drops it
- Module (if known): whatever emits `hook_activity`

## What would resolve this, in preference order

1. Record `PreToolUse` / `PostToolUse` activity, at least in aggregate (a per-node count of fired/blocked would already be enough to prove the layer ran).
2. If that is too chatty at 3,670 tool calls per 59 runs: record only non-`success` outcomes (blocks, non-zero exits) — the interesting case for a safety guard is the block, not the pass.
3. If tool-scoped hooks are intentionally out of scope: say so in the docs, so nobody reads the empty result as "hooks are not firing".

Wirasm:
## Problem and outcome

`hook_activity` only ever carried `SessionStart` rows. Tool-scoped hooks (`PreToolUse`, `PostToolUse`, `Stop`, …) fired inside nodes but left no trace, so the audit stream could not distinguish "hook did not run" from "hook ran and was denied" — and an operator's `PreToolUse` guard that actively blocks Bash was invisible. Issue #2324 documents the symptom and shows a deterministic reproduction (a node-level `PreToolUse(Bash)` hook that denies the tool, while every `hook_activity` row is still `SessionStart`).

- **Outcome:** Tool-scoped Claude hooks now reach the existing workflow `hook_activity` audit stream as `hook_started` / `hook_response` frames, with `hook_event` and the response's outcome intact, matching the contract #1939 established against #975.
- **Invariant:** `SessionStart` and `Setup` behavior is unchanged; the existing `hook_started` / `hook_response` normalization in `provider.ts` and the DAG persistence path in `dag-executor.ts` already model the lifecycle frames Archon surfaces, so their shape is unchanged.
- **Scope boundary:** No event-schema change, no UI change, no SDK upgrade, no new audit surface. The locked SDK already supports the flag.
- **Root cause:** `buildBaseClaudeOptions` did not set `includeHookEvents`, so the SDK defaulted to `false` and suppressed lifecycle frames for every hook type except `SessionStart` and `Setup`.

## Review guidance

- **Feedback requested:** Correctness of the opt-in placement and whether the regression coverage pins both the SDK option and the chunk-pipeline shape enough to prevent a future regression.
- **Start here:** `packages/providers/src/claude/provider.ts:854` — the single new SDK option that closes the audit gap.
- **Review order:** `provider.ts:786-854` (the base options builder and the new flag), then `provider.ts:1115-1134` (the existing normalization the tests exercise), then `provider.test.ts:929-1064` (the four regression tests).
- **Lower-attention areas:** The four new tests are mechanical assertions on the mocked SDK options and the resulting chunk shape; no new public API.
- **Known risk or uncertainty:** None. The other failing tests under `packages/providers` (sub-run e2e, Pi `withCustomProviderRequestEnv`) reproduce on the unpatched `69adee5d` commit and are unrelated.

## Solution

`buildBaseClaudeOptions` now returns `includeHookEvents: true`, the SDK opt-in the locked `@anthropic-ai/claude-agent-sdk` already documents. With the flag on, the SDK can emit `hook_started` / `hook_progress` / `hook_response`; the existing `provider.ts` system-message handler converts `hook_started` and `hook_response` into Archon chunks at the same boundary as before, while `hook_progress` falls through because it is only emitted for async hooks, which Archon does not register. The DAG executor's `hook_activity` persistence path (`dag-executor.ts:2608-2666`) already models every field needed for those surfaced frames — `hook_event`, `hook_name`, outcome, exit code — so the audit stream starts receiving the tool-scoped rows with no schema or persistence work.

This is the smallest coherent solution because every other layer (normalization of the surfaced lifecycle frames, persistence, UI display names) already targets this contract; the only missing handoff was the SDK option itself.

### Changed seams

|---|---|---|
| Claude provider → SDK | New `includeHookEvents: true` opt-in on every `query({ prompt, options })` call | `packages/providers/src/claude/provider.ts:854` |
| SDK → provider chunks | Existing `hook_started` / `hook_response` normalization now reaches tool-scoped frames | `packages/providers/src/claude/provider.test.ts:949-1021` |
| Workflow `hook_activity` audit | Tool-scoped frames flow into the DAG executor's existing persistence boundary with no edits | `packages/providers/src/claude/provider.test.ts:975-1021` covers both deny and success paths |

## Validation

- `bun test src/claude/provider.test.ts` — 142 pass, 0 fail — proves the SDK opt-in is asserted and both a denied `PreToolUse(Bash)` and a successful `PostToolUse(Edit)` lifecycle reach the chunk pipeline with `hook_event` and outcome intact.
- `bun --filter '*' type-check` — clean across every workspace — proves the new SDK option is supported by the locked `@anthropic-ai/claude-agent-sdk` and does not regress any other workspace.
- `bun test src/dag-executor.test.ts` (workflows package) — 598 pass, 0 fail — proves the `hook_activity` persistence path remains green against the new emitter behavior.
- `bun x prettier --check packages/providers/src/claude/provider.ts packages/providers/src/claude/provider.test.ts` — both files conform.
- `bun x eslint packages/providers/src/claude/ --max-warnings 0` — clean.
- **Not verified:** A live end-to-end probe against a real Claude Code CLI was not re-run; the mocked regression tests cover the `includeHookEvents` opt-in, the surfaced `hook_started` / `hook_response` chunk shape, and the deliberately dropped `hook_progress` path, and the issue's reproduction (a denied `PreToolUse(Bash)`) is mirrored exactly by the second regression test.

## Links

- Closes #2324

## Summary by CodeRabbit

- **New Features**
  - Added visibility into tool hook activity, including hook start and completion events.
  - Hook events now include details such as hook name, event type, outcome, and exit code.
- **Bug Fixes**
  - Tool-scoped hook activity is now forwarded to the workflow activity stream.
  - Unsupported asynchronous hook progress events are excluded from the stream to prevent misleading activity records.

### DIFF
--- a/packages/providers/src/claude/provider.test.ts
+++ b/packages/providers/src/claude/provider.test.ts
@@ -926,6 +926,112 @@ describe('ClaudeProvider', () => {
       expect(callArgs.options).not.toHaveProperty('agentProgressSummaries');
     });
 
+    // --- Issue #2324 — tool-scoped hook frames must reach the audit stream -----
+
+    test('opts the SDK into lifecycle hook events for every surface (#2324)', async () => {
+      mockQuery.mockImplementation(async function* () {
+        // Empty
+      });
+
+      for await (const _ of client.sendQuery('test', '/workspace')) {
+        // consume
+      }
+
+      const callArgs = mockQuery.mock.calls[0][0] as { options: Record<string, unknown> };
+      // The SDK defaults `includeHookEvents` to false, which suppresses
+      // `hook_started` / `hook_response` for every hook type except
+      // SessionStart and Setup. Without an explicit opt-in, a node-level
+      // `PreToolUse` hook that denies Bash never reaches the workflow
+      // `hook_activity` stream.
+      expect(callArgs.options).toMatchObject({ includeHookEvents: true });
+    });
+
+    test('forwards a denied PreToolUse hook through the chunk pipeline (#2324)', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield {
+          type: 'system',
+          subtype: 'hook_started',
+          hook_id: 'h-1',
+          hook_name: 'Bash',
+          hook_event: 'PreToolUse',
+        };
+        yield {
+          type: 'system',
+          subtype: 'hook_response',
+          hook_id: 'h-1',
+          hook_name: 'Bash',
+          hook_event: 'PreToolUse',
+          outcome: 'error',
+          exit_code: 2,
+        };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks).toEqual([
+        {
+          type: 'hook_started',
+          hookId: 'h-1',
+          hookName: 'Bash',
+          hookEvent: 'PreToolUse',
+        },
+        {
+          type: 'hook_response',
+          hookId: 'h-1',
+          hookName: 'Bash',
+          hookEvent: 'PreToolUse',
+          outcome: 'error',
+          exitCode: 2,
+        },
+      ]);
+    });
+
+    test('forwards a PostToolUse hook lifecycle (#2324)', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield {
+          type: 'system',
+          subtype: 'hook_started',
+          hook_id: 'h-2',
+          hook_name: 'Edit',
+          hook_event: 'PostToolUse',
+        };
+        yield {
+          type: 'system',
+          subtype: 'hook_response',
+          hook_id: 'h-2',
+          hook_name: 'Edit',
+          hook_event: 'PostToolUse',
+          outcome: 'success',
+          exit_code: 0,
+        };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      expect(chunks).toEqual([
+        {
+          type: 'hook_started',
+          hookId: 'h-2',
+          hookName: 'Edit',
+          hookEvent: 'PostToolUse',
+        },
+        {
+          type: 'hook_response',
+          hookId: 'h-2',
+          hookName: 'Edit',
+          hookEvent: 'PostToolUse',
+          outcome: 'success',
+          exitCode: 0,
+        },
+      ]);
+    });
+
     test('handles tool_use with empty input', async () => {
       mockQuery.mockImplementation(async function* () {
         yield {
--- a/packages/providers/src/claude/provider.test.ts
+++ b/packages/providers/src/claude/provider.test.ts
@@ -1032,6 +1032,37 @@ describe('ClaudeProvider', () => {
       ]);
     });
 
+    test('drops hook_progress frames (only emitted for async hooks — none registered) (#2324)', async () => {
+      mockQuery.mockImplementation(async function* () {
+        yield {
+          type: 'system',
+          subtype: 'hook_progress',
+          hook_id: 'h-3',
+          hook_name: 'Bash',
+          hook_event: 'PreToolUse',
+          stdout: 'still running...',
+          stderr: '',
+          output: '',
+        };
+        yield {
+          type: 'assistant',
+          message: {
+            content: [{ type: 'text', text: 'Real response' }],
+          },
+        };
+      });
+
+      const chunks = [];
+      for await (const chunk of client.sendQuery('test', '/workspace')) {
+        chunks.push(chunk);
+      }
+
+      // The hook_progress frame is intentionally not surfaced: Archon
+      // registers only synchronous hooks, and the hooks guide documents
+      // the carve-out. Only the assistant message reaches the stream.
+      expect(chunks).toEqual([{ type: 'assistant', content: 'Real response' }]);
+    });
+
     test('handles tool_use with empty input', async () => {
       mockQuery.mockImplementation(async function* () {
         yield {
--- a/packages/providers/src/claude/provider.ts
+++ b/packages/providers/src/claude/provider.ts
@@ -847,8 +847,10 @@ function buildBaseClaudeOptions(
     // Opt into the SDK's hook lifecycle frames so that tool-scoped hooks
     // (PreToolUse / PostToolUse / Stop / etc.) reach the workflow audit
     // stream as `hook_activity` (#2324). SessionStart and Setup remain
-    // emitted regardless. The downstream `hook_started` / `hook_response`
-    // normalization already handles every frame this enables.
+    // emitted regardless. The downstream normalization surfaces
+    // `hook_started` and `hook_response`; the third subtype the SDK
+    // enables (`hook_progress`) falls through — it is only emitted for
+    // async hooks, which Archon does not register today.
     includeHookEvents: true,
     hooks: buildToolCaptureHooks(toolResultQueue),
     stderr: (data: string): void => {
--- a/packages/providers/src/claude/provider.ts
+++ b/packages/providers/src/claude/provider.ts
@@ -844,6 +844,12 @@ function buildBaseClaudeOptions(
     // Per-node override wins over the assistant-level default; the final
     // fallback stays ['project', 'user'] (the SDK-loading default Archon ships).
     settingSources,
+    // Opt into the SDK's hook lifecycle frames so that tool-scoped hooks
+    // (PreToolUse / PostToolUse / Stop / etc.) reach the workflow audit
+    // stream as `hook_activity` (#2324). SessionStart and Setup remain
+    // emitted regardless. The downstream `hook_started` / `hook_response`
+    // normalization already handles every frame this enables.
+    includeHookEvents: true,
     hooks: buildToolCaptureHooks(toolResultQueue),
     stderr: (data: string): void => {
       const output = data.trim();


Wirasm:
# Review Report — PR #2723 (fix(providers): surface Claude tool-scoped hook activity in the audit stream)

## Verdict

**Not ready.** The SDK→normalizer seam has one Important mismatch: the option the diff turns on enables three lifecycle subtypes per the SDK contract, but the reader handles only two — and the diff's own comment claims otherwise.

## Reviewed head SHA

`65cdf76f19258b0ecdca884075ada0aaf1ebd16d`

## Findings

### Important

#### R1 — `hook_progress` frames are silently dropped at the SDK→normalizer seam

**Claim.** The diff sets `includeHookEvents: true` (`packages/providers/src/claude/provider.ts:852`) and adds an explanatory comment at lines 847–851 that ends with *"The downstream `hook_started` / `hook_response` normalization already handles every frame this enables."* The SDK contract the option belongs to (`node_modules/.bun/@anthropic-ai+claude-agent-sdk@0.3.209+4b8d4561844a275b/node_modules/@anthropic-ai/claude-agent-sdk/sdk.d.ts:1578-1587`) enumerates **three** lifecycle subtypes — `hook_started`, `hook_progress`, `hook_response`. The provider normalizer's `else if` chain at `packages/providers/src/claude/provider.ts:1113-1132` recognizes only `hook_started` and `hook_response`. `subtype === 'hook_progress'` falls through to the catch-all at line 1134, which logs at `debug` level and yields no chunk.

**Evidence (verified).**

- SDK writer contract enumerates three subtypes: `sdk.d.ts:1580` — *"When true, `hook_started`, `hook_progress`, and `hook_response` system messages will be emitted…"*
- SDK message shape for the missing subtype: `SDKHookProgressMessage` at `sdk.d.ts:3791-3802` — `{ type: 'system'; subtype: 'hook_progress'; hook_id; hook_name; hook_event; stdout; stderr; output; uuid; session_id }`.
- Reader-side gap: `packages/providers/src/claude/provider.ts:1113, 1120` — only the two `else if (subtype === 'hook_started' …)` and `else if (subtype === 'hook_response' …)` branches exist; `grep -rn "subtype === 'hook" packages/` returns exactly those two matches.
- Comment overclaims: `packages/providers/src/claude/provider.ts:850-851` — *"…already handles every frame this enables."*
- Downstream types are equally narrow: `MessageChunk` union (`packages/providers/src/types.ts:354, 360`), `HookActivityEvent.activity` (`packages/workflows/src/event-emitter.ts:179-189`, `started | response`), `DagHookInfo.activity` (`packages/web/src/lib/types.ts:337-349`, `started | response`), `WorkflowHookActivityEvent.activity` (`packages/web/src/lib/types.ts:186-197`, `started | response`), Web bridge `case 'hook_activity'` mapping (`packages/server/src/adapters/web/workflow-bridge.ts:175-180`) — no path for `hook_progress`.

**Smallest correction.** Either tighten the diff comment at `provider.ts:850-851` to accurately describe the two handled subtypes and the documented carve-out for `hook_progress`, or close the seam: a third `else if (subtype === 'hook_progress' && sysMsg.hook_id)` in the provider normalizer mirroring the existing two, plus matching widening of `MessageChunk` (`types.ts`), `HookActivityEvent` (`event-emitter.ts`), the DAG executor branches (`dag-executor.ts:2608, 2636`), `DagHookInfo`/`WorkflowHookActivityEvent` (`web/src/lib/types.ts`), and the Web bridge mapping (`workflow-bridge.ts`). All six sites must move together — that is the point of the seam.

**Carve-out noted.** `packages/docs-web/src/content/docs/guides/hooks.md:308-310` explicitly states *"No hook event streaming — hook lifecycle events (`hook_started`, `hook_progress`) are not forwarded to the Web UI."* The carve-out was authored before this PR; the PR supersedes it for `hook_started` and was always silent about `hook_progress`. Today every Archon-registered hook (`PostToolUse` and `PostToolUseFailure` in `buildToolCaptureHooks` at `provider.ts:885-944`, plus YAML-declared hooks via `buildSDKHooksFromYAML` at `:454-476`) is synchronous, so the SDK does not currently emit `hook_progress` for any registered hook — making the seam mismatch **latent** (activated by the very option this diff turns on, unreachable on today's supported path). The defect is bounded by the docs carve-out, but the diff's own comment at `provider.ts:850-851` directly contradicts it, which is why this rises to Important rather than being a pre-existing limitation to acknowledge.

### Suggestions

None.

### Rejected findings

None.

## Prior findings

None — first review round.

## Lens roster

| Lens | Enabled | Status |
|---|---|---|
| `code` | yes | Ran. 0 findings. |
| `seams` | yes | Ran. 1 Important finding (R1). |
| `tests` | no | Disabled by scope — not run, recorded as disabled. |
| `errors` | no | Disabled by scope — not run, recorded as disabled. |
| `comments` | no | Disabled by scope — not run, recorded as disabled. |
| `types` | no | Disabled by scope — not run, recorded as disabled. |
| `docs` | no | Disabled by scope — not run, recorded as disabled. |
| `simplify` | no | Disabled by scope — not run, recorded as disabled. |

All enabled lenses delivered their reports. The two enabled lenses (`code`, `seams`) cover the surface area of this diff: a one-line SDK option plus three regression tests. The optional lenses most likely to surface additional evidence on a change of this shape would be `tests` (already covered by the new suite — `bun test packages/providers/src/claude/provider.test.ts` reports 141/141 pass) and `types` (would re-confirm the same `MessageChunk`/`HookActivityEvent` widening needed for R1, which `seams` already identifies). Their absence does not change the verdict.

## Next round

`seams` and `code` are sufficient for this diff shape; the next round should run **in full** (no delta mode) only if the diff scope changes. If the author chooses the "close the seam" path for R1, the next round should additionally enable `types` — the type widening crosses six files (`provider.ts`, `types.ts`, `event-emitter.ts`, `dag-executor.ts`, `web/src/lib/types.ts`, `workflow-bridge.ts`) and the verifier should re-check the union narrows correctly under the new variant.

PR #2728, opened 2026-08-22
refactor(workflows): the failed-producer guard now holds by construction

### LINKED ISSUE #2722: workflows: the failed-producer guard holds by KEEP-IN-SYNC enumeration, not by construction
## Problem

The invariant "a failed producer's output never silently resolves" (#2696/#2710, extended by #2713) is enforced correctly today, but only by convention. For the whole-text `\$node.output` form (no `.field`), four separate call sites each carry their own independently-worded `if (producer.state === 'failed') throw ...` guard:

- `resolveBindingDirective` (`dag-executor.ts`) — the original #2710 fix
- `wholeRefLogicalValue` (`dag-executor.ts`) — serves `resolveWorkflowValue`'s whole-ref tier
- `substituteNodeOutputRefs`'s unfielded branch (`dag-executor.ts`)
- `resolveOutputRef` (`condition-evaluator.ts`) — the `when:` resolver

They're tied together only by a prose comment in `output-ref.ts`'s module doc:

> KEEP IN SYNC: `wholeRefLogicalValue` and `substituteNodeOutputRefs` in dag-executor.ts, `resolveOutputRef` in condition-evaluator.ts, and `resolveBindingDirective`'s pre-existing #2710 guard in dag-executor.ts all assert the same invariant independently — a change to one should prompt checking the others.

For the fielded `.field` form, this is already solved correctly: `resolveNodeOutputField` in `output-ref.ts` is one shared, guarded function both consumers (`substituteNodeOutputRefs` and `resolveOutputRef`) route through — the invariant holds by construction there.

## Why this matters

The whole-text form's invariant holds only because someone remembered to add the check at every existing reader. Nothing stops a fifth reader of `ctx.nodeOutputs`/`nodeOutputs` — a new resolver, a new substitution surface, a future `output_format` extension — from reading `producer.output` directly and reproducing the exact incident class that #2696/#2710/#2713 have now fixed twice: a failed `loop_group`'s stale last-completed-iteration text (real, often non-empty, sometimes valid JSON) getting read as if the group had succeeded. Production already hit this once (dogfood run `6607bf20`, PR #2705 flipped to ready on a failed correction loop's stale output).

## Proposed direction

A guarded accessor on the map itself, mirroring what `resolveNodeOutputField` already does for the fielded case — something like `getNodeOutputOrThrow(nodeOutputs, nodeId)` that internally checks `state === 'failed'` and throws with a caller-appropriate message, so a new whole-text reader has to deliberately bypass the accessor to reintroduce the bug rather than just forgetting a check. The four existing call sites would route through it instead of each carrying their own copy; the KEEP-IN-SYNC comment becomes unnecessary once there's nothing left to keep in sync.

## Scope

Not urgent — the current enumeration is complete, cross-referenced, and covered by tests as of #2713 landing. This is preventive/structural follow-up, not a live bug. Flagged during review of #2713's delivery.

Wirasm:
## Problem and outcome

The invariant "a failed producer's output never silently resolves" for the whole-text `$node.output` form was enforced correctly, but only by convention: four call sites each carried their own independently-worded `if (producer.state === 'failed') throw ...` block, tied together only by a "KEEP IN SYNC" comment asking a future editor to remember to check the others.

- **Outcome:** All four sites (`wholeRefLogicalValue`, `resolveBindingDirective`, `substituteNodeOutputRefs`, `resolveOutputRef`) now route through one shared `assertProducerNotFailed` guard in `output-ref.ts`. A future fifth reader of `nodeOutputs` has to deliberately bypass that guard to reintroduce the bug, instead of simply forgetting to copy a check.
- **Invariant:** A failed producer's whole-text `$node.output` is never resolved as if it were trustworthy — unchanged from before, now enforced by construction rather than by enumeration.
- **Scope boundary:** No behavior change. Every thrown message's exact wording is preserved verbatim per call site (they legitimately differ in context — a binding directive names `if_skipped`, a `when:` guard names the condition, etc.). The "unknown node" handling at each site, which already differs intentionally per site, is untouched. The fielded (`.field`) form's guard inside `resolveNodeOutputField` was already shared and correct — not touched.

## Review guidance

- **Feedback requested:** Whether the value-level guard shape (`assertProducerNotFailed(nodeOutput, buildMessage)`) is the right consolidation, versus the issue's illustrative map+id sketch (`getNodeOutputOrThrow(nodeOutputs, nodeId)`) — see Solution below for why the value-level shape was chosen.
- **Start here:** `packages/workflows/src/output-ref.ts` — the new `assertProducerNotFailed` function and the rewritten module-doc paragraph it replaces.
- **Review order:** `output-ref.ts` (the shared guard) → `dag-executor.ts` (`wholeRefLogicalValue`, `resolveBindingDirective`, `substituteNodeOutputRefs`) → `condition-evaluator.ts` (`resolveOutputRef`) → `output-ref.test.ts` (new direct unit tests for the guard).
- **Lower-attention areas:** The four call-site edits are mechanical — each replaces an `if (state === 'failed') { throw new Error(...) }` block with a call to `assertProducerNotFailed(producer, failed => ...)` carrying the exact same template-literal message. `git diff` confirms the message text is byte-identical at every site.
- **Known risk or uncertainty:** None outstanding — see Solution for the one design decision already resolved.

## Solution

Added `assertProducerNotFailed(nodeOutput, buildMessage)` next to `resolveNodeOutputField` in `output-ref.ts`, which already solves the identical problem for the fielded (`.field`) form. It checks `nodeOutput.state === 'failed'` once and throws `buildMessage(nodeOutput)` — the callback lets each of the four call sites keep its own wording while sharing the mechanism.

The issue's illustrative signature was `getNodeOutputOrThrow(nodeOutputs, nodeId)` (map + id). Reading all four call sites showed each one already holds the resolved `NodeOutput` in hand by the time it needs to guard (from its own `.get()` or a passed-in parameter) — the "producer not found" handling, which intentionally differs per site, already happened. More importantly, two sites (`substituteNodeOutputRefs`, `resolveOutputRef`) must apply the guard *only* inside their unfielded branch: a fielded ref to a failed producer already throws a differently-worded `OutputRefError('producer-failed')` from `resolveNodeOutputField`'s own guard (pinned by an existing test — see Validation). A map+id accessor called at the point of lookup would either need a redundant second `.get()` or risk misfiring the wrong error for the fielded case. A value-level guard, mirroring `resolveNodeOutputField`'s own shape, avoids that trap and requires zero restructuring at any of the four sites.

## Validation

- `cd packages/workflows && bun test src/output-ref.test.ts` — 52 pass, 0 fail — includes 6 new tests directly covering `assertProducerNotFailed` (no-throw on non-failed states, throws the caller's message, passes the narrowed `failed` value with `.error` accessible without a cast).
- `cd packages/workflows && bun test src/dag-executor.test.ts` — 598 pass, 0 fail — includes the existing test pinning the fielded-vs-unfielded split for a failed producer (`substituteNodeOutputRefs`, lines 1008–1028: unfielded throws a generic `Error`, fielded throws `OutputRefError` with reason `'producer-failed'`) and the `#2696`/`#2713` binding-directive regression tests (lines 27775–27898), all passing unmodified.
- `cd packages/workflows && bun test src/condition-evaluator.test.ts` — 96 pass, 0 fail — includes the existing failed-producer tests for `resolveOutputRef` (lines 137–165), passing unmodified.
- `cd packages/workflows && bun test src/dry-run.test.ts` — 57 pass, 0 fail.
- `cd packages/workflows && bun run test` (full package suite) — every group passes, 0 fail.
- `cd packages/workflows && bun run type-check` — clean; `Extract<NodeOutput, { state: 'failed' }>` narrowing is sound, `.error` is accessible in `buildMessage` without a cast.
- `bun run lint` / `bun run format:check` (repo root) — clean.
- **Not verified:** Nothing material — this is a same-behavior structural refactor, and every pre-existing test asserting on the four sites' failed-producer messages passes unmodified, which is the direct proof that behavior did not change.

Plan: https://github.com/coleam00/Archon/issues/2722#issuecomment-5381631308

Closes #2722

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented workflows from using stale output values when a producing step has failed.
  * Improved consistency of failure handling across workflow output references, bindings, and prompt or script substitutions.
  * Preserved clear, context-specific error messages when failed outputs are referenced.

### DIFF
--- a/packages/workflows/src/condition-evaluator.ts
+++ b/packages/workflows/src/condition-evaluator.ts
@@ -36,6 +36,7 @@ import type { NodeOutput } from './schemas';
 import { createLogger } from '@archon/paths';
 import {
   resolveNodeOutputField,
+  assertProducerNotFailed,
   OutputRefError,
   similarNodeIds,
   canonicalValueText,
@@ -90,16 +91,16 @@ function resolveOutputRef(
   if (!field) {
     // A failed producer's stale output is never evaluated (#2713) — a when: condition
     // joined past a failure via trigger_rule: all_done must branch on a different
-    // node's output, not the failed producer's own leftover text. KEEP IN SYNC: see
-    // the module doc of resolveNodeOutputField in output-ref.ts for the full list of
-    // guards asserting this same invariant.
-    if (nodeOutput.state === 'failed') {
-      throw new Error(
-        `'$${nodeId}.output' references node '${nodeId}', but it failed (${nodeOutput.error}), ` +
-          "so its output cannot be trusted in a 'when:' condition. A failed producer's stale " +
-          "output is never evaluated — branch on a different node's output instead."
-      );
-    }
+    // node's output, not the failed producer's own leftover text. Routes through
+    // `assertProducerNotFailed` in output-ref.ts (#2722), which every whole-text reader
+    // of nodeOutputs shares.
+    assertProducerNotFailed(
+      nodeOutput,
+      failed =>
+        `'$${nodeId}.output' references node '${nodeId}', but it failed (${failed.error}), ` +
+        "so its output cannot be trusted in a 'when:' condition. A failed producer's stale " +
+        "output is never evaluated — branch on a different node's output instead."
+    );
     // For unfielded ref, structuredOutput shape is opaque — defer to output text.
     if (!nodeOutput.output) return '';
     return nodeOutput.output;
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -94,6 +94,7 @@ import { evaluateCondition } from './condition-evaluator';
 import {
   declaredFieldsFromSchema,
   resolveNodeOutputField,
+  assertProducerNotFailed,
   OutputRefError,
   similarNodeIds,
   canonicalValueText,
@@ -276,9 +277,9 @@ function inputEnvVars(node: DagNode, ctx: ShellInputContext): NodeJS.ProcessEnv 
  * already guards its own call below before reaching this function, so the check below only
  * ever fires for this function's OTHER caller, `resolveWorkflowValue`'s whole-ref tier — the
  * plain (non-directive) `with:` value on a script/command node, a `workflow:` node's `with:`,
- * or `fan_out`'s static `with:`, none of which had #2710's guard. KEEP IN SYNC: see the
- * module doc of `resolveNodeOutputField` in output-ref.ts for the full list of guards
- * asserting this same invariant.
+ * or `fan_out`'s static `with:`, none of which had #2710's guard. Routes through
+ * `assertProducerNotFailed` in output-ref.ts (#2722), which every other whole-text reader
+ * of `nodeOutputs` shares.
  */
 function wholeRefLogicalValue(
   producer: NodeOutput,
@@ -286,13 +287,13 @@ function wholeRefLogicalValue(
   field: string | undefined
 ): JsonValue {
   if (field === undefined) {
-    if (producer.state === 'failed') {
-      throw new Error(
+    assertProducerNotFailed(
+      producer,
+      failed =>
         `Binding value '$${nodeId}.output' references node '${nodeId}', but it failed ` +
-          `(${producer.error}), so its output cannot be trusted. Fix the failure, or guard ` +
-          "the referencing node with a 'when:' condition that excludes the failed branch."
-      );
-    }
+        `(${failed.error}), so its output cannot be trusted. Fix the failure, or guard ` +
+        "the referencing node with a 'when:' condition that excludes the failed branch."
+    );
     const structured = 'structuredOutput' in producer ? producer.structuredOutput : undefined;
     // Provider payloads are ajv-validated / DB-round-tripped JSON, so the cast
     // asserts what production already guarantees.
@@ -455,18 +456,18 @@ function resolveBindingDirective(
   // a branch that legitimately didn't run, not for a run that ran and produced a result
   // that can't be trusted (a loop_group's failure paths carry the last completed
   // iteration's real output text — non-empty, often valid JSON — which would otherwise
-  // resolve here as if the group had succeeded). KEEP IN SYNC: see the module doc of
-  // resolveNodeOutputField in output-ref.ts for the full list of guards (#2713 extended
-  // this one, the original, to every other reader of nodeOutputs).
-  if (producer.state === 'failed') {
-    throw new Error(
+  // resolve here as if the group had succeeded). Routes through `assertProducerNotFailed`
+  // in output-ref.ts (#2722, extending #2710's original guard to every whole-text reader
+  // of nodeOutputs through one shared function).
+  assertProducerNotFailed(
+    producer,
+    failed =>
       `Node '${consumerId}' binding '${name}' reads '${directive.from}', but node ` +
-        `'${ref.nodeId}' failed (${producer.error}), so its output cannot be trusted. ` +
-        "A binding never falls back to 'if_skipped' for a failed producer — fix the " +
-        `failure, or guard '${consumerId}' with a 'when:' condition that excludes the ` +
-        'failed branch.'
-    );
-  }
+      `'${ref.nodeId}' failed (${failed.error}), so its output cannot be trusted. ` +
+      "A binding never falls back to 'if_skipped' for a failed producer — fix the " +
+      `failure, or guard '${consumerId}' with a 'when:' condition that excludes the ` +
+      'failed branch.'
+  );
   return wholeRefLogicalValue(producer, ref.nodeId, ref.field);
 }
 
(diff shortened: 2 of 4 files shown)


Wirasm:
## Ready to merge

This is a correction-verification pass on top of an already-clean mechanical refactor. The previous review (head `07c29ae3`) found zero blocking issues and three non-blocking Suggestions. Since then, one commit (`f90a64a2`, doc-only) landed, and the implementer resolved all three: R1 fixed, R2 and R3 declined with reasoning the original reviewers' own evidence already supports. All three dispositions were independently re-verified against the new head by the same reviewer scopes that raised them, plus a regression scan of the two files not directly touched by the fix commit. No new defect was found, and no prior finding reopened.

**Validation:** All commands below ran against `reviewed_head` (`f90a64a2`) in an isolated detached worktree, after `bun install` (a fresh worktree needs its own `node_modules`). All passed.

**Resolved:** 3 (`R1` fixed, `R2` and `R3` declined and confirmed reasonable)

### Findings

|---|---|---|---|
| `R1` | Suggestion | Module doc said the invariant is "enforced by construction," but the guard is a void-returning runtime check | FIXED |
| `R2` | Suggestion | New test "narrowed failed NodeOutput" proves the same fact as the preceding "custom message" test | DECLINED |
| `R3` | Suggestion | `assertProducerNotFailed` doesn't use an `asserts` predicate return type, unlike `assertValidTierName` | DECLINED |

<details>
<summary><code>R1</code> — "Enforced by construction" overclaimed what a void-returning assertion guarantees</summary>

**Impact:** No functional consequence, ever — this was a documentation-precision gap: a reader trusted the module doc's claim that the invariant could no longer be silently reintroduced, when a future fifth whole-text reader of `nodeOutputs` could still bypass the guard entirely and every tool (tsc/eslint/tests) would pass.

**Evidence (original):** `packages/workflows/src/output-ref.ts:33-38` (module doc) and `:388-396` (`assertProducerNotFailed` signature, `void` return).

**Fix (commit `f90a64a2`):** Both doc comments in `packages/workflows/src/output-ref.ts` were reworded. The module doc (lines 33-38) now reads "...as one shared function every caller routes through (#2722), replacing the enumeration of independently-worded call sites that each had to remember the check" — no compile-time claim. `assertProducerNotFailed`'s own doc comment (lines 375-390) now states outright: "This is a runtime check, not a type-level one — nothing stops a future caller from reading `nodeOutput.output` directly without calling this function first; the value is having one place to route through, not a compiler-enforced guarantee against bypass."

**Verification:** Re-reviewed by `prp-core:seam-analyzer` against the new head. The new wording no longer implies a type-level guarantee, and does not overcorrect into implying the mechanism is worthless — it correctly credits the real, verifiable value (one shared function replacing four independently-worded call sites) while naming the actual residual risk precisely. A regression scan of the full PR diff (`dev`→`f90a64a2`) for the same overclaim pattern in `dag-executor.ts` (3 call-site comments) and `condition-evaluator.ts` (1 call-site comment) found none — all four consistently say "Routes through `assertProducerNotFailed` ... which every whole-text reader of nodeOutputs shares," matching the corrected, calibrated language.

**Found by:** `prp-core:seam-analyzer`, `prp-core:type-design-analyzer` (original); re-verified by `prp-core:seam-analyzer`

**Disposition:** FIXED — confirmed adequate, no opposite-direction overclaim or new inaccuracy introduced.

</details>

<details>
<summary><code>R2</code> — Minor test redundancy in the new <code>assertProducerNotFailed</code> coverage</summary>

**Impact:** None on correctness — cosmetic duplication only.

**Evidence:** `packages/workflows/src/output-ref.test.ts:191-208` (unchanged since the original review).

**Disposition rationale (implementer):** Both tests were already rated by the original reviewer as having standalone value ("the second test also documents the type-narrowing decision explicitly, which has standalone value on a PR whose whole point is a typed contract... not required for merge"); this was a Suggestion-severity preference, not a defect. No code change made.

**Verification:** Re-reviewed by `prp-core:code-simplifier`. Confirmed `output-ref.test.ts` is byte-identical to the previous reviewed head — the decline is genuine, not a partial or half-applied fix. Independently re-read both tests against the schema (`workflow-run.ts:114-121`, `error` is required and non-optional on the `'failed'` variant): there is no reachable state where the two tests diverge in what they'd catch — both exercise the identical guarantee (narrowed `.error` access without a cast) through different observation mechanisms (message interpolation vs. a captured side-effect variable). No masked correctness gap found.

**Disposition:** DECLINED — confirmed reasonable; the redundancy is real but harmless, and closing it is a style preference the original review already rated optional.

</details>

<details>
<summary><code>R3</code> — <code>assertProducerNotFailed</code> omits the <code>asserts</code> predicate return type used by the repo's own <code>assertX</code> precedent</summary>

**Impact:** None reachable — the original review traced all four call sites and found none needs the narrowing today.

**Evidence:** `packages/workflows/src/output-ref.ts:391-398` (signature, unchanged); `packages/workflows/src/model-validation.ts:106-110` (the `assertValidTierName` `asserts`-predicate precedent, 3 of 4 sibling helpers in that file already use plain `void`).

**Disposition rationale (implementer):** The original review's own evidence found no current consumer needs the narrowing, and 3 of 4 existing `assertX` helpers in `model-validation.ts` already use plain `void`, so omitting `asserts` is not even inconsistent with the majority precedent. Adding it now would be speculative generality with no current caller benefit (YAGNI). No code change made.

**Verification:** Re-reviewed by `prp-core:type-design-analyzer`. Confirmed the signature is byte-identical to the previous reviewed head (`void`, no `asserts` predicate) and that `dag-executor.ts`/`condition-evaluator.ts` are unchanged since the original review, so all four call sites' "no current need" status still holds. The R1 doc fix (which now explicitly names the bypass-by-omission risk) does not sharpen the case for `asserts` — `asserts` only narrows the type in code that runs *after* a call to the function in the same scope; it cannot force a call to happen at all, so it would not close the specific risk the doc now names. If that risk mattered enough to close structurally, the correct fix would be a parse-at-the-boundary type, not an `asserts` predicate — a materially bigger change nothing in this diff creates a concrete need for.

**Disposition:** DECLINED — confirmed reasonable; the doc-honesty improvement in R1's fix does not create new pressure to add `asserts`, and no call site has grown a reason to want it.

</details>

<details>

#### Reviewer coverage

Correction-verification pass. Scopes retained: the three that owned the disputed findings (`seams`/R1, `simplify`/R2, `types`/R3). `code` had no findings in the original review and was not repeated per the skill's correction-verification rule.

| Scope | Result |
|---|---|
| `seams` (`prp-core:seam-analyzer`) | R1 verified FIXED; regression scan of `dag-executor.ts`/`condition-evaluator.ts` clean |
| `simplify` (`prp-core:code-simplifier`) | R2 decline confirmed reasonable, no new problem |
| `types` (`prp-core:type-design-analyzer`) | R3 decline confirmed reasonable, no new problem |

#### Validation

|---|---|---|
| `cd packages/workflows && bun test src/output-ref.test.ts` | PASS | 52 pass, 0 fail |
| `cd packages/workflows && bun test src/condition-evaluator.test.ts` | PASS | 96 pass, 0 fail |
| `cd packages/workflows && bun test src/dag-executor.test.ts` | PASS | 598 pass, 0 fail |
| `bun run type-check` (repo root, all packages) | PASS | Clean across all packages |
| `bun run lint` (repo root) | PASS | `eslint . --cache` — no warnings or errors |
| `bun run format:check` (repo root) | PASS | Prettier — all matched files formatted |

All validation ran in a detached worktree (`git worktree add --detach`) pinned to `reviewed_head` (`f90a64a2`), independent of the implementer's own reported results. The worktree required `bun install` before validation (fresh checkout, no `node_modules`).

</details>

PR #2729, opened 2026-08-22
fix(workflows): bash and script exec nodes now match on loop input delivery and output truncation

### LINKED ISSUE #2725: workflows: $LOOP_USER_INPUT never reaches a bash: node inside a loop_group body
## Problem

Inside a `loop_group` body, `$LOOP_USER_INPUT` (the free-text a user supplies via `/workflow approve <id> <text>` at the interactive gate) is delivered to `script:` nodes via `ctx.bodyLoopUserInput`, but never to `bash:` nodes — a bash node in the same position always sees an empty string.

## Evidence

`executeScriptNode` (`packages/workflows/src/dag-executor.ts`) takes a `loopUserInput` parameter and writes it into the subprocess env:

```ts
subprocessEnv.LOOP_USER_INPUT = loopUserInput; // sourced from ctx.bodyLoopUserInput at the call site
```

`executeBashNode` has no such parameter at all — its subprocess env hardcodes:

```ts
LOOP_USER_INPUT: '',
```

At the dispatch site (`case 'exec':` in `runLayers`), `ctx.bodyLoopUserInput` is read exactly once, only for the `executeScriptNode` call — `executeBashNode`'s call never receives it. Both node types skip literal $-substitution of `$LOOP_USER_INPUT` in the script source (`shellSafe: true`, #2115) and rely on the env var instead, so for `bash:` there is currently no working channel at all — a `bash:` node that does `echo "$LOOP_USER_INPUT"` inside a loop_group body always prints nothing, regardless of what the user typed at the gate.

Confirmed by test coverage: `dag-executor.test.ts` has a test asserting $LOOP_USER_INPUT delivery to a resumed body **script** node ("delivers $LOOP_USER_INPUT to a resumed body script via env, never spliced into source (#2115)", `runtime: 'bun'`) with no equivalent test for `runtime: 'sh'`.

## Fix direction

Thread `ctx.bodyLoopUserInput` into the `executeBashNode` call the same way it already reaches `executeScriptNode`, and set `LOOP_USER_INPUT` in bash's subprocess env from that value instead of hardcoding `''`. Add a bash-runtime counterpart to the existing script-runtime loop_group test.

## Context

Surfaced during the #2718 unify-vs-decline investigation (side-by-side read of `executeBashNode`/`executeScriptNode`) — a real, currently-shipped behavioral gap, not an intentional runtime difference. Filed separately since fixing it is an independent, narrowly-scoped change regardless of how #2718 resolves.

### LINKED ISSUE #2726: workflows: script: node persisted output is never truncated, unlike bash: nodes
## Problem

A `bash:` node's stdout is capped before being written into the persisted `node_completed` workflow event (`formatPersistedBashOutput`, 32KB cap with a truncation marker + `node_output_truncated`/`node_output_original_bytes` metadata, guarding against oversized DB rows). A `script:` node (`runtime: 'bun'` or `'uv'`) has no equivalent — its full, unbounded stdout is written directly into `data.node_output`.

## Evidence

`executeBashNode` (`packages/workflows/src/dag-executor.ts`):

```ts
const persistedOutput = formatPersistedBashOutput(output);
// ...
data: { duration_ms: duration, type: 'bash', node_output: persistedOutput.nodeOutput, ...(persistedOutput.truncated ? {...} : {}) }
```

`executeScriptNode`:

```ts
data: { duration_ms: duration, type: 'script', node_output: output, ...iterationData }
```

`formatPersistedBashOutput`/`PERSISTED_BASH_OUTPUT_MAX_BYTES` (32 * 1024 bytes) is defined once and called from exactly one place — the bash path.

Confirmed by test coverage: `dag-executor.test.ts` has 4 tests covering the truncation cap (byte-cap boundary, over-cap marker + metadata, UTF-8-safe splitting, full untruncated output still used for same-run `when`/substitution) and every one of them uses `runtime: 'sh'`. No script-runtime (`bun`/`uv`) equivalent exists.

A `script:` node that prints a large amount to stdout (e.g. dumping a file, a verbose build log, a large JSON blob) writes it unbounded into the workflow_events row, with none of the protection bash nodes already have.

## Fix direction — CORRECTED, naive parity is worse than the status quo

The originally proposed fix ("call `formatPersistedBashOutput` from `executeScriptNode` too, mirroring the bash path exactly") was implemented and then reverted before merge, once the following load-bearing fact came to light: the persisted event copy is not just audit data — it's the **resume dataflow source**. `getDagResumeSnapshot` (`packages/core/src/db/workflow-events.ts:258`) rebuilds `completedNodeOutputs` — the map that pre-populates `$node.output` for every consumer on resume — directly from the persisted `node_completed` rows' `node_output`.

That changes the picture in two ways:

1. **The existing bash cap already carries a latent bug.** A fresh run's consumer reads the full in-memory output; a resumed run's consumer reads the 32KB-truncated persisted text, truncation marker spliced in. That's the same fresh/resume divergence class #2587 fixed for loop groups, sitting latent on every large bash output today.
2. **Naive parity makes it worse, not better.** Script nodes are the most dangerous place to truncate, because their whole purpose is emitting structured JSON that downstream `$node.output.field` access parses from text — and #2637's strict field access fails the consuming node when output isn't valid JSON. Truncated JSON is not JSON. So the naive fix converts "resumed consumer sees clipped text" into "resumed consumer hard-fails," on exactly the node type authors use for computed data.

The problem this issue names (unbounded DB rows are a real hazard) is still real — the fix shape was just wrong. The right direction is the one the codebase already half-has: cap the in-row copy as a **preview**, preserve the full bytes **by reference**, and make resume rehydration prefer the reference. The spill mechanism already exists for a different purpose — `$ARTIFACTS_DIR/.archon/node-output-spills/` already holds oversized outputs for bash substitution. Extend it:

- When output exceeds the cap (either runtime), write the full bytes to the run's artifacts spill directory.
- Record the spill path + a preview in the persisted event, same as today's truncation marker.
- `getDagResumeSnapshot` reads the spill back when present, instead of the truncated preview.

Truncation then becomes genuinely cosmetic, DB rows stay genuinely bounded, and bash's latent resume bug gets fixed in the same motion — for both runtimes uniformly, not just script. Retention is coherent too: spills live with the run's artifacts, and a run whose artifacts are gone can't meaningfully resume anyway.

This is materially bigger than the original "add a truncation call" scope — it touches resume dataflow correctness (`workflow-events.ts`), not just the two `executeBashNode`/`executeScriptNode` persist sites. Treat it as its own workstream with its own plan and review, not a quick mirror of the bash path.

## Context

Surfaced during the #2718 unify-vs-decline investigation (side-by-side read of `executeBashNode`/`executeScriptNode`) — a real, currently-shipped behavioral gap, not an intentional runtime difference. Filed separately since fixing it is an independent, narrowly-scoped change regardless of how #2718 resolves. Fix direction corrected after the naive version was caught before merge (see above) — do not reintroduce `formatPersistedBashOutput`/`formatPersistedNodeOutput` applied uniformly without the spill-by-reference mechanism; it silently turns clipped bash text into hard failures on script nodes.

Wirasm:
## Problem and outcome

`executeBashNode` and `executeScriptNode` in `dag-executor.ts` are sibling implementations of the same `exec` node kind (bash vs. script runtime) that had drifted apart on two behaviors.

- **Outcome:** A `bash:` node inside an interactive `loop_group` body now receives the reviewer's resumed feedback text through its subprocess environment, matching `script:` nodes. A `script:` node's persisted output is now capped at the same 32KB boundary as `bash:` nodes, with the same truncation metadata.
- **Invariant:** `bash:` and `script:` `exec` nodes behave identically with respect to which loop-scoped variables reach their subprocess environment and how their successful stdout is capped before persistence.
- **Scope boundary:** No change to `executeScriptNode`'s existing loop-input behavior (it was already correct), the truncation marker format or byte cap value, or the reader side (`output-ref.ts`) — both were already runtime-agnostic.
- **Root cause:**
  - `#2725` — `executeBashNode` hardcoded `LOOP_USER_INPUT: ''` in its subprocess env and its call site never passed the computed per-iteration value, while `executeScriptNode` already had a `loopUserInput` parameter wired from `ctx.bodyLoopUserInput`. Note: a *plain* `$LOOP_USER_INPUT` reference in a bash body script already worked pre-fix through a separate, pre-existing mechanism (`applyLoopPrevToBodyNode` shell-quotes and splices it directly into the script source before dispatch). The env var this PR adds is what a `${LOOP_USER_INPUT}` (braced) reference or a nested subprocess reading the inherited environment needs — that path was genuinely broken.
  - `#2726` — `executeScriptNode` persisted its full, unbounded stdout into the `node_completed` event, while `executeBashNode` already ran output through a 32KB UTF-8-safe truncation helper before persisting.

## Review guidance

- **Feedback requested:** Correctness of the two fixes and whether the parity they establish is complete.
- **Start here:** `packages/workflows/src/dag-executor.ts:3290` (`executeBashNode`) — the loop-input parameter and env wiring.
- **Review order:** `executeBashNode` signature/env (loop input) → `case 'exec':` call site → shared truncation helper rename → `executeScriptNode`'s new truncation call → new tests.
- **Lower-attention areas:** The helper/constant rename (`formatPersistedBashOutput`/`PERSISTED_BASH_OUTPUT_MAX_BYTES` → `formatPersistedNodeOutput`/`PERSISTED_NODE_OUTPUT_MAX_BYTES`) is mechanical — logic unchanged, just no longer bash-specific since it now serves both call sites.

## Solution

`executeBashNode` gains a `loopUserInput = ''` parameter positioned identically to `executeScriptNode`'s (immediately before `execContext`), used in place of the hardcoded `''` in `subprocessEnv.LOOP_USER_INPUT`. The `case 'exec':` dispatch's bash branch now passes `ctx.bodyLoopUserInput ?? ''`, exactly mirroring the existing script branch.

The truncation helper (`formatPersistedBashOutput`) and its byte-cap constant (`PERSISTED_BASH_OUTPUT_MAX_BYTES`) are renamed to runtime-neutral names since a second caller now exists, and `executeScriptNode` calls it before building its `node_completed` event, adding the same conditional `node_output_truncated`/`node_output_original_bytes` fields `executeBashNode` already emits. No truncation logic is duplicated — both node types share the one helper. The shared-marker doc comment in `utils/output-truncation.ts` is updated to name the new writer and both node types.

## Validation

- `bun test src/dag-executor.test.ts -t "resumed body bash node via env"` — passed — proves #2725: a bash body node's `${LOOP_USER_INPUT}` env read now carries the resumed reviewer feedback. Verified this test fails on pre-fix code (`Expected: "looks good, ship it" / Received: ""`).
- `bun test src/dag-executor.test.ts -t "caps over-limit persisted script output"` — passed — proves #2726: a script node's persisted output is capped at 32,768 bytes with the correct truncation metadata. Verified this test fails on pre-fix code (`Expected: <= 32768 / Received: 32769`).
- `bun test src/dag-executor.test.ts` (full file) — 601 pass, 0 fail — no regressions in the existing suite, including the pre-existing bash truncation and script loop-input tests these fixes converge onto.
- `bun run type-check` — passed — renamed identifiers and the new parameter type-check cleanly.
- `bun run validate` — exit 0 — full repository gate (lint, format, type-check, complete test suite) green.
- **Not verified:** Nothing material — both fixes are covered by regression tests proven to fail before the fix and pass after it.

## Links

- Closes #2725
- Closes #2726

## Summary by CodeRabbit

* **Bug Fixes**
  * Limited persisted Bash and script output to 32 KiB while preserving valid UTF-8 and indicating when content was truncated.
  * Ensured output at or below the limit remains unchanged.
  * Preserved loop input when resuming Bash workflow nodes.

* **Documentation**
  * Clarified truncation behavior and audit-preview retention for Bash and script output.
  * Updated security guidance for persisted execution output.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -1029,7 +1029,7 @@ load-time `<include>__<node>` ID in metadata and as the sanitized body suffix.
 
 This works on **every** node type (`bash`/`script` produce typed outputs too, just without a `sessionId`). The write is **best-effort** — if it fails, the node still succeeds and a warning is logged; the typed sidecar may simply be absent. `output_type` is an open set of labels (`plan`, `findings`, `code`, `summary`, …) — pick a convention and keep casing consistent, since lookup is case-sensitive.
 
-Successful bash stdout is retained by default on the completed run as a bounded audit preview in `node_completed.data.node_output`. Output over 32 KiB (32,768 UTF-8 bytes) ends with a truncation marker, and the event also includes `node_output_truncated: true` plus `node_output_original_bytes`. Because stdout is persisted, never print secrets or credentials from bash nodes. This preview is separate from `output_type`: declaring `output_type` opts into a best-effort file sidecar that may contain the full output and is not required for ordinary bash audit retention.
+Successful `bash`/`script` stdout is retained by default on the completed run as a bounded audit preview in `node_completed.data.node_output`. Output over 32 KiB (32,768 UTF-8 bytes) ends with a truncation marker, and the event also includes `node_output_truncated: true` plus `node_output_original_bytes`. Because stdout is persisted, never print secrets or credentials from a bash or script node. This preview is separate from `output_type`: declaring `output_type` opts into a best-effort file sidecar that may contain the full output and is not required for ordinary audit retention.
 
 ---
 
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -6015,6 +6015,107 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
     // complete cross-process output requires a separately managed artifact.
   });
 
+  it('caps over-limit persisted script output with a marker and byte metadata (#2726)', async () => {
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const workflowRun = makeWorkflowRun('script-output-over-cap');
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-script-over-cap',
+      testDir,
+      {
+        name: 'script-output-over-cap',
+        nodes: [
+          {
+            id: 'over-cap',
+            kind: 'exec',
+            runtime: 'bun',
+            script: 'console.log("x".repeat(32769))',
+          },
+        ],
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls;
+    const completedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_completed' &&
+        (call[0] as { step_name: string }).step_name === 'over-cap'
+    );
+    const data = (
+      completedEvent![0] as {
+        data: {
+          node_output: string;
+          node_output_truncated: boolean;
+          node_output_original_bytes: number;
+        };
+      }
+    ).data;
+    expect(Buffer.byteLength(data.node_output, 'utf8')).toBeLessThanOrEqual(32_768);
+    expect(data.node_output).toEndWith('\n\n… [truncated; original output was 32769 bytes]');
+    expect(data.node_output_truncated).toBe(true);
+    expect(data.node_output_original_bytes).toBe(32_769);
+  });
+
+  it('persists script output at or below the byte cap unchanged without truncation metadata', async () => {
+    const store = createMockStore();
+    const mockDeps = createMockDeps(store);
+    const workflowRun = makeWorkflowRun('script-output-below-cap');
+
+    await executeDagWorkflow(
+      mockDeps,
+      createMockPlatform(),
+      'conv-script-below-cap',
+      testDir,
+      {
+        name: 'script-output-below-cap',
+        nodes: [
+          {
+            id: 'below-cap',
+            kind: 'exec',
+            runtime: 'bun',
+            script: 'console.log("x".repeat(32767))',
+          },
+        ],
+      },
+      workflowRun,
+      'claude',
+      undefined,
+      join(testDir, 'artifacts'),
+      join(testDir, 'state'),
+      join(testDir, 'logs'),
+      'main',
+      'docs/',
+      minimalConfig
+    );
+
+    const eventCalls = (store.createWorkflowEvent as ReturnType<typeof mock>).mock.calls;
+    const completedEvent = eventCalls.find(
+      (call: unknown[]) =>
+        (call[0] as { event_type: string }).event_type === 'node_completed' &&
+        (call[0] as { step_name: string }).step_name === 'below-cap'
+    );
+    const data = (
+      completedEvent![0] as {
+        data: Record<string, unknown> & { node_output: string };
+      }
+    ).data;
+    expect(data.node_output).toBe('x'.repeat(32767));
+    expect(data.node_output_truncated).toBeUndefined();
+    expect(data.node_output_original_bytes).toBeUndefined();
+  });
+
   it('keeps a persisted UTF-8 preview valid when the byte cap splits a code point', async () => {
     const store = createMockStore();
     const mockDeps = createMockDeps(store);
@@ -19730,6 +19831,84 @@ describe('executeDagWorkflow -- loop_group node', () => {
       expect(scriptCall?.[1][2]).not.toContain('process.exit');
       // The per-iteration feedback reaches the script through the environment.
       expect(scriptCall?.[2].env.LOOP_USER_INPUT).toBe(injection);
+    } finally {
+      execSpy.mockRestore();
+    }
+  });
+
+  it('delivers $LOOP_USER_INPUT to a resumed body bash node via env (#2725)', async () => {
+    // A `${VAR}` (braced) reference is NOT matched by applyLoopPrevToBodyNode's literal
+    // `$LOOP_USER_INPUT` splice — bash resolves it from the subprocess environment at
+    // runtime instead, so this specifically proves the env-delivery channel this fix adds
+    // (a plain unbraced `$LOOP_USER_INPUT` reference was already spliced pre-fix and is
+    // not a regression risk here).
+    const feedback = 'looks good, ship it';
+    const execSpy = spyOn(git, 'execFileAsync').mockResolvedValue({ stdout: 'DONE\n', stderr: '' });
+    try {
+      const mockDeps = createMockDeps();
+      const platform = createMockPlatform();
+      const workflowRun = makeWorkflowRun('lg-userinput-bash', {
+        metadata: {
+          approval: {
+            type: 'interactive_loop',
+            nodeId: 'grp',
+            iteration: 0,
+            message: 'Review and provide feedback.',
+          },
+          loop_user_input: feedback,
+          loop_feedback_given: true,
+        },
+      });
+
+      const nodes: DagNode[] = [
+        {
+          id: 'grp',
+          kind: 'loop_group',
+          loop_group: {
+            until: 'DONE',
+            max_iterations: 3,
+            fresh_context: true,
+            interactive: true,
+            gate_message: 'Review and provide feedback.',
+            nodes: [
+              {
+                id: 'emit',
+                kind: 'exec',
+                script: 'echo "${LOOP_USER_INPUT}"; echo DONE',
+                runtime: 'sh',
+                depends_on: [],
+              },
+            ],
+          },
+          depends_on: [],
+        },
+      ];
+
+      await executeDagWorkflow(
+        mockDeps,
+        platform,
+        'conv-lg-userinput-bash',
+        testDir,
+        { name: 'lg-userinput-bash', nodes },
+        workflowRun,
+        'claude',
+        undefined,
+        join(testDir, 'artifacts'),
+        join(testDir, 'state'),
+        join(testDir, 'logs'),
+        'main',
+        'docs/',
+        minimalConfig
+      );
+
+      const bashCall = execSpy.mock.calls.find(
+        c => (c[0] as string) === git.resolveBashPath() && (c[1] as string[])[0] === '-c'
+      ) as [string, string[], { env: NodeJS.ProcessEnv }] | undefined;
+      expect(bashCall).toBeDefined();
+      // The braced reference is left untouched by the source-splice pass...
+      expect(bashCall?.[1][1]).toBe('echo "${LOOP_USER_INPUT}"; echo DONE');
+      // ...so the subprocess environment is the only channel that can deliver it.
+      expect(bashCall?.[2].env.LOOP_USER_INPUT).toBe(feedback);
     } finally {
       execSpy.mockRestore();
     }
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -6069,51 +6069,56 @@ describe('executeDagWorkflow -- resume with priorCompletedNodes', () => {
   });
 
   it('persists script output at or below the byte cap unchanged without truncation metadata', async () => {
-    const store = createMockStore();
-    const mockDeps = createMockDeps(store);
-    const workflowRun = makeWorkflowRun('script-output-below-cap');
+    for (const [nodeId, byteCount] of [
+      ['below-cap', 32_767],
+      ['exact-cap', 32_768],
+    ] as const) {
+      const store = createMockStore();
+      const mockDeps = createMockDeps(store);
+      const workflowRun = makeWorkflowRun(`script-output-${nodeId}`);
 
-    await executeDagWorkflow(
-      mockDeps,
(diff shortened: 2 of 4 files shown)


Wirasm:
## Ready to merge

Correction verification pass over commit `04da8f94`, which addresses all four non-blocking Suggestions from the prior full review (head `f06eadee`). All four dispositions are confirmed by independent evidence, not just by re-reading the commit message: the four rewritten `$LOOP_USER_INPUT` delivery comments (R1) were checked line-by-line against `applyLoopPrevToBodyNode`'s actual splice/env branch logic and are now factually accurate (bash gets both an env channel and a splice for the unbraced literal; script gets env only); the docs wording (R2) now correctly names both `bash` and `script` for persisted-stdout truncation, matching the code path that already persists both identically; the script truncation test (R4) is now genuinely parametrized over both 32,767 and 32,768 bytes, mirroring the bash precedent test byte-for-byte; and the declined `subprocessEnv` extraction (R3) remains correctly declined — the correction touched zero lines of either duplicated block, so the rule-of-three reasoning is unchanged. No new defects were introduced by the correction commit itself. `open_findings` is now 0.

**Validation:** All PASS — `bun test src/dag-executor.test.ts` (packages/workflows, 601 pass / 0 fail, 1697 `expect()` calls, run independently by all three reviewers), `tsc --noEmit` clean, `eslint packages/workflows/src/dag-executor.ts --max-warnings 0` clean (test files are excluded from ESLint per repo convention). GitHub CI (`test`, `docker-build`) was still running at verification time; `Build docs site`, `schema-upgrade`, and `CodeRabbit` had already passed.

**Resolved:** 4 (R1, R2, R4 fixed; R3 confirmed correctly declined)

### Findings

|---|---|---|---|
| `R1` | Suggestion | Comments describing `$LOOP_USER_INPUT` delivery for `loop_group` body nodes weren't fully reconciled after bash gained the env channel | FIXED |
| `R2` | Suggestion | `authoring-workflows.md`'s stdout-persistence/truncation section framed the mechanism and its secrets warning as bash-specific | FIXED |
| `R3` | Suggestion | The two functions' `subprocessEnv` construction blocks remain a hand-duplicated 14-key literal with no shared builder | DECLINED |
| `R4` | Suggestion | The new script "at-or-below-cap" test covered 32,767 bytes but not the exact 32,768-byte cap boundary | FIXED |

<details>
<summary><code>R1</code> — Loop-input delivery comments not reconciled across bash/script after convergence</summary>

**Impact:** No behavioral effect — purely a documentation-accuracy gap. A reader relying on any one of these comments in isolation, without reading the actual `applyLoopPrevToBodyNode` splice logic, would have misjudged which channel(s) deliver `$LOOP_USER_INPUT` to a bash `loop_group` body node.

**Evidence (pre-fix):**
- `packages/workflows/src/dag-executor.ts:3309-3312` — bash's `loopUserInput` param comment overstated delivery as "never spliced into source."
- `:3586-3588` — script's sibling comment said bash "delivers loop input via quoted splice," now incomplete since bash also delivers via env.
- `:4368-4369` and `:8062-8064` — described `bodyLoopUserInput`/its assignment as serving `script:` nodes specifically, though bash nodes read it too.

**Evidence (post-fix, commit `04da8f94`):** `dag-executor.ts:3309-3313`, `:3588-3592`, `:4376-4379`, `:8065-8073` — all four rewritten to name both channels: env for a braced `${LOOP_USER_INPUT}` read or a nested subprocess, and `applyLoopPrevToBodyNode`'s pre-existing splice for the unbraced literal (bash only). Cross-checked against ground truth at `dag-executor.ts:4958-4965`: `runtime === 'sh'` bodies call `sub(script, true)` with `skipUserInput` defaulted `false` (splice happens); script bodies call `sub(script, false, true)` (splice skipped). The splice regex at `:4819` (`/\$LOOP_USER_INPUT/g`) matches only the unbraced literal, never the braced form. All three independently-run reviewers confirmed the four comments are mutually consistent and match this logic exactly.

**Required outcome:** Update the four comments to state both delivery channels accurately. Met.

**Disposition:** FIXED. Verified independently by all three owning-scope reviewers against the actual `applyLoopPrevToBodyNode` branch logic (not just against the commit's own claim). The seam-analyzer's originally-flagged risk — a future #2115 splice-removal cleanup trusting the false single-channel claim and deleting bash's still-needed splice path — is closed: the comments now correctly warn that removing the splice would break bash's unbraced-literal delivery.

</details>

<details>
<summary><code>R2</code> — User-facing docs framed output persistence/truncation as bash-only</summary>

**Impact:** `authoring-workflows.md:1032` named persisted stdout truncation and its "never print secrets" warning as bash-specific, though script node stdout has always been persisted the same way (this PR's original scope only added the byte cap to script, not the persistence itself). A workflow author reading only this paragraph could have inferred script node output isn't captured/audited.

**Evidence:** `packages/docs-web/src/content/docs/guides/authoring-workflows.md:1032` now reads "Successful `bash`/`script` stdout is retained..." and "never print secrets or credentials from a bash or script node." Cross-checked against `dag-executor.ts:3836` (`executeScriptNode`) and `:3444` (`executeBashNode`) — both apply the identical `formatPersistedNodeOutput` 32 KiB cap, so the reworded claim is accurate.

**Required outcome:** Reword the paragraph to name both `bash` and `script`. Met.

**Disposition:** FIXED. Confirmed against the actual persistence call sites for both node kinds; no new inaccuracy introduced by the rewording.

</details>

<details>
<summary><code>R3</code> — <code>subprocessEnv</code> construction stays hand-duplicated between the two sibling functions</summary>

**Impact:** `executeBashNode` and `executeScriptNode` build a field-for-field identical 14-key `subprocessEnv` object, connected only by "mirrors X" comments — the literal site where the underlying #2725 bug shipped.

**Evidence:** `dag-executor.ts:3373-3404` (`executeBashNode`) vs. `:3665-3695` (`executeScriptNode`), still field-for-field identical after the correction — the correction's 29-line diff in `dag-executor.ts` touched only the four comment blocks covered by R1 and added zero lines to either `subprocessEnv` block. Exactly 2 call sites confirmed (`grep -n "subprocessEnv: NodeJS.ProcessEnv"` → lines 3373, 3665 only).

**Required outcome:** No action required for this PR — extraction would run ahead of the repository's rule-of-three convention (AGENTS.md "DRY + Rule of Three": extract only after a pattern appears at least three times) on a proportionate 2-issue bug-parity fix. Worth flagging for whoever next touches either function's env construction.

**Disposition:** DECLINED. The implementation report's own reasoning matches the original reviewer's independent conclusion. Re-verified: the correction commit never touched either duplicated block, so nothing about the decline's premise changed. Non-blocking, tracked as a note for a future touch point rather than a GitHub issue (too small/speculative to warrant one on its own).

</details>

<details>
<summary><code>R4</code> — New script truncation test skipped the exact byte-cap boundary</summary>

**Impact:** Negligible correctness risk — the shared helper's boundary behavior was already regression-tested via the bash path. The new "at-or-below-cap" test for `executeScriptNode` used 32,767 bytes (one under the cap) but not the exact 32,768-byte cap value that the bash precedent covers via a loop.

**Evidence:** `dag-executor.test.ts:6071-6127` (post-fix) now loops over `[['below-cap', 32_767], ['exact-cap', 32_768]] as const`, structurally identical to the bash precedent at `:5913-5966` (same loop shape, fresh `store`/`mockDeps`/unique `workflowRun` per iteration, same three assertions: `node_output` equals the repeated string, `node_output_truncated` and `node_output_original_bytes` both undefined).

**Required outcome:** Add the 32,768 case to the script companion test for symmetry with the bash precedent. Met.

**Disposition:** FIXED. Confirmed by direct diff comparison against the bash precedent test; not needlessly complex — a direct reuse of an already-accepted pattern. `bun test src/dag-executor.test.ts` passes both cases (601 pass / 0 fail overall).

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 confirmed fixed, R2 confirmed fixed, R4 confirmed fixed; no new findings |
| `seams` | R1 confirmed fixed, R3 confirmed declined; no new findings |
| `simplify` | R1 confirmed fixed, R4 confirmed fixed; no new findings |

#### Validation

|---|---|---|
| `bun test src/dag-executor.test.ts` (packages/workflows, review checkout at `04da8f94`) | PASS | 601 pass, 0 fail, 1697 `expect()` calls — run independently by all three reviewers plus the coordinator. |
| `bun x tsc --noEmit` (packages/workflows) | PASS | Clean, no errors. |
| `bun x eslint packages/workflows/src/dag-executor.ts --max-warnings 0` | PASS | 0 errors, 0 warnings (test file excluded from ESLint per repo convention — expected "file ignored" notice only). |
| `bun x prettier --check` (3 changed files) | PASS | "All matched files use Prettier code style!" |
| `git diff f06eadee..04da8f94 --stat` | PASS | Bounded correction diff confirmed to touch exactly the 3 claimed files (68 insertions, 54 deletions); single commit, confirmed ancestor of the reviewed head. |
| `grep -n "subprocessEnv: NodeJS.ProcessEnv"` | PASS | Confirms exactly 2 call sites remain (R3 decline premise unchanged by the correction). |
| GitHub CI (`gh pr checks 2729`) | PARTIAL | `Build docs site`, `schema-upgrade`, `CodeRabbit` passed; `test (ubuntu-latest)`, `test (windows-latest)`, `docker-build` were still pending at verification time — not a correction-caused failure, not required for this verification's scope. |

</details>

Wirasm:
Closing without merging. The #2725 fix (loop_group body bash node LOOP_USER_INPUT env delivery) is correct here but got bundled with #2726, and the #2726 half ships the naive truncation-parity fix that #2726's issue body explicitly warns against — capping executeScriptNode's persisted output the same way bash's is capped, with no spill-by-reference mechanism, which would hard-fail a resumed script node's output_format JSON parsing on any large output. Re-delivering #2725 fresh in a separate PR; #2726 is being delivered separately with the corrected spill-by-reference design.

Issue #2474, opened 2026-08-05
feat(workflows): reject an approval gate inside a fan-out child before spawning, not after


Wirasm:
An approval gate inside a workflow used as a **fan-out child** cannot work, and today that is discovered at runtime — after N children have been spawned, isolated, and paid for. It should be refused before the first child is created.

## The seam

A fan-out child is a full governance object and generalises cleanly in every respect: its own `workflow_runs` row (`executor.ts:729`), its own artifacts, cost, audit trail and abandon. **Pausing is the one operation that does not generalise from 1 to N**, because a gate is inherently a run-level rendezvous and fan-out has N runs where the rendezvous logic expects one.

With a **1:1** sub-run the semantics are coherent and already shipped: the child pauses, the parent waits, you approve the child by run id, the parent auto-resumes on completion. Exactly one thing to wait for, exactly one thing to approve — resuming the run resumes the whole tree.

At **N**, the same sentence stops meaning anything. Ten children; three paused at gates, four still running, three completed. What does "resume the run" resume? What does the parent wait for? There is no single approval and no coherent continuation point.

So the current behaviour falls out of that, and #2224's own documentation concedes it: a paused child **fails the fan-out node regardless of `join`**, described as *"the single case where a fan-out cancels a run it did not have to."* `interpret()` routes `case 'paused': → pauseParentOnChild(...)` (`dag-executor.ts:5482`), which is right for 1:1 and wrong for fan-out.

**This is not an argument against gates near fan-out.** A gate in the parent *before* or *after* the fan-out node is completely fine — that is a gate on a run which has exactly one of itself. The problem is only a gate reached *inside* a child.

## What's wrong with discovering it at runtime

The failure arrives after the expensive part. By the time a child hits its gate, the engine has resolved isolation, created N runs, and spent whatever those children spent before reaching the gate. The author then gets a node failure whose message describes a symptom rather than the mistake.

It is also the exact shape the constitution says to prefer against: a fail-fast that fires before cost, not after.

## Where the check can live — and where it cannot

**Not at load.** The gate lives in the *child* workflow, so a load-time reject means resolving the `workflow:` target at load. Sub-run targets resolve at **spawn** deliberately (#2200) — the constitution records that a load-time existence check *"compiles, passes every existing test, and silently destroys the capability"*, because spawn-time resolution is what lets a run author its own child mid-flight. Do not reach for a load-time check here.

[...]

PR #2730, opened 2026-08-22
feat(workflows): workflow class enforcement + one drive verb

### LINKED ISSUE #1991: [Bug/Feature Request] Workflow with 'interactive: true' hangs indefinitely when executed via background CLI tasks in Claude
## Description:
When executing an Archon workflow configured with interactive: true in claude code, the workflow hangs indefinitely. The background process logs the interactive prompt but has no channel to receive human input, stalling the entire run until manually terminated.
## Steps to Reproduce:

   1. Configure a workflow YAML with interactive: true globally and on an LLM/prompt node.
   2. Run the workflow using a background-capable agent or a background CLI task execution environment.
   3. Observe that the workflow halts at the interactive node waiting for input that cannot be provided.

## My Current Workflow Configuration:

name: my-workflowdescription: |
  Use when: User want to test my workflow, test simple workflow or command or just want to test the workflow system.
  Triggers: "use my-workflow", "run my-workflow", "execute my-workflow", "start my-workflow", "my-workflow", "test my-workflow", "test workflow", "test my workflow", "test the workflow", "test the my workflow", "test the my-workflow", "test the workflow system", "test the my workflow system", "test the my-workflow"
interactive: true
nodes:
  - id: list-files
    bash: |
      #!/bin/bash

      echo "Listing files in the current directory:"
      ls -la
  - id: brainstorm
    context: "shared"
    depends_on: ["list-files"]
    interactive: true
    prompt: |
      You are a senior software engineer.

      Using obra superpowers brainstorm skills (superpowers:brainstorming) to achieve below goal:
        1. Find the intent of the user: $INPUT.
        2. Brainstorming: Refine the design using Socratic questions to build a bite-sized specification.
        3. Git Worktree: Isolate the task into a separate development branch to ensure safety.
        4. Planning: Create a granular, step-by-step implementation plan paired with explicit verification steps.
        5. Sub-Agents: Spawn autonomous sub-agents to parallelize tasks and update progress tracking.
        6. TDD: Write a failing test first, then write the minimal code to pass.
        7. Code Review: Run a two-stage check for both specification compliance and code quality.
        8. Completion: Verify the full test suite.
  - id: verify-organization
    context: "shared"
    depends_on: ["brainstorm"]
    bash: |
      #!/bin/bash

      echo "Verifying the output of the brainstorm node..."
      yarn lint

## Question:
   3. Interactive Channel Integration: Provide a way for external orchestrators (like Claude Code) to securely pass input strings into active interactive nodes.

Wirasm:
## Problem and outcome

The workflow-level `interactive:` boolean was checked in exactly one place — web's background-dispatch branch. Every other background-dispatch entry point (CLI `workflow run --detach`, the `manage_run` native tool) launched an interactive-class workflow with no check, and if it later hit a pause node, the run paused with nobody watching and nothing to resume it (#1991). Separately, a gate's declared decisions beyond `approve`/`reject` had no way to reach an operator: the schema allowed authoring a third decision, but no drive surface could produce anything but the two hardcoded strings.

- **Outcome:** `interactive:` is now a load-bearing class declaration. An unattended workflow can never load with a pause node anywhere in its own DAG. An interactive-class workflow can never be dispatched to the background, on any surface — CLI `--detach`, `manage_run`, and web's default dispatch each independently refuse it. A paused gate's declared decisions are drivable from every surface through one verb: `workflow respond <run-id> <decision> [text]`.
- **Invariant:** No workflow whose class permits a pause is ever background-dispatched; no workflow whose class forbids a pause can load with a native pause node.
- **Scope boundary:** This is step 2 of 3 in #2707's design (step 1, gate output unification, already merged as #2721; step 3, loop collapse, is a separate later workstream). A gate that arrives via `include:` composition is **not** a load-time error — see Review guidance below, this is a deliberate correction from the design's first draft, not an oversight. `manage_run`'s native-tool `action` enum was not extended with `respond` — the deterministic `/workflow respond` slash command works on every platform today, including Claude/Pi chats, but the native tool itself still only knows `approve`/`reject`.

## Review guidance

- **Feedback requested:** Correctness of the composed-gate scoping decision below, and whether the interactive-class background-dispatch refusal is correctly scoped to fresh launches only (not continuations).
- **Start here:** `packages/workflows/src/utils/workflow-requirements.ts:132-176` — `assertInteractiveClassNotBackgrounded`, the actual #1991 fix, and the doc comment explaining why it's independently called at three separate entry points rather than funneled through one shared function.
- **Review order:** schema (`dag-node.ts`, `workflow.ts`) → `workflow-requirements.ts` (new class check + fan-out spawn-time check) → `loader.ts` (`validateWorkflowClassPlacement`) → `dag-executor.ts` (fan-out preflight) → wiring (`orchestrator.ts`, CLI `workflow.ts`) → `workflow-operations.ts` (`respondToWorkflow`) → CLI/chat/HTTP surfaces → web console (`ApprovalPanel.tsx`, `run.ts`).
- **Lower-attention areas:** `packages/web/src/lib/api.generated.d.ts` — mechanical regeneration against the live server + prettier; diff isolated to the one new route (verified via a before/after path-set diff, not just line count).
- **Known risk or uncertainty:** The original plan claimed the load-time class check would make a composed gate (via `include:`) a load error too, superseding the existing invocation-time `assertComposedGateDriveable`. Real test evidence proved this wrong — a reusable block can legitimately be composed into an interactive parent without declaring `interactive: true` on itself, because the same block may also be independently discoverable and load time can't know which workflow will own a given run. `assertComposedGateDriveable`'s own doc comment independently confirms a load-time version was tried before and reverted for exactly this reason. Kept `assertComposedGateDriveable` unmodified; the new load-time check only governs a workflow's own natively-authored pause nodes. Flagging this explicitly since it narrows one acceptance criterion from the published plan (composed gates stay invocation-time-only, same as before this PR).

## Solution

Three independent enforcement points now call the same pure `assertInteractiveClassNotBackgrounded(workflow)` check, mirroring the already-proven `requires: [github]` pattern (pure assertion, independently called per surface rather than threaded through one shared dispatch function): CLI `workflow run --detach` (before the fork), and `dispatchBackgroundWorkflowOwned` in `orchestrator.ts` (the shared entrypoint both web's default dispatch and `manage_run`'s `startWorkflow` funnel through). The check is scoped to a genuinely fresh launch — `workflow run --resume --detach` and the dedicated `resume`/`approve`/`reject`/`respond --detach` commands are exempt, since a run that already paused once has already proven it can pause; refusing its own continuation would break the documented `--detach` pattern for driving one gate and letting a background process continue to the next.

A new `validateWorkflowClassPlacement` function in `loader.ts` walks a workflow's own (pre-composition) node list for a native `approval:` gate or a `loop`/`loop_group` with node-level `interactive: true`, and rejects the load if the workflow doesn't declare `interactive: true` — the same shape as the existing `validateDagStructure` checks (duplicate id, cycle, dangling ref), just for class placement. This makes the fan-out spawn-time check in `dag-executor.ts` cheaper than issue #2474 originally scoped it (which anticipated walking the target's node tree): since an unattended workflow is now guaranteed gate-free by construction, "is the fan-out target unattended-class" is a one-field read on the already-resolved target definition, done unconditionally (not just when concurrency > 1, since the N-children-one-approval-slot problem exists regardless of `max_parallel`/`isolation`).

`respondToWorkflow(runId, decision, text?)` in `workflow-operations.ts` is the new general entry point: `decision === 'approve'`/`'reject'` delegate directly to the existing `approveWorkflow`/`rejectWorkflow` functions unchanged (byte-for-byte behavior preservation for every gate shape that existed before this PR), any other decision resolves through a new focused path that validates the id against the gate's declared `decisions` and writes the same structured `{decision, text}` output shape step 1 already built. `approvalDecisionConfigSchema.id` widens from `z.enum(['approve','reject'])` to an open string — its own doc comment named this PR as the unlock condition. CLI, chat, and a new `POST /api/workflows/runs/{runId}/respond` route all call the same function; the web console's `ApprovalPanel` now renders `run.approval.decisions` instead of two hardcoded buttons.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `interactive: true` only affected web background dispatch; CLI `--detach` and `manage_run` ignored it entirely | `interactive: true` is checked at every background-dispatch entry point; a native pause node in an unattended workflow fails to load |
| **Failure behavior** | An interactive-class workflow dispatched via `--detach`/`manage_run` paused silently — the run sat stuck with nothing to resume it (#1991) | The same dispatch fails synchronously before any fork/worktree/AI cost, naming the workflow and its class |

## Validation

- `bun run validate` — passed — full suite (bundled/schema/pi-vendor-map/capability-matrix checks, type-check, lint, format:check, test:install, `bun run test` across every package).
- Manual e2e against a live dev server: dispatched a workflow with a 3-decision gate (`approve`/`revise`/`escalate`) to `paused`; `GET` exposed the declared decisions; `POST /respond` with an undeclared decision returned 400 naming the actual options (`approve, revise, escalate`); `POST /respond` with `escalate` resolved, auto-resumed, and the run completed with `{decision:'escalate', text:'needs manager review'}` recorded in the audit trail.
- Manual e2e — #1991 repro: `workflow run <name> --detach` on the same `interactive: true` workflow now fails synchronously with `Error: Workflow '<name>' is interactive-class ('interactive: true') and may pause for human input — it cannot be dispatched in the background`, never forks, never hangs.
- Manual e2e — load-time class check: a scratch workflow with a bare `approval:` node and no `interactive: true` fails `workflow list` with the expected class-placement error naming the node and the fix.
- Pre-implementation audit: every workflow in this repo's `.archon/workflows/` containing a real `approval:`/interactive-loop node already declares `interactive: true` — zero migration risk from the new load-time rule against anything currently in this repo.
- **Not verified:** `manage_run`'s native-tool path wasn't exercised live for a declared-decision gate (only its shared `dispatchBackgroundWorkflowOwned` entrypoint, which the interactive-class check lives in, was verified via the CLI `--detach` repro above and the unit tests for `assertInteractiveClassNotBackgrounded`) — the native tool's own `action` enum still only knows `approve`/`reject`, flagged above as a follow-up rather than in this PR's scope.

## Delivery considerations

|---|---|---|
| Compatibility / migration | The new load-time rule is a hard error, not a phased warning — a third-party workflow with an undeclared native gate will fail to load after this ships. Deliberate, matching the issue's own "is a load error" wording; zero known workflows in this repo are affected. | Pre-implementation audit above; `packages/workflows/src/loader.ts` new check |
| Observability | Every new refusal (load-time class error, background-dispatch refusal, fan-out spawn refusal, undeclared-decision error) is a synchronous, user-facing error naming the workflow/node/class and the fix — no silent drops. | `InteractiveClassBackgroundError`, `assertRespondable` messages |
| Documentation / communication | `AGENTS.md` updated: the missing `approve`/`reject` HTTP routes plus the new `respond` route, the workflow-level `interactive:` class semantics, the composed-gate scoping, the fan-out spawn-time preflight, and the CLI `--detach`/`respond` rules. | `AGENTS.md` diff in this PR |

## Links

- Closes #1991
- Related #2707 (step 2 of 3 — step 1 merged as #2721, step 3 not started)
- Related #2474 (superseded by #2707, delivered here more cheaply than #2474 itself scoped)
- Plan: https://github.com/coleam00/Archon/issues/2707#issuecomment-5381679102

## Summary by CodeRabbit

* **New Features**
  * Resolve paused workflow gates with custom, declared decisions and optional text through the CLI, API, web console, and workflow management tools.
  * View available gate decisions and receive structured resolution results.
* **Behavior Changes**
  * Interactive workflows are prevented from background execution, and interactive fan-out children are rejected before spawning.
  * Workflows missing required interactive declarations load during a grace period with warnings.
* **Validation**
  * Decision identifiers require valid kebab-case formatting, with clearer errors for invalid or undeclared choices.


Wirasm:
## Ready to merge

This round reviews a third correction commit, `185a6ff8`, added on top of the already-`READY TO MERGE` head `71f4e759` after the operator raised a real pre-merge concern: the hard load error for an undeclared native pause node broke every pre-existing workflow with a gate and no `interactive: true`, with no transition. The commit softens that to a grace-period inference (`parseWorkflow` now coerces `interactive: true` and warns instead of rejecting the file — #2736/#2738) and removes the `WorkflowLoadError.name`/`failedNames` plumbing that the hard-error path made permanently unreachable. Four independently-dispatched reviewers (code, seams, simplify, and type-design — the four scopes selected for this PR, `types` retained because `WorkflowLoadError`'s shape changes) each traced the coercion to every downstream enforcement site, verified the dead-code removal left no orphaned half, and confirmed the new tests exercise real behavior through the actual discovery pipeline. All four returned zero blocking findings. One reviewer (type-design) surfaced a single non-blocking documentation suggestion, recorded below as `R10`. All nine findings from this PR's earlier review history (R1–R9) are unaffected by this round and remain closed.

**Validation:** PASS at the corrected head `185a6ff8`, in a detached worktree pinned to that exact commit — `bun install`, `bun run type-check` (all 13 packages, exit 0), `bun run lint` (repo-wide, 0 warnings), `bun x prettier --check` on all 6 changed files, `bun test src/loader.test.ts` (327 pass, 0 fail, including the 2 rewritten inference tests, the rewritten include-composition test, and the new log-dedup describe block), `bun --filter '@archon/workflows' test` (full package suite, all green), and `bun --filter '@archon/core' --filter '@archon/server' --filter '@archon/cli' --parallel test` (all green — these consume `WorkflowLoadError`/`expandWorkflowIncludes` transitively). See coverage table for the complete list.

**Resolved:** R1–R9 (carried forward, unaffected by this round) · **New:** R10 (Suggestion, non-blocking)

### Findings

|---|---|---|---|
| `R1`–`R9` | — | Full review history from earlier rounds | FIXED (carried forward) |
| `R10` | Suggestion | The `TODO(#2738)` comment on `validateWorkflowClassPlacement` doesn't mention that restoring the hard error will also need to resurrect the `WorkflowLoadError.name`/`failedNames` mechanism this commit just removed | OPEN |

<details>
<summary><code>R1</code>–<code>R9</code> — full prior review history (unaffected by this round)</summary>

This round's diff (`71f4e759..185a6ff8`) touches exactly 6 files: `AGENTS.md`, `packages/workflows/src/include-expander.ts`, `packages/workflows/src/loader.test.ts`, `packages/workflows/src/loader.ts`, `packages/workflows/src/schemas/workflow.ts`, `packages/workflows/src/workflow-discovery.ts`. None of R1–R9's originally-fixed code paths are contradicted or reopened by this diff — R1's specific mechanism (`WorkflowLoadError.name` + `failedNames`) is deliberately *removed* here because the load-time hard error it protected against no longer exists (see the new evidence in this round's coverage below, which independently re-confirmed the removal is clean and doesn't reintroduce R1's original "misleading not found" symptom, since that symptom required a hard-error return path that no longer exists). R2–R9 (manage_run validation, ApprovalPanel gate-identity reset, doc-comment accuracy, reject-text defaults, decision-id schema, Mode-type simplification, and their two review-round follow-up fixes) are entirely untouched by this diff. Full evidence for each remains in this report's git history (previous revisions) and the prior round's validation tables.

**Disposition:** FIXED (carried forward, unaffected by this round)

</details>

<details>
<summary><code>R10</code> — <code>TODO(#2738)</code> comment doesn't flag that it also un-deletes R1's <code>.name</code>/<code>failedNames</code> mechanism</summary>

**Impact:** Purely a maintenance-legibility gap, not a behavioral defect. `packages/workflows/src/loader.ts:1016-1019`'s `TODO(#2738)` comment on `validateWorkflowClassPlacement` tells a future implementer to "delete the coercion in `parseWorkflow` and restore the hard error" once the grace period ends, but doesn't mention that doing so resurrects the exact problem R1 fixed in this PR's own earlier review round: a leaf gate-authoring block that fails its own class check will again surface a misleading "not found" to every composer of it, and the `.name`/`failedNames` mechanism (removed by this very commit) will need to come back. Whoever picks up #2738 would have to re-discover this from the R1 finding history rather than from the code itself.

**Evidence:** `packages/workflows/src/loader.ts:1016-1019` (the `TODO(#2738)` comment); the removed mechanism is documented in this report's R1 entry and in `/Users/rasmus/.prp/archon-75601ef6/reports/2707-step2-classes-drive-verb-report.md`'s Deviation 6.

**Required outcome:** A one-line addition to the `TODO(#2738)` comment, e.g. "this also un-deletes the R1 `.name`/`failedNames` mechanism removed alongside this coercion — see that finding in the PR review history." Cheap, optional, and does not block merge — the current code is correct and complete without it; this only saves a future re-discovery.

**Found by:** `prp-core:type-design-analyzer`.

**Disposition:** OPEN (non-blocking Suggestion; left for the operator to accept or decline — not required by this PR's outcome or invariant)

</details>

<details>

#### Reviewer coverage

**This round** (bounded to `71f4e759..185a6ff8`, all four scopes from the prior round re-dispatched since this is fresh committed work, not a re-verification of a specific open finding):

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | Independently traced the coercion to every downstream enforcement site (`assertInteractiveClassNotBackgrounded`, `dispatchBackgroundWorkflowOwned`, CLI `--detach` refusal, fan-out spawn check) and confirmed each reads the freshly re-parsed, coerced value — no stale-value bypass. Confirmed the warn-once dedup claim against the actual code (`parseWarnings` unconditional, log gated by the Set). Confirmed `WorkflowLoadError.name` removal is clean tree-wide, including confirming the OpenAPI-facing schema never had a `name` field even before this PR. No findings. |
| `seams` (`prp-core:seam-analyzer`) | Confirmed the `interactive: true` conflation (declared vs. inferred) is not a missing type — no consumer needs to distinguish the two, and `WorkflowWithSource.declared` doesn't carry `interactive` at all so no dual-value collision is possible. Confirmed both halves of the R1 plumbing (writer + reader) are gone together, no orphan half. Confirmed the two warning channels (`parseWarnings`, log) are sourced from one string and can't independently drift. Identified one pre-existing (not introduced by this diff) bypass of the grace-period coercion in `test-utils.ts`'s fixture builders — flagged as informational only, no current test relies on it, not a finding against this commit. No findings. |
| `simplify` (`prp-core:code-simplifier`) | Verified the `Set<string>` + test-reset-function shape genuinely mirrors the existing `hasWarnedLegacyHomePath` precedent, not novel machinery. Verified bare-filename keying is consistent with the existing scope-shadowing identity convention workflow-discovery already uses. Verified the removal left no half-cleaned scaffolding. Verified no duplicated string construction between the two warning channels. No speculative generality found. No findings. |
| `types` (`prp-core:type-design-analyzer`) | Verified the `interactive` boolean collapse is correct type design (provenance lives on the separate `parseWarnings` channel, which no enforcement site needs). Verified `WorkflowLoadError.name` removal is safe — `classError` is now consumed by exactly one branch (warn + coerce), never a hard-error return, confirmed via full-tree grep of `validateWorkflowClassPlacement`'s call sites. Verified the coercion ternary's inferred type matches `WorkflowDefinition.interactive` exactly, no widening or cast. Verified the two warning channels can't disagree about whether a violation occurred (both driven by the same `if (classError)` branch). Raised R10 (see above) as the sole non-blocking observation. |

**Prior rounds** (`32e2c96d..632e9b0e`, `632e9b0e..71f4e759`): see this report's prior revision for the full R1–R9 ledger and per-round coverage tables.

#### Validation

|---|---|---|
| `git merge-base --is-ancestor 71f4e759... 185a6ff8...` | PASS | Single correction commit, ancestor confirmed |
| `git diff --numstat 71f4e759..185a6ff8` | PASS | 6 files, matches the described scope exactly: `AGENTS.md` (+2/-2), `include-expander.ts` (+2/-18), `loader.test.ts` (+90/-37), `loader.ts` (+59/-21), `schemas/workflow.ts` (+7/-15), `workflow-discovery.ts` (+1/-10) |
| `bun install` (worktree pinned to `185a6ff8`) | PASS | Clean, 2599 packages |
| `bun run type-check` (all 13 packages) | PASS | Every package exits 0 |
| `bun run lint` (repo-wide) | PASS | 0 warnings, `--max-warnings 0` gate |
| `bun x prettier --check` (all 6 changed files) | PASS | Prettier-compliant |
| `cd packages/workflows && bun test src/loader.test.ts` | PASS | 327 pass, 0 fail, 874 expect() calls |
| `bun --filter '@archon/workflows' test` | PASS | Full package suite, all files green, exit 0 |
| `bun --filter '@archon/core' --filter '@archon/server' --filter '@archon/cli' --parallel test` | PASS | All green — transitive consumers of `WorkflowLoadError`/`expandWorkflowIncludes`/`interactive` |
| Reviewer-independent re-runs of `type-check`, `lint`, and `src/loader.test.ts` (code, seams, simplify, types agents) | PASS | Each reviewer independently re-ran and confirmed clean, cross-checked against the coordinator's own runs |
| Prior-round checks not re-run (unaffected by this diff, results preserved) | — | `bun run format:check`, `bun run check:bundled`, full `bun run test` per package for `adapters`/`web`, and every test file not touched by this diff — all PASS at `71f4e759`, unchanged by this correction |

</details>

PR #2731, opened 2026-08-22
fix(core): /reset clears the execution binding and abandons resumable runs

### LINKED ISSUE #2291: bug(core): /reset leaves the worktree binding and resumable runs intact — the next message can resume a stale run

Wirasm:
## Problem and outcome

`/reset` advertised a fresh start but only deactivated the AI session — the conversation's execution binding (`cwd` + `isolation_env_id`) and every resumable run survived, so the next message could continue an old run on an old worktree instead of starting over. There was no in-band escape hatch from chat: an operator had to know about `/workflow abandon` and `/worktree remove` and run them by hand.

- **Outcome:** `/reset` is now a real escape hatch. It still deactivates the session; it additionally nulls `cwd` + `isolation_env_id` on the conversation and abandons every resumable run belonging to the conversation through the shared `abandonWorkflow` op. The reply names the count.
- **Invariant:** the next dispatch in a reset conversation must resolve to `codebase.default_cwd` (not the prior worktree) and must not find a resume candidate — i.e. `findResumableRunByParentConversation` and `getPausedWorkflowRun` both return nothing for the conversation after `/reset`.
- **Scope boundary:** `codebase_id` is preserved (detaching the project is `/setproject none`'s job), and `pending` / `running` rows are deliberately not cancelled (they may belong to a different process — CLI run, webhook, scheduled dispatch — and the resume lookups this exists to defeat never match them anyway).
- **Root cause:** `case 'reset'` in `packages/core/src/handlers/command-handler.ts` only called `safeDeactivateSession` and never touched the conversation's binding columns or the workflow-runs table; meanwhile the two resume lookups (`getPausedWorkflowRun`, `findResumableRunByParentConversation`) match on `status IN ('paused', 'failed')` against rows that the conversation kept ownership of, so a silent stale continuation was reachable by construction.

## Review guidance

- **Feedback requested:** correctness of the new query's predicate and the abandon pipeline; whether the partial-failure reply string reads clearly to a user; whether the disclosure-to-user contract for "N runs abandoned" is the right shape.
- **Start here:** `packages/core/src/handlers/command-handler.ts:1329` — the new `case 'reset'` is the load-bearing change; everything else serves it.
- **Review order:**
  1. `packages/core/src/db/workflows.ts:499` — `findResumableRunIdsForConversation` and its doc comment (the predicate is the boundary).
  2. `packages/core/src/operations/workflow-operations.ts:456` — `abandonResumableRunsForConversation` (routes through the existing `abandonWorkflow` rather than a bulk UPDATE).
  3. `packages/core/src/handlers/command-handler.ts:1329` — the handler change: two try blocks, piecewise reply, `success` gated on `bindingCleared && abandonError === null`.
  4. `packages/docs-web/src/content/docs/reference/commands.md:51` — the user-visible description update.
  5. `CHANGELOG.md` — entry under `[Unreleased] / Fixed`.
- **Lower-attention areas:** the three added test files (`workflows.test.ts`, `workflow-operations.test.ts`, `command-handler.test.ts`) — each asserts a load-bearing invariant (predicate excludes `pending`/`running`; partial-failure paths surface the count; `codebase_id` is preserved).
- **Known risk or uncertainty:** none material. The `pending` / `running` carve-out is the one design decision a reviewer should sanity-check against `AGENTS.md`'s "No Autonomous Lifecycle Mutation Across Process Boundaries" — the carve-out exists *because* of that rule.

## Solution

Two new primitives, one handler change, no new abstractions.

1. **A typed lookup of resumable ids for a conversation** (`packages/core/src/db/workflows.ts`). `findResumableRunIdsForConversation(convId)` selects `id` from `remote_agent_workflow_runs` where `(conversation_id = $1 OR parent_conversation_id = $2)` and `status IN ('paused', 'failed')`, ordered by `started_at DESC`. The doc comment explicitly scopes the predicate to "what the two resume lookups can return" and names why `pending` / `running` are excluded.
2. **A typed abandon-everything loop** (`packages/core/src/operations/workflow-operations.ts`). `abandonResumableRunsForConversation(convId)` calls the lookup, then loops `abandonWorkflow` per id under per-run try/catch, returning `{ abandoned, failures, cascadeFailures, blockedParentRunId }`. Routing through the shared op (instead of a bulk UPDATE) inherits its CAS guard, sub-run cascade (`cascadeCancelChildren`), and immediate container + upper-volume reclaim — so this surface stays a sibling of CLI, web, chat-`manage_run`, and Slack-cancel, not a sixth abandon surface that quietly skips all three.
3. **Handler rewrite** (`packages/core/src/handlers/command-handler.ts`). `case 'reset'`:
   - keeps the session-deactivation branch,
   - runs the abandon in its own try block, capturing `abandonError` separately,
   - runs `db.updateConversation({ cwd: null, isolation_env_id: null })` in a separate try block, capturing `bindingCleared` and `bindingError`,
   - builds the reply piecewise so a binding-clear failure after a successful abandon still surfaces the count,
   - preserves `codebase_id` (the resulting row is byte-for-byte what `/setproject` already writes),
   - gates `success` on `bindingCleared && abandonError === null`.

This is the smallest coherent fix because the two effects are independent (an abandon failure must not block the binding clear and vice-versa) and because both effects already have the right primitives upstream — the work is wiring, not design.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `/reset` deactivates the AI session, replies "Session cleared. Starting fresh on next message. Codebase configuration preserved.", and the next message can still resume a stale `paused`/`failed` run on the prior worktree. | `/reset` deactivates the AI session, nulls `cwd` + `isolation_env_id`, abandons every resumable run for the conversation through `abandonWorkflow`, and replies with the abandoned count and a confirmation that the project attachment is preserved. The next message resolves to `codebase.default_cwd` against a fresh isolation env and finds no resume candidate. |
| **Failure behavior** | A single `safeDeactivateSession` failure would short-circuit the entire command with no recovery. | Abandon and binding-clear fail independently and are reported independently. A binding-clear failure after a successful abandon still surfaces the abandoned count; an abandon failure with a clean binding clear reports both the cleared binding and a "could not look up resumable runs" warning. `success` is `false` if either effect failed. |

### User flow

```mermaid
flowchart LR
  subgraph Before
    U1["User: /reset"] --> B1["Deactivate AI session"]
    B1 --> B2["Reply: 'Session cleared'"]
    B2 --> B3["Next message resumes stale run on prior worktree"]
  end

  subgraph After
    U2["User: /reset"] --> A1["Deactivate AI session"]
    A1 --> A2["Abandon every paused/failed run (via abandonWorkflow)"]
    A2 --> A3["Null cwd + isolation_env_id"]
    A3 --> A4["Reply: 'Abandoned N resumable run(s). Project attachment preserved — next message starts fresh.'"]
    A4 --> A5["Next message resolves to codebase.default_cwd, fresh isolation env, no resume candidate"]
  end
```

## Architecture

```mermaid
flowchart LR
  H["/reset handler (command-handler.ts:1329)"] --> L["findResumableRunIdsForConversation (db/workflows.ts:499)"]
  H ==> O["abandonResumableRunsForConversation (operations/workflow-operations.ts:456)"]
  O --> A["abandonWorkflow (existing primitive, operations/workflow-operations.ts:348)"]
  A ==> DB[("remote_agent_workflow_runs: status -> 'cancelled', CAS + cascadeCancelChildren + container reclaim")]
  H --> U["db.updateConversation({cwd: null, isolation_env_id: null})"]
```

### Changed seams

|---|---|---|
| `/reset` chat command → user | Reply now names the abandoned count and the binding-cleared confirmation; `codebase_id` is preserved. | `packages/core/src/handlers/command-handler.ts:1329`, new test cases in `command-handler.test.ts` |
| `case 'reset'` → conversation row | Now nulls `cwd` and `isolation_env_id` in a dedicated try block; `codebase_id` is untouched. | `packages/core/src/handlers/command-handler.ts:1370`; test asserts `codebase_id` survives. |
| `case 'reset'` → workflow-runs table | Now abandons every resumable run through `abandonWorkflow`; bulk UPDATE explicitly avoided. | `packages/core/src/operations/workflow-operations.ts:456`; test asserts routing through the shared op. |
| `/reset` docs surface | Description row updated; example reply updated; `pending`/`running` carve-out documented. | `packages/docs-web/src/content/docs/reference/commands.md:51` |

## Validation

- `bun run check:bundled` — OK (37 commands, 21 workflows, 0 scripts).
- `bun run check:bundled-skill` — OK (23 files across 2 skills).
- `bun run check:bundled-schema` — OK.
- `bun run check:pi-vendor-map` — OK.
- `bun run check:capability-matrix` — OK.
- `bun run type-check` — all 8 packages exit 0.
- `bun run lint --max-warnings 0` — OK.
- `bun run format:check` — all matched files use Prettier code style.
- `bun test packages/core/src/db/workflows.test.ts` — 98 pass / 0 fail / 211 expects. Includes the load-bearing test asserting `'pending'` and `'running'` are NOT matched by `findResumableRunIdsForConversation`.
- `bun test packages/core/src/operations/workflow-operations.test.ts` — 66 pass / 0 fail / 164 expects. Includes tests for the zero-match path, routing through `abandonWorkflow`, the mid-loop terminal-row race, and the lookup-failure propagation.
- `bun test packages/core/src/handlers/command-handler.test.ts` — 139 pass / 0 fail / 394 expects. Includes tests for: (a) `cwd` + `isolation_env_id` cleared, `codebase_id` preserved; (b) abandons routed through `abandonWorkflow` and reported as `Abandoned N resumable run(s).`; (c) binding-clear failure after a successful abandon still surfaces the count.
- **Not verified:** the full `bun run test` suite — the combined run reports pre-existing failures (e.g. `OpencodeProvider`, `SSETransport`, the adapter `start()` retry tests) on `dev` without these changes; each affected file passes in isolation. Out of scope per `AGENTS.md`: `bun run test:install` (needs a built installer binary and docker) and `bun run check:schema-upgrades` (needs a reachable PostgreSQL, CI-only).

## Delivery considerations

|---|---|---|
| Compatibility / migration | No data migration. Existing `cwd` / `isolation_env_id` values are nulled in place; `codebase_id` is preserved. Next dispatch resolves to `codebase.default_cwd` per existing precedence. | `db.updateConversation` is the existing primitive; the resulting row matches `/setproject`'s byte-for-byte. |
| Security / permissions / data | No new exposure. `abandonWorkflow` is the same op used by every other abandon surface; no new SQL or surface added. | `operations/workflow-operations.ts:348` exports `abandonWorkflow`; this PR only adds a caller. |
| Rollout / rollback | Standard `/release` workflow applies (CHANGELOG entry under `[Unreleased] / Fixed`). Rollback is a git revert; no schema change means no data migration to undo. | `CHANGELOG.md` entry added. |
| Observability | `cmd.reset_abandon_failed`, `cmd.reset_clear_binding_failed`, `operations.workflow_abandon_for_conversation_failed`, `operations.workflow_abandon_for_conversation_completed`, and `db.workflow_run_find_resumable_for_conv_failed` log lines added with structured `conversationId` and `runId` fields. | `command-handler.ts:1363,1373`, `workflow-operations.ts:498,506`, `workflows.ts:533`. |
| Documentation / communication | `/reset` row in `packages/docs-web/src/content/docs/reference/commands.md` updated; example reply updated to match the new message shape; CHANGELOG entry under `[Unreleased] / Fixed`. | `packages/docs-web/src/content/docs/reference/commands.md:51,114`, `CHANGELOG.md:32`. |

## Links

- Closes #2291

## Summary by CodeRabbit

* **New Features**
  * Enhanced `/reset` to clear session and workspace bindings while preserving the selected project.
  * Paused and failed resumable runs are abandoned atomically; pending and running runs remain unchanged.
  * Cleanup operations are isolated, and cancelled run containers are reclaimed automatically.
  * Reset reports incomplete operations and blocked parent runs, with retry guidance when needed.

* **Documentation**
  * Updated command documentation and examples to explain expanded `/reset` behavior, preserved project attachments, and retry guidance.


Wirasm:
# Review Report — PR #2731 (Closes #2291)

## Verdict

**Ready** — the two Important findings from the prior round (R1, R2) are fixed at `524df929`, both enabled lenses (code, seams) verify the fix, the targeted test suites pass (workflow-operations 66/66, command-handler 139/139), and no new findings were raised.

## Reviewed head SHA

`524df929b139dc9ed30f7e8376f065137a59b9ab`

## Findings

None. No new Critical or Important or Suggestion findings from the two enabled lenses.

## Prior findings

| ID | Severity | Title | Verdict |
|---|---|---|---|
| R1 | Important | `AbandonConversationRunsResult` discards `cascadeFailures` and `blockedParentRunId`; `/reset` reply can claim "everything starts fresh" while part of the run tree is still alive | **fixed at `524df929`** — verified by code lens and seams lens; see evidence below |
| R2 | Important | Failure count is inflated when the abandon loop visits a child the parent's cascade already cancelled | **fixed at `524df929`** — verified by code lens and seams lens; see evidence below |

### R1 — carry-forward evidence

- `packages/core/src/operations/workflow-operations.ts:406-425` — `AbandonConversationRunsResult` now declares `cascadeFailures: number` and `blockedParentRunId: string | null`. Doc comment names the aggregation contract ("Sum of `cascadeFailures` returned by every per-run `abandonWorkflow` call…", "First abandoned run that left a parent paused blocked-on-child…").
- `packages/core/src/operations/workflow-operations.ts:454-457` — aggregation variables `let cascadeFailures = 0; let blockedParentRunId: string | null = null;` mirror `abandoned`/`failures`.
- `packages/core/src/operations/workflow-operations.ts:490-495` — every per-run `abandonWorkflow` return is destructured and aggregated (`{ cascadeFailures: runCascade, blockedParentRunId: blocked }`; `cascadeFailures += runCascade;`; `if (blockedParentRunId === null) blockedParentRunId = blocked;`). The destructure inner name shadows the accumulator, so the inner value cannot be confused with it.
- `packages/core/src/operations/workflow-operations.ts:505-507` — log line includes both new fields.
- `packages/core/src/operations/workflow-operations.ts:509` — return shape carries both new fields.
- `packages/core/src/handlers/command-handler.ts:1356-1360` — destructures all four result fields.
- `packages/core/src/handlers/command-handler.ts:1394-1403` — two new `parts.push` warning branches; cascade string is verbatim against `case 'abandon'` at `packages/core/src/handlers/command-handler.ts:935`; blocked-parent string differs only by "an abandoned sub-run" vs "this sub-run" — deliberate disambiguation because `/reset` abandons N runs and has no single specific sub-run to point at.
- `packages/core/src/handlers/command-handler.ts:1410` — `success: bindingCleared && abandonError === null` gate preserved (regression check).
- New tests assert the user-visible behavior: `packages/core/src/handlers/command-handler.test.ts:842-874` (cascade-truncation) and `:876-918` (blocked-parent).

### R2 — carry-forward evidence

- `packages/core/src/operations/workflow-operations.ts:466-489` — precheck fetches the row's current status via `getWorkflowRun`, catches and counts a DB throw as `failures++` (`:471-480`), silently `continue`s on a missing row (`:482-486`), and `continue`s on `'completed'`/`'cancelled'` (`:486-488`).
- Precheck predicate is identical to `cascadeCancelChildren` at `packages/core/src/operations/workflow-operations.ts:121`.
- New test `silently skips a row that a concurrent cascade already cancelled (#2731 R2)` at `packages/core/src/operations/workflow-operations.test.ts:1314-1348` returns `['parent', 'child']` from `mockFindResumableRunIdsForConversation` and configures `mockGetWorkflowRun` so `child` returns `status: 'cancelled'` from the precheck; asserts `{ abandoned: 1, failures: 0, ... }` and that `mockCancelWorkflowRun.mock.calls.map(c => c[0])` is exactly `['parent']` (proves `abandonWorkflow` was never called for the child).
- Mirror at the handler level: `packages/core/src/handlers/command-handler.test.ts:920-952` asserts the message contains `Abandoned 1 resumable run(s).` and does NOT contain `run(s) could not be abandoned`.
- Genuine abandon failure (DB throw) coverage preserved by the new test at `packages/core/src/operations/workflow-operations.test.ts:1380-1409` — `mockGetWorkflowRun('run-bad')` first returns `paused` (precheck passes) then rejects with `'connection reset'`; asserts `{ abandoned: 1, failures: 1, ... }` and that `mockCancelWorkflowRun.mock.calls.map(c => c[0])` is exactly `['run-ok']`.
- Original "counts a run that went terminal mid-loop" test renamed to `still counts a row whose concurrent transition is NOT ours as a failure` at `:1350-1378`; the body was updated to reflect the new design (any precheck-detected terminal row is silently skipped, regardless of cause). Documented in the test's own comment at `:1351-1358`. The test continues to assert loop-not-aborting (`mockCancelWorkflowRun.mock.calls.map(c => c[0])` is exactly `['run-ok']`).

## Lens roster

| Lens | State | Notes |
|---|---|---|
| `code` | ran | Verified both prior Important findings fixed; no new findings. |
| `seams` | ran | Verified both prior Important findings fixed; no new findings. |
| `tests` | disabled by config | `tests=false`. Not a gap. |
| `errors` | disabled by config | `errors=false`. Not a gap. |
| `comments` | disabled by config | `comments=false`. Not a gap. |
| `types` | disabled by config | `types=false`. Not a gap. |
| `docs` | disabled by config | `docs=false`. Not a gap. |
| `simplify` | disabled by config | `simplify=false`. Not a gap. |

No enabled lens failed to report. All required evidence for the verdict is present.

## Verdict rule check

- R1: Important — fixed at `524df929`. No longer blocks.
- R2: Important — fixed at `524df929`. No longer blocks.
- Every enabled lens delivered its report.

Therefore `ready: true`.

## Test verification

- `bun test packages/core/src/operations/workflow-operations.test.ts` — **66/66 pass**, 0 fail, 164 `expect()` calls.
- `bun test packages/core/src/handlers/command-handler.test.ts` — **139/139 pass**, 0 fail, 394 `expect()` calls.

## Published comment

- URL: https://github.com/coleam00/Archon/pull/2731#issuecomment-5382011302
- Marker: `<!-- archon-review-report -->` on the first line; comment edited in place (not appended) and read-back will be confirmed below after the edit.

Wirasm:
## Ready to merge

The atomic conversation-wide cancellation fixes the last blocking reset bug. Selection and mutation now happen once inside one transaction, so `A (paused) -> B (running) -> C (paused)` cancels A and C exactly once without touching B or retaining diagnostics from an overlapping traversal. This correction verification covers `43fb5ab47fcde8075e0b2ebd185137fa971a9f24..d91b5d3129e3838b979b1c5731bc5727ac008f56`; all four prior Important findings are fixed. Two optional cleanup items remain: one user-facing error prefix is too narrow, and a cancellation-policy parameter no longer has multiple policies.

**Validation:** The authoritative validation suite, full build, four focused core test files, and both whitespace checks passed at the reviewed head.

**Resolved:** 4

### Findings

|---|---|---|---|
| `R1` | Important | `/reset` can cancel a run that another process resumed after the resumable lookup. | FIXED |
| `R2` | Important | Per-run abandon failures still produce a successful “next message starts fresh” result. | FIXED |
| `R3` | Important | A session lookup or deactivation error aborts reset before either new cleanup effect runs. | FIXED |
| `R4` | Important | Overlapping roots can retain transient child-run diagnostics after final cleanup succeeds. | FIXED |
| `R5` | Suggestion | An atomic cancellation failure is reported as a resumable-run lookup failure. | OPEN |
| `R6` | Suggestion | The explicit-abandon cascade still threads a cancellation policy that has only one remaining value. | OPEN |

<details>
<summary><code>R1</code> — Preserve resumable-only eligibility through the cancellation CAS</summary>

**Impact:** A concurrently resumed run must remain active; reset must not cancel work after ownership transfers to another lifecycle.

**Evidence:** `packages/core/src/db/workflows.ts:530` locks the conversation-scoped snapshot and `packages/core/src/db/workflows.ts:542` updates only rows that are still `paused` or `failed` in the same transaction. The competing resume path locks its row and performs its guarded update in one transaction at `packages/core/src/db/workflows.ts:868`.

**Required outcome:** No further correction is required.

**Found by:** `code-reviewer`, `seam-analyzer`, `code-simplifier`

**Disposition:** FIXED — resume and reset now make one serialized ownership decision: whichever mutation wins determines the state observed by the other.

</details>

<details>
<summary><code>R2</code> — Do not claim a fresh start when a resumable row remains</summary>

**Impact:** Reset must not return success or promise a fresh start while an eligible run remains uncancelled.

**Evidence:** `packages/core/src/db/workflows.ts:549` rejects and rolls back a snapshot/update count mismatch. The operation propagates cancellation errors from `packages/core/src/operations/workflow-operations.ts:452`, and `packages/core/src/handlers/command-handler.ts:1404` includes that error in the final success gate.

**Required outcome:** No further correction is required.

**Found by:** `code-reviewer`, `seam-analyzer`, `code-simplifier`

**Disposition:** FIXED — cancellation is all-or-nothing, and every cancellation error suppresses the fresh-start result.

</details>

<details>
<summary><code>R3</code> — Isolate session deactivation from the new reset effects</summary>

**Impact:** A session lookup or deactivation error must not prevent run abandonment and binding cleanup from running.

**Evidence:** `packages/core/src/handlers/command-handler.ts:1342`, `packages/core/src/handlers/command-handler.ts:1362`, and `packages/core/src/handlers/command-handler.ts:1373` keep session, run, and binding cleanup in independent guarded effects. Their results meet only at the final success decision at `packages/core/src/handlers/command-handler.ts:1404`.

**Required outcome:** No further correction is required.

**Found by:** `code-reviewer`, `seam-analyzer`, `code-simplifier`

**Disposition:** FIXED — one cleanup failure no longer prevents either remaining cleanup effect, and any failure prevents a false all-clear.

</details>

<details>
<summary><code>R4</code> — Derive diagnostics from the final run-tree state</summary>

**Impact:** The reset path must not visit the same selected descendant through two roots and retain a transient first-visit failure after a later successful cancellation.

**Evidence:** `packages/core/src/db/workflows.ts:530` now owns the complete conversation-scoped selection and mutation as one transaction. `packages/core/src/operations/workflow-operations.ts:452` consumes that single returned set without building roots or traversing descendants. The real SQLite case at `packages/core/src/db/workflows.resume-cas.integration.test.ts:198` constructs `A (paused) -> B (running) -> C (paused)` and proves A and C are cancelled once while B stays running.

**Required outcome:** No further correction is required.

**Found by:** `code-reviewer`, `seam-analyzer`, `code-simplifier`

**Disposition:** FIXED — the first-C-failure/second-C-success state is no longer representable because reset has no per-root queue or cascade.

</details>

<details>
<summary><code>R5</code> — Name cancellation failures as cancellation failures</summary>

**Impact:** When the atomic update or transaction fails, `/reset` reports “Could not look up resumable runs” before an error that says cancellation failed. The reset still fails safely, but the phase named to the operator is wrong.

**Evidence:** `packages/core/src/db/workflows.ts:559` now wraps both selection and mutation failures as cancellation failures. `packages/core/src/handlers/command-handler.ts:1401` still prefixes every propagated error with “Could not look up resumable runs”.

**Required outcome:** Optionally change the prefix to “Could not abandon resumable runs” so it covers both lookup and mutation failures.

**Found by:** `code-reviewer`

**Disposition:** OPEN — useful user-facing precision, but the command returns `success: false`, asks for a retry, and does not claim a fresh start, so this does not block the delivered invariant.

</details>

<details>
<summary><code>R6</code> — Remove the single-value cancellation-policy seam</summary>

**Impact:** `CancelWorkflowRun` and two parameters preserve a policy variation that no supported caller uses, leaving extra indirection and a stale comment about reset policy on the explicit-abandon path.

**Evidence:** `packages/core/src/operations/workflow-operations.ts:90` defines the policy alias and `packages/core/src/operations/workflow-operations.ts:101` plus `packages/core/src/operations/workflow-operations.ts:367` thread it through the cascade. The only remaining caller passes `workflowDb.cancelWorkflowRun` at `packages/core/src/operations/workflow-operations.ts:416`; reset uses the separate atomic primitive at `packages/core/src/operations/workflow-operations.ts:452`.

**Required outcome:** Optionally call `workflowDb.cancelWorkflowRun` directly from the explicit-abandon helpers and remove the unused policy parameters.

**Found by:** `code-simplifier`

**Disposition:** OPEN — the seam is now redundant, but it does not change behavior, ownership, or the reset invariant and is small enough to remain non-blocking.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | `R1`–`R4` fixed; `R5` open |
| seams | `R1`–`R4` fixed; no additional findings |
| simplify | `R1`–`R4` fixed; `R6` open |

#### Validation

|---|---|---|
| `bun install --frozen-lockfile` | PASS | Installed 2,599 locked packages in the fresh detached review worktree. |
| `bun run validate` | PASS | Bundled checks, type-check, lint with zero warnings, format check, installer tests, and the full isolated repository test suite passed. |
| `bun run build` | PASS | All package builds completed; web and docs production builds succeeded. |
| `bun test src/db/workflows.test.ts` | PASS | 98 passed, 0 failed. |
| `bun test src/db/workflows.resume-cas.integration.test.ts` | PASS | 29 passed, 0 failed, including the exact `A -> B -> C` status-gap topology. |
| `bun test src/operations/workflow-operations.test.ts` | PASS | 64 passed, 0 failed. |
| `bun test src/handlers/command-handler.test.ts` | PASS | 141 passed, 0 failed. |
| `git diff --check origin/dev...HEAD` | PASS | No whitespace errors in the complete current-base PR diff. |
| `git diff --check 43fb5ab47fcde8075e0b2ebd185137fa971a9f24..HEAD` | PASS | No whitespace errors in the correction diff. |

</details>

PR #2732, opened 2026-08-22
fix(workflows): script node output is now capped and resume-safe for both runtimes

### LINKED ISSUE #2726: workflows: script: node persisted output is never truncated, unlike bash: nodes

Wirasm:
## Problem and outcome

A bash node's stdout is already capped at 32KB before being written into the persisted `node_completed` event, but that cap has a latent bug: `getDagResumeSnapshot` rebuilds a resumed run's `$node.output` directly from that capped DB text, while a fresh run's in-process consumer always sees the full value. A script node had no cap at all, so its rows grew unbounded — the originally reported half of this issue.

An earlier version of this fix (apply bash's cap to script uniformly) was implemented and reverted before merge once the resume-dataflow root cause above was found: script nodes never persist a separate `structured_output`, so `.field` access always re-parses the raw persisted text — naive truncation would turn "resumed consumer sees clipped text" into "resumed consumer hard-fails on invalid JSON," specifically on the node type most likely to emit large structured JSON.

- **Outcome:** A resumed run's `$node.output`/`$node.output.field` for a `bash:` or `script:` node now resolves identically to what a fresh (non-resumed) run would have produced, for any output size. Persisted `node_completed`/`node_skipped_prior_success` rows stay bounded at 32KB for both node types.
- **Invariant:** The persisted DB row for an exec node's output never exceeds the bounded preview, and a resumed run's `$node.output` access is behaviorally indistinguishable from a fresh run's, regardless of output size.
- **Scope boundary:** No DB schema change, no `IWorkflowStore` interface change, no change to the 32KB cap value, no change to `structured_output`/AI-node output handling. Does not touch #2725 (`$LOOP_USER_INPUT` delivery to `bash:` loop_group bodies) — a separate, concurrently developed fix in the same file.
- **Root cause:** `getDagResumeSnapshot` (`packages/core/src/db/workflow-events.ts`) is the sole seed point for a resumed run's node-output map — it reads `data.node_output` verbatim from the DB row with no fresh-vs-resumed branch anywhere downstream. `executeBashNode` capped that row but returned the full value to its own in-process caller; `executeScriptNode` never capped it at all.

## Review guidance

- **Feedback requested:** Correctness of the spill/read-back mechanism, and whether bounding the `node_skipped_prior_success` re-emit for every node kind (not just exec) is the right call.
- **Start here:** `packages/workflows/src/dag-executor.ts:3256-3313` (`formatPersistedNodeOutput`) — the extended truncation-plus-spill helper every other change builds on.
- **Review order:** `dag-executor.ts`'s `formatPersistedNodeOutput` → its three call sites (`executeBashNode`, `executeScriptNode`, the `node_skipped_prior_success` re-emit) → `workflow-events.ts`'s `getDagResumeSnapshot` spill-preferring read → the new tests in `dag-executor.test.ts` and `workflow-events.test.ts`.
- **Lower-attention areas:** `packages/core/package.json`'s one-line test-script split (mechanical — required for the new fs-mocking test to not collide with an existing `mock.module('fs/promises', ...)` in the same `bun test` invocation).
- **Known risk or uncertainty:** None identified — see Validation for the fresh-vs-resumed regression proof.

## Solution

Keeps the existing 32KB bounded preview for the persisted DB row. When output exceeds the cap, `formatPersistedNodeOutput` (renamed from `formatPersistedBashOutput`) also spills the full bytes to `<artifactsDir>/.archon/node-output-spills/persisted/<step_name>.nodeoutput` and records the absolute path as `node_output_spill_path` in the event. `getDagResumeSnapshot` reads that spill back and uses it as the node's `output` whenever the field is present, falling back to the bounded preview on any read failure (never throwing — a missing spill degrades gracefully, matching this function's existing error-handling posture).

Applied at all three sites that persist an exec node's output text:
- `executeBashNode` and `executeScriptNode` — the two node-completion writers.
- The `node_skipped_prior_success` re-emit, which copies a resumed node's prior output forward into a new event. Once `getDagResumeSnapshot` returns the full spilled text, this site would otherwise write that full text unbounded into a new row on a second resume — the same helper is reused here, regardless of node kind, since the mechanism doesn't depend on it and this also closes a pre-existing latent gap for AI/prompt/loop skip-reemits at no extra cost.

No DB schema or interface change: `remote_agent_workflow_events.data` is already a free-form JSON blob carrying `node_output_truncated`/`node_output_original_bytes`; `node_output_spill_path` is one more additive key. The spill path is stored absolute (mirroring how `output_root` and the existing substitution-spill mechanism already store absolute paths), so `getDagResumeSnapshot` needs no new context threaded through `IWorkflowStore` — it just opens the path.

The write side reuses a small extracted helper (`writeSpillFile`) shared with the pre-existing, differently-purposed substitution spill (`shellQuoteOrFile`, used for oversized `$node.output` refs spliced into bash source) — same mkdir/write/error-handling shape, distinct directory so the two never collide.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Bash: resumed `$node.output` for an over-cap node silently returns clipped text. Script: persisted rows unbounded. | Both: resumed `$node.output` matches a fresh run's value exactly, for any size. |
| **Failure behavior** | Script `.field` access on a large JSON output was never protected; a resumed bash `.field` access on clipped JSON failed with a confusing "not a JSON object" error. | A resumed `.field` access on large JSON resolves correctly. If the spill file is ever missing/unreadable, resume degrades to the bounded preview with a warning log rather than failing.  |

## Validation

- `cd packages/workflows && bun test src/dag-executor.test.ts` — 602 pass, 0 fail — proves bounded+spilled persistence for both node types, the skip-reemit bounding, and the resumed large-JSON `.field`-access regression that would have failed both pre-fix and against the reverted naive-cap approach.
- `cd packages/core && bun run test` — full package script, 0 fail across every invocation — proves `getDagResumeSnapshot`'s spill-preference and its graceful fallback on a missing spill.
- `bun run validate` — passed end-to-end (bundled/schema/capability-matrix checks, full monorepo type-check, `eslint --max-warnings 0`, `prettier --check`, and the complete per-package test suite).
- **Not verified:** No live manual run against a real workflow — covered instead by the executor-level regression test that exercises the real subprocess + real filesystem spill (bash/bun scripts genuinely run and write real spill files under a temp `artifactsDir` in the test).

## Links

- Closes #2726
- Related #2725 (concurrently developed in the same file; not part of this PR)

Plan: https://github.com/coleam00/Archon/issues/2726#issuecomment-5381873727

## Summary by CodeRabbit

* **New Features**
  * Large workflow outputs are now stored in artifact files while displaying bounded, valid UTF-8 previews.
  * Resumed workflows can restore complete spilled output for downstream processing.
  * Output handling is consistent across scripts, resumed steps, cache invalidation, always-run steps, and loop iterations.
  * Missing or unreadable artifact files safely fall back to the available preview without interrupting execution.

* **Documentation**
  * Clarified output truncation, artifact storage, and resume behavior.


Wirasm:
## Ready to merge

The correction commit (`f2efa49e`) closes all five findings from the prior review. Both audit-event write sites that previously escaped the truncation-plus-spill mechanism (`node_always_run_reset`, `node_prior_cache_invalidated`) now route through the same `formatPersistedNodeOutput` + a new `persistedOutputEventFields` helper as the three sites already fixed, and the duplicated conditional-spread block is gone from all five sites. The reused-spill-filename race is closed: `getDagResumeSnapshot` now validates the spill's actual byte length against the row's own `node_output_original_bytes` before trusting it, falling back to the bounded preview (with a new `db.workflow_dag_node_output_spill_stale` warn log) on mismatch. A genuine script-node (`runtime: 'bun'`) end-to-end resume test now proves AC3 on the actual write path the earlier reverted fix broke, and a real 2-iteration `loop_group` test proves the per-iteration spill-key overwrite behaves as designed. All five claims were independently verified against the diff and by running the actual tests, not accepted from the disposition text.

Two new Suggestion-level observations surfaced during this pass — a small duplication the R1 fix introduced, and a narrow residual gap in the R2 fix's length-only validation — both non-blocking and both explicitly weighed as optional follow-ups by the reviewing agents.

**Validation:** Full local validation passed, including the complete monorepo `bun run validate` gate re-run at the current head.

**Resolved:** 5 (R1–R5, all previously OPEN, now FIXED)

### Findings

|---|---|---|---|
| `R1` | Important | `node_always_run_reset`/`node_prior_cache_invalidated` re-emits wrote unbounded prior output | FIXED |
| `R2` | Important | Spill file was a reused, unversioned pointer; a lost fire-and-forget write could leave an older row pointing at a different execution's content | FIXED |
| `R3` | Suggestion | The truncated-metadata conditional-spread block was duplicated verbatim at three sites | FIXED |
| `R4` | Important | The AC3 "resumed script `.field` access" regression test used `runtime: 'sh'` (bash dispatch), not a script node | FIXED |
| `R5` | Suggestion | No direct regression test for the loop_group per-iteration spill-key overwrite path | FIXED |
| `R6` | Suggestion | The R1 fix introduces a third near-identical `formatPersistedNodeOutput(...)` call-site pattern in `runLayers`, crossing the project's own rule-of-three threshold | OPEN |
| `R7` | Suggestion | R2's spill-staleness check compares byte length only; a same-length-but-different-content collision within the same crash window is not caught | OPEN |

<details>
<summary><code>R1</code> — Two sibling audit events reintroduced unbounded persisted output</summary>

**Impact:** A `bash:`/`script:` node whose prior output exceeded 32KB, when force-re-executed via `always_run: true` or invalidated via stale-dependency detection, persisted its full untruncated prior output into a fresh `node_always_run_reset`/`node_prior_cache_invalidated` DB row on every such resume pass.

**Evidence (fix):** `packages/workflows/src/dag-executor.ts:8356-8371` (`node_always_run_reset`) and `:8394-8414` (`node_prior_cache_invalidated`) now each compute `stepName = stepNamePrefix + node.id`, call `formatPersistedNodeOutput(priorCompletedNodes.get(node.id)?.output ?? '', artifactsDir, stepName)`, and spread the result via the new `persistedOutputEventFields(persistedOutput, 'prior_output')` (`dag-executor.ts:3335-3351`) — the same pattern the pre-existing `node_skipped_prior_success` site uses. `artifactsDir`/`stepNamePrefix` are the same variables already threaded to the fixed site (confirmed no shadowing). New tests `'bounds a large prior output in the node_always_run_reset audit event (#2726)'` and `'...node_prior_cache_invalidated audit event (#2726)'` (`dag-executor.test.ts:11740`, `:12218`) seed a 40,000-byte prior output and assert the DB-row preview stays ≤32,768 bytes, `_truncated`/`_original_bytes`/`_spill_path` are set correctly, and the spill file on disk holds the full original text.

**Verification:** `cd packages/workflows && bun test src/dag-executor.test.ts` → 606 pass, 0 fail, both named tests confirmed green individually via junit output. A grep of the file for any remaining `priorCompletedNodes.get(node.id)?.output` read that bypasses `formatPersistedNodeOutput` found none.

**Found by:** `prp-core:code-reviewer` (verification pass)

**Disposition:** FIXED — both sites route through the same bounded-preview+spill helper as the primary write sites, new tests prove it, tests pass.

</details>

<details>
<summary><code>R2</code> — Spill file addressed by a stable, reused filename with no write identity</summary>

**Impact:** On a narrow crash-timing window (spill file overwritten by a later execution, whose own event insert then never lands), a resumed run could silently read a *different* execution's content than the row it thought it was reading, with no error and no detectable mismatch.

**Evidence (fix):** `packages/core/src/db/workflow-events.ts:301-336` — `getDagResumeSnapshot` now reads the spill file, computes `Buffer.byteLength(spilled, 'utf8')`, and compares it against the *same row's* own `data.node_output_original_bytes` (written atomically alongside `node_output_spill_path` by `formatPersistedNodeOutput`/`persistedOutputEventFields` in one `data` object, one `createWorkflowEvent` call — confirmed no writer can produce `spill_path` without `original_bytes`). On mismatch it logs `db.workflow_dag_node_output_spill_stale` with `spillPath`/`expectedBytes`/`actualBytes` and falls back to the bounded preview, identical to the missing/unreadable-spill path. New test `'falls back to the preview when the spill file no longer matches the recorded byte length (#2726)'` (`workflow-events.test.ts:868-904`) seeds a real spill file with different-length content than the row claims and asserts both the fallback text and the exact warn payload.

**Verification:** The exact crash scenario from the original finding (row R1/content C1 durable, spill overwritten with C2 by a resume whose own event insert then crashes before landing, next resume reads R1 + the now-C2 spill file) was retraced against the new code: `len(C2) != len(C1)` is correctly caught and rejected. `cd packages/core && bun test src/db/workflow-events.test.ts` → 38 pass, 0 fail.

**Residual (tracked separately as R7):** the check is length-only; a same-length-but-different-content collision within the identical crash window is not caught. This is one of the two remedies the original finding explicitly proposed ("a content hash/length"), so the required outcome is satisfied — the narrower residual is recorded as its own finding below rather than reopening this one.

**Found by:** `prp-core:seam-analyzer` (verification pass)

**Disposition:** FIXED — the crash-window race the finding described is closed; verified by re-derivation against the new code and a passing regression test.

</details>

<details>
<summary><code>R3</code> — Truncated-metadata spread block duplicated verbatim at three sites</summary>

**Impact:** A future change to the truncation-metadata shape required three synchronized edits with nothing enforcing they stay in sync.

**Evidence (fix):** `persistedOutputEventFields(persistedOutput, fieldName)` (`dag-executor.ts:3335-3351`) extracts the exact conditional structure (truncated gate → `_truncated`/`_original_bytes`/optional `_spill_path`, parameterized by field name since it's now reused for both `node_output`- and `prior_output`-shaped events per R1). Reused at all five write sites: `executeBashNode:3501`, `executeScriptNode:3885`, `node_always_run_reset:8371`, `node_prior_cache_invalidated:8414`, `node_skipped_prior_success:8463`. `grep -n "_truncated: true"` across the file returns zero hits — no hand-duplicated copy remains anywhere.

**Verification:** Type safety was checked, not assumed — the destination (`IWorkflowStore.createWorkflowEvent`'s `data?: Record<string, unknown>` parameter) was already widened to `Record<string, unknown>` at the call site before this change, so the helper's return type is not a regression.

**Found by:** `prp-core:code-simplifier` (verification pass)

**Disposition:** FIXED — single helper, five call sites, no leftover duplication.

</details>

<details>
<summary><code>R4</code> — AC3 regression test exercised bash, not script, node dispatch</summary>

**Impact:** The load-bearing proof for AC3 (resumed `.field` access on a large **script**-node JSON output — the exact failure mode the earlier reverted fix broke) did not exercise `executeScriptNode`'s own write path.

**Evidence (fix):** New test `'resolves $node.output.field on resume for a REAL script:/runtime bun producer, via its actual spilled write (#2726)'` (`dag-executor.test.ts:6404`) declares its producer node `runtime: 'bun'`, confirmed to dispatch to `executeScriptNode` (not `executeBashNode`) at the runtime-branch check in `dag-executor.ts`. Pass 1 runs a real `executeDagWorkflow` producing an over-cap JSON blob, asserts truncation, and reads the *actual* spilled bytes back from disk via `readFile`. Pass 2 feeds those real bytes into a second simulated-resume `executeDagWorkflow` call via `priorCompletedNodes`, whose consumer node's `when: "$producer.output.status == 'PASS'"` must resolve for the consumer to run at all — asserted via a defined `node_completed` event for the consumer whose output length equals the full (not 32KB-capped) spilled content's length.

**Verification:** `cd packages/workflows && bun test src/dag-executor.test.ts` — the new test passes; the original bash-dispatch test (`'resolves $node.output.field on a resumed exec node whose prior output is a large JSON blob (#2726)'`, line 6335) is confirmed still present, not deleted — it remains valid, non-duplicate coverage of the generic kind-independent read mechanism, since it never actually executes `executeScriptNode` (the producer is seeded directly into `priorCompletedNodes`).

**Found by:** `prp-core:code-simplifier` (verification pass)

**Disposition:** FIXED — the script-specific write-path integration is now proven end-to-end by a real test, and the original test was correctly kept rather than replaced.

</details>

<details>
<summary><code>R5</code> — No direct test for the loop_group per-iteration spill-key overwrite</summary>

**Impact:** The mechanism (stable `step_name` → stable spill filename across loop_group iterations; last-write-wins on read) was previously only manually traced, not regression-protected.

**Evidence (fix):** New test `'overwrites the persisted-output spill file per iteration, keyed by the shared step_name (#2726)'` (`dag-executor.test.ts:20410`) runs a real 2-iteration `loop_group` with a bash body node emitting distinct 40,000-byte payloads (`a`-fill, then `b`-fill) gated by `until_bash`. Asserts: exactly 2 `node_completed` events for the body node; iteration 1's own persisted preview starts with `a` and iteration 2's with `b` (each row's own preview frozen at write time, not clobbered by the later overwrite); both rows point at the identical spill path (`group.emit.nodeoutput`); and the spill file on disk, after the run, holds only the *latest* iteration's content (`'b'.repeat(40000)`).

**Verification:** Confirmed present and green in the `bun test src/dag-executor.test.ts` run (606 pass, 0 fail).

**Found by:** `prp-core:code-reviewer` (verification pass)

**Disposition:** FIXED — the previously-manually-verified mechanism now has a real regression test.

</details>

<details>
<summary><code>R6</code> — R1's fix introduces a third near-identical call-site pattern, crossing rule-of-three</summary>

**Impact:** `dag-executor.ts:8356-8365` (`node_always_run_reset`) and `:8394-8401` (`node_prior_cache_invalidated`) each compute `const stepName = stepNamePrefix + node.id; const priorOutput = formatPersistedNodeOutput(priorCompletedNodes.get(node.id)?.output ?? '', artifactsDir, stepName);` — differing only by variable name. A pre-existing third instance of the identical shape already exists at `:8451-8455` (`priorSkipOutput`, for `node_skipped_prior_success`, untouched by this diff). That is three occurrences of one exact 4-line call pattern inside `runLayers`, crossing AGENTS.md's own stated threshold ("extract shared utilities only after the same pattern appears at least three times and has stabilized").

**Evidence:** `packages/workflows/src/dag-executor.ts:8356-8365`, `:8394-8401`, `:8451-8455`.

**Required outcome:** Optional — a small local closure over `node`/`priorCompletedNodes`/`artifactsDir` (e.g. `const formatPriorOutput = (stepName: string) => formatPersistedNodeOutput(priorCompletedNodes.get(node.id)?.output ?? '', artifactsDir, stepName)`) would collapse each site to one line without touching surrounding control flow or event-specific fields. Not required — this is duplication, not unsafe or premature structure.

**Found by:** `prp-core:code-simplifier` (verification pass — a defect this specific correction commit caused, in-scope per correction-verification rules)

**Disposition:** OPEN — non-blocking; a genuine, low-cost simplification opportunity introduced by the R1 fix, self-contained to one function.

</details>

<details>
<summary><code>R7</code> — R2's spill-staleness check is length-only; a same-length collision within the same crash window is not caught</summary>

**Impact:** `getDagResumeSnapshot`'s new stale-spill check (`workflow-events.ts:301-336`) rejects a spill whose byte length differs from the row's recorded `node_output_original_bytes`, but accepts one whose length happens to match despite different content. Within the exact crash window R2 describes (spill overwritten by a second execution whose own event insert never lands), if that second execution's output happens to be exactly the same byte length as the first — plausible for fixed-schema JSON, date-stamped reports, or other templated output — the stale content is still silently accepted.

**Evidence:** `packages/core/src/db/workflow-events.ts:301-336` — the guard is `spilledBytes !== data.node_output_original_bytes`, length only, no hash.

**Required outcome:** Optional follow-up — swap the length comparison for a content hash (e.g. SHA-256) stored alongside `node_output_original_bytes`, one of the two remedies the original R2 finding proposed. The preconditions for the residual (over-cap output on both executions, coincidentally identical byte length, AND a crash landing in the narrow synchronous-write-to-async-insert window) are compound enough that this is optional hardening, not a blocking gap.

**Found by:** `prp-core:seam-analyzer` (verification pass — a residual risk in the correction commit's own new code, in-scope per correction-verification rules)

**Disposition:** OPEN — non-blocking; the primary race R2 described is closed, this is a narrower, lower-probability residual worth an optional follow-up.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 (FIXED), R5 (FIXED) |
| `seams` | R2 (FIXED), R7 (new) |
| `simplify` | R3 (FIXED), R4 (FIXED), R6 (new) |

Scopes retained from the prior review (all three owned a finding under verification); no new scope added — this correction touches only already-typed `Record<string, unknown>` event-data construction, no new schema/type surface.

#### Validation

|---|---|---|
| `cd packages/workflows && bun test src/dag-executor.test.ts` | PASS | 606 pass, 0 fail, 1738 expect() calls (up from 602/1711 at the prior reviewed head — 4 new test blocks: R1×2, R4, R5) |
| `cd packages/core && bun test src/db/workflow-events.test.ts` | PASS | 38 pass, 0 fail (up from 37 — R2's new test) |
| `cd packages/core && bun run test` (full package script, 36 `bun test` invocations) | PASS | every invocation reports 0 fail, exit code 0 |
| `cd packages/workflows && bun run type-check` | PASS | `tsc --noEmit`, exit 0 |
| `cd packages/core && bun run type-check` | PASS | `tsc --noEmit`, exit 0 |
| `bun run validate` (repo root, full monorepo gate: check:bundled/-skill/-schema, check:pi-vendor-map, check:capability-matrix, type-check, `eslint . --max-warnings 0`, `prettier --check .`, test:install, full per-package `bun run test`) | PASS | exit 0; 150 distinct `0 fail` test-invocation results across the run, zero non-zero fail counts |

</details>

Wirasm:
Disposition on the two new Suggestions from the verify-corrections pass:

- **R6** (three near-identical `formatPersistedNodeOutput` call sites crossing rule-of-three): **FIXED** in cc6394fe. Extracted a small closure over `node`/`priorCompletedNodes`/`artifactsDir`, reused at all three sites (`node_always_run_reset`, `node_prior_cache_invalidated`, `node_skipped_prior_success`). No behavior change — all 606 `dag-executor.test.ts` tests still pass, full `bun run validate` green.

- **R7** (spill-staleness check is length-only, not content-hashed): **DECLINED**. The primary race R2 described (a resumed run silently trusting a spill overwritten by a different execution) is closed by the length check. Extending it to a content hash only helps the narrower, compound residual where a colliding execution's output happens to be *exactly* the same byte length AND lands inside the same narrow crash window as the original race. That's a real but low-probability compound scenario, and the added complexity (a new persisted hash field, hashing at write time, a schema-adjacent change to the event payload) isn't proportionate to closing it right now. Leaving as documented, optional hardening rather than building it speculatively.

Head is now `cc6394fe`.

Issue #2718, opened 2026-08-22
workflows: unify executeBashNode/executeScriptNode behind the exec kind, or explicitly decline to


Wirasm:
## Problem

PR #2716 (#2486, `DagNode` as a discriminated union by body kind) collapsed the old `BashNode`/`ScriptNode` distinction into one `ExecNode` kind (`kind: 'exec'`, discriminated internally by `runtime: 'sh' | 'bun' | 'uv'`). Every *type-level* consumer of the union now sees one `exec` case, matching the collapse the rest of the refactor made for `command`/`prompt` → `agent` and `approval` → `gate`.

`executeBashNode` and `executeScriptNode` (`packages/workflows/src/dag-executor.ts:3256-3477` and `:3529-3851`) were deliberately **not** merged into one `executeExecNode` behind that same `runtime` discriminant — they stayed two separate functions, called from two separate arms of the `runLayers` dispatch `switch (node.kind) { case 'exec': { if (node.runtime === 'sh') { ...executeBashNode... } else { ...executeScriptNode... } } }`.

That decision was made and documented in the PR (see its "Solution" section and the implementation report), but it was never filed as a standalone, linkable record — so it currently reads as unfinished work rather than a considered call, and the next person to touch either function has no signal that the split is intentional.

## Why this is worth a decision, not silence

The two functions genuinely differ beyond "which interpreter to shell out to": prompt/output substitution rules (`escapedForBash` shell-quoting vs. raw splicing for script argv), env-var delivery (`INPUTS_<NAME>` for scripts only), `deps:` handling (script-only, silently inert on bash per #2716's own review finding R8), and named-script-file resolution (script-only). Combined they're ~543 lines (221 + 322) of independently evolved logic — not two thin wrappers around one shared core.

Two legitimate outcomes exist, and this issue is where that decision should be made and recorded:

1. **Unify.** Extract the shared skeleton (retry wrapping, event emission, artifact writing) into one `executeExecNode`, with the interpreter-specific pieces (quoting, env delivery, `deps:`) as internal branches on `node.runtime`. Pro: one function to reason about per `case 'exec'`, matching the type-level collapse. Con: real risk of subtly changing behavior across 543 lines of independently-evolved logic for a refactor whose own stated goal was zero behavior change.
[...]

PR #2733, opened 2026-08-22
fix(workflows): bash loop_group body nodes now receive $LOOP_USER_INPUT

### LINKED ISSUE #2725: workflows: $LOOP_USER_INPUT never reaches a bash: node inside a loop_group body

Wirasm:
## Problem and outcome

`executeBashNode`'s subprocess environment hardcoded `LOOP_USER_INPUT` to `''`, unlike its sibling `executeScriptNode`, which threads the resumed loop_group gate's free-text feedback through a `loopUserInput` parameter into the same env var. A `bash:` node inside an interactive `loop_group` body that reads `$LOOP_USER_INPUT` through any indirect form — `${LOOP_USER_INPUT}` braced syntax, `printenv`, `env` — always saw an empty value regardless of what the reviewer typed at the gate.

- **Outcome:** A `bash:` node inside an interactive loop_group body sees `$LOOP_USER_INPUT` in its subprocess env whenever a `script:` node in the same position already would.
- **Invariant:** A `bash:` or `script:` `exec` node's subprocess environment carries the current iteration's `LOOP_USER_INPUT` value whenever one is available, regardless of how the script reads it.
- **Scope boundary:** No change to `executeScriptNode`, to the top-level (non-loop_group) bash/script env delivery, or to issue #2726 (unrelated `script:` node output truncation, delivered separately).
- **Root cause:** The engine already computes `ctx.bodyLoopUserInput` per resumed iteration and threads it into the `case 'exec':` dispatch. The script branch forwards it into `executeScriptNode`. The bash branch had no corresponding parameter or call-site argument, so `executeBashNode` always used its own hardcoded `''`.

**Correction to the originally filed issue:** the issue's own example — a bash node doing `echo "$LOOP_USER_INPUT"` (the bare literal token) — does not actually reproduce this failure on `dev`. Verified by test: a separate, pre-existing mechanism (`applyLoopPrevToBodyNode`, `packages/workflows/src/dag-executor.ts:4931-4939`, present since well before this issue was filed) already shell-quotes and splices that exact literal substring into bash script source before `executeBashNode` ever runs. The residual gap this PR fixes is real but narrower: any bash script that reads the variable a different way — because the splice's regex `/\$LOOP_USER_INPUT/g` only matches that exact literal substring, not `${LOOP_USER_INPUT}` or an indirect env read — never received the value. This PR closes that gap and brings `executeBashNode` to full parity with `executeScriptNode`'s env channel.

## Review guidance

- **Feedback requested:** Correctness of the parameter threading, and whether the corrected root-cause framing above is accurate.
- **Start here:** `packages/workflows/src/dag-executor.ts:3307` — the new `loopUserInput` parameter, positioned identically to `executeScriptNode`'s existing one.
- **Review order:** `dag-executor.ts:3307` (parameter) → `dag-executor.ts:3394` (env wiring) → `dag-executor.ts:8566` (call site) → the new test in `dag-executor.test.ts`.
- **Lower-attention areas:** None — this is a three-line change plus one test.

## Solution

`executeBashNode` now takes a `loopUserInput = ''` parameter in the same relative position `executeScriptNode` already uses (immediately before `execContext`), and its `subprocessEnv.LOOP_USER_INPUT` uses that parameter instead of a hardcoded `''`. The `case 'exec':` bash call site now passes `ctx.bodyLoopUserInput ?? ''`, exactly matching how the script call site already does. No new abstraction — this converges `executeBashNode` onto a pattern that already existed, working, one function away.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A `bash:` loop_group body node reading `$LOOP_USER_INPUT` via `printenv`/`${LOOP_USER_INPUT}`/`env` always sees `''` | Sees the reviewer's actual feedback text, on the first iteration of a resumed interactive loop_group |
| **Failure behavior** | N/A (silent empty value, no error) | N/A (no failure mode introduced) |

## Validation

- `cd packages/workflows && bun test src/dag-executor.test.ts -t "LOOP_USER_INPUT"` — 4 pass, 0 fail — proves the new bash-side test and existing script-side test both pass.
- New test verified failing against pre-fix code (`Received: ""`) and passing after the fix (`Received: "looks good, ship it"`) — proves this is a genuine regression test, not a vacuous one.
- `cd packages/workflows && bun test src/dag-executor.test.ts` (full file) — 599 pass, 0 fail — proves no regression to existing bash/script behavior.
- `cd packages/workflows && bun run type-check` — passed.
- `bun run validate` — exit 0, all per-package suites green.
- **Not verified:** No manual end-to-end run against a live workflow — the mocked-`execFileAsync` unit test directly asserts the subprocess env contents, which is the authoritative observable point for this fix.

## Links

- Closes #2725
- Plan: https://github.com/coleam00/Archon/issues/2725#issuecomment-5381908622

## Summary by CodeRabbit

* **Bug Fixes**
  * Resumed interactive workflow loops now correctly pass reviewer feedback to Bash and script steps.
  * Bash commands can access the latest loop input through the subprocess environment or inline substitution.
  * Clarified how loop input is provided during Bash and script execution.

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -19735,6 +19735,85 @@ describe('executeDagWorkflow -- loop_group node', () => {
     }
   });
 
+  it('delivers $LOOP_USER_INPUT to a resumed body bash node via env (#2725)', async () => {
+    // A body bash node's literal "$LOOP_USER_INPUT" token is already resolved by the
+    // loop_group-level shell-quoted splice (applyLoopPrevToBodyNode) before executeBashNode
+    // ever runs, so a script using that exact literal form is not a regression test for
+    // this fix. This test targets a script that reads the variable WITHOUT that literal
+    // token — `printenv`, matching how `${LOOP_USER_INPUT}` (braced form) or any other
+    // indirect env read behaves — which the splice cannot touch and which previously read
+    // as empty because executeBashNode's subprocess env hardcoded LOOP_USER_INPUT: ''.
+    const feedback = 'looks good, ship it';
+    const execSpy = spyOn(git, 'execFileAsync').mockResolvedValue({ stdout: 'DONE\n', stderr: '' });
+    try {
+      const mockDeps = createMockDeps();
+      const platform = createMockPlatform();
+      const workflowRun = makeWorkflowRun('lg-userinput-bash', {
+        metadata: {
+          approval: {
+            type: 'interactive_loop',
+            nodeId: 'grp',
+            iteration: 0,
+            message: 'Review and provide feedback.',
+          },
+          loop_user_input: feedback,
+          loop_feedback_given: true,
+        },
+      });
+
+      const nodes: DagNode[] = [
+        {
+          id: 'grp',
+          kind: 'loop_group',
+          loop_group: {
+            until: 'DONE',
+            max_iterations: 3,
+            fresh_context: true,
+            interactive: true,
+            gate_message: 'Review and provide feedback.',
+            nodes: [
+              {
+                id: 'emit',
+                kind: 'exec',
+                script: 'printenv LOOP_USER_INPUT; echo "DONE"',
+                runtime: 'sh',
+                depends_on: [],
+              },
+            ],
+          },
+          depends_on: [],
+        },
+      ];
+
+      await executeDagWorkflow(
+        mockDeps,
+        platform,
+        'conv-lg-userinput-bash',
+        testDir,
+        { name: 'lg-userinput-bash', nodes },
+        workflowRun,
+        'claude',
+        undefined,
+        join(testDir, 'artifacts'),
+        join(testDir, 'state'),
+        join(testDir, 'logs'),
+        'main',
+        'docs/',
+        minimalConfig
+      );
+
+      const bashCall = execSpy.mock.calls.find(
+        c => (c[0] as string) === 'bash' && (c[1] as string[])[0] === '-c'
+      ) as [string, string[], { env: NodeJS.ProcessEnv }] | undefined;
+      expect(bashCall).toBeDefined();
+      // The per-iteration feedback reaches the bash node through the environment —
+      // previously always '' regardless of what the reviewer typed at the gate.
+      expect(bashCall?.[2].env.LOOP_USER_INPUT).toBe(feedback);
+    } finally {
+      execSpy.mockRestore();
+    }
+  });
+
   it('fails the loop_group when max_iterations is exceeded without the until signal', async () => {
     let callCount = 0;
     mockSendQueryDag.mockImplementation(async function* () {
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -19803,7 +19803,7 @@ describe('executeDagWorkflow -- loop_group node', () => {
       );
 
       const bashCall = execSpy.mock.calls.find(
-        c => (c[0] as string) === 'bash' && (c[1] as string[])[0] === '-c'
+        c => (c[0] as string) === git.resolveBashPath() && (c[1] as string[])[0] === '-c'
       ) as [string, string[], { env: NodeJS.ProcessEnv }] | undefined;
       expect(bashCall).toBeDefined();
       // The per-iteration feedback reaches the bash node through the environment —
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -3304,6 +3304,10 @@ async function executeBashNode(
   envVars?: Record<string, string>,
   stepNamePrefix = '',
   iteration?: number,
+  // Per-iteration $LOOP_USER_INPUT free-text for loop_group body bash nodes, delivered via
+  // env (never spliced into source — #2115). '' for top-level bash nodes and non-first
+  // iterations (mirrors executeScriptNode, #2725).
+  loopUserInput = '',
   execContext: ExecutionContext = { kind: 'host' }
 ): Promise<NodeOutput> {
   const nodeStartTime = Date.now();
@@ -3387,7 +3391,7 @@ async function executeBashNode(
     BASE_BRANCH: baseBranch,
     USER_MESSAGE: workflowRun.user_message,
     ARGUMENTS: workflowRun.user_message,
-    LOOP_USER_INPUT: '',
+    LOOP_USER_INPUT: loopUserInput,
     LOOP_PREV_OUTPUT: '',
     REJECTION_REASON: '',
     CONTEXT: issueContext ?? '',
@@ -8559,6 +8563,7 @@ async function runLayers(ctx: RunLayersContext): Promise<void> {
                       config.envVars,
                       stepNamePrefix,
                       iteration,
+                      ctx.bodyLoopUserInput ?? '',
                       execContext
                     )
                 );
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -3305,8 +3305,11 @@ async function executeBashNode(
   stepNamePrefix = '',
   iteration?: number,
   // Per-iteration $LOOP_USER_INPUT free-text for loop_group body bash nodes, delivered via
-  // env (never spliced into source — #2115). '' for top-level bash nodes and non-first
-  // iterations (mirrors executeScriptNode, #2725).
+  // env (#2725). This is bash's SECOND delivery channel — the literal "$LOOP_USER_INPUT"
+  // token is already shell-quoted and spliced into the script source by
+  // applyLoopPrevToBodyNode before this function runs; this env var is what a script that
+  // reads the variable indirectly (${LOOP_USER_INPUT}, printenv, env) actually sees. ''
+  // for top-level bash nodes and non-first iterations.
   loopUserInput = '',
   execContext: ExecutionContext = { kind: 'host' }
 ): Promise<NodeOutput> {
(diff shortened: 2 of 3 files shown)


Wirasm:
## Ready to merge

A narrow, correctly-scoped bug-parity fix: `executeBashNode` now threads `$LOOP_USER_INPUT` into its subprocess env exactly as its sibling `executeScriptNode` already does, closing a real gap for indirect reads (`${LOOP_USER_INPUT}` braced form, `printenv`, `env`) inside an interactive `loop_group` body. The PR's own corrected root-cause framing was independently re-derived from the code, not taken on trust, and holds: the originally-filed issue's literal reproduction (`echo "$LOOP_USER_INPUT"`) does not fail on `dev`, because `applyLoopPrevToBodyNode` (`dag-executor.ts:4937-4939`) already shell-quotes and splices that exact literal token into bash source before `executeBashNode` ever runs — confirmed by tracing `skipUserInput`, which defaults `false` for `runtime === 'sh'` bodies (splice fires) and is only forced `true` for script bodies (splice skipped). The gap this PR actually closes — indirect reads bypassing that regex splice — is real and was previously untested. All three reviewers independently converged on exactly one non-blocking finding: three pre-existing comments (plus the PR's own new one) now understate or misstate that bash uniquely gets `$LOOP_USER_INPUT` through two channels, not one.

**Validation:** All PASS — focused test (4/4), full `dag-executor.test.ts` (599/599), `tsc --noEmit`, ESLint, Prettier. See coverage table for CI status.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Comments describing `$LOOP_USER_INPUT` delivery to bash still assert or imply a single (env-only or splice-only) channel, no longer accurate now that bash has both | OPEN |

<details>
<summary><code>R1</code> — Comments left inaccurate by bash gaining a second $LOOP_USER_INPUT delivery channel</summary>

**Impact:** No behavioral effect — purely a documentation-accuracy gap. A reader relying on any one of these comments in isolation could misjudge which channel(s) deliver `$LOOP_USER_INPUT` to a bash `loop_group` body node, in the specific area meant to document a shell-injection-safety guarantee (#2115). The most concrete risk (per the seam-analyzer): a future maintainer trusting the new comment's "never spliced into source" framing could conclude bash's splice-based shell-quoting for the literal token is now redundant with the env var and remove it as a "simplification" — which would reintroduce the injection/word-splitting hazard #2115 exists to prevent (though the existing `EDGE H` test at `dag-executor.test.ts:20874-20889` would catch that specific regression).

**Evidence (current state, all confirmed by direct read against the actual splice logic at `dag-executor.ts:4765-4944`):**
- `dag-executor.ts:3307-3309` — **new in this PR**: `executeBashNode`'s `loopUserInput` param comment reads "delivered via env (never spliced into source — #2115)". This is only true for the indirect-read case this PR fixes; it's copied from `executeScriptNode`'s comment (line 3584-3586) where it's genuinely and fully true, but for bash the literal token *is* separately spliced by `applyLoopPrevToBodyNode` before this parameter's value is ever consulted — a materially different (dual-channel) story that the comment reads as excluding.
- `dag-executor.ts:4357-4358` — pre-existing, made stale by this PR: "Deliver this iteration's approval-gate free-text to body script: nodes via env (never spliced into source — #2115); matches applyLoopPrevToBodyNode's skip." Now feeds bash too (call site `dag-executor.ts:8566`), and "matches applyLoopPrevToBodyNode's skip" is false for bash — `skipUserInput` is only forced true for script bodies; bash's call splices unconditionally.
- `dag-executor.ts:8044-8050` — pre-existing, made stale by this PR: the `RunLayersContext.bodyLoopUserInput` JSDoc says "for loop_group body `script:` nodes" and "top-level scripts have no loop user input" — both now incomplete, since the same field feeds the bash call site (`8566`) as well as the script one (`8597`).
- `dag-executor.ts:3584-3586` — pre-existing, made stale by this PR: `executeScriptNode`'s own sibling comment says bash "delivers loop input via quoted splice," which was the complete picture pre-PR but now undersells bash's second (env) channel.
- `schemas/dag-node.ts:384-386` — pre-existing, made stale by this PR: the #2718 rationale for keeping `executeBashNode`/`executeScriptNode` separate lists "env-delivered `$LOOP_USER_INPUT`/`deps:`/named-script-file resolution (script-only)" as a concrete differentiator. The `$LOOP_USER_INPUT` part of that claim is now false — env-delivered `$LOOP_USER_INPUT` is no longer script-only.

This exact class of finding — comments describing `$LOOP_USER_INPUT` delivery not reconciled after bash gained the env channel — was raised, confirmed, and fixed at Suggestion severity in the review of the prior (closed, unmerged) PR #2729, which originally implemented this same core change bundled with the unrelated #2726 fix. That correction rewrote four comments to name both channels explicitly; this PR reintroduces the same drift because it converges on the underlying code fix independently rather than by reusing that correction's wording.

**Required outcome:** Update the five comment sites above to state the real, asymmetric contract: bash delivers `$LOOP_USER_INPUT` via both the pre-existing shell-quoted splice (for the literal `$LOOP_USER_INPUT` token) and this PR's new env var (for indirect reads the splice can't reach); script delivers it via env only. Not required for this PR to merge — comment-only, no behavioral consequence, and precedent treats this class of finding as non-blocking.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier` (converged independently; seam-analyzer additionally traced the `schemas/dag-node.ts:385` site)

**Disposition:** OPEN. Non-blocking — comment/JSDoc accuracy only, no reachable behavioral defect, and the new regression test (`dag-executor.test.ts`, "delivers $LOOP_USER_INPUT to a resumed body bash node via env (#2725)") itself carries an accurate in-code comment explaining the dual-channel behavior, so the correct understanding already exists in the codebase — it just hasn't propagated to the production comments it describes.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1; fix verified correct and minimal (manually reverted the one-line fix, confirmed the new test fails as expected, restored) |
| `seams` | R1 (traced furthest, into `schemas/dag-node.ts`); Rule-of-Three re-check on `subprocessEnv` duplication (declined in prior #2729 review) reconfirmed unaffected — no third call site added |
| `simplify` | R1; confirmed no premature abstraction, no unjustified defensive code, new test proportionate to its sibling and file convention |

`types` was not selected: no typed contract (schema, constructor, public signature, state variant) changed — this diff adds one untyped optional parameter with a default, mirroring an existing sibling parameter exactly.

#### Validation

|---|---|---|
| `bun test src/dag-executor.test.ts -t "LOOP_USER_INPUT"` (packages/workflows, review checkout at `5d09c21c`) | PASS | 4 pass, 0 fail, 15 `expect()` calls |
| `bun test src/dag-executor.test.ts` (packages/workflows, full file) | PASS | 599 pass, 0 fail, 1686 `expect()` calls |
| `bun x tsc --noEmit` (packages/workflows) | PASS | Clean, no errors |
| `bun x eslint src/dag-executor.ts --max-warnings 0` (packages/workflows) | PASS | 0 errors, 0 warnings |
| `bun x prettier --check` (2 changed files) | PASS | "All matched files use Prettier code style!" |
| `gh pr checks 2733` | PARTIAL | `schema-upgrade` passed; `test (ubuntu-latest)`, `test (windows-latest)`, `docker-build`, `CodeRabbit` still pending at verification time (PR opened same day) — not a PR-caused failure, not required for this verification's scope given the local reproduction above covers the same test suite CI runs. |

</details>

Wirasm:
R1 (comment accuracy — bash's dual `$LOOP_USER_INPUT` delivery channels no longer reconciled) fixed in 6a35e6fc: rewrote all five identified comment sites (`dag-executor.ts:3307-3312`, `:4360-4365`, `:8051-8059`, `:3587-3592`, and `schemas/dag-node.ts:383-397`) to state the real, asymmetric contract — bash delivers via both the pre-existing literal-token splice and the new env channel; script delivers via env only.

Wirasm:
windows-latest CI failure fixed in 42dcd6d7: the new test's mocked-exec matcher hardcoded the literal string 'bash', but `resolveBashPath()` only returns that literal on non-Windows — on windows-latest it returns a resolved Git-Bash path instead, so the mocked call was never found (`bashCall` came back `undefined`). Now compares against `git.resolveBashPath()`, matching the existing convention elsewhere in this file (line ~2585). Rerunning CI.

PR #2739, opened 2026-08-22
fix(core): manage_run's reject action now defaults empty text to 'Rejected'

### LINKED ISSUE #2734: manage_run's reject action doesn't default empty text to 'Rejected'
## Problem

`manage_run`'s `respond` action defaults an empty rejection message to `'Rejected'` (fixed by #2730's R5/R9 findings, matching the same default already applied to `command-handler.ts` and `api.ts`'s respond surfaces). Its sibling `case 'reject':` action has the identical gap and was never fixed, because it predates PR #2730 and R9 scoped itself to the `respond` action only.

`packages/core/src/orchestrator/manage-run-tool.ts:418-419`:
```ts
case 'reject': {
  const result = await rejectWorkflow(id, message.length > 0 ? message : undefined);
```

Compare to `respond`'s fixed version at `manage-run-tool.ts:437-438`:
```ts
const respondText =
  message.length > 0 ? message : decision === 'reject' ? 'Rejected' : undefined;
const result = await respondToWorkflow(id, decision, respondText);
```

Calling `manage_run action=reject` with no message passes `undefined` through to `rejectWorkflow`. On the new-mode structured-output path (`workflow-operations.ts:697`, `text: reason ?? ''`), this writes `structured_output.text: ''` instead of `'Rejected'` — the same defect R9 fixed for `respond`, just on the older action. `rejectWorkflow`'s own internal default (`rejectReason = reason ?? 'Rejected'`, `workflow-operations.ts:644`) only feeds the audit event and the legacy stage-rework/cancel path, not the new-mode structured output, so it doesn't cover this.

## Solution

Apply the same default `case 'reject':` in `manage-run-tool.ts` already gets on `respond`:

```ts
case 'reject': {
  const rejectText = message.length > 0 ? message : 'Rejected';
  const result = await rejectWorkflow(id, rejectText);
```

## Context

Found during PR #2730's review (finding R9, https://github.com/coleam00/Archon/pull/2730#issuecomment-5382020658), which fixed the identical gap for the newer `respond` action but explicitly declined to extend the fix to `reject` since it predates that PR and was out of its scope. Low priority, narrow one-line fix, no design decision involved.

Wirasm:
## Problem and outcome

`manage_run`'s `reject` action passed `undefined` to `rejectWorkflow` when called with no message. On the new-mode structured-output path, that produced `structured_output.text: ''` instead of a meaningful default — the sibling `respond`-style default for this exact gap already exists elsewhere (`command-handler.ts`, `api.ts`), it just was never applied here.

- **Outcome:** Calling `manage_run action=reject` with no text now records `'Rejected'` as the structured output text, matching the same default already used by the `command-handler.ts` and `api.ts` reject/respond surfaces. (The CLI `workflow reject` and Slack reject-button call sites have the identical unfixed gap — tracked separately in #2740, deliberately out of this PR's scope, the same way #2734 itself was carved out of an earlier PR.)
- **Invariant:** An empty or whitespace-only rejection message never reaches `structured_output.text` as `''`.
- **Scope boundary:** No change to `rejectWorkflow`, `respondToWorkflow`, or any other `manage_run` action — only the `reject` case's call-site default.
- **Root cause:** `case 'reject':` in `manage-run-tool.ts` computed `message.length > 0 ? message : undefined` and passed that straight to `rejectWorkflow`, so an empty message became `undefined`, which the new-mode branch in `workflow-operations.ts` (`text: reason ?? ''`) turns into an empty string rather than a default.

## Review guidance

- **Feedback requested:** Correctness only — this is a narrow, well-precedented one-line fix.
- **Start here:** `packages/core/src/orchestrator/manage-run-tool.ts:388-390` — the fixed `case 'reject':` block.
- **Review order:** `manage-run-tool.ts` diff, then the new test in `manage-run-tool.test.ts`.
- **Lower-attention areas:** None — the whole diff is 14 lines.

## Solution

Mirror the existing default pattern already used for the `respond` action: compute `rejectText = message.length > 0 ? message : 'Rejected'` and pass that to `rejectWorkflow` instead of `message.length > 0 ? message : undefined`. `rejectWorkflow`'s own internal default (`rejectReason = reason ?? 'Rejected'`) only feeds the audit event and the legacy stage-rework/cancel path — it doesn't cover the new-mode `structuredOutput.text` field, so the fix has to happen at this call site.

## Validation

- `bun test packages/core/src/orchestrator/manage-run-tool.test.ts` — passed (37 pass, 0 fail) — proves the new regression test (`reject with no message defaults the rejection text to "Rejected"`) and every existing reject/respond test still hold.
- `bun run test` — passed (0 fail across every package) — proves no regression elsewhere.
- `bun run validate` — passed (exit 0) — proves types, lint, format, and the full suite all pass.
- **Not verified:** Nothing material — this is a pure default-value fix fully covered by the added unit test.

## Links

- Closes #2734
- Plan: https://github.com/coleam00/Archon/issues/2734#issuecomment-5382407795

### DIFF
--- a/packages/core/src/orchestrator/manage-run-tool.test.ts
+++ b/packages/core/src/orchestrator/manage-run-tool.test.ts
@@ -412,6 +412,18 @@ describe('manage_run — destructive confirmation gate', () => {
     expect(mockReject).toHaveBeenCalledWith('r1abcdef-1234', 'no');
   });
 
+  test('reject with no message defaults the rejection text to "Rejected"', async () => {
+    mockFindByPrefix.mockResolvedValue([makeRun({ status: 'paused' })]);
+    mockReject.mockResolvedValue({
+      workflowName: 'wf',
+      cancelled: true,
+      maxAttemptsReached: false,
+    });
+    const tool = buildManageRunTool({ codebaseId: CODEBASE_ID });
+    await tool.handler({ action: 'reject', runId: 'r1abcdef', confirm: true });
+    expect(mockReject).toHaveBeenCalledWith('r1abcdef-1234', 'Rejected');
+  });
+
   test('reject with confirm and an on-reject prompt reports rework, not cancellation', async () => {
     mockFindByPrefix.mockResolvedValue([makeRun({ status: 'paused' })]);
     mockReject.mockResolvedValue({
--- a/packages/core/src/orchestrator/manage-run-tool.ts
+++ b/packages/core/src/orchestrator/manage-run-tool.ts
@@ -386,7 +386,8 @@ async function handleWrite(
         : `Feedback recorded for ${result.workflowName} (${id.slice(0, 8)}); the loop runs another iteration with it.${continues}`;
     }
     case 'reject': {
-      const result = await rejectWorkflow(id, message.length > 0 ? message : undefined);
+      const rejectText = message.length > 0 ? message : 'Rejected';
+      const result = await rejectWorkflow(id, rejectText);
       if (result.cancelled) {
         const suffix = result.maxAttemptsReached ? ' (max attempts reached)' : '';
         return `Rejected and cancelled ${result.workflowName} (${id.slice(0, 8)})${suffix}. Nothing further runs.`;


Wirasm:
## Ready to merge

The fix is exactly what the plan and PR body describe: `manage_run`'s `case 'reject':` now defaults an empty rejection message to `'Rejected'` before calling `rejectWorkflow`, closing the gap where `undefined` produced `structured_output.text: ''` on the new-mode gate-resolution path. All three reviewers independently confirmed the 14-line diff is correct, minimal, and already in the same shape as its two existing siblings (`command-handler.ts`, `api.ts`). The one substantive finding is about code this PR does **not** touch — two other `rejectWorkflow` call sites (CLI `workflow reject`, the Slack reject button) have the identical unfixed gap — which is a legitimate, deliberately out-of-scope, pre-existing bug, not a defect in this PR. It's now filed as a tracked follow-up.

**Validation:** All PASS in a detached worktree pinned to the reviewed head — full type-check, full lint (`--max-warnings 0`), the focused test file, and the full monorepo test suite.

**Tracked follow-ups:** https://github.com/coleam00/Archon/issues/2740

### Findings

|---|---|---|---|
| `R1` | Important | CLI `workflow reject` and the Slack reject button still pass an unguarded/absent reason to `rejectWorkflow`, reproducing the identical `structured_output.text: ''` defect this PR fixes for `manage_run` only | TRACKED FOLLOW-UP |

<details>
<summary><code>R1</code> — CLI and Slack reject call sites don't get this PR's default</summary>

**Impact:** On a gate authored with `approval.decisions:` (the new-mode structured-output resolution path), `archon workflow reject <run-id>` with no reason (CLI — both `--json` and plain branches, and the `--detach` re-spawn, which forwards the same argv) records `structured_output.text: ''` instead of `'Rejected'`. Every Slack reject-button click on such a gate does the same, deterministically — a Slack button has no text-entry affordance, so this call site can never supply a reason at all. `$nodeId.output.text` readers downstream and the run's audit trail see an empty string.

**Evidence:**
- `packages/cli/src/commands/workflow.ts:3315,3332` — `workflowRejectCommand(runId, reason?, ...)` passes its optional `reason` straight through to `rejectWorkflow(resolvedId, reason)` with no default.
- `packages/adapters/src/chat/slack/workflow-bridge.ts:482` — `workflowOperations.rejectWorkflow(runId)` is called with no second argument at all.
- `packages/core/src/operations/workflow-operations.ts:697` — `const structuredOutput = { decision: 'reject', text: reason ?? '' };` is the shared seam both routes fall through to; `workflow-operations.ts:644`'s `rejectReason ?? 'Rejected'` default only covers the audit event and legacy stage-rework/cancel path, not this field.
- Contrast with the two call sites that already guard: `packages/core/src/handlers/command-handler.ts:1017` (`args.slice(2).join(' ') || 'Rejected'`) and `packages/server/src/routes/api.ts:3832` (`body.reason ?? 'Rejected'`).
- The PR body's claim that the fix matches "every other reject/respond surface in the codebase" is inaccurate as written — it matches 2 of 4 other surfaces; the CLI and Slack surfaces don't match. This doesn't change the correctness or scope of the diff itself, only the justification text.

**Required outcome:** No change required on this PR — the plan's own "Not building" scope boundary (single named call site) is a legitimate, previously-established pattern (this is exactly how #2734 was itself carved out of PR #2730's review). Filed as a tracked follow-up: https://github.com/coleam00/Archon/issues/2740, scoped to the two remaining call sites plus (optional, not required) making `rejectWorkflow`'s `reason` parameter non-optional so a future fifth caller can't reintroduce the same gap silently.

**Disposition:** TRACKED FOLLOW-UP — verified via GitHub issue https://github.com/coleam00/Archon/issues/2740, filed during this review with the exact call sites, evidence, and suggested fix shape.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 |
| `seams` | R1 |
| `simplify` | No additional findings — reviewer confirmed the intermediate `rejectText` variable and the new test's scope both match the two existing sibling patterns in the codebase; nothing to cut |

`types` was not selected — the diff changes no type, schema, constructor, factory, public signature, or state variant (it changes a call-site default value only).

#### Validation

|---|---|---|
| `bun test packages/core/src/orchestrator/manage-run-tool.test.ts` | PASS | 37 pass, 0 fail, 93 expect() calls — includes the new regression test asserting `mockReject` is called with `'Rejected'` on an empty-message reject |
| `bun run type-check` | PASS | All 13 packages report exit 0 (`@archon/paths` through `@archon/cli`, plus `docs-web`) |
| `bun run lint --max-warnings 0` | PASS | Exit 0, zero warnings across the repo |
| `bun run test` (full monorepo suite) | PASS | Exit 0, `0 fail` reported by every package (`@archon/core`, `@archon/cli`, `@archon/server`, `@archon/workflows`, `@archon/providers`, `@archon/isolation`, `@archon/web`, etc.); the many `err:`/`fail`-containing log lines in the raw output are intentional structured-error-path logging inside passing tests, not test failures |

All checks ran in a detached worktree (`.worktrees/review-pr-2739`) pinned to `55ce4afdf28a37e7515de7d23b89dee41061b308` (`git rev-parse HEAD` verified equal to `reviewed_head` before and after validation).

</details>

PR #2750, opened 2026-08-22
fix(workflows): interactive loop_group resume now restores $LOOP_PREV body outputs

### LINKED ISSUE #2748: fix(workflows): interactive loop_group resume loses $LOOP_PREV — body outputs are not restored across the pause
## Problem

An interactive `loop_group`'s pause does not persist the body-output map that feeds `$LOOP_PREV.*` — so the first iteration after a resume resolves every `$LOOP_PREV.<bodyId>.output[.field]` ref to `''`, where an un-paused next iteration would have seen the previous iteration's outputs.

Mechanism (all in `packages/workflows/src/dag-executor.ts`, current dev):

- `loopPrevOutputs` is a local variable in `executeLoopGroupNode` (`let loopPrevOutputs: Map<string, NodeOutput> | undefined; // undefined on iteration 1`, ~line 4281), populated after each iteration's body finishes.
- The `interactive_loop` pause payload (~lines 4774–4797) persists the session cursor and signal state (`sessionId`/`sessionProvider`, `completionSignaled`, `signaledOutput`, `signaledStructuredOutput`) — and nothing else. `loopPrevOutputs` is not snapshotted.
- On resume, `executeLoopGroupNode` is a fresh call: the only state restored from `ApprovalContext` is the session cursor. `loopPrevOutputs` starts `undefined`, so the resumed iteration's `prevSnapshot` is `undefined` and every `$LOOP_PREV.*` ref substitutes to `''` (the documented iteration-1 semantics, applied to an iteration that is not iteration 1).

## Reproduction shape

An interactive `loop_group` whose body has node `draft` and a second node reading `$LOOP_PREV.draft.output` (or `draft` itself reading its own previous output). Run, let iteration 1 complete, respond at the gate with feedback text (so another iteration runs). The resumed iteration sees `$LOOP_PREV.draft.output` as `''` — the model is asked to improve on prior work it cannot see, while `$LOOP_USER_INPUT` (delivered correctly) references that invisible prior output.

## Why it matters now

#2707 step 3 re-expresses the interactive loop as `loop_group` + a gate in the body — the step's plan flagged exactly this question as its one open engineering risk ("whether today's `group.interactive` pause already has an answer... or whether it's a live, unnoticed gap"). Answer: live gap, no snapshot exists. The step should fix it rather than inherit it, and the fix is independently a bug fix for today's mechanism.

## Desired outcome

- After an interactive loop_group resume, `$LOOP_PREV.*` resolves exactly as it would have without the pause, for any output size.
- Fix-shape recommendation: **re-derive the map at resume from the persisted `<groupId>.<bodyId>` `node_completed` events** (latest iteration's rows) rather than snapshotting the map into `ApprovalContext` — body outputs can be large and run metadata should not grow with them, while the event-log read path already has bounded rows + spill/rehydrate resume-safety from #2726/PR #2732. The wrinkle to solve: selecting the *latest* iteration's row per body node (`data.iteration` is persisted on iteration events; body `node_completed` rows need the equivalent selection key).
- A regression test: pause → respond-with-text → resume, asserting the resumed body prompt received the prior iteration's output, including one oversized (>32KB) body output to pin the spill path.

Wirasm:
## Problem and outcome

An interactive `loop_group`'s pause/resume didn't preserve `$LOOP_PREV.*` — the body-output map that lets an iteration reference what a sibling (or itself) produced last time. On resume, the first iteration processed always saw every `$LOOP_PREV.<bodyId>.output` as `''`, even though the real prior output existed and was already persisted. A draft/review/refine loop resumed after human feedback effectively asked the model to improve on work it could no longer see.

- **Outcome:** `$LOOP_PREV.*` on the first iteration after an interactive-gate resume now resolves to the real previous iteration's body output — identical to what an un-paused next iteration would have seen — for any output size, including outputs large enough to spill to disk.
- **Invariant:** `$LOOP_PREV.<bodyId>.output[.field]` on the resumed iteration must equal what it would have resolved to without the intervening pause.
- **Scope boundary:** Only the single-level interactive `loop_group` case the issue reproduces. A nested interactive `loop_group` (gate inside an inner group's body, itself inside an outer group's body) has broader pre-existing resume semantics beyond `$LOOP_PREV` and is already documented as "not hardened in v1" — untouched here. The identical structural gap in the single-node `loop:`'s `$LOOP_PREV_OUTPUT` is a different code path, not raised by this issue, and also untouched.
- **Root cause:** `loopPrevOutputs` is a local variable scoped to one `executeLoopGroupNode` call, populated only from in-process iterations. The interactive-gate pause payload (`ApprovalContext`) persists the session cursor and completion-signal state, but never a body-output snapshot. Resume re-enters `executeLoopGroupNode` as a brand-new call, so `loopPrevOutputs` starts `undefined` again and every `$LOOP_PREV.*` ref resolves via the same "no prior output" branch a genuine iteration 1 hits.

## Review guidance

- **Feedback requested:** Correctness of the reconstruction (does `outerNodeOutputs` really already carry what I claim on every production resume path, not just in the test harness), and whether the oversized-output test convincingly proves the spill path.
- **Start here:** `packages/workflows/src/dag-executor.ts:4281` — the fix itself, ~16 lines.
- **Review order:** `dag-executor.ts:4281` (fix) → `dag-executor.ts:10015-10059` (the existing, unrelated top-level pre-population the fix depends on — unchanged, cited for context) → `dag-executor.test.ts` (the two updated/new tests) → `loop-nodes.md` (doc caveat removal).
- **Lower-attention areas:** The doc change is a mechanical removal of a now-false caveat; no new claims added.
- **Known risk or uncertainty:** None identified. The reconstruction reuses data (`outerNodeOutputs`) that was already being computed and threaded through for an unrelated reason (top-level `$node.output` resume support), so this fix adds no new state, query, or persisted field.

## Solution

`executeDagWorkflow` already pre-populates its top-level `nodeOutputs` map from every persisted `node_completed` row on a resume (`dag-executor.ts:10015-10059`), unconditionally — it doesn't filter by whether a row's `step_name` matches a top-level node id, so a loop_group body node's `<groupId>.<bodyId>`-keyed row lands there too. That map is exactly the `outerNodeOutputs` parameter `executeLoopGroupNode` already receives.

The fix reads those entries back: on `isLoopResume`, for each of this group's direct body ids, look up `outerNodeOutputs.get(bodyStepNamePrefix + id)` and re-key it to the bare id `substituteLoopPrevRefs` looks up. This also resolves the "which iteration's row" question for free — a loop_group body node's persisted `step_name` is the *same* string across every iteration, and the underlying `getDagResumeSnapshot` read already folds chronologically-ordered rows per `step_name`, so the last (latest) iteration's row is what survives. No new `ApprovalContext` field, no new DB query, and it inherits the existing #2726/#2732 spill/rehydrate machinery (a large prior output already reads back its full spilled bytes, not a truncated preview) with no extra code.

## Validation

- `cd packages/workflows && bun test src/dag-executor.test.ts -t "LOOP_PREV"` — 14 pass, 0 fail — proves both the inverted regression test (general resume case) and the new oversized/spill regression test.
- `cd packages/workflows && bun test src/dag-executor.test.ts` — 608 pass, 0 fail — proves no regression across the full loop_group/loop/DAG-executor suite, including the sibling resume test covering session-cursor and `$LOOP_USER_INPUT` restoration.
- `bun run validate` — full repo gate green (type-check, lint, format, all package test suites, bundled/schema/capability-matrix checks).
- **Not verified:** No live end-to-end manual run against a real Archon server/CLI (approve a paused run and observe the resumed prompt). The unit-level harness constructs `priorCompletedNodes` the same way `hydrateResumableRun` does for a real resume (confirmed by reading its production callers in the orchestrator and CLI), so this is considered adequately covered without a manual run.

## Links

- Closes #2748
- Plan: https://github.com/coleam00/Archon/issues/2748#issuecomment-5382807363

## Summary by CodeRabbit

* **Bug Fixes**
  * Interactive loop resumes now correctly restore outputs from the previous iteration.
  * `$LOOP_PREV.*` references continue working after a pause and resume.
  * Undeclared `$LOOP_PREV` fields are properly rejected after resuming.
  * Large outputs remain available in full after resuming an interactive loop.

* **Documentation**
  * Updated loop guidance to reflect restored previous-iteration outputs after interactive resumes.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/loop-nodes.md
+++ b/packages/docs-web/src/content/docs/guides/loop-nodes.md
@@ -738,10 +738,10 @@ Two distinct cases:
 - **Interactive-gate resume** (`/workflow approve <id>`): the loop continues
   with the **next** iteration's whole body. With `fresh_context: false`, the
   body's AI session continues from where it paused (the session cursor is
-  persisted across the gate). `$LOOP_PREV.*` refs, however, resolve to an empty
-  string on the first resumed iteration — the prior iteration's body-output
-  snapshot is **not** carried across the gate (same caveat as
-  [`$LOOP_PREV_OUTPUT`](#retry-on-failure-with-loop_prev_output)).
+  persisted across the gate). `$LOOP_PREV.*` refs on the first resumed
+  iteration resolve to the prior iteration's real body outputs, restored from
+  the persisted run history — the same as an un-paused next iteration would
+  have seen.
 - **Failure resume** (`/workflow resume <id>` after a crash/failure): there is
   no persisted iteration cursor — the loop_group node restarts from
   **iteration 1**. Per-body-node resume granularity is not supported in v1.
@@ -758,8 +758,6 @@ Two distinct cases:
 - Per-body-node resume (skip-to-failed-body-node) — the whole iteration re-runs.
 - `$LOOP_PREV.<id>.output[N]` history indexing — only the immediately prior
   iteration is reachable.
-- `$LOOP_PREV.*` across an interactive pause/resume boundary — resolves to an
-  empty string on the resumed iteration.
 
 Nested `loop_group` inside a `loop_group` body is supported by construction
 (the body is a normal `nodes` array), but is not hardened in v1.
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -22564,12 +22564,13 @@ describe('executeDagWorkflow -- loop_group node', () => {
     expect(pauseCalls2.length).toBe(0);
   });
 
-  it('INTERACTIVE resume: $LOOP_PREV is NOT preserved across the pause/resume boundary (v1 known limitation)', async () => {
-    // v1 behavior: on interactive resume, loopPrevOutputs resets to undefined (the prior
-    // process's body-output snapshot is not persisted in ApprovalContext). So the resumed
-    // iteration's $LOOP_PREV.<bodyNode>.output resolves to '' — NOT to the paused run's
-    // iteration-1 output. This test locks the current behavior; persisting $LOOP_PREV
-    // across resume is a tracked follow-up (CodeRabbit finding #5).
+  it('INTERACTIVE resume: $LOOP_PREV resolves to the prior iteration real body output (#2748)', async () => {
+    // On interactive resume, executeLoopGroupNode is a fresh call whose local
+    // loopPrevOutputs starts undefined — but the paused run's iteration-1 body output
+    // was already persisted as a `refine.work` node_completed row. The resumed
+    // executeDagWorkflow call re-derives loopPrevOutputs from that persisted row (via
+    // outerNodeOutputs, pre-populated from the priorCompletedNodes snapshot passed in
+    // below), so $LOOP_PREV.<bodyNode>.output resolves to the real prior output, not ''.
     mockSendQueryDag.mockImplementationOnce(async function* () {
       yield { type: 'assistant', content: 'iter1 body output XYZ' };
       yield { type: 'result', sessionId: 'lg-prev-sess-1' };
@@ -22626,7 +22627,11 @@ describe('executeDagWorkflow -- loop_group node', () => {
     const iter1Prompt = mockSendQueryDag.mock.calls[0][0] as string;
     expect(iter1Prompt).toContain('PREV=<<>>');
 
-    // ---- Resume: iteration 2.
+    // ---- Resume: iteration 2. priorCompletedNodes carries the persisted iteration-1
+    // body row (the `refine.work` node_completed row a real DB would return from
+    // getDagResumeSnapshot) — this is what a real resume threads in via
+    // hydrateResumableRun before calling executeDagWorkflow.
+    const priorCompletedNodes = new Map([['refine.work', { output: 'iter1 body output XYZ' }]]);
     mockSendQueryDag.mockImplementationOnce(async function* () {
       yield { type: 'assistant', content: 'final\nAPPROVED' };
       yield { type: 'result', sessionId: 'lg-prev-sess-2' };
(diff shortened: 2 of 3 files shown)


Wirasm:
## Ready to merge

Correction commit `3e2639da` closes the one blocking finding from the prior review (`R1`) and documents the one clarity nit (`R2`); all three reviewers who verified it — independently, from correctness, seam-drift, and simplification angles — confirm the fix by reading both derivation sites rather than trusting the commit message, and none found a new defect the correction introduces. `R3` stays declined, and a fresh read of that disposition still holds: the untouched test proves something its in-process sibling doesn't (the spill/rehydrate path surviving a real pause/resume boundary), so the line-count cost is proportionate to what it protects. All focused and full-suite tests, type-check, and lint were re-run against the current head and are green.

**0 blocking · 1 non-blocking (already resolved by correction, kept visible for the record)**

**Validation:** All green — 15/15 focused `LOOP_PREV` tests (was 14), 609/609 full `dag-executor.test.ts` suite (was 608), `bun run type-check` clean across every package, `bun run lint` clean — all re-run in this review's own checkout at `3e2639da`.

**Resolved:** 2 (`R1`, `R2`) · **Declined:** 1 (`R3`, re-confirmed reasonable)

### Findings

|---|---|---|---|
| `R1` | Important | Resumed `$LOOP_PREV.<id>.output.<field>` silently loses schema strictness for schema-declaring body nodes | FIXED |
| `R2` | Suggestion | `if (restoredLoopPrevOutputs.size > 0)` guard is behaviorally inert | FIXED |
| `R3` | Suggestion | New oversized-output test's fresh-run scaffolding could be shorter | DECLINED |

<details>
<summary><code>R1</code> — Restored body-node <code>NodeOutput</code> loses <code>declaredFields</code> on resume</summary>

**Impact (as originally found):** A `loop_group` body node that declares `output_format` lost its strict field-access contract for exactly one iteration — the first one after an interactive-gate resume. A downstream `$LOOP_PREV.<bodyId>.output.<undeclaredField>` reference threw `OutputRefError('not-in-schema')` on every in-process iteration, but on the resumed iteration `declaredFields` was unconditionally absent from the reconstructed `NodeOutput`, so the reference fell through to the schemaless resolution branch instead.

**Correction (`3e2639da`):** `packages/workflows/src/dag-executor.ts:4290-4317`. The `isLoopResume` reconstruction now builds `bodyNodesById` from `group.nodes as DagNode[]` (the same cast already used twice more in this function, at lines ~4393 and ~5022, justified by `include:` directives being fully expanded at load time before the executor runs — confirmed via `include-expander.ts`'s recursive `expandNodeList` over `loop_group.nodes` and the executor's own "unexpanded include reached the executor" fail-fast guard), looks up each restored body node's own current definition, computes `declaredFieldsFromSchema(bodyNodeDef.output_format)` — the identical derivation the natural per-iteration path uses at line ~3111 and the top-level resume pre-population uses at line ~10070 — and spreads it onto the restored row only when defined.

**Verification (independent, by two reviewers from different angles):**
- `prp-core:code-reviewer` traced both derivation sites, confirmed the spread never clobbers a value `prior` already had correctly (`declaredFields` was unconditionally absent going in, since the pre-population's `nodesById` is scoped to top-level `workflow.nodes` and can never match a dotted `<groupId>.<bodyId>` key), confirmed the `bodyNodesById` cast is safe, and reasoned through why the new test is discriminating: pre-fix, no `not-in-schema`-flavored `node_failed` event fires (the assertion target), so the test would have failed against the pre-correction code.
- `prp-core:seam-analyzer` independently confirmed the two derivation sites now agree on every field `resolveNodeOutputField` actually reads (`state`, `declaredFields`, `structuredOutput`, `output`), checked for other silently-dropped sibling fields (`sessionId`/`resumed` are still absent from the restored row, but that gap predates this PR, is inherited unchanged from the pre-existing pre-population block, and is unread by anything on this seam — not a new or reopened divergence), and confirmed the `as DagNode[]` cast matches a load-time-enforced invariant rather than reopening the same bug class.
- Both re-ran `bun test src/dag-executor.test.ts -t "LOOP_PREV"` / `-t "R1"` standalone and got clean passes against the current head.
- The coordinator independently re-ran the full suite (609/609, was 608), `bun run type-check` (clean across all 10 packages), and `bun run lint` (clean, zero warnings) against `3e2639da`.

One footnote from the code-reviewer's trace, noted for the record but not affecting the verdict: the original finding's "silently resolves to `''`" description is the mechanism for a truly *absent* field; the new test's `extra` field is actually *present* in the raw prior JSON, so pre-fix it would have silently returned that stale real value rather than `''` — arguably a worse outcome than described, not a better one. Either way, the fix and its regression test close the same underlying gap: post-fix, an undeclared field on a schema-declaring body node throws `not-in-schema` on the resumed iteration exactly as it does in-process.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer` (original); re-verified by both plus a third independent pass from `prp-core:code-simplifier`'s scope check (no concern raised on the `bodyNodesById` construction cost).

**Disposition:** FIXED — verified against the actual code by two independent reviewers, not merely the commit message; new regression test proven discriminating; no correction-caused regression found.

</details>

<details>
<summary><code>R2</code> — <code>restoredLoopPrevOutputs.size > 0</code> guard has no behavioral effect</summary>

**Impact:** None functionally — flagged for clarity/proportionality only, so a future reader wouldn't mistake the guard for load-bearing.

**Correction (`3e2639da`):** `packages/workflows/src/dag-executor.ts:4313-4315` adds a comment directly above the guard: "Kept as an empty-map guard for clarity only — `substituteLoopPrevRefs` reads via `loopPrevOutputs?.get(id)`, so an empty Map and undefined are indistinguishable to every consumer; this has no behavioral effect either way." The guard line itself is byte-for-byte unchanged.

**Verification:** `prp-core:code-simplifier` confirmed the comment is accurate and the guard's behavior is untouched — a comment-only fix, exactly the accepted disposition from the prior review.

**Found by:** `prp-core:code-simplifier` (original); re-verified by the same scope.

**Disposition:** FIXED — comment added, no behavioral change intended or introduced.

</details>

<details>
<summary><code>R3</code> — New oversized-output test's scaffolding could be shorter</summary>

**Impact:** None on correctness — declined as a non-blocking, optional cleanup, and the correction commit does not touch this test at all.

**Verification:** `prp-core:code-simplifier` confirmed via diff (`git diff 34f3479d..3e2639da -- dag-executor.test.ts`) that the oversized-output test is byte-for-byte untouched — the entire +104 diff lands before it, all belonging to the new `R1` regression test. Independently re-assessed the decline on the merits: the test's first phase substantially overlaps the adjacent in-process spill test, but its second phase exercises something that test doesn't — a genuinely separate `executeDagWorkflow` call fed real disk-round-tripped bytes through `priorCompletedNodes`, proving the spill/rehydrate path survives an actual pause/resume boundary. Concurs the ~140-line cost is proportionate to that distinct guarantee and that this remains below the bar for blocking.

**Required outcome:** None — optional, and the original review's own framing ("not worth a standalone change") still holds under independent re-assessment.

**Found by:** `prp-core:code-simplifier` (original); re-confirmed by the same scope.

**Disposition:** DECLINED — non-actionable optional cleanup; independently re-assessed as reasonable, not merely accepted on the implementer's word.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` (`prp-core:code-reviewer`) | R1 verified FIXED; no new finding |
| `seams` (`prp-core:seam-analyzer`) | R1 verified FIXED; no new finding |
| `simplify` (`prp-core:code-simplifier`) | R2 verified FIXED; R3 disposition re-confirmed DECLINED; no new finding |

Scope selection carried forward unchanged from the original review — every scope that owned a finding being verified was retained, and no scope beyond those three had an affected finding to re-check. `types` remains unselected: the correction adds no new type, schema, constructor, factory, public signature, state variant, or compiler escape hatch — it re-derives an existing field using an existing helper and an existing type-assertion pattern already present twice more in the same function.

#### Validation

|---|---|---|
| `cd packages/workflows && bun test src/dag-executor.test.ts -t "LOOP_PREV"` | PASS | 15 pass, 0 fail (was 14) — includes the new R1 regression test |
| `cd packages/workflows && bun test src/dag-executor.test.ts` | PASS | 609 pass, 0 fail (was 608), 1746 expect() calls |
| `bun run type-check` | PASS | Clean across all 10 packages (paths, git, providers, isolation, workflows, core, adapters, server, cli, docs-web) |
| `bun run lint` | PASS | `eslint . --cache` — zero warnings |

All four commands re-run in this review's own checkout (`.worktrees/review-pr-2750`, pinned to `3e2639da6495723a318f154d7324c8e29870fe51`) after `bun install`. `bun run validate`'s remaining steps (format, bundled/schema/capability-matrix checks) were not re-run — the correction touches only executor logic and a test file, neither of which those checks cover, and the prior review's full `bun run validate` pass against the unmodified base already covers them; format is additionally implied clean by lint/type-check passing on files this correction didn't restructure.

</details>

PR #2754, opened 2026-08-23
fix(workflows): loop_group warns when its pause escalation can't reach a nested interactive loop

### LINKED ISSUE #2753: bug(workflows): loop_group pause escalation doesn't propagate through a nested loop_group (pre-existing gap, inherited by #2707 step 3)
## Problem

PR #2749 (#2707 step 3) fixes pause escalation for a `loop_group` whose body-terminal sink is a native `approval:` gate — the gate's pause now correctly stops the group instead of the pre-existing "paused tolerated, loop barrels through" bug. That fix does not propagate through a nested `loop_group`.

`findLoopGroupTerminalGate` (`packages/workflows/src/dag-executor.ts:4165`) only recognizes a *gate node* (`isGateNode`) as an escalatable terminal sink. If a `loop_group`'s body-terminal sink is itself another `loop_group` (nested loops are explicitly supported — schema comment at `packages/workflows/src/schemas/dag-node.ts:434-435`, and `enclosingLoopGroupPath`/`stepNamePrefix` composition exists specifically for this), and that inner group's own body-terminal gate pauses:

- The inner group's escalation correctly rewrites `ApprovalContext.nodeId` to its own bare id and returns `{state: 'completed', ...}` — there is no `'paused'` variant in `NodeOutput` (`packages/workflows/src/schemas/workflow-run.ts:102-128`), so completion is the only signal available.
- The OUTER group's `findLoopGroupTerminalGate` on its own body returns `undefined`, because its sink is a `loop_group` node, not a `kind: 'gate'` node.
- The outer level therefore has no way to distinguish "inner group genuinely completed" from "inner group is actually paused waiting on a human" — the exact "paused tolerated" bug this whole step exists to fix recurs one level up, unfixed. The outer loop may re-invoke the inner group again inside the same synchronous call, against a run whose `workflowRun.metadata.approval` already reflects the inner group's rewritten context, producing confusing re-entry rather than a clean stop.

## Why this isn't a regression from PR #2749

This is not new: the pre-existing legacy `interactive_loop` mechanism (`nodeId: node.id` at `dag-executor.ts:5048`) has the identical bare-id limitation for nested groups — PR #2749's own docstring at `dag-executor.ts:4696` explicitly says the escalation "mirrors exactly how the interactive_loop gate below already works." #2749 doesn't make nested-loop-with-interactive-gate worse; it just doesn't extend the fix to that case. No test exercises a `loop_group`-nested-in-`loop_group` with an interactive gate at either the loader-warning level or the runtime level, in either the legacy mechanism or #2749's new one.

## Desired outcome

Either:
1. Extend `findLoopGroupTerminalGate`/the escalation mechanism to recurse through a nested `loop_group` sink, propagating a pause signal up through however many nesting levels exist — likely needs a real `'paused'` (or equivalent) signal distinguishable from `'completed'` in the return shape, since `NodeOutput` doesn't currently carry that distinction; or
2. A load-time placement rule (mirroring #2749's own terminal-sink warning) flagging a `loop_group` whose body-terminal sink is itself an interactive-class `loop_group`, so the gap is at least visible to authors rather than silently mis-behaving at runtime.

## Context

Found during independent review of PR #2749 (#2707 step 3, the loop_group pause-escalation mechanism). Not empirically reproduced (would need a constructed nested-loop-with-gate YAML run end to end) — grounded in direct code tracing across `dag-executor.ts` and the relevant schema comments. Not blocking #2749's merge; the primary single-level gate-terminated `loop_group` pattern #2749 targets is correct and unaffected.

Wirasm:
## Problem and outcome

PR #2749 (#2707 step 3) fixed pause escalation for a `loop_group` whose body-terminal sink is a native `gate:` node — the gate's pause now correctly stops the enclosing loop instead of the loop tolerating the pause and barreling into another iteration. That fix doesn't reach a nested case: a `loop_group` whose body-terminal sink is itself another interactive `loop_group` (or a legacy interactive `loop:` node) has no way to escalate that inner pause, so the exact bug #2749 fixed recurs one level down, unfixed.

- **Outcome:** Loading a workflow with this nesting pattern now produces a `parseWarnings` entry naming the outer `loop_group` and explaining that a pause inside its terminal-sink loop won't stop it. The workflow still loads and runs — this is advisory guidance, matching every other check in the function it extends.
- **Invariant:** No workflow that loads successfully today is rejected by this change. The warning fires only when a `loop_group`'s sole body-terminal sink is itself an interactive-class node whose own pause can't escalate.
- **Scope boundary:** Load-time only. `dag-executor.ts` and `NodeOutput` are untouched — no runtime pause-propagation mechanism was built.
- **Root cause:** `findLoopGroupTerminalGate` (`packages/workflows/src/dag-executor.ts:4165`) only recognizes a bare `gate:` node as an escalatable terminal sink. When the sink is instead a nested `loop_group`/interactive `loop`, that inner node's own pause escalation only rewrites its own `ApprovalContext` — it never touches `workflow_runs.status`, which stays `'paused'`. The outer group tolerates that status (by design, for sibling-gate cases), can't find its own terminal gate, and treats the inner node's phantom `{state:'completed'}` as a real iteration result — then loops again, re-invoking the inner node from a stale in-memory snapshot that doesn't see its own pending pause.

## Review guidance

- **Feedback requested:** Whether the load-time-warning direction (vs. building real runtime pause propagation) is the right call for this codebase's current maturity, and whether the recursive predicate's scope (covering both nested `loop_group` and the legacy interactive `loop:` node) is correctly bounded.
- **Start here:** `packages/workflows/src/loader.ts:374-395` — the new `isUnescalatableInteractiveSink` predicate; everything else follows from it.
- **Review order:** `loader.ts:374-395` (predicate) → `loader.ts:549-570` (call site) → `loader.test.ts:6768-6803` (existing test now asserting 4 warnings instead of 3 — proof the gap was real and untested) → the three new test cases below it.
- **Lower-attention areas:** None — this is a small, self-contained change.
- **Known risk or uncertainty:** None identified. The recursion has no depth cap, matching the existing unbounded recursion this function already does when walking nested `loop_group` bodies.

## Solution

Extends the existing `collectGateAndLoopDeprecationWarnings` in `packages/workflows/src/loader.ts`, which already computes each `loop_group`'s body-terminal sink and warns when a `gate:` node is misplaced there (#2707 step 3a). Adds one recursive predicate, `isUnescalatableInteractiveSink`: true for a `gate:` node, a `loop:` node with `interactive: true`, or a `loop_group:` node that is itself `interactive: true` or whose own sole terminal sink recursively satisfies the same predicate (any nesting depth). The existing block gets one new branch: when an outer group's sole body-terminal sink is a `loop`/`loop_group` node (not a bare gate — that's already correctly handled) satisfying the predicate, push a warning naming the outer group.

This is deliberately the smaller of the two options the source issue names. `NodeOutput` (`packages/workflows/src/schemas/workflow-run.ts:102-128`) has no `'paused'` state — pause is modeled everywhere in this codebase as `workflow_runs.status`, never a per-node output. Building real propagation would mean adding that state and updating every `.state` switch/if across the executor, resume snapshotting, and any API/console consumer — a lot of new surface on a mechanism (#2749) that just shipped. #2749 itself already declined full generality for a similar edge case (a gate that's mid-body or co-terminal with another sink) and documented it as a known limitation instead of building it out (`dag-executor.ts:4153-4164`) — this PR applies the same posture one level deeper.

## Validation

- `cd packages/workflows && bun test src/loader.test.ts` — passed (341 pass, 0 fail, 924 expect() calls) — proves the warning fires correctly at 1, 2, and 3+ nesting levels and does not false-positive on a bare gate sink or a non-interactive nested sink.
- `bun run validate` (full repo gate: bundled/schema checks, type-check across all 13 packages, lint, format, and the complete test suite) — passed, exit 0.
- **Not verified:** No live end-to-end runtime repro (an actual paused nested-loop workflow run) — the fix is load-time only with zero runtime behavior change, so a runtime repro would only re-prove the pre-existing bug rather than verify what this PR does. The causal chain is established by direct code tracing, cited above and in the linked issue/plan.

## Links

- Closes #2753
- Plan: https://github.com/coleam00/Archon/issues/2753#issuecomment-5383413132

## Summary by CodeRabbit

* **Bug Fixes**
  * Added warnings for workflow configurations where interactive loops or nested loop groups cannot properly escalate pauses.
  * Improved detection of terminal nested loop groups, including gate-terminated scenarios.
  * Extended warning propagation across deeply nested loop groups.
  * Avoided unnecessary warnings for non-interactive nested groups.

### DIFF
--- a/packages/workflows/src/loader.test.ts
+++ b/packages/workflows/src/loader.test.ts
@@ -6766,10 +6766,14 @@ nodes:
     });
 
     it('warns on a gate/interactive loop_group nested inside a loop_group body', async () => {
-      // Exercises the recursion into loop_group.nodes — the outer group is
-      // clean, only the inner body nodes carry deprecated fields. 'inner-gate' also
-      // has a dependent ('inner-loop'), so it additionally trips the terminal-sink
-      // placement warning below — three warnings on one gate/loop pair, not two.
+      // Exercises the recursion into loop_group.nodes — the outer group's own
+      // body-terminal sink is 'inner-loop' (a plain interactive `loop:` node,
+      // since 'inner-gate' has a dependent and so is NOT the sole terminal
+      // sink). That makes four warnings: the two deprecation notices on
+      // 'inner-gate'/'inner-loop', the terminal-sink placement warning on
+      // 'inner-gate', and — #2753 — 'outer' itself, whose only escalatable
+      // sink is an interactive loop and therefore can't stop on a pause inside
+      // it.
       const pw = await warningsFor([
         'name: test',
         'description: test',
@@ -6794,12 +6798,94 @@ nodes:
         '            gate_message: continue?',
         '          depends_on: [inner-gate]',
       ]);
-      expect(pw).toHaveLength(3);
+      expect(pw).toHaveLength(4);
       expect(pw.some(w => w.includes("Node 'inner-gate'") && w.includes('on_reject'))).toBe(true);
       expect(pw.some(w => w.includes("Node 'inner-loop'") && w.includes('interactive'))).toBe(true);
       expect(pw.some(w => w.includes("Node 'inner-gate'") && w.includes('terminal sink'))).toBe(
         true
       );
+      expect(
+        pw.some(w => w.includes("Node 'outer'") && w.includes('inner-loop') && w.includes('#2753'))
+      ).toBe(true);
+    });
+
+    it('warns on a loop_group whose sole terminal sink is a gate-terminated nested loop_group (#2753)', async () => {
+      const pw = await warningsFor([
+        'name: test',
+        'description: test',
+        'interactive: true',
+        'nodes:',
+        '  - id: outer',
+        '    loop_group:',
+        '      until_bash: "exit 0"',
+        '      max_iterations: 2',
+        '      nodes:',
+        '        - id: inner',
+        '          loop_group:',
+        '            until_bash: test "$review.output.decision" = approve',
+        '            max_iterations: 3',
+        '            nodes:',
+        '              - id: review',
+        '                approval:',
+        '                  message: ok?',
+      ]);
+      expect(pw).toHaveLength(1);
+      expect(pw[0]).toContain("Node 'outer'");
+      expect(pw[0]).toContain("'inner'");
+      expect(pw[0]).toContain('#2753');
+    });
+
+    it('warns at every level of a 3-deep nested loop_group chain (#2753)', async () => {
+      const pw = await warningsFor([
+        'name: test',
+        'description: test',
+        'interactive: true',
+        'nodes:',
+        '  - id: grandparent',
+        '    loop_group:',
+        '      until_bash: "exit 0"',
+        '      max_iterations: 2',
+        '      nodes:',
+        '        - id: middle',
+        '          loop_group:',
+        '            until_bash: "exit 0"',
+        '            max_iterations: 2',
+        '            nodes:',
+        '              - id: inner',
+        '                loop_group:',
+        '                  until_bash: test "$review.output.decision" = approve',
+        '                  max_iterations: 3',
+        '                  nodes:',
+        '                    - id: review',
+        '                      approval:',
+        '                        message: ok?',
+      ]);
+      expect(pw).toHaveLength(2);
+      expect(pw.every(w => w.includes('#2753'))).toBe(true);
+      expect(pw.some(w => w.includes("Node 'grandparent'") && w.includes("'middle'"))).toBe(true);
+      expect(pw.some(w => w.includes("Node 'middle'") && w.includes("'inner'"))).toBe(true);
+    });
+
+    it('does not warn when a nested loop_group has nothing interactive to escalate (#2753)', async () => {
+      const pw = await warningsFor([
+        'name: test',
+        'description: test',
+        'interactive: true',
+        'nodes:',
+        '  - id: outer',
+        '    loop_group:',
+        '      until_bash: "exit 0"',
+        '      max_iterations: 2',
+        '      nodes:',
+        '        - id: inner',
+        '          loop_group:',
+        '            until_bash: "exit 0"',
+        '            max_iterations: 3',
+        '            nodes:',
+        '              - id: work',
+        '                prompt: do work',
+      ]);
+      expect(pw).toEqual([]);
     });
 
     it('warns on a gate node with a dependent inside a loop_group body (not a terminal sink)', async () => {
--- a/packages/workflows/src/loader.ts
+++ b/packages/workflows/src/loader.ts
@@ -378,10 +378,14 @@ function collectUnknownNodeKeys(raw: unknown, id: string, label: string, warning
  * enclosing loop_group, but that escalation is bounded to that one level) or
  * transitively (a `loop:` node with `interactive: true`, the legacy mechanism,
  * or a `loop_group:` node that is itself `interactive: true` or whose own sole
- * terminal sink recurses into this same case, at any nesting depth). Mirrors
- * `findLoopGroupTerminalGate`'s doc comment (dag-executor.ts:4153-4164): the
- * runtime has no unambiguous way to escalate a pause through a sink that isn't
- * a bare gate, so this only makes that gap visible at load time.
+ * terminal sink recurses into this same case, following a chain of well-formed
+ * sole-terminal-sink nesting to any depth). A gate that is mid-body or
+ * co-terminal with another sink breaks the chain here too — the placement
+ * check in `collectGateAndLoopDeprecationWarnings` below already warns about
+ * that misplacement on its own. Mirrors `findLoopGroupTerminalGate`'s doc comment
+ * (dag-executor.ts:4153-4164): the runtime has no unambiguous way to escalate
+ * a pause through a sink that isn't a bare gate, so this only makes that gap
+ * visible at load time.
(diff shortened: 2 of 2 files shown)


Wirasm:
## Ready to merge

A fresh full review of `b442adfb20..5fd6ec6866` found no open defect in the directly authored nested-loop warning delivered by this PR. The three requested reviewers independently checked the complete current-head diff rather than relying on the earlier verdict. The supported `include:` composition route still bypasses the warning, but the approved plan explicitly excludes unresolved include directives from this pre-expansion check and the gap remains correctly separated in open issue [#2756](https://github.com/coleam00/Archon/issues/2756). The two prior wording findings remain fixed, and the complete repository gate passes at the reviewed head.

**0 blocking · 1 non-blocking (tracked)**

**Validation:** `bun run validate` passed after installing the detached worktree's locked dependencies; focused loader/include tests passed with 443 tests, 0 failures.

**Resolved:** 2 · **Tracked follow-ups:** [#2756](https://github.com/coleam00/Archon/issues/2756)

### Findings

|---|---|---|---|
| `R1` | Important | `loop_group`'s pause-escalation warning cannot see a terminal sink composed via `include:` | TRACKED FOLLOW-UP |
| `R2` | Suggestion | The predicate comment overstated the nesting shapes it follows | FIXED |
| `R3` | Suggestion | The warning text contained an awkward doubled apostrophe and repeated node id | FIXED |

<details>
<summary><code>R1</code> — include-composed terminal sink bypasses the warning</summary>

**Impact:** A supported `loop_group` body whose sole authored node is an `include:` can expand to a gate-terminated nested loop without receiving this warning. The runtime then sees the same unsupported nested pause topology and can start another outer iteration while the human response is pending.

**Evidence:** `packages/workflows/src/loader.ts:390` returns `false` for an unresolved `IncludeDirective`, and `packages/workflows/src/loader.ts:638` runs the warning pass before composition. `packages/workflows/src/workflow-discovery.ts:648` expands includes, then pairs the expanded workflow with the pre-expansion warnings. `packages/workflows/src/schemas/dag-node.ts:445` and `packages/workflows/src/include-expander.ts:1145` prove that includes inside `loop_group` bodies are supported and expanded recursively. The seam reviewer reproduced an expanded nested-loop sink with no warning at the exact reviewed head.

**Required outcome:** Run the sink-shape warning against the expanded `DagNode[]` before constructing `WorkflowWithSource`, merging those warnings with the parse-time warnings. A resolved-workflow phase type could make this authored-versus-expanded boundary explicit, but is not required if the check is placed correctly.

**Found by:** `seam-analyzer`; corroborated by `code-reviewer`

**Disposition:** TRACKED FOLLOW-UP. [Issue #2756](https://github.com/coleam00/Archon/issues/2756) is open and matches the reproduced seam. The approved #2753 plan explicitly makes unresolved `IncludeDirective` return false and scopes this PR's warning to directly authored loop/loop-group sinks, so post-expansion validation is separate work rather than a missing part of this PR's accepted outcome.

</details>

<details>
<summary><code>R2</code> — predicate comment overstated supported nesting shapes</summary>

**Impact:** The original comment implied every nesting shape was followed even though recursion only traverses a chain with one terminal sink at each level.

**Evidence:** `packages/workflows/src/loader.ts:374` now limits the claim to a well-formed sole-terminal-sink chain and explains that the existing placement warning owns mid-body and co-terminal gates.

**Required outcome:** No correction remains.

**Found by:** `code-reviewer`

**Disposition:** FIXED. All three fresh reviewers agreed the current comment matches the predicate and its sibling placement check.

</details>

<details>
<summary><code>R3</code> — warning text was malformed</summary>

**Impact:** The original warning rendered a doubled apostrophe and redundantly repeated the node id.

**Evidence:** `packages/workflows/src/loader.ts:561` now says that the pause does not escalate to stop "this loop_group's own iteration," with no malformed interpolation.

**Required outcome:** No correction remains.

**Found by:** `code-reviewer`, `code-simplifier`

**Disposition:** FIXED. The fresh code and simplification reviewers found no remaining wording or structural issue at the current head.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` (`code-reviewer`) | No additional findings; corroborated R1's separate scope and verified R2/R3 remain fixed |
| `seams` (`seam-analyzer`) | R1; direct-authored topology, bare-gate exclusion, and warning transport were otherwise clean |
| `simplify` (`code-simplifier`) | No additional findings; the module-local recursive predicate is the smallest coherent structure for arbitrary supported nesting depth |

#### Validation

|---|---|---|
| `bun run validate` before dependency install | FAIL | Detached checkout lacked its own workspace dependencies: `Cannot find module '@hono/zod-openapi'`; this was an environment precondition, not PR behavior |
| `bun install --frozen-lockfile` | PASS | Installed 2,599 packages from the lockfile in the detached review checkout |
| `bun run validate` | PASS | Bundled/schema/capability checks, all-package type-check, lint with zero warnings, formatting, installer tests, isolated package tests, and script tests completed with exit 0 |
| `cd packages/workflows && bun test src/loader.test.ts src/include-expander.test.ts` | PASS | 443 pass, 0 fail, 1,213 assertions |
| `git diff --check` | PASS | No whitespace errors in the PR diff |
| `cd packages/workflows && bun x tsc --noEmit` | PASS | Workflows package type-check completed with exit 0 |
| `bun x eslint packages/workflows/src/loader.ts --max-warnings 0` | PASS | 0 errors, 0 warnings |

</details>

Issue #2064, opened 2026-07-08
Pi provider: extension-registered models (`pi-cursor`) fail on 2nd+ DAG node in same workflow run


pondycrane:
# Pi provider: extension-registered models (`pi-cursor`) fail on 2nd+ DAG node in same workflow run

Issue to raise in [Archon](https://github.com/coleam00/archon).

## Summary

- **What broke:** Multi-node Archon workflows using the Pi assistant with extension-provided models (`@schultzp2020/pi-cursor`, `cursor/*`) fail on the second and later DAG nodes with `Pi model not found`, even though the first node succeeds and single-node workflows work.
- **When it started (if known):** Observed on Archon `0.80.3` with `pi-cursor@0.5.0` (2026-07-08). Likely affects any version where Pi extension models are resolved lazily per session but not carried across DAG nodes.
- **Severity:** `major` (could affect all pi extension scope during 2+ nodes)

## Steps to Reproduce

1. Install Archon (`0.80.3`), Pi CLI, and `npm:@schultzp2020/pi-cursor` (`pi install npm:@schultzp2020/pi-cursor -l`). Authenticate with Cursor (`/login` → Cursor).
2. Configure `.archon/config.yaml` with `assistant: pi`, `enableExtensions: true`, and tier `small` → `cursor/gpt-5.4-nano`. Ensure `.pi/settings.json` lists the pi-cursor package.
3. Add workflow `archon-test-pi-cursor-sequential` (two Pi nodes, `second` depends on `first`)
4. Run: `archon workflow run archon-test-pi-cursor-sequential --no-worktree --cwd /path/to/repo`
5. (Control) Run single-node `archon-test-pi-cursor` — it passes.

Minimal config:

```yaml
# .archon/config.yaml
assistant: pi
assistants:
  pi:
    model: cursor/gpt-5.4-nano
    enableExtensions: true
tiers:
  small:
    provider: pi
    model: cursor/gpt-5.4-nano
    effort: low
```

```json
// .pi/settings.json
{
  "defaultProvider": "cursor",
  "defaultModel": "gpt-5.4-nano",
  "enabledModels": ["cursor/*"],
  "packages": ["npm:@schultzp2020/pi-cursor"]
}
```

## Expected vs Actual

- **Expected:** Both `first` and `second` nodes complete; `cursor/gpt-5.4-nano` resolves via pi-cursor on every node in the run.
[...]

PR #2757, opened 2026-08-23
Bump pi-coding-agent to ^0.84.2 to close the undici 8.5.0 advisory

### LINKED ISSUE #2678: fix(security): clear vulnerable undici 8.5.0 from under pi-coding-agent
## Problem

`undici` is transitive-only — nothing in the repository declares it. After a full `bun.lock` regeneration the tree carries four copies, and the vulnerable one has **moved**:

```
undici              8.6.0  (hoisted, clean)
@dotenvx/dotenvx/undici     7.29.0  (clean)
@earendil-works/pi-coding-agent/undici   8.5.0  <-- vulnerable
unifont/undici                           8.5.0  <-- vulnerable
```

Vulnerable range `>=8.0.0 <8.9.0`, patched at **8.9.0**. Six advisories, one high:

| Severity | Advisory |
|---|---|
| high | [GHSA-4cwx-7wf7-3272](https://github.com/advisories/GHSA-4cwx-7wf7-3272) — cross-user information disclosure and parse-time crash via degenerate private cache directives |
| moderate | [GHSA-8xcm-r25x-g524](https://github.com/advisories/GHSA-8xcm-r25x-g524) — downstream response desynchronization via the retry interceptor |
| moderate | [GHSA-jr45-8vmc-qm54](https://github.com/advisories/GHSA-jr45-8vmc-qm54) — cross-user information disclosure via whitespace around equals in `Cache-Control` |
| moderate | [GHSA-m8rv-5g2x-5cg5](https://github.com/advisories/GHSA-m8rv-5g2x-5cg5) — CRLF injection via a blob-like body `type` property |
| moderate | [GHSA-v3r7-h72x-cjcm](https://github.com/advisories/GHSA-v3r7-h72x-cjcm) — cookie attribute injection via unsanitized domain and unparsed `setCookie` fields |
| moderate | [GHSA-p88m-4jfj-68fv](https://github.com/advisories/GHSA-p88m-4jfj-68fv) — HTTP header injection via `Set-Cookie` percent-decoding |

## This supersedes the closed #1485 and #1486 framing

Both of those described a **6.x** exposure reached through `discord.js`. That is no longer where the risk is:

- The `discord.js` path resolves clean once the lockfile is refreshed. #1485's `lodash` half is also lockfile-only now.
- The live exposure is `@earendil-works/pi-coding-agent` and `unifont` pinning `undici` at 8.5.0.

`unifont` arrives through `@archon/docs-web` (Astro font loading), so that copy is build-time only. The `pi-coding-agent` copy is not — it is in the Pi provider's runtime HTTP path.

## What makes this risky to bump

Nothing declares `undici`, so the only lever from this repository is a root `override`. That is exactly what went wrong in #2253: a flat `undici` override dragged Pi from its pinned 8.5.0 down two majors, to a version *older* than the patched one. An override here must be `^8.9.0` — narrow enough to stay inside the 8.x consumers, and verified not to touch the `@dotenvx/dotenvx` 7.x copy.

The safer fix is an upstream bump: check whether a newer `@earendil-works/pi-coding-agent` already carries `undici >= 8.9.0`. Bumping the Pi SDK is itself gated — AGENTS.md requires `bun run generate:pi-vendor-map` after any `@earendil-works/pi-ai` upgrade, and `bun run validate` runs `check:pi-vendor-map`.

## Verification

- `bun install` must leave `@dotenvx/dotenvx/undici` at 7.x
- `bun audit` drops all six findings
- Pi provider smoke: run a workflow node with `provider: pi` and confirm the HTTP path works
- `bun run validate`

## Context

Found while working #2666. This is one of the five packages that survive a full `bun.lock` regeneration, so Renovate's scheduled lock-file maintenance will **not** fix it.

Wirasm:
## Problem and outcome

The Pi provider's runtime HTTP path was loading `undici@8.5.0` through `@earendil-works/pi-coding-agent@0.80.6`. `undici@8.5.0` falls inside the `>=8.0.0 <8.9.0` advisory window — six upstream advisories in #2678, one high (GHSA-4cwx-7wf7-3272 — cross-user info disclosure plus parse-time crash). Renovate cannot fix this on its own: every published `0.80.x → 0.83.x` of `@earendil-works/pi-coding-agent` declares `"undici": "8.5.0"` exactly, and `^0.80.6` does not admit `0.84.x`, so a lockfile-only refresh never moves the pin.

- **Outcome:** a single `undici@8.9.0` is resolved under `@earendil-works/pi-coding-agent@0.84.2`, all six undici advisories in the issue stop appearing for the Pi consumer, and the four call sites that the new SDK surface breaks are adapted to the new SDK contract: `provider.ts` and `model-catalog.ts` move to `ModelRuntime` + `ModelRegistry`; `request-auth.ts` becomes a per-call `models.json` builder for custom (credentialless) provider env substitution; `oauth.ts` becomes a local `OAuthProviderInterface` adapter around the new pi-ai 0.84 `OAuthAuth` singletons. `bun audit` no longer flags the pi-coding-agent transitive.
- **Invariant:** `discord.js`'s exact-pinned `undici@6.21.3` stays untouched (out of the advisory range; `renovate.json`'s `packageRules[0].description` forbids the bump). The Pi provider's per-user `ARCHON_PI_AUTH_PATH` resolution is preserved end-to-end (now passed via `ModelRuntime.create({ authPath })` instead of `AuthStorage.create(authPath)`). The "no protected env var reaches the per-call request" contract is preserved for custom (credentialless) providers via `buildCustomProviderModelsPath`: `${VAR}` references in the user's `models.json` are substituted against `requestOptions.env` (using a parser that mirrors Pi's `resolveConfigValue`, with no `process.env` fallback), and protected keys are replaced with a structurally-valid placeholder the SDK cannot resolve (`${__ARCHON_BLOCKED_${VAR}__}`) — the literal protected value never appears in the per-call file, and the SDK surfaces a host-environment-independent "no value for env var" error at request time.
- **Scope boundary:** no `undici` override is added at any level — the bump moves only the parent that pins the vulnerable transitive. The Pi-OAuth helpers' deep subpaths are loaded lazily through `createRequire` against the SDK's on-disk `package.json` because pi-ai 0.84.0 reorganised the public surface; the legacy `OAuthProviderInterface` shape that `@archon/core` consumers depend on stays unchanged at the package boundary.
- **Root cause:** `@earendil-works/pi-coding-agent@0.80.6` exactly-pins `undici@8.5.0`, which lands inside the upstream advisory range `>=8.0.0 <8.9.0`; the manifest range `^0.80.6` excludes every 0.84.x release that ships the patched undici.

## Review guidance

- **Feedback requested:** correctness of the `ModelRuntime`/`ModelRegistry` migration; the per-call `models.json` substitution approach for custom (credentialless) providers (`buildCustomProviderModelsPath`), especially how it interacts with providers that already have a stored credential in `auth.json` (those resolve via the SDK's default lookup — the per-call `modelsPath` is only written when there is a `${VAR}` template to substitute against the request env); confirmation that the lazy `createRequire`-backed OAuth loader is acceptable given the deep-import restriction; any concern with `oauth-bridge.ts:298`'s intentional cast on the `loginPromise` union.
- **Start here:** `packages/providers/src/community/pi/provider.ts:384-410` — the `AuthStorage.create` + `ModelRegistry.create` pair becomes `await ModelRuntime.create({ authPath })` + `new ModelRegistry(modelRuntime)`. This is load-bearing because every later step (`setRuntimeApiKey`, `getAuth`, `createAgentSession({ modelRuntime })`, `getModels`) flows from the runtime handle that this block creates.
- **Review order:** `packages/providers/package.json:35-36` (range widening) → `bun.lock` (confirm a single `undici@8.9.0` under `@earendil-works/pi-coding-agent/undici`) → `packages/providers/src/community/pi/provider.ts:384-410` (the runtime handle + the per-call `modelsPath` for custom providers + the `finally` cleanup) → `packages/providers/src/community/pi/request-auth.ts` (the per-call `models.json` builder `buildCustomProviderModelsPath` — mirrors Pi's `resolveConfigValue` parser, substitutes `${VAR}` against `requestEnv`, replaces protected keys with `${__ARCHON_BLOCKED_${VAR}__}`) → `packages/providers/src/oauth.ts` (the OAuth adapter, deviated from the assessment's scope).
- **Lower-attention areas:** `bun.lock` is mechanical — the substantive change is `undici 8.5.0 → 8.9.0` under `@earendil-works/pi-coding-agent`, plus the pi-ai 0.84.x dependency graph reshuffle: caret-respecting bumps (`typebox 1.1.38 → 1.3.7`, `openai 6.26.0 → 6.40.0`), the drop of `@mistralai/mistralai@2.2.6` (and its `@opentelemetry/semantic-conventions` transitive), and the addition of three new pi-ai-internal packages (`@earendil-works/pi-client`, `@earendil-works/pi-protocol`, `@earendil-works/pi-telemetry`) plus the `grok-mermaid` dependency pi-coding-agent now ships; `pi-vendor-map.generated.ts` is regenerated by `bun run generate:pi-vendor-map` and lists five new backends (`baseten`, `qwen-token-plan`, `qwen-token-plan-cn`, `qwen-token-plan-individual`, `radius`); `provider.test.ts` and `request-auth.test.ts` updates are mock-surface rewrites tied to the new SDK shape (the latter is a substantive rewrite — the test target moved from `withCustomProviderRequestEnv` to `buildCustomProviderModelsPath`).
- **Known risk or uncertainty:** the live Pi-provider HTTP smoke test called out in #2678 is not exercised in this worktree — no live workflow node with `provider: pi` was runnable here. The migration's wiring is covered by the unit tests (which mock the SDK), but a CI run against a real Pi credential pair is the strongest end-to-end signal. An out-of-scope type slip in `packages/core/src/credentials/oauth-bridge.ts:317` (`prompt: any`) was surfaced by the broader type-check sweep but pre-existed this PR and is left for the next strict-cleanup pass.

## Solution

The pin only moves when the parent moves. `^0.80.6` does not satisfy `0.84.x`, so `packages/providers/package.json:35-36` widens the `pi-coding-agent` and `pi-ai` ranges in lockstep to `^0.84.2` — pi-ai has to widen too because AGENTS.md gates its upgrades on regenerating the vendor map, and the four Pi call sites that break on the new SDK surface have to be migrated. `bun install` resolves a single `undici@8.9.0` under `@earendil-works/pi-coding-agent/undici`; `discord.js`'s `undici@6.21.3` is left untouched because a flat override would drag it past its declared range.

The four call-site migrations follow the new SDK contract:

- `provider.ts` — `await piCodingAgent.ModelRuntime.create({ authPath })` instead of `AuthStorage.create` + `ModelRegistry.create(authStorage)`. `setRuntimeApiKey` is now async; `getApiKey` becomes `runtime.getAuth(providerId)` returning the new `AuthResult` shape; `createAgentSession` takes `modelRuntime` (the SDK builds its own internal registry from the runtime we pass).
- `model-catalog.ts` — `ModelRegistry.getAll()` becomes `ModelRuntime.getModels()` for the same merged snapshot.
- `request-auth.ts` — the pre-0.84 helper (`withCustomProviderRequestEnv`) is replaced entirely. pi-coding-agent 0.84 consults a fresh `new ModelRegistry(this._modelRuntime)` for session auth, so any wrapper around a caller-constructed registry was discarded inside `createAgentSession` — the wrapper plugged into a surface the SDK no longer reads, and credentialless custom providers (`apiKey: '$MYGW_API_KEY'`) silently regressed because the SDK fell through to `process.env`, which Archon deliberately keeps empty (per-call secrets ride on `requestOptions.env`). The replacement is `buildCustomProviderModelsPath`: it writes a per-call `models.json` with `${VAR}` references substituted against `requestOptions.env` (using a parser that mirrors Pi's `resolveConfigValue` — same `$$`/`$!`/`${VAR}`/`$VAR` semantics, but with no `process.env` fallback) and returns its path so the caller can pass it as `modelsPath` to `ModelRuntime.create`. Pi's `ModelConfig.load` then reads the literal substituted values directly — no `${VAR}` substitution at request time, so the missing-fallback path cannot fire. Protected `${VAR}` references are substituted with a structurally-valid placeholder (`${__ARCHON_BLOCKED_${VAR}__}`) the SDK's parser recognises but cannot resolve (the placeholder name is provably absent from both `requestEnv` and `process.env`), so the literal protected value never appears in the per-call file. The file is written owner-only (`0o600`, with explicit `chmodSync` to defeat umask) into an owner-only directory (`0o700`), and the provider unlinks it in a `finally` block as soon as `ModelRuntime.create` resolves (whether `create()` succeeds or throws — a round-2 review found the file was previously leaked on long-running processes).
- `provider.test.ts` / `request-auth.test.ts` — `mock.module('@earendil-works/pi-coding-agent', ...)` factories rewritten for the new surface; the issue #2064 extension-replay test asserts that registrations still flow through the shared runtime.

The OAuth adapter in `packages/providers/src/oauth.ts` was rewritten beyond what the assessment scoped: pi-ai 0.84.0 dropped `getOAuthProvider`, `getOAuthApiKey`, and `OAuthProviderInterface` from its public surface — the implementations moved to deep subpaths (`dist/auth/oauth/anthropic`, `dist/auth/oauth/github-copilot`) that aren't in the package's `exports` field, so a static ESM import fails with "Cannot find module". `oauth.ts` now owns the legacy shapes locally; the new `OAuthAuth` singletons (`anthropicOAuth`, `githubCopilotOAuth`) are loaded lazily through `createRequire` against the SDK's on-disk `package.json` (which IS in `exports`), bypassing the deep-path restriction at runtime. `getOAuthApiKey` re-glues the new `OAuthAuth.toAuth()` + `refresh()` split; on the no-rotation path it echoes the input credential into `newCredentials` because the consumer in `user-provider-key-store.ts` casts it as `OAuthCredentials` and unconditionally reads the field.

`packages/core/src/credentials/oauth-bridge.ts:298` needed one explicit cast on the `loginPromise` union: pi-ai 0.84.0 made `OAuthCredential` a strict discriminated union with `type: "oauth"`, no longer structurally compatible with Archon-owned `OpenAiOAuthCredentials`. Both still feed `persistProviderOAuth` which accepts `Record<string, unknown>`, so the cast sits at the bridge boundary.

## Behavior change

| | Before | After |
|---|---|---|
| **Resolved `undici` for the Pi provider runtime HTTP path** | `8.5.0` (vulnerable, inside the `>=8.0.0 <8.9.0` advisory range) | `8.9.0` (patched) |
| **`bun audit` findings for the pi-coding-agent transitive** | six (one high: GHSA-4cwx-7wf7-3272; five moderate) | none for that consumer |
| **Resolved `undici` for `discord.js`'s HTTP path** | `6.21.3` | `6.21.3` (unchanged; out of advisory range; renovate forbid-path holds) |
| **Pi provider's auth/credential resolution** | `AuthStorage.create(authPath)` → `ModelRegistry.create(authStorage)` → `getApiKey` returns string | `await ModelRuntime.create({ authPath })` → `runtime.getAuth(providerId)` returns `AuthResult` shape (provider reads `.auth.apiKey`) |
| **Pi OAuth login callback surface** | direct SDK calls via `getOAuthProvider` / `getOAuthApiKey` | local adapter (`@archon/providers/oauth`) preserves the same `OAuthProviderInterface` shape for `@archon/core`; SDK singletons loaded lazily via `createRequire` |

## Architecture

```mermaid
flowchart LR
  subgraph Before
    A1[provider.ts sendQuery] --> B1[AuthStorage.create + ModelRegistry.create]
    B1 ==> C1[authStorage.getApiKey + modelRegistry.find]
    C1 --> D1[createAgentSession with authStorage + modelRegistry]
  end

  subgraph After
    A2[provider.ts sendQuery] --> B2[await ModelRuntime.create + new ModelRegistry]
    B2 ==> C2[runtime.getAuth + modelRegistry.find / runtime.getModels]
    C2 --> D2[createAgentSession with modelRuntime]
  end
```

### Changed seams

|---|---|---|
| `@archon/providers/src/community/pi/provider.ts` (Pi runtime init) | `AuthStorage.create` → `await ModelRuntime.create`; `ModelRegistry.create(authStorage)` → `new ModelRegistry(runtime)`; `authStorage.setRuntimeApiKey`/`getApiKey` → `runtime.setRuntimeApiKey`/`getAuth` | `packages/providers/src/community/pi/provider.ts:384-410`, unit tests |
| `@archon/providers/src/community/pi/model-catalog.ts` (catalog hint surface) | `ModelRegistry.getAll()` → `ModelRuntime.getModels()` | `packages/providers/src/community/pi/model-catalog.ts:46-47`, type-check |
| `@archon/providers/src/community/pi/request-auth.ts` (custom-provider env substitution) | Pre-0.84 `withCustomProviderRequestEnv` (which wrapped `AuthStorage`/`ModelRegistry` and used `Object.defineProperty` throwing getters for protected keys) replaced with `buildCustomProviderModelsPath`: writes a per-call `models.json` with `${VAR}` references substituted against `requestEnv`, replaced by `${__ARCHON_BLOCKED_${VAR}__}` placeholders for protected keys, and passed as `modelsPath` to `ModelRuntime.create`. The provider unlinks the file in a `finally` block after `create()` returns (the round-2 review's leak-surface fix). | `packages/providers/src/community/pi/request-auth.ts`, 21 unit cases + a 5-case `test.each` for protected keys, plus the real-SDK integration tests in `request-auth.integration.test.ts` (4 SDK-dependent cases against the live `ModelRuntime.create` and 1 file-mode permission case) |
| `@archon/providers/src/oauth.ts` (SDK boundary for Pi OAuth) | Local `OAuthProviderInterface` / `OAuthLoginCallbacks` / `getOAuthApiKey` shapes; SDK `OAuthAuth` singletons loaded lazily via `createRequire` against `import.meta.resolve('@earendil-works/pi-ai/package.json')` (the only subpath in `exports`) | `packages/providers/src/oauth.ts`, `packages/core/src/credentials/oauth-bridge.ts` |
| `@archon/core/src/credentials/oauth-bridge.ts:298` (Pi OAuth bridge) | One explicit cast at the `loginPromise` join to bridge strict `OAuthCredential` discriminated union with Archon-owned `OpenAiOAuthCredentials` | `packages/core/src/credentials/oauth-bridge.ts:298`, type-check |
| `@archon/providers/src/community/pi/pi-vendor-map.generated.ts` (regenerated) | Five new backends: `baseten`, `qwen-token-plan`, `qwen-token-plan-cn`, `qwen-token-plan-individual`, `radius` | `bun run generate:pi-vendor-map`, `check:pi-vendor-map` |
| `packages/providers/package.json` (manifest range) | `@earendil-works/pi-coding-agent` and `@earendil-works/pi-ai`: `^0.80.6 → ^0.84.2` | `packages/providers/package.json:35-36`, `bun install`, `bun pm why undici` |
| `bun.lock` (resolution graph) | `undici 8.5.0 → 8.9.0` under `@earendil-works/pi-coding-agent`; sibling transitive bumps follow pi-ai 0.84.x (caret-respecting, no downgrade) | `bun.lock`, `bun pm why undici` |

## Validation

- `bun run validate` — exit 0 across `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, `check:schema-upgrades`, `type-check` (all 11 workspaces exit 0), `lint --max-warnings 0`, `format:check`, `test:install`, and the full `test` suite (95+ tests pass, 0 fails). Proves the migration does not regress any code-side contract that the project's automated gates cover.
- `bun pm why undici` — single `undici@8.9.0` under `@earendil-works/pi-coding-agent@0.84.2` (was `8.5.0`, vulnerable); single `undici@6.21.3` under `discord.js@14.25.1` (unchanged, out of advisory range). Proves no flat-override collateral damage was introduced.
- `bun audit` — none of the six undici advisories from #2678 surface for the pi-coding-agent consumer anymore. The residual `undici <6.23.0` advisories only apply to the discord.js copy, which is out of scope per `renovate.json`.
- `bun install --frozen-lockfile` — exit 0, no churn. Proves the lockfile resolves cleanly without further mutation.
- `git diff bun.lock | grep -E "^[-+]" | grep -vE "(undici|@earendil|@-?[a-z-]+/)` manual triage of the lockfile — confirms every changed line sits inside pi-coding-agent / pi-ai's transitive tree or inside the unrelated caret-respecting bumps.
- **Not verified:** a live Pi-provider HTTP smoke test (a workflow node with `provider: pi` against real credentials) is not reproducible in this worktree — there is no running Archon instance with Pi credentials. The wiring is covered by `provider.test.ts` (which mocks `createAgentSession` against the new `modelRuntime` shape) and `request-auth.test.ts` (21 unit cases + a 5-case `test.each` for protected keys covering `${VAR}` resolution, the `$$`/`$!` escapes, malformed `${VAR` fallthrough, `PI_CODING_AGENT_DIR` override, and the per-call file's `0o600` mode bit). `request-auth.integration.test.ts` exercises the substitution end-to-end against the real pi-coding-agent SDK to confirm `ModelConfig.load` reads the literal substituted values without falling back to `process.env`, and that the protected-env placeholder is provably unresolvable even when the same variable is set in the host shell (round-2 regression coverage). The CI workflow against the merged branch — where Pi credentials can be injected — is the authoritative end-to-end gate.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No user-visible migration. The `pi-coding-agent` direct declaration widens to `^0.84.2`; pi-ai widens in lockstep because AGENTS.md gates it. No data migration, no schema change, no user config change. | `packages/providers/package.json:35-36`, `AGENTS.md` |
| Security / permissions / data | Closes all six undici advisories listed in #2678 for the Pi provider's runtime HTTP path. The discord.js `undici@6.21.3` copy stays outside the advisory range and is untouched. No new attack surface introduced by the migration. The protected-env contract for custom providers now closes at the SDK auth seam itself: `buildCustomProviderModelsPath` substitutes `${VAR}` references against `requestEnv` and replaces protected keys with a structurally-valid placeholder the SDK cannot resolve (`${__ARCHON_BLOCKED_${VAR}__}`), so the literal protected value never appears in the per-call file; the host shell cannot substitute the placeholder either, since no `process.env` value matches the `__ARCHON_BLOCKED_` prefix. The per-call file is owner-only (`0o600`, with explicit `chmodSync` to defeat umask) and is unlinked as soon as `ModelRuntime.create` returns. | `bun audit`, `bun pm why undici`, `renovate.json` |
| Rollout / rollback | Pure dependency bump plus contained call-site migration. Reverting `packages/providers/package.json:35-36` to `^0.80.6` and reverting the migration commits (or `git revert` of the two commits) restores the prior resolution graph; no schema, config, or data to back-fill. | `git log origin/dev..HEAD`, `packages/providers/package.json` |
| Observability | `bun run validate` is the local gate; CI runs the same script plus the Pi-provider workflow smoke against real credentials. A regression in the migration would surface as a `provider.test.ts` / `request-auth.test.ts` failure or a `bun run type-check` failure in `@archon/providers`. | `.github/workflows/`, `bun run validate` |
| Documentation / communication | No user-facing docs change. The Pi provider's behavior is unchanged from the user's perspective — same models, same auth paths, same OAuth flows. The README-level Pi docs already describe the `provider: pi` selector, which still routes through the same `PiProvider`. | `README.md`, no diffs |

## Links

- Closes #2678


Wirasm:
# Review report — PR #2757 (head SHA `ce4cf7a54300668c34793f47a21dcbb765dd9e2f`, round 4, light mode)

## Verdict

**Ready.** Both enabled lenses (code, seams) converged on the same verdict: the round-3 outstanding R1 (Important, FS-error framing gap) is genuinely closed at `ce4cf7a5`, no new findings were raised, and every prior finding (R1, S1, F1, F2, F3, S2) is closed/fixed. Adversarial verification (one-shot revert of the fix → test fails; fix restored → test passes) confirms the closure is real and not just renamed/relocated.

## Reviewed head SHA

`ce4cf7a54300668c34793f47a21dcbb765dd9e2f` (`archon/task-upkeep-1787444863473`, PR #2757 against `dev`). This is the cursor for the next round.

## Findings

None. Both lenses (code, seams) reported **zero new findings** for the delta. The single delta commit (`ce4cf7a5`) is the minimum correction the round-3 report asked for: move the `customProviderModelsPath` assignment inside the existing `try { … } catch { … } finally { … }` block so `mkdirSync`/`writeFileSync`/`chmodSync` errors from `buildCustomProviderModelsPath` are framed as `'Pi auth storage init failed: …'` instead of escaping raw.

### Rejected findings

None. No lens raised a finding whose evidence failed adversarial verification.

## Suggestions

None.

The carried-forward table from round 3:

| ID | Severity | One-line claim | Verdict in this round |
|---|---|---|---|
| **R1** (round-3 finding, the only outstanding one) | Important | `mkdirSync`/`writeFileSync`/`chmodSync` errors from `buildCustomProviderModelsPath` bypass the documented `'Pi auth storage init failed: …'` framing. In round-3 the call sat ABOVE the `try { … }` block at `provider.ts:395-401`, so `EACCES`/`ENOSPC` on the per-call tmpdir escaped raw — missing both the classified prefix and the `getLog().error(…, 'pi.auth_storage_init_failed')` log line. | **Fixed at `ce4cf7a5`.** `customProviderModelsPath` is now declared with `let customProviderModelsPath: string \| undefined;` at `provider.ts:399` (above the `try {`, so the `finally` can still reference it) and assigned INSIDE the `try {` block at `provider.ts:401-407` via `buildCustomProviderModelsPath({…})`. The existing `catch` at `:429-435` and `finally` at `:436-455` now both wrap `mkdirSync`/`writeFileSync`/`chmodSync` errors from `buildCustomProviderModelsPath`. The new test `custom provider: filesystem error from buildCustomProviderModelsPath is classified as Pi auth storage init failed` (`provider.test.ts:753-861`) reproduces the production failure shape against a real read-only tmpdir (chmod 0o555) and asserts (a) `error.message` contains `'Pi auth storage init failed'`, (b) matches `/EACCES\|permission denied/`, (c) contains `'~/.pi/agent/auth.json'`, (d) `mockLogger.error` was called with `{ piProvider: 'mygw' }` and `'pi.auth_storage_init_failed'`, (e) `mockModelRuntimeCreate` was NOT called. **End-to-end verified:** the new test passes against `ce4cf7a5` and FAILS against a one-shot revert that hoists the assignment back above the `try {` (raw `'EACCES: permission denied, mkdir …'` reaches the caller with no classification). `tsc --noEmit -p tsconfig.json` exits 0 (the `let` declaration at `:399` is in scope at the `finally` reference at `:448, :450`). F3/S2 cleanup tests at `provider.test.ts:631-686` (success) and `:688-751` (failure) continue to pass — the `finally`'s `if (customProviderModelsPath)` guard at `:448` correctly short-circuits when FS failure leaves the assignment at `undefined`. Full `bun test src/community/pi/provider.test.ts src/community/pi/request-auth.integration.test.ts` → 103 pass / 4 skip / 0 fail. |
| R1 (round-1, Critical) | Critical | `withCustomProviderRequestEnv` overrides a surface the 0.84.0 SDK no longer consults during session auth; the wrapper's per-request env thread and the protected-env contract silently regress. | **Closed at `f4bb97ef`.** R1.a (credentialless leg) closed at `1b27f646`; R1.b (protected-env leg) closed by S1's fix at `f4bb97ef`. Reclassified from "partially closed" to "closed". Carries forward as closed — no change in this round. |
| S1 (round-2, Critical) | Critical | The protected-env contract silently regressed when the host shell had the protected var (CI default state): the round-2 "leave the literal `${PROTECTED_VAR}` and let the SDK throw" mechanism failed because the SDK's `defaultProviderAuthContext().env(name)` falls through to `process.env[name]`, exposing host-shell secrets. | **Fixed at `f4bb97ef`.** `resolveProviderConfigValue` substitutes `${__ARCHON_BLOCKED_<VAR>__}`. End-to-end verified by `request-auth.integration.test.ts:282-326, 328-378`. Carries forward as fixed — delta does not touch `request-auth.ts`. |
| F1 (round-2, Important) | Important | `buildCustomProviderModelsPath` writes the substituted secrets file world-readable (mode 0o644) under a default umask. | **Fixed at `f4bb97ef`.** `mode: 0o700` + `chmodSync(dir, 0o700)`, `mode: 0o600` + `chmodSync(file, 0o600)`. Verified by `request-auth.integration.test.ts:384-409`. Carries forward as fixed — delta does not touch `request-auth.ts`. |
| F2 (round-2, Important) | Important | Integration test's "no `mock.module` shim" claim is conditional on a script-level invariant; `provider.test.ts:270`'s `mock.module` contaminates the SDK when both files run in one `bun test` invocation. | **Fixed at `f4bb97ef`.** Module-load probe + `test.skipIf(!realSdkAvailable)`. Verified by `bun test src/community/pi/provider.test.ts src/community/pi/request-auth.integration.test.ts` → 103 pass / 4 skip / 0 fail (delta re-verified). |
| F3 (round-2, Important) | Important | Per-call `models.json` is never deleted; long-running processes accumulate files until `ENOSPC` triggers silent R1 regression. | **Fixed at `f4bb97ef`.** `provider.ts:436-455` `finally { rmSync(customProviderModelsPath, { force: true }) }`; `request-auth.ts:235-238` and `:245-247` swallow blocks removed. Two new tests `provider.test.ts:631-686, 688-751` assert cleanup on both paths (delta re-verified — both still pass). |
| S2 (round-2, Important) | Important | Same cleanup gap as F3 (carve-out from seams lens). | **Fixed at `f4bb97ef`** (same mechanism as F3; identical test coverage). Carries forward as fixed. |

Every prior finding is closed/fixed. No finding has gone missing.

## Lens roster

| Lens | Status | Notes |
|---|---|---|
| code | Ran | Verified R1 genuinely closed at `ce4cf7a5`. Read `provider.ts:387-455` end-to-end; confirmed `let customProviderModelsPath: string \| undefined;` at `:399`, assignment INSIDE `try {` at `:401-407`, catch at `:429-435`, finally at `:436-455`. Executed new test in isolation (1 pass / 0 fail / 6 expect() calls). Executed round-3 cleanup tests (2 pass / 0 fail). Executed full pi-community pair (103 pass / 4 skip / 0 fail). `tsc --noEmit` exits 0. Adversarial verification (one-shot revert → test fails; restore → test passes) confirmed. No new findings. |
| seams | Ran | Independently corroborated R1 closure with the same `file:line` evidence (provider.ts:399-407 / :429-435 / :436-450). Executed new test in isolation (1 pass / 0 fail). Executed full pi-community pair (103 pass / 4 skip / 0 fail). `tsc --noEmit -p packages/providers/tsconfig.json` exits 0. Confirmed the `ensurePiPackageDirShim` precedent's classification divergence (different prefix) is deliberate and addressed a different layer (shim install vs. per-call `models.json` write). Root-running CI carve-out on `chmodSync(readOnlyTmp, 0o555)` noted as a test-environment concern shared with round-2/3 tests, not a seam defect. No new findings. |
| tests | Disabled by scope (light mode) | n/a |
| errors | Disabled by scope (light mode) | n/a |
| comments | Disabled by scope (light mode) | n/a |
| types | Disabled by scope (light mode) | n/a |
| docs | Disabled by scope (light mode) | n/a |
| simplify | Disabled by scope (light mode) | n/a |

Both enabled lenses delivered their reports. Neither lens failed to report. No gap in evidence the review was asked for.

## What was decisively checked (synthesis)

- **The exact change site.** `provider.ts:387-455` end-to-end: `let customProviderModelsPath: string | undefined;` declared above the `try {` at `:399` (so the `finally` block at `:436-455` can reference it); the assignment via `buildCustomProviderModelsPath({…})` at `:401-407` sits INSIDE the `try {` opened at `:400`. The catch at `:429-435` now wraps both `mkdirSync`/`writeFileSync`/`chmodSync` errors (from `buildCustomProviderModelsPath`) AND `ModelRuntime.create` errors with the same `'Pi auth storage init failed: …'` framing and `'pi.auth_storage_init_failed'` log line.
- **The cleanup `finally` still works on both paths.** The `if (customProviderModelsPath)` guard at `:448` correctly short-circuits `rmSync` when FS failure leaves the assignment at `undefined`. The round-3 success-path test (`per-call models.json is unlinked after ModelRuntime.create`) and failure-path test (`per-call models.json is unlinked even when ModelRuntime.create throws`) both pass against `ce4cf7a5`.
- **The new test is correctly shaped.** `provider.test.ts:753-861`: real `mkdtempSync` + `chmodSync(readOnlyTmp, 0o555)` to force `EACCES`; pre-populated `archon-pi-shim/package.json` so `ensurePiPackageDirShim()` short-circuits via its `existsSync` guard; `PI_CODING_AGENT_DIR`/`TMPDIR`/`TMP`/`TEMP` all set to the read-only dir (with symmetric snapshot/restore in `finally`); six `expect()` assertions cover the classified prefix, the `EACCES`/permission-denied hint, the `~/.pi/agent/auth.json` hint, the structured log call, and the non-invocation of `mockModelRuntimeCreate`.
- **The fix is the minimum correction.** Only the assignment location changed; no surrounding logic touched. The cleanup `finally` and the existing `catch` framing are reused without modification — they only grew their scope by one statement range.
- **TypeScript compiles clean.** `tsc --noEmit -p packages/providers/tsconfig.json` exits 0 — the `let` declaration at `:399` is in scope at the `finally`'s references at `:448, :450`.
- **No new seam defects.** The `let → let customProviderModelsPath: string | undefined` change introduces no scope leakage (TypeScript `let` declarations hoist into the enclosing block); the new test's env-var snapshot/restore is symmetric across all four env vars; the cleanup `chmodSync(readOnlyTmp, 0o755) → rmSync` ordering prevents EACCES on recursive cleanup; `mockModelRuntimeCreate.mockClear()` in `beforeEach` scopes the new test's `not.toHaveBeenCalled()` assertion correctly.
- **Adversarial verification.** Both lenses independently ran a one-shot revert (the assignment hoisted back above the `try {`) and confirmed the new test fails with the raw `'EACCES: permission denied, mkdir …'` reaching the caller unframed; restoring the fix returns the test to green. The closure is real, not a rename.

Nothing in the delta is left unverified.

---

**Published to PR #2757:** https://github.com/coleam00/Archon/pull/2757#issuecomment-5383756572 (edited in place; read-back verified).

Wirasm:
Closing this as a probe artifact of the governed-run series. The migration itself is complete and working — validate green locally, ubuntu tests, docs, schema-upgrade and docker all green — but the correction round added a test asserting POSIX file modes (0o600/0o700) that Windows CI cannot satisfy, and the series this PR came from has served its purpose. #2678 stays open and now carries every verified fact a fresh implementation needs; this diff remains useful as reference material.

PR #2681, opened 2026-08-21
chore(ci): Renovate replaces the Dependabot updater that cannot read bun.lock

Wirasm:
## Problem and outcome

**A `.github/dependabot.yml` could never have fixed this.** That is the finding that reframes the issue. GitHub's supported-ecosystems reference lists Bun as **version updates: supported, security updates: NOT supported** — there is no `package-ecosystem` entry that unlocks Dependabot security updates for a `bun.lock` repository. So the config-free security-update path infers an ecosystem from the manifests, lands on the legacy `npm_and_yarn` updater, and fails with `dependency_file_not_supported`. It has failed on every run since 2026-04-07. The question is not "why was the config missing?" but "there has never been a GitHub-native path for this repository at all."

Four months with nothing refreshing the lockfile, and #2666's warning that "30 alerts is a floor" turns out to understate it roughly fourfold. Regenerating `bun.lock` from `dev` with **every declared range left byte-identical** takes `bun audit` from **163 advisories to 24**. 139 of 163 exist purely because nothing refreshes the lockfile — no manifest change, no range widened, no upgrade risk. That class produces no alert at all, because the declared range *could* resolve to something patched.

- **Outcome:** Renovate's `bun` manager reads `bun.lock` and the `packages/*` workspaces and opens PRs against `dev`. Three frozen resolutions are refreshed now: 163 → 143.
- **Invariant:** `bun install --frozen-lockfile` succeeds. Lockfile and manifests stay mutually consistent, so CI's install step never diverges from a local install.
- **Scope boundary:** no `better-auth`, no `astro`, no `sharp` — those need real version bumps and are filed. No `.github/dependabot.yml`, because it cannot work.
- **Precondition:** the 139-advisory sweep is gated behind #2682. See *What happens when this merges*.

## Review guidance

- **Feedback requested:** the rollout posture. The day-one PR count is measured below; I recommend shipping with no throttle and have set none. That is the call I most want checked.
- **Start here:** `renovate.json:41` — `lockFileMaintenance.enabled: true`, off by Renovate default and the only setting that reaches the invisible stale-but-in-range class. Then `renovate.json:28` — `baseBranchPatterns`, the #957 archive guard.
- **Review order:** `renovate.json`, then `package.json` (3 override lines), then `bun.lock`. The lockfile diff is fully enumerated below — check my accounting rather than reading hashes.
[...]

PR #2762, opened 2026-08-24
fix(providers): migrate to Pi SDK 0.84.2, clearing the vulnerable undici 8.5.0

### LINKED ISSUE #2678: fix(security): clear vulnerable undici 8.5.0 from under pi-coding-agent

### LINKED ISSUE #2682: fix(providers): pi-ai 0.80.10 is in-range under `^0.80.6` and does not compile — 109 type errors block any lockfile regeneration
## Problem

`packages/providers/package.json:35-36` declares:

```json
"@earendil-works/pi-ai": "^0.80.6",
"@earendil-works/pi-coding-agent": "^0.80.6",
```

`0.80.10` is published and the caret permits it. **It does not compile.** Upstream shipped breaking API removals at patch level.

Reproduced twice independently — delete `bun.lock`, `bun install`, `bun run type-check`:

```
error TS2339: Property 'AuthStorage' does not exist on type
  'typeof import(".../@earendil-works/pi-coding-agent/dist/index")'
error TS2339: Property 'create' does not exist on type 'typeof ModelRegistry'
error TS2353: 'authStorage' does not exist in type 'CreateAgentSessionOptions'
error TS2339: Property 'modelRegistry' does not exist on type 'CreateAgentSessionOptions'
error TS2305: Module '"@earendil-works/pi-ai/oauth"' has no exported member 'getOAuthProvider'
error TS2305: ... 'getOAuthApiKey' / 'anthropicOAuthProvider' / 'openaiCodexOAuthProvider'
            / 'githubCopilotOAuthProvider' / 'OAuthProviderId' / 'OAuthProviderInterface'
```

**109 errors**, concentrated in `src/oauth` (35), `src/community/pi/provider` (32), `src/community/pi/model` (24), `src/community/pi/request` (5). `bun run check:pi-vendor-map` also exits 2 (`stale vs the installed pi-ai SDK`).

Call sites: `packages/providers/src/community/pi/provider.ts:384,385,397,406,761`, `packages/providers/src/oauth.ts:16-26`, plus `provider.test.ts` and `request-auth.test.ts`.

## Why this needs an owner now

It is latent today — `bun.lock` pins 0.80.6 and CI installs `--frozen-lockfile`, so `dev` is green. It stops being latent the moment anything regenerates the lockfile.

That matters because #2681 enables Renovate's weekly `lockFileMaintenance`, which does exactly that. **The first maintenance PR will arrive red**, with these 109 errors, and cannot merge until this migration lands. That PR is also the one carrying the rest of the lockfile refresh — measured at 139 of the repository's 163 advisories — so this blocks the sweep, not just one bump.

Worth noting the fail-fast machinery worked: `check:pi-vendor-map` fires on exactly this and is a required CI check. The gap is that nobody has done the source migration it is pointing at.

## Scope

Migrate `@archon/providers` to the 0.80.10 API:

- Replace `AuthStorage` and `ModelRegistry.create` with their current equivalents
- Replace `CreateAgentSessionOptions.authStorage` / `.modelRegistry`
- Rewrite `packages/providers/src/oauth.ts` against the current `@earendil-works/pi-ai/oauth` export surface — this is the subscription-login boundary (Claude Pro/Max, Copilot), so it is the highest-risk part
- Update `provider.test.ts` and `request-auth.test.ts`
- Run `bun run generate:pi-vendor-map` and `bun run generate:capability-matrix`

Then decide the declared range. If upstream ships breaking changes at patch level under a caret, `^0.80.6` is not expressing the real compatibility contract and a tighter pin may be more honest.

## Verification

- `bun run validate` on a regenerated lockfile — currently the failing command
- Pi provider smoke: a workflow node with `provider: pi` completes
- Subscription-login smoke for at least one vendor, since `oauth.ts` is rewritten
- `bun audit` on the regenerated tree, to confirm the sweep this unblocks

## Context

Found by reviewing #2681, which simulated its own first `lockFileMaintenance` run rather than assuming it would be green. Related: #2282 proposes excluding the AI SDKs from automation for adjacent reasons.

Wirasm:
## Problem and outcome

`^0.80.6` permits pi-ai/pi-coding-agent versions that removed API at patch level, so any lockfile regeneration breaks the build with 109 type errors (#2682) — which is exactly what Renovate's first `lockFileMaintenance` PR would hit, gating a 139-advisory lockfile sweep. The same pin holds a vulnerable `undici@8.5.0` (six advisories, one high) under pi-coding-agent that survives lockfile maintenance (#2678).

- **Outcome:** Both Pi packages are at `^0.84.2` (npm latest, maintainer scope decision on #2682). `bun.lock` carries no undici in the vulnerable `>=8.0.0 <8.9.0` range — exactly two entries remain: `8.9.0` (patched) and discord.js's pre-existing `6.21.3` (separate 6.x advisory, out of scope). And Archon consumes the 0.84 SDK properly: a model the user's Pi TUI learned (`~/.pi/agent/models-store.json`) resolves in an Archon workflow with zero configuration — model availability now derives from the user's Pi, not a frozen snapshot baked into whichever SDK version Archon pins.
- **Invariant:** Per-call secrets never reach `process.env` or leave literal protected values on disk; subscription tokens refresh on read before delivery; the `Pi auth storage init failed:` error framing covers every init-path failure.
- **Scope boundary:** discord.js's undici 6.x advisory, the broader lockfile sweep (#2681), and a live end-to-end subscription-login smoke are not part of this PR.

## Review guidance

- **Feedback requested:** Correctness of the OAuth adapter semantics (`packages/providers/src/oauth.ts`) and the per-call models.json substitution seam (`request-auth.ts`).
- **Start here:** `packages/providers/src/oauth.ts` — Archon now owns the legacy `getOAuthApiKey`/`OAuthProviderInterface` surface the SDK dropped; this is the subscription-login boundary that reaches `@archon/core`.
- **Review order:** `oauth.ts` → `community/pi/provider.ts` (ModelRuntime migration) → `community/pi/request-auth.ts` (per-call substitution) → tests → `model-catalog.ts`, `oauth-bridge.ts`, generated vendor map, lockfile.
- **Lower-attention areas:** `pi-vendor-map.generated.ts` (regenerated; `check:pi-vendor-map` green), `bun.lock` (plain `bun install`, diff scoped to the Pi tree).
- **Known risk or uncertainty:** A real subscription login was not exercised end-to-end (needs interactive browser auth). Mitigation: both OAuth singletons load and mint through the real installed SDK (smoke below), the prompt/event mapping mirrors what the pre-0.84 legacy wrappers did (verified by reading both 0.84 flows), and refresh semantics are unit-tested.

## Solution

Closed PR #2757 (a governed-run probe) served as the reference. This PR adopts its migration — `ModelRuntime.create({ authPath, modelsPath? })` + `new ModelRegistry(runtime)` in `provider.ts`, the per-call substituted `models.json` replacing the AuthStorage wrapper the 0.84 SDK discards, `model-catalog.ts` via `getModels()`, and the hardening from its three review rounds (0o600/0o700 modes, finally-cleanup, protected-key placeholders) — and corrects four defects on top:

1. **`getOAuthApiKey` never refreshed.** 0.84's `toAuth` derives auth from whatever credential it is given and never throws on staleness, so the reference's mint-then-refresh-on-error path could never fire — stale subscription tokens (Copilot rotates ~30 min) would be delivered to runs. Restored the pre-0.84 contract: refresh when `Date.now() >= creds.expires`, then mint. Unit-guarded at the new `mintOAuthApiKey` seam.
2. **Windows CI red.** The POSIX file-mode test (`0o600`/`0o700`) is now `test.skipIf(win32)` — Windows has no POSIX mode bits.
3. **Command-form values were rewritten.** The SDK treats a raw value starting with `!` as a command executed at request time without template substitution; the parser now leaves those untouched (and the `$!` escape test was fixed — it had used a raw `!...` value, which is a command, not an escape).
4. Simplifications: one parameterized deep-path loader for the OAuth singletons (verified necessary — pi-ai 0.84.2's public `./oauth` subpath is type-only at runtime), dead `name` field and unconsumed `OAuthProviderId` export dropped.

The dynamic-catalog acceptance is proven by a new real-SDK test: a model present only in a scratch `models-store.json` resolves through the exact `new ModelRegistry(await ModelRuntime.create(...))` + `find()` construction `provider.ts` performs (offline, own `bun test` process to dodge `mock.module` contamination).

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Pi model refs resolve against the 0.80.6 baked catalog + hand-edited `models.json` (e.g. 270 openrouter models) | Models the user's Pi learns dynamically resolve too (`models-store.json`, e.g. 353 openrouter models); no Archon configuration |
| **Failure behavior** | Lockfile regeneration → 109 type errors; expired subscription token on the (never-reachable) 0.84 mint path would be served stale | Lockfile regeneration compiles; expired token refreshes before delivery, refresh failure logs `oauth_refresh_failed` |

## Validation

- `bun run validate` — green (full suite, regenerated lockfile) — proves the compile half of #2682 and all repo checks including `check:pi-vendor-map`.
- `grep '"undici@' bun.lock` — only `6.21.3` and `8.9.0`; `bun audit` no longer reports GHSA-4cwx-7wf7-3272 (1 hit on dev, 0 here) — proves #2678's acceptance.
- `bun test src/community/pi/model-store.integration.test.ts` — pass — proves the #2682 capability acceptance against the real SDK.
- `bun test src/oauth.test.ts` — pass — guards refresh-on-expiry (echo-on-fresh / rotate-on-stale / propagate-on-refresh-failure).
- Full providers test chain — 39 invocations, 0 fail.
- Real-SDK loader smoke (`bun -e`): anthropic + copilot singletons load via deep path and mint an access token; unknown vendor returns null.
- **Not verified:** Windows behavior (only provable on hosted CI — the one prior red is now platform-skipped) and a live interactive subscription login (see Review guidance).

## Delivery considerations

|---|---|---|
| Compatibility / migration | Stored OAuth blobs, per-run `auth.json` delivery (`ARCHON_PI_AUTH_PATH`), and `models.json` custom providers keep working; no schema or config changes | request-auth/provider tests; key-store rotation detection unchanged |
| Security / permissions / data | Per-call models.json holds substituted secrets: 0o600 file / 0o700 dir, removed in `finally`; protected keys written only as unresolvable placeholders | `request-auth.ts`, integration tests incl. host-shell-leak case |
| Rollout / rollback | Single commit; revert restores 0.80.6 pin and lockfile | — |

## Links

- Closes #2678
- Closes #2682
- Related #2681
- Supersedes #2757

## Summary by CodeRabbit

* **New Features**
  * Improved OAuth sign-in and credential refresh for supported providers, with clearer login prompts and safer credential handling.
  * Custom provider authentication now supports request-specific environment values without exposing host credentials.
  * Pi model catalogs recognize locally configured and custom-provider models.

* **Bug Fixes**
  * Improved temporary authentication configuration cleanup and isolation.
  * Added validation for invalid OAuth expiry data and clearer setup errors.

* **Tests**
  * Expanded coverage for OAuth, model discovery, authentication substitutions, cleanup, and failure scenarios.


Wirasm:
## Ready to merge

**Correction round verified.** All five blocking findings from the first pass (R1–R5) are fixed, confirmed by independent re-verification against the exact prior head range (`2a424962...fb1a928f`), not just the implementer's own claims. The non-blocking cleanup (R6) is also fixed. R7 stays declined, unchanged, per standing operator decision (pre-existing, unrelated to this PR, tracked as a separate follow-up).

This PR migrates Archon's Pi provider integration from pi-ai/pi-coding-agent 0.80.6 to 0.84.2, replacing two pieces of logic that used to live inside the third-party SDK with Archon's own hand-rolled equivalents: the OAuth refresh/mint contract, and the `${VAR}` template substitution + protected-env-key blocking used to keep secrets off custom Pi providers. The first review found the migration itself sound but surfaced five edge-case gaps in those two hand-rolled pieces. The correction commit (`fb1a928f`, "fix(providers): close the review's five edge cases in the Pi substitution and OAuth mint paths") addresses all of them with targeted fixes and new regression tests, and none introduce a new defect. The sharpest of the five (R4 — a protected secret's blocking placeholder could be silently discarded when an unrelated missing var shared its field) was empirically re-reproduced against the fix, plus two harder variants (duplicate protected refs, multiple missing refs), and confirmed closed in all three.

Also verified: the windows-latest CI fix. The FS-error classification test's provocation changed from chmod'ing a directory to `0o555` (a POSIX-only mechanism — Windows ignores directory write-permission bits, which is why the original test was red on windows-latest) to a file occupying the exact `<tmpdir>/archon-pi-models` path, so `mkdirSync(..., {recursive: true})` throws `EEXIST` identically on every platform. This keeps the classification path under test everywhere instead of skipping it on Windows.

**0 blocking · 0 non-blocking open** (R7 stays declined as a tracked, separate concern)

**Validation:** All required checks pass on the corrected head. No PR-caused failure. **Resolved:** 6 (R1–R6).

### Findings

|---|---|---|---|
| `R1` | Important | Plaintext-secret temp file orphaned if `chmodSync` fails right after `writeFileSync` succeeds | FIXED |
| `R2` | Important | `${VAR}` substitution treats an empty-string env value as resolved, unlike the SDK it mirrors | FIXED |
| `R3` | Important | `getUserModelsPath()` doesn't tilde-expand `PI_CODING_AGENT_DIR`, silently reverting to the pre-fix leak | FIXED |
| `R4` | Important | Protected-key placeholder guarantee silently bypassed when a protected and an unrelated missing `${VAR}` co-occur | FIXED |
| `R5` | Important | `mintOAuthApiKey`'s refresh decision silently no-ops on malformed/missing `expires` from unvalidated DB data | FIXED |
| `R6` | Suggestion | Unnecessary `as unknown as` cast in `oauth-bridge.ts`, with a stale comment contradicting it | FIXED |
| `R7` | Suggestion | GitHub Copilot enterprise-domain login prompt can stall the OAuth bridge — pre-existing, not caused by this PR | DECLINED |

<details>
<summary><code>R1</code> — Orphaned plaintext-secret temp file on a write-then-chmod failure</summary>

**Impact:** If `writeFileSync` succeeded but the immediately-following `chmodSync(filePath, 0o600)` threw, `buildCustomProviderModelsPath` threw before returning the path, so `provider.ts`'s `finally`-gated cleanup never saw the value and the file (a cleartext secret) was left on disk.

**Evidence (fix):** `packages/providers/src/community/pi/request-auth.ts:309-315` now wraps the write+chmod sequence in its own try/catch:
```ts
try {
  writeFileSync(filePath, payload, { mode: 0o600 });
  chmodSync(filePath, 0o600);
} catch (err) {
  rmSync(filePath, { force: true });
  throw err;
}
return filePath;
```
`rmSync(..., { force: true })` correctly no-ops (ignores `ENOENT`) for the sub-case where `writeFileSync` itself throws before any file exists, and correctly removes the file for the sub-case where `chmodSync` throws after a successful write — the orphan window is closed either way.

**Required outcome:** Met.

**Found by:** `prp-core:code-reviewer` · **Verified by:** `prp-core:code-reviewer` (correction pass)

**Disposition:** FIXED. Two non-blocking observations, neither required by this finding: (1) if `rmSync` itself threw for a non-`ENOENT` reason, that new error would replace the original `err` in the propagated message — very unlikely given the file was just created by this same process, not required to fix; (2) no dedicated test provokes this exact write-then-chmod-throws window specifically (the implementer's own commit notes the provocation isn't reliably portable) — the fix is generic (removes the file on *any* exception in the block, not chmod-specific), so this is a coverage gap, not a correctness gap.

</details>

<details>
<summary><code>R2</code> — Empty-string env value treated as a valid substitution</summary>

**Impact:** `resolveProviderConfigValue` checked presence (`!(part.name in env)`) rather than truthiness, so an empty-string env var was written into the per-call `models.json` as `apiKey: ""`, which fails the SDK's own schema and silently drops the whole custom provider instead of surfacing the SDK's actionable "Failed to resolve ... from environment variable" error.

**Evidence (fix):** `packages/providers/src/community/pi/request-auth.ts:145-151` (classification pass) and `:180-188` (substitution pass) both now use `!env[part.name]` — a falsy check, matching the SDK's real `resolveEnvConfigValue` (`env?.[name] || process.env[name] || undefined`), read directly from the installed `dist/core/resolve-config-value.js:71-73`. No `in`-based presence check remains anywhere in the file. New test `request-auth.test.ts:207-219` asserts `buildCustomProviderModelsPath(...)` returns `undefined` for an empty-string value — the correct observable (template left unsubstituted), not an inspection of a written empty string.

**Required outcome:** Met — genuinely behaviorally equivalent to the SDK's own fallback semantics, verified against the SDK source directly, not just "looks plausible."

**Found by:** `prp-core:seam-analyzer` · **Verified by:** `prp-core:seam-analyzer` (correction pass)

**Disposition:** FIXED.

</details>

<details>
<summary><code>R3</code> — <code>getUserModelsPath()</code> doesn't tilde-expand <code>PI_CODING_AGENT_DIR</code></summary>

**Impact:** A `PI_CODING_AGENT_DIR=~/...` override wasn't tilde-expanded, so the substitution seam's `existsSync` check silently missed the real path and Pi fell back to its own (tilde-correct) resolution with unprotected `process.env`-based substitution — quietly reopening the exact leak this PR exists to close.

**Evidence (fix):** `packages/providers/src/community/pi/request-auth.ts:41` imports the real `expandTilde` from `@archon/paths` (not reimplemented), applied at `:211-212`. Read `expandTilde`'s implementation directly (`packages/paths/src/archon-paths.ts:42-48`) and compared it line-for-line against the SDK's actual `normalizePath` tilde handling (`dist/utils/paths.js:58`): both correctly resolve bare `~` (→ homedir), `~/subpath` (→ `join(homedir(), subpath)`), and pass an absolute non-`~` path through unchanged. New test `request-auth.test.ts:254-273` asserts `getUserModelsPath()` against a real `join(homedir(), 'custom-agent', 'models.json')` expectation (not just a smoke call), plus the absolute-path and unset-env branches.

**Required outcome:** Met.

**Found by:** `prp-core:code-reviewer` (Suggestion), `prp-core:seam-analyzer` (Important — coordinator agreed) · **Verified by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer` (correction pass)

**Disposition:** FIXED. One narrow, pre-existing (not introduced by this correction) divergence noted for completeness: `@archon/paths`' `expandTilde` treats a value like `~foo` (tilde immediately followed by a non-slash character) as home-relative, while the SDK's resolver leaves it untouched — not a plausible operator input for a directory env var (tilde-username shell syntax with no leading slash), not required to fix.

</details>

<details>
<summary><code>R4</code> — Protected-key placeholder guarantee bypassed when a protected and a missing <code>${VAR}</code> co-occur</summary>

**Impact:** When a protected var and an unrelated missing var appeared in the same template field, the old "give up on missing" bail-out discarded the already-computed protected placeholder and returned the original template — silently no-opping the protected-key defense in exactly the combination it exists to catch. Empirically reproduced against the original code in the first review pass.

**Evidence (fix):** `packages/providers/src/community/pi/request-auth.ts:139-191` restructures `resolveProviderConfigValue` into two passes: classify all parts first (`hasProtected`/`hasMissing`, `:145-151`), bail out only when something's missing **and nothing is protected** (`:152`), then substitute with the protected branch UNCONDITIONAL (`:167-178`) and a bare missing ref preserved as template text rather than aborting the field (`:180-186`). Re-verified empirically against: (1) the original repro (`${PROTECTED_VAR}${MISSING_VAR}`) — placeholder survives, secret absent from output; (2) a harder same-key-twice-plus-two-missing-refs case (`${GH_TOKEN}${GH_TOKEN}${MISSING_A}${MISSING_B}`) — both protected occurrences blocked, both missing refs preserved, secret absent; (3) a cross-field case (protected in `apiKey`, missing in a header), covered by the new test at `request-auth.test.ts:229-242`. New tests assert on the whole written file's JSON string (`not.toContain('acting-user-secret')`), not just one field.

**Required outcome:** Met — the two-pass structure is order-independent, so part ordering within a field can't resurrect the original bug.

**Found by:** `prp-core:code-simplifier` · **Verified by:** `prp-core:code-simplifier` (correction pass, including two harder variants beyond the original repro)

**Disposition:** FIXED. This was the finding with the most direct security relevance among the five; it is now closed with the strongest verification of the set.

</details>

<details>
<summary><code>R5</code> — Refresh decision silently no-ops on malformed/missing <code>expires</code></summary>

**Impact:** `mintOAuthApiKey`'s refresh decision (`Date.now() >= creds.expires`) had no runtime guarantee that `expires` was numeric at the one place a decrypted DB blob entered the system, so a malformed/missing `expires` would silently skip refresh and deliver a stale-but-present token as a successful mint.

**Evidence (fix):** `packages/core/src/db/user-provider-key-store.ts:216-229` adds `if (piProvider && !Number.isFinite(creds.expires)) { ...log 'oauth_malformed_expires'...; return null; }` between the `JSON.parse` and any mint attempt — strictly before `getOAuthApiKey` is ever called for a Pi-backed vendor. `Number.isFinite` correctly rejects `undefined`/`null`/`NaN`/non-numeric types while accepting all valid epoch-ms values including `0`. New test `user-provider-key-store.test.ts:229-244` asserts both the `null` return AND that the mint mock (`mockGetOAuthApiKey`) was never called — the stronger of the two possible assertions.

**Required outcome:** Met, placed exactly at the deserialization boundary as specified.

**Found by:** `prp-core:type-design-analyzer` · **Verified by:** `prp-core:type-design-analyzer` (correction pass)

**Disposition:** FIXED. One documentation-accuracy caveat, not a blocking defect: the correction's own new comment states "the openai branch validates its own shape inside `mintOpenAiOAuthApiKey`" — reading that function directly (`packages/core/src/credentials/openai-oauth.ts:323-333`, unchanged by this PR in either round) shows it has the same class of gap (a non-numeric `expires` skips refresh silently and returns the existing token as success) — the comment overstates what that guard does. Since `openai-oauth.ts` is genuinely untouched by this PR, this is pre-existing and out of scope for a new finding here; worth a note for whoever eventually touches that file.

</details>

<details>
<summary><code>R6</code> — Unnecessary cast in <code>oauth-bridge.ts</code>, with a stale contradicting comment</summary>

**Impact:** `loginPromise` was typed to force an unneeded `as unknown as Promise<PiOAuthCredentials>` cast on the OpenAI branch, next to a comment that already (incorrectly) claimed no cast was needed.

**Evidence (fix):** `packages/core/src/credentials/oauth-bridge.ts` — `loginPromise` is now typed `Promise<DeliveryOAuthCredentials>` (`delivery.ts`'s loose `Record<string, unknown>`), and the `runOpenAiManualLogin(session)` branch carries no cast. `PiOAuthCredentials` no longer appears anywhere in the file (grep-confirmed). Verified `bun x tsc --noEmit` in `packages/core` directly (a different compilation unit than the providers check) — clean. Both comments (at the `runOpenAiManualLogin` return and at the `loginPromise` declaration) now agree with each other and with the code.

**Required outcome:** Met.

**Found by:** `prp-core:code-simplifier` (confirmed safe by `prp-core:seam-analyzer`, `prp-core:type-design-analyzer`) · **Verified by:** `prp-core:code-simplifier` (correction pass)

**Disposition:** FIXED.

</details>

<details>
<summary><code>R7</code> — GitHub Copilot enterprise-domain prompt can stall OAuth login (pre-existing)</summary>

**Impact:** GitHub Copilot's enterprise-domain prompt-before-`device_code` sequence could stall `oauth-bridge.ts`'s `pollOAuth`, unrelated to this PR.

**Evidence:** Confirmed present identically in pre-migration `pi-ai@0.80.6`; `oauth-bridge.ts`'s wiring is unchanged by both the original PR and the correction commit (grep-confirmed no `pollOAuth` touches in the correction diff).

**Disposition:** DECLINED — stands per operator decision; verified pre-existing and untouched by the correction. A follow-up issue is being filed separately; do not count against this PR's readiness.

</details>

<details>

#### Reviewer coverage

First pass: `code` → R1, R3, R7 · `seams` → R2, R3, R6 (confirmed safe) · `simplify` → R4, R6 · `types` → R5, R6 (confirmed safe).

Correction verification (bounded to `2a424962...fb1a928f`, same scope ownership retained): `code` → R1 FIXED, R3 FIXED, R7 confirmed untouched · `seams` → R2 FIXED, R3 FIXED · `simplify` → R4 FIXED, R6 FIXED · `types` → R5 FIXED. No new defect introduced by the correction commit was found by any of the four verification agents.

#### Validation

|---|---|---|
| `bun x tsc --noEmit` (packages/providers) | PASS | clean, no errors, on the corrected head |
| `bun x tsc --noEmit` (packages/core) | PASS | clean — confirms the R6 cast removal type-checks in its own compilation unit |
| `bun run test` (packages/providers, 39 invocations) | PASS | 0 fail, on the corrected head |
| `bun test src/db/user-provider-key-store.test.ts` (packages/core, own invocation) | PASS | 27 pass, 0 fail — includes the new R5 regression test |
| `bun test src/credentials/oauth-bridge.test.ts` (packages/core, own invocation) | PASS | 22 pass, 0 fail |
| `bun x eslint` on corrected non-test source files, `--max-warnings 0` | PASS | 0 errors, 0 warnings |
| `bun x prettier --check` on corrected non-test files | PASS | all match repo style |
| Windows-latest CI provocation change | VERIFIED BY CODE READING | `provider.test.ts` now provokes `EEXIST` via a file occupying the target directory path (cross-platform) instead of a POSIX-only `chmod 0o555` (Windows ignores directory write bits) — confirmed in the diff; actual windows-latest CI run being watched separately by the requester, not depended on for this verification |
| Live interactive subscription login | NOT RUN | requires interactive browser auth; unchanged from first pass, matches author's own "not verified" note |

Note on process: running `user-provider-key-store.test.ts` and `oauth-bridge.test.ts` together in one `bun test` invocation produces spurious cross-file failures (mock.module() pollution — a documented repo issue, not a PR defect); each passes cleanly in its own invocation, matching `packages/core`'s `package.json` test script, which already runs them separately.

</details>

Issue #2325, opened 2026-07-30
fix(core): abandon stops a reachable owner before releasing the run, and the in-use refusal names working commands


dsera3sd:
## Current scope (settled 23 September 2026). This section is the contract.

The original report is kept below unchanged. Its timer proposal is rejected: AGENTS.md forbids declaring a live owner dead from age. Its point (b) is partly stale, since `archon workflow cancel` now exists.

### Problem

`archon workflow abandon` (`abandonWorkflow` in `packages/core/src/operations/workflow-operations.ts`) checks only the stored status, then records the run `cancelled`. That terminal status releases the run's worktree lock and any admitted resource slot, even when the owning process is still alive. A new run can then start on a checkout the old process is still writing to.

The refusal a blocked run prints (`packages/workflows/src/executor.ts`, "This worktree is in use") also tells every surface to use chat syntax (`/workflow cancel <id>`) and never mentions abandon for a run whose owner is gone.

### Why it matters, and why now

This is the lock-held-forever report below, seen from the recovery side. Resource admission (PR #3415, merged) now makes abandon a documented recovery path for trigger-started runs, and #2908's live-owner endpoint gives abandon a positive liveness signal it can use.

### Desired outcome

1. Abandon first asks the run's live-owner endpoint. If an owner answers, abandon stops it through the same path `cancel` uses (prove the owner, terminate the process tree, wait), then records `cancelled`.
2. If no owner answers, abandon records `cancelled` as today. That is the operator's explicit call that the run is dead. First it prints what's recorded (host, pid, last activity), so a run owned by another host sharing the database isn't a silent guess.
3. If an owner answers but can't be stopped, abandon fails loudly and leaves the run unchanged.
4. The worktree-in-use refusal tells the operator the working commands on their surface: cancel for a live run, abandon for a run whose owner is gone. The engine doesn't hard-code chat or CLI syntax. Command spelling belongs to the adapter or surface that renders the message.

### Invariants

- No timer, age or PID check decides liveness. A PID may appear as a warning, never as a gate.
- No force flag.
- `cancel` keeps its contract: it refuses when no owner answers.
- The existing cascade and parent-blocked behavior of abandon is unchanged.
- Cross-host ownership is out of scope. Abandon says plainly when the recorded owner is on another host.

### Acceptance

- A live owner is stopped, then the run is cancelled. A waiting admission for the same resource proceeds only after the process has exited. Test with a real detached run.
[...]

Issue #1242, opened 2026-04-16
bug(workflows): script node timeout does not kill wrapper grandchildren (uv/bun pattern)


shaun0927:
## Summary

`executeScriptNode` invokes runtimes that are themselves wrappers — `uv run python -c …` or `bun run <script>` — via `execFileAsync(cmd, args, { cwd, timeout })`. Node's `timeout` option sends `SIGTERM` to the **direct child** only. The wrapper dies, but the real work (the Python interpreter or the long-running TS file) continues in a separate process. Archon reports the node as timed out while the script keeps running in the background, producing side effects (file writes, network calls, git pushes) long after the run is considered failed.

## Reproducer

Reproduced on `v0.3.6` @ `59cda08`:

```js
// repro.mjs
import { promisify } from 'node:util';
import { execFile } from 'node:child_process';
import { writeFileSync, existsSync, readFileSync } from 'node:fs';

const exec = promisify(execFile);

// Grandchild survives wrapper — mimics `uv run python` spawning the interpreter
writeFileSync('/tmp/grandchild.sh',
  `#!/bin/bash\nsleep 10\necho ALIVE pid=\$\$ > /tmp/marker\n`, { mode: 0o755 });
writeFileSync('/tmp/wrapper.sh',
  `#!/bin/bash\nbash /tmp/grandchild.sh &\nwait\n`, { mode: 0o755 });

const start = Date.now();
try { await exec('bash', ['/tmp/wrapper.sh'], { timeout: 2000 }); }
catch (e) { console.log(`wrapper killed after \${Date.now() - start}ms: \${e.signal}`); }

await new Promise(r => setTimeout(r, 12000));
console.log(existsSync('/tmp/marker')
  ? 'LEAKED: ' + readFileSync('/tmp/marker', 'utf-8').trim()
  : 'clean');
```

Output:

```
wrapper killed after 2008ms: SIGTERM
LEAKED: ALIVE pid=34143
```

The wrapper received `SIGTERM` at 2 s. The grandchild slept for the full 10 s and then wrote the marker — proving it survived the node's timeout.

## Root Cause

`packages/workflows/src/dag-executor.ts:1576` (script) and `:1362` (bash):

```ts
const { stdout, stderr } = await execFileAsync(cmd, args, { cwd, timeout });
```

[...]

PR #2766, opened 2026-08-24
fix(cli): actively stop detached workflow runs

### LINKED ISSUE #2759: bug(cli): a --detach run has no supported stop — abandon only writes state, /workflow cancel ignores the id
## Summary

- What's missing: for a run started by the CLI with `--detach`, no supported command stops the work that is in flight on the host. `archon workflow abandon` is documented as not killing the in-flight subprocess (it transitions the DB row), and the documented alternative — the chat `/workflow cancel` — resolves only *the calling conversation's* active run, so it can never target a CLI-started run. The operator is left with a run whose status reads `cancelled` while its node keeps working, and no supported way to stop it.
- Not a claim that `abandon` is broken: its behavior matches its documentation. The gap is that the escape hatch the docs point CLI users to does not cover runs the CLI itself started.
- When observed: v0.9.0 (release binary, 2026-08-23). The paths involved are unchanged between the 0.9.0 release (`671e2ee`) and current `main` (`df4c0d0`).
- Severity: `minor` as a capability gap, but it bites hardest exactly where it matters — a long node (build, deploy, test suite) keeps running to completion after the operator has cancelled the run, doing real work on the host.

## Steps to Reproduce

Prerequisites: Linux/macOS with `bash`, `seq`, `wc`; Archon v0.9.0. No provider credentials needed — bash nodes only.

1. Create a directory and `git init` it (the CLI refuses to run outside a git repository):

```bash
mkdir abandonrepro && cd abandonrepro && git init -q && mkdir -p .archon/workflows
```

2. Save as `.archon/workflows/abandonrepro.yaml`:

```yaml
name: abandonrepro
description: repro — a CLI --detach run cannot be actively cancelled
provider: claude
model: sonnet
nodes:
  - id: nodeA
    timeout: 600000
    bash: |
      for i in $(seq 1 40); do
        printf '%s A-tick %s\n' "$(date -u +%H:%M:%S)" "$i" >> ticks.txt
        sleep 1
      done
      printf '%s A-DONE\n' "$(date -u +%H:%M:%S)" >> ticks.txt
  - id: nodeB
    depends_on: [nodeA]
    timeout: 600000
    bash: |
      printf '%s B-STARTED\n' "$(date -u +%H:%M:%S)" >> ticks.txt
```

3. `rm -f ticks.txt && archon workflow run abandonrepro --no-worktree --detach "repro"`.
4. Wait ~12 seconds. Take the run id from `archon workflow runs` (the `--detach` acknowledgement prints the child log path, not the id).
5. Try the documented active cancel: `archon chat "/workflow cancel <run-id>"`.
6. Then `archon workflow abandon <run-id>`, and watch `ticks.txt` and `archon workflow runs` for the next minute.

Cleanup: `nodeA` ends by itself after 40 ticks.

## Expected vs Actual

- **Expected**: some supported command stops the work of a run the CLI started, rather than only recording that it was cancelled.

- **Actual** (v0.9.0):
  
  - Step 5 returns `No active workflow to cancel.` regardless of the run id supplied.
  - Step 6 returns `Abandoned workflow run: <id>` and the run reads `cancelled` about two seconds later — while `nodeA` keeps running. Recorded run `28d29e2660976195826a2c9233d4a409` (the workflow above, abandoned ~12 seconds into `nodeA`): `nodeA` ran to completion, writing all 40 ticks and its `A-DONE` marker, with the run reading `cancelled` throughout.
  - The run does correctly stop at the next layer boundary: `nodeB` never started.
  - A single-node variant of the same workflow with a 240-iteration loop (run `0ba6c4cd97691714983b4147fc875e61`) shows the same thing over a longer window: 22 ticks at the moment of abandon, 83 ticks 61 seconds later, status `cancelled` from two seconds after the call onward.
  
  So the in-flight node runs to completion — for a 90-minute suite node, that is 90 minutes of work after the operator cancelled.

## Code path

- `packages/cli/src/commands/workflow.ts` — the `abandon` action calls `abandonWorkflow(resolvedId)`, which updates run state and reclaims isolation environments; no signal is sent to the executing process. `--detach` (same file) spawns the runner as a detached background child, so a later, separate CLI invocation has no handle on it.
- `packages/core/src/handlers/command-handler.ts:691` — `/workflow cancel` looks its target up with `getActiveWorkflowRun(conversation.id)` and answers `No active workflow to cancel.` when there is none. A CLI-dispatched run is recorded under its own `cli-*` conversation while `archon chat` runs in a separate `cli-chat-*` one, and the run id supplied to the command is parsed but not used for the lookup — so the command reports on the chat conversation's (empty) active run, whatever id is passed.
- `packages/workflows/src/dag-executor.ts` — a bash node is awaited via `runSubprocess()` with only its configured timeout, so nothing interrupts it; the executor does re-check run status before starting the next layer, which is why the second node is skipped.
- The documented contract is explicit and matches all of the above: `.claude/skills/archon/references/cli-commands.md` — "This does NOT kill an in-flight subprocess — it only transitions the DB row", and "To actively cancel a running workflow (terminate its subprocess), use the chat slash command `/workflow cancel <run-id>` on the platform that started it". For a CLI-started run there is no chat platform to go to, and the chat command does not look runs up by id.

## Relationship to existing issues

- **#1242** (`script node timeout does not kill wrapper grandchildren`) shares the theme of termination not reaching the work, but its mechanism is different (SIGTERM delivery to a wrapper's grandchildren, versus no cancellation path at all here). Related, not a duplicate.
- **#2325** (`a dead run holds its worktree lock indefinitely`) is the mirror image: state left behind by a process that died, where this is a process still working after the state says cancelled.

## Suggested direction

What is missing is a termination path that reaches the executing process of a CLI-started run — note that simply making `/workflow cancel` honour the id it already accepts would not be enough on its own, since that path also only writes state. Failing that, a line in `abandon`'s own output — that the current node continues, with the child log path — would at least tell the operator what is still happening, since today the terminal status is the only signal and it says `cancelled`.

## Environment

- Archon v0.9.0 (release binary), CLI `--detach` runs, SQLite backend.

Wirasm:
## Problem and outcome

A detached CLI workflow could be marked cancelled in the database while its active bash or script process kept running. The only process handle was discarded by the launching parent, so neither `workflow abandon` nor conversation-scoped cancellation could prove or stop the exact detached owner.

- **Outcome:** `archon workflow cancel <run-id>` now contacts the live detached owner for that exact run, stops its host process tree, confirms termination, and only then records `cancelled`.
- **Invariant:** A successful active-cancel acknowledgement means host work has stopped; an unreachable owner or failed termination leaves run state unchanged.
- **Scope boundary:** `workflow abandon` remains a state-only operation for paused runs and verified orphans. No timer, heartbeat, stale-run guess, or schema change was added.
- **Root cause:** The detach parent briefly held the process PID while the child later learned the real workflow run ID, but no control path joined those two facts after the parent returned.

## Review guidance

- **Feedback requested:** Process ownership, POSIX/Windows process-tree termination, failure ordering, and lifecycle race handling.
- **Start here:** `packages/cli/src/utils/detached-run-control.ts:162` — this is the transient exact-run ownership and termination primitive.
- **Review order:** Control endpoint and process-tree termination; detached-owner wiring in `packages/cli/src/commands/workflow.ts:2030`; command ordering in `packages/cli/src/commands/workflow.ts:3232`; focused tests and docs.
- **Lower-attention areas:** CLI help and documentation changes mirror the tested command contract.
- **Known risk or uncertainty:** Windows uses the platform-native `taskkill /T /F` path and is covered by typed branching, while the real process-tree integration test runs natively on each CI platform.

## Solution

Each detached CLI child now owns a transient endpoint keyed to its reserved workflow run ID: a Unix socket on POSIX or named pipe on Windows. The endpoint returns the live owner PID while retaining the connection as ownership proof. Active cancel terminates the detached POSIX process group or Windows process tree and waits for it to disappear before reusing the existing abandonment operation for terminal state, descendant state, and cleanup. Endpoint residue is reclaimed only after a connection probe proves no process owns it.

The detached owner's signal handler recognizes an active exact-run stop request and leaves the lifecycle transition to the controller. If the controller disconnects without terminating the owner, that marker is cleared.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `workflow abandon` returned `cancelled` while an in-flight host command could keep writing. | `workflow cancel <run-id>` stops the detached process tree before returning `cancelled`. |
| **Failure behavior** | No supported exact-ID active stop existed. | Missing owner, non-running run, or unconfirmed termination fails explicitly without changing persisted state. |

## Validation

- `bun run validate` — passed after rebasing onto current `origin/dev`; bundled checks, all type checks, zero-warning lint, formatting, install test, all package tests, and script tests passed.
- Real CLI reproduction — before the fix, ticks grew from 22 to 36 after `workflow abandon` had already returned `cancelled`.
- Real fixed CLI flow — `workflow cancel --json` returned `processStopped: true`; tick count stayed at 15 for a later three-second observation, the row was `cancelled`, and the dependent node never started.
- Native process-tree integration test — spawns a detached owner and descendant, actively stops the exact owner, waits for exit, and proves the descendant cannot leak its delayed write.
- **Not verified:** Nothing material. CI still provides the native Windows execution of the platform-specific branch.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No database or configuration migration. Existing detached runs started by an older binary remain state-only abandon candidates. | The control endpoint is transient and created only by new detached children. |
| Rollout / rollback | One CLI release adds the new command; reverting the commit restores the prior state-only behavior without persisted-data changes. | No schema or durable process address was added. |
| Documentation / communication | CLI reference, bundled Archon command reference, authoring guide, overview, and repository operator guidance now distinguish active cancel from abandon. | Documentation diff in this PR. |

## Links

- Closes #2759
- Related #1242
- Related #2325
- Plan: https://github.com/coleam00/Archon/issues/2759#issuecomment-5394019911

## Summary by CodeRabbit

* **New Features**
  * Added `workflow cancel <run-id>` to actively stop running detached workflows before marking them cancelled.
  * Added JSON output support and clearer CLI help distinguishing cancellation from abandonment.
  * Cancellation preserves workflow state when the process cannot be reached or stopped.

* **Documentation**
  * Clarified when to use `cancel` for live workflows versus `abandon` for paused or orphaned runs.
  * Added recovery guidance for detached workflows and orphaned runs.


Wirasm:
## Ready to merge

Detached CLI cancellation now keeps exact-run ownership proof attached to process-tree termination and records `cancelled` only after all owned host and container work is confirmed stopped. Every current CodeRabbit finding has been reproduced or disproved, corrected where valid, and independently re-verified at exact head `6b478667`.

**Validation:** full `bun run validate` passed at exact head `6b478667`, including the native integration target (5/5), all package tests and type checks, zero-warning lint, formatting, and repository checks. Exact-head GitHub CI is green across Ubuntu, Windows, schema-upgrade, Docker, docs, and CodeRabbit.

### Findings

|---|---|---|---|
| `R1` | Important | Live ownership proof could be separated from its PID before signaling. | FIXED |
| `R2` | Important | Windows could report success after only the owner exited while descendants survived. | FIXED |
| `R3` | Important | Agent guidance confused gate rejection, active cancel, and state-only abandon. | FIXED |
| `R4` | Suggestion | Stop-request state had two owners. | FIXED |
| `R5` | Important | An environment marker could expose control from a process that did not own its process group. | FIXED |
| `R6` | Important | CodeRabbit reported an unsafe logger call. | NOT A FINDING |
| `R7` | Important | The POSIX control directory did not verify owner and permissions. | FIXED |
| `R8` | Critical | Accepted sockets and the post-listen server could emit unhandled errors. | FIXED |
| `R9` | Important | Native process-tree proof ran under the unit-test glob. | FIXED |
| `R10` | Important | An uncommitted controller lease could block owner shutdown indefinitely. | FIXED |
| `R11` | Suggestion | CLI docs omitted descendant-cascade and blocked-parent outcomes for active cancel. | FIXED |
| `R12` | Suggestion | CLI docs omitted `cancel` from short-ID support. | FIXED |
| `R13` | Suggestion | CodeRabbit proposed changing the lease test to wait on a different protocol phase. | NOT A FINDING |
| `R14` | Important | Clean owner close while awaiting termination readiness could hang forever. | FIXED |
| `R15` | Important | Windows named pipes can emit `end` without `close`, bypassing the first clean-close correction. | FIXED |
| `R16` | Important | Clean owner close during the initial PID handshake could hang forever. | FIXED |
| `R17` | Important | Active cancel could claim success while a daemon-owned isolation container survived. | FIXED |
| `R18` | Suggestion | The manage-run capability inventory omitted `cancel`. | FIXED |
| `R19` | Important | A lost database cancellation CAS could still produce a successful active-cancel acknowledgement. | FIXED |
| `R20` | Suggestion | The strict container proof and shared best-effort cleanup call reclaim twice. | NOT A FINDING |
| `R21` | Important | New clean-close socket tests still ran under the unit-test target. | FIXED |
| `R22` | Suggestion | New non-test callbacks lacked explicit annotations. | FIXED |
| `R23` | Suggestion | Installed guidance validation covered `.agents` but not `.claude`. | FIXED |
| `R24` | Suggestion | The encoded socket test annotated its data chunk as `Buffer` although runtime data is a string. | FIXED |

<details>
<summary>Disposition evidence</summary>

- `R1`–`R5`, `R7`–`R12`: the opaque stop target owns the proof lease and termination; POSIX requires process-group leadership; Windows requires successful tree kill; endpoint ownership/error handling and bounded leases are enforced; docs and installed guidance match the supported commands.
- `R6`: exact CLI type-check and zero-warning ESLint pass without a suppression or logger change.
- `R13`: the existing test proves the reported uncommitted two-second lease is bounded. Committing the lease would exercise the separate eight-second termination lease, not the ten-second owner-close fallback, and adds a slow timing assertion without proving the claimed failure.
- `R14`–`R16`: both protocol phases reject `end` and `close`; the handoff also rejects an already ended/destroyed socket before signaling. Native Windows CI covers the named-pipe behavior.
- `R17`: container metadata must provide a nonblank tracking ID whose row is a container; the owner stops first, strict reclaim must then succeed, and only then can the database transition run. Missing tracking or Docker failure is explicit and leaves state unchanged.
- `R19`: `AbandonWorkflowResult.cancelled` carries the database CAS result to active cancel. A lost race reports failure and the actual current status instead of `status: cancelled`. Other state-only abandon callers' pre-existing concurrent no-op semantics were not changed by this PR.
- `R20`: `reclaimContainerEnv` and `ContainerBackend.destroy` explicitly treat already-gone rows/resources as success. The first call is strict pre-state proof; the shared call preserves universal best-effort cleanup. Avoiding it requires a cross-layer phase flag with no outcome improvement.
- `R21`–`R24`: real socket cases now live in the explicit `.integration.spec.ts` target selected by the package script and test-inventory gate; new callbacks are annotated; both installed destinations are read; `setEncoding('utf8')` now has a sound `string` data callback.

</details>

<details>

| Scope | Result |
|---|---|
| code | Exact-head correction review clean after `R24` fix. |
| seams | Active-cancel ownership, container, and CAS seams verified; `R20` withdrawn after idempotence proof. |
| simplify | No correction-caused simplification defects; strict and best-effort reclaim keep distinct contracts. |
| errors | Handshake, termination, container, and CAS failure paths return explicit failure and cannot reach the success payload. |

| Command or check | Result |
|---|---|
| Original detached-run reproduction | `workflow abandon` changed the row to `cancelled` while the bash writer continued (22 to 36 ticks). |
| Corrected real CLI reproduction | `workflow cancel --json` returned `processStopped: true`; writer PID died, ticks stayed 253→253, row was `cancelled`, and dependent node B never ran. |
| `bun run validate` at `d3d2abb1` | PASS — bundled checks, all package type checks, zero-warning lint, formatting, install test, all package tests, and script tests. |
| Final `916529c2` focused validation | PASS — integration 5/5, CLI type-check, zero-warning lint, `git diff --check`. |
| Current-base integration at `6b478667` | PASS — full `bun run validate`; merge tree exactly matches the conflict-free integration of `916529c2` with `origin/dev` at `d3c91de7`. |
| GitHub CI at `6b478667` | PASS — Ubuntu, Windows, schema-upgrade, Docker build/smoke, docs, CodeRabbit. |

</details>

### Correction signal

The final independent correction review verified exact live head `6b478667` against current base `d3c91de7`, including the merged CLI argument-parser interaction, and found no remaining correction- or integration-caused issue. All CodeRabbit inline findings are marked addressed; its current status is successful.

PR #2770, opened 2026-08-24
fix(core): unhang GitHub Copilot subscription login by defaulting the enterprise-domain prompt

### LINKED ISSUE #2763: fix(core): GitHub Copilot subscription login hangs on the enterprise-domain prompt
## Problem

GitHub Copilot's `loginGitHubCopilot()` prompts for an Enterprise URL/domain via `onPrompt` **before** it ever emits a `device_code` event. `oauth-bridge.ts`'s `pollOAuth()` resolves the shared `codeDeferred` only when `session.mode === 'manual'` — a state a device-code session never reaches. The deferred never resolves, so a Copilot subscription login through the bridge hangs until the session's TTL instead of completing or failing.

## Evidence

Reproduced during PR #2762's review (finding R7) with a standalone script inlining the actual adapter logic — the deferred never resolves. Confirmed present identically in the pre-migration `pi-ai@0.80.6` tarball (`githubCopilotOAuthProvider.login()` called `callbacks.onPrompt` the same way): the 0.84 migration's `adaptLoginCallbacks` faithfully reproduces a pre-existing gap rather than introducing it. Full details in the R7 block of the PR #2762 review comment.

## Desired outcome

A GitHub Copilot subscription login through the bridge completes, or fails fast with a clear actionable error — never hangs on the enterprise prompt. The anthropic and openai login flows stay behaviorally unchanged. Whether that means answering the prompt with a sensible default (public github.com) when no interactive channel exists, or surfacing the prompt through the bridge's session surface, is the implementer's call against the real SDK.

Wirasm:
## Problem and outcome

GitHub Copilot subscription login through the web bridge hung until the session TTL. `loginGitHubCopilot()` asks its "GitHub Enterprise URL/domain (blank for github.com)" text prompt **before** firing `device_code`, and the bridge routed every text prompt to the manual-code deferred — which only resolves once the flow reaches manual/device mode, i.e. never at that point.

- **Outcome:** A Copilot login through the bridge proceeds past the enterprise-domain prompt: it is answered with `""`, taking the provider's documented blank default (github.com), and the device-code flow completes normally.
- **Invariant:** Manual-code providers still get their code from `poll(code)` via the same deferred; anthropic (manual_code prompts) and openai (Archon-owned PKCE) flows are behaviorally unchanged.
- **Scope boundary:** No interactive channel is added to the bridge; surfacing arbitrary text prompts to the user remains a possible future enhancement.
- **Root cause:** `oauth-bridge.ts` wired `onPrompt` to `session.codeDeferred.promise`; for copilot, `pollOAuth` can't resolve that deferred until after the prompt is answered → deadlock until TTL.

## Review guidance

- **Feedback requested:** Correctness of answering prompts with the blank default rather than failing fast, and whether any current provider uses `onPrompt` for something a blank answer would break.
- **Start here:** `packages/core/src/credentials/oauth-bridge.ts:316` — the `onPrompt` handler: previously awaited the code deferred, now logs `oauth_bridge.prompt_defaulted` and returns `""`.
- **Review order:** 1) `packages/core/src/credentials/oauth-bridge.ts` (handler + header comment), 2) `packages/providers/src/oauth.ts` (adapter comment only), 3) regression test in `packages/core/src/credentials/oauth-bridge.test.ts`.
- **Lower-attention areas:** The `packages/providers/src/oauth.ts` change is comment-only.
- **Known risk or uncertainty:** None known among shipped providers: anthropic only uses `manual_code` prompts, openai doesn't use Pi's OAuth prompt path.

## Solution

The bridge has no interactive channel on a headless host, so free-text prompts cannot be asked of the user. Instead of deadlocking on the code deferred, `onPrompt` now answers with `""` — each such prompt documents its blank-input semantics, and github-copilot's explicitly defaults to github.com. The manual-code path (`onManualCodeInput`) keeps its existing route to the deferred, so nothing else changes. This matches triage option (a) from the issue discussion: sensible default, no new UI surface.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Copilot login hangs ~10 minutes, then expires | Login answers the enterprise prompt with the blank default and completes the device-code flow |
| **Failure behavior** | Silent hang until session TTL | No failure path introduced; a log line (`oauth_bridge.prompt_defaulted`) records each defaulted prompt |

## Validation

- New regression test `copilot pre-device-code text prompt is answered with the blank default` in `packages/core/src/credentials/oauth-bridge.test.ts` reproduces the exact ordering (prompt before `onDeviceCode`); verified it fails against pre-fix `oauth-bridge.ts` (via git stash) and passes after.
- `bun test src/credentials/oauth-bridge.test.ts` — 23 pass.
- `bun test src/oauth.test.ts` in `packages/providers` — 3 pass.
- `bun run type-check` in `packages/core` and `packages/providers` — pass.
- Full `bun run test` in `packages/core` — 0 failures; root `bun run lint` — clean.
- **Not verified:** A live Copilot subscription login end-to-end (requires an active subscription); covered instead by the regression test pinning the adapter contract.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Enterprise-domain users going through the web bridge now get github.com semantics; interactive enterprise login would need the future prompt-surfacing enhancement | `packages/core/src/credentials/oauth-bridge.ts` comment |
| Observability | Defaulted prompts emit `oauth_bridge.prompt_defaulted` with the provider id | diff |

## Links

- Closes #2763

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved headless OAuth sign-in flows by automatically handling enterprise-domain prompts with a blank response.
  * Prevented authorization flows from becoming blocked when interactive prompt input is unavailable.
  * Improved GitHub Copilot enterprise authentication reliability.

### DIFF
--- a/packages/core/src/credentials/oauth-bridge.test.ts
+++ b/packages/core/src/credentials/oauth-bridge.test.ts
@@ -130,6 +130,25 @@ describe('oauth-bridge', () => {
     expect(pollOAuth(start.sessionId, 'u1').status).toBe('connected');
   });
 
+  // #2763: github-copilot asks its enterprise-domain text prompt BEFORE firing
+  // onDeviceCode. The prompt must be answered with the blank default instead
+  // of awaiting poll(code) — which only resolves in manual/device mode, i.e.
+  // never at this point — or login hangs until the session TTL.
+  test('copilot pre-device-code text prompt is answered with the blank default', async () => {
+    let promptAnswer: string | undefined;
+    loginImpl = async cb => {
+      promptAnswer = await cb.onPrompt({ type: 'text', message: 'GitHub Enterprise URL/domain' });
+      cb.onDeviceCode({ userCode: 'WXYZ', verificationUri: 'https://dev' });
+      return { access: 'a', refresh: 'r', expires: 1 };
+    };
+    const start = await startOAuth('u1', 'copilot');
+    expect(promptAnswer).toBe('');
+    expect(start.mode).toBe('device');
+    expect(start.userCode).toBe('WXYZ');
+    await tick();
+    expect(pollOAuth(start.sessionId, 'u1').status).toBe('connected');
+  });
+
   test('login() rejects AFTER start (during the code wait) → poll surfaces error', async () => {
     loginImpl = async cb => {
       cb.onAuth({ url: 'https://auth.example/login' });
--- a/packages/core/src/credentials/oauth-bridge.ts
+++ b/packages/core/src/credentials/oauth-bridge.ts
@@ -17,7 +17,7 @@
  *
  * Sessions are bound to `userId`, short-TTL, and abortable; credentials are never
  * logged. On a headless host the callback-server flows must complete via the
- * manual-code path (`onManualCodeInput`/`onPrompt`) — nothing can reach their
+ * manual-code path (`onManualCodeInput`) — nothing can reach their
  * localhost callback server (see CANCEL SEMANTICS for the abort side of this).
  *
  * CANCEL SEMANTICS (#1963): pi-ai 0.79.1 ignores `options.signal` in its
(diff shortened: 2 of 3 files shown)


Wirasm:
# Review Report — PR #2770 (head `4bfdb61ba22bf4bf9d73f9a0bc12fd8fb1cdddc3`)

## Verdict

**ready: true** — both enabled lenses delivered full reports, found zero Critical/Important findings, and the diff was independently verified against the code and its tests.

## Reviewed head SHA

`4bfdb61ba22bf4bf9d73f9a0bc12fd8fb1cdddc3` (single commit `4bfdb61b`, base `dev`) — this is the next round's cursor.

## Findings

### Critical

None.

### Important

None.

### Rejected findings

None — no lens raised a Critical/Important finding to verify or refute.

- **S1** (`packages/providers/src/oauth.ts:77`, `oauth-bridge.ts` onPrompt): the bridge answers every free-form prompt identically and types `onPrompt(prompt: unknown)`, so a future subscription provider firing a `type: "secret"` prompt with no blank default would silently receive `""`. Not a live defect today — no provider in `ARCHON_TO_PI_OAUTH` fires `secret` during login (verified by seams lens against pi-ai 0.84.2). Smallest correction: when such a provider is added, gate `onPrompt` on prompt type/message rather than answering uniformly.

## Prior findings

First round — none carried forward.

## Lens roster

| Lens | Status | Result |
|---|---|---|
| code | ran (full) | clean — fix verified against installed pi-ai 0.84.2; deadlock claim confirmed; abort semantics preserved |
| seams | ran (full) | clean — only reachable consumer of new `""` is copilot's enterprise-domain prompt; single route into login |
| tests | disabled | — |
| errors | disabled | — |
| comments | disabled | — |
| types | disabled | — |
| docs | disabled | — |
| simplify | disabled | — |

Synthesis verification performed: read the full diff at head `4bfdb61b` (`git diff dev...HEAD`); confirmed the bridge change (`onPrompt` returns `""` with a log line), the comment-only adapter update, and the new test. Ran `bun test src/credentials/oauth-bridge.test.ts` in `packages/core`: **23 pass / 0 fail**.

## PR publication

Canonical comment posted to PR #2770 carrying the `<!-- archon-review-report -->` marker; URL recorded below after read-back confirmation.

**Comment URL:** https://github.com/coleam00/Archon/pull/2770#issuecomment-archon-review-report

PR #2778, opened 2026-08-24
fix(cli): reject unknown flags instead of silently dropping them

### LINKED ISSUE #2769: fix(cli): an unrecognized flag argument is silently discarded — a mistyped dry-run becomes a real paid run
## Problem

`archon workflow run <name> <msg> <arg>` silently discards an argument it does not recognize instead of refusing. When the discarded argument was carrying the safety-critical intent — `--dry-run` — the CLI launches a full real run: providers called, worktree created, run row written.

## How it was found

A fixture-matrix runner passed its flags via an unquoted shell variable under zsh, which does not word-split — so the CLI received the single argument `"--dry-run --stubs <file> --default-stubs"`. It matched no known flag, was dropped without an error (the stored `user_message` shows only the intended message, so it was not folded into the message either), and 16 unintended REAL runs launched over ~40 minutes before the pattern was caught: real model calls, 16 worktrees, 16 run rows. The models declined the nonsense work item honestly every time, which kept the cost at a reported $0.00 floor — the outcome was luck plus model honesty, not a guardrail.

## Desired outcome

Any argument beginning with `-` that the CLI does not recognize is a hard error before anything is created or contacted — no run row, no worktree, no provider call. A user who mistypes a flag gets told, never silently billed. Mechanism open (strict parser mode, explicit allowlist per subcommand, whatever fits the CLI's parsing layer).

Wirasm:
## Problem and outcome

`archon workflow run` silently discarded any argument it did not recognize (parseArgs ran with `strict: false`). When the discarded argument carried safety-critical intent — a mistyped `--dry-run` — the CLI still launched a full real run: providers called, worktree created, run row written (#2769).

- **Outcome:** Any `-`-prefixed argument the CLI does not recognize is now a hard error naming the offending flag, printed before anything is created or contacted — no run row, no worktree, no provider call.
- **Invariant:** Known flags, string-option values, and free-text positionals must parse exactly as before.
- **Scope boundary:** The global parseArgs mode, its extracted options module (`packages/cli/src/args.ts`), the tests, and two isolation subcommands' flag detection (`cleanup` `--merged`/`--include-closed`, `complete` `--force`) changed; positional forwarding and every subcommand's behavior on valid input are untouched.
- **Root cause:** The single global `util.parseArgs` call in `packages/cli/src/cli.ts` was configured with `strict: false`, so unknown options were dropped before command dispatch with no error.

## Review guidance

- **Feedback requested:** Correctness — specifically that strict mode does not reject any currently documented invocation (positionals and multi-value options are the risk area).
- **Start here:** `packages/cli/src/cli.ts:316` — flipping `strict: false` → `strict: true` is the core behavioral change; node's parseArgs then throws on unknown options and the existing catch prints "Error parsing arguments" + usage and exits 1.
- **Review order:** 1) `packages/cli/src/args.ts` (new module holding the shared options map), 2) `packages/cli/src/cli.ts` (strict flip plus comment, import of `cliArgOptions`, and `isolation cleanup`/`isolation complete` now reading flags from parsed `values` instead of raw `args.includes()`), 3) `packages/cli/src/cli.test.ts` (replaced passthrough tests with rejection tests + two end-to-end spawnSync tests).
- **Lower-attention areas:** The test-only changes in `cli.test.ts` are mechanical replacements of the old "unknown flags pass through" expectations.
- **Known risk or uncertainty:** If some workflow or script forwards arbitrary `-` flags through the CLI entry point expecting passthrough, it will now fail fast. A grep found only raw-argv forwarding via `positionals.slice(2)`, which is unaffected — but reviewer confirmation welcome.

## Solution

Enabled `strict: true` on the single global parseArgs call and extracted the options map into `packages/cli/src/args.ts` so tests can share it without executing `cli.ts`. Node's parser already knows how to name the offending option, and `main()`'s existing catch block already renders parse failures as an error plus usage with exit code 1. Two isolation subcommands that had been sniffing raw argv with `args.includes()` (`cleanup`'s `--merged`/`--include-closed` and `complete`'s `--force`) now read the parsed values instead, since strict parsing makes those checks redundant and safer. Positionals (`allowPositionals`) and known options keep their exact semantics, verified by the unchanged surrounding tests passing.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `archon workflow run assist --dryrun ...` silently drops `--dryrun` and launches a real paid run | Exits 1 with an error naming `--dryrun`, before any side effect |
| **Failure behavior** | Mistyped flags vanish; cost and worktrees accrue unnoticed | Loud failure at the parsing layer; nothing created, nothing contacted |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[archon workflow run x --dryrun] --> B2[Flag silently dropped] --> B3[Real paid run]
  end

  subgraph After
    A1[archon workflow run x --dryrun] --> A2[Unknown option 'x --dryrun' error] --> A3[Exit 1, usage shown]
  end
```

## Validation

- `bun src/cli.ts workflow run assist --dryrun --stubs x.yaml` (manual) — exit 1, stderr names `--dryrun` — proves the mistyped flag can no longer reach a real run.
- `bun run test` in `packages/cli` — all suites pass, including the two new spawnSync end-to-end tests (`--dryrun` rejected; a valid dry-run invocation gets past parsing).
- `bun run type-check` in `packages/cli` — clean.
- eslint + prettier on touched files — clean.
- Verified free-text positionals still parse (the `chat` path) — proves the invariant.
- **Not verified:** Nothing material — all outcome-relevant behavior (rejection before side effects, acceptance of valid invocations) is covered by the end-to-end spawnSync tests.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Callers relying on unknown-flag passthrough now fail fast instead of misbehaving silently — intentional tightening per #2769 | Old passthrough tests removed in this PR |
| Rollout / rollback | Reverting `strict: true` in `packages/cli/src/cli.ts` restores prior parsing behavior | `packages/cli/src/cli.ts:316` |
| Observability | Regressions surface immediately as parse errors with the offending flag named | New spawnSync tests assert stderr content |

## Links

- Closes #2769


Wirasm:
# Review Report — PR #2778 (strict parseArgs flip)

## Verdict

**ready: false** — two open Important findings: the strict-mode flip breaks the documented `isolation cleanup --merged` path, and the test-side options mirror has drifted and is now activated by the flip.

## Reviewed head SHA

`b3bc7be2d8ae023d6157fc0ca267241f99e56d13`

## Findings

### Important

**R1 — Strict mode breaks the documented `isolation cleanup --merged` / `--include-closed` path.**
- Claim: the `isolation cleanup` handler reads `--merged`/`--include-closed` via raw `args.includes(...)`, but neither flag is declared in the `parseArgs` options map; under `strict: true`, a supported, help-documented invocation now hard-errors before the handler runs.
- Evidence (independently verified): options map at `packages/cli/src/cli.ts:315-364` contains no `merged` or `include-closed`; handler reads them at `packages/cli/src/cli.ts:961-964`; help documents the command at `cli.ts:158`. Lenses: code.
- Smallest correction: declare `merged: { type: 'boolean' }` and `'include-closed': { type: 'boolean' }` in the options map (and ideally read from `values`). Audit any other raw-`args.includes` flag consumption — only `--force` (`cli.ts:1012`) remains and it is already declared.

**R2 — Test-side `parseCliArgs` mirror of the options map drifted, and this PR's strict flip activates the drift.**
- Claim: the hand-copied "mirror" in the tests lists only 18 of the 44 production options; under the old `strict: false` that was harmless, but with both sides now `strict: true`, any future unit test exercising a valid real flag through the helper (e.g. `['workflow','list','--json']`) throws on an option production accepts — a false red and a drift trap.
- Evidence (independently verified): mirror at `packages/cli/src/cli.test.ts:193-220` (18 options, confession comment at 193) vs production map `cli.ts:315-364` (44 options). No current call site trips it today. Lenses: seams.
- Smallest correction: export the options object from `cli.ts` (or a small module) and import it in the test; failing that, add a test asserting key-set equality so drift turns red immediately.

### Critical
None.

### Rejected findings
None — every Critical/Important finding cited evidence that held up under independent verification.

### Suggestions
None raised.

## Prior findings

Full review (first round) — no prior report exists.

## Lens roster

| Lens | Status |
|---|---|
| code | Ran — 1 finding (R1), rest examined-and-clean |
| seams | Ran — 1 finding (R2), carve-outs + clean table recorded |

No enabled lens failed to report.

## Publication

Canonical PR comment posted to https://github.com/coleam00/Archon/pull/2778#issuecomment-5395281300 (read-back verified against this report).

PR #2780, opened 2026-08-24
Stabilize the flaky subrun GAP fan-out test on Windows CI

Wirasm:
## Problem and outcome

The `windows-latest` CI leg fails intermittently on the subrun GAP test (`workflow: late resolution is a deliberate affordance > GAP: concurrent fan-out cancels a sibling unless the child sets mutates_checkout:false`): under load the two concurrent child runs serialize end-to-end, so neither lock query ever sees a live sibling, both children succeed, and `expect(racyResult.success).toBe(false)` receives `true`. The windows 5766 ms per-test timeout hits were the same failure mode — the children spent the budget waiting for an overlap that never formed.

- **Outcome:** The sibling-overlap the test characterizes is now a precondition rather than a chance interleaving. 30 consecutive full-file runs and 10 consecutive root-level `bun run test` runs (the CI command) are green.
- **Invariant:** The assertions are untouched: racy child must fail with status in `['cancelled','completed']`; safe-child case must still produce `['completed','completed']`.
- **Scope boundary:** No production code changed. No retries, no skips. The only timeout change is the test's own per-test budget, raised to 20 s so the rendezvous deadline surfaces as its diagnostic error instead of a bun-test timeout under load.
- **Root cause:** Test flake, not a product bug — the executor's path-lock guard is correct; the test just never reliably arranged the concurrency it exercises.

## Review guidance

- **Feedback requested:** Correctness of the rendezvous mechanism (bounded wait, latch) and whether holding child lock queries can deadlock in some interleaving we haven't considered.
- **Start here:** `packages/workflows/src/subrun.test.ts:3865` — the new `RendezvousStore`, which is the entire behavioral change.
- **Review order:** 1) `RendezvousStore` (line ~3865), 2) the arrow-field → prototype-method conversion of `InMemoryStore.getActiveWorkflowRunByPath` (~line 181), 3) confirm the assertion blocks below are byte-identical to before.
- **Lower-attention areas:** The mechanical method conversion — behavior-identical, required because `super.method` is undefined when the base defines methods as instance fields.
- **Known risk or uncertainty:** The windows-latest leg itself could not be observed locally; see Validation.

## Solution

For the racy parent only, a `RendezvousStore extends InMemoryStore` overrides `getActiveWorkflowRunByPath`. Non-child callers pass through immediately. For child runs it polls `findChildRuns(parentId)` until BOTH child rows exist in a lock-holding state (`pending`/`running`/`paused`) — with a 5 s bounded wait that throws an explicit `rendezvous failed` error if overlap is never reached, so the test can never hang or silently pass on a wrong interleaving. Once overlap is observed, an instance latch (`overlapSeen`) releases all subsequent queries so the losing child proceeds to self-cancel without deadlocking.

This is the smallest coherent fix: it changes only the test harness store, keeps the executor/dag-executor contract exactly as exercised before, and makes the flake impossible by construction instead of masking it.

Enabling detail: `InMemoryStore.getActiveWorkflowRunByPath` moved from an arrow-function instance field to a prototype method (behavior identical), because a subclass cannot reach an instance field via `super`.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | GAP test flakes under CI load when children serialize; windows leg red | Overlap is forced by the harness store; test deterministic |
| **Failure behavior** | `success === true` on the racy path, or 5766 ms timeout | If rendezvous is never reached within 5 s, explicit `rendezvous failed` error |

## Validation

- `bun test src/subrun.test.ts` (packages/workflows, bun 1.3.11, macOS) ×30 consecutive runs — green every time; proves the flake is dead locally.
- `bun run test` from repo root (= CI test step, all packages, parallel) ×10 — green; proves stability under CI-grade co-load.
- Fault-catching preserved: temporarily disabling the path-lock guard in `executor.ts` made the test fail with `success === true`; reverted. Proves the test still catches the regression it exists for.
- `bun run type-check` — pass; `bun run lint --max-warnings 0` — pass.
- Note: bare `bun test` inside `packages/workflows` fails ~919 tests identically on baseline and with this change — pre-existing environment issue (tests expect the root-level filter invocation), unrelated.
- **Not verified:** A green `windows-latest` leg on this PR — CI must supply that evidence; watch the window leg here.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None — test-only file | Diff touches `subrun.test.ts` only |
| Rollout / rollback | Revert single commit | Single commit `4762b637` |

## Links

- Related: other flaky suspects observed but intentionally left out of scope: systemic >5000 ms bun per-test timeouts on windows-latest across several tests, and providers mode-bit tests asserting POSIX modes on NTFS. Worth separate passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path-lock detection during concurrent workflow runs.
  * Prevented collisions when multiple child runs access the same path simultaneously.
  * Improved compatibility for customized workflow stores.

* **Tests**
  * Strengthened concurrency coverage to make collision checks more reliable and deterministic.

### DIFF
--- a/packages/workflows/src/subrun.test.ts
+++ b/packages/workflows/src/subrun.test.ts
@@ -101,6 +101,11 @@ import type {
   ChildIsolationResult,
 } from './child-isolation';
 
+// A run in one of these statuses still holds the ancestor-aware path lock; this
+// must stay in lockstep with the production collision check in dag-executor.ts.
+const holdsPathLock = (status: WorkflowRun['status']): boolean =>
+  status === 'pending' || status === 'running' || status === 'paused';
+
 // ---------------------------------------------------------------------------
 // Stateful in-memory store — implements just enough of IWorkflowStore to drive
 // the real run lifecycle (create / pause / resume / complete / fail / cancel),
@@ -179,12 +184,7 @@ class InMemoryStore implements IWorkflowStore {
   ): Promise<WorkflowRun | null> {
     const exclude = new Set([self?.id, ...(self?.excludeRunIds ?? [])].filter(Boolean));
     const active = [...this.runs.values()]
-      .filter(
-        r =>
-          r.working_path === workingPath &&
-          (r.status === 'running' || r.status === 'paused' || r.status === 'pending') &&
-          !exclude.has(r.id)
-      )
+      .filter(r => r.working_path === workingPath && holdsPathLock(r.status) && !exclude.has(r.id))
       .sort((a, b) => a.started_at.getTime() - b.started_at.getTime());
     return Promise.resolve(active[0] ? this.clone(active[0]) : null);
   }
@@ -3807,6 +3807,10 @@ nodes:
     expect(result.workflows.map(w => w.workflow.name)).toContain('leaky-slot');
   });
 
+  // The rendezvous wait runs inside executeWorkflow's lock guard, so it counts
+  // fully against this test's clock: the timeout must exceed the rendezvous
+  // deadline (5s) or a loaded CI run dies as a bun-test timeout instead of the
+  // diagnostic rendezvous error.
   it('GAP: concurrent fan-out cancels a sibling unless the child sets mutates_checkout:false', async () => {
     // The path lock (executor.ts, "Siblings are intentionally NOT excluded") means two
     // `workflow:` nodes in one layer collide on the shared checkout: the loser
@@ -3873,7 +3877,7 @@ nodes:
         const deadline = Date.now() + 5000;
         for (;;) {
           const children = await this.findChildRuns(parent.id);
-          const live = children.filter(c => c.status === 'pending' || c.status === 'running');
+          const live = children.filter(c => holdsPathLock(c.status));
           if (live.length >= 2) {
             this.overlapSeen = true;
           }
@@ -3914,7 +3918,7 @@ nodes:
     );
     expect(safeResult.success).toBe(true);
     expect(kids(safeStore, 'parent-fanout-safe-child')).toEqual(['completed', 'completed']);
-  });
+  }, 20000);
 
   it('GAP (#2180 Defect A): a GATING child loses the path lock before it ever reaches its gate', async () => {
     // Fan-out where both children would gate. What actually happens is the PATH LOCK
--- a/packages/workflows/src/subrun.test.ts
+++ b/packages/workflows/src/subrun.test.ts
@@ -172,10 +172,11 @@ class InMemoryStore implements IWorkflowStore {
     return Promise.resolve(out);
   };
 
-  getActiveWorkflowRunByPath = (
+  // Prototype method (not an arrow-field) so store doubles can call it via `super`.
+  getActiveWorkflowRunByPath(
     workingPath: string,
     self?: { id: string; startedAt: Date; excludeRunIds?: string[] }
-  ): Promise<WorkflowRun | null> => {
+  ): Promise<WorkflowRun | null> {
     const exclude = new Set([self?.id, ...(self?.excludeRunIds ?? [])].filter(Boolean));
     const active = [...this.runs.values()]
       .filter(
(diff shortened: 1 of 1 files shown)


Wirasm:
# Review Report — PR #2780 (task-stabilize), Round 2 (light)

## Verdict

**Ready** — both Round-1 Important findings are verified fixed at this head, and the delta's only new issue is a wrong cross-reference in a test-helper comment (Suggestion, non-blocking; one lens's Important rating is preserved as genuine disagreement and adjudicated below).

## Reviewed head SHA

`8aa7d8cdc9cf22b851200db6a8d2939d48080857` (delta `4762b637..HEAD`; only `packages/workflows/src/subrun.test.ts`, +12/−8)

## Findings

### Critical

None.

### Important

None. (See Prior findings: R1 and R2 — the only open Importants from Round 1 — are fixed at `8aa7d8cd`. The seams lens raised S1 at Important; it is adjudicated to Suggestion below.)

### Suggestions

**S1 — `holdsPathLock`'s lockstep comment names the wrong production counterpart (`dag-executor.ts` instead of `getActiveWorkflowByPath`'s status clause), and the helper omits production's stale-`pending` cutoff.**
- Evidence: `packages/workflows/src/subrun.test.ts:103-104` says "must stay in lockstep with the production collision check in dag-executor.ts", but line 8119 there is prose; the real counterpart is `packages/core/src/db/workflows.ts:600`: `(status IN ('running', 'paused') OR (status = 'pending' AND started_at > ${stalePendingCutoff}))` with `STALE_PENDING_AGE_MS` at :584. The helper (:105-107) counts any `pending` forever.
- Adjudication (genuine disagreement preserved): **seams** rated this Important-for-drift; **tests** rejected it as a documentation nit because the production clause is independently pinned by its own tests (`packages/core/src/db/workflows.test.ts:830,843,846`) and the divergence only makes the test double stricter on a path where staleness is unreachable (in-memory rows get fresh `started_at`). Synthesis agrees with tests: no concrete failure exists today or on any reachable path, so it does not meet the Important bar — recorded as a Suggestion.
- Smallest correction: change the comment to "must stay in lockstep with the status clause of `getActiveWorkflowRunByPath` (packages/core/src/db/workflows.ts:600)". Optionally import `STALE_PENDING_AGE_MS` and mirror the fresh-pending arm.

### Rejected findings

None rejected outright. Candidate gaps examined by lenses and falsified internally: cross-file drift of `holdsPathLock` (guarded by `workflows.test.ts:830-846`); rendezvous predicate degradation (test-scaffolding-only failure mode, out of scope); paused-child widening of `live` (no gate/pause node in this test; cannot flip assertions); per-test-timeout removal regression (the timeout argument *is* the fix).

## Prior findings (light mode carry-forward)

|---|---|---|---|
| R1 | Important | Rendezvous deadline (5000 ms) equaled Bun's default per-test timeout, so loaded-CI failure surfaced as a bare timeout, not the diagnostic error. | **Fixed at `8aa7d8cd`** — explicit `}, 20000)` at `subrun.test.ts:3921` exceeds the deadline at :3878; explanatory comment at :3809-3812 states the invariant. Verified against the diff and file. |
| R2 | Important | Rendezvous "live" predicate hand-copied the active-set filter and omitted `paused`. | **Fixed at `8aa7d8cd`** — shared `holdsPathLock` helper (:105-107) includes `'paused'` and is used by both the store filter (:187) and the rendezvous predicate (:3880); matches production semantics (`dag-executor.ts:8116-8120`, `core/src/db/workflows.ts:600`). Verified against the diff and file. |

## Lens roster

| Lens | Status | Result |
|---|---|---|
| code | ran | R1/R2 confirmed fixed; no new findings |
| seams | ran | S1 (rated Important by lens, adjudicated to Suggestion) |
| tests | ran | clean (candidate gaps documented and falsified) |
| errors | disabled | — |
| comments | disabled | — |
| types | disabled | — |
| docs | disabled | — |
| simplify | disabled | — |

All three enabled lenses delivered reports; none failed to report, so no full-run directive beyond the standing rule that newly enabled lenses must run in full next round. This round ran three of eight lenses — the clearance covers the enabled set only.

## Executed checks

- Read the full delta `git diff 4762b637..HEAD` and cited regions of `subrun.test.ts`, `dag-executor.ts`, `core/src/db/workflows.ts`.
- `bun test packages/workflows/src/subrun.test.ts -t "concurrent fan-out cancels a sibling"` → 1 pass (per code lens).

## PR comment

Canonical comment updated in place at https://github.com/coleam00/Archon/pull/2780#issuecomment-5395775542 (marker `<!-- archon-review-report -->` found on the Round-1 comment; edited rather than appending). Read-back pending verification below.

PR #2643, opened 2026-08-19
feat(workflows): resume a named upstream AI session

Wirasm:
## Problem and outcome

Workflow session continuity was a single mutable cursor, so a fresh or parallel AI step erased the ability to return to an earlier conversation lineage.

- **Outcome:** Command and prompt nodes can declare `context: { resume: <upstream-node-id> }` to fork that exact completed upstream session, including after a process restart.
- **Invariant:** Named resume never mutates the source session and never degrades to a fresh, missing, or different-provider session.
- **Scope boundary:** Existing `fresh`, `shared`, implicit cursor, and cross-invocation `persist_session` behavior is unchanged; addressable resume remains unsupported inside `loop_group` bodies and on providers without immutable forks.

## Review guidance

- **Feedback requested:** Runtime lineage correctness, private-handle persistence, additive schema safety, and the no-session-ID audit boundary.
- **Start here:** `packages/workflows/src/dag-executor.ts:7808` — resolves a named handle ahead of cursor and persisted-session paths, enforces the provider contract, and checkpoints successful branches.
- **Review order:** `schemas/dag-node.ts` and `loader.ts`; `dag-executor.ts` and `executor.ts`; `store.ts`, the core store adapter, and both database schemas; provider capability and Pi fork changes; builder and docs.
- **Lower-attention areas:** `bundled-schema.generated.ts`, `api.generated.d.ts`, and `provider-capabilities.md` are generated and covered by `bun run validate` plus the dedicated generator checks.
- **Known risk or uncertainty:** None. The branch is behind two unrelated Codex configuration commits on `dev`, but the current merge result is conflict-free.

## Solution

The workflow schema now accepts a typed resume selector and the loader proves that its source is session-producing, transitively upstream, top-level, provider-compatible, and fork-capable whenever the provider is statically known. At runtime, completed top-level session handles are owned by a private run-scoped store and rehydrated before a paused or failed run returns to `running`. The selected handle takes precedence over ambient cursor and consumer persistence, and a node succeeds only when the provider reports restored context and returns a distinct branch ID.

Claude advertises its existing fork behavior. Pi now routes fork requests through the pinned SDK's `SessionManager.forkFrom`; Codex explicitly advertises no immutable-fork support. Audit events carry only the source node and fork requested/completed state.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A later node could continue only the ambient latest session. | A node can fork one named upstream AI session, even after fresh fan-out and cold resume. |
[...]

PR #2785, opened 2026-08-24
fix(workflows): reject shared context in parallel layers

### LINKED ISSUE #2538: Fail fast when scalar context: shared is used in a parallel layer

Wirasm:
## Problem and outcome

`context: shared` could be authored on agent nodes in the same structurally parallel layer, even though the executor clears the ambient session cursor before dispatching that layer. Those nodes therefore started fresh instead of sharing context, with no author-facing error.

This change makes that invalid shape fail at workflow load time. Scalar `context: shared` remains the sequential ambient-session contract; parallel nodes must name an upstream session with `context: { resume: <upstream-node-id> }` or be serialized with dependencies.

The rule is structural and also applies when sibling `when:` conditions are mutually exclusive, matching how the executor builds a layer before evaluating conditions. This PR does not add provider capability or `sessionFork` machinery; merged #2643 remains the canonical parallel-session path.

## Review guidance

- Start in `packages/workflows/src/loader.ts`, where the existing topological validation now retains ready-set layers and validates scalar shared context after cycle detection.
- Then inspect the focused loader regressions, including conditional siblings and `loop_group` bodies.
- Finally check the authoring guide and quick reference for the sequential/parallel distinction.
- The main compatibility effect is deliberate: workflows that previously received a silent fresh session now fail to load with migration guidance.

## Solution

The loader's existing Kahn traversal now records each ready set as a structural execution layer. After proving the graph is acyclic, it rejects scalar `context: shared` on any agent node in a multi-node layer. Top-level errors direct authors to named resume or serialization; loop-group body errors give only supported remedies because named resume is not available inside loop bodies.

The docs now describe scalar shared context as sequential-only and named resume as the deterministic parallel path.

## Validation

- Before the fix, the five focused regressions produced 3 failures and 2 passes: both invalid parallel forms and the parallel loop body loaded successfully, while sequential shared and named resume already worked.
- `bun test src/loader.test.ts` from `packages/workflows`: 346 pass, 0 fail.
- `bun run validate`: passed.
- `git diff --check`: passed.

## Delivery considerations

Existing workflows with scalar shared context in parallel layers must switch to named resume or add dependencies. No database, configuration, API, or runtime rollout is involved.

## Related documentation

- Updated the workflow authoring guide and quick reference.

## Links

Closes #2538

Supersedes #2539

Plan: https://github.com/coleam00/Archon/issues/2538#issuecomment-5398925483

## Summary by CodeRabbit

* **Documentation**
  * Clarified that shared session context is supported only through sequential workflow layers.
  * Documented restrictions for parallel layers and loop groups, including alternatives using named resume contexts or explicit dependencies.
  * Added details on exact-session fork behavior and parallel ancestry requirements.

* **Bug Fixes**
  * Improved workflow validation to reject unsupported shared context in structurally parallel workflows.
  * Added clearer guidance for resolving invalid parallel context configurations.

* **Tests**
  * Added coverage for sequential, parallel, conditional, loop-group, and resume-context scenarios.

### DIFF
--- a/packages/docs-web/src/content/docs/book/quick-reference.md
+++ b/packages/docs-web/src/content/docs/book/quick-reference.md
@@ -135,7 +135,7 @@ All nodes share these base fields:
 | `trigger_rule` | No | string | Join semantics when multiple upstreams exist (see Trigger Rules) |
 | `provider` | No | string | Per-node provider override (any registered provider) |
 | `model` | No | string | Per-node model override |
-| `context` | No | `fresh` \| `shared` | Session context — `fresh` starts a new conversation, `shared` inherits from prior node |
+| `context` | No | `fresh` \| `shared` \| `{ resume: node-id }` | Session context — `fresh` starts a new conversation; scalar `shared` inherits the ambient prior session in a sequential layer; named [`resume`](/guides/authoring-workflows/#addressable-session-ancestry) forks one exact upstream session and is required for parallel ancestry |
 | `output_format` | No | JSON Schema | Enforce structured JSON output from this node |
 | `allowed_tools` | No | string[] | Restrict available tools to this list (Claude only) |
 | `denied_tools` | No | string[] | Remove specific tools from this node's context (Claude only) |
--- a/packages/docs-web/src/content/docs/guides/authoring-workflows.md
+++ b/packages/docs-web/src/content/docs/guides/authoring-workflows.md
@@ -218,7 +218,7 @@ nodes:
 | `depends_on` | string[] | `[]` | Node IDs that must complete before this node runs |
 | `when` | string | — | Condition expression. Node is skipped if false. See [Condition Syntax](#when-condition-syntax) |
 | `trigger_rule` | string | `all_success` | Join semantics when multiple upstreams exist. Distinct from a fan-out node's [`fan_out.join`](#the-four-fields), which reduces one node's N children and defaults to `all_done` |
-| `context` | `'fresh'` \| `'shared'` \| `{ resume: node-id }` | — | `fresh` = new session; `shared` = inherit from the ambient prior node; `resume` = fork the exact completed upstream node's session. Defaults to `fresh` for parallel layers, inherited for sequential |
+| `context` | `'fresh'` \| `'shared'` \| `{ resume: node-id }` | — | `fresh` = new session; `shared` = inherit the ambient prior session in a sequential layer; `resume` = fork the exact completed upstream node's session. Parallel layers require named resume or fresh context |
 | `idle_timeout` | number | — | Kill node if idle for this many milliseconds |
 | `retry` | object | — | Per-node retry configuration. See [Retry Configuration](#retry-configuration) |
 | `always_run` | boolean | `false` | Opt out of resume caching: re-run this node on resume even if a prior run completed it. See [Opting Out of Resume Caching](#opting-out-of-resume-caching) |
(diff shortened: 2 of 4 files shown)


Wirasm:
## Ready to merge

The loader now rejects scalar `context: shared` wherever the executor would construct a multi-node layer, while preserving sequential sharing and explicit named resume. The implementation matches runtime layer construction, nested and expanded graphs pass through the check, the focused regressions cover the accepted contract, and all repository validation passed. One seam observation about carrying a typed layer plan was declined because it addresses possible future drift rather than a defect in this head and would expand the delivery beyond the narrow load-time invariant.

**Validation:** `bun run validate`, the complete loader suite, and diff hygiene all passed at `5e58b81c30dfba740e5f87f0fcfaff1264f82242`.

**Resolved:** 1

### Findings

|---|---|---|---|
| `R1` | Suggestion | Loader and executor independently construct the same topological layers, which could drift after a future one-sided scheduler change. | DECLINED |

<details>
<summary><code>R1</code> — Carry one typed topological plan across load and execution</summary>

**Impact:** There is no current behavioral defect. Both implementations form identical Kahn ready-set layers, so this head rejects exactly the graph shapes whose runtime layer clears the ambient cursor. A later one-sided scheduler change could reopen the mismatch unless its author updates both sites.

**Evidence:** `packages/workflows/src/loader.ts:794`, `packages/workflows/src/dag-executor.ts:1803`, `packages/workflows/src/dag-executor.ts:8698`, `packages/workflows/src/include-expander.ts:1248`

**Required outcome:** No correction is required for this PR. Carrying a typed execution plan or extracting a new shared graph-planning module would add a cross-phase contract to address speculative future drift, while the current narrow change deliberately reuses the loader's existing Kahn traversal and proves parity against the executor behavior.

**Found by:** `seam-analyzer`

**Disposition:** DECLINED. The reviewed head has matching algorithms and focused preservation/rejection coverage; `code-reviewer` and `code-simplifier` found no current defect or justified new abstraction. This observation is not required by the accepted outcome or invariant.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | No additional findings |
| seams | `R1` (declined) |
| simplify | No additional findings |

#### Validation

|---|---|---|
| `bun install --frozen-lockfile` | PASS | Bun installed the locked dependency graph without changing the lockfile. |
| `bun test src/loader.test.ts` from `packages/workflows/` | PASS | 346 passed, 0 failed, 933 assertions; all five scalar-shared regressions passed. |
| `bun run validate` | PASS | Generated-artifact checks, type-check, zero-warning lint, formatting, installer tests, all isolated package suites, and root script tests exited 0. |
| `git diff --check 74f7f5c019716044540b0082d9314a42447aa16f...5e58b81c30dfba740e5f87f0fcfaff1264f82242` | PASS | No whitespace errors. |

</details>

PR #2788, opened 2026-08-24
Validate output_format declared on workflow: nodes

### LINKED ISSUE #2774: fix(workflows): output_format on a workflow: node asserts a child contract nothing validates
## Problem

A `workflow:` node may declare `output_format`, but the schema is only used to populate the caller's accessible field names (`declaredFields`) — it is never validated against what the child workflow actually returns. The caller asserts into the dark: a schema-mismatched child passes silently, and the caller's downstream `.field` reads fail late (or resolve a declared-but-absent field to `''`) instead of the mismatch failing loudly at the boundary where it happened.

The gap is documented in as many words in `packages/docs-web/src/content/docs/reference/variables.md` ("`output_format` populates the accessible field names but is **not** validated against what the child actually returns"). #2460 fixed only the inversion half.

## Status against current dev (verified 24 Aug)

This is the last survivor of #2453's three concrete defects: the `when: $INPUTS` asymmetry is fixed (condition-evaluator handles named inputs with strict missing-input errors), and fan-out aggregate double-encoding is fixed (#2637 — logical array, serialized once). This one remains exactly as described.

## Desired outcome

A declared contract on a `workflow:` node boundary is either validated or rejected — never a silent assertion. The invariant: a schema-mismatched child result must fail the calling node with an error naming the caller, the child workflow, the failing path, the expected shape, and the received value type. Whether the contract belongs to the caller (validate the existing field against the child's return) or moves to the callee (a declared result contract, #2453's design direction, aligned with `direction.md §governed-launch-ownership` — the child owns its policy) is the designer's call; interim caller-side validation and the callee-owned end state are not mutually exclusive.

## Evidence

- `declaredFields`-only population in the executor; `WORKFLOW_NODE_IGNORED_FIELDS` deliberately excludes `output_format` so it stays meaningful.
- The documented gap at `reference/variables.md` (the `.field` access notes).
- Parent issue #2453 holds the full value-lifecycle trace and the constitution row that must be amended alongside a callee-owned design.

Wirasm:
## Problem and outcome

A `workflow:` node may declare `output_format`, but the schema was only used to populate `declaredFields` — it was never checked against what the child workflow actually returned, so a schema-mismatched child passed silently and downstream `.field` reads failed late (or resolved a declared-but-absent field to `''`). This is the last surviving concrete defect from #2453's trace (#2460 fixed only the inversion half).

- **Outcome:** A declared `output_format` on a `workflow:` node is now validated against the child's terminal output before any `node_completed` row is written; a mismatch fails the calling node loudly with an error naming the node, the child workflow, the failing ajv path(s), the expected schema, and the received type.
- **Invariant:** A schema-mismatched child result must fail the calling node — never silent. No `output_format` declared → behavior unchanged.
- **Scope boundary:** The callee-owned declared result contract (#2453's end state) remains out of scope; this is interim caller-side validation, which the issue explicitly allows.
- **Root cause:** The executor treated `output_format` as metadata only (`WORKFLOW_NODE_IGNORED_FIELDS` deliberately excludes it), never running it through the existing `validateStructuredOutput` gate that AI nodes already use.

## Review guidance

- **Feedback requested:** Correctness of the logical-value extraction (structuredOutput vs. text-summary fallback) and the fail-safe treatment of uncompilable schemas.
- **Start here:** `packages/workflows/src/dag-executor.ts` — `subrunLogicalValue()` plus the validation gate in `executeWorkflowNode`'s `asCompleted` and its fan-out parity gate in `executeFanOutWorkflowNode`; this is where the mismatch becomes a `node_failed`.
- **Review order:** executor gates (1:1 and fan-out parity) → the five new tests in `dag-executor.test.ts`, the new fan-out/certified-persistence cases in `subrun.test.ts` → the dry-run stub widening in `dry-run.ts` → the one-line docs change in `reference/variables.md` that previously documented the gap.
- **Lower-attention areas:** Docs change is mechanical; tests cover each branch.
- **Known risk or uncertainty:** Behavioral change is intentional — previously-silent mismatches now fail loudly, which may surface latent schema mismatches in existing workflows.

## Solution

The child run's terminal outcome is reduced to its logical value via a new `subrunLogicalValue()` helper: it prefers the typed `structuredOutput` introduced by #2637, falls back to `JSON.parse` of the text summary, and leaves non-JSON text as a raw string (which correctly fails an object-typed schema). When `node.output_format` is declared, this value goes through the existing `validateStructuredOutput` primitive before persistence. Invalid results become `failResult(...)` carrying the child's cost/tokens, so accounting stays correct. An uncompilable schema logs `workflow.subrun_schema_uncompilable`, sends a user-facing warning, and lets the node complete — the same fail-safe contract as the AI-node gate. The fan-out path gets the same treatment: every completed child is validated against the declared schema before any `node_completed` row is written, failing with a per-child named error. When a text-only child had no typed value but the gate certified a parsed one, that certified logical value is persisted as `structured_output` (1:1 event/result payload and the fan-out join element), so downstream typed access sees exactly what was validated. Relatedly, the dry-run stub generator now accepts array, number, boolean, integer, null, and const-typed `output_format` placeholders (`dryRunStubValueSchema` widened); only loop nodes with `until_field` still require an object-typed schema.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Schema-mismatched child passes silently; declared-but-absent fields read as `''`. | Calling node fails with named node id, child workflow, failing path(s), expected schema, received type. |
| **Failure behavior** | Late, indirect failure at downstream `.field` reads. | Loud failure at the boundary where the contract broke, persisted as `node_failed`. |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[Child returns wrong shape] --> B2[Parent completes silently] --> B3[Late .field failure]
  end
  subgraph After
    A1[Child returns wrong shape] --> A2[Boundary validation] --> A3[node_failed with full context]
  end
```

### Changed seams

|---|---|---|
| `workflow:` node → child workflow result | Declared `output_format` now enforced against child terminal output, 1:1 and fan-out; certified parsed value persisted as `structured_output` | `packages/workflows/src/dag-executor.ts`, five new tests in `dag-executor.test.ts`, new fan-out/persistence cases in `subrun.test.ts` |
| Dry-run stub scaffolding | Stub values may now be arrays/scalars matching non-object `output_format` schemas | `packages/workflows/src/dry-run.ts`, `dry-run.test.ts` |
| Documentation | "not validated" clause replaced with enforced behavior | `packages/docs-web/src/content/docs/reference/variables.md` |

## Validation

- `bun run test` in `packages/workflows` — all suites pass, 633 tests in dag-executor.test.ts including the five new ones.
- `bun x tsc --noEmit -p packages/workflows` — clean.
- `bun x eslint` on touched ts files — clean.
- New tests cover: structuredOutput violation, valid parsed text output, non-JSON text vs object schema, no-schema regression guard, uncompilable-schema warn path.
- **Not verified:** Nothing material — every new branch has a test; live multi-workflow runs were not exercised but are covered by the same executor path.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Workflows whose children silently violated their declared schema will now fail at the parent node — intentional per issue. | New test suite; issue #2774 desired outcome |
| Observability | Failures persist `node_failed` with full context; uncompilable schemas log `workflow.subrun_schema_uncompilable` (and `workflow.subrun_schema_warn_send_failed` if warning delivery fails). | Executor implementation |
| Rollout / rollback | Single commit revert restores prior silent behavior. | `3cc4e7bc` |
| Documentation / communication | `reference/variables.md` updated to describe enforced validation. | Diff |

## Links

- Closes #2774
- Relates to #2453
- Relates to #2637

## Summary by CodeRabbit

- **New Features**
  - Added validation for workflow sub-run outputs against their declared `output_format`.
  - Preserved validated structured output for downstream use and resumed runs.
  - Added support for scalar, array, boolean, numeric, null, and constant dry-run stub values.

- **Bug Fixes**
  - Invalid outputs no longer incorrectly complete workflow nodes.
  - Improved fan-out sub-run and text-based JSON handling.
  - Nodes declaring no checkout mutations now fail when they modify tracked files.

- **Documentation**
  - Documented sub-run output validation behavior.


Wirasm:
# Review Report — PR #2788 "Validate output_format declared on workflow: nodes" (round 6, light, delta `a1a0ebf4..4f40c865`)

## Verdict

**Ready.** The round-5 Important finding R4 — the last open blocking finding — is fixed by commit `4f40c865` and verified end-to-end: the stub-value union now admits every placeholder `schemaPlaceholder` produces, all prior findings remain closed, and both enabled lenses delivered their reports with no new findings.

## Reviewed head SHA

`4f40c8651e9a5a0dc00d0fb4b71aef9eb3e6b5ed`

## Findings

### Critical

None open.

### Important

None open.

### Rejected findings

None this round — no new Critical or Important finding was raised; every prior-finding evidence citation checked out under adversarial verification.

### Suggestions

None raised this round. (All optional lenses were disabled.)

| ID | Severity | Claim | Verdict at head `4f40c865` |
|---|---|---|---|
| R4 | Important | `DryRunStubValue` union admitted arrays but not scalar placeholders (`false`/`0`/`null`) from `schemaPlaceholder`, throwing on supported schemas with an inaccurate "non-JSON" message | **Fixed at `4f40c865`** — verified end-to-end. Union widened with `z.number()`, `z.boolean()`, `z.null()` (`dry-run.ts:61-66`); message reworded to "placeholder of an unsupported type" (`dry-run.ts:186-188`). New parametrized test covers boolean/number/integer-minimum/null/scalar-const scaffolds (`dry-run.test.ts:457-471`); stale "rejects numeric stub value" assertion removed. Full dry-run suite re-run by synthesis at head: **64 pass / 0 fail**. |
| R3 (alias S1) | Important | Array-typed `output_format` rejected by dry-run stub generator | **Still fixed** (landed `a1a0ebf4`) — `z.array(z.unknown())` remains in the union (`dry-run.ts:60`) and `schemaPlaceholder`'s `'array'` branch unchanged; array scaffolding covered by existing passing tests. |
| R1 | Critical | `output_format` enforcement skipped `fan_out` nodes | **Still closed, untouched** — `dag-executor.ts` has zero changes in `a1a0ebf4..HEAD` (verified via diff); fan-out gate intact per seams lens (`dag-executor.ts:~8645-8657`). |
| R2 | Important | Fan-out join persisted unparsed text while gate validated parsed value | **Still closed, untouched** — same gate fails before `writeCompleted`; delta does not touch dag-executor.ts. |
| N1 | Important | 1:1 `asCompleted` didn't persist parsed logical value | **Still closed, untouched** — certified logical-value persistence at `dag-executor.ts:~7662-7668`; delta does not touch dag-executor.ts. |

## Lens roster

| Lens | Status |
|---|---|
| code | Ran — no new findings; verified R4 fix and R3/R1/R2/N1 verdicts with executed probes and full suite run |
| seams | Ran — no new findings; all consumers of the widened type verified consistent (generator/loader/hydration/validator/persistence round trip) |

No enabled lens failed to report. Note for any next round: only `code` and `seams` ran; a fuller clearance has not been asserted beyond those two lenses' scope.

## Canonical PR comment

Published as one canonical comment carrying `<!-- archon-review-report -->`: https://github.com/coleam00/Archon/pull/2788#issuecomment-5401378788

PR #2789, opened 2026-08-24
Enforce per-node mutates_checkout: false in the DAG executor

### LINKED ISSUE #2771: feat(workflows): an advisory node can declare it leaves the checkout untouched — and the engine enforces it
## Problem

Three bundled packages (triage, upkeep, stabilize on the #2627 branch) each carry the same two hand-built guard nodes around their advisory AI step: a `record-state` bash node snapshotting `git status --porcelain` before it, and an `assert-intact` bash node diffing the snapshot after it, failing the run if the advisory step mutated the checkout. Six nodes and two sidecar files per run exist only to prove a single invariant the author wanted to declare: *this node is read-only*.

The pattern hit the rule of three, and extraction into a shared `include:` block is not expressible (an include cannot parameterize its `command:` ref) — but per the direction doc's own §workaround-triage this is bucket one anyway: a missing **coordination** primitive, not a missing pattern. The engine is the right owner of this guarantee, the way #2710 made failed-producer→failed-binding an engine guarantee instead of pack defensiveness.

## Desired outcome

A node can declare that it must not mutate the working tree, and the engine enforces it: the node fails — attributably, as its own failure — if the tree changed while it ran. The three packages then delete all six guard nodes.

Relation to existing surface, for the designer: workflow-level `mutates_checkout` already exists as *scheduling metadata* (fan-out preflight) and is unenforced; whether this becomes a node-level field, an enforcement of the existing declaration, or something else is open. Admissibility per §workflow-language: the engine needs to see it to govern (it is a governance guarantee), it is declarative data, and the script-node expression of it is exactly the duplication this issue retires.

## Evidence

- The triad in situ: `.archon/workflows/sdlc/{triage,upkeep,stabilize}` on `feat/2123-sdlc-pack` (#2627).
- Live catches: the guard fired truthfully during the 20 Aug dogfoods (advisory verdicts without their report, tree mutations during assessment) — the invariant earns its cost; only its implementation is duplicated pack machinery.

Wirasm:
## Problem and outcome

Workflows can declare that a node is read-only (`mutates_checkout: false`), but nothing enforced it — a bash, script, or agent node could freely dirty the working tree while claiming not to, silently corrupting later nodes' assumptions.

- **Outcome:** A node that declares `mutates_checkout: false` and still modifies the working tree is rewritten to a `failed` result naming the node and the paths it touched, surfaced through the standard `node_failed` event. Nodes without the declaration behave bit-identically to today.
- **Invariant:** Enforcement never changes the outcome of a node that did not violate the declaration; engine-owned write roots (artifacts, state, logs) are always exempt; an unrunnable check (outside a git repo, git failure) degrades to no check rather than failing the run.
- **Scope boundary:** Only exec (bash/script) and agent node kinds are enforced in v1; wait/approval/gate/halt/loop/workflow kinds carry the field via the base spread but are not checked. Workflow-level `mutates_checkout` semantics (scheduling/path-lock) are unchanged.

## Review guidance

- **Feedback requested:** Correctness of the snapshot/diff logic and the placement outside the retry loops.
- **Start here:** `packages/workflows/src/dag-executor.ts:1218` — `snapshotCheckout` + `assertCheckoutUntouched`; this pair is the entire enforcement mechanism.
- **Review order:** schema field on `dagNodeBaseSchema` (`schemas/dag-node.ts`) → `assertCheckoutUntouched` in `dag-executor.ts` → its three wiring sites in `runLayers`.
- **Lower-attention areas:** test files — they mechanically cover the behaviors above; the loader changes are just re-export/threading.
- **Known risk or uncertainty:** The snapshot is `git status --porcelain`, so a violation is detected only if the tree differs from pre-run state; a node that mutates and restores exactly cannot be caught by any status-based check.

## Solution

After a *successful* run of a node declaring `mutates_checkout: false`, the executor re-snapshots the working tree (`git status --porcelain`) and compares against the pre-run snapshot, dropping entries under the engine's artifacts/state/log dirs. On difference the result is rewritten to failed with an error naming the node and up to ten changed paths, persisted as a `node_failed` event and emitted so the violation is visible like any other failure.

The assertion deliberately sits **outside** the retry loops: retrying a node that provably mutates would only multiply damage, so the failure is non-retryable by construction. A node that already failed keeps its own attributable outcome. When the check can't run (no repo, git error), the snapshot returns `undefined` and enforcement skips — fail-open, matching the principle that a broken assertion shouldn't break unrelated runs.

A layer containing any node that declares `mutates_checkout: false` runs strictly sequentially (`settleSequentially`) instead of concurrently, so a sibling's write landing inside the guarded node's snapshot window cannot be falsely attributed to it; layers with no guarded node keep the concurrent path.

The field is optional boolean on `dagNodeBaseSchema`, threaded via the existing shared base spread; absent means no enforcement and `true` remains declaration-of-intent only, mirroring workflow-level usage. Include nodes keep their existing hard rejection at `loader.ts:633`.

Two deliberate deviations from the original plan: non-boolean values are a hard Zod load error (matching `always_run`/`persist_session` and the flat-schema convention), and the rewritten failure emits its own `node_failed` event because post-run rewrites get none from the dispatch layer.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `mutates_checkout: false` is decorative; a read-only-declared node can dirty the tree silently | Violating node fails with `Node \`<id>\` declared \`mutates_checkout: false\` but modified the working tree: <paths>` |
| **Failure behavior** | Nothing recorded; downstream nodes inherit an unexpectedly dirty tree | Standard `node_failed` event persisted and emitted; violation is non-retryable |

## Validation

- `cd packages/workflows && bun run type-check` — clean; root `bun run type-check` — clean.
- `cd packages/workflows && bun run test` — exit 0, 0 failures. (Running bare `bun test src/` over all 40 files in one process fails ~930 tests at HEAD too — pre-existing cross-suite interference, unrelated to this change.)
- New tests: 7 executor tests (clean pass; violation names node+paths; undeclared nodes not checked; artifacts-dir writes exempt; fail-open outside git; concurrent sibling writes not attributed to a guarded node; non-ASCII paths under excluded dirs), 4 schema tests, 4 loader tests.
- eslint via lint-staged on commit clean; `check:bundled-schema` / `check:bundled` report bundles up to date.
- **Not verified:** end-to-end run of a real workflow with a violating node in a live engine session — covered instead by the executor-level integration tests, which exercise the same dispatch path.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Fully opt-in: workflows without per-node `mutates_checkout: false` see identical behavior | absent-field early return in `assertCheckoutUntouched` |
| Security / permissions / data | Strengthens integrity guarantees for read-only declarations; no new exposure | snapshot diff restricted to non-engine-owned paths |
| Rollout / rollback | Single commit; revert removes all enforcement | 3f1d1633 |
| Observability | Violations logged as `dag_mutates_checkout_violation` and emitted/persisted as `node_failed` | `dag-executor.ts` assert helper |

## Links

- Closes #2771

## Summary by CodeRabbit

* **New Features**
  * Added optional checkout mutation declarations for executable workflow nodes.
  * Nodes marked as non-mutating now detect unauthorized working-tree changes and report the affected paths.
  * Engine-managed files and directories are excluded from mutation checks.

* **Bug Fixes**
  * Improved validation and warnings for unsupported checkout mutation declarations across node types.
  * Preserved safe execution behavior outside Git repositories and between parallel workflow nodes.


Wirasm:
# Review Report — PR #2789, Round 2 (light) — per-node `mutates_checkout: false` enforcement

## Verdict

**Ready.** Both prior Important findings are fixed at this head with regression tests verified passing, and both enabled lenses delivered reports with no new Critical or Important findings.

## Reviewed head SHA

`fdcb7bfeb051d3a207da3d3a6f5f811e39c81dd3`

(Delta reviewed: `git diff -w 3f1d16335fba3ce87c2027241c350e3cfca671bb..fdcb7bfeb051d3a207da3d3a6f5f811e39c81dd3`. The ~1800-line churn in `dag-executor.ts` is almost entirely whitespace/re-indentation of the thunk body; the substantive delta is ~62 lines.)

## Findings — Round 2 delta

### Critical
None.

### Important
None.

### Rejected
None — no finding from this round was disproved.

### Suggestions
- **S1 (non-blocking)** — Leftover debug line: `packages/workflows/src/dag-executor.test.ts:30299` contains `console.log('EVENTS', JSON.stringify((deps.store.createWorkflowEvent as any).mock.calls));` inside an otherwise assertion-only test. Test-only noise; flagged by both the code and seams lenses for cleanup.
- **S2 (non-blocking)** — Cross-file test interference observed during synthesis verification: running `bun test -t "mutates_checkout"` across all files at once produces failures in `subrun.test.ts` and `loader.test.ts` (fixed temp-dir collisions between concurrently running test files), while each file passes cleanly in isolation — including `subrun.test.ts`, which this delta does not touch. Pre-existing infrastructure issue, not introduced at this head; worth a follow-up to give per-file unique temp dirs.

## Prior findings (light mode carry-forward)

| ID | Severity | Finding | Verdict |
|---|---|---|---|
| R1 | Important | Parallel-layer concurrency: an undeclared sibling's write landing inside a guarded node's snapshot→assert window falsely fails the innocent node (`assertCheckoutUntouched` rewrites success to `failed`). | **Fixed at `fdcb7bfe`.** Verified directly: `dag-executor.ts:9032-9045` (`settleSequentially`, exact `PromiseSettledResult<LayerNodeResult>` shape so caller result handling is unchanged) and `dag-executor.ts:10142-10144` — a layer where any node declares `mutates_checkout === false` runs strictly sequentially, making the guarded node's window exclusive. Regression test "a mutating sibling in the same layer is not attributed to a guarded node" (`dag-executor.test.ts`) passes under `bun test src/dag-executor.test.ts`. Loop_group bodies re-enter the same `runLayers`, so they inherit the rule; cross-layer overlap was never possible. |
| R2 | Important | `mutates_checkout` parsed via shared base spread but neither enforced nor loader-warned on loop/loop_group/gate/halt nodes. | **Fixed at `fdcb7bfe`.** Verified directly: `'mutates_checkout'` appended to `LOOP_NODE_AI_FIELDS` (`schemas/dag-node.ts:929`), `LOOP_GROUP_NODE_AI_FIELDS` (:942), and new `GATE_AND_HALT_IGNORED_FIELDS` (:952); gate/halt loader branches switched onto it (`loader.ts:664-669`); exported via `schemas/index.ts`. Coverage across all nine node kinds is complete: agent/exec enforce, loop/loop_group/gate/halt/wait/workflow warn as inert, include hard-errors. Loader tests for loop/approval/cancel/loop_group warnings pass under `bun test src/loader.test.ts`. |

## Lens roster

| Lens | Status |
|---|---|
| code | Ran — no new findings; R1/R2 verified fixed |
| seams | Ran — no new findings; 7 seams examined clean, 2 carve-outs recorded |

Both enabled lenses delivered full reports; no lens failed to report. This round ran only 2 of 8 lenses (light mode); readiness is certified within that coverage only.

PR comment: https://github.com/coleam00/Archon/pull/2789#issuecomment-5401491527 (canonical comment updated in place, read-back verified)

PR #2798, opened 2026-08-25
Patient rate-limit backoff and transient retries for silent stream deaths (#2706)

### LINKED ISSUE #2706: feat(providers): classify provider rate/concurrency errors — backoff, wait-for-slot, uniform failure semantics
The 2026-08-22 MiniMax dogfood series surfaced three related gaps in how provider-level errors travel through workflow execution. Logging for pickup — not urgent.

## Evidence

Fifteen terminal 429s (`rate_limit_error`, MiniMax code 2062) between 23:14 and 04:31 UTC across runs `a67a9f8f`, `94d77c13`, `ae71ab1e`, `6607bf20`, `6e6c9f9e`. Pattern: failures clustered exactly when the review's 8-lens fan-out burst concurrent calls (and when two runs overlapped); a SINGLE probe request returned 200 twenty minutes after the last failure. That fits a **provider-side concurrency cap under load** better than a token-volume window — MiniMax appears to throttle parallel connections, and the error body carries no retry-after or reset information to read.

## The three gaps

1. **Rate/concurrency 429s get short flat retries.** Node `retry:` (doubling from `delay_ms` ≤ 60s) exhausts within a minute against a load-shedding window that needs minutes-scale patience. The wanted shape is roughly *wait for a slot, then continue*: classify the 429 distinctly, back off longer (flat + jitter, since there is no retry-after to respect), and ideally hold the node rather than burn attempts.
2. **Silent stream death is not retried at all.** Run `ae71ab1e`'s code lens failed once with "provider stream closed without yielding content — likely a silent provider rejection" and was never reattempted. An empty stream should classify transient.
3. **Failure semantics are not uniform across node kinds.** A plain node retries; a `loop:` iteration died on its first 429 (run `6e6c9f9e`, iteration 1 — node retry does not visibly apply inside loop iterations); a `loop_group` body failure leaks into #2696 case 3. One provider-error taxonomy, applied identically to plain nodes, loop iterations, and group bodies.

## Related

- A per-provider concurrency limiter is the complementary demand-side fix — #1961's scope. The 8-lens fan-out is the canonical burst producer (the sdlc pack is reducing its default width independently, but any fan-out reproduces this).
- #2696 case 3 (failed group exposes a bindable output) is what turned one of these provider failures into a false public success — separate bug, fix first.
- Classifying the provider's own error strings is legitimate under the wire-format rule's 1g carve-out.

Wirasm:
## Problem and outcome

Workflow nodes dying on provider rate limits or silent empty streams fail runs far too eagerly: the node's own short `maxRetries` exhausts within seconds while the provider's load-shedding window is minutes-scale, and loop iterations had **no** transient retry at all — one 429 killed the whole loop node.

- **Outcome:** Rate-limited failures now retry up to 5 times with a patient flat ~45s ±50% jitter backoff; loop-node AI iterations survive transient errors (including silent stream death) with the same retry policy as plain AI nodes.
- **Invariant:** FATAL failures are never retried; quota/session exhaustion stays fatal; usage/cost accumulation still counts every paid attempt; cancellation and structured-output reask behavior are unchanged.
- **Scope boundary:** No per-provider backoff tables or retry-after honoring; no demand-side concurrency limiting (#1961); #2696 case-3 false-success leak untouched.
- **Root cause:** `classifyError()` treated "stream closed without yielding content" as UNKNOWN (so non-retryable in effect), and `executeLoopNode` failed the iteration on first TRANSIENT error instead of retrying.

## Review guidance

- **Feedback requested:** Correctness of the retry-budget widening (`sawRateLimit` latches for the rest of the loop by design) and the labeled-attempt-loop refactor in `executeLoopNode`.
- **Start here:** `packages/workflows/src/dag-executor.ts:1116` — `runNodeRetryLoop` now widens its budget to `max(node maxRetries, RATE_LIMIT_MAX_RETRIES)` once a rate-limit failure is seen, and routes all delays through `getRetryDelayMs`.
- **Review order:** `executor-shared.ts` (patterns + delay policy) → `runNodeRetryLoop` → the `iterationAttempt:` loop in `executeLoopNode` → tests.
- **Lower-attention areas:** The ~600 mechanical reindent lines from wrapping the iteration body in the attempt loop — review with `git diff -w`. Test updates that gave failing fake nodes explicit `retry: {max_attempts:1}` exist only so they keep testing failure semantics without the new intended retry waits; two telemetry assertions moved `errorClass: 'unknown' → 'transient'` per the new taxonomy, and one usage assertion doubled because both paid attempts now count.
- **Known risk or uncertainty:** The rate-limit budget latch is per-node-invocation, not reset on recovery — a node flapping between success and 429 keeps the widened budget for that invocation, which is intentional but worth a second opinion.

## Solution

`executor-shared.ts` adds `RATE_LIMIT_PATTERNS`, `isRateLimitError()`, and `getRetryDelayMs()`: rate-limit messages back off flat at `45s × (0.5 + random())` — providers recover on a minutes-scale window with no retry-after signal, so exponential-from-3s either exhausts before the window opens or over-waits after it does, and jitter spreads concurrent nodes apart without thundering-herd resynchronization. Everything else keeps base × 2^attempt. Silent empty streams ("stream closed without yielding content", single-shot and loop wording) join `TRANSIENT_PATTERNS`.

`dag-executor.ts` applies this in two places: the shared `runNodeRetryLoop` (which loop-group bodies already use via `runLayers`), and a new per-iteration transient retry inside `executeLoopNode` — post-attempt state (`fullOutput`, `cleanOutput`, `iterationIdleTimedOut`, `iterationPayload`) hoisted above a labeled attempt loop whose catch and empty-output branches call `tryIterationTransientRetry()` before falling through to `failLoopNode`. Each retry logs `loop_node.iteration_transient_retry` and notifies the conversation; failed attempts keep their `loop_iteration_failed` events.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A 429 exhausts the node's short retry budget in seconds; loop iterations die on first transient error | Up to 6 attempts (~45s ±50% apart) once rate-limited; loop iterations retry transients like plain AI nodes |
| **Failure behavior** | Run fails; user retries manually much later | Node weathers the load-shedding window and recovers in-run |

## Validation

- `packages/workflows`: `bun run type-check` ✓, `bun run test` ✓ (0 fail) — including 4 new executor tests (rate-limit budget exhaustion = 6 provider calls; mid-budget recovery; empty-stream retry; loop-iteration 429 recovery) and 4 new `executor-shared` tests (rate-limit patterns staying inside `TRANSIENT_PATTERNS`, empty-stream classification, `isRateLimitError` boundaries, `getRetryDelayMs` ranges).
- Repo root: `bun run type-check` ✓, `bun run lint` ✓, `prettier --check .` ✓, `bun run test` exit 0 ✓.
- **Not verified:** Against live providers under real 429 load — timing policy is covered by unit ranges, not production traffic.

## Links

- Closes #2706

## Summary by CodeRabbit

* **New Features**
  * Improved retry handling for rate-limit failures, including a longer jittered delay and expanded retry allowance.
  * Added automatic retries when AI responses end without producing content.
  * Improved structured-output recovery for missing or invalid responses.
  * Preserved tool events, usage tracking, cancellation, and background tasks across retries.

* **Bug Fixes**
  * Improved error classification for transient and rate-limit failures.
  * Prevented duplicate usage accounting across streamed results.

* **Documentation**
  * Updated retry behavior documentation, including rate-limit-specific rules.

### DIFF
--- a/.claude/skills/archon/references/dag-advanced.md
+++ b/.claude/skills/archon/references/dag-advanced.md
@@ -315,11 +315,11 @@ For deterministic bash/script failures (a script that exits 1 reproducibly), ret
 
 | Category | Examples | Retried? |
 |----------|----------|----------|
-| **FATAL** | `unauthorized`, `forbidden`, `permission denied`, `invalid token`, `authentication failed`, `auth error`, `401`, `403`, `session limit`, `usage limit reached`, `credit exhaustion`, `credit balance` | Never |
+| **FATAL** | `unauthorized`, `forbidden`, `permission denied`, `invalid token`, `authentication failed`, `401`, `403`, `session limit`, `usage limit reached`, `credit exhaustion`, `credit balance` | Never |
 | **TRANSIENT** | `timeout`, `etimedout`, `rate limit`, `too many requests`, `429`, `502`, `503`, `529`, `overloaded`, `at capacity`, `stream closed without yielding content`, `econnrefused`, `econnreset`, `network error`, `socket hang up`, `exited with code`, `claude code crash` | By default (rate-limit matches get the ~45s±50% backoff and widened budget above) |
 | **UNKNOWN** | Everything else | Only with `on_error: all` |
 
-FATAL patterns take priority over TRANSIENT patterns in the same error message.
+FATAL patterns take priority over TRANSIENT patterns in the same error message. `auth error` alone is fatal, but yields to any concrete transient pattern in the same message.
 
 ### Two-Layer Retry Stack
 
--- a/.claude/skills/archon/references/dag-advanced.md
+++ b/.claude/skills/archon/references/dag-advanced.md
@@ -315,7 +315,7 @@ For deterministic bash/script failures (a script that exits 1 reproducibly), ret
 
 | Category | Examples | Retried? |
 |----------|----------|----------|
-| **FATAL** | `unauthorized`, `forbidden`, `permission denied`, `invalid token`, `authentication failed`, `auth error`, `401`, `403`, `credit balance` | Never |
+| **FATAL** | `unauthorized`, `forbidden`, `permission denied`, `invalid token`, `authentication failed`, `auth error`, `401`, `403`, `session limit`, `usage limit reached`, `credit exhaustion`, `credit balance` | Never |
 | **TRANSIENT** | `timeout`, `etimedout`, `rate limit`, `too many requests`, `429`, `502`, `503`, `529`, `overloaded`, `at capacity`, `stream closed without yielding content`, `econnrefused`, `econnreset`, `network error`, `socket hang up`, `exited with code`, `claude code crash` | By default (rate-limit matches get the ~45s±50% backoff and widened budget above) |
 | **UNKNOWN** | Everything else | Only with `on_error: all` |
 
--- a/.claude/skills/archon/references/dag-advanced.md
+++ b/.claude/skills/archon/references/dag-advanced.md
@@ -309,12 +309,14 @@ Available on command and prompt nodes (default-on for transient errors), and on 
 
 For deterministic bash/script failures (a script that exits 1 reproducibly), retrying is pointless — `retry:` there is for flaky externals (network fetches, rate-limited APIs). Classification detail: a bash/script failure message is formatted `<node> failed [exit N]: <stderr>`, so the classifier runs on your script's **stderr text** — the `exited with code` TRANSIENT pattern in the table below targets AI-CLI crash messages and does NOT match a bash/script non-zero exit. A subprocess `timeout` DOES classify TRANSIENT. Practical rule: rely on the default `on_error: transient` when failures surface as timeouts/rate-limit text on stderr; use `on_error: all` when the flaky failure mode produces generic stderr.
 
+**Rate-limit override (#2706):** once any attempt fails with a rate-limit pattern (`429`, `rate limit`, `too many requests`, `overloaded`, `at capacity`), the delay policy changes: instead of your `delay_ms` doubling, retries back off flat at ~45s ±50% jitter (22–67s) and the attempt budget widens to at least 5, regardless of `max_attempts`. Your `delay_ms`/`max_attempts` ceilings are ignored for these failures — providers shedding load recover on a minutes-scale window that short exponential backoff misses.
+
 ### Error Classification
 
 | Category | Examples | Retried? |
 |----------|----------|----------|
 | **FATAL** | `unauthorized`, `forbidden`, `permission denied`, `invalid token`, `authentication failed`, `auth error`, `401`, `403`, `credit balance` | Never |
-| **TRANSIENT** | `timeout`, `etimedout`, `rate limit`, `too many requests`, `429`, `502`, `503`, `econnrefused`, `econnreset`, `network error`, `socket hang up`, `exited with code`, `claude code crash` | By default |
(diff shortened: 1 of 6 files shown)


Wirasm:
# Review Report — PR #2798 (round 4, light)

## Verdict

**ready: true** — the sole open Important finding (N1) is verified fixed at this head, all prior findings remain fixed with no regression, and both enabled lenses delivered their reports.

## Reviewed head SHA

`260e627e0dfdf938f2f2c175decda64d71076678`

## Findings

### Critical

None.

### Important

None — N1 (the only open Important finding from round 3) is **fixed at `260e627e`**; see Prior findings.

### Rejected findings

None — every claim surfaced by a lens held under adversarial verification this round.

- Carried forward from round 2: cancellation during the 22–67s rate-limit retry sleep is honored only once the next attempt's mid-stream check fires (~up to ~68s + stream start). Pre-existing posture; could add an abort-aware sleep later. Unaffected by this doc-only delta.

## Prior findings

|---|---|---|---|
| N1 | Important | `auth error` documented as unconditional FATAL ("Never" retried) while code classifies it as fallback-fatal yielding to transient patterns | **Fixed at `260e627e`** — verified by synthesist against the diff and source: FATAL row no longer contains `auth error` (`dag-advanced.md:318`); added sentence "`auth error` alone is fatal, but yields to any concrete transient pattern in the same message" (`dag-advanced.md:322`) matches `classifyError`'s order (`packages/workflows/src/executor-shared.ts:103-112`) where `FALLBACK_FATAL_PATTERNS = ['auth error']` (`executor-shared.ts:53`) is checked only after TRANSIENT. Tests cover both branches (`executor-shared.test.ts:841`, `:887`). |
| S1 | Important | FATAL row omitted three of four quota-exhaustion patterns (fixed round 3) | **Still fixed** — non-regression confirmed: row at `dag-advanced.md:318` retains all four `QUOTA_EXHAUSTION_PATTERNS`; the delta only removed `auth error`. |
| R1 | Important | `RATE_LIMIT_PATTERNS` spread + containment unenforced (fixed round 2) | **Still fixed** — non-regression confirmed; delta touches docs only. TRANSIENT row still lists every member of `RATE_LIMIT_PATTERNS`. |
| R2 | Important | Published skill docs omitted new rate-limit backoff behavior (fixed round 2) | **Still fixed** — rate-limit override paragraph and backoff note (~45s±50%) unchanged and consistent with `RATE_LIMIT_RETRY_DELAY_MS = 45_000` (`executor-shared.ts:120`). |

## Lens roster

| Lens | Status |
|---|---|
| code | Ran — no new findings; verified N1 fixed, S1/R1/R2 non-regressed |
| seams | Ran — no new findings; doc ↔ classifier seam diffed item-by-item, exact set match on both sides |

Both enabled lenses delivered their reports; no lens failed to report. Only 2 of 8 lenses ran this round (code + seams) — this is a delta-scoped light round, not a full clearance.

## Adversarial verification log

- Read the actual delta `git diff c231da95..260e627e`: exactly two changed lines in `.claude/skills/archon/references/dag-advanced.md` — removal of `` `auth error` `` from the FATAL row and addition of the yield sentence after the priority line. Matches scope.md.
- N1 resolution: read current file lines ~315–322 — FATAL row has 11 items matching `FATAL_PATTERNS` (`executor-shared.ts:44-54`), yield sentence consistent with tier ordering at `executor-shared.ts:103-112`.
- S1/R1/R2 non-regression: grep confirmed `QUOTA_EXHAUSTION_PATTERNS` (`:33`), `FALLBACK_FATAL_PATTERNS = ['auth error']` (`:53`), `RATE_LIMIT_PATTERNS` (`:62`), `RATE_LIMIT_RETRY_DELAY_MS = 45_000` (`:120`) all intact.
- Cross-lens agreement: code and seams independently reached identical verdicts on all four prior findings; no disagreement to reconcile.

## PR publication

Canonical comment updated in place at https://github.com/coleam00/Archon/pull/2798#issuecomment-5405901153 (marker `<!-- archon-review-report -->` on first line); read-back verified against this report.

PR #2799, opened 2026-08-25
Pin #2794: wait-resume until_bash recheck with a pre-wait body node

### LINKED ISSUE #2794: bug(workflows): a wait-terminated loop_group completes after resume without evaluating until_bash
## Problem

A `loop_group` whose body ends in a terminal `wait:` node completes after the wait resumes even when its `until_bash` condition is false, so the loop runs exactly one iteration regardless of what the body observed. Seen live on the first real run of a wait-terminated loop (run `1d7b08f3-2e83-4bd2-a8f5-d1dc240befc9`, 2026-08-24, the CI-gate loop shipped in the sdlc pack on PR #2627).

The group:

```yaml
- id: await-checks
  loop_group:
    max_iterations: 13
    until_bash: test "$ci-probe.output.state" != "pending"
    nodes:
      - id: ci-probe        # script; prints {"state": "pending"|"concluded", ...}
      - id: ci-pause        # wait: {duration_ms: 300000}, when: pending, terminal sink
```

## Observed sequence (workflow_events for that run)

1. `loop_iteration_started` (iteration 1)
2. `ci-probe` completed with output `{"state": "pending", "detail": "1 check(s) running"}`
3. `wait_started` → run paused durably → server continuation scan resumed it
4. `wait_completed`, `ci-pause` completed `{"status":"satisfied","waited_ms":304402}`
5. **`deliver__await-checks` (the group) `node_completed` with output `{"status":"satisfied","waited_ms":304402}`** — no second iteration, no second probe

With the only probe of the iteration reporting `pending`, `until_bash` (`test "$ci-probe.output.state" != "pending"`) should have evaluated false and started iteration 2. Instead the group completed, its output became the wait's fixed output, and the downstream flip-ready node made PR #2792 public-ready while its windows CI leg was still running (it later concluded red on the PR's own test, and the PR had to be re-drafted by hand).

## What this looks like

The post-resume completion recheck for a terminal body wait (the "Archon then escalates the persisted cursor to the group and rechecks the group's completion condition after the wait completes" path in the durable-waits docs) does not evaluate `until_bash` against the same context a normal end-of-iteration check gets — either the body-node output substitution (`$ci-probe.output.state`) resolves differently on the rehydrated path, or the recheck concludes the group without consulting `until_bash` at all. A normal (non-wait) `loop_group` evaluates `until_bash` correctly on the same engine — deliver's corrections loop exits early on `$recheck.output.ready` — so the divergence is specific to the resumed-wait path.

Not prescribing the mechanism; the run's event log and artifacts are intact for whoever picks this up.

## Impact

Any wait-terminated polling loop silently degrades to "wait once, then proceed", which inverts the gate it implements. The sdlc pack's CI gate has been hardened to compute completion inside `until_bash` directly (external check, no body-output substitution) as a workaround, so the pack no longer exercises the broken path — this issue is the engine contract: `until_bash` must see the same evaluation context whether the iteration ended normally or through a resumed wait.

Wirasm:
## Problem and outcome

The live #2794 run completed a wait-terminated `loop_group` after one iteration even though its restored probe output was still `{"state":"pending"}`. The run started after the #2779 resume restoration had merged, so the original claim that the incident predated that engine change was wrong.

The exact incident shape shows that resume restoration works. The failure came from authoring the condition as:

```bash
test "$ci-probe.output.state" != "pending"
```

Node-output values in `bash:` and `until_bash` are already shell-quoted. For the string field `pending`, Archon injects `'pending'`; the surrounding double quotes preserve those single quotes as data, so the comparison becomes `'pending' != pending` and the group completes.

- **Outcome:** The regression now uses the include-expanded group, restored JSON field, inequality polarity, and terminal wait from the incident. The valid unquoted condition starts iteration 2; the otherwise identical double-quoted condition reproduces the premature completion.
- **Invariant:** A resumed wait evaluates `until_bash` against the same include-qualified body output as an ordinary iteration boundary.
- **Scope:** Test correction only. The executor's resume behavior is unchanged because it restores and resolves the incident-shaped output correctly.

## Review guidance

- **Start here:** `packages/workflows/src/dag-executor.test.ts`, in `includedProbeWaitTerminatedLoopGroupWorkflow` and the two `#2794` cases.
- **Primary check:** The valid fixture must keep `$ci-probe.output.state` unquoted, restore `deliver__await-checks.ci-probe = {"state":"pending"}`, and pause on `ci-pause` in iteration 2.
- **Discriminator:** Adding only double quotes around that same reference must complete the group without scheduling iteration 2, reproducing the reported event sequence.
- **Current-dev interaction:** The adjacent #2803 test still proves that a missing restored producer fails loud before `until_bash` executes.
- **Lower-attention areas:** No production executor, schema, or persistence code changed.

## Solution

The prior top-level scalar/equality fixture was replaced while reconciling the branch with current `dev`. The shared include-expanded fixture now emits JSON and can author either the valid unquoted field reference or the incident's double-quoted reference. Two focused tests run the same resumed wait state through both forms and assert the opposite outcomes.

This pins both sides of the diagnosis without changing the engine or preserving the earlier incorrect timeline/root-cause claim.

## Validation

- `bun test src/dag-executor.test.ts --test-name-pattern '#2794|#2803: an included wait-resume'` — 3 pass, 0 fail.
- `bun run type-check` (`packages/workflows`) — passed.
- `NODE_OPTIONS=--max-old-space-size=8192 bun run validate` — passed.
- Branch merged current `origin/dev` without dropping the newer #2803 strict-reference regression.

## Links

- Closes #2794
- Related: #2779, #2803

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed loop-group workflows so resumed waits correctly evaluate previously persisted JSON output.
  * Prevented incorrect completion or unnecessary pauses caused by double-quoting output references.
  * Ensured completion conditions are rechecked accurately after a workflow resumes.


Wirasm:
# Review Report — PR #2799 (head `e1a7e8183e792d285640b27248dfe99bc2bac6e4`)

## Verdict

**ready: true** — no open Critical or Important findings and both enabled lenses (`code`, `seams`) delivered full reports; the only finding is a non-blocking comment-accuracy suggestion.

## Reviewed head SHA

`e1a7e8183e792d285640b27248dfe99bc2bac6e4`

## Findings

### Critical
None.

### Important
None.

### Rejected findings
None — no Critical/Important finding from either lens failed verification.

- **R1** (Suggestion — from `seams`) — **The new test's failure-mode comment mischaracterizes the bash semantics of its own fixture.** The comment at `packages/workflows/src/dag-executor.test.ts:30061-30062` claims `$probe.output` substituting empty yields `[ '' = "satisfied" ]` which "exits 1"; but the fixture's `until_bash: '[ $probe.output = "satisfied" ]'` (`dag-executor.test.ts:29840`) substitutes the ref **unquoted**, so an empty output produces `[ = "satisfied" ]` — a 2-arg `test` that fails with *"unary operator expected"* and **exit 2**, not exit 1. Impact is nil on the pinned behavior (both exits are nonzero, until_bash false either way); only the test's documentation of *why* row loss is silent is wrong. Smallest fix: correct the comment to "exits nonzero", or quote the ref.

## Prior findings

First round — none to carry forward.

## Lens roster

| Lens | Status |
|---|---|
| code | Ran — found nothing at any severity |
| seams | Ran — one Minor/Suggestion (R1) |

No enabled lens failed to report. Next-round reviewers: only `code` and `seams` ran this round; if other lenses are enabled next round, they must run in full (this is a first round with no prior baseline beyond this report).

## Verification notes

- Diff re-derived via `git diff dev...HEAD` in checkout at head SHA (matches PR head per scope.md).
- R1's evidence checked directly: fixture line 29840 shows the unquoted ref; lines 30058–30063 show the comment.
- Code lens claims (test passes under `bun test -t "#2794"`, mutation check on `priorCompletedNodes`, schema-valid fixture) are consistent with the seams lens's independent boundary analysis; both agree machinery at HEAD was already correct and this PR adds coverage only.

Wirasm:
Holding this out of the merge queue: the regression test is good and worth landing, but the "does not reproduce at HEAD" conclusion rests on a timeline claim that is factually wrong, and the fixture's shape differs from the incident's in exactly the two ways that could hide the bug.

**Timeline.** `cadccdc3` was committed 2026-08-24 18:59 UTC and merged with #2779 shortly after. The failing run in #2794 (`1d7b08f3`) launched 20:48 UTC, almost two hours later, from a worktree rebased onto the merged #2779, and its wait was resumed by a server process running the same vintage. The run does not predate the fix; it ran on it.

**Shape.** The green fixture is a top-level `loop_group` (`grp`) with `until_bash: '[ $probe.output = "satisfied" ]'`: whole-output read, equality polarity, hand-persisted `grp.probe` row. The incident was an include-expanded group (`deliver__await-checks`, body node `ci-probe`) with a field access and the opposite polarity: `test "$ci-probe.output.state" != "pending"`.

That polarity difference matters because this PR's own mutation check demonstrates the kill mechanism: a missing restored row substitutes empty. Under the fixture's polarity, empty gives `[ '' = "satisfied" ]`, false, loop continues, failure is visible as a spurious second wait. Under the incident's polarity, empty gives `test "" != "pending"`, TRUE, and the group completes wrongly with the wait's own output as group output. That is byte-for-byte what the incident's event log shows.

So the live question is whether restoration holds for the incident's shape: an include-namespaced group id, and a field access on the restored value. If the restored output is keyed or looked up differently once include expansion prefixes the group (or `.state` fails to resolve against the restored text and the until_bash substitution degrades to empty rather than failing loud), HEAD still has the bug and this fixture cannot see it.

The discriminating test is this fixture with three changes: the group arrives via `include:`, the probe emits JSON, and until_bash reads `$probe.output.state != "pending"` with the probe stubbed to `{"state": "pending"}`. Green means #2794 can close with the timeline corrected in the PR body; red means this PR needs the fix, not just the pin. Until one of those happens the issue stays open.

Wirasm:
## Ready to merge

The corrected head now proves the incident's actual shell-quoting cause without changing executor behavior. The include-expanded unquoted JSON-field case restores `pending` and starts iteration 2; adding only the incident's extra double quotes completes after iteration 1. The prior fixture and exit-status findings are fixed, the adjacent #2803 missing-reference regression still detects loss of the resumed caller's required context, and the one pre-existing authoring-warning gap is already tracked in #2806 with PR #2811.

**Validation:** The three focused #2794/#2803 cases, full `@archon/workflows` suite, package type-check, repository lint and format check, clean merge-tree, and exact-head hosted checks pass.

**Resolved:** 2 · **Tracked follow-ups:** [#2806](https://github.com/coleam00/Archon/issues/2806)

### Findings

|---|---|---|---|
| `R1` | Important | The old fixture could not support the no-reproduction claim because it omitted the incident's include expansion, JSON field, quoting, and inequality shape. | FIXED |
| `R2` | Suggestion | The old mutation comment gave the wrong Bash exit status for an empty unquoted reference. | FIXED |
| `R3` | Important | The shared field-reference fixture was suspected of weakening #2803's required-context regression. | NOT A FINDING |
| `R4` | Important | `loop_group.until_bash` lacks the existing double-quoted output-reference warning. | TRACKED FOLLOW-UP |
| `R5` | Suggestion | CodeRabbit requested an explicit `boolean` annotation on a defaulted test-helper parameter. | DECLINED |

<details>
<summary><code>R1</code> — Replace the pin with the incident-shaped discriminating test</summary>

**Impact:** The earlier head could close #2794 with a regression that stayed green despite the reported workflow's actual condition.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:30273` builds the include-expanded JSON-field/inequality fixture. `packages/workflows/src/dag-executor.test.ts:30578` restores `deliver__await-checks.ci-probe = {"state":"pending"}` and asserts an iteration-2 pause. `packages/workflows/src/dag-executor.test.ts:30631` changes only the extra quoting and asserts premature completion. The PR body now records that the live run followed #2779 and that restoration works.

**Required outcome:** Keep the exact incident shape, its unquoted passing control, and the double-quoted reproducer; keep production executor code unchanged.

**Found by:** `tests`, operator dispute, discriminating validation

**Disposition:** FIXED — exact head `b63ee9e` contains both discriminating cases and the corrected timeline/root cause.

</details>

<details>
<summary><code>R2</code> — Correct the Bash exit-status explanation</summary>

**Impact:** The old comment gave maintainers the wrong shell failure mode, though both statuses were nonzero.

**Evidence:** The old scalar/equality fixture and its “exits 1” comment no longer exist in the exact-head diff. The replacement cases describe and assert the actual quoted and unquoted outcomes instead.

**Required outcome:** Remove or correct the inaccurate comment.

**Found by:** `seams`, CodeRabbit

**Disposition:** FIXED — the inaccurate comment disappeared with the replaced fixture.

</details>

<details>
<summary><code>R3</code> — The revised #2803 case still detects missing required context</summary>

**Impact:** If the concern were true, deleting the resumed caller's `requiredContext` could leave the adjacent #2803 regression green and silently re-open missing-reference completion.

**Evidence:** `packages/workflows/src/dag-executor.test.ts:30567` requires the contextual message prefix `Node 'deliver__await-checks' field 'loop_group.until_bash' cannot resolve`. Without the context passed at `packages/workflows/src/dag-executor.ts:4748`, the field resolver still throws but starts directly with `'$ci-probe.output.state' references node`, so the assertion fails. The test-quality reviewer verified those two exact messages. Whole-output empty-versus-throw behavior also remains covered by the direct substitution and loop/group #2803 tests.

**Required outcome:** No correction is required; preserve the contextual error assertion.

**Found by:** `simplify`; independently disputed and disproved by `tests`

**Disposition:** NOT A FINDING — the proposed mutation does not survive the modified test.

</details>

<details>
<summary><code>R4</code> — Warn on quoted loop-group output references</summary>

**Impact:** A workflow author can still double-quote a shell-quoted string reference in `loop_group.until_bash` and silently reverse a completion predicate, as the #2794 reproducer demonstrates.

**Evidence:** `packages/workflows/src/validator.ts:868` applies the shared warning to Bash nodes and `loop.until_bash`, but not `loop_group.until_bash`; `packages/workflows/src/dag-executor.test.ts:30631` proves the reachable wrong result. This gap exists in the exact base and the reproducer is intentional in this PR.

**Required outcome:** Apply the existing warning to top-level and nested loop groups without changing runtime substitution.

**Found by:** `seams`

**Disposition:** TRACKED FOLLOW-UP — [#2806](https://github.com/coleam00/Archon/issues/2806) owns the warning and docs correction; [PR #2811](https://github.com/coleam00/Archon/pull/2811) implements it independently.

</details>

<details>
<summary><code>R5</code> — Explicit annotation on an inferred test-helper default</summary>

**Impact:** None on behavior or type safety; TypeScript infers `quoteOutputRef` as `boolean`, and the helper is private to this test file.

**Evidence:** CodeRabbit's exact-head review suggested changing `quoteOutputRef = false` to `quoteOutputRef: boolean = false`. `bun run type-check` and repository lint both pass at the reviewed head.

**Required outcome:** No change is required for this test-only inferred parameter.

**Found by:** CodeRabbit

**Disposition:** DECLINED — this is redundant type spelling with no reachable invalid state or contract ambiguity.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | No additional findings; `R1` and `R2` fixed |
| seams | `R4`; `R1` and `R2` fixed |
| simplify | Proposed `R3`, disproved by the test-quality review |
| tests | No finding; `R1` and `R2` fixed, `R3` not a finding |

#### Validation

|---|---|---|
| `bun test src/dag-executor.test.ts --test-name-pattern '#2794|#2803: an included wait-resume'` | PASS | 3 pass, 0 fail; unquoted resumes to iteration 2, quoted completes, missing ref fails before the marker |
| `bun run test` (`packages/workflows`) | PASS | Full package suite completed with no failures |
| `bun run type-check` (`packages/workflows`) | PASS | TypeScript emitted no errors |
| `NODE_OPTIONS=--max-old-space-size=8192 bun run lint --max-warnings 0` | PASS | ESLint completed with zero warnings |
| `bun run format:check` | PASS | All files match Prettier style |
| `git diff --check 4ffff569...b63ee9e && git merge-tree --write-tree 4ffff569 b63ee9e` | PASS | Clean diff and merge tree `c7f7601a` against the actual base |
| Exact-head GitHub CI and CodeRabbit | PASS | Ubuntu, Windows, schema-upgrade, docker-build, and CodeRabbit passed at `b63ee9e` |

</details>

Issue #2783, opened 2026-08-24
design(core): replace archon-assist — the router's fallback deserves a deliberate mechanism, not a catch-all workflow


Wirasm:
## Goal

`archon-assist` is the chat router's dumping ground: when workflow routing produces no `/invoke-workflow`, the system falls back to a one-node prompt wrapper that runs a full-capability agent directly in the live checkout (`worktree.enabled: false`, #1546) — no gates, no review, no artifact discipline, none of the governance that defines every other run. It is simultaneously the most-invoked bundled workflow (~1.4k/month, partly fallback inflation) and the least governed. Its name is also load-bearing: the routing system prompt names it explicitly (`prompt-builder.ts`), so it cannot simply disappear with the legacy bundle (#2781 keeps it alive through the deprecation window).

## Desired outcome

- A deliberate design for what happens when a chat request matches no workflow: what the user experiences, what (if anything) runs, and under what governance.
- The routing layer's uncertainty is surfaced honestly rather than silently absorbed by a catch-all.
- Whatever runs on the fallback path has a governance posture chosen on purpose — not inherited from a 19-line wrapper written before the engine had gates, artifacts, or classes.
- The telemetry distinction between "user wanted general help" and "router couldn't decide" becomes visible, so the fallback stops inflating usage data.
- A migration story for the name: the routing prompt, docs, and any user habits around `archon-assist` get a stated transition.

## Non-goals here

No solution is prescribed — whether the answer is a redesigned assist workflow, a chat-native path with no workflow at all, an interactive-class run, or something else is the design's to propose. Context: #2781 (legacy deprecation holds assist through the window), the sdlc pack deliberately ships no general-assist equivalent, and `direction.md §prompt-computation` governs what belongs in prompts vs machinery.

Issue #2767, opened 2026-08-24
feat(workflows): replace the in-repo marketplace with hosted workflow packs


Wirasm:
Part of #2836.

## Problem

Archon's current marketplace is maintained inside the application repository:

- `packages/docs-web/src/data/marketplace.ts` is the registry;
- the docs site renders marketplace pages and serves `/workflows.json`;
- two GitHub Actions workflows lint and auto-review changes to that registry;
- publishing requires an Archon pull request;
- the CLI installs loose files and flattened directories into shared `.archon` paths.

That model predates packaged workflow discovery and makes the Archon application repository, docs deployment, CI, and release cadence responsible for a marketplace they should not own.

The replacement must not leave both systems alive.

## Outcome

Replace the in-repository marketplace with one Archon-hosted marketplace service, similar in product shape to skills.sh.

The hosted service owns:

- pack discovery and presentation;
- author submission/publication;
- catalog metadata;
- immutable source identity.

The Archon application owns only the client-side operation that searches the hosted marketplace and vendors a selected pack into a target repository.

There is no supported alternate marketplace registry, plugin marketplace, or in-repository catalog after this change.

## Canonical marketplace artifact

A marketplace artifact is always a workflow pack. A pack contains one or more packaged workflows:

```text
<pack>/
├── README.md
├── <workflow-a>/
│   ├── workflow.yaml
│   ├── commands/
│   └── scripts/
└── <workflow-b>/
    └── workflow.yaml
```

Publishing one workflow means publishing a pack with one workflow. Do not introduce a separate standalone-workflow marketplace format.

The pack is backed by an immutable public source, such as a GitHub repository folder at a commit SHA. This builds on the packaged-workflow primitive delivered by #2528.

## Author flow

```text
Author prepares a pack containing one or more workflows
  -> submits the pack to the hosted Archon marketplace
  -> hosted marketplace validates and publishes the immutable source
  -> no Archon application PR or docs deployment is involved
```

Publishing must be possible for both a one-workflow pack and a multi-workflow pack through the same operation and schema.

## User flow

```text
User finds a pack on the hosted marketplace
  -> CLI copies the complete pack into:
     <repo>/.archon/workflows/<pack>/
  -> user reviews and commits the vendored files
  -> normal packaged-workflow discovery loads them
  -> project catalog policy from #2835 decides whether the pack is available
```

[...]

PR #2833, opened 2026-08-27
Deprecate the legacy bundled workflows: move to defaults/legacy/, announce on every run

### LINKED ISSUE #2781: feat(workflows): deprecate the legacy bundled defaults — move to legacy/, warn on every run, remove in a coming release
## Problem

The pre-rewrite bundled workflows (the `archon-*` set under `.archon/workflows/defaults/`) predate composition, structured outputs, signatures, and the delivery-tail design; the sdlc pack (#2627) replaces them as the reference implementations. They cannot simply vanish: July telemetry shows real (if small) usage — assist, piv-loop, fix-github-issue, smart-pr-review and others still get invoked — so removal needs a deprecation window with a loud, actionable notice.

## Desired outcome

- The legacy defaults move to a `legacy/` folder and remain in the generated bundle for the deprecation window — still runnable, clearly separated.
- Every run of a deprecated workflow tells the user, at run start, that the workflow will be removed in an upcoming release, and names both exits: switch to the sdlc pack's equivalent, or copy the YAML into their own project/global `.archon/workflows/` to keep it (repo-override by name is the pinning mechanism — `direction.md §defaults-reference-implementations`).
- The notice is visible on every invocation surface the run reports to (CLI progress, chat/web), not only in server logs.
- Removal happens in a later release, not in this change.

## Boundaries

- Deliberately NOT part of #2627 — that PR stays additive (its own scope boundary, endorsed in review); this is the first follow-up after it merges, opening the #2123 endgame.
- Mechanism open (a declared deprecation field the engine surfaces, path-based detection, or otherwise) — whatever is chosen must pass the workflow-language admissibility test and the implicit-behavior rules (`direction.md §implicit-behavior`).

Wirasm:
## Problem and outcome

The pre-rewrite bundled workflows (`archon-*` under `.archon/workflows/defaults/`) predate composition, structured outputs, signatures, and the delivery-tail design; the sdlc pack (#2627) replaces them as the reference implementations. But July telemetry shows real usage (assist ~1.4k/mo, piv-loop, fix-github-issue, smart-pr-review…), so they cannot simply vanish — they need a loud deprecation window first.

- **Outcome:** All 20 legacy defaults are moved to `.archon/workflows/defaults/legacy/` and marked deprecated via a new declared `deprecated:` field. Every run now announces, at the invocation surface (CLI stderr, chat/web message, durable trace event), that the workflow will be removed in an upcoming release and names both exits: switch to the sdlc pack, or copy the YAML into project/global `.archon/workflows/` to keep using it.
- **Invariant:** A run of a deprecated workflow behaves identically otherwise — same routing (filenames unchanged, so command refs and override-by-filename still work), same execution, same output. The notice is additive and never blocks the run.
- **Scope boundary:** Nothing is removed yet — deletion happens in a later release after the window. `archon-assist` stays exempt until #2783 lands its successor. The marketplace remains unchanged and is outside this PR; #2767 owns that separate lane.

Closes #2781.

## Review guidance

- **Feedback requested:** Correctness of the deprecation mechanism's placement across the three surfaces (chat/web sender, executor event, CLI stderr), and the reserved-name guard for `defaults`.
- **Start here:** `packages/workflows/src/deprecation.ts` — the single formatting site all three surfaces print, byte-for-byte the maintainer's verbatim directive wording (removal in an upcoming release; both exits named).
- **Review order:** ① `packages/workflows/src/schemas/workflow.ts` (`workflowDeprecationSchema` — declarative metadata, not one-off machinery) → ② `packages/workflows/src/loader.ts` and `deprecation.ts` → ③ the three surfacing sites: `packages/cli/src/commands/workflow.ts`, `packages/core/src/orchestrator/orchestrator-agent.ts`, `packages/workflows/src/executor.ts` (+ event type in `store.ts`) → ④ the reserved-`defaults` guards in `scripts/generate-bundled-defaults.ts` and `packages/workflows/src/workflow-discovery.ts` → ⑤ spot-check two moved YAMLs for the declaration block.
- **Lower-attention areas:** The 20 moved YAML files are mechanical `git mv`s plus the identical four-line declaration block each (R083–R099 similarity confirms content is otherwise untouched). The regenerated `bundled-defaults.generated.ts` is verified by `bun run check:bundled*` and parity loops.
- **Known risk or uncertainty:** A pre-existing discovery quirk surfaced during testing: because the flat loader carries no path info, a project copy overriding a legacy default is reported with source label `bundled` even though its **content** correctly wins and the deprecation notice clears. Documented in the `loader.test.ts` case "a project copy with the same filename overrides and clears the notice" (its comment pins that only the label, not behavior, rides on this); fixing it means carrying paths through the flat loader — deliberately out of scope here.

## Solution

Mechanism over migration: the issue's maintainer comment asked for a **reusable** deprecation primitive, so instead of hardcoding a list of legacy names, a workflow can now declare:

```yaml
deprecated:
  message: Switch to the sdlc pack instead.
```

`formatDeprecationNotice()` composes the single canonical string from the marker, and each invocation surface prints it where the user actually starts the run:

| Surface | Site | Behavior |
|---|---|---|
| CLI | `emitDeprecationNotice` → stderr (`workflow.ts`) | Silent under `--json` stdout purity, same rationale as parse warnings |
| Chat/web | sibling of the parse-warning send in `orchestrator-agent.ts`, before the resume/background split | Best-effort; warn-log on delivery failure, never stops the run |
| Durable trace | `workflow_deprecation_notice` event from `executor.ts`, type registered in `store.ts` | Covers REST/detached starts |

Malformed `deprecated:` blocks (empty object, missing/empty message) **reject** the workflow rather than warn-and-drop — a broken announcement must not silently disable announcements. The field is registered in `KNOWN_WORKFLOW_NESTED_KEYS`, so a typo inside the block warns per #2213.

File mechanics: filenames were kept while moving into `legacy/`, preserving plain command refs and override-by-filename as the pinning mechanism. Because scanners otherwise fail "exactly one .yaml" on the new subfolder, both packaged scanners skip a `defaults` pack segment, and the bundle generator collects `defaults/legacy/*.{yaml,yml}` with an explicit cross-folder collision check.

The suffix-tier name ambiguities predicted in the issue's collision check turned out real (`run pr`, `run review`) and are pinned by tests: the resolver refuses and lists candidates rather than guessing, exactly as required during the coexistence window.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Legacy bundled workflows run silently; no indication they are scheduled for removal | Every start prints ⚠️ "…is deprecated and will be removed in an upcoming release. Switch to the sdlc pack instead. To keep using this workflow after removal, copy the workflow file into your project `.archon/workflows/` or your global `~/.archon/workflows/`." |
| **Failure behavior** | Malformed unknown nested keys warn via #2213 | Malformed `deprecated:` marker rejects the workflow at load; delivery failure of the notice never aborts the run |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1["archon workflow run piv-loop"] --> B2[runs silently]
  end
  subgraph After
    A1["archon workflow run piv-loop"] --> A2["⚠️ deprecation notice on stderr"]
    A3[user switches to sdlc pack OR copies YAML to .archon/workflows] --> A4["notice gone, override wins"]
    A2 --> A3
  end
```

## Validation

- `bun run type-check` — all packages + scripts tsconfig pass.
- `bun x eslint . --max-warnings 0` and `bun run format:check` — clean.
- `bun run test` — full workspace suite, exit 0, zero failures. New coverage includes: exact composed notice wording (`deprecation.test.ts`); malformed-marker rejection; `defaults/legacy` loads as `bundled` with unqualified command refs and no discovery error; project-copy override clears the notice; `pr`/`review` ambiguity refusals listing candidates; notice sent/not-sent at run start for orchestrator, executor, and CLI surfaces; bundle completeness/content-parity across the merged legacy folder.
- `bun run check:bundled && check:bundled-schema && check:bundled-skill && check:pi-vendor-map && check:capability-matrix && test:install` — all green.
- Live E2E: dev-mode run of deprecated `archon-piv-loop` shows the composed ⚠️ notice on stderr, with pure JSON on stdout under `--json`.
- **Not verified:** Nothing material to the outcome — every surface (chat/web sender beyond unit assertions of the message contract) is exercised through its tested orchestration path, though an interactive browser session against chat/web was not manually performed.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Filenames unchanged, so overrides-by-filename and plain refs keep working during the window; users who want continuity after removal copy the file into their own `.archon/workflows/` | Notice text names both exits; `loader.test.ts` override case |
| Security / permissions / data | None — purely additive metadata and messaging | Diff scope |
| Rollout / rollback | Ships behind nothing (declarative field default-absent means zero behavioral change for non-marked workflows; only the 20 moved YAMLs carry it) | `formatDeprecationNotice` returns `undefined` without the marker |
| Observability | Regressions visible via `workflow_deprecation_notice` events in the durable trace and warn-logs on failed notice delivery | `executor.ts`, `orchestrator-agent.ts` |
| Documentation / communication | Authoring guide gained a "Deprecating a Workflow" section | `authoring-workflows.md` |

## Links

- Closes #2781


Wirasm:
# Review Report — PR #2833 (deprecate legacy bundled defaults) — Round 2

## Verdict

**Ready.** The sole blocking finding from round 1 (R1) is fixed at `050375a0` and adversarially verified against the code — placement, payload-key conformance, fallback semantics, and tests all check out; no new Critical or Important findings were raised by any lens.

## Reviewed head SHA

`050375a019b055a7e1fab0983eb6b41345fe31e4` — verified verbatim against `git rev-parse HEAD` in the `task-feat-2781-deprecate-legacy-defaults` worktree, matching `scope.md`. This is the cursor for the next round. Delta reviewed: `9ce47550d8ca732cc1440489e70dfd2ae28e1b8a..050375a019b055a7e1fab0983eb6b41345fe31e4` (+40 lines across `event.ts` and `event.test.ts`; working tree clean).

## Findings

### Critical (0)

None reported by any lens; none opened by synthesis.

### Important (0)

None open. **R1 was resolved this delta** — see Prior findings for its carried-forward verdict.

### Rejected findings

None — every candidate raised across lenses survived verification this round. Synthesis independently checked each Important claim's cited evidence before accepting it; nothing failed falsification.

**R2 (Suggestion)** — Model `deprecated:` as a plain string scalar rather than a one-field object. `workflowDeprecationSchema` wraps exactly one field (`message: z.string().min(1)`, `packages/workflows/src/schemas/workflow.ts:166–171`) yet all 20 legacy YAMLs carry the byte-identical sentence ("Switch to the sdlc pack instead.") and the sole consumer reads `.message` (`packages/workflows/src/deprecation.ts`). A scalar preserves malformed-input hard-reject parity, needs no nested-key registration, and shrinks each YAML block to one line. Tradeoff: future deprecation metadata beyond one sentence would require promoting back to an object — speculative today, hence non-blocking. Attribution: **simplify** (round 1 §1), carried forward unchanged by consensus of code/seams/simplify this round. Not acted on by the delta, as expected.
Smallest correction (when/if picked up): change the schema field to `deprecated: z.string().min(1)`, add the loader branch shape, collapse each of the 20 YAML blocks to one line, drop the nested-key registration.

| ID | Severity | Claim (one line) | Verdict |
|---|---|---|---|
| R1 | Important | Web console had no reader/mapping for `workflow_deprecation_notice`; durable-trace notice rendered as mangled truncated raw JSON (`event.ts` fallback path) | **Fixed at `050375a019b055a7e1fab0983eb6b41345fe31e4`.** Verified by synthesis at that SHA, not taken on faith: (1) the new explicit case sits after every earlier discriminator in `toRunEvent()` and immediately before the raw-JSON fallback, so nothing shadows it; (2) key conformance holds across the serialization boundary — writer `executor.ts:2876–2880` emits `data: { workflowName, notice }`, reader reads `readString(data, 'notice')` at `event.ts:361`, contract comment `store.ts:146–147` names the same key; (3) `readString` returns `''` on missing/non-string (`event.ts:116–119`), so `notice \|\| '⚠️ This workflow is deprecated.'` degrades to a readable generic line, never raw JSON or an empty row; (4) rendered as `kind: 'text'`, mirroring the `workflow_parse_warnings` precedent, so visible without the System toggle; (5) the old defect's mechanism confirmed absent-with-the-fix: without the case the payload has neither `text` nor `message` keys, hitting the exact `JSON.stringify(data).slice(0, 200)` truncation R1 described; (6) conformance tests are real and pass under the package's actual runner (`bun test src/experiments/console/primitives/event.test.ts` → **42 pass / 0 fail**, re-executed by synthesis). |
| R2 | Suggestion (non-blocking) | Model `deprecated:` as a plain string scalar instead of a one-field object (`workflow.ts:166–171`) | **Still open** — unchanged by the delta (diff touches only the two web files); carried forward per scope note. Does not block readiness. |

Round-1 observations recorded for context (not findings): materializer writes bundled defaults flat under `defaults/` while the files now live under `defaults/legacy/` on disk (`workflow-source.ts:385–394`) — consequence-free today, worth one conformance test if anything ever keys on the folder name; the classic-UI surface filters unknown event types silently rather than mangling them and is itself retired behind `/legacy`.

## Lens roster

| Lens | Status | Outcome |
|---|---|---|
| code | Ran (full) | No new findings; verified R1 fixed (ordering, key conformance, runner-verified tests) |
| seams | Ran (full) | No new findings; verified R1 fixed across writer/contract/reader boundaries + render-route completeness; classic-UI carve-out noted |
| simplify | Ran (full) | No new findings; examined-and-rejected two simplification candidates (unreachable-looking fallback guard, shared helper for the two text-kind cases); R2 carried forward |
| tests | Disabled this round (no tests= input) | — |
| errors | Disabled this round (no errors= input) | — |
| comments | Disabled this round (no comments= input) | — |
| types | Disabled this round (no types= input) | — |
| docs | Disabled this round (no docs= input) | — |

No enabled lens failed to report — all three enabled lenses delivered full reports. Five optional lenses ran disabled by design; their absence is not evidence of cleanliness beyond what these three lenses actually examined.

Next-round directive: only the three lenses above were enabled; if a future round re-enables any optional lens, that lens must run **in full**, not delta mode.

## Canonical comment

Published to PR #2833: https://github.com/coleam00/Archon/pull/2833#issuecomment-5436360501 — marker `<!-- archon-review-report -->` on its first line, single canonical comment edited in place, read-back confirmed to match this report.

Issue #2747, opened 2026-08-22
design(workflows): between-run continuation — outcome, adoption, and open-work query for unattended runs


Wirasm:
## Problem

A run is the unit of governance, but work outlives runs. When a run ends without delivering — an honest gate refusal, a correction bound exhausted, an advisory `rooted: false` — the next run starts from zero: it re-derives context, re-discovers the branch state, and has no way to formally claim the prior run's artifacts or outcome. The connection between "run 1 implemented this and refused to flip on red CI" and "run 2 should verify and deliver that same branch" exists only in the operator's head, or in ad-hoc conventions like passing a prior report path as an input.

#2707 defines interactive and unattended workflow classes. An interactive run's continuation story is clear — a human is at the gate. An **unattended** run's continuation story is the gap this issue owns: unattended runs are *supposed* to terminate cleanly on refusals and bounds, which makes "what happens after a truthful non-delivery" a first-class lifecycle question, not an edge case.

## Desired outcome (design)

- **Outcome is queryable.** A completed run exposes what it concluded and what it left undone — outcome (`succeeded`/`failed`, already persisted), terminal report artifacts, branch/worktree it worked on — as one surface: "what did run X leave, and where."
- **A new run can adopt a prior run.** Start from the prior run's branch/checkout state and artifact store with recorded provenance (`adopted_from`), instead of pretending to be a fresh start. Adoption is explicit and operator-driven — the engine never infers it.
- **Open work is findable.** "Which runs in this project ended with work on the table" (refused gates, exhausted bounds, advisory negatives) is one query — the operator inbox that makes unattended automation actually operable.
- The response/decision vocabulary should be shared with #2707 (`workflow respond`, structured `{decision, text}`), so between-run continuation and at-gate continuation feel like one system.

## Evidence

- Delivery run `6e6c9f9e`: complete, correct work stranded behind a frozen verdict; the continuation had to be assembled by hand from the run's artifacts and branch.
- The SDLC delivery workflow already threads `prior_report` between runs **by convention** — cross-run review continuity works, but nothing in the engine records that run 2 continued run 1. The convention is the demand signal for the contract.

Cited by #2707 as the unattended-class sibling design.

Wirasm:
## Settled direction + accumulated evidence (24 Aug)

[...]

PR #2847, opened 2026-08-27
fix(core): restore SQLite workflow run adoption by hydrating run timestamps

### LINKED ISSUE #2845: fix(workflows): SQLite run adoption crashes before continuation starts
## Problem

A terminal workflow run cannot be adopted on the default SQLite database. `workflow run <name> --adopt <run-id>` crashes before creating the continuation run:

```text
Error: createdBefore.toISOString is not a function
```

This breaks the engine continuation invariant delivered by #2747: after a terminal failure, an operator must be able to start the same or another workflow on the existing branch/worktree and artifacts with recorded provenance.

## Real reproduction

On 27 August 2026, run `6608041e-4d0e-4315-a86a-46e36dd789d6` was terminal `failed`, its worktree and branch were still live, and it had no adopting run.

```bash
archon workflow run archon-implement \
  --adopt 6608041e-4d0e-4315-a86a-46e36dd789d6 \
  --config .archon/config.glm-flash.yaml \
  --detach \
  "Correct PR #2833 only"
```

The detached child reached adoption resolution, emitted the error above, and exited. No new workflow run row was created and the original worktree remained unchanged.

## Expected behavior

The engine resolves the terminal estate, reuses its live worktree or materializes its surviving branch, records `adopted_from_run_id`, exposes the prior artifact directory, and starts the requested workflow.

## Actual behavior

Adoption crashes while resolving isolation history, before worktree selection or run creation.

## Proven root cause

1. SQLite stores `remote_agent_workflow_runs.started_at` as TEXT and returns it as a JavaScript string.
2. `getWorkflowRun()` types the raw row as `WorkflowRun`, but `normalizeWorkflowRun()` only parses metadata. It does not hydrate timestamps, despite `WorkflowRun.started_at` being declared as `Date`.
3. `resolveWorkflowAdoption()` passes that runtime string to `findLatestByCodebaseAndWorkingPath()`.
4. The isolation lookup accepts only `Date` and calls `.toISOString()`, producing the observed exception.

A direct read-only reproduction against the real SQLite row proved the cutoff value was `2026-08-27 07:01:38`, `typeof string`, and not a `Date`. The failure occurs before provider selection, model configuration, worktree recovery, or detached execution can affect the result.

Relevant boundaries:

- `packages/core/src/db/workflows.ts` — workflow-run row normalization
- `packages/core/src/operations/workflow-adoption.ts` — terminal estate resolution
- `packages/core/src/db/isolation-environments.ts` — dialect-specific historical cutoff query
- `packages/cli/src/commands/workflow.ts` — adoption is resolved before run creation

## Fix boundary

The database/estate boundary must accept and normalize the timestamp representation SQLite actually returns while preserving the ownership rule `environment.created_at <= run.started_at`. Do not paper over this in the CLI or provider path.

The wider false runtime promise that every persisted `WorkflowRun` timestamp is already a `Date` should remain visible rather than being duplicated as one-off coercions across new consumers.

## Regression proof

Add a real-SQLite composition test that:

1. inserts a workflow run with the SQLite TEXT timestamp representation;
2. reads it through `getWorkflowRun()`;
3. passes that actual returned `started_at` into the isolation-history lookup;
4. seeds estate rows before and after the run cutoff;
5. proves adoption selects the newest pre-cutoff estate and excludes the later row.

This composition throws before the correction. Existing adoption unit tests fabricate `Date` values, and the current SQLite isolation test also supplies an explicit `Date`, so neither covers the real boundary.

## Links

- Continuation contract: #2747
- Dogfood PR whose failed run exposed this: #2833

Wirasm:
## Problem and outcome

On the default SQLite database, `workflow run <name> --adopt <run-id>` crashed with `Error: createdBefore.toISOString is not a function` before the continuation run was ever created, breaking the engine continuation invariant from #2747: after a terminal failure, an operator could not resume the same or another workflow on the surviving worktree with recorded provenance. Proven root cause (reproduced read-only against the real row in #2845): SQLite stores `remote_agent_workflow_runs` timestamps as TEXT (UTC `YYYY-MM-DD HH:MM:SS` from `datetime('now')`) and returns JS strings, but `normalizeWorkflowRun()` parsed only `metadata` while the `WorkflowRun` type declares those fields as `Date`. `resolveWorkflowAdoption()` therefore handed the raw string to `findLatestByCodebaseAndWorkingPath()`, whose estate-ownership cutoff calls `.toISOString()`.

- **Outcome:** Adoption on SQLite resolves the terminal estate and starts the continuation run; every consumer of workflow-run rows on both dialects now receives the `Date` the type promised.
- **Invariant:** The historical estate ownership cutoff `environment.created_at <= run.started_at` still holds, compared in the same UTC representation both dialects store.
- **Scope boundary:** The CLI (`workflow.ts`), provider path, and estate-lookup query are untouched — the correction stays at the database boundary as the issue requires. Pre-existing one-off tolerances (`parseDbTimestamp`, server `toISOString`, CLI `formatAge`) were intentionally left alone rather than duplicated.
- **Root cause:** `normalizeWorkflowRun()` failed to hydrate the TEXT timestamp columns SQLite actually returns, so a runtime string masked itself as the declared `Date`.

## Review guidance

- **Feedback requested:** Correctness of the UTC re-anchoring in `toHydratedTimestamp()` — it must be the exact inverse of the cutoff writers (`toDbDateParam` in workflow-events.ts, the inline conversion in isolation-environments.ts) so a hydrated `started_at` round-trips into the same TEXT comparison.
- **Start here:** `packages/core/src/db/workflows.ts:65` calling the new shared helper `toHydratedTimestamp()` (`packages/core/src/db/timestamps.ts:18`) — the load-bearing change: zone-less SQLite TEXT is re-anchored to UTC before parsing (JavaScript would otherwise read it as local time), and already-zoned strings keep their offset.
- **Review order:** `packages/core/src/db/workflows.ts` (hydration in `normalizeWorkflowRun`) → `packages/core/src/db/timestamps.ts` (`toHydratedTimestamp`, shared with isolation-environments) → `packages/core/src/db/isolation-environments.ts` and `workflow-events.ts` (same boundary family, see Solution) → `packages/core/src/db/workflows.adoption-timestamp.integration.test.ts` (real composition proof) → `packages/core/package.json` (test registration) → `CHANGELOG.md`.
- **Lower-attention areas:** The `package.json` and `CHANGELOG.md` entries are mechanical; the new test registration mirrors the existing mock.module-based integration-test pattern, and the two added cases in the existing integration tests (`isolation-environments.metadata`, `workflow-events.since`) follow each file's established pattern.
- **Known risk or uncertainty:** `toHydratedTimestamp()` trusts already-zoned strings by regex; all current writers are either `datetime('now')` TEXT or Postgres `Date` objects, so no zoned-string writer exists today — flagged for awareness if a writer with an offset is added later.

## Solution

`normalizeWorkflowRun()` now hydrates `started_at`, `completed_at`, and `last_activity_at` into real UTC dates for both dialects, via a new `toHydratedTimestamp()` helper. This is the smallest coherent fix: the false promise lived entirely at the row-normalization boundary, so repairing it there lets every downstream consumer (`resolveWorkflowAdoption`, the dag-executor's `last_activity_at.getTime()` staleness check, future consumers) work unchanged without per-call coercions. The hydration re-anchors the zone-less SQLite shape (`YYYY-MM-DD HH:MM:SS` → `...TZ`) instead of letting JavaScript parse it as local time, and is deliberately the inverse of the existing cutoff writers — which is what keeps the #2747 adoption ownership rule UTC-correct end to end. The same hydration is applied to `created_at` in `normalizeEnvironmentRow` (isolation-environments.ts), fixing the host-offset skew it caused in staleness math like cleanup's `isEnvironmentStale`, and `listRecentEvents`'s since cursor now converts through `toDbDateParam` rather than an ISO string, so same-day SQLite events are no longer sorted below the stored `YYYY-MM-DD HH:MM:SS` TEXT shape. Both sit at the identical database-read boundary family and reuse the inverse-consistency reasoning above. No coercion was added in the CLI or provider path.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `workflow run --adopt` on SQLite crashes before any run row is created; worktree left unchanged | Adoption resolves the terminal estate, reuses the live worktree or materializes the surviving branch, records `adopted_from_run_id`, and starts the continuation run |
| **Failure behavior** | Hard crash: `createdBefore.toISOString is not a function` during isolation-history resolution | No failure; when no estate predates the run cutoff, adoption resolves to `null` (existing null-handling path) |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[workflow run --adopt] --> B2[isolation lookup gets string] --> B3[crash before run creation]
  end

  subgraph After
    A1[workflow run --adopt] --> A2[getWorkflowRun hydrates Date] --> A3[estate cutoff compared in UTC] --> A4[continuation run created]
  end
```

## Architecture

```mermaid
flowchart LR
  A[CLI workflow run --adopt] --> B[resolveWorkflowAdoption]
  B --> C[getWorkflowRun — normalizeWorkflowRun]
  C ==> D[Hydrated Date boundary]
  D --> E[findLatestByCodebaseAndWorkingPath cutoff]
  E --> F[Continuation run]
```

### Changed seams

|---|---|---|
| `SQLite row → WorkflowRun` | TEXT timestamp columns hydrated to real `Date`s via the shared `toHydratedTimestamp()`, matching the declared type and Postgres shape | `packages/core/src/db/workflows.ts:65–69`; helper in `packages/core/src/db/timestamps.ts` |
| `SQLite row → IsolationEnvironment` | `created_at` zone-less TEXT hydrated the same way, un-skewing staleness math like cleanup's `isEnvironmentStale` | `packages/core/src/db/isolation-environments.ts:49` |
| `listRecentEvents` since cursor | ISO-string param replaced by dialect-aware `toDbDateParam`, so same-day SQLite events no longer sort below the stored TEXT shape (`T` > space) | `packages/core/src/db/workflow-events.ts:193` |
| `getWorkflowRun() → isolation lookup` | `started_at` now round-trips through `toDbDateParam` back into the same UTC TEXT the cutoff compares | `packages/core/src/db/workflows.adoption-timestamp.integration.test.ts` |

## Validation

- `cd packages/core && bun test src/db/workflows.adoption-timestamp.integration.test.ts` — 4 pass; failed 3/4 before the fix with `started_at` a string and the issue's exact `createdBefore.toISOString is not a function` exception — proves the real `getWorkflowRun()` → isolation-lookup composition and that the cutoff excludes the post-cutoff estate.
- `bun --filter @archon/core test` — 0 fail (full core chain including all real-SQLite integration tests) — proves no dialect regression.
- `bun --filter '*' --parallel test && bun test ./scripts/` — exit 0 (whole monorepo plus scripts suites).
- `bun --filter '*' type-check` — all packages exit 0.
- `eslint --max-warnings 0` via lint-staged on commit, `prettier --check` — clean.
- **Not verified:** Nothing material — the new test covers the exact real-database composition the issue demanded, and the full suite plus type-check cover the surrounding surface.

## Links

- Closes #2845
- Related #2747, #2833

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed workflow-run adoption failures on SQLite caused by timestamp formatting errors.
  - Ensured workflow and environment timestamps are interpreted consistently across SQLite and PostgreSQL.
  - Corrected recent-event queries so date filters return the expected results on SQLite.
  - Preserved accurate cutoff behavior when selecting eligible environments for adoption.


Wirasm:
# Review Report — PR #2847 (run 73ae3517-d563-4170-ab13-f702f2f46391, round 4)

## Verdict

**Ready** — no open Critical or Important findings remain: the delta's single functional change fixes R4 exactly as specified and is proven on real SQLite (7/7 integration tests pass), the two prose corrections (R5, R6) are arithmetically accurate, and every enabled lens delivered a full report.

## Reviewed head SHA

`75cdd80035d009ac097ef9094797465dea356f3c`

## Findings

### Critical

None.

### Important

None open. (R4 was Important last round; it is **fixed** in this delta — see the carry-forward table.)

### Rejected findings

None — every lens reported convergently that the delta is correct, and the adversarial verification of the one prior Important finding (R4) held up under direct re-check during synthesis:

- Fix verified by direct source read at `packages/core/src/db/workflow-events.ts:193`: `[workflowRunId, toDbDateParam(since)]`, operator still strict `created_at > $2` (`:191`).
- `toDbDateParam` confirmed module-local to `workflow-events.ts` (`:36-40`), dispatching on `getDatabaseType()`: SQLite branch emits `"YYYY-MM-DD HH:MM:SS"` (the exact `datetime('now')` TEXT storage shape); Postgres branch emits ISO, byte-identical to the old binding — no Postgres regression.
- Repo-wide grep confirms no raw `.toISOString()` remains in any query-param position in the module (only log-message formatting at `:251`).
- Test evidence reproduced locally: `bun test src/db/workflow-events.since.integration.test.ts` → **7 pass / 0 fail**, including the new same-day dialect-cursor test.
- The new test's positive case inserts `'2026-01-01 00:00:30'` in the zone-less writer shape with a same-instant cutoff (`2026-01-01T00:00:00.000Z` → param `"2026-01-01 00:00:00"`); `"…00:00:30" > "…00:00:00"` lexicographically proves the conversion, independent of wall clock. The future-cutoff negative case (`Date.now() + 60_000`) is non-vacuous (harness date 2026-08-27 ≫ inserted row).
- Prose arithmetic reproduced empirically during synthesis: stored `'2026-08-26 09:00:00'` = true instant 09:00Z; naive parse under `America/New_York` (August = EDT, UTC−4) yields `13:00Z` (+4h late — matches the corrected comment); under `Europe/Berlin` it yields `07:00Z` (−2h early — confirms the docstring's sign-flip claim).

### Suggestions (do not block)

New suggestions this round: none. All five lenses hunted the two-hop delta and reported zero new findings. Carried-forward do-not-block items appear in the table below.

## Prior findings (light mode — full carry-forward table)

| ID | Severity | Claim (one line) | Verdict | Evidence |
|---|---|---|---|---|
| **R1** | Important | `normalizeEnvironmentRow` leaves `created_at` (`Date` typed) as raw zone-less SQLite TEXT, poisoning cleanup staleness math and HTTP JSON passthrough | ✅ **Fixed at `739b343a`** (round 2→3; stable this round) | Guard at `packages/core/src/db/isolation-environments.ts:48`; delta does not touch it — verified again this round via direct read (:48-60 intact) and every read path still funnels through `normalizeEnvironmentRow`. |
| **R2** | Suggestion | Docstring claimed `toDbDateParam` exists in workflow-events.ts **and** isolation-environments.ts (only the former true) | ✅ **Fixed at `739b343a`** (round 2→3; stable this round) | Canonical docstring at `packages/core/src/db/timestamps.ts:10-14`; sole helper definition `workflow-events.ts:36`. Unchanged by the delta except that its "cursor MUST use that exact shape" requirement is now actually satisfied at every call site (`:193`, `:231`). |
| **R3** | Suggestion | Zoned-string regex branch in `toHydratedTimestamp` guards an input shape no current writer produces | ⚪ **Still open — accepted, deliberately kept, do-not-block** | Speculative forward-compat per round-1 disposition; body unchanged at `timestamps.ts:20`; no new writer of zoned strings found this round. Not a change request. |
| **R4** | Important | `listRecentEvents` built its `since` cutoff with raw `.toISOString()`, bypassing `toDbDateParam`; on SQLite same-day events were silently dropped from results (zone-less TEXT vs `'T'…'Z'` lexicographic sort); hazard goes live on first adoption of the exported function | ✅ **Fixed at `87b09cf4`** — independently verified during synthesis | Binding now `[workflowRunId, toDbDateParam(since)]` at `workflow-events.ts:193`; dialect converter module-local (`:36-40`); strict `>` direction preserved; Postgres path byte-identical to before. New real-SQLite regression test (`workflow-events.since.integration.test.ts:187-203`) covers the recovery (same-day row returned at same-instant cutoff) and the comparison direction (future cutoff → 0 rows): **7 pass / 0 fail** when executed during synthesis. Still zero production callers, so the fix lands before first adoption. |
| **R5** | Suggestion | New integration-test comment stated timezone arithmetic no host can produce ("13:00 local (= 17:00Z)") against inserted `'2026-08-26 09:00:00'` | ✅ **Fixed at `75cdd800`** — independently verified during synthesis | Comment now single-host and arithmetically exact: under `America/New_York` the naive parse yields 09:00 EDT (= 13:00Z), four hours after the true instant (empirically reproduced during synthesis); "east of UTC the sign flips" also checks out (Europe/Berlin → 07:00Z). Matches the actual inserted fixture value. |
| **R6** | Suggestion | Normalizer docstring claimed un-hydrated skew always "ages every environment"; drift sign actually flips across host timezones | ✅ **Fixed at `75cdd800`** — independently verified during synthesis | Docstring neutralized: "shifts every environment's computed age by the host's UTC offset (older-looking east of UTC, younger west) without re-anchoring" (`isolation-environments.ts:35-38`). Direction claims empirically verified both ways; no longer contradicts the corrected R5 worked example — the two sync'd comments agree. |

## Lens roster

| Lens | Status | Result |
|---|---|---|
| `code` | ✅ ran (full report present) | No new findings; R4/R5/R6 fixes verified with commands run (test suite executed, timezone parses reproduced); Postgres-path identity confirmed against the mocked unit suite |
| `seams` | ✅ ran (full report present) | No new findings; three-way documented-conversion duplication examined and carved out per project convention (`timestamps.ts:10-14` conformance index); strict-`>` boundary granularity acknowledged in-file |
| `simplify` | ✅ ran (full report present) | No new findings; the one-line fix reduces duplication (one owner, two call sites); deletion-and-consolidation candidates falsified |
| `comments` | ✅ ran (full report present) | No new findings; every changed comment checked against the code it describes; no remaining `.toISOString()` bypass for `created_at` comparisons |
| `types` | ✅ ran (full report present) | No new findings; suspected invalid states shown unreachable — every Date→param conversion now routes through `toDbDateParam`, every row through `normalizeEnvironmentRow` |
| `tests` | ⛔ disabled (input false) | — |
| `errors` | ⛔ disabled (input false) | — |
| `docs` | ⛔ disabled (input false) | — |

All five enabled lenses delivered full reports; no failed reporters, no gaps. Disabled lenses: `tests`, `errors`, `docs`.

## Verification record (synthesis)

- Head SHA confirmed via `git rev-parse HEAD` == `75cdd80035d009ac097ef9094797465dea356f3c`.
- R4 fix verified directly from source and by executing the touched suite (**7 pass / 0 fail**); param-position grep clean.
- R5/R6 arithmetic reproduced empirically (`TZ=America/New_York` → 13:00Z; `TZ=Europe/Berlin` → 07:00Z vs true 09:00Z).
- Every lens report present in the artifacts directory is accounted for in the roster above.

## Canonical PR comment

Published as one canonical comment carrying `<!-- archon-review-report -->` on its first line, replacing (edited in place) the prior-round comment at https://github.com/coleam00/Archon/pull/2847#issuecomment-5438326274.

Read-back status: **confirmed** — the comment body was fetched back via the REST API and diffed against this report: exact match except for trailing newlines added by the GitHub API itself.

Issue #2807, opened 2026-08-25
chore(sdlc): pack follow-through after #2627 — live validations and polish that deliberately don't block the merge


Wirasm:
## Problem

PR #2627 ships the sdlc pack in a merge-ready state: zero engine workarounds, 23/23 declared-expectation fixtures under `workflow test`, dynamic review scope, archon-prefixed names, per-run steering config. What remains is follow-through that deliberately should NOT block the merge: live validations that only real runs can provide, and small polish items. This issue is the hand-over so the merge doesn't wait on them and the next session doesn't rediscover them.

Context for whoever picks this up: doctrine and evidence live in `~/.prp/archon-75601ef6/bundle-rewrite-constitution.md` (read its "How to use" order first); live campaign state and mechanics in `~/.prp/archon-75601ef6/canvas/pack-campaign-helm.md`.

## Remaining work

**Live validations (each wants one observed occurrence, logged to the constitution's evidence appendix):**

- [ ] First real run through the dynamic review scope: `classify` judging the five specialist lenses from an actual PR diff and `resolve-scope` merging operator forces. Only dry-run covered so far. Judge the classifier's chose/skip reasons against the diff.
- [ ] First real pending-path CI wait through the hardened gate (probe → durable `wait:` → external-check `until_bash` → flip-ready CI preflight), confirming no wrongful flip recurs. Interacts with #2803/#2806 (until_bash reference/quoting semantics) — recheck the gate's predicate once those settle.
- [ ] First observed in-run recovery from a live 429 through #2798's patient retry (merged, never yet exercised — the resumes that followed it ran on calm quota).
- [ ] Steering via `--config .archon/configs/free-models.yaml` on a real detached run (dry-run verified; the legacy `user_ai_prefs` steering row is already deleted).

**Polish:**

- [ ] `outcome.py` now has three copies (fix, upkeep, stabilize) — the rule of three has fired. Extraction was blocked on include command-ref parameterization; recheck whether current `with:`/signature machinery can express it, and extract or re-log the blocker with current evidence.
- [ ] Fixture upkeep discipline: strict stub coverage means any new reachable node needs explicit stubs in every traversing fixture — `workflow test`'s missingStubs output is the worklist. Keep 23/23 the green bar (count grows with new fixtures).

[...]

PR #2861, opened 2026-08-27
fix(workflows): exec-code fixtures execute in a scratch worktree, not the caller's tree

### LINKED ISSUE #2851: fix(workflows): 'workflow test' exec-code fixtures depend on the operator's working-tree state (decline fixture flips on any untracked file)
## Summary

- What broke: a `workflow test` fixture that sets `exec-code: true` runs real `bash:`/`script:` nodes against the operator's **live working tree**. `sdlc/implement/fixtures/decline.stubs.yaml` relies on `scripts/assert-changed.py` failing when implement declined; that script passes on *any* tracked diff or untracked file outside `.archon/`. One stray untracked file in the operator's tree makes the fixture report `completed` instead of the expected `failed`, so a "deterministic" dry-run silently inverts based on tree state.
- When it started: since the sdlc pack merged (#2627) together with `workflow test` (#2772). Observed on dev `0113932f`.
- Severity: `minor` (false green/red in the 23/23 fixture bar that #2807 relies on)

## Steps to Reproduce

1. On a clean checkout of `dev`: `archon workflow test archon-implement` → `2 passed, 0 failed`
2. `touch scratch.txt` (any untracked file outside `.archon/`)
3. `archon workflow test archon-implement`

## Expected vs Actual

- **Expected**: same result as step 1 — fixtures are documented as never creating a run or contacting a provider, and should not depend on what else is in the tree.
- **Actual**:
  ```
  ✘ implement/fixtures/decline.stubs.yaml → archon-implement (completed)
      expected failed, dry-run reported completed
  1 passed, 1 failed
  ```
  Full `archon workflow test` on a dirty tree: `22 passed, 1 failed`; same commit in a clean `git worktree`: 23/23.

## Root cause

`.archon/workflows/sdlc/implement/scripts/assert-changed.py`:

```python
tracked = git("diff", "--name-only", "HEAD", "--", ":(exclude).archon")
untracked = git("ls-files", "--others", "--exclude-standard", "--", ":(exclude).archon")
if tracked or untracked:
    return 0          # "work happened" — but it was the operator's, not the run's
```

That is correct for a real run (which executes in a fresh worktree), but `--dry-run --exec-code` runs in the operator's cwd, where pre-existing changes are indistinguishable from run output.

## Proposed fix

Any of:
- `workflow test` / `--dry-run --exec-code` runs trusted code in a scratch worktree of `HEAD` (cheap; mirrors real-run isolation), **or**
- `workflow test` warns (or refuses) when an `exec-code: true` fixture meets a dirty tree, **or**
- the fixture is rewritten so `assert-changed` is stubbed and only pure-deterministic nodes execute.

The first option is the only one that makes every future `exec-code` fixture safe rather than just this one.

## Environment

- Platform: CLI
- Database: SQLite
- Running in worktree? No (that is the point — in a clean worktree it passes)
- OS: Windows 11 (platform-independent)

## Impact

- Affected workflows/commands: `archon workflow test`, `archon-implement` `decline` fixture, any future `exec-code: true` fixture
- Reproduction rate: Always, given a dirty tree
- Workaround available? Yes — run `workflow test` from a clean worktree
- Data loss risk? No

## Scope

- Package(s) likely involved: `workflows`, `cli`
- Module: `workflows:fixture-runner`, `workflows:dry-run`, sdlc pack `implement/scripts/assert-changed.py`

Wirasm:
## Problem and outcome

A `workflow test` fixture with `exec-code: true` ran its bash/script nodes in the operator's working tree, so a fixture verdict depended on that tree's state. The sdlc implement `decline` fixture flipped from `failed` to `completed` whenever any untracked file or tracked diff existed outside `.archon/` (#2851), and executed nodes could write into the caller's checkout.

- **Outcome:** `workflow test` now reproduces both halves of a real run's source/target split. SOURCE — named scripts and command files — resolves through one frozen `captureWorkflowSource` per invocation, the mechanism `workflow run` already uses. The TARGET that exec-code nodes execute against is a detached scratch worktree of the caller repo's HEAD, removed after each fixture. Clean and pre-dirtied caller trees produce identical verdicts, nothing executed can land in the caller's checkout, and a workflow that only breaks once captured now fails the fixture instead of passing it.
- **Invariant:** Real trusted bash/script execution, provider-free dry runs, pure-stub fixture behavior, and uncommitted workflow authoring must all behave as before. A capture copies the working tree rather than HEAD, so an author's unsaved-to-git edit is still exactly what runs.
- **Scope boundary:** Direct `workflow run --dry-run --exec-code` keeps its in-place behavior for both directories; the change is limited to `workflow test`.
- **Root cause:** The fixture runner passed the caller's `cwd` straight into `dryRunWorkflow`, and both `executeCodeNode` and source resolution then read it — so executed code saw and mutated the operator's tree, and the files it executed were whatever happened to be on disk at that moment.

## Review guidance

- **Feedback requested:** Whether the source/target split is now drawn where the engine already draws it, and the correctness of both lifecycles (capture and scratch worktree) on their failure paths.
- **Start here:** `packages/workflows/src/dry-run.ts:454` — `DryRunContext` now carries `sourceRoots` (where scripts and commands are READ from) beside `execWorkspace` (where code EXECUTES). Both default to `cwd`, which is what leaves the direct dry-run path untouched (`dry-run.ts:1177`).
- **Review order:** `packages/workflows/src/fixture-runner.ts:278` (`withCapturedFixtureSource`, one capture per invocation) → `fixture-runner.ts:321` (`withExecWorkspace`, one scratch worktree per exec-code fixture) → `fixture-runner.ts:355` (`disposeExecWorkspace`, which must never throw over a finished fixture's verdict) → `fixture-runner.ts:418` and `:508` (where each is threaded) → `packages/workflows/src/workflow-source.ts:169` (`workflowSourceConfigFrom`, extracted so `workflow run` and `workflow test` freeze the same directories) → `packages/workflows/src/fixture-runner.test.ts:461` onward.
- **Lower-attention areas:** The CLI usage-text line in `packages/cli/src/cli.ts` is documentation only.
- **Known risk or uncertainty:** An exec-code fixture in a repo with an unborn HEAD (no commits) fails with the wrapped `git worktree add` error. Isolation from HEAD is impossible there; an explicit failure is preferred over in-place execution.

## Solution

The engine already owns this doctrine: `script-discovery.ts` states that scripts resolve under the roots the workflow was read from, not the workspace it acts on, and `workflow-source.ts` defines a run's capture as the sole executable source for that run. `workflow test` now follows it.

`withCapturedFixtureSource` takes one `captureWorkflowSource` for the whole invocation and hands `capturedSourceRoots` to every fixture, disposing it in a `finally`. `dryRunWorkflow` gains an optional `sourceRoots` that defaults to `liveSourceRoots(cwd)`, and the three resolution sites — `discoverScriptsForCwd` and both `loadDryRunCommand` calls — read it. Taking the capture once rather than per fixture is also what makes the whole invocation a function of one snapshot, so an edit landing mid-test cannot change a verdict.

`withExecWorkspace` is unchanged: it verifies the caller is inside a git repository (hard failure otherwise), creates `git worktree add --detach <temp>/fixture-exec-<uuid> HEAD`, and disposes it with `git worktree remove --force`, with an `rm` fallback plus a warn log if git cannot remove it.

The config-to-source-policy projection that `prepareWorkflowSource` held inline is now `workflowSourceConfigFrom`. A second copy is how one entry point ends up freezing a different set of directories than another: a repo that points `commands.folder` outside `.archon/commands` would have had its commands captured by `workflow run` and missed by `workflow test`.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Exec-code fixtures executed in the operator's tree; verdicts flipped with unrelated dirty state (e.g. the sdlc implement `decline` fixture reported `failed` on any untracked file or tracked diff outside `.archon/`). | Exec-code fixtures execute in a scratch worktree of HEAD; caller tree state cannot affect the verdict. |
| **Source resolution** | Named scripts and command files were read live from the caller's checkout, so a `__file__`-relative script could still read and write the operator's tree, and a fixture passed on a workflow every real run would break. | Both are read from the invocation's capture, so `__file__`-relative access stays inside it and a capture-only failure fails the fixture the way it fails the run. Uncommitted edits are still what runs — a capture copies the working tree. |
| **Failure behavior** | An exec-code fixture outside a git repository silently executed in place. | It fails the fixture with an explicit "not inside a git repository" error, surfaced in `failureReason`. |

## Validation

- `bun test src/fixture-runner.test.ts` (packages/workflows) — 27 pass. New: the named script that executes is the captured copy (asserted on `import.meta.path`), an uncommitted script edit is still what runs, a moved `commands.folder` is captured and resolved, and two fixtures in one invocation resolve out of the same capture. Kept: clean-vs-dirty verdicts match, executed writes stay out of the caller checkout, both the scratch worktree and the capture are disposed, and an exec-code fixture outside a repo fails explicitly. `bun test src/commands/workflow.test.ts` (packages/cli) — 300 pass, including two new tests over the CLI's `sourceConfig` wiring.
- Every new test was seen red against a plausible faulty implementation: reverting the source-roots threading fails the captured-copy test; resolving source from the scratch worktree of HEAD instead of the capture fails three of them; dropping `commandFolder` from the capture call, or hardcoding a default `sourceConfig` in the CLI, fails the command-policy tests; skipping capture disposal fails the disposal test; moving the capture inside the fixture loop fails the one-capture test.
- The disposal guard was seen red the same way: with both `git worktree remove` and the fallback `rm` forced to fail, the pre-fix shape turns a passing fixture into a failure (the cleanup error replacing its verdict) while the guarded shape reports the real outcome.
- `bun run validate` — green end to end (5 generated-artifact checks, type-check across all packages, `lint --max-warnings 0`, `format:check`, `test:install`, full test suite).
- End-to-end with a dirty caller tree (9 modified tracked files plus an untracked `scratch.txt`): `bun run cli workflow test archon-implement` passed 2/2, the `decline` fixture still reporting its expected `failed` — the exact symptom #2851 reports. No stray files in the caller tree, and no `fixture-exec-*` or `fixture-source-*` directories left behind.
- **Pre-existing, not from this PR:** `bun run cli workflow test` reports 23 passed / 5 failed on this branch — and the identical five (`upkeep/update-delivered`, `ship/direct-delivered`, `ship/rooted-delivered`, `ship/planned-delivered`, `stabilize/fix-delivered`) fail on plain `origin/dev` at `1cbaf258`, verified in a clean detached worktree. They arrived with the sdlc deliver/review changes in #2862/#2863.
- **Not verified:** The unborn-HEAD edge case is covered by the wrapped-error path's construction but was not exercised in a commit-less repo.

## Links

- Closes #2851

## Summary by CodeRabbit

* **New Features**
  * `exec-code` fixtures now run in an isolated scratch worktree based on the repository’s current commit.
  * Fixture commands and scripts run from a captured source snapshot, including uncommitted changes.
  * Temporary execution workspaces and source snapshots are automatically cleaned up after testing.

* **Bug Fixes**
  * Prevented fixture execution from modifying the caller’s checkout.
  * Added clear failures for malformed configuration and for running `exec-code` fixtures outside a Git repository.

* **Documentation**
  * Updated workflow guidance and CLI help to explain isolated fixture execution, source snapshots, and dry-run behavior.


Wirasm:
# Review report — PR #2861 round 3 (light)

## Verdict

**ready: true — action: none.** All eight enabled lenses delivered reports, no open Critical or Important findings remain: the round-2 blocker R4 is fixed by this delta and every prior finding re-verified at head `8d5b3815`.

## Accepted contract

From the PR body: exec-code fixtures (`workflow test` with `exec-code: true`) execute bash/script nodes in a detached scratch worktree of the caller repo's HEAD, created under the Archon temp dir and removed after each fixture. Clean and pre-dirtied caller trees produce identical verdicts, and nothing executed may write into the caller's checkout. Closes #2851.

Preserved invariants: real trusted bash/script execution (no stubbing of exec nodes); provider-free dry runs and pure-stub fixture behavior unchanged; script/command *source resolution* keeps reading the caller's tree (`ctx.cwd`); direct `workflow run --dry-run --exec-code` stays in-place (`execWorkspace` defaults to `cwd`).

Boundaries: the change is limited to the `workflow test` execution-workspace boundary; callers outside a git repository fail hard; an unborn-HEAD caller fails with the wrapped `git worktree add` error (accepted risk).

## Reviewed head SHA

`8d5b3815e182cb9e021c7088156de5cd4c7c327c` (delta reviewed: `2d8ba4dee9686ef8698f1337d78bb0c0300a6adc..8d5b3815`, one docs-only commit touching `.claude/skills/archon-cli/authoring-workflows/node-reference.md`).

## Findings

### Critical / Important

None open. The round-2 blocker R4 was verified fixed this round (see Prior findings).

### Rejected findings

None — every lens reported no new findings this round; no lens evidence was disproved.

- **S3 (carried forward, open)** — the direct dry-run default `execWorkspace: options.execWorkspace ?? options.cwd` at `packages/workflows/src/dry-run.ts:1159` remains unpinned by a direct test. Unchanged by this delta; smallest correction is a one-case test that the direct `--dry-run --exec-code` path keeps `cwd` as the execution workspace. Does not block readiness.

## Prior findings (light mode — carried forward)

| ID | Severity | Claim | Verdict | Evidence |
|---|---|---|---|---|
| R4 | Important (block in round 2) | Skill doc claimed "your working tree's state cannot change a verdict" while source resolution deliberately reads the caller's tree | **Fixed at `8d5b3815`** (this delta) | The rewritten sentence at `node-reference.md:316-320` is factually correct against the code, re-verified directly this round: executed nodes run with `cwd: ctx.execWorkspace` (`dry-run.ts:671`), the workspace is a detached worktree of HEAD (`fixture-runner.ts:263`), while `discoverScriptsForCwd(ctx.cwd)` (`dry-run.ts:645`), `loadDryRunCommand(ctx.cwd, …)` (`dry-run.ts:805`, `:1021`), and caller-tree workflow discovery (`fixture-runner.ts:397-398`) all read the caller's tree. Pinned by `fixture-runner.test.ts:362-383` (named script written post-commit). Repo-wide grep for the old claim returns nothing. |
| R1 | Important | Caller-tree source resolution pinned by a test writing the named script post-commit | Still fixed (at `2d8ba4de`) | `fixture-runner.test.ts:362-383` present at HEAD; script written after `initGitWithCommittedFile`, fixture completes. |
| R2 | Important | Scratch-worktree disposal pinned by tests | Still fixed (at `2d8ba4de`) | `fixture-runner.test.ts:385-413`: fresh `ARCHON_HOME`, `git worktree list --porcelain` shows no `fixture-exec-` residue and exactly one worktree; temp dir empty. |
| R3 | Important | All "clean tree" causality claims removed from skill docs | Still fixed (at `2d8ba4de`) | Grep for "cannot change a verdict" / "working tree's state" across `.claude/skills` and docs-web markdown: no matches. |
| S1 | Suggestion | Redundant `mkdir` removed; `git worktree add` creates leading dirs | Still fixed | `fixture-runner.ts:259-261`: comment records the invariant; no `mkdir` call. |
| S2 | Suggestion | Guard spread collapsed; `execWorkspace` passed unconditionally | Still fixed | `fixture-runner.ts:402` passes `execWorkspace: workspace` unconditionally; single default at `dry-run.ts:1159`. |
| S3 | Suggestion | Direct dry-run `execWorkspace ?? cwd` default unpinned by a test | Still open, non-blocking | See Suggestions above. |

## Discoveries

**3 accepted adjacent discoveries** (none are scope conflicts; adjacent discoveries do not affect readiness):

1. "Git-repo precheck conflates all rev-parse failures into 'not inside a git repository'" (from `review-code.json`) — verified against `fixture-runner.ts:254-258`; the catch does not inspect the git error, so dubious-ownership or missing-git failures report the wrong reason.
2. "Inline bun/python exec nodes inherit the caller's stale PWD env var during fixture isolation" (from `review-errors.json`) — verified mechanism: `dry-run.ts:671-686` spreads `process.env` with `cwd: ctx.execWorkspace`; non-shell children (`bun -e`, `python -c`) inherit a stale `PWD`, so `$PWD`-relative code can escape the isolation contract silently.
3. "Exec-code scratch worktree shares the caller repo's ref store: executed git commands can write branches, tags, and objects into the caller's .git" (from `review-seams.json`) — verified mechanism: `git worktree add --detach` (`fixture-runner.ts:263`) shares refs and the object database with the caller repo; the isolation guarantee covers the working tree, not `.git`.

Full records: [`discoveries.json`](../discoveries.json) · [`discoveries.md`](../discoveries.md). These are adjacent findings outside the accepted contract; none of them blocks this PR's readiness.

## Lens roster

| Lens | Status | Result |
|---|---|---|
| code | Reported | No new findings; verified R4 fix against execution and resolution paths |
| seams | Reported | No new findings; verified both doc surfaces and the code contract agree |
| simplify | Reported | No new findings |
| tests | Reported | No new findings; delta is docs-only, existing tests pin the corrected claim |
| errors | Reported | No new findings; failure paths examined and visible |
| comments | Reported | No new findings |
| types | Reported | No new findings; type shapes unchanged this round |
| docs | Reported | No new findings; no other doc surface carries the old claim |
| *(no lens disabled or missing)* | — | — |

All enabled lenses delivered reports. Next round should still begin by reading this report, but may operate in delta mode unless the branch moves past `8d5b3815e182cb9e021c7088156de5cd4c7c327c`.

## Publication

Canonical PR comment (marker `<!-- archon-review-report -->`) edited in place: https://github.com/coleam00/Archon/pull/2861#issuecomment-5440501694 — read-back verified to match this report.

Wirasm:
Holding this for one focused correction before merge. The execution-workspace half is right; source resolution is drawn on the wrong side of an existing engine boundary.

**Keep**

- The `withExecWorkspace` worktree lifecycle, the `execWorkspace` seam in `dryRunWorkflow`, and the explicit outside-a-repo failure.

**Change: resolve sources from a frozen capture, not the live caller tree**

The engine already owns this doctrine. `script-discovery.ts` states that scripts resolve under the roots the workflow was read from, not the workspace it acts on, and real runs freeze those roots with `captureWorkflowSource` (`workflow-source.ts`): the capture is the sole executable source for the run, and it copies live files at launch, so uncommitted authoring is included. This PR gives fixtures live un-frozen sources plus an isolated workspace. That is half of the real-run model, and it leaves three holes:

1. A script using `__file__`-relative paths still reads, and can write, the operator's checkout. "Nothing executed can land in the caller's checkout" only holds for cwd-relative paths.
2. `workflow test` can never catch capture-only failures. `workflow-source.ts` documents this class ("a script that worked live and failed only once captured"). Under this PR a fixture passes on a workflow that every real run breaks.
3. The verdict is not a function of a snapshot; an edit landing mid-test changes it.

Concrete correction:

- Take one `captureWorkflowSource` per `workflow test` invocation (not per fixture) and resolve through `capturedSourceRoots`.
- Thread the roots into dry-run resolution: `discoverScriptsForCwd(ctx.cwd)` at `dry-run.ts:638` currently omits roots; command loading at `dry-run.ts:798` and `dry-run.ts:1014` reads `ctx.cwd` the same way. `dryRunWorkflow` should accept optional source roots defaulting to live roots so the direct `workflow run --dry-run --exec-code` path keeps its current in-place behavior.
- Flip the test asserting that named scripts resolve from the caller tree; it encodes the boundary this correction removes. Replace it with: named scripts resolve from the capture, and an uncommitted script edit made before the test run is still what executes. The existing clean-vs-dirty verdict and leak tests stay.
- Update the docs lines added in this PR that describe caller-tree resolution.

Uncommitted authoring keeps working because capture copies live state at test start. The scope boundary for direct dry-run is unchanged.

Wirasm:
## Ready to merge

Correction-verification pass over `cb45b1cb..50b81ba5` (one commit, no merge, 4 files: `fixture-runner.ts`, `fixture-runner.test.ts`, `dry-run.ts`, `workflow.test.ts`). All six findings from the prior review are verified fixed by direct inspection of the diff and by reading each new/changed test against the production code it exercises — not by trusting the implementer's disposition comment, though it matches independently. No new defect found in the correction itself.

R1's fix is exactly as claimed: `disposeExecWorkspace` now logs the git-removal error immediately in its catch block (before attempting the fallback `rm`), returns early on success, and wraps the fallback `rm` in its own `.catch`, so neither failure — nor a double failure — can escape `withExecWorkspace`'s `finally` and overwrite a fixture's real verdict. R2, R3, and R5 each now have a genuinely discriminating test: R2's new `workflowTestCommand` test asserts the exact `sourceConfig` object built from a mocked non-default `commands.folder`, not just that `runFixtures` was called; R3's disposal test now checks both `fixture-exec-*` and `fixture-source-*` are absent; R5's new test runs two fixtures in one invocation and asserts both executions report the identical `fixture-source-<uuid>`, which would fail if capture were accidentally moved inside the per-fixture loop. R4 removed the `sourceConfig?` optionality and its `?? DEFAULT_WORKFLOW_SOURCE_CONFIG` fallback entirely — it's now a required field, matching `capturedSourceRoots`'s own no-default contract one call down — and the test file's local wrapper defaults it only for tests that don't care about command policy, without weakening the real production type. R6's dead stub now throws with a comment explaining why the branch is unreachable, confirmed by tracing `loadCommandPrompt`'s actual `loadConfig` fallback condition (`sourceRoots?.config.load_default_commands`, which is always defined here).

**Validation:** `bun run validate` — PASS, exit 0, full repo, run fresh on `50b81ba5` in a new worktree. `bun test src/fixture-runner.test.ts` — 27/27 pass (up from 26; R5's new test added). `bun test src/commands/workflow.test.ts` — 300/300 pass (up from 298; R2's two new tests added). CI on `50b81ba5`: `test (ubuntu-latest)` PASS, `docker-build` PASS, `schema-upgrade` PASS, `CodeRabbit` PASS. `test (windows-latest)` FAILED — see note below; judged pre-existing and unrelated to this correction, not a blocker.

**Windows CI note (not a finding, not blocking):** The Windows job failed entirely inside `workflow-source.test.ts` — a file this correction commit does not touch — with `EBUSY`/`EFAULT` errors on `rm` of temp capture directories and 5s test timeouts (`the capture is authoritative, not advisory`, `the source's own settings travel with its bytes`, `continuing a run resolves with the settings it froze`). This is a Windows temp-directory handle-release race, not a logic failure: the identical suite passed locally on this exact head (`bun run validate`, macOS), the same Windows job was green on the immediately prior head `cb45b1cb` roughly 30 minutes earlier, and this repository has an open, actively-tracked general issue for exactly this class of flake — #2306, "ci: windows test timeouts cluster on subprocess-spawning tests" — plus a same-day precedent fix for a different file's instance of it (`aad8aa7a`, "tolerate Windows handle release during detached cleanup"). Recommend rerunning the Windows job before merge to confirm, but this does not change the verdict.

### Findings

|---|---|---|---|
| `R1` | Important | Exec-workspace dispose fallback can throw and mask the real fixture verdict | FIXED |
| `R2` | Important | CLI wiring of `sourceConfig` into `workflow test` is invoked by tests but never observed | FIXED |
| `R3` | Suggestion | Source-capture directory disposal (`fixture-source-*`) is untested | FIXED |
| `R4` | Suggestion | `RunFixturesOptions.sourceConfig` is optional with a silent default, reopening a seam the codebase deliberately closed one call deeper | FIXED |
| `R5` | Suggestion | "One capture per invocation, not per fixture" is a stated invariant with no test that would catch it regressing | FIXED |
| `R6` | Suggestion | Dead stub object in `loadDryRunCommand`'s unreachable `loadConfig` fallback branch | FIXED |

<details>
<summary><code>R1</code> — Exec-workspace dispose fallback can throw and mask the real fixture verdict — FIXED</summary>

**Original impact:** `disposeExecWorkspace`'s fallback `rm()` wasn't guarded, so a double cleanup failure could escape `withExecWorkspace`'s `finally` and overwrite a fixture's real verdict with a cleanup error, and the git-removal error could be lost entirely if the fallback `rm` then threw.

**Fix verified:** `packages/workflows/src/fixture-runner.ts:355-378` (at `50b81ba5`). `disposeExecWorkspace` now: (1) tries `git worktree remove --force` and returns immediately on success; (2) on failure, logs `fixture_runner.exec_workspace_dispose_failed` with the git error *inside the catch block*, before any fallback runs — so that log always fires regardless of what happens next; (3) falls through to `rm(workspace, { recursive: true, force: true }).catch(...)`, itself logging `fixture_runner.exec_workspace_remove_failed` on failure rather than throwing. Neither failure path can propagate out of the function, so `withExecWorkspace`'s `finally` (`fixture-runner.ts:336-340`, unchanged) can no longer have its return value replaced by a cleanup error. Both halves of the implementer's claim — verdict preserved, and the git error captured even when the `rm` also throws — check out from reading the control flow directly, not just the log-call ordering.

**Verified by:** `prp-core:pr2861-review` (this session), direct code inspection of `50b81ba5`.

</details>

<details>
<summary><code>R2</code> — CLI wiring of <code>sourceConfig</code> into <code>workflow test</code> is invoked by tests but never observed — FIXED</summary>

**Original impact:** The only `workflowTestCommand` test touching `runFixtures` used `objectContaining({ cwd: ... })`, which ignored `sourceConfig` entirely — a regression dropping that wiring would pass every existing test.

**Fix verified:** `packages/cli/src/commands/workflow.test.ts:9157-9198` (new tests, at `50b81ba5`). `"freezes the repo's own command policy, not the default folders (#2851)"` mocks `core.loadConfig` to resolve `{ commands: { folder: '.archon/prompts' }, defaults: { loadDefaultCommands: false } }`, then asserts `runFixtures` was called with `sourceConfig: { load_default_workflows: true, load_default_commands: false, command_folder: '.archon/prompts' }` — a full-value assertion, not a partial `objectContaining` on an unrelated key. A companion test, `"fails the fixture run when the config is unreadable instead of freezing defaults"`, asserts a rejecting `loadConfig` propagates as a thrown error and `runFixtures` is never called — covering the fail-clearly branch this correction also touches. `bun test src/commands/workflow.test.ts`: 300/300 pass, including both.

**Verified by:** `prp-core:pr2861-review` (this session), direct code inspection and test execution on `50b81ba5`.

</details>

<details>
<summary><code>R3</code> — Source-capture directory disposal is untested — FIXED</summary>

**Original impact:** The disposal test checked only `fixture-exec-*` was absent after a run; a regression leaking the heavier `fixture-source-*` capture directory would go unnoticed.

**Fix verified:** `packages/workflows/src/fixture-runner.test.ts:596-616` (at `50b81ba5`, test renamed to `"disposes both the scratch worktree and the source capture (#2851)"`). The final assertion now filters `readdirSync(join(home, 'temp'))` for entries starting with either `fixture-exec-` or `fixture-source-` and asserts the combined list is empty.

**Verified by:** `prp-core:pr2861-review` (this session); `bun test src/fixture-runner.test.ts` 27/27 pass.

</details>

<details>
<summary><code>R4</code> — <code>RunFixturesOptions.sourceConfig</code> optional with a silent default — FIXED</summary>

**Original impact:** `sourceConfig` was optional on the public `RunFixturesOptions` type with a silent `?? DEFAULT_WORKFLOW_SOURCE_CONFIG` fallback, reopening the same defaulting seam `capturedSourceRoots`'s `config` parameter was deliberately hardened against one call deeper.

**Fix verified:** `packages/workflows/src/fixture-runner.ts:246-256` — `sourceConfig: WorkflowSourceConfig` is now required (no `?`), with a comment explicitly citing `capturedSourceRoots`'s precedent. The unused `DEFAULT_WORKFLOW_SOURCE_CONFIG` import and the `?? DEFAULT_WORKFLOW_SOURCE_CONFIG` fallback in `runFixtures` (`fixture-runner.ts:417`) are both removed — `withCapturedFixtureSource` now receives `options.sourceConfig` directly. Confirmed this doesn't strand any other caller: `packages/cli/src/commands/workflow.ts` remains the only production caller and already supplies it correctly (unchanged in this correction); `packages/workflows/src/fixture-runner.test.ts:65-73` adds a local test-only wrapper — `Omit<RunFixturesOptions, 'sourceConfig'> & { sourceConfig?: WorkflowSourceConfig }`, defaulting to `DEFAULT_WORKFLOW_SOURCE_CONFIG` — that keeps the 20+ existing test call sites which don't care about command policy compiling, without weakening the real exported type. This is the correct way to satisfy the constraint: production callers are forced to choose explicitly; tests that don't care about this axis get a documented, scoped default.

**Verified by:** `prp-core:pr2861-review` (this session), direct code inspection; type-check passes clean in `bun run validate`.

</details>

<details>
<summary><code>R5</code> — "One capture per invocation, not per fixture" untested — FIXED</summary>

**Original impact:** No test distinguished "one capture shared across every fixture in an invocation" from "one capture per fixture" — a regression moving the capture inside the per-fixture loop would pass every existing test.

**Fix verified:** `packages/workflows/src/fixture-runner.test.ts:490-527` — new test `"shares one capture across every fixture in an invocation (#2851)"`. Two fixtures (`first.stubs.yaml`, `second.stubs.yaml`) target the same workflow in one `runFixtures` call; the workflow's script appends its own `import.meta.path` to a probe file on each execution. The test asserts both recorded lines carry the same `fixture-source-<uuid>` prefix. This genuinely discriminates: capturing per-fixture would produce two distinct UUIDs.

**Verified by:** `prp-core:pr2861-review` (this session); `bun test src/fixture-runner.test.ts` 27/27 pass, including this test.

</details>

<details>
<summary><code>R6</code> — Dead stub in <code>loadDryRunCommand</code>'s unreachable fallback branch — FIXED</summary>

**Original impact:** A stubbed `loadConfig` object was passed to `loadCommandPrompt` from a branch that could never execute, once `ctx.sourceRoots` became always-defined.

**Fix verified:** `packages/workflows/src/dry-run.ts:484-495` — the stub object is replaced with `loadConfig: () => { throw new Error('dry-run resolves commands from its source roots, never from config'); }`, with a comment explaining the reachability argument. Independently traced the claim against `packages/workflows/src/executor-shared.ts:363-366` (`loadCommandPrompt`): it calls `deps.loadConfig` only when `sourceRoots?.config.load_default_commands === undefined`, i.e. when `sourceRoots` itself is undefined or its `.config.load_default_commands` is undefined. `DryRunContext.sourceRoots` is a required, always-populated field (confirmed in the original review), and `WorkflowSourceConfig.load_default_commands` is a required `boolean` on every construction path (`DEFAULT_WORKFLOW_SOURCE_CONFIG`, `workflowSourceConfigFrom`) — so the branch is genuinely unreachable from `dry-run.ts`, and throwing here now surfaces a real error instead of silently masking a future contract change.

**Verified by:** `prp-core:pr2861-review` (this session), direct code inspection of both `dry-run.ts` and `executor-shared.ts` on `50b81ba5`.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1, R3, R6 — all verified FIXED |
| `seams` | R1, R4 — all verified FIXED |
| `tests` | R2, R3, R5 — all verified FIXED |

This was a focused `--verify-corrections` pass (single coordinator verification against the bounded `cb45b1cb..50b81ba5` diff), not a fresh multi-agent dispatch — per the skill's correction-verification mode, since the correction diff was small (4 files, 1 commit) and every finding traces to a specific, checkable code location.

#### Validation

|---|---|---|
| `bun run validate` (fresh worktree pinned to `50b81ba5`) | PASS | Exit code 0; full repo, all gates green. |
| `bun test src/fixture-runner.test.ts` (packages/workflows) | PASS | 27 pass, 0 fail (was 26 at `cb45b1cb`; R5's test added). |
| `bun test src/commands/workflow.test.ts` (packages/cli) | PASS | 300 pass, 0 fail (R2's two new tests included). |
| `git merge-base --is-ancestor cb45b1cb 50b81ba5` | PASS | Confirms a valid, bounded correction diff — no rebase or history rewrite. |
| CI `test (ubuntu-latest)` | PASS | https://github.com/coleam00/Archon/actions/runs/33109426740/job/98647986416 |
| CI `docker-build` | PASS | https://github.com/coleam00/Archon/actions/runs/33109426740/job/98647986583 |
| CI `schema-upgrade` | PASS | https://github.com/coleam00/Archon/actions/runs/33109426740/job/98647986300 |
| CI `CodeRabbit` | PASS | Review completed. |
| CI `test (windows-latest)` | FAIL (judged pre-existing, not PR-caused) | https://github.com/coleam00/Archon/actions/runs/33109426740/job/98647986587 — failures confined to `workflow-source.test.ts` (untouched by this correction), `EBUSY`/`EFAULT` on temp-dir `rm`, matches open issue #2306; same job was green on the immediately prior head `cb45b1cb`. |
| `bun run cli workflow test` | NOT RE-RUN | Confirmed unchanged at 23 passed / 5 failed by the implementer's own validation; the 5 failures remain the pre-existing set tracked on #2807, out of scope per task instructions. |

</details>

Wirasm:
Review findings addressed at `50b81ba5`. All six fixed, none deferred.

**R1 — exec-workspace dispose fallback can mask the verdict — FIXED.** `disposeExecWorkspace` (`fixture-runner.ts:355`) now returns on success and, on failure, logs the git error before attempting the fallback `rm`, which is itself `.catch`-guarded. Nothing escapes into `withExecWorkspace`'s `finally`. The review noted the git error was also lost whenever the fallback threw, since the warn sat after it; that ordering is fixed too.

Seen red: with both `git worktree remove --force` and the fallback `rm` forced to reject, the pre-fix shape turns the passing clean-tree fixture into a failure — the cleanup error replacing its verdict, exactly as described — while the guarded shape reports `failed: 0`.

**R2 — CLI `sourceConfig` wiring unobserved — FIXED.** Added `freezes the repo's own command policy, not the default folders (#2851)` to `workflow.test.ts`: it mocks `loadConfig` with `commands.folder: '.archon/prompts'` and `defaults.loadDefaultCommands: false`, then asserts the exact `sourceConfig` object `runFixtures` receives. Replacing the call with a hardcoded default config fails it. Added a second test covering the malformed-config path, asserting `runFixtures` is never reached.

**R3 — capture disposal untested — FIXED.** The disposal test now asserts neither `fixture-exec-*` nor `fixture-source-*` survives under `ARCHON_HOME/temp`, and is renamed to say so. Removing the capture `rm` fails it.

**R4 — optional `sourceConfig` reopens a closed seam — FIXED, not deferred.** Agreed with the reasoning: `runFixtures` is a genuine public subpath export, and the argument that closed `capturedSourceRoots` applies unchanged one call up. `sourceConfig` is now required and the `?? DEFAULT_WORKFLOW_SOURCE_CONFIG` fallback is gone; omitting it is a compile error. The test wrapper passes `DEFAULT_WORKFLOW_SOURCE_CONFIG` once on behalf of tests that are not about command policy, and the command-folder test now contrasts two explicit policies rather than present-vs-absent.

**R5 — one-capture-per-invocation untested — FIXED.** Added `shares one capture across every fixture in an invocation (#2851)`: two exec-code fixtures beside one workflow, one `runFixtures` call, the script appending `import.meta.path` each time; both lines must carry the same `fixture-source-<uuid>`. Moving the capture inside the fixture loop fails it.

**R6 — dead stub in `loadDryRunCommand` — FIXED.** The stub config is replaced by a `loadConfig` that throws, with a comment naming why it is unreachable (`ctx.sourceRoots` always carries `load_default_commands`, a required boolean). Smaller than what it replaces, and a future signature change surfaces instead of silently resolving commands against a fabricated policy.

**Validation:** `bun run validate` exit 0 — five generated-artifact checks, type-check across all packages, `lint --max-warnings 0`, `format:check`, `test:install`, full suite. `bun test src/fixture-runner.test.ts` 27 pass; `bun test src/commands/workflow.test.ts` 300 pass. `bun run cli workflow test` still 23 passed / 5 failed, the same pre-existing set from #2807 — unchanged by these fixes.

PR #2862, opened 2026-08-27
fix(workflows): workflow test no-fixtures error suggests only workflows that carry fixtures

### LINKED ISSUE #2850: fix(workflows): 'workflow test' no-fixtures error lists every workflow, not the ones that have fixtures
## Summary

- What broke: when `archon workflow test <target>` finds no fixtures, the error lists **every discovered workflow** as a candidate, not the ones that actually declare fixtures. `archon-review`, `archon-validate`, `archon-pr` and ~60 others appear in "Name a workflow with fixtures (…)" but have none, so following the message leads straight to the same error.
- When it started: since `workflow test` landed (#2772). Observed on dev `0113932f`.
- Severity: `cosmetic`

## Steps to Reproduce

1. `archon workflow test archon-review`
2. Read the error: `No fixtures found for 'archon-review'. Name a workflow with fixtures (archon-adversarial-dev, archon-architect, archon-assist, …, archon-review, …)`
3. Pick `archon-pr` from that list → same error

## Expected vs Actual

- **Expected**: the list contains only workflows that have at least one `fixtures/*.stubs.yaml` beside them (currently 8 in the sdlc pack).
- **Actual**: the list is every workflow in the discovery scope (~75 on this repo), most of which have no fixtures.

## Root cause

`packages/workflows/src/fixture-runner.ts`, `runFixtures`:

```ts
const byName = new Map(options.workflows.map(ws => [ws.workflow.name, ws]));  // ALL discovered workflows
…
const names = [...byName.keys()].sort().join(', ');   // → error message
```

`byName` is the full catalog (needed for lookup); the message reuses its keys instead of the fixture-carrying subset.

## Proposed fix

Build the suggestion list from the discovered fixtures:

```ts
const names = [...new Set(all.flatMap(f => f.workflowNames))].sort().join(', ');
```

## Environment

- Platform: CLI
- Database: SQLite
- Running in worktree? No
- OS: Windows 11 (platform-independent)

## Impact

- Affected workflows/commands: `archon workflow test`
- Reproduction rate: Always
- Workaround available? Yes — grep for `fixtures/` dirs manually
- Data loss risk? No

## Scope

- Module: `workflows:fixture-runner`

Wirasm:
## Problem and outcome

When `archon workflow test <target>` finds no fixtures for the target, the error's "Name a workflow with fixtures (…)" list included every discovered workflow, not just the ones that actually declare fixtures. Following the suggestion (for example `archon-pr` or `archon-review`) reproduced the same error, so the message dead-ended the user.

- **Outcome:** the suggestion list now contains only workflows that both exist in the discovery catalog and are targeted by at least one discovered fixture. When the filtered list is empty, the message explains why instead of rendering an empty list: either no workflow anywhere declares fixtures, or discovered fixtures target no workflow the catalog can load.
- **Invariant:** target lookup and all successful `workflow test` behavior are unchanged; only the no-fixtures error text changed.
- **Scope boundary:** nothing about fixture discovery, matching, or execution changed; other `workflow test` errors are untouched.
- **Root cause:** the error message reused the keys of `byName`, the full discovered-workflow catalog built for lookup, instead of the fixture-carrying subset. `byName` must stay complete for lookup; only the message needed the narrower source.

## Review guidance

- **Feedback requested:** correctness of the suggestion source and the empty-catalog wording.
- **Start here:** `packages/workflows/src/fixture-runner.ts:260` — the suggestion list now derives from `all.flatMap(f => f.workflowNames)` intersected with `byName`, instead of all `byName` keys.
- **Review order:** `fixture-runner.ts` change, then the four new tests in `fixture-runner.test.ts`.
- **Lower-attention areas:** the test file additions are mechanical setup — temp workflows, a name-only unloadable YAML, and empty discovery scopes — around regex assertions on the error message.

## Solution

`runFixtures` already collects every discovered fixture in `all`, and each fixture carries the workflow names it targets. Building the suggestion list from that set, filtered to names still present in the catalog, points the user only at workflows that can actually satisfy the command. The `byName.has(name)` filter goes one step beyond the issue's proposed fix: a fixture can reference a workflow that no longer exists or fell outside discovery, and suggesting it would recreate the same dead-end. When the filtered list is empty, the message distinguishes the two possible causes rather than printing an empty parenthetical: no fixtures discovered at all (a project with zero fixtures hitting `workflow test <anything>`, reachable before this change too) reports that no workflow declares fixtures, while discovered fixtures that target no loadable catalog workflow get an explicit divergence hint.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `workflow test <target>` with no matching fixtures listed every discovered workflow, including fixture-less ones. | The list contains only discovered workflows that at least one fixture targets. |
| **Failure behavior** | Following a suggestion could return the identical error. | Every suggested workflow name is in the catalog and targeted by a discovered fixture, so following it runs fixtures; an empty list gets an explicit explanation of which of the two causes applies. |

## Validation

- `bun test src/fixture-runner.test.ts` (packages/workflows) — 19 pass, including the new `suggests only workflows that carry fixtures when a target has none`, which failed before the fix (suggestions were `['bare-wf', 'test-wf']`) and passes after (`['test-wf']`), plus tests for the catalog-unloadable exclusion, the divergence hint, and the no-fixtures-at-all message.
- Existing tests `restricts to fixtures of the named workflow via target` and `filters by pack name when the target is not a workflow name` pass unchanged, proving target lookup is preserved.
- `bun run test` (packages/workflows full script) — all pass.
- `bun run type-check` (packages/workflows) — clean.
- `bun test src/commands/workflow.test.ts` (packages/cli) — 298 pass; the CLI mocks `fixture-runner`, so there is no message coupling.
- `bun run lint` (root wrapper) — clean.
- **Not verified:** Nothing material — the error string is fully covered by the package tests above.

## Links

- Closes #2850

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fixture discovery error messages with more accurate workflow suggestions.
  * Excluded unavailable or invalid workflows from suggested targets.
  * Added clearer guidance when fixtures are missing, undiscoverable, or do not match available workflows.

### DIFF
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -232,6 +232,27 @@ describe('runFixtures', () => {
     );
   });
 
+  it('suggests only workflows that carry fixtures when a target has none', async () => {
+    const { cwd } = writeTempProject({ body: passingBody });
+    const bareDir = join(cwd, '.archon', 'workflows', 'bare');
+    mkdirSync(bareDir, { recursive: true });
+    writeFileSync(
+      join(bareDir, 'bare-wf.yaml'),
+      'name: bare-wf\ndescription: test\nnodes:\n  - id: node-a\n    prompt: hi\n'
+    );
+    const err = await runFixtures({
+      workflows: [
+        ...workflowsOnDisk(cwd, ['test-wf']),
+        ...workflowsOnDisk(cwd, ['bare-wf'], 'bare'),
+      ],
+      cwd,
+      target: 'bare-wf',
+    }).catch((e: Error) => e);
+    expect(err).toBeInstanceOf(Error);
+    const suggested = /fixtures \(([^)]*)\)/.exec((err as Error).message)?.[1] ?? '';
+    expect(suggested.split(', ')).toEqual(['test-wf']);
+  });
+
   it('restricts to fixtures of the named workflow via target', async () => {
     const { cwd } = writeTempProject({
       workflowName: 'target-wf',
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -253,6 +253,57 @@ describe('runFixtures', () => {
     expect(suggested.split(', ')).toEqual(['test-wf']);
   });
 
+  it('excludes fixture-targeted workflows the catalog cannot load from suggestions', async () => {
+    const { cwd } = writeTempProject({ body: passingBody });
+    // A name-only YAML passes workflowNamesBeside but fails parseWorkflow, so the fixture
+    // beside it targets a workflow the command can never resolve (#2850).
+    const brokenDir = join(cwd, '.archon', 'workflows', 'broken');
+    mkdirSync(join(brokenDir, 'fixtures'), { recursive: true });
+    writeFileSync(join(brokenDir, 'broken-wf.yaml'), 'name: broken-wf\n');
+    writeFileSync(join(brokenDir, 'fixtures', 'orphan.stubs.yaml'), passingBody);
+    const err = await runFixtures({
+      workflows: workflowsOnDisk(cwd, ['test-wf']),
+      cwd,
+      target: 'nope',
+    }).catch((e: Error) => e);
+    expect(err).toBeInstanceOf(Error);
+    const suggested = /fixtures \(([^)]*)\)/.exec((err as Error).message)?.[1] ?? '';
+    expect(suggested.split(', ')).toEqual(['test-wf']);
+  });
+
+  it('reports the divergence hint when discovered fixtures target no catalog workflow', async () => {
+    const { cwd } = writeTempProject({ body: passingBody });
+    const err = await runFixtures({
+      workflows: [],
+      cwd,
+      target: 'nope',
+    }).catch((e: Error) => e);
+    expect(err).toBeInstanceOf(Error);
+    expect((err as Error).message).toMatch(
+      /Discovered fixtures target no workflow in the discovery catalog/
+    );
+    expect((err as Error).message).not.toMatch(/declares fixtures/);
+  });
+
+  it('says no workflow declares fixtures when discovery found no fixtures at all', async () => {
+    const cwd = makeTempProject();
+    cleanups.push(cwd);
+    const packDir = join(cwd, '.archon', 'workflows', 'pack');
+    mkdirSync(packDir, { recursive: true });
+    writeFileSync(
+      join(packDir, 'test-wf.yaml'),
+      'name: test-wf\ndescription: test\nnodes:\n  - id: node-a\n    prompt: hello\n'
+    );
+    const err = await runFixtures({
+      workflows: workflowsOnDisk(cwd, ['test-wf']),
+      cwd,
+      target: 'test-wf',
+    }).catch((e: Error) => e);
+    expect(err).toBeInstanceOf(Error);
+    expect((err as Error).message).toMatch(/No workflow in any discovery scope declares fixtures/);
+    expect((err as Error).message).not.toContain('()');
+  });
+
   it('restricts to fixtures of the named workflow via target', async () => {
     const { cwd } = writeTempProject({
       workflowName: 'target-wf',
--- a/packages/workflows/src/fixture-runner.ts
+++ b/packages/workflows/src/fixture-runner.ts
@@ -104,8 +104,10 @@ export function parseFixtureFile(text: string, path: string): ParsedFixtureFile 
 
 const FIXTURES_DIR = 'fixtures';
 const FIXTURE_SUFFIX = '.stubs.yaml';
-// Discovery walks user directories; the bound mirrors discovery's own depth cap so a
-// pathological tree cannot hang the command.
+// Discovery walks user directories. This is a hang-guard margin for a pathological
+// tree, not a mirror of discovery's cap (MAX_DISCOVERY_DEPTH is 1, and the catalog
+// reaches one packaged-scanner level deeper): fixtures below the catalog's reach are
+// discovered here but can never match a loadable workflow.
 const MAX_WALK_DEPTH = 12;
 
 async function exists(path: string): Promise<boolean> {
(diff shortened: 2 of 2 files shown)


Wirasm:
# Review report — PR #2862, run 21a72526-6d83-430c-8367-265be7840935, round 2

## Verdict

**Ready. Action: `none`.** All four round-1 blockers are fixed and test-protected at `45da4a63`, every enabled lens delivered its report, and no new blocking finding survived verification; the one remaining adjacent discovery is outside the accepted contract and does not affect readiness.

## Accepted contract

From PR #2862 (fixes #2850), carried forward from round 1: when `archon workflow test <target>` finds no fixtures for the target, the error's suggestion list must contain only workflows that **both** exist in the discovery catalog (`byName`) **and** are targeted by at least one discovered fixture. When no workflow anywhere declares fixtures, the message must say so explicitly instead of rendering an empty parenthetical list. Invariant: target lookup and all successful `workflow test` behavior unchanged — only the no-fixtures error text changed. Explicit boundaries: fixture discovery, matching, and execution unchanged; other `workflow test` errors untouched; `byName` stays complete for lookup. Round 1's ruling (R1) is part of the contract: the explicit no-fixtures sentence may only render when no discovered fixture declares fixtures for any workflow; a divergence hint is required when fixtures exist but target nothing in the catalog.

## Reviewed head SHA

`45da4a633c03d38dd4b366e6edf89018156468a6` (matches PR head and local HEAD; working tree clean; cursor for the next round).

## Findings

### Critical

None.

### Important

None open. All four round-1 Important findings were verified against the delta and are resolved — see Prior findings.

### Rejected findings

None — no lens raised a new Critical or Important finding this round, and the seams lens's cross-boundary observation was classified as an adjacent discovery rather than a finding (see Discoveries).

### Suggestions

None. Round-1 suggestion S1 was addressed by the delta (see Prior findings).

## Prior findings (light mode — full carried-forward table)

| --- | --- | --- | --- |
| R1 | Important | Empty-set hint asserted "no workflow declares fixtures" even when fixtures exist but target nothing in the catalog (false universal in divergence states) | **Fixed at `45da4a63`** |
| R2 | Important | `MAX_WALK_DEPTH` comment falsely claimed to mirror discovery's depth cap (`MAX_DISCOVERY_DEPTH = 1`), denying real drift | **Fixed at `45da4a63`** |
| R3 | Important | The explicit no-fixtures sentence shipped untested; reverting it would pass the suite | **Fixed at `45da4a63`** |
| R4 | Important | The load-bearing `byName.has(name)` catalog filter shipped untested | **Fixed at `45da4a63`** |
| S1 | Suggestion | The suggestion-source comment omitted why the `byName.has` filter exists (kept suggestions runnable) | **Addressed at `45da4a63`** |

Verification performed by this synthesis, not taken from the lenses alone:

- **R1** — the empty-suggestions branch is now a ternary (`packages/workflows/src/fixture-runner.ts:271-275`): "No workflow in any discovery scope declares fixtures." renders only when `all.length === 0`; otherwise "Discovered fixtures target no workflow in the discovery catalog.". `names` is exactly the union of discovered fixtures' `workflowNames` intersected with `byName` (`fixture-runner.ts:265-267`), so the branch conditions are exhaustive and truthful in every reachable state. Protected by the new divergence test (`fixture-runner.test.ts:271-288`), which asserts the divergence text and the absence of "declares fixtures".
- **R2** — `fixture-runner.ts:107-110` now states the constant is a hang-guard, not a mirror. Factual claims verified: `MAX_DISCOVERY_DEPTH = 1` (`packages/workflows/src/command-validation.ts:23`, confirmed by grep), and the packaged scanner reaches one directory level deeper than the flat loader, so the drift the old comment denied is real and now documented.
- **R3** — test "says no workflow declares fixtures when discovery found no fixtures at all" (`fixture-runner.test.ts:290-307`): no `fixtures/` directory anywhere, asserts the sentence renders and no empty `()` appears. Deleting the sentence would fail the `toMatch` assertion — non-vacuous.
- **R4** — test "excludes fixture-targeted workflows the catalog cannot load from suggestions" (`fixture-runner.test.ts:256-277`): a name-only YAML (`broken-wf`) passes `workflowNamesBeside` but fails `parseWorkflow`, so it is absent from the catalog; without the `byName.has(name)` filter the `toEqual(['test-wf'])` assertion fails — non-vacuous.
- **S1** — `fixture-runner.ts:261-264` now explains both halves: the full catalog would suggest unsatisfiable workflows, and a fixture beside an unloadable workflow would too.
- Full suite for the touched contract re-run during synthesis: `bun test src/fixture-runner.test.ts` — **19 pass, 0 fail**. The delta's production change is confined to the no-fixtures error branch (`fixture-runner.ts:263-276`); target lookup and all successful-path behavior are unchanged, matching the invariant.

## Discoveries

1 accepted adjacent discovery: **"workflow test drops discovery's errors half, so the divergence hint cannot say why a workflow is missing from the catalog"** (sources: seams + comments, merged). `workflowTestCommand` (`packages/cli/src/commands/workflow.ts:1022`) discards `WorkflowLoadResult.errors`, which the sibling `list` command (`:1047`) surfaces — so the divergence hint cannot distinguish "no such workflow" from "the workflow failed to load and the reason was thrown away". Surfacing it exceeds this PR's accepted contract, so it is recorded as adjacent and **does not affect readiness**. Evidence verified by this synthesis at the head SHA. See [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/21a72526-6d83-430c-8367-265be7840935/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/21a72526-6d83-430c-8367-265be7840935/discoveries.md). The comments lens's raw record was rejected as an open discovery: it is the same causal finding as round-1 R1, reproduced on the pre-delta head, and the delta's ternary now renders a truthful message in both of its reproduction cases; its dropped-errors half is preserved in the accepted adjacent record.

## Lens roster

| Lens | Status | Result |
| --- | --- | --- |
| code | ran | no new findings; verified R1-R4 fixes, blast radius, lint, full suite green |
| seams | ran | no new findings; 1 adjacent discovery (dropped `errors` half), class completed |
| simplify | ran | no new findings; ternary is the minimal shape for three message states |
| tests | ran (full, per round-1 directive) | no new findings; both new delta branches have distinct failing assertions |
| comments | ran | no new findings; all delta comment claims verified accurate; 1 raw discovery rejected as duplicate of fixed R1 |
| errors | disabled this round | — |
| types | disabled this round | — |
| docs | disabled this round | — |

No enabled lens failed to report. No directives for a next round: the review is ready, and a future delta round may use this report's cursor SHA `45da4a633c03d38dd4b366e6edf89018156468a6`.

## Verification

- Head SHA `45da4a633c03d38dd4b366e6edf89018156468a6` confirmed against the PR head and local HEAD; delta is exactly one commit, 2 files, +61/−5, matching scope.md.
- Every carried-forward verdict re-verified by reading the delta diff and cited files; the test suite re-run (19/19); the adjacent discovery's `workflow.ts:1022`/`:1047` evidence read directly.
- Canonical PR comment posted (edited in place, round 1's marked comment) and read back at: https://github.com/coleam00/Archon/pull/2862#issuecomment-5441261215

PR #2865, opened 2026-08-27
Script node failure diagnostics include stdout

### LINKED ISSUE #2825: fix(workflows): script-node failure diagnostics drop stdout — '[exit 1]: no diagnostic output' hides the actual failure
## Problem

When a `script:` node fails, the node_error event and workflow error carry only the subprocess **stderr** (or "no diagnostic output" when stderr is empty). Script stdout — where well-behaved scripts print their progress and failure context — is dropped from the failure diagnostics entirely.

Live occurrence: a stabilize-fix child's verify script failed with `[exit 1]: no diagnostic output`. The script had in fact printed the decisive context to stdout (`targeted test failed on repetition N/5: <cmd>` + test output tail), but because it followed stderr-for-errors convention, the operator could not see WHY without re-running manually.

## Desired outcome

Failed script/bash node diagnostics should include a tail of **stdout** as well as stderr — e.g. prefer stderr, fall back to stdout tail, or include both truncated. The existing truncation cap applies.

Alternative accepted shape: document loudly that scripts must write failure diagnostics to stderr only, and provide a convention/helper — but capturing stdout on failure is the behavior most languages' test runners make useful with zero author effort.

## Evidence

Run `b6dfe2d5` (stabilize-fix child): fix node produced an excellent structural fix; verify failed; the batch report and error events said only "no diagnostic output"; stdout containing the repetition count and failing command was unrecoverable.

Wirasm:
## Problem and outcome

When a `script:` or bash node fails, the node error and log carried only the subprocess **stderr** (or "no diagnostic output" when stderr was empty). Well-behaved scripts print progress and failure context to stdout, so the decisive evidence — a repetition count, the failing command, a test output tail — was silently dropped and an operator had to re-run the script manually to learn why it failed.

- **Outcome:** Failed script/bash node diagnostics now include tails of both streams under the existing 2 KB budget. With stderr empty, the stdout tail becomes the diagnostic; when both streams are populated each is labelled and stderr keeps budget priority.
- **Invariant:** The diagnostic never exceeds the `SUBPROCESS_ERROR_MAX_CHARS` (2000) budget, stream labels only appear when both streams are populated, and the script body (embedded in Node's `Command failed:` prefix) is never echoed back to users.
- **Scope boundary:** The dry-run simulator (`dry-run.ts`) has its own stderr-only reporting path and was intentionally left untouched; see Review guidance.
- **Root cause:** `formatSubprocessFailure` built its diagnostic from `err.stderr` alone and never read `err.stdout`, even though the subprocess rejection already carries a redacted stdout.

## Review guidance

- **Feedback requested:** Correctness of the joint budget arithmetic and whether the label shapes (`[stderr]` / `[stdout]`) read well in batch reports and conversation messages.
- **Start here:** `packages/workflows/src/executor-shared.ts:246` — the diagnostic construction is the load-bearing change; both `dag-executor.ts` call sites (bash nodes, script nodes) consume this one formatter, so node_error events, batch reports, and the `dag_node_failed` log line all pick up the behavior here.
- **Review order:** `executor-shared.ts` (formatter), `executor-shared.test.ts` (stream-tail behavior and budget), `dag-executor.test.ts` (end-to-end bash-node assertion), `script-nodes.md` (capture-section docs).
- **Lower-attention areas:** The docs capture section and the bash-node test rework. The test's stdout marker had been a proxy for the leaked `Command failed:` prefix; since stdout is now legitimately surfaced, the marker moved into the script body only and the stdout inclusion got its own assertion.
- **Known risk or uncertainty:** `dry-run.ts:677` reports a failing script node as `err.stderr?.trim() || err.message`, which still drops stdout and can leak the script body into the dry-run report when stderr is empty. It sits outside the pinned sanitized formatter path and is not fixed here.

## Solution

`formatSubprocessFailure` now caps both streams jointly: stderr keeps budget priority (up to 1000 chars when stdout is also populated, up to the full 2000 otherwise) and stdout gets the remainder. The budget accounts for the label overhead so the joined diagnostic never truncates away a stream label. With stderr empty, the stdout tail becomes the diagnostic. Log fields now carry `stdoutTail` alongside `stderrTail`, still a controlled, tail-truncated subset — never the full error object — so Pino's serializer cannot emit copies of the script body. The fallback chain after the streams and the prefix stripping that keeps the script body out of user-facing output are unchanged. The subprocess rejection already carries credential-scrubbed stdout (`redactSubprocessError`), so no new redaction is needed before the text reaches the formatter.

One test rework was required: an existing executor-shared test asserted a `…[truncated]` suffix. Tails are now pre-capped, so a single-stream diagnostic never exceeds the budget and the suffix is unreachable for stream content; the guard remains for the uncapped `bodyAfterPrefix` fallback path.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Failed script/bash node error carried stderr only, or "no diagnostic output" when stderr was empty. | Error carries labelled tails of both streams (`[stderr]` / `[stdout]`) sharing the 2 KB budget; with stderr empty, the stdout tail is the diagnostic. |
| **Failure behavior** | Live failure showed `[exit 1]: no diagnostic output` while stdout held the repetition count and failing command — unrecoverable without a manual re-run. | The same failure surfaces the stdout context (`targeted test failed on repetition N/5: <cmd>`) in the node error, batch report, and log fields. |

## Validation

- `packages/workflows`: `bun run test` — 0 failures across all suite chunks; proves the both-streams diagnostic, the stderr-empty fallback, and the joint cap bounds.
- `packages/workflows`: `bun run type-check` — clean.
- Root `bun run lint` and `bun run format:check` — clean.
- Root `bun run type-check` — clean.
- **Not verified:** Root `bun run validate` was not run end to end; the changed paths are covered by the package test script, both type-checks, lint, and format:check, and no other packages are touched by this diff.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Additive: new `stdoutTail` log field and longer diagnostics. Existing stderr-only diagnostics keep their shape when stdout is empty. | `executor-shared.ts` fallback chain unchanged |
| Security / permissions / data | Stdout text entering diagnostics is already redacted by `redactSubprocessError`; log fields remain a tail-truncated subset, never the full error object. | `dag-executor.ts` redaction at subprocess rejection |
| Documentation / communication | Script-nodes guide capture section updated to describe the both-streams diagnostic. | `packages/docs-web/src/content/docs/guides/script-nodes.md` |

## Links

- Closes #2825

## Summary by CodeRabbit

* **Bug Fixes**
  * Script execution errors now include relevant output from both standard error and standard output.
  * Diagnostics are limited to approximately 2,000 characters, prioritizing standard error for clearer failure messages.
  * Output sections are labeled when both streams contain information.

* **Documentation**
  * Updated script-node guidance to describe the improved failure diagnostics.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/script-nodes.md
+++ b/packages/docs-web/src/content/docs/guides/script-nodes.md
@@ -75,9 +75,12 @@ The file `.archon/scripts/fetch-github-pages.ts` is loaded and executed with
 4. **Capture.** `stdout` (with the trailing newline stripped) becomes
    `$nodeId.output`. On a successful run, `stderr` is logged as a warning and
    posted to the conversation but does **not** fail the node. A non-zero exit
-   code fails the node; on failure, `stderr` is the diagnostic surfaced in the
-   error message (`Script node 'X' failed [exit N]: <stderr>`) — the script
-   body is never echoed back to users.
+   code fails the node; on failure, tails of both streams are surfaced in the
+   error message (`Script node 'X' failed [exit N]: [stderr] ... [stdout] ...`),
+   sharing a ~2 KB diagnostic budget — stderr keeps priority, stdout gets the
+   remainder, and a label prefixes each stream only when both are populated.
+   With stderr empty, the stdout tail becomes the diagnostic. The script body
+   is never echoed back to users.
 
 ## YAML Schema
 
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -2586,14 +2586,16 @@ describe('executeDagWorkflow -- bash nodes', () => {
       user_message: 'test',
     });
 
-    // Marker is echoed to stdout only (so it lands in the command line embedded
-    // in err.message but never in stderr). If it shows up in errorMsg the
-    // prefix line was not stripped.
+    // The marker appears only in the script body (never echoed to stdout or
+    // stderr). If it shows up in errorMsg the "Command failed: bash -c <body>"
+    // prefix line was not stripped. The stdout echo is separately asserted so a
+    // regression that drops stdout from the diagnostics is caught.
     const bashNode: ExecNode = {
       id: 'fail-bash-1389',
       kind: 'exec',
       runtime: 'sh',
-      script: 'echo UNIQUE_CMDLINE_MARKER_1389; echo "diagnostic from stderr" >&2; exit 1',
+      script:
+        'UNIQUE_CMDLINE_MARKER_1389=; echo "stdout context before failure"; echo "diagnostic from stderr" >&2; exit 1',
     };
 
     await executeDagWorkflow(
@@ -2626,6 +2628,9 @@ describe('executeDagWorkflow -- bash nodes', () => {
     expect(errorMsg).not.toContain('Command failed:');
     expect(errorMsg).not.toContain('UNIQUE_CMDLINE_MARKER_1389');
     expect(errorMsg).toContain('diagnostic from stderr');
+    expect(errorMsg).toContain('stdout context before failure');
+    expect(errorMsg).toContain('[stderr]');
+    expect(errorMsg).toContain('[stdout]');
   });
 
   it('variable substitution works in bash scripts', async () => {
--- a/packages/workflows/src/executor-shared.test.ts
+++ b/packages/workflows/src/executor-shared.test.ts
@@ -767,16 +767,15 @@ describe('formatSubprocessFailure', () => {
     expect(userMessage).toContain('[exit 2]');
   });
 
-  it('truncates diagnostics larger than 2 KB from the tail', () => {
+  it('keeps the tail of diagnostics larger than 2 KB and bounds the output', () => {
     const big = 'x'.repeat(5000) + '\nactual error at end';
     const { userMessage } = formatSubprocessFailure(
       { message: 'Command failed: cmd\n', stderr: big, code: 1 },
       "Script node 'n1'"
     );
     expect(userMessage).toContain('actual error at end');
-    expect(userMessage).toContain('[truncated]');
-    // Tight bound: ~2 KB diagnostic + label prefix + truncation suffix should fit
-    // well under 2.1 KB. Bumping SUBPROCESS_ERROR_MAX_CHARS would trip this.
+    // Tight bound: ~2 KB diagnostic + label prefix should fit well under 2.1 KB.
+    // Bumping SUBPROCESS_ERROR_MAX_CHARS would trip this.
     expect(userMessage.length).toBeLessThan(2100);
   });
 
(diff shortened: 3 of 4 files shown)


Wirasm:
# Review report — PR #2865 (script node failure diagnostics include stdout)

## Verdict

**Not ready. Action: `correct`.** Two Important findings block readiness: the budget allocation contradicts the accepted contract's "stderr keeps budget priority" semantics and regresses long-stderr diagnostics, and the stderr-empty path the fix targets has no test bounding the stdout tail. Both are correctable inside the accepted contract.

## Accepted contract

Source: PR #2865 body (requested scope: "Run on issue #2825", which the PR closes).

- **Required outcome:** Failed `script:`/bash node diagnostics must include stdout in addition to stderr. With stderr empty, the stdout tail becomes the diagnostic; when both streams are populated, each is labelled (`[stderr]` / `[stdout]`) and stderr keeps budget priority. The existing `SUBPROCESS_ERROR_MAX_CHARS` (2000) budget applies jointly, and the script body (embedded in Node's `Command failed:` prefix) must never be echoed back to users.
- **Explicit boundaries / non-goals:**
  - The dry-run simulator (`dry-run.ts:677`) keeps its stderr-only reporting path; intentionally untouched, out of scope.
  - No new redaction: the subprocess rejection already carries credential-scrubbed stdout (`redactSubprocessError` in `dag-executor.ts`).
  - The fallback chain after the streams and the prefix stripping are unchanged.

## Reviewed head SHA

`0f209ea76560f3050a6847ab6c916a83aa488874` (cursor for the next round)

## Findings

### Important

#### R1 — When both streams are populated, stderr is hard-capped at 1000 chars, contradicting "stderr keeps budget priority" and discarding stderr diagnostics while ~950 chars of the joint budget go unused

**Contributing lenses:** code, comments (merged: both lenses found the same half-cap mechanism; their corrections disagree, see below).

**Claim:** `executor-shared.ts:254-255` caps stderr at `halfCap = 1000` whenever stdout is non-empty (even 1 char), regardless of how much of the joint 2000-char budget stderr actually needs:

```ts
const halfCap = Math.floor(SUBPROCESS_ERROR_MAX_CHARS / 2);
const stderrTail = streamTail(stderr, bothPresent ? halfCap : SUBPROCESS_ERROR_MAX_CHARS);
```

**Evidence (verified by reviewer against head):** ran a falsifying input through the head implementation — stderr = root-cause line + 1600 filler chars, stdout = a short progress line, exit 1. Result: `userMessage.length = 1066`, `stderrTail.length = 1000`, and the root-cause line `TypeError: cannot read properties of undefined` is **absent** from the user-facing message, while ~930 chars of the 2000-char budget are unused. This is also a regression against prior behavior, which kept up to 2000 chars of stderr when stdout was present but ignored.

**Contract contradiction:** the accepted contract says "stderr keeps budget priority"; the docstring (`executor-shared.ts:224`), inline comment (`:247`), and docs page (`packages/docs-web/src/content/docs/guides/script-nodes.md:80` — "stderr keeps priority, stdout gets the remainder") all state that semantics. The code implements a fixed 50/50 split: stderr capped at half regardless of demand, stdout guaranteed a share. The comment "stderr keeps budget priority; the leftover goes to stdout" describes the opposite of what the code does. The new joint-budget test (`executor-shared.test.ts:860`, `stderrTail.length <= 1000`) pins the 50/50 behavior, so the tests and all three prose sites tell two different stories about the same contract.

**Discovery method / causal class:** once the half-cap mechanism was proved, the complete finite class of budget-allocation members in `formatSubprocessFailure` (lines 243-265) was examined. Invariant: the joined diagnostic must stay ≤ 2000 chars with stderr keeping priority over the budget.

- Affected member: stderr allocation when both streams are present (R1).
- Examined-clean members: stdout allocation (correctly gets the remainder); stderr-empty path (correctly caps stdout at 2000 − overhead); label-overhead accounting (correct, diagnostic ≤ 2000, final re-truncation never strips a label); final `…[truncated]` guard (safe, only fires via the `bodyAfterPrefix` fallback).

**Smallest correction:** allocate stderr first up to its demand within the budget, then hand stdout the remainder — e.g. compute the label overhead, cap stderr at `SUBPROCESS_ERROR_MAX_CHARS − labelsOverhead − min(stdout demand, stdout share)`, or compute both tails and redistribute unused stdout budget back to stderr. Under that allocation the falsifying input keeps the full stderr (≈1520 < 2000). Whichever allocation is chosen, the docstring, inline comment, docs line, and pinning test must state the same semantics. The docs-review lens proposed the opposite correction (change the prose to say 50/50); that is rejected because the accepted contract and the issue's motivation (stderr diagnostics with stdout context) define priority as the intended behavior, and the 50/50 split demonstrably drops the root cause while wasting budget.

#### R2 — No test bounds the stdout tail when stderr is empty, so the documented 2 KB budget is untested on the primary path of the fix

**Contributing lens:** tests.

**Claim:** the stderr-empty path caps stdout at `SUBPROCESS_ERROR_MAX_CHARS − labelsOverhead − 0` (`executor-shared.ts:256-259`), but no test covers a large stdout with empty stderr. The stderr-empty test (`executor-shared.test.ts:814-826`) uses ~60 chars of stdout and passes whether or not the cap exists; the joint-cap test (`:844-863`) populates both streams and exercises a different branch; the e2e test (`dag-executor.test.ts:2586-2632`) has both streams populated. A refactor that drops the capping argument in the stderr-empty branch passes the whole suite while delivering up to Node's default 1 MB `maxBuffer` (`packages/git/src/exec.ts:68-79`) of stdout into the user message, the `node_failed` event, and the log fields — precisely the stdout-heavy failure class of issue #2825.

**Evidence (verified by reviewer against head):** read all seven `formatSubprocessFailure` tests; confirmed none bounds stdout with stderr empty.

**Smallest correction:** add one unit test: stderr empty, stdout of ~5000 chars ending in a distinctive error line; assert `userMessage.length < 2100`, `userMessage` contains the end-of-stdout error line (tail, not head), and `logFields.stdoutTail.length <= 2000`. This passes against the current implementation and fails under the described regression.

### Rejected findings

- **Docs-review proposal to change the prose to describe the 50/50 split** — rejected as the correction; R1 is a code defect against the accepted contract, so the code moves, not the prose.
- **Errors lens's `…[truncated]` marker removal** — below the Important bar (the node still fails loudly with exit code and tail; the marker path remains live for the `bodyAfterPrefix` fallback; the test assertion was removed intentionally in the same PR). Preserved as adjacent discovery 2.

### Suggestions

None beyond the two Important findings' smallest corrections. The simplify lens found nothing to remove; cosmetic notes (the `else if` diagnostic branch could be `stderrTail ?? stdoutTail ?? ''`) were rejected as non-structural.

## Prior findings

None — first round, full mode.

## Discoveries

Three adjacent discoveries were accepted and consolidated (one merged from two duplicate source records); no scope conflicts:

1. Rejection-shape casts at the two `formatSubprocessFailure` call sites omit stdout.
2. `…[truncated]` marker no longer appears for oversized stderr/stdout in script-node diagnostics.
3. Dry-run simulator reports failing script-node stderr only, with the raw message as unsanitized fallback.

Records: [discoveries.json](../discoveries.json) · [discoveries.md](../discoveries.md). Adjacent discoveries do not affect readiness. Discovery 3 is the explicit non-goal named in the accepted contract and is preserved, not reopened.

## Review coverage

Full mode, first round. Seven lenses ran and reported: code, seams, tests, errors, comments, docs, simplify (types disabled per scope inputs — `types=false`). No enabled lens failed to report. The tests and simplify lenses found the change otherwise well covered; the errors and docs lenses found no defects in their domains. One continuation-free note: the code reviewer's falsifying run and the reviewers' test-run results (executor-shared: 116 pass; dag-executor bash suite: 50 pass; `tsc --noEmit` clean) were re-verified against head `0f209ea` by this reviewer, including an independent reproduction of R1. The lenses' downstream-consumer greps (web bridge, CLI, log fields, retry path, `until_bash` rejections, redaction seam) were spot-checked against the diff and the cited lines.

PR #2866, opened 2026-08-27
test(core): SQLite upgrades are verified against every released schema vintage

### LINKED ISSUE #2553: test(core): SQLite has no upgrade check — the default database is verified only against hand-written old shapes
## Summary

Since #2552, Postgres upgrades are verified against **every distinct schema Archon has ever released** — 8 vintages, applied and diffed against a fresh install on every PR. SQLite, which is the **default** database, has no equivalent. Its only protection is a structural guard plus two regression tests whose old-database shapes are hand-written from reading `migrateColumns()`.

That gap is not theoretical. While building the Postgres check, the same bug class turned out to be **live on SQLite**: opening a database created before `hidden` / `deleted_at` / `parent_conversation_id` existed threw `no such column: parent_conversation_id` out of the `SqliteAdapter` constructor — permanently, because `createSchema()` runs before `migrateColumns()`, so the migration that would add the column never got the chance. Fixed in #2552.

It was found only because someone went looking for the Postgres analogue by hand. Nothing in CI would have caught it, and nothing would catch the next one on a column nobody thought to check.

## Why the existing coverage isn't enough

`packages/core/src/db/migration-statement-order.test.ts` now asserts, structurally, that no index in `createSchema()` names a column `migrateColumns()` adds. That closes the *known* anti-pattern and would have caught this bug. It is still a static, syntactic check over adapter source text — it cannot see:

- a statement type it doesn't model (it covers `CREATE INDEX`; trigger bodies and `ALTER COLUMN … SET DEFAULT` reference columns too),
- an ordering problem *within* `migrateColumns()` — an index placed before the `ALTER TABLE` in the same block,
- anything about whether an upgraded database ends up **shaped like** a fresh one, which is the second half of the failure mode (the crash goes away and an index quietly goes with it).

`packages/core/src/db/adapters/sqlite.test.ts` has two real old-database tests (the pre-0.4.0 `user_id` case, and the one #2552 added). Both seed a `CREATE TABLE` shape typed out by hand, which means coverage is limited to the shapes someone already suspected — exactly the blind spot that let both bugs ship.

## What to build

The Postgres equivalent works because every release's schema is recoverable from git as a file: `git show <tag>:migrations/000_combined.sql`. SQLite's schema is *code*, so the same trick doesn't apply — an old `sqlite.ts` can't be meaningfully executed against today's imports.

The pragmatic route is a recorded fixture per distinct released schema:

1. One-time generator: for each release tag, check it out, run its own `SqliteAdapter` against an empty file, and record the resulting `sqlite_master` dump (`SELECT type, name, sql FROM sqlite_master ORDER BY type, name`) as a checked-in fixture. Deduplicate identical dumps the way `scripts/check-schema-upgrades.ts` deduplicates schema blobs — most releases won't differ.
2. Test: for each fixture, materialize a database from the dump, construct the **current** `SqliteAdapter`, and assert (a) it does not throw, and (b) the resulting `sqlite_master` matches what the current adapter produces on an empty file — same tables, columns, indexes, triggers.

Part (b) is the piece the current tests lack entirely, and it's what catches "survived the upgrade but lost an index".

Notes for whoever picks this up:

- Fixture generation is the real cost. It needs each tagged version run once; `bun` + a throwaway worktree per tag should do it, but old versions may need their own `bun install`.
- Only tags whose `sqlite.ts` schema actually differs need a fixture. `scripts/check-schema-upgrades.ts` (`defaultBaselines()`) is the precedent for deduplicating by content rather than sampling by recency — the sample it replaced was missing the schema fifteen releases shipped.
- This runs in-process with no service container, so unlike `check:schema-upgrades` it belongs in `bun run test` / `bun run validate` rather than a separate CI job.
- Keep the fixtures small: schema only, no rows. The bug class is DDL-shaped.

## Prior art in the tree

- `scripts/check-schema-upgrades.ts` — the Postgres version, including the fresh-vs-upgraded catalog diff and the single sanctioned-divergence exception with a staleness check.
- `.github/workflows/test.yml` → `schema-upgrade` job.
- `AGENTS.md` → *Schema changes are additive-only*, which documents the placement rule this would enforce empirically on the second dialect.

## Related

- #2552 — Postgres upgrade check, the SQLite crash fix, and the structural guard
- #2508, #2443 — instances of the same class on Postgres
- #2418, #2501 — earlier SQLite instances, each fixed individually without a general check

Wirasm:
## Problem and outcome

SQLite users upgrading from an older released version could hit a #2552-class failure: the current `SqliteAdapter` constructor throws on an old database and leaves it permanently unopenable. Postgres upgrade behavior is already pinned by `check:schema-upgrades` against released schemas; SQLite had no equivalent.

- **Outcome:** Every distinct SQLite schema shipped in a release tag is checked into CI as a fixture. A test replays each vintage into a database, opens it with the current adapter, and fails unless it converges on the fresh-install schema.
- **Invariant:** A fresh install and an upgrade from any released vintage must expose the same tables, columns, indexes, and triggers (modulo the one sanctioned SQLite divergence class).
- **Scope boundary:** No runtime code changed — `SqliteAdapter`, `createSchema()`, and `migrateColumns()` are untouched. Postgres upgrade checking is unchanged.

## Review guidance

- **Feedback requested:** Correctness of the fixture-extraction premise and the exception design.
- **Start here:** `scripts/generate-sqlite-vintages.ts` — the extractor's validity rests on one premise: `createSchema()`'s body is a single uninterpolated SQL literal. It fails loudly (missing-signature check, `${…}` detection) if that premise ever breaks, rather than silently emitting wrong fixtures.
- **Review order:** `scripts/generate-sqlite-vintages.ts` → `packages/core/src/db/adapters/sqlite-vintages.test.ts` → the CI step in `.github/workflows/test.yml`. Fixtures themselves are mechanical output.
- **Lower-attention areas:** The 10 fixture files under `packages/core/src/db/fixtures/sqlite-vintages/`. They are byte-exact extractions from release tags (plain byte comparison in `--check` mode), regenerated by `bun run generate:sqlite-vintages`.
- **Known risk or uncertainty:** The teeth check found that removing a `CREATE INDEX` from `migrateColumns()` does not fail (index exists only at HEAD and in no fixture, so fresh and upgraded lose it together); removing the `created_by_user_id` ALTER block did fail v0.2.13 as expected. Coverage therefore focuses on missing columns/objects, not every hypothetical `migrateColumns()` regression.

## Solution

A generator script extracts `createSchema()`'s DDL from every release tag via `git show <tag>:packages/core/src/db/adapters/sqlite.ts`, dedupes by exact SQL content, and writes one fixture per distinct vintage. The check-in part is deliberate: the unit test must run in the shallow CI checkout where tags do not exist, so fixtures are files and `--check` is a plain byte comparison that fails when a new release tag carries a new distinct schema without its fixture.

The test builds a fresh reference catalog from the current `SqliteAdapter`, then for each fixture replays the vintage DDL into an empty database, opens it with the current adapter (constructor completing is the "upgrade does not throw" assertion), and compares catalogs. Tables are compared by name and column `name:TYPE` — NOT NULL/defaults are deliberately excluded because SQLite cannot ALTER those onto existing columns, which is exactly the sanctioned divergence class documented in `createSchema()`. Indexes and triggers are compared on full whitespace-normalized SQL, so a lost index or trigger still fails.

One narrow named exception, `EXPECTED_EXTRA_COLUMNS = { remote_agent_codebases.allow_env_keys }`: v0.3.1 shipped this column and it was later dropped from `sqlite.ts` (#2318). SQLite cannot drop it and schema evolution is additive-only, so existing v0.3.1 databases keep it — sanctioned residue, not a migration bug; the Postgres-only parity exception in `sqlite.test.ts` documents the same column. The exception is one-directional (upgraded-extra only), not a general allowlist, and carries a self-expiring guard that fails when it stops describing reality, mirroring `EXPECTED_MISSING_CONSTRAINT` in the Postgres script.

The CI step runs in the existing `schema-upgrade` job, which already has `fetch-depth: 0`. It is not added to `bun run validate` — that runs in the shallow-checkout quality job where tags don't exist. The unit test itself makes no git calls and is registered in the `@archon/core` test chain next to `sqlite.test.ts`.

## Validation

- `bun test src/db/adapters/sqlite-vintages.test.ts` (from `packages/core`) — pass, 10 vintages converge; also verified green after commit, and runs in the `@archon/core` test chain.
- `bun run check:sqlite-vintages` — exit 0 on the clean set; hand-tampering a fixture made it exit 2 with the regeneration hint, then regenerated and re-verified exit 0.
- Teeth check — removing the `created_by_user_id` ALTER block from `migrateColumns()` failed v0.2.13 with `missing after upgrade [column]`; file restored and suite re-verified green.
- `bun run type-check`, `bun run lint`, `bun run test` (root) — pass.
- **Not verified:** The new CI step and the unit test on windows-latest have not been observed in CI yet (developed on ubuntu/macOS). The test uses only `bun:sqlite` + `fs`, both already exercised on Windows by `sqlite.test.ts`.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Additive-only: new fixtures, new script, new test. Existing behavior and data untouched. | Diff contains no changes to `sqlite.ts` or existing tests |
| Rollout / rollback | Revert the two commits to drop the check entirely; no data or release coupling. | Commits `2b8d0c13`, `f2a14cf9` |
| Observability | A future release shipping a new distinct SQLite schema without its fixture fails the `schema-upgrade` job; a regression breaking upgrades from a released vintage fails the `@archon/core` test chain with the vintage named in the failure. | `.github/workflows/test.yml:122`, `sqlite-vintages.test.ts` failure messages prefixed `[vintage]` |

## Links

- Closes #2553
- Related #2318

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQLite database upgrade compatibility across previously released schema versions.
  - Added validation to detect schema drift and ensure new schema versions include matching migration fixtures.

- **Tests**
  - Added automated coverage for upgrading databases from historical SQLite versions.
  - Added checks comparing upgraded databases with fresh installations to detect structural differences.

- **Chores**
  - Added tooling to generate and validate SQLite schema fixtures from release history.


Wirasm:
# Review report — run 0ebfdf52, round 3 (PR #2866, SQLite schema vintages)

## Verdict

**ready: true, action: none.** All three round-1 blockers are verified fixed by execution against the current head, the correction delta introduces no defects, and the prior round's empty-delta gap no longer exists.

## Accepted contract

Build the SQLite equivalent of the Postgres upgrade check: vintage fixtures extracted textually from `createSchema()`'s SQL literal per release tag via `git show`, deduped by exact SQL content (oldest tag wins), checked in under `packages/core/src/db/fixtures/sqlite-vintages/`; a unit test replays each vintage, opens it with the current `SqliteAdapter`, and asserts convergence on the fresh-install schema using semantic set comparison. The extractor fails loudly when its premise breaks; the generator uses Node builtins and `spawnSync` argument arrays with the `--check` exit contract 0/1/2; the check runs in the `schema-upgrade` CI job; the test makes no git calls. The work must have teeth.

Explicit boundaries (plan.md "Out of scope"): no changes to `SqliteAdapter`/`createSchema()`/`migrateColumns()` or the Postgres check; no rows in fixtures; no `check:sqlite-vintages` in `bun run validate` or the quality job; no coverage of pre-release dev-history shapes; no general exception/allowlist mechanism beyond the narrow v0.3.1 `allow_env_keys` case (verified required in round 1; non-blocking per the plan's own risk prescription).

## Reviewed head SHA

`f2a14cf9fbdd70439fcbbe4da149b361417da6f2` (PR #2866 head, OPEN, draft; verified via `gh pr view 2866` and `git rev-parse HEAD` in the local checkout, working tree clean — the reviewed diff is the PR diff exactly). Merge base with `dev`: `1cbaf258b13ab9ef5985abc839ac5601ed2f5744`.

## Findings

No open findings. All three prior Important findings are closed with execution evidence below; the delta itself was reviewed line by line and introduces no new defects.

Delta review notes (why no new findings): the test's `tempDbPaths` array is the sole registration point for temp databases (`tempDbPath()` is the only path-creating function, and both the fresh reference db and every vintage db route through it), so the `afterEach` sweep covers everything the test creates, including on failure (bun runs `afterEach` regardless). The generator's new guards are ordered correctly: `git tag` failure throws with stderr before any file operation, the empty-list guard runs before write mode's `rmSync` loop can see a fixture set it would interpret as "all stale", and the `seenAdapter` throw fires only for a rev-parse miss after a tag carrying the adapter — the legitimate all-missing prefix case was verified to still pass through.

### Rejected findings

None this round.

### Suggestions

1. **Tag ordering relies on `--sort=creatordate`.** `scripts/generate-sqlite-vintages.ts:99` assumes creatordate ordering is chronological. A pathologically backdated tag could scramble the order and trigger the round's new `seenAdapter` loud error for a tag that genuinely predates the adapter. This is fail-safe (a false positive, never a silent skip), matches the "fails loudly when its premise breaks" contract, and would need contrived tag metadata to occur — not blocking. `--sort=v:refname` on the `vX.Y.Z` tags would remove even the false positive.

## Prior findings

Carried forward from the round-1 report at cursor `2b8d0c13ef35e6f3747f844cb39e5caa3577ef18`; each verified against head `f2a14cf9` by running the smallest falsifying command.

| ID | Severity | Claim | Verdict this round |
|----|----------|-------|--------------------|
| R1 | Important | Vintage test leaked temp SQLite databases into the source tree; the single `currentDbPath` was overwritten per call and `afterEach` unlinked only the last `.db`, never the `-wal`/`-shm` sidecars. | **Fixed — verified.** `sqlite-vintages.test.ts:29-37` now appends every path to `tempDbPaths`, and `:149-159` sweeps every recorded path plus `-wal`/`-shm`. Ran the suite: 1 pass, 0 fail, and `ls src/db/adapters/ | grep test-sqlite-vintage` finds zero leaked files afterwards. |
| R2 | Important | Generator proceeded silently when `git tag` failed or returned nothing; write mode then deleted every checked-in fixture and exited 0. | **Fixed — verified both branches.** `generate-sqlite-vintages.ts:101-111` throws on `!ok` with stderr and on an empty tag list. Executed: `GIT_DIR=/tmp/nonexistent-repo-for-probe bun scripts/generate-sqlite-vintages.ts --check` → exit 1, "git tag failed"; same against a freshly initialized empty bare repo → exit 1, "git tag listed no tags", in both `--check` and write mode, with all 10 fixtures untouched (`ls …/sqlite-vintages/ | wc -l` = 10). |
| R3 | Important | A `rev-parse` miss was classified as "tag predates the adapter", so a future adapter-path rename silently ended vintage coverage with a green check. | **Fixed — verified with a control.** `generate-sqlite-vintages.ts:120-133`: a miss on a tag newer than tags carrying the file now throws. Executed with a temp copy outside the repo re-pointing `ADAPTER_REPO_PATH` at `packages/core/src/test/setup.ts` (present in v0.2.13, deleted from v0.9.0) with extraction stubbed to isolate the seam → exit 1, "missing from a tag newer than tags that carry it". Control probe with a path missing from all 23 tags correctly stayed in the legitimate prefix case (exit 2 staleness report, no throw). |

## Discoveries

No new discoveries this round. The round-1 consolidation stands: 4 accepted discoveries — three adjacent (independent tag enumerators; disclosed teeth-check limitation; guarded `allow_env_keys` duplication) and one non-blocking scope conflict (the plan's "no allowlist" premise is falsified by the v0.3.1 vintage; the shipped narrow exception is required and follows the plan's own risk prescription). Full lists: [discoveries.json](../discoveries.json) and [discoveries.md](../discoveries.md). Adjacent discoveries do not affect readiness, and the scope conflict is resolved inside the plan's documented contingency — no replan.

## Review coverage

Continuation review, round 3. One continuation reviewer verified the prior report (rounds 1–2, full review of PR #2866) and examined the correction delta `2b8d0c13ef35e6f3747f844cb39e5caa3577ef18..f2a14cf9fbdd70439fcbbe4da149b361417da6f2`. The original round-1 selected concerns were: code, seams, simplify (required), plus errors, comments, types (optional; tests and docs disabled). All three open blockers were verified by execution (commands above); the delta's changed lines were read in full in both files, with the test file and generator read end to end at head. The specialist lenses did not rerun. Evidence not inspectable: CI execution of the new `schema-upgrade` step and of the unit test on windows-latest remains unobserved (no local runner for it; recorded since round 1). No other evidence gaps.

PR #2876, opened 2026-08-27
test(core): wait on the conversation lock instead of the clock

### LINKED ISSUE #2689: test(core): ConversationLockManager FIFO test is a wall-clock race and fails intermittently on windows-latest
## Problem

`ConversationLockManager > queued messages process in order after completion` (`packages/core/src/utils/conversation-lock.test.ts:130`) fails intermittently on `windows-latest`. Observed failing on PR #2681 at `9ca72833`, in a commit whose diff touches only `renovate.json`, `CHANGELOG.md`, `AGENTS.md`, `CONTRIBUTING.md` and two `.archon/workflows/maintainer/*.yaml` files — nothing under `packages/core/`. The same test passed on the immediately preceding head, `7acbc5ca`.

```
@archon/core:test | (fail) ConversationLockManager > queued messages process in order after completion [125.00ms]
@archon/core:test |  1 fail
@archon/core:test | Exited with code 1
```

Ubuntu passed in the same run. So does `bun run validate` locally on macOS, repeatedly.

## Cause

The test has no synchronisation — it asserts ordering inside a fixed wall-clock budget:

```ts
const manager = new ConversationLockManager(10);
const executionOrder: number[] = [];

manager.acquireLock('test-conv', async () => {
  executionOrder.push(1);
  await new Promise(resolve => setTimeout(resolve, 30));   // 30ms
});

await new Promise(resolve => setTimeout(resolve, 10));      // 10ms

manager.acquireLock('test-conv', async () => {
  executionOrder.push(2);
  await new Promise(resolve => setTimeout(resolve, 10));    // 10ms
});
manager.acquireLock('test-conv', async () => {
  executionOrder.push(3);
});

await new Promise(resolve => setTimeout(resolve, 100));     // hope it finished
expect(executionOrder).toEqual([1, 2, 3]);
```

Three `acquireLock` calls are fired without awaiting anything they return, and the assertion runs after an unconditional 100 ms sleep. The work is ~40 ms, so the margin is ~60 ms of scheduler slack. That is comfortable on a Linux runner and not on a loaded Windows one, where `setTimeout` granularity and process scheduling are both coarser — the failure took 125 ms, i.e. it overshot the window rather than producing wrong ordering.

This is the "gradient" shape AGENTS.md describes for Windows CI: genuine cost meeting a fixed budget, rather than a bounded stall at Bun's 5000 ms timeout. Note the test is **not** asserting a timing property — it is asserting FIFO ordering. The clock is only being used as a substitute for waiting on completion.

## Suggested fix

Await completion rather than sleeping. If `acquireLock` returns a promise, collect and `await Promise.all([...])` before asserting; the ordering invariant is then tested with no wall-clock dependency at all and the test gets faster as a side effect. If it does not return one, the smallest honest fix is a poll-until-`executionOrder.length === 3` with a generous ceiling, so the budget bounds failure rather than success.

Raising the 100 ms would bury it rather than fix it, and would disarm the only signal that catches a real ordering regression.

Worth checking the sibling tests in the same file (`error in handler does not prevent queue processing` at 78 ms, `stats show correct active conversation IDs` at 141 ms) for the same pattern while in there.

## Impact

A required check fails on unrelated PRs, which costs a re-run and, worse, trains reviewers to re-run red Windows jobs without reading them. `test (windows-latest)` is a required check and `fail-fast: false` is deliberately set so both legs always report (`.github/workflows/test.yml:14-18`), so this is exactly the kind of noise that setup exists to keep meaningful.

## Context

Found while working #2666 / PR #2681. Not caused by that PR; re-running the job cleared it.

Wirasm:
## Problem and outcome

`ConversationLockManager > queued messages process in order after completion` fails intermittently on `windows-latest` and blocks unrelated PRs. It asserts FIFO ordering after an unconditional 100ms sleep that has to cover ~40ms of handler work, so a loaded runner exhausts the budget before the queue drains. The reported failure took 125ms and the ordering was never wrong — only the wall-clock margin ran out.

- **Outcome:** the file contains no `setTimeout` and no `Date.now` call. Every wait is a wait on a state the manager reports, so no amount of scheduler pressure can change the result. It runs in ~35ms instead of ~800ms.
- **Invariant:** the tests still fail on a real regression. FIFO drain order, per-conversation serialisation, the concurrency cap, lock release, error containment, and the freed-capacity wakeup are each proven by a mutation that turns the suite red.
- **Scope boundary:** `packages/core/src/utils/conversation-lock.ts` is unchanged. No test budget or timer value was raised anywhere.
- **Root cause:** `acquireLock` resolves with the acquisition status, not with handler completion, so the original tests dropped the handler promises entirely and used sleeps as a stand-in for completion. The clock was substituting for a signal that the handlers themselves can produce.

## Review guidance

- **Feedback requested:** whether the microtask-drain helper is the right waiting primitive here, and whether the reworked assertions still cover what the originals intended.
- **Start here:** `packages/core/src/utils/conversation-lock.test.ts:39-66` — `drainUntil` and its two wrappers are the load-bearing change; every other edit follows from them.
- **Review order:** the three helpers at the top, then `queued messages process in order after completion` (the test named in #2689), then `multiple conversations process concurrently` (the one that asserted scheduler jitter).
- **Lower-attention areas:** `initializes with correct maxConcurrent` and `getStats returns empty state initially` are untouched; they had no timing dependency.
- **Known risk or uncertainty:** `drainUntil` bounds itself at 100 microtask ticks. That is a bound on failure, not on success — the manager's queue handoff and cleanup are entirely microtask-driven, so a correct implementation converges in a handful of ticks and a broken one fails immediately with a message naming the state it never reached.

## Solution

Handlers are now gated: each one appends its label to a shared log when it starts and stays in flight until the test releases it. That gives the test both halves of what the sleeps were faking — a start signal it can wait for, and control over when work finishes.

Waiting became "drain microtasks until the manager reports this state" rather than "sleep long enough". Ordering is asserted against the recorded start log instead of being assumed from a wall-clock margin, so a wrong pick fails on an ordering diff rather than blocking on a handler that never runs. Concurrency is asserted from the manager's own `active` count and `activeConversationIds` with all three handlers deliberately still in flight, which is a stronger claim than the `maxDiff < 20ms` jitter check it replaces. Queueing decisions are asserted against the `LockAcquisitionResult` status the manager already returns and nothing previously tested.

One test changed its claim rather than only its mechanism: `stats show correct active conversation IDs` only ever checked the set at one moment, which the concurrency test now covers. It became `stats stop reporting a conversation once it finishes`, which checks that the id set tracks completion and not just starts.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | 10 hardcoded sleeps pace 9 tests; assertions run after a fixed budget | No timers; each assertion runs when the manager reports the awaited state |
| **Failure behavior** | A slow runner fails the FIFO test at 125ms with correct ordering; a real FIFO break also fails, indistinguishably | A slow runner cannot fail it; a real FIFO break fails in 0.26ms with an ordering diff |

## Validation

- `bun test src/utils/conversation-lock.test.ts` — 9 pass, 0 fail, 20 consecutive runs, 32–40ms each — proves stability.
- Same file, 5 runs under 8 saturating CPU processes — 9 pass, 61–71ms — proves the result does not depend on scheduler slack.
- Reproduced the original failure mechanism: with `setTimeout` patched so one 30ms timer is delivered at 130ms, the old FIFO test fails at 115ms with `expected [1,2,3], received [1]` — ordering intact, budget expired, matching the 125ms CI signature. The rewritten file passes unchanged under the same patched clock.
- Mutation testing against `conversation-lock.ts`, each mutation reverted afterwards:

  | Mutation | Result |
  |---|---|
  | `queue.shift()` → `queue.pop()` (LIFO drain) | `queued messages process in order after completion` fails in 0.26ms on the ordering diff |
  | drop the "already active" check | 3 tests fail; first is `expected "queued-conversation", received "started"` |
  | drop the capacity check | `queues message when at max capacity` fails: `expected "queued-capacity", received "started"` |
  | drop `activeConversations.delete` | 7 tests fail with `manager never reached expected state: …` |
  | drop the handler `.catch` | `error in handler does not prevent queue processing` fails on the escaped rejection |
  | drop the `processGlobalQueue` wakeup | `queues message when at max capacity` fails: `3 handler(s) started, saw [conv-1,conv-2]` |

- `bun run lint` — clean. `bun run validate` — exit 0 (bundled/schema/pi-vendor/capability checks, type-check, lint at `--max-warnings 0`, format check, installer tests, full test suite).
- **Not verified:** the fix is not observed green on `windows-latest` itself; that only happens once CI runs this branch. The mechanism is host-independent — a file with no timers cannot lose a timing race — and the reproduction above shows the old test failing under exactly the delivery delay the issue diagnoses.

## Links

- Closes #2689
- Related #2306

## Summary by CodeRabbit

* **Tests**
  * Improved automated coverage for conversation locking and handler execution ordering.
  * Added deterministic validation that conversations are released correctly after all processing completes.
  * Reduced timing-related test variability, supporting more reliable verification of conversation activity behavior.

### DIFF
--- a/packages/core/src/utils/conversation-lock.test.ts
+++ b/packages/core/src/utils/conversation-lock.test.ts
@@ -1,5 +1,70 @@
 import { ConversationLockManager } from './conversation-lock';
 
+/**
+ * Nothing in this file sleeps. `acquireLock` resolves with the acquisition
+ * status, not with handler completion, so the handlers report their own starts
+ * into a log and the test decides when each one finishes. Waiting is always
+ * "drain microtasks until the manager reached this state", and ordering is
+ * asserted against the log rather than assumed from a wall-clock margin.
+ */
+
+interface GatedHandler {
+  /** Lets the handler finish. */
+  release: () => void;
+  handler: () => Promise<void>;
+}
+
+/**
+ * A handler the test drives explicitly: it appends its label to `log` when it
+ * starts and stays in flight until released. `fail` makes it reject on release,
+ * which exercises the manager's error path without a timing assumption.
+ */
+function gate(log: string[], label: string, { fail = false } = {}): GatedHandler {
+  let release!: () => void;
+  const released = new Promise<void>(resolve => {
+    release = resolve;
+  });
+  return {
+    release,
+    handler: async () => {
+      log.push(label);
+      await released;
+      if (fail) throw new Error(`handler ${label} failed`);
+    },
+  };
+}
+
+/**
+ * The manager releases a lock and hands off to the next queued message entirely
+ * on the microtask queue, so draining microtasks converges without touching the
+ * clock. The bound turns a broken handoff into an immediate failure rather than
+ * a hang.
+ */
+async function drainUntil(predicate: () => boolean, expectation: string): Promise<void> {
+  for (let tick = 0; tick < 100; tick++) {
+    if (predicate()) return;
+    await Promise.resolve();
+  }
+  throw new Error(`manager never reached expected state: ${expectation}`);
+}
+
+/**
+ * Waits for one more handler to start, whichever one the manager picked. The
+ * caller then asserts the log, so picking the wrong one fails on the ordering
+ * assertion instead of blocking on a handler that never runs.
+ */
+function drainUntilStarted(log: string[], count: number): Promise<void> {
+  return drainUntil(() => log.length >= count, `${count} handler(s) started, saw [${log}]`);
+}
+
+/** Waits for the manager to have no active conversations and nothing queued. */
+function drainUntilIdle(manager: ConversationLockManager): Promise<void> {
+  return drainUntil(() => {
+    const stats = manager.getStats();
+    return stats.active === 0 && stats.queuedTotal === 0;
+  }, 'idle (no active conversations, empty queues)');
+}
+
 describe('ConversationLockManager', () => {
   test('initializes with correct maxConcurrent', () => {
     const manager = new ConversationLockManager(5);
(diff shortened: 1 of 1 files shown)


Wirasm:
## Ready to merge

This is a single-file determinism rewrite of `packages/core/src/utils/conversation-lock.test.ts` (closes #2689) that replaces 10 hardcoded `setTimeout`/`Date.now` waits with a gated-handler pattern and a 100-tick-bounded `drainUntil(predicate)` microtask drain. Four independent reviewers (code, seams, simplify, tests) and my own mutation testing all converge: the rewrite is correct, the abstraction is proportionate rather than premature, no behavioral coverage was lost (several assertions got strictly stronger), and the design correctly treats handler-start as the only observation point production callers never get either. Two non-blocking Suggestions remain open — a cosmetic no-op drain call and a pre-existing untested fairness path that this PR did not touch.

**Validation:** `packages/core` `bun run test` (full chain, includes the target file) — PASS, exit 0. `packages/core` `bun x tsc --noEmit` — PASS, exit 0. Repo-root `bun run lint` — PASS, exit 0. Target file run 77 times across coordinator + 3 reviewer agents (12 + 35 + 25 + 5) — 9/9 pass every run, 33–56ms, zero flakes. CI on the PR: `test (ubuntu-latest)`, `test (windows-latest)` (the platform that reported the original flake), `schema-upgrade`, `docker-build`, CodeRabbit — all SUCCESS.

### Findings

|---|---|---|---|
| `R1` | Suggestion | A few `drainUntilStarted` calls right after a same-tick `acquireLock` are no-ops (predicate already true) | OPEN |
| `R2` | Suggestion | Cross-conversation fairness ordering in the global capacity queue has no test, old or new (pre-existing gap) | OPEN |

<details>
<summary><code>R1</code> — redundant drain on the immediate-execution path</summary>

**Impact:** None on correctness. At `conversation-lock.test.ts:95, 142, 170`, `await drainUntilStarted(log, N)` follows an `await manager.acquireLock(...)` on the path where `acquireLock` executes the handler synchronously before returning (`conversation-lock.ts:76-109`) — the log entry is already pushed by the time the drain call runs, so its predicate is true on the first check and it never actually awaits a microtask. This is a harmless self-check, not a bug, but it's mildly inconsistent with the file's opening comment that frames every wait as "drain microtasks until the manager reached this state."

**Evidence:** `packages/core/src/utils/conversation-lock.test.ts:95,142,170`; traced against the synchronous branch at `packages/core/src/utils/conversation-lock.ts:76-109`.

**Required outcome:** No correction required. Optional: a one-line comment noting these particular drains are same-tick confirmations rather than genuine waits, if a future reader might otherwise assume every drain call blocks.

**Disposition:** OPEN (non-blocking suggestion; does not affect merge readiness).

</details>

<details>
<summary><code>R2</code> — untested cross-conversation promotion order under capacity</summary>

**Impact:** `processGlobalQueue` (`conversation-lock.ts:183-196`) promotes the first conversation with a pending queue when a global slot frees up, iterating `this.messageQueues` (a `Map`, insertion-ordered). No test — old or new — ever has two *different* conversation IDs queued simultaneously purely for the capacity reason, so nothing proves arrival-order fairness across competing conversations is preserved. A refactor that swapped the `Map` for a plain object, for example, would silently reorder capacity-queued conversations whose IDs look numeric, and every current test would still pass.

**Evidence:** `packages/core/src/utils/conversation-lock.ts:183-196`; existing coverage at `packages/core/src/utils/conversation-lock.test.ts:133-159` only ever queues one excess conversation, so it proves promotion happens but not that it picks correctly among competitors.

**Required outcome:** No correction required in this PR — it is a pre-existing gap in behavior this PR does not touch (the PR's own scope is the determinism rewrite, not new coverage). Worth a small follow-up test (`maxConcurrent = 1`; queue two distinct conversations for capacity; assert promotion order) if anyone touches `processGlobalQueue` next.

**Disposition:** OPEN (non-blocking suggestion; not required by this PR's stated outcome or invariant, so left open rather than declined or force-tracked into a new issue).

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1; independently confirmed scope (test-file-only), 35 stable runs, type-check and lint clean, and the LIFO mutation caught in 0.23ms |
| `seams` | No additional findings — exhaustively traced all 6 production `acquireLock` call sites (github/gitlab/gitea adapters, `api.ts`, and 3 in `index.ts` for Discord/Slack/Telegram) and confirmed none observe handler completion, matching what the test's gated-handler design stands in for; confirmed `getStats()` is the same accessor two production health endpoints already consume |
| `simplify` | No findings — judged the `gate`/`drainUntil` extraction as due (replaces 10 duplicated ad hoc sleeps) rather than premature; measured the real handoff depth empirically at 3 microtask ticks against a 100-tick bound (~33x headroom, not a tuned magic number) |
| `tests` | R2; test-by-test comparison against `dev`'s version confirmed no assertion was dropped or weakened, several were strengthened (status-code assertions, post-completion ID removal), and traced that the concurrency test's `active === 3` assertion cannot pass under a full-serialization or macrotask-delay bug |

#### Validation

|---|---|---|
| `bun test src/utils/conversation-lock.test.ts` (repeated) | PASS | 77 total runs across coordinator (12) and 3 reviewer agents (35+25+5); 9/9 pass every time, 33–56ms, zero flakes |
| `bun run test` (packages/core) | PASS | exit 0, full chain including the target file |
| `bun x tsc --noEmit` (packages/core) | PASS | exit 0, no diagnostics |
| `bun run lint` (repo root) | PASS | exit 0, all 12 package globs clean |
| Mutation: `queue.shift()` → `queue.pop()` (LIFO) | caught | ordering test fails in 0.22–0.26ms with a readable diff (coordinator + code-reviewer both reproduced) |
| Mutation: drop "conversation already active" check | caught | 3 tests fail with `expected "queued-conversation", received "started"` |
| Mutation: drop `activeConversations.delete` (idle predicate never holds) | caught | 7 tests fail fast (~35ms total) via the named `drainUntil` timeout message, not a hang |
| Mutation: serialize all conversations globally (break true concurrency) | caught | `multiple conversations process concurrently` fails with `expected "started", received "queued-conversation"` |
| `git diff dev...HEAD --name-only` | confirmed | exactly one file changed, `conversation-lock.test.ts`; production `conversation-lock.ts` byte-identical to `dev` |
| `grep -E "setTimeout\|Date\.now\|setInterval\|Bun\.sleep"` on the test file | confirmed | no matches |
| CI: `test (ubuntu-latest)`, `test (windows-latest)`, `schema-upgrade`, `docker-build`, CodeRabbit | PASS | all SUCCESS on head `25edb029` |

</details>

PR #2877, opened 2026-08-27
test(cli): the detached stop-lease test no longer depends on the wall clock

Wirasm:
## Problem and outcome

One test in the detached-run-control suite measured how long a real 2,000 ms socket idle timeout took and required the result to be at least 2,000. Windows CI read 1,997 and failed the build ([run 33074248136](https://github.com/coleam00/Archon/actions/runs/33074248136/job/98524179771)). Nothing was wrong with the code under test.

- **Outcome:** the file no longer measures wall-clock time anywhere. The same contract is proven by the order of observable events.
- **Invariant:** while a controller retains an uncommitted stop lease, `close()` must not finish shutting the owner down; and the owner must release that lease on its own, without the controller cooperating.
- **Scope boundary:** test-only. No production file changed, no timer, timeout or budget was raised or added, and the test still runs in ~2.1 s. The `afterEach` and every cleanup line were deliberately left untouched — see the note on #2874 below.
- **Root cause:** not clock granularity on its own. The assertion's entire safety margin is a couple of milliseconds, and it requires the clock domain driving the socket timeout to agree with `Date.now()` to within that. Neither promises it. Instrumenting the original assertion over 8 runs on an idle machine:

  ```
  Date.now=2005 (headroom 5)  performance.now=2004.767  skew=-0.233
  Date.now=2003 (headroom 3)  performance.now=2002.571  skew=-0.429
  Date.now=2003 (headroom 3)  performance.now=2002.654  skew=-0.346
  Date.now=2008 (headroom 8)  performance.now=2008.113  skew=+0.113
  Date.now=2002 (headroom 2)  performance.now=2002.647  skew=+0.647
  Date.now=2003 (headroom 3)  performance.now=2002.794  skew=-0.206
  Date.now=2002 (headroom 2)  performance.now=2001.672  skew=-0.328
  Date.now=2003 (headroom 3)  performance.now=2002.968  skew=-0.032
  ```

  The floor clears by 2–8 ms with the machine doing nothing else, and CI came in 3 ms under. The same table rules out the cheap fix: `performance.now()` has the same headroom and disagrees with `Date.now()` by up to 0.65 ms in both directions, so a higher-resolution clock would have kept the flake.

## Review guidance

- **Feedback requested:** correctness — specifically, whether the reachability probe is a fair substitute for the removed lower bound.
- **Start here:** `packages/cli/src/utils/detached-run-control.integration.spec.ts:189-198` — the ordering proof that replaces the two elapsed-time assertions.
- **Review order:** the test body first, then `endpointIsReachable` at `packages/cli/src/utils/detached-run-control.integration.spec.ts:45-63`.
- **Lower-attention areas:** the two import/constant deletions are consequences of the assertions going away — `DETACHED_RUN_IPC_TIMEOUT_MS` and `LEASE_RELEASE_SLACK_MS` had no other use in the file.
- **Known risk or uncertainty:** the probe must complete inside the 2 s lease window. It is a local socket connect issued roughly 30 ms in, so the margin is about 66×, versus the 3 ms it replaces.

## Solution

The test opens a controller connection, sends `stop`, and then never commits and never hangs up, so the owner's own idle timeout is the only thing that can release `close()`. Rather than time that timeout, the test now watches where `close()` has got to.

`close()` waits on the retained lease **before** it stops the server or unlinks the endpoint. That ordering makes endpoint reachability a direct, untimed report of shutdown progress: while the lease is held the endpoint still answers a connection, and once `close()` returns it refuses. A `close()` that ignored the lease would have called `server.close()` synchronously before its promise was even handed back, so the probe would already be refused.

```ts
const closing = owner.close();
expect(await endpointIsReachable(endpointPath)).toBe(true);   // parked on the lease
expect(owner.isStopRequested()).toBe(true);

await closing;
expect(owner.isStopRequested()).toBe(false);                  // the owner released it
expect(await endpointIsReachable(endpointPath)).toBe(false);  // shutdown completed
```

`endpointIsReachable` is a connect-or-refuse probe: refusal is the endpoint's own answer about whether it is still serving, which is why it needs no clock.

The rest of the file was already condition-driven — `waitFor()` polls a predicate and both detached-subprocess tests await observable effects (the ready file appearing, the descendant's death, the child's `exit`) rather than sleeping. There are no fixed sleeps to remove.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | asserts `2000 <= Date.now() delta < 3000` around a real timer | asserts the endpoint is reachable while `close()` is parked and unreachable after it returns |
| **Failure behavior** | fails on 2–8 ms of clock-domain disagreement, reporting a number rather than a defect | fails only when `close()` skips the lease or never releases it |

## Validation

- `bun test src/utils/detached-run-control.integration.spec.ts` × 8 — 5 pass / 0 fail every run, 2.13–2.15 s with no drift — proves stability.
- **Red proof, `close()` ignores retained stop leases** (`if (stopSockets.size > 0)` → `if (false)`): fails at `expect(await endpointIsReachable(endpointPath)).toBe(true)` with `Expected: true / Received: false`, in **32.66 ms** — proves the lower bound the elapsed-time floor used to carry is still enforced, and now fails fast with a readable diff instead of after a 2 s wait.
- **Red proof, the lease idle timeout never arms** (`socket.setTimeout(IPC_TIMEOUT_MS, …)` → `socket.setTimeout(0)` on the lease branch): the test times out; both gate assertions pass first, so only the release failed — proves the boundedness half. Without the idle timeout `close()` falls back to `OWNER_CLOSE_WAIT_MS` (10 s), which is a 5× separation from the passing time, not a 3 ms one.
- Both injected faults reverted; `git diff origin/dev...HEAD --name-only` returns the one spec file, so production behavior provably did not change.
- `bun run test` in `packages/cli` — all 13 legs, 0 fail.
- `bun run lint` — clean across all 12 package globs. `bun x tsc --noEmit` in `packages/cli` — exit 0.
- Gate coverage was checked rather than assumed: `tsc --listFiles` includes this file (the tsconfig excludes only `**/*.test.ts`, not `*.spec.ts`), and a deliberately injected `const dead: any = 1` produced a real `@typescript-eslint/no-explicit-any` error before removal. Both gates genuinely cover the change.
- **Not verified:** the Windows runner itself. The failure is reproduced here as a margin measurement rather than a platform reproduction, and the fix removes the measurement entirely, so there is no remaining Windows-specific behavior to confirm.

## Notes for the reviewer

**Overlaps PR #2874** (`fix/2306-deterministic-test-cleanup`), which is still open and edits this same file. Its hunks are the import block, the `afterEach` → `trackTempRoot` replacement, and the two `mkdtempSync` call sites. Mine are the import block, the `LEASE_RELEASE_SLACK_MS` → `endpointIsReachable` replacement, and the lease test body. The only textual overlap is the import block, and resolution is mechanical: keep #2874's `trackTempRoots` import and `afterEach` removal, keep my removal of `DETACHED_RUN_IPC_TIMEOUT_MS`. No cleanup or teardown line was touched here. Whichever merges second takes a trivial rebase.

**`DETACHED_RUN_IPC_TIMEOUT_MS` now has no external consumer.** This spec was its only importer; it is still used inside the module via `const IPC_TIMEOUT_MS = DETACHED_RUN_IPC_TIMEOUT_MS`. Un-exporting it and collapsing the two names would be coherent subtraction, but it edits the module that governs process termination for a cosmetic gain, and #2874 still imports the symbol in its version of this file. Left deliberately, flagged rather than done.

**#2859 stays open.** It cites two CLI instances. This PR fixes the lease-floor one. The other — `packages/cli/src/cli.test.ts:173`, `keeps a detached config handoff outside repo env overrides` — is still live on `dev` as a `spawnSync` test under the default 5 s budget; PR #2860, which attempted it, is closed and its `beforeAll` fan-out never landed. No closing keyword for that reason.

## Links

- Related #2859
- Related #2306

## Summary by CodeRabbit

* **Tests**
  * Improved detached-run shutdown coverage by verifying control endpoint availability during lease handling.
  * Replaced timing-dependent checks with reliable state and ordering assertions.
  * Confirmed shutdown requests and endpoint closure occur in the expected sequence.


Wirasm:
## Ready to merge

This rewrite of the lease-floor test in `detached-run-control.integration.spec.ts` is sound. The
diff touches exactly one file (`git diff origin/dev...HEAD --name-only`), and the causal claim it's
built on checks out against the real `close()` implementation: `close()` unconditionally awaits the
retained-lease wait before it destroys sockets, calls `server.close()`, or unlinks the endpoint path
(`detached-run-control.ts:256-284`), so the endpoint-reachability ordering the test asserts is
guaranteed by async/await sequencing, not a lucky timing window. Both fault injections in the PR
body were reproduced independently in this review with matching results, and CI ran the actual test
green on both platforms across two heads. The one finding — a new connect-probe helper that omits an
explicit timeout its two counterparts both carry — is real but did not produce any observed failure
across 8+ local runs, a full-workspace-contention run, or four CI job executions, so it's a
Suggestion, not a blocker.

**Validation:** all green — CI (Windows 8m50s, Ubuntu 3m34s, docker-build, schema-upgrade,
CodeRabbit) and independently rerun locally (`bun run validate` exit 0; target spec 8/8 stable at
2.13s; both fault injections reproduced).

### Findings

|---|---|---|---|
| `R1` | Suggestion | New `endpointIsReachable` probe omits the explicit timeout its production and sibling-test counterparts both carry | OPEN |

<details>
<summary><code>R1</code> — new connect-probe helper drops the timeout bound its two counterparts both carry</summary>

**Impact:** `endpointIsReachable` (`detached-run-control.integration.spec.ts:49-63`) is a third,
independent implementation of "probe this control endpoint": production's `canConnect`
(`detached-run-control.ts:110-127`) binds it with `socket.setTimeout(IPC_TIMEOUT_MS, () =>
settle(false))`, and an identically-named helper already exists in a sibling spec file
(`workflow-terminal-event.integration.spec.ts:49-63`) with a 250ms bound. The new helper has no
`setTimeout` at all — if the probe socket never emitted `'connect'` or `'error'`, the promise would
hang indefinitely, and the test would fail only via Bun's generic 5000ms per-test default rather than
a targeted assertion.

**Evidence:** I could not turn this into an actual failure. Ran the target spec 8x back-to-back
immediately after a fresh `bun install` (2.13s every time, no anomaly on the first run); ran the full
workspace suite (`bun run test`, all 10 packages fanned out via `bun --filter '*' --parallel test`,
real CPU contention with `cli.test.ts` measurably slowing from 25.6s to 27.2s under the same load) —
the target spec's timing was unaffected, still exactly `5 pass, 0 fail, [2.13s]`. A standalone probe
against a nonexistent socket path resolved in 4.5ms. CI ran the exact test green on both platforms
across two heads (pre-merge `cf9afa4c`: Windows 2016ms; final `a5d89a76`: Windows 2031ms, Ubuntu
2005.04ms) — margins of roughly 3x against the 5000ms default, consistent with a local Unix
socket/named pipe connect resolving near-instantly (no network stack involved) rather than being
timing-sensitive.

**Required outcome:** add `probe.setTimeout(IPC_TIMEOUT_MS, () => settle(false))` to
`endpointIsReachable`, mirroring `canConnect`. This is a one-line hardening for consistency with the
two existing patterns in the same package, not a correctness fix — I found no reachable path where
the omission produces different observable behavior on any platform this repository targets.

**Disposition:** OPEN (non-blocking). The seam-analyzer initially rated this Important, backed by a
claimed reproduction of the spec timing out at ~5000ms "on the very first invocation." Asked to
reproduce it again with the literal output: it confirmed it could not. Its original failure was the
very first `bun test` invocation in that worktree; the 13 runs since (3 immediately after, 3 more
under synthetic CPU load, 5 more on request — the last batch under genuine contention from a
long-running background process, load average 6.91-7.58 on 11 cores) all passed at ~2.13-2.18s. It
now attributes the original failure to unrecorded environmental noise, not the code, and agrees with
downgrading to Suggestion. Combined with my own evidence (fresh install, 8x repetition,
full-workspace contention, four CI job runs across two platforms and two heads, all passing with
~3x margin), the missing timeout bound is a real, verified inconsistency worth the one-line fix, but
nothing found — by either of us — shows it causing a reachable failure. Does not block this PR.

A related, already-disclosed tradeoff: the old test's explicit upper-bound assertion
(`elapsedMs < ...`) is gone, so a regression that fell through to the 10s `OWNER_CLOSE_WAIT_MS`
backstop would now surface as a generic Bun timeout instead of a targeted assertion. The PR body
states this plainly and explains why: asserting it explicitly would reintroduce the exact wall-clock
boundary this lane exists to remove. I verified the primary invariant this test protects — `close()`
respecting the retained lease — still fails fast and precisely (32ms, `Expected: true / Received:
false`) when broken. Not logged as a separate finding; it's an accepted, reasoned tradeoff rather
than an oversight.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | `R1` |
| `seams` | `R1` |
| `simplify` | No additional findings |

#### Validation

|---|---|---|
| `git diff origin/dev...HEAD --name-only` | PASS | exactly one file: `packages/cli/src/utils/detached-run-control.integration.spec.ts`; +37/−14 |
| `bun test src/utils/detached-run-control.integration.spec.ts` × 8 (isolated, post fresh install) | PASS | 5 pass / 0 fail every run, 2.13s each, no drift |
| Fault injection 1 — `close()` ignores retained lease (`stopSockets.size > 0` → `false`) | PASS (caught) | fails in 32.48ms at `expect(await endpointIsReachable(endpointPath)).toBe(true)`, `Expected: true / Received: false`; reverted, `git status` clean after |
| Fault injection 2 — lease idle timeout never arms (`socket.setTimeout(IPC_TIMEOUT_MS, ...)` → `socket.setTimeout(0)`) | PASS (caught) | times out at 5001.52ms after 2 gate assertions pass; reverted, `git status` clean after |
| `bun run test` (root, full workspace, real parallel contention) | PASS (target spec) | target spec unaffected at 2.13s under contention that slowed `cli.test.ts` from 25.6s→27.2s |
| `bun run validate` | PASS | exit 0, no nonzero fail counts across any package |
| `bun x tsc --noEmit` (packages/cli) | PASS | exit 0 |
| `bun run lint` (root) | PASS | clean across all 12 package globs |
| CI — `test (windows-latest)`, final head | PASS | run 33121477699; target spec `(pass) ... [2031.00ms]` (job log, `gh api .../jobs/98689131872/logs`) |
| CI — `test (ubuntu-latest)`, final head | PASS | run 33121477699; target spec `(pass) ... [2005.04ms]` (job log, `gh api .../jobs/98689131616/logs`) |
| CI — `docker-build`, `schema-upgrade`, CodeRabbit | PASS | `gh pr checks 2877` |
| Live head re-check before publication | PASS | `gh pr view 2877` headRefOid still `a5d89a76...`, state OPEN, not merged |

</details>

PR #2878, opened 2026-08-27
fix(core): SqliteAdapter finalizes prepared statements so close() releases the db file

### LINKED ISSUE #2875: fix(core): SqliteAdapter never finalizes prepared statements — close() leaves a zombie connection pinning the db file
## Problem

`bun:sqlite`'s `Database.close()` calls `sqlite3_close_v2`, which does not release the database file while any prepared statement on the connection is unfinalized — the connection lingers as a zombie until GC finalizes the statements. `SqliteAdapter` prepares statements it never finalizes: `initSchema()` prepares ~8 at construction (sqlite.ts:185, 214, 253, 267, 287, 321, 338, …) and `query()` prepares on every call.

Measured during the PR #2866 investigation (probe evidence: https://github.com/coleam00/Archon/issues/2306#issuecomment-5445101081): 66 file descriptors open on temp databases at teardown; finalizing only test-owned statements left 55 (all adapter-owned); only a forced GC pass reached 0. POSIX `unlink` tolerates the zombies; Windows returns EBUSY — which is exactly how it surfaced (the vintage test failing on windows-latest only).

## Production consequence

`await adapter.close()` resolving does NOT mean the database file is released. On Windows, any delete/move/replace of the db file right after close can fail; on every platform, native statement handles accumulate until GC. Tests currently work around it with `Bun.gc(true)` before cleanup (PR #2866's vintage test — the workaround is load-bearing there, not decoration).

## Desired outcome

The adapter finalizes every statement it prepares: either finalize-after-use in `query()` and friends, or track prepared statements and finalize them in `close()` before closing the connection. After `await adapter.close()` resolves, the database file is immediately deletable on Windows with no GC assistance.

## Invariant

Closing the adapter releases every native resource it acquired. No caller needs `Bun.gc` to make close() mean closed.

## Regression proof

- A test that opens an adapter, runs schema init + a few queries, closes it, and immediately unlinks the db file — must pass on windows-latest without any GC call.
- The `Bun.gc(true)` workaround in `sqlite-vintages.test.ts` can then be removed (leave the bounded retry backstop).

## Notes

Probe ladder (bun 1.3.11): `db.run`/`db.exec` leak nothing; one unfinalized `db.prepare` pins 4 fds; `close(true)` on a live statement throws "database is locked". Check whether statement caching is worth keeping for hot paths while still finalizing on close.

Wirasm:
## Problem and outcome

`bun:sqlite`'s `Database.close()` does not release the database file while any prepared statement on the connection is unfinalized — the connection lingers as a zombie until GC. `SqliteAdapter` prepared statements it never finalized (one per `query()` call, plus roughly ten in `initSchema()` helpers), so `await adapter.close()` resolving did not mean the database file was released. On POSIX, native statement handles accumulated until GC; on Windows, any delete/move/replace of the db file right after close could fail with EBUSY, which is how the bug surfaced in CI.

- **Outcome:** After `await adapter.close()` resolves, the database file is immediately deletable — no GC assistance needed on any platform.
- **Invariant:** Closing the adapter releases every native resource it acquired.
- **Scope boundary:** No statement caching, no close-time tracking set, no change to `db.run`/`db.exec` call sites (a probe on bun 1.3.11 showed they leak nothing). The `Bun.gc(true)` workaround in `sqlite-vintages.test.ts` is intentionally not removed here — that file is not on this base; it ships via #2866 and the workaround can be removed there after this lands.
- **Root cause:** `SqliteAdapter` called `db.prepare()` directly and never called `stmt.finalize()`; `sqlite3_close_v2` leaves the connection alive until unfinalized statements are collected.

## Review guidance

- **Feedback requested:** Correctness of the finalize-in-`finally` pattern and its coverage — every `prepare()` in the class should now route through the wrappers.
- **Start here:** `packages/core/src/db/adapters/sqlite.ts:146` — the three private wrappers (`prepareAll`, `prepareGet`, `prepareRun`) that finalize in a `finally` block, so native handles are released whether the operation succeeds or throws.
- **Review order:** `sqlite.ts` wrappers → their call sites in `query()` and the `initSchema()` helpers → the two regression tests in `sqlite.test.ts`.
- **Lower-attention areas:** The `initSchema()` call-site edits are mechanical one-to-one swaps of `db.prepare(...).get()/.all()` for the wrapper equivalents.
- **Known risk or uncertainty:** Finalize-per-use adds one finalize per operation on hot paths. This was accepted over the issue's alternative (track-and-finalize-on-close) because the adapter has no statement cache — it prepares fresh per call — and close-time tracking would mean an unbounded set for no benefit.

## Solution

Every `prepare()` in `SqliteAdapter` now funnels through `prepareAll`, `prepareGet`, or `prepareRun`. Each wrapper calls `stmt.finalize()` in a `finally`, covering the error path as well — a statement that prepares cleanly but fails on execution is still finalized. `query()` (SELECT, INSERT-with-RETURNING, and standard write branches) and the schema-init helpers (`hasAnyArchonTable`, `recordSchemaVersion`, `migrateColumns`) use the wrappers. `db.run`/`db.exec` calls are untouched: they allocate no statement handles, confirmed by probe.

Two regression tests were added to `sqlite.test.ts`:

1. "finalizes every statement it prepares by the time close() resolves" — wraps `Database.prototype.prepare` (scoped to the test, restored in `finally`) to count preparations and finalizations; runs schema init, an insert, a select, a failing statement (primary-key violation, exercising the `finally` branch), then `close()`; asserts `finalized === prepared > 0`. Verified to fail on the pre-fix code (27 pass / 1 fail with the fix stashed) and pass after (28 pass / 0 fail).
2. "close() leaves the database file immediately deletable" — closes the adapter then `unlinkSync`es the db file with no GC call. Passes trivially on POSIX, but fails with EBUSY on windows-latest if any statement is unfinalized, which is where the real proof lands.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `close()` resolves while the connection stays alive as a zombie until GC; native handles accumulate | `close()` releases all statement handles; the db file is immediately deletable |
| **Failure behavior** | Delete-after-close fails with EBUSY on Windows; test suites need `Bun.gc(true)` before cleanup | Delete-after-close succeeds with no GC call on any platform |

## Validation

- `cd packages/core && bun run test` — 0 failures; `sqlite.test.ts` runs 28 pass / 0 fail with the fix and 27 pass / 1 fail without it, proving the regression test reproduces the bug.
- `bun run type-check` — exit 0.
- `bun run lint --max-warnings 0` — exit 0.
- `bun run validate` (full chain including generated-file checks and all package tests) — exit 0.
- Probe evidence (bun 1.3.11, macOS): `db.run`/`db.exec` leak nothing; one unfinalized `db.prepare` leaves an open fd on the db file after `close()`; forced GC releases it.
- **Not verified:** The Windows EBUSY failure mode is proven by CI (windows-latest) rather than locally; the second test is written so a zombie connection fails it there.

## Links

- Closes #2875
- Relates to #2306 (probe evidence)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQLite database cleanup so prepared statements are finalized reliably when the database closes.
  * Database files can now be deleted immediately after closing, including on Windows.
  * Preserved existing query, migration, and schema behavior while preventing resource locks from failed operations.

### DIFF
--- a/packages/core/src/db/adapters/sqlite.test.ts
+++ b/packages/core/src/db/adapters/sqlite.test.ts
@@ -2,7 +2,7 @@ import { describe, test, expect, afterEach } from 'bun:test';
 import { SqliteAdapter } from './sqlite';
 import { getSchemaSQL } from '../bundled-schema';
 import { APP_VERSION, readSchemaVersion } from '../schema-version';
-import { Database } from 'bun:sqlite';
+import { Database, type Statement } from 'bun:sqlite';
 import { unlinkSync } from 'fs';
 import { join } from 'path';
 
@@ -1077,3 +1077,72 @@ function raw_query(dbPath: string, sql: string): unknown[] {
     raw.close();
   }
 }
+
+describe('SqliteAdapter native-resource finalization (#2875)', () => {
+  let adapter: SqliteAdapter | undefined;
+
+  afterEach(async () => {
+    if (adapter) {
+      await adapter.close();
+      adapter = undefined;
+    }
+    for (const suffix of ['', '-wal', '-shm']) {
+      try {
+        unlinkSync(currentDbPath + suffix);
+      } catch {
+        /* may not exist */
+      }
+    }
+  });
+
+  test('finalizes every statement it prepares by the time close() resolves', async () => {
+    const originalPrepare = Database.prototype.prepare;
+    let prepared = 0;
+    let finalized = 0;
+    Database.prototype.prepare = function (
+      this: Database,
+      ...args: Parameters<typeof originalPrepare>
+    ) {
+      const stmt = originalPrepare.apply(this, args) as Statement & {
+        finalize: () => unknown;
+      };
+      prepared++;
+      const nativeFinalize = stmt.finalize.bind(stmt);
+      stmt.finalize = () => {
+        finalized++;
+        return nativeFinalize();
+      };
+      return stmt;
+    };
+
+    try {
+      adapter = createTestDb();
+      await insertCodebase(adapter, 'cb-finalize');
+      await adapter.query('SELECT id FROM remote_agent_codebases WHERE id = $1', ['cb-finalize']);
+
+      // A statement that prepares cleanly but fails on execution (primary key
+      // violation) must still be finalized (this is the finally-branch of the fix).
+      await expect(insertCodebase(adapter, 'cb-finalize')).rejects.toThrow();
+
+      await adapter.close();
+      adapter = undefined;
+
+      expect(prepared).toBeGreaterThan(0);
+      expect(finalized).toBe(prepared);
+    } finally {
+      Database.prototype.prepare = originalPrepare;
+    }
+  });
+
+  test('close() leaves the database file immediately deletable', async () => {
+    adapter = createTestDb();
+    await insertCodebase(adapter, 'cb-delete');
+    await adapter.query('SELECT id FROM remote_agent_codebases WHERE id = $1', ['cb-delete']);
+    await adapter.close();
+    adapter = undefined;
+
+    // Passes trivially where POSIX unlink tolerates open handles, but on
+    // windows-latest this fails with EBUSY while any statement is unfinalized.
+    unlinkSync(currentDbPath);
+  });
+});
--- a/packages/core/src/db/adapters/sqlite.ts
+++ b/packages/core/src/db/adapters/sqlite.ts
@@ -8,6 +8,9 @@ import type { IDatabase, QueryResult, SqlDialect } from './types';
 import { createLogger } from '@archon/paths';
 import { APP_VERSION } from '../schema-version';
 
+/** bun:sqlite's get() returns null (not undefined) when the query has no row. */
+type Maybe<T> = T | null;
+
 /** Lazy-initialized logger (deferred so test mocks can intercept createLogger) */
 let cachedLog: ReturnType<typeof createLogger> | undefined;
 function getLog(): ReturnType<typeof createLogger> {
@@ -149,10 +152,10 @@ export class SqliteAdapter implements IDatabase {
     }
   }
 
-  private prepareGet(sql: string, params: SQLQueryBindings[] = []): unknown {
+  private prepareGet<T>(sql: string, params: SQLQueryBindings[] = []): Maybe<T> {
     const stmt = this.db.prepare(sql);
     try {
-      return stmt.get(...params);
+      return stmt.get(...params) as Maybe<T>;
     } finally {
       stmt.finalize();
     }
@@ -242,9 +245,9 @@ export class SqliteAdapter implements IDatabase {
       return;
     }
     try {
-      const existing = this.prepareGet(
+      const existing = this.prepareGet<{ app_version: string }>(
         'SELECT app_version FROM remote_agent_schema_version WHERE id = 1'
-      ) as { app_version: string } | undefined;
+      );
 
       if (!existing) {
         this.db.run(
@@ -347,9 +350,9 @@ export class SqliteAdapter implements IDatabase {
       // not on every open: an unconditional DROP + CREATE rebuilds the index
       // each time the adapter is constructed, and leaves a window where the
       // index is gone if the re-create fails.
-      const existingCodebaseIdx = this.prepareGet(
+      const existingCodebaseIdx = this.prepareGet<{ sql: string | null }>(
         "SELECT sql FROM sqlite_master WHERE type = 'index' AND name = 'idx_conversations_codebase'"
-      ) as { sql: string | null } | undefined;
+      );
       if (existingCodebaseIdx && !(existingCodebaseIdx.sql ?? '').includes('deleted_at IS NULL')) {
         this.db.run('DROP INDEX idx_conversations_codebase');
       }
--- a/packages/core/src/db/adapters/sqlite.ts
+++ b/packages/core/src/db/adapters/sqlite.ts
@@ -67,8 +67,7 @@ export class SqliteAdapter implements IDatabase {
       const sqliteParams = reorderedParams as SQLQueryBindings[];
 
       if (isSelect) {
-        const stmt = this.db.prepare(convertedSql);
-        const rows = stmt.all(...sqliteParams) as T[];
+        const rows = this.prepareAll<T>(convertedSql, sqliteParams);
         return { rows, rowCount: rows.length };
       } else {
         const upperSql = sql.toUpperCase();
@@ -78,8 +77,7 @@ export class SqliteAdapter implements IDatabase {
         // RETURNING results, and its lastInsertRowid is unreliable when
         // ON CONFLICT DO UPDATE fires.
         if (upperSql.includes('RETURNING') && upperSql.includes('INSERT')) {
-          const stmt = this.db.prepare(convertedSql);
-          const rows = stmt.all(...sqliteParams) as T[];
+          const rows = this.prepareAll<T>(convertedSql, sqliteParams);
           return { rows, rowCount: rows.length };
         }
 
@@ -93,9 +91,8 @@ export class SqliteAdapter implements IDatabase {
         }
 
         // Standard INSERT/UPDATE/DELETE without RETURNING
-        const stmt = this.db.prepare(convertedSql);
-        const result = stmt.run(...sqliteParams);
-        return { rows: [], rowCount: result.changes };
+        const rowCount = this.prepareRun(convertedSql, sqliteParams);
+        return { rows: [], rowCount };
       }
     } catch (error) {
       const err = error as Error;
@@ -134,6 +131,40 @@ export class SqliteAdapter implements IDatabase {
 
   async close(): Promise<void> {
     this.db.close();
+  }
+
+  /**
+   * bun:sqlite does not finalize statements automatically when the connection
+   * closes: one unfinalized statement keeps the database file pinned until GC
+   * (delete-after-close fails with EBUSY on Windows). So every prepare() in
+   * this class must funnel through these wrappers, which finalize the
+   * statement within the operation's scope (#2875).
+   */
+  private prepareAll<T>(sql: string, params: SQLQueryBindings[] = []): T[] {
+    const stmt = this.db.prepare(sql);
+    try {
+      return stmt.all(...params) as T[];
+    } finally {
+      stmt.finalize();
+    }
+  }
+
+  private prepareGet(sql: string, params: SQLQueryBindings[] = []): unknown {
+    const stmt = this.db.prepare(sql);
+    try {
+      return stmt.get(...params);
+    } finally {
+      stmt.finalize();
+    }
+  }
+
+  private prepareRun(sql: string, params: SQLQueryBindings[] = []): number {
+    const stmt = this.db.prepare(sql);
+    try {
+      return stmt.run(...params).changes;
+    } finally {
+      stmt.finalize();
+    }
   }
 
   /**
@@ -181,9 +212,10 @@ export class SqliteAdapter implements IDatabase {
 
   /** True when core Archon tables already exist — i.e. this is not a fresh database. */
   private hasAnyArchonTable(): boolean {
-    const row = this.db
-      .prepare("SELECT name FROM sqlite_master WHERE type = 'table' AND name = ?")
-      .get('remote_agent_codebases');
+    const row = this.prepareGet(
+      "SELECT name FROM sqlite_master WHERE type = 'table' AND name = ?",
+      ['remote_agent_codebases']
+    );
     return row !== null && row !== undefined;
   }
 
@@ -210,9 +242,9 @@ export class SqliteAdapter implements IDatabase {
       return;
     }
     try {
-      const existing = this.db
-        .prepare('SELECT app_version FROM remote_agent_schema_version WHERE id = 1')
-        .get() as { app_version: string } | null;
+      const existing = this.prepareGet(
+        'SELECT app_version FROM remote_agent_schema_version WHERE id = 1'
+      ) as { app_version: string } | undefined;
 
       if (!existing) {
         this.db.run(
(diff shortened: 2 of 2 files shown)


Wirasm:
# Review report — run c7ce3f42, round 2

## Verdict

**Ready. Action: `none`.** The single blocking finding from round 1 is resolved exactly by the delta, the correction introduced nothing new, and all required evidence was verified at the new head.

## Accepted contract

Fix #2875 (via PR #2878): `SqliteAdapter` must finalize every prepared statement so that after `await adapter.close()` resolves the database file is immediately deletable, on every platform, including the error path (prepare succeeds, execution throws).

Explicit boundaries: no statement caching; no close-time tracking set; no change to `db.run`/`db.exec` call sites (probed on bun 1.3.11: they leak nothing); the `Bun.gc(true)` workaround in `sqlite-vintages.test.ts` stays untouched here (ships via #2866).

## Reviewed head SHA

`204e8ec283bbf62115096ea922e505c3ad016251` — verified: local checkout HEAD equals it, `gh pr view 2878` reports `headRefOid` equal to it (PR OPEN, DRAFT), and the working tree is clean. This is the next round's cursor. Delta reviewed: `git diff 703cb9e685567c60cf9e802a33b5ae7459e51b23..204e8ec283bbf62115096ea922e505c3ad016251` (one commit, one file: `packages/core/src/db/adapters/sqlite.ts`, +9/−6).

## Findings

No open Critical or Important findings. No rejected findings this round. No Suggestions.

The delta (one commit) fixes round-1 finding R1 by making the wrapper carry the type, with one documented deviation that does not change the finding's outcome:

- `prepareGet` is now `private prepareGet<T>(sql, params): Maybe<T>` where `type Maybe<T> = T | null` (`packages/core/src/db/adapters/sqlite.ts:12`, `:155-158`). The JSDoc on the alias states bun:sqlite's `get()` returns `null` on no row — confirmed against `node_modules/bun-types/sqlite.d.ts:658` (`get(...params: ParamsType): ReturnType | null`).
- Both affected call sites drop their caller-side `| undefined` casts and pass explicit type parameters: `recordSchemaVersion` uses `this.prepareGet<{ app_version: string }>(...)` (`sqlite.ts:248`) and the codebase-index check uses `this.prepareGet<{ sql: string | null }>(...)` (`sqlite.ts:353`). Both consumers' truthy guards now check a truthful `| null` sentinel.
- Deviation: the review proposed an inline `T | null` return; the implementer used the named alias `Maybe<T>` because the bare form trips `@typescript-eslint/no-unnecessary-type-parameters`. The alias preserves the invariant (truthful `null` sentinel, no caller casts), and `bun run lint` passes at the new head, which empirically confirms the rationale. Accepted.
- The third `prepareGet` call site (`hasAnyArchonTable`, `sqlite.ts:218`) is untouched per S2 — it predates the PR and dual-guards both sentinels (`row !== null && row !== undefined`, `sqlite.ts:222`), which remains true. Non-blocking, unchanged from round 1.

Verification run at `204e8ec2`: `bun test src/db/adapters/sqlite.test.ts` 28 pass / 0 fail (including the finalize-counting and delete-after-close tests that prove the accepted outcome); root `bun run type-check` passes across all packages; root `bun run lint` passes with no errors.

## Prior findings

| ID | Severity | Verdict at `204e8ec2` | Evidence |
|----|----------|----------------------|----------|
| R1 | Important | Fixed | The false `\| undefined` sentinel and both unchecked caller casts are gone; the wrapper now carries `Maybe<T> = T \| null` matching `bun-types/sqlite.d.ts:658`. Smallest correction applied in full, via a documented alias. |
| S1 | Suggestion | Resolved by R1's correction | `prepareGet` is now generic, matching `prepareAll<T>`'s pattern; both call sites dropped their `as` casts. |
| S2 | Suggestion | Left as-is (still non-blocking) | `hasAnyArchonTable`'s dual guard (`sqlite.ts:222`) predates the PR and stays truthful. No action required. |

## Discoveries

2 accepted discoveries, both adjacent and carried forward unchanged; neither affects readiness. No scope conflicts. Both raw producer records (`discoveries/review-code.json`, `discoveries/implement.json`) were re-validated against their cited evidence and remain supported.

1. dev branch: `cli.test.ts` uses `rmSync` without importing it, failing the shared test workflow's ubuntu job for every open PR including #2878 — not caused by this PR's diff.
2. The `Bun.gc(true)` workaround in `sqlite-vintages.test.ts` (PR #2866 branch) can be dropped once #2878 lands; the file lives only on that branch.

Records: [discoveries.json](../discoveries.json) and [discoveries.md](../discoveries.md). Adjacent discoveries do not affect readiness.

## Review coverage

Continuation review (round 2). One continuation reviewer verified the round-1 report (`703cb9e6`) and reviewed the correction delta (`703cb9e6..204e8ec2`) in full. Canonical PR comment: https://github.com/coleam00/Archon/pull/2878#issuecomment-5445774061 (updated in place; read-back verified). The original selected concerns from round 1 were code, seams, and simplify, plus comments and types; tests, errors, and docs were disabled by inputs. The specialist fan-out was not rerun; round-1 coverage stands. No evidence was uninspectable: the diff, both call sites, the bun-types declaration, the test suite, type-check, and lint were all verified directly at the new head.

Wirasm:
## Ready to merge

`SqliteAdapter` never finalized the statements it prepared, so `bun:sqlite`'s `close()` left a zombie connection pinning the database file until GC — silently tolerated on POSIX, EBUSY on Windows. The fix routes every `db.prepare()` call site in the class (10 of them, across `query()` and the `initSchema()`/`migrateColumns()` helpers) through three new private wrappers (`prepareAll`, `prepareGet`, `prepareRun`) that finalize the statement in a `finally` block. I independently confirmed every remaining `.prepare(` call in the file lives inside those three wrappers — no missed call site — and that the diff makes no schema/DDL changes. I reverted just `sqlite.ts` to the pre-fix version and reran the new regression test locally: it fails exactly as claimed (`finalized: 0` vs `prepared: 15`), proving the test catches the real bug. CI's Windows job (run 33121700783) shows both new tests actually executing and passing there — the platform where the bug is observable without forcing GC. An earlier CI attempt on this branch failed, but only on an unrelated `ReferenceError: rmSync is not defined` in `packages/cli/src/cli.test.ts`, not on anything sqlite-related. Full core suite, type-check, lint, and format all pass locally against the exact reviewed head. Three specialist review agents (code, seams, simplify) found no Critical or Important defects in the production fix; the only findings are cosmetic type-safety cleanups and a pre-existing, unrelated latent flakiness risk in the test file's own helper functions.

**0 blocking · 4 non-blocking**

**Validation:** full core suite, type-check, lint, and format all pass locally at the reviewed head; regression test proven to fail on pre-fix code and pass on the fix; CI green including Windows, confirmed by job-log grep that the new tests actually ran there.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Pre-existing test helpers (`raw_pragma`/`raw_indexes`/`raw_query`) open their own raw `Database` and never finalize their `.prepare()`d statement — same defect class this PR fixes, in a file it already touches | OPEN |
| `R2` | Suggestion | Strategy choice (finalize-per-use) diverges from the issue thread's later-pinned "manual track-and-finalize [on close]" language; PR discloses and justifies it, and the outcome is proven correct, but the maintainer comment specifically named the other approach | OPEN |
| `R3` | Suggestion | `hasAnyArchonTable()`'s `row !== undefined` check is dead code now that `prepareGet` returns `T \| null` only | OPEN |
| `R4` | Suggestion | `Maybe<T> = T \| null` alias is avoidable — `bun-types`' native `Database.prepare<T, Params>()` generics would type `.all()`/`.get()` without a hand-rolled cast | OPEN |

<details>
<summary><code>R1</code> — Test-only helpers leak the same unfinalized-statement pattern the PR fixes</summary>

**Impact:** `raw_pragma`, `raw_indexes`, and `raw_query` (`packages/core/src/db/adapters/sqlite.test.ts:1050-1079`, unchanged by this PR) each open a fresh `new Database(dbPath, { readonly: true })`, call `.prepare(sql).all()`, and `raw.close()` in a `finally` — but never call `stmt.finalize()`. Per Bun's own documented `close()` semantics (the same semantics this PR's fix relies on), that statement stays alive until GC, so `raw.close()` doesn't actually release the file. These helpers are called from several schema-parity tests that immediately follow with `unlinkSync(currentDbPath)` in the shared `afterEach` — the exact failure shape (`close()` then immediate unlink) this PR's own regression test targets, just via a second, separate connection to the same path rather than the adapter under test. A sibling helper in the same file, `columnsOf` (lines 58-67), does this correctly (`stmt.finalize(); raw.close();`), showing the pattern was already known.

**Evidence:** `packages/core/src/db/adapters/sqlite.test.ts:1050-1079` (missing `finalize()`) vs. `:58-67` (`columnsOf`, correct pattern); CI evidence shows no observed failure from this yet, so it is a latent flakiness risk (JS GC timing dependent) rather than a proven live break.

**Required outcome:** Add `stmt.finalize()` to the three helpers, matching `columnsOf`. Cheap (three one-line additions) and directly on-topic for a PR that already establishes "every `prepare()` in this file must be finalized" as the invariant, but it is test-only code with no production impact — not required for this PR to merge.

**Disposition:** OPEN — non-blocking. Worth a quick follow-up commit or a fast-follow issue; does not affect the correctness of the production fix.

</details>

<details>
<summary><code>R2</code> — Chosen strategy diverges from the issue thread's pinned language</summary>

**Impact:** Issue #2875's body offered two acceptable strategies ("either finalize-after-use in `query()` and friends, or track prepared statements and finalize them in `close()`"). A later comment from the issue author (the same person this review reports to), posted after upstream Bun research, said: "Archon pins 1.3.11 ... so on the current toolchain the viable fix is **manual track-and-finalize** (track statements the adapter prepares; finalize in `close()` before closing)." The PR implements finalize-per-use via wrapper methods instead — `close()` itself (`sqlite.ts:135-137`) is untouched and never sees a tracked set. The PR description discloses this explicitly in its "known risk or uncertainty" section and gives a technical rationale (the adapter has no statement cache, so tracking a set for close-time finalization would add a data structure for no benefit over finalizing eagerly).

**Evidence:** issue #2875 comment (2026-08-27T21:46:08Z) vs. `sqlite.ts:135-171` (no tracking set, `close()` unchanged, finalize happens inline in `prepareAll`/`prepareGet`/`prepareRun`). Functionally, the chosen approach satisfies the invariant either way: because every statement is finalized synchronously right after use, none are ever outstanding by the time `close()` runs, so the acceptance contract ("after `close()` resolves, the file is immediately deletable, no GC") holds regardless of which of the two strategies is used. Proven by the reverted-code regression test and by the passing Windows CI leg.

**Required outcome:** No code change required — the delivered outcome meets the issue's invariant and regression proof under either strategy. Flagging so the maintainer can explicitly confirm the deviation from the pinned wording is acceptable, since the comment reads as a specific recommendation rather than just ruling out `close(true)`.

**Found by:** review coordinator (cross-referencing issue #2875's comment thread against the diff)

**Disposition:** OPEN — non-blocking, for maintainer acknowledgment only.

</details>

<details>
<summary><code>R3</code> — Dead null-check after tightening `prepareGet`'s type</summary>

**Impact:** `hasAnyArchonTable()` (`sqlite.ts:217-223`) still checks `row !== null && row !== undefined`, but `prepareGet`'s return type is now `Maybe<T> = T | null` — `undefined` is unreachable. Purely cosmetic; no functional effect.

**Evidence:** `packages/core/src/db/adapters/sqlite.ts:12` (`Maybe<T>` definition), `:218-222` (`hasAnyArchonTable`).

**Required outcome:** Simplify to `row !== null` whenever the file is next touched. Not worth a dedicated commit.

**Disposition:** OPEN — non-blocking.

</details>

<details>
<summary><code>R4</code> — Hand-rolled `Maybe<T>` duplicates bun-types' own generics</summary>

**Impact:** `Maybe<T> = T | null` (`sqlite.ts:12`) is used exactly twice (its own definition and `prepareGet`'s signature/cast). `bun-types@1.3.10`'s `Database.prepare<ReturnType, ParamsType>()` and `Statement<ReturnType>` already type `.get()` as `ReturnType | null` and `.all()` as `ReturnType[]`. Passing the generic through at the `db.prepare<T, SQLQueryBindings[]>(sql)` call sites in `prepareAll`/`prepareGet` would remove both the local alias and the two `as` casts, aligning with AGENTS.md's preference for well-inferred types over hand-rolled escape hatches.

**Evidence:** `packages/core/src/db/adapters/sqlite.ts:11-12`, `:146-162`; verified by the reviewing agent that the generic form compiles cleanly against the repo's installed `bun-types@1.3.10`.

**Required outcome:** Optional cleanup — swap to the SDK's own generics next time this file is touched.

**Disposition:** OPEN — non-blocking.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R3; also independently confirmed the wrapper methods preserve exact return shapes/params for every call site, and that `finally { stmt.finalize() }` cannot mask the original error (bun-types documents `finalize()` as safe to call multiple times, no throw path; regression test exercises the throwing branch and confirms the original error still propagates) |
| `seams` | R1; also confirmed no other file in the repo constructs a `bun:sqlite` `Database` and calls `.prepare()` directly outside `sqlite.ts`/`sqlite.test.ts`, and that the Postgres adapter has no counterpart `Statement`/`finalize` concept to mirror (pg.Pool has no exposed prepared-statement handle) |
| `simplify` | R4; also judged the three-wrapper split as justified (distinct return shapes: `T[]` / `T\|null` / `number`; collapsing would push normalization into every caller) and the test's `Database.prototype.prepare` monkey-patching as proportionate (no public API exists to observe unfinalized native handles; scoped and restored in the same test) |

#### Validation

|---|---|---|
| `bun test src/db/adapters/sqlite.test.ts` (packages/core) | PASS | 28 pass / 0 fail at reviewed head |
| Same test, `sqlite.ts` reverted to pre-fix (dev) | FAIL as expected | 27 pass / 1 fail — `finalized: 0` vs `prepared: 15`, proving the regression test catches the bug |
| `bun run test` (packages/core) | PASS | 0 fail across all files in the package |
| `bun x tsc --noEmit` (packages/core) | PASS | exit 0 |
| `bun run lint --max-warnings 0` (repo root) | PASS | no lint errors across all packages |
| `bun x prettier --check` (changed files) | PASS | "All matched files use Prettier code style!" |
| CI: `test (windows-latest)`, run 33121700783 | PASS | job log confirms both new tests ran and passed: "finalizes every statement it prepares by the time close() resolves [46.00ms]", "close() leaves the database file immediately deletable [32.00ms]" |
| CI: earlier run 33120033229 (same branch, pre type-fix commit) | FAIL (unrelated) | failure was `ReferenceError: rmSync is not defined` in `packages/cli/src/cli.test.ts:130`; the two new sqlite tests passed in that same run's Windows job log |

</details>

Wirasm:
R2 sign-off (maintainer-session): the finalize-per-use strategy is accepted deliberately. My comment on #2875 sketched track-and-finalize-at-close as the viable route, but the binding constraint was only that `close(true)` is unusable on the Bun 1.3.11 pin. Finalize-per-use satisfies the invariant more strongly (no statement is ever outstanding, so the zombie window never opens rather than closing at close()), costs nothing versus the adapter's existing prepare-per-call pattern, and was independently verified by review. R1 (the test file's own raw_* helpers carrying the same class) stands as a recorded follow-up candidate; R3/R4 cosmetic.

PR #2879, opened 2026-08-27
fix(core): isolation cleanup is pinned by live runs, not conversation history

### LINKED ISSUE #2868: fix(core): isolation cleanup --merged never cleans — historical conversation rows pin every environment
## Problem

Tonight's sweep on the maintainer install: `archon isolation cleanup --merged` → **0 cleaned, 56 skipped**, almost all with "still used by N conversation(s)" (two others legitimately dirty). The merged-cleanup path is inert in practice and the worktree estate grows monotonically.

Mechanism: `getRemovalBlocker` (`packages/core/src/services/cleanup-service.ts:444-457`) calls `getConversationsUsingEnv` (`packages/core/src/db/isolation-environments.ts:249`), which is `SELECT id FROM remote_agent_conversations WHERE isolation_env_id = $1` — it counts every conversation row ever created, with no liveness or terminality filter. Every CLI-launched run creates a conversation that persists, so every environment is pinned forever. The same unfiltered check also gates the single-env removal path (`cleanup-service.ts:285`) and `cleanupToMakeRoom`, which reuses the merged logic.

## Desired outcome (operator decision, 2026-08-27)

A merged environment is cleanable regardless of historical conversation references. The only thing that pins an environment is **live work**: a non-terminal workflow run (running/paused/pending) owning it. Historical conversations are data, not locks.

## Non-negotiable invariants

1. Never remove an environment with a non-terminal workflow run attached.
2. The uncommitted-changes skip stays exactly as is.
3. Removal keeps the database coherent: historical conversations and runs may still reference the environment as data; follow the existing removal path's semantics rather than deleting rows others reference.
4. Cleanup never mutates run or conversation lifecycle state — it removes dead estate only (AGENTS.md lifecycle-ownership rule).
5. Skip reasons distinguish a live-run block from the dirty-worktree block; the historical-reference reason disappears.

## Regression proof

- A merged environment referenced only by terminal runs/conversations is cleaned.
- A merged environment with one running or paused run is skipped, with a live-run reason.
- A dirty merged environment is skipped exactly as before.
- The sweep that motivated this (56 environments, dozens merged and terminal) goes from 0 cleaned to cleaning everything except live and dirty ones.

## Evidence

Maintainer sweep 2026-08-27: 0 cleaned / 56 skipped; run `b4e9597b` and two siblings live at the time were correctly not candidates (unmerged branches); the #2851 delivery worktree — merged as `9f8d1a5c` an hour earlier, fully terminal — was pinned by its conversation row.

Wirasm:
## Problem and outcome

`archon isolation cleanup --merged` reported 0 cleaned, 56 skipped on a maintainer install — almost all with "still used by N conversation(s)". The merged-cleanup path was inert in practice and the worktree estate grew monotonically.

- **Outcome:** A merged or stale environment is skipped only while a workflow run can still claim it. Historical conversation rows no longer pin anything, so `--merged` actually cleans.
- **Invariant:** Cleanup still never removes an environment that live work depends on, and still never mutates run or conversation lifecycle state.
- **Scope boundary:** The uncommitted-changes skip is untouched. `cleanupToMakeRoom` (limit-pressure path) inherits the new semantics unchanged via `cleanupMergedWorktrees`.
- **Root cause:** `getRemovalBlocker` counted every `remote_agent_conversations` row with a matching `isolation_env_id`. Every CLI-launched run leaves a conversation row behind, so every environment — merged or stale — was pinned forever by history, not by work.

## Review guidance

- **Feedback requested:** Correctness of the live-run matching (both attachment routes, both database dialects) and of the operator decision this encodes: conversation references are data, not locks.
- **Start here:** `packages/core/src/db/isolation-environments.ts` — `UNCLAIMABLE_WORKFLOW_STATUSES` defines when a run lets go of its environment, and `getLiveRunOwningEnv` replaces `getConversationsUsingEnv` as the single lock query every autonomous cleanup path now goes through (scheduled merged/stale sweeps, the scheduled path-missing branch, `cleanupMergedWorktrees`, `cleanupStaleWorktrees`, `onConversationClosed`, the `isolation cleanup` CLI command, `reconcileGhosts`, and the container reaper).
- **Review order:** `UNCLAIMABLE_WORKFLOW_STATUSES` + `getLiveRunOwningEnv` → `getRemovalBlocker` in `packages/core/src/services/cleanup-service.ts` → `onConversationClosed` → the two scheduled-sweep branches → `reconcileGhosts` and `cleanupContainerEnvironments`.
- **Lower-attention areas:** Skip-reason strings, log event names, docs prose, and comment updates — mechanical, verified by unit tests and doc-line inspection.
- **Known risk or uncertainty:** A failed run never leaves `failed`, so its environment stays pinned until an operator removes it explicitly (`archon isolation remove`) or resumes and finishes the run. That is deliberate — see "Failed runs keep their estate" below — but it means the estate is not fully self-draining. On the maintainer install this is 22 of 63 otherwise-releasable environments.

## Review corrections (round 1)

All four findings from the first review round are addressed:

- **R1 (Critical) — Postgres parse failure:** `getLiveRunOwningEnv` compared one untyped `$1` against a text extract and `conversations.isolation_env_id` (UUID on Postgres) in one OR, so Postgres rejected the query at parse time and the lock check never ran on Postgres deployments — the #2868 symptom on a different layer. The Postgres branch now casts the UUID column to text (`c.isolation_env_id::text = $1`), keeping the shared parameter text-typed across both OR branches. Coverage: new real-Postgres integration test `packages/core/src/db/isolation-environments.live-run.postgres.integration.test.ts` (opt-in via `ARCHON_TEST_PG_URL`; creates and drops its own scratch database; fails 3/3 without the cast) and a new `postgres-parity` CI job in `.github/workflows/test.yml` with a `postgres:17-alpine` service — the only place the Postgres dialect branch executes, since unit tests mock the pg driver and the main test job has no Postgres service.
- **R2 (Important) — check-before-null ordering:** `onConversationClosed` nulled the closing conversation's `isolation_env_id` before checking for a live owning run. A top-level run attaches to its env solely through that reference, so the check matched nothing and the env could be removed under a running or paused run. The live-run check now runs before the null-out and returns early; when a live run owns the env, the conversation reference and `cwd` are left intact until the run goes terminal and the scheduled sweeps clean up. Regression test asserts neither `destroy` nor `updateConversation` fires; it fails under the old ordering.
- **R3 (Important) — stale shipped docs:** `cli-internals.md` guard-check box and Code pointer now name `getRemovalBlocker()` and the live-run guard; both "conversations reference it" clauses in `.claude/docs/isolation-and-worktree-guide.md` now say a non-terminal workflow run owns the env.
- **R4 (Important) — stale docstring/comments:** the `onConversationClosed` docstring and the `// No other users` comment no longer claim a conversation-count contract.

## Review corrections (round 3)

The external review at `57e0477b` returned three blocking findings. All three are the same invariant — prove no run can still claim an environment before destroying it — reached through the same primitive.

- **R1 (Critical) — a failed run's estate could be destroyed with no recovery path.** "Live" was defined as non-terminal only, but `failed` is in *both* `TERMINAL_WORKFLOW_STATUSES` and `RESUMABLE_WORKFLOW_STATUSES`. Once released, `removeEnvironment` → `WorktreeProvider.destroy` force-deletes the **local** branch (`git branch -D`, independent of merge state and of `deleteRemoteBranch`), and `--adopt`'s Lane 3 fallback checks `localBranchExists`, so it dies with it. Committed-but-unpushed work on a failed run — the normal shape for an agent that commits incrementally and fails before opening a PR — was unrecoverable, on the always-on 6-hour scheduler. The exclusion set is now derived: `UNCLAIMABLE_WORKFLOW_STATUSES = TERMINAL_WORKFLOW_STATUSES − RESUMABLE_WORKFLOW_STATUSES` (today `completed`, `cancelled`), so a new resumable status keeps its pin automatically.
- **R2 (Important) — `reconcileGhosts` had no lock.** It marks any active environment `destroyed` when its worktree path is missing, and `listEnvironments()` calls it *before* the per-item guard in both `isolation cleanup` commands — so a row ghosted there never reached that guard. It now applies the same check as the scheduled sweep's path-missing branch and leaves the row active, visible in `isolation list` with its missing path.
- **R3 (Important) — the container reaper used a structurally weaker check.** `getRunByIsolationEnvId` takes the newest run row and only then tests terminality, so a newer terminal run can shadow an older claimable one and `docker rm -f` proceeds against a still-owned environment. The reaper now calls `getLiveRunOwningEnv`, which filters status inside the query. `listContainerEnvironments` keeps `getRunByIsolationEnvId` — it displays the newest run, which is a reporting concern, not a lock.

Non-blocking: **R4** (SQL-text assertions in `isolation-environments.test.ts`) is fixed — the assertions pinned one dialect branch of a dialect-parallel query while the two integration tests already prove behavior on both; only the parameter list is asserted now, which is what the fix actually changes. **R5** (`RemovalBlocker` if/else not compiler-exhaustive) is declined: it is pre-existing, the union is closed, and it is speculative defense against a third variant that does not exist. **R6** (`/worktree remove` chat command) is deferred to a follow-up issue — it is an explicit single-conversation human action, outside this PR's autonomous-cleanup contract, alongside `isolation remove` and codebase deletion.

### Failed runs keep their estate

This is a deliberate narrowing of #2868's stated outcome. The issue names the pinning set as "running/paused/pending"; invariant 4 of that same issue says cleanup "removes dead estate only". By the codebase's own vocabulary a resumable run's estate is not dead, and the two clauses conflict. This PR resolves the conflict in favour of the invariant, because the failure is unrecoverable and silent.

The cost is measured, not assumed. A read-only probe of the maintainer install (below) shows 22 of the 63 environments the narrow lock would release are held by a failed run. `--merged` still goes from cleaning 0 to releasing 41; the estate drains, just not completely. Draining failed-run estate on a longer or explicit timer is a separate operator decision with its own invariants, not a side effect of this fix.

## Solution

Per the operator decision in #2868, the only signal that pins an environment is live work — but "live" means a run that can still act on the estate, not merely one that has not finished: running, pending, paused, or failed-and-still-resumable.

`getLiveRunOwningEnv` is one dialect-aware query that joins `remote_agent_workflow_runs` to `remote_agent_conversations` and matches the env through both routes the code stamps: the run's own `metadata.isolation_env_id` (container runs, sub-run child worktrees) and the worker conversation's `isolation_env_id` (top-level runs). It orders by `started_at DESC` — the same column `getRunByIsolationEnvId` already orders by — and returns the newest live run's `{id, status}` or null. The Postgres branch casts the UUID conversation column to text so the shared parameter stays single-typed.

Every autonomous route into environment destruction now goes through this one check:

- `getRemovalBlocker` returns a `live_run` blocker (`run <short-id> is <status>`) instead of `in_use` with a conversation count. It drives the scheduled merged sweep, the scheduled stale sweep, `cleanupMergedWorktrees`, and `cleanupStaleWorktrees`.
- `onConversationClosed` (the single-env removal path the issue cites) checks for a live owning run *before* clearing the closing conversation's reference, so the env cannot lose its only pin.
- The scheduled path-missing branch and the `isolation cleanup` CLI loop check before destroying.
- `reconcileGhosts` checks before flipping a missing-worktree row to `destroyed`.
- `cleanupContainerEnvironments` checks before `docker rm -f`.

The routes deliberately left out are the explicit human ones — `/worktree remove`, `archon isolation remove`, and codebase deletion. They are an operator naming an environment, not a background sweep choosing one.

`getConversationsUsingEnv` became dead after these swaps and was removed — no compatibility path remains to mislead the next reader.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Merged/stale envs skipped with "still used by N conversation(s)" for every historical conversation row. | Merged/stale envs skipped only with "run <short-id> is <status>" while a run can still claim them. |
| **Failure behavior** | `--merged` never cleaned; worktree count grew monotonically. | Environments whose runs have all completed or been cancelled are cleaned on the next sweep. A failed run's environment is kept, so its branch survives for `--resume`/`--adopt`. |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[archon isolation cleanup --merged] --> B2[Every env pinned by historical conversation rows] --> B3[0 cleaned]
  end
  subgraph After
    A1[archon isolation cleanup --merged] --> A2{Can a run still claim env? running/pending/paused/failed} -->|yes| A3[skip: run is status]
    A2 -->|no| A4[cleaned]
  end
```

## Validation

- `bun run validate` at repo root — pass (generated-artifact checks, type-check across all packages, `lint --max-warnings 0`, prettier, `test:install`, full suite).
- Real-SQLite integration test `packages/core/src/db/isolation-environments.live-run.integration.test.ts` (8 cases) proves: terminal-only history does not pin; a running run via conversation pins; a paused run via run-metadata stamp pins; a run-less conversation does not pin; **a failed run pins**; **completed and cancelled runs do not**; and **a newer terminal run does not shadow an older claimable one** (the shape R3's old primitive got wrong). Registered in `packages/core/package.json` test chain.
- Real-Postgres parity test `...live-run.postgres.integration.test.ts` covers both attachment routes, the terminal-only case, and the failed-run case on real Postgres. Verified it fails 3/3 with the UUID cast reverted; runs in the `postgres-parity` CI job (skips locally without `ARCHON_TEST_PG_URL`). Passing locally against Postgres 18.
- **Each new guard was observed failing before it was trusted.** Reverting the exclusion set to raw `TERMINAL_WORKFLOW_STATUSES` reds exactly the failed-run test (7 pass / 1 fail) and nothing else. Dropping the status filter from the query so it orders-then-filters reds the shadow test with the other filter tests (3 pass / 5 fail). Removing the `reconcileGhosts` guard reds only the new ghost test (9 pass / 1 fail). Making the reaper ignore `failed` reds only the new container test (61 pass / 1 fail).
- Unit tests: live-run skip with distinguishing reason, a merged env with only historical references is cleaned, scheduled-sweep reason strings, the round-1 ordering regression, `reconcileGhosts` keeps a claimable env, and the reaper skips a failed run's container.
- **Live install, read-only probe (the 0-of-56 scenario):** all three lock definitions were run against the maintainer install's real `archon.db` (read-only; nothing removed). 64 active environments. The old conversation-count lock pins all 64. The shipped narrow lock pinned 1. The corrected lock pins 23 — the 1 genuinely running run plus 22 environments held by a failed, still-resumable run. So `--merged` goes from 0 cleaned to 41 environments released into the ordinary merged/stale and uncommitted-changes filters. Two of those 22 were `running` an hour earlier in the same session and had already turned `failed`: under the narrow definition the next 6-hour sweep would have `git branch -D`'d them. That is R1 reproduced on real data, not a hypothetical.
- **Not verified:** A destructive `archon isolation cleanup --merged` against the maintainer install — several of its environments host live agent work and the command has no dry-run flag, so only the read-only probe above was run.

## Delivery considerations

|---|---|---|
| Observability | Log event names changed: `skip_merged_still_in_use`/`skip_stale_still_in_use` → `skip_merged_live_run`/`skip_stale_live_run`, and `env_still_in_use` → `env_has_live_run`. New events: `skip_path_missing_live_run` (scheduled path-missing branch) and `isolation.ghost_kept_for_live_run` (`reconcileGhosts`). `container_env_reaped` no longer carries `runId` — the reap now only happens when no run can claim the env, so the field was structurally always null. Update any dashboards or alerts keyed on these names. | `packages/core/src/services/cleanup-service.ts`, `packages/cli/src/commands/isolation.ts`, `packages/core/src/operations/isolation-operations.ts` |
| Estate retention | A failed run's environment is now kept indefinitely — `failed` is a terminal status, so nothing releases it on its own. Operators drain it explicitly with `archon isolation remove`, or by resuming the run to completion. On an install that accumulates failed runs this is the dominant remaining source of estate growth. | `UNCLAIMABLE_WORKFLOW_STATUSES` in `packages/core/src/db/isolation-environments.ts` |

## Links

- Closes #2868

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented isolation environments that workflow runs can still claim—including running, pending, paused, and resumable failed runs—from being removed during cleanup.
  * Cleanup now reports environments skipped because they remain in use.
  * Missing-worktree environments remain active when a workflow run can still claim them.
  * Historical conversation references no longer block cleanup after workflows finish.

* **Documentation**
  * Clarified cleanup protection and resumable failed-run behavior.

* **Tests**
  * Added SQLite and PostgreSQL integration coverage for environment ownership and cleanup behavior.


Wirasm:
# Review report — run cd0ae3e4, round 2 (issue #2868 / PR #2879)

## Verdict

**Ready — action: `none`.** Both round-1 blockers are fixed correctly at head `2b7a2196`: every destruction route for a stale or merged environment now refuses when a non-terminal workflow run owns it, and every fixed route has a regression test. The full relevant suites pass in this checkout.

## Accepted contract

Source: PR #2879 body (per scope.md, unchanged from round 1).

- **Required outcome:** `archon isolation cleanup --merged` (and the stale sweep, including the CLI default-mode sweep) must actually clean merged/stale isolation environments. A merged or stale environment is skipped only while a non-terminal workflow run owns it (status outside `TERMINAL_WORKFLOW_STATUSES`). Historical `remote_agent_conversations` rows are data, not locks, and must not pin an environment.
- **Invariant:** Cleanup never removes an environment that live work depends on, and never mutates run or conversation lifecycle state.
- **Explicit boundaries:** uncommitted-changes skip untouched; `cleanupToMakeRoom` inherits via `cleanupMergedWorktrees` without separate rework; no live verification against the maintainer install (integration tests stand in); log event renames (`skip_merged_still_in_use` → `skip_merged_live_run`, etc.) are intentional, not findings.

## Reviewed head SHA

`2b7a21960e220c13e63c248df049d649f4ab3ae6` — matches local checkout HEAD (`2b7a2196`), working tree clean. This is the cursor for the next round.

## Findings

### Critical

None. Round-1 R1 is fixed (see prior-findings table).

### Important

None. Round-1 R2 is fixed (see prior-findings table).

### Suggestions (non-blocking, unchanged from round 1)

- **S1:** `isolation-environments.test.ts:374-385` re-asserts the SQL text and parameter array of `getLiveRunOwningEnv` against a mocked driver — a third synchronized copy of the query. The behavioral parity tests catch regressions without it. Not addressed in this delta; still open as a suggestion.
- **S2:** the scheduled sweep's else-branch treats any non-`live_run` blocker as `uncommitted_changes` for log-event selection (cleanup-service.ts). The two-variant union is closed, so nothing is reachable today; an `else if` would make a future variant a compile error. Not addressed in this delta; still open as a suggestion.

### Correction-delta review (new defects introduced by the correction)

None found. Specific checks that came back clean:

- `isolationCleanupCommand` (packages/cli/src/commands/isolation.ts:128-138): the `getLiveRunOwningEnv` check sits before the destroy call inside the loop; skip increments the new summary counter; the post-loop container reaper was already examined clean in round 1 and is unchanged.
- Scheduled path-missing branch (packages/core/src/services/cleanup-service.ts:488-501): the live-run check precedes `removeEnvironment` and uses `continue` correctly; the merged and stale branches below keep their round-1 `getRemovalBlocker` guards.
- The CLI calls `isolationDb.getLiveRunOwningEnv` (the same query the sweeps use, filtered by `TERMINAL_WORKFLOW_STATUSES`), so semantics cannot diverge from the scheduled sweeps' definition of "live".
- Docs sentence in cli.md accurately describes the lock and why quiet conversations still pass the activity filter.
- The delta touches no producer, schema, or wire contract; no caller of the modified functions changed behavior beyond the intended skips.

### Causal class closure (R1, from round 1)

Invariant: no removal route destroys an environment a non-terminal run owns. Mechanism: a route to `provider.destroy`/`removeEnvironment` that bypasses the live-run lock. Discovery method (round 1): enumerate every route into environment destruction repo-wide.

- Affected member (a), CLI default-mode sweep — **fixed this round** (`isolation.ts:128`, verified in code and by new CLI test `skips a stale environment owned by a live run without destroying it`).
- Affected member (b), scheduled path-missing branch — **fixed this round** (cleanup-service.ts:491-501, verified in code and by new test `skips path-missing environments owned by a live run`).
- Examined clean (round 1, re-confirmed unchanged in this delta): scheduled merged and stale branches, `cleanupMergedWorktrees`, `cleanupStaleWorktrees`, `onConversationClosed`, CLI `--merged` route, container reaper, `reclaimContainerEnv` (intentional post-cancel reclaim), operator-typed `/worktree remove` and `isolation remove` (explicit operator action, outside the autonomous-cleanup contract).

## Prior findings

Continuation from round 1 (cursor `3eee6d64ca1a4075bafb421cf660cac4b3be5ad2` → `2b7a21960e220c13e63c248df049d649f4ab3ae6`):

| ID | Severity | Claim | Verdict at `2b7a2196` |
|---|---|---|---|
| R1 | Critical | CLI default-mode `archon isolation cleanup` destroyed stale environments with no live-run guard; scheduled path-missing branch same class, bounded harm | **Fixed.** CLI loop now checks `isolationDb.getLiveRunOwningEnv(env.id)` before each destroy and skips with `skip_stale_live_run` (isolation.ts:128-138); path-missing branch checks before `removeEnvironment` with `skip_path_missing_live_run` (cleanup-service.ts:491-501). Both verified by reading the current code; proof tests pass (`bun test src/services/cleanup-service.test.ts`: 61 pass / 0 fail; `packages/cli` suite: 20 pass / 0 fail). The optional shared-guard export was not taken; the direct `isolationDb` call reuses the exact same lock function the sweeps use, so the divergence risk R1 described is addressed without the refactor. Uncommitted-changes boundary respected (no new skip added to the CLI merged route). |
| R2 | Important | No regression test proved a live owning run blocks the stale sweep paths | **Fixed.** Three new core tests verified present and passing: `cleanupStaleWorktrees > skips stale worktrees owned by a live run`, `runScheduledCleanup > skips stale environments with a live owning run`, `runScheduledCleanup > skips path-missing environments owned by a live run`; plus two new `isolationCleanupCommand` CLI tests (destroy when no live run; skip without destroy when a live run owns the env). The tests assert skip reasons, empty `removed`, and that the removal path was never reached. Implementation.md additionally reports mutation checks (short-circuiting `getLiveRunOwningEnv` fails the new tests); the passing mutated-behavior claim is consistent with the test assertions I read. |
| S1 | Suggestion | SQL-text assertion is a third synchronized copy | Not addressed; non-blocking, remains a suggestion. |
| S2 | Suggestion | `else if` hardening in log-event selection | Not addressed; non-blocking, remains a suggestion. |

## Discoveries

2 accepted adjacent discoveries, unchanged from round 1 and re-validated at head `2b7a2196` (the delta touches none of their evidence paths; re-checked that no source path issues `DELETE FROM remote_agent_conversations` and that `softDeleteConversation` remains the only deletion implementation):

1. Local Postgres parity evidence (18.0) and cast-claim falsification.
2. Run-conversation hard delete would silently drop the live-run pin; the repo's documented posture keeps it unreachable.

Links: [discoveries.json](../../discoveries.json) · [discoveries.md](../../discoveries.md). These are adjacent to the accepted contract and do not affect readiness. No scope conflicts were proved; no `replan` conditions apply.

## Review coverage

Continuation review, round 2. One continuation reviewer verified the round-1 report's findings against the current code and reviewed the full delta (`3eee6d64..2b7a2196`, 5 files, +199/−6) plus the causal-class members the correction touches. The original selected concerns from round 1 were code, seams, and simplify (enabled: tests, comments, types, docs; disabled: errors) — they were not rerun as separate lenses this round. Evidence inspected directly: both corrected production sites, the five new tests and their assertions, both test suites run in this checkout, the docs change, and the unchanged status of all examined-clean causal-class members. Nothing required for this verdict was uninspectable.

Wirasm:
All four findings from the review report are corrected on `bfb0dc9b` and `3eee6d64` (now pushed):

- **R1:** `getLiveRunOwningEnv`'s Postgres branch now casts `c.isolation_env_id::text = $1`, keeping the shared parameter single-typed. Covered by a new real-Postgres parity test (`isolation-environments.live-run.postgres.integration.test.ts`, opt-in via `ARCHON_TEST_PG_URL`, scratch-database isolation) plus a new `postgres-parity` CI job with a postgres:17-alpine service — the unit suite mocks the pg driver, so CI is where this dialect branch actually executes. Verified the test fails 3/3 with the cast reverted.
- **R2:** `onConversationClosed` now checks `getLiveRunOwningEnv` before nulling the closing conversation's `isolation_env_id`, so a top-level run's only pin survives. Regression test asserts neither `destroy` nor `updateConversation` fires under a live run.
- **R3/R4:** shipped docs, the `onConversationClosed` docstring, and the stale comment updated to the run-based contract.

Validation rerun: full root test suite, type-check, lint --max-warnings 0, format check, bundled-schema check — all pass. PR body updated accordingly; still draft.

Wirasm:
Reconciled with `dev` at `dd2838c4` (~14 PRs had merged since this branch forked). Head is now `57e0477b`, `MERGEABLE`.

**Conflicts — both purely additive, both sides kept:**

- `.github/workflows/test.yml` — `dev` appended a `Check SQLite vintage fixtures` step to the existing `schema-upgrades` job, and this PR added a new `postgres-parity` job right after it. Adjacent lines, so git could not separate them. Kept `dev`'s step inside `schema-upgrades`, then the `postgres-parity` job. Both jobs are green on the new head.
- `packages/core/package.json` — both sides extended the `test` chain: `dev` added `src/db/adapters/sqlite-vintages.test.ts`, this PR added the two `isolation-environments.live-run*.integration.test.ts` invocations. Took `dev`'s line and re-inserted the two live-run entries in their original position.

Nothing else conflicted. The merged head's diff against `dev` is the same 13 files / +700 −81 this PR carried before the merge, so nothing was lost either way.

**On #2878:** `getLiveRunOwningEnv` and every query this PR touches go through `pool.query`, which routes into the new `prepareAll`/`prepareGet`/`prepareRun` wrappers. The PR adds no adapter code, so no raw `db.prepare` was reintroduced. The real-SQLite integration test constructs an actual `SqliteAdapter`, so it exercises the query through those wrappers.

**Re-verified at `57e0477b`:**

- `bun run validate` — pass (bundled/schema/vendor-map/capability checks, type-check across all 13 packages, `lint --max-warnings 0`, prettier, `test:install`, full parallel suite).
- `packages/core` full chain — pass, every sub-invocation `0 fail`.
- `cleanup-service.test.ts` 61/0, `packages/cli isolation.test.ts` 24/0, live-run SQLite integration 4/0.
- Postgres parity test — 3/0 against local PostgreSQL 18, and the `postgres-parity` CI job passes on this head.

**Live-install evidence for the issue's fourth regression proof.** Both lock queries were run read-only against the maintainer install's `archon.db` — nothing removed:

```
active environments:             64
pinned by OLD conversation lock: 64
pinned by NEW live-run lock:      3

still pinned (live runs):
  - archon/task-archon-piv-loop-1787816956312       -> run 832640a0 is paused
  - archon/task-feat-2884-discovery-terminal-report -> run 65f152bb is running
  - archon/task-feat-2885-cleanup-drift-check       -> run 36935e89 is running
```

All three are genuinely live agent worktrees. The other 61 are released by the lock and fall through to the ordinary merged/stale and uncommitted-changes filters. A destructive `isolation cleanup --merged` was deliberately not run: several of those environments host live work right now, and the command has no dry-run flag. The PR body's Validation section is updated to say this instead of the old blanket "Not verified" line.

Wirasm:
## Ready to merge

**Round 3 correction verification.** All three blocking findings (R1 Critical, R2/R3 Important) are fixed on `d415a65d`, and I independently reproduced every claim in the author's dispositions comment rather than taking it on faith: read the actual diff for each fix, ran the affected test suites myself (isolated per file, matching the project's mock-isolation convention), and re-ran every "seen-red" mutation the author reported — reverting each fix in a scratch copy of the file and confirming it reds exactly the test the author named, then restoring cleanly. All four mutations discriminated as claimed. I also re-ran the read-only probe against the maintainer's real `archon.db` myself: the corrected lock now pins 23 of 65 active environments (22 failed-but-resumable + 1 paused; the author's snapshot showed 1 running instead of 1 paused, expected drift on a live install between two probes taken hours apart — the corrected-lock count of 23 matches exactly). R4 (test redundancy) is fixed honestly — not just trimmed. R5 stays declined for a sound reason (closed union, cosmetic, no live variant to defend against). R6 is deferred to a real, correctly-scoped, currently open follow-up issue (#2893), not silently dropped. CI is 7/7 green on `d415a65d`, type-check is clean, and no schema/DDL changes appear anywhere in the round-3 delta.

**Operator-pending, not a finding:** the fix for R1 has a real, disclosed product cost — a `'failed'` run's environment is now pinned indefinitely (nothing auto-releases it, since `'failed'` never transitions on its own), so `--merged` goes from "0 cleaned" to "41 released" rather than the "63" a purely non-terminal definition would have produced, and draining that estate requires an explicit `archon isolation remove` or resuming the run to completion. This is the correct resolution of a genuine tension inside issue #2868's own text (its pinning set says "running/paused/pending", but its invariant 4 says cleanup "removes dead estate only" — and a resumable run's estate isn't dead by the codebase's own vocabulary). It's clearly disclosed in the PR body's Delivery considerations table and in the shipped docs. Nothing here needs a code change; it's a scope call for the operator to confirm they're fine trading some estate growth for not silently destroying resumable work.

**0 blocking · 0 non-blocking** (2 resolved via disclosed decline/follow-up)

**Validation:** all four fixes independently re-verified with fresh test runs and fresh mutations in a new worktree (see coverage table). CI on `d415a65d`: 7/7 green, including `postgres-parity`.

### Findings

|---|---|---|---|
| `R1` | Critical | Cleanup can force-delete the local branch and worktree of a `'failed'`-but-resumable run, with no recovery path | FIXED |
| `R2` | Important | `reconcileGhosts()` marks an env `destroyed` on a missing worktree with no live-run check, ahead of the CLI commands this PR fixes | FIXED |
| `R3` | Important | Container reap (`cleanupContainerEnvironments`) still answers "is there a live run" via the older, structurally weaker `getRunByIsolationEnvId` | FIXED |
| `R4` | Suggestion | Mocked SQL-text test is a redundant third copy of the query, and only ever exercises the Postgres branch | FIXED |
| `R5` | Suggestion | `RemovalBlocker` if/else isn't compiler-exhaustive (pre-existing, inherited, cosmetic) | DECLINED |
| `R6` | Suggestion | `/worktree remove` chat command has no live-run check either (pre-existing, narrower, explicit human action) | TRACKED FOLLOW-UP |

<details>
<summary><code>R1</code> — Cleanup can permanently destroy a resumable failed run's only estate</summary>

**Impact:** `getLiveRunOwningEnv`'s exclusion set is `TERMINAL_WORKFLOW_STATUSES = ['completed', 'failed', 'cancelled']` (`packages/workflows/src/schemas/workflow-run.ts:123-127`), so a `'failed'` run is never returned as "live." But `RESUMABLE_WORKFLOW_STATUSES = ['failed', 'paused']` (same file, `:130-133`) is the codebase's own definition of a run a user can still act on — `resumeWorkflow` (`packages/core/src/operations/workflow-operations.ts:354-362`) and the server's resume route (`packages/server/src/routes/api.ts:3708`) both accept `'failed'`. The CLI's `--resume` flag is documented as reusing "the existing worktree from the failed run" (`packages/cli/src/commands/workflow.ts:1517-1519`).

Once `getRemovalBlocker` returns `null` for a `'failed'`-owned environment, `removeEnvironment` (`cleanup-service.ts:377-385`) calls `provider.destroy`, whose `WorktreeProvider.destroy` (`packages/isolation/src/providers/worktree.ts:322-324`) unconditionally force-deletes the **local** branch via `deleteBranchTracked` → `git branch -D <branch>` (`worktree.ts:388`) whenever a `branchName` is supplied — which `removeEnvironment` always does — independent of whether `deleteRemoteBranch` was requested and independent of merge state. Only pre-existing uncommitted changes block this; committed-but-unpushed work (the common shape for an agent that commits incrementally and fails before opening a PR) does not.

The designed fallback, `--adopt`, does not recover this either: `resolveWorkflowAdoption`'s Lane 3 (`packages/core/src/operations/workflow-adoption.ts:187-201`) checks `branchExists` → `localBranchExists` (`:78-80`), which is **local-only**. Once cleanup deletes the local branch, Lane 3 throws `"neither its worktree nor its branch ... still exists. There is nothing left to continue — start a fresh run."` There is no remaining recovery path.

Trigger requires no special configuration: `startCleanupScheduler()` runs unconditionally on every server start (`packages/server/src/index.ts:329`) on a 6-hour interval, and the stale-sweep's staleness check (`isEnvironmentStale`, based on last-commit date and `STALE_THRESHOLD_DAYS = 14`) has nothing to do with run status.

**Evidence:** `packages/core/src/db/isolation-environments.ts:268-283` (exclusion set), `packages/isolation/src/providers/worktree.ts:322-324,382-392` (unconditional local branch deletion), `packages/core/src/operations/workflow-adoption.ts:78-80,187-201` (adopt fallback also fails), `packages/workflows/src/schemas/workflow-run.ts:122-133` (`TERMINAL_WORKFLOW_STATUSES` vs `RESUMABLE_WORKFLOW_STATUSES` both include `'failed'` but mean different things). Independently confirmed against the maintainer's real `~/.archon/archon.db` (read-only probe, nothing written): of the 61 active environments the new lock currently releases, 17 have a `'failed'` run as their newest attached run (e.g. `archon/task-fix-2645-detached-exact-resume`, run `92cd40b2...`). `cleanup-service.test.ts` never stubs `getLiveRunOwningEnv`'s absence with a `'failed'`-status run (checked via `grep`), confirming this is an untested gap, not a deliberate, verified tradeoff. This also sits in tension with issue #2868's own invariant 4 ("Cleanup never mutates run or conversation lifecycle state — it removes dead estate only") — by the codebase's own vocabulary, a resumable run's estate is not dead estate.

**Required outcome:** Exclude only the statuses that are both terminal and non-resumable from "live" — e.g. derive the exclusion set as `TERMINAL_WORKFLOW_STATUSES.filter(s => !RESUMABLE_WORKFLOW_STATUSES.includes(s))` (today: `['completed', 'cancelled']`) at `isolation-environments.ts:270-272`, rather than raw `TERMINAL_WORKFLOW_STATUSES`. No new type or `RemovalBlocker` variant needed — the existing `run <id> is failed` display already reads correctly. This is a one-line, single-call-site fix reusing vocabulary the codebase already treats as authoritative for resumability. Alternatively, if the operator wants `'failed'` environments cleaned on a longer/explicit timer, that needs to be a deliberate, separately-invariant-checked decision, not a side effect of this PR's narrowing.

**Found by:** `prp-core:type-design-analyzer` (full causal chain including the adopt-fallback failure); independently corroborated by the reviewing agent's own live-database probe.

**Disposition:** FIXED on `d415a65d`. `packages/core/src/db/isolation-environments.ts:262-264` now derives `UNCLAIMABLE_WORKFLOW_STATUSES = TERMINAL_WORKFLOW_STATUSES.filter(status => !RESUMABLE_WORKFLOW_STATUSES.includes(status))` — verified this evaluates to exactly `['completed', 'cancelled']` — and both the SQL exclusion placeholders and the bound parameters are seeded from that derived array, so a future addition to `RESUMABLE_WORKFLOW_STATUSES` automatically keeps its pin with no second edit. Verified with my own mutation: reverting the constant to raw `TERMINAL_WORKFLOW_STATUSES` reds exactly 1 of 8 tests in `isolation-environments.live-run.integration.test.ts` — `a failed run still pins the env — it remains resumable` — matching the author's reported result precisely. New direct behavioral tests exist on both dialects (SQLite and real Postgres 18, the latter confirmed by running it myself locally against `postgres:18`) proving a `'failed'` run pins its env while `'completed'`/`'cancelled'` release it. Re-ran the read-only probe against the maintainer's real `archon.db` myself: the corrected lock now pins 23 of 65 active environments (22 `failed` + 1 `paused`); the author's snapshot reported 1 `running` instead of 1 `paused` for that last slot — expected drift, since the same run had already progressed on the live install between the two probes, taken hours apart. The count (23) and the dominant `failed` bucket (22) match exactly.

</details>

<details>
<summary><code>R2</code> — <code>reconcileGhosts()</code> destroys the DB row's live-run signal with no live-run check</summary>

**Impact:** `reconcileGhosts` (`packages/core/src/operations/isolation-operations.ts:46-72`) unconditionally marks any active environment `destroyed` when `worktreeExists()` returns false — no call to `getLiveRunOwningEnv` anywhere in the function. It's called from `listEnvironments()` (`:111-134`), which both `isolationCleanupCommand` (`packages/cli/src/commands/isolation.ts:100`, via `listContainerEnvironments`/CLI plumbing) and `isolationCleanupMergedCommand` (`isolation.ts:193`) — the exact `archon isolation cleanup` / `--merged` commands issue #2868 is about — call before their own per-item live-run guard ever runs, so an already-ghosted row never reaches that guard. This is the identical hazard the scheduled sweep's own new "path missing" branch explicitly guards against (`cleanup-service.ts:490-504`: "Even with the directory gone, marking the env destroyed invalidates the live run's resume handle"), but the same protection was never applied here, even though this file is the shared business logic the CLI commands sit on top of.

**Evidence:** `packages/core/src/operations/isolation-operations.ts:46-72,111-134`; not among this PR's 13 changed files (confirmed via `git diff dev...HEAD --name-only`) — pre-existing, but in the exact feature area and causal class this PR is fixing, and reachable from the commands the PR's own description centers on.

**Required outcome:** Add the same `getLiveRunOwningEnv` guard `reconcileGhosts` is missing, mirroring the scheduled sweep's path-missing branch — skip the status flip (and report it, if the caller has a place to surface that) when a non-terminal run still owns the environment.

**Found by:** `prp-core:code-reviewer`.

**Disposition:** FIXED on `d415a65d`. `reconcileGhosts` now calls `getLiveRunOwningEnv` before flipping status, and `continue`s without incrementing the reconciled count when a run still owns the environment — logged as `isolation.ghost_kept_for_live_run`. The env is left `active` in the DB with its missing path, matching the scheduled sweep's existing path-missing behavior, and `EnvironmentListData` was not widened with a new field — the row's continued presence in `isolation list` (with a missing path) is the operator-visible signal, same pattern the codebase already uses elsewhere. The guard covers `cleanupStaleEnvironments` too (confirmed: both call sites go through the same `reconcileGhosts`). Verified with my own mutation: removing the guard reds exactly 1 of 10 tests in `isolation-operations.test.ts` — `leaves a missing worktree active when a run can still claim it` — matching the author's reported result precisely.

</details>

<details>
<summary><code>R3</code> — Container reap still uses the older, structurally weaker live-run check</summary>

**Impact:** `cleanupContainerEnvironments` (`packages/core/src/services/cleanup-service.ts:163-216`) determines liveness via `workflowDb.getRunByIsolationEnvId` (`packages/core/src/db/workflows.ts:450-461`), which orders by `started_at DESC` and takes the single newest row **before** any status filter, then checks only that row for terminality — rather than the new `getLiveRunOwningEnv`, which filters to non-terminal status first. `getLiveRunOwningEnv`'s own doc comment claims the metadata attachment route (the one container runs use) is exactly what this function covers, but the container reap path was never switched to it. If more than one run row ever attaches to the same container env id — plausible via `create()`'s upsert-by-`(codebase_id, workflow_type, workflow_id)` semantics reactivating an env row for a fresh run while an earlier run for that identity is still non-terminal — a newer terminal row can shadow an older live one, and `backend.destroy()` (`docker rm -f` container + volume) proceeds against a still-owned environment.

**Evidence:** `packages/core/src/services/cleanup-service.ts:163-216` vs `packages/core/src/db/workflows.ts:450-461` vs `packages/core/src/db/isolation-environments.ts:257-284`. Trigger condition (multiple run rows on one container env, mixed liveness) is structurally plausible but not confirmed with an end-to-end repro — the query-design divergence itself is directly verified from both function bodies.

**Required outcome:** Point `cleanupContainerEnvironments` at `getLiveRunOwningEnv` in place of `getRunByIsolationEnvId` + the manual terminal check — the metadata-route branch already matches container attachment, so only the call site changes.

**Found by:** `prp-core:seam-analyzer`.

**Disposition:** FIXED on `d415a65d`. `cleanupContainerEnvironments` now calls `isolationEnvDb.getLiveRunOwningEnv` in place of `workflowDb.getRunByIsolationEnvId` + the manual terminal check; the H3 fail-closed try/catch around the lookup is unchanged. Verified the "can only become more protective" claim by reading both queries side by side: `getLiveRunOwningEnv` matches container runs through the same metadata route `getRunByIsolationEnvId` used, plus the conversation route (which never populates for container envs today since `getIsolationProvider()` is hardcoded to `WorktreeProvider`) — so behavior for container envs is unchanged in the common case and strictly safer in the shadow scenario. `container_env_reaped` correctly dropped its `runId` log field with an honest comment explaining why (the reap now only runs when no run can claim the env, so the field was structurally always null) — not a silent regression. Verified with two of my own mutations: (1) making the reaper ignore a `'failed'` owner reds exactly 1 of 62 tests in `cleanup-service.test.ts` — `never reaps a FAILED run's container (still resumable)` — matching the author's report; (2) reverting `getLiveRunOwningEnv`'s query to the old order-then-filter shape (newest row first, status checked after) reds the new shadow-proof test `a newer terminal run does not shadow an older claimable one` in `isolation-environments.live-run.integration.test.ts`, directly proving the query-design fix this finding required. (My mutation's overall fail count differed from the author's reported 3/5 — a different mutation implementation naturally reds a different set of pre-existing tests that depend on run-ordering timing — but the load-bearing shadow test discriminates correctly in both, which is what R3 actually required.)

</details>

<details>
<summary><code>R4</code> — Mocked SQL-text test is redundant and doesn't test what it appears to</summary>

**Impact:** `isolation-environments.test.ts:365-395` (specifically the `toContain` assertions at `:372-385`) re-asserts `getLiveRunOwningEnv`'s literal SQL text — a third synchronized copy alongside the two new real-database integration tests, which already prove the actual behavior (including the exact Postgres UUID-cast fix this duplication exists to protect). Worse than the PR's own round-2 review described it: the file hardcodes `getDatabaseType: () => 'postgresql'` for every test in it (`:12`), so this "unit" test can only ever exercise the Postgres dialect branch — it never touches the SQLite `json_extract(...)` branch its presence implies it guards.

**Evidence:** `packages/core/src/db/isolation-environments.test.ts:12,372-385`.

**Required outcome:** Drop the SQL-text `toContain` assertions; keep the outcome assertions (`result` equals `{id, status}` on a hit, `null` on empty) — the real-SQLite and real-Postgres integration tests already prove behavior for both dialects.

**Found by:** `prp-core:code-simplifier` (also flagged as S1 in the PR's own round-2 review; independently re-confirmed here, plus the postgres-only-hardcoding detail).

**Disposition:** FIXED on `d415a65d`. The `toContain` SQL-text assertions (join clause, `IN` list, `ORDER BY`) are gone; only the outcome (`result` equals `{id, status}`) and the bound-parameter list (`['env-123', 'completed', 'cancelled']` — critically, `'failed'` is correctly absent) are asserted, with a comment explaining why (dialect-hardcoded file; real SQLite and Postgres integration tests prove behavior on both branches). This is an honest removal, not a weakened one — verified by reading the diff directly.

</details>

<details>
<summary><code>R5</code> — <code>RemovalBlocker</code> if/else is not compiler-exhaustive</summary>

**Impact:** `cleanup-service.ts:528,563` — `if (blocker.reason === 'live_run') {...} else {...}` treats any non-`live_run` blocker as `uncommitted_changes` for log-event selection. `RemovalBlocker` has exactly 2 variants today so this is behaviorally exhaustive, but a future third variant would silently misroute the log event name rather than fail to compile. Confirmed pre-existing: this exact two-branch shape existed on `dev` before this PR (discriminating the old `'in_use'` vs `'uncommitted_changes'`) — the PR renamed the variant and swapped fields but inherited the pattern rather than introducing it.

**Evidence:** `packages/core/src/services/cleanup-service.ts:444-446,528,563`.

**Required outcome:** Optional — a `switch` with an exhaustiveness assertion would future-proof this, but it's defensive machinery against a hypothetical third state that doesn't exist yet.

**Found by:** `prp-core:code-simplifier` (also flagged as S2 in the PR's own round-2 review; independently re-confirmed as pre-existing, not introduced by this PR).

**Disposition:** DECLINED, sound reason. Author's stated rationale: pre-existing pattern, closed 2-variant union, affects only log-event selection (never the removal decision), and there is no actual third variant to defend against today — an exhaustiveness assertion here would be speculative defense-in-depth. I agree: this matches AGENTS.md's "do not add... without a current requirement or concrete failure mode," and the finding's own original writeup already scoped it as cosmetic and non-blocking.

</details>

<details>
<summary><code>R6</code> — <code>/worktree remove</code> chat command has no live-run check</summary>

**Impact:** `packages/core/src/handlers/command-handler.ts:578-604` destroys the conversation's own environment (`provider.destroy` + `updateStatus('destroyed')`) with no `getLiveRunOwningEnv` gate. Same unmigrated pattern as R2, but lower risk: it's an explicit single-conversation human action, not an autonomous background sweep, and it's outside the consumer list this PR's own description names.

**Evidence:** `packages/core/src/handlers/command-handler.ts:578-604`; not among this PR's 13 changed files.

**Required outcome:** Same guard as R2, when convenient — not urgent given it requires deliberate human action to trigger.

**Found by:** `prp-core:code-reviewer`.

**Disposition:** TRACKED FOLLOW-UP — verified issue [#2893](https://github.com/coleam00/Archon/issues/2893) (open), which correctly restates the finding, correctly scopes the fix to routing through the same `getLiveRunOwningEnv` primitive (including the resumable-terminal pinning from R1's fix), and correctly frames it as the same causal class at a different, lower-risk (explicit human action) call site. Deferral is reasonable given the narrow trigger.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R2, R6 |
| `seams` | R3 |
| `simplify` | R4, R5 |
| `types` | R1 |

#### Validation — first pass (`57e0477b`)

|---|---|---|
| `bun test src/services/cleanup-service.test.ts` (packages/core) | PASS | 61 pass / 0 fail, 125 expect() calls |
| `bun test src/db/isolation-environments.test.ts` (packages/core) | PASS | 31 pass / 0 fail |
| `bun test src/db/isolation-environments.live-run.integration.test.ts` (packages/core) | PASS | 4 pass / 0 fail (real bun:sqlite) |
| `bun test src/commands/isolation.test.ts` (packages/cli) | PASS | 24 pass / 0 fail |
| `bun x tsc --noEmit` (packages/core) | PASS | clean |
| `bun x tsc --noEmit` (packages/cli) | PASS | clean |
| Mutation check: `getLiveRunOwningEnv` forced to return `null` | DISCRIMINATES | Real-SQLite integration test: 2/4 fail (correctly red); mocked `cleanup-service.test.ts`: still 61/0 (stubs the function, doesn't exercise it — expected) |
| Read-only probe vs `~/.archon/archon.db` (old lock vs new lock, all 64 active envs) | MATCHES PR CLAIM | 64/64 pinned by old lock, 3/64 pinned by new lock, identical run IDs/statuses to the PR's reconciliation comment |
| `git diff dev...HEAD --stat` | MATCHES PR CLAIM | 13 files, +700/−81, exactly as stated in the PR body and reconciliation comment |
| `grep -rn getConversationsUsingEnv` | CLEAN | zero remaining references anywhere in the repo |
| Schema/DDL check | CLEAN | no schema changes anywhere in the diff |
| CI on `57e0477b` | PASS | 7/7 green, including new `postgres-parity` job |

#### Validation — round-3 correction verification (`d415a65d`), fresh worktree, independently re-run

|---|---|---|
| `bun test src/db/isolation-environments.live-run.integration.test.ts` | PASS | 8 pass / 0 fail (real bun:sqlite; 4 new cases) |
| `ARCHON_TEST_PG_URL=... bun test src/db/isolation-environments.live-run.postgres.integration.test.ts` | PASS | 4 pass / 0 fail, run against a real local PostgreSQL 18 |
| `bun test src/services/cleanup-service.test.ts` | PASS | 62 pass / 0 fail |
| `bun test src/operations/isolation-operations.test.ts` | PASS | 10 pass / 0 fail |
| `bun test src/db/isolation-environments.test.ts` | PASS | 31 pass / 0 fail |
| `bun test src/commands/isolation.test.ts` (packages/cli) | PASS | 24 pass / 0 fail |
| `bun x tsc --noEmit` (packages/core, packages/cli) | PASS | clean, both packages |
| Mutation — R1: exclusion set reverted to raw `TERMINAL_WORKFLOW_STATUSES` | DISCRIMINATES | 7 pass / 1 fail in the SQLite integration test — exactly `a failed run still pins the env`, matching the author's report |
| Mutation — R3: query reverted to order-then-filter (old `getRunByIsolationEnvId` shape) | DISCRIMINATES | Shadow-proof test `a newer terminal run does not shadow an older claimable one` reds, directly proving the query-design fix (my mutation implementation differed slightly from the author's, so the total fail count differed — 6/2 vs their 3/5 — but the load-bearing test reds in both) |
| Mutation — R2: `reconcileGhosts` live-run guard removed | DISCRIMINATES | 9 pass / 1 fail in `isolation-operations.test.ts` — exactly the new ghost-guard test, matching the author's report |
| Mutation — R3: container reaper made to ignore `'failed'` owners | DISCRIMINATES | 61 pass / 1 fail in `cleanup-service.test.ts` — exactly the new container-reap test, matching the author's report |
| Re-run read-only probe vs `~/.archon/archon.db`, corrected lock | MATCHES PR CLAIM | 23/65 active envs pinned (22 `failed`, 1 `paused`) — count and dominant bucket match the author's reported 23 (22 failed + 1 running); the running-vs-paused difference is expected drift between two probes taken hours apart on a live install |
| Schema/DDL check on the `57e0477b..d415a65d` delta | CLEAN | no schema/migration/`.sql` files touched, no `CREATE`/`ALTER TABLE` added |
| `git diff 57e0477b..d415a65d --stat` | MATCHES PR CLAIM | 12 files, all within the fixes' stated scope (isolation-environments.ts + tests, cleanup-service.ts + tests, isolation-operations.ts + tests, one comment-only line in isolation.ts, docs) — nothing unrelated smuggled in |
| Causal-class closure re-check: `grep -rn` every `removeEnvironment`/`updateStatus(...,'destroyed')`/`backend.destroy`/`provider.destroy` call site across `packages/core`, `packages/cli`, `packages/server` | CONFIRMED | Every autonomous route is now guarded (directly or via an already-guarded caller); the remaining unguarded sites are explicit human actions (`/worktree remove` → R6/#2893, `archon isolation remove`, codebase deletion) or structurally safe (`orchestrator.ts:188` rolls back an env before any run can be linked to it; `listContainerEnvironments` is read-only display). No missed autonomous route found. |
| CI on `d415a65d` | PASS | 7/7 green, including `postgres-parity` |

</details>

Wirasm:
Round-3 findings addressed on `d415a65d`. R1 was real and already latent on this machine — details below.

## Dispositions

| ID | Disposition | Note |
|---|---|---|
| R1 Critical | **FIXED** | Exclusion set derived as `TERMINAL_WORKFLOW_STATUSES − RESUMABLE_WORKFLOW_STATUSES` (today `completed`, `cancelled`). |
| R2 Important | **FIXED** | `reconcileGhosts` checks the lock before flipping a missing-worktree row to `destroyed`. |
| R3 Important | **FIXED** | Container reaper swapped to `getLiveRunOwningEnv`. |
| R4 Suggestion | **FIXED** | SQL-text assertions dropped; only the parameter list is asserted. |
| R5 Suggestion | **DECLINED** | Pre-existing, closed union, log-event selection only — speculative defense against a variant that does not exist. |
| R6 Suggestion | **DEFERRED** | Explicit human action, outside the autonomous-cleanup contract. Follow-up issue being filed. |

## R1, reproduced on real data

I re-ran the probe with all three lock definitions against the maintainer install's real `archon.db` (read-only, nothing written). 64 active environments:

```
pinned by the OLD conversation lock: 64
pinned by the SHIPPED narrow lock:    1
pinned by the CORRECTED lock:        23   (1 running + 22 failed-but-resumable)
released by the CORRECTED lock:      41
```

Two of those 22 — `task-feat-2884-discovery-terminal-report` (run `65f152bb`) and `task-feat-2885-cleanup-drift-check` (run `36935e89`) — were `running` an hour earlier and reported as live in my previous reconciliation comment. They have since turned `failed`. Under the shipped narrow definition the next 6-hour sweep would have released them and run `git branch -D` on both branches. So this was not a latent risk waiting on an unusual trigger; it was one scheduler tick away on the reviewer's own example install.

I verified every supporting claim in the finding against source before changing anything: `git branch -D` is unconditional whenever `branchName` is supplied (`worktree.ts:322-324,382-392`), `removeEnvironment` always supplies it (`cleanup-service.ts:382`), and adoption Lane 3 checks `localBranchExists` (`workflow-adoption.ts:78-80,187-201`), so `--adopt` cannot recover it. All three hold.

## The tension this creates, stated plainly

Issue #2868 names the pinning set as "running/paused/pending". Its invariant 4 says cleanup "removes dead estate only". By the codebase's own vocabulary a resumable run's estate is not dead, so the two clauses disagree. I resolved it toward the invariant because the failure mode is silent and unrecoverable, but the cost is real and worth an explicit decision: **a failed run never leaves `failed`, so its environment is now pinned until an operator removes it or resumes the run to completion.** `--merged` goes from 0 cleaned to 41 released rather than 63. On an install that accumulates failed runs, that is the dominant remaining source of estate growth, and draining it deserves its own decision rather than riding along here. Called out in the PR body's Delivery considerations.

## R2 — ghost semantics

Taken, because the primitive slots in without changing what the operator sees in a confusing way. Skipping the flip leaves the env active with a missing path, which is exactly what the scheduled sweep's own path-missing branch already does. The env stays visible in `isolation list` with its missing path — that is the operator's signal. I did not widen `EnvironmentListData` with a skipped counter: `listEnvironments` is a list operation, not a cleanup report, and the `isolation.ghost_kept_for_live_run` log event carries the run id and status. The guard also covers `cleanupStaleEnvironments`, which calls `reconcileGhosts` too.

## R3 — clean swap, with one consequence

`getLiveRunOwningEnv` matches both attachment routes where `getRunByIsolationEnvId` matched only run metadata, so the check can only become more protective. The call-site try/catch is unchanged, so the H3 fail-closed contract on a DB error still holds. `listContainerEnvironments` deliberately keeps `getRunByIsolationEnvId` — it displays the newest run, which is reporting, not locking.

One consequence: `container_env_reaped` no longer logs `runId`. After the swap the reap only runs when the lock returns null, so the field was structurally always null; recovering it would mean a second query purely for a log line.

## Causal-class closure

I enumerated every route into estate destruction across `packages/core`, `packages/cli` and `packages/server`. Every autonomous route now holds the lock: the scheduled merged/stale/path-missing sweeps, `cleanupMergedWorktrees`, `cleanupStaleWorktrees`, `onConversationClosed`, the `isolation cleanup` CLI loop, `reconcileGhosts`, and the container reaper. Deliberately outside it: `/worktree remove` (R6), `archon isolation remove`, and codebase deletion (`api.ts:3187`) — all explicit operator actions naming a specific environment. Also examined and correctly unguarded: `orchestrator.ts:188`, which rolls back an environment it just created when linking fails, before any run can own it.

## Naming

I kept the name `getLiveRunOwningEnv` even though "live" now covers a failed run. The name is load-bearing across the `live_run` blocker variant, five log event names, CLI output and shipped docs that this PR's body already documents as an observability contract; renaming ripples through all of it for a preference. Instead the meaning is pinned in one place — the derived `UNCLAIMABLE_WORKFLOW_STATUSES` constant — and every comment and doc line that said "non-terminal run" now says "a run that can still claim it".

## Validation

`bun run validate` — pass. Full `packages/core` and `packages/cli` chains — pass. `cleanup-service.test.ts` 62/0, `isolation-operations.test.ts` 10/0, `isolation-environments.test.ts` 31/0, real-SQLite live-run integration 8/0, real-Postgres parity 4/0 against PostgreSQL 18.

Each new guard was observed failing before I trusted it:

| Mutation | Result |
|---|---|
| Exclusion set reverted to raw `TERMINAL_WORKFLOW_STATUSES` | 7 pass / 1 fail — exactly the failed-run test |
| Status filter dropped from the query (orders, then filters) | 3 pass / 5 fail — includes the new shadow test, i.e. it catches R3's old query shape |
| `reconcileGhosts` guard removed | 9 pass / 1 fail — only the new ghost test |
| Reaper made to ignore `failed` | 61 pass / 1 fail — only the new container test |

The shadow test (`a newer terminal run does not shadow an older claimable one`) is the direct regression proof for R3's structural point: two runs on one env, the older paused and the newer completed, and the lock must return the paused one.

Issue #2575, opened 2026-08-17
Windows CI: the two C5 tests in migrate-state-dir.test.ts time out at 5000ms under parallel load


Wirasm:
## Summary

- **What broke:** the two `describe('subdirectory invocation (C5)')` tests in
  `scripts/migrate-state-dir.test.ts` intermittently time out on `test (windows-latest)`, failing CI
  on unrelated PRs.
- **When it started:** unknown; observed hitting several PRs on 2026-08-17 within the same hour
  (#2569 and #2571 both, 17 minutes apart, same test, same duration).
- **Severity:** `major` — it is not a product bug, but it red-lights unrelated PRs and trains
  everyone to ignore a red Windows check, which is how a real Windows regression gets waved through.

## Steps to Reproduce

1. Open any PR and let `test (windows-latest)` run while other suites are running in parallel.
2. Observe `scripts/migrate-state-dir.test.ts` fail on `migrates the PROJECT root when invoked from a
   subdirectory` and/or its sibling C5 test.
3. Re-run the job with no code change — it passes.

Reproduction rate is load-dependent, not deterministic. It passed on re-run against *strictly more
code* on the same branch and OS, which is the cleanest evidence available that the diff is not the
cause.

## Expected vs Actual

- **Expected:** these tests either pass, or fail with an assertion that names what is wrong.
- **Actual:** they fail at **5016.00 ms** and **5015.00 ms** — Bun's default 5000 ms per-test timeout
  plus overhead. A timeout wearing a failure's clothes.

## Root cause

`withProjectSandbox` (`scripts/migrate-state-dir.test.ts:430`) has exactly two callers, at `:464` and
`:481` — the only two tests in the file that use it, and the only two that fail. Per test it:

1. spawns `bun -e` (`:449`) with a script that does `await import('@archon/core/db/codebases')` and
   `createCodebase(...)` — pulling in the whole core module graph and creating a **real SQLite
   database** in a temp `ARCHON_HOME`, schema applied on connect as every process that opens the DB
   does;
2. then the test body spawns a **second** subprocess via `runIn(...)`, which reopens that database
   and applies schema again.

So each of these two tests pays two cold Bun process starts, two full module-graph loads, and two
schema applies — inside a 5000 ms budget, on the slowest runner, while `bun --filter '*' --parallel
test` has ten package suites competing for the same cores (#2306).

The other 17 tests in the file spawn only the lightweight script and never touch `@archon/core`,
which is exactly why only these two time out.

[...]

PR #2881, opened 2026-08-27
test: stop the C5 fixture spawning a second Bun process

Wirasm:
## Problem and outcome

Two Windows CI tests keep failing at hard boundaries they should never be near. The two subdirectory-invocation ("C5") tests in `scripts/migrate-state-dir.test.ts` time out against Bun's 5000 ms default because each starts *two* cold Bun processes, and only one of them is the thing being tested. Separately, the connect probe added by #2877 was the only one of three near-identical probes in the repo with no timeout bound.

- **Outcome:** each C5 test now starts one cold Bun process instead of two, and the spec's probe is production `canConnect` rather than a third copy of it. The C5 pair runs in ~160 ms locally, down from ~250 ms, on a machine where process startup is nearly free.
- **Invariant:** the subprocess that runs `scripts/migrate-state-dir.ts` is the contract under test and stays a real process — this PR only removes the *setup* spawn. Both C5 tests must still fail when the script's guards break, and the sandbox must resolve the same dialect on the write and the read side.
- **Scope boundary:** no timer, timeout, or budget is raised anywhere. The only production change is exporting `canConnect` so the spec can import it; it adds no behavior. The third item in scope, `cli.test.ts`'s detached-config handoff test, is deliberately left unchanged — see below.
- **Root cause:** the C5 fixture spawned `bun -e` to import `@archon/core`'s whole module graph purely to write one registry row. The row itself measures 3-6 ms of work.

## Review guidance

- **Feedback requested:** whether the fixture's raw `INSERT` is the right trade against `createCodebase`, and whether leaving the `cli.test.ts` test unchanged is the right call.
- **Start here:** `scripts/migrate-state-dir.test.ts:480` — the setup spawn replaced by an in-process adapter write.
- **Review order:** the `withProjectSandbox` change first, then `childEnv` at `scripts/migrate-state-dir.test.ts:119` (which keeps that write and the script's read on the same dialect), then the `canConnect` export and the duplicate it replaces.
- **Known risk or uncertainty:** the C5 overrun is a `windows-latest` phenomenon and does not reproduce on macOS. One Windows job on this branch did get past the `@archon/workflows` rotation and measured the pair at 625 ms and 703 ms against the 5000 ms budget, but that is one sample of a load-dependent failure — see Validation.

## Solution

`withProjectSandbox` needs one row in the codebase registry so the script's `findCodebaseByPathPrefix` lookup has something to climb to. It used to get that row by spawning `bun -e` with a snippet importing `@archon/core/db/codebases`, because that module reaches the database through the module-level connection singleton — which, in-process, would pin the first test's temp `ARCHON_HOME` and then fail with `SQLITE_IOERR_VNODE` once the directory was torn down. That hazard is real, but a subprocess is not the only way around it.

The row now goes in through a directly constructed `SqliteAdapter`, closed in a `finally` before the sandbox goes away. It is the same adapter the script's own lookup opens, so the fixture cannot drift from the schema, and it takes an explicit path so no singleton is involved. Only `name` and `default_cwd` are written, because those are the only columns `resolveRepoProjectIdentity` reads; `ai_assistant_type` and `kind` fall to the column defaults, which are the values `createCodebase` supplied anyway.

That makes the sandbox's dialect matter. The fixture writes SQLite, but the migration-script child used to inherit `DATABASE_URL`, so under the documented Postgres mode the child would read a registry the fixture never wrote to and both C5 tests would fail looking like a climbing regression. `childEnv` now forces `DATABASE_URL: ''` for every child, pinning the dialect the way `ARCHON_HOME` already pins the destination. It also stops the fixture writing test rows into a contributor's real database, which the old `bun -e` subprocess did. The one test that wants the Postgres branch passes a real DSN back through `extraEnv`.

Forced empty rather than deleted, and that distinction is load-bearing: a spawned Bun child re-runs automatic `.env` loading from its own cwd and fills in every key the passed environment omits, so a repo-root `.env` — which is how Postgres mode is configured — puts a deleted key straight back. An empty string is a key that is present, which dotenv leaves alone, and `getDatabaseType()` reads it as falsy. `packages/cli/src/cli.test.ts` already suppresses its own inherited keys this way. The first attempt at this fix used `delete` and looked green only because the worktree it was verified in had no `.env`.

Opening the adapter announces its schema init at info, so the fixture silences the logger around that block and restores the prior level in the same `finally`. `setLogLevel` mutates the process-wide root logger and `bun test ./scripts/` runs all five files in one process, so it is scoped rather than set once at module load.

For the probe, the spec no longer carries its own copy: `canConnect` is exported from `detached-run-control.ts` under the same "Exported for testing" convention `calculatePortOffset` uses, and the spec imports it. That removes the third divergent copy rather than making it consistent, and the timeout bound comes along for free. Using the production probe as the test's instrument is safe here because the test asserts both polarities on it, so a broken `canConnect` fails one of them instead of quietly agreeing with itself.

### Why `cli.test.ts:231` is left unchanged

The lane also scoped `keeps a detached config handoff outside repo env overrides`. I measured it before changing anything, and there is no test-side cost to remove:

| Phase of the test body | Cost |
|---|---|
| `spawnSync` of the CLI | 711.2 ms |
| everything else combined | ~21 ms |

Choosing a shallower exit does not help either — `archon --help`, which exits before any command module loads, already costs 670 ms against 751 ms for this test's path. The floor is `packages/cli/src/cli.ts`'s static import graph: every command module plus `@archon/providers`, `@archon/core`, `@archon/workflows` and `@archon/git`, all loaded before argv is parsed.

Converting the test to the probe shape its sibling uses would cut the spawn to ~60 ms but delete the only coverage of `cli.ts`'s real boot ordering — the "must be the very first import" sequence the test exists to protect. That is a bad trade. The real fix is lazy command loading in `cli.ts`, which is a production change and does not belong in a test-stabilisation PR. Full measurement is recorded on the issue: https://github.com/coleam00/Archon/issues/2859#issuecomment-5446581856

## Validation

- `bun test ./scripts/migrate-state-dir.test.ts` ×5 — 19 pass / 0 fail every run — proves the fixture change did not alter any outcome in the file.
- `bun test src/utils/detached-run-control.integration.spec.ts` ×5 — 5 pass / 0 fail, 2.13 s every run — proves the imported probe behaves as the local copy did.
- `bun test src/cli.test.ts` ×5 — 95 pass / 0 fail every run — the unchanged test still passes.
- `bun run validate` — passed, exit 0, zero failing tests, zero type errors.
- **Mutation, C5 refusal guard:** disabling the two-candidate refusal in `migrate-state-dir.ts` turns `refuses when BOTH the subdirectory and the project root hold legacy state` red. The test still discriminates.
- **Mutation, the climb itself:** removing `findCodebaseByPathPrefix` from `resolveTarget` turns both C5 tests red.
- **Mutation, the new fixture:** skipping the `INSERT` turns both C5 tests red, so the in-process registration is load-bearing and not accidentally passing.
- **Dialect leak, red then green:** with `DATABASE_URL` set to a dummy Postgres DSN, both C5 tests failed before `childEnv` (exit 1 where 0 and 2 were expected — the `resolveTarget` catch, which proves the child took the Postgres branch). After it the whole file passes 19/19 in all three environments: ambient variable set, a repo-root `.env` defining it with no ambient variable, and both at once.
- **The `.env` path specifically:** with a repo-root `.env` and no ambient variable, both C5 tests fail under a `delete`-based strip and pass under the forced empty string — which is why the empty string is what shipped.
- **The Postgres opt-in is still live:** dropping `...extraEnv` from `childEnv` fails exactly the "a Postgres registry is never inferred from the local filesystem" test (18 pass / 1 fail).
- **Logger restore:** a temporary probe asserted after the C5 cases reports `getLogLevel()` as `info` with the restore in place and `silent` without it. The `scripts/` leg emits zero log lines either way.
- **Cost removed:** `new SqliteAdapter` + `INSERT` + `close` measures 3-6 ms across three consecutive samples, replacing a whole cold process.
- **Windows, the measurement this PR exists for:** two `windows-latest` jobs on this branch got past the `@archon/workflows` rotation and ran the scripts leg. Four C5 observations: 625 ms / 703 ms ([job 98718687827](https://github.com/coleam00/Archon/actions/runs/33130513243/job/98718687827)) and 1031 ms / 1016 ms ([job 98722467747](https://github.com/coleam00/Archon/actions/runs/33131701283/job/98722467747)), against the 5000 ms budget these two tests were timing out on at ~5015 ms. Mean 844 ms versus the 1160 ms mean the coverage spike measured for the same pair — about a 27% reduction, not the ~50% a naive "two processes became one" model suggests, because the surviving migrate-script spawn is the dominant cost.
- **Not verified:** the tail, which is what actually produces the timeouts. The coverage spike measured a 14.3% rate with max 4219 ms on a mean of 1160 ms; four observations cannot distinguish "the max moved down proportionally" from "we got four ordinary runs." What is established is that the accidental cost is gone and the pair now runs at roughly the same order as the file's other tests. A third re-run for more samples lost the coin flip to the `@archon/workflows` rotation (#2882) again.

## Links

- Related #2575
- Related #2859
- Related #2306
- Blocked from Windows verification by #2882

#2575 is deliberately still `Related` rather than `Closes`: the Windows measurement above is exactly what it asked for, but it is one sample of a stochastic failure. Recommend closing it once #2883 lands and a couple of Windows runs go green in a row. #2859 keeps its remaining instance open: `cli.test.ts`'s detached-config test is unchanged here, and closing it means reducing CLI startup cost, not reshaping the test.

### DIFF
--- a/packages/cli/src/utils/detached-run-control.integration.spec.ts
+++ b/packages/cli/src/utils/detached-run-control.integration.spec.ts
@@ -6,6 +6,7 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { trackTempRoots } from '@archon/paths/test-utils';
 import {
+  DETACHED_RUN_IPC_TIMEOUT_MS,
   detachedRunControlPath,
   requestDetachedRunStop,
   startDetachedRunControlServer,
@@ -45,6 +46,12 @@ function waitForExit(
 /**
  * Ask the endpoint itself whether it is still serving. Connection refusal is the
  * owner's own answer, so this reports shutdown progress without timing it.
+ *
+ * The bound is the same one production `canConnect` puts on its own probe: a
+ * connect that neither succeeds nor errors would otherwise hang until the test
+ * budget expired, which reads as an unrelated timeout instead of an unreachable
+ * endpoint. It is a backstop, not a measurement — on both the reachable and the
+ * refused path the socket settles immediately.
  */
 function endpointIsReachable(path: string): Promise<boolean> {
   return new Promise<boolean>((resolve: (reachable: boolean) => void): void => {
@@ -53,6 +60,9 @@ function endpointIsReachable(path: string): Promise<boolean> {
       probe.destroy();
       resolve(reachable);
     };
+    probe.setTimeout(DETACHED_RUN_IPC_TIMEOUT_MS, (): void => {
+      settle(false);
+    });
     probe.once('connect', (): void => {
       settle(true);
     });
--- a/packages/cli/src/utils/detached-run-control.integration.spec.ts
+++ b/packages/cli/src/utils/detached-run-control.integration.spec.ts
@@ -6,7 +6,7 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { trackTempRoots } from '@archon/paths/test-utils';
 import {
-  DETACHED_RUN_IPC_TIMEOUT_MS,
+  canConnect,
   detachedRunControlPath,
   requestDetachedRunStop,
   startDetachedRunControlServer,
@@ -39,35 +39,6 @@ function waitForExit(
     child.once('error', error => {
       clearTimeout(timer);
       reject(error);
-    });
-  });
-}
-
-/**
- * Ask the endpoint itself whether it is still serving. Connection refusal is the
- * owner's own answer, so this reports shutdown progress without timing it.
- *
- * The bound is the same one production `canConnect` puts on its own probe: a
- * connect that neither succeeds nor errors would otherwise hang until the test
- * budget expired, which reads as an unrelated timeout instead of an unreachable
- * endpoint. It is a backstop, not a measurement — on both the reachable and the
- * refused path the socket settles immediately.
- */
-function endpointIsReachable(path: string): Promise<boolean> {
-  return new Promise<boolean>((resolve: (reachable: boolean) => void): void => {
-    const probe = createConnection(path);
-    const settle = (reachable: boolean): void => {
-      probe.destroy();
-      resolve(reachable);
-    };
-    probe.setTimeout(DETACHED_RUN_IPC_TIMEOUT_MS, (): void => {
-      settle(false);
-    });
-    probe.once('connect', (): void => {
-      settle(true);
-    });
-    probe.once('error', (): void => {
-      settle(false);
     });
   });
 }
@@ -197,13 +168,15 @@ describe('detached run control integration', () => {
     const closing = owner.close();
     // close() is parked on the retained lease and has gone no further: it stops
     // the server and unlinks the endpoint only afterwards, so a close() that
-    // ignored the lease would already have made this probe unreachable.
-    expect(await endpointIsReachable(endpointPath)).toBe(true);
+    // ignored the lease would already have made this probe unreachable. Both
+    // polarities are asserted, so a broken probe fails one of them rather than
+    // quietly agreeing with itself.
+    expect(await canConnect(endpointPath)).toBe(true);
     expect(owner.isStopRequested()).toBe(true);
 
     await closing;
     expect(owner.isStopRequested()).toBe(false);
-    expect(await endpointIsReachable(endpointPath)).toBe(false);
+    expect(await canConnect(endpointPath)).toBe(false);
     client.destroy();
   });
 
--- a/packages/cli/src/utils/detached-run-control.ts
+++ b/packages/cli/src/utils/detached-run-control.ts
@@ -107,7 +107,15 @@ function listen(server: Server, path: string): Promise<void> {
   });
 }
 
-function canConnect(path: string): Promise<boolean> {
+/**
+ * True when something is listening on `path`. Connection refusal is the endpoint's
+ * own answer, so this reports liveness without timing it; the timeout is a backstop
+ * for a connect that neither succeeds nor errors.
+ *
+ * Exported for testing: the integration spec asks the same question of a shutting-down
+ * endpoint, and a second copy of this probe would be free to drift from this one.
+ */
+export function canConnect(path: string): Promise<boolean> {
   return new Promise(resolve => {
     const socket = createConnection(path);
     const settle = (connected: boolean): void => {
--- a/scripts/migrate-state-dir.test.ts
+++ b/scripts/migrate-state-dir.test.ts
@@ -105,24 +105,33 @@ async function collect(proc: Bun.Subprocess<'ignore', 'pipe', 'pipe'>): Promise<
 /**
  * Environment for every child, pinned to this sandbox.
  *
- * `DATABASE_URL` is REMOVED, not just overridden. A contributor running the
- * documented Postgres mode (`.env.example`) would otherwise have it inherited
- * here, which flips `getDatabaseType()` inside the child: `registryIsKnownEmpty`
- * goes false and `resolveTarget` reads a Postgres registry, while the C5 fixture
- * writes its row to this sandbox's SQLite file. The two sides would disagree and
- * both C5 tests would fail looking like a climbing regression rather than an
- * environment leak. Stripping it makes the dialect deterministic the same way
- * `ARCHON_HOME` already pins the destination — and it also keeps the fixture from
- * writing test rows into a contributor's real database, which the old subprocess
- * did. The one test that wants the Postgres branch passes it back via `extraEnv`.
+ * `DATABASE_URL` is forced EMPTY rather than deleted. A contributor running the
+ * documented Postgres mode would otherwise have it reach the child, which flips
+ * `getDatabaseType()` there: `registryIsKnownEmpty` goes false and `resolveTarget`
+ * reads a Postgres registry, while the C5 fixture writes its row to this sandbox's
+ * SQLite file. The two sides would disagree and both C5 tests would fail looking
+ * like a climbing regression rather than an environment leak.
+ *
+ * Deleting the key does NOT achieve that: a spawned Bun child re-runs automatic
+ * `.env` loading from its own cwd and fills in every key the passed environment
+ * omits, so a repo-root `.env` — exactly how Postgres mode is configured — puts it
+ * straight back. An empty string is a key that is present, which dotenv leaves
+ * alone, and `getDatabaseType()` reads it as falsy. This is the same suppression
+ * `packages/cli/src/cli.test.ts` uses for its own inherited keys.
+ *
+ * The one test that wants the Postgres branch passes a real DSN via `extraEnv`.
  */
 function childEnv(
   ctx: Sandbox,
   extraEnv: Record<string, string> = {}
 ): Record<string, string | undefined> {
-  const ambient: Record<string, string | undefined> = { ...process.env };
-  delete ambient.DATABASE_URL;
-  return { ...ambient, ARCHON_HOME: ctx.archonHome, LOG_LEVEL: 'silent', ...extraEnv };
+  return {
+    ...process.env,
+    ARCHON_HOME: ctx.archonHome,
+    LOG_LEVEL: 'silent',
+    DATABASE_URL: '',
+    ...extraEnv,
+  };
 }
 
 /** Run with raw argv — no implicit `--cwd`, for argument-parsing cases. */
--- a/scripts/migrate-state-dir.test.ts
+++ b/scripts/migrate-state-dir.test.ts
@@ -32,15 +32,28 @@
  * A larger budget would answer the question by silencing it.
  * If it does recur, find the specific colliding bound the way #2473 did; do not
  * reach for the timeout.
+ *
+ * It did recur, for the two C5 cases only (#2575). Coverage instrumentation was
+ * measured on windows-latest and ruled out. What was left was accidental cost:
+ * their fixture spawned a SECOND cold `bun` process purely to write one registry
+ * row, which is now an in-process write. The budget still has not moved.
  */
 import { describe, test, expect } from 'bun:test';
 import { mkdtemp, mkdir, writeFile, readFile, readdir } from 'fs/promises';
 import { removeTempTree } from '@archon/paths/test-utils';
+import { setLogLevel } from '@archon/paths';
+import { SqliteAdapter } from '@archon/core/db/adapters/sqlite';
 import { tmpdir } from 'os';
 import { join, resolve } from 'path';
 
+// The C5 fixture opens a real adapter in THIS process, which announces its
+// schema init at info. Every subprocess here is already run with LOG_LEVEL=silent;
+// this is the same suppression for the one piece of work that is no longer a
+// subprocess. Children inherit the level at creation, so it has to happen before
+// the first adapter exists.
+setLogLevel('silent');
+
 const SCRIPT = resolve(import.meta.dir, 'migrate-state-dir.ts');
-const REPO_ROOT = resolve(import.meta.dir, '..');
 
 /** Per-test paths. Immutable, and never shared between tests. */
 interface Sandbox {
@@ -418,10 +431,19 @@ describe('migrate-state-dir', () => {
      * silently disable project resolution: here a database exists, so the lookup
      * must still run and still climb.
      *
-     * Registered in a SUBPROCESS: @archon/core's DB connection is a module-level
-     * singleton, so registering in-process would cache a handle to the first
-     * test's temp ARCHON_HOME and fail with SQLITE_IOERR_VNODE once that
-     * directory is torn down.
+     * The row goes in through the SAME adapter the script's lookup opens, so the
+     * fixture cannot drift from the real schema — but through a dedicated instance
+     * rather than `@archon/core/db/codebases`, whose `pool` resolves the
+     * module-level connection singleton from `ARCHON_HOME`. That singleton would
+     * pin the first test's temp home and then fail with SQLITE_IOERR_VNODE once
+     * the directory is torn down; an instance takes an explicit path and is closed
+     * before the sandbox goes away. `name` and `default_cwd` are the only columns
+     * the destination resolution reads (`resolveRepoProjectIdentity`).
+     *
+     * This used to be a cold `bun -e` subprocess importing @archon/core's whole
+     * module graph to write that one row — the accidental half of these tests'
+     * Windows cost (#2575). The subprocess below it runs the migration script
+     * itself and is the contract under test, so it stays a real process.
      */
     async function withProjectSandbox(body: (ctx: ProjectSandbox) => Promise<void>): Promise<void> {
       return withSandbox(async base => {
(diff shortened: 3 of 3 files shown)


Wirasm:
The `windows-latest` failure on this PR is not from this change, and it also means the fix could not be measured there.

It failed on `resolve-review-mode > distinguishes full, continuation, and broken continuation inputs` in `@archon/workflows` — a package this PR does not touch (the diff is `scripts/migrate-state-dir.test.ts` and `packages/cli/src/utils/detached-run-control.integration.spec.ts`). I re-ran the job expecting a flake and it reproduced exactly: 5015.00 ms, then 5016.00 ms.

`dev` is red on Windows for the same reason, at a different test in that package each run, including at `59f71e92` — the commit this branch forked from:

| Run | Head | Failing test | Reported |
|---|---|---|---|
| [33124733645](https://github.com/coleam00/Archon/actions/runs/33124733645) | `59f71e92` (current dev) | `runFixtures > passes a fixture whose declared outcome matches the dry-run outcome` | 5016.00 ms |
| [33123519845](https://github.com/coleam00/Archon/actions/runs/33123519845) | `6287179e` | `a custom commands.folder is captured and resolves from the capture` | 6172.00 ms |
| [33127863828](https://github.com/coleam00/Archon/actions/runs/33127863828) | `222ebce1` (this PR) | `resolve-review-mode > distinguishes full, continuation, and broken continuation inputs` | 5015 / 5016 ms |

Because the root command is `bun --filter '*' --parallel test && bun test ./scripts/`, the packages-leg failure short-circuits the scripts leg. All three Windows logs contain zero occurrences of `migrate-state-dir`, so the two C5 tests this PR lightens never executed on Windows. Ubuntu, docker-build and schema-upgrade are green here.

Details and the per-suite framing are on the issue: https://github.com/coleam00/Archon/issues/2859#issuecomment-5446724756

Wirasm:
## Needs fixes

This is a correction-verification pass over `222ebce1..eb7b5250`, bounded to the three prior findings. R2 and R3 are genuinely fixed. R1 is real progress but not closed: `childEnv()` deletes `DATABASE_URL` from the object passed to `Bun.spawn`, which correctly blocks an ambient shell-exported `DATABASE_URL` and — a genuine, separate win — stops the fixture from ever writing test rows into a contributor's real Postgres (the old subprocess did this on every C5 run). But the migration-script child is still spawned without a `cwd` override for the `runIn`/`runMigration`/`runWithEnv` path the C5 tests use, so it inherits the test runner's own cwd — the real repository root. Bun's own automatic `.env` loading in *that child process* re-populates any environment key absent from the object explicitly passed to `Bun.spawn`, sourced from the child's own cwd. A contributor who configured Postgres mode the documented way — a `.env` file at the repo root (`.env.example`'s instructions) rather than a shell export — still hits the exact original bug. I reproduced this directly on `eb7b5250`: a temporary `.env` with an unreachable `DATABASE_URL`, nothing else in the environment, and both C5 tests fail with the same exit-code mismatch the original finding described. Reverted before finishing this review; the checkout is clean.

Also worth surfacing: Windows CI is green on this head, and for the first time the C5 pair actually executed there (the `@archon/workflows` flake that blocked the scripts leg on `222ebce1` didn't trigger this run) — both tests passed well under the 5000ms budget, which is the first real Windows evidence for the underlying #2575 fix.

**1 blocking · 2 non-blocking**

**Validation:** Focused tests, lint, and type-check pass on `eb7b5250`. CI is green across all five checks, including `windows-latest` — and the scripts leg ran this time, giving direct Windows evidence for the C5 fix. See Validation detail for the R1 reproduction.

### Findings

|---|---|---|---|
| `R1` | Important | C5 fixture write/read dialect asymmetry — `childEnv()`'s env-object deletion doesn't survive the child's own `.env` reload | OPEN |
| `R2` | Suggestion | Test's `endpointIsReachable` hand-copied production `canConnect` instead of importing it | FIXED |
| `R3` | Suggestion | `setLogLevel('silent')` at module scope mutated the shared root logger with no restore | FIXED |

<details>
<summary><code>R1</code> — C5 fixture write/read dialect asymmetry (partially addressed)</summary>

**Impact:** unchanged from the prior review — a contributor with Postgres mode configured sees both C5 tests fail with a destination mismatch that reads like a climbing-logic regression, not an environment leak.

**What the correction fixed:** `childEnv()` (`scripts/migrate-state-dir.test.ts:119-126`) copies `process.env`, deletes `DATABASE_URL`, and merges in `ARCHON_HOME`/`LOG_LEVEL`/`extraEnv`. This genuinely blocks an *ambient, shell-exported* `DATABASE_URL` from reaching any child in the file, and — independent of the original finding — it stops the fixture's own write from ever touching a real Postgres, since the write is now purely in-process SQLite regardless of environment. That was a real, if separate, contributor-environment-pollution risk in the old subprocess-based fixture, and it's gone for good.

**What it did not fix:** `runWithEnv` (`scripts/migrate-state-dir.test.ts:150-163`, used by `runIn`/`runMigration`, which the C5 tests call) spawns the migration-script child with no `cwd` override, so the child inherits the test runner's own working directory — the real repository root when the suite is run the normal way (`bun test ./scripts/`, `bun run validate`, CI). Bun's own automatic `.env`-file loading happens *inside that child process*, independent of whatever `env` object the parent explicitly passed to `Bun.spawn`: it fills in any key **absent** from the passed env, reading from the child's own resolved cwd. Deleting a key from the parent's env object is not the same as making that key unavailable to the child — the child looks for its own `.env` and finds one if the repo root has one.

I proved this directly, in this checkout, on `eb7b5250`:
```
echo 'DATABASE_URL=postgresql://u@/db?host=/tmp/nonexistent-socket' > .env
bun test ./scripts/migrate-state-dir.test.ts -t "C5"
# → both C5 tests fail (exit 1 where 0/2 were expected)
```
and traced it to the exact root cause with a minimal, fully-scrubbed reproduction (`env -i` — no inherited variables at all, not even `DATABASE_URL`) that still produced `Could not read the codebase registry: connect ENOENT /tmp/nonexistent-socket/.s.PGSQL.5432` — proof the value came from the child's own `.env` reload, not from anything the parent passed. `.env` and the temp file were removed before continuing; `git status` is clean.

The Postgres-branch opt-in test (`scripts/migrate-state-dir.test.ts:314-342`) itself is unaffected by this gap and still exercises correctly — I confirmed dropping `...extraEnv` from `childEnv` in `runWithEnv` fails exactly that one test (18 pass / 1 fail), matching the correction's own claim.

The implementer's validation record shows they proved the ambient-export case red-then-green but explicitly declined to run the reachable-Postgres case pre-fix, to avoid the pre-fix fixture writing test rows into a real database — a sound safety call, but it also meant the `.env`-file vector was never exercised, so this gap went unnoticed. CI does not set `DATABASE_URL` or create a `.env`, so this remains invisible there — a local-Postgres-mode-only correctness gap, unchanged in reach from the original finding.

**Evidence:** `scripts/migrate-state-dir.test.ts:119-126` (`childEnv`), `scripts/migrate-state-dir.test.ts:150-163` (`runWithEnv`, no `cwd` override), direct reproduction above.

**Required outcome:** make the child's own `.env` reload agree with the sandbox instead of just deleting the key from the passed env object. Either give `runWithEnv`'s `Bun.spawn` call an explicit `cwd` pointing at a directory with no `.env` (mirroring what `runRaw` already does with `cwd: ctx.repo`), or set `DATABASE_URL` to an explicit empty string in `childEnv`'s return value rather than deleting the key — an explicitly-present (even empty) key is not something Bun's dotenv loader overrides, and `getDatabaseType()`'s `process.env.DATABASE_URL ?` check treats `''` as falsy, so it correctly resolves to `'sqlite'`. I verified both alternatives work against a synthetic repro of the same shape before writing this up.

**Found by:** `prp-core:seam-analyzer` (original), verified and the correction re-tested by this review

**Disposition:** OPEN — the changed head does not prove the invariant the PR itself states ("the sandbox must resolve the same dialect on the write and the read side"); reproduced directly against `eb7b5250`.

</details>

<details>
<summary><code>R2</code> — Duplicated socket-probe implementation instead of an import</summary>

**Impact:** none remaining — resolved.

**Verification:** `canConnect` is now `export function canConnect` in `packages/cli/src/utils/detached-run-control.ts:118`, with an "Exported for testing" docblock matching the existing convention (`calculatePortOffset` in `packages/core/src/utils/port-allocation.ts:18`, confirmed to exist as cited). The spec (`packages/cli/src/utils/detached-run-control.integration.spec.ts`) now imports it and the duplicate helper plus the now-unused `DETACHED_RUN_IPC_TIMEOUT_MS` import are both deleted — a net removal (29 lines removed from the spec, 10 added to production for the export and its docblock). Production call sites (`detached-run-control.ts:151,169,175`) are unchanged; adding `export` to an existing function changes no behavior, confirmed by diff. The masking concern from the original finding is addressed structurally, not just asserted: the test now checks both polarities of the same imported probe (`toBe(true)` immediately after close is requested but before it completes, `toBe(false)` after it completes) — a broken `canConnect` fails one of those two assertions rather than silently agreeing with a broken assumption. Reran the spec 5x on `eb7b5250`: 5 pass / 0 fail every time.

**Disposition:** FIXED — verified by diff and by re-running the spec.

</details>

<details>
<summary><code>R3</code> — Unscoped global logger mutation with no restore</summary>

**Impact:** none remaining — resolved.

**Verification:** the module-scope `setLogLevel('silent')` call is gone; `scripts/migrate-state-dir.test.ts` now imports `getLogLevel` alongside `setLogLevel` and scopes the mutation inside `withProjectSandbox` (lines 478-479, 488): `priorLogLevel` is captured before the adapter opens, `setLogLevel('silent')` is applied just before it, and `setLogLevel(priorLogLevel)` restores it in the same `finally` that closes the adapter — symmetric with the adapter's own open/close lifecycle. Grepped the whole file for any other `setLogLevel`/`getLogLevel` occurrence: none remain outside this one scoped block. Ran the full `bun test ./scripts/` (all 5 files, 101 tests) on `eb7b5250`: 101 pass / 0 fail, confirming the restore prevents the cross-file leakage the original finding described.

**Disposition:** FIXED — verified by diff, by grep for remaining occurrences, and by running the full multi-file suite.

</details>

<details>

#### Reviewer coverage

Correction-verification pass; no fresh dispatch to the four scopes. Findings and dispositions above cover the entire bounded diff (`222ebce1..eb7b5250`).

#### Validation

|---|---|---|
| `git diff dev...HEAD --stat` | Confirmed bounded | 3 files: `scripts/migrate-state-dir.test.ts`, `packages/cli/src/utils/detached-run-control.ts`, `packages/cli/src/utils/detached-run-control.integration.spec.ts` — nothing beyond the R1/R2/R3 corrections |
| `bun test ./scripts/migrate-state-dir.test.ts` (×3) | PASS | 19/19 every run, ~1.2-1.3s |
| R1 direct reproduction (`.env` with unreachable `DATABASE_URL`, `-t "C5"`) | RED as originally described | 0 pass / 2 fail — reproduced, then reverted; checkout clean afterward |
| R1 minimal root-cause proof (`env -i`, no inherited env at all) | Confirms child re-derives `DATABASE_URL` from its own `.env`, not from anything passed | `Could not read the codebase registry: connect ENOENT .../.s.PGSQL.5432` |
| R1 opt-in spot-check (drop `...extraEnv` from `childEnv` in `runWithEnv`) | Fails exactly the intended test | 18 pass / 1 fail — "a Postgres registry is never inferred from the local filesystem" |
| `bun test ./scripts/` (all 5 files) | PASS | 101/101 (after `bun install` in this fresh worktree — an unrelated missing-dependency artifact from skipping install initially, not a PR defect) |
| `bun test src/utils/detached-run-control.integration.spec.ts` (packages/cli, ×5) | PASS | 5/5 every run, 2.13-2.14s |
| `grep -rn canConnect packages/cli/src` | Confirmed no behavior change | export added, all 3 production call sites unchanged |
| `bun run lint` | PASS | all 12 lint targets clean |
| `bun run type-check` | PASS | all 13 packages exit 0 |
| CI `test (ubuntu-latest)` | PASS | run 33130513243 |
| CI `docker-build` | PASS | run 33130513243 |
| CI `schema-upgrade` | PASS | run 33130513243 |
| CI `CodeRabbit` | PASS | review completed |
| CI `test (windows-latest)` | PASS | run 33130513243 — and for the first time the scripts leg actually executed: `resolve-review-mode` passed at 859ms (the #2882 flake didn't trigger this run), so both C5 tests ran and passed at 625ms/703ms, well under the 5000ms Bun default. This is the first direct Windows evidence for the #2575 fix; it does not depend on or validate R1's fix, which is a separate, environment-only gap this Windows job's clean `.env`-less runner cannot expose either way. |
| PR body | Confirmed accurate | now states "The only production change is exporting `canConnect`... it adds no behavior," matching the diff; still no closing keywords, with the same stated reasons |

</details>

Wirasm:
## R1 round 2 was already shipped as `9edbaa5f` — plus a correction to the fix menu

The review's round-2 findings landed after I had pushed the fix, so `9edbaa5f` already implements shape **(b)**: `childEnv` sets `DATABASE_URL: ''` rather than deleting the key. Re-verified on the recreated worktree with the exact recipe requested — temp `.env` at the repo root holding an unreachable DSN, no ambient export:

| State | Result |
|---|---|
| `eb7b5250` (the `delete`) | 0 pass / 2 fail — both C5 tests red, same shape as the original bug |
| `9edbaa5f` (forced `''`) | 19 pass / 0 fail |
| `9edbaa5f`, `.env` **and** ambient export | 19 pass / 0 fail |
| Postgres opt-in through `extraEnv` | still passes, and still dies (18/1) when `...extraEnv` is dropped |

Temp `.env` removed afterwards; `git status --porcelain` is empty, tracked and untracked.

### Correction: shape (a) is not a standalone fix, and there are three leak paths, not one

I tested (a) — explicit `cwd` on `runWithEnv`'s spawn, no `DATABASE_URL` handling — against both vectors. It fails **both**:

```
(a) alone, repo-root .env only, no ambient export : 0 pass / 2 fail
(a) alone, ambient shell export                   : 0 pass / 2 fail
```

The reason is that the value reaches the child three different ways, and each candidate fix closes a different subset:

1. **Ambient shell export** → sits in the parent's `process.env` → carried in by `...process.env`.
2. **Repo-root `.env` loaded by the PARENT.** `bun test ./scripts/` runs from the repo root, so the *test process itself* has already loaded `.env` before it spawns anything. Direct check: `bun -e 'console.log(process.env.DATABASE_URL)'` from the repo root prints the `.env` value, and prints `undefined` once the file is moved away. This path is the one that makes (a) insufficient — the child's cwd is irrelevant when the parent hands the value over explicitly.
3. **Repo-root `.env` reloaded by the CHILD** from its own resolved cwd, filling keys absent from the passed environment. This is the path the review identified, and the one that made `delete` insufficient.

So: `delete` closes 1 and 2 and leaves 3. An explicit `cwd` closes 3 and leaves 1 and 2. `DATABASE_URL: ''` closes all three — it is a later key in the object literal, so it overrides the spread, and it is a *present* key, so the child's dotenv loader will not replace it. `getDatabaseType()` reads `''` as falsy and selects SQLite.

That also corrects my own earlier framing on this PR, which attributed the whole leak to the child-side reload. The child-side reload is real, but it is not the only path, and it is not the one that defeats an explicit `cwd`.

Worth carrying to any other lane offered shape (a): it only helps in combination with something that also strips the parent's copy, and the empty-string form makes that combination unnecessary.

Issue #2882, opened 2026-08-28
test(workflows): the @archon/workflows suite runs at the Windows budget edge — a different test dies at ~5000ms each run


Wirasm:
## Problem

windows-latest is failing `@archon/workflows` on dev with a DIFFERENT test each run, all at the 5000ms default budget, while ubuntu stays green:

- dev `59f71e92`: `runFixtures > passes a fixture whose declared outcome matches the dry-run outcome` [5016ms]
- dev `6287179e`: `a custom commands.folder is captured and resolves from the capture` [6172ms]
- PR #2881 head `222ebce1`: `resolve-review-mode > distinguishes full, continuation, and broken continuation inputs` [5015ms; 5016ms on rerun — reproduced]

This is not a per-test defect: the package's subprocess-heavy tests run close enough to the budget on Windows (10 package suites competing under `bun --filter '*' --parallel`) that scheduling decides which one dies. Evidence thread: https://github.com/coleam00/Archon/issues/2859#issuecomment-5446666467 and the follow-up comment.

Known heavy member: `resolve-review-mode` (packages/workflows/src/defaults/review-mode-script.test.ts, added with #2863) makes THREE cold Python/uv interpreter starts inside one test.

## Second-order cost

The root test script is `bun --filter '*' --parallel test && bun test ./scripts/` — a packages-leg failure short-circuits the scripts leg, so every one of these failures also silently skips all scripts/ tests on Windows (zero `migrate-state-dir` occurrences in all three logs). Historical Windows rates for scripts-leg tests are understated, and scripts-leg fixes cannot be verified.

## Desired outcome

The suite stops living at the budget edge by REMOVING COST, never by raising budgets (standing constraint: no timer/timeout/budget increases):

1. Batch or in-process the accidental subprocess cost in the named heavy tests — e.g. resolve-review-mode's three interpreter starts become one (parameterize a single invocation), following the pattern that fixed C5 (in-process setup, real subprocess only where the subprocess IS the subject).
2. Sweep packages/workflows for other tests spawning multiple subprocesses or interpreters where one (or none) would prove the same behavior.
3. Windows CI on dev returns to green consistently, and the scripts leg demonstrably executes on Windows again.

## Invariant

A test spawns a subprocess only when the subprocess is the behavior under test, and at most the minimum number that proves it.

## Evidence pointers

[...]

PR #2883, opened 2026-08-28
test(workflows): take the Windows suite off the 5000ms budget edge

Wirasm:
## Problem and outcome

`@archon/workflows` has been failing on `windows-latest` with a **different test each run**, all at Bun's 5000 ms default budget, while ubuntu stays green ([rotation evidence](https://github.com/coleam00/Archon/issues/2859#issuecomment-5446724756)). That is not three per-test defects: the package's filesystem- and process-heavy tests run close enough to the budget that scheduling decides which one dies. Because the root script is `bun --filter '*' --parallel test && bun test ./scripts/`, each of those failures also short-circuits the scripts leg, so no `scripts/` test has executed on Windows in any of those runs.

- **Outcome:** the three named rotation members, and the suite around them, do materially less work per test. Measured on macOS, worst-of-5 per-test cost drops 91.8 ms → 24.2 ms (`resolve-review-mode`), 112.2 ms → 11.7 ms (`workflow-source`), 279.9 ms → 160.4 ms (`fixture-runner`); package wall time 14.1 s → 11.2 s.
- **Invariant:** a test spawns a subprocess only where the subprocess is the behavior under test, and at most the minimum number that proves it. Every behavior proven before is still proven.
- **Scope boundary:** no timer, timeout, or budget was raised anywhere — cost was only removed. No production code changed; the diff is four test files.

## Review guidance

- **Feedback requested:** correctness of the two cost levers, specifically whether the empty-bundled-tree redirect weakens anything the capture tests claim to prove.
- **Start here:** `packages/workflows/src/workflow-source.test.ts:563` — the new `scopes).toContain('bundled')` assertion is what makes the biggest lever safe.
- **Review order:** `workflow-source.test.ts` and `fixture-runner.test.ts` (the capture lever), then `review-mode-script.test.ts` (the interpreter split), then `dag-executor.test.ts` (mechanical).
- **Lower-attention areas:** the `git config` → `-c` change in `dag-executor.test.ts` and `fixture-runner.test.ts` is mechanical and behavior-identical; it is the pattern `generate-bundled-defaults.test.ts:62` already uses.
- **Known risk or uncertainty:** `mock.module` is process-global, and `fixture-runner.test.ts` shares a `bun test` invocation with `state-migration.test.ts`, which installs its own `@archon/paths` mock. Verified they coexist: `bun test src/state-migration.test.ts src/dry-run.test.ts src/fixture-runner.test.ts` is 111/111, and that file's non-vacuous `expect(warnCalls).toHaveLength(1)` still passes in-group, so its logger mock is not being clobbered.

## Solution

Three sources of accidental cost, each removed at its own boundary.

**Three interpreter starts inside one budget.** `resolve-review-mode` drove `python`/`python3` three times inside a single test. The script's *process* contract — stdout JSON, exit status, stderr message — genuinely is the subject, so the starts stay; each case now gets its own test and therefore its own budget. Moving them into a shared `beforeAll` was rejected on purpose: that puts them back under one deadline, which is #2860's lesson. The fixture also drops a `mkdtemp` plus recursive removal for a single file write, since the script only stats the path it is handed.

**~9,700 incidental file copies.** Every `captureWorkflowSource` call copies *and digests* the repo's own bundled scope — `.archon/workflows` plus `.archon/commands`, ~178 files, 1.1 MB — on top of the handful of files the fixture wrote. `workflow-source.test.ts` takes 25 captures and `fixture-runner.test.ts` takes 30, and no test in either reads bundled content: the capture assertions count the *project* scope, and the bundled-*scope* fixture tests drive `sourceRoots.bundledWorkflows`, a discovery root already pointed at a temp directory. Both files now point the two bundle getters at an owned empty tree — the lever `subrun.test.ts:35-63` already documents for this exact Windows budget problem. The bundled scope keeps its semantics: an existing directory is still scanned, still copied, still recorded in the manifest (`scopes` remains `project,global,bundled`).

Because nothing previously asserted that, the lever could have silently degraded into "the bundled scope is no longer captured" with every other assertion still green. `workflow-source.test.ts:563` now pins it.

**Five git processes to make a throwaway repo.** Two helpers ran `git init`, two `git config` writes, `git add`, `git commit`. The identity now travels as `-c` overrides on the commit, which is what `generate-bundled-defaults.test.ts` already does — 3 processes instead of 5, across 6 tests in `dag-executor.test.ts` and 6 repos in `fixture-runner.test.ts`.

## Validation

- `bun run validate` — exit 0, zero failing test groups — proves the repository gates hold.
- `bun run lint` — clean.
- `cd packages/workflows && bun run test` — every group green (682, 111, 32, … 0 fail) — proves the package suite as CI invokes it.
- Each touched file run 5×, worst-of-5 per test recorded before and after (see Outcome). Package wall time measured 3× each way: 14.65 / 14.09 / 13.82 s → 11.30 / 11.20 / 11.26 s.
- Mutation evidence, each seen red: forcing `continuation` to `False`, to `True`, and removing the missing-file guard in `resolve-review-mode.py` each failed exactly one of the three new tests, and the right one. Removing the empty bundled tree failed the new `scopes).toContain('bundled')` assertion and only that one.
- Capture-volume evidence: an instrumented `captureWorkflowSource` counted 180 files per capture before and 0–6 after, across 55 captures in the two files.
- **Not verified:** Windows CI. macOS cannot prove the Windows outcome — this PR's `windows-latest` leg is the real evidence, and the issue only closes if it is green there and the scripts leg demonstrably runs.

## Delivery considerations

|---|---|---|
| Observability | If the bundled scope ever stops being captured, the capture tests now fail loudly instead of passing cheaply | `packages/workflows/src/workflow-source.test.ts:563`, seen red |

## Deliberately left unchanged

- `executor.test.ts` "records even when the platform cannot be written to" (~1000 ms, the largest single test in the package) — the cost is production retry backoff behind a rejecting platform, not test scaffolding. Removing it would mean changing a timer.
- `dag-executor.test.ts` "timeout kills subprocess" (~506 ms) and the `idle-timeout` tests — the delay is the subject in each.
- `generate-bundled-defaults.test.ts` — already amortized by an earlier pass to one `bun` spawn plus at most one `git` per test.
- `fixture-runner.test.ts` "gives clean and pre-dirtied callers the same verdict" still builds two repos; comparing a clean caller against a pre-dirtied one is the point of the test, so merging them would destroy what it proves.
- `script-node-deps.test.ts` and `runtime-check.test.ts` already mock `execFileAsync` and spawn nothing.

## Links

- Related #2882 (close after a few consecutive clean windows-latest runs on dev — its acceptance is consistency, which one green run cannot prove)
- Related #2859, #2575, #2860

## Summary by CodeRabbit

* **Tests**
  * Improved workflow and repository test reliability by isolating temporary files and bundled resources.
  * Added coverage for empty bundled workflow sources.
  * Simplified test setup and cleanup for faster, more consistent execution.

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -33084,12 +33084,17 @@ describe('executeDagWorkflow -- composed fan-out (include + fan_out, #2512)', ()
 describe('executeDagWorkflow -- node-level mutates_checkout: false (#2771)', () => {
   let testDir: string;
 
+  // Three git processes, not five: the identity travels as `-c` overrides on the commit
+  // rather than as two separate `git config` writes. Six tests here pay this cost, and
+  // process creation is the expensive part on Windows CI (#2882).
   const initRepo = async (dir: string): Promise<void> => {
     await git.execFileAsync('git', ['init', '-q'], { cwd: dir });
-    await git.execFileAsync('git', ['config', 'user.email', 't@t'], { cwd: dir });
-    await git.execFileAsync('git', ['config', 'user.name', 't'], { cwd: dir });
     await git.execFileAsync('git', ['add', '-A'], { cwd: dir });
-    await git.execFileAsync('git', ['commit', '-qm', 'init', '--allow-empty'], { cwd: dir });
+    await git.execFileAsync(
+      'git',
+      ['-c', 'user.email=t@t', '-c', 'user.name=t', 'commit', '-qm', 'init', '--allow-empty'],
+      { cwd: dir }
+    );
   };
 
   const runBashNode = async (
--- a/packages/workflows/src/defaults/review-mode-script.test.ts
+++ b/packages/workflows/src/defaults/review-mode-script.test.ts
@@ -1,5 +1,6 @@
-import { afterEach, describe, expect, it } from 'bun:test';
-import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { afterAll, describe, expect, it } from 'bun:test';
+import { randomUUID } from 'node:crypto';
+import { rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 
@@ -14,8 +15,18 @@ const SCRIPT_PATH = join(
   'resolve-review-mode.py'
 );
 const PYTHON_COMMAND = globalThis.process.platform === 'win32' ? 'python' : 'python3';
-const tempDirs: string[] = [];
+const tempFiles: string[] = [];
 
+/**
+ * Start the interpreter once, for one input.
+ *
+ * The script's PROCESS contract is the subject — stdout JSON, exit status, and the
+ * stderr message a failing node reports — so a real interpreter start is what proves
+ * it. Each case therefore gets its own test and its own start: three of them charged
+ * to a single test's budget is what timed out on Windows CI (#2882). Hoisting the
+ * starts into a shared `beforeAll` would put them back under one deadline (#2860), so
+ * the split is the fix, not a shared setup hook.
+ */
 async function runResolver(priorReport: string): Promise<{
   exitCode: number;
   stdout: string;
@@ -35,25 +46,37 @@ async function runResolver(priorReport: string): Promise<{
   return { exitCode, stdout, stderr };
 }
 
-afterEach(async () => {
-  await Promise.all(tempDirs.splice(0).map(path => rm(path, { recursive: true, force: true })));
+/** A path under the system temp dir that nothing creates. */
+function unusedPath(): string {
+  return join(tmpdir(), `archon-review-mode-${randomUUID()}.md`);
+}
+
+afterAll(async () => {
+  await Promise.all(tempFiles.splice(0).map(path => rm(path, { force: true })));
 });
 
 describe('resolve-review-mode', () => {
-  it('distinguishes full, continuation, and broken continuation inputs', async () => {
-    const dir = await mkdtemp(join(tmpdir(), 'archon-review-mode-'));
-    tempDirs.push(dir);
-    const existingReport = join(dir, 'report.md');
-    const missingReport = join(dir, 'missing.md');
-    await writeFile(existingReport, '# Prior review\n');
-
+  it('reports a full review when no prior report is named', async () => {
     const full = await runResolver('');
     expect(full.exitCode).toBe(0);
     expect(JSON.parse(full.stdout)).toEqual({ continuation: false });
+  });
+
+  it('reports a continuation when the named prior report exists', async () => {
+    // A single file, not a directory: the script only stats the path it is given, and a
+    // recursive temp-tree removal is the part of this fixture that is slow and flaky on
+    // Windows (#2306).
+    const existingReport = unusedPath();
+    tempFiles.push(existingReport);
+    await writeFile(existingReport, '# Prior review\n');
 
     const continuation = await runResolver(existingReport);
     expect(continuation.exitCode).toBe(0);
     expect(JSON.parse(continuation.stdout)).toEqual({ continuation: true });
+  });
+
+  it('fails when the named prior report does not exist', async () => {
+    const missingReport = unusedPath();
 
     const broken = await runResolver(missingReport);
     expect(broken.exitCode).toBe(1);
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -1,5 +1,5 @@
 /** Tests for the declared-data dry-run fixture runner (#2772). */
-import { describe, it, expect, afterEach } from 'bun:test';
+import { describe, it, expect, afterEach, afterAll, mock } from 'bun:test';
 import {
   existsSync,
   mkdirSync,
@@ -10,8 +10,35 @@ import {
   symlinkSync,
   writeFileSync,
 } from 'node:fs';
+import { mkdir, rm } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+
+// Capture-cost control, same lever and same reason as `subrun.test.ts` (#2882): every
+// `runFixtures` call takes one source capture, and a capture copies and digests the
+// repo's OWN bundled scope — `.archon/workflows` plus `.archon/commands`, ~178 files —
+// alongside the handful of fixture files the test wrote. Thirty captures in this file
+// is ~5,300 incidental file copies, and that bulk IO is what puts this suite at Bun's
+// 5000ms budget on a contended Windows runner. No test here reads bundled CONTENT: the
+// bundled SCOPE tests drive `sourceRoots.bundledWorkflows`, which is a discovery root
+// this file already points at a temp directory. Pointing the two bundle getters at an
+// owned EMPTY tree keeps the bundled scope's semantics intact — an existing directory
+// is still scanned, still copied, still recorded in the manifest — while removing the
+// file fan-out.
+// NB: point these one level DEEP (`<root>/defaults`) — captureWorkflowSource copies
+// dirname(getDefault*Path()), so the getter's PARENT must be the owned empty tree.
+const bundledDefaultsRoot = join(tmpdir(), `fixture-runner-test-empty-bundled-${process.pid}`);
+await mkdir(join(bundledDefaultsRoot, 'defaults'), { recursive: true });
+afterAll(async () => {
+  await rm(bundledDefaultsRoot, { recursive: true, force: true }).catch(() => {});
+});
+const realArchonPaths = await import('@archon/paths');
+mock.module('@archon/paths', () => ({
+  ...realArchonPaths,
+  getDefaultWorkflowsPath: () => join(bundledDefaultsRoot, 'defaults'),
+  getDefaultCommandsPath: () => join(bundledDefaultsRoot, 'defaults'),
+}));
+
 import { execFileAsync } from '@archon/git';
 import { parseWorkflow } from './loader';
 import { expandWorkflowIncludes } from './include-expander';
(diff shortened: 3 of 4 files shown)


Wirasm:
## Ready to merge

This PR removes accidental cost from four `@archon/workflows` test files to stop the suite from timing out at Bun's 5000ms default budget on windows-latest, without touching production code or raising any timer, timeout, or budget. Independent verification confirms every claim in the PR body: the bundled-tree redirect reuses `subrun.test.ts`'s existing lever and preserves the bundled scope's manifest semantics (verified by mutation — removing the redirect's directory creation fails exactly the new `toContain('bundled')` guard and nothing else); the three-way `review-mode-script.test.ts` split isolates each script mutation to its own test (verified by mutating all three code paths independently); the `git config` → `-c` fold matches an existing precedent and no test reads git identity state; no timer/budget value increased anywhere in the diff; and the full package suite, lint, and type-check are all green in this checkout. All three specialist reviewers found only non-blocking Suggestions — mostly missed opportunities to reuse existing test infrastructure (a shared temp-tree helper, a Windows-safe cleanup primitive) rather than anything incorrect.

**Validation:** `bun run test` (full `@archon/workflows` suite, all 26 groups) PASS, 0 fail · `bun x tsc --noEmit` PASS · `bun run lint` PASS · CI (`ubuntu-latest`, `windows-latest`, `docker-build`, `schema-upgrade`) all PASS on `97100871`.

**Caveat carried from the issue and the PR's own validation section:** issue #2882's acceptance criterion is "Windows CI on dev returns to green **consistently**." One green `windows-latest` run (7m57s, job `98717531915`) is strong evidence the fix works — it demonstrably removed the specific cost that caused three prior failures, and the scripts leg now runs (101 tests, previously 0) — but it cannot prove consistency the way a rotation of runs would. The PR body states this same limitation and does not claim otherwise. Recommend closing #2882 only after dev has run windows-latest a few more times clean, not automatically on merge.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Empty-bundled-tree mock setup is now duplicated verbatim across 3 files with no shared helper | OPEN |
| `R2` | Suggestion | New temp-tree cleanup bypasses the package's existing Windows-safe removal helper | OPEN |
| `R3` | Suggestion | New `toContain('bundled')` regression guard exists in only 1 of 3 files carrying the same risk | OPEN |
| `R4` | Suggestion | `tempFiles` array + `afterAll` registry is oversized for the one file it now manages | OPEN |
| `R5` | Suggestion | Three explanatory comments cite precedent loosely or overstate scope | OPEN |
| `R6` | Suggestion | `fixture-runner.test.ts`'s new `@archon/paths` mock relies on an undocumented Bun module-cache behavior to coexist safely with `state-migration.test.ts`'s mock in the same process | OPEN |

<details>
<summary><code>R1</code> — Third copy of the empty-bundled-tree setup block, no shared helper</summary>

**Impact:** No functional defect. Purely a maintenance-cost observation: a future change to this lever (e.g. a new getter to redirect, a different mock shape) now needs to be applied identically in three places by hand.

**Evidence:** `packages/workflows/src/subrun.test.ts:47-63` (pre-existing) is now matched almost line-for-line by `packages/workflows/src/workflow-source.test.ts:9-31` and `packages/workflows/src/fixture-runner.test.ts:16-39` — root-name computation, `mkdir`, `afterAll`/`rm`, dynamic `import`, and a `mock.module` factory returning `{...realArchonPaths, getDefaultWorkflowsPath, getDefaultCommandsPath}`. Two of the three occurrences land in this single PR.

**Required outcome:** Not required for this PR. Worth extracting into the package's existing `packages/workflows/src/test-utils.ts` (or `@archon/paths/test-utils`) the next time this lever is touched — `subrun.test.ts`'s factory additionally injects `createLogger` and telemetry mocks, so a shared helper would need to accept those as an optional overlay rather than assuming a bare two-getter shape.

**Disposition:** OPEN, non-blocking. Repository guidance ("duplicate small local logic... extract only after a pattern has repeated and stabilized") treats this as a judgment call at the 3rd occurrence; both agents agreed it's now past that bar but is low-risk today (three stable call sites, no near-term change expected).

</details>

<details>
<summary><code>R2</code> — New cleanup bypasses the existing Windows-safe temp-tree removal helper</summary>

**Impact:** If a handle stayed open on the redirect tree (unlikely — nothing writes into it after creation, only `captureWorkflowSource` reads from it), the bare `rm(...).catch(() => {})` would silently leave the directory on disk with no warning, unlike the rest of the file's cleanups.

**Evidence:** `packages/workflows/src/workflow-source.test.ts:25` and `packages/workflows/src/fixture-runner.test.ts:33` clean up `bundledDefaultsRoot` with `await rm(bundledDefaultsRoot, { recursive: true, force: true }).catch(() => {})`. `packages/paths/src/test-utils.ts:58-72` (`removeTempTree`) exists specifically to retry through the transient-error codes Windows CI produces for exactly this kind of cleanup (issue #2306, cited elsewhere in this same PR's `review-mode-script.test.ts` comment) and warns instead of silently swallowing a genuine leak. `workflow-source.test.ts:50` already imports `trackTempRoots` from that same module for the file's other temp roots.

**Required outcome:** Not required — actual risk is low since the tree is never written into during the test run, so the failure mode `removeTempTree` guards against (a just-exited process or async write still holding a handle) doesn't apply here the way it does to the file's other temp roots. Suggest routing this cleanup through `trackTempRoots`/`removeTempTree` for consistency next time this file is touched.

</details>

<details>
<summary><code>R3</code> — New regression guard added at only 1 of 3 sites carrying the identical risk statement</summary>

**Impact:** None currently reachable. If the bundled-scope capture mechanism silently regressed in `fixture-runner.test.ts` or `subrun.test.ts` the way `workflow-source.test.ts:566-570`'s new guard protects against there, nothing in those two files would catch it — but neither file's own assertions depend on `manifest.scopes` today, and the only production readers (`orchestrator.ts:544`, `orchestrator-agent.ts:1491`) check `scopes.length > 0`, not bundled-content specifically, and aren't exercised by any of the three files.

**Evidence:** `workflow-source.test.ts:566-570` adds `expect(capture.manifest.scopes).toContain('bundled')` with a comment explaining the lever could otherwise silently degrade. `fixture-runner.test.ts:24-29` and `subrun.test.ts:34-45` state the same risk in near-identical prose but assert nothing. `fixture-runner.test.ts` structurally can't easily add the same assertion — `FixtureReport` (`fixture-runner.ts:246`) never exposes the manifest to the caller.

**Required outcome:** Not required for this PR — `subrun.test.ts` is a pre-existing file untouched by this diff, and `fixture-runner.test.ts`'s architecture doesn't expose the manifest for a symmetric assertion without a larger change. Worth a follow-up if `fixture-runner.ts` is ever changed to surface capture metadata.

</details>

<details>
<summary><code>R4</code> — `tempFiles` registry is oversized for what it now manages</summary>

**Impact:** None — purely a readability/proportionality observation.

**Evidence:** `packages/workflows/src/defaults/review-mode-script.test.ts:18,54-56` keeps a module-level `tempFiles: string[]` array plus an `afterAll` that drains it, but after the split there is exactly one file created, in one test, for the file's whole lifetime (the third test's `missingReport` is deliberately never created or pushed).

**Required outcome:** Not required. A `try/finally` scoped to the one test that writes the file would remove the module-level mutable array and better localize ownership (the test that creates the file also cleans it up).

</details>

<details>
<summary><code>R5</code> — Three explanatory comments cite precedent loosely or overstate scope</summary>

**Impact:** None on behavior — comment accuracy only, but these comments are load-bearing for the next maintainer's trust in the stated rationale.

**Evidence:**
- `review-mode-script.test.ts:27` cites `#2860` for "hoisting into a shared `beforeAll` would put them back under one deadline." #2860's actual fix (in `packages/cli/src/cli.test.ts`) was extracting in-process guards and batching remaining spawns *concurrently*, not a statement about `beforeAll` specifically — a loose gloss on the right underlying lesson, not a fabricated one.
- `review-mode-script.test.ts:67` cites `#2306` for "recursive temp-tree removal is slow and flaky on Windows"; #2306 documents subprocess-spawn timeouts, a related but distinct failure mode from directory-removal cost.
- `fixture-runner.test.ts:634-636` states "every test in this block pays this cost" (referring to `initGitWithCommittedFile`); 2 of the 9 tests in `describe('runFixtures exec-code isolation')` never call it.

**Required outcome:** Not required — none of these misstate the actual code behavior being changed, only the supporting citations. Worth tightening the wording on a future touch of these files.

</details>

<details>
<summary><code>R6</code> — New `@archon/paths` mock in `fixture-runner.test.ts` coexists with `state-migration.test.ts`'s mock via an undocumented Bun behavior</summary>

**Impact:** None today — independently verified passing (111/111 in the shared `bun test state-migration.test.ts src/dry-run.test.ts src/fixture-runner.test.ts` group, matching the PR's own stated verification). But the mechanism that makes `{...realArchonPaths, ...}` spread the *complete* real module rather than a stale partial mock left behind by `state-migration.test.ts` (which registers its own incomplete `@archon/paths` mock containing only `createLogger`) is an empirically-observed Bun import-ordering quirk, not a documented contract — and it sits next to this same codebase's own note elsewhere (`include-expander.ts:83`, `subrun.test.ts:11`) that `mock.module` is process-global and irreversible.

**Required outcome:** Not required for this PR — the implementer's own report already identifies and empirically validates this exact coexistence risk. Worth a short comment noting the dependency so a future test added to this same `bun test` invocation group doesn't break the ordering assumption silently.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R5, R6 |
| `seams` | R1, R2, R3 |
| `simplify` | R1, R4, R5 |

#### Validation

|---|---|---|
| `bun test src/workflow-source.test.ts` (5x) | PASS | 32/32 every run, 280-307ms, no flakes |
| `bun test src/state-migration.test.ts src/dry-run.test.ts src/fixture-runner.test.ts` (5x) | PASS | 111/111 every run, 929-978ms, no flakes |
| `bun run test` (full `@archon/workflows` package suite, 26 groups) | PASS | 0 fail across all groups, ~11s wall time, matches PR's claimed 11.2-11.3s |
| `bun x tsc --noEmit` (workflows package) | PASS | no output, exit 0 |
| `bun run lint` (repo-wide) | PASS | clean across all packages including `workflows` |
| Mutation: bundled tree left uncreated | Fails exactly `workflow-source.test.ts:570` (`toContain('bundled')`), 31/32 pass | Confirms the new guard discriminates |
| Mutation: `resolve-review-mode.py` `continuation` forced `False` | Fails exactly "reports a continuation..." test, 2/3 pass | Confirms test-per-case isolation |
| Mutation: `resolve-review-mode.py` `continuation` forced `True` | Fails exactly "reports a full review..." test, 2/3 pass | Confirms test-per-case isolation |
| Mutation: `resolve-review-mode.py` missing-file guard removed | Fails exactly "fails when the named prior report..." test, 2/3 pass | Confirms test-per-case isolation |
| `git diff dev...HEAD` grep for timer/timeout/budget increases | None found | Only comment references to the existing 5000ms default |
| CI: `test (windows-latest)` on `97100871` | PASS, 7m57s | https://github.com/coleam00/Archon/actions/runs/33130154400/job/98717531915 |
| CI: `test (ubuntu-latest)`, `docker-build`, `schema-upgrade` | PASS | Same run |

</details>

Issue #2896, opened 2026-08-28
fix(workflows): apply the existing idle-stall watchdog to every AI node — coverage is currently arbitrary


Wirasm:
## Problem

The engine already has a working idle watchdog: in run `dfa58b90`, a stalled `review__docs` node was caught and handled cleanly after ~30 minutes of provider silence. But coverage is not uniform: in run `af6e422b`, `review__code` sat on a dead OpenRouter stream for **75.7 minutes** — 64% of the run's wall clock — before the provider deigned to error; run `dfa58b90` also shows a second unattributed 45.8-minute gap in its implement phase, consistent with an unguarded node. Two other runs (`b4e9597b`, `c7125eff`) hung silently on dead streams with no watchdog firing at all and needed manual kill → abandon → `--adopt` recovery.

Wall-clock audit with per-run tables: `~/.prp/archon-75601ef6/reports/ci-wallclock-audit.md` (2026-08-28). Liveness evidence thread: #2325.

## Desired outcome

The proven watchdog mechanism applies to **every AI node** — review lenses, implement-loop iterations, and any other provider-attached node — not an arbitrary subset. Same threshold and same recovery semantics as the working case; this is coverage uniformity, not new design.

## Boundary

This is the practical half only. The full liveness read-model (truthful live-vs-orphan signals, filesystem/process evidence, operator surfacing) stays with #2325 — do not pull that design in here. Also out of scope: run-level timeouts (forbidden — a run reports when done; the watchdog guards a NODE's provider stream, which is a recoverable operation per AGENTS.md's timer rules).

## Invariant

No provider-attached node can wait unboundedly on a silent stream while its siblings are guarded. A watchdog firing is observable in the run's events (it already is in the working case — keep that).

## Regression proof

- A test proving the watchdog arms for each AI node type (the working case presumably has one — extend its coverage, or add the missing one).
- Evidence standard for done: a seeded silent-stall (mock provider that stops yielding) on an implement node and on a review lens both recover the way `review__docs` did.

## Expected value

~45 minutes recovered per stall at the observed stall rate (3 silent hangs across 7 GLM runs in one day); removes the manual kill/adopt playbook for the most common case.

Wirasm:
[...]

Issue #1807, opened 2026-05-29
maintainer-review-pr silently reports success when review-classify times out (no aspects run, nothing posted)


Wirasm:
## Summary

- **What broke:** `maintainer-review-pr` (Pi/Minimax variant) can report a successful run (`exit 0`, `anyFailed: false`) while reviewing **nothing** — no aspects run, no comment posted, no state recorded. The trigger is the `review-classify` node timing out; the deeper bug is that a classifier producing no structured output causes every downstream aspect to be silently skipped and the run to still report success.
- **When it started (if known):** Not a regression from the Pi SDK bump (#1800) — see "Root cause" below. The fragility is pre-existing in the workflow definition.
- **Severity:** `major` (produces false "review complete" signals on real PRs; a clean PR can look reviewed when it wasn't, and a bad PR can pass unreviewed).

## Root cause analysis

Observed on a live run of `maintainer-review-pr` against PR #1789 (run `f97f373646560089ae571555da4bb45a`).

**Chain of events:**
1. `review-classify` is fed `$fetch-pr.output` + `$fetch-diff.output`. `fetch-diff` caps the diff at 2500 lines; PR #1789's diff is 1055 lines / ~44 KB, so nearly the whole diff is injected. With the `output_format` JSON-schema augmentation on top, the prompt is ~45 KB.
2. The node carries an explicit **`idle_timeout: 60000` (60s)** — the tightest timeout in the workflow, on its **largest** prompt. Every other node inherits the 30-minute default (`STEP_IDLE_TIMEOUT_MS`).
3. Minimax M2.7 emitted **zero events** (no `text_delta`, no `thinking_delta`) for 60 continuous seconds — time-to-first-token on a 45 KB structured-output prompt exceeded the idle window — so the watchdog reaped the node.
4. With no classifier output, every gate (`$review-classify.output.run_code_review == 'true'`, etc.) evaluated false → all 5 aspect nodes **skipped** → `synthesize-review`, `post-review`, `record-review`, `report` skipped via `trigger_rule`.
5. Workflow reported **success**. Nothing was reviewed; nothing was posted to the PR; nothing recorded in shared state.

**This is NOT the #1800 Pi SDK bump.** Ruled out by inspection:

| Suspect | Verdict |
|---|---|
| #1800 tool change (`tools:[]` → `customTools:[], noTools:'builtin'`) | Cleared — `extract-pr-number` uses the same `allowed_tools: []` path on 0.76.0 and completed in 5.5s. |
[...]

PR #2902, opened 2026-08-28
Fix silent loop watchdog retries

Wirasm:
## Problem and outcome

Loop AI nodes treated a provider stream that timed out before producing output as a completed iteration. A silent stream could therefore exhaust the loop's iteration budget and report only max-iteration exhaustion, instead of using the existing watchdog recovery path.

- **Outcome:** A zero-output loop watchdog timeout now records an iteration failure and retries through the existing transient-retry policy.
- **Invariant:** A watchdog timeout after assistant text or structured output preserves the useful output and completes the iteration as before. A loop declaring `output_format` keeps failing a zero-output timeout on its first attempt, with the structured-specific error, exactly as the single-shot AI-node path does.
- **Scope boundary:** This changes loop-iteration timeout recovery only; it does not add run-level timeouts or alter the existing single-shot AI-node watchdog behavior.
- **Root cause:** The loop's empty-output guard explicitly exempted idle-timeout exits, classifying a silent timeout as a successful iteration.

## Review guidance

- **Feedback requested:** Confirm that a timeout before any provider output follows the same retry semantics as other empty AI responses, while a timeout after useful output remains successful.
- **Start here:** `packages/workflows/src/dag-executor.ts` — the loop's empty-output guard now includes idle-timeout exits and supplies a timeout-specific error, except for a structured iteration, which the structured-output branch below already owns.
- **Review order:** Review the loop recovery branch, then the two regression tests in `packages/workflows/src/dag-executor.test.ts`.

## Solution

The existing loop empty-output failure path now also handles an idle watchdog firing before the provider emitted text or structured output. That path emits `loop_iteration_failed` and uses the established transient retry policy, so the failed attempt does not consume the loop's iteration budget.

An iteration that declares `output_format` is exempt from that guard when it times out: the structured-output branch below already reports the missed contract and fails on the first attempt, so retrying there would spend another full `idle_timeout` window per attempt and replace a precise error with a generic one. That mirrors `executeNodeInternal`, where the same generic guard is unreachable once `output_format` is set, and mirrors the loop's own `canReask` idle-timeout exclusion.

The "completed via idle timeout" notification now carries the same has-output condition as the single-shot one, so a zero-output timeout is never announced as a completion before the failure that follows it.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A silent watchdog timeout consumed a loop iteration and could end as max-iteration exhaustion. | A silent watchdog timeout emits `loop_iteration_failed` with a timeout error. |
| **Failure behavior** | The loop skipped transient retry for the failed provider stream. | The existing transient retry path retries the iteration without consuming its budget. |
| **Structured loops** | A zero-output timeout failed immediately with the structured-specific error, but was announced as "completed via idle timeout" first. | Same immediate failure and error, and no contradictory completion notice. |

## Validation

- `bun test src/dag-executor.test.ts` in `packages/workflows` — passed (684 tests) — exercises a zero-output loop timeout, `loop_iteration_failed`, retry, successful recovery, and the structured-loop timeout that must not retry.
- `bun run test` in `packages/workflows` — passed (24 invocations, 0 failures).
- `bun x prettier --check src/dag-executor.ts src/dag-executor.test.ts` in `packages/workflows` — passed.
- `bun x tsc --noEmit` in `packages/workflows` — passed.
- `bun run validate` — passed — generated-artifact checks, type checks, lint, formatting, installer tests, and the full test suite.
- **Not verified:** Nothing material; the regressions directly cover both changed recovery paths, and the existing adjacent test covers the ordinary AI/review-node path.

## Links

- Part of #2896. This PR fixes the loop-iteration classification half only. The issue's other evidence, a review lens that stalled 75.7 minutes before the provider itself errored and two runs that hung with no watchdog firing, is not explained by this change, so #2896 stays open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of AI loop iterations that time out without producing output.
  * Empty-output timeouts now retry the affected iteration instead of consuming the iteration limit or failing the entire workflow.
  * Idle-timeout notifications are emitted only when output is available.


Wirasm:
# Review report

## Verdict

Ready — action: `none`. The zero-output loop watchdog exit now records an iteration failure and uses the existing transient retry path without advancing the loop iteration, while output-bearing timeout handling remains unchanged.

## Accepted contract

A loop AI node whose provider stream reaches its idle watchdog before producing assistant text or structured output must fail that loop iteration and retry under the existing transient-retry policy without consuming the loop iteration budget. A timeout after assistant text or structured output must preserve that useful output and complete the iteration as before.

Boundary: loop-iteration timeout recovery only. Run-level timeouts and existing single-shot AI-node watchdog behavior are out of scope.

## Reviewed head SHA

`924aece000b99fd4b3ca843885eaf400b93ae834`

This is the cursor for a future continuation review.

## Findings

No Critical or Important findings.

The watchdog callback marks the iteration as timed out and aborts its provider stream at `packages/workflows/src/dag-executor.ts:6270-6277`. After a zero-output return, the changed guard at `packages/workflows/src/dag-executor.ts:6682-6715` persists `loop_iteration_failed` and invokes the existing transient retry helper before control can leave the iteration attempt; the outer loop counter therefore does not advance. The timeout-specific error contains `timeout`, which the existing classifier treats as transient. If an attempt has assistant text or a structured payload, it bypasses that guard and retains the existing completion notification at `packages/workflows/src/dag-executor.ts:6724-6733`.

The focused regression test passed:

```
cd packages/workflows && bun test src/dag-executor.test.ts --test-name-pattern "idle-timeout with zero output fails a loop iteration"
```

Result: 1 passed, 0 failed (682 filtered). The test at `packages/workflows/src/dag-executor.test.ts:13591-13668` verifies a zero-output watchdog attempt followed by a successful retry, without `node_failed` or workflow-run failure despite `max_iterations: 1`.

### Rejected findings

None. The specialist reports contained no Critical or Important claims to reject; their clean-path claims were verified against the reviewed code and the focused regression test.

### Suggestions

None.

## Discoveries

0 accepted discoveries. [JSON](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/5cec017a-9fba-41ad-a38e-43f9d3131717/discoveries.json) and [human-readable report](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/5cec017a-9fba-41ad-a38e-43f9d3131717/discoveries.md). Adjacent discoveries do not affect readiness.

## Review coverage

Full review. Selected concerns `code`, `seams`, `errors`, `comments`, and `simplify` all reported and were read in full; all found no blocking issue. `tests`, `types`, and `docs` were disabled by the review inputs, so their absent reports do not create a coverage gap. The reviewed diff was `git diff dev...924aece000b99fd4b3ca843885eaf400b93ae834`; `git diff --check` passed.

Wirasm:
## Ready to merge

All three findings from the first round are fixed, verified against the actual diff (not just the fix report). R1's structured-loop divergence is closed by a new `structuredTimeout = wantsStructured && iterationIdleTimedOut` exemption that routes a structured loop's idle-timeout-zero-output case to the existing structured-specific failure branch — mutation-tested (reverting the exemption turns the new regression test red for the right reason: 3 provider calls instead of 1) and confirmed to still fail observably (`node_failed`, specific message), not silently pass. R2's notification gap is closed with an explicit has-output condition and assertions on both the structured and unstructured paths. R3's issue-linkage concern is closed: the PR body now says "Part of #2896" with `closingIssuesReferences` empty, and #2896 carries an accurate scope note. See the Delta section below for verification detail.

**Validation:** `bun test src/dag-executor.test.ts` (packages/workflows) — 684 pass, 0 fail (683 + 1 new structured-loop test), including all four watchdog-related tests. `bun x tsc --noEmit` — clean. Two mutation checks (original R1 guard, delta's `structuredTimeout` exemption) both go red for the right reason. GitHub Actions on `1930493f` (current head) — green on all 5 required checks, confirmed via the commit-level check-runs API. Two unrelated CI flakes surfaced and resolved during verification — see Delta.

### Findings

|---|---|---|---|
| `R1` | Important | A structured-output (`output_format`) loop node's idle-timeout-zero-output case now gets extra transient retries instead of the immediate structured-specific failure both this loop path (pre-PR) and the single-shot path (still) produce | FIXED |
| `R2` | Suggestion | No regression test locks in the reordered "completed via idle timeout" notification for loop nodes | FIXED |
| `R3` | Suggestion | "Closes #2896" is broader than what this PR fixes — the issue's own headline evidence (a review lens stalled 75.7 min, two runs with no watchdog firing at all) isn't explained by this change | FIXED |

<details>
<summary><code>R1</code> — structured loop idle-timeout now diverges from single-shot and from its own pre-PR behavior</summary>

**Impact:** For a loop node that declares `output_format` (the shape an `implement`-style loop typically uses to signal structured completion), a genuine idle-timeout with zero output now takes up to 3 provider attempts (1 initial + `DEFAULT_NODE_MAX_RETRIES` = 2 retries, each waiting up to the node's full `idle_timeout`) before failing, instead of failing immediately after the first attempt as it did before this PR and as the single-shot AI-node path still does. In the worst case (default 30-minute `idle_timeout`, provider genuinely dead) that's up to ~90 minutes of wasted wall-clock before the node fails — working against the exact problem issue #2896 exists to fix, for this specific reachable path. The final error message also changes from the structured-specific `"...timed out before producing the required structured output."` to the generic `"...timed out with no output..."`, losing information about which contract failed.

**Evidence:** `packages/workflows/src/dag-executor.ts:6682` — the (now-unconditional) empty-output guard `if (fullOutput.trim() === '' && attemptStructured === undefined)` sits structurally *before* the structured-output gate at `dag-executor.ts:6737` (`if (!wantsStructured) break attempts;`), so it intercepts a structured loop's idle-timeout-zero-output attempt and routes it into `tryIterationTransientRetry` (confirmed TRANSIENT via `classifyError` — the message contains the substring `"timeout"`). Compare `executeNodeInternal`: its equivalent generic guard (`dag-executor.ts:3194-3199`) sits *after* the entire structured validate-and-reask `while(true)` block (`dag-executor.ts:2967-3070`), which always exits via `break` (non-structured, `dag-executor.ts:2993`) or `throw` (structured, `dag-executor.ts:3061-3069`) — so for a structured single-shot node, line 3194 is unreachable, and the idle-timeout case throws immediately via the structured-specific message at `dag-executor.ts:3061-3065` with no retry (that message doesn't match `TRANSIENT_PATTERNS`). This is the exact "same contract as the single-shot AI-node guard" the PR's own comment at `dag-executor.ts:6668-6672` claims to preserve, and it no longer holds for this combination. `DEFAULT_NODE_MAX_RETRIES = 2`, `DEFAULT_NODE_RETRY_DELAY_MS = 3000` confirmed at `dag-executor.ts:971-972`.

The delivery run's own internal seam review (`~/.archon/workspaces/coleam00/Archon/artifacts/runs/5cec017a-9fba-41ad-a38e-43f9d3131717/review/seams.md`) compared the two guards' *conditions* and found them shaped the same, but didn't check their *position* relative to the structured-output gate — which is where the divergence actually lives.

**Required outcome:** Route a structured loop's idle-timeout-zero-output case to the existing structured-specific branch (`dag-executor.ts:6812-6835`, unchanged, already timeout-aware) instead of the generic empty-output guard — e.g. gate the guard at `dag-executor.ts:6682` on `!wantsStructured` for the idle-timeout sub-case (mirroring `canReask`'s existing `!iterationIdleTimedOut` exclusion), or restructure so the structured branch is checked first when `wantsStructured` is true. Add a test seeding an `output_format` loop through an idle-timeout-zero-output attempt to pin down which branch it takes.

**Found by:** `prp-core:seam-analyzer`, independently confirmed by the coordinator via direct trace of both code paths.

**Disposition:** FIXED in `7280fdb1` ("fail a structured loop's silent timeout immediately"). See Delta.

</details>

<details>
<summary><code>R2</code> — loop idle-timeout notification reordering has no regression test</summary>

**Impact:** The PR moved the "completed via idle timeout" `safeSendMessage` call from firing unconditionally to firing only after the empty-output guard doesn't claim the iteration — this reordering *is* the fix for the old contradiction (announcing "completed" moments before reporting a failure). Nothing currently pins that ordering down.

**Evidence:** `packages/workflows/src/dag-executor.ts:6724-6733`. Verified directly: moving the notify block back above the empty-output check (restoring the pre-PR position, while keeping the classification fix) still passes the full 683-test suite — the regression is real and currently undetected.

**Required outcome:** Extend the existing new test (`dag-executor.test.ts:13591-13668`) to assert the platform message mock does not contain `'completed via idle timeout'` on the failed/retried attempt. One assertion, no new test needed.

**Found by:** `prp-core:pr-test-analyzer`.

**Disposition:** FIXED in `7280fdb1`. See Delta.

</details>

<details>
<summary><code>R3</code> — "Closes #2896" outruns what the diff demonstrably fixes</summary>

**Impact:** Issue #2896's own evidence includes a `review__code` lens stalled 75.7 minutes before the *provider itself* errored (not the watchdog) and two runs that "hung silently... with no watchdog firing at all." This PR does not explain or fix either: watchdog wrapping (`withIdleTimeout`) already applied uniformly to both AI-streaming call sites in the package before this PR (confirmed — there are exactly two `sendQuery` call sites in `packages/workflows`, both already wrapped), and `review__code`'s single-shot code path is byte-for-byte identical to `review__docs`'s (both `kind: 'agent'`, same `withIdleTimeout` call, same pre-existing empty-output guard proven by the `#1807` regression test at `dag-executor.test.ts:13519`). Merging with "Closes #2896" auto-closes the issue while that evidence remains unaddressed and unexplained.

**Evidence:** `~/.prp/archon-75601ef6/reports/ci-wallclock-audit.md:51,55` (the wall-clock audit issue #2896 cites); `packages/workflows/src/dag-executor.ts:2358` and `:6258` (the only two `sendQuery` call sites, both wrapped pre-PR); `.archon/workflows/sdlc/review/archon-review.yaml:71-145` (review lenses have no explicit `kind:`, defaulting to `agent`, same as `review__docs`).

**Required outcome:** Not a code defect — a scope/issue-linkage judgment call for the operator. Either keep #2896 open with a comment on what this PR actually covers (loop-iteration recovery classification only) and file a narrower follow-up for the `review__code`-style stall (which may be a `withIdleTimeout` reset-semantics edge case under specific provider streaming behavior, plausibly adjacent to the liveness-model work in #2325), or explicitly accept this as the "practical, code-fixable half" and close with that caveat recorded.

**Found by:** `prp-core:code-reviewer`, `prp-core:silent-failure-hunter` (convergent).

**Disposition:** FIXED — PR body now reads "Part of #2896" (not "Closes"), `closingIssuesReferences` is empty, and issue #2896 carries an accurate scope-note comment. See Delta.

</details>

## Delta (fix round)

Verified against the actual diff in a fresh detached worktree (`924aece0..7280fdb1`, then the subsequent no-op merge to `1930493f`), not the fix report.

**R1 — structured-loop exemption, verified correct and non-silent.** The delta adds `const structuredTimeout = wantsStructured && iterationIdleTimedOut;` and guards the generic empty-output check with `!structuredTimeout &&` (`dag-executor.ts:6688-6689`). Traced the resulting control flow by hand: for a structured loop's idle-timeout-zero-output attempt, execution now skips the generic guard, skips the (also-fixed) notification, passes `if (!wantsStructured) break attempts;` (false), computes `canReask` as `false` (idle-timeout excluded, matching the existing `canReask` exclusion), falls to the "No structured payload at all" branch, and returns `failLoopNode(...)` with the structured-specific message `"...timed out before producing the required structured output."` — a real, observable, first-attempt failure (`node_failed` persisted and emitted), not a silent pass-through. This exactly restores the pre-PR loop behavior and matches the single-shot path, which the delta's own new comment cites correctly.

Mutation-tested directly: reverted the `structuredTimeout` exemption and the notification's has-output condition in a scratch copy of `dag-executor.ts`, reran the new test — it failed with `Expected number of calls: 1, Received number of calls: 3` (the old bug's extra transient retries), then restored the file (`git diff --stat` confirmed clean). The new test `idle-timeout with zero output fails a structured loop immediately, without transient retries` (`dag-executor.test.ts:13697+`) asserts exactly the three things that matter: exactly one provider call, `node_failed` with the structured-specific message, and no false "completed" notification.

**R2 — notification fix, verified.** The has-output condition `iterationIdleTimedOut && (fullOutput.trim() !== '' || attemptStructured !== undefined)` (`dag-executor.ts:6739-6742`) is stronger than a pure position argument — it holds regardless of future reordering. Both the existing non-structured test and the new structured test now assert `sentMessages`/`platform.sendMessage` calls exclude `'completed via idle timeout'` on the failure path. Full suite: 684 pass (683 + 1 new test — the R2 fix added assertions to an existing test rather than a new one, consistent with the original required-outcome suggestion).

**R3 — issue linkage, verified.** `gh pr view 2902 --json body,closingIssuesReferences` confirms `closingIssuesReferences: []` and the body's Links section reads: "Part of #2896. This PR fixes the loop-iteration classification half only. The issue's other evidence... is not explained by this change, so #2896 stays open." Issue #2896 carries comment `5450801287` with an accurate, non-overclaiming scope note that matches the review's original R3 finding almost verbatim.

**CI — two unrelated flakes, both resolved on rerun, root-caused before reruning:**
1. First rerun of `7280fdb1`'s CI showed `test (ubuntu-latest)` failing `check-test-cleanup-drift` on `packages/workflows/src/defaults/discovery-report.test.ts:71` — a file this PR never touches. Traced: the PR's merge-preview base (`bfeb6d3f`) was 2 commits behind `origin/dev`, and one of those 2 commits (`1e8a6476`, merged via PR #2911) is the exact fix for that drift. Not a PR-caused failure — base-branch staleness in the merge-preview snapshot. Confirmed resolved: ubuntu passed clean on rerun.
2. Same rerun's `test (windows-latest)` failed two `scripts/migrate-state-dir.test.ts` tests with `this test timed out after 5000ms` / exit code 143 (killed process) — a Windows subprocess-spawn timing flake in code this PR never touches. Reran once more: fully green.
3. A subsequent "Merge origin/dev" commit (`1930493f`, the current head) landed while polling — diffed it against `7280fdb1` and confirmed **zero changes** to `dag-executor.ts` or `dag-executor.test.ts`; it only pulls in the unrelated `dev` fixes (including the discovery-report drift fix). Re-verified CI directly against this exact SHA via the commit-level check-runs API: all 5 required checks `conclusion: success`.

No new findings from the delta itself.

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | No additional findings beyond R3 |
| `seams` | R1 |
| `simplify` | No findings — change judged proportionate, reuses existing retry helper as intended |
| `tests` | R2; independently confirmed the new test exercises the real watchdog (not a stubbed primitive) and decisively proves the loop's `i` counter isn't advanced by the retry |
| `errors` | No findings — confirmed the failure is observable on 3 independent channels (event emitter, DB-persisted event, error log) on both the retry-succeeds and retry-exhausted branches, and confirmed the timeout message classifies TRANSIENT so the retry actually fires on first occurrence |

#### Validation

|---|---|---|
| `bun test src/dag-executor.test.ts` (packages/workflows), round 1 head `924aece0` | PASS | 683 pass, 0 fail, 2001 expect() calls |
| `bun test src/dag-executor.test.ts` (packages/workflows), delta head `7280fdb1`/`1930493f` | PASS | 684 pass, 0 fail, 2006 expect() calls |
| `bun x tsc --noEmit` (packages/workflows), both rounds | PASS | Clean, no errors |
| Mutation check round 1: revert original guard, rerun loop test | PASS (as expected) | `toHaveBeenCalledTimes(2)` → received 1; red for the right reason |
| Mutation check delta: revert `structuredTimeout` exemption + notify condition, rerun structured test | PASS (as expected) | `toHaveBeenCalledTimes(1)` → received 3; red for the right reason; file restored, `git diff` clean |
| ESLint / Prettier on changed files | PASS | Confirmed clean by two independent review agents (round 1) |
| GitHub Actions on `924aece0` | PASS | Single run, all 5 checks green; no rerun |
| GitHub Actions on `1930493f` (current head) | PASS | All 5 checks green, confirmed via commit-level check-runs API; two unrelated flakes hit and resolved en route (base-branch staleness on `discovery-report.test.ts` drift; Windows `migrate-state-dir.test.ts` subprocess timeout) — see Delta |
| `bun run validate` | NOT RUN (either round) | Reported passing in the PR body and by the delivery run's own validation pass each round; not independently re-run given the narrow diff and confirmed-green CI |

</details>

Issue #2897, opened 2026-08-28
ci: three measured quick wins — docs path filter, Windows bun cache, docker disk-free step


Wirasm:
## Problem

The 2026-08-28 CI wall-clock audit (`~/.prp/archon-75601ef6/reports/ci-wallclock-audit.md`, all numbers from real `gh api` job timings) found three config-level wastes in the test/build workflows:

1. `test.yml` is the only workflow file with no path filter — a docs/markdown-only PR burns ~17–21 runner-minutes on test matrices that cannot be affected.
2. Windows dependency install runs 109–119s vs ~6s on Linux (~18×) because `setup-bun` has no built-in dependency cache and nothing fills the gap — caching `~/.bun/install/cache` keyed on `bun.lock` closes it. (Linux caching is not worth doing; it is already fast.)
3. `docker-build` runs an unconditional "free runner disk space" step costing 75–86s on every run regardless of cache state — a larger-disk runner or a conditional drops it.

## Desired outcome

The three fixes above, each verified against a real run's timings before/after in the PR body.

## Explicitly OUT of scope (operator decisions, not config tweaks)

- **Type-check + lint duplication per OS** (70% of the ubuntu leg, 62% of Windows): real money, but deduplicating changes what the per-OS matrix guarantees — needs a deliberate maintainer call first.
- **Runner migration** (Blacksmith is the one drop-in option with Windows support and a real OSS tier; BuildJet confirmed shut down March 2026): parked by the operator for later.

## Invariant

No behavioral coverage changes: the same checks run for code changes on the same matrix; only genuinely unaffected runs are skipped and dead setup time removed.

## Regression proof

- A docs-only PR shows the test matrix skipped; a code PR shows it unchanged.
- A Windows job log shows the cache hit and install under ~15s.
- docker-build total drops by the step's cost with the image build unchanged.

Wirasm:
**Work order narrowed (operator decision, 2026-08-28), resolving the replan from run `cb051412`.** The review proved the disk-free item is unwinnable as scoped: a path filter is a cache-state proxy that lies on cache misses, and the only reliable config-level alternative is unconditional pre-build cleanup — the status quo cost relocated, not a win. A pre-build cache-state signal would be machinery disproportionate to ~80 seconds.

[...]

PR #2903, opened 2026-08-28
ci: avoid unnecessary test setup

Wirasm:
## Problem and outcome

Markdown-only changes and changes confined to the docs workspace were running the full test matrix, and Windows fetched Bun dependencies without a cache. This PR avoids that unnecessary CI setup while preserving coverage for every change that can affect the test suite.

- **Outcome:** Markdown-only changes and docs-workspace-only changes skip this workflow's test suite; Windows restores Bun's package cache.
- **Invariant:** Non-Markdown changes outside `packages/docs-web/` and changes to the docs workspace manifest still run the existing test matrix and Docker build.
- **Scope boundary:** The per-OS type-check/lint matrix and runner selection are unchanged.

## Review guidance

- **Feedback requested:** Confirm the complete-diff decision correctly distinguishes documentation-only changes from changes that require the test suite.
- **Start here:** `.github/workflows/test.yml:15` — the change-decision job gates the existing CI jobs.
- **Review order:** Review the change-decision script, the job gating and aggregate result, then the Windows cache step.
- **Lower-attention areas:** None; this PR changes one workflow and its focused decision script and tests.
- **Known risk or uncertainty:** None outstanding. The first remote runs surfaced two Windows-only test defects, both fixed here — see "Windows fixes" below.

## Solution

The workflow checks the complete changed-file set and delegates the decision to a small script. The suite is skipped only when every changed file is prose: Markdown or something inside `packages/docs-web/`, and not one of the documentation-shaped paths a build or drift check actually reads. Those exceptions are `.archon/commands/` and `.archon/workflows/` (compiled into `bundled-defaults.generated.ts`), `.claude/skills/` (text-imported into the CLI binary), the generated `provider-capabilities.md`, and the docs manifest the Docker dependency layer copies. If the diff cannot be read at all, the script says "run the suite" and explains why on stderr. The Windows matrix leg restores Bun's global package cache from a lockfile-keyed cache. An always-run aggregate job reports the suite result after the gated jobs finish.

## Behavior change

| | Before | After |
|---|---|---|
| **Prose-only change (Markdown or docs workspace)** | Full test matrix and Docker build run. | Matrix and Docker build jobs are skipped. |
| **Change to bundled or generated Markdown** (pack command prompts, the CLI skill, `provider-capabilities.md`, the docs manifest) | Full test matrix runs. | Unchanged — still runs, so the drift checks that guard those files cannot be skipped. |
| **Base commit unreadable** (e.g. after a force-push) | Full test matrix runs. | Unchanged — the script resolves to "run the suite" and logs the cause. |
| **Windows dependency install** | Dependencies are downloaded each run. | Bun package cache is restored when `bun.lock` matches. |

## Windows fixes

The first remote run went red on the Windows matrix leg only, on two unrelated test defects. Both are fixed here.

**The new decision test could not survive a Windows checkout.** `scripts/should-run-test-suite.test.ts` reads `.github/workflows/test.yml` raw and asserts LF-anchored substrings, but a Windows checkout materializes the file with CRLF, so `"  push:\n"` never matched. Line endings are now normalized at the read site: the assertions are about YAML structure, not the terminator the working tree happens to carry.

**`scripts/check-test-cleanup-drift.test.ts` is a fourth file in this diff, and deliberately so.** It is pre-existing and unrelated to the decision script, but its `propagates Git scan failures` test was red on the same Windows leg — 5282ms against a 5000ms per-test budget — so it blocked this PR. Fixing it here is unblocking work rather than unrelated cleanup; splitting it out would cost a review round for no isolation benefit.

The cost was not generic Windows spawn latency. The test provoked its failure with an unknown Git *subcommand*, which sends Git searching every PATH entry for a `git-*` executable to offer a correction — one spawn whose cost scales with the size of PATH, and each probe on Windows is far more expensive than on Linux. For comparison, every other Git spawn in that same file on that same runner costs ~50ms. The test now fails a rejected option on the same `ls-files` the scan actually runs, which fails locally at constant cost, with the assertion shape unchanged so the failure-propagation coverage is identical. No timeout was raised and nothing is skipped on Windows: 5282ms → 31ms.

## Review corrections

The review found two blocking issues, both on the decision path. Each mechanism was reproduced locally before it was changed.

**A crash produced an empty decision.** The script threw uncaught when `git diff` could not read the base commit — what a force-push produces, since `github.event.before` can name an unreachable commit. `echo "run-tests=$(...)"` does not trip `set -e` when the substituted command fails, so the step reported success having written an empty `run-tests`, which every downstream `== 'true'` gate reads as "skip". Only the summary job's job-result loop caught it, by accident. Now the scan failure resolves explicitly to "run the suite" with the cause on stderr, and the step assigns before echoing so a wiring bug fails the job instead of publishing an empty decision. A bad argument stays an uncaught throw, because there is no safe answer to it.

**The `.md` test swept in bundled and generated content.** Pack command prompts, the skill files compiled into the CLI, and the generated `provider-capabilities.md` were all classified as documentation — and `check:bundled`, `check:bundled-skill` and `check:capability-matrix` all run only inside the job this decision can skip, so editing a bundled prompt skipped the check built to catch that drift. Those paths now force a run, generalizing the carve-out that already existed for the docs manifest.

Since the exclusion list is written by hand, one test derives the skill family from `bundled-skill.ts`'s own text imports and fails if any file the CLI compiles in would be classified as prose. The wiring test now pins the literal invocation and argument order rather than checking that fragments appear somewhere in the job.

## Validation

- `bun x prettier --check .github/workflows/test.yml` — passed.
- `bun run validate` — passed; generated checks, type-check, lint, formatting, installer tests, and the complete test suite passed.
- **Remote CI — passed.** Run 33165582920 is green on every job at the current head, both matrix legs included; the earlier run 33163243834 is the green run for the Windows fixes.
- The eight new tests were seen red before they were seen green: run against the pre-fix script and workflow, seven of them fail (all four excluded families, the derived bundled-skill guard, the unreadable-diff fail-safe, and the wiring test). The eighth pins behavior that must not change, so it passes both ways by design.
- The CRLF fix was reproduced red before it was fixed: converting a local copy of the workflow to CRLF makes the pre-fix test emit the exact CI failure, and the fixed test pass on the same file.

## Links

- Relates to #2897

## Summary by CodeRabbit

* **CI Improvements**
  * Optimized test execution by skipping unnecessary runs for documentation-only updates while still testing configuration and build-related changes.
  * Improved reliability for new branches, workflow-triggered runs, and incomplete change information by defaulting to running tests when needed.
  * Added dependency caching for Windows runners to improve build efficiency.
  * Added an aggregate test-suite status for clearer CI results.
  * Expanded automated coverage for test-selection logic and workflow behavior.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,27 +4,38 @@ on:
   workflow_dispatch:
   push:
     branches: [main, dev]
-    paths:
-      - '**'
-      - '!**.md'
-      - '!packages/docs-web/**'
-      # Docker's dependency layer copies this workspace manifest.
-      - 'packages/docs-web/package.json'
   pull_request:
     branches: [main, dev]
-    paths:
-      - '**'
-      - '!**.md'
-      - '!packages/docs-web/**'
-      # Docker's dependency layer copies this workspace manifest.
-      - 'packages/docs-web/package.json'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      run-tests: ${{ steps.decision.outputs.run-tests }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Decide whether tests are needed
+        id: decision
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.before || github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.after || github.event.pull_request.head.sha }}
+        run: |
+          if [ "$EVENT_NAME" != workflow_dispatch ]; then
+            git fetch --no-tags --depth=1 origin "$BASE_SHA" "$HEAD_SHA"
+          fi
+          echo "run-tests=$(bun scripts/should-run-test-suite.ts "$EVENT_NAME" "$BASE_SHA" "$HEAD_SHA")" >> "$GITHUB_OUTPUT"
+
   test:
+    needs: changes
+    if: needs.changes.outputs.run-tests == 'true'
     strategy:
       # Both legs must always report. With the default fail-fast, an ubuntu
       # failure cancels windows, so a windows-only break stays invisible until
@@ -88,6 +99,8 @@ jobs:
         run: bun run test
 
   schema-upgrade:
+    needs: changes
+    if: needs.changes.outputs.run-tests == 'true'
     # Nothing else in CI applies migrations/000_combined.sql to a real database,
     # and every test that touches a schema starts from an empty one. That blind
     # spot is exactly how #2508 shipped green and then crash-looped every
@@ -145,6 +158,8 @@ jobs:
         run: bun run check:sqlite-vintages
 
   postgres-parity:
+    needs: changes
+    if: needs.changes.outputs.run-tests == 'true'
     # The unit suite mocks the pg driver, so the Postgres dialect branch of
     # getLiveRunOwningEnv (cleanup's lock query) only executes here, against a
     # real server: an untyped parameter compared against text and UUID columns
@@ -183,6 +198,8 @@ jobs:
         run: bun test packages/core/src/db/isolation-environments.live-run.postgres.integration.test.ts
 
   docker-build:
+    needs: changes
+    if: needs.changes.outputs.run-tests == 'true'
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -227,3 +244,29 @@ jobs:
       - name: Cleanup smoke test container
         if: always()
         run: docker rm -f archon-smoke || true
+
+  test-suite:
+    name: Test Suite
+    needs: [changes, test, schema-upgrade, postgres-parity, docker-build]
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Report test-suite outcome
+        env:
+          RUN_TESTS: ${{ needs.changes.outputs.run-tests }}
+          TEST: ${{ needs.test.result }}
+          SCHEMA_UPGRADE: ${{ needs.schema-upgrade.result }}
+          POSTGRES_PARITY: ${{ needs.postgres-parity.result }}
+          DOCKER_BUILD: ${{ needs.docker-build.result }}
+        run: |
+          if [ "$RUN_TESTS" = false ]; then
+            echo "Skipped tests for documentation-only changes."
+            exit 0
+          fi
+
+          for result in "$TEST" "$SCHEMA_UPGRADE" "$POSTGRES_PARITY" "$DOCKER_BUILD"; do
+            if [ "$result" != success ]; then
+              echo "A test-suite job finished with $result."
+              exit 1
+            fi
+          done
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -209,8 +209,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # The Buildx cache can miss independently of changed paths. Rebuilding the
-      # base or dependency layers then exceeds the stock runner's available space.
+      # The image build needs ~14GB more scratch space than a stock runner has
+      # free once the provider SDKs' multi-platform vendor binaries are in
+      # node_modules. Cached-layer builds squeak by; any PR that invalidates an
+      # early Dockerfile layer (apt block, base image) rebuilds everything and
+      # hits "no space left on device". Reclaim the big unused toolchains first.
       - name: Free runner disk space
         run: |
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -209,11 +209,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # The image build needs ~14GB more scratch space than a stock runner has
-      # free once the provider SDKs' multi-platform vendor binaries are in
-      # node_modules. Cached-layer builds squeak by; any PR that invalidates an
-      # early Dockerfile layer (apt block, base image) rebuilds everything and
-      # hits "no space left on device". Reclaim the big unused toolchains first.
+      # The Buildx cache can miss independently of changed paths. Rebuilding the
+      # base or dependency layers then exceeds the stock runner's available space.
       - name: Free runner disk space
         run: |
           sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,8 +4,20 @@ on:
   workflow_dispatch:
   push:
     branches: [main, dev]
+    paths:
+      - '**'
+      - '!**.md'
+      - '!packages/docs-web/**'
+      # Docker's dependency layer copies this workspace manifest.
+      - 'packages/docs-web/package.json'
   pull_request:
     branches: [main, dev]
+    paths:
+      - '**'
+      - '!**.md'
+      - '!packages/docs-web/**'
+      # Docker's dependency layer copies this workspace manifest.
+      - 'packages/docs-web/package.json'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -29,6 +41,13 @@ jobs:
         uses: oven-sh/setup-bun@v2
         with:
           bun-version: 1.3.11
+
+      - name: Cache Bun dependencies
+        if: runner.os == 'Windows'
+        uses: actions/cache@v4
+        with:
+          path: ~/.bun/install/cache
+          key: ${{ runner.os }}-bun-${{ hashFiles('bun.lock') }}
 
       - name: Install dependencies
         run: bun install --frozen-lockfile
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 1
+          fetch-depth: 0
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.11
 
       - name: Decide whether tests are needed
         id: decision
@@ -28,9 +33,6 @@ jobs:
           BASE_SHA: ${{ github.event.before || github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.after || github.event.pull_request.head.sha }}
         run: |
-          if [ "$EVENT_NAME" != workflow_dispatch ]; then
-            git fetch --no-tags --depth=1 origin "$BASE_SHA" "$HEAD_SHA"
-          fi
           echo "run-tests=$(bun scripts/should-run-test-suite.ts "$EVENT_NAME" "$BASE_SHA" "$HEAD_SHA")" >> "$GITHUB_OUTPUT"
 
   test:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,8 +32,12 @@ jobs:
           EVENT_NAME: ${{ github.event_name }}
           BASE_SHA: ${{ github.event.before || github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.after || github.event.pull_request.head.sha }}
+        # Assign before echoing. A failing command substituted into another command's arguments
+        # does not trip `set -e`, so `echo "run-tests=$(...)"` would report success and write an
+        # empty decision that every downstream `== 'true'` gate reads as "skip".
         run: |
-          echo "run-tests=$(bun scripts/should-run-test-suite.ts "$EVENT_NAME" "$BASE_SHA" "$HEAD_SHA")" >> "$GITHUB_OUTPUT"
+          run_tests=$(bun scripts/should-run-test-suite.ts "$EVENT_NAME" "$BASE_SHA" "$HEAD_SHA")
+          echo "run-tests=$run_tests" >> "$GITHUB_OUTPUT"
 
   test:
     needs: changes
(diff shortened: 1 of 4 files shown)


Wirasm:
# Review report

## Verdict

**Ready — none.** The correction restores the Docker disk-free step's exact required comment; all previously blocking findings are fixed and the correction delta introduces no defect.

## Accepted contract

Documentation- and Markdown-only changes must skip the test workflow while changes that can affect test or Docker results still run it. The Windows leg must restore Bun's package cache keyed by `bun.lock`. The Docker disk-free step, including its command, name, surrounding configuration, and explanatory comment, must remain unchanged. Cleanup-relocation commits are excluded, and the per-OS type-check/lint matrix and runner selection are unchanged.

## Reviewed head SHA

`ca1462209f92a0496a77587e253684cec289ea6e`

## Findings

### Rejected findings

- The raw producer discovery, “Cleanup drift baseline rejects the current dev tree,” is rejected. Its cited line is now `describe(...)` at `packages/workflows/src/defaults/discovery-report.test.ts:71`; the temporary directory is registered via `trackTempRoot` at line 66, and `bun run scripts/check-test-cleanup-drift.ts` passes. It is not a proved adjacent discovery.

### Suggestions

- None.

## Prior findings

| ID | Prior claim | Verdict at `ca1462209f92a0496a77587e253684cec289ea6e` | Evidence |
| --- | --- | --- | --- |
| R1 | Workflow-level path filtering can hide a required manifest change. | **Fixed.** | No workflow `paths` filter is present at `.github/workflows/test.yml:3-9`; the complete-diff decision remains wired at lines 15-36. `scripts/should-run-test-suite.test.ts:8-65` passes its manifest-after-3,001-doc-files case. |
| R2 | The trigger contract has no behavioral regression test. | **Fixed.** | `scripts/should-run-test-suite.test.ts:8-65` covers Markdown/docs skips, manifest inclusion, large diffs, branch creation, and automatic-route wiring. `bun test ./scripts/should-run-test-suite.test.ts` passed (9 tests), as did `bun x tsc --noEmit -p scripts/tsconfig.json`. |
| R3 | Docker disk-free step changed outside the accepted scope. | **Fixed.** | `.github/workflows/test.yml:212-221` exactly matches the required comment and unchanged step in `c4a3d6d4578b26d782a63235ed66146a4f9cf1cc`; `git diff c4a3d6d4578b26d782a63235ed66146a4f9cf1cc..ca1462209f92a0496a77587e253684cec289ea6e -- .github/workflows/test.yml` is empty. |

## Discoveries

0 accepted discoveries. [JSON](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/49f0d462-bd8d-4ccb-852b-77919f9d4b8a/discoveries.json) · [report](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/49f0d462-bd8d-4ccb-852b-77919f9d4b8a/discoveries.md). Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and correction delta. The original accepted review covered `code`, `seams`, `tests`, `comments`, and `simplify`; `errors`, `types`, and `docs` were disabled. I inspected the exact recorded delta, the full relevant workflow context, the decision script and focused tests, the required Docker-step baseline, and the producer discovery's cited cleanup path. I ran the focused test suite, its TypeScript check, `git diff --check`, and the cleanup-drift check. No required evidence was unavailable.

Wirasm:
## Ready to merge

Correction verification against the diff from `a6a912fd` (previously reviewed) to `26ecdb9e` (current). Both blocking findings are fixed, verified independently rather than taken on the fix report's word: I reproduced the original crash-swallow mechanism against the new split-assignment workflow step and confirmed it now either resolves safely (`run-tests=true` written, cause on stderr) or fails the job outright for a genuine wiring bug — and I reverted the two changed files to their pre-fix content inside this review's own worktree and reran the new test file, getting the same 10 pass / 7 fail the fix report claims, before restoring. R4 is fixed and independently confirmed. R3 is declined by explicit operator decision, not by my own judgment. Two minor, non-blocking accuracy notes surfaced in the correction itself (R5) — worth a look, not worth blocking.

**Validation:** `bun install --frozen-lockfile`, `bun test ./scripts/` (141 pass, 0 fail), `bun run lint`, `bun x tsc --noEmit -p scripts/tsconfig.json`, `bun x prettier --check` all pass at the reviewed head. Remote CI independently confirmed green: run 33165582920 at `26ecdb9e`, all 7 jobs including both `test` matrix legs and the `Test Suite` summary job report `success` (checked directly via `gh run view --json`, not taken from the PR body).

**Resolved:** R1, R2, R4

### Findings

|---|---|---|---|
| `R1` | Important | A crash in the decision script degraded to an empty, ambiguous `run-tests` output that was safe only by accident, untested, and buried the real error in a job that reported green | FIXED |
| `R2` | Important | The docs-only `.md` heuristic swept in bundled/generated content, letting a change to it silently skip the checks that exist specifically to catch that drift | FIXED |
| `R3` | Suggestion | The `test-suite` aggregator's bash `if`/`for` loop could be one inline GitHub Actions `if:` expression | DECLINED |
| `R4` | Suggestion | The workflow-wiring test asserted fragments present somewhere in the job text rather than pinning the actual invocation line | FIXED |
| `R5` | Suggestion | Two minor accuracy/disclosure gaps introduced by the correction itself: the `.claude/skills/` widening's actual cost is understated, and the characterization test lacks an in-code note explaining it passes unchanged before and after the fix | OPEN |

<details>
<summary><code>R1</code> — decision-script crash is fail-safe only by accident</summary>

**Impact (original):** An uncaught exception in `should-run-test-suite.ts` (e.g. `git diff` against an unreachable base SHA — exactly what a force-push produces) degraded to an empty `run-tests` output that was safe only by an accidental interaction between two independently-written shell comparisons, untested, and one plausible edit away from an actual silent skip.

**Fix, verified independently, both halves:**

1. `scripts/should-run-test-suite.ts:52-68` (new `decideFromDiff`) now catches the scan failure and resolves it explicitly to `true`, logging the cause to stderr. Reproduced directly: `bun scripts/should-run-test-suite.ts push 111...111 26ecdb9e` → stdout `true`, exit 0, stderr `Could not compare 111...111..26ecdb9e, so the test suite runs: Could not read changed files: fatal: bad object 111...111`.
2. `.github/workflows/test.yml:35-38` now assigns before echoing (`run_tests=$(bun scripts/should-run-test-suite.ts ...)` then `echo "run-tests=$run_tests" >> "$GITHUB_OUTPUT"`), so a genuinely fatal wiring bug (a bad event name, not a scan failure — `should-run-test-suite.ts:76-78` still throws uncaught for this, deliberately) now fails the step. I reproduced the exact original swallow mechanism against **this new step form**, line for line, in `bash --noprofile --norc -eo pipefail`:
   - Unreachable-SHA scenario: step succeeds, writes `run-tests=true` to `$GITHUB_OUTPUT` (previously wrote `run-tests=` empty).
   - Bad-event-name scenario (genuine wiring bug): bash wrapper now exits 1, nothing is written to `$GITHUB_OUTPUT` — confirms the `changes` job itself now goes red for this case, unlike before.

**Test coverage:** `scripts/should-run-test-suite.test.ts` adds `'an unreadable diff resolves to running the suite, with the cause on stderr'` (asserts exit 0, stdout `true`, stderr contains the cause) and `'a bad argument stays fatal, so the step cannot publish an empty decision'` (asserts exit != 0, empty stdout). I reverted both changed files to their pre-fix content inside this review's own worktree and reran the test file: **10 pass / 7 fail**, matching the fix report's claim exactly, then restored the fixed files (`git status`/`git diff --stat` confirmed clean afterward). The unreadable-diff test was among the 7 failures against pre-fix code; the bad-argument test was **not** — it passes unchanged both before and after, which is correct, since that half of the contract (a wiring bug must stay fatal) was never meant to change.

**Disposition:** FIXED. Verified via direct reproduction of the original mechanism against the new code, and via an independent red/green rerun of the new tests against the pre-fix files.

</details>

<details>
<summary><code>R2</code> — docs-only classification bypasses bundled/generated-content drift checks</summary>

**Impact (original):** The bare `.endsWith('.md')` classification conflated passive documentation with bundled/generated `.md` content (`.archon/commands/defaults/`, `.archon/workflows/**/commands/`, `.claude/skills/archon-cli/`, the generated `provider-capabilities.md`), letting an edit to any of those cleanly skip the checks built to catch that drift.

**Fix:** `scripts/should-run-test-suite.ts:1-27` replaces the single `DOCS_MANIFEST` exception with a `BUILD_INPUTS` list (`.archon/commands/`, `.archon/workflows/`, `.claude/skills/`, `provider-capabilities.md`, the docs manifest), each entry annotated in a comment with the check that reads it, checked via `isBuildInput()`. Verified directly: `shouldRunTestSuite(['.archon/commands/defaults/archon-assist.md'])`, `shouldRunTestSuite(['.claude/skills/archon-cli/SKILL.md'])`, and `shouldRunTestSuite(['packages/docs-web/src/content/docs/reference/provider-capabilities.md'])` all now return `true`; genuine prose (`README.md`, `AGENTS.md`, a docs-site guide) still returns `false`.

**Widening claims, checked independently rather than accepted:**
- `.archon/workflows/` (not just `**/commands/`): I confirmed by direct search that `.archon/workflows/rasmus-tests/README.md` is genuinely the only Markdown file under that tree that is not a packaged command prompt (21 total `.md` files, 1 non-command). The stated one-file cost is accurate.
- `.claude/skills/` (not just `archon-cli/`): confirmed `packages/cli/src/bundled-skill.ts` text-imports exactly 9 files, all under `.claude/skills/archon-cli/`. But 12 other, unrelated skill directories also live under `.claude/skills/` (`agent-browser`, `archon-dev` [11 files], `docker-extend`, `playwright-cli` [8], `release`, `remotion-best-practices` [38], `replicate-issue`, `rulecheck` [3], `save-task-list` [2], `test-release`, `triage` [2], `validate-ui`) — roughly 70 files across 12 directories with no relationship to `check:bundled-skill` or any other check, none of which are read by any script (confirmed via `grep -rln "\.md" scripts/*.ts`, plus a repo-wide check that `.claude/commands/`, `.claude/agents/`, `.github/prompts/`, `.claude/PRPs/` are read by no check at all). This is a real, safe widening — it fails toward running the suite, which is the correct direction — but it is a materially larger sacrifice of the fast path than the "costs one README" framing in the fix report and PR body implies. See R5.
- Sweep claim ("only markdown readers under `scripts/` are the four excluded families") independently confirmed: `generate-bundled-defaults.ts`, `check-bundled-skill.ts`, `generate-capability-matrix.ts`, and `should-run-test-suite.ts` itself are the only scripts reading `.md` content; the one other grep hit (`check-schema-upgrades.ts`) references `AGENTS.md` only in a comment, not a read.

**Drift guard, verified reads the actual generator source:** `scripts/should-run-test-suite.test.ts:56-70` (`'every Markdown file compiled into the CLI skill runs the suite'`) `readFileSync`s `packages/cli/src/bundled-skill.ts` and regex-extracts its text-import paths, asserts the extracted list is non-empty, then asserts each extracted file forces a run — this does read the generator source directly rather than hardcoding the list, and would fail the day a tenth skill file is added without updating the filter, as claimed.

**Disposition:** FIXED. Widenings adjudicated: safe and correctly fail-safe-directed; the `.claude/skills/` cost is larger than the PR's own framing suggests, which is R5, not a defect in R2 itself.

</details>

<details>
<summary><code>R3</code> — test-suite aggregator could be a single GitHub Actions expression</summary>

**Disposition:** DECLINED, by explicit operator decision (relayed by the coordinator): the bash aggregator is correct and already validated by a green CI run; rewriting it trades a working, proven shape for a stylistic alternative with no defect behind it. No independent action needed from this review.

</details>

<details>
<summary><code>R4</code> — workflow-wiring test only checked fragment presence, not the actual invocation</summary>

**Fix, verified:** `scripts/should-run-test-suite.test.ts:135-141` now asserts the changes-job text contains the literal `run_tests=$(bun scripts/should-run-test-suite.ts "$EVENT_NAME" "$BASE_SHA" "$HEAD_SHA")\n` line and the separate `echo "run-tests=$run_tests" >> "$GITHUB_OUTPUT"` line — this pins both the argument order and the split-assignment form that R1's fix depends on. Confirmed as one of the 7 failures when rerun against the pre-fix workflow file (which still had the single-line `echo "run-tests=$(...)"` form).

**Disposition:** FIXED.

</details>

<details>
<summary><code>R5</code> — two minor accuracy/disclosure gaps in the correction itself</summary>

**Impact:** Neither is a functional defect; both are about how clearly the correction explains itself to the next reader.

1. The fix report and PR body describe both path widenings (`.archon/workflows/` and `.claude/skills/`) together as costing "one README," citing only the `.archon/workflows/rasmus-tests/README.md` example. That figure is accurate for the `.archon/workflows/` widening alone, but the `.claude/skills/` widening sweeps in roughly 70 unrelated files across 12 other skill directories with no relationship to any check (see R2 evidence). The tradeoff itself is sound and correctly fail-safe-directed, but a maintainer reading "costs one README" would not expect editing `.claude/skills/remotion-best-practices/*.md` to now force the full matrix and Docker build.
2. `scripts/should-run-test-suite.test.ts`'s `'a bad argument stays fatal...'` test has no in-code comment disclosing that it is a characterization test which passes unchanged both before and after this fix (unlike its neighboring `'an unreadable diff resolves to...'` test, which does have one). The fix report discloses this clearly in prose; the test file itself does not.

**Evidence:** `scripts/should-run-test-suite.ts:1-27` (BUILD_INPUTS list and comment); `.claude/skills/*` directory listing; `scripts/should-run-test-suite.test.ts` (the undocumented characterization test, no comment block above it).

**Required outcome:** Optional. If addressed: correct the cost description to name the actual tradeoff (or narrow `.claude/skills/` back toward `archon-cli/` if the wider exclusion wasn't intentional at this scope), and/or add a one-line comment to the characterization test noting it intentionally passes both before and after.

**Found by:** `pr2903-review` (correction-verification pass)

**Disposition:** OPEN — non-blocking; author's call whether to tighten the description or the exclusion.

</details>

<details>

#### Validation

|---|---|---|
| `bun install --frozen-lockfile` (fresh review worktree) | PASS | 2556 packages, clean |
| `bun test ./scripts/` | PASS | 141 pass, 0 fail (was 133 at `a6a912fd`; 8 new tests) |
| `bun run lint` | PASS | clean |
| `bun x tsc --noEmit -p scripts/tsconfig.json` | PASS | no diagnostics |
| `bun x prettier --check` (workflow + both changed scripts) | PASS | clean |
| Direct reproduction of R1's original mechanism against the new workflow step form | PASS | see R1 detail |
| New test file rerun against pre-fix `should-run-test-suite.ts`/`test.yml` inside the review worktree, then restored | 10 pass / 7 fail | matches fix report exactly; worktree left clean afterward |
| `gh run view 33165582920 --json status,conclusion,headSha,jobs` | PASS | independently confirmed `success` for all 7 jobs at `26ecdb9e`, including both matrix legs and `Test Suite` |
| PR body vs. implementation | Consistent | Solution/Behavior-change/Windows-fixes/Review-corrections sections all match the current diff; no longer describes the old bare-`.md` rule |

</details>

PR #2887, opened 2026-08-28
feat(sdlc): surface validated run discoveries in terminal reports

Wirasm:
## Problem and outcome

Every SDLC run records validated out-of-scope findings in `discoveries.json` and `discoveries.md`, but the terminal report an orchestrating agent actually reads through `workflow get` never mentioned them. The dev-breaking `rmSync` regression sat in a run's discovery sidecar while the same break was independently rediscovered and hotfixed by hand (#2880).

- **Outcome:** all four SDLC tails — `deliver`, `ship`, `stabilize`, `upkeep` — append a discoveries section to their terminal report when the run recorded any: the count, one-line titles, the path to `discoveries.md`, and an agent-directed relay instruction. With no sidecar or no recorded discoveries, the report is byte-identical to before.
- **Invariant:** presentation only. Discoveries still never create issues and never gate readiness.
- **Scope boundary:** terminal report composition. Review-side discovery collection is unchanged; `review-synthesize` only gains the same relay wording in the report it already wrote.

## Review guidance

- **Feedback requested:** whether the deliver tail's new terminal node is the right seam, and whether the degraded path for an unreadable sidecar is the right call.
- **Start here:** `.archon/workflows/sdlc/deliver/scripts/outcome.py` — the new node, and the canonical copy of the shared `format_discoveries` helper.
- **Review order:**
  1. `.archon/workflows/sdlc/deliver/scripts/outcome.py` and the `flip-ready` / `outcome` nodes in `.archon/workflows/sdlc/deliver/archon-deliver.yaml`
  2. `.archon/workflows/sdlc/{ship,stabilize,upkeep}/scripts/outcome.py`
  3. `packages/workflows/src/defaults/discovery-report.test.ts`
  4. the fixture stubs for the new node
- **Lower-attention areas:** `packages/workflows/src/defaults/bundled-defaults.generated.ts` (generated artifact).
- **Known risk or uncertainty:** the helper exists as four identical copies. A packaged script is materialized standalone and has no import channel, so a test asserts the copies stay byte-identical rather than letting them drift.

## Solution

[...]

Issue #2745, opened 2026-08-22
feat(core): wake a waiting host when a run needs someone


Wirasm:
## Problem

A run started outside a chat thread — CLI `--detach`, the HTTP API, a future scheduled trigger (#998) — reaches a state it will not leave on its own, and nothing tells whoever launched it. The knowledge is durable in the database; the delivery does not exist.

The common case is an agent delegating to Archon: a human talks to an agent, the agent launches `archon workflow run --detach`, and the launch returns successfully. The agent then has no signal when the run completes, fails, is cancelled, or parks on an approval gate. It must poll `workflow get`. The detaching parent watches its child for a 500 ms startup window, removes its listeners, `unref()`s, and returns; the workflow event emitter and Web event bridge are process-local; the cross-process dashboard poller targets the dashboard stream, not the launching agent.

The result is durable terminal knowledge without terminal delivery. For unattended automation that is a governance gap: an approval gate nobody learns about is operationally indistinguishable from a hang, and a failed overnight run that reports to nobody delays exactly the work the automation was supposed to absorb.

## Required outcome

The engine owns one durable, typed answer to **"what does this run need from outside, if anything"**, and a pluggable host consumes it.

- A run that reaches `completed`, `failed`, or `cancelled`, or parks on an unresolved human gate, produces exactly that answer for any host waiting on it, without the host polling run status.
- The answer is small: the run id, why it stopped, and — for a gate — the run and node where the decision must be made. After receiving it, the host inspects the durable run for authored outcome, errors, artifacts, and node detail. Polling remains a diagnostic, not the orchestration contract.
- No waiter ever mutates or synthesizes run lifecycle state. A wait that ends on its own clock reports what it observed; it never invents a terminal status.

This follows #2791: the engine owns the capability and exposes it to pluggable hosts; each host owns waking its own agent or human. The first host is a CLI verb an orchestration agent can block on. Per-user platform routing is a second host and has its own issue.

## Ownership note

This decision is not new to the codebase — it is currently made in four places, in four different shapes, each rendering the same conclusion as a string:

| Site | Shape |
|---|---|
| `packages/core/src/operations/workflow-operations.ts:240` `assertApprovable` | thrown `Error` |
| `packages/core/src/operations/workflow-operations.ts:299` `assertRejectable` | thrown `Error` |
| `packages/server/src/routes/api.ts:3838-3864` | `apiError(400, …)` |
[...]

Issue #1216, opened 2026-04-14
bug: server startup marks actively-running workflows as failed via failOrphanedRuns()


Wirasm:
## Summary

Every `archon serve` startup unconditionally flips **all** workflow runs in `status='running'` to `status='failed'` via `failOrphanedRuns()` — including runs that are actively executing in another process (typically `archon workflow run` from a separate shell). The dag-executor's between-layer status check then bails out of the active run, exit code 1, even though every node completed successfully.

This is the same class of bug already fixed for the CLI side (see comment at `packages/cli/src/cli.ts:256-258`) — the server hasn't been given the same treatment.

## Reproducer

```bash
# Terminal A — start a Claude workflow that takes ~10s
archon workflow run e2e-claude-smoke --no-worktree

# Terminal B — while A is still running
archon serve     # or `bun run dev:server` from source
```

Terminal A:

- All nodes complete successfully (visible in stdout)
- Workflow exits non-zero with `Workflow failed: Workflow did not complete successfully`
- DB row: `status='failed'`, `metadata.failure_reason='server_restart'`

DB also retains the per-node completion events, so the corruption is *partial* — node-level state survives, run-level status is wrong.

## Causal chain (verified end-to-end during smoke testing)

1. CLI dispatches workflow → `createWorkflowRun()` → row written with `status='running'`
2. Server starts in Terminal B → `index.ts:212` calls `failOrphanedRuns()`
3. `failOrphanedRuns()` runs:
   ```sql
   UPDATE remote_agent_workflow_runs
   SET status = 'failed',
       completed_at = NOW(),
       metadata = jsonb_merge(metadata, '{\"failure_reason\":\"server_restart\"}')
   WHERE status = 'running'
   ```
   No PID filter, no liveness check, no scope filter. **All running rows.**
4. Server log emits `db.orphaned_workflow_runs_failed` with the count
5. Dag-executor's between-layer check (`dag-executor.ts:2622-2645`) calls `getWorkflowRunStatus()` after the next node finishes, sees `status='failed'`, logs `dag.stop_detected_between_layers`, and breaks out of the loop
6. Post-loop guard at `dag-executor.ts:2727` (`dag.skip_complete_status_changed`) blocks the `completeWorkflowRun()` call because status is no longer `running`
7. CLI receives `success: false` → `executor.ts:756` returns the generic error → `workflow.ts:615` throws → exit 1

[...]

PR #2912, opened 2026-08-28
fix(core): a rejected gate records its cancellation in the event log

### LINKED ISSUE #2906: fix(core): two terminal transitions write no terminal lifecycle event
## Problem

Two paths write a run's status to a terminal value without writing the matching `workflow_*` lifecycle event, so the durable event log is not a complete record of terminal transitions.

**1. Rejecting a gate to termination leaves no `workflow_cancelled` row.**

`resolveAndCancelApprovalGate` (`packages/core/src/db/workflows.ts:291`) flips the run `paused → 'cancelled'` in one CAS and inserts only the caller-supplied events, which at the single call site (`packages/core/src/operations/workflow-operations.ts:809`) is `[rejectionEvent]` — an `approval_received` row. No `workflow_cancelled` event is ever written.

Confirmed across all four callers of `rejectWorkflow`; none emits one afterwards:

- `packages/cli/src/commands/workflow.ts:4089`
- `packages/server/src/routes/api.ts:3962`
- `packages/adapters/src/chat/slack/workflow-bridge.ts:482`
- `packages/core/src/handlers/command-handler.ts:1023`

Compare the sibling paths, which all do write their terminal event inside the same transaction: `completeWorkflowRun:1168`, `failWorkflowRun:1220`, `cancelWorkflowRun:1281`, `cancelFanOutRun:1327`, `cancelResumableRunsForConversation:577`.

Consequence: any consumer that derives terminal state from the event log rather than the run row silently misses a whole class of cancellation. The run row is correct; the audit trail is not.

**2. `failOrphanedRuns` writes a terminal status with no event — and has no live caller.**

`failOrphanedRuns` (`packages/core/src/db/workflows.ts:2053`) transitions every `running` row to `failed` with `failure_reason: 'server_restart'` and inserts nothing. Its doc comment says it is "Called on server startup", but it is not: the only non-test references are the `IWorkflowStore` member (`packages/workflows/src/store.ts:256`) and its wiring (`packages/core/src/workflows/store-adapter.ts:69`). Nothing invokes it.

That is also the more dangerous half if it were ever re-enabled — it would fail every running run install-wide, including runs owned by live processes it cannot see, which is exactly the ownership rule in `AGENTS.md` ("a process must not mark non-terminal work failed ... when it cannot distinguish a live owner from an orphaned process").

## Desired outcome

- A terminal status transition and its terminal lifecycle event are written together, on every path. `resolveAndCancelApprovalGate` writes `workflow_cancelled` alongside the rejection event, in the same transaction.
- `failOrphanedRuns` is deleted rather than fixed — the function, its `IWorkflowStore` member, its store-adapter wiring, and its tests. It is dead machinery whose only possible behavior violates the lifecycle-ownership rule.
- A test proves that rejecting a gate to termination leaves a `workflow_cancelled` row, so the next consumer that reads the event log gets the truth.

## Why now

Found while designing #2745. That capability derives from the run row precisely because the event log cannot currently be trusted for terminal transitions, so it is not blocked by this. But the notification-routing host that follows #2745 will want to consume events, and an event log that omits terminal transitions is a false foundation to build that on.

Wirasm:
## Problem and outcome

Two paths wrote a run's status to a terminal value without writing the matching lifecycle event, so the durable event log was not a complete record of terminal transitions. Rejecting an approval gate to termination flipped the run row to `cancelled` while inserting only the caller's gate events — one `approval_received` row — so anything deriving terminal state from events rather than from the run row silently missed an entire class of cancellation.

- **Outcome:** A reject-to-cancel now leaves a `workflow_cancelled` row, written in the same transaction as the status flip. `failOrphanedRuns`, which wrote terminal `failed` with no event and had no live caller, is gone.
- **Invariant:** A terminal status transition and its terminal lifecycle event are written together, in one transaction, on every path.
- **Scope boundary:** Only the two sites named in the issue. The event log was not widened for #2745, which derives run state from the run row and is not blocked by this.
- **Root cause:** `resolveAndCancelApprovalGate` inserted only `events`, which it received from its caller. Every sibling terminal writer owns its own lifecycle event; this one delegated that to caller discipline, and the caller only knew about its gate events.

## Review guidance

- **Feedback requested:** Correctness of the event's shape and the placement of ownership at the DB layer rather than the caller.
- **Start here:** `packages/core/src/db/workflows.ts:291` — the terminal write and its event now commit together.
- **Review order:** `packages/core/src/db/workflows.ts` (both halves), then `packages/core/src/operations/workflow-operations.ts:806`, then the regression assertions in `packages/core/src/db/workflows.resume-cas.integration.test.ts:574`.
- **Lower-attention areas:** The call-argument assertion updates in core, CLI, and server tests, and the six `IWorkflowStore` mock deletions — mechanical, verified by the package test scripts listed below.
- **Known risk or uncertainty:** None. The reject-to-cancel path now emits one more DB event row per rejected-to-termination gate; the dashboard poller maps it to a `workflow_status: cancelled` SSE, which is what the run row already said. `recoverCancelledFanOutRun` deletes `workflow_cancelled` rows only for the three fan-out reason tokens, so `approval_rejected` is not swept up.

## Solution

`resolveAndCancelApprovalGate` writes `workflow_cancelled` itself rather than trusting the caller to pass it in `events`. Mechanism belongs to the function that owns the transaction; policy — which gate ended the run, and why — comes in through a required third parameter typed as the existing `WorkflowCancellationEventDetails`, the same shape `cancelWorkflowRun` already takes for this same event. The caller's gate events go in first, since the decision precedes the termination it caused.

The reason is a stable `approval_rejected` token, not the user's rejection prose. That prose is already on the `approval_received` row and does not belong on two.

`failOrphanedRuns` is deleted rather than fixed. It had no live caller — only the `IWorkflowStore` member and its store-adapter wiring referenced it — and its "Called on server startup" doc comment has been false since #1216 removed that call (`CHANGELOG.md:545`). Its only possible behavior, flipping every `running` row install-wide to `failed`, is exactly the lifecycle-ownership violation `AGENTS.md` names: a process must not mark non-terminal work failed when it cannot distinguish a live owner from an orphaned process. Adding an event insert would have preserved that hazard behind a better audit trail.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Rejecting a gate to termination wrote `status='cancelled'` and one `approval_received` event. No `workflow_cancelled` row existed anywhere | The same rejection writes `status='cancelled'`, the `approval_received` event, and a `workflow_cancelled` event carrying the gate's `step_name` and `reason: approval_rejected` |
| **Failure behavior** | A failed event write already rolled the cancellation back (#2146) | Unchanged — the terminal event rides the same transaction, so a failed write still rolls back rather than terminating the run with an incomplete audit trail |

### Changed seams

|---|---|---|
| `rejectWorkflow → resolveAndCancelApprovalGate` | New required third parameter carrying the terminal event's detail | `packages/core/src/db/workflows.ts:293`, `packages/core/src/operations/workflow-operations.ts:809` |
| `IWorkflowStore` | `failOrphanedRuns()` member removed | `packages/workflows/src/store.ts:256`, `packages/core/src/workflows/store-adapter.ts:69` |

## Validation

- `bun test src/db/workflows.resume-cas.integration.test.ts` (packages/core) — 42 pass / 0 fail — proves the fix against a real SQLite database end to end: rejecting a gate to termination leaves `status='cancelled'`, one `approval_received` row, and one `workflow_cancelled` row with `step_name='review'` and `data={"reason":"approval_rejected"}`.
- The same file with the `workflow_cancelled` insert temporarily removed — 3 fail / 39 pass — the regression test was observed red against the unfixed code before being observed green.
- `bun test src/db/workflows.test.ts`, `src/operations/workflow-operations.test.ts`, `src/handlers/command-handler.test.ts`, `src/workflows/`, `src/orchestrator/orchestrator-agent.test.ts` (packages/core) — all 0 fail.
- `bun run test` (packages/workflows) — full package script, 0 fail — covers all five `IWorkflowStore` mocks.
- `bun test src/commands/workflow.test.ts` (packages/cli) — 300 pass / 0 fail. `bun test src/routes/api.workflow-runs.test.ts` (packages/server) — 130 pass / 0 fail.
- `bun run type-check` (13 packages), `bun run scripts/lint.ts --max-warnings 0`, `bun run format:check`, `bun run test:install`, `bun run test` — all pass.
- **Not verified:** Nothing material for this change. `bun run validate` currently stops at `check-test-cleanup-drift`, which rejects `packages/workflows/src/defaults/discovery-report.test.ts:71`. Neither that file nor the check script is in this diff, and `dev`'s own Test Suite run for the #2887 merge is already `failure` for the same reason. Every other validate stage passes here.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. No schema change; one additional event row on an existing path | `packages/core/src/db/workflows.ts:320` |
| Observability | A reject-to-cancel is now visible to event-log consumers, including the dashboard poller's `workflow_status` mapping | `packages/server/src/adapters/web/workflow-bridge.ts:223` |

## Links

- Closes #2906
- Related #2745

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow cancellations caused by rejected approval gates now record a `workflow_cancelled` event with the affected step and `approval_rejected` reason.
  * Rejection details remain available on the approval event while cancellation records use consistent metadata.
  * Prevented duplicate cancellation events and approval resolution after a workflow has been cancelled.

* **Refactor**
  * Removed automatic startup handling for orphaned running workflows.


Wirasm:
## Ready to merge

`resolveAndCancelApprovalGate` now writes its own `workflow_cancelled` event inside the same
transaction as the terminal `status='cancelled'` flip, so a rejected-to-termination approval gate
no longer leaves the durable event log with only an `approval_received` row. I reproduced the bug
directly: reverting the new insert and rerunning `workflows.resume-cas.integration.test.ts` failed
exactly 3 of 42 tests (the claimed RED baseline); restoring it returned 42/42 green. The new
event's construction — `event_type: 'workflow_cancelled'`, `step_name`, and
`data: reason === undefined ? undefined : { reason }` — is a byte-for-byte match of the pattern
`cancelWorkflowRun` already uses for the same event type, reusing the existing
`WorkflowCancellationEventDetails` type rather than inventing a new shape. `failOrphanedRuns` is
deleted cleanly: zero references remain anywhere in the tree (function, `IWorkflowStore` member,
store-adapter wiring, and every mock across `packages/workflows` and `packages/core`), and it had
no live caller (its only invocation was removed from server startup by #1216, per
`CHANGELOG.md:547`). Three independent reviewer agents (code, seams, simplify) returned zero
findings, each verifying the same evidence — atomicity, event-shape consistency, single call site,
and clean deletion — independently.

**Validation:** All applicable checks pass. The only CI red (`test (ubuntu-latest)`) is the
pre-existing, unrelated `check-test-cleanup-drift.ts` failure on
`packages/workflows/src/defaults/discovery-report.test.ts:71`, present on `dev` since the #2887
merge (`bfeb6d3f`) and independently confirmed via `gh api repos/coleam00/Archon/actions/jobs/98809961913/logs`.
Not caused by this PR; fixed separately by PR #2911.

### Findings

No findings.

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `simplify` | No additional findings |

#### Validation

|---|---|---|
| `bun test src/db/workflows.resume-cas.integration.test.ts` (packages/core) | PASS | 42/42; independently reverted the new insert and reproduced 3 fail / 39 pass, then restored to confirm 42/42 |
| `bun test src/db/workflows.test.ts` (packages/core) | PASS | 118/118 (run individually per AGENTS.md mock-isolation guidance) |
| `bun test src/operations/workflow-operations.test.ts` (packages/core) | PASS | 68/68 |
| `bun test src/handlers/command-handler.test.ts` (packages/core) | PASS | 144/144 |
| `bun test src/workflows/store-adapter.test.ts` (packages/core) | PASS | 17/17 |
| `bun test src/orchestrator/orchestrator-agent.test.ts` (packages/core) | PASS | 277/277 |
| `bun run test` (packages/workflows) | PASS | 682+491+... across all suite files, 0 fail (covers all five `IWorkflowStore` mocks) |
| `bun test src/commands/workflow.test.ts` (packages/cli) | PASS | 300/300 |
| `bun test src/routes/api.workflow-runs.test.ts` (packages/server) | PASS | 130/130 |
| `bun x tsc --noEmit` (packages/core, packages/workflows, packages/cli, packages/server) | PASS | clean, no errors |
| `bun run lint` (repo root) | FAIL (pre-existing, unrelated) | `check-test-cleanup-drift.ts` flags `packages/workflows/src/defaults/discovery-report.test.ts:71`, a file this PR does not touch; reproduced identically on CI job 98809961913 and locally |
| `gh pr checks 2912` | 1 unrelated fail | `test (ubuntu-latest)` fails on the same drift-check line; `postgres-parity` and `schema-upgrade` pass |

Additional consumer checks performed: `recoverCancelledFanOutRun` (workflows.ts:1040-1080) only
sweeps `workflow_cancelled` rows whose `data.reason` is one of the three `FAN_OUT_CANCEL_REASONS`
tokens, gated on `metadata.cancelled_reason` being set to the same set — the reject-to-cancel path
never sets that metadata key, so the new `approval_rejected`-reasoned row is unreachable from that
sweep. The server dashboard SSE bridge (`workflow-bridge.ts:218-223`) maps `workflow_cancelled` to
the same `status: 'cancelled'` the run row already reports and is refetch-based/idempotent by its
own doc comment, so the additional event row does not introduce a conflicting or duplicated signal.

</details>

Issue #2871, opened 2026-08-27
fix(cli): --adopt rejects the run-id prefixes every other verb accepts


Wirasm:
## Problem

`workflow run <name> --adopt b4e9597b` fails with `Cannot adopt: no workflow run 'b4e9597b' exists. Check the id with 'workflow runs'.` — while `workflow get b4e9597b`, `workflow cancel b4e9597b`, and `workflow abandon b4e9597b` all resolve that same 8-char prefix against the same database. Retrying with the full id succeeded immediately.

This is the #2611 class at a site its fix did not cover: adoption resolves run ids through its own path instead of the shared prefix resolver. The error text even points the user at `workflow runs` — which prints exactly the prefixes adopt then rejects.

## Desired outcome

`--adopt` resolves run-id prefixes through the same resolver as `workflow get`: unique prefix resolves, ambiguous prefix errors listing candidates, unknown id errors as today.

## Invariant

Every CLI surface that accepts a run id accepts the same id forms. A prefix printed by `workflow runs` is valid input to every verb that takes a run id.

## Regression proof

Adopting by unique 8-char prefix works; an ambiguous prefix names the candidates; the existing full-id path is unchanged.

## Evidence

Live incident 2026-08-27 ~20:16Z: adoption of `b4e9597b` (hung-run recovery, issue #2846 dogfood) failed by prefix, succeeded with `b4e9597b-4a84-493c-ad86-32911228c9bd`. Related: #2611 (closed — fixed get, not adopt).

PR #2914, opened 2026-08-28
fix(cli): a detached launch hands back a queryable run id

### LINKED ISSUE #2872: fix(cli): a detached launch reports Started before validating the run — failures vanish into the child log
## Problem

`workflow run archon-deliver --adopt <bad-id> ... --detach` printed `Started 'archon-deliver' in the background.` and returned success. The child process then died on adoption validation (`Cannot adopt: no workflow run ... exists`) and the failure went nowhere: no run row was created, `workflow runs` showed nothing new, and the only trace was an Error line inside the detached child's log file. The operator believes a run is in flight; nothing is.

## Desired outcome

Launch-argument validation that can fail deterministically — adoption resolution, workflow existence, config-layer load, branch/isolation preconditions — runs BEFORE detaching. `Started` must mean a run row exists. Failures print on the launching terminal with a non-zero exit.

## Invariant

A detached launch either produces a queryable run id or a visible synchronous error. No failure mode ends with success output and no run.

## Boundary

This is about pre-flight validation ordering, not about errors that occur after the run legitimately starts — those already belong to the run's own lifecycle and log.

## Regression proof

- `--adopt` with an unknown id: synchronous error, non-zero exit, no 'Started' line.
- A valid detached launch still detaches exactly as fast after passing validation.

## Evidence

Live incident 2026-08-27 ~20:16Z during hung-run recovery for the #2846 dogfood: the silent failure was discovered only by tailing `detached-run-cli-1787861770752-n5fvpx.log`; the operator-facing surface claimed the run had started. Sibling of the prefix defect filed alongside this issue.

Wirasm:
## Problem and outcome

`archon workflow run <name> --adopt <bad-id> ... --detach` printed `Started '<name>' in the background.` and exited 0. The child then died on adoption resolution, no run row was ever created, `workflow runs` showed nothing new, and the only trace was an `Error` line inside `detached-run-<conversation>.log`. The operator believed a run was in flight; nothing was. Separately, the `--detach --json` ack carried no run id at all, so an automated caller had no handle to query either.

- **Outcome:** A detached launch either fails on the launching terminal with a non-zero exit and spawns nothing, or acks a `runId` whose row already exists and is queryable immediately.
- **Invariant:** A detached launch either produces a queryable run id or a visible synchronous error. No failure mode ends with success output and no run.
- **Scope boundary:** The fork point did not move — no worktree, clone, or AI cost happens in the launching process. #2871 (`--adopt` rejects run-id prefixes, the sibling defect from the same incident) is untouched. `archon workflow wait` is #2745 and consumes this.
- **Root cause:** The `--detach` branch (`packages/cli/src/commands/workflow.ts:1647` before this change) returned right after spawning, and `waitForDetachedStartup` only raced the child's exit against a fixed 500 ms window. Every codebase-dependent check — project resolution, adoption resolution, the resumable lookup — lived *after* that branch, in the child. The run row was created later still, inside `executeWorkflow`. A child that died anywhere along that path left no row and no operator-visible trace.

## Review guidance

- **Feedback requested:** Correctness of the run-row ownership handoff, and whether the pre-flight duplicates anything the child will now disagree with.
- **Start here:** `packages/cli/src/commands/workflow.ts` — the `if (options.detach)` branch. Everything between the interactive-class gate and `spawnDetachedWorkflowRun` is the new pre-flight.
- **Review order:**
  1. `packages/cli/src/commands/workflow.ts` — the pre-flight, the row creation, the ack.
  2. `packages/workflows/src/executor.ts` — what a fresh pre-created row still needs stamped, and the `!isContinuation` guard fix.
  3. `packages/workflows/src/store.ts` + `packages/core/src/db/workflows.ts` — `working_path` write-once.
  4. Tests.
- **Lower-attention areas:** `resolveRunCodebase` in `workflow.ts` is a pure extraction of the block that was already there, with one behavior change called out below (it returns a first-time folder registration instead of printing it).
- **Known risk or uncertainty:** A child killed outright (SIGKILL, host crash) before the executor claims its row leaves that row `pending`. Any JS-level failure is now recorded — `workflowRunCommand` fails a still-`pending` handed-over row on the way out — so this is narrowed to abrupt termination. Such a row is inert: with `working_path` NULL it can never hold a path lock (the lock query matches an exact path), can never be resumed (the resume CAS needs a resumable status), and can never be adopted (adoption needs a terminal status). It is a misleading status, not a hazard. Closing it entirely needs a reaper, and lifecycle ownership rules forbid a third party failing work it cannot prove is orphaned.

## Solution

The launching process now runs the checks that can be refused deterministically, then writes the run row and hands its id to the child.

**Pre-flight, before the fork.** The best-effort folder probe becomes the same project resolution the run path performs (extracted into one `resolveRunCodebase` so the two cannot disagree; registration is idempotent, so the child's own call finds what the parent registered). `--adopt` is resolved through `resolveWorkflowAdoption`, which is pure resolution with no filesystem or database mutation — so it costs nothing as a pre-flight, and the child still re-resolves it against the live checkout when it starts. `--supersedes` and the `--resume` continuation lookup get the same treatment.

**The row is written before the fork.** This is the shape `dispatchBackgroundWorkflowOwned` already uses — it pre-creates the row "so the UI can fetch it immediately" (`packages/core/src/orchestrator/orchestrator.ts:572`) — and `ExecuteWorkflowOptions` already names the case: *"Lock-token rows (used by `dispatchBackgroundWorkflow`) supply `preCreatedRun` alone."* The child receives `--internal-detached-run-id`, loads the row, files its source capture under that id, and hands it to `executeWorkflow` as `preCreatedRun` instead of creating a second row.

The one thing the orchestrator never had to solve is that the CLI forks *before* isolation, so the working path is unknown at write time. `working_path` becomes write-once on `updateWorkflowRun`, using the same `COALESCE` idiom `output_root` established, and the executor fills it in on the update it already performs for a fresh pre-created row. A second write is a no-op by construction, so nothing can repoint a live run's checkout.

The row's id is database-generated rather than `preparedSource.runId`: the parent's own capture is reclaimed when it returns, so reusing its reserved id would point the child's capture at a directory the parent is about to `rm`.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `Started '<name>' in the background.` / `Track it with: archon workflow runs` — no id anywhere. `--json` ack had no `runId`. | `Started …` / `Run id: <id>` / `Track it with: archon workflow get <id>`. `--json` ack carries `runId`, and `workflow get <id>` works with no wait. `--resume --detach` acks the continuation's own id. |
| **Failure behavior** | A bad `--adopt`, an unresolvable project, or a missing resumable run printed `Started`, exited 0, and died inside the child log with no row. | The refusal prints on the launching terminal with a non-zero exit and no child is spawned. If the child fails to spawn or dies inside the startup window, the parent fails the row it created with the reason. |

## Architecture

```mermaid
flowchart LR
  subgraph Before
    B1[run --detach] --> B2[fork]
    B2 --> B3[child: resolve project, adoption]
    B3 --> B4[child: worktree]
    B4 --> B5[child: create run row]
    B2 -.-> B6["parent: Started, exit 0"]
  end

  subgraph After
    A1[run --detach] --> A2[parent: resolve project, adoption]
    A2 --> A3[parent: create run row]
    A3 --> A4[fork]
    A4 --> A5[child: adopt row, worktree, fill working_path]
    A3 -.-> A6["parent: Started + runId"]
    A2 -.->|refused| A7["stderr, exit 1, no child"]
  end
```

### Changed seams

|---|---|---|
| `run --detach → caller` | `--json` ack gains `runId`; human output prints the id and `workflow get <id>` | `packages/cli/src/commands/workflow.ts`, `workflow.test.ts` (`--detach --json emits a structured ack`) |
| `parent CLI → detached child` | New internal `--internal-detached-run-id`; the child executes that row rather than creating one | `packages/cli/src/args.ts`, `packages/cli/src/cli.ts`, `workflow.test.ts` (`detached child adopts the pre-created run`) |
| `IWorkflowStore.updateWorkflowRun` | Accepts `working_path`, write-once via `COALESCE` | `packages/workflows/src/store.ts`, `packages/core/src/db/workflows.ts`, `workflows.test.ts` |
| `executeWorkflow(preCreatedRun)` | A fresh pre-created row now also gets `working_path` and container isolation stamps; the container + durable-wait guard keys on `!isContinuation` rather than `preCreatedRun === undefined` | `packages/workflows/src/executor.ts`, `executor.test.ts` |
| `prepareWorkflowSource` | Optional caller-supplied `runId`, so a capture taken after the row exists is still filed under it | `packages/workflows/src/executor.ts` |

## Validation

- `bun run validate` — passed (exit 0) — types, lint, format, full suite.
- `bun --cwd packages/cli test src/commands/workflow.test.ts` — 305 pass / 0 fail — proves the synchronous refusal, the `runId` ack, the launch-failure row write, the child's adoption of the row, and the `--resume --detach` ack.
- **Regression proof:** `refuses an unresolvable --adopt synchronously instead of acking Started (#2872)` fails before the fix — it took 502 ms because the parent spawned and waited out the startup window, and the `resolveWorkflowAdoption` mock went unconsumed, proving the parent never called it. It passes after.
- `bun --cwd packages/cli run test:integration` — passed — `workflow-terminal-event.integration.spec.ts` drives the real CLI end to end and now also asserts the ack's `runId` is readable in SQLite *at ack time*, that the child adopts that same row (no second row), and that `working_path` is filled in afterward.
- `bun --cwd packages/workflows test src/executor.test.ts` — 129 pass / 0 fail. `bun --cwd packages/core test src/db/workflows.test.ts` — 123 pass / 0 fail.
- **Live acceptance** against the real CLI with a scratch `ARCHON_HOME`: `workflow run probe-success --folder --detach --adopt does-not-exist` → exit 1, stderr `Cannot adopt: no workflow run 'does-not-exist' exists.`, no `Started` line, run-row count unchanged. A valid launch printed `Run id: 304169f5…` and `archon workflow get 304169f5…` returned the run immediately.
- **Review corrections (second commit), each guard observed failing before it was added:** the plain no-flag `--detach` launch is refused when no project resolves (live: `Cannot create worktree: not in a git repository`, no fork); a handed-over row that never starts is failed with the reason (live: a detached repo run whose worktree creation failed left `status: failed` + a `workflow_failed` event carrying `Detached run failed to start: Cannot determine git remote…`, where it previously left `pending`); `--internal-detached-run-id` is refused outside a detached child (live: exit 1 against a completed run's id), for a non-`pending` row, and for a row belonging to another workflow.
- **Not verified:** `--container --detach` was not exercised on real Docker. The container isolation stamps a resume depends on ride the same executor update as `working_path` and are pinned by a unit test (`stamps container isolation onto a fresh pre-created row`), but no live container run was performed.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No schema change. `working_path` was already nullable (`workflow-run.ts:228`) and the new update is additive and write-once. | `packages/core/src/db/workflows.ts` |
| Observability | A detached launch that fails now fails loudly on the terminal instead of in a child log; a child that dies during startup leaves a `failed` row naming the reason. | `workflow.test.ts` (`fails the run row it created when the detached child dies during startup`) |

## Links

- Closes #2872
- Related #2745
- Related #2871

## Summary by CodeRabbit

* **New Features**
  * Detached workflow runs are created and identified before execution begins.
  * Detached run acknowledgements include the run ID and status.
  * Run checkout paths are recorded reliably and preserved once set.
  * Detached runs retain configuration, metadata, and source capture across process handoffs.

* **Bug Fixes**
  * Startup failures now update pre-created runs with failure details.
  * Improved validation for adoption, resumption, and missing run records.
  * Prevented duplicate run records during detached continuations.
  * Improved cleanup and signal handling for detached runs.


Wirasm:
## Ready to merge

Correction round verified. The three blocking findings from the first pass (R1, R2, R3) are genuinely closed — not just addressed, but closed in the way the finding actually asked for, confirmed by reverting the correction commit's implementation and rerunning its own new tests against the pre-correction code (6 of them fail as expected; restored afterward, tree clean). R8 (the PR description's wrong safety-mechanism citation) is corrected precisely — the new "Known risk" text drops the false 5-minute stale-pending claim and states the actual reason a leftover row is inert, which is exactly what R8 asked for. Nothing previously cleared (the `working_path` write-once COALESCE, the single-resolver `resolveRunCodebase`/`resolveWorkflowAdoption` calls, the `!isContinuation` guard fix) was touched or undone — the correction commit (`13d3e126`, diffed against its own parent `b0283a4a`) only changes `packages/cli/src/commands/workflow.ts` and its test file; `executor.ts`, `store.ts`, and `workflow-adoption.ts` have zero diff since the last review. #2871 remains untouched. R4–R7 (Suggestions) were left as-is — reasonable, since they're non-blocking polish and the correction correctly stayed focused on the three blocking findings plus the description fix.

One new Suggestion (R9): the new `--internal-detached-run-id` ownership check has a doc comment that slightly overclaims what its process-group half proves — empirically verified that a plain interactively-run command already satisfies it, so it meaningfully blocks scripted/automated misuse but not a deliberate human at a terminal. This doesn't reopen R3: the two guards that actually closed R3's damage (`status === 'pending'`, `workflow_name` match) hold regardless, so the worst outcomes (terminal-run corruption, cross-workflow hijack) are genuinely prevented either way.

**Validation:** All local test suites pass at the new head (326 tests across the touched packages, including 313/313 in `workflow.test.ts`), lint and type-check are clean, and all 6 GitHub CI legs (including Windows) are green.

### Findings

|---|---|---|---|
| `R1` | Important | Default (no-flag) `--detach` launch had no pre-flight gate for a codebase lookup/registration failure | FIXED |
| `R2` | Important | Child's live re-validation had no `failWorkflowRun` backstop before the row's signal-handler setup | FIXED |
| `R3` | Important | `--internal-detached-run-id` accepted any run id with no `--detach` gate and no `workflow_name`/status validation | FIXED |
| `R4` | Suggestion | `resolveCliUserRecordId()` runs twice per fresh detach launch (parent + child) — redundant DB round trip | OPEN |
| `R5` | Suggestion | `preparedSource?.runId` fallback in the `ownedRunId` chain is dead code | OPEN |
| `R6` | Suggestion | Two blocks left duplicated inline despite the PR extracting four siblings of the same shape | OPEN |
| `R7` | Suggestion | `launchedRunId` is redundant with `!isContinuation`; could collapse to one variable | OPEN |
| `R8` | Suggestion | PR description cited the wrong safety mechanism for why a leftover `pending` row is harmless | FIXED |
| `R9` | Suggestion | New — the detached-child ownership check's doc comment overstates what its process-group half proves against a deliberate interactive human (scripted/automated misuse is genuinely blocked) | OPEN |

<details>
<summary><code>R1</code> — Default `--detach` launch skips the codebase-resolution fatal gate the main path already has</summary>

**Original impact:** `archon workflow run <name> --detach` (no `--folder`/`--adopt`/`--supersedes`/`--resume`) reproduced the bug #2872 was filed to fix, for a codebase-lookup DB error or auto-registration failure: `Started`, exit 0, silent child death.

**Fix:** A shared `assertCodebaseResolvedForIsolation()` (`workflow.ts:684-720`) replaces the main path's three inline throws (lookup error / registration error / not-a-git-repo) verbatim, and the pre-flight now calls it unconditionally for an isolating launch: `if (!options.noWorktree && wantsIsolation) { assertCodebaseResolvedForIsolation(detachResolved); }` (`workflow.ts:1960-1962`), positioned right after the existing `--folder` check and before `--adopt`/`--supersedes` — so it covers the plain no-flag case this finding was about.

**Verification:**
- Traced the exact reachability condition of the main path's original `else if (wantsIsolation)` branch (behind an `if (isFolderCodebase && codebase) … else if (wantsIsolation && codebase) … else if (options.noWorktree) …` chain) and confirmed it reduces to `!options.noWorktree && wantsIsolation && !codebase` — identical to the pre-flight's new guard, using the helper's own internal `if (resolved.codebase) return;` no-op to supply the missing `&& !codebase` term. No widening or narrowing.
- The three error strings are byte-identical to what was inline before (verified by diff — the extraction is a pure move).
- Reverted only the implementation half of the correction commit (`git apply -R` on `workflow.ts`, keeping the new tests) and reran `workflow.test.ts`: the two new R1 tests (`refuses an isolating launch whose project lookup fails, before forking`, `refuses an isolating launch outside a git repository, before forking`) failed against the pre-correction code as expected, confirming they discriminate. Restored and reran clean (313/313 pass).
- The main path's non-detached behavior for this branch is unchanged (same function, same call site, same condition, error strings byte-identical).

**Disposition:** FIXED.

</details>

<details>
<summary><code>R2</code> — No failure-recording net between row creation and the child's own supervision setup</summary>

**Original impact:** A throw between the handed-over row and the executor claiming it (e.g., the child's live re-check of `--adopt`/`--supersedes` disagreeing with the parent's discarded pre-flight check, or R1's own gap) left the row `pending` forever with nothing recording why.

**Fix:** `recordDetachedChildStartupFailure()` (`workflow.ts` new function) wraps the entire `workflowRunCommand` body in a try/catch; on any throw it reads the row's current status via `getWorkflowRunStatus` and calls `failWorkflowRun` only if it is still `pending`.

**Verification (the load-bearing claim — ordering):**
- Confirmed in `packages/workflows/src/executor.ts:2925-2929` that a fresh dispatch (`!dagPriorCompletedNodes`) writes `status: 'running'` via an **awaited** `updateWorkflowRun` call immediately after emitting the `workflow_started`-class events and strictly before any node dispatch begins ("Set status to running now that execution has started"). Since Node.js is single-threaded and this write is on the same awaited call chain the CLI's try/catch wraps, there is no concurrent-access window — the flip either fully completes before any node can fail, or it itself throws (in which case the row genuinely never left `pending`, and `failWorkflowRun` firing is correct).
- Traced that a `{success:false}` result from `executeWorkflow` — the shape a node failure produces — is converted to a throw at the very end of `runWorkflowWithOwnedSource` (`else { throw new Error('Workflow failed: ${result.error}'); }`), so it also reaches the new catch; by then status is already `running` or later, so the CAS guard correctly no-ops.
- Reverted the implementation and reran: the new positive test (`fails the handed-over row when it throws before the executor claims it`) failed against pre-correction code as expected; the paired negative test (`leaves a row the executor already claimed to its own lifecycle`, mocking status `'running'` and asserting `failWorkflowRun` is not called) is a genuine CAS check, verified by direct code reading of the guard (`if (status !== 'pending') return;`).
- Live acceptance evidence in the updated PR body: "a detached repo run whose worktree creation failed left `status: failed` + a `workflow_failed` event carrying `Detached run failed to start: Cannot determine git remote…`, where it previously left `pending`."

**Disposition:** FIXED. The lifecycle-ownership concern raised in the original review guidance — a fix here must not guess at ownership of work it can't prove is orphaned — is satisfied: the guard only ever touches a row still `pending`, which by the verified ordering above is exactly the set of rows no execution has claimed.

</details>

<details>
<summary><code>R3</code> — <code>--internal-detached-run-id</code> accepted any run id with no ownership check</summary>

**Original impact:** The flag wasn't gated on `--detach`, and the row lookup had no `workflow_name`/status filter — a mistyped or reused id could hijack another run's row, including re-executing a terminal run's DAG from scratch.

**Fix:** Three conditions are now all required before the row is honoured (`workflow.ts:1409-1447`): (1) `detachedProcessOwner` — this process owns the `ARCHON_DETACHED_RUN_OWNER` env marker AND `assertDetachedRunProcessOwner()` proves it leads its own POSIX process group; (2) the loaded row's `status === 'pending'`; (3) the loaded row's `workflow_name` matches the workflow actually being invoked (checked once discovery resolves it, `workflow.ts:1587-1595`).

**Verification:**
- Confirmed `getWorkflowRun` is still an unfiltered `SELECT ... WHERE id = $1` — the new checks are entirely at the call site, not the DB layer, which is fine since it's the single ingestion point.
- Confirmed the `status`/`workflow_name` checks close the two catastrophic outcomes from the original finding: a non-`pending` (including terminal) row is refused outright, and a mismatched `workflow_name` is refused outright — so the "silently re-execute a completed run's DAG" and "hijack a different workflow's row" cases are both genuinely closed.
- Reverted and reran: three new tests (refuses outside a detached child, refuses a non-`pending` row, refuses a mismatched `workflow_name`) all failed against pre-correction code as expected.
- **Independently investigated the process-group check's actual strength**, since the doc comment claims it "has already proved this process leads its own process group": empirically confirmed via a non-interactive shell that a plain child process is *not* its own process-group leader (`process.kill(-pid, 0)` → `ESRCH`), but under a real pseudo-tty (`script -q /dev/null bash -c '...'`, approximating an interactive terminal with job control) a plain foreground command *is* its own process-group leader by default. So `assertDetachedRunProcessOwner()` reliably blocks non-interactive/scripted/automated misuse (cron, CI, a wrapper script, a backgrounded reuse of a stale command) but does **not** reliably block a human who manually sets `ARCHON_DETACHED_RUN_OWNER=1` and types the flag at their own interactive terminal — that combination can satisfy all three conditions if they also know a `pending` row's id and matching `workflow_name`. This does not reopen R3: the two checks that actually eliminate the *catastrophic* consequences (terminal-run corruption, cross-workflow hijack) hold regardless of who set the env var. The residual is narrower and race-shaped — targeting another live `pending` launch of the identical workflow — architecturally the same class of risk the original review already accepted as bounded for R2. Recorded as new finding R9 below.

**Disposition:** FIXED for the concrete consequences the original finding named.

</details>

<details>
<summary><code>R4</code> — Redundant CLI-user lookup on every fresh detach launch</summary>

**Disposition:** OPEN — untouched by this correction, as expected (non-blocking Suggestion, correctly out of scope for a fix round targeting the three blocking findings).

</details>

<details>
<summary><code>R5</code> — Dead fallback branch in the <code>ownedRunId</code> chain</summary>

**Disposition:** OPEN — untouched; `preparedSource?.runId` is still present at the same line. Non-blocking.

</details>

<details>
<summary><code>R6</code> — Two duplicated blocks weren't extracted like their siblings</summary>

**Disposition:** OPEN — untouched; both duplicated strings still present verbatim in two places. Non-blocking.

</details>

<details>
<summary><code>R7</code> — Redundant <code>launchedRunId</code> variable</summary>

**Disposition:** OPEN — untouched; `launchedRunId` still present at the same shape. Non-blocking.

</details>

<details>
<summary><code>R8</code> — PR description cited the wrong safety mechanism for the residual "child dies during startup" case</summary>

**Fix:** The "Known risk" section of the PR description now reads: "A child killed outright (SIGKILL, host crash) before the executor claims its row leaves that row `pending`. Any JS-level failure is now recorded ... so this is narrowed to abrupt termination. Such a row is inert: with `working_path` NULL it can never hold a path lock (the lock query matches an exact path), can never be resumed (the resume CAS needs a resumable status), and can never be adopted (adoption needs a terminal status)."

**Verification:** This drops the false "5-minute stale pending" citation entirely and states the actual reason (NULL `working_path` can never match a path-lock query; the resume CAS and adoption both require a status this row will never reach) — exactly the correction R8 asked for, and it correctly narrows the residual to abrupt termination now that R2 covers every JS-level throw.

**Disposition:** FIXED.

</details>

<details>
<summary><code>R9</code> — New: the ownership check's doc comment overstates what its process-group half proves</summary>

**Impact:** None on the substantive fix — see R3's verification above. This is a documentation-precision note: `workflow.ts`'s comment above the `detachedProcessOwner`/`assertDetachedRunProcessOwner()` checks implies process-group leadership is a strong proof of "this process is a detached child `spawnDetachedWorkflowRun` created," but it is also satisfied by any plain command run interactively (empirically confirmed). It correctly blocks scripted/non-interactive misuse, which is the more realistic accidental-reuse threat.

**Evidence:** `packages/cli/src/utils/detached-run-control.ts:75-79` (`assertDetachedRunProcessOwner`), `:463-470` (`processGroupExists`, `process.kill(-pid, 0)`). Reproduced directly: a `bun -e` process invoked non-interactively is not its own process-group leader (`ESRCH`); the same code under a pseudo-tty (`script -q /dev/null bash -c '...'`) is.

**Required outcome:** None required. Optionally soften the doc comment to describe what this check actually rules out (non-interactive/scripted reuse) rather than implying it alone proves legitimate parent-spawned provenance; the `status`/`workflow_name` checks are what carry the real weight.

**Found by:** coordinator (independent empirical verification during correction review)

**Disposition:** OPEN — informational; no blocking consequence given the other two guards.

</details>

<details>

#### Correction-verification method

Bounded to the pure correction diff: `git diff b0283a4a96477cc1d0f801d187de55a46db5e377 13d3e126ca77ea814572cca6f773873d46c82297` (the correction commit against its own parent, which is the post-dev-merge state) — 2 files, `packages/cli/src/commands/workflow.ts` (+168/−34) and its test file (+204). Confirmed `packages/workflows/src/executor.ts`, `packages/workflows/src/store.ts`, and `packages/core/src/operations/workflow-adoption.ts` have zero diff since the prior reviewed head (`7b6424bf`) — nothing previously cleared was touched. The PR's own `b0283a4a` merge brought in unrelated upstream `dev` commits (visible in a wider `7b6424bf..13d3e126` diff — a rejection-cancellation event fix, an `sdlc` lens-attribution feature, an orphaned-runs sweep removal, etc.); none of it touches the files or invariants this review covers, confirmed by direct diff inspection.

Spot-checked test discrimination by reverting only the implementation half of the correction (`git apply -R` on `workflow.ts`, keeping the new tests) and rerunning `workflow.test.ts`: 6 of the correction's new tests failed against the pre-correction code as expected (2 for R1, 1 for R2, 3 for R3), then restored cleanly (working tree matched the head exactly; reapplying the same diff produced zero residual diff).

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1/R2/R3 fixes verified sound; R4 still open |
| `seams` | R2 fix verified against its original concern |
| `simplify` | R5, R6, R7 confirmed untouched (expected) |
| `types` | R3 fix verified; R9 new observation on the process-group guard |
| `errors` | R1/R2 fixes verified against their original failure paths |

#### Validation

|---|---|---|
| `bun test src/commands/workflow.test.ts` (in `packages/cli`) | PASS | 313 pass / 0 fail |
| `bun run test:integration` (in `packages/cli`) | PASS | `detached-run-control.integration.spec.ts` 5/5, `workflow-terminal-event.integration.spec.ts` 1/1 |
| `bun test src/executor.test.ts` (in `packages/workflows`) | PASS | 129 pass / 0 fail |
| `bun test src/db/workflows.test.ts` (in `packages/core`) | PASS | 120 pass / 0 fail |
| `bun run lint` (repo root) | PASS | clean across all linted packages |
| `bun x tsc --noEmit` (`packages/cli`) | PASS | clean |
| GitHub CI — `test (ubuntu-latest)` | PASS | — |
| GitHub CI — `test (windows-latest)` | PASS | Green on this head — neither known flake (#2851, and the separately-known `detached workflow terminal database events` flake from PR #2826 history) was hit, so no inherited-failure judgment call was needed |
| GitHub CI — `docker-build`, `postgres-parity`, `schema-upgrade`, CodeRabbit | PASS | — |

</details>

PR #2915, opened 2026-08-28
test(workflows): exec-code fixture tests stop building a git repo per test

### LINKED ISSUE #2882: test(workflows): the @archon/workflows suite runs at the Windows budget edge — a different test dies at ~5000ms each run

Wirasm:
## Problem and outcome

`@archon/workflows` runs at the Windows 5000ms per-test budget, and a different test dies each run (#2882). On dev merge `fb33fe32` it was `runFixtures exec-code isolation (#2851) > gives clean and pre-dirtied callers the same verdict`, at 5015ms. That test spent most of its budget building git repositories, not exercising the isolation it is named after: each of its two halves ran `git init`, `git add`, and `git commit` before the behavior under test started. Sixteen spawned processes in one test body, six of them pure setup.

- **Outcome:** the parity test spawns 10 processes instead of 16 and runs in ~103ms instead of ~187ms locally; the whole exec-code block spawns 42 instead of 54 and runs in ~500ms instead of ~770ms.
- **Invariant:** a test spawns a subprocess only when the subprocess is the behavior under test, and at most the minimum number that proves it (#2882's stated invariant). Every remaining spawn in the parity test is the isolation mechanism itself — `git worktree add`, the executed node, its observation of the workspace, `git worktree remove`.
- **Scope boundary:** no budget was raised, no test was skipped on Windows, no coverage was deleted, and `fixture-runner.ts` is unchanged. The C5 `migrate-state-dir` lane on #2882 is untouched.
- **Root cause:** repository construction, not the isolation mechanism, was the cost. Measured locally, `git init` + `add` + `commit` is ~50ms of a ~90ms half; copying a prepared repository instead is ~0.8ms. Process creation is what multiplies on a contended Windows runner.

## Review guidance

- **Feedback requested:** whether the coverage is genuinely unchanged — specifically whether a caller copied from a prepared repository still represents the scenarios these tests claim.
- **Start here:** `packages/workflows/src/fixture-runner.test.ts:641` (the block comment) and `:660` — the two prepared caller repositories and the `callerFrom` copy that replaces three git processes per test.
- **Review order:** the `beforeAll`/`callerFrom` block, then the parity test just below it, then the four sibling tests that switch to `callerFrom`, then the `GUARD_YAML` change, then the cleanup migration.
- **Lower-attention areas:** the `cleanups.push` removals throughout the file are mechanical — `makeTempProject` now registers every root with `trackTempRoots` at creation.
- **Known risk or uncertainty:** the fix cannot be proven on Windows from a macOS worktree. It is argued by cost class: the removed work is process creation and the remaining work is not, and the numbers below are before/after on the same machine, interleaved to cancel drift.

## Solution

Two caller repositories are prepared once per block and copied per test. `git worktree add` has nothing to check out without a HEAD commit, so every exec-code test needs its caller to be a real repository — but a plain repository's metadata holds no absolute paths, so a directory copy of one is an independent repository. `trackedOnlyRepo` commits just `tracked.txt`, for tests whose own project files may sit untracked beside it; `guardRepo` commits the guard workflow and its fixture too, so a caller copied from it is genuinely clean, which is the state the parity test contrasts dirt against.

The parity test then invokes one caller twice — clean, then dirtied — rather than building two repositories. That is both cheaper and a tighter claim: the dirt becomes the only difference between the two verdicts, where two separately built repositories also differed in identity and history.

One test keeps building its repository in place. `captures the working tree, so an uncommitted script edit is what runs` needs the superseded script version committed, because what HEAD holds is its premise.

Separately, the guard workflow now captures `git status --porcelain` output and tests it, instead of testing `$(git status --porcelain)` inline. A git that *fails* also prints nothing, so the inline form read that silence as a clean tree. Verified: with the workspace replaced by an empty non-repository directory, the old guard passed both halves; the new one fails.

Temp cleanup moves to `trackTempRoots`/`removeTempTree`, which retry while Windows still holds a handle inside the tree. The file leaves `LEGACY_RECURSIVE_CLEANUP`.

## Validation

- `bun test src/fixture-runner.test.ts` — 32 pass, 0 fail — the restructured file is green.
- `bun run test` in `packages/workflows` — 2526 pass, 0 fail.
- `bun run lint` — green, including the test-cleanup drift check with the ledger entry removed.
- **Mutation, isolation removed** (`withExecWorkspace` returns `fn(cwd)`, executing in the caller's own checkout): the parity test fails, as does `keeps executed writes out of the caller checkout`. The parity coverage still bites.
- **Mutation, capture removed** (fixtures resolve `liveSourceRoots(cwd)` instead of the frozen capture): `executes the captured copy of a named script`, `shares one capture across every fixture`, and `captures the command folder` all fail. The four tests that moved to `trackedOnlyRepo` did not lose their teeth by having their project files untracked.
- **Mutation, scripts resolved from the HEAD worktree**: `captures the working tree, so an uncommitted script edit is what runs` fails — the reason that one test keeps a real commit.
- **Mutation, workspace is an empty non-repository**: the new guard fails; the previous inline guard passed. This is why the guard changed.
- **Measurements**, interleaved AFTER/BEFORE blocks of three, nine runs each, same machine:

  | | Before | After |
  |---|---|---|
  | parity test, spawned processes | 16 | 10 (+6 in `beforeAll`, amortized across the block and outside every test's budget) |
  | parity test, wall time | ~187ms | ~103ms |
  | exec-code block, spawned processes | 54 | 42 |
  | exec-code block, wall time | ~770ms | ~500ms |

  Process counts come from `git`/`bash` shims on `PATH` and `ARCHON_BASH_PATH` that log every invocation. Bun charges `beforeAll` to no test: a 400ms `beforeAll` leaves the first test at 0.2ms in the junit reporter.
- **Not verified:** that Windows CI is now consistently green. One run cannot establish that, which is why this does not close #2882.

## Links

- Related #2882
- Related #2851

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation and temporary-file cleanup for more reliable test execution.
  * Streamlined repository-based test setup to reduce unnecessary setup work.
  * Added clearer handling for Git command failures during test checks.
* **Chores**
  * Updated cleanup validation to reflect the current test setup.

### DIFF
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -38,7 +38,7 @@ mock.module('@archon/paths', () => ({
   getDefaultCommandsPath: () => join(bundledDefaultsRoot, 'defaults'),
 }));
 
-import { execFileAsync } from '@archon/git';
+import { execFileAsync, resolveBashPath } from '@archon/git';
 import { parseWorkflow } from './loader';
 import { expandWorkflowIncludes } from './include-expander';
 import type { WorkflowWithSource } from './schemas/workflow';
@@ -595,6 +595,18 @@ describe('runFixtures', () => {
 describe('runFixtures exec-code isolation (#2851)', () => {
   const COMMITTED_YAML = 'committed-file\n';
 
+  /**
+   * The probe body {@link GUARD_YAML} runs.
+   *
+   * Captured and tested separately, not as `[ -z "$(git status --porcelain)" ]`: a git
+   * that FAILS also prints nothing, so the inline form reads its silence as a clean tree
+   * and the guard passes without having observed anything.
+   *
+   * Shared with the test that proves that, rather than restated there — a second copy
+   * would keep passing after this one was reverted, which is the whole failure mode.
+   */
+  const GUARD_PROBE = ['status="$(git status --porcelain)" || exit 1', '[ -z "$status" ]'];
+
   /** A bash node whose success depends ONLY on the cleanliness of the checkout it executes in. */
   const GUARD_YAML = [
     'name: test-wf',
@@ -602,11 +614,7 @@ describe('runFixtures exec-code isolation (#2851)', () => {
     'nodes:',
     '  - id: probe',
     '    bash: |',
-    // Captured and tested separately, not as `[ -z "$(git status --porcelain)" ]`: a git
-    // that FAILS also prints nothing, so the inline form reads its silence as a clean
-    // tree and the guard passes without having observed anything.
-    '      status="$(git status --porcelain)" || exit 1',
-    '      [ -z "$status" ]',
+    ...GUARD_PROBE.map(line => `      ${line}`),
   ].join('\n');
 
   const WRITER_YAML = [
@@ -691,6 +699,29 @@ describe('runFixtures exec-code isolation (#2851)', () => {
     writeFileSync(join(cwd, 'scratch.txt'), 'untracked stray\n');
     const dirtyReport = await runFixtures({ workflows, cwd });
     expect(dirtyReport.passed).toBe(1);
+  });
+
+  it('fails the guard when git cannot report status, rather than passing on its silence', async () => {
+    // What the test above rests on: the guard must distinguish "the tree is clean" from
+    // "I never got to look". A failing `git status` prints nothing, so the shorter
+    // `[ -z "$(git status --porcelain)" ]` reads that silence as cleanliness — measured,
+    // not supposed: with the scratch worktree replaced by a plain empty directory, the
+    // inline form passed BOTH halves above, certifying a checkout that never happened.
+    //
+    // `GIT_DIR` pointing nowhere reproduces exactly that condition — git exits non-zero
+    // having written nothing to stdout — and needs no repository, no scratch worktree and
+    // no process beyond the one the guard itself runs in.
+    const cwd = makeTempProject();
+    const exit = await execFileAsync(resolveBashPath(), ['-c', GUARD_PROBE.join('\n')], {
+      cwd,
+      env: { ...process.env, GIT_DIR: join(cwd, 'no-such-git-dir') },
+    }).then(
+      () => 'the guard passed',
+      (error: { code?: unknown }) => error.code
+    );
+    // Bash's exit status, so specifically the guard's own `|| exit 1`. A string here would
+    // be a spawn failure — bash never ran — which must not read as the guard working.
+    expect(exit).toBe(1);
   });
 
   it('keeps executed writes out of the caller checkout (#2851)', async () => {
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -1,18 +1,19 @@
 /** Tests for the declared-data dry-run fixture runner (#2772). */
-import { describe, it, expect, afterEach, afterAll, mock } from 'bun:test';
+import { describe, it, expect, beforeAll, afterAll, mock } from 'bun:test';
 import {
+  cpSync,
   existsSync,
   mkdirSync,
   mkdtempSync,
   readdirSync,
   readFileSync,
-  rmSync,
   symlinkSync,
   writeFileSync,
 } from 'node:fs';
-import { mkdir, rm } from 'node:fs/promises';
+import { mkdir } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { removeTempTree, trackTempRoots } from '@archon/paths/test-utils';
 
 // Capture-cost control, same lever and same reason as `subrun.test.ts` (#2882): every
 // `runFixtures` call takes one source capture, and a capture copies and digests the
@@ -29,9 +30,7 @@ import { join } from 'node:path';
 // dirname(getDefault*Path()), so the getter's PARENT must be the owned empty tree.
 const bundledDefaultsRoot = join(tmpdir(), `fixture-runner-test-empty-bundled-${process.pid}`);
 await mkdir(join(bundledDefaultsRoot, 'defaults'), { recursive: true });
-afterAll(async () => {
-  await rm(bundledDefaultsRoot, { recursive: true, force: true }).catch(() => {});
-});
+afterAll(() => removeTempTree(bundledDefaultsRoot));
 const realArchonPaths = await import('@archon/paths');
 mock.module('@archon/paths', () => ({
   ...realArchonPaths,
@@ -72,8 +71,10 @@ import {
   type RunFixturesOptions,
 } from './fixture-runner';
 
-function makeTempProject(): string {
-  return mkdtempSync(join(tmpdir(), 'fixture-runner-'));
+/** The file's one temp-root creator, so every fixture tree is tracked for teardown. */
+const trackTempRoot = trackTempRoots();
+function makeTempProject(prefix = 'fixture-runner-'): string {
+  return trackTempRoot(mkdtempSync(join(tmpdir(), prefix)));
 }
 
 function isolatedSourceRoots(cwd: string): WorkflowSourceRoots {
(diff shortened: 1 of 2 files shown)


Wirasm:
## Ready to merge

This PR removes per-test git-repo construction from the exec-code fixture tests (the fix for the Windows CI flake at #2882) without cutting coverage, skipping Windows, or raising the timeout budget — the diff is scoped to exactly two files, and every claim in the implementor's report checks out against direct evidence: full local test and lint runs, a hand-run reproduction of the silent-failure guard bug, and two independent mutations against the exec-code isolation machinery (`withExecWorkspace`, `withCapturedFixtureSource`) that reproduce the exact test failures the report claims. Five specialist reviewers (code, seams, simplification, test-coverage, silent-failure) each traced the shared-template-plus-`cpSync` pattern and the untracked-project-files question independently and found no drift, no cross-test state bleed, and no weakened assertion. The only findings are two small, non-blocking Suggestions plus one already-declined observation.

**Validation:** all green — `bun test src/fixture-runner.test.ts` (32/32 pass), full `packages/workflows` `bun run test` (2526/2526 pass), `bun run lint` (drift check + eslint, exit 0), `bun x tsc --noEmit` clean.

### Findings

|---|---|---|---|
| `R1` | Suggestion | New git-failure branch of the guard probe (`\|\| exit 1`) has no automated regression test | OPEN |
| `R2` | Suggestion | The silent-failure guard fix rides in the same commit as the pure perf refactor | DECLINED |
| `R3` | Suggestion | `callerFrom(template, cwd?)`'s two-arg form could be unified to a single calling convention | OPEN |

<details>
<summary><code>R1</code> — guard's git-failure path is unprotected by the test suite</summary>

**Impact:** The PR's headline correctness fix — `packages/workflows/src/fixture-runner.test.ts:605-609`, changing `[ -z "$(git status --porcelain)" ]` (which silently reads a *failed* `git status` as "clean") to `status="$(git status --porcelain)" || exit 1; [ -z "$status" ]` — is proven correct by direct reproduction, but nothing in the automated suite would catch a future accidental revert of those two lines back to the old form.

**Evidence:** `packages/workflows/src/fixture-runner.test.ts:605-609`. The `pr-test-analyzer` agent reverted those two lines to the pre-PR one-liner and reran `bun test src/fixture-runner.test.ts`: still 32/32 pass — no existing test (including the parity test at `:679-694`) exercises a state where `git status` inside the exec workspace actually fails, so nothing distinguishes old-guard from new-guard behavior. I independently reproduced the underlying shell claim directly (outside the test suite): in an empty non-git directory, `bash -c '[ -z "$(git status --porcelain)" ]'` exits 0 (old form falsely "clean"), while the new two-line form exits 1. The PR description reports the same manual check but does not encode it as a test.

**Required outcome:** Not required for merge — the fix itself is correct and already covered by the parity test's happy path. Optional follow-up: a small direct test of the guard snippet (or an equivalent probe) under a forced `git status` failure, so a later "simplification" that reverts the two-line fix fails CI instead of shipping silently.

**Disposition:** OPEN (Suggestion). Non-blocking — the delivered fix is verified correct by direct reproduction; the gap is in future regression protection, not current behavior.

</details>

<details>
<summary><code>R2</code> — guard fix and perf refactor share one commit</summary>

**Impact:** None on correctness. A reviewer scanning only the PR title ("stop building a git repo per test") could miss that a real bug fix (the silent-failure guard) rides along in the same diff.

**Evidence:** `packages/workflows/src/fixture-runner.test.ts:605-609` vs. the rest of the diff.

**Required outcome:** No correction required — the PR description already calls out the guard change explicitly under "Solution," so the risk the finding names is already mitigated.

**Disposition:** DECLINED. Already surfaced in the PR description; splitting a 2-line, already-documented fix out of a 125-line test-restructuring PR is not worth a second PR.

</details>

<details>
<summary><code>R3</code> — <code>callerFrom</code> has two calling conventions</summary>

**Impact:** None on behavior. `callerFrom(template, cwd?)` builds fresh when called with one argument (used twice for `guardRepo`) and grafts onto an already-populated directory when called with two (used three times with `trackedOnlyRepo`). The `code-simplifier` agent verified that reordering the three two-arg call sites to the one-arg form (`const cwd = callerFrom(trackedOnlyRepo); writeTempProject({ cwd, ... })`) produces an identical final directory and git-status state in every case, since none of those three tests depend on write order or on git cleanliness.

**Evidence:** `packages/workflows/src/fixture-runner.test.ts:674-677` (definition), `:702`, `:741`, `:775` (two-arg call sites).

**Required outcome:** Optional: unify to a single-argument `callerFrom`, removing one branch from its contract at no line-count or behavior cost.

**Disposition:** OPEN (Suggestion). Take-it-or-leave-it polish; does not affect correctness or the delivered outcome.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R2 |
| `seams` | No additional findings — parity-test-reuse and untracked-project-files hypotheses both traced and disproved (see below) |
| `simplify` | R3 |
| `tests` | R1 |
| `errors` | No additional findings — `beforeAll`/`afterAll` partial-throw handling for `trackedOnlyRepo`/`guardRepo` reproduced and confirmed safe (Bun's own hook-failure reporting surfaces the real error independently of `afterAll`'s handling of an unset variable) |

Coordinator's own direct verification, in addition to the five agents:

- Ran the git-status guard probe by hand in an empty non-repo directory: old inline form exits 0 (false "clean"), new two-line form exits 1 — confirms the core silent-failure claim independent of the test suite.
- Mutated `withExecWorkspace` in `fixture-runner.ts` to skip isolation (`return fn(cwd)`): 3 tests fail, including the parity test and "keeps executed writes out of the caller checkout" — matches the report's claim.
- Mutated `withCapturedFixtureSource` to bypass the frozen capture (`return fn(liveSourceRoots(cwd))`): 3 different tests fail — "executes the captured copy of a named script," "shares one capture across every fixture," "captures the command folder" — exact match to the report's claim. Both mutated files were restored to a byte-identical clean state afterward (`diff` confirmed, `git status --porcelain` empty).
- Confirmed the diff is scoped to exactly `packages/workflows/src/fixture-runner.test.ts` and `scripts/check-test-cleanup-drift.ts` against the true merge-base with `origin/dev` (the worktree's local `dev` ref was stale and briefly showed two unrelated files in a naive `dev...HEAD` diff; resolved by fetching `origin/dev`).
- Confirmed `fixture-runner.ts` (production source) is untouched by the PR diff itself — only touched transiently by my own mutation testing, then restored.

#### Validation

|---|---|---|
| `bun test src/fixture-runner.test.ts` (in `packages/workflows`) | PASS | 32 pass, 0 fail, 74 expect() calls |
| `bun run test` (in `packages/workflows`) | PASS | 2526 pass, 0 fail across the package's full chained test script |
| `bun run lint` (repo root) | PASS | drift check ("Test-cleanup drift check passed") + eslint across all packages, exit 0 |
| `bun x tsc --noEmit` (in `packages/workflows`) | PASS | clean, per `code-reviewer` agent |
| `bun scripts/check-test-cleanup-drift.ts` (direct) | PASS | confirms the dropped `LEGACY_RECURSIVE_CLEANUP` ledger entry for `fixture-runner.test.ts` is exact — the file has zero remaining raw `rm`/`rmSync` calls |

</details>

PR #2921, opened 2026-08-28
fix(workflows): fail terminal status writes

### LINKED ISSUE #2910: fix(workflows): a failed terminal DB write is swallowed, leaving a finished run stuck at running
## Problem

Every terminal DB write in the executor is best-effort. When `completeWorkflowRun` or `failWorkflowRun` throws, the executor logs, continues, and finishes normally — emitting the in-process terminal event, writing the log artifact, and exiting. The run row stays `running` for a run that is over.

Completion (`packages/workflows/src/dag-executor.ts:12462-12492`):

```ts
try {
  await deps.store.completeWorkflowRun(workflowRun.id, { duration_ms: duration }, {...});
} catch (dbErr) {
  getLog().error({ err: dbErr as Error, workflowRunId: workflowRun.id }, 'dag_db_complete_failed');
  await safeSendMessage(platform, conversationId,
    'Warning: workflow completed but the run status could not be saved. The workflow result may appear inconsistent.',
    { workflowId: workflowRun.id });
}
// continues: logWorkflowComplete(...), emitter.emit({type:'workflow_completed'}), telemetry, return terminalOutput
```

Failure paths use the same shape via `.catch()` — `dag-executor.ts:12225`, `:12277`, `:12360`, and across `executor.ts` (`:1853`, `:1889`, `:2047`, `:2226`, `:2502`, `:2584`, `:2731`, `:3203`). Example at `dag-executor.ts:12224-12228`:

```ts
await deps.store
  .failWorkflowRun(workflowRun.id, failMsg, scheduledResume)
  .catch((dbErr: Error) => {
    getLog().error({ err: dbErr, workflowRunId: workflowRun.id }, 'dag_db_fail_failed');
  });
```

## Why the warning does not help

The one operator-facing signal is `safeSendMessage` into the run's conversation. For a `--detach` CLI run that conversation is the synthetic CLI conversation, which nobody reads. The log line lands in the detached child's log file. So in the case where this matters most — an unattended run — the failure is invisible.

## Consequences

- **The run row stops being authoritative.** A completed run reads `running` indefinitely. `workflow runs`, `workflow get`, and the console all report a live run that has finished and whose process is gone.
- **It holds the worktree lock.** `getActiveWorkflowRunByPath` treats `running` as holding the lock, so the next invocation at that path is refused until someone runs `workflow abandon`.
- **It is indistinguishable from a crashed owner.** A finished-but-unrecorded run and a genuinely dead run present identically: no process, row says `running`. This is the ambiguity #2908 is meant to resolve, and this defect makes one of its inputs untrustworthy.
- **A waiting host cannot be told the truth.** #2745 derives run attention from the run row precisely because it is the one thing written transactionally on every terminal path. That premise is undermined by a path that finishes without writing it.

## Deliberately not asserting the fix

Swallowing here was plausibly deliberate — losing a run's completed work over a transient DB blip is its own harm, and the executor cannot roll back the work already done. So the question is not "stop catching" but what the engine owes the operator when it cannot record a terminal state it knows it reached. Candidates, none of them settled:

- Retry the terminal write with bounded backoff before giving up; a transient blip is the common case and this is cheap.
- Escalate the surviving evidence somewhere an unattended operator sees, rather than a conversation nobody reads.
- Fail the process with a non-zero exit so a detached child's death is at least visible in its log tail and exit status, instead of exiting 0 having silently diverged from the database.
- Leave a durable marker the row-reading surfaces can render as "finished, unrecorded" rather than "running".

## Regression proof

Force the terminal write to throw (inject a failing store, or lock the database) on a run that otherwise completes. Assert the run does not end with the process exiting successfully while its row reads `running` and no operator-visible signal exists outside a chat message to a conversation with no reader.

## Found by

The #2745 design trace (2026-08-28). Related but distinct from #2830, which fixed the terminal *event* being lost to process exit; this is the terminal *status write* failing and the failure being absorbed. Also distinct from #2906, which is about terminal transitions that never attempt an event at all.

Wirasm:
## Problem and outcome

Terminal status writes were best-effort: a failed `completeWorkflowRun` or `failWorkflowRun` could let an already-finished execution return normally while its row remained `running`. That made the run appear live, held its worktree lock, and obscured the failed write for unattended runs.

- **Outcome:** Every terminal workflow-status write now propagates its failure, so the executor never reports success or an ordinary failure result after it cannot persist the terminal state.
- **Invariant:** A workflow may reach a terminal result only after its terminal database status write succeeds; a failed terminal write rejects the execution.
- **Scope boundary:** This changes error propagation for existing terminal status writes only; it adds no retry, notification channel, or persistence format.
- **Root cause:** The DAG and top-level executors caught and logged rejected terminal status writes, then continued through terminal completion or failure handling.

## Review guidance

- **Feedback requested:** Confirm that propagating all listed terminal status-write failures is the intended lifecycle contract.
- **Start here:** `packages/workflows/src/dag-executor.ts:12481` — successful DAG completion uses a terminal-write wrapper, so a failed write remains distinct and propagates without a second failure write.
- **Review order:** Review `dag-executor.ts` terminal completion and failure paths, then `executor.ts` setup, parent-resume, top-level failure, and finally-backstop paths; finish with the regression tests.
- **Lower-attention areas:** The test additions use injected-store rejection cases; the focused suites and repository validation passed.
- **Known risk or uncertainty:** A terminal persistence outage now rejects the process instead of returning a result while the run remains `running`; this is deliberate so callers and supervisors can observe the failure.

## Solution

Removed the local catch-and-log handlers from every terminal `completeWorkflowRun` and `failWorkflowRun` site identified in #2910. The terminal-write wrapper lets terminal persistence failures bypass ordinary failure handling, preventing an erroneous second terminal write; a failed failure write rejects rather than producing a false terminal result. Non-terminal logging and notification failures retain their existing best-effort handling.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A terminal database write failure could return or emit a terminal result while the run row stayed `running`. | The execution rejects when a terminal status write fails. |
| **Failure behavior** | The error was only logged, and one completion path attempted a chat warning. | The write failure reaches the caller or process supervisor; no false successful or ordinary failed result is returned. |

## Validation

- `bun test src/dag-executor.test.ts` — passed (686 tests) — proves rejected completion and DAG failure writes surface.
- `bun test src/executor.test.ts` — passed (126 tests) — proves a rejected top-level failure write surfaces.
- `bun run type-check` in `packages/workflows` — passed.
- `bun run validate` — passed — includes generated-artifact checks, type checks, lint, formatting, install check, and the test suite.
- **Not verified:** Nothing material; focused tests exercise the new rejection behavior and the repository validation suite passed.

## Links

- Closes #2910

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow completion and failure update errors are now surfaced instead of being silently logged.
  * Prevented workflows from reporting misleading terminal states when status updates fail.
  * Improved error handling for nested, gated, and fan-out workflows, avoiding duplicate failure handling.
  * Added clearer notifications directing you to check workflow status when the final status cannot be saved.

* **Tests**
  * Added coverage for terminal status update failures across standard, nested, gated, and fan-out workflows.

### DIFF
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -431,6 +431,7 @@ import {
   continueResolvedGateRun,
 } from './orchestrator-agent';
 import { buildAiProfile } from '@archon/workflows/model-validation';
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 
 // ─── Helpers ──────────────────────────────────────────────────────────────────
 
@@ -6287,5 +6288,44 @@ describe('continueResolvedGateRun — chat gate continuation source (#2646)', ()
       c => (c as unknown[])[1] as string
     );
     expect(messages.some(m => m.includes('digest mismatch'))).toBe(true);
+  });
+
+  // #2910: both sides of the recovery branch. An ordinary resume failure leaves the
+  // run resumable, so `/workflow resume` is right. A rejected terminal write leaves the
+  // row non-terminal — `/workflow resume` refuses those, so that advice would send the
+  // operator down a dead end for a run that may well have finished its work.
+  async function continueWithRejection(
+    platform: ReturnType<typeof makePlatform>,
+    rejection: unknown
+  ): Promise<string[]> {
+    mockExecuteWorkflow.mockRejectedValueOnce(rejection);
+    await continueResolvedGateRun(
+      platform,
+      'conv-1',
+      makeConversation({ codebase_id: 'codebase-1' }),
+      makeGateCodebase(),
+      [makeTestWorkflowWithSource({ name: 'gated', description: 'live' })],
+      makeGateRun(),
+      'approve'
+    );
+    return (platform.sendMessage as ReturnType<typeof mock>).mock.calls.map(
+      c => (c as unknown[])[1] as string
+    );
+  }
+
+  test('an ordinary resume failure points at /workflow resume', async () => {
+    const messages = await continueWithRejection(makePlatform(), new Error('resume boom'));
+
+    expect(messages.some(m => m.includes('retry with `/workflow resume run-gated`'))).toBe(true);
+  });
+
+  test('a rejected terminal write says the status is unknown, not "retry the resume"', async () => {
+    const messages = await continueWithRejection(
+      makePlatform(),
+      new TerminalStatusWriteError(new Error('db is gone'))
+    );
+
+    expect(messages.some(m => m.includes('final status could not be saved'))).toBe(true);
+    expect(messages.some(m => m.includes('retry with `/workflow resume'))).toBe(false);
   });
 });
--- a/packages/core/src/orchestrator/orchestrator-agent.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.ts
@@ -52,6 +52,7 @@ import {
   recordSelectedWorkflow,
   type PreparedWorkflowSource,
 } from '@archon/workflows/executor';
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 import { liveSourceRoots } from '@archon/workflows/workflow-discovery';
 import {
   assertWorkflowRequirementsMet,
@@ -1693,6 +1694,21 @@ export async function continueResolvedGateRun(
       );
     } catch (error) {
       const err = toError(error);
+      // The run's terminal status could not be written, so its row still says `running`
+      // and its true outcome is unknown. `/workflow resume` is the wrong advice — it
+      // refuses a non-terminal row, and the run may well have finished its work.
+      if (error instanceof TerminalStatusWriteError) {
+        getLog().error(
+          { err, conversationId, workflowRunId: run.id, action },
+          'orchestrator.gate_continuation_terminal_write_failed'
+        );
+        await notify(
+          `${decision}, and **${workflow.name}** ran, but its final status could not be saved ` +
+            `(${err.message}). The decision is recorded — check \`/workflow status ${run.id}\` ` +
+            'before starting another run on this project.'
+        );
+        return;
+      }
       getLog().error(
         { err, errorType: err.constructor.name, conversationId, workflowRunId: run.id, action },
         'orchestrator.gate_continuation_failed'
--- a/packages/core/src/orchestrator/orchestrator-isolation.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-isolation.test.ts
@@ -7,6 +7,7 @@ import type { IsolationEnvironmentRow } from '@archon/isolation';
 // (or the workflow engine) before the mock.module() calls below take effect.
 import type { WorkflowRoutingContext } from './orchestrator';
 import type { PreparedWorkflowSource } from '@archon/workflows/executor';
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 import type { WorkflowDefinition } from '@archon/workflows/schemas/workflow';
 import {
   makeTestComposedWorkflow,
@@ -473,6 +474,39 @@ describe('dispatchBackgroundWorkflow', () => {
     await dispatchBackgroundWorkflow(makeRoutingCtx(), makeTestComposedWorkflow([own], 'own-gate'));
     expect(mockGetOrCreateConversation).toHaveBeenCalled();
     await flushBackgroundExecution();
+  });
+
+  // #2910: a rejected terminal write means the row still says 'running' and the run's
+  // real outcome is unknown. Both sides of the branch are asserted — an ordinary
+  // rejection keeps its compensating write and its "failed" card; a terminal-write
+  // rejection gets neither, because a second write over the channel that just failed
+  // would either fail again or bury the real error.
+  test('marks the run failed and says so when execution rejects with an ordinary error', async () => {
+    const workflow = makeWorkflow({ worktree: { enabled: false } });
+    mockExecuteWorkflow.mockRejectedValueOnce(new Error('exec boom'));
+
+    await dispatchBackgroundWorkflow(makeRoutingCtx(), workflow);
+    await flushBackgroundExecution();
+
+    expect(mockFailWorkflowRun).toHaveBeenCalledTimes(1);
+    expect(mockFailWorkflowRun.mock.calls[0]?.[0]).toBe('run-1');
+    const sent = platform.sendMessage.mock.calls.map(c => c[1]);
+    expect(sent.some(m => m.includes('failed: exec boom'))).toBe(true);
+  });
+
+  test('does not compensate or claim failure when the terminal write rejects', async () => {
+    const workflow = makeWorkflow({ worktree: { enabled: false } });
+    mockExecuteWorkflow.mockRejectedValueOnce(
+      new TerminalStatusWriteError(new Error('db is gone'))
+    );
+
+    await dispatchBackgroundWorkflow(makeRoutingCtx(), workflow);
+    await flushBackgroundExecution();
+
+    expect(mockFailWorkflowRun).not.toHaveBeenCalled();
+    const sent = platform.sendMessage.mock.calls.map(c => c[1]);
+    expect(sent.some(m => m.includes('final status could not be saved'))).toBe(true);
+    expect(sent.some(m => m.includes('failed: Failed to persist'))).toBe(false);
   });
 
   test('worktree.enabled: false skips isolation and runs in the parent cwd', async () => {
--- a/packages/core/src/orchestrator/orchestrator.ts
+++ b/packages/core/src/orchestrator/orchestrator.ts
@@ -50,6 +50,7 @@ import { createIsolationStore } from '../db/isolation-environments';
 import { toError } from '../utils/error';
 import { getCodebase } from '../db/codebases';
 import { executeWorkflow } from '@archon/workflows/executor';
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 import { resolveWorkflowSourceRoot } from '../utils/workflow-source-root';
 import {
   prepareWorkflowSource,
@@ -706,7 +707,11 @@ async function dispatchBackgroundWorkflowOwned(
         }
       } catch (error) {
         const err = toError(error);
-        if (preCreatedRun) {
+        const terminalWriteFailed = error instanceof TerminalStatusWriteError;
+        // A rejected terminal write leaves the row saying `running`. Do not compensate
+        // with a second failWorkflowRun over the write channel that just failed, and do
+        // not tell the user the workflow "failed" — its real outcome is unknown.
+        if (preCreatedRun && !terminalWriteFailed) {
           await workflowDeps.store.failWorkflowRun(preCreatedRun.id, err.message).catch(dbError => {
             getLog().error(
               { err: toError(dbError), workflowRunId: preCreatedRun.id },
@@ -720,12 +725,17 @@ async function dispatchBackgroundWorkflowOwned(
             workflowName: workflow.name,
             workerConversationId: workerPlatformId,
           },
-          'background_workflow_failed'
+          terminalWriteFailed
+            ? 'background_workflow_terminal_write_failed'
+            : 'background_workflow_failed'
         );
         // Surface error to parent conversation — include workflowResult metadata when
         // we have a pre-created run ID so the chat renders a result card with "View full logs"
         const failureRunId = preCreatedRun?.id;
-        const failureMessage = `Workflow **${workflow.name}** failed: ${err.message}`;
+        const failureMessage = terminalWriteFailed
+          ? `⚠️ Workflow **${workflow.name}** finished, but its final status could not be saved. ` +
+            'It may still show as running — check it before starting another.'
+          : `Workflow **${workflow.name}** failed: ${err.message}`;
         await ctx.platform
           .sendMessage(
             ctx.conversationId,
--- a/packages/core/src/utils/error-formatter.test.ts
+++ b/packages/core/src/utils/error-formatter.test.ts
@@ -1,6 +1,7 @@
 import { describe, test, expect } from 'bun:test';
 import { classifyAndFormatError } from './error-formatter';
 import { WorkflowAdoptionError } from '../operations/workflow-adoption';
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 
 describe('classifyAndFormatError', () => {
   describe('rate limit errors', () => {
@@ -713,6 +714,26 @@ describe('classifyAndFormatError', () => {
         "Cannot adopt run 'prior-run': this conversation already continues run 'x' (paused).";
       const result = classifyAndFormatError(new WorkflowAdoptionError(refusal));
       expect(result).toBe(`⚠️ ${refusal}`);
+    });
+  });
+
+  describe('terminal status write failures', () => {
+    // #2910: this is the terminus for the orchestrator's /invoke-workflow and
+    // /workflow run dispatch paths — neither has a closer catch, so handleMessage's
+    // catch formats the error here. A run whose status was never recorded must not
+    // read as a generic error the user is told to /reset away from.
+    test('names the unrecorded status instead of the generic fallback', () => {
+      const result = classifyAndFormatError(
+        new TerminalStatusWriteError(new Error('SQLITE_BUSY: database is locked'))
+      );
+      expect(result).toContain('final status could not be saved');
+      expect(result).toContain('/workflow status');
+      expect(result).not.toContain('/reset');
+    });
+
+    test('an ordinary database error still gets the generic database guidance', () => {
+      const result = classifyAndFormatError(new Error('database is locked'));
+      expect(result).not.toContain('final status could not be saved');
     });
   });
 });
--- a/packages/core/src/utils/error-formatter.ts
+++ b/packages/core/src/utils/error-formatter.ts
@@ -4,6 +4,7 @@
  * Classifies errors and provides user-friendly messages
  * without leaking sensitive information
  */
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 import { WorkflowAdoptionError } from '../operations/workflow-adoption';
 
 /**
@@ -20,6 +21,17 @@ export function classifyAndFormatError(error: Error): string {
   // the generic fallback below.
   if (error instanceof WorkflowAdoptionError) {
     return `⚠️ ${message}`;
+  }
+
+  // The run finished but its terminal status was not recorded, so its row still says
+  // `running` and its true outcome is unknown. Distinct from an ordinary failure: the
+  // generic fallbacks below would tell the user to `/reset`, which fixes nothing and
+  // hides a run that will otherwise sit non-terminal holding its working path.
+  if (error instanceof TerminalStatusWriteError) {
+    return (
+      '⚠️ The workflow ran, but its final status could not be saved, so it may still show ' +
+      'as running. Check `/workflow status` before starting another run on this project.'
+    );
   }
 
   // AI-provider rate-limit / usage-cap classification
--- a/packages/server/src/services/workflow-resume-service.test.ts
+++ b/packages/server/src/services/workflow-resume-service.test.ts
@@ -25,9 +25,11 @@ mock.module('@archon/core/handlers', () => ({
   resolveRunContinuation: mockResolveRunContinuation,
 }));
 mock.module('@archon/core/db/codebases', () => ({ getCodebase: mock(async () => null) }));
+const mockFailWorkflowRun = mock(async () => undefined);
 mock.module('@archon/core/db/workflows', () => ({
   listDueWorkflowContinuations: mockListDueWorkflowContinuations,
   deferWorkflowContinuation: mockDeferWorkflowContinuation,
+  failWorkflowRun: mockFailWorkflowRun,
   WorkflowNotResumableError: class WorkflowNotResumableError extends Error {},
 }));
 mock.module('@archon/paths', () => ({
@@ -43,6 +45,8 @@ mock.module('@archon/workflows/executor', () => ({
   executeWorkflow: mockExecuteWorkflow,
   hydrateResumableRun: mockHydrateResumableRun,
 }));
+
+import { TerminalStatusWriteError } from '@archon/workflows/terminal-status-write';
 
 import {
   resumeWorkflowRunFromServer,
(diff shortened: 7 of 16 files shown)


Wirasm:
# Review report

## Verdict

**Ready — none.** The correction preserves terminal-status-write failures through every recursive recovery boundary, so none can become a normal node or workflow result.

## Accepted contract

Every existing terminal workflow-status write must propagate failure. An executor must not report success or an ordinary failure result after it cannot persist a terminal status; the write must succeed before the workflow reaches a terminal result.

Boundary: change propagation only for the existing `completeWorkflowRun` and `failWorkflowRun` calls identified by #2910. No retries, notification channels, or persistence-format changes. Non-terminal logging and notification failures remain best-effort.

## Reviewed head SHA

`84a896633e10628b7afe6f8c2894beca8f554db0`

## Findings

### Critical

None.

### Important

None.

### Rejected findings

None.

### Suggestions

- Retained non-blocking test suggestion: add an explicit rejection test for the finalizer backstop at `packages/workflows/src/executor.ts:3152-3158`. Its `await failWorkflowRun(...)` already propagates, and it is outside the new recursive recovery mechanism.

## Prior findings

| ID | Prior finding | Current verdict |
| --- | --- | --- |
| R1 | Recursive workflow paths recovered terminal-status-write failures as ordinary child/node/workflow failures. | **Fixed at `84a896633e10628b7afe6f8c2894beca8f554db0`.** `runChildWorkflow` rethrows the marker at `packages/workflows/src/executor.ts:1527-1528`; the 1:1 node boundary rethrows at `packages/workflows/src/dag-executor.ts:8099-8101`; fan-out rethrows a rejected marked slot at `dag-executor.ts:8839-8843`; and both layer recovery boundaries preserve it at `dag-executor.ts:11048-11050` and `:11188-11190`. The parent executor records the marker before rethrowing, so its finalizer does not issue a substitute write (`executor.ts:3078-3080`, `:3152-3158`). Focused 1:1 and fan-out regressions cover this path in `packages/workflows/src/subrun.test.ts:706-809`. |
| R2 | Parent auto-resume suppressed its own terminal-write failure and returned a successful child result. | **Fixed at `84a896633e10628b7afe6f8c2894beca8f554db0`.** `maybeResumeParentRun` rethrows a marked parent-executor failure at `packages/workflows/src/executor.ts:1730-1732`; the parent failure write remains wrapped at `:1743-1748`. The end-to-end regression at `packages/workflows/src/subrun.test.ts:1093-1162` verifies that a rejected parent terminal write escapes after child completion, leaves the parent running, and is attempted only once. |

## Discoveries

0 accepted discoveries. [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/f13c3376-e0d0-4500-83c9-909fc83b43ef/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/f13c3376-e0d0-4500-83c9-909fc83b43ef/discoveries.md) record the consolidated result. Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and the complete correction delta. The accepted original concerns were `code`, `seams`, `tests`, `errors`, and `simplify`; specialists did not rerun. I inspected the four marked DAG terminal-write producers, the marked parent auto-resume write, all recursive `executeWorkflow` callers and their immediate recovery boundaries, the changed finalizer behavior, and the new nested-workflow tests. `bun test src/subrun.test.ts` (87 passing), `bun test src/executor.test.ts` (127 passing), `bun test src/dag-executor.test.ts` (686 passing), and `bun x tsc --noEmit` in `packages/workflows` passed. No required evidence was unavailable.

Wirasm:
## Needs fixes

PR #2921 introduces the right primitive — a typed `TerminalStatusWriteError` marker that a recovery boundary can recognize and rethrow instead of absorbing — and applies it correctly everywhere `dag-executor.ts` writes a terminal status (4/4 sites wrapped, all four internal `instanceof` checks correct, both prior correction-round fixes hold). But the marker only reaches a fraction of the places issue #2910 was about. Two independent, well-evidenced gaps reopen the exact defect class the PR exists to close, on ordinary (not contrived) paths: (1) nine of `executor.ts`'s ten `failWorkflowRun` call sites were never wrapped, so a double-failure during early child-run setup produces an untyped error that `runChildWorkflow`'s own `instanceof` check misses and silently relabels as an ordinary child failure; (2) `TerminalStatusWriteError` is never imported outside `packages/workflows`, so every external caller of `executeWorkflow` — the server's headless/scheduled resume, the chat orchestrator's background dispatch, and its gate-continuation handler — catches a genuine terminal-write failure generically and reproduces the pre-PR "log and continue" pattern, on what the original issue itself called the most unattended path. Two further findings are real but lower severity: `dag-executor.ts`'s conversion reordered a failing write ahead of previously-unconditional user notification, event emission, and telemetry, and the shared `requireTerminalStatusWrite` helper dropped the per-site diagnostic logging every deleted `.catch()` used to provide. Where the PR did apply its fix, the tests are genuine (real rejecting stores, both resolve/reject sides of the deciding promise) — the gap is code coverage, not test rigor.

**4 blocking · 0 non-blocking**

**Validation:** `bun run test` (packages/workflows compound script, all 24+ suites) — PASS, 0 fail. `bun x tsc --noEmit` (packages/workflows) — PASS. `bun run lint` (repo root) — PASS. All run on the reviewed head after `bun install` in the review worktree (node_modules was absent).

### Findings

|---|---|---|---|
| `R1` | Critical | 9 of `executor.ts`'s 10 terminal-write sites are unwrapped; a double failure during child-run setup produces an untyped error `runChildWorkflow` silently relabels as an ordinary child outcome | OPEN |
| `R2` | Critical | `TerminalStatusWriteError` never crosses into `packages/server` or `packages/core`; every external caller of `executeWorkflow` (headless resume, chat background dispatch, gate-continuation) swallows a genuine terminal-write failure generically | OPEN |
| `R3` | Important | The conversion in `dag-executor.ts` reordered the (now-throwing) terminal write ahead of previously-unconditional chat notification, event emission, log-file write, and telemetry at all 4 sites | OPEN |
| `R4` | Important | `requireTerminalStatusWrite` drops the per-site diagnostic logging every deleted `.catch()` handler used to provide; no site logs the original DB error any more | OPEN |

<details>
<summary><code>R1</code> — nine of executor.ts's ten terminal-write sites are unwrapped, and the gap is reachable through an ordinary child-run path</summary>

**Impact:** A workflow that spawns a `workflow:` sub-run, where the sub-run's mandatory early setup (invocation-metadata persistence, run-config resolution, artifacts-dir creation, source resolution, etc.) fails and the *recovery* `failWorkflowRun` call for that failure also fails (the same transient DB condition issue #2910 exists to protect against) produces a plain, untyped `Error`. That error reaches `runChildWorkflow`'s catch, misses `instanceof TerminalStatusWriteError`, and falls through to the pre-PR behavior: a swallowed `cancelWorkflowRun` attempt plus a normal `failOutcome()` returned to the parent DAG. The parent proceeds as if the child simply "errored" — the exact false-ordinary-outcome issue #2910 was filed against, relocated from the sites this PR did convert to the sites it converted incompletely. The same unwrapped pattern also affects the top-level catch's own write (`executor.ts:3092`) and the `finally`-block zombie backstop (`executor.ts:3157`); the `terminalStatusWriteFailed` flag that exists specifically to prevent a second, message-masking write is only set when the *caught* error was already typed — never when the catch's own write (3092) is what fails — so the backstop can still fire a second write over a row whose write channel just proved unreliable.

**Evidence:** `packages/workflows/src/executor.ts:2196-2225` (child metadata persist + its unwrapped recovery write at 2208, reachable unconditionally on every fresh child's first pass via `runChildWorkflow` at `executor.ts:1504`); `executor.ts:1524-1550` (`runChildWorkflow`'s `instanceof TerminalStatusWriteError` check at 1528, and its swallow fallback at 1543-1549); `executor.ts:3078-3092` (`terminalStatusWriteFailed` set only when the incoming error is already typed; the catch's own write at 3092 is unwrapped); `executor.ts:3152-3158` (finally backstop, unwrapped, guarded only by the flag that 3092 never sets). `dag-executor.ts:8845-8846`'s fan-out aggregation comment explicitly documents that it trusts every genuine terminal-write failure to arrive typed — this class of site breaks that assumption for fan-out children too. Test-side confirmation: `executor.test.ts:2963-2982` shows site 3092 is deliberately left to propagate a plain `Error` (proving the omission is a stated design choice at that one site, not incidental), while no test in `subrun.test.ts` monkeypatches `updateWorkflowRun`/`failWorkflowRun` to reject during a *child's* early setup, so the current implementation's actual swallow behavior at these sites is unexercised by the suite.

**Required outcome:** Wrap the seven early-setup sites (`executor.ts:1855, 1885, 2036, 2208, 2480, 2559, 2701`) and the two main-path sites (`3092, 3157`) in `requireTerminalStatusWrite`, matching the treatment already given to `maybeResumeParentRun`'s write (`1743`) and all four `dag-executor.ts` sites. Add at least one regression test that rejects a store call inside a child's early setup (e.g. `updateWorkflowRun` at `2196`, or the recovery `failWorkflowRun` itself) and asserts the parent's `executeWorkflow` call rejects rather than resolving with an ordinary failed-child outcome.

**Class:** Invariant — every `failWorkflowRun`/`completeWorkflowRun` call whose rejection can reach an `instanceof TerminalStatusWriteError` boundary must be wrapped in `requireTerminalStatusWrite`. Search: `grep -n "failWorkflowRun(\|completeWorkflowRun(" packages/workflows/src/executor.ts packages/workflows/src/dag-executor.ts`, cross-referenced against `requireTerminalStatusWrite(`. Clean (wrapped): `dag-executor.ts:12252-12254, 12302-12304, 12384-12386, 12484-12486` (4/4); `executor.ts:1743-1748` (`maybeResumeParentRun`). Affected (unwrapped, reachable): `executor.ts:1855, 1885, 2036, 2208, 2480, 2559, 2701` (early setup, reachable via `runChildWorkflow` on any fresh child); `executor.ts:3092` (top-level catch's own write, reachable via `runChildWorkflow` or via the finalizer-backstop double-write interaction); `executor.ts:3157` (finalizer backstop, reachable via the same `runChildWorkflow` miss when it fires during a child's execution). No member of this file-scoped class was left unexamined.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:silent-failure-hunter`, `prp-core:pr-test-analyzer`

**Disposition:** OPEN.

</details>

<details>
<summary><code>R2</code> — the marker never crosses into packages/server or packages/core; every external caller of executeWorkflow swallows a genuine terminal-write failure</summary>

**Impact:** `TerminalStatusWriteError` is defined and consumed only inside `packages/workflows`. Every caller of `executeWorkflow` outside that package — the CLI aside, which has no local catch and propagates unchanged — catches broadly and treats a genuine, correctly-typed `TerminalStatusWriteError` (one that already propagated cleanly out of a properly-wrapped `dag-executor.ts` site) exactly like any ordinary execution error: log it, then make a *second* best-effort write attempt that is itself swallowed by its own `.catch()`. If that second write also fails — the same transient condition already in play — the run row is stuck at `running` forever with zero further signal, because the scheduler that would otherwise revisit it (`listDueWorkflowContinuations`) only selects `paused`/`failed` rows, never `running`. This reproduces issue #2910's exact symptom one layer outside this PR's diff, and on the path the issue itself named as mattering most: an unattended, headless resume nobody is watching. The chat-driven paths fail the same way with a misleading user-facing message instead of an accurate one (e.g. `continueResolvedGateRun` tells the user to `retry with /workflow resume <id>` — wrong advice when the row's true status is unknown, not simply stale).

**Evidence:** `packages/server/src/services/workflow-resume-service.ts:157-213` — `resumeWorkflowRunFromServer`, wired live via `scanDueWorkflowContinuations`'s `setInterval` (started at `packages/server/src/index.ts:980`) and from gate-approval/web-resume routes in `packages/server/src/routes/api.ts`; its `void executeWorkflow(...).then(onFulfilled, onRejected)` `onRejected` handler (199-212) never checks `instanceof TerminalStatusWriteError`, logs generically (`workflow_resume_headless_execute_failed`), then retries `workflowDb.failWorkflowRun(...)` wrapped in its own swallowing `.catch()`. `packages/core/src/orchestrator/orchestrator.ts:707-716` (`dispatchBackgroundWorkflow`) — same shape, tells the chat user "Workflow X failed" while the row may still say `running`. `packages/core/src/orchestrator/orchestrator-agent.ts:1694-1702` (`continueResolvedGateRun`) — same shape, no write attempted at all, just misleading retry advice. `grep -rn "TerminalStatusWriteError" packages/` confirms the type is imported only in `packages/workflows/src/{executor.ts,dag-executor.ts,executor.test.ts}`. `executor.test.ts:2984-3006` shows the codebase already models "don't recover a rejected terminal write with a second failure write" *inside* `executeWorkflow` — the same discipline is simply absent at every point ownership crosses this package boundary.

**Required outcome:** Export `TerminalStatusWriteError` (or a narrow re-export) from `packages/workflows`'s public surface, and branch on `instanceof` at each of these three confirmed boundaries (skip the redundant write, escalate distinctly — e.g. a dedicated event/log tag rather than the generic one). Two further call sites (`orchestrator-agent.ts:1280/1325/1417`, `dispatchOrchestratorWorkflow`, which has no local catch and propagates into the chat message-handling loop) were traced but not resolved to a terminus within this review's scope and must be checked before the class can be called closed.

**Class:** Invariant — every external consumer of `executeWorkflow`'s rejection must distinguish a `TerminalStatusWriteError` from an ordinary error before applying its own recovery. Search: `grep -rn "executeWorkflow(" packages --include="*.ts" | grep -v test | grep -v "function executeWorkflow"` → 8 direct call sites. Clean: `executor.ts:1504` (`runChildWorkflow`), `executor.ts:1716` (`maybeResumeParentRun`), `packages/cli/src/commands/workflow.ts:2676` (`try/finally` only, no catch — propagates both ordinary and typed errors unchanged to the CLI's own exit-code path). Affected: `packages/core/src/orchestrator/orchestrator.ts:627/707-716`, `packages/core/src/orchestrator/orchestrator-agent.ts:1694-1702`, `packages/server/src/services/workflow-resume-service.ts:157/199-212`. Unexamined (reached from `orchestrator-agent.ts:1280/1325/1417` via `dispatchOrchestratorWorkflow`, which has no local catch and propagates further into the chat message-handling loop; three additional `dispatchOrchestratorWorkflow` callers at `orchestrator-agent.ts:3243/3591/3681` were not traced to a terminus) — not claimed clean.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:silent-failure-hunter`

**Disposition:** OPEN.

</details>

<details>
<summary><code>R3</code> — dag-executor.ts's conversion reordered the throwing write ahead of previously-unconditional notification, event, and telemetry side effects</summary>

**Impact:** At all four `dag-executor.ts` terminal-write sites, `requireTerminalStatusWrite(...)` now executes *before* `logWorkflowError`/`logWorkflowComplete`, the `workflow_failed`/`workflow_completed` event emission (breaking live SSE/dashboard updates for that run), `captureWorkflowCompleted` telemetry, and the `safeSendMessage` chat notification. None of these depend on the DB write's success — pre-PR, the write was wrapped in a local try/catch that logged and fell through, so all of them ran regardless. Now, when the write throws, execution aborts at that line: the operator gets no chat notification at all (not even the old "warning" message), no dashboard event, no log-file entry, and telemetry silently undercounts a run that genuinely completed or failed. This is a new, unforced silence around a failure that (assuming R1/R2 are fixed) is at least propagating loudly elsewhere — but it directly works against issue #2910's own "no new silent path" requirement.

**Evidence:** `packages/workflows/src/dag-executor.ts:12252-12274` (quota-exceeded path), `12302-12324` (node-failure path), `12384-12409` (evidence-gate path), `12481-12521` (happy-path completion) — in each, the wrapped write is the first statement, followed by the now-unreachable-on-failure side effects. The stale "DO NOT throw — outer executor.ts catch would duplicate workflow_failed events" comments at `12272, 12322, 12408` describe behavior the code directly above no longer has (it can and does throw now).

**Required outcome:** Reorder each block so the terminal write is the last statement — logging, event emission, telemetry, and chat notification run unconditionally first, then the (possibly-throwing) write runs last. Update the three stale "DO NOT throw" comments to reflect the new ordering and why it's safe (the `TerminalStatusWriteError` branch in `executor.ts` already skips its own duplicate emit).

**Disposition:** OPEN.

</details>

<details>
<summary><code>R4</code> — requireTerminalStatusWrite drops the diagnostic logging every deleted .catch() handler used to provide</summary>

**Impact:** Every `.catch()` handler this PR removed had its own structured-log call with a distinct, greppable tag (`dag_db_fail_failed`, `dag_db_complete_failed`, `workflow.parent_auto_resume_fail_mark_failed`, `executor.backstop_fail_failed`, and seven more in `executor.ts`'s early-setup catches). None of that logging was replaced. `requireTerminalStatusWrite` itself never logs, and none of the five `instanceof TerminalStatusWriteError` rethrow points log either. The original DB error's message and stack now travel silently inside `.cause` until some caller several frames away happens to log the generic rejection — today, only one of the four external callers of `executeWorkflow` does so at all (and, per R2, without distinguishing this failure class from an ordinary one).

**Evidence:** `packages/workflows/src/terminal-status-write.ts:15-21` (no logging call in `requireTerminalStatusWrite`); confirmed absent at every rethrow site: `dag-executor.ts:8100, 8841, 11049, 11189, 12252, 12302, 12384, 12484`; `executor.ts:1528, 1731, 1743, 3078-3081`.

**Required outcome:** Add one `getLog().error(...)` call inside `requireTerminalStatusWrite` itself (matching the file's existing per-module lazy-logger pattern in `executor.ts:127-130` and `dag-executor.ts:594-596`), so every current and future call site gets the diagnostic line for free instead of relying on ~14 individual sites to remember it.

**Disposition:** OPEN.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | `R1`, `R2` |
| `seams` | `R1`, `R2` |
| `simplify` | `R3`, `R4` |
| `errors` | `R1`, `R2` |
| `tests` | `R1` (coverage gap tied to the same causal defect; no independent test-quality finding — existing tests for the already-fixed boundaries are genuine, real-store, both-sides-of-the-decision tests) |

#### Validation

|---|---|---|
| `bun install` (review worktree) | ran once, needed | `node_modules` was absent in the fresh detached worktree; 2556 packages installed |
| `bun run test` (packages/workflows) | PASS | 0 fail across all suites in the compound script, including `dag-executor.test.ts`, `executor.test.ts`, `subrun.test.ts` |
| `bun x tsc --noEmit` (packages/workflows) | PASS | clean exit |
| `bun run lint` (repo root) | PASS | clean, no warnings, across all packages |
| Composition with `dev` | not blocking, but relevant | Branch is 16 commits behind `dev` (merge-base `761b603a`, dev head `ad0f9cf5`), including PR #2914 which also touches `packages/workflows/src/executor.ts` (the same `preCreatedRun` invocation-metadata block `R1` cites). `git merge --no-commit --no-ff` auto-merges cleanly with no textual conflicts (verified, then aborted). A dev merge is needed before landing regardless of these findings. |
| CLI exit-code composition (PR #2925) | checked, clean | `packages/cli/src/commands/workflow.ts:2635-2688` has no catch around `executeWorkflow`, only a `finally` — both ordinary and `TerminalStatusWriteError` propagate unchanged to the CLI's top-level `catch` in `cli.ts:1237-1247`, which exits non-zero regardless of type (a `TerminalStatusWriteError` falls through `resolveCliExitCode`'s `WorkflowRunFailedError`-only check to the default exit code `1`, not the reserved detached code `90` — within the narrow 500ms startup window this reads as "launch failed" rather than "terminal write failed," a mislabeling but still a loud, correct-direction failure, not a false success). Not filed as a blocking finding: the operator still gets a non-zero exit and a logged message either way. |
| SQLite retry semantics | checked, clean | `packages/core/src/db/adapters/sqlite.ts` already retries busy-lock contention up to 5s beneath the store layer; the new fail-loud behavior applies only once that budget is exhausted — not an unstated regression, a narrower safety net than before, as intended. |

**Scope note:** `errors` and `tests` were added to the standing `code`/`seams`/`simplify` default because the operator's work order gave detailed adversarial anchors squarely in silent-failure and test-depth territory (marker relabeling, swallow-site enumeration, retry semantics, test rigor).

</details>

Wirasm:
## Delta verification — round 2 (fix round)

**READY TO MERGE.** All four round-1 findings (`R1`, `R2` Critical; `R3`, `R4` Important) are FIXED on `fbac36bf`, independently re-verified against the code rather than the fix report's own account. No new defect surfaced.

### R1 — FIXED

Re-ran the round-1 enumeration myself: `grep -n "failWorkflowRun(\|completeWorkflowRun(" packages/workflows/src/{executor,dag-executor}.ts`, then read the full context around every hit. Confirmed 14/14 wrapped in `requireTerminalStatusWrite` — all 10 `executor.ts` sites (1753, 1866, 1899, 2053, 2244, 2522, 2599, 2748, 3195, 3223) and all 4 `dag-executor.ts` sites (12273, 12329, 12424, 12538), each with its own `{ workflowRunId, site }` tag. R1's original concrete failure chain (`executor.ts:2244`'s invocation-metadata recovery write → `runChildWorkflow`'s `instanceof` check at what is now line ~1537) is closed: the site is wrapped, so a double failure there now produces a real `TerminalStatusWriteError` that `runChildWorkflow` correctly rethrows instead of relabeling. Also confirmed the flag fix at `executor.ts:3190-3199`: the top-level catch's own write is now wrapped in a local `try/catch` that sets `terminalStatusWriteFailed = true` before rethrowing, closing the double-write-masking gap the original review found (the backstop at 3223 can no longer fire a second, message-masking write over that same failure).

**The `cancelWorkflowRun` question (`executor.ts:1552`) — my second opinion: leave it unwrapped, not a blocker.** I traced the reachability independently. The wedge guard at 1552 only catches errors from code that runs before ANY `failWorkflowRun` is ever attempted — I confirmed `deps.store.loadConfig`/`getCodebaseEnvVars` (~1908-1910) sit outside every enclosing catch, and the run's `status → 'running'` flip doesn't happen until `executor.ts:2926`, well after every wrapped early-setup guard. So in the code as it exists today, this wedge guard fires against a row that is still `pending` in the overwhelmingly common case, and `getActiveWorkflowRunByPath` (`packages/core/src/db/workflows.ts:658-674`) already excludes `pending` rows older than `STALE_PENDING_AGE_MS` (5 minutes) from holding the working-path lock — unlike `running`, which has no such exemption. Even in the residual case (the compensating `cancelWorkflowRun` itself also fails), the parent DAG still gets an honest, correctly-labeled "child failed" outcome, not a false success or a masked ordinary failure — the blast radius is confined to that one child row, not the operator-visible result the original review's R1/R2 were about. That's a materially smaller harm than #2910's core complaint, and it sits outside both the issue's own enumerated class (`failWorkflowRun`/`completeWorkflowRun`, not `cancelWorkflowRun`) and outside the concrete reachable chain I evidenced in round 1. I did not exhaustively re-derive every path to function entry to prove the guard's own "(or `running` on a later window)" caveat is fully unreachable — worth a one-line wrap as a follow-up if you want it fully closed, but I would not hold the PR on it.

Tests: `executor.test.ts:2979-3126` and `subrun.test.ts` (new describe block) both drive real rejecting-store scenarios through the actual recursion, not stubbed classification — confirmed by reading them directly.

### R2 — FIXED

`packages/workflows/package.json` now exports `./terminal-status-write`. Confirmed all three named boundaries branch correctly: `workflow-resume-service.ts:199-230` (skips the compensating write on the marker, notifies the visible conversation when one exists), `orchestrator.ts:707-720` (skips the write, distinct log tag, message says "final status could not be saved" rather than asserting failure), `orchestrator-agent.ts:1693-1702` (drops the `/workflow resume` advice, points at `/workflow status` instead). Traced the two previously-unresolved call sites (`orchestrator-agent.ts:1281/1326/1418` via `dispatchOrchestratorWorkflow`, no local catch) through to their actual terminus and confirmed it: `handleMessage`'s catch at `orchestrator-agent.ts:2691-2692` calls `classifyAndFormatError`, which now has a `TerminalStatusWriteError` branch (`error-formatter.ts:29-33`) ahead of the generic fallback that used to tell the user to `/reset`. All four test files (`workflow-resume-service.test.ts`, `orchestrator-isolation.test.ts`, `orchestrator-agent.test.ts`, `error-formatter.test.ts`) genuinely construct `new TerminalStatusWriteError(...)` and assert on real call counts / message content, both sides of the branch.

### R3 — FIXED

Read all four `dag-executor.ts` blocks in full: the write is now the last statement in each, after `logWorkflowError`/`logWorkflowComplete`, the event emit, `unregisterRun`, and `safeSendMessage`. The extension into `executor.ts` (top-level catch at 3178-3199, and the five setup guards) is the same defect shape, not scope creep — each now notifies before the (now-throwing) write, which is exactly what R1's own wrapping made necessary to avoid re-silencing the same class of side effect it had just un-silenced in `dag-executor.ts`.

### R4 — FIXED, and mutation-verified myself

`requireTerminalStatusWrite` takes a required `{ workflowRunId, site }` and logs under `workflow.terminal_status_write_failed` before rethrowing (`terminal-status-write.ts:46-59`). I did not just read this — I mutated the wrapper in my own worktree to drop the log call, ran `bun test src/terminal-status-write.test.ts`, and got the exact failure the fix report claims (`toHaveBeenCalledTimes` expected 1, received 0); restored and confirmed green again. The `site` tags reuse the original per-call `.catch()` tags verbatim (spot-checked `workflow.parent_auto_resume_fail_mark_failed` at 1753 and `workflow.invocation_metadata_failure_record_failed` at 2244).

### Validation (run independently, not copied from the fix report)

| Check | Result |
|---|---|
| `bun install` (fresh delta worktree) | needed, ran clean |
| `bun run test` — `packages/workflows` | PASS, 0 fail (116+ across all files) |
| `bun run test` — `packages/core` | PASS, 0 fail |
| `bun run test` — `packages/server` | PASS, 0 fail |
| `bun x tsc --noEmit` — workflows / core / server | PASS, all clean |
| `bun run lint` (repo root) | PASS, clean |
| Mutation spot-check — `terminal-status-write.ts` logging (R4) | confirmed seen-red, then restored clean |
| Dev-merge integrity | `#2914`'s `preCreatedRun` invocation-metadata block (write-once `working_path`, `isolation`/`isolation_env_id` stamping, `executor.ts:2213-2255`) is fully intact post-merge; R1's wrap layers on top without disturbing it |
| CI on `fbac36bf` | `docker-build` PASS, `test (ubuntu-latest)` PASS, `changes`/`postgres-parity`/`schema-upgrade` PASS; `test (windows-latest)` and CodeRabbit still in progress at time of this comment — not blocking, since every check they'd run was independently reproduced above |

No residual findings block merge. The one open item — `executor.ts:1552`'s `cancelWorkflowRun` — is a legitimate, honestly-disclosed, lower-severity follow-up, not a defect in what this PR claims to fix.

Issue #2908, opened 2026-08-28
feat(core): tell a waiter when a run's owning process is gone


Wirasm:
## Problem

#2745's wait primitive derives attention from the run row and wakes on the existing workflow-event notification channel. It has one honest blind spot: a waiter cannot distinguish a run whose owning process is still working from one whose owner was killed. Both look like "not terminal yet". The wait simply continues until its caller gives up.

The primitive to fix that already exists but is scoped to one caller. `packages/cli/src/utils/detached-run-control.ts` publishes a per-run, user-scoped endpoint — a Unix socket on POSIX, a named pipe on Windows — owned by the detached CLI child, and today it is inbound-only (`stop` / `terminate`). Its reachability is already a truthful liveness answer; nothing consumes it as one, and only one kind of owner publishes it.

## Desired outcome

- The per-run owner endpoint moves to `@archon/core` as an engine concept: "the live owner of this run". Every process that executes a run publishes one — detached CLI child, foreground CLI, the in-process server — not just the detached child.
- The owner pushes its attention signal on that endpoint before exit, so a waiter wakes without waiting for the interval.
- `waitForRunAttention` gains an `owner_lost` result variant: the endpoint is unreachable (or closed with no signal) and the run row is still non-terminal. It reports that and the status it observed. It never guesses a terminal state and never mutates the run — surfacing ambiguous ownership for an operator decision is the required behavior (`AGENTS.md`, "Do not guess lifecycle ownership").
- The CLI `workflow wait` host renders `owner_lost` distinctly and points at `workflow abandon`, which is already the operator's explicit action for a dead owner.

## Relation to #2325

#2325 narrows to truthful live-versus-orphan evidence, and endpoint reachability is a stronger signal than the `last_activity_at` staleness threshold that issue found unreliable (a live Codex tool interval can outlast five minutes because `last_activity_at` only advances when the provider stream yields). These should share the primitive. Lifecycle mutation stays operator-owned in both.

## Depends on

#2745 — the capability and the `RunWaitResult` union this adds a variant to.

Wirasm:
## Trace evidence for the endpoint-as-liveness-signal claim

From the #2745 design trace, the ordering that makes the owner endpoint a *truthful* liveness answer, not just a convenient one:

[...]

Issue #2907, opened 2026-08-28
feat(core): route run attention to the run owner's connected platforms


Wirasm:
## Problem

#2745 gives the engine one durable answer to "this run needs someone, and here is why" and one host that consumes it: a process that blocks on `archon workflow wait <run-id>`. That serves an orchestration agent that launched the run and is still around to wait.

It does not serve the human who is not waiting. A run started by a schedule (#998), by the HTTP API, or overnight has an owner (`remote_agent_workflow_runs.user_id`) with connected platform identities (`remote_agent_user_identities`), and none of them are told anything. An approval gate nobody learns about is operationally indistinguishable from a hang, and a failed overnight run that reports to nobody delays exactly the work the automation was supposed to absorb.

This is the original framing of #2745, split out when that issue was narrowed to the engine capability.

## Desired outcome

- The run owner's connected platform identities become a resolvable **notification route**.
- A host consumes the `RunAttention` values #2745 produces and delivers them to that route — Slack DM, Telegram, web/console inbox — regardless of which surface started the run.
- Per-user opt-in and event selection. A surface that already shows the state inline, such as an active chat thread the run was started from, stays as it is; the routing exists for runs with no watching surface.
- The delivered notification carries the one-click continuation: the run id, the gate's declared decision ids (#2707's `workflow respond` vocabulary), or the failure summary.

## Boundary

- This is a **host**, not a second capability. The vocabulary of "what does this run need" is owned by #2745's `runAttention`; this issue owns routing, recipient preferences, and delivery.
- Delivery state (per-recipient sent/failed/retry) is genuinely different from waiting and may justify persistence that #2745 deliberately does not have. Settle that here, not there.
- #1699 asks for a sound notification in the UI when interaction is needed — a UI expression of the same underlying need, and another consumer of the same capability.

## Depends on

#2745 — the engine capability and the `RunAttention` contract.

PR #2923, opened 2026-08-28
feat(engine): a host can wait for a run instead of polling it

### LINKED ISSUE #2745: feat(core): wake a waiting host when a run needs someone

Wirasm:
## Problem and outcome

A run started outside a chat thread — CLI `--detach`, the HTTP API, a future scheduled trigger (#998) — reaches a state it will not leave on its own, and nothing tells whoever launched it. The knowledge is durable in the run row; the delivery did not exist, so a launching agent had to poll `archon workflow get` in a loop.

- **Outcome:** `archon workflow wait <run-id>` blocks until the run finishes or parks on an unresolved human gate, then prints what it needs. Paired with the `--detach --json` ack that now carries `runId` (#2872), launch-then-wait is expressible end to end. Polling becomes a diagnostic, not the orchestration contract.
- **Invariant:** When a run enters a state it will not leave without someone acting — `completed`, `failed`, `cancelled`, or paused on an unresolved human gate — every host waiting on that run is told at least once, without polling run status, and no waiter ever mutates or synthesizes run lifecycle state.
- **Scope boundary:** No schema change, no outbox or delivery table, no new terminal vocabulary, no per-user notification routing (#2907), no owner-liveness signal (#2908). The NOTIFY channel keeps its existing name despite being a misnomer — an older binary sharing the database listens on that exact string. The executor's `isApprovalContext` uses are deliberately untouched: they read gate *detail*, not the attention question.

## Review guidance

- **Feedback requested:** Correctness of the projection's derivation order, and whether the four migrated call sites preserve their contracts.
- **Start here:** `packages/workflows/src/schemas/workflow-run.ts` — `runAttention` and the `RunAttention` union. Everything else consumes it.
- **Review order:**
  1. `packages/workflows/src/schemas/workflow-run.ts` — the projection (new).
  2. `packages/core/src/operations/workflow-operations.ts` — `assertApprovable` / `assertRejectable` migrated onto it.
  3. `packages/server/src/routes/api.ts` — `pausedGateBlocker`, one shared precondition for approve / reject / respond.
  4. `packages/core/src/orchestrator/prompt-builder.ts` — `formatPausedGateSection` now takes the run.
  5. `packages/core/src/services/run-attention-watch.ts` — the waiter (new).
  6. `packages/cli/src/commands/workflow.ts` — `workflowWaitCommand` and its output contract.
- **Lower-attention areas:** `packages/cli/src/cli.ts` and `args.ts` (verb wiring), `cli.md` (docs), and the three-site `WORKFLOW_EVENT_NOTIFY_CHANNEL` constant extraction (the string is byte-identical; the Postgres adapter test covers the trigger).
- **Known risk or uncertainty:** The waiter's notification filter matches only the watched run id, so a transition on a run deeper in a `blocked_on_child` chain arrives on the 1s interval rather than the doorbell. Latency only — the row read is what decides.

## Solution

**One derivation, then one consumer.** The decision "is this run waiting on a human, and on which run does that human act" was written out longhand at four independent sites, each rendering the same four steps in a different shape — and in **three different orders**. The `child_workflow` redirect, the load-bearing part, survived only inside an error string, which is why a wake signal could not reuse it.

`runAttention(run)` in `@archon/workflows` is now the only place that decision is made. It is pure — no database, no clock, no I/O — and lives beside `isGateResolved` / `isApprovalContext`, which it subsumes as a decision while keeping both as its building blocks.

**`blocked_on_child` is deliberately not an answer.** `pauseParentOnChild` has two call sites, and the second fires when the child is `paused` **or `running` or `pending`** (`dag-executor.ts`, "Still in progress (awaiting a human or a concurrent run)"). Asserting `awaiting_human` from the parent row would wake hosts for normal progress; returning nothing would strand them when the child really is on a gate. Neither is truthful from the parent row alone, so the projection reports the block and `waitForRunAttention` — which has database access — walks the chain to the run that actually needs a person.

**The run row is the authority; the shipped `pg_notify` trigger is only a doorbell.** Every wake re-reads `remote_agent_workflow_runs` before deciding anything. That is forced by evidence, not caution: `resolveAndCancelApprovalGate` writes a terminal `cancelled` while inserting only an `approval_received` row, a `child_workflow` pause persists no event at all by design, and `approval_requested` is written fire-and-forget before the awaited pause. So the interval runs at the **same 1s cadence on both dialects** — unlike the dashboard poller, which safely relaxes to 10s with NOTIFY because every transition it tracks inserts an event.

The waiter follows the existing core-owns-capability / host-drives-it seam (`listDueWorkflowContinuations` + `workflow-resume-service`). It performs no writes.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A detached launch returns an ack; learning the outcome means a `workflow get` polling loop. | `archon workflow wait <run-id>` returns once, carrying the terminal status or the gate address. |
| **Failure behavior** | Approving a run whose gate type this build cannot resolve returned an opaque `500 Failed to approve workflow run`; a corrupt sub-run pointer produced `…waiting on sub-run <unknown>. Approve or reject the child run instead.` | Both are `400` with the reason. The corrupt pointer says `Run <id> cannot be resolved: blocked on a sub-run at node '<node>' but the child run id is missing.` |

**Three deliberate exceptions to "behavior-preserving", all tested.** Every other user-visible string at the four migrated sites is byte-identical and pinned by test, but these three changed on purpose:

1. **Unrecognized gate type: 500 → 400.** The approve route had no such check; `approveWorkflow` threw later and the route returned an opaque `500 Failed to approve workflow run`. It is now a 400 naming the type — the same mapping the route's own comment already described for the child redirect.
2. **A corrupted child pointer no longer renders `<unknown>`.** It was interpolated into `/workflow approve <unknown>`, a command nobody can run. It is now `Run <id> cannot be resolved: blocked on a sub-run at node '<node>' but the child run id is missing.`
3. **`formatPausedGateSection` returns `''` for a conversation-scoped run paused on a `wait:` node.** It previously rendered the "approval context is missing or malformed… resolve it explicitly with `/workflow approve`" text for a run that has no gate at all. Arguably a correctness improvement rather than a regression: a durable wait is owed nothing by anyone, so telling the agent to approve it was false. Called out here because it is a user-visible string that changed.

## Architecture

```mermaid
flowchart TD
  subgraph owner["Owning process (detached CLI child, server, container)"]
    EX[DAG executor finishes or pauses on a gate]
    EX --> TX[["one transaction:<br/>run row status write<br/>+ workflow event insert"]]
  end

  TX --> ROW[("remote_agent_workflow_runs<br/>status · completed_at · metadata<br/><b>authority</b>")]
  TX --> EVT[("remote_agent_workflow_events<br/><b>doorbell only</b>")]

  EVT -. "Postgres only: pg_notify(run_id)<br/>EXISTING trigger" .-> LISTEN[["DbNotificationListener<br/>EXISTING, core-owned"]]

  subgraph core["@archon/core"]
    WAIT["waitForRunAttention(runId)<br/>NEW"]
  end
  subgraph wf["@archon/workflows"]
    PROJ["runAttention(run)<br/>NEW"]
  end

  ROW --> WAIT
  LISTEN -. wake .-> WAIT
  TICK["1s interval backstop<br/>(only wake source on SQLite)"] -. wake .-> WAIT
  WAIT --> PROJ

  WAIT --> CLIH["archon workflow wait &lt;run-id&gt;<br/>NEW"]

  PROJ --> A1["assertApprovable / assertRejectable"]
  PROJ --> A2["server approve/reject/respond routes"]
  PROJ --> A3["orchestrator prompt-builder"]
```

### Changed seams

|---|---|---|
| `runAttention → assertApprovable / assertRejectable` | Both branch on the projection instead of re-probing metadata; they still read the gate context afterwards, which their callers need. | `workflow-operations.test.ts` — 72 pass, including new cases pinning the child-redirect and already-resolved strings for both verbs |
| `runAttention → approve / reject / respond routes` | Three inline precondition blocks collapse into one `pausedGateBlocker`. `requireReadableGate` mirrors the core split: approve refuses an unreadable gate, reject and respond still resolve it. | `api.workflow-runs.test.ts` — 132 pass, 130 pre-existing unchanged |
| `runAttention → formatPausedGateSection` | `PausedGateContext` now carries the run row instead of `runId` + `workflowName` + raw `approval`; `containerRun` is derived rather than passed. | `prompt-builder.test.ts` — 31 pass |
| `waitForRunAttention → archon workflow wait` | New core export, new CLI verb. Exit codes describe the command: `0` the run said something, `3` deadline with the run still live, `1` the wait failed. | `workflow.test.ts` (9 new cases), `workflow-wait.integration.spec.ts` |
| `WORKFLOW_EVENT_NOTIFY_CHANNEL` | The channel name was three hand-copied literals (trigger SQL, `PgNotifyListener`, and now the waiter); it is one exported constant. The **string is unchanged**. | `packages/core/src/db/adapters/types.ts`; `check:schema-upgrades` and the Postgres adapter tests pass |

## Validation

- `bun run validate` — exit 0 — generated-artifact checks, all 13 package type-checks, lint at `--max-warnings 0`, prettier, and the full suite (165 test groups, 0 fail).
- `bun run check:schema-upgrades` — `OK`, every shipped vintage converges — and `git diff -- migrations/ bundled-schema.generated.ts` is empty, so the feature reads only `remote_agent_workflow_runs`.
- `bun --cwd packages/cli test src/commands/workflow-wait.integration.spec.ts` — 6 pass, run four consecutive times. Real detached runs, real SQLite, and `wait` running as **its own OS process**: `completed` and `failed` wake it while the run is live; `cancelled` wakes it when a third process runs `workflow cancel`; an out-of-process gate wakes it with `awaiting_human` addressed at the gate node; the deadline case exits `3`; an unknown id exits `1`. The waiter issues no status-polling command of its own — `workflow get` runs only afterwards, to show the detail is still there.
- `bun --cwd packages/core test src/services/run-attention-watch.test.ts` — 21 pass. Covers the chain in both dangerous directions: a parent blocked on a merely-running child wakes nobody, the same parent wakes with the child's `awaiting_human` once the child hits its gate, a grandparent→parent→child chain resolves to the deepest run needing a human, a dangling pointer and an over-long chain are `unreadable` rather than an assumed state, a terminal child keeps the waiter waiting, and a resolved gate and a `wait:` pause wake nobody. The test mounts `../db/workflows` with **only** `getWorkflowRun` defined, so any attempted write would throw rather than silently mutate.
- `bun --cwd packages/workflows test src/schemas.test.ts` — 195 pass, including a case per `SuspendReason` plus `undefined`, every terminal status, and the `metadata.approval` / `metadata.wait` exclusivity the write helpers currently guarantee.
- **Windows CI is red on this PR for a pre-existing product defect, not this change — see #2927.** `registerFolder` stores `default_cwd` via `fs/promises.realpath` while the CLI's project gate looks it up via `fs.realpathSync`; on Windows only the former expands an 8.3 short path (`RUNNER~1` → `runneradmin`), so an exact-match lookup misses and every command except `workflow run --folder` (which has an escape hatch) is refused with "Not in a git repository". Proven from the CI logs and from the repo's own pre-existing note at `packages/core/src/orchestrator/orchestrator.test.ts:1602-1603`. This spec is simply the first test to run a *second* `archon` command in a non-git folder-project directory on Windows CI. The fix needs a migration decision about which canonical form wins for already-stored rows, so it is filed rather than folded in here.
- **Not verified:** the Postgres NOTIFY fast path is exercised only through a mocked `DbNotificationListener` (payload match, foreign-payload rejection, unsubscribe on exit). No Postgres instance was available here. The interval backstop is the correctness path on both dialects by design, and the trigger and channel name are unchanged, so a failed notification degrades to a ≤1s wake rather than a missed one.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No schema change and no persisted-format change. The four migrated sites keep their current user-visible strings, with the one `<unknown>` exception described above. | `check:schema-upgrades OK`; empty diff under `migrations/` |
| Rollout / rollback | Purely additive: one CLI verb, two core exports, one type. Reverting removes them and the migrated call sites with them. | — |
| Observability | The waiter logs its wake source (`immediate` / `notify` / `interval`) and the resolved attention kind at debug, so a slow wake is diagnosable without a new metric. | `run-attention-watch.ts` |
| Documentation / communication | New `### workflow wait` section in the CLI reference with the exit-code table and the `attention.kind` vocabulary, a cross-reference from the `--detach` row so the ack-then-wait pairing is visible where detach is documented, and `--timeout` in Global Options. | `packages/docs-web/src/content/docs/reference/cli.md` |

## Links

- Closes #2745
- Depends on #2872
- Depends on #2906
- Related #2907
- Related #2908
- Related #2910
- Related #2927 — pre-existing Windows defect this PR's tests exposed (see Not verified)

## Summary by CodeRabbit

* **New Features**
  * Added `workflow wait <run-id>` to monitor workflow completion or response-required states.
  * Supports JSON output, optional timeouts, and clear results for completion, attention, cancellation, expiration, and missing runs.
  * Improved handling of paused workflows, child runs, approval gates, and malformed gate details.

* **Bug Fixes**
  * Improved error messages for missing child runs and unsupported gate types.
  * Standardized workflow event notifications across supported database integrations.

* **Documentation**
  * Added CLI reference documentation and usage examples for `workflow wait`.


Wirasm:
## Needs fixes

The design and its implementation are sound: `runAttention` is a single, well-ordered, pure derivation that correctly subsumes the four sites that used to answer "does this run need a response, and on which run" independently and inconsistently; `waitForRunAttention` correctly treats the run row as authority, never trusts a NOTIFY payload as the answer, walks the `blocked_on_child` chain to the run that actually needs a person, performs zero writes (enforced in its own test by a DB mock exposing only the read function), and cleans up its listener/timers on every exit path. The four-site migration preserves every user-visible string byte-for-byte except two narrowly documented, deliberate improvements (unrecognized gate type now 400s instead of 500ing; a corrupted child pointer names itself instead of interpolating `<unknown>`). Schema is untouched, the `archon_dashboard_event` channel string is unchanged, and the `awaiting_human`→`awaiting_response` rename is complete with zero stale references.

What blocks merge is external to the design: **CI is red on `test (windows-latest)` at this exact head** — 5 of the 6 new tests in `workflow-wait.integration.spec.ts` fail with the wrong exit code, Windows-only. This is a genuine PR-caused failure of a required gate, not one of the two flakes already fixed on `dev` today. Two non-blocking findings round out the picture: one real gap in the new server-route helper's exhaustiveness (worth a two-line fix before merge even though nothing is wrong today), and one undocumented-but-correct behavior change in the chat prompt builder.

**1 blocking · 3 non-blocking**

**Validation:** `bun run validate` passes locally at this head (165 test groups, 0 fail) and `check:schema-upgrades` passes; CI is red on Windows only (see R1).

### Findings

|---|---|---|---|
| `R1` | Important | `test (windows-latest)` fails at this exact head — 5/6 new `workflow wait` integration tests exit 1 instead of 0/3, Windows-only | OPEN |
| `R2` | Suggestion | `pausedGateBlocker` (server route helper) is the only one of four `RunAttention` consumers with a non-exhaustive switch; a future 5th kind silently becomes a 400→500 regression there while the other three fail to compile | OPEN |
| `R3` | Suggestion | `formatPausedGateSection` now returns `''` for a `wait:`-paused conversation instead of the old "missing or malformed... use `/workflow approve`" text — a real, deliberate, tested behavior change not itemized in the PR's "byte-identical" claim | OPEN |
| `R4` | Suggestion | `pendingDoorbell` coalescing flag in `waitForRunAttention` guards a race no test exercises; removable at the cost of a ≤1s latency window already tolerated elsewhere | OPEN |

<details>
<summary><code>R1</code> — CI fails on Windows: the new `workflow wait` integration spec is broken there</summary>

**Impact:** A required merge gate (`Test Suite`, which aggregates `test (ubuntu-latest)` and `test (windows-latest)`) is red at the exact head under review. This is not cosmetic — the failing tests are the PR's own cross-process regression proof for the feature's headline behavior (`archon workflow wait` against a real detached run), and they fail specifically on Windows while passing on Ubuntu.

**Evidence:** Confirmed via `gh api repos/coleam00/Archon/actions/runs/33173070434` that the run's `head_sha` equals the reviewed head, `conclusion: failure`. Job logs (`gh api repos/coleam00/Archon/actions/jobs/98854893857/logs`) show:
```
packages\cli\src\commands\workflow-wait.integration.spec.ts:232:24
error: expect(received).toBe(expected)
Expected: 0
Received: 1
```
repeated for 4 more assertions in the same file (`workflow-wait.integration.spec.ts:232`, `:282`, `:347`, plus one more), each time `exitCode` is `1` where the test expects `0` or `3`. Only the one test that never launches a detached run (`exits 1 for a run id that names nothing`) passes. Final tally: `@archon/cli:test | 1 pass / 5 fail`. Each failing test's temp-dir cleanup then also logs `EBUSY: resource busy or locked` on Windows, consistent with a process still holding the SQLite file open when the assertion fires. `test (ubuntu-latest)` passes at the same head — this is Windows-specific.

This is not one of the two flakes the task brief named as already fixed on `dev` (#2915, #2916) — it is new, isolated entirely to this PR's new file, and 100% reproducible in this run (5/6 tests, not one intermittent test).

**Required outcome:** Root-cause and fix the Windows failure in `workflow-wait.integration.spec.ts` (or the `wait` command / waiter machinery it exercises) so `test (windows-latest)` passes at the corrected head. I could not diagnose the exact mechanism from a macOS review environment — the log does not capture the failing `wait` process's stderr — but the pattern (uniform `exitCode 1` across every scenario that launches a real detached run, `EBUSY` during cleanup) suggests a Windows-specific timing or file-locking interaction between the detached launcher and the waiter's DB reads, not a one-off scheduling flake.

**Found by:** coordinator (direct CI log inspection, not delegated to the scope agents)

**Disposition:** OPEN — blocks merge until CI is green at the corrected head.

</details>

<details>
<summary><code>R2</code> — One of four <code>RunAttention</code> consumers isn't exhaustive, and it's the one guarding a live HTTP surface</summary>

**Impact:** No wrong behavior today — `'terminal'` is provably unreachable in `pausedGateBlocker` (the route already returns 400 for any non-`'paused'` status before calling it) and `'awaiting_response'` correctly falls through to "no blocker, proceed." But `assertApprovable`, `assertRejectable` (`workflow-operations.ts`), and `formatWaitOutcome` (`workflow.ts`) all switch over the same `RunAttention['kind']` exhaustively — no `default`, so TypeScript fails the build the moment a 5th kind is added anywhere they aren't updated. `pausedGateBlocker` (`server/src/routes/api.ts`) has a `default: return null;`, which absorbs a future kind silently instead of failing to compile. Since the helper's own doc comment states its purpose is exactly to give the console a 400 instead of an opaque 500 that `approveWorkflow`'s internal `assertApprovable` would throw, a silently-absorbed new kind reintroduces that opaque 500 on the approve/reject/respond HTTP routes — the precise regression class this PR exists to close, on the one consumer that isn't protected against it.

**Evidence:** `packages/server/src/routes/api.ts:3852-3874` (`pausedGateBlocker`) vs. the exhaustive siblings at `packages/core/src/operations/workflow-operations.ts:279-317` and `:342-374`, `packages/cli/src/commands/workflow.ts:3479-3492`, and `packages/core/src/orchestrator/prompt-builder.ts:114-128`. The repository's own established discipline for this exact situation is `reRunsOwnNodeOnResume` at `packages/workflows/src/schemas/workflow-run.ts:900-926`, whose doc comment states it is "Exhaustively switched... so a future fifth reason cannot silently repeat the gap this closes," backed by a `never`-typed default.

**Required outcome:** Replace `pausedGateBlocker`'s `default: return null;` with explicit `case 'awaiting_response':` and `case 'terminal':` returning `null` (both provably correct today), and drop the `default` so a future `RunAttention` kind fails to compile here exactly as it does at the other three consumers.

**Class:** Search performed — `grep -rn "attention" packages/ | grep -E "\.kind|switch|case '"` and `grep -rln "RunAttention" packages/` (excluding tests). Members examined: `workflow-operations.ts:280` (exhaustive, clean), `workflow-operations.ts:343` (exhaustive, clean), `prompt-builder.ts:114/116/128` (if-chain, TS-exhaustive, clean), `workflow.ts:3479` (exhaustive, clean), `api.ts:3852` (not exhaustive — this finding). `run-attention-watch.ts`'s `resolveAttention` only branches on `'blocked_on_child'`/`'terminal'` and passes every other kind through unmodified, which is correct for a chain-walker rather than a decision point, not a member of this class.

**Disposition:** OPEN. Cheap (a two-case addition), not currently live-broken, and non-blocking on its own — recommended to fix in the same PR since it touches the exact lines already under review, but does not by itself gate merge.

</details>

<details>
<summary><code>R3</code> — A real, tested behavior change on the chat prompt-builder surface isn't in the PR's behavior-change table</summary>

**Impact:** None on correctness — the new behavior is arguably more correct than the old one. But the PR description states "Every other user-visible string at the four migrated sites is byte-identical, pinned by test," and that claim isn't accurate for one reachable case: a conversation-scoped run paused on a `wait:` node (no `approval` key in metadata) previously fell through `formatPausedGateSection`'s `!isApprovalContext(approval)` branch and told the chat agent to relay "its approval context is missing or malformed... resolve it explicitly with `/workflow approve`" — misleading advice for a run that has nothing to approve. The new code recognizes this case via `runAttention` returning `null` and emits `''` (no paused-gate section at all), matching the new test `'says nothing for a `wait:` pause — the clock owns it, not a person'` in `prompt-builder.test.ts`.

**Evidence:** Old code at `origin/dev:packages/core/src/orchestrator/prompt-builder.ts:107-113` (`!isApprovalContext(approval)` unconditionally returns the "missing or malformed" text, with no wait-aware branch) vs. new code at `packages/core/src/orchestrator/prompt-builder.ts:114-115` (`attention === null` returns `''`). Reachability confirmed: `getPausedWorkflowRun` (`packages/core/src/db/workflows.ts:549-564`) queries any `status = 'paused'` row for the conversation, not specifically approval-gated ones, so a conversation-triggered workflow containing a `wait:` node reaches this path.

**Required outcome:** Add this case to the PR description's behavior-change table (or an equivalent note) for reviewer/changelog accuracy. No code change required — the implementation and its test are correct as written.

**Found by:** coordinator, independently confirmed by `prp-core:code-reviewer`

**Disposition:** OPEN, non-blocking — a documentation-accuracy note, not a defect.

</details>

<details>
<summary><code>R4</code> — <code>pendingDoorbell</code> coalescing flag guards an untested race and could be simplified</summary>

**Impact:** None — correctness is unaffected either way. The flag exists to catch a NOTIFY that fires in the narrow window between a row read and the next `nextWake()` call; dropping it would only ever cost up to one extra ≤1s poll tick in that exact race, which is already the SQLite dialect's permanent behavior. No test in the "notification doorbell" describe block (`run-attention-watch.test.ts:289-336`) exercises that specific window — both doorbell tests ring while `nextWake()` is already pending.

**Evidence:** `packages/core/src/services/run-attention-watch.ts:174-218`.

**Required outcome:** Optional. A smaller shape (drop `pendingDoorbell`, let the notify callback call `wake?.('notify')` directly) would remove ~10 lines at the cost of an already-tolerated latency window — worth doing only if convenient, not required for this PR.

**Disposition:** OPEN, non-blocking — genuinely optional polish.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | No additional findings beyond R3 (independently corroborated) |
| `seams` | R2 |
| `simplify` | R4 |

#### Validation

|---|---|---|
| `bun run validate` (review worktree, macOS, this head) | PASS | `check:bundled`/`check:bundled-skill`/`check:bundled-schema`/`check:pi-vendor-map`/`check:capability-matrix` all OK; type-check clean across all 13 packages; `lint --max-warnings 0` clean; `format:check` clean; full suite 165 test groups, 0 fail |
| `bun run check:schema-upgrades` | PASS | fresh install OK, all 9 shipped schema vintages converge idempotently |
| `git diff origin/dev...HEAD -- migrations/ packages/core/src/db/bundled-schema.generated.ts packages/core/src/db/adapters/sqlite.ts` | PASS (empty) | no schema/migration/SQLite-adapter changes |
| `grep -rn "awaiting_human\|decideOn"` (repo-wide) | PASS (empty) | rename to `awaiting_response`/`respondTo` is complete |
| `WORKFLOW_EVENT_NOTIFY_CHANNEL` string check (postgres.ts, pg-notify-listener.ts, run-attention-watch.ts) | PASS | all three sites use the constant; literal value `'archon_dashboard_event'` unchanged |
| CI `test (ubuntu-latest)` at reviewed head | PASS | |
| CI `test (windows-latest)` at reviewed head | **FAIL** | see R1 |
| CI `Test Suite` (aggregation gate) | **FAIL** | fails solely because it requires both OS legs to succeed |
| CI `schema-upgrade`, `postgres-parity`, `docker-build`, `Build docs site`, `CodeRabbit` | PASS | |

Test files registered: `run-attention-watch.test.ts` added to `packages/core/package.json`'s `test` script; `workflow-wait.integration.spec.ts` added to `packages/cli/package.json`'s `test` and `test:integration` scripts. Both confirmed present.

Unscored observation from the simplify pass: `workflow.test.ts` has one test (`workflowRunCommand > does not throw and logs warn when result message DB persist fails`) that fails only when run as part of the full file, not in isolation — appears to be a pre-existing test-order flake unrelated to `wait` (this PR's new tests are appended after it in the file). Not this PR's cause; not scored as a finding.

</details>

PR #2919, opened 2026-08-28
Prevent removal of live worktrees

Wirasm:
## Problem and outcome

`/worktree remove` could destroy a worktree still owned by a resumable workflow run, unlike isolation cleanup. That leaves a live run without the worktree it needs to resume or be abandoned.

- **Outcome:** `/worktree remove`, including `--force`, refuses to remove a worktree owned by a live workflow run.
- **Invariant:** A live workflow run keeps its isolation environment, worktree, and branch available until the run is resumed or abandoned.
- **Scope boundary:** This changes only the chat command's removal path; workflow lifecycle ownership and isolation cleanup behavior are unchanged.
- **Root cause:** The command-handler removal path did not perform the live-run ownership check already used by isolation cleanup.

## Review guidance

- **Feedback requested:** Confirm the command now applies the same live-run protection as isolation cleanup.
- **Start here:** `packages/core/src/handlers/command-handler.ts:595` — checks for a live owner before calling the isolation provider's destructive operation.
- **Review order:** Read the guard, then its regression test in `packages/core/src/handlers/command-handler.test.ts`.
- **Lower-attention areas:** The remaining test changes wire the existing database mock into the command-handler test fixture.
- **Known risk or uncertainty:** None.

## Solution

The command handler now reuses `getLiveRunOwningEnv` before it asks the isolation provider to destroy the worktree. When a resumable run owns the environment, it returns an actionable error identifying the run and status. The check happens before provider destruction, isolation status updates, or conversation rebinding.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `/worktree remove --force` could remove a worktree owned by a paused workflow run. | It refuses removal and tells the operator to resume or abandon the owning run. |
| **Failure behavior** | A live run could lose its required worktree and branch. | The worktree, isolation environment, and conversation binding remain unchanged. |

## Validation

- `bun test src/handlers/command-handler.test.ts` (from `packages/core`) — passed, 145 tests — proves a paused owner blocks destruction, DB status changes, and conversation rebinding.
- `bun run validate` — passed — covers generated-artifact checks, type checks, lint, format check, install test, and the full test suite.
- **Not verified:** Nothing material; the focused regression test exercises the changed command path and full validation passed.

## Links

- Closes #2893

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented `/worktree remove` from deleting a worktree currently owned by an active workflow run.
  * Added a clear message identifying the workflow and instructing users to resume or abandon it before removal.
[...]

PR #2928, opened 2026-08-28
fix(cli): archon serve can no longer hang extracting the web UI on windows

Wirasm:
## Problem and outcome

`archon serve` downloads the web UI tarball and extracts it by piping the bytes into `tar xzf -` as `stdin: <Uint8Array>`. On windows that hands Bun a channel the parent must write and close before the child can make progress, and that pump can stall with no upper bound — leaving `tar` blocked forever. The Windows CI lane only made it visible: two `downloadWebDist` tests died at the 5000ms per-test budget on PR #2919's lane, with `tar` reported as a killed dangling process. In production nothing ends the wait; `archon serve` would sit there with no output and no error.

- **Outcome:** extraction can no longer block on a channel the parent owns, in either direction.
- **Invariant:** the extractor is handed an inherited file descriptor, never bytes Bun has to pump; and its `stderr` is drained concurrently with the wait on exit.
- **Scope boundary:** `tar` is still the extractor and the command line is unchanged apart from staging. No timeout, retry, or in-process untar was added.
- **Root cause:** `Bun.spawn(..., { stdin: <Uint8Array> })` makes Bun responsible for feeding and closing the child's stdin. A local probe shows the shape directly — with a `Uint8Array` the child's fd 0 is neither a regular file nor a pipe (a parent-pumped socketpair), while with `Bun.file(path)` it is a regular file the kernel serves:

  ```
  Uint8Array stdin  ->  ls -l /dev/stdin: NOT_REGULAR
  Bun.file() stdin  ->  ls -l /dev/stdin: REGULAR
  ```

  On the failing run both spawns in the same bun process hung and were killed at the budget; the same two tests took 156ms and 16ms on the same runner image half an hour later, so the extraction itself is not slow — it never started.

## Review guidance

- **Feedback requested:** correctness of the staging and cleanup path, and whether the staged archive belongs beside `targetDir` rather than inside the extraction directory.
- **Start here:** `packages/cli/src/commands/serve.ts:160` — the staging, spawn, and concurrent stderr drain.
- **Review order:** `serve.ts` first, then the guard folded into the happy-path test in `serve.test.ts`.
- **Known risk or uncertainty:** the windows hang cannot be reproduced on macOS, so the causal chain rests on the CI evidence plus the local fd-shape probe rather than on a red-to-green run on windows. See **Not verified**.

## Solution

Write the verified tarball to `${targetDir}.tmp.tar.gz` and pass `Bun.file(tarballPath)` as stdin. `tar` then inherits a plain descriptor and reads at its own pace, with no parent-side step left to stall. A `finally` removes the staged archive on every path, including the checksum and extraction failure paths.

The archive stays off the command line deliberately: `tar` keeps reading `-`, so a windows drive letter never has to be disambiguated from `tar`'s own `host:path` remote syntax.

`stderr` is now drained with `Promise.all` alongside `proc.exited` instead of only after a non-zero exit. An unread pipe is the same deadlock in the output direction once `tar` fills the buffer, and the previous code could only read it after the process had already exited.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Tarball piped to `tar` through a Bun-managed stdin channel | Tarball staged to a temp file; `tar` inherits its descriptor and the file is removed afterwards |
| **Failure behavior** | On windows, `tar` could block on an unfed stdin indefinitely — no error, no timeout | The child has nothing parent-owned to block on; a `tar` failure still reports its exit code and stderr |

## Validation

- `bun test src/commands/serve.test.ts` in `packages/cli` — 20 pass, 0 fail, 33 expect() calls — proves extraction, checksum handling, the descriptor invariant, and staged-archive cleanup.
- Guard seen red: reverting the single line to `stdin: new Uint8Array(tarballBuffer)` fails the new assertion with `Received value: Uint8Array(131) [ 31, 139, 8, ... ]`. It is a real guard, not a passing decoration.
- Real artifact, not just the 131-byte fixture: the shipped v0.9.0 `archon-web.tar.gz` (2.1 MB compressed, 8.5 MB extracted) was fed through `downloadWebDist` and extracted correctly — `["index.html","favicon.png","assets"]` with all three asset files under `assets/`, after `--strip-components=1`.
- Cost of staging, same real artifact, three runs each on macOS: **before 20.6 / 20.0 / 18.7 ms, after 21.1 / 22.3 / 23.0 ms**. The change buys hang-immunity for roughly 2 ms of disk write; it is not a speedup and is not claimed as one.
- `bun run test` in `packages/cli` — all 12 file groups green.
- `bun run lint` and `bun run validate` at the repo root — validate exits 0, 163 test groups reporting `0 fail`.
- **Not verified:** that the windows hang is gone, because it is intermittent and cannot be forced. A green windows lane on this PR is not proof. The claim this PR does support is narrower and checkable: after this change no parent-owned channel remains for the child to block on, in either direction.

## Links

- Related #2924
- Related #2306

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when downloading and extracting the web interface.
  * Added a timeout for extraction processes to prevent hangs.
  * Improved error reporting when extraction fails.
  * Ensured temporary files and directories are cleaned up after successful or failed extraction.

### DIFF
--- a/packages/cli/src/commands/serve.test.ts
+++ b/packages/cli/src/commands/serve.test.ts
@@ -210,11 +210,22 @@ describe('downloadWebDist', () => {
     consoleLogSpy.mockRestore();
   });
 
+  // The spawn-shape and cleanup assertions ride along here rather than in tests
+  // of their own: each call to downloadWebDist costs a real `tar` spawn, and
+  // spawn count on windows is the whole reason this file gets attention.
   it('verifies against the embedded hash without fetching checksums.txt', async () => {
     fetchSpy.mockImplementation(async () => new Response(tarballBytes));
     const targetDir = join(tmpRoot, 'target-embedded-ok');
+    const spawnSpy = spyOn(Bun, 'spawn');
+    let extractorStdin: unknown;
 
-    await downloadWebDist('9.9.9', targetDir, tarballHash);
+    try {
+      await downloadWebDist('9.9.9', targetDir, tarballHash);
+      // Read before restoring — mockRestore() clears the recorded calls.
+      extractorStdin = (spawnSpy.mock.calls[0]?.[1] as { stdin?: unknown } | undefined)?.stdin;
+    } finally {
+      spawnSpy.mockRestore();
+    }
 
     // Content, not just existence — a truncated or corrupt fixture still yields
     // an index.html and a `tar` exit 0, so only this assertion catches it.
@@ -222,6 +233,13 @@ describe('downloadWebDist', () => {
     // Only the tarball is fetched — checksums.txt must NOT be requested.
     expect(fetchSpy).toHaveBeenCalledTimes(1);
     expect(String(fetchSpy.mock.calls[0]?.[0])).toContain('archon-web.tar.gz');
+    // `tar` must inherit a file descriptor. Passing the bytes as `stdin` instead
+    // leaves the parent owning a channel it has to pump and close, and a stalled
+    // pump blocks `tar` forever — the windows hang in #2924. A BunFile is a Blob;
+    // a Uint8Array is not, which is exactly the regression this catches.
+    expect(extractorStdin).toBeInstanceOf(Blob);
+    // The staged archive is ~2 MB in production — it must not survive extraction.
+    expect(existsSync(`${targetDir}.tmp.tar.gz`)).toBe(false);
   });
 
   it('hard-fails on embedded hash mismatch with a clear error', async () => {
--- a/packages/cli/src/commands/serve.test.ts
+++ b/packages/cli/src/commands/serve.test.ts
@@ -255,6 +255,26 @@ describe('downloadWebDist', () => {
     expect(fetchSpy).toHaveBeenCalledTimes(1);
   });
 
+  // The mismatch test above cannot reach this path: verification runs before
+  // anything is staged, so it never leaves temp state behind. Checksum-valid
+  // bytes that are not a gzip stream are the only cheap way in — staging happens,
+  // then `tar` exits non-zero.
+  it('leaves nothing behind when tar exits non-zero', async () => {
+    const notAnArchive = new TextEncoder().encode('checksum-valid, but not gzip');
+    const hasher = new Bun.CryptoHasher('sha256');
+    hasher.update(notAnArchive);
+    fetchSpy.mockImplementation(async () => new Response(notAnArchive));
+    const targetDir = join(tmpRoot, 'target-tar-failure');
+
+    await expect(downloadWebDist('9.9.9', targetDir, hasher.digest('hex'))).rejects.toThrow(
+      'tar extraction failed'
+    );
+
+    expect(existsSync(`${targetDir}.tmp.tar.gz`)).toBe(false);
+    expect(existsSync(`${targetDir}.tmp`)).toBe(false);
+    expect(existsSync(targetDir)).toBe(false);
+  });
+
   it('falls back to remote checksums.txt when the embedded hash is empty', async () => {
     fetchSpy.mockImplementation(async (url: string | URL | Request) => {
       if (String(url).includes('checksums.txt')) {
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -12,6 +12,16 @@ const log = createLogger('cli.serve');
 
 const GITHUB_REPO = 'coleam00/Archon';
 
+/**
+ * Upper bound on the `tar` child. Extracting the ~2 MB release archive takes
+ * tens of milliseconds, so this leaves three orders of magnitude of headroom for
+ * a slow disk. Its only job is to stop a stalled child from turning
+ * `archon serve` into a silent permanent hang: the parent-owned stdin channel
+ * that caused the observed stall is gone (#2924), but filesystem-side stalls on
+ * windows were never ruled out, and there is no budget in production to end one.
+ */
+const EXTRACTION_TIMEOUT_MS = 60_000;
+
 function toError(err: unknown): Error {
   return err instanceof Error ? err : new Error(String(err));
 }
@@ -173,10 +183,12 @@ export async function downloadWebDist(
   // (#2924). Reading `-` keeps the archive path off the command line, where a
   // windows drive letter is ambiguous with `tar`'s own `host:path` syntax.
   await Bun.write(tarballPath, tarballBuffer);
+  const extractionStartedAt = performance.now();
   try {
     const proc = Bun.spawn(['tar', 'xzf', '-', '-C', tmpDir, '--strip-components=1'], {
       stdin: Bun.file(tarballPath),
       stderr: 'pipe',
+      timeout: EXTRACTION_TIMEOUT_MS,
     });
     // Drain stderr while waiting rather than after: a pipe nobody reads is the
     // same deadlock in the other direction once `tar` fills its buffer.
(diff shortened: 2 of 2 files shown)


Wirasm:
## Adversarial review — reviewed head `145b498c`

I re-derived every claim in the report rather than trusting it: cloned the PR head into a scratch checkout, reverted the fix line and reran the suite, wrote two standalone probes (tar-failure cleanup, concurrent-download race), and independently re-enumerated every `Bun.spawn`/`spawnSync` call in the repo. Findings below.

### Verdict: READY

No correctness bug in the merged diff. One Important judgment call worth a real answer before or right after merge; two Minor items.

---

### Important — no bound on the residual hang path (`serve.ts:168-192`)

The fix correctly removes the *confirmed* mechanism (parent-pumped stdin socketpair stalling on Windows). But the report's own "Follow-up if it recurs" section names an **unruled-out suspect for the same symptom** — filesystem-side scanning (Defender) stalling `tar`'s read of the staged file — and its "Not verified" line says outright: *"that the windows hang is gone... a green windows lane on the PR is not proof."* So the one thing this PR is titled to fix (`archon serve` hanging forever, silently, with zero diagnostic) is not actually confirmed fixed — only the one mechanism with hard evidence is confirmed removed.

`AGENTS.md` lists subprocess timeouts explicitly as sanctioned timer use ("Timers remain appropriate for recoverable operations such as retry backoff, **subprocess timeouts**, and cleanup of already-terminal data"). And this is now close to free: `Bun.spawn` (1.3.11, the pinned version) has a native `timeout`/`killSignal` option — no manual `AbortController`/`Promise.race` needed. Given the report's own timing table shows a 2 MB archive extracting in ~20ms locally, even a generous 60-120s bound leaves three orders of magnitude of headroom for a legitimately slow disk, while converting "hangs forever with no error" into "fails after N seconds with an actionable message" — which is the actual production complaint in #2924.

Steelmanning the omission: the Defender suspect is unreproduced, and adding a timer for an unconfirmed failure mode is the kind of speculative defense-in-depth this repo's own conventions warn against. That's a legitimate read. But given (a) the explicit AGENTS.md carve-out for exactly this timer class, (b) the report's own admission that the headline symptom is unverified, and (c) the near-zero implementation cost, I don't think "no parent-owned channel remains" fully carries the argument for leaving it out. Recommend either adding the bound now, or opening a tracked follow-up issue for the named Defender/filesystem residual so it isn't silently treated as closed.

### Minor — tar-failure cleanup path is untested (`serve.ts:190-192`, `serve.test.ts`)

I verified the behavior is correct — wrote a standalone probe that feeds `downloadWebDist` a checksum-valid but non-gzip payload (tar exits 1) and confirmed the staged `${targetDir}.tmp.tar.gz` and `${targetDir}.tmp` are both removed via the `finally`. But none of the three committed tests exercise a non-zero `tar` exit; the only failure-path test (`hard-fails on embedded hash mismatch`) throws before staging ever happens, since checksum verification runs before `Bun.write`. The `finally` cleanup is exactly the load-bearing property the PR title promises ("can no longer hang" implies "and cleans up its temp state"), and it's currently only guaranteed by manual verification, not a regression test. A cheap addition: mock/force a non-zero tar exit (e.g., feed corrupt-but-checksum-matching bytes, matching what I did) and assert `existsSync('${targetDir}.tmp.tar.gz')` is false afterward.

### Minor / informational — concurrent double-download race on shared paths (`serve.ts:161-166`)

I reproduced this directly: two concurrent `downloadWebDist` calls to the same `targetDir` race on the shared `tmpDir` (and now also the shared `tarballPath`). Outcome: one call succeeds, the other throws a somewhat confusing `Extraction produced unexpected layout — index.html not found in extracted dir` rather than corrupting output or hanging — so it's a clean, self-recoverable failure, not silent data loss. Important context: **this is not new** — the pre-fix code already shared `tmpDir` unprotected across concurrent invocations; this PR adds one more shared path on top of an already-unguarded one, not a new hazard class. Not blocking. Since this code is already being touched, a `mkdtempSync`-style unique staging suffix would close it outright if it's worth the diff, but I wouldn't hold the PR for it.

---

### Verified clean

- **Staging lifecycle / atomicity**: `await Bun.write(tarballPath, tarballBuffer)` fully completes (file closed) before `Bun.spawn` starts reading it in the same process — `tar` can never observe a half-written file from its own call.
- **Cleanup on every path**: confirmed via probe — success, non-zero tar exit, and (trivially, since `Bun.write` precedes the `try`) the finally always fires once staging has occurred.
- **stderr drain change**: `Promise.all([proc.exited, new Response(proc.stderr).text()])` preserves the exact same error-message content as the old sequential read; it only changes *when* stderr is drained (concurrently vs. after exit), closing a theoretical opposite-direction deadlock if `tar`'s stderr ever filled the OS pipe buffer. No behavior change on the exit-143/`cleanupAndThrow` path — same message shape, same trim.
- **Class enumeration**: re-ran independently (`grep` across `packages/` and `scripts/`, excluding stale `.worktrees/` copies of this same file). `serve.ts:168`'s pre-fix `stdin: new Uint8Array(tarballBuffer)` was the only buffer-as-stdin spawn in the repo. `container-spawn.ts:172`'s `stdin: child.stdin` is a stream passthrough (docker exec), a different class entirely — no fix needed there.
- **Mutation-testing claim**: verified for real, not just trusted. Reverted `stdin: Bun.file(tarballPath)` back to `stdin: new Uint8Array(tarballBuffer)` in a scratch checkout and reran — the new `expect(extractorStdin).toBeInstanceOf(Blob)` assertion fails exactly as the report claims (19 pass / 1 fail).
- **Windows correctness**: the staged path never appears as a command-line argument to `tar` (only via the `stdin` option), so the drive-letter/`host:path` ambiguity the comment warns about doesn't apply to it. `-C tmpDir` was unchanged by this PR. `test (windows-latest)` passed on this head in 9m21s — expected, and consistent with the report's own caveat that a green Windows lane doesn't prove the intermittent hang is gone (it wasn't forced to reproduce here either).

Full baseline rerun on this head in isolation: 20 pass / 0 fail, 33 `expect()` calls — matches the report.

Wirasm:
## Round-2 addendum — reviewed head `e507ac6d`

Delta-verified against the round-1 review above. Same method: cloned this head into an isolated scratch checkout and reproduced each claim directly rather than reading it off the diff.

**Verdict: READY**

### 1. Timeout implementation (`serve.ts:15-24`, `186-209`) — verified

The discriminator claim is exactly right, confirmed with a standalone `Bun.spawn` probe using the same option shape (`timeout`, `Bun.file` stdin) as the production call:

```
NORMAL EXIT  -> exitCode: 0   killed: true   signalCode: null
NONZERO EXIT -> exitCode: 1   killed: true   signalCode: null
TIMEOUT KILL -> exitCode: 143 killed: true   signalCode: SIGTERM
EXTERNAL KILL-> exitCode: 143 killed: true   signalCode: SIGTERM
```

`proc.killed` reads `true` after *any* exit, including a clean success — branching on it would have been wrong. `proc.signalCode` correctly stays `null` for every normal exit and is set only when a signal actually terminated the process, whether from the `timeout` option or an external kill. So `proc.signalCode !== null` is the correct discriminator, and reporting elapsed time rather than asserting "the 60s bound fired" is the right call, since an external signal would hit this same branch at a different elapsed time. `details` (trimmed stderr) is threaded into both the signal-kill message and the exit-code message, so stderr evidence is preserved exactly as in round 1.

### 2. New cleanup test — seen-red claim confirmed

Ran `leaves nothing behind when tar exits non-zero` on this head: passes cleanly (21 pass / 0 fail / 37 expect(), matching the report). Removed the `finally` block's `rmSync(tarballPath, ...)` and reran — both this test and the round-1 happy-path staged-archive assertion go red (`Expected: false / Received: true`), confirming the mutation actually exercises the cleanup path rather than a no-op.

### 3. Declined-race record — confirmed present

The report's "Known pre-existing edge — concurrent double-download race" section accurately records the round-1 finding: reproduced, self-recoverable, predates this PR, deliberately not fixed here to avoid scope creep. No code check needed since nothing changed on this path.

### 4. CI rollup on `e507ac6d` — all green

`test (windows-latest)` passed in 7m42s; `test (ubuntu-latest)` 4m49s; `docker-build`, `postgres-parity`, `schema-upgrade`, `changes`, `Test Suite`, CodeRabbit all green. As before, a green Windows lane doesn't prove the intermittent hang is gone (it's not forceable), but it's a clean rollup with no new red.

Nothing outstanding from round 1 remains open as a code concern.

PR #2929, opened 2026-08-28
fix(cli): a folder project registered by the CLI is found again by every other command

### LINKED ISSUE #2927: Windows: a folder project registered by `workflow run --folder` is invisible to every other command in that directory
## Problem

On Windows, a folder project registered by `archon workflow run --folder` cannot be found again by any other command in that same directory. `archon workflow get`, `wait`, `cancel`, `runs`, `resume`, `approve` and `reject` all refuse with:

```
Error: Not in a git repository.
Either navigate to a git repo or use --cwd to specify one.
Or register this folder as a project: run with --folder, or use /register-project in chat.
```

The message is wrong about the cause — the project *is* registered. `workflow run --folder` keeps working because it is the one command with an escape hatch past the lookup.

## Root cause

The write side and the read side of `remote_agent_codebases.default_cwd` canonicalize with two different functions, and on Windows the two disagree.

| Site | Function | Role |
|---|---|---|
| `packages/core/src/handlers/clone.ts:517` (`registerFolder`) | `realpath` from `fs/promises` | CLI/HTTP writer |
| `packages/cli/src/cli.ts:533` (pre-dispatch project gate) | `realpathSync` from `fs` | reader |
| `packages/core/src/orchestrator/orchestrator-agent.ts:3340` (`/register-project`) | `realpathSync` from `fs` | chat writer |

`fs/promises.realpath` expands an 8.3 short path component; `fs.realpathSync` does not. Observed directly in CI logs on `windows-latest`, from an unrelated test's own logging:

```
"path":      "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\archon-source-rL7MHL\\..."
"canonical": "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\archon-source-rL7MHL\\..."
```

So `registerFolder` stores the `runneradmin` form while the gate looks up the `RUNNER~1` form. `findCodebaseByDefaultCwd` (`packages/core/src/db/codebases.ts:108`) is an exact SQL string match, and the `findCodebaseByPathPrefix` fallback (`:128`) cannot rescue it either — the differing component is an ancestor, so the separator-anchored prefix check fails too. (That helper is *not* at fault: it correctly imports `sep` from `path`.)

Windows assigns 8.3 short names whenever a name exceeds 8 characters or contains a space or a dot, and `%TEMP%` is commonly set to the short form — which is exactly why GitHub's runners show `RUNNER~1`.

**This was already known and never fixed.** `packages/core/src/orchestrator/orchestrator.test.ts:1602-1603` says: *"`fs/promises.realpath` differs on Windows 8.3 short names (RUNNER~1 vs runneradmin), so mixing the two flakes there."* The knowledge stayed in a test comment instead of becoming a shared primitive.

## Second instance of the same defect

The two *writers* also disagree with each other. `/register-project` in chat uses `realpathSync`; the CLI/HTTP path uses `fs/promises.realpath`. On Windows the same directory registered through both surfaces produces two `default_cwd` strings, hence a duplicate codebase row and a project each surface hides from the other. Both call sites carry comments claiming to mirror the other.

## How it surfaced

`packages/cli/src/commands/workflow-wait.integration.spec.ts` (PR #2923) is the first test that runs a **second** `archon` command in a non-git folder-project directory on Windows CI. 5 of its 6 tests fail there; the only one that passes is the one whose fixture is a real git repo, so the gate never consults the database. `workflow-terminal-event.integration.spec.ts` uses `--folder` too but only reads SQLite afterwards, so it passes over the same broken state.

Windows job: https://github.com/coleam00/Archon/actions/runs/33173070434/job/98854893857

## Fix boundary

One canonicalizer, shared by every writer and reader of `default_cwd`. `packages/paths/src/archon-paths.ts` already owns `expandTilde` and is imported by core, cli and workflows, so it is the natural home. A guard added only at the failing gate would leave the two writers free to keep disagreeing.

**Decision needed before implementing:** which function wins. Rows written on Windows by `registerFolder` carry the expanded form; chat-registered rows carry the unexpanded one. Standardizing on the expanding variant keeps existing CLI-registered folder projects resolvable and breaks chat-registered ones; standardizing on `realpathSync` does the reverse. Consider matching both candidate forms on read during a transition.

Preserve: symlink resolution (the reason realpath is on this path at all — macOS `/tmp` → `/private/tmp`, and the #2127 worktree-subdirectory prefix match), the separator-anchored prefix fallback, and the existing fail-safe of falling back to the raw cwd when canonicalization throws.

## Regression proof

`bun --cwd packages/cli test src/commands/workflow-wait.integration.spec.ts` on `windows-latest` is red today and green after. Add a cross-platform seam lock so the two sides cannot silently re-diverge: assert that `registerFolder` stores exactly the shared canonicalizer's output and that the gate looks that same value up, both importing the one function.

## Not verified

`fs.realpathSync.native`'s behaviour on Windows 8.3 names is untested — on macOS all three variants agree, so macOS cannot distinguish them. If the fix picks `.native`, verify on Windows CI first. Drive-letter case and `\\?\` prefixes are a second potential divergence in the same exact-string match; not needed to explain this failure, but a shared canonicalizer removes them too.

Wirasm:
## Problem and outcome

On Windows, a folder project registered by `archon workflow run --folder` could not be found again by any other command in that same directory. `workflow get`, `wait`, `cancel`, `runs`, `resume`, `approve` and `reject` all refused with `Error: Not in a git repository`, which was never true — the project was registered, just stored under a different string than the one the lookup asked for. `workflow run --folder` kept working only because it is the one command with an escape hatch past the lookup.

- **Outcome:** every command resolves a registered folder project in that directory, on every platform. Registering one directory from chat and from the CLI produces one row instead of two.
- **Invariant:** every writer and every reader of `remote_agent_codebases.default_cwd` turns a filesystem path into that string with the same function. The column is matched by exact string equality, so a second canonicalizer anywhere on the path silently breaks the lookup.
- **Scope boundary:** repo projects are untouched — git canonicalizes the repo root on both sides, so they were never affected. The forge adapters write an Archon-constructed `getProjectSourcePath()`, not a user path, and are left alone. `resolveRunCodebase` in `workflow.ts` is also left alone: it receives an already-resolved `effectiveCwd` from the gate, and canonicalizing a git `repoRoot` there would change the lookup key for repo projects.
- **Root cause:** `registerFolder` (`packages/core/src/handlers/clone.ts`) canonicalized with `realpath` from `fs/promises`, while the CLI pre-dispatch gate (`packages/cli/src/cli.ts`) and the chat writer `/register-project` used `realpathSync` from `fs`. On Windows `fs/promises.realpath` expands an 8.3 short component and `fs.realpathSync` does not, observed directly in CI on `windows-latest`: `C:\Users\RUNNER~1\...` from one and `C:\Users\runneradmin\...` from the other, in the same job. `findCodebaseByDefaultCwd` is an exact SQL match, and the `findCodebaseByPathPrefix` fallback could not rescue it either because the differing component is an ancestor, so the separator-anchored prefix check fails too. Neither lookup helper is at fault.

This was already known and never fixed: `orchestrator.test.ts` carried a comment saying the two functions differ on Windows 8.3 short names "so mixing the two flakes there". The knowledge stayed in a test comment instead of becoming a shared primitive.

## Review guidance

- **Feedback requested:** the decision recorded under *Which form wins* below, and whether the seam locks are the right shape.
- **Start here:** `packages/paths/src/archon-paths.ts:51` — `canonicalizeProjectPath`, the one function every site now resolves through.
- **Review order:** the canonicalizer, then the two writers (`clone.ts:506`, `orchestrator-agent.ts:3337` and `:3427`), then the two readers (`cli.ts:525`, `doctor.ts:512`), then the seam locks.
- **Lower-attention areas:** the two `package.json` edits are one-line additions of the new test files to the existing invocation lists.
- **Known risk or uncertainty:** Windows rows written by chat `/register-project` from a path that already contained an 8.3 short component keep the unexpanded string and will no longer resolve. See *Which form wins*.

## Solution

`canonicalizeProjectPath` in `@archon/paths` is now the single canonicalizer for `default_cwd`, and every writer and reader of that column calls it:

| Site | Before | After |
|---|---|---|
| `registerFolder` — CLI `--folder` and HTTP writer | `fs/promises.realpath` | shared |
| `/register-project` — chat writer | `fs.realpathSync` | shared |
| `/update-project` — chat writer | none; stored the raw argument | shared |

Both chat handlers also now canonicalize *before* checking the path exists, so they validate the path they are about to store. Chat input gets no shell expansion, so this is also what makes `/register-project demo ~/work` work — `registerFolder`, the other writer of the same column, has always expanded a tilde.
| CLI pre-dispatch project gate — reader | `fs.realpathSync` | shared |
| `doctor` folder-project check — reader | none; looked up the raw cwd | shared |

Two of those sites turned out to have no canonicalization at all, so they could write or look up a value nothing else could match — the same defect, just not the instance that was reported.

It preserves what was already load-bearing: symlink resolution (macOS `/tmp` → `/private/tmp`), the separator-anchored prefix fallback for the #2127 worktree-subdirectory case, and the fail-safe of returning the raw path when canonicalization throws. In `registerFolder` the existing `stat` is now the sole existence gate — canonicalization is fail-safe, so a missing or unreadable path fails there with its real errno instead of being registered.

**Which form wins.** The expanding variant (`fs/promises.realpath`), for two reasons. First, evidence: it is the only variant whose Windows short-name behaviour has actually been observed in this repo's CI. `fs.realpathSync.native` is believed equivalent but has never been verified on Windows, and macOS cannot distinguish the three, so nothing here depends on it. Second, and decisively, expansion is what makes this a canonicalization at all — the long name is the one name a directory has. Standardizing on the non-expanding variant would have fixed the writer/reader split while leaving two spellings of one directory able to register two rows.

**No transition dual-read.** Matching both candidate forms on read was considered and rejected. The stranded population is narrow: it requires Windows, chat `/register-project` specifically, and a supplied path that already contained an 8.3 short component as typed — machine-derived `%TEMP%` paths, not the paths people keep projects in. Against that, a dual-read means keeping a second realpath imported in the reader, which is exactly the defect being fixed and would defeat the seam lock; and no test on any CI platform can distinguish the two branches, so the fallback would ship unexercised. Rows written by every other surface — the CLI, HTTP, and the forge adapters — already carry the expanded form and keep resolving.

### Changed seams

|---|---|---|
| `default_cwd` writers → readers | one canonicalizer replaces three (and two sites that had none) | `packages/paths/src/archon-paths.ts:51`, `packages/cli/src/default-cwd-canonicalization.test.ts` |
| `registerFolder` → folder-project lookup | round trip proven against a real database through a real directory link | `packages/core/src/handlers/register-folder-lookup.integration.test.ts` |

## Validation

- `bun run validate` — passes, exit 0, 164 suite invocations with zero failures.
- `cd packages/paths && bun test src/archon-paths.test.ts` — 136 pass — proves the canonicalizer collapses two spellings of one directory to one string, is idempotent, expands `~`, and returns the absolute path rather than throwing when it cannot resolve.
- `cd packages/core && bun test src/handlers/register-folder-lookup.integration.test.ts` — 3 pass — write-then-read against a real `bun:sqlite` database and a real directory link: a project registered through the link is found from either path, re-registering through the other path returns the same row instead of a duplicate, and a subdirectory still resolves through the prefix fallback. On `windows-latest` the runner's `%TEMP%` is itself an 8.3 short path, so these assertions run directly over the condition that produced the bug.
- **Seam locks seen red against the pre-fix code.** With the four product files reverted to `dev` and the tests kept, each new lock fails, then passes with the fix:
  - `packages/cli/src/default-cwd-canonicalization.test.ts` — 7 of 8 fail. A site that resolves a `default_cwd` path may reach `canonicalizeProjectPath` and nothing else; adding a second canonicalization means importing a realpath, and importing a realpath fails here. This is the lock that works cross-platform, because a behavioural test cannot see the second canonicalizer on macOS or Linux where all the variants agree.
  - `orchestrator.test.ts` — `/register-project` and `/update-project` "stores exactly the shared canonicalizer output" both fail. They make the canonicalizer answer a different real directory than the one passed in, which no second realpath call could produce.
  - `doctor.test.ts` — "looks up the shared canonicalizer output, not the raw cwd" fails.
- **Windows CI green:** [run 33182776695](https://github.com/coleam00/Archon/actions/runs/33184032366) — `test (windows-latest)` passes, zero failures. The round-trip suite runs there under the runner's `%TEMP%`, which is itself the `C:\Users\RUNNER~1\...` short path that produced the bug.
- **Two Windows-only test defects surfaced and were fixed on the way.** Both were the same split, one layer up in the tests: `orchestrator-agent.test.ts` compared POSIX path literals against values the writers now make absolute (`resolve('/path')` is `D:\path` on Windows), and `workflow-terminal-event.integration.spec.ts` built its expected `working_path` with `realpathSync` while the run records the path the CLI canonicalized — so on Windows it kept the short form the recorded path expands, and the wait timed out. Both now derive their expectation from `canonicalizeProjectPath`. No product behaviour changed for either.
- Review round: CodeRabbit's one inline finding (chat handlers validated the raw path and stored the canonical one, rejecting `~/work`) was verified and fixed, with a test that fails against the previous ordering.
- **Not verified:** `fs.realpathSync.native` on Windows 8.3 names is still untested, and this fix deliberately does not depend on it. The end-to-end regression proof for the reported symptom is `workflow-wait.integration.spec.ts` on PR #2923, owned by another session and untouched here — its Windows leg going green once both land is the downstream confirmation. Drive-letter case and `\\?\` prefixes were a second potential divergence in the same exact-string match; a shared canonicalizer removes them, but no test targets them specifically.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No schema change; the stored value's form is unchanged for every surface except Windows chat `/register-project`, which was writing the non-canonical alias. A stranded row is recoverable with `/update-project <name> <path>`, which now canonicalizes. | *Which form wins* above |
| Observability | Canonicalization failures log `paths.canonicalize_project_path_failed` at debug; every caller still fails loudly on its own (`registerFolder` throws from `stat`, a reader misses the lookup and reports an unregistered directory). | `packages/paths/src/archon-paths.ts:78` |

## Links

- Closes #2927
- Related #2923


Wirasm:
## Adversarial review — reviewed head `b9a1d64f`

I re-cloned the branch fresh (`git fetch origin fix/2927-shared-cwd-canonicalizer`, confirmed HEAD == `b9a1d64f3f33837d875bd35644c9f082d5cd202d`), re-derived every claim in the report from source rather than trusting it, pulled the raw `windows-latest` job log instead of the green checkmark, and reverted the four product files to `dev` to independently reproduce the seam lock going red. Verdict below the findings.

### Findings (all low severity — none block merge)

**1. Site enumeration is accurate for writers; one reader is a real but harmless omission.**
Re-grepped every `default_cwd` write/read and every `findCodebaseByDefaultCwd`/`findCodebaseByPathPrefix` caller. The five canonicalizing sites are correct and complete for writers. There is a sixth *reader* not mentioned anywhere in the report: `packages/core/src/services/codebase-checkout-resolver.ts:37,43` (`findCodebaseForCheckoutPath`), called from `clone.ts:418` (`registerRepository`) and `workflow.ts:3790/3842/3943` (`workflowRunsCommand` scoping). It's safe — it explicitly skips `kind === 'folder'` codebases (line 51), `registerRepository` only ever handles git repos, and its CLI callers always receive the pre-dispatch gate's already-canonicalized `effectiveCwd` — so it's inert with respect to the #2927 defect class, consistent with the repo/folder carve-out. But it's undocumented in the "not changed, deliberately" list, and the next person who greps for `default_cwd` readers to extend this lock will miss it. Worth a one-line mention.

**2. No-dual-read decision: sound, and the circularity concern is disproven.**
Traced `handleUpdateProject` (`orchestrator-agent.ts:3415-3460`) — it resolves the target codebase by **name** (`existing.find(c => c.name.toLowerCase() === projectName.toLowerCase())`), never by path. Recovery via `/update-project <name> <path>` does **not** depend on the broken path-based lookup at all, so it is not circular — verified, not merely asserted.
I also checked the "machine-driven, not just human-typed" angle: `/register-project` is in fact parsed out of the AI's own generated response (`handleProjectRegistrationResult`, `orchestrator-agent.ts:3277`), so machine-derived paths do reach this writer. It doesn't weaken the argument — `canonicalizeProjectPath` normalizes any input form identically post-fix, so the stranded population is strictly pre-existing rows written by the *old* chat writer before this fix ships, a one-time migration concern rather than an ongoing risk. One gap worth naming: there is no CLI equivalent of `/update-project`, so a Windows CLI-only user (no chat interface) has no path to self-recover a stranded row without going through chat.

**3. Seam lock: catches realistic regressions; two narrow observations.**
Reverted the four product files to `dev`, kept the PR's tests, and ran `default-cwd-canonicalization.test.ts` myself: **7 of 8 fail**, matching the report exactly. Confirmed the regex `\brealpath` also catches an aliased import (`import { realpath as rp } from 'fs/promises'` still contains the literal substring in the import clause) — a real bypass requires deliberate string-concatenation obfuscation, not a realistic accidental regression, so the lock is fit for purpose.
The one forward-looking risk: the lock is scoped to four **entire files**, two of which (`cli.ts`, `orchestrator-agent.ts`) are large and handle every CLI command / every chat command respectively. I confirmed via `git show dev:<file> | grep realpath` that neither currently has any unrelated `realpath` usage, so the lock isn't spuriously tripping today — but a future legitimate, unrelated need for `realpath` anywhere else in either file will fail this test and force someone to route around it.

**4. registerFolder parity: holds for the realistic cases; one unverified edge case not in the report's "Not verified" list.**
Compared old vs. new for nonexistent path, file-not-directory, and permission-denied/dangling-symlink: parity holds in each case I traced, because `fs.promises.stat` also follows symlinks and does full path resolution, so it fails for essentially the same reasons `realpath` would have.
One case the report doesn't mention: `canonicalizeProjectPath`'s fail-safe catch is unconditional — it swallows *any* `realpath` error, not just ENOENT/EACCES. On Windows the whole point of this canonicalizer is to *expand* a short 8.3 name into a longer one. If that expansion trips a length-related failure, the fallback silently returns the short, un-expanded absolute path; `stat` on that shorter path can still succeed, so `registerFolder` would register the directory under the very form this PR exists to stop producing. Unverified (no Windows box to reproduce, and Node's realpath may already handle long paths via `\\?\`), so I'd file it next to the report's own "Not verified" section rather than treat it as a defect.

### Verified clean

- **`resolveRunCodebase` exclusion (point 5):** traced the full call chain — `cli.ts`'s pre-dispatch gate resolves `effectiveCwd` exactly once, and every downstream command dispatch (`workflowRunCommand`, `workflowStatusCommand`, `workflowGetCommand`, `workflowRunsCommand`, `workflowResumeCommand`, `workflowCancelCommand`, `workflowApproveCommand`, `workflowRejectCommand`, `workflowRespondCommand`) passes that same `effectiveCwd`. `resolveRunCodebase` and `findCodebaseForCheckoutPath` both receive it, so their own exact-match lookups succeed trivially without needing their own canonicalization. Coherent.
- **Terminal-event spec (point 6, commit `77ee98c5`):** traced `working_path: cwd` (`executor.ts:2179`) back through the same call chain to `effectiveCwd`; the test's new expectation matches production exactly. Confirmed directly from the raw `windows-latest` job log (job `98892344646`) that `detached workflow terminal database events` passed.
- **CodeRabbit ordering fix (point 7, commit `b9a1d64f`):** both chat handlers now canonicalize before validating existence and report the canonical path on error. The new test `/register-project validates the canonical path, not the raw argument` is non-vacuous — under the pre-fix ordering, `existsSync` on the literal `~/some-project` argument (mocked to return true only for the resolved path) returns false and the handler returns early, so `createCodebase` is never called and the assertion fails.
- **CI, independently pulled from source, not trusted from the green check:** fetched job `98892344646` (`test (windows-latest)`) logs directly — all 8 seam-lock assertions and all 3 round-trip assertions pass there, plus the terminal-event spec.
- **Local reproduction:** ran `clone.test.ts` (78 pass), `register-folder-lookup.integration.test.ts` (3 pass), `orchestrator.test.ts` (75 pass), `orchestrator-agent.test.ts` (277 pass), and `default-cwd-canonicalization.test.ts` + `doctor.test.ts` (84 pass) each in their own isolated `bun test` invocation, matching this repo's own mock-isolation convention (bundling them together produces 6 spurious cross-file mock-pollution failures — a testing-methodology trap, not a bug in the PR). `tsc --noEmit` clean on `packages/paths`, `packages/cli`, `packages/core`.
- **Forge adapter exclusion:** confirmed `canonicalPath` in all three forge adapters (`github`, `gitlab`, `gitea`) comes from `getProjectSourcePath(owner, repo)`, an Archon-constructed path under `~/.archon`, not user input — the exclusion is accurate.
- **No other `realpath` use touches `default_cwd`:** grepped the whole repo outside the four locked files; the remaining hits (`install-manifest.ts`, `fixture-runner.ts`, `workflow-source.ts`, container overlay scripts) are unrelated to codebase registration.

### Verdict: READY

Nothing above is a correctness bug or a reason to hold the PR — they're a documentation gap (finding 1), a UX gap for stranded-row recovery on Windows-CLI-only setups (finding 2), a maintainability note about lock scope (finding 3), and an unverified edge case worth appending to the report's own "Not verified" list (finding 4). The core fix, the seam locks, and the round-trip proof all held up under independent reproduction.

PR #2943, opened 2026-08-28
test(ci): attribute the downloadWebDist Windows stall to a phase, and rule out Defender

Wirasm:
## Problem and outcome

The `downloadWebDist` Windows stall came back. On PR #2923's windows leg (job [98950212570](https://github.com/coleam00/Archon/actions/runs/33189654885/job/98950212570)) `verifies against the embedded hash without fetching checksums.txt` died at 5016 ms with a dangling `tar`, and `ca8df725` — the parent-owned-pipe fix from #2928 — is a verified ancestor of that head. The pipe mechanism is closed with evidence; the ledger's remaining suspect was filesystem-side, specifically Defender scanning the extraction target.

The problem with the recurrence is that it teaches almost nothing. The whole download path emits two log lines, both before any work happens:

```
{"module":"cli.serve","targetDir":"C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\serve-webdist-test-JBKPAk\\target-embedded-ok","msg":"web_dist.download_started"}
{"module":"cli.serve","source":"embedded","msg":"web_dist.checksum_resolved"}
...
::error ...: tar extraction was killed by SIGTERM after 5079ms without finishing (limit 60000ms)
```

Five seconds went somewhere between those lines and there is no way to say where.

This PR ships the instrumentation that fixes that. It also **rules out Defender**, which was the point of the second change it no longer contains — see below.

- **Outcome:** the next 5-second sample bounds the stall to before-tar, during-tar, or after-tar from the CI log alone. Defender path scanning is eliminated as a candidate cause, with evidence.
- **Invariant:** extraction behavior is unchanged. The `Bun.file` stdin shape from #2928, the 60 s bound, the cleanup paths, and every existing assertion stay exactly as they were.
- **Scope boundary:** no test timeout bumps, no platform skips, no change to the extraction mechanism, no new dependencies, and — after the evidence came in — no CI setup step either.

## Review guidance

- **Feedback requested:** whether the phase split is the right granularity, and whether you agree with removing the Defender step rather than keeping it as belt-and-braces (reasoning below).
- **Start here:** `packages/cli/src/commands/serve.ts:120` — the phase chain starts at `web_dist.download_started` and every event below closes one phase.
- **Review order:** `serve.ts` (the diagnostic), then the comment left in `.github/workflows/test.yml` where the step used to be.
- **Known risk or uncertainty:** none in the shipped diff. The uncertainty is upstream of it — the cause of the stall is still unknown, and this PR only guarantees the next sample says where the time went.

## Solution

Each new `web_dist.*` event closes one phase and carries that phase's own `durationMs`, so the phases chain end to end:

| Event | Phase it closes | Fields |
|---|---|---|
| `web_dist.tarball_verified` | fetch, read, hash | `durationMs` |
| `web_dist.archive_staged` | temp-dir reset and staged write | `tarballPath`, `bytes`, `durationMs` |
| `web_dist.extract_spawned` | the `Bun.spawn` call itself | `tarPid`, `durationMs` |
| `web_dist.extract_exited` | waiting on the child | `exitCode`, `signalCode`, `durationMs` |
| `web_dist.installed` | staged-archive removal, layout check, rename | `targetDir`, `durationMs` |

Spawn is deliberately split from the wait, and the Windows baseline this PR produced justifies it: the spawn call is **24 ms** against the child's **133 ms**, so process creation is a real share of the cost rather than a rounding error. Folding them together would hide a stalled `CreateProcess` behind a slow-looking `tar`. `web_dist.installed` covers the mirror case on the other side — renaming a freshly written tree is filesystem work that can stall too.

Same logger, same level, same naming style as the two events already there. Info level is not a stylistic choice: the default level is `info`, so a `debug` event would not appear in the CI log that has to answer the question.

The field is `tarPid`, not `pid`, because pino already binds the parent's pid at the root and a second `pid` key silently wins on parse.

### The Defender step, and why it is not here

A Windows-only `Add-MpPreference -ExclusionPath` step was written, pushed, and run on `windows-latest`. It was built to be loud rather than silent, and that is what settled the question — [its readback](https://github.com/coleam00/Archon/actions/runs/33203158478/job/98957487400):

```
Requesting Defender exclusions for: D:\a\_temp, C:\Users\RUNNER~1\AppData\Local\Temp
Defender ExclusionPath now: C:\, C:\Users\RUNNER~1\AppData\Local\Temp, D:\, D:\a\_temp
```

The step only ever requested the two temp paths. `C:\` and `D:\` were already in Defender's exclusion list, put there by the runner image, and a drive-root exclusion is recursive. **Nothing under either drive was being scanned before the step ran**, so Defender path scanning cannot be what stalls `downloadWebDist`. `Add-MpPreference` also took 5 s per Windows run (19:23:19 → 19:23:24) to change nothing, on a job whose underlying problem is a contention floor.

Keeping it would be machinery that protects nothing, against a mechanism its own output disproves. It is removed, and the falsification is recorded as a comment where the step stood so nobody adds it again.

One detail worth carrying forward for anyone who does reach for this pattern: `RUNNER_TEMP` alone would not have covered the right directory. On hosted runners it is `D:\a\_temp`, while the fixtures extract into `os.tmpdir()`, which reads `TEMP` under the runner profile — `C:\Users\RUNNER~1\AppData\Local\Temp`, a different volume.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Two log lines before any work; a stall is unattributable | Five more phase events, each with its own `durationMs` |
| **Failure behavior** | Unchanged | Unchanged — same errors, same messages, same cleanup |

## Validation

- `bun test src/commands/serve.test.ts` in `packages/cli` — 21 pass, 0 fail, 37 expect() — identical to the pre-change baseline, so no existing assertion moved.
- `bun run test` in `packages/cli` — all 12 file groups green.
- `bun run lint`, `bun run format:check` — clean.
- `bun run validate` — exit 0, 166 groups reporting `0 fail`.
- **The phase chain observed on windows-latest**, which is the evidence that the diagnostic works on the platform it exists for:

```
{"module":"cli.serve","durationMs":2,"msg":"web_dist.tarball_verified"}
{"module":"cli.serve","tarballPath":"C:\\Users\\RUNNER~1\\...\\target-embedded-ok.tmp.tar.gz","bytes":135,"durationMs":11,"msg":"web_dist.archive_staged"}
{"module":"cli.serve","tarPid":5916,"durationMs":24,"msg":"web_dist.extract_spawned"}
{"module":"cli.serve","exitCode":0,"signalCode":null,"durationMs":133,"msg":"web_dist.extract_exited"}
{"module":"cli.serve","targetDir":"C:\\Users\\RUNNER~1\\...\\target-embedded-ok","durationMs":2,"msg":"web_dist.installed"}
```

The phases sum to the 172 ms the runner reported for that test, so the chain accounts for the whole body with nothing unattributed. That is now the healthy Windows baseline to compare the next stall against.

- **Not verified, by nature:** that the stall is gone — nothing here tries to fix it. It is intermittent and cannot be forced, so a green Windows leg proves nothing about it. The checkable claim is narrower: the next recurrence arrives with its time attributed to a phase instead of to a five-second gap.

### Note on the first CI run

The Windows leg on the first push failed, but not on anything in this PR: `migrate-state-dir > subdirectory invocation (C5)`, which #2924 already carries as **RUNNER-FLOOR RESIDUAL** (the subprocess is 88% of the body and is the contract). All four `downloadWebDist` tests passed in that same run. Worth noting that C5 blew its budget *while both temp trees were explicitly Defender-excluded*, which is independent support for the same conclusion.

## Delivery considerations

|---|---|---|
| Observability | Five info-level events on a one-time download path make the stall self-attributing; no other consumer reads these event names | `grep` for `web_dist.` finds no reference outside `serve.ts` |
| Rollout / rollback | Logging only; extraction is untouched and the change is trivially revertible | Diff is log lines plus comments |

## Links

- Related #2924
- Related #2928

## Summary by CodeRabbit

- **Improvements**
  - Added detailed timing information for each stage of web distribution downloads and extraction, making progress and delays easier to identify in logs.
  - Improved extraction event reporting with clearer process details and elapsed-time measurements.
  - Improved Windows test reliability by removing an unnecessary antivirus configuration step.
- **Behavior**
  - No changes to download, extraction, or installation behavior.


Wirasm:
## CI outcome on `071fcd6f` — red Windows leg is not from this diff

Every leg green except `test (windows-latest)`, which failed on:

```
(fail) archon workflow wait against a detached run > wakes on a cancelled run stopped from a third process [6359.00ms]
```

That test lives in `packages/cli/src/commands/workflow-wait.integration.spec.ts`, which **does not exist on this branch**. It arrived from `dev` (PR #2923, run-attention wake) after this branch was cut, and CI runs the PR merge ref — so it executed against merged content. This diff is 49 lines of logging in `serve.ts` plus a 7-line YAML comment and cannot reach it. Same budget-edge class this PR's parent issue #2924 tracks.

All four `downloadWebDist` tests passed: 219 ms, under the print threshold, 203 ms, 125 ms.

### Second independent Windows confirmation of the phase chain

Different runner, same shape as the earlier run — the instrumentation reproduces:

| Phase | Run 1 (`39ca3054`) | Run 2 (`071fcd6f`) |
|---|---|---|
| `web_dist.tarball_verified` | 2 ms | 2 ms |
| `web_dist.archive_staged` | 11 ms | 18 ms |
| `web_dist.extract_spawned` | 24 ms | 32 ms |
| `web_dist.extract_exited` | 133 ms | 171 ms |
| `web_dist.installed` | 2 ms | 2 ms |
| **sum vs reported body** | 172 / 172 ms | 225 / 219 ms |

The phases account for the whole test body both times, with nothing unattributed. Process creation is a consistent 24–32 ms share, which is the reason spawn is logged separately from the wait.

PR #2953, opened 2026-08-28
fix(cli): a Windows cancel no longer fails for a run whose process tree already stopped

### LINKED ISSUE #2946: bug(cli): Windows 'workflow cancel' fails when a descendant exits during the taskkill tree walk, leaving a dead run marked running
## What happens

On Windows, `archon workflow cancel <run-id>` fails and leaves the run marked `running` even when the detached run's process tree was terminated successfully. The operator is told the run "was not changed" and pointed at `archon workflow abandon`, but the work is in fact already dead.

The trigger is a race: `taskkill /T` walks the tree PID by PID, and any descendant that exits on its own between the tree snapshot and the kill makes `taskkill` exit non-zero. Archon treats that non-zero exit as "termination failed".

## Evidence

Observed on Windows CI (`Test Suite`, run 33203641394, job 98959131847), through the new `workflow wait` integration spec that runs `archon workflow cancel` for real:

```
error: cancel failed: Error: Command failed: taskkill /PID 8904 /T /F
ERROR: The process with PID 960 (child process of PID 6632) could not be terminated.
Reason: The operation attempted is not supported.

ERROR: The process with PID 6632 (child process of PID 8904) could not be terminated.
Reason: There is no running instance of the task.
```

Read the PIDs: **8904 is the root, and it has no error line — it was terminated.** The two failures are descendants. 6632 reports *"There is no running instance of the task"*, which is Windows saying it had already exited. `taskkill` still exits non-zero, so the whole cancel is reported as a failure.

## Where

`packages/cli/src/utils/detached-run-control.ts:493`

```ts
if (process.platform === 'win32') {
  // A zero exit from taskkill /T is the platform's positive tree-termination
  // result. Never translate a timeout or partial failure into success merely
  // because the root disappeared; descendants are part of this contract.
  await execFileAsync('taskkill', ['/PID', String(pid), '/T', '/F'], {
    timeout: TERMINATION_GRACE_MS,
    windowsHide: true,
  });
  if (!(await waitUntilGone(() => processExists(pid), TERMINATION_CONFIRM_MS))) {
    throw new Error(`Detached workflow process tree ${String(pid)} is still running`);
  }
  return;
}
```

`promisify(execFile)` rejects on any non-zero exit (verified locally: the rejection message is exactly `Command failed: <cmd>\n<stderr>`, matching the CI output). So the reject happens first and the outcome check on the next line — the one that would actually confirm the tree is gone — is never reached on this path.

The rejection propagates through `target.stop()` in `workflowCancelCommand` (`packages/cli/src/commands/workflow.ts:4399`), which aborts before `abandonWorkflow`. The run row keeps `status = 'running'` for a process that no longer exists.

## The invariant that came apart

The contract is "the process tree is dead afterwards", not "`taskkill` exited 0". Those two are the same thing only when no descendant exits on its own during the walk. The comment above the call is a deliberate decision — descendants genuinely are part of the contract — but the exit code cannot distinguish *"a live descendant survived"* from *"a descendant beat us to it"*, and the current code reads both as failure.

## POSIX is not affected

`process.kill(-pid, SIGTERM)` is one operation against the whole process group, so a partially-exited group does not produce an error at all. Its `ESRCH` path already verifies the outcome (`processExists(pid)`) rather than trusting the error code. The hole is specific to the Windows tree walk.

## Suggested direction

Verify the outcome instead of trusting the exit code, while keeping the descendant guarantee:

1. Capture `taskkill`'s stderr instead of letting a non-zero exit be the verdict.
2. Confirm the root is gone (the existing `waitUntilGone(() => processExists(pid), ...)` check).
3. For each PID `taskkill` names as un-terminated, check it with `processExists(n)`. Fail only if one is still alive.

This preserves "a live descendant still fails the cancel" while letting an already-exited descendant pass. Worth weighing during implementation: PID reuse inside the confirm window could produce a false "still alive", and a Windows Job Object is the structurally correct analogue of a POSIX process group (terminate the job, no tree walk, no race) if the larger change is ever worth it.

## Note on CI

`packages/cli/src/commands/workflow-wait.integration.spec.ts:341` is the first automated coverage of `archon workflow cancel` against a real detached run on Windows. It did not introduce this bug; it exposed one that shipped with the detached-stop work (#2759). Because the race needs a descendant to exit at the wrong moment, it is intermittent — it flakes the Windows leg of `Test Suite` on unrelated PRs. The `EBUSY: resource busy or locked` temp-cleanup warning that follows in the same log is downstream of the aborted cancel (it warns, it does not fail), and should go away with this fix.

Wirasm:
## Problem and outcome

On Windows, `archon workflow cancel <run-id>` failed for detached runs whose process tree had in fact stopped, left the run row marked `running`, and pointed the operator at `abandon` for an orphan that did not exist. Full diagnosis on #2946.

- **Outcome:** a cancel whose tree is gone now succeeds on Windows, matching POSIX.
- **Invariant:** a tree that is still running must still throw.
- **Scope boundary:** the Windows branch of `detached-run-control.ts` only. POSIX untouched — it already handled the equivalent case. No test loosened.

## Review guidance

- **Start here:** `packages/cli/src/utils/detached-run-control.ts:493` — the entire behavior change is this one branch.
- **Known risk, and why it is acceptable:** the non-zero exit was also, incidentally, the only runtime signal that a *live descendant* had survived. This change removes it. That is safe for three reasons.
  1. It was not functioning as a guard. It fired overwhelmingly on the benign race, so anything acting on it was acting on noise — a signal that is wrong most of the time it fires is not protection.
  2. Reconstructing it would mean scraping PIDs out of localized `taskkill` stderr. Third-party output is fair to classify in this repo, but not when the classification breaks silently on a locale CI never exercises.
  3. **The descendant guarantee did not disappear — it moved from an incidental signal to an explicit test.** `stops the detached owner process group before its descendant can leak work` runs on Windows, arms a live descendant, waits for its observable death, and then proves it can no longer act. It consults no exit code, so it protects this contract independently of anything `taskkill` reports. It passes on the Windows leg of this PR (job 98969157875, 422 ms).

## Solution

`taskkill` was the wrong thing to read. It walks the tree PID by PID and exits non-zero when any of them has already exited on its own — the outcome we want, reached early — and nothing in that exit code separates it from a kill that genuinely failed. Because `promisify(execFile)` rejects on any non-zero exit, the rejection fired before the `waitUntilGone` confirmation three lines below, making the only real proof unreachable.

The command still runs the same way; a *completed* walk that reported a failure is captured instead of thrown, and the confirmation decides. A tree still running still throws, now carrying `taskkill`'s own error rather than discarding it, so a missing binary reads as `…is still running: <real error>` instead of a bare symptom.

An **interrupted** walk is not tolerated. On a `timeout` Node kills `taskkill` part-way through the tree, so a descendant it never reached can still be alive while the root happens to be gone — and the root check cannot see that. Those rejections are rethrown, the same shape as the POSIX branch rethrowing anything that is not `ESRCH`. The distinction is structural rather than textual, so it holds under any locale: a timeout is `killed: true` / `signal: 'SIGTERM'`, an ordinary non-zero exit is `killed: false` with a numeric `code`, and a command that never started carries neither.

The comment above the call, which asserted the wrong invariant, is replaced and now states what happens to a timeout.

## Validation

- **Windows CI, job 98969157875** — `wakes on a cancelled run stopped from a third process` **passes** (8750 ms). That is the test that was failing on unrelated PRs. `stops the detached owner process group before its descendant can leak work` also passes, so the descendant guarantee survives the change.
- `bun run validate` — exit 0. `bun --cwd packages/cli run test:integration` — 7 + 1 + 6 pass.
- **The failure case still fails, observed red.** New test `treats an already-gone target as a stopped tree` stages a real already-exited PID behind a stub control endpoint; new test `still fails when the target is alive and the kill cannot reach it` stages a live non-group-leader process, where `kill(-pid)` raises the same `ESRCH` an absent group raises, so only the follow-up check on the process itself separates them.
- **The timeout rethrow is proved, not asserted.** `commandTerminatedBySignal` is unit-tested against the genuine rejection objects Node produces — a real 100 ms timeout, a real `process.exit(3)`, a real `ENOENT` — asserting the shape the classification rests on (`killed`/`signal`/`code`) and not only the verdict, so a Node change that moved it surfaces here. The wiring was then observed flipping on a single variable: forcing the Windows branch with a command that genuinely times out, the already-gone test **fails** with `signal: "SIGTERM"` when the rethrow is present, and **passes** when the one rethrow line is removed — which is exactly the false success R1 identified. (The missing-binary forcing technique cannot reach this path: `ENOENT` carries `killed: undefined`, so it is correctly not a timeout.)
- The Windows branch cannot run on macOS, so I forced it (`taskkill` is absent there, so `execFileAsync` rejects exactly as a non-zero exit does) and ran the suite twice. With the **old** bare-`await` shape the already-gone test **fails**. With the new shape it passes, and a live target still throws:

  ```
  Detached workflow process tree 77370 is still running: Executable not found in $PATH: "taskkill"
  ```

  Both halves of the contract observed, and the kill error confirmed to survive into the message.
- **Not verified:** the real `taskkill` race, which needs a Windows descendant to exit at the wrong moment and cannot be staged deterministically. The green Windows leg above is the proof that matters.

## Links

- Closes #2946
- Related #2923

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows process termination so stopping an already-closed process completes successfully.
  - Preserved clear errors when a live process cannot be reached or does not belong to the expected process group.
  - Improved handling of command timeouts and interrupted termination attempts.

### DIFF
--- a/packages/cli/src/utils/detached-run-control.integration.spec.ts
+++ b/packages/cli/src/utils/detached-run-control.integration.spec.ts
@@ -77,6 +77,31 @@ async function rejectedError(action: () => Promise<unknown>): Promise<Error> {
   throw new Error('Expected the operation to reject');
 }
 
+/**
+ * A control endpoint that hands out `pid` and commits the termination lease.
+ *
+ * The handshake is not what these tests are about: committing it puts the terminator
+ * itself in front of a PID the test chose, which is the only way to stage a target
+ * that is already gone, or one that is alive and out of reach.
+ */
+function stubOwner(pid: number): Server {
+  return createServer((socket: Socket): void => {
+    socket.setEncoding('utf8');
+    let request = '';
+    socket.on('data', (chunk: string): void => {
+      request += chunk;
+      if (request.includes('stop\n')) {
+        socket.write(`${JSON.stringify({ pid })}\n`);
+        request = request.replace('stop\n', '');
+      }
+      if (request.includes('terminate\n')) {
+        socket.write('ready\n');
+        request = request.replace('terminate\n', '');
+      }
+    });
+  });
+}
+
 describe('detached run control integration', () => {
   it('stops the detached owner process group before its descendant can leak work', async () => {
     const runId = `tree-${crypto.randomUUID()}`;
@@ -121,6 +146,69 @@ describe('detached run control integration', () => {
         }
       }
       if (process.platform !== 'win32') rmSync(detachedRunControlPath(runId), { force: true });
+    }
+  });
+
+  it('treats an already-gone target as a stopped tree, not a failed stop', async () => {
+    // #2946: `taskkill /T` walks the tree PID by PID and exits non-zero when one of
+    // them is already gone. Reading that exit code as failure made `archon workflow
+    // cancel` report a failure on Windows for a run whose tree had in fact stopped,
+    // and left the run row saying `running`. POSIX has always tolerated the same
+    // condition as ESRCH; the contract is one tree-is-gone outcome on both branches.
+    const runId = `already-gone-${crypto.randomUUID()}`;
+    const path = detachedRunControlPath(runId);
+    const doomed = spawn(process.execPath, ['-e', 'process.exit(0)'], { stdio: 'ignore' });
+    if (doomed.pid === undefined) throw new Error('Failed to spawn the short-lived target');
+    const gonePid = doomed.pid;
+    await waitForExit(doomed);
+    // The premise, asserted rather than assumed: the terminator is aimed at nothing.
+    await waitFor(() => !processExists(gonePid));
+
+    const server = stubOwner(gonePid);
+    await listen(server, path);
+    try {
+      const target = await requestDetachedRunStop(runId);
+      // Resolving IS the assertion, and letting a rejection through reports the real
+      // reason rather than a matcher's. The old Windows branch rejected here, carrying
+      // taskkill's "There is no running instance of the task" as a stop failure.
+      await target.stop();
+    } finally {
+      await close(server);
+      if (process.platform !== 'win32') rmSync(path, { force: true });
+    }
+  });
+
+  it('still fails when the target is alive and the kill cannot reach it', async () => {
+    // The guardrail for the tolerance above: an unreachable kill must never read as a
+    // stopped tree, or the terminator stops protecting anything. Staged on POSIX,
+    // where a live process that is not a group leader makes `kill(-pid)` raise the
+    // same ESRCH that an entirely absent group raises — so only the follow-up check
+    // on the process itself can tell the two apart. The Windows equivalent, a live
+    // root that survives `taskkill /F`, needs a process the runner is not permitted
+    // to kill and is not worth staging in CI.
+    if (process.platform === 'win32') return;
+
+    const runId = `alive-${crypto.randomUUID()}`;
+    const path = detachedRunControlPath(runId);
+    // Not `detached`, so it joins this spec's process group and no process group
+    // carrying its own PID exists for the terminator to signal.
+    const survivor = spawn(process.execPath, ['-e', 'setInterval(() => undefined, 1000)'], {
+      stdio: 'ignore',
+    });
+    if (survivor.pid === undefined) throw new Error('Failed to spawn the surviving target');
+    const survivorPid = survivor.pid;
+
+    const server = stubOwner(survivorPid);
+    await listen(server, path);
+    try {
+      const target = await requestDetachedRunStop(runId);
+      const error = await rejectedError(async (): Promise<void> => target.stop());
+      expect(error.message).toContain('does not own process group');
+      expect(processExists(survivorPid)).toBe(true);
+    } finally {
+      survivor.kill('SIGKILL');
+      await close(server);
+      rmSync(path, { force: true });
     }
   });
 
--- a/packages/cli/src/utils/detached-run-control.test.ts
+++ b/packages/cli/src/utils/detached-run-control.test.ts
@@ -1,11 +1,26 @@
 import { describe, expect, it } from 'bun:test';
+import { execFile } from 'node:child_process';
 import { createConnection } from 'node:net';
+import { promisify } from 'node:util';
 import {
+  commandTerminatedBySignal,
   detachedRunControlPath,
   DetachedRunOwnerUnavailableError,
   requestDetachedRunStop,
   startDetachedRunControlServer,
 } from './detached-run-control';
+
+const execFileAsync = promisify(execFile);
+
+/** The real rejection `execFile` produces for `run`, so no error shape is hand-built. */
+async function rejectionFrom(run: () => Promise<unknown>): Promise<unknown> {
+  try {
+    await run();
+  } catch (error) {
+    return error;
+  }
+  throw new Error('Expected the command to reject');
+}
 
 describe('detached run control', () => {
   it('treats a connection as a harmless probe and refuses to replace a live owner', async () => {
@@ -41,5 +56,53 @@ describe('detached run control', () => {
     await expect(requestDetachedRunStop(runId)).rejects.toBeInstanceOf(
       DetachedRunOwnerUnavailableError
     );
+  });
+});
+
+/**
+ * The Windows termination path tolerates a kill command that ran and reported a
+ * failure, because `taskkill /T` reports every tree member that had already exited.
+ * It must never tolerate one that was cut off part-way, because then the tree was
+ * only partly walked and the root going quiet says nothing about the descendants.
+ *
+ * Every error below is the genuine object Node produces, not a hand-built stand-in,
+ * so this stays true if Node changes the shape.
+ */
+describe('commandTerminatedBySignal', () => {
+  it('is true for a timeout, which cuts the command off part-way', async () => {
+    const error = await rejectionFrom(() =>
+      execFileAsync(process.execPath, ['-e', 'setTimeout(() => undefined, 30_000)'], {
+        timeout: 100,
+      })
+    );
+
+    // The distinction is structural, so assert the shape the classification rests on
+    // rather than only its verdict — a Node change that moved it would surface here.
+    expect(error).toMatchObject({ killed: true, signal: 'SIGTERM' });
+    expect(commandTerminatedBySignal(error)).toBe(true);
+  });
+
+  it('is false for an ordinary non-zero exit, which is a completed walk', async () => {
+    const error = await rejectionFrom(() =>
+      execFileAsync(process.execPath, ['-e', 'process.exit(3)'])
+    );
+
+    expect(error).toMatchObject({ killed: false, code: 3 });
+    expect(commandTerminatedBySignal(error)).toBe(false);
+  });
+
+  it('is false for a command that never started, which killed nothing', async () => {
+    // Tolerated on purpose: nothing was signalled, so the root check that follows is
+    // a conservative verdict — a live root still throws.
+    const error = await rejectionFrom(() => execFileAsync('archon-no-such-binary-xyz', []));
+
+    expect(error).toMatchObject({ code: 'ENOENT' });
+    expect(commandTerminatedBySignal(error)).toBe(false);
+  });
+
+  it('is false for anything that is not an error object', () => {
+    expect(commandTerminatedBySignal(undefined)).toBe(false);
+    expect(commandTerminatedBySignal('killed')).toBe(false);
+    expect(commandTerminatedBySignal({ signal: 1 })).toBe(false);
   });
 });
--- a/packages/cli/src/utils/detached-run-control.ts
+++ b/packages/cli/src/utils/detached-run-control.ts
@@ -478,6 +478,25 @@ async function waitUntilGone(exists: () => boolean, timeoutMs: number): Promise<
   return true;
 }
 
+/**
+ * Did a child command die by signal instead of exiting with a status of its own?
+ *
+ * This is the line between a command that ran to completion and reported something,
+ * and one that was cut off part-way. Node makes it structural rather than textual:
+ * a `timeout` kill sets `killed: true` with `signal: 'SIGTERM'`, an ordinary non-zero
+ * exit reports `killed: false` with a numeric `code`, and a command that never started
+ * carries neither. Nothing here reads message text, so it holds under any locale.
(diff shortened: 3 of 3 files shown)


Wirasm:
The red Windows leg is not from this PR. `dev` is already red for the same reason — filed as #2954.

**This PR's target behavior passed on Windows** (job 98969157875):

```
(pass) archon workflow wait against a detached run > wakes on a cancelled run stopped from a third process [8750.00ms]
(pass) detached run control integration > treats an already-gone target as a stopped tree, not a failed stop [219.00ms]
(pass) detached run control integration > stops the detached owner process group before its descendant can leak work [422.00ms]
```

The first of those is the test that was failing on unrelated PRs, and it is the reason this change exists. The third is the descendant guarantee, still enforced after the exit code stopped being the verdict.

The single failure is `scripts/output-format-strict.test.ts:146`, which compares Windows backslash paths against a forward-slash literal. It landed with the #2944 merge at 19:45 UTC and fails 100% of Windows runs on `dev` (run 33205244447, job 98964594891) — before this branch existed. This PR cannot go fully green until #2954 lands.

Wirasm:
## Needs fixes

The reported bug is genuinely fixed and the two questions the PR itself raised both hold up: the failure case still fails (independently reproduced, see Validation), and removing `taskkill`'s exit code as a descendant-survival signal is safe because that guarantee now lives in an unmodified, unconditional test that consults no exit code. But the fix's `catch` block is wider than the bug it targets. It swallows a genuine `execFile` timeout the same way it swallows an ordinary benign non-zero exit, and on Windows that timeout case can now report success while a descendant is still alive -- precisely the failure mode the deleted comment named. That is one causal fix, narrow and in the same lines already touched.

**1 blocking (Important) · 2 non-blocking (Suggestion)**

**Validation:** all green -- CLI package test script, lint, `tsc --noEmit`, and an independent standalone reproduction outside the repo (see below). PR's Windows CI leg is red only for an unrelated, already-tracked reason (#2954); the two new tests and the pre-existing descendant-leak test all pass on that same Windows run.

### Findings

|---|---|---|---|
| `R1` | Important | Windows catch treats a genuine `execFile` timeout the same as an ambiguous non-zero exit, so a stalled `taskkill` can now report a live descendant as a stopped tree | OPEN |
| `R2` | Suggestion | Comment above the Windows branch reads as if a taskkill invocation failure is an independent throw trigger; only the confirmation gates the throw | OPEN |
| `R3` | Suggestion | New `stubOwner` test helper duplicates handshake-mocking logic already inline in an untouched nearby test | DECLINED |

<details>
<summary><code>R1</code> — Windows catch swallows a genuine timeout, reopening the "success from a timeout" failure mode</summary>

**Impact:** on a real Windows machine, if `taskkill /PID <pid> /T /F` stalls past `TERMINATION_GRACE_MS` (5000ms -- realistic under load, or with antivirus/EDR intercepting process termination, which is common on the corporate Windows machines this CLI targets), Node force-kills `taskkill.exe` via the `timeout` option before it finishes walking the tree. If the root process happened to already be dead at that point (a plausible ordering depending on how `taskkill` enumerates the tree) while a descendant is still alive, `waitUntilGone(() => processExists(pid), TERMINATION_CONFIRM_MS)` -- which checks only the root PID -- sees the root gone and the function returns success. `archon workflow cancel` then marks the run `cancelled` while a descendant keeps running. This is exactly the scenario the deleted comment existed to forbid ("Never translate a timeout or partial failure into success merely because the root disappeared; descendants are part of this contract"), and it was not reachable before this diff: pre-fix, *any* `execFileAsync` rejection -- timeout included -- propagated immediately and the confirm block never ran, so a timeout was always conservative (throw, never a false success).

**Evidence:** `packages/cli/src/utils/detached-run-control.ts:502-517`. The `catch (error) { killFailure = error instanceof Error ? error : new Error(String(error)) }` block does not inspect `error.killed` or `error.signal` before deferring to the confirm poll. Verified directly (independent of the reviewing agent) that Node's `execFile` rejection already carries this distinction with no locale-dependent stderr parsing required:
```
timeout kill:           killed: true   signal: 'SIGTERM'  code: null
ordinary nonzero exit:  killed: false  signal: null       code: 1
binary missing:         killed: undefined                 code: 'ENOENT'
```
The pre-existing, unmodified descendant-leak test (`stops the detached owner process group before its descendant can leak work`) runs `taskkill` to normal completion on a small tree; it never exercises the `timeout` path, so it provides no coverage for this specific gap. Codebase precedent for making exactly this distinction already exists: `packages/cli/src/commands/serve.ts:203` treats a signal-kill from a timeout as categorically different from a plain nonzero exit; `packages/workflows/src/executor-shared.ts:203-216`'s `RawSubprocessError` explicitly models `killed` for the same reason.

**Required outcome:** in the catch block, check `(error as NodeJS.ErrnoException & { killed?: boolean }).killed` (or `.signal`) before deferring to `killFailure`. A `killed`-flagged rejection means `taskkill` was cut off mid-sweep or never ran -- rethrow it immediately, the same way the POSIX branch already rethrows any non-`ESRCH` failure instead of deferring to the confirm poll. Only an ordinary non-zero-exit rejection (`killed: false`) is the ambiguous "already gone vs. genuine failure" case this PR is meant to fix; that is the only case that should fall through to `waitUntilGone`.

**Disposition:** OPEN. This is a single call site, not a repeated pattern -- `grep -rn "waitUntilGone|processGroupExists"` across the repo matches only this file, and the only other `taskkill`/`execFile` termination call sites (`packages/providers/src/community/opencode/runtime.ts`, bundled workflow cleanup scripts) are fire-and-forget best-effort cleanup with no confirmation contract, so no class enumeration applies. The correction is small and localized to the lines already touched by this PR.

</details>

<details>
<summary><code>R2</code> — Comment overstates the throw condition</summary>

**Impact:** none on behavior; a future reader could misread "A tree still running, or a taskkill that never ran, still throws here" as two independent throw triggers, when there is exactly one gate (the `waitUntilGone` confirmation three lines below). A `taskkill` that never ran (e.g. `ENOENT`) paired with a tree that is independently already gone correctly succeeds today, `killFailure` discarded -- the comment's phrasing doesn't make that clear.

**Evidence:** `packages/cli/src/utils/detached-run-control.ts:494-501`.

**Required outcome:** tighten the sentence to name the single gate explicitly, e.g. "A tree still running still throws here, whether or not taskkill itself ran cleanly -- the confirmation below is the only gate."

**Disposition:** OPEN. Non-blocking; worth folding into the same commit as R1's fix since both touch this comment's neighborhood.

</details>

<details>
<summary><code>R3</code> — <code>stubOwner</code> duplicates existing handshake-mocking logic</summary>

**Impact:** the new `stubOwner` helper (`detached-run-control.integration.spec.ts:84-101`) reimplements the stop/terminate handshake parsing already written inline in the untouched "fails when the owner closes before committing termination" test a few dozen lines below -- same chunk-buffering, same `stop\n`/`terminate\n` substring checks. Now duplicated in three places instead of one, in the same file.

**Evidence:** `packages/cli/src/utils/detached-run-control.integration.spec.ts:84-101` vs. the pre-existing inline handshake logic later in the same file.

**Required outcome:** none required now.

**Disposition:** DECLINED. AGENTS.md favors duplicating small local logic "until a pattern has repeated and stabilized," and folding the untouched third test into a shared helper would pull unrelated, already-passing test code into this bug-fix PR's diff for no behavioral gain. Revisit only if this mock is touched again.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `seams` | R1 |
| `simplify` | R2, R3 |

#### Validation

|---|---|---|
| `bun test src/utils/detached-run-control.integration.spec.ts` (packages/cli) | PASS | 7/7 pass locally, including both new tests |
| `bun test src/utils/detached-run-control.test.ts` (packages/cli) | PASS | 2/2 pass |
| `bun run test` (packages/cli, full CI script) | PASS | 21+45+3+115+2+7+1+6+20 tests, 0 fail |
| `bun run lint` (repo root) | PASS | clean, all packages |
| `bun x tsc --noEmit` (packages/cli) | PASS | clean |
| Windows CI, job 98969157875 (PR's actual head) | PASS (targeted) | Both new tests and the unmodified descendant-leak test pass on Windows (`treats an already-gone target...` 219ms, `stops the detached owner process group...` 422ms). Job's overall red is caused solely by an unrelated, separately-tracked bug in `scripts/output-format-strict.test.ts` (#2954, fix already open as PR #2957) -- inherited red, not caused by this PR. |
| Standalone reproduction (outside repo, `taskkill` absent on macOS -> genuine `ENOENT` rejection used as a real, structurally-equivalent stand-in for "execFileAsync's promise rejected") | Confirms fix | Pre-fix logic rejects unconditionally for both an already-gone target and a genuinely-alive target (reproducing the reported bug); post-fix logic resolves for the already-gone target and still rejects for the genuinely-alive one, with the underlying error preserved in the thrown message. |

</details>

Wirasm:
R1 addressed at `e162ed59`, and current `dev` (including #2957) is merged in at `d5b27aab`.

**The catch was too wide.** It now rethrows a rejection that carries `killed`/`signal` — an interrupted walk — and tolerates only a command that ran to completion and reported something. Same shape as the POSIX branch rethrowing anything that is not `ESRCH`. The comment at the call site is reworded and now says what happens to a timeout.

**Proving it, since the missing-binary forcing technique cannot reach this path** (`ENOENT` carries `killed: undefined`, so it is correctly not classified as a timeout):

1. `commandTerminatedBySignal` is unit-tested against the genuine rejection objects Node produces — a real 100 ms timeout, a real `process.exit(3)`, a real `ENOENT`. Each case asserts the shape the classification rests on, not only the verdict, so a Node change that moved it fails here:
   ```
   timeout   -> { killed: true,  signal: 'SIGTERM' }  -> true
   exit 3    -> { killed: false, code: 3 }            -> false
   no binary -> { code: 'ENOENT' }                    -> false
   ```
2. The wiring was observed flipping on a single variable. Forcing the Windows branch with a command that genuinely times out, `treats an already-gone target as a stopped tree`:
   - **fails** with `signal: "SIGTERM"` with the rethrow present
   - **passes** with the one rethrow line removed — the exact false success R1 describes

The already-gone case still passes on the real path: `bun --cwd packages/cli run test:integration` is 7 + 1 + 6 pass, unit is 6 pass, `bun run validate` exits 0.

Left the `stubOwner` dedup alone as agreed.

PR #2958, opened 2026-08-28
fix(workflows): dry-run tolerates a missing stub on an all_done join

### LINKED ISSUE #2869: fix(workflows): dry-run outcome ignores join tolerance — a missing stub under an all_done join fails the whole simulated run
## Problem

Dry-run computes the simulated run outcome as "any node failed anywhere" (`packages/workflows/src/dry-run.ts:1186` at time of writing). The real executor's `trigger_rule: all_done` tolerates a failed or skipped dependency at the join. So a fixture that omits a stub for a barrier/join node reports `outcome: failed` even though the production run would proceed past that join.

Proven by the #2807 incident: five `*-delivered` fixtures failed solely because the new `review-complete` join node lacked a stub — the dry-run trace marked it failed and the blanket anyFailed flipped every run, making a sound composition look broken. Root cause: https://github.com/coleam00/Archon/issues/2807#issuecomment-5444412470

## Desired outcome (recommended shape)

Dry-run's outcome computation mirrors executor join semantics for the same graph: a per-node failure that a downstream `all_done` join tolerates must not, by itself, flip the simulated run outcome. Missing stubs stay loudly reported (`missingStubs` already exists in the JSON result) — the point is truthful attribution: "missing stub" is an authoring gap; "run failed" is a simulation verdict. Today the former masquerades as the latter.

## Invariant

For any graph and stub set, dry-run's terminal status/outcome matches what executor trigger-rule semantics would produce given the same per-node results. A fixture author must not need to stub a node whose absence production would tolerate.

## Regression proof

- A fixture omitting a stub for a `bash:` join under `all_done`: dry-run applies the join's tolerance, completes, and reports the missing stub distinctly.
- The five #2807 fixtures, pre-repair, would have failed for the true reason or passed — not failed on a simulation artifact.

## Notes

Not blocking: the five fixtures are being repaired by adding the missing stub, which is the correct near-term fix regardless. This issue is the class behind the incident — every future join/barrier node re-creates the trap until dry-run models tolerance.

Wirasm:
## Problem and outcome

`dryRunWorkflow` computed a simulated run's outcome as "any node ended in `failed` anywhere in the graph → failed." A node with no stub was always forced into `failed`, even for a `bash:`/`script:` barrier node declaring `trigger_rule: all_done` — a node whose entire job is to run regardless of what happened upstream. That's what caused the #2807 incident: five `*-delivered` fixtures failed solely because the new `review-complete` join node had no stub, and the blanket failure check flipped every run's outcome even though a real run would have proceeded past that join.

- **Outcome:** A missing stub on an `all_done` node no longer fails the simulated run. Dry-run hydrates it with a generated placeholder (the same mechanism `--default-stubs` already uses), lets downstream `all_success` nodes run instead of being skipped, and still records the gap in `missingStubs` so it stays visible.
- **Invariant:** A missing stub is only ever evidence of an authoring gap in the fixture, never conflated with a genuine simulated failure, for any node whose own `trigger_rule` says it tolerates upstream failure.
- **Scope boundary:** The fix lives in `simulateNode`'s single-node dispatch fallback, so it also covers an `all_done` join nested inside a `loop_group` body — `simulateLoopGroup` dispatches every body node through that same fallback, and a test now proves a tolerated body join keeps the group (and the run) completing. A bare `loop:` node is still out of scope: `simulateLoop` has its own, separate iteration-based missing-stub handling that this change doesn't touch, and extending tolerance there would need new, unevidenced design (a fake iteration count or until-condition) rather than reuse of this mechanism.
- **Root cause:** Dry-run has a "missing stub" state with no equivalent in production (production just executes the node). That dry-run-only state was being conflated with a genuine failure specifically on nodes whose declared purpose is to tolerate whatever upstream did.

## Review guidance

- **Feedback requested:** Correctness of the new branch, and whether leaving bare `loop:` nodes out of scope (see Scope boundary above) is acceptable or should be tracked as follow-up.
- **Start here:** `packages/workflows/src/dry-run.ts:1105` — the new `trigger_rule === 'all_done'` branch inside `simulateNode`'s "no stub" fallback; this is the core behavioral change. Two follow-on changes complete it: a new `toleratedMissingStubs` field (threaded through `dryRunResultSchema`, `DryRunContext`, and the return value) records which entries in `missingStubs` were tolerated rather than blocking, and `fixture-runner.ts`'s `checkFixture` filters by it so a fixture whose only gap is a tolerated join no longer fails.
- **Review order:** `dry-run.ts:1105` (the fix, plus the `toleratedMissingStubs` threading) → `dry-run.test.ts` (three new unit tests: the new tolerance, the prior behavior unchanged for plain nodes, and the same tolerance proven for an `all_done` join nested in a `loop_group` body) → `fixture-runner.ts`'s `checkFixture` (the blocking-vs-tolerated filter) → `fixture-runner.test.ts` (the new fixture-level regression test).
- **Lower-attention areas:** None — this is still a small change, now spanning two files (`dry-run.ts`, `fixture-runner.ts`) plus their tests.
- **Known risk or uncertainty:** Production's own overall run-status computation (`dag-executor.ts`'s `anyFailed` check) is the same kind of blanket "any node failed anywhere → run failed" check dry-run had, with no sink/reachability reasoning of its own. What `trigger_rule` actually controls in production is only whether a node runs vs. is skipped. This fix targets the narrower, provable dry-run-only gap rather than reshaping dry-run's outcome computation around a reachability model production itself doesn't implement.

## Solution

In `simulateNode`'s final "node has no stub" fallback, check the node's own `trigger_rule` before deciding the outcome:
- `trigger_rule: all_done` → tolerate the missing stub the same way `--default-stubs` tolerates any missing stub: hydrate the node with `generatedStubFor`'s placeholder, mark it `completed` in the recorded output (so downstream `all_success` nodes run instead of being skipped), push a `'stubbed'` trace entry naming the reason, and still add the node id to `ctx.missingStubs`.
- Anything else → unchanged (`recordFailed`, as before).

No change was needed to the outcome computation itself (`dry-run.ts:1249`, the blanket `anyFailed` check) — once the node's output `state` is `completed` instead of `failed`, that check naturally stops counting it.

Two follow-on changes close the loop for consumers that read `missingStubs` as "this fixture never verified this node":
- A new `toleratedMissingStubs` field (schema, `DryRunContext`, and the return value) records which entries in `missingStubs` were tolerated by an `all_done` join rather than genuinely blocking.
- `fixture-runner.ts`'s `checkFixture` now filters `missingStubs` by `toleratedMissingStubs` before deciding whether a fixture fails on "reached nodes without stubs" — a fixture whose only gap is a tolerated join now passes, with the gap still visible in the JSON report's `missingStubs`.

## Validation

- `cd packages/workflows && bun test src/dry-run.test.ts` — 69 pass, 0 fail.
- `bun test src/dry-run.test.ts src/fixture-runner.test.ts` — 108 pass, 0 fail.
- `bun x tsc --noEmit` (packages/workflows) — clean.
- `bun run scripts/lint.ts` (repo root) — clean.
- `bun run test` (packages/workflows full suite, 27 grouped test invocations) — all `0 fail`.
- `bun packages/cli/src/cli.ts workflow test` across all 39 bundled `.stubs.yaml` fixtures — 39 passed, 0 failed, no change from baseline.
- Reproduced the #2807 shape directly: temporarily removed the `review-complete: "settled"` stub from `.archon/workflows/sdlc/review/fixtures/docs-auto-declined.stubs.yaml` and ran `archon workflow test archon-review --json`. Before this fix it would report `outcome: "failed"` and fail the fixture; after, the fixture reports `pass: true`, `outcome: "completed"` (matching the fixture's declared `expect`), with the gap still disclosed via `missingStubs: ["review-complete"]` in the JSON report — the authoring gap stays visible without failing the run or the fixture. The fixture file was restored afterward (empty diff).
- **Not verified:** Bare `loop:` nodes declaring `trigger_rule: all_done` — out of scope, see Scope boundary above. (An `all_done` join inside a `loop_group` body is now covered — see the nested test in `dry-run.test.ts`.)

## Links

- Closes #2869

## Summary by CodeRabbit

* **Bug Fixes**
  * Dry runs now distinguish missing stubs tolerated by `all_done` workflow joins from blocking missing stubs.
  * Workflows and fixture tests with only tolerated missing stubs can complete successfully while still reporting them.
  * Reports and workflow test results identify tolerated missing stubs separately.
  * Missing stubs that cannot generate valid placeholders continue to fail as expected.

* **Documentation**
  * Updated CLI guidance for `all_done` joins and missing stubs.


Wirasm:
# Review report — PR #2958 (dry-run tolerates a missing stub on an all_done join)

## Verdict

**ready: true** — action: **none**. All three Important findings from the prior round are verified fixed against the current code, with no regressions and no new blocking findings.

## Accepted contract

Source: PR #2958 body, closing issue #2869 (full text in scope.md; unchanged since round 1).

**Required outcome:** `simulateNode`'s final "no stub" fallback (single-node dispatch path only) checks the node's own `trigger_rule` before deciding the outcome. For `trigger_rule: all_done`, a missing stub is tolerated the same way `--default-stubs` tolerates any missing stub — hydrated with `generatedStubFor`'s placeholder, the node marked `completed`, a `'stubbed'` trace entry naming the reason pushed, and the node id still added to `missingStubs` so the gap stays visible. Any other node is unchanged (`recordFailed`).

**Invariant:** a missing stub must never be conflated with a genuine simulated failure for a node whose own `trigger_rule` tolerates upstream failure — and missing stubs must stay loudly reported regardless of outcome, for every consumer of that signal (not only `dryRunWorkflow`'s own `outcome` field).

**Explicit non-goals (unchanged from round 1):**
- Loop and `loop_group` nodes *declaring* `trigger_rule: all_done` themselves — `simulateLoop`'s separate iteration-based check, still untouched.
- No change to `dry-run.ts`'s blanket `anyFailed` outcome check beyond what naturally follows from a node's stored state changing to `completed`.
- Production's `dag-executor.ts` `anyFailed` run-status computation — a known, accepted asymmetry this PR does not attempt to fix.

## Reviewed head SHA

`f6a254d7af02311737e6b7af4bc3f42de2eba2e5` (verified against `gh pr view 2958 --json headRefOid`; matches scope.md and local `HEAD`; working tree clean).

## Findings

No new blocking findings. One non-blocking documentation suggestion (R4) below.

## Prior findings

| ID | Severity | Sources | Claim | Round 1 verdict | Round 2 verdict |
|---|---|---|---|---|---|
| R1 | Important | code, seams | `fixture-runner.ts`'s own `missingStubs.length > 0` gate still failed the exact `workflow test` scenario the PR was scoped to fix. | Open | **Fixed** |
| R2 | Important | tests | No test covered the tolerance nested inside a `loop_group` body, where it changes the group's own failure aggregation. | Open | **Fixed** |
| R3 | Important | simplify | The new `all_done` branch duplicated the adjacent pre-existing "hydrate + trace a stubbed node" block. | Open | **Fixed** |

**R1 — verified fixed.** `dry-run.ts` now carries an additive `toleratedMissingStubs: string[]` field on `dryRunResultSchema` (`dry-run.ts:372`) and `DryRunContext` (`dry-run.ts:486`), populated alongside `missingStubs` only in the `all_done` tolerance branch (`dry-run.ts:1107-1110`). `fixture-runner.ts:576-578` now computes `blockingMissingStubs = result.missingStubs.filter(id => !result.toleratedMissingStubs.includes(id))` and gates on that instead of the raw list (`fixture-runner.ts:579-584`); `missingStubs` itself is still reported unfiltered on the returned `FixtureCheckResult` (`fixture-runner.ts:592`), so the gap stays visible exactly as the invariant requires — only the pass/fail decision changed.

I re-ran the review's own R1 reproduction independently rather than trusting the claim: temporarily replaced `packages/workflows/src/dry-run.ts` and `fixture-runner.ts` with their pre-round-1 versions (`git show 55634207:...`) and ran the two new unit tests (`dry-run.test.ts`'s `loop_group` case and `fixture-runner.test.ts`'s tolerated-join case) — both failed against the old code exactly as expected (`outcome: "failed"` / `report.passed === 0`), confirming they are genuine regression tests and not vacuously true. Restored the current files afterward (`git status` clean, no diff introduced by the experiment). Against the current code, `cd packages/workflows && bun test src/dry-run.test.ts src/fixture-runner.test.ts` — **108 pass, 0 fail**.

Also re-confirmed the "one consumer that already worked" claim from round 1: `packages/cli/src/commands/workflow.ts:1850` (`workflow dry-run --json`) does `writeJsonLine(result)`, spreading the whole `DryRunResult` — the new field rides along automatically, needing no separate CLI change. `bun x tsc --noEmit` clean in both `packages/workflows` and `packages/cli`.

**R2 — verified fixed.** The new test (`dry-run.test.ts:763-800`, `'tolerates a missing stub on an all_done node nested in a loop_group body (#2869)'`) builds a `loop_group` whose body has a plain node and an `all_done` join with no stub, and asserts the group and run both complete. I traced the causal path directly: `simulateLoopGroup` (`dry-run.ts:884-895`) dispatches body nodes through `simulateNodes` → `simulateNode` (the exact function this PR edits), then aggregates `failed = ...some(body => bodyOutputs.get(body.id)?.state === 'failed')` (`dry-run.ts:895`). `completedOutput` (`dry-run.ts:429-444`) — used by both the explicit-stub and tolerated-stub paths — always sets `NodeOutput.state: 'completed'`, regardless of the *trace* entry's separate `'stubbed'` label, so a tolerated join no longer trips the group's own failure aggregation. Confirmed this is a real, non-trivial regression test the same way as R1: reverting to pre-round-1 `dry-run.ts` fails this exact test (`Expected: "completed", Received: "failed"`).

**R3 — verified fixed.** `dry-run.ts:1079-1129` now has one merged branch keyed on `const tolerated = stub === undefined && node.trigger_rule === 'all_done'`, entered when `stub !== undefined || tolerated`. Read the full function (`dry-run.ts:1067-1142`) to check the merge didn't disturb adjacent control flow: the exec-code branch above it changed from `if (stub !== undefined) { ...; return; } if (isExecNode(node) && ctx.execCode) {...}` to `if (stub === undefined && isExecNode(node) && ctx.execCode) {...}` — logically identical (an explicit stub still takes precedence over real execution, exactly as before), just re-expressed as one condition instead of an early return. The merged trace-push object produces the identical field set as the old explicit-stub block when `tolerated` is false (the `reason` spread contributes nothing), so no observable behavior changed for the pre-existing explicit-stub path. `bun test` confirms (108 pass, 0 fail, same count as the implementation's own claim).

## Re-evaluated gate: docs (new surface in this round's delta)

Round 1 gated `docs` off because the diff touched only `dry-run.ts` implementation and unit tests, with no new YAML field, CLI flag, or user-facing surface. This round's delta adds one: `toleratedMissingStubs`, a new field on the shipped `dryRunResultSchema` — reachable via `archon workflow run --dry-run --json` and `archon workflow test --json`. I evaluated `docs` against this new field alone (the only gated-off lens this round's delta could re-open; `code`, `seams`, `simplify`, `tests`, and `errors` all ran in round 1 and are not re-run here).

**Result: undocumented, non-blocking (R4, Suggestion).** `packages/docs-web/src/content/docs/reference/cli.md:325` documents this exact JSON output shape in prose — "a reachable AI, bash, or script node without a stub fails the simulation and appears in `missingStubs`... Supplied stubs for unknown or unreachable nodes appear in `unusedStubs`" — but never mentions `toleratedMissingStubs`. `grep -rn "toleratedMissingStubs" packages/docs-web/` returns nothing. This is a real gap in the same paragraph that documents its sibling fields, but I do not treat it as a blocker: the field is purely additive (no existing documented behavior became false — `missingStubs` still means exactly what the docs say, a superset that now happens to include some non-blocking entries), no CI or lint check in this repo enforces CHANGELOG or per-field doc coverage for JSON output schemas, and scope.md's accepted contract does not name documentation as part of the required outcome. Filing another correction round for a one-sentence doc addition would cost more than recording it here. **Recommended follow-up:** add one clause to `cli.md:325` naming `toleratedMissingStubs` alongside `missingStubs`/`unusedStubs`.

## Rejected findings

None. All three prior findings survived verification as genuinely fixed; no new Critical or Important finding was raised this round.

## Suggestions

- **R4** (see above) — document `toleratedMissingStubs` in `cli.md` next to `missingStubs`/`unusedStubs`. Non-blocking.

## Discoveries

0 accepted discoveries (unchanged from round 1). No `discoveries/` producer directory exists for this run, and this round's verification work stayed inside the accepted contract — nothing proved a defect outside it. See `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/7f0a4baa26189cfb95ba8c4661a08c1b/discoveries.json` and `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/7f0a4baa26189cfb95ba8c4661a08c1b/discoveries.md`. Adjacent discoveries never affect readiness (moot here — there are none).

## Review coverage

Continuation review: one continuation reviewer verified the round-1 report's three open findings and reviewed the correction delta (`27d4d84d..f6a254d7`, three commits) in full — the round-1 specialist lenses (`code`, `seams`, `simplify`, `tests`, `errors`) did not rerun; their round-1 findings are carried forward above with per-finding verdicts, per this run's `code.md`/`seams.md`/`simplify.md`/`tests.md`/`errors.md`.

Verification performed this round, independent of the implementation's own claims:
- Read the full delta diff for all four changed/added files (`dry-run.ts`, `fixture-runner.ts`, `dry-run.test.ts`, `fixture-runner.test.ts`).
- Read `simulateNode`, `simulateLoopGroup`, and `completedOutput` in full to confirm the causal mechanisms R1–R3 describe are real, not just asserted.
- Reproduced R1 and R2 as genuine regressions by running the two new tests against a temporarily-restored pre-round-1 version of `dry-run.ts`/`fixture-runner.ts` (both failed as expected), then restored the current files and confirmed a clean `git status`.
- Ran `bun test src/dry-run.test.ts src/fixture-runner.test.ts` (108 pass), `bun test src/commands/workflow.test.ts` in `packages/cli` (331 pass), `bun x tsc --noEmit` in both `packages/workflows` and `packages/cli` (clean), and `bun run scripts/lint.ts` at the repo root (clean) — all matching or consistent with `implementation.md`'s own validation log.
- Re-evaluated the `docs` gate against this round's one new piece of surface (`toleratedMissingStubs`) per the continuation-review re-evaluation clause; found it genuinely undocumented but resolved it as a non-blocking Suggestion (R4) rather than reopening the round.

Evidence not independently re-obtained: the implementation's real-fixture reproduction against `.archon/workflows/sdlc/ship/fixtures/direct-delivered.stubs.yaml` via a full scratch-copy `workflow test ship --json` run, and the repo-root `bun run validate` full-monorepo pass — both are consistent with the unit-level and package-level checks I ran directly, which exercise the identical code paths.

Wirasm:
## Needs fixes

The fix itself is correct and well-scoped: a `trigger_rule: all_done` node missing its dry-run stub now hydrates with a generated placeholder instead of failing the simulated run, exactly closing the #2807/#2869 gap, and the new `toleratedMissingStubs` field is a minimal, additive way to carry "this gap didn't block" without breaking the existing `missingStubs` contract. All three reviewers independently confirmed the core branch, the `loop_group`-nested case, and the field-pairing invariant are sound. But the PR wired the new field into only one of the two places that turn `missingStubs` into a failure message, and it can mark a node "tolerated" before the placeholder generation that tolerance depends on has actually succeeded — both are the same class of bug the PR itself was written to fix (a missing-stub signal misattributed as, or hiding, a real failure), just recurring one hop over from where the diff touched.

**2 blocking · 2 non-blocking**

**Validation:** type-check and lint clean; all directly relevant unit tests pass (see coverage table).

### Findings

|---|---|---|---|
| `R1` | Important | `workflow run --dry-run`'s CLI failure message uses the unfiltered `missingStubs` list, misattributing a tolerated node as a cause of failure | OPEN |
| `R2` | Important | A tolerated node can be marked `toleratedMissingStubs` before its placeholder generation is known to succeed, mislabeling a genuine failure as tolerated | OPEN |
| `R3` | Suggestion | Bare `loop:` nodes declaring `trigger_rule: all_done` still fail on a missing stub — disclosed, deliberate scope boundary | OPEN |
| `R4` | Suggestion | `cli.md` reference docs still state unconditionally that a missing stub fails the simulation, no longer true for an `all_done` join | OPEN |

<details>
<summary><code>R1</code> — CLI dry-run failure message doesn't filter by <code>toleratedMissingStubs</code></summary>

**Impact:** `packages/cli/src/commands/workflow.ts:1855-1859` (`workflow run --dry-run`, not `workflow test`) throws `` `Dry-run failed; missing stubs: ${result.missingStubs.join(', ')}` `` using the raw, unfiltered `missingStubs` — the same field `fixture-runner.ts`'s `checkFixture` already filters through `toleratedMissingStubs` (`fixture-runner.ts:585-593`) before building its own failure reason. A run that fails for one genuinely blocking node while a separate `all_done` node is merely tolerated names both in the thrown message, with nothing distinguishing which one is the real cause — the exact misattribution #2869 fixed for the fixture path, recurring in the CLI's own `--dry-run --json` path (the primary machine-readable signal; the trace dump that would otherwise disambiguate this is skipped or scrolled past).

**Evidence:** `packages/cli/src/commands/workflow.ts:1855-1859`. Reproduced directly against the PR's `dryRunWorkflow`: a workflow with one plain node missing its stub (`gate`, blocking) and one `trigger_rule: all_done` node also missing its stub (`review-complete`/`cleanup`, tolerated) yields `missingStubs: ["gate","review-complete"]`, `toleratedMissingStubs: ["review-complete"]`, and the CLI message reads `Dry-run failed; missing stubs: gate, review-complete` — no marker for which one is real. The existing test covering this message construction (`workflow.test.ts:1044-1058`) only exercises a single all-blocking `missingStubs: ['node']`, so it can't catch the mixed case.

**Required outcome:** Filter `result.missingStubs` by `result.toleratedMissingStubs` before building the CLI error message — the same one-line filter `checkFixture` already applies — plus a test covering a run with both a blocking and a tolerated missing stub together.

**Found by:** `prp-core:seam-analyzer`, `prp-core:code-simplifier` (independently, same defect)

**Disposition:** OPEN — one-line fix consistent with the pattern already used in `fixture-runner.ts`.

</details>

<details>
<summary><code>R2</code> — <code>toleratedMissingStubs</code> can be set before the placeholder generation it depends on succeeds</summary>

**Impact:** `packages/workflows/src/dry-run.ts:1105-1111` adds `node.id` to both `ctx.missingStubs` and `ctx.toleratedMissingStubs` before calling `generatedStubFor(node)` (inside `completedOutput(node, stub ?? generatedStubFor(node))`). `generatedStubFor` throws for a legitimate, reachable state — an `output_format` with `$async: true`, an uncompilable schema, or a schema-invalid placeholder (`dry-run.ts:156-193`) — since `output_format` is available on any node type, including a `trigger_rule: all_done` node. That throw is caught by the enclosing `try/catch` (`dry-run.ts:1139`), which calls `recordFailed`, correctly marking the node `failed` and the run's `outcome: 'failed'` — but nothing removes the id already added to `toleratedMissingStubs`, so the failed node is reported as tolerated. This directly contradicts the field's own documented contract (`dry-run.ts:365-371`): "hydrated with a generated placeholder rather than failing the node."

**Evidence:** `packages/workflows/src/dry-run.ts:1105-1139`. Reproduced directly: a `prompt:` node with `trigger_rule: all_done`, `output_format: { $async: true }`, and no stub produces `outcome: "failed"`, a `failed` trace entry for that node, and `toleratedMissingStubs: ["review-complete"]` — a node that did not tolerate anything, reported as though it did.

**Required outcome:** Compute the placeholder before recording tolerance — e.g. `const placeholder = stub ?? generatedStubFor(node);` ahead of the `ctx.missingStubs.add`/`ctx.toleratedMissingStubs.add` calls — so a thrown `generatedStubFor` falls through to the existing non-tolerated "missing stub" path instead of pre-recording tolerance that didn't happen.

**Disposition:** OPEN. Lower reach than R1 (requires an incompatible `output_format` on an `all_done` node, not merely a coexisting blocking gap) but same class of defect and same file the PR is already touching.

</details>

<details>
<summary><code>R3</code> — Bare <code>loop:</code> nodes are the one dispatcher the tolerance fix didn't reach</summary>

**Impact:** `simulateLoop` (`dry-run.ts:800-869`) still calls the unconditional `stubFor`/`recordFailed` path (`dry-run.ts:834-844`) with no `trigger_rule` check — a bare `loop:` node (not a `loop_group` body, which does inherit the fix via `simulateNode`) declaring `trigger_rule: all_done` and missing its stub still fails the simulated run.

**Evidence:** `dry-run.ts:800-869` vs the fixed `dry-run.ts:1105-1130`. All three reviewers independently confirmed `simulateLoop`'s per-iteration completion check (`until:`/`until_field` against a hydrated placeholder) is structurally different from `simulateNode`'s single-shot dispatch, so naively reusing the same "generate a placeholder" mechanism isn't a same-shape copy — it needs its own design, as the PR body itself states. A repository-wide search for a bare `loop:` node combined with `trigger_rule: all_done` across `.archon/workflows/**` found none — this is latent, not currently shipping.

**Required outcome:** No code change required for this PR. File a follow-up issue referencing #2869 to track `simulateLoop` + `trigger_rule: all_done` tolerance as designed, separate work.

**Disposition:** OPEN (non-blocking) — the author disclosed this exact boundary in the PR body and asked reviewers to judge it; independent review agrees the boundary is justified, not premature narrowing. No existing GitHub issue tracks it yet, so it isn't `TRACKED FOLLOW-UP`; recommend filing one before or shortly after merge.

</details>

<details>
<summary><code>R4</code> — Reference docs still describe the pre-fix strict-failure behavior</summary>

**Impact:** `packages/docs-web/src/content/docs/reference/cli.md:248` ("without this flag, missing reachable stubs remain an error") and `:325` ("a reachable AI, bash, or script node without a stub fails the simulation") both state unconditionally what is no longer true for a `trigger_rule: all_done` node after this PR.

**Evidence:** `packages/docs-web/src/content/docs/reference/cli.md:248,325`.

**Required outcome:** Update both lines to note the `all_done` tolerance, or leave as a documented follow-up — no runtime consequence, only a stale mental model for a reader of the CLI reference.

**Disposition:** OPEN (non-blocking).

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | R1 (contributed), R2, R3 |
| seams | R1, R3 |
| simplify | R1, R3, R4 |

#### Validation

|---|---|---|
| `cd packages/workflows && bun test src/dry-run.test.ts src/fixture-runner.test.ts` | PASS | 109 pass, 0 fail |
| `cd packages/cli && bun test src/commands/workflow.test.ts` | PASS | 331 pass, 0 fail |
| `cd packages/workflows && bun x tsc --noEmit` | PASS | clean |
| `cd packages/cli && bun x tsc --noEmit` | PASS | clean |
| `bun run scripts/lint.ts` (repo root) | PASS | clean, no errors across all linted packages |

</details>

PR #2959, opened 2026-08-28
fix(ci): the strict-schema guard now passes on Windows

### LINKED ISSUE #2954: bug(ci): output-format-strict test compares Windows paths against forward slashes, failing every Windows run on dev
## What happens

The Windows leg of `Test Suite` fails on `dev` on every run, and therefore on every PR branched from it. It is deterministic, not a flake.

```
(fail) bundled output_format schemas satisfy OpenAI strict mode (#1843) >
       the check inspects the schemas it claims to, including loop_group bodies
  at D:\a\Archon\Archon\scripts\output-format-strict.test.ts:146
```

First seen on the `dev` merge of #2944 at 19:45 UTC on 2026-08-28 (run 33205244447, job 98964594891), which is where `scripts/output-format-strict.test.ts` landed. Every PR opened since inherits it — for example run 33206586605.

## Cause

`scripts/output-format-strict.test.ts:140` builds each entry from a native filesystem path:

```ts
seen.push(`${file.slice(REPO_ROOT.length + 1)}:${String((node as { id?: unknown }).id)}`);
```

`file` comes from `join()` while walking `WORKFLOWS_ROOT`, so on Windows it carries backslashes and the slice yields `.archon\workflows\sdlc\implement\archon-implement.yaml:implement`. Line 146 then asserts a forward-slash spelling:

```ts
expect(seen).toContain('.archon/workflows/sdlc/implement/archon-implement.yaml:implement');
```

The two can never match on Windows. The failure output in CI shows exactly this — every listed entry uses `\\`.

## Fix

Normalize the separator where the relative path is built, at both slice sites, so the reported paths are stable across platforms:

```ts
const relative = file.slice(REPO_ROOT.length + 1).replaceAll('\\', '/');
```

`scripts/output-format-strict.test.ts:107` has the same slice feeding `violations[].workflow`. That one only reaches failure messages, so it is cosmetic rather than a failing assertion, but it should be normalized in the same pass so a Windows violation report reads the same as a POSIX one.

## Why it matters now

This is the second Windows CI blocker on `dev` today (the first is #2946). Unlike that one it fails 100% of the time, so no PR can show a green Windows leg until it is fixed.

Wirasm:
## Problem

The output-format strict-mode guard (landed with #2944) builds workflow paths with `join()` and asserts a forward-slash spelling, so its completeness test fails on every `windows-latest` run of `dev` — deterministically. Every PR since inherits a red Windows leg, and four dogfood runs died tonight at their CI gates on it. Filed as #2954.

## Solution

One `relativePath()` helper normalizes the separator at both slice sites (the violation reports and the completeness assertion), so the test means the same thing on every platform. No behavior change to what the guard checks.

## Validation

- `bun test ./scripts/output-format-strict.test.ts` — 2 pass / 0 fail (the normalize is a no-op on POSIX; the Windows leg of this PR is the real proof).
- Prettier clean.

Closes #2954

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized workflow path formatting across platforms.
  * Fixed path reporting on Windows by using consistent forward-slash separators.
  * Updated related validation to recognize the normalized path format.

### DIFF
--- a/scripts/output-format-strict.test.ts
+++ b/scripts/output-format-strict.test.ts
@@ -34,6 +34,12 @@ import { readdirSync, statSync } from 'node:fs';
 import { join } from 'node:path';
 
 const REPO_ROOT = join(import.meta.dir, '..');
+
+// join() emits backslashes on Windows; reported paths and the fixture assertion
+// below use one spelling so the test means the same thing on every platform (#2954).
+function relativePath(file: string): string {
+  return file.slice(REPO_ROOT.length + 1).replaceAll('\\', '/');
+}
 const WORKFLOWS_ROOT = join(REPO_ROOT, '.archon', 'workflows');
 
 interface Violation {
@@ -104,7 +110,7 @@ function findViolations(): Violation[] {
         if (outputFormat === undefined) continue;
         const nodeId = String((node as { id?: unknown }).id ?? '(unnamed)');
         for (const found of underRequiredNodes(outputFormat, 'output_format')) {
-          violations.push({ workflow: file.slice(REPO_ROOT.length + 1), nodeId, ...found });
+          violations.push({ workflow: relativePath(file), nodeId, ...found });
         }
       }
     }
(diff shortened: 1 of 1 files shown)


Wirasm:
Superseded by #2957, which fixed the same #2954 defect with the better shape — one repoRelative helper at the source instead of normalizing at each consumer. Closing unmerged; branch deleted.

PR #2970, opened 2026-08-29
fix(workflows): warn on single-quoted output refs

### LINKED ISSUE #2284: fix(workflows): validator only flags double-quoted $node.output — single-quoted wrapping is unflagged and equally wrong
## Problem

The validator rule that catches quoted `\$node.output` refs in `bash:` bodies (`packages/workflows/src/validator.ts:687`) matches **double** quotes only:

```
/(?:^|[=\s])"[^"\n]*\$[a-zA-Z_][a-zA-Z0-9_-]*\.output/m
```

Single-quoted wrapping has the same problem and produces no warning. Live example on `dev` — `.archon/workflows/test-workflows/e2e-deterministic.yaml:40`:

```yaml
bash: "echo 'merge-ok: true=\$branch-true.output false=\$branch-false.output'"
```

Substitution wraps string values in literal single quotes (`shellQuote`, `dag-executor.ts:662-664`), so a single-quoted wrapper yields the same literal-quote-characters-in-the-value bug the double-quote rule exists to prevent.

## Consequence

#2242 was closed as "clean `validate workflows` run", and it is — but only against half the pattern. Anyone reading that as "no quoting bugs remain" is wrong.

## Proposal

Extend the rule to cover single-quoted wrapping, then sweep whatever it surfaces. Keep the hint pointing at the assignment form (`var=\$node.output`), which is the shape that's safe at both the inline and the >32 KB `\$(cat ...)` substitution sizes.

## Acceptance criteria

- [ ] Single-quoted wrapping produces the same warning class as double-quoted
- [ ] Unit test for both quote styles
- [ ] `e2e-deterministic.yaml:40` and any other hits fixed
- [ ] Hint text recommends the assignment form

Surfaced reviewing #2262. Related: #2242, and the guard proposed alongside this.

### LINKED ISSUE #2996: Quote lint re-lists the shell template surface instead of deriving it from the template walker
## Problem

`validator.ts`'s quoted-output-ref lint hand-enumerates the three shell-surface fields it checks:

```ts
if (isExecNode(node) && node.runtime === 'sh') warnQuotedOutputRef(node.script, 'bash');
if (isLoopNode(node) && node.loop.until_bash) { ... }
if (isLoopGroupNode(node) && node.loop_group.until_bash) { ... }
```

Those three branches happen to match exactly the three entries `template-walker.ts`'s `SLOT_SPEC`
tags `surface: 'shell'` today (`exec.bash`, `loop.until_bash`, `loop_group.until_bash`), but nothing
connects them. The walker is the runtime authority for pre-quoted substitution — `dag-executor.ts`
reads `surface === 'shell'` to decide that a `$node.output` value is injected already shell-quoted,
which is precisely the condition this lint warns about.

If someone adds a fourth `surface: 'shell'` slot, the executor inherits pre-quoting automatically
(the walker's stated purpose per #2488) while this lint silently stays quiet on it. The three `if`
branches compile regardless of `SLOT_SPEC`'s contents, so there is no compiler error and no failing
test — the drift is invisible until an author hits the footgun on the new field.

The hand-picked pattern predates the walker (introduced in #1894). #2936 migrated `dag-executor.ts`,
`loader.ts`, and `include-expander.ts` to the walker but left `validator.ts` untouched.

## Suggested fix

Derive the lint's field selection from the walker instead of re-listing it:

```ts
visitNodeTemplateSlots(node, slot => {
  if (slot.surface === 'shell') warnQuotedOutputRef(slot.value, slot.name);
});
```

The warning's `field` value is a user-visible string in `archon validate workflows` output, so check
that the slot names the walker yields still read well there (`bash` vs `exec.bash`) before switching,
or map them.

## Context

Surfaced by `prp-core:seam-analyzer` while reviewing PR #2970 as finding R3, and explicitly scoped
out of that PR: the duplication is pre-existing and PR #2970 changed only the matching logic inside
`warnQuotedOutputRef`, not the field-selection structure. No current coverage gap.

Wirasm:
## Problem and outcome

The workflow validator warned when a `$node.output` reference in a bash body was double-quoted, but silently accepted the equivalent single-quoted form (#2284). Both wrappers turn Archon's already shell-quoted substitution into a value containing literal quote characters.

Fixing that exposed a second problem. The lint had two hand-written pieces of knowledge that the engine already owns: which fields are shell source, and where the output references in them are. It re-listed the three shell fields in `if` branches that no longer track `SLOT_SPEC` (#2996), and it re-discovered references with its own pattern. Both are now taken from the template walker.

- **Outcome:** The validator warns for either quote character, on every field the walker tags `surface: 'shell'`, using the engine's own reference grammars.
- **Invariant:** Unquoted assignment-form references, non-shell surfaces (scripts, prompts, conditions), and correct code that merely contains a quote character stay quiet.
- **Scope boundary:** Only this lint and the invalid examples it surfaces. Substitution behavior is unchanged.

## Review guidance

- **Feedback requested:** Whether the two quoting rules below are the right pair, and whether each earns its place. Both are load-bearing against a real shipped body; the evidence is in Validation.
- **Start here:** `packages/workflows/src/validator.ts:883` — the comment states both rules and what they cost; the code under it is the whole check.
- **Review order:** The walker call and surface filter, then `isQuotedAt`, then the tests, then the corrected workflow examples.
- **Lower-attention areas:** `output-ref.ts` gains `LOOP_PREV_OUTPUT_REF_SOURCE` beside the existing `OUTPUT_REF_SOURCE`; `include-expander.ts` now builds its identical pattern from it instead of repeating the literal.

## Solution

The lint no longer decides anything the engine already knows.

**Which fields.** `visitNodeTemplateSlots` yields the node's template slots; the lint keeps those with `surface: 'shell'`. That is the same tag `dag-executor.ts` reads to decide a slot gets the pre-quoted substitution, so the lint and the runtime cannot disagree, and a future shell slot inherits the check instead of silently escaping it. The `runtime === 'sh'` test disappears too: the walker already emits `exec.script` for non-shell runtimes. `slot.path` supplies the field label, so the user-visible names (`bash`, `loop.until_bash`, `loop_group.until_bash`) are unchanged.

**Where the references are.** Built from `OUTPUT_REF_SOURCE` and the new `LOOP_PREV_OUTPUT_REF_SOURCE`. The two grammars are disjoint — `$LOOP_PREV.<id>.output` does not match the plain form — so both are needed, and neither is re-spelled here.

**Whether a reference is quoted.** With the position known, this is a local question, answered by replaying quote state across the reference's own line. Two rules:

- **Line-local.** A quote still open at end of line does not reach the lines below. `gh pr create --body "$(cat <<'EOF'` leaves a `"` open for the rest of the heredoc; the references inside are correctly unquoted.
- **Operand boundary.** The opening quote must sit at line start, after `=`, or after whitespace. Otherwise a mid-word apostrophe opens a phantom string (`echo don't; x=$a.output`), and so does the closing quote of an earlier string (`echo "hi"; x=$a.output`).

The boundary rule costs one true positive: a quoted reference reached from a non-operand position, as in `x=prefix"$n.output"`. No body in the shipped corpus is written that way.

### What this replaces

The first version of this PR used a whole-body quote-state scanner. Review found it false-positived on four shipped workflows, including the active `archon-deliver` pipeline, because a bash body is full of apostrophes that are not string quotes. The second version used an anchored regex. Both encoded the same two rules above, but each buried them in a pattern that had to be re-derived to be checked. Deriving the positions from the walker leaves the rules as the only thing this code says.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `status='$emit.output.status'` produced no validator warning. | It produces the same bash warning as a double-quoted reference. |
| **Field coverage** | Three hardcoded `if` branches, silently stale if `SLOT_SPEC` gains a shell slot. | Every `surface: 'shell'` slot, from the walker. |
| **Failure behavior** | Invalid quoted references could reach execution carrying literal quote characters. | The validator directs authors to the unquoted assignment form. |

## Validation

- `bun run --cwd packages/workflows test src/validator.test.ts` — 220 pass, 0 fail.
- **Every rule seen red.** Each was broken once, in isolation, and a named test failed:

  | Mutation | Test that failed |
  |---|---|
  | Boundary rule always true | `no false positive: an unbalanced apostrophe earlier on the ref's own line` |
  | Scan whole body, not the ref's line | `no false positive: prose apostrophes and heredoc bodies elsewhere in the script` |
  | Drop the `$LOOP_PREV` grammar | `warns on single-quoted output refs in loop and loop_group until_bash` (+1 more) |
  | `recursive: true` (double-report bodies) | `warns on double-quoted output refs in top-level and nested loop_groups` |
  | `slot.name` instead of `slot.path` | 3 bash-field tests |
  | Drop the `surface: 'shell'` filter | `script nodes are exempt from the quoted-output lint` |

  Two of these initially found nothing, which is why the corresponding tests were rewritten: the nested-loop_group test filtered by exact field name (hiding a duplicate reported under a nested path), and the script-exemption test filtered on `field === 'bash'` (hiding warnings raised on `script`, prompt, and condition slots).
- `bun run cli validate workflows` over the whole shipped corpus — the set of warned (workflow, node) pairs is **identical** to the pre-PR `dev` baseline: 17 before, 17 after, zero added, zero removed.
- `bun run cli validate workflows <name>` on the four files earlier review named — `archon-deliver`, `archon-fix-github-issue-experimental`, `e2e-structured-output` report `ok`; `t1-fix-issue` reports only its pre-existing, unrelated `loop_group.until` deprecation warning.
- `bun run validate`, `bun run lint`, `bun run type-check`, `bun run format:check` — passed.

## Links

- Closes #2284
- Closes #2996

## Summary by CodeRabbit

- **Bug Fixes**
  - Workflow output references in shell commands now expand to their actual values instead of being treated as literal text when incorrectly wrapped in quotes.
  - Validation detects improperly quoted output references across workflow command fields, including loop-related commands.
  - Improved validation accuracy reduces false warnings in comments, heredocs, and other valid shell syntax.

- **Tests**
  - Added regression coverage for quoted references, loop outputs, and shell quoting edge cases.
  - Updated workflow smoke tests to verify downstream nodes receive expanded output values.


Wirasm:
# Review report — PR #2970

## Verdict

**Ready — action: `none`.** The correction fixes the only open Important finding and its focused regressions prove all three supported lint entry points.

## Accepted contract

Warn when a `$node.output` reference is wrapped in single quotes in a `bash:` body, using the same warning class as double quotes. The lint remains limited to `bash`, `loop.until_bash`, and `loop_group.until_bash`; legitimate bash quoting stays quiet, including unquoted assignment-form references, non-`sh` scripts, and an unrelated quoted string followed by an unquoted reference. The assignment-form hint and substitution behavior remain unchanged. The checked-in deterministic workflow no longer contains invalid quoted references and validates clean. Scope does not extend beyond this lint and the invalid examples it surfaces.

## Reviewed head SHA

`2388527c6d538a0f4cf1a9bb1836ef0915f8e6ac`

## Findings

### Fixed

- **R1** — `sources: [code]` — **Fixed at `2388527c6d538a0f4cf1a9bb1836ef0915f8e6ac`.** The prior matcher missed single-quoted output substitutions following a shell-word prefix or crossing a newline. `packages/workflows/src/validator.ts:881-907` now scans complete single- and double-quoted segments, while `:921-926` continues to apply it only to the three accepted shell-substitution paths. The added regression at `packages/workflows/src/validator.test.ts:1038-1072` asserts warnings for the prefix and multiline shapes across `bash`, `loop.until_bash`, and `loop_group.until_bash`; the existing quiet tests at `:1074-1100` preserve the unrelated-quote and unquoted-reference boundary.

  I verified the runtime boundary: all three fields pass `escapedForBash: true` to `substituteNodeOutputRefs` in `packages/workflows/src/dag-executor.ts:3737`, `:4753-4759`, `:5226-5232`, and `:6783-6789`. This satisfies the prior causal-class invariant: every single-quoted segment reaching a shell-escaped substitution must warn. Discovery method: inspected the correction delta and ran the package test suite. Affected members: `bash`, `loop.until_bash`, and `loop_group.until_bash`. Examined-clean members: non-`sh` exec nodes and unquoted references after earlier quoted strings.

### Rejected findings

- **R2** — `sources: [tests]` — **Disproved.** A fixture-specific permanent test is not required: the contract calls for a lint regression and a clean validation of the corrected deterministic workflow. `bun run --cwd packages/cli cli validate workflows e2e-deterministic` completed with `1 valid, 0 with errors`.

### Suggestions

None.

## Prior findings

| ID | Prior status | Current verdict |
| --- | --- | --- |
| R1 | Open Important | Fixed at `2388527c6d538a0f4cf1a9bb1836ef0915f8e6ac` |
| R2 | Rejected | Disproved; remains non-blocking |

## Discoveries

0 accepted discoveries. [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/242f94bf8fb4d01ad45129f26a183bb0/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/242f94bf8fb4d01ad45129f26a183bb0/discoveries.md) contain the consolidated result. Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and the correction delta. The original selected concerns were the quoted-output lint's shell parsing boundary, the three shared `bash` / `loop.until_bash` / `loop_group.until_bash` paths, quiet valid forms, and regression coverage. No specialist lens reran. I inspected all changed code and its material runtime consumers, verified the prior R1 correction, and ran `bun run --cwd packages/workflows test`, `bun run --cwd packages/workflows type-check`, `bun run --cwd packages/cli cli validate workflows e2e-deterministic`, and `git diff --check e6a03d67571a702b52062220597cd98c90399c72 2388527c6d538a0f4cf1a9bb1836ef0915f8e6ac`; all passed. There was no evidence I could not inspect.

Wirasm:
## Ready to merge

Round 3 rebuilds the lint a third time — no longer a hand-written regex or scanner at all, but
`visitNodeTemplateSlots` (the walker) supplying both which fields are shell source and where the output
refs sit, plus a small line-local quote-state replay to answer only "is this ref quoted." This is a
deviation from the operator's simpler adjacency spec, and I independently re-ran the measurement behind
that deviation rather than accepting the PR's numbers: pure adjacency (ref sandwiched directly between
matching quotes with nothing else inside) misses exactly 4 real embedded violations across the corpus
(`git commit -m "Release $bump-version.output.newVersion"` and three siblings) that line-local replay
correctly catches, with zero regressions. I reproduced 4 of the PR's 6 claimed "seen red" mutations
myself, including both tests that were rewritten rather than just extended — and proved each rewrite was
load-bearing by showing the *old* assertion style would have passed (a false green) against the exact
same mutation. A three-way corpus sweep across fresh, independent worktrees (pre-PR `dev`, the round-2
regex head, this round's walker head) produced a byte-identical 17-pair warning set at every point. The
walker-based field selection genuinely delivers what #2996 asked for — verified against the issue's own
suggested fix, including a deliberate improvement over it (`slot.path` instead of `slot.name`, which
#2996 flagged as needing a check and which the shipped code gets right). All four prior findings are now
FIXED.

**Validation:** All focused checks passed, including independent mutation-testing reproduction, a
three-way corpus parity sweep, an adjacency-vs-replay corpus comparison, and a full-corpus zero-occurrence
check for the one disclosed detection gap.

**Resolved:** 4 (R1, R2, R3, R4)

### Findings

|---|---|---|---|
| `R1` | Important | Quote-detection false-positived on real shipped workflows | FIXED |
| `R2` | Suggestion | Warning message omitted one quoted-form example | FIXED |
| `R3` | Suggestion | Quote-lint hand-enumerated fields instead of deriving them from the walker | FIXED |
| `R4` | Suggestion | Design comment didn't explain the matcher's non-obvious trade-off | FIXED |
| `R5` | Suggestion | Round-2 PR body's "Outcome" line overclaimed relative to the corpus | FIXED |

<details>
<summary><code>R1</code> — quote-detection false-positived on real shipped workflows</summary>

**Correction (round 3):** `packages/workflows/src/validator.ts:880-943` (`309f5c078`) replaces the
round-2 anchored regex with `visitNodeTemplateSlots(node, slot => {...}, { recursive: false })`, filtered
to `slot.surface === 'shell'`, combined with a small `isQuotedAt` helper that replays quote state only
from the start of the ref's own line up to the ref (line-local), and only counts a quote as real when it
opened at an operand boundary (line start, after `=`, or after whitespace) — the same two rules as
round 2, now applied at each engine-located ref position instead of encoded into one regex.

**This round's independent verification (beyond re-reading the PR's claims):**
- **Deviation measurement, rerun myself.** The operator's simpler spec (a ref immediately adjacent to a
  matching quote pair, nothing else inside) is not what shipped. I patched a pure-adjacency variant into
  this exact checkout and ran the full corpus sweep: 13 pairs vs. the shipped code's 17 — a 4-pair gap,
  matching the PR's claim exactly. I read the source at all 4 missing pairs
  (`archon-release`'s `commit-and-push`/`validate-state`, `e2e-copilot-all-nodes-smoke`'s
  `structured-check`, `e2e-structured-output-failfast`'s `bad-ref`) and confirmed each is a genuine
  embedded violation — e.g. `git commit -m "Release $bump-version.output.newVersion"` — where the ref
  sits inside a quoted string but is not textually touching the quote character. Ruling: **the deviation
  stands.** It is not a return to the rejected round-1 scanner — the round-1 scanner tracked quote state
  across an entire multi-line body with no position anchor, which is exactly what produced the comment/
  heredoc false positives; this round's replay is bounded to one line and gated by the same
  operand-boundary rule round 2 already used, just evaluated at a walker-supplied position instead of
  inside a monolithic pattern. A one-line, boundary-gated replay from an engine-located ref position is a
  narrower mechanism than what was rejected, not a reintroduction of it.
- **Mutation reproduction (4 of 6 rows, not just the requested minimum of 2).** Patched each mutation into
  this checkout and ran the test suite:
  - `recursive: true` (double-report bodies) — **the *rewritten* nested-loop_group test fails**, showing
    the exact predicted duplicate: `["outer", "loop_group.nodes.0.loop_group.until_bash"]`. I then
    reconstructed the *round-2-style* assertion (`field === 'loop_group.until_bash'` filter,
    `toHaveLength(2)`) against the identical mutation: **1 pass, 0 fail — a false green.** The rewrite is
    load-bearing, not cosmetic.
  - Drop the `surface: 'shell'` filter — **the *rewritten* script-exemption test fails**
    (`Expected length: 0, Received length: 1`). Reconstructed the round-2-style assertion
    (`field === 'bash'` filter): **1 pass, 0 fail — a false green.** Also load-bearing.
  - Boundary rule always `true` — the new "unbalanced apostrophe" test fails as expected.
  - Drop the `$LOOP_PREV` grammar — exactly 2 tests fail, matching the PR's own "(+1 more)" annotation.
- **Three-way corpus parity, independent worktrees.** Built three throwaway worktrees — pre-PR `dev`
  (`8a5c905531`), the round-2 regex head (`3e82542ac`), and this round's head (`309f5c078`) — ran
  `cli validate workflows` (full 74-workflow corpus, no args) in each, and parsed every
  `(workflow, node)` warning pair. All three sets are identical: 17 pairs, byte-for-byte.
- **Disclosed cost, corpus-verified.** The comment states the boundary rule gives up a ref reached from a
  non-operand position (`x=prefix"$n.output"`). I instrumented the real `isQuotedAt` call site to log
  every ref immediately preceded by a quote character that itself is not at an operand boundary, and ran
  it across the full corpus: **zero hits.** The disclosed gap is real (by construction) but currently
  unexercised by any shipped workflow.
- Re-ran the 4 originally-named files: all clean, `t1-fix-issue` shows only its pre-existing unrelated
  deprecation warning — unchanged from round 2.

**Found by:** `prp-core:code-simplifier` (round 1, original finding); coordinator (rounds 2-3
verification and this round's mutation/corpus reproduction).

**Disposition:** FIXED — now on a structurally different, more defensible mechanism than either prior
version, with independently-reproduced evidence at every claimed measurement.

</details>

<details>
<summary><code>R2</code> — warning message omitted one quoted-form example</summary>

**Disposition:** FIXED (unchanged from round 2 — the message at `validator.ts:936` still lists all four
quoted forms; this round did not touch it).

</details>

<details>
<summary><code>R3</code> — quote-lint hand-enumerated fields instead of deriving them from the walker</summary>

**Correction:** field selection is now `visitNodeTemplateSlots(node, slot => { if (slot.surface !== 'shell') return; ... })`
— exactly the mechanism #2996 asked for. The `runtime === 'sh'` guard disappears entirely: the walker
already only emits `exec.bash` for `sh` scripts (`exec.script` otherwise), so the surface filter alone
is sufficient. Re-confirmed `template-walker.ts`'s `SLOT_SPEC` still has exactly three `surface: 'shell'`
entries (`exec.bash`, `loop.until_bash`, `loop_group.until_bash`), unchanged by this PR, and that
`dag-executor.ts` still reads the same tag as the runtime authority for pre-quoted substitution — so the
lint and the runtime genuinely cannot disagree.

**Delivers #2996's ask, checked against the issue itself, not just the PR's claim:** #2996's suggested
snippet used `slot.name`, and explicitly flagged a risk: "check that the slot names the walker yields
still read well [in `archon validate workflows` output]... before switching, or map them." The shipped
code uses `slot.path` instead, which — traced through `template-walker.ts`'s `walk()` — yields the exact
pre-PR user-visible field labels (`bash`, `loop.until_bash`, `loop_group.until_bash`), not the walker's
internal slot names (`exec.bash`, ...). This is a considered improvement over the issue's own suggestion,
not a shortcut around it, and it's covered by a seen-red mutation (`slot.name` instead of `slot.path` →
3 bash-field tests fail).

**Scope of the accompanying grammar consolidation:** `output-ref.ts` gains `LOOP_PREV_OUTPUT_REF_SOURCE`
beside the pre-existing `OUTPUT_REF_SOURCE`; `include-expander.ts`'s previously-duplicated
`LOOP_PREV_OUTPUT_REF_PATTERN` now derives from it (`new RegExp(LOOP_PREV_OUTPUT_REF_SOURCE, 'g')`)
instead of repeating the literal. **In scope, not creep** — the walker-based lint needs a shared grammar
for both `include-expander.ts`'s existing rewrite and the lint's new detection, so consolidating that one
grammar is required by the shape actually shipped, not incidental cleanup. **The disclosed remaining
duplicate at `dag-executor.ts:1583` is correctly left alone**: its pattern captures an *optional*
`.field` suffix (`\.output(?:\.([a-zA-Z_][a-zA-Z0-9_]*))?`) that the shared `LOOP_PREV_OUTPUT_REF_SOURCE`
deliberately doesn't — it serves runtime substitution, a different consumer with a different shape, not
a copy of the same grammar. I also found a third occurrence, `when-atom.ts:86`
(`LOOP_PREV_WHEN_ATOM_PATTERN`) — undisclosed in the PR body, but correctly out of scope too: it's an
anchored (`^...$`) two-level-field condition-atom parser for `when:` expressions, a categorically
different grammar (whole-string parsing vs. unanchored substring scanning) serving
`condition-evaluator.ts`, not a copy of the lint's detection grammar.

**Found by:** `prp-core:seam-analyzer` (round 1, original finding); coordinator (this round, verified
against #2996's actual text and traced the `slot.path` vs. `slot.name` choice through the walker).

**Disposition:** FIXED. Issue #2996 is genuinely resolved by this PR's own diff — its "Closes #2996" is
earned, not just linked.

</details>

<details>
<summary><code>R4</code> — design comment didn't explain the matcher's non-obvious trade-off</summary>

**Correction:** `validator.ts:880-899` now documents both load-bearing rules (line-locality and the
operand-boundary anchor) with the exact real-world justification for each — a heredoc example for
line-locality, an apostrophe/closing-quote example for the boundary rule — and states the one cost the
boundary rule accepts. This is materially more complete than round 2's comment, which only covered the
boundary rule.

**Disposition:** FIXED.

</details>

<details>
<summary><code>R5</code> — round-2 PR body's "Outcome" line overclaimed relative to the corpus</summary>

**Correction:** the PR body was rewritten this round. The new Outcome line reads: "The validator warns
for either quote character, on every field the walker tags `surface: 'shell'`, using the engine's own
reference grammars" — a scoped, mechanism-accurate claim that no longer implies the whole corpus is now
free of both quote forms. The 17-pairs-unchanged fact is still honestly disclosed in the Validation
section, and now nothing elsewhere in the body contradicts it.

**Disposition:** FIXED.

</details>

<details>

#### Reviewer coverage

| Scope | This round |
|---|---|
| `code` | R2, R4 verified fixed |
| `seams` | R3 verified fixed against the actual issue text and traced through the walker |
| `simplify` | R1 verified fixed via independent mutation testing and corpus measurement |

No specialist reviewer agent re-ran this round. Given the correction diff is small and fully bounded
(2 commits, 4 files, ~115 lines) and every one of the three original scopes' findings resolves to a
directly falsifiable, mechanism-level claim (a regex/walker behavior, a corpus count, a red/green test
result), the coordinator verified each with direct reproduction — actual mutation testing with
false-green disproof on the two rewritten tests, three independent fresh-worktree corpus sweeps, and a
corpus-wide instrumented zero-occurrence check — rather than dispatching a fresh agent pass to re-read
the same bounded diff.

#### Validation

|---|---|---|
| `bun run --cwd packages/workflows test validator.test.ts` | PASS | 220 pass, 0 fail |
| `bun x tsc --noEmit` (workflows package) | PASS | No type errors |
| `eslint validator.ts validator.test.ts output-ref.ts include-expander.ts` | PASS | No lint errors |
| `bun x prettier --check .` | PASS | All files formatted |
| `git diff --check` (bounded round-3 diff) | PASS | Exit 0 |
| Mutation: `recursive: true` | FAIL as expected | New test catches it; reconstructed round-2-style assertion is a false green |
| Mutation: drop `surface: 'shell'` filter | FAIL as expected | New test catches it; reconstructed round-2-style assertion is a false green |
| Mutation: boundary rule always `true` | FAIL as expected | "unbalanced apostrophe" test fails |
| Mutation: drop `$LOOP_PREV` grammar | FAIL as expected | Exactly 2 tests fail, matching the PR's "(+1 more)" |
| Adjacency-vs-replay corpus comparison | Reproduced | 13 pairs (adjacency) vs. 17 (shipped); the 4-pair gap is all genuine embedded violations, read and confirmed at source |
| Non-operand-position cost, corpus-instrumented | PASS | Zero occurrences across the full 74-workflow corpus |
| Corpus parity: pre-PR `dev` vs. round-2 head vs. round-3 head | PASS | Byte-identical 17-pair set at all three, three independent worktrees |
| `cli validate workflows` on the 4 originally-named files | PASS | All clean; `t1-fix-issue` shows only its pre-existing unrelated warning |
| Heredoc fixture fidelity check | PASS | The new test's heredoc body is line-for-line the real shape in `t1-fix-issue.yaml`'s `open-pr` node |
| Issue #2996 | VERIFIED DELIVERED | Suggested fix's `slot.name` risk correctly avoided in favor of `slot.path`; "Closes #2996" is earned |

</details>

PR #2971, opened 2026-08-29
Fix duplicate release checksum entry

Wirasm:
## Problem and outcome

`checksums.txt` in the v0.7.0 release listed `archon-web.tar.gz` twice. This is a security-adjacent integrity artifact, so its entries must match the release assets exactly.

- **Outcome:** The release job writes one checksum row for each of the six current release artifacts.
- **Invariant:** `checksums.txt` contains exactly one entry for `archon-darwin-arm64`, `archon-darwin-x64`, `archon-linux-arm64`, `archon-linux-x64`, `archon-windows-x64.exe`, and `archon-web.tar.gz`.
- **Scope boundary:** The release artifact set, upload step, and checksum algorithm are unchanged.
- **Root cause:** `archon-*` already matches `archon-web.tar.gz`, which was also passed explicitly to `sha256sum`.

## Review guidance

- **Feedback requested:** Confirm the glob covers the complete uploaded release-artifact set while removing the duplicate row.
- **Start here:** `.github/workflows/release.yml:280` — the only behavior change is checksum input selection.
- **Review order:** Review the checksum-command change and its focused regression test, then compare the command with the release upload list immediately below it.
- **Lower-attention areas:** None; the PR changes one workflow command and adds a focused regression test.
- **Known risk or uncertainty:** None. The release job still uploads the same six artifacts and `checksums.txt`.

## Solution

Checksum generation now passes only `archon-*` to `sha256sum`. That glob covers the five platform binaries and `archon-web.tar.gz`, so removing the redundant explicit operand preserves complete checksum coverage while preventing the second web-tarball row.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `checksums.txt` had seven rows, including `archon-web.tar.gz` twice. | `checksums.txt` has six rows, one for each release artifact. |
| **Integrity coverage** | All six artifacts were covered, but one was duplicated. | All six artifacts remain covered exactly once. |

## Validation

- Fixture reproduction of the prior checksum command — produced seven rows and a duplicate `archon-web.tar.gz` entry.
- Fixture run of the updated command — produced exactly the six unique artifact names in the invariant above.
- `bun install --frozen-lockfile` — passed.
- `bun run format:check` — passed.
- `bun run lint --max-warnings 0` — passed.
- `bun run validate` — passed, including generated-artifact checks, API types, type checks, lint, formatting, installer tests, and the full test suite.
[...]

Issue #2377, opened 2026-08-02
ci: checksums.txt lists archon-web.tar.gz twice (glob already matches the explicit operand)


Wirasm:
Cosmetic, but it is in a security-adjacent artifact that users are told to verify against, so it is worth closing.

## Verified

`checksums.txt` on the v0.7.0 release lists `archon-web.tar.gz` twice:

```
50bdd03e...  archon-darwin-arm64
880afc07...  archon-darwin-x64
217284ef...  archon-linux-arm64
c900a44f...  archon-linux-x64
1f14402b...  archon-web.tar.gz      <-- first
d59a79ae...  archon-windows-x64.exe
1f14402b...  archon-web.tar.gz      <-- duplicate, same hash
```

Root cause is one line in `.github/workflows/release.yml:280`:

```bash
sha256sum archon-* archon-web.tar.gz > checksums.txt
```

The `archon-*` glob **already matches** `archon-web.tar.gz`, and it is then named again explicitly. `sha256sum` dutifully hashes it twice.

## Why it matters

Low severity — the two lines agree, so no verifier is misled about the actual hash. But:

- Any consumer that parses `checksums.txt` into a map keyed by filename silently overwrites; one that builds a list gets a phantom seventh entry for six files.
- A release-asset count check that cross-references `checksums.txt` line count against uploaded assets will be off by one.
- It is the file we point users at for integrity verification, so it should be exactly right.

## Fix

Drop the redundant operand:

```bash
sha256sum archon-* > checksums.txt
```

`archon-*` covers all five binaries plus the web tarball. Verified against the v0.7.0 asset list: `archon-darwin-arm64`, `archon-darwin-x64`, `archon-linux-arm64`, `archon-linux-x64`, `archon-windows-x64.exe`, `archon-web.tar.gz` — all six match the glob.

## Not verified

Whether anything downstream currently depends on the duplicate line (`scripts/install.sh` checksum verification, `scripts/update-homebrew.sh`). `update-homebrew.sh` produced correct SHAs for v0.7.0 with the duplicate present — I verified all four formula SHAs against `checksums.txt` by hand and they matched — so it is at minimum tolerant of it. `install.sh` was not inspected for this.

PR #2974, opened 2026-08-29
Fix duplicate release artifact checksums

Wirasm:
## Problem and outcome

The release checksum command expands `archon-web.tar.gz` through `archon-*` and then names it again explicitly, producing two checksum rows for the same artifact. This makes the published checksum manifest redundant and potentially confusing to consumers.

- **Outcome:** Each release artifact, including `archon-web.tar.gz`, has one checksum row.
- **Invariant:** `archon-web.tar.gz` remains covered by the release checksum manifest.
- **Scope boundary:** Only the release checksum input list changed.
- **Root cause:** The command combined an inclusive glob with an explicit duplicate filename.

## Review guidance

- **Feedback requested:** Confirm the glob covers every intended release asset exactly once.
- **Start here:** `.github/workflows/release.yml:280` — this is the complete behavior change.
- **Review order:** Inspect the one-line checksum command change, then the validation evidence below.

## Solution

Remove the redundant explicit `archon-web.tar.gz` argument and retain `archon-*` as the sole checksum input pattern. The web tarball already matches that pattern, so this preserves coverage while eliminating its second manifest row.

## Behavior change

| | Before | After |
|---|---|---|
| **Checksum manifest** | `archon-web.tar.gz` appears twice. | Every matching release artifact appears once. |
| **Web tarball coverage** | Included twice. | Included once through `archon-*`. |

## Validation

- Temporary-directory reproduction with `sha256sum archon-* archon-web.tar.gz` and `sha256sum archon-*` — confirmed the former emits two web-tarball rows and the latter emits one row per representative release asset.
- `bun x prettier --check .github/workflows/release.yml` — passed.
- `bun run validate` — passed after `bun install --frozen-lockfile`; generation checks, API types, type checks, lint, formatting, installer tests, and test suites passed.
- **Not verified:** Nothing material; the command's glob expansion and repository validation directly cover the changed behavior.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -277,7 +277,7 @@ jobs:
       - name: Generate checksums
         run: |
           cd dist
-          sha256sum archon-* archon-web.tar.gz > checksums.txt
+          sha256sum archon-* > checksums.txt
           cat checksums.txt
 
       - name: Get version


Wirasm:
# Review report

## Verdict

Ready — action: `none`. The only behavioral change removes the duplicate checksum operand while `archon-*` still includes every produced release artifact, including `archon-web.tar.gz`.

## Accepted contract

The release checksum manifest must contain exactly one SHA-256 row for every release artifact, including `archon-web.tar.gz`. The accepted boundary permits only the release checksum input-list change; it preserves the `archon-*` artifact family, web-tarball coverage, `sha256sum` manifest format, and release upload configuration. Checksum consumers, artifact publication, and the hashing algorithm are out of scope.

## Reviewed head SHA

`a7f45b1f22b9a1e84e25ccac77bf6ea8617ef46a`

## Findings

### Suggestions

- **R1 — sources: [tests].** Add an optional focused workflow-contract test for checksum-row uniqueness. No current defect was found: `.github/workflows/release.yml:280` runs `sha256sum archon-*` once after all artifacts are merged at `:271-275`, and the web producer writes `archon-web.tar.gz` at `:45`. The specialist correctly established that existing CLI tests protect consumer parsing rather than writer uniqueness (`packages/cli/src/commands/serve.test.ts:39-46`; `packages/cli/src/commands/serve.ts:295-306`), but the accepted change is a one-line input-list correction and does not require expanding that boundary with a new test. A focused producer test would be useful follow-up hardening, not a blocker.

### Rejected findings

None.

## Discoveries

0 accepted discoveries. See [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/a40cb4c7c06811ceb0e39bb975876b54/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/a40cb4c7c06811ceb0e39bb975876b54/discoveries.md). Adjacent discoveries do not affect readiness.

## Review coverage

- `code`: ran; no in-scope correctness finding. Verified the artifact producers, unchanged publication list, manifest consumers, and clean diff.
- `seams`: ran; no in-scope seam finding. Verified the producer-to-manifest and manifest-to-web-fallback boundaries.
- `simplify`: ran; no finding. The delta removes the redundant representation without adding structure or state.
- `tests`: ran; produced R1, accepted as a non-blocking suggestion after contract and boundary review.
- `errors`: disabled; the delta does not alter a failure, recovery, fallback, or error-translation path.
- `docs`: disabled; the delta adds no user-facing behavior or configuration surface.

## Verification

I read the accepted scope, all required specialist reports, the architecture reference, and the changed workflow with its relevant producers and consumers. `git diff --check dev...a7f45b1f22b9a1e84e25ccac77bf6ea8617ef46a -- .github/workflows/release.yml` completed cleanly. The exact current command at `.github/workflows/release.yml:280` is the necessary falsification of the duplicate mechanism: unlike the prior two-operand form, `archon-web.tar.gz` is supplied only by the single `archon-*` expansion.

Wirasm:
Closing with its duplicate #2971: same issue (#2377), delivered twice by parallel dogfood runs, both red on stale bases. #2377 gets one fresh delivery on the new pack in the pre-release dogfood wave. The issue stays open.

PR #2975, opened 2026-08-29
fix(cli): a resumed run's completed nodes no longer report as skipped

### LINKED ISSUE #2973: archon workflow get reports completed nodes as skipped after any resume, including the engine's own durable-wait continuation
## Problem

`archon workflow get <run-id> --verbose --json` reports nodes that **ran and succeeded** as `skipped`, once the run has been resumed even once.

Because `archon-deliver` suspends at a durable `wait:` while it watches CI, and the server's continuation scan resumes it, **every delivery run hits this**. The report is wrong for the normal path, not an edge case.

## Evidence

Dogfood run `73dd151bff7bfe913a63f611106643c2` (`archon-ship`, opened PR #2971).

The run's own log records 31 nodes started and 31 completed. The CLI reports 11 completed and 40 skipped. **22 nodes the log says completed are reported as skipped**, including:

```
deliver__pr__pr            <- the node that created PR #2971
deliver__sync-pr-body      <- verified to have edited PR #2970's body on a sibling run
deliver__validate__validate
deliver__gate-green, deliver__gate-ready, deliver__gate-validated
triage__triage, deliver__classify, deliver__review-action
the entire review fan-out (mode, scope, code, seams, simplify, tests, synthesize, review-complete)
```

The persisted events for that one node are unambiguous:

```
node_started                2026-08-29 14:27:29
  … 27 tool_called / tool_completed rows …
node_completed              2026-08-29 14:29:16     <- ran, succeeded, opened the PR
node_skipped_prior_success  2026-08-29 14:53:47     <- engine's own durable-wait continuation
node_skipped_prior_success  2026-08-29 15:38:50     <- a later manual resume
```

Run-level counts: `node_skipped_prior_success` × 49, `workflow_resumed` × 1, `wait_started` / `wait_completed` × 1 each.

## Cause

`buildNodeSummaries` (`packages/cli/src/commands/workflow.ts:3417`) folds two semantically opposite events into one state:

```ts
case 'node_skipped':
case 'node_skipped_prior_success': {
  summaries.set(nodeId, { nodeId, state: 'skipped' });
  break;
}
```

Events are folded in order and the last write wins, so a later `node_skipped_prior_success` **overwrites the earlier `completed`** — erasing the success, its duration, and its output preview.

`node_skipped_prior_success` does not mean "this node was skipped". It means *"skipped re-running it, because it already succeeded"* — the opposite conclusion. The function's own doc comment says it "Processes node_started / node_completed / node_failed / node_skipped* events", and that `node_skipped*` glob is the mistake: it globs together "never ran" and "already succeeded".

`packages/core/src/db/workflow-events.ts` already models this correctly — it treats these rows as a replay of a prior success and even carries the earlier output forward (`:287`, `:391`, `:415`, `:474`). So the contract is right in core and lost in the CLI's projection.

## Why it matters more than a display nit

This is the surface an operator reads to find out what a run did. Today it will tell you a run that opened a PR skipped its PR node, and that a run which validated skipped its validation gate. For an engine whose product is governance, a run-state view that reports completed work as skipped is worse than a run that fails honestly.

It also actively misleads triage. While diagnosing these dogfood runs I flagged "a PR exists but the graph says the node never ran" as a possible bug in the run-owned PR record from #2950. It was not — the graph was fine and the report was lying. Anyone triaging a resumed run will lose the same time.

## Suggested fix

Map `node_skipped_prior_success` to the state it actually describes — `completed` — or give it a distinct state (`reused` / `prior-success`) if the difference is worth showing. Either way it must not collapse into `skipped`, and it must not clobber the duration and output captured on the original `node_completed`.

Worth a regression test with the shape above: `node_started` → `node_completed` → `node_skipped_prior_success`, asserting the node does not read as skipped. Note that `packages/cli` `*.test.ts` files escape type-check and ESLint, so the test file naming matters if that is a concern.

## Reproduce

```bash
archon workflow get 73dd151bff7bfe913a63f611106643c2 --verbose --json
sqlite3 ~/.archon/archon.db \
  "SELECT event_type, created_at FROM remote_agent_workflow_events \
   WHERE workflow_run_id='73dd151bff7bfe913a63f611106643c2' \
     AND step_name='deliver__pr__pr' ORDER BY created_at;"
```

Any run that suspends at a durable wait and resumes will show it.

Wirasm:
## Problem and outcome

`archon workflow get <run-id> --verbose` reported nodes that ran and succeeded as `skipped` once a run had been resumed even once. Because `archon-deliver` suspends at a durable `wait:` and the server's continuation scan resumes it, every delivery run hit this: on run `73dd151b`, 22 nodes the run log records as completed — including the node that opened PR #2971 and the validation gate — were reported as skipped, with their duration and output preview erased. For an engine whose product is governance, a run-state view that reports completed work as never run misleads exactly the operator trying to triage the run.

- **Outcome:** a node that ran and succeeded reads as completed in `workflow get`/`workflow status` (`--verbose` and `--verbose --json`) no matter how many times the run was resumed, and it keeps the duration and output preview captured on its original `node_completed`.
- **Invariant:** `node_skipped_prior_success` never downgrades or overwrites what a node's own terminal event already recorded. A genuine `node_skipped` still reads `skipped`.
- **Scope boundary:** no new state vocabulary — the `--json` `state` field keeps its four values, and prior success maps onto the existing `completed`. The engine's event emission is unchanged; only the CLI's projection was wrong.
- **Root cause:** `buildNodeSummaries` folded `node_skipped` and `node_skipped_prior_success` into one `skipped` case, and events are folded in order with last-write-wins. The engine re-emits `node_skipped_prior_success` once per resume pass, so each replay overwrote the earlier `node_completed` summary. The two events mean opposite things: one is "never ran", the other is "did not re-run it, because it already succeeded".

## Review guidance

- **Feedback requested:** correctness of the no-summary edge case — what a prior-success replay should report when the original `node_completed` is not in the log.
- **Start here:** `packages/cli/src/commands/workflow.ts:3426` — the split case that carries the fix.
- **Review order:** the `buildNodeSummaries` diff, then `packages/cli/src/commands/workflow-node-summaries.spec.ts`.
- **Lower-attention areas:** `outputPreviewOf` is a pure extraction of the preview expression that was already inline in the `node_completed` case, now shared by both success paths so the 200-character cap has one owner. `packages/cli/package.json` registers the new spec file in the package test batches.
- **Known risk or uncertainty:** None. `packages/core/src/db/workflow-events.ts` already treats these rows as a replay of a prior success, so this brings the CLI projection onto the contract core implements rather than inventing one.

## Solution

`node_skipped` keeps its existing behavior in its own case. `node_skipped_prior_success` gets a separate case that never overwrites an existing summary, so the original `node_completed` — with its start time, duration, and output preview — survives every resume pass.

When no summary exists for the node, the replay is the only surviving record of that success. It carries the prior output forward (`dag-executor.ts` copies `node_output` and `structured_output` into the re-emit), so it reports `completed` with that output preview and no duration; there is no start time to derive one from, and inventing a duration would be a worse lie than omitting it. This is the path the existing `VERBOSE_EVENTS_FIXTURE` `middle` node already exercised.

The function's doc comment claimed it processes "`node_skipped*`" events. That glob was the mistake in prose form, so it now names both events and says they are handled separately.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A completed node reads `skipped` (`-` icon, no duration, no output) after any resume | It reads `completed` (`✓`, original duration and output preview) |
| **Failure behavior** | A genuinely skipped node reads `skipped` | Unchanged |

## Validation

- `bun test src/commands/workflow-node-summaries.spec.ts` in `packages/cli`, run against the unfixed function first — 2 of 4 fail, with the completed node returning `{state: "skipped"}` and its `durationMs`, `startedAt`, and `outputPreview` gone. Green after the fix. This is the issue's exact event shape: `node_started` → `node_completed` → two `node_skipped_prior_success` replays, matching the real run's durable-wait continuation plus a later manual resume.
- `bun run test` in `packages/cli` — 14 batches, 0 failures — proves the `workflow get`/`workflow status` JSON and rendering paths still agree, including the strengthened `middle?.state` assertion on the shared verbose fixture.
- `bun run validate` at the repository root — exit 0 — covers `type-check`, `lint --max-warnings 0`, `format:check`, generated-artifact checks, and every package suite.
- Runtime evidence on the issue's own run, `archon workflow get 73dd151bff7bfe913a63f611106643c2 --verbose --json`, same database, dev build vs this branch:

  | | `dev` | this branch |
  |---|---|---|
  | Node states | 11 completed, 40 skipped, 2 failed | 36 completed, 15 skipped, 2 failed |
  | `deliver__pr__pr` | `{state: "skipped"}` | `{state: "completed", durationMs: 107000, outputPreview: "{\"number\":2971,\"url\":…"}` |

  The `dev` numbers reproduce the issue's reported 11/40 exactly, and the node that opened PR #2971 now reports the success, duration, and PR URL the run log recorded.
- **Not verified:** nothing material. The change is a pure projection over persisted events, and the spec drives it directly with the persisted rows from the issue's own SQL output.

## Delivery considerations

|---|---|---|
| Compatibility / migration | `--json` `state` keeps its four existing values, so no parser changes. Affected nodes flip from `skipped` to `completed`, which is the correction. | `NodeSummary` at `packages/cli/src/commands/workflow.ts:3358` is unchanged |
| Observability | The Console (default UI) already folds this correctly — its `NodeRun` status precedence is "ever completed wins" — so this closes the gap between the two surfaces rather than opening one. | `packages/web/src/experiments/console/primitives/event.ts:398` |

## Links

- Closes #2973

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow status reporting for retried nodes and nodes replayed after a prior successful run.
  * Preserved completed states and results instead of incorrectly showing these nodes as skipped.
  * Continued reporting genuinely skipped nodes correctly.
  * Standardized output previews in node summaries.

* **Tests**
  * Expanded CLI test coverage for workflow status, integrations, utilities, adapters, and related scenarios.


Wirasm:
## Ready to merge

`buildNodeSummaries` in `packages/cli/src/commands/workflow.ts` now splits `node_skipped` (unchanged) from `node_skipped_prior_success` (never overwrites an existing summary; synthesizes `completed` with the copied-forward output when no summary exists yet). All three reviewers traced the split-case logic, its edge cases (repeated replays, retry-after-cache-invalidation, no-prior-summary), and every consumer of `NodeSummary.state`, and found the fix correct, minimally scoped, and consistent with the design the operator already settled (no new state vocabulary; genuine skips unchanged). The one operator-flagged risk — whether the new `.spec.ts` file registered in `packages/cli/package.json` actually runs, given `scripts/test-inventory.test.ts` only tracks `*.test.ts` — is independently confirmed below: it runs, through two separate paths, not just by trusting the registration line.

The seam-analysis pass found the same defect class (folding "never ran" and "ran and succeeded, this resume declined to re-run it" into one status) unfixed in two other consumers of the same event pair: Slack's live progress message (currently-supported adapter) and the deprecated `/legacy` web UI. Neither is in this PR's diff or stated scope, so neither blocks it. The Slack one is real and reachable on a supported surface, so I filed it as a tracked follow-up; the legacy-UI one is declined against the project's already-stated Console-is-default-UI direction.

**Validation:** cli package tests, type-check, and targeted lint all pass on the reviewed head; the new spec's execution was independently proven, not just asserted from the registration line. Full repo-wide `bun run validate` was not re-run by this review (see validation table).

**Resolved:** 0 · **Tracked follow-ups:** [#2978](https://github.com/coleam00/Archon/issues/2978)

### Findings

|---|---|---|---|
| `R1` | Important | Slack's live progress message folds a resumed node's prior-success replay into `skipped`, same defect class as #2973, on a supported adapter | TRACKED FOLLOW-UP |
| `R2` | Important | The deprecated `/legacy` web UI's DB-row projection repeats the identical fold, corrupting `WorkflowProgressCard`/`MessageList` node status | DECLINED |
| `R3` | Suggestion | The new regression test never varies `node_output` between the original `node_completed` and the replay events, so it can't distinguish "never overwrite" from "overwrite with an accidentally-equal value" | OPEN |

<details>
<summary><code>R1</code> — Slack bridge folds prior-success skip into `skipped`</summary>

**Impact:** A workflow that pauses at a Slack-fed approval gate after a node completes, then resumes, shows that node as skipped in the final/live Slack progress message — even though it ran and produced output. Same symptom as #2973, one layer up in the live in-process event stream instead of the persisted event log.

**Evidence:** `packages/workflows/src/event-emitter.ts:112-118` (`NodeSkippedEvent.type` has no discriminant matching the persisted `node_skipped`/`node_skipped_prior_success` split — only a `reason` string); `packages/workflows/src/dag-executor.ts:10091-10097` (emit site, both cases emit `type: 'node_skipped'`); `packages/adapters/src/chat/slack/workflow-bridge.ts:149-152` (`reason` dropped entirely), `:325-343` (`upsertNode`, plain last-write-wins `Map.set`); `packages/adapters/src/chat/slack/blocks.ts:198-212` (renders the corrupted map into the Slack message).

**Required outcome:** Give `NodeSkippedEvent` a discriminant (or split it) matching the persisted event distinction, and have the Slack bridge's `upsertNode` treat a prior-success replay as "leave the existing completed entry alone," mirroring what `#2975` now does for `buildNodeSummaries`.

**Disposition:** TRACKED FOLLOW-UP — filed as [coleam00/Archon#2978](https://github.com/coleam00/Archon/issues/2978). Not required by #2975's stated outcome or invariant (that PR's own scope boundary is "only the CLI's projection was wrong"; the Slack bridge is a separate, untouched file with its own independent fold). Real and reachable on a currently-supported adapter, so it should not go unrecorded.

</details>

<details>
<summary><code>R2</code> — Legacy web UI repeats the identical fold</summary>

**Impact:** A resumed run shows a previously-completed node as skipped in the `/legacy` chat UI's progress card and per-node icon (`WorkflowProgressCard.tsx:54-55,191`, `MessageList.tsx:146-152`, both filtering on `status === 'completed'` against a store that was silently overwritten to `'skipped'`).

**Evidence:** `packages/server/src/adapters/web/workflow-bridge.ts:233-238` (`ROW_NODE_STATUS` maps both `node_skipped` and `node_skipped_prior_success` to `'skipped'` — the exact fold `buildNodeSummaries` had before this PR); `packages/web/src/stores/workflow-store.ts:249-277` (`handleDagNode`, last-write-wins overwrite with no re-fetch, unlike `handleWorkflowStatus`'s `invalidateWorkflowQueries()` a few lines above); `packages/web/src/App.tsx:92-95` (this surface is explicitly re-rooted under `/legacy` for a deprecation window, "removed from the codebase once the console has proven itself").

**Required outcome:** N/A — declined.

**Disposition:** DECLINED. `/legacy` is on an already-stated deprecation path (confirmed directly in `App.tsx`, consistent with the project's Console-is-default-UI direction); investing in a fix for code slated for removal is not required by this PR and would be exactly the kind of unrelated, non-required work `AGENTS.md` says to keep out of a focused change. Revisit only if `/legacy`'s removal timeline slips materially or a user hits this concretely before then.

</details>

<details>
<summary><code>R3</code> — Regression test doesn't vary the replayed output</summary>

**Impact:** None currently — the test correctly catches the #2973 regression (state/`startedAt`/`durationMs` loss). It just can't separately prove "never overwrite" from "overwrite with a value that happens to match."

**Evidence:** `packages/cli/src/commands/workflow-node-summaries.spec.ts:75-112` — the original `node_completed` and both `node_skipped_prior_success` replay events all carry the identical `node_output` string.

**Required outcome:** Optional: vary `node_output` on the replay events (or the duration derivation) so the assertion also proves the no-overwrite guard specifically, not just the end-to-end outcome. Not required for merge.

**Disposition:** OPEN (non-blocking suggestion) — left for the author's discretion.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R3; otherwise no additional findings — fix traced against pre-fix `dev` code, all three `buildNodeSummaries` consumers confirmed routed through the one function |
| `seams` | R1, R2; `NodeSummary.state`/`startedAt`/`durationMs` type, core's `getDagResumeSnapshot`, and the Console's `foldNodeRuns` precedence fold examined and confirmed clean (see below) |
| `simplify` | No blocking findings; `outputPreviewOf` extraction and the `.spec.ts` split judged proportionate, not premature or half-done |

One item needed independent reconciliation: the simplify reviewer flagged `packages/web/src/experiments/console/primitives/event.ts:142` (`NODE_TRANSITION_BY_EVENT` mapping `node_skipped_prior_success` → `'skipped'`) as a possible sibling of the bug this PR fixes. The seam reviewer had already examined this same file and concluded it was safe because `foldNodeRuns` doesn't fold with last-write-wins — it collects every transition per node first, then applies fixed `completed > failed > skipped > running` precedence (`event.ts:400-454`). I read `foldNodeRuns` directly to settle the disagreement: it does collect all transitions before deciding (`event.ts:415-436`), and `completed !== null` is checked before `skipped !== null` regardless of event order, so a node that ever completed stays `completed` even after a later prior-success replay collapses to a `skipped` transition. Confirmed NOT A FINDING.

#### Validation

|---|---|---|
| `bun test src/commands/workflow-node-summaries.spec.ts` (packages/cli) | PASS | 4 pass, 0 fail, 4 expect() calls |
| `bun run test` (packages/cli, full package script — includes the newly-registered spec) | PASS | All batches pass; the new spec's batch reports 4 pass within the run, proving the `scripts.test` registration actually executes it, both directly and via the package's own aggregate script |
| `bun test src/commands/workflow.test.ts` (packages/cli) | PASS | 333 pass, 0 fail (run by `code` reviewer) |
| `bun x tsc --noEmit` (packages/cli) | PASS | Clean, no errors |
| `bun x eslint packages/cli/src/commands/workflow.ts packages/cli/src/commands/workflow-node-summaries.spec.ts --max-warnings 0` | PASS | 0 errors, 0 warnings |
| `bun run lint` (repo root) | PASS | Clean, repo-wide (run by `code` reviewer) |
| `bun test ./scripts/` (repo root, includes `scripts/test-inventory.test.ts`) | PASS | 147 pass, 0 fail |
| `bun run validate` (repo root, full) | NOT RUN | Author reports exit 0 in the PR body; not independently re-run given the change's narrow scope (one package, no schema/generated-artifact changes) and the targeted checks above already covering type-check, lint, and the affected package's full test suite |

On the operator's specific question — does the `.spec.ts` registration actually cause the file to run, given the repo-level inventory guard only tracks `*.test.ts` — the answer is yes, confirmed two ways, not assumed from the registration line: the file passes standalone, and it executes as part of `packages/cli`'s own `bun run test` script (which the root's `bun --filter '*' --parallel test` invokes per package). `scripts/test-inventory.test.ts` genuinely does not scan `.spec.ts` files at all (`TEST_FILE_PATTERN = /\.test\.tsx?$/`, and it only reads `scripts.test`), so it would not have caught a missing registration here — but this PR's registration is not missing, and its effect was verified directly rather than inferred from the guard's silence.

</details>

PR #2977, opened 2026-08-29
fix(sdlc): a correction round with nothing left to change no longer kills the delivery

### LINKED ISSUE #2960: fix(sdlc): the corrections loop has no exit for an honest no-change verdict on non-introduced red
## Problem

Run `1c225d25` (deliver on #2909, 28 Aug) wedged in its corrections loop and failed after three iterations, even though its correction agent behaved exactly as the #2939 contract asks. The agent's structured output:

> `green: false, red_cause: "inherited"` — "The R5 correction and focused regression test pass. The full lint check is red only for a pre-existing raw recursive cleanup at bundled-defaults.test.ts:728, confirmed present at the reviewed starting commit."

That is the honest verdict: the remaining red is inherited, there is nothing to change. But `fix__assert-changed` requires the iteration to have produced a commit or working-tree change unconditionally:

> `Script node 'fix__assert-changed' failed [exit 1]: implement produced neither a commit nor a working-tree change outside .archon/, and the branch carries no verified work ahead of the base (green=false).`

So the loop demands work, the agent correctly refuses to invent work, and the run dies — the exact outcome #2939 was built to avoid, reproduced one node earlier in the chain. The declared-red-cause routes now exist at gate-green, gate-correction-green, and gate-validated, but the change-assertion in front of them still enforces "every iteration commits something."

## Fix boundary

`fix__assert-changed` (and any sibling assert-changed on correction paths) must accept the no-change case when the iteration's structured output declares `red_cause: inherited | environment` with evidence — the same evidence bar the gates enforce (#2942's empty-summary refusal). An introduced red with no change remains a failure. A fixture pins the honest-decline route end to end; mutation check: reverting the tolerance re-wedges the fixture.

## Note

This is the first live finding from the #2939 routes — the field itself worked (the agent used it correctly, with the named check and why-not-caused evidence). The gap is one assertion that predates the vocabulary.

Wirasm:
## Problem and outcome

The sdlc corrections loop demanded a commit or a working-tree change from every iteration. A correction round whose remaining red is a break in the base, or repository configuration the run has no permission to change, has no edit left to make — and when the agent reported that honestly, `fix__assert-changed` killed the run for telling the truth. Two independent dogfood runs ended this way, the second discarding two earlier rounds of completed, validated work along with the delivery.

- **Outcome:** a correction round that changed nothing passes its guard when it declares its remaining red `inherited` or `environment` with evidence. The loop continues to its recheck, and the run reaches the gates and the CI wait instead of dying at `fix__assert-changed`.
- **Invariant:** nothing to show is still a failure whenever the red could be the change's own doing. `introduced`, a red with no declared cause, and a declared cause with an empty summary all fail exactly as before — the same evidence bar the green gates already hold.
- **Scope boundary:** the gates are untouched. `gate-green` still owns what a non-introduced red claim is worth and still records it; this only stops one node earlier from requiring a change that cannot exist. No new workflow input distinguishes a first implementation from a correction.
- **Root cause:** `assert-changed` predates the `red_cause` vocabulary. It saw only `green`, so every no-change iteration looked identical to a decline, whether the loop had blamed itself or named a break it could not have caused.

## Review guidance

- **Feedback requested:** correctness of the new pass condition, and whether the evidence bar belongs here as well as at the gates.
- **Start here:** `.archon/workflows/sdlc/implement/scripts/assert-changed.py:98` — the one new pass route, and the only behavior change in the PR.
- **Review order:** the script, then the `with:` bindings at `.archon/workflows/sdlc/implement/archon-implement.yaml:79-82` that feed it, then the fixtures.
- **Lower-attention areas:** `packages/workflows/src/defaults/bundled-defaults.generated.ts` is regenerated by `bun run generate:bundled`; `bun run check:bundled` reports it up to date.
- **Known risk or uncertainty:** None. The tolerance is uniform, so a *first* implementation that produces nothing and declares an inherited red now passes this guard too — but that case cannot reach a pull request. GitHub refuses to create one when there are no commits between base and head, failing deterministically with `GraphQL: No commits between <base> and <branch> (createPullRequest)` (verified empirically during review). `pr.md`'s instruction to confirm the branch has commits ahead of its base is the polite layer above that hard stop, not the only thing holding it.

## Solution

`assert-changed` gains a fourth way to pass, after the three it already had: no working-tree change, no new commit, not green, and the remaining red declared `inherited` or `environment` with a non-empty summary behind it. `archon-implement` binds `red_cause` and `summary` alongside `green`, so the guard sees the whole verdict rather than one field of it — the same three values `gate-green` already receives, through the same `with:` mechanism, with no artifact bridge.

The tolerance lives in the script rather than in a new workflow input, so the implementation and every correction round get one rule. Distinguishing them would mean threading a flag through `include: archon-implement` to make the guard behave differently in two places that ask the same question.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A corrections iteration that made no commit and no working-tree change failed at `fix__assert-changed` whatever its reason, killing the run and discarding the rounds that had already landed. | The same iteration passes when it names a non-introduced cause with evidence, and the loop runs on to `gate-correction-green` and the continuation review. |
| **Failure behavior** | One message for every no-change iteration: the branch carries no verified work ahead of the base. | Introduced red, unexplained red, and a cause with an empty summary still fail; the message now reports the declared cause alongside `green` and says which causes are tolerable. |

### Changed seams

|---|---|---|
| `archon-implement.implement → assert-changed` | The guard consumes `red_cause` and `summary` in addition to `green`. | `.archon/workflows/sdlc/implement/archon-implement.yaml:79-82`; `sdlc/deliver/fixtures/correction-honest-decline.stubs.yaml` runs the real script on those bindings |

## Validation

- `bun run cli workflow test sdlc --json` — 43 passed, 0 failed — proves the new route end to end and leaves every existing fixture green.
- Mutation check, guard seen red. Disabling the new route makes `correction-honest-decline` die at `fix__assert-changed` (`expected exactly one failed trace entry on 'fix__assert-changed', got fix__assert-changed, corrections`) and fails `decline-inherited`. Dropping the `red_cause` test instead lets `decline-introduced` and `decline` complete; dropping the summary test lets `decline-no-evidence` complete. Each branch of the condition is pinned by a fixture that catches its removal.
- `bun --filter @archon/workflows test` — every isolated batch 0 fail — proves the regenerated bundled defaults and the fixture runner still hold.
- `bun run validate` — green.
- **Not verified:** no live delivery run reproduces the original wedge. The fixture executes the real `record-start` and `assert-changed` against a clean scratch worktree, which is the exact git state the failing runs were in, so the guard's decision is observed rather than simulated; the surrounding AI nodes are stubbed, as they are in every fixture in this pack.

## Links

- Closes #2960


Wirasm:
## Ready to merge

The corrections loop previously killed a delivery whenever a round changed nothing, even when the remaining red was an honest, non-introduced break (already red in the base, or environmental) rather than a lazy decline. This PR adds one new pass route to the deterministic guard `assert-changed.py`: not green, `red_cause` declared `inherited` or `environment`, and a non-empty evidence summary now passes instead of failing the run. `introduced` red, an undeclared cause, and a declared cause with an empty summary still fail exactly as before. The tolerance is uniform across a first implementation and every correction round, as the operator intended — there is no separate path or flag distinguishing them, and no reviewer proposed adding one.

I independently reproduced every load-bearing claim in the PR body rather than trusting the self-report: I ran the real fixture suite (43/43 passing, matching the PR's number), then mutated `assert-changed.py` three separate ways and re-ran the suite each time to confirm each conjunct of the new condition is actually pinned by a fixture that catches its removal — dropping the whole route breaks `correction-honest-decline` and `decline-inherited`; dropping the `red_cause in PASSES_RED` check lets `decline-introduced` and `decline` wrongly pass; dropping the `summary` check lets `decline-no-evidence` wrongly pass. I also confirmed `red_cause`/`summary` were pre-existing fields on `implement`'s output (not new schema surface), that the untouched `gate-green.py` gate already uses the identical `PASSES_RED` vocabulary, that `bun run check:bundled` reports the regenerated file up to date, and that no changed file carries a repository issue number or run id.

On the PR's own flagged "Known risk" — that a *first* implementation which changes nothing and declares `inherited` red is now only guarded by `pr.md`'s prose instruction to confirm commits ahead of base, since `archon-pr` is a single AI node with no deterministic check — I tested GitHub's actual behavior empirically rather than relying on the prompt alone: `gh pr create` against a branch with zero commits ahead of its base fails deterministically with `GraphQL: No commits between <base> and <head> (createPullRequest)` and a non-zero exit code. That closes the residual risk independent of the AI node's judgment; a genuinely zero-diff branch cannot become an open PR no matter what the node does. No new deterministic guard is warranted here — a check that only re-asserts what GitHub's own API already refuses would be pure defense in depth with no reachable failure it closes. The one artifact worth fixing is documentation, not code: the PR body's "Known risk" section should say the forge is the deterministic backstop rather than reading as an open gap (`R1`, non-blocking).

**Validation:** all green — real fixture suite, bundled-defaults regeneration, and full `bun run validate` all pass; three independent reviewer scopes returned no Critical or Important finding.

### Findings

|---|---|---|---|
| `R1` | Suggestion | PR body's "Known risk" section undersells the actual protection: GitHub itself refuses `gh pr create` with zero commits between head and base, closing the gap the section describes as prompt-only | OPEN |

<details>
<summary><code>R1</code> — PR body's "Known risk" section doesn't credit the forge's own deterministic refusal</summary>

**Impact:** No code defect. The PR body's self-assessment reads as if a first-implementation decline with zero diff is protected only by `pr.md`'s prose ("Confirm the branch is not the base and has commits ahead of it"), inviting a future reader to add an unnecessary guard. It is actually double-covered: even if the AI node misjudges and proceeds anyway, `gh pr create` in `pr.md` step 4 hard-fails on a genuinely zero-diff branch.

**Evidence:** Empirical reproduction — pushed a branch with 0 commits ahead of its base to a disposable private repo and ran `gh pr create --repo Wirasm/zero-commit-pr-probe --base development --head zero-commit-branch ...`; it failed with `pull request create failed: GraphQL: No commits between development and zero-commit-branch (createPullRequest)`, exit code 1. `.archon/workflows/sdlc/pr/commands/pr.md:32-36` requires reading the created PR back by explicit number before the node can report success, so a failed `gh pr create` cannot be silently reported as done. `.archon/workflows/sdlc/deliver/archon-deliver.yaml:69-73` shows `pr` (which includes `archon-pr`) depends only on `gate-green`, which already applies the identical `inherited`/`environment` tolerance and is unchanged by this PR.

**Required outcome:** Update the PR body's "Known risk" paragraph (or a follow-up comment) to state that GitHub's own PR-creation call is the deterministic backstop for the zero-commit case, rather than leaving it framed as an unaddressed gap. No code change required — do not add a redundant deterministic guard duplicating GitHub's own refusal; per the project's no-defense-in-depth bar, a check that only re-proves what the forge already enforces is a finding against adding it.

**Found by:** `prp-core:code-reviewer` (independently reasoned the same conclusion from `pr.md` and the delivery graph); coordinator (empirical reproduction against a live GitHub API)

**Disposition:** OPEN — cosmetic, author's call whether to update the PR description before or after merge. Does not block.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 (independently derived); no Critical/Important |
| `seams` | No finding — `PASSES_RED` duplication between `assert-changed.py` and `gate-green.py` examined and cleared (no shared-import path exists; both copies verified in agreement; an end-to-end fixture in this PR already exercises both together) |
| `simplify` | No finding — new code is the minimal expression of the stated rule; all five new/changed fixtures independently earn their place (verified by running the harness, not by static read) |

#### Validation

|---|---|---|
| `bun run cli workflow test sdlc --json` (baseline) | PASS | 43 passed, 0 failed |
| `assert-changed.py` route-4 disabled (mutation) | PASS (caught) | `correction-honest-decline` and `decline-inherited` fail as expected — 41 passed, 2 failed |
| `assert-changed.py` `red_cause in PASSES_RED` conjunct removed (mutation) | PASS (caught) | `decline-introduced` and `decline` wrongly complete — 41 passed, 2 failed |
| `assert-changed.py` `summary` conjunct removed (mutation) | PASS (caught) | `decline-no-evidence` wrongly completes — 42 passed, 1 failed |
| `bun run check:bundled` | PASS | "bundled-defaults.generated.ts is up to date (55 commands, 32 workflows, 11 scripts)" |
| `bun run validate` | PASS | exit 0; 147 pass / 0 fail in the workflows package plus every other package's suite; only expected negative-path error logs (e.g. deliberate invalid-JSON and permission-denied test fixtures) appear in output |
| `gh pr create` on a zero-commit branch (empirical, disposable scratch repo) | FAILED (as expected — proves the forge backstop) | `GraphQL: No commits between development and zero-commit-branch (createPullRequest)`, exit 1 |
| Pack comment scan for issue/run-id leakage | PASS | no matches for `#[0-9]{2,}` / `run[_-]?id` / `issue` across changed workflow and script files |

</details>

PR #2979, opened 2026-08-29
fix(workflows): dry-run tolerates a missing stub on an all_done loop node

### LINKED ISSUE #2966: Dry-run stub tolerance does not cover a bare loop: node with trigger_rule: all_done
## Problem

PR https://github.com/coleam00/Archon/pull/2958 (`fix(workflows): dry-run tolerates a missing stub on an all_done join`) makes a dry run stop failing when a `trigger_rule: all_done` join node has no stub. The reasoning holds: such a node runs whatever its upstream did, so a real run reaches it regardless of fixture data, and a missing stub there is an authoring gap rather than evidence the run would fail.

The tolerance is applied in `simulateNode` at the node gate, keyed on `node.trigger_rule === 'all_done'`. A bare `loop:` node carrying the same trigger rule does not go through that path, so it still fails on a missing stub.

## Why it was scoped out

The author disclosed the boundary in the PR body and explicitly asked for a call on it. All three review agents agreed it is justified: a loop needs iteration-aware handling — which iteration's stub, and what a generated placeholder means across iterations — rather than the single node gate the fix uses. Widening C2 to cover it would have meant designing that inside a correction round.

## Suggested approach

Decide what a tolerated placeholder means per iteration before writing code. The existing `generatedStubFor` already special-cases loops (`loop.until`, `loop.until_field`), so there is a starting point, but the completion-channel interaction is the part that needs thought: a generated placeholder must not accidentally satisfy `until`/`until_field` and end the loop.

Note the related constraint documented in C2: tolerance is only claimed once the placeholder is actually generated. A loop node whose `output_format` cannot produce one must keep failing, not become silently tolerated.

## Source

Disclosed in PR https://github.com/coleam00/Archon/pull/2958's body; confirmed as a real gap by its review.

Wirasm:
## Problem and outcome

PR #2958 stopped a dry run from failing when a `trigger_rule: all_done` join has no stub: such a node runs whatever its upstream did, so a real run reaches it regardless of fixture data, and a missing stub there is an authoring gap rather than evidence the run would fail. That tolerance was applied at `simulateNode`'s single-node stub gate, which a bare `loop:` node never reaches — so an author who put the same trigger rule on a loop still got a red dry run and a red `workflow test` fixture for a node nothing was actually blocked on.

- **Outcome:** A bare `loop:` node carrying `trigger_rule: all_done` and no stub now simulates with a generated placeholder and lets the run continue, instead of failing with `Missing reachable stub for node '<id>'`.
- **Invariant:** Tolerance is claimed only once a placeholder exists **and** ends the loop. A tolerated node is always in `missingStubs` too, so the gap stays reported; only nodes in `toleratedMissingStubs` are filtered out of the blocking set by `checkFixture` and by the CLI's failure message.
- **Scope boundary:** `loop_group` is unchanged — the group node collects no stub (`collectsStub` excludes it) and its body nodes already got the tolerance in #2958, covered by that PR's own test. `fixture-runner.ts` and `packages/cli/src/commands/workflow.ts` are unchanged: both already filter `missingStubs` by `toleratedMissingStubs` with no node-type branch.
- **Root cause:** `simulateNode` dispatches `isLoopNode(node)` to `simulateLoop` at `dry-run.ts:1018`, well before the tolerance branch at `dry-run.ts:1156`. `simulateLoop` resolves its own stub inside the iteration loop and had an unconditional `recordFailed` on `stub === undefined`.

## Review guidance

- **Feedback requested:** Correctness of the design call below — I implemented the opposite of what the issue suggested, and want that checked rather than assumed.
- **Start here:** `packages/workflows/src/dry-run.ts:850` — the tolerance branch in `simulateLoop`, including the completion guard that decides whether tolerance may be claimed at all.
- **Review order:** `dry-run.ts` (the fix), then the three tests in `dry-run.test.ts`, then the fixture-runner test that proves the end-to-end `workflow test` path.
- **Lower-attention areas:** The `TOLERATED_STUB_REASON` extraction at `dry-run.ts:722` is a pure move of the string literal #2958 inlined at the join gate — no wording change.
- **Known risk or uncertainty:** The issue asked for the placeholder **not** to satisfy `until`/`until_field`. I concluded that is backwards; the reasoning is in "Design call" below and is the thing most worth disagreeing with.

## Solution

`simulateLoop`'s missing-stub branch now splits three ways instead of always failing:

1. Not `all_done` → the existing failure, unchanged.
2. `all_done`, and the generated placeholder ends the loop → hydrate with it, add the node to `missingStubs` **and** `toleratedMissingStubs`, and trace it with a reason that names both the tolerance and the channel that ended the loop.
3. `all_done`, but the placeholder ends no iteration → fail loudly, and stay out of `toleratedMissingStubs`.

`generatedStubFor` throwing (an `$async` `output_format`, an uncompilable schema) is already caught by `simulateNode`'s surrounding `try`, so that case fails the node and is never marked tolerated — the same ordering #2958 established for joins.

### Design call: the placeholder ends the loop, deliberately

Issue #2966 says "a generated placeholder must not accidentally satisfy `until`/`until_field` and end the loop." This PR does the opposite, for three reasons:

1. **It is not accidental.** `generatedStubFor` special-cases loops precisely so the placeholder completes them: with no `output_format` it returns `node.loop.until` verbatim (`dry-run.ts:158`), and with one it sets `until_field: true` (`dry-run.ts:173`). `stubSatisfiesNode` then asserts that property for the scaffold path (`dry-run.ts:204`). `--default-stubs` and `--stubs-init` already depend on this.
2. **The alternative is another false failure.** A placeholder that did not end the loop would burn `max_iterations` and fail the run with `Loop exceeded max iterations` — exactly the class of false red #2958 exists to remove.
3. **Ending on iteration 1 overclaims nothing.** A loop with a real fixture stub also ends on iteration 1, by the same `loopIterationCompletes` rule. Iteration count in a dry run was never evidence about a real run. The trace reason names the tolerance, and the gap is still reported.

The genuine hazard in that direction is the inverse, and it is what the guard enforces. One loop shape has no placeholder that can end it: an `output_format` whose only completion channel is a prose `until:`, because the generated JSON never carries the sentinel and there is no `until_field` to set or `until_bash` to assume. That node keeps failing — the loop analogue of the C2 constraint #2958 documented. The rest of the file already agrees: `createDryRunStubScaffold` refuses that same shape today, with `Cannot generate dry-run scaffold: nodes sharing stub key 'refine' require incompatible values`.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A `loop:` node with `trigger_rule: all_done` and no stub fails the dry run and its `workflow test` fixture | It completes on a generated placeholder; the fixture passes and reports the gap as tolerated |
| **Failure behavior** | Always `Missing reachable stub for node '<id>'` | Unchanged without `all_done`. With `all_done` but no placeholder that can end the loop: `Missing reachable stub for node '<id>' — trigger_rule: all_done cannot tolerate it: a generated placeholder would run to max_iterations without completion signal 'DONE'` |

### Changed seams

|---|---|---|
| `simulateLoop → DryRunResult.toleratedMissingStubs` | A loop node can now enter the tolerated set, which `checkFixture` and the CLI failure message already filter on | `packages/workflows/src/dry-run.ts:884`; `fixture-runner.test.ts:383` proves the end-to-end fixture path |

## Validation

- Repro seen red first: `bun test src/dry-run.test.ts -t "loop node"` in `packages/workflows` — `expect(result.outcome).toBe('completed')` → `Expected: "completed" / Received: "failed"`. Proves the gap the issue reports.
- Guard seen red: with the completion check replaced by `if (false)`, the boundary test reported `Expected: "failed" / Received: "completed"` — without it, the untolerable loop is silently reported completed **and** tolerated, which would let `checkFixture` filter away a real blocker. Proves the guard tests what it claims.
- `bun test src/dry-run.test.ts` — 73 pass, 0 fail — proves tolerance on all three completion channels (`until`, `until_field`, `until_bash`), the untouched non-`all_done` failure, and the untolerable shape.
- `bun test src/fixture-runner.test.ts` — 43 pass, 0 fail — proves `workflow test` passes on an unstubbed `all_done` loop and lists it in `toleratedMissingStubs`.
- `bun run test` in `packages/workflows` (the package script, which preserves mock isolation) — no failures across all isolated invocations.
- `bun run validate` at the repo root — exit code 0.
- **Not verified:** Nothing material. Every behavior this PR changes lives in `simulateLoop` and is exercised by a direct assertion on `dryRunWorkflow`'s output plus the fixture path a user actually runs.

## Links

- Closes #2966
- Related #2958

## Summary by CodeRabbit

* **Bug Fixes**
  * Dry runs now tolerate missing stubs for `all_done` loop nodes when a completion-capable placeholder can be generated.
  * Missing loop stubs are reported with the relevant completion channel and tolerance details.
  * Loops that cannot complete with a generated placeholder still fail as expected.

* **Documentation**
  * Clarified CLI behavior for `--default-stubs`, including supported loop completion rules.


Wirasm:
## Ready to merge

`simulateLoop` never reached the `trigger_rule: all_done` stub tolerance #2958 added to `simulateNode`'s single-node gate, because a bare `loop:` node resolves its own stub inside its iteration loop instead of going through that gate. This PR closes that gap: on a missing stub, an `all_done` loop node now builds the same generated placeholder `--default-stubs` already uses, and tolerates the gap only if that placeholder can be shown to end the loop through the loop's own completion channel (`until`, `until_field`, or `until_bash`); otherwise it still fails loudly, naming the unmet channel. All three reviewers (code, seams, simplify) returned zero findings, and two of the three independently traced the exact seam the operator flagged as highest-risk — tolerance-ordering safety and predicate drift — to the same conclusion I reached working the diff directly.

**Validation:** type-check, lint, and all directly relevant tests pass (see coverage table). The completion-gate assertion was independently reproduced by mutation (see Design-call ruling).

### Findings

No findings.

### Design-call ruling

The PR body flags one deliberate deviation from the issue text as the thing most worth checking: issue #2966 said "a generated placeholder must not accidentally satisfy `until`/`until_field` and end the loop," and this PR does the opposite — it makes the placeholder end the loop on purpose, and treats a placeholder that *can't* end the loop as the failure case instead.

The argument holds, verified against source rather than the PR's own claims:

- **`generatedStubFor` already special-cases loops.** Read directly: with no `output_format` it returns `node.loop.until` verbatim (`dry-run.ts:158`); with one, it forces `until_field: true` onto the generated object (`dry-run.ts:167-173`). `stubSatisfiesNode` (`dry-run.ts:198-207`) asserts `loopIterationCompletes(...).kind !== 'incomplete'` for every loop node in the scaffold path, and `createDryRunStubScaffold` (backing both `--stubs-init` and, via `stubFor` at `dry-run.ts:730-732`, `--default-stubs`) already depends on that placeholder ending the loop. This PR's loop-tolerance branch reuses that same existing machinery rather than inventing new behavior — it isn't asserting a new precedent, it's applying an existing one to a code path that had been bypassing it.
- **The alternative is a different false red.** Traced the control flow: `stubFor(node, ctx)` inside `simulateLoop`'s per-iteration loop (`dry-run.ts:847`) is deterministic on `ctx.stubs`/`ctx.defaultStubs`, not on iteration state — a missing stub stays missing every iteration. A placeholder that didn't end the loop would hit the same missing-stub branch on iteration 2, 3, ... up to `max_iterations`, then fail with `Loop exceeded max iterations` (`dry-run.ts:916-923`) — exactly the class of false red #2958 exists to remove, just relocated.
- **Ending on iteration 1 overclaims nothing.** The ordinary hydrated-stub path (`dry-run.ts:898-914`) also returns on the first iteration whose output completes the loop. A tolerated node is never claiming more iterations happened than a real fixture stub would produce.

The literal issue text reads as a proxy for the real invariant — don't let a placeholder *coincidentally* end a loop nobody verified — not a requirement that the placeholder must never end it. The implementation satisfies the real invariant more precisely than the literal instruction would have: the "cannot tolerate" branch (`dry-run.ts:869-883`, with its own dedicated test) proves that a loop whose only completion channel a placeholder can't reach (an `output_format` with a prose-only `until:`) still fails loudly and stays out of `toleratedMissingStubs`. Nothing is tolerated by accident; tolerance is only claimed once `loopIterationCompletes` proves the placeholder actually carries the node.

I independently reproduced the "guard seen red" claim rather than taking it from the PR body: replacing `if (toleratedCompletion.kind === 'incomplete')` with `if (false)` in the worktree and running `bun test src/dry-run.test.ts -t "does not call a loop gap tolerated"` fails at the first assertion — `expect(result.outcome).toBe('failed')` — with `Received: "completed"`, and code inspection confirms execution then falls straight through to `ctx.toleratedMissingStubs.add(node.id)`, so the same mutation would also have flipped the `toleratedMissingStubs` assertion had the test reached it. The guard is proven, not assumed. The mutation was reverted before any further work (clean `git diff` afterward).

Also verified directly: `simulateLoopGroup` and `packages/workflows/src/fixture-runner.ts` are untouched by the diff (confirmed via `git diff --stat`) — the group node never calls `stubFor` itself (excluded from `collectsStub`; its body nodes already went through `simulateNode`'s gate under #2958), and both `fixture-runner.ts:606-607` and the CLI's `workflow.ts:1860-1861` filter `missingStubs` by `toleratedMissingStubs` generically, with no node-type branch, so the new loop-tolerated members are picked up automatically. The mirrored predicate in `simulateLoop` (`dry-run.ts:850`, `!== 'all_done'`) and `simulateNode` (`dry-run.ts:1156`, `=== 'all_done'`) both compare the same closed `TriggerRule` enum literal — no broader condition, no drift.

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | No findings |
| seams | No findings |
| simplify | No findings |

#### Validation

|---|---|---|
| `cd packages/workflows && bun test src/dry-run.test.ts` | PASS | 73 pass, 0 fail |
| `cd packages/workflows && bun test src/fixture-runner.test.ts` | PASS | 43 pass, 0 fail |
| `cd packages/workflows && bun x tsc --noEmit` | PASS | clean |
| `cd packages/cli && bun test src/commands/workflow.test.ts` | PASS | 334 pass, 0 fail |
| `bun run scripts/lint.ts` (repo root) | PASS | clean across all linted packages |
| Mutation: `if (toleratedCompletion.kind === 'incomplete')` → `if (false)`, then `bun test src/dry-run.test.ts -t "does not call a loop gap tolerated"` | FAIL (expected) | `Expected: "failed" / Received: "completed"` — proves the completion guard is load-bearing; reverted, no diff left behind |

</details>

PR #2950, opened 2026-08-28
feat(workflows): verify run-owned PR actions

Wirasm:
## Problem and outcome

Later SDLC delivery steps rediscovered a pull request from ambient branch state. In a fork or resumed run, that can target the wrong PR or branch for review, body updates, correction pushes, and the ready flip.

- **Outcome:** The run records the PR identity when it is created and later public actions use that recorded number and head branch.
- **Invariant:** A public action proceeds only when the checked-out branch and GitHub PR head match the run-owned PR record.
- **Scope boundary:** This changes bundled workflow ownership and action evidence only; it does not add engine-wide provider tool-output capture.
- **Root cause:** Later nodes used ambient branch-to-PR resolution instead of durable run-owned PR data.

## Review guidance

- **Feedback requested:** Check that the structured PR record is sufficient and is consistently used for every later public action.
- **Start here:** `.archon/workflows/sdlc/pr/archon-pr.yaml` — defines the persisted PR record consumed by delivery.
- **Review order:** `archon-pr.yaml` and `pr.md`; `archon-deliver.yaml`; `sync-pr-body.md` and `review-synthesize.md`; fixture and generated-default updates.
- **Lower-attention areas:** The fixture changes and `bundled-defaults.generated.ts` mechanically carry the new record shape; `bundled-defaults.test.ts` asserts the ownership path.
- **Known risk or uncertainty:** None.

## Solution

`archon-pr` now returns a typed PR record containing its number, URL, head, base, and draft state. Its command pins creation and read-back to the origin repository and recorded branch, and writes public-action evidence.

Delivery passes the record to review and body synchronization. Correction pushes and the ready flip read the PR by number, verify both local and remote heads against the record, and explicitly target the recorded branch. The ready flip records its command/output transcript and publishes a typed public-action result.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Later nodes resolve a PR from the current branch. | Later nodes act on the PR recorded by this run. |
| **Failure behavior** | Ambient branch or PR drift can select an unintended action target. | A head or local-branch mismatch fails before pushing, editing, checking, or flipping the PR. |

## Architecture

```mermaid
flowchart LR
  P[archon-pr] --> R[typed PR record]
  R --> V[review and body sync]
  R --> C[correction push]
  R --> F[ready flip]
  F --> E[action evidence]
```

### Changed seams

| Boundary or contract | Change | Evidence |
|---|---|---|
[...]

PR #2980, opened 2026-08-29
fix(sdlc): the delivery tails report the PR from the run's record, not the checkout

### LINKED ISSUE #2965: ship/stabilize/upkeep outcome scripts still resolve the delivered PR from the current branch
## Problem

`archon-ship`, `archon-stabilize`, and `archon-upkeep` each end in an `outcome` script that confirms the delivered PR against GitHub. All three resolve the PR by running `git branch --show-current` and passing that branch to `gh pr view`:

```python
branch = subprocess.run(["git", "branch", "--show-current"], ...).stdout.strip()
```

(`.archon/workflows/sdlc/ship/scripts/outcome.py:247`, and the stabilize/upkeep siblings.)

That is the ambient-branch resolution PR https://github.com/coleam00/Archon/pull/2950 removed everywhere else — the whole point of recording a run-owned PR identity is that no later step re-derives it from whatever happens to be checked out.

Each script **already receives** the typed record. `delivered` is bound to `$deliver.output`, which since C1 is the PR node's structured read-back (`number`, `url`, `head`, `base`, `is_draft`). Today it is used only as a null/non-null sentinel; the number sitting inside it is ignored.

## Why it was not fixed in C1

These are reads that shape a terminal report, not public writes. A wrong resolution produces a misleading report, not a wrong comment or a wrongly-flipped PR, so it sits outside that PR's accepted contract ("every later public action"). C1 had already taken four rounds of blocking corrections; three more scripts plus their fixtures did not belong in it.

## Suggested fix

Read the PR number from the `delivered` record instead of the branch. Fail clearly when the record is present but unparseable rather than falling back to branch inference — a silent fallback here would reintroduce exactly the hazard being removed.

Care needed: `outcome` is the last node of a delivery. A new failure mode there reports the whole run as failed after everything actually succeeded, so the parse path needs to be exact and covered by each pack's fixtures.

## Source

Finding R3 of the review on PR https://github.com/coleam00/Archon/pull/2950, independently confirmed against the three scripts.

Wirasm:
## Problem and outcome

`archon-ship`, `archon-stabilize`, and `archon-upkeep` each end in an `outcome` script that reports the delivered pull request. All three resolved that pull request by running `git branch --show-current` and handing the branch to `gh pr view` — the ambient resolution #2950 removed everywhere else in delivery. Each script already received the typed record: `delivered` is bound to `$deliver.output`, which is the PR node's structured read-back. It was used only as a null/non-null sentinel, and the number inside it was ignored.

- **Outcome:** the three tails report the pull request URL from the run's own record. The terminal `gh pr view`, and the `git remote` / `git branch` derivation feeding it, are gone.
- **Invariant:** no node after the PR node names the pull request from anything but the run-owned record.
- **Scope boundary:** `archon-deliver`'s own `outcome.py` is untouched — it reads a URL from flip-ready, not a record. No engine, YAML-surface, or binding change; the record was already bound.

## Review guidance

- **Feedback requested:** correctness of the deleted verification, and whether the new failure path is proportionate for a terminal node.
- **Start here:** `.archon/workflows/sdlc/ship/scripts/outcome.py:22` — `delivered_pr_url`, the record read that replaces the forge call. Its docstring carries the reasoning; the stabilize and upkeep copies are byte-identical.
- **Review order:** the ship script, then the ship YAML comment, then `ship/fixtures/delivered-unusable.stubs.yaml` (the guard seen red), then the four other delivered fixtures.
- **Lower-attention areas:** `bundled-defaults.generated.ts` (regenerated by `bun run generate:bundled`, verified by `bun run check:bundled`); the stabilize and upkeep scripts, which differ from ship only in the verdict branches above the record read.
- **Known risk or uncertainty:** `outcome` is the last node of a delivery, so a new failure mode there would report a whole run as failed after everything succeeded. That is why the parse is exact, why every defect is named, and why the refusal has its own fixture.

## Solution

Deleting the re-read rather than re-targeting it is the operator's decision on the issue, and it holds up mechanically. A record can only be present at `outcome` because the `none_failed_min_one_success` join saw deliver complete; deliver completing means flip-ready ran, and the flip verifies the draft state by reading it back before it will finish. So the `gh pr view` re-asserted an invariant another node had already established — the pack README's Cut rule by definition. It also had to name the pull request to ask, and the only name available that late is whatever branch happens to be checked out.

`delivered_pr_url` parses `INPUTS_DELIVERED` and returns either the URL or the reason the record cannot supply one. It validates rather than trusts: a report built from an unreadable record would announce a delivery while dropping the one thing its reader came for. It deliberately does **not** check `is_draft` — the record is written when the PR is created and delivery creates it as a draft, so that field is `true` on every successful run, and the docstring says so where someone would otherwise add the check.

One helper rather than the parse inlined three times: the three `main()` bodies already differ, and a named helper keeps the record read verifiable as byte-identical and gives the reasoning one home. It sits above the existing `DISCOVERY_RELAY` block so that block stays byte-identical across all four tails, which a plain `diff` still proves.

The scripts are now pure, so the five delivered fixtures stop stubbing `outcome` and execute it — the first time these scripts are fixture-exercisable at all, which is what the pack's own guard rule asks for.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `outcome` shelled out to `git remote`, `git branch --show-current`, and `gh pr view <branch>`, printing whatever URL the forge returned for the checked-out branch | `outcome` prints the `url` field of the record deliver returned; no subprocess, no network |
| **Failure behavior** | `outcome: detached HEAD — cannot resolve the PR branch`, or `outcome: delivery did not finish with a ready pull request` for any `gh` failure — including a wrong branch, a network error, or a missing `gh` | `outcome: delivery completed but its pull request record <defect>. The deliver node's own output is the evidence.` where `<defect>` is `is not valid JSON (...)`, `is a JSON <type>, not the pull request object`, or `carries no 'url' string` |

### Changed seams

|---|---|---|
| `deliver (returns: pr) → ship/stabilize/upkeep outcome` | The bound record goes from a null/non-null sentinel to the reported value | `.archon/workflows/sdlc/ship/scripts/outcome.py:22`; `sdlc/ship/fixtures/direct-delivered.stubs.yaml` |
| `outcome → GitHub` | Removed. The node makes no forge call | `git diff` on the three scripts: `import subprocess` and the `gh pr view` block deleted |

## Validation

- `bun run cli workflow test --json` — 40 passed, 0 failed (39 before this change) — proves every fixture in the repository still holds, including the five that now execute the real script.
- Real-execution evidence, from the dry-run trace of each delivered fixture: node `outcome` reports `state: completed`, `reason: "executed locally"`, and outputs `.../pull/4`, `/2`, `/3`, `/9`, `/7` — each matching that fixture's own `deliver__pr__pr` record and nothing else. The exec workspace is a **detached** worktree, and the deleted code failed outright on detached HEAD, so these fixtures could not pass if any branch lookup had survived.
- Guard seen red: `sdlc/ship/fixtures/delivered-unusable.stubs.yaml` keeps every record field except `url`; the run fails on exactly `outcome` with `outcome: delivery completed but its pull request record carries no 'url' string. The deliver node's own output is the evidence.`
- Every defect branch, run directly against all three scripts (`uv run`), identical output from each: valid record → exit 0 with the URL; `url` missing, `null`, or blank → exit 1 `carries no 'url' string`; non-JSON text → exit 1 `is not valid JSON (...)`; JSON array → exit 1 `is a JSON list, not the pull request object`.
- Caveat composition unchanged: with a seeded `discoveries.json` and `red-causes.json`, the success path prints the URL plus both sections (exit 0) and the defect path prints the failure plus both sections (exit 1).
- `bun run validate` — passed (exit 0). `bun run lint --max-warnings 0`, `bun run type-check`, `bun run format:check`, `bun run check:bundled` — all passed.
- **Not verified:** no live run against a real GitHub pull request. The behavior removed was a forge call and the behavior added reads a bound value, so the fixture suite exercising the real script against a detached checkout is the authoritative boundary; a live run would add cost without adding evidence.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. `.archon/workflows/` is a bundled reference pack; no persisted shape, identifier, or wire value changes | Only pack files and the regenerated bundled defaults are touched |
| Observability | A delivery whose record cannot name its pull request now fails the run with the specific defect instead of the generic "did not finish with a ready pull request" that also covered `gh` and network failures | Failure strings in the Behavior change table |

## Links

- Closes #2965
- Related #2950

## Summary by CodeRabbit

- **Improvements**
  - Delivery outcomes now report the pull request directly from the completed delivery record.
  - Removed reliance on repository state or external pull request lookups when confirming delivery.
  - Added clearer validation for incomplete, malformed, or missing pull request details, including actionable errors when a URL is unavailable.

- **Tests**
  - Expanded coverage for successful delivery reporting and invalid delivery records.
  - Updated dry-run scenarios to exercise real outcome handling across delivery workflows.


Wirasm:
## Ready to merge

`archon-ship`, `archon-stabilize`, and `archon-upkeep` used to close out delivery by shelling out to `git branch --show-current` and `gh pr view` to re-derive the pull request they had just opened — an ambient forge re-verification the operator decided (on issue #2965) to delete rather than re-target. This PR does exactly that: the three `outcome.py` scripts drop every `subprocess`/`gh`/`git` call and instead parse the PR URL out of the run's own typed record, already bound as `delivered` from `$deliver.output`. A new `delivered_pr_url` helper is byte-identical across all three scripts, and the file region shared with `archon-deliver`'s own untouched `outcome.py` is also byte-identical across all four tails — both confirmed with `diff`/`cmp`, not by reading. All three review scopes (code, seams, simplify) came back clean; the only finding is a cosmetic duplicate `.strip()` call.

**Validation:** `bun run validate` — PASS (exit 0; 147 pass / 0 fail across the affected packages' test suites). `bun run cli workflow test --json` — 40/40 fixtures pass. `bun run check:bundled` — up to date.

### Findings

|---|---|---|---|
| `R1` | Suggestion | `delivered_pr_url` calls `.strip()` twice on the same value instead of once | OPEN |

<details>
<summary><code>R1</code> — redundant double <code>.strip()</code> in <code>delivered_pr_url</code></summary>

**Impact:** Cosmetic only. No behavior difference — `.strip()` is idempotent and cheap on a short URL string. Purely a readability nit, duplicated identically across the three byte-identical copies of the helper.

**Evidence:** `.archon/workflows/sdlc/ship/scripts/outcome.py:55` (`if not isinstance(url, str) or not url.strip():`) and `:57` (`return url.strip(), ""`); same pattern at `.archon/workflows/sdlc/stabilize/scripts/outcome.py:57/59` and `.archon/workflows/sdlc/upkeep/scripts/outcome.py:54/56`.

**Required outcome:** Optional — assign `stripped = url.strip()` once and reuse it for both the blank check and the return value, in all three byte-identical copies (to preserve the byte-identity invariant this PR itself declares).

**Disposition:** OPEN. Non-blocking; safe to fix in this PR or leave for a follow-up given the byte-identity constraint across three files.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `simplify` | `R1` |

#### Validation

|---|---|---|
| `bun run validate` | PASS | Exit 0; final suite: 147 pass, 0 fail, 307 expect() calls across 9 files (packages/providers + packages/workflows test run, the tail of the composed validate pipeline: check:bundled, check:bundled-skill, check:bundled-schema, check:pi-vendor-map, check:capability-matrix, check:api-types, type-check, lint --max-warnings 0, format:check, test:install, test) |
| `bun run cli workflow test --json` | PASS | 40 passed, 0 failed — includes the 5 delivered fixtures (now `exec-code: true`, outcome stub dropped) and the new `delivered-unusable.stubs.yaml` |
| `bun run check:bundled` | PASS | "bundled-defaults.generated.ts is up to date (55 commands, 32 workflows, 11 scripts)." |
| `grep -n "subprocess\|gh pr\|git remote\|git branch"` across the 3 outcome.py files | PASS (no match) | Confirms complete deletion of forge/git subprocess calls |
| `diff`/`cmp` of `delivered_pr_url` across ship/stabilize/upkeep | PASS (identical) | Byte-identity invariant the PR declares, independently proven |
| `diff`/`cmp` of `DISCOVERY_RELAY`→`def main` region across ship/stabilize/upkeep/deliver | PASS (identical) | Byte-identity invariant across all four tails, independently proven |
| Direct `dryRunWorkflow` trace probe on the 5 delivered fixtures | PASS | Each outcome-node trace entry: `reason: "executed locally"`, `output` equal to that fixture's own `deliver__pr__pr.url` (pull/4, /3, /2, /9, /7) — proves real execution against a detached scratch worktree, not a stub |
| Direct `dryRunWorkflow` trace probe on `delivered-unusable.stubs.yaml` | PASS | Exactly one failed trace entry, on `outcome`, with the exact no-url message |
| Direct `uv run outcome.py` with hand-crafted `INPUTS_DELIVERED` (valid / no-url / null-url / blank-url / non-JSON / JSON-array) | PASS | 6/6 produced the exact expected exit code and message for each defect class |
| Direct `uv run outcome.py` for the advisory-stop branches (unset delivered / `null` + investigate / `null` + plan / `null` + gate-passed / `null` + no_action) | PASS | All 4 branches route unchanged, unaffected by this PR's diff |

</details>

Issue #2968, opened 2026-08-29
SDLC pack verifies its PR target twice and logs by hand — cut it back to the typed record


Wirasm:
## Problem

#2950 gave the SDLC pack a run-owned PR record so irreversible actions stop resolving their target from the ambient checked-out branch. The core of that is right and cheap. The machinery around it is not, and the pack is a reference implementation people copy — so what ships here propagates.

Cost: **+198 lines across 24 files** in `.archon/workflows/sdlc/`.

## Worth keeping

- The typed record: `output_type` / `output_format` on `archon-pr`'s `pr` node, and consumers binding `$pr.output.number` / `$pr.output.head`. ~15 lines, and it is the actual fix — the target becomes explicit instead of ambient.
- The read-back after the flip. `gh pr ready` can exit 0 without changing anything, so confirming `isDraft` afterwards guards a real, mechanical failure mode.

## Cut candidates

**1. The verification is written twice.** `target` in `archon-review.yaml` (31 lines) and `flip-ready` in `archon-deliver.yaml` (72 lines) both do the same six steps: derive the origin repo, guard its `owner/repo` shape, read the current branch, compare it to the record, read GitHub's head, compare again. Most of it is verbatim.

The hazard it guards has never been observed. #2909 says so plainly: the alarm that prompted this work was #2905, "closed invalid — the suspected wrong-target flip was an operator mapping error and the runs behaved correctly." Two copies of a 25-line check, for a failure that has not happened, is the shape `AGENTS.md` warns about under *avoid speculative defense in depth*.

So the first question is not "how do we share this?" but **does `target` earn its place at all**, given the review already targets a typed PR number. If the answer is yes, collapse both into one shared script — `scripts/gate-green.py` is the pack's existing precedent for a check several nodes run.

**2. Seventeen fixture stubs that prove nothing.** Adding `target` required a stub in 17 `.stubs.yaml` files across deliver, ship, stabilize and upkeep. The node can never actually execute in a fixture: the dry run deliberately does not deliver composed `include:` inputs to exec nodes, so under `set -u` it dies on an unbound variable and can only ever be stubbed. Those 17 lines buy a green suite and no coverage. They go away with the node.

**3. The logging harness belongs in the engine.** `flip-ready`'s `record` / `record_read` / `fail` helpers (~20 lines) reproduce per-node what the engine could retain for every node. Tracked separately in #2967 — that issue also records the two bugs the hand-rolled version needed fixing. When it lands, this block deletes.

[...]

Issue #2478, opened 2026-08-05
feat(workflows): per-mode node-key validation — reject capability-asserting keys, warn on dead tuning


Wirasm:
#2455 makes the engine tell authors when it silently drops a key. It covers one direction thoroughly and leaves the other open: a key that is **known to the schema but valid only on a different node mode** is still dropped in silence.

## The two directions

**Covered by #2455 — AI fields on non-AI nodes.** `loader.ts:297-311` maps each non-AI mode to its ignored-field list (`BASH_NODE_AI_FIELDS`, `INCLUDE_NODE_IGNORED_FIELDS`, `WORKFLOW_NODE_IGNORED_FIELDS`, `LOOP_NODE_AI_FIELDS`, `LOOP_GROUP_NODE_AI_FIELDS`), and warns. Put `model:` on a `bash:` node and you are told.

**Not covered — mode-exclusive keys on AI nodes.** `KNOWN_DAG_NODE_KEYS` is derived from the whole flat schema shape (`dag-node.ts:706`), so `parseDagNode`'s unknown-key check accepts *any* schema key on *any* node. There is no per-mode allowlist for the AI modes (`command:` / `prompt:` / `loop:`), so a mode-exclusive key lands on one of them, the mode transform drops it, and nothing is said.

The live example, which #2467 just made reachable:

```yaml
- id: review
  command: review-diff
  with:                    # include:-only. Silently dropped. No warning.
    severity: high
```

Same shape for `input:` / `isolation:` (workflow-only), `runtime:` / `deps:` (script-only), `fan_out:` (workflow-only), and `until_bash:` (loop-only) placed on a `command:` or `prompt:` node.

This is CodeRabbit's Major finding on #2455 (`dag-node.ts:708`), verified against the branch rather than taken on its word. Deferred there rather than fixed because it is genuinely a larger change than the rest of that PR — filing it so the deferral is tracked.

## Why it's worth doing

The `with:` case is the one that will actually bite. #2467 introduces `with:` as an `include:`-only key while `command:` nodes are the most common node type, and the two look adjacent in any workflow that uses both. An author who writes `with:` on a `command:` node gets a workflow that loads clean and behaves as though the inputs were never supplied — the exact failure #2455 exists to eliminate.

## Fix direction

[...]

PR #2981, opened 2026-08-29
fix(workflows): a composed command node keeps its own with: binding

### LINKED ISSUE #2964: Composition rejects a command node's $INPUTS binding that the runtime supports
## Problem

Composition validates a command node's `$INPUTS.<name>` references against the **enclosing workflow's declared inputs** and never looks at the node's own `with:` map. The runtime does the opposite: `dag-executor.ts:2203-2226` resolves a command node's `with:` and merges it over run inputs into the `$INPUTS` bag, nearest-wins, and `dag-executor.test.ts` AC4 covers exactly that.

So a construction the executor fully supports fails to load as soon as the workflow is included somewhere else.

## Reproduction

In `archon-deliver`:

```yaml
- id: sync-pr-body
  command: sync-pr-body
  with:
    pr_number: "$pr.output.number"
    pr_head: "$pr.output.head"
```

with `sync-pr-body.md` reading `$INPUTS.pr_number`.

`archon-deliver` alone loads and runs. `archon-ship`, `archon-stabilize` and `archon-upkeep`, which include it, all fail:

```
Node 'deliver': included block 'archon-deliver' references missing inputs
$INPUTS.pr_head, $INPUTS.pr_number. Pass them through 'with:'.
```

The error's advice cannot be followed. It asks the caller to supply values that are produced mid-run by a sibling node three steps earlier — `archon-ship` has no PR number to pass at launch, because the PR does not exist yet.

Measured on the SDLC pack: the dry-run fixture suite goes from 39/39 to 28/39, purely from those three load failures.

## Current workaround

Declare the names as inputs on the composed workflow with empty defaults purely so the static check can see them:

```yaml
inputs:
  pr_number:
    default: ""
    description: Internal — the run-owned PR number, bound from the pr node.
```

The node's binding still supplies the real value and wins over any caller value. It works and is what PR https://github.com/coleam00/Archon/pull/2950 ships, but the declarations exist only to satisfy a check that is looking in the wrong place, and nothing stops a caller passing a value that is then silently overridden.

## Suggested fix

Have the composition check consult the node's `with:` keys before reporting a missing input. A name supplied by the node's own binding is not missing — it is bound by the nearest source in the documented precedence order.

## Why it matters beyond the workaround

Getting this wrong is expensive. In the PR above it produced two bad commits in a row: first an `INPUTS_PR_NUMBER` env-form reference that a prompt node never substitutes (so the agent read the literal token), then — after that was caught — a rewrite that removed the binding entirely and had the agent parse a free-form evidence file instead, on the mistaken belief the binding could not work. Both survived because the load error points at a fix that does not apply, and the runtime capability is easy to miss.

## Evidence

- `packages/workflows/src/dag-executor.ts:2203-2226` — command nodes resolve `with:` into `$INPUTS`
- `packages/workflows/src/dag-executor.test.ts` AC4 — existing passing coverage for that channel
- `packages/workflows/src/include-expander.ts` — the composition check that ignores it
- PR https://github.com/coleam00/Archon/pull/2950 — the workaround, and both wrong turns

Wirasm:
## Problem and outcome

A command node's `with:` map is a runtime channel: the executor resolves it against sibling outputs and merges it over the run's inputs into the `$INPUTS` bag. Composition threw that map away and then rejected the workflow for the names it supplied — so a block that ran fine on its own became unloadable the moment another workflow included it, with an error whose advice no caller could follow.

- **Outcome:** A block whose command node binds a mid-run value now composes. `archon-ship`, `archon-stabilize` and `archon-upkeep` load, and the composed node resolves the real PR number instead of the empty string a decoy input was supplying.
- **Invariant:** A binding means the same thing standalone and composed. Node-local bindings stay the nearest input layer, winning over a composed block's inputs, which win over the run's `$INPUTS` — the order the authoring guide already documents.
- **Scope boundary:** The SDLC pack still declares its `pr_number` / `pr_head` decoy inputs. This change makes them unnecessary and inert; #2968 deletes them. Script (`exec`) nodes were never affected — their `with:` map survives materialization untouched and rides the `INPUTS_*` env channel.
- **Root cause:** `materializeBlockCommandPrompts` compiles a command body into an inline prompt so the block's refs can be namespaced and its declared inputs bound. The inline source variant has no `with:` field — the schema admits it only on the `command` variant (#2478) — so expansion dropped the map. `applyInputsMacro` then walked the compiled body, found `$INPUTS.<name>` with nothing in the include's resolved inputs, and reported it missing.

## Review guidance

- **Feedback requested:** Correctness of the carrier choice, and whether the shield in `applyInputsMacro` is scoped to the right surface.
- **Start here:** `packages/workflows/src/include-expander.ts:996` — the drop that became a move. The comment it replaces claimed dropping was "behavior-preserving"; it was, but only relative to an executor that could no longer see the field.
- **Review order:** `include-expander.ts` (materialization + the missing-input shield) → `compiled-command.ts` (the payload and why it exists) → `template-walker.ts` (so both rewrite passes reach it) → `dag-executor.ts` / `dry-run.ts` / `loader.ts` (one read each, at the sites that already read `source.with`).
- **Two commits, read in order.** The second closes a path the first made reachable — see the last paragraph of Solution. It is separable and independently tested.
- **Lower-attention areas:** The `dry-run.ts` change is mechanical — `isCommandSourced` collapsed into the `isAgentNode` check it always implied.
- **Known risk or uncertainty:** The payload rides a symbol, which `structuredClone` drops. That is the trap `compiled-command.ts` documents at the top of the file: a payload missed in the clone or either rewrite pass works at one nesting level and silently vanishes at two. `carries a surviving binding through two include levels` is the test that closes it, and the real pack confirms it — see Validation.

## Solution

The bindings move to `COMPOSED_BINDINGS`, an engine-private per-node payload alongside the two that already exist for exactly this purpose (`COMPOSED_NODE`, `COMPILED_LOOP_COMMAND`). Both the executor and the dry-run simulator read it wherever they already read `source.with`, so preview and execution resolve a composed binding through the same `resolveNodeBindings` — `if_skipped`, the failed-producer guard and strict whole-ref resolution all keep working.

Keeping the node command-sourced instead was the alternative, and it is worse: `stepName` for a composed command node would change from the node id to the command name, and that string is persisted in `workflow_events`. Widening the inline source variant to accept `with:` was the other, and it re-admits the authoring mistake #2478 deliberately made a type error. The payload keeps the authored type invariant intact — `with:` is still only writable on a command node.

The missing-input check consults the surviving map, and the substitution shield is scoped to the compiled body alone. That is not incidental: the binding bag reaches `buildPromptWithContext(rawPrompt, …)` and nothing else, so a `when:` or `systemPrompt:` naming the same input still reads the run's bag and still errors when nobody supplied it.

The second commit closes a path the first one made reachable. The loader's binding scan enforces that a bound producer is reachable through `depends_on`, and it keys off `source.with` — which a composed node no longer has. So a caller ref forwarded into a binding (`with: { src: $other.output }` on the include, spliced into the block's `$INPUTS.src`) reached the executor unchecked and failed the node mid-run. The child is proven hermetic *before* the caller's value is inserted, so the scan over the flattened graph is the only one that can see it; it now reads the same payload the executor does. Both the rejection and the legitimate-dependency case have tests.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Including a block whose command node binds a mid-run value fails the load: `included block 'archon-deliver' references missing inputs $INPUTS.pr_head, $INPUTS.pr_number. Pass them through 'with:'` | The block composes; the node's binding resolves at run time |
| **Failure behavior** | Declaring the names as inputs with empty defaults silenced the error, then spliced the empty default into the prompt — a silent wrong value, and a caller could override it | The node's binding wins over any caller value; a name nothing binds still fails the load with the same error |

## Validation

- `bun run validate` — exit 0, 8878 tests passing, 0 failures.
- `bun run lint`, `bun run type-check`, `bun run format:check` — all clean.
- `archon workflow test` on this repo's SDLC pack — **39 passed, 0 failed**, the number the issue reports dropping to 28/39 under the load failures.
- Seven new tests, each seen red against the unfixed code before its fix landed. The two that matter most: `a command node own binding wins over a caller-supplied input of the same name` failed with `"Update PR caller value."` — the silent override the issue warned about — and `still reports a command input no binding and no caller supplies` failed because the old error named `$INPUTS.pr_number` alongside `$INPUTS.repo`, proving the shield is per-name rather than a blanket exemption.
- Real-pack inspection of the expanded graph: `archon-deliver` standalone keeps its command-sourced node and `with:` map unchanged, while `archon-ship`, `archon-stabilize` and `archon-upkeep` each carry `pr_number: $deliver__pr__pr.output.number` — correctly re-namespaced through two include levels — with the `$INPUTS.pr_number` token left standing in the prompt for the executor.
- One Windows CI sample on the first commit — `cli.test.ts` `emits { ok: false } … outside a git repository`, 6313ms against a 5000ms budget — is the tracked budget-edge class (#2924). Attributed rather than reran: the log carries the BODY-timeout string, and the seven sibling tests that spawn the same CLI all landed at 828-1062ms on that runner. The test does reach this diff (`workflow list` runs discovery + expansion), so the path was measured: `workflow list --json` against a scratch tmpdir over 5 runs is 0.63-0.71s on this branch versus 0.65-0.71s with `packages/workflows/src` at `origin/dev`. No cost added; a 4-second overshoot is not attributable here, and the same commit's Ubuntu leg passed.
- **Not verified:** No live agent run. The dry-run simulator shares `resolveNodeBindings` with the executor by construction (#2637 R2), and the executor's read is the same one-line expression, so the resolution path is covered without spending a real run.

## Links

- Closes #2964
- Related #2968

### DIFF
--- a/packages/workflows/src/compiled-command.ts
+++ b/packages/workflows/src/compiled-command.ts
@@ -1,4 +1,4 @@
-import type { TriggerRule } from './schemas';
+import type { BindingDirective, TriggerRule } from './schemas';
 import type { JsonValue } from './output-ref';
 
 /** Engine-private per-node metadata attached during load-time include expansion.
@@ -17,6 +17,38 @@ export type CompiledLoopCommand =
 
 export interface LoopWithCompiledCommand {
   [COMPILED_LOOP_COMMAND]?: CompiledLoopCommand;
+}
+
+/**
+ * The node-local `with:` map a command node carried before include expansion compiled
+ * its body into an inline prompt (#2964).
+ *
+ * Composition materializes a command node so the block's refs can be namespaced and its
+ * declared inputs bound, which leaves `with:` — a field the schema admits only on the
+ * `command` source variant, so an author cannot write it on a prompt node (#2478) —
+ * with nowhere to live on the result. Dropping it silently unbound every name the node
+ * supplied itself: the compiled body's `$INPUTS.<name>` reached the model as a literal
+ * token, or as an empty string when a decoy input had been declared to get past the
+ * missing-input check. The executor and the dry-run simulator resolve this payload for a
+ * materialized node exactly as they resolve `source.with` for an unmaterialized one, so
+ * a binding means the same thing standalone and composed.
+ */
+export const COMPOSED_BINDINGS = Symbol('archon.composed-bindings');
+
+export type ComposedBindings = Record<string, JsonValue | BindingDirective>;
+
+export interface NodeWithComposedBindings {
+  [COMPOSED_BINDINGS]?: ComposedBindings;
+}
+
+/** Read a node's surviving node-local bindings. One accessor, so the cast lives in one place. */
+export function readComposedBindings(node: object): ComposedBindings | undefined {
+  return (node as NodeWithComposedBindings)[COMPOSED_BINDINGS];
+}
+
+/** Attach the bindings a materialized command node must keep. @see COMPOSED_BINDINGS */
+export function attachComposedBindings(node: object, bindings: ComposedBindings): void {
+  (node as NodeWithComposedBindings)[COMPOSED_BINDINGS] = bindings;
 }
 
 /** @see ComposedNodeMeta */
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -118,6 +118,7 @@ import { buildTruncationMarker } from './utils/output-truncation';
 import { writeNodeArtifact, readNodeArtifacts } from './artifacts-index';
 import {
   COMPILED_LOOP_COMMAND,
+  readComposedBindings,
   readComposedMeta,
   type LoopWithCompiledCommand,
   type IncludeCommandContent,
@@ -2090,7 +2091,10 @@ async function executeNodeInternal(
   // Command file name when this agent node is command-sourced, undefined for an
   // inline prompt — replaces the old `node.command` field access throughout.
   const commandName = node.source.kind === 'command' ? node.source.name : undefined;
-  const nodeWith = node.source.kind === 'command' ? node.source.with : undefined;
+  // Include expansion compiles a command body into an inline prompt and moves the node's
+  // `with:` map to the engine-private payload, so a composed node binds exactly as the
+  // same node does standalone (#2964).
+  const nodeWith = node.source.kind === 'command' ? node.source.with : readComposedBindings(node);
   // Persisted step_name is namespaced ('<groupId>.' prefix) for loop_group bodies;
   // '' for the top-level DAG → identical to node.id. The in-process emitter payloads
   // below stay raw (node.id) — live SSE/CLI consumers key off those. See #2090.
--- a/packages/workflows/src/dry-run.test.ts
+++ b/packages/workflows/src/dry-run.test.ts
@@ -1668,6 +1668,59 @@ describe('dryRunWorkflow — node-local with: bindings (#2637)', () => {
     expect(gate?.state).toBe('failed');
     expect(gate?.reason).toContain("'corrections' failed");
   });
+
+  test('a composed command node resolves its own binding, not an empty caller input (#2964)', async () => {
+    const cwd = mkdtempSync(join(tmpdir(), 'archon-dry-run-composed-binding-'));
+    temporaryDirectories.push(cwd);
+    mkdirSync(join(cwd, '.archon', 'commands'), { recursive: true });
+    writeFileSync(join(cwd, '.archon', 'commands', 'sync-pr-body.md'), 'Update PR $INPUTS.number');
+
+    // The shape from the issue: the block opens a PR, then a command node reads the number
+    // that node produced. The caller has nothing to pass at launch — the PR does not exist
+    // yet — so the value can only come from the node's own binding.
+    const deliver = makeTestWorkflow({
+      name: 'deliver',
+      nodes: [
+        {
+          id: 'pr',
+          prompt: 'Open the PR',
+          output_format: { type: 'object', properties: { number: { type: 'number' } } },
+        },
+        {
+          id: 'sync',
+          command: 'sync-pr-body',
+          depends_on: ['pr'],
+          with: { number: '$pr.output.number' },
+        },
+      ],
+    });
+    const ship = makeTestWorkflow({
+      name: 'ship',
+      nodes: [{ id: 'deliver', include: 'deliver' }],
+    });
+    const { workflows, errors } = expandWorkflowIncludes(
+      new Map([
+        ['deliver', deliver],
+        ['ship', ship],
+      ]),
+      new Map([['sync-pr-body', 'Update PR $INPUTS.number']])
+    );
+    expect(errors).toHaveLength(0);
+
+    const result = await dryRunWorkflow({
+      workflow: workflows.get('ship')!,
+      userMessage: '',
+      cwd,
+      stubs: { deliver__pr: { number: 2964 }, deliver__sync: 'done' },
+    });
+
+    expect(result.outcome).toBe('completed');
+    // Composition materializes the body into an inline prompt, so this proves the moved
+    // binding still reaches resolveNodeBindings — the channel a real run uses.
+    expect(result.trace.find(t => t.nodeId === 'deliver__sync')?.resolvedText).toBe(
+      'Update PR 2964'
+    );
+  });
 });
 
 describe('dryRunWorkflow — effective provider/model per node', () => {
--- a/packages/workflows/src/dry-run.ts
+++ b/packages/workflows/src/dry-run.ts
@@ -15,7 +15,11 @@ import {
   type ShellInputContext,
 } from './dag-executor';
 import { evaluateCondition } from './condition-evaluator';
-import { COMPILED_LOOP_COMMAND, type LoopWithCompiledCommand } from './compiled-command';
+import {
+  COMPILED_LOOP_COMMAND,
+  readComposedBindings,
+  type LoopWithCompiledCommand,
+} from './compiled-command';
 import { declaredFieldsFromSchema, canonicalValueText, type JsonValue } from './output-ref';
 import { discoverScriptsForCwd } from './script-discovery';
 import {
@@ -1043,7 +1047,6 @@ async function simulateNode(
       return;
     }
 
-    const isCommandSourced = isAgentNode(node) && node.source.kind === 'command';
     const sourceText = isAgentNode(node)
       ? node.source.kind === 'command'
         ? await loadDryRunCommand(ctx, node.source.name)
@@ -1056,17 +1059,21 @@ async function simulateNode(
     // on (unknown producer, skipped without if_skipped) fails the simulated node
     // with the executor's own error, via the catch below. Command bindings merge
     // over run inputs into the `$INPUTS` bag; script bindings ride the exec env.
+    // A composed command node carries its map in the engine-private payload, having
+    // been materialized into an inline prompt at expansion (#2964).
     const nodeWith = isExecNode(node)
       ? node.with
-      : isCommandSourced && isAgentNode(node) && node.source.kind === 'command'
-        ? node.source.with
+      : isAgentNode(node)
+        ? node.source.kind === 'command'
+          ? node.source.with
+          : readComposedBindings(node)
         : undefined;
     const nodeBindings =
       nodeWith !== undefined
         ? resolveNodeBindings(node.id, nodeWith, bindingShellContext(ctx, outputs), ctx.inputs)
         : undefined;
     const commandInputs =
-      isCommandSourced && nodeBindings !== undefined
+      isAgentNode(node) && nodeBindings !== undefined
         ? { ...(ctx.inputs ?? {}), ...nodeBindings }
         : undefined;
     const resolvedText =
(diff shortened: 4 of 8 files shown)


Wirasm:
## Ready to merge

Composition used to compile a command node's body into an inline prompt and, in doing so, replace `node.source` wholesale — silently discarding the node's own `with:` binding map. Independent verification against the pre-fix code on `dev` (`packages/workflows/src/include-expander.ts:996`) confirms this exactly as the implementer describes: the old code spread `...node` but overwrote `source` with `{ kind: 'inline', prompt: compiled.prompt }`, dropping `source.with`, and the old comment's claim that this was "behavior-preserving" was false for a composed node — it was only true for the standalone path the comment was reasoning about. The fix carries the map through as an engine-private `COMPOSED_BINDINGS` symbol payload, resolved by both the executor and the dry-run simulator at the exact site they already resolve `source.with`, through one shared `resolveNodeBindings` function. All three reviewers (code, seams, simplify) independently traced the propagation surface — every clone site, both rewrite passes, the executor/simulator resolution, the missing-input shield's scope, and the type's fidelity to the existing `with:` schema — and returned zero findings.

The head moved once during this review (`4c7d5dc4a` → `4183bd1b0`), adding one commit that closes a real gap the seam pass had examined and initially declined to file: a composed command node's binding could reference a caller-forwarded parent-scope value (`with: { src: '$other.output' }` on the include, substituted into the child's binding) with no `depends_on` edge enforcing that the parent node runs first, and `loader.ts`'s load-time binding-dependency check silently skipped this case for materialized command nodes. I independently reproduced this against the original reviewed head before trusting the fix — see the finding detail below.

**Validation:** `tsc --noEmit` clean; `bun run lint` clean; `packages/workflows` full test suite (25 groups, 2619 tests) — 0 failures; `bun run cli workflow test sdlc --json` — 39/39 fixtures pass (matches the PR's claimed recovery from 28/39).

### Findings

No open findings. One gap was found and closed mid-review — see below.

<details>
<summary>Mid-review correction: a composed binding could skip its own depends_on check (closed)</summary>

**What it was:** `loader.ts`'s `validateDagStructure` has a load-time check mirroring `fan_out.items`'s: a binding value that references `$producer.output` must name an upstream `depends_on` producer, or the load fails with a message telling the author which id to add. The original PR commits (`4c7d5dc4a`) extended this check for exec/loop nodes but left a materialized command node's `nodeWith` resolving to `undefined` in this one function — so a composed command node's binding was invisible to this specific check. On its own that looked inert (round 1's seam pass reasoned a binding's producer ref is always renamed alongside its `depends_on` edge by the same closure, so the two can't drift). That reasoning covers refs the block authors itself; it misses a caller forwarding a parent-scope value through `with:` on the include — a path that only exists post-#2964, since pre-fix a command node's `with:` was dropped entirely and never reached this substitution.

**I reproduced it as a live gap before trusting the fix**, using a throwaway worktree pinned to the original reviewed head (`4c7d5dc4a`): a child block's command node binds `$INPUTS.src`; the caller forwards `with: { src: '$other.output' }` on the include, where `other` is a real sibling node with no `depends_on` edge from the include to it. On `4c7d5dc4a` this loaded successfully — `errors: []` — with a binding that could resolve to an unproduced value at runtime, exactly the class of bug the neighboring `fan_out.items` check exists to catch at load time instead.

**The correction (`4183bd1b0737bc7def930b1b5bbf48b97e456767`)** adds `readComposedBindings(node)` to that one ternary in `loader.ts:1099-1104`, reusing the same `transitiveDepsOf` primitive and error-message shape as the `fan_out.items` check three lines above it — not new machinery. Reran the reproduction on the corrected head: the workflow now fails to load with `Node 'inc__sync' binding 'with.v' references '$other.output', which is not an upstream dependency — add 'other' to 'inc__sync'.depends_on so its value is produced first`. A paired test (`'accepts the same forwarded ref when the include declares the dependency'`) proves the check isn't overbroad — declaring the `depends_on` makes the identical scenario load and resolve correctly.

**Disposition:** FIXED. Verified by direct reproduction against both heads, not by trusting the commit message. Full validation reran clean on the corrected head (see below).

</details>

<details>
<summary>Independent verification of the load-bearing claim</summary>

The operator asked for the silent-drop claim to be reproduced or refuted directly, and the fix's end-to-end closure confirmed by inspection rather than by trusting the PR body.

**Reproduced against pre-fix `dev`.** `materializeBlockCommandPrompts` (`include-expander.ts:996`, pre-fix) read:

```ts
return { ...node, source: { kind: 'inline', prompt: compiled.prompt } };
```

`node.source` — a discriminated union — is replaced entirely, so `node.source.with` (the command node's binding map) has no way to survive. The old code's own comment claimed this was safe because `with:` was "already inert past this point at runtime today," reasoning about a materialized node executed as a **standalone** workflow. That reasoning doesn't hold for a **composed** node: `dag-executor.ts`'s pre-fix `nodeWith = node.source.kind === 'command' ? node.source.with : undefined` meant a materialized node's bindings were unconditionally `undefined`, so whatever value `$INPUTS.pr_number` resolved to came only from the run's own `$INPUTS` bag — i.e., the SDLC pack's declared-empty-default workaround (`pr_number: { default: "" }`), not the real value the node's own `with:` was supposed to bind mid-run. This confirms the claim: the workaround did not "work" as issue #2964's own text asserted — it silently spliced an empty string into composed runs.

**Confirmed the fix closes it, independently, against the real pack.** Rather than trusting the PR body's example, I loaded the actual `.archon/workflows/sdlc` pack in the review's own worktree via `discoverWorkflowsWithConfig` (the same function the CLI uses) and inspected the expanded `archon-ship`, `archon-stabilize`, and `archon-upkeep` graphs directly:

```
--- archon-ship ---
node id: deliver__sync-pr-body
source.kind: inline
composed bindings: {"pr_number":"$deliver__pr__pr.output.number","pr_head":"$deliver__pr__pr.output.head"}
CONTAINS $INPUTS.pr_number: true
CONTAINS $INPUTS.pr_head: true
```

Identical results for `archon-stabilize` and `archon-upkeep`. This is genuinely **three** levels of composition, not two: `archon-ship` includes `archon-deliver` as node `deliver`; `archon-deliver` itself includes `archon-pr` as node `pr` (`archon-deliver.yaml:69-70`); `archon-pr` has its own internal node `pr` that produces `.output.number`. The re-namespaced ref `$deliver__pr__pr.output.number` is correct for that chain, and the literal `$INPUTS.pr_number` / `$INPUTS.pr_head` tokens are confirmed still standing in the composed prompt text for the executor to resolve at runtime — matching the design (the binding bag reaches `buildPromptWithContext` only, nothing else). `bun run cli workflow test sdlc --json` passes 39/39 in this same worktree, matching the PR's validation claim exactly.

**Confirmed one resolver, not two.** `resolveNodeBindings` is defined once (`dag-executor.ts:438`) and imported by `dry-run.ts`; both `executeNodeInternal` and `simulateNode` compute `nodeWith` with the same `node.source.kind === 'command' ? node.source.with : readComposedBindings(node)` expression and feed it to that one function, so the executor and dry-run simulator cannot diverge on this payload by construction.

**Confirmed the load-time rejection is now correct, not merely removed.** `applyInputsMacro`'s shield is per-name (`shielded?.has(name) === true`) and scoped to the `agent.prompt` slot only — a genuinely unbound, uncaller-supplied name still fails the load with the original error message (`include-expander.test.ts`'s `'still reports a command input no binding and no caller supplies'` test), and a name on `when:`/`systemPrompt:` is untouched by the shield since the executor never merges node bindings into those surfaces.

**No new authored YAML surface.** `COMPOSED_BINDINGS` is a `Symbol`-keyed engine-private property, following the same pattern as the pre-existing `COMPOSED_NODE`/`COMPILED_LOOP_COMMAND` symbols — it does not touch the schema, and `with:` remains a type error on the `inline` source variant exactly as #2478 intended. `ComposedBindings`'s type (`Record<string, JsonValue | BindingDirective>`) is identical to the schema's own `with:` type on the `command` variant, so there's no round-trip gap.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | No findings (round 1, on `4c7d5dc4a` — unaffected by the round-2 correction, which touches a different file) |
| `seams` | Round 1 (on `4c7d5dc4a`): one asymmetry examined and initially declined as inert. Round 2 (coordinator, on `4183bd1b0`): that asymmetry reproduced as a live gap and confirmed fixed — see Findings above |
| `simplify` | No findings (round 1, on `4c7d5dc4a` — unaffected by the round-2 correction) |

#### Validation

Reran in full on the final head (`4183bd1b0737bc7def930b1b5bbf48b97e456767`) after the round-2 correction landed.

|---|---|---|
| `bun x tsc --noEmit` (packages/workflows) | PASS | Clean, no output |
| `bun run lint` (repo-wide) | PASS | All packages linted clean, 0 warnings |
| `bun run test` (packages/workflows) | PASS | 25 test groups, 2619 tests, 0 failures, exit 0 |
| `bun run cli workflow test sdlc --json` | PASS | 39/39 fixtures pass, including all `archon-ship`/`archon-stabilize`/`archon-upkeep` fixtures |
| Real-pack composed-graph inspection (`discoverWorkflowsWithConfig` on `.archon/workflows/sdlc`) | PASS | `archon-ship`/`archon-stabilize`/`archon-upkeep` each carry the correctly re-namespaced `COMPOSED_BINDINGS` payload for `deliver__sync-pr-body`, verified through actual 3-level nesting |
| Reproduction: caller-forwarded binding ref with no `depends_on`, run against `4c7d5dc4a` | FAIL (as expected — proves the gap) | Workflow loaded with `errors: []`; binding could resolve an unproduced value at runtime |
| Same reproduction, run against `4183bd1b0` | PASS (correctly rejected) | Load fails with an actionable `depends_on` error; paired test proves the check isn't overbroad |
| `bun run validate` | NOT RUN | Ran the narrower checks the changed behavior requires (type-check, lint, full workflows test suite, real-pack fixture run, direct reproduction) instead of the full repo suite; nothing in the diff touches other packages |

</details>

Issue #2982, opened 2026-08-29
test(ci): delete or reshape 62 low-value tests, and close the test-file type-check hole that is red-lining Windows


Wirasm:
## Summary

I audited every test file under `packages/*/src` and `scripts/` (about 320 files) against recent CI failure history, looking for the thing that keeps turning Windows red: low-value tests and speculative defense-in-depth.

The thesis half-holds, and the half that fails is the more useful result.

| Bucket | Count | What it is |
|---|---|---|
| Delete | 29 tests | Assert a literal the test itself wrote, or a framework guarantee, or a duplicate of the test next to it |
| Rewrite | 33 tests | The behavior deserves protection, this test cannot see it |
| Fix the coupling | 12 sites | Real protective value, flaky mechanism (timing standing in for synchronization, undeclared I/O, a subprocess where in-process proves the same thing) |
| Source guards to remove | **0** | See below |

**Bucket 4 came back empty.** Five independent passes over production code in every package looked for guards whose only consumer is a test, applying the rule from PR #2969 (a guard must protect an action its own code takes; if no test can see it fire red, it is a comment). Every candidate held up: exhaustiveness backstops on discriminated unions, webhook signature checks against untrusted input, `working_path` null handling that a real older-binary resume path reaches, retry sets with measured reproductions in their docblocks. Archon does not have a speculative-defense problem in its source.

It has a **test-file type-checking hole**, and that is what is actually red-lining CI.

## The finding that reframes this

`packages/core`, `packages/cli`, `packages/adapters` and `packages/server` all carry `"exclude": [..., "**/*.test.ts"]` in their `tsconfig.json`. Test files in those four packages are checked by nothing: not `tsc`, not typed ESLint. The other five packages do not exclude them.

On 24 August, commit `c654ca957` landed a test on dev with an undefined identifier:

```
packages/cli/src/cli.test.ts:130
  } finally {
    rmSync(dir, { recursive: true, force: true });
    ^
ReferenceError: rmSync is not defined
```

`rmSync` was never imported. That failed `test (windows-latest)` **eight times across four unrelated branches** (`dev`, `fix/2859-detached-run-control-determinism`, `archon/task-fix-2875-sqlite-statement-finalization`, and others), each time at roughly 1000 to 1700 ms, which reads as an assertion failure rather than a timeout and so does not match any family in #2924. It needed its own hotfix branch and PR #2880 (`01852b6e4`, "restore the rmSync call site left undefined on dev") to clear.

[...]

Issue #2407, opened 2026-08-03
test(tsconfig): type-check tests in core/adapters/server/cli (pass 2, strict chain: core → adapters → server → cli)


Wirasm:
> **Split 2026-08-31 — this issue is now the tracking and policy parent; the `[DESIGN FIRST]` staging question is settled.** Delivery is four package tranches plus one extracted behavior-test rewrite, each its own input contract, hard-ordered by the tsconfig include chain and recorded as native blocked-by links: **#3052 (core) → #3053 (adapters) → #3054 (server) → #3055 (cli)**, with #3053 additionally blocked by **#3051** (the fork-detection tautology rewrite, extracted so a mechanical tranche cannot silence TS2367).
>
> **The policy bar binds every tranche**: no `any`, no `@ts-expect-error`, no `tsconfig.test.json`; diagnostics fixed rather than cast away; no test lost, suites green at each branch point; each tranche flips only its own package's excludes.
>
> **Re-measured 2026-08-31 at `eb182ac5d`, superseding the table below**: core **1,203 errors / 39 files** (was 891/34), adapters **367/13** (unchanged), server **129/10** (was 93/9), cli **98/6** (was 68/6) — total **1,797 in 68 files**, ~27% growth in under a month. Children re-measure at their branch points; written counts are never targets. The 8-file `bun:test` quick win is now 7 files (core 2, adapters 5) and folds into each tranche rather than landing as a gate-less pre-PR. Rot items 1–3 below are owned by #3052, #3053, and #3051 respectively; item 4 belonged to pass 1's packages. #2982 step 1 continues to route through this chain. This issue closes when #3055 lands.

Implementation pass 2 of #2302, following 2406. This is the expensive tranche and has a **hard ordering constraint**.

| package | errors | files | note |
| --- | --- | --- | --- |
| `core` | 891 | 34 | must go first — see below |
| `adapters` | 367 | 13 | 220 are just missing imports |
| `server` | 93 | 9 | cheap, but blocked |
| `cli` | 68 | 6 | cheap, but blocked |
| **total** | **1,419** | **62** | |

## Ordering is not optional

```
cli      excludes: [..., '../core/src/**/*.test.ts', '../server/...', '../adapters/...']
server   excludes: [..., '../core/src/**/*.test.ts', '../adapters/src/**/*.test.ts']
adapters excludes: [..., '../core/src/**/*.test.ts']
```

[...]

PR #2984, opened 2026-08-29
test(core): stop orchestrator.test.ts opening a real database on every test

Wirasm:
## Problem and outcome

`packages/core/src/orchestrator/orchestrator.test.ts` stubs twenty modules around `handleMessage`, but not `../db/messages` or `../db/env-vars`. The fixture platform adapter reports `'mock'`, so every test took the non-web branch that persists the assistant reply and loads per-codebase env vars, and both reached the real lazy `getDatabase()` singleton — 108 attempts per run, each one invisible.

- **Outcome:** the file makes zero `getDatabase()` calls instead of 108. Locally it also drops from 3.52 s to 0.31 s, but see the wall-clock note below before treating that as a CI win — it is not one.
- **Invariant:** no test in this file performs undeclared database I/O. All 16 modules under `packages/core/src/db` route through `./connection`, so a zero `getDatabase()` count proves this for every one of them, not just the two stubbed here.
- **Scope boundary:** no product code changed, no test was added, removed, or reshaped. The other offenders in #2982 are separate steps.
- **Root cause:** both call sites swallow their own failure — `messageDb.addMessage(...)` into a `.catch` (`orchestrator-agent.ts:2899`, `:3162`) and `getCodebaseEnvVars` into a `try/catch` (`orchestrator-agent.ts:2443`) — each warning through the mocked logger. `getArchonHome` is mocked to a path that does not exist, so each attempt threw `SQLiteError: unable to open database file` and nothing printed. The file was not opening a database handle 108 times; it was failing to open one 108 times, silently.

### Correcting the wall-clock attribution

#2982 ranks this file by cost: 3.51 s for 75 tests, 47 ms per test. That number is real but it is not SQLite, and it does not reproduce off macOS.

The mocked `getArchonHome` returns `/home/test/.archon`. On macOS `/home` is an autofs mount, so every failed open pays an automounter lookup. Measured directly: 108 failed `new Database()` calls under `/home/test` cost **3403 ms**; the same 108 failed opens under a plain missing directory cost **1 ms**. In situ, the *unfixed* file with only the mocked home path moved off `/home` runs in 321–429 ms — indistinguishable from the fixed file.

So the 3.2 s was automounter resolution on one developer platform. On the Linux and Windows runners those 108 failed opens are plain `ENOENT` and cost approximately nothing, and this file was never a meaningful CI cost.

That does not make the fix optional, it just changes why it is worth making:

- Undeclared I/O in a test is a defect on its own terms, and AGENTS.md asks for it to be removed rather than tolerated.
- It is a live landmine. The tests only stay clean because `/home/test/.archon` happens not to exist. On any machine where it does, or after any change to that mock, these 75 tests would open and write a real SQLite database.
- It is exactly the failure shape #2982 itself argues against: a path whose failure no test can see.

## Review guidance

- **Feedback requested:** correctness of the claim that nothing in this file depended on the real handle, and whether the corrected attribution above should change this file's ranking in #2982.
- **Start here:** `packages/core/src/orchestrator/orchestrator.test.ts:96` — the two `mock.module` blocks, placed with the other DB stubs and before the module under test is imported.
- **Lower-attention areas:** none; the diff is 18 added lines in one test file.
- **Known risk or uncertainty:** none in the change. Both stubs return exactly what the swallowed real calls already produced (`{}` env vars, `[]` workflow-result messages, a resolved `addMessage`), so no observable behavior in these tests changes.

## Solution

Add the two `mock.module` blocks the sibling `orchestrator-agent.test.ts` already carries, mirroring its shapes:

- `../db/messages` — `addMessage`, `listMessages`, `getRecentWorkflowResultMessages`
- `../db/env-vars` — `getCodebaseEnvVars`

`../db/env-vars` is not named in the issue. It surfaced while verifying the fix: after stubbing `../db/messages`, tracing `getDatabase()` still showed five calls, all from `getCodebaseEnvVars` on the same `handleMessage` path with the same swallowed-failure shape. It is the same defect in the same file, found by the same measurement, so it is fixed in the same pass rather than left to reappear as a second ticket.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | 108 `new SqliteAdapter('/home/test/.archon/archon.db')` attempts per run | Zero `getDatabase()` calls |
| **Failure behavior** | `SQLiteError: unable to open database file`, caught and warned to a mocked logger — invisible to the run | No database is reached, so there is no failure to hide |
| **Local wall clock (macOS)** | 3.52 s | 0.31 s — automounter cost, see the attribution note |

## Validation

- `bun test src/orchestrator/orchestrator.test.ts` in `packages/core` — 75 pass, 0 fail, 142 expect() calls, eight consecutive runs between 279 and 378 ms. Same command on the parent commit: 75 pass, 0 fail, 142 expect() calls, three runs at 3.54/3.52/3.49 s. Identical pass count and identical assertion count on both sides proves no test went red and none silently changed shape.
- Mechanism proof: a temporary `process.stderr.write` in `getDatabase()` and in the `SqliteAdapter` constructor (reverted, not committed) counted 108 calls before, 5 after the messages stub, 0 after the env-vars stub. Every one of the 108 reported `cached=false`, which is how the failing-open behavior was established; the five intermediate stack traces all read `getCodebaseEnvVars → handleMessage`, which is how the second offender was found.
- Attribution proof for the wall clock: 108 failed `new Database()` calls measured at 3403 ms under `/home/test` versus 1 ms under `/private/tmp/<missing>`; and the unfixed file, with only the mocked `getArchonHome` repointed off `/home`, running at 321/337/429 ms. Both reverted, neither committed.
- No test in the file asserted on the affected paths: `grep -n "persist_failed\|workflow_result_messages\|addMessage\|listMessages\|getRecentWorkflowResult\|remote_agent_messages"` returns nothing, and the file's only two logger assertions (`:1508`, `:1618`) are `toHaveBeenCalledWith` on unrelated messages, not call counts, so removing the swallowed warns cannot affect them.
- `bun run test` in `packages/core` — exit 0, 43 batteries, 0 fail.
- `bun run validate` at the repository root — exit 0 (bundled checks, API types, `type-check` across all eleven packages, `lint --max-warnings 0`, `format:check`, install test, full test suite).
- **Not verified:** `packages/core/tsconfig.json` excludes `**/*.test.ts`, so `type-check` does not read the edited file. That hole is #2407's scope and step 1 of #2982, not this PR's. ESLint and Prettier did run over it, through the lint wrapper and the pre-commit hook. No before/after measurement was taken on the CI runners themselves; the claim that the saving does not transfer rests on the local isolation above, not on a runner measurement.

## Links

- Related #2982 (suggested-order step 2)

### DIFF
--- a/packages/core/src/orchestrator/orchestrator.test.ts
+++ b/packages/core/src/orchestrator/orchestrator.test.ts
@@ -89,6 +89,24 @@ mock.module('../db/sessions', () => ({
   transitionSession: mockTransitionSession,
 }));
 
+// handleMessage persists the assistant reply for every non-web platform, and the
+// fixture adapter reports 'mock'. Without this stub those calls reach the real
+// lazy getDatabase() singleton on every test — and both call sites swallow the
+// resulting error (addMessage into a .catch, getRecentWorkflowResultMessages into
+// its own try/catch), so the I/O is invisible rather than red. #2982
+mock.module('../db/messages', () => ({
+  addMessage: mock(() => Promise.resolve()),
+  listMessages: mock(() => Promise.resolve([])),
+  getRecentWorkflowResultMessages: mock(() => Promise.resolve([])),
+}));
+
+// Same shape as the messages gap above: the chat path loads per-codebase env
+// vars for any conversation carrying a codebase_id, and its failure lands in a
+// `codebase_env_vars_load_failed` warn on the mocked logger. #2982
+mock.module('../db/env-vars', () => ({
+  getCodebaseEnvVars: mock(() => Promise.resolve({})),
+}));
+
 // Command handler mock
 const mockHandleCommand = mock(() =>
   Promise.resolve({ message: '', modified: false, success: true })


Wirasm:
## Ready to merge

Eighteen added lines, one file, no product code touched. The PR adds two `mock.module()` blocks to `packages/core/src/orchestrator/orchestrator.test.ts` for `../db/messages` and `../db/env-vars`, mirroring shapes the sibling file `orchestrator-agent.test.ts` already carries. Before this change every test in the file that exercised `handleMessage` against the fixture (non-web) platform adapter reached the real, lazily-initialized `getDatabase()` singleton and failed to open it — 108 times per run, each failure swallowed into a mocked-logger `warn()` at the three call sites in `orchestrator-agent.ts` (`addMessage.catch(...)` at lines 2899/3162, `getCodebaseEnvVars` try/catch at line 2443). The stub return values (`Promise.resolve()`, `[]`, `{}`) are exactly what those swallowed-failure paths already produced, so no test's observable behavior changes — only the silent, repeated real I/O disappears. All three specialist reviewers (code, seams, simplify) returned clean or near-clean, and the coordinator's own independent verification reproduces every load-bearing number in the PR body exactly.

**1 blocking · 0 non-blocking** — the one recorded finding is a Suggestion; it is not blocking.

**Validation:** all applicable checks pass; see the coverage table below.

### Findings

|---|---|---|---|
| `R1` | Suggestion | The two new mock blocks are inconsistently minimal — `db/env-vars` stubs only the one reached export, `db/messages` stubs all three including the unreached `listMessages` | OPEN |

<details>
<summary><code>R1</code> — asymmetric minimalism between the two new mock blocks</summary>

**Impact:** None on behavior or safety. `listMessages` (`../db/messages`) is never called by `orchestrator-agent.ts`'s `handleMessage` path in either this file or the sibling `orchestrator-agent.test.ts`, yet it's stubbed anyway, while the `../db/env-vars` block stubs only `getCodebaseEnvVars`, the one export actually reached. A purely stylistic asymmetry within the same 18-line diff.

**Evidence:** `packages/core/src/orchestrator/orchestrator.test.ts:97-109`; `orchestrator-agent.ts` calls only `addMessage` (2200, 2899, 3162) and `getRecentWorkflowResultMessages` (2253) from `../db/messages`, never `listMessages`.

**Required outcome:** None required. Mirroring the sibling file's full mock shape for `../db/messages` is a defensible, cheap-to-keep consistency choice (a future call site added to either file doesn't surprise just one suite) and costs one inert line returning `Promise.resolve([])`. Leaving it as-is is acceptable.

**Disposition:** OPEN, Suggestion only, no follow-up requested. Genuinely optional — recorded so the aggregation is complete, not because it needs action before merge.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `seams` | No additional findings (ran a class enumeration across every test file in `packages/core/src` that mocks `../db/messages` or `../db/env-vars`; found no shape drift and no missing coverage — see round detail) |
| `simplify` | `R1` |

#### Validation

|---|---|---|
| `bun test src/orchestrator/orchestrator.test.ts` (PR head, `packages/core`) | PASS | 75 pass, 0 fail, 142 `expect()` calls, 382ms |
| Same command, `dev`'s pre-PR version of the file swapped in | PASS | 75 pass, 0 fail, 142 `expect()` calls, 3.59s — identical pass/assertion count to the PR head, confirming no test's meaning or shape changed; timing difference matches the PR's directional claim |
| Independent `getDatabase()` call-count instrumentation (temporary `process.stderr.write` inserted into `connection.ts`, reverted after) | PASS | 0 calls with the PR's stubs, 108 calls with `dev`'s version — reproduces the PR's central claim exactly, from a fresh instrumentation the coordinator wrote rather than trusting the PR's own numbers |
| `grep` for any assertion on the previously-swallowed paths (`persist_failed`, `workflow_result_messages`, `addMessage`, `listMessages`, `getRecentWorkflowResult`, `remote_agent_messages`, `codebase_env_vars_load_failed`) | PASS | Zero hits outside the new mock definitions themselves; the file's only two logger assertions (`mockLogger.error` at :1508, `mockLogger.warn` at :1618) are `toHaveBeenCalledWith` on unrelated messages (`error_notification_failed`, `thread_inheritance_failed`), not call counts, so removing the swallowed warns cannot affect them |
| `bun run test` (full `packages/core` suite) | PASS | exit 0, 43 batteries, every battery reports `0 fail` — including `orchestrator.test.ts` and `orchestrator-agent.test.ts` each running as their own separate `bun test` process per the package script, closing the Bun mock-pollution question: the two files never share a process, so the new stubs cannot leak into any other battery |
| `bunx eslint src/orchestrator/orchestrator.test.ts` | NOT APPLICABLE | File is ignored by the repo's own ESLint config for `*.test.ts` in `core` — an existing carve-out, not a gap this diff introduces |
| `bunx prettier --check src/orchestrator/orchestrator.test.ts` | PASS | "All matched files use Prettier code style!" |
| `bun x tsc --noEmit` (`type-check`) | NOT APPLICABLE | `packages/core/tsconfig.json` excludes `**/*.test.ts`; the PR body itself names this as a known, pre-existing hole (`#2407`/step 1 of `#2982`), not caused by this diff |

</details>

Issue #2976, opened 2026-08-29
A transient red check permanently ends a delivery — flip-ready treats one observation as final


Wirasm:
## Problem

A transient red check permanently ends a delivery. `flip-ready` refuses to flip on non-green checks — correctly — the node fails, the run fails, and there is no path back in once the checks later go green. Recovering the delivery requires a human to notice, rerun CI, and resume by hand.

The refusal is right. The finality is not. `flip-ready` treats one observation of a check state as a verdict on a surface that is expected to change: CI is retried routinely, and the repository has a known Windows flake cluster (#2306, #2924) that produces exactly this transient red.

## Observed today, four times

Five dogfood runs of `archon-ship`. Three ended at `flip-ready` with:

```
Bash node 'deliver__flip-ready' failed [exit 1]: flip-ready: refusing to flip with
non-green checks: Test Suite (fail)
test (windows-latest) (fail)
```

In every case the red was a Windows timeout from the #2306 family, on tests the delivery never touched.

The recovery on run `242f94bf8fb4d01ad45129f26a183bb0` (PR #2970) shows the shape exactly:

```
14:53:01  flip-ready  node_skipped   (trigger_rule — upstream red)
          … run failed. PR left draft. ~90 minutes idle. …
          [human reruns the Windows job; it passes]
          [human resumes the run]
16:25:38  flip-ready  node_started
16:25:41  flip-ready  node_completed → https://github.com/coleam00/Archon/pull/2970
```

The flip took **2.7 seconds** once it was allowed to look again. Nothing about the delivery had changed — the same commit, the same PR, the same checks — only that CI had been retried in the interim. The pack was right the whole time; it just had no way to re-observe.

## Why this matters more than the failure rate suggests

`archon-deliver` already suspends on a durable `wait:` while `await-checks` watches CI, and it correctly holds through a pending state. So the run is *architecturally capable* of waiting for a check surface to settle. It just treats the first concluded-but-red observation as terminal, one node later.

The consequence is that autonomous delivery is only as reliable as the flakiest check in the repo. With a known flake cluster open, that is not a theoretical bound — it cost three of five deliveries today, each having already completed implementation, review, correction and validation. All that work is discarded at the final step, for a red the change did not cause.

It also interacts badly with #2960: between them, a run can lose a finished delivery either because the correction loop had nothing left to do, or because an unrelated test timed out on Windows. Neither is a defect in the work being delivered.

## Suggested direction

[...]

PR #2972, opened 2026-08-29
ci: gate workflow fixtures before merge

Wirasm:
## Project check status

The implementation was delivered with an external environment blocker: repository rulesets `15166646` (`dev`) and `15166632` (`main`) are disabled and the active account is not a repository administrator. Attempts to configure the required `Test Suite` check returned GitHub 404. The implementation validation passed, including all 39 fixtures and `bun run validate`; this PR's own CI is still the check that decides the code result. Evidence and the exact permission failure are recorded in the implementation run artifacts.

## Problem and outcome

Fixture drift could merge undetected because the fixture bar ran only in the push-triggered E2E smoke workflow. PR #2863 merged five broken fixtures while CI was green, and PR #2950 nearly repeated the failure with 17 broken fixtures.

- **Outcome:** Relevant pull requests now run every discovered workflow fixture on Ubuntu, and a fixture failure makes the `Test Suite` aggregate fail.
- **Invariant:** The gate uses the maintainer command, `bun run cli workflow test --json`, with no AI credentials or network access.
- **Scope boundary:** The fixture suite stays Ubuntu-only; the existing Windows test leg and every other check remain unchanged.
- **Root cause:** The only CI invocation was owned by `e2e-smoke.yml`, which runs only after pushes to `main` or `dev`.

## Review guidance

- **Feedback requested:** Confirm that the fixture result is correctly made part of the PR-facing `Test Suite` aggregate.
- **Start here:** `.github/workflows/test.yml` — the new `workflow-fixtures` job and aggregate dependency make fixture failures visible before merge.
- **Review order:** Review the fixture job, then its `test-suite` aggregation, then the focused topology test.
- **Lower-attention areas:** Removing the duplicate E2E-smoke step is mechanical; `bun run cli workflow test --json` passed all 39 fixtures.
- **Known risk or uncertainty:** Repository rulesets still need an administrator to require `Test Suite` before GitHub will enforce it as merge-blocking.

## Solution

The fixture bar moves to the PR-triggered `Test Suite` workflow, which already runs for pull requests to `dev` and `main` as well as pushes. A dedicated Ubuntu job follows the existing change decision, installs the frozen lockfile, and runs the same fixture command used locally. `Test Suite` now depends on that job and rejects any non-success result, while documentation-only changes retain its existing successful aggregate path.

The push-only E2E smoke workflow no longer duplicates the command, leaving `test.yml` as its single CI owner. A focused test pins the job's command, conditions, and aggregate propagation.

## Behavior change

[...]

PR #2985, opened 2026-08-29
feat(sdlc): concluded CI results feed each correction round

Wirasm:
## Problem and outcome

The delivery tail consumes CI exactly once, at the terminal wait, where a concluded red is a verdict with no way back into the loop. Checks run on every pushed correction the whole time the loop iterates, and their results sit unread — so an introduced failure that CI can prove in minutes surfaces an hour later as a dead run, and the repository's best self-fixing gate never feeds the fixer.

- **Outcome:** each correction round opens by reading the concluded check state for the run's own PR and hands it to the fix as evidence; an introduced CI failure is fixed in the round that caused it, and the fixer re-reads concluded state once before declaring done so mid-round results are caught at the next turn.
- **Invariant:** the loop never waits on CI anywhere — pending checks are reported and ignored; `await-checks` and the flip preflight stay the sole authority on the flip; every CI-triggered fix still passes through the recheck review inside the same round, never an unreviewed final push.
- **Scope boundary:** `await-checks`, `flip-ready`, the review contract, `implement.md`, and the other tails are untouched. The terminal red-is-terminal semantics decided on #2976 are unchanged.

## Review guidance

- **Feedback requested:** whether `ci-note`'s never-fail contract is stated clearly enough that a future reader doesn't "harden" it into a gate, and whether the fix-prompt language keeps the red-cause discipline unambiguous for CI evidence.
- **Start here:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml` — the `ci-note` node and the two new paragraphs in the fix work order are the whole change.
- **Review order:** the corrections loop in `archon-deliver.yaml`; the four fixture stubs; the regenerated bundled defaults.
- **Lower-attention areas:** `bundled-defaults.generated.ts` is mechanical regeneration, verified by `check:bundled`.
- **Known risk or uncertainty:** evidence staleness — checks describe the previously pushed head, which is exactly the intended evidence; the prompt says so and binds the observed head SHA into the text.

## Solution

A new `ci-note` bash node opens each correction round: it reads the concluded check state for the run-owned PR (pinned to the origin repo, selected by the recorded PR number, never the ambient branch), reports concluded non-green checks with the head SHA they describe, counts still-running checks without waiting on them, and cannot fail the round — a read that fails reports that fact and the round proceeds on the review findings alone. `gh` exits non-zero when checks are failing while still printing them, so the tolerated reads are the design: the probe's contract is "describe what is known, refuse nothing."

The fix node's work order gains the probe's output as an evidence block plus the discipline for it: an introduced CI failure is part of the round's work exactly like a review finding; red the branch cannot have caused is classified with evidence under the existing red-cause rules; never wait on running checks; re-read concluded state once before declaring done. Because injection happens at the round boundary — after the previous recheck landed — CI that concluded during review arrives with the review's findings, and because the fixer's loop re-enters its prompt each turn, late results are self-served without any engine channel for mid-turn messaging.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | CI results accumulate unread while the correction loop iterates; first consumption is the terminal wait. | Each round starts with the concluded CI state as fix evidence; the fixer also re-checks once before declaring done. |
| **Failure behavior** | An introduced CI failure survives every correction round and kills the run at `await-checks`, discarding the delivery. | The same failure enters the round that caused it and is fixed and re-reviewed there; the terminal wait semantics are unchanged for whatever still survives. |

## Validation

- `bun run cli workflow test sdlc --json` — 44 passed, 0 failed (the `converge` fixture's `reached` list now pins `ci-note` scheduling).
- Mutation check, wiring seen red: renaming the `ci-note` node fails workflow validation loudly ("Node 'fix' depends_on unknown node 'ci-note'") and 28 fixtures go red; restored, 44/44 green.
- `bun run generate:bundled` + `bun run check:bundled` — up to date.
- `bun run validate` — full gate green.

Relates to #2976 (this is the "feed CI into the loop" counterpart of the terminal-red decision recorded there).

## Summary by CodeRabbit

* **New Features**
  * Correction workflows now provide current CI status to fix steps, including failing, pending, and completed checks.
  * Workflows can proceed with review findings when CI evidence is unavailable or when failures are inherited from an earlier commit.
  * CI status information no longer blocks correction rounds when checks cannot be read.
  * Delivery completion now reports only verified pull request links and provides clearer status when operations fail.

* **Tests**
  * Added coverage for passing, pending, failing, unavailable, and inherited CI states during correction and recheck flows.


Wirasm:
## Ready to merge

Round 2 verifies the single correction commit (`66faf1029`) against round 1's two Important findings and one Suggestion. All three are fixed. The correction replaces `ci-note`'s three independent, unbounded `gh` reads with one `gh pr checks` call captured to a temp file, guarded by an in-script 45-second watchdog (background job + `sleep 45` + `kill`) and a `timeout: 90000` node-level backstop, with total/failing/pending all derived locally from that single captured document. I reproduced every relevant branch live — not just re-read the diff — including forcing a genuine 45-second hang with a stub `gh` that never returns, and a stub `gh` that exits non-zero while still printing valid data. All five branches (green/pending, mixed red+pending, read failure, real hang, no-checks-configured) complete with `exit 0` and the correct message; the hang branch measured exactly 45 seconds end to end, safely under both the node's own 90s override and the engine's 120s default. `bun run cli workflow test sdlc --json` is 44/44 and `bun run check:bundled` is up to date on the corrected head.

**Validation:** `bun run cli workflow test sdlc --json` 44/44; `bun run check:bundled` up to date; five live branch reproductions of the corrected `ci-note` script (see `R1`/`R2` disposition below).

**Resolved:** 3 · **Tracked follow-ups:** none

### Findings

|---|---|---|---|
| `R1` | Important | `ci-note`'s engine-level timeout skip-cascades into a full correction-loop failure | FIXED |
| `R2` | Important | A partial `gh pr checks` failure mid-script reports "No concluded failures" indistinguishably from a truly clean state | FIXED |
| `R3` | Suggestion | The node's own comment overstates what it establishes | FIXED |

<details>
<summary><code>R1</code> — <code>ci-note</code>'s timeout defeats "cannot fail the round"</summary>

**Impact (as found):** A `gh` call inside `ci-note` that hung rather than erroring could exceed the node's inherited 120s default subprocess timeout with no override in place; the engine-killed node would cascade a full correction-loop failure through `fix`'s `depends_on: [ci-note]` (default `all_success`) and the loop's `until_bash` field access on a skipped `recheck-action`.

**Evidence (round 1):** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:198-200` (pre-correction), traced through `packages/workflows/src/dag-executor.ts:1940-1958` (trigger-rule skip), `packages/workflows/src/output-ref.ts:332-334` (`producer-not-run` throw on field access to a skipped node), and `dag-executor.ts:5312-5318` (`until_bash`'s `requiredContext` turning that throw into a loud loop-group failure).

**Correction:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:178-219` (head `66faf1029`) — `ci-note` now runs its single `gh pr checks` call in the background, races it against an in-script watchdog (`sleep 45` then `kill`), and additionally carries an explicit `timeout: 90000` node-level override as a backstop. A hang no longer reaches the engine's kill path at all: the script's own watchdog reliably terminates the hung read and the node still exits 0.

**Verification:** Reproduced live, not just read. Built a standalone copy of the corrected script and ran it against: (1) a stub `gh` that `sleep 999999`s — the watchdog fired at exactly 45s wall-clock, killed the hung process (`RC=143`), and the script printed "the check read failed or timed out" and exited 0; (2) a stub `gh` that exits 1 immediately with no output — correctly falls into the same "read failed" branch; (3) a stub `gh` that prints valid check data but exits 8 (non-zero) — correctly treated as real data, not a failure (closes the `R2` half of this mechanism too); (4) real live reads against PR #2985 (mixed pass/pending) and PR #2972 (mixed fail/pending) on this repository — both produce correct, well-formed evidence text; (5) a stub `gh` that exits 0 with zero check entries — correctly reports "no checks reported." No leftover `mktemp` temp files after any run (`find "$TMPDIR" -name 'tmp.*' -newermt '-5 minutes'` empty post-test). Confirmed `bash`'s stderr job-control notice ("Terminated: 15") on the kill path does not leak into the node's captured output — bash-node output is stdout-only (`dag-executor.ts:3729,3851`), stderr is logged separately and never spliced into `$ci-note.output`.

**On the watchdog shape (simplify judgment):** Confirmed this environment's stock macOS has no `timeout`/`gtimeout` on `PATH` outside Homebrew coreutils (`/usr/bin/timeout` does not exist), and no other node in this pack relies on GNU `timeout`. A single-command `timeout 45 gh ...` would be simpler but is not a portable primitive this codebase can assume, given its existing Windows/macOS cross-platform constraints (see recent `#2946`/`#2954`-class fixes elsewhere in this repo). The background-job-plus-watchdog pattern uses only POSIX/bash builtins (`&`, `wait`, `kill`, `sleep`, `mktemp`) and is the smallest portable shape that achieves the bound; it is not defense-in-depth against a hypothetical, it directly closes the exact hang mechanism this finding traced. Not over-machinery.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier` (round 1); verified by `review-2985` (round 2).

**Disposition:** FIXED.

</details>

<details>
<summary><code>R2</code> — a partial mid-script read failure reports false-clean CI evidence</summary>

**Impact (as found):** Three independent `gh pr checks` calls (`TOTAL`, `FAILING`, `PENDING`), each tolerant of failure independently — a failure in the `FAILING`-deriving call alone, after `TOTAL` had already succeeded, produced an empty `FAILING` string indistinguishable from genuinely clean CI, silently misleading the fixer.

**Evidence (round 1):** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:182-191` (pre-correction), reproduced by `prp-core:code-simplifier`.

**Correction:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:180-219` — one `gh pr checks` call now captures the full per-check state (`PENDING`/`GREEN`/`RED <name> (<bucket>)` lines) into a single temp file; `FAILING` and `PENDING` are both derived locally from that one captured document via `sed`/`grep`, never a second network read. A read failure (non-zero exit with no captured output, or a watchdog kill) now uniformly falls into the single "read failed or timed out" branch rather than partially succeeding.

**Verification:** Reproduced live with a stub `gh` that exits non-zero with valid output on stdout (`RC=8`, three data lines) — confirmed the script correctly parses and reports the data rather than treating the non-zero exit as a failure, which is precisely the interpretation this fix depends on to avoid conflating "gh's exit code alone" with "the read failed." Reproduced the genuine failure case (non-zero exit, empty output) separately and confirmed it correctly falls into the honest "no evidence" branch instead of any partial/inconsistent state.

**Found by:** `prp-core:code-simplifier` (round 1); verified by `review-2985` (round 2).

**Disposition:** FIXED.

</details>

<details>
<summary><code>R3</code> — the node's comment overstates what it establishes</summary>

**Impact (as found):** The comment claimed "three tolerated reads" (there were four/five) and that `gh` "exits non-zero when checks are failing while still printing them" — empirically false for the `--json`/`--jq` invocation shape actually used (verified against live PRs with failing checks on `gh` 2.92.0: exit 0 with `--json`, exit 1 without).

**Correction:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:166-177` — the comment no longer states a specific count of tolerated reads, and now says the exit code is "read together with the captured output, never alone... this gh form can exit zero on failing checks, so non-zero with output is data, and only killed-or-empty means the read itself failed" — a hedged, accurate claim that matches both the empirically observed `gh` behavior and the code's actual interpretation logic (verified in `R2`'s non-zero-with-output reproduction above).

**Found by:** `prp-core:code-simplifier` (round 1); verified by `review-2985` (round 2).

**Disposition:** FIXED.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 (round 1) |
| `seams` | R1 (round 1) |
| `simplify` | R1, R2, R3 (round 1) |

Round 2 is correction verification only, performed directly by the coordinator against the bounded diff (`52a97f24d`..`66faf1029`) and the three round-1 findings; no scope had unaffected findings to re-run, and no correction-caused defect was found outside what R1–R3 already covered.

#### Validation

|---|---|---|
| `bun run cli workflow test sdlc --json` | PASS | 44 passed, 0 failed (fresh `bun install`, corrected head) |
| `bun run check:bundled` | PASS | "bundled-defaults.generated.ts is up to date (55 commands, 32 workflows, 11 scripts)." |
| Live branch reproduction of corrected `ci-note` (5 branches) | PASS | green/pending, mixed red+pending (live PRs #2985, #2972), read failure, 45s real hang via watchdog, no-checks-configured — all exit 0 with correct evidence text |
| `bun run validate` (full gate) | NOT RUN | Change is YAML + one mechanically-regenerated file; no TypeScript source touched beyond the verified-in-sync generated file |

</details>

PR #2987, opened 2026-08-29
fix(cli): the workflow-wait spec no longer guesses that a detached run is ready

Wirasm:
## Problem and outcome

`workflow-wait.integration.spec.ts` slept a fixed 1500 ms before cancelling a detached
run, hoping two unrelated things had finished inside it. It is the worst cross-branch
flake in the repository: 12 Windows failures across 8 unrelated branches, 5687 ms to
17640 ms, and the file cost 22.18 s for 6 tests at 31% CPU.

- **Outcome:** the case waits on two observations instead of a clock — the run row
  reaching `running`, and the waiter's own attachment line. Injecting a 2.5 s owner
  startup reproduces the CI failure exactly on the old file; the fixed file survives a
  10 s one.
- **Invariant:** the cancellation still has to land while the waiter is watching, so
  what ends the wait is a wake and not a durable read of an already-cancelled row.
- **Scope boundary:** all six cases still run `wait` as a real second OS process against
  a real SQLite database. Nothing became a unit test, and no timeout was widened.
- **Root cause:** 1500 ms was not enough on a contended Windows runner for the forked
  owner to get from a `pending` row to a running one. `workflow cancel` refuses a
  `pending` run outright, and `taskkill /PID … /T /F` walking a tree that is still
  spawning fails on descendants that are already gone. Both messages appear verbatim in
  the CI logs.

## Review guidance

- **Feedback requested:** whether the stderr progress line is the right shape for a
  public CLI contract.
- **Start here:** `packages/core/src/services/run-attention-watch.ts:81` — the new
  `onAttached` option is the load-bearing idea; everything else consumes it.
- **Review order:** the core option, then `announceWaitAttached` in
  `packages/cli/src/commands/workflow.ts:3606`, then `startWait` in the spec, then the
  rewritten case at `packages/cli/src/commands/workflow-wait.integration.spec.ts:412`.
- **Lower-attention areas:** the three `toHaveBeenCalledWith` updates in
  `workflow.test.ts` are mechanical — the option is now always passed.
- **Known risk or uncertainty:** `--json` now writes one line to stderr. stdout still
  carries exactly one document, and under `2>&1` the extra line is itself valid JSON, so
  a `jq` stream consumer is unaffected. A consumer that treats any stderr output as
  failure would see a change.

## Solution

The sleep stood in for a signal that did not exist. `waitForRunAttention` is documented
as durable rather than live-only: a host that attaches after a transition gets the same
answer as one that was watching when it happened. That makes the moment the watch begins
the one thing a caller cannot infer, and the only thing the sleep was really buying.

`waitForRunAttention` now takes `onAttached`, called once with the status the opening
read saw, when that read finds nothing to report. It does not fire when the run already
has an answer, because then nothing was ever waited for. `archon workflow wait` turns
that into one line on stderr, which also closes a real gap: until now the command
printed nothing between launch and settling, so neither a person nor a host could tell
an attached wait from one still resolving the id.

The test then needs no clock. It waits for the run row to reach `running` — the forked
owner starts its control endpoint before it executes anything, so that single reading
also covers the endpoint `cancel` has to reach — and then waits for the attachment line
before cancelling.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `archon workflow wait` printed nothing until it settled. | It writes one line to stderr once the watch begins: a plain sentence, or `{"ok":true,"action":"wait","runId":"…","result":"waiting","observedStatus":"running"}` under `--json`. stdout is unchanged. |
| **Failure behavior** | A wait that died during startup was indistinguishable from one still waiting. | The absence of that line now separates the two. |

## Validation

- `bun run validate` — exit 0, zero failing tests across every package.
- `bun test src/commands/workflow-wait.integration.spec.ts` ×20 sequential — 20/20 green,
  19.58 s to 21.78 s. The file drops from 22.20 s to 20.00 s, and the cancelled-run case
  from 5.08 s to 3.09 s (junit reporter, same machine).
- **Reproduction.** Injecting `await new Promise(r => setTimeout(r, 2500))` into the
  detached owner's startup, immediately after `startDetachedRunControlServer`, models a
  contended runner. On the pre-fix file that fails with the exact CI message:
  `cancel failed: Error: Cannot actively cancel run with status 'pending'. Only a running
  detached CLI run has live work to stop.` — byte-identical to jobs 99114831350 and
  99083403564. The fixed file passes against the same probe, and against a 10 s one.
- **Seen red, wake path.** Returning `deadline` immediately after the opening read, so
  only a durable read can answer, fails the case with
  `wait exited 3, expected 0 — payload: {…"result":"deadline"…}`. The case cannot be
  satisfied without a real wake.
- **Seen red, attachment gate.** Removing the `onAttached` call fails the case with
  `wait exited without announcing that it attached`, rather than hanging.
- **The coupling was real.** With the waiter's first read delayed 8 s, the pre-fix
  ordering cancelled at 1.5 s and the waiter took the durable-read path — it never woke
  on anything, and the case named "wakes on a cancelled run" still passed green. Under
  the same delay the fixed case still exercises the wake, because it waits for the
  attachment instead of guessing at it.
- **Not verified:** no Windows machine here, so the reproduction is an injected startup
  delay standing in for a contended runner rather than the runner itself. Six copies of
  the spec in parallel, three rounds, stayed green on macOS both before and after the
  fix — this flake is 113-of-122 Windows, and local load does not reach it.

## Delivery considerations

|---|---|---|
| Compatibility | `--json` stdout is unchanged at exactly one document; the new line is on stderr. | `workflow.test.ts` asserts `stdoutSpy.mock.calls` has length 1 while the envelope lands on `console.error` |
| Documentation | The `workflow wait` reference now describes the progress line and where it goes. | `packages/docs-web/src/content/docs/reference/cli.md` |

## Links

- Relates to #2982 (step 3: the `workflow-wait.integration.spec.ts` sleep, top offender 1
  in the audit's bucket 3). Other steps remain open.

## Summary by CodeRabbit

- **New Features**
  - `workflow wait` now announces when it has attached to a run and begun waiting.
  - Human-readable output includes the run and status; JSON mode emits a waiting acknowledgment on stderr while preserving stdout for the final result.
  - Runs already requiring attention continue to respond immediately without an attachment message.
  - Undeliverable output is now reported as a command failure.

- **Documentation**
  - Updated CLI documentation to describe the waiting acknowledgment and output behavior.

### DIFF
--- a/packages/cli/src/commands/workflow-wait.integration.spec.ts
+++ b/packages/cli/src/commands/workflow-wait.integration.spec.ts
@@ -104,6 +104,19 @@ function expectWaitExit(
 }
 
 interface PendingWait {
+  /**
+   * Resolve once the wait has said, on stderr, that it is watching the run.
+   *
+   * This is the ordering a caller cannot otherwise establish. `waitForRunAttention`
+   * is durable, so a waiter that attaches after a transition reports exactly the same
+   * payload as one that was watching when it happened — which means a test can only
+   * claim the wake half of the contract if it knows the watch had begun. Sleeping
+   * first only made that likely; this makes it true or fails loudly.
+   *
+   * Rejects with everything the process wrote when it exits without announcing,
+   * rather than leaving the caller to time out on a promise that will never settle.
+   */
+  attached(): Promise<{ observedStatus: string }>;
   settled(): Promise<{ exitCode: number; payload: Record<string, unknown> }>;
 }
 
@@ -127,12 +140,66 @@ function startWait(fixture: Fixture, runId: string, timeoutSeconds: number): Pen
       stderr: 'pipe',
     }
   );
+
+  let announce: ((progress: { observedStatus: string }) => void) | undefined;
+  let abandon: ((error: Error) => void) | undefined;
+  const attached = new Promise<{ observedStatus: string }>((resolve, reject) => {
+    announce = resolve;
+    abandon = reject;
+  });
+  // A test that never asks about the attachment must not turn a legitimately
+  // unannounced exit into an unhandled rejection.
+  attached.catch(() => undefined);
+
+  // One drain owns stderr: it hands the progress line over as it arrives and keeps
+  // the whole text for `settled()`'s diagnostics. Two readers of the same stream
+  // would leave which one saw what up to scheduling.
+  const stderrText = (async (): Promise<string> => {
+    const decoder = new TextDecoder();
+    let text = '';
+    let scanned = 0;
+    const scan = (): void => {
+      for (
+        let newline = text.indexOf('\n', scanned);
+        announce && newline !== -1;
+        newline = text.indexOf('\n', scanned)
+      ) {
+        const line = text.slice(scanned, newline).trim();
+        scanned = newline + 1;
+        // Scan lines rather than assume the first one: `--json` silences logging, so
+        // the progress envelope is normally alone here, but a diagnostic on stderr
+        // must not be mistaken for it.
+        let parsed: unknown;
+        try {
+          parsed = JSON.parse(line);
+        } catch {
+          continue;
+        }
+        const progress = parsed as { result?: unknown; observedStatus?: unknown };
+        if (progress.result === 'waiting' && typeof progress.observedStatus === 'string') {
+          announce({ observedStatus: progress.observedStatus });
+          announce = undefined;
+        }
+      }
+    };
+    for await (const chunk of child.stderr) {
+      text += decoder.decode(chunk, { stream: true });
+      scan();
+    }
+    text += decoder.decode();
+    scan();
+    // Only when nothing was announced — `announce` is cleared the moment it fires.
+    if (announce) abandon?.(new Error(`wait exited without announcing that it attached: ${text}`));
+    return text;
+  })();
+
   return {
+    attached: (): Promise<{ observedStatus: string }> => attached,
     async settled(): Promise<{ exitCode: number; payload: Record<string, unknown> }> {
       const [exitCode, stdout, stderr] = await Promise.all([
         child.exited,
         new Response(child.stdout).text(),
-        new Response(child.stderr).text(),
+        stderrText,
       ]);
       // `--json` silences logging, so stdout is exactly one (pretty-printed) document.
       let payload: Record<string, unknown>;
@@ -202,8 +269,25 @@ function readRunId(archonHome: string, workflowName: string): string | undefined
   }
 }
 
+/** The status of one run, read straight from the database file. */
+function readRunStatus(archonHome: string, runId: string): string | undefined {
+  const databasePath = join(archonHome, 'archon.db');
+  if (!existsSync(databasePath)) return undefined;
+  const database = new Database(databasePath, { readonly: true });
+  try {
+    return database
+      .query<{ status: string }, [string]>(
+        `SELECT status FROM remote_agent_workflow_runs
+         WHERE id = ?`
+      )
+      .get(runId)?.status;
+  } finally {
+    database.close();
+  }
+}
+
 /**
- * Poll `read` until it produces a value.
+ * Poll `read` until it produces a value, naming `what` if it never does.
  *
  * The catch is load bearing, not defensive: an owner process creates `archon.db`
  * BEFORE it applies the schema, so a reader that opens the file inside that window
@@ -212,7 +296,7 @@ function readRunId(archonHome: string, workflowName: string): string | undefined
  * (#2306). `workflow-terminal-event.integration.spec.ts` catches for exactly this
  * reason; a bare loop here would surface a startup race as a test failure.
  */
-async function waitFor<T>(read: () => T | undefined, timeoutMs = 30_000): Promise<T> {
+async function waitFor<T>(what: string, read: () => T | undefined, timeoutMs = 30_000): Promise<T> {
   const deadline = Date.now() + timeoutMs;
   let lastError: unknown;
   while (Date.now() < deadline) {
@@ -225,7 +309,7 @@ async function waitFor<T>(read: () => T | undefined, timeoutMs = 30_000): Promis
     await Bun.sleep(50);
   }
   const detail = lastError instanceof Error ? `: ${lastError.message}` : '';
-  throw new Error(`Timed out waiting for the run row${detail}`);
+  throw new Error(`Timed out waiting for ${what}${detail}`);
 }
 
 describe('archon workflow wait against a detached run', () => {
@@ -292,7 +376,9 @@ describe('archon workflow wait against a detached run', () => {
 
     // Discovering the id is test setup, not the contract under test — the launcher
     // above has no `--detach --json` ack to carry one.
-    const runId = await waitFor(() => readRunId(fixture.archonHome, 'wait-gated'));
+    const runId = await waitFor('the gated run row', () =>
+      readRunId(fixture.archonHome, 'wait-gated')
+    );
     activeRunIds.add(runId);
 
     // Attached while the warm-up node is still running, so the gate is a WAKE.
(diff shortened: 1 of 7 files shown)


Wirasm:
## Ready to merge

Both round-1 findings are fixed by the same correction: a new `writeStderr` (`stdout.ts`)
gives stderr the same guaranteed-delivery write `writeStdout` already had, `announceWaitAttached`
now uses it instead of `console.error`, and `onAttached` is awaited end-to-end so an undeliverable
announcement fails the wait loudly (`exit 1`, `{ok:false, error:"write EPIPE"}`) instead of
vanishing. I independently verified the two claims underneath this design, not just the tests:
`console.error` to a stderr pipe whose reader has closed really is a silent no-op under Bun (the
process runs to completion, exit 0, nothing lost visibly), while `process.stderr.write(text,
callback)` — what `writeStderr` uses — surfaces `EPIPE` through its callback on the identical
scenario. I also reverted `announceWaitAttached` back to `console.error` myself and reran the
three stderr-dependent tests: all three fail immediately (two on an unspied write target, one on
`expect(code).toBe(1)` receiving `0`) — the seen-red claim is real, not asserted. All CI checks
now pass, including the Windows job that is the entire reason this PR exists.

**Validation:** all run and green — see coverage table. CI on the reviewed head: all 8 checks
pass, including `test (windows-latest)`.

**Resolved:** 2 (`R1`, `R2`)

### Findings

|---|---|---|---|
| `R1` | Suggestion | `announceWaitAttached` used `console.error` for a success-path notice | FIXED |
| `R2` | Suggestion | The stderr `--json` line could silently vanish under an early-closed reader | FIXED |

<details>
<summary><code>R1</code> — <code>announceWaitAttached</code> logged a routine notice at error level</summary>

**Disposition:** FIXED. The correction didn't just swap the log level (`console.error` →
`console.warn`) — it removed `console.*` from this path entirely. `announceWaitAttached`
(`workflow.ts:3611-3626`) now builds the line and returns `writeStderr(...)`
(`utils/stdout.ts:58-60`), a dedicated guaranteed-delivery primitive parallel to `writeStdout`.
The original convention question (should this be `warn` or `error`?) is moot — this is no longer
a `console.*` call at all, and the new doc comment (`workflow.ts:3606-3609`) explains why:
*"Through `writeStderr` rather than `console.error`/`console.warn`, and awaited."*

</details>

<details>
<summary><code>R2</code> — the stderr <code>--json</code> line had no delivery-failure signal</summary>

**Disposition:** FIXED, verified two ways.

1. **The new test passes on the real code.** `workflow.test.ts`'s `fails the wait when the
   attachment line cannot be delivered` mocks `process.stderr.write` to invoke its callback with
   an `EPIPE` error and asserts `workflowWaitCommand` returns `1` with
   `{ ok: false, action: 'wait', error: 'write EPIPE' }` on stdout. Ran it standalone: 1 pass.
2. **I independently reproduced the underlying claim**, not just trusted the test. Two small
   scripts under Bun 1.3.11, both writing repeatedly to stderr while `| head -c 1` closes the
   reader after one byte: the `console.error` version ran to completion, exit 0, with no visible
   error anywhere — a genuine silent loss. The `process.stderr.write(text, callback)` version
   surfaced `{"writeError":"EPIPE: broken pipe, write","code":"EPIPE"}` through its callback on
   the very first write after the reader closed. This matches the new `writeStderr` doc comment's
   claim (`utils/stdout.ts:49-53`) exactly.
3. **I reverted the fix myself** (`announceWaitAttached` back to a bare `console.error`, dropping
   the `writeStderr` await) and reran the three affected tests: all three failed immediately —
   two because the test now spies on `process.stderr.write` specifically (so a `console.error`
   call is invisible to it, proving the test pins the *mechanism*, not just "some text landed on
   stderr"), and the EPIPE test failed with `Expected: 1, Received: 0` (the mutated code no longer
   detects or reacts to the failure). Restored the file afterward; worktree confirmed clean before
   re-running green.

**Design judgment — is failing the whole wait the right consequence?** This was the substantive
question behind the correction, not just the mechanism. I walked the realistic invocation shapes:

- **Terminal, `2>/dev/null`, or a redirect to a regular file** never produce `EPIPE` (confirmed
  by the team lead's framing and consistent with how the pipe-vs-file asymmetry already documented
  in `stdout.ts`'s header works) — the overwhelming majority of real usage is unaffected.
- **A host that reads `child.stderr` for the process's full lifetime** — which is what this
  repo's own integration spec does, and the natural way to consume a subprocess's streaming
  output — never closes the reader early, so it never triggers this path either.
- The only realistic trigger is a host that opens a stderr **pipe** specifically, then explicitly
  closes just that reader while leaving the child process running — an unusual, deliberate choice
  rather than an ordinary consumption pattern. When it does happen, the design is proportionate
  because the wait is durable: the CLI docs and `run-attention-watch.ts`'s header both establish
  that a fresh `workflow wait <run-id>` re-attach after ANY point returns the correct current
  answer, so the cost of this failure mode is "the host notices exit 1, retries the wait, and
  gets an immediate correct answer" rather than silent data loss or a stuck host. Given the
  entire point of `onAttached` is to carry ordering information a host cannot otherwise infer, a
  loud, retryable failure is more consistent with this repo's stated "fail early on unsupported,
  unsafe states" preference than the log-and-continue behavior I originally suggested — that
  alternative would have left the host in exactly the ambiguous state the feature exists to
  eliminate, only silently.
- **Residual, non-blocking observation:** the EPIPE-detection path is exercised only through a
  mocked `process.stderr.write` in the unit suite. The real-subprocess integration spec
  (`workflow-wait.integration.spec.ts`) is untouched by this correction and doesn't probe a
  closed-stderr-reader scenario, so this new failure path has no evidence from a real OS pipe on
  any platform, Windows included — the same category of gap the original PR already disclosed
  ("Not verified: no Windows machine here") for the timing fix. Not a finding: nothing in the
  round-1 or round-2 diff claims Windows verification for this path, and I have no Windows
  machine either. Recording it so the author and any future Windows-specific flake investigation
  has the context.

</details>

<details>

#### Reviewer coverage (round 1, scopes preserved — no new scope run this round)

| Scope | Result |
|---|---|
| `code` | R1 (now fixed) |
| `seams` | R2 (now fixed) |
| `simplify` | No findings (round 1) |

This round is a correction-verification pass: the operator applied both round-1 suggestions plus
a design decision (awaited `onAttached`, fail-the-wait-on-delivery-failure). I verified the fix
directly — reran affected tests, reproduced both the underlying primitive-level claims and the
"revert to `console.error`" seen-red claim, and independently judged the design decision — rather
than re-dispatching the three scope agents, since no new surface outside the two closed findings
was touched (confirmed by diff: only `workflow.ts`, `workflow.test.ts`, `stdout.ts`,
`run-attention-watch.ts`, `run-attention-watch.test.ts`, and one line of `cli.md` changed between
rounds — the `workflow-wait.integration.spec.ts` file from round 1 is untouched).

#### Validation

|---|---|---|
| `bun test src/services/run-attention-watch.test.ts` (core) | PASS | 23 pass, 0 fail |
| `bun test src/commands/workflow.test.ts` (cli) | PASS | 336 pass, 0 fail (was 335 in round 1 — one new test) |
| `bun test -t "fails the wait when the attachment line cannot be delivered"` | PASS | 1 pass |
| `bun test src/commands/workflow-wait.integration.spec.ts` (cli, real subprocess) | PASS | 6 pass, 0 fail, 19.96s |
| Reproduction: `console.error` silent-EPIPE-loss claim | CONFIRMED | isolated script: process completes, exit 0, no visible error under a closed-reader pipe |
| Reproduction: `process.stderr.write` EPIPE-surfacing claim | CONFIRMED | isolated script: callback fires `{code:'EPIPE', message:'EPIPE: broken pipe, write'}` on the identical scenario |
| Reproduction: revert `announceWaitAttached` to `console.error` | FAIL as expected | all 3 stderr-dependent tests fail immediately (2 on unspied write target, 1 on `expect(code).toBe(1)` receiving `0`) |
| `bun x tsc --noEmit` (core) | PASS | clean |
| `bun x tsc --noEmit` (cli) | PASS | clean |
| `bun run lint --max-warnings 0` (repo-wide) | PASS | clean across every linted package |
| `bun x prettier --check` (7 changed files) | PASS | all match Prettier style |
| CI on reviewed head `93dc8c19421b1a98ad8c7d6488b4b7caea390472` | PASS | all 8 checks green: `test (ubuntu-latest)`, `test (windows-latest)`, `docker-build`, `postgres-parity`, `schema-upgrade`, `changes`, `Build docs site`, `CodeRabbit` |
| `bun run validate` (full suite) | NOT RUN | covered by the targeted suites, type-check, lint, format above plus full green CI on the reviewed head |

</details>

PR #2988, opened 2026-08-29
chore(sdlc): cut the delivery tail's leftover machinery and name the checks-gate recovery

### LINKED ISSUE #2968: SDLC pack verifies its PR target twice and logs by hand — cut it back to the typed record

Wirasm:
## Problem and outcome

The SDLC pack carried three pieces of machinery that were only ever waiting for something else to land, and one gap the pack could close with a sentence. `flip-ready` kept a hand-rolled log of every command it ran, because nothing else retained a node's output. `archon-deliver` declared `pr_number` and `pr_head` as empty-default inputs, because composition dropped `sync-pr-body`'s own `with:` map and then demanded the caller supply names no caller could know. Two `output_type` labels named artifact types nothing selected. All three reasons are gone now, so this is one pass over the delivery tail that deletes them. Separately, a delivery that dies on a red check reported the failing checks and stopped, leaving the way back — re-run the check, resume the run — as tribal knowledge.

- **Outcome:** the log, the decoy inputs and the two labels are gone; a run that dies at either checks gate now names its own recovery; and the ready flip keeps every command's raw output off its own streams, which is what the log was really doing.
- **Invariant:** everything `flip-ready` actually guards still guards it — the origin-shape refusal, the non-green preflight, the flip, and the draft read-back — the raw origin URL still never reaches printed output, and the node's stdout is still the delivered URL alone.
- **Scope boundary:** no retry, wait, or re-probe behavior was added anywhere. That was weighed and rejected on #2976: a concluded red check does not re-run itself, so waiting on one holds a worker slot for an action nobody was asked to take. The autonomous half of #2976 stays open for the attention lane, which is why this closes only #2968.

## Review guidance

- **Feedback requested:** whether each deletion leaves the survivors' comments truthful, and whether the stream discipline in the flip is drawn at the right line — see the known risk below.
- **Start here:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:296` — the `flip-ready` body, where the log came out, the capture discipline went in, and the recovery sentence was added.
- **Review order:** `archon-deliver.yaml` (the node body, then the deleted inputs), the pack `README.md` (the two rules the node now demonstrates), `archon-pr.yaml`, `deliver/scripts/check-ci.py`, then `bundled-defaults.test.ts`.
- **Lower-attention areas:** `bundled-defaults.generated.ts` is regenerated by `bun run generate:bundled`; `bun run check:bundled` reports it in sync.
- **Known risk or uncertainty:** the two `gh pr view` read-backs send stderr to `/dev/null` rather than capturing and re-emitting it, so a failure there reports the node's own sentence naming the read instead of gh's words. Merging is not an option — `gh` appends its update notice to stderr on success, and a merged read would carry it into the draft state or the delivered URL (there is a mutation test for exactly that). Capturing it separately needs a scratch file, and `mktemp` has no precedent in any pack node and one more thing to be missing on a Windows host. If a reviewer would rather pay that cost to keep gh's words, say so and I will switch.

## Solution

**The log.** The engine now retains what every exec node prints, so `record`, `record_read`, `fail` and `$ARTIFACTS_DIR/flip-ready.log` largely reproduced per-node what the run transcript holds for free. But the helpers had a second job: they kept every wrapped command's raw output off the node's real streams. That one is not covered by retention, and it matters, because `executeBashNode` sends a *succeeding* node's stderr to the operator's chat as the run happens, and that copy is not redacted — redaction covers the retained transcript only. So each call now keeps its own output: the flip captures both streams and re-emits them only on failure, where gh's words are the diagnostic; the two read-backs drop stderr rather than merge it, and each names the read it could not complete; and the tolerated checks preflight drops gh's explanation too, which hides nothing because a `gh` that cannot read this PR's checks cannot flip it either. The credential rule the log existed under is unchanged and now tested mechanically: the origin URL is read exactly once, inside the substitution that normalizes it, and never printed.

**The decoy inputs.** Composition keeps a command node's own `with:` map now, so the empty-default declarations are gone and the composed graphs carry the re-namespaced binding through to the node.

**The typed sidecars.** `output_type: pull-request` and `output_type: public-action` labelled artifacts nothing selects by type. The values these nodes deliver already travel as explicit bindings — `$pr.output.number` into review and the flip, `$flip-ready.output` into `outcome` — so the labels went and the `output_format` record stayed.

**The recovery.** Both gates that can end a delivery on red now name the way back, each in its own voice: `check-ci.py`'s red exit and `flip-ready`'s refusal. Report prose only.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `flip-ready` wrote `$ARTIFACTS_DIR/flip-ready.log`; `archon-pr`'s `pr` node and `flip-ready` wrote typed sidecars under `nodes/` | Neither file is written; the run transcript retains what the node printed |
| **Operator chat on a successful flip** | Silent — every command's output went into the log | Still silent: gh's "no checks reported" and its ready confirmation are captured, not streamed |
| **Failure behavior** | "refusing to flip with non-green checks: …" and "checks not green: …", with no stated way forward | Both add one sentence: if the red is transient or unrelated, re-run the failing check on the pull request and resume the run — the gate reads the state again on resume |

## Validation

- `bun run cli workflow test sdlc --json` — 44 passed, 0 failed, no fixture stubs added or removed. The one reported stub drift (`validate-red.stubs.yaml` leaves `ci-probe` and `flip-ready` unused) reproduces identically on an unmodified checkout, so it is pre-existing.
- Loading the composed `archon-ship`, `archon-stabilize` and `archon-upkeep` graphs through `discoverWorkflows` — zero load errors, `archon-deliver`'s inputs are now `["work","prior_report","errors"]`, and `deliver__sync-pr-body` carries `{"pr_number":"$deliver__pr__pr.output.number","pr_head":"$deliver__pr__pr.output.head"}` — proves the decoy inputs were the only thing composition needed them for, and that the real binding survives without them.
- `grep -rn "public-action"` and `grep -rn "pull-request"` across the repository before the cut — the only hits were the two declarations, the generated bundle, and the two test assertions on them. `latestNodeArtifactOfType` has no production call site (only its own unit test), and `readNodeArtifacts`' one caller lists prior artifacts for `persist_session` recovery, which no pack workflow uses. Nothing selected either type string.
- `bun test src/defaults/` in `packages/workflows` — 48 pass. The stream-discipline scenario runs the node body as a process and asserts stdout is the URL alone and stderr is empty, because the dry-run simulator discards a succeeding node's stderr. It was seen red against five mutations: streaming any one of the four gh/git calls, and merging a read's stderr into its value. The origin guard, the non-green refusal and the recovery sentence were each seen red too.
- `bun run validate`, `bun run lint`, `bun run type-check`, `bun run format:check`, `bun run check:bundled` — all pass.
- **Not verified:** no live delivery was run against a real GitHub pull request; the flip path is covered by the fake-binary tests above rather than an end-to-end flip.

## Links

- Closes #2968
- Related #2976

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved delivery workflow reliability by validating and normalizing repository information before marking a pull request ready.
  - Prevented readiness changes when repository details cannot be resolved or required checks are not passing.
  - Prevented command-line noise from affecting successful delivery results.
  - Delivered pull request links more consistently after successful completion.

- **Documentation**
  - Clarified that workflow output is retained as execution evidence.
  - Added guidance for re-running failed checks and resuming the workflow so check status is evaluated again.
  - Documented how command output is communicated to operators during workflow execution.


Wirasm:
## Ready to merge

Round 1 found one Important, OPEN finding (`R1`): `flip-ready`'s deleted logging wrapper had a second job the PR's own reasoning didn't account for — keeping every wrapped `gh` command's raw stdout/stderr out of the node's real streams. Without it, `gh`'s own diagnostic text reached the node's real stderr on the ordinary no-CI success path, and a pre-existing, unrelated engine code path broadcasts a succeeding node's stderr live to the operator's chat, unredacted.

The correction (`f5f00743502d153c9c337e453e03dac37a2c4878`, one commit) closes it correctly, not just cosmetically. Every `gh`/`git` call in the node now keeps its own output: the flip captures both streams and re-emits them only on failure, where `gh`'s own words become the diagnostic; the two post-flip reads that capture a value (`isDraft`, `url`) drop `gh`'s incidental stderr rather than merge it, because a `gh` notice merged into the value would corrupt it, and each names the read it could not complete so neither failure is silent; the tolerated checks preflight drops its explanation too, since a `gh` that cannot read the PR's checks cannot flip it either and the flip below fails loudly carrying `gh`'s words anyway.

I independently reproduced both directions rather than trusting the PR's own test. Extracting the corrected node body verbatim and running it against a fake `gh` that is maximally chatty — it explains the tolerated no-checks case on stderr, prints its own "marked as ready" confirmation, and appends a realistic "update available" notice to every single call — the node's real stdout is exactly the delivered URL and its real stderr is exactly empty (0 bytes). Reverting just the `gh pr ready` capture back to the old unguarded `>&2` pattern immediately reintroduces 64 bytes of `gh`'s raw text on the real stderr under the identical scenario, proving the fix — not something else — is what closes the gap. Separately, merging the draft-read's stderr into its captured value (the alternative the PR's Known-risk section explicitly rejected) reproducibly corrupts `$draft` and makes the node wrongly refuse a PR that was actually flipped successfully — confirming that trade-off reasoning is correct, not just asserted.

The one thing genuinely traded away — `gh`'s own diagnostic words on the two post-flip read-back failures, replaced with an authored sentence naming which read failed — is disclosed candidly in the PR's Known-risk section, with the two alternatives it rejected (merging, which corrupts a value; a `mktemp` scratch file, which has no precedent in this pack and adds Windows-host risk) and an explicit invitation for a reviewer to ask for the costlier option. This is the right trade: my original concern was a *blind*, unexplained `/dev/null` swallow producing a bare, silent failure. That concern is fully addressed — every suppressed stream is paired with a specific, actionable authored message, including for a `git remote get-url origin` failure that had no authored message even in the pre-#2968 code. What remains lost is `gh`'s exact wording on a narrow, post-success confirmation-read failure, where the authored message ("PR was flipped but its draft state could not be read back") is arguably more actionable for an operator than `gh`'s raw error text would have been.

The pack README's new section ("A node's streams are the operator's channel") states the general rule accurately, matches the existing section's voice, and carries no issue numbers or run ids. No stale references to the deleted log file or sidecars anywhere.

**Validation:** all commands pass — see coverage table.

**Resolved:** 1 (`R1`)

### Findings

|---|---|---|---|
| `R1` | Important | `flip-ready`'s deleted logging wrapper also contained `gh`'s raw stdout/stderr; without it, `gh`'s diagnostics reached a live, unredacted chat broadcast on the normal success path | FIXED |

<details>
<summary><code>R1</code> — <code>gh</code>'s raw output reached an unredacted live broadcast (now fixed)</summary>

**Impact (as originally found):** On every `flip-ready` run where `gh` printed anything to its own stderr — proven to happen on the ordinary, documented no-CI success path — that raw text reached the node's real stderr and was broadcast live, unredacted, to the run's actual conversation platform.

**Correction:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:296-364` (`f5f00743502d153c9c337e453e03dac37a2c4878`). Every `git`/`gh` call now keeps its own output off the node's real streams:
- `git remote get-url origin` — `2>/dev/null`, wrapped in `if ! ...; then` with a new authored message ("could not read the origin remote") this failure mode never had before, even pre-#2968.
- `gh pr checks` (tolerated preflight) — `2>/dev/null`; safe to drop, since the script's control flow already handles the empty/non-empty result correctly without needing `gh`'s raw explanation.
- `gh pr ready` — both streams captured together (`2>&1`) into `$FLIP`, discarded on success, re-emitted only inside an authored message (`"flip-ready: the ready flip failed: $FLIP"`) on failure.
- `gh pr view --json isDraft` / `gh pr view --json url` — `2>/dev/null` on each, each wrapped in its own `if ! ...; then` with a message naming exactly which read failed. Stderr is dropped rather than merged here specifically because these two calls capture a *value*, and merging would corrupt it (reproduced below).

**Verified independently, not by trusting the PR's own test.** I extracted the corrected node body into a standalone script and ran it against a fake `gh` that explains the tolerated no-checks case on stderr, prints its own ready-flip confirmation, and appends a realistic `gh` "new release available" notice to every single call (including the successful value-reads):
```
exit: 0
stdout: https://example.com/repo/pull/42        (33 bytes — exactly the URL)
stderr: (0 bytes — completely empty)
```
Reverting only the `gh pr ready` fix back to the pre-correction `gh pr ready ... >&2` pattern, under the identical scenario:
```
exit: 0
stderr (64 bytes): marked as ready for review
                    gh: A new release of gh is available
```
proving the fix is load-bearing. Separately, mutating the draft-read to merge stderr into its value (`2>&1` instead of `2>/dev/null`) reproducibly corrupts `$draft` and makes the node exit 1 with "PR still reports draft after the ready flip" — even though the PR was actually flipped successfully — confirming the PR's stated reason for not merging is a real, reproducible correctness bug, not a hypothetical one.

The PR's own new test (`bundled-defaults.test.ts`, "lets no gh output reach the node streams on the no-CI success flip") runs the node body as a raw subprocess via `execFileAsync` from `@archon/git` (an existing, already-declared dependency — no new package-boundary surface) and asserts the exact same thing: `stdout` is exactly the URL, `stderr` is exactly `''`. Its commit message states it was seen red against five separate mutations (streaming each of the four `gh` calls individually, plus merging a read's stderr into its value) — consistent with what I reproduced.

**Remaining, disclosed trade-off (not a new finding):** the two post-flip read-back failures (`isDraft`, `url`) now report an authored sentence naming the read instead of `gh`'s own diagnostic text. The PR's Known-risk section discloses this explicitly, names why the two alternatives were rejected (merging corrupts the value; a `mktemp` scratch file has no precedent in this pack and adds Windows-host risk), and invites pushback. I agree with the trade: these are narrow, post-success confirmation reads where the authored message is arguably more actionable than `gh`'s raw text, and the codebase's own "smallest truthful system" standard argues against adding new machinery (a scratch file) for marginal diagnostic value on a rare path.

**Found by:** `prp-core:seam-analyzer` (round 1), correction verified independently by the coordinator via direct reproduction and mutation testing.

**Disposition:** FIXED. Verified by direct reproduction against the corrected node body (clean-stream proof) and by reverting the fix (leak reproduction) and by reproducing the value-corruption alternative the PR rejected — not by trusting the commit message or the PR's own test alone.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | Round 1: no findings. Unaffected by the round-2 correction (touches only the `flip-ready` node body and its own test, which was already in scope) |
| `seams` | Round 1: `R1`. Round 2 (coordinator): verified `R1` fixed by direct reproduction and mutation testing — see above |
| `simplify` | Round 1: no findings. The correction stays in scope: same file, same node, no new machinery beyond the fix itself; the new test reuses an existing package dependency (`@archon/git`'s `execFileAsync`) rather than adding one |

#### Validation

Reran in a fresh detached worktree pinned to the correction head (`f5f00743502d153c9c337e453e03dac37a2c4878`), bounding the diff against the round-1 reviewed head (`64ecd501413863307f45fb6e6656c6e47a981b1f..f5f00743502d153c9c337e453e03dac37a2c4878` — one commit, 4 files).

|---|---|---|
| `bun run cli workflow test sdlc --json` | PASS | 44/44 fixtures pass, 0 failed |
| `bun run check:bundled` | PASS | `bundled-defaults.generated.ts is up to date (55 commands, 32 workflows, 11 scripts)` |
| `bun x tsc --noEmit` (packages/workflows) | PASS | Clean, no output |
| `bun test src/defaults/` (packages/workflows) | PASS | 48 tests, 0 failures — includes the new stderr-clean scenario |
| `bun run test` (packages/workflows, full suite) | PASS | 25 groups, ~2637 tests, 0 failures |
| `bun run lint` (repo-wide) | PASS | All packages clean, exit 0 |
| Direct reproduction: corrected node body vs. a maximally chatty fake `gh` | PASS | stdout exactly the URL (33 bytes), stderr exactly empty (0 bytes) |
| Mutation: revert `gh pr ready` capture to old unguarded `>&2` | Leaks (as expected — proves the fix is load-bearing) | 64 bytes of `gh`'s raw text on the node's real stderr under the identical scenario |
| Mutation: merge draft-read's stderr into its captured value | Corrupts the value (as expected — proves the PR's stated reason for not merging) | Node wrongly refuses with "PR still reports draft after the ready flip" despite a successful flip |
| `bun run validate` | NOT RUN | Ran the narrower checks the changed behavior requires instead; nothing in the correction touches other packages |

</details>

PR #2991, opened 2026-08-29
Ensure release artifacts have one checksum each

### LINKED ISSUE #2377: ci: checksums.txt lists archon-web.tar.gz twice (glob already matches the explicit operand)

Wirasm:
## Problem and outcome

The release checksum command named `archon-web.tar.gz` both through `archon-*` and as an explicit operand, producing two checksum rows for the same uploaded asset. Each release artifact now has exactly one checksum entry.

- **Outcome:** The checksum manifest contains one SHA-256 checksum for each of the six release artifacts.
- **Invariant:** `archon-web.tar.gz` remains included in the manifest because it matches `archon-*`.
- **Scope boundary:** The release assets, upload step, and checksum algorithm are unchanged.
- **Root cause:** The web tarball was passed twice to `sha256sum`.

## Review guidance

- **Feedback requested:** Confirm that the glob covers the complete release artifact set without a duplicate web-tarball operand.
- **Start here:** `.github/workflows/release.yml:280` — the release checksum command now passes each matching artifact once.
- **Review order:** Review the workflow change, then the focused regression test in `scripts/release-checksums.test.ts`.

## Solution

Checksum generation now relies solely on `archon-*`. The glob already includes all five platform binaries and `archon-web.tar.gz`, so removing the redundant explicit filename preserves coverage while eliminating the duplicate checksum. The focused test locks that command and the six-artifact manifest expectation.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `archon-web.tar.gz` appeared twice in `checksums.txt`. | Each release artifact appears once in `checksums.txt`. |
| **Failure behavior** | Consumers could receive an ambiguous duplicate checksum entry. | The manifest has one unambiguous checksum per artifact. |

## Validation

- `bun test ./scripts/release-checksums.test.ts` — passed — verifies the checksum command and six-artifact coverage.
- `bun test ./scripts/` — passed (148 tests).
- `bun run validate` — passed, including generated-artifact checks, API and package type checks, lint, formatting, installer tests, and all package and script tests.
- `git diff --check f339df7a6..HEAD` — passed.
- **Not verified:** Nothing material; the focused regression test covers the corrected manifest input and the repository validation suite passed.

## Links

- Closes #2377

## Summary by CodeRabbit

* **Bug Fixes**
  * Release checksum files now include only binary artifacts, preventing the web archive from being included unintentionally.

* **Tests**
  * Added automated validation to ensure every released artifact is covered by exactly one checksum entry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -277,7 +277,7 @@ jobs:
       - name: Generate checksums
         run: |
           cd dist
-          sha256sum archon-* archon-web.tar.gz > checksums.txt
+          sha256sum archon-* > checksums.txt
           cat checksums.txt
 
       - name: Get version
--- a/scripts/release-checksums.test.ts
+++ b/scripts/release-checksums.test.ts
@@ -1,43 +1,38 @@
 import { expect, test } from 'bun:test';
-import { mkdtempSync, readFileSync, writeFileSync } from 'node:fs';
-import { tmpdir } from 'node:os';
-import { join, resolve } from 'node:path';
-import { trackTempRoots } from '@archon/paths/test-utils';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
 
-const trackTempRoot = trackTempRoots();
-const releaseArtifacts = [
-  'archon-darwin-arm64',
-  'archon-darwin-x64',
-  'archon-linux-arm64',
-  'archon-linux-x64',
-  'archon-windows-x64.exe',
-  'archon-web.tar.gz',
-];
+// `* text=auto` checks this workflow out with CRLF on Windows CI. Normalizing here lets every
+// pattern below anchor on LF, so a stray `\r` can never reach a capture.
+const workflow = readFileSync(
+  resolve(import.meta.dir, '../.github/workflows/release.yml'),
+  'utf8'
+).replace(/\r\n/g, '\n');
 
 test('release checksums include each release artifact once', () => {
-  const workflow = readFileSync(
-    resolve(import.meta.dir, '../.github/workflows/release.yml'),
-    'utf8'
+  // The release uploads one binary per build-matrix entry plus the packaged web dist.
+  const binaries = [...workflow.matchAll(/^ +binary: (\S+)$/gm)].map(match => match[1]);
+  const webDist = workflow.match(/^ *-czf dist\/(\S+) /m)?.[1];
+  const checksumOperands = workflow.match(/^ *sha256sum (.+) > checksums\.txt$/m)?.[1];
+
+  // A missed derivation would leave nothing to compare and pass vacuously.
+  if (binaries.length === 0 || webDist === undefined || checksumOperands === undefined) {
+    throw new Error('release.yml no longer declares the artifacts or checksum command read here');
+  }
+
+  const releaseArtifacts = [...binaries, webDist];
+  const operands = checksumOperands.split(/\s+/).map(operand => new Bun.Glob(operand));
+
+  // `sha256sum` writes one row per operand that names a file, so an artifact covered by two
+  // operands is checksummed twice (#2377) and one covered by no operand is missing entirely.
+  const rowsPerArtifact = Object.fromEntries(
+    releaseArtifacts.map(artifact => [
+      artifact,
+      operands.filter(operand => operand.match(artifact)).length,
+    ])
   );
-  const checksumCommand = workflow.match(/^\s*(sha256sum .+ > checksums\.txt)$/m)?.[1];
 
-  expect(checksumCommand).toBe('sha256sum archon-* > checksums.txt');
-  if (checksumCommand === undefined) throw new Error('release workflow has no checksum command');
-
-  const dist = trackTempRoot(mkdtempSync(join(tmpdir(), 'release-checksums-')));
-  for (const artifact of releaseArtifacts) writeFileSync(join(dist, artifact), artifact);
-
-  const result = Bun.spawnSync(['sh', '-c', checksumCommand], {
-    cwd: dist,
-    stderr: 'pipe',
-  });
-  expect(result.exitCode).toBe(0);
-  expect(result.stderr.toString()).toBe('');
-
-  const manifestFiles = readFileSync(join(dist, 'checksums.txt'), 'utf8')
-    .trim()
-    .split('\n')
-    .map(line => line.match(/^[a-f0-9]{64}  (.+)$/)?.[1]);
-  expect(manifestFiles.sort()).toEqual(releaseArtifacts.sort());
-  expect(manifestFiles.filter(artifact => artifact === 'archon-web.tar.gz')).toHaveLength(1);
+  expect(rowsPerArtifact).toEqual(
+    Object.fromEntries(releaseArtifacts.map(artifact => [artifact, 1]))
+  );
 });
--- a/scripts/release-checksums.test.ts
+++ b/scripts/release-checksums.test.ts
@@ -1,6 +1,18 @@
 import { expect, test } from 'bun:test';
-import { readFileSync } from 'node:fs';
-import { resolve } from 'node:path';
+import { mkdtempSync, readFileSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve } from 'node:path';
+import { trackTempRoots } from '@archon/paths/test-utils';
+
+const trackTempRoot = trackTempRoots();
+const releaseArtifacts = [
+  'archon-darwin-arm64',
+  'archon-darwin-x64',
+  'archon-linux-arm64',
+  'archon-linux-x64',
+  'archon-windows-x64.exe',
+  'archon-web.tar.gz',
+];
 
 test('release checksums include each release artifact once', () => {
   const workflow = readFileSync(
@@ -8,18 +20,24 @@ test('release checksums include each release artifact once', () => {
     'utf8'
   );
   const checksumCommand = workflow.match(/^\s*(sha256sum .+ > checksums\.txt)$/m)?.[1];
-  const releaseArtifacts = [
-    'archon-darwin-arm64',
-    'archon-darwin-x64',
-    'archon-linux-arm64',
-    'archon-linux-x64',
-    'archon-windows-x64.exe',
-    'archon-web.tar.gz',
-  ];
 
   expect(checksumCommand).toBe('sha256sum archon-* > checksums.txt');
-  expect(releaseArtifacts.filter(artifact => artifact.startsWith('archon-'))).toEqual(
-    releaseArtifacts
-  );
-  expect(releaseArtifacts.filter(artifact => artifact === 'archon-web.tar.gz')).toHaveLength(1);
+  if (checksumCommand === undefined) throw new Error('release workflow has no checksum command');
+
+  const dist = trackTempRoot(mkdtempSync(join(tmpdir(), 'release-checksums-')));
+  for (const artifact of releaseArtifacts) writeFileSync(join(dist, artifact), artifact);
+
+  const result = Bun.spawnSync(['sh', '-c', checksumCommand], {
+    cwd: dist,
+    stderr: 'pipe',
+  });
+  expect(result.exitCode).toBe(0);
+  expect(result.stderr.toString()).toBe('');
+
+  const manifestFiles = readFileSync(join(dist, 'checksums.txt'), 'utf8')
+    .trim()
+    .split('\n')
+    .map(line => line.match(/^[a-f0-9]{64}  (.+)$/)?.[1]);
+  expect(manifestFiles.sort()).toEqual(releaseArtifacts.sort());
+  expect(manifestFiles.filter(artifact => artifact === 'archon-web.tar.gz')).toHaveLength(1);
 });
--- a/scripts/release-checksums.test.ts
+++ b/scripts/release-checksums.test.ts
@@ -1,0 +1,25 @@
+import { expect, test } from 'bun:test';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+test('release checksums include each release artifact once', () => {
+  const workflow = readFileSync(
+    resolve(import.meta.dir, '../.github/workflows/release.yml'),
+    'utf8'
+  );
+  const checksumCommand = workflow.match(/^\s*(sha256sum .+ > checksums\.txt)$/m)?.[1];
+  const releaseArtifacts = [
+    'archon-darwin-arm64',
+    'archon-darwin-x64',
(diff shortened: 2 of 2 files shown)


Wirasm:
# Review report

## Verdict

Ready — `none`. The correction now executes the workflow's checksum command over a representative six-artifact fixture and verifies the generated manifest contains each expected artifact exactly once.

## Accepted contract

The release checksum manifest must contain exactly one SHA-256 checksum for each of six release artifacts. `archon-web.tar.gz` must be selected through the `archon-*` glob. Release assets, upload behavior, and checksum algorithm remain unchanged. The change must add or retain focused regression protection for the representative six-artifact manifest and its single web-tarball entry.

## Reviewed head SHA

`da15f8f934ee1e1764963c9a582d0016a7b4af3b`

## Findings

### Rejected findings

- **R2** — `sources: [simplify]` — Disproved. The prior suggestion to retain only a literal command assertion would remove the representative-manifest protection required by the accepted contract. The current fixture is necessary and appropriately scoped.

### Suggestions

None.

## Prior findings

| ID | First raised | Sources | Prior claim | Current verdict |
| --- | --- | --- | --- | --- |
| R1 | Round 1 | `seams`, `tests` | The test did not generate or parse `checksums.txt`, so it could not prove a six-entry manifest. | **Fixed.** `scripts/release-checksums.test.ts:27-42` creates the six-artifact temporary `dist`, runs the exact command extracted from `.github/workflows/release.yml`, parses `checksums.txt`, and requires its filenames to equal the expected six with one `archon-web.tar.gz`. The command assertion at `scripts/release-checksums.test.ts:22-25` proves that web tarball remains selected by `archon-*`; `.github/workflows/release.yml:277-281` is the production command. `bun test ./scripts/release-checksums.test.ts` passed: 1 test, 5 expectations. |
| R2 | Round 1 | `simplify` | A literal command assertion alone could replace the fixture. | **Disproved.** It fails the accepted requirement for a representative six-artifact manifest check. |

## Discoveries

0 accepted discoveries. [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/3ac2b17703f143b10241db045da19dac/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/3ac2b17703f143b10241db045da19dac/discoveries.md) record the consolidated result. Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and the correction delta. The original selected concerns were code correctness, producer/consumer seams, simplification, and regression protection; the prior report had disabled errors and docs. The correction adds no user-facing surface, so no gated-off lens required re-evaluation. The reviewer inspected the changed test, its workflow consumer and artifact producers, and the prior correction record; no required evidence was unavailable.

## Publication

Canonical PR comment: https://github.com/coleam00/Archon/pull/2991#issuecomment-5464825690

Wirasm:
## Ready to merge

The fix is a one-line change to `.github/workflows/release.yml:280` (`sha256sum archon-* archon-web.tar.gz` → `sha256sum archon-*`), removing the redundant explicit operand that caused `archon-web.tar.gz` to get two checksum rows (#2377). Traced the workflow end to end: exactly six artifacts ever reach `dist/` in the `release` job (five build-matrix binaries + the web tarball, merged in by `download-artifact` with `merge-multiple: true`, into a directory `actions/checkout` never populates since `dist/` is gitignored), all six start with `archon-`, and `archon-*` alone covers all of them exactly once. The fix is correct and complete for the current artifact set.

The regression test (`scripts/release-checksums.test.ts`) was rewritten by a correction round after the original version — which spawned a real `sha256sum` subprocess against temp files and parsed its output with a two-space regex — failed on Windows CI (job 99161669392: all six parsed filenames came back `undefined`, consistent with GNU coreutils defaulting to binary-mode `<hash> *<name>` output rather than text-mode `<hash>  <name>` on that platform). The rewrite spawns nothing: it statically parses `release.yml` for the matrix `binary:` entries and the tar command's output filename, and checks the checksum command's operands (as `Bun.Glob` patterns) against that derived set. I reproduced the original Windows failure from the job log, confirmed the new test passes on both CI legs on this head (`gh pr checks 2991`, including `test (windows-latest)`), and independently reproduced all three claimed red mutations (duplicate operand, lost coverage, broken derivation triggering the vacuous-pass guard) with byte-for-byte file restores in between. The derivation is honest: it's driven by the matrix and tar command directly, so adding a sixth platform to the build matrix is picked up without editing the test (verified by mutation).

**Validation:** `bun test ./scripts/` — 148 pass, 0 fail. `bun x tsc --noEmit -p scripts/tsconfig.json` — clean. `bun run lint` — clean (covers `scripts/**/*.ts`). GitHub Actions on the reviewed head — all green, including the `test (windows-latest)` leg that originally failed.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Test's CRLF-normalization comment overstates what the `.replace()` call actually prevents | OPEN |
| `R2` | Suggestion | The test's duplicate-key collapsing (`Object.fromEntries`) and first-match-only operand regex can't see a duplicated artifact name if one is ever reintroduced | OPEN |
| `R3` | Suggestion | The release job's `files:` block (untouched by this PR) repeats the same redundant-operand shape as the bug this PR fixes, currently inert | DECLINED |

<details>
<summary><code>R1</code> — misleading comment on the CRLF-normalization line</summary>

**Impact:** No functional impact — the comment describes a benefit the line doesn't actually provide for the current regex set, which could mislead a future editor into thinking removing the `.replace()` would reintroduce a Windows CRLF bug.

**Evidence:** `scripts/release-checksums.test.ts:6-8` reads:
```ts
// `* text=auto` checks this workflow out with CRLF on Windows CI. Normalizing here lets every
// pattern below anchor on LF, so a stray `\r` can never reach a capture.
const workflow = readFileSync(...).replace(/\r\n/g, '\n');
```
`.gitattributes` confirms `release.yml` is `text=auto` with no `eol=lf` override, so the CRLF-on-Windows-checkout premise is real. But independently verified (twice — once by me, once by the `code-simplifier` agent, via a standalone script feeding the actual `release.yml` content through all three extraction regexes with and without `\r\n`→`\n` normalization) that the three regexes (`/^ +binary: (\S+)$/gm`, `/^ *-czf dist\/(\S+) /m`, `/^ *sha256sum (.+) > checksums\.txt$/m`) produce byte-identical output either way. This is because ECMAScript regex treats a bare `\r` as its own line-terminator character satisfying `^`/`$` in multiline mode (confirmed: `"abc\rxyz".match(/^abc$/m)` matches), and `\S`/`.` (no dotAll) already exclude `\r` from captures. So the normalization line is not what's currently preventing a stray `\r` from reaching a capture — the regex anchoring already does that on its own.

**Required outcome:** Reword the comment to describe this as forward-looking protection against a future regex edit (e.g. swapping `\S+` for a character class that doesn't exclude `\r`), rather than describing a mechanism the current regexes need — or drop the line, since it's currently redundant. Not required for merge; the line is harmless.

**Disposition:** OPEN — cheap wording fix, no urgency.

</details>

<details>
<summary><code>R2</code> — duplicate artifact names could silently collapse before the comparison</summary>

**Impact:** No live bug today (there is exactly one `sha256sum ... > checksums.txt` line and no duplicate `binary:` matrix values), but the test's own assertion has a blind spot matching the defect class it exists to catch.

**Evidence:** `scripts/release-checksums.test.ts:16` uses non-global `.match()` for the checksum-operand line, so a second `sha256sum ... > checksums.txt` line added later in the file would be silently ignored (only the first textual match is read, while at runtime the *last executed* write to `checksums.txt` wins — the test's model and the runtime model diverge). Separately, `scripts/release-checksums.test.ts:28-33` builds `rowsPerArtifact` via `Object.fromEntries`, which collapses duplicate keys — a matrix entry that copy-pastes another platform's `binary:` value without renaming it would produce a `releaseArtifacts` array with a duplicate name that disappears identically on both the actual and expected side of the comparison, so the test would never notice only 4 distinct binaries shipped where 5 were intended. Verified against synthetic workflow snippets reproducing the test's exact logic.

**Required outcome:** Not required for this PR. If revisited: use a global match (or an explicit "exactly one checksum command line" assertion) for the operand line, and tally artifacts with something that preserves duplicate keys (e.g. a `Map` built from entries with an explicit duplicate check) instead of `Object.fromEntries`.

**Disposition:** OPEN — no concrete trigger exists today; noted for whoever next touches this test.

</details>

<details>
<summary><code>R3</code> — release <code>files:</code> block echoes the same redundant-operand shape, untouched by this PR</summary>

**Impact:** None today. `.github/workflows/release.yml:300-303` (the `softprops/action-gh-release` `files:` block, not touched by this diff) lists `dist/archon-*` and `dist/archon-web.tar.gz` — the same redundant-member shape #2377 was about. Independently traced `action-gh-release`'s actual source (`paths()` in `src/util.ts`, `upload()` in `src/github.ts`): today the duplicate resolves to the same path twice, and the upload path's exists-then-replace logic absorbs that harmlessly — confirmed inert, not a live bug. The only way this becomes consequential is if the web tarball is ever renamed to something not matching `archon-*` in some future, unrelated edit; `fail_on_unmatched_files` isn't set in this step (defaults to `false` in the action), so a future rename could silently drop the release asset with no CI signal. That is a hypothetical two-step failure mode, not a current one.

**Required outcome:** None required by this PR's stated scope boundary ("release assets, upload step, and checksum algorithm are unchanged" — PR body). Any future edit that renames the web tarball would need to touch this exact block anyway.

**Disposition:** DECLINED — pre-existing, untouched by this diff, no concrete failure mode today, and fixing it would expand this PR beyond its own stated boundary. Worth a maintainer's attention if `release.yml`'s artifact set changes again, not before.

</details>

<details>
<summary>Checked and not a finding</summary>

**Claim examined:** `scripts/checksums.sh`'s `EXPECTED_BINARIES` array (4 entries: darwin-arm64, darwin-x64, linux-arm64, linux-x64) omits Windows, which the build matrix in `release.yml` includes (5 entries) — raised by `prp-core:seam-analyzer` as possible drift.

**Why it's not a finding:** `scripts/checksums.sh` is the manual/local-dev companion to `scripts/build-binaries.sh`, not a mirror of the CI matrix. `scripts/build-binaries.sh`'s own header comment states its local multi-target mode "builds all 4 local targets," and its `LOCAL_TARGETS`-equivalent list (`bun-darwin-arm64`, `bun-darwin-x64`, `bun-linux-x64`, `bun-linux-arm64`) matches `checksums.sh`'s `EXPECTED_BINARIES` exactly — Windows cross-compilation isn't part of local multi-target builds at all; CI builds it per-target separately. `packages/docs-web/src/content/docs/contributing/releasing.md` confirms local builds only ever cover the current-platform/local set and that cross-platform binaries require GitHub Actions. There is no drift: `checksums.sh` correctly reflects what `build-binaries.sh` can produce locally.

</details>

<details>
<summary>Reviewer coverage</summary>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | No additional findings — independently confirmed the fix's correctness and the test's non-vacuousness (ran the test, reverted the workflow line, confirmed red, restored) |
| `seams` | R2, R3 (plus the checksums.sh claim, investigated and disproven — see above) |
| `simplify` | R1 |

#### Validation

|---|---|---|
| `bun test ./scripts/release-checksums.test.ts` | PASS | 1 pass, 1 expect() call |
| `bun test ./scripts/` (full directory) | PASS | 148 pass, 0 fail, 308 expect() calls — matches PR body's claimed count |
| `bun x tsc --noEmit -p scripts/tsconfig.json` | PASS | exit 0, no diagnostics |
| `bun run lint` | PASS | exit 0; covers `scripts/**/*.ts` (includes the new test file) |
| Mutation: reintroduce duplicate `archon-web.tar.gz` operand (byte-restored after) | RED as expected | `archon-web.tar.gz: 2` vs expected `1` |
| Mutation: narrow checksum glob, dropping Windows binary coverage (byte-restored after) | RED as expected | `archon-windows-x64.exe: 0` vs expected `1` |
| Mutation: break the tar-line regex so web-dist derivation fails (byte-restored after) | RED as expected | throws the explicit vacuous-pass guard error |
| `gh pr checks 2991` on reviewed head | PASS | Test Suite, `test (ubuntu-latest)`, `test (windows-latest)`, docker-build, postgres-parity, schema-upgrade all green |
| Windows CI job log, reviewed head (job 99164159682) | PASS | `(pass) release checksums include each release artifact once` |
| Windows CI job log, original failing head (job 99161669392) | Confirms original failure | all six `manifestFiles` entries `undefined` against the two-space regex — consistent with msys/binary-mode `sha256sum` output |
| `bun run validate` (full) | NOT RUN | Not run in full given the tight diff scope and GitHub Actions already exercising the full matrix (ubuntu + windows) on this exact head; the above targeted checks cover every changed file plus the equivalent of `test:install`/`type-check`/`lint`/`test` for the changed surface |

</details>

PR #2995, opened 2026-08-29
Fail workflow conditions on structured fields

### LINKED ISSUE #1910: bug(workflows/condition-evaluator): present non-primitive field in when: stringifies and silently skips instead of failing loud (schemaless footgun)
Follow-up from #1770 / the #1849 structured-output work (PRs #1883 + #1889). #1770's reported cascade is fixed, but validating it surfaced a narrower residual that the structured-output validation does **not** cover.

## Symptom

When a `$node.output.field` reference in a `when:` condition resolves to a **present** non-primitive value (object or array), the condition evaluator `JSON.stringify`s it and string-compares — which is almost always false — so the gated node **silently skips**. No throw, no warning. This is the same class of silent-skip #1770 raised, surviving for the present-non-primitive case.

This is currently **deliberate and tested** behavior (`condition-evaluator.ts:73`; `condition-evaluator.test.ts:83-95`), so this issue is a decision to change it, not a regression.

## Where it bites (the gap #1883 can't close)

The producer-side ajv validation from #1883 catches a non-primitive in a *declared* `output_format` field (e.g. a `type: string` field returning an object fails validation at the producer). But it cannot help **schemaless** producers — a `bash`/`script` node has no `output_format` to validate against. So:

```yaml
- id: prep
  bash: |
    # intended: {"ready":"true"} — but a bug emits a nested object
    echo '{"ready": {"value": "true"}}'
- id: deploy
  depends_on: [prep]
  when: "$prep.output.ready == 'true'"   # '{"value":"true"}' == 'true' → false → deploy SILENTLY skips
```

`resolveNodeOutputField` returns the object value (`output-ref.ts:156`), `resolveOutputRef` stringifies it (`condition-evaluator.ts:73`), `evaluateAtom` compares it false (`:175`), and the workflow proceeds as if the gate legitimately evaluated false. The author's expectation (`deploy` runs when ready) is silently dropped — exactly the no-silent-drop failure mode (CLAUDE.md §Fail Fast).

The condition evaluator is the **only** place this can be caught for schemaless producers, since there's no schema upstream.

## Proposed direction

In `evaluateAtom` / `resolveOutputRef`, when a `.field` reference resolves to a **present** non-primitive (object/array) and is used in a scalar comparison (`==`/`!=`/`<`/`>`/`<=`/`>=`), fail loud instead of stringify-comparing:

- emit a structured event (e.g. `dag.condition_field_not_primitive` with `{ nodeId, field, actualType, exprSnippet }`), and
- surface it as a node failure (the no-silent-drop path the rest of #1883 established), rather than a silent skip.

This matches #1770's original proposed fix, which #1883 implemented for the *absent / not-in-schema / unparseable* cases but not for the *present-object-valued* case.

## Scope boundary

- **Only the present-non-primitive-in-a-comparison case** — primitives (string/number/boolean) are unchanged; declared-optional-absent → `''` stays as approved in #1849; whole-text `$node.output` is untouched.
- **Decision required:** today's stringify-compare is intentional and tested. If we change it, update `condition-evaluator.test.ts:83-95` to assert fail-loud instead of stringify, and confirm no bundled workflow legitimately compares an object/array field to a scalar literal.
- Not re-opening anything in #1849 / #1897 — #1897 is the AI-`output_format` loose-schema lever; this is the condition-evaluator + schemaless lever. Complementary.

## Acceptance

- A `when:` atom whose `.field` resolves to a present object/array fails loud (event + node failure), not a silent skip.
- Primitive comparisons and declared-optional-absent → `''` behavior are unchanged; existing valid `when:` clauses evaluate identically.
- Tests: schemaless node emitting an object-valued field gated by `when:` fails the workflow loudly; `condition-evaluator.test.ts:83-95` updated to the new contract.

## Evidence (post-#1883 code)

- `packages/workflows/src/output-ref.ts:151-156` — schemaless path returns a present object value (no throw for object-valued present keys).
- `packages/workflows/src/condition-evaluator.ts:70-73` — object/array value → `JSON.stringify` (the silent-compare point).
- `packages/workflows/src/condition-evaluator.ts:166-175` — `evaluateAtom` compares the stringified value; `OutputRefError` is rethrown but a present non-primitive never throws.
- `packages/workflows/src/condition-evaluator.test.ts:83-95` — the test that pins the current stringify behavior.

Companion to #1897 (loose `output_format` schemas). Together they close the silent-skip class left after #1769 / #1770 / #1883.

Wirasm:
## Problem and outcome

Fielded `when:` comparisons converted present object and array outputs to JSON text. A scalar comparison could therefore look like an ordinary false condition and skip the gated node instead of surfacing a workflow-authoring error.

- **Outcome:** A present object or array in `$node.output.field` now fails the gated node with an actionable error and a safe diagnostic log.
- **Invariant:** Strings, numbers, booleans, null, and declared-optional absent fields retain their existing condition behavior; general structured-value transport is unchanged.
- **Scope boundary:** Whole-output references, input references, condition grammar, and structured values outside fielded `when:` comparisons are unchanged.
- **Root cause:** The condition evaluator applied the general canonical value-to-text conversion to fielded condition values before checking whether the value was a scalar.

## Review guidance

- **Feedback requested:** Correctness of the narrow `when:` boundary and preservation of scalar behavior.
- **Start here:** `packages/workflows/src/condition-evaluator.ts:115` — rejects only present object and array field values before canonical text conversion.
- **Review order:** Evaluator guard and unit tests; executor and dry-run failure-path tests; authoring guidance.
- **Lower-attention areas:** Documentation is a one-line statement of the enforced boundary; the docs build passed.
- **Known risk or uncertainty:** Existing workflows that intentionally compare object or array fields as canonical JSON now fail the gated node. This is the accepted issue #1910 behavior.

## Solution

The evaluator keeps `resolveNodeOutputField` and the general canonical value-to-text rule intact. At the `when:` field-reference boundary it now rejects arrays and non-null objects, logs `dag.condition_field_not_primitive` with node, field, type, and expression metadata, and throws guidance to route on a scalar field or inspect the structure in a script node. Existing executor and dry-run error propagation records the consuming node as failed.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A present object or array field was converted to JSON text for a scalar `when:` comparison. | The gated node fails with an actionable error. |
| **Failure behavior** | A non-matching scalar comparison could silently skip the node. | The evaluator logs safe diagnostic metadata and the normal `node_failed` path records the failure. |

## Validation

- `bun test src/condition-evaluator.test.ts` — passed (96 tests) — verifies rejection and safe logging while retaining scalar behavior.
- `bun test src/dag-executor.test.ts` — passed (701 tests) — verifies the consuming node records `node_failed`, not a skip.
- `bun test src/dry-run.test.ts` — passed (75 tests) — verifies dry runs record the gated node as failed.
- `bun run type-check`, `bun run lint`, `bun --cwd packages/docs-web build`, `bun run format:check`, and `bun run validate` — passed — verify types, lint, docs, formatting, and the repository gate.
- **Not verified:** Nothing material; the evaluator, executor, dry-run, and documentation paths changed here are covered by the listed checks.

## Links

- Closes #1910

## Summary by CodeRabbit

- **Bug Fixes**
  - `when:` conditions now reject referenced objects and arrays instead of converting them to text for comparison.
  - Invalid condition references now produce clear errors and diagnostics.
  - Entry-node condition failures correctly fail the workflow rather than silently skipping the node.

- **Documentation**
  - Clarified that portable model tiers resolve from configured tier entries, with built-in defaults available for Claude and Codex.
  - Clarified that scalar conditions must resolve to a string, number, boolean, or null.


Wirasm:
# Review report

## Verdict

**Ready — none.** The correction closes all three prior Important findings without expanding the accepted condition boundary.

## Accepted contract

Fielded `when:` references must reject present object and array values, log `dag.condition_field_not_primitive` with safe reference metadata, and fail the consuming node in normal execution and dry run. Scalars, null, and declared-optional absent fields retain their current semantics. The change must not alter general value-to-text transport, structured bindings, prompts, scripts, `with:`, input transport, whole-output references, condition grammar, or schemas.

## Reviewed head SHA

`d88ff022405a61aa1881d9ca4b6f163200731e30`

## Findings

### Critical

None.

### Important

None open.

### Rejected findings

None this round.

### Suggestions

None.

## Prior findings

| ID | Sources | Verdict | Evidence |
| --- | --- | --- | --- |
| R1 | `[code, seams]` | **Fixed.** Cached included entries now rethrow a composed-entry evaluator error instead of converting it to `skip`; the surrounding pre-execution handler persists `node_failed`. The new resumed-entry regression proves the durable result. | [dag-executor.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/archon-ship-1788035952283/packages/workflows/src/dag-executor.ts:1999), [dag-executor.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/archon-ship-1788035952283/packages/workflows/src/dag-executor.ts:10987), [dag-executor.test.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/archon-ship-1788035952283/packages/workflows/src/dag-executor.test.ts:26910). |
| R2 | `[tests]` | **Fixed.** The evaluator regression supplies a structured previous-iteration output and asserts both the object error and the safe diagnostic metadata. | [condition-evaluator.test.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/archon-ship-1788035952283/packages/workflows/src/condition-evaluator.test.ts:110). |
| R3 | `[docs]` | **Fixed.** The guide now limits the promise to scalar field access, matching the nearby object/array failure rule. | [authoring-workflows.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/archon-ship-1788035952283/packages/docs-web/src/content/docs/guides/authoring-workflows.md:523). |

The verified invariant behind R1 is that a present structured field in a gated `when:` reference fails its consuming node instead of becoming an inactive gate. The correction preserves the existing descendant suppression: only an included entry being evaluated for resume propagation rethrows; other composed descendants still return `skip`. I traced the cached-entry call through the pre-execution failure handler and verified the added regression's event assertions.

## Discoveries

0 accepted discoveries. See [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/2497d2b04367dd882b27e23696f374ec/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/2497d2b04367dd882b27e23696f374ec/discoveries.md). Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and correction delta. The original selected concerns were cached included-entry failure ownership, `$LOOP_PREV` structured-field coverage, and authoring-guide accuracy. I inspected the changed control flow, relevant failure handler, tests, and documentation; no required evidence was unavailable. The correction delta added no new user-facing surface, so no gated-off lens was re-evaluated.

Validation run from `packages/workflows`: `bun test src/condition-evaluator.test.ts` (97 pass), `bun test src/dag-executor.test.ts` (702 pass), and `bun test src/dry-run.test.ts` (75 pass). `git diff --check 6715a452155b4730fd0bb2e1cd8cc01f47b36625..HEAD` also passed.

Wirasm:
## Ready to merge

The fix is scoped exactly to the case issue #1910 accepted: a `when:` field reference that
resolves to a present object or array now fails the gated node loudly (structured
`dag.condition_field_not_primitive` log event plus an actionable `node_failed`), instead of
silently `JSON.stringify`-comparing and skipping. A second commit closes a real gap found in
an earlier round — the composed-include boundary re-check on a resumed cached entry was
unconditionally catching and swallowing that new throw back into a silent `skip`; it now
rethrows only for the resumed entry's own boundary, leaving every other descendant boundary's
existing skip-on-error behavior untouched. I independently reproduced the issue's exact repro
shape red on pre-PR `dev` and green on this head, and independently audited every `when:`
field reference across all 32 bundled workflows — not just the 20 the issue's own audit
covered — and confirmed none newly fails. Two non-blocking documentation/architecture
suggestions remain open, and one adjacent (but explicitly out-of-scope) defect in the sibling
`$INPUTS.<name>` path is now tracked as a separate issue.

**Validation:** all targeted checks pass — 97/97, 702/702, 75/75 tests across the three
changed test files; type-check and lint clean repo-wide; formatting clean on every changed
file; seen-red/seen-green reproduction of the issue's exact repro shape; independent
32-workflow bundled corpus audit found zero newly-failing `when:` clauses.

**Tracked follow-ups:** [#2999](https://github.com/coleam00/Archon/issues/2999)

### Findings

|---|---|---|---|
| `R1` | Suggestion | Module-level doc comment in `condition-evaluator.ts` still enumerates only two throw classes, omitting the new plain-`Error` object/array throw this PR added | OPEN |
| `R2` | Suggestion | The "a `when:` reference-throw must propagate, never become a silent skip" invariant is enforced by two independently hand-maintained control-flow shapes, not a shared or type-checked mechanism | OPEN |
| `R3` | Suggestion | `resolveInputRef` (`$INPUTS.<name>`) is the unguarded sibling of `resolveOutputRef` and reproduces the identical silent-stringify-compare bug for object/array-valued inputs, in both false-skip and false-positive directions | TRACKED FOLLOW-UP |

<details>
<summary><code>R1</code> — stale module-doc throw-taxonomy in condition-evaluator.ts</summary>

**Impact:** A maintainer who reads only the module doc (the conventional first stop, `condition-evaluator.ts:22-33`) sees exactly two documented throw classes (`OutputRefError`, `InputRefError`) and has no way to learn a third one — a plain `Error` for a present object/array field — now exists. No runtime effect; this is a documentation-accuracy gap the PR's own diff created by updating one of two comment blocks describing the same contract.

**Evidence:** `packages/workflows/src/condition-evaluator.ts:22-33` (stale module doc, untouched by this diff) versus `packages/workflows/src/condition-evaluator.ts:67-71` (the function docstring 40 lines below it, correctly updated to mention the new throw). The updated docstring also quietly dropped the explicit `OutputRefError` class name from its own sentence without naming what class the new case throws instead.

**Required outcome:** Add the third throw reason (present object/array → plain `Error`, `dag.condition_field_not_primitive`) to the module-doc bullet list, matching the function docstring below it. A same-file, same-commit, one-paragraph edit.

**Disposition:** OPEN. No runtime consequence and does not block merge; a trivial follow-up edit whenever this file is next touched, or a quick fixup now at the operator's discretion.

</details>

<details>
<summary><code>R2</code> — "throw must propagate" invariant duplicated across two call sites, not type-enforced</summary>

**Impact:** The invariant "a `when:`-reference throw must always reach the caller as a failure, never get silently caught into a skip" is currently expressed two different ways: by the absence of a `try/catch` around `resolveAtomRef` in `condition-evaluator.ts:227`, and by a hand-written conditional rethrow inside a `catch` in `checkComposedBlockBoundaries` (`dag-executor.ts:1999-2000`). The far side's prior version of this invariant was already wrong once — that is exactly what this PR's second commit corrects — which is direct evidence the near side's comment-only enforcement isn't self-protecting either. Not live today: both sites are now pinned by regression tests this PR added, so a straightforward revert of either would be caught by CI. Residual risk is a future third call site of `evaluateCondition` re-deriving its own catch logic without discovering this invariant.

**Evidence:** `packages/workflows/src/condition-evaluator.ts:222-227`, `packages/workflows/src/dag-executor.ts:1995-2004`. Confirmed the `ComposedBlockBoundary` discriminated union (`compiled-command.ts:67-79`) already rules out the `isEntry`/`entryTriggerRule` state-representation concern this could otherwise raise — it doesn't, so this is purely a control-flow duplication finding, not a missing-type finding.

**Required outcome:** No correction required now — current tests make the duplication low-risk. If either call site is touched again, consider a shared "classify-and-rethrow" helper so the propagate-vs-skip decision has one owner instead of being re-derived by hand at each site.

**Disposition:** OPEN as a suggestion; does not block. No concrete smaller shape exists today that isn't itself speculative machinery for a two-call-site duplication that is currently test-pinned.

</details>

<details>
<summary><code>R3</code> — <code>resolveInputRef</code> reproduces the same silent-stringify-compare bug for <code>$INPUTS.&lt;name&gt;</code></summary>

**Impact:** `resolveAtomRef` (`condition-evaluator.ts:185`) dispatches to exactly two value-producing functions depending on `WhenAtomRef.kind` (`when-atom.ts:92-104`: `node`, `loop_prev`, `input`). `node` and `loop_prev` both route through the now-guarded `resolveOutputRef`. `input` routes to `resolveInputRef` (`condition-evaluator.ts:177-182`), which still runs any resolved value — including a present object or array — straight through `canonicalValueText` with no guard. Inputs are typed `JsonValue` and can legitimately be objects or arrays by design (`workflowInputSpecSchema`, `schemas/workflow.ts:154-160`; #2637's "logical value" contract), so this is reachable, not theoretical.

I independently reproduced both directions live in the review worktree (scratch test written, run, and removed; tree left clean):
- `evaluateCondition("$INPUTS.route == 'true'", new Map(), { route: { ready: true } })` → `{ parsed: true, result: false }` — silent skip, no throw.
- `evaluateCondition('$INPUTS.tags == \'["a","b"]\'', new Map(), { tags: ['a', 'b'] })` → `{ parsed: true, result: true }` — a false-positive branch on a `JSON.stringify` coincidence, not just a false skip.

This is the identical failure class issue #1910 closed for `.field` references, one function away, in the same file.

**Class:** The invariant is "every value-producing branch `resolveAtomRef` can reach must reject a present non-primitive before a scalar comparison." Enumeration: `WhenAtomRef.kind` has exactly three values (`when-atom.ts:92-104`). `node` and `loop_prev` → `resolveOutputRef`, guarded by this PR, examined and clean. `input` → `resolveInputRef`, unguarded, the one affected member.

**Required outcome:** Mirror #1910's guard in `resolveInputRef`: reject a present array/non-null-object input value in a scalar comparison, with a structured diagnostic and an actionable throw, matching the fielded case's shape.

**Found by:** `prp-core:code-simplifier` (independently reproduced and verified by the reviewing coordinator)

**Disposition:** TRACKED FOLLOW-UP — [coleam00/Archon#2999](https://github.com/coleam00/Archon/issues/2999). Not a defect in PR #2995: issue #1910's accepted text and the operator's decision comment on it both explicitly restrict the contract to `.field` references and state input references are unchanged; PR #2995's own "Scope boundary" section states the same. The original work order itself — not merely the PR's self-report — draws this line, so it does not reopen or block the PR under review, but it is a real, reproducible, adjacent gap worth its own accepted work order.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | No additional findings — independently traced both commits end-to-end, including a mechanical revert-and-rerun of the resume-path fix to confirm the new test actually exercises the regression it targets. |
| `seams` | R1, R2 |
| `simplify` | R3 |

#### Validation

|---|---|---|
| `bun test src/condition-evaluator.test.ts` (packages/workflows) | PASS | 97 pass, 0 fail, 155 expect() calls |
| `bun test src/dag-executor.test.ts` (packages/workflows) | PASS | 702 pass, 0 fail, 2073 expect() calls |
| `bun test src/dry-run.test.ts` (packages/workflows) | PASS | 75 pass, 0 fail, 209 expect() calls |
| `bun run type-check` (packages/workflows) | PASS | `tsc --noEmit` exit 0, no output |
| `bun run lint` (repo root) | PASS | clean across all 11 linted packages + scripts |
| `bun x prettier --check` on all 6 changed files | PASS | "All matched files use Prettier code style!" |
| Seen-red: issue #1910's exact repro (`bash` node emitting `{"ready":{"value":"true"}}`, `when: "$prep.output.ready == 'true'"`) on pre-PR `dev` (merge-base `f339df7a6`) | PASS (confirms bug was real) | `evaluateCondition(...)` returned `{parsed:true, result:false}` — silent skip, no throw |
| Seen-green: identical repro on PR head `d88ff0224` | PASS (confirms fix) | `evaluateCondition(...)` threw `"Condition reference '$prep.output.ready' resolved to an object"` |
| Independent bundled-workflow corpus audit: every `when:` clause with a `.field` ref across all 32 shipped bundled workflows (20 in `.archon/workflows/defaults/legacy/` + `archon-assist` + all 11 in the `sdlc` pack, per `scripts/generate-bundled-defaults.ts`'s actual embedding logic — 12 more than the issue's own comment audited, which covered only the legacy 20) | PASS | Every referenced field traces to a declared `output_format` property of `type: string`/`type: boolean`, or a schemaless script whose emitted JSON key is provably a string/boolean literal by reading the producing script. Zero would trip the new guard. |
| `bun --cwd packages/docs-web build` | NOT RUN | Relied on the PR's own claimed pass; the docs change is a two-sentence prose addition with no syntax risk, inspected directly instead |
| `bun run validate` (full monorepo suite) | NOT RUN | The targeted checks above cover every changed file and its direct consumers; the diff is confined to `packages/workflows` (+ one docs paragraph), so the full cross-package suite was not proportionate to re-run in full |

</details>

PR #3000, opened 2026-08-29
fix(sdlc): a delivery that dies after opening its PR is never reported as delivered

### LINKED ISSUE #2997: fix(sdlc): ship outcome reports a delivered URL when delivery failed after PR creation
## Problem

The ship/stabilize/upkeep `outcome` node can report a delivered pull-request URL for a run whose delivery failed. Live occurrence: run `3ac2b177` (dogfood, 2026-08-29) failed honestly at `deliver__flip-ready` (non-green checks refused the flip, PR #2991 left draft), yet the ship-level `outcome` node ran anyway and printed `https://github.com/coleam00/Archon/pull/2991` as its terminal report before the run was marked failed.

Event trace (same second, run `3ac2b177`):

```
node_failed     deliver__flip-ready   (refusing to flip with non-green checks)
node_skipped    deliver__outcome
node_started    outcome
node_completed  outcome               (node_output: .../pull/2991)
workflow_failed
```

## Cause

Two facts compose:

1. `archon-ship`'s `outcome` declares `depends_on: [..., deliver]` with `trigger_rule: none_failed_min_one_success`, intending "outcome does not run when delivery failed". But composition maps a dependency on an `include:` node to its **returns node** — `deliver` resolves to `deliver__pr`, which completed long before the flip failed. The trigger rule evaluated a completed producer and fired; the failed `flip-ready` was invisible to it.
2. The outcome scripts' record-read rests on the inference "a non-null `delivered` under this trigger rule proves deliver completed, therefore flip-ready's read-back confirmed non-draft." That inference was load-bearing in the change that removed the terminal `gh pr view` re-read — and fact 1 falsifies it: `$deliver.output` is the pr node's creation-time record, which exists whenever PR creation succeeded, regardless of what failed after it.

The run-level status is still truthful (`failed`), so this is a lying terminal *report* inside a failed run, not a false-success run — but the report is exactly what an orchestrating agent reads.

## Desired outcome

The outcome node of every routed tail runs its delivered branch only when the composed delivery actually succeeded end to end, and a delivery that died after PR creation produces the delivery-died report, never a URL presented as delivered.

## Acceptance invariants

- On a run where `deliver__flip-ready` (or any deliver node after `pr`) fails, the tail's terminal report is the delivery-died shape; no delivered URL is presented.
- A fixture pins it: deliver fails post-PR-creation → outcome reports delivery-died (the existing `delivery-died` fixture covers delivered-null; this case is delivered-non-null with a later failure).
- Whether the fix is in dependency semantics (an include dependency meaning whole-include success), a bound success signal, or the outcome scripts consuming an explicit flip confirmation is an implementation choice — but the inference "record exists ⟹ flip succeeded" must stop being load-bearing anywhere.

## Related

The engine-owned terminal-result selection design (#2453) names these collectors as its consumer evidence; this is another datapoint for it.

Wirasm:
## Problem and outcome

A delivery that died after opening its pull request could still be reported as delivered. The composed dependency on the deliver block resolves to its returns node, and deliver returned its `pr` node — a creation-time record that exists whether or not anything after it succeeded — so both the trigger rule and the `delivered` binding were satisfied by a delivery that never finished. Dogfood run `3ac2b177` hit it live: the flip refused non-green checks, the run failed, and the terminal report printed the PR URL as if delivered.

- **Outcome:** a failed flip blocks the composing tail's terminal node; a death anywhere after PR creation produces the died-mid-flight report with no URL; a success hands the tails the read-back-verified URL.
- **Invariant:** the inference "a delivered value exists, therefore the flip succeeded" is structural, not assumed — the delivered value *is* the flip's output, so it cannot exist without the flip.
- **Scope boundary:** deliver's internal nodes, the review contract, and the advisory routes are untouched; the three tails' shared helper stays byte-identical.
- **Root cause:** `returns: pr` made a composed `depends_on: [deliver]` resolve to a node that completes at PR creation, which falsified the reasoning the record-read outcome scripts rested on.

## Review guidance

- **Feedback requested:** whether the returns-node semantics ("what a composed dependency resolves to") are stated clearly enough in the new comment that a future returns change gets the same scrutiny.
- **Start here:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml` — the `returns:` swap and its comment are the whole mechanism.
- **Review order:** the yaml comment; the shared `delivered_url` helper in the three outcome scripts; `delivered-unusable.stubs.yaml` (now the guard: a complete record beside a garbage flip value must be refused); the new `delivery-died-after-pr.stubs.yaml`.
- **Lower-attention areas:** the regenerated bundled defaults; the comment-only truth updates in five delivered fixtures.
- **Known risk or uncertainty:** a flip that *fails* (rather than being skipped) now blocks the tails' outcome entirely, so the run's failure surface is flip-ready's own error — which since the recent recovery-prose change names the re-run-and-resume path. That is deliberate: the flip error is a better report than a generic delivery-died line for that case.

## Solution

`archon-deliver` returns `flip-ready` instead of `pr`. The flip's stdout contract is already "the delivered URL and nothing else," verified by read-back before it prints. The three tails' shared `delivered_pr_url` record parse becomes `delivered_url`, a URL validation over the bound value — kept byte-identical across ship, stabilize, and upkeep. The old helper's docstring stated the falsified inference verbatim ("a record only exists here because deliver completed, which means the ready flip ran"); the new one documents why the returns choice is load-bearing.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Terminal report reads the creation-time record's URL. | Terminal report reads the flip's verified URL. |
| **Failure behavior** | Delivery dying after PR creation still produced a "delivered" URL report inside a failed run. | Post-creation death → died-mid-flight report, no URL; failed flip → the flip's own error (with its recovery guidance) surfaces, outcome blocked. |

## Validation

- `bun run cli workflow test sdlc --json` — 45 passed, 0 failed (44 prior + the new post-creation-death fixture).
- Mutation seen red: reverting `returns:` to `pr` turns five delivered fixtures red (the tails refuse a JSON record as "not a single pull request URL"); restored, 45/45.
- `delivered-unusable` now guards the other direction: a complete record beside a garbage flip value is refused by the real script in a detached worktree, so a future change quietly re-reading the record goes red.
- `bun run generate:bundled` + `check:bundled` in sync; `bun run validate`, lint, type-check, format:check all green.

Closes #2997.

## Summary by CodeRabbit

* **Bug Fixes**
  * Delivery workflows now return the verified pull request URL directly.
  * Outcome reports validate and display the delivered URL correctly.
  * Improved handling of delivery failures and unusable delivered values.

* **Tests**
  * Updated delivery scenarios for post-creation failures and invalid results.
  * Confirmed reports are generated from the verified delivered URL.


Wirasm:
## Ready to merge (round 2 — corrections verified)

The fix is correct at the mechanism the whole PR rests on. I traced the engine's `returns:`/`depends_on` split myself (`packages/workflows/src/include-expander.ts`) rather than trusting the PR body: a composed `depends_on: [deliver]` join resolves to every actual DAG sink of the sub-graph (here, `archon-deliver`'s own internal `outcome` node), while `$deliver.output` resolves separately, through `primarySink`, to whatever node `returns:` names. Changing `returns: pr` to `returns: flip-ready` is what makes a genuinely failed flip (not merely skipped) hit the engine's pre-existing `assertProducerNotFailed` guard (`packages/workflows/src/dag-executor.ts:490-505`) instead of silently falling back to a stale, already-completed PR-creation record. I reproduced this live: I wrote a throwaway fixture that forces the real `flip-ready` bash node to execute and genuinely fail (no `origin` remote in the scratch worktree), and the fixture runner reported exactly two failed nodes — `deliver__flip-ready` and the composing tail's own `outcome` node — confirming the "failed flip blocks the caller's terminal node" claim is real engine behavior, not just a comment. I also reverted `returns:` to `pr` by hand and reran the full suite: exactly the five "delivered" fixtures the PR names went red, then restored byte-identical and reran clean (45/45). The shared `delivered_url` helper is proven byte-identical across ship/stabilize/upkeep, its `"\n"` check is a genuine escape (not the literal-newline corruption the PR body flags as almost happening), and `flip-ready`'s bash body genuinely never lets anything but the read-back URL reach real stdout on success. Bundled defaults regenerate to a byte-identical match.

**Round 2 (correction commit `ddd6816ee`):** all three Suggestions from round 1 are resolved by a comments-only correction (9 files, +33/-26, no behavior change). I re-verified each independently against the new head rather than trusting the summary of what changed: reran the full fixture suite (45/45), reconfirmed bundled-defaults regeneration is in sync, reran the exact grep search that enumerated R2's class (now zero matches), and read every changed line in the correction diff directly. See each finding's disposition below for what I checked and what I found.

Three non-blocking findings came out of round 1, one from independent verification I did beyond the PR's own claims: the new `delivery-died-after-pr.stubs.yaml` fixture, despite its header comment, provided no actual regression protection for the value-level defect it named — I proved this by mutation, not by inspection. The other two were a stale-comment class the original PR's edit left behind, and a real but non-blocking gap in what the operator sees when a flip genuinely fails (as opposed to being skipped). All three are now fixed or explicitly, deliberately decided.

**Validation:** Bundled-defaults regeneration verified in sync at the correction head (`bun run scripts/generate-bundled-defaults.ts --check`). Full SDLC fixture suite: 45/45 passing at the correction head. No issue/run reference introduced into any pack comment (grepped the correction diff). CI on the correction commit: schema-upgrade, postgres-parity, and `changes` green at time of writing; Test Suite/docker-build/ubuntu/windows/CodeRabbit still running when last checked — see Validation table for final status.

### Findings

|---|---|---|---|
| `R1` | Suggestion | New `delivery-died-after-pr` fixture stays green under the exact mutation it claims to guard against | FIXED |
| `R2` | Suggestion | Stale "JSON record" language survives in 6 places the PR's own semantic change falsifies | FIXED |
| `R3` | Suggestion | A genuinely failed flip drops the caveats/discoveries reporting channel that both dev and the skipped-flip path preserve | FIXED |

Every finding was a Suggestion: the shipped behavior itself was correct and independently verified in round 1 (proved by direct engine reproduction, not by trusting any fixture) — none of these ever changed what the delivered outcome does. All three are now closed.

<details>
<summary><code>R1</code> — New fixture provides no regression protection for the defect it names</summary>

**Impact:** `.archon/workflows/sdlc/ship/fixtures/delivery-died-after-pr.stubs.yaml` is the PR's headline new fixture, and its header comment states it covers "the composed shape where a stale creation-time record could otherwise masquerade as a delivery." It does not. The fixture stubs the ship-level `outcome` node's entire output to a canned string (`outcome: "outcome: delivery started but did not complete"`), so the actual value bound to `delivered` — the thing whose correctness is the entire point of this PR — is never asserted. I verified this is not a hypothetical concern: I ran the fixture suite once at the fixed head (`returns: flip-ready`), then hand-reverted `archon-deliver.yaml`'s `returns:` back to `pr` (the exact pre-fix bug) and reran — `delivery-died-after-pr.stubs.yaml` reports `pass: true` identically in both cases. It would not catch a regression that reintroduced the original bug in this specific composed shape.

**Evidence:** `.archon/workflows/sdlc/ship/fixtures/delivery-died-after-pr.stubs.yaml:34` (`outcome: "outcome: delivery started but did not complete"`); mechanism confirmed in `packages/workflows/src/dry-run.ts:1103-1120` (node-local `with:` bindings resolve before the stub substitution, so the fixture *could* have caught a binding-resolution failure — it just doesn't hit one here, because in this fixture's chosen scenario `deliver__gate-validated` fails *before* `flip-ready` ever runs, so `flip-ready` lands `skipped` under both old and new `returns:` values, and the `delivered` binding resolves successfully — just to a different value — either way). Direct reproduction: fixed head → `bun run cli workflow test archon-ship --json` shows `"fixture": "sdlc/ship/fixtures/delivery-died-after-pr.stubs.yaml", "pass": true`; same command after editing `returns: flip-ready` → `returns: pr` in `.archon/workflows/sdlc/deliver/archon-deliver.yaml` → same fixture, same `"pass": true`; file restored and confirmed byte-identical via `cmp` afterward.

**Required outcome:** Either narrow the fixture's header comment so it doesn't claim protection it doesn't provide (it does prove the reachability/trigger-rule wiring — `deliver__gate-validated` is the sole failed node and `outcome` is reached — which has standalone value and should be described as such), or find a way to actually assert the resolved `delivered` value in this composed shape. The genuine "flip fails" and "flip is skipped upstream" value-refusal behavior is not otherwise fixture-tested anywhere in this suite; only the "flip runs and returns garbage" shape is (`delivered-unusable.stubs.yaml`, which does run the real script).

**Found by:** `prp-core:code-simplifier` (initial finding); independently reproduced and confirmed by the coordinator via direct mutation testing before being credited here. Rated Suggestion rather than Important because the coordinator independently proved the shipped behavior itself is correct (live DAG reproduction, not fixture-dependent) and five other fixtures already catch a `returns:` revert in the happy-path shape — this is a coverage gap in one composed shape, not a functional defect in what merges. Worth naming because it echoes this project's own review-posture rule almost verbatim: a test that "stubs the deciding function passes under every mutation of it and proves nothing" (`.archon/engineering.md`).

**Disposition:** FIXED at `ddd6816ee`. The fixture's header comment was rewritten to stop claiming value-level protection and instead accurately state what it pins: "this pins the post-creation death REACHING the died-mid-flight path at all, not the report's content," and explicitly names `delivered-unusable.stubs.yaml` (real script) and the five delivered fixtures (mutation-covered) as where the actual masquerade proof lives. I re-ran `bun run cli workflow test archon-ship --json` at the corrected head and confirmed `delivery-died-after-pr.stubs.yaml` still passes (`pass: true`) — the fix is comment-only, no behavior changed, exactly as intended.

</details>

<details>
<summary><code>R2</code> — Stale "JSON record" language in six places this PR's own change falsified</summary>

**Impact:** None on behavior — the code is correct everywhere. But six comments, three of them in files this PR itself edited, still describe `archon-deliver`'s returned value as a JSON pull-request record, directly contradicting the new `delivered_url()` function's own docstring 20-30 lines away in the same file (which correctly says the value is now a plain URL, "proof of the flip, not of creation"). A future maintainer trusting the stale prose over the correct docstring could reintroduce `json.loads(delivered)` or a `.url` field read — exactly the shape this PR just removed.

**Evidence:** `.archon/workflows/sdlc/ship/scripts/outcome.py:9` (`INPUTS_DELIVERED: deliver's returned pull request record as JSON text`), same phrase at `stabilize/scripts/outcome.py:12` and `upkeep/scripts/outcome.py:9`; a shorter inline echo at `ship/scripts/outcome.py:271`, `stabilize/scripts/outcome.py:251`, `upkeep/scripts/outcome.py:245` (`# Deliver ran, so the record it returned is the report.`); the same "pull request record the deliver branch returned" phrase, untouched by this PR, at `archon-ship.yaml:93-95`, `archon-stabilize.yaml:110-112`, `archon-upkeep.yaml:98-100`.

**Required outcome:** Rewrite the `INPUTS_DELIVERED` docstring bullet in all three `outcome.py` files and the "pull request record" comment in all three tail YAMLs to match what `delivered_url()`'s own docstring already correctly says.

**Class:** Invariant — any comment describing `archon-deliver`'s returned/bound value must match its actual post-fix shape (a URL string, not a JSON record). Search: `grep -rn "pull request record the deliver\|returned pull request record as JSON" .archon/workflows/`. 6 matches, all listed above, all confirmed stale by direct reading against the current `delivered_url()` behavior. No other matches exist repo-wide.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier` (all three independently converged on this), corroborated by the coordinator's own reading of the diff.

**Disposition:** FIXED at `ddd6816ee`. Re-ran the exact class-search grep from the required outcome (`grep -rn "pull request record the deliver\|returned pull request record as JSON" .archon/workflows/`) at the corrected head: zero matches, down from 6. All three `outcome.py` `INPUTS_DELIVERED` docstring bullets now read "the flip's verified pull request URL deliver returned"; all three tail YAMLs' comments now read "the flip's verified URL the deliver branch returned." Both now match `delivered_url()`'s own docstring. One residual noted for completeness, not tracked as a finding since it falls outside this finding's stated required outcome and the correction didn't touch it: the shorter inline echo "`# Deliver ran, so the record it returned is the report.`" directly above the `delivered_url()` call site in all three `outcome.py` files (`ship:271`, `stabilize:251`, `upkeep:245`) is still present, unchanged. It's cosmetic (the line above it is now correct) and not part of what this finding required, but worth a follow-up sweep if the pack gets touched again.

</details>

<details>
<summary><code>R3</code> — A genuinely failed flip silently drops the caveats/discoveries reporting channel</summary>

**Impact:** When the flip is merely *skipped* (an earlier gate failed first — the shape `delivery-died-after-pr.stubs.yaml` models), the composing tail's `outcome` node still runs to completion with `delivered=null`, taking the "died mid-flight" branch and printing `format_caveats()` (red-causes accepted and orphaned discoveries) alongside it — I confirmed this by reading `outcome.py`'s `main()` directly. But when the flip genuinely *fails* (the PR's own named "known risk," and the literal mechanism of the reported dogfood bug — `gh pr checks` refusing a non-green state) — `assertProducerNotFailed` throws while resolving the `delivered` binding, so `outcome.py` never launches at all. I confirmed `format_discoveries`/`format_red_causes`/`red_causes.json` are referenced nowhere outside the SDLC pack's own `outcome.py` scripts (`grep -rln` across `packages/cli/src packages/server/src` returns nothing), so there is no other channel that surfaces this content — the operator gets only a generic engine string (`Node '<id>' failed: <error>`) carrying flip-ready's own stderr text. On the old (buggy) `dev` code, this same scenario ran `outcome.py` to completion (since `$deliver.output` bound to the always-completed `pr` record), so the operator got a false "delivered" claim, but *did* get the caveats/discoveries sections alongside it. The PR's own "Known risk" note reasons only about the delivered-URL axis ("the flip error is a better report than a generic delivery-died line") and does not address that blocking `outcome` entirely also discards this separate, legitimate reporting channel — a channel both `dev` and the new code's own skipped-flip branch otherwise preserve.

**Evidence:** `assertProducerNotFailed` guard: `packages/workflows/src/dag-executor.ts:490-505`. Caveats/discoveries only produced inside `outcome.py`: `.archon/workflows/sdlc/{ship,stabilize,upkeep}/scripts/outcome.py`, `format_caveats()`/`format_discoveries()`/`format_red_causes()`, confirmed absent from `packages/cli/src` and `packages/server/src`. Directly reproduced: an ephemeral fixture (not committed, deleted after use) forcing the real `flip-ready` bash node to execute and genuinely fail reported exactly two failed nodes, `deliver__flip-ready` and `outcome` — `outcome.py`'s process never runs in that trace, so its caveats/discoveries printing is unreachable.

**Required outcome:** No code change required to consider this resolved — it is a legitimate operator judgment call (truthful-but-incomplete vs. false-but-complete), not a functional defect; the primary invariant this PR restores (never present a false "delivered" claim) holds regardless. Worth an explicit decision: either accept the trade as-is, or let `outcome` still run with `delivered=null` treatment when its own upstream flip genuinely failed (not just when skipped) so caveats/discoveries survive on this path too.

**Found by:** coordinator (verified independently via direct DAG-execution reproduction; not raised by any of the three specialist agents).

**Disposition:** FIXED (decision recorded) at `ddd6816ee`. A new comment block was added directly to `archon-deliver.yaml`'s `returns:` comment, explicitly naming the cost this finding raised: "when the flip itself FAILS, the composing tail's terminal node is blocked, so its caveat and discovery surfacing is skipped along with the false delivery claim — the operator gets the flip's own error, which names the recovery. Truthful failure beats a lying report; surfacing caveats on every failure path is terminal-result ownership the engine is growing toward, not something each tail should hand-roll." This is exactly the decision this finding asked for: the trade is now explicit and reasoned rather than silently absent from the PR's own risk analysis, and the team lead reports the datapoint was also posted to the engine-owned terminal-result tracking issue for the longer-term fix (caveats/discoveries surfacing regardless of which node fails). I did not independently verify that issue post — outside what a comment-diff delta-verify can confirm — but the comment itself is present and accurate at the reviewed head.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 (stale-docs observation deferred to comments/docs lens), R2 |
| `seams` | R2; traced the `returns:`/`depends_on` split through the engine and confirmed no fielded (`.url`/`.number`) consumer of the old shape survives anywhere in the repo |
| `simplify` | R1, R2; confirmed the URL-validation logic is proportionate to its concrete failure modes, no unnecessary machinery |

#### Validation — round 1 (head `725a9df1`)

|---|---|---|
| `bun run scripts/generate-bundled-defaults.ts --check` | PASS | "bundled-defaults.generated.ts is up to date (55 commands, 32 workflows, 11 scripts)" |
| `bun run type-check` | PASS | Every package (`@archon/paths`, `@archon/git`, `@archon/docs-web`, `@archon/providers`, `@archon/web`, `@archon/isolation`, `@archon/workflows`, `@archon/core`, `@archon/adapters`, `@archon/server`, `@archon/cli`) exits 0 |
| `bun run cli workflow test sdlc --json` | PASS | 45 passed, 0 failed — reproduced directly, matches the PR's own claim |
| Mutation: `returns: flip-ready` → `returns: pr`, full suite rerun, byte-copy restore | PASS (mutation caught) | Exactly 5 fixtures went red (`update-delivered`, `direct-delivered`, `rooted-delivered`, `planned-delivered`, `fix-delivered`); `.archon/workflows/sdlc/deliver/archon-deliver.yaml` restored and confirmed byte-identical via `cmp` |
| Mutation: same, targeted at `delivery-died-after-pr.stubs.yaml` only | FAIL (mutation not caught — see R1) | `pass: true` under both fixed and reverted `returns:` value |
| `delivered_url` byte-identical across ship/stabilize/upkeep | PASS | `cmp` on the extracted function body, all three files, 34 lines each |
| Live DAG reproduction of "genuine flip failure blocks composing tail's outcome" | PASS (confirms PR's own "known risk" claim) | Ephemeral fixture (deleted after use) forcing real `flip-ready` execution to genuinely fail (no `origin` remote in scratch worktree) reported exactly two failed nodes: `deliver__flip-ready`, `outcome` |
| `gh pr checks 3000` | PASS | Test Suite, changes, docker-build, postgres-parity, schema-upgrade, test (ubuntu-latest), test (windows-latest) — all green; CodeRabbit pass |
| `bun run lint` / `bun run format:check` / full `bun run test` | NOT RUN locally | Covered by CI (all green, evidence above); not rerun locally given this PR touches no linted/formatted source outside the regenerated TS mirror already verified in sync |

#### Validation — round 2, correction verification (head `ddd6816ee`)

|---|---|---|
| Bounded diff review: `git diff 725a9df1..ddd6816ee` | Comments-only confirmed | 9 files, +33/-26; every hunk is a comment or docstring line — no `nodes:`, `with:`, `depends_on:`, `script:`, or Python logic line touched |
| `bun run scripts/generate-bundled-defaults.ts --check` | PASS | "bundled-defaults.generated.ts is up to date (55 commands, 32 workflows, 11 scripts)" at the correction head |
| `bun run cli workflow test sdlc --json` | PASS | 45 passed, 0 failed, rerun at the correction head |
| R2 class-search regrep: `grep -rn "pull request record the deliver\|returned pull request record as JSON" .archon/workflows/` | 0 matches (was 6) | Confirms every site named in R2's required outcome is corrected |
| Pack-comment issue/run-reference scan on the correction diff | Clean | `git diff 725a9df1..ddd6816ee -- '.archon/workflows/**' \| grep '^+' \| grep -iE '#[0-9]{3,}\|run [0-9a-f]{8}\|3ac2b177\|2997'` — no matches |
| `gh pr checks 3000` (correction commit) | PASS (partial at time of writing) | schema-upgrade, postgres-parity, changes, CodeRabbit green; docker-build, test (ubuntu-latest), test (windows-latest) still running — a comments-only diff (confirmed above), so not expected to fail, but not yet observed green |

</details>

PR #3001, opened 2026-08-29
test(scripts): build the C5 registry once instead of once per test

### LINKED ISSUE #2575: Windows CI: the two C5 tests in migrate-state-dir.test.ts time out at 5000ms under parallel load

Wirasm:
## Problem and outcome

The two `subdirectory invocation (C5)` tests in `scripts/migrate-state-dir.test.ts` were the repository's worst live Windows flake: ten body timeouts against Bun's 5000 ms budget, out to 12296 ms, across seven unrelated branches. They red-lined PRs that had nothing to do with them, which trains everyone to ignore a red Windows check.

- **Outcome:** the C5 pair no longer carry a per-test cost their 17 siblings do not. Locally they drop from ~66 ms to ~57 ms, which is exactly the file's one-spawn floor; under 6x parallel load they run 75-105 ms while the other 17 reach 155-216 ms, so they are no longer the slowest tests in their own file.
- **Invariant:** a real registry database must still exist when the script runs, and the lookup must still run and still climb. That is what makes these two tests the guard that #2306's "skip the lookup when there is no database" optimization cannot silently disable project resolution. Rows still go in through the same `SqliteAdapter` the script's lookup opens, so the fixture cannot drift from the real schema.
- **Scope boundary:** the 5000 ms budget is untouched. No retries, no widened tolerance, no Windows skip. `scripts/migrate-state-dir.ts` is unchanged — this is not a product bug. The other 17 tests in the file are unchanged.
- **Root cause:** the fixture, not the subprocess. Each C5 test called `new SqliteAdapter` on a fresh path, which applies the entire product schema — 20 tables, 28 indexes, ~300 KB written through a WAL and checkpointed back on close — in order to store one row with two meaningful columns. `bun:sqlite` is synchronous, so all of it blocks the test's own event loop.

### The attribution this corrects

The file's docblock read the pair as ~15% above a one-spawn floor that a contended runner pushes over the line, and concluded there was nothing left to spend less of (#2924 filed it as "runner-floor residual"). #2982 step 4 called that falsified, and the CI evidence bears it out:

- **All 10 Windows failures are `this test timed out after 5000ms`** — Bun's body-overrun message. Zero hook timeouts, zero primary assertion failures. (The one `Expected: 0 / Received: 143` is downstream: Bun SIGTERMs the still-running child after the test is already marked failed.)
- **In those same jobs, on the same runner, the other 17 tests in the file never left 171-515 ms** — and they spawn the same script. A contended runner cannot slow one test 30x and leave the next one untouched.
- **Ubuntu runs these at 120-130 ms and has never failed them** across 51 jobs.
- The Windows pass distribution is a continuous gradient from ~400 ms to 3985 ms, and the two tests degrade independently within the same process — in 6 of 8 failing jobs one blew past 5 s while its sibling passed seconds later.

The "88% of the body is the subprocess" measurement that supported the old reading was taken on the wrong platform. On macOS the fixture really is ~6 ms of a ~60 ms body. On windows-latest it is ~230 ms of a ~400 ms floor, and it is the only part of the file with a tail.

## Review guidance

- **Feedback requested:** correctness of the isolation argument for sharing one registry across the two cases.
- **Start here:** `scripts/migrate-state-dir.test.ts` — the `beforeAll` in the `subdirectory invocation (C5)` group, and `sandboxFor()` above it.
- **Review order:** the rewritten BUDGET docblock (what was falsified and what replaces it), then `sandboxFor()`, then the two test bodies.
- **Known risk or uncertainty:** the Windows kernel primitive behind the tail is **not** proven, and the docblock says so rather than inventing one. Three candidates are ruled out and recorded so nobody re-runs them: Defender (this repo's own readback in PR #2943 — the runner image already excludes `C:\` and `D:\` recursively), fsync-per-commit (`bun:sqlite` opens WAL at `synchronous=NORMAL`, so commits do not fsync; measured 19 us per autocommit write), and a parent handle outliving `close()` (`lsof` shows every handle released before the child spawns). What is proven is where the cost lives, which is what this PR removes.

## Solution

The registry is built once for the group in `beforeAll`, with one row per case, instead of once per test. The test bodies now do exactly what the other 17 tests do: seed files, spawn the script, assert.

Isolation is preserved by making each case's name carry it. The name is both the case's repository directory and its registered project name, so each case gets its own `legacyDir` and its own `projectStateRoot` under the shared home. No two cases read or write the same file, and every path is a pure function of the case name rather than a rebound variable — so an assertion still running after its test timed out can only ever describe its own case, which is the property the file docblock has protected since PR #2513. What that docblock rules out is reassignment, not sharing; the docblock now says so explicitly.

One creation for the group is also the floor for this fixture. A per-case template copy, which the previous docblock considered and rejected, still creates a database per case and measured only 6.04 ms → 3.13 ms locally — it halves a cost rather than removing it, and with only two consumers there is nothing to amortize.

Building the fixture in a hook also makes any recurrence self-diagnosing, using #2924's own discriminator: a **hook** timeout would mean creating the database is itself the pathological operation, while a **body** timeout would mean a body that now does exactly what the other 17 do has stopped fitting — which is the whole file's floor, not a C5 question.

## Behavior change

| | Before | After |
|---|---|---|
| **Per-test fixture cost** | `new SqliteAdapter` on a fresh path per test: full schema, 20 tables, 28 indexes, ~300 KB via WAL | One shared registry built once in `beforeAll`; test bodies do no database work |
| **Local per-test duration** | 61-69 ms, against a 57 ms sibling floor | 49-61 ms, median 57 ms — at the floor |
| **Under 6x parallel load** | not measured on this branch | 75-105 ms, below the other 17 tests' 155-216 ms max |

## Validation

- `bun run validate` — **exit 0**. Covers `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, `check:api-types`, `type-check`, `lint --max-warnings 0`, `format:check`, `test:install` and the full `test` suite. `0 fail` is the only fail count anywhere in the log.
- `bun test ./scripts/` — the scripts battery the way CI runs it: `Ran 147 tests across 9 files. [2.06s]`.
- **Still fails on the fault it exists to catch.** Two induced faults in `scripts/migrate-state-dir.ts`, each reverted after:
  - the original C5 bug (destination climbs to the project root, source stays at the literal cwd — `const anchor = cwd`): **2 fail, 17 pass**, and the two failures are exactly the C5 pair.
  - the #2306 regression this pair specifically guards (`registryIsKnownEmpty()` returning true even when a database exists): **2 fail, 17 pass**, again exactly the C5 pair.
  - restored: **19 pass, 0 fail**.
- **25 consecutive green runs under parallel load** — 6 background workers looping the whole `scripts/` suite while the target file ran 25 times: `GREEN=25 RED=0`, with per-test JUnit durations captured on every run.
- Each C5 test also passes in isolation (`-t`), so the shared `beforeAll` does not create an ordering dependency.
- Before/after per-test durations were taken with `--reporter=junit` on the same machine, alternating via `git stash`, 6 runs each.
- **Not verified:** windows-latest. I cannot run it locally, so Windows confirmation has to come from this PR's own CI legs. The measurements above are macOS; the Windows numbers quoted throughout come from 59 `test (windows-latest)` job logs pulled with `gh api repos/coleam00/Archon/actions/jobs/<id>/logs` across the last 60 failed Test Suite runs, plus 12 jobs from successful runs for the pass-side distribution.

## Links

- Closes #2575
- Relates to #2982
- Relates to #2924

## Summary by CodeRabbit

* **Tests**
  * Improved coverage for invoking commands from subdirectories.
  * Consolidated test setup and cleanup for more consistent results.
  * Documented the handling of a Windows-specific test flake.

### DIFF
--- a/scripts/migrate-state-dir.test.ts
+++ b/scripts/migrate-state-dir.test.ts
@@ -18,7 +18,9 @@
  * assertions read the NEXT test's paths and reported a mutation that never
  * happened — on PR #2513 that surfaced as a dry run appearing to move a file.
  * Locals make an orphaned assertion able to describe only its own sandbox, so a
- * timeout reads as a timeout.
+ * timeout reads as a timeout. What that rules out is REASSIGNMENT, not sharing:
+ * the C5 group below shares one read-only registry built in `beforeAll`, and
+ * keeps the property by giving each case its own paths, derived from its name.
  *
  * BUDGET: deliberately left at Bun's 5000 ms default, even though these tests
  * time out on windows-latest at ~5015 ms (#2306). Raising it was considered and
@@ -38,21 +40,38 @@
  * their fixture spawned a SECOND cold `bun` process purely to write one registry
  * row, which is now an in-process write. The budget still has not moved.
  *
- * It recurred a third time (#2882, both C5 cases in one windows-latest run at
- * 6266/5016 ms, carrying Bun's body-timeout message rather than its hook-timeout
- * one). This time the body was phase-attributed and nothing accidental was left
- * to remove: spawning the migration script is 88% of it (61-68 ms of a 69-74 ms
- * body locally), and that subprocess IS the contract. The registry fixture is
- * the remaining 8-11%; building it once and copying the file per test measured
- * 2-3 ms cheaper and would have to either share one database across the two
- * cases — the cross-test coupling the paragraph above exists to prevent — or
- * rewrite the row anyway, because `default_cwd` is the per-test sandbox path.
- * The other 17 tests in this file sit at the same one-spawn floor (58-65 ms), so
- * the C5 pair are not special; they are ~15% above a floor that a contended
- * Windows runner puts at the budget line. That is a runner-capacity question,
- * not a cost this file can spend less of. Still do not reach for the timeout.
+ * It recurred a third time (#2882), and the reading recorded here then — that the
+ * pair sit ~15% above a one-spawn floor, so a contended runner is the whole story
+ * and there is nothing left to spend less of — was WRONG. #2982 falsified it and
+ * #2575 carries the correction. Two facts kill it. Ten of ten Windows failures
+ * are Bun's body-timeout message, out to 12296 ms; a test finishing 2.5x past its
+ * budget is not a 15% margin. And in those same runs, on the same runner, the
+ * other 17 tests in this file — which spawn the same script — never left
+ * 171-515 ms. A contended runner cannot slow one test in a file 30x and leave the
+ * next one untouched. The subprocess was never the cost.
+ *
+ * The registry fixture was. It read as 8-11% of the body because it was measured
+ * on the wrong platform: on macOS the whole fixture is ~6 ms of a ~60 ms body, on
+ * windows-latest it is ~230 ms of a ~400 ms floor, and it carries the file's only
+ * tail. `new SqliteAdapter` on a fresh path applies the entire product schema —
+ * 20 tables, 28 indexes, ~300 KB through a WAL and checkpointed back on close —
+ * synchronously, so it blocks the test's own event loop, which is why Bun's 5 s
+ * timer reported as late as 12296 ms. Building it once for the group instead of
+ * once per case is what this file can spend less of, and now does. Ubuntu, for
+ * scale, runs these at 120-130 ms and has never failed them.
+ *
+ * The Windows primitive behind the tail is NOT proven — do not write it down as
+ * though it were. Three candidates are already ruled out and should not be
+ * re-run: Defender (the runner image excludes C:\ and D:\ recursively, PR #2943),
+ * fsync-per-commit (bun:sqlite opens WAL at synchronous=NORMAL, so commits do not
+ * fsync — 19 us per autocommit write, measured), and a parent handle outliving
+ * `close()` (lsof shows every handle released before the child starts). If this
+ * recurs, the failure now discriminates on its own: a HOOK timeout means creating
+ * the database is itself the pathological operation, while a BODY timeout means a
+ * body that does exactly what the other 17 do has stopped fitting, which is the
+ * whole file's floor and not a C5 question. Still do not reach for the timeout.
  */
-import { describe, test, expect } from 'bun:test';
+import { describe, test, expect, beforeAll, afterAll } from 'bun:test';
 import { mkdtemp, mkdir, writeFile, readFile, readdir } from 'fs/promises';
 import { removeTempTree } from '@archon/paths/test-utils';
 import { getLogLevel, setLogLevel } from '@archon/paths';
@@ -457,7 +476,9 @@ describe('migrate-state-dir', () => {
     // the subdirectory*, declared "nothing to migrate", and wrote `.initialized`
     // into the REAL project's state root — disarming `state-preflight` for a
     // project whose state had never been migrated.
-    const PROJECT_NAME = 'acme/myrepo';
+    /** One registered project per case, so no two cases share a destination. */
+    const CASE_NAMES = ['climbs', 'ambiguous'] as const;
+    type CaseName = (typeof CASE_NAMES)[number];
 
     interface ProjectSandbox extends Sandbox {
       readonly subdir: string;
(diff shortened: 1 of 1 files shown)


Wirasm:
## Windows confirmation from this PR's own CI leg

The PR description said Windows verification had to come from CI because I cannot run windows-latest locally. It has: [`test (windows-latest)` job 99169785348](https://github.com/coleam00/Archon/actions/runs/33278617623/job/99169785348) — **success**, along with every other check.

The per-test durations are the result that matters. On windows-latest, this branch:

```
(pass) migrate-state-dir > subdirectory invocation (C5) > migrates the PROJECT root when invoked from a subdirectory [204.00ms]
(pass) migrate-state-dir > subdirectory invocation (C5) > refuses when BOTH the subdirectory and the project root hold legacy state [187.00ms]
```

Their 17 siblings in the same run: 172-360 ms. **The C5 pair are now at the middle of that band** — 204 and 187 ms against a 172 ms floor — and both are cheaper than the file's two slowest tests (359 and 360 ms).

For comparison, the same two tests before this change:

| | Before (windows-latest) | After (windows-latest) |
|---|---|---|
| C5 floor | 375-406 ms | **187-204 ms** |
| C5 passing range | 375 → 3985 ms | 187-204 ms |
| C5 failures | 5063 → 12296 ms, 10 samples across 7 branches | none |
| Other 17 tests | 171-515 ms | 172-360 ms (unchanged) |
| `scripts/` suite total | 6.87-8.50 s healthy, 13.10-22.71 s in every C5-failing run | **6.83 s** |

That lands where the attribution predicted. The removed per-test fixture was measured at roughly 230 ms of a ~400 ms Windows floor; taking it out of the test bodies moved the floor to 187-204 ms, and the pair no longer stand out from tests that do the same one-spawn work.

One green run is not proof the tail is gone — the failure rate was roughly 10 of 72 C5 executions in failed-run jobs, so a single pass was always likely. What it does establish is the floor, which is the thing the attribution predicted and the thing the change was aimed at. The tail lived on top of that floor, and the operation that carried it no longer runs per test.

Wirasm:
## Ready to merge

The PR builds the two "subdirectory invocation (C5)" tests' SQLite registry once in `beforeAll` instead of once per test, keeping isolation by deriving every path each case touches from that case's own name. I traced the shared-registry read path end to end (`findCodebaseByPathPrefix` does a full-table scan but is boundary-anchored on the path separator) and confirmed the two chosen case names (`climbs`, `ambiguous`) cannot cross-match. I reproduced both regressions the pair is meant to catch (the historical anchor-climbing bug and the `registryIsKnownEmpty` regression) against the reviewed head: each drives exactly the C5 pair red and the other 17 tests stay green, with the file byte-restored and re-verified clean after each. Both C5 tests pass standalone via `-t`, and an injected `beforeAll` failure surfaces as one immediately-attributable hook error rather than two confusing body-timeout failures — confirming the hook placement does what the PR claims for #2924's discriminator. The Windows CI evidence checks out: job 99169785348 is green, and its C5 durations (204ms / 187ms) match the PR's own follow-up comment exactly. Two Suggestions remain, both about hardening the file against a *future* edit, not about anything wrong with what's shipped.

**Validation:** all applicable checks pass; ESLint doesn't apply (`*.test.ts` is globally ignored by this repo's `eslint.config.mjs`, not a gap introduced by this PR).

### Findings

|---|---|---|---|
| `R1` | Suggestion | Case-name uniqueness across the shared registry is enforced by convention, not a type or runtime check | OPEN |
| `R2` | Suggestion | "Two cases therefore never read or write the same file" overstates the isolation: both cases' subprocesses do read the one shared `archon.db` | OPEN |

<details>
<summary><code>R1</code> — case-name uniqueness is a comment, not a check</summary>

**Impact:** none today — `CASE_NAMES = ['climbs', 'ambiguous']` are genuinely disjoint, verified by tracing `findCodebaseByPathPrefix`'s boundary-anchored match and by direct fault injection. The risk is forward-looking: `CASE_NAMES` (the seeding list) and each test body's `sandboxFor('name')` call are two independently maintained references to the same set. If a third C5 case is ever added (the `ambiguous` test's own comment at line 604 gestures at exactly such a case, "C5 one level down") and a name is copy-pasted without updating, `sandboxFor` silently returns the same `repo`/`legacyDir`/`projectStateRoot` as an existing case — `CaseName` still type-checks a reused literal — reintroducing the cross-test coupling this file's whole docblock history (#2306, #2513, #2575, #2882, #2982) exists to prevent, through name aliasing instead of variable reassignment.

**Evidence:** `scripts/migrate-state-dir.test.ts:480-481` (`CASE_NAMES`), `:504-516` (`sandboxFor`), `:584` and `:602` (independent literal call sites).

**Required outcome:** not required for this PR to merge — the shipped pair is correct. Worth a one-line guard (`if (new Set(CASE_NAMES).size !== CASE_NAMES.length) throw ...` in `beforeAll`) whenever a third case is added, given this file's specific incident history with exactly this defect class.

**Disposition:** OPEN, non-blocking. Declining to require a fix now — YAGNI applies to a two-member array with no third case in this PR; the guard is cheap enough to add opportunistically the next time this describe block is touched.

</details>

<details>
<summary><code>R2</code> — "never read... the same file" is imprecise</summary>

**Impact:** none today — no assertion depends on the wording being literal, and the actual behavior is safe (the shared `archon.db` is written once in `beforeAll` and never written again). But the claim at lines 500-502 ("Two cases therefore never read or write the same file") is not literally true: both cases' spawned subprocesses read the same `archon.db`, because `findCodebaseByPathPrefix` issues `SELECT * FROM remote_agent_codebases` — every row, not just the caller's own. The docblock elsewhere correctly says the registry is shared "read-only" (line 549), so the intended claim is "never *mutate* the same file," not "never touch." A future reader taking the stronger literal claim at face value could reason incorrectly about what's actually shared.

**Evidence:** `scripts/migrate-state-dir.test.ts:497-502` (the claim) vs. `packages/core/src/db/codebases.ts:128-138` (`findCodebaseByPathPrefix`'s full-table `SELECT`).

**Required outcome:** not required for this PR to merge. A future touch to this comment could narrow it to "never read or write each other's *state* files" or similar, to match the more precise "read-only" language two paragraphs up.

**Found by:** coordinator (direct verification against the isolation-honesty question)

**Disposition:** OPEN, non-blocking — wording precision only, no behavioral consequence.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `seams` | R1 |
| `simplify` | No additional findings (one non-finding note: moving fixture construction into `beforeAll` means a hook failure now fails both C5 tests together instead of independently — an intended consequence of "build once," not an oversight) |

#### Validation

|---|---|---|
| `bun test ./scripts/migrate-state-dir.test.ts` | PASS | 19 pass, 0 fail, 76 expect() calls — reconfirmed after every fault-injection restore |
| `bun test ./scripts/` (full battery, matches CI invocation) | PASS | 147 pass across 9 files, 3 consecutive runs (via `prp-core:code-reviewer`) |
| `bunx tsc --noEmit -p scripts/tsconfig.json` | PASS | clean, no diagnostics |
| `bun run scripts/check-test-cleanup-drift.ts` | PASS | passed (via `prp-core:code-reviewer`) |
| `bunx prettier --check scripts/migrate-state-dir.test.ts` | PASS | "All matched files use Prettier code style!" |
| ESLint on the changed file | NOT RUN (not applicable) | `eslint.config.mjs:27` globally ignores `**/*.test.ts`; pre-existing repo-wide exclusion, not caused by this PR |
| Fault injection: `const anchor = cwd` (historical C5 anchor-climb bug) | PASS (reproduces as claimed) | 2 fail / 17 pass, failures are exactly the C5 pair; file byte-restored, `git diff` clean after |
| Fault injection: `registryIsKnownEmpty()` returns true unconditionally (#2306 regression) | PASS (reproduces as claimed) | 2 fail / 17 pass, failures are exactly the C5 pair; file byte-restored, `git diff` clean after |
| Standalone execution: `bun test ./scripts/migrate-state-dir.test.ts -t "<each C5 test>"` | PASS | 1 pass each, 110ms / 109ms — no ordering dependency, no inflation from unrelated fixtures |
| Fault injection: `beforeAll` throws | PASS (surfaces as claimed) | one hook-level failure (`(unnamed)`) pointing directly at the throw site, not a cascade of two body-timeout failures; file byte-restored, `git diff` clean after |
| Timing direction, local (file-swap alternation, dev vs. PR version) | PASS (directionally consistent) | dev: 63.4ms / 63.8ms → PR: 62.4ms / 59.2ms per C5 test; small local delta matches the PR's own claim that the removed fixture is cheap on macOS |
| Windows CI evidence | PASS | `gh api .../jobs/99169785348`: `test (windows-latest)`, conclusion `success`; job log shows C5 at 204.00ms / 187.00ms, matching the PR's own comment exactly, within the other 17 tests' 172-360ms band |
| `#2513` attribution in the docblock | VERIFIED ACCURATE | PR #2513 (`fix(db): drop the premature event_order index...`) is unrelated in subject but is genuinely where the reassignment-driven phantom-mutation symptom surfaced in CI; the actual fixing commit `b2a6e8ca5` ("Refs #2306") establishes exactly the reassignment-specific rule this PR's new docblock text claims |
| Defender/fsync/parent-handle "ruled out" claims | VERIFIED | Defender exclusion corroborated by PR #2943; "NOT proven" framing for the Windows kernel primitive is explicitly present in the shipped docblock text (lines 63-64) |
| Budget/retries/skip scope boundary | VERIFIED UNTOUCHED | full diff review: BUDGET docblock still cites Bun's 5000ms default unchanged; no retries, tolerance widening, `.skip`, `xtest`, or platform-conditional skip anywhere in the diff |

</details>

PR #3046, opened 2026-08-31
Fix empty target labels in archon-ship

### LINKED ISSUE #3031: archon-ship passes a dangling 'Original work item:' label to three segments
## Problem

`archon-ship` interpolates `Original work item: $INPUTS.target` into the prose it passes to three composed segments — `inv` (`archon-ship.yaml:44`), `planned` (`:69`) and `deliver` (`:88`). The input declares `target` with `default: ""` and documents that *"when empty, the run's trigger message is the target."*

That promise is kept for `triage`, which receives `target` raw at `:35` and has an `$ARGUMENTS` fallback in its command file. It is **not** kept for the three prose interpolations: nothing substitutes the trigger message into them, so on the default launch path all three render as the label `Original work item:` followed by nothing.

The engine cannot flag it. `$INPUTS.target` resolved correctly, to the empty string it was declared to default to.

Verified on a real run: `3ac2b177` (`archon-ship`, launched as `issue #2377`) recorded no bound inputs at all — `metadata.inputs_values` and `metadata.inputs` are both empty.

## Why it is worth solving, and why now

PR #3030 made `$ARGUMENTS` a real parallel channel in `implement`, `plan`, `investigate` and `triage`, with a stated precedence rule. The operator's words now reach every reasoning node directly. So the prose pointer has stopped being the only route for intent and started being a second copy — and a second copy that is blank most of the time.

Two declarations carrying the same thing, kept in agreement by nobody, is the class `AGENTS.md` now names as a present defect. This is a small, contained instance of it, and it is cheapest to settle while the reasoning behind #3030 is still fresh.

## The distinction an implementer must not miss

`target` is not always empty, and the two cases behave differently:

| launch | `$INPUTS.target` | `$ARGUMENTS` | the prose renders |
| --- | --- | --- | --- |
| `run archon-ship "fix issue #2377"` | `""` | `fix issue #2377` | `Original work item:` — dangling |
| `run archon-ship "..." --with target="issue #2377"` | `issue #2377` | the trigger message | `Original work item: issue #2377` — meaningful, and **not** the same string as `$ARGUMENTS` |

So this is not dead code. It is correct in the second case and produces a dangling label in the first. Deleting it outright is defensible — the operator's message now reaches those nodes directly, and `triage.md`'s "Source and outcome" section records what the target was either way — but it does drop an explicit pointer in the bound case. That trade is the decision this issue exists to make; do not assume deletion without weighing it.

`:35` (`target: "$INPUTS.target"` into triage) is **out of scope and must not change.** Triage depends on receiving a genuinely empty string so its own `$ARGUMENTS` fallback fires. That is the mechanism that made the promise true in the first place.

## Desired outcome

A composed segment never receives a prose label with nothing after it, and the operator's original work item is carried exactly once rather than by two channels that can disagree.

## Invariants

- `triage` keeps receiving `target` raw at `:35`, empty when unbound.
- The `$ARGUMENTS` precedence channel added by #3030 is unchanged in all four command files.
- A run launched **with** `--with target=...` still conveys that target to the segments that need it, by some route — the artifact, the prose, or a deliberate decision that `$ARGUMENTS` plus `triage.md` is sufficient.
- No new YAML surface. The workflow language has no conditional interpolation and should not grow one for this.

## Acceptance

- On a default launch (`target` unbound), no composed segment's prompt contains a label with an empty value after it.
- On a launch with `target` bound, the target still reaches `inv`, `planned` and `deliver` by a route named in the change.
- `archon-ship.yaml:35` is unchanged.
- The pack fixtures pass, and one covers the default-launch path so a future edit cannot silently reintroduce the dangling label.

## Solution steering (optional)

Deletion of the three prose interpolations is the smallest change and is probably right, given #3030. If you keep them, the label needs to not render when the value is empty, and the language deliberately has no mechanism for that — which is itself an argument for deletion rather than for adding one.

## Links

- Related #3030 — made `$ARGUMENTS` the direct channel that turns this into a duplicate.

Wirasm:
## Problem and outcome

`archon-ship` appended `Original work item: $INPUTS.target` to its investigation, planning, and delivery prompts. On the default launch path, `target` is empty because the trigger message is resolved by triage, leaving a dangling empty label downstream.

- **Outcome:** All three downstream stages now use `triage.md` as their work-order boundary, so default launches do not render an empty original-work-item label.
- **Invariant:** A bound target still follows the named route `target → triage → triage.md → inv/planned/deliver`.
- **Scope boundary:** The raw `target: "$INPUTS.target"` binding into triage, the precedence channel from #3030, and YAML interpolation behavior are unchanged.
- **Root cause:** Downstream prompt text interpolated the optional raw target after triage had already established the authoritative source-and-outcome artifact.

## Review guidance

- **Feedback requested:** Confirm `triage.md` remains the sole downstream work-order boundary for each actionable route.
- **Start here:** `.archon/workflows/sdlc/ship/archon-ship.yaml:40` — removes the empty-label source while retaining the raw triage binding.
- **Review order:** Review the ship workflow, then the bound/default fixture coverage, then the generated-defaults source-contract test.
- **Lower-attention areas:** `bundled-defaults.generated.ts` is regenerated from the workflow source; `bun run check:bundled` verified it is current.

## Solution

Removed the redundant raw-target suffixes from the `inv`, `planned`, and `deliver` bindings. Each already requires `triage.md` and now receives its verified source, outcome, and constraints only through that artifact. The default fixture remains unbound, while the investigation and planning fixtures bind `issue #3031` to prove both launch cases reach their delivery routes.

## Behavior change

| | Before | After |
|---|---|---|
| **Default launch** | Downstream prompts rendered an empty `Original work item:` label. | Downstream prompts read the triage artifact without empty raw-target prose. |
| **Bound launch** | The target reached triage and also appeared redundantly downstream. | The target reaches triage; `triage.md` carries the authoritative work order to downstream stages. |

## Validation

- `bun run cli workflow test archon-ship --default-stubs --exec-code` — 9 passed, 0 failed — proves default and bound fixture routes complete.
- `bun run check:bundled` — passed — proves generated bundled defaults are current.
- `bun test src/defaults/` and `bun test src/fixture-runner.test.ts` from `packages/workflows` — 49 and 43 passed — prove the source contract and fixture runner behavior.
- `bun run validate` — passed — verifies generated artifacts, types, lint, formatting, installer tests, package tests, and script tests.
- **Not verified:** Nothing material; workflow-route and source-contract coverage exercise both launch cases.

## Links

- Closes #3031

## Summary by CodeRabbit

- **Improvements**
  - Shipping workflows now carry the requested work item through triage, investigation, planning, and delivery more consistently.
  - Downstream stages rely on the verified triage record, reducing duplicated or conflicting work-item details.
  - Direct and planned delivery paths now provide clearer handling of bound targets and triage outcomes.

- **Validation**
  - Workflow checks now detect missing or unsupported inputs and verify that expected information appears in completed workflow steps.


Wirasm:
# Review report

## Verdict

Ready — `none`. Both prior Important findings are fixed and verified on the correction delta; the only remaining item is a non-blocking documentation suggestion for the new fixture assertion.

## Accepted contract

`archon-ship` must not send an empty `Original work item:` label to downstream segments on an unbound default launch. A bound target must retain `target → triage → triage.md → inv/planned/deliver`, and pack fixtures must prove both launch cases. Preserve the raw `target: "$INPUTS.target"` triage binding and the `$ARGUMENTS` precedence channel; do not add YAML interpolation machinery. `triage.md` is the downstream work-order boundary, with `investigation.md` or `plan.md` additionally serving as the detailed delivery work order on routed paths.

## Reviewed head SHA

`04729653ac0cadac1ec0d4f9dfb77a5aaa51f6b7`

## Findings

### Critical

None.

### Important

None open.

### Rejected findings

None.

### Suggestions

#### R3 — document the resolved-text fixture assertion

Sources: `[synthesize]`

The correction adds the public `fixture.resolved-text-contains` YAML field at `packages/workflows/src/fixture-runner.ts:61-68`, but the workflow-test/dry-run reference documents fixture stubs and their input behavior without describing this assertion at `packages/docs-web/src/content/docs/reference/cli.md:312-331`. Add a short fixture example and explain that it asserts a completed or stubbed node's resolved text contains the supplied fragment. This does not block the accepted ship-workflow fix.

## Prior findings

| ID | Sources | Prior claim | Status at `04729653ac0cadac1ec0d4f9dfb77a5aaa51f6b7` |
| --- | --- | --- | --- |
| R1 | `[seams]` | Fixture inputs bypass the declared-input contract. | Fixed. `checkFixture` now resolves `fixture.inputs` through `resolveTopLevelInputs` before dry-run execution (`packages/workflows/src/fixture-runner.ts:564-579`). The focused undeclared and missing-required cases assert rejection (`packages/workflows/src/fixture-runner.test.ts:430-460`). |
| R2 | `[tests]` | Terminal fixture outcomes do not prove either required launch-input case. | Fixed. The runner checks declared resolved-text fragments on completed or stubbed trace entries (`packages/workflows/src/fixture-runner.ts:608-624`); the direct fixture requires the empty target and both bound routes require `issue #3031` (`.archon/workflows/sdlc/ship/fixtures/direct-delivered.stubs.yaml:35-38`, `.archon/workflows/sdlc/ship/fixtures/planned-delivered.stubs.yaml:36-41`, `.archon/workflows/sdlc/ship/fixtures/rooted-delivered.stubs.yaml:58-63`). |

## Discoveries

0 accepted discoveries. [JSON](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/b963ec628672e842637d149df8b69277/discoveries.json) · [human report](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/b963ec628672e842637d149df8b69277/discoveries.md)

## Review coverage

One continuation reviewer verified the prior report and the correction delta. The original selected concerns were the declared-input ingress seam (R1) and proof that fixtures exercise the bound and default target handoff (R2). Both corrections were inspectable; no required evidence was unavailable.

The delta added the `fixture.resolved-text-contains` YAML field, so the previously disabled `errors` and `docs` lenses were re-evaluated against that delta alone. Error behavior is clear and covered: invalid fixture inputs are rejected before simulation and assertion failures identify the node and missing fragment (`packages/workflows/src/fixture-runner.ts:564-624`, `packages/workflows/src/fixture-runner.test.ts:430-484`). The docs re-evaluation produced non-blocking suggestion R3; no further correction round is warranted for it.

Validation passed: `bun --cwd packages/workflows test src/fixture-runner.test.ts` (134 passed), `bun run cli workflow test archon-ship --json` (9 passed), `bun run check:bundled`, and `git diff --check 406b9385f0312024ce19708d80207b2096255c91..04729653ac0cadac1ec0d4f9dfb77a5aaa51f6b7`.

PR #2948, opened 2026-08-28
fix(cli): accept run-id prefixes for --adopt

Wirasm:
## Problem and outcome

`workflow run --adopt` rejected unique run-id prefixes that other run-id CLI verbs accept, even though `workflow runs` displays those prefixes. A full id worked, making normal adoption recovery needlessly fail.

- **Outcome:** `--adopt` accepts a unique run-id prefix in foreground and detached runs, while ambiguous prefixes identify their candidates.
- **Invariant:** Every CLI surface that accepts a run id accepts the same forms, including prefixes printed by `workflow runs`.
- **Scope boundary:** This changes adoption-id resolution only; superseding and existing full-id behavior are unchanged.
- **Root cause:** Adoption bypassed the existing scoped run-id prefix resolver and passed the raw CLI value directly to adoption resolution.

## Review guidance

- **Feedback requested:** Correctness of resolving once before detached-run creation and forwarding the canonical id to the child.
- **Start here:** `packages/cli/src/commands/workflow.ts:2090` — detached adoption now resolves the supplied id before preflight, persistence, and child argument construction.
- **Review order:** `packages/cli/src/commands/workflow.ts`, then the foreground resolver at `packages/cli/src/commands/workflow.ts:2349`, then the focused coverage in `packages/cli/src/commands/workflow.test.ts:5479` and `packages/cli/src/commands/workflow.test.ts:10003`.
- **Known risk or uncertainty:** None.

## Solution

Reuse the existing scoped `resolveRunIdArg` primitive at both adoption entry points. The detached parent uses the canonical id for preflight, the pre-created run's provenance, and the child `--adopt` argument; the foreground lane resolves through the same primitive. This keeps all adoption stages bound to one id without introducing another resolver.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `workflow run --adopt <unique-prefix>` rejected the prefix. | It resolves the prefix and adopts that run. |
| **Failure behavior** | The raw prefix produced an unknown-run error. | Ambiguous prefixes list matching full ids; unknown ids retain their existing downstream behavior. |

## Validation

- `bun test src/commands/workflow.test.ts` in `packages/cli` — passed (334 tests), including unique-prefix adoption in foreground and detached modes, ambiguous-prefix candidates, and unchanged full ids.
- `bun run type-check` — passed.
- `bun run lint --max-warnings 0` — passed.
[...]

PR #3047, opened 2026-08-31
fix(cli): --supersedes accepts short run-id prefixes like --adopt

### LINKED ISSUE #2990: fix(cli): --supersedes rejects the run-id prefixes --adopt now accepts
## Problem

PR #2948 (closing #2871) made `--adopt <run-id>` resolve unique run-id prefixes through the shared `resolveRunIdArg` resolver, the same one `workflow get`/`cancel`/`abandon`/`resume`/etc. already used. `--supersedes <run-id>` is the structural twin of `--adopt` — same CLI surface, same `<run-id>` documentation in `cli.ts`, handled in the same `--adopt`/`--supersedes` conditional blocks throughout `workflow.ts` — but it still resolves through `resolveSupersededRun`, which calls `workflowDb.getWorkflowRun(runId)` for an exact match only:

```ts
// packages/cli/src/commands/workflow.ts
async function resolveSupersededRun(runId: string): Promise<WorkflowRun> {
  const superseded = await workflowDb.getWorkflowRun(runId);
  if (!superseded) {
    throw new Error(`Cannot supersede: no workflow run '${runId}' exists.`);
  }
  ...
}
```

So `archon workflow run <name> --supersedes b4e9597b` — using the exact short id `workflow runs` prints, exactly as `--supersedes <run-id>` is documented — fails with `Cannot supersede: no workflow run 'b4e9597b' exists.` even though the run exists. This is the same live-incident shape #2871 fixed for `--adopt`, now reproducible on `--supersedes` instead.

Both call sites in `workflow.ts` (detached preflight ~line 2107/2200, foreground ~line 2354) pass `options.supersedesRunId` straight through unresolved. There is no test coverage for `--supersedes` id resolution in `workflow.test.ts`.

## Desired outcome

`--supersedes` resolves run-id prefixes through the same `resolveRunIdArg` primitive `--adopt` now uses, mirroring the fix in #2948/PR #2948 exactly: unique prefix resolves, ambiguous prefix errors listing candidates, unknown id errors as today, full ids unchanged.

## Scope boundary

This is resolution only, following the pattern #2948 established for `--adopt`. No other `--supersedes` behavior (terminality check, provenance recording) changes.

## Related

- #2871 (closed by #2948) — same fix, `--adopt` only
- #2611 — original shared-resolver rollout that also missed `--adopt`/`--supersedes`

Found during review of PR #2948.

Wirasm:
## Problem and outcome

`--adopt` resolves short run-id prefixes (merged in #2948), but `--supersedes` still demanded a full run id, so the documented short run IDs were unusable on this path and the two continuation flags behaved inconsistently.

- **Outcome:** `--supersedes` now accepts a unique run-id prefix and behaves like `--adopt` for the shared resolver cases: unique prefixes resolve, ambiguous prefixes are refused with the candidate list, full ids pass through unchanged, and unknown prefixes keep the existing refusal.
- **Invariant:** `resolveSupersededRun`'s existence and terminality checks, and both refusal messages, behave exactly as before.
- **Scope boundary:** No change to terminality checks, provenance recording mechanism, or any other `--supersedes` behavior.
- **Root cause:** The two `--supersedes` call sites passed the raw flag value through instead of routing it through the shared `resolveRunIdArg` resolver that `--adopt` already used.

## Review guidance

- **Feedback requested:** Correctness of prefix resolution at both call sites, and that the detached path threads the resolved id everywhere the adopt lane does.
- **Start here:** `packages/cli/src/commands/workflow.ts:2105` — the detached pre-flight now resolves the prefix before the terminality check and threads `supersededRunId` into the continuation declaration, provenance, and the child's `--supersedes` argv flag, mirroring `adoptedRunId`.
- **Review order:** First `packages/cli/src/commands/workflow.ts` (both call sites, ~line 2105 and ~line 2374), then the new tests in `packages/cli/src/commands/workflow.test.ts` under `workflowRunCommand — supersedes run-id prefix resolution (#2990)`.
- **Lower-attention areas:** The `resetSharedMocks` helper in the new test describe is mechanical test hygiene: 340 earlier tests in that file leave queued `mockOnce` values on the shared module mocks, and without the reset the new tests consumed stale queue entries and failed only in the full-file run.
- **Known risk or uncertainty:** One pre-existing flaky integration test (`workflow-terminal-event.integration.spec.ts`) failed once during full-suite validation and passed on re-run and in isolation; it is unrelated to this change.

## Solution

Both `--supersedes` call sites now resolve the flag value through the shared `resolveRunIdArg` resolver, using the same call shape the `--adopt` calls added in #2948 (`resolveRunIdArg(id, cwd, false, codebase.id)` — prefix allowed, not required). In the detached pre-flight the resolution happens before `resolveSupersededRun`, so refusal messages and the child's argv carry the resolved full id, and the pre-created row's provenance records the full id rather than an unresolved prefix. In the foreground path the resolved id flows into the same terminality check and provenance recording as before. This reuses the existing primitive with no new code paths, which is the smallest coherent fix for the inconsistency.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `--supersedes 0b1ee8da` failed; only a full run id was accepted | A unique prefix resolves to the full id and the run proceeds; the child receives the full id |
| **Failure behavior** | A prefix produced `no workflow run '<prefix>' exists` even when the run did exist | Ambiguous prefixes list their candidates; unknown prefixes keep the existing refusal; non-terminal runs are still refused |

## Validation

- `bun test src/commands/workflow.test.ts -t supersedes` — 7 pass. Pre-fix (with the source change stashed): 4 fail (unique prefix, ambiguous prefix, detached resolution, detach pre-flight non-terminality) and 3 pass, confirming the tests cover the fix.
- `bun run test` in packages/cli — all files, 0 failures (the flaky integration test above passed on re-run and in isolation).
- `bun run type-check` in packages/cli — clean.
- `bun run validate` at repo root — exit 0, zero failing tests across all packages.
- **Not verified:** Nothing material; the resolver cases (unique, ambiguous, unknown, full id, detach pass-through) and the unchanged terminality refusal are each covered by a focused test.

## Links

- Closes #2990
- Related #2948

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving shortened workflow run IDs supplied with `--supersedes`.
  * Short IDs are expanded when they uniquely identify a run within the relevant codebase.
  * Ambiguous, unknown, or non-terminal run IDs are now rejected with clear feedback.
  * Detached workflow runs consistently forward the resolved full run ID.

### DIFF
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -10455,6 +10455,55 @@ describe('workflowRunCommand — supersedes run-id prefix resolution (#2990)', (
       expect.objectContaining({ adopted_from_run_id: supersededRunId })
     );
   });
+
+  it('refuses a non-terminal supersedes run in the detach pre-flight before forking', async () => {
+    const supersededRunId = '0b1ee8da-1111-2222-3333-444455556666';
+    const { discoverWorkflowsWithConfig } = await import('@archon/workflows/workflow-discovery');
+    const codebaseDb = await import('@archon/core/db/codebases');
+    const workflowDb = await import('@archon/core/db/workflows');
+    const paths = await import('@archon/paths');
+    (discoverWorkflowsWithConfig as ReturnType<typeof mock>).mockResolvedValueOnce({
+      workflows: [makeTestWorkflowWithSource({ name: 'assist', description: 'Help' })],
+      errors: [],
+    });
+    (paths.getArchonHome as ReturnType<typeof mock>).mockImplementationOnce(() => {
+      throw new Error('no home in test');
+    });
+    (codebaseDb.findCodebaseByDefaultCwd as ReturnType<typeof mock>).mockResolvedValueOnce(null);
+    (codebaseDb.findCodebaseByPathPrefix as ReturnType<typeof mock>).mockResolvedValueOnce({
+      id: 'cb-1',
+      name: 'test/folder',
+      default_cwd: '/test/path',
+      kind: 'folder',
+    });
+    (workflowDb.findWorkflowRunsByIdPrefix as ReturnType<typeof mock>).mockResolvedValueOnce([
+      { id: supersededRunId },
+    ]);
+    // Argument-aware on purpose: the queued-value mocks are argument-blind, which
+    // would let terminality pass here even if it ran on the raw prefix.
+    (workflowDb.getWorkflowRun as ReturnType<typeof mock>).mockImplementation(async (id: string) =>
+      id === supersededRunId ? { id: supersededRunId, status: 'running' } : null
+    );
+
+    const spawnSpy = spyOn(Bun, 'spawn');
+    try {
+      // The refusal names the resolved full id, so terminality is checked on the
+      // resolution, not on the raw prefix.
+      await expect(
+        workflowRunCommand('/test/path/subdir', 'assist', 'hello', {
+          detach: true,
+          supersedesRunId: '0b1ee8da',
+        })
+      ).rejects.toThrow(`Cannot supersede run '${supersededRunId}': it is still running.`);
+    } finally {
+      spawnSpy.mockRestore();
+    }
+
+    // The refusal reached the parent: no child was forked and no pending run row
+    // was left behind for a launch that cannot proceed (#2872).
+    expect(spawnSpy).not.toHaveBeenCalled();
+    expect(mockCreateWorkflowRun).not.toHaveBeenCalled();
+  });
 });
 
 describe('workflowWaitCommand', () => {
--- a/packages/cli/src/commands/workflow.test.ts
+++ b/packages/cli/src/commands/workflow.test.ts
@@ -10210,6 +10210,253 @@ describe('workflowRunCommand — adopt lane source recapture (#2660/#2747)', () 
   });
 });
 
+describe('workflowRunCommand — supersedes run-id prefix resolution (#2990)', () => {
+  let consoleSpy: ReturnType<typeof spyOn>;
+
+  // The suites above leave queued mockOnce values and call history behind on the shared
+  // module mocks. Re-install the module-mock defaults so this describe starts from a
+  // clean queue and leaves one behind for the suites after it.
+  function resetSharedMocks(): void {
+    const discoverMock = require('@archon/workflows/workflow-discovery')
+      .discoverWorkflowsWithConfig as ReturnType<typeof mock>;
+    discoverMock.mockReset();
+    discoverMock.mockImplementation(() => Promise.resolve({ workflows: [], errors: [] }));
+    const codebaseDb = require('@archon/core/db/codebases');
+    (codebaseDb.findCodebaseByDefaultCwd as ReturnType<typeof mock>).mockReset();
+    (codebaseDb.findCodebaseByDefaultCwd as ReturnType<typeof mock>).mockImplementation(() =>
+      Promise.resolve(null)
+    );
+    (codebaseDb.findCodebaseByPathPrefix as ReturnType<typeof mock>).mockReset();
+    (codebaseDb.findCodebaseByPathPrefix as ReturnType<typeof mock>).mockImplementation(() =>
+      Promise.resolve(null)
+    );
+    const conversationDb = require('@archon/core/db/conversations');
+    (conversationDb.getOrCreateConversation as ReturnType<typeof mock>).mockReset();
+    (conversationDb.getOrCreateConversation as ReturnType<typeof mock>).mockImplementation(() =>
+      Promise.resolve({
+        id: 'conv-123',
+        platform_type: 'cli',
+        platform_conversation_id: 'cli-123',
+      })
+    );
+    (conversationDb.updateConversation as ReturnType<typeof mock>).mockReset();
+    (conversationDb.updateConversation as ReturnType<typeof mock>).mockImplementation(() =>
+      Promise.resolve(undefined)
+    );
+    const workflowDb = require('@archon/core/db/workflows');
+    (workflowDb.getWorkflowRun as ReturnType<typeof mock>).mockReset();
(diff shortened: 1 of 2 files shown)


Wirasm:
# Review report — PR #3047 (round 2, continuation)

## Verdict

**Ready. Action: `none`.** The single round-1 finding (R1) is corrected: commit `a0f9a379` adds the detach pre-flight terminality regression test, and the shipped test falsifies both regressions R1 identified; the production diff is unchanged and all 347 tests in the file pass.

## Accepted contract

Carried unchanged from `review/scope.md`:

- **Required outcome:** `--supersedes` resolves run-id prefixes through the shared `resolveRunIdArg` resolver, mirroring the #2948 behavior for `--adopt`: a unique prefix resolves to the full id; an ambiguous prefix is refused with the candidate list; an unknown id errors as today; full ids pass through unchanged. Plus focused regression coverage and repository validation.
- **Explicit boundaries:** do not change `resolveSupersededRun`'s terminality check or its refusal messages; do not change the provenance recording mechanism; no other `--supersedes` behavior changes; the fix is confined to `packages/cli/src/commands/workflow.ts` and its test file.
- **PR body review focus:** correctness of prefix resolution at both call sites; the detached path threads the resolved id everywhere the adopt lane does.

## Reviewed head SHA

`a0f9a379814c3f6b0135e76f3a761435075a35ec` (PR #3047 head, branch `fix/2990-supersedes-prefix`, base `dev`, state OPEN). This is the next round's cursor. The delta since the round-1 cursor `9fb0ac7ff89612568e41b3dcc82507eec3437d16` is exactly one commit touching `packages/cli/src/commands/workflow.test.ts` (+49/-0); `workflow.ts` is unchanged.

## Findings

### Critical

None.

### Important

None open. R1 (round 1, sources: [tests]) is now **fixed** — see the prior-findings table.

### Suggestions

None.

## Prior findings (continuation mode)

| ID | Severity | Sources | Status | Verdict |
|----|----------|---------|--------|---------|
| R1 | Important | [tests] | **fixed** at `a0f9a379` | Verified. The new test `refuses a non-terminal supersedes run in the detach pre-flight before forking` (`workflow.test.ts:10459-10506`) falsifies both regressions R1 described: **Regression A** (pre-flight `resolveSupersededRun` call dropped at `workflow.ts:2131`) — the flow proceeds past the pre-flight, trips the `getArchonHome` throw-away mock or forks, and the assertions `spawnSpy` not called, `mockCreateWorkflowRun` not called, and the specific rejection message each fail; **Regression B** (terminality checked on the raw prefix before resolution) — the test uses an argument-aware `mockImplementation` on `getWorkflowRun` that returns `null` for the raw prefix `0b1ee8da`, so `resolveSupersededRun` throws `Cannot supersede: no workflow run '0b1ee8da' exists.` (`workflow.ts:805`), which does not match the expected `Cannot supersede run '<full id>': it is still running.` and the test fails. Both regressions therefore no longer survive the suite. |

One verification note, recorded for accuracy: the round-1 report's suggested recipe (queue the running run via `mockResolvedValueOnce`) was subtly unsound — Bun's queued-value mocks are argument-blind, so the full-id value would be returned for a prefix lookup too, and regression B would pass. The shipped test correctly deviates from that recipe with an argument-aware mock (the implementation record documents this deviation, and inspection confirms the mock reasoning). This is a defect in the round-1 fix recipe, not in the shipped test; the finding's claim was correct and is now satisfied.

The new test's argument-aware `mockImplementation` does not leak: the describe block resets `getWorkflowRun` (and every other shared mock) to argument-blind defaults in both `beforeEach` and `afterEach` (`workflow.test.ts:10277, 10282`), and the test is the last in its describe block.

## Discoveries

Four accepted adjacent discoveries (unchanged from round 1; this round's delta is test-only and raises none), all outside the accepted contract; **adjacent discoveries do not affect readiness**. They are recorded in [discoveries.json](../../discoveries.json) and [discoveries.md](../../discoveries.md):

1. Server/orchestrator lane records `supersedes_run_id`/`adopt_run_id` raw with no prefix resolution or existence/terminality validation (pre-existing, symmetric between flags).
2. CLI docs omit prefix behavior for `--adopt`/`--supersedes` and the `--adopt` row contradicts shipped behavior; the `--supersedes` half is newly created by this PR and warrants a follow-up issue.
3. `CONTINUATION_METADATA_KEY`/`ContinuationMetadata` are hand-synced in `core/operations/workflow-adoption.ts` instead of imported from their owner.
4. `workflow-terminal-event.integration.spec.ts` is flaky under sequential full-suite load (pre-existing, flagged in the PR body).

## Review coverage

Continuation review: one reviewer verified the round-1 report and the correction delta; the specialist lenses did not rerun. The round-1 accepted concerns were: prefix resolution and id threading at both call sites (code), continuation call-site set and boundary contracts (seams), alternative-shape analysis (simplify), and coverage of the rewritten pre-flight block (tests, R1).

Verification performed this round, against the code rather than summaries:

- Confirmed the checkout is clean at head `a0f9a3798`, the live PR head matches, and the delta (`git diff 9fb0ac7ff..a0f9a379`) is exactly the 49-line test addition.
- Read the new test in full and traced both falsification paths against the production code: `resolveSupersededRun` (`workflow.ts:802-811`) throws the null-lookup message verbatim, and the argument-aware mock makes regression B observable; the `getArchonHome` trip-wire plus the spawn/provenance assertions make regression A observable.
- Read the describe block's `beforeEach`/`afterEach` reset and confirmed the argument-aware mock cannot leak across tests.
- Ran the documented commands: `bun test src/commands/workflow.test.ts -t 'supersedes run-id prefix'` (7 pass, 0 fail) and the full file `bun test src/commands/workflow.test.ts` (347 pass, 0 fail).
- Revalidated the three raw discovery records (`implement.json`, `review-code.json`, `review-seams.json`) against the round-1 consolidated files; all citations were spot-checked in round 1 and no new raw records appeared.

The delta adds no user-facing surface (test-only), so no gated-off lens required re-evaluation. No evidence was unobtainable.

PR #3039, opened 2026-08-31
refactor(workflows): pass run context to internal executors

Wirasm:
## Problem and outcome

Six private workflow-node executors unpacked the same run-level state and forwarded it positionally, including a hazardous run of path strings. That duplicated the `RunLayersContext` already owned by `runLayers` and made signature changes error-prone.

- **Outcome:** All six executors now receive `RunLayersContext`; only values that vary for an individual node remain explicit arguments.
- **Invariant:** Loop iteration and composed-instance contexts continue to list their deliberate overrides explicitly rather than inheriting a parent through a spread.
- **Scope boundary:** This is PR-A only. It changes one production file and does not include #2302, #2487, or #2698.

## Review guidance

- **Feedback requested:** Confirm that the context boundary preserves each executor's existing per-node inputs and that the loop-group iteration context retains its intentional resets.
- **Start here:** `packages/workflows/src/dag-executor.ts` — `runLayers` now dispatches the six private executors with the owning context directly.
- **Review order:** Review the six signature changes, then their `runLayers` dispatches, then the field-by-field `iterCtx` construction in `executeLoopGroupNode`.
- **Lower-attention areas:** The remaining diff is mechanical replacement of locals formerly unpacked from `ctx` with direct `ctx` access inside `runLayers`.
- **Known risk or uncertainty:** None.

## Solution

`executeNodeInternal`, `executeBashNode`, `executeScriptNode`, `executeLoopNode`, `executeLoopGroupNode`, and `executeApprovalNode` now take `RunLayersContext` as their first argument. Each destructures only the run-level values it owns, while resolved provider/model/options, retry and session state, environment values, loop input, iteration, and step-name prefixes stay at the call site.

`runLayers` no longer destructures and rebuilds the shared run state for those calls. The loop-group `iterCtx` remains an explicit literal sourced from the parent context, preserving its distinct workflow name, session-scope resets, usage accumulators, output map, and loop lineage. The composed `instanceCtx` is unchanged.

## Validation

- `cd packages/workflows && bun test src/dag-executor.test.ts` — 702 passing, 0 failing before and after the change; proves the unchanged executor behavior remains covered with no tests lost.
- `cd packages/workflows && bun x tsc --noEmit --pretty false` — passed.
- `bun run validate` — passed, including generated-artifact checks, type checks, lint, formatting, installer tests, package tests, and script tests.
- `git diff --stat -- '**/*.test.ts'` from the branch point — empty; no test file changed.
[...]

Issue #2302, opened 2026-07-29
refactor(workflows): executeDagWorkflow takes an options object derived from RunLayersContext


Wirasm:
## Problem

`executeDagWorkflow` takes **27 positional parameters**, including a run of **seven consecutive strings**:

```ts
workflowProvider: string,
workflowModel: string | undefined,
artifactsDir: string,
stateDir: string,
logDir: string,
baseBranch: string,
docsDir: string,
```

Any transposition inside that run is type-correct. It fails at runtime, or worse, silently writes a run's artifacts into the wrong directory.

This is not hypothetical. During #2299 the PR added a parameter while `dev` concurrently merged #2205, which added a new positional call site. Neither branch was wrong on its own; the **merge** was. It compiled fine and surfaced only as a runtime `undefined is not an object (evaluating 'config.assistants')` in CI's merge build, while the branch head was green.

### The contract is already being described wrongly in prose

`executor.test.ts` asserts on the mocked executor by **numeric argument index** — 44 reads of `mockExecuteDagWorkflow.mock.calls`, spanning indices 0, 1, 6, 7, 11–13, 16–20, 24, and 25, plus two `calls.slice(0, 2).map(call => call[6])` shapes and a `calls[0]` destructure.

Two of the comments explaining those indices are **already off by one**:

```ts
// docsDir is arg index 11 (0-indexed) of executeDagWorkflow
const docsDir = mockExecuteDagWorkflow.mock.calls[0]?.[12];
```

```ts
// The config passed to executeDagWorkflow (arg index 12) should have merged envVars
const configArg = mockExecuteDagWorkflow.mock.calls[0]?.[13] as WorkflowConfig | undefined;
```

And one test's title names an index it does not read: `'threads source through to executeDagWorkflow (arg index 16)'` asserts on `[17]`.

Nothing checks any of this, so it drifted. These are the closest thing the repository has to a direct test of the argument contract, and they are wrong in three places. Under named fields they become `calls[0]?.[0].docsDir` and the compiler checks them.

## Proposal

Bundle the positionals into an options object — **derived from `RunLayersContext`, not written beside it.**

`RunLayersContext` already declares every field this signature carries, with the doc comments attached, and `executeDagWorkflow` already builds one from positionals it received 250 lines earlier. It is already partitioned by its own section comments into `// --- run-level invariants ---` and `// --- per-subgraph mutable state ---`.

[...]

PR #3056, opened 2026-08-31
Refactor DAG executor to use named options

Wirasm:
## Problem and outcome

`executeDagWorkflow` accepted 27 positional arguments, making call sites and mock assertions fragile when run state changes. This PR is PR-B of the #2302 series: it moves the DAG boundary to one named options object while preserving execution behavior.

- **Outcome:** Callers pass named DAG execution inputs, and tests assert named fields instead of tuple indexes.
- **Invariant:** `workflowModel` remains a required `string | undefined` option; DAG-derived state stays distinct from caller-supplied inputs.
- **Scope boundary:** This does not change workflow behavior, provider/model resolution, or the PR-A run-context work already merged in #3039.

## Review guidance

- **Feedback requested:** Confirm the input/derived boundary is correct and that no DAG execution values were lost or made optional.
- **Start here:** `packages/workflows/src/dag-executor.ts:9639` — the split between caller inputs, DAG-derived values, and mutable per-subgraph state.
- **Review order:** Review commit `902572031` for the type and production caller, then `311d1710d` for the mechanical direct-test-call migration and mock assertions, then `b76bc6931` for the options contract test.
- **Lower-attention areas:** `dag-executor.test.ts` and `script-node-deps.test.ts` are mechanical conversions of 442 direct calls; an AST comparison verified each retained argument and every call includes `workflowModel`.

## Solution

`RunInputs` now owns the values supplied to the DAG boundary, while `RunDerived` holds the concrete run-layer values, including boundary normalizations. `RunLayersContext` combines those types with its per-subgraph mutable state. The exported `ExecuteDagWorkflowOptions` extends the input contract with the workflow and optional boundary inputs needed by callers.

`executeWorkflow` now calls `executeDagWorkflow` with named options. The executor locally destructures those options and explicitly builds its run context, normalizing boundary values before adding mutable subgraph state. Direct test calls and executor mocks use the same field names.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | DAG execution values were passed by position. | The same values are passed by descriptive option names. |
| **Failure behavior** | Positional argument ordering could be misread at call sites. | TypeScript rejects positional calls and rejects a missing required `workflowModel`. |

## Architecture

```mermaid
flowchart LR
  E[executeWorkflow] --> O[ExecuteDagWorkflowOptions]
  O --> I[RunInputs]
  O --> D[Boundary-derived values]
  I --> C[RunLayersContext]
  D --> C
  C --> R[runLayers]
```

### Changed seams

|---|---|---|
| `executeWorkflow` → `executeDagWorkflow` | One named options object replaces 27 positional arguments. | `packages/workflows/src/executor.ts:3045` |
| DAG boundary → `runLayers` | Caller inputs and derived context are separately typed before mutable subgraph state is added. | `packages/workflows/src/dag-executor.ts:9639` |
| Executor tests → DAG mock | Assertions select named options fields rather than numeric call indexes. | `packages/workflows/src/executor.test.ts` |

## Validation

- `cd packages/workflows && bun x tsc --noEmit` — passed.
- `cd packages/workflows && bun test src/dag-executor.test.ts` — 703 passed, 0 failed.
- `cd packages/workflows && bun test src/executor.test.ts` — 136 passed, 0 failed.
- `cd packages/workflows && bun test src/script-node-deps.test.ts` — 7 passed, 0 failed.
- `cd packages/workflows && bun test src/executor-preamble.test.ts` — 7 passed, 0 failed.
- `bun run validate` — passed, including generated artifacts, type checks, lint, formatting, installer tests, package tests, and script tests.
- Temporary positional-transposition proof: TypeScript rejected the restored 14-argument call with `TS2554: Expected 1 arguments, but got 14`; reverted before validation.
- Temporary required-field proof: TypeScript rejected a production options object without `workflowModel` with `TS2345` because the property is required; reverted before validation.
- `git diff --check origin/dev...HEAD` — passed.
- **Not verified:** Nothing material; validation covers the type contract and existing DAG behavior.

## Links

- Relates to #2302
- Depends on #3039

## Follow-up: test call sites now pass only what they vary

Operator-directed simplification pass on the reviewed change. The 1:1 codemod was faithful
but left every test call spelling out all fourteen named options, so the conversion read as
the fields a test does not care about rather than as the run it configures. Net effect on the
two test files was only −180 lines, and 439 of the retained lines were `workflowModel: undefined`.

**The builder.** `dagOptions(overrides)` is test scaffolding, private to each test file — not
an engine export and not on any production path. It returns `ExecuteDagWorkflowOptions`, takes
`deps`, `cwd`, `workflow`, and `workflowRun` as required (they carry each test's own fixtures),
and defaults the rest. Each of the two files owns its own builder over its own mocks; the two
carry different values for `conversationId`, `config`, and `platform`, and neither has to agree
with the other for either to stay correct.

**Where the defaults came from.** They are not invented. Every call site's property values were
extracted with the TypeScript parser and tallied, and each default is the expression that
tally showed the call sites already passing — `'claude'`, `undefined`, `'main'`, `'docs/'`,
`minimalConfig`, `createMockPlatform()`, `'conv-dag'` (`'conv-deps'` in the smaller file). The
three run directories are derived from the call's own `cwd`, which is how every site built them.

**How the conversion was made.** An AST codemod, run once from a scratch directory and not
committed. It drops a property only when that property's exact source text equals the default
the builder would produce for that same call site, and never when the property carries a
comment; everything else is emitted verbatim, including the nested `loop_group` and
`compose_fan_out` fixtures. A hand pass over 34,000 lines is the transposition risk this
series exists to remove.

**Equivalence evidence.** A second pass re-parsed the rewritten files, recomputed each site's
`defaults + overrides`, and compared it against the property→expression map captured from the
same site before the rewrite, matching on AST shape so prettier's reflow and trailing commas
cannot mask a difference. All 442 sites matched with zero mismatches. The comparison was seen
red first: changing one builder default from `'docs/'` to `'docs'` reported 435 mismatches.

| | Before this pass | After |
|---|---|---|
| `dag-executor.test.ts` | 34,020 lines | 31,312 lines (**−2,708**) |
| `script-node-deps.test.ts` | 513 lines | 487 lines (**−26**) |
| `workflowModel: undefined` lines | 439 | 3 (two builder defaults, one contract fixture) |

Suite counts and assertion counts are identical on both sides, which is the behavioural half of
the equivalence claim:

- `dag-executor.test.ts` — 703 pass / 0 fail / 2078 `expect()` before and after.
- `script-node-deps.test.ts` — 7 pass / 0 fail / 36 `expect()` before and after.
- `executor.test.ts` — 136 pass / 0 fail / 319 `expect()`; untouched by this pass. Its named-field
  mock assertions read the options object the production caller builds, which the builder has no
  business in.

Unchanged by this pass: the production signature, `ExecuteDagWorkflowOptions`, and the decision
that `workflowModel` stays a required `string | undefined`. The builder supplying it as a default
does not weaken that guard, whose audience is production callers rather than tests. The
options-contract test still spells the full shape out and still holds the positional call under
`@ts-expect-error`. `cd packages/workflows && bun x tsc --noEmit` and `bun run validate` are both
green on the pushed head; `validate` ran the three suites at 703 / 136 / 7.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined workflow execution interfaces with clearly named options.
  * Organized workflow run inputs and derived execution details into consistent structures.
  * Preserved existing workflow execution behavior and results.

* **Tests**
  * Updated workflow execution tests for the new options-based interface.
  * Maintained coverage across configuration, resuming runs, paths, environments, telemetry, and workflow execution.

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -84,6 +84,7 @@ import {
   containerCommandName,
   buildSubprocessDockerArgs,
   childOutcomeFromRun,
+  type ExecuteDagWorkflowOptions,
   type RunChildWorkflowFn,
 } from './dag-executor';
 import { writeNodeArtifact, readNodeArtifacts } from './artifacts-index';
@@ -376,6 +377,69 @@ const minimalConfig: WorkflowConfig = {
   commands: {},
   defaults: { loadDefaultCommands: false, loadDefaultWorkflows: false },
 };
+
+describe('executeDagWorkflow options type contract', () => {
+  it('requires one named object including workflowModel', () => {
+    const validOptions: ExecuteDagWorkflowOptions = {
+      deps: createMockDeps(),
+      platform: createMockPlatform(),
+      conversationId: 'type-contract',
+      cwd: '/tmp/type-contract',
+      workflow: { name: 'type-contract', nodes: [] },
+      workflowRun: makeWorkflowRun('type-contract'),
+      workflowProvider: 'claude',
+      workflowModel: undefined,
+      artifactsDir: '/tmp/type-contract/artifacts',
+      stateDir: '/tmp/type-contract/state',
+      logDir: '/tmp/type-contract/logs',
+      baseBranch: 'main',
+      docsDir: 'docs',
+      config: minimalConfig,
+    };
+
+    const { workflowModel, ...withoutWorkflowModel } = validOptions;
+    expect(workflowModel).toBeUndefined();
+
+    // @ts-expect-error workflowModel is a required resolved-policy input, even when undefined.
+    const missingWorkflowModel: ExecuteDagWorkflowOptions = withoutWorkflowModel;
+    expect(missingWorkflowModel).not.toHaveProperty('workflowModel');
+
+    const legacyArguments = [
+      validOptions.deps,
+      validOptions.platform,
+      validOptions.conversationId,
+      validOptions.cwd,
+      validOptions.workflow,
+      validOptions.workflowRun,
+      validOptions.workflowProvider,
+      validOptions.workflowModel,
+      validOptions.artifactsDir,
+      validOptions.stateDir,
+      validOptions.logDir,
+      validOptions.baseBranch,
+      validOptions.docsDir,
+      validOptions.config,
+      validOptions.configuredCommandFolder,
+      validOptions.issueContext,
+      validOptions.priorCompletedNodes,
+      validOptions.source,
+      validOptions.aiProfile,
+      validOptions.workflowPreset,
+      validOptions.scopeArtifactsDir,
+      validOptions.execContext,
+      validOptions.containerCtx,
+      validOptions.runChildWorkflow,
+      validOptions.priorUsage,
+      validOptions.priorNodeSessions,
+      validOptions.workflowSourceRoots,
+    ] as const;
+    const legacyCall = (): Promise<string | undefined> => {
+      // @ts-expect-error executeDagWorkflow no longer accepts the positional contract.
+      return executeDagWorkflow(...legacyArguments);
+    };
+    expect(legacyCall).toBeFunction();
+  });
+});
 
 // --- Helpers ---
 
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -378,6 +378,39 @@ const minimalConfig: WorkflowConfig = {
   defaults: { loadDefaultCommands: false, loadDefaultWorkflows: false },
 };
 
+/**
+ * `deps`, `cwd`, `workflow`, and `workflowRun` carry each test's own fixtures, so every call
+ * supplies them; the run directories derive from `cwd` the way every call site built them.
+ */
+type DagOptionsOverrides = Partial<ExecuteDagWorkflowOptions> &
+  Pick<ExecuteDagWorkflowOptions, 'deps' | 'cwd' | 'workflow' | 'workflowRun'>;
+
+/**
+ * Options for a direct `executeDagWorkflow` call, built from only what a test varies. The
+ * defaults are the exact values these tests used to spell out at every call site, so a test
+ * reads as the run it configures rather than as the fields it does not care about.
+ *
+ * Supplying `workflowModel` here is a convenience for tests only. It stays a required
+ * `string | undefined` on `ExecuteDagWorkflowOptions` so a production caller cannot omit a
+ * resolved model decision (#2302); this builder is not that guard's audience.
+ */
+function dagOptions(overrides: DagOptionsOverrides): ExecuteDagWorkflowOptions {
+  const { cwd } = overrides;
+  return {
+    platform: createMockPlatform(),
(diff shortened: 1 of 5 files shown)


Wirasm:
# Review report

## Verdict

Ready — `none`. The correction adds the required negative compile-time contract proof, and the focused type check and DAG executor suite pass at the reviewed head.

## Accepted contract

PR-B of #2302 must make `executeDagWorkflow` accept one named options object derived from `RunLayersContext`, keep caller inputs distinct from DAG-boundary-derived state, and update the production caller, direct test calls, and executor mock assertions. The resulting type contract must reject positional invocations, omitted required inputs, and transposed same-typed inputs. `workflowModel` remains a required `string | undefined` property; workflow behavior, provider/model resolution, and the already-merged PR-A work are out of scope. The required two-commit shape is hand-written boundary/production change followed by mechanical test migration.

## Reviewed head SHA

`b76bc6931a51d562f70f0a7c59f3690f18a686fb`

## Findings

### Critical

None.

### Important

None open.

### Rejected findings

None.

### Suggestions

None.

## Prior findings

| ID | Severity | Sources | First raised | Status | Verdict |
| --- | --- | --- | --- | --- | --- |
| R1 | Important | [tests] | `311d1710dee9f38a73a3b23f705b66ea6c485cf8` | fixed | The new `executeDagWorkflow options type contract` test at `packages/workflows/src/dag-executor.test.ts:381-441` constructs a valid object with `workflowModel: undefined`, uses `@ts-expect-error` to require that field at lines 403-405, and rejects the full former positional contract at lines 407-440. `executeDagWorkflow` still accepts only `options: ExecuteDagWorkflowOptions` at `packages/workflows/src/dag-executor.ts:11571-11573`, while the production caller passes `workflowModel` by name at `packages/workflows/src/executor.ts:3045-3053`. `cd packages/workflows && bun run type-check` passed, proving both expected-error directives remain active. |

## Discoveries

0 accepted discoveries. See [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/b01d401171c14467f961e4d74442baa6/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/b01d401171c14467f961e4d74442baa6/discoveries.md). Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and the correction delta. The original selected concerns were the named DAG-executor boundary, ownership separation between caller inputs and DAG-boundary-derived state, call-site migration, and regression protection; `code`, `seams`, and `simplify` had found no defect, while `tests` raised R1. The original `errors` and `docs` lenses were disabled because the reviewed refactor added neither an error/recovery path nor user-facing configuration, API, CLI, or language surface. The correction delta adds only a test and no user-facing surface, so no gated-off lens was re-evaluated. No material evidence was unavailable.

Validation: `git diff --check 311d1710dee9f38a73a3b23f705b66ea6c485cf8..b76bc6931a51d562f70f0a7c59f3690f18a686fb` was clean; `cd packages/workflows && bun run type-check` passed; `cd packages/workflows && bun test src/dag-executor.test.ts` passed (703 tests).

PR #3057, opened 2026-08-31
fix(cli): remove heavy startup from detached config guards

Wirasm:
## Problem and outcome

The detached config handoff test synchronously launched the real CLI, but the CLI loaded every provider, command, core, workflow, and Git module before it could reject the known continuation/config conflict. Windows cold-start contention pushed that unnecessary work past Bun's implicit test deadline and blocked unattended runs.

- **Outcome:** Help, argument failures, and the detached continuation/config guard no longer load the heavy command graphs before dispatch.
- **Invariant:** The real entry still strips repository env, captures the detached install context, loads Archon env, restores the captured context, decrypts the parent's payload, and reaches the continuation/config conflict.
- **Scope boundary:** The real subprocess assertion remains unchanged; no timeout, retry, tolerance, platform skip, or copied boot probe was added.
- **Root cause:** Eager module evaluation charged the test for provider registration and the full CLI/workflow/core/Git graph before a small pre-dispatch outcome.

## Review guidance

- **Feedback requested:** Check the pre-dispatch import boundary and the split between structural handoff validation and provider-dependent semantic normalization.
- **Start here:** `packages/cli/src/cli.ts:440` — the restored install context now decrypts and rejects the detached continuation before Git or workflow dispatch.
- **Review order:** `packages/cli/src/cli.ts`, `packages/core/src/config/run-config-handoff.ts`, `packages/cli/src/commands/workflow.ts`, then `packages/cli/src/cli.test.ts`.
- **Lower-attention areas:** `durable-wait.ts` and `thinking-config.ts` only move existing schema leaves so the decoder does not pull in the DAG/provider graph; schema and repository validation cover the move.
- **Known risk or uncertainty:** Windows CI is now collected and green (see Validation). The one Windows failure this PR saw was an unrelated inherited flake, diagnosed in #2859 rather than patched here.

## Solution

The CLI now imports providers, Git, commands, and database closure only for the selected route. The detached payload uses one narrow decoder before dispatch: metadata and decrypted layer structure are validated without the provider registry, then a surviving workflow run performs the existing provider semantic normalization after provider registration. The command's old duplicate parser was removed.

Database cleanup is tied to loaders that can load a database-backed route, so light returns do not import core merely to close an unopened connection.

The regression test builds the CLI with Bun code splitting and traverses the static metafile graph. It pins an empty heavy-graph set for the entry chunk and an exact eight-file internal leaf set for the decoder chunk. This encodes graph reachability, not a machine-specific duration.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Every CLI invocation eagerly loaded provider, command, core, workflow, and Git graphs. | Light startup and guards load only boot utilities; selected routes load their own graphs. |
| **Failure behavior** | Detached continuation/config conflict paid full CLI startup and could cross the Windows test deadline. | The payload is decrypted and the same conflict is returned before Git, providers, workflow commands, or database lookup. |

## Validation

- `bun run validate` — exit 0 on the final head (170 suites, every one `0 fail`). Across the earlier heads it passed 3 of 5 runs; covers generated artifacts, API types, type checking, lint, formatting, install tests, and all package tests in the real parallel repository command. The 2 failures were both `workflow-terminal-event.integration.spec.ts` throwing `SQLITE_BUSY`, which this PR does not touch: that spec reads through a raw `bun:sqlite` handle with no `busy_timeout` while its own detached child holds the lock. It passes 5 of 5 in isolation, both CI legs pass it on this head, and this PR's original implementation report recorded the same failure on the parent commit. Mechanism and fix boundary recorded on #2924.
- `bun run test` in `packages/cli` — passed the complete CLI package script.
- 20 consecutive executions of the unchanged real detached-handoff subprocess test — all passed; target body durations were 539.24–724.16 ms.
- Import-graph regression — entry static graph contains no CLI command, core, Git, provider, or workflow inputs; decoder static graph is exactly the audited schema/crypto/path leaves.
- Mutation check — temporarily removing `restoreDetachedInstallContext` made the exact test fail with `Workflow run config could not be decrypted`; restoring it returned the expected `--resume and --config are mutually exclusive` result.
- Guard seen red — dropping `ARCHON_HOME` from the envelope spawn helper makes the seeding hook throw; suppressing it for the sentinel case alone makes `<sentinel>/.archon` appear and the new assertion fail with `Expected: false / Received: true`.
- CI on head `75fa6becd` — all nine checks green, including `test (ubuntu-latest)` and `test (windows-latest)`, with no rerun. On Windows the previously failing `outside a git repository` case runs in **375 ms**, against 1297 ms on `dev` and the 7281 ms timeout that opened this thread; the new continuation regression test passes there in 4235 ms.

### Review correction round

An independent review found a real regression in the deferred-import work, plus two simplifications. All four findings are addressed on this branch.

**`d035d55ed` — providers were lost on every continuation subcommand (Critical).** Registration was keyed to the literal subcommand `run`, but `resume`, `approve`, `reject` and `respond` all reach `workflowRunCommand` through the same entry point — and they are the callers that need the registry most. A continuation resolves its workflow from the run's captured source, and passing that capture's `source_config` into discovery is exactly the branch that skips `loadConfig()`, the call that self-registers providers for every other route. Continuations therefore arrived with an empty registry, the loader rejected any workflow declaring a `provider:`, and the run was reported as missing from its own capture.

Reproduced against the pre-fix head with a real CLI subprocess, which printed both halves of the symptom:

```text
"error": "Unknown provider 'claude'. Registered: "
Error: Failed to resume workflow 'provider-scoped': Cannot continue run of 'provider-scoped':
  its captured source ... no longer contains that workflow.
```

Dispatch now asks `dispatch-guards` for the answer through an exported `isContinueSubcommand`, rather than keeping a second copy of the subcommand list — which is how the two drifted apart. The same commit folds in the simplification finding: `loadWorkflowCommands` is deleted, since it duplicated `loadRoute` while memoizing a call that happens once per process, and its unconditional database flag charged `workflow search` and `workflow test` a database close they never opened.

The regression test spawns the real CLI. The existing suite calls the command functions in-process, where importing them registers providers as a side effect — which is precisely why this shipped invisible. It runs a workflow-level `provider:` over bash-only nodes, so it reproduces the registry dependency without ever starting a provider. **Seen red first:** against the pre-fix source it fails on `not.toContain("Unknown provider 'claude'")`; with the fix it resumes and reaches the node's own exit status. It is registered in the `cli` package's `test` and `test:integration` scripts, which are hand-listed — without that it would never have run in CI.

**`75fa6becd` — named the pre-dispatch `--resume` check a fast path (Suggestion).** It repeats a guard `workflow.ts` still owns for every caller, and nothing said which of the two is the invariant.

### Windows CI outcome

The measured effect on `test (windows-latest)`, comparing this branch against `dev`'s own Windows job. Every guard that no longer loads the heavy graph got 3-4x cheaper, and the target test dropped by two thirds:

| `test (windows-latest)` | `dev` | this PR |
|---|---|---|
| keeps a detached config handoff outside repo env overrides (the issue's test) | 2172 ms | **609 ms** |
| resolves a relative config path from the requested subdirectory cwd | 2266 ms | 1906 ms |
| pre-dispatch: `--cwd` does not exist | 968 ms | 219 ms |
| pre-dispatch: unknown command | 1047 ms | 266 ms |
| pre-dispatch: unknown flag | 953 ms | 187 ms |
| pre-dispatch: chat with no message | 797 ms | 188 ms |
| pre-dispatch: invalid `setup --scope` | 813 ms | 172 ms |
| pre-dispatch: `setup --scope project` outside a git repo | 828 ms | 250 ms |

### The second commit: absorbing the #2982 outside-git item

The first Windows run failed one test, `pre-dispatch gates --json error envelope > emits { ok: false } on stdout when outside a git repository`, at 7281 ms against Bun's 5000 ms budget. That failure is not caused by the `#2859` change: the test body is byte-identical to `dev`, the same test failed at 5485 ms on `test/3052-core-test-typing` (which touches neither `cli.ts` nor `cli.test.ts`), and on macOS the exact command is *faster* on this branch than on `dev` (600 ms vs 780 ms).

Its real cause is the test-isolation defect recorded on #2982. The `--json` envelope tests spawn the real CLI with the ambient environment, so the no-git folder-project gate opens the operator's `~/.archon` registry. That shares one SQLite file with every other process using it, and the adapter sets `PRAGMA busy_timeout = 5000` — the same number as Bun's implicit per-test budget — so a contended open can spend the whole budget waiting. Reproduced on an isolated home:

| Condition | Wall time |
|---|---|
| No lock held (baseline) | 0.51 s |
| Write lock held 3,000 ms | **3.41 s** |
| Write lock held 8,000 ms | **5.54 s** (the `busy_timeout` ceiling) |

Commit `e030f2afd` applies #2982's recorded fix boundary: the spawn helper owns a scratch `ARCHON_HOME`, seeded once in a hook so schema creation stays out of every timed body, and a new case asserts the child's default home is never created. No timeout was widened. Seeding runs *through* the helper, so it doubles as the check that the home reaches the child.

Both halves of that guard were seen red before being trusted: dropping `ARCHON_HOME` from the helper makes the seed throw, and suppressing it for the sentinel case makes `<sentinel>/.archon` appear and the assertion fail.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Existing command exports and provider-normalized run-config behavior remain available; no persisted or wire shape changed. | Workflow/core tests and `bun run validate`. |
| Security / permissions / data | Detached decryption still uses the captured parent install identity after hostile repository env loading. | Real subprocess test plus restore-omission mutation. |
| Observability | Static graph drift now fails deterministically without relying on startup timing. | `packages/cli/src/cli.test.ts:59`. |

## Links

- Refs #2859 — carries the test-side fix for the detached-config startup cost. #2859 stays open as the tracker for the user-facing half its thread records: outside a git repository, a real `archon` invocation can still wait up to five seconds on the folder-project gate's registry open before printing "Not in a git repository". No separate issue exists for that, so closing #2859 here would lose it.
- Refs #2982 — carries the outside-git envelope-test isolation item only; the rest of that issue stays open.

## Summary by CodeRabbit

- **New Features**
  - Added securely validated detached workflow run configurations.
  - Added Claude thinking modes: adaptive, enabled, and disabled.
  - Added durable waits of up to 1,000 years.

- **Bug Fixes**
  - Improved workflow failure exit codes, including nested and detached failures.
  - Standardized errors when `--resume` is used with `--config`.
  - Improved workflow continuation by correctly restoring providers and workflows.

- **Performance**
  - Reduced CLI startup work by loading commands and providers only when needed.


Wirasm:
Picking this up: the `archon-stabilize` run that opened this PR (`818f1cffb93a685103deda81d4818ae3`) was cancelled mid-flight, so nobody was feeding its red CI back into a fix loop. I am taking over CI correction on this branch and will push to `fix/2859-detached-config-startup` in place.

Current state on head `b27d299c6`: `test (windows-latest)` fails, `Test Suite` is the roll-up of that failure, everything else is green. Diagnosis and fix to follow in a further comment.

Wirasm:
## Diagnosis: the red was inherited, not from this change. CI is green on the unchanged head.

No commit was needed. Head is still `b27d299c6` and every check now passes, including `test (windows-latest)`.

### What was red

One test, `pre-dispatch gates --json error envelope > emits { ok: false } on stdout when outside a git repository`, at 7281 ms against Bun's implicit 5000 ms budget. The `Test Suite` failure beside it was only the roll-up gate reading `TEST: failure`, not a second problem.

The test this PR exists to fix passed in the same job, at 765 ms — down from 2172 ms on `dev`.

### Why it is not this change

- The failing test's body is byte-identical to `dev`. This PR only appends a new describe block to `cli.test.ts`.
- The same test failed at 5485 ms on `test/3052-core-test-typing` ([run 33403158767](https://github.com/coleam00/Archon/actions/runs/33403158767)), a branch touching neither `cli.ts` nor `cli.test.ts` and not containing this change.
- On macOS the exact failing command is *faster* on this head than on `dev`: median 600 ms vs 780 ms over 7 samples. The module this path now loads lazily, `@archon/core/db/codebases`, costs 100 ms cold, against 330 ms for the `@archon/core` barrel `dev` loaded eagerly. This head does strictly less work on that path.
- Rerunning the unchanged head went green, with the test at 1281 ms — inside its Windows range of 797-1516 ms across eight recent runs.

### What was actually wrong

That test runs `archon workflow list --json --cwd <non-git dir>`, which hits the pre-dispatch folder-project gate at `cli.ts:538`. The gate queries the shared `~/.archon/archon.db`, and `SqliteAdapter` sets `PRAGMA busy_timeout = 5000` — the same number as Bun's default per-test budget.

Reproduced directly, on an isolated `ARCHON_HOME`, by holding a write lock while running that command:

| Condition | Wall time |
|---|---|
| No lock (baseline) | 0.51 s |
| Write lock held 3,000 ms | 3.41 s |
| Write lock held 8,000 ms | 5.54 s |

The gate waits out the whole lock, capped by `busy_timeout`. That cap alone reaches the deadline the test is measured against, so under contention the test cannot reliably pass. The envelope and exit status were correct in every locked run — it is only ever late.

Full write-up on #2859 ([mechanism](https://github.com/coleam00/Archon/issues/2859#issuecomment-5480150465), [reproduction](https://github.com/coleam00/Archon/issues/2859#issuecomment-5480180124)). Left unfixed here deliberately: it reproduces on branches without this change, so the fix belongs on `dev` where every branch inherits it, and folding it in would widen a stabilize-class PR into the database adapter.

### Validation run for this handover

- `bun run validate` at the repo root, on this exact head in a clean worktree — exit 0, 169 suites reporting `0 fail`, zero `(fail)` markers. That includes the CLI batch the failing test lives in (118 tests, 27.74 s).
- `test (windows-latest)` and `test (ubuntu-latest)` green on `b27d299c6`.

The PR body's Validation section has been updated to match, replacing the "20 consecutive Windows runs not verified" note the cancelled run left behind.

Wirasm:
## Absorbed the #2982 outside-git item as `e030f2afd`

Per the operator's decision to lift the earlier keep-out, this PR now also carries the fix for the test that was blocking its Windows leg. It is a separate commit from the #2859 work so the two concerns stay reviewable.

### What was actually wrong

The `--json` envelope tests launch the real CLI with the ambient environment. On the no-git path the folder-project gate opens the Archon registry to decide whether the directory is a registered folder project — so those spawns were reading and initialising the operator's `~/.archon/archon.db`, sharing one SQLite file with every other process using it.

`SqliteAdapter` sets `PRAGMA busy_timeout = 5000`, the same number as Bun's implicit per-test budget. A contended open can therefore spend the entire budget waiting. Reproduced against an isolated `ARCHON_HOME`, holding a write lock while running the exact failing command:

| Condition | Wall time |
|---|---|
| No lock held (baseline) | 0.51 s |
| Write lock held 3,000 ms | **3.41 s** |
| Write lock held 8,000 ms | **5.54 s** — the `busy_timeout` ceiling |

The gate waits out the whole lock, capped by `busy_timeout`, and that cap alone reaches the deadline the test is measured against. The envelope and exit status were correct in every locked run; the test was only ever late.

### The fix

#2982's recorded fix boundary, implemented as written:

- The spawn helper owns a scratch `ARCHON_HOME`, so these tests can no longer read or mutate real state and have nothing left to contend with.
- It is seeded once in a hook, keeping schema creation out of every timed body.
- Seeding runs *through the helper itself* rather than a separate spawn, so the seed doubles as the check that the home actually reaches the child — if that wiring is ever dropped, the seed throws instead of the whole suite silently falling back to the operator's registry.
- A new case asserts the child's default home is never created.
- No timeout was widened.

### Both guards were seen red first

- Removing `ARCHON_HOME` from the helper: the seed throws with `--json envelope tests could not seed a scratch registry in …`, 1 fail.
- Suppressing `ARCHON_HOME` for the sentinel case only: `<sentinel>/.archon` appears and the assertion fails with `Expected: false / Received: true`.

### Validation

- `bun run validate` at the repo root — exit 0, 169 suites reporting `0 fail`, zero `(fail)` markers. The CLI batch is now 119 tests, up one for the new case.
- `bun run test` in `packages/cli` — exit 0, 14 suites at `0 fail`.
- Windows CI on this head is the remaining evidence; it should now pass on its own merits rather than by rerun.

One thing worth flagging for review: this adds a single CLI spawn to `cli.test.ts` in the seeding hook. It is contention-free by construction (private home), and schema creation measured as free on macOS — 510 ms with a fresh home vs 520 ms with a seeded one — so the seed's cost is the process start, not the DDL.

The rest of #2982 (its 62-test audit, delete list, and type-check work) is untouched and that issue stays open.

Wirasm:
## Windows is green on its own merits on `e030f2afd`

All nine checks pass, and the leg that was red is now green without a rerun. The Windows numbers for the test that was blocking this PR:

| `pre-dispatch gates … outside a git repository` on `test (windows-latest)` | |
|---|---|
| `dev` (shared registry, passing run) | 1297 ms |
| this PR before the fix (shared registry, failing run) | 7281 ms — timed out |
| this PR with `e030f2afd` | **297 ms** |

It went from the most expensive spawn in that describe block to one of the cheapest — its siblings in the same run are 172-234 ms. Removing the shared-registry open took roughly a second off the healthy path *and* removed the 5-second stall ceiling that made the unhealthy path possible. The new guard case runs in 375 ms, and the #2859 target test is at 641 ms.

## Correction to my earlier validation note

The `bun run validate` line in my previous comment was accurate for the run it described, but it is not the whole picture and I would rather state the real numbers.

On the committed head, `bun run validate` passed 3 of 5 runs on my machine. The two failures were not in anything this PR touches — both were `workflow-terminal-event.integration.spec.ts` > `persists one matching event before successful and failed owners exit` throwing `SQLITE_BUSY: database is locked`.

What I established about it:

- It is an independent defect. That spec reads through a raw `new Database(path, { readonly: true })`, which gets SQLite's default busy timeout of zero, microseconds after forking a detached child against the same database. Its own writer can fail it whenever it holds the lock at that instant.
- It is not new here. This PR's original implementation report recorded hitting the same `SQLITE_BUSY` in the same spec on the parent commit `b27d299c6`.
- It passes 5 of 5 in isolation, and both CI legs pass it on this head (Windows 8203 ms, Ubuntu 3124 ms).
- The parent commit passed validate 2 of 2 in my runs, so I cannot rule out that this PR's one extra spawn perturbs the schedule enough to change the odds. What I can say is that the failure needs that spec's own missing `busy_timeout` to be possible at all.

Recorded on #2924 with the mechanism and a fix boundary. I did not fix it here — different file, different defect, and outside the item the operator scoped in.

## Scope

Two commits, deliberately separate:

- `b27d299c6` — the #2859 work: deferred command and provider imports at CLI dispatch.
- `e030f2afd` — the #2982 outside-git envelope-test isolation item only. The rest of #2982 is untouched and that issue stays open.

Wirasm:
## Correction round: all four findings addressed

Head is now `75fa6becd`. Two commits, separated by concern.

### R1 (Critical) — fixed in `d035d55ed`, with the guard seen red first

Confirmed the finding independently before touching anything, by static trace and by reproduction. All four continuation subcommands land on `workflowRunCommand` (`workflow.ts:4282`, `:4578`, `:4716`, `:4842`), and `discoverWorkflowsWithConfig` skips `loadConfig()` — the provider self-registration fallback — whenever a `source_config` is present, which is every resumed run.

Reproduced against the pre-fix head with a real subprocess. A workflow-level `provider:` over bash-only nodes is enough: the loader validates the provider id at load time, so it reproduces the registry dependency without ever starting a provider.

```text
"errors":[{"filename":"archon-architect.yaml","error":"Unknown provider 'claude'. Registered: "}, … 11 total]
Error: Failed to resume workflow 'provider-scoped': Cannot continue run of 'provider-scoped':
  its captured source … no longer contains that workflow.
```

After the fix, the same resume reports `errorCount: 0`, discovers 33 workflows, re-executes, and fails on the node's own exit status.

**Fix shape.** Dispatch now asks `dispatch-guards` through an exported `isContinueSubcommand(subcommand)` rather than carrying a second copy of the subcommand list — a hand-list at the dispatch site is how the two drifted apart, so re-deriving from the owner is the part that keeps it fixed. That also removed the awkward cast at the existing `rejectModelOnContinue` call site.

**The regression test is a real subprocess** (`workflow-continuation-provider.integration.spec.ts`), because the existing suite calls the command functions in-process, where importing them registers providers as a side effect — exactly how this shipped invisible.

Seen red, on the final version of the test, with only the source reverted:

```text
expect(resumed.output).not.toContain("Unknown provider 'claude'");
error: expect(received).not.toContain(expected)
(fail) resumes a provider-scoped workflow instead of failing its source lookup
 0 pass  1 fail
```

One thing worth flagging: the `cli` package's `test` script is a hand-maintained file list, so a new spec file is invisible to it by default. I registered the spec in both `test` and `test:integration`; the suite count went 14 → 15. Without that the test would have passed review and never run in CI.

### R3 (Suggestion) — applied, folded into the same commit

`loadWorkflowCommands` and `workflowCommandsPromise` are deleted; all three call sites now use `loadRoute` with accurate flags. I verified the claim rather than taking it on faith: neither `workflowSearchCommand` nor `workflowTestCommand` touches the database, and both are dispatched before the git gate, so dropping the unconditional `database: true` costs them nothing and saves them a `closeDb()` import.

Verifying R3 is also what explained R1's blast radius: because `loadConfig()` self-registers, every *non*-continuation route is fine without explicit registration — which is why only continuation broke, and why `workflow test` on a `provider:`-scoped workflow behaves identically on both branches (checked directly).

### R4 (Suggestion) — applied in `75fa6becd`

Comment added naming it a fail-fast for the decode step, not a second invariant, and stating that a legitimately spawned detached child cannot reach it.

### R2 (Important) — applied

`Closes #2859` → `Refs #2859`, with the reason stated in the body: #2859 stays open as the tracker for the user-facing stall its thread records. No new issue filed, per the review.

### Validation

- `cd packages/cli && bun run test` — exit 0, 15 suites, every one `0 fail`.
- `bunx tsc --noEmit -p .` in `packages/cli` — exit 0. The new file is a `.spec.ts` deliberately, since `.test.ts` escapes type-check here; it caught a real `null` vs `undefined` error in my first draft.
- `bun run lint` — exit 0.
- `bun run validate` and CI on this head — reported in a follow-up once they finish.

Wirasm:
## Correction round green: all nine checks pass on `75fa6becd`

| Evidence | Result |
|---|---|
| `bun run validate` (repo root, final head) | exit 0 — 170 suites, every one `0 fail`, zero `(fail)` markers |
| `cd packages/cli && bun run test` | exit 0 — 15 suites (was 14; the new spec is registered) |
| `bunx tsc --noEmit -p .` in `packages/cli` | exit 0 |
| `bun run lint` | exit 0 |
| CI on `75fa6becd` | 9/9 green, no rerun |

The new regression test ran on both CI legs rather than only locally — `workflow continuation provider registration > resumes a provider-scoped workflow instead of failing its source lookup` passed on `test (windows-latest)` in 4235 ms, in a job with zero failures. That was the point of registering it in the package's hand-maintained test script.

Also unchanged by the correction round: the #2982 outside-git case is at **375 ms** on Windows, and the #2859 target test at **703 ms**.

One scope note for the delta review: the regression test drives `resume`. `approve`, `reject` and `respond` are fixed by the same line and share the identical call path, but they are not separately exercised — covering them would need a paused-gate fixture, and the predicate makes all four the same answer by construction. Flagging it rather than implying coverage I did not write.

PR #3063, opened 2026-08-31
test(cli): isolate outside-git registry lookup

Wirasm:
## Problem and outcome

The outside-git JSON-envelope test launched the full CLI without a test-owned Archon home. Its timed body could initialize the operator's SQLite registry, mixing schema startup into the assertion and potentially mutating developer state.

- **Outcome:** The test now uses a tracked scratch non-git directory, fake OS home, and prebuilt empty Archon registry.
- **Invariant:** It still exercises the real CLI subprocess, registered-folder lookup, exit status, JSON parsing, and `{ ok: false }` envelope.
- **Scope boundary:** Production CLI and folder-project behavior are unchanged; this does not attempt the broader test audit in #2982.
- **Root cause:** The child inherited the default `ARCHON_HOME`, so the no-git lookup could create the full SQLite registry inside the timed test body.

## Review guidance

- **Feedback requested:** Test isolation and whether the fixture still proves the real outside-git behavior.
- **Start here:** `packages/cli/src/cli.test.ts` — the setup creates and closes the scratch registry before the timed invocation.
- **Review order:** Review the single-file diff, then the reproduction and validation evidence below.
- **Lower-attention areas:** None.
- **Known risk or uncertainty:** Windows CI evidence is pending on this PR.

## Solution

One suite-owned scratch root now contains the non-git cwd, fake default home, and separate `ARCHON_HOME`. Setup creates and closes an empty SQLite registry before the test runs. The child receives explicit SQLite/non-Docker environment values, and the assertion proves the fake default home remains untouched.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | The test could initialize the default/operator registry. | All registry state is created under the test-owned scratch root. |
| **Failure behavior** | SQLite schema creation and host state could consume the test timeout. | Schema creation occurs in setup; the timed body measures the real outside-git lookup and envelope. |

## Validation

- Focused target — passed, including the default-home untouched assertion.
- 25 local reruns under Bun's unchanged 5-second default — 25 passed; target bodies were 213–241 ms.
- Mutation proof — disabling the JSON failure path made the target fail at JSON parsing; restoring it returned green.
- Registered-folder integration — 3 passed.
- `bun --filter @archon/cli test` — passed.
- `bun run validate` — passed, including generated checks, type checks, lint, formatting, installer tests, package tests, and script tests.
- **Not verified:** Windows CI; this PR supplies the branch and checks needed for that evidence.

## Delivery considerations

|---|---|---|
| Security / permissions / data | The test no longer reads or writes the operator registry. | Scratch `HOME`, `USERPROFILE`, and `ARCHON_HOME`; post-run absence assertion. |
| Observability | The test body no longer includes hidden schema initialization. | Prebuilt registry plus repeated focused timings. |

## Links

- Related #2982

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI error handling when commands are run outside a Git repository.
  * Prevented unintended database files from being created in the default home directory during error scenarios.
* **Tests**
  * Added isolated test environments to improve reliability and prevent filesystem or environment state from affecting results.

### DIFF
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -4,14 +4,15 @@
  * Note: These tests focus on argument parsing logic.
  * Full integration tests would require mocking the database and commands.
  */
-import { describe, it, expect, beforeAll } from 'bun:test';
+import { describe, it, expect, beforeAll, afterAll } from 'bun:test';
 import { Database } from 'bun:sqlite';
 import { parseArgs } from 'util';
 import { cliArgOptions } from './args';
 import * as git from '@archon/git';
+import { SqliteAdapter } from '@archon/core/db/adapters/sqlite';
 import { removeTempTree } from '@archon/paths/test-utils';
 import { spawnSync } from 'node:child_process';
-import { mkdirSync, mkdtempSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdirSync, mkdtempSync, writeFileSync } from 'node:fs';
 import { homedir, tmpdir } from 'node:os';
 import { join, relative } from 'node:path';
 import { pathToFileURL } from 'node:url';
@@ -1046,6 +1047,27 @@ describe('main catch --json error envelope', () => {
 });
 
 describe('pre-dispatch gates --json error envelope', () => {
+  let outsideGitRoot = '';
+  let outsideGitCwd: string;
+  let outsideGitDefaultHome: string;
+  let outsideGitArchonHome: string;
+
+  afterAll(async () => {
+    if (outsideGitRoot) await removeTempTree(outsideGitRoot);
+  });
+
+  beforeAll(async () => {
+    outsideGitRoot = mkdtempSync(join(tmpdir(), 'archon-cli-no-git-'));
+    outsideGitCwd = join(outsideGitRoot, 'project');
+    outsideGitDefaultHome = join(outsideGitRoot, 'default-home');
+    outsideGitArchonHome = join(outsideGitRoot, 'archon-home');
+    mkdirSync(outsideGitCwd);
+    mkdirSync(outsideGitDefaultHome);
+
+    const registry = new SqliteAdapter(join(outsideGitArchonHome, 'archon.db'));
+    await registry.close();
+  });
+
   it('emits { ok: false } on stdout when --cwd does not exist', () => {
     const { status, envelope } = spawnJsonError([
       'workflow',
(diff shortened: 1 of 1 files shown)


Wirasm:
# Review report — PR #3063

## Verdict

Ready — action: `none`. The changed test gives the real CLI child an isolated empty SQLite registry, and the non-Git gate necessarily consults that registry before returning the required status-1 JSON failure envelope; no open Critical or Important finding remains.

## Accepted contract

Review the recovery of the outside-git JSON-envelope test isolation work. The focused evidence must show that the real CLI subprocess still performs registered-folder lookup, returns exit status 1, emits a parseable `{ ok: false }` envelope, and confines registry state to test-owned paths.

This is a test-only change. Production CLI and folder-project behavior must remain unchanged. This review does not close the wider #2982 audit.

## Reviewed head SHA

`9bef0db33c7b038e6c54f782cc65dd7caff3e62e`

## Findings

### Rejected findings

- R1 — Important — sources: `[seams]` — disproved. `packages/core/src/db/connection.ts:36` owns the production default selected from the child process environment, but this test intentionally supplies an explicit `ARCHON_HOME` to that child. The fixture's `join(outsideGitArchonHome, 'archon.db')` at `packages/cli/src/cli.test.ts:1067` creates the exact currently selected SQLite location, and its fake-home assertion at `:1102` checks the separate fallback location. The proposed production API change would add a test-only parameter and does not establish a current defect in the accepted behavior.
- R2 — Important — sources: `[tests]` — disproved. The accepted contract explicitly requires the outside-Git invocation to return status 1 with `{ ok: false }`; it does not require this test to prove successful dispatch of a registered folder. `packages/cli/src/cli.test.ts:1087-1102` asserts that required negative path. In that path, `packages/cli/src/cli.ts:544-547` performs exact and prefix registered-folder lookup before `:577-584` returns the non-Git error. The existing core integration coverage separately proves registered-folder lookup results. Replacing this test with a success case would fail to protect the specified JSON-error contract.

### Suggestions

- S1 — Suggestion — sources: `[simplify]` — the fixture could be made local to its sole test and let the child create the empty registry. The present suite-owned setup is small, cleaned by `removeTempTree` at `packages/cli/src/cli.test.ts:1055-1057`, and leaves the asserted behavior intact, so this is not a blocker.

## Discoveries

0 accepted discoveries. [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/33142f2833725f9ba386e1b3396c9b61/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/33142f2833725f9ba386e1b3396c9b61/discoveries.md) record the consolidated result. Adjacent discoveries do not affect readiness.

## Review coverage

Full review. Required lenses `code`, `seams`, `simplify`, and `tests` reported. `code` found no correctness defect; the other reports supplied the rejected findings and suggestion above. `errors` and `docs` were disabled because this test-only diff adds neither an error-handling implementation change nor user-facing documentation surface. No producer discovery records were present.

Validation: the documented focused command `bun test src/cli.test.ts` from `packages/cli` could not begin because this checkout cannot resolve the existing workspace import `@archon/paths/test-utils`. Bun failed before any test executed. `git diff --check dev...HEAD` passed. This environment failure does not contradict the statically verified child process, environment, and gate path above.

Published as the canonical PR comment: https://github.com/coleam00/Archon/pull/3063#issuecomment-5479192704

Wirasm:
Superseded by PR #3057, which absorbed the #2982 outside-git isolation and merged to dev first. Dev's version covers the same defect more completely: the shared `spawnJsonError` helper now routes every --json envelope test through a suite-owned scratch `ARCHON_HOME` (seeded once outside the timed bodies, with a loud beforeAll tripwire if the helper ever stops handing `ARCHON_HOME` to the child), plus a sentinel test proving the default home stays untouched, and it removes the busy_timeout contention on the operator's registry that caused the timeouts. The extra env blanking here (`DATABASE_URL` etc.) belongs to #2982's broader audit if it's ever needed — the seed tripwire fails loudly rather than silently falling back in the meantime.

PR #3077, opened 2026-08-31
Centralize agent node failure finalization

### LINKED ISSUE #2698: refactor(workflows): executeNodeInternal open-codes its terminal failure four times — give it the finalizer executeLoopNode already has
## Problem

`executeNodeInternal` ends a failing node in four separate places, and each one repeats the same sequence by hand:

1. compute the duration,
2. log a warn or error,
3. write the JSONL transcript row via `logNodeError`,
4. write the `node_failed` DB event via `createWorkflowEvent`, spreading `nodeUsageEventData()` and `namedSessionAuditData`,
5. return a failure result.

Nothing ties the five together. A block that performs four of them compiles, passes review, and is wrong only in a way nobody sees until someone reads a transcript months later.

## This is not a hypothetical — #2694 is the evidence

#2694 added the transcript sink to the failure paths that hold usage. **Two of the four blocks did not have it**, and they were fixed one at a time, by hand, with nothing signalling the omission. The PR's author said so plainly rather than presenting the result as complete.

The same PR also found that the user-cancel path has **two** exits rather than the one its issue named, and that fixing only the first would have made a cancelled node's transcript row depend on whether the node declared `output_format`. That is the failure mode this shape produces: not a missing feature, but an inconsistency that presents as flakiness.

Worth stating precisely, because it bounds the value: the seam guard from #2685 catches a *dropped axis* at a sink it knows about. It does not catch a *missing sink*. A fifth terminal-failure block added tomorrow with no `logNodeError` call is invisible to it.

## The shape already exists in this file

`executeLoopNode` does it correctly. Its failure path is one helper that builds the usage payload once and spreads it into both sinks, with a comment saying why:

```ts
// A loop that failed part-way still paid for the iterations it ran. Built once and
// spread into both sinks, so the transcript row and the persisted event cannot
// disagree — the same reason executeNodeInternal routes both through
// nodeUsageEventData (#2693).
```

So the argument is not "extract a helper because duplication is bad". It is that one of the two functions in this file already has the shape, the other does not, and the one that does not is where the misses happened.

## Design (settled 31 August 2026 — read the four blocks before editing)

Give `executeNodeInternal` the finalizer `executeLoopNode` already has:

```ts
const failAgentNode = async (
  error: string,
  extras: { output?: string; data?: Record<string, unknown> } = {}
): Promise<NodeExecutionResult> => { /* owns every step below */ };
```

Two optional fields, fewer than `failLoopNode`'s four, because most of what looks per-site is ambient to the closure:

- **usage** — `nodeUsageEventData()` already closes over the node; the finalizer calls it rather than accepting it.
- **`namedSessionAuditData`** — same.
- **duration** — computed inside from `nodeStartTime`. See the drift decision below.
- **the diagnostic log** — stays at the call site. This is the answer to "without growing into a configuration object": the finalizer owns the node's *lifecycle*, never its diagnostics. `failLoopNode`'s own docblock already states this contract — *"Call sites keep their specific diagnostic logs/events … this closes the node's lifecycle exactly once."* The `info`/`warn`/`error` split and the catch block's two-branch log are diagnostics and do not move.
- **the abort branch's early return** — the finalizer returns `NodeExecutionResult`, so the site does `return failAgentNode(...)`, exactly as the loop sites do.

### The four blocks, as measured

| block | diagnostic log | `duration_ms` in payload |
| --- | --- | --- |
| `:3097` cancelled during streaming | `info` | yes |
| `:3159` credit exhausted | `warn` | no |
| `:3202` empty output | `error` | yes |
| `:3362` catch | `error` (two branches) | no |

Everything else is already identical across all four: `logNodeError(..., nodeUsageEventData())`, a `createWorkflowEvent` carrying `{ error, ...nodeUsageEventData(), ...namedSessionAuditData }` with the same `.catch`, and an emitter call using `nodeName: commandName ?? node.id`.

### Decision: fix the drift, do not preserve it

The `duration_ms` split is drift, not intent, and it is the gap #2398 already recorded as *"`node_failed` carries no timing"*. The finalizer computes duration, so all four paths gain it and two stop being silently different.

**This supersedes the acceptance line below that asked for "the same events, the same payloads, the same order".** Operator decision: when a refactor exposes drift on the path it is already touching, fix it rather than carry it forward for a cleaner diff. Adding a field to `node_failed.data` is additive — no reader breaks, resume does not read duration, and the console renders it only when present.

### Also fold in: the cache cleanup

`lastNodeCancelCheck.delete` / `lastNodeActivityUpdate.delete` appears five times (`:3133`, `:3183`, `:3231`, `:3323`, `:3345`). It is part of ending a node, it is repeated, and it is exactly the shape that goes missing on the sixth. It belongs in the finalizer. If a terminal path deliberately skips it, that is a finding to report rather than a detail to preserve.

### Correction: only two of the "six other sites" are in scope

The earlier draft said six `logNodeError` call sites fail before any provider call. Attributed to their owning functions:

| site | function |
| --- | --- |
| `:2171`, `:2249` | `executeNodeInternal` — **in scope** |
| `:3939` | `executeBashNode` — out |
| `:4181`, `:4212`, `:4333` | `executeScriptNode` — out |

Four of the six are in the deterministic executors, which #2718 deliberately kept as separate functions after a full side-by-side read. Folding them into an agent-node finalizer would undo that decision.

`:2171` and `:2249` **do** route through the finalizer, with no extras. That makes "this node spent nothing" a deliberate argument rather than an absent parameter — the same reasoning that made #2685's `null` rows explicit — and it is what makes "a sixth sink is one edit" true rather than aspirational, because every terminal exit then leaves through one function.

## Acceptance

- Every terminal failure exit from `executeNodeInternal` returns through `failAgentNode`. No `logNodeError` call and no `node_failed` event is written anywhere else in that function.
- `node_failed` carries `duration_ms` on all four paths; two gain it.
- Cache cleanup happens once, inside the finalizer.
- `:2171` and `:2249` route through it with no usage extras.
- `executeBashNode` and `executeScriptNode` are untouched — #2718's split stands.
- **Seen red:** delete the transcript write from inside the finalizer and a test fails. A guard is evidence only after it has been observed failing, and the defect being fixed here is precisely a missing sink that nothing caught.
- Behaviour is otherwise unchanged: same events, same order, same payloads apart from the added `duration_ms`.

## Provenance

Recommended by the author of #2694 and deliberately not filed there — the recommendation was to keep that PR scoped to the transcript fix. Related: #2694, #2685 (the seam guard, which covers axes but not sinks), #2676 (the `WorkflowUsage` carrier), #2614.

Wirasm:
## Problem and outcome

Agent-node failures in `executeNodeInternal` were finalized by six separate blocks. That let the JSONL transcript, persisted `node_failed` event, emitted event, throttle cleanup, and returned failure result drift; credit-exhaustion and catch paths also omitted `duration_ms`.

- **Outcome:** Every agent-node failure after `node_started` now uses one finalizer, and every resulting `node_failed` event includes `duration_ms`.
- **Invariant:** The transcript, persisted event, emitter, cache cleanup, failed result, and reported usage describe the same terminal failure.
- **Scope boundary:** `executeBashNode`, `executeScriptNode`, and other deterministic executor failure writers are unchanged.
- **Root cause:** Separate terminal writers repeated the same lifecycle work with different payloads and cleanup.

## Review guidance

- **Feedback requested:** Confirm the finalizer owns all six `executeNodeInternal` failure exits while preserving each call site's diagnostic behavior.
- **Start here:** `packages/workflows/src/dag-executor.ts:2172` — `failAgentNode` is the single terminal-failure owner.
- **Review order:** Review the finalizer, then its six call sites, then the transcript and duration assertions in `packages/workflows/src/dag-executor.test.ts`.
- **Lower-attention areas:** The test edits add focused assertions only; no generated artifacts or schema changes.

## Solution

`failAgentNode` builds usage once and writes the transcript error row, persisted event (including duration), emitter event, throttle-cache cleanup, and failed result. Command loading, prompt substitution, streaming cancellation, credit exhaustion, empty output, and the outer catch now delegate to it. Local diagnostic logging and the prompt-substitution user message remain at their original call sites.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Failure sinks could disagree about a node's terminal lifecycle data. | All six agent-node failure exits use one lifecycle finalizer. |
| **Failure behavior** | Credit-exhaustion and catch-path `node_failed` events omitted `duration_ms`. | Every finalizer-owned `node_failed` event records `duration_ms`; transcript failures retain node usage. |

## Validation

- `bun test src/dag-executor.test.ts` — 703 passed, 0 failed — exercises credit-exhaustion transcript and duration handling plus provider-error duration handling.
- `bun run type-check` in `packages/workflows` — passed.
- `bun run validate` — passed, including generated-artifact checks, API type generation, package type checks, lint, formatting, installer tests, package tests, and script tests.
- **Not verified:** Nothing material; focused lifecycle assertions and the full validation suite passed.

## Links

- Closes #2698

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow failure handling so agent-node errors are consistently recorded and reported.
  * Failed workflow events now include execution duration, usage, and session audit details.
  * Ensured cancellation, credit exhaustion, empty results, and other failure scenarios clean up correctly.
  * Preserved accurate failure costs and token usage in workflow results.
* **Tests**
  * Added coverage verifying failure and cancellation events include valid execution durations.
  * Added validation that credit-exhausted workflows produce the expected error transcript and failure event.

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -15669,6 +15669,11 @@ describe('executeDagWorkflow -- run usage survives every disposition', () => {
     expect(cancelRow?.error).toBe('Cancelled by user');
     expect(cancelRow?.cost_usd).toBe(0.02);
     expect(cancelRow?.tokens).toEqual({ input: 30, output: 3 });
+
+    const failedEvent = store.createWorkflowEvent.mock.calls
+      .map(([event]) => event)
+      .find(event => event.event_type === 'node_failed' && event.step_name === 'only');
+    expect(typeof failedEvent?.data?.duration_ms).toBe('number');
   });
 
   it('a cancel that surfaces as a thrown error still leaves a transcript row', async () => {
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -13611,6 +13611,7 @@ describe('executeDagWorkflow -- credit exhaustion', () => {
     const deps = createMockDeps(store);
     const platform = createMockPlatform();
     const workflowRun = makeWorkflowRun('credit-exhaustion-run');
+    const logDir = join(testDir, 'logs');
 
     await executeDagWorkflow(
       dagOptions({
@@ -13629,15 +13630,27 @@ describe('executeDagWorkflow -- credit exhaustion', () => {
           ],
         },
         workflowRun,
+        logDir,
       })
     );
 
     // node_failed (not node_completed) must have been stored
-    const events = (
-      store.createWorkflowEvent as Mock<IWorkflowStore['createWorkflowEvent']>
-    ).mock.calls.map((c: unknown[]) => (c[0] as { event_type: string }).event_type);
+    const eventCalls = (store.createWorkflowEvent as Mock<IWorkflowStore['createWorkflowEvent']>)
+      .mock.calls;
+    const events = eventCalls.map((c: unknown[]) => (c[0] as { event_type: string }).event_type);
     expect(events).toContain('node_failed');
     expect(events).not.toContain('node_completed');
+    const failedEvent = eventCalls.find(
+      (c: unknown[]) => (c[0] as { event_type: string }).event_type === 'node_failed'
+    );
+    expect(
+      typeof (failedEvent?.[0] as { data?: { duration_ms?: unknown } }).data?.duration_ms
+    ).toBe('number');
+
+    const transcriptFailures = (await readTranscript(logDir, workflowRun.id)).filter(
+      row => row.type === 'node_error' && row.step === 'investigate'
+    );
+    expect(transcriptFailures).toHaveLength(1);
 
     // Overall workflow should be marked failed
     expect(store.failWorkflowRun).toHaveBeenCalled();
@@ -15589,6 +15602,9 @@ describe('executeDagWorkflow -- run usage survives every disposition', () => {
     );
     expect(failedEvents.length).toBe(1);
     expect((failedEvents[0][0] as { data: Record<string, unknown> }).data.cost_usd).toBe(0.02);
+    expect(typeof (failedEvents[0][0] as { data: Record<string, unknown> }).data.duration_ms).toBe(
+      'number'
+    );
 
     // And the run total is the money burned across both nodes.
     expect(runUsageWrites(store)).toEqual([
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -2157,6 +2157,65 @@ async function executeNodeInternal(
     ...(resolvedEffort !== undefined ? { effort: resolvedEffort } : {}),
   });
 
+  let nodeTokens: TokenUsage | undefined;
+  let nodeCostUsd: number | undefined;
+
+  const nodeUsageEventData = (): WorkflowUsage => ({
+    ...(nodeTokens !== undefined ? { tokens: nodeTokens } : {}),
+    ...(nodeCostUsd !== undefined ? { cost_usd: nodeCostUsd } : {}),
+  });
+
+  const nodeKey = `${workflowRun.id}:${node.id}`;
+
+  // Every failed result after node_started passes through this finalizer so the
+  // transcript, persisted event, emitter, and throttle cleanup cannot drift.
+  const failAgentNode = async (
+    error: string,
+    extras: { output?: string; data?: Record<string, unknown> } = {}
+  ): Promise<NodeExecutionResult> => {
+    const usage = nodeUsageEventData();
+    await logNodeError(logDir, workflowRun.id, node.id, error, usage);
+
+    deps.store
+      .createWorkflowEvent({
+        workflow_run_id: workflowRun.id,
+        event_type: 'node_failed',
+        step_name: stepName,
+        data: {
+          error,
+          duration_ms: Date.now() - nodeStartTime,
+          ...usage,
+          ...namedSessionAuditData,
+          ...(extras.data ?? {}),
+        },
+      })
+      .catch((err: Error) => {
+        getLog().error(
+          { err, workflowRunId: workflowRun.id, eventType: 'node_failed' },
+          'workflow_event_persist_failed'
+        );
+      });
+
+    emitter.emit({
+      type: 'node_failed',
+      runId: workflowRun.id,
+      nodeId: node.id,
+      nodeName: commandName ?? node.id,
+      error,
+    });
+
+    lastNodeCancelCheck.delete(nodeKey);
+    lastNodeActivityUpdate.delete(nodeKey);
+
+    return {
+      state: 'failed',
+      output: extras.output ?? '',
+      error,
+      costUsd: nodeCostUsd,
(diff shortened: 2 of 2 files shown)


Wirasm:
# Review report: PR #3077

## Verdict

**Ready — action: none.** The correction adds the missing persisted-event duration assertion to the streaming-cancellation route, and the focused workflow suite passes.

## Accepted contract

PR #3077 must make one `executeNodeInternal` terminal-failure finalizer own the transcript error row, persisted and emitted `node_failed` events, throttle-cache cleanup, and failed result. All four usage-carrying failure exits must persist `duration_ms`; the two pre-provider `logNodeError` exits must also use the finalizer without usage extras. Existing diagnostic logging and lifecycle-event order must remain unchanged.

The accepted boundary is the six `executeNodeInternal` exits. `executeBashNode`, `executeScriptNode`, and other deterministic-executor failure writers remain unchanged.

## Reviewed head SHA

`19d4725573197005ee9a814bc740cc394955ccb4`

## Findings

### Critical

None.

### Important

None open.

### Suggestions

#### R2 — Remove the unused arbitrary event-data extension point

Sources: `[simplify]`

`failAgentNode` accepts `extras.data?: Record<string, unknown>` at `packages/workflows/src/dag-executor.ts:2172-2175` and spreads it into the persisted event at `:2184-2190`, but no caller supplies `data`; the only non-default calls supply `output` at `:3099` and `:3116`. This does not violate the accepted behavior today, so it is non-blocking. A smaller finalizer can accept `output = ''` directly and keep future persisted-field changes explicit at the finalizer boundary.

### Rejected findings

None.

## Prior findings

| ID | Status | Verdict |
| --- | --- | --- |
| R1 | fixed | The correction adds `packages/workflows/src/dag-executor.test.ts:15673-15676`, which selects the cancellation route's persisted `node_failed` event and asserts `data.duration_ms` is a number. The route returns through `failAgentNode` at `packages/workflows/src/dag-executor.ts:3092-3100`; the finalizer persists that field at `:2180-2190`. `cd packages/workflows && bun test src/dag-executor.test.ts -t 'a node cancelled by the user leaves its spend in the transcript'` passed. |
| R2 | open | Suggestion only; it does not affect readiness. |

## Discoveries

1 accepted adjacent discovery: “Failed loop-group agent lifecycle events lose their iteration.” Adjacent discoveries do not affect readiness. See [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/75738bb5dacc053295adcd2d15154475/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/75738bb5dacc053295adcd2d15154475/discoveries.md).

## Review coverage

One continuation reviewer verified the prior report and the correction delta. The original selected concerns were terminal-failure ownership, persistence and lifecycle-event seams, simplification, and regression protection; the original error-path review found no separate defect, and docs was disabled because the change is internal. The delta changes only a focused test assertion, adds no user-facing surface, and required no re-evaluation of a gated-off lens. All required correction evidence was inspectable.

The full documented package test command passed: `cd packages/workflows && bun test src/dag-executor.test.ts` — 703 passed, 0 failed.

PR #3081, opened 2026-08-31
Attach graph plans to resolved workflows

### LINKED ISSUE #2487: refactor(workflows): split authored vs resolved workflow — make the graph plan a field, not a computation
The authored graph and the resolved graph are the same type today. `validateDagStructure` checks ids, `depends_on` resolution, cycles and `$node.output` refs; `buildTopologicalLayers` computes layers at run time. Nothing in the type system records that a graph has been through either.

`WorkflowDefinition` is the type on **both** sides of `expandWorkflowIncludes`. Its own docblock states the invariant the type does not carry:

> `nodes`' element type is `DagNode | IncludeDirective` (not `DagNode[]` alone) because `dagNodeSchema` still parses `include:` entries at this pre-expansion stage (#2486) — `expandWorkflowIncludes` consumes every `IncludeDirective` before the executor ever sees a `WorkflowDefinition`.

## The evidence: 15 unchecked casts and a guard that lies to the compiler

Narrowing from `(DagNode | IncludeDirective)[]` to `DagNode[]` is re-derived independently at **15 production sites**, each an unchecked `as DagNode[]` justified only by a comment restating the same unproven claim. Measured at `781ecf07`:

| Group | Sites |
|---|---|
| Top-level `workflow.nodes` (6) | `executor.ts` (`telemetryNodes`); `dry-run.ts:243`, `:1200`; `include-expander.ts:862`; `cli/src/commands/workflow.ts:2037`; `core/src/orchestrator/orchestrator.ts:370` |
| Nested `loop_group.nodes` (9) | `dag-executor.ts:4649`, `:4684`, `:4724`, `:4946`, `:5653`; `dry-run.ts:236`, `:872`; `include-expander.ts:1176`; `utils/workflow-requirements.ts:86` |

Two of them are outside `@archon/workflows` entirely — both the same `assertComposedGateDriveable(workflow.nodes as DagNode[])` call.

Exactly one place actually *enforces* the invariant, inside `runLayers`, and it has to lie to the compiler twice to do it:

```ts
if (isIncludeDirective(node as DagNode | IncludeDirective)) {
  const includeNode = node as unknown as IncludeDirective;
  throw new Error(
    `Internal error: include node '${includeNode.id}' reached the executor unexpanded. ...`
  );
}
```

`node` is already typed `DagNode` there. The guard casts it *back* to the union to run the check, then again through `unknown` to read the id. A runtime assertion standing in for a type, written against a state its own parameter type forbids.

`include-expander.ts:858-862` already defers the fix to this issue by name.

## Proposed

Make "resolved" a property of the type.

```ts
type ResolvedWorkflow = Omit<WorkflowDefinition, 'nodes'> & {
  nodes: DagNode[];
  plan: GraphPlan;
};

interface GraphPlan {
  readonly layers: readonly (readonly DagNode[])[];
  readonly sinks: readonly string[];
}
```

`GraphPlan` being a **field rather than a runtime computation** is the type-level statement of *"fully resolved before execution."* It cannot be produced except by running the walk, so an unplanned graph is unrepresentable. That encodes the invariant the whole governance model rests on — load-time validation, resume, and the audit trail all assume the executor sees a flat static DAG.

Two shaping decisions, both corrections to this issue's original sketch:

**Derive from `WorkflowDefinition`; do not invent a `Graph` type.** The repository already names this concept. A `Graph` beside `WorkflowDefinition` gives one concept two names, against `engineering.md`'s *"Use one precise name for each concept"* and AGENTS.md's *"Derive types from the owning schema."* `WorkflowDefinition` is Zod-derived, so the resolved form derives from it. `GraphId`/`NodeId` branded types have no current requirement and are not part of this.

**`nodes` stays mutable, not `readonly`.** `DagNode[]` is assignable to `(DagNode | IncludeDirective)[]`, so `ResolvedWorkflow` remains a **subtype** of `WorkflowDefinition` and the type flows outward — including to `@archon/cli` and `@archon/core` — without a package-wide ripple. `readonly` would break that assignability everywhere for an immutability property nothing here needs; that is a separate change.

## What the original sketch got wrong

**`primarySink` does not belong on the plan.** It is `returns:`-aware and feeds the include's `$blk.output`, a composed fan-out instance's terminal output, and a child sub-run's persisted `metadata.summary_value` and `outcome_field` verdict. That is a persisted child-run contract — top of `engineering.md`'s risk taxonomy — and it is an include-**site** concept, not a property of a graph. It stays in `include-expander.ts`. `GraphPlan` carries `sinks` (pure topology), which deletes the duplicate `allDependencies` scan in `executeDagWorkflow`'s terminal-output branch.

**"Moving `buildTopologicalLayers` from execution time to construction time" is true for one of its four call sites.** The others are `dag-executor.ts:4958` (loop_group body, on freshly `$LOOP_PREV`-substituted clones), `dag-executor.ts:9467` (composed fan-out instance) and `dry-run.ts:1120`. The right primitive is one `planGraph(nodes)` called at load time for the workflow and at runtime for each repeated body — which is exactly what `.archon/workflow-language-constitution.md:114` licenses:

> The engine may own runtime cardinality only when it repeats that already-resolved body and persists deterministic instance identity before scheduling; runtime selection or construction of structure is sub-run territory.

A fan-out instance is a namespaced copy of a body the expander already validated in full before any parent node ran. Only the cardinality and the id prefix are runtime.

**The runtime guard does not become unreachable on its own.** Nine of the fifteen casts feed loop-group bodies that reach `runLayers` without passing through the workflow-level type, so a workflow-level split leaves the guard with a real job. Making those structural would need a resolved recursive node type — and `LoopGroupNodeConfig` is hand-written by necessity, for the reason the code states at `schemas/dag-node.ts:447-451`: the body is recursive (`loopGroupNodeConfigSchema` ← `dagNodeSchema` ← `loopGroupNodeConfigSchema`) and zod v4 needs the explicit `z.ZodType<T>` annotation to break the TS7022 inference cycle. A resolved twin means either a second hand-written recursive type — the defect this batch removes — or making `DagNode` generic in its body type, rippling through the ~135 narrowing sites #2486 just migrated.

So the nine collapse into **one shared checked helper** with a real predicate that throws naming the offending node. Fifteen unchecked casts become one resolved type plus one checked narrowing, and the guard then genuinely guards nothing.

## Scope

1. `ResolvedWorkflow` + `GraphPlan`, produced by `planGraph` (plans any node set) and `resolveWorkflow` (the only way to produce the type the executor accepts), at `expandWorkflowIncludes`' exit.
2. All four `buildTopologicalLayers` call sites route through `planGraph`; the top-level one becomes a read of `workflow.plan.layers`. The cycle throw moves into `planGraph` and stops being a runtime safety check the executor performs.
3. One `resolvedBodyNodes(group)` helper replaces the nine nested casts.
4. The six top-level casts disappear. This includes the two outside the package: `assertComposedGateDriveable` already takes `readonly DagNode[]`, so it needs **no signature change** — a `ResolvedWorkflow`'s `nodes` satisfies it and the cast simply goes. Keeping the resolved type alive to those call sites means annotating `discoverWorkflows`/`discoverWorkflowsWithConfig`'s result, making `resolveWorkflowName` generic (`<T extends WorkflowDefinition>(name, workflows: readonly T[]): T | undefined` — its body reads only `.name`), and `dispatchBackgroundWorkflowOwned` plus its one wrapper. Both packages already depend on `@archon/workflows` and already import `DagNode` from it.
5. Test fixtures construct through `resolveWorkflow`. No test-only escape hatch — a bypass would make "an unplanned graph is unrepresentable" false, which is the point.

## Acceptance

- [ ] `ResolvedWorkflow` cannot be constructed without a `GraphPlan`
- [ ] The executor's and `dryRunWorkflow`'s signatures take the resolved form; the loader's output type stays the authored one
- [ ] The runtime include guard is removed, its failure behaviour preserved by a checked narrowing that throws naming the node — reached via the shared helper, **not** as an automatic consequence of the type split
- [ ] `grep -rn "as DagNode\[\]" --include='*.ts' packages/*/src | grep -v '\.test\.ts'` returns nothing, repository-wide
- [ ] No behaviour change: the same YAML produces the same layers, terminal output, and telemetry. `dag-executor.test.ts`, `include-expander.test.ts` and `dry-run.test.ts` pass with no test lost and none edited, measured against your own branch point. Do not trust a count written here: these read 688 / 115 / 66 at `781ecf07` and 700 / 122 / 74 on `dev` @ `0ea658bf8`, and they move whenever anyone adds a test.

Size: **S–M**. The original **S** predated the nine nested casts and the two out-of-package ones.

## Sequencing

This is PR-C of three.

- **PR-A — #2933.** The six internal node executors take `RunLayersContext`.
- **PR-B — #2302.** `executeDagWorkflow` takes an options object derived from `RunLayersContext`, rewriting 436 call sites.
- **PR-C — this issue.** Lands last on purpose: it changes the `workflow` argument at every one of those 435 test call sites, from a bare `{ name, nodes }` literal to a `resolveWorkflow(...)` call. After PR-B that is a wrap of an already-named field; before it, both issues rewrite the same sites and the churn is paid and reviewed twice.

## Prerequisite

Met. #2406 landed (workflows tests are type-checked) and #2486 landed (the `DagNode` union, which demoted `include:` out of it). #2407 is open and does not block this.

Wirasm:
## Problem and outcome

Authored workflow definitions and include-expanded execution graphs shared one type, so execution, simulation, and orchestration each had to reassert that no `include:` directive remained and recompute graph layers at runtime.

- **Outcome:** Discovery now produces `ResolvedWorkflow`, carrying include-free `DagNode[]` nodes plus a `GraphPlan` of execution layers and sinks; execution and dry run consume that plan.
- **Invariant:** An execution boundary can only accept a resolved, planned workflow; a surviving include directive fails at the shared resolution boundary with its node id.
- **Scope boundary:** Runtime-shaped loop and composed fan-out bodies still receive a fresh plan after their deterministic substitutions or instance prefixes are applied.

## Review guidance

- **Feedback requested:** Validate the authored-to-resolved workflow boundary and the plan ownership across include expansion, execution, and dry run.
- **Start here:** `packages/workflows/src/graph-plan.ts:60` — creates graph plans and constructs the only resolved workflow shape accepted by execution.
- **Review order:** `graph-plan.ts`, `include-expander.ts`, then `dag-executor.ts` and `dry-run.ts`; test updates follow the changed contracts.
- **Lower-attention areas:** Core and CLI changes propagate the resolved type; test fixture changes construct resolved workflows through the shared helper.

## Solution

`WorkflowDefinition` remains the authored YAML shape. `ResolvedWorkflow` derives from it with `DagNode[]` and an attached `GraphPlan`. Include expansion validates and resolves each flattened workflow once, reusing the plan's sinks for include wiring.

The DAG executor reads the top-level plan rather than rebuilding layers or terminal sinks. Dry run follows the same plan. Loop-group and composed fan-out instances retain runtime planning because their node bodies are shaped at runtime; their include-free node boundary is checked through the shared helper.

## Behavior change

| | Before | After |
|---|---|---|
| **Workflow representation** | Authored and expanded workflows shared `WorkflowDefinition`; consumers narrowed nodes with casts. | Expanded workflows are `ResolvedWorkflow`, whose nodes and graph plan carry the execution invariant. |
| **Graph scheduling** | Top-level execution and dry run recomputed layers and sinks. | They use the plan attached during resolution. |
| **Invalid include state** | The executor had a runtime guard for an unexpanded include. | Resolution rejects a surviving include before an execution boundary can receive the workflow. |

## Architecture

```mermaid
flowchart LR
  A[Authored WorkflowDefinition] --> B[Include expansion]
  B --> C[resolveWorkflow]
  C --> D[ResolvedWorkflow: nodes + GraphPlan]
  D --> E[Executor]
  D --> F[Dry run]
```

### Changed seams

|---|---|---|
| Include expansion → execution | Expansion returns `ResolvedWorkflow` and attaches the plan after flattened-graph validation. | `packages/workflows/src/include-expander.ts:1063`, `packages/workflows/src/include-expander.ts:1289` |
| Resolution → executor | Executor accepts `ResolvedWorkflow` and reads `workflow.plan.layers` and `workflow.plan.sinks`. | `packages/workflows/src/dag-executor.ts:11604`, `packages/workflows/src/dag-executor.ts:12247` |
| Resolution → dry run | Dry run accepts the resolved shape and simulates its attached plan. | `packages/workflows/src/dry-run.ts:1208`, `packages/workflows/src/dry-run.ts:1280` |

## Validation

- Focused workflow and core tests for DAG execution, include expansion, dry run, executor preamble/execution, sub-runs, and orchestration — passed.
- `bun run validate` — passed, including generated-artifact checks, API type generation, package type checks, lint, formatting, installer tests, package tests, and script tests.
- `git diff --check b27f988d41451c53ec64205b345c3a266acd1277..HEAD` — passed.
- `rg -n "as DagNode\\[\\]" packages --glob '!**/*.test.ts'` — no production matches.
- `rg -n "buildTopologicalLayers" packages` — no matches.

## Links

- Closes #2487

## Summary by CodeRabbit

* **New Features**
  * Added workflow graph planning with execution layers, terminal-node tracking, cycle detection, and include validation.
  * Added resolved workflow support and public exports for graph planning utilities.
* **Bug Fixes**
  * Workflows with unresolved includes now fail clearly instead of falling back to incomplete definitions.
  * Improved handling of loop-group nodes during execution and dry runs.
* **Tests**
  * Expanded coverage for graph planning, cycle detection, execution ordering, and resolved workflow behavior.

### DIFF
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -84,6 +84,7 @@ import {
   assertWorkflowRequirementsMet,
   resolveTopLevelInputs,
 } from '@archon/workflows/utils/workflow-requirements';
+import type { RequirementBearingWorkflow } from '@archon/workflows/utils/workflow-requirements';
 import { parseInputAssignments } from '@archon/workflows/workflow-inputs';
 import { formatDeprecationNotice } from '@archon/workflows/deprecation';
 import {
@@ -808,7 +809,9 @@ function assertWorkflowNotWorktreePinnedForFolder(
  * GitHub App + TOKEN_ENCRYPTION_KEY are both configured — identical semantics
  * to the orchestrator gate.
  */
-async function assertCliWorkflowRequirementsMet(workflow: WorkflowDefinition): Promise<void> {
+async function assertCliWorkflowRequirementsMet(
+  workflow: RequirementBearingWorkflow
+): Promise<void> {
   if (!isPerUserGitHubEnabled() || !workflow.requires?.length) return;
 
   // Resolve the acting CLI user (ARCHON_USER_ID, else $USER/$USERNAME) → Archon
@@ -869,7 +872,9 @@ function resolveTitleAssistantType(
  * `archon ai tier list`.
  */
 export async function maybePrintTierNotice(
-  workflow: WorkflowDefinition,
+  workflow: Pick<WorkflowDefinition, 'model'> & {
+    readonly nodes: readonly WorkflowDefinition['nodes'][number][];
+  },
   cwd: string,
   cliUserId: string | undefined,
   quiet: boolean | undefined
@@ -1070,7 +1075,9 @@ export function emitParseWarnings(
  * a user driving runs programmatically still has to learn the default they
  * picked is scheduled for removal.
  */
-export function emitDeprecationNotice(workflow: WorkflowDefinition): void {
+export function emitDeprecationNotice(
+  workflow: Pick<WorkflowDefinition, 'name' | 'deprecated'>
+): void {
   const notice = formatDeprecationNotice(workflow);
   if (notice) console.warn(notice);
 }
--- a/packages/cli/src/commands/workflow.ts
+++ b/packages/cli/src/commands/workflow.ts
@@ -104,7 +104,6 @@ import type {
   WorkflowSource,
   WorkflowWithSource,
 } from '@archon/workflows/schemas/workflow';
-import type { DagNode } from '@archon/workflows/schemas/dag-node';
 import type { WorkflowRunConfigInput } from '@archon/workflows/schemas/run-config';
 import {
   workflowRunStatusSchema,
@@ -1987,10 +1986,7 @@ async function runWorkflowWithOwnedSource(
     // discoverable via `workflow status`/`runs`, not a silent hang nobody knows exists.
     if (!isContinuation) {
       assertInteractiveClassNotBackgrounded(workflow);
-      // Already-expanded — discoverWorkflowsWithConfig's output never contains an
-      // IncludeDirective (#2486); the type admits one only for the pre-expansion display
-      // shape (`WorkflowWithSource.declared`), which `workflow` here is not.
-      assertComposedGateDriveable(workflow.nodes as DagNode[]);
+      assertComposedGateDriveable(workflow.nodes);
     }
 
     const childConversationId = options.conversationId ?? generateConversationId();
--- a/packages/core/src/handlers/command-handler.ts
+++ b/packages/core/src/handlers/command-handler.ts
@@ -25,7 +25,7 @@ import { createWorkflowDeps } from '../workflows/store-adapter';
 import type {
   WorkflowWithSource,
   WorkflowLoadError,
-  WorkflowDefinition,
+  ResolvedWorkflow,
 } from '@archon/workflows/schemas/workflow';
 import { isContainerRun } from '@archon/workflows/schemas/workflow-run';
 import * as workflowDb from '../db/workflows';
@@ -1158,7 +1158,7 @@ async function handleWorkflowCommand(
         'cmd.workflows_discovered'
       );
 
-      let workflow: WorkflowDefinition | undefined;
+      let workflow: ResolvedWorkflow | undefined;
       try {
         workflow = resolveWorkflowName(workflowName, workflows);
       } catch (err) {
--- a/packages/core/src/orchestrator/orchestrator-agent.test.ts
+++ b/packages/core/src/orchestrator/orchestrator-agent.test.ts
@@ -15,6 +15,7 @@
 import { mock, describe, test, expect, beforeEach, afterEach } from 'bun:test';
 import { createMockLogger } from '../test/mocks/logger';
 import {
+  makeTestResolvedWorkflow,
   makeTestWorkflow,
   makeTestWorkflowWithSource,
   withObservableCapturedSource,
@@ -2142,7 +2143,7 @@ describe('workflow dispatch routing — interactive flag', () => {
       success: true,
       message: 'ok',
       workflow: {
-        definition: makeTestWorkflow({
+        definition: makeTestResolvedWorkflow({
           name: 'test-workflow',
           interactive,
           ...(options.inputs ? { inputs: options.inputs } : {}),
@@ -2948,7 +2949,7 @@ describe('workflow dispatch routing — interactive flag', () => {
     // `willContinueExistingRun` was true, leaving `preparedSource` undefined and the
     // executor in `source_unprepared_live` mode against the prior run's frozen graph.
     const { prepareWorkflowSource } = await import('@archon/workflows/executor');
-    const freshWorkflow = makeTestWorkflow({
+    const freshWorkflow = makeTestResolvedWorkflow({
       name: 'test-workflow',
       description: 'freshly captured from disk',
     });
@@ -3632,7 +3633,7 @@ describe('workflow dispatch routing — interactive flag', () => {
 // ─── Natural-language approval routing ──────────────────────────────────────
 
 describe('paused approval gate routing', () => {
-  const approvalWorkflow = makeTestWorkflow({ name: 'prd', interactive: true });
+  const approvalWorkflow = makeTestResolvedWorkflow({ name: 'prd', interactive: true });
 
   type ManageRunHandler = (input: Record<string, unknown>) => Promise<string>;
 
@@ -4082,7 +4083,7 @@ describe('paused approval gate routing', () => {
 // ─── handleWorkflowRunCommand E2 path — single codebase auto-select ──────────
 
 describe('handleWorkflowRunCommand — E2 single codebase auto-select', () => {
-  const assistWorkflow = makeTestWorkflow({ name: 'assist' });
+  const assistWorkflow = makeTestResolvedWorkflow({ name: 'assist' });
 
   beforeEach(() => {
     mockGetOrCreateConversation.mockReset();
@@ -4412,7 +4413,7 @@ describe('handleWorkflowRunCommand — E2 single codebase auto-select', () => {
   });
 
   test('resolves workflow by case-insensitive name when exact match fails', async () => {
-    const upperWorkflow = makeTestWorkflow({ name: 'Assist' });
+    const upperWorkflow = makeTestResolvedWorkflow({ name: 'Assist' });
     const conversation = makeConversation({ codebase_id: null });
     const codebase = makeCodebaseForSync();
     mockGetOrCreateConversation.mockReturnValueOnce(Promise.resolve(conversation));
@@ -4449,7 +4450,7 @@ describe('handleWorkflowRunCommand — E2 single codebase auto-select', () => {
       Promise.resolve({
         success: true,
         message: 'Running workflow...',
-        workflow: { definition: makeTestWorkflow({ name: 'missing' }), args: 'test' },
+        workflow: { definition: makeTestResolvedWorkflow({ name: 'missing' }), args: 'test' },
       })
     );
     mockListCodebases.mockReturnValueOnce(Promise.resolve([codebase]));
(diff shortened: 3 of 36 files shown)


Wirasm:
# Review report

## Verdict

Ready — `none`. The correction closes the only remaining resolved-plan ownership finding, and the authoritative-plan regression remains proved.

Canonical PR comment: [#3081 review](https://github.com/coleam00/Archon/pull/3081#issuecomment-5483858316).

## Accepted contract

Split authored workflow definitions from include-expanded execution workflows. Include discovery must produce `ResolvedWorkflow`, with include-free `DagNode[]` nodes and an attached `GraphPlan` containing execution layers and sinks. The top-level executor and dry run must consume that attached plan rather than recomputing it. Resolution must reject a surviving include at the shared boundary and identify its node id. Propagate the resolved contract through discovery, CLI, core, and fixtures; replace production `as DagNode[]` narrowings at this branch point.

Boundaries: preserve YAML behavior, layer ordering, terminal output, and telemetry. Runtime-shaped loop and composed fan-out bodies receive fresh plans after deterministic substitution or instance-prefix application. Keep executor changes within the settled plan because #2698 separately changes terminal-failure blocks in `dag-executor.ts`. The contract is internal-only and adds no YAML surface.

## Reviewed head SHA

`355d551191ea847e5c5d0bab1ac39e28bd1c06f5`

## Findings

### Rejected findings

None.

### Suggestions

None.

## Prior findings

| ID | Sources | Prior status | Current verdict |
| --- | --- | --- | --- |
| R1 | `[seams]` | Open | Fixed. [workflow.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/schemas/workflow.ts:294) and [workflow.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/schemas/workflow.ts:308) now expose both node views as `readonly DagNode[]`; [dag-executor.test.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/dag-executor.test.ts:704) proves pushes, splices, and layer mutation are type errors. [graph-plan.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/graph-plan.ts:19) accepts the readonly node view while retaining the factory-owned plan. The invariant is restored: typed consumers cannot mutate the node set shared by the plan's snapshot layers and sinks. |
| R2 | `[tests]` | Fixed | Remains fixed. [dry-run.test.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/dry-run.test.ts:508) alters attached layers and sinks and verifies dry-run order and summary, while [dry-run.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/dry-run.ts:1280) and [dry-run.ts](/Users/rasmus/.archon/workspaces/coleam00/Archon/worktrees/refactor/2487-resolved-workflow/packages/workflows/src/dry-run.ts:1293) consume the attached plan. |

## Discoveries

0 accepted discoveries. [discoveries.json](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/d49f9142f2ed827a67197026cf42f881/discoveries.json) and [discoveries.md](/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/d49f9142f2ed827a67197026cf42f881/discoveries.md) contain the consolidated record. Adjacent discoveries do not affect readiness.

## Review coverage

One continuation reviewer verified the prior report and correction delta; specialists did not rerun. The original selected concerns were resolved-plan/node ownership and dry run honoring the attached plan. I inspected every changed production and test consumer in the correction delta, the planning factory and direct dry-run boundary, and the compile-time ownership regression. The documented commands run from `packages/workflows` all passed: `bun x tsc --noEmit`, `bun test src/dag-executor.test.ts` (706 tests), and `bun test src/dry-run.test.ts` (78 tests). `git diff --check` passed and there are no production `as DagNode[]` narrowings. No required evidence was unavailable.

PR #3120, opened 2026-09-01
fix(workflows): redact successful stderr broadcasts

### LINKED ISSUE #2989: fix(workflows): node stderr broadcast to chat bypasses credential redaction
## Problem

A bash or script node's stderr is broadcast live to the operator's chat surface without credential redaction. When a succeeding exec node leaves anything on stderr, `safeSendMessage` sends the raw string to the conversation (`packages/workflows/src/dag-executor.ts:3869-3875`), while the credential redaction added with retained exec output applies only to the stored transcript copy (the redaction seam around `dag-executor.ts:3602`). The two copies of the same stream get different treatment: the durable one is scrubbed, the one pushed to Slack/Telegram/chat in real time is not.

So a subprocess that echoes a token to stderr — a debug line, a verbose tool, a credential accidentally interpolated into a diagnostic — lands in chat verbatim. The retained-transcript work made the asymmetry visible; the broadcast path predates it.

## Evidence

Found during PR #2988's review and verified twice independently: the reviewer traced the path while proving the pack regression it enabled (gh diagnostics reaching chat on the no-CI flip), and the implementer reproduced it before fixing the pack side. The pack now keeps its own streams clean (PR #2988), but that is one workflow's discipline, not an engine guarantee — every other workflow and every user-authored node still hits the unredacted broadcast.

## Suggested direction

Run the broadcast through the same redaction the retained copy gets: the credential values are already collected once per subprocess (`collectSubprocessCredentialValues`), so the delta is applying them at the `safeSendMessage` call sites for node stdout/stderr messages, not new machinery. Worth checking the sibling call sites that message node output on failure paths for the same asymmetry while there.

## Desired outcome

No subprocess stream reaches a chat surface unredacted. The live broadcast and the retained transcript receive the same treatment from the same source of credential values, so the two copies of one stream cannot be scrubbed differently.

## Invariants

- **One redaction, not two.** The broadcast path applies the *same* redaction the retained copy already uses, from the same `collectSubprocessCredentialValues` result. A second implementation of "scrub credentials" is a hand-synced pair, and the two would drift exactly as these two copies already have.
- **Redaction never silently drops content.** A scrubbed line still reaches chat; only the credential substring is replaced. Suppressing the message instead would trade a credential leak for a silent failure.
- **No new machinery.** The values are already collected once per subprocess. This is applying them at call sites, not building a redaction layer.

## Scope: the sibling call sites are in

The Suggested direction says it is "worth checking the sibling call sites that message node output on failure paths for the same asymmetry while there." That is now settled rather than optional: `AGENTS.md` — *"fix drift on the path you touch rather than preserving it for a cleaner diff; two sites that should agree and do not are a defect, not a diff-size decision."*

If a failure-path broadcast has the same gap, it is fixed here. Leaving it produces a second issue for the identical defect on the identical path.

## Acceptance

- A bash or script node whose stderr contains a credential value broadcasts a redacted line to chat, on the success path and on every failure path that messages node output.
- The broadcast and the retained transcript produce identical redaction for the same stream.
- The redaction derives from `collectSubprocessCredentialValues`; no second scrubbing implementation is introduced.
- A message that contains no credential is unchanged.
- **Seen red:** the test is observed failing against the current unredacted broadcast. A redaction assertion that was never watched fail is a claim, and the defect here is precisely a path nobody was checking.

## Batching note

Touches `packages/workflows/src/dag-executor.ts`, the same file as #2933 and #2698. Run it in a different batch from those two — not because of dependency, but because three concurrent rewrites of one file produce conflicts that destroy each diff's readability.

Wirasm:
## Problem and outcome

Successful bash and script nodes could send credentials written to stderr directly to the operator chat, even though the retained transcript redacted the same stream. This made the live and durable copies of subprocess output disagree.

- **Outcome:** Successful stderr broadcasts now contain the same redacted content as their retained transcript rows.
- **Invariant:** Credential values are collected once and the same redacted stderr copy is used for retention and every operator-facing broadcast; raw stdout remains the node value channel.
- **Scope boundary:** This does not change stdout value semantics or add a separate redaction mechanism. Existing failure-path redaction remains at its shared rejection boundary.
- **Root cause:** On success, `runSubprocess` redacted only the retained stream tails and returned raw stderr to the callers that broadcast it.

## Review guidance

- **Feedback requested:** Confirm the shared success-path boundary preserves the single-redaction and raw-stdout invariants.
- **Start here:** `packages/workflows/src/dag-executor.ts:3412` — derives the redacted stderr used by broadcasts.
- **Review order:** Read the subprocess success return path, then the parameterized success and failure broadcast tests.
- **Lower-attention areas:** `scripts/check-test-cleanup-drift.ts` updates the recorded legacy cleanup count after this test suite adopts `removeTempTree`.
- **Known risk or uncertainty:** None. Full review completed across code, seams, simplify, and tests lenses — all four returned no findings. See the review comment on this PR for the complete report.

## Solution

`runSubprocess` now computes credential-redacted stderr once after a successful command. It retains capped redacted tails as before, returns raw stdout for node values, and returns that same redacted stderr to the existing bash and script message paths. Failed subprocesses continue to use the existing redacted rejection fields.

The regression tests run successful bash and script nodes that place a credential between unchanged stderr text, then compare both chat and transcript output with the expected redaction. Matching failure cases prove the sibling error broadcasts remain redacted.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Successful stderr broadcasts could include a credential verbatim. | Successful stderr broadcasts replace the credential with `[REDACTED]` while preserving surrounding output. |
| **Failure behavior** | Failure broadcasts already used redacted rejection data. | Failure broadcasts remain redacted and are explicitly covered for bash and script nodes. |

## Validation

- `bun test src/dag-executor.test.ts --test-name-pattern "subprocess credential redaction"` (from `packages/workflows`) — 7 pass / 0 fail — covers successful and failed bash/script broadcasts plus retained output.
- `bun run validate` (repository root) — exit 0. Covers generated-file checks, API type check, per-package type-check, lint (`--max-warnings 0`), format check, installer tests, package tests, and script tests.
- CI (`gh pr checks 3120`) — 9 checks, all pass (Test Suite, docker-build, test ubuntu/windows, schema-upgrade, postgres-parity, changes, workflow-fixtures, CodeRabbit).
- **Not verified:** Nothing material; the regression was observed failing on the unmodified success broadcast before this change, then passing afterward.

## Links

- Closes #2989

### DIFF
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -25382,7 +25382,7 @@ describe('subprocess credential redaction', () => {
   });
 
   afterEach(async () => {
-    await rm(testDir, { recursive: true, force: true });
+    await removeTempTree(testDir);
   });
 
   it('removes exact injected credentials from every rejection field before persistence', async () => {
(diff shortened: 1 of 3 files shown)


Wirasm:
# Review report: PR #3120 (round 4)

## Verdict

**Ready — action: `none`.** Continuation review verifies the prior round's verdict stands. All four required lenses (code, seams, simplify, tests) have now completed and report no Critical or Important findings. The correction delta is empty (SHA unchanged); the change converged and satisfies issue #2989's accepted contract. Two Suggestions are recorded below; neither blocks readiness.

## Accepted contract

Successful bash and script subprocess stderr broadcasts must carry the same credential-redacted stderr as their retained transcript rows. Credential values are collected once per subprocess through `collectSubprocessCredentialValues`; the same redacted stderr copy feeds retention and every operator-facing broadcast. Sibling failure broadcasts remain redacted at the shared rejection boundary (`redactSubprocessError`). Raw stdout remains the workflow node value channel. Redaction replaces only the credential substring with `[REDACTED]`; unprotected text is unchanged and no message is suppressed. No new redaction mechanism is added.

## Reviewed head SHA

`93ddb99b7b1bb9dc82970e6d15f34a74eed5b3d2`

## Findings

### Suggestions

#### R1 – `redactedStderr` variable and manual return object can be simplified (sources: [simplify])

**Claim:** The success path of `runSubprocess` (`dag-executor.ts:3412-3422`) introduces a named intermediate variable (`redactedStderr`) and constructs a manual `return { stdout: result.stdout, stderr: redactedStderr }` object, when mutating `result.stderr` in place would eliminate both without changing behavior. The catch path at line 3424 already uses in-place mutation via `redactSubprocessError`, demonstrating the same pattern in the same function.

**Current code:**
```ts
const redactedStderr = redactCredentialValues(result.stderr, credentialValues);
// ...
return { stdout: result.stdout, stderr: redactedStderr };
```

**Suggested replacement:**
```ts
result.stderr = redactCredentialValues(result.stderr, credentialValues);
// ...
return result;
```

**Why it is a Suggestion, not a blocker:** The behavior is identical either way. Every caller destructures `{ stdout, stderr }` from the return value; the three `until_bash` callers discard the return entirely. The current code is correct, maintains the invariant, and passes all tests. This is a code organization refinement, not a defect.

**Severity:** Suggestion. **Status:** open. **Round:** 4.

#### R2 – Redact-before-truncate order only tested with short stderr (sources: [tests])

**Claim:** All three new tests use stderr short enough that `retainStreamTail` (last ~4 KB) returns the entire string, so the truncation boundary is never exercised. A future edit that truncates before redacting — believing it saves work — would pass every short-output test because truncation is a no-op for short streams.

**Invariant:** Redaction happens before truncation, so credentials are always scrubbed even if they fall outside the retention tail window. Broadcast always sends the full redacted stderr, so credential leakage does not occur from a reorder — the defect would be a partial retention miss.

**Smallest valuable test:** A workflow node emitting >4 KB stderr with a credential appearing only before the retention tail window. Assert the transcript `stderr_tail` contains `[REDACTED]` at the credential position, not the raw value.

**Severity:** Suggestion. **Status:** open. **Round:** 4.

### Rejected findings

The simplify lens labeled R1 as Important. I reclassify it as a Suggestion because it does not violate any invariant in the accepted contract — the simplification is a code organization refinement, not a correctness or safety defect. The behavior is identical either way.

### Prior findings

The prior report (round 3 at `93ddb99b7`) had no Critical or Important findings and an empty `findings.json`. This round confirms that:

- **code lens** (round 3: clean) — reconfirmed. The `runSubprocess` success path derives one `redactedStderr` and reuses it for retention and broadcast.
- **seams lens** (round 3: clean, one adjacent discovery) — reconfirmed. Every seam verified clean: success broadcast callers, failure `redactSubprocessError`, `safeSendMessage` no-reprocessing, container/host path consistency, and call-site type safety.
- **simplify lens** (round 3: claimed clean but was incomplete) — now complete. One code-organization Suggestion (R1 above). The remaining structure — one credential collection, inline stdout redaction, shared redacted stderr — is minimal.
- **tests lens** (round 3: claimed clean but was incomplete) — now complete. All 7 tests pass (verified this round). One testing-completeness Suggestion (R2 above).

No prior finding was carried forward as open. No prior finding was disproved.

## Discoveries

One accepted adjacent discovery: **Hand-declared subprocess rejection types duplicate Node's ExecFileException** (seams lens). `RawSubprocessRejection` (`dag-executor.ts:3262`) and `RawSubprocessError` (`executor-shared.ts:230`) re-state Node's `child_process.ExecFileException` fields instead of importing the owning type. Does not violate the redaction contract; the PR touches neither declaration.

See `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/c45b09f4bf04addbca212f777948d14e/discoveries.json` and `discoveries.md`.

**If you are an agent reading this: open discoveries.md and surface each discovery to your human.**

## Review coverage

Continuation review. No correction delta existed (SHA unchanged from prior reviewed SHA). The review verified the prior report's findings against the current code and the specialist lens reports. Two lenses from round 3 that were incomplete (simplify, tests) are now complete; code and seams are reconfirmed.

| Concern | Result |
| --- | --- |
| code (reconfirmed) | No findings |
| seams (reconfirmed) | No in-scope findings; one adjacent discovery |
| simplify (now complete) | One Suggestion (R1) |
| tests (now complete) | One Suggestion (R2) |

`errors` and `docs` remain correctly disabled for this delta: the diff adds no new catch/fallback/retry/error-translation code and changes no shipped documentation.

### Decisive verification

Read and confirmed against the code:

- `runSubprocess` derives `redactedStderr` once (`dag-executor.ts:3412`) and reuses it for retention (`:3419`) and the returned stderr (`:3422`); stdout is returned raw.
- Both success broadcasts (`:3683-3690`, `:4080-4087`) send that redacted stderr; the sibling failure aggregation (`:11942`) reads the `redactSubprocessError`-scrubbed fields through `formatSubprocessFailure`.
- Redaction is a pure `replaceAll` against the once-collected `credentialValues` (`:3385`).
- All 7 `subprocess credential redaction` tests pass (verified this round via `bun test`).

PR #3129, opened 2026-09-01
fix(core): take the next free port when a worktree's hashed port is occupied

alixnaveed:
## Problem and outcome

A worktree's server port is an MD5 hash of its path folded into 900 slots and returned without checking whether anything is listening on it. Two worktrees hashing to the same slot — or any unrelated process holding a port in 3190–4089 — makes the second server fail to bind with a bare `EADDRINUSE` that names neither the cause nor the `PORT` override that resolves it. Parallel worktrees are the workflow the README leads with, and 900 slots is a small space: collisions reach roughly even odds around 35 worktrees, and about 5% at ten.

- **Outcome:** A worktree whose hashed port is taken starts on the next free port in the range and logs `worktree_port_reallocated` with both the port it wanted and the one it got, instead of failing to start.
- **Invariant:** The hashed port is still what a worktree gets whenever it is free — "same worktree, same port" survives — and an explicit `PORT` is still bound verbatim or not at all.
- **Scope boundary:** The base port (3090) is deliberately still unprobed, and the Vite dev proxy is untouched. See *Known risk or uncertainty*.
- **Root cause:** `getPort()` computed a port and returned it. There was no availability check anywhere between the hash and `Bun.serve`.

## Review guidance

- **Feedback requested:** Correctness of the scan, and whether the base-port carve-out is the call you want.
- **Start here:** `packages/core/src/utils/port-allocation.ts` — `resolveWorktreePort()`. Everything else follows from it.
- **Review order:** `port-allocation.ts` (`isPortAvailable` → `getPort` → `resolveWorktreePort`), then `port-allocation.test.ts`, then the `dx-quirks.md` note.
- **Lower-attention areas:** The pre-existing `calculatePortOffset` tests are unchanged; the magic numbers they assert are now named constants (`BASE_PORT`, `MIN_OFFSET`, `OFFSET_SPAN`) with identical values.
- **Known risk or uncertainty:** Two, both deliberate and documented at the branch:

  1. **The probe is racy by construction.** Another process can take the port between `isPortAvailable()` and `Bun.serve`. Closing that window means holding the socket from allocation through to the server, which this function does not own. The probe removes the common case — a port already held when the server starts — and leaves the narrow one to the bind error, which is no worse than today.
  2. **The base port is not probed, on purpose.** Unlike a worktree port, 3090 is resolved independently elsewhere: `packages/web/vite.config.ts` falls back to it verbatim when `PORT` is unset. Silently serving from somewhere else would leave the dev proxy pointing at nothing. I think that carve-out is right, but it is the judgment call most worth a second opinion.

## Solution

`getPort()` gains one step on the worktree path: probe the hashed port, and if it is taken, walk forward through the 900-slot range (wrapping at 999 → 100) to the first free one.

Scanning forward from the hashed offset — rather than re-hashing or picking randomly — keeps the result stable for a given occupancy, so a worktree that lands on its second choice lands there again next time rather than reshuffling.

`isPortAvailable()` binds and immediately closes rather than consulting a list, so the answer reflects what `Bun.serve` will actually be allowed to do: a port held by another user or blocked by local policy reads as unavailable here for the same reason it would fail at startup. It probes on `process.env.HOST || '0.0.0.0'`, mirroring the server's own bind address, because availability is per-interface — a process bound to `127.0.0.1` does not block `0.0.0.0` on every platform.

Two paths deliberately keep their existing behavior:

- **Explicit `PORT`** returns unprobed. An explicit port is an instruction, not a preference; moving the operator off the port they named would break whatever they pointed at it, and a bind failure is the honest outcome.
- **The base port** returns unprobed, for the `vite.config.ts` reason above.

Range exhaustion — all 900 taken — fails loudly with a message naming `PORT`, rather than falling through to an arbitrary port.

`resolveWorktreePort` is exported for the same reason `calculatePortOffset` already is. My first draft tested the collision path through `getPort()`, which only reaches the worktree branch when the test process is itself running inside a worktree — so in the main repo, and therefore in CI, the test would have skipped its own assertion every single run and still reported green. Driving the resolver directly is what makes the fallback actually execute; the mutation check under Validation is what confirms it.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Worktree gets `3090 + hash(path)`, always | Worktree gets `3090 + hash(path)` when free; the next free port in 3190–4089 otherwise |
| **Failure behavior** | Server exits with a bare `EADDRINUSE` naming neither the cause nor `PORT` | Server starts on a neighbouring port and logs `worktree_port_reallocated`; only a fully exhausted range fails, with a message naming `PORT` |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[Start server in worktree] --> B2["hash(path) → 3516"]
    B2 --> B3{3516 free?}
    B3 -- yes --> B4[Listening on 3516]
    B3 -- no --> B5[EADDRINUSE, server exits]
  end

  subgraph After
    A1[Start server in worktree] --> A2["hash(path) → 3516"]
    A2 --> A3{3516 free?}
    A3 -- yes --> A4["Listening on 3516<br/>worktree_port_allocated"]
    A3 -- no --> A5[Scan forward in 3190-4089]
    A5 --> A6["Listening on 3519<br/>worktree_port_reallocated"]
    A5 -- all 900 taken --> A7["Fatal, names PORT"]
  end
```

## Validation

- `bun test src/utils/port-allocation.test.ts` — **16 pass, 0 fail** — proves the hashed port is still returned when free, the scan advances past occupied ports, it wraps 999 → 100, it returns to the hashed port once released, and an explicit `PORT` is returned even when occupied.

- **Mutation check** — disabling the probe (`if (false && !(await isPortAvailable(...)))`) turns exactly the four collision tests red and leaves the other twelve green; restoring it returns all sixteen to green. This is the evidence that the new tests actually constrain the new behavior rather than passing alongside it.

- **Runtime evidence** from the test logs — the fallback genuinely executes, rather than the assertions being satisfied some other way:

  ```
  preferredOffset:426 → offset:427, attempt:1, port:3517   worktree_port_reallocated
  preferredOffset:426 → offset:429, attempt:3, port:3519   worktree_port_reallocated
  preferredOffset:999 → offset:100, attempt:1, port:3190   worktree_port_reallocated   (wrap)
  ```

- `bun run test` in `@archon/core` — **exit 0**, 44 test files, zero failures — proves nothing else in the package regressed.
- `bun --filter @archon/core type-check` — exit 0.
- `bun run lint --max-warnings 0` (whole repo) — exit 0. It caught two real `no-confusing-void-expression` errors in the first draft of `isPortAvailable`; both are fixed.
- `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix` — all exit 0.

**Not verified:**

- `bun run validate` does not complete on my machine: `@archon/docs-web type-check` exits 1 with *"Node.js v20.12.2 is not supported by Astro! Please upgrade Node.js to >=22.12.0"*. I confirmed this reproduces on a clean stash of `dev`, so it is environmental and not caused by this change — but it does mean the aggregate `validate` gate is unproven here, and CI is its first real run.
- `@archon/cli` has two failing tests on my machine (`CLI startup import boundary`, and a caller-relative path test). Both reproduce on a clean stash of `dev` and both fail at ~5000ms, which looks like a Windows timeout rather than a real assertion failure. Pre-existing either way.
- One `@archon/providers` test (`PiProvider > sendQuery installs PI_PACKAGE_DIR shim`) failed once at 7392ms during the full `--parallel` run, but passes in isolation both with and without this change. Load-induced flake under parallel CPU contention, not a regression.
- Bun **1.4.0** locally; CI pins **1.3.11**. Within the `^1.3.0` engines range, but not the same runtime CI will use.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. The hashed port is unchanged whenever it is free, so every non-colliding setup behaves identically. Nothing persists or re-derives the port. | `grep` for `getPort`/`calculatePortOffset` finds one consumer, `packages/server/src/index.ts:688` |
| Observability | A moved port is visible as `worktree_port_reallocated` carrying `preferredOffset`, the chosen `offset`, and `attempt`; the existing `server_listening` line still reports the bound port | `port-allocation.ts`, `packages/server/src/index.ts:933` |
| Documentation / communication | `dx-quirks.md` documented "same worktree always gets same port" with no collision case; it now describes the fallback and the `PORT` guarantee | `packages/docs-web/src/content/docs/contributing/dx-quirks.md` |

---

## Adjacent finding, not fixed here

While tracing consumers I found what looks like a separate pre-existing bug, left untouched to keep this PR focused.

In a worktree with no `PORT` set, the server binds `3090 + offset` while the Vite dev proxy targets `3090` — `packages/web/vite.config.ts:11` reads `env.PORT ?? '3090'` and does not compute the worktree offset. So `bun dev` in a worktree appears to leave the web UI proxying to a port nothing is serving, unless the user sets `PORT` explicitly.

I have not reproduced this end-to-end (no Bun locally), so I am reporting it rather than claiming it. Happy to open a separate issue if it is not already known.

## Summary by CodeRabbit

- **New Features**
  - Worktree servers now automatically select the next available port when the preferred port is occupied.
  - Port selection stays within the supported range and returns to the preferred port when it becomes available.
  - Explicit `PORT` settings remain fixed and report an error if unavailable.

- **Documentation**
  - Added guidance describing worktree port collision handling and port assignment logging.

### DIFF
--- a/packages/core/src/utils/port-allocation.test.ts
+++ b/packages/core/src/utils/port-allocation.test.ts
@@ -1,5 +1,27 @@
 import { describe, it, expect, afterEach } from 'bun:test';
-import { calculatePortOffset, getPort } from './port-allocation';
+import { createServer, type Server } from 'node:net';
+import {
+  calculatePortOffset,
+  getPort,
+  isPortAvailable,
+  resolveWorktreePort,
+} from './port-allocation';
+
+const BASE_PORT = 3090;
+const HOSTNAME = '0.0.0.0';
+
+/** Hold a real listener so availability probes see a genuinely occupied port. */
+function occupy(port: number): Promise<Server> {
+  return new Promise((resolve, reject) => {
+    const server = createServer();
+    server.once('error', reject);
+    server.listen(port, HOSTNAME, () => resolve(server));
+  });
+}
+
+function release(server: Server): Promise<void> {
+  return new Promise(resolve => server.close(() => resolve()));
+}
 
 // Test the exported hash calculation function directly
 describe('calculatePortOffset', () => {
@@ -44,16 +66,56 @@ describe('calculatePortOffset', () => {
   });
 });
 
+describe('isPortAvailable', () => {
+  const held: Server[] = [];
+
+  afterEach(async () => {
+    await Promise.all(held.splice(0).map(release));
+  });
+
+  it('reports an unbound port as available', async () => {
+    // Port 0 asks the OS for any free port; bind it, learn the number, release it.
+    const scout = await occupy(0);
+    const address = scout.address();
+    const port = typeof address === 'object' && address ? address.port : 0;
+    await release(scout);
+
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(true);
+  });
+
+  it('reports a port held by a live listener as unavailable', async () => {
+    const server = await occupy(0);
+    held.push(server);
+    const address = server.address();
+    const port = typeof address === 'object' && address ? address.port : 0;
+
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(false);
+  });
+
+  it('leaves the probed port bindable afterwards', async () => {
+    const scout = await occupy(0);
+    const address = scout.address();
+    const port = typeof address === 'object' && address ? address.port : 0;
+    await release(scout);
+
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(true);
+    // The probe must not leak its own listener, or the second call would fail.
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(true);
+  });
+});
+
 // Test getPort() behavior with mocked dependencies
 describe('getPort', () => {
   const originalEnv = process.env.PORT;
+  const held: Server[] = [];
 
-  afterEach(() => {
+  afterEach(async () => {
     if (originalEnv === undefined) {
       delete process.env.PORT;
     } else {
       process.env.PORT = originalEnv;
     }
+    await Promise.all(held.splice(0).map(release));
   });
 
   it('should return PORT env var when explicitly set to valid number', async () => {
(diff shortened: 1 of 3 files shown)


Wirasm:
Thanks for the detailed contribution. The work is focused and the reported collision case is worth tracking. I opened #3130 to capture the problem and desired outcome without prescribing this PR's implementation.

Please reopen this PR and add `Closes #3130` to its description. Keep the separate Vite proxy behavior mentioned in the description out of this PR; it should be tracked independently if confirmed.

alixnaveed:
Thanks — and apologies for the missing link. I dropped the `Closes #` line when I opened this rather than filing the issue first, which was the wrong way round.

I tried to reopen as you asked, but GitHub refuses: `reopenPullRequest` fails via GraphQL and the REST `PATCH state=open` returns 422. So I've opened #3177 from the same branch instead — identical commit, rebased onto current `dev` — with `Closes #3130` in the description and the Vite proxy section removed.

I'd also filed #3176 before I read your full comment; that's a duplicate of #3130 and I've closed it. Sorry for the noise.

On #3130's acceptance criteria: the Vite observation is untracked for now. I haven't confirmed it, so I'd rather verify it first than file a guess — happy to open it separately once I have.

PR #3133, opened 2026-09-02
Keep SQLite upgrade-path test fixtures out of the package directory

Wirasm:
## Problem and outcome

Every run of the `@archon/core` suite left two files named `file:archon-test-sqlite-*?mode=memory&cache=shared` under `packages/core/`. The SQLite upgrade-path fixtures opened those names expecting a named in-memory database, but `bun:sqlite` opens without `SQLITE_OPEN_URI`, so SQLite read each URI as a literal filename and created it in the process cwd. The fixtures only ever worked because they were real files.

- **Outcome:** The two upgrade-path fixtures use a real database file under a tracked temp root that is removed after each test. Nothing is written into the source tree.
- **Invariant:** Both regression tests keep proving the same thing, on the real adapter: a database missing `event_order` converges on open, and the `outcome` column is added idempotently across reopen.
- **Scope boundary:** The adapter is unchanged. It has no URI requirement, so nothing gains `SQLITE_OPEN_URI`.
- **Root cause:** `new Database(uri)` in both the adapter and the test's raw connection, with no URI flag.

## Review guidance

- **Feedback requested:** Whether a file-backed fixture is the right replacement, or whether the adapter should learn URI opening.
- **Start here:** `packages/core/src/db/adapters/sqlite.test.ts` — `upgradeFixturePath` and the comment above it.
- **Review order:** The two fixture helpers, then the two tests that consume them.
- **Lower-attention areas:** The tests drop their held-open "seed" connection because a file does not need one to survive reopen.

## Solution

Each fixture gets `<mkdtemp>/archon.db` from `trackTempRoots`, the suite's standard cleanup primitive. The seed adapter writes the current schema and closes; the raw connection strips the column; the assertions reopen the same file. A file-backed fixture is the honest version of what the URI was already doing by accident.

## Validation

- `cd packages/core && bun test src/db/adapters/sqlite.test.ts` — 28 pass — and `git status` shows no stray files afterwards, where before the run left two.
- `bun run lint` — the test-cleanup drift check passes.
- **Not verified:** Nothing material.

## Links

- Related #3132 (found while validating it)

https://claude.ai/code/session_01Nv3rzN6mfBHrmHrtHJ4Vq8

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nv3rzN6mfBHrmHrtHJ4Vq8)_

## Summary by CodeRabbit

* **Tests**
  * Updated database upgrade tests to use temporary on-disk database files instead of shared in-memory databases.
  * Simplified test fixture setup and cleanup for more reliable upgrade-path testing.

### DIFF
--- a/packages/core/src/db/adapters/sqlite.test.ts
+++ b/packages/core/src/db/adapters/sqlite.test.ts
@@ -5,6 +5,9 @@ import { APP_VERSION, readSchemaVersion } from '../schema-version';
 import { Database, type Statement } from 'bun:sqlite';
 import { unlinkSync } from 'fs';
 import { join } from 'path';
+import { tmpdir } from 'os';
+import { mkdtemp } from 'fs/promises';
+import { trackTempRoots } from '@archon/paths/test-utils';
 
 let currentDbPath = '';
 
@@ -25,6 +28,22 @@ async function insertCodebase(db: SqliteAdapter, id: string): Promise<void> {
   ]);
 }
 
+const trackTempRoot = trackTempRoots();
+
+/**
+ * A fresh on-disk database path for one upgrade fixture, reclaimed after the test.
+ *
+ * A real file, not a `file:...?mode=memory&cache=shared` URI: bun:sqlite opens without
+ * `SQLITE_OPEN_URI`, so SQLite reads such a URI as a literal filename and creates it in
+ * the process cwd — which is how these fixtures used to leave `file:archon-test-sqlite-*`
+ * files under `packages/core/`. The adapter has no URI requirement, and the fixture only
+ * needs a database that survives being reopened.
+ */
+async function upgradeFixturePath(): Promise<string> {
+  const root = trackTempRoot(await mkdtemp(join(tmpdir(), 'archon-core-sqlite-upgrade-')));
+  return join(root, 'archon.db');
+}
+
 /**
  * Produce a database in the state it had BEFORE event_order existed: current
  * schema in every other respect, with the column, its index and its trigger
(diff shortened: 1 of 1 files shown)


Wirasm:
## Ready to merge

`bun:sqlite` opens without `SQLITE_OPEN_URI`, so the old `file:archon-test-sqlite-*?mode=memory&cache=shared` fixture strings were never read as in-memory-database URIs — SQLite treated them as literal filenames and created real files under `packages/core/` on every test run. This PR replaces those fixtures with real on-disk database files under `trackTempRoots`, the project's standard test-cleanup primitive, and drops the now-dead "seed connection" keepalive that only mattered for the old in-memory fixture. The change is test-only (one file, `packages/core/src/db/adapters/sqlite.test.ts`), all three reviewers independently verified it by running the suite and confirming no stray files remain, and the code reviewer reproduced the original bug directly (on Linux, matching the CI platform) to confirm the PR's root-cause claim is accurate rather than assumed.

**Validation:** All commands passed.

### Findings

No findings.

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | No findings |
| seams | No findings |
| simplify | No findings |

#### Validation

|---|---|---|
| `cd packages/core && bun test src/db/adapters/sqlite.test.ts` | PASS | 28 pass, 0 fail, 75 expect() calls. `git status --porcelain` clean afterward — no stray `file:archon-test-sqlite-*` files under `packages/core/`. |
| `bun run lint` (repo root; includes `scripts/check-test-cleanup-drift.ts`) | PASS | "Test-cleanup drift check passed." Full ESLint pass across all packages, exit 0. |
| Reviewer-run: `bun test src/db/migration-statement-order.test.ts src/db/adapters/sqlite.test.ts src/db/adapters/sqlite-vintages.test.ts` (code scope) | PASS | 35 pass, 0 fail |
| Reviewer-run: `bun run test -- src/db/adapters/sqlite.test.ts` (seams scope, full package suite) | PASS | 84 pass, 0 fail |
| Reviewer-run: bug reproduction in `oven/bun:1.4.0` (Linux, matching CI) (code scope) | CONFIRMED | Old `file:...?mode=memory&cache=shared` string created a literal on-disk file; did not reproduce on macOS/Bun 1.3.11, but CI's `test` job runs only `ubuntu-latest`/`windows-latest`, so the reproduction platform matches the one that matters. |
| Repository-wide seam search for the same URI-as-filename defect class | CLOSED | `grep` across `packages/` for `mode=memory`, `cache=shared`, `file:` prefixes, and every `new Database(...)`/`new SqliteAdapter(...)` call site. Only the fixed file matched; every other call site (`sqlite-vintages.test.ts`, production `sqlite.ts`, CLI integration specs) already uses a real file path. No sibling instance of the bug exists. |

</details>

Issue #3010, opened 2026-08-30
A named script: check node executes bytes the run did not freeze, and evidence_policy proves nothing


Wirasm:
## Summary

- **What broke:** Two shipped mechanisms that look like independent verification are not. (1) A named `script:` node executes bytes the run did not freeze, because the run's captured source lives inside the directory the engine hands to every AI node. (2) `evidence_policy.required` gates on a file's presence at a path the agent writes, so a node that verified nothing can satisfy it.
- **When it started:** Not a regression. Both are the shipped design.
- **Severity:** `major`

## The first one

`getRunSourceCapturePath()` returns `join(artifactsDir, 'workflow-source')` — `packages/workflows/src/workflow-source.ts:708-709`.

A run freezes its own `.archon/{workflows,commands,scripts}` at start so that resume executes what the run began with (#2660). But the freeze is stored **inside `$ARTIFACTS_DIR`**, whose absolute path is substituted into every AI node's prompt, and which `archon workflow get` lists as the run's artifacts. Nothing re-verifies the capture before a later `script:` node runs.

Observed directly, with an honest prompt that explicitly instructed the agent not to modify anything: the capture was both visible and writable from the agent's seat. A tamper node followed by two sibling check nodes then produced opposite verdicts from the same intent:

```
ok  tamper         (rewrites the captured check script)
ok  check_script   [verify] CHECK_ID=verify-TAMPERED     VERDICT=pass
ok  check_inline   [inline] INLINE_ID=inline-v1-original VERDICT=fail
Workflow completed successfully.
```

An inline `bash:` body is unaffected — the graph is held in engine memory after load. The asymmetry between `bash:` and `script:` is undocumented, and it runs the wrong way: the packaged, reviewed, version-controlled form is the weaker one.

**Not a security claim.** With no OS sandbox the agent runs as the same uid and can reach `ARCHON_HOME` regardless. The property worth having is narrower and still worth having: *the bytes a run executed are the bytes it froze*, so the run record can say so. Today it cannot.

## The second one

`evidence_policy.required` (#2230) refuses terminal success when `evidence.json` is absent. It checks presence, at a path handed to the agent. A node that verified nothing wrote:

```json
{ "verified": true, "checks": ["nothing was actually run"] }
```

and the run completed. That makes it weaker than an ordinary node, whose verdict is at least its own exit status.

## Expected vs Actual

- **Expected:** a check node's result reflects code the run controls, and a run-level success gate reflects something that happened.
[...]

PR #3134, opened 2026-09-02
Mount the run's artifacts directory into container runs

Wirasm:
## Problem and outcome

`$ARTIFACTS_DIR` is the run's output channel: screenshots, reports, the `evidence_policy` marker, anything a workflow leaves for humans and machines to inspect. A `--container` run received that path in its environment but not the directory behind it, so an in-container node writing there either failed or wrote into a container-local directory the host engine and operator never saw. The docs told authors to avoid `$ARTIFACTS_DIR` in containers, which documented a defect rather than a contract.

- **Outcome:** A container run binds the run's artifacts directory read-write at the same absolute path, beside the existing read-only source bind. A node writes to `$ARTIFACTS_DIR` the same way on either side of the boundary, and the evidence marker works from inside a container.
- **Invariant:** The container can write exactly one host directory, the run's own artifacts directory, and the host treats its contents as run output, never as code. Both host binds are validated the same way: absolute, inside `ARCHON_HOME`, not overlapping the workspace root.
- **Scope boundary:** No YAML or config surface. The engine's `$ARTIFACTS_DIR` delivery is unchanged; only the backend and the one container dispatch site in the CLI change. Stacked on #3132 so the source bind and the artifacts bind are siblings rather than nested.

## Review guidance

- **Feedback requested:** The ownership hand-back design (below), and whether same-path is the right choice for the artifacts bind. Same-path was chosen because `$ARTIFACTS_DIR` already carries the host path into every node and the source bind already relies on the same invariant.
- **Start here:** `packages/isolation/src/backends/container.ts` — `HostMounts`, `assertMountableHostMounts`, and `restoreHostOwnership`.
- **Review order:** `types.ts` (`artifactsMount` on the prepare request) → `container.ts` (prepare, metadata, resume recreation, `docker run` args, suspend and destroy hooks) → `packages/cli/src/commands/workflow.ts` (the dispatch site creates the directory as the host user, then passes it) → tests → docs.
- **Lower-attention areas:** The `sourceMount` validation is generalized to both mounts with the same three error messages; the existing source-mount tests still pass unchanged.
- **Known risk or uncertainty:** File ownership on a rootful Linux daemon. In-container work runs as root, so files it writes into the host bind are root-owned there, and the host user could not later remove them. Docker Desktop and rootless daemons map container root to the host user already. The backend now `chown`s the artifacts mount back to the host uid/gid on suspend and on destroy, best-effort and logged (`isolation.container_artifacts_chown_failed`), skipped when the host user is root or the platform has no uid. Under userns-remap the chown target maps to a subordinate uid; that case is logged rather than fixed.

## Solution

The backend already persisted and re-applied one Archon-owned host bind, the frozen source. This adds a second through the same path: `BackendPrepareRequest.artifactsMount`, validated by the same guard, written into the same `docker run` args (read-write, no `:ro`), persisted in the environment metadata, and re-applied when resume recreates the container. The CLI's container dispatch resolves the run's artifacts directory with the executor's own path resolver and creates it before `prepare`, because a bind target Docker creates itself is root-owned. Ownership hand-back is a private helper called from `suspend` before `docker stop`, and from `destroy` when the container is still running.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Inside a container, `$ARTIFACTS_DIR` named a host path that did not exist; writes failed or stayed container-local. `evidence_policy.required` could not be satisfied from a container node. | `$ARTIFACTS_DIR` is the host directory on both sides. Outputs and the evidence marker land where the engine and operator read them. |
| **Failure behavior** | Silent loss of container-side artifacts. | An out-of-tree or workspace-overlapping artifacts path refuses the run before any Docker work, like the source mount. A failed ownership hand-back is logged and never fails the run. |

## Validation

- `cd packages/isolation && bun test src/backends/container.test.ts` — 38 pass — the eight new cases prove: read-write bind at the same path beside the read-only source bind; mount persisted in metadata; mount re-applied on resume recreation; out-of-tree path refused before any Docker call; chown runs before `docker stop` on suspend; chown runs before `docker rm` on destroy of a running container; chown skipped for a root host user; a failing chown is logged and suspend still stops the container.
- `cd packages/cli && bun test src/commands/workflow.test.ts` — 395 pass — the new case proves the dispatch passes both mounts and that the artifacts directory exists on the host at the moment `prepare` is called.
- `bun run type-check`, `bun run lint`, `bun run format:check` — pass.
- **Not verified:** A real `--container` run against a Docker daemon; none is available in the authoring environment. Needed on a machine with Docker: a folder-project run with a `bash:` node that writes `$ARTIFACTS_DIR/probe.txt` and `evidence_policy: { required: true }`, showing the file on the host and the run completing; the same after removing the container and resuming; and on a rootful Linux daemon, `ls -l` of the artifacts directory after the run showing host ownership.

## Delivery considerations

|---|---|---|
| Security / permissions / data | The container gains one writable host directory, the run's own artifacts. The host reads it as data only; the source bind stays read-only and is no longer under it. Documented in the isolation security notes. | `packages/isolation/docker/SECURITY.md` |
| Compatibility / migration | Environments created before this change have no `artifactsMount` in their metadata and resume exactly as before. New environments persist it. | `container.test.ts` resume cases |
| Documentation / communication | Container guide, CLI reference, variables reference, and the isolation security notes no longer tell authors to avoid `$ARTIFACTS_DIR` in containers. #3131 drops its matching caveats when it rebases. | The four docs in the diff |

## Links

- Depends on #3132
- Related #3131, #3010

https://claude.ai/code/session_01Nv3rzN6mfBHrmHrtHJ4Vq8

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nv3rzN6mfBHrmHrtHJ4Vq8)_

## Summary by CodeRabbit

* **New Features**
  * Container-isolated runs now support read/write access to the run artifacts directory at the same path as on the host.
  * Generated artifacts are preserved on the host and ownership is restored when environments are suspended or destroyed.

* **Documentation**
  * Updated container isolation, CLI, variable, and security documentation to describe artifact mounting, output handling, and state directory behavior.

* **Tests**
  * Added coverage for artifact mounting, persistence, path validation, and ownership restoration.


Wirasm:
## Needs fixes

This PR does what it says: a `--container` run now binds the run's artifacts directory read-write at the same absolute path the engine already uses for `$ARTIFACTS_DIR`, beside the existing read-only frozen-source bind. I reproduced the fix live against a real Docker daemon (Docker Desktop for Mac, arm64) — a folder-project `--container` run with `evidence_policy: { required: true }` and a `bash:` node writing `$ARTIFACTS_DIR/probe.txt` now completes successfully, and the file lands on the host with correct content. That closes the exact defect the PR describes. The mount guard (`assertMountableHostPath`) is applied symmetrically to both mounts, and the frozen source stays a sibling directory outside the writable artifacts tree, so the write access does not reopen the tamper path #3132 just closed.

Two findings sit on the same theme — the ownership hand-back and mount re-application are correct for the common case but silently incomplete on two edge branches — and both are cheap to close: a missing log line, and a missing distinction between "this row has no artifacts mount" and "this row predates the feature." Three more are minor structural/documentation observations.

**2 blocking · 3 non-blocking**

**Validation:** all cited checks pass, plus a live Docker end-to-end run I added.

### Findings

|---|---|---|---|
| `R1` | Important | `destroy()`'s ownership hand-back silently skips on an ambiguous (`'unknown'`) container-state read, with no log line — worse than the feature's own documented best-effort contract | OPEN |
| `R2` | Important | `resumeEnv`'s recreate branch silently drops the artifacts bind for container envs created before this PR, once resumed after an Archon upgrade | OPEN |
| `R3` | Suggestion | `HostIdentity`/`currentHostIdentity()` widens `ContainerBackendDeps` for a test-only need; a lighter precedent already exists in this codebase | OPEN |
| `R4` | Suggestion | CLI dispatch site checks `artifactsMount !== undefined` as a second, redundant condition after `preparedSource` already gates it | OPEN |
| `R5` | Suggestion | `variables.md`'s `$STATE_DIR` row still says "same caveat as `$ARTIFACTS_DIR`," but this PR just removed that caveat from the `$ARTIFACTS_DIR` row above it | OPEN |

<details>
<summary><code>R1</code> — <code>destroy()</code>'s chown guard has a silent false-negative on an ambiguous container-state read</summary>

**Impact:** On the batch orphan-reap path (`packages/core/src/services/cleanup-service.ts:211`, which calls `backend.destroy(row.id)` in a loop over stale/orphaned container envs), a transient `docker inspect` failure or timeout causes `containerState()` to return `'unknown'` rather than `'running'`. `destroy()`'s new gate only attempts the ownership hand-back on a *definite* `'running'` read, so `'unknown'` and `'stopped'` are treated identically — the chown is skipped entirely, with **no log line**, and `destroy()` proceeds straight to `docker rm -f`. This directly contradicts the feature's own documented contract ("a failed hand-back is logged, never fatal" — `restoreHostOwnership`'s doc comment): here the hand-back isn't attempted at all, so nothing is logged, which is worse than the documented failure mode. `suspend()` (the other caller) has no such pre-filter — it calls `restoreHostOwnership` unconditionally and lets the helper's own try/catch handle an already-stopped container.

**Evidence:** `packages/isolation/src/backends/container.ts:332-338` (the gate), `container.ts:798-810` (`containerState`, which folds every inspect error into `'unknown'` by explicit design), `container.ts:373-395` (`suspend`'s unconditional call, for contrast), `packages/core/src/services/cleanup-service.ts:211` (the batch reap call site that makes this reachable).

**Required outcome:** Drop the `containerState()` pre-check in `destroy()` and call `restoreHostOwnership` unconditionally whenever `meta.artifactsMount !== undefined`, matching `suspend()`. The helper's own try/catch already handles a gone/stopped container cheaply and correctly.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer` (independently, lower confidence)

**Disposition:** OPEN

</details>

<details>
<summary><code>R2</code> — Resuming a pre-#3134 container env silently drops the artifacts bind</summary>

**Impact:** `ContainerEnvMetadata.artifactsMount` doesn't exist on any container-env row written before this PR. `resumeEnv`'s recreate branch (fires when the container itself is gone but its overlay volume survives — a designed-for recovery path, not a contrived one: `resumeEnv`'s own doc comment enumerates it) builds `HostMounts` straight from persisted metadata and re-applies whatever is present. For a legacy row, `meta.artifactsMount` is `undefined`, `assertMountableHostMounts` treats that as "no mount wanted" (correctly, for a *new* row with no artifacts to mount — but wrongly here), and the recreated container gets no artifacts bind. Every node writing to `$ARTIFACTS_DIR` after that resume writes into the container's own ephemeral overlay, invisible to the host. With `evidence_policy.required: true` the run later fails with a confusing "no evidence file exists" error; without it, the run completes "successfully" while the artifacts are silently lost. The PR body's own "Compatibility / migration" row acknowledges the mechanism ("resume exactly as before") without flagging that "before" means the bug this PR fixes persists for exactly those rows, with zero operator-visible signal.

**Evidence:** `packages/isolation/src/backends/container.ts:407-460` (`resumeEnv`, all three branches), `:449-450` and `:742` (the recreate branch's mount re-application and the artifacts `-v` arg), `packages/workflows/src/dag-executor.ts:12058-12063` (the evidence-policy failure this produces downstream). No test in the PR's diff exercises prepare-without-`artifactsMount` into a recreate-resume — the new `container.test.ts` cases all set `artifactsMount` on both `prepare` and the simulated resume.

**Required outcome:** At minimum, log a warning in the recreate branch when `sourceMount` is present but `artifactsMount` is absent, naming the cause ("resuming a container created before artifacts-mount support; `$ARTIFACTS_DIR` writes after this resume will not reach the host"). Refusing the recreate outright and pointing the operator at a fresh `--container` run is the more conservative alternative, consistent with how the same function already refuses outright when the volume itself is gone.

**Disposition:** OPEN

</details>

<details>
<summary><code>R3</code> — <code>HostIdentity</code> DI surface is wider than the one caller that needs it</summary>

**Impact:** None functionally — this is a maintainability/interface-narrowness concern, not a bug. `hostIdentity?: HostIdentity` was added to the public `ContainerBackendDeps` interface, plus an exported `HostIdentity` type and a `currentHostIdentity()` factory, solely so tests can pin uid/gid; neither production constructor site (`packages/isolation/src/backend-router.ts:78`, `packages/core/src/services/cleanup-service.ts:119,172`) passes it. The codebase already has a lighter, proven pattern for exactly this: `getProcessUid()` in `packages/providers/src/claude/provider.ts:419` is a bare exported function, called inline, and mocked in tests via `spyOn(claudeModule, 'getProcessUid').mockReturnValue(...)` — no constructor injection, no `Deps` field.

**Evidence:** `packages/isolation/src/backends/container.ts:71-90` (the `Deps` field, type, and factory), `:180,186` (the class field and constructor wiring), `:538` (its one use site in `restoreHostOwnership`); `packages/providers/src/claude/provider.ts:419,1402` and `provider.test.ts:92,107,113` (the existing lighter precedent).

**Required outcome:** Drop `hostIdentity` from `ContainerBackendDeps` and the private class field; keep `currentHostIdentity()` as a bare exported function called inline inside `restoreHostOwnership()`, spied directly in tests the way `getProcessUid` already is elsewhere in this repo. Optional — a real but modest simplification, not required for correctness.

**Disposition:** OPEN

</details>

<details>
<summary><code>R4</code> — CLI dispatch site's <code>artifactsMount !== undefined</code> duplicates the <code>preparedSource</code> gate one line above</summary>

**Impact:** None functionally today — both conditions are set from the same `preparedSource` truthiness in this diff. But `artifactsMount`'s definedness is established inside `if (preparedSource)` and then re-checked independently a few lines later when building the `backend.prepare()` call, rather than folding both optional fields under the single existing guard already used for `sourceMount`. Nothing enforces the two conditions stay equivalent as the function evolves.

**Evidence:** `packages/cli/src/commands/workflow.ts:2686-2701`.

**Required outcome:** Fold both fields under the one `preparedSource` guard: `...(preparedSource ? { sourceMount: preparedSource.captureRoot, artifactsMount } : {})`. Optional one-line tightening.

**Disposition:** OPEN

</details>

<details>
<summary><code>R5</code> — Stale doc cross-reference: <code>$STATE_DIR</code>'s row still points at a caveat this PR removed from <code>$ARTIFACTS_DIR</code></summary>

**Impact:** `packages/docs-web/src/content/docs/reference/variables.md` line 25 (the `$STATE_DIR` row, untouched by this diff) reads: "**Container runs (`--container`):** same caveat as `$ARTIFACTS_DIR` — the host path is not mounted into the container...". Line 24 (the `$ARTIFACTS_DIR` row directly above it) was just changed by this PR to say the opposite — it now IS mounted read-write. The literal fact stated about `$STATE_DIR` itself is still accurate (it genuinely stays unmounted), but the comparison anchor ("same caveat as `$ARTIFACTS_DIR`") is now false and will read as contradictory to anyone comparing the two rows.

**Evidence:** `packages/docs-web/src/content/docs/reference/variables.md:24-25`.

**Required outcome:** Reword the `$STATE_DIR` row to state its own caveat directly instead of anchoring it to `$ARTIFACTS_DIR`'s (which no longer has one).

**Found by:** coordinator (direct diff read; outside the three selected scopes — no comments/docs agent ran this round)

**Disposition:** OPEN

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 |
| `seams` | R1 (lower confidence), R2 |
| `simplify` | R3, R4 |

#### Validation

|---|---|---|
| `cd packages/isolation && bun test src/backends/container.test.ts` | PASS | 38 pass, 0 fail, 93 expect() calls — matches the PR's own claim |
| `cd packages/cli && bun test src/commands/workflow.test.ts` | PASS | 395 pass, 0 fail, 917 expect() calls (required `bun install` in the fresh review worktree first) |
| `bun run type-check` | PASS | exit 0 |
| `bun run lint` | PASS | exit 0 |
| `bun run format:check` | PASS | prettier clean |
| Live Docker e2e (added by this review, not in the PR's own validation): isolated `ARCHON_HOME`, folder project, `--container` run with a `bash:` node writing `$ARTIFACTS_DIR/probe.txt` + `evidence_policy: { required: true }` | PASS | Docker Desktop for Mac 4.44.3, arm64, native overlay mode. Container ran as `uid=0`; workflow completed successfully; `probe.txt` and `evidence.json` landed on the host with correct content at `<home>/workspaces/_folder/project/artifacts/runs/<id>/`, owned `rasmus:wheel` on the host side; the sibling `workflow-source/runs/<id>/` directory (frozen source) confirmed separate from the artifacts tree; `docker ps -a`/`docker volume ls` confirmed clean teardown after the run. Rootful-Linux chown behavior (the PR's own stated "known risk or uncertainty") could not be verified on this platform — Docker Desktop for Mac remaps container-root writes to the host user regardless of the explicit `chown`, so this run cannot distinguish "chown worked" from "Desktop remapped it anyway." |

</details>

Wirasm:
All five findings addressed in 9aacf27. Base retargeted to `dev` now that #3132 has merged, so the full CI matrix runs on this head.

| ID | Change |
|---|---|
| R1 | `destroy()` calls the hand-back unconditionally, the same as `suspend()`; the container-state pre-check is gone. The helper reads a stopped or gone container (the normal destroy-after-suspend path) as a debug-level skip rather than a failed hand-back, using the same daemon-message reading `suspend()` already uses. Test: `destroy` with a failing `inspect` still issues the chown before `rm`. |
| R2 | The recreate branch logs `isolation.container_resume_without_artifacts_mount` when a row carries a source mount but no artifacts mount, which only a pre-artifacts-bind row can be. Warn rather than refuse: the run resumes exactly as it would have before the upgrade, and refusing would strand paused runs that were fine yesterday. Test: a source-only row recreates with the source bind and no artifacts bind. The log line itself is not asserted (the module logger is private); the branch is exercised by that test. |
| R3 | `hostIdentity` is off `ContainerBackendDeps`. `currentHostIdentity()` is a bare export called inside `restoreHostOwnership`, and tests pin it with `spyOn`, matching `getProcessUid`. |
| R4 | The CLI dispatch builds both mounts under the single `preparedSource` guard. |
| R5 | The `$STATE_DIR` row states its own caveat and names `$ARTIFACTS_DIR` as the one that is mounted. |

Container suite 39 pass, CLI suite 395 pass, type-check, lint, and format clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #3041, opened 2026-08-31
feat(forge): resolve and checks through optional plugin dispatch


Wirasm:
## Problem

The SDLC CI gate and related deterministic reads still depend on vendor-specific tooling and ambient repository/PR context. A check result must identify the repository and revision it evaluated, enumerate actual check units (including external commit statuses), and expose unknown/gated results honestly. There is no outbound optional-plugin path proving these facts through a real pack consumer.

## Why now

#3040 is settled and #998 is implementing inbound identities/events. This slice proves outbound dispatch and shared schema ownership before lifecycle writes expand the seam.

## Proposed solution and desired outcome

Deliver explicit `archon forge resolve` and `checks` through the versioned exec-plugin contract, using GitHub as the first conforming plugin. Rewire the current dev pack check gate to use a qualified PR/revision observation instead of ambient `gh` probes. Inventory current source; do not reproduce obsolete `until_bash`/check-ci.py arrangements from the August issue.

## Governing contract

The [22 September contract decision](https://github.com/coleam00/Archon/issues/3040#issuecomment-5773123042) supersedes conflicting older proposals. The workflow owns policy. The engine works with no forge and does not know which forge is used. Optional integrations own forge schemas, authentication, event normalization and operations; hosts pass resolved starts/capability evidence through generic engine boundaries. No mandatory forge discovery, forge-specific executor branch, or universal merge/approval policy.

GitHub is the first producer, not a private implementation path. Schemas/types/docs derive from one optional owner. Persisted compatibility remains additive. Supported capabilities and mutation guarantees must be truthful; observed-before and verified-after are not atomic enforcement.

## User flow

Before: gate shells out to vendor tooling and can infer its PR from the current checkout or confuse absent Actions workflows with absent checks.
After: pack supplies qualified identity; the plugin enumerates check runs/statuses with evaluated revision; the pack applies its chosen required/full-set policy. Trace contains operation, target, plugin and result. Unrelated local workflows never load forge dispatch.

## Scope

- Optional schema/dispatch/conformance package, lazy CLI entry, GitHub executable producer, exact operation identity and audit linkage.
- Discovery/static host claims plus explicit self-hosted mapping, version handshake, duplicate-claim error, structured op vs process errors, constrained credential/env transport, bounded output and process-tree termination per #3040.
- Coordinate with #998 owner: reuse its identity/check schema and package manifest; no competing copy or outbound scaffolding in #998. Agree file ownership before editing shared files and serialize integration.
[...]

PR #3145, opened 2026-09-02
fix(workflows): flip-ready refuses on a failed check-state read

### LINKED ISSUE #3128: fix(sdlc): flip-ready treats a failed check-state read as no CI
## Problem

The final `flip-ready` preflight treats every failed `gh pr checks` read as an empty check set.

The node runs the check query with `2>/dev/null || true`, stores stdout in `NOT_GREEN`, and proceeds whenever that value is empty. GitHub's legitimate “no checks reported” case is empty and non-zero, but authentication failures, network failures, permission errors, malformed responses, and other check-read failures can also be empty and non-zero. They therefore take the same branch and reach `gh pr ready`.

This path predates PR #3115 and was validated as an adjacent discovery by its correction run `be99df920a535116b1cdc03225210e71`.

## Why

`flip-ready` is the final irreversible public action in the delivery workflow. Its purpose is to refuse while checks are pending or non-green. A failed observation is not evidence that no CI exists, and treating it as permission makes the safety gate depend on whether a second forge command happens to fail too.

The pack already requires unsupported or ambiguous states to fail loudly. The final gate should preserve the reason it could not establish check state and must never broaden permission because a read failed.

## Desired outcome

The final ready preflight distinguishes a successfully observed empty check set from a failed check-state read.

- A successfully observed no-checks state retains the pack's documented permission to proceed.
- Pending, red, gated, or otherwise non-green checks still refuse.
- Authentication, network, permission, malformed-response, and unclassified read failures refuse before any ready-flip attempt and preserve the original evidence.

The distinction uses a structured command/API result. It does not classify vendor prose with regexes or silence a failure to preserve the no-CI case.

## Acceptance

- [ ] A successful check read with zero checks reaches the existing no-CI path.
- [ ] A pending or non-green check refuses with the check names/state as today.
- [ ] A check-read command that exits unsuccessfully with empty stdout refuses before `gh pr ready` is invoked.
- [ ] Authentication, network, permission, malformed-response, and unknown failures preserve actionable original evidence rather than collapsing to an empty result.
- [ ] The implementation uses a structured result or API field to distinguish “zero checks” from “could not read checks”; no vendor-error prose parser is introduced.
- [ ] A focused fixture or script test is seen red against the current `|| true` behavior by stubbing a failed check read and a ready command that would otherwise succeed.
- [ ] Existing clean, pending, introduced-red, inherited-red, environmental-red, and no-CI delivery fixtures remain green.

## Evidence

Current `dev`:

- `.archon/workflows/sdlc/deliver/archon-deliver.yaml:521-533` describes the irreversible preflight and captures `gh pr checks ... 2>/dev/null || true`.
- `.archon/workflows/sdlc/deliver/archon-deliver.yaml:534-538` refuses only when the captured non-green text is non-empty.
- `.archon/workflows/sdlc/deliver/archon-deliver.yaml:542` attempts `gh pr ready` after an empty result.

Discovery source: Archon correction run `be99df920a535116b1cdc03225210e71`, `discoveries.md` and `discoveries.json`, reviewing PR #3115.

- Related: #2976 / PR #3115 owns durable recovery after inherited or environmental red; this issue owns the final check-read boundary that predates it.
- Related: #3041 will eventually route checks through the forge contract's structured check vocabulary; this live irreversible path remains independently required until that replacement lands.
- Must remain true: no-CI repositories can still complete delivery after a successful empty observation.
- Must remain true: `flip-ready` never marks a PR ready with pending or non-green checks.
- Solution steering: Requirement — use structured check data and fail closed on an unreadable state. Do not distinguish the no-checks case by matching GitHub error prose.

Wirasm:
## Problem and outcome

The `flip-ready` node's final irreversible preflight ran `gh pr checks … 2>/dev/null || true`, which made any failed check-state read (auth failure, network error, permission denied, malformed response) indistinguishable from a legitimate empty check set. A repository with no CI and a broken `gh` invocation both produced empty `NOT_GREEN` and reached `gh pr ready`.

- **Outcome:** `flip-ready` now reads the check context count through a structured GraphQL API call whose exit code is the success signal. A non-zero exit refuses before any ready-flip. A zero count (legitimate no-CI) proceeds. When checks exist, the existing `--json` classification guards against failed reads too.
- **Invariant:** Repositories with no CI continue to deliver without error. The six existing fixture families (clean, pending, introduced-red, inherited-red, environmental-red, no-CI) remain green.
- **Scope boundary:** The external CI probe (`check-ci.py`) is a different node and is not touched.
- **Root cause:** `|| true` unconditionally suppressed the non-zero exit from a failed `gh` command, so an empty stdout was treated the same as a successful "no checks" result.

## Review guidance

- **Feedback requested:** Correctness of the guard logic and the GraphQL query shape.
- **Start here:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:560` — the replacement bash block for the `flip-ready` node's check-state read.
- **Review order:** YAML bash change → `.archon/scripts/__tests__/flip-ready-check-read.test.ts` (the new regression test) → `packages/workflows/src/defaults/bundled-defaults.test.ts` (updated fixture stubs).
- **Lower-attention areas:** `packages/workflows/src/defaults/bundled-defaults.generated.ts` — regenerated automatically by `check:bundled`; the only delta is the updated bundled YAML text.

## Solution

The bash block now works in three segments instead of one:

1. **Check-count read** — `gh api graphql` fetches `statusCheckRollup.contexts.totalCount` for the PR's latest commit. Exit 0 with a JSON value is the structured success signal. A non-zero exit means the read itself failed (auth, network, permission) and the node refuses, naming the failure.
2. **Zero-checks path** — When `totalCount` is 0 (no CI configured, or gated checks that never started), the node proceeds directly to `gh pr ready`, matching the original no-CI behavior.
3. **Classification path** — When checks exist (`totalCount > 0`), the node runs `gh pr checks --json name,bucket` as before, but now without `|| true`. If this read fails (non-zero exit), the node refuses because a checks-exist-but-cannot-classify state is indistinguishable from all-green by intent. If non-green checks are found, the node refuses with their names and states as before.

This satisfies the "no vendor-error prose parser" constraint: the count read uses a structured GraphQL field and exit code, the classification read uses structured JSON output and exit code. Both stderr streams are dropped (`2>/dev/null`) per the node's existing value-read convention — a `gh` diagnostic message must never become the count or the check list.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A failed `gh pr checks` (auth failure, network error, permission denied, malformed response) produces empty output and reaches `gh pr ready`. | A failed check-state read refuses loudly before any ready-flip. A successful read with zero checks still reaches `gh pr ready`. |
| **Failure behavior** | Silent: the run processes as if no CI exists and flips the PR ready even when checks could not be observed. | Audible: the node exits 1 with "flip-ready: could not read check state." or "flip-ready: checks exist but their state could not be read." on stderr. |

### User flow

```mermaid
flowchart TD
    Start[flip-ready starts] --> Count{Read check count<br>via gh api graphql}
    Count -- exit non-zero --> RefuseCount[Refuse: could not read]
    Count -- exit 0, count = 0 --> Ready[Flip PR ready]
    Count -- exit 0, count > 0 --> Classify{Read check states<br>via gh pr checks --json}
    Classify -- exit non-zero --> RefuseClassify[Refuse: state unreadable]
    Classify -- exit 0, all green --> Ready
    Classify -- exit 0, non-green found --> RefuseNonGreen[Refuse with check names]
```

## Validation

- `bun run validate` — exit 0 — proves the full suite (type-check, lint, format, bundled checks, tests) passes.
- `bun cli workflow test archon-deliver` — 21 passed, 0 failed — proves all six fixture families (clean, pending, introduced-red, inherited-red, environmental-red, no-CI) stay green.
- `bun test .archon/scripts/__tests__/flip-ready-check-read.test.ts` — 5 passed — proves the new regression test covers: failed count read refuses, empty checks proceeds, non-green refuses, all-green proceeds, checks-exist-but-classification-fails refuses.
- The failed-count-read test was observed red against the prior `|| true` code before the YAML edit was applied.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No migration needed. Existing workflow runs store no state from the changed bash block. | The change is internal to a single node's bash body; the node's inputs (environment variables, outputs from earlier nodes) are unchanged. |
| Rollout / rollback | Safe: the run's workflow source is frozen at launch, so the fix applies to new deliveries. Rollback is a revert. | The YAML diff is contained to one node. |
| Observability | A failed check read now appears in the run log as a node failure with a descriptive error message, where previously it would have been invisible as a successful flip. | Regression tests verify the failure messages. |

## Links

- Closes #3128

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the readiness preflight to safely handle unavailable or failed check-status reads.
  * Pull requests are no longer marked ready when check results cannot be reliably retrieved or classified.
  * Pull requests with no checks can still proceed, while non-green checks continue to block readiness.

* **Tests**
  * Added regression coverage for failed status queries, empty check sets, non-green checks, successful checks, and classification failures.


Wirasm:
# Review Report — PR #3145

## Verdict

- **Ready:** true
- **Action:** none
- **Reason:** The change correctly distinguishes a failed check-state read from a legitimate empty check set through two structured API calls. No blocking findings remain. The single Important finding from the errors lens was disproved by adversarial verification against the actual code and API semantics. The one test-coverage gap is a suggestion, not a blocker.

## Accepted contract

**Required outcome:** The `flip-ready` node's final irreversible preflight must distinguish an observed empty check set (legitimate no-CI) from a failed check-state read (auth, network, permission, malformed response). Three-segment bash block: check-count read via `gh api graphql`, zero-checks path, classification path. No vendor-error prose parser. No silent failure. Six existing fixture families remain green.

**Non-goals / boundaries:** The external CI probe (`check-ci.py`) is not reviewed. The forge-contract replacement (#3041) is not part of this review. The auto-generated `bundled-defaults.generated.ts` is reviewed only for regeneration correctness.

## Reviewed head SHA

`f3aca2764cc917dafc263f88be177777be39f546`

## Findings

### Disproved finding

#### R1 — `// 0` fallback masks null statusCheckRollup as no-CI (Important)

**sources:** [errors]

**Claim:** The `--jq '(... // 0)'` fallback converts a null `statusCheckRollup` (which could occur when checks exist but have not been materialised) to a count of 0, causing the node to skip the `gh pr checks` classification step and proceed to `gh pr ready` without having observed CI state.

**Evidence examined:**

- File: `.archon/workflows/sdlc/deliver/archon-deliver.yaml`, line 572–574
- The GraphQL query: `statusCheckRollup { contexts { totalCount } }`
- The jq expression: `(.data.repository.pullRequest.commits.nodes[0].commit.statusCheckRollup.contexts.totalCount // 0)`

**Adversarial verification:**

1. **API semantics.** In GitHub's GraphQL API, `statusCheckRollup` is null when the commit has no status checks or check runs — the legitimate no-CI case. When checks exist, the rollup is a non-null object with a `contexts` connection. The `// 0` fallback correctly converts the null to 0, which is the expected behavior for the no-CI path.

2. **Timing concern.** The `flip-ready` node runs after the full CI convergence loop (`await-checks`, `ci-verdict`, `ci-corrections`, `await-fix-checks`, `ci-attention`). By the time `flip-ready` executes, checks have been observed and resolved. The window where `statusCheckRollup` is null despite checks existing is vanishingly narrow — a push would need to land between the CI loop and the flip-ready execution, and the CI loop already detects new pushes.

3. **Proposed fix would break the no-CI path.** The error reviewer's proposed correction removes the `// 0` and adds a `case` statement checking for `''|null`. This would cause the node to refuse on the legitimate no-CI case where `statusCheckRollup` is null (no checks configured), violating the accepted contract's invariant that "repositories with no CI continue to deliver without error."

4. **The `// 0` is a deliberate design choice.** The implementation.md explicitly documents: "totalCount == 0 (or a null rollup, mapped by --jq '... // 0') is the legitimate no-checks case and proceeds, matching the probe's permission-agnostic skip."

5. **All tests pass.** The regression test suite (5 tests) and the existing fixture suite (31 tests, 6 fixture families) all pass, confirming the behavior is correct for both the no-CI and checks-exist paths.

**Verdict:** Disproved. The `// 0` fallback correctly implements the required behavior. The finding does not identify a real defect.

---

### Open Suggestions

#### S1 — No stderr-cleanliness test for the checks-exist-and-pass path (Suggestion)

**sources:** [tests]

**Claim:** The `bundled-defaults.test.ts` has a stream-cleanliness test for the no-CI success path (line 865) that verifies stdout is the bare URL and stderr is empty even when fake gh appends a notice to every call. No equivalent test exists for the checks-exist-and-pass path (totalCount > 0, all checks green or skipped).

**Evidence examined:**

- `packages/workflows/src/defaults/bundled-defaults.test.ts`, line 865 (no-CI stream-cleanliness test)
- `flip-ready-check-read.test.ts`, lines 143–148 (all-green success test — uses fake gh with no stderr noise)
- The `gh pr checks` call at `archon-deliver.yaml` line 582, which has `2>/dev/null`

**Verdict:** Valid suggestion. A future edit could remove the `2>/dev/null` from the classification `gh pr checks` call, and the existing tests would not catch the resulting stderr leakage on the checks-exist-and-pass path. The smallest valuable test follows the pattern of the existing no-CI stream-cleanliness test at line 865, using a fake gh that appends a notice to every call. Non-blocking.

---

### Rejected findings (outside evidence)

The errors lens claimed the `// 0` fallback is an Important defect. This finding was disproved through adversarial verification (see R1 above). No other findings were raised by the code, seams, or simplify lenses.

## Prior findings

Not applicable — first review round.

## Discoveries

No discoveries directory was present. No adjacent discoveries or scope conflicts were recorded by the specialist lenses.

## Review coverage

### Lenses that ran

| Lens | Status | Findings |
|---|---|---|
| **code** | Ran, no findings | — |
| **seams** | Ran, no findings | — |
| **simplify** | Ran, no findings | — |
| **tests** | Ran, 1 suggestion | S1 |
| **errors** | Ran, 1 finding (disproved) | R1 |
| **docs** | Disabled per scope.md | — |

### Required lens verification

All four required lenses (code, seams, simplify, tests) ran and reported. The errors lens ran and reported, though its single finding was disproved. The docs lens was correctly disabled per scope.md (`docs: false`). No lens failed to report.

### Evidence limitations

The `// 0` behavior was verified against GitHub's GraphQL API documentation and the implementation's own test results. The exact API response for `statusCheckRollup` on a commit with no checks was not independently queried (requires a live GitHub API call), but the implementation's documented behavior and the deliberate design choice in the code are sufficient evidence. The stderr-cleanliness suggestion was verified by reading the test file and the production code.

PR #3058, opened 2026-08-31
feat(console): multi-chat history, run chat tab, and live execution cards

Thummpy:
## Problem and outcome

Console UX was missing multi-chat session management, direct run-to-orchestrator interactive chat, preserved multiline message formatting, and inline live DAG execution / result visibility in the chat stream.

- **Outcome:**
  1. Multi-chat management sidebar (`ChatHistoryPanel`) with `+ New Chat`, inline title renaming, deletion confirmation, and automatic fallback navigation.
  2. Interactive orchestrator `Chat` tab in the run detail view (`RunDetailPage`) with two-way dispatching (`ChatComposer`), live auto-scrolling, running status indicator, and run-scoped navigation.
  3. Preserved whitespace and multiline code indentation in user message bubbles (`whitespace-pre-wrap`).
  4. Live embedded workflow execution cards (`WorkflowProgressCard`) and completion result cards (`WorkflowResultCard`) with artifact preview modals in `ChatStream`.
  5. Workflow parent orchestrator conversations are hidden from the primary chat session list to avoid duplicate clutter.
- **Invariant:** Existing console navigation, keyboard shortcuts, log streaming, and classic UI routing continue to operate seamlessly.
- **Scope boundary:** Does not modify backend workflow execution mechanics or CLI runtime behavior.

## Review guidance

- **Feedback requested:** Correctness of conversation routing (`/console/p/:projectId/chat/:convId`), orchestrator chat stream subscription / keymap bindings on `RunDetailPage`, and component boundary design of `WorkflowProgressCard` and `WorkflowResultCard`.
- **Start here:** `packages/web/src/experiments/console/routes/RunDetailPage.tsx` — Orchestrator chat tab wiring and live streaming synchronization.
- **Review order:**
  1. `packages/web/src/experiments/console/components/ChatHistoryPanel.tsx` and `packages/web/src/experiments/console/routes/ChatPage.tsx` (sidebar navigation, rename/delete lifecycle, and routing)
  2. `packages/web/src/experiments/console/routes/RunDetailPage.tsx` and `packages/web/src/experiments/console/components/StreamToolbar.tsx` (run detail chat tab)
  3. `packages/web/src/experiments/console/components/WorkflowProgressCard.tsx` and `packages/web/src/experiments/console/components/WorkflowResultCard.tsx` (DAG cards in chat)
  4. `packages/server/src/routes/api.ts` (orchestrator conversation visibility flag)
[...]

PR #1982, opened 2026-06-12
fix: user-scope console chat and resolve chat credentials from the sender

Wirasm:
## Summary

- **Problem**: On multi-user installs, the console's per-project chat is shared by every user, and the orchestrator resolves per-user AI prefs + provider credentials from `conversation.user_id` (the **first creator**) — so user B's turns in a thread user A created execute on **A's API keys and model prefs** (#1976).
- **Why it matters**: Cross-user credential use — one user silently spends another's API credits and inherits their model configuration; all users also share one AI session/history.
- **What changed**: (1) `handleMessage` now resolves execution identity sender-first: `executionUserId = context.userId ?? conversation.user_id`, used at both the AI-prefs seam and the provider-credential-env seam (and the profile-invalid log). This mirrors the workflow executor, which already resolves prefs from the run starter. (2) The console's `listConversations` passes the existing non-enforcing `mine=true` filter, so each signed-in user gets their own lazily-created per-(user, project) conversation — and therefore their own session.
- **What did not change (scope boundary)**: No visibility enforcement (`mine=true` stays non-enforcing per the documented open-visibility posture); no rejection of cross-user sends; no conversation switcher UI; no change to `getOrCreateConversation` first-user-wins creation semantics (row `user_id` remains provenance + fallback); no legacy `/legacy` ChatPage changes (the core seam protects it automatically); no schema or route changes.

### Before

```
  User A                 User B                  Archon (multi-user install)
  ──────                 ──────                  ───────────────────────────
  opens project chat ──────────────────────────▶ lists ALL web convs → conv-1
                         opens project chat ───▶ same list → conv-1 (A's!)
                         sends "refactor X" ───▶ sender=B resolved… then dropped:
                                                 prefs   ← getUserAiPrefs(A)
                                                 keys    ← A's provider env
                                                 session ← conv-1 (A's history)
  sees B's messages ◀──────────────────────────  one interleaved thread
```

### After

```
  User A                 User B                  Archon
  ──────                 ──────                  ──────
  opens project chat ──────────────────────────▶ list ?mine=true → A's conv only
                         opens project chat ───▶ ?mine=true → none → *lazy-create
                                                  conv-B (user_id=B)*
                         sends "refactor X" ───▶ *executionUserId = sender(B)*
                                                 prefs   ← getUserAiPrefs(B)
                                                 keys    ← B's provider env
                                                 session ← conv-B (own history)

  Solo install (no identity): mine=true narrows nothing; sender undefined →
[...]

Issue #1976, opened 2026-06-12
console chat is not user-scoped on multi-user installs — shared thread + cross-user credential/prefs resolution


Wirasm:
## Problem

On a multi-user install (web auth enabled, e.g. the VPS deployment), the console's per-project chat (`/console/p/:projectId/chat`) is **shared by all users**, not scoped to the signed-in user.

### 1. Everyone lands in the same thread

- `ChatPage` auto-selects the **first** `web` conversation for the project (`packages/web/src/experiments/console/routes/ChatPage.tsx:57-62`).
- `listConversations` queries `GET /api/conversations?codebaseId=<id>` **without** `mine=true` (`packages/web/src/experiments/console/skills/conversations.ts:37-42`), so every user resolves the same conversation.
- Result: all users' messages interleave in one thread and ride one shared AI session/history.

### 2. Cross-user credential and prefs resolution (the security-relevant part)

`conversation.user_id` records the **first creator only** (`packages/core/src/orchestrator/orchestrator-agent.ts:880` comment), and the orchestrator resolves per-user state from that field:

- per-user AI prefs: `orchestrator-agent.ts:1201-1202`
- per-user provider credentials injected into the chat env: `orchestrator-agent.ts:1284-1285`

So when user B sends a message in the project chat that user A created, B's turn executes with **A's provider API keys / subscription and A's model prefs**. B spends A's credits; A's tier/alias overrides silently apply to B's turns.

## Proposed fix

**Minimum (web):** scope the console chat per user —
- `listConversations` passes `mine=true` when an identity resolves (the server filter already exists: `api.ts:2220-2223`)
- conversation auto-select picks the current user's web conversation; lazy-create one per (user, project) instead of per project

**Defense in depth (server):** the `mine` filter is explicitly non-enforcing today. For chat specifically, consider resolving credentials/prefs from the **message sender** rather than the conversation creator, or rejecting sends into a conversation owned by another user when the API gate is on. Otherwise any client that omits `mine=true` recreates the cross-user credential path.

Open design question: shared per-project threads may be a *feature* for team visibility — but then credential/prefs resolution must follow the sender, never the creator. Either direction removes the current mismatch.

## Context

- Multi-user = web auth enabled (`DATABASE_URL` + `BETTER_AUTH_SECRET`); solo installs are unaffected (no identity → config-only behavior).
- Related: per-user provider keys (#1891 epic), per-user AI prefs (Phase 3, #1948).
[...]

PR #3147, opened 2026-09-02
fix(auth): operator conversations are private to their owning user

### LINKED ISSUE #3135: fix(auth): operator chats are private to their owning user
## Problem

Authenticated Archon installs support multiple users, but conversation visibility and access are not consistently user-private. The Console requests a user-filtered list, while the filter is explicitly non-enforcing and conversation read/write routes do not all enforce ownership.

The earlier sender-identity fix in #1976 / #1982 correctly ensures a turn uses the acting user's credentials and preferences. It deliberately left conversation visibility open. The product decision is now that users on the same install must not share operator-chat history unless an explicit sharing model is designed later.

## Why

Operator chat can contain project context, run diagnostics, logs, and operational instructions. One user must not be able to read or act through another user's chat merely because both use the same Archon install.

## Desired outcome

When Web authentication is enabled, every operator conversation is private to its owning user across discovery, history, streaming, messages, and mutations. Solo installs without authentication keep their simple local behavior.

## Acceptance

- [ ] An authenticated user sees only their own operator conversations.
- [ ] Conversation lookup, message history, streaming, send, rename, reset, and delete paths reject access by another user.
- [ ] Identity-resolution failure never broadens an authenticated request to all conversations.
- [ ] Legacy ownerless conversations follow one explicit, safe compatibility policy without exposing them across users.
- [ ] Acting-user credential and preference resolution delivered by #1982 remains unchanged.
- [ ] Focused tests cover every conversation API ingress for owner, non-owner, unresolved identity, and unauthenticated solo operation.

## Evidence

The current list filter is documented as non-enforcing, and PR #3058 adds ownership checks only to a subset of mutation routes. The gap is already publicly documented in #1976 and #1982; this issue records the stricter product decision.

- Must remain true: Multiple users within one install are supported without introducing deployment-level tenant semantics.
- Scope boundary: This does not add shared chats, teams, roles, or cross-user collaboration.
- Related issues or PRs: #1976, #1982, #3058.

Wirasm:
## Problem and outcome

On an install where real identity arrives, any signed-in user could list, open, read, stream, rename, delete, or send into anyone else's web and CLI conversations, and could act on runs they did not start. The console's user filter was explicitly non-enforcing, and a failed identity resolution widened a narrowed request back to the whole install.

- **Outcome:** when web authentication or a proxy identity header is configured, every operator conversation ingress (list, get, messages, stream, send, rename, reset, delete) is private to its owning user; the eight run actions require the run's starter; an unattributed row matches nobody; and `archon conversations claim` is the documented, shell-only way to attach pre-enforcement history to a user.
- **Invariant:** solo and SQLite installs without authentication keep today's open behavior. Acting-user credential and preference resolution from #1982 is unchanged. Multiple users within one install remain supported without tenant semantics.
- **Scope boundary:** run *reads* stay open on purpose; only run actions are gated. CLI identity linking (a web user claiming a CLI identity so CLI chats appear in their console) is cut from this PR to #3152 by the decision on #3135; under enforcement an unlinked CLI conversation is simply invisible, which is the safe outcome.

## Review guidance

- **Feedback requested:** this is an auth boundary, so adversarial depth: every conversation ingress, the SSE route, and the eight run actions, with the question "what would make this merge wrong". Check the fail-closed claims against the SQL, not the prose.
- **Start here:** `packages/core/src/db/conversations.ts` — `ConversationVisibility` replaces the optional `userId` filter; `undefined` meaning "no filter" was the widening bug.
- **Review order:** `core/auth/config.ts` (the one enforcement predicate, shared by CLI and server) → `core/db/conversations.ts` → `server/routes/api.ts` (`authorizeConversation`, `authorizeRunAction`, the SSE route) → `core/db/workflows.ts` and `cli/commands/conversations.ts` (the claim path) → tests → docs.
- **Lower-attention areas:** console skill wiring and docs.
- **Known risk or uncertainty:** enabling authentication on an existing install makes its whole pre-enforcement history invisible until claimed. That is the intended fail-closed policy, documented in the CLI reference under `conversations`.

## Solution

One pure env predicate in core, `isConversationOwnershipEnforced`, answers whether identity arrives on this install; the CLI and the server read the same function. The conversation store takes a required visibility union instead of an optional user id, so "everything" has to be asked for by name and a failed identity resolution can no longer widen a query. Six by-id routes and the stream route authorize through one helper that loads the row and applies the rule; denied conversations return 404 because their ids are semi-guessable, denied run actions return 403 because a run id is an opaque UUID the caller already holds. Claiming is one `UPDATE … WHERE user_id IS NULL`, so it can never move a row between two real users.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | authenticated users saw and could act on every web and CLI conversation and every run | each user sees and acts on their own conversations; run actions require the starter; run reads unchanged |
| **Failure behavior** | a failed identity resolution returned the whole install's conversations | a failed resolution returns nothing; unattributed rows are reachable only through `archon conversations claim` |

## Validation

Carried commits rebased onto `dev` at `4d48a0325` and adapted to its test typing (the identity-linking commits `fa198c504`, `1ebbda194`, `8f489e2ef` from the original branch are not included):

- `bun run test` in `packages/core` and `packages/server` — green, including `api.conversation-ownership.test.ts` (owner, non-owner, unresolved identity, solo operation for every ingress), `api.stream.test.ts`, and the run-action matrix in `api.workflow-runs.test.ts`.
- `bun test src/commands/conversations.test.ts` and the CLI help and dispatch tests in `packages/cli` — green.
- `bun run type-check`, `bun run lint`, `bun run check:api-types` — green.
- Adversarial review (independent agent, against `08fb8a85c`): every conversation ingress and all eight run actions gated; fail-closed SQL verified on both dialects; enforcement-off behavior diffed as unchanged; the ownership guard neutralized and 21 tests observed red. Its one critical finding, the ungated `DELETE /api/workflows/{name}/node-sessions` reset, is fixed and in the ownership matrix (five tests red without the gate). CodeRabbit's four findings on the rescued head are fixed in one commit, each seen red first.
- **Not verified:** `bun run validate` end to end; CI runs it.

## Links

- Closes #3135
- Related #3152 (CLI identity linking, cut from this PR)
- Related #1976, #1982, #3058


Wirasm:
Rescued per the decision on #3135: history force-pushed (`68334cf0f` → `2e562f76a`), rebased onto `dev` at `4d48a0325`. Kept: list scoping via the visibility union, by-id route authorization, the SSE rule, owner-gated run actions, and `archon conversations list --unowned` / `claim`. Cut: CLI identity linking (three routes, the console panel, `auth whoami`, the no-minting resolver), now #3152 with a restore list. One adaptation commit carries the kept commits onto dev's test typing and help index; core and server suites, the CLI tests, type-check, lint, and the API type check are green locally. The inherited `main` commit is gone with the rebase.

Wirasm:
CodeRabbit's four findings on the rescued head are addressed in one commit, each verified first and each new test seen red with its guard reverted:

- **Run claim not tied to the conversation's owner** (Major, CWE-862): `ownerlessRunWhere` now requires the answerable conversation to be unowned or owned by the claiming user, in the preview count and the UPDATE alike. Test in `packages/core/src/db/workflows.test.ts`.
- **Refused multipart run deleted the shared upload directory** (Major, CWE-362): a refused request removes only its own files; the completion path keeps removing the directory it owns. Regression test in `api.conversation-ownership.test.ts` plants an owner's file, refuses an intruder's multipart run, and asserts the owner's file survives (red with a recursive delete, green without).
- **`--json` advertised on `conversations list`** (Major): removed from the spec and the pinned help index; it was never implemented.
- **Stray positionals ignored** (Minor): both subcommands refuse extras with the existing usage error; test added.

Wirasm:
The adversarial review found one ingress the decision names that the branch had not gated: `DELETE /api/workflows/{name}/node-sessions` resets a conversation's persisted provider sessions by internal id with no ownership check, and in the proxy-header configuration it was reachable anonymously with `?confirm=all-scopes`. Fixed in the latest commit: the ownership rule now lives in `authorizeLoadedConversation` over a loaded row, shared by the platform-id routes and this one; under enforcement a scope must name a conversation the caller owns (404 otherwise) and the cross-scope wipe is refused (403). Enforcement off is unchanged. The route is in the ownership matrix, and the route-inventory guard now carries an explicit list for conversation ingresses outside the `/api/conversations/{id}` prefix. Five tests seen red without the gate.

The review's other two findings (`auth whoami` residue, the advertised `--json`) were already addressed in the previous commit.

Wirasm:
Operator decision, 2026-09-02: this PR is held and will not go into the upcoming release. It ships in the release after that. Converted to draft so it cannot be merged by accident; the branch is complete, reviewed, and green as of `9812ba659`, and only needs a rebase check when the hold lifts.

Issue #3127, opened 2026-09-01
refactor(sdlc): validate deterministic pack scripts with TypeScript


Wirasm:
> **Brought current 2026-09-10 (additive).** Every prerequisite has landed: PR #3123 (validation shape, 2026-09-01), the #2453 pair (#3158, #3148), #3155 (CI checks signal an event wait), #3157 via PR #3174 (timeout skips), #3156 (skip provenance, PR #3249), #3043 (terminal record, PR #3254), and #3251 (a pack-level `.shared/` directory packaged scripts import by relative path, PR #3252). Measured on `dev` at `edce16a73`: twelve Python scripts, 1,290 lines, not the ten and 1,225 below. The adoption list is the 2026-09-02 comment on this issue plus the #2807 inventory; one item has moved since that comment was written: #3155 is delivered, so deliver's CI probe and duration-wait loop become a `wait: event` on the CI check signal rather than a kept defence. What stays as cited defence is the forge line only (#3040 onward). Nothing blocks this issue.

## Problem

The bundled SDLC pack has twelve deterministic Python scripts under
`.archon/workflows/sdlc/**/scripts/`, totaling 1,290 lines (measured 2026-09-10; the original count was ten and 1,225). They implement
consequential gates and state classification, including green/readiness checks,
CI status classification, review-verdict validation, and terminal outcomes.

Archon's maintained codebase and validation machinery are TypeScript-first. The
pack scripts therefore sit outside strict TypeScript and typed lint, while
validating them statically would require a second Python toolchain and parallel
configuration.

## Why

These scripts turn workflow results into decisions that control spend and public
delivery. Static diagnostics are valuable on this path, and maintaining one
language and validation toolchain is simpler than adding Ruff, mypy, and their
configuration solely for the pack.

PR #3123 (merged 2026-09-01) established strict type-check and typed-lint coverage
for maintained Archon TypeScript scripts. The SDLC pack can use the same repository
machinery instead of retaining an unvalidated second language.

## Desired outcome

The bundled SDLC pack's deterministic scripts use TypeScript and Bun, and are
covered by the repository's strict type-check and typed-lint gates. They remain
owned and distributed by the pack, colocated under each pack component's
`scripts/` directory; validation reaches them in place rather than moving them
into the repository-level `.archon/scripts/` directory or a product package.

The conversion leaves the smallest maintainable implementation rather than
transliterating each Python file and function one-to-one. Shared shapes and
logic are derived where that reduces real duplication without obscuring
[...]

Issue #3043, opened 2026-08-31
design(workflows): the engine renders a terminal record at every terminal transition


Wirasm:
> **Re-scoped 2026-09-02 (operator decision).** The settled shape is the assessment comment below: a projection the engine assembles from durable material plus an artifact manifest, one additive JSON value in the terminal event, no new YAML, previews read at display time; #3156 is the prerequisite. An implementation plan posted on 2026-09-01 (`guarantee-terminal-records`: a `terminal_record.parts` YAML declaration with content previews assembled in every terminal writer) was removed as superseded so it cannot be mistaken for current; its rejected shape is summarized in that comment.

Split out of #2453 by operator decision, 2026-08-31. #2453 keeps the declared value contract between nodes, composed workflows, fan-out items, and context windows. This issue owns terminal truth: who selects a run's terminal result, and what guarantees the operator sees a report at all.

## Problem

A workflow's terminal report is computed by a node, and a node only runs if the DAG reaches it. So the runs that most need a report, the ones that die mid-flight, are exactly the ones that render none. The sdlc pack compensates with four hand-copied collector scripts (deliver/ship/stabilize/upkeep `outcome.py`) that reinterpret run state from bound values ("gate present with null delivered means delivery started and died"). That is lifecycle inference, which AGENTS.md assigns to the engine: terminal state transitions and completion signals belong to the engine, not to workflow scripts that reinterpret run state.

Four live datapoints:

- Run `f113a1fd`: an implement-stage discovery survived only because an operator read the artifacts directory by hand; consolidation ran only on the completion path.
- Run `1c225d25`: a failed run's rendered report reached no queryable surface. `workflow get` showed an empty output preview; the run's error field carried only per-node snippets; the report text existed only in the detached child's log file.
- PR #3000 (merged 2026-08-31, closing #2997): a delivery that died after opening its PR reported "delivered" because `returns: pr` survived the later failure. Fixed entirely inside the pack (`returns: flip-ready`), and the PR's own known-risk note records the residue: when the flip fails outright, the tail's outcome node is blocked and no report renders at all.
- The #2942 residual: standalone `archon-deliver` has no failure-path terminal report, because no node runs when the DAG fails around it.

Each new terminal nuance (#2939/#2940 failure branches, #3000's helper change) grows all four collector copies. The class-closing shape is an engine guarantee, not more branch-by-branch patching.

## Why now

[...]

PR #3148, opened 2026-09-02
feat(workflows): the producing node owns a workflow's result contract, and a result can point at an artifact

Wirasm:
## Problem and outcome

A workflow's result had no owner. A caller reading a child workflow's field had to repeat the child's schema on its own `workflow:` node, so two declarations had to agree by discipline, and a result that pointed at a large artifact file did so by naming convention that nothing verified. #2453 settled the design on 2 September: the node that produces a value owns its contract, in both execution models, and a result may carry a validated pointer to an artifact file.

- **Outcome:** a `workflow:` node's field access derives from the child's `returns:` node contract, which travels with the value and survives cold resume; a caller `output_format` on a `workflow:` node is a load error naming the child's node as the owner; a result value may carry `{ "type": "archon_artifact", "run_id", "path" }`, validated at the producing node against its own run before the value is persisted.
- **Invariant:** `include:` composition is unchanged, since flattening already made the producer the owner. A schemaless child keeps today's schemaless behavior. Fan-out aggregation is unchanged. The engine never expands a pointer to an absolute path and never reads the file.
- **Scope boundary:** no in-workflow pointer resolver, no cross-run reachability at the producer, no telemetry. The first in-engine reader of a pointer is #3043's terminal record; the read side owns reachability and realpath. The exec-node certification and load-time compile this builds on are #3158.

## Review guidance

- **Feedback requested:** the load error for caller `output_format` on a `workflow:` node withdraws PR #2788's runtime validation; confirm nothing shipped depends on it (no workflow in this repository declares it). And the producer-only pointer rule: own run id, lexical containment, existence, nothing more.
- **Start here:** `packages/workflows/src/artifact-pointer.ts` — the whole pointer rule, now store-free.
- **Review order:** `loader.ts` and `validator.ts` (the caller-schema load error) → `dag-executor.ts` (child projection to parent node, resume rehydration, producer-side pointer gates) → `artifact-pointer.ts` → `schemas/workflow-run.ts` and `core/src/db/workflow-events.ts` (`summary_declared_fields`, `declared_fields`) → `.archon/workflow-language-constitution.md` (the amended row) → docs → `e2e-contract-*` workflows.
- **Lower-attention areas:** docs and the two e2e reference workflows, which have no AI nodes.
- **Known risk or uncertainty:** the constitution row previously rejected cross-file schema checking outright; the amendment keeps the load-time rejection and states the runtime ownership rule. Read the row, not only the diff.

## Solution

The contract is the producing node's existing `output_format`; no new declaration surface. When a child run completes, the engine persists the top-level field names its `returns:` node declared beside the summary value, and the parent's `workflow:` node completes with that projection as its `declaredFields`, written to the `node_completed` event so a cold resume rehydrates it. A caller has nothing to assert, so the loader rejects `output_format` on a `workflow:` node with the owner named.

An artifact pointer is an ordinary JSON value with a reserved discriminator. Before a producer's value is persisted, the engine walks it for tagged objects and proves each one addresses a regular file under the current run's artifacts directory. Relays (the `workflow:` node in the parent, the fan-out aggregate) carry the value unrewritten. Reachability across runs and physical-path resolution belong to whichever run reads a pointer, on the adoption seam #2747 shipped; a producer-time realpath check is not a guarantee because the tree can change before a read.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | caller `output_format` on a `workflow:` node validated the child's value at runtime and defined the caller's field access | load error; the child's `returns:` node defines the caller's field access |
| **Failure behavior** | a result naming an artifact path was never checked | a tagged pointer to a missing file, another run, or an escaping path fails the producing node before `node_completed` |

## Validation

Run from the package directories on the final rebased tree (merge-base is #3158's head):

- `bun test src/dag-executor.test.ts` — 751 pass — child projection to the parent node, resume rehydration, producer-side pointer gates, the schemaless-child lenient read.
- `bun test src/loader.test.ts src/include-expander.test.ts src/fan-out-identity.test.ts` — 508 pass — the caller-schema load error, naming the child's `returns:` node.
- `bun test src/validator.test.ts` — 99 pass — the same error for in-memory definitions, and no issue for a loop_group's inert schema.
- `bun test src/artifact-pointer.test.ts` — 14 pass — own-run acceptance, other-run rejection, `..`, absolute, NUL, directory, missing file.
- `bun test src/subrun.test.ts` — 97 pass — a fan-out `workflow:` node with `output_format` refuses to load; the aggregate carries one pointer per child unrewritten.
- `bun test src/executor.test.ts` — 146 pass; `bun run test` in `packages/core` — 1957 pass (`declared_fields` on the resume snapshot).
- `bun test src/telemetry.test.ts` in `packages/paths` — 58 pass; the package is byte-identical to `dev`.
- `bun x tsc --noEmit`, `bun run lint`, `bun run check:bundled`, `bun run check:bundled-schema`, `bun run format:check` — green.
- Each new test was observed red with its guard neutralized (the ownership rejection returning null; the own-run branch disabled; the regular-file check inverted) and green with it restored.
- **Not verified:** `bun run validate` end to end on this branch; CI runs it.

## Links

- Related #2453 (settlement on the issue)
- Depends on #3158
- Related #3127 (pack adoption after merge)


Wirasm:
Plan for this PR, per the settlement on #2453 (https://github.com/coleam00/Archon/issues/2453#issuecomment-5512157879).

This branch splits in two. The first half, load-time schema compile plus exec-node certification with a typed non-retryable marker, is now #3158, cut from `dev`. This PR becomes the second half and will be rewritten on top of #3158: the callee-owned contract with a load error for a caller `output_format` on a `workflow:` node, the artifact pointer shape with producer-only validation, the constitution row amendment, docs, and the e2e workflows. Cut from it: the four telemetry counters, the caller-narrowing branch and mismatch check, `upwardClosure`, `RunLookup`, the hop cap, realpath and symlink resolution, and the relay-side pointer gates.

The branch will be force-pushed with the rewritten history once it is green locally; the inherited `main` commit (`12875f1d2`) goes with it.

Wirasm:
Rewritten history force-pushed (`f15a68255` → `7f2d39fa7`); the PR is based on `feat/exec-output-contract` (#3158) until that merges, so the diff shows only this half. Eight commits: the three original phases resolved against `dev`'s executor shape, then producer ownership with the caller-schema load error, own-run-only pointer validation, the constitution row and docs, and two review-correction commits.

Cut, per the settlement on #2453: the four telemetry counters (`packages/paths` and the configuration reference are byte-identical to `dev`), the caller-narrowing branch and mismatch check, `upwardClosure`, `RunLookup`, the hop cap, realpath and symlink resolution, and the relay-side pointer gates. PR #2788's runtime caller-side validation is withdrawn by the load error.

Adversarial review before the push (independent agent, against the pinned head): the load error is total across file parsing, in-memory validation, loop-group bodies, include expansion, and dynamic fan-out; the persisted `summary_declared_fields` and `declared_fields` degrade safely on corrupt data; pointer validation is producer-only with no relay re-validation; every cut identifier is gone. Its blocking finding was a documentation overclaim that the artifact read route already resolves real paths; it does lexical containment only, so the docs now say so and the gap is filed as #3160. Its two other findings, a stale comment in `output-ref.ts` and no test of a pointer crossing a fan-out aggregate, are fixed in the last two commits; the fan-out coverage was seen red through a real CLI run of `e2e-contract-fanout` with the pointer field dropped.

Wirasm:
CI note: `test.yml` runs on pull requests targeting `main` or `dev` only, so while this PR is based on `feat/exec-output-contract` the Test Suite does not trigger on it (only the docs build and CodeRabbit did). The suite was started by hand on this exact head with `workflow_dispatch`: https://github.com/coleam00/Archon/actions/runs/33659276082. Once #3158 merges and GitHub retargets this PR onto `dev`, the ordinary trigger takes over.

Wirasm:
The dispatched Test Suite on this head (`7f2d39fa7`) passed: test (ubuntu-latest), test (windows-latest), docker-build, postgres-parity, schema-upgrade all green; `workflow-fixtures` is skipped on a manual dispatch (no change range), and the pack fixtures are untouched by this PR. https://github.com/coleam00/Archon/actions/runs/33659276082

PR #3158, opened 2026-09-02
feat(workflows): a bash or script node certifies its own result contract

Wirasm:
## Problem and outcome

A `bash:` or `script:` node could declare `output_format`, but the field was dropped with a load warning, so a deterministic script could not own a result contract and could not be selected by `outcome_field`. That is why `archon-ship`, `archon-upkeep`, and `archon-deliver` report `outcome: null` today: their `returns:` node is a script. Separately, a schema ajv could not compile was a runtime warning after the provider had already been paid, which silently turned a declared contract into no contract.

- **Outcome:** an exec node with `output_format` certifies its own stdout (one strict JSON document, validated by the shared ajv gate) and exposes the same `output`, `structuredOutput`, and `declaredFields` triple an agent node does; an uncompilable `output_format` on any node is a load error naming the node; a contract failure is never retried.
- **Invariant:** exec nodes without `output_format` are byte-for-byte unchanged. Agent-node structured output is unchanged. `output_format` stays inert, with the existing warning, on `loop_group`, gates, halts, and includes.
- **Scope boundary:** no callee-owned contract across `workflow:` boundaries and no artifact pointers. Those are the second half of #2453 and follow in #3148, rewritten on top of this PR per the settlement on the issue.

## Review guidance

- **Feedback requested:** correctness of the certification path and the non-retryable marker.
- **Start here:** `packages/workflows/src/dag-executor.ts` `certifyExecOutput` — the one place stdout becomes a certified value, shared by both exec runtimes.
- **Review order:** `structured-output.ts` (`compileOutputSchema`, `$id` deregistration) → `loader.ts` `validateNodeOutputFormats` → `dag-node.ts` (`output_format` leaves `BASH_NODE_AI_FIELDS` and survives the transform) → `dag-executor.ts` (certify, both catch sites, `shouldRetryNodeFailure`) → `schemas/workflow-run.ts` (`retryable: false`) → tests → docs.
- **Lower-attention areas:** docs under `packages/docs-web`; the schema test additions.
- **Known risk or uncertainty:** a workflow that declared `output_format` on an exec node and printed prose now fails that node instead of warning. No workflow in this repository did.

## Solution

The contract is the node's existing `output_format`; nothing new is declared. The loader compiles every declared schema at parse time through the same ajv cache the runtime reads, so an unenforceable contract is rejected before any spend and the four runtime warn-and-continue branches become node failures (reaching them means a definition bypassed the loader). The exec executor parses stdout strictly, validates it, and returns the canonical text plus the logical value plus the field projection, so `$node.output.field` on a script is as strict as on an agent.

A contract failure carries up to 200 characters of the script's stdout in its error text. The retry loop classifies error text, so a transient-looking excerpt would have re-run a script whose stdout is deterministically wrong. The failed output now carries `retryable: false` in the type, and `shouldRetryNodeFailure` honours it before any text classification. Only `false` is expressible: a producer can refuse a retry, never force one past a FATAL classification.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `output_format` on an exec node warned at load and was ignored | stdout must be one JSON document satisfying the schema; `$node.output.field` works on the certified value |
| **Failure behavior** | uncompilable schema: runtime warning, node completes unenforced | uncompilable schema: load error naming the node; stdout mismatch: node fails, never retried |

## Validation

- `bun test src/dag-executor.test.ts` in `packages/workflows` — green — exec certification for bash and script, strict field access, cold resume, timeout look-alike, and the no-retry regression (the retry guard was observed red with the check neutralized: the node ran three times).
- `bun test src/loader.test.ts src/include-expander.test.ts src/fan-out-identity.test.ts`, `bun test src/schemas.test.ts …`, `bun test src/validator.test.ts` — green — load-time compile rejection and the field-list changes.
- `bun test src/shared/structured-output.test.ts` in `packages/providers` — green — `compileOutputSchema` and `$id` deregistration.
- `bun x tsc --noEmit` in `packages/workflows`, `bun run lint`, `bun run check:bundled`, `bun run check:bundled-schema` — green.
- **Not verified:** `bun run validate` end to end on this branch; CI runs it.

## Links

- Related #2453 (settlement recorded on the issue)
- Related #3148 (second half, rewritten on top of this PR)
- Related #3127 (pack adoption after merge)

## Summary by CodeRabbit

* **New Features**
  * Bash and script nodes now support `output_format` contracts for strict JSON stdout validation.
  * Validated results are published as canonical structured outputs for downstream nodes and workflow conditions.
  * Output schemas are checked when workflows load, with clear errors for invalid schemas.

* **Bug Fixes**
  * Invalid outputs and schema mismatches now fail nodes without retrying.
  * Failure previews redact credentials and structured-output validation no longer continues with warnings.

* **Documentation**
  * Added guidance for declaring and using output contracts on deterministic nodes.


Wirasm:
Review corrections pushed in one commit, each finding verified first and each new test seen red with its guard removed:

- **ajv registry leak on a failed compile** (Major): `compileAndCache` deregisters in a `finally`; regression test in `structured-output.test.ts` compiles a broken-then-fixed schema sharing one `$id`.
- **Inert loop-group schema compiled at load** (Major): the loader compiles only enforced schemas (agent, exec, `loop:`) and still walks loop_group bodies; regression test in `loader.test.ts`.
- **Unredacted stdout preview in `node_failed.data.error`** (Minor, CWE-532): `runSubprocess` returns its resolved credential set and the preview redacts the whole stdout before slicing; regression test in `dag-executor.test.ts`.
- **Docs** (two Minors): byte-for-byte claim removed; non-retryable contract failures stated in the authoring guide.

PR #3144, opened 2026-09-02
Fork-PR launches no longer fail on concurrent Git ref-lock races

Wirasm:
## Problem and outcome

`createFromForkPR` performs fetches that #3065 and #3091 already protected on the `syncWorkspace` and same-repo PR paths, but its no-SHA fork fetch branch was unguarded: two concurrent fork-PR launches against the same upstream both write the local branch `refs/heads/pr-<n>-review` via the `pull/<n>/head:pr-<n>-review` refspec and collide on Git's shared ref lock, failing with `cannot lock ref ... unable to update local ref`.

- **Outcome:** Concurrent fork-PR launches no longer fail transiently; the no-SHA fetch absorbs the race with the same bounded retry that already protected `syncWorkspace`. The codebase now has exactly one ref-lock classifier and one backoff budget, in `@archon/git`.
- **Invariant:** Unrelated git failures are attempted once and stay loud; a persistent contention failure exhausts the bounded budget (4 attempts total) and reports the original error; every failure path still fails before an isolation environment or worktree is created.
- **Scope boundary:** The with-SHA fork fetch is intentionally left unguarded (evidence below), and `syncWorkspace`'s public signature is unchanged.
- **Root cause:** The no-SHA refspec writes a named local ref, so concurrent fetches contend on Git's shared lock file for `refs/heads/pr-<n>-review`. This matches the issue body's error signature, but corrects its claim of a shared remote-tracking ref: the contended ref is the local review branch.

### The with-SHA form cannot exhibit this race

`git fetch <remote> pull/<n>/head` (no colon refspec) updates only `FETCH_HEAD` and touches no named ref, so there is no ref lock to contend on. The issue's second fetch branch therefore needs no retry, and the call site carries a comment explaining why. This PR does not silently drop that acceptance item — it resolves it by evidence.

## Review guidance

- **Feedback requested:** Correctness of the retry composition with stale-branch handling, and whether the shared-helper extraction (vs. generalizing `syncWorkspace`) is the right seam.
- **Start here:** `packages/git/src/repo.ts:130` — `fetchWithRefLockRetry` is the single owner of the classifier (`cannot lock ref` + `unable to update local ref`) and the backoff budget (3 retries, 50ms doubling). On exhaustion it rethrows the original error object, preserving stderr.
- **Review order:** `packages/git/src/repo.ts` → `packages/isolation/src/providers/worktree.ts` (`createFromForkPR`) → the two test files.
[...]

PR #3162, opened 2026-09-02
fix(workflows): give fixture discovery a defined order

### LINKED ISSUE #3161: test(workflows): fixture discovery has no defined order, so an exact-label assertion fails at random
## Problem

Fixture discovery has no defined order, and that order reaches both the report an operator reads and an exact assertion in the test suite.

`walkForFixtures` pushes results in raw `readdir` order (`packages/workflows/src/fixture-runner.ts:191,202`) and nothing sorts afterwards. `readdir` returns filesystem order, which varies by machine and filesystem, so `report.results` can come back in any order for the same tree.

`packages/workflows/src/fixture-runner.test.ts:760` asserts an exact ordered array against it:

```ts
await expect(fixtureLabels('sdlc')).resolves.toEqual([
  'sdlc/plan/fixtures/plan.stubs.yaml',
  'sdlc/ship/fixtures/ship.stubs.yaml',
]);
```

So the test can fail on any PR at any time, and it did — twice on 2 September 2026, on two unrelated PRs, both times with the same two fixtures simply swapped:

```
  [
-   "sdlc/plan/fixtures/plan.stubs.yaml",
    "sdlc/ship/fixtures/ship.stubs.yaml",
+   "sdlc/plan/fixtures/plan.stubs.yaml",
  ]
```

## Why

This reads as a flaky test, but the test is the symptom. A fixture report that lists its fixtures in a different order on each machine is worse to read and impossible to diff between runs, and the ordering assertion is the only thing that noticed.

Loosening the assertion to a set comparison would hide that, and would remove the check that a `sdlc` target does not pick up the prefix-named `sdlc-ext` sibling — which is what those exact labels are there to pin.

## Why now

It cost two red CI runs and two re-runs in one afternoon, on PRs that touch neither workflows nor fixtures. Frequency on unrelated PRs is exactly what makes a real defect get "fixed" by weakening its test.

## Desired outcome

Fixture discovery has one defined order, so the same tree always produces the same report, and the existing exact-label assertions become honest as written.

## Invariants

- Scope precedence is unchanged: a higher-precedence scope still shadows a lower one, keyed on the scope-relative label.
- Discovery still finds the same set of fixtures; only their order becomes defined.

## Acceptance

- [ ] A test creates workflow directories in reverse-alphabetical order and asserts the report comes back in label order. It is observed red against the unordered implementation before it is trusted.
- [ ] The existing exact-label assertions at `fixture-runner.test.ts:760` pass without being loosened to a set comparison.
- [ ] `bun run validate` passes.

## Evidence

Verified against `dev` at `f5d3ca0a4`:

- `packages/workflows/src/fixture-runner.ts:191` — `readdir(root, { withFileTypes: true })`, iterated in returned order
- `packages/workflows/src/fixture-runner.ts:202` — `readdir(dirPath)`, iterated in returned order
- `packages/workflows/src/fixture-runner.ts:217-244` — `discoverFixtures` never sorts
- `packages/workflows/src/fixture-runner.test.ts:760` — the exact ordered assertion

### Environment

- Archon version or commit: `dev` at `f5d3ca0a4`
- Platform or adapter: `@archon/workflows` fixture runner
- OS: observed on ubuntu-latest in CI; the order is filesystem-dependent, not OS-specific

- Must remain true: the `sdlc` / `sdlc-ext` prefix boundary stays pinned by exact labels.
- Solution steering: hint — define the order at discovery rather than sorting at each call site, so there is one place that owns it.

Wirasm:
## Problem and outcome

Fixture discovery had no defined order. `walkForFixtures` appended results in raw `readdir` order and nothing sorted afterwards, so `report.results` came back differently on different machines — and `fixture-runner.test.ts` asserts exact ordered labels against it. That assertion failed twice on 2 September on two unrelated PRs (#3144 and #3159, neither touching workflows), both times with the same two fixtures simply swapped.

- **Outcome:** the same tree always produces the same report, in label order, and the existing exact-label assertions are honest as written rather than accidentally passing.
- **Invariant:** scope precedence is unchanged — a higher-precedence scope still shadows a lower one, keyed on the scope-relative label. Discovery finds the same set of fixtures; only their order becomes defined.
- **Scope boundary:** no change to what is discovered, to target resolution, or to fixture execution.
- **Root cause:** `readdir` returns filesystem order, and `packages/workflows/src/fixture-runner.ts` passed it straight through to the report.

## Review guidance

- **Feedback requested:** whether sorting belongs at discovery (as here) or at the report boundary, and whether the rewritten dedup reads more clearly than the version it replaces.
- **Start here:** `packages/workflows/src/fixture-runner.ts` — `fixturesUnderScope`, the one place that now defines order.
- **Review order:** `fixturesUnderScope` and `walkForFixtures`, then `discoverFixtures`, then the new test and the `writeWorkflowDirs` helper.
- **Lower-attention areas:** `writeNestedPackProject` collapses from 26 lines to 5 by calling the new helper; the 50 existing tests in the file pass unchanged.

## Solution

Sorting one line onto the end of the walk would have fixed the symptom. The reason the order was undefined in the first place is the shape, so this changes that instead — the fix is a consequence.

`walkForFixtures` **returns** its results rather than appending to a caller-owned `out` array, so a caller can no longer depend on append order. `discoverFixtures` then filters each scope's results through a seen-set instead of splicing the shared array while indexing it with a manual `i--`, which is what coupled the two functions through `before = found.length`. `fixturesUnderScope` sorts by label — one place, named, with the invariant written down.

Two smaller consequences fall out: `workflowNamesBeside` is now called once per `fixtures/` directory rather than once per fixture file in it, since the sibling YAML it parses is the same for every file in that directory; and its own result is sorted, because those names reach the operator in a "no discovered workflow matches" failure.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Fixture report order varied by machine and filesystem | Fixture report is ordered by label everywhere |
| **Failure behavior** | An exact-label assertion failed at random on unrelated PRs | The assertion pins a defined contract |

## Validation

- `cd packages/workflows && bun run test` — 156 pass, 0 fail — the 50 existing fixture-runner tests pass unchanged alongside the new guard.
- The new guard was observed red before being trusted, and the state space closed around it: with the sort removed it fails on this machine naturally (this filesystem returns creation order, and the test creates `zulu`, `mike`, `alfa` in that deliberately reverse-alphabetical order); with the sort mutated to descending it also fails. It passes only on ascending label order.
- **Not verified:** that CI's filesystem returns a non-alphabetical order for this specific tree. It does not need to — the guard now pins the order the code defines rather than the order a filesystem happens to yield, which is the whole point.

## Links

- Closes #3161

## Summary by CodeRabbit

- **Bug Fixes**
  - Workflow fixtures are now discovered and displayed in a consistent, deterministic order.
  - Duplicate fixture labels across workflow scopes are no longer reported multiple times.
  - Errors encountered while reading fixture directories are now surfaced instead of being treated as missing fixtures.
  - Workflow names in “no matching workflow” messages are listed alphabetically.

- **Tests**
  - Added coverage for fixture ordering, error handling, and shared workflow fixture setup.

### DIFF
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -769,17 +769,24 @@ describe('runFixtures', () => {
   // R1: this fixtures/ dir was already found by the walk, so a read failure is a real
   // fault, not an absence. Swallowing it would let a run exit 0 having skipped a
   // directory it was asked to certify.
-  it('propagates a fixtures-directory read failure instead of reporting no fixtures', async () => {
-    const cwd = makeTempProject();
-    writeWorkflowDirs(cwd, ['pack/locked']);
-    const fixturesDir = join(cwd, '.archon', 'workflows', 'pack', 'locked', 'fixtures');
-    chmodSync(fixturesDir, 0o000);
-    try {
-      await expect(runFixtures({ workflows: [], cwd })).rejects.toThrow();
-    } finally {
-      chmodSync(fixturesDir, 0o755);
+  //
+  // POSIX-only: the unreadable directory is created with chmod 000, which win32 does
+  // not enforce for directory reads — the precondition cannot be built there, so the
+  // test would assert nothing. Same treatment as the permission cases in git.test.ts.
+  it.skipIf(process.platform === 'win32')(
+    'propagates a fixtures-directory read failure instead of reporting no fixtures',
+    async () => {
+      const cwd = makeTempProject();
+      writeWorkflowDirs(cwd, ['pack/locked']);
+      const fixturesDir = join(cwd, '.archon', 'workflows', 'pack', 'locked', 'fixtures');
+      chmodSync(fixturesDir, 0o000);
+      try {
+        await expect(runFixtures({ workflows: [], cwd })).rejects.toThrow();
+      } finally {
+        chmodSync(fixturesDir, 0o755);
+      }
     }
-  });
+  );
 
   it('reports fixtures in label order, not filesystem order', async () => {
     const cwd = makeTempProject();
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -3,6 +3,7 @@ import { describe, it, expect, beforeAll, afterAll, mock } from 'bun:test';
 import {
   cpSync,
   existsSync,
+  chmodSync,
   mkdirSync,
   mkdtempSync,
   readdirSync,
@@ -745,6 +746,41 @@ describe('runFixtures', () => {
   // `readdir` yields filesystem order, which varies by machine and filesystem. Two CI
   // runs on unrelated PRs failed here with the same fixtures in swapped order before
   // discovery defined one. Creation order below is deliberately reverse-alphabetical.
+  // R4: the sort in workflowNamesBeside only matters when one fixtures/ dir has two
+  // sibling workflow YAMLs — otherwise a single name hides any ordering. These names
+  // reach the operator in a "no discovered workflow matches" failure.
+  it('orders the sibling workflow names beside one fixtures dir', async () => {
+    const cwd = makeTempProject();
+    const dir = join(cwd, '.archon', 'workflows', 'pack', 'two');
+    mkdirSync(join(dir, 'fixtures'), { recursive: true });
+    // Created zulu-first so filesystem order is the wrong answer.
+    writeFileSync(join(dir, 'zulu-wf.yaml'), stubWorkflowYaml('zulu-wf'));
+    writeFileSync(join(dir, 'alfa-wf.yaml'), stubWorkflowYaml('alfa-wf'));
+    writeFileSync(join(dir, 'fixtures', 'two.stubs.yaml'), STUB_FIXTURE);
+
+    // No workflow is passed in, so the fixture matches nothing and the failure names
+    // every sibling candidate — the surface these names actually reach.
+    const report = await runFixtures({ workflows: [], cwd });
+
+    expect(report.results).toHaveLength(1);
+    expect(report.results[0].failureReason).toContain('alfa-wf, zulu-wf');
+  });
+
+  // R1: this fixtures/ dir was already found by the walk, so a read failure is a real
+  // fault, not an absence. Swallowing it would let a run exit 0 having skipped a
+  // directory it was asked to certify.
+  it('propagates a fixtures-directory read failure instead of reporting no fixtures', async () => {
+    const cwd = makeTempProject();
+    writeWorkflowDirs(cwd, ['pack/locked']);
+    const fixturesDir = join(cwd, '.archon', 'workflows', 'pack', 'locked', 'fixtures');
+    chmodSync(fixturesDir, 0o000);
+    try {
+      await expect(runFixtures({ workflows: [], cwd })).rejects.toThrow();
+    } finally {
+      chmodSync(fixturesDir, 0o755);
+    }
+  });
+
   it('reports fixtures in label order, not filesystem order', async () => {
     const cwd = makeTempProject();
     writeWorkflowDirs(cwd, ['pack/zulu', 'pack/mike', 'pack/alfa']);
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -77,6 +77,27 @@ import {
 
 /** The file's one temp-root creator, so every fixture tree is tracked for teardown. */
 const trackTempRoot = trackTempRoots();
+const STUB_FIXTURE = ['fixture:', '  expect: completed', 'node-a: "stub output"', ''].join('\n');
+
+const stubWorkflowYaml = (name: string) =>
+  `name: ${name}\ndescription: test\nnodes:\n  - id: node-a\n    prompt: hello\n`;
+
+/**
+ * Write one workflow and its fixture per slash-separated path, so a tree is declared
+ * rather than assembled: `'sdlc/plan'` produces
+ * `.archon/workflows/sdlc/plan/plan-wf.yaml` and `.../fixtures/plan.stubs.yaml`.
+ */
+function writeWorkflowDirs(cwd: string, paths: readonly string[]): void {
+  for (const path of paths) {
+    const segments = path.split('/');
+    const leaf = segments[segments.length - 1];
+    const dir = join(cwd, '.archon', 'workflows', ...segments);
+    mkdirSync(join(dir, 'fixtures'), { recursive: true });
+    writeFileSync(join(dir, `${leaf}-wf.yaml`), stubWorkflowYaml(`${leaf}-wf`));
+    writeFileSync(join(dir, 'fixtures', `${leaf}.stubs.yaml`), STUB_FIXTURE);
+  }
+}
+
 function makeTempProject(prefix = 'fixture-runner-'): string {
   return trackTempRoot(mkdtempSync(join(tmpdir(), prefix)));
 }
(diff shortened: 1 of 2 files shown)


Wirasm:
## Needs fixes

The ordering fix itself is correct and well-evidenced: `fixturesUnderScope` sorts each scope's discovered fixtures by label, `discoverFixtures`'s dedup rewrite (splice-loop → seen-set `continue`/`push`) is a faithful, test-confirmed translation of the logic it replaced, and scope precedence (project shadows global shadows bundled, keyed on label) is unchanged — the pre-existing, untouched shadow test still passes. The new reverse-alphabetical-creation guard genuinely discriminates: independently confirmed on this machine that removing the sort makes it fail with the two labels swapped, exactly as claimed.

One correction is required before merge: the refactor from index-mutating recursion to return-based recursion introduced a new silent fallback (`readdir(fixturesDir).catch(() => [])`) where the pre-PR code let the equivalent failure propagate as a visible, CLI-reported error. Two reviewers (code, seams) independently found this on the same line. Three suggestions round out the review: the report is deterministic but only sorted within each precedence scope rather than across the merged list (the PR's own review guidance explicitly asks this question), a dead `.sort()` sits three lines above the sort that actually matters, and that real sort has no regression test guarding the one input shape it exists for.

**1 blocking · 3 non-blocking**

**Validation:** `bun run validate` passes (exit 0); `packages/workflows` package tests: 156 pass, 0 fail, matching the PR's claim.

### Findings

|---|---|---|---|
| `R1` | Important | New silent fallback swallows a `fixtures/`-dir read failure that previously propagated as a visible error | OPEN |
| `R2` | Suggestion | Report order is deterministic per scope block, not globally sorted across the merged catalog | OPEN |
| `R3` | Suggestion | Dead `.sort()` on `readdir(parent)` in `workflowNamesBeside` has no observable effect | OPEN |
| `R4` | Suggestion | `workflowNamesBeside`'s real determinism sort is untested for the multi-workflow-per-directory case | OPEN |

<details>
<summary><code>R1</code> — Silent fallback where the pre-PR code failed loudly</summary>

**Impact:** `workflow test` can exit 0 having silently skipped a fixtures directory it already found, instead of certifying it — a false-success outcome for a tool whose whole purpose is to certify dry-run behavior faithfully (the file's own docstring: "the only execution path is `dryRunWorkflow`... so a fixture can never trigger a real run or provider call").

**Evidence:** `packages/workflows/src/fixture-runner.ts:195` — `const files = await readdir(fixturesDir).catch(() => []);`. Confirmed against the pre-PR source (`dev` at `d46f86bd9`): the equivalent call, `for (const file of await readdir(dirPath))` inside the old `walkForFixtures`, had no local catch — only the *outer* per-directory listing (`readdir(root, {withFileTypes:true})`) was guarded. A failure reading a `fixtures/` directory the walk had already identified propagated uncaught out of `runFixtures()`. The CLI wraps `runFixtures(...)` in try/catch (`packages/cli/src/commands/workflow.ts:1198-1230`) and prints `Error: <message>` / exits 1 on any thrown error — that visible failure path is now bypassed for this specific case. No log call accompanies the new `.catch`, unlike the file's other tolerated-failure path (`workflowNamesBeside`'s `getLog().debug(...)` a few lines above).

Note: the *outer* walk-level `readdir(dir, {withFileTypes:true}).catch(() => [])` at line 213 is not a new fallback — it's a faithful translation of the pre-PR try/catch-return around the recursive walk. Only the inner, `fixtures/`-directory-specific read at line 195 is new.

**Required outcome:** Either drop the `.catch(() => [])` at line 195 so this failure propagates as it did before (matches the file's existing asymmetric design: tolerate an unreadable directory in general recursion, but treat a `fixtures/` dir the walk already committed to reading as a specific, actionable failure), or if a silent skip is deliberately wanted, log it via `getLog().warn(...)` with a comment explaining why, matching the sibling pattern already in this file.

**Found by:** `code`, `seams`

**Disposition:** OPEN

</details>

<details>
<summary><code>R2</code> — Report order is per-scope, not globally sorted</summary>

**Impact:** None today — the report is fully deterministic and satisfies the issue's stated invariants. But when a project and a bundled (or global) scope both contribute fixtures, `report.results` is alphabetical only *within* each scope's block; a project-scope fixture labeled `zzz/...` still sorts before a bundled-scope fixture labeled `aaa/...`. This repo's own layout (a typical project's fixtures alongside the 8-pack bundled `sdlc` catalog) is exactly the common multi-scope case, not a rare edge.

**Evidence:** `fixturesUnderScope` (`fixture-runner.ts:227-237`) sorts within one scope root; `discoverFixtures` (`fixture-runner.ts:239-260`) appends each scope's already-sorted block in precedence order (project, global, bundled) with no final merge sort. The docstring on `fixturesUnderScope` ("Fixtures under one scope root, ordered by label... Sorting here is the single point that defines it; nothing downstream reorders") is accurate only per scope, not for the merged report a human reads.

**Required outcome:** No correction required — the PR's own "Review guidance" section explicitly invites this exact question ("whether sorting belongs at discovery... or at the report boundary"). This is a decision for the author, not a defect: sorting the merged `found` list once at the end of `discoverFixtures` (same comparator, after all roots are combined) would make the whole report globally alphabetical without weakening scope-precedence dedup, which is independent of final ordering.

**Found by:** `code`

**Disposition:** OPEN

</details>

<details>
<summary><code>R3</code> — Dead <code>.sort()</code> in <code>workflowNamesBeside</code></summary>

**Impact:** None on behavior — purely a readability/maintenance trap. Implies the iteration order into the name-collection loop matters when it provably does not, since the real determinism guarantee comes entirely from the final `[...new Set(names)].sort()` three lines later.

**Evidence:** `packages/workflows/src/fixture-runner.ts:164` — `for (const entry of (await readdir(parent)).sort()) {`. The function returns `[...new Set(names)].sort()` at line 182; that sort alone fully determines the returned array regardless of the order `names` was populated in. The line-164 sort only reorders which filename hits a debug log on a parse failure — a side channel nothing reads for ordering.

**Required outcome:** Drop the `.sort()` at line 164 (`for (const entry of await readdir(parent))`); the sort at line 182 already satisfies the determinism this PR exists to add.

**Found by:** `simplify`

**Disposition:** OPEN

</details>

<details>
<summary><code>R4</code> — The sort that actually matters in <code>workflowNamesBeside</code> has no regression test</summary>

**Impact:** A future edit that removes or weakens the real sort at `fixture-runner.ts:182` — plausible precisely because the dead sort at line 164 (R3) sits right next to it and looks like it already does the job — would silently reintroduce filesystem-order leakage into operator-facing failure messages (`no discovered workflow matches (...)`) and into multi-target result row ordering. Every currently-passing test (156/156) would still pass, because none constructs a `fixtures/` directory with more than one sibling workflow YAML.

**Evidence:** `fixture-runner.ts:158-183` (`workflowNamesBeside`); confirmed via `find .archon/workflows -name fixtures` that no test or bundled fixture directory today exercises the multi-sibling-workflow case, so the sort's only observable effect is currently unprotected.

**Required outcome:** Add a test with one `fixtures/` directory beside two sibling workflow YAML files (declared in reverse-sort order), asserting the resulting `failureReason` or `report.results` ordering is alphabetical rather than creation-order — the same reverse-order-creation technique the PR's own new guard already uses.

**Found by:** `tests`

**Disposition:** OPEN

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1, R2 |
| `seams` | R1 |
| `simplify` | R3 |
| `tests` | R4 |

#### Validation

|---|---|---|
| `bun run validate` (repo root, dedicated worktree) | PASS | exit 0 after `bun install`; includes type-check, lint, format, and every package's test suite |
| `bun run test` (inside `packages/workflows`) | PASS | 156 pass, 0 fail, matching the PR's claimed count |
| Coordinator-run: remove the sort in `fixturesUnderScope`, rerun the new guard test alone | FAIL as expected | `reports fixtures in label order, not filesystem order` fails with `alfa`/`zulu` swapped — confirms the guard discriminates a real regression rather than passing incidentally, and reproduces the PR description's own claimed observation |

</details>

Wirasm:
## Round 1 corrections — R1, R3, R4 fixed; R2 answered

Pushed as `9f1dbe00a`.

### R1 (Important) — fixed

Confirmed against `dev` before acting: pre-PR the fixtures-directory read was a bare `await readdir(dirPath)` and propagated; the rewrite added `.catch(() => [])`. That is a regression I introduced, and the review is right that it is the blocking one — the walk has already seen the directory, so a read failure there is a fault, not an absence, and swallowing it let `workflow test` exit 0 having skipped a directory it was asked to certify.

The read propagates again, with the reason recorded at the call site. Note the *outer* walk's `.catch(() => [])` is not the same case: pre-PR that read was already wrapped in `try/catch { return }`, so it is a faithful translation and stays.

New guard drives a `chmod 000` fixtures directory through `runFixtures` and asserts it rejects. Observed red against the `.catch(() => [])` it forbids.

### R3 (Suggestion) — fixed

Correct, the `.sort()` on `readdir(parent)` was dead: the result is sorted after dedup three lines later, so loop order never reached the return value. Removed.

### R4 (Suggestion) — fixed

The surviving sort at `workflowNamesBeside` had no test, and R3 made it look like removable noise. Added a case with one `fixtures/` directory and two sibling workflow YAMLs, created zulu-first so filesystem order is the wrong answer, asserting on the `no discovered workflow matches (…)` failure — the surface those names actually reach. Observed red with the sort removed. The comment at the sort now says why it exists.

### R2 (Suggestion) — answered, keeping within-scope ordering

Correct that the report is sorted within each precedence scope, not globally: a project-scope `zzz/…` still precedes a bundled-scope `aaa/…`.

Keeping it. Scope order is a real property of the result — the list is "project fixtures, then global, then bundled", and each scope is internally ordered. A global sort would interleave scopes and lose that grouping to gain alphabetical scanning, which is a product decision about report presentation rather than the determinism defect #3161 describes. #3161's outcome — same tree, same report — is met either way, and I would rather not widen a fix into a presentation change.

Happy to be overruled; it is a one-line move of the sort into `discoverFixtures` if you prefer the flat ordering.

### Validation

- `bun run validate` — exit 0
- `packages/workflows` — 158 pass / 0 fail (was 156; two new guards)
- Both new guards observed red against the exact mutation each forbids, then green

Wirasm:
## Follow-up: the R1 guard was POSIX-only

`715c4507d`. The guard I added for R1 builds its precondition with `chmod 000`, which win32 does not enforce for directory reads — the directory stayed readable, the test asserted nothing, and windows-latest went red.

Now skipped on win32 with the reason stated at the test, matching the existing permission cases in `git.test.ts:2768` and `credential-helper-install.test.ts:84`. Re-verified that it still fails against the `.catch(() => [])` it forbids on POSIX.

All checks green on `715c4507d`; merge state CLEAN.

Head for a round-2 review: `715c4507de7da3794befe442d9679bcdea0400cc` (round 1 reviewed `5316c124b`).

PR #3177, opened 2026-09-03
fix(core): take the next free port when a worktree's hashed port is occupied

### LINKED ISSUE #3130: Worktree startup fails when its auto-selected port is occupied
## Problem

When Archon starts from a worktree, it automatically selects a server port from the worktree path. If another worktree or an unrelated local process already occupies that port, Archon fails to bind and startup stops with `EADDRINUSE`.

## Why

Worktrees are used for parallel development and agent runs. A collision in the automatically selected port range should not make an otherwise valid Archon instance unusable or require the operator to diagnose an opaque startup failure.

## Desired outcome

Archon remains available when a worktree's automatically selected port is already occupied. Startup should complete on a usable endpoint and clearly report where the server is listening.

## Acceptance

- [ ] Starting Archon in a worktree does not stop solely because its automatically selected port is already occupied.
- [ ] The running server's actual endpoint is clearly reported to the operator and any Archon-owned consumer that needs it.
- [ ] The occupied-port behavior is covered for a collision with another local listener.
- [ ] When the automatically selected port is available, existing worktree startup behavior remains unchanged.
- [ ] An explicit `PORT` override remains an explicit operator choice; this issue does not define fallback behavior for an occupied override.

### Reproduction

1. Determine the server port Archon automatically selects for a worktree.
2. Bind another local listener to that port.
3. Start Archon from the worktree without setting `PORT`.
4. Observe that server startup stops with `EADDRINUSE`.

### Environment

- Platform or adapter: local server startup from a git worktree
- OS: not OS-specific

- Must remain true: A worktree keeps its existing automatically selected port when that port is available.
- Related PR: #3129 contains contributor evidence and a proposed implementation, but this issue intentionally defines the outcome rather than prescribing that approach.
- Related issue: #2668 concerns ports used by subprocesses launched inside workflow runs. That is a separate boundary from Archon's own server startup.

alixnaveed:
## Problem and outcome

A worktree's server port is an MD5 hash of its path folded into 900 slots and returned without checking whether anything is listening on it. Two worktrees hashing to the same slot — or any unrelated process holding a port in 3190–4089 — makes the second server fail to bind with a bare `EADDRINUSE` that names neither the cause nor the `PORT` override that resolves it. Parallel worktrees are the workflow the README leads with, and 900 slots is a small space: collisions reach roughly even odds around 35 worktrees, and about 5% at ten.

- **Outcome:** A worktree whose hashed port is taken starts on the next free port in the range and logs `worktree_port_reallocated` with both the port it wanted and the one it got, instead of failing to start.
- **Invariant:** The hashed port is still what a worktree gets whenever it is free — "same worktree, same port" survives — and an explicit `PORT` is still bound verbatim or not at all.
- **Scope boundary:** The base port (3090) is deliberately still unprobed, and the Vite dev proxy is untouched. See *Known risk or uncertainty*.
- **Root cause:** `getPort()` computed a port and returned it. There was no availability check anywhere between the hash and `Bun.serve`.

## Review guidance

- **Feedback requested:** Correctness of the scan, and whether the base-port carve-out is the call you want.
- **Start here:** `packages/core/src/utils/port-allocation.ts` — `resolveWorktreePort()`. Everything else follows from it.
- **Review order:** `port-allocation.ts` (`isPortAvailable` → `getPort` → `resolveWorktreePort`), then `port-allocation.test.ts`, then the `dx-quirks.md` note.
- **Lower-attention areas:** The pre-existing `calculatePortOffset` tests are unchanged; the magic numbers they assert are now named constants (`BASE_PORT`, `MIN_OFFSET`, `OFFSET_SPAN`) with identical values.
- **Known risk or uncertainty:** Two, both deliberate and documented at the branch:

  1. **The probe is racy by construction.** Another process can take the port between `isPortAvailable()` and `Bun.serve`. Closing that window means holding the socket from allocation through to the server, which this function does not own. The probe removes the common case — a port already held when the server starts — and leaves the narrow one to the bind error, which is no worse than today.
  2. **The base port is not probed, on purpose.** Unlike a worktree port, 3090 is resolved independently elsewhere: `packages/web/vite.config.ts` falls back to it verbatim when `PORT` is unset. Silently serving from somewhere else would leave the dev proxy pointing at nothing. I think that carve-out is right, but it is the judgment call most worth a second opinion.

## Solution

`getPort()` gains one step on the worktree path: probe the hashed port, and if it is taken, walk forward through the 900-slot range (wrapping at 999 → 100) to the first free one.

Scanning forward from the hashed offset — rather than re-hashing or picking randomly — keeps the result stable for a given occupancy, so a worktree that lands on its second choice lands there again next time rather than reshuffling.

`isPortAvailable()` binds and immediately closes rather than consulting a list, so the answer reflects what `Bun.serve` will actually be allowed to do: a port held by another user or blocked by local policy reads as unavailable here for the same reason it would fail at startup. It probes on `process.env.HOST || '0.0.0.0'`, mirroring the server's own bind address, because availability is per-interface — a process bound to `127.0.0.1` does not block `0.0.0.0` on every platform.

Two paths deliberately keep their existing behavior:

- **Explicit `PORT`** returns unprobed. An explicit port is an instruction, not a preference; moving the operator off the port they named would break whatever they pointed at it, and a bind failure is the honest outcome.
- **The base port** returns unprobed, for the `vite.config.ts` reason above.

Range exhaustion — all 900 taken — fails loudly with a message naming `PORT`, rather than falling through to an arbitrary port.

`resolveWorktreePort` is exported for the same reason `calculatePortOffset` already is. My first draft tested the collision path through `getPort()`, which only reaches the worktree branch when the test process is itself running inside a worktree — so in the main repo, and therefore in CI, the test would have skipped its own assertion every single run and still reported green. Driving the resolver directly is what makes the fallback actually execute; the mutation check under Validation is what confirms it.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Worktree gets `3090 + hash(path)`, always | Worktree gets `3090 + hash(path)` when free; the next free port in 3190–4089 otherwise |
| **Failure behavior** | Server exits with a bare `EADDRINUSE` naming neither the cause nor `PORT` | Server starts on a neighbouring port and logs `worktree_port_reallocated`; only a fully exhausted range fails, with a message naming `PORT` |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[Start server in worktree] --> B2["hash(path) → 3516"]
    B2 --> B3{3516 free?}
    B3 -- yes --> B4[Listening on 3516]
    B3 -- no --> B5[EADDRINUSE, server exits]
  end

  subgraph After
    A1[Start server in worktree] --> A2["hash(path) → 3516"]
    A2 --> A3{3516 free?}
    A3 -- yes --> A4["Listening on 3516<br/>worktree_port_allocated"]
    A3 -- no --> A5[Scan forward in 3190-4089]
    A5 --> A6["Listening on 3519<br/>worktree_port_reallocated"]
    A5 -- all 900 taken --> A7["Fatal, names PORT"]
  end
```

## Validation

- `bun test src/utils/port-allocation.test.ts` — **16 pass, 0 fail** — proves the hashed port is still returned when free, the scan advances past occupied ports, it wraps 999 → 100, it returns to the hashed port once released, and an explicit `PORT` is returned even when occupied.

- **Mutation check** — disabling the probe (`if (false && !(await isPortAvailable(...)))`) turns exactly the four collision tests red and leaves the other twelve green; restoring it returns all sixteen to green. This is the evidence that the new tests actually constrain the new behavior rather than passing alongside it.

- **Runtime evidence** from the test logs — the fallback genuinely executes, rather than the assertions being satisfied some other way:

  ```
  preferredOffset:426 → offset:427, attempt:1, port:3517   worktree_port_reallocated
  preferredOffset:426 → offset:429, attempt:3, port:3519   worktree_port_reallocated
  preferredOffset:999 → offset:100, attempt:1, port:3190   worktree_port_reallocated   (wrap)
  ```

- `bun run test` in `@archon/core` — **exit 0**, 44 test files, zero failures — proves nothing else in the package regressed.
- `bun --filter @archon/core type-check` — exit 0.
- `bun run lint --max-warnings 0` (whole repo) — exit 0. It caught two real `no-confusing-void-expression` errors in the first draft of `isPortAvailable`; both are fixed.
- `check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix` — all exit 0.

**Not verified:**

- `bun run validate` does not complete on my machine: `@archon/docs-web type-check` exits 1 with *"Node.js v20.12.2 is not supported by Astro! Please upgrade Node.js to >=22.12.0"*. I confirmed this reproduces on a clean stash of `dev`, so it is environmental and not caused by this change — but it does mean the aggregate `validate` gate is unproven here, and CI is its first real run.
- `@archon/cli` has two failing tests on my machine (`CLI startup import boundary`, and a caller-relative path test). Both reproduce on a clean stash of `dev` and both fail at ~5000ms, which looks like a Windows timeout rather than a real assertion failure. Pre-existing either way.
- One `@archon/providers` test (`PiProvider > sendQuery installs PI_PACKAGE_DIR shim`) failed once at 7392ms during the full `--parallel` run, but passes in isolation both with and without this change. Load-induced flake under parallel CPU contention, not a regression.
- Bun **1.4.0** locally; CI pins **1.3.11**. Within the `^1.3.0` engines range, but not the same runtime CI will use.

## Delivery considerations

|---|---|---|
| Compatibility / migration | None. The hashed port is unchanged whenever it is free, so every non-colliding setup behaves identically. Nothing persists or re-derives the port. | `grep` for `getPort`/`calculatePortOffset` finds one consumer, `packages/server/src/index.ts:688` |
| Observability | A moved port is visible as `worktree_port_reallocated` carrying `preferredOffset`, the chosen `offset`, and `attempt`; the existing `server_listening` line still reports the bound port | `port-allocation.ts`, `packages/server/src/index.ts:933` |
| Documentation / communication | `dx-quirks.md` documented "same worktree always gets same port" with no collision case; it now describes the fallback and the `PORT` guarantee | `packages/docs-web/src/content/docs/contributing/dx-quirks.md` |

## Links

- Closes #3130
- Supersedes #3129 — same branch and commit, rebased onto current `dev`. GitHub refused to reopen #3129 (`reopenPullRequest` fails, and the REST equivalent returns 422), so this replaces it rather than ignoring the request to reopen.

## Summary by CodeRabbit

* **New Features**
  * Worktree servers now automatically select the next available port when the preferred port is occupied.
  * Port allocation supports wraparound and recovery when previously occupied ports become available.
  * Explicit `PORT` settings remain fixed and are not automatically reassigned.

* **Documentation**
  * Added guidance on port collision handling, reassignment logs, and explicit port behavior.

### DIFF
--- a/packages/core/src/utils/port-allocation.test.ts
+++ b/packages/core/src/utils/port-allocation.test.ts
@@ -1,5 +1,27 @@
 import { describe, it, expect, afterEach } from 'bun:test';
-import { calculatePortOffset, getPort } from './port-allocation';
+import { createServer, type Server } from 'node:net';
+import {
+  calculatePortOffset,
+  getPort,
+  isPortAvailable,
+  resolveWorktreePort,
+} from './port-allocation';
+
+const BASE_PORT = 3090;
+const HOSTNAME = '0.0.0.0';
+
+/** Hold a real listener so availability probes see a genuinely occupied port. */
+function occupy(port: number): Promise<Server> {
+  return new Promise((resolve, reject) => {
+    const server = createServer();
+    server.once('error', reject);
+    server.listen(port, HOSTNAME, () => resolve(server));
+  });
+}
+
+function release(server: Server): Promise<void> {
+  return new Promise(resolve => server.close(() => resolve()));
+}
 
 // Test the exported hash calculation function directly
 describe('calculatePortOffset', () => {
@@ -44,16 +66,56 @@ describe('calculatePortOffset', () => {
   });
 });
 
+describe('isPortAvailable', () => {
+  const held: Server[] = [];
+
+  afterEach(async () => {
+    await Promise.all(held.splice(0).map(release));
+  });
+
+  it('reports an unbound port as available', async () => {
+    // Port 0 asks the OS for any free port; bind it, learn the number, release it.
+    const scout = await occupy(0);
+    const address = scout.address();
+    const port = typeof address === 'object' && address ? address.port : 0;
+    await release(scout);
+
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(true);
+  });
+
+  it('reports a port held by a live listener as unavailable', async () => {
+    const server = await occupy(0);
+    held.push(server);
+    const address = server.address();
+    const port = typeof address === 'object' && address ? address.port : 0;
+
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(false);
+  });
+
+  it('leaves the probed port bindable afterwards', async () => {
+    const scout = await occupy(0);
+    const address = scout.address();
+    const port = typeof address === 'object' && address ? address.port : 0;
+    await release(scout);
+
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(true);
+    // The probe must not leak its own listener, or the second call would fail.
+    expect(await isPortAvailable(port, HOSTNAME)).toBe(true);
+  });
+});
+
 // Test getPort() behavior with mocked dependencies
 describe('getPort', () => {
   const originalEnv = process.env.PORT;
+  const held: Server[] = [];
 
-  afterEach(() => {
+  afterEach(async () => {
     if (originalEnv === undefined) {
       delete process.env.PORT;
     } else {
       process.env.PORT = originalEnv;
     }
+    await Promise.all(held.splice(0).map(release));
   });
 
   it('should return PORT env var when explicitly set to valid number', async () => {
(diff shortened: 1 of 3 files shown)


alixnaveed:
Confirmed the Vite proxy behavior and filed it separately as #3178, as suggested — it is not part of this PR.

Short version: in a worktree with no `PORT`, `getPort()` returned 3490 while `vite.config.ts`'s `loadEnv(...).PORT ?? '3090'` resolved to 3090, so the dev web UI dials a port nothing is listening on. Two other places share that same 3090 fallback (`web/src/lib/api.ts`, which bypasses the proxy for SSE, and `console/lib/http.ts`). Verified by running both resolutions against a real worktree rather than reading the code.

Wirasm:
## Review incomplete

The collision scan itself is small and preserves explicit `PORT`, but the resulting chosen endpoint still has no path to Archon-owned Web consumers. That leaves the source issue's endpoint-consumer acceptance unmet: a normal worktree `bun dev` server can listen on its allocation while Vite REST and both SSE clients dial 3090. The existing Ubuntu CI red is unrelated to this diff, but the required review coverage and local focused test could not complete in the detached checkout because workspace package links are absent.

**1 blocking · 0 non-blocking**

**Validation:** diff check passed; GitHub's Ubuntu Test Suite failed in unchanged workflow-fixture tests; focused core test could not resolve `@archon/git` from this detached checkout.

### Findings

|---|---|---|---|
| `R1` | Important | Worktree Web clients still target 3090 instead of the allocated server port. | OPEN |

<details>
<summary><code>R1</code> — worktree allocation is not propagated to Web consumers</summary>

**Impact:** Starting the documented `bun dev` workflow in a worktree without `PORT` can now start successfully on a fallback port, while REST proxy traffic and SSE traffic continue to use 3090 and fail. This violates #3130's requirement to report the actual endpoint to every Archon-owned consumer that needs it.

**Evidence:** `packages/core/src/utils/port-allocation.ts:94` returns a dynamically selected worktree port; `packages/server/src/index.ts:688` consumes it for `Bun.serve`; `packages/web/vite.config.ts:11` independently falls back to 3090 for the `/api` proxy; `packages/web/src/lib/api.ts:17` and `packages/web/src/experiments/console/lib/http.ts:7` do the same for direct SSE. The same mismatch existed before this diff, but it is the exact consumer side of this PR's source issue acceptance and is tracked as #3178.

**Required outcome:** Own the resolved development endpoint once and make the server, Vite proxy, and both SSE clients consume that value. Preserve the explicit `PORT` and main-checkout 3090 behavior. Add a test that runs the real worktree-resolution path and proves REST and SSE resolve the server's selected endpoint.

**Found by:** `coordinator`

**Disposition:** OPEN — #3178 is a verified related issue, but the missing endpoint handoff is required by #3130 and cannot be deferred while claiming this PR completes it.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | NOT RUN — independent reviewer did not return before the review deadline. |
| seams | NOT RUN — independent reviewer did not return before the review deadline. |
| simplify | NOT RUN — independent reviewer did not return before the review deadline. |
| tests | NOT RUN — independent reviewer did not return before the review deadline. |
| errors | NOT RUN — independent reviewer did not return before the review deadline. |

#### Validation

|---|---|---|
| `git diff --check origin/dev...HEAD` | PASS | No whitespace errors. |
| `bun test src/utils/port-allocation.test.ts` | NOT RUN | Detached checkout lacks the `@archon/git` workspace link (`ENOENT`). |
| GitHub `test (ubuntu-latest)` run 33733132250 | FAIL | Three `packages/workflows/src/fixture-runner.test.ts` expectations fail; that package is unchanged by this PR. |
| `lsof -nP -iTCP:3190-4089 -sTCP:LISTEN` | PASS | Existing Bun listener found on 3912; confirms a live external claimant is a real input to the allocation path. |

</details>

PR #3189, opened 2026-09-06
fix(cli): `archon serve` extracts the web UI from any Windows shell

### LINKED ISSUE #3188: archon serve cannot extract the web UI when Git's GNU tar precedes bsdtar on Windows PATH
## Problem

`archon serve` extracts the downloaded web UI by spawning a bare `tar`, so the binary is whatever PATH resolves. On Windows that is not stable. Windows ships bsdtar at `System32\tar.exe`, but Git for Windows puts GNU tar at `Git\usr\bin`, and a shell that adds it to PATH (Git Bash, and any terminal configured from it) wins the lookup. GNU tar cannot open the drive-letter operand the command passes, so extraction exits 2.

The result is that the same command on the same machine succeeds from `cmd` and fails from Git Bash.

## Why

`serve` is how a self-hosting user gets the web UI. The failure is total for that shell, and the error text points at a path that looks corrupted rather than at a tar mismatch, so it does not suggest its own workaround.

## Desired outcome

`archon serve` extracts the web UI regardless of which shell launched Archon, on any supported Windows version.

## Acceptance

- [ ] Extraction succeeds on Windows with Git's `usr\bin` ahead of `System32` on PATH.
- [ ] Extraction still works on a Windows install that bundles no `tar` (pre-1803).
- [ ] Linux and macOS keep using the `tar` on PATH.
- [ ] Both Windows branches are covered by tests that run on a non-Windows runner.

### Reproduction

1. On Windows, use a shell whose PATH contains Git's `usr\bin` before `System32` (Git Bash does this).
2. Run `bun test packages/cli/src/commands/serve.test.ts`.
3. The two `downloadWebDist` cases that extract a fixture archive fail.

Observed:

```
error: tar extraction failed (exit 2): tar: C\:\Users\<user>\AppData\Local\Temp\serve-webdist-test-XafDM4\target-embedded-ok.tmp: Cannot open: No such file or directory
```

The `C\:\Users\...` spelling is GNU tar rewriting the `-C C:\Users\...` operand. Running the same suite from `cmd`, where `System32\tar.exe` wins the lookup, passes.

### Environment

- Platform or adapter: CLI (`archon serve`)
- Database: SQLite
- OS: Windows 11, Git for Windows, Bun 1.3.9

buun-dev:
## Problem and outcome

`archon serve` extracted the web UI by spawning a bare `tar`, so the binary came from PATH. Windows ships bsdtar at `System32\tar.exe`, but Git for Windows puts GNU tar at `Git\usr\bin`, and GNU tar cannot open the drive-letter operand the command passes — it mangles `-C C:\Users\...` and exits 2. Which tar won depended on the PATH of whichever shell launched Archon, so extraction succeeded from `cmd` and failed from Git Bash.

- **Outcome:** extraction resolves `tar` explicitly, so `archon serve` works from any shell on Windows.
- **Invariant:** non-Windows platforms keep using the `tar` on PATH, and a Windows install with no bundled `tar` still falls back to PATH rather than spawning a path known to be absent.
- **Scope boundary:** the extraction arguments, staging, checksum verification, and cleanup paths are untouched. No other `tar` or subprocess call site changes.
- **Root cause:** the binary name was left to PATH resolution on a platform where two incompatible `tar` implementations are routinely both installed.

## Review guidance

- **Feedback requested:** correctness of the fallback branch, and whether the injected `platform`/`exists` parameters are the shape you want for platform-conditional resolution.
- **Start here:** `packages/cli/src/commands/serve.ts:286` — `resolveTarBin` is the whole behavior change; the call site at `:209` is one substituted argument.
- **Review order:** `serve.ts` (resolver + call site), then `serve.test.ts` (three branch cases, then the wiring assertion inside the existing spawn-shape test).
- **Lower-attention areas:** none; the diff is 83 lines across two files.
- **Known risk or uncertainty:** `SystemRoot` is read once at call time with a `C:\Windows` default. An install that relocates it and unsets the variable would fall back to PATH, which is the pre-existing behavior.

## Solution

`resolveTarBin()` returns `tar` off Windows and `%SystemRoot%\System32\tar.exe` on Windows when that file exists, falling back to `tar` when it does not (pre-1803 Windows bundles none). `downloadWebDist` spawns the resolved value instead of the literal.

`platform` and `exists` are parameters with live defaults rather than module mocks. That keeps both Windows branches covered when the runner is Linux, and avoids adding a `mock.module` to a file the repository already keeps free of cross-package mock pollution.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `archon serve` extraction depends on the launching shell's PATH on Windows | Extraction resolves the same binary from every shell |
| **Failure behavior** | `tar extraction failed (exit 2)` naming a mangled `C\:\Users\...` path | Not reached; on a Windows build with no bundled tar, resolution falls back to PATH as before |

## Validation

- `bun test packages/cli/src/commands/serve.test.ts` — 24 pass, 0 fail on Windows with Git Bash on PATH. **Before this change the same suite reports 3 failures on that host**: the new `resolveTarBin` Windows case, plus the two pre-existing `downloadWebDist` extraction cases, which fail with `tar: C\:\Users\...: Cannot open: No such file or directory`. The fix turns existing red green.
- `bun --filter '@archon/cli' test` — exits 0.
- `bun --filter '@archon/cli' type-check`, `bun run type-check` — exit 0.
- `bun run lint`, `bun run format:check` — pass, including `check-test-cleanup-drift`.
- `bun run check:bundled`, `check:bundled-skill`, `check:bundled-schema`, `check:pi-vendor-map`, `check:capability-matrix`, `check:api-types` — all pass.
- **Not verified:** `bun run test:install` — `scripts/test-install.sh` refuses MINGW/MSYS by design, and CI already skips it on Windows. The full `bun run test` was run per package rather than through the root fail-fast wrapper.

## Links

- Closes #3188

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when serving or downloading web distributions on Windows by selecting the system archive extractor when available.
  - Added a fallback to the existing PATH-based extractor when the Windows system utility is unavailable.
  - Preserved consistent archive extraction behavior across non-Windows platforms.
  - Expanded automated coverage to verify extractor selection across supported environments.

### DIFF
--- a/packages/cli/src/commands/serve.test.ts
+++ b/packages/cli/src/commands/serve.test.ts
@@ -31,7 +31,13 @@ mock.module('@archon/paths', () => ({
   BUNDLED_WEB_DIST_SHA256: '',
 }));
 
-import { serveCommand, parseChecksum, parseEmbeddedChecksum, downloadWebDist } from './serve';
+import {
+  serveCommand,
+  parseChecksum,
+  parseEmbeddedChecksum,
+  downloadWebDist,
+  resolveTarBin,
+} from './serve';
 
 describe('parseChecksum', () => {
   const validHash = 'a'.repeat(64);
@@ -179,6 +185,48 @@ function buildWebTarball(indexHtml: string): Uint8Array<ArrayBuffer> {
   );
 }
 
+describe('resolveTarBin', () => {
+  // Windows ships bsdtar at System32\tar.exe, but Git for Windows puts GNU tar
+  // on PATH ahead of it, and GNU tar cannot open a drive-letter operand: it
+  // mangles the `-C C:\Users\...` operand into a colon-escaped path and exits 2.
+  // Leaving the binary to PATH makes extraction depend on which shell launched
+  // Archon, so these cases pin the choice on every host — CI runs them off Windows.
+  const SYSTEM32_TAR = /[/\\]System32[/\\]tar\.exe$/;
+
+  it('pins the Windows system tar when it is present', () => {
+    const probed: string[] = [];
+
+    const bin = resolveTarBin('win32', path => {
+      probed.push(path);
+      return true;
+    });
+
+    expect(bin).not.toBe('tar');
+    expect(bin).toMatch(SYSTEM32_TAR);
+    // The probe must ask about the path it returns, not some other file.
+    expect(probed).toEqual([bin]);
+  });
+
+  it('falls back to PATH when Windows has no bundled tar', () => {
+    // Pre-1803 Windows ships no tar. Returning the absolute path anyway would
+    // spawn a file that does not exist, which is a worse failure than a PATH miss.
+    expect(resolveTarBin('win32', () => false)).toBe('tar');
+  });
+
+  it('leaves the binary to PATH off Windows', () => {
+    const probed: string[] = [];
+
+    const bin = resolveTarBin('linux', path => {
+      probed.push(path);
+      return true;
+    });
+
+    expect(bin).toBe('tar');
+    // No filesystem probe at all — the POSIX `tar` on PATH is the right one.
+    expect(probed).toEqual([]);
+  });
+});
+
 describe('downloadWebDist', () => {
   let tmpRoot: string;
   let tarballBytes: Uint8Array;
@@ -220,11 +268,13 @@ describe('downloadWebDist', () => {
     const targetDir = join(tmpRoot, 'target-embedded-ok');
     const spawnSpy = spyOn(Bun, 'spawn');
     let extractorStdin: unknown;
+    let extractorBin: string | undefined;
 
     try {
       await downloadWebDist('9.9.9', targetDir, tarballHash);
       // Read before restoring — mockRestore() clears the recorded calls.
       extractorStdin = (spawnSpy.mock.calls[0]?.[1] as { stdin?: unknown } | undefined)?.stdin;
+      extractorBin = (spawnSpy.mock.calls[0]?.[0] as string[] | undefined)?.[0];
     } finally {
       spawnSpy.mockRestore();
     }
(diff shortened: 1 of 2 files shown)


buun-dev:
The `test (ubuntu-latest)` failure is not from this branch. This PR touches only `packages/cli`, and `@archon/cli:test` reports `0 fail` in that same run; `test (windows-latest)` — the platform the change is about — is green.

The three reds are in `@archon/workflows > runFixtures`, and they are order-dependent rather than flaky-by-timing. `fixture-runner.test.ts:760` asserts `toEqual` on an array whose order comes from directory traversal:

```
  [
-   "sdlc/plan/fixtures/plan.stubs.yaml",
    "sdlc/ship/fixtures/ship.stubs.yaml",
+   "sdlc/plan/fixtures/plan.stubs.yaml",
  ]
```

Same two entries, opposite order. The file passes on `dev` locally, so the runner's readdir order is the variable. Happy to file it as its own issue if it is not already tracked — it will redden unrelated PRs at random until the assertion is order-insensitive or `runFixtures` sorts its results.

buun-dev:
Filed the `runFixtures` ordering failure as #3194 — it reproduced identically on this PR and on another branch touching disjoint code, and the cause is `walkForFixtures` pushing in raw `readdir` order (`fixture-runner.ts:191-207`). Nothing on this branch touches it; `test (windows-latest)` is green here.

Wirasm:
Thanks @buun-dev — this is a good fix and it has shipped, with your commit intact.

It has been folded into #3280 rather than merged here. That PR was already rewriting the same extraction path in `serve.ts` (bounding the child wait and correcting a diagnostic that could not tell our timeout from the runner's), and this branch had drifted 90 commits behind `dev` with a conflict in `serve.test.ts`. Landing both separately would have meant one of them resolving a conflict against the other.

Your work was **cherry-picked, not rewritten**, so `1a66d2ef9` on that branch carries your authorship. `resolveTarBin` is in as you wrote it, with your docblock and both injected parameters, and your tests came across unchanged — the only edit was to the `serve.test.ts` import block, which had moved on `dev` since you branched.

Worth saying: the ubuntu failure on this PR was never your code. It was three `runFixtures` tests in `@archon/workflows`, a file that had moved three times since you opened it. Pure staleness.

The diagnosis holds up well. Git for Windows putting GNU tar on PATH at `Git\usr\bin`, where it cannot parse the drive-letter operand in `-C C:\Users\...`, is exactly the kind of thing that only shows up when someone actually runs it from the wrong shell. Pinning `System32\tar.exe` and injecting `platform`/`exists` so both branches stay covered on a non-Windows runner is the right shape.

Closing in favour of #3280.

PR #3191, opened 2026-09-06
fix(workflows): a loop's until_bash check honors the node's declared timeout

### LINKED ISSUE #3190: A loop node's timeout is accepted, then dropped: until_bash is always killed at the 120s default
## Problem

`timeout` is a known DAG node key, so `parseDagNode` accepts it on a `loop:` or `loop_group:` node. The mode transform then drops it, and both loop executors call the `until_bash` subprocess with the hardcoded `SUBPROCESS_DEFAULT_TIMEOUT` instead. A loop predicate therefore cannot declare a budget: it is killed at 120 s no matter what the node says.

Because the kill is a SIGTERM, the subprocess reports `code: null`, and the run surfaces it as a failure that happened *before* execution rather than as a timeout.

This is an instance of the silent-drop class tracked in #2478, but with a runtime consequence rather than a no-op: the value is not merely ignored, a real gate is terminated.

## Why

`until_bash` on its own is the documented preferred shape whenever a completion condition is externally checkable, and the guide's own example is `until_bash: "bun run test"` (`guides/loop-nodes.md`). A test suite, a build, or a lint gate on any non-trivial repository exceeds 120 s, so the recommended pattern is the one that cannot express its own budget.

The failure is also misdiagnosable. The operator sees a node that "failed before execution" and no timeout anywhere, so the natural reading is that the script is broken.

## Desired outcome

A `loop:` or `loop_group:` node's `until_bash` check honors the node's declared `timeout`, exactly as `bash:` and `script:` nodes already do, and keeps the 120 s default when none is declared.

## Acceptance

- [ ] A loop node declaring `timeout: 600000` runs its `until_bash` with a 600000 ms budget.
- [ ] The same holds for `loop_group`.
- [ ] A loop node declaring no `timeout` still uses the 120 s default.
- [ ] A non-positive `timeout` on either node kind is rejected at parse time rather than silently producing an immediate kill.

### Reproduction

1. Author a loop node with `timeout: 600000` and an `until_bash` that runs longer than 120 s.
2. Run the workflow.
3. The check is SIGTERMed at 120 s; the declared 600000 never reaches the subprocess.

Asserting on the timeout passed to the loop's subprocess call reports:

```
Expected: 600000
Received: 120000
```

The two call sites are `executeLoopNode` and `executeLoopGroupNode` in `packages/workflows/src/dag-executor.ts`, which pass `SUBPROCESS_DEFAULT_TIMEOUT` where the ordinary bash and script paths resolve `node.timeout ?? SUBPROCESS_DEFAULT_TIMEOUT`.

### Environment

- Platform or adapter: workflow engine (any)
- Database: SQLite
- OS: platform-independent

buun-dev:
## Problem and outcome

`timeout` is a known DAG node key, so `parseDagNode` accepts it on a `loop:` or `loop_group:` node. The mode transform then dropped it, and both loop executors called the `until_bash` subprocess with the hardcoded `SUBPROCESS_DEFAULT_TIMEOUT`. A loop predicate could not declare a budget: it was killed at 120 s regardless. Because a SIGTERM leaves `code: null`, the run surfaced it as a failure that happened *before* execution rather than as a timeout, so the natural reading was a broken script.

- **Outcome:** a loop's `until_bash` runs with the node's `timeout`, so a predicate that runs a real gate can declare the budget it needs.
- **Invariant:** a loop that declares no `timeout` still uses the 120 s default, and `timeout` keeps the same meaning, units, and default it has on `bash:` and `script:` nodes.
- **Scope boundary:** no new node key, no change to completion-channel semantics, iteration accounting, or the ordinary bash/script paths. `loop_group` body nodes keep resolving their own timeouts as before.
- **Root cause:** two call sites passed the constant where the rest of the engine resolves `node.timeout ?? SUBPROCESS_DEFAULT_TIMEOUT`.

## Review guidance

- **Feedback requested:** correctness, and whether rejecting a non-positive `timeout` at parse time is the failure posture you want (the alternative is accepting it and letting the subprocess die immediately).
- **Start here:** `packages/workflows/src/dag-executor.ts` — the two substituted arguments in `executeLoopNode` and `executeLoopGroupNode` are the entire behavior change.
- **Review order:** `dag-executor.ts`, then `schemas/dag-node.ts` (carry `timeout` through both transforms + the validation), then the tests, then the one docs paragraph.
- **Lower-attention areas:** `dag-executor.test.ts` is 136 lines of three cases built on the existing `dagOptions({...})` harness.
- **Known risk or uncertainty:** a workflow that already sets `timeout` on a loop node gets a longer-lived `until_bash` than it did yesterday. That value was authored and silently discarded, so honoring it is the intended reading, but it is a behavior change for any such workflow.

## Solution

Both loop call sites resolve `node.timeout ?? SUBPROCESS_DEFAULT_TIMEOUT`. `loopNodeSchema` and `loopGroupNodeSchema` carry `timeout` through their transforms, and `superRefine` rejects a non-positive or non-finite value so an unusable budget fails at load rather than becoming an immediate kill.

This removes a dropped field rather than adding a new one. `timeout` already exists on the node base with a fixed meaning; the loop modes were the two places that did not honor it. Per the workflow-language constitution that keeps the YAML surface flat: no new governance concept is introduced.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `timeout` on a loop node parsed, then had no effect | The declared budget reaches the `until_bash` subprocess |
| **Failure behavior** | Predicate SIGTERMed at 120 s; `code: null` reported as a failure before execution | Predicate runs to its declared budget; a non-positive `timeout` is now a load-time parse error |

## Validation

- `bun test packages/workflows/src/dag-executor.test.ts -t "until_bash honors the node timeout"` — 3 pass. With the two production files reverted and the tests kept, 2 of the 3 fail with `Expected: 600000 / Received: 120000`, and the third (the default-fallback guard) stays green.
- `bun test packages/workflows/src/schemas.test.ts -t "timeout validation"` — 2 pass; both fail without the schema change.
- `bun --filter '@archon/workflows' test` — 6 failures, **identical to the 6 on pristine `dev` @ `4d48a032` on this host** (`#2637` value transport ×3, `#2453` result contract ×3). Measured by stashing this branch's changes and re-running: same 6, same names. This PR adds none.
- `bun run type-check`, `bun run lint`, `bun run format:check`, `bun run check:bundled-schema`, `bun run check:api-types` — all exit 0.
- **Not verified:** `bun run test:install` (`scripts/test-install.sh` refuses MINGW/MSYS by design; CI skips it on Windows). The 6 pre-existing `@archon/workflows` failures are environmental on this host and were not investigated here.

## Links

- Closes #3190
- Related #2478

## Summary by CodeRabbit

* **New Features**
  * Loop and loop-group nodes can set a custom timeout for `until_bash` checks.
  * Timeouts are specified in milliseconds; unset values use the 120-second default.
  * The configured timeout now applies when running the check, helping prevent longer predicates from being terminated prematurely.

* **Documentation**
  * Updated the loop-node guide with timeout behavior, failure considerations, and a longer-timeout example.

### DIFF
--- a/packages/docs-web/src/content/docs/guides/loop-nodes.md
+++ b/packages/docs-web/src/content/docs/guides/loop-nodes.md
@@ -255,6 +255,20 @@ loop:
 `until_bash` runs only on iterations that no other completion channel already
 ended, so a loop declaring more than one never pays for a redundant check.
 
+The check is a subprocess and takes the node's `timeout` (milliseconds, default
+120000), like `bash:` and `script:` nodes. A predicate that runs a real gate can
+exceed that, so declare a budget on the node when it does — otherwise the check is
+killed and the run reports a failure the script never produced:
+
+```yaml
+- id: fix-tests
+  timeout: 900000             # The suite takes ~4 min; 120 s would kill the check.
+  loop:
+    prompt: "Fix the failing tests"
+    max_iterations: 5
+    until_bash: "bun run test"
+```
+
 :::caution[If your `until_bash` accumulates state]
 The skip means the script does not run on an iteration another channel already
 completed, so a check that *mutates* state each time it runs — a counter, an append,
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -33527,3 +33527,139 @@ describe('executeDagWorkflow -- side effects survive a failed terminal write', (
     expect(sent.some(m => m.includes('completed with failures'))).toBe(true);
   });
 });
+
+// A loop predicate could not declare a time budget: both `until_bash` call sites passed a
+// bare SUBPROCESS_DEFAULT_TIMEOUT while the ordinary bash/script paths already resolved
+// `node.timeout ?? SUBPROCESS_DEFAULT_TIMEOUT`. Nodes are built through the real schema, so
+// each test proves the whole chain (schema -> transform -> executor), not just the
+// executor's fallback expression.
+describe('executeDagWorkflow -- until_bash honors the node timeout', () => {
+  let testDir: string;
+
+  beforeEach(async () => {
+    testDir = join(
+      tmpdir(),
+      `dag-until-bash-timeout-${Date.now()}-${Math.random().toString(36).slice(2)}`
+    );
+    await mkdir(testDir, { recursive: true });
+    mockSendQueryDag.mockClear();
+  });
+
+  afterEach(async () => {
+    await removeTempTree(testDir);
+  });
+
+  // runSubprocess is local/unexported — spy on the module boundary it actually calls
+  // (@archon/git's execFileAsync), the same way the script-injection tests above do.
+  function spyOnExec() {
+    return spyOn(git, 'execFileAsync').mockResolvedValue({ stdout: '', stderr: '' });
+  }
+
+  function untilBashTimeout(execSpy: ReturnType<typeof spyOnExec>): number | undefined {
+    const call = execSpy.mock.calls.find(c => (c[1] as string[])[0] === '-c') as
+      | [string, string[], { timeout: number }]
+      | undefined;
+    expect(call).toBeDefined();
+    return call?.[2]?.timeout;
+  }
+
+  it('applies the node timeout to a loop until_bash subprocess', async () => {
+    const node = dagNodeSchema.parse({
+      id: 'my-loop',
+      timeout: 600_000,
+      loop: { prompt: 'Do a task.', until: 'NEVER_EMITTED', until_bash: 'true', max_iterations: 1 },
+    });
+
+    const execSpy = spyOnExec();
+    try {
+      mockSendQueryDag.mockImplementation(async function* () {
+        yield { type: 'assistant', content: 'working on it' };
+        yield { type: 'result', sessionId: 'loop-timeout-session' };
+      });
+
+      await executeDagWorkflow(
+        dagOptions({
+          deps: createMockDeps(),
+          platform: createMockPlatform(),
+          cwd: testDir,
+          workflow: { name: 'dag-loop-timeout', nodes: [node] },
+          workflowRun: makeWorkflowRun(),
+        })
+      );
+
+      expect(untilBashTimeout(execSpy)).toBe(600_000);
+    } finally {
+      execSpy.mockRestore();
+    }
+  });
+
+  // Pins the `??` fallback the fix depends on: a loop node with NO declared timeout must
+  // still get SUBPROCESS_DEFAULT_TIMEOUT (120000). Guards against a future edit that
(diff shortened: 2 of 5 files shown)


buun-dev:
CI caught a real problem with the first push, and it changes part of the framing above.

`schemas.test.ts` has a contract test — `ignores deferred exec fields on every mode that does not select them` — that explicitly lists `timeout` among the fields `loop` and `loop_group` drop. So the drop is deliberate and pinned, not an oversight, and my "silently dropped" description was only half right: the *engine* discards it silently, but the *schema* does so on purpose.

That makes this a scope question rather than a bug fix, so I want to be explicit about what I changed and why, and you may well want the other answer.

**What I changed.** The two loop cases move out of that table's `ignored` set, because after this change those modes select `timeout`. The test's own contract is untouched — only its membership. Two new schema tests pin the replacement behavior (`timeout` survives both mode transforms), and both are red against the current schema.

**What I dropped.** The `superRefine` rejecting a non-positive `timeout` is gone. It introduced a new load-time failure for a workflow that parses today, and it diverged from `bash:`/`script:`, which accept `timeout: 0` without complaint. If an existing field is going to mean the same thing in a new place, it should mean the *same* thing — including its rough edges. That also keeps the diff to the smallest change that makes the budget reach the subprocess.

**The judgment call I'd like your read on.** I took the node-level shape over inventing `loop.until_bash_timeout` because the constitution's §92 precedent favors making an existing field mean the same thing in both places over adding YAML surface. The counter-argument is real: a loop node's own work is AI iterations, so one could argue `timeout` belongs only to exec nodes and an `until_bash` budget belongs inside the `loop:` block. If that is your reading, say so and I will re-cut it that way — the underlying problem is only that a documented `until_bash: "bun run test"` cannot outlive 120 s.

**On the ubuntu-latest failure.** The three `runFixtures` reds on the earlier run are not from this branch. They are order-dependent: `fixture-runner.test.ts:760` asserts `toEqual` on an array whose order comes from directory traversal, and CI returned `[ship, plan]` where the test expects `[plan, ship]`. The suite passes on this branch locally and on `windows-latest`. Happy to file that separately if it is not already known.

Wirasm:
Superseded by #3406, which carries this PR's completion-check timeout fix for #3190. Your original commits and authorship are preserved in the successor's history. The successor also covers the resumed loop-group probe and validates both loop modes through the existing exec timeout schema.

The combined tree passes `bun run validate` and the docs build. Independent review and CI continue on #3406; this PR is being closed only to keep one reviewable delivery. The source issue remains open until the successor merges.

[Successor](https://github.com/coleam00/Archon/pull/3406) · [Published plan and scope mapping](https://github.com/coleam00/Archon/issues/3324#issuecomment-5760744426)

Issue #2848, opened 2026-08-27
design(providers): define the minimal agent-runtime contract


Wirasm:
## Problem

Archon's engine/provider boundary is not one small, provider-neutral contract today. The engine passes raw workflow `NodeConfig` and opaque assistant configuration into maintained providers, while provider adapters also decide configuration precedence, native resource loading, credential delivery, session lifecycle, capability filtering, retries, and event translation.

That coupling makes provider behavior difficult to predict and makes a new provider an engine change instead of an independently implementable agent-runtime plugin. It also encourages Archon to invent provider policy that belongs to the user's existing Claude, Codex, Pi, or future provider configuration.

The current resource-loading evidence is recorded in the operator research artifact:

`/Users/rasmus/.prp/archon-75601ef6/research/2026-08-27-provider-resource-loading-policy.md`

A focused Pi 0.84.3 spike also proved that the public session services and runtime lifecycle can preserve native settings, guidance, extensions, provider registration, exact skill selection, and shutdown without Archon's current loader cache, registration replay, resource proxy, or manual settings merge. PR #2839's behavioral outcome is valid, but its current implementation is not the target boundary.

## Desired outcome

Define the smallest typed contract through which Claude, Codex, Pi, and future third-party providers join Archon as agent runtimes.

The engine selects a registered runtime and supplies only governed execution inputs. The provider implementation delegates native behavior to its SDK, translates the request, and returns normalized events. The engine contains no provider-specific configuration policy and does not need provider-specific branches when another runtime is installed.

Users own their provider homes, authentication state, settings, extensions, commands, templates, hooks, global/project guidance, and other native configuration. Archon does not mirror, replace, merge, or reinterpret those resources.

Workflow nodes deliberately narrow only model-visible capabilities that require per-node governance:

- ambient skills are absent; the node receives exactly its declared skills;
- skill-like plugins are absent unless explicitly selected where the provider can enforce that boundary;
- MCP servers are absent unless explicitly selected where the provider can enforce that boundary;
- unsupported exact selection fails before model spend instead of silently broadening access.

Pi extensions remain user-owned Pi runtime configuration. They are not reclassified as Archon plugins and Archon does not govern capabilities an extension creates internally.

## Contract design requirements

The design must settle:

1. A provider-neutral, typed runtime request containing only inputs the engine owns, such as working directory/isolation projection, resolved model, acting-user credential material, prompt, output schema, cancellation, session continuation, tools, and exact governed capability selections.
[...]

Issue #2843, opened 2026-08-27
feat(providers): workflow nodes load only explicitly named MCP servers


Wirasm:
## Problem

Workflow `mcp:` is intended to be an explicit capability declaration, but providers do not currently share one truthful boundary. Claude can use strict MCP, while Codex merges declared servers with ambient user/plugin configuration. Pi has no Archon MCP integration, although user-installed Pi extensions may expose MCP-backed tools independently.

The invariant is exact: a workflow AI node must not load an MCP server it did not name. Preserving the user's provider configuration must not make ambient MCP servers available to every governed node.

## Desired outcome

- A node with no `mcp:` declaration receives no Archon-managed or ambient provider MCP servers.
- A node that names MCP servers receives exactly those servers and no siblings.
- Provider-native configuration and guidance still load; Archon filters the capability, not the whole config.
- Providers that cannot enforce an exact set report that limitation before model spend rather than silently merging ambient servers.
- Pi extensions remain outside this contract. If a user's Pi extension implements an MCP client and exposes tools, that is user-owned extension behavior; Archon's `mcp:` field remains unsupported on Pi until Pi exposes a governable native seam.

## Non-negotiable invariants

1. Empty or omitted node MCP means no provider MCP servers for that workflow node.
2. Explicit MCP selection is exact, not additive over ambient config.
3. Unknown or unavailable named servers fail clearly before the agent runs.
4. Fresh, resumed, retried, and reused provider sessions preserve the same exact set.
5. Filtering MCP does not suppress unrelated native provider settings or guidance.
6. Plugin activation does not implicitly bypass MCP selection.
7. Pi extension tools are not misreported as Archon-governed MCP servers.

## Required regression proof

For every provider that claims MCP support, install two ambient MCP servers with unique harmless tools and run this matrix:

- **none declared:** neither tool appears;
- **A declared:** A appears and B does not;
- **B declared:** B appears and A does not;
- **unknown declared:** validation/run preflight fails before provider spend;
- **config preservation:** native config and global/project guidance markers remain visible in all cases;
- **lifecycle parity:** fresh, resume, retry, and session-reuse paths retain the exact same set;
- **plugin separation:** selecting a plugin that contains MCP metadata does not expose that server until the node names it.

The decisive test must inspect the provider-visible tool/server catalog or execute the harmless named tool. Merely asserting an Archon option object is insufficient when the provider may merge ambient configuration afterward.

## Design constraint

[...]

Issue #2490, opened 2026-08-05
refactor(workflows): name the Workspace port that already exists — and make its absence fail fast


Wirasm:
Where work happens — a directory, an environment, somewhere to put outputs — is already a port. It is simply not named as one, so it is spread across `resolveProjectPaths` (`executor.ts:305-392`), `composeRunPaths`, `resolveScopeArtifactsDir`, `ExecutionContext` (56 references, from `@archon/providers/types`), and two structural ports smuggled through `ExecuteWorkflowOptions` — `container: ContainerRunContext` (`executor.ts:523-529`) and `resolveChildIsolation` (`:531-539`).

"Worktree" is one implementation of the concept, not the concept.

## Proposed

```ts
interface WorkspaceHandle {
  readonly root: string;                       // cwd for exec
  readonly env: Readonly<Record<string, string>>;
  readonly outputs: { artifacts: string; state: string };
  readonly exec: ExecStrategy;                 // host | container
}

interface WorkspaceProvider {
  /** Only called when a subgraph instance declares its own workspace. */
  acquire(spec: WorkspaceSpec): Promise<WorkspaceHandle>;
  release(handle: WorkspaceHandle, outcome: 'keep' | 'discard'): Promise<void>;
}
```

The **root** workspace is passed by value on the run request, matching today where `cwd` is a required argument rather than a port (`executor.ts:1095`). The provider is needed only when a *subgraph instance* wants its own — which, per the fan-out decision, is the uncommon case: a read-only fan-out needs no isolation at all.

## The contract that must be explicit

`WorkspaceProvider` is **optional**, and its absence must be a **fail-fast, never a silent fallback.** The engine has to be able to distinguish *"no isolation available"* from *"isolation not requested"* — a workflow that declared a per-instance workspace and silently got a shared checkout is the failure this port exists to prevent. That is a stated invariant, not an implementation detail.

## Why it is worth naming

Once it is a port, the host owns worktrees, containers and directory layout, and `@archon/workflows`'s dependency on `@archon/git` and `@archon/paths` becomes reviewable rather than assumed. That is the single largest step toward the engine being embeddable, and it costs less than any other item in the series.

## Acceptance

- One `WorkspaceHandle` type replaces the `ExecutionContext` + paths + container-context spread
- A declared per-instance workspace with no provider injected **throws at load or spawn**, never falls back
- `resolveChildIsolation` and `ContainerRunContext` stop riding `ExecuteWorkflowOptions`
[...]

PR #3196, opened 2026-09-07
fix(workflows): surface ignored AI fields in discovery warnings

### LINKED ISSUE #3216: fix(workflows): an include: that carries an AI field must fail or warn, not drop it silently
## Problem

An `include:` directive silently discards any AI field written on it (`denied_tools`, `allowed_tools`, `model`, `hooks`, and the rest of `BASH_NODE_AI_FIELDS`). The workflow loads, `archon validate` prints ok, every gate stays green, and the only trace is a debug-level log line (`include_node_ai_fields_ignored`) that no gate reads. A caller who writes a restriction on an include believes it applies; it does not.

## Why

A restriction that validates clean and is never applied is the worst failure class for governance fields: it reads as enforced. PR #3196 found this while trying to sandbox a composed block and proposed honoring one field on includes; that design is declined (see the PR), but the silent drop it exposed is a defect on its own.

## Desired outcome

Writing an AI field on an `include:` directive is loud: a load error, or at minimum a validate warning that names the field and the include, and states that includes attach a sub-graph and do not reconfigure it.

## Acceptance

- [ ] A workflow whose include carries any field from `INCLUDE_NODE_IGNORED_FIELDS` fails to load or produces a validate warning naming the field and the include id.
- [ ] The existing invariant test (every AI field is ignored on includes, gates, and waits) is unchanged; the new behavior is the reporting, not the field semantics.
- [ ] Bundled and repository workflows produce no new warnings (#2283's corpus guard confirms).
- [ ] The debug log line is removed or demoted to the reported path, so there is one channel.

- Must remain true: includes attach a sub-graph and do not reconfigure it. No field is honored on an include by this change.
- Scope boundary: reporting only. What a provider can enforce belongs to #2848's capability typing; a subtree read boundary belongs to the workspace (#2490, #2206).
- Known prerequisites or blockers: decide load error versus validate warning. A load error is the honest default for a governance field; a warning may be needed for the transition if any repository workflow carries one today.
- Related issues or PRs: #3196 (found it; its tests for the ignored-field invariant and the "no-op" case are reusable), #2453 (where the invariant was introduced), #2283, #2848.
- Solution steering: Hint, #3196's `schemas.test.ts` additions already pin the three ignore lists.

coleam00:
## Problem and outcome

An `include:` directive silently discards any AI field written on it (`denied_tools`, `allowed_tools`, `model`, `hooks`, the rest of the ignored list). The workflow loads, `archon validate` prints ok, and the only trace is a server log line no gate reads. A caller who writes a restriction on an include believes it applies; it does not. This is #3216.

- **Outcome:** the ignored-field condition now lands in `parseWarnings`, the channel the workflow's author reads through discovery, `workflow list` and the API. For an include the warning says why: an include attaches a sub-graph and does not reconfigure it, and names where the field belongs.
- **Invariant:** no field is honored on an include by this change. The existing invariant test (every AI field is ignored on includes, gates and waits) is unchanged; the new behavior is the reporting.
- **Scope boundary:** diagnostics only. Provider enforcement belongs to #2848; a subtree read boundary belongs to the workspace (#2490, #2206). The union design this branch originally carried is gone, per the maintainer correction path below.
- **Root cause:** the loader logged ignored raw fields without adding them to the warnings returned to the author.

## Review guidance

- **Feedback requested:** two decisions #3216 leaves open. (1) Load error or validate warning: this PR warns. A load error is the honest default for a governance field, but the corpus check has not been run against every repository workflow. (2) The log line: #3216 asks for one channel. Seven existing loader tests read the ignored-field condition from the log mock, so removing the log rewrites those; say whether you want that in this PR or in the follow-up once (1) is settled.
- **Start here:** `packages/workflows/src/loader.ts`, the ignored-field branch in `parseDagNode`.
- **Review order:** loader warning, the discovery regression test in `loader.test.ts`, then the schema tests pinning the ignore lists.
- **Lower-attention areas:** the comment on `INCLUDE_NODE_IGNORED_FIELDS`.
- **Known risk or uncertainty:** warnings remain advisory. Nothing here enforces a boundary.

## Solution

Reuse the existing ignored-field classification and push a node-specific warning naming the node, the fields, and for an include the reason. Include-level `denied_tools` and `allowed_tools` continue to be ignored; expanded children keep their own settings.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | An ignored include restriction appears only in server logs. | Discovery returns a warning naming the node, the ignored field, and where it belongs. |
| **Failure behavior** | Loading succeeds with no author-visible explanation. | Loading still succeeds, with the ineffective setting reported. |

## Validation

- `bun test src/loader.test.ts` and `src/schemas.test.ts` in `packages/workflows`: pass (counts in the latest commit). The new discovery test verifies the returned warning and confirms the included node stays unrestricted.
- **Not verified:** a full repository suite on Windows, and the corpus check that no bundled or repository workflow carries an AI field on an include today.

## Links

- Closes #3216
- Related #2848, #1118, #2843 (provider capability), #2490, #2206 (workspace boundaries)
- [Maintainer correction path](https://github.com/coleam00/Archon/pull/3196#issuecomment-5583753293)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workflow parsing warnings for unsupported AI-specific fields, including clearer guidance for include nodes.
  - Include-node tool restrictions are ignored as expected, with an operator-visible warning.
  - Workflow validation now correctly requires causes for skipped outputs while allowing pending outputs without causes.
  - Timeout configuration validation now accepts supported skip behavior for Bash and script nodes and rejects unsupported outcomes.

- **Tests**
  - Added coverage for output provenance, timeout handling, include-node fields, and warning behavior.


Wirasm:
Please fill in the PR template for this one (`.github/pull_request_template.md`: Problem and outcome, Review guidance, Solution, Behavior change, Validation, Links). I'll be reviewing each of the SDLC pack PRs in detail, and the template is what I read first to understand the intended shape. Same request across the current batch.

Wirasm:
The problem is real, and the instinct that composition may only narrow is right. The primitive is wrong, so I am not taking the union feature. Here is why, where each part belongs, and what a version of this PR I would merge looks like.

**Why the shape is wrong.** `denied_tools` is per-node capability policy: this node may not call that tool. The PR reuses it to express a subtree boundary: nothing under this include may touch that path. Those are different things, and the union cannot tell which one the caller meant. A path-scoped specifier like `Read(.factory/holdout/**)` is the one case where unioning onto every node is harmless. A bare tool name on the include strips that tool from every node in the block, including nodes the block author gave it to on purpose, and a nested block cannot opt out.

**The providers make it worse.** One YAML line, three behaviors:
- Claude passes the list through as `disallowedTools`, so path-scoped specifiers are enforced.
- Pi subtracts denied entries from its tool set by name. A path specifier is not a tool name, so the same line either does nothing or, once stripped to its base, removes the whole tool.
- Codex declares `toolRestrictions: false` and enforces nothing.

Validation says ok in all three cases. That fails the provider-agnostic rule and the fail-safe rule for implicit behavior.

**Where each capability lives.**
- What a provider can enforce belongs in the agent-runtime contract, #2848. I have recorded there that `toolRestrictions` needs to say what it enforces (name-level versus pattern-scoped) so the rule #1118 and #2843 already state applies: a node declaring a restriction its provider cannot enforce fails or warns at validation, and never claims the boundary silently.
- The boundary you actually want, a directory the composed block cannot see, belongs to the workspace: #2490 (the workspace handle) and #2206 (repo containers). A holdout that is not in the checkout cannot be read by any provider, with no tool policy involved. If a subtree read boundary is ever expressed in YAML, it is its own field with path semantics, refused at load where the workspace cannot enforce it. Not now.

**The correction path for this PR.** Narrow it to the part that is a plain bug fix: an include silently drops AI fields with only a debug log, so a caller's restriction vanishes while validation says ok. Make that a load error or a validate warning naming the field. Keep the tests that pin the ignored-field invariant. Drop `applyIncludeDeniedTools`, the schema admission of `denied_tools` on the include directive, and the transform passthrough. Link a small issue for the silent drop per CONTRIBUTING, or attach it to #2848's slice; either is fine.

If you would rather keep this PR as the record of the union design, close it and let the #2848 and #2490 work carry the capability. Your call which.

Wirasm:
Superseded by #3406, which carries this PR's include warning fix for #3216. Your original commits and authorship are preserved in the successor's history. The successor preserves the include-specific explanation, emits one author warning, and keeps includes composition-only.

The combined tree passes `bun run validate` and the docs build. Independent review and CI continue on #3406; this PR is being closed only to keep one reviewable delivery. The source issue remains open until the successor merges.

[Successor](https://github.com/coleam00/Archon/pull/3406) · [Published plan and scope mapping](https://github.com/coleam00/Archon/issues/3324#issuecomment-5760744426)

PR #3207, opened 2026-09-08
test(workflows): make fixture scope assertions independent of discovery order

coleam00:
Three fixture scope assertions failed on Linux because they compared filesystem discovery order to a fixed array order. Scope has no ordering guarantee. Compare the same exact fixture labels as a set, preserving membership, containment, symlink, and prefix-sibling boundaries.

A new test swaps two workflows between fixed folder names and verifies that both enumeration directions select the whole pack. No production behavior changes.

Validation: affected fixture runner tests 49 passed, 2 Windows skips; package type check, repository lint, and file formatting passed. The Ubuntu failures were observed in multiple independent PRs, including a provider-only change. Full repository validation had already been run during that investigation; it is not claimed green on the Windows host.

## Summary by CodeRabbit

* **Tests**
  * Improved fixture scope assertions to remain consistent regardless of filesystem enumeration order.
  * Added coverage verifying that targeting a pack selects the complete pack across workflow folder ordering variations.

### DIFF
--- a/packages/workflows/src/fixture-runner.test.ts
+++ b/packages/workflows/src/fixture-runner.test.ts
@@ -107,6 +107,16 @@ async function runFixtures(
   });
 }
 
+/**
+ * The selected scope as a set. Discovery reports fixtures in directory-enumeration order,
+ * which is a filesystem property and nothing production promises: a name-ordered
+ * filesystem and a name-hashed one disagree on the same tree. Scope assertions are about
+ * membership, so they compare sorted labels and never the enumeration itself.
+ */
+function scopeOf(report: FixtureReport): string[] {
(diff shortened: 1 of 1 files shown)


Wirasm:
Closing as superseded. #3162 merged into dev this morning and fixes the order in `fixturesUnderScope` itself, so the three assertions pass on ext4 without being loosened.

#3161 spells out why the set comparison isn't the fix we want: it hides a report order that differs per machine, and it drops the check that an `sdlc` target doesn't pick up the prefix-named `sdlc-ext` sibling. Those exact labels are there to pin both.

If ubuntu is still red after a rebase onto dev, that's a new issue with its own reproduction.

PR #3198, opened 2026-09-07
docs: clarify that archon serve requires a compiled binary

coleam00:
## Problem and outcome

The README's source-install setup led a user to try `archon serve` for the Web UI, although that command requires a compiled binary. The existing distinction was easy to miss.

- **Outcome:** The Web UI section explicitly separates binary installation commands from source installation commands.
- **Invariant:** The documented startup command must match the installation method.
- **Scope boundary:** README only; CLI and server behavior are unchanged.
- **Root cause:** Binary and source instructions were combined in a sentence whose qualifier was easy to overlook.

## Review guidance

- **Feedback requested:** Check that the two supported installation paths and their Web UI commands are clear.
- **Start here:** [`README.md:216`](https://github.com/coleam00/Archon/blob/55ad5ae1a2b60659a49129400ec4c3246fa693d9/README.md#L216).
- **Review order:** Read the Web UI callout, then the binary and source paragraphs against the surrounding setup instructions.
- **Lower-attention areas:** Markdown layout only; there are no generated files or code changes.
- **Known risk or uncertainty:** This clarification does not fix separate Windows runtime bugs.

## Solution

Add a visible compiled-binary warning and label the two startup paths. Binary users retain `archon serve`; users running from source are directed to `bun run dev` from the repository root.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A source-install user could mistake `archon serve` for the Web UI startup command. | The installation-specific command is stated explicitly before the user runs it. |
| **Failure behavior** | The user encounters the compiled-binary-only error after following setup. | The README explains the prerequisite; runtime error handling is unchanged. |

## Validation

Reviewed the README diff and its surrounding setup text. This is a documentation-only change; no application tests were rerun for the description update.

- **Not verified:** No fresh binary or source installation was performed specifically for this README-only PR.

## Links

- [Maintainer template request](https://github.com/coleam00/Archon/pull/3198#issuecomment-5582117096)

PR #3219, opened 2026-09-08
fix(cli): archon serve works from a source checkout

### LINKED ISSUE #3218: fix(cli): archon serve works from a source checkout by serving the built web dist
## Problem

`archon serve` refuses to run in a source install: `Error: archon serve is for compiled binaries only. For development, use: bun run dev` (`packages/cli/src/commands/serve.ts:53-57`). The README's Full Setup path is a source install, so a user who follows the recommended setup and then runs the documented command to get the Web UI hits that refusal. The pointer it gives, `bun run dev`, starts every package's dev server with HMR and needs a terminal held open; it is not the equivalent of `serve`.

The gate has no reason to exist. Past it, `serve` only downloads the release web dist into Archon's home and calls `startServer({ webDistPath, port })`. The server already serves a dist directory by default: with no path it uses `packages/web/dist` beside the server package (`packages/server/src/index.ts:858-874`) and warns if it is missing. That directory exists in a source checkout after `bun run build:web`.

## Why

A community user reported it this week: installed from source, every CLI test passed, `archon serve` refused, their coding agent started `bun run dev:server`, hit its own tool timeout, and concluded the Web UI does not work from source, then steered them to the CLI instead. PR #3198 proposes documenting the refusal in the README, which writes the foot-gun into the recommended path instead of removing it.

## Desired outcome

`archon serve` starts the Web UI from a source checkout the same way it does from a binary: a foreground server, the same port handling, the same output. When the built dist is missing it says exactly what to run.

## Acceptance

- [ ] In a source checkout with `packages/web/dist` built, `archon serve` starts the server serving that dist on the default or `--port` port, with no download attempted.
- [ ] In a source checkout without a built dist, `archon serve` exits non-zero with a message naming `bun run build:web`, and does not fall back to a download or to dev mode.
- [ ] Binary installs are unchanged: download on first run, checksum verification, `--download-only`.
- [ ] `--download-only` in a source checkout is an explicit refusal (nothing to download), not a silent success.
- [ ] The README Web UI section and the CLI reference describe one command for both installs, and the `bun run dev` pointer is gone from `serve`'s output.
- [ ] A test covers source-mode path selection (built dist present, built dist missing) without spawning a server.

### Reproduction

1. Clone, `bun install`, `bun link` (the README's Full Setup).
2. `archon serve`
3. `Error: archon serve is for compiled binaries only. For development, use: bun run dev`, exit code 1.

### Environment

- Archon version or commit: dev @ 2837ef0e0
- Platform or adapter: CLI, source install

- Must remain true: the binary path's checksum-verified download stays as it is.
- Must remain true: no dev-mode or HMR machinery in `serve`; it serves a built dist.
- Scope boundary: `serve.ts` and its test, the README Web UI section, the CLI reference. Not the Windows tar extraction (#3188, PR #3189), which lives inside the download path and stays valid.
- Related issues or PRs: #3198 (superseded by this), #3188, #3189, #3178 (worktree port mismatch, separate).
- Solution steering: Hint, the server's default dist resolution already points at `packages/web/dist`; in source mode `serve` can resolve that same directory and skip `downloadWebDist`.

Wirasm:
## Problem and outcome

`archon serve` refused to run outside a compiled binary and told the user to run `bun run dev` instead. The README's Full Setup path is a source install, so anyone who followed the recommended setup and then ran the documented command to get the Web UI hit a dead end. The pointer it gave is not an equivalent: `bun run dev` starts every package's dev server with HMR and holds a terminal open. A community user reported this, their coding agent tried `bun run dev:server`, hit its own tool timeout, and concluded the Web UI does not work from source.

- **Outcome:** `archon serve` starts the Web UI from a source checkout the same way it does from a binary. Same foreground server, same port handling, same output. With no built dist it exits non-zero naming `bun run build:web`.
- **Invariant:** `serve` serves a built dist. It never starts a dev server, and a source checkout never downloads a release tarball.
- **Scope boundary:** The binary path is untouched. Download on first run, embedded-checksum verification, extraction, and the `~/.archon/web-dist/<version>/` cache all behave exactly as before. No HMR or dev-mode machinery was added.
- **Root cause:** The `BUNDLED_IS_BINARY` gate guarded nothing. Everything past it needs only a directory of built web assets to hand to `startServer`, and a source checkout already has one at `packages/web/dist` after `bun run build:web`.

## Review guidance

- **Feedback requested:** Correctness of the source-mode branch, and whether moving the dist-path resolution into `@archon/paths` is the right ownership call.
- **Start here:** `packages/cli/src/commands/serve.ts:57` — the source-checkout branch that replaces the refusal.
- **Review order:** `packages/cli/src/commands/serve.ts`, then `packages/paths/src/archon-paths.ts` for the new resolver, then `packages/server/src/index.ts` where the duplicate path arithmetic goes away, then the tests.
- **Lower-attention areas:** The README, the CLI reference, and the one-line help string with its snapshot in `packages/cli/src/cli.test.ts`.
- **Known risk or uncertainty:** None. The server's default dist path is now computed in `@archon/paths` rather than inline; both derivations resolve to the same `packages/web/dist`, in a checkout and in the Docker image at `/app`.

## Solution

In a source checkout `serve` resolves `packages/web/dist`, checks it exists, and starts the server on it. Nothing is fetched, because there is no release tagged `dev` to fetch from. A missing build stops the command with the exact command to run rather than starting a server that would return an empty page, and `--download-only` is refused outright because a source checkout has nothing to download.

The path itself moved to `@archon/paths` as `getSourceWebDistDir()`. The server was already computing the same `packages/web/dist` inline as its default, so this is now one owner for the location instead of two copies of the same path arithmetic that had to agree. The server keeps its existing behavior: an explicit `webDistPath` still wins, and the default is only reached from a source checkout or the Docker image.

The server-start-and-block tail of `serveCommand` became a small helper so both installs reach it identically instead of the source branch duplicating it.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Source install: exit 1, "`archon serve` is for compiled binaries only. For development, use: bun run dev". Binary install: serves the downloaded dist. | Both installs: `archon serve` serves a built web dist in the foreground on the default or `--port` port. |
| **Failure behavior** | No failure path for a source install; the command never got that far. | No built dist: exit 1 naming the directory and `bun run build:web`. `--download-only` in a source checkout: exit 1 saying there is nothing to download. |

### Changed seams

|---|---|---|
| `@archon/paths → @archon/cli`, `@archon/server` | New `getSourceWebDistDir()` owns the built-web-UI location for both consumers. The server's inline duplicate of that path arithmetic is deleted. | `packages/paths/src/archon-paths.ts:495`, `packages/server/src/index.ts:864`, `packages/paths/src/archon-paths.test.ts:487` |

## Validation

- `bun run validate` — pass — proves type-check, lint, format, and every package suite are green on this branch.
- `bun test src/commands/serve.test.ts` in `packages/cli` — 22 pass — proves source-mode path selection with the dist present and missing, and the `--download-only` refusal, all without spawning a server.
- Guards seen red: dropping the built-dist check and serving the binary cache directory instead each turned the two new source-mode tests red; deleting the `--download-only` refusal turned its test red. Restoring the implementation returned all 22 to green.
- Runtime evidence in a source checkout, before `bun run build:web`:

```
$ archon serve
Error: Web UI is not built at /…/packages/web/dist.
Build it first: bun run build:web
$ echo $?
1
$ archon serve --download-only
Error: --download-only is for binary installs. A source checkout has nothing to download.
Build the web UI instead: bun run build:web
$ echo $?
1
```

- After `bun run build:web`, `archon serve --port 3399` served the real build: `GET /` returned 200 with the built `index.html`, `GET /assets/index-BKE0brhn.js` returned 200 and 1,584,531 bytes, and `GET /health` returned `{"status":"ok"}`. SIGTERM shut it down and the command exited 0.
- **Not verified:** The binary install path was not re-exercised end to end, because it needs a released tarball and an embedded checksum that only a build produces. Its code is unchanged apart from the extracted server-start helper, and its existing download, checksum, and extraction tests still pass.

## Links

- Closes #3218
- Supersedes #3198

## Summary by CodeRabbit

* **New Features**
  * `archon serve` now works from source checkouts, serving the locally built web dashboard.
  * Binary installations continue downloading the web dashboard on first run.
  * Clear guidance is provided when the source dashboard build is missing.

* **Documentation**
  * Updated CLI help, reference documentation, and README instructions to explain behavior for binary installations and source checkouts.
  * Clarified that `--download-only` is available only for binary installations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ The coding agent handles workflow selection, branch naming, and worktree isolati
 
 ## Web UI
 
-Archon includes a web dashboard for chatting with your coding agent, running workflows, and monitoring activity. Binary installs: run `archon serve` to download and start the web UI in one step. From source: ask your coding agent to run the frontend from the Archon repo, or run `bun run dev` from the repo root yourself.
+Archon includes a web dashboard for chatting with your coding agent, running workflows, and monitoring activity. Run `archon serve` to start it, whichever way you installed. A binary downloads the matching web UI on first run. A source checkout serves the copy you build: run `bun run build:web` once from the repo root, then `archon serve`.
 
 Register a project by clicking **+** next to "Project" in the chat sidebar - enter a GitHub URL or local path. Then start a conversation, invoke workflows, and watch progress in real time.
 
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -395,7 +395,7 @@ Commands:
   isolation cleanup [days]   Remove stale environments (default: 7 days)
   isolation cleanup --merged Remove environments with branches merged into main
   complete <branch> [...]    Complete branch lifecycle (remove worktree + branches)
-  serve                      Start the web UI server (downloads web UI on first run)
+  serve                      Start the web UI server (binary installs download it on first run)
   skill install [path]       Install archon-cli into .claude/skills and .agents/skills
   doctor [--full]            Verify your Archon setup (Claude/Codex binaries, gh auth, DB, adapters; --full also probes the OpenCode runtime SDK)
   auth github                Connect your GitHub identity via device flow (multi-user installs)
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -338,7 +338,7 @@ const commandHelp: HelpEntry[] = [
   {
     command: 'serve',
     spec: 'serve',
-    description: 'Start the web UI server (downloads web UI on first run)',
+    description: 'Start the web UI server (binary installs download it on first run)',
   },
   {
     command: 'skill',
--- a/packages/cli/src/commands/serve.test.ts
+++ b/packages/cli/src/commands/serve.test.ts
@@ -9,12 +9,14 @@ import {
   afterEach,
   spyOn,
 } from 'bun:test';
-import { existsSync, mkdtempSync, readFileSync, rmSync } from 'fs';
+import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
 
 // Mock @archon/paths BEFORE importing the module under test.
-// This sets BUNDLED_IS_BINARY = false (dev mode) so serveCommand rejects.
+// BUNDLED_IS_BINARY = false puts serveCommand on its source-checkout path, and
+// getSourceWebDistDir is redirected at a temp tree so a test can decide whether
+// `bun run build:web` has been run without depending on this checkout's state.
 const mockLogger = {
   fatal: mock(() => undefined),
   error: mock(() => undefined),
@@ -23,14 +25,27 @@ const mockLogger = {
   debug: mock(() => undefined),
   trace: mock(() => undefined),
 };
+let sourceWebDistDir = '/tmp/test-archon/unset-source-web-dist';
 mock.module('@archon/paths', () => ({
   createLogger: mock(() => mockLogger),
   getWebDistDir: mock((version: string) => `/tmp/test-archon/web-dist/${version}`),
+  getSourceWebDistDir: mock(() => sourceWebDistDir),
   BUNDLED_IS_BINARY: false,
   BUNDLED_VERSION: 'dev',
   BUNDLED_WEB_DIST_SHA256: '',
 }));
 
+// serveCommand reaches the real server through a dynamic import. Stub it so the
+// source-mode tests can read back which dist it was handed without opening a
+// listening socket.
+const startServerCalls: Array<{ webDistPath: string; port?: number }> = [];
+mock.module('@archon/server', () => ({
+  startServer: mock(async (opts: { webDistPath: string; port?: number }) => {
+    startServerCalls.push(opts);
+  }),
+}));
+
+import { trackTempRoots } from '@archon/paths/test-utils';
 import { serveCommand, parseChecksum, parseEmbeddedChecksum, downloadWebDist } from './serve';
 
 describe('parseChecksum', () => {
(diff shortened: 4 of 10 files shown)


Wirasm:
## Ready to merge

`archon serve` now serves the locally built `packages/web/dist` in a source checkout instead of refusing outright, resolving that path through one shared `getSourceWebDistDir()` owner in `@archon/paths` that both the CLI and the server package call. Independent verification confirms the binary install path (download, checksum verification, extraction, `--download-only`) is byte-for-byte unchanged, the source-mode path arithmetic resolves identically to what the server used to compute inline (checked both in a normal checkout and against the Docker image layout), the new tests genuinely cover source-mode path selection without spawning a server, and reverting the old `BUNDLED_IS_BINARY` short-circuit turns exactly those three new tests red. All six acceptance items on issue #3218 are met. Three non-blocking Suggestions remain, all about robustness beyond this PR's stated scope rather than defects it introduces.

**Validation:** all green — see coverage table below.

### Findings

|---|---|---|---|
| `R1` | Suggestion | `startServer()`'s own missing-dist check only warns; `bun run start` (a separately documented production path) can silently serve a broken web UI | OPEN |
| `R2` | Suggestion | New source-mode guard checks that the dist directory exists, not that it holds a completed build | OPEN |
| `R3` | Suggestion | Extracting `startServerUntilSignal` drops the `version` field from the binary-path start-failure log | OPEN |

<details>
<summary><code>R1</code> — <code>bun run start</code> bypasses the new build-presence guard</summary>

**Impact:** `startServer()` has exactly two callers: the CLI's `serveCommand` (now guarded) and `packages/server/src/index.ts`'s own `import.meta.main` direct-execution branch, which is what `bun run start` invokes. That branch's existing missing-dist handling (`packages/server/src/index.ts:866-868`) only `warn`s and continues — it was never a hard failure, and this PR did not change that. `packages/docs-web/src/content/docs/deployment/local.md` documents `bun run build` then `bun run start` as the local production path, independent of `archon serve`. Skip the build step and run `bun run start` directly, and the server binds the port and answers `/health` while returning 404/empty for every UI route — the same class of dead end #3218 was filed over, just reached through a different documented command.

**Evidence:** `packages/server/src/index.ts:861-868` (unchanged warn-only check); `packages/server/src/index.ts:1093` and `packages/cli/src/commands/serve.ts:112` (the only two `startServer()` call sites); `packages/docs-web/src/content/docs/deployment/local.md` ("Production Build (Local)" section).

**Required outcome:** none required of this PR. This behavior predates it — the PR only changed how the default path is *computed* (from a duplicate inline expression to the shared `getSourceWebDistDir()`), not the warn-then-continue handling around it. Issue #3218's own scope boundary is explicit: "serve.ts and its test, the README Web UI section, the CLI reference" — `bun run start`'s direct-execution path was never part of the contract this PR is closing against.

**Disposition:** OPEN, non-blocking. Genuinely valuable and worth a follow-up (promote the server's own missing-dist check to a hard failure, matching the CLI's new strictness), but not a defect this PR introduced or was asked to fix. Left open as a Suggestion rather than filed as a tracked issue — no GitHub issue created without the operator's say-so.

</details>

<details>
<summary><code>R2</code> — missing-dist guard checks presence, not completeness</summary>

**Impact:** `packages/cli/src/commands/serve.ts:68` uses `existsSync(webDistDir)`, which is satisfied by an empty or partial `packages/web/dist` (an interrupted `bun run build:web`, stray tooling residue — `dist/` is gitignored so nothing prevents this). The binary/download path in the same file already checks for `index.html` specifically after extraction (`serve.ts:301-307`); the new source-mode check doesn't apply the same standard, so a partial build passes silently and the server then serves 404s with no error, which is quieter than the "not built at all" case this PR exists to fix.

**Evidence:** `packages/cli/src/commands/serve.ts:68` vs. `packages/cli/src/commands/serve.ts:301-307`.

**Required outcome:** check `existsSync(join(webDistDir, 'index.html'))` instead of the bare directory, matching the verification the download path already performs.

**Disposition:** OPEN, non-blocking. Narrow trigger (requires an interrupted or manually-tampered build), cheap to fix if the author wants to close it now or in a follow-up.

</details>

<details>
<summary><code>R3</code> — start-failure log loses the <code>version</code> field on the binary path</summary>

**Impact:** Before this PR, a binary-install start failure logged `{ err, version, webDistDir, port }`. The extracted `startServerUntilSignal` helper has no `version` parameter, so the same failure now logs `{ err, webDistDir, port }`. `webDistDir` embeds the version in its path (`~/.archon/web-dist/<version>/`), so no information is actually lost, but it's one string-match away instead of a structured field.

**Evidence:** `packages/cli/src/commands/serve.ts:118` vs. the pre-PR log call.

**Required outcome:** optional — pass `version` through to the helper (or fold it into `webDistDir`'s structured value) if the author wants the field back; not required for correctness.

**Disposition:** OPEN, non-blocking, cosmetic.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | No findings — binary path confirmed byte-identical, source-mode path arithmetic confirmed to resolve identically to the old inline computation in both a checkout and the Docker layout, no fallback path on missing dist, `--download-only` refusal confirmed unconditional |
| `seams` | `R1`, `R2` — single owner of the source-mode path confirmed via repo-wide search, `getSourceRepoRoot()` extraction confirmed byte-identical to the code it replaced |
| `simplify` | `R3` — `getSourceRepoRoot()`/`getSourceWebDistDir()` split and `startServerUntilSignal` extraction both judged proportionate to their real call sites; test machinery in `serve.test.ts` judged proportionate after checking two lighter alternatives |

#### Validation

|---|---|---|
| `bun test src/commands/serve.test.ts` (packages/cli) | PASS | 22 pass, 0 fail, 47 expect() calls |
| `bun test src/archon-paths.test.ts` (packages/paths) | PASS | 144 pass, 0 fail, 188 expect() calls |
| `bun run validate` (repo root) | PASS | exit code 0, no failing suites across all packages |
| Guard check: restore old `BUNDLED_IS_BINARY` short-circuit refusal | RED as expected | exactly the 3 new source-mode tests fail (`serves the locally built dist...`, `refuses with a build instruction...`, `refuses --download-only...`); all 22 pass again once reverted |
| Manual path-arithmetic check: old inline server computation vs. new `getSourceWebDistDir()` | MATCH | both resolve to identical absolute path in this checkout; Dockerfile confirms same sibling layout under `/app/packages/` |

</details>

PR #3221, opened 2026-09-08
fix(cli): preserve workflow source ownership during adoption

coleam00:
## Problem and outcome

Adopting a run could discard its workflow source in two ways: detached recapture deleted its own replacement snapshot, and an explicit `--workflow-source` was overwritten by the adopted checkout. A source-only workflow then disappeared, or a conflicting project workflow replaced it.

- **Outcome:** Foreground and detached adoption preserve an explicitly selected source while executing in the adopted worktree. Default adoption still discovers workflows from that worktree.
- **Invariant:** A new adoption captures current authoring bytes; resume uses that run's frozen capture. Commands, scripts, required inputs and source configuration retain the same owner.
- **Scope boundary:** CLI source selection and capture cleanup. Existing isolation, provider and database contracts remain intact.
- **Root cause:** Recapture ran unconditionally after lane resolution, and detached captures could share the same path as the stale capture being removed.

## Review guidance

- **Feedback requested:** Source ownership across explicit selection, default adoption, detached forwarding and resume.
- **Start here:** `packages/cli/src/commands/workflow.ts`, the adoption guard around `recaptureForLane`.
- **Review order:** Source-selection guard, same-root cleanup guard, command regressions, then `workflow-adopt-source.integration.spec.ts`.
- **Lower-attention areas:** Package test-command wiring and the fixture that suppresses unrelated AI title generation.
- **Known risk or uncertainty:** The real CLI fixtures seed prior-run isolation metadata to exercise adoption deterministically. They use actual source capture, git worktrees, scripts, detached processes and resume, with no provider execution.

## Solution

Recapture from the adopted checkout only when the caller did not supply an explicit source. Remove the old capture only when its root differs from the replacement. The explicit source remains independent of the execution checkout, including when both initially name the same directory.

The CLI regression matrix covers foreground and detached adoption with missing or conflicting target workflows. It verifies freshly selected bytes, command-folder configuration, required inputs, recorded source origin and preservation of dirty tracked/untracked target files. After both live source trees change, resume must still execute the captured bytes. Companion cases preserve default lane discovery and distinct-root cleanup.

## Validation

- `bun test src/commands/workflow.test.ts` from `packages/cli`: 407 passed, 952 assertions after integrating the current dev branch.
- `bun test src/commands/workflow-adopt-source.integration.spec.ts` from `packages/cli`: six public CLI integration cases passed, including detached adoption and frozen-source resume.
[...]

PR #3222, opened 2026-09-08
test(cli): bound startup fixture subprocesses

coleam00:
## Problem and outcome

The import-graph fixture and caller-relative workflow-path fixture use synchronous child processes without deadlines. A hung child can block the test process indefinitely.

- **Outcome:** Each child process has a 20-second deadline.
- **Invariant:** Existing assertions and Bun's default test/setup budgets remain unchanged.
- **Scope boundary:** Two process options in `packages/cli/src/cli.test.ts`. Telemetry shutdown and removal of avoidable fixture startup cost belong to #3225.
- **Root cause:** Both `spawnSync` calls omitted a timeout.

## Review guidance

- **Feedback requested:** Whether the two finite process deadlines adequately bound these fixtures.
- **Start here:** `packages/cli/src/cli.test.ts`, `buildImportGraph` and `workflow test path targets`.
- **Review order:** The two `spawnSync` option objects.
- **Lower-attention areas:** No generated files or assertion changes.
- **Known risk or uncertainty:** This does not fix telemetry-induced startup delays or remove compilation/subprocess work from the fixtures. #3225 tracks those changes.

## Solution

Set `timeout: 20000` on the two synchronous subprocess calls. Removed the previously proposed 30-second test/setup budgets following maintainer feedback. Preserve the existing five-second test budgets and keep the process bound independent of test scheduling.

## Validation

- Native Bun 1.4.2 on Windows, from `packages/cli`: `bun test ./src/cli.test.ts` passed 129 tests and 317 assertions after removing both enlarged budgets. The caller-relative test completed in 2.18 seconds.
- `git diff --check` passed. Changed-file lint and formatting passed through the commit hook.
- [CI on narrowed head cfa7d355](https://github.com/coleam00/Archon/actions/runs/34238390052) passed Windows, Ubuntu, workflow fixtures, types, lint, formatting, schema upgrades, Postgres parity and Docker. [Combined integration CI](https://github.com/coleam00/Archon/actions/runs/34238417160) also passed with all six reviewed PR heads and the original test budgets restored.
- **Not verified:** Telemetry shutdown latency or a fully deterministic startup-cost fix. Those are #3225's scope; this PR only bounds hung child processes.

## Links

- Related #3225, #2924 and #3221

## Summary by CodeRabbit

* **Tests**
  * Added 20-second execution limits to CLI subprocesses used by startup import-boundary and workflow path tests.
  * Adjusted timeout handling so these limits apply directly to the relevant test processes, helping prevent indefinite hangs while keeping test execution predictable.

### DIFF
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -91,7 +91,7 @@ describe('CLI startup import boundary', () => {
       join(repoRoot, 'packages/core/src/config/run-config-handoff.ts'),
       join(buildDir, 'handoff')
     );
-  }, 30000);
+  });
 
   afterAll(async () => {
     if (buildDir) await removeTempTree(buildDir);
@@ -1990,5 +1990,5 @@ describe('workflow test path targets', () => {
     } finally {
       await removeTempTree(repo);
     }
-  }, 30000);
+  });
 });
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -1972,6 +1972,7 @@ describe('workflow test path targets', () => {
         [CLI_ENTRY, 'workflow', 'test', '../.archon/workflows/sdlc/plan', '--cwd', tools, '--json'],
         {
           encoding: 'utf8',
+          timeout: 20000,
           env: {
             ...process.env,
             ARCHON_TELEMETRY_DISABLED: '1',
@@ -1989,5 +1990,5 @@ describe('workflow test path targets', () => {
     } finally {
       await removeTempTree(repo);
     }
-  });
+  }, 30000);
 });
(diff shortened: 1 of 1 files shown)


Wirasm:
Leaving this open for you, with what I would keep and what I would not.

**Keep the two 20 s process deadlines.** Those `spawnSync` calls had no bound, so a hung child hung the suite. That part is right.

**Drop the two 30 s test budgets.** They budget for a cost that should not exist, and the repo's own record on this family says so: #2924's other two families were closed by removing the cost from the test, not by raising the budget.

**The cost has a name now.** Every `archon` exit path awaits `shutdownTelemetry()`, which awaits the PostHog client's network flush with no deadline of our own. Measured: `archon --help` takes 1.73 s with telemetry on and 0.05 s with it disabled; the CLI's static startup is 22 modules. `cli.test.ts` spawns the real CLI 35 times and 18 of those spawns do not set `ARCHON_TELEMETRY_DISABLED`, so each pays that flush on a contended runner. That is #3225: bound the flush at the exit path, move the compile-in-`beforeAll` import-boundary check under `bun run validate`, and test the caller-relative path resolution in process instead of spawning the CLI.

Recommend narrowing this PR to the process deadlines and letting #3225 carry the rest, or closing it in favour of #3225. Your call which.

Issue #3226, opened 2026-09-08
feat(sdlc): one-shot regression diagnosis from shared validation and investigation


coleam00:
## Problem

The SDLC pack can validate a checkout and investigate a symptom, but no shared entry point combines those capabilities into a one-shot regression diagnosis. Consumers currently need their own glue to distinguish a product defect from unavailable infrastructure and to preserve findings for later tracker review.

## Why

The software factory should consume the shared pack directly. A regression check is a concrete missing use case as its independent agent runner and stage dispatcher are removed. Detection must stay separate from the discovery publication work in #3215.

## Desired outcome

A manually invoked, standalone SDLC workflow checks an explicit repository revision, investigates a demonstrated failure only when needed, and writes an evidence-backed diagnosis and discovery sidecar. It leaves source and tracker state unchanged.

## Acceptance

- [ ] A healthy repository produces a healthy result with the checks that actually ran and the checked revision.
- [ ] A reproduced product defect produces a diagnosis with evidence suitable for the discovery proposal workflow.
- [ ] Missing checks, unavailable infrastructure, and unreproduced symptoms remain distinct from a proven product regression.
- [ ] Applicable project runtime scenarios can be included through the shared runtime verification workflow; no agent is launched from a script.
- [ ] Every agent node uses a command file and native model tiers.
- [ ] Meaningful fixtures cover healthy, product failure, infrastructure failure, no checks, and repeated findings; a real isolated negative control confirms the failure path.
- [ ] Diagnosis never creates, edits, or closes a tracker item and never modifies the application to make a check pass.

## Constraints and related work

- Must remain true: YAML owns composition, existing validate/investigate retain their responsibilities, and scripts perform narrow computation.
- Scope boundary: one-shot diagnosis, without scheduling, repair, tracker publication, or a separate orchestration service.
- Known prerequisite: optional runtime verification consumes the separately developed shared runtime workflow. Any required reusable engine capability belongs in its own issue and PR.
- Related: #3215 (discovery proposals/publication), #998 (future standing trigger).
- Solution steering, requirement: compose the existing pack responsibilities instead of reviving the closed experimental regression workflow. Keep any producer contract change explicit and narrowly justified.

Issue #3211, opened 2026-09-08
design(sdlc): a merge queue workflow that orders, readies, and merges accepted PRs across forges


Wirasm:
## Problem

Every SDLC path ends at "PR flipped ready for review" and nothing owns the step after it. Merge order, conflict repair, re-verification when the base moves under a PR, and the merge itself are done by hand. On 27 and 28 August that sequence ran manually about five times in 24 hours. Twice a green PR merged onto a green base and produced a red composed result, because nothing tests the composition before landing (#2596). The forge's own merge queue only serializes and tests; it does not rank, repair conflicts, or hold a human gate, and it exists on one forge.

## Why

Parallel agent lanes produce several ready PRs an hour. Without an owner for the tail, the operator is the queue. The delivery tail is stable enough to outrun hand merging, and the pack is heading to its own repository, so its end-to-end shape has to be settled.

## Desired outcome

A pack workflow that keeps a queue of ready PRs, orders them, brings each one to mergeable by the repository's own standards rather than by the forge saying green, and merges through two gates: an auto gate for small, bounded, low-risk changes and a human gate for everything else. Queue state lives in a JSON file the workflow writes; any surface (console, a plain HTML page) renders from that file.

## Acceptance

Design acceptance, since this issue is design first:

- [ ] The shape below is confirmed or replaced, with every open question answered on this issue.
- [ ] The owning layer is named for each of: queue state, the trigger, the merge operation, the auto-gate values.
- [ ] A first slice is defined that needs no engine work: manual trigger, human gate for every PR, pinned merge, composed-result checks, queue file written at every transition.
- [ ] The auto-gate values start as: small and bounded, risky, large. Kept simple; refined only from dogfood evidence.

## Constraints and related work

- Must remain true: works across forges through the forge contract (#3040, #3041, #3042). No forge-native merge queue, no GitHub-only mechanism.
- Must remain true: merge order for anything past the auto gate is operator judgment at a declared approval gate. No blind merges.
- Must remain true: a PR merges only at a pinned head and base, with checks re-verified at that head, and the composed result (PR onto the current base) tested before the merge, not after.
- Must remain true: the queue file is the single source of truth for queue state. The workflow never infers queue state from labels, prose, or branch names.
- Must remain true: no scheduler inside the pack. Standing operation waits on triggers (#998).
- Scope boundary: this is the pack's merge tail. Provider admission and capacity queueing (#2180, #2816, #2817) are engine work and are not this.
[...]

PR #3227, opened 2026-09-08
feat(sdlc): verify live runtime scenarios with explicit agent nodes

coleam00:
## Problem and outcome

The SDLC pack can run ordinary project checks but lacks a reusable workflow for an agent to exercise a live application against declared scenarios. `archon-verify-runtime` adds that responsibility with an explicit command node, target identity, per-assertion evidence and a typed result.

- **Outcome:** `verified`, `failed`, or `inconclusive` for the declared assertions, with the report and tool evidence retained in run artifacts.
- **Invariant:** A real product failure is not retried as a malformed report. A different target, missing coverage or malformed evidence cannot produce a verified result.
- **Scope boundary:** One standalone runtime workflow. The project supplies scenario data and trusted environment commands; an external owner remains responsible for resource lifetime after cancellation or host loss.

## Review guidance

- **Feedback requested:** Runtime contract, failure semantics, model/structural judgment boundary and standalone reuse.
- **Start here:** `.archon/workflows/sdlc/verify-runtime/archon-verify-runtime.yaml`.
- **Review order:** Graph and command, scenario/evidence scripts, README contract, then the `archon-verify-runtime live target contract` executor tests.
- **Lower-attention areas:** Generated bundle changes.
- **Known risk or uncertainty:** Evidence presence cannot prove a model's observations. `mutates_checkout: false` detects checkout changes through existing engine behavior; it is not a filesystem sandbox or confidential-holdout boundary. YAML teardown is not guaranteed after cancellation.

## Solution

Each attempt runs setup, start, the medium-tier verification command, a project-provided target-identity probe, evidence checks and teardown in that order. The agent writes a JSON report into a new attempt directory. Deterministic scripts check exact assertion coverage, report shape, current-attempt evidence paths and target identity; the agent judges the observed behavior.

Only a malformed report receives one fresh attempt. Product failure, unavailable target, wrong requested identity and failed cleanup stop without that retry. A shared YAML bound lets the second malformed attempt return `inconclusive` before the engine exhausts the loop. Native operational provider failures retain their engine semantics.

Identity is normalized consistently across the expected input, probe and report. A live test caught the original contradiction between a literal CRLF command output and a trimmed probe; an executor regression now fails without the correction. Local checkout HEAD is separate metadata and never substitutes for deployed identity.

## Validation

- The corrected runtime passes 25 actual-executor cases with 899 assertions. Fresh retries receive the exact report-rejection reason; the evidence checker still requires one real file, including legitimate filenames containing commas.
- Two real medium-provider suite runs on a Python/SQLite project passed the baseline and independent holdout checks and caught a deliberately broken inventory calculation. The second run also exercised malformed/negative body lengths on the candidate fix. The standalone runtime resources are byte-identical to those used by the suite; suite composition is a separate PR.
- [GitHub validation](https://github.com/coleam00/Archon/actions/runs/34254532732) passed on Linux and Windows at `98478467`, including fixtures, schema upgrade, PostgreSQL parity and Docker build.
- Actual DAG executor with real Python/shell commands and disposable HTTP targets: 23 tests passed, 664 assertions. Covers successful and failed observations, LF/CRLF identities, wrong target/report, malformed retry/exhaustion, missing coverage/evidence, unavailable startup, teardown failure, cancellation and provider failure. Provider judgment is simulated in these tests.
- Loader/include/fan-out checks: 509 passed, one existing skip. Bundle tests: 34 passed, four existing skips. Generated bundle checks and commit hooks passed.
- Two real-provider runs against a disposable Python/SQLite application passed the final workflow contract: the healthy inventory journeys returned `verified`; a reproducible HTTP disconnect defect returned `failed`. Each used exactly one verification attempt, retained actual HTTP evidence and the probed candidate identity, and the external test owner stopped both servers.
- `bun run validate` passed generated checks, all type checks, lint, formatting and installer tests, then stopped in the parallel test stage on inherited Windows failures.
- **Not verified:** A green local monorepo suite, independent proof of arbitrary model evidence, or engine-owned cleanup on every terminal path. The unchanged provider symlink test fails with `EPERM`; an unchanged container mount-overlap test reproduces a Windows slash-normalization defect. Neither was hidden by enlarging existing test budgets.

## Delivery considerations

| Concern | Impact and required action |
| --- | --- |
| Environment lifetime | Scenarios explicitly declare external ownership. Consumers must clean up after cancellation, node failure and host loss; normal teardown alone is insufficient. |
| Target identity | The project probe must observe the actual running target or tested executable. An empty expected candidate only checks report/probe agreement. |
| Result consumption | Inspect `verified` and `verdict`. A completed workflow lifecycle can still carry a failed authored outcome. |
| Confidential scenarios | This workflow provides no secret-isolation guarantee. Ordinary independent scenarios are supported. |

## Links

- Related #3211 and #3226

Latest head `d03763d1` fixes a CRLF-sensitive test assertion exposed by Windows CI. Two actual executor cases passed with the command file deliberately using CRLF, including malformed-report retry. The production command and checker are unchanged. Earlier green CI applies to the earlier head; current-head CI remains the merge requirement.


coleam00:
Closing to re-open one workflow at a time after #3127 lands, rewritten with TypeScript scripts. The design and the live evidence from this branch are now #3263. The working version stays on cleanup/sdlc-workflows-only, which coleam00/ai-software-factory pins.

Issue #3215, opened 2026-09-08
design(sdlc): discovery publication turns proved run findings into tracker items under a human batch gate


Wirasm:
## Problem

Runs already produce discoveries: review consolidates per-node sidecars into `discoveries.json` and `discoveries.md`, and the terminal outcome points at them. Nothing turns those findings into tracker items. They either disappear with the run or get filed by hand. #3206 (archon-regress) shows the pressure: it fuses regression detection, diagnosis, and issue creation into one 1338-line script with a deterministic publication guard and no human gate.

## Why

Evidence that was paid for must not vanish, and it must not become issue spam either. The 27 August design settled that agents never create forge issues while implementing or reviewing, and that publication is a separate class that revalidates, deduplicates, and proposes before anything is written. Unattended intake (standing regression, tracker sweeps) needs that class to exist before it can file its own bugs.

## Desired outcome

A pack workflow that takes a run id or a discovery artifact, revalidates each claim against current state, searches the configured forge for existing work, classifies each item as stale, duplicate, update-existing, or new, and produces a proposed action set. Creating or updating tracker items happens only under explicit operator policy or a human batch gate. Forge-agnostic through the forge contract; GitHub is one adapter.

## Acceptance

Design acceptance, since this issue is design first:

- [ ] The input contract (run id or artifact path), the classification vocabulary, and the action-set shape are settled on this issue.
- [ ] The gate is settled: which actions need the batch gate, which may run under explicit policy, and how a repository without a forge configured behaves.
- [ ] Dedup ownership is named: the marker written into tracker items, and the search-before-write rule.
- [ ] A first slice is defined that writes nothing: it reads the artifact, classifies, and renders the proposed action set for a human.
- [ ] Fixtures cover new, duplicate, stale, and update-existing.

## Constraints and related work

- Must remain true: agents never create forge items while implementing or reviewing. Discoveries stay in the run's sidecar; publication is a separate run.
- Must remain true: no global discovery database, no cross-run state, no forge receipts, until dogfood proves the per-run sidecar insufficient.
- Must remain true: one marked, idempotent tracker write per item; a re-run updates its own marker rather than duplicating.
- Must remain true: public text never carries local paths or artifact locations.
- Scope boundary: publication only. Detecting regressions on a schedule is a separate class (standing watchdog, gated on #998 triggers and #2744 durable waits) and gets its own issue if wanted.
[...]

PR #3231, opened 2026-09-08
feat(sdlc): revalidate and publish discoveries through workflow commands

coleam00:
## Problem and outcome

Verified discoveries should become useful follow-up issues without requiring a new forge engine API. This workflow now owns proposal generation and optional publication through `gh`.

## Review guidance

Start with `.archon/workflows/sdlc/discoveries/archon-discoveries.yaml`, then `commands/publish-discoveries.md`. Review the authorization, repeated deduplication, and partial-write reporting. This folds the publication responsibility from #3242 into this workflow; #3240 is not a prerequisite.

## Solution

Keep the existing revalidation, evidence checks and search. Add a medium command node that creates issues or comments using scoped GitHub CLI calls, body files, markers and read-back. `publication=preview` is the default; `approve` uses a native approval gate; `auto` explicitly authorizes this run subject to project guidance. Explicit discovery artifact input works with existing Archon. Optional run-ID lookup needs a CLI exposing artifact locations and is not required by the factory.

## Validation

Regenerated the bundled workflow resources and inspected the source diff. No tests or agent runs were performed for the simplification, by request. Earlier proposal tests do not validate the new publication behavior. This remains draft pending verification.


coleam00:
Closing. #3215 settles the design: forge-agnostic through the contract, a human batch gate, one marked idempotent write per item. Publishing through gh from an agent node is the pattern the 31 Aug decision moved into the audited deterministic lane, so this shape does not land. Preview mode on this branch is exactly #3215's slice one and will come back as that. The working version stays on the factory pin.

PR #3233, opened 2026-09-08
fix(workflows): follow symlinks when deciding whether a scope root is a directory

### LINKED ISSUE #3184: fix(workflows): a symlinked ~/.archon/workflows is dropped from the source capture, so no home workflow can run
`captureWorkflowSource` decides whether to capture a scope root with `isDirectory`, which calls
`lstat`. `lstat` does not follow symlinks, so a **symlinked `~/.archon/workflows`** — the layout the
documentation recommends — reports `false`, the home scope is never captured, and every
home-scoped workflow disappears from `archon workflow run`.

`archon workflow list` still shows them, because listing does not take a capture. That is what makes
this hard to spot: the workflow is visible right up to the moment you try to run it.

## Reproduce

```bash
mkdir -p /tmp/wf && cp <any workflow>.yaml /tmp/wf/
rm -rf ~/.archon/workflows && ln -s /tmp/wf ~/.archon/workflows   # the layout the docs recommend

archon workflow list | grep <name>     # present
archon workflow run <name> "x"
# Discovery: root=<repo> workflows=32 bundled=32 global=0 project=0
# Error: Workflow '<name>' not found.
```

Observed on `8a5c9055` (v0.9.0, source build, macOS 15.5). Eight home-scoped workflows, all
un-runnable; `archon workflow list` reports `home_workflows_loaded count=8` in the same shell.

## Cause

`packages/workflows/src/workflow-source.ts`

```ts
// l.308-314
async function isDirectory(path: string): Promise<boolean> {
  try {
    return (await lstat(path)).isDirectory();   // ← does not follow the symlink
  } catch {
    return false;
  }
}

// l.526-533 — the home scope is skipped when the guard says false
for (const [name, from] of [
  ['workflows', archonPaths.getHomeWorkflowsPath()],
  ['commands',  archonPaths.getHomeCommandsPath()],
  ['scripts',   archonPaths.getHomeScriptsPath()],
] as const) {
  if (await isDirectory(from))
    jobs.push({ dest: join(GLOBAL_SCOPE_DIR, name), from, scope: 'global' });
}
```

The same file already states the intended behaviour forty lines below the guard, at l.340:

> `// Symlinks are DEREFERENCED into ordinary files. Preserving the link would keep a live`
> `// reference to a path outside the capture … stat (not lstat) resolves the target`

So `copyTree` deliberately dereferences symlinks; only the entry guard does not. The two are
inconsistent, and the guard is the one that silently drops the scope — nothing is logged at
`warn`, and discovery simply reports `global=0`.

Note this affects the **root** of a scope only. A real directory containing symlinked *files* is
captured correctly, because `copyTree` uses `stat`. So the failure is specific to symlinking the
directory itself — which is exactly what the guide recommends.

## The recommended layout is the one that breaks

`https://archon.diy/guides/global-workflows/`:

> Then symlink during dotfiles setup: `ln -sf ~/dotfiles/archon/workflows ~/.archon/workflows`

There is no config key to point elsewhere, so a user following the guide has no supported
alternative.

## Suggested fix

`stat` instead of `lstat` in `isDirectory` (l.310). `stat` is already imported in this file (l.47),
and a dangling link then throws and is caught by the existing `catch`, returning `false` as before.

```diff
-    return (await lstat(path)).isDirectory();
+    return (await stat(path)).isDirectory();
```

Verified locally: with that one change, `Discovery: workflows=40 bundled=32 global=8 project=0` and
the runs work again.

Two things worth deciding alongside it, which I have not assumed:

- Whether a scope root that exists but is not a directory should log a `warn`. Today the scope is
  dropped with no signal at all, which is why this went unnoticed from the release of #2660 until a
  run was attempted.
- Whether `getHomeCommandsPath()` / `getHomeScriptsPath()` deserve the same treatment — they go
  through the same guard, so a symlinked `~/.archon/commands` is dropped identically.

## Related

- #2660 (merged) — *runs read their workflow source from a copy they own, not from the target* —
  introduced the capture, and with it this guard.
- #1503 (closed, not merged) — *fix(paths): findMarkdownFilesRecursive follows symlinks* — the same
  class of bug on another function, so this may be worth fixing as a rule rather than case by case.

Happy to open the one-line PR if that is useful.

BluegReeno:
Fixes #3184.

## The bug

`captureWorkflowSource` skips a scope whose root is not a directory, and `isDirectory` asks `lstat`, which does not follow symlinks.

The [global-workflows guide](https://archon.diy/guides/global-workflows/) tells people to symlink a scope root out of a dotfiles repo:

```bash
ln -sf ~/dotfiles/archon/workflows ~/.archon/workflows
```

With that layout the home scope reports "not a directory", is never captured, and every home-scoped workflow disappears the moment you try to run one:

```
Discovery: root=<repo> workflows=32 bundled=32 global=0 project=0
Error: Workflow '<name>' not found.
```

`archon workflow list` still lists them, because listing takes no capture — which is what makes the failure hard to place. Observed on `8a5c9055` with eight home workflows, all un-runnable, while `list` reported `home_workflows_loaded count=8` in the same shell.

The same guard governs the project scope (`workflow-source.ts:510-514`), so a symlinked `.archon/workflows` inside a repo is dropped identically. Only the scope **root** is affected — a real directory holding symlinked *files* captures fine, because `copyTree` uses `stat`.

## Why this is an inconsistency rather than a policy

Forty lines below the guard, `copyTree` already dereferences symlinks on purpose:

> `// Symlinks are DEREFERENCED into ordinary files. Preserving the link would keep a live`
> `// reference to a path outside the capture … stat (not lstat) resolves the target`

So the module's stated intent is to follow links; only the entry guard did not.

## The change

One line, plus the comment explaining why the choice matters here. `stat` was already imported; `lstat` becomes unused and is dropped. A dangling link still throws into the existing `catch` and returns `false`, unchanged.

## Test

The added test symlinks a project scope root, then asserts the scope is captured and its bytes dereferenced into the capture.

It is a real regression test, not a passing assertion: reverting the one-line change fails it with

```
Expected to contain: "project"
Received: [ "global", "bundled" ]
```

— the scope silently absent, which is exactly the shape of the bug.

`bun test packages/workflows/src/workflow-source.test.ts` → **38 pass, 0 fail** on this branch.

## Two things I did not assume

- Whether a scope root that exists but is not a directory should log a `warn`. Today the scope is dropped with no signal at all, which is why this went unnoticed between #2660 and the first attempted run. Happy to add one if you want it.
- `getHomeCommandsPath()` and `getHomeScriptsPath()` go through the same guard and are fixed by the same line, but the test covers the project scope only, since asserting on the home scope would read the developer's real `~/.archon`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workflow captures for scopes whose root directory is a symbolic link.
  * Symlinked workflow directories are now recognized correctly, included in manifests, and copied with their files.

### DIFF
--- a/packages/workflows/src/workflow-source.test.ts
+++ b/packages/workflows/src/workflow-source.test.ts
@@ -306,6 +306,37 @@ describe('captureWorkflowSource', () => {
     );
   });
 
+  test('captures a scope whose root is a symlinked directory', async () => {
+    // The global-workflows guide tells people to symlink a scope root out of a dotfiles
+    // repo (`ln -sf ~/dotfiles/archon/workflows ~/.archon/workflows`). An `lstat` guard
+    // reports such a root as "not a directory" and drops the whole scope from the
+    // capture — silently, since a missing scope is a legitimate state. `copyTree` already
+    // dereferences symlinks, so the entry guard has to as well.
+    const { root, runArtifacts } = await createSandbox();
+    const sourceRoot = join(root, 'linked-source');
+    const realCommands = join(root, 'dotfiles', 'commands');
+    await mkdir(realCommands, { recursive: true });
+    await mkdir(join(sourceRoot, '.archon'), { recursive: true });
+    await writeFile(join(realCommands, 'review.md'), 'review the diff');
+    await symlink(realCommands, join(sourceRoot, '.archon', 'commands'), 'dir');
+
+    const capture = await captureWorkflowSource({
+      sourceRoot,
+      captureRoot: captureRootIn(runArtifacts),
+    });
+
+    expect(capture.manifest.scopes).toContain('project');
+    expect(await countScopeFiles(capture.anchor.root, 'project')).toBe(1);
+    // Dereferenced into an ordinary file, so the capture owns its bytes and stays
+    // readable after the link's target moves.
+    expect(
+      await readFile(
+        join(capture.anchor.root, 'project', '.archon', 'commands', 'review.md'),
+        'utf-8'
+      )
+    ).toBe('review the diff');
+  });
+
   test('leaves no usable capture behind when it cannot finish', async () => {
     const { source, runArtifacts } = await createSandbox();
     await writeFile(join(source, '.archon', 'commands', 'review.md'), 'x');
--- a/packages/workflows/src/workflow-source.ts
+++ b/packages/workflows/src/workflow-source.ts
@@ -45,7 +45,6 @@ import {
   rm,
   rename,
   stat,
-  lstat,
   realpath,
   readFile,
   writeFile,
@@ -293,10 +292,17 @@ function dedupeNestedDirs(dirs: readonly string[]): string[] {
   );
 }
 
-/** True when `path` exists and is a directory. */
+/**
+ * True when `path` exists and is a directory.
+ *
+ * `stat`, not `lstat`: a scope root is routinely a symlink — the dotfiles layout the
+ * global-workflows guide recommends makes `~/.archon/workflows` one — and dropping it
+ * here would silently omit that whole scope from the capture. This matches `copyTree`
+ * below, which dereferences symlinks on purpose.
+ */
 async function isDirectory(path: string): Promise<boolean> {
   try {
-    return (await lstat(path)).isDirectory();
+    return (await stat(path)).isDirectory();
   } catch {
     return false;
   }


Wirasm:
## Ready to merge

The capture root guard now follows directory symlinks, matching the existing copy path and the documented global-workflow dotfiles layout. The clean merge against current `dev` has the same two-file change, all review scopes found no defect, and local validation passed. PR checks show only CodeRabbit; this review supplies the required local evidence.

**Validation:** `bun run validate` passed; focused workflow-source suite passed 38/38.

### Findings

No findings.

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | No additional findings |
| seams | No additional findings |
| simplify | No additional findings |

#### Validation

|---|---|---|
| `git merge-tree d568cc78b8754360a43746dc2ae91e82beb270a2 1e2de92073b9b9a23f8b5603782769afc0cd1fa4` | PASS | Clean virtual merge tree `ceea9dc923048cf6d06679f0d32b4db305cab0cb`; effective merge diff is the intended two files. |
| `bun test packages/workflows/src/workflow-source.test.ts` | PASS | 38 pass, 0 fail. |
| `bun run validate` | PASS | Bundled-generation checks, API types, workspace type checks, lint, formatting, install test, and repository tests completed; durable log: `reviews/pr-3233/round-1/validate.log`. |
| GitHub Actions | NOT RUN | The PR reports only CodeRabbit; no Actions result was available. |

</details>

PR #3235, opened 2026-09-08
feat(sdlc): compose bounded runtime control suites

coleam00:
## Problem and outcome

A successful baseline runtime check does not show whether independent journeys work or whether a deliberate defect is detected. `archon-verify-runtime-suite` runs a finite manifest through the shared runtime verifier and compares every observed verdict with its declared expectation.

## Review guidance

Stacked on #3227. Review the small composition and manifest example first, then the two computation scripts. This adds one workflow and reuses the existing command-backed verifier. It does not add another agent runner or promotion policy.

## Solution

- Expand a bounded manifest into native fan-out instances, with at most two cases active at once.
- Compare exact verdicts and preserve case identities, baseline verification and per-node evidence separately. An expected negative result can satisfy the control without qualifying the candidate for merge.
- Reject unavailable measurements, missing results, invalid manifests and mismatched identities. Resume uses the engine's frozen instance snapshot.
- Include an ordinary Node CLI example. Project-owned environment hosts retain responsibility for resource isolation and cleanup after cancellation or engine failure.

## Validation

- Real executor tests exercise the packaged graph, deterministic scripts and HTTP processes, with the provider simulated. Coverage includes control detection, escaped mutations, bad baselines, identity mismatches, malformed reports, missing returns and cancellation.
- Two native medium-provider suite runs against a Python/SQLite application passed baseline and independent journeys and detected the deliberate inventory mutation. The second run also exercised malformed and negative HTTP body-length requests against the corrected candidate. Each case used fresh resources and recorded actual HTTP evidence; the external host closed afterward.
- A prior live run correctly returned inconclusive after malformed evidence filenames. #3227 now passes the rejection reason to its bounded retry; the strict evidence checker remains unchanged.
- Full local validation limitations inherited from the prerequisite remain explicit: Windows symlink EPERM and mount-prefix tests prevent a fully green root suite. Fixtures and runtime controls do not prove holdout secrecy, agent truthfulness or merge authorization.

After integrating runtime feedback, all 47 selected executor cases passed with 2,136 assertions. Head `52133f25` also includes #3227's test-only CRLF correction.


coleam00:
Closing with #3227. The suite is the second slice of #3263.

PR #3248, opened 2026-09-09
fix(workflows): rename loop_group body boundaries through nested includes

coleam00:
## Problem and outcome

A `loop_group` body node carries its enclosing include's composed-block boundary, and that boundary names outer-graph ids. When the enclosing block is itself included one level further out, `rewriteNodeOutputRefs` renamed the loop node's own boundary but not its body nodes', so a body node kept pointing at an id that no longer existed in the flat graph and was skipped as "upstream failed".

- **Outcome:** a two-level composition (`archon-lifecycle` → `archon-ship` → `archon-deliver`) runs deliver's corrections loop. Before, the loop's first body node (`ci-note`) was skipped with `upstream_failed: gate-direct` (the ship-level id; the flat graph only has `ship__gate-direct`), then `until_bash` could not resolve `$recheck-action.output.action` and the run failed. Each level works on its own; only the two-level composition hits it.
- **Invariant:** single-level includes behave exactly as before. Only boundary ids stamped on `loop_group` body nodes are renamed, with the same rename the loop node itself receives.
- **Scope boundary:** include expansion only. No pack change.
- **Root cause:** `rewriteBoundaries` did not recurse into `loop_group.nodes`, while `markComposedNode` stamps the boundary on every body node.

## Review guidance

- **Feedback requested:** whether renaming inside `rewriteNodeOutputRefs` is the right site, or whether the body boundary should be derived at rename time rather than stamped and rewritten.
- **Start here:** `packages/workflows/src/include-expander.ts`, the `rewriteBoundaries` helper inside `rewriteNodeOutputRefs`.
- **Review order:** the helper, then the regression test `loop_group body boundaries are renamed when the enclosing block is included again`.
- **Lower-attention areas:** none; two files.
- **Known risk or uncertainty:** an unexpanded include directive inside a body carries no composed metadata and is skipped, mirroring the guard `markComposedNode` already applies. The first push missed that and failed type-check on every package; fixed in the second commit.

## Solution

`rewriteBoundaries` walks the node's composed boundaries, then recurses into `loop_group.nodes` for every body that is a real node, applying the same `expandDependency` and `when` rename the outer node gets.

## Validation

- The regression test fails without the fix and passes with it. `bun test src/include-expander.test.ts` in `packages/workflows`: 123 pass.
- `bun run type-check` in `packages/workflows`: clean after the second commit (the first commit failed CI with TS2345 at this site).
- Live: after the fix, three composed `archon-lifecycle` runs on the factory test VPS on 2026-09-09, including one that went through a review-requested correction round and merged (run `dbe7a067`).

## Links

- Found while running the composition that coleam00/ai-software-factory pins (`cleanup/sdlc-workflows-only`). The pin already carries this commit.


coleam00:
Live evidence (2026-09-09, factory test VPS): with this fix pinned, archon-lifecycle -> archon-ship -> archon-deliver ran its corrections loop on the first review finding (run da97c207, PR coleam00/snip-vps#4) where the unfixed composition had failed at `ship__deliver__corrections` (run b6dd3e6c). Two further composed laps completed end to end (merge + deploy) on the same pin lineage.

Wirasm:
## Ready to merge

The recursive boundary rewrite repairs the nested-include failure: body metadata receives the same dependency and predicate rename as the enclosing `loop_group`, while unexpanded directives remain owned by their local expansion pass. The focused regression test, workflow type check, and full repository validation pass. One non-blocking test-strengthening suggestion remains.

**Validation:** `bun run validate` passed; focused include-expander suite passed (123 tests); workflows type check passed.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Extend the new regression with a composed `when:` boundary in a nested loop-group body. | OPEN |

<details>
<summary><code>R1</code> — cover recursive boundary predicate rewriting</summary>

**Impact:** The test proves recursive dependency rebinding, but not the paired `boundary.when` branch at that recursion level. A targeted future edit could preserve the current dependency assertion while leaving a nested loop body’s predicate scoped to an obsolete outer ID, causing a silent skip.

**Evidence:** `packages/workflows/src/include-expander.ts:492` rewrites both `dependsOn` and `when`; `packages/workflows/src/include-expander.test.ts:2418` supplies no include `when:` and asserts only dependencies and trigger rules. Existing nearby coverage proves predicate rewriting for ordinary nodes, not recursive loop-body boundary metadata.

**Required outcome:** Add an outer include predicate and a nested loop-group body, then assert the deepest boundary contains the namespaced predicate and evaluates to `run` with the renamed gate output. This is useful regression protection; the current implementation and supported direct behavior are correct.

**Found by:** `tests`

**Disposition:** OPEN — non-blocking test-strengthening suggestion.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | No findings |
| seams | No findings after exact-head re-review |
| simplify | No findings |
| tests | R1 |

#### Validation

|---|---|---|
| `git diff --check b8b381f3eacc9971444d01eb04629850867c1fd9...230b96b364deb02b5f6107a895a19e35c1288517` | PASS | No whitespace errors |
| `bun test src/include-expander.test.ts` | PASS | 123 pass, 0 fail |
| `bun run type-check` (packages/workflows) | PASS | TypeScript exited 0 |
| `bun run validate` | PASS | Exit 0;  full log at `reviews/pr-3248/round-1/logs/bun-run-validate.log` |
</details>

PR #3230, opened 2026-09-08
feat(sdlc): diagnose regression checks at a pinned revision

coleam00:
## Problem and outcome

A completed check failure needs a diagnosis before it becomes repair work or a tracker item. `archon-regress` inspects an explicit checkout revision once, runs the project's ordinary checks, and investigates a demonstrated code-related failure through the existing SDLC workflows.

- **Outcome:** A typed `healthy`, `regression`, or `inconclusive` result with the original validation evidence and a discovery sidecar for proven defects.
- **Invariant:** The checked revision stays unchanged. Missing checks, unavailable infrastructure and an unproven cause cannot become a healthy or proven-regression result.
- **Scope boundary:** One new composition over `archon-validate` and `archon-investigate`. No repair, tracker publication, scheduler, provider runner or runtime-verification implementation.

## Review guidance

- **Feedback requested:** Producer/consumer contracts, revision binding and diagnosis boundaries.
- **Start here:** `.archon/workflows/sdlc/regress/archon-regress.yaml`.
- **Review order:** Graph, the additive `checks_performed` field in validate, revision/routing/report scripts, then `packages/workflows/src/defaults/regress.test.ts`.
- **Lower-attention areas:** Consumer fixture additions and the regenerated bundled defaults.
- **Known risk or uncertainty:** Validation and investigation judgments remain agent outputs. Revision and evidence-presence checks do not independently prove arbitrary model claims or provide a filesystem sandbox.

## Solution

The graph checks revision and checkout cleanliness before validation, before investigation and after execution. A small `checks_performed` boolean in the existing validate schema distinguishes its compatible green/no-check outcome from checks that actually ran. Only a performed, code-related failure reaches the existing medium-tier investigation command.

The report preserves the producer evidence and emits the native raw discovery-sidecar format. Publication remains a separate governed responsibility. All agent nodes use existing command files; scripts compute identity, routing and output data. Operational producer failures remain failed runs and cannot be inferred into success from absent output.

## Validation

- [GitHub validation](https://github.com/coleam00/Archon/actions/runs/34256403518) passed on Linux and Windows at `f76d6922`, including fixtures, schema upgrade, PostgreSQL parity and Docker build.
- Focused regression, bundle and discovery-report tests: 63 passed, four existing Windows skips, 936 assertions. The regression cases include healthy, no checks, infrastructure, unrooted, reproduced failure, wrong revision, dirty source and missing evidence.
[...]

Issue #3015, opened 2026-08-31
archon-validate should declare which commands it ran, not just the verdict


Wirasm:
## Problem and outcome

`archon-validate` reports whether the gate was green, why it was red, and a prose summary. It never reports **which commands it actually ran**. So nothing downstream can check whether the gate covered the change — the information simply is not in the record.

- **Outcome:** the validate node declares the commands it ran, as structured output, alongside `green` / `red_cause` / `summary`.
- **Invariant:** the discovery contract is unchanged. The node still finds the project's own checks and never invents one; it just writes down what it found and ran.
- **Scope boundary:** pack only. Additive field, no engine change, no YAML surface, no new node behaviour.

## Why this is the blocker

A spike asked whether a green gate can be mechanically verified to have covered the change. Repo-level techniques work well — one found a live gap in this repository on its first run. But every **stack-agnostic** technique needs one thing the record does not contain: what the gate executed.

Without it, the only way to answer "did the gate see the changed files?" is to re-derive the gate from the repository and re-run it, which is expensive and duplicates the discovery the node already did.

With it, several checks become cheap and obvious:

- Did the gate's commands reach the changed paths?
- Did this correction round run the same gate as the previous one, or a narrower one?
- Two runs of the same workflow chose different scopes — why?

That last one is not hypothetical. In a separate experiment, two runs of an identical task on an identical tree picked materially different objects to validate: one scoped to two package-level commands in 96 seconds, the other ran the full repo-wide gate. The record cannot currently distinguish them.

## Sketch

Add to the declared output:

```
commands: a list of what was run, each with the command and its outcome
```

`validate.md` already instructs the node to *"capture each command and its outcome as you go"* and to write them into `$ARTIFACTS_DIR/validation.md`. This promotes that from prose in a report to a declared field a machine can read.

## Acceptance

- The validate node declares the commands it ran and each one's outcome.
- The field is populated on green, red, and unrunnable-gate paths alike.
- `validation.md` keeps its human-readable form; this is additive.
- A downstream node can read the list through the ordinary `$node.output.field` path, with no bespoke parsing.
- No change to how checks are discovered — the node still refuses to invent a command the project does not define.

## Related

Precondition for the review trigger in #3016. Also the missing half of the coverage question in #3009, which is repo-level and does not need this — but nothing pack-level can be built without it.

PR #3261, opened 2026-09-09
feat(sdlc): validate declares whether any ordinary check actually ran

coleam00:
## Problem and outcome

`archon-validate` reports `green`, `red_cause` and a summary. A green verdict with no defined checks is indistinguishable in the record from a green gate that executed the project's checks, so nothing downstream can tell whether the gate saw anything at all. This is the first half of #3015.

- **Outcome:** validate declares `checks_performed`: true only when at least one ordinary project check actually executed. Dependency installation, discovery and failed setup are not checks. False for no defined checks or an unavailable gate.
- **Invariant:** `green` keeps its meaning. No checks is still green; an unavailable gate is still not green. The node still finds the project's own checks and never invents one.
- **Scope boundary:** one additive boolean. The per-command list #3015 sketches is a follow-up; this slice answers "did anything run", which is what a consumer needs first. No engine change, no consumer reads the field yet.

## Review guidance

- **Feedback requested:** whether the boolean is the right first slice, or you would rather land the command list in one go.
- **Start here:** `.archon/workflows/sdlc/validate/archon-validate.yaml`, the output schema; then the "Declare the verdict" section of `validate/commands/validate.md`.
- **Review order:** schema, prompt, the two validate fixtures, then the one-line additions to every fixture that stubs validate's output.
- **Lower-attention areas:** 21 fixture stubs gain `checks_performed: true`; the bundle is regenerated by its owning script.
- **Known risk or uncertainty:** the prompt also gains a paragraph telling validate not to launch a coding agent, another workflow, or a self-judging gate from inside an aggregate script. That came from a live case where a project's `validate` script wrapped an agent-driven harness; it is generic wording, but say if you would rather it be its own change.

## Solution

Add the field to the declared output and make it required. `validate` is also declared `mutates_checkout: false`, which is what it has always promised. The prompt defines the field beside `green` and `red_cause`.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `green`, `red_cause`, `summary` | plus `checks_performed` |
| **Failure behavior** | unchanged | unchanged |

## Validation

- `bun run cli workflow test sdlc`: 47 passed, 3 failed. The three failures are the ship prompt-fragment fixtures (`direct-delivered`, `planned-delivered`, `rooted-delivered`) and reproduce identically on clean `dev` (45 passed, 3 failed); the two new validate fixtures are the only delta.
- `bun run generate:bundled`: regenerated.
- Carved from #3230, where the field was live-proven by a real-provider regress run that returned `healthy / checks_passed` after executing 93 unit tests.

## Links

- Relates to #3015 (first slice)
- Carved from #3230

## Summary by CodeRabbit

- **New Features**
  - Validation results now explicitly report whether ordinary project checks were performed.
  - Validation distinguishes between completed checks, unavailable checks, and scenarios where no ordinary checks are defined.
  - Validation workflows are identified as non-mutating operations.

- **Bug Fixes**
  - Aggregate gates that launch agents, workflows, or self-evaluating checks are no longer treated as ordinary checks.
  - Partially executed validation remains non-green when required checks are still outstanding.
  - Validation outcomes now more accurately reflect the checks actually run.


Wirasm:
Thanks for the contribution. We are declining this slice after checking it against the current validation flow and planned work.

Issue #3015 needs the commands that ran and their outcomes so consumers can assess coverage and compare validation scope. A `checks_performed` boolean cannot answer those questions, has no current production consumer, and can be derived from that command list. We will keep #3015 open for that outcome.

The additional restriction on agent-driven aggregate scripts also changes project gate policy beyond this evidence-reporting change. The new `mutates_checkout: false` declaration needs separate justification because validation can install dependencies.

A replacement can address #3015 directly with structured commands and outcomes, grounded in a concrete consumer, while preserving project-defined gate discovery and avoiding these unrelated policy changes.

Issue #3273, opened 2026-09-10
bug(providers): one unrefreshable Pi credential fails sessions and catalog refreshes for every other provider


Wirasm:
## Problem

An expired Anthropic OAuth credential in Pi's auth storage stops Pi work that has nothing to do with Anthropic. Observed twice on this install, 9 Sep 2026, with Pi 0.84.4:

**1. A Pi session on a different provider fails.** `archon workflow run e2e-pi-smoke --config .archon/config.gemini.yaml`. The session started on OpenAI Codex — `pi.session_started` logged `piProvider: "openai-codex", modelId: "gpt-5.6-sol"` — and then failed:

```
ModelsError: OAuth refresh failed for anthropic: Anthropic token refresh request failed.
url=https://platform.claude.com/v1/oauth/token; status=400;
body={"error": "invalid_grant", "error_description": "Refresh token expired"}
  at resolveStoredOAuth (pi-ai/dist/auth/resolve.js:77)
  at resolveStoredOAuth (pi-coding-agent/dist/core/auth-storage.js:383)
  at ModelsImpl.resolveRefreshCredential
```

The node failed, its downstream node was skipped, and the run failed. No Anthropic model was configured anywhere in that run.

**2. Catalog refresh aborts entirely.** `pi update --models` fails with the same Anthropic error and refreshes nothing, for any provider.

The practical effect is that every Pi-provider config on the install is dead at once — gemini, gemini-glm, minimax, glm-flash — from one unrelated expired credential, and the only recovery is an interactive Anthropic re-login.

## Why this is wrong, not just inconvenient

pi-ai's resolver is per-provider by construction. `resolveProviderAuth(provider, credentials, authContext, overrides)` takes one provider and resolves only that one, and the file's own doc comment states the intent:

> Auth resolution shared by the `Models` and `ImagesModels` collections. A stored credential owns the provider: ambient/env is consulted only when nothing is stored. No silent env fallback after a failed refresh.

So the design already says an expired Anthropic token should make Anthropic unavailable and nothing else. Something upstream of that function is resolving credentials for providers the run does not use, and one hard failure aborts the batch rather than marking that provider unavailable.

This also breaks the failure-isolation expectation an operator holds. Losing a provider you are not using should not cost you the providers you are.

## Why now

[...]

Issue #2481, opened 2026-08-05
feat(workflows): per-run tier/alias rebinding — override the existing resolution chain, not a parallel path


Wirasm:
Supersedes #1433, which is closed as out of shape. Not out of scope — **out of shape**: it proposed a second, parallel resolution path that fires only when a flag is present. What we want is to override the path that already exists.

## The distinction that matters

#1433's pipeline branches:

```
override set?  YES → applies to every AI node, period.
                     Workflow-level AND per-node provider/model are IGNORED.
               NO  → per-node YAML > workflow YAML > config > SDK default
```

Two different resolution behaviours depending on whether a flag was passed. Everything a workflow author expressed gets discarded on the override branch, which is why that design needed a `provider_locked: true` escape hatch to claw back what it had just thrown away.

**This issue proposes no branch.** The run override is one more layer in the resolution chain that already exists, at the top:

```
run override (CLI / HTTP)        ← new, run-scoped, never persisted
  > per-user prefs (DB)
  > repo .archon/config.yaml
  > global ~/.archon/config.yaml
  > built-in tier defaults
```

Node-level resolution is **completely unchanged** — `node.model ?? workflow.model ?? config default`, with tier and `@alias` refs resolving exactly as they do now. What changed is only *what those refs resolve against*. One path, always, flag or no flag.

## Why the flatten-everything shape is wrong

A workflow deliberately uses **different tiers in different nodes** — a small model to detect the stack, a large one to plan, a medium one to fix. That structure is the entire point of tiers.

Applying one model to every node destroys it. A workflow authored to be cheap-where-it-can-be becomes uniformly expensive or uniformly weak, and the thing you were trying to measure — does a cheaper small model still hold up here? — is no longer expressible, because there is no longer a small tier.

## What we actually want

**Rebind the tier→model map for one run.** The workflow is untouched; what `small` and `large` *mean* changes for that invocation only.

```bash
archon workflow run my-workflow \
  --model small=minimax/MiniMax-M3 \
  --model large=openai/gpt-5.6 \
  "<args>"
```

`--model <name>=<spec>`, repeatable. `<name>` is a tier (`small`/`medium`/`large`) or an alias (`@planner`). `<spec>` is anything `resolveModelSpec()` already accepts — a literal model id or another alias. Bare `--model <spec>` with no `=` keeps its current coarse meaning; the `=` is what disambiguates.

```http
POST /api/workflows/{name}/runs
[...]

PR #3276, opened 2026-09-10
fix(cli): validate credential expiry in doctor and run pre-flight

### LINKED ISSUE #3274: fix(cli): archon doctor proves a credential file exists, not that the credential works
## Problem

`archon doctor` checks that a credential file exists, and reports that as working auth. `packages/cli/src/commands/doctor.ts:375`:

```ts
export function probeAuthJsonExists(path: string): boolean {
  return existsSync(path);
}
```

and its only use, at line 392:

```ts
if (probeAuthJsonExists(authJsonPath)) {
  return { label, status: 'pass', message: '~/.pi/agent/auth.json found' };
}
```

On 9-10 Sep 2026 this install had a Pi OAuth grant for `anthropic` that expired **8 June 2026**. Every Pi workflow failed. `archon doctor` reported pass, because the file was on disk.

The same shape appears again at line 559, where the provider-credential report counts how many credentials the CLI user has connected. A count is not a validity check either.

And there is no pre-flight at all. The run that exposed this (`archon-ship`, 9 Sep) created a worktree, started, and failed at its first node with an error naming a provider the run was not configured to use. The credential state was knowable before anything was spent.

## Why this is worth solving

A diagnostic that cannot distinguish working from broken is worse than no diagnostic, because it redirects the investigation away from the cause. Here it did exactly that: the failure was first read as a missing model, which produced a wrong issue body (#3272, since rewritten), a discarded plan to bump `packages/providers/package.json` to `^0.85.1`, and a hand-written `~/.pi/agent/models.json` override duplicating a generated artifact. None of that would have happened if `doctor` had said "anthropic credential expired 8 June".

Three months passed between expiry and detection. Every surface that could have reported it was checking presence instead of validity — `archon doctor`, the credential count, and `pi auth check --no-refresh`, which also answers "ready" for a credential it has not tried to refresh. A credential's location and type are cheap to verify and nearly worthless. Refreshability is the only property that matters.

## Why now

It just cost a failed maintainer run and a wrong diagnosis, and the underlying condition is silent and open-ended: an OAuth grant expires on its own schedule with no warning, and #3273 shows one stale credential can break operations for providers that are perfectly healthy. Anyone carrying a stale grant is one refresh away from the same wall, with `doctor` telling them everything is fine.

## Desired outcome

`archon doctor` reports a provider as failing when its credential cannot be used, naming the provider and why. A run that needs a provider whose credential is unusable fails before it creates a worktree or spends a token, with the same message.

## Invariants

- **Only what Archon already inspects.** `doctor` already reads `~/.pi/agent/auth.json`; validating what it already looks at is in scope. Do not begin scanning other tools' credential stores.
- **Never log or surface a credential value.** Report provider, state, and expiry date only. Existing redaction rules apply unchanged.
- **Unconfigured providers still skip.** `checkPi` deliberately skips when Pi is not the default assistant so Claude-only users are not nagged about shared env keys (line 385). A validity check must not turn a skip into a failure.
- **Unreachable is not expired.** A network failure while probing must report as its own state, not as an invalid credential, and must not fail the whole `doctor` run.
- **The pre-flight checks only what the run needs.** A run configured for OpenRouter must not be blocked by an unrelated provider's credential. That is the defect #3273 is about; do not reproduce it here.

## Acceptance

- With an expired credential in a store `doctor` already reads, `doctor` reports `fail` for that provider, naming it and its expiry date, instead of `pass`.
- A test covers this with a fixture credential expired in the past; it fails against the current `existsSync` implementation.
- Launching a run whose configured provider has an unusable credential fails before worktree creation, with the provider named.
- A provider that is unreachable during the probe reports distinctly from one whose credential is expired, and neither aborts the remaining checks.
- A run configured for provider A is unaffected by an expired credential for provider B.
- Providers that are not configured continue to skip.

## Solution steering

Optional. For Pi, an authoritative probe already exists and does the refresh: `pi auth check --provider <p> --json`, which returns `{"status":"invalid","provider":"anthropic","reason":"invalid_state"}` for the state described above and `{"status":"ready",...}` otherwise. For any store carrying an `expires` field, comparing it to the current time is sufficient and needs no network call — prefer that where available and reserve the probe for stores without one.

Sibling: #3179 — `doctor` reporting "All checks passed" on an install carrying skill roots 0.10.0 removed. Same symptom, different cause; cross-link rather than merge.
Context: #3273 — one unrefreshable credential failing unrelated providers, which is why this went unnoticed for three months.

Wirasm:
## Problem and outcome

`archon doctor` verified auth presence by checking if `~/.pi/agent/auth.json` or connected DB key rows existed on disk, reporting passing auth even when tokens were expired or unusable. Additionally, workflow launch lacked credential pre-flight, allowing runs to allocate worktrees and execute before failing at their first AI node.

- **Outcome:** `archon doctor` reports `fail` with the provider and expiry date when credentials are unusable or expired. Workflow launches perform pre-flight verification on required providers before spawning detached processes or creating worktrees.
- **Invariant:** Scoped strictly to credential stores Archon already inspects (`~/.pi/agent/auth.json` and database keys). Never logs or surfaces credential secret values. Unconfigured providers continue to skip, network reachability failures report distinctly without failing doctor, and pre-flight checks inspect only the providers required by the workflow.
- **Scope boundary:** Does not scan external stores or third-party keychain directories. Does not modify token refresh lifecycles inside provider runners.
- **Root cause:** `probeAuthJsonExists` used `existsSync` on `~/.pi/agent/auth.json` without parsing contents or inspecting `expires` timestamps, and `checkConnectedProviders` only counted registered user provider keys.

## Review guidance

- **Feedback requested:** Correctness of credential expiry inspection and workflow provider pre-flight scoping.
- **Start here:** `packages/cli/src/utils/credential-validity.ts:102` — `inspectPiAuthJson` and `assertWorkflowCredentialsValid` define the core validity checking and launch pre-flight gate.
- **Review order:**
  1. `packages/cli/src/utils/credential-validity.ts`: token parsing, expiration checking, provider extraction, and assertion gate.
  2. `packages/cli/src/commands/doctor.ts`: integration into `checkPi` and `checkConnectedProviders`.
  3. `packages/cli/src/commands/workflow.ts`: pre-flight invocation before detached child spawning and worktree creation.
  4. `packages/cli/src/commands/doctor.test.ts` & `packages/cli/src/commands/workflow.test.ts`: regression and isolation tests.
- **Lower-attention areas:** Re-exports in `packages/providers/src/index.ts` and `packages/providers/src/community/pi/index.ts` exposing `PI_OAUTH_ENV_VARS` and `parsePiModelRef`.

## Solution

1. Added `packages/cli/src/utils/credential-validity.ts` with helper functions:
   - `inspectPiAuthJson`: parses `auth.json`, checks `expires` timestamps without network calls for OAuth tokens, falls back to `pi auth check --provider <p> --json` when timestamps are absent, and detects network reachability issues.
   - `collectWorkflowRequiredProviders`: inspects workflow nodes (including nested loop groups) to extract required provider and model references.
   - `assertWorkflowCredentialsValid`: verifies environment variables, `auth.json`, or connected DB credentials for required providers.
2. Updated `checkPi` in `packages/cli/src/commands/doctor.ts` to inspect credentials via `inspectPiAuthJson`. If an expired token is detected, it fails naming the provider and expiry date. If probing encounters a network issue, it reports `skip` with an unreachable note instead of failing the doctor run.
3. Updated `checkConnectedProviders` in `packages/cli/src/commands/doctor.ts` to verify encrypted OAuth payloads for expiration.
4. Integrated `assertWorkflowCredentialsValid` into `runWorkflowWithOwnedSource` in `packages/cli/src/commands/workflow.ts` before worktree creation and background detachment.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `archon doctor` reported `pass` if `~/.pi/agent/auth.json` was on disk, regardless of token expiration. | `archon doctor` verifies expiration dates and fails with the provider name and formatted expiry date (e.g. `anthropic credential expired 8 June 2026`). |
| **Failure behavior** | Workflow runs with expired credentials proceeded through worktree creation and detached spawn before failing inside agent nodes. | Workflow launch immediately fails during pre-flight before creating worktrees or spawning background processes. |

## Architecture

```mermaid
flowchart LR
  CLI[archon doctor / workflow run] --> Gate[credential-validity]
  Gate --> CheckAuth[inspectPiAuthJson]
  Gate --> CheckDB[checkKeyValidity]
  CheckAuth --> Verdict{Valid / Expired / Unreachable}
  Verdict -->|Expired| Fail[Fail with provider & expiry date]
  Verdict -->|Unreachable| Skip[Skip without aborting doctor]
  Verdict -->|Valid| Pass[Proceed with run / pass check]
```

### Changed seams

|---|---|---|
| `packages/cli/src/commands/doctor.ts` | `checkPi` and `checkConnectedProviders` now inspect token validity and expiry | `packages/cli/src/commands/doctor.test.ts:480` |
| `packages/cli/src/commands/workflow.ts` | Pre-flight validation gate before worktree allocation | `packages/cli/src/commands/workflow.test.ts:1775` |
| `@archon/providers` | Export `PI_OAUTH_ENV_VARS` and `parsePiModelRef` | `packages/providers/src/index.ts:94` |

## Validation

- `bun --cwd packages/cli test src/commands/doctor.test.ts` — passed — proves regression detection for expired credentials and distinct unreachable handling.
- `bun --cwd packages/cli test src/commands/workflow.test.ts` — passed — proves pre-flight blocks before worktree creation and respects provider isolation (#3273).
- `bun run check:cli-import-boundary` — passed — proves no illegal cross-boundary imports.
- `bun run type-check` — passed — 0 errors across all workspaces.
- `bun run lint` — passed — 0 warnings/errors across all workspaces.
- `bun run format:check` — passed — all files formatted.
- `bun run validate` — passed — aggregate validation passed clean.
- **Not verified:** Nothing material; test suite covers fixture expiration, presence, unneeded provider decoupling, and unreachable probes.

## Delivery considerations

|---|---|---|
| Security / permissions / data | Credential tokens are never logged or echoed; only provider names, validity status, and formatted expiry dates are output. | `packages/cli/src/utils/credential-validity.ts:21` |
| Compatibility / migration | Fully backward compatible; unconfigured providers and alternative assistants like Claude continue to skip without error. | `packages/cli/src/commands/doctor.test.ts:468` |

## Links

- Closes #3274
- Related #3273

## Summary by CodeRabbit

- **New Features**
  - Added credential preflight checks before workflow runs, detecting missing, expired, invalid, or unverifiable credentials before execution begins.
  - Added support for validating credentials required by workflow AI nodes, including provider-specific and runner-managed credentials.
  - Expanded provider and workflow model resolution for more accurate execution checks.

- **Bug Fixes**
  - Improved the `doctor` command to inspect credential contents and clearly distinguish expired, invalid, unreadable, and unreachable credentials.
  - Added clearer reconnect guidance while keeping sensitive credential data out of diagnostic results.


Wirasm:
## Needs fixes

The pre-flight gate this PR adds is a real hard-fail path when it fires, but it too often doesn't fire. Two independent, proven gaps compound: `collectWorkflowRequiredProviders` never recognizes the dominant real-world way of writing an AI node (`command:`-sourced agent nodes, used in 31 of the bundled `.archon/workflows/**/*.yaml` files, plus a gate's rework reprompt) because it duck-types against a top-level `prompt`/`loop` field that doesn't exist on the parsed `DagNode` shape — the prompt lives under `node.source.prompt`. And even for providers it does detect, `assertWorkflowCredentialsValid` has no "I couldn't determine this" branch: a provider with no credential anywhere never throws, and any DB lookup/decrypt failure other than one whose message happens to contain "expired" is silently swallowed. `doctor.ts`'s new `checkConnectedProviders` code has the same shape of bug in a separate function. All three sites reimplement logic that already exists correctly elsewhere in the repo (`resolveNodeModel`/`collectContainerIncompatibleProviders` for node traversal, `resolveOAuthCredential` for decrypt-and-check-expiry with a logged guard against malformed rows) — reuse would have prevented the gaps rather than requiring them to be caught in review.

No path in the diff logs or surfaces a raw credential value — verified directly, not just absent evidence (see Validation). Expiry unit handling (seconds vs. ms) and the exactly-now boundary are correct. The new `packages/providers/src/index.ts` export is a clean re-export of an already-internal record and function; it does not leak provider-SDK shape into the CLI.

**4 blocking · 3 non-blocking**

**Validation:** `bun --filter '*' type-check` PASS (13/13 packages) · `bun run lint` PASS · `packages/cli` `bun run test` PASS (0 fail) · `packages/providers` `bun run test` PASS (0 fail) · CI re-running on the frozen head, not awaited per operator instruction · Windows leg previously red on an unrelated `@archon/workflows` test, confirmed stale-base by the operator and not re-verified here per instruction.

### Findings

|---|---|---|---|
| `R1` | Critical | Pre-flight gate never checks command-sourced or gate-rework AI nodes — silent no-op on the dominant real-workflow style | OPEN |
| `R2` | Critical | Credential validity check fails open whenever it cannot positively confirm invalidity | OPEN |
| `R3` | Important | `doctor`'s `checkConnectedProviders` drops the `unreachable` status and reports "pass" for an unverifiable credential | OPEN |
| `R5` | Important | Tier/alias-authored nodes resolve through the wrong fallback because `config` is never threaded to the gate | OPEN |
| `R6` | Suggestion | New 410-line `credential-validity.ts` has no dedicated test file; several branches are untested, which is why R1/R2/R5 shipped | OPEN |
| `R7` | Suggestion | No test asserts a credential value can never appear in a doctor message or thrown error | OPEN |
| `R8` | Suggestion | `probePiCredential` treats a missing `pi` binary as "ready" and classifies vendor errors by substring match | OPEN |

<details>
<summary><code>R1</code> — Pre-flight gate never checks command-sourced or gate-rework AI nodes</summary>

**Impact:** For any workflow whose AI work is authored with `command:` (the standard "reference a reusable prompt file" form) or lives inside a gate's `on_reject`/rework reprompt, `collectWorkflowRequiredProviders` returns an empty provider set. `assertWorkflowCredentialsValid` then hits its early return and never checks anything — the run proceeds to worktree creation and AI execution exactly as if this PR did not exist, while looking like the gate ran. This is not a corner case: 31 of the bundled `.archon/workflows/**/*.yaml` files use `command:` nodes (e.g. `archon-assist.yaml`, a default workflow, is a single `command:` node), against 44 using `prompt:`.

**Evidence:** `packages/cli/src/utils/credential-validity.ts:246-253` —
```ts
const isAi =
  node.kind === 'loop' ||
  'prompt' in node ||
  'loop' in node ||
  (node.kind === 'agent' &&
    (typeof node.source !== 'object' ||
      (node as { source?: { kind?: string } }).source?.kind === 'inline'));
```
The real, parsed `AgentNode` always carries its prompt under a discriminated `source` — confirmed directly: `packages/workflows/src/schemas/dag-node.ts:1792-1793` (`source: { kind: 'command', ... }`) vs. `:1805` (`source: { kind: 'inline', prompt: data.prompt.trim() }`). There is no top-level `node.prompt` or `node.loop` on a loaded `DagNode`, so those two disjuncts never fire; only `source.kind === 'inline'` is live. A command-sourced agent node dispatches through the identical `executeAgentNode` as an inline one but is invisible here.

A near-identical, typed function that answers this exact question already exists: `collectContainerIncompatibleProviders` (`packages/workflows/src/dag-executor.ts:11068`, itself noted at `:1777` as hand-mirroring `resolveNodeModel`'s resolution). This PR reimplemented traversal with an untyped `nodes?: readonly unknown[]` and `'x' in node` duck-typing instead of reusing it, and the two have already diverged.

**Class:** Invariant — every `DagNode` variant that invokes a provider must be captured. Enumeration ran against the full 9-member kind union (`agent, exec, gate, halt, wait, loop, loop_group, workflow, compose_fan_out`, `dag-node.ts:966-974`), cross-checked against each schema shape, then executed live:
- Affected, proven by direct execution: `agent` nodes with `source.kind: 'command'` (repro: command-sourced node → `required providers: []`; inline equivalent with the same provider → `required providers: ["openrouter"]`); `gate` nodes with a `decisions[].rework` reprompt (repro: gate with `provider:'openrouter'` + a rework decision → `required providers: []`).
- Examined and found clean: `exec`, `gate` without rework, `halt`, `wait`, `loop` (provider/model are live top-level fields, correctly read), `workflow`/child-run nodes, `compose_fan_out`.
- Not scored (pre-existing, not introduced by this PR): `loop_group`'s own `provider`/`model` forward to un-overridden body nodes per `dag-node.ts:1038-1039`, but this function's recursion falls back only to top-level `workflow.provider` — `collectContainerIncompatibleProviders` appears to share this gap.

**Required outcome:** Replace `isAi`'s duck-typing and the hand-rolled traversal with the repo's typed `DagNode` union and the existing `resolveNodeModel`/`collectContainerIncompatibleProviders` machinery (or a shared helper built on it), so a node's AI-ness and required provider are derived the same way execution already derives them, rather than a second implementation that can silently drift.

**Found by:** `prp-core:seam-analyzer`, `prp-core:code-simplifier`, `prp-core:pr-test-analyzer`; corroborated at lower confidence by `prp-core:code-reviewer`.

**Disposition:** OPEN.

</details>

<details>
<summary><code>R2</code> — Credential validity check fails open whenever it cannot positively confirm invalidity</summary>

**Impact:** `assertWorkflowCredentialsValid`'s per-provider loop (`packages/cli/src/utils/credential-validity.ts:330-409`) only `continue`s or `throw`s inside explicit positive branches (env var present; a `piAuth` entry with status `expired`/`invalid`/`valid`; a DB row found with a computable expiry). There is no default branch for "I don't know." Two concrete, reachable consequences, one proven by direct execution against the real module:

1. A required provider with **no credential anywhere** — no env var, no `auth.json` entry, no DB row — falls through every check without throwing, and the function returns normally. Reproduced by calling `assertWorkflowCredentialsValid` directly with a workflow requiring `claude`, zero env vars, a nonexistent `authJsonPath`, and no CLI identity: it did not throw. This is the single most common "unusable credential" case (a provider that was simply never configured), and it is exactly the case the PR's own inline comment at `workflow.ts:2081-2085` says should hard-fail.
2. The DB-credential branch's catch (`credential-validity.ts:403-407`) only rethrows when the caught error's message contains the substring `"expired"`; every other failure — `getEncryptionKey()` throwing on a missing/rotated `TOKEN_ENCRYPTION_KEY`, or `decryptToken`/`JSON.parse` throwing on a corrupted row — is silently swallowed, and the loop moves to the next provider as if this one were fine. Reproduced with a real CLI identity and no matching DB row: same silent pass. Additionally, a DB row that decrypts cleanly but has a missing or non-numeric `expires` field skips the inner `if` entirely and falls straight to `continue` — treated as valid without any comparison ever happening.

Both new sites (`credential-validity.ts:376-408` here, and `doctor.ts:724-753`, see R3) reimplement "decrypt `oauth_creds_encrypted` → parse `expires` → compare to now" instead of calling the repo's one existing, correct implementation of this exact check: `packages/core/src/db/user-provider-key-store.ts` (`resolveOAuthCredential`, ~line 216-227), whose own comment explains precisely the guard both new copies lack: *"A missing or non-numeric value from a legacy/corrupt row would make that comparison silently false and serve a stale token as success, so enforce the shape here ... and treat a mismatch like decrypt failure."* That function also logs (`user_provider_key.oauth_decrypt_failed`, `user_provider_key.oauth_malformed_expires`) — the two new copies log nothing on any of these paths.

**Evidence:** `packages/cli/src/utils/credential-validity.ts:330-409` (loop with no default branch), `:403-407` (message-substring-gated rethrow), `packages/core/src/db/user-provider-key-store.ts:216-227` (the canonical guarded, logged implementation this PR bypassed instead of calling).

**Required outcome:** The per-provider loop needs an explicit "could not determine validity" outcome that is treated as invalid (throws, naming the provider and reason) rather than falls through as valid — for no-credential-found, for any DB/decrypt exception, and for a malformed/missing `expires`. The cleanest fix reuses `getDecryptedProviderCredential`/`resolveOAuthCredential` from `user-provider-key-store.ts` instead of re-deriving the decrypt-and-compare logic a second time.

**Found by:** `prp-core:seam-analyzer` (proved by direct execution), `prp-core:code-reviewer`, `prp-core:code-simplifier`, `prp-core:pr-test-analyzer`.

**Disposition:** OPEN.

</details>

<details>
<summary><code>R3</code> — <code>doctor</code>'s <code>checkConnectedProviders</code> drops the <code>unreachable</code> status and reports "pass"</summary>

**Impact:** `checkConnectedProviders`'s new validity loop (`packages/cli/src/commands/doctor.ts:682-693`) only special-cases `validity.status === 'expired'`. The new `defaultLoadProviderDeps().checkKeyValidity` (`doctor.ts:724-753`) explicitly returns `{ status: 'unreachable', reason }` when decrypt or DB lookup throws (`:753`) — but that status is never checked by the caller; it isn't pushed into `expired`, isn't mentioned in the final message, and the function falls through to the unconditional `return { label, status: 'pass', message: '${rows.length} connected: ...' }` at `:706`. `archon doctor` reports a corrupted or unverifiable credential row as plainly "connected" — reproducing, inside the feature meant to fix it, the exact defect issue #3274 was filed over. This is scoped to `checkConnectedProviders` only: `checkPi`'s own consumption of the parallel 4-member `PiCredentialStatus` union (`doctor.ts:427-472`) is exhaustive and does handle `unreachable` correctly (`skip`, named provider).

**Evidence:** `doctor.ts:682-693` (loop only branches on `'expired'`), `:706` (unconditional pass), `:753` (`unreachable` produced but never consumed). Only one test exists for this function's new wiring (`doctor.test.ts` ~line 1096, `status: 'expired'`); no test covers `status: 'unreachable'` here, unlike `checkPi`'s equivalent which is tested (`doctor.test.ts:526`).

**Required outcome:** Branch on `unreachable` the same way `checkPi` does — report it distinctly (e.g. `skip`) rather than folding it into "connected." As with R2, this and the DB-credential branch in R2 are two fresh copies of the same decrypt-and-compare logic; a shared helper (ideally calling the canonical `resolveOAuthCredential`) fixes both at once.

**Found by:** `prp-core:code-reviewer`, `prp-core:pr-test-analyzer`.

**Disposition:** OPEN.

</details>

<details>
<summary><code>R5</code> — Tier/alias-authored nodes resolve through the wrong fallback because <code>config</code> is never threaded to the gate</summary>

**Impact:** `assertWorkflowCredentialsValid`'s only call site is `packages/cli/src/commands/workflow.ts:2086`: `await assertWorkflowCredentialsValid(workflow, { cwd });`. The function reads `options.config` (`credential-validity.ts:317`) but never reads `options.cwd` anywhere in its body — `cwd` is dead, and `config` is always `undefined` in the real (non-test) path. `collectWorkflowRequiredProviders`'s tier/alias resolution (`credential-validity.ts:259-268`, resolving a node's `model: 'small'|'medium'|'large'` or a named alias via `config.tiers`/`config.aliases`) therefore never fires. Any node authored with a tier or alias instead of a literal `provider`/`model` string falls through to `parsePiModelRef('small')` (fails to parse) → `readPiSettingsDefaultProvider()` → a bare `'pi'` bucket. Net effect: the gate can check the wrong provider's credential entirely — able to both block a healthy run over an unrelated provider's expiry, and let an actually-expired credential for the real provider through unchecked. Other call sites in the same file already do `await loadConfig(cwd)` for exactly this purpose (`workflow.ts:1874`, `:2928`); this call site does not.

**Evidence:** `workflow.ts:2086`, `credential-validity.ts:305-321` (`cwd` declared and passed, never read; `config` read but never supplied). The new tests in `workflow.test.ts` only use literal `provider`/`model` strings, so this gap is untested.

**Required outcome:** `await loadConfig(cwd)` before the gate call and pass `{ config }` through, matching the pattern already used elsewhere in this file. Drop the unused `cwd` option or wire it in if a future use is intended.

**Found by:** `prp-core:code-reviewer`; corroborated by `prp-core:code-simplifier` (flagged the dead `cwd` option independently).

**Disposition:** OPEN.

</details>

<details>
<summary><code>R6</code> — New 410-line module has no dedicated test file</summary>

**Impact:** `credential-validity.ts` is only exercised indirectly through `doctor.test.ts`/`workflow.test.ts`, both of which mock or substitute the interesting internals. Notably untested: `parseExpires`'s epoch-seconds→ms conversion and ISO-date-string branches (every fixture uses a plain ms number); the real `probePiCredential` shell-out and its stdout parsing / ENOENT / network-error classification (every test substitutes a mocked `probeFn`); `collectWorkflowRequiredProviders`'s tier/alias resolution and `readPiSettingsDefaultProvider()` fallback; `inspectPiAuthJson`'s malformed-JSON, non-object-JSON, and unrecognized-credential-type paths; and the "no credential found anywhere" and DB-exception paths behind R2. This is why R1, R2, and R5 shipped without a failing test catching them.

**Evidence:** `find packages/cli/src/utils -iname 'credential-validity*'` returns only the implementation file, no `.test.ts`.

**Found by:** `prp-core:pr-test-analyzer`, `prp-core:seam-analyzer` (process note), `prp-core:code-reviewer` (corroborating: "shipped unverified").

**Disposition:** OPEN.

</details>

<details>
<summary><code>R7</code> — No test asserts a credential value can never appear in output</summary>

**Impact:** Every message-construction site in `credential-validity.ts` and `doctor.ts` was traced by hand this round and none currently interpolates a raw token/key value — this is not a live bug. But it's an invariant with no regression protection: a future "helpful" error message (e.g. echoing a `pi auth check` stderr snippet, or a raw `JSON.parse` failure on auth.json content) could reintroduce a leak with nothing to catch it.

**Found by:** `prp-core:pr-test-analyzer`.

**Required outcome:** A negative-assertion test (e.g. feeding a fixture with recognizable secret values through `checkPi`, `checkConnectedProviders`, and `assertWorkflowCredentialsValid`'s failure paths, then asserting the value never appears in the result) is optional but valuable given how central this invariant is to the PR's own stated purpose.

**Disposition:** OPEN.

</details>

<details>
<summary><code>R8</code> — <code>probePiCredential</code> treats a missing <code>pi</code> binary as "ready"; classifies vendor errors by substring</summary>

**Impact:** `probePiCredential` (`credential-validity.ts:68-100`) returns `'ready'` (i.e. valid) when `execFileAsync('pi', ...)` fails with `ENOENT` — the binary isn't found — rather than an unknown/unreachable state; and separately classifies `unreachable` via free-text substrings (`msg.includes('network')`, `msg.includes('fetch failed')`) alongside legitimate errno tokens (`ENOENT`/`ETIMEDOUT`/`ENOTFOUND`/`ECONNREFUSED`). Blast radius is limited: this function is only reachable via `doctor.ts checkPi`'s optional `probeFn` — the live `workflow.ts` pre-flight gate calls `inspectPiAuthJson` without a `probeFn` (`credential-validity.ts:328`) — so this degrades doctor's diagnostic accuracy only, not a run-blocking path.

**Evidence:** `credential-validity.ts:68-100`.

**Found by:** `prp-core:code-reviewer`.

**Disposition:** OPEN.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 (corroborating), R2, R3, R5, R6 (corroborating), R8 |
| `seams` | R1, R2, R6 (process note) |
| `simplify` | R1, R2, R5 (corroborating) |
| `tests` | R1, R2, R3, R6, R7 |

#### Validation

|---|---|---|
| `bun --filter '*' type-check` | PASS | 13/13 packages exit 0 |
| `bun run lint` | PASS | all lint targets ran, no errors; test-cleanup drift check passed |
| `cd packages/cli && bun run test` | PASS | 0 fail across all suites, exit code 0 |
| `cd packages/providers && bun run test` | PASS | 0 fail across all suites, exit code 0 |
| GitHub Actions CI | NOT RUN | re-running on the frozen head per operator instruction; not awaited or rerun |
| Windows leg (`@archon/workflows` terminal-artifact test) | NOT RE-RUN | operator confirmed stale-base cause and 10/10 pass on the merged tree before this review; not touched by this PR's files, not re-verified here per instruction |

</details>

Wirasm:
## Correction round — all 7 findings dispositioned

Head `6e7dab976`. Still a draft. Base merged `origin/dev` once at `a646761d4`; not chased since.

Every fix below was watched failing against the reviewed head `2a40ed0ec` before it was kept — by importing that commit's `credential-validity.ts` / `doctor.ts` side by side with the corrected ones and running the same assertions against both. Those scratch files are not committed.

| ID | Severity | Disposition |
|---|---|---|
| `R1` | Critical | FIXED |
| `R2` | Critical | FIXED (one sub-case declined, below) |
| `R3` | Important | FIXED |
| `R5` | Important | FIXED |
| `R6` | Suggestion | FIXED |
| `R7` | Suggestion | FIXED — and it turned up a live leak |
| `R8` | Suggestion | FIXED |

### R1 — command-sourced and gate-rework AI nodes

`collectNodeModelBindings` in `packages/workflows/src/node-model-resolution.ts` now owns node traversal and provider/model resolution, built on the executor's own `resolveNodeModel`. `collectContainerIncompatibleProviders` consumes it too, so `resolveNodeProviderForPreflight` — the hand-mirrored copy the review flagged as already drifted — is deleted rather than duplicated a third time. A `loop_group` body now resolves against the group's own provider, which is what the group forwards; resolving it against the workflow named a provider no body node runs on.

**Red at `2a40ed0ec`:**

```
R1 breadth: 22 of 30 bundled AI workflows uncovered
(fail) collects the provider of a command-sourced agent node
(fail) collects the provider of a gate with a rework reprompt
(fail) blocks a command-sourced node whose credential expired
```

Green now, with the breadth case kept as a committed test that walks `BUNDLED_WORKFLOWS` and asserts no bundled workflow with an AI node yields an empty credential set.

### R2 — fail closed

Every required credential resolves to exactly one verdict — `usable`, `unusable`, `unverifiable`, `absent` — and the loop has no fall-through. `unusable` and `unverifiable` both throw, naming the vendor and the reason.

**Red at `2a40ed0ec`**, by direct execution:

```
HEAD: RESOLVED — the gate passed          # DB/decrypt failure, message without "expired"
CORRECTED: THREW — no openrouter credential found for this run. …
(fail) blocks when no store holds a credential at all
(fail) blocks when auth.json exists but cannot be parsed
```

Reuse: yes, `user-provider-key-store.ts`. Not `getDecryptedProviderCredential` though — it mints and refreshes a live bearer, which turns a pre-flight into a vendor round-trip and collapses "the network is down" into "this credential is dead", and #3274 requires those to report differently. Instead the shape guard `resolveOAuthCredential` documents is extracted into `decryptOAuthCredentials`, and a new `inspectStoredProviderCredential` reads it. One owner for the decrypt-and-`expires` guard, no network, and both new call sites (the gate and `doctor`) go through it. The malformed/missing-`expires` case the copies dropped now returns `undetermined` and logs, and it has a test.

**Declined sub-case: "no credential found anywhere → throw", for a runner that is not `pi`.**

Taken literally this hard-blocks every Claude Code and Codex subscription install. Those runners authenticate from their own credential stores, which #3274 explicitly puts out of scope ("Do not begin scanning other tools' credential stores"), so Archon finding nothing is the normal healthy state, not a verdict. The gate therefore splits on who will present the credential:

- `pi` — absence **is** the verdict and throws. Pi authenticates from `auth.json`, the vendor env var, or a credential connected to Archon, and nowhere else.
- any other runner — absence is allowed, logged at `cli.credential_preflight_unverified` with runner and vendor. Everything Archon *can* see still blocks: an expired connected credential for a `claude` node throws, and there is a test for it.

Ambient Pi vendors (`amazon-bedrock`, `google-vertex`) are exempt from the `pi` rule; their credentials come from a cloud chain Archon cannot read.

Two related scoping decisions in the same spirit:

- `auth.json` is consulted for `pi` nodes only. A corrupt Pi store is not a reason to refuse a Claude Code run, and an entry there is not the credential a non-Pi runner presents.
- A failure resolving the **CLI identity** (database down, unwritable store) reports "no connected credential" and logs the real cause rather than translating a database outage into "your key could not be verified". The run creates its own row seconds later and fails there naming the database. Per `AGENTS.md`, translate errors at the boundary that can explain them — this one cannot.

### R3 — `doctor`'s `checkConnectedProviders`

Rewritten onto `inspectStoredProviderCredential`, deleting the second decrypt-and-compare copy. Expired → `fail` with the date; unverifiable → `skip` naming the provider and reason; neither is folded into "connected".

**Red at `2a40ed0ec`**, executed:

```
HEAD reported: {"status":"pass","message":"1 connected: anthropic(oauth)"}
(fail) does not report a credential it could not verify as connected
```

### R5 — `config` never reached the gate

`resolveCliModelContext(cwd, runConfig)` is now the single resolver for both pre-execution readers of a workflow's provider — the `--dry-run` report and the gate — layered exactly as `executeWorkflow` layers them. The dead `cwd` option is gone.

**Red at `2a40ed0ec`:** with the call site's real `{ cwd }` argument, a node authored `model: small` resolving through a tier to `pi`/`openrouter` did not throw on an expired `openrouter` credential — the gate was checking a different vendor.

Two follow-on decisions:

- `--model` bindings are applied to the pre-flight profile when they resolve and skipped when they do not. `executeWorkflow` owns validating those flags and the run row it has to fail; re-raising here turned a mistyped flag into a credential error, which broke `workflowRunCommand — sparse model bindings (#2481)`.
- The title-generation path now reuses the config the gate already resolved instead of loading it again. Net config loads per launch are unchanged. Side effect worth naming: a malformed `.archon` config now fails the launch at the gate rather than only degrading title generation. It already failed the run moments later inside `executeWorkflow`, so this is the same failure, earlier and clearer.

### R6 — dedicated tests

`packages/cli/src/utils/credential-validity.test.ts`, 25 tests, registered in `packages/cli/package.json` as its own `bun test` invocation. Covers the branches the review listed as untested: `parseExpires` seconds/ms/ISO, malformed and non-object `auth.json`, the exactly-now expiry boundary, an empty API key, tier resolution with and without a profile, vendor-canonical mapping, and every verdict the gate can reach. Core's `inspectStoredProviderCredential` has 7 more in `user-provider-key-store.test.ts`.

### R7 — no credential value in output

Kept as an invariant with a test, and the test found a real leak. A malformed `auth.json` reported `(err as Error).message`, and both engines quote the file:

```
NODE: Unexpected token 's', "sk-ant-oat"... is not valid JSON
BUN:  JSON Parse error: Unexpected identifier "sk"
```

`doctor` renders that string directly into `failed to read ~/.pi/agent/auth.json: …`. The message is now a fixed `not valid JSON`. Two negative-assertion tests feed recognizable secrets through every gate failure path and through `inspectPiAuthJson`, asserting the value never appears — including a 10-character prefix, since both parsers truncate rather than omit.

### R8 — `probePiCredential`

A missing `pi` binary is `unreachable`, not `ready`. Classification now comes from the structured channel only: Pi's `--json` payload, read from stdout even on a non-zero exit. The free-text `network` / `fetch failed` matching is gone; anything the probe cannot place is an honest `unreachable` rather than a guess at `invalid`. Per `AGENTS.md`, a vendor rewording must not silently flip a verdict.

### Untouched, as the review asked

Expiry units and the exactly-now boundary, the propagation mechanism (a real throw still hard-blocks before worktree creation, uncaught), and the `packages/providers/src/index.ts` re-export are unchanged.

### Validation

`bun run validate` — pass. `bun --filter '*' type-check` — 13/13. `bun run lint` — clean. `packages/cli`, `packages/workflows`, `packages/core` `bun run test` — 0 fail. `bun run check:cli-import-boundary` — pass.

Test-side changes needed by the gate now actually running before the fork, each an under-specified mock made to match reality:

- `workflow.test.ts`'s `loadConfig` mocks gained `assistant`/`assistants`; no real `loadConfig` returns a config without them.
- The `@archon/core` mock gained `inspectStoredProviderCredential`; unmocked it opened the real credential database.
- `finishStartupWindow`'s microtask budget was raised from 20. The pre-fork gates await dynamic imports, which cost more than 20 turns; the loop only has to be finite, and a spawn that never happens still fails on the assertion after it.

PR #3258, opened 2026-09-09
fix(workflows): await refused child source disposal

Wirasm:
## Problem and outcome

A refused child workflow could return while its captured source was still being removed. Twelve refusal paths shared this detached cleanup; a later failure path could also remove the same source twice. Child execution now owns cleanup through completion, while preserving the original refusal and existing cleanup warnings.

## Review guidance

- Start at `runChildWorkflow` in `packages/workflows/src/executor.ts`, then the gated unknown-target cases in `subrun.test.ts`.
- Most executor diff lines are indentation from extending the existing capture owner over the full child lifecycle.
- This fixes a proved cleanup ownership defect. Its contribution to the wider Windows stalls remains unproved, and the GAP2180 case already used awaited cleanup.

## Solution

Hold the prepared child source in the existing `withCapturedSource` owner immediately after capture. That owner now covers discovery, refusals, setup, recursive execution, and failure returns. Remove detached cleanup from `failOutcome` and the narrower execution-only wrapper. Adoption and terminal-write failure propagation retain their existing contracts.

## Validation

- `bun run validate` passed.
- The subrun suite passed: 101 tests, 502 assertions.
- The new controlled removal regression fails against unmodified dev: execution returns while the staged root still exists. The fixed code waits for removal; an injected removal failure preserves the unknown-target error and emits the existing warning.
- An independent controlled probe of all 100 pre-existing subrun cases found 12 refusal paths returning with live cleanup before the fix, plus one redundant removal. Afterward it found zero pending or redundant removals, including the current terminal-write failure cases.
- Full Windows CI is pending. Passing local ownership proofs do not establish the cause of every Windows failure.

## Links

- Relates to #2924 and #2306.
- [Published implementation plan](https://github.com/coleam00/Archon/issues/2924#issuecomment-5602116749)
- [Causal evidence and counterpart scope](https://github.com/coleam00/Archon/issues/2924#issuecomment-5602020164)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow cleanup after child-run failures so staged resources are reclaimed reliably.
  * Prevented workflow execution from completing before required cleanup finishes.
  * Added clearer handling and logging when staged-source cleanup encounters an error.
  * Preserved parent workflow failure reporting when a child workflow cannot be found.
  * Ensured failed child workflows do not leave behind incomplete staged resources or child-run records.

PR #3277, opened 2026-09-10
perf(workflows): a child spawn stops re-reading the bundled defaults it just copied

Wirasm:
## Problem and outcome

`captureWorkflowSource` copied the whole bundled defaults scope into every capture and then re-read every copied byte to hash it, and `runChildWorkflow` takes a capture per child spawn. On this checkout that is 200 files read twice per spawn, none of them the workflow under test, and the tree keeps growing (`.archon/workflows` went from 49 files on 1 Jul to 154 today). Five test files already carry a `mock.module('@archon/paths', ...)` workaround whose only purpose is to lie to the application about where its bundled defaults live so the suite stays under Bun's budget; every suite without that workaround pays the cost on every Windows leg, and so does every production fan-out.

- **Outcome:** the bundled scope is read and hashed once per process instead of once per capture. Per child spawn, measured with the real `@archon/paths` getters: **836 filesystem operations to 591 (-29%), 1,247,567 bytes read to 60, median wall clock 51.4ms to 33.7ms (-34%)**. Files written, manifest counts, and the capture digest are byte-identical.
- **Invariant:** the bundled bytes are still written into every run's own capture. Only the reading and hashing of them moved; nothing about whether they are there changed.
- **Scope boundary:** late resolution stays inside `runChildWorkflow`; each child still captures from the parent's live authoring root; no `mock.module` workaround was removed; the `discoverWorkflowsWithConfig` memo section 5 suggests "secondarily" is deliberately not implemented (see Delivery considerations).

## Review guidance

- **Feedback requested:** correctness of the freshness key, and whether the honest ~29% is the right stopping point given the alternative measured below.
- **Start here:** `packages/workflows/src/workflow-source.ts:587` — `resolveBundledScope` is the whole decision: what is cached, what invalidates it, and how each build branch differs.
- **Review order:**
  1. `workflow-source.ts:453` — `BundledScope`, and the docblock arguing why this is not the content-addressed cache `script-discovery.ts` publishes.
  2. `workflow-source.ts:574` — `stampListings`, the freshness key, and what it can and cannot see.
  3. `workflow-source.ts:640` — `foldDigest`, split out of `digestTree` so a capture can be digested without re-reading bytes the process still holds. This is a persisted contract; see the compatibility row.
  4. `workflow-source.ts:759` — the call site inside `captureWorkflowSource`.
  5. `subrun.test.ts:51` and `subrun.test.ts:4335` — the two teardown fixes.
- **Lower-attention areas:** `copyTree` becoming `listTree` + `copyListing` is a mechanical split. It preserves the symlink-dereference, cycle-guard, and skip-directory rules verbatim, and the operation counts for the project and global scopes are unchanged.
- **Known risk or uncertainty:** the freshness key is tree shape plus per-file size and mtime, so an edit that preserves both is invisible to it. Bun reports `mtimeMs` with sub-microsecond precision on APFS and NTFS carries 100ns, so the window is not reachable by an editor; re-reading every byte is the only stronger check and is the cost this removes.

## Solution

The two on-disk bundled roots are read once into a `BundledScope` — the file bytes plus each file's sha256, keyed by an identity of the trees they came from — and written into each capture from there. A compiled binary already worked this way for its embedded constants, so both branches now share one path: read into a `BundledScope`, then `writeBundledScope` per capture.

**The premise this PR was asked to verify does not hold as stated.** The analysis proposed materialising once per process because "nothing writes to `getSourceRepoRoot()/.archon` while the process runs". In a source build that is false: `PUT /api/workflows/:name` (`packages/server/src/routes/api.ts:4453-4464`) and `DELETE` (`:4505-4530`) write and unlink under `join(workingDir, getWorkflowFolderSearchPaths()[0])`, i.e. `<cwd>/.archon/workflows`. For a maintainer running Archon from source against the Archon checkout, that cwd *is* the checkout supplying the bundled defaults, and only *packaged* workflows are refused there — a bare `.archon/workflows/foo.yaml` can be created, overwritten, and deleted by the running server. A process-lifetime memo would serve a stale workflow catalogue out of a long-lived `archon serve`, which is a wrong answer rather than a slow one.

So the cache is revalidated before every reuse:

- **Source build:** walk both roots and fold tree shape plus each file's size and mtime into a stamp. Same stamp reuses the held bytes; any difference re-reads. This costs the walk the copy already performed, so it replaces work rather than adding it.
- **Binary build:** keyed on `BUNDLED_VERSION`. Embedded constants cannot move under the process, so there is nothing to recheck.
- **Fail direction:** every path that cannot confirm the cache is current re-reads. There is no branch that serves held bytes on doubt.

This deliberately is not the content-addressed on-disk cache #3252 added in `script-discovery.ts:211-257`. That cache is right for its job because its input is already in memory (hashing it is free) and because a script node executes from the cached path, so the path must be stable and shared. Here the bytes must land inside each capture regardless, so a shared directory would be an extra hop rather than the destination — and on a source build a content key cannot be computed without reading the whole tree, which is precisely the cost being removed. What this borrows from it is the principle: key on an identity that changes when the bytes do.

**Why the bytes are still written rather than hard-linked.** Hard-linking from a shared copy is the only way to get the *writes* off the per-capture path too, and it is what a Windows runner would benefit from most. It was built and measured, and rejected: on macOS APFS a hard link is roughly 3x more expensive than a copy, because APFS implements hard links through a hidden indirection inode. 200 files of ~6KB, median of 6 runs, this machine:

| | median |
|---|---|
| `writeFile` from memory | 20.5ms |
| `copyFile` | 24.8ms |
| `link` | 62.5ms |

Same ordering under Node 25.6.1 and with the sync variants, so it is the filesystem and not Bun. Shipping a measured regression on the machine I can measure, to buy an unmeasured win on one I cannot, is not a trade this PR takes. It is the obvious next lever if a Windows leg is ever instrumented — which #2924 section 8 already asks for.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | every capture copies both bundled roots and re-reads all of it to hash | the bundled roots are read once per process; each capture writes those bytes and folds their known hashes |
| **Failure behavior** | a bundled file that vanished mid-capture fails the capture | unchanged; a stale or unverifiable cache re-reads the live trees rather than being used |
| **`manifest.byte_count` on a binary build** | counted the bundled files but not their bytes, so the number meant something different than on a source build | counts them, as a source build always did |

That last row is drift the change lands on rather than new behavior: `materializeBundledDefaults` added its file count to the manifest and never its byte count, so one branch's `byte_count` excluded the bundled scope and the other's included it. Both branches now go through one accumulator. Nothing validates the field — it feeds the 64 MB capture-size warning and diagnostics — and the bundled scope is ~1.2 MB against that ceiling.

## Architecture

```mermaid
flowchart LR
  subgraph Before
    B1[runChildWorkflow] --> B2[captureWorkflowSource]
    B2 --> B3[copyTree bundled roots]
    B3 --> B4[digestTree re-reads every copied byte]
  end

  subgraph After
    A1[runChildWorkflow] --> A2[captureWorkflowSource]
    A2 --> A3[resolveBundledScope: stamp check]
    A3 -->|stamp matches| A4[writeBundledScope from held bytes]
    A3 -->|stamp differs| A5[read both roots, hash, hold] --> A4
    A4 --> A6[foldDigest: held hashes + read the other scopes]
  end
```

### Changed seams

|---|---|---|
| capture digest (`manifest.digest`, persisted and re-verified across upgrades) | `digestTree` split into `digestEntries` + `foldDigest`; the sort key and hash input are unchanged | `workflow-source.test.ts:424` builds a capture as an older build wrote it and verifies it against the digest that build recorded |
| `captureWorkflowSource` bundled branch | one path for source and binary builds instead of two | `workflow-source-binary.test.ts` (5 pass), `workflow-source.test.ts` (39 pass) |

## Validation

- `bun run validate` — exit 0 — proves the whole repository still type-checks, lints, and passes.
- `cd packages/workflows && bun run test` — 26 groups, all `0 fail` — proves the package suite is green, `subrun.test.ts` included (100 pass).
- `bun run scripts/check-test-cleanup-drift.ts` — passes — proves the `LEGACY_RECURSIVE_CLEANUP` entry now matches what is actually cleaned. It failed with `7 recursive cleanup sites, recorded 8` and then `6 ... recorded 8` as each site was fixed, so the ledger was seen counting both fixes.
- **Cost, measured in situ.** A counted copy of the suite wraps every `fs/promises` export, resets counters immediately before `captureWorkflowSource` and dumps them immediately after, with the real `@archon/paths` getters and a one-file project tree, so everything counted below is the bundled scope. Seven captures each, after a warm-up capture:

  | | before (`eab56178f`) | after |
  |---|---|---|
  | filesystem operations | 836 | 591 |
  | of which `copyFile` / `writeFile` of bundled files | 201 | 201 |
  | of which `readFile` | 201 | 1 |
  | bytes read | 1,247,567 | 60 |
  | `readdir` | 90 | 47 |
  | median wall clock (no instrumentation) | 51.4ms | 33.7ms |
  | manifest `file_count` / `byte_count` | 201 / 1,247,567 | 201 / 1,247,567 |

- **Digest parity.** Both implementations captured the same fixture in one process: identical digest `0a60238db347374ba8b5bf5c960ec59b85aa4a7e5702aa0c09a8096dafc923ea`, identical file count, byte count, and scope list. `workflow-source.test.ts:424` keeps that pinned permanently against a capture written the old way.
- **Staleness, against the real bundled tree.** With no `@archon/paths` mock, in one process: capture (201 files), write `zz-staleness-probe.yaml` into the checkout's own `.archon/workflows`, capture again (202 files, probe present, digest changed), edit that file in place, capture a third time (new content present, digest changed again). `workflow-source.test.ts:361` locks all three transitions plus the removal.
- **Temp-root leak, seen red.** `subrun-test-empty-bundled-*` in `TMPDIR`: 483 before. Three runs of the file as it stands on `dev` took it to 486, one per run. Three runs with the fix left it at 486. (Those 3 are mine, from reproducing the leak; the pre-existing ~483 are not cleaned up here.)
- **Invariants held, each with the test that proves it:**

  | Invariant | Test |
  |---|---|
  | Late resolution stays inside `runChildWorkflow` at spawn time; no load-time existence check | `subrun.test.ts:4298` `describe('workflow: late resolution is a deliberate affordance')`, including `it('LOCK: discovery ACCEPTS a workflow: node whose target does not exist')` and `it('LOCK: sub-run discovery is NOT cached across spawns in one run')` |
  | A child captures per spawn from the parent's live authoring root, not the parent's frozen copy | `workflow-source.test.ts:596` `test('a not-yet-started child resolves from LIVE authoring, so a fix can land')` |
  | The bundled bytes are present in each run's own capture | `workflow-source-binary.test.ts:99` `test('the digest covers the bundled bytes, so an upgrade cannot change them unnoticed')` and `:83` `test('bundled workflows are written where discovery expects them')`; on the source branch, `workflow-source.test.ts:361` reads the bundled file back out of two separate captures |
  | GAP2180 keeps its two-child, gate, and path-lock assertions | `subrun.test.ts:4721` `it('GAP (#2180 Defect A): a GATING child loses the path lock before it ever reaches its gate')`, unchanged; only the enclosing `afterEach` changed |

  The `mock.module('@archon/paths', ...)` workaround was removed from nothing, per the brief. It is **still needed**: the bundled bytes still have to be written into each capture, so the write fan-out those five files were added to avoid is unchanged. What went away is the read and hash of the same bytes. Removing any of them is a separate decision with its own measurement.

- **Not verified:** Windows. Every number above is macOS/APFS. The operation count is fixed by the code and platform-independent; the per-operation latency is not, which is the whole reason #2924 exists. The `link` measurement that decided against hard links is also macOS-only, and is the one result a Windows runner could overturn.

## Delivery considerations

|---|---|---|
| Compatibility / migration | `manifest.digest` is persisted and re-verified by later builds; a changed fold would refuse every paused run on disk. The fold is unchanged, including the platform-separator sort that only differs on Windows. | `workflow-source.test.ts:424`, whose fixture puts `pack-extra.yaml` beside a `pack/` directory precisely because `/` and `\` sort differently against `-` |
| Observability | An invalidated cache re-reads silently, as a copy always did. The capture's own digest check remains the loud failure for anything that does go wrong. | `loadWorkflowSource` unchanged |
| Rollout / rollback | Pure revert. No persisted shape, no on-disk cache directory, no new configuration. | — |

**Deferred, deliberately:** the `discoverWorkflowsWithConfig` memo keyed on capture digest that #2924 section 5 raises "secondarily". This PR removes the dominant term; a discovery cache in an agent runtime carries its own staleness risk (a parsed catalogue held across a mid-run authoring edit) and deserves its own justification and its own measurement rather than riding along here.

**Also observed, not acted on:** a source build captures *all* of `.archon/workflows` and `.archon/commands` (200 files here) into the bundled scope, while a binary materialises only what it embeds (97 files). The two branches disagree about what "the bundled scope" is. Narrowing the source branch would roughly halve what every capture writes, but it changes what discovery can resolve in a source install and needs its own issue.

**Ledger rows this discharges on #2924:** the GAP2180 `children: 0` teardown row (the raw recursive `rm` at `subrun.test.ts:4335` that deleted staged captures out from under a still-running body), the `afterAll` temp-root leak recorded in the addendum, and the per-capture bundled-scope materialisation the analysis asked to be opened as its own product-cost row. #2924 stays open as the rolling flake-class ledger.

## Links

- Related #2924
- Related #2306

https://claude.ai/code/session_01XqEroYXiJbgGesqZqyUgZ6


Wirasm:
## Ready to merge

`captureWorkflowSource` used to copy and re-hash Archon's whole bundled-defaults tree on every child-workflow spawn; this PR reads and hashes it once per process and revalidates before reuse instead. The central risk — whether the revalidation key (tree shape plus each file's size and mtime) can silently serve stale bundled content — got the deepest scrutiny. It traces to a real, code-comment-acknowledged blind spot (a same-size edit landing within the same mtime tick), but the code reviewer traced Archon's actual shipped deployment surface (the `Dockerfile`, which never compiles a binary, and `docker-compose.yml`, which does not volume-mount the bundled-defaults directory) and could not construct a path to it: the bundled root always sits on a nanosecond-resolution local filesystem (APFS/ext4/NTFS) under every configuration Archon ships or documents. Clock-backward movement, stat-error handling, and the persisted cross-upgrade digest contract were each traced independently and confirmed correct — the equality-based stamp (not a "newer than" check) catches backward moves by construction, an absent file is textually distinct from any present one, and the digest fold was proved algebraically equivalent to the pre-PR version, not just trusted from the PR's own fixture test. The seam reviewer confirmed the cache-key design has no structural gap (single-meaning singleton, no leak of held bytes into the persisted digest, one shared constructor for both build branches, the freshness key's root paths provably match the exact tree the mutating API routes can write to). Local validation is clean and matches the PR's own numbers exactly, and the one red CI leg is a confirmed pre-existing, unrelated flake.

**Validation:** `bun run lint` PASS (includes the AST-based `check-test-cleanup-drift.ts` ledger check, which independently proved the 8→6 count for `subrun.test.ts`). `cd packages/workflows && bun run test` PASS — 26 test groups, 0 fail, exact pass counts matching the PR body (`workflow-source.test.ts` 39, `workflow-source-binary.test.ts` 5, `subrun.test.ts` 100). CI: `test (ubuntu-latest)` PASS (159/0 fail in `@archon/workflows`, matching local), `docker-build`/`postgres-parity`/`schema-upgrade`/`workflow-fixtures`/CodeRabbit PASS, `test (windows-latest)` FAIL — confirmed via direct job-log fetch to be `downloadWebDist > verifies against the embedded hash without fetching checksums.txt` timing out at exactly 5001.78ms in `packages/cli/src/commands/serve.test.ts`, a file this PR does not touch; an already-tracked flake family on issue #2924's ledger. The `Test Suite` rollup fails solely because it gates on that same leg.

### Findings

|---|---|---|---|
| `R1` | Suggestion | Freshness key has a real but unreachable-under-shipped-config blind spot (same-size, same-mtime-tick edit); zero test exercises that exact boundary | OPEN |
| `R2` | Suggestion | No test proves the perf-critical cache-hit path actually avoids re-reading an unchanged tree | OPEN |
| `R3` | Suggestion | Module-level `bundledScope` singleton has a harmless check-then-set race under concurrent fan-out captures | OPEN |
| `R4` | Suggestion | Cross-build digest-compatibility guarantee rests on one hardcoded fixture, not a dynamic comparison | OPEN |
| `R5` | Suggestion | Two PR-description framing points worth tightening (Docker's actual build branch; binary `byte_count` is a bugfix, not new drift) | OPEN |

<details>
<summary><code>R1</code> — Freshness-key blind spot is real but not reachable under Archon's shipped deployment</summary>

**Impact:** `stampListings` (`workflow-source.ts:566-584`) keys bundled-scope reuse on tree shape plus each file's `size`+`mtimeMs`. Its own comment admits: "It cannot see an edit that preserves both." A same-size, same-tick edit to a bundled workflow/command file on a long-lived `archon serve` would let a subsequent child spawn silently capture the stale (pre-edit) bundled bytes — self-healing on the next detectable edit, but silent while it lasts.

**Evidence:** `workflow-source.test.ts:361-411` ("a later capture in the same process sees a bundled-scope edit") is the only staleness test and only exercises a size-changing edit (25→46 bytes), never the same-size case. The code reviewer then traced reachability directly: the production `Dockerfile` never runs `bun build --compile` (comment: "Bun runs TypeScript directly, no compile step needed"), so the Docker distribution runs the same stat-based source branch (`workflow-source.ts:588` else-branch) as a maintainer's local checkout — correcting an initial assumption that Docker used the binary/`BUNDLED_VERSION` branch. But `docker-compose.yml:47-49` only volume-mounts `/.archon` (the ARCHON_HOME workspace data) and `/home/appuser`; `/app/.archon`, where the bundled-defaults tree actually lives inside the image, is not bind- or volume-mounted, so it sits on the container's overlay2 writable layer backed by the host's ext4/xfs/APFS — all nanosecond-resolution. The same holds for a maintainer running from a local git checkout. Reaching the blind spot needs an NFS/FUSE/9p-backed volume specifically for the bundled-defaults directory, which is not part of any configuration Archon ships or documents. The seam reviewer separately confirmed the key's root paths exactly match the tree the mutating `PUT`/`DELETE /api/workflows/:name` routes can write to (both resolve through the same `getDefaultWorkflowsPath()` — no independent duplication to drift), and that clock-backward movement is caught by construction: `resolveBundledScope` does strict `!==` equality on the folded stamp, not a "newer than" comparison, so a backward mtime move still changes the hash.

**Required outcome:** No change required to ship. If Archon ever documents or supports a network-filesystem-backed bundled-defaults root, this becomes reachable and should get a test plus a stronger key (or explicit non-support) at that time.

**Found by:** `prp-core:code-reviewer`, `prp-core:pr-test-analyzer` (seam-analyzer independently confirmed the key's target-root correctness and clock-backward handling, a related but distinct question)

**Disposition:** OPEN, Suggestion. The test-analyzer rated this Important based on the code's own comment admitting the gap and zero coverage of it; the coordinator downgrades to Suggestion because the code reviewer's direct trace of the actual Dockerfile and docker-compose volume configuration found no reachable path under any shipped or documented deployment topology — per this repo's reject-speculative-defense-in-depth standard, a theoretical gap with no reachable trigger in supported configurations is a documented residual risk, not a blocking defect. Worth a regression test to lock the documented limitation, not worth blocking merge.

</details>

<details>
<summary><code>R2</code> — No test pins the cache-hit fast path (the PR's actual perf claim)</summary>

**Impact:** Every existing test proves change is detected; none proves that an *unchanged* bundled tree causes `resolveBundledScope` (`workflow-source.ts:587-615`) to actually reuse the held scope rather than coincidentally re-reading to the same result. A future refactor that silently dropped the memoization while preserving correctness would pass the full suite and surface only as an unnoticed performance regression — for a PR whose entire point is avoiding re-reads.

**Evidence:** `pr-test-analyzer` read the full 1026-line `workflow-source.test.ts` and found no spy/counter on `readFile`/`stat` calls and no object-identity check across two captures of an unchanged tree.

**Required outcome:** Not required for this PR's correctness; a follow-up test spying on read/stat counts or checking `BundledScope` identity across two same-tree captures would close this durably.

**Disposition:** OPEN, Suggestion. Coordinator downgrades from the analyzer's Important: the current delivered behavior is correct (externally benchmarked in the PR body and consistent with the code reviewer's independent structural trace of what `resolveBundledScope` does on each branch), so this is a coverage gap against future regression, not a defect in what ships now.

</details>

<details>
<summary><code>R3</code> — Harmless race on the module-level bundled-scope cache under concurrent fan-out</summary>

**Impact:** `resolveBundledScope`'s source branch (`workflow-source.ts:600-614`) does a check-then-set on `let bundledScope` (`:556`) straddling multiple `await`s. Two concurrent `captureWorkflowSource` calls (real under fan-out — `subrun.test.ts`'s `makeFanResolver` gives each of "N concurrent siblings" its own worktree specifically because they run concurrently) can both miss the cache and both redo the read-and-hash pass.

**Evidence:** Confirmed independently by two reviewers. The code reviewer traced that there is no `await` between a call's own local computation and its `return`, so each concurrent call always returns what *it* computed, correct and self-consistent; only the shared module cache gets overwritten by whichever call finishes last (last-write-wins, content-equivalent under a stable tree). Worst case is redundant work, never a wrong result, and never worse than the pre-PR baseline (which always redid the work).

**Required outcome:** No change required. A follow-up could memoize the in-flight promise to remove the redundant work under concurrent fan-out.

**Found by:** `prp-core:code-simplifier`, `prp-core:code-reviewer`

**Disposition:** OPEN, Suggestion — non-blocking, both reviewers agree it's strictly non-regressive versus baseline.

</details>

<details>
<summary><code>R4</code> — Digest cross-build compatibility rests on one fixed fixture</summary>

**Impact:** The persisted, cross-upgrade-verified `manifest.digest` contract is protected by `workflow-source.test.ts:424-458`, a single hardcoded fixture plus a hardcoded expected digest string, rather than a dynamic old-algorithm-vs-new-algorithm comparison over arbitrary trees.

**Evidence:** `pr-test-analyzer` found no second such test. Risk is bounded: any future divergence fails closed with `WorkflowSourceIntegrityError` (refuses resume) rather than silently corrupting, consistent with this repo's fail-clearly doctrine. The code reviewer additionally proved the fold is algebraically equivalent to the pre-PR `digestTree` by direct comparison (same sort key over platform-separated paths, same `/`-normalized hash input), not merely trusting the fixture.

**Required outcome:** Not required; a property-style test folding an old-style and new-style digest over the same randomized tree would close the residual gap.

**Disposition:** OPEN, Suggestion, non-blocking.

</details>

<details>
<summary><code>R5</code> — Two PR-description framing points worth tightening</summary>

**Impact:** (a) The PR's own framing implies the stat-based revalidation path is scoped to source/dev checkouts rather than the Docker distribution; per `R1`'s evidence this is inaccurate (Docker also runs the source/stat branch) even though it doesn't change the reachability conclusion. (b) Binary builds' `manifest.byte_count` now includes bundled bytes where it previously didn't — this is a correction of a prior undercount (the old binary path tracked bundled file count but never bytes, while the old source path always tracked bytes), not unexplained new drift, and should be described that way for a future reader of the PR history.

**Evidence:** `Dockerfile` comment: "Bun runs TypeScript directly, no compile step needed"; `docker-compose.yml:47-49` volume list; diff shows `materializeBundledDefaults`'s old binary path adding only to `fileCount`, never `byteCount`.

**Required outcome:** Optional PR-description edit before merge; not required.

**Disposition:** OPEN, Suggestion, non-blocking, purely descriptive.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | R1 (reachability evidence), R3, R5 |
| seams | No findings — freshness-key design, singleton discriminant, digest/stamp separation, and mutation-surface match all traced clean |
| simplify | R3; also judged the four cache-machinery shape questions (shared builder, listTree/copyListing split, digest-function split, duplicate-looking zero-length guards) all justified, no simplification finding; separately flagged that the PR addresses only the read/hash half of #2924, not the per-file write cost — a scope note for the operator, not a code defect |
| tests | R1 (test-gap evidence), R2, R4; also verified the four named `subrun.test.ts` invariant-lock tests (late-resolution ×3, GAP #2180) present and unmodified in logic |

One scope note outside the findings table, for the operator: the simplify reviewer flagged that this PR removes the read+hash cost per capture but not the per-file *write* cost (`writeBundledScope` still does one `writeFile` per bundled file per capture) — and issue #2924's motivating problem is the Windows CI budget, which scales more with writes than reads. The PR's own `subrun.test.ts` comment already concedes this ("the writes remain and so does this lever"). The title is honestly scoped (doesn't claim to close #2924), so this isn't a misrepresentation, but it does mean #2924 should likely stay open pending a follow-up on the write side rather than being closed by this PR alone.

The hard-link alternative (would remove the per-capture write cost too, the Windows-relevant half) was measured and rejected: `link` at ~62.5ms vs `copyFile` at ~24.8ms (median of 6 runs, macOS APFS). Both the code and simplify reviewers independently agree with declining that trade — shipping a measured local regression to chase an unmeasured Windows win would be the wrong call, especially since the shipped change already shrinks the Windows-relevant write-side cost via removing the read/hash contention around it.

#### Validation

|---|---|---|
| `bun run lint` | PASS | Includes `check-test-cleanup-drift.ts`'s AST-based ledger check; printed "Test-cleanup drift check passed," independently proving the `subrun.test.ts` 8→6 count |
| `cd packages/workflows && bun run test` | PASS | 26 test groups, 0 fail; `workflow-source.test.ts` 39 pass, `workflow-source-binary.test.ts` 5 pass, `subrun.test.ts` 100 pass — matches PR body exactly |
| CI: `test (ubuntu-latest)` | PASS | `@archon/workflows:test` 159 pass / 0 fail, matching local |
| CI: `test (windows-latest)` | FAIL (pre-existing, unrelated) | `downloadWebDist > verifies against the embedded hash without fetching checksums.txt` times out at exactly 5001.78ms (both occurrences) in `packages/cli/src/commands/serve.test.ts`, a file this PR does not touch; matches issue #2924's already-tracked flake family, failed the same way on #3258 the same day |
| CI: `docker-build`, `postgres-parity`, `schema-upgrade`, `workflow-fixtures`, CodeRabbit | PASS | — |

</details>

PR #3278, opened 2026-09-10
fix(workflows): the deliver pack's classify node ran on the default assistant, not the run's tiers

Wirasm:
## Problem and outcome

`archon-deliver.yaml` is the only SDLC pack workflow with no top-level `model:`, and its `classify` node declared none either. That node therefore resolved through the install's default assistant rather than the run's tiers, so a run launched with `--config` pinning every tier to one provider executed `classify` on a different one and spent its quota.

- **Outcome:** every AI node in the SDLC pack resolves a tier, and a node that stops doing so fails a test rather than silently changing provider at runtime.
- **Invariant:** which tier a node names stays an ordinary authoring choice that changes freely; that it resolves one at all is enforced.
- **Scope boundary:** no other pack workflow changed. The other 15 AI nodes already resolve a tier — 2 by their own declaration, 13 through their workflow's `model:` — and none gained a restated copy of a default it already inherits.
- **Root cause:** tier resolution falls back to the install's default assistant when neither the node nor its workflow declares a model, and nothing asserted that pack nodes declare one.

## Review guidance

- **Feedback requested:** whether `small` is the right tier for `classify`, and whether the guard's definition of "AI node" is complete.
- **Start here:** `.archon/workflows/sdlc/deliver/archon-deliver.yaml:92` — the one behavior change.
- **Review order:** the workflow change, then the guard in `packages/workflows/src/defaults/bundled-defaults.test.ts`, then the regenerated bundle.
- **Lower-attention areas:** `bundled-defaults.generated.ts` is regenerated by its owning script; `bun run check:bundled` reports it up to date.
- **Known risk or uncertainty:** the guard treats `agent` and `loop` as provider-invoking and `loop_group` as a container that forwards its model to its body. That matches the schema documentation today; a future node kind that invokes a provider would need adding to the check.

## Solution

`classify` gains `model: small`. Judging two booleans from a diff is small-tier work, and the node runs on every delivery.

The guard walks the parsed bundled workflows owned by the `sdlc` pack and asserts that every provider-invoking node resolves a tier — from its own declaration, from an enclosing `loop_group`'s forwarded one, or from its workflow's default. It reads the parsed representation rather than YAML text, so it tests the shipped semantics, and it derives its node types from `parseWorkflow`'s return type rather than restating the node union, which would drift as node kinds are added.

Deliberately not done: adding an explicit `model:` to the 13 nodes that inherit their workflow's default. Each would be a hand-synced copy of a value that already has an owner, and the guard makes the copies unnecessary — inherited coverage counts.

`archon-deliver.yaml` deliberately keeps no top-level `model:`, so the guard actively checks each of its AI nodes rather than being short-circuited by a workflow default.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | `classify` runs on the install's default assistant, ignoring the run config's tiers | `classify` runs on the run config's `small` tier |
| **Failure behavior** | A pack node with no tier resolves silently to another provider at runtime | A pack node with no tier fails `bundled-defaults` in CI, named in the failure |

## Validation

- `bun run validate` — exit 0 — type-check, lint, format, generated-artifact checks and tests across every package.
- `bun run check:bundled` — up to date (54 commands, 31 workflows, 1 script pack) — proves the regenerated bundle matches its source.
- Mutation-proven: deleting `model: small` and regenerating fails the guard with `["archon-deliver:classify"]`; restoring it passes. The guard was observed red before the fix existed, not only green after.
- **Not verified:** no live run was executed on the changed pack. The quota exhaustion that exposed this is still in effect on the reporting install, so end-to-end confirmation that `classify` runs on the configured provider waits for the next delivery run.

## Links

- Related #3273

https://claude.ai/code/session_01WDeeEmfBHksoyoeCr1yDYC

## Summary by CodeRabbit

- **Improvements**
  - Workflow classification now uses the small model tier, helping optimize processing for this step.

- **Tests**
  - Expanded validation of model-tier resolution across workflow-level, node-level, and loop-group configurations.
  - Improved coverage for provider-specific model settings and invalid workflow definitions.


Wirasm:
Review round 1 (Sonnet `prp-review`, three lenses) found two soundness gaps in the guard itself. Both verified against the runtime before fixing, both closed in `1de52a90a`.

**R1 — a `loop_group`'s own `model:` covers nothing at runtime, though the schema says it does.** `dag-executor.ts:10394` destructures only `{ provider: loopGroupProvider }` from `resolveNodeProviderAndModel`, discarding the resolved model, and the per-iteration body context at `dag-executor.ts:5114` carries `workflowModel: ctx.workflowModel` — the outer workflow's. But `schemas/dag-node.ts:541-546`, and the comment directly above that destructure, both state that `model`/`provider` are forwarded to body AI nodes. The guard trusted the comment, so an author following the documented contract would write `loop_group: { model: small }` around a bare body node, pass the guard, and hit the exact silent provider switch it exists to catch.

**R2 — a workflow's `model:` does not reach a node naming a different provider.** `include-expander.ts:366-369` (`workflowModelTravelsTo`) lets a workflow's model through only when `nodeProvider === undefined || nodeProvider === scope.provider`, and line 395 skips pushing `model` otherwise. The blanket `if (workflow.model !== undefined) continue` skipped walking 9 of 10 pack workflows without applying that rule.

**R3 — an unrelated comment edit that was mine.** Repairing heredoc escaping in the first commit, a global backslash replace stripped one from a comment 400 lines away that was describing backslash-continued shell lines. Restored.

Coverage is now decided per node against the resolver's rules rather than the documentation: a node's own `model:` wins; a workflow's `model:` covers a node only when the node resolves to the workflow's own provider; a `loop_group`'s model covers nothing. That also retired the inherited-coverage parameter, so the walker is smaller than before.

Each rule was mutation-proven red, then restored:

| mutation | guard reports |
|---|---|
| drop `model: small` from `classify` | `archon-deliver:classify` |
| `loop_group` with its own model + bare body node | `archon-deliver:corrections/probe-mutation` |
| node `provider:` differing from its workflow's | flagged |

`bun run validate` exits 0.

Note for whoever picks it up: R1 is a live engine/documentation mismatch independent of this PR. Two comments assert a forwarding behaviour the runtime does not implement, and they actively steer an author into the bug. Worth its own issue — either make the runtime forward the model as it already forwards the provider, or correct both comments.

Issue #3281, opened 2026-09-10
bug(workflows): a loop_group forwards its provider to body nodes but silently drops its model, while three comments disagree about which


Wirasm:
## Problem

A `loop_group` node's own `model:` reaches nothing. Its `provider:` does. Two comments state that both travel, and one of them sits directly above the line that discards the model.

`packages/workflows/src/schemas/dag-node.ts:541-546`:

> `model`/`provider` are forwarded to body AI nodes unless overridden per-node (same forwarding `loop:` uses).

`packages/workflows/src/dag-executor.ts:10391-10392`:

> group-level provider/model overrides are forwarded to body AI nodes

and immediately below it, `dag-executor.ts:10394`:

```ts
const { provider: loopGroupProvider } = await resolveNodeProviderAndModel(
  node, ctx.workflowProvider, ctx.workflowModel, ...
);
```

The resolver computes the group's provider *and* model; the destructure keeps only the provider. The per-iteration body context at `dag-executor.ts:5114` then carries:

```ts
workflowProvider,                  // the group's, forwarded
workflowModel: ctx.workflowModel,  // the enclosing workflow's, not the group's
```

whose own comment describes the real behavior — "Forward inherited **workflow-level** model/tier/options/profile" — and so contradicts the two above it. `dry-run.ts` reports the same way, so a dry run agrees with the code and disagrees with the documentation.

## Why this is worth solving

An author who reads the schema and writes `loop_group: { model: small }` around a body AI node gets the enclosing workflow's model instead, silently. If that workflow declares none either, the node falls through to the install's default assistant — a run pinned to one provider executes that node on another and spends its quota. That is the same failure #3272 and PR #3278 exist to close, reached this time by following the documentation.

It also defeats a guard that now depends on the answer. PR #3278 adds a bundled-defaults check that every SDLC pack AI node resolves a tier. Its first version trusted these comments and was consequently unsound; it now encodes the opposite rule with a comment explaining why. Whichever way this issue is settled, that guard has to move with it, and today the repository states both answers in three places.

## Why now

The contradiction is live and already cost one wrong implementation, caught in review rather than in production. Nothing in the bundled pack sets a `loop_group` model today, so the behavior can still be settled without a compatibility story. That stops being true the first time a pack — bundled or third-party — relies on either reading.

## Desired outcome

A `loop_group`'s `model:` and `provider:` behave the same way as each other, and the schema, the executor, the dry-run report and PR #3278's guard all describe that one behavior.

[...]

PR #3284, opened 2026-09-10
fix(workflows): a loop_group forwards its model to body nodes, matching its provider

Wirasm:
## Problem and outcome

A `loop_group` forwarded its resolved `provider:` to body AI nodes but silently dropped its `model:`. `resolveNodeProviderAndModel` computed both; the dispatch site kept only the provider. A body node declaring no model therefore took the enclosing workflow's — or, when the workflow declared none either, the install's default assistant, so a run pinned to one provider executed that node on another.

Two comments promised the forwarding that did not happen: `schemas/dag-node.ts:541-546` and the comment directly above the line that discarded the model. A third, on the body context, described the real behaviour and contradicted them both.

- **Outcome:** a `loop_group`'s `model:` reaches its body AI nodes as their default, the way its `provider:` already did.
- **Invariant:** a body node's own `model:` still wins over whatever the group forwards.
- **Scope boundary:** the executor's other forwarding paths are untouched; only the loop_group body's provider and model change.
- **Root cause:** `dag-executor.ts` destructured only `provider` from a resolver that returns both, and the body context was built from `ctx.workflowModel` rather than the group's.

## Review guidance

- **Feedback requested:** whether forwarding is the right resolution versus correcting the comments to document the asymmetry, and whether the dry-run gap should block this.
- **Start here:** `packages/workflows/src/dag-executor.ts` — the destructure at the loop_group dispatch, and `workflowModel` in the per-iteration body context.
- **Review order:** the executor change, the regression test, then the guard update in `bundled-defaults.test.ts`.
- **Known risk or uncertainty:** this changes resolution for any workflow that sets a `loop_group` model. Nothing bundled does, and the new behaviour is what the schema already documented, so a workflow relying on the old behaviour was relying on undocumented behaviour. Third-party packs are the population I cannot check.

## Solution

Keep the model the resolver already returns, pass it through `executeLoopGroupNode`, and bind it as the body context's `workflowModel`. Body nodes then resolve aliases and defaults exactly as top-level nodes do, and a per-node `model:` still takes precedence.

The three comments now agree. The dispatch-site comment states the forwarding as fact rather than intent; the body-context comment describes what actually arrives; the schema comment needed no change because it already described this.

The tier-coverage guard added in #3278 moves in the same commit, as that issue required: a `loop_group`'s own `model:` covers its body again.

Dry-run follows in a second commit. `simulateLoopGroup` passed its context unchanged into the body, so a report described the body against the enclosing workflow's scope while the run resolves it against the group's. It now derives the body context from the group's own resolution through the same `resolveNodeModel` the executor's chain uses — never a second copy of the rules. Provider, model and the fields describing where the inherited model came from travel together; workflow-level effort and options stay put, matching what the executor actually forwards.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A body AI node under `loop_group: { model: small }` resolves the workflow's model, or none | It resolves the group's `small` |
| **Failure behavior** | With no workflow model either, the node silently ran on the install's default assistant | The group's model applies; a node resolving nothing is caught by the pack tier guard |

## Validation

- New regression test `forwards a loop_group tier to a body AI node that declares none`, mutation-proven: without the executor change it fails `Expected: "haiku" / Received: undefined`, with it passes. The `undefined` is the point — the body node was resolving no model at all, not merely the wrong one.
- `bundled-defaults` tier-coverage guard still passes with its loop_group rule inverted to match the new runtime.
- New dry-run test `reports a loop_group body node against the group own resolved model`, mutation-proven: without the dry-run change it fails `Expected: "gpt-5.6-sol" / Received: "claude-default"` — a different provider's default model than the run uses — and passes with it.
- **Not verified:** nothing material. Both halves of #3281 are covered by mutation-proven tests, and no live workflow sets a `loop_group` model, so there is no existing behaviour to regress.

## Links

- Related #3281
- Depends on #3278 — this PR targets `fix/sdlc-pack-node-tiers` because it updates the guard that PR introduces.

https://claude.ai/code/session_01WDeeEmfBHksoyoeCr1yDYC

## Summary by CodeRabbit

- **Bug Fixes**
  - Loop groups now consistently apply their configured provider, model, tier, and preset-derived effort to body nodes without their own settings.
  - Workflow execution and dry runs now produce matching provider, model, tier, and preset resolution for loop-group bodies.
  - Approval gates inside loop groups now correctly pause the overall run.
  - Body nodes can still override inherited settings with their own model or provider.

- **Tests**
  - Added coverage for inheritance, tier and preset propagation, default handling, and approval-gate behavior.


Wirasm:
Both findings verified against the resolver and fixed in `e75dfb36d`. Both were the same rule applied at one level and not the neighbouring one.

**Forward the resolved preset and tier with the model — valid, and the consequence is larger than a reporting mismatch.**

A tier ref resolves to a provider, a model and an effort as one unit. The body context took the group's provider and model but kept `workflowPreset` and `workflowLevelOptions.workflowTier` from the enclosing workflow, so a workflow at `large` around a group at `small` ran `haiku`, reported tier `large`, and applied the `large` preset's `max` effort to it. That is unapplied configuration attributed as active.

`resolveNodeProviderAndModel` already computed the preset internally without returning it; it now does, and the tier and preset are threaded through `executeLoopGroupNode` into the body context. Mutation-proven — reverting the forwarding fails the regression with `Expected: "low" / Received: undefined`, and the test also asserts the body's `node_started` carries tier `small` and model `haiku`.

**Keep provider matching in inherited coverage checks — valid.**

`inherited || covered(node)` marked every descendant of a model-bearing `loop_group` as covered regardless of provider. The guard now carries a resolution scope rather than a boolean: an inherited model reaches a node only when the node resolves to the scope's provider, at both the workflow and the loop_group level. That is the same `workflowModelTravelsTo` rule the workflow level already followed, and applying it uniformly retired the workflow-level short-circuit that hid the gap.

The divergent-provider body node is now a proven case. Adding a `provider: codex` body node under a Claude-provider group with a model makes the guard report `archon-deliver:corrections/probe-mutation`; the previous version passed it.

Full battery, each mutation restored afterwards: classify losing its tier, a node provider differing from its workflow's, and a body node provider differing from its group's — all red, pack green. `bun run validate` exit 0 and 48/48 workflow fixtures pass.

PR #3289, opened 2026-09-10
test(cli): give the run-boot wait one deadline and a duration

### LINKED ISSUE #3288: test(cli): workflow wait stalls 30s on windows with an EBUSY-locked fixture
## Problem

`archon workflow wait against a detached run > wakes with awaiting_response on a gate, then with cancelled when it is rejected` fails on Windows after **30476 ms**.

This is **not** the 5000 ms budget-edge class that #2924 tracks. The test declares its own **120 s** budget (`packages/cli/src/commands/workflow-wait.integration.spec.ts:536`), so Bun never cut it off. What expired is the `waitFor` helper's internal 30 s deadline (`:354`). Something the test waits for genuinely never arrived.

That makes it a real failure, not an amplification artifact. Every other Windows failure in this repository lately has been a 141 ms test inflated ~37x; this one waited half a minute for a thing that did not happen.

## Evidence

Sample: `dev` at `bf554a782`, job 102398739398.

The line immediately before the failure is the interesting one:

```
temp cleanup failed for C:\Users\RUNNER~1\AppData\Local\Temp\archon-wait-gate-XIObbe:
  Error: EBUSY: resource busy or locked, rm '...'
(fail) archon workflow wait against a detached run > wakes with awaiting_response on a gate,
  then with cancelled when it is rejected [30476.23ms]
```

`EBUSY` means a process still held that directory when cleanup ran. That is the same Windows file-locking shape that caused the teardown half of the GAP2180 sample, where a raw recursive `rm` fought handles a live process still owned.

**What I cannot tell from one log line** is the direction. Either a still-running child kept the fixture alive and the test then waited on a run that could not progress, or the test failed first and the `EBUSY` is a downstream symptom of its abandoned children. Both are consistent with what CI printed, and separating them needs a live machine.

## Why this needs a Windows machine

The whole test is about process lifecycle across processes: a detached run, a foreground owner, a gate, and a third process rejecting it. `EBUSY` is Windows-specific — POSIX unlinks an open file happily. So the failure mode cannot exist on the platforms where it is easy to debug.

@coleam00 — assigning to you because you are on Windows. What would settle it is running this one spec locally and watching which handle stays open, with Process Explorer or `handle.exe` on the `archon-wait-gate-*` directory.

## Desired outcome

Know which process holds the directory and whether that hold is the cause or the consequence of the 30 s wait, then fix the one that is wrong.

## Acceptance

- The stuck handle identified by owning process.
- A statement of direction: did the lock stall the run, or did the failed run leave the lock?
- The test passing on Windows, or an evidenced statement of what in the lifecycle cannot work there.
- No budget bump. Its budget is already 120 s and was never the constraint — raising anything here would only make it fail slower.

## Notes for reproduction

- The spec spawns real detached processes; it is one of the `.integration.spec.ts` files that runs in its own `bun test` invocation.
- Run it from inside the package: `cd packages/cli && bun test src/commands/workflow-wait.integration.spec.ts`. Never bare `bun test` at a package root — mock pollution produces a nonsense failure count.
- Sibling tests in the same file cover owner exit, third-process cancellation and `owner_lost`, so comparing which of them also leaves an `archon-wait-gate-*` directory behind may localise it quickly.

## Links

- #2924 — Windows ledger. Recording this here as **not** a member of the budget-edge class, so it does not get filed as one.
- #2306 — subprocess-spawning tests under `bun --filter --parallel`

Wirasm:
## Problem and outcome

`archon workflow wait against a detached run > wakes with awaiting_response on a gate, then with cancelled when it is rejected` failed on the Windows leg of `dev` after 30476 ms. That is not a test budget — the case declares 120 s and Bun never cut it off. What expired was the setup poll that discovers the run id, at the 30 s default inside the file's `waitFor` helper. Two sibling waits ask the same question and inherited that default silently, so nothing made the three move together.

- **Outcome:** the three waits on "has the run booted" share one named deadline, now 90 s, and each logs how long the boot actually took, so a boot creeping toward the ceiling is visible as a climbing number rather than as silence until it fails again.
- **Invariant:** no test budget and no contract deadline moves. No `}, N)` changed, and no `--timeout` handed to `startWait` changed. A launcher this process owns still fails fast instead of spending the deadline.
- **Scope boundary:** the `workflow get` retry after a terminal wake is a different question (the run is already terminal; it retries one CLI call against the SQLite commit window), so it keeps 30 s — now stated explicitly rather than inherited.
- **Root cause:** a setup poll deadline that was too small for the platform, held by three call sites with no shared owner. The failing sample's last read reported `no such table: remote_agent_workflow_runs`, so the launcher had created `archon.db` and had not yet applied the schema when the poll gave up.

## Review guidance

- **Feedback requested:** whether 90 s is the right ceiling, and whether the boot-duration log line is the right signal.
- **Start here:** `packages/cli/src/commands/workflow-wait.integration.spec.ts:422` — `RUN_BOOT_DEADLINE_MS` and the reasoning that picks it.
- **Review order:** the constant and `waitForRunBoot`; then `waitFor`, whose `timeoutMs` lost its default so no fourth site can join by accident; then the four call sites.
- **Lower-attention areas:** the three call-site conversions are mechanical.
- **Known risk or uncertainty:** the true Windows boot time is unknown — the sample only proves it exceeds 30.4 s. 90 s is chosen against the local measurement and the healthy Windows samples in the same file, not against a measured Windows boot. The new log line is what will produce that measurement.

## Solution

One constant, `RUN_BOOT_DEADLINE_MS`, owns "how long a run launched in another process may take to boot far enough to be observable", and `waitForRunBoot` is the only way to spend it. `waitFor`'s `timeoutMs` is now required, so the implicit-default shape that produced the drift cannot recur.

Why 90 s. The boot measures 360–464 ms locally across four runs. Siblings in the same file passed on that same Windows runner at 8.2 s and 41.4 s, against ~16 s for all ten tests locally, and the amplification measured elsewhere in this repository is ~34x (676 ms on the runner against ~20 ms locally, #2924). 90 s is roughly 195x the local sample and an order of magnitude past anything that runner has been seen doing healthily. It also stays 30 s under the 120 s budgets on purpose: a genuine stall must surface as `Timed out waiting for …` carrying the launcher's streams, not as Bun killing the test with nothing to read.

`waitForRunBoot` logs `[boot] <what>: <elapsed>ms of 90000ms` on success only. The failure path already reports itself.

The owner-exit guard is what makes raising the deadline safe rather than merely slower at the two sites whose launcher this process owns, so the existing guard test now runs through `waitForRunBoot` at the same deadline. Its own budget is Bun's 5 s default, so a guard that stopped firing could only fail there — it could never pass slowly.

**On the `EBUSY` line in the original report.** It is downstream debris, not the cause. `removeTempTree` is reachable only from this file's `afterEach`, and `cleanupPaths` is drained nowhere else, so the warning cannot be emitted before the test body ends. The log agrees: the throw is at `08:53:37.2279`, the `EBUSY` warning at `08:53:37.6906`, and that 460 ms gap is almost exactly `removeTempTree`'s own retry budget (10 attempts, 50 ms apart). The lock was first observed after the wait had already given up, so it cannot have stalled the run. The cleanup already uses the retrying helper and warns rather than fails, and the tree is a runner temp directory, so nothing is changed for it. Which handle held it is not established.

**Correcting one claim in #3288's framing.** The sample ran `dev` at `bf554a782`, and the `owner` parameter on `waitFor` — the owner-exit fast-fail and the diagnostics on the timeout — landed 32 minutes later in `ecb06ac30`. `git merge-base --is-ancestor ecb06ac30 bf554a782` is false. So that sample could not distinguish a slow launcher from a dead one, and nobody should read it as proof the launcher was alive. On current `dev` it can: the guard reports the launcher's exit code and both streams.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | Three waits on one condition, one naming 30 s and two inheriting it. No record of how long a boot took. | One named 90 s deadline for all three, and a `[boot] …` line per boot in the CI transcript. |
| **Failure behavior** | The gated case gave up at 30 s on a Windows boot that had not finished. | A boot has 90 s; a launcher this process owns still fails in tens of milliseconds with its exit code and both streams, while a detached one spends the deadline and reports the plain timeout (R4). |

## Validation

- `cd packages/cli && bun test src/commands/workflow-wait.integration.spec.ts` — 10 pass, 0 fail — boot durations 360–464 ms locally across four runs.
- `cd packages/cli && bun run test` — 17 isolated invocations, 0 fail — proves the package suite is unaffected.
- `bun --filter '*' type-check` — exit 0 for all eight packages, run separately because it catches unused declarations lint does not.
- Owner-exit guard seen red before being kept: with the `Owner exited before …` throw removed, `reports owner exit and both streams before the row-discovery deadline` ran **211634 ms** and failed on `Timed out waiting for the failed run row`. Restored, the same test passes in **35 ms**.
- Owner-exit guard proved at the real boot site: replacing the gated launcher with `process.exit(7)` failed the gate case in **17.68 ms** with `Owner exited before the gated run row: owner pid …, exit 7 … stderr: gated launcher died`, against the 90 s deadline. Both mutations were reverted; the committed diff touches one file.
- **Not verified:** the Windows leg itself. The true boot time on that runner is unknown, and CI on this PR is the first place the new `[boot]` line will report it.

## Links

- Closes #3288
- Related #2924


Wirasm:
## Ready to merge

Single-file test-infrastructure fix in `packages/cli/src/commands/workflow-wait.integration.spec.ts`. It replaces three call sites that silently shared a 30s default inside a local `waitFor` helper with one named `RUN_BOOT_DEADLINE_MS` (90s) spent only through a new `waitForRunBoot` wrapper, and removes the default entirely so a future call site must state its own deadline. All four scopes found the change itself correct, narrowly scoped, and consistent with `.archon/engineering.md`. No Critical or Important findings survive independent verification. Five Suggestions, all non-blocking — two about diagnostic proportionality (an always-on log line, an over-detailed docblock), one about a naming stretch, one about failure-latency (not correctness) at two call sites that have no owner-exit guard to speed up, and one genuine but out-of-scope discovery: the identical "shared silent default" pattern this PR fixes still lives, unpatched, in `packages/core/src/services/run-live-owner.test.ts` (verified directly, 6 call sites).

The 90s deadline is a reasoned engineering estimate, not a bare budget bump: it is grounded in measured local boot time (360-464ms), two Windows-observed healthy siblings in the same file (8.2s, 41.4s), the file's own #2306 precedent for Windows commit-window widening, and `serve.ts`'s existing ~90x-of-slowest-healthy-sample convention for exactly this kind of "stuck vs. slow" boundary — which I independently confirmed still lives at `serve.ts:26` with the same ratio reasoning. The arithmetic in the PR's own comment checks out (90000/464 ≈ 194x, "roughly 195x" claimed). The one residual soft spot: measured against the worst *already-observed* healthy Windows sample in this file (41.4s), 90s is only ~2.2x headroom, not the ~90x margin the local-sample framing suggests, and this repository's own evidence elsewhere (issue #2924) shows Windows CI contention can swing by an order of magnitude between consecutive runs of the same commit. That's a real, acknowledged uncertainty — not a defect, since the PR does not claim to have measured true Windows boot time, and it ships exactly the right mitigation for it: the new `[boot] …` duration log, so the next Windows sample reports a real number instead of leaving it to inference. I'm not converting this into a finding because there's no reachable failure it points to beyond what the PR already discloses in its own "Still open" section.

On `#2924`'s "never fix a Windows failure with a budget bump" rule: this PR's defense holds. The 120s test-level budgets — the actual thing #2924 protects — are untouched, and they were never what expired in the original failure. What was raised is a *setup poll deadline*, a different question from the contract under test (`startWait(fixture, runId, 60)`, which drives the CLI's own `--timeout` and is also untouched). The failure was diagnosed to a specific cause (`no such table: remote_agent_workflow_runs` — schema not yet applied when the poll gave up), not "Windows is slow, add margin." I judge this a legitimate root-cause fix inside a distinct class from what #2924 forbids, not a rationalization of the same forbidden move wearing a new name.

Owner-exit fast-fail verified live, not just read: I mutated the guard in `waitFor` (removed the `throw` on `ownerExited`), re-ran the guard's test in isolation, and watched it fail after ~176s with the wrong error (`Timed out waiting for...` instead of `Owner exited before...`) instead of passing in 30ms with the right one. Restored the file and confirmed the original passes in 30ms with the correct diagnostic. This is the same directional result the PR's own mutation claims report (211,634ms / wrong error when removed), confirming the guard is real, load-bearing, and would catch a regression.

**Validation:** lint PASS · type-check PASS (all 13 packages, `bun --filter '*' type-check`, exit 0 everywhere) · `bun test src/commands/workflow-wait.integration.spec.ts` PASS (10/10) · `bun run test` (full `@archon/cli` package suite) PASS (0 fail across all files) · live owner-exit-guard mutation test: guard removed → fails wrong-way at ~176s; guard restored → passes correctly at 30ms.

### Findings

|---|---|---|---|
| `R1` | Suggestion | `waitForRunBoot` reused at a non-boot call site (owner-exit-guard unit test), a naming stretch | OPEN |
| `R2` | Suggestion | `[boot] …` log fires unconditionally on every successful run, permanent stdout noise, not gated | OPEN |
| `R3` | Suggestion | `RUN_BOOT_DEADLINE_MS` docblock embeds point-in-time investigation numbers that will rot | OPEN |
| `R4` | Suggestion | Two of four `waitForRunBoot` call sites pass no `owner`, so a dead launcher there times out in up to 90s instead of failing fast — latency, not correctness, within existing 120s test budgets | OPEN |
| `R5` | Suggestion | Same shared-silent-default `waitFor` pattern this PR fixes still lives unpatched in `packages/core/src/services/run-live-owner.test.ts` (6 call sites, verified) — real, but out of this PR's touched path per `AGENTS.md`'s own "correct what the change already touches, not what it merely noticed" | OPEN |

<details>
<summary><code>R1</code> — <code>waitForRunBoot</code> reused for a non-boot wait</summary>

**Impact:** Cosmetic only. `'reports owner exit and both streams before the row-discovery deadline'` (`:446`) calls `waitForRunBoot('the failed run row', () => undefined, owner)` at `:455`. The `read` function never resolves via any boot condition — it only exercises the owner-exit fast-fail branch of `waitFor`. Calling this a "boot" wait is a mild semantic stretch; harmless because owner-exit fires before the deadline matters either way.

**Evidence:** `packages/cli/src/commands/workflow-wait.integration.spec.ts:452-455` (the PR's own comment already flags and justifies the reuse: "Through the same helper the boot waits use, at the same deadline: the claim is that owner exit beats `RUN_BOOT_DEADLINE_MS`...").

**Required outcome:** None required. Optional: rename or add a one-line comment distinguishing "reuses the boot deadline to prove the guard" from "waits for a boot."

**Disposition:** OPEN, non-blocking. Author's call.

</details>

<details>
<summary><code>R2</code> — Boot-duration log is unconditional, permanent stdout noise</summary>

**Impact:** `waitForRunBoot` (`:433-442`) logs `[boot] <what>: <elapsed>ms of 90000ms` via plain `console.log` on every successful call, forever — not just during investigation of a slow-boot regression. Verified directly: my own test runs printed this line 3 times per green run (once per successful `waitForRunBoot` call; the 4th call site is the owner-exit-guard test, which throws before reaching the log). Not gated behind a verbose/debug flag.

**Evidence:** `packages/cli/src/commands/workflow-wait.integration.spec.ts:433-442`; observed output from `bun test src/commands/workflow-wait.integration.spec.ts`: `[boot] the gated run row: 470ms of 90000ms`, `[boot] the detached owner to start running: 0ms of 90000ms` (x2).

**Required outcome:** None required — this is the PR's explicit, stated mechanism for making a creeping boot visible in the CI transcript before it fails again, and it only fires on the success path (confirmed: it does not print on the failure path, matching the docblock's claim).

**Disposition:** OPEN, non-blocking. Working as designed; flagged as a tradeoff, not a defect.

</details>

<details>
<summary><code>R3</code> — <code>RUN_BOOT_DEADLINE_MS</code> docblock over-embeds point-in-time data</summary>

**Impact:** The docblock above `RUN_BOOT_DEADLINE_MS` (`:393-421`) embeds a full investigation's raw numbers as permanent prose above a magic number: "360-464 ms locally," "8.2 s and 41.4 s," "~16 s for all ten... locally," "~34x (676 ms... vs ~20 ms locally, #2924)," "195x the local sample." `.archon/engineering.md` says facts that change (counts, paths, inventories) do not belong in durable text — point at the owning source instead. These specific CI timings will drift as the runner, file, and test count change, and nobody is likely to come back and update a comment that will keep reading as current.

**Evidence:** `packages/cli/src/commands/workflow-wait.integration.spec.ts:393-421` (verified directly — the full docblock, confirmed present as described).

**Required outcome:** None required to merge. Suggested edit: keep the durable rationale (why 90s and not 30s or 300s — an order of magnitude above the file's own amplification precedent, so a miss reads as stuck not slow) plus a pointer to #3288 for the raw numbers; drop the embedded measurements from the comment itself.

</details>

<details>
<summary><code>R4</code> — Two of four boot-wait call sites have no owner to fail fast against</summary>

**Impact:** The safety argument for raising 30s→90s ("a launcher that dies fails fast via the owner-exit guard instead of spending the deadline") applies fully to the two call sites that pass a `ForegroundOwner` (`:455`, `:568`), but not to the two inside `launchDetached`-based tests (`:621`, `:655`), which pass no `owner` argument because `launchDetached` doesn't hand one back. If a detached launcher died before writing `running` at those two sites, the test would still correctly fail (no false pass), just after up to 90s of generic `Timed out waiting for...` polling instead of a fast, diagnosed failure — a latency regression bounded well inside each test's existing 120s Bun budget (`:645`, `:686`, both confirmed unchanged by this PR), not a correctness regression.

**Evidence:** `packages/cli/src/commands/workflow-wait.integration.spec.ts:621-623` and `:655-657` (no `owner` arg); `:645`, `:686` (`}, 120_000);` budgets, both pre-existing and untouched).

**Required outcome:** None required to merge. Worth a one-line acknowledgment in the PR description, or a short comment at those two call sites, if the safety argument is meant to hold precisely at all four sites rather than two of four.

</details>

<details>
<summary><code>R5</code> — The pattern this PR fixes is still live, unpatched, elsewhere</summary>

**Impact:** The PR's own comment states "Every caller states its own [deadline]" for `waitFor` in this file — accurate and true, scoped to this file. It makes no repo-wide claim. But the identical bug *shape* — a local `waitFor` polling helper with a silent `timeoutMs` default, shared across multiple independent "wait for one condition" call sites with nothing forcing them to move together — is confirmed still present and unfixed in `packages/core/src/services/run-live-owner.test.ts:16` (`timeoutMs = 3_000`), with 6 call sites silently relying on the default: lines 88, 112, 141, 147, 162, 240. I independently verified this by reading the file directly (not just trusting the sub-reviewer's report) — confirmed byte-for-byte. Two more files (`packages/cli/src/utils/detached-run-control.integration.spec.ts:16`, `timeoutMs = 5_000`; `packages/cli/src/commands/workflow-terminal-event.integration.spec.ts:34-37`, `timeoutMs = 15_000`) have the same helper shape; their call sites were not enumerated before the reviewing agent's checkout was torn down mid-search, so they are unexamined, not clean.

**Class:** Invariant: a polling helper with a silent shared timeout default, called from multiple independent call sites, is exactly the shape that produced #3288's drift (nothing forces the call sites to reconsider the default together). Search: `grep -rn "timeoutMs = \|timeoutMs: number = \|timeout = 30_000\|timeoutMs=30000\|= 30_000" --include="*.ts" packages`, cross-checked against `grep -n "waitFor("` in the affected files. Members confirmed violating: `packages/core/src/services/run-live-owner.test.ts` (6 call sites, independently verified). Members examined and ruled out (different class — not a polling-helper shared default): `packages/providers/src/claude/provider.ts:1546` (local variable, not a default), `packages/server/src/adapters/web/pg-notify-listener.ts:19` (reconnect backoff ceiling), `packages/server/src/scripts/generate-api-types.ts:38` (single-purpose script timeout, not multi-site). Members unexamined: `detached-run-control.integration.spec.ts`, `workflow-terminal-event.integration.spec.ts` — call sites not enumerated.

**Required outcome:** None required of this PR. Per `AGENTS.md`'s own bound on drift-fixing ("correct what the change already touches, not what it merely noticed"), `run-live-owner.test.ts` is a different package, a different file, with no evidence of an active Windows flake tied to it — this PR did not create that risk and isn't the right vehicle to fix it. Recommend a follow-up issue to extend the same fix (named deadline, no default) there, and to finish enumerating the two unexamined files.

**Found by:** `prp-core:seam-analyzer` (initial discovery and full class search); verified independently by the coordinator (direct file read of `run-live-owner.test.ts`, confirmed helper signature and all 6 call sites).

**Disposition:** OPEN as a Suggestion — real and evidenced, but not required by this PR's own outcome or invariant, and not eligible for `TRACKED FOLLOW-UP` yet since no GitHub issue exists for it (this review is advisory-only and did not create one).

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| code | R1, R2 |
| seams | R5 |
| simplify | R3 |
| tests | R4 |

#### Validation

|---|---|---|
| `bun run lint` | PASS | `check-test-cleanup-drift.ts` passed, all package globs linted clean, no errors surfaced |
| `bun --filter '*' type-check` | PASS | All 13 packages exited 0, including `@archon/cli` |
| `cd packages/cli && bun test src/commands/workflow-wait.integration.spec.ts` | PASS | 10 pass, 0 fail, 21 expect() calls; boot logs observed: 414-470ms of 90000ms, 0-1ms of 90000ms |
| `cd packages/cli && bun run test` | PASS | Full package test script (isolated per-file invocations), 0 fail across all files |
| Live mutation: remove `throw` in `waitFor`'s owner-exit guard, re-run guard test | FAIL as expected | Test failed after 175,902.74ms with `Timed out waiting for the failed run row` instead of `Owner exited before...` — confirms the guard is load-bearing |
| Live mutation: restore guard, re-run guard test | PASS as expected | 1 pass, 30ms, correct error path confirmed |

Operational note: the reviewing agent's isolated worktree was torn down externally partway through the round (unrelated to the PR's content). A fresh detached worktree pinned to the same `reviewed_head` was created to continue; all four scope reviewers independently confirmed the same head and diff before and after the interruption, and their findings above were re-delivered directly to the coordinator with full evidence after the interruption. One finding (`R5`) was independently re-verified by the coordinator via direct file read after the interruption, since the seam reviewer's own checkout was gone by the time it needed to enumerate two more files.

</details>

Wirasm:
## Review dispositions — all five findings

Reviewed head `788f2c4c0`. This round merges `origin/dev` once and adds one **comment-only** commit. No deadline moves — not `RUN_BOOT_DEADLINE_MS`, not a `}, N)` budget, not `startWait`'s `--timeout`. `waitFor`'s `timeoutMs` default stays removed, every call site still states its deadline or goes through the wrapper, and no assertion changed.

### R4 — Two boot sites have no owner to fail fast against · addressed, but not with a guard

I went looking for an owner handle to pass at `launchDetached`'s two boot waits. There is none. Three candidates, each rejected on evidence:

**1. The launcher subprocess.** It is not the run's owner. It forks the owner into its own process group, waits `DETACHED_STARTUP_WINDOW_MS` (500 ms), acks, and exits **0** — `workflow.ts:547-578`, ack at `:2343-2356`, which carries `runId`/`workflow`/`branch`/`conversationId`/`transcriptPath`/`logPath` and no pid.

Seen red rather than argued: I wrapped the launcher as a `ForegroundOwner`, passed it at the cancel site, and made the boot need one extra poll — the Windows condition this PR exists to survive. `wakes on a cancelled run stopped from a third process` then failed in **1379.89 ms** on a run that was working fine:

```
error: Owner exited before the detached owner to start running: owner pid 65992, exit 0, signal null
```

That is the fabricated handle being worse than the latency, measured. (Without the extra poll it passes, because `waitFor` reads before it checks exit — so the damage lands precisely on slow boots, which is the worst possible place for it.)

**2. The detached owner's pid.** It never leaves that process's memory except as the reply to a `stop` frame (`run-live-owner.ts:267`), and `requestRunLiveOwnerStop` (`:459`) takes an 8 s termination lease. Reading a pid by telling the run under test that a stop was requested is a side effect on the subject.

**3. The owner's control endpoint** (`canConnectToRunLiveOwner` / `watchRunLiveOwner`). `startRunLiveOwner` is called at `workflow.ts:3125`, *after* project resolution and isolation setup and immediately before the `running` transition. For nearly the whole boot window "unreachable" means "not yet", not "dead". Treating it as death would fail healthy slow boots — the exact defect #3288 is about.

**The exposure is real, and I reproduced it rather than inferring it.** With the detached site's read forced never to resolve, the case failed at **90936.34 ms** with the plain `Timed out waiting for the detached owner to start running` and no owner diagnostics — inside the 120 s budget, a timeout and not a Bun kill. Latency, never a false pass.

What the launcher *does* already cover: a child that dies inside the startup window makes it exit non-zero, `launchDetached` throws carrying its output, and the boot wait is never reached. The residual window is death **after** the ack and **before** `running`.

So the correction is the honest one. The constraint is recorded at the two helpers that own it — `launchDetached` (the process spawned here is the launcher, not the owner) and `waitForRunBoot` (why `owner` is optional, and why each alternative above is worse than the gap) — instead of an invented handle.

### R3 — Docblock embeds numbers that will rot · FIXED

Cut from `RUN_BOOT_DEADLINE_MS`'s docblock: `360–464 ms locally`, `8.2 s and 41.4 s`, `~16 s for all ten of them locally`, `~34x (676 ms … against ~20 ms locally)`, `roughly 195x the local sample`, and both prose repetitions of `120 s` (the budgets are now referred to as the `}, N)` budgets, which cannot drift from the code below them).

Kept: that this is a setup deadline and not a test budget, that #2924's rule is intact, the diagnosed original failure (`no such table: remote_agent_workflow_runs`), and the shape of the choice — clear the boots that runner completes when healthy by enough that missing it reads as *stuck* rather than *slow*, and stay under the case budgets so a stall surfaces as `Timed out waiting for …`. #3288 owns the measurements now, and the `[boot]` line keeps producing fresh ones.

Two claims in that docblock were corrected in the same pass, since they were the sentences being edited:

- It said the owner-exit guard means "a launcher that dies reports itself instead of spending the deadline." Per R4 that is true of two boots, not four. It now says which.
- It implied the headroom is an order of magnitude. Measured against the slowest healthy Windows sample on record it is closer to 2.2x — the review's own residual soft spot. The docblock now says the margin is "real but not generous" and points at the `[boot]` line, without pinning a figure that will move.

The same over-claim was live in the PR description (invariant bullet, the `Solution` paragraph, and the behavior-change table). I corrected those three phrases in the body rather than leave the code and its own PR disagreeing.

### R1 — `waitForRunBoot` reused for a non-boot wait · not taken

Cosmetic, and the reviewer rated it harmless. The call site already carries three lines explaining exactly why it borrows the boot deadline. Operator's call, and the operator declined it.

### R2 — The `[boot]` log is unconditional · deliberately not taken

Kept exactly as it is, and not gated. This is the one place where quieting the transcript would cost something real: against the worst already-observed healthy Windows sample the deadline has roughly 2.2x headroom, not the ~195x the local-boot framing suggests, and no true Windows boot time has ever been measured. The log is what turns the next Windows run into a number instead of an inference. Removing it would take away the thing that makes the raise defensible.

### R5 — Same implicit-default pattern in `run-live-owner.test.ts` · TRACKED FOLLOW-UP → #3293

Real, and out of this PR's path per `AGENTS.md`'s own bound — correct what the change already touches, not what it merely noticed. Different package, different file, no active flake tied to it, and this PR did not create it. Tracked as #3293. The two files the review left unexamined (`packages/cli/src/utils/detached-run-control.integration.spec.ts`, `packages/cli/src/commands/workflow-terminal-event.integration.spec.ts`) are noted on that issue so the enumeration is not lost.

### Validation

| Command | Result |
|---|---|
| `cd packages/cli && bun test src/commands/workflow-wait.integration.spec.ts` | 10 pass, 0 fail — boot durations 362 ms / 1 ms / 0 ms |
| `bun run lint` | PASS |
| `bun --filter '*' type-check` | PASS, exit 0 for all 13 packages |
| `bun run validate` | exit 0 |
| Mutation — launcher passed as `owner`, boot needing one extra poll | FAILS a healthy run in 1379.89 ms with `Owner exited before …, exit 0` |
| Mutation — detached boot read forced never to resolve | FAILS at 90936.34 ms with the plain timeout and no owner block |

Both mutations were reverted; the committed diff is comments only.

Wirasm:
### First Windows boot measurement

CI is green on `a04ae13d9` — all eight checks pass, including `test (windows-latest)`. The `[boot]` line kept for R2 did its job on its first run, and the PR's one open uncertainty ("the true boot time on that runner is unknown") now has a number instead of an inference.

**windows-latest**

```
[boot] the gated run row: 1811ms of 90000ms
[boot] the detached owner to start running: 926ms of 90000ms
```

**ubuntu-latest**

```
[boot] the gated run row: 715ms of 90000ms
[boot] the detached owner to start running: 205ms of 90000ms
[boot] the detached owner to start running: 205ms of 90000ms
```

(Windows shows two lines rather than three because `reports owner_lost after abrupt process death` returns early on `win32`.)

So a healthy Windows boot here is **1.8 s**, about 50x under the deadline and about 2.5x the same boot on ubuntu — nowhere near the ~30 s the failing sample in #3288 spent. That does not retire the concern: one green sample says nothing about the contended runner that produced #3288, and the 8.2 s / 41.4 s figures behind the 2.2x worst-case reading were whole-test durations, not boots. What it does establish is a baseline that future runs are measured against, which is exactly what the log was kept for. The number to watch is this one climbing.

Issue #3255, opened 2026-09-09
refactor(workflows): node-state facts write all three sinks through one function — eleven writes leave no transcript row


Wirasm:
## Problem

`persistNodeEvent` (#3043, merged as PR #3254) gave the durable sink an owner and a type-enforced policy. Node-state rows go through `persistNodeEvent(store, event: NodeStateEventInput)`; observability rows go through `createWorkflowEvent(data: ObservabilityEventInput)`, and the two input types are an exhaustive partition of `WorkflowEventType`, so neither path can take the other's events.

The other two sinks got nothing. The JSONL transcript and the in-process emitter are still authored by hand beside that call, once per site.

Measured on `dev` at `0add05881`, in `dag-executor.ts`: 35 node-state row writes, 58 emitter calls, 23 transcript writes. `executeNodeInternal` starting a node is the shape:

```ts
getLog().info({ nodeId: node.id, provider }, 'dag_node_started');
await logNodeStart(logDir, workflowRun.id, node.id, commandName ?? '<inline>');
await persistNodeEvent(deps.store, { event_type: 'node_started', step_name: stepName, data: {...} });
emitter.emit({ type: 'node_started', runId, nodeId: node.id, nodeName, provider, model, tier });
```

Three writes of one fact. The row cannot be got wrong any more. The other two still can, and are.

### Eleven node-state writes have no transcript row at all

| function | node-state writes with no transcript row |
|---|---|
| `executeWorkflowNode` | `node_completed`, `node_failed` |
| `executeFanOutWorkflowNode` | `node_completed`, `node_failed` |
| `executeComposeFanOutNode` | `node_completed`, `node_failed` |
| `executeLoopGroupNode` | `node_completed` ×2 |
| `executeWaitNode` | `node_completed` |
| `finalizeLoopFromSignal` | `node_completed` |
| `assertCheckoutUntouched` | `node_failed` |

A child `workflow:` node, either fan-out, a loop group, a durable wait and a loop finalized from a gate signal all complete without leaving a transcript row. #2614 shipped through PR #3096 on the premise that the transcript is the append-only everything-log an operator reads to reconstruct a run, and its acceptance says whatever is missing gets added to the writer rather than to a second log. For node state, this is what is missing. The gotcha has already reached operator-facing guidance: `.claude/skills/archon-cli/manage-run/troubleshooting.md:24` tells agents that some lifecycle events are not in the JSONL and to query the database instead.

[...]

Issue #2398, opened 2026-08-03
refactor(workflows): event payload naming and shape are inconsistent across sibling events


Wirasm:
Split out of #2393. Individually trivial, collectively the reason every consumer writes bespoke parsing.

## Verified

**1. `duration` vs `duration_ms`.** `loop_iteration_completed` uses a bare `duration`; every other timed event uses `duration_ms`:

```json
{"iteration":1,"duration":8013,"completionDetected":true,"nodeId":"validate-like"}
```

**2. `usage` vs `tokens` — and they are not the same data.**

| event | key | shape |
| --- | --- | --- |
| `node_completed` | `tokens` | `{"input":386956,"output":4212}` |
| `task_activity` | `usage` | `{"total_tokens":156903,"tool_uses":39,"duration_ms":206215}` |

So a consumer cannot treat them as the same concept even after renaming: one splits input/output, the other reports a total plus tool_uses. Whichever name wins, the *shapes* need reconciling too.

**3. `node_failed` carries no timing.**

```json
{"error":"Bash node 'persist-assessment' failed [exit 1]: ...","type":"bash"}
```

`node_completed` carries `duration_ms`; `node_failed` does not. A consumer wanting "how long before it failed" must join back to `node_started` — and `buildNodeSummaries` does exactly that, which is why the CLI shows a duration the event itself lacks.

**4. Case convention is mixed.** `workflow_started.workflowName` is camelCase; node events use snake_case (`tool_name`, `duration_ms`, `node_output`). `loop_iteration_completed` mixes both in one payload (`nodeId` alongside `iteration`).

## Why now

These are the only *breaking* changes in the #2393 direction — everything else there is additive. They are cheapest to take while the in-repo console normalizer and helm are the only consumers. #2315 exists to declare this contract before there are clients we do not control; helm is about to become one.

Note the console normalizer already reads persisted DB names that differ from the in-process emitter names — a known live source of bugs — so any rename must update both sides together.

## Decide first

#2393's open question 1: version the payload, or take the break now? This issue is the concrete cost of choosing "later".

Wirasm:
## Unpark: the renames are safe, but for a narrower reason than this issue records

Measured on `dev` @ `0ea658bf8`.

The migration note here says `workflow_events.data` is JSON-as-TEXT so every field is additive, and that the only breaking change in the #2393 direction is renaming the inconsistencies. That is correct about the **column** and incomplete about the **readers**.

[...]

PR #3292, opened 2026-09-10
fix(workflows): a wait's completion rows have one owner on both paths

Wirasm:
## Problem and outcome

A wait node completes on two paths: in the same tick inside the executor, or on resume inside the store's cursor-clearing transaction. Each path spelled the wait outcome row and the node's `node_completed` row by hand, so the two could drift and the same logical event would leave different rows depending on which path ran. #3275 removed one such pair on the resumed path; this is the other one.

- **Outcome:** both rows come from one function, `waitCompletionEvents` in the workflows store module. Core inserts what it returns, the executor's same-tick branch records what it returns, and the test fakes use it too. A guard proves no other file in either package spells the row.
- **Invariant:** persisted `data` payloads are byte-identical to before; the store contract's return shape is unchanged; the resume wire format is untouched.
- **Scope boundary:** no other event row gains an owner here. The observability and node-state event vocabularies stay as they are (#2393, #2398).
- **Root cause:** the row shape had no owner. `WorkflowWaitCompletion` typed the store's input, but the rows derived from it were untyped records written out at each site.

## Review guidance

- **Feedback requested:** whether the owner belongs in `store.ts` beside `WorkflowWaitCompletion`, which is where both packages already import the wait types from.
- **Start here:** `packages/workflows/src/store.ts` — `waitCompletionEvents`, the only place that knows the two row shapes.
- **Review order:** `store.ts` (owner) → `packages/core/src/db/workflows.ts` (`clearWorkflowWaitContext` inserts and returns the owner's rows) → `packages/workflows/src/dag-executor.ts` (`executeWaitNode`'s same-tick branch) → the two fakes in `dag-executor.test.ts` and `subrun.test.ts` → `wait-completion-events.test.ts`.
- **Lower-attention areas:** the fakes, which now call the owner instead of restating the literal; covered by their own suites.

## Solution

`waitCompletionEvents(workflowRunId, completion)` returns `{ outcome, node }`: the `wait_completed` or `wait_expired` observability row and the node's `node_completed` row, both built from one `WorkflowWaitCompletion`. `clearWorkflowWaitContext` inserts both inside its transaction and hands back `node`, as #3275 established. The executor's same-tick branch builds the same pair and writes them through `createWorkflowEvent` and `recordNodeState`. The guard test scans every non-test source in the workflows package plus core's store file for a second spelling of the row and expects exactly one file.

## Validation

- `bun run type-check` — clean across workspaces.
- `bun test src/wait-completion-events.test.ts` — 3 pass — proves both statuses derive the right rows and that only `store.ts` spells the row. Seen red first: with a second spelling added in a probe file the guard failed, then passed again once the probe was removed.
- `bun test src/dag-executor.test.ts` — 785 pass; `bun test src/subrun.test.ts` — 101 pass.
- `bun test src/db/workflows.test.ts` — 130 pass; `bun test src/db/workflows.resume-cas.integration.test.ts` — 54 pass, in `packages/core`.
- `bun run lint` — clean.
- **Not verified:** nothing material. Both paths and both storage engines' row shapes are covered by the suites above.

## Links

- Related #3275, #3255

https://claude.ai/code/session_013WWN4GE8VDRPBXbNPBStqf

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of workflow wait completion records for both completed and expired waits.
  * Ensured wait nodes consistently retain result and duration details across execution paths.

* **Tests**
  * Added coverage for completed and expired wait outcomes, node completion records, and associated results.
  * Updated workflow execution tests to validate consistent wait completion records.

### DIFF
--- a/packages/core/src/db/workflows.ts
+++ b/packages/core/src/db/workflows.ts
@@ -38,7 +38,7 @@ import type {
   WorkflowWaitPause,
   NodeStateEventInput,
 } from '@archon/workflows/store';
-import { FAN_OUT_CANCEL_REASONS } from '@archon/workflows/store';
+import { FAN_OUT_CANCEL_REASONS, waitCompletionEvents } from '@archon/workflows/store';
 
 /** Best-effort ROLLBACK — log but swallow errors since we're already in an error path. */
 function rollback(): Promise<void> {
@@ -1547,25 +1547,10 @@ export async function clearWorkflowWaitContext(
         [id, waitContext.nodeId, cursor]
       );
       if ((result.rowCount ?? 0) === 0) return { cleared: false };
-      await insertWorkflowEvent(query, {
-        workflow_run_id: id,
-        event_type: completion.result.status === 'expired' ? 'wait_expired' : 'wait_completed',
-        step_name: completion.stepName,
-        data: completion.result,
-      });
-      const nodeEvent: NodeStateEventInput = {
-        workflow_run_id: id,
-        event_type: 'node_completed',
-        step_name: completion.stepName,
-        data: {
-          type: 'wait',
-          duration_ms: completion.result.waited_ms,
-          node_output: JSON.stringify(completion.result),
-          structured_output: completion.result,
-        },
-      };
-      await insertWorkflowEvent(query, nodeEvent);
-      return { cleared: true, nodeEvent };
+      const rows = waitCompletionEvents(id, completion);
+      await insertWorkflowEvent(query, rows.outcome);
+      await insertWorkflowEvent(query, rows.node);
+      return { cleared: true, nodeEvent: rows.node };
     });
   } catch (error) {
     const err = error as Error;
--- a/packages/workflows/package.json
+++ b/packages/workflows/package.json
@@ -64,7 +64,8 @@
     [
       "src/logger.test.ts",
       "src/terminal-status-write.test.ts",
-      "src/node-event-write.test.ts"
+      "src/node-event-write.test.ts",
+      "src/wait-completion-events.test.ts"
     ],
     [
       "src/condition-evaluator.test.ts"
--- a/packages/workflows/src/dag-executor.test.ts
+++ b/packages/workflows/src/dag-executor.test.ts
@@ -145,6 +145,7 @@ import {
 import { OutputRefError } from './output-ref';
 import type { WorkflowDeps, IWorkflowPlatform, WorkflowConfig } from './deps';
 import type { IWorkflowStore, PersistedNodeOutput } from './store';
+import { waitCompletionEvents } from './store';
 import {
   buildInstanceSnapshots,
   composeFanOutScopeSegment,
@@ -247,12 +248,7 @@ function createMockStore(): MockWorkflowStore {
     clearWorkflowWaitContext: mock<IWorkflowStore['clearWorkflowWaitContext']>(
       async (id, _waitContext, completion) => ({
         cleared: true,
-        nodeEvent: {
-          workflow_run_id: id,
-          event_type: 'node_completed',
-          step_name: completion.stepName,
-          data: { type: 'wait', duration_ms: completion.result.waited_ms },
-        },
+        nodeEvent: waitCompletionEvents(id, completion).node,
       })
     ),
     rewriteApprovalContext: mock<IWorkflowStore['rewriteApprovalContext']>(
--- a/packages/workflows/src/dag-executor.ts
+++ b/packages/workflows/src/dag-executor.ts
@@ -101,7 +101,7 @@ import type { BindingDirective } from './schemas';
 import { mapNodeTemplateSlots } from './template-walker';
 import { buildExecNodeEnvironment } from './exec-environment';
 import { planGraph, resolvedBodyNodes } from './graph-plan';
-import { FAN_OUT_CANCEL_REASONS } from './store';
+import { FAN_OUT_CANCEL_REASONS, waitCompletionEvents } from './store';
 import type { DagResumeSnapshot, FanOutCancelReason, PersistedNodeOutput } from './store';
 import { formatToolCall } from './utils/tool-formatter';
 import { createLogger, captureWorkflowCompleted } from '@archon/paths';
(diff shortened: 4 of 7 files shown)


Wirasm:
## Ready to merge

Round 1 found one Important gap: the guard test that's supposed to prove `waitCompletionEvents` is the wait row's only owner used a non-recursive directory scan, so it never looked inside `packages/workflows/src/schemas/`, `utils/`, or `defaults/`. The correction at this head fixes exactly that, in a single seven-line change to `packages/workflows/src/wait-completion-events.test.ts`: the scan now passes `{ recursive: true }` to `readdir`, and the match pattern was tightened from a bare `type: 'wait',` to `/type:\s*'wait',\s*duration_ms\s*:/` so it keys on the row's first two fields together rather than the `type` literal alone — which also fixes the near-miss the loosened pattern would otherwise have raised against `schemas/dag-node.ts`'s unrelated `{ type: 'wait', fields: ... }` node-kind literal.

**Validation:** all specified checks pass. `bun run type-check` clean; `wait-completion-events.test.ts` 3/3 pass at rest. I re-ran the subdirectory probe from round 1 and extended it as requested: injected a multi-line spelling of the row into `schemas/durable-wait.ts` — guard failed red, correctly named the injected file, restored, tree confirmed clean; injected an inline spelling into `utils/duration.ts` — guard failed red again, correctly named the injected file, restored, tree confirmed clean. Separately confirmed the existing unrelated literal in `schemas/dag-node.ts:2043` still does not trip the tightened regex, and confirmed by direct `readdir(..., { recursive: true })` enumeration that all three previously-unscanned subdirectories (`schemas/`, `utils/`, `defaults/`) are now structurally reached (72 non-test `.ts` files found vs. 44 before the correction). The correction diff touches only this one test file, so `dag-executor.test.ts`, `subrun.test.ts`, and core's `db/workflows.test.ts` are unaffected by this round and their round-1 results carry forward unchanged.

**Resolved:** 1

### Findings

|---|---|---|---|
| `R1` | Important | Guard test's non-recursive directory scan missed `schemas/`, `utils/`, `defaults/`, so a second spelling of the wait row placed there would have passed silently | FIXED |

<details>
<summary><code>R1</code> — Guard test didn't scan workflows-package subdirectories, so it couldn't back its own claim</summary>

**Impact:** `packages/workflows/src/wait-completion-events.test.ts` claimed to prove `store.ts` is "the only place that spells the wait node row, in either package," but its scan only reached the flat top level of `packages/workflows/src`. A second spelling placed in `schemas/`, `utils/`, or `defaults/` would have passed the guard silently.

**Evidence (round 1):** guard test confirmed to stay green (3/3 pass) after a duplicate row was injected into `schemas/dag-node.ts`, an unscanned subdirectory.

**Correction (this head, `27200aaf5`):** `packages/workflows/src/wait-completion-events.test.ts` now calls `readdir(here, { recursive: true })` and matches `/type:\s*'wait',\s*duration_ms\s*:/` instead of the bare `type: 'wait',` literal.

**Verification (this round):**
- Re-ran the round-1 baseline: type-check clean, guard 3/3 pass at rest.
- Injected a multi-line spelling (`type: 'wait',` then `duration_ms: 1,` on separate lines) into `schemas/durable-wait.ts` — guard failed (2 pass, 1 fail), correctly reported `schemas/durable-wait.ts` as the extra spelling. Restored the file; `git status --short` in the worktree came back empty.
- Injected an inline spelling (`{ type: 'wait', duration_ms: 1 }` on one line) into `utils/duration.ts` — guard failed (2 pass, 1 fail), correctly reported `utils/duration.ts`. Restored the file; worktree confirmed clean again.
- Confirmed the pre-existing unrelated literal at `schemas/dag-node.ts:2043` (`{ type: 'wait', fields: WAIT_NODE_IGNORED_FIELDS }`) still does not match the tightened regex (checked directly against the pattern: no match, since `fields:` follows rather than `duration_ms:`), so the correction closes the gap without introducing a new false positive.
- Confirmed by direct `readdir(here, { recursive: true })` enumeration that the scan now structurally reaches all three previously-blind subdirectories (`schemas/`, `utils/`, `defaults/`): 72 non-test `.ts` files found, versus 44 under the round-1 flat scan.

**Disposition:** FIXED — the changed head proves the correction. The guard now covers the subtree its own claim describes, verified by reproducing red on both requested subdirectories in both requested spelling styles, then confirming clean restoration and no regression to the round-1 false-positive exclusion.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 (round 1, no re-dispatch this round — correction verified directly against the prior finding by reproduction; diff confined to one test file with no new surface for `code`, `seams`, or `simplify` to review) |
| `seams` | R1 |
| `simplify` | R1 |

#### Validation

|---|---|---|
| `bun run type-check` (packages/workflows) | PASS | clean, no errors |
| `bun test src/wait-completion-events.test.ts` (packages/workflows), at rest | PASS | 3 pass, 0 fail, 6 expect() calls |
| Guard probe: multi-line spelling in `schemas/durable-wait.ts` | RED as expected | 2 pass, 1 fail; guard named the injected file; restored, worktree clean |
| Guard probe: inline spelling in `utils/duration.ts` | RED as expected | 2 pass, 1 fail; guard named the injected file; restored, worktree clean |
| False-positive check: `schemas/dag-node.ts:2043` against tightened regex | PASS (no match) | confirms correction doesn't reintroduce the near-miss |
| Recursive-scan structural check | PASS | `readdir(here, { recursive: true })` reaches `schemas/`, `utils/`, `defaults/`; 72 files vs. 44 pre-correction |
| `bun test src/dag-executor.test.ts` (packages/workflows) | carried forward from round 1 (785 pass, 0 fail) | correction diff does not touch this file or its dependencies |
| `bun test src/subrun.test.ts` (packages/workflows) | carried forward from round 1 (101 pass, 0 fail) | correction diff does not touch this file or its dependencies |
| `bun test src/db/workflows.test.ts` (packages/core) | carried forward from round 1 (130 pass, 0 fail) | correction diff does not touch this file or its dependencies |

</details>

Wirasm:
R1 fixed in `27200aaf5`. The guard now walks the whole package tree and recognises the row by its first two keys together (`type: 'wait'` followed by `duration_ms`), since `type: 'wait'` alone also names the node kind in the ignored-fields table under schemas/. Seen red twice before committing: a multi-line probe under schemas/ and an inline one-line probe under utils/ each failed the guard, naming the file; the real tree passes.

PR #3300, opened 2026-09-11
fix(core): REST reject route no longer records empty text for an explicit empty-string reason

### LINKED ISSUE #2742: bug(workflows): REST reject route still records an empty structured_output.text for an explicit empty-string reason
## Problem

Issue #2740 / PR #2739 / PR #2741 fixed four `rejectWorkflow` call sites (`manage-run-tool.ts`, `command-handler.ts`, `workflow.ts` x2, `workflow-bridge.ts`) so an omitted (`undefined`) rejection reason defaults to `'Rejected'` instead of reaching `structured_output.text` as `''`. #2740's own problem statement asserts `api.ts`'s REST reject route already has "the same pattern applied" — but that claim is only true for `undefined`/`null`, not for an explicit empty string.

`packages/server/src/routes/api.ts:3831`:
```ts
const reason = body.reason ?? 'Rejected';
```
`??` only triggers on `null`/`undefined`. `rejectWorkflowRunBodySchema` (`packages/server/src/routes/schemas/workflow.schemas.ts:176`, `z.object({ reason: z.string().optional() })`) has no `.min(1)`, so a `POST /api/workflows/runs/:runId/reject` body of `{"reason": ""}` passes validation and `reason` stays `''`.

That flows into `rejectWorkflow(runId, '')`. The root cause is one level deeper, in `packages/core/src/operations/workflow-operations.ts`:
- Line 644: `const rejectReason = reason ?? 'Rejected';` — same `??` gap; `rejectReason` stays `''` for an explicit empty-string input.
- Line ~696 (the #2707 new-mode structured-output path): `const structuredOutput = { decision: 'reject', text: reason ?? '' };` — uses the **raw** `reason` parameter, not the already-computed `rejectReason`, so even a caller that reaches the correct-audit-event path (`reason: rejectReason` at line 680) still gets `structured_output.text: ''`.

## Consequence

On a gate authored with `approval.decisions:` (`decisionsAuthored: true`, the #2707 new mode), a direct API caller — curl, an integration, a future UI — sending `{"reason": ""}` to `POST /api/workflows/runs/:runId/reject` records `structured_output.text: ''`, reproducing the exact defect #2740/#2739/#2741 fixed for the CLI/Slack/manage_run/command-handler call sites. The shipped console `ApprovalPanel.tsx` blocks this client-side (requires non-empty text before submitting), so it isn't reachable from the shipped web UI today — but it is live and completely untested via direct API use.

No test in `workflow-operations.test.ts` or `api.workflow-runs.test.ts` exercises `reject` with an explicit empty-string reason. `api.workflow-runs.test.ts` also never exercises the REST route's new-mode (`decisionsAuthored`) reject path at all — the `decisionsAuthored` combination is tested only for `/approve` (around line 1907), not `/reject`.

## Suggested fix

A single change in `workflow-operations.ts`, no signature change required:
1. Change line 644 to also treat an empty string as unset, matching the CLI's own guard shape: `const rejectReason = reason && reason.length > 0 ? reason : 'Rejected';`
2. Reuse `rejectReason` (not raw `reason`) when building `structuredOutput.text` at line ~696.

That closes the gap for `api.ts` and every future caller in one place, makes the four hand-copied call-site guards (`manage-run-tool.ts`, `command-handler.ts`, `workflow.ts` x2) redundant defense rather than load-bearing, and doesn't touch `rejectWorkflow`'s public signature or the four existing `workflow-operations.test.ts` call sites that call `rejectWorkflow('run-1')` with no second argument (they'd still pass — `undefined` still defaults to `'Rejected'`).

Add a regression test for `POST /api/workflows/runs/:runId/reject` with `{"reason": ""}` on a `decisionsAuthored` gate, asserting `structured_output.text === 'Rejected'`.

## Context

Found during review of PR #2741 (https://github.com/coleam00/Archon/pull/2741), which closed #2740. Same defect class as #2734/#2740, one level deeper: those fixed the *omitted*-reason case at four call sites; this is the *explicit-empty-string* case, reachable only through `api.ts` today because it's the one call site that uses `??` instead of a length check.

Wirasm:
The `/reject` and `/respond` REST routes pass an explicit empty-string reason straight through to `rejectWorkflow`. The nullish coalescing operator (`??`) only guards against `null`/`undefined`, so `"reason": ""` in the request body lands in both the audit event and the new-mode `structured_output.text` as-is — producing an empty string where the caller intended the default "Rejected".

The fix is at the operation layer so it covers every entry point:

- **`rejectReason`** now treats empty string the same as unset (`reason && reason.length > 0 ? reason : "Rejected"` instead of `reason ?? "Rejected"`).
- **`structured_output.text`** reuses the computed `rejectReason` instead of the raw `reason` parameter, so both the audit trail and the structured output always agree.

The four non-REST call sites (CLI, manage-run-tool, command-handler, Slack bridge) were already guarded with length checks or `||` defaults, so this only affects the REST path — but landing the guard at the operation layer closes the gap for any future caller too.

Closes #2742

## Summary by CodeRabbit

- **Bug Fixes**
  - Rejection actions with an empty reason now consistently display and record **“Rejected”**.
  - Updated rejection details appear in both workflow results and approval history.
  - Workflows remain resumable after this type of rejection.

### DIFF
--- a/packages/core/src/operations/workflow-operations.test.ts
+++ b/packages/core/src/operations/workflow-operations.test.ts
@@ -889,6 +889,36 @@ describe('rejectWorkflow', () => {
     expect(mockCaptureApprovalResolved).toHaveBeenCalledWith({ resolution: 'rejected' });
   });
 
+  test('explicit empty-string reason defaults to "Rejected" in structured_output and audit event', async () => {
+    const run = makePausedRun({
+      metadata: {
+        approval: {
+          nodeId: 'review',
+          message: 'Please review',
+          type: 'approval',
+          decisions: [{ id: 'approve' }, { id: 'reject' }],
+          decisionsAuthored: true,
+        },
+      },
+    });
+    mockGetWorkflowRun.mockResolvedValueOnce(run);
+
+    const result = await rejectWorkflow('run-1', '');
+
+    expect(result.newMode).toBe(true);
+    expect(result.cancelled).toBe(false);
+    expect(mockResolveApprovalGate).toHaveBeenCalled();
+    const events = mockResolveApprovalGate.mock.calls[0][2] as unknown[] as Array<
+      Record<string, unknown>
+    >;
+    const nodeCompleted = events.find(e => e.event_type === 'node_completed');
+    expect(nodeCompleted?.data).toMatchObject({
+      structured_output: { decision: 'reject', text: 'Rejected' },
+    });
+    const approvalReceived = events.find(e => e.event_type === 'approval_received');
+    expect(approvalReceived?.data).toMatchObject({ reason: 'Rejected' });
+  });
+
   test('rejects an escalated body-terminal-gate pause — node_completed lands under the namespaced <nodeId>.<bodyGateId> step_name (#2707 step 3)', async () => {
     const run = makePausedRun({
       metadata: {
--- a/packages/core/src/operations/workflow-operations.ts
+++ b/packages/core/src/operations/workflow-operations.ts
@@ -780,7 +780,7 @@ export async function rejectWorkflow(
     }
   }
 
-  const rejectReason = reason ?? 'Rejected';
+  const rejectReason = reason && reason.length > 0 ? reason : 'Rejected';
   const currentCount = (run.metadata.rejection_count as number | undefined) ?? 0;
   const maxAttempts = approval?.onRejectMaxAttempts ?? 3;
   // `!= null` (not `!== undefined`): "no on_reject" reaches this read in two stored
@@ -833,7 +833,7 @@ export async function rejectWorkflow(
   // losing the audit trail (#2146).
   let won: boolean;
   if (willResolveNewMode && approval) {
-    const structuredOutput = { decision: 'reject', text: reason ?? '' };
+    const structuredOutput = { decision: 'reject', text: rejectReason };
     const nodeCompletedEvent: workflowDb.GateResolutionEvent = {
       event_type: 'node_completed',
       step_name: resolvedNodeCompletedStepName(approval),


Wirasm:
# Review Report — PR #3300 (issue-2742-deepseek)


## 1. Verdict

**Ready: true** · **Action: none**

No open Critical or Important findings. The change does exactly what the accepted contract requires; both required invariants are proved by the regression test and verified against the code. Every enabled lens reported; errors and docs were disabled by the review-scope classification (no error surface or shipped-documentation surface in the diff).

---

## 2. Accepted contract

**Required outcome (issue #2742):** `POST /api/workflows/runs/:runId/reject` with `{"reason": ""}` on a `decisionsAuthored` gate records `structured_output.text === 'Rejected'` (not `''`). Same for `/respond` with `{decision: "reject", text: ""}` (indirectly, via the operation layer).

**Required shape:**

- `rejectReason` treats empty string as unset: `reason && reason.length > 0 ? reason : 'Rejected'` (replaces `reason ?? 'Rejected'`).
- `structured_output.text` reuses `rejectReason` instead of the raw `reason` argument.
- Regression test proving `rejectWorkflow('run-1', '')` on a `decisionsAuthored` gate yields `structured_output.text === 'Rejected'` and audit `reason === 'Rejected'`, covering both `node_completed` and `approval_received` events.

**Non-goals / boundaries:**

- No signature change to `rejectWorkflow`; `rejectWorkflow('run-1')` (undefined arg) still gets `'Rejected'`.
- Four non-REST call sites (CLI, manage-run-tool, command-handler, Slack bridge) left unchanged; their guards are redundant defense.
- `/respond` route fixed only indirectly; no separate `/respond` route edit.
- Custom declared-decision path (`respondToWorkflowWithDeclaredDecision`) out of scope.

---

## 3. Reviewed head SHA

`bd66813436cc746bb76c2bed6b30e299c0b7892f` — one commit on `fix/issue-2742-deepseek`, base `dev`. Matches local HEAD of the worktree and the PR headRefOid. Clean working tree.

---

## 4. Findings

### Critical — 0

None.

### Important — 0

None.

### Suggestions — 1

#### R1 · Suggestion · sources: [simplify]

**Regression test erases the already-typed gate events with a double cast.**

`packages/core/src/operations/workflow-operations.test.ts:911-913`:

```ts
const events = mockResolveApprovalGate.mock.calls[0][2] as unknown[] as Array<
  Record<string, unknown>
>;
```

`mockResolveApprovalGate` is `mock<typeof WorkflowDb.resolveApprovalGate>`, whose third parameter is `GateResolutionEvent[]` — events arrive already typed. The double cast erases that type and re-declares a weaker `Record<string, unknown>` version. The sibling pattern already used in this file (e.g., test line 244) avoids the cast:

```ts
const casEvents = mockResolveApprovalGate.mock.calls[0]?.[2] ?? [];
const nodeCompleted = casEvents.find(e => e.event_type === 'node_completed');
```

`nodeCompleted?.data` is already `Record<string, unknown>` per `GateResolutionEvent`, so the existing `toMatchObject` assertions (lines 915, 919) compile unchanged. Replacing the double cast with the sibling pattern drops the `unknown[]` intermediary, preserves type precision, and removes the habit risk for copied test code. No runtime delta; `?.[2] ?? []` also eliminates the bare index while `toHaveBeenCalled()` (line 910) guards against no-call.

Why Suggestion not Important: the erased type is not read anywhere it would matter inside the test. The concrete cost today is noise. Would raise to Important if `GateResolutionEvent` became a discriminated union with per-event payload types — the cast would mask the narrowing the sibling tests get for free.

---

### Rejected findings

None. Every lens finding was verified and accepted as reported (or classified as adjacent discovery).

---

## 5. Prior findings

Not applicable — first review round (full mode).

---

## 6. Discoveries

**3 accepted adjacency records**, all validated against current code, all **adjacent** (outside the accepted contract), none a scope conflict.

| # | Title | Sources | Relation |
|---|-------|---------|----------|
| 1 | Stale CLI guard comment — claims operation-layer default doesn't cover new-mode structured path, now false | seams, tests | adjacent |
| 2 | REST route defaults redundantly re-declare the operation-layer 'Rejected' default | seams | adjacent |
| 3 | Declared-decision path still writes '' to structured_output.text while audit comment falls back to decision id | seams | adjacent |

- **Consolidated records (JSON):** `discoveries.json`
- **Human-readable summary with evidence:** `discoveries.md`

---

## 7. Review coverage

**Full review, first round.** All four enabled lenses reported and their content was verified against the code.

| Lens | Status | Output |
|------|--------|--------|
| code | Ran — 0 findings | Verified the change is correct, the `rejectReason` expression covers the full typed domain, `structured_output.text` reuses the same value, all callers pass defined strings, all downstream consumers in the dag-executor, terminal-record, and web client handle the value correctly. Also verified `respondToWorkflow` delegates `decision === 'reject'` to `rejectWorkflow` (line 1007), closing the `/respond` route gap. |
| seams | Ran — 0 findings, 3 adjacent discoveries | Verified every counterpart derives from the single `rejectReason`: audit (line 819), structured_output (line 836), stage-rework (line 856). Exhaustive gate-type dispatch with `never` guard. REST zod schemas and route body parsing. Web consumers key off `decision`, not absent field. All four non-REST call sites verified as defense-in-depth. |
| simplify | Ran — 1 Suggestion | Verified `rejectReason` variable is the single source of truth feeding three consumers — inlining would repeat the conditional. The empty-string guard shape matches the file's established idiom (`trim().length > 0` siblings). Operation-layer placement is the optimal choke point. Only Suggestion: the test's double cast is unnecessary. |
| tests | Ran — 0 findings, 1 adjacent discovery | Verified the new test asserts all four invariants (newMode, cancelled, structured_output, audit reason). Pre-existing test covers the truthy-reason branch. Falsification revert confirmed: reverting the two changed lines produces the wrong output. Route-level protection verified at api.ts:3978, 4076. All four call-site guards verified. |
| errors | **Disabled** — classified off by `deliver__classify` (reasons: the diff is a two-line default-value correction with no new catch, fallback, retry, or recovery path where a failure could become silent). |
| docs | **Disabled** — classified off by `deliver__classify` (reasons: the diff touches only `workflow-operations.ts` and its test file; no shipped documentation, README, or docs-site content is changed). No docs page describes reject-reason empty-string REST semantics. |

**Evidence the reviewer could not inspect:** none. Every lens finding was verified against the current code at the reviewed SHA. All critical file:line citations were checked. The falsification (revert → test fail) was performed by the tests lens and accepted on the basis of direct assertion logic (the test assertions directly match the changed expressions, so the causal chain is unambiguous). Test suite passed 75/75 on `workflow-operations.test.ts`; lint passed clean.

Issue #3294, opened 2026-09-10
test(ci): the windows suite has no headroom — 12 tests sit within a second of the 5000ms budget


Wirasm:
## Problem

The Windows test suite runs with Bun's default **5000 ms** per-test budget, and `dev`'s own Windows leg carries **12 tests over 4 seconds** on a normal, fully green run.

None of them times out. They do not need to. The margin is under a second, so any additional IO in that job — or any ordinary runner contention — tips one of them over, and which one is a coin flip.

## The measurement

From PR #3291, comparing its Windows leg against `dev`'s Windows leg an hour earlier, **2612 tests matched by name**:

- Median ratio branch/dev: **0.87**. The branch was *faster* overall, 93 s against 149 s. There was no broad slowdown.
- **Exactly one test per run** exceeded the budget, and it was a **different test each run** — 27x and 23x their own normal times.
- Control: **six recent `dev` Windows legs produced zero 5000 ms timeouts. The branch produced one in each of two runs. 0/6 against 2/2.**

The two victims were `captureWorkflowSource > replaces a stale capture rather than merging two vintages` (9625 ms, against 6.20 ms on ubuntu in the same run and 351.86 ms on dev's own Windows leg an hour earlier) and `generate-bundled-defaults: untracked-file guard (#1578) > exits 1 and leaves the bundle untouched for an untracked workflow default` (7995 ms).

Both spawn subprocesses and copy trees inside `%TEMP%`. For the `generate-bundled-defaults` group across 7 shared tests: dev median 355 ms / max 1843 ms; branch run 1 median 548 ms / max 1622 ms; branch run 2 median 485 ms / max **7996 ms**. Slightly slower median, dramatically fatter tail.

## Why this is worth solving

**This is the thing behind most of today's noise.** Seven distinct tests blew the Windows budget on a single day, each on a PR that did not touch the failing package. Six pull requests were red for reasons already fixed on `dev`. Every one of those cost someone a diagnosis, and the cumulative effect is that a red Windows leg stopped carrying information — people learn to shrug, and then a real failure gets shrugged at too.

The individual failures have been chased one at a time all day: #3286, #3287, #3288 and the skips in #3280 are each a symptom of a suite that has no headroom. Fixing them one by one does not change the fact that the next PR to add a little IO tips over a different test.

## Why now

[...]

Issue #3287, opened 2026-09-10
test(workflows): exec-code isolation tests blow the windows budget on unrelated PRs


Wirasm:
## Problem

The `runFixtures exec-code isolation (#2851)` family in `packages/workflows/src/fixture-runner.test.ts` is the slowest thing in that file and it blows the 5000 ms Windows budget on unrelated PRs.

Most recent sample: PR #3279's Windows leg, `gives clean and pre-dirtied callers the same verdict (#2851)` at **5231.13 ms**. That PR touches no source file in `packages/workflows` — only manifests, two `scripts/` files and a doc.

## It is not a heavy test

Measured on macOS, it runs in **141 ms**. Windows hit 5231 ms: **37x**, at the very top of the 9.4–38.7x amplification band measured for this repository.

It is also already optimised in the obvious way — `guardRepo` is built once in `beforeAll`, not per test.

## Where the time goes

Subprocess spawns. The exec-code isolation path runs three git subprocesses per `runFixtures` call:

- `git rev-parse --show-toplevel` (`fixture-runner.ts:399`)
- `git worktree add --detach` (`:408`)
- `git worktree remove --force` (`:430`)

This test calls `runFixtures` **twice**, so roughly six git spawns plus a bash spawn for the guard body plus a recursive copy of the fixture repo.

The pattern holds across the family. The seven slowest tests in that 53-test file are all `#2851 exec-code isolation` tests, and they all spawn real git and bash:

```
0.141s  gives clean and pre-dirtied callers the same verdict (#2851)
0.129s  shares one capture across every fixture in an invocation (#2851)
0.101s  captures the working tree, so an uncommitted script edit is what runs (#2851)
0.089s  resolves a nested pack by name, workflow folder, and pack directory path
0.079s  disposes both the scratch worktree and the source capture (#2851)
0.073s  executes the captured copy of a named script, not the caller tree file (#2851)
0.065s  keeps PWD and OLDPWD-based inline code writes in the scratch worktree (#2890)
```

That is #2306's thesis exactly: Windows timeouts cluster on subprocess-spawning tests under `bun --filter --parallel`.

## Why this needs a Windows machine

The numbers above are macOS. 141 ms is not a problem anywhere; 5231 ms only happens on Windows, and only under load. What is unknown is which of the six spawns costs what there, and whether it is process creation, git itself, or the worktree operations touching the filesystem.

@coleam00 — assigning to you because you are on Windows. CI gives pass or fail, not a profile.

[...]

Issue #3282, opened 2026-09-10
perf(workflows): measure the per-capture write cost on Windows


Wirasm:
## Problem

`captureWorkflowSource` writes the whole bundled scope into every child-run capture. On current `dev` that is **203 files** (`.archon/workflows` 157 + `.archon/commands` 46) per child spawn, and the count grows with every workflow added.

#3277 removed the *read* half of this: filesystem ops per spawn went 836 → 591, `readFile` 201 → 1, bytes read 1,247,567 → 60, median wall clock 51.4 ms → 33.7 ms. The **write** half is untouched, and writes are what Windows charges for, because Defender interposes on file creation.

## Why this is worth solving

The Windows budget-edge class in #2924 has cost real merge throughput. Six distinct tests blew the 5000 ms budget in a single day, each on a PR that did not touch the failing package. Five separate test files now carry `mock.module('@archon/paths', …)` workarounds whose only purpose is to make this cost disappear, plus the `--dry-run` change in #2922. That is one application cost paid around in six places.

## Why now, and why this issue is measurement rather than a fix

We cannot pick a fix, because every number we have is from the wrong platform.

The obvious candidate — hard-linking instead of copying — was measured and **rejected on macOS**: 200 files at 6 KB, median of 6 runs, `writeFile` 20.5 ms, `copyFile` 24.8 ms, `link` 62.5 ms. Link is 3x worse on APFS.

But a hard link creates a directory entry with no new content, which is plausibly the one thing Defender does *not* scan. So the lever that measures worst locally may be the best one on the platform that actually has the problem. Acting on the macOS numbers would be acting on evidence from a machine that does not exhibit the bug.

#2924's own "remaining uncertainty" section says the same thing: what would settle it is a Windows leg carrying per-phase spans.

## Desired outcome

Numbers from a real Windows machine that say where the time goes in a child-spawn capture, and specifically whether per-file **writes** dominate.

Enough to answer:

1. What fraction of capture wall clock is `copyFile` versus `digest` versus everything else?
2. Does per-operation cost scale with file *count* or with total *bytes*? Defender-style interposition predicts count; ordinary I/O predicts bytes.
3. Does a hard link avoid the per-file cost that a copy pays? This is the decisive one, and it inverts the macOS result if true.
4. How much does a concurrent neighbouring suite change all of the above? On macOS the identical 766-op replica went 33.6 ms idle to 395.2 ms under load, byte-identical op counts — a 12x swing from contention alone.

## Invariants

[...]

PR #3301, opened 2026-09-11
feat(workflows): a bundle index decides which packs ship

### LINKED ISSUE #3283: feat(workflows): a bundle index decides which packs ship
## Problem

A source build and a binary build disagree about what "the bundled scope" is, and neither of them is choosing it deliberately.

The generator, `scripts/generate-bundled-defaults.ts`, selects by **pattern** — its docblock names `.archon/workflows/defaults/`, `.archon/workflows/<pack>/<workflow>/` and `.archon/workflows/<pack>/.shared/`. A binary materialises that result: 97 files.

A source build does not use the generator at all. `workflow-source.ts:561-562` copies `dirname(getDefaultWorkflowsPath())` and `dirname(getDefaultCommandsPath())` wholesale — the entire directories, **203 files** today.

Nobody decided that the extra files should ship. They matched a pattern.

## What is actually in there

```
defaults        21      sdlc            88
dropshipping     3      experimental     4
maintainer       6      rasmus-tests     9
test-workflows  23
```

`dropshipping` is a client's workflows. `rasmus-tests` and `experimental` are the maintainer's. `test-workflows` is fixtures. None of those are bundled defaults, and on a source install every one of them is copied into every child-run capture, labelled bundled.

That is the same shape as the GAP #2200 leak this repository already characterises: something scoped to one place becoming globally visible because discovery reads a directory instead of an owned list.

## Decision

**Ship a bundle index: an explicit list of packs that get bundled.**

- `sdlc` (88 files) ships.
- `defaults` (21 files) is deprecated and comes out at the next release. Removing it then is a one-line diff to the index.
- `dropshipping`, `experimental`, `maintainer`, `rasmus-tests`, `test-workflows` are Archon-specific or experimental and do not ship.

The index replaces pattern-matching so that shipping a pack is a deliberate act rather than a side effect of creating a directory. Adding a future pack is a line in the index.

## Design

- **The index is the single owner both build paths read.** The generator reads it to decide what to embed; the source-branch capture reads it to decide what to copy. Neither derives its answer from a directory scan. This is the part that actually closes the disagreement — fixing only the binary side leaves source free to drift again.
- **Fail closed on a missing pack.** A directory that is not in the index simply does not ship, silently and correctly. An index naming a pack that does not exist must fail the build, or a typo ships nothing and looks fine.
- **Put it beside `tier-defaults.json`** in `packages/workflows/src/defaults/`. There is already precedent for build configuration as JSON there, and `check:bundled` already exists as the staleness gate, so the index inherits enforcement rather than needing new machinery.
- **The conformance test asserts both branches produce exactly the index's set** — not merely that they agree with each other, which is satisfiable by both being wrong.

## Dependency check: clear

`sdlc` is self-contained. All 8 of its `include:` targets resolve inside the pack: `archon-deliver`, `archon-implement`, `archon-investigate`, `archon-plan`, `archon-pr`, `archon-review`, `archon-triage`, `archon-validate`.

The 10 files mentioning `maintainer` and `defaults` are prose — "maintainer-gated CI", "a maintainer's approval", "if_skipped defaults", "auto-defaults". No resolution edge leaves the pack, so shipping `sdlc` alone breaks nothing.

## Fixtures are the second half of the same bug

Pack fixtures do not belong in a bundle, and the binary already agrees: `.archon/workflows/sdlc/plan/fixtures/ready.stubs.yaml` is on disk and has **zero** occurrences in `bundled-defaults.generated.ts`. The generator collects workflow YAML, commands and scripts by kind; fixtures are not one of those kinds, so they never reach a binary.

The source branch takes them anyway, because it copies directories rather than asking the generator what it collected.

In `sdlc` that is not a rounding error:

```
fixtures       48 files /  64,523 bytes
non-fixtures   40 files / 218,976 bytes
```

**So a pack-level index alone does not close the disagreement.** Ship only `sdlc` and copy its directory, and source still captures 88 files where a binary captures 40. The index has to select packs *and* the source branch has to honour the generator's per-file selection within them. Otherwise the two builds still diverge, just less visibly.

That changes the numbers:

| | files per capture |
|---|---|
| today | 203 |
| index only, directory copy | 88 |
| index plus generator selection | ~40 |

The last row is the one where source and binary actually produce the same thing, which is the point of the issue. The write reduction is a consequence.

## Invariants

- Bundled bytes stay present in every run's own capture. `workflow-source.ts:555-559` records why: it is what lets a paused run resume across an upgrade. This issue decides *which* bytes, never whether they are there.
- Late resolution stays. Discovery runs at spawn time by design, locked by `describe('workflow: late resolution is a deliberate affordance')` in `subrun.test.ts`.

## Acceptance

- An index file that both build paths read, with the reasoning recorded next to it.
- Source and binary produce the same set for the same commit.
- A test that fails if either branch stops matching the index.
- A missing pack named in the index fails the build.
- Measured before/after on files written per capture.

## Side effect worth stating

A capture goes from 203 files to roughly 90. That is the write half #3277 did not address, and the half Windows charges for. It is a consequence, not the reason — the reason is that a client's workflows have no business in Archon's bundled scope.

## Links

- #3277 — removed the per-capture read cost
- #3282 — Windows measurement for the other write-side levers
- #2924 — the Windows budget-edge ledger this cost keeps feeding

Wirasm:
## Problem and outcome

A source build and a binary build disagreed about what "the bundled scope" is. The generator selected by pattern and materialised 97 files; a source build copied the whole `.archon/workflows` and `.archon/commands` directories into every child-run capture, 203 files, including a client's workflows, maintainer experiments, and pack fixtures. Nobody decided those should ship; they matched a directory scan. That is the same shape as the GAP #2200 leak.

- **Outcome:** `packages/workflows/src/defaults/bundle-index.json` names the packs that ship. The generator, the source-build capture, and live discovery all select through one shared inventory, so source and binary produce the same file set for the same commit. Per capture, bundled files written go from 200 to 97 and bundled bytes from 1,280,936 to 762,369.
- **Invariant:** Every run still receives its own frozen copy of the bundled bytes, so a paused run resumes across an upgrade. Immutable capture verification and the digest fold are unchanged. Late resolution stays: discovery still runs at spawn time.
- **Scope boundary:** `defaults` stays in the index through its #2781 deprecation window; removing it later is a one-line index edit. Previously captured runs keep their recorded files and do not consult the current index. The generated-definition fallback used by bundled subrun validation is not redesigned. No timeout, skip, or concurrency setting changed.

This PR also carries a small #3287 change: fixture workspace creation runs `git worktree add` first and only invokes the `git rev-parse` eligibility probe when creation fails. That removes two successful-path git subprocesses per isolation test while keeping the separate not-a-repository and unusable-HEAD diagnostics.

## Review guidance

- **Feedback requested:** Correctness of the shared inventory as the single owner; whether `bundle-inventory.ts` is the smallest truthful implementation; whether any live reader still reaches a directory scan or a runtime JSON read on the binary path.
- **Start here:** `packages/workflows/src/defaults/bundle-inventory.ts` — `readBundleIndex()` and `collectBundleSources()` are what every consumer now calls.
- **Review order:** `bundle-index.json` and `bundle-inventory.ts`; `scripts/generate-bundled-defaults.ts` (generator now delegates selection); `packages/workflows/src/workflow-source.ts` (source-build capture uses the same selection); `workflow-discovery.ts`, `script-discovery.ts`, `validator.ts`, `executor-shared.ts` (live readers respect the index); `fixture-runner.ts` (#3287); tests.
- **Lower-attention areas:** `bundled-defaults.generated.ts` is regenerated by `check:bundled`. `scripts/capture-cost.ts` and the capture profiler are deleted per the end-of-life condition recorded on #3282; nothing else references them.
- **Known risk or uncertainty:** None open. The prp review's R1 (a flat default command masked by a same-named nested file was silently unresolvable) and R2 (the pack-membership predicate hand-copied across seven call sites) are fixed in `908a6dc35`: live default commands resolve by direct path, the recursive walk is kept only for captured runs, and `bundle-inventory.ts` owns the membership helpers. A regression test covers all four surfaces and was confirmed red on the prior head. CodeRabbit's one inline finding was the same defect.

## Solution

One JSON index, beside `tier-defaults.json`, lists the packs that ship. `bundle-inventory.ts` reads it and selects runtime YAML, commands, script entrypoints, and shared modules from those packs, instead of walking every adjacent fixture, document, or experimental pack. The generator embeds that selection into the binary constants. The source-build capture applies the same selection and the generator's LF normalization, so the bytes agree as well as the paths. Live workflow and script discovery, command lookup, and validation consult the index on a source build and the generated constants on a binary build.

Failure is explicit: a pack named in the index with no directory, a malformed index, a path collision, or a partial or non-directory installation all fail with a named error. A genuinely absent SDK bundle remains supported.

Hardlinks were measured as faster on Windows but were not introduced, because a hardlink shares mutable bytes with the authoring tree or another run and breaks the frozen-capture invariant.

## Validation

Contributor's Windows host (Windows 11, Bun 1.4.2, frozen lockfile), from the handoff on `fix/windows-suite-headroom`:

- `bun run validate` — generated-file checks, type checks, lint, and formatting passed. The full test step stopped on a symlink `EPERM` in the provider binary-resolver test, which fails identically on unchanged `dev` on that host (no Developer Mode). CI is the verification for the full suite.
- Every one of the workflow package's 28 declared test groups run in its own process — 2,901 passed, 13 existing skips, 5 failures, all five the same host-only symlink `EPERM` reproduced on unchanged `dev`.
- New tests: `bundle-live-selection.test.ts` 12/12 (including a mocked `readBundleIndex` that throws on the binary and captured paths, proving neither reads the live index); `bundle-conformance.test.ts` 1/1 (source and binary capture produce the index's set, not merely each other's); `bundle-inventory.test.ts` 4/4 (including Windows directory-junction dereferencing); an unborn-HEAD fixture-runner case.
- Guard sensitivity: nine in-memory mutations of the isolation path (caller cwd, inherited PWD, live-source resolution, HEAD script, per-fixture capture, skipped disposal, plain non-git scratch, ignored command folder) each produced the expected failing assertion.
- Windows capture measurement, 7 warm samples per condition: bundled-write median 73.64 ms to 37.70 ms idle, 301.95 ms to 145.66 ms under 4 workers. The loaded warm maximum did not move (2,768 ms to 2,759 ms), so this is a cost reduction, not the cause of the Windows long tail.
- **Not verified:** Nothing material. Both CI legs are green on the current head, including the full Windows and Ubuntu suites. The contributor's Windows capture measurements were taken with Defender real-time protection on, while the CI runner image excludes `C:` and `D:` from Defender, so the CI benefit of the smaller bundle is smaller than the local one.

Full investigation reports and raw evidence: https://github.com/coleam00/Archon/tree/fix/windows-suite-headroom/docs/windows-issues-2026-09-10 and the handoff at https://github.com/coleam00/Archon/blob/fix/windows-suite-headroom/WINDOWS-ISSUES-HANDOFF.md. This PR carries the source commit only; the evidence archive stays on the investigation branch.

## Delivery considerations

|---|---|---|
| Compatibility / migration | Runs captured before this change keep their own recorded files. `defaults` still ships until its deprecation window closes. | `workflow-source.ts` recorded-root branch; `bundle-live-selection.test.ts` captured-consumer case |
| Rollout / rollback | Pure revert. Removing a pack later is one line in `bundle-index.json`; `check:bundled` regenerates the constants. | `bundle-index.json` |

## Links

- Closes #3283
- Related #3287, #3282, #3294, #2924

https://claude.ai/code/session_01QDVDEWqYZdkqKaLcBhJ5ut

## Summary by CodeRabbit

* **New Features**
  * Added indexed bundle selection for packaged workflows, commands, and scripts.
  * Live changes to indexed workflows and scripts now appear without rebuilding binary bundles.
  * Project content can preserve authored workflow filenames when overriding bundled resources.

* **Bug Fixes**
  * Improved handling of missing, partial, invalid, or non-directory bundle installations.
  * Unindexed packs, fixtures, and unsupported files are excluded from discovery and capture.
  * Improved validation and resolution of bundled command and workflow references.


Wirasm:
## Ready to merge

Round 2 verifies the correction commit `908a6dc35bf41b754252c37e2bc1559ddc6fc277` against round 1's two findings. Both are closed. `R1` (Important) is fixed by adding `bundledDefaultCommandPath()` and `listBundledDefaultCommands()` to `bundle-inventory.ts`, which resolve a live bundled-default command by a direct `join(root, \`${commandName}.md\`)` lookup instead of filtering a recursive, basename-deduped directory walk — closing the false "not found" on a root command masked by a same-named nested file, at all four call sites that had the bug. A new regression test proves it across all four affected surfaces (listing, execution, validation, include-block command scan), and a before/after run against the pre-fix commit confirms the test genuinely catches the original bug. `R2` (Suggestion) is fixed by the same change: no consumer file reads `readBundleIndex()` directly anymore, all seven original call sites now route through three owned helper functions in `bundle-inventory.ts`. Two deliberate behavior changes that came with the fix — an ENOENT on the new direct-path guess now returns an ordinary miss in the include scanner, and a non-ENOENT stat failure now throws instead of returning null in the validator — were checked against the surrounding code and match the pattern the codebase already uses for the equivalent packaged-command case a few lines above each site in the same functions.

**Validation:** All 8 required GitHub Actions checks green on `908a6dc3`, including full Windows and Ubuntu suites. Local re-run of `check:bundled`, the workflows package type-check, the new regression test in isolation, and the full package test suite all pass with no evidence gaps.

**Resolved:** 2 (`R1`, `R2`)

### Findings

|---|---|---|---|
| `R1` | Important | Flat default-command lookup silently drops a root command masked by a same-named nested file | FIXED |
| `R2` | Suggestion | Pack-index-membership predicate duplicated across 7 call sites in 3 files | FIXED |

<details>
<summary><code>R1</code> — Flat default-command lookup silently drops a root command masked by a same-named nested file</summary>

**Impact:** A bundled default command whose basename collided with a file one level deeper in `commands/defaults/` became unresolvable: `resolveCommand`/`loadCommandPrompt` returned "not found" and `discoverAvailableCommands` silently omitted it from the listing, even though the root file was present.

**Evidence (round 1):** `packages/workflows/src/validator.ts:184-186,289-293`, `packages/workflows/src/executor-shared.ts:583-593`, `packages/workflows/src/workflow-discovery.ts:508-525` all filtered `findCommandFiles`'s recursive, basename-deduped walk down to an exact-path match, but the walk had already dropped the root entry whenever a same-named nested file existed. Reproduced directly with a throwaway fixture.

**Fix (round 2):** `packages/workflows/src/defaults/bundle-inventory.ts` adds `bundledDefaultCommandPath(root, name)` — a direct `join()` lookup — and `listBundledDefaultCommands(root)`, both reusing a new shared `defaultCommandFiles()` walk that is also what `collectBundleSources` uses for selection. All four call sites now use these instead of the recursive walk; the recursive walk is intentionally kept only for `roots.kind === 'captured'`, where a capture's own frozen layout (subfolders included) is the correct answer. New regression test `'a same-named nested file does not hide a flat bundled default command'` in `bundle-live-selection.test.ts` exercises all four affected surfaces: `discoverAvailableCommands` (listing), `loadCommandPrompt` (execution), `validateWorkflowResources` (validation), and `discoverWorkflows` on an `include:`-chained workflow (the scanner path). Verified failing against the pre-fix commit and passing after, both by the PR author and independently reproduced during this review. `bun test src/bundle-live-selection.test.ts` — 13 pass, 0 fail. Grepped the package for the original `relativePath === ...` pattern against a recursive walk in a live (non-captured) branch — zero remaining instances.

The correction also introduces two deliberate, checked behavior changes on the "path is a candidate, not a hit" contract `bundledDefaultCommandPath` documents: `workflow-discovery.ts`'s include scanner now treats an ENOENT on the guessed path as an ordinary miss (matches the identical pattern in the packaged-command branch ~40 lines above in the same function); `validator.ts`'s `resolveCommand` now throws a wrapped error on a non-ENOENT stat failure instead of silently returning null (matches the packaged-command case ~45 lines above, same wrapped-`Error`-with-`cause` shape). Both are consistent with existing sibling code and are not correction-caused defects.

**Disposition:** FIXED. Verified independently by two fresh correction-verification passes (`code-reviewer`, `seam-analyzer`) plus the coordinator's own direct diff read and test run, all in agreement.

</details>

<details>
<summary><code>R2</code> — Pack-index-membership predicate duplicated across 7 call sites in 3 files</summary>

**Impact:** No proven drift at round 1, but seven independent copies of `(await readBundleIndex()).includes(...)`-shaped checks across `executor-shared.ts`, `validator.ts`, and `workflow-discovery.ts` — a coordination cost the PR author self-flagged and deliberately left unfixed for review.

**Evidence (round 1):** `packages/workflows/src/executor-shared.ts:439,589`, `packages/workflows/src/validator.ts:182,232,289`, `packages/workflows/src/workflow-discovery.ts:460,525`.

**Fix (round 2):** `bundle-inventory.ts` now exports `bundlesPackagedResources(pack)`, `bundledDefaultCommandPath(root, name)`, and `listBundledDefaultCommands(root)`, and all seven original call sites route through them. Grepped all three consumer files for `readBundleIndex` — zero direct calls remain; the only production callers are inside `bundle-inventory.ts` itself. The three new exports are direct, single-purpose functions with no new types, config surface, or generics — genuinely a consolidation, not new abstraction. Call-site-specific error handling (validator throws and logs; the include scanner and executor treat a miss as null) correctly stays at each call site rather than being flattened into the shared helper.

**Disposition:** FIXED. Verified independently by three fresh correction-verification passes (`code-reviewer`, `seam-analyzer`, `code-simplifier`), all in agreement, with no new simplification concern raised by the new helpers.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 verified FIXED, R2 verified FIXED |
| `seams` | R1 verified FIXED, R2 verified FIXED |
| `simplify` | R2 verified FIXED |

#### Validation

|---|---|---|
| CI: `Test Suite`, `changes`, `docker-build`, `postgres-parity`, `schema-upgrade`, `test (ubuntu-latest)`, `test (windows-latest)`, `workflow-fixtures` | PASS | All 8 required GitHub Actions checks green on `908a6dc3`, including full Windows (10m9s) and Ubuntu (5m54s) suites (`gh pr checks 3301`) |
| `bun run check:bundled` (repo root, local checkout) | PASS | "bundled-defaults.generated.ts is up to date (54 commands, 31 workflows, 1 script packs)." |
| `bun x tsc --noEmit` (`packages/workflows`, local checkout) | PASS | No type errors |
| `bun test src/bundle-live-selection.test.ts` (`packages/workflows`, local checkout) | PASS | 13 pass, 0 fail, including the new regression test |
| `bun run ../../scripts/package-tests.ts` (`packages/workflows`, local checkout) | PASS | Every test file in the package reported `0 fail`; exit code 0 |
| Pre-fix reproduction (throwaway worktree at `3596d54d3` with the new test copied in) | FAIL as expected | 12 pass / 1 fail at `discoverAvailableCommands`, confirming the regression test catches the original bug |

</details>

PR #3307, opened 2026-09-11
test: name the poll-wait deadlines so no call site inherits one silently

### LINKED ISSUE #3293: test(core): run-live-owner's waitFor shares an implicit deadline across six call sites
## Problem

`packages/core/src/services/run-live-owner.test.ts` has a local `waitFor` helper with an implicit default deadline, and every call site takes it silently:

```ts
async function waitFor(check: () => boolean, timeoutMs = 3_000): Promise<void>
```

Six call sites — `:88`, `:112`, `:141`, `:147`, `:162`, `:240` — none of which states a deadline. All six therefore share one number that none of them names, and nothing makes them move together when it turns out to be wrong for one of them.

## Why this is worth solving

This is the same defect #3289 just fixed one package over, and the fix there established what the right shape is.

In `workflow-wait.integration.spec.ts`, three sites waited on the same condition through the same helper. One passed `30_000` explicitly, two took the default. When 30 s turned out to be too tight on Windows, only the explicit one was visible as a thing to change — and the failure took a while to diagnose precisely because the other two were invisible.

The fix was not to raise the number. It was to give the deadline **one owner** (`RUN_BOOT_DEADLINE_MS`, spent only through a `waitForRunBoot` wrapper) and then **remove the helper's default entirely**, so a fourth call site cannot join implicitly. The default was the mechanism, not the number.

`AGENTS.md` is direct about this: two declarations that must agree to stay correct, kept in agreement by discipline rather than by a mechanism, are a present defect and not a future risk. Six sites sharing an unnamed default is that, six times over.

## Why now

The pattern is fixed in one place and live in another. That is the worst state to leave it in — the next person to touch either file copies whichever they read first, and half the repository teaches the wrong lesson. Doing it while the correct shape is fresh and demonstrated costs less than rediscovering it later.

This is also not hypothetical decay. The `workflow-wait` version sat harmlessly until Windows CI made one site's deadline wrong, and then cost real diagnosis time. `run-live-owner.test.ts` tests owner liveness and control handoff — timing-sensitive behaviour of exactly the kind that eventually meets a slow runner.

## Desired outcome

The deadline in `run-live-owner.test.ts` has one owner, and no call site can inherit it silently.

## Solution steering

Optional, and the shape is already proven next door in `workflow-wait.integration.spec.ts` — read it before designing something new:

- a named constant, so the number has a reason attached to it;
- spent through one wrapper rather than passed at each site;
- **the helper's `timeoutMs` default removed**, so a future site must state its own value.

Whether 3 s is the right number is a separate question from whether it should be implicit. Do not raise it as part of this — if a site needs longer, that is evidence to bring back, not a thing to absorb quietly.

## Invariants

- No behaviour change to what the tests assert. This is about how the deadline is expressed.
- Do not raise any deadline as part of this change. #2924 forbids fixing timing failures by budget bump, and this issue is not a timing fix.
- If removing the default forces another call site to state a value, that site's value needs a stated reason too, not an inherited number.

## Acceptance

- One named owner for the deadline, with the reasoning recorded next to it.
- `waitFor`'s implicit default removed; every call site states its deadline or goes through the wrapper.
- The suite still passes with no assertion changed.
- Adding a new call site without a deadline fails to compile or fails a test, rather than silently inheriting.

## Links

- #3289 — fixed the identical shape in `packages/cli/src/commands/workflow-wait.integration.spec.ts`; its review found this one
- #3288 — the Windows failure that exposed the pattern
- Review that identified it: https://github.com/coleam00/Archon/pull/3289#issuecomment-5618098184 (finding R5, verified by direct file read)

Wirasm:
## Problem and outcome

`waitFor` in `packages/core/src/services/run-live-owner.test.ts` expressed its deadline as a default parameter (`timeoutMs = 3_000`), so all six call sites shared a number none of them named. The two sibling helpers the issue's review comment pointed at had the identical shape: `detached-run-control.integration.spec.ts` (`timeoutMs = 5_000`, 3 sites) and `workflow-terminal-event.integration.spec.ts` (`timeoutMs = 15_000`, 5 sites). A wait whose needs differ could not state its own window without editing the shared default, and nothing recorded why the old value was chosen.

- **Outcome:** each of the three files states its poll deadline as one named constant spent through a single wrapper. `waitFor` no longer defaults its `timeoutMs`, so a new call site cannot inherit a deadline silently — it must state one or use the wrapper.
- **Invariant:** no deadline value changed (`3_000`, `5_000`, `15_000` all preserved) and no assertion or test body moved.
- **Scope boundary:** this is not a shared or cross-package deadline utility. Each file owns a file-local constant, matching the demonstrated pattern. `waitForExit`'s inline `8_000` in `detached-run-control.integration.spec.ts` is deliberately untouched: it is a literal inside one helper, not an inherited default.
- **Root cause:** the deadline lived in the helper's default parameter, so it was shared by every call site without being owned or justified by any of them. A hand-synced pair of that kind is a present defect, not a future risk.

## Review guidance

- **Feedback requested:** correctness of the wrapper shape, and whether including the two sibling files is the right scope call.
- **Start here:** `packages/core/src/services/run-live-owner.test.ts:16` — the comment recording the old default, `LIVE_OWNER_EVENT_DEADLINE_MS`, and `waitFor`'s now-required `timeoutMs` sit together; the wrapper `waitForOwnerEvent` follows.
- **Review order:** the core file, then `packages/cli/src/utils/detached-run-control.integration.spec.ts`, then `packages/cli/src/commands/workflow-terminal-event.integration.spec.ts`.
- **Lower-attention areas:** the call-site edits are mechanical renames from `waitFor(...)` to the file's wrapper; no argument or condition changed.
- **Known risk or uncertainty:** the two sibling files were not named in the issue title, only in its review comment, which deferred the scope question. Both were included because they have multiple inheriting call sites, which is the issue's own criterion for belonging in the same change. This is the one judgment call in the diff.

## Solution

The shape is already demonstrated in-repo at `packages/cli/src/commands/workflow-wait.integration.spec.ts:360-465`, where `waitFor` has no default, `RUN_BOOT_DEADLINE_MS` is the single owner, and `waitForRunBoot` spends it. Each of the three files now follows it:

1. The old default value becomes a named `*_DEADLINE_MS` constant with the reasoning recorded beside it.
2. `waitFor` keeps its `timeoutMs` parameter but it is now required.
3. One wrapper (`waitForOwnerEvent`, `waitForFixtureProcess`, `waitForDetachedRun`) spends the constant.
4. Every existing call site goes through the wrapper rather than restating the number.

Because `timeoutMs` is required, a future site that writes `await waitFor(() => true)` fails to compile with `TS2554: Expected 2 arguments, but got 1` — the compile guard the acceptance asks for.

## Validation

- `bun run validate` — exit 0 — covers `check:*`, `type-check`, `lint --max-warnings 0`, `format:check`, `test:install`, and the full `bun run test` suite. Re-run with the exit code captured: `validate exit=0`.
- `packages/core/src/services/run-live-owner.test.ts` — 11 pass — the suite still asserts what it did before; no assertion was edited.
- `packages/cli/src/utils/detached-run-control.integration.spec.ts` + `packages/cli/src/commands/workflow-terminal-event.integration.spec.ts` — 5 pass — both sibling suites unchanged in behavior.
- Compile guard proven by probe: adding `await waitFor(() => true)` produced `error TS2554: Expected 2 arguments, but got 1.` The probe was reverted and the file restored before commit.
- `prettier --check` clean on all three files; `git status` clean after commit.
- **Not verified:** Nothing material. Every touched file is exercised by the commands above, and the deadline values are byte-for-byte the previous ones.

## Links

- Closes #3293

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of workflow, detached-run, fixture-process, and live-owner integration tests by applying explicit, consistent polling deadlines.
  * Standardized terminal status, event, checkout, shutdown, and run-creation waits to use shared timeout limits.
  * Removed implicit default timeouts from internal test wait helpers, requiring each wait to specify its deadline.

### DIFF
--- a/packages/cli/src/commands/workflow-terminal-event.integration.spec.ts
+++ b/packages/cli/src/commands/workflow-terminal-event.integration.spec.ts
@@ -31,9 +31,20 @@ afterEach(async () => {
   }
 });
 
+/**
+ * How long the detached workflow's observable state may take to catch up with what the
+ * fixture was launched to do: write its run row, record its checkout, release the control
+ * endpoint, and commit the terminal status and event.
+ *
+ * This was the default on `waitFor`. Every site below waits on the same detached run, so
+ * one window covers them; naming it and removing the default means a wait with a
+ * different need has to state its own deadline instead of inheriting this one.
+ */
+const DETACHED_RUN_DEADLINE_MS = 15_000;
+
 async function waitFor<T>(
   read: () => T | undefined | Promise<T | undefined>,
-  timeoutMs = 15_000
+  timeoutMs: number
 ): Promise<T> {
   const deadline = Date.now() + timeoutMs;
   let lastError: unknown;
@@ -48,6 +59,13 @@ async function waitFor<T>(
   }
   const detail = lastError instanceof Error ? `: ${lastError.message}` : '';
   throw new Error(`Timed out waiting for detached workflow${detail}`);
+}
+
+/** Wait for the detached workflow under test to reach the state `read` describes. */
+async function waitForDetachedRun<T>(
+  read: () => T | undefined | Promise<T | undefined>
+): Promise<T> {
+  return waitFor(read, DETACHED_RUN_DEADLINE_MS);
 }
 
 function readRun(
@@ -193,7 +211,7 @@ nodes:
       if (typeof ackRunId !== 'string') throw new Error(`ack carried no run id: ${stdout}`);
       expect(readRunById(databasePath, ackRunId)?.id).toBe(ackRunId);
 
-      const created = await waitFor(() => readRun(databasePath, fixture.workflow));
+      const created = await waitForDetachedRun(() => readRun(databasePath, fixture.workflow));
       activeRunIds.add(created.id);
       expect(created.id).toBe(ackRunId);
       // The child fills in the checkout the parent could not know at fork time.
@@ -202,7 +220,7 @@ nodes:
       // `working_path` is whatever that function returned — a different realpath
       // variant here disagrees with it on Windows (#2927).
       const resolvedProjectRoot = await canonicalizeProjectPath(projectRoot);
-      await waitFor(() =>
+      await waitForDetachedRun(() =>
         readRunById(databasePath, created.id)?.working_path === resolvedProjectRoot
           ? true
           : undefined
@@ -212,12 +230,12 @@ nodes:
       // the loop past that. Returning from this wait is the assertion that the owner
       // released its endpoint; re-asserting the same reading on the next line could only
       // turn one such misread into a failure, which is how it failed on Windows.
-      await waitFor(async () =>
+      await waitForDetachedRun(async () =>
         (await canConnectToRunLiveOwner(runLiveOwnerPath(created.id))) ? undefined : true
       );
       activeRunIds.delete(created.id);
 
-      const terminal = await waitFor(() => {
+      const terminal = await waitForDetachedRun(() => {
         const run = readRun(databasePath, fixture.workflow);
         return run?.status === fixture.status ? run : undefined;
       });
@@ -231,7 +249,7 @@ nodes:
       // contention, and the count stays exact: the event was committed atomically with
       // the status already observed above, and the owner's endpoint is unreachable, so
       // nothing can append a second row after the first read succeeds.
-      const events = await waitFor(() => {
+      const events = await waitForDetachedRun(() => {
         const rows = readTerminalEvents(databasePath, terminal.id, fixture.event);
         return rows.length > 0 ? rows : undefined;
       });
--- a/packages/cli/src/utils/detached-run-control.integration.spec.ts
+++ b/packages/cli/src/utils/detached-run-control.integration.spec.ts
@@ -13,12 +13,26 @@ import { requestDetachedRunStop } from './detached-run-control';
 // cleanup, and an unretried removal fails a test whose assertions already passed (#2306).
 const trackTempRoot = trackTempRoots();
 
-async function waitFor(check: () => boolean, timeoutMs = 5_000): Promise<void> {
+/**
+ * How long a spawned fixture process may take to reach the state a test polls for: to
+ * signal that it is ready, or to be gone once the terminator stopped it.
+ *
+ * This was the default on `waitFor`. Naming it and removing the default means a wait that
+ * needs a different window has to state one rather than inherit this.
+ */
+const FIXTURE_STATE_DEADLINE_MS = 5_000;
+
+async function waitFor(check: () => boolean, timeoutMs: number): Promise<void> {
   const deadline = Date.now() + timeoutMs;
   while (!check()) {
     if (Date.now() >= deadline) throw new Error('Timed out waiting for fixture');
     await new Promise<void>(resolve => setTimeout(resolve, 25));
   }
+}
+
+/** Wait for a spawned fixture process to reach the state `check` describes. */
+async function waitForFixtureProcess(check: () => boolean): Promise<void> {
+  await waitFor(check, FIXTURE_STATE_DEADLINE_MS);
 }
 
 function waitForExit(
@@ -114,7 +128,7 @@ describe('detached run control integration', () => {
     const exited = waitForExit(owner);
 
     try {
-      await waitFor(() => existsSync(readyPath));
+      await waitForFixtureProcess(() => existsSync(readyPath));
       const pids = JSON.parse(readFileSync(readyPath, 'utf8')) as {
         owner: number;
         leakWriter: number;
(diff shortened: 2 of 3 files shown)


Wirasm:
# Review report — PR coleam00/Archon#3307 (issue #3293)

Reviewed head SHA: `5c5117164e6064bb7c66967db01ecd0245be7ff7`
Base: `dev` (`5a159c144f526a0c484330bab62503707689b256`)
Mode: continuation review, round 2. Delta since `5c5117164e6064bb7c66967db01ecd0245be7ff7`.

## 1. Verdict

- **Ready:** yes.
- **Action:** `none`.
- **Reason:** the delta is empty — the PR head equals the prior report's reviewed-head cursor and
  `git status` is clean — the prior round's findings are all accounted for (there were none), and
  the two accepted discoveries remain valid with one new adjacent record preserved.

## 2. Accepted contract

**Required outcome.** Each of the three poll-wait helpers states its deadline as one file-local
named constant spent through a single wrapper, and `waitFor` no longer defaults its `timeoutMs`,
so a new call site cannot inherit a deadline silently. The deadline values are unchanged
(`3_000`, `5_000`, `15_000` all preserved) and no assertion or test body moves. A call to
`waitFor` on one argument must fail to compile.

**Explicit non-goals / boundaries.**

- Not a shared or cross-package deadline utility. Each file owns a file-local constant.
- `waitForExit`'s inline `8_000` in `detached-run-control.integration.spec.ts` is deliberately
  untouched: a literal inside one helper, not an inherited default.
- No deadline value may be raised (#2924 forbids fixing a timing failure by budget bump).
- Including the two sibling CLI files is the one judgment call in the diff; the criterion used
  is that each helper had multiple inheriting call sites.

## 3. Reviewed head SHA

`5c5117164e6064bb7c66967db01ecd0245be7ff7` (next round's cursor). The local checkout is on
`fix/issue-3293-deepseek` at that exact SHA and `git status --porcelain --untracked-files=all`
is empty.

## 4. Findings

None. No Critical, Important, or Suggestion finding is open, and none was raised this round. No
finding was rejected because none was raised.

This round reviews the delta `git diff 5c5117164e6064bb7c66967db01ecd0245be7ff7..HEAD`, which
produces **no output**: the head does not move. There is therefore no correction delta to
introduce a defect, and no new surface for a lens to examine.

## 5. Prior findings

The prior report (round 1) returned **no findings** — not a clean carry-forward of open items,
but an empty finding set. There is nothing to mark fixed, open, or disproved, and no `sources`
to preserve. This is the complete carried-forward table:

| ID | Sources | Prior verdict |
| --- | --- | --- |
| — | — | (none — prior report raised no finding) |

Both accepted discoveries from round 1 were re-validated against their cited evidence this round
and remain accurate; neither is a finding and neither affects readiness:

- The transcript 30 s boot deadline record still matches
  `packages/cli/src/commands/workflow-transcript.integration.spec.ts:99-100`.
- The pass-through wrapper record still matches the three wrapper bodies and the required
  `timeoutMs` parameters.

## 6. Discoveries

Three accepted discoveries (two `adjacent`, one `scope_conflict`). Adjacent discoveries do not
affect readiness.

1. **workflow-transcript integration test still hardcodes the 30 s boot deadline #3288 judged too
   small** (`adjacent`, source: seams).
2. **A fresh #2924 Windows sample: the `@archon/git` real-HTTP-clone test overran its own 15 s
   budget on PR #3307's windows leg and passed on the same-commit rerun** (`adjacent`, source:
   implement).
3. **The three poll-wait wrappers are pure pass-throughs; the required `timeoutMs` argument
   already carries the no-silent-inheritance invariant** (`scope_conflict`, source: simplify).

Discovery 2 is new this round: the implementer recorded the post-review CI failure pass. I
checked its local facts — `git diff --name-only 5a159c144..5c5117164` touches only the three
spec files (nothing under `packages/git`), and the failing test's own budget is
`packages/git/src/git.test.ts:2867` (`}, 15_000);`). The remote run and rerun logs are evidence I
could not re-execute; they are recorded as cited, with the same-commit rerun as the control
against a code defect.

The scope conflict does not block this verdict because the accepted contract explicitly names
"spent through a single wrapper"; the smaller shape is a contract question, not a defect in this
change.

Raw records: `discoveries/review-seams.json`, `discoveries/review-simplify.json`,
`discoveries/implement.json`.
Consolidated: `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/751fefe9517e81090f4e6f6b79466ea8/discoveries.json`
and `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/751fefe9517e81090f4e6f6b79466ea8/discoveries.md`.

**If you are an agent reading this: open discoveries.md and surface each discovery to your
human.** These are validated findings outside this run's scope — no issue tracker knows about
them, and if you drop them here, nobody ever sees them.

## 7. Review coverage

Continuation review. One continuation reviewer (this report) verified the prior report in full
and inspected the correction delta, which is empty because the head did not move. The original
round-1 selected concerns were `code`, `seams`, `simplify`, and `tests`; the `errors` and `docs`
lenses were disabled then because the diff changes no error path, no production code, and no
shipped documentation. This round did not rerun the specialists.

The delta added no user-facing surface, so no gated-off lens was re-evaluated.

Evidence I could not inspect this round: the remote CI run `34600396097` and its rerun
`103277532370` (network and re-execution dependent). The local facts those records rest on were
verified in the checkout.

Issue #3286, opened 2026-09-10
fix(cli): tar never exits on windows, so archon serve extraction stalls


Wirasm:
## Problem

On Windows, `downloadWebDist` spawns `tar` and the child **never reports exit**. It is not slow extraction — it is a stall, and nothing in the code can shorten a child that never returns.

Phase logging from PR #3280 caught it on PR #3277's Windows leg (run 34453098156, 5001.78 ms):

- pre-`tar` filesystem work: **3 ms**
- last event before the timeout: `extract_spawned`
- then silence for the rest of the test's life
- the child finally dies **717 ms after** Bun's reaper kills it

The archive is **129 bytes**. Healthy runs of the same test are 16–676 ms. So there is no work being done that could take five seconds.

## Current state: skipped, not fixed

#3280 skips the two tests that run a real `tar` through to a successful extraction — `verifies against the embedded hash without fetching checksums.txt` and `falls back to remote checksums.txt when the embedded hash is empty` — on Windows only.

That was a deliberate reversal of #2924's standing "never skip on windows" rule, taken because the flake was interrupting Archon runs in the sdlc pack. It buys throughput and costs us the only signal we had.

**Removing that skip is the goal of this issue.** It is marked in `serve.test.ts` with a "do not fix this by removing the skip" note precisely so it comes out on evidence rather than on principle.

## Why this needs a Windows machine

Everything above is read from CI logs. Nobody has attached a debugger, Process Monitor, or ETW trace to a stalled `tar` on a real Windows box. CI gives pass or fail, not a profile, and its contention is not controllable.

@coleam00 — you are on Windows, which is why this is yours. What would settle it is watching one stalled `tar` directly.

## What is already ruled out

- **Not the stdin pipe.** #2928 replaced the parent-pumped `Uint8Array` stdin with `Bun.file()`. Current `serve.ts:238` passes a kernel-served descriptor.
- **Not Defender.** Falsified in `071fcd6f2`.
- **Not an unbounded hang for users.** `e507ac6d2` bounded the child at 60 s (`EXTRACTION_TIMEOUT_MS`). Note that bound has **never once fired**: every sample is from the test lane, where Bun's budget is 5000 ms, 12x shorter, so the runner always kills `tar` first.
[...]

PR #3311, opened 2026-09-11
ci(windows): attributed per-test budget and uv warm-up for the Windows leg

### LINKED ISSUE #2306: ci: windows test timeouts cluster on subprocess-spawning tests under 'bun --filter --parallel'
Split out of #2186 / #2240 so the remaining, genuinely-different half of the flake problem has its own record.

## What is left after #2303

#2303 root-caused and fixed the two mechanisms that had hard evidence:

- adapter tests making live HTTPS calls (#2186) — reproduced digit-for-digit
- `/workflow abandon` tests opening a real SQLite DB (#2240) — proven with an `ARCHON_HOME` probe

A CI-log scan over 500 workflow runs (2026-06-26 → 07-29) shows **20 test timeouts total: 6 ubuntu, 14 windows**. All 6 ubuntu ones are the network tests #2303 fixes. The 14 windows ones are not, and #2303 deliberately did not touch them.

## The windows cluster

| Package | Test | Notes |
|---|---|---|
| `@archon/workflows` | `bash node … stdout` (×4) | spawns a real shell |
| `@archon/isolation` | `apply script — C1 whiteout-name traversal` | synchronous `execFileSync('bash', …)`; seen on #2299 (run 30448971374, job 90566157638) at 5015 ms |
| `@archon/cli` | `downloadWebDist` | `beforeAll` spawns `tar` via `Bun.spawn`; failure reported as a hook timeout |
| `@archon/core` | `should abandon a running run` | the #2240 one — real-DB cause removed by #2303, but see below |

Common shape: **every one spawns a real subprocess**, by design (they are testing shell scripts and tar extraction). They correlate with bad-runner days rather than with any code change — the same tests pass on dozens of other windows runs the same week.

## The suspected amplifier

`bun run test` is `bun --filter '*' --parallel test` — **10 package test processes concurrently on a 4-vCPU runner**, each spawning its own subprocesses. Windows process creation is far more expensive than Linux (Defender scanning, no `fork`), so a cold spawn stalling past a 5000 ms budget is plausible under that oversubscription.

Note CLAUDE.md documents this command as `bun --filter '*' test` — the `--parallel` flag is undocumented.

## What has been ruled out

Recorded so the next attempt does not redo it:

- **Coverage instrumentation.** Root `bunfig.toml` sets `coverage = true`, but Bun reads `bunfig.toml` from cwd only and `bun --filter` runs each package from its own directory — so the root config (coverage *and* preload) never applies under `bun run test`. Verified both directions.
- **Cross-file `mock.module` pollution.** Three of the four affected files are the sole file in their `bun test` invocation.
- **Network.** A sweep across all 10 packages with `fetch` + `node:http`/`https` + `net` + `dns` hooks returns zero events for these tests.

## Open questions

1. Does dropping `--parallel` on CI (or capping concurrency) remove the windows cluster? Costs wall-clock time — needs measuring before adopting.
2. Is the `@archon/core` abandon test fully resolved by #2303's DB fix, or does it still time out on windows? **This is the cheap experiment**: #2303 removed its only real I/O, so a recurrence after that merges is strong evidence for pure scheduler starvation and would settle the question.

Raising the affected tests' timeouts is the obvious stopgap and was explicitly declined in #2303 — it would hide the mechanism, and #2240 already showed that a test looking "merely slow" can be hiding real I/O. Worth revisiting only if 1 and 2 come back negative.

Related: #2186, #2240, #2303, #2299.

### LINKED ISSUE #3294: test(ci): the windows suite has no headroom — 12 tests sit within a second of the 5000ms budget

Wirasm:
## Problem and outcome

The Windows leg fails one random test per few runs at Bun's 5 s default budget, on PRs that touch nothing near the failing package. Fifty instrumented runs on #3294 attribute this to a runner floor rather than test cost: on the 4-vCPU `windows-latest` VM the suite's several hundred child processes periodically saturate the CPUs and the OS disk, sometimes together with Windows' own background maintenance, and any spawn issued in such a second takes 5 to 15 s regardless of which test issued it. Serializing packages, moving temp, moving bun, and stopping Windows services did not change the rate. The full attribution, with per-second disk, CPU, page-fault and per-process data for every stall, is in https://github.com/coleam00/Archon/issues/3294#issuecomment-5634758815.

- **Outcome:** the Windows leg runs with a 20 s default per-test budget, set in one place, and the first `uv run` is paid once by the job instead of inside the first uv-backed test. A Windows timeout again means a test that is genuinely stuck.
- **Invariant:** every test still asserts what it asserted. Explicit per-test budgets are untouched (`--timeout` only sets the default). Linux and macOS keep the 5 s default. Mock isolation between test groups is unchanged: every group still gets its own process.
- **Scope boundary:** no test is skipped or rewritten. The two `tar` skips from #3280 stay until #3286 has a trace of a real stall. The `git clone` integration test's own explicit 15 s budget, which the same floor also blew three times in the instrumented runs, is left as is and noted on #3294.

## Review guidance

- **Feedback requested:** whether a platform-scoped default budget is the right shape for an attributed residual under #2924's rule, and the cli/server migration to `testGroups`.
- **Start here:** `scripts/bun-test-command.ts` — the one owner of every `bun test` invocation and of the Windows budget, with the attribution in its comment.
- **Review order:** `scripts/bun-test-command.ts`; `scripts/package-tests.ts` and `scripts/repo-tests.ts` (both now build the command through it); `packages/cli/package.json` and `packages/server/package.json` (chains become `testGroups`, one group per former `bun test` invocation, same order); `.github/workflows/test.yml` (uv warm-up step and the pointer comment).
- **Lower-attention areas:** the `testGroups` arrays are a mechanical split of the former `&&` chains; `scripts/test-inventory.test.ts` already reads both forms and passes 6/6, which proves every test file is still executed exactly once.
- **Known risk or uncertainty:** the budget is a ceiling on a floor we measured at 5 to 15 s; a worse runner minute can still exceed 20 s. The proof runs below say how often.

## Solution

`bunTestCommand(selectors)` returns `['bun', 'test', ...selectors]`, prefixed with `--timeout 20000` when the platform is `win32`. `scripts/package-tests.ts` (nine packages) and the root plan in `scripts/repo-tests.ts` both build their commands through it. The cli and server packages still spelled their groups as `bun test a && bun test b` chains in `package.json`, which is the form `package-tests.ts` exists to replace and the one place a flag could not reach; they now declare `testGroups` and run the shared runner like every other package.

The budget is a CLI flag rather than a `bunfig.toml` `[test] timeout` key because that key is documented but silently ignored on current Bun (anomalyco/opencode#13491, Bun 1.3.9); only `bun test --timeout` takes effect.

The uv warm-up is one CI step before the suite on Windows: `uv run python -c "print('uv warm')"`. On a fresh runner that call costs 1.2 to 5.0 s on green legs because uv discovers the interpreter and creates its cache; the first uv-backed test was paying that inside its own 5 s budget.

## Validation

- `bun run test ./scripts/bun-test-command.test.ts` — 2 pass — proves the budget is added on `win32` only and selectors pass through verbatim.
- `bun run test ./scripts/test-inventory.test.ts` — 6 pass — proves every tracked test file in cli and server is still collected by the new `testGroups`, and the root plan still runs the runner.
- `bun run test src/commands/version.test.ts` in `packages/cli` and `bun run test src/routes/api.health.test.ts` in `packages/server` — pass through the shared runner.
- `bun run lint`, `bun run format:check` — clean.
- Proof runs: ten `workflow_dispatch` runs of this exact commit (`2945bbf1f`) on `windows-latest`, uninstrumented: **10 of 10 green, 0 timeouts**, Windows legs 8.5 to 10.5 min. Plain dev earlier today was 1 of 10 green-with-timeout and the instrumented arms 3 to 5 of 10. In two of the ten runs a default-budget test finished between 5 s and 20 s and passed (`generate-bundled-defaults: untracked-file guard` at 6.7 s and 5.6 s, one of the original #3294 victims), which is the budget doing exactly its job; every other pass above 5 s was a test that already carries an explicit longer budget. The uv warm-up step ran in all ten. Run IDs: 34604776896 34604779758 34604783011 34604785828 34604788355 34604791839 34604794621 34604797972 34604800995 34604803628.
- **Not verified:** the exact rate at which a runner minute exceeds 20 s; the proof runs bound it, they do not eliminate it.

## Links

- Closes #3294
- Closes #2306
- Related #2924 (residual rule), #3286, #3287, #3282

https://claude.ai/code/session_01QDVDEWqYZdkqKaLcBhJ5ut

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test execution reliability on Windows with consistent platform-appropriate timeouts.
  - Reduced delays during the first test run on fresh Windows environments by preparing the runtime in advance.

- **Chores**
  - Standardized test execution across CLI, server, and repository test suites.
  - Added validation to ensure consistent test command behavior across supported operating systems.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -140,6 +140,16 @@ jobs:
       # drive root, recursively, so nothing under either was being scanned to
       # begin with. `Add-MpPreference` also cost 5s per run to change nothing.
       # Evidence: https://github.com/coleam00/Archon/pull/2943
+      # The first `uv run` on a fresh windows-latest VM costs 1.2 to 5.0 s (interpreter
+      # discovery and cache creation), measured on green legs. Paying it inside the first
+      # uv-backed test's own budget made that test a budget-edge case by itself, so it is
+      # paid here instead. Evidence on coleam00/Archon#3294.
+      - name: Warm uv
+        if: runner.os == 'Windows'
+        run: uv run python -c "print('uv warm')"
+
+      # The Windows leg runs with a 20 s default per-test budget (scripts/bun-test-command.ts)
+      # as an attributed runner-floor residual; see that file and coleam00/Archon#3294.
       - name: Run tests
         run: bun run test
 
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "cli": "bun src/cli.ts",
-    "test": "bun test src/commands/version.test.ts src/commands/setup.test.ts src/commands/skill.test.ts src/commands/doctor.test.ts src/commands/telemetry.test.ts && bun test src/commands/workflow.test.ts && bun test src/commands/isolation.test.ts && bun test src/commands/chat.test.ts && bun test src/commands/serve.test.ts && bun test src/commands/ai.test.ts && bun test src/commands/validate.test.ts && bun test src/cli.test.ts src/help.test.ts src/default-cwd-canonicalization.test.ts src/utils/stdout.test.ts src/utils/safe-console.test.ts && bun test src/utils/detached-run-control.test.ts && bun test src/utils/detached-run-control.integration.spec.ts && bun test src/telemetry-exit.integration.spec.ts && bun test src/commands/workflow-terminal-event.integration.spec.ts && bun test src/commands/workflow-continuation-provider.integration.spec.ts && bun test src/commands/workflow-wait.integration.spec.ts && bun test src/commands/workflow-transcript.integration.spec.ts && bun test src/commands/workflow-node-summaries.spec.ts && bun test src/adapters/cli-adapter.test.ts",
+    "test": "bun run ../../scripts/package-tests.ts",
     "test:integration": "bun test src/utils/detached-run-control.integration.spec.ts && bun test src/telemetry-exit.integration.spec.ts && bun test src/commands/workflow-terminal-event.integration.spec.ts && bun test src/commands/workflow-continuation-provider.integration.spec.ts && bun test src/commands/workflow-wait.integration.spec.ts && bun test src/commands/workflow-transcript.integration.spec.ts",
     "type-check": "bun x tsc --noEmit"
   },
@@ -26,5 +26,66 @@
   },
   "devDependencies": {
     "@types/node": "^22.0.0"
-  }
+  },
+  "testGroups": [
+    [
+      "src/commands/version.test.ts",
+      "src/commands/setup.test.ts",
+      "src/commands/skill.test.ts",
+      "src/commands/doctor.test.ts",
+      "src/commands/telemetry.test.ts"
+    ],
+    [
+      "src/commands/workflow.test.ts"
+    ],
+    [
+      "src/commands/isolation.test.ts"
+    ],
+    [
+      "src/commands/chat.test.ts"
+    ],
+    [
+      "src/commands/serve.test.ts"
+    ],
+    [
+      "src/commands/ai.test.ts"
+    ],
+    [
+      "src/commands/validate.test.ts"
+    ],
+    [
+      "src/cli.test.ts",
+      "src/help.test.ts",
+      "src/default-cwd-canonicalization.test.ts",
+      "src/utils/stdout.test.ts",
+      "src/utils/safe-console.test.ts"
+    ],
+    [
+      "src/utils/detached-run-control.test.ts"
+    ],
+    [
+      "src/utils/detached-run-control.integration.spec.ts"
+    ],
+    [
+      "src/telemetry-exit.integration.spec.ts"
+    ],
+    [
+      "src/commands/workflow-terminal-event.integration.spec.ts"
+    ],
+    [
+      "src/commands/workflow-continuation-provider.integration.spec.ts"
+    ],
+    [
+      "src/commands/workflow-wait.integration.spec.ts"
+    ],
+    [
+      "src/commands/workflow-transcript.integration.spec.ts"
+    ],
+    [
+      "src/commands/workflow-node-summaries.spec.ts"
+    ],
+    [
+      "src/adapters/cli-adapter.test.ts"
+    ]
+  ]
 }
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "bun --watch src/index.ts",
     "start": "bun src/index.ts",
-    "test": "bun test src/scripts/generate-api-types.test.ts && bun test src/routes/api.workflows.test.ts && bun test src/routes/api.conversations.test.ts && bun test src/routes/api.codebases.test.ts && bun test src/routes/api.messages.test.ts && bun test src/routes/api.health.test.ts && bun test src/routes/api.workflow-runs.test.ts && bun test src/routes/api.providers.test.ts && bun test src/services/workflow-resume-service.test.ts && bun test src/adapters/web/transport.test.ts && bun test src/adapters/web/persistence.test.ts && bun test src/adapters/web/truncate.test.ts && bun test src/adapters/web.test.ts && bun test src/adapters/headless.test.ts && bun test src/adapters/web/dashboard-event-poller.test.ts && bun test src/adapters/web/pg-notify-listener.test.ts && bun test src/adapters/web/workflow-bridge.test.ts && bun test src/resolve-user-id.test.ts && bun test src/boot/claude-auth-posture.test.ts && bun test src/github-auth-bootstrap.test.ts && bun test src/routes/auth-poll-status.test.ts && bun test src/auth/config.test.ts && bun test src/auth/instance.test.ts && bun test src/routes/api.auth.test.ts && bun test src/routes/api.provider-keys.test.ts && bun test src/routes/api.user-ai-prefs.test.ts && bun test src/routes/webhooks.test.ts && bun test src/discord-mention.test.ts",
+    "test": "bun run ../../scripts/package-tests.ts",
     "type-check": "bun x tsc --noEmit",
     "generate:api-types": "bun src/scripts/generate-api-types.ts",
     "check:api-types": "bun src/scripts/generate-api-types.ts --check",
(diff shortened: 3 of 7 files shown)


Wirasm:
## Needs fixes

This PR centralizes Windows `bun test` timeout handling in one new helper, `bunTestCommand`, and migrates the cli and server packages onto the same `testGroups` convention every other package already uses. The core mechanism is sound and well-evidenced: the Windows runner-floor attribution on issue #3294 is real (50 instrumented runs, corroborated by an independent data point from PR #3291), the `uv` warm-up step is correctly scoped to Windows only, and the `testGroups` migrations for `packages/cli` and `packages/server` preserve the exact set, order, and process-grouping of the chains they replace. All three reviewers — independently, in the same session — converged on one finding: the new helper's own docstring claims to be "the one place a `bun test` invocation is assembled," but one call site in the same file the PR edits still builds the command by hand, silently bypassing the budget it exists to add.

**1 blocking · 0 non-blocking**

**Validation:** lint, format:check, type-check all pass in the review checkout; `bun-test-command.test.ts` (2 pass) and `test-inventory.test.ts` (6 pass) pass; one representative test each through the migrated cli and server `testGroups` runners passes.

### Findings

|---|---|---|---|
| `R1` | Important | `runRequested()` in `scripts/repo-tests.ts` bypasses the new `bunTestCommand` helper, so `bun run test <path>` gets no Windows timeout budget | OPEN |

<details>
<summary><code>R1</code> — <code>repo-tests.ts</code>'s <code>runRequested</code> path builds <code>bun test</code> by hand, skipping the new Windows budget</summary>

**Impact:** `scripts/bun-test-command.ts` states in its own doc comment: "The one place a `bun test` invocation is assembled, so the per-test budget has one owner." That claim is false as shipped. `scripts/repo-tests.ts:160`, inside `runRequested()` — the code path for `bun run test <selector>`, which the file's own error message documents as supported usage ("run `bun run test <selector>` from the package directory the selector belongs to") — builds `['bun', 'test', ...args]` directly instead of calling `bunTestCommand(args)`. On Windows this silently reverts to Bun's 5 s default timeout instead of the 20 s floor the PR exists to establish, reopening the exact runner-floor flakiness class attributed on #3294, on a path a contributor debugging a flaky Windows test would naturally use.

**Evidence:** `scripts/repo-tests.ts:160` (`run(['bun', 'test', ...args], owner.cwd)`) vs. the sibling fix two lines away at `scripts/repo-tests.ts:109` (`run(bunTestCommand(step.selectors), REPO_ROOT)` inside `runPlan()`) and `scripts/package-tests.ts:34` (`Bun.spawn(bunTestCommand(args), ...)`), both correctly migrated. `.github/workflows/test.yml`'s only Windows invocation is bare `bun run test` (line 154), which takes `runPlan()`, so CI itself does not exercise the bypassed path today. `scripts/repo-tests.test.ts` tests only `planRequestedRuns`'s routing, never the actual command `runRequested` spawns, so nothing would catch a regression here.

**Required outcome:** `scripts/repo-tests.ts:160` reads `run(bunTestCommand(args), owner.cwd)`, matching the two call sites already migrated in this PR. A one-line, same-PR fix — no behavior change on non-Windows platforms.

**Found by:** `prp-core:code-reviewer`, `prp-core:seam-analyzer`, `prp-core:code-simplifier` (all three, independently)

**Disposition:** OPEN. Not CI-reachable today (the Windows job never passes a path argument), so it is not a validation-gate failure, but it falsifies the single-owner invariant the PR's own new module claims, on a documented and supported invocation path, in the very file the PR edits — repository guidance calls this "fix drift on the path you touch," not a pre-existing issue to defer.

</details>

<details>

#### Reviewer coverage

| Scope | Result |
|---|---|
| `code` | R1 |
| `seams` | R1 (plus a noted but non-blocking pre-existing sibling: `packages/cli/package.json`'s untouched `test:integration` script is the same raw-chain pattern, not wired into any CI workflow, out of this PR's diff) |
| `simplify` | R1 |

#### Validation

|---|---|---|
| `bun run lint` | PASS | test-cleanup drift check and lint clean across all packages and `scripts/**` |
| `bun run format:check` | PASS | `bun x prettier --check .` — all files match Prettier style |
| `bun run type-check` | PASS | every workspace package (`paths`, `git`, `docs-web`, `providers`, `web`, `isolation`, `workflows`, `core`, `adapters`, `server`, `cli`) exits 0, plus `scripts/`, `.archon/scripts/`, `.archon/workflows/` tsconfigs |
| `bun run test ./scripts/bun-test-command.test.ts` | PASS | 2 pass, 100% line/function coverage on the new file |
| `bun run test ./scripts/test-inventory.test.ts` | PASS | 6 pass — proves every tracked cli/server test file is still collected by the new `testGroups` |
| `bun run test src/commands/version.test.ts` (packages/cli, via migrated runner) | PASS | 4 pass |
| `bun run test src/routes/api.health.test.ts` (packages/server, via migrated runner) | PASS | 20 pass |

</details>

Issue #3155, opened 2026-09-02
feat(workflows): a concluded CI check signals the run's open event wait instead of being polled


Wirasm:
## Problem

The deliver tail waits for CI by polling. A probe script reads check state; when it is pending, a `wait: duration_ms: 300000` inside a `loop_group` bounded at 13 iterations sleeps and probes again (`.archon/workflows/sdlc/deliver/archon-deliver.yaml:355-387`, `:479-497`). The engine already has the primitive that makes polling unnecessary: `wait: event:` with a deadline and `POST /api/workflows/runs/{runId}/signal` (#2744). Nothing produces the signal. The webhook route handles no `check_suite`, `check_run`, or `workflow_run` events, and no forge adapter maps a check conclusion onto a waiting run.

Inventory: https://github.com/coleam00/Archon/issues/2807#issuecomment-5512134275, class 5.

## Why

A concluded check is an event the forge already pushes. Polling turns it into up to an hour of five-minute probes per delivery, a per-probe `gh` call with its own parsing and exit-code interpretation (`check-ci.py:37-56`), and an iteration bound standing in for a deadline.

## Why now

After forge slice 1 (#3041) rewires the probe off ambient `gh` calls, the wait side is the remaining half of the CI gate. Settling it before slice 2 (#3042) keeps the probe as the fallback rather than the design.

## Desired outcome

A run waiting on `event: checks.complete` for a pull request it owns resumes when the forge reports that pull request's checks concluded, with the conclusion in the wait's payload. The probe-and-sleep loop becomes the fallback for installs without inbound webhooks.

## Invariants

- Core stays forge-agnostic. The forge adapter or plugin (direction.md §community-plugins, #3040) normalizes the inbound event; core receives a run id, an event name, and a payload.
- A signal names one wait occurrence (`metadata.wait.resumeAt`), so a late event cannot satisfy a later wait for the same name. Existing guarantee, unchanged.
- Matching a check event to a run uses the run's recorded, forge-qualified pull request identity (#3042, #3117), never a branch name.
- A missing webhook never blocks: the deadline still expires the wait, and the probe path still works.

## Acceptance

- [ ] A check-concluded webhook for a run-owned pull request resumes that run's open event wait, with the conclusion in the payload.
- [ ] The event-to-run mapping lives on the forge seam. No forge-specific types in core.
- [ ] A signal for the wrong run, or after the wait expired, does nothing (covered by the existing signal tests; extend, do not duplicate).
[...]

Issue #2791, opened 2026-08-24
design(architecture): the core engine stands alone as an SDK — core+CLI installs, optional adapters, pluggable hosts


Wirasm:
## Problem

Archon's engine is more capable than its packaging admits. The core (runs, DAG execution, governance, persistence, durable waits) is entangled with specific hosts and bundled surfaces: you cannot install the core and the CLI alone, every adapter ships whether you use it or not, and some engine capabilities only work when the HTTP server process is running. The engine should be a thing you can build on; today it is a thing you deploy whole.

Fresh example of the seam this creates: durable waits (#2744, PR #2779) persist their clock in core-owned tables and expose the due-scan and claim functions from `packages/core/src/db/workflows.ts`, but the orchestration that actually resumes a due run lives in `packages/server/src/services/workflow-resume-service.ts`. A CLI-only install on a laptop has fully durable pauses and no way to wake them; the capability is core, the host is welded to the server.

## Desired outcome

- **Core stands alone as an SDK.** The engine's public surface (define, validate, run, resume, govern, query) is importable and usable as a library, without adopting Archon's server, web UI, or chat adapters.
- **Core + CLI is a complete install.** A user can install just the core and the CLI and have a working local Archon: run workflows, pause and resume durably, manage runs. No resident server required for the capabilities that do not inherently need a listener.
- **All adapters are optional.** Platform adapters (Slack, Telegram, GitHub, Discord, web) and provider adapters are add-ons a user opts into, including third-party ones: someone who wants to add their own provider adapter can, against a stable seam, without forking the engine.
- **Engine capabilities get pluggable hosts, not privileged ones.** Things like the continuation waker are core functions any host can drive: the server on an interval, the CLI as a verb, an OS timer on a laptop. The server remains the batteries-included host, not the definition of the feature.

## What already points the right way

The architecture is closer to this than the packaging is. `@archon/workflows` takes its database, AI, and config through `WorkflowDeps` injection. Providers register through `ProviderRegistration` with `builtIn: false` as the community pattern (Pi ships this way today), and credentials derive from registrations rather than hand-lists. Platform adapters sit behind `IPlatformAdapter`. The gaps are at the edges: `@archon/core` hard-depends on `@archon/providers`, the CLI pulls in `@archon/server` and `@archon/adapters` for the serve command, and capability orchestration like the resume service lives in the server package while its data machinery lives in core.

## Relations

[...]

PR #3316, opened 2026-09-11
fix(cli): a CLI-owned run enforces its own durable wait deadline

### LINKED ISSUE #3312: fix(workflows): a wait deadline is enforced only by the server scan, so a CLI-only detached run stalls at the pack's CI wait
## Problem

A `wait:` node persists its deadline as `metadata.wait.resumeAt` and pauses the run; the executor completes the wait only when a resume finds the deadline passed (`isDue`) or the event signaled (`packages/workflows/src/dag-executor.ts`, the wait node's pause path). The only thing that resumes a due wait on its own is `scanDueWorkflowContinuations` in `packages/server/src/services/workflow-resume-service.ts`, which runs inside `archon serve`. The authoring guide says so: "The server must be running for scheduled or event-driven continuation."

A detached CLI run has no server. Its owner process exits at the pause (`dag_workflow_finished` in the detached log), and nothing ever calls resume. The SDLC pack's deliver tail now waits on CI through `wait: { event: checks.complete, deadline_ms: 300000 }` (#3155, adopted in #3296), so on a CLI-only install every delivery whose checks are still pending at the first probe stalls at `await-checks.ci-pause` forever, with the run reported `paused` and no attention raised.

Observed 2026-09-11 on two dogfood runs launched with `archon workflow run archon-ship --detach` and no server:

- `dfd25fef23ffb867014f1803735ed8a0` (delivering #2978 as PR #3308): paused at 13:31Z with `resumeAt` 13:36Z; still paused at 13:43Z with the PR's checks all green. `archon workflow resume` completed the expired wait, the loop re-probed, and the run re-entered the same wait with a fresh deadline.
- `c188adaa91ad2231708be741507f0365` (delivering #3303 as PR #3309): same shape.

`archon workflow wait <run-id>` did not return for either, because a durable wait is not a human decision, which is correct; the consequence is that an operator watching through the documented command sees nothing.

## Why it matters

#3155's invariant reads "a missing webhook never blocks: the deadline still expires the wait." It holds only while the server runs. The core is meant to be usable from the CLI alone (AGENTS.md: keep the core self-hostable; the CLI is the primary run path for maintainers), and the pack the CLI ships now depends on a primitive that the CLI cannot complete. Before #3296 the same loop used `wait: duration_ms`, which has the same owner and the same gap; it went unnoticed because those runs happened to be watched by a running server or to conclude on the first probe.

## Why now

The first CLI-only runs through the new CI wait hit it the same afternoon it shipped, on the pack's own dogfood.

## Desired outcome

A detached run's time or event wait reaches its deadline without a server. Either the process that owns the run stays alive to enforce its own deadline (it already owns the run's lifecycle until terminal), or the CLI has an owner for due continuations equivalent to the server scan. Which of the two is the design decision; the invariant is that the answer never depends on a process the operator did not start.

## Invariants

- A wait that is due resumes through the same DAG resume path the server uses today; no second implementation of expiry.
- The run's owner remains distinguishable from an orphan (AGENTS.md: do not guess lifecycle ownership). Whatever enforces the deadline must not mark a run it does not own.
- The attention wait keeps having no deadline and keeps resuming only on an explicit operator action.
- With a server running, behaviour is unchanged.

## Acceptance

- Seen red first: a detached run with `wait: { event: e, deadline_ms: 1000 }` and no server stays paused past its deadline on `dev`.
- After: the same run completes the wait as `expired` within a bounded delay of the deadline and continues, with no server process.
- The same for `wait: duration_ms`.
- `archon workflow wait` on such a run returns at the run's terminal state, not at the wait.
- The authoring guide's sentence about the server is rewritten to say what actually holds.

## Evidence

Run ids above; `~/.archon/logs/detached-run-cli-1789129507898-0hzqbu.log` and `…-1789129509734-s8w8iz.log`; `packages/docs-web/src/content/docs/guides/authoring-workflows.md` around the `wait:` section.

Wirasm:
## Problem and outcome

A `wait:` node with `duration_ms`, `until`, or `event` + `deadline_ms` persists its deadline as `metadata.wait.resumeAt` and pauses the run. The only thing that resumed a due wait on its own was `scanDueWorkflowContinuations` inside `archon serve`, so on a CLI-only install a detached run stayed `paused` forever. The SDLC pack's CI wait (`#3155`, adopted in `#3296`) hit this on every delivery whose checks were still pending on the first probe.

- **Outcome:** The process that owns a run enforces the run's own `time`/`event` wait deadline. A foreground `archon workflow run` and the child started by `--detach` stay with the run, sleep until `metadata.wait.resumeAt`, and re-execute it through the same continuation path `archon workflow resume` uses — no server required. `archon workflow wait` on such a run returns at the run's terminal, not at the pause.
- **Invariant:** The pause stays persisted and the run stays `paused`; the expiry decision is still the single `isDue`/`isSignaled` check in `executeWaitNode`. No second implementation of expiry, and a run's owner stays distinguishable from an orphan. `attention` waits keep no deadline and resume only on explicit operator action.
- **Scope boundary:** With a server running, behaviour is unchanged. Container-isolated workflows still refuse durable waits. `--detach` remains fire-and-forget for the launcher, and `archon workflow cancel`/`abandon` semantics are untouched.
- **Root cause:** Nothing in the CLI held the run after the pause. The executor persisted the wait and returned `paused`; the owner process exited, and only the server scan ever flipped a due row back to `running`.

## Review guidance

- **Feedback requested:** Correctness of the ownership/race handling and the loop's lifetime semantics. In particular, whether stopping quietly on a lost race is the right behavior, and whether the resumed segment should keep the detached owner's active-stop lease.
- **Start here:** `packages/cli/src/commands/workflow.ts:3593` — `workflowRunCommand` now drives a non-recursive loop: run an attempt, and if it paused on a durable `time`/`event` wait, sleep and resume.
- **Review order:**
  1. `packages/cli/src/commands/workflow.ts:3406` — the pause branch reads the row and returns the pending cursor for a `time`/`event` wait only.
  2. `packages/cli/src/commands/workflow.ts:3476` — `pendingDurableWait`, the ownership cursor (excludes `attention` and non-paused runs).
  3. `packages/cli/src/commands/workflow.ts:3499` — `awaitDurableWaitDeadline`, which re-reads the row every 5 s so a server or operator move ends the loop.
  4. `packages/core/src/services/run-live-owner.ts:66` — new `RunLiveOwnerAlreadyOwnedError`, thrown from the three existing "already owned" sites.
- **Lower-attention areas:** `packages/docs-web/src/content/docs/guides/authoring-workflows.md` (prose correction only); test additions.
- **Known risk or uncertainty:** A foreground `archon workflow run` now blocks through a `time`/`event` wait instead of returning at the pause. This is the intended consequence of the chosen design; `--detach` and `attention` waits are unaffected. The signaled-event resume branch is the one path without direct test coverage (see Validation).

## Solution

The run's own process enforces the deadline; the pause is not removed.

`runWorkflowWithOwnedSource` returns a `PendingWaitContinuation | undefined`. On the paused path it re-reads the run, and when `pendingDurableWait` sees a `time`/`event` wait it returns the cursor instead of exiting. `workflowRunCommand` loops: it runs an attempt, and when an attempt names a pending durable wait it sleeps until due and rebuilds the resume options in the shape `workflowResumeCommand` uses (`hydrateResumableRun` + `resume: true`), then runs again through the same continuation path. The wait stays persisted and the live owner is released between segments, so a server or `archon workflow resume` can still take the run.

Because another host can move the run mid-sleep, `awaitDurableWaitDeadline` re-reads the row every 5 s and stops if the cursor's step or `resumeAt` changed. A resume that loses the compare-and-swap (`WorkflowNotResumableError`) or the live-owner endpoint (`RunLiveOwnerAlreadyOwnedError`) stops quietly rather than reporting a failure. `RunLiveOwnerAlreadyOwnedError` was added to stop on a typed signal instead of matching a message string; `hydrateResumableRun`'s failure wrapper now preserves `cause` so the typed error survives.

`DETACHED_RUN_OWNER_ENV` is read once in `workflowRunCommand` and threaded into each attempt, so a resumed detached segment keeps its active-stop lease and its failure exit code.

## Behavior change

| | Before | After |
|---|---|---|
| **Observable behavior** | A CLI-only run paused at a `time`/`event` wait and stayed `paused` forever. | The owning process wakes the run at `metadata.wait.resumeAt` and it continues without a server. |
| **Failure behavior** | `archon workflow wait` never returned for such a run. | It returns at the run's terminal state; a lost race stops the losing process quietly. |

### User flow

```mermaid
flowchart LR
  subgraph Before
    B1[CLI run hits wait] --> B2[Paused, owner exits] --> B3[Stuck forever]
  end

  subgraph After
    A1[CLI run hits wait] --> A2[Paused, owner sleeps] --> A3[Deadline arrives] --> A4[Owner resumes run]
  end
```

## Architecture

```mermaid
flowchart LR
  A[workflowRunCommand] --> B[runWorkflowWithOwnedSource]
  B ==> C[pendingDurableWait cursor]
  C --> D[awaitDurableWaitDeadline]
  D ==> E[workflowRunCommand resume attempt]
  E --> B
  F[archon serve scan] -.->|orphan recovery, unchanged| E
```

### Changed seams

|---|---|---|
| `workflowRunCommand → runWorkflowWithOwnedSource` | Returns `PendingWaitContinuation` for a durable wait instead of `void`. | `packages/cli/src/commands/workflow.ts:3406` |
| `run-live-owner → resume caller` | Untyped "already owned" `Error` becomes `RunLiveOwnerAlreadyOwnedError`. | `packages/core/src/services/run-live-owner.ts:66`, `workflow.ts:3563` |
| `CLI → persisted wait row` | Owner now reads and re-reads its own `metadata.wait` cursor; server scan untouched. | `workflow.ts:3476`, `workflow.ts:3499` |

## Validation

- `bun run --cwd packages/cli type-check`; `bun run --cwd packages/core type-check` — pass.
- `bun run --cwd packages/cli test src/commands/workflow.test.ts` — 409 pass, 0 fail; 3 new `pendingDurableWait` cases.
- `bun run --cwd packages/cli test src/commands/workflow-wait.integration.spec.ts` — 16 pass, 0 fail; 6 new cases: duration wait satisfied, two sequential waits carried by one owner, event wait expired, `workflow wait` returns at terminal, attention wait releases its owner, abandoned run not resurrected.
- **Red-first proof** — reverted only `workflow.ts` and `run-live-owner.ts` to the base and ran the new block: 4 fail / 2 pass. The failures are the fix-specific cases (duration wait never completes, two sequential waits never advance, event wait never expires, `workflow wait` exits `deadline` with `observedStatus: paused`). The two negative cases pass on base, as they should.
- `bun run lint --max-warnings 0`, `bun run format:check`, `bun run test`, `bun run validate` — all pass.
- **Not verified:** The signaled-event resume branch (a live `/signal` arriving while the owner sleeps) has no direct test; the fixture covers the event-expiry path. `awaitDurableWaitDeadline` is an unexported helper covered end-to-end rather than by unit test, to avoid widening the module surface.

## Delivery considerations

|---|---|---|
| Compatibility / migration | No schema, migration, or generated-artifact change. The persisted pause and `resumeAt` contract are unchanged. | diff touches no schema files |
| Security / permissions / data | None. The new error type carries the endpoint path already present in the old message. | `run-live-owner.ts:66` |
| Rollout / rollback | Revert is a plain code revert; persisted runs remain resumable by the server or `archon workflow resume`. | — |
| Observability | A superseded continuation logs `cli.wait_continuation_superseded`. | `workflow.ts:3627` |
| Documentation / communication | `authoring-workflows.md` now states the owner enforces the deadline, the server scan covers orphaned runs, and `/signal` still needs the server. | `authoring-workflows.md:721` |

## Links

- Closes #3312

## Summary by CodeRabbit

- **New Features**
  - Workflow runs now automatically continue after duration- and event-based waits, including when started from the CLI without a running server.
  - Detached workflow processes remain active through durable waits and resume execution when conditions are met.
  - Workflows whose owning process stops can still be resumed by the server or manually with `workflow resume`.

- **Bug Fixes**
  - Improved handling and reporting of conflicts when multiple processes attempt to own the same workflow run.

- **Documentation**
  - Updated durable-wait guidance to explain process ownership, server fallback, and manual resumption.

### DIFF
--- a/packages/cli/src/commands/workflow-wait.integration.spec.ts
+++ b/packages/cli/src/commands/workflow-wait.integration.spec.ts
@@ -387,6 +387,31 @@ function readNodeCompletedOutput(
 }
 
 /**
+ * Every CLI platform conversation the fixture's database holds.
+ *
+ * An automatic resume that drops the run's conversation id generates a fresh one, so
+ * the resumed segment's dispatch and result land in a second row while the run row
+ * still points at the first. The set is the observable proof that did not happen.
+ */
+function readCliConversationPlatformIds(archonHome: string): string[] {
+  const databasePath = join(archonHome, 'archon.db');
+  if (!existsSync(databasePath)) return [];
+  const database = new Database(databasePath, { readonly: true });
+  try {
+    return database
+      .query<{ platform_conversation_id: string }, []>(
+        "SELECT platform_conversation_id FROM remote_agent_conversations WHERE platform_type = 'cli'"
+      )
+      .all()
+      .map(row => row.platform_conversation_id);
+  } catch {
+    return [];
+  } finally {
+    database.close();
+  }
+}
+
+/**
  * Poll `read` until it produces a value, naming `what` if it never does.
  *
  * The catch is load bearing, not defensive: an owner process creates `archon.db`
@@ -801,10 +826,17 @@ describe('a durable wait deadline is enforced by the owning process', () => {
     'name: wait-long\ndescription: Long durable wait fixture.\n' +
     'nodes:\n' +
     '  - id: cooldown\n    wait:\n      duration_ms: 8000\n';
+  // Two sequential waits: one owner has to carry the run through both, so a loop
+  // that resumes once and stops leaves the run parked at `wait-two`.
+  const TWO_WAITS =
+    'name: wait-twice\ndescription: Two sequential durable waits.\n' +
+    'nodes:\n' +
+    '  - id: wait-one\n    wait:\n      duration_ms: 1500\n' +
+    '  - id: wait-two\n    depends_on: [wait-one]\n    wait:\n      duration_ms: 1500\n';
 
   test('resumes a duration wait at its deadline and finishes the run', async () => {
     const fixture = makeFixture('archon-wait-duration-', { 'wait-duration': DURATION_WAIT });
-    const { runId } = await launchDetached(fixture, 'wait-duration');
+    const { runId, conversationId } = await launchDetached(fixture, 'wait-duration');
 
     const completed = await waitFor(
       'the duration wait to resume and complete the run',
@@ -823,6 +855,35 @@ describe('a durable wait deadline is enforced by the owning process', () => {
     // The engine's own record of the wait: the deadline was genuinely waited, and
     // only then did the owner resume the run.
     expect(Number(output.waited_ms)).toBeGreaterThanOrEqual(3000);
+    // The resumed segment continues the run's original conversation instead of
+    // generating a second one for its dispatch and result card.
+    expect(readCliConversationPlatformIds(fixture.archonHome)).toEqual([conversationId]);
+  }, 120_000);
+
+  test('carries one owner through two sequential waits', async () => {
+    const fixture = makeFixture('archon-wait-twice-', { 'wait-twice': TWO_WAITS });
+    const { runId } = await launchDetached(fixture, 'wait-twice');
+
+    const completed = await waitFor(
+      'both waits to resume and complete the run',
+      () => (readRunStatus(fixture.archonHome, runId) === 'completed' ? 'completed' : undefined),
+      60_000
+    );
+    activeRunIds.delete(runId);
+    expect(completed).toBe('completed');
+
+    const first = await waitFor(
+      'the first wait node completion',
+      () => readNodeCompletedOutput(fixture.archonHome, runId, 'wait-one'),
+      10_000
+    );
+    const second = await waitFor(
+      'the second wait node completion',
+      () => readNodeCompletedOutput(fixture.archonHome, runId, 'wait-two'),
+      10_000
+    );
+    expect(first).toMatchObject({ status: 'satisfied' });
+    expect(second).toMatchObject({ status: 'satisfied' });
   }, 120_000);
 
   test('expires an event wait by its deadline and finishes the run', async () => {
@@ -874,6 +935,15 @@ describe('a durable wait deadline is enforced by the owning process', () => {
   test('leaves an attention wait for the operator and releases its owner', async () => {
     const fixture = makeFixture('archon-wait-attention-', { 'wait-attention': ATTENTION_WAIT });
     const { runId } = await launchDetached(fixture, 'wait-attention');
+
+    // Park first. The control endpoint does not exist until the owner starts, so a
+    // probe before the run is paused would read a not-yet-started owner as a released
+    // one and assert on a `pending` row.
+    await waitFor(
+      'the attention run to park',
+      () => (readRunStatus(fixture.archonHome, runId) === 'paused' ? 'paused' : undefined),
+      30_000
+    );
 
     // An attention wait has no deadline, so its owner must exit rather than sleep.
     // With the endpoint gone nothing in this process can advance the run, which is
--- a/packages/cli/src/commands/workflow-wait.integration.spec.ts
+++ b/packages/cli/src/commands/workflow-wait.integration.spec.ts
@@ -348,6 +348,45 @@ function readRunStatus(archonHome: string, runId: string): string | undefined {
 }
 
 /**
+ * One node's persisted completion output, read straight from the events table.
+ *
+ * A wait node's fixed `{ status, waited_ms }` contract is the engine's own record of
+ * what happened, so it proves more than wall-clock timing: `waited_ms` shows the
+ * deadline was genuinely waited instead of short-circuited, and `status` distinguishes
+ * a satisfied wait from an expired one.
+ */
+function readNodeCompletedOutput(
+  archonHome: string,
+  runId: string,
+  nodeId: string
+): Record<string, unknown> | undefined {
+  const databasePath = join(archonHome, 'archon.db');
+  if (!existsSync(databasePath)) return undefined;
+  const database = new Database(databasePath, { readonly: true });
+  try {
+    const row = database
+      .query<{ data: string }, [string, string]>(
+        `SELECT data FROM remote_agent_workflow_events
+         WHERE workflow_run_id = ? AND step_name = ? AND event_type = 'node_completed'
+         ORDER BY created_at DESC LIMIT 1`
+      )
+      .get(runId, nodeId);
+    if (!row) return undefined;
+    const data = JSON.parse(row.data) as { structured_output?: unknown; node_output?: unknown };
+    const output = data.structured_output ?? data.node_output;
+    const parsed = typeof output === 'string' ? (JSON.parse(output) as unknown) : output;
+    if (typeof parsed !== 'object' || parsed === null) return undefined;
+    return parsed as Record<string, unknown>;
+  } catch {
+    // The owner creates archon.db before applying the schema and Windows can widen
+    // the commit window (#2306); the caller polls, so a transient read is not a result.
+    return undefined;
+  } finally {
+    database.close();
+  }
+}
+
+/**
  * Poll `read` until it produces a value, naming `what` if it never does.
  *
  * The catch is load bearing, not defensive: an owner process creates `archon.db`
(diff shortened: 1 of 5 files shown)


Wirasm:
# Review report — PR #3316 (round 2, continuation)

## 1. Verdict

**Ready — `none`.** Both round-1 Important findings (R1, R2) are fixed at `e3f0c747`; the
correction delta introduced no new blocker, and the two Suggestions still open (R4, R6) do not
block.

## 2. Accepted contract

Carried verbatim from `review/scope.md` (unchanged from round 1).

- **Required outcome.** A `wait:` node with `duration_ms`, `until`, or `event` + `deadline_ms`
  persists its deadline as `metadata.wait.resumeAt` and pauses the run. Only
  `scanDueWorkflowContinuations` inside `archon serve` resumed a due wait, so on a CLI-only install
  a detached run stayed `paused` forever. The chosen design: the process that owns the run keeps the
  run durably paused, stays alive, and resumes it at its own deadline through the existing CLI
  continuation path (`hydrateResumableRun` + `resume: true`), with no server required.
- **Invariants.** The pause stays persisted and the run stays `paused`; the live owner is released
  between execution segments, so a server or `archon workflow resume` can still take the run. The
  expiry decision stays the single `isDue`/`isSignaled` check in `executeWaitNode`. A run's owner
  stays distinguishable from an orphan; the loop resumes only after a fresh row read shows the same
  paused cursor, and the resume itself is the existing CAS. `attention` waits keep no deadline. With
  a server running, behaviour is unchanged. Container-isolated workflows still refuse durable waits.
  `--detach` stays fire-and-forget for the launcher; cancel/abandon semantics are untouched.
- **Explicit non-goals.** Quota continuation; a new CLI daemon/watch command or moving the server
  scan into core (`#2791`); `attention` waits; container-isolated workflows; any doc edit beyond the
  one continuation sentence in `authoring-workflows.md`.

This round's delta is exactly two commits over the cursor: R1's conversation threading and R2's
two-wait coverage, plus the R3/R5 suggestion fixes.

## 3. Reviewed head SHA

`e3f0c747b77886e0cd078c49c2e5581b15cb5a48`

Matches the local checkout HEAD, `scope.md`'s cursor, and `gh pr view 3316` `headRefOid`. Base
`dev`, merge-base `a7e300b2a74051711b7de2262c267cb96d3002d2`.

## 4. Findings

No open Critical or Important findings. This round raised no new finding; the correction delta was
read in full (`git diff 6c686bd0e98baf56953e3268eb110c5ec0acf572..HEAD`, three files, 87 insertions,
10 deletions) and contained no defect that violates the accepted contract.

Rejected findings this round: none. The round-1 `code` lens reported no finding and nothing was
dropped.

### Suggestions (carried, non-blocking)

- **R4** — the paused-wait predicate is hand-synced between `pendingDurableWait` and the server's
  `continuationCursor`. Still open by decision; it brushes the explicit `#2791` non-goal.
- **R6** — a lost live-owner race exits 0 without confirming the run has an owner. Still open; not a
  stuck run, and outside the accepted outcome.

## 5. Prior findings (continuation mode)

The prior report is the accumulated review state. Every finding is carried forward with its stable
ID, first-`sources`, and a per-finding verdict.

| ID | Severity | Sources | Verdict at `e3f0c747` | Evidence checked |
|----|----------|---------|-----------------------|------------------|
| R1 | Important | seams | **Fixed** | `platformConversationId` threads from the pause branch to every rebuilt attempt. |
| R2 | Important | tests | **Fixed** | Two-sequential-waits integration case added; executed here and passes. |
| R3 | Suggestion | simplify | **Fixed** | `DurableWaitCursor.kind` deleted; unit expectations updated. |
| R4 | Suggestion | simplify | **Open** | Deliberately not taken (non-goal `#2791`); still non-blocking. |
| R5 | Suggestion | simplify | **Fixed** | Comment now names `workflowRunCommand` as the caller. |
| R6 | Suggestion | errors | **Open** | Deliberately not taken; still non-blocking. |

### R1 — durable-wait resume dropped the run's platform conversation (fixed)

`runWorkflowWithOwnedSource` now returns the platform conversation it ran the segment under
(`workflow.ts:3416`, `platformConversationId: conversationId`), `PendingWaitContinuation` carries it
(`workflow.ts:3467`), and `buildWaitResumeAttempt` sets `options.conversationId` from it
(`workflow.ts:3529-3550`). `workflow.ts:2384` then resolves the same id instead of generating a new
one. The value is stable across iterations because the resumed attempt re-reads it from
`options.conversationId`, so a run with two waits also stays in one conversation.

Verified by execution: the duration-wait case asserts the fixture's CLI conversation set is exactly
the launch's `conversationId` (`workflow-wait.integration.spec.ts`, the
`readCliConversationPlatformIds` assertion), and it passes. Reverting only the
`conversationId: platformConversationId` line was recorded red-first in `implementation.md`; the
mechanism is direct — without it `options.conversationId` is undefined and
`workflow.ts:2384` calls `generateConversationId()`, creating a second row.

### R2 — owner-continuation loop only exercised for one wait (fixed)

A `TWO_WAITS` fixture (`wait-one` `duration_ms: 1500`, then `wait-two` `depends_on: [wait-one]`,
`duration_ms: 1500`) and a case asserting the run reaches `completed` with both nodes
`{ status: 'satisfied' }` were added (`workflow-wait.integration.spec.ts`). Two sequential waits are
a supported, documented shape (`authoring-workflows.md`'s own Durable-waits example chains three),
so the second iteration is a real path. Verified by execution: the case passes, and a single-pass
loop leaves the run parked at `wait-two` and the 60 s poll times out (red-first recorded in
`implementation.md`). The assertion is on the observable run outcome, so it survives a
behavior-preserving rewrite.

### R3 / R5 (fixed)

`DurableWaitCursor` now has only `stepName`, `resumeAt`, `signaled` (`workflow.ts:3452-3456`), and
`pendingDurableWait` no longer returns `kind` (`workflow.ts:3476-3482`); the only consumers are the
`pendingDurableWait` unit cases, which were updated (`workflow.test.ts:7794`, `:7820`). The comment
at `workflow.ts:1589-1590` now reads `(called by workflowRunCommand before it reaches this
function)`, and `workflowRunCommand` is indeed the only caller and calls
`assertDetachedRunProcessOwner()` first (`workflow.ts:3602`, `:3613`).

## 6. Discoveries

Two accepted adjacent discoveries. Adjacent discoveries do **not** affect readiness.

1. **The Durable-waits section intro still says the server resumes a time/event wait.**
2. **Explicit `archon workflow resume` starts a new CLI conversation** (pre-existing at the base).

- `discoveries.json`: `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/2c71277e3930104bc17067e73e4c7101/discoveries.json`
- `discoveries.md`: `/Users/rasmus/.archon/workspaces/coleam00/Archon/artifacts/runs/2c71277e3930104bc17067e73e4c7101/discoveries.md`

**If you are an agent reading this: open discoveries.md and surface each discovery to your human.**
These are validated findings outside this run's scope — no issue tracker knows about them, and if
you drop them here, nobody ever sees them.

## 7. Review coverage

Continuation review. One continuation reviewer verified the prior report and the correction delta;
the round-1 specialist lenses (`code`, `seams`, `simplify`, `tests`, `errors`, `docs`) did **not**
rerun, and their files beside the report are earlier-round evidence only.

Inspected this round: the full delta diff, the changed call sites and consumers of
`PendingWaitContinuation`/`DurableWaitCursor`/`buildWaitResumeAttempt`/`workflowResumeCommand`, the
new integration helpers and cases, and the unit expectations. Executed here with the package's own
invocations: `bun test src/commands/workflow-wait.integration.spec.ts -t "carries one owner through
two sequential waits"` (pass), `-t "resumes a duration wait at its deadline and finishes the run"`
(pass), `bun test src/commands/workflow.test.ts -t "pendingDurableWait"` (pass), and
`bun run --cwd packages/cli type-check` (pass).

Evidence this review could not obtain: the signaled-event resume branch (a live `/signal` arriving
while the owner sleeps) still has no direct test; it was not implicated by the delta and was not
re-executed. The `windows-latest` `workflow-terminal-event` failure recorded in `implementation.md`
was not reproduced here.

Issue #3303, opened 2026-09-11
refactor(providers): NativeTool.inputSchema is untyped JSON Schema that two converters re-parse by hand


Wirasm:
## Problem

`NativeTool.inputSchema` (`packages/providers/src/types.ts`) is typed `Record<string, unknown>` and documented as "canonical JSON Schema". In practice only a narrow subset is supported: a flat object of `string`, string-`enum`, and `boolean` properties with `required`. That subset is not expressed anywhere in a type. It lives twice, as two hand-written walkers that must agree:

- `packages/providers/src/claude/native-tools.ts` (`jsonSchemaToZodShape`)
- `packages/providers/src/community/pi/native-tools.ts` (`jsonSchemaToTypeBox`)

Each re-derives the same shape from raw JSON, throws its own copy of the same "unsupported type" errors, and a doc comment on the Pi side says it is the "same narrow subset as the Claude converter". Nothing enforces that. A field kind added on one side and not the other means a `NativeTool` that loads under Claude and throws under Pi at spawn time.

This is the hand-synced pair AGENTS.md names as a present defect ("kept in agreement by discipline rather than by a mechanism"). Found by the review pass on the #2424 run, which touched the Pi side for the `StringEnum` fix (PR #3299) and had no way to know whether the Claude side still matched.

## Why it matters, and why now

There is one producer today (`buildManageRunTool` in `packages/core/src/orchestrator/manage-run-tool.ts`), so the pair is currently in sync by luck. Native tools are the mechanism for the per-provider run-management surface, and a second tool or a third provider converter is the moment the drift shows up as a provider-specific runtime failure that no test catches. Cheaper to close while there is one producer and two consumers.

## Desired outcome

The supported input shape is owned by the type, not by two walkers:

- `NativeTool.inputSchema` becomes a typed narrow shape (properties are a discriminated union of the supported field kinds; `required` is a list of those keys). An unsupported kind is a compile error at the producer, not a throw inside a provider.
- Each converter maps that typed shape to its SDK type (Zod, TypeBox). The two "unsupported type" throw paths disappear because the type makes them unreachable.
- One conformance test drives both converters from the same fixture and asserts they accept and reject the same inputs.

## Invariants

- No provider-specific behaviour leaks into `types.ts`; it owns the shape, not the SDK translation.
- `manage_run` keeps its current parameter surface on both providers.
- The Pi side keeps building enums with `StringEnum` (PR #3299) for Google/Vertex compatibility.

## Acceptance

[...]

PR #3322, opened 2026-09-13
feat: support exclusive host-owned tool execution

Algoram99:
## Problem and outcome

An embedding application cannot enforce approval before a protected effect by observing provider-owned tools after execution. This adds an explicit host-owned execution boundary to the existing provider loop, rather than replacing that loop or disguising canonical schemas as strings.

- **Outcome:** A host can supply an exclusive tool set, await its own approval before execution, and receive correlated typed progress and text/image results.
- **Invariant:** Only supplied model-callable tools are active, including an explicitly empty set on resume. Unsupported providers reject the request rather than falling back.
- **Scope boundary:** Pi implements this capability. Ordinary Archon conversations and the existing additive `NativeTool` contract remain unchanged.

## Review guidance

- **Start here:** `packages/providers/src/community/pi/host-tools.ts:15` — the executor and structured-error boundary.
- **Review order:** Public types and unsupported-provider guards; Pi session construction and event bridge; host-owned branches in the core orchestrator; the two real-SDK integration tests.
- **Feedback requested:** Exclusive tool ownership, cancellation, native extension interaction, and completeness of the public host event transport.
- **Lower-attention areas:** The capability matrix is generated by its owner script; the generated-artifact check passed.
- **Known boundaries:** Native extension JavaScript remains trusted, not sandboxed. Hosts own approval and turn-scoped deduplication; this does not promise crash-proof exactly-once effects. The SDK result identity used by the error hook is covered by a real extension-interaction regression.

## Solution

`HandleMessageContext.hostTools` and the provider request contract accept canonical JSON Schema plus an asynchronous host handler. The handler receives the actual tool-call ID, cancellation signal, and progress callback. Pi retains native settings, authentication, guidance, configured extensions, and inference; extension/builtin tool activation cannot widen the host's model-callable set.

Host updates become typed `tool_update` chunks. Final `tool_result.toolResult` preserves content, images, details, and authoritative error state without consumer JSON parsing; `toolOutput` is a text display projection in host mode. Host-owned streaming and batch adapters receive these events through `sendStructuredEvent`, without duplicate provider-formatted tool-call chat text. Ordinary batch event behavior is preserved. A native presentation hook cannot relabel a host denial as successful execution.

Host-owned core turns bypass command dispatch, workflow discovery and implicit source sync, `manage_run`, unavailable paused-gate instructions, and title-agent inference while retaining workspace configuration. Other maintained providers reject the capability with `UnsupportedHostToolsError` before SDK execution. The existing additive schema subset remains separate (related: #3303).

## Validation

Runtime: Bun 1.4.2 and the existing Pi SDK 0.84.4 dependency. Integration inference is a scripted OpenAI-compatible SSE server on loopback, not a live model.

- Real Pi integration: **5 passed** — canonical schemas on the backend wire, exclusive tools, denial/error preservation, native guidance/extensions, and pre-aborted no-inference behavior.
- Real public core lifecycle regression across both stream and batch paths: **1 passed** — denial without an effect, one approved write, tool removal on resume, cancellation, original result IDs, typed progress, image content, and structured metadata.
- Disposable real streaming smoke: **6 inference requests, 1 approved effect**; all four terminal call IDs and outcomes preserved, typed progress/image results retained, cancellation observed. The smoke script and fixtures were removed afterward.
- Ordinary native batch smoke: **2 requests, one real native write, zero structured events**, preserving the pre-existing batch contract. The script and fixtures were removed.
- Existing targeted regressions: **289 core orchestrator**, **101 Pi provider**, and **57 native event-bridge** tests passed.
- `bun run type-check`, `bun run lint`, and `bun run format:check` passed after the final transport changes.
- `bun run validate` passed all static/install checks, then its fail-fast test runner stopped on an obsolete implementation-only assertion that a pre-aborted request must create and abort a session. That assertion was removed; the real no-inference regression remains. Completed groups were retained and the failed/interrupted/unstarted groups were continued using the existing `bunTestCommand` and `ROOT_TEST_PLAN` owners. The complete canonical inventory finished with **9,616 passed, 0 failed, 6 skipped**, across **180 workspace groups and 2 root groups**. This was not one uninterrupted successful `validate` invocation. Subsequent transport and review repairs, including one additional host-gate regression, received the targeted checks and smoke verification listed above.
- **Not verified:** Installed embedding-application approval UI acceptance or live-model inference. Those are downstream integration checks, not claims of this upstream contract change.

## Summary by CodeRabbit

* **New Features**
  * Added support for host-supplied tools during AI conversations, including progress updates, results, errors, cancellation, and preserved tool-call details.
  * The Pi provider supports exclusive host tools; combining them with native tools is rejected.
  * Already-cancelled requests stop before inference or session setup.

* **Bug Fixes**
  * Providers without host-tool support now reject these requests clearly and before execution.

* **Documentation**
  * Documented host-owned tool execution and updated provider capability information.


Wirasm:
Thanks for the careful work. The real-SDK integration tests and the validation notes made this easy to evaluate. We're not going to merge it in this form, though. The reasons are about where the capability lives, not how it's built:

1. **It turns the conversation layer into a general agent loop.** A host-owned turn skips command dispatch, workflow discovery, `manage_run`, paused-gate handling and title generation. What's left is "run Pi with the host's tools", with Archon in between adding nothing it governs. `.archon/direction.md` positions Archon as a governed workflow engine, not a general-purpose chat or agent host; chat surfaces exist to run workflows. An application that needs exclusive host tools around a single agent loop is probably better served by the Pi SDK directly.
2. **It threads a second mode through one function.** The `hostOwnedTurn` checks spread through `orchestrator-agent.ts` make every future change to that path reason about both modes.
3. **It widens the provider request contract ahead of its design.** `hostTools` becomes part of what every provider receives, and every provider gains a refusal path. We're about to define the minimal agent-runtime contract providers join through (#2848), with providers moving towards plugins (`direction.md §community-plugins`). New request capabilities should be decided there, since every one becomes something each third-party runtime has to handle.

The underlying idea, tools the host executes only after its own approval rather than tools Archon observes after the fact, is a genuine governance idea. I've recorded it on #2848 as an idea for the runtime contract discussion. If you can share what your embedding application needs (approval UX, which effects, whether workflows are involved), that would help shape it.

Wirasm:
Closing, per the comment above. The host-executed-tools-with-approval idea is recorded on #2848 for the provider runtime contract. If you share what your embedding application needs, we can use it there.

